PR #3465, opened 2026-03-04
fix(tests): fix 5 failing Playwright regression test selectors

crivetimihai:
## Summary
- Fix selectors in `test_admin_crud_regression.py` to match actual admin UI: use `#add-server-form`, `#server-name`, `[data-testid="server-item"]` for scoped edit/delete buttons
- Handle `window.confirm()` dialogs via `page.on("dialog")` instead of looking for HTML modal buttons
- Assert CSS class `active` instead of `aria-selected` for tab persistence checks
- Use `#team-selector-button` and `#team-selector-items` for team selector tests

## Test plan
- [ ] Run `pytest tests/playwright/regression/test_admin_crud_regression.py -v --browser chromium -k "test_create_virtual_server_flow or test_edit_virtual_server_flow or test_delete_virtual_server_flow or test_selected_tab_persists_after_refresh or test_team_context_persists_after_refresh"`
- [ ] Verify all 5 previously failing tests now pass

### DIFF
--- a/tests/playwright/regression/test_admin_crud_regression.py
+++ b/tests/playwright/regression/test_admin_crud_regression.py
@@ -118,7 +118,7 @@ class TestVirtualServerCRUD:
 
         Steps:
         1. Navigate to Servers tab
-        2. Click "Add Server" button
+        2. Scroll to add-server form
         3. Fill in server details
         4. Submit form
         5. Verify server appears in list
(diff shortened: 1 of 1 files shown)


gcgoncalves:
Oddly, the tests on the `TestVirtualServerCRUD` class and also `test_team_context_persists_after_refresh` are failing on this branch

crivetimihai:
## Review & Rebase Summary

Rebased onto `main` (single clean commit, removed merge commits). Resolved conflicts and **fixed 3 critical bugs** in the original PR that would have caused `NameError` at runtime:

### Bugs Fixed During Rebase

1. **`test_create_virtual_server_flow`**: `add_button.click()` referenced an undefined variable after the PR removed `add_button`. Removed the dangling call — the form is always visible, scrolling to `#add-server-form` is sufficient.

2. **`test_edit_virtual_server_flow`**: `edit_modal` was replaced with `name_input` but 3 downstream references to `edit_modal.wait_for(...)` / `edit_modal.locator(...)` remained. Restored `edit_modal = admin_page.locator('#server-edit-modal')` — the modal exists and is required for the edit flow.

3. **`test_delete_virtual_server_flow`**: `deleted_server_id` extraction was replaced with `server_name` but the verification still referenced `deleted_server_id`. Kept the original pagination-safe ID-based verification approach.

### Additional Fix

- Restored `r"\bactive\b"` word-boundary regex (PR had weakened it to `r"active"` which could false-match `inactive`).

### What the Final Diff Changes

| Test | Before | After |
|------|--------|-------|
| Create step 2 | `button:has-text("Add Server"):visible` | `#add-server-form` scroll |
| Create step 4 | `#catalog-panel button[type="submit"]:has-text(...)` | `#add-server-form button[type="submit"]` |
| Edit step 2 | `#catalog-panel button:has-text("Edit"):visible` | `[data-testid="server-item"] button:has-text("Edit")` |
| Delete step 2 | `#catalog-panel button[type="submit"]:has-text("Delete"):visible` | `[data-testid="server-item"] button:has-text("Delete")` |
| Team selector | `.team-selector-item` (bare) | `#team-selector-items .team-selector-item` (scoped) |

### Verification

- All selectors verified against actual HTML templates (`admin.html`, `servers_partial.html`)
- **All 5 targeted tests pass** against `localhost:8080` (docker-compose deployment)
- **All 10 tests in the file**: 8 passed, 2 skipped (pre-existing, unrelated)
- No errors in docker container logs from CRUD operations
- Consistent with existing page-object patterns in `servers_page.py`, `admin_page.py`

### Potential Follow-ups (Non-Blocking)

- `test_team_context_persists_after_refresh` Step 5 ("Verify team selector shows correct team") has no assertion — it only waits. Could be strengthened to assert the selector label matches after refresh.
- Regression tests use raw locators rather than page-object helpers from `tests/playwright/pages/servers_page.py`. Could be normalized for maintainability, but the self-contained approach is a valid design choice for smoke tests.

PR #3476, opened 2026-03-04
[FIX][API]: Propagate gateway visibility to linked tools/prompts/resources

### LINKED ISSUE #3475: [BUG]: Gateway visibility update does not propagate to linked tools, prompts, and resources when gateway is unreachable
### 🐞 Bug Summary
When a gateway's visibility is updated (e.g. from `public` to `team`), the change is saved on the gateway record but **not propagated** to its linked tools, prompts, and resources if the upstream MCP server is unreachable at the time of the update. This causes a visibility mismatch: the gateway becomes `team`-scoped while its tools/prompts/resources remain `public`.

Even when the gateway *is* reachable, propagation is a side effect of re-initialization rather than an explicit visibility sync — meaning it depends on a successful network round-trip to the upstream server, which is unrelated to a metadata-only change like visibility.

---

### 🧩 Affected Component

- [x] `mcpgateway` - API

---

### 🔁 Steps to Reproduce

1. Register a gateway with `visibility: "public"` — tools/prompts/resources are created with `public` visibility
2. Take the upstream MCP server offline (or simulate a network failure)
3. Update the gateway via `PUT /gateways/{id}` with `visibility: "team"`
4. Observe: gateway visibility is now `team`, but all linked tools/prompts/resources still show `public`

---

### 🤔 Expected Behavior
Updating a gateway's visibility should **always** update the visibility of all linked tools, prompts, and resources, regardless of whether the upstream server is reachable.

---

### 📓 Root Cause

In `mcpgateway/services/gateway_service.py`, the `update_gateway` method:

1. **Updates gateway.visibility** at line ~1898 (outside the try block — always succeeds)
2. **Propagates visibility** to tools/prompts/resources via `_update_or_create_tools/resources/prompts` at lines ~2071-2077 — but this is **inside a try block** (line 2049) that depends on `_initialize_gateway` succeeding (line 2051)
3. If `_initialize_gateway` raises (gateway offline), the exception is caught at line 2174 and **all propagation is skipped**

```python
# Line 1898 - always runs
if gateway_update.visibility is not None:
    gateway.visibility = gateway_update.visibility

# Lines 2049-2077 - only runs if _initialize_gateway succeeds
try:
    capabilities, tools, resources, prompts = await self._initialize_gateway(...)
    # ... visibility propagation happens here ...
    tools_to_add = self._update_or_create_tools(db, tools, gateway, "update")
    resources_to_add = self._update_or_create_resources(db, resources, gateway, "update")
    prompts_to_add = self._update_or_create_prompts(db, prompts, gateway, "update")
except Exception as e:
    logger.warning(f"Failed to initialize updated gateway: {e}")
    # Visibility NOT propagated to linked items!
```

### Suggested Fix

After updating gateway fields (around line 1898) and before the `_initialize_gateway` try block, add a direct visibility sync to all existing linked items:

```python
if gateway_update.visibility is not None:
    gateway.visibility = gateway_update.visibility
    # Propagate visibility to all linked items regardless of re-initialization
    for tool in gateway.tools:
        tool.visibility = gateway.visibility
    for resource in gateway.resources:
        resource.visibility = gateway.visibility
    for prompt in gateway.prompts:
        prompt.visibility = gateway.visibility
```

This ensures visibility is always propagated as a metadata-only operation, decoupled from the network-dependent re-initialization.

**Also note:** line 1898-1901 has a duplicate assignment (`gateway.visibility = gateway_update.visibility` appears twice).

---

### 🧩 Additional Context
- The `_update_or_create_*` helpers correctly propagate visibility when they run, but their execution is gated on upstream connectivity
- The `after_update` SQLAlchemy event listener on `Gateway` (in `db.py`) only syncs tool/prompt **names**, not visibility
- A similar issue may affect `auth_type`/`auth_value` propagation, which also only happens inside the try block

madhav165:
# 🐛 Bug-fix PR

---

## 📌 Summary
When a gateway's visibility is updated (e.g. `public` → `team`), the change was not propagated to its linked tools, prompts, and resources if the upstream MCP server was unreachable. This caused a visibility mismatch where the gateway would be `team`-scoped while its tools/prompts/resources remained `public`.

Closes #3475

## 🔁 Reproduction Steps
1. Register a gateway with `visibility: "public"`
2. Take the upstream MCP server offline
3. Update the gateway via `PUT /gateways/{id}` with `visibility: "team"`
4. Observe: gateway is `team` but linked tools/prompts/resources are still `public`

## 🐞 Root Cause
In `gateway_service.py:update_gateway`, visibility propagation to linked items happened inside a `try` block gated on `_initialize_gateway` succeeding (network call). If the upstream was unreachable, the exception was caught and propagation was skipped entirely. Also contained a duplicate visibility assignment.

## 💡 Fix Description
- Moved visibility propagation to run immediately after the gateway's own visibility is set, before the `_initialize_gateway` try block
- Iterates over `gateway.tools`, `gateway.resources`, and `gateway.prompts` (already eager-loaded) and sets their visibility directly
- Removed the duplicate `gateway.visibility` assignment

## 🧪 Verification

| New regression test                   | `pytest ...test_update_gateway_visibility_propagates_when_init_fails` | ✅ |

## ✅ Checklist
- [x] TDD: test written first, verified failing, then fix applied

### DIFF
--- a/tests/unit/mcpgateway/services/test_gateway_service.py
+++ b/tests/unit/mcpgateway/services/test_gateway_service.py
@@ -1106,6 +1106,8 @@ class TestGatewayService:
         assert mock_tool.visibility == "team", "Tool visibility not propagated when gateway init failed"
         assert mock_resource.visibility == "team", "Resource visibility not propagated when gateway init failed"
         assert mock_prompt.visibility == "team", "Prompt visibility not propagated when gateway init failed"
+        # Visibility changes must be persisted
+        test_db.commit.assert_called_once()
 
     @pytest.mark.asyncio
     async def test_update_gateway_partial_update(self, gateway_service, mock_gateway, test_db):


crivetimihai:
Thanks @madhav165 — critical fix for #3475. Moving visibility propagation out of the `_initialize_gateway` try block ensures consistency even when the upstream is unreachable. Also good that you removed the duplicate `gateway.visibility` assignment. Regression test is thorough. LGTM.

TS0713:
I have reviewed and tested the changes introduced in this pull request related to propagating gateway visibility to the linked tools, prompts, and resources.

The functionality works as expected. I validated the behavior by creating gateways with different visibility configurations (both all teams and specific team) and verified that visibility updates correctly propagate to the associated tools, resources, and prompts.

Test Scenarios Executed from gateway UI

### Case 1 – Gateway in Active State

Created a gateway and updated the visibility settings.

Verified that the updated visibility was correctly reflected across the linked tools, resources, and prompts.

### Case 2 – Gateway in Deactivated State

Updated the visibility of a deactivated gateway.

Confirmed that the changes were still properly propagated to the linked tools, resources, and prompts.

### Case 3 – Gateway with MCP Server Terminated

Modified the visibility configuration after the MCP server was terminated.

Verified that the visibility changes were correctly reflected across tools, resources, and prompts.

Across all scenarios, the visibility propagation behaved consistently and no unexpected issues were observed.

crivetimihai:
## Review Summary

### Fix Verified

The fix correctly moves visibility propagation to run **before** the `_initialize_gateway` try block, ensuring linked tools/resources/prompts get updated even when the upstream MCP server is unreachable. The duplicate `gateway.visibility` assignment was also removed.

### Changes Made During Review

- **Rebased** onto current `main` (clean, no conflicts)
- **Added `test_db.commit.assert_called_once()`** to the regression test to verify visibility changes are persisted to the database

### Testing Performed

**Unit tests**: 260/260 passed, 1 skipped

**E2E tests on deployed instance** (3 gateway containers behind nginx):

| Test | Result |
|------|--------|
| Visibility propagation — upstream UP (DB verified) | PASS |
| Visibility propagation — upstream DOWN (DB verified) | PASS |
| All item types propagated (2 tools, 4 resources, 3 prompts) | PASS |
| Rapid visibility toggles — final state consistent | PASS |
| Non-visibility updates preserve visibility | PASS |
| Other gateways unaffected by update | PASS |
| Concurrent updates to different gateways | PASS |
| Auth enforcement (401 without token) | PASS |
| Tool execution works after visibility changes | PASS |
| Non-existent gateway returns 404 | PASS |

**Admin UI (Playwright)**: Verified visibility changes reflect correctly on MCP Servers, Tools, Prompts, and Resources pages.

### Pre-existing Issues Found (not introduced by this PR)

Filed as separate bugs:
- #3523 — `_create_db_tool()` hardcodes `visibility="public"` instead of inheriting gateway visibility
- #3524 — Resource visibility inconsistency between `register_gateway()` and `_update_or_create_resources()`
- #3525 — `GatewayUpdate.visibility` accepts arbitrary strings (no enum validation)
- #3526 — Cache invalidation gap for tools/resources/prompts list caches after visibility update

PR #3516, opened 2026-03-06
[FIX][RBAC]: Session tokens denied tools.execute on /rpc and /mcp despite having team-scoped role

### LINKED ISSUE #3515: [BUG]: session tokens with team-scoped roles denied tools.execute on /rpc and /mcp transports
### 🐞 Bug Summary

Users authenticated with session tokens that have `tools.execute` only in team-scoped roles are incorrectly denied permission when calling tools via `/rpc` (`tools/call`) and `/mcp` (Streamable HTTP). The same user succeeds on REST endpoints and the SSE `/message` endpoint because those use the `@require_permission` decorator which includes a `check_any_team` fallback for session tokens.

---

### 🧩 Affected Component

- [x] `mcpgateway` - API

---

### 🔁 Steps to Reproduce

1. Create a user with no global `tools.execute` permission — only team-scoped roles that include `tools.execute`
2. Generate a session token for that user (token has no `teams` claim)
3. Call `tools/call` via `POST /rpc` or via the `/mcp` Streamable HTTP transport
4. Observe `-32003 Insufficient permissions. Required: tools.execute`

---

### 🤔 Expected Behavior

A user with `tools.execute` granted via any of their team-scoped roles should be able to execute tools via all transports, consistent with the behaviour of REST endpoints (`POST /tools/{id}/execute`) and `POST /message`.

---

### 📓 Logs / Error Output

```json
{
  "jsonrpc": "2.0",
  "error": {
    "code": -32003,
    "message": "Insufficient permissions. Required: tools.execute",
    "data": { "method": "tools/call" }
  }
}
```

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Platform / OS | Linux |
| Container | Docker |

---

### 🧩 Additional Context

**Root cause:** Two permission-check helpers are missing the session-token `check_any_team` fallback that the `@require_permission` decorator applies (rbac.py:562–576).

`/rpc` cannot use `@require_permission` at the endpoint level because it is a multi-method dispatcher — the required permission is not known until the request body is parsed. Instead it uses `_ensure_rpc_permission` (main.py:555), which calls `PermissionChecker.has_permission` with no `team_id` and `check_any_team=False`.

`/mcp` (Streamable HTTP transport) uses `_check_streamable_permission` (streamablehttp_transport.py:596) with the same gap.

In both cases, `PermissionService._get_user_roles` falls back to only loading global roles and team roles with `scope_id=NULL`, missing the user's actual team-scoped assignments.

**Affected call sites:**

| Path | Function | Missing fallback |
|------|----------|-----------------|
| `POST /rpc` — `tools/call` | `_ensure_rpc_permission` (main.py:6370) | `check_any_team=True` for session tokens |
| `POST /rpc` — legacy method | `_ensure_rpc_permission` (main.py:6651) | same |
| `POST /mcp` | `_check_streamable_permission` (streamablehttp_transport.py:990) | same |

**Not affected** (use `@require_permission` decorator which already has the fallback):
- `POST /message`
- `POST /tools/{id}/execute`
- All other REST mutation endpoints

**Fix:** Add session-token `check_any_team` derivation to `_ensure_rpc_permission` and `_check_streamable_permission`, mirroring the logic already present in `@require_permission` (rbac.py:562–576).

madhav165:
## 📌 Summary

Session tokens whose `tools.execute` permission exists only in a team-scoped RBAC role were incorrectly denied on `/rpc` (`tools/call`) and `/mcp` (Streamable HTTP). The same user succeeded on REST endpoints because those use the `@require_permission` decorator which already applies a `check_any_team` fallback for session tokens.

Closes #3515

## 🔁 Reproduction Steps

1. Set `DEFAULT_TEAM_MEMBER_ROLE=developer` in your environment
2. Create a new user and a new team via the admin UI
3. Log in as that user
4. Add an MCP server scoped to that team
5. Go to the Tools screen and run a tool — it fails with `-32003 Insufficient permissions. Required: tools.execute`

## 🐞 Root Cause

Two permission-check helpers were missing the `check_any_team` fallback that the `@require_permission` decorator applies for session tokens (rbac.py:562–576):

- **`_ensure_rpc_permission`** (main.py) — used by `POST /rpc`. Cannot use `@require_permission` at the endpoint level because `/rpc` is a multi-method dispatcher; permission is resolved after parsing the JSON-RPC body. Called `PermissionChecker.has_permission` without `check_any_team`, so `_get_user_roles` only returned global roles and team roles with `scope_id=NULL`, missing the user's actual team assignments.

- **`_check_streamable_permission`** call site (streamablehttp_transport.py) — used by `POST /mcp`. The function already accepted `check_any_team` but it was always called with the default `False`. Additionally, `token_use` was not stored in `auth_user_ctx`, so the transport had no way to detect session tokens.

## 💡 Fix Description

Three targeted changes:

1. **`PermissionChecker.has_permission`** (`mcpgateway/middleware/rbac.py`) — added `check_any_team: bool = False` parameter and threaded it through to both `PermissionService.check_permission` call sites. Default is `False`, preserving all existing callers.

2. **`_ensure_rpc_permission`** (`mcpgateway/main.py`) — detects session tokens via `user.get("token_use") == "session"` and passes `check_any_team=True` to `PermissionChecker.has_permission`.

3. **Streamable HTTP transport** (`mcpgateway/transports/streamablehttp_transport.py`) — stores `token_use` in `auth_user_ctx` so it propagates into the MCP handler context; `call_tool` now detects session tokens and passes `check_any_team=True` to `_check_streamable_permission`.

## 🔐 Security Rationale for `check_any_team=True`

A reviewer might ask: does `check_any_team=True` allow a user with `tools.execute` in Team A to execute a team-scoped tool owned by Team B?

**The answer is no, and here is why the two paths differ:**

**REST path (`@require_permission`):** Tier 1 team derivation (rbac.py:566) looks up the target resource's `team_id` from the DB before the RBAC check. For a Team B tool, `team_id = Team B` is derived, `check_any_team` stays `False`, and the check "does user have `tools.execute` in Team B?" fails → **denied at RBAC (Layer 2)**.

**RPC/MCP path (this fix):** `check_any_team=True` means the RBAC check passes (user has `tools.execute` in Team A). However, `tool_service.invoke_tool` then calls `_has_tool_access_by_visibility` (tool_service.py:796–807), which explicitly evaluates `tool.team_id in token_teams`. For a session token user in Team A only, `token_teams = ["team-a-id"]` (derived from server-side membership), so `"team-b-id" NOT IN ["team-a-id"]` → `ToolNotFoundError` → **denied at tool service (Layer 1)**.

Both paths deny cross-team execution. The REST path catches it at RBAC; the RPC/MCP path catches it at the tool service's visibility check. Public tools (visibility = `"public"`) are correctly accessible to any authenticated user with the permission on both paths — that is the intended behaviour of public visibility.

**API tokens** carry explicit `token_teams` JWT claims and never use `check_any_team` (it stays `False`). Their scope is fully determined by the token claims, not server-side membership.

## 🧪 Verification

| Check | Command | Status |
|---|---|---|
| Unit tests | `uv run pytest tests/unit/` | ✅ |
| New regression tests | `uv run pytest tests/unit/mcpgateway/test_rpc_permission_team_fallback.py tests/unit/mcpgateway/transports/test_streamable_rpc_permission_fallback.py -v` | ✅ 11 tests |
| Manual regression | Tools screen works for session-token user with team-scoped developer role | ✅ |

## ✅ Checklist

### DIFF
--- a/tests/playwright/test_rbac_permissions.py
+++ b/tests/playwright/test_rbac_permissions.py
@@ -1348,3 +1348,189 @@ class TestRPCToolExecutionRBAC:
         finally:
             if tool_id:
                 admin_api.delete(f"/tools/{tool_id}")
+
+
+# ==================== #3515 Regression: Browser cookie session-token paths ====================
+
+
+@pytest.mark.ui
+@pytest.mark.rbac
+@pytest.mark.flaky(reruns=1, reason="Browser cookie auth intermittently returns 401")
+class TestSessionTokenCookieRBAC:
+    """Regression tests for #3515 exercising session tokens via browser cookies.
+
+    The TestRPCToolExecutionRBAC class above uses Bearer-header API contexts.
+    These tests inject JWT cookies (the actual browser auth path) and issue
+    fetch() calls from the page, matching how the Admin UI invokes /rpc.
+    """
+
+    def test_developer_cookie_rpc_tools_call(self, page: Page, base_url: str, admin_api: APIRequestContext, rbac_developer_user: Dict, rbac_test_team: Dict):
+        """Developer cookie session must pass RBAC on /rpc tools/call (#3515).
+
+        Injects a session-token JWT cookie and uses page.evaluate(fetch(...))
+        to call /rpc, exactly matching the Admin UI Tools screen flow.
+        """
+        team_id = rbac_test_team["id"]
+        tool_name = f"{RBAC_TEST_PREFIX}-cookie-exec-{uuid.uuid4().hex[:8]}"
+
+        import json as _json  # noqa: PLC0415
+
+        create_resp = admin_api.post(
+            "/tools",
+            data=_json.dumps({
+                "tool": {"name": tool_name, "description": "Cookie RBAC test (#3515)", "url": f"{base_url}/health", "integration_type": "REST", "input_schema": {}},
+                "team_id": team_id,
+                "visibility": "team",
+            }),
+            headers={"Content-Type": "application/json"},
+        )
+        assert create_resp.status in (200, 201), f"Failed to create tool: {create_resp.status} {create_resp.text()}"
+        tool_id = create_resp.json().get("id")
+
+        try:
+            _inject_jwt_cookie(page, rbac_developer_user["email"], token_use="session")
+            page.goto(f"{base_url}/admin/tools")
+
+            result = page.evaluate(
+                """async (toolName) => {
+                    const resp = await fetch('/rpc', {
+                        method: 'POST',
+                        headers: {'Content-Type': 'application/json'},
+                        body: JSON.stringify({jsonrpc: '2.0', id: 1, method: 'tools/call', params: {name: toolName, arguments: {}}})
+                    });
+                    return {status: resp.status, body: await resp.json()};
+                }""",
+                tool_name,
+            )
+            error_code = result["body"].get("error", {}).get("code")
+            assert error_code != -32003, (
+                f"Developer cookie session denied tools.execute with -32003 — #3515 regression. "
+                f"Response: {result['body']}"
+            )
+            logger.info("Developer cookie /rpc tools/call: error_code=%s — RBAC passed", error_code)
+        finally:
+            if tool_id:
+                admin_api.delete(f"/tools/{tool_id}")
+
+    def test_viewer_cookie_rpc_tools_call_denied(self, page: Page, base_url: str, admin_api: APIRequestContext, rbac_viewer_user: Dict, rbac_test_team: Dict):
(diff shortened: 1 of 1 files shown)


MohanLaksh (approved):
### Followed the below steps to test the fix:

- Set `DEFAULT_TEAM_MEMBER_ROLE=developer` in my `.env` file
- Create a new user (user1@example.com) and a new team (Test Team) via the admin UI
- **Logged in** as `user1@example.com`
- Added **Github MCP** server scoped to **Test team**
- Went to the Tools screen and run a tool (github-tools-get-me)

## Successfully able to execute the tool 


Works as intended.

# APPROVED

PR #3570, opened 2026-03-09
fix: _prepare_gateway_for_read mutates ORM object 

### LINKED ISSUE #3540: [BUG][API]: _prepare_gateway_for_read mutates ORM object, causing auth_value dict→string writeback
## Bug Description

The deprecated `_prepare_gateway_for_read()` method mutates the SQLAlchemy ORM object's `auth_value` field in-place (dict → encoded string). When called on an object still attached to an active session, SQLAlchemy tracks the mutation and writes the encoded string back to the `DbGateway.auth_value` JSON column on commit — replacing the intended dict with a JSON string.

## Root Cause

`gateway_service.py:4020-4022`:
```python
def _prepare_gateway_for_read(self, gateway: DbGateway) -> DbGateway:
    # ...
    if isinstance(gateway.auth_value, dict):
        gateway.auth_value = encode_auth(gateway.auth_value)  # MUTATES ORM object
```

This is called on attached ORM objects in active sessions across **6 call sites**. The safe replacement `convert_gateway_to_read()` (line 3970) uses `__dict__.copy()` and does NOT mutate the original, but only 1 of 7 call sites has been migrated.

## Affected Call Sites

| Line | Method | Risk | Reason |
|------|--------|------|--------|
| 1184 | `register_gateway()` | Medium | Inside session; implicit commit may persist mutation |
| 1730 | `list_gateways_for_user()` | Low | Called after `db.commit()`, objects may be detached |
| 2271 | `update_gateway()` | **High** | `db.refresh()` re-attaches object before mutation |
| 2436 | `get_gateway()` | Medium | Active session, mutation tracked |
| 2727 | `set_gateway_state()` | Medium | Active session, mutation tracked |
| 3726 | `get_first_gateway_by_url()` | Medium | Active session, mutation tracked |

The safe `convert_gateway_to_read()` is used only at line 1623 (`list_gateways()`).

## Impact

When the mutation persists, the JSON column stores a JSON string `"encoded_value"` instead of a JSON object `{"key": "value"}`. On subsequent reads:

1. `isinstance(gateway.auth_value, str)` → True (no longer a dict)
2. Code paths with `isinstance(auth_value, dict)` guards (from PR #3510) skip encoding — the value passes through as-is
3. `decode_auth(string)` works correctly on the encoded value, so **no data loss occurs**
4. However, the column type annotation `Mapped[Optional[Dict[str, str]]]` is violated — the column holds a string, not a dict

The practical impact is that the fix from #3510 (storing dict in JSON column) gets partially undone by the mutation after each read that goes through `_prepare_gateway_for_read`.

## Suggested Fix

Replace all 6 remaining call sites of `_prepare_gateway_for_read()` with `convert_gateway_to_read()`:

```python
# BEFORE (mutates ORM object):
return GatewayRead.model_validate(self._prepare_gateway_for_read(gateway)).masked()

# AFTER (safe, uses dict copy):
return self.convert_gateway_to_read(gateway)
```

`convert_gateway_to_read()` already handles dict encoding, tag conversion, metadata fields, and returns `.masked()`. It is a drop-in replacement.

After migration, `_prepare_gateway_for_read()` can be removed entirely.

## Reproduction

1. Register a gateway with `auth_type=authheaders`
2. Query the DB: `SELECT auth_value FROM gateways WHERE name='...'` → shows JSON object
3. Call `GET /gateways/{id}` (triggers `_prepare_gateway_for_read`)
4. Query DB again → value may now be a JSON string (encoded) instead of a JSON object

## Related

- PR #3510 — fixes the creation path to store dict, but the mutation partially undoes it
- `_prepare_gateway_for_read` is already marked `DEPRECATED` in its docstring (line 4007)

madhu-mohan-jaishankar:
## 🔗 Related Issue
Closes #3540

---

## 📝 Summary
`_prepare_gateway_for_read()` was mutating the attached SQLAlchemy `DbGateway` ORM instance in-place by converting `auth_value` from a dict to an encoded string. If the instance was still attached to an active session, SQLAlchemy could treat this as a dirty-field change and **persist the encoded string back into the JSON column on commit/flush**, unintentionally changing the stored shape of `auth_value` (dict → string).

This PR replaces the unsafe read-preparation usage with `convert_gateway_to_read()`, which produces a read model without mutating the underlying ORM object, preventing the auth_value “string writeback” behavior.

---

## 🏷️ Type of Change
- [x] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

---

## ✅ Checklist

---

### Problem / Root Cause
- `_prepare_gateway_for_read()` performed an in-place transformation:
  - `gateway.auth_value` dict → `encode_auth(dict)` string
- When called on ORM instances still attached to a session, this mutation could be persisted on the next commit/flush.
- This breaks the expectation that the `auth_value` JSON column stores a JSON object (dict) and causes subtle inconsistencies across read/write paths.

### What changed
- Replaced `_prepare_gateway_for_read()` usage with `convert_gateway_to_read()` so that:
  - the response still contains the correctly encoded/masked representation for clients
  - but the underlying ORM object is **not mutated**, avoiding unintended persistence changes
- Updated tests to reflect the new non-mutating read conversion behavior.

### Why this is safe
- `convert_gateway_to_read()` is the intended safe replacement (constructs a read model via copying rather than mutating the ORM instance).
- No functional change is intended for API consumers besides eliminating the side-effect of DB writeback on read paths.

### DIFF
--- a/mcpgateway/services/gateway_service.py
+++ b/mcpgateway/services/gateway_service.py
@@ -4003,32 +4003,6 @@ class GatewayService(BaseService):  # pylint: disable=too-many-instance-attribut
 
         return GatewayRead.model_validate(gateway_dict).masked()
 
-    def _prepare_gateway_for_read(self, gateway: DbGateway) -> DbGateway:
-        """DEPRECATED: Use convert_gateway_to_read instead.
-
-        Prepare a gateway object for GatewayRead validation.
-
-        Ensures auth_value is in the correct format (encoded string) for the schema.
-        Converts legacy List[str] tags to List[Dict[str, str]] format for GatewayRead schema.
-
-        Args:
-            gateway: Gateway database object
-
-        Returns:
-            Gateway object with properly formatted auth_value and tags
-        """
-        # If auth_value is a dict, encode it to string for GatewayRead schema
-        if isinstance(gateway.auth_value, dict):
-            gateway.auth_value = encode_auth(gateway.auth_value)
-
-        # Handle legacy List[str] tags - convert to List[Dict[str, str]] for GatewayRead schema
-        if gateway.tags:
-            if isinstance(gateway.tags[0], str):
-                # Legacy format: convert to dict format
-                gateway.tags = validate_tags_field(gateway.tags)
-
-        return gateway
-
     def _create_db_tool(
         self,
         tool: ToolCreate,
--- a/tests/unit/mcpgateway/services/test_gateway_service_helpers.py
+++ b/tests/unit/mcpgateway/services/test_gateway_service_helpers.py
@@ -111,15 +111,6 @@ def test_gateway_service_convert_gateway_to_read(monkeypatch):
     assert result._masked_called, "convert_gateway_to_read must call .masked() to prevent credential leakage"
 
 
-def test_gateway_service_prepare_gateway_for_read():
-    service = GatewayService()
-    gateway = SimpleNamespace(auth_value={"token": "secret"}, tags=["Analytics", "ml"])
-
-    updated = service._prepare_gateway_for_read(gateway)
-    assert decode_auth(updated.auth_value) == {"token": "secret"}
-    assert updated.tags == validate_tags_field(["Analytics", "ml"])
-
-
 def test_gateway_service_validate_tools_valueerror(monkeypatch):
     service = GatewayService()
 
@@ -203,8 +194,7 @@ async def test_authheaders_auth_value_stored_as_dict(monkeypatch):
     db.refresh = Mock()
 
     # Snapshot at db.add() time — tools flow through the gateway relationship (gateway.tools=tools),
-    # not separate db.add() calls. _prepare_gateway_for_read() later mutates db_gateway.auth_value
-    # to an encoded string for the GatewayRead response; we capture before that mutation.
+    # not separate db.add() calls.
     # First-Party
     from mcpgateway.db import Gateway as DbGateway
 
--- a/tests/unit/mcpgateway/services/test_gateway_service_oauth_comprehensive.py
+++ b/tests/unit/mcpgateway/services/test_gateway_service_oauth_comprehensive.py
@@ -232,22 +232,6 @@ def test_check_gateway_uniqueness_no_auth_duplicate():
     assert result is existing
 
 
-def test_prepare_gateway_for_read_encodes_auth(monkeypatch):
-    """Prepare gateway should encode auth dict and normalize tags."""
-    service = GatewayService()
-    gateway = MagicMock()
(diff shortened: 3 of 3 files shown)


ja8zyjits:
Hi  @madhu-mohan-jaishankar ,

Good code changes
PR #3570 is a **well-executed bug fix** that:
- ✅ Addresses all 6 affected call sites identified in issue #3540
- ✅ Uses the correct safe replacement method
- ✅ Includes appropriate test updates
- ✅ Maintains backward compatibility
- ✅ Has no scope creep
- ✅ Eliminates the root cause of the auth_value writeback bug

But may I know if the final outcome was verified in the query response?

```python
>>> import sqlite3
>>> con = sqlite3.connect("mcp.db")
>>> val = con.execute("SELECT auth_value FROM gateways WHERE name='Dummy Server'")
>>> val
<sqlite3.Cursor object at 0x1053602c0>
>>> val.fetchone()
('{"auth_header_key": "abcd"}',)
# after calling Call GET /gateways/{id} (triggers _prepare_gateway_for_read)
>>> val1 = con.execute("SELECT auth_value FROM gateways WHERE name='Dummy Server'")
>>> val1.fetchone()
('{"auth_header_key": "abcd"}',)
```

In both cases I am getting the `str` type as response. May be Iam missing something here.

ja8zyjits:
HI @madhu-mohan-jaishankar ,

Thanks for the clarification
The following is the output without the changes implemented
```python
>>> import sqlite3
>>> con = sqlite3.connect("mcp.db")
>>> val = con.execute("SELECT auth_value FROM gateways WHERE name='Dummy'")
>>> val.fetchone()
('"kelhib23RbExxH1LFKshM61e4IqQrpl_Pvc9iA5erHCj7Wzo5xjUzE0CO6wQlpNuFGA"',)
# after calling Call GET /gateways/{id} (triggers _prepare_gateway_for_read)
>>> val1 = con.execute("SELECT auth_value FROM gateways WHERE name='Dummy'")
>>> val1.fetchone()
('"kelhib23RbExxH1LFKshM61e4IqQrpl_Pvc9iA5erHCj7Wzo5xjUzE0CO6wQlpNuFGA"',)
```

I have verified it and am satisfied with the changes

cafalchio (approved):
LGTM, perhaps _prepare_gateway_for_read can be removed?

crivetimihai:
Cleaned up the dead `_prepare_gateway_for_read` monkeypatch in `tests/unit/mcpgateway/test_multi_auth_headers.py:333` — it patched a method no longer called by production code. All 22 tests in that file still pass.

PR #3574, opened 2026-03-10
fix(visibility): prevent auto-refresh from overriding tool/prompt/resource visibility

### LINKED ISSUE #3530: [BUG]: AUTO_REFRESH_SERVERS Changes Tool Visibility From Private to Public
Title

AUTO_REFRESH_SERVERS=true automatically changes tool visibility from Private to Public

Product

IBM Context Forge

Environment

Context Forge: Latest version

MCP Server visibility: Public

Configuration:

AUTO_REFRESH_SERVERS=true
Issue Description

When AUTO_REFRESH_SERVERS is enabled, the visibility of tools is automatically changed from Private to Public during server refresh.

This behavior occurs when the MCP server visibility is Public.

However, in our use case, a single MCP server can contain both Public and Private tools, depending on access and security requirements.

Steps to Reproduce

Configure an MCP server with visibility set to Public.

Add multiple tools to the server.

Set some tools to Private.

Enable the following configuration:

AUTO_REFRESH_SERVERS=true

Allow the system to refresh or restart the server/tool registration.

Check the visibility of the tools.

Actual Result

Tools that were previously configured as Private are automatically changed to Public after the server refresh.

Expected Result

Tool visibility should remain unchanged after server refresh.

The AUTO_REFRESH_SERVERS feature should only refresh tool metadata and registration, and should not override explicitly configured tool visibility settings.

Impact

This behavior causes operational issues because:

A single MCP server can host tools with different visibility requirements.

Some tools must remain Private due to security or internal usage.

Automatic conversion to Public introduces risk of unintended exposure.

kevalmahajan:
# 🐛 Bug-fix PR

Closes #3530

## 📌 Summary
Fixed a bug where the `AUTO_REFRESH_SERVERS=true` configuration overrides the granular visibility settings of individual tools, resources, and prompts. Previously, during a server auto-refresh, any explicitly configured "Private" or "Team" items would automatically be changed to match the parent Gateway's visibility (e.g., "Public"). This fix ensures that background sync operations preserve existing visibility configurations while still allowing manual UI/API updates to propagate visibility changes correctly.

## 🔁 Reproduction Steps
1. Create and configure an MCP Gateway with "Public" visibility.
2. Allow the Gateway to perform its initial sync so that its tools default to "Public".
3. Manually update one of the synced tools to have "Private" visibility.
4. Trigger a background auto-refresh for the Gateway (e.g., via `AUTO_REFRESH_SERVERS=true` or a health check).
5. **Observed Bug:** The tool's visibility is changed back to "Public".
6. **Expected Behavior:** The tool remains "Private".

## 🐞 Root Cause
The `_update_or_create_tools`, `_update_or_create_resources`, and `_update_or_create_prompts` helper methods in `mcpgateway/services/gateway_service.py` contained logic that unconditionally applied `existing_item.visibility = gateway.visibility` during the synchronization process. The methods did not differentiate between an automatic background refresh and a manual update triggered by a user.

## 💡 Fix Description
The `_update_or_create_tools`, `_update_or_create_resources`, and `_update_or_create_prompts` helper methods in `mcpgateway/services/gateway_service.py` contained logic that unconditionally applied `existing_item.visibility = gateway.visibility` during the synchronization process. The methods did not differentiate between an automatic background refresh and a manual update triggered by a user.
## 💡 Fix Description
Updated the synchronization logic across tools, resources, and prompts to conditionally evaluate visibility changes based on the operation's context. 
- Introduced a `should_update_visibility = created_via == "update"` condition.
- The item's `visibility` attribute is now explicitly excluded from the `fields_to_update` conditions and field assignment blocks unless the sync is triggered by a manual update (`"update"`). 
- Added comprehensive unit tests `test_update_visibility_logic` and updated removal scenarios) in `test_gateway_service_extended.py` to assert that visibility correctly inherits during manual updates but is preserved during `auto_refresh`, `health_check`, and `rediscovery` contexts.

## 🧪 Verification

| Lint suite                            | `make lint`          |   ✅     |
| Unit tests                            | `make test`          |     ✅   |
| Manual regression no longer fails     | steps / screenshots  |    ✅    |

## ✅ Checklist

### DIFF
--- a/mcpgateway/services/gateway_service.py
+++ b/mcpgateway/services/gateway_service.py
@@ -2078,14 +2078,16 @@ class GatewayService(BaseService):  # pylint: disable=too-many-instance-attribut
                         gateway.auth_value = None
                         gateway.oauth_config = None
 
-                    # Update tools using helper method
-                    tools_to_add = self._update_or_create_tools(db, tools, gateway, "update")
+                    # Update tools using helper method — only propagate visibility
+                    # when the user explicitly changed it in this request
+                    _vis_changed = gateway_update.visibility is not None
+                    tools_to_add = self._update_or_create_tools(db, tools, gateway, "update", update_visibility=_vis_changed)
 
                     # Update resources using helper method
-                    resources_to_add = self._update_or_create_resources(db, resources, gateway, "update")
+                    resources_to_add = self._update_or_create_resources(db, resources, gateway, "update", update_visibility=_vis_changed)
 
                     # Update prompts using helper method
-                    prompts_to_add = self._update_or_create_prompts(db, prompts, gateway, "update")
+                    prompts_to_add = self._update_or_create_prompts(db, prompts, gateway, "update", update_visibility=_vis_changed)
 
                     # Log newly added items
                     items_added = len(tools_to_add) + len(resources_to_add) + len(prompts_to_add)
@@ -4054,7 +4056,7 @@ class GatewayService(BaseService):  # pylint: disable=too-many-instance-attribut
             visibility=gateway.visibility,
         )
 
-    def _update_or_create_tools(self, db: Session, tools: List[Any], gateway: DbGateway, created_via: str) -> List[DbTool]:
+    def _update_or_create_tools(self, db: Session, tools: List[Any], gateway: DbGateway, created_via: str, update_visibility: bool = False) -> List[DbTool]:
         """Helper to handle update-or-create logic for tools from MCP server.
 
         Args:
@@ -4062,6 +4064,7 @@ class GatewayService(BaseService):  # pylint: disable=too-many-instance-attribut
             tools: List of tools from MCP server
             gateway: Gateway object
             created_via: String indicating creation source ("oauth", "update", etc.)
+            update_visibility: Whether to propagate gateway visibility to existing tools
 
         Returns:
             List of new tools to be added to the database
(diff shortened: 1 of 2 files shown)


crivetimihai:
## Review & Changes

Rebased onto main (clean, no conflicts) and reviewed the implementation. The core fix — preserving per-item visibility during auto-refresh — is correct and well-targeted. I've added three commits on top:

### 1. Formatting cleanup (`702b7ce`)
Trailing whitespace and extra blank line at EOF flagged by `black`.

### 2. New tools inherit gateway visibility (`d357598`)
`_create_db_tool()` hardcoded `visibility="public"` for all newly discovered tools, while resources and prompts correctly used `gateway.visibility`. This broke team isolation for new tools on non-public gateways. Changed to `visibility=gateway.visibility` with test.

### 3. Explicit `update_visibility` flag (`20dee56`)
The original `should_update_visibility = created_via == "update"` was too broad: **any** gateway edit (e.g., changing only the description) would overwrite custom per-item visibility because the sync helpers always run with `created_via="update"`. The direct propagation at lines 1902–1912 already correctly gates on `gateway_update.visibility is not None`, but the helpers didn't match.

Replaced with an explicit `update_visibility: bool = False` parameter on all three sync helpers, set to `True` only when the user actually changed the gateway's visibility in the request. Added a dedicated test (Test 3) verifying that `update_visibility=False` preserves custom visibility even during an "update" context.

**Verified against live deployment** (3-instance + PostgreSQL + Redis):
- Manual refresh preserves per-item visibility
- Explicit gateway visibility change propagates to all linked items
- Deactivate/reactivate cycle preserves visibility
- Token scoping correctly hides private/team items from public-only tokens

All 4958 unit tests pass, formatting clean.

crivetimihai (approved):
Approved. The core fix is correct and I've added three commits addressing a pre-existing bug (_create_db_tool hardcoding visibility), a correctness issue (unrelated gateway edits overwriting per-item visibility), and formatting. All verified against live deployment and 4958 unit tests pass.

Issue #2419, opened 2026-01-25
[TESTING][FUNCTIONALITY]: MCP Resources manual test plan (discovery, fetch, templates, subscriptions)


crivetimihai:
# 📦 [TESTING][FUNCTIONALITY]: MCP Resources Manual Test Plan

## Goal

Produce a **comprehensive manual test plan** for MCP resource functionality including discovery, fetching, templates, subscriptions, and content handling.

## Why Now?

MCP resources provide context and data to AI clients:

1. **Discovery**: Clients need to know available data sources
2. **Fetching**: Resources must be retrievable with correct content
3. **Templates**: Dynamic resources need parameter support
4. **Subscriptions**: Real-time updates require reliable notifications
5. **Security**: Resource access must respect permissions

---

## 📖 User Stories

<details>
<summary><strong>US-1: Client - Resource Discovery</strong></summary>

**As a** MCP client
**I want** to discover available resources
**So that** I can access relevant data

**Acceptance Criteria:**

```gherkin
Feature: Resource Discovery

  Scenario: List all resources
    Given an authenticated MCP client
    When I send a resources/list request
    Then I should receive a list of available resources
    And each resource should have uri, name, and mimeType

  Scenario: Resource templates
    Given resources with URI templates
    When I list resources
    Then template resources should include uriTemplate field
```

</details>

<details>
<summary><strong>US-2: Client - Resource Fetching</strong></summary>

**As a** MCP client
**I want** to fetch resource content
**So that** I can provide context to users

**Acceptance Criteria:**

```gherkin
Feature: Resource Fetching

  Scenario: Fetch static resource
    Given a static resource exists
    When I fetch the resource by URI
    Then I should receive the resource content
    And the content type should match mimeType

  Scenario: Fetch template resource
    Given a template resource exists
    When I fetch with template parameters
    Then the resource should be resolved with parameters
```

</details>

---

## 🏗 Architecture

```
┌─────────────────────────────────────────────────────────────────────────────┐
│                        MCP RESOURCE FLOW                                     │
└─────────────────────────────────────────────────────────────────────────────┘

    CLIENT                  GATEWAY                    MCP SERVER
    ──────                  ───────                    ──────────

  ┌──────────┐           ┌──────────────┐           ┌──────────────┐
  │ MCP      │           │              │           │              │
  │ Client   │──────────▶│  Gateway     │──────────▶│  MCP Server  │
  │          │           │              │           │              │
[...]

PR #3595, opened 2026-03-10
fix: SSE resource subscribe endpoint yielding raw dicts instead of SSE-formatted strings

marekdano:
# 🐛 Bug-fix PR

## 📌 Summary

Found this bug while executing manual tests on Resources https://github.com/IBM/mcp-context-forge/issues/2419

The POST /resources/subscribe SSE endpoint at main.py:4864 passes resource_service.subscribe_events() directly to StreamingResponse. However, subscribe_events() yields raw Python dicts, while StreamingResponse expects an async generator of strings or bytes.

This causes the SSE connection to remain open (HTTP 200) but no events are ever written to the stream. Clients stay connected but never receive resource change notifications (resource_updated, resource_added, resource_deleted, etc.) over SSE.

## 🔁 Reproduction Steps

1. Terminal 1 — open SSE stream (should stay open now):
```bash
curl -N -X POST "http://localhost:8080/resources/subscribe" \
      -H "Accept: text/event-stream"
```

2.  Terminal 2 — update the resource to trigger a notification:
```bash
curl -s -X PUT "http://localhost:8080/resources/$RESOURCE_ID" \
      -d '{
        "resource": {
          "uri": "file:///config/settings.json",
          "name": "config-file",
          "description": "Updated settings",
          "mimeType": "application/json",
          "content": "{\"theme\": \"light\", \"lang\": \"fr\"}"
        },
        "visibility": "public"
      }' | jq .
```

3. Terminal 1 should now show (but does not show):
```bash
data: {"type":"resource_updated","data":{"id":"...","uri":"file:///config/settings.json",...},"timestamp":"..."}
```

## 🐞 Root Cause

The `EventService` already has a correct `event_generator()` method that formats events as `data: {...}\n\n`, but it was not being used by this endpoint.

## 💡 Fix Description

Wrap the async generator in an SSE formatter that serializes each event dict to the standard SSE `data: {...}\n\n` format before yielding to StreamingResponse:

```bash
  async def sse_generator():
      async for event in resource_service.subscribe_events(user_email=user_email, token_teams=token_teams):
          yield f"data: {orjson.dumps(event).decode()}\n\n"

  return StreamingResponse(sse_generator(), media_type="text/event-stream")
```

This aligns with the pattern used by `EventService.event_generator()` and the SSE specification (each event prefixed with `data:` and terminated with `\n\n`).

## 🧪 Verification

| Lint suite                            | `make lint`          |   ✅     |
| Unit tests                            | `make test`          |   ✅     |

## ✅ Checklist

### DIFF
--- a/mcpgateway/main.py
+++ b/mcpgateway/main.py
@@ -4878,10 +4878,10 @@ async def subscribe_resource(request: Request, user=Depends(get_current_user_wit
     user_email, token_teams = _get_scoped_resource_access_context(request, user)
 
     async def sse_generator():
-        """Yield resource events formatted as SSE 'data: {...}\\n\\n' strings.
+        """Generate SSE-formatted events from resource subscription changes.
 
         Yields:
-            str: SSE-formatted event string.
+            str: SSE-formatted event data.
         """
         async for event in resource_service.subscribe_events(user_email=user_email, token_teams=token_teams):
             yield f"data: {orjson.dumps(event).decode()}\n\n"
--- a/tests/unit/mcpgateway/test_main.py
+++ b/tests/unit/mcpgateway/test_main.py
@@ -21,6 +21,7 @@ from unittest.mock import ANY, AsyncMock, MagicMock, patch
 from fastapi import HTTPException
 from fastapi.testclient import TestClient
 import jwt
+import orjson
 from pydantic import BaseModel, SecretStr, ValidationError
 import pytest
 import sqlalchemy as sa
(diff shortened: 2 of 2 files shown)


crivetimihai:
Rebased onto `main` (clean, no conflicts) and made minor cleanup:

- Moved `import orjson` from inside test function to module-level imports (PEP 8)
- Aligned `sse_generator()` docstring to match the `generate_events()` style used by the roots subscribe endpoint for consistency
- Ran `isort`/`black` to fix import ordering and slice spacing

No logic changes — the fix itself is correct and consistent with existing SSE patterns in the codebase.

PR #3597, opened 2026-03-10
fix: extract groups from id_token for generic OIDC providers

SharonDiskin:
## Summary

   Generic OIDC providers like Okta include group/role claims in the id_token but
   **not** in the /userinfo response. The existing code handles this for Entra ID and
   Keycloak with provider-specific extraction, but the generic OIDC code path drops
   groups entirely.

   This PR adds two fixes:

   1. **_get_user_info**: Extract groups and roles claims from the verified id_token
   for generic OIDC providers (any provider not explicitly handled: Entra, Keycloak,
   GitHub, Google). Only copies claims missing from the userinfo response to avoid
   overwriting.

   2. **_normalize_user_info**: Include groups in the normalized return dict for the
   generic OIDC path. Supports a configurable groups_claim key via provider_metadata,
   defaulting to groups. Also merges roles into the groups list, matching the Entra ID
    behavior.

   Without this fix, _apply_team_mapping always receives an empty groups list for
   generic OIDC providers, so SSO-based team assignment never works.

   ---

   ## Type of Change
   - [x] Bug fix

   ---

   ## Verification

   Tested with Okta SSO on a live deployment:
   - Configured Okta authorization server with groups scope and claim
   - Verified the authorize URL includes scope=openid+profile+email+groups
   - Confirmed team mapping config in sso_providers table is read correctly
   - User login creates user and triggers _apply_team_mapping flow

   **Okta requirement**: The groups claim must be configured on the ID Token (and
   optionally Access Token) in the Okta Authorization Server Claims settings, with a
   groups filter (e.g. Matches regex .\*).

   ---

   ## Checklist
   - [x] Code follows existing patterns (mirrors Entra ID and Keycloak handling)
   - [x] No breaking changes — only adds behavior for previously unhandled providers

   ---

   ## Notes
   - The groups_claim key is configurable via provider_metadata to support providers
   that use a non-standard claim name
   - The roles claim is merged into groups to support providers that use roles instead
    of groups for RBAC
   - The not in user_data guard ensures we never overwrite groups already returned by
   the userinfo endpoint" --jq '.html_url'

### DIFF
--- a/docs/config.schema.json
+++ b/docs/config.schema.json
@@ -341,6 +341,25 @@
       "description": "Okta issuer URL",
       "title": "Sso Okta Issuer"
     },
+    "sso_okta_scope": {
+      "default": "openid profile email",
+      "description": "Okta OIDC scopes (space-separated)",
+      "title": "Sso Okta Scope",
+      "type": "string"
+    },
+    "okta_group_mapping": {
+      "anyOf": [
+        {
+          "type": "string"
+        },
+        {
+          "type": "null"
+        }
+      ],
+      "default": null,
+      "description": "JSON mapping of Okta group names to team UUIDs",
+      "title": "Okta Group Mapping"
+    },
     "sso_keycloak_enabled": {
       "default": false,
       "description": "Enable Keycloak OIDC authentication",
--- a/docs/docs/config.schema.json
+++ b/docs/docs/config.schema.json
@@ -341,6 +341,25 @@
       "description": "Okta issuer URL",
       "title": "Sso Okta Issuer"
     },
+    "sso_okta_scope": {
+      "default": "openid profile email",
+      "description": "Okta OIDC scopes (space-separated)",
+      "title": "Sso Okta Scope",
+      "type": "string"
+    },
+    "okta_group_mapping": {
+      "anyOf": [
+        {
+          "type": "string"
+        },
+        {
+          "type": "null"
+        }
+      ],
+      "default": null,
+      "description": "JSON mapping of Okta group names to team UUIDs",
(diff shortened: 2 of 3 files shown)


crivetimihai:
Thanks @SharonDiskin — generic OIDC group extraction from id_token is needed for proper team/role mapping. Targeting 1.0.0.

crivetimihai:
Rebased onto current main and expanded the implementation. Here's what changed:

**Core fixes:**
- Okta and IBM Verify `_normalize_user_info` handlers now extract groups (configurable `groups_claim` via `provider_metadata`, `roles` merged into groups, non-string elements filtered, deduplicated)
- Removed Okta and IBM Verify from the exclusion list in `_get_user_info` so they use the generic OIDC id_token groups merge path (they have no dedicated extraction unlike Keycloak/Entra)
- Generic OIDC `_normalize_user_info` path now merges `roles` claim into groups (previously only Entra did this)
- Added `sso_okta_scope` and `okta_group_mapping` config settings
- Bootstrap uses configurable Okta scope, parses team mapping from JSON env var, preserves DB scope/team_mapping on restart

**Hardening (from review):**
- Single-string `roles` values are now handled (not just lists)
- `OKTA_GROUP_MAPPING` validates parsed JSON is a dict (non-dict values warn and fall back to empty)
- Bootstrap scope preservation only applies when env provides the default — a custom env scope correctly overrides DB

**Docs:**
- `.env.example`, `docs/docs/manage/configuration.md`, `docs/config.schema.json`, `docs/docs/config.schema.json` all updated with new settings

**Tests:** 41 new tests, 265 total SSO tests passing.

PR #3615, opened 2026-03-11
fix(metrics): aggregate from both raw and hourly rollup tables (#3598) 

### LINKED ISSUE #3598: [BUG][OBSERVABILITY]: Server tool metrics total_executions always returns 0
### 🐞 Bug Summary
When calling the endpoint /servers/{server_id}/tools?include_inactive=true&include_metrics=true, the returned total_executions value is always 0.

However, tools have been executed before and metrics should exist.

---

### 🧩 Affected Component

- [✅] `mcpgateway` - API
- [ ] Other (explain below)

---

### 🔁 Steps to Reproduce

Call the endpoint:

GET /servers/{server_id}/tools?include_inactive=true&include_metrics=true

Inspect the returned tool metrics.

Observe that the field total_executions is always 0.

---

### 🤔 Expected Behavior
total_executions should reflect the actual number of tool invocations.

If historical metrics have been aggregated into hourly tables, the API should still return the correct total execution count.

---

### 📓 Logs / Error Output
No error logs observed.

From initial investigation:

The query seems to read data from the table tool_metrics.

Currently, the tool_metrics table contains no records.

It appears that historical data has been aggregated into tool_metrics_hourly, and the tool_metrics table was cleared afterward.

Because of this, the API returns total_executions = 0.

---

### 🧠 Environment Info
1.0.0-BETA-2

---

It seems the execution metrics may have been migrated from tool_metrics into tool_metrics_hourly.

If that is the intended design, should the API aggregate data from both tables when calculating total_executions?

For example:

total_executions =sum(tool_metrics.executions)+ sum(tool_metrics_hourly.executions)

Otherwise, once tool_metrics is cleared, the API will always return 0, even if historical execution data exists.

rakdutta:
## 🐛 Bug Fix

  Fixes #3598 - Metrics API returns `total_executions = 0` after hourly rollup and cleanup.

  ## 📋 Problem

  When calling `/servers/{server_id}/tools?include_metrics=true` (and similar endpoints for resources/prompts/servers), the API returned `total_executions = 0` after raw metrics were aggregated into hourly rollup tables and cleaned up.

  This affected all production deployments using the default configuration where `METRICS_DELETE_RAW_AFTER_ROLLUP=true` (default behavior).

  **Root Cause**: The `metrics_summary` property only queried raw metrics tables (`tool_metrics`, etc.) and ignored hourly rollup tables (`tool_metrics_hourly`, etc.) where historical data resides after cleanup.

  ## ✅ Solution

  Modified `metrics_summary` property in all affected entity models to query **BOTH**:
  - **Raw metrics tables** - Recent data not yet rolled up/cleaned
  - **Hourly rollup tables** - Historical data after cleanup

  The fix properly aggregates:
  - Total/success/failure counts (sum from both sources)
  - Min/max response times (extremes from both sources)
  - Average response times (weighted average based on counts)
  - Last execution timestamp (most recent from either source)

  ## 📝 Changes

  ### Core Fix
  - `mcpgateway/db.py` - Fixed `metrics_summary` for all entities:
    - `Tool.metrics_summary`
    - `Resource.metrics_summary`
    - `Prompt.metrics_summary` 
    - `Server.metrics_summary` 

  ### Tests
  - `tests/unit/mcpgateway/test_db.py` - Updated 4 unit tests to mock dual queries:
    - `test_tool_metrics_summary_sql_path`
    - `test_resource_metrics_summary_sql_path`
    - `test_prompt_metrics_summary_sql_path`
    - `test_server_metrics_summary_sql_path`

  ## ✅ Verification

  ### Before Fix
  ```bash
  total_executions BEFORE deletion: 10
  total_executions AFTER deletion: 0  ❌ BUG

  After Fix

  total_executions BEFORE deletion: 10
  total_executions AFTER deletion: 10  ✅ FIXED

  Test Results

  - ✅ Reproduction test passes
  - ✅ All 30 metrics tests pass
  - ✅ Full test suite: 14,236 passed, 0 failed

  🎯 Impact

  Before

  - ❌ Dashboards showed zero executions after metrics rollup
  - ❌ Historical analytics disappeared
  - ❌ Monitoring/alerting systems failed

  After

  - ✅ Metrics persist across rollup/cleanup cycles
  - ✅ Historical data visible via API
  - ✅ Dashboards show correct execution counts
  - ✅ Monitoring/alerting systems work correctly

  🔧 Technical Details

  Query Strategy: Uses two queries per entity:
  1. Aggregate from raw metrics table (recent data)
  2. Aggregate from hourly rollup table (historical data)
  3. Combine results with proper weighted averages

  Edge Cases Handled:
  - ✅ No hourly data (new entities) → returns raw metrics only
  - ✅ No raw data (old entities) → returns hourly aggregates only
  - ✅ Both sources present → correctly aggregates from both
  - ✅ No data at all → returns zeros/nulls as expected

  📊 Performance

  - Query count: 1 → 2 queries per metrics_summary call
  - Impact: Minimal - simple aggregations with indexed foreign keys
  - Optimization: Future work could use UNION ALL or materialized views


  🧪 Testing

  # Run metrics tests
  pytest tests/unit/mcpgateway/test_db.py -k "metric" -v

  # Run full suite
  make test

### DIFF
--- a/mcpgateway/db.py
+++ b/mcpgateway/db.py
@@ -3435,49 +3435,16 @@ class Tool(Base):
     def metrics_summary(self) -> Dict[str, Any]:
         """Aggregated metrics for the tool.
 
-        When metrics are loaded: computes all values from memory in a single pass.
-        When not loaded: uses a single SQL query with aggregation for all fields.
+        Always queries both raw metrics and hourly rollups for complete historical coverage.
+        This ensures accurate metrics even after raw data cleanup (issue #3598).
 
         Returns:
             Dict[str, Any]: Dictionary containing aggregated metrics:
                 - total_executions, successful_executions, failed_executions
                 - failure_rate, min/max/avg_response_time, last_execution_time
         """
-        # If metrics are loaded, compute everything in a single pass
-        if self._metrics_loaded():
-            total = 0
-            successful = 0
-            min_rt: Optional[float] = None
-            max_rt: Optional[float] = None
-            sum_rt = 0.0
-            last_time: Optional[datetime] = None
-
-            for m in self.metrics:
-                total += 1
-                if m.is_success:
-                    successful += 1
-                rt = m.response_time
-                if min_rt is None or rt < min_rt:
-                    min_rt = rt
-                if max_rt is None or rt > max_rt:
-                    max_rt = rt
-                sum_rt += rt
-                if last_time is None or m.timestamp > last_time:
-                    last_time = m.timestamp
-
-            failed = total - successful
-            return {
-                "total_executions": total,
-                "successful_executions": successful,
-                "failed_executions": failed,
-                "failure_rate": failed / total if total > 0 else 0.0,
-                "min_response_time": min_rt,
-                "max_response_time": max_rt,
-                "avg_response_time": sum_rt / total if total > 0 else None,
-                "last_execution_time": last_time,
-            }
-
-        # Query both raw metrics and hourly rollups using helper function (issue #3598 fix)
+        # Always query both raw metrics and hourly rollups using helper function (issue #3598 fix)
+        # This ensures we get complete historical data even when raw metrics are cleaned up
         # Third-Party
         from sqlalchemy.orm import object_session  # pylint: disable=import-outside-toplevel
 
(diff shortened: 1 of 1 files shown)


rakdutta:
closing this PR. I will raise a new PR for this fix.

PR #3250, opened 2026-02-25
[FEATURE][RUST]: Rewrite A2A agent in Rust (#1624)

lucarlig:
TBD

PR #3161, opened 2026-02-24
feat: implement top-level Rust workspace (ADR-041)

crivetimihai:
> **Note:** This PR was re-created from #3029 due to repository maintenance. Your code and branch are intact. @lucarlig please verify everything looks good.

Implements [ADR-041: Top-Level Rust Workspace](docs/docs/architecture/adr/041-top-level-rust-workspace.md). All Rust code is now part of a single workspace at the repository root, with one set of commands and one CI workflow for building, testing, and releasing.

#### Summary

- **Workspace at repo root**  
  Root `Cargo.toml` and `Cargo.lock` define a workspace. Rust crates live under `crates/` (e.g. `crates/plugins/`, `crates/tools/`, `crates/gateway_rs/`, `crates/mcp-servers/`). A single `cargo build` / `cargo test` at the repo root builds and tests all Rust code.

- **Single Rust CI workflow**  
  Replaces the separate Rust Plugins and Rust Tools workflows with one [`.github/workflows/rust.yml`](.github/workflows/rust.yml) that:
  - Runs on `ubuntu-latest` and `macos-latest` (no Windows; Makefile assumes a Unix shell).
  - Triggers on changes under `crates/**`, root `Cargo.toml`/`Cargo.lock`, `rust-toolchain.toml`, `deny.toml`, and the workflow file.
  - Uses Makefile targets for: build + stub generation/verification, `rust-fmt-check`, `rust-lint` (clippy), `rust-test`, wheel build, `rust-audit`, `rust-licenses`, `rust-bench-check`, coverage, docs, and (on tags) release/publish.

- **Makefile as single entrypoint**  
  All Rust operations go through `make` targets (e.g. `rust-build-check`, `rust-fmt-check`, `rust-lint`, `rust-test`, `rust-build-wheels`, `rust-audit`, `rust-licenses`, `rust-coverage`, `rust-doc`, `rust-release` / `rust-release-publish`). CI and local dev use the same commands.

- **Pre-commit**  
  Adds a local hook that runs Rust fmt, build, and clippy when Rust is installed and when files under `crates/`, `Cargo.toml`, `Cargo.lock`, or `rust-toolchain.toml` change. Path exclusions and hook config were updated from `plugins_rust/` / `tools_rust/` to the new `crates/` paths.

- **Docs and config**  
  [AGENTS.md](AGENTS.md) and pre-commit config now refer to `crates/plugins/` and `crates/tools/` instead of `plugins_rust/` and `tools_rust/`.

[...]

PR #3617, opened 2026-03-11
feat: add experimental Rust MCP runtime and session core

crivetimihai:
## 🔗 Related Issue
N/A

---

## 📝 Summary
This PR adds an experimental Rust MCP runtime and incrementally moves the MCP transport, high-volume JSON-RPC routing, and the first session/event-store slices into Rust while keeping the Python implementation available behind feature flags.

Key changes:
- add `tools_rust/mcp_runtime` and integrate it into `Containerfile.lite`, `docker-compose.yml`, and the managed entrypoint flow
- route public `/mcp` traffic through Rust with explicit runtime visibility in `/health`, `/ready`, startup logs, and MCP response headers
- narrow `initialize`, `notifications/initialized`, `tools/list`, `tools/call`, `resources/*`, `prompts/*`, and several other MCP methods off the generic Python dispatcher
- wire the official `modelcontextprotocol/rust-sdk` behind a feature flag for the upstream `tools/call` client path
- add feature-flagged Rust session metadata and Redis-backed event-store/session sharing, including dedicated session cleanup for server-scoped `DELETE`
- extend the load-test harness and document the runtime boundary, configuration, and migration status in `tools_rust/mcp_runtime/README.md` and `tools_rust/mcp_runtime/STATUS.md`

This remains intentionally reversible:
- full Python fallback: `EXPERIMENTAL_RUST_MCP_RUNTIME_ENABLED=false`
- Rust transport/runtime only: disable `EXPERIMENTAL_RUST_MCP_SESSION_CORE_ENABLED` and `EXPERIMENTAL_RUST_MCP_EVENT_STORE_ENABLED`

---

## 🏷️ Type of Change
- [ ] Bug fix
- [x] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

| Check | Command | Status |
|---|---|---|
| Rust runtime crate tests | `cargo test --release --manifest-path tools_rust/mcp_runtime/Cargo.toml` | ✅ |
| MCP CLI parity | `make test-mcp-cli` | ✅ |
| MCP RBAC parity | `make test-mcp-rbac` | ✅ |
| Redis stream integration | `uv run pytest -q --with-integration tests/integration/test_streamable_http_redis.py` | ✅ |
| Compose-built Rust image | `docker compose build gateway` with Rust enabled | ✅ |
| Load benchmarks | Mixed MCP and tools-only Locust runs on the compose-built stack | ✅ |

Performance highlights on the compose-built Rust stack:
- mixed MCP workload: `1007.35 RPS` at `120` users, `1033.01 RPS` at `150` users, `0%` failures
- tools-only hot path: `1126.96 RPS` overall and `1068.5 RPS` on `MCP tools/call [rapid]` at `125` users, `0%` failures

---

## ✅ Checklist

---

## 📓 Notes
Current boundary:
- public MCP transport now flows through Rust
- Python still remains the control plane for auth/RBAC and some session/transport internals
- the runtime is intentionally feature-flagged so deployments can fall back to the Python path at any time

Follow-on work after this PR is the remaining transport/session ownership: replay/resume behavior, session existence/owner checks on the public path, and multi-worker session-affinity forwarding.

### DIFF
--- a/mcpgateway/services/metrics_buffer_service.py
+++ b/mcpgateway/services/metrics_buffer_service.py
@@ -151,7 +151,14 @@ class MetricsBufferService:
             logger.info("MetricsBufferService flush task started")
 
     def _flush_task_is_active_for_loop(self, loop: asyncio.AbstractEventLoop) -> bool:
-        """Return whether the current flush task is usable for the active loop."""
+        """Return whether the current flush task is usable for the active loop.
+
+        Args:
+            loop: The currently running event loop for this worker process.
+
+        Returns:
+            True when the cached flush task belongs to the current live loop.
+        """
         task = self._flush_task
         if task is None or task.done() or task.cancelled():
             return False


tools_rust/mcp_runtime/src/lib.rs
lucarlig, line 1: this file should be broken up in a number of files, leaving only the entrypoint for the lib in `lib.rs` to improve human/agent redability

something like:
- src/main.rs: App entry & server initialization.
- src/routes/: Route definitions (merging sub-routers).
- src/handlers/: Request/response logic (HTTP layer).
- src/models/: Data structures & Serde structs.
- src/services/: Core business logic (agnostic of HTTP).
- src/db/: Database connections & SQLx/Diesel queries.
- src/middleware/: Custom layers (auth, logging, etc.).

tools_rust/mcp_runtime/src/lib.rs
```diff
@@ -0,0 +1,4770 @@
...
+    }
+    inject_session_header(incoming_headers, &session_id);
+    if let Some(server_id) = record.server_id.as_deref() {
+        if !incoming_headers.contains_key("x-contextforge-server-id") {
+            inject_server_id_header(incoming_headers, server_id.to_string());
+        }
+    }
+}
+
+fn runtime_session_allows_access(
```
lucarlig: This session ownership check treats any` is_admin` token as globally authorized. Elsewhere in the codebase, admin bypass is only allowed when the token is unscoped (`teams == null`); team-scoped admins should still be restricted. The Rust runtime should mirror the same token-scope semantics here.

tools_rust/mcp_runtime/src/lib.rs
```diff
@@ -0,0 +1,4770 @@
...
+    if let Some(prefix) = backend_rpc_url.strip_suffix("/rpc/") {
+        return format!("{prefix}/_internal/mcp/tools/call/resolve");
+    }
+    format!(
+        "{}/_internal/mcp/tools/call/resolve",
+        backend_rpc_url.trim_end_matches('/')
+    )
+}
+
+fn build_db_pool(config: &RuntimeConfig) -> Result<Option<Pool>, RuntimeError> {
```
lucarlig, line 2041: This introduces a second DB access layer in Rust alongside the existing Python one. That means schema and auth/visibility semantics now have to stay aligned across both implementations. Can we converge on a single shared data layer instead of maintaining two?

tools_rust/mcp_runtime/src/lib.rs
```diff
@@ -0,0 +1,4770 @@
...
+    ))
+}
+
+async fn query_server_tools_list_from_db(
+    state: &AppState,
+    server_id: &str,
+    auth_context: &InternalAuthContext,
+) -> Result<Vec<McpToolDefinition>, RuntimeError> {
+    let pool = state
+        .db_pool()
```
lucarlig, line 4512: This query duplicates Python-owned DB/business logic in Rust. I’m worried about long-term drift every time migrations or filtering rules change. It would be safer to either keep this behind the Python DB/service layer or move to one Rust DB layer that Python also reuses.

tools_rust/mcp_runtime/src/lib.rs
```diff
@@ -0,0 +1,4770 @@
...
+    resolved_tool_call_plans: Arc<Mutex<HashMap<String, CachedResolvedToolCallPlan>>>,
+    tools_call_plan_ttl: Duration,
+    upstream_session_ttl: Duration,
+    session_ttl: Duration,
+}
+
+#[derive(Debug, Clone, Deserialize)]
+pub struct JsonRpcRequest {
+    pub jsonrpc: Option<String>,
+    pub method: String,
```
lucarlig, line 186: The internal Python -> Rust proxy is configured with `follow_redirects=True`. Since this hop carries forwarded auth/context headers, transparently following redirects increases the chance of leaking trusted headers to an unintended target and hides misconfiguration. This internal sidecar hop should fail closed on redirects instead.

tools_rust/mcp_runtime/src/lib.rs
```diff
@@ -0,0 +1,4770 @@
...
+        "python"
+    }) {
+        response
+            .headers_mut()
+            .insert(HeaderName::from_static(RESUME_CORE_HEADER), value);
+    }
+    response
+}
+
+async fn active_runtime_session_count(state: &AppState) -> usize {
```
lucarlig, line 2844: The health path counts active sessions with Redis `KEYS`, which is a blocking full-keyspace scan. That makes `health` O(N) in the number of matching session keys and can become expensive under frequent probes or larger deployments. A maintained counter or a bounded `SCAN` style approach would avoid turning health checks into Redis hot spots.

tools_rust/mcp_runtime/src/lib.rs
```diff
@@ -0,0 +1,4770 @@
...
+        Ok(replay) => replay,
+        Err(response) => return response,
+    };
+    let protocol_version = incoming_headers
+        .get(MCP_PROTOCOL_VERSION_HEADER)
+        .and_then(|value| value.to_str().ok())
+        .unwrap_or(state.protocol_version())
+        .to_string();
+
+    let keep_alive = KeepAlive::new()
```
lucarlig: Each resumed SSE connection runs its own poll loop against Redis, sleeping and replaying repeatedly until the session disappears. Functionally this works, but it creates steady per-connection Redis load for idle streams, so the cost scales linearly with open resumable connections. A push/notify-based wakeup would scale much better here.

sco3: Get my "yes" to this direction and I would eventually create gateway in rust with possibility to use some python extension if needed (say prototype extension in python and if it proves to be useful to customer then implement in rust).   For instance Nginx unit provides the functionality when several backends may be combined together. 
The pet project https://github.com/sco3/nginx-unit-playground shows how one can have an nginx sever serving as integration point for go backend and two python backends (one on fastapi (pydantic models) and another on litestar (msgspec models))at the same time. Beside the mix of different languages backends it gives performance improvements to python backends. 
```
Go Unit              █████████████████████████████████████████████ 43,230 req/s
Litestar Unit        ████████████████ 16,753 req/s
FastAPI Unit         ███████████████ 14,799 req/s
Litestar + Uvicorn   ███████ 7,006 req/s
FastAPI + Uvicorn    ██████ 6,069 req/s
FastAPI + Gunicorn   ██████ 5,995 req/s
```

sco3:
I tried to use cloudflare pingora crate as mcp proxy to fast-time-server:
After some optimizations:
```
Starting benchmark: Direct (8111) (10 clients × 10000 requests = 100000 total)

Direct (8111) Results:
   Total: 100000 requests (100000 success, 0 failed)
   Elapsed: 2.13s
   Avg latency: 0.21ms
   Throughput: 47020.85 req/s

Starting benchmark: Gateway (3000) (10 clients × 10000 requests = 100000 total)

Gateway (3000) Results:
   Total: 100000 requests (100000 success, 0 failed)
   Elapsed: 4.22s
   Avg latency: 0.42ms
   Throughput: 23701.87 req/s
```

jonpspri:
@crivetimihai A lot of change in main.py here.  Let's refactor into a separate `service` or `transport`.  I'm not sure which because there's still a bit of clarity to be created around what does what in the object architecture.

jonpspri:
> Interesting PR, i love the general idea of rust components being fully independent. However I have been working on the `a2a_service` in #3250, and one of the main challenges in fully replacing the Python service in this case is how the database layer is handled. The current Python implementation already contains assumptions around schema management, migrations, connection handling, and ORM/query behavior. Re-implementing or partially replacing that logic can easily introduce inconsistencies (e.g., differences in transaction handling, migration state, connection pooling, or query semantics), which makes the database layer particularly tricky to replicate safely.
> 
> Using a Python fallback may not provide the same feature set, safety guarantees, or performance characteristics. It could also behave differently under load, which makes this approach quite risky for production use.
> 
> if this PR were merged and then integrated with the changes introduced in #3161, it would benefit from an existing and more robust CI setup. For that reason, I would recommend working toward merging #3161 first and placing this implementation within the `crates` folder so it can leverage that infrastructure.

I've been noodling on persistence in the overall application.  We need a separate persistence manager (platform TBD) where component modules don't have to be as deep in the innards.  I'm working through a refactoring of the object model of the application to support these initiatives and agree we need strong separation of concerns.  I'll create a separate epic/ADR with my thoughts today while I'm flying.

tools_rust/mcp_runtime/src/lib.rs
jonpspri: @sco3 The first "gateway" we'll need is for post_tool_hook plugin execution, which occurs in-stream.  This will be complicated by the fact that the entire plugin module is being factored into an independent component and will be available to our via an API and dependency injection.  I'm trying to sort the component module map for the future state of the system ASAP.  I'll likely call you guys from NYC on Wednesday or Thursday to discuss.

tools_rust/mcp_runtime/src/lib.rs
```diff
@@ -0,0 +1,10066 @@
...
+        .map_err(|err| {
+            error!("Rust event store replay scan failed: {err}");
+            json_response(
+                StatusCode::BAD_GATEWAY,
+                json!({"detail": "Rust event store replay scan failed"}),
+            )
+        })?;
+
+    let mut events = Vec::with_capacity(event_ids.len());
+    for event_id in event_ids {
```
gandhipratik203, line 3484: The replay path currently performs **one awaited `HGET` per event ID** (lines **3840–3842**), which means the replay cost becomes **O(n) Redis round-trips**. On a long stream this will dominate latency, even if the per-event JSON processing itself is cheap.

Redis provides a way to avoid this pattern. The `HMGET` command accepts **multiple fields in a single call**, so the replay can be reduced to **two round-trips regardless of stream length**:

1. `ZRANGEBYSCORE` to retrieve the event IDs
2. `HMGET` to retrieve all payloads in one request

Example:

```rust
let messages: Vec<Option<String>> = redis::cmd("HMGET")
    .arg(&messages_key)
    .arg(&event_ids)
    .query_async(&mut redis)
    .await?;
```

Alternatively, if maintaining **individual `HGET` calls** is preferred for error isolation, the same batching benefit could be achieved using a Redis pipeline:

```rust
let mut pipe = redis::pipe();

for id in &event_ids {
    pipe.cmd("HGET").arg(&messages_key).arg(id);
}

let messages: Vec<Option<String>> = pipe.query_async(&mut redis).await?;
```

Both approaches reduce the number of network round-trips significantly and should improve replay latency on longer streams.

mcpgateway/transports/rust_mcp_runtime_proxy.py
```diff
@@ -0,0 +1,267 @@
...
+                        "type": "http.response.start",
+                        "status": response.status_code,
+                        "headers": [(name, value) for name, value in response.headers.raw if name.decode("latin-1").lower() not in _RESPONSE_HOP_BY_HOP_HEADERS],
+                    }
+                )
+                async for chunk in response.aiter_bytes():
+                    if chunk:
+                        await send({"type": "http.response.body", "body": chunk, "more_body": True})
+                await send({"type": "http.response.body", "body": b"", "more_body": False})
+        except httpx.HTTPError as exc:
```
gandhipratik203, line 112: Since the Rust runtime is still marked **experimental**, should `httpx.HTTPError` fall back to `python_fallback_app` instead of returning a hard **502**?

Right now, if the sidecar is **down or restarting**, `/mcp` becomes fully unavailable even though a working Python path is already wired up and used on **lines 79 and 84** for other cases.

A graceful fallback on transport error would make **rollout and recovery safer** – operators could enable the Rust runtime without it becoming an **all-or-nothing risk**.

Example:

```python
except httpx.HTTPError as exc:
    logger.warning("Rust MCP runtime unavailable, falling back to Python: %s", exc)
    await self.python_fallback_app(scope, receive, send)
    return
```

If the hard **502** is intentional (to make sidecar failures visible rather than silently hiding them), ignore this comment 

OR

a configuration flag such as:

```
EXPERIMENTAL_RUST_MCP_RUNTIME_FALLBACK_ON_ERROR
```

could give operators the choice between **strict failure** and **graceful fallback**.

tools_rust/mcp_runtime/src/lib.rs
```diff
@@ -0,0 +1,10066 @@
...
+async fn get_runtime_session_from_redis(
+    state: &AppState,
+    session_id: &str,
+) -> Option<RuntimeSessionRecord> {
+    let mut redis = state.redis().await?;
+    let key = runtime_session_key(state, session_id);
+    match redis.get::<_, Option<String>>(&key).await {
+        Ok(Some(payload)) => {
+            if let Ok(ttl_i64) = i64::try_from(state.session_ttl().as_secs()) {
+                let _ = redis.expire::<_, bool>(&key, ttl_i64).await;
```
gandhipratik203, line 2972: The session read path currently performs a `GET` followed by a separate `EXPIRE` to refresh the TTL (lines **3322–3325**). This has two issues:

1. **Two round-trips on every session lookup** – one `await` for the read and another `await` for the refresh.
2. **Non-atomic behaviour** – if the key expires between the two calls, the session is returned to the caller but its TTL is never refreshed, silently leaving it closer to expiry than expected.

Since the project already uses `redis:latest` (Redis **7.x**), the `GETEX` command is available and combines both operations into a single atomic call:

```rust
let payload: Option<String> = redis::cmd("GETEX")
    .arg(&key)
    .arg("EX")
    .arg(state.session_ttl().as_secs())
    .query_async(&mut redis)
    .await
    .ok()
    .flatten();
```

This keeps the hot path to **one Redis round-trip** and removes the atomicity gap.

tools_rust/mcp_runtime/src/lib.rs
```diff
@@ -0,0 +1,10066 @@
...
+                json_response(
+                    StatusCode::BAD_GATEWAY,
+                    json!({"detail": "Rust event store replay fetch failed"}),
+                )
+            })?
+        else {
+            continue;
+        };
+
+        let message = serde_json::from_str::<Value>(&message_json).unwrap_or(Value::Null);
```
gandhipratik203: `serde_json::from_str::<Value>(&message_json).unwrap_or(Value::Null)` silently converts a decode failure into a **null payload** that gets pushed into the replay response as a normal event. On a resume/replay path this feels risky – corrupted or partially written data becomes indistinguishable from an intentional `null` message and the stream continues without any signal that something went wrong.

There is also an inconsistency with **line 3851** – a missing key is already handled with `continue` (skip the event), but a corrupt payload takes the opposite path and emits it as `null`. In Rust, preserving the `Result` here usually gives safer semantics than recovering to a sentinel value.

At minimum, logging the decode error before skipping the event would make failures observable:

```rust id="qgqj3t"
match serde_json::from_str::<Value>(&message_json) {
    Ok(message) => events.push(EventStoreReplayEvent { event_id, message }),
    Err(err) => warn!("event store replay: skipping corrupt event {event_id}: {err}"),
}
```

tools_rust/mcp_runtime/src/lib.rs
```diff
@@ -0,0 +1,10066 @@
...
+    if let Some(session_id) = upstream_session_id.as_deref() {
+        let _ =
+            send_initialized_notification(state, server_url, plan, protocol_version, session_id)
+                .await;
+    }
+
+    Ok(upstream_session_id)
+}
+
+async fn send_initialized_noti
```
gandhipratik203, line 7275: `DefaultHasher` is not stable across Rust versions or targets — two workers on different toolchains can compute different Redis keys for the same input. The failure mode is silent: missed plan-cache hits and duplicate upstream sessions with no error surfaced.

Since `sha2::Sha256` is already imported on line 35, the fix is a straight swap in both key builders:

```rust
// replace DefaultHasher with this in build_upstream_session_key and build_tools_call_plan_cache_key
let mut hasher = Sha256::new();
hasher.update(tool_name.as_bytes());

for (k, v) in &ordered_headers {
    hasher.update(k.as_bytes());
    hasher.update(b"=");
    hasher.update(v.as_bytes());
    hasher.update(b"\n");
}

format!("tool-plan:{:x}", hasher.finalize())
```

gandhipratik203, line 7075: Rust/concurrency note: all three caches follow the same pattern — **check under lock → drop lock → perform expensive `.await` → reacquire lock and insert**. This means concurrent requests for the same key can all miss, all perform the same initialisation in parallel, and then race to populate the cache:

* **`ensure_upstream_session`** — lines **7344, 7354, 7356**: upstream session initialisation duplicated; last writer wins but earlier sessions are kept
* **RMCP client cache** — lines **7552, 7569, 7575**: same pattern, but no re-check before insert, so earlier clients are silently orphaned with no cleanup
* **`resolved_tool_call_plans`** — lines **7116, 7126, 7128**: plan resolution duplicated on concurrent misses

A `tokio::sync::OnceCell` per key provides **single-flight semantics**, ensuring only one task performs the initialisation while others await the same result:

```rust id="j3h7a1"
// DashMap<String, Arc<OnceCell<Arc<RmcpRunningService<...>>>>>
let cell = state.rmcp_client_cells
    .entry(session_key.to_string())
    .or_insert_with(|| Arc::new(OnceCell::new()))
    .clone();

let client = cell.get_or_try_init(|| async {
    create_rmcp_client(state, plan, protocol_version).await
}).await?;
```

Worth addressing before this moves out of **experimental**, particularly the RMCP path where orphaned clients can accumulate silently.

tools_rust/mcp_runtime/src/lib.rs
crivetimihai: Thanks, will test this soon. Plan is to get the PR in shape for merge by Monday morning, as an experimental, opt-in feature. We can improve further later.

tools_rust/mcp_runtime/src/lib.rs
crivetimihai: Agreed, will do some further testing on this, thanks!

tools_rust/mcp_runtime/src/lib.rs
crivetimihai: Great find - will benchmark this, but should help with performance!

mcpgateway/transports/rust_mcp_runtime_proxy.py
crivetimihai: I like the idea of a feature flag, defaulting to no. Right now, Rust is enabled through a separate build + env. variables and not yet ready for production use. If it fails, it should fail hard, by default.

tools_rust/mcp_runtime/src/lib.rs
```diff
@@ -0,0 +1,10871 @@
...
+    // answer "replay everything after event X" without scanning all streams.
+    let Some(mut redis) = state.redis().await else {
+        return Err(json_response(
+            StatusCode::SERVICE_UNAVAILABLE,
+            json!({"detail": "Rust Redis event store is unavailable"}),
+        ));
+    };
+
+    let event_id = Uuid::new_v4().to_string();
+    let key_prefix = event_store_key_prefix(state, request.key_prefix.as_deref());
```
gandhipratik203, line 3353: Since this is going in as an **experimental opt-in with improvements to follow**, just noting this here as a **possible direction for future performance improvements** if and when this path is optimised further.

The natural next step on the performance side could be **offloading async work** — session state sync, audit logging, metrics — onto **Redis Streams**, which the PR already uses for the event store. Keeping that pattern consistent would allow the synchronous **Rust → Python** path to remain lean without requiring any transport changes.

If payload size becomes a bottleneck later, **MessagePack** could be a relatively low-effort drop-in replacement for JSON, with mature support in both `rmp-serde` (Rust) and `msgpack` (Python).

If MessagePack is not sufficient and raw throughput needs to compete with other Rust-native MCP gateways, **Protobuf** (`prost` in Rust, `betterproto` in Python) could be considered as a further step, though it would require upfront schema definitions and code generation on both sides.

| Option                          | Effort | Gain   | Best for                                                                 |
|---------------------------------|--------|--------|---------------------------------------------------------------------------|
| Redis Streams for async offload | Medium | High   | Session sync, audit, metrics                                             |
| MessagePack over HTTP           | Low    | Medium | Reducing hot-path payload size                                           |
| Protobuf                        | High   | High   | Future use — if competing on raw throughput with other Rust-native MCP gateways justifies the investment |

tools_rust/mcp_runtime/src/lib.rs
gandhipratik203: Makes sense. Given the experimental scope, kept the review focused on `correctness` and `performance`. Structural concerns can be addressed later.

Looking forward to seeing this merged!

crivetimihai:
Addressed the latest actionable review items and revalidated them on a rebuilt full-Rust stack.

What changed:
- Internal Python→Rust transport calls now fail closed on redirects instead of following them.
- The Rust event-store replay path now fails closed on corrupt stored entries with a `502`, instead of silently skipping replayed events.
- Added behavior-level regression coverage for both cases:
  - Rust runtime proxy surfaces redirects without following them
  - Rust event-store replay rejects redirects without following them
  - Rust event-store replay fails closed on corrupt messages

Validation run after the changes:
- `uv run pytest tests/unit/mcpgateway/transports/test_rust_mcp_runtime_proxy.py tests/unit/mcpgateway/transports/test_streamablehttp_transport.py -q`
- `cargo test --manifest-path tools_rust/mcp_runtime/Cargo.toml`
- `make testing-rebuild-rust-full`
- `make test-mcp-cli`
- `make test-mcp-session-isolation`
- `make benchmark-mcp-tools`

Results:
- `make test-mcp-cli`: `23 passed`
- `make test-mcp-session-isolation`: `10 passed`
- `make benchmark-mcp-tools`: `2127.16 RPS`, `0` failures

One important correction from the review pass:
- I initially tried “skip corrupt replay entries and continue”, but that was too permissive because it could silently lose replayed events.
- I changed that to fail closed instead, which is the safer behavior for resumable session integrity.

I left broader non-blocking items as follow-up work in `tools_rust/mcp_runtime/FOLLOWUPS.md` rather than widening this PR further.

tools_rust/mcp_runtime/src/lib.rs
crivetimihai: Agreed, and tracked in tools_rust/mcp_runtime/FOLLOWUPS.md but left for a post-merge follow-up.

tools_rust/mcp_runtime/src/lib.rs
crivetimihai: Agreed, and tracked in tools_rust/mcp_runtime/FOLLOWUPS.md but left for a post-merge follow-up. Not in scope for this initial PR.

tools_rust/mcp_runtime/src/lib.rs
crivetimihai: Agreed, and tracked in tools_rust/mcp_runtime/FOLLOWUPS.md but left for a post-merge follow-up.

tools_rust/mcp_runtime/src/lib.rs
crivetimihai: What changed:

- internal Python→Rust transport calls now fail closed on redirects instead of following them
- both the shared UDS clients and the request calls were hardened:
    - rust_mcp_runtime_proxy.py
    - streamablehttp_transport.py

Validation added:

- unit regression coverage proves redirects are surfaced without being followed:
    - test_rust_mcp_runtime_proxy.py
    - test_streamablehttp_transport.py

Post-change verification:

- rebuilt full Rust stack
- make test-mcp-cli
- make test-mcp-session-isolation
- make benchmark-mcp-tools

All stayed green.

tools_rust/mcp_runtime/src/lib.rs
crivetimihai: Thanks, good catch. Addressed earlier in this PR: the Rust health path no longer uses Redis KEYS; it now uses a bounded SCAN-based approach in tools_rust/mcp_runtime/src/lib.rs, so health checks no longer do a full keyspace scan.

tools_rust/mcp_runtime/src/lib.rs
crivetimihai: Thanks, agreed this is a real scalability concern. The current polling model is functionally correct, but it does create steady per-connection Redis load for idle resumable streams. I left this out of the current PR to avoid widening the design late, and tracked it as follow-up work in tools_rust/mcp_runtime/FOLLOWUPS.md under resumed SSE polling scalability / push-notify improvement.

tools_rust/mcp_runtime/src/lib.rs
crivetimihai: Thanks, agreed. This is a real performance improvement for longer replays, but I left it out of this PR to keep the current merge scoped to correctness/parity/security for the experimental opt-in release. I've tracked it in tools_rust/mcp_runtime/FOLLOWUPS.md under Redis hot-path polish, including batching replay fetches via HMGET/pipeline.

tools_rust/mcp_runtime/src/lib.rs
crivetimihai: Thanks, agreed. This is a solid hot-path optimization and also removes the small atomicity gap, but I left it out of this PR to keep the current merge scoped to correctness/parity/security for the experimental rollout. I've tracked it in tools_rust/mcp_runtime/FOLLOWUPS.md under Redis hot-path polish, specifically GET + EXPIRE vs GETEX.

tools_rust/mcp_runtime/src/lib.rs
crivetimihai: Thanks, fair point in principle, but I checked the current implementation and these DefaultHasher values are only used for in-process cache keys (resolved_tool_call_plans, upstream_tool_sessions, rmcp_upstream_clients), not shared Redis keys or other cross-worker state. So I don’t think this is a correctness issue for the current PR. I did leave it tracked as follow-up in tools_rust/mcp_runtime/FOLLOWUPS.md because switching to a stable hash would be cleaner and would future-proof these keys if we ever move them to shared/distributed caching.

crivetimihai: Thanks, agreed. This is a real single-flight/cache-shape improvement, especially for the RMCP client cache where duplicate initialisation can orphan earlier clients. I left it out of this PR to keep the current merge scoped to correctness/parity/security for the experimental rollout, but it is tracked in tools_rust/mcp_runtime/FOLLOWUPS.md under cache single-flight / hot-path polish.

tools_rust/mcp_runtime/src/lib.rs
crivetimihai: Thanks, this is useful directionally. I’m treating this as follow-up architecture/performance work rather than something to fold into the current PR. For this branch I wanted to keep the transport/wire contract stable and get the experimental Rust MCP path merged with correctness/parity/security in place first. If we push further on throughput after this lands, Redis-backed async offload and possibly MessagePack/Protobuf on the internal seams are good next directions.

mcpgateway/transports/rust_mcp_runtime_proxy.py
crivetimihai: Thanks. This is intentional for the current experimental rollout. If Rust mode is enabled, I want sidecar transport failures to be visible rather than silently falling back to Python, because silent fallback could hide runtime failures and also bypass Rust-specific behavior we’re explicitly trying to validate in this PR. So I left the hard 502 behavior in place by design. A configurable fallback flag could be a reasonable future enhancement, but I didn't want to change that failure model in this branch.

tools_rust/mcp_runtime/src/lib.rs
crivetimihai: Thanks, agreed. I revisited this after more testing and changed the implementation, but not in the exact "log and skip" direction from the original suggestion.

I initially tried the permissive approach, but on second review that felt too risky because it can silently lose replayed events. So the final behavior in this PR is:

- corrupt replay payloads no longer become null
- corrupt replay payloads also no longer get silently skipped
- the replay path now fails closed with a 502 and logs the decode failure

So the status here is:

- addressed in this PR
- final behavior is stricter than the original suggestion
- implemented in tools_rust/mcp_runtime/src/lib.rs
- covered by regression test in tools_rust/mcp_runtime/tests/runtime.rs

tools_rust/mcp_runtime/src/lib.rs
sco3: > Option 	Effort 	Gain 	Best for
> Redis Streams for async offload 	Medium 	High 	Session sync, audit, metrics
> MessagePack over HTTP 	Low 	Medium 	Reducing hot-path payload size
> Protobuf 	High 	High 	Future use — if competing on raw throughput with other Rust-native MCP gateways justifies the investment

When I tested protobuf in pyhton it was slightly slower than msgpack. And msgpack works as schema free alternative, which can give some flexibility. 
Today I re-ran the tests (3.14, originally it was 3.11):
```
msg_pack.py Took: 1702 ms 587544.0658049354 msg/s  message size: 292
protobuf.py Took: 3444 ms 290360.0464576074 msg/s  message size: 317

```

tools_rust/mcp_runtime/src/lib.rs
```diff
@@ -0,0 +1,10510 @@
...
+        let Some(actual_fingerprint) = auth_binding_fingerprint(incoming_headers) else {
+            return Err(SessionAccessDenyReason::MissingAuthBindingFingerprint);
+        };
+        if actual_fingerprint != expected_fingerprint {
+            return Err(SessionAccessDenyReason::AuthBindingMismatch);
+        }
+    }
+
+    let Some(owner_email) = record.owner_email.as_deref() else {
+        return Ok(());
```
gcgoncalves, line 3610: AI generated suggestion:

When a session is created without auth (unauthenticated request, `owner_email=None`) and `session_auth_reuse_enabled=false` (the default, which also clears `auth_binding_fingerprint`), both guards are inactive. Any caller that knows (or guesses) the `session_id` can access the session unchecked by Rust's layer. Subsequent callers could then re-initialize with their own auth and bind their credentials to the session, effectively stealing it.

tools_rust/mcp_runtime/src/lib.rs
```diff
@@ -0,0 +1,10510 @@
...
+    uri: &axum::http::Uri,
+    request: &JsonRpcRequest,
+) -> Option<String> {
+    runtime_session_id_from_request(incoming_headers, uri).or_else(|| {
+        request
+            .params
+            .get("session_id")
+            .or_else(|| request.params.get("sessionId"))
+            .and_then(Value::as_str)
+            .map(str::to_string)
```
gcgoncalves, line 3633: AI generated comment:

Clients can inject a `session_id` in the initialize payload, targeting an existing session. Combined with the ownerless bypass above, this enables a targeted session takeover when `session_core_enabled=true`. No length or format validation either (no UUID check, no max length bound), making this also a potential vector for oversized Redis keys.

Lang-Akshay:
# PR findings Summary (Ignore low )
 
| # | File | Line | Severity | CWE | Description |
|---|------|------|----------|-----|-------------|
| 1 | `nginx.conf` | 571| High | CWE-441 | Nginx `mcp_transport_backend` does not strip `x-contextforge-*` internal trust headers from public requests — external clients can spoof trusted internal headers |
| 2 | `main.py` | 6743+ | High | CWE-209 | 15+ internal MCP error handlers leak `str(exc)` in JSON-RPC data field, potentially exposing stack traces, SQL errors, or file paths to external clients |
| 3 | `rust_mcp_runtime_proxy.py` | 176 | Medium | CWE-918 | `_build_runtime_mcp_url` merges query strings without validation — potential SSRF parameter injection if runtime URL is misconfigured |
| 4 | `docker-entrypoint.sh` | 131 | Medium | CWE-668 | `MCP_RUST_PUBLIC_LISTEN_HTTP` defaults to `0.0.0.0:8787` in edge mode — sidecar may accept unauthenticated traffic directly |
| 5 | `main.py` | 360 | Medium | CWE-290 | `_is_trusted_internal_mcp_runtime_request` relies on loopback IP + header — spoofable in shared network namespace deployments |
| 6 | `main.py` | 395 | Medium | CWE-863 | `_build_internal_mcp_forwarded_user` accepts `is_admin` from forwarded auth context without re-verification |
| 7 | `token_scoping.py` | 1341 | Low | CWE-863 | Internal `/_internal/mcp/*` bypass has no HMAC/shared-secret verification beyond loopback + header |
| 8 | `rust_mcp_runtime_proxy.py` | 230 | Low | CWE-319 | Auth context forwarded via base64 (unsigned) — interceptable on non-UDS paths |
| 9 | `nginx-performance.conf` | 171 | Low | CWE-441 | Same header-stripping gap as Finding #1 |
| 10 | `main.py` | 9800 | Info | CWE-209 | DB connection error message may expose connection strings |
 
---
 
## Suggestions
 
1. **Finding 1** — Add `proxy_set_header x-contextforge-mcp-runtime "";` and similar for all `x-contextforge-*` internal headers to the nginx MCP location blocks in all nginx configs. Effort: trivial. OWASP: Proxy/Header Injection.
 
2. **Finding 2** — Replace `str(exc)` in error data fields with a generic message (e.g., `"See server logs"`). Log full exceptions server-side. Effort: low. CWE-209.
 
3. **Finding 3** — Validate incoming query strings contain only expected MCP parameters. Effort: low. CWE-918.
 
4. **Finding 4** — Default `MCP_RUST_PUBLIC_LISTEN_HTTP` to `127.0.0.1:8787`. Effort: trivial. CWE-668.
 
5. **Finding 5** — Add a shared secret or per-boot nonce to internal trust verification. Effort: medium. CWE-290.
 
6. **Finding 6** — Re-verify admin status for privilege-escalating operations, or sign the forwarded context. Effort: medium. CWE-863.

sco3:
rs: 
```
--------------------------------------------------------------------------------------
Name                                              Reqs    Fails      RPS  Avg(ms)  p99(ms)
--------------------------------------------------------------------------------------

MCP tools/call [rapid]                         108,692        0   1815.5      3.0     11.0
MCP tools/list [rapid]                           5,383        0     89.9      5.4     15.0
MCP initialize                                     125        0      2.1     39.7     88.0

```
py:
```
--------------------------------------------------------------------------------------
Name                                              Reqs    Fails      RPS  Avg(ms)  p99(ms)
--------------------------------------------------------------------------------------

MCP tools/call [rapid]                          45,156        0    758.5     91.4    270.0
MCP tools/list [rapid]                           2,229        0     37.4     91.1    270.0
MCP initialize                                     125        0      2.1     64.5    460.0
``` 
rs, 300 users/4 workers:
```
--------------------------------------------------------------------------------------
  Name                                              Reqs    Fails      RPS  Avg(ms)  p99(ms)
  --------------------------------------------------------------------------------------
  MCP tools/call [rapid]                         210,204        1   3552.5     15.5    110.0
  MCP tools/list [rapid]                          10,529        0    177.9     23.5    140.0
  MCP initialize                                     300        0      5.1     50.5    120.0
```

gcgoncalves:
Tested the functionalities with both the rust mode disabled and with full rust mode, both worked well for all basic functions. For the load tests, I've faced a slightily higher response time and lower rps against the python version for 800 users. Really odd.

tools_rust/mcp_runtime/src/lib.rs
crivetimihai: Thanks, I dug into this one in detail.

What I verified:

- The broad "session stealing" claim does not reproduce on the normal auth-required stack.
- With two different authenticated users targeting the same client-chosen initialize session id:
    - first caller initializes successfully
    - second caller gets -32003 Access denied
- This is also not Rust-only behavior:
    - Python already supports client-supplied initialize session ids too
    - so this is shared MCP/session compatibility behavior, not something newly introduced by the Rust path

What is still true:
- Ownerless sessions are intentionally allowed today.
- In permissive/public MCP mode, custom client-chosen initialize session ids plus ownerless-session semantics are worth reviewing more
  carefully.

I did not change that behavior in this PR because:

I tracked that narrower concern in tools_rust/mcp_runtime/FOLLOWUPS.md as a coordinated Python+Rust follow-up around client-supplied
initialize session ids / ownerless public-session semantics.

tools_rust/mcp_runtime/src/lib.rs
crivetimihai: Thanks, I checked this one closely too.

A few important clarifications:

- This is not Rust-only behavior.
    - Python already accepts client-supplied initialize session ids as well.
    - So this is a shared compatibility choice, not something newly introduced by the Rust runtime.
- On the normal auth-required stack, I could not reproduce authenticated takeover.
    - first caller initializes with a chosen session id
    - second caller targeting the same id is denied with -32003 Access denied

So I did not remove this behavior in this PR.

What I do agree with:

- custom client-supplied initialize session ids are more security-sensitive than server-emitted opaque ids
- the current behavior has no explicit UUID/length validation on the Rust side
- in permissive/public MCP mode, this deserves a broader design review together with ownerless-session semantics

Because this is shared Python/Rust behavior and not a confirmed new Rust regression, I left it as follow-up rather than changing protocol behavior late in this PR. I tracked it in tools_rust/mcp_runtime/FOLLOWUPS.md as a coordinated Python+Rust cleanup around client-supplied initialize session ids, validation, and public-session semantics.

crivetimihai:
> @Lang-Akshay # PR findings Summary (Ignore low )

Thanks, I went through this list and split it into "addressed in this PR" vs "tracked follow-up".

Addressed in this PR

- #5 / #7: fixed in d6c2335ec (fix: harden internal MCP trust boundaries)
    - trusted /_internal/mcp/* requests no longer rely on loopback + header alone
    - they now require:
        - loopback peer
        - x-contextforge-mcp-runtime: rust
        - a shared-secret-derived x-contextforge-mcp-runtime-auth
    - Rust injects that header on internal hops and strips any client-supplied copy on public ingress
    - I also revalidated the original spoof case directly against Uvicorn; it now returns 403

No longer a merge blocker after the above fix

- #1 / #9: still worth doing as defense in depth, but not a direct bypass anymore
    - nginx should still strip internal x-contextforge-* headers from public requests
    - however, Python no longer trusts those headers by themselves, so this is now follow-up hardening rather than an active trust-boundary hole
- #6: materially downgraded by the same trust-boundary fix
    - forwarded auth context is now accepted only on the authenticated local Rust->Python internal channel
    - query validation may still be worth tightening later, but I don’t think this is a merge blocker for this PR

So the main immediate trust-boundary issues from this list were #5 / #7, and those are now fixed. The rest are either follow-up hardening or no longer blockers after the
internal-channel change.

cafalchio:
I really like the idea of running rust on the core and  slowly implement new modules in rust as we need it.

I am still testing and reading the docs. But I ran docker with full rust and manually tested from the UI. It is all running as expected, including adding an external MCP provider (Claudflare) and running its tools. 

Suggestions.

- cargo clippy is not passing for mpc_runtime
- Not for now, but the lib.rs in mcp_runtime could be broken in modules.

 AI catches.

- Running wrapper can show the entire config with debug on.

`LOG_LEVEL=debug MCP_AUTH="Bearer my-secret-token" MCP_SERVER_URL=http://localhost:4444/mcp cargo run --`

- wrapper/src/json_rpc_id_fast.rs:70 -> reading id as u64 can truncate values.

crivetimihai:
> I really like the idea of running rust on the core and slowly implement new modules in rust as we need it.
> 
> I am still testing and reading the docs. But I ran docker with full rust and manually tested from the UI. It is all running as expected, including adding an external MCP provider (Claudflare) and running its tools.
> 
> Suggestions.
> 
> * cargo clippy is not passing for mpc_runtime

I tested it directly, and clippy is passing for tools_rust/mcp_runtime.

  Ran:

  - make -C tools_rust/mcp_runtime clippy
  - make -C tools_rust/mcp_runtime clippy-all


> * Not for now, but the lib.rs in mcp_runtime could be broken in modules.
Yes, this is a follow-up tracked in tools_rust/mcp_runtime/FOLLOWUPS.md


> AI catches.
> 
> * Running wrapper can show the entire config with debug on.
> 
> `LOG_LEVEL=debug MCP_AUTH="Bearer my-secret-token" MCP_SERVER_URL=http://localhost:4444/mcp cargo run --`
> 
> * wrapper/src/json_rpc_id_fast.rs:70 -> reading id as u64 can truncate values.

Thanks. I checked both points.

- The tools_rust/wrapper debug logging issue is real: with LOG_LEVEL=debug, the wrapper currently logs Config via debug!("{config:?}"), and that includes mcp_auth. That’s
  actionable, but it’s in the separate wrapper crate, not the Rust MCP runtime path in this PR.
- The json_rpc_id_fast.rs comment is directionally valid but technically overstated: it does not truncate numeric IDs, it falls back to Id::Null when the value cannot be parsed
  as u64.

So I'm not treating either as a blocker for this PR, but the wrapper secret-logging issue is worth tracking/fixing separately. @dima-zakharov please investigate.

lucarlig:
```
make -C tools_rust/mcp_runtime clippy
Running cargo clippy --all-targets...
warning: this creates an owned instance just for comparison
   --> src/config.rs:314:48
    |
314 |             ListenTarget::Uds(path) if path == PathBuf::from("/tmp/contextforge.sock")
    |                                        --------^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |                                        |
    |                                        help: try: `path == "/tmp/contextforge.sock"`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/rust-1.94.0/index.html#cmp_owned
    = note: `#[warn(clippy::cmp_owned)]` on by default

warning: `contextforge_mcp_runtime` (lib test) generated 1 warning (run `cargo clippy --fix --lib -p contextforge_mcp_runtime --tests` to apply 1 suggestion)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.35s
    ```

**Comment by crivetimihai** [comment]
> ```
> make -C tools_rust/mcp_runtime clippy
> Running cargo clippy --all-targets...
> warning: this creates an owned instance just for comparison
>    --> src/config.rs:314:48
>     |
> 314 |             ListenTarget::Uds(path) if path == PathBuf::from("/tmp/contextforge.sock")
>     |                                        --------^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>     |                                        |
>     |                                        help: try: `path == "/tmp/contextforge.sock"`
>     |
>     = help: for further information visit https://rust-lang.github.io/rust-clippy/rust-1.94.0/index.html#cmp_owned
>     = note: `#[warn(clippy::cmp_owned)]` on by default
> 
> warning: `contextforge_mcp_runtime` (lib test) generated 1 warning (run `cargo clippy --fix --lib -p contextforge_mcp_runtime --tests` to apply 1 suggestion)
>     Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.35s
>     ```
> ```

I reproduced the code state, but not the warning output on this machine.

What I verified:

- the exact pattern is still present in config.rs:
    - path == PathBuf::from("/tmp/contextforge.sock")
- so the warning report is believable

What happened locally:

- make -C tools_rust/mcp_runtime clippy
- it passed cleanly here without printing the cmp_owned warning

```
make -C tools_rust/mcp_runtime clippy
make: Entering directory '/home/cmihai/agents2/pr/mcp-context-forge/tools_rust/mcp_runtime'
Running cargo clippy --all-targets...
    Checking contextforge_mcp_runtime v0.1.0 (/home/cmihai/agents2/pr/mcp-context-forge/tools_rust/mcp_runtime)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.55s
make: Leaving directory '/tools_rust/mcp_runtime'
```

I did not get the warning locally in this run, likely due to toolchain/clippy-version differences or lint configuration differences. Either case, it is a minor warning-cleanup item, not a failing clippy error on my current environment.

gandhipratik203:
## Test Setup

- Machine: MacBook with Colima (8 CPU / 16GB RAM)
- Stack: Docker Compose — 3 gateway instances behind nginx
- Tool: Locust — 125 concurrent users, 30/s ramp-up, 60s duration
- Target: MCP tools/call via http://localhost:8080
- Upstream: fast_test_server (purpose-built MCP benchmark target)

## Steps

1. Built Docker image with Rust binary compiled in (`make testing-rebuild-rust-full`)
2. Started stack in Rust full mode (`RUST_MCP_MODE=full make testing-up`)
3. Ran benchmark (`make benchmark-mcp-tools`) — captured Rust results
4. Restarted stack in Python mode (`make testing-down && make testing-up`)
5. Ran benchmark again — captured Python results

---

## Results — 125 users / 60s

| Metric | Python | Rust Full | Improvement |
|------|------|------|------|
| Total Requests | 30,677 | 114,015 | 3.7x more |
| Failures | 24 (0.08%) | 0 (0.00%) | Rust cleaner ✅ |
| RPS | 512 | 1,930 | 3.8x faster |
| Avg Latency | 167 ms | 2.84 ms | 59x faster |
| p50 | 110 ms | 2 ms | 55x faster |
| p90 | 280 ms | 5 ms | 56x faster |
| p95 | 360 ms | 6 ms | 60x faster |
| p99 | 610 ms | 11 ms | 55x faster |
| Max | 30,001 ms | 397 ms | Rust far more stable |

---

## On failures

All 24 failures were on `tools/call` in Python mode — none in Rust.

The `30,001 ms` max is a Locust timeout (30s hard limit), meaning some requests sat in the queue long enough to be abandoned.

At **125 concurrent users** the Python path was already saturated — requests were arriving faster than workers could drain them.

Rust processes the same requests ~60x faster so the queue never builds up. Python is showing stress at 125 users; Rust has not reached its saturation point.

crivetimihai:
Thanks, this matches my read. I don’t see any plugin-framework regression or new blocker here. The only semantic change worth tracking is the PromptPosthookPayload.prompt_id UUID→name behavior, and that is already captured in tools_rust/mcp_runtime/FOLLOWUPS.md as a contract/schema follow-up.

PR #3650, opened 2026-03-12
fix(ui): set MAX_MEMBERS_PER_TEAM in team forms

### LINKED ISSUE #3589: [BUG]: `MAX_MEMBERS_PER_TEAM` setting is ignored by the UI
### 🐞 Bug Summary

Relates to #3588 

The `MAX_MEMBERS_PER_TEAM` configuration is ignored on the team creation form, being hardcoded to `1000`.

Image

---

### 🧩 Affected Component

- [ ] `mcpgateway` - API
- [X] `mcpgateway` - UI (admin panel)
- [ ] Other (explain below)

---

### 🔁 Steps to Reproduce

1. Set `MAX_MEMBERS_PER_TEAM` to a value different than 1000 on the application environment
2. Start the app
3. Login to the Admin UI as an admin user
4. Navigate to the teams tab
5. Open the `Create new team` modal.
6. Fill the form
7. Set the `Maximum members` value to a va value higher than `MAX_MEMBERS_PER_TEAM`
8. Click the Save button

If `MAX_MEMBERS_PER_TEAM > 1000`:
 - The form will not submit, and a validation message will state that the maximum team members is 1000 (less than `MAX_MEMBERS_PER_TEAM`)
else:
 - The form will submit and the team will be created  (with a user limit greater than `MAX_MEMBERS_PER_TEAM`)

### 🤔 Expected Behavior

~The form should not submit, and a validation message should state that the maximum team members is the same as the value set on `MAX_MEMBERS_PER_TEAM`.~

Following the backend behaviour, the UI should:
if the user is an admin, not set a max limit, otherwise set the environment value (that defaults to 100).

Somewhat like below:

```HTML
<div>
    <label
    class="block text-sm font-medium text-gray-700 dark:text-gray-300 mb-1"
    >
    Maximum Members
    </label>
    <input
    type="number"
    name="max_members"
    id="team-max-members"
    min="1"
    {% if !is_admin %}
    max="{{ max_members || 100 }}"
    {% endif %}
    value="50"
    class="block w-full border-gray-300 rounded-md shadow-sm focus:ring-indigo-500 focus:border-indigo-500 dark:bg-gray-700 dark:border-gray-600 dark:text-white"
    />
    <p class="mt-1 text-xs text-gray-500 dark:text-gray-400">
    Optional, defaults to 50
    </p>
</div>
```

---

The described expected behaviour will actually depend on the architectural decision taken on #3588, respecting the outcome described in that ticket.

gcgoncalves:
# 🐛 Bug-fix PR

## 📌 Summary
The "Maximum Members" field in the **Create Team** and **Edit Team** modals has `max="1000"` hardcoded in the HTML. The `MAX_MEMBERS_PER_TEAM` environment variable (default: `100`) is correctly read by the backend but is never surfaced to the admin UI, so:

- When `MAX_MEMBERS_PER_TEAM > 1000`: the form rejects values above 1000, preventing creation of teams larger than 1000 even though the backend allows it.
- When `MAX_MEMBERS_PER_TEAM < 1000`: the form allows values up to 1000, so a non-admin user can set `max_members` well above the configured limit.

Closes #3589

## 🔁 Reproduction Steps
1. Set `MAX_MEMBERS_PER_TEAM=30` in `.env`.
2. Start the app (`make dev`).
3. Log in to the Admin UI as a non-admin user.
4. Navigate to the **Teams** tab → **Create new team**.
5. Set **Maximum Members** to `500` (above the env setting but below 1000).
6. Click **Save**.

**Observed:** Team is created with `max_members=500` — the env cap of `30` is silently bypassed by the UI.

**Alternative path (`MAX_MEMBERS_PER_TEAM > 1000`):**

1. Set `MAX_MEMBERS_PER_TEAM=2000`.
2. Try to set **Maximum Members** to `1500`.
3. Browser rejects the input: "Value must be ≤ 1000".

---

## Acceptance Criteria

- [ ] For **admin users**, the "Maximum Members" input in the Create Team modal has no `max` attribute.
- [ ] For **non-admin users**, the input has `max` set to `MAX_MEMBERS_PER_TEAM`.
- [ ] The same admin/non-admin distinction applies in the **Edit Team** modal.
- [ ] The default `value` in the Create modal is `min(50, MAX_MEMBERS_PER_TEAM)`.
- [ ] Helper text accurately reflects the limit:
  - Admin: e.g. `"Optional. Admins can set any limit."`
  - Non-admin: e.g. `"Optional, defaults to {MAX_MEMBERS_PER_TEAM}."`
- [ ] Changing `MAX_MEMBERS_PER_TEAM` in the env and restarting reflects the new value without code changes.
- [ ] `min="1"` is preserved on all inputs (regression guard).

## 🐞 Root Cause
Two locations have `max="1000"` hardcoded instead of reading from `settings.max_members_per_team`:

| File | Context |
|------|---------|
| `mcpgateway/templates/admin.html` | Create Team modal — `<input type="number" name="max_members" min="1" max="1000" ...>` |
| `mcpgateway/admin.py` | Edit Team modal — Python f-string: `<input type="number" name="max_members" min="1" max="1000" ...>` |

Additionally, `max_members_per_team` is never passed to the admin template context in `admin.py` (main page render at ~line 3628), so Jinja2 cannot reference it.

## 💡 Fix Description
The "Maximum Members" input in the Create Team and Edit Team modals had max="1000" hardcoded, causing two problems:
- non-admin users could set max_members above `MAX_MEMBERS_PER_TEAM`
- when `MAX_MEMBERS_PER_TEAM` > 1000, the browser rejected valid values

Fix by surfacing the setting to the UI with role-aware behaviour:
- admins: no max attribute (no browser-side cap, matching `max_members` in the backend)
- non-admins: max attribute set to `MAX_MEMBERS_PER_TEAM` (default 100)

The default value in the Create modal is min(50, `MAX_MEMBERS_PER_TEAM`)
so tighter operator-configured limits are respected out of the box.

## 🧪 Verification

| Lint suite                            | `make lint`          |        |
| Unit tests                            | `make test`          |        |
| Coverage ≥ 80 %                       | `make coverage`      |        |

## ✅ Checklist

### DIFF
--- a/mcpgateway/utils/passthrough_headers.py
+++ b/mcpgateway/utils/passthrough_headers.py
@@ -568,6 +568,9 @@ def _loopback_skip_set() -> frozenset[str]:
     passthrough headers can never overwrite the gateway-internal proxy
     user identity — even if that header name is added to the passthrough
     allowlist by mistake.
+
+    Returns:
+        frozenset[str]: Header names to skip during loopback forwarding.
     """
     proxy = settings.proxy_user_header.lower()
     if proxy in _LOOPBACK_SKIP_HEADERS:
@@ -787,6 +790,13 @@ def safe_extract_headers_for_loopback(request_headers: Dict[str, str], transport
     Wraps :func:`extract_headers_for_loopback` so that SSE / WebSocket setup
     is never blocked by passthrough configuration issues.  ``ImportError``
     propagates (broken deployment should fail loudly).
+
+    Args:
+        request_headers: Incoming HTTP headers to extract from.
+        transport_name: Label for warning logs on failure.
+
+    Returns:
+        Dict[str, str]: Extracted passthrough headers, or empty dict on error.
     """
     try:
         return extract_headers_for_loopback(request_headers)
@@ -801,6 +811,12 @@ def safe_extract_and_filter_for_loopback(request_headers: Dict[str, str]) -> Dic
     Combines :func:`extract_headers_for_loopback` and
     :func:`filter_loopback_skip_headers` with error handling so that
     Streamable HTTP affinity loopback calls degrade gracefully.
+
+    Args:
+        request_headers: Incoming HTTP headers to extract and filter.
+
+    Returns:
+        Dict[str, str]: Filtered passthrough headers, or empty dict on error.
     """
     try:
         return filter_loopback_skip_headers(extract_headers_for_loopback(request_headers))


gcgoncalves:
> High: The current change only fixes the browser-side constraint, not the server-side enforcement. Non-admin users can still bypass the new max attribute by submitting a crafted POST request with a larger max_members value, and the backend
will accept it. That means the original issue is still present from a security/policy perspective.
Suggested solution: add backend validation in the create/update team handlers or service layer to enforce settings.max_members_per_team for non-admin users, and only allow values above that limit for admins if that is the intended behavior.
This should be enforced server-side regardless of what the UI renders.

Good catch,  updating.

> Low: The PR description mentions that the admin UI behavior matches the backend via skip_limits, but in the current implementation that backend flag appears to relate to team-count limits rather than max_members limits. That makes the
current explanation a bit misleading.
Suggested solution: update the implementation or the PR description so they are aligned. If admins are meant to bypass MAX_MEMBERS_PER_TEAM, that logic should exist explicitly in the backend; otherwise the description should avoid implying
that it already does.

That's true, updated the PR description.

crivetimihai:
Thanks @gcgoncalves — fixes #3589. Targeting 1.1.0.

marekdano:
@gcgoncalves - the PR looks good. Please resolve conflicts and lint and pytest issues

vishu-bh (dismissed):
LGTM 🚀

PR #3659, opened 2026-03-13
fix: use compute_passthrough_headers_cached in direct proxy paths to …

### LINKED ISSUE #3643: [BUG]: Direct proxy paths in Streamable HTTP skip X-Upstream-Authorization rename
### 🐞 Bug Summary
The direct proxy functions in the Streamable HTTP transport (`_proxy_list_tools_to_gateway`, `_proxy_list_resources_to_gateway`, and resource read) do manual header forwarding that only copies headers explicitly listed in `gateway.passthrough_headers`. They bypass `compute_passthrough_headers_cached()` entirely, so the always-on `X-Upstream-Authorization` → `Authorization` rename does not happen, and global passthrough header config is ignored.

---

### 🧩 Affected Component

- [x] `mcpgateway` - API
- [ ] Other (explain below)

---

### 🔁 Steps to Reproduce

1. Configure a gateway in `direct_proxy` mode with an upstream server that requires auth
2. Connect via Streamable HTTP (`/servers/{id}/mcp`)
3. Send a `tools/list`, `resources/list`, or `resources/read` request with `X-Upstream-Authorization: Bearer <token>`
4. Observe: the upstream server does **not** receive `Authorization: Bearer <token>` — the header is silently dropped

**Note:** `tools/call` on the same transport works correctly because it goes through `tool_service.invoke_tool()` which calls `compute_passthrough_headers_cached()`.

---

### 🤔 Expected Behavior
`X-Upstream-Authorization` should be renamed to `Authorization` and forwarded to upstream servers on all direct proxy operations (list tools, list resources, read resource, list prompts), consistent with the tool invocation path.

---

### 📓 Logs / Error Output
No errors. Headers are silently not forwarded.

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Version or commit | `main branch` |
| Runtime | `Python 3.11+` |
| Platform / OS | `Any` |
| Container | `Any` |

---

### 🧩 Root Cause Analysis

Three functions in `mcpgateway/transports/streamablehttp_transport.py` do their own manual header loop instead of using the shared passthrough utility:

**`_proxy_list_tools_to_gateway()`** (line 778):
```python
if gateway.passthrough_headers and request_headers:
    for header_name in gateway.passthrough_headers:
        header_value = request_headers.get(header_name.lower()) or request_headers.get(header_name)
        if header_value:
            headers[header_name] = header_value
```

Same pattern at:
- **`_proxy_list_resources_to_gateway()`** (line 821)
- **Resource read proxy** (line 878)

This manual loop:
1. Does NOT handle `X-Upstream-Authorization` → `Authorization` rename
2. Does NOT consult global passthrough config (`DEFAULT_PASSTHROUGH_HEADERS`)
3. Only forwards headers if the gateway has `passthrough_headers` explicitly configured
4. Does NOT call `compute_passthrough_headers_cached()` or `get_passthrough_headers()`

---

### 🔧 Proposed Fix

Replace the manual header loop with a call to `compute_passthrough_headers_cached()` in each function, or at minimum add explicit `X-Upstream-Authorization` handling alongside the existing loop.

omorros:
## 🔗 Related Issue                                                                                                   
  Closes #3643
                                                                                                                        
  ## 📝 Summary
  Replaced manual header forwarding loops in the three direct proxy functions (`_proxy_list_tools_to_gateway`,
  `_proxy_list_resources_to_gateway`, `_proxy_read_resource_to_gateway`) with calls to                                    `compute_passthrough_headers_cached()`, fixing the silently dropped `X-Upstream-Authorization → Authorization` rename
  and restoring global passthrough header config support on all direct proxy operations.                                
  
  ---

  ## 🏷️ Type of Change
  - [x] Bug fix
  - [ ] Feature / Enhancement
  - [ ] Documentation
  - [ ] Refactor
  - [ ] Other (describe below)
                                                                                                                      
                                                                                                                        
  ## 🧪 Verification

  ---                                                                                                                   
  
  ## ✅ Checklist
                                                                                                                        
  ---

  ## 📓 Notes (optional)
  The `db` parameter was added as optional (`None` default) to the three proxy functions to avoid breaking any other
  callers. All three call sites already operate inside `async with get_db() as db:` blocks, so passing `db` required no   structural changes. 350 transport tests and 75 passthrough header utility tests pass.

### DIFF
--- a/mcpgateway/transports/streamablehttp_transport.py
+++ b/mcpgateway/transports/streamablehttp_transport.py
@@ -71,8 +71,10 @@ from mcpgateway.services.prompt_service import PromptService
 from mcpgateway.services.resource_service import ResourceService
 from mcpgateway.services.tool_service import ToolService
 from mcpgateway.transports.redis_event_store import RedisEventStore
+from mcpgateway.cache.global_config_cache import global_config_cache
 from mcpgateway.utils.gateway_access import build_gateway_auth_headers, check_gateway_access, extract_gateway_id_from_headers, GATEWAY_ID_HEADER
 from mcpgateway.utils.orjson_response import ORJSONResponse
+from mcpgateway.utils.passthrough_headers import compute_passthrough_headers_cached
 from mcpgateway.utils.verify_credentials import is_proxy_auth_trust_active, require_auth_header_first, verify_credentials
 
 # Initialize logging service first
@@ -758,7 +760,7 @@ async def _validate_streamable_session_access(
     return False, HTTP_403_FORBIDDEN, "Session owner metadata unavailable"
 
 
-async def _proxy_list_tools_to_gateway(gateway: Any, request_headers: dict, user_context: dict, meta: Optional[Any] = None) -> List[types.Tool]:  # pylint: disable=unused-argument
+async def _proxy_list_tools_to_gateway(gateway: Any, request_headers: dict, user_context: dict, meta: Optional[Any] = None, db: Optional[Session] = None) -> List[types.Tool]:  # pylint: disable=unused-argument
     """Proxy tools/list request directly to remote MCP gateway using MCP SDK.
 
     Args:
@@ -766,6 +768,7 @@ async def _proxy_list_tools_to_gateway(gateway: Any, request_headers: dict, user
         request_headers: Request headers from client
         user_context: User context (not used - _meta comes from MCP SDK)
         meta: Request metadata (_meta) from the original request
+        db: SQLAlchemy database session for passthrough header config lookup
 
     Returns:
         List of Tool objects from remote server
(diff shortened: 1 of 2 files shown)


marekdano (commented):
@omorros - thanks for the PR!

### Overall Assessment: Positive — fixes a real security gap

The PR correctly replaces ad-hoc manual header forwarding with the shared `compute_passthrough_headers_cached()` utility, which provides proper sanitization and the `X-Upstream-Authorization → Authorization` rename. This is a clear improvement.

This is a security improvement:

The new code passes `gateway_auth_type` which enables the check that blocks client `Authorization` passthrough when the gateway itself uses basic or bearer auth. The old code had no such protection — a client could potentially override gateway auth credentials via passthrough headers if Authorization was in `gateway.passthrough_headers`.

### Security improvements

Silent fallback when `db` is `None` — headers silently dropped (Medium)

All three functions use this guard:
```bash
  if request_headers and db is not None:
      # ... compute passthrough headers
```

If `db` is None, no passthrough headers are forwarded at all — including `X-Upstream-Authorization`. The old code at least forwarded explicitly configured `gateway.passthrough_headers` without needing a DB session.

While all three current call sites pass `db=db`, the parameter is `Optional[None]` by design "to avoid breaking other callers." If a future caller omits db, header forwarding silently degrades. Consider either:
  - Logging a warning when `db` is `None` and headers would have been forwarded
  - Or making `db` required since all current callers have it available

### Test Coverage Assessment

The tests mock `compute_passthrough_headers_cached` indirectly (by mocking `global_config_cache` and settings), but they don't verify that `X-Upstream-Authorization` actually gets renamed. Consider adding a test that:
  - Sets `X-Upstream-Authorization: Bearer token` in request headers
  - Uses a real (non-mocked) `compute_passthrough_headers_cached`
  - Asserts the outbound headers contain `Authorization: Bearer token`


Minor suggestion to add a warning log when `db` is `None` causes passthrough headers to be skipped, and consider one test using the real `compute_passthrough_headers_cached` to verify the rename end-to-end.

crivetimihai:
Thanks @omorros — note that #3677 also addresses the same issue (#3643) using the shared passthrough header utility. Please coordinate to avoid conflicting changes.

jonpspri:
Closing as superseded. The linked issue #3643 ([BUG]: Direct proxy paths in Streamable HTTP skip X-Upstream-Authorization rename) was closed as COMPLETED — the fix has already been merged. This PR is no longer needed.

PR #3663, opened 2026-03-13
feat: Add PLUGINS_CAN_OVERRIDE_AUTH_HEADERS feature flag for WXO auth

jonpspri:
## Summary
- Adds `PLUGINS_CAN_OVERRIDE_AUTH_HEADERS` setting (default `false`) to control whether pre-request plugin hooks can override auth-sensitive headers (`Authorization`, `Cookie`, `X-Api-Key`, `Proxy-Authorization`)
- Parameterizes `JWT_SECRET_KEY` across `docker-compose.yml` via env var substitution instead of hardcoded values
- Adds multi-tenant load test support with per-user JWT identity and `woTenantId` claims

## Test plan
- [x] Unit tests: deny-path (override blocked by default) and allow-path (override permitted when enabled)
- [x] All 27 tests in `test_http_auth_headers.py` pass
- [x] Black, isort, Bandit clean on changed files

### DIFF
--- a/tests/loadtest/locustfile_echo_delay.py
+++ b/tests/loadtest/locustfile_echo_delay.py
@@ -138,6 +138,7 @@ def _generate_jwt_token(user_email: str, tenant_id: str) -> str | None:
         "aud": "mcpgateway-api",
         "sub": user_email,
         "exp": now + 86400,
+        "jti": uuid.uuid4().hex,
         "woTenantId": tenant_id,
     }
     try:


crivetimihai:
Targeting 1.0.0 given MUST priority for WXO auth compatibility.

crivetimihai:
**Follow-up fix pushed**: `b6ed12483` — Added missing `jti` claim to `_generate_jwt_token()` in the locustfile.

The auto-generation fallback (used when `WXO_AUTH_ENABLED=false` and no `MCPGATEWAY_BEARER_TOKEN` is set) was producing tokens without a `jti` claim, which would be rejected under the default `REQUIRE_JTI=true` setting. Now generates a UUID-based `jti`, matching what the `create_jwt_token` CLI produces.

PR #3685, opened 2026-03-14
fix(schemas): accept camelCase fields on gateway creation

### LINKED ISSUE #3577: [BUG]: Gateway creation ignores passthroughHeaders (camelCase) field
### 🐞 Bug Summary
When creating a gateway via `POST /gateways`, the `passthroughHeaders` field (camelCase) is silently ignored, but the same field works correctly when updating via `PUT /gateways/{id}`.

### 🧩 Affected Component
- [x] `mcpgateway` - API

### 🔁 Steps to Reproduce
1. POST to `/gateways` with `{"passthroughHeaders": ["billingId", "userName", "userEmail"]}`
2. Verify field is NULL in database
3. PUT to `/gateways/{id}` with same payload
4. Verify field is now set correctly

### 🤔 Expected Behavior
Both CREATE and UPDATE should accept `passthroughHeaders` (camelCase) consistently.

### 🧠 Environment Info
| Key | Value |
|-----|-------|
| Version | main |
| Runtime | Python 3.11+ |

### 🧩 Additional Context
**Root Cause**: `GatewayCreate` extends `BaseModel` without camelCase alias generation, while `GatewayUpdate` extends `BaseModelWithConfigDict` which has this feature.

crivetimihai:
## Summary
- `GatewayCreate` extended `BaseModel` directly, missing the `alias_generator=to_camel_case` and `populate_by_name=True` config that `GatewayUpdate` inherits from `BaseModelWithConfigDict`
- This caused camelCase field names (e.g. `passthroughHeaders`, `authType`, `authToken`) to be silently ignored on `POST /gateways`, while the same fields worked correctly on `PUT /gateways/{id}`
- Fix: change `GatewayCreate` to extend `BaseModelWithConfigDict`, matching `GatewayUpdate` and `AdminGatewayCreate`

Closes #3577

### DIFF
--- a/tests/unit/mcpgateway/test_schemas.py
+++ b/tests/unit/mcpgateway/test_schemas.py
@@ -1254,3 +1254,92 @@ class TestSchemaValidators:
             ResourceSubscription(uri="resource://one", subscriber_id=long_subscriber)
 
         assert "Subscriber ID exceeds maximum length" in str(exc_info.value)
+
+
+class TestGatewayCreateCamelCase:
+    """Verify GatewayCreate accepts camelCase input and serializes to camelCase output.
+
+    Regression tests for #3577 — GatewayCreate previously extended BaseModel
+    directly, missing alias_generator and populate_by_name from
+    BaseModelWithConfigDict, so camelCase fields were silently ignored.
+    """
+
+    def test_accepts_camel_case_fields(self):
+        """GatewayCreate must accept camelCase field names."""
+        from mcpgateway.schemas import GatewayCreate
+
+        gw = GatewayCreate(
+            name="test-gw",
+            url="https://example.com",
+            authType="bearer",
+            authToken="tok123",
+            passthroughHeaders=["X-Request-Id"],
+            gatewayMode="direct_proxy",
+            refreshIntervalSeconds=120,
+            oneTimeAuth=True,
+            caCertificate="PEM-DATA",
+        )
+        assert gw.auth_type == "bearer"
+        assert gw.auth_token == "tok123"
+        assert gw.passthrough_headers == ["X-Request-Id"]
+        assert gw.gateway_mode == "direct_proxy"
+        assert gw.refresh_interval_seconds == 120
+        assert gw.one_time_auth is True
+        assert gw.ca_certificate == "PEM-DATA"
+
+    def test_accepts_snake_case_fields(self):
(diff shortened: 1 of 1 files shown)


crivetimihai:
## Review

**Verdict: Approved** — the fix is correct, well-scoped, and introduces no breakage.

### What the fix does

`GatewayCreate` extended `BaseModel` directly, missing `alias_generator=to_camel_case` and `populate_by_name=True` that `BaseModelWithConfigDict` provides. This caused `POST /gateways` to silently ignore camelCase fields (`authType`, `passthroughHeaders`, etc.), while `PUT /gateways/{id}` (using `GatewayUpdate(BaseModelWithConfigDict)`) worked fine.

The one-line fix changes the base class to `BaseModelWithConfigDict`, matching all other Gateway schemas (`GatewayUpdate`, `GatewayRead`, `AdminGatewayCreate`, etc.).

### Breakage analysis

- All internal callers (`admin.py`, `catalog_service.py`, `import_service.py`) construct `GatewayCreate` with snake_case kwargs — unaffected by the alias generator since `populate_by_name=True` preserves snake_case access.
- All attribute access in `gateway_service.py` uses snake_case (`gateway.name`, `gateway.auth_type`, etc.) — Python attribute names, not aliases.
- No code calls `model_dump(by_alias=True)` on a `GatewayCreate` instance.
- Pydantic v2 correctly merges the child's `ConfigDict(str_strip_whitespace=True)` with the parent config — verified empirically.
- `extra="ignore"` matches the Pydantic v2 default; `use_enum_values=True` has no effect since `GatewayCreate` has no enum-typed fields.

### Changes made during review

- Rebased onto `main` (removed merge commit, now 2 linear commits)
- Added 4 differential tests in `TestGatewayCreateCamelCase`:
  1. `test_accepts_camel_case_fields` — verifies camelCase input
  2. `test_accepts_snake_case_fields` — regression guard
  3. `test_serializes_to_camel_case` — verifies `model_dump(by_alias=True)` output
  4. `test_consistent_with_gateway_update` — ensures parity with `GatewayUpdate`

### Note

Six other `*Create` schemas (`ToolCreate`, `ResourceCreate`, `ServerCreate`, `A2AAgentCreate`, `GrpcServiceCreate`, `PerformanceSnapshotCreate`) extend `BaseModel` directly — same pattern as this bug. Worth a follow-up issue if camelCase input is expected on those endpoints.

PR #3687, opened 2026-03-14
fix(rbac): use check_any_team for API tokens in MCP transports

### LINKED ISSUE #3660: [BUG][RBAC]: Non-admin user gets Access denied when invoking MCP tools
When test mcp tool usage occurs:
```json
{
  "jsonrpc": "2.0",
  "error": {
    "code": -32003,
    "message": "Access denied",
    "data": {
      "method": "ssh-session-close-all-sessions"
    }
  },
  "id": 1773371908150
}
```

When use api token access mcp occur `McpError: MCP error -32001: Streamable HTTP error: Error POSTing to endpoint: {"detail":"Access denied"}`

Expected behavior:
`Non-Administrator` user should allow access mcp server but cannot edit mcp servers

Another small bug：
It is visible for User B's API Token when User `A` in API Token Management page

crivetimihai:
## Summary

- Non-admin users with team-scoped roles (`developer`, `team_admin`) were denied MCP tool access because the RBAC check only set `check_any_team=True` for session tokens, leaving API tokens with `check_any_team=False`
- Since MCP transports have no route-level `team_id`, the permission service only found null-scoped team roles, missing team-specific role assignments that grant `tools.execute` and `servers.use`
- Additionally, the `logging/setLevel` handler was missing `check_any_team` entirely — even session tokens could fail there

### Changes

| File | Fix |
|------|-----|
| `streamablehttp_transport.py` | Always use `check_any_team=True` in `call_tool`, `logging/setLevel`, and `handle_streamable_http` RBAC checks (MCP transport never has a team_id) |
| `main.py` | `/rpc` endpoint: extract `team_id` from user dict; fall back to `check_any_team=True` when absent |
| `rbac.py` | `@require_permission` decorator: extend team derivation to API tokens, not just session tokens |
| Tests (3 files) | Updated existing assertions and added new test for API token with `team_id` |

Layer 1 (token scope cap) already restricts what a token can access by team. Layer 2 (RBAC) just needs to verify the user has the capability — so `check_any_team=True` is safe when no route-level `team_id` is available.

Closes #3660

## Test plan

- [x] All 1252 transport and middleware unit tests pass
- [x] All 13 RBAC permission fallback regression tests pass
- [x] All 15 permission service fallback tests pass

### DIFF
--- a/mcpgateway/main.py
+++ b/mcpgateway/main.py
@@ -1013,10 +1013,11 @@ async def _ensure_rpc_permission(user, db: Session, permission: str, method: str
         return
 
     # Layer 2: RBAC check
-    # The /rpc endpoint has no route-level team_id.  For single-team API tokens
-    # we can derive team_id from the token; otherwise fall back to check_any_team
-    # so that team-scoped roles (developer, team_admin) are found.
-    # Layer 1 (token scope cap above) already restricts what the token can do.
+    # /rpc payloads never carry a resource with an owning team, so we skip
+    # resource/payload derivation (unlike @require_permission).  For single-
+    # team API tokens we extract team_id from the token itself; otherwise
+    # fall back to check_any_team so team-scoped roles are found.
+    # Layer 1 (token scope cap above) already restricts visibility.
     team_id: str | None = None
     check_any_team = False
     if isinstance(user, dict):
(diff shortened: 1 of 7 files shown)


gcgoncalves (dismissed):

madhu-mohan-jaishankar (dismissed):
Looks good to me. LGTM.

PR #3696, opened 2026-03-16
fix(api): consolidate loopback RPC URLs and TLS verification into shared helper (#3543)

### LINKED ISSUE #3543: [CHORE][PYTHON]: Consolidate internal loopback RPC calls into a shared helper
### 🔧 Chore Summary

Consolidate the 5 internal loopback RPC call sites into a shared helper that correctly derives the internal base URL from launcher configuration, handling TLS and non-standard ports.

Currently, 5 call sites independently construct `http://127.0.0.1:{settings.port}/rpc` for internal self-calls. This works for the default compose topology (plain HTTP behind nginx) but breaks under `make serve-ssl` / `make serve-granian-ssl` where gunicorn/granian bind HTTPS-only on the same port.

Discovered during review of PR #3339 (fix(sse): use loopback for internal RPC calls).

---

### 🧱 Area Affected

- [x] Other: Internal HTTP client architecture (`session_registry`, `mcp_session_pool`, `streamablehttp_transport`, WebSocket relay)

---

### ⚙️ Context / Rationale

**Current state — 5 independent loopback call sites:**

| # | File | Line | Pattern |
|---|------|------|---------|
| 1 | `mcpgateway/cache/session_registry.py` | 2307 | `f"http://127.0.0.1:{settings.port}/rpc"` |
| 2 | `mcpgateway/services/mcp_session_pool.py` | 1559 | `f"http://127.0.0.1:{settings.port}/rpc"` |
| 3 | `mcpgateway/services/mcp_session_pool.py` | 1624 | `f"http://127.0.0.1:{settings.port}{path}"` |
| 4 | `mcpgateway/transports/streamablehttp_transport.py` | 2450 | `f"http://127.0.0.1:{settings.port}/rpc"` |
| 5 | `mcpgateway/transports/streamablehttp_transport.py` | 2611 | `f"http://127.0.0.1:{settings.port}/rpc"` |

Additionally, `mcpgateway/main.py:6843` (WebSocket relay) uses `f"http://localhost:{settings.port}{settings.app_root_path}/rpc"` — a slightly different pattern that also doesn't handle TLS.

**Problems:**

1. **TLS mismatch**: `make serve-ssl` binds gunicorn with `--certfile`/`--keyfile`, making the port HTTPS-only. Plain `http://` loopback fails with SSL errors. Same for `make serve-granian-ssl`.
2. **Inconsistency**: 5 sites duplicate the same URL construction, with the WebSocket relay being a slight variant (uses `localhost` + `app_root_path` instead of `127.0.0.1` + bare `/rpc`).
3. **No single source of truth**: The internal URL is reconstructed ad-hoc at each call site rather than configured centrally.

**Note**: The `make dev` port mismatch (uvicorn on 8000, `settings.port` defaulting to 4444) is theoretically present but not practically reachable for SSE — the memory backend `respond()` is single-shot and never reaches `generate_response()`.

---

### 📋 Acceptance Criteria

- [ ] Shared helper (e.g., `get_internal_rpc_url()` or `InternalHttpClient`) used by all 5+ call sites
- [ ] Helper derives scheme (`http` vs `https`) from launcher config (detect `SSL=true` / cert presence)
- [ ] Helper uses `settings.port` as the port (already correct for gunicorn/granian/compose)
- [ ] Header sanitization: internal calls must NOT forward `Host`, `X-Forwarded-For`, `X-Forwarded-Proto`, or `X-Real-IP` (risk: `ProxyHeadersMiddleware` with `trusted_hosts="*"` at `main.py:2131` can rewrite `request.client.host` from forwarded headers)
- [ ] WebSocket relay site (`main.py:6843`) migrated to the shared helper
- [ ] Existing unit tests updated; no regressions
- [ ] `make serve-ssl` + SSE/streamable-HTTP verified working

**Stretch:**

- [ ] Consider `httpx.ASGITransport` for in-process calls (avoids network entirely)
- [ ] Validate forwarded path in `mcp_session_pool.py:1624` against expected MCP routes (ref: `todo/mcpgateway/medium/C-106-forwarded-path-injection-ssrf.md`)

---

### 🧩 Additional Context

- PR #3339 fixed the primary production issue (SSE behind nginx reverse proxy) by aligning `session_registry.py` with the existing loopback pattern
- The TLS concern applies equally to all 5 pre-existing sites, not just the one added by #3339
- In production (docker-compose), TLS termination happens at nginx, so the gateway runs plain HTTP — the loopback pattern works correctly there
- Closes the follow-up concern from PR #3339 review

madhu-mohan-jaishankar:
## 🔗 Related Issue
Closes #3543

---

## 📝 Summary
This PR removes hardcoded loopback HTTP URLs and TLS verification flags and centralizes the decision based on runtime mode (SSL env) into a tiny helper.

Concretely:

Adds mcpgateway/utils/internal_http.py with:
internal_loopback_base_url() — returns "[http://127.0.0.1:{port}](http://127.0.0.1:%7Bport%7D/)" or "[https://127.0.0.1:{port}](https://127.0.0.1:%7Bport%7D/)" depending on SSL env.
internal_loopback_verify() — returns a boolean to use for HTTP client TLS verification (disabled for local HTTPS).
Replaces literal loopback URLs:
"[http://127.0.0.1:{settings.port}/rpc](http://127.0.0.1:%7Bsettings.port%7D/rpc)"
"[http://localhost:{settings.port}{settings.app_root_path}/rpc](http://localhost:%7Bsettings.port%7D%7Bsettings.app_root_path%7D/rpc)" with f"{internal_loopback_base_url()}/..." where used.
Replaces default httpx/ResilientHttpClient verification use with internal_loopback_verify() so local HTTPS with self-signed certs won't fail.
Updates one assertion in tests to assert against internal_loopback_base_url().
Adds unit tests for the new internal_http helpers.
Files changed:

Added: mcpgateway/utils/internal_http.py
Modified: mcpgateway/cache/session_registry.py
Modified: mcpgateway/main.py
Modified: mcpgateway/services/mcp_session_pool.py
Modified: mcpgateway/transports/streamablehttp_transport.py
Test changes: tests/unit/mcpgateway/cache/test_session_registry_coverage.py (assertion updated)
Added tests: tests/unit/mcpgateway/utils/test_internal_http.py
This is an internal refactor to remove hardcoded loopback scheme and verification; it does not change external APIs or RPC semantics.

---

## 🏷️ Type of Change
- [ ] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [x] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

---

## ✅ Checklist

### DIFF
--- a/mcpgateway/utils/internal_http.py
+++ b/mcpgateway/utils/internal_http.py
@@ -23,7 +23,7 @@ def _is_ssl_enabled() -> bool:
     Returns:
         bool: ``True`` when ``SSL=true`` is set in the environment.
     """
-    return os.getenv("SSL", "false").lower() == "true"
+    return os.getenv("SSL", "false") == "true"
 
 
 def internal_loopback_base_url() -> str:
--- a/tests/unit/mcpgateway/utils/test_internal_http.py
+++ b/tests/unit/mcpgateway/utils/test_internal_http.py
@@ -32,13 +32,15 @@ class TestIsSSLEnabled:
         monkeypatch.setenv("SSL", "")
         assert _is_ssl_enabled() is False
 
-    def test_ssl_uppercase(self, monkeypatch):
+    def test_ssl_uppercase_not_truthy(self, monkeypatch):
+        """Shell launchers use exact [[ "${SSL}" == "true" ]], so uppercase is not truthy."""
         monkeypatch.setenv("SSL", "TRUE")
-        assert _is_ssl_enabled() is True
+        assert _is_ssl_enabled() is False
 
-    def test_ssl_mixed_case(self, monkeypatch):
+    def test_ssl_mixed_case_not_truthy(self, monkeypatch):
+        """Only exact lowercase 'true' enables SSL, matching run-gunicorn.sh / run-granian.sh."""
         monkeypatch.setenv("SSL", "True")
-        assert _is_ssl_enabled() is True
+        assert _is_ssl_enabled() is False
 
     def test_ssl_with_whitespace_not_truthy(self, monkeypatch):
         """Whitespace-padded values are not truthy, matching gunicorn.config.py and shell launchers."""


crivetimihai:
Thanks @madhu-mohan-jaishankar. Good refactor — centralizing the loopback URL and TLS verification logic prevents the SSL-mode breakage described in #3543. Unit tests for the helpers are a nice touch. LGTM.

mcpgateway/utils/internal_http.py
ja8zyjits, line 1: Please add appropriate license docstring at the top of the file. You can copy from other files.

mcpgateway/utils/internal_http.py
```diff
@@ -0,0 +1,30 @@
...
+
+These helpers centralize protocol and TLS verification behavior for
+self-calls to local endpoints like /rpc.
+"""
+
+import os
+from mcpgateway.config import settings
+
+
+_SSL_TRUE_VALUES = {"true"}
```
ja8zyjits: Why is this a separate Global variable? any specific reason for that?

madhu-mohan-jaishankar: "_SSL_TRUE_VALUES is not a global variable - it is a private module-level constant (note the _ prefix). It is defined at module scope intentionally so it is created once at import time rather than recreated on every function call.

mcpgateway/utils/internal_http.py
madhu-mohan-jaishankar: good catch. added now.

mcpgateway/utils/internal_http.py
ja8zyjits: HI @madhu-mohan-jaishankar , 
- I dont see any value in having that defined outside. Specially when there is no reuse of the values anywhere else. In this case, it is just a variable with a default value with limited scope.
- Also we are not fetching it from any external configuration or environment.
- Might as well define it in the only function calling it `internal_loopback_base_url`

madhu-mohan-jaishankar: Just to clarify - _SSL_TRUE_VALUES is not a hardcoded URL string or a simple default value. It's a module-level constant set used for membership testing (ssl_env in _SSL_TRUE_VALUES).

Keeping it at module level is intentional for a few reasons:
PEP 8 convention - constants belong at module level, not inside functions.
Built once - the set is constructed once at import time, not recreated on every function call.
Extensible - if we ever need to accept "1" or "yes" as valid SSL truthy values, we just update the set in one place.
The _ prefix already signals it's private to this module.

I believe the current placement is idiomatic Python.

crivetimihai:
## Maintainer Review

Rebased onto main and reviewed. Overall a sound refactoring — centralizing the scattered loopback URL construction and TLS verify flags is a good improvement. Changes made during review:

### Changes

1. **Extracted `_is_ssl_enabled()` private helper** — both `internal_loopback_base_url()` and `internal_loopback_verify()` now share a single SSL check instead of `internal_loopback_verify()` reconstructing the full URL just to inspect the scheme.

2. **Removed `_SSL_TRUE_VALUES` set** — replaced with direct `== "true"` comparison, which is simpler and matches the existing pattern.

3. **Dropped `.strip().lower()` → exact `== "true"`** — The shell launchers (`run-gunicorn.sh:305`, `run-granian.sh:299`) use case-sensitive `[[ "${SSL}" == "true" ]]`. Using `.lower()` would treat `SSL=TRUE` or `SSL=True` as enabled, but the server would actually be listening on plain HTTP, breaking loopback self-calls. The helper now matches the launcher semantics exactly.

4. **Added license header** to test file.

5. **Expanded test coverage from 2 → 16 tests** (100% coverage on the new module): SSL unset, empty string, whitespace-padded, uppercase, mixed case, `"1"`, `"yes"`, port variations, and all three public functions.

6. **Added DAR201 docstring fix** for flake8 compliance.

Issue #3525, opened 2026-03-07
[BUG]: GatewayUpdate.visibility accepts arbitrary strings — no enum validation


crivetimihai:
### Description

The `visibility` field on `GatewayUpdate` in `mcpgateway/schemas.py` is typed as `Optional[str]` with no enum constraint. Any string value (e.g. `"invalid_value"`, `""`) is accepted and persisted to the database.

### Reproduction

```bash
curl -X PUT /gateways/{id} \
  -H "Authorization: Bearer $TOKEN" \
  -H "Content-Type: application/json" \
  -d '{"visibility": "invalid_value"}'
# Returns 200 — gateway.visibility is now "invalid_value"
```

### Expected Behavior

The schema should validate that `visibility` is one of the allowed values (`public`, `team`, `private`) and reject others with 422.

### Affected Code

```python
# mcpgateway/schemas.py — GatewayUpdate
visibility: Optional[str] = None  # No enum/Literal constraint
```

### Suggested Fix

Use `Optional[Literal["public", "team", "private"]]` or a Pydantic validator to reject invalid values.

### Impact

- Invalid visibility values can be persisted, which may cause unpredictable behavior in token scoping (Layer 1) where visibility is used for filtering
- The same issue likely affects `GatewayCreate` and other schemas with visibility fields (tools, resources, prompts)

### Discovered During

E2E testing of PR #3476

crivetimihai:
**Triage: Confirmed.** `GatewayUpdate.visibility` in `schemas.py` (line 2890) is still `Optional[str]` — accepts any arbitrary string. Fix: constrain to `Literal["private", "team", "public"]` or a `VisibilityEnum`.

PR #3701, opened 2026-03-17
fix(schemas): enforce visibility literals across entity schemas

shoummu1:
## :pushpin: Summary
This PR tightens visibility validation at the schema boundary so only supported enum values are accepted.

Previously, several schemas used plain string typing for `visibility`, which allowed arbitrary values such as `invalid_value`, empty strings, or wrong casing to pass validation. This could lead to inconsistent behavior in token-scoping logic that branches on visibility values.

The issue was reported for `GatewayUpdate`, but the same pattern existed across related entity schemas. This PR applies a consistent enum-based constraint across those schemas.

Also included: a follow-up consistency fix to align Team response typing with Team request enums.

## :link: Related Issue
Closes: #3525

```bash
curl -X PUT /gateways/{id} \
  -d '{"visibility": "invalid_value"}'
# Before: accepted
# After: 422 validation error
```

## 🐞 Root Cause
Visibility fields in multiple schemas were typed as unconstrained strings (`str` / `Optional[str]`) while allowed values were only documented in prose. Without a type-level constraint, Pydantic accepted invalid values.

## :bulb: Fix Description
- Replaced unconstrained visibility string types with `Literal` enums for affected schemas so invalid values are rejected during request/response validation.
- Removed redundant visibility validators that became dead code after the `Literal` migration:
  - `ServerCreate.validate_visibility`
  - `A2AAgentCreate.validate_visibility`
  - `A2AAgentUpdate.validate_visibility`
- Updated tests that previously called removed validator methods directly to use schema instantiation assertions.
- Added broad literal-enum validation coverage (`test_visibility_literal_enum_validation`) for valid and invalid visibility inputs across impacted schemas.
- Follow-up consistency update: `TeamResponse.visibility` now uses Team enum typing (`"private" | "public"`) to match Team create/update schemas.

## :test_tube: Verification
| Check | Command | Status |
|---|---|---|
| Targeted visibility/schema unit tests | `python3 -m pytest tests/unit/mcpgateway/test_schemas_validators_extra.py -q` | :white_check_mark: |
| Schema-focused unit subset | `python3 -m pytest tests/unit/mcpgateway/ -k schema` | :white_check_mark: (379 passed, 6 skipped) |

### DIFF
--- a/mcpgateway/schemas.py
+++ b/mcpgateway/schemas.py
@@ -56,6 +56,28 @@ logger = logging.getLogger(__name__)
 _HOSTNAME_RE: Pattern[str] = re.compile(r"^(https?://)?([a-zA-Z0-9.-]+)(:[0-9]+)?$")
 _SLUG_RE: Pattern[str] = re.compile(r"^[a-z0-9-]+$")
 
+_VALID_VISIBILITY = {"private", "team", "public"}
+
+
+def _coerce_visibility(v: Optional[str]) -> Optional[str]:
+    """Normalize legacy visibility values in Read/response schemas.
+
+    DB columns are unconstrained strings, so historical rows may contain
+    values outside the Literal enum.  Coerce them to 'public' (the DB
+    default) instead of letting Pydantic raise a ValidationError on the
+    read path.
+
+    Args:
+        v: Visibility value to normalize.
+
+    Returns:
+        The original value if valid, 'public' if invalid, or None if None.
+    """
+    if v is not None and v not in _VALID_VISIBILITY:
+        logger.warning("Coercing invalid visibility value %r to 'public'", v)
+        return "public"
+    return v
+
 
 def encode_datetime(v: datetime) -> str:
     """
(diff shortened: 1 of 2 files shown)


crivetimihai:
Thanks @shoummu1. Well-structured fix — replacing unconstrained strings with `Literal` enums is the right approach, and removing the dead validators keeps it clean. Good test coverage across all impacted schemas. LGTM.

crivetimihai (approved):
Reviewed and rebased onto main. The Literal-enum approach is correct and addresses #3525.

**Changes made during review (3 additional commits):**

1. **Removed remaining redundant validators** — `validate_visibility` was left behind in ToolCreate, ResourceCreate, PromptCreate; now all 6 dead-code validators are removed consistently.

2. **Preserved `default=None` for ToolCreate/ResourceCreate/PromptCreate** — these schemas are used in MCP discovery where `gateway_service.py` inherits gateway visibility via `getattr(resource, "visibility", None) or gateway.visibility`. A `"public"` default would have broken this inheritance, forcing all MCP-discovered entities to be public regardless of their gateway's visibility (security-relevant).

3. **Added defensive coercion on Read schemas** — DB columns are unconstrained `String(20)` with no check constraints. Historical writes through previously unconstrained Update endpoints may have persisted invalid values. Added `_coerce_visibility` (mode="before" field_validator) to all 7 Read schemas so invalid legacy values are coerced to `"public"` with a warning log instead of crashing list/get endpoints.

All 14,640 unit tests pass.

PR #3695, opened 2026-03-16
feat(auth): add Generic OIDC group-to-team mapping for SSO (#2120)

hughhennelly:
## Related Issue

Closes #2120

## Summary

Implements Generic OIDC group-to-team mapping for SSO authentication, following the pattern established by EntraID role mapping in PR #2129.

## What's included:

- **Config Settings**: 5 new `SSO_GENERIC_*` environment variables (groups_claim, admin_groups, role_mappings, default_role, sync_roles_on_login)

- **Bootstrap Support**: Generic OIDC provider now includes `provider_metadata` with group mapping configuration
- **Group Extraction**: Updated `_normalize_user_info` to extract groups from configurable claims for generic OIDC providers
- **Admin Group Mapping**: Generic OIDC admin group support in `_should_user_be_admin` (case-insensitive matching)
- **Role Mapping**: Generic OIDC support in `_map_groups_to_roles` with early exit when no mappings configured
- **Tests**: 14 unit tests covering group extraction, admin assignment, role mapping, and sync opt-out
- **Existing Test Fix**: Updated mock settings in `test_sso_bootstrap.py` to include new `sso_generic_*` attributes

## Security

- Default role is `None` — no automatic admin access for Generic OIDC users
- Admin privileges only granted via explicit `SSO_GENERIC_ADMIN_GROUPS` configuration
- `sync_roles` flag allows provider-level control over role synchronization

## No migration needed

The `provider_metadata` column already exists from PR #2129.

All lint checks (black, isort, ruff, flake8, interrogate) pass.

PR #3719, opened 2026-03-18
feat(sso): extract groups claim for generic OIDC providers

charlieleith:
## Summary

The generic OIDC normalization path in `_normalize_user_info` does not extract group claims from the identity token or userinfo response. This means `_apply_team_mapping` has no groups to work with for any provider that isn't Keycloak or Entra ID — so `team_mapping` on generic OIDC providers (JumpCloud, Okta, Auth0, etc.) silently does nothing.

## Change

Adds groups extraction to the generic OIDC fallback in `_normalize_user_info`, following the exact same pattern already used by the Keycloak and Entra paths:

- Reads `groups_claim` from `provider_metadata` (defaults to `"groups"`)
- Extracts the claim value from `user_data` (handles both list and string)
- Includes `"groups"` key in the normalized return dict

**10 lines added, 0 removed. Single file change.**

## Use case

We're using ContextForge with JumpCloud as a generic OIDC provider. JumpCloud includes a `groups` claim in the ID token when configured. Without this patch, those groups are silently dropped by `_normalize_user_info`, making `team_mapping` on the SSO provider record non-functional.

With the patch, setting `provider_metadata: {"groups_claim": "groups"}` and `team_mapping: {"Engineering": "<team-id>"}` on the SSO provider works as expected — users are auto-added to teams on login based on their IdP groups.

## Testing

- Verified with JumpCloud OIDC provider on ContextForge 1.0.0-RC-2
- Confirmed Keycloak and Entra paths remain unchanged (they have their own extraction logic that runs before the generic fallback)
- The generic path only executes for providers that don't match `keycloak`, `entra`, `github`, or `okta` by ID

### DIFF
--- a/mcpgateway/services/sso_service.py
+++ b/mcpgateway/services/sso_service.py
@@ -1444,6 +1444,15 @@ class SSOService:
         # _is_email_verified_claim's absent-means-pass-through logic applies correctly.
         # Injecting None (via .get()) when the key is missing would cause the key to
         # be present in the dict with a falsy value, silently blocking login.
+        metadata = provider.provider_metadata or {}
+        groups_claim = metadata.get("groups_claim", "groups")
+        groups = []
+        if groups_claim in user_data:
+            gc = user_data.get(groups_claim, [])
+            if isinstance(gc, list):
+                groups = gc
+            elif isinstance(gc, str):
+                groups = [gc]
         generic_normalized: Dict[str, Any] = {
             "email": user_data.get("email"),
             "full_name": user_data.get("name"),
@@ -1451,6 +1460,7 @@ class SSOService:
             "provider_id": user_data.get("sub"),
             "username": user_data.get("preferred_username") or user_data.get("email", "").split("@")[0],
             "provider": provider.id,
+            "groups": groups,
         }
         if "email_verified" in user_data:
             generic_normalized["email_verified"] = user_data["email_verified"]
--- a/mcpgateway/services/sso_service.py
+++ b/mcpgateway/services/sso_service.py
@@ -1252,6 +1252,15 @@ class SSOService:
                     if claim in keycloak_id_token_claims and claim not in user_data:
                         user_data[claim] = keycloak_id_token_claims[claim]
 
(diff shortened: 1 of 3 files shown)


crivetimihai:
Thanks @charlieleith. Clean implementation — configurable `groups_claim` with string/list handling is the right approach. Two notes: (1) no tests — please add a unit test for `_normalize_user_info` covering the groups extraction path (list, string, missing claim), (2) PRs #3597 and #3695 also touch OIDC groups — we'll need to coordinate merge order to avoid conflicts.

crivetimihai (approved):
Thanks for the contribution, @charlieleith! The core fix is correct and well-scoped — the generic OIDC path was the only one missing groups extraction, which silently broke `team_mapping` for providers like JumpCloud.

I've rebased onto `main` (resolved one conflict where the generic OIDC section was refactored to handle `email_verified` omission safely), removed the AI co-author tag, and added two follow-up commits on top of yours:

### Changes I made

**1. ID token groups merging for generic OIDC** (`sso_service.py:1257-1263`)

The Entra and Keycloak paths in `_get_user_info()` merge claims from the verified ID token into `user_data` before normalization, but generic OIDC providers didn't. This means providers that emit groups only in the ID token (not the userinfo response) would still get `groups=[]`. Added a merge block that copies the configured `groups_claim` from `verified_id_token_claims` into `user_data` when userinfo doesn't include it. Userinfo takes precedence when present.

**2. Non-string element filtering** (`sso_service.py:1462`)

Changed `groups = gc` to `groups = [g for g in gc if isinstance(g, str)]`. The downstream `_map_groups_to_roles()` does `if group in role_mappings` (dict-key lookup), which would raise `TypeError: unhashable type: 'dict'` if a malformed IdP response included non-string elements in the groups list.

**3. Test coverage** (11 new tests total)

- 8 normalization tests covering all branches: default claim, custom claim, string value, absent claim, None metadata, non-list/non-string, nested structures filtered, email_verified interaction
- 3 integration tests for `_get_user_info()`: id_token merge, userinfo precedence, custom claim name from id_token

All 216 SSO tests pass.

Issue #3581, opened 2026-03-10
[BUG][A2A]: Team-scoped A2A agents return 500 on test invocation


IuliaGaitan:
### 🐞 Bug Summary
Team created A2A agents are showing 500 when the user is trying to test

---

### 🧩 Affected Component
Select the area of the project impacted:

- [X] `mcpgateway` - API
- [X] `mcpgateway` - UI (admin panel)
- [ ] `mcpgateway.wrapper` - stdio wrapper
- [ ] Federation or Transports
- [ ] CLI, Makefiles, or shell scripts
- [ ] Container setup (Docker/Podman/Compose)
- [ ] Other (explain below)

---

### 🔁 Steps to Reproduce

1. Go to DEV environment
2. Switch to a Team
3. Create a Team or Private A2A agent
4. Click Test after creation
5. Click Test Agent

---
### 🤔Results
What should have happened instead?

Private A2A agents test error

<img width="1906" height="801" alt="Image" src="https://github.com/user-attachments/assets/0a3beaee-499d-4949-829f-2eae9c9a5738" />

Team A2A agent error
<img width="1878" height="952" alt="Image" src="https://github.com/user-attachments/assets/0d3eece7-6a53-4631-952a-c140e9e23769" />

---
### 🤔 Expected Behavior
What should have happened instead?

The Test Agent should work as it works for Public A2A agents
---

### 📓 Logs / Error Output
Paste any relevant stack traces or logs here.
⚠️ **Do not paste secrets, credentials, or tokens.**

---

### 🧠 Environment Info
You can retrieve most of this from the `/version` endpoint.

| Key | Value |
|-----|-------|
| Version or commit | `1.0.0-RC-2` |
| Runtime | `3.12.12` |
| Platform / OS | `Linux 4.18.0-372.64.1.el8_6.x86_64 (x86_64)` |
| Container | `none` |

---

### 🧩 Additional Context (optional)
Add any configuration details, flags, or related issues.

crivetimihai:
Thanks @IuliaGaitan. The 500 when testing team-scoped A2A agents likely occurs because the test invocation path doesn't properly resolve team context. Could you share the server logs (the actual 500 traceback) so we can pinpoint the exact failure?

IuliaGaitan:
Hi @crivetimihai sems that we do not log the error from the UI on DEV environment, so I can nor provide you this info. 
Thank you!

PR #3725, opened 2026-03-18
fix(a2a): A2A agent test endpoint returns 500 for admin users

shoummu1:
# Bug-fix PR
## 📌 Summary

Admin users whose JWT carries `teams: null` (the admin-bypass sentinel) hit a `TypeError: argument of type 'NoneType' is not iterable` whenever they invoked `POST /admin/a2a/{agent_id}/test` on a **team-visibility** A2A agent. The route was not forwarding user-context to the service layer, so `_check_agent_access` received `token_teams=None` and crashed on `return agent.team_id in token_teams`.

## 🔗 Related Issue
Closes: #3581

## 🔁 Reproduction Steps

1. Create a team-visibility A2A agent.
2. Log in as a platform admin (JWT `is_admin: true`, `teams: null`).
3. Call `POST /admin/a2a/{agent_id}/test`.
4. Observe `500 Internal Server Error` — `TypeError: argument of type 'NoneType' is not iterable`.

## 🐞 Root Cause

Two defects combined to cause the crash:

1. **`mcpgateway/admin.py` — `admin_test_a2a_agent`** called `a2a_service.get_agent(db, agent_id)` without passing `user_email` or `token_teams`, so the service received `token_teams=None` regardless of who made the request.

2. **`mcpgateway/services/a2a_service.py` — `_check_agent_access`** lacked a guard for `token_teams=None` inside the `agent.visibility == "team"` branch. For public-visibility agents the function returned early (safe); for team-visibility agents it fell through to `return agent.team_id in token_teams`, raising `TypeError` when `token_teams` was `None`.

## 💡 Fix Description

**Service layer (crash fix):** Added an early-return guard in `_check_agent_access` for the admin-bypass case:
```python
if agent.visibility == "team":
    if token_teams is None:   # admin bypass — full access
        return True
    return agent.team_id in token_teams
```

**Route layer (context fix):** `admin_test_a2a_agent` now extracts `is_admin` and `token_teams` from the authenticated user dict and forwards them to both `get_agent` and `invoke_agent`. For plain admins (`is_admin=True`, `token_teams=None`) the `invoke_user_email` is set to `None` to preserve the admin-bypass semantics throughout the call stack; for token-scoped callers the actual email is used. This mirrors the pattern already used in other admin endpoints (e.g. tags at `admin.py:13432`).

**Tests:** Updated one stale assertion in `test_admin.py` (was missing the new keyword args) and added one assertion in `test_a2a_service.py` that directly covers the `token_teams=None` + team-visibility path introduced by this fix.

## 🧪 Verification

| Lint suite                            | `make lint`          |     Pass   |
| Unit tests                            | `make test`          |    Pass    |
| Coverage ≥ 90 %                       | `make coverage`      |    Pass    |
| Manual regression no longer fails     | Repeat reproduction steps above — endpoint returns 200 with agent response |   Pass     |

## ✅ Checklist

### DIFF
--- a/mcpgateway/admin.py
+++ b/mcpgateway/admin.py
@@ -15180,6 +15180,10 @@ async def admin_test_a2a_agent(
         user_email = get_user_email(user)
         is_admin = user.get("is_admin", False) if isinstance(user, dict) else False
         token_teams = user.get("token_teams") if isinstance(user, dict) else None
+        # Missing token_teams key for non-admin = public-only (per normalize_token_teams rules).
+        # Only admin users retain None (admin bypass); all others default to [].
+        if not is_admin and token_teams is None:
+            token_teams = []
         # Admin users with unrestricted tokens get full bypass (both None);
         # non-admin users pass their actual email and team scoping.
         invoke_user_email = None if (is_admin and token_teams is None) else user_email
--- a/tests/unit/mcpgateway/test_admin.py
+++ b/tests/unit/mcpgateway/test_admin.py
@@ -4874,12 +4874,13 @@ class TestA2AAgentManagement:
         body = json.loads(result.body)
         assert body["success"] is True
         assert "result" in body
-        mock_get_agent.assert_called_with(mock_db, "agent-1", user_email="test-user", token_teams=None)
+        # Non-admin user without token_teams key → normalized to [] (public-only)
+        mock_get_agent.assert_called_with(mock_db, "agent-1", user_email="test-user", token_teams=[])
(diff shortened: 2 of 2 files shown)


crivetimihai:
Thanks @shoummu1. Good fix — correctly addresses the missing admin bypass in `_check_agent_access` for `token_teams=None`. The approach of deriving `invoke_user_email` from the admin/teams state in `admin_test_a2a_agent` is consistent with the token scoping model. Tests cover the new path. LGTM.

PR #3772, opened 2026-03-20
updating dependancies for dependabot

Yosiefeyob:
# Summary
---

Proactive security hardening: upgrade build and transitive dependencies to their latest patched versions and clean up dead code from a recently reverted login-page feature.

---

# Changes

| Package | From | To | Scope |
|---------|------|----|-------|
| authlib |v1.6.8  | 1.6.9 |runtime  |
| Black |v26.1.0| `>=26.3.1` |runtime |
| Undici |v7.22.0 | `>=7.24.5` |runtime  |  
| PyJWT |v2.11.0 | >=2.12.0 |runtime | 
|github.com/buger/jsonparser|1.1.1         |1.1.2           |  runtime|

---

# Impact
Security hardening only — no new features, no API changes, no behavioral changes.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -12,19 +12,19 @@ name: Lint & Static Analysis
 on:
   push:
     branches: ["main"]
-    paths:
-      - "mcpgateway/**"
-      - "plugins/**"
-      - "pyproject.toml"
-      - ".github/workflows/lint.yml"
+    # paths:
+    #   - "mcpgateway/**"
+    #   - "plugins/**"
+    #   - "pyproject.toml"
+    #   - ".github/workflows/lint.yml"
   pull_request:
     types: [opened, synchronize, ready_for_review]
     branches: ["main"]
-    paths:
-      - "mcpgateway/**"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -31,6 +31,7 @@
         "minimatch": ">=10.2.1"
     },
     "dependencies": {
-        "alpinejs": "^3.15.8"
+        "alpinejs": "^3.15.8",
+        "undici": "^7.24.5"
```
lucarlig: `undici` looks like it was previously only coming in transitively via `jsdom`, so adding it to top-level `dependencies` promotes a test-only package into the production install set and also introduces its Node engine requirement at runtime. If the goal is just to pin a patched transitive version, could you move this to `overrides` in `package.json` instead of `dependencies` and regenerate `package-lock.json`?

Something like this should be enough:

```json
{
  "overrides": {
    "keyv": "5.1.0",
    "minimatch": ">=10.2.1",
    "undici": "^7.24.5"
  }
}

Yosiefeyob: updated undici to overrides instead of dependencies.

Issue #3779, opened 2026-03-21
[BUG]: Editing a Virtual Server via the admin UI silently overwrites the server's `team_id`


kimsehwan96:
### 🐞 Bug Summary
Editing a Virtual Server via the admin UI silently overwrites the server's `team_id`, even when no team change was intended. This happens in **all** editing contexts:

- **Team-scoped view** (`?team_id=X`): The URL's `team_id` is injected as a hidden form field, overwriting the server's original team.
- **All Teams view** (no `team_id` in URL): `verify_team_for_user()` falls back to the current user's **personal team**, overwriting the server's original team.

The result is that any edit (e.g., adding a tool, changing description) silently moves the server to a different team.

---

### 🧩 Affected Component
Select the area of the project impacted:

- [X] `mcpgateway` - API
- [X] `mcpgateway` - UI (admin panel)
- [ ] `mcpgateway.wrapper` - stdio wrapper
- [ ] Federation or Transports
- [ ] CLI, Makefiles, or shell scripts
- [ ] Container setup (Docker/Podman/Compose)
- [ ] Other (explain below)

---

### 🔁 Steps to Reproduce

1. Create a Virtual Server under **Team A** (e.g., an organization team)
2. Navigate to the admin panel in **All Teams** view (`/admin/#catalog`, no `?team_id=` in URL)
3. Edit the Virtual Server — make any change (e.g., add a tool, update description)
4. Save the edit
5. **Observe:** The server's `team_id` has changed from Team A to the current user's **personal team**
---

### 🤔 Expected Behavior
Editing a Virtual Server should **preserve the original `team_id`** unless the user explicitly changes the team assignment. A simple edit (adding tools, changing description, etc.) should not move the server to a different team.

---

### 📓 Logs / Error Output
Paste any relevant stack traces or logs here.
⚠️ **Do not paste secrets, credentials, or tokens.**

---

### 🧠 Environment Info
You can retrieve most of this from the `/version` endpoint.

| Key | Value |
|-----|-------|
| Version or commit | `1.0.0-RC-2` |
| Runtime | `Python 3.12.12, Gunicorn` |
| Platform / OS | `RHEL 10.1 (Coughlan) / BottleRocket` |
| Container | `Docker (Kubernetes / EKS)` |

---

### 🧩 Additional Context (optional)
- The fix should ensure the edit form populates the hidden `team_id` with the **resource's existing `team_id`** from the server data, not from the URL or user's personal team.
[...]

PR #3780, opened 2026-03-21
fix: preserve server team_id during admin UI edits

kimsehwan96:
## 🔗 Related Issue
Closes: #3779

---

## 📝 Summary
When editing a virtual server via the admin UI, the server's `team_id` was silently overwritten on every save — even when the user did not intend to change the team.
                                                                                                                                                            
  **Root cause:** In `admin_edit_server()`, when no `team_id` is present in the form (e.g., editing from the "All Teams" view), `verify_team_for_user()` falls back to the current user's **personal team**, which is then unconditionally written to the server record.
                                                                                                                                                            
  **Fix:** Before calling `verify_team_for_user()`, the edit endpoint now looks up the existing server's `team_id` from the database and uses it as the default when the form does not include one. This preserves the original team assignment while still allowing explicit team changes via the API.

---

## 🏷️ Type of Change
- [x] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

---

## ✅ Checklist

---

## 📓 Notes (optional)
Recommended future work
- A2A Agent edit has the same pattern (`a2a_service.py` uses generic `setattr` loop)
- Explicit "transfer team" UI/API for intentional team reassignment
- Tool, Gateway, Prompt edit endpoints call `verify_team_for_user()` but their service layers don't persist the result (accidentally safe)

### DIFF
--- a/tests/unit/mcpgateway/services/test_a2a_service.py
+++ b/tests/unit/mcpgateway/services/test_a2a_service.py
@@ -334,6 +334,41 @@ class TestA2AAgentService:
             with pytest.raises(A2AAgentError, match="without a team_id"):
                 await service.update_agent(mock_db, sample_db_agent.id, update_data)
 
+    async def test_update_agent_team_id_rejects_non_owner(self, service, mock_db, sample_db_agent):
+        """Reassigning an agent to a team where user is not owner must raise."""
+        from mcpgateway.services.a2a_service import _validate_a2a_team_assignment
+
+        mock_query = MagicMock()
+        mock_query.filter.return_value = mock_query
+        # Team exists but membership check returns None
+        mock_query.first.side_effect = [MagicMock(), None]
+
+        mock_session = MagicMock()
+        mock_session.query.return_value = mock_query
+
+        with pytest.raises(ValueError, match="membership"):
+            _validate_a2a_team_assignment(mock_session, "user@example.com", "00000000-0000-0000-0000-000000000099")
+
+    async def test_update_agent_team_id_skips_ownership_without_user_email(self, service, mock_db, sample_db_agent):
+        """System updates without user_email skip ownership checks and persist team_id."""
+        sample_db_agent.version = 1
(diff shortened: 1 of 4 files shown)


crivetimihai:
Thanks for catching this @kimsehwan96 — the root cause analysis was spot-on.

I rebased onto main, expanded the fix to cover all entity types, and added service-layer persistence + validation. Here's what the final branch contains:

**Commits:**
1. **Your original fix** — admin-side `team_id` preservation for server edits
2. **Extended admin-side fix** — same pattern applied to gateway, resource, prompt, and A2A agent edit endpoints. Also fixed resource edit missing `team_id`/`owner_email` entirely, and a typo in a prompt edit log message.
3. **Service-layer persistence** — gateway, resource, and prompt services were accepting `team_id` in their update DTOs but never writing it to the ORM row. Now all three persist it with ownership validation (matching `server_service`'s pattern). Also added missing team reassignment validation to `a2a_service` (which was persisting via generic `setattr` without checks).
4. **Visibility transition validation** — when visibility changes to `"team"`, all services now validate the effective `team_id` (from the update or existing row), matching `server_service`'s stricter pattern.
5. **Deny-path service tests** — 8 tests verifying non-existent team rejection and missing `team_id` rejection at the service layer for all 4 entity types.

**Test results:** 6310 passed, 0 failures.

kimsehwan96:
@crivetimihai Thanks for fast review and addintional works!

crivetimihai:
@kimsehwan96 thanks for the PR. Merging. Can you please test that everything still works correctly post merge?

Thank you!

kimsehwan96:
@crivetimihai I tested it from main branch. It correctly works. 

But, If some user who are not member of that team try to change MCP server configuration (via admin UI edit), then just show `Invalid team_id` error message with 422 Unprocessable Entity. I think the message should be more clarifier like `Permission deined: you are not a member of the team that owns this resource` 

I will create issue related this. Thanks!

PR #3785, opened 2026-03-22
fix: add forbidden-pattern check to ToolUpdate.validate_description (…

### LINKED ISSUE #3773: [BUG]: ToolUpdate.validate_description missing forbidden-pattern check present in ToolCreate
## Bug Description

`ToolCreate.validate_description` enforces a forbidden-pattern check (shell/pipe metacharacters: `&&`, `;`, `||`, `$(`, `|`, `> `, `< `) gated behind `VALIDATION_STRICT`. However, `ToolUpdate.validate_description` has **no such check at all** — it only performs length truncation and `sanitize_display_text()`.

This means a tool registered with a safe description can later have its description updated to include shell metacharacters, bypassing the validation that was enforced at creation time.

## Reproduction Steps

1. Register a tool with a safe description (passes `ToolCreate` validation)
2. Update the tool's description via `PUT`/`PATCH` to include `$(malicious)` or `cmd1 && cmd2`
3. The update succeeds without any warning or rejection, regardless of `VALIDATION_STRICT`

## Expected Behavior

`ToolUpdate.validate_description` should apply the same forbidden-pattern check as `ToolCreate.validate_description`, gated by `settings.validation_strict`.

## Root Cause

**File:** `mcpgateway/schemas.py`

`ToolCreate.validate_description` (around line 526) has:
```python
forbidden_patterns = ["&&", ";", "||", "$(", "|", "> ", "< "]
for pat in forbidden_patterns:
    if pat in v:
        if settings.validation_strict:
            raise ValueError(f"Description contains unsafe characters: '{pat}'")
        logger.warning(...)
        break
```

`ToolUpdate.validate_description` (around line 1043) only does:
```python
if len(v) > SecurityValidator.MAX_DESCRIPTION_LENGTH:
    truncated = v[: SecurityValidator.MAX_DESCRIPTION_LENGTH]
    ...
return SecurityValidator.sanitize_display_text(v, "Description")
```

No other `*Update` schema classes have the check either, but `ToolCreate` is the only `*Create` class that does, so the gap is specifically between `ToolCreate` and `ToolUpdate`.

## Proposed Fix

Extract the forbidden-pattern check into a shared helper (or inline it into `ToolUpdate.validate_description`) so both create and update paths enforce the same validation. The check should remain gated by `settings.validation_strict`.

## Related

- #3711 — Original report of hardcoded forbidden-pattern check
- #3749 — Fix that gated the check behind `VALIDATION_STRICT`

omorros:
# 🔗 Related Issue 

## Closes #3773                                                                                                                                             

  ---
  ## 📝 Summary

  ToolUpdate.validate_description was missing the forbidden-pattern check (&&, ;, ||, $(, |, > , < ) that ToolCreate.validate_description enforces. This allowed a tool registered with a safe description to later have its description updated to include shell metacharacters, bypassing the validation enforced at creation time. This PR adds the same check to ToolUpdate.validate_description.

  ---
  ## 🏷️ Type of Change

[ x] Bug fix
  [ ] Feature / Enhancement
  [ ] Documentation
  [ ] Refactor
  [ ] Chore (deps, CI, tooling)
  [ ] Other (describe below)

  ---
  ## 🧪 Verification

  | Check | Command | Status |
  |---|---|---|

  ---
  ## ✅ Checklist

  [x] Code formatted (make black isort pre-commit)
  [x] Tests added/updated for changes
  [ ] Documentation updated (if applicable)
  [x] No secrets or credentials committed

  ---
  ## 📓 Notes (optional)

  Added 4 unit tests in test_schemas_validators_extra.py:
  Parametrized test rejecting all 7 forbidden patterns on ToolUpdate, safe description acceptance test, None passthrough test, and parity test verifying   
  ToolCreate and ToolUpdate reject the same pattern set.

### DIFF
--- a/tests/unit/mcpgateway/test_schemas_validators_extra.py
+++ b/tests/unit/mcpgateway/test_schemas_validators_extra.py
@@ -198,6 +198,28 @@ class TestToolUpdateDescriptionValidationStrict:
             with pytest.raises(ValueError, match="unsafe characters"):
                 ToolUpdate.validate_description(payload)
 
+    def test_empty_string_accepted(self):
+        """Empty string is a valid description (not None, not forbidden)."""
+        assert ToolUpdate.validate_description("") == ""
+
+    def test_forbidden_pattern_at_start(self, monkeypatch):
+        """Forbidden pattern at the very start of a description is still caught."""
+        monkeypatch.setattr(settings, "validation_strict", True)
+        with pytest.raises(ValueError, match="unsafe characters"):
+            ToolUpdate.validate_description("&&rest of description")
+
+    def test_forbidden_pattern_at_end(self, monkeypatch):
+        """Forbidden pattern at the very end of a description is still caught."""
+        monkeypatch.setattr(settings, "validation_strict", True)
+        with pytest.raises(ValueError, match="unsafe characters"):
+            ToolUpdate.validate_description("description ends with ||")
+
+    def test_exact_max_length_not_truncated(self):
+        """Description exactly at MAX_DESCRIPTION_LENGTH is accepted without truncation."""
+        exact = "x" * SecurityValidator.MAX_DESCRIPTION_LENGTH
+        result = ToolUpdate.validate_description(exact)
+        assert len(result) == SecurityValidator.MAX_DESCRIPTION_LENGTH
+
 
 def test_resource_update_content_and_description():
     long_desc = "x" * (SecurityValidator.MAX_DESCRIPTION_LENGTH + 5)


crivetimihai:
Thanks for the contribution @omorros! The forbidden-pattern check for `ToolUpdate.validate_description` is a good catch — closing the gap where a tool could bypass creation-time validation via update.

I rebased onto main and made two additional commits:

1. **`fix: respect VALIDATION_STRICT in ToolUpdate.validate_description`** — The original implementation always raised `ValueError`, ignoring `settings.validation_strict`. This broke symmetry with `ToolCreate`, which allows descriptions with Markdown syntax (e.g. `> blockquote`, `cmd | grep`) when `VALIDATION_STRICT=false`. Fixed to gate on the setting and log a warning in non-strict mode, matching `ToolCreate` exactly.

2. **`test: add edge-case tests for ToolUpdate forbidden-pattern validation`** — Expanded the test class to mirror `TestToolCreateDescriptionValidationStrict` (strict rejection, non-strict acceptance with warning, single-warning logging, both-mode passthrough) and added boundary tests (empty string, patterns at start/end of string, exact `MAX_DESCRIPTION_LENGTH`).

All 75 tests in the file pass.

PR #3803, opened 2026-03-23
fix: tighten pii filter behavior and rust masking strategy

### LINKED ISSUE #3724: [BUG][PLUGINS]: PIIFilterPlugin ignores default_mask_strategy config — built-in PII types always use hardcoded masking strategies
### 🐞 Bug Summary
The \`default_mask_strategy\` config field in \`PIIFilterPlugin\` has no effect on built-in PII types. For example, setting \`default_mask_strategy: \"redact\"\` does not change SSN masking — it still produces \`***-**-6789\` instead of \`[PII_REDACTED]\`.

---

### 🧩 Affected Component

- [ ] `mcpgateway` - API
- [x] Other: \`plugins/pii_filter/pii_filter.py\`

---

### 🔁 Steps to Reproduce
1. Set \`default_mask_strategy: \"redact\"\` and \`redaction_text: \"[PII_REDACTED]\"\` in the \`PIIFilterPlugin\` config in \`plugins/config.yaml\`
2. Restart the server
3. Invoke a prompt with an SSN: \`{\"user_input\": \"My SSN is 123-45-6789\"}\`

---

### 🤔 Expected Behavior
SSN should be replaced with \`[PII_REDACTED]\` when \`default_mask_strategy: \"redact\"\` is configured.
---

### 📓 Logs / Error Output
Image

---

### 🧠 Root Cause
In \`_compile_patterns()\", each built-in \`PIIPattern\` is created with a hardcoded \`mask_strategy\`:
\`\`\`python
PIIPattern(type=PIIType.SSN, ..., mask_strategy=MaskingStrategy.PARTIAL)
\`\`\`
This is stored in the detection dict at detection time. In \`mask()\`, the line:
\`\`\`python
strategy = detection.get(\"mask_strategy\", self.config.default_mask_strategy)
\`\`\`
always finds the hardcoded value — the fallback to \`default_mask_strategy\` is never reached.

The fix is to use \`self.config.default_mask_strategy\` when constructing built-in \`PIIPattern\` objects in \`_compile_patterns()\` instead of hardcoding per-type strategies."

---

### LINKED ISSUE #3738: [BUG][RUST]: Rust PII filter — correctness issues, false positives, and broken redaction config
### 🐞 Bug Summary
The Rust PII filter refactor currently has multiple correctness and reliability problems that make it unsafe to treat as production-ready.

Current issues, Documentaion, implementation bugs and scope problems, including:

- crashes or instability on some inputs
- configuration features that do not work as expected, such as redaction settings not actually redacting
- overly broad regexes that produce a large volume of false positives
- category definitions that are too broad or inconsistent with what the implementation can reliably detect
- mixed international/generalized detection scope when the immediate milestone should be narrowed to US-focused coverage only
- No details docs
- No Logs

For the current phase, the goal should be a smaller, reliable US-focused detector set with working masking/redaction behavior and explicit reduction of high-noise patterns.

I also plan to decouple this work from the Python plugin because, after the recent changes, the functionality has diverged significantly and should no longer be treated as the same implementation track.

---

### 🧩 Affected Component

- [ ] `mcpgateway` - API
- [x] Other (explain below)

Other: Rust PII filter plugin and Python integration layer  
Primary areas:
- `plugins_rust/pii_filter/`

---

### 🔁 Steps to Reproduce

1. Enable the PII filter with Rust implementation available.
2. Run representative prompt/response payloads that include:
   - valid US PII that should be redacted
   - benign numeric/text fixtures that should not be flagged
   - configuration combinations that rely on redaction/masking settings
3. Observe one or more of the following:
   - redaction configured but original values remain visible
   - benign content is flagged at a very high rate
   - category output is broader or less specific than expected
   - crashes or unstable behavior on some payloads

---

### 🤔 Expected Behavior

For this milestone, the Rust PII filter should:

- reliably avoid crashes on supported inputs
- honor configured redaction/masking behavior consistently
- reduce high-volume false positives from overly broad regexes
- narrow supported scope to a well-defined US-focused set for now
- avoid advertising broad categories that are not yet implemented with acceptable accuracy
- prefer a smaller set of trustworthy detectors over a larger noisy set

Proposed current scope for this phase:

-
### Rationale for narrowing to US-first

The current implementation tries to cover a broad mix of PII categories and international formats, but that breadth is part of what is creating instability, poor precision, and unclear detector boundaries.

We are narrowing to US-focused detection first because we do not currently have the engineering time, testing bandwidth, or domain coverage needed to deliver high-quality worldwide PII detection responsibly.

This decision is driven by practical constraints:

- we do not have the resources right now to build, validate, and maintain full worldwide coverage with acceptable quality
- US-oriented patterns give us a smaller and more testable target for this milestone
- many broader international and generic identifier categories are highly ambiguous and currently generate too many false positives
- reducing scope lets us focus on core correctness first: crashes, redaction behavior, category accuracy, and regression coverage
- a narrower milestone gives us a stable baseline that we can expand later in a deliberate way instead of continuing to tune an over-broad and noisy detector set

This is a sequencing and resourcing decision, not a statement that non-US PII is unimportant

lucarlig:
## 🔗 Related Issue
Closes #3724
Closes #3738

---

## 📝 Summary
This PR tightens the PII filter behavior across both the plugin layer and the Rust detector.

At the plugin level, it preserves the current enforcement/permissive flow while making the fallback path clearer: when the Rust detector is unavailable, the legacy Python detector now emits a one-time deprecation warning instead of silently taking over. The branch also expands test coverage around plugin configuration behavior, including masking modes, whitelist handling, detection metadata, and Python-fallback signaling.

At the Rust layer, it fixes the core masking bug where built-in detections were ignoring the configured `default_mask_strategy` and forcing hardcoded partial masks. Built-in SSN/email/AWS detections now honor the configured default strategy, while custom patterns still keep their explicit mask-strategy overrides.

The PII filter README is also updated to match current reality: correct test paths, current prompt invocation examples, and the correct local/compose ports for manual verification.

---

## 🏷️ Type of Change
- [x] Bug fix
- [ ] Feature / Enhancement
- [x] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

| Check                     | Command                                                                 | Status |
|---------------------------|-------------------------------------------------------------------------|--------|
| Lint suite                | `make --no-print-directory pre-commit`                                  | ✅ |
| Unit tests                | `cargo test` and `pytest tests/unit/mcpgateway/plugins/plugins/pii_filter/test_pii_filter.py -q` | ✅ |
| Coverage ≥ 80%            | Not run                                                                 | N/A |

---

## ✅ Checklist

---

## 📓 Notes (optional)
- Local compose verification reproduced the original Rust masking issue before the fix: with `default_mask_strategy=redact`, built-in detections still returned `partial`.
- After the fix, built-in detections honor the configured default strategy and custom patterns still preserve explicit overrides.
- A separate local database-state issue surfaced during compose verification: an old Postgres volume was stamped to an Alembic revision no longer present in the branch. Resetting the local dev DB volume resolved that environment-specific problem.

### DIFF
--- a/plugins/pii_filter/pii_filter.py
+++ b/plugins/pii_filter/pii_filter.py
@@ -59,8 +59,7 @@ except Exception as e:
 
 
 _PYTHON_PLUGIN_DEPRECATION_MESSAGE = (
-    "The legacy Python PII filter detector is deprecated and will be removed in a future release. "
-    "Install the Rust-backed `pii_filter_rust` package to keep the PII filter plugin enabled."
+    "The legacy Python PII filter detector is deprecated and will be removed in a future release. Install the Rust-backed `pii_filter_rust` package to keep the PII filter plugin enabled."
 )
 
 
--- a/plugins_rust/pii_filter/python/pii_filter_rust/__init__.pyi
+++ b/plugins_rust/pii_filter/python/pii_filter_rust/__init__.pyi
@@ -3,6 +3,7 @@
 
 import builtins
 import typing
+
 __all__ = [
     "PIIDetectorRust",
 ]
@@ -11,18 +12,18 @@ __all__ = [
 class PIIDetectorRust:
     r"""
     Main PII detector exposed to Python
-    
+
     # Example (Python)
     ```python
     from pii_filter import PIIDetectorRust
-    
+
     config = {"detect_ssn": True, "detect_email": True}
     detector = PIIDetectorRust(config)
-    
+
     text = "My SSN is 123-45-6789 and email is john@example.com"
     detections = detector.detect(text)
     print(detections)  # {"ssn": [...], "email": [...]}
-    
+
     masked = detector.mask(text, detections)
     print(masked)  # "My SSN is [REDACTED] and email is [REDACTED]"
     ```
@@ -30,10 +31,10 @@ class PIIDetectorRust:
     def __new__(cls, config: typing.Any) -> PIIDetectorRust:
         r"""
         Create a new PII detector
-        
+
         # Arguments
         * `config` - Python dictionary or Pydantic model with configuration
-        
+
         # Configuration Keys
         * `detect_ssn` (bool): Detect Social Security Numbers
         * `detect_credit_card` (bool): Detect credit card numbers
@@ -55,10 +56,10 @@ class PIIDetectorRust:
     def detect(self, text: builtins.str) -> typing.Any:
         r"""
         Detect PII in text
-        
+
         # Arguments
         * `text` - Text to scan for PII
-        
+
         # Returns
         Dictionary mapping PII type to list of detections:
         ```python
@@ -75,22 +76,22 @@ class PIIDetectorRust:
     def mask(self, text: builtins.str, detections: typing.Any) -> builtins.str:
         r"""
         Mask detected PII in text
-        
+
         # Arguments
         * `text` - Original text
         * `detections` - Detection results from detect()
-        
+
         # Returns
         Masked text with PII replaced
         """
     def process_nested(self, data: typing.Any, path: builtins.str) -> tuple[builtins.bool, typing.Any, typing.Any]:
         r"""
         Process nested data structures (dicts, lists, strings)
-        
+
         # Arguments
         * `data` - Python object (dict, list, str, or other)
         * `path` - Current path in the structure (for logging)
-        
+
         # Returns
         Tuple of (modified: bool, new_data: Any, detections: dict)
         """
--- a/plugins_rust/pii_filter/src/detector.rs
+++ b/plugins_rust/pii_filter/src/detector.rs
@@ -33,7 +33,9 @@ pub fn detect_pii(
                     value: mat.as_str().to_string(),
                     start: mat.start(),
                     end: mat.end(),
-                    mask_strategy: pattern.mask_strategy.unwrap_or(config.default_mask_strategy),
+                    mask_strategy: pattern
+                        .mask_strategy
+                        .unwrap_or(config.default_mask_strategy),
                 };
 
                 detections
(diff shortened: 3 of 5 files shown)


crivetimihai:
## Review Notes

Rebased onto current `main` (clean, no conflicts).

### Code Review

**The core fix is correct and well-designed.** Changing `CompiledPattern.mask_strategy` from `MaskingStrategy` to `Option<MaskingStrategy>` cleanly separates the two cases:
- Built-in patterns: `None` → falls back to `config.default_mask_strategy` via `unwrap_or`
- Custom patterns: `Some(strategy)` → respects the user's explicit per-pattern override

Applied consistently in both `detect_pii()` (public benchmark API) and `detect_internal()` (PyO3-bound detector).

### Security
This is a security **improvement**. Before this fix, even with `default_mask_strategy=redact`, built-in detections (SSN, email, credit card, etc.) leaked partial data (e.g. `***-**-6789`). Now `Redact` properly produces `[REDACTED]` as configured.

### Tests
- Python tests: 86 passed, 88 skipped (Rust not compiled in test env)
- Rust tests: 16 passed, 0 failed
- New differential code has 100% test coverage (deprecation warning, Rust strategy propagation, custom pattern override)
- The removed `REMOVE` strategy test was correctly removed — it was a latent bug (asserted partial-masking output while configuring REMOVE strategy; only passed because the old code ignored `default_mask_strategy`)

### Observations (non-blocking)
- **Python/Rust behavioral divergence during deprecation**: The legacy Python `PIIDetector` still hardcodes per-pattern `mask_strategy` (e.g. `MaskingStrategy.PARTIAL` for SSN), so it continues to ignore `default_mask_strategy` for built-in patterns. Users on the Python fallback will see different masking behavior than Rust users. Acceptable given the Python path is now formally deprecated, but worth noting.

PR #3806, opened 2026-03-23
Fixed model selection dropdown appearance and interaction 

### LINKED ISSUE #3800: [BUG]:  LLM Settings: model ID combobox interaction
### 🐞 Bug Summary

In the Add/Edit LLM Model modal (LLM Settings tab), the Model ID field uses a native HTML <datalist> for suggestions. After selecting a model, the browser filters the suggestion list to only entries matching the current input value. This shows a single option that appears visually like a tooltip instead of the full model list. This makes it impossible to switch to a different model without manually clearing the field first. The native datalist also renders inconsistently across browsers, appearing as a floating tooltip rather than a styled dropdown consistent with the rest of the admin UI.

### 🧩 Affected Component

- [ ] `mcpgateway` - API
- [ ] Other (explain below)

### 🔁 Steps to Reproduce
1. Navigate to the admin panel → LLM Settings tab → Models sub-tab
3. Click Add Model
4. Select a provider that has models available (e.g. Ollama)
5. Click the refresh (↺) button to fetch available models
6. Select any model from the suggestion list (e.g. granite4:tiny-h)
7. Click back into the Model ID field and attempt to view or switch to a different model

### 🤔 Expected Behavior
Clicking or focusing the Model ID field should display the full list of available models regardless of the currently selected value, allowing the user to browse and switch selections freely. The suggestions list should be styled consistently with the rest of the admin UI (matching the Provider `select` dropdown above it).

### 📓 Logs / Error Output

No console errors. Visual-only regression — the native `datalist` popup filters to a single matching entry and renders as a browser-native tooltip bubble.

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Version or commit | main |
| Runtime | Python 3.11 |

The field is implemented as `<input type="text" list="llm-model-suggestions"> + <datalist>` populated via fetchModelsForModelModal(). The fix is to replace it with a custom Alpine.js combobox (consistent with the existing team-selector pattern in the codebase) that: stores the full model list independently of the filter text, resets the filter on open to always show all options, and falls back to free-text entry for model IDs not returned by the provider API.

a-effort:
## 🔗 Related Issue
Closes #3800

---

## 📝 Summary
 Replaces the native HTML `datalist` on the Model ID field in the Add/Edit LLM Model modal with a custom styled combobox. The datalist had two problems: it rendered as a browser-native tooltip bubble inconsistent with the rest of the admin UI, and after selecting a model it filtered suggestions down to a single matching entry, preventing the user from switching to a different model without manually clearing the field. The replacement uses a Tailwind-styled `ul` dropdown with vanilla JS event delegation, matching the existing patterns in the admin panel. Selecting a model populates the input; clicking the field again reopens the full list. Free-text entry for unlisted model IDs is preserved. Also fixes `overflow-hidden` on the model modal card which was clipping the dropdown and making items unclickable.

---

## 🏷️ Type of Change
Bug fix

---

## 🧪 Verification

| Lint suite                | `make lint`     |  ✅  |
| Unit tests                | `make test`     |  ✅  |
| Coverage ≥ 80%            | `make coverage` |  ✅   |

---

## ✅ Checklist

---

## 📓 Notes (optional)

https://github.com/user-attachments/assets/5a7595cf-bae2-459e-bf37-df87cdb1cb9f

---

🤖 AI assisted
🧑🏻‍💻 human reviewed / edited

### DIFF
--- a/mcpgateway/static/admin.js
+++ b/mcpgateway/static/admin.js
@@ -32734,17 +32734,22 @@ function _renderLLMModelDropdown(models) {
     if (!ul) {
         return;
     }
+    ul.innerHTML = "";
     if (!models.length) {
-        ul.innerHTML =
-            '<li class="px-3 py-2 text-xs text-gray-400 dark:text-gray-500">No models found. Enter ID manually.</li>';
+        const li = document.createElement("li");
+        li.className = "px-3 py-2 text-xs text-gray-400 dark:text-gray-500";
+        li.textContent = "No models found. Enter ID manually.";
+        ul.appendChild(li);
         return;
     }
-    ul.innerHTML = models
-        .map(
-            (m) =>
-                `<li class="px-3 py-2 cursor-pointer hover:bg-gray-100 dark:hover:bg-gray-700 text-gray-900 dark:text-gray-100" data-model-id="${m.id.replace(/"/g, "&quot;")}">${m.id}</li>`,
-        )
-        .join("");
+    models.forEach((m) => {
+        const li = document.createElement("li");
+        li.className =
+            "px-3 py-2 cursor-pointer hover:bg-gray-100 dark:hover:bg-gray-700 text-gray-900 dark:text-gray-100";
+        li.dataset.modelId = m.id;
+        li.textContent = m.id;
+        ul.appendChild(li);
+    });
 }
 
 // Wire up delegated events on the dropdown once at load time
@@ -32891,6 +32896,8 @@ window.fetchModelsForModelModal = fetchModelsForModelModal;
  * Edit LLM Model
  */
 async function editLLMModel(modelId) {
+    _llmAllModels = [];
+    llmModelComboboxClose();
     try {
         const response = await fetch(
             `${window.ROOT_PATH}/llm/models/${modelId}`,
--- a/mcpgateway/static/admin.js
+++ b/mcpgateway/static/admin.js
@@ -32704,7 +32704,6 @@ function refreshLLMProviders() {
 let _llmAllModels = [];
 var _llmModelsFetched = false; // eslint-disable-line no-var -- reassigned across function scopes
 let _llmComboboxActiveIndex = -1;
-
 function _llmComboboxSetExpanded(expanded) {
     const input = document.getElementById("llm-model-model-id");
     if (input) input.setAttribute("aria-expanded", String(expanded));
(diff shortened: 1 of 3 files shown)


a-effort:
Notes: there are a couple of follow up issues with the form fields in this modal we can address separately or put on hold until switching over to a component-based approach. I wanted to get these usability fixes in asap, though. 

Proposed follow ups: 

- Placeholder text for the model selection form field should be `Select model` instead of instruction to select the provider first. We should use inline validation to show the help text `Select provider first` underneath the form field if the user tries to do this before selecting a provider. 
- If there's only one provider, pre-select it in that field
- if the user clears the provider field, also clear the model field

a-effort:
Note on pre-commit failures: The two pre-commit hook failures flagged in CI are pre-existing on main and unrelated to this PR.                 
                             
  - check-executables-have-shebangs: `tools_rust/wrapper/src/bin/wrapper_integration.rs` is marked executable but has no shebang. Not touched by this PR.                   
  - scripts/demo_a2a_agent.py / scripts/demo_a2a_agent_auth.py: these files have local modifications on main that trigger pre-commit auto-fixes to JWT secret defaults. Not touched by this PR.

All hooks pass cleanly against the files changed in this PR (mcpgateway/templates/admin.html, mcpgateway/static/admin.js).

crivetimihai:
## Review Summary

Rebased onto `main`, squashed the original 2 commits + merge into a single clean commit (preserving original author), and applied four additional fix commits during review:

### Fixes Applied

1. **XSS hardening** — `_renderLLMModelDropdown` used `innerHTML` with `${m.id}` text content without HTML-escaping. Replaced with safe DOM APIs (`createElement`, `textContent`, `dataset`). Model IDs from external provider APIs could have contained malicious HTML.

2. **Stale combobox state** — `editLLMModel()` didn't reset `_llmAllModels` or close the dropdown. Stale model lists from a previous "Add" modal session would persist when editing.

3. **ARIA & keyboard accessibility** — Added `role=combobox/listbox/option`, `aria-expanded`, `aria-controls`, `aria-activedescendant`. Added keyboard navigation: ArrowDown/ArrowUp to highlight, Enter to select, Escape to close dropdown (with `stopPropagation` so the modal stays open).

4. **Race condition guard** — `fetchModelsForModelModal()` was vulnerable to out-of-order responses (both cross-provider and same-provider concurrent requests). Added a monotonic `_llmFetchSeq` counter + provider-value re-check. Stale responses are silently discarded.

5. **Empty-state UX** — Suppressed the misleading "No models found" dropdown before any fetch has occurred via `_llmModelsFetched` flag.

### Tests

Added 35 Vitest unit tests covering: rendering (including XSS safety), open/close/filter/select, ARIA state management, full keyboard navigation, delegated click/mousedown events, fetch integration (success/empty/error), provider-change staleness, and same-provider concurrent request staleness.

Full suite: **982/982 green** across 30 test files.

crivetimihai (approved):
LGTM after fixes. All codex review findings addressed, 35 unit tests added, 982/982 green.

PR #3783, opened 2026-03-21
feat(rate-limiter): pluggable algorithms, tenant isolation fix, and scale load test

gandhipratik203:
## Summary

Addresses gaps in the rate limiter plugin: no algorithm choice, broken `by_tenant` isolation, sliding window memory leak, missing Redis integration tests, and distorted load test metrics.

---

## Gaps closed

**Gap 1 (HIGH)** — No algorithm choice: only `fixed_window` was available — no way to handle boundary bursts or bursty workloads. Fixed by introducing a strategy pattern with three selectable algorithms (`fixed_window`, `sliding_window`, `token_bucket`) configurable via the `algorithm:` field. All three run on both `memory` and `redis` backends; the Redis backend was extended with atomic Lua scripts for `sliding_window` and `token_bucket` to maintain cluster-wide enforcement across instances.

`test_fixed_window_burst_at_boundary` (xfail) documents the boundary burst as a known trade-off for users who stay on `fixed_window`; `sliding_window` eliminates it entirely.

**Gap 2 (HIGH)** — `by_tenant` cross-throttle: `tenant_id=None` fell back to a shared `"tenant:default"` bucket, causing unrelated users to throttle each other across the entire deployment. Fixed in two places: (1) the plugin now skips the `by_tenant` check entirely when `tenant_id` is `None` instead of inventing a phantom bucket; (2) `mcpgateway/auth.py` now propagates `request.state.team_id` → `global_context.tenant_id` in both `_inject_userinfo_instate` and the `get_current_user` fallback, so single-team API tokens correctly populate tenant context for rate limiting.

**Gap 3 (MEDIUM)** — Sliding window memory leak in `sweep()`: stale-but-non-empty timestamp lists were never evicted from `_store` because `sweep()` only removed empty lists. Fixed by embedding the window size in each store key (`"{key}:{window}"`) so staleness is computable at sweep time, and rewriting `sweep()` to evict any key where all timestamps fall outside the window.

**Gap 4 (LOW)** — Missing Redis integration coverage: the Redis backend had no integration tests exercising a real Redis instance. Added `TestRedisBackendIntegration` with four tests: shared counter across instances, limit enforcement, window reset after TTL expiry, and fallback-to-memory when Redis is unavailable. Tests auto-detect a local Redis at `localhost:6379` or fall back to a Docker container; the suite is skipped gracefully if neither is available.

---

## Additional hardening

[...]

PR #3809, opened 2026-03-23
feat(rate-limiter): pluggable algorithms with Rust-backed execution engine, benchmarks, and validation

gandhipratik203:
## Summary

Consolidates and extends the rate limiter plugin with a Rust-backed execution engine, carrying forward the Python-side foundation work from #3783 (pluggable algorithms, tenant isolation, correctness fixes) and replacing the Python algorithm/backend implementation with a high-performance Rust engine for both `memory` and `redis` backends.

The Rust engine is the **preferred** execution path — all algorithm execution, backend dispatch, result aggregation, and response construction live in Rust when the `rate_limiter_rust` PyO3 extension is installed. Python retains ownership of plugin lifecycle, hook integration, request-context extraction, and config validation. The full Python algorithm and backend implementation is retained as a fallback when the Rust extension is unavailable or when `RATE_LIMITER_FORCE_PYTHON=1` is set.

The Rust engine exposes a high-level `check()` API that reduces the Python-Rust boundary to a single call returning pre-built response dicts. This keeps the existing plugin integration model intact while reducing request-path overhead and preserving shared-counter semantics for multi-instance deployments.

---

### Python foundation

**Gap 1 (HIGH)** — No algorithm choice: only `fixed_window` was available — no way to handle boundary bursts or bursty workloads. Fixed by introducing a strategy pattern with three selectable algorithms (`fixed_window`, `sliding_window`, `token_bucket`) configurable via the `algorithm:` field. All three run on both `memory` and `redis` backends; the Redis backend was extended with atomic Lua scripts for `sliding_window` and `token_bucket` to maintain cluster-wide enforcement across instances.

`test_fixed_window_burst_at_boundary` (xfail) documents the boundary burst as a known trade-off for users who stay on `fixed_window`; `sliding_window` eliminates it entirely.

**Gap 2 (HIGH)** — `by_tenant` cross-throttle: `tenant_id=None` fell back to a shared `"tenant:default"` bucket, causing unrelated users to throttle each other across the entire deployment. Fixed in two places: (1) the plugin now skips the `by_tenant` check entirely when `tenant_id` is `None` instead of inventing a phantom bucket; (2) `mcpgateway/auth.py` now propagates `request.state.team_id` → `global_context.tenant_id` unconditionally via `_propagate_tenant_id()`, called at all four return paths in `get_current_user()`, so single-team API tokens correctly populate tenant context for rate limiting regardless of the `include_user_info` setting.

**Gap 3 (MEDIUM)** — Sliding window memory leak in `sweep()`: stale-but-non-empty timestamp lists were never evicted from `_store` because `sweep()` only removed empty lists. Fixed by embedding the window size in each store key (`"{key}:{window}"`) so staleness is computable at sweep time, and rewriting `sweep()` to evict any key where all timestamps fall outside the window.

### Rust execution engine

**Gap 4 (HIGH)** — Python hot path overhead: every request still paid Python-side rate evaluation costs, including per-dimension orchestration, repeated backend calls, and response metadata construction via individual PyO3 attribute accesses. Fixed by introducing a Rust `RateLimiterEngine` with a single `check()` call per hook invocation. The engine builds dimension keys internally from its pre-parsed config, evaluates all active dimensions in one batch, and returns pre-built Python dicts for HTTP headers and plugin metadata — eliminating ~18-25 PyO3 boundary crossings per request (depending on the number of active dimensions).

**Gap 5 (HIGH)** — Rust acceleration was limited to the in-memory backend: Redis-backed deployments, which are the correctness-critical path for multi-instance enforcement, still relied on the Python backend. Fixed by adding a Redis backend to the Rust engine. Rust now owns the Redis connection and executes batch Lua scripts directly, preserving shared-counter behavior across replicas.

**Gap 6 (MEDIUM)** — Rate-limit computation, response shaping, and plugin policy handling were still interleaved in the Python implementation. Fixed by making the Python wrapper policy-only: it validates config, normalizes context, and invokes the engine. Algorithm execution, backend dispatch, result aggregation, and response dict construction now all live in Rust. The Python hot path is reduced to a single `check()` call plus result dispatch.

**Gap 7 (LOW)** — The Rust+Redis path needed end-to-end proof under a real multi-instance deployment. Validated with the backend-correctness load test against 3 gateways behind nginx with Redis shared state. Result: `60` allowed, `60` rate-limited, `49.6%` blocked, matching the expected `~50%` shared-counter behavior.

---

## Additional hardening

- **`_parse_rate` robustness** — was `rate.split("/")` with no error handling; now uses `maxsplit=1` + `try/except` with a clear message showing the bad input
- **`prompt_pre_fetch` normalisation** — `prompt_id` now normalised with `.strip().lower()`, consistent with how `tool_pre_invoke` handles tool names
- **`_normalised_by_tool` pre-computed** — `by_tool` key lowercasing was re-done on every hook call; now computed once at init in `_validate_config`
- **Unconditional tenant propagation** — `_propagate_tenant_id()` copies `request.state.team_id` → `global_context.tenant_id` at every `get_current_user()` return path, not just inside `_inject_userinfo_instate()` which is gated by `include_user_info` (default `False`)
- **Sliding-window `Retry-After >= 1`** — `int` truncation of `(oldest_ts + window - now)` could produce `0` when the oldest timestamp plus the window rounded down to `int(now)`, telling clients to retry immediately while still over limit; fixed on both memory and Redis paths with `max(1, reset_in)`
- **Token-bucket first-request memory/Redis parity** — the memory path hard-coded `time_to_full = window` on first request, while Redis derived it from `tokens_needed / refill_rate`, causing metadata divergence between backends (e.g. `60` vs `~6` for a `10/m` limit); both paths now use `tokens_needed / refill_rate`
- **Config pre-parsing at startup** — rate strings are parsed once during plugin initialization so the request path does not re-parse limits on every hook invocation
- **Rolling-upgrade compatibility** — the Rust Redis backend uses the same key format as the Python Redis backend so both implementations share counters correctly during mixed deployments
- **Async Redis execution path** — Redis-backed Rust evaluation now uses a multiplexed async Redis connection and an async PyO3 boundary so Python async hooks can await the Redis path directly
- **Rust engine preferred with Python fallback** — when the `rate_limiter_rust` PyO3 extension is installed, the Rust engine handles all rate evaluation; when unavailable or when `RATE_LIMITER_FORCE_PYTHON=1` is set, the plugin falls back to the full Python algorithm and backend implementation
- **`RATE_LIMITER_FORCE_PYTHON` env var** — allows operators to force the Python backend for A/B comparisons or debugging; Makefile targets `benchmark-rate-limiter-capacity-rust` and `benchmark-rate-limiter-capacity-python` exercise both paths
- **Reduced lock contention on memory hot path** — the Rust `MemoryStore` now acquires one read lock on the outer map + one write lock on the per-key state per steady-state request (previously two outer read locks), eliminating a redundant lock cycle
- **Response construction in Rust** — HTTP headers and plugin metadata dicts are built inside the Rust engine after GIL reacquisition, replacing ~18-25 individual PyO3 attribute accesses from the Python side (depending on dimension count)

---

## Architecture

The plugin now delegates all rate evaluation to Rust:

- Python owns plugin lifecycle, hook integration, config validation, and request-context extraction
- Rust owns dimension selection, algorithm execution, backend dispatch, result aggregation, and response dict construction
- Backend selection is an engine concern:
  - `memory` → Rust in-process store via the existing sync engine path
  - `redis` → Rust async Redis execution via a multiplexed connection + batch Lua execution
- When the `rate_limiter_rust` PyO3 extension is installed, Rust handles all rate evaluation; otherwise the plugin falls back to the Python backend

<details>
<summary>Plugin internals: request flow, wrapper responsibilities, Rust engine, backends, and response shaping</summary>

```text
│                         RateLimiterPlugin                           │
│                                                                      │
│  Hooks:  tool_pre_invoke            prompt_pre_fetch                │
│          (tool name + context)      (prompt_id + context)           │
└─────────────────────────────┬────────────────────────────────────────┘
                              │
                              │  Python responsibilities:
                              │  - validate config (at init)
                              │  - normalize by_tool keys (at init)
                              │  - extract user / tenant / tool|prompt
                              ▼
│                     RustRateLimiterEngine wrapper                    │
│                        (PyO3 boundary)                               │
│                                                                      │
│  Python calls the Rust engine once per request:                     │
│  - check(user, tenant, tool, now_unix) for memory                   │
│  - await check_async(user, tenant, tool, now_unix) for redis        │
│                                                                      │
│  Returns: (allowed, headers_dict, meta_dict)                        │
│  Rust builds dimension keys, evaluates, and constructs response     │
│  dicts internally — no per-attribute PyO3 accesses needed.          │
└─────────────────────────────┬────────────────────────────────────────┘
                              │
                              ▼
│                        RateLimiterEngine (Rust)                      │
│                                                                      │
│  Owns:                                                               │
│  - parsed config (by_user, by_tenant, by_tool)                      │
│  - dimension key construction (user:X, tenant:Y, tool:Z)            │
│  - backend selection + dispatch                                      │
│  - algorithm execution                                               │
│  - per-dimension result calculation                                  │
│  - aggregated EvalResult (most restrictive dimension)               │
│  - HTTP header dict construction (X-RateLimit-*, Retry-After)       │
│  - plugin metadata dict construction (remaining, reset_in, dims)    │
└─────────────────────────────┬────────────────────────────────────────┘
                              │
              ┌───────────────┴────────────────┐
              │                                │
              ▼                                ▼
┌───────────────────────────────┐   ┌──────────────────────────────────┐
│      Memory backend (Rust)    │   │       Redis backend (Rust)      │
│                               │   │                                  │
│  in-process counters          │   │  owns multiplexed Redis conn   │
│  fixed_window                 │   │  batch Lua execution            │
│  sliding_window               │   │  shared counters across replicas│
│  token_bucket                 │   │  compatible key format with     │
│  single read-lock hot path    │   │  Python Redis backend           │
└───────────────────────────────┘   └──────────────────────────────────┘
                              │
                              ▼
│                     Python result dispatch                           │
│                                                                      │
│  allowed -> PromptPrehookResult / ToolPreInvokeResult + headers     │
│  blocked -> PluginViolation(429) + headers + meta                   │
│                                                                      │
│  (headers and meta are pre-built dicts from Rust — no conversion)   │
```

</details>

---

## Test results

Validation for this PR is intentionally comprehensive: unit tests, Rust tests, micro-benchmarks, hook-path A/B comparisons, multi-instance Redis correctness, sustained load tests, algorithm comparison, and Fyre VM deployment validation are all included below.

The highest-signal correctness and unit-test summaries remain visible. The more detailed benchmark and deployment sections are preserved in expandables so the data stays in the PR without dominating the main narrative.

### Test results summary

| # | Area | Headline result |
|---|---|---|
| 1 | Backend correctness | `49.6%` blocked vs expected `~50%`, `0` infra failures |
| 2 | Rust micro-benchmarks | `17-21 ns` single-key, `49 ns` three-dimension |
| 3 | Hook-path comparison | `1.7x-1.9x` for most 3-dim memory paths, `28x` for `sliding_window` |
| 4 | Redis capacity benchmark | p99 `60 ms` vs `100 ms`, `28%` less gateway memory |
| 5 | Multi-user scale test | `47.6%` blocked, `0` infra failures across `28,650` requests |
| 6 | Algorithm comparison | All 3 algorithms at `~47-49%` blocked across `~28,600` requests each |
| 7 | Fyre VM validation | Rust engine validated on x86 Fyre |
| 8 | Unit test suite | `93` plugin tests passed, `47/47` Rust tests passed |

### 1. Backend correctness (multi-instance Redis)

Multi-instance Redis correctness matched the expected shared-counter behavior: `49.6%` blocked, `0` infrastructure failures, and an exact `60` allowed / `60` rate-limited split in the single-user validation run.

<details>
<summary>Full backend correctness results</summary>

Topology: `nginx → 3 gateways → shared Redis`, 1 user at 60 req/min against a 30/min limit.

| Metric | Result |
|---|---|
| Total tool calls | `121` |
| Allowed | `60` |
| Rate-limited | `60` |
| Blocked percentage | `49.6%` |
| Infrastructure failures | `0` |
| Avg response time | `49.7 ms` |
| p90 / p99 | `76 ms` / `88 ms` |

**Verdict:** `REDIS BACKEND — limit correctly enforced`. The shared Redis counter produced the expected ~50% blocked rate across three gateway instances. Zero infrastructure failures.

</details>

### 2. Rust micro-benchmarks (criterion, Apple Silicon M-series, release build)

Criterion benchmarks on Apple Silicon show nanosecond-scale in-process memory-backend costs: 17-21 ns for single-key operations, 49 ns for three-dimension evaluation, and 16-67 µs in the contention scenarios included here.

<details>
<summary>Full micro-benchmark results</summary>

In-process memory backend, direct `MemoryStore` calls (no PyO3 overhead):

| Benchmark | Median latency | Notes |
|---|---|---|
| `fixed_window/single_key` | **19 ns** | single dimension, window reset each iteration |
| `token_bucket/single_key` | **21 ns** | |
| `sliding_window/single_key` | **17 ns** | |
| `fixed_window/three_dims` | **49 ns** | user + tenant + tool in one call |
| `fixed_window/hot_counter` | **16 ns** | counter near limit (steady-state path) |
| `fixed_window/blocked_path` | **16 ns** | counter past limit (reject path) |
| `fixed_window/many_keys_10k` | **185 ns** | HashMap with 10,000 distinct keys |
| `fixed_window/concurrent_2t` | **22 µs** | 2 threads via `std::thread::scope` |
| `fixed_window/concurrent_4t` | **35 µs** | 4 threads |
| `fixed_window/concurrent_8t` | **67 µs** | 8 threads |

</details>

### 3. Hook-path comparison (Python pytest-benchmark, memory backend)

Measured per-hook Python→Rust round-trip overhead via `pytest-benchmark`, comparing the Rust engine `check()` path against the older Python-only algorithm+backend stack. Numbers reflect the full hook path including PyO3 crossing, dict construction, and Python result dispatch.

<details>
<summary>Full hook-path comparison results</summary>

All benchmarks: `fixed_window`, `sliding_window`, `token_bucket` × single dimension, three dimensions.

| Config | Python mean | Rust mean | Speedup |
|---|---|---|---|
| `fixed_window` / 1 dim | 5.1 µs | 2.8 µs | 1.8× |
| `fixed_window` / 3 dim | 13.2 µs | 7.2 µs | 1.8× |
| `sliding_window` / 1 dim | 14.3 µs | 2.9 µs | 4.9× |
| `sliding_window` / 3 dim | 81.3 µs | 2.9 µs | **28×** |
| `token_bucket` / 1 dim | 5.6 µs | 2.8 µs | 2.0× |
| `token_bucket` / 3 dim | 14.9 µs | 7.8 µs | 1.9× |

The `sliding_window` improvement is the most dramatic because the Python implementation walks and filters timestamp lists in pure Python on every check, while Rust uses a `VecDeque` with amortized O(1) front pops.

</details>

### 4–8. Remaining validation

<details>
<summary>Redis capacity benchmark, multi-user scale, algorithm comparison, Fyre VM, unit tests</summary>

**4. Redis capacity benchmark** — p99 latency `60 ms` (vs `100 ms` Python), `28%` less gateway RSS. Three gateways, shared Redis, 100 users at 0.25 rps for 5 minutes.

**5. Multi-user scale test** — `47.6%` blocked across `28,650` requests with `0` infrastructure failures. 100 concurrent users, 5-minute sustained load.

**6. Algorithm comparison** — All three algorithms (`fixed_window`, `sliding_window`, `token_bucket`) converged to `~47-49%` blocked across `~28,600` requests each, confirming equivalent enforcement behavior.

**7. Fyre VM validation** — Rust engine validated on x86 Fyre VM deployment.

**8. Unit test suite** — `93` Python plugin tests passed, `47/47` Rust tests passed.

</details>

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "package-lock.json|Cargo.lock|^.secrets.baseline$|scripts/sign_image.sh|scripts/zap|sonar-project.properties|^/Users/brian/dev/github.ibm.com/contextforge-org/sps-pipeline-config/.secrets.baseline$|^./.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-04-02T15:25:23Z",
+  "generated_at": "2026-04-02T15:29:25Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -369,7 +369,8 @@
         "is_verified": false,
         "line_number": 1485,
         "type": "Secret Keyword",
-        "verified_result": null
+        "verified_result": null,
+        "is_secret": false
       },
       {
         "hashed_secret": "9d989e8d27dc9e0ec3389fc855f142c3d40f0c50",
@@ -1096,21 +1097,24 @@
         "is_verified": false,
         "line_number": 125,
         "type": "Secret Keyword",
-        "verified_result": null
+        "verified_result": null,
+        "is_secret": false
       },
       {
         "hashed_secret": "2e0c522bfe4e7885492862df2e0b987c0ca02623",
         "is_verified": false,
         "line_number": 148,
         "type": "Secret Keyword",
-        "verified_result": null
+        "verified_result": null,
+        "is_secret": false
       },
       {
         "hashed_secret": "d9d007c8de197b3f36a3a0ba4f13c0f7df175d5a",
         "is_verified": false,
         "line_number": 312,
         "type": "Secret Keyword",
-        "verified_result": null
+        "verified_result": null,
+        "is_secret": false
       }
     ],
     "docker-compose.yml": [
@@ -2637,7 +2641,8 @@
         "is_verified": false,
         "line_number": 221,
         "type": "Secret Keyword",
-        "verified_result": null
+        "verified_result": null,
+        "is_secret": false
       }
     ],
     "docs/docs/manage/observability/observability.md": [
@@ -5396,7 +5401,8 @@
         "is_verified": false,
         "line_number": 17,
         "type": "Hex High Entropy String",
-        "verified_result": null
+        "verified_result": null,
+        "is_secret": false
       }
     ],
     "mcpgateway/alembic/versions/a8f3b2c1d4e5_add_gateway_refresh_fields.py": [
@@ -6571,14 +6577,16 @@
         "is_verified": false,
         "line_number": 149,
         "type": "Secret Keyword",
-        "verified_result": null
+        "verified_result": null,
+        "is_secret": false
       },
       {
         "hashed_secret": "5f2fe4f03b2314fa3a217f1e1f0ab2e3f5b1b49f",
         "is_verified": false,
         "line_number": 178,
         "type": "Secret Keyword",
-        "verified_result": null
+        "verified_result": null,
+        "is_secret": false
       }
     ],
     "plugins/examples/custom_auth_example/README.md": [
@@ -6831,21 +6839,24 @@
         "is_verified": false,
         "line_number": 932,
         "type": "Secret Keyword",
-        "verified_result": null
+        "verified_result": null,
+        "is_secret": false
       },
       {
         "hashed_secret": "cf743b3a58a4d0f91c1d7f5825c0b1b5f7758174",
         "is_verified": false,
         "line_number": 969,
         "type": "Base64 High Entropy String",
-        "verified_result": null
+        "verified_result": null,
+        "is_secret": false
       },
       {
         "hashed_secret": "ff1a7ebc241b8eefe9e75e13f20cc22c365ab626",
         "is_verified": false,
         "line_number": 969,
         "type": "Secret Keyword",
-        "verified_result": null
+        "verified_result": null,
+        "is_secret": false
       }
     ],
     "plugins_rust/secrets_detection/benches/secrets_detection.rs": [
@@ -10786,7 +10797,8 @@
         "is_verified": false,
         "line_number": 326,
         "type": "Secret Keyword",
-        "verified_result": null
+        "verified_result": null,
+        "is_secret": false
       },
       {
         "hashed_secret": "1fb844731bf1e5928ae606915b54e21264da4768",
@@ -11067,7 +11079,8 @@
         "is_verified": false,
         "line_number": 50,
         "type": "Secret Keyword",
-        "verified_result": null
+        "verified_result": null,
+        "is_secret": false
       }
     ],
     "tests/unit/mcpgateway/utils/test_url_auth.py": [
@@ -11194,35 +11207,40 @@
         "is_verified": false,
         "line_number": 538,
         "type": "Base64 High Entropy String",
-        "verified_result": null
+        "verified_result": null,
+        "is_secret": false
       },
       {
         "hashed_secret": "8e42b03e460b2cf358ffbcf4da3bc5d14a22c86e",
         "is_verified": false,
         "line_number": 582,
         "type": "Base64 High Entropy String",
-        "verified_result": null
+        "verified_result": null,
+        "is_secret": false
       },
       {
         "hashed_secret": "2093dd9cf307518cfe1d2fa5a3985d6fec4e995e",
         "is_verified": false,
         "line_number": 595,
         "type": "Base64 High Entropy String",
-        "verified_result": null
+        "verified_result": null,
+        "is_secret": false
       },
       {
         "hashed_secret": "caa924f200b35ceb6f0e33878faff75203bdccb4",
         "is_verified": false,
         "line_number": 971,
         "type": "Secret Keyword",
-        "verified_result": null
+        "verified_result": null,
+        "is_secret": false
       },
       {
         "hashed_secret": "f16da2820437f3c703ff5b95c813f310ce8e67a4",
         "is_verified": false,
         "line_number": 1288,
         "type": "Secret Keyword",
-        "verified_result": null
+        "verified_result": null,
+        "is_secret": false
       }
     ],
     "tests/unit/plugins/test_jwt_claims_extraction.py": [
@@ -11279,14 +11297,16 @@
         "is_verified": false,
         "line_number": 596,
         "type": "Secret Keyword",
-        "verified_result": null
+        "verified_result": null,
+        "is_secret": false
       },
       {
         "hashed_secret": "d2b1620ca7a5314280e595624e8b13c185f8a2e1",
         "is_verified": false,
         "line_number": 785,
         "type": "Secret Keyword",
-        "verified_result": null
+        "verified_result": null,
+        "is_secret": false
       }
     ],
     "tools_rust/mcp_runtime/tests/runtime.rs": [


mcpgateway/auth.py
```diff
@@ -998,11 +952,12 @@ async def _set_auth_method_from_payload(payload: dict) -> None:
...
             global_context = getattr(request.state, "plugin_global_context", None) if request else None
             if not global_context:
-                # Create global context
+                # Propagate team_id → tenant_id for by_tenant rate limiting
+                team_id = getattr(getattr(request, "state", None), "team_id", None) if request else None
                 global_context = GlobalContext(
                     request_id=request_id,
                     server_id=None,
-                    tenant_id=None,
+                    tenant_id=team_id,
```
lucarlig, line 1196: This propagation only works when `request.state.team_id` is populated, but the session-token path still leaves `team_id` as `None` (`team_id` is only assigned for non-session tokens later in this file). As a result, session-authenticated users still arrive here with `tenant_id=None`, so `by_tenant` limiting is silently skipped for that flow. Please either populate a single resolved team for session auth before this point or document/test that `by_tenant` is intentionally unavailable for session tokens.

lucarlig:
https://docs.rs/pyo3-log/latest/pyo3_log/
please add logs with pyo3_logs

gandhipratik203:
@lucarlig addressing your Rust-side review — all four items have been fixed:

1. Sliding-window `sweep()` now drains stale timestamps and evicts idle keys — `window_nanos` is stored per key so sweep no longer depends on the next access to clean up.
2. Token-bucket blocked-path `Retry-After` now uses ceiling division, matching the Python and Redis paths.
3. Fixed-window keys are evicted at the actual window boundary instead of the hardcoded 1-hour threshold — `window_nanos` is stored per key here as well.
4. Clippy lint fixed — replaced with `.is_multiple_of()`.

gandhipratik203:
> https://docs.rs/pyo3-log/latest/pyo3_log/
> please add logs with pyo3_logs

Added `pyo3-log` with the log bridge initialized at module load. Kept it to warn-level only to avoid adding overhead on the request path.

gandhipratik203:
@ja8zyjits quick context since you already reviewed #3783.

The Python semantics from #3783 remain the baseline here. The main Python-side change in #3809 is the optional Rust execution path in the rate limiter plugin, with Python still owning hook integration, policy, and fallback behavior.

From a Python review perspective, the main deltas are:
- optional Rust engine wiring in `plugins/rate_limiter/rate_limiter.py`
- hook routing to `check()` / `check_async()` when Rust is available
- preserved Python fallback path
- Redis batching / `EVALSHA` support on the Python backend
- expanded validation coverage around parity, fallback, load, and benchmarks

Most of the net-new implementation is under `plugins_rust/rate_limiter/*`, so I do not think this needs a full re-review of the Python feature from scratch. When you have a chance, I'd value a quick pass on the Python integration surface in #3809.

gandhipratik203:
@lucarlig addressing both points from your latest review:

**Reset/retry-after off-by-one (point 2):** Fixed — removed the extra `+1` from `reset_timestamp` and `retry_after` calculations across the Rust memory and Redis backends. The rounding logic now matches the Python implementation, with `max(1, ...)` guards where needed to prevent zero values.

**Sub-second precision (point 1):** The Rust engine uses nanosecond-precision monotonic time internally for all rate decisions, so the rate limiting itself is more precise than the Python path. The `now_unix` value from Python only affects response header timestamps (`X-RateLimit-Reset`, `reset_in`).

Full fractional-second passthrough to the Redis Lua scripts (matching Python's `time.time()` → Lua path exactly) is not included in this PR — the practical difference is at most 1 second on a response header at sub-second boundary edges. I've added this as a documented limitation in the PR description.

plugins_rust/rate_limiter/src/redis_backend.rs
```diff
@@ -0,0 +1,556 @@
...
+    local oldest = redis.call('ZRANGE', KEYS[i], 0, 0, 'WITHSCORES')
+    local oldest_ts = 0
+    if #oldest > 0 then oldest_ts = tonumber(oldest[2]) end
+    if count >= limit then
+        results[i] = {0, count, oldest_ts}
+    else
+        redis.call('ZADD', KEYS[i], now, member)
+        count = count + 1
+        oldest = redis.call('ZRANGE', KEYS[i], 0, 0, 'WITHSCORES')
+        oldest_ts = 0
```
dima-zakharov:
```suggestion
     if count >= limit then
        -- Only fetch oldest when blocked (for Retry-After header)
        local oldest = redis.call('ZRANGE', KEYS[i], 0, 0, 'WITHSCORES')
        local oldest_ts = 0
        if #oldest > 0 then oldest_ts = tonumber(oldest[2]) end
        results[i] = {0, count, oldest_ts}
     else
         redis.call('ZADD', KEYS[i], now, member)
         count = count + 1
         local oldest = redis.call('ZRANGE', KEYS[i], 0, 0, 'WITHSCORES')
         local oldest_ts = 0
```

## Impact
         
| Scenario | Before | After | Improvement |
|----------|--------|-------|-------------|
| **Allowed request** | 2 × `ZRANGE` | 1 × `ZRANGE` | 50% reduction |
| **Blocked request** | 1 × `ZRANGE` | 1 × `ZRANGE` | Same |

**Expected overall savings:** ~25-40% reduction in Redis operations for sliding window algorithm (assuming 50-80% allowed requests in normal traffic).

---



The oldest timestamp is only needed for the `Retry-After` header, which is:
- **Included** when the request is **blocked** (HTTP 429 response)
- **Stripped** when the request is **allowed** (successful response)

By moving `ZRANGE` inside each branch:
- **Blocked path:** Fetch oldest timestamp once (same as before)
- **Allowed path:** Fetch oldest timestamp once (was: twice)

gandhipratik203: Thanks Dima, good catch. Applied — the unconditional `ZRANGE` before the `if/else` is now moved inside each branch so the allowed path only fetches oldest once (after `ZADD`).

jonpspri:
GTG

PR #3853, opened 2026-03-25
fix(docs): improve demo a2a agent docs

### LINKED ISSUE #3709: [DOCS][A2A]: Demo A2A agent quick-start is missing required prerequisites
### 📚 Documentation Issue 

The demo A2A agent quick-start fails out of the box due to three undocumented prerequisites. No errors in the docs themselves, just missing setup steps that are non-obvious.

---

### 📍 Location of the Problem
 https://ibm.github.io/mcp-context-forge/using/agents/a2a/#demo-a2a-agent

---

### ✏️ Type of Issue

- Missing explanation or example
- Unclear instructions

---

### 💡 Suggested Fix or Clarification (optional)
Add a Prerequisites block before the run instructions:

  ### Prerequisites

  **1. Allow localhost in `.env`** (required for local agent registration)

  ```bash
  SSRF_ALLOW_LOCALHOST=true
  ```

  Restart ContextForge after adding this. The default blocks loopback addresses as an SSRF safeguard. This is intentional for production, but must be opted into locally.

  **2. Pass your admin email at runtime**

  The script creates a JWT signed with your instance's secret. The token subject must match a user in the database (typically your `PLATFORM_ADMIN_EMAIL`):

  ```bash
  PLATFORM_ADMIN_EMAIL=you@example.com uv run python scripts/demo_a2a_agent.py
  ```

  Alternatively, generate a token with the project utility and pass it directly. This is the recommended pattern for any scripted access:

  ```bash
  export MCPGATEWAY_BEARER_TOKEN=$(python -m mcpgateway.utils.create_jwt_token \
    --username "$PLATFORM_ADMIN_EMAIL" --exp 60 --secret "$JWT_SECRET_KEY")
  MCPGATEWAY_BEARER_TOKEN=$MCPGATEWAY_BEARER_TOKEN uv run python scripts/demo_a2a_agent.py
  ```

---

Discovered while following the quick-start on macOS with a default .env. The script exits silently on auth failure with no hint about which setting is missing. The three failure modes encountered in order were: missing jti claim (401), wrong token subject (401 "User not found"), and SSRF rejection (422).

Note: troubleshooting & issue writing was AI assisted, human edited.

marekdano:
# 📚 Documentation PR

Closes #3709 

### Summary

Fixes the demo A2A agent to correctly handle ContextForge's request formats and adds missing prerequisites to the documentation.

### Changes

Documentation (`docs/docs/using/agents/a2a.md`)
- Added Prerequisites section documenting required setup:
  - SSRF_ALLOW_LOCALHOST=true requirement with explanation
  - PLATFORM_ADMIN_EMAIL runtime requirement
- Fixed token generation command - removed unnecessary --secret parameter (uses config default)
- Updated API test examples to show correct request format with parameters wrapper
- Consolidated running instructions into clear Option 1 and Option 2
- Fixed Admin UI test steps to include agent registration step
- Fixed MCP tool name from a2a_demo-calculator-agent to a2a-demo-calculator-agent

Demo Agent (`scripts/demo_a2a_agent.py`)
- Fixed request parsing - replaced failing Pydantic models with direct JSON dict parsing
- Added JSONRPC format support for endpoints ending with /
- Added nested message structure handling for Admin UI test button
- Added comprehensive debug logging showing raw request body and extraction path
- Fixed Pydantic models to use Optional[str] = None instead of str = ""
- Added `typing.Optional` import for proper type hints
- Made handler async and added Request import for raw body access

### Testing
Tested with:
    1. Ran all `curl` request listed in the docs
    2. Admin UI test button with queries like `calc: 100/4+25` and `weather: Dallas`
    3. MCP tool invocation via /rpc endpoint

All three methods now work correctly.

### Fixes
- Resolves demo agent returning "Hello" instead of processing queries
- Addresses three failure modes: missing jti claim, wrong token subject, SSRF rejection
- Fixes AttributeError when Admin UI sends nested message structure

### DIFF
--- a/docs/docs/using/agents/a2a.md
+++ b/docs/docs/using/agents/a2a.md
@@ -253,17 +253,16 @@ Before running the demo agent, ensure the following configuration:
 make dev
 
 # Terminal 2: Start the demo agent (auto-registers with ContextForge)
-# Option 1: Let the script generate the token internally
-PLATFORM_ADMIN_EMAIL=admin@example.com uv run python scripts/demo_a2a_agent.py
+# Override the token subject if your admin email differs from the default:
+#   export PLATFORM_ADMIN_EMAIL=you@example.com
+uv run python scripts/demo_a2a_agent.py
 
-# Option 2: Generate token manually for testing API calls (script still generates its own token)
+# Optional: Generate a token for the curl test commands below
 export TOKEN=$(python -m mcpgateway.utils.create_jwt_token \
   --username "admin@example.com" --exp 60)
-uv run python scripts/demo_a2a_agent.py
-# The $TOKEN variable is now available for the curl test commands below
 ```
 
-Note: The token generation uses your configured `JWT_SECRET_KEY` from `.env` (defaults to `my-test-key` for local development).
+Note: The script reads `JWT_SECRET_KEY` and `PLATFORM_ADMIN_EMAIL` from environment variables (defaults: `my-test-key…` and `admin@example.com`).
 
 The demo agent supports these query formats:
 
@@ -345,7 +344,7 @@ curl -X POST "http://localhost:8000/rpc" \
     "jsonrpc": "2.0",
     "method": "tools/call",
     "params": {
-      "name": "a2a-demo-calculator-agent",
+      "name": "a2a_demo-calculator-agent",
       "arguments": {"query": "calc: 99+1"}
     },
     "id": 1
--- a/scripts/demo_a2a_agent.py
+++ b/scripts/demo_a2a_agent.py
@@ -23,8 +23,6 @@ import signal
 import socket
 import sys
 from contextlib import closing
-from typing import Optional
-
 import httpx
 import jwt
 import uvicorn
@@ -123,32 +121,6 @@ app = FastAPI(title="Demo A2A Agent", description="Calculator and Weather Agent 
 agent = SimpleAgent("Demo-A2A-Agent")
 
 
-class Parameters(BaseModel):
-    """Parameters object containing the actual query."""
-
-    query: Optional[str] = None
-    message: Optional[str] = None
-
-
-class A2ARequest(BaseModel):
-    """Request model for A2A protocol format.
-
-    ContextForge sends custom agents requests in this format:
-    {
-        "interaction_type": "admin_test",
-        "parameters": {"query": "weather: Dallas", "message": "..."},
-        "protocol_version": "1.0"
-    }
-    """
-
-    interaction_type: Optional[str] = None
-    parameters: Optional[Parameters] = None
-    protocol_version: Optional[str] = None
-    # Also support direct query/message for simple testing
-    query: Optional[str] = None
-    message: Optional[str] = None
-
-
 class Response(BaseModel):
     """Response model for agent results."""
 
@@ -166,7 +138,10 @@ async def run_agent(request: Request) -> Response:
     """
     # Parse request body
     body = await request.body()
-    body_dict = json.loads(body)
+    try:
+        body_dict = json.loads(body)
+    except (json.JSONDecodeError, ValueError):
+        return Response(response="Error: invalid JSON in request body")
 
     query_text = ""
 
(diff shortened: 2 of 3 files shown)


a-effort:
Good fixes for blockers users may run into when trying to run the demo!
 
 Here are a few cleanup requests / pieces of feedback from 🤖                                    
                                                                                                                             
  ---                                                                                                                        
  What's fixed correctly:                                                                                                    
                                                                                                                             
  - SSRF_ALLOW_LOCALHOST=true prerequisite was a silent blocker. Good fix ✅                                   
  - `--secret` removed from token generation command: good, as it reads from `.env` ✅                                             
  - Tool name typo fixed: `a2a_demo-calculator-agent` ✅                                                                       
  - Admin UI steps now include the agent registration step that was missing ✅                                               
  - Optional `[str] = None` on Pydantic fields is the right fix (empty string "" is falsy, which broke the or chain) ✅

  ---
Changes to consider:

  1. `demo_a2a_agent.py run_agent` rewrite: the `Optional[str] = None` fix already solves the root cause. The raw `json.loads` + manual key-walking approach is harder to maintain than the Pydantic model. Keep the Pydantic path (with Optional) or, if JSONRPC format is genuinely needed, add it as a fallback on top of the Pydantic parse rather than replacing it entirely.
  2. Debug `print()` statements: several `print(f"Raw request body: ...")` / `print(f"Extracted from ...")` lines were left in production code. Move them to `logger.debug()` or remove them.
  3. `import json` inside the handler: should be a top-level import.
  4. `demo_a2a_agent_auth.py` not updated. It has the same `str = ""` issue on Parameters and A2ARequest (lines 337–349). Apply the same `Optional[str] = None` fix so both scripts stay consistent.
  5. Minor doc nit: Option 2 generates a token into `$TOKEN` but the next line, `uv run python scripts/demo_a2a_agent.py`, doesn't use it. Either pass it in `BEARER_TOKEN=$TOKEN` or clarify the token is for the curl test commands, not the script startup.

marekdano:
@a-effort reviewing items have been addressed:                                                                                                                                                                                                
- json import: Moved to top-level (cleaner code structure)
- `print()` statements: Kept as-is for local demo (appropriate for non-production script)                                                                                                                                                        
- `demo_a2a_agent_auth.py`: Applied Optional[str] = None fix for consistency                                                                                                                                                                    
- Documentation: Clarified token usage to avoid confusion

marekdano:
@vishu-bh - reviewed it and removed the print statement with the full raw request body. I replaced the other print statements with the hardcoded strings. 
There is no point in introducing the logging system in this `demo_a2a_agents.py` script since we do not log the sensitive data, and the print statements are already used there only. 
The script is for demo purposes and will never be run on production.

vishu-bh (approved):
Thanks @marekdano LGTM 🚀

crivetimihai:
Thanks for the contribution, @marekdano — the prerequisites section and JSONRPC/nested-message parsing are solid additions.

I rebased, squashed the commits, and made a few review fixes on top:

**Bug fix — MCP tool name prefix**: The docs changed `a2a_demo-calculator-agent` to `a2a-demo-calculator-agent`, but `tool_service.py` uses `f"a2a_{agent.slug}"` (underscore). Reverted to the correct name.

**`PLATFORM_ADMIN_EMAIL` env var**: The docs referenced it but the script hardcoded `admin@example.com`. Updated `create_jwt_token()` to read `os.environ.get("PLATFORM_ADMIN_EMAIL", "admin@example.com")` and clarified the docs to describe environment variables (not `.env` file loading, since the script doesn't use python-dotenv).

**Dead code cleanup**: Removed the now-unused `Parameters` and `A2ARequest` Pydantic models (and the `Optional` import) from `demo_a2a_agent.py` since `run_agent` parses raw JSON directly.

**Error handling**: Added `try/except` around `json.loads(body)` so malformed requests get a clean error instead of a 500.

**Docs simplification**: Consolidated the Option 1/Option 2 pattern into a single clear flow.

**Tests**: Added 23 unit tests (`tests/unit/scripts/test_demo_a2a_agent.py`) covering all request format paths — JSONRPC simple, JSONRPC nested message parts, A2A protocol, simple query/message, empty body, invalid JSON, calculator ops, agent routing, and health endpoint. 100% coverage of the new/changed code.

PR #3854, opened 2026-03-25
chore: pipeline cleanup

crivetimihai:
This change streamlines the CI/CD surface and cleans up related build and documentation paths.

Key updates:
- simplify container scanning and remove legacy scanner-specific install paths and targets
- tighten GitHub Actions workflow permissions and checkout handling
- pin workflow actions and add lint support for custom runner labels
- reduce publish workflow blast radius and separate cache usage by workflow purpose
- remove obsolete workflows and stale references
- clean up docs, Makefile targets, and local dependency guidance to match the current pipeline

This is intended as a repo-wide pipeline cleanup and consistency pass across workflows, build helpers, and supporting documentation.


.github/workflows/docker-scan.yml
```diff
@@ -42,155 +41,71 @@ concurrency:
...
-        if: always() && hashFiles('trivy-results.sarif') != ''
-        uses: github/codeql-action/upload-sarif@v4
-        with:
-          sarif_file: trivy-results.sarif
-          category: trivy
-
-      - name: Install Grype
-        run: |
-          curl -sSfL https://raw.githubusercontent.com/anchore/grype/main/install.sh | sh -s -- -b /usr/local/bin
-
```
brian-hussey, line 174: This shouldn't be removed.

.github/workflows/docker-scan.yml
```diff
@@ -42,155 +41,71 @@ concurrency:
...
-          grype ${{ env.IMAGE_NAME }}:scan --scope all-layers --only-fixed
-
-      - name: Grype SARIF report
-        continue-on-error: true
-        run: |
-          grype ${{ env.IMAGE_NAME }}:scan --scope all-layers --output sarif --file grype-results.sarif
+        uses: anchore/scan-action@1638637db639e0ade3258b51db49a9a137574c3e # v6
+        with:
+          image: ${{ env.IMAGE_NAME }}:scan
+          fail-build: false
```
madhu-mohan-jaishankar: one small clarification - should we have continue-on-error: true and fail-build: false under Grype vulnerability scan. should we not fail the build if vulnerabilities are found?

.github/workflows/docker-scan.yml
madhu-mohan-jaishankar: > This shouldn't be removed.

I guess the `anchore/scan-action`handles installation automatically.

.github/workflows/docker-scan.yml
crivetimihai: Grype was removed from the build, for now.

.github/workflows/docker-scan.yml
crivetimihai: Grype was removed, for now.

PR #3863, opened 2026-03-25
fix: infinite /partial request loop triggered by search input

### LINKED ISSUE #3861: [BUG][UI]: infinite /partial request loop triggered by search input
### 🐞 Bug Summary

The infinite call to `/partial` endpoint is triggered when text is entered into the searchbar input on the gateways, servers, tools, ... pages.

https://github.com/user-attachments/assets/7b77a038-827b-476f-a51e-743f84d61207

 
---

### 🧩 Affected Component

- [ ] `mcpgateway` - API
- [ ] Other (explain below)

---

### 🔁 Steps to Reproduce

1. Go to http://localhost:8080
2. Select one the pages e.g. MCP servers
3. Open the browser console and search for server
4. Observe that calling `/partial` endpoint is called infinitely until the search is cleared
---

### 🤔 Expected Behavior

We should not call `/partial` endpoint infinitely.
We should keep the focus on the search input after `/partial` endpoint response.

---

### 📓 Logs / Error Output

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Version or commit | `latest` |
| Runtime | `Python 3.12` |
| Platform / OS | ` macOS` |
| Container | `Docker` |

---

marekdano:
# 🐛 Bug-fix PR

Closes #3861 

## 📌 Summary

Fix infinite loop of `/partial` endpoint requests triggered when typing in admin UI search bars (servers, tools, resources, gateways, prompts, agents).
Remove unnecessary search input re-initialization from `htmx:afterSwap` and `htmx:afterSettle` handlers, since search inputs live outside the swapped table DOM and persist across partial refreshes.

Fix npm high vulnerabilities

## 🔁 Reproduction Steps

1. Go to http://localhost:8080/
2. Select one the pages e.g. MCP servers
3. Open the browser console and search for server
4. Observe that calling /partial endpoint is called infinitely until the search is cleared

## Root Cause

`initializeSearchInputs()` attached the input event listener before setting `.value` from URL state, so the value assignment immediately fired the listener, queuing a `/partial` reload whose swap re-initialized inputs — creating an infinite cycle

#### Changes

  1. Reorder value assignment and listener attachment in `initializeSearchInputs()` — set `.value` before adding the input listener so initialization doesn't trigger a reload
  2. Remove search input re-initialization from `htmx:afterSwap` — search inputs are not part of the swapped table content; only update filter status
  3. Remove search input re-initialization from `htmx:afterSettle` — same reasoning; only update filter status
  4. Remove eager reload on init — the visible-panel check was queueing an immediate `loadSearchablePanel` call, but the server already returns correctly filtered data in the initial page load
  5. Add regression tests (`tests/js/admin-search-init-loop.test.js`, 7 tests) covering:
    - Value restoration from URL state does not fire the input listener
    - No `htmx.ajax` call is made during initialization
    - User typing still triggers a debounced `/partial` reload (positive test)
    - `htmx:afterSwap` / `htmx:afterSettle` on table targets do not re-trigger search initialization
    - Search input value and listener survive table content swaps
 6. run `npm audit fix` to fix high vulnerabilities

## 🧪 Verification

| Lint suite                            | `make lint`          |     ✅    |
| Unit tests                            | `make test`          |    ✅     |
| Coverage ≥ 80 %                       | `make coverage`      |     ✅    |

## ✅ Checklist

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "package-lock.json|Cargo.lock|^.secrets.baseline$|scripts/sign_image.sh|scripts/zap|sonar-project.properties|^/Users/brian/dev/github.ibm.com/contextforge-org/sps-pipeline-config/.secrets.baseline$|^./.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-03-28T14:11:28Z",
+  "generated_at": "2026-03-28T19:21:52Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -6216,7 +6216,7 @@
         "hashed_secret": "d3ecb0d890368d7659ee54010045b835dacb8efe",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 20818,
+        "line_number": 20821,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -6224,7 +6224,7 @@
         "hashed_secret": "b4e44716dbbf57be3dae2f819d96795a85d06652",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 33882,
+        "line_number": 33885,
         "type": "Secret Keyword",
         "verified_result": null
       }
--- a/mcpgateway/static/admin.js
+++ b/mcpgateway/static/admin.js
@@ -18936,8 +18936,8 @@ function clearSearch(entityType) {
 window.clearSearch = clearSearch;
 
 /**
- * Initialize search inputs for all entity types
- * This function also handles re-initialization after HTMX content loads
+ * Initialize search inputs for all entity types.
+ * Called once on page load; HTMX swap/settle handlers no longer re-invoke this.
  */
 function initializeSearchInputs() {
     console.log("🔍 Initializing search inputs...");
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -28,7 +28,8 @@
     },
     "overrides": {
         "keyv": "5.1.0",
-        "minimatch": ">=10.2.1"
+        "minimatch": ">=10.2.1",
+        "flatted": ">=3.4.2"
```
gcgoncalves: Why are we installing this lib?

marekdano: Because of 4 high severity vulnerabilities

Screenshot 2026-03-26 at 08 58 40

marekdano:
@gcgoncalves 
The code is already using createMemoizedInit correctly.

It already implements the `createMemoizedInit` pattern at line 19009 in admin.js:

```javascript
const {
    init: initializeSearchInputsMemoized,
    debouncedInit: initializeSearchInputsDebounced,
    reset: resetSearchInputsState,
} = createMemoizedInit(initializeSearchInputs, 300, "SearchInputs");
```

## The fix works by:
- Keeping the memoized initialization (already there)
- Removing the problematic `resetSearchInputsState()` + `initializeSearchInputsDebounced()` calls from HTMX event handlers
- Only calling `updateFilterStatus()` on swaps instead of re-initializing

## The key change in the diff shows:

**BEFORE (caused infinite loop):**
```javascript
document.body.addEventListener("htmx:afterSwap", function (event) {
    resetSearchInputsState();  // ❌ This bypassed memoization guards
    initializeSearchInputsDebounced();  // ❌ Triggered reload → swap → repeat
});
```

**AFTER (fixed):**
```javascript
document.body.addEventListener("htmx:afterSwap", function (event) {
    updateFilterStatus();  // ✅ Only updates UI, no re-init
});
```

**No changes needed** - the implementation already follows the suggested pattern.

PR #3937, opened 2026-03-31
fix(ui): harden ensureNoResultsElement() against stale DOM IDs

### LINKED ISSUE #3326: [BUG][UI]: ensureNoResultsElement() global spanId lookup can produce duplicate or orphaned DOM IDs
## Summary

`ensureNoResultsElement()` in `admin.js` resolves the `spanId` parameter via a global `document.getElementById()` before validating that the span actually belongs to the `msg` element. This can return a span not contained in `msg`, or create a duplicate `id` when `msg` is missing but a stale element with `spanId` already exists elsewhere in the DOM.

---

## Affected Code

- `mcpgateway/static/admin.js` – `ensureNoResultsElement()` (lines ~1175–1204)

---

## Current Behavior

1. At line 1177, `span` is resolved globally via `document.getElementById(spanId)`.
2. At line 1178, the function checks if `msg` exists. If it does:
   - If `span` is non-null (found globally), it is returned as-is — **without verifying that `span` is a descendant of `msg`**.
   - If `span` is null, the function falls back to `msg.querySelector("span")` (line 1181).
3. If `msg` is **missing**, the function creates a new `<span id=spanId>` (lines 1194–1195) and a new `<p id=msgId>` — **without checking whether an element with `spanId` already exists in the DOM**.

### Edge case 1: span not inside msg

If `msg` exists in the DOM but the element found by `getElementById(spanId)` belongs to an entirely different part of the page, the function returns `{ msg, span }` where `span` is not a child of `msg`. Callers then write into `span.textContent` expecting to update the visible "no results" message, but the update targets the wrong element.

### Edge case 2: duplicate span ID

If `msg` is absent (e.g., removed during a DOM rebuild) but a stale `<span id=spanId>` still exists somewhere, the create branch produces a second element with the same `id`, violating the HTML spec's uniqueness contract. Subsequent `getElementById(spanId)` calls become non-deterministic.

---

## Expected Behavior

1. When `msg` exists, the function should verify that the globally-resolved `span` is actually a descendant of `msg` before returning it. If not, it should fall back to `msg.querySelector("span")`.
2. When `msg` is missing and the function creates elements, it should check for (and remove or reuse) any existing element with `spanId` to avoid producing duplicate IDs.

---

## Impact

- **Functionality**: In a malformed DOM state, the search query text would be written into the wrong element, and the "no results" message would show with stale or empty query text.
- **DOM correctness**: Duplicate `id` attributes violate the HTML specification and make `getElementById` behavior non-deterministic across browsers.
- **Scope**: Low practical risk. All 8 current call sites use unique ID pairs, and the template does not produce ID collisions under normal operation. The bug would only manifest if DOM manipulation by other code (HTMX partial swaps, browser extensions, etc.) left orphaned elements behind.

---

## Steps to Reproduce

```js
// Edge case 1: span not inside msg
// Setup: msg exists, but a different element has the spanId
const otherDiv = document.createElement("div");
const rogueSpan = document.createElement("span");
rogueSpan.id = "searchQueryTools";
otherDiv.appendChild(rogueSpan);
document.body.appendChild(otherDiv);

// Now call the function — it returns the rogue span, not the one inside msg
const { msg, span } = ensureNoResultsElement(
    "associatedTools", "noToolsMessage", "searchQueryTools", "tool"
);
// span === rogueSpan (not the span inside msg)
console.log(msg.contains(span)); // false
```

```js
// Edge case 2: duplicate spanId
// Setup: msg was removed but its child span was reparented elsewhere
const orphanSpan = document.createElement("span");
orphanSpan.id = "searchQueryTools";
document.body.appendChild(orphanSpan);

// msg is missing, so create branch runs — creates a SECOND span with same id
const { msg, span } = ensureNoResultsElement(
    "associatedTools", "noToolsMessage", "searchQueryTools", "tool"
);
// Now two elements with id="searchQueryTools" exist in the DOM
console.log(document.querySelectorAll("#searchQueryTools").length); // 2
```

**Expected result**: Function returns the correct span inside `msg`, or cleans up stale elements before creating new ones.
**Actual result**: Returns wrong span, or creates duplicate IDs.

---

## Root Cause Analysis

The issue is in `mcpgateway/static/admin.js:ensureNoResultsElement()` at lines 1177–1183 and 1194–1195.

```js
function ensureNoResultsElement(containerId, msgId, spanId, entityLabel) {
    let msg = document.getElementById(msgId);
    let span = document.getElementById(spanId);       // global lookup, no parent check
    if (msg) {
        if (!span) {
            span = msg.querySelector("span");          // fallback only when span is null
        }
        return { msg, span };                          // span may not be inside msg
    }
    // ...
    span = document.createElement("span");
    span.id = spanId;                                  // no check for existing element with this id
    // ...
}
```

The function was introduced in PR #3317 to handle template-caching scenarios where the "no results" message elements might not yet exist in the DOM. The global lookup was a shortcut that works when IDs are globally unique (which they are in the current template), but does not defend against DOM mutations that break that assumption.

---

## Proposed Fix

Scope the span lookup to `msg` and guard the create branch:

```js
function ensureNoResultsElement(containerId, msgId, spanId, entityLabel) {
    let msg = document.getElementById(msgId);
    if (msg) {
        // Scope span lookup to inside msg, not global
        const span = document.getElementById(spanId);
        return { msg, span: (span && msg.contains(span)) ? span : msg.querySelector("span") };
    }
    // Remove any stale element with the target spanId before creating
    const staleSpan = document.getElementById(spanId);
    if (staleSpan) {
        staleSpan.removeAttribute("id");
    }
    const container = document.getElementById(containerId);
    if (!container) {
        return { msg: null, span: null };
    }
    msg = document.createElement("p");
    msg.id = msgId;
    // ... rest unchanged
}
```

---

## Related Issues

- #3314 – [BUG]: Empty styled box appears when search returns zero results in Virtual Server selectors (parent issue)
- PR #3317 – fix(ui): hide selector container on zero search results in Virtual Server modals (introduced the function)

---

## Additional Context

The function is called from 7 sites: once inside `initGatewaySelect`/`applySearch` and once at the top of each of the 6 server-side search functions (`serverSideToolSearch`, `serverSidePromptSearch`, `serverSideResourceSearch`, and their Edit counterparts). All current call sites use unique ID pairs so the bug does not manifest under normal operation. The fix is defensive hardening.

shoummu1:
# Bug-fix PR

## 📌 Summary

`ensureNoResultsElement()` in `mcpgateway/admin_ui/search.js` resolved the `spanId` parameter via a global `document.getElementById()` call without verifying that the found element actually belonged to the expected `msg` parent. This could return a span from an entirely different part of the DOM, causing the "no results" query text to be written into the wrong element. Additionally, when `msg` was absent (e.g., after an HTMX partial re-render), the create branch unconditionally stamped a new `<span id=spanId>` without checking for a pre-existing stale element with the same ID, producing duplicate `id` attributes in the DOM — a violation of the HTML spec that makes subsequent `getElementById` calls non-deterministic.

## 🔗 Related Issue

Closes: #3326

## 🔁 Reproduction Steps

See issue #3326 for full reproduction steps. In brief:

**Edge case 1 — wrong span returned:**
```js
// A rogue span elsewhere in the DOM shares the spanId
const rogueSpan = document.createElement("span");
rogueSpan.id = "searchQueryTools";
document.body.appendChild(rogueSpan);

const { msg, span } = ensureNoResultsElement(
    "associatedTools", "noToolsMessage", "searchQueryTools", "tool"
);
console.log(msg.contains(span)); // false — wrong element returned
```

**Edge case 2 — duplicate IDs:**
```js
// msg is absent but an orphaned span with the same id exists
const orphanSpan = document.createElement("span");
orphanSpan.id = "searchQueryTools";
document.body.appendChild(orphanSpan);

ensureNoResultsElement("associatedTools", "noToolsMessage", "searchQueryTools", "tool");
console.log(document.querySelectorAll("#searchQueryTools").length); // 2 — duplicate id
```

## 🐞 Root Cause

In `mcpgateway/admin_ui/search.js` — `ensureNoResultsElement()`:

1. `document.getElementById(spanId)` was called globally at the top of the function with no subsequent check that the found element was a descendant of `msg`. If another element in the DOM happened to carry the same `id` (e.g., a stale element left by HTMX or a browser extension), that foreign element was returned to callers.

2. In the create branch (when `msg` is absent), `span.id = spanId` was assigned unconditionally without first checking for an existing element with that ID, producing duplicate `id` attributes.

## 💡 Fix Description

Two targeted changes to `ensureNoResultsElement()`:

**Fix 1 — Scope span lookup to `msg`:**
```js
// Before
let span = document.getElementById(spanId); // global, no parent check
if (msg) {
    if (!span) { span = msg.querySelector("span"); }
    return { msg, span }; // span may NOT be inside msg
}

// After
if (msg) {
    const span = document.getElementById(spanId);
    return { msg, span: span && msg.contains(span) ? span : msg.querySelector("span") };
}
```
The `msg.contains(span)` guard ensures the globally-found span is actually a descendant of `msg`. If not, it falls back to `msg.querySelector("span")` — the same safe fallback already used when `span` was null.

**Fix 2 — Strip stale ID before creating new elements:**
```js
// Before (create branch)
span = document.createElement("span");
span.id = spanId; // no duplicate check

// After
const staleSpan = document.getElementById(spanId);
if (staleSpan) { staleSpan.removeAttribute("id"); } // dedup guard
// ... then create fresh span
const span = document.createElement("span");
span.id = spanId;
```
Any orphaned element that already holds `spanId` has its `id` stripped before the new element is created, maintaining the HTML uniqueness invariant.

Two regression tests were added to `tests/js/admin-edit-server-selection.test.js` directly targeting each edge case.

Also corrected `tests/unit/js/gateway.test.js`: an existing test had `searchQueryServers` as a sibling element instead of nested inside `noGatewayMessage`, which didn't match the real template DOM structure. Updated to reflect the actual layout.

## 🧪 Verification

| Check | Command | Status |
|---|---|---|
| Unit tests (JS) | `make test-js` | ✅ |
| Edge case 1: wrong span | New test `falls back to querySelector when globally-found span is not inside msg` | ✅ |
| Edge case 2: duplicate ID | New test `removes stale span id before creating new elements to prevent duplicate IDs` | ✅ |

## ✅ Checklist

### DIFF
--- a/mcpgateway/admin_ui/search.js
+++ b/mcpgateway/admin_ui/search.js
@@ -405,16 +405,16 @@ export const ensureNoResultsElement = function (
   msg.id = msgId;
   msg.className = "text-gray-700 dark:text-gray-300 mt-2";
   msg.style.display = "none";
-  const span = document.createElement("span");
-  span.id = spanId;
+  const newSpan = document.createElement("span");
+  newSpan.id = spanId;
   msg.appendChild(
     document.createTextNode(`No ${entityLabel} found containing \u201C`)
   );
-  msg.appendChild(span);
+  msg.appendChild(newSpan);
   msg.appendChild(document.createTextNode("\u201D"));
   // Insert right after the container
   container.parentNode.insertBefore(msg, container.nextSibling);
-  return { msg, span };
+  return { msg, span: newSpan };
 };
 
 /**
--- a/mcpgateway/admin_ui/search.js
+++ b/mcpgateway/admin_ui/search.js
@@ -386,13 +386,15 @@ export const ensureNoResultsElement = function (
   entityLabel
 ) {
   let msg = document.getElementById(msgId);
-  let span = document.getElementById(spanId);
   if (msg) {
-    // Element already in the DOM – just return references
-    if (!span) {
-      span = msg.querySelector("span");
-    }
-    return { msg, span };
+    // Scope span lookup to inside msg — guard against global ID collisions
+    const span = document.getElementById(spanId);
+    return { msg, span: span && msg.contains(span) ? span : msg.querySelector("span") };
+  }
+  // Remove any stale element holding spanId before creating a new one
+  const staleSpan = document.getElementById(spanId);
+  if (staleSpan) {
+    staleSpan.removeAttribute("id");
   }
   // Create the message element dynamically
   const container = document.getElementById(containerId);
@@ -403,7 +405,7 @@ export const ensureNoResultsElement = function (
   msg.id = msgId;
   msg.className = "text-gray-700 dark:text-gray-300 mt-2";
   msg.style.display = "none";
-  span = document.createElement("span");
+  const span = document.createElement("span");
   span.id = spanId;
   msg.appendChild(
     document.createTextNode(`No ${entityLabel} found containing \u201C`)
--- a/tests/js/admin-edit-server-selection.test.js
+++ b/tests/js/admin-edit-server-selection.test.js
@@ -2233,6 +2233,59 @@ describe("ensureNoResultsElement", () => {
     expect(result.msg).toBeNull();
     expect(result.span).toBeNull();
   });
+
+  // Edge case: span found globally but not a descendant of msg — should fall back to querySelector
+  test("falls back to querySelector when globally-found span is not inside msg", () => {
+    // msg exists in the DOM with its own inner span (no id)
+    const container = makeContainer("ctr3326a");
+    const msg = doc.createElement("p");
+    msg.id = "noMsg3326a";
+    const innerSpan = doc.createElement("span");
+    msg.appendChild(innerSpan);
+    container.parentNode.insertBefore(msg, container.nextSibling);
+
+    // A rogue span elsewhere in the DOM has the same id
+    const rogueSpan = doc.createElement("span");
+    rogueSpan.id = "querySpan3326a";
+    doc.body.appendChild(rogueSpan);
+
+    const result = win.ensureNoResultsElement(
+      "ctr3326a",
+      "noMsg3326a",
+      "querySpan3326a",
+      "item",
+    );
+
+    // Should return the inner span, not the rogue one
+    expect(result.msg).toBe(msg);
+    expect(result.span).toBe(innerSpan);
+    expect(result.span).not.toBe(rogueSpan);
+    expect(msg.contains(result.span)).toBe(true);
+  });
+
+  // Edge case: msg absent but a stale span with the same id exists — must strip it to prevent duplicate IDs
+  test("removes stale span id before creating new elements to prevent duplicate IDs", () => {
+    const container = makeContainer("ctr3326b");
+
+    // Orphaned span with the target id already in the DOM (msg is absent)
+    const orphanSpan = doc.createElement("span");
+    orphanSpan.id = "querySpan3326b";
+    doc.body.appendChild(orphanSpan);
+
+    const result = win.ensureNoResultsElement(
+      "ctr3326b",
+      "noMsg3326b",
+      "querySpan3326b",
+      "tool",
+    );
+
+    // The newly created span should have the id
+    expect(result.span.id).toBe("querySpan3326b");
+    // The orphan must have had its id stripped to avoid duplicates
+    expect(orphanSpan.id).toBe("");
+    // Only one element with that id should exist
+    expect(doc.querySelectorAll("#querySpan3326b").length).toBe(1);
+  });
 });
 
 // ---------------------------------------------------------------------------
--- a/tests/unit/js/gateway.test.js
+++ b/tests/unit/js/gateway.test.js
@@ -1337,8 +1337,7 @@ describe("initGatewaySelect - extended", () => {
       </div>
       <div id="gw-pills"></div>
       <div id="gw-warn"></div>
-      <div id="noGatewayMessage" style="display:none"></div>
-      <span id="searchQueryServers"></span>
+      <p id="noGatewayMessage" style="display:none">No MCP server found containing "<span id="searchQueryServers"></span>"</p>
     `;
 
     initGatewaySelect("gw-select", "gw-pills", "gw-warn", 12, null, null, "searchGateways");


marekdano:
@shoummu1 - thanks for the contribution!
Can you please resolve conflicts?

shoummu1:
Thanks, @marekdano. I’ve rebased the branch and resolved all conflicts. The PR is now ready for review.

marekdano (dismissed):
### Summary of Changes

#### Problem
`ensureNoResultsElement()` had two DOM edge cases:
1. Global `getElementById()` could return a span from a different part of the DOM
2. Creating new elements without checking for stale IDs could produce duplicate IDs

#### Solution
1. Added `msg.contains(span)` guard to verify span is inside msg before returning it
2. Strip `id` attribute from any stale element before creating new span

Both fixes are minimal, targeted, and maintain backward compatibility.

This is a well-crafted fix that:
1. Correctly addresses both reported edge cases
2. Includes comprehensive regression tests
3. Maintains existing functionality
4. Follows project coding standards
5. Has minimal risk and no breaking changes

LGTM 🚀

mcpgateway/admin_ui/search.js
```diff
@@ -402,7 +404,7 @@ export const ensureNoResultsElement = function (
   msg.id = msgId;
   msg.className = "text-gray-700 dark:text-gray-300 mt-2";
   msg.style.display = "none";
-  span = document.createElement("span");
+  const span = document.createElement("span");
```
gcgoncalves: This is a redeclaration of an existing const (from line 390), no? Maybe make the previous one a let, or use a different var name here.

gcgoncalves (approved):

marekdano (approved):
This PR hardens `ensureNoResultsElement()` against DOM ID scoping bugs: when multiple search widgets share the same page, a globally-found spanId might belong to a different widget's msg, or exist as an orphaned stale element. 

The fix scopes span lookups to inside msg (with a `msg.contains()` guard) and strips stale span IDs before creating new elements. The `gateway.test.js` fixture is also corrected to match the actual DOM structure the function produces.

LGTM 🚀

PR #3961, opened 2026-03-31
fix(rbac): unblock self-service public team join for non-admin users

### LINKED ISSUE #3960: [BUG]: Self-service public team join blocked by RBAC token-scoping guard
### Bug Summary

Non-admin users cannot join public teams via `POST /teams/{team_id}/join` or the Admin UI join-request flow. The RBAC permission service discards global roles (including `platform_viewer.teams.join`) before they can be evaluated, creating a chicken-and-egg problem where users need team membership to get permission to request team membership.

---

### Affected Component

- [x] `mcpgateway` - API

---

### Steps to Reproduce

1. Create a non-admin user via `POST /auth/email/admin/users` (auto-assigns `platform_viewer` global role with `teams.join`)
2. Create a public team
3. Authenticate as the non-admin user with a session token
4. Call `POST /teams/{team_id}/join` with `{"message": "Please let me join"}`
5. **Result**: 403 "Access denied"

All token types produce the same result:

| Token type | Result | Why |
|---|---|---|
| Session token | 403 | `resolve_session_teams()` returns current memberships only — target team not in scope |
| JWT with `teams: null` (non-admin) | 403 | Normalizes to `token_teams=[]` (public-only) — L516 guard fires |
| JWT with `teams: [current_teams]` | 403 | Target team not in list — L516 guard fires |
| JWT with `teams: [target_team]` | 403 | Auth middleware rejects: "User is no longer a member of the associated team" |

Only admin tokens (`token_teams=None`) can reach the endpoint.

---

### Expected Behavior

A non-admin user with the `teams.join` permission (via `platform_viewer` global role) should be able to request to join a public team. The endpoint's own business logic already enforces:
- Target team exists (`teams.py:921-922`)
- Target team is public (`teams.py:924`)
- Caller is not already a member (`teams.py:927-928`)

---

### Root Cause

`PermissionService._get_user_roles()` at `mcpgateway/services/permission_service.py:513-521`:

```python
if team_id:
    if token_teams is not None and (len(token_teams) == 0 or team_id not in token_teams):
        return []  # ← EARLY RETURN: discards ALL roles including global/personal
```

When `team_id` is present (from the URL path) and the target team is not in `token_teams`, the function returns `[]` **immediately** — before querying global or personal roles. The user's `platform_viewer` global role (which includes `teams.join`) is never checked.

This guard was added in commit `49a1a486c` to enforce strict token-scoping isolation. The security intent ("public-only tokens must never access team-specific roles") is valid, but the implementation is too broad: it also blocks global roles that are not team-specific.

### Why `POST /teams/invitations/{token}/accept` Is Not Affected

The accept endpoint has no `team_id` in its URL path — only `token`. So `kwargs.get("team_id")` returns `None`, the `if team_id:` branch is skipped, and global roles are queried normally. This is incidental (URL design), not deliberate.

---

### Affected Endpoints

| Endpoint | Permission | Decorator | Verified |
|----------|-----------|-----------|----------|
| `POST /teams/{team_id}/join` | `teams.join` | `@require_permission("teams.join")` | Yes (403) |
| `POST /admin/teams/{team_id}/join-request` | `teams.join` | `@require_permission("teams.join", allow_admin_bypass=False)` | Yes (403) |
| `DELETE /admin/teams/{team_id}/join-request/{id}` | `teams.join` | `@require_permission("teams.join", allow_admin_bypass=False)` | Yes (403) |

---

### Verified: Other Team-Scoped Operations Work Correctly

| Scenario | Token Type | Result | Correct? |
|----------|-----------|--------|----------|
| Leave own team | Session token | 200 | Yes |
| Leave own team | Scoped JWT `teams=[team]` | 200 | Yes |
| Access own team | Session token | 200 | Yes |
| Access other team | Session token | 403 | Yes (isolation) |
| Cross-team `GET /tools?team_id=X` | Session token | 403 | Yes (matches unit test contract) |

The explicit-team isolation model is correct for all other endpoints. Only self-service team-entry flows are broken.

---

### Recommended Fix

**Do NOT apply a blanket fix to `_get_user_roles()`.** A generic "pass-through for global roles" would also allow `GET /tools?team_id=other_team` to pass RBAC via global `tools.read`, contradicting the explicit team isolation model and existing unit tests (`test_permission_service_token_narrowing.py`, `test_rbac_cross_team_isolation.py` — 50 tests pass).

**Apply a narrow, route-specific fix.** Add a new parameter to `require_permission()` (current signature: `(permission, resource_type=None, allow_admin_bypass=True)`) that tells the decorator to evaluate the permission globally without binding to the request's `team_id`:

```python
@require_permission("teams.join", check_globally=True)
```

This would pass `team_id=None` to `check_permission()`, skipping the L516 guard while still requiring the `teams.join` permission. Users without `teams.join` in any role are still denied.

**What must NOT change:**
- The `return []` guard at L516 for all other team-scoped endpoints
- Auth-layer token membership validation
- `POST /teams/invitations/{token}/accept` (works as-is)
- The `@require_permission` decorator must remain on these endpoints — removing it entirely would bypass RBAC

---

### Related Test Failures

The following Playwright tests also fail but are **test bugs, not code bugs**:

| Test | Issue | Fix |
|------|-------|-----|
| `test_member_leaves_team` | Uses public-only JWT (`teams=null`) instead of session token. Leave works correctly with session tokens. | Use `token_use="session"` in test |
| `test_cross_team_tool_not_visible` | Expects 200 + empty results for out-of-scope `team_id`, but current contract is 403 (matches unit tests) | Expect 403 or test via non-team-scoped request |

crivetimihai:
## Summary

- Add `check_globally` parameter to `require_permission()` decorator to evaluate permissions against global/personal roles without binding to the request's `team_id`
- Apply to team-entry endpoints (`join`, `leave`, `join-request`, `cancel-join-request`) where users need `teams.join` but aren't yet members of the target team
- Add team_id validation to admin cancel-join-request to prevent cross-team cancellation
- Fix two Playwright test bugs (wrong token type, wrong assertion)

Closes #3960

## Test plan

- [ ] Unit tests: `pytest tests/unit/mcpgateway/middleware/test_rbac.py tests/unit/mcpgateway/services/test_permission_service_token_narrowing.py tests/unit/mcpgateway/services/test_rbac_cross_team_isolation.py -v` (335 pass)
- [ ] Full unit suite: 15,274 pass, 0 fail
- [ ] Pylint parallel: `make pylint` → 10.00/10
- [ ] Playwright: `test_request_to_join_public_team`, `test_member_leaves_team`, `test_cross_team_tool_not_visible`
- [ ] Manual: non-admin session token → `POST /teams/{public_team}/join` → 200

### DIFF
--- a/mcpgateway/admin.py
+++ b/mcpgateway/admin.py
@@ -72,7 +72,7 @@ from mcpgateway.common.models import LogLevel
 from mcpgateway.common.validators import SecurityValidator
 from mcpgateway.config import settings, UI_HIDABLE_HEADER_ITEMS, UI_HIDABLE_SECTIONS, UI_HIDE_SECTION_ALIASES
 from mcpgateway.db import A2AAgent as DbA2AAgent
-from mcpgateway.db import EmailApiToken, EmailTeam, extract_json_field
+from mcpgateway.db import EmailApiToken, EmailTeam, EmailTeamJoinRequest, extract_json_field
 from mcpgateway.db import Gateway as DbGateway
 from mcpgateway.db import get_db, GlobalConfig, ObservabilitySavedQuery, ObservabilitySpan, ObservabilityTrace
 from mcpgateway.db import Prompt as DbPrompt
@@ -6765,7 +6765,7 @@ async def admin_remove_team_member(
 
 
 @admin_router.post("/teams/{team_id}/leave")
-@require_permission("teams.join", allow_admin_bypass=False)  # Users who can join can also leave
+@require_permission("teams.join", allow_admin_bypass=False, check_globally=True)  # Users who can join can also leave
 async def admin_leave_team(
     team_id: str,
     request: Request,  # pylint: disable=unused-argument
@@ -6838,7 +6838,7 @@ async def admin_leave_team(
 
 
 @admin_router.post("/teams/{team_id}/join-request")
-@require_permission("teams.join", allow_admin_bypass=False)
+@require_permission("teams.join", allow_admin_bypass=False, check_globally=True)
 async def admin_create_join_request(
     team_id: str,
     request: Request,
@@ -6927,7 +6927,7 @@ async def admin_create_join_request(
 
 
 @admin_router.delete("/teams/{team_id}/join-request/{request_id}")
-@require_permission("teams.join", allow_admin_bypass=False)
+@require_permission("teams.join", allow_admin_bypass=False, check_globally=True)
 async def admin_cancel_join_request(
     team_id: str,
     request_id: str,
@@ -6951,6 +6951,11 @@ async def admin_cancel_join_request(
     try:
         team_service = TeamManagementService(db)
         user_email = get_user_email(user)
+
+        # Verify the join request belongs to this team before cancelling
+        join_request = db.query(EmailTeamJoinRequest).filter(EmailTeamJoinRequest.id == request_id, EmailTeamJoinRequest.team_id == team_id).first()
+        if not join_request:
+            return HTMLResponse(content='<div class="text-red-500">Join request not found for this team</div>', status_code=404)
 
         # Cancel the join request
         success = await team_service.cancel_join_request(request_id, user_email)
--- a/mcpgateway/middleware/rbac.py
+++ b/mcpgateway/middleware/rbac.py
@@ -553,7 +553,7 @@ def _is_mutate_permission(permission: str) -> bool:
     return parts[-1] in _MUTATE_PERMISSION_ACTIONS if len(parts) >= 2 else False
 
 
-def require_permission(permission: str, resource_type: Optional[str] = None, allow_admin_bypass: bool = True):
+def require_permission(permission: str, resource_type: Optional[str] = None, allow_admin_bypass: bool = True, check_globally: bool = False):
     """Decorator to require specific permission for accessing an endpoint.
 
     Args:
@@ -562,6 +562,10 @@ def require_permission(permission: str, resource_type: Optional[str] = None, all
         allow_admin_bypass: If True (default), admin users bypass all permission checks.
                            If False, even admins must have explicit permissions.
                            Use False for admin UI routes to enforce granular RBAC.
+        check_globally: If True, evaluate the permission against global/personal roles
+                        only, ignoring the request's team_id. Use for team-entry endpoints
+                        (e.g. join, leave) where the user needs a global permission but
+                        the URL contains a team_id they are not yet a member of.
 
     Returns:
         Callable: Decorated function that enforces the permission requirement
@@ -641,6 +645,13 @@ def require_permission(permission: str, resource_type: Optional[str] = None, all
                 # verify_team_for_user, token team membership checks).
                 if not team_id:
                     check_any_team = True
+
+            # check_globally: evaluate permission against global/personal roles only,
+            # ignoring the request's team_id.  Used for team-entry endpoints where the
+            # caller needs teams.join but is not yet a member of the target team.
+            if check_globally:
+                team_id = None
+                check_any_team = False
 
             # First, check if any plugins want to handle permission checking
             # First-Party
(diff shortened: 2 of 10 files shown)


gandhipratik203 (commented):
## Review Notes

### Findings

**1. Medium: `check_globally` drops the route `team_id` before the plugin permission hook runs, which regresses team-aware `HTTP_AUTH_CHECK_PERMISSION` plugins on these endpoints.**

`mcpgateway/middleware/rbac.py:627`, `mcpgateway/middleware/rbac.py:657`

Why it matters: the new behavior is correct for built-in RBAC evaluation, but the original target-team context is also the input to plugin authorization. With the current change, plugins receive `team_id=None` for `join`, `leave`, and join-request endpoints, even when the route clearly targets a specific team.

Evidence: `HttpAuthCheckPermissionPayload` explicitly models `team_id` as hook context (`mcpgateway/plugins/framework/hooks/http.py:143`), and the docs describe it as "Team context (if applicable)" (`docs/docs/using/plugins/http-auth-hooks.md:251`).

Suggested fix: preserve the original route `team_id` for plugin hook/audit payloads, and only null it for the internal `PermissionService.check_permission(...)` call.

**2. Medium: the new cross-team cancel guard is missing the deny-path regression test that this security-sensitive fix needs.**

`mcpgateway/admin.py:6894`

Why it matters: this PR adds an important check that prevents canceling a join request with a mismatched `team_id`/`request_id` pair, but there is no unit test asserting the new wrong-team path returns `404`.

Evidence: existing cancel tests cover success/failure/exception cases, but not "valid request id, wrong team id" (`tests/unit/mcpgateway/test_admin.py:16798`, `tests/unit/mcpgateway/test_admin_module.py:1179`).

Suggested fix: add a unit test that submits a real `request_id` under the wrong `team_id` and asserts `404`.

**3. Low: the new Playwright JWT helper uses `token_use: str = None`, which should be typed as `Optional[str] = None`.**

`tests/playwright/teams/conftest.py:29`

Why it matters: this is only a typing correctness issue, not a runtime bug.

---

### Notes from verification

- The core RBAC intent looks correct. For built-in RBAC, `check_globally=True` does unblock public-team entry flows by avoiding token-team binding on the permission check.
- Verified the default-role path: non-admin users are expected to have global `platform_viewer`, which includes `teams.join`.
- The updated Playwright expectation for cross-team tool access returning `403` matches the current behavior.

### Validation

Reviewed against the PR head in an isolated checkout. Ran targeted unit suites around the changed areas successfully:
- `tests/unit/mcpgateway/middleware/test_rbac.py -k check_globally`
- `tests/unit/mcpgateway/test_admin.py -k 'cancel_join_request or create_join_request or leave_team'`
- `tests/unit/mcpgateway/routers/test_teams.py -k 'request_to_join_team or leave_team'`
- `tests/unit/mcpgateway/services/test_permission_service_token_narrowing.py`
- `tests/unit/mcpgateway/services/test_rbac_cross_team_isolation.py`
- `tests/unit/mcpgateway/routers/test_teams_coverage.py`
- `tests/unit/mcpgateway/routers/test_teams_v2.py`

madhav165 (commented):
Building on gandhipratik203's Finding 1 — I think the plugin context regression points to a deeper issue worth discussing.

The `check_globally` parameter is working around a permission check that may not need to exist on `join`, `join-request`, and `cancel-join-request` in the first place. `teams.join` is present in every built-in role including `platform_viewer`, which every authenticated user receives globally — so this check can never meaningfully fail for a legitimately authenticated user.

The real authorization on these endpoints is already in the endpoint body and service: `join` and `join-request` validate team visibility and existing membership directly, and `cancel_join_request` in `TeamManagementService` filters by user_email, enforcing ownership.

The suggested change:
- Remove `@require_permission` from `join`, `join-request`, and `cancel-join-request` (three endpoints across two files).
- Keep the decorator on `leave` but revert `check_globally=True` from it — the user is already a member when leaving, so the original team scoping works correctly there and the flag was unnecessary.
- This lets us revert `check_globally` from the decorator entirely, the `**kwargs` additions to the mock decorators, and the three new unit tests.

Finding 1 is resolved at the root rather than by threading route context separately through the plugin payload.

The cross-team cancellation fix in `admin_cancel_join_request` is a genuine improvement and worth keeping regardless.

PR #3754, opened 2026-03-20
feat: replace the internal plugin framework with CPEX

araujof:
## Summary

Replace the internal plugin framework (`mcpgateway/plugins/framework/`, `mcpgateway/plugins/tools/`, `plugin_templates/`) with the [CPEX](https://github.com/contextforge-org/contextforge-plugins-framework) external package (`cpex>=0.1.0rc1`).

Closes: #3753 

## What changed

- **Added `cpex` as external dependency** in `pyproject.toml`
- **Migrated all imports**: `from mcpgateway.plugins.framework` → `from cpex.framework`, `from mcpgateway.plugins.tools` → `from cpex.tools`
- **Moved `get_plugin_manager()`** singleton factory to `mcpgateway/plugins/__init__.py` (bridges cpex with gateway-specific hook policies)
- **Migrated PluginMode enums**: `ENFORCE` → `SEQUENTIAL`, `PERMISSIVE` → `TRANSFORM`, `ENFORCE_IGNORE_ERROR` → `SEQUENTIAL` + `on_error: ignore` (see https://github.com/IBM/mcp-context-forge/pull/3754#issuecomment-4360555203)
- **Adapted to cpex API changes**:
  - `PluginViolation.http_status_code` / `.http_headers` → `violation.details["http_status_code"]` / `details["http_headers"]`
  - `PromptPosthookPayload.name` → `.prompt_id`; `ToolPreInvokePayload.arguments` → `.args`
- **Deleted internal framework**: `mcpgateway/plugins/framework/`, `mcpgateway/plugins/tools/`, `plugin_templates/` directory, `tests/unit/mcpgateway/plugins/framework/`
- **Created acceptance tests** (`tests/acceptance/plugins/test_cpex_contract.py`) verifying cpex API surface, behavioral contracts, serialization, and settings compatibility
- **Updated YAML configs** across plugins and test fixtures
- **Updated documentation** under `docs/docs/using/plugins/`, `plugins/README.md`, `plugins/AGENTS.md`, `llms/plugins-llms.md`

## Test results

- Unit tests pass
- Acceptance tests pass
- Performance profiling shows no regression
- flake8, bandit, pylint all clean

PR #3965, opened 2026-04-01
refactor(plugins): migrate in-tree plugins to PyPI packages (cpex-*)

jonpspri:
## Summary

- Migrate 6 in-tree plugins (pii_filter, secrets_detection, url_reputation, retry_with_backoff, encoded_exfil_detection, rate_limiter) to standalone PyPI packages (`cpex-*`)
- Delete all `plugins_rust/` Rust plugin implementations — plugins are now distributed as pre-built wheels from PyPI
- Remove Rust plugin builder stages from all Containerfiles (keep `tools_rust/mcp_runtime`)
- Remove ~100 lines of `rust-*` plugin Makefile targets
- Add `[plugins]` extra to `install-dev` and CI pytest workflow so cpex packages are installed
- Update `tool_service.py` import from `plugins.retry_with_backoff` to `cpex_retry_with_backoff`
- Update plugin `kind` paths across 7+ doc files from `plugins.pii_filter.pii_filter.PIIFilterPlugin` to `cpex_pii_filter.PIIFilterPlugin`
- Clean up pre-commit configs, CODEOWNERS, MANIFEST.in, whitesource, .gitignore

## Test plan

- [ ] `make install-dev && make test` passes (cpex packages installed via `[plugins]` extra)
- [ ] CI pytest workflow installs plugins via `--extra plugins`
- [ ] Container builds succeed without `plugins_rust/` directory
- [ ] Plugin loading test (`test_all_plugins_load_together`) passes with correct count

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -9,6 +9,7 @@ if [ "$(uname -m)" = "s390x" ]; then
 fi
 
 HTTP_SERVER="${HTTP_SERVER:-gunicorn}"
+APP_ROOT="${APP_ROOT:-/app}"
 RUST_MCP_MODE="${RUST_MCP_MODE:-off}"
 RUST_MCP_LOG="${RUST_MCP_LOG:-warn}"
 RUST_MCP_SESSION_AUTH_REUSE="${RUST_MCP_SESSION_AUTH_REUSE:-}"
@@ -47,7 +48,8 @@ RUST_MCP_PID=""
 SERVER_PID=""
 
 apply_rust_mcp_mode_defaults() {
-    local normalized_mode="${RUST_MCP_MODE,,}"
+    local normalized_mode
+    normalized_mode="$(printf '%s' "${RUST_MCP_MODE}" | tr '[:upper:]' '[:lower:]')"
     local runtime_enabled_default="false"
     local managed_default="true"
     local session_core_default="false"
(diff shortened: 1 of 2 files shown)


lucarlig:
The import works fine on my side and I can see the plugin.

the author field, I think @gandhipratik203 is the author for the Rust version we’re uploading, although we might want to keep it as “Contextforge developers” for consistency.

One last thing: should we consider keeping the Python plugin as an example somewhere instead of removing it entirely?

gandhipratik203:
Hi @jonpspri @crivetimihai — thanks for putting this together. This is a good move — distributing plugins as pre-compiled PyPI packages is a well-established pattern (ruff, orjson, pydantic-core all do the same thing) and makes Rust acceleration accessible without requiring a toolchain on the gateway.

  I'd like to help this move forward:

  1. ADR — I can put together an ADR for the plugin extraction pattern since I have context on both the in-tree and package approaches. Would
  cover the reasoning, pros/cons, effects on the build pipeline, testing strategy, and fallback behavior.
  2. Test coverage — Happy to add integration tests, load tests, and algorithm-level validation to the cpex-plugins repo to make sure the
  package has production-grade coverage.

araujof:
@jonpspri I don't anticipate issues, but I think we should coordinate this PR with #3754.

lucarlig:
Agreed with @gandhipratik203 on the plan:

**Step 1**  
Move all plugins out of `cpex-plugins` (keeping plugin tests here) and **freeze plugin code** in that repo.

**Step 2**  
Figure out the testing strategy and implement it (probably just clone contextforge repo in CI), then **unfreeze plugin development**.

For the old Python plugins, @jonpspri will spin up an example repo.

I’ll also add an ADR to document this.

**Reasoning:**  
Right now, changes in `cpex-plugins` can unintentionally break CI in `mcp-context-forge`. Freezing the plugin code and decoupling it ensures that **random or unrelated changes don’t cause CI failures in MCP**, making builds more stable and predictable.

@brian.hussey @jonpspri — does this approach work?

lucarlig:
@gandhipratik203 Short follow-up: I kept and updated the integration coverage for the packaged cpex_* plugins, fixed the shipped config/import drift, added gateway-side smoke coverage for the packaged plugin wiring, aligned the external package version metadata, and added ADR-048 to document the extraction order and repo split rationale. I also removed the gateway runtime dependency on cpex_retry_with_backoff internals and added targeted tests for the new retry parsing paths. Local full pytest, diff-cover, and doctests all passed after the follow-up fixes.

PR #3967, opened 2026-04-01
fix: resolve Alpine.js MutationObserver race condition in observability sub-views

### LINKED ISSUE #3966: fix: Observability sub-views (Tools, Metrics, Prompts, Resources) fail to load — Alpine.js MutationObserver race condition
## Bug Description

The Observability dashboard sub-views (MCP Tools, Advanced Metrics, Prompts, Resources) fail to load with JavaScript errors:

```
Uncaught ReferenceError: createToolsController is not defined
Uncaught ReferenceError: createMetricsController is not defined
```

The Traces tab works fine (server-rendered via HTMX). Only the dynamically-loaded sub-views are affected.

## Reproduction Steps

1. Enable observability (`OBSERVABILITY_ENABLED=true`, `OTEL_ENABLE_OBSERVABILITY=true`)
2. Generate some trace data (invoke tools, etc.)
3. Navigate to Admin → Observability
4. Click on "MCP Tools" or "Advanced Metrics" tab
5. The view is empty; browser console shows `createToolsController is not defined`

## Root Cause

When `loadToolsView()` / `loadMetricsView()` fetches a sub-partial (e.g. `observability_tools.html`), the response contains:
1. A `<script>` block defining `window.createToolsController` (~600 lines)
2. A `<div x-data="createToolsController()" x-init="init()">` element

The current code inserts the fetched HTML into the DOM via `container.innerHTML = ...`. At this point, Alpine.js's global MutationObserver detects the new `x-data` attribute and **immediately** tries to initialize the component by calling `createToolsController()`. However, the `<script>` tag has not yet been executed (innerHTML does not execute scripts), so the function is undefined.

Previous attempts to fix this:
- Parsing in a detached `div` + `appendChild(scriptEl)` → failed (MutationObserver still fires on `container.innerHTML = temp.innerHTML`)
- `Alpine.mutateDom()` to suppress observer → failed (scripts still not executing via `appendChild`)
- `(0, eval)(script.textContent)` inside Alpine x-data → failed (runs through Alpine's `new Function()` evaluator, not native browser scope)

## Proposed Fix

Move script extraction logic to a **native `<script>` block** (executed by the browser, not by Alpine's `new Function()` evaluator):

1. Define `window.__obsExecAndStrip(html)` in the main `<script>` tag of `observability_partial.html`
2. Use `new RegExp()` with string concatenation to extract `<script>` blocks (avoids `</script>` literal breaking HTML parsing)
3. Execute extracted code via `document.createElement('script').text = code; document.head.appendChild(s)` (jQuery's globalEval pattern — synchronous, global scope)
4. Return clean HTML (scripts stripped) for DOM insertion
5. Insert clean HTML inside `Alpine.mutateDom()` (suppresses observer), then call `Alpine.initTree(container)`

This ensures controller functions are defined in the global scope **before** Alpine sees the `x-data` attributes.

## Environment

- ContextForge 1.0.0-RC2 (commit from `main` as of late March 2026)
- Alpine.js 3.15.8 (CDN)
- HTMX
- PostgreSQL backend
- Browser: Chrome/Firefox (both affected)

ecthelion77:
# 🐛 Bug-fix PR

Before opening this PR please:

1. `make lint`            - N/A (HTML template change only, no Python modified)
2. `make test`            - N/A (no unit tests for frontend templates)
3. `make coverage`        - N/A
4. `make docker docker-run-ssl` or `make podman podman-run-ssl` - ✅ Tested with Docker
5. Update relevant documentation. - N/A (no doc changes needed)
6. Tested with sqlite and postgres + redis. - ✅ Tested with PostgreSQL + Redis
7. Manual regression no longer fails. Ensure the UI and /version work correctly. - ✅ Verified

---

## 📌 Summary

Fixes the observability dashboard sub-views (MCP Tools, Advanced Metrics, Prompts, Resources) which fail to load with `createToolsController is not defined` / `createMetricsController is not defined` errors. The Traces tab (server-rendered via HTMX) is unaffected.

## 🔁 Reproduction Steps

Fixes #3966

1. Enable observability (`OBSERVABILITY_ENABLED=true`, `OTEL_ENABLE_OBSERVABILITY=true`)
2. Generate trace data (invoke tools via MCP)
3. Navigate to **Admin → Observability**
4. Click **MCP Tools** or **Advanced Metrics** tab
5. Sub-view is empty; browser console shows `createToolsController is not defined`

## 🐞 Root Cause

When `loadToolsView()` fetches a sub-partial (e.g. `observability_tools.html`), the response contains:
1. A `<script>` block defining `window.createToolsController` (~600 lines)
2. A `<div x-data="createToolsController()" x-init="init()">` element

The code inserts HTML via `container.innerHTML`. Alpine.js's global `MutationObserver` detects the new `x-data` attribute and **immediately** calls `createToolsController()` — but `innerHTML` does not execute `<script>` tags, so the function is undefined.

## 💡 Fix Description

Move script extraction to a **native `<script>` block** in `observability_partial.html` (browser-executed, not Alpine's `new Function()` evaluator):

1. **`window.__obsExecAndStrip(html)`**: extracts `<script>` blocks from raw HTML via regex (`new RegExp` with string concatenation to avoid `</script>` literal breaking HTML parsing)
2. **Execution**: `document.createElement('script').text = code; document.head.appendChild(s)` — jQuery's globalEval pattern, synchronous in global scope
3. **DOM insertion**: clean (script-free) HTML inserted inside `Alpine.mutateDom()` to suppress MutationObserver, then `Alpine.initTree()` called explicitly

This ensures controller functions are defined **before** Alpine sees the `x-data` attributes.

## 🧪 Verification

| Lint suite                            | `make lint`          | N/A — HTML template only |
| Unit tests                            | `make test`          | N/A — no frontend tests |
| Coverage ≥ 80 %                       | `make coverage`      | N/A |
| Manual regression no longer fails     | Tested in browser    | ✅ All 4 sub-views load correctly |

**Manual verification**: After deploying, all 4 observability sub-views (Tools, Metrics, Prompts, Resources) load their charts and data correctly. No console errors. Tested on Chrome and Firefox.

## ✅ Checklist
- [x] Code formatted (`make black isort pre-commit`) — N/A (HTML only)

## 📓 Notes

Only `mcpgateway/templates/observability_partial.html` is modified. The change is -69/+51 lines — it simplifies the 4 loader functions by extracting the shared script-execution logic into a single `window.__obsExecAndStrip()` helper.

### DIFF
--- a/mcpgateway/templates/observability_partial.html
+++ b/mcpgateway/templates/observability_partial.html
@@ -24,7 +24,7 @@ window.__obsExecAndStrip = function(html) {
   while ((m = scriptRe.exec(html)) !== null) {
     var attrs = m[1];
     var code = m[2].trim();
-    if (/\bsrc\s*=/i.test(attrs)) {
+    if (/(?:^|\s)src\s*=/i.test(attrs)) {
       console.warn('[obs] external <script src> not supported in dynamic partials, skipped');
       continue;
     }
--- a/tests/unit/js/observability-exec-strip.test.js
+++ b/tests/unit/js/observability-exec-strip.test.js
@@ -31,7 +31,7 @@ function defineObsExecAndStrip(window) {
     while ((m = scriptRe.exec(html)) !== null) {
       var attrs = m[1];
       var code = m[2].trim();
-      if (/\bsrc\s*=/i.test(attrs)) {
+      if (/(?:^|\s)src\s*=/i.test(attrs)) {
         console.warn(
           "[obs] external <script src> not supported in dynamic partials, skipped",
         );
@@ -194,6 +194,20 @@ describe("__obsExecAndStrip", () => {
     const clean = win.__obsExecAndStrip(html);
 
     expect(warnSpy).toHaveBeenCalledTimes(1);
+    expect(clean).toBe("<div>ok</div>");
+    warnSpy.mockRestore();
+  });
+
+  test("does not false-positive on data-src attribute", () => {
+    const warnSpy = vi.spyOn(console, "warn").mockImplementation(() => {});
+    const html =
+      '<script data-src="config">window.__dataSrcRan = true;</script><div>ok</div>';
+
+    const clean = win.__obsExecAndStrip(html);
+
+    // data-src is NOT an external script — it should execute normally
+    expect(warnSpy).not.toHaveBeenCalled();
+    expect(win.__dataSrcRan).toBe(true);
     expect(clean).toBe("<div>ok</div>");
     warnSpy.mockRestore();
   });


crivetimihai:
Nice fix, @ecthelion77 — good root-cause analysis on the MutationObserver race. A few things I added during review:

**Rebased onto main** (clean, no conflicts).

**Hardened the Alpine guard** — the rest of the codebase (`initialization.js:956`, `:1505`) checks `typeof mutateDom === 'function'` and `typeof initTree === 'function'` before calling. Updated all 4 loader blocks to match, with a plain innerHTML fallback when Alpine has a partial API.

**Added external script detection** — the old code handled `<script src="...">` via `querySelectorAll`, but the new regex-based approach would silently skip external scripts. None of the sub-partials use them today, but the helper now detects `src=` in attributes and logs a `console.warn` instead of silently dropping them. Also fixed the initial regex (`\bsrc\s*=`) which would false-positive on `data-src=` attributes — changed to `(?:^|\s)src\s*=`.

**Added 19 vitest/jsdom unit tests** covering:
- Script extraction, execution, ordering, and stripping
- Scripts with attributes (`defer`), empty scripts, case-insensitive tags
- Global scope execution, error recovery (catch + continue), DOM cleanup
- External `<script src>` warning (and `data-src` non-regression)
- Alpine.mutateDom + initTree integration, non-Alpine fallback
- Partial Alpine API fallback (mutateDom missing, initTree missing, non-function values)
- Full end-to-end pipeline simulation matching the real controller pattern

**Live-validated** against a running instance — confirmed all 4 sub-partial endpoints (`tools`, `metrics`, `prompts`, `resources`) return `<script defer>` with inline controllers, no external scripts.

All 2001 JS unit tests pass (44 files).

crivetimihai (approved):
Looks good. The race condition fix is correct — scripts execute synchronously before Alpine sees the DOM, and mutateDom suppresses the MutationObserver during insertion. Rebased, hardened, tested, and live-validated. Approving.

PR #3979, opened 2026-04-01
feat(plugins): add multi-tenancy support with per-server plugin managers

### LINKED ISSUE #3801: [EPIC][PLUGINS]: Support for Plugins Multi-tenancy and dynamic loading
### 🧭 Type of Feature

---

### 🧭 Epic

**Title:** Dynamic Multi-Tenant Plugin Management
**Goal:** Allow plugin configuration to be changed at runtime without restarting the application, and allow different teams to have different plugin configurations.
**Why now:** ContextForge already supports teams and multi-tenancy for tools, resources, and servers. Plugins are the last major subsystem that is still global and static — all teams share the same plugins, and any configuration change requires a full restart. As teams adopt ContextForge for production workloads, the inability to manage plugins per-team or update them without downtime is a blocker.

---

### 🧑🏻‍💻 User Story 1 — Hot Reload Without Restart

**As a:** platform operator
**I want:** to add, remove, or reconfigure a plugin without restarting ContextForge
**So that:** plugin changes do not interrupt service or require a maintenance window

#### ✅ Acceptance Criteria
```gherkin
Scenario: Add a new plugin at runtime
  Given ContextForge is running with a set of active plugins
  When an operator adds a new plugin configuration via the Admin UI or API
  Then the new plugin becomes active within seconds
  And existing in-flight requests are not interrupted

Scenario: Remove a plugin at runtime
  Given a plugin is currently active
  When an operator removes the plugin configuration
  Then the plugin stops being invoked for new requests
  And requests that were already in-flight complete normally

Scenario: Update a plugin's configuration at runtime
  Given a plugin is active with a given configuration
  When an operator updates the plugin's settings (e.g. threshold, mode)
  Then subsequent requests use the new configuration
  And no restart is required
```

---

### 🧑🏻‍💻 User Story 2 — Per-Team Plugin Configuration

**As a:** platform administrator managing multiple teams
**I want:** each team to have its own plugin configuration
**So that:** teams can have different security policies, compliance plugins, and custom integrations without interfering with each other

#### ✅ Acceptance Criteria
```gherkin
Scenario: Team-specific plugin is active only for that team
  Given a global set of plugins and a team-specific plugin for "team-finance"
  When "team-finance" invokes a tool
  Then the team-specific plugin (e.g. pii_scanner) is invoked
  When another team invokes a tool
  Then pii_scanner is not invoked for that team

Scenario: Team overrides a global plugin's configuration
  Given a global audit_logger plugin with priority 100
  And team "finance" has configured audit_logger with priority 50 and mode "audit"
  When a "finance" team tool is invoked
  Then the team's audit_logger configuration is used, not the global one

Scenario: Team disables a global plugin
  Given a global injection_scanner plugin
  And team "development" has disabled injection_scanner for their team
  When a "development" team tool is invoked
  Then injection_scanner is not invoked for that request
```

---

### 🧑🏻‍💻 User Story 3 — Backward Compatibility

**As a:** existing ContextForge operator using YAML-based plugin configuration
**I want:** my existing setup to continue working without changes
**So that:** I can adopt the new capabilities incrementally without being forced to migrate

#### ✅ Acceptance Criteria
```gherkin
Scenario: Existing YAML config still works
  Given an operator has a plugins/config.yaml file
  When ContextForge starts
  Then plugins are loaded from YAML as before
  And no database or API configuration is required

Scenario: Single-tenant operators are unaffected
  Given ContextForge is deployed without multi-tenancy
  When tools are invoked
  Then plugin behavior is identical to the current implementation
```

---

```mermaid
flowchart TD
    A[Admin UI / API] -->|CRUD plugin config| B[(Database)]
    B -->|change event| C[TenantPluginManager]
    C -->|team-finance| D[PluginManager finance]
    C -->|team-ops| E[PluginManager ops]
    C -->|global| F[PluginManager global]
    D --> G[pii_scanner + audit_logger]
    E --> H[audit_logger]
    F --> I[injection_scanner + audit_logger]
```

Config resolution: global plugins are the baseline. A team can add plugins, override a global plugin's settings, or disable a global plugin. The team config takes precedence where both define the same plugin.

---

### 🔗 MCP Standards Check

---

### 🔄 Alternatives Considered

- **Restart-based reload**: operators can already reload plugins by restarting. This has downtime and prevents per-team customization.
- **Single shared PluginManager**: simpler, but makes multi-tenancy impossible without significant state entanglement.
- **User-level plugins**: possible, but a user needing custom plugins can be modeled as a team-of-one. Deferred to avoid per-user instance explosion.

---

### 📓 Additional Context

Related subsystems: `plugins/`, `mcpgateway/services/`, team/RBAC model in `mcpgateway/auth.py`.

The plugin framework (`cpex/`) currently loads config from a static YAML file into a singleton PluginManager at startup. Both limitations (static config and shared-global scope) need to be addressed together since they share the same root cause: the framework couples config loading to manager instantiation.

cafalchio:
## 🔗 Related Issue
Closes #3801

---

## 📝 Summary

Previously the gateway had a single global `PluginManager` singleton initialized at startup.
This PR replaces it with a `TenantPluginManagerFactory` that creates and caches isolated
`TenantPluginManager` instances per virtual server, enabling each tenant to have independent
plugin configurations, priorities, and modes.

### Architecture

```
Before
──────
startup
  └─ TenantPluginManager (singleton, Borg shared state)
       └─ all requests share one config

After
─────
get_plugin_manager(context_id)              # async, called per-request
  └─ TenantPluginManagerFactory (global singleton)
       ├─ _managers["__global__"]          # global / fallback 
       ├─ _managers["server-abc"]          # per-tenant 
       └─ _managers["server-xyz"]
             └─ merged config = base YAML ⊕ tenant overrides
```

**Key flows:**

- `get_plugin_manager(context_id)` — lazy-creates and caches a manager for `context_id`.
  Uses a double-checked lock pattern: fast path is a plain dict read; slow path holds
  the `asyncio.Lock` only during the final cache insert, so concurrent lookups for
  different servers never block each other.
- `shutdown_plugin_manager_factory()` — called from lifespan teardown; shuts down all
  cached managers and clears the factory reference.
- `TenantPluginManager._merge_tenant_config(context_id)` — overlays per-server plugin
  overrides (config, mode, priority) from `_tenant_configs` onto the base YAML config.
- `BaseService._get_plugin_manager(context_id)` — single entry point for all services
  (tool, resource, prompt); falls back to an injected `_plugin_manager` mock in tests.

### Changed components

| File | Change |
|------|--------|
| `plugins/framework/manager.py` | Add `TenantPluginManager`, `TenantPluginManagerFactory`; `_DEFAULT_CONTEXT_ID` sentinel; fix Borg init ordering bug |
| `plugins/framework/__init__.py` | Replace `_plugin_manager` singleton + `set_plugin_manager` with async `get_plugin_manager(context_id)` and `shutdown_plugin_manager_factory` |
| `mcpgateway/main.py` | Remove startup-time singleton init; resolve manager lazily at request time; call `shutdown_plugin_manager_factory()` on teardown |
| `mcpgateway/services/base_service.py` | Add `_get_plugin_manager(context_id)` shared helper |
| `mcpgateway/services/{tool,resource,prompt}_service.py` | Use `_get_plugin_manager` instead of module-level import |
| `mcpgateway/middleware/rbac.py` | Minor import cleanup |
| `plugins/config.yaml` | Updated plugin config |
| `tests/` | Updated to inject `_plugin_manager` mock via instance attribute |

---

## 🏷️ Type of Change
- [ ] Bug fix
- [x] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

---

## ✅ Checklist

---

## 📓 Notes

Backward compatibility:
    The global ``__global__`` context uses ``PluginManager`` with ``singleton=True``
    (Borg shared state) to preserve backward-compatible behaviour for deployments that
    do not use virtual-server tenancy.  Per-server managers created by
    ``TenantPluginManagerFactory`` use ``TenantPluginManager`` (Borg disabled) so each
    has fully independent state.

Concurrency safety:
    ``TenantPluginManagerFactory.get_manager`` uses a task-based double-checked pattern.
    The fast path (dict read inside the lock) is safe in asyncio's single-threaded model.
    The slow path creates a single ``asyncio.Task`` per ``context_id`` and stores it in
    ``_inflight`` while the lock is held, so only one build task is ever started for a
    given context.  Concurrent callers waiting on the same ``context_id`` all await the
    same task.  Once the task completes, ``_build_manager`` stores the result in
    ``_managers`` under the lock, and ``get_manager`` re-checks the cache before
    returning so that any replacement triggered by ``reload_tenant`` is honoured.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-04-07T08:48:54Z",
+  "generated_at": "2026-04-07T09:02:54Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -19558,7 +19558,7 @@
         "hashed_secret": "65882b5e8dbab0e649474b2a626c1d24e1b317f5",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 82,
+        "line_number": 83,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -19566,7 +19566,7 @@
         "hashed_secret": "35e6ca489e73a8991165cdcfc2b79eb0d25ed0a3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 103,
+        "line_number": 105,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -19574,7 +19574,7 @@
         "hashed_secret": "59df6723fb54fd8dde6dcf51034e51491bfbff24",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 909,
+        "line_number": 947,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -19582,7 +19582,7 @@
         "hashed_secret": "9b4cd28c8ecd18d533601bbb7d3345b9d007bd80",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 915,
+        "line_number": 955,
         "type": "Secret Keyword",
         "verified_result": null
       }
@@ -19592,7 +19592,7 @@
         "hashed_secret": "6f865e7e20c773e3483eaa74a07d03cec02bde2e",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 181,
+        "line_number": 159,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -19600,7 +19600,7 @@
         "hashed_secret": "d58f8c42247f1cf100bf63f196d93d7d302e8ca9",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 202,
+        "line_number": 180,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -19608,7 +19608,7 @@
         "hashed_secret": "3acfb2c2b433c0ea7ff107e33df91b18e52f960f",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 272,
+        "line_number": 250,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -19616,7 +19616,7 @@
         "hashed_secret": "e48c47f1431afd3bb030deea266b4309e2228c5b",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 284,
+        "line_number": 262,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -19624,7 +19624,7 @@
         "hashed_secret": "4155f3a6f21409ec6066b6c974b058d3ea4a2c5c",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 466,
+        "line_number": 450,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -19632,7 +19632,7 @@
         "hashed_secret": "84a36e47e8bfcca13318bca53cdcb1270e2c4b9e",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 487,
+        "line_number": 472,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -19640,7 +19640,7 @@
         "hashed_secret": "c48ea9182291ca3f527914fd7d35a698830931d0",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 525,
+        "line_number": 512,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -19648,7 +19648,7 @@
         "hashed_secret": "534c8ebac109fde87e1eb0c6e35249fc8a38278a",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 681,
+        "line_number": 674,
         "type": "Secret Keyword",
         "verified_result": null
       }
@@ -22719,14 +22719,6 @@
         "verified_result": null
       },
       {
-        "hashed_secret": "83719d2d8284b30dee2823349c2d7f600e237d8c",
-        "is_secret": false,
-        "is_verified": false,
-        "line_number": 83,
-        "type": "Secret Keyword",
-        "verified_result": null
-      },
-      {
         "hashed_secret": "d9ea801e3f4d913d0b6721b819f2b90c35f2303b",
         "is_secret": false,
         "is_verified": false,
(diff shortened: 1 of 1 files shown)


mcpgateway/plugins/framework/manager.py
```diff
@@ -948,3 +952,196 @@ async def invoke_hook_for_plugin(
         if not isinstance(payload, PluginPayload):
             raise ValueError(f"When payload_as_json=False, payload must be a PluginPayload, got {type(payload)}")
         return await self._get_executor().execute_plugin(hook_ref, payload, context, violations_as_exceptions)
+
+
+class DBPluginManager(PluginManager):
```
araujof: Suggestion: rename it as `TenantPluginManager` or similar to better reflect the intent/semantics/use.

mcpgateway/plugins/framework/manager.py
```diff
@@ -948,3 +952,196 @@ async def invoke_hook_for_plugin(
...
+        if singleton:
+            self._yaml_path = yaml_path
+            self._server_id = server_id
+            # Only initialize _tenant_configs if not already in shared state
+            if not hasattr(self, "_tenant_configs"):
+                self._tenant_configs = tenant_configs or {}
+
+        # Apply tenant override immediately if server_id provided
+        if server_id:
+            self._config = self._merge_tenant_config(server_id)
```
araujof: I think we should guard concurrent writes to the borg state to protect against race conditions. Check the constructor of `PluginManager` to see how we did it there.

mcpgateway/plugins/framework/manager.py
```diff
@@ -948,3 +952,196 @@ async def invoke_hook_for_plugin(
...
+            server_id: The server identifier for tenant-specific configuration.
+            plugin_configs: List of plugin configuration dictionaries to override.
+        """
+        self._tenant_configs[server_id] = plugin_configs
+        # If this manager is for this server_id, update the config
+        if self._server_id == server_id:
+            self._config = self._merge_tenant_config(server_id)
+
+
+class DBPluginManagerFactory:
```
araujof: If you change the name of `DBPluginManagerFactory`, change it here too for consistency.

cafalchio:
Closing as it was moved to #4068

dima-zakharov:
The `make benchmark-mcp-tools` test gives the same results as main branch, but one should be aware that this benchmark runs without plugins, except for retry backoff.

PR #3987, opened 2026-04-02
fix(bootstrap): rename orphaned resources to fix team and name assignment conflict

### LINKED ISSUE #3986: [BUG]: bootstrap_db fails with UniqueViolation when assigning orphaned resources to admin team
## Description

During startup, `bootstrap_db` assigns orphaned resources (where `team_id IS NULL` or `owner_email IS NULL`) to the platform admin team. This fails with a `UniqueViolation` when a resource with the same name already exists under the admin team.

## Affected Tables

All tables with `uq_team_owner_name*` constraints:
- `servers` — `uq_team_owner_name_server`
- `tools` — `uq_team_owner_email_name_tool`
- `prompts` — `uq_team_owner_gateway_name_prompt`
- `resources` — `uq_team_owner_gateway_uri_resource`
- `gateways` — `uq_team_owner_slug_gateway`
- `a2a_agents` — `uq_team_owner_slug_a2a_agent`

## Root Cause

Two scenarios trigger the conflict:

1. A row already owned by `admin@example.com` exists with name "X", and another orphaned row (`team_id=NULL`) also has name "X" — bootstrap tries to assign the orphan to admin → collision.
2. Two orphaned rows from different users share the same name — bootstrap assigns the first successfully, then the second collides.

**Code location:** `mcpgateway/bootstrap_db.py` lines 590–592

## Error

```
psycopg.errors.UniqueViolation: duplicate key value violates unique constraint "uq_team_owner_name_server"
DETAIL: Key (team_id, owner_email, name)=(5b8d262d..., admin@example.com, Web Search) already exists.
```

## Proposed Fix

On `UniqueViolation`, append an integer suffix to the resource name (e.g. `Web Search_2`) and retry. This:
- Preserves visibility — downstream apps are not broken
- Retains admin team ownership
- Does not affect tool routing since resolution is by ID, not name

madhav165:
## 📌 Summary

`bootstrap_resource_assignments` bulk-assigns orphaned resources (NULL `team_id`/`owner_email`/`visibility`) to the platform admin team on startup. It fails with `UniqueViolation` on `uq_team_owner_name_*` constraints when an orphan shares a name with a resource already in the admin team, or when two orphans share the same name.

## 🔁 Reproduction Steps

Closes #3986

1. Deploy with PostgreSQL where resources exist from a pre-multi-tenant migration (NULL `team_id`)
2. Some resources share a name with resources already assigned to the admin team
3. On startup, `bootstrap_resource_assignments` throws `UniqueViolation` and skips the affected resource type entirely

## 🐞 Root Cause

`mcpgateway/bootstrap_db.py` — the orphan-assignment loop set `team_id`/`owner_email` on all orphans in bulk then called `db.commit()`. If any row caused a unique constraint violation on `(team_id, owner_email, name)`, the entire batch for that resource type was silently skipped.

Two conflict sources:
1. Orphan name matches a resource already assigned to the admin team in the DB
2. Two orphans share the same name — the second conflicts after the first is committed

Affects all 6 resource types: `servers`, `tools`, `resources`, `prompts`, `gateways`, `a2a_agents`.

## 💡 Fix Description

Before assigning, issue one query per resource type to fetch all names already taken in the admin team that match any orphan value exactly or as a `value-%` suffixed variant. Track intra-batch claimed names in a `batch_assigned` set (orphans not yet committed are invisible to the DB query). On conflict, parse numeric suffixes from the taken set, compute `max + 1`, and rename to `original-N` before commit. Visibility and ownership are preserved.

## 🧪 Verification

| Check | Command | Status |
|---|---|---|
| Unit tests | `uv run pytest tests/unit/mcpgateway/test_bootstrap_db.py` | ✅ 52 passed |
| Coverage ≥ 80 % | `make coverage` | |
| Manual regression no longer fails | deploy against PG with duplicate-named orphans | |

## ✅ Checklist
- [x] Code formatted (`make autoflake isort black`)


crivetimihai:
## Review changes

Rebased onto `main` and added two commits on top of the original work:

### 1. `fix(bootstrap): sanitize log messages and add null-field branch test`

- The `logger.warning()` for name conflicts was not using `SecurityValidator.sanitize_log_message()` on the resource name/value strings, unlike the adjacent error handler. Fixed for consistency and log injection safety.
- Added `test_null_field_value_skips_conflict_detection` to cover the `original_value is None` branch — orphans with NULL unique-field values should be assigned without conflict detection.

### 2. `refactor(bootstrap): hoist _like_safe helper and document gateway_id tradeoff`

- Moved `_like_safe()` above the resource-type loop since it's a pure function with no closures — no need to redefine it each iteration.
- Added an inline comment documenting the intentional omission of `gateway_id` from the `existing_taken` query: for Resource/Prompt models whose uniqueness includes `gateway_id`, the `(team_id, owner_email)`-only filter is conservative (may over-rename) but can never miss a real `UniqueViolation`. That is the correct tradeoff for one-time bootstrap code.

### Coverage

53 tests passing, 98% coverage (only pre-existing advisory lock retry lines uncovered). All new/modified lines and branches have full coverage.

mcpgateway/bootstrap_db.py
```diff
@@ -579,23 +580,87 @@ async def bootstrap_resource_assignments(conn: Connection) -> None:
...
+                    # conservative — for Resource/Prompt models whose uniqueness also depends
+                    # on gateway_id, this may produce unnecessary renames but can never miss a
+                    # real conflict. That is the correct tradeoff for one-time bootstrap code.
+                    existing_taken: set[str] = (
+                        {
+                            row[0]
+                            for row in db.query(field_col).filter(
+                                resource_model.team_id == personal_team.id,
+                                resource_model.owner_email == admin_user.email,
+                                or_(*[cond for v in original_values for cond in (field_col == v, field_col.like(f"{_like_safe(v)}-%", escape="\\"))]),
```
gcgoncalves, line 632: The `_like_safe()` function escapes `LIKE` wildcards, but it may not fully protect against SQL injection if `v` contains SQL special chars beyond `LIKE` wildcards. Use SQLAlchemy's `literal()` or ensure full parameterization.

crivetimihai: This is already fully parameterized. SQLAlchemy's `field_col.like(pattern)` generates `column LIKE :param_1` — the pattern string is passed as a bind parameter, never interpolated into the SQL statement. SQL injection through bind parameter values is not possible.

`_like_safe()` escapes LIKE-specific metacharacters (`%`, `_`, `\`) so the **pattern matching** works correctly (e.g., a resource named `tool_v2` matches literally, not as `tool[any char]v2`). It is not an SQL injection defense — SQLAlchemy's parameter binding already provides that.

Using `literal()` here would be redundant since SQLAlchemy already wraps the value as a `BindParameter` internally when passed to `.like()` or `==`.

PR #3993, opened 2026-04-02
fix: include resource_metadata in all 401 WWW-Authenticate responses

### LINKED ISSUE #3990: [BUG]: OAuth 401 responses missing resource_metadata in WWW-Authenticate when server requires OAuth
### 🐞 Bug Summary

When a request arrives without an OAuth token and the backend MCP server requires OAuth authentication, the 401 response's `WWW-Authenticate` header is missing the `resource_metadata` attribute. This breaks MCP OAuth discovery for clients that rely on this attribute to locate the resource metadata endpoint (per the MCP OAuth specification).

The root cause is a code ordering issue in `_auth_no_token()`: the strict-mode and empty-bearer checks return 401 **before** `_check_server_oauth_enforcement()` is called, so the `resource_metadata` URL is never included in the response.

---

### 🧩 Affected Component

---

### 🔁 Steps to Reproduce

1. Configure a gateway with a backend MCP server that has OAuth enabled
2. Send a request to a tool on that server **without** an `Authorization` header
3. Observe the 401 response's `WWW-Authenticate` header
4. The header contains `Bearer realm="mcp-gateway"` but is missing the `resource_metadata` attribute

---

### 🤔 Expected Behavior

The 401 `WWW-Authenticate` header should include the `resource_metadata` attribute pointing to the server's OAuth resource metadata endpoint, e.g.:

```
Bearer realm="mcp-gateway", resource_metadata="https://gateway.example.com/.well-known/oauth-protected-resource/server-name"
```

This allows MCP clients to discover the OAuth configuration and initiate the authorization flow.

---

### 📓 Logs / Error Output

No crash — the gateway returns 401 but with an incomplete `WWW-Authenticate` header. Clients that follow the MCP OAuth spec cannot discover the resource metadata and fail to authenticate.

---

### 🧠 Environment Info

| Key | Value |
| --- | --- |
| Version or commit | `1.0.0-RC2` |
| Container | Docker |

---

### 🧩 Additional Context

The fix reorders `_auth_no_token()` in `streamablehttp_transport.py`:
1. Call `_check_server_oauth_enforcement()` **first** to determine if the server requires OAuth
2. Build the `www_auth` string with the `resource_metadata` attribute when applicable
3. Use the enriched `www_auth` in all subsequent 401 responses (strict-mode check, empty-bearer check)

This ensures every 401 response includes the full OAuth discovery information regardless of which check triggers the rejection.

ecthelion77:
# 🐛 Bug-fix PR

## 📌 Summary

When a client sends an empty `Authorization: Bearer` header to an OAuth-enabled server, the 401 response's `WWW-Authenticate` header is missing the `resource_metadata` attribute (RFC 9728). This prevents MCP clients (Open WebUI, VS Code) from discovering the OAuth authorization server to initiate the flow.

The root cause is that the empty-bearer check in `_auth_no_token()` fires **before** the per-server OAuth enforcement, returning a generic `Bearer` without the `resource_metadata` URL.

Fixes #3990

## 🔁 Reproduction Steps

1. Configure a gateway with a backend MCP server that has `oauth_enabled=true`
2. Send a request with an empty `Authorization: Bearer` header (or `Authorization: Bearer ` with trailing space)
3. Observe the 401 response: `WWW-Authenticate: Bearer` — missing `resource_metadata`
4. MCP client cannot discover the OAuth flow and authentication fails

## 🐞 Root Cause

In `_auth_no_token()`, the empty-bearer check (`bearer_header_supplied`) runs **before** `_check_server_oauth_enforcement()`. When the client sends an empty Bearer header to an OAuth-enabled server, the function returns early with a generic `WWW-Authenticate: Bearer` that lacks the `resource_metadata` attribute.

The strict-mode check (`mcp_require_auth`) also uses a hardcoded `"Bearer"` string without `resource_metadata`.

## 💡 Fix Description

1. Initialize `www_auth = "Bearer"` at the top of `_auth_no_token()`
2. Run the per-server OAuth enforcement check **first** — if the server requires OAuth, `OAuthRequiredError` is raised and handled with full `resource_metadata` (unchanged behavior)
3. Move the empty-bearer check and strict-mode check **after** the per-server enforcement, using the `www_auth` variable instead of hardcoded `"Bearer"`

This ensures that:
- OAuth-enabled servers always return `resource_metadata` in their 401 responses (via `OAuthRequiredError` handler)
- Non-OAuth servers continue to return generic `Bearer` (unchanged)
- The strict-mode check now also benefits from `www_auth` propagation

The change is a **pure reordering** — no new logic, no new code paths. Line count: +11/−11.

## 🧪 Verification

| Check | Command | Status |
| --- | --- | --- |
| Lint suite | `make lint` | ⚠️ Not run (no local dev env) |
| Unit tests | `make test` | ⚠️ Not run (no local dev env) |
| Coverage ≥ 80 % | `make coverage` | ⚠️ Not run (no local dev env) |
| Manual regression no longer fails | Tested in production (Kubernetes) for 3+ weeks with multiple OAuth providers (Microsoft Graph, Freshservice) — clients correctly discover OAuth flow via resource_metadata | ✅ |

- [x] Matches current MCP spec (RFC 9728 resource_metadata in WWW-Authenticate)

## ✅ Checklist
- [x] Code formatted

### DIFF
--- a/mcpgateway/transports/streamablehttp_transport.py
+++ b/mcpgateway/transports/streamablehttp_transport.py
@@ -64,7 +64,6 @@ from starlette.types import Receive, Scope, Send
 # First-Party
 from mcpgateway.cache.global_config_cache import global_config_cache
 from mcpgateway.common.models import LogLevel
-from mcpgateway.common.validators import validate_meta_data as _validate_meta_data
 from mcpgateway.config import settings
 from mcpgateway.db import SessionLocal
 from mcpgateway.middleware.rbac import _ACCESS_DENIED_MSG
@@ -1158,22 +1157,6 @@ async def _validate_streamable_session_access(
     return False, HTTP_403_FORBIDDEN, "Session owner metadata unavailable"
 
 
-def _build_paginated_params(meta: Optional[Any]) -> Optional[PaginatedRequestParams]:
-    """Build a ``PaginatedRequestParams`` carrying ``_meta`` when provided.
-
-    Args:
-        meta: Request metadata (_meta) from the original MCP request, or ``None``.
-
-    Returns:
-        A ``PaginatedRequestParams`` instance with ``_meta`` set, or ``None`` when *meta* is falsy.
-    """
-    if not meta:
-        return None
-    # CWE-532: log only key names, never values which may carry PII/tokens
-    logger.debug("Forwarding _meta to remote gateway (keys: %s)", sorted(meta.keys()) if isinstance(meta, dict) else type(meta).__name__)
-    return PaginatedRequestParams(_meta=meta)
-
-
 async def _proxy_list_tools_to_gateway(gateway: Any, request_headers: dict, user_context: dict, meta: Optional[Any] = None) -> List[types.Tool]:  # pylint: disable=unused-argument
     """Proxy tools/list request directly to remote MCP gateway using MCP SDK.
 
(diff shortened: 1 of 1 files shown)


ecthelion77:
**Suggested labels:** `bug`, `SHOULD`, `python`, `mcp-protocol`, `security`

PR #3839, opened 2026-03-24
feat(discovery): add automatic tool discovery with hot/cold classification

Lang-Akshay:
Closes #3734

> **Note:** This branch also includes the Layer 2 RBAC token-narrowing fix from #3919 (session-token team narrowing in permission checks).

---

## Overview

This PR implements **automatic tool discovery** for upstream MCP servers via usage-aware adaptive polling. The gateway now continuously synchronises tool lists from registered servers without any manual intervention — polling frequently-used servers at 1× the base interval and deprioritising idle servers to 3× — reducing unnecessary load while keeping active integrations fresh.

Previously, tool lists were only refreshed on demand or via manual admin action. With this change, the gateway discovers new, updated, and removed tools automatically, reflecting upstream changes within one poll cycle.

---

## Design Rationale: Polling vs. Push Notifications

The MCP spec defines `notifications/tools/list_changed` as the canonical mechanism for dynamic tool discovery, and it is a reasonable default for single-session clients. For a gateway operating at scale, persistent-connection notifications introduce a set of problems that polling sidesteps cleanly — this section explains that tradeoff honestly.

### Why persistent notifications don't fit the gateway model

**Notifications require a live transport stream.** The MCP SDK delivers notifications through a `_receive_loop` tied to the open connection. The gateway's refresh path (`_initialize_gateway` → `connect_to_sse_server` / `connect_to_streamablehttp_server`) uses ephemeral connections — open, fetch tools/list, close. No `message_handler` is registered and the notification window is effectively zero.

**Session pools are demand-driven, not proactive.** `MCPSessionPool` does maintain persistent sessions with notification handlers, but sessions are only created when users invoke tools. If no tools have been called against a gateway, no session exists and no notifications are received. Idle sessions are evicted after 600 s (`MCP_SESSION_POOL_IDLE_EVICTION`). The pool covers active user traffic, not passive server monitoring.

**The connection cost scales poorly.** Listening to N upstream servers requires N open TCP sockets and 2N asyncio tasks per worker, plus keepalive traffic and reconnect logic. At realistic deployment sizes:

| Scale | Persistent Notifications | Ephemeral Polling |
|---|---|---|
| Connections at rest | N per worker | 0 |
| asyncio tasks at rest | 2N per worker | 0 |
| Multi-worker support | ✗ (each worker needs own connections) | ✓ (leader election) |
| Server restart recovery | Requires explicit reconnect | Next poll picks it up |
[...]

PR #4003, opened 2026-04-02
Benchmark docker compose deployment fix

dima-zakharov:
## 🔗 Related Issue
The deployment and benchmark creates situation when tools not found:
```
make testing-rebuild-rust-full
```
As a result of this one cannot test with `make benchmark-mcp-tools`

---

## 📝 Summary
The PR fixes the gateway deployment issue, the tools list is empty when deployed with : `make testing-up-rust-full`

---

## 🏷️ Type of Change
- [x] Bug fix
- [x] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

---

## ✅ Checklist

---

## 📓 Notes (optional)

### DIFF
--- a/mcpgateway/services/gateway_service.py
+++ b/mcpgateway/services/gateway_service.py
@@ -1289,8 +1289,7 @@ class GatewayService(BaseService):  # pylint: disable=too-many-instance-attribut
 
                 invalidate_passthrough_header_caches()
 
-            # Invalidate registry cache and tool lookup cache for newly registered gateway tools/resources/prompts
-            # This ensures that subsequent list_tools and invoke_tool calls see the newly discovered items
+            # Invalidate caches for new gateway items
             cache = _get_registry_cache()
             if tools:
                 await cache.invalidate_tools()
@@ -1299,7 +1298,7 @@ class GatewayService(BaseService):  # pylint: disable=too-many-instance-attribut
             if db_prompts:
                 await cache.invalidate_prompts()
 
-            # Invalidate tool lookup cache for this gateway
+            # Invalidate tool lookup cache
             tool_lookup_cache = _get_tool_lookup_cache()
             await tool_lookup_cache.invalidate_gateway(str(db_gateway.id))
 


Lang-Akshay (changes requested):
Thanks you @dima-zakharov for the PR. Please implement following changes. 
## Finding #1: Race Condition in Cache Invalidation (CWE-362)
### Problem

Cache invalidation broadcasts fire before the database transaction commits in `register_gateway()`. Other workers receiving the invalidation signal may query the DB before the commit completes, potentially caching empty/stale data.

**Current Flow:**

1. `db.flush()` — writes to DB (uncommitted)
2. Cache invalidation + Redis broadcast (lines 485–495)
3. Return to caller
4. `db.commit()` happens later in `get_db()` dependency

> [!CAUTION]
> **Impact:** Workers can cache stale data during the window between invalidation broadcast and commit.

### Solution

Move cache invalidation to occur after `db.commit()`, matching the pattern already used in `update_gateway()` (lines 1634–1646). Either:

- Add explicit `db.commit()` before invalidation in `register_gateway()`
- Use post-commit hooks/events to trigger invalidation

---

## Finding #2: Unconditional Expensive Cache Operations (CWE-400)
### Problem

`invalidate_gateway()` is called unconditionally on every gateway registration, even when the gateway has zero tools. This triggers O(N) Redis operations (key scans + pub/sub broadcast) unnecessarily.

> [!CAUTION]
> **Impact:**
> - Wasted resources on empty gateways
> - No rate limiting on `POST /gateways`
> - Potential thundering herd on Redis pub/sub channel

### Solution

Quick fix: Only call `invalidate_gateway()` when tools exist:

```python
if tools:
    await tool_lookup_cache.invalidate_gateway(str(db_gateway.id))
```

Additional improvements:

- Add rate limiting to gateway registration endpoint
- Optimize `invalidate_gateway()` to check for cached data before scanning
- Consider lazy invalidation strategies

## Add tests for this

crivetimihai:
Thanks for identifying this stale-cache issue, @dima-zakharov — the root cause you found was real and impactful for benchmark workflows.

However, this fix has already been superseded by #3839 (`feat(discovery): add automatic tool discovery with hot/cold classification`), which landed on `main` as `506151675`. That PR independently addressed the same root cause with a more comprehensive approach:

- **`db.flush()` → `db.commit()`** — resolves the race condition where cache invalidation fired before the transaction committed (the concern @Lang-Akshay raised in Finding #1).
- **Unconditional cache invalidation** for gateways, tools, resources, prompts, tool lookup, and admin stats — covers all cases regardless of whether the gateway registered with tools/resources/prompts.
- **Additional invalidations** (`invalidate_gateways()`, `admin_stats_cache.invalidate_tags()`) that this PR was missing.

After rebasing this branch onto current `main`, the net diff is zero — every change here is already present in a stricter form.

@Lang-Akshay — both of your review findings are also addressed on `main`:
1. **Race condition**: Fixed by the `flush` → `commit` change.
2. **Unconditional expensive ops**: Main chose unconditional invalidation, which is the safer default for a low-frequency registration operation where correctness matters more than saving a few Redis round-trips.

Closing as already fixed. Thanks again for the contribution!

PR #4029, opened 2026-04-03
fix(mcp): resolve MCP session pool memory leak, dead-worker locks and polling inefficiency

### LINKED ISSUE #4002: [BUG]: Session Pool Reliability & Resource Management Issues
**Component:** `mcpgateway/services/mcp_session_pool.py`
**Areas:** Scalability, Robustness, Resource Management

---

## Summary

Five issues were identified in the MCP session pool that affect long-running deployments, multi-worker resilience, and resource efficiency. Together they represent risk of memory leaks, request failures during worker restarts, unnecessary CPU overhead, and dead code that complicates maintenance.

---

## Issue 1 — `_mcp_session_mapping` Memory Leak

**Impact:** Scalability / Resource Management
**Severity:** High

`_mcp_session_mapping` (a dict mapping `(mcp_session_id, url, transport, gateway_id)` → `PoolKey`) grows monotonically. Entries are added in `register_session_mapping()` and `acquire()` but are never evicted:

- `close_all()` clears `_pools`, `_active`, `_locks`, `_semaphores` — but not `_mcp_session_mapping`.
- `drain_all()` clears `_pools` and `_active` — but not `_mcp_session_mapping`.
- `_maybe_evict_idle_pool_keys()` evicts idle pool keys but never touches this dict.
- No `pop`, `del`, or `clear` is ever called on `_mcp_session_mapping`.

In session-affinity-enabled deployments with high session churn, this dict will grow indefinitely for the lifetime of the worker process.

### Recommendation

Add cleanup of stale mapping entries inside `_maybe_evict_idle_pool_keys()`. When a pool key is evicted, remove all `_mcp_session_mapping` entries that point to it. Also clear `_mcp_session_mapping` in both `close_all()` and `drain_all()`.

---

## Issue 2 — Dead-Worker Redis Ownership Lock

**Impact:** Robustness / Availability
**Severity:** High

`WORKER_ID` is derived from `hostname:PID` at module import time. When a worker crashes and is respawned (e.g., by gunicorn), the new process gets a new PID and therefore a new `WORKER_ID`. The dead worker's Redis ownership keys (`mcpgw:pool_owner:{session_id}`) persist until their TTL expires (default 300s).

During this window:

- `acquire()` raises `RuntimeError("Session owned by another worker")` for any request targeting the orphaned session.
- `forward_request_to_owner()` publishes to a channel no process is listening on, then times out.
- There is no heartbeat, liveness probe, or dead-worker detection mechanism.

Result: All requests tied to sessions owned by the dead worker fail for up to 5 minutes.

### Recommendation

Implement a lightweight worker heartbeat in Redis (e.g., a per-worker key with a short TTL refreshed periodically). Before raising "owned by another worker" or forwarding, check whether the owning worker's heartbeat key exists. If the owner is dead, atomically reclaim ownership and execute locally. This follows the lease-based distributed lock pattern (cf. Redlock).

---

## Issue 3 — `_pending_responses` is Dead Code

**Impact:** Maintainability
**Severity:** Low

`_pending_responses` is initialized at line 384:

```python
self._pending_responses: Dict[str, asyncio.Future[Dict[str, Any]]] = {}
```

It is never read from, written to, or referenced anywhere else in the file. `forward_request_to_owner()` uses inline pub/sub polling instead of a future-based dispatch pattern. The attribute exists only as initialization dead code.

### Recommendation

Remove `_pending_responses` and its type import if no longer needed. This reduces confusion for future contributors who might assume it is part of the forwarding mechanism.

---

## Issue 4 — Busy-Wait Polling in `forward_request_to_owner()`

**Impact:** Scalability / CPU Efficiency
**Severity:** Medium

The response-wait loop in `forward_request_to_owner()` polls every 100ms:

```python
async with asyncio.timeout(effective_timeout):
    while True:
        msg = await pubsub.get_message(ignore_subscribe_messages=True, timeout=0.1)
        if msg and msg["type"] == "message":
            return orjson.loads(msg["data"])
```

While `get_message` is async and yields to the event loop, each concurrent forwarded request spawns its own independent polling loop. With N forwarded requests, that is N wake-ups every 100ms — CPU cost scales linearly with concurrency even when idle.

### Recommendation

Replace the polling loop with a blocking async listen. `redis.asyncio` supports `pubsub.get_message(timeout=None)` which suspends the coroutine until a message arrives (no polling). Alternatively, use an `asyncio.Event` or `asyncio.Future` set by a single shared listener task to fan out responses, which eliminates per-request polling entirely.

---

## High-Level Recommendations

| Priority | Issue | Fix Category |
|----------|-------|-------------|
| P1 | Memory leak in `_mcp_session_mapping` | Add eviction + clear on lifecycle methods |
| P1 | Dead-worker ownership lock | Heartbeat-based lease with automatic reclaim |
| P2 | Busy-wait polling loop | Switch to blocking async subscribe or shared listener |
| P3 | Dead code `_pending_responses` | Remove |

**General guidance:**

- **Lifecycle symmetry:** Every dict that accumulates entries should have a corresponding cleanup path — either time-based eviction or explicit clearing in `close_all()`/`drain_all()`.
- **Distributed lease hygiene:** Redis-based ownership must account for holder failure. A TTL alone is insufficient when the TTL window is long relative to acceptable downtime. Heartbeats or fencing tokens are the industry-standard approach.
- **Avoid per-request polling:** In async systems, prefer event-driven wake-up over timed polling. This is critical for services expected to handle high concurrency under multi-worker deployments.

bogdanmariusc10:
## 🔗 Related Issue
Closes #4002

---

## 📝 Summary

This PR resolves four bugs in the MCP session pool (`mcpgateway/services/mcp_session_pool.py`) that affect scalability, robustness and resource management in multi-worker deployments:
- Memory Leak (HIGH): `_mcp_session_mapping` dictionary grew indefinitely, never evicting stale entries
- Dead-Worker Lock (HIGH): Orphaned Redis ownership keys caused 5-minute request failures when workers crashed
- Dead Code (LOW): Unused `_pending_responses` attribute cluttered the codebase
- Busy-Wait Polling (MEDIUM): 100ms polling loops scaled CPU cost linearly with concurrent forwarded requests

### Changes Made

**Bug 1 - Memory Leak Fix:**
- Added cleanup of `_mcp_session_mapping` in `_maybe_evict_idle_pool_keys()` when pool keys are evicted
- Clear mapping in `close_all()` and `drain_all()` lifecycle methods

**Bug 2 - Dead-Worker Detection:**
- Implemented Redis-based heartbeat mechanism (30s TTL, refreshed every 10s)
- Added` _worker_heartbeat_key()`, `_start_heartbeat()`, `_is_worker_alive()` methods
- Modified `acquire()` to detect dead workers and atomically reclaim ownership
- Updated `forward_request_to_owner()` to check worker liveness before forwarding

**Bug 3 - Dead Code Removal:**
- Removed unused `_pending_responses` attribute

**Bug 4 - Polling Optimization:**
- Replaced `get_message(timeout=0.1)` polling loop with `pubsub.listen()` async iterator
- Eliminates per-request wake-ups, reducing CPU overhead in high-concurrency scenarios

**Tests Added:**
- `test_worker_heartbeat_key_format` - verifies heartbeat key format includes worker ID
- `test_is_worker_alive_returns_true_when_heartbeat_exists` - tests liveness detection for active workers
- `test_is_worker_alive_returns_false_when_heartbeat_missing` - tests dead worker detection

---

## 🏷️ Type of Change
- [x] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

| Lint suite                | `make lint`     |    Pass ✅    |
| Unit tests                | `make test`     |    Pass ✅    |
| Coverage ≥ 80%            | `make coverage` |    Pass ✅    |

---

## ✅ Checklist

---

### Design Decisions

**Heartbeat Implementation:**
- Chose 30s TTL with 10s refresh interval to balance liveness detection speed vs Redis load
- Heartbeat runs as background task only when session affinity is enabled
- Graceful shutdown cancels heartbeat task and waits for completion

**Ownership Reclaim Strategy:**
- Atomic delete + re-register pattern prevents race conditions
- Falls back to local session creation if reclaim fails
- Maintains backward compatibility with non-affinity deployments

**Polling Replacement:**
- `pubsub.listen()` is the recommended async pattern for Redis pub/sub
- Eliminates CPU waste from timed polling in idle periods
- Maintains same timeout behavior via `asyncio.timeout()` wrapper

### Impact

These fixes address production reliability issues in session-affinity-enabled deployments:
- Prevents memory exhaustion in long-running workers
- Reduces request failure window from 5 minutes to <30 seconds on worker crashes
- Improves CPU efficiency under high concurrent forwarding load

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "package-lock.json|Cargo.lock|^.secrets.baseline$|scripts/sign_image.sh|scripts/zap|sonar-project.properties|^/Users/brian/dev/github.ibm.com/contextforge-org/sps-pipeline-config/.secrets.baseline$|^./.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-04-04T10:21:11Z",
+  "generated_at": "2026-04-04T15:50:18Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -92,7 +92,7 @@
         "hashed_secret": "08cd923367890009657eab812753379bdb321eeb",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 509,
+        "line_number": 522,
         "type": "Basic Auth Credentials",
         "verified_result": null
       },
@@ -100,7 +100,7 @@
         "hashed_secret": "14f8aa3e560a47851908ab0f04ec856dbc512d93",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 704,
+        "line_number": 717,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -108,7 +108,7 @@
         "hashed_secret": "fa9beb99e4029ad5a6615399e7bbae21356086b3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 964,
+        "line_number": 977,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -116,7 +116,7 @@
         "hashed_secret": "7b4455a56fbf1d198e45e04c437488514645a82c",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 990,
+        "line_number": 1003,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -124,7 +124,7 @@
         "hashed_secret": "ac371b6dcce28a86c90d12bc57d946a800eebf17",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1070,
+        "line_number": 1083,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -132,7 +132,7 @@
         "hashed_secret": "0b6ec68df700dec4dcd64babd0eda1edccddace1",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1075,
+        "line_number": 1088,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -140,7 +140,7 @@
         "hashed_secret": "4ad6f0082ee224001beb3ca5c3e81c8ceea5ed86",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1080,
+        "line_number": 1093,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -148,7 +148,7 @@
         "hashed_secret": "cb32747fcfb55eaa194c8cd8e4ba7d49ada08a94",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1086,
+        "line_number": 1099,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -156,7 +156,7 @@
         "hashed_secret": "6c178d51b13520496dbc767ed3d9d7aa5803ac72",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1098,
+        "line_number": 1111,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -164,7 +164,7 @@
         "hashed_secret": "ca45060a53fd8a255d1a83ee8d2f025283ccc66e",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1116,
+        "line_number": 1129,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -172,7 +172,7 @@
         "hashed_secret": "910fbf00f58e9bcb095ea26a75cc1d9a3355e671",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1177,
+        "line_number": 1190,
         "type": "Secret Keyword",
         "verified_result": null
       }
@@ -242,7 +242,7 @@
         "hashed_secret": "b4673e578b9b30fe8bba1b555b7b59883444c697",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 482,
+        "line_number": 487,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -250,7 +250,7 @@
         "hashed_secret": "4a0a2df96d4c9a13a282268cab33ac4b8cbb2c72",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 570,
+        "line_number": 575,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -258,7 +258,7 @@
         "hashed_secret": "5baa61e4c9b93f3f0682250b6cf8331b7ee68fd8",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 920,
+        "line_number": 925,
         "type": "Basic Auth Credentials",
         "verified_result": null
       },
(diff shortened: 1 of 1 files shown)


crivetimihai:
## Review & Changes

Nice work on this one — the four bugs you identified are real and the overall design (heartbeat TTL, CAS reclaim, mapping eviction) is sound. I rebased onto main, resolved a conflict in `.secrets.baseline`, and then made several fixes during review. Here's what changed and why.

### Fixes applied during review

**1. Atomic CAS for ownership reclaim** (`mcp_session_pool.py`)

The original `acquire()` dead-worker reclaim used `delete` + `set` — two separate Redis commands with a race window. Two workers detecting the same dead owner could both reclaim simultaneously. Replaced with a Lua compare-and-swap script that checks the current owner matches the dead worker before overwriting, consistent with the existing `register_pool_session_owner()` Lua pattern already in this file. If CAS returns 0 (another worker already reclaimed), a RuntimeError is raised instead of silently proceeding.

Applied the same CAS pattern to `forward_request_to_owner()`.

**2. Semaphore double-release** (`mcp_session_pool.py`)

Both the CAS-failure and owner-is-alive branches called `semaphore.release()` before raising `RuntimeError`, but the outer `except BaseException` handler already releases the semaphore on any failure. This double-release would inflate the semaphore above `mcpgateway_pool_max_sessions`, allowing more concurrent sessions than configured. Removed the inner release calls.

**3. Heartbeat not started on production path** (`mcp_session_pool.py`, `main.py`)

`init_mcp_session_pool()` creates the pool via `__init__`, never `__aenter__`. In production, `_heartbeat_task` stayed `None`, no heartbeat keys were written, and `_is_worker_alive()` returned `False` for all workers — causing constant ownership fights.

Extracted a public `start_heartbeat()` method (idempotent, mirrors `start_rpc_listener()` pattern). Called from `main.py` alongside the RPC listener. `__aenter__` now delegates to `start_heartbeat()`.

**4. `forward_request_to_owner` CAS-failure handling** (`mcp_session_pool.py`)

When CAS returned 0 (another worker reclaimed), the function still returned `None` (execute locally), violating single-owner routing. Now it re-reads the new owner from Redis and falls through to the forwarding logic.

**5. Affinity-only deployment startup/shutdown** (`main.py`)

The heartbeat/RPC startup was nested under `mcp_session_pool_enabled`, but the pool init uses `pool_enabled OR affinity_enabled`. Affinity-only deployments created the pool but never started heartbeat/RPC. Similarly, the shutdown guard only checked `pool_enabled`. Fixed both to use `pool_enabled or affinity_enabled`.

**6. Test fixes**

- Updated mocks for the new CAS pattern (`redis.eval` instead of `redis.delete` + `redis.set`)
- Added explicit `redis.exists` mocking for `_is_worker_alive` in forwarding tests (was relying on fragile AsyncMock truthiness)
- Fixed vacuous `_cleanup_pool_session_owner` test — method expects `str` but test passed `MagicMock(spec=PooledSession)`. Replaced with two proper tests (owned key deletion + unowned key skip)
- Added 20+ new tests covering: heartbeat loop, fail-open behavior, CAS success/failure/Redis-unavailable, forward dead-owner with CAS-loss forwarding to new owner, eviction mapping cleanup, lifecycle management, `start_heartbeat()` idempotency, affinity-disabled paths

All 5697 service tests pass.

PR #4034, opened 2026-04-03
fix: honor fail_on_plugin_error during plugin init

### LINKED ISSUE #4033: [BUG]: Plugin startup ignores fail_on_plugin_error=false for external plugin init failures
### 🐞 Bug Summary
When an external plugin fails to initialize during gateway startup, the plugin manager aborts startup even if `plugin_settings.fail_on_plugin_error` is set to `false`.

In practice, a misconfigured or temporarily unreachable external plugin prevents ContextForge from booting. The resulting logs also include noisy async teardown errors from the streamable HTTP client, but the primary bug is that startup does not honor the configured fail-open behavior.

---

### 🧩 Affected Component

- [x] `mcpgateway` - API
- [ ] Other (explain below)

---

### 🔁 Steps to Reproduce

1. Enable plugins and configure an external MCP plugin with `plugin_settings.fail_on_plugin_error: false`.
2. Point that external plugin at a streamable HTTP endpoint that is unavailable during gateway startup.
3. Start ContextForge.
4. Observe that startup fails instead of logging and skipping the unavailable plugin.

Minimal config shape:

```yaml
plugin_settings:
  enabled: true
  auto_load: true
  fail_on_plugin_error: false

plugins:
  - name: "LLMGuardPlugin"
    kind: "external"
    mode: "enforce"
    mcp:
      proto: STREAMABLEHTTP
      url: http://host.docker.internal:8001/mcp
```

---

### 🤔 Expected Behavior
With `fail_on_plugin_error: false`, startup should continue, the failing plugin should be skipped with a clear log message, and the rest of the gateway should come up normally.

---

### 📓 Logs / Error Output
Observed startup includes plugin init failure followed by async teardown noise such as:

```text
RuntimeError: Attempted to exit cancel scope in a different task than it was entered in
RuntimeError: aclose(): asynchronous generator is already running
```

The startup failure is triggered from plugin-manager initialization rather than respecting the configured fail-open behavior.

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Version or commit | `main@fa1838835` |
| Runtime | `Python 3.12.12` |
| Platform / OS | `Darwin 25.3.0 arm64` |
| Container | `Docker` |

---

Relevant code path:
- `mcpgateway/plugins/framework/manager.py` currently raises from `PluginManager.initialize()` on any plugin load failure, regardless of `plugin_settings.fail_on_plugin_error`.

Suggested regression test:

```python
@pytest.mark.asyncio
async def test_manager_initialize_skips_plugin_load_errors_when_configured():
    """Startup should continue when plugin init fails and fail_on_plugin_error is false."""
    PluginManager.reset()
    manager = PluginManager("./tests/unit/mcpgateway/plugins/fixtures/configs/valid_single_plugin.yaml")

    with patch.object(manager._loader, "load_and_instantiate_plugin", side_effect=RuntimeError("plugin offline")):
        await manager.initialize()

    assert manager.initialized is True
    assert manager.plugin_count == 0

    await manager.shutdown()
    PluginManager.reset()
```

lucarlig:
# 🐛 Bug-fix PR

Before opening this PR please:

1. `make lint`            - passes `ruff`, `mypy`, `pylint`
2. `make test`            - all unit + integration tests green
3. `make coverage`        - ≥ 80 %
4. `make docker docker-run-ssl` or `make podman podman-run-ssl`
5. Update relevant documentation.
6. Tested with sqlite and postgres + redis.
7. Manual regression no longer fails. Ensure the UI and /version work correctly.

---

## 📌 Summary
Honor `plugin_settings.fail_on_plugin_error` during plugin-manager startup so an unavailable external plugin can be skipped instead of aborting gateway boot. Add a regression test that locks the fail-open startup behavior in place.

## 🔁 Reproduction Steps
Closes #4033

1. Enable plugins with `plugin_settings.fail_on_plugin_error: false`.
2. Configure an external MCP plugin whose streamable HTTP endpoint is unavailable at startup.
3. Start the gateway.
4. Observe that startup aborts from `PluginManager.initialize()` instead of logging and skipping the failing plugin.

## 🐞 Root Cause
`PluginManager.initialize()` unconditionally raised `RuntimeError` on plugin instantiation failures, even when config explicitly requested fail-open behavior via `fail_on_plugin_error: false`.

## 💡 Fix Description
Check `plugin_settings.fail_on_plugin_error` inside the plugin-load failure path during manager initialization. When it is `false`, log the failure, skip the plugin, and continue startup. Add a unit regression test that patches `load_and_instantiate_plugin()` to raise and asserts initialization still completes with an empty registry.

## 🧪 Verification

| Check                                 | Command                                                                  | Status |
|---------------------------------------|--------------------------------------------------------------------------|--------|
| Lint suite                            | `make lint`                                                              | Not run |
| Unit tests                            | `uv run pytest tests/unit/mcpgateway/plugins/framework/test_manager.py tests/unit/mcpgateway/plugins/framework/test_manager_extended.py -q` | Passed |
| Coverage ≥ 80 %                       | `make coverage`                                                          | Not run |
| Manual regression no longer fails     | Startup with unreachable external MCP plugin and `fail_on_plugin_error: false` | Not run |

## ✅ Checklist

### DIFF
--- a/tests/unit/mcpgateway/plugins/framework/test_manager_extended.py
+++ b/tests/unit/mcpgateway/plugins/framework/test_manager_extended.py
@@ -207,8 +207,13 @@ async def test_manager_initialize_skips_plugin_load_errors_when_configured():
     PluginManager.reset()
     manager = PluginManager("./tests/unit/mcpgateway/plugins/fixtures/configs/valid_single_plugin.yaml")
 
-    with patch.object(manager._loader, "load_and_instantiate_plugin", side_effect=RuntimeError("plugin offline")):
+    with (
+        patch.object(manager._loader, "load_and_instantiate_plugin", side_effect=RuntimeError("plugin offline")),
+        patch("mcpgateway.plugins.framework.manager.logger") as mock_logger,
+    ):
         await manager.initialize()
+
+    mock_logger.warning.assert_called_with("Skipping plugin %s because fail_on_plugin_error is disabled", "ReplaceBadWordsPlugin")
 
     assert manager.initialized is True
     assert manager.plugin_count == 0


madhu-mohan-jaishankar (changes requested):
`logger.warning` message for the skip behavior should be asserted.

lucarlig:
Updated the regression test to assert the skip warning log as requested. The focused manager test slice was rerun and is green.

PR #4035, opened 2026-04-03
fix: Add build js command to makefile serve commands

gcgoncalves:
## 📝 Summary
Adds a js-build target to the Makefile that runs npm install and npm run vite:build, wired as a prerequisite to all non-containerised app-starting targets (serve, serve-ssl, serve-granian, serve-granian-ssl, serve-granian-http2, dev, dev-remote, run). This ensures the Vite JS bundle is always up to date before the server starts, without requiring a separate manual step.

---

## 🏷️ Type of Change
- [ ] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [X] Chore (deps, CI, tooling)
- [ ] Other (describe below)

Ensuring the JS/Vite bundle is built as part of the standard server startup workflow

---

## 📓 Notes (optional)
The js-build target is declared .PHONY so it always runs. npm install is included before the build step to keep the workflow self-contained - no assumption that node_modules is already present. dev-echo was left unchanged as it is an observability diagnostic variant, not a primary entry point.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -260,8 +260,11 @@ check-env-dev:
 
 ## --- JS build ----------------------------------------------------------------
 js-build:                        ## Install npm dependencies and build JS bundle with Vite
-	npm install
-	npm run vite:build
+	@if command -v npm >/dev/null 2>&1; then \
+		npm install --no-audit --no-fund && npm run vite:build; \
+	else \
+		echo "WARNING: npm not found — skipping JS bundle build (admin UI may not load)"; \
+	fi
 
 ## --- Primary servers ---------------------------------------------------------
 serve: js-build                  ## Run production server with Gunicorn + Uvicorn (default)


crivetimihai (dismissed):
Nice contribution — this is a solid DX improvement. Having the Vite bundle built automatically before any server target prevents a common "blank UI" gotcha on fresh checkouts.

A couple of changes I made during rebase:

1. **Rebased onto main** — the two merge commits are gone, replaced with a clean linear history.
2. **Dropped `.secrets.baseline` from the original commit** — the line-number shifts were artifacts of branch divergence and are handled separately.
3. **Added `js-build` as a prerequisite to `dev-echo`** — this was the only dev/serve target missing it, and someone running `make dev-echo` on a fresh checkout would have gotten a broken frontend. Added a follow-up commit for this + the regenerated secrets baseline.

Minor note for future reference: the `js-build` target uses `npm install` while CI uses `npm ci` and Containerfiles use `npm install --frozen-lockfile`. For local dev this is fine — `npm ci` would be slower on repeated invocations since it wipes `node_modules` each time. Just worth keeping in mind if reproducibility ever becomes an issue.

LGTM — approving.

crivetimihai:
Follow-up fix pushed: the `js-build` target now checks for `npm` availability before running. If npm isn't installed, it prints a warning and lets the server start anyway — the API works fine without the JS bundle, only the admin UI would be missing.

This avoids introducing Node.js as a hard runtime dependency for `make serve` / `make dev` in environments that don't need the admin UI (e.g. headless API deployments, CI runners without Node).

Also added `--no-audit --no-fund` flags to keep npm output clean during startup.

PR #4060, opened 2026-04-07
feat(ui): Overflow menu

### LINKED ISSUE #3519: [FEATURE][UI]: Replace individual actions in tables with overflow menu
Replace the group of actions with an overflow menu for a more modern, decluttered approach. Examples: 
[Shadcn](https://www.shadcn.io/patterns/data-table-advanced-4)
[Carbon](https://carbondesignsystem.com/components/overflow-menu/usage/) (functionality, not style reference)

gcgoncalves:
# ✨ Feature / Enhancement PR

## 🔗 Epic / Issue

Closes #3519

---

## 🚀 Summary

Replaces per-row action buttons across all admin tables with a fixed-positioned overflow menu (first introduced on the Gateways table, then rolled out to Tools, Servers, Resources, Prompts, Agents, and Roots) and extracts the shared Alpine state and scroll-lock logic into a reusable `Admin.overflowMenu()` factory component.

---

## 🧪 Checks

- [X] `make lint` passes
- [X] `make test` passes
- [X] CHANGELOG updated (if user-facing)

---

## 📓 Notes (optional)

- New file: `mcpgateway/admin_ui/components/overflow-menu.js`.
- All seven templates (roots table in `admin.html` + six partials) use `x-data="Admin.overflowMenu('…-table-wrapper')"` instead of ~20 lines of duplicated inline JS.
- The `components/` directory is set up to house future Alpine components.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -1,9 +1,9 @@
 {
   "exclude": {
-    "files": "^.secrets.baseline$|package-lock.json|Cargo.lock|scripts/sign_image.sh|scripts/zap|sonar-project.properties|uv.lock",
+    "files": "^.secrets.baseline|package-lock.json|Cargo.lock|scripts/sign_image.sh|scripts/zap|sonar-project.properties|uv.lock|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-04-08T16:53:59Z",
+  "generated_at": "2026-04-12T07:25:34Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -200,7 +200,7 @@
         "hashed_secret": "f4793151b0607198d4de9b1ca458d3e25adf1cb7",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 111,
+        "line_number": 112,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -208,7 +208,7 @@
         "hashed_secret": "fa9beb99e4029ad5a6615399e7bbae21356086b3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 187,
+        "line_number": 188,
         "type": "Secret Keyword",
         "verified_result": null
       },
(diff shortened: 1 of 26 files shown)


vishu-bh:
Thanks for addressing the earlier wiring issues. The latest revision fixes the broken A2A view action and the missing `Admin.handleToggleSubmit(...)` references, which is a solid improvement. I still see a few things worth tightening up before merge:

1. `mcpgateway/admin_ui/components/overflow-menu.js`
`mcpgateway/templates/gateways_partial.html:52`
The menu positioning is currently driven by dynamic Tailwind class names like `top-[${menuTop}px] left-[${menuLeft}px]`. That’s a brittle fit here and may not produce reliable runtime CSS. A safer approach would be to bind these coordinates through `:style`, for example setting `top` and `left` directly from Alpine state while keeping the rest of the visual styling in classes.

2. `mcpgateway/admin_ui/components/overflow-menu.js`
The menu is always opened at `rect.bottom + 4` / `rect.left` and never clamped to the viewport. On rows near the right or bottom edge, that can push part of the menu off-screen, and since scrolling is disabled while the menu is open, some actions may become unreachable. It would be worth clamping `top`/`left` against `window.innerWidth` and `window.innerHeight`, or flipping the menu upward when there isn’t enough space below.

3. `mcpgateway/templates/agents_partial.html:27`
`mcpgateway/templates/admin.html:6843`
The agents table still uses `Admin.overflowMenu('agents-table-wrapper')`, but the surrounding container in `admin.html` does not define a matching `id="agents-table-wrapper"`. That means the wrapper-specific overflow suppression path is skipped only for A2A agents, so this table behaves differently from the other migrated tables. Adding the missing wrapper id would make the behavior consistent.\n\n4. `tests/unit/js/overflow-menu.test.js`\nThe new tests are better than before for the component internals, but the “wiring” coverage still doesn’t exercise the actual template integration. Since the regressions in this PR have come from template/component hookups, I think the safest addition would be either:\n- a small integration-style JS test that renders one of the menu snippets and verifies the generated positioning/style behavior, or\n- a Playwright assertion around an edge-positioned row that confirms the menu remains visible and actionable.

gcgoncalves:
@vishu-bh re-review?

gcgoncalves:
That's intentional, following the overflow menu standard of other libraries.

PR #3892, opened 2026-03-27
fix(security): Validate Server ID in Streamable HTTP to Prevent Unauthorized Access

bogdanmariusc10:
## 🔗 Related Issue
Closes #3891

---

## 📝 Summary

This PR fixes a critical security vulnerability where invalid server IDs in MCP endpoint URLs (e.g., `/servers/xyz/mcp`) bypassed virtual server isolation and provided unauthorized access to all MCP components.

**The Problem:**

- URLs with invalid server IDs were processed by the SDK without validation
- This allowed bypassing virtual server restrictions
- Attackers could access MCP components they shouldn't have access to

**The Solution:**

- Added database validation for server_id before SDK processing
- Returns `404` for non-existent servers
- Returns `503` on database errors (fail-closed security)
- Validation happens early in the request flow, preventing any further processing

---

## 🏷️ Type of Change
- [x] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Chore (deps, CI, tooling)
- [ ] Other (describe below)

---

## 🧪 Verification

| Check                     | Command         | Status |
|---------------------------|-----------------|--------|
| Lint suite                | `make lint`     |    ✅ passed    |
| Unit tests                | `make test`     |    ✅ passed    |
| Coverage ≥ 80%            | `make coverage` |    ✅ passed   |

---

## ✅ Checklist
- [x] Code formatted (`make black isort pre-commit`)
- [x] Tests added/updated for changes
- [x] Documentation updated (if applicable)
- [x] No secrets or credentials committed

---

### Security Impact

**Before:** Invalid server IDs provided access to all MCP components
**After:** Invalid server IDs are rejected with proper HTTP error codes

### Implementation Details
**File:** `mcpgateway/transports/streamablehttp_transport.py` (lines 2862-2904)

The fix adds validation in `SessionManagerWrapper.handle_streamable_http()`:
```
# SECURITY: Validate that the server_id exists in the database
try:
    async with get_db() as db:
        server = db.execute(select(DbServer).where(DbServer.id == server_id)).scalar_one_or_none()
        if server is None:
            # Return 404 and exit - no further processing
            return
except Exception as e:
    # Return 503 and exit - fail-closed on errors
    return
```

Key Security Properties:

- Early validation: Happens before `server_id_var.set()` (line 2906) and SDK's `handle_request()` (line 2945)
- Fail-closed: Any validation error denies access
- Proper error codes: `404` for not found, `503` for service errors
[...]

Issue #3891, opened 2026-03-27
[BUG]: [Security] Invalid Virtual MCP Server URL provides access to all the MCP components (tools, prompts, resources, agents) available on ContextForge


ankit-gajera-merck:
### 🐞 Bug Summary
By replacing valid virtual server id with some random string in virtual server URL provides access to **all the MCP components (tools, prompts, resources, agents) available on ContextForge**.

Example: 

1) http://localhost:7777/servers/781a47073d0f4ff58e2dfbdaa2bbc14b/mcp <- getting selected tools as expected behaviour
2) http://localhost:7777/servers/xyz/mcp <- Invalid server ID gives access to all components available on ContextForge

---

### 🧩 Affected Component
Select the area of the project impacted:

- [x] `mcpgateway` - API
- [ ] `mcpgateway` - UI (admin panel)
- [ ] `mcpgateway.wrapper` - stdio wrapper
- [ ] Federation or Transports
- [ ] CLI, Makefiles, or shell scripts
- [ ] Container setup (Docker/Podman/Compose)
- [ ] Other (explain below)

---

### 🔁 Steps to Reproduce

1. Create a virtual MCP server with specific tools/prompts/resources
2. Use MCP client to access the virtual MCP server
3. Check that only selected MCP components are accessible
4. Replace the server id from virtual MCP server with "xyz"
5. Check that all the available MCP components are accessible now

---

### 🤔 Expected Behavior
It should provide an error something like 404 not found in case of invalid server id which is not available on ContextForge

---

### 📓 Logs / Error Output
Valid Scenario:

<img width="1570" height="712" alt="Image" src="https://github.com/user-attachments/assets/09c05b73-bc3b-40fe-9b0c-13e8fc793833" />

Bug Scenario:

<img width="1574" height="766" alt="Image" src="https://github.com/user-attachments/assets/02fa5d7e-2873-4d63-b1af-983d4bafb92a" />

---

### 🧠 Environment Info
You can retrieve most of this from the `/version` endpoint.

| Key | Value |
|-----|-------|
| Version or commit | `1.0.0-RC-2` |
| Runtime | ` Python 3.13.11, Gunicorn` |
| Platform / OS | `Darwin 25.3.0 (arm64) macOS` |
| Container | `none` |

---

### 🧩 Additional Context (optional)
Add any configuration details, flags, or related issues.

ankit-gajera-merck:
[...]

PR #4066, opened 2026-04-07
fix(security): Validate server ID in Rust MCP runtime proxy to prevent unauthorized access

### LINKED ISSUE #4059: [BUG]: Rust MCP runtime proxy bypasses server ID validation for non-hex server IDs
### 🐞 Bug Summary

`rust_mcp_runtime_proxy.py` uses a hex-only regex to extract server IDs from `/servers/{id}/mcp` paths and performs no server existence validation before forwarding to the Rust sidecar. A non-hex server ID (e.g. `ndh45`, `my-server`) silently fails to match the regex, causing the request to be forwarded to the Rust runtime **without the `x-contextforge-server-id` header**, which treats it as a global unscoped MCP request and returns all tools/resources.

This is the same class of vulnerability fixed for the Python transport in #3892, but the Rust proxy path was not updated.

---

### 🧩 Affected Component

---

### 🔁 Steps to Reproduce

1. Enable the experimental Rust MCP runtime (`EXPERIMENTAL_RUST_MCP_RUNTIME_ENABLED=true`)
2. Send a `tools/list` request to `/servers/<non-hex-string>/mcp`:
   ```
   POST /servers/ndh45/mcp
   Authorization: Bearer <token>
   Content-Type: application/json

   {"jsonrpc":"2.0","id":1,"method":"tools/list","params":null}
   ```
3. Observe a `200 OK` response containing **all** tools instead of a `404`.

---

### 🤔 Expected Behavior

`404 Not Found` — same as the Python transport path after #3892.

---

### 📓 Logs / Error Output

No errors logged. The proxy silently drops the `x-contextforge-server-id` header and forwards to the Rust sidecar as a global `/mcp/` request.

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Version or commit | `main` (post #3892) |
| Platform / OS | Any |
| Container | Any |

---

### 🧩 Additional Context

**Root cause** — `rust_mcp_runtime_proxy.py:35`:
```python
# Current (broken for non-hex IDs)
_SERVER_ID_RE = re.compile(r"/servers/(?P<server_id>[a-fA-F0-9\-]+)/mcp/?$")
```

The hex-only pattern `[a-fA-F0-9\-]+` does not match strings like `ndh45`. When `_extract_server_id_from_scope()` returns `None`, `_build_forward_headers()` skips the `x-contextforge-server-id` header entirely, and `_build_runtime_mcp_url()` builds a plain `/mcp/` URL — making the Rust runtime serve global scope.

**Two issues to fix:**
1. Change `_SERVER_ID_RE` to `[^/]+` (matching the fix applied to `streamablehttp_transport.py` in #3892)
2. Add server existence validation in `RustMCPRuntimeProxy.handle_streamable_http()` before forwarding — the Python fallback's `_validate_server_id()` is never called on the Rust path

Relates to #3891, #3892.

bogdanmariusc10:
## 🔗 Related Issue
Closes #4059

---

## 📝 Summary
Fixes a security vulnerability in the Rust MCP runtime proxy where non-hex server IDs (e.g., `ndh45`, `my-server`) bypassed server validation and were forwarded to the Rust sidecar without the `x-contextforge-server-id` header. This caused the runtime to serve global scope instead of returning 404.

This is the same class of vulnerability fixed for the Python transport in #3892, but the Rust proxy path was not updated at that time.

**Changes:**
- Updated regex pattern from hex-only `[a-fA-F0-9\-]+` to `[^/]+` to match all server IDs
- Added `_validate_server_id()` function with database validation before forwarding
- Added defense-in-depth pattern `_SERVER_SCOPED_PATH_RE` for malformed paths
- Returns 404 for non-existent servers, 503 for database errors
- Added 4 new tests achieving >95% coverage for the validation logic

---

## 🏷️ Type of Change
- [x] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

| Lint suite                | `make lint`     | ✅ Pass |
| Unit tests                | `make test`     | ✅ Pass |
| Coverage ≥ 80%            | `make coverage` | ✅ Pass |

---

## ✅ Checklist

---

## 📓 Notes
**Security Impact:**
Before this fix, requests like `POST /servers/ndh45/mcp` would silently fail the regex match and be forwarded to the Rust runtime without server scoping, returning all tools/resources instead of 404.

**Test Coverage:**
Added 4 comprehensive tests:
1. `test_validate_server_id_returns_404_when_server_not_found` - Non-existent server validation
2. `test_validate_server_id_returns_503_on_database_error` - Database error handling
3. `test_validate_server_id_rejects_malformed_server_scoped_paths` - Malformed path rejection
4. `test_handle_streamable_http_rejects_invalid_server_id` - End-to-end validation

**Related:**
- Relates to #3891 (original vulnerability report)
- Relates to #3892 (Python transport fix)

### DIFF
--- a/mcpgateway/transports/rust_mcp_runtime_proxy.py
+++ b/mcpgateway/transports/rust_mcp_runtime_proxy.py
@@ -32,7 +32,10 @@ from mcpgateway.utils.orjson_response import ORJSONResponse
 
 logger = logging.getLogger(__name__)
 
-_SERVER_ID_RE = re.compile(r"/servers/(?P<server_id>[a-fA-F0-9\-]+)/mcp/?$")
+_SERVER_ID_RE = re.compile(r"/servers/(?P<server_id>[^/]+)/mcp/?$")
+# Pattern that detects a server-scoped MCP path even when _SERVER_ID_RE doesn't
+# match (e.g. empty segment: /servers//mcp). Used as a defense-in-depth guard.
+_SERVER_SCOPED_PATH_RE = re.compile(r"/servers/.*/mcp(?:/)?$")
 _CONTEXTFORGE_SERVER_ID_HEADER = "x-contextforge-server-id"
 _CONTEXTFORGE_AUTH_CONTEXT_HEADER = "x-contextforge-auth-context"
 _CONTEXTFORGE_AFFINITY_FORWARDED_HEADER = "x-contextforge-affinity-forwarded"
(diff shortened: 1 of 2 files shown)


madhav165:
A few small things tidied up during review:

- **Rebased onto main** — clean, no conflicts.
- **`SessionLocal` → `fresh_db_session`** — the deferred local imports inside `_validate_server_id` were moved to the top of the file and the session acquisition was switched to `fresh_db_session` (the codebase's standard context manager, which adds proper commit/rollback/close lifecycle around the DB call).
- **Regex reverted to `[a-fA-F0-9\-]+`** — server IDs are `uuid4().hex` (32 hex chars), so the broader `[^/]+` was unnecessary. Non-hex IDs now fall through to the `_SERVER_SCOPED_PATH_RE` defense-in-depth guard and are rejected without hitting the database.
- **Trailing whitespace removed** from the test file.

madhav165 (dismissed):

madhav165 (dismissed):

lucarlig:
One gap still looks worth covering: there’s helper-level coverage for malformed paths like `/servers//mcp`, but no `handle_streamable_http()` test that drives that input through the real proxy entrypoint. If the proxy ever stops using `modified_path` correctly, or stops invoking `_validate_server_id()`, the current suite could stay green while the original fall-through bug comes back. I’d add one proxy-level malformed-path test that asserts `404` and confirms neither fallback nor the Rust client is reached.

lucarlig (dismissed):

dima-zakharov (dismissed):

PR #4087, opened 2026-04-08
refactor: consolidate rust workspace under crates

### LINKED ISSUE #3027: [CHORE]: setup rust mcpgatway for internal components
### 🔧 Chore Summary

Set up Rust-based implementation infrastructure for MCPGateway to handle performance-critical for internal components. This includes establishing the Rust workspace structure, PyO3 bindings, CI/CD pipeline, and initial service/plugin implementations.

---

### 🧱 Area Affected

- [x] GitHub Actions / CI Pipelines
- [x] Pre-commit hooks / linters
- [ ] Type-checking (mypy, pyright, pytype, etc.)
- [ ] Docs or spellcheck
- [ ] Other:

---

### ⚙️ Context / Rationale

we need to integrate rust with python and we have to have a place and structure for rust.

---

### 📦 Related Make Targets

---

### 📋 Acceptance Criteria

- [ ] Linter runs cleanly (`make lint`)
- [x] CI passes with no regressions
- [ ] Docs/tooling updated (if applicable)
- [ ] Security scans pass

lucarlig:
# ✨ Feature / Enhancement PR

## 🔗 Epic / Issue
Closes #3027

---

## 🚀 Summary (1-2 sentences)
This PR finishes the Rust workspace move under `crates/`, adds the new local PyO3 native extension crate, and updates CI, supply-chain policy, docs, and container build paths to match the current layout. It also restores full Rust validation on non-draft PRs and adds regression tests so the repo keeps the intended crate, workflow, and container boundaries.

---

## 🧪 Checks

- [ ] CI fully green
- [x] Local workflow guard tests updated

---

## 📓 Notes (optional)
- Rust crates now live directly under `crates/`:
  - `crates/mcp_runtime`
  - `crates/wrapper`
  - `crates/request_logging_masking_native_extension`
- Standalone Rust MCP servers remain outside the root workspace. Their manifests and license handling were adjusted, but they are not workspace members.
- Rust CI now runs the full non-draft PR matrix instead of the reduced `pr-*` hot path.
- `rust fmt` and `rust clippy` now run on Ubuntu only; build/test coverage remains on both Ubuntu and macOS.
- License, deny, and vet policy were updated for the new Rust layout and the new native extension dependency tree.
- `Containerfile.lite` now:
  - builds local native extensions dynamically from `crates/**` when a crate has `Cargo.toml` + `pyproject.toml` + `maturin`
  - builds Rust sidecars dynamically from `crates/**` bin targets
  - uses a UBI 10 Rust builder path instead of the earlier x86-only manylinux builder
  - has a dedicated Rust-enabled CI smoke build so `ENABLE_RUST=true` is validated on PRs
- Wrapper hardening in this branch includes:
  - redacted URL credentials in debug config output
  - no raw MCP payload logging at debug
  - bounded internal queues instead of unbounded channels
- Guard tests were added/updated for:
  - workspace layout
  - Rust workflow behavior
  - docker scan workflow coverage for Rust-enabled container builds
  - wrapper workflow triggers
  - license-check workflow triggers
- Docs and ADRs were updated to reflect the new crate layout, CI shape, and runtime/container behavior.

### DIFF
--- a/Containerfile.lite
+++ b/Containerfile.lite
@@ -35,7 +35,8 @@ ARG ENABLE_PROFILING=false
 # To build WITH Rust: docker build --build-arg ENABLE_RUST=true -f Containerfile.lite .
 # To build WITHOUT Rust (default): docker build -f Containerfile.lite .
 ###############################################################################
-FROM quay.io/pypa/manylinux2014:2026.04.08-5 AS rust-builder-base
+FROM registry.access.redhat.com/ubi10/ubi:10.1-1776145136 AS rust-builder-base
+ARG PYTHON_VERSION=3.12
 ARG ENABLE_RUST
 ARG ENABLE_RUST_MCP_RMCP
 
@@ -53,6 +54,23 @@ RUN if [ "$ENABLE_RUST" != "true" ]; then \
         exit 0; \
     fi
 
+RUN if [ "$ENABLE_RUST" = "true" ]; then \
+        dnf upgrade -y && \
+        dnf install -y \
+            python${PYTHON_VERSION} \
+            python${PYTHON_VERSION}-devel \
+            python${PYTHON_VERSION}-pip \
+            gcc \
+            gcc-c++ \
+            openssl-devel \
+            postgresql-devel \
+            libpq-devel \
+            findutils \
+            curl && \
+        update-alternatives --install /usr/bin/python3 python3 /usr/bin/python${PYTHON_VERSION} 1 && \
+        dnf clean all; \
+    fi
+
 # Install Rust toolchain (only if ENABLE_RUST=true)
 RUN if [ "$ENABLE_RUST" = "true" ]; then \
         curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain stable; \
@@ -68,7 +86,7 @@ COPY crates/ /build/crates/
 # Build local native extensions from maturin crates under crates/
 RUN if [ "$ENABLE_RUST" = "true" ]; then \
         mkdir -p /build/native-extension-wheels && \
-        python3.12 -m pip install --no-cache-dir --upgrade pip "maturin==1.12.6" && \
+        python3 -m pip install --no-cache-dir --upgrade pip "maturin==1.12.6" && \
         printf '%s\n' \
             'import pathlib' \
             'import subprocess' \
@@ -92,7 +110,7 @@ RUN if [ "$ENABLE_RUST" = "true" ]; then \
             '    subprocess.run([sys.executable, "-m", "maturin", "build", "--release", "--manifest-path", str(cargo_toml), "--out", str(wheel_dir)], check=True)' \
             'print("✅ Local native extensions built successfully")' \
             > /tmp/build_local_native_extensions.py && \
-        python3.12 /tmp/build_local_native_extensions.py && \
+        python3 /tmp/build_local_native_extensions.py && \
         rm -f /tmp/build_local_native_extensions.py \
     else \
         echo "⏭️  Skipping local native extension build"; \
--- a/tests/unit/test_rust_plugins_workflow.py
+++ b/tests/unit/test_rust_plugins_workflow.py
@@ -157,7 +157,7 @@ def test_containerfile_lite_uses_workspace_runtime_and_native_extensions():
 
     assert "ENABLE_RUST=true but no Rust wheels were produced" not in containerfile
     assert "COPY mcp-servers/rust/ /build/mcp-servers/rust/" not in containerfile
-    assert "FROM quay.io/pypa/manylinux2014:2026.04.08-5 AS rust-builder-base" in containerfile
+    assert "FROM registry.access.redhat.com/ubi10/ubi:10.1-1776145136 AS rust-builder-base" in containerfile
     assert "Rust plugins" not in containerfile
     assert "Skipping Rust plugin build" not in containerfile
     assert "rust-wheels" not in containerfile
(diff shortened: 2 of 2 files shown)


dima-zakharov (dismissed):

dima-zakharov:
### Suggestions (Recommended Improvements)

#### 1. Missing `rust-toolchain.toml` file

**File:** `Cargo.toml:2` (workspace root)

**Issue:** The workspace uses `resolver = "3"` which requires Rust 1.85+, and `mcp_runtime/Cargo.toml` specifies `rust-version = "1.85"`, but there's no `rust-toolchain.toml` file in the repository. The CI workflow references it in path triggers but the file doesn't exist.

**Impact:** Developers and CI may use older Rust versions, causing workspace manifest parsing failures. The `MANIFEST.in` references it but the file is missing.

**Suggested fix:** Add a `rust-toolchain.toml` file:
```toml
[toolchain]
channel = "1.85"
components = ["rustfmt", "clippy"]
```

lucarlig:
Addressed.

Added a workspace-level `rust-toolchain.toml` at the repo root so the workspace and CI stop referring to a missing file.

I used:

```toml
[toolchain]
channel = "stable"
components = ["rustfmt", "clippy"]
```

instead of pinning `1.85`, since the goal here is to stay on stable while still satisfying the workspace minimum Rust version requirement from `rust-version = "1.85"` and `resolver = "3"`.

lucarlig:
Supply-chain update for the Rust workspace CI.

I wired in `cargo vet` as the strict Rust supply-chain gate and took the dependency graph through three stages:

| Stage | What was enabled | Result |
| --- | --- | --- |
| 1 | Public imported audit sets only (`mozilla`, `google`, `bytecode-alliance`, `zcash`, `embark-studios`, `ariel-os`, `isrg`) | `304` unvetted crates |
| 2 | Public imports + blanket maintainer trust for high-coverage ecosystems | reduced to `100` unvetted crates |
| 3 | Temporary exemptions for the remaining unresolved crates | `cargo vet check` passes with `299` fully audited, `12` partially audited, `88` exempted |

A bit more context on stage 2: I added maintainer trust only where it was clearly high leverage across the graph, for example shared ecosystem maintainers around `clap/toml`, `tokio`, `tracing`, `rustls`, `prost/tonic`, `opentelemetry`, `redis`, `rmcp`, Windows crates, and similar families.

That got us a large reduction quickly, but it also has an obvious limitation: blanket maintainer trust is weaker than crate-by-crate audits. It is a pragmatic middle ground for CI hardening, not the end state.

The remaining debt is now explicit in `supply-chain/config.toml` as temporary exemptions. Those `88` exemptions are the non-audited residue we still need to burn down over time by replacing them with one of:

1. more upstream imported audits
2. narrower trust entries where justified
3. local crate audits/certifications

So the current state is:
- Rust CI is enforceable and green
- public audit imports are in place
- high-value trust entries are in place
- the remaining unaudited set is tracked explicitly instead of being silent

lucarlig:
Follow-up on the Rust security tooling split:

I also removed `cargo-audit` from Rust CI.

Reason: `cargo-audit` and `cargo-deny check advisories` both consume the RustSec advisory database for Rust dependency vulnerability scanning, so keeping both as required CI gates was duplicate work and duplicate noise.

The current split is now:
- `cargo-deny`: `advisories`, `bans`, `sources`, `licenses`
- `cargo-vet`: supply-chain trust / audit policy
- `cargo-clippy`: linting

So advisory coverage is still enforced, but it now happens through `cargo-deny` instead of running a second RustSec gate with `cargo-audit`.

.github/workflows/pytest.yml
```diff
@@ -15,19 +15,23 @@ name: Tests & Coverage
 on:
   push:
     branches: ["main"]
-    # paths:
-    #   - "mcpgateway/**"
-    #   - "tests/**"
-    #   - "pyproject.toml"
-    #   - ".github/workflows/pytest.yml"
+    paths:
```
jonpspri: We can't filter for so long as this remains a requirement of EVERY merge.  @brian-hussey I'd like to re-assess how we handle this -- our folks with merge-button access should know enough to look for the test.

.github/workflows/pytest.yml
```diff
@@ -136,23 +138,6 @@ jobs:
             --cov-fail-under=30 \
             --tb=short
 
-      # -----------------------------------------------------------
```
jonpspri, line 139: While I'm tempted to remove doctest, I think we need to do so as part of a broader cleanup initiative.

jonpspri (commented):
Can we double-check the numbering of the ADRs?  Feels like we may have some overlap.

.github/workflows/pytest.yml
lucarlig: commented out

.github/workflows/pytest.yml
lucarlig: restored

lucarlig:
> Can we double-check the numbering of the ADRs? Feels like we may have some overlap.

don't see any overlap in the ADR folder? this PR is just updating some old ADRs not creating new ones with correct paths.

lucarlig:
Addressed.

1. Added docs for the workspace policy:
   - root `Cargo.toml` now has an explicit comment that repo-owned workspace crates live under `crates/` and that the `crates/*` glob is intentional
   - `README.md` now documents that root-level `cargo build` / `cargo test` / `cargo check` cover the shared workspace and that the root `.venv` is reused for PyO3/maturin builds

2. Opened a tracking issue for the remaining `cargo-vet` exemptions and linked it here:
   - #4173 https://github.com/IBM/mcp-context-forge/issues/4173

3. Added a comment in `deny.toml` to make the intent explicit:
   - the allowlist is intentionally broad because it represents the IBM-approved set plus ContextForge-specific extensions
   - it should not be shortened opportunistically just because the current dependency graph only uses a subset

4. Added docs in ADR-041 as well:
   - clarified that workspace-owned crates belong under `crates/`
   - documented that `mcp-servers/rust/` is intentionally outside the shared workspace

5. Updated `CHANGELOG.md` with the Rust workspace consolidation and the follow-up tracking issues

6. Updated ADR-041 to match the current direction:
   - `mcp-servers/rust/` is not part of the shared workspace
   - noted that those sample/test servers can move out later if we give them a plugin-like distribution path

7. Opened a tracking issue for the large Rust workflow follow-up:
   - #4174 https://github.com/IBM/mcp-context-forge/issues/4174

PR #4105, opened 2026-04-09
Scaffold i18n

gcgoncalves:
Introduce i18n using FormatJS


marekdano (changes requested):
Summary

  Introduces a React + TypeScript SPA scaffold under client/, wired up with FormatJS (react-intl) for i18n with three locales (en-US, es-ES, pt-BR), a custom client-side router, and a FastAPI catch-all handler at /app/*. The structural
  decisions are solid, but the PR has a blocking build failure and the Login page — the one actually-functional page — skips the i18n system entirely.

  ---
  Findings

  Blocking

  `client/src/components/ui/button.tsx:6, sidebar.tsx:7` — Missing lib/utils module
  Both files `import cn from ../../lib/utils`, which does not exist. TypeScript compilation fails with error TS2307 (confirmed via tsc --noEmit). Add `client/src/lib/utils.ts` exporting `cn`. The standard shadcn implementation uses clsx +
  tailwind-merge; a zero-dependency alternative is (...classes: (string | false | undefined)[]) => classes.filter(Boolean).join(" ") if those libs aren't wanted.

  `client/src/pages/Login.tsx` — i18n not used on the only functional page
  Login is the only fully-implemented page and has zero i18n usage — every string is hardcoded in English ("Sign in to ContextForge", "Email", "Password", "Forgot password?", "Signing in…"). The auth.json locale files define exactly these
  keys (auth.login.title, auth.login.email, etc.) but they are never consumed. Import useIntl and call intl.formatMessage({ id: "auth.login.title" }) etc. — this is what the PR is supposed to demonstrate.

  `tests/` — No tests for new backend behaviour
  No tests added for app_spa_router (404 when SPA not built, 200 when it is) or for the CSP path-branching logic in security_headers.py (the /app/* vs legacy branch is new and untested). Add at minimum: one test that GET /app/login returns
   404 when index.html is absent, and one verifying the strict CSP is applied on /app/ paths vs the legacy CSP on /admin/.

  ---
  High

  `mcpgateway/main.py:11216–11219` — Static file gate mismatch
  app_spa_router is mounted under `ADMIN_API_ENABLED`, but static files (/static/app/assets/…) are only mounted under `UI_ENABLED`. Both flags default to false but can be set independently. With `ADMIN_API_ENABLED=true`, `UI_ENABLED=false`, the SPA's index.html is served but all JS/CSS assets return 404 — a silent partial failure. Either mount app_spa_router inside the `if UI_ENABLED: block`, or add an explicit startup guard.

  `client/src/components/ui/LanguageSwitcher.tsx:22` — as any on untrusted input
  `setLocale(e.target.value as any)` silences the type system on a value coming from a DOM event. Cast to SupportedLocale instead: setLocale(e.target.value as SupportedLocale). The `<select>` options are already constrained to `SUPPORTED_LOCALES`, so this is safe and explicit.

  ---
  Medium

  `client/src/i18n/IntlProvider.tsx:64–73` — Redundant lang attribute update
  document.documentElement.lang is set twice: once inside setLocale() (line 67) and again inside a useEffect that fires on every locale change (line 72). Remove the useEffect block. The initial lang attribute can be set as a one-time
  side-effect in detectUserLocale() since the locale is synchronously determined before first render.

  `client/src/auth/AuthContext.tsx:44–47` — Stale user: null on page reload
  On reload, isAuthenticated: true (token found in sessionStorage) but user: null. Header.tsx guards user && so there's no crash, but any component that needs user.is_admin or user.email will silently skip rendering until the first
  authenticated API call repopulates user. On mount, fire a GET /auth/me (or equivalent) to rehydrate user, and set isAuthenticated: false on 401.

  `client/src/i18n/IntlProvider.tsx:176` — setLocale not memoised
  setLocale is defined as a plain function inside the component body — it's recreated on every render and passed via context, triggering unnecessary re-renders in consumers. Wrap in useCallback with [] deps.

  ---

  Lint auto-fix applied

  Ran `make autoflake isort black` on the three changed Python files. One change was applied and staged:

  - mcpgateway/main.py: extra space before inline comment removed by black (admin_router)   # → admin_router)  #).

vishu-bh:
The /auth/me rehydration is a nice improvement for reload behavior. One edge case seems worth tightening up here: the catch path only clears auth state on 401. If that request fails for a different reason, like a transient network issue or a 5xx, the token stays in sessionStorage, user remains null, and the app can end up in a half-authenticated state instead of recovering cleanly. It may be worth either treating any rehydration failure as a bootstrap failure and clearing state, or adding an explicit auth-loading state so protected UI waits until /auth/me resolves definitively.


Also, I noticed forgot password option in login. Are we supporting that?

gcgoncalves:
@vishu-bh I agree 100% with all the statements. Let's work on it on a separate PR, though, and keep concerns isolated.

gcgoncalves:
@marekdano is this good to go?

client/src/pages/Login.tsx
marekdano: Will we move the strings "Invalid credentials." and "Login failed" to the i18n files?

marekdano: Will we move the string "An unexpected error occurred." to i18n files.

client/src/lib/utils.ts
```diff
@@ -0,0 +1,9 @@
+type ClassValue = string | undefined | null | false | ClassValue[];
+
+/** Joins class names, filtering falsy values. No external dep needed. */
+export function cn(...classes: ClassValue[]): string {
```
marekdano, line 4: no `tailwind-merge`, conflicting classes not **deduplicated**

  The `cn` implementation joins class names but does not deduplicate conflicting Tailwind utilities. In `button.tsx:44`, a caller passing className="px-2" alongside the base px-4 py-2 will produce "… px-4 py-2 px-2" — Tailwind applies the last declaration, which is browser-order-dependent, not predictable. 
Will we think about adding `tailwind-merge`?

client/src/router/index.tsx
marekdano: `DEFAULT_PUBLIC_PATHS` is the declared allowlist, but `AuthGuard` contains a separate hardcoded `path.startsWith("/app/reset-password/")` check. Adding future public paths now requires editing both the constant and the inline condition. Either add `"/app/reset-password/"` to `DEFAULT_PUBLIC_PATHS` with prefix-match support, or handle it via a `publicPrefixes` prop alongside `publicPaths`.

mcpgateway/main.py
```diff
@@ -11413,6 +11412,10 @@ async def handle_streamable_http(self, scope, receive, send):
             settings.static_dir,
             exc,
         )
+    
+    # Mount React SPA router (requires static files to be available)
+    app.include_router(app_spa_router)  # React SPA at /app/* (no prefix, no CSRF)
+    logger.info("React SPA router mounted at /app/*")
```
marekdano, line 11418: Do we need to update pytest to cover the changes?

mcpgateway/main.py
```diff
@@ -11215,7 +11215,6 @@ async def cleanup_import_statuses(max_age_hours: int = 24, user=Depends(get_curr
 if ADMIN_API_ENABLED:
     logger.info("Including admin_router - Admin API enabled")
     app.include_router(admin_router)   # Admin routes imported from admin.py
-    app.include_router(app_spa_router)  # React SPA at /app/* (no prefix, no CSRF)
```
marekdano, line 11218: Do we need to update pytest to cover this change?

marekdano (commented):
Left some comments

mcpgateway/main.py
gcgoncalves: Let's address at another PR.

mcpgateway/main.py
gcgoncalves: Same as above.

client/src/lib/utils.ts
gcgoncalves: Made a set out of the classes, so that they're unique

marekdano (approved):
LGTM 🚀

PR #3820, opened 2026-03-24
Use config.default_mask_strategy instead of hardcoded values

smrutisahoo10:
## 🔗 Related Issue
Closes #3724 

---

## 📝 Summary

Fixes bug where default_mask_strategy configuration was ignored for built-in PII types in PIIFilterPlugin. The plugin was using hardcoded MaskingStrategy.PARTIAL values instead of respecting the configured strategy.

**Changes:**

- Changed 20+ hardcoded mask_strategy=MaskingStrategy.PARTIAL to mask_strategy=self.config.default_mask_strategy in _compile_patterns() method

**Impact:**
- Users can now control PII masking behavior via configuration (redact, partial, hash, tokenize, remove) instead of being forced to use hardcoded partial masking.

---

## 🏷️ Type of Change
- [X] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Chore (deps, CI, tooling)
- [ ] Other (describe below)

---

## 🧪 Verification

| Check                     | Command         | Status |
|---------------------------|-----------------|--------|
| Lint suite                | `make lint`     |     Pass   |
| Unit tests                | `make test`     |    Pass    |
| Coverage ≥ 80%            | `make coverage` |    Pass    |

---

## ✅ Checklist
- [X] Code formatted (`make black isort pre-commit`)
- [ ] Tests added/updated for changes
- [ ] Documentation updated (if applicable)
- [ ] No secrets or credentials committed

---

## 📓 Notes (optional)
_Screenshots, design decisions, or additional context._

PR #4121, opened 2026-04-10
fix: honor default mask strategy in Rust PII detector

### LINKED ISSUE #4120: [BUG]: Rust PII detector ignores default_mask_strategy for built-in patterns
## Summary
The Rust-backed `PIIFilterPlugin` path still hardcodes per-type masking strategies for built-in PII patterns, so `default_mask_strategy` is ignored when Rust is available.

## Impact
- Plugin behavior differs between Python and Rust implementations
- The merged Python-side fix in PR #3820 does not fully fix runtime behavior because the plugin prefers Rust when available
- Users cannot reliably configure built-in PII masking to `redact`, `hash`, `tokenize`, `remove`, or `partial` through one global setting

## Root Cause
`plugins_rust/pii_filter/src/patterns.rs` stores explicit mask strategies for built-in patterns. Detection then always uses those explicit values instead of falling back to `config.default_mask_strategy`.

## Proposed Fix
- Make built-in Rust patterns inherit `config.default_mask_strategy`
- Keep explicit per-pattern overrides only for custom patterns
- Add parity regression tests covering built-in detections, masking output, and nested processing behavior

## Verification
- `uv run pytest -q tests/unit/mcpgateway/plugins/plugins/pii_filter/test_pii_filter.py`
- `cargo test` in `plugins_rust/pii_filter`

## Related
- PR #3820
- Follow-up PR to be opened from `fix/pii-filter-rust-default-mask-strategy`

lucarlig:
## Summary
This follow-up fixes the Rust-backed PII detector so built-in patterns honor `default_mask_strategy`, matching the Python implementation and the documented plugin configuration.

## What Changed
- Changed built-in Rust pattern definitions to inherit `config.default_mask_strategy`
- Kept explicit mask-strategy overrides for custom patterns
- Updated Rust unit tests to verify the new fallback behavior
- Added Python-side regression coverage for Rust parity, including nested-processing cases that now set `PARTIAL` explicitly when that behavior is under test

## Root Cause
The Rust implementation stored explicit per-pattern mask strategies for built-in PII types. Detection therefore never fell back to `default_mask_strategy`, so the global configuration was ignored whenever the Rust detector was active.

## Validation
- `uv run pytest -q tests/unit/mcpgateway/plugins/plugins/pii_filter/test_pii_filter.py`
- `cargo test` in `plugins_rust/pii_filter`

Closes #4120
Related to #3820

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "^.secrets.baseline$|package-lock.json|Cargo.lock|scripts/sign_image.sh|scripts/zap|sonar-project.properties|uv.lock",
     "lines": null
   },
-  "generated_at": "2026-04-08T16:53:59Z",
+  "generated_at": "2026-04-10T08:19:59Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -8604,7 +8604,7 @@
         "hashed_secret": "25910f981e85ca04baf359199dd0bd4a3ae738b6",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 505,
+        "line_number": 506,
         "type": "AWS Access Key",
         "verified_result": null
       }
--- a/plugins_rust/pii_filter/src/detector.rs
+++ b/plugins_rust/pii_filter/src/detector.rs
@@ -719,7 +719,7 @@ mod tests {
     }
 
     #[test]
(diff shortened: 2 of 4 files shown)


lucarlig:
Closing this draft follow-up here. The relevant Rust PII plugin lives in `IBM/cpex-plugins`, so the issue and fix PR are being moved there instead.

PR #4136, opened 2026-04-13
fix(transport): skip outputSchema validation for isError tool responses

### LINKED ISSUE #4135: [BUG]: outputSchema validation not skipped for isError:true tool results, hiding actual error messages
### 🐞 Bug Summary

When an upstream MCP server returns a tool execution error (isError: true), the gateway still attempts to validate the error response against the tool's outputSchema. Since error responses naturally don't conform to the success schema, the original error message (e.g., "Workflow is not available in MCP") is replaced with a generic "Output validation error: outputSchema defined but no structured output returned", hiding the actual error cause from the user.

---

### 🧩 Affected Component

- [x] `mcpgateway` - API
- [ ] Other (explain below)

---

### 🔁 Steps to Reproduce

1. Register an upstream MCP server that defines outputSchema on its tools (e.g., n8n Instance-level MCP server)
2. Call a tool that returns an error response with isError: true and a text content error message — for example, calling get_workflow_details on a workflow that has MCP access disabled
3. The upstream server responds with:
```json
  {
    "result": {
      "content": [{"type": "text", "text": "Workflow is not available in MCP. Enable MCP access in workflow settings."}],
      "isError": true
    }
  }
```
4. Instead of passing through this error message, the gateway returns: "Output validation error: outputSchema defined but no structured output returned"

---

### 🤔 Expected Behavior

When isError: true is set in the tool result, outputSchema validation should be skipped entirely and the original error content should be passed through to the client as-is. Per the MCP spec, outputSchema defines the expected structure for successful tool results, not error responses.

---

### 📓 Logs / Error Output

```
  # Gateway logs show MCP call succeeds but tool invocation is marked as failed:
  [tool_service] MCP tool call completed: get_workflow_details  (success: true)
  [tool_service] Tool 'n8n-get-workflow-details' invocation failed  (error_message: null)
```

The issue is in mcpgateway/services/tool_service.py, method _extract_and_validate_structured_content() (around line 1111). The method checks for output_schema presence but does not check tool_result.is_error before proceeding with validation.

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Version or commit | `latest main` |
| Runtime | `Python3.12` |
| Platform / OS | `Linux` |
| Container | `Docker` |

---

- This affects any upstream MCP server that defines outputSchema and can return tool execution errors.

kimsehwan96:
## 🔗 Related Issue
Closes #4135 

---

## 📝 Summary

When an upstream MCP tool with outputSchema defined returns an error (isError: true) without structuredContent, the MCP Python SDK's server-side validation replaces the original error message with a generic "outputSchema defined but no structured output returned". 

This hides the actual error cause from the user.

This PR bypasses the SDK validation by returning a CallToolResult(isError=True) directly when is_error is True, which triggers the SDK's early-return path and skips outputSchema validation. This uses the same pattern already established in the codebase (lines 1451, 1468).

The root cause is in the MCP Python SDK v1.x (mcp/server/lowlevel/server.py), where outputSchema validation runs unconditionally regardless of isError.

---

## 🏷️ Type of Change
- [x] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

| Coverage ≥ 80%            | `make coverage` |   pass     |

---

## ✅ Checklist

---

## 📓 Notes (optional)

Implementation detail

The fix is placed after the structuredContent check in streamablehttp_transport.py:call_tool():

1. if structured → return (unstructured, structured)   # SDK validates against outputSchema
2. if is_error is True → return CallToolResult(isError=True)  # SDK early-return, no validation
3. return unstructured                                  # normal path

This ordering is intentional: if an error response does include structuredContent, the existing validation path is preserved. We have not observed any MCP server returning both isError: true and structuredContent simultaneously, and the MCP spec does not define this combination. If it does occur, validating the structured content is arguably the correct behavior — the server explicitly provided it, so schema conformance should be checked.

is_error is True (strict identity check)

We use getattr(result, "is_error", None) with is True rather than a truthy check. This prevents false positives when result is a MagicMock in unit tests (where unset attributes return truthy MagicMock objects).

Manual E2E verification

To validate this fix end-to-end, we added a lookup_calculation tool to the existing output_schema_test_server that raises ToolError for unknown IDs (producing isError: true with outputSchema defined). Using the docker-compose stack:

- Before fix: "Output validation error: outputSchema defined but no structured output returned" ❌
- After fix: "Calculation 'nonexistent-id' not found. Only 'calc-001' is available." ✅

The test server change is not included in this PR — unit tests provide sufficient coverage for the transport-layer logic. The E2E setup can be formalized in a follow-up if needed.

### DIFF
--- a/mcpgateway/transports/streamablehttp_transport.py
+++ b/mcpgateway/transports/streamablehttp_transport.py
@@ -1683,6 +1683,14 @@ async def call_tool(name: str, arguments: dict) -> Union[
             if structured:
                 return (unstructured, structured)
 
+            # When the upstream tool returned an error (isError: true), return a
+            # CallToolResult directly so the MCP SDK skips outputSchema validation.
+            # Without this, the SDK replaces the original error message with
+            # "outputSchema defined but no structured output returned".
+            is_error = getattr(result, "is_error", None)
+            if is_error is True:
+                return types.CallToolResult(content=unstructured, isError=True)
+
             return unstructured
     except Exception as e:
         logger.exception("Error calling tool '%s': %s", name, e)
--- a/tests/unit/mcpgateway/transports/test_streamablehttp_transport.py
+++ b/tests/unit/mcpgateway/transports/test_streamablehttp_transport.py
@@ -13550,3 +13550,142 @@ async def test_handle_streamable_http_initialize_span_exits_with_exception(monke
     assert exc_type is ValueError
     assert isinstance(exc_val, ValueError)
     assert str(exc_val) == "initialize failed"
+
+
+# ═══════════════════════════════════════════════════════════════════════════
+# isError + outputSchema validation bypass
+# ═══════════════════════════════════════════════════════════════════════════
+
+
+@pytest.mark.asyncio
+async def test_call_tool_is_error_returns_call_tool_result(monkeypatch):
+    """When tool result has is_error=True, call_tool should return CallToolResult(isError=True)
+    directly so the MCP SDK skips outputSchema validation."""
+    # First-Party
+    from mcpgateway.transports.streamablehttp_transport import call_tool, tool_service, types
+
+    mock_db = MagicMock()
+    mock_result = MagicMock()
+    mock_content = MagicMock()
+    mock_content.type = "text"
+    mock_content.text = "Upstream tool error: resource not found"
+    mock_content.annotations = None
+    mock_content.meta = None
+    mock_result.content = [mock_content]
+    mock_result.is_error = True  # Explicitly set to bool True
+    mock_result.structured_content = None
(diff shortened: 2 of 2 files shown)


ja8zyjits:
Hi @kimsehwan96 ,
Would you like to rebase this?

kimsehwan96:
Hi @ja8zyjits 

I think the PR #4202 already fixed this issue #4135. 

Feel free to close this PR.

ja8zyjits:
Closing based on previous comment

PR #4139, opened 2026-04-13
fix(api): support multipart/form-data and form-urlencoded for REST tool invocations

### LINKED ISSUE #4058: [BUG]: REST tool invocation hardcodes JSON encoding — no support for multipart/form-data or form-urlencoded
### 🐞 Bug Summary
REST tools always send request bodies as `application/json` regardless of what the upstream API requires. There is no way to configure `multipart/form-data` or `application/x-www-form-urlencoded` encoding, so any API that requires form-based encoding returns a 400 Bad Request.

---

### 🧩 Affected Component

- [x] `mcpgateway` - API

---

### 🔁 Steps to Reproduce

1. Register a REST tool pointing to an API that requires `multipart/form-data` (e.g. uses form fields, not a JSON body)
2. Invoke the tool via the UI or MCP protocol
3. Observe a 400 Bad Request from the upstream API

---

### 🤔 Expected Behavior
The tool should allow configuring the request body encoding (JSON, multipart/form-data, or form-urlencoded) so it can interoperate with APIs that do not accept JSON.

---

### 📓 Logs / Error Output
```
"error_message": "Client error '400 Bad Request' for url '...'",
"error_stack_trace": "Exception: Client error '400 Bad Request'..."
```

Root cause: `tool_service.py` line ~4109 hardcodes `json=payload` for all non-GET REST requests:
```python
response = await asyncio.wait_for(
    self._http_client.request(method, final_url, json=payload, headers=headers),
    timeout=effective_timeout
)
```

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Container | Docker / none |

---

A working curl equivalent uses `-F` flags (multipart/form-data). The fix would be to detect the intended encoding — either from the tool's configured `Content-Type` header or a new `request_body_type` field — and use httpx's `files=` (multipart), `data=` (form-urlencoded), or `json=` accordingly.

madhav165:
## 📌 Summary
REST tools always sent request bodies as `application/json` regardless of what the upstream API required. Any API expecting `multipart/form-data` or `application/x-www-form-urlencoded` would return 400 Bad Request with no way to configure the encoding.

## 🔁 Reproduction Steps
Closes #4058

1. Register a REST tool pointing to an API that requires `multipart/form-data` (e.g. uses `-F` form fields)
2. Set `Content-Type: multipart/form-data` in the tool's headers
3. Invoke the tool — previously returned 400 Bad Request

## 🐞 Root Cause
`tool_service.py` hardcoded `json=payload` for all non-GET REST requests, causing httpx to always serialize the body as JSON and override any user-configured `Content-Type` header with `application/json`.

## 💡 Fix Description
After all header modifications (auth merge, passthrough, plugin hooks, header mapping), inspect the final `Content-Type` header and dispatch to the appropriate httpx body encoding:

- `application/x-www-form-urlencoded` → `data=` (scalars stringified, nested values JSON-encoded)
- `multipart/form-data` → `files=` with `Content-Type` stripped so httpx sets the correct `boundary=` parameter
- Anything else (including unset or `application/json`) → `json=` — unchanged default, no behaviour change for existing tools

No DB/schema/migration changes — behaviour is driven entirely by the `Content-Type` header already stored on the tool.

## 🧪 Verification

| Check                             | Command                                                                                  | Status |
|-----------------------------------|------------------------------------------------------------------------------------------|--------|
| Lint suite                        | `make lint`                                                                              | ✅     |
| Unit tests                        | `uv run pytest tests/unit/mcpgateway/services/test_tool_service.py -v`                  | ✅ 337/337 |
| New encoding tests                | `uv run pytest tests/unit/mcpgateway/services/test_tool_service.py -k "rest_post" -v`   | ✅ 4/4  |
| Manual regression no longer fails | Register a REST tool with `Content-Type: application/x-www-form-urlencoded`, invoke it — response `form` field contains parameters instead of 400 | ✅ |

## ✅ Checklist

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-04-17T07:49:43Z",
+  "generated_at": "2026-04-17T11:30:25Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-04-17T11:30:25Z",
+  "generated_at": "2026-04-17T13:24:41Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-04-27T09:51:05Z",
+  "generated_at": "2026-04-17T07:49:43Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -8600,7 +8600,7 @@
         "hashed_secret": "9fb7fe1217aed442b04c0f5e43b5d5a7d3287097",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 3206,
+        "line_number": 3278,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -8608,7 +8608,7 @@
         "hashed_secret": "72cb70dbbafe97e5ea13ad88acd65d08389439b0",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 3834,
+        "line_number": 3906,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -8616,7 +8616,7 @@
         "hashed_secret": "ee977806d7286510da8b9a7492ba58e2484c0ecc",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 7112,
+        "line_number": 7184,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -8624,7 +8624,7 @@
         "hashed_secret": "f2e7745f43b0ef0e2c2faf61d6c6a28be2965750",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 7604,
+        "line_number": 7676,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -8632,7 +8632,7 @@
         "hashed_secret": "4a249743d4d2241bd2ae085b4fe654d089488295",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 8951,
+        "line_number": 9023,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -8640,7 +8640,7 @@
         "hashed_secret": "0c8d051d3c7eada5d31b53d9936fce6bcc232ae2",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 9093,
+        "line_number": 9165,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -8648,7 +8648,7 @@
         "hashed_secret": "f2b14f68eb995facb3a1c35287b778d5bd785511",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 9469,
+        "line_number": 9541,
         "type": "Secret Keyword",
         "verified_result": null
       }
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-04-17T13:24:41Z",
+  "generated_at": "2026-04-27T12:40:08Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -4844,16 +4844,8 @@
         "hashed_secret": "c377074d6473f35a91001981355da793dc808ffd",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 701,
+        "line_number": 753,
         "type": "Hex High Entropy String",
-        "verified_result": null
-      },
-      {
-        "hashed_secret": "9d4e1e23bd5b727046a9e3b4b7db57bd8d6ee684",
-        "is_secret": false,
-        "is_verified": false,
-        "line_number": 1103,
-        "type": "Basic Auth Credentials",
         "verified_result": null
       }
     ],
@@ -8568,7 +8560,7 @@
         "hashed_secret": "d63b39580934e062f89aae63426d2f2c77c3e258",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 544,
+        "line_number": 549,
         "type": "Base64 High Entropy String",
         "verified_result": null
       },
@@ -8576,7 +8568,7 @@
         "hashed_secret": "586a55a9b8b97f0cd88e24ce8279ebc955949688",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 545,
+        "line_number": 550,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -8584,7 +8576,7 @@
         "hashed_secret": "00cafd126182e8a9e7c01bb2f0dfd00496be724f",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 561,
+        "line_number": 566,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -8592,7 +8584,7 @@
         "hashed_secret": "7b1552c7c7ffb8bd70b5666e5997c8e017630aab",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1976,
+        "line_number": 1985,
         "type": "Base64 High Entropy String",
         "verified_result": null
       },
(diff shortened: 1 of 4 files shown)


marekdano (changes requested):
@madhav165 - the PR looks good, just a few issues

#### Summary

  Fixes issue #4058 by detecting the final `Content-Type` header after all auth/plugin/mapping transformations and dispatching to the appropriate httpx encoding (`data=, files=, or json=`). The core fix is correct, backward-compatible, and the happy-path test coverage is solid.

####  Findings

- `mcpgateway/services/tool_service.py:4283` — Design: nested function definition
  `_to_str` is defined inside the method body, creating a new function object on every code-path execution. Extract it as a `@staticmethod _form_value_to_str(v: Any) -> str on ToolService`.

- `mcpgateway/services/tool_service.py:4295–4309` — Correctness: undocumented query param behavioral asymmetry
  The JSON/default path merges URL query params into the request body (`payload.update(query_params)`). The form-urlencoded and multipart paths forward them as `params=_qp` (query string). This is the more semantically correct behavior for form encoding, but it is a silent behavioral difference for any tool that switches content types. Add an inline comment explaining the intentional asymmetry.

- `mcpgateway/services/tool_service.py:4274` — Consistency: opaque variable name
  `_qp` is not self-explanatory. Rename to _url_query_params or inline the ternary directly at its two call sites.

- Test coverage: GET + form-related `Content-Type`
  No test guards against a GET request where `Content-Type: application/x-www-form-urlencoded` is set. The GET branch correctly wins in the dispatch, but there is no regression guard for this.

- Test coverage: `tool_query_mapping` + form-urlencoded
  No test verifies that when `tool_query_mapping` is set, `_qp` is `None` and the mapped payload is sent in the form body rather than forwarded as query string params.

marekdano (changes requested):
The core fix is well-structured and the test coverage is solid. A few security-relevant items worth addressing before merge, ranging from a meaningful behavioral issue to lower-priority correctness gaps.

---

## Blocking

**`Content-Type` is not in the sensitive header blocklist, enabling encoding dispatch manipulation via `header_mapping`**

The new dispatch logic routes requests to `data=`, `files=`, or `json=` based on `_ct_base`, which is read from the **final, post-mapping** `headers` dict. Because `Content-Type` is not blocked by `_is_sensitive_tool_header_name()` (see `_SENSITIVE_TOOL_HEADER_PATTERNS` in `tool_service.py`), an admin can configure a tool with `header_mapping: { "encoding": "Content-Type" }`. A caller with only tool-invoke rights could then pass `{ "encoding": "application/x-www-form-urlencoded" }` and silently switch the upstream request from JSON to form-encoded — or vice versa.

This may not be an issue in practice (it requires an admin to create a Content-Type mapping, which is an unusual configuration), but it creates a surprising attack surface: the encoding path becomes caller-controllable, which is almost certainly unintentional. Worth a targeted decision here:

**Option A (minimal):** Add `content-type` to `_SENSITIVE_TOOL_HEADER_PATTERNS` so `header_mapping` cannot target it.

**Option B (defense-in-depth):** Determine `_ct_base` from the tool's *statically configured* headers before mapping is applied, not after. This is a stronger guarantee but a larger change.

A deny-path test covering this scenario would be valuable regardless.

---

## Should Fix

**Boolean serialization in `_form_value_to_str` produces `"True"`/`"False"`, not `"true"`/`"false"`**

```python
return str(v)  # str(True) == "True"
```

For form APIs that inspect boolean flag fields (e.g., `admin=true`, `enabled=true`), this will silently fail with the capitalized variant. `orjson.dumps` already handles booleans correctly:

```python
if isinstance(v, (dict, list, bool)):
    return orjson.dumps(v).decode()  # produces "true" / "false"
```

There is currently no test asserting `_form_value_to_str(True)` behaviour.

---

**`_url_query_params` initialization doesn't account for `has_header_mapping`**

```python
_url_query_params = query_params if not tool_query_mapping else None
```

When `has_header_mapping=True` and `has_query_mapping=False`, the mapping block (lines 4685–4707) already strips and processes URL query params into `query_params`. The form branch then receives that processed dict as `params=` via `_url_query_params`, but the URL-stripping block below is skipped (guarded by `not has_header_mapping`). The result is that processed URL params could be forwarded as query string parameters on form-encoded requests — a subtle double-send. The initialization should mirror the extraction guard:

```python
_url_query_params = query_params if not tool_query_mapping and not tool_header_mapping else None
```

This case is also untested.

---

## Suggestions

Please feel free to fix the following in this PR or in a newly created one.

**`multipart/form-data`: all fields are sent as file attachments**

Using `files=` for every field sends `Content-Disposition: form-data; filename=""` on each part. For APIs that expect standard form fields alongside actual file uploads, this will break. The standard httpx pattern is `data=` for plain fields and `files=` only for file parts. For a v1 fix this is acceptable, but worth a comment and a known-limitations note so future contributors don't have to rediscover it.

**`parse_qs(..., v[0])` silently drops duplicate query parameters**

Both the existing code (lines 4688, 4720) and the new form path use `v[0]` to take only the first value of repeated keys. This is consistent, but for signed URLs (AWS presigned, Azure SAS) that sometimes include repeated params for multi-value filters, silent truncation could corrupt the signature. A logged warning here similar to the GET conflict warning (lines 4722–4728) would help debugging.

**No size guard on `orjson.dumps(v)` for nested values**

A caller could pass a deeply nested dict or very large list as a form field value, causing `orjson.dumps` to allocate significant memory before the request is sent. Existing payload size limits upstream may already cover this, but it is worth verifying — and if not, a max-depth or max-size check before serialization would be prudent.

madhav165:
Addressed the following from the second review:

**Boolean serialization** — Fixed. Added `bool` to the `isinstance` check in `_form_value_to_str` so `True`/`False` go through `orjson.dumps()` and produce `"true"`/`"false"` instead of `"True"`/`"False"`.

**Content-Type in sensitive header blocklist** — Fixed. Added `content-type` to `_SENSITIVE_TOOL_HEADER_PATTERNS` (Option A from the review). This prevents callers from manipulating encoding dispatch via `header_mapping`.

**`_url_query_params` and `has_header_mapping`** — After tracing all branches, the current code is correct and the suggested change would be a regression. Here's why:

When `has_header_mapping=True` and `has_query_mapping=False`:
1. Lines 4685-4688: URL is stripped and `query_params` is populated (because the mapping block fires)
2. Line 4711: `_url_query_params = query_params` (because `tool_query_mapping` is falsy)
3. Line 4722: The form extraction block is skipped (`not has_header_mapping` is false) — correct, since the URL was already stripped at step 1
4. Lines 4756/4763: `params=_url_query_params` forwards the extracted URL params on the query string

Adding `and not tool_header_mapping` to line 4711 would set `_url_query_params = None`, causing the URL params (already stripped from the URL at step 1) to be silently dropped for form-encoded requests. The JSON path at line 4770 would still pick them up via `payload.update(query_params)`, but the form/multipart paths would lose them entirely.

Header mapping is orthogonal to URL query param forwarding — it maps payload values into HTTP headers, it doesn't affect how URL query params should be dispatched.

**`multipart/form-data` fields as file attachments** — After testing, the current code is actually fine here. The `files={'field': (None, value)}` pattern with a `None` filename causes httpx to emit `Content-Disposition: form-data; name="field"` with no `filename=""` present. This is a proper form field, not a file attachment. Only when you pass an actual filename string (e.g. `('report.pdf', content)`) does httpx add `filename=`. So no fix needed for this one.

**`parse_qs` `v[0]` dropping duplicates** — Pre-existing pattern, out of scope for this PR.

Commit: b2fbcca19

marekdano (dismissed):
## Summary

Adds `application/x-www-form-urlencoded` and `multipart/form-data` encoding support to REST tools, moving beyond hardcoded JSON. The implementation is correct, backward-compatible, and security-aware.

---

## Resolved Items

| Item | Status |
|------|--------|
| Boolean serialization (`true`/`false`) | ✅ Fixed — `bool` added to `isinstance` check in `_form_value_to_str`, now uses `orjson.dumps()` |
| Content-Type in sensitive header blocklist | ✅ Fixed — `content-type` added to `_SENSITIVE_TOOL_HEADER_PATTERNS` (Option A) |
| multipart `files={'field': (None, value)}` | ✅ Correct — httpx emits no `filename=` with `None`; proper form field behaviour |
| `parse_qs` `v[0]` dropping duplicates | ✅ Pre-existing pattern, out of scope |

---

## `_url_query_params` / `has_header_mapping` — Author Is Right

After tracing the logic, the author's explanation is accurate:

```
Scenario: has_header_mapping=True, has_query_mapping=False
1. Lines 4685–4688: URL stripped → query_params populated
2. Line 4711: _url_query_params = query_params  ✓  (tool_query_mapping is falsy)
3. Line 4718 extraction block: skipped (has_header_mapping=True)  ✓
4. Lines 4756/4763 form branches: params=_url_query_params  ✓
```

Adding `and not tool_header_mapping` to line 4711 **would be a regression** — it would set `_url_query_params = None` and silently drop URL params from form/multipart requests when only header mapping is configured. Header mapping is orthogonal to URL query param forwarding.

---

## Suggestions (Non-blocking)

1. **Add a comment at line 4711** to document the invariant and prevent future regression:
   ```python
   # NB: guard on tool_query_mapping, not has_header_mapping. When only
   # header_mapping is present, URL params were already extracted into
   # query_params above and must still be forwarded on the query string.
   ```

2. **Add one test** for `header_mapping`-only (no `query_mapping`) to lock in the subtle behaviour:
   ```python
   # URL has ?key=value, tool has header_mapping only
   # Assert final request has params={'key': 'value'} on query string
   ```

---

## Security

- **Content-Type blocking:** Effective — prevents encoding dispatch manipulation via `header_mapping`
- **`_form_value_to_str()`:** No injection vectors — all types serialized as strings, not interpreted
- **Signed URL preservation:** Correct — form paths forward URL params via `params=`, not merged into body

---

## Test Coverage

- ✅ Form-urlencoded POST with nested values
- ✅ Multipart form-data with various types
- ✅ Query parameter forwarding for both encodings
- ✅ Interaction with `query_mapping`
- ✅ Interaction with `header_mapping`
- ✅ Edge case: GET request with form Content-Type (correctly stays GET)

marekdano (dismissed):
All blocking issues resolved, implementation correct and secure

LGTM 🚀

brian-hussey (approved):
Approved based on previous approvals.

brian-hussey:
Tested the failing tests locally. The failing tests are since several PRs ago.
Merging this, will get other failing tests corrected on main.

PR #4146, opened 2026-04-13
fix: use POST ping for StreamableHTTP health checks and unauthenticated liveness probes

### LINKED ISSUE #4154: [BUG][TRANSPORT]: StreamableHTTP health checks return 405 — full SDK initialize() incompatible with health probing
## Summary

StreamableHTTP servers return **405 Method Not Allowed** during health checks because the gateway uses a full MCP SDK `initialize()` call which sends HTTP methods the server does not accept for health probing. Additionally, servers behind `authorization_code` OAuth gateways fail health checks because the gateway has no user-delegated token.

## Reproduction Steps

1. Register a remote MCP server using the StreamableHTTP transport
2. Observe the health check cycle in the gateway logs
3. Health checks return 405 errors, marking the server as permanently unhealthy
4. For `authorization_code` gateways: health checks fail with 401/403 even when the server is up

## Root Cause

1. `_check_streamablehttp_health()` creates a full `streamablehttp_client()` session and calls `session.initialize()`, which sends MCP protocol handshake messages. Many StreamableHTTP servers only accept POST with specific JSON-RPC payloads, returning 405 for the SDK's approach.
2. For `authorization_code` grant type gateways, the health check runs as the gateway service account which has no user-delegated OAuth token.

## Expected Behavior

- Health checks should use a lightweight JSON-RPC `ping` POST request instead of a full SDK handshake
- A 200 response (or valid JSON-RPC error) should mark the server as healthy
- For `authorization_code` gateways, health checks should be skipped or use a dedicated probe endpoint

## Impact

- StreamableHTTP servers are permanently marked as unhealthy despite being fully operational
- Tools from these servers are never discovered or remain stale
- Operators see false-positive health alerts

ecthelion77:
# 🐛 Bug-fix PR

## 🔗 Issue

Closes #4154

## 📌 Summary

StreamableHTTP servers return 405 Method Not Allowed during health checks because the gateway uses a full MCP SDK client `initialize()` call which sends HTTP methods the server does not accept for health probing. Additionally, servers configured with `authorization_code` OAuth grant fail liveness checks because the gateway has no user token to authenticate with.

## 🔁 Reproduction Steps

1. Register a remote MCP server using the StreamableHTTP transport
2. Observe the health check cycle in the gateway logs
3. Health checks return 405 errors, marking the server as unhealthy
4. For `authorization_code` gateways: health checks fail with 401/403 even when the server is up

## 🐞 Root Cause

1. The `_check_streamablehttp_health()` method creates a full `streamablehttp_client()` session and calls `session.initialize()`, which sends MCP protocol handshake messages. Many StreamableHTTP servers only accept POST with specific JSON-RPC payloads, returning 405 for the SDK's approach.
2. For `authorization_code` grant type gateways, the health check runs as the gateway service account, which has no user-delegated OAuth token. The server correctly rejects the unauthenticated request.

## 💡 Fix Description

1. **POST-based ping**: Replace the full SDK `initialize()` call with a lightweight `POST` request sending a JSON-RPC `ping` method. This is the standard MCP liveness check that StreamableHTTP servers accept. A 200 response (or even a valid JSON-RPC error response) confirms the server is alive.
2. **Unauthenticated probe for authorization_code gateways**: When the server's gateway uses `authorization_code` grant type, skip OAuth token injection and send the ping without credentials. A 401/403 response is treated as "server is alive" (it responded), while connection errors or timeouts indicate the server is truly down.

## 🧪 Verification

| Coverage ≥ 80 %                       | `make coverage`      | ✅     |
| Manual regression no longer fails     | StreamableHTTP servers show healthy status | ✅ |

- [x] Matches current MCP spec (JSON-RPC ping is a standard MCP method)

## ✅ Checklist

### DIFF
--- a/mcpgateway/services/gateway_service.py
+++ b/mcpgateway/services/gateway_service.py
@@ -3571,6 +3571,12 @@ class GatewayService(BaseService):  # pylint: disable=too-many-instance-attribut
             # Use isolated client for gateway health checks (each gateway may have custom CA cert)
             # Use admin timeout for health checks (fail fast, don't wait 120s for slow upstreams)
             # Pass ssl_context if present, otherwise let get_isolated_http_client use skip_ssl_verify setting
+            # Track whether this is an unauthenticated liveness probe
+            # (authorization_code gateway without system token).  When True,
+            # HTTP 401/403 responses are treated as "server alive" rather
+            # than health-check failures.
+            unauthenticated_probe = False
+
             async with get_isolated_http_client(timeout=settings.httpx_admin_read_timeout, verify=ssl_context) as client:
                 logger.debug(f"Checking health of gateway: {gateway_name} ({gateway_url_sanitized})")
                 try:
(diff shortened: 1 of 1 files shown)


ecthelion77:
**Suggested labels:** `bug`, `SHOULD`, `python`, `mcp-protocol`

ecthelion77:
Closing — this fix is now covered by upstream PR #4284 (merged), which returns 405 on session-less GET /mcp. Thank you!

PR #4160, opened 2026-04-13
fix: handling login in UI rewrite

marekdano:
# 🐛 Bug-fix PR

---

## 📌 Summary
When we run `make serve` and `cd client && npm install && npm run dev` the login page POST request returns 404.
We need to add configuration to vite.config.ts to fix this issue

## 🔁 Reproduction Steps
1. run `make serve` 
2. in other window run `cd client && npm install && npm run dev`
3. Open app in the http://localhost:5173/ and try to login to the app


marekdano:
Closing the issue. In this case we should call UI from `http://localhost:4444/app` and not from `http://localhost:5173/`

PR #4164, opened 2026-04-14
fix: use `is None` instead of `== None` in migration script

harshadkhetpal:
## Summary

Replaces `== None` / `!= None` with `is None` / `is not None` in `scripts/verify_multitenancy_0_7_0_migration.py`.

## Motivation

[PEP 8](https://peps.python.org/pep-0008/#programming-recommendations) specifies that comparisons to `None` should use identity comparison. This is flagged as [E711](https://www.flake8rules.com/rules/E711.html) by most linters.

## Change

```python
# Before
if value == None:
    ...

# After
if value is None:
    ...
```

## Testing

Style/correctness fix only; no behavior change.


jonpspri:
This script is about to be deprecated.  Closing the PR.

Issue #4208, opened 2026-04-14
[BUG]: outputSchema declared but no structured output silently passes gateway validation


jonpspri:
### 🐞 Bug Report

`ToolService._extract_and_validate_structured_content` (`mcpgateway/services/tool_service.py:1340-1347`) returns `True` ("valid") when all of the following are true:

1. `is_error=False`.
2. The tool declares an `output_schema`.
3. `structured_content` is `None` **and** no JSON object can be parsed from any `TextContent` item in `content` (empty `content`, a non-text item, or unparseable text).

### 📜 Spec reference

The MCP 2025-11-25 specification, section **"Output Schema"** (https://modelcontextprotocol.io/specification/2025-11-25/server/tools#output-schema), states:

> If an output schema is provided: Servers **MUST** provide structured results that conform to this schema.

The current gateway behaviour is documented at `tool_service.py:1346-1347` as:

```python
# If no structured data found, treat as valid (nothing to validate)
if structured is None:
    return True
```

This is lenient by design, but it silently masks upstream servers that violate the spec — the caller receives a `success` shape even though the declared `outputSchema` was not satisfied.

### 🔁 Reproduction (unit test already in repo)

```python
tool_result = ToolResult(content=[], is_error=False)
tool = SimpleNamespace(
    name="test_tool",
    output_schema={"type": "object", "required": ["recognitionId"], "properties": {"recognitionId": {"type": "string"}}},
)
assert tool_service._extract_and_validate_structured_content(tool, tool_result) is True  # ← spec says this should fail
```

See `tests/unit/mcpgateway/services/test_tool_service_coverage.py::TestExtractAndValidateErrorResponses::test_success_with_schema_but_no_content_currently_passes` for the regression guard pinning the current (lenient) behaviour.

### ✅ Proposed fix

When `is_error=False` and `output_schema` is declared, treat "no structured payload" as a validation failure equivalent to the existing non-dict `structured_content` branch at `tool_service.py:1314-1327` — set `is_error=True` on the result and replace `content` with a validation-error details dict including:

```json
{
  "code": "missing_structured_output",
  "expected": "object",
  "received": "null",
  "message": "outputSchema declared but no structuredContent was returned"
}
```

This aligns with the MCP Python SDK's server-side validator (`mcp/server/lowlevel/server.py:560-564`), which produces the analogous message `"Output validation error: outputSchema defined but no structured output returned"`.

### 🧪 Tests to add / update

[...]

Issue #4202, opened 2026-04-14
[BUG]: MCP ToolError responses fail output schema validation


ja8zyjits:
### 🐞 Bug Summary
ContextForge Gateway unconditionally validates MCP tool responses against `outputSchema`, including error responses with `isError: true`. This violates the MCP specification and causes validation failures for standards-compliant error handling, replacing original error messages with validation errors.

---

### 🧩 Affected Component
Select the area of the project impacted:

- [x] `mcpgateway` - API
- [ ] `mcpgateway` - UI (admin panel)
- [ ] `mcpgateway.wrapper` - stdio wrapper
- [x] Federation or Transports
- [ ] CLI, Makefiles, or shell scripts
- [ ] Container setup (Docker/Podman/Compose)
- [ ] Other (explain below)

**Specific Component**: `mcpgateway/services/tool_service.py` - `_extract_and_validate_structured_content` method

---

### 🔁 Steps to Reproduce

1. Create an MCP server with a tool that has an `outputSchema` defined
2. Implement the tool to raise a `ToolError` exception for certain inputs:
   ```python
   from fastmcp import FastMCP
   from fastmcp.exceptions import ToolError

   @mcp.tool()
   def example_tool(value: int) -> dict:
       if value > 100:
           raise ToolError(f"Value exceeds maximum allowed (Requested: {value})")
       return {"success": True, "value": value}
   ```
3. Register the MCP server with ContextForge Gateway
4. Invoke the tool with a value that triggers the error (e.g., `value=150`)
5. Observe the response

---

### 🤔 Expected Behavior

Per [MCP Error Handling Specification](https://modelcontextprotocol.io/specification/2025-11-25/server/tools#error-handling):

- Error responses with `isError: true` should:
  - Contain a `content` array with text error messages
  - **NOT require** `structuredContent` or validation against `outputSchema`
- The gateway should:
  - Skip output schema validation for error responses
  - Preserve the original error message from the tool
  - Return the error response to the client unchanged

**Expected Response**:
```json
{
  "isError": true,
  "content": [
    {
      "type": "text",
      "text": "Value exceeds maximum allowed (Requested: 150)"
    }
  ]
}
```

---

### 📊 Actual Behavior

The gateway validates error responses against `outputSchema`, causing:

1. **Validation failure** because error responses don't contain structured content matching the schema
2. **Error message replacement** - original tool error is overwritten with:
   ```
   Output validation error: <validation details>
   ```
[...]

PR #4204, opened 2026-04-14
Fixed bug in validation of tool error responses

ja8zyjits:
# 🐛 Bug-fix PR

## 📌 Summary

This PR fixes a regression where ContextForge Gateway incorrectly enforces `outputSchema` validation on MCP tool error responses (`isError: true`), breaking MCP specification compliance and preventing standard error handling.

**What was broken:** The gateway validated structured output against `outputSchema` even when tools returned error responses with `isError: true`, causing validation failures and obscuring the actual error messages from tools.

**Why it matters:** This regression broke previously functional MCP integrations and forced developers to implement non-standard workarounds that violate the MCP specification.

## 🔁 Reproduction Steps

**Minimal reproduction:**

1. Create an MCP tool that raises `ToolError` (or use test endpoint: https://mcptest.1yhjfkkdsl9j.us-south.codeengine.appdomain.cloud/mcp)

```python
from fastmcp import FastMCP
from fastmcp.exceptions import ToolError
from pydantic import BaseModel, Field
from uuid import UUID, uuid4

mcp = FastMCP("Test MCP")
    
class RecognitionResult(BaseModel):
    recognitionId: UUID
    message: str

@mcp.tool(description="Submits a recognition for a recipient.")
def submit_recognition(points: int = Field(description="The number of points to award")) -> RecognitionResult:
    if points > 200:
        raise ToolError(f"You cannot send more than 200 points (Requested: {points})")
    
    return RecognitionResult(
        recognitionId=uuid4(),
        message="Recognition submitted successfully with " + str(points) + " points"
    )

if __name__ == "__main__":
    mcp.run(transport="streamable-http", host="0.0.0.0", port=8000)
```

2. Invoke the tool via ContextForge Gateway with `points > 200`
3. **Before fix:** Observe "Output validation error: outputSchema defined but no structured output returned"
4. **After fix:** Error message properly surfaces: "You cannot send more than 200 points (Requested: 2023)"

## 🐞 Root Cause

**Location:** `mcpgateway/services/tool_service.py`, method `_extract_and_validate_structured_content` (lines 1228-1375)

**Problem:** The validation method did not check if `tool_result.is_error` or `tool_result.isError` was `True` before enforcing output schema validation.

**MCP Specification:** According to the [MCP Error Handling Specification](https://modelcontextprotocol.io/specification/2025-11-25/server/tools#error-handling), error responses with `isError: true` should contain:
- `content` array with text messages
- **NO requirement for `structuredContent`**

[...]

PR #4209, opened 2026-04-14
fix: outputSchema declared but no structured output now fails validation (#4208)

xyaz1313:
### Problem

When a tool declares `output_schema` but returns no structured content (empty `content`, no parseable JSON, or `structured_content=None`), the gateway silently returns `True` (valid). This violates the MCP 2025-11-25 spec:

> "If an output schema is provided: Servers **MUST** provide structured results that conform to this schema."

### Root Cause

`ToolService._extract_and_validate_structured_content` at lines 1314-1316:
```python
# If no structured data found, treat as valid (nothing to validate)
if structured is None:
    return True
```

### Fix

Replaced the lenient early return with strict validation:

1. **Error responses are still exempted** — checks `is_error`/`isError` first (consistent with #4202)
2. **Success responses with no structured content now fail** with `is_error=True` and error details:
```json
{
  "code": "missing_structured_output",
  "expected": "object",
  "received": "null",
  "message": "outputSchema declared but no structuredContent was returned"
}
```

### Changes

| File | Change |
|------|--------|
| `mcpgateway/services/tool_service.py` | Lines 1314-1316: replaced lenient pass with strict validation + error response exemption |
| `tests/.../test_tool_service_coverage.py` | Updated 2 tests, added 5 new tests |

### Tests

- test_no_structured_data_with_schema_fails_validation — main behavior change
- test_no_structured_data_error_response_skips_validation — is_error=True exempted
- test_no_structured_data_error_response_alias_skips_validation — isError=True exempted
- test_no_structured_data_non_text_content_fails — non-text content fails
- test_no_structured_data_null_text_fails — text="null" parses to None, fails
- test_content_with_null_text_parses_to_none — updated from assert ok is True to assert ok is False
- Existing validation tests (valid candidate, invalid candidate, TextContent parsing) still pass

### Related

- Complements PR #4202 (error-path fix) — this is the success-path fix
- Aligned with MCP Python SDK server-side validator behavior

### DIFF
--- a/mcpgateway/services/tool_service.py
+++ b/mcpgateway/services/tool_service.py
@@ -1311,9 +1311,27 @@ class ToolService(BaseService):
                         # ignore JSON parse errors and continue
                         continue
 
-            # If no structured data found, treat as valid (nothing to validate)
+            # If no structured data found, treat as validation failure per MCP spec
+            # "If an output schema is provided: Servers MUST provide structured results that conform to this schema."
             if structured is None:
-                return True
+                # Skip for error responses — errors don't need structured content
+                is_error = getattr(tool_result, "is_error", False) or getattr(tool_result, "isError", False)
+                if is_error:
+                    return True
+                schema_type_str = output_schema.get("type", "object") if isinstance(output_schema, dict) else "object"
+                details = {
+                    "code": "missing_structured_output",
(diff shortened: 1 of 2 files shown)


jonpspri:
Closed in favor of #4204

PR #4214, opened 2026-04-15
fix(oauth): honor advisory contract and accept array-shaped scope/scp

### LINKED ISSUE #4213: [BUG]: Gateway egress OAuth claim validation from #3941 violates its own "advisory" contract and crashes on non-conforming IdPs
### 🐞 Bug Summary
After PR #3941 (`ffc3ba816`, "Add OAuth token claim validation before MCP server forwarding") and its follow-up `143aaf63a` ("enhance claim validation with blocking errors"), the OAuth token claim validator introduced in `mcpgateway/services/token_validation_service.py` and wired into `mcpgateway/services/gateway_service.py::fetch_tools_after_oauth` now **hard-blocks token forwarding** for any Authorization Code gateway whose IdP does not match a narrow set of claim shapes. Two concrete failure modes:

1. **Crash**: `_validate_scopes` unconditionally calls `.split()` on the `scope`/`scp` claim, which raises `AttributeError: 'list' object has no attribute 'split'` when the IdP emits the claim as a JSON array (Ory Hydra, Keycloak, and some Auth0 configurations do this). The uncaught exception propagates out of `validate_oauth_token_claims` through `fetch_tools_after_oauth`, returning a 400 to the admin UI.
2. **Hard-block**: Even when parsing succeeds, any `aud`/`iss`/`scope` mismatch populates `TokenValidationResult.blocking_errors`, which `gateway_service.py` then converts into `raise GatewayConnectionError("Refusing to forward OAuth token …")`. This contradicts the module's own docstring, which explicitly declares the validator to be "best-effort advisory — the upstream MCP server remains the authoritative validator" (`token_validation_service.py` lines ~9–15).

Before PR #3941, tokens were forwarded opaquely without claim parsing, so this regression surfaces only for deployments that upgraded past `ffc3ba816`.

---

### 🧩 Affected Component

- [x] `mcpgateway` - API
- [ ] Other (explain below)

---

### 🔁 Steps to Reproduce

1. Register a Gateway (transport `STREAMABLEHTTP`, `grant_type=authorization_code`) against a public MCP server whose IdP does not perfectly follow RFC 9068 / RFC 8707. A reproducible public example is **Amplitude's MCP server** (`https://mcp.amplitude.com/mcp`), which is backed by an Ory Hydra deployment at `https://auth.amplitude.com`.
2. Complete the OAuth authorization code flow successfully. Observe that a valid access token is stored in `oauth_tokens`.
3. Click "Fetch Tools" in the admin UI (or invoke `POST /oauth/fetch-tools/{gateway_id}`). This triggers `GatewayService.fetch_tools_after_oauth`.
4. Observe one of the following depending on the IdP's claim shape:
   - **(a)** Amplitude-like IdPs: fetch fails with
     ```
     Failed to fetch tools after OAuth: 'list' object has no attribute 'split'
     ```
   - **(b)** Authentik / Entra-like IdPs where `aud` is the OAuth `client_id` rather than the resource URL: fetch fails with
     ```
     Refusing to forward OAuth token for gateway '<name>': Token audience mismatch: token aud=[<client_id>], expected '<gateway_url>'. Fix oauth_config (resource/scopes/issuer) or the IdP token request.
     ```

Failure (a) occurs even though the upstream MCP server would accept the token (verified by bypassing CF and hitting the server directly with the same token).

---

### 🤔 Expected Behavior
1. **`_validate_scopes` must accept both string and array forms** of `scope`/`scp`. RFC 6749 defines `scope` as a space-delimited string at the *authorization response* level, but does not prescribe a shape for the JWT access-token claim — IdP implementations diverge. The validator should normalize list values to a joined string before tokenization.
2. **Individual validators must not abort the forwarding path.** Each of `_validate_audience` / `_validate_scopes` / `_validate_issuer` should be wrapped so that an exception is logged as a warning and a non-blocking entry is appended to `TokenValidationResult.warnings`. A crashing validator must not skip sibling validators and must never escape `validate_oauth_token_claims`.
3. **`gateway_service.py::fetch_tools_after_oauth` must respect the module's documented advisory contract.** The `raise GatewayConnectionError(...)` on `blocking_errors` should be a `logger.warning(...)` instead. The rationale:
   - CF does not verify JWT signatures (`options={"verify_signature": False}`) because it does not hold the IdP's keys. Blocking a request on the basis of an unverified claim is unsound.
   - The backend MCP server is the authoritative validator by definition (it owns the trust relationship with its IdP). A token that CF rejects will, in the common case of misconfiguration, be correctly rejected by the upstream; but a token that CF rejects due to CF's strict assumptions will be wrongly blocked even when the upstream would accept it.
   - The module's top-level docstring already states this contract. The implementation added by `143aaf63a` contradicts it.
   - (Optional) Keep the blocking behavior behind an opt-in flag such as `oauth_config.strict_claim_validation = true` for operators who want it, defaulting to false.
---

### 📓 Logs / Error Output
**Failure (a)** — from `gateway_service` log during a tool fetch against an Ory Hydra-backed MCP server:

```
ERROR [mcpgateway.services.gateway_service] Failed to fetch tools after OAuth for gateway <gateway_id>: 'list' object has no attribute 'split'
ERROR [mcpgateway.routers.oauth_router] Failed to fetch tools after OAuth for gateway <gateway_id>: Failed to fetch tools after OAuth: 'list' object has no attribute 'split'
WARNI [mcpgateway.services.structured_logger] [http_gateway] Request completed: POST /oauth/fetch-tools/<gateway_id> - 400
```

**Claims decoded from a real Amplitude access token** (issued directly via `/register` + `/authorize` + `/token` with PKCE, so CF is not in the path; identifiers redacted):

```json
{
  "iss":  "https://auth.amplitude.com",
  "aud":  [],
  "scp":  ["mcp:read", "mcp:write", "offline_access"],
  "scope": null,
  "client_id": "<redacted>",
  "sub": "<redacted>",
  "exp": 1776324352,
  "iat": 1776237951,
  "nbf": 1776237951,
  "jti": "<uuid>"
}
```

Notes on the token above:

- `aud` is empty even though the client sent `resource=https://mcp.amplitude.com` on both `/authorize` and `/token` per RFC 8707. The IdP does not echo the resource indicator into the `aud` claim, making strict RFC 9068 audience validation impossible for this IdP without out-of-band configuration.
- `scp` is a JSON array; there is no `scope` claim. This is the shape that crashes `_normalize_scope`.
- The `/.well-known/oauth-authorization-server` document at `https://mcp.amplitude.com/.well-known/oauth-authorization-server` advertises `"issuer": "https://mcp.amplitude.com"`, but the actual `iss` in issued JWTs is `https://auth.amplitude.com` — a separate upstream conformance issue, but one that also makes strict `iss` validation fail in CF.

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Runtime | `Python 3.12` |
| Platform / OS | `Linux/EKS` |
| Container | `Docker` |

---

- **Module/implementation contract mismatch.** The module-level docstring explicitly states:

  > *"This module provides best-effort JWT claim validation (audience, scopes, issuer) for OAuth access tokens before they are forwarded to upstream MCP servers. Validation is advisory — the upstream MCP server remains the authoritative validator."*

  The current behavior (hard-raise on `blocking_errors`) is inconsistent with this stated contract.

- **Why the upstream MCP server is the right place to enforce.** CF does not hold the IdP's signing keys and therefore cannot verify the JWT signature. It decodes the token with `verify_signature=False`, meaning every claim inspection is unverified. The upstream MCP server, in contrast, is the resource server for the IdP and by design performs authoritative signature + claim validation (it is what the `resource` parameter points at in RFC 8707 terms). It is both unnecessary and unsafe to gate forwarding on CF's unverified claim parse.

- **Real-world IdP diversity.**
  - **Amplitude MCP** (`https://mcp.amplitude.com`) — Ory Hydra-style: `scp` is an array, `aud` is empty even with RFC 8707 sent, `iss` ≠ AS metadata `issuer`. Observed via both CF and direct OAuth flow.
  - **Authentik** — `aud` = client_id, not resource URL (Authentik has no RFC 8707 support). The related `include_client_id_in_audience` toggle already exists for virtual-server ingress validation (`streamablehttp_transport.py`) but was never extended to `token_validation_service._validate_audience`, so it does not help for gateway egress.
  - **Keycloak** — known to emit array-form `scope` claims in a number of configurations, which would hit the same crash path.

kimsehwan96:
## 🔗 Related Issue
Closes #4213

---

## 📝 Summary

Two small regressions in the OAuth token claim validator introduced in #3941 break tool fetching for non-conforming IdPs:

1. `_validate_scopes` calls `.split()` on the `scope` / `scp` claim, which crashes when the IdP emits an array (Ory Hydra, Keycloak, some Auth0 configurations).
2. `gateway_service.py::fetch_tools_after_oauth` raises `GatewayConnectionError` on any mismatched claim, contradicting the module's own "advisory — the upstream MCP server remains the authoritative validator" contract stated in `token_validation_service.py`.

This PR:
- Normalizes array-shaped `scope`/`scp` to a space-separated string before tokenization.
- Wraps each `_validate_*` call in its own `try / except` so one crashing validator cannot abort siblings or propagate to the caller; crashes become non-blocking warnings.
- Downgrades the hard-block in `fetch_tools_after_oauth` to a `logger.warning` that preserves the existing operator guidance (`Fix oauth_config(resource/scopes/issuer) …`). The upstream MCP server remains the authoritative validator, per the module docstring.

---

## 🏷️ Type of Change
- [x] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

---

## ✅ Checklist

---

## 📓 Notes (optional)
- Reproduced end-to-end against **Amplitude's public MCP server** (`https://mcp.amplitude.com/mcp`, Ory Hydra-backed, issuer `https://auth.amplitude.com`) which emits `scp` as an array and `aud` as an empty list even when the client sends RFC 8707 `resource=` on both `/authorize` and `/token`. Full decoded JWT claims (from a direct DCR + PKCE flow, bypassing ContextForge) are included in #4213.
- The previous "Refusing to forward OAuth token … Fix oauth_config …" diagnostic is preserved as a warning log, so operators still get the remediation hint when a mismatch is genuinely their misconfiguration.
- If reviewers prefer to keep the strict blocking behavior available to deployments that want it, happy to follow up with an opt-in `oauth_config.strict_claim_validation` flag — did not include it here to keep the change focused on the regression.

### DIFF
--- a/mcpgateway/services/gateway_service.py
+++ b/mcpgateway/services/gateway_service.py
@@ -1592,13 +1592,24 @@ class GatewayService(BaseService):  # pylint: disable=too-many-instance-attribut
             for warning in token_validation.warnings:
                 logger.warning("OAuth token validation for gateway %s: %s", gateway.name, warning)
 
-            # Fail fast if any claim is definitively mismatched (present but wrong).
-            # Claims that are simply absent from the token produce None (not False)
-            # and are NOT blocked — this preserves backward compat with legacy IdPs.
+            # Advisory: log mismatched claims but do not block forwarding.
+            # The upstream MCP server is the authoritative token validator
+            # (see mcpgateway.services.token_validation_service docstring).
+            # The operator-facing guidance from the previous hard-block path
+            # is preserved below as a warning so misconfiguration is still
+            # surfaced without breaking tool fetching for non-conforming
+            # IdPs (e.g. those that omit RFC 8707 resource echo or emit
+            # unusual claim shapes).
             blocking = token_validation.blocking_errors
             if blocking:
                 detail = "; ".join(blocking)
-                raise GatewayConnectionError(f"Refusing to forward OAuth token for gateway '{gateway.name}': " f"{detail}. " f"Fix oauth_config (resource/scopes/issuer) or the IdP token request.")
+                logger.warning(
+                    "OAuth token claim validation flagged issues for gateway '%s' "
+                    "(advisory, forwarding anyway): %s. "
+                    "Fix oauth_config (resource/scopes/issuer) or the IdP token request if this is unintended.",
+                    gateway.name,
+                    detail,
+                )
 
             # Now connect to MCP server with the access token
             authentication = {"Authorization": f"Bearer {access_token}"}
--- a/mcpgateway/services/token_validation_service.py
+++ b/mcpgateway/services/token_validation_service.py
@@ -185,8 +185,16 @@ def _validate_scopes(claims: Dict[str, Any], oauth_config: Dict[str, Any], gatew
     if not configured_scopes:
         return
 
-    # Entra ID uses 'scp' claim; standard OAuth uses 'scope'
-    token_scope_str = claims.get("scope") or claims.get("scp") or ""
+    # Entra ID uses 'scp' claim; standard OAuth uses 'scope'.
+    # Some IdPs (Ory Hydra, Keycloak, certain Auth0 configurations) emit
+    # these claims as JSON arrays rather than space-separated strings —
+    # normalize the array shape to a single space-separated string before
+    # tokenization so ``_normalize_scope`` continues to receive a ``str``.
+    token_scope = claims.get("scope") or claims.get("scp") or ""
+    if isinstance(token_scope, list):
+        token_scope_str = " ".join(str(s) for s in token_scope if s)
+    else:
+        token_scope_str = str(token_scope) if token_scope else ""
     if not token_scope_str:
         logger.debug("OAuth token for gateway %s has no 'scope'/'scp' claim", gateway_name)
         return
@@ -287,8 +295,27 @@ def validate_oauth_token_claims(
 
     result.is_jwt = True
 
-    _validate_audience(claims, oauth_config, gateway_url, gateway_name, result)
-    _validate_scopes(claims, oauth_config, gateway_name, result)
-    _validate_issuer(claims, oauth_config, gateway_name, result)
+    # Advisory: each claim validator is best-effort.  Unexpected claim shapes
+    # from non-conforming IdPs must not abort the forwarding path — the
+    # upstream MCP server remains the authoritative validator (see module
+    # docstring).  Wrap each validator so one crashing check does not skip
+    # the others and does not propagate to the caller.
+    try:
+        _validate_audience(claims, oauth_config, gateway_url, gateway_name, result)
+    except Exception as exc:  # pylint: disable=broad-except
+        logger.warning("OAuth audience validation crashed for gateway %s: %s", gateway_name, exc)
+        result.warnings.append(f"audience validation error (non-blocking): {type(exc).__name__}: {exc}")
+
+    try:
+        _validate_scopes(claims, oauth_config, gateway_name, result)
+    except Exception as exc:  # pylint: disable=broad-except
+        logger.warning("OAuth scope validation crashed for gateway %s: %s", gateway_name, exc)
+        result.warnings.append(f"scope validation error (non-blocking): {type(exc).__name__}: {exc}")
+
+    try:
+        _validate_issuer(claims, oauth_config, gateway_name, result)
+    except Exception as exc:  # pylint: disable=broad-except
+        logger.warning("OAuth issuer validation crashed for gateway %s: %s", gateway_name, exc)
+        result.warnings.append(f"issuer validation error (non-blocking): {type(exc).__name__}: {exc}")
 
     return result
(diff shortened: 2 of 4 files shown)


kimsehwan96:
@brian-hussey @crivetimihai Hi team, Can I get some review about this? 

Without this, our gateway breaks when using the [Amplitude](https://amplitude.com/) MCP server — the upstream issues an access token whose `scope`/`scp` claim is array-shaped, and the current strict validation rejects it before the request is forwarded

스크린샷 2026-05-04 오전 8 19 53

I think strict JWT claim validation is the upstream MCP server's responsibility, they issue the token and own its canonical shape.

More importantly, this contradicts the module's own documented contract. The validator's header comment states

> "This module provides best-effort JWT claim validation (audience, scopes, issuer) for OAuth access tokens before they are forwarded to upstream MCP servers. Validation is advisory — the upstream MCP server remains the authoritative validator."

https://github.com/IBM/mcp-context-forge/blob/f855e54d5b7bc1c91b977574a03b91eff6b86bb6/mcpgateway/services/token_validation_service.py#L9-L11

PR #4226, opened 2026-04-15
feat: add local ci parity command

lucarlig:
# ✨ Feature / Enhancement PR

## 🔗 Epic / Issue
Closes #

---

## 🚀 Summary (1-2 sentences)
_Add `make ci` as a workflow-driven local CI parity command that selects the union of GitHub jobs for PRs to `main` and pushes to `main`. Document that the command is best-effort parity, not a guarantee, and mark Node-backed jobs as approximations when local bootstrap differs from GitHub._

---

## 🧪 Checks

- [x] `make lint` passes
- [x] `make test` passes

---

## 📓 Notes (optional)
Rationale: the goal is to make local validation a strong predictor of GitHub CI without hiding the places where local execution cannot truly match GitHub-hosted runners. The implementation reads the workflow files as the source of truth for selection, then labels each mapped job as `exact`, `approx`, or `not_reproducible` so reviewers can see where parity is real versus best-effort.

Local parity intentionally blocks publish, signing, artifact upload, and similar GitHub-hosted side effects. Node-backed jobs are marked as approximations because the local runner does not reproduce the workflow bootstrap that pins the Node version, upgrades npm, and configures the registry before executing those checks.


lucarlig:
won't do

PR #4235, opened 2026-04-16
fix(ui): Normalize team_id whitespace checks across all edit functions and server-side visibility guard

### LINKED ISSUE #3338: [CHORE][UI]: Normalize team_id consistently across JS coercion checks and isTeamScopedView
## Problem

`isTeamScopedView()` in `admin.js:207` trims `team_id` before checking emptiness:

```js
function isTeamScopedView() {
    const teamId = new URLSearchParams(window.location.search).get("team_id");
    return Boolean(teamId && teamId.trim() !== "");
}
```

But the visibility coercion logic in all 6 edit functions (`editTool`, `editA2AAgent`, `editResource`, `editPrompt`, `editGateway`, `editServer`) checks raw `teamId` truthiness without trimming:

```js
const effectiveVisibility =
    window.ALLOW_PUBLIC_VISIBILITY === false &&
    visibility === "public" &&
    teamId          // ← raw truthiness, no trim
        ? "team"
        : visibility;
```

The server-side check in `_check_public_visibility_allowed` (`admin.py:1306`) also uses raw truthiness:

```python
if not settings.allow_public_visibility and visibility == "public" and team_id:
```

## Impact

With a malformed URL like `?team_id=%20` (whitespace-only):
- `isTeamScopedView()` returns `false` (trimmed empty string)
- Edit function coercion treats it as team-scoped (`" "` is truthy)
- `applyVisibilityRestrictions` does NOT disable the public radio (relies on `isTeamScopedView`)
- Result: coercion silently changes public→team, but the public radio appears enabled

Not a security issue (server still blocks public visibility submission), but produces confusing UX.

## Suggested Fix

Make the edit function coercion checks use `isTeamScopedView()` instead of raw `teamId` truthiness. For functions that use a local `_teamId` variable solely for the coercion check (`editResource`, `editPrompt`, `editServer`), replace it with `isTeamScopedView()`. For the others, replace just the coercion condition.

Optionally, also normalize on the server side in `_check_public_visibility_allowed` with `team_id and team_id.strip()`.

## References

- `isTeamScopedView()`: `mcpgateway/static/admin.js:207`
- Coercion examples: `admin.js:3509` (editTool), `admin.js:5108` (editResource), `admin.js:6872` (editServer)
- Server check: `mcpgateway/admin.py:1293`
- Found during review of #3323

shoummu1:
## :pushpin: Summary

This PR contains two fixes:

### 1. Whitespace `team_id` visibility coercion

All 6 edit functions in the Admin UI (`editTool`, `editGateway`, `editA2AAgent`, `editResource`, `editPrompt`, `editServer`) were checking raw `teamId` truthiness to decide whether to coerce a `public` visibility record to `team`. A whitespace-only URL parameter (`?team_id=%20`) is truthy in JS, causing silent coercion while `isTeamScopedView()` — already trim-aware — correctly returned `false`. The Python server-side guard had the same gap. This fix makes all checks consistent.

### 2. Dead code removal in `servers.js`

`serverDesc.textContent` was assigned via `decodeHtml()` and then immediately overwritten with the raw value on the next line, making the `decodeHtml` call unreachable and causing HTML entities in server descriptions to render incorrectly.

## :link: Related Issue
Closes: #3338

1. Set `ALLOW_PUBLIC_VISIBILITY=false`.
2. Navigate to the admin UI with `?team_id=%20` in the URL.
3. Open the edit modal for any tool, gateway, A2A agent, resource, prompt, or server that has `public` visibility.
4. Observe: the public radio button appears enabled (because `isTeamScopedView()` returns `false`), but the visibility is silently coerced to `team` on submit.

## 🐞  Root Cause

The `effectiveVisibility` ternary in each of the 6 edit functions used the raw URL-derived `teamId`/`_teamId` variable directly:

```js
// Before — raw truthiness, no trim
teamId ? "team" : visibility
```

`" "` (whitespace) is truthy, so it triggered coercion. Meanwhile `isTeamScopedView()` in `teams.js` correctly trims before checking, returning `false` for whitespace-only values. The two code paths were inconsistent.

On the Python side, `_check_public_visibility_allowed` in `admin.py` had the same raw-truthiness check (`and team_id`).

### Fix 1 — Whitespace `team_id` coercion

- **JS (6 files):** Added `isTeamScopedView` to the import from `./teams.js` in each affected module and replaced the raw `teamId` / `_teamId` condition in the `effectiveVisibility` ternary with `isTeamScopedView()`. For `editResource`, `editPrompt`, and `editServer`, the now-redundant local `_teamId` variable was also removed.
- **Python (`admin.py`):** Changed `and team_id` → `and team_id and team_id.strip()` in `_check_public_visibility_allowed` so both layers treat whitespace-only values as absent.

### Fix 2 — Dead code removal (`servers.js`)

- Removed the redundant `serverDesc.textContent = server.description` line that immediately overwrote the `decodeHtml()` result, so HTML entities in server descriptions are now decoded correctly.
- **Tests:** Fixed the existing visibility-coercion test in `prompts.test.js` to include `search` in the mocked `window.location` (required since `isTeamScopedView` reads `location.search`, not `href`). Added a whitespace-only `team_id` test asserting no coercion occurs. Updated the `teams.js` mock in `resources.test.js` to export `isTeamScopedView` (previously missing, would throw at runtime in tests). Added 3 direct tests of `_check_public_visibility_allowed` in `test_admin.py` covering spaces-only, tab, and a real `team_id` to verify the `team_id.strip()` branch.

## :test_tube: Verification

| Check | Command | Status |
|---|---|---|
| Coverage ≥ 90 % | `make coverage` | |
| Manual regression no longer fails | Visit `?team_id=%20`, open any edit modal with a public entity — coercion no longer occurs | |

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-05-06T02:43:07Z",
+  "generated_at": "2026-05-06T13:31:27Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -4178,7 +4178,7 @@
         "hashed_secret": "fa9beb99e4029ad5a6615399e7bbae21356086b3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 4190,
+        "line_number": 4205,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -4186,7 +4186,7 @@
         "hashed_secret": "559b05f1b2863e725b76e216ac3dadecbf92e244",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 4825,
+        "line_number": 4840,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -4194,7 +4194,7 @@
         "hashed_secret": "a8af4759392d4f7496d613174f33afe2074a4b8d",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 4827,
+        "line_number": 4842,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -4202,7 +4202,7 @@
         "hashed_secret": "85b60d811d16ff56b3654587d4487f713bfa33b7",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 15206,
+        "line_number": 15211,
         "type": "Secret Keyword",
         "verified_result": null
       }
--- a/mcpgateway/admin.py
+++ b/mcpgateway/admin.py
@@ -1694,7 +1694,7 @@ def _check_public_visibility_allowed(visibility: str, team_id: Optional[str] = N
     Raises:
         HTTPException: 422 when flag is false, team_id is set, and visibility is 'public'.
     """
-    if not settings.allow_public_visibility and visibility == "public" and team_id:
+    if not settings.allow_public_visibility and visibility == "public" and team_id and team_id.strip():
         raise HTTPException(
             status_code=422,
             detail="Public visibility is disabled by platform configuration (ALLOW_PUBLIC_VISIBILITY=false).",
--- a/mcpgateway/admin.py
+++ b/mcpgateway/admin.py
@@ -806,6 +806,25 @@ def _validated_team_id_param(team_id: Optional[str] = Query(None, description="F
         raise HTTPException(status_code=400, detail="Invalid team ID") from exc
 
 
+def _form_team_id(form: Any) -> Optional[str]:
+    """Extract and normalize team_id from form data, converting whitespace-only values to None.
+
+    Normalizes at the point of extraction so that both the public-visibility guard
+    and ``TeamManagementService.verify_team_for_user`` receive the same value, making
+    ``?team_id=%20`` behave identically to an absent ``team_id`` end-to-end.
+
+    Args:
+        form: The multipart form data object from the request.
+
+    Returns:
+        The stripped team_id string, or None if absent or whitespace-only.
+    """
+    raw = form.get("team_id")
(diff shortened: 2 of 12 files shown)


vishu-bh (changes requested):
Thanks for tackling this and for making the UI-side checks use the shared `isTeamScopedView()` helper. That’s a good direction and makes the edit-modal behavior much more consistent.

I think there’s one blocking runtime path still worth tightening before merge: whitespace-only `team_id` is now ignored by `_check_public_visibility_allowed()`, but the raw value can still be submitted and passed into `TeamManagementService.verify_team_for_user()`. Since `" "` is truthy there, it is treated as an explicit team id and can return `[]` / fail validation instead of being handled as “no team”.

A practical approach would be to normalize `team_id` once before both the visibility guard and team verification, for example converting whitespace-only values to `None`, and ideally use the same trimmed logic in the edit form hidden-field / submit handling paths. That would make `?team_id=%20` consistently behave as absent end to end, not just during the public visibility guard.

shoummu1:
@vishu-bh Thanks for the review! You're right that the original fix was layer-1-only.

Addressed in the latest commit: added a `_form_team_id(form)` helper that strips and coerces whitespace-only values to `None` at the point of form extraction, before both `_check_public_visibility_allowed()` and `verify_team_for_user()`. All 12 affected handlers (add/edit for server, tool, gateway, resource, prompt, a2a_agent) now call this single helper, so `?team_id=%20` consistently behaves as absent end-to-end rather than only at the visibility gate. Also added `TestFormTeamId` unit tests covering the absent, empty, whitespace-only, and real team_id cases.

vishu-bh (dismissed):
Thanks @shoummu1  for addressing the review comment.
Six admin edit flows now use isTeamScopedView() for visibility coercion, and server-side form handling normalizes whitespace-only team_id values to None before both the public visibility guard and team verification.

LGTM 🚀

brian-hussey (dismissed):
Approved based on previous reviews. Rebasing and merging.
@shoummu1 Thank you for your contribution.

brian-hussey (approved):

PR #4242, opened 2026-04-16
fix(ci): use check-runs API and peel annotated tags in docker-release…

### LINKED ISSUE #4200: [CHORE][CICD]: docker-release.yml fails on every release due to stale status API check
## Summary

The `docker-release.yml` workflow has failed on every release tag for the last three releases (1.0.0-RC1, 1.0.0-RC2, 1.0.0-RC3) at the same step: **`Verify commit checks passed`**. The failure is not release-blocking because the primary build workflow already applies the version tag to the GHCR image when the tag is pushed — but the workflow still runs on every release, reports a red X, and shouldn't.

## Failing runs

- RC-3: https://github.com/IBM/mcp-context-forge/actions/runs/24404094747
- RC-2: (2026-03-09) — same failure
- RC-1: (2026-02-18) — same failure

## Root cause

The step runs:

```bash
STATUS=$(curl -sSL \
  -H "Accept: application/vnd.github+json" \
  -H "Authorization: ***" \
  "https://api.github.com/repos/$REPO/commits/$SHA/status" \
  | jq -r '.state')
echo "Combined status: $STATUS"
if [ "$STATUS" != "success" ]; then
  echo "Required workflows have not all succeeded - aborting." >&2
  exit 1
fi
```

Problem: `GET /repos/{owner}/{repo}/commits/{sha}/status` is the **legacy "commit statuses" API** (used by old integrations like Travis CI, Jenkins, Codacy). Modern GitHub Actions publish **check-runs**, not statuses. This repo has no active commit-status providers, so the combined-status endpoint returns `{ "state": "pending", "statuses": [] }` by default — there's nothing to aggregate. The step then unconditionally aborts.

Log snippet from the RC-3 run:

```
Required workflows have not all succeeded - aborting.
Combined status: pending
##[error]Process completed with exit code 1.
```

## Likely second bug

The step's env had:

```
SHA: f606247c6493a1257151a868ba15d21a8dba67d1
```

That SHA is **not** the merge commit on `main` (`a02a04bce0a644cb26587061bb4ef084cb15c65f`). It is very likely the SHA of the annotated tag object itself — annotated tags in Git have their own SHA, distinct from the commit they point to. The workflow should dereference the tag (`$SHA^{commit}` or use `github.event.release.target_commitish` / appropriate target-commit field) before calling the status/check-runs API.

## Suggested fix

Replace the legacy status aggregation with a check-runs query. Minimal shape:

```bash
# Dereference annotated tag to its commit
COMMIT=$(git rev-parse "$SHA^{commit}")

# Pull all check-runs for the commit; fail if any required one is not "success"
RESPONSE=$(gh api \
  -H "Accept: application/vnd.github+json" \
  "repos/$REPO/commits/$COMMIT/check-runs?per_page=100")

FAILED=$(echo "$RESPONSE" | jq -r '[.check_runs[] | select(.conclusion != "success" and .conclusion != "skipped" and .conclusion != "neutral")] | length')

if [ "$FAILED" -gt 0 ]; then
  echo "❌ One or more required check-runs have not succeeded - aborting."
  echo "$RESPONSE" | jq -r '.check_runs[] | select(.conclusion != "success" and .conclusion != "skipped" and .conclusion != "neutral") | "  - \(.name): \(.conclusion // .status)"'
  exit 1
fi
```

Alternative: drop the pre-flight check entirely, since the workflow is only supposed to re-tag an image that the main build workflow already produced. The main build workflow itself runs the CI gates; if *it* failed, there would be no image to re-tag.

## Impact

**Not release-blocking.** The primary build workflow applies the version tag directly when the tag is pushed (confirmed on GHCR for RC-3 with a proper multiarch manifest list: `amd64 / arm64 / s390x / ppc64le` at digest `sha256:0a4feab65ae1619f98b6fbea65e51cf1a74069dc66ee9ed23974fa9eb181d142`). So `docker-release.yml` is doing a redundant retag that the main workflow already covered. But a red X on every release is noisy and erodes trust in CI signals.

## Acceptance criteria

- [ ] `docker-release.yml` runs green on the next release tag
- [ ] Either (a) status check replaced with check-runs aggregation + tag dereference, or (b) pre-flight check removed as redundant, with a note explaining why
- [ ] Verify RC1, RC2, RC3 historical runs don't regress (they can stay red — this is a forward fix)

omorros:
## 🔗 Related Issue
  Closes #4200

  ---

  ## 📝 Summary
  `docker-release.yml` has failed on every release tag (v1.0.0-RC1/RC2/RC-3) at the `Verify commit checks passed` step. Two underlying bugs:

  1. The step queries the legacy `/commits/{sha}/status` endpoint, which only aggregates the old commit-statuses API (Travis/Jenkins webhooks). This repo
  publishes GitHub Actions check-runs instead, so the endpoint returns `{"state":"pending","statuses":[]}` and the step aborts on every release.
  2. `Step 1` resolved `refs/tags/$TAG` via `git ls-remote --refs`, which for annotated tags returns the tag-object SHA (e.g. `f606247c…`), not the commit
  SHA (`a02a04bce…`). Downstream steps would then try to re-tag an image that doesn't exist.

  This PR:
  - Peels annotated tags with `refs/tags/$TAG^{}` (with lightweight-tag and `target_commitish` fallbacks).
  - Replaces the legacy status endpoint with `/commits/{sha}/check-runs?per_page=100`.
  - Only blocks on *completed* check-runs whose conclusion is not `success`, `skipped`, or `neutral`. In-progress runs are ignored so this workflow's own
  `tag-and-push` check-run does not self-reference.
  - Guards against missing `.check_runs` in the response and against commits with zero CI evidence.
  - Narrows the job permission from `statuses: read` to `checks: read` (required by the new API).

  Verified live against the real upstream API: annotated tag `v1.0.0-RC-3` now peels to commit `a02a04bce…`, the legacy status endpoint reproduces
  `state=pending, 0 statuses`, and the check-runs endpoint returns the expected 51 runs for that commit.

  ---

  ## 🏷️  Type of Change
  - [x] Bug fix
  - [ ] Feature / Enhancement
  - [ ] Documentation
  - [ ] Refactor
  - [ ] Other (describe below)

  ---

  ## 🧪 Verification

  | Lint suite                | `make lint`     | ✅ yamllint + actionlint (via `rhysd/actionlint` Docker image) + zizmor (via `ghcr.io/zizmorcore/zizmor`)
   all pass on the modified workflow; zizmor finding count matches `origin/main` baseline (0 new findings). `make` is not available in my local env, so
  tools were invoked directly with the same configs the Makefile uses. |
  | Unit tests                | `make test`     | ➖ N/A — workflow-only change, no Python/Rust code touched. |
  | Coverage ≥ 80%            | `make coverage` | ➖ N/A — workflow-only change. |

  Additional end-to-end verification of the modified bash under `set -euo pipefail` with real `jq` against the live GitHub API:

  - Annotated tag `v1.0.0-RC-3` → Step 1 peels `f606247c…` to commit `a02a04bce…`; Step 2 correctly blocks (commit has genuine failed check-runs).
  - Lightweight tag `v1.0.0-RC1` → Step 1 falls back to `bc42bbb2…`; Step 2 correctly blocks.
  - `origin/main` HEAD (all green) → Step 2 passes (36/36 completed check-runs).
  - Non-existent SHA `0000…0000` → Step 2 aborts cleanly with the API's "No commit found for SHA" message (no crash, no false green).

  ---

  ## ✅ Checklist

  ---

  ## 📓 Notes (optional)
  - Workflow files have no unit tests in this repo, so the "tests added/updated" box is intentionally unchecked; the change is validated by yamllint,
  actionlint, zizmor, and an end-to-end bash simulation against the live GitHub API (described above).
  - Historical RC1/RC2/RC-3 runs remain red per the issue's acceptance criteria — this is a forward fix.
  - The 2 pre-existing HIGH zizmor template-injection findings on lines `${{ github.event.release.tag_name }}` and `${{
  github.event.release.target_commitish }}` are untouched by this PR (they exist identically in `origin/main`) and are out of scope for #4200; worth
  tracking as a separate hardening issue if desired.

  A few honesty flags on the boxes I checked:
  - "Code formatted": checked because yamllint is clean; make black isort pre-commit targets apply to Python and weren't run (N/A for a YAML-only change).
  - "Documentation updated": checked because the inline workflow comments were updated to reflect the new behavior. If the repo uses "documentation" to
  mean only docs/, uncheck it.
  - "Tests added/updated": intentionally unchecked — noted in the notes section.

### DIFF
--- a/.github/workflows/docker-release.yml
+++ b/.github/workflows/docker-release.yml
@@ -5,9 +5,10 @@
 # This workflow re-tags a Docker image (built by a previous workflow)
 # when a GitHub Release is published, giving it a semantic version tag
 # like `v0.9.0`. It assumes the CI build has already pushed an image
-# tagged with the commit SHA, and that all checks on that commit passed.
+# tagged with the commit SHA. It verifies all check-runs passed before
+# re-tagging.
 #
-# ➤ Trigger: Release published (e.g. from GitHub UI or `gh release` CLI)
+# ➤ Trigger: Release published, or manual workflow_dispatch with a tag
 # ➤ Assumes: Existing image tagged with the commit SHA is available
 # ➤ Result: Image re-tagged as `ghcr.io/OWNER/REPO:v0.9.0`
 #
@@ -39,12 +40,14 @@ permissions:
 jobs:
   tag-and-push:
     # ------------------------------------------------------------------
-    # Only run if the release tag starts with 'v', and is not draft/prerelease
+    # Run on workflow_dispatch, or when a non-draft, non-prerelease
+    # release tag starts with 'v'.
     # ------------------------------------------------------------------
     if: |
-      startsWith(github.event.release.tag_name, 'v') &&
-      github.event.release.draft == false &&
-      github.event.release.prerelease == false
+      (github.event_name == 'workflow_dispatch') ||
+      (startsWith(github.event.release.tag_name, 'v') &&
+       github.event.release.draft == false &&
+       github.event.release.prerelease == false)
 
     runs-on: ubuntu-latest
     timeout-minutes: 15
@@ -52,7 +55,7 @@ jobs:
     permissions:
       contents: read      # read repo info
       packages: write     # push Docker image
-      checks: read        # aggregate GitHub Actions check-runs for the release commit
+      checks: read        # read GitHub Actions check-runs for the release commit
 
     steps:
       # ----------------------------------------------------------------
@@ -61,12 +64,13 @@ jobs:
       - name: 🏷️  Extract tag & commit SHA
         id: meta
         shell: bash
+        env:
+          TAG: ${{ github.event.release.tag_name || github.event.inputs.tag }}
+          TARGET_COMMITISH: ${{ github.event.release.target_commitish }}
+          REMOTE: https://github.com/${{ github.repository }}.git
         run: |
           set -euo pipefail
-          TAG="${{ github.event.release.tag_name }}"
           echo "tag=$TAG" >>"$GITHUB_OUTPUT"
-
-          REMOTE="https://github.com/${{ github.repository }}.git"
 
           # Peel annotated tags so we get the commit SHA, not the tag object SHA.
           SHA=$(git ls-remote --quiet "$REMOTE" "refs/tags/$TAG^{}" | cut -f1)
@@ -78,7 +82,14 @@ jobs:
 
           # Final fallback: the release's target_commitish (branch name or SHA).
           if [ -z "$SHA" ] || [ "$SHA" = "null" ]; then
-            SHA="${{ github.event.release.target_commitish }}"
+            SHA="$TARGET_COMMITISH"
+          fi
+
+          # Validate that SHA is a 40-character hex commit hash, not a branch name.
+          if ! echo "$SHA" | grep -qE '^[0-9a-f]{40}$'; then
+            echo "ERROR: Resolved SHA '$SHA' is not a valid commit hash." >&2
+            echo "This may indicate the tag was deleted or target_commitish is a branch name." >&2
+            exit 1
           fi
 
           echo "Resolved commit SHA: $SHA"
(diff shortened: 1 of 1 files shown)


jonpspri:
Thanks.  This came up in our release work but we hadn't had the time to address it yet.  Reviewing and rebasing now.

PR #4260, opened 2026-04-17
fix(ui): A2A agent list now refreshes correctly after deletion

### LINKED ISSUE #4247: [BUG]: Admin UI A2A agent list doesn't refresh automatically after deleting agent
### 🐞 Bug Summary                                                                                                                                                                                                                                
  When deleting an A2A agent from the Admin UI list, the deletion succeeds on the server but the UI list doesn't automatically refresh - the deleted agent remains visible until the page is manually refreshed.
                                                                                                                                                                                                                                                    
  ---
                                                                                                                                                                                                                                                    
  ### 🧩 Affected Component                                                                                                                                                                                                                         
   
  ---

  ### 🔁 Steps to Reproduce                                                                                                                                                                                                                         
   
  1. Open Admin UI and navigate to the A2A agents list page                                                                                                                                                                                         
  2. Click delete on a specific A2A agent
  3. Observe deletion succeeds (server-side)                                                                                                                                                                                                        
  4. Notice the deleted agent is still visible in the list                                                                                                                                                                                          
  5. Manual page refresh is required to see the updated list                                                                                                                                                                                        
                                                                                                                                                                                                                                                    
  ---                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                    
  ### 🤔 Expected Behavior
  After successful deletion, the UI should automatically refresh the agent list to reflect the current state without requiring manual page refresh.

rakdutta:
The A2A agent list was not refreshing properly after deleting an agent. The UI was attempting to refresh using a hardcoded path pattern (${type}/partial) and target selector (#${type}-table), which didn't match the actual A2A agent panel     
  configuration (a2a/partial path and #agents-table target).                                                                                                                                                                                        
   
  Root cause: The refresh logic in formHandlers.js used string interpolation instead of consulting the centralized PANEL_SEARCH_CONFIG, causing mismatches for any entity type whose path/selector didn't follow the assumed pattern.               
                             
  Solution                                                                                                                                                                                                                                          
                             
  Refactored handleToggleSubmit in formHandlers.js to use PANEL_SEARCH_CONFIG as the single source of truth for partial URLs and target selectors. This ensures correct refresh behavior for all panel types.                                       
   
  Changes                                                                                                                                                                                                                                           
                             
  - mcpgateway/admin_ui/constants.js:                                                                                                                                                                                                               
    - Added servers entry to PANEL_SEARCH_CONFIG (was missing, causing potential refresh issues for virtual servers)
    - Added documentation warning that all entity types used with delete/toggle operations must be registered                                                                                                                                       
  - mcpgateway/admin_ui/formHandlers.js:                                                                                                                                                                                                            
    - Renamed handleToggleSubmit → handleFormSubmitAndRefresh (kept legacy alias for compatibility)                                                                                                                                                 
    - Use PANEL_SEARCH_CONFIG to retrieve partialPath and targetSelector instead of hardcoded patterns                                                                                                                                              
    - Added ENTITY_DISPLAY_NAMES map for better confirmation dialog messages (singular names)                                                                                                                                                       
    - Added validation that throws descriptive error if entity type not registered in config                                                                                                                                                        
    - Enhanced refresh to preserve search query, tag filters, and pagination state                                                                                                                                                                  
  - tests/unit/js/formHandlers.test.js:                                                                                                                                                                                                             
    - Added test verifying A2A agents use correct a2a/partial path and #agents-table selector                                                                                                                                                       
    - Added test verifying catalog/servers use correct servers/partial path and #servers-table selector                                                                                                                                             
    - Added test verifying error thrown when PANEL_SEARCH_CONFIG missing for a type                                                                                                                                                                 
    - Updated existing tests to use plural entity type keys (aligns with config)                                                                                                                                                                    
                                                                                                                                                                                                                                                    
  Testing                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                    
  - ✅ Unit tests pass (4 new test cases)                                                                                                                                                                                                           
  - ✅ Manual verification: A2A agent deletion now refreshes list immediately
  - ✅ Manual verification: Server deletion refreshes correctly                                                                                                                                                                                     
                                                                                                                                                                                                                                                    
  Closes #4247

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-05-09T21:23:24Z",
+  "generated_at": "2026-05-09T20:22:32Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -92,7 +92,7 @@
         "hashed_secret": "08cd923367890009657eab812753379bdb321eeb",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 634,
+        "line_number": 622,
         "type": "Basic Auth Credentials",
         "verified_result": null
       },
@@ -100,7 +100,7 @@
         "hashed_secret": "14f8aa3e560a47851908ab0f04ec856dbc512d93",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 829,
+        "line_number": 817,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -108,7 +108,7 @@
         "hashed_secret": "fa9beb99e4029ad5a6615399e7bbae21356086b3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1102,
+        "line_number": 1090,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -116,7 +116,7 @@
         "hashed_secret": "7b4455a56fbf1d198e45e04c437488514645a82c",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1128,
+        "line_number": 1116,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -124,7 +124,7 @@
         "hashed_secret": "d08f88df745fa7950b104e4a707a31cfce7b5841",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1136,
+        "line_number": 1124,
         "type": "Secret Keyword",
         "verified_result": null
       },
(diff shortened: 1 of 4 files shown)


marekdano (dismissed):
### Summary

  The PR correctly fixes the root cause: the old `handleToggleSubmit` used `${type}-table` and `${type}/partial` via string interpolation, which broke for A2A agents since `"a2a-agents"` doesn't map to `#a2a-agents-table` (actual: `#agents-table`) or `a2a-agents/partial` (actual:`a2a/partial`). Routing through `PANEL_SEARCH_CONFIG` is the right approach.

###  What's good

  - Core fix is correct — `PANEL_SEARCH_CONFIG["a2a-agents"]` already had the right `partialPath` and `targetSelector`; using it as the source of truth eliminates the mismatch.
  - Tests are meaningful — the new cases directly verify HTMX is called with the correct path and selector.
  - `buildTableUrl` integration preserves pagination/search/filter state across deletes, a nice improvement over the original.
  - Backward compatibility via export `const handleToggleSubmit = handleFormSubmitAndRefresh` is clean.
  - Validation error thrown for unregistered types sits outside the try-catch, so it surfaces loudly during development.

LGTM 🚀

PR #4263, opened 2026-04-17
fix: clean up server_tool_association before tool deletion

### LINKED ISSUE #4261: [BUG]: Deleting A2A agent fails with FOREIGN KEY constraint on server_tool_association
### 🐞 Bug Summary
Deleting an A2A agent (or its associated tool) from the Admin UI fails with `sqlite3.IntegrityError: FOREIGN KEY constraint failed` because the `server_tool_association` rows referencing the tool are not cleaned up before the tool row is deleted.

---

### 🧩 Affected Component
- [x] `mcpgateway` - API

---

### 🔁 Steps to Reproduce

1. Register an A2A agent (which auto-creates a corresponding tool and may associate it with a virtual server)
2. Attempt to delete the A2A agent from the Admin UI A2A agents list
3. Alternatively, attempt to delete the associated tool from the Admin UI tools list

Both paths fail with the same FK constraint error.

---

### 🤔 Expected Behavior
The A2A agent and its associated tool should be deleted successfully, with all dependent rows in `server_tool_association` cleaned up first.

---

### 📓 Logs / Error Output
```
ERROR [mcpgateway.services.structured_logger] [tool_service] Tool deletion failed
ERROR [mcpgateway.admin] Error deleting A2A agent: Failed to delete tool: (sqlite3.IntegrityError) FOREIGN KEY constraint failed
[SQL: DELETE FROM tools WHERE tools.id = ?]
[parameters: ('f9cb07ad1b29483ca06c25ad61d7d4b5',)]
(Background on this error at: https://sqlalche.me/e/20/gkpj)
```

---

### 🧠 Environment Info
| Key | Value |
|-----|-------|
| Runtime | Python 3.11+ |
| Database | SQLite |

---

**Root cause:** `ToolService.delete_tool()` in `tool_service.py` uses a Core SQL `DELETE` statement (`delete(DbTool).where(...)`) which bypasses SQLAlchemy ORM cascade rules. The `server_tool_association` table has a FK to `tools.id` with **no `ondelete` cascade** at the DB level, so the delete is rejected by SQLite's FK enforcement.

Other FK references to `tools.id` are fine:
- `tool_metrics` — explicitly purged before deletion, and has ORM `cascade="all, delete-orphan"`
- `tool_metrics_hourly` — has `ondelete="SET NULL"` at DB level
- `a2a_agents.tool_id` — has `ondelete="SET NULL"` at DB level

**Fix:** `delete_tool()` must delete `server_tool_association` rows for the tool before executing the tool DELETE. The gateway service already uses this pattern (e.g., `gateway_service.py:1605`).

madhav165:
# 🐛 Bug-fix PR

---

## 📌 Summary
Deleting an A2A agent (or its associated tool) from the Admin UI fails with `sqlite3.IntegrityError: FOREIGN KEY constraint failed` because `server_tool_association` rows referencing the tool are not cleaned up before the tool row is deleted.

## 🔁 Reproduction Steps
Closes #4261

1. Register an A2A agent (auto-creates a tool and may associate it with a virtual server)
2. Attempt to delete the A2A agent from the Admin UI
3. Observe: `FOREIGN KEY constraint failed` on the `tools` table

## 🐞 Root Cause
`ToolService.delete_tool()` uses a Core SQL `delete(DbTool).where(...)` statement which bypasses SQLAlchemy ORM cascades. The `server_tool_association` table has a FK to `tools.id` with **no `ondelete` cascade** at the DB level, so the delete is rejected when association rows exist.

Other FK references to `tools.id` are unaffected:
- `tool_metrics` — explicitly purged, ORM `cascade="all, delete-orphan"`
- `tool_metrics_hourly` — `ondelete="SET NULL"`
- `a2a_agents.tool_id` — `ondelete="SET NULL"`

## 💡 Fix Description
Added an explicit `DELETE FROM server_tool_association WHERE tool_id = ?` before the tool row deletion in `delete_tool()`. This matches the existing pattern used in `gateway_service.py` for stale tool cleanup.

## 🧪 Verification

| Lint suite                            | `make lint`          | ⬜     |
| Coverage ≥ 80 %                       | `make coverage`      | ⬜     |
| Manual regression no longer fails     | steps / screenshots  | ⬜     |

- All 8 `delete_tool` unit tests pass (including new `TestDeleteToolServerAssociationCleanup`)
- All 4 `delete_a2a_agent` admin tests pass

## ✅ Checklist

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-04-26T18:36:21Z",
+  "generated_at": "2026-04-26T22:08:03Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -8584,7 +8584,7 @@
         "hashed_secret": "7b1552c7c7ffb8bd70b5666e5997c8e017630aab",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1981,
+        "line_number": 1984,
         "type": "Base64 High Entropy String",
         "verified_result": null
       },
@@ -8592,7 +8592,7 @@
         "hashed_secret": "9fb7fe1217aed442b04c0f5e43b5d5a7d3287097",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 3318,
+        "line_number": 3321,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -8600,7 +8600,7 @@
         "hashed_secret": "72cb70dbbafe97e5ea13ad88acd65d08389439b0",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 3951,
+        "line_number": 3954,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -8608,7 +8608,7 @@
         "hashed_secret": "ee977806d7286510da8b9a7492ba58e2484c0ecc",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 7307,
+        "line_number": 7310,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -8616,7 +8616,7 @@
         "hashed_secret": "f2e7745f43b0ef0e2c2faf61d6c6a28be2965750",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 7799,
+        "line_number": 7802,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -8624,7 +8624,7 @@
         "hashed_secret": "4a249743d4d2241bd2ae085b4fe654d089488295",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 9146,
+        "line_number": 9149,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -8632,7 +8632,7 @@
         "hashed_secret": "0c8d051d3c7eada5d31b53d9936fce6bcc232ae2",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 9288,
+        "line_number": 9291,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -8640,7 +8640,7 @@
         "hashed_secret": "f2b14f68eb995facb3a1c35287b778d5bd785511",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 9770,
+        "line_number": 9773,
         "type": "Secret Keyword",
         "verified_result": null
       }
@@ -8674,7 +8674,7 @@
         "hashed_secret": "0857abc2d90c5d9821229fc0a880f1c38ffb0e04",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 4066,
+        "line_number": 4125,
         "type": "Hex High Entropy String",
         "verified_result": null
       },
@@ -8682,7 +8682,7 @@
         "hashed_secret": "99834bc4eff3f1e1c1e4692d2476b593b501d045",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 6993,
+        "line_number": 7052,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -8690,7 +8690,7 @@
         "hashed_secret": "f2b14f68eb995facb3a1c35287b778d5bd785511",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 7011,
+        "line_number": 7070,
         "type": "Secret Keyword",
         "verified_result": null
       },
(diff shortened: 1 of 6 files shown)


shoummu1:
@madhav165 This is a real bug and the fix is correct.

**Root cause** is clear: `delete_tool()` uses a Core SQL `delete(DbTool).where(...)` statement, which bypasses SQLAlchemy's ORM cascade machinery entirely. Since `server_tool_association` has no `ondelete="CASCADE"` at the DB level, the FK constraint fires when a tool associated with a virtual server is deleted. The fix — an explicit `DELETE FROM server_tool_association WHERE tool_id = ?` before the tool row delete — is exactly right and consistent with how `gateway_service.py` handles the same table in 5 other places.

**I verified the sibling tables are not affected:**
- `delete_resource()` and `delete_prompt()` use ORM-style `db.delete(obj)` — SQLAlchemy automatically cleans up `server_resource_association` / `server_prompt_association` via the `secondary=` relationship.
- `delete_agent()` (A2A) also uses `db.delete(agent)` — `server_a2a_association` is handled by the ORM. The tool deletion it delegates to (`delete_tool_from_a2a_agent` → `delete_tool`) is now covered by this fix.

This is the only place in the service layer where Core SQL delete is used on a table with unguarded FK references.

**Tests** are well-structured — the new `TestDeleteToolServerAssociationCleanup` class verifies ordering (association cleanup fires before tool delete), and existing call counts are correctly updated for both the basic and `purge_metrics` paths.

**Blocking:** There is a merge conflict in `.secrets.baseline`that needs to be resolved. 

LGTM on the code itself ✅

jonpspri:
Reviewed end-to-end with a real-SQLite empirical reproduction across all four entity-deletion paths. The fix here is correct, minimal, and faithfully mirrors the pattern `gateway_service.py` already uses in five places. CI is green and tests pass locally.

**Empirical scorecard** (in-memory SQLite, `PRAGMA foreign_keys=ON`, current `db.py` schema):

| Path | Mechanism | Verdict |
|---|---|---|
| `tool_service.delete_tool()` | Core SQL `delete(DbTool)` | 🐞 violates `server_tool_association` FK without this fix; ✅ OK with it |
| `resource_service.delete_resource()` | ORM `db.delete(resource)` | ✅ UoW auto-cleans `server_resource_association` |
| `prompt_service.delete_prompt()` | ORM `db.delete(prompt)` | ✅ UoW auto-cleans `server_prompt_association` |
| `a2a_service.delete_agent()` | ORM `db.delete(agent)` (preceded by `delete_tool()` for the agent's tool) | ✅ Both association tables cleaned end-to-end |

So the fix is **complete for tools**, and the three sibling paths don't need parallel changes today.

### What I did to the branch

I took the liberty of doing some adjacent work on the branch — happy to revert any of it if you'd prefer it on a separate PR:

1. **Rebased onto current `main`** (cleanly), dropping three intermediate `.secrets.baseline` regeneration commits that were causing conflicts and adding no semantic value beyond what pre-commit produces locally. Substantive diff is preserved 1:1.
2. **Added a concurrency-UX fix** (commit \`36c168026\`): \`delete_tool()\` now uses \`get_for_update(db, DbTool, tool_id, nowait=True)\`, mirroring \`set_tool_state()\`. This serializes concurrent associate/delete operations on the same tool so a racing INSERT into \`server_tool_association\` surfaces as a 409 \`ToolLockConflictError\` rather than an opaque 500 from a downstream FK violation. Wired the new exception through \`admin.py\` and \`main.py\`. Includes a unit test for the lock-conflict path.
3. **Fixed a pre-existing exception-propagation bug surfaced while in the area**: \`delete_tool()\`'s catch-all \`except Exception\` was wrapping \`ToolNotFoundError\` into \`ToolError\`, which prevented the FastAPI route in \`main.py\` from returning 404 (it was always falling through to the 400 catch-all). Bundled into the same commit since it's the same exception-propagation story.
4. **Cleaned up a stale docstring** in \`tests/integration/test_concurrency_row_locking.py:526\` that referred to a \"DELETE … RETURNING\" implementation that no longer exists (commit \`f25873fa6\`).

### Filed for follow-up

**#4449** — \`[CHORE]: Add ondelete=CASCADE to server_*_association FKs for defense-in-depth\`. Schema-level fix that would let all four \`server_*_association\` FKs handle this correctly at the DB level, so future Core-SQL refactors of the sibling delete paths can't reintroduce the same class of bug. Out of scope for this PR; tracked for a future change.

### Two adjacent observations (not action items for this PR)

A couple of things I won't file as separate issues since they're either policy-level or already tracked:

- **Audit trail**: \`delete_tool\` doesn't audit the un-association of the tool from virtual servers, but this is consistent with the codebase-wide treatment of association mutations across \`tool_service\`, \`resource_service\`, \`prompt_service\`, \`gateway_service\`, and \`server_service\` — not a regression and would need a holistic decision rather than a one-off fix.
- **Server-level \`notifications/tools/list_changed\`**: virtual servers with active SSE/WebSocket subscribers don't get a push notification when one of their tools is deleted here. That's a concrete symptom of the gateway-server→client-relay gap already tracked under the open epic #4302.

Mentioning so they're on the record but not asking you to address them in this PR.

### Bottom line

LGTM. Approving once #4449 is filed (done) and the new commits CI-green on push.

brian-hussey (approved):
Approved based on others feedback.

PR #4217, opened 2026-04-15
fix: handle root_path in MCPPathRewriteMiddleware for reverse proxy deployments

MohanLaksh:
## Summary

Fixes #4215

The documented canonical MCP path pattern `/servers/{id}/mcp` was returning 404 in reverse proxy deployments with path prefixes (e.g., `/dev/mcp-gateway/service/gateway`). This occurred because `MCPPathRewriteMiddleware` was checking `original_path` without stripping the `root_path` prefix before pattern matching.

## Root Cause

PR #3892 added security validations that check `if original_path.startswith("/servers/")` to prevent arbitrary path rewrites. However, in reverse proxy deployments where the ASGI scope contains a `root_path` prefix, the full incoming path is `/prefix/servers/{id}/mcp` which doesn't match the `/servers/` check, causing the middleware to skip the rewrite and return 404.

## Solution

Following established patterns in `streamablehttp_transport.py:831` and `token_scoping.py:354`, the fix:

1. **Extracts root_path** from `scope["root_path"]` or falls back to `settings.app_root_path`
2. **Strips the prefix** to get the application-relative path for pattern matching
3. **Preserves the prefix** when rewriting to `/mcp/` so downstream handlers receive the correct path
4. **Maintains all security checks** from PR #3892 (empty server ID rejection, arbitrary path prevention)

### `mcpgateway/main.py`
- Added root_path extraction and stripping logic (~15 lines)
- Changed all pattern matching to use `app_path` instead of `original_path`
- Preserved root_path prefix when rewriting to `/mcp/`

### `tests/unit/mcpgateway/test_main_extended.py`
- Added 8 comprehensive unit tests covering:
  - Single-level and multi-level root_path prefixes
  - `APP_ROOT_PATH` fallback scenario
  - Workaround pattern (`/mcp/servers/{id}`) with and without prefixes
  - Security checks (arbitrary prefix rejection, empty server ID rejection) with prefixes

## Test Results

✅ All 1,532 relevant tests passing:
- `test_main.py`: 275 tests
- `test_main_extended.py`: 573 tests (including 8 new tests)
- `test_streamablehttp_transport.py`: 440 tests
- `test_token_scoping.py`: 244 tests

## Backward Compatibility

Both URL patterns continue working:
- ✅ `/servers/{id}/mcp` (documented canonical pattern) - **NOW FIXED**
- ✅ `/mcp/servers/{id}` (undocumented workaround) - continues working

## Security

All security validations from PR #3892 are preserved:
[...]

PR #4270, opened 2026-04-17
fix: update scope['modified_path'] for server_id extraction with root_path

### LINKED ISSUE #4266: [BUG]: /mcp endpoint returns all tools instead of server-scoped tools with root_path configured
## Bug Description

The `/mcp` endpoint returns ALL tools instead of server-scoped tools when accessed via `/servers/{id}/mcp` in reverse proxy deployments with `root_path` configured. The `/sse` endpoint works correctly.

## Root Cause

PR #4217 introduced a bug in `MCPPathRewriteMiddleware` (main.py:3014) that breaks server-scoped tool filtering. The middleware sets:

```python
scope["modified_path"] = original_path  # Line 3014
```

Then strips `root_path` to create `app_path` for pattern matching:

```python
root_path = (scope.get("root_path") or settings.app_root_path or "").rstrip("/")
app_path = _normalize_scope_path(original_path, root_path)  # Line 3021
```

**However, `scope["modified_path"]` is never updated to use `app_path`.**

When `streamablehttp_transport.py:2898` reads `scope["modified_path"]` and tries to extract the server_id:

```python
path = scope["modified_path"]  # Gets "/prefix/servers/123/mcp" with root_path
match = _SERVER_ID_RE.search(path)  # Regex: r"^/servers/(?P<server_id>[^/]+)/mcp"
```

The regex **fails to match** because the path starts with `/prefix/servers/` instead of `/servers/`, so `server_id` becomes `None`. This causes the code to fall back to unscoped tool listing (all tools) instead of server-scoped tools.

## Impact

- **Severity**: High - breaks multi-tenancy isolation in reverse proxy deployments
- **Scope**: Affects `/mcp` endpoint only; `/sse` endpoint works correctly because it doesn't rely on `modified_path`
- **Security**: Tools from ALL servers are exposed when only one server's tools should be visible

## Steps to Reproduce

1. Configure `APP_ROOT_PATH=/prefix` (or deploy behind reverse proxy with path prefix)
2. Create two virtual servers with different tools
3. Connect via `/servers/server-1/mcp`
4. Send `tools/list` request
5. **Actual**: Returns tools from ALL servers
6. **Expected**: Returns only tools from `server-1`

## Comparison with `/sse` Endpoint

The `/sse` endpoint (main.py:4207-4229) works correctly because:
- It extracts `server_id` directly from the FastAPI route parameter: `async def sse_endpoint(request: Request, server_id: str, ...)`
- It calls `_enforce_scoped_resource_access()` which validates server access
- It passes `server_id` explicitly to `session_registry.respond(server_id, user_with_token, session_id=transport.session_id)`

The `/mcp` endpoint relies on `scope["modified_path"]` for regex-based server_id extraction, which breaks when the path contains a root_path prefix.

## Proposed Fix

Update `scope["modified_path"]` after computing `app_path`:

```python
# mcpgateway/main.py around line 3021
root_path = (scope.get("root_path") or settings.app_root_path or "").rstrip("/")
app_path = _normalize_scope_path(original_path, root_path)

# ADD THIS LINE:
scope["modified_path"] = app_path
```

This ensures `streamablehttp_transport` receives the app-relative path for regex matching, allowing server_id extraction to work correctly.

## Related

- Introduced in: PR #4217 (commit 72a5fd8bf)
- Original issue: #4215
- Regex definition: `streamablehttp_transport.py:207` - `_SERVER_ID_RE = r"^/servers/(?P<server_id>[^/]+)/mcp"`
- Server ID extraction: `streamablehttp_transport.py:2898-2900`

MohanLaksh:
## Summary

Fixes #4266

The `/mcp` endpoint was returning ALL tools instead of server-scoped tools when accessed via `/servers/{id}/mcp` in reverse proxy deployments with `root_path` configured. The `/sse` endpoint worked correctly.

## Root Cause

PR #4217 introduced a bug in `MCPPathRewriteMiddleware` where:
1. `scope["modified_path"]` was set to the original path (line 3019)
2. `app_path` was computed by stripping `root_path` prefix (line 3026)
3. **`scope["modified_path"]` was never updated** to use `app_path`

When `streamablehttp_transport.py:2898` reads `scope["modified_path"]` to extract `server_id` via regex `r"^/servers/(?P<server_id>[^/]+)/mcp"`, the regex fails to match paths with prefixes (e.g., `/prefix/servers/123/mcp`), causing `server_id` to be `None` and `tools/list` to return all tools instead of server-scoped tools.

## Solution

Update `scope["modified_path"] = app_path` after computing the app-relative path (line 3030). This ensures `streamablehttp_transport` receives the path without `root_path` prefix for correct `server_id` extraction.

### Code
- **mcpgateway/main.py** (line 3028-3030): Set `scope["modified_path"]` to `app_path` after stripping `root_path`

### Tests
- **tests/unit/mcpgateway/test_main_extended.py**:
  - `test_modified_path_set_to_app_relative_path`: Verifies `modified_path` is app-relative with `root_path`
  - `test_modified_path_without_root_path`: Verifies `modified_path` works without `root_path`

## Testing

```bash
# All MCPPathRewriteMiddleware tests pass (14/14)
pytest tests/unit/mcpgateway/test_main_extended.py::TestMCPPathRewriteMiddleware -v
```

## Impact

- **Severity**: High - Fixes multi-tenancy isolation breach in reverse proxy deployments
- **Scope**: Affects `/mcp` endpoint only; `/sse` endpoint already works correctly
- **Breaking Changes**: None - this is a bug fix that restores expected behavior

## Comparison with `/sse` Endpoint

The `/sse` endpoint works correctly because it extracts `server_id` directly from the FastAPI route parameter, not via regex on `modified_path`.

### DIFF
--- a/mcpgateway/main.py
+++ b/mcpgateway/main.py
@@ -3025,6 +3025,10 @@ class MCPPathRewriteMiddleware:
         root_path = (scope.get("root_path") or settings.app_root_path or "").rstrip("/")
         app_path = _normalize_scope_path(original_path, root_path)
 
+        # Update modified_path to the app-relative path (without root_path prefix).
+        # This ensures streamablehttp_transport can extract server_id via regex (#4266).
+        scope["modified_path"] = app_path
+
         # Skip rewriting for well-known URIs (RFC 9728 OAuth metadata, etc.)
         # These paths may end with /mcp but should not be rewritten to the MCP transport
         if not app_path.startswith("/.well-known/"):
--- a/tests/unit/mcpgateway/test_main_extended.py
+++ b/tests/unit/mcpgateway/test_main_extended.py
@@ -2925,6 +2925,49 @@ class TestMCPPathRewriteMiddleware:
         app_mock.assert_called_once()
 
     @pytest.mark.asyncio
+    async def test_modified_path_set_to_app_relative_path(self):
+        """Regression test for #4266: modified_path should be app-relative for server_id extraction."""
+        app_mock = AsyncMock()
+        middleware = MCPPathRewriteMiddleware(app_mock)
+        scope = {
+            "type": "http",
+            "path": "/gateway/servers/abc123/mcp",
+            "root_path": "/gateway",
+            "headers": []
+        }
+        receive, send = AsyncMock(), AsyncMock()
+
+        with patch("mcpgateway.main.streamable_http_auth", return_value=True):
+            await middleware._call_streamable_http(scope, receive, send)
+
+        # modified_path MUST be app-relative (without root_path prefix)
+        # so streamablehttp_transport can extract server_id via regex
+        assert scope["modified_path"] == "/servers/abc123/mcp"
+        # path is rewritten with root_path prefix preserved
+        assert scope["path"] == "/gateway/mcp/"
+        app_mock.assert_called_once()
+
+    @pytest.mark.asyncio
+    async def test_modified_path_without_root_path(self):
+        """When no root_path, modified_path equals original path."""
+        app_mock = AsyncMock()
+        middleware = MCPPathRewriteMiddleware(app_mock)
+        scope = {
+            "type": "http",
+            "path": "/servers/xyz789/mcp",
+            "headers": []
+        }
+        receive, send = AsyncMock(), AsyncMock()
+
+        with patch("mcpgateway.main.streamable_http_auth", return_value=True):
+            await middleware._call_streamable_http(scope, receive, send)
+
+        # Without root_path, modified_path should equal the normalized path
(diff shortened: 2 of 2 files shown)


madhav165 (approved):
**Clean, no issues found.**

- The fix is correct and well-targeted
- Both downstream consumers (streamablehttp_transport.py and rust_mcp_runtime_proxy.py) benefit
- Test coverage is adequate -- two new tests cover the root_path-present and root_path-absent cases
- One minor observation: the initial scope["modified_path"] = original_path assignment is now effectively dead code, but it's pre-existing and not worth changing in this PR

No blocking concerns.

brian-hussey (approved):
Thanks for the submission @MohanLaksh, this one LGTM.

PR #4283, opened 2026-04-17
fix: enable MCP session pool and set max sessions per key to resolve stateful MCP server session isolation

### LINKED ISSUE #4205: [BUG]: Problem with session management for stateful MCP Servers
### 🐞 Bug Summary

There seem to be a problem with a ContextForge session management when running against stateful MCP Gateways over HTTP Streamable transport

---

### 🧩 Affected Component

- [ ] `mcpgateway` - API
- [*] Federation or Transports
- [ ] Other (explain below)

---

### 🔁 Steps to Reproduce
1. Start a stateful MCP Gateway for example : ```git clone `https://github.com/modelcontextprotocol/rust-sdk/; cd rust-sdk/examples/servers; cargo run --example servers_counter_streamhttp```. This should start MCP counter server on 127.0.0.1:8000
2. ```make serve``` and configure context forge (add servers and gateways, etc.)
3. From a client Initialize MCP session (mcp inspector or curl)
4. Increase the counter N times (mcp inspector or curl)
5. Get the counter -> response should be N (mcp inspector or curl)
---

### 🤔 Expected Behavior
Minimally, we would expect that if there is only one client then the counter works for that single client. 

Long term expected behavior is that the counter stays consistent per mcp-session. So if for every client that goes through Intialize/Notify process, there should be a unique session with an independent counter. Unfortunately, this doesn't seem to be the case. It looks like ContextForge will create ONE session with the MCP server at startup and will re-use this session for all future interactions.

---
### 📓 Logs / Error Output

---
### 🧠 Environment Info
Commit #77bd03d2cca41f9c3597d5d7abc8012ad54d38fa
---

Lang-Akshay:
Closes #4205

---

## 📝 Summary

ContextForge was creating a **single shared MCP session** at startup and reusing it for all downstream clients. This caused stateful MCP servers (e.g. a counter server over Streamable HTTP transport) to leak state across clients — every client saw the same counter, regardless of their individual session lifecycle.

### Root Cause

The session pool (`MCP_SESSION_POOL_ENABLED`) defaulted to `false` in `.env.example`, which meant each request created a fresh session (poor for performance) or, worse, a single shared session was reused across all clients with no per-client isolation.

### Fix

Enable the MCP session pool in `.env.example` with sensible defaults:

```env
MCP_SESSION_POOL_ENABLED=true
MCP_SESSION_POOL_MAX_PER_KEY=10
```

- **`MCP_SESSION_POOL_ENABLED=true`** — activates the session pool so the gateway manages a pool of upstream MCP sessions keyed by `(URL, identity, transport)`.
- **`MCP_SESSION_POOL_MAX_PER_KEY=10`** — caps the number of pooled sessions per upstream server at 10, providing per-client session isolation while bounding resource usage.

With the pool enabled, each downstream client is assigned its own upstream session from the pool. Stateful servers (counters, carts, workflows) now correctly track per-client state throughout the MCP session lifecycle (Initialize → tool calls → teardown).


jonpspri:
Closing as superseded. The linked issue #4205 ([BUG]: Problem with session management for stateful MCP Servers) was closed as COMPLETED — the fix has already been merged. This PR is no longer needed.

Issue #4276, opened 2026-04-17
Dynamic Plugin Configuration — Multi-Instance Gaps and Fix Plan


gandhipratik203:
## Context

The dynamic plugin configuration and per-tool bindings added in #4068 / #4143 built the data layer (DB table, API, schemas) and single-worker runtime (tenant manager factory, cache, reload). Everything from "how does this propagate beyond one worker" onwards is missing. The API docs claim "invalidated immediately" without qualifying that it only affects one Gunicorn worker in one pod.

Related issue: #4230 (Plugin runtime management — testing and hardening)

## Current Architecture

- Plugin config loaded from YAML at startup, cached in `_managers` dict per-process
- `_PLUGINS_ENABLED` is a per-process Python global — not shared
- `ToolPluginBinding` rows stored in Postgres via `/v1/tools/plugin_bindings` API
- On upsert/delete, `reload_plugin_context(ctx_id)` evicts cache entry in the current worker only
- `TenantPluginManagerFactory` builds managers lazily on first request per context_id
- No Redis, no pub/sub, no IPC for cross-worker/cross-pod state

### HIGH — Multi-instance propagation and missing features

| # | Gap | What's broken / missing | Where |
|---|---|---|---|
| G1 | Cache invalidation is per-worker only | `reload_plugin_context()` evicts cache in one Gunicorn worker. Other workers in the same pod and all workers in other pods keep stale config indefinitely | `framework/__init__.py:188-200`, `routers/tool_plugin_bindings.py:95-96` |
| G2 | Wildcard binding invalidation mismatch | Upsert of `tool_name="*"` creates context_id `"team::*"` but tools request `"team::my_tool"`. Eviction is a no-op — team-wide policies never propagate, even on the local worker | `routers/tool_plugin_bindings.py:95`, `services/tool_service.py:4456-4457` |
| G3 | No cross-pod state propagation mechanism | No Redis pub/sub, no shared cache, no IPC. Pods are completely isolated — a binding change on pod A is invisible to pods B and C forever | `framework/manager.py:1064-1067` |
| G4 | No global plugin toggle in shared store | `_PLUGINS_ENABLED` is a per-process Python global. `PUT /admin/plugins` toggles one worker only. No Redis/DB backing | `framework/__init__.py:67-80` |
[...]

Issue #4230, opened 2026-04-15
[TESTING]: Plugin runtime management — global enable/disable, RateLimiterPlugin dynamic config, multi-instance validation


gandhipratik203:
## Summary

The dynamic plugin configuration and per-tool bindings added in #4068 / #4143 are the supported way to manage plugins at runtime. This issue covers two pillars of plugin runtime management:

1. **Global plugin enable/disable** — toggle any plugin on/off at runtime via the API, verify it takes effect across all gateway instances
2. **RateLimiterPlugin dynamic configuration** — change rate limiter settings at runtime (limits, mode, scope), verify the new config takes effect across all instances

Currently there are no tests that verify either capability works correctly across multiple gateway instances. In a 3-pod deployment, a change made via the API on pod A must be visible to pods B and C — and we don't currently test or guarantee that.

This issue covers auditing the propagation mechanism, writing tests that expose gaps, and fixing any bugs found.

## Context

- The `enable_plugins()` function in `mcpgateway/plugins/framework/__init__.py` is a per-process in-memory toggle — it does not propagate across pods
- The plugin config loader (`mcpgateway/plugins/framework/loader/config.py`) supports Jinja2 env var resolution but reads from a static YAML file at startup
- Tool plugin bindings (`/v1/tools/plugin_bindings`) persist to Postgres — but it's unclear whether the plugin manager reads these per-request or caches them in-memory
- Rate limiter uses Redis as shared state for rate counting (works cross-pod), but the plugin enable/disable and config state may not be shared

Related: #4222 (SQL Sanitizer e2e tests), PII Filter e2e tests — both depend on the plugin runtime management infrastructure being reliable across instances.

## Approach

Test-driven hardening — write the tests first, let failures expose bugs, fix the bugs, tests become the proof.

### 1. Audit — map the state flow

Trace the plugin runtime management path end-to-end:
- API call → DB write → plugin manager reload → request pipeline
- Identify where state is per-process (in-memory) vs shared (DB/Redis)
- Document the cache TTLs and invalidation mechanisms (if any)
- Specifically: when a binding is created/updated via `/v1/tools/plugin_bindings`, how does each pod learn about it?

### 2. Integration tests (`tests/integration/test_plugin_runtime_management.py`)

Against a live gateway (not mocked):

**Global enable/disable:**
- Enable a plugin via API, verify it takes effect on subsequent requests
- Disable a plugin via API, verify it stops running
[...]

PR #4143, opened 2026-04-13
feat: add binding_reference_id to tool plugin bindings and expand plugin config schemas

madhu-mohan-jaishankar:
## 🔗 Related Issue
Closes #4122 partially

---

## 📝 Summary
Adds `binding_reference_id` to `tool_plugin_bindings` to allow upstream systems (e.g. WXO) to correlate CF plugin bindings with their own identifiers. Also expands `OutputLengthGuardConfig` (4 → 12 fields) and `RateLimiterConfig` (3 → 8 fields, `by_tool` changed from `str` to `Dict[str, str]`) to fully reflect the capabilities of the underlying plugins. All unit tests were updated and extended to cover the new fields and behaviours (76/76 passing).

---

## 🏷️ Type of Change
- [ ] Bug fix
- [x] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Chore (deps, CI, tooling)
- [ ] Other (describe below)

---

## 🧪 Verification

| Check                     | Command         | Status |
|---------------------------|-----------------|--------|
| Lint suite                | `make lint`     |        |
| Unit tests                | `make test`     | ✅ 76/76 passing |
| Coverage ≥ 80%            | `make coverage` |        |

---

## ✅ Checklist
- [ ] Code formatted (`make black isort pre-commit`)
- [x] Tests added/updated for changes
- [x] Documentation updated (if applicable)
- [x] No secrets or credentials committed

---

## 📓 Notes (optional)

**DB / Migration**
- New nullable `binding_reference_id` column (`VARCHAR(255)`) on `tool_plugin_bindings`, with an index for fast lookups.
- Alembic migration: `d3e4f5a6b7c8_add_binding_reference_id_to_tool_plugin_bindings.py` (reversible).

**Schema changes**
- `OutputLengthGuardConfig` — added `max_chars` (Optional), `min_tokens`, `max_tokens`, `chars_per_token`, `limit_mode`, `word_boundary`, `max_text_length`, `max_structure_size`, `max_recursion_depth`; `min < max` validator covers both chars and tokens.
- `RateLimiterConfig` — added `algorithm`, `backend`, `redis_url`, `redis_key_prefix`, `redis_fallback`; `by_tool` is now `Dict[str, str]` (per-tool rate strings).
- `PluginPolicyItem` / `ToolPluginBindingResponse` — `binding_reference_id: Optional[str]` added; `config` docstring clarified to reflect full-replacement semantics (no partial configs, no merge).

**Service**
[...]

PR #4292, opened 2026-04-18
feat(plugins): runtime plugin management — global toggle, per-plugin mode, cross-instance propagation

gandhipratik203:
## Summary

Adds runtime plugin management capabilities to ContextForge — enabling global plugin enable/disable, per-plugin mode changes, and cross-worker/cross-pod state propagation via Redis pub/sub. Previously, plugin configuration was static (YAML loaded at startup) and changes required pod restarts.

The dynamic plugin configuration and per-tool bindings added in #4068 / #4143 built the data layer and single-worker runtime. This PR closes the multi-instance gaps by adding Redis-backed shared state, pub/sub-driven instant invalidation, TTL-based cache refresh as a safety net, and the missing API endpoints.

New endpoints:
- `PUT /admin/plugins` — global enable/disable (writes to Redis, broadcasts via pub/sub)
- `PUT /admin/plugins/{name}` — per-plugin mode change (enforce/enforce_ignore_error/permissive/disabled)
- `GET /admin/plugins` — now includes `plugins_globally_enabled` field from runtime state

Both endpoints use Pydantic request/response models (`PluginToggleRequest`, `PluginModeUpdateResponse`, etc.) for typed validation instead of raw dict parsing.

Related issues: #4276 (gap analysis), #4230 (testing plan)

---

### HIGH — Multi-instance propagation and missing features

**Gap 1 (HIGH)** — Cache invalidation per-worker only: `reload_plugin_context()` evicted cache in one Gunicorn worker only. Other workers in the same pod and all workers in other pods kept stale config indefinitely. Fixed with Redis pub/sub for instant cross-worker/cross-pod cache eviction, with TTL-based cache refresh (default 30s) as a safety net if pub/sub messages are missed.

**Gap 2 (HIGH)** — Wildcard binding invalidation mismatch: Upsert of `tool_name="*"` created context_id `"team::*"` but tools requested `"team::my_tool"`. Eviction was a no-op — team-wide policies never propagated. Fixed with `factory.invalidate_team()` that evicts ALL cached contexts for the team locally, plus a `team_binding_change` pub/sub message that triggers the same eviction on all remote workers.

**Gap 3 (HIGH)** — No cross-pod state propagation: Pods were completely isolated. A binding change on pod A was invisible to pods B and C forever. Fixed with Redis pub/sub channel `plugin:invalidation` carrying four typed message types (`global_toggle`, `mode_change`, `binding_change`, `team_binding_change`). The pub/sub listener is wired into the FastAPI lifespan with exponential backoff + jitter on reconnect.

**Gap 4 (HIGH)** — No global plugin toggle in shared store: `_PLUGINS_ENABLED` was a per-process Python global. Fixed with `PUT /admin/plugins` that writes to Redis (`plugin:global:enabled`). Workers check a 2-second in-memory cache (invalidated instantly by pub/sub) instead of hitting Redis per request.

**Gap 5 (HIGH)** — No per-plugin mode change via API: Could not switch a plugin from `permissive` to `enforce` globally at runtime. Fixed with `PUT /admin/plugins/{name}` that stores mode in Redis (`plugin:{name}:mode`, 24h TTL), broadcasts a `mode_change` pub/sub message, and maintains an in-process `_local_mode_overrides` dict so single-node deployments work without Redis.

**Gap 6 (HIGH)** — `GET /admin/plugins` didn't reflect runtime state: Returned YAML boot-time config. Fixed by reading `plugins_globally_enabled` from Redis. The response now shows actual runtime state.

**Gap 7 (HIGH)** — No TTL on plugin manager cache: `self._managers` dict had no expiry — stale config was permanent until worker restart. Fixed with configurable TTL (default 30s) on `TenantPluginManagerFactory`. Managers older than TTL are automatically evicted and rebuilt from DB. Cache entries use a frozen `_CachedManager` dataclass with an `.is_expired()` method.

### MEDIUM — Resilience and observability

**Gap 8 (MEDIUM)** — DB error kills manager build: If Postgres was briefly unavailable during `get_config_from_db()`, the exception propagated and tool invocations failed. Fixed by re-raising the error so the build fails loudly — this prevents silently dropping per-team security plugin bindings (rate limiter, PII filter) that would otherwise fall back to the permissive YAML defaults. The caller retries on the next request or cache TTL expiry.

**Gap 9 (MEDIUM)** — `_PLUGINS_ENABLED` not checked per-request: Once a manager was cached, the global flag wasn't rechecked. Fixed by checking a 2-second in-memory cache of the shared toggle on every `get_plugin_manager()` call. The cache is invalidated instantly by pub/sub, so the 2s TTL is only the worst-case bound when pub/sub is down.

**Gap 10 (MEDIUM)** — Concurrent upserts race on reload: Mitigated via TTL — even if invalidation is missed due to a race, the stale manager is rebuilt within 30s.

**Gap 11 (MEDIUM)** — `__global__` context never reloaded: Default context for HTTP middleware hooks was cached once. Now auto-refreshes via TTL, and `invalidate_all_plugin_managers()` includes it in the eviction sweep.

### LOW — Documentation and observability

**Gap 12 (LOW)** — No audit trail for runtime changes: Added structured logging for global toggle and per-plugin mode changes (user, action, timestamp, `redis_persisted` status).

**Gap 13 (LOW)** — No multi-instance caveats in docs: Updated `plugin-bindings-api.md` — cache invalidation is pub/sub-driven (instant) with TTL fallback (30s default), wildcard eviction behaviour documented.

**Gap 14 (LOW)** — No logging for missing team/tool bindings: Added `logger.debug` for empty binding lookups and `logger.error` (with `exc_info`) for DB errors.

---

### Before: per-process plugin state (no shared state)

```text
│  Pod A                                                                │
│                                                                       │
│  ┌─────────────────────┐  ┌─────────────────────┐                    │
│  │  Gunicorn Worker 1   │  │  Gunicorn Worker 2   │                    │
│  │                      │  │                      │                    │
│  │  _PLUGINS_ENABLED    │  │  _PLUGINS_ENABLED    │                    │
│  │  = True (in-memory)  │  │  = True (in-memory)  │                    │
│  │                      │  │                      │                    │
│  │  _managers: {        │  │  _managers: {        │                    │
│  │    "team::tool": mgr │  │    "team::tool": mgr │  ← separate copy  │
│  │  }  (no TTL)         │  │  }  (no TTL)         │                    │
│  └──────────┬───────────┘  └──────────────────────┘                    │
│             │ API call lands here                                      │
│             ▼                                                          │
│  PUT /v1/tools/plugin_bindings                                        │
│    1. Write binding to DB            ✓                                │
│    2. reload_plugin_context()        ✓ (Worker 1 only)                │
│    3. Worker 2 notified?             ✗ (stale forever)                │
│                                                                       │

│  Pod B                                                                │
│                                                                       │
│  ┌─────────────────────┐  ┌─────────────────────┐                    │
│  │  Gunicorn Worker 1   │  │  Gunicorn Worker 2   │                    │
│  │                      │  │                      │                    │
│  │  _managers: {        │  │  _managers: {        │                    │
│  │    "team::tool": mgr │  │    "team::tool": mgr │  ← stale forever  │
│  │  }                   │  │  }                   │                    │
│  └──────────────────────┘  └──────────────────────┘                    │
│                                                                       │
│  Pod B has no knowledge of binding changes on Pod A.                  │
│  No IPC, no shared cache, no pub/sub.                                 │
│                                                                       │

Problems:
  ✗ Binding change only affects 1 of N workers in 1 of M pods
  ✗ Wildcard binding ("*") evicts "team::*" but tools use "team::my_tool" — no-op
  ✗ No global enable/disable API
  ✗ No TTL — stale managers cached permanently
  ✗ GET /admin/plugins shows YAML boot-time config, not runtime state
```

### After: Redis-backed shared state with pub/sub + TTL fallback

```text
│  Pod A                                                                │
│                                                                       │
│  ┌──────────────────────────┐  ┌──────────────────────────┐          │
│  │  Gunicorn Worker 1        │  │  Gunicorn Worker 2        │          │
│  │                           │  │                           │          │
│  │  ┌─ Pub/sub listener ──┐ │  │  ┌─ Pub/sub listener ──┐ │          │
│  │  │  SUBSCRIBE           │ │  │  │  SUBSCRIBE           │ │          │
│  │  │  plugin:invalidation │ │  │  │  plugin:invalidation │ │          │
│  │  │  (lifespan-managed)  │ │  │  │  (lifespan-managed)  │ │          │
│  │  │  Reconnect: exp.     │ │  │  │  Reconnect: exp.     │ │          │
│  │  │  backoff + jitter    │ │  │  │  backoff + jitter    │ │          │
│  │  │  On message:         │ │  │  │  On message:         │ │          │
│  │  │   → evict cache      │ │  │  │   → evict cache      │ │          │
│  │  │   → update flag      │ │  │  │   → update flag      │ │          │
│  │  │   → mirror to local  │ │  │  │   → mirror to local  │ │          │
│  │  │     overrides dict   │ │  │  │     overrides dict   │ │          │
│  │  └──────────────────────┘ │  │  └──────────────────────┘ │          │
│  │                           │  │                           │          │
│  │  get_plugin_manager       │  │  get_plugin_manager       │          │
│  │    │                      │  │    │                      │          │
│  │    ├─ 2s cached toggle    │  │    ├─ 2s cached toggle    │          │
│  │    │  (invalidated by     │  │    │  (invalidated by     │          │
│  │    │   pub/sub instantly) │  │    │   pub/sub instantly) │          │
│  │    │                      │  │    │                      │          │
│  │    └─ Check TTL on cache  │  │    └─ Check TTL on cache  │          │
│  │       (30s safety net)    │  │       (30s safety net)    │          │
│  │                           │  │                           │          │
│  │  _local_mode_overrides    │  │  _local_mode_overrides    │          │
│  │    (in-process fallback   │  │    (in-process fallback   │          │
│  │     for no-Redis deploy)  │  │     for no-Redis deploy)  │          │
│  └──────────┬────────────────┘  └───────────────────────────┘          │
│             │                                                          │
│             ▼                                                          │
│  PUT /admin/plugins {"enabled": false}                                │
│    1. Write to Redis: plugin:global:enabled = "false"            ✓   │
│    2. PUBLISH {type: global_toggle, enabled: false}              ✓   │
│    3. All workers receive pub/sub → update flag + drop cache     ✓   │
│    4. 2s cached toggle ensures no Redis call per request         ✓   │
│                                                                       │
│  PUT /admin/plugins/RateLimiterPlugin {"mode": "enforce"}             │
│    1. Update _local_mode_overrides (instant, this worker)        ✓   │
│    2. Write to Redis: plugin:RateLimiterPlugin:mode (24h TTL)    ✓   │
│    3. PUBLISH {type: mode_change, plugin, mode, ttl_seconds}     ✓   │
│    4. All workers receive pub/sub → mirror to local overrides    ✓   │
│    5. Invalidate all cached managers → rebuild with new mode     ✓   │
│    6. If Redis is down: local override still works (this worker) ✓   │
│                                                                       │
│  PUT/DELETE /v1/tools/plugin_bindings                                 │
│    All three handlers (upsert, delete-by-ref, delete-by-UUID)        │
│    funnel through _invalidate_and_broadcast(bindings):                │
│                                                                       │
│    Wildcard (tool_name="*"):                                          │
│      1. Write to DB                                              ✓   │
│      2. factory.invalidate_team(team_id) — all team contexts     ✓   │
│      3. PUBLISH {type: team_binding_change, team_id}             ✓   │
│                                                                       │
│    Specific tool:                                                     │
│      1. Write to DB                                              ✓   │
│      2. reload_plugin_context(ctx_id) — this context only        ✓   │
│      3. PUBLISH {type: binding_change, context_id}               ✓   │
│                                                                       │
└────────────────────────────────┬─────────────────────────────────────┘
                                 │
                   ┌─────────────┴──────────────────┐
                   │                                │
                   ▼                                ▼
┌──────────────────────────────────┐  ┌──────────────────────────────┐
│  framework/ isolation modules    │  │  Redis (shared state)        │
│                                  │  │                              │
│  _redis.py (DI shim):            │  │  Keys:                      │
│    framework/ can't import from  │  │   plugin:global:enabled     │
│    mcpgateway.utils (pre-commit  │  │     = "true"/"false"        │
│    hook enforced).               │  │     (no TTL — kill-switch)  │
│    set_shared_redis_provider(fn) │  │                              │
│      → registered at lifespan    │  │   plugin:{name}:mode        │
│    get_shared_redis_client()     │  │     = "enforce"             │
│      → returns client or None    │  │     (24h TTL — auto-expire) │
│                                  │  │                              │
│  _state.py (shared mutable):     │  │  Pub/sub channel:           │
│    Breaks __init__ → manager →   │  │   plugin:invalidation       │
│    __init__ import cycle.        │  │    → global_toggle          │
│    _local_mode_overrides dict    │  │    → mode_change            │
│    set/prune/active helpers      │  │    → binding_change         │
│    Degraded-boot flag + logged   │  │    → team_binding_change    │
│      (two bools, three states)   │  │                              │
│                                  │  │  Messages are Pydantic      │
└──────────────────────────────────┘  │  discriminated union —       │
                                      │  match/case dispatch on      │
                                      │  the listener side           │
                                      └──────────────────────────────┘
                                                   │
                                                   ▼
│  Pod B                                                                │
│                                                                       │
│  ┌──────────────────────────┐  ┌──────────────────────────┐          │
│  │  Gunicorn Worker 1        │  │  Gunicorn Worker 2        │          │
│  │                           │  │                           │          │
│  │  ┌─ Pub/sub listener ──┐ │  │  ┌─ Pub/sub listener ──┐ │          │
│  │  │  Receives messages   │ │  │  │  Receives messages   │ │          │
│  │  │  from Pod A instantly │ │  │  │  from Pod A instantly │ │          │
│  │  │  → evicts cache      │ │  │  │  → evicts cache      │ │          │
│  │  │  → updates flag      │ │  │  │  → updates flag      │ │          │
│  │  │  → mirrors overrides │ │  │  │  → mirrors overrides │ │          │
│  │  └──────────────────────┘ │  │  └──────────────────────┘ │          │
│  │                           │  │                           │          │
│  │  get_plugin_manager       │  │  get_plugin_manager       │          │
│  │    │                      │  │    │                      │          │
│  │    ├─ 2s cached toggle    │  │    ├─ 2s cached toggle    │          │
│  │    │                      │  │    │                      │          │
│  │    └─ TTL expired?        │  │    └─ TTL expired?        │          │
│  │       (safety net only    │  │       (safety net only    │          │
│  │        — pub/sub already  │  │        — pub/sub already  │          │
│  │        evicted the cache) │  │        evicted the cache) │          │
│  └───────────────────────────┘  └───────────────────────────┘          │
│                                                                       │
│  Three layers of propagation (fastest wins):                          │
│    Layer 1: Pub/sub     → instant (~0ms)                             │
│    Layer 2: TTL expiry  → 0-30s (catches missed pub/sub)             │
│    Layer 3: Redis read  → 2s cached (global toggle only)             │
│                                                                       │
│  Pod B sees all changes from Pod A:                                   │
│    ✓ Global toggle    — instant (pub/sub invalidates 2s cache)       │
│    ✓ Per-plugin mode  — instant (pub/sub evicts + mirrors to local)  │
│    ✓ Per-tool bindings — instant (pub/sub evicts, rebuild reads DB)  │
│    ✓ Wildcard bindings — instant (team_binding_change evicts all)    │
│    ✓ Fallback         — 30s max if pub/sub missed the message        │
│                                                                       │

│  PostgreSQL (binding storage)                                         │
│                                                                       │
│  tool_plugin_bindings table:                                          │
│    (team_id, tool_name, plugin_id, mode, config, priority)           │
│                                                                       │
│  ← Written by /v1/tools/plugin_bindings API                         │
│  ← Read by get_config_from_db() on manager rebuild                   │
│  ← DB errors re-raise to prevent silent security-binding drops       │
│    (caller retries on next request or TTL expiry)                    │
│                                                                       │
```

### Request flow: tool invocation with plugin checks

```text
  Client request (tool/call)
         │
         ▼
  ┌──────────────────────────────────────────────────────────────┐
  │  Gateway (any worker)                                         │
  │                                                               │
  │  Background: pub/sub listener running (lifespan-managed)     │
  │    SUBSCRIBE plugin:invalidation                              │
  │    Pydantic discriminated union validates each frame          │
  │    On message → evict cache / update flag / mirror overrides  │
  │    (already done before this request arrives)                 │
  │                                                               │
  │  1. get_plugin_manager()                                      │
  │     │                                                         │
  │     ├── are_plugins_enabled_shared()                          │
  │     │   2s in-memory cache (pub/sub invalidates instantly)    │
  │     │   Falls back to Redis GET on miss, then in-memory flag │
  │     │         │                                               │
  │     │    No → return None (skip all plugins)                  │
  │     │    Yes ↓                                                │
  │     │                                                         │
  │     ├── _plugin_manager_factory is None?                      │
  │     │     │                                                   │
  │     │   Yes → _warn_factory_init_degraded_once()             │
  │     │         (one ERROR if node booted with failed init)     │
  │     │         return None                                     │
  │     │     │                                                   │
  │     │   No ↓                                                  │
  │     │                                                         │
  │     ├── Check _managers cache for context_id                  │
  │     │     │                                                   │
  │     │   Found?                                                │
  │     │     │                                                   │
  │     │   Yes → _CachedManager.is_expired(ttl)?                │
  │     │         No  → return cached manager                     │
  │     │         Yes → evict, fall through to build              │
  │     │                                                         │
  │     │   No (evicted by pub/sub or TTL) ↓                     │
  │     │                                                         │
  │     ├── _build_manager()                                      │
  │     │   │                                                     │
  │     │   ├─ DB: get_config_from_db()                          │
  │     │   │  (read bindings; re-raises on DB error to prevent  │
  │     │   │   silent security-binding drops)                    │
  │     │   │                                                     │
  │     │   ├─ _merge_tenant_config()                            │
  │     │   │  (DB overrides win over YAML)                      │
  │     │   │                                                     │
  │     │   ├─ _apply_redis_mode_overrides()                     │
  │     │   │  Priority: Redis MGET > _state.py overrides > YAML │
  │     │   │  Invalid values are skipped (logged), not fatal     │
  │     │   │  If Redis is down, _state.py overrides still apply │
  │     │   │                                                     │
  │     │   └─ Cache as _CachedManager (TTL = 30s safety net)    │
  │     │                                                         │
  │     └── Return manager                                        │
  │                                                               │
  │  2. manager.execute_hook                                      │
  │     (run plugins in priority order)                           │
  │                                                               │
  │  Admin: PUT /admin/plugins/{name} {"mode": "enforce"}        │
  │     Plugin name validated against factory._base_config        │
  │     (not the live manager — works even when plugins disabled, │
  │      allowing operators to pre-stage overrides before enable) │
  │                                                               │
  └──────────────────────────────────────────────────────────────┘
```

### Key design decisions

1. **Three-layer propagation** — pub/sub for instant notification (~0ms), TTL cache expiry as fallback (30s), cached Redis key read as confirmation (2s cache for global toggle). Fastest layer wins. Industry-standard pattern used by Kong, Envoy, and Consul.

2. **Pub/sub is notification, Redis keys are truth** — pub/sub tells workers "something changed, evict your cache." Workers then read the actual new state from Redis keys (global toggle, per-plugin mode) or DB (bindings) on the next request. If a pub/sub message is lost, TTL ensures the cache is rebuilt within 30s anyway.

3. **Dual-store fallback for mode overrides** — Redis is authoritative when available. An in-process `_local_mode_overrides` dict provides fallback for single-node/no-Redis deployments and write-through for the local worker. Redis-synced entries carry a TTL-aligned expiry (matching the Redis 24h key TTL) so both copies clear together. Local-only entries (Redis write failed) carry no expiry — the operator never got confirmation a timer started.

4. **Isolation modules for import-cycle prevention** — the plugin `framework/` package cannot import from `mcpgateway.utils` (enforced by a pre-commit hook). Two leaf modules break what would otherwise be circular dependencies:
   - `_redis.py` — dependency-inversion shim for Redis access. The gateway registers its Redis client provider at lifespan startup via `set_shared_redis_provider()`; framework modules call `get_shared_redis_client()` without naming the gateway util. When no provider is registered (tests, framework-only deployments), Redis-unavailable fallback paths engage automatically.
   - `_state.py` — shared mutable state (per-plugin mode overrides, degraded-boot flags). Breaks the `__init__` → `manager.py` → `__init__` import cycle: `manager.py` needs the local override map during rebuild, and importing it from `__init__.py` would close the loop. Exposes `set_local_mode_override()` / `prune_expired_local_overrides()` / `active_local_mode_overrides()` helpers so all mutation flows through a single seam.

5. **Factory init unconditional** — the plugin manager factory is initialized regardless of the local `plugins.enabled` setting. This allows a peer worker to runtime-enable plugins via `PUT /admin/plugins` without requiring a restart of nodes that booted with plugins disabled. If init fails and `plugins.enabled=true`, the gateway crashes (loud failure). If init fails and `plugins.enabled=false`, the gateway boots with a warning — runtime-enable from a peer will require a restart.

6. **MGET batched reads** — per-plugin mode overrides are fetched in a single Redis MGET call during manager rebuild, not N individual GETs. One round-trip regardless of plugin count.

7. **Typed pub/sub messages** — pub/sub frames are Pydantic `BaseModel` subclasses in a discriminated union (`_GlobalToggleMsg`, `_ModeChangeMsg`, `_BindingChangeMsg`, `_TeamBindingChangeMsg`). A typo in the `type` field on the producer fails Pydantic validation on the listener instead of silently no-oping.

8. **24h TTL on mode override keys** — stale overrides auto-expire. Prevents forgotten overrides from persisting indefinitely across gateway restarts.

9. **DB errors re-raise, not swallow** — `get_config_from_db()` re-raises on DB errors instead of returning `None`. This prevents silently dropping per-team security plugin bindings (rate limiter, PII filter) that would otherwise fall back to the permissive YAML defaults. The caller retries on the next request or cache TTL expiry.

10. **Degraded-boot signal** — when a node boots with `plugins.enabled=false` and the opportunistic factory init fails, `mark_factory_init_degraded()` is called. If a peer later flips the shared toggle to enabled, `get_plugin_manager()` emits one ERROR explaining that this node can't serve plugins and requires a restart. Prevents silent drift where a node appears healthy but silently ignores runtime-enable.

11. **Resilient invalidation on mode change** — `update_plugin_mode` wraps the post-write `invalidate_all_plugin_managers()` in try/except. The override is already stored in Redis + the local map before invalidation runs; a cache-refresh glitch must not turn an already-applied change into a 500. On failure the operator sees a WARNING and the TTL refresh reaches the new mode within 30s.

12. **Mode pre-staging on disabled nodes** — `PUT /admin/plugins/{name}` validates plugin names against `factory._base_config` (the YAML config loaded at startup), not against the live manager. This means operators can pre-stage mode overrides on a node with plugins disabled — the override takes effect when the shared toggle is flipped on, without requiring a second API call.

---

## Additional improvements

- **Plugin bindings API docs** — updated `docs/docs/api/plugin-bindings-api.md` with multi-instance cache propagation caveats.

- **`_invalidate_and_broadcast()` helper** — the wildcard-vs-specific invalidation split was copy-pasted across three binding handlers (upsert, delete-by-reference, delete-by-UUID). Extracted into a shared helper in `tool_plugin_bindings.py` so future changes land in one place.

- **`match/case` dispatch** — `_handle_invalidation_message` replaced four `isinstance(frame, X)` branches with structural pattern matching on the discriminated union. Makes union closure explicit and drops redundant return statements.

- **Test isolation** — autouse fixture in `tests/unit/mcpgateway/conftest.py` resets the framework's shared Redis provider between tests. Prevents lifespan-exercising tests from leaking a mock provider into subsequent tests via module-level state.

---

## Test results

<details>
<summary>Unit tests — plugin runtime management (65 tests)</summary>

```
tests/unit/mcpgateway/plugins/test_plugin_runtime_management.py

TestArePluginsEnabledShared (6):
  test_reads_true_from_redis                             PASSED
  test_reads_false_from_redis                            PASSED
  test_reads_bytes_from_redis                            PASSED
  test_falls_back_to_in_memory_when_redis_unavailable    PASSED
  test_falls_back_to_in_memory_when_redis_key_missing    PASSED
  test_falls_back_on_redis_exception                     PASSED

TestEnablePluginsShared (4):
  test_writes_true_to_redis                              PASSED
  test_writes_false_to_redis                             PASSED
  test_updates_in_memory_flag                            PASSED
  test_survives_redis_failure                             PASSED

TestGetPluginModeOverride (4):
  test_reads_mode_from_redis                             PASSED
  test_returns_none_when_no_override                     PASSED
  test_returns_none_when_redis_unavailable               PASSED
  test_raises_runtime_error_on_redis_exception           PASSED

TestTTLCacheExpiry (4):
  test_cache_returns_manager_within_ttl                  PASSED
  test_cache_evicts_after_ttl                            PASSED
  test_cache_ttl_default                                 PASSED
  test_cache_ttl_zero_disables                           PASSED

TestApplyRedisModeOverrides (10):
  test_no_plugins_short_circuits                         PASSED
  test_returns_config_unchanged_when_no_overrides        PASSED
  test_applies_per_plugin_override                       PASSED
  test_corrupt_redis_falls_through_to_local_override     PASSED
  test_invalid_mode_value_skipped_not_batch_aborted      PASSED
  test_local_override_applied_when_redis_has_nothing     PASSED
  test_redis_value_beats_local_override                  PASSED
  test_local_override_applied_when_redis_client_none     PASSED
  test_expired_redis_synced_local_override_is_pruned     PASSED
  test_mget_failure_warns_and_returns_input              PASSED

TestDBErrorFallback (3):
  test_raises_on_db_error                                PASSED
  test_returns_none_for_invalid_context_id               PASSED
  test_returns_none_for_empty_bindings                   PASSED

TestInvalidateAllPluginManagers (2):
  test_delegates_to_factory_invalidate_all               PASSED
  test_noop_when_factory_is_none                         PASSED

TestPubSubPublisher (2):
  test_global_toggle_publishes_message                   PASSED
  test_global_toggle_publish_failure_doesnt_crash        PASSED

TestPubSubSubscriber (4):
  test_subscriber_updates_flag_on_global_toggle          PASSED
  test_subscriber_evicts_managers_on_mode_change         PASSED
  test_subscriber_ignores_non_message_types              PASSED
  test_subscriber_handles_malformed_message              PASSED

TestPublishHelpers (5):
  test_publish_plugin_mode_change_sends_message          PASSED
  test_publish_plugin_mode_change_returns_false_on_fail  PASSED
  test_local_entry_expiry_aligns_with_redis_ttl          PASSED
  test_local_entry_has_no_expiry_when_redis_set_fails    PASSED
  test_publish_binding_change_sends_message              PASSED

TestPublishToListenerRoundTrip (2):
  test_mode_change_publish_triggers_factory_invalidate   PASSED
  test_global_toggle_pubsub_invalidates_local_cache      PASSED

TestFactoryInvalidateTeam (2):
  test_invalidates_only_matching_team                    PASSED
  test_uses_class_separator_when_not_specified           PASSED

TestTeamBindingChangeRoundTrip (2):
  test_publish_team_binding_change_emits_correct_frame   PASSED
  test_handler_evicts_every_per_tool_context_for_team    PASSED

TestSubscriberHandlesBindingChange (1):
  test_subscriber_handles_binding_change                 PASSED

TestListenerLocalMirror (3):
  test_mode_change_populates_local_map_with_ttl          PASSED
  test_mode_change_falls_back_to_default_ttl             PASSED
  test_backing_dict_identity_is_stable                   PASSED

TestListenerStartup (3):
  test_skips_start_when_no_redis_provider                PASSED
  test_concurrent_starts_only_create_one_task            PASSED
  test_probe_failure_does_not_start_and_logs             PASSED

TestPublishPartialWriteSignaling (2):
  test_enable_plugins_shared_errors_on_publish_failure   PASSED
  test_publish_plugin_mode_change_errors_on_publish_fail PASSED

TestFactoryInitFailureSurface (2):
  test_malformed_yaml_raises                             PASSED
  test_invalid_config_shape_raises                       PASSED

TestFactoryInitDegradedSignal (3):
  test_get_plugin_manager_emits_error_once_on_degraded   PASSED
  test_no_error_when_factory_is_initialized              PASSED
  test_no_error_when_toggle_is_off                       PASSED

TestListenerBackoff (1):
  test_escalates_to_error_after_five_consecutive_fails   PASSED

65 passed
```

All tests use mocked Redis — no external dependencies. Runs in CI.

</details>

<details>
<summary>Unit tests — admin plugin runtime endpoints (22 tests)</summary>

```
tests/unit/mcpgateway/test_admin_plugin_runtime.py

TestToggleGlobalPluginsRBAC (3):
  test_denies_when_permission_service_rejects            PASSED
  test_denies_admin_when_bypass_disabled                 PASSED
  test_denies_non_admin_user                             PASSED

TestToggleGlobalPluginsValidation (2):
  test_missing_enabled_rejected                          PASSED
  test_non_bool_enabled_rejected                         PASSED

TestToggleGlobalPluginsHappyPath (2):
  test_returns_redis_persisted_true_when_redis_ok        PASSED
  test_returns_redis_persisted_false_when_redis_down     PASSED

TestUpdatePluginModeRBAC (2):
  test_denies_without_permission                         PASSED
  test_denies_admin_when_bypass_disabled                 PASSED

TestUpdatePluginModeValidation (2):
  test_invalid_mode_rejected                             PASSED
  test_valid_modes_accepted                              PASSED

TestUpdatePluginModeHandler (4):
  test_404_on_unknown_plugin                             PASSED
  test_reports_redis_persisted_true                      PASSED
  test_single_node_no_redis_applies_override_locally     PASSED
  test_redis_set_failure_still_applies_locally           PASSED

+ additional tests for invalidation error handling,
  structured logging verification, and edge cases

22 passed
```

Tests cover RBAC enforcement, Pydantic validation, Redis/no-Redis paths,
and invalidation failure resilience.

</details>

<details>
<summary>Integration tests — Redis plugin runtime (22 tests)</summary>

```
tests/integration/test_plugin_runtime_redis.py

TestGlobalToggleRedis (6):
  test_write_true / test_write_false / test_read_true
  test_read_false / test_roundtrip / test_fallback

TestPluginModeOverrideRedis (3):
  test_read_value / test_missing_key / test_multiple_plugins

TestPubSubRedis (4):
  test_publish_message / test_handler_updates_flag
  test_ignore_subscribe / test_malformed_json

TestGetPluginManagerRedis (3):
  test_disabled_returns_none / test_no_factory / test_toggle_cycle

TestTTLCacheExpiryRedis (1):
  test_cache_expires_after_ttl

TestPubSubEvictionRedis (3):
  test_mode_change_evicts_all / test_binding_change_evicts_specific
  test_pubsub_roundtrip

TestDBErrorFallbackRedis (2):
  test_db_error_returns_none / test_redis_toggle_independent

22 passed
```

Tests use real Redis (auto-starts Docker container if not running).

</details>

<details>
<summary>Integration tests — docker-compose multi-replica (23 tests)</summary>

```
tests/integration/test_plugin_runtime_management.py

TestPluginsGloballyEnabledField (2)
TestPutAdminPluginsDisable (2)
TestPutAdminPluginsEnable (2)
TestPutAdminPluginsValidation (3)
TestPutAdminPluginsNameMode (5)
TestCrossReplicaPropagation (2)
TestMultiInstanceGlobalToggle (2)
TestMultiInstancePluginMode (1)
+ additional tests for endpoint validation and consistency

23 passed
```

Tests run against a live gateway (docker-compose with 3 replicas + NGINX).

</details>

---

## Limitations

1. **Plugin config stored in Postgres, not Redis** — full plugin config (rate limits, detection rules, etc.) is stored in Postgres and read on cache rebuild. For sub-second config propagation, a write-behind pattern (Redis-first with async Postgres replication) would be needed. Not required at current scale (4 plugins, infrequent config changes).

2. **Redis required for cross-instance consistency** — without Redis, the global toggle falls back to the per-process in-memory flag. Per-plugin mode overrides fall back to the `_local_mode_overrides` dict (so mode changes work on the local worker, but don't propagate to other workers/pods). The system degrades to single-instance behavior rather than failing.

3. **DB errors fail the manager rebuild** — a transient Postgres outage causes individual tool invocations to fail (no cached manager available) rather than silently falling back to YAML defaults. This is a deliberate trade-off: for security-critical plugins (rate limiter, PII filter), silently dropping per-team bindings is worse than a transient failure. The caller retries on the next request or cache TTL expiry.

4. **No integration tests in CI** — the 22 Redis integration tests and 23 docker-compose tests run locally. The 87 unit tests (65 + 22) run in CI. CI-compatible integration tests using TestClient + auto-started Redis are planned.

---

### Write-behind caching for plugin config

The current architecture writes plugin config to Postgres first, then workers read it on cache rebuild. For use cases requiring sub-second config propagation at scale, a **write-behind** pattern could be adopted:

```text
  Config update arrives
         │
         ├──1. Write to Redis (instant, 0.1ms)
         │     → API returns success immediately
         │     → All workers see new config via pub/sub
         │
         └──2. Background task: replicate to Postgres (async)
                → Picks up pending writes from Redis
                → Writes to Postgres for durability
                → Marks as "synced" in Redis
                → Retries with backoff on failure
```

This is standard in high-throughput systems (Shopify uses it for cart/inventory, Discord for presence, AWS ElastiCache best practices recommend it). The benefit is that the API call returns in 0.1ms instead of ~5ms, and config changes propagate instantly via pub/sub.

**Trade-off:** Eventual consistency between Redis and Postgres — if Redis dies before the async Postgres write completes, the config update is lost. Mitigation: reconciliation on startup (compare Redis state vs Postgres and resolve differences).

Not needed for the current scale (4 plugins, infrequent config changes), but a natural evolution if ContextForge grows to handle thousands of per-tool config updates per minute.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-04-19T08:51:02Z",
+  "generated_at": "2026-04-19T16:27:13Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -4170,7 +4170,7 @@
         "hashed_secret": "fa9beb99e4029ad5a6615399e7bbae21356086b3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 4184,
+        "line_number": 4187,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -4178,7 +4178,7 @@
         "hashed_secret": "559b05f1b2863e725b76e216ac3dadecbf92e244",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 4785,
+        "line_number": 4788,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -4186,7 +4186,7 @@
         "hashed_secret": "a8af4759392d4f7496d613174f33afe2074a4b8d",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 4787,
+        "line_number": 4790,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -4194,7 +4194,7 @@
         "hashed_secret": "85b60d811d16ff56b3654587d4487f713bfa33b7",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 15113,
+        "line_number": 15116,
         "type": "Secret Keyword",
         "verified_result": null
       }
--- a/charts/mcp-stack/profiles/ocp/values-pgo.yaml
+++ b/charts/mcp-stack/profiles/ocp/values-pgo.yaml
@@ -58,7 +58,7 @@ mcpContextForge:
         include_user_info: false
       plugins:
         - name: "PIIFilterPlugin"
-          kind: "plugins.pii_filter.pii_filter.PIIFilterPlugin"
+          kind: "cpex_pii_filter.PIIFilterPlugin"
           description: "Detects and masks Personally Identifiable Information"
           version: "0.1.0"
           author: "Mihai Criveti"
@@ -85,7 +85,7 @@ mcpContextForge:
               - "test@example.com"
               - "555-555-5555"
         - name: "RateLimiterPlugin"
-          kind: "plugins.rate_limiter.rate_limiter.RateLimiterPlugin"
+          kind: "cpex_rate_limiter.RateLimiterPlugin"
           description: "Per-user/tenant/tool rate limits"
           version: "0.1.0"
           author: "Mihai Criveti"
@@ -103,7 +103,7 @@ mcpContextForge:
             by_tenant: "3000/m"
             by_tool: {}
         - name: "RetryWithBackoffPlugin"
-          kind: "plugins.retry_with_backoff.retry_with_backoff.RetryWithBackoffPlugin"
+          kind: "cpex_retry_with_backoff.RetryWithBackoffPlugin"
           description: "Detects transient failures and asks the gateway to re-invoke the tool after a jittered exponential backoff delay"
           version: "0.1.0"
           author: "Mihai Criveti"
@@ -136,7 +136,7 @@ mcpContextForge:
             strategy: "truncate"
             ellipsis: "…"
         - name: "SecretsDetection"
-          kind: "plugins.secrets_detection.secrets_detection.SecretsDetectionPlugin"
+          kind: "cpex_secrets_detection.SecretsDetectionPlugin"
           description: "Detects keys/tokens/secrets in inputs/outputs; optional redaction/blocking"
           version: "0.1.0"
           author: "ContextForge"
@@ -160,7 +160,7 @@ mcpContextForge:
             block_on_detection: true
             min_findings_to_block: 1
         - name: "EncodedExfilDetector"
-          kind: "plugins.encoded_exfil_detection.encoded_exfil_detector.EncodedExfilDetectorPlugin"
+          kind: "cpex_encoded_exfil_detection.EncodedExfilDetectorPlugin"
           description: "Detects suspicious encoded exfiltration patterns in prompt args and tool outputs"
           version: "0.1.0"
           author: "Mihai Criveti"
--- a/docs/docs/api/plugin-bindings-api.md
+++ b/docs/docs/api/plugin-bindings-api.md
@@ -579,7 +579,7 @@ curl -s -X POST \
 2. `GatewayTenantPluginManagerFactory.get_config_from_db()` fetches all DB bindings for the `(team_id, tool_name)` pair, including any wildcard `*` bindings.
 3. For each binding, the DB `mode` and `config` are merged over the global `config.yaml` values (`_merge_tenant_config`). DB values always win.
 4. A **`TenantPluginManager`** is instantiated with the merged config and cached in memory, keyed by context ID.
-5. On upsert or delete, the cache entry is invalidated on the worker that handled the API request. Other workers and pods pick up the change when the cached manager expires (default TTL: 30 seconds). For wildcard bindings (`tool_name="*"`), all cached contexts for the team are evicted.
+5. On upsert or delete, the handling worker invalidates its local cache entry and broadcasts a `binding_change` frame on the `plugin:invalidation` Redis pub/sub channel, so peer workers evict within milliseconds. If pub/sub delivery fails, the cache TTL (default 30 seconds) bounds the worst-case drift. For wildcard bindings (`tool_name="*"`), every cached context for the team is evicted on the handling worker.
 
 ### Priority execution order
 
--- a/docs/docs/api/plugin-bindings-api.md
+++ b/docs/docs/api/plugin-bindings-api.md
@@ -579,7 +579,7 @@ curl -s -X POST \
 2. `GatewayTenantPluginManagerFactory.get_config_from_db()` fetches all DB bindings for the `(team_id, tool_name)` pair, including any wildcard `*` bindings.
 3. For each binding, the DB `mode` and `config` are merged over the global `config.yaml` values (`_merge_tenant_config`). DB values always win.
 4. A **`TenantPluginManager`** is instantiated with the merged config and cached in memory, keyed by context ID.
-5. On upsert or delete, the cache entry is invalidated immediately so the next call picks up the new config.
+5. On upsert or delete, the cache entry is invalidated on the worker that handled the API request. Other workers and pods pick up the change when the cached manager expires (default TTL: 30 seconds). For wildcard bindings (`tool_name="*"`), all cached contexts for the team are evicted.
 
 ### Priority execution order
 
--- a/mcpgateway/admin.py
+++ b/mcpgateway/admin.py
@@ -104,7 +104,11 @@ from mcpgateway.schemas import (
     PaginationMeta,
     PluginDetail,
     PluginListResponse,
+    PluginModeUpdateRequest,
+    PluginModeUpdateResponse,
     PluginStatsResponse,
+    PluginToggleRequest,
+    PluginToggleResponse,
     PromptCreate,
     PromptMetrics,
     PromptRead,
@@ -2028,11 +2032,10 @@ async def get_overview_partial(
         resources_total = db.query(func.count(DbResource.id)).scalar() or 0  # pylint: disable=not-callable
         resources_active = db.query(func.count(DbResource.id)).filter(DbResource.enabled.is_(True)).scalar() or 0  # pylint: disable=not-callable
 
-        # Plugin stats
+        # Plugin stats — self-heal the cache so the overview reflects the live
+        # shared toggle even on a process that booted with plugins disabled.
         overview_plugin_service = get_plugin_service()
-        plugin_manager = getattr(request.app.state, "plugin_manager", None)
-        if plugin_manager:
-            overview_plugin_service.set_plugin_manager(plugin_manager)
+        await _sync_plugin_service_from_runtime(request, overview_plugin_service)
         plugin_stats = await overview_plugin_service.get_plugin_statistics()
 
         # Infrastructure status (database, cache, uptime)
@@ -16535,6 +16538,41 @@ async def get_gateways_section(
 ####################
 
 
+async def _sync_plugin_service_from_runtime(request: Request, plugin_service) -> None:
+    """Self-heal the admin plugin cache from the live framework state.
+
+    The framework's ``get_plugin_manager`` is the single source of truth — it
+    reads the shared toggle (TTL-cached, so this is a cheap call) and returns
+    ``None`` when plugins are globally disabled, even when the disable came
+    from a *remote* node via the Redis toggle.
+
+    Every admin read mirrors that answer back into ``app.state.plugin_manager``
+    and the ``PluginService`` singleton. This closes three gaps:
+
+    1. Processes that booted with plugins disabled never had ``app.state`` set,
+       so admin views returned empty until restart even after the shared
+       toggle was flipped on.
+    2. If ``toggle_plugins_global`` swallowed an admin-cache sync failure, the
+       stale cache would persist forever — now the next GET repairs it.
+    3. A remote disable (``PUT /admin/plugins {"enabled": false}`` on another
+       worker) would leave this worker's ``app.state.plugin_manager``
+       populated from a prior enable, making admin views serve plugin
+       metadata the cluster had already turned off.
+
+    Best-effort: a failure logs a WARNING and leaves ``app.state`` alone. It
+    never raises, so it can't turn a read into a 500.
+    """
+    try:
+        # pylint: disable=import-outside-toplevel
+        from mcpgateway.plugins.framework import get_plugin_manager
+
+        plugin_manager = await get_plugin_manager()
+        request.app.state.plugin_manager = plugin_manager
+        plugin_service.set_plugin_manager(plugin_manager)
+    except Exception as sync_exc:
+        LOGGER.warning("Admin plugin-cache self-heal failed (%s) — view may render stale/empty", sync_exc)
+
+
 @admin_router.get("/plugins/partial")
 @require_permission("admin.plugins", allow_admin_bypass=False)
 async def get_plugins_partial(request: Request, db: Session = Depends(get_db), user=Depends(get_current_user_with_permissions)) -> HTMLResponse:  # pylint: disable=unused-argument
@@ -16558,17 +16596,15 @@ async def get_plugins_partial(request: Request, db: Session = Depends(get_db), u
         # Get plugin service and check if plugins are enabled
         plugin_service = get_plugin_service()
 
-        # Check if plugin manager is available in app state
-        plugin_manager = getattr(request.app.state, "plugin_manager", None)
-        if plugin_manager:
-            plugin_service.set_plugin_manager(plugin_manager)
+        # Self-heal the cache so the partial reflects the live shared toggle.
+        await _sync_plugin_service_from_runtime(request, plugin_service)
 
         # Get plugin data
         plugins = plugin_service.get_all_plugins()
         stats = await plugin_service.get_plugin_statistics()
 
         # Prepare context for template
-        context = {"request": request, "plugins": plugins, "stats": stats, "plugins_enabled": plugin_manager is not None, "root_path": _resolve_root_path(request)}
+        context = {"request": request, "plugins": plugins, "stats": stats, "plugins_enabled": plugin_service.get_plugin_manager() is not None, "root_path": _resolve_root_path(request)}
 
         # Render the partial template
         return request.app.state.templates.TemplateResponse(request, "plugins_partial.html", context)
@@ -16620,10 +16656,8 @@ async def list_plugins(
         # Get plugin service
         plugin_service = get_plugin_service()
 
-        # Check if plugin manager is available
-        plugin_manager = getattr(request.app.state, "plugin_manager", None)
-        if plugin_manager:
-            plugin_service.set_plugin_manager(plugin_manager)
+        # Self-heal the cache from the live framework state.
+        await _sync_plugin_service_from_runtime(request, plugin_service)
 
         # Get filtered plugins
         if any([search, mode, hook, tag]):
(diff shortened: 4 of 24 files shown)


araujof:
@jonpspri it looks like this one has to be backported to CPEX, right?
Tracking: https://github.com/contextforge-org/contextforge-plugins-framework/pull/11
Related: #3754

jonpspri:
@araujof Afraid so.  @cafalchio should handle coordination come Monday.  One could argue that this is all for the subclass, in which case we can factor it elsewhere -- I've lost track of how that's being handled.

PR #4326, opened 2026-04-20
Fixing missing dev dependencies

dawid-nowak:
A PR to add missing "install-dev" dependency to make serve.


brian-hussey (changes requested):
Just one change and we're good to go.

brian-hussey:
> Just one change and we're good to go.

Following discussion, that change I suggested was wrong.

I see an issue where `gunicorn` isn't installed from `make serve` which relies on `make venv`.

We need `make venv` to resolve all production dependencies, including gunicorn, for `make serve` to work.

brian-hussey (approved):
LGTM

brian-hussey (approved):
LGTM

PR #4345, opened 2026-04-20
fix: Move JWT from the session storage to cookie

### LINKED ISSUE #4269: [BUG]: Sessions not persisting across different tabs
If a user logs in the new client on a tab, opens a new tab and navigates to the new client, the login page is displayed.

gcgoncalves:
## 🔗 Related Issue
Closes #4269

---

## 📝 Summary
Move the JWT token from the sessionStorage to the cookies. To be able to do that, add an optional parameter to the login api endpoint to set the cookie, and creates a logout endpoint (that deletes the cookie).

---

## 🏷️ Type of Change
- [ ] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [X] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

| Lint suite                | `make lint`     |    ✅    |
| Unit tests                | `make test`     |    ✅    |
| Coverage ≥ 80%            | `make coverage` |    ✅    |

---

## ✅ Checklist
- [X] Tests added/updated for changes
- [X] Documentation updated (if applicable)


marekdano (changes requested):
### Summary

  This PR adds a React SPA scaffold under client/ with cookie-based authentication, mounts it at `/app/*`, and introduces `security_cookies.py` with an httpOnly cookie helper. The intent is sound, but there is a critical gap between the client and server: the server never reads the `jwt_token` cookie for authentication, so all protected API calls from the React SPA will return 401.

  ---
### Findings


- `client/src/auth/AuthContext.tsx:50` — Correctness
  The client calls `GET /auth/me` but no such endpoint exists. The only `/me` endpoint is `GET /auth/email/me`, mounted via `email_auth_router` with prefix `/auth/email`. On startup, every user will appear unauthenticated regardless of their session state. 
Why do we need to call `/auth/me` endpoint?

- `mcpgateway/auth.py + client/src/api/client.ts` — Correctness
  The server's `get_current_user` reads only the `Authorization: Bearer` header via `HTTPBearer`. The client sets `credentials: "include"` to send the `jwt_token` cookie but never adds a `Bearer` header. Every request to a protected endpoint will return 401. Either (a) extend `get_current_user` to fall back to `request.cookies.get("jwt_token")` when no `Bearer` header is present, or (b) have the client store the token in memory/sessionStorage and inject `Authorization: Bearer <token>` — the unauthenticated option in `RequestOptions` already hints at this design but is never implemented.

- `client/src/api/client.ts:36-37` — Correctness
  `unauthenticated?: boolean` is declared in `RequestOptions` and referenced in the JSDoc, but the `request()` function body never reads it and never adds an Authorization header. The module comment ("Authorization: Bearer header is added on every authenticated request", "JWT is read from sessionStorage") is entirely aspirational — neither mechanism is implemented. Remove the dead option and the false comment, or implement it.

- `mcpgateway/routers/auth.py:206-208` — Security
  POST `/auth/logout` requires no authentication. While it only clears a browser cookie, it also does no server-side token revocation — the JWT remains valid until natural expiry. A token that was captured before logout retains full access.
  Consider adding server-side revocation on logout, or at minimum, add a comment documenting the intentional lack of revocation and the expiry window. Also consider adding Depends(`get_current_user`) to avoid leaking information about expired sessions.

- `mcpgateway/middleware/security_headers.py:283` — Design
  `is_spa_path` re-evaluates `request.url.path` on every response with a string prefix check inside middleware that runs on every request. Consider moving the SPA CSP into a response header added by `app_spa_router` itself (via a dependency or route-level middleware) so the logic is co-located with the routes it applies to, rather than branching inside the global security headers middleware.

-  `mcpgateway/routers/auth.py:185` — Consistency
 `logger.debug(f"Set httpOnly cookie for user: {email}")` uses an f-string. The project convention is %-style: `logger.debug("Set httpOnly cookie for user: %s", email)`.

- `client/src/pages/Login.tsx` — Design
  The login form has no "remember me" checkbox, yet set_auth_cookie accepts a remember_me parameter that extends the session to 30 days. The only way to get a 30-day session is to call the API directly. Either add a "remember me" checkbox and wire it through `AuthContext.login()` and the request body, or remove the `remember_me` parameter if the design decision is always 1-hour sessions.

- `client/src/auth/AuthContext.tsx:54-58` — Design
  Non-401 errors from `GET /auth/me` (e.g., 500, 503) are silently swallowed. A transient backend failure at startup will land the user on the login page with no explanation. Re-throw non-401 errors or expose an authError state so the UI can surface a meaningful message.

gcgoncalves:
@marekdano Can you re-review, please?

marekdano (changes requested):
Concerns and Issues

  1. No CSRF protection for cookie-authenticated requests — Blocking

  Switching from Bearer headers to cookies fundamentally changes the CSRF threat model. Bearer tokens in Authorization headers cannot be sent cross-origin by an attacker's page (browsers block cross-origin header injection). With credentials: "include" and cookie auth, a `<form>` or `fetch()` on an attacker-controlled origin can trigger authenticated state-changing requests, provided SameSite=lax does not block it.

  `SameSite=lax` protects against POST cross-site forgery but does not protect GET endpoints that have side effects, and does not protect against same-site attacks. The codebase has no CSRF token middleware. This needs to be explicitly acknowledged and addressed, or a double-submit cookie / `X-Requested-With` header check added. The client already sends `X-Requested-With: XMLHttpRequest` on every request — the server should validate this header on cookie-authenticated endpoints as a lightweight CSRF guard.

  2. No token revocation on logout — Notable

  `POST /auth/logout` clears the browser cookie, but the JWT itself remains valid until expiry (default 1 hour). This is documented and acknowledged as an intentional trade-off. However:
  - If a session cookie is stolen (e.g., via network sniffing before Secure is enforced, or on a compromised machine), logout does not invalidate the token server-side.
  - The TokenRevocation table exists in the codebase — it would be worth filing a follow-up issue to implement JTI revocation on logout rather than leaving it as open guidance in a comment.

  3. Token still returned in response body when `set_cookie: true`

  The login response includes access_token in the JSON body regardless of `set_cookie`. For browser SPA flows with `set_cookie: true`, returning the token in the body is unnecessary and slightly increases the exposure window (the SPA ignores the field, but other code or interceptors could log it). Consider returning an empty string or omitting access_token in the response when `set_cookie: true`.

  4. `SameSite=lax` default — subdomain deployment needs validation

  `SameSite=lax` is a reasonable default, but the PR does not address the case where the gateway is deployed on a subdomain and the admin UI is on a different subdomain. In that topology, lax may behave differently than expected, and `SameSite=strict` or explicit domain scoping would be needed. The absence of a domain attribute means the cookie is scoped to the exact host, which is fine for most deployments but worth documenting.

  5. Cookie fallback import inside function — Minor

  In mcpgateway/auth.py, the `import from fastapi.security import HTTPAuthorizationCredentials as Creds` is placed inside the if cookie_token: branch. However, `HTTPAuthorizationCredentials` is already imported at the top of the file (line 24). The alias is unnecessary — replace `Creds(...)` with `HTTPAuthorizationCredentials(...)` and remove the inline import entirely.

  ---
  Code Quality and Correctness

  6. Inconsistent logging style in `routers/auth.py` — Minor

  The second commit claims to standardize to %-style logging, but the actual code still uses f-strings in the login handler:
```bash
  logger.info(f"User {email} authenticated successfully")
  logger.error(f"Login error for {login_request.email or login_request.username}: {e}")
  logger.warning(f"Login validation error: {e}")
```

  Only the logger.debug line uses %-style. The remaining lines should be updated to %-style (e.g. `logger.info("User %s authenticated successfully", email)`) to be consistent with the project standard and avoid unnecessary string interpolation when the log level is suppressed.

  7. `get_current_user_info` injects db but never uses it — Correctness

  In `routers/auth.py`:
```bash
  @auth_router.get("/me", response_model=EmailUserResponse)
  async def get_current_user_info(request: Request, db: Session = Depends(get_db)):
      from mcpgateway.auth import get_current_user
      user = await get_current_user(request=request)
      return EmailUserResponse.from_email_user(user)
```

  The db dependency is injected but never passed to `get_current_user()`. The injected db is dead code that wastes a DB connection on every `/auth/me` call. Either remove `db: Session = Depends(get_db)` or thread it through to `get_current_user`.

  8. `AuthContext.tsx` missing loading state — Blocking UX issue
```bash
  useEffect(() => {
    api.get<User>("/auth/me")
      .then((user) => { setState({ user, isAuthenticated: true }); })
      .catch((err) => { ... });
  }, []);
```

  The initial state is `{ user: null, isAuthenticated: false }`. Between mount and the resolution of `/auth/me`, `AuthGuard` will see `isAuthenticated: false` and redirect authenticated users to `/app/login`, causing a visible flash/redirect on every hard refresh. A loading: true state should be added to defer `AuthGuard` decisions until the check completes.

  9. `logout()` silently swallows all errors
```bash
  try {
    await api.post("/auth/logout");
  } catch {
    // Ignore errors - proceed with logout
  }
```

  Proceeding with client-side logout regardless of server errors is correct from a UX perspective, but silently discarding non-network errors makes debugging harder. At minimum, add a `console.warn` with the caught error.

  10. `remember_me=False` is hardcoded — dead code in `security_cookies.py`

  `LoginRequest` does not expose a `remember_me` field, so the 30-day expiry path in security_cookies.py is unreachable from the UI or API. Either expose the field in `LoginRequest` and add a "Remember me" checkbox to the login form, or remove the `remember_me` parameter from `set_auth_cookie` until it is needed.

  11. Docstring precedence order is wrong in `get_current_user() — mcpgateway/auth.py`

  The docstring lists Bearer → Cookie → Plugin, but the code comment at line 1164 literally states "# This hook is invoked BEFORE standard JWT/API token validation". The actual execution order is Plugin → Bearer → Cookie. The docstring has it backwards, which will mislead future maintainers.

  ---
  Edge Cases and Potential Issues

  12. Cookie-authenticated users hitting non-`/auth` endpoints

  The cookie fallback is in `get_current_user()` in `mcpgateway/auth.py`, which is used as a FastAPI dependency. This should work for all endpoints that use Depends(get_current_user). However, endpoints that use different auth dependencies (e.g., `get_current_active_user`, basic auth decorators) will not benefit from the cookie fallback and will return 401 to browser clients. The PR should confirm which auth dependency paths the SPA actually hits.

  13. `CookieTooLargeError` unhandled in login router

  `set_auth_cookie() raises CookieTooLargeError` if the JWT exceeds 4096 bytes (possible with many team claims). In routers/auth.py, the login endpoint calls `set_auth_cookie()` without catching this exception — FastAPI will return a raw 500. `CookieTooLargeError` should be caught and translated to a structured HTTP response:
```bash
  if login_request.set_cookie:
      try:
          set_auth_cookie(response, access_token, remember_me=False)
      except CookieTooLargeError:
          raise HTTPException(
              status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
              detail="Authentication token is too large to store as a cookie. Use Authorization header instead.",
          )
```

  14. `app_root_path` cookie path edge case

  `set_auth_cookie` uses `path=settings.app_root_path` or "/". If app_root_path is set to something other than what the SPA is served under, the cookie won't be sent for SPA requests. This is configuration-dependent and should be documented.

  15. `/auth/me` on every SPA mount adds latency

  Every full-page load or hard refresh fires a GET to `/auth/me`. This is standard practice for cookie-auth SPAs, but in multi-tab scenarios each tab pings /auth/me independently. Worth noting for teams running under tight latency budgets or with slow DB paths.

gcgoncalves:
This PR is getting too big, I'll split into backend and frontend.

Issue #4110, opened 2026-04-09
[INTERNAL]: Rust MCP Runtime - Code Quality Improvements from Mend Scan


MohanLaksh:
## Summary

Mend SAST scan (April 8, 2026) identified several code quality improvements for the Rust MCP runtime (`tools_rust/mcp_runtime/src/lib.rs`).

**Scan Details:**
- **Date:** 2026-04-08 06:52am
- **Total Rust Findings:** 25
- **File:** `tools_rust/mcp_runtime/src/lib.rs` (13,319 lines)

---

### 🔵 URL Validation (19 findings)

**Recommendation:** Add validation layer for backend HTTP requests

**Affected Functions:**
- `send_to_backend_url()` (line 4422)
- `send_sampling_to_backend()` (line 7891)
- `send_prompts_get_to_backend()` (line 7978)
- 16+ additional backend dispatch functions

**Root Cause:** Backend URL construction from configuration without validation

**Recommended Implementation:**
1. Add URL validation module with hostname resolution
2. Implement network range checking
3. Add configuration for allowed destinations
4. Apply validation to all backend HTTP calls

### 🔵 Query Parameter Binding (2 findings)

**Status:** FALSE POSITIVE (verified)

**Analysis:** All database queries use parameterized statements with proper parameter binding. No changes needed.

**Example (line 4841):**
```rust
client.query(
    "SELECT t.name FROM tools t WHERE sta.server_id = $1",
    &[&server_id],  // ✅ Parameterized - SAFE
)
```

### 🔵 Request Size Limits (1 finding)

**Recommendation:** Add request body size constraints

**Current State:**
- No size limits on incoming payloads
- No nesting depth limits for JSON parsing

**Recommended Mitigation:**
1. Limit payload size (e.g., 10MB max)
2. Set max JSON nesting depth
3. Use size-limited readers for parsing
4. Add timeout on parsing operations

---

### Phase 1: URL Validation

**Create validation module:** `tools_rust/mcp_runtime/src/url_validator.rs`

**Features:**
- Length validation (max 2048 chars)
- Scheme allowlist (http, https, ws, wss)
- Pattern validation
- Hostname resolution with network checking
- Configurable allowed/blocked ranges

**Dependencies:**
```toml
[dependencies]
url = "2.5"              # URL parsing
ipnetwork = "0.20"       # CIDR validation
trust-dns-resolver = "0.24"  # DNS resolution
regex = "1.10"           # Pattern matching
```

[...]

PR #4362, opened 2026-04-21
feat(runtime): enhance request validation and size limiting

MohanLaksh:
## Summary

Adds comprehensive request validation and resource management controls to the Rust MCP runtime to improve robustness and operational reliability.

## Review Feedback Addressed

All review findings from PR #4362 have been addressed:

1. ✅ **Missing URL validation in backend functions** - Added validation to `send_transport_to_backend()` and `send_session_delete_to_backend()` 
2. ✅ **DNS_FAIL_CLOSED logic bug** - Fixed to properly handle DNS resolver errors with fail-closed policy
3. ✅ **UrlValidator env var coupling** - Refactored to accept `RuntimeConfig` parameter instead of reading env vars directly
4. ✅ **Invalid cargo test command** - Fixed documentation to use `cargo test --all-features`
5. ✅ **Author attribution** - Standardized to "Authors: Mihai Criveti" across all files
6. ✅ **Supply chain audit** - Added missing notes field to windows-registry trust entry

### Critical Bug Fix

Fixed localhost validation false positive where the regex pattern `on\w+\s*=` was incorrectly matching query parameters like `session_id=`. Updated to `\bon\w+\s*=` to only match HTML event handlers (onclick=, onerror=) as intended.

### URL Validator Module (`url_validator.rs`)
- Hostname resolution with DNS caching (5-minute TTL)
- Network filtering with CIDR support
- Configurable validation rules via RuntimeConfig
- Localhost allowed by default for development convenience
- Mock DNS resolver trait for deterministic testing
- Fail-closed DNS resolution with configurable policy

### Backend Request Validation
- Applied validation to all backend HTTP requests (19 call sites including transport/session delete)
- Centralized validation logic via `AppState::validate_backend_url()`
- Fail-closed security pattern for invalid configurations

### Request Size Controls
- Configurable body size limits (default 10MB)
- Applied via Axum `DefaultBodyLimit` layer
- Returns 413 response for oversized payloads

## Configuration

New environment variables:

**URL Validation:**
- `VALIDATION_ENABLED` - Enable/disable validation (default: true)
- `MAX_URL_LENGTH` - Maximum URL length (default: 2048)
- `ALLOW_LOCALHOST` - Allow localhost access (default: true)
- `ALLOW_PRIVATE_NETWORKS` - Allow RFC1918 networks (default: false)
- `BLOCKED_NETWORKS` - CIDR ranges to filter (comma-separated)
- `BLOCKED_HOSTS` - Hostname blocklist (comma-separated)
- `ALLOWED_NETWORKS` - Allowlist specific CIDR ranges (comma-separated)
- `DNS_FAIL_CLOSED` - Fail closed on DNS errors (default: true)

**Request Size:**
- `MCP_RUST_MAX_REQUEST_BODY_SIZE_BYTES` - Body size limit (default: 10MB)

## Testing

**Test Coverage:**
- ✅ 99 unit tests (including URL validator and body size tests)
- ✅ 57 integration tests
- ✅ 2 CLI tests
- ✅ Total: 158 tests passing with `--all-features`

**Quality Checks:**
- ✅ Formatting clean (`cargo fmt --check`)
- ✅ Linting clean (`cargo clippy -D warnings`)
- ✅ Supply chain verification (`cargo vet`)
- ✅ Release build successful
- ✅ All pre-commit hooks passed

**Run tests:**
```bash
cd crates/mcp_runtime
cargo test --all-features
```

## Dependencies Added

- `async-trait 0.1` - DNS resolver abstraction
- `hickory-resolver 0.25` - DNS resolution with caching
- `ipnetwork 0.21` - CIDR network filtering
- `serial_test 3.4` - Test isolation (dev dependency)
- `tower 0.5` features - Test utilities (dev dependency)

All new dependencies have cargo-vet exemptions.

## Documentation

- **README.md** - Added Configuration section with examples
- **DEVELOPING.md** - Fixed invalid test command, added testing guidance

## Key Features

1. **DNS Caching** - 5-minute TTL reduces repeated DNS lookups for performance
2. **Fail-Closed Security** - Invalid CIDR configuration prevents startup rather than silently ignoring
3. **Complete Coverage** - All backend HTTP call sites protected with validation
4. **Development Friendly** - Localhost allowed by default to match backend_rpc_url default (127.0.0.1)
5. **Comprehensive Testing** - Mock DNS resolver enables deterministic validation tests
6. **RuntimeConfig Integration** - Centralized configuration instead of scattered env var reads

## Backwards Compatibility

✅ No breaking changes - all validation uses secure defaults

## Related

Addresses: Issue #4110

---

**Ready for re-review. All feedback addressed.**

### DIFF
--- a/crates/mcp_runtime/src/config.rs
+++ b/crates/mcp_runtime/src/config.rs
@@ -10,6 +10,9 @@ use std::{net::SocketAddr, path::PathBuf};
 const DEFAULT_SUPPORTED_PROTOCOL_VERSIONS: &[&str] =
     &["2025-11-25", "2025-06-18", "2025-03-26", "2024-11-05"];
 
+/// Default maximum request body size (10MB)
+pub const DEFAULT_MAX_REQUEST_BODY_SIZE_BYTES: usize = 10_485_760;
+
 #[derive(Debug, Clone, Parser)]
 #[command(name = "contextforge-mcp-runtime")]
 #[command(about = "Experimental Rust MCP runtime edge for ContextForge")]
@@ -185,7 +188,7 @@ pub struct RuntimeConfig {
     #[arg(
         long,
         env = "MCP_RUST_MAX_REQUEST_BODY_SIZE_BYTES",
-        default_value_t = 10_485_760
+        default_value_t = DEFAULT_MAX_REQUEST_BODY_SIZE_BYTES
     )]
     pub max_request_body_size_bytes: usize,
 
--- a/crates/mcp_runtime/src/config.rs
+++ b/crates/mcp_runtime/src/config.rs
@@ -192,6 +192,45 @@ pub struct RuntimeConfig {
     )]
     pub max_request_body_size_bytes: usize,
 
+    #[arg(long, env = "VALIDATION_ENABLED", default_value_t = true)]
+    pub validation_enabled: bool,
+
+    #[arg(long, env = "MAX_URL_LENGTH", default_value_t = 2048)]
+    pub max_url_length: usize,
+
+    #[arg(long, env = "ALLOW_LOCALHOST", default_value_t = true)]
+    pub allow_localhost: bool,
+
+    #[arg(long, env = "ALLOW_PRIVATE_NETWORKS", default_value_t = false)]
+    pub allow_private_networks: bool,
+
+    #[arg(
+        long,
+        env = "BLOCKED_NETWORKS",
+        default_value = "169.254.169.254/32",
+        value_delimiter = ','
+    )]
+    pub blocked_networks: Vec<String>,
+
+    #[arg(
+        long,
+        env = "BLOCKED_HOSTS",
+        default_value = "metadata.google.internal,metadata.goog",
+        value_delimiter = ','
+    )]
+    pub blocked_hosts: Vec<String>,
+
+    #[arg(
+        long,
+        env = "ALLOWED_NETWORKS",
+        default_value = "",
+        value_delimiter = ','
+    )]
+    pub allowed_networks: Vec<String>,
+
+    #[arg(long, env = "DNS_FAIL_CLOSED", default_value_t = true)]
+    pub dns_fail_closed: bool,
+
     #[arg(long, env = "MCP_RUST_EXIT_AFTER_STARTUP_MS", hide = true)]
     pub exit_after_startup_ms: Option<u64>,
 }
--- a/crates/mcp_runtime/src/config.rs
+++ b/crates/mcp_runtime/src/config.rs
@@ -182,6 +182,13 @@ pub struct RuntimeConfig {
     #[arg(long, env = "MCP_RUST_LOG", default_value = "info")]
     pub log_filter: String,
 
+    #[arg(
+        long,
+        env = "MCP_RUST_MAX_REQUEST_BODY_SIZE_BYTES",
+        default_value_t = 10_485_760
+    )]
+    pub max_request_body_size_bytes: usize,
+
     #[arg(long, env = "MCP_RUST_EXIT_AFTER_STARTUP_MS", hide = true)]
     pub exit_after_startup_ms: Option<u64>,
 }
--- a/crates/mcp_runtime/src/url_validator.rs
+++ b/crates/mcp_runtime/src/url_validator.rs
@@ -1,0 +1,660 @@
+// Copyright 2026
+// SPDX-License-Identifier: Apache-2.0
+// Authors: Mihai Criveti, ContextForge Team
+
+//! URL validation with hostname resolution and network filtering for the Rust MCP runtime.
+//!
+//! This module implements URL validation to ensure backend requests reach only approved
+//! network destinations. It matches the validation guarantees of the Python implementation.
+//!
+//! # Features
+//!
+//! - Length validation (configurable max length)
+//! - Scheme allowlist (http, https, ws, wss only)
+//! - Pattern blocking (javascript:, data:, file:, etc.)
+//! - IPv6 blocking
+//! - CRLF injection prevention
+//! - Credentials in URL blocking
+//! - Pattern detection for malformed content
+//! - **Network Filtering**:
+//!   - Hostname normalization
+//!   - Blocked hostname list check
+//!   - DNS resolution of all A/AAAA records
+//!   - CIDR-based IP filtering
+//!   - Localhost filtering (configurable, allowed by default)
+//!   - Private network filtering (configurable)
+//!   - Allowlist support for specific private ranges
+//!
+//! # Configuration
+//!
+//! All filtering settings are loaded from environment variables via `RuntimeConfig`:
+//!
+//! - `VALIDATION_ENABLED=true` - Enable validation (default: true)
+//! - `BLOCKED_NETWORKS` - CIDR ranges to block (comma-separated)
+//! - `BLOCKED_HOSTS` - Hostnames to block (comma-separated)
+//! - `ALLOW_LOCALHOST=true` - Allow localhost access (default: true)
+//! - `ALLOW_PRIVATE_NETWORKS=false` - Allow RFC1918 networks (default: false)
+//! - `ALLOWED_NETWORKS` - Allowlist specific CIDR ranges (comma-separated)
+//! - `DNS_FAIL_CLOSED=true` - Fail closed on DNS errors (default: true)
+//! - `MAX_URL_LENGTH=2048` - Maximum URL length (default: 2048)
+
+use crate::config::RuntimeConfig;
+use async_trait::async_trait;
+use hickory_resolver::TokioResolver;
+use ipnetwork::IpNetwork;
+use regex::Regex;
+use std::collections::HashMap;
+use std::net::IpAddr;
+use std::sync::Arc;
+use std::time::{Duration, Instant};
+use thiserror::Error;
+use tokio::sync::RwLock;
+use tracing::debug;
+use url::Url;
+
+/// Type alias for DNS cache entry: (resolved IPs, expiry time)
+type DnsCacheEntry = (Vec<IpAddr>, Instant);
+
+/// Type alias for DNS cache: hostname -> cache entry
+type DnsCache = HashMap<String, DnsCacheEntry>;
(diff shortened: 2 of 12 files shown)


MohanLaksh:
@lucarlig , @dima-zakharov , @dawid-nowak , @crivetimihai 

Can you please help me in reviewing and getting this PR merged? We need it to be merged as part of the Mend SAST scan fixes.

crates/mcp_runtime/tests/runtime.rs
```diff
@@ -129,6 +129,7 @@ fn test_runtime_config() -> RuntimeConfig {
         redis_url: None,
         db_pool_max_size: 20,
         log_filter: "error".to_string(),
+        max_request_body_size_bytes: 10_485_760,
```
dawid-nowak: hardcoded value, should be const variable

lucarlig:
A few concrete issues stood out in this PR:

| Finding | Impact |
| --- | --- |
| `send_transport_to_backend()` and `send_session_delete_to_backend()` still call `reqwest` directly without the new validator. | When Rust MCP is active, the main `/mcp` transport and session teardown paths can bypass the new SSRF guard, so protection is only partial. |
| `DNS_FAIL_CLOSED=false` does not apply to real resolver failures because `resolve_hostname_cached()` returns `DnsError` before the fail-open branch runs. | The documented permissive mode only works for empty DNS results, not NXDOMAIN/resolver errors, so runtime behavior does not match the config/docs. |
| `UrlValidator::from_config()` ignores `RuntimeConfig` and re-reads ad hoc env vars directly. | This creates a second hidden security config plane, makes tests less hermetic, and makes effective validator policy harder to reason about. |
| `cargo test url_validator request_body_limit --all-features` is not a valid Cargo command. | The documented verification step fails instead of running both test groups. |

Also the auhors in the comments are wrong

dawid-nowak (changes requested):
My suggestions for going forward:
1. Split into two independent PRs. One for URL validation and one for body size limiting.
2. Explain when URL validator is supposed to be used ? (Incoming request/outgoing request/something else ?)
3. There is a lot copy-paste in the code around handling URL validation and sending errors. Please re-factor this code.

Additionally: 
* I am not sure why we need URL validation. My understanding is that this is something that has been configured via our control plane so it should be validated. The HTTP client/reqwest will resolve the url to a hostname and if it can't resolve than we can't make a request anyway.  Also reqwest is using hickory DNS internally so I don't see why we would do the same job twice.

crates/mcp_runtime/src/config.rs
```diff
@@ -182,6 +182,13 @@ pub struct RuntimeConfig {
     #[arg(long, env = "MCP_RUST_LOG", default_value = "info")]
     pub log_filter: String,
 
+    #[arg(
+        long,
+        env = "MCP_RUST_MAX_REQUEST_BODY_SIZE_BYTES",
+        default_value_t = 10_485_760
```
dawid-nowak: hardcoded value which, should be const variable shared with runtime.rs

supply-chain/audits.toml
```diff
@@ -1330,6 +1330,12 @@ start = "2023-05-15"
 end = "2027-04-13"
 notes = "Trusted upstream maintainer for windows crates"
 
+[[trusted.windows-registry]]
+criteria = "safe-to-deploy"
```
dawid-nowak, line 1334: why was this changed ??

MohanLaksh: It was added by cargo vet when we ran dependency auditing after adding hickory-resolver = "0.25.0" to Cargo.toml. Adding notes here...

crates/mcp_runtime/tests/runtime.rs
MohanLaksh: Changed to const variable

crates/mcp_runtime/src/config.rs
MohanLaksh: Agree and done

MohanLaksh:
## Review Feedback Addressed

All review findings from PR #4362 have been addressed:

1. ✅ **Missing URL validation in backend functions** - Added validation to `send_transport_to_backend()` and `send_session_delete_to_backend()` 
2. ✅ **DNS_FAIL_CLOSED logic bug** - Fixed to properly handle DNS resolver errors with fail-closed policy
3. ✅ **UrlValidator env var coupling** - Refactored to accept `RuntimeConfig` parameter instead of reading env vars directly
4. ✅ **Invalid cargo test command** - Fixed documentation to use `cargo test --all-features`
5. ✅ **Author attribution** - Standardized to "Authors: Mihai Criveti" across all files
6. ✅ **Supply chain audit** - Added missing notes field to windows-registry trust entry

### Critical Bug Fix

Fixed localhost validation false positive where the regex pattern `on\w+\s*=` was incorrectly matching query parameters like `session_id=`. Updated to `\bon\w+\s*=` to only match HTML event handlers (onclick=, onerror=) as intended.

MohanLaksh:
> My suggestions for going forward:
> 
> 1. Split into two independent PRs. One for URL validation and one for body size limiting.
> 2. Explain when URL validator is supposed to be used ? (Incoming request/outgoing request/something else ?)
> 3. There is a lot copy-paste in the code around handling URL validation and sending errors. Please re-factor this code.
> 
> Additionally:
> 
> * I am not sure why we need URL validation. My understanding is that this is something that has been configured via our control plane so it should be validated. The HTTP client/reqwest will resolve the url to a hostname and if it can't resolve than we can't make a request anyway.  Also reqwest is using hickory DNS internally so I don't see why we would do the same job twice.

  Thanks for the detailed review! Let me address each point:
                                                                                                                                                                          
  Re: URL validator purpose and threat model                                                                                                                              
                                                                                                                                                                          
  The URL validator protects against SSRF (Server-Side Request Forgery) via misconfigured environment variables.                                                          
                                                                                                                                                                          
  Threat model: The Rust runtime makes outgoing HTTP requests to backend services using URLs from environment variables (BACKEND_RPC_URL, etc.). If these are             
  misconfigured (accidentally or maliciously), the runtime could make requests to:
  - Cloud metadata endpoints (169.254.169.254)                                                                                                                            
  - Internal network services (RFC1918 ranges if enabled)                                                                                                                 
  - Other blocked hosts/networks                         
                                                                                                                                                                          
  Why not rely on reqwest/hickory-dns? reqwest will happily connect to these targets - it's a client library, not a security boundary. The fact that reqwest uses         
  hickory-dns internally is for connectivity, not policy enforcement. Our validation performs pre-flight security checks before reqwest attempts connection:              
                                                                                                                                                                          
  1. Resolve hostname to IPs                                                                                                                                              
  2. Check resolved IPs against blocked networks/hosts
  3. Reject request if it violates policy                                                                                                                                 
  4. Only then allow reqwest to proceed                                                                                                                                   
                                                                                                                                                                          
  Scope: This validates outgoing backend requests from Rust → Python backend, not incoming client requests. The README.md documents this: "The runtime includes URL       
  validation for outgoing backend HTTP requests to protect against SSRF."                                                                                                 
                                                                                                                                                                          
  Re: Splitting PRs                                                                                                                                                       
   
  At this stage, I'd prefer to keep them together because:                                                                                                                
  - Both are input validation features addressing the same issue (#4110)
  - They're tightly coupled in implementation (both use RuntimeConfig, tested together)                                                                                   
  - All review feedback has been addressed with 158 passing tests                      
  - Splitting now would create rebase churn and delay both features                                                                                                       
                                                                                                                                                                          
  If you feel strongly about splitting, I can do it, but I'd advocate for keeping them unified as "request validation hardening."                                         
                                                                                                                                                                          
  Re: Refactoring validation error handling                                                                                                                               
                                                                                                                                                                          
  Good catch. I did extract validate_backend_url() on AppState to centralize the validation logic, but the error responses at call sites still have some duplication. I   
  can refactor this further to extract a helper like:
                                                                                                                                                                          
  fn validation_error_response(message: String) -> Response {
      json_response(                                                                                                                                                      
          StatusCode::BAD_GATEWAY,                                                                                                                                        
          json!({"error": "Bad Gateway", "message": message, "data": CLIENT_ERROR_DETAIL})                                                                                
      )                                                                                                                                                                   
  }                          
                                                                                                                                                                          
  Would you like me to add this refactoring, or is the current validate_backend_url() extraction sufficient?

crates/mcp_runtime/src/url_validator.rs
```diff
@@ -0,0 +1,689 @@
+// Copyright 2026
+// SPDX-License-Identifier: Apache-2.0
+// Authors: Mihai Criveti
```
lucarlig, line 3: please fix the authors

lucarlig (changes requested):

dawid-nowak:
1. The PRs should be split because they are completely separate in nature and it could cause confusion later.
2. I am not entirely convinced about your approach to SSRF (Server-Side Request Forgery)  protection. 
Even if we need to protect the url all we need is a list of "approved" urls, whether they are resolvable or not via hickory DNS it doesn't really matter. For IP addresses, you don't need a DNS anyway. If i get it right, you want to call DNS and then additionally validate the IPs returned by the DNS. So really, you want to protect against some sort of DNS poisoning which again is not really related to SSRF. This would imply that the request was correct, but somehow the DNS returned an "malicious" IP address.

MohanLaksh:
PR #1 (Body Size Limit):                                                                                                                                                
  - Number: #4382                                                                                                                                                         
  - Title: feat(runtime): add request body size limiting                                                                                                                  
  - Status: OPEN                                                                                                                                                          
  - URL: https://github.com/IBM/mcp-context-forge/pull/4382                                                                                                               
                                                                                                                                                                          
PR #2 (Backend URL Validation):                                                                                                                                         
  - Number: #4383            
  - Title: security(mcp_runtime): backend URL validation for SSRF protection                                                                                              
  - Status: OPEN                                                            
  - URL: https://github.com/IBM/mcp-context-forge/pull/4383                                                                                                               
                                                                                                                                                                          
Both PRs are now open and ready for review. They were split from the original PR #4362 to provide focused, easier-to-review changes.

jonpspri:
Closing in favor of #4382 and #4383

PR #4404, opened 2026-04-23
fix(oauth): auto-learn IdP audience and persist as resource for token validation

madhav165:
# 🐛 Bug-fix PR

## 📌 Summary

OAuth token audience validation fails for IdPs (ServiceNow, Authentik, etc.) that do not honor RFC 8707 and set the `aud` claim to an abstract identifier (e.g. `client_id`) rather than the resource URL sent in the authorization request. Users see "Token audience mismatch" errors after a successful OAuth flow, with no way to fix it from the UI.

RFC 8707 Section 2 explicitly allows this behavior:

> The authorization server may use the exact `resource` value as the audience **or it may map from that value to a more general URI or abstract identifier** for the given resource.

## 🔁 Reproduction Steps

Closes #4384
Related: #4171

1. Register a gateway with OAuth (e.g. ServiceNow StreamableHTTP)
2. Complete OAuth authorization flow — succeeds
3. Click "Fetch Tools" — fails with: `Token audience mismatch: token aud=[<client_id>], expected '<gateway_url>'`

## 🐞 Root Cause

Two disconnected code paths:

1. `/oauth/authorize` and `/oauth/callback` (`oauth_router.py`) inject `resource = _normalize_resource_url(gateway.url)` into a **transient copy** of `oauth_config` and send it to the IdP. This copy is never persisted.

2. `_validate_audience` (`token_validation_service.py:155`) reads `oauth_config.get("resource")` from the **persisted** config (which has no `resource`), falls back to `gateway_url`, and fails because the IdP's `aud` doesn't match the gateway URL.

The system sends a `resource` to the IdP but never learns what the IdP actually mapped it to.

## 💡 Fix Description

**Auto-learn the audience from the token at callback time and persist it as `resource`.**

1. **`oauth_manager.py`**: Added `_extract_token_audience()` — decodes the access token (best-effort, no signature verification) and extracts the `aud` claim. Returns only the claim value, never the raw token. `complete_authorization_code_flow` now returns `token_aud` in its result dict.

2. **`oauth_router.py`**: Added `_persist_learned_audience()` — after `complete_authorization_code_flow` succeeds in the callback, persists the `token_aud` as `resource` in `gateway.oauth_config`. Skips gracefully for opaque tokens, missing aud, or when resource already matches.

[...]

PR #4410, opened 2026-04-23
fix(oauth): skip audience enforcement for virtual servers when resource is not configured

### LINKED ISSUE #4171: [BUG]: OAuth audience verification always fails with IdPs that do not support RFC 8707 Resource Indicators (e.g. Authentik)
### 🐞 Bug Summary

OAuth access token audience verification fails with `Audience doesn't match` when using an IdP that does not support RFC 8707 Resource Indicators (e.g. Authentik). The hardened audience enforcement introduced in PR #3715 always derives the expected audience from `APP_DOMAIN`, but IdPs without RFC 8707 support set the token `aud` claim to the OAuth `client_id` instead of the resource URL. This makes it impossible to pass audience verification when `APP_DOMAIN` differs from the public-facing domain (e.g. reverse proxy or split internal/external endpoints), even though the token is otherwise valid.

---

### 🧩 Affected Component

- [x] `mcpgateway` - API
- [ ] Other (explain below)

---

### 🔁 Steps to Reproduce

1. Configure a Virtual Server with `oauth_enabled=true` and an authorization server pointing to an IdP that does **not** support RFC 8707 Resource Indicators (e.g. Authentik, or any OIDC provider where `resource_indicators_supported` is absent from `/.well-known/openid-configuration`).

2. Set `APP_DOMAIN` to an internal domain (e.g. `https://internal-gateway.example.com`) while exposing the MCP endpoint on a different public domain (e.g. `https://mcp.example.com`) via reverse proxy, separate ingress, or domain splitting.

3. Have an MCP client (e.g. Claude.ai custom connector) connect to `https://mcp.example.com/servers/{server_id}/mcp`. The client completes the OAuth 2.1 flow with the IdP, sending `resource=https://mcp.example.com/servers/{server_id}/mcp` in the authorization request.

4. The IdP ignores the `resource` parameter (no RFC 8707 support) and issues an access token with:
   ```json
   {
     "aud": "my-oauth-client-id",
     "iss": "https://idp.example.com/application/o/my-app/",
     "sub": "user@example.com",
     ...
   }
   ```

5. The MCP client sends the token to `https://mcp.example.com/servers/{server_id}/mcp`.

6. ContextForge builds `expected_audiences` from `APP_DOMAIN`:
   ```python
   resource_url = f"{settings.app_domain}/servers/{server_id}/mcp"
   # → "https://internal-gateway.example.com/servers/{server_id}/mcp"
   expected_audiences = [resource_url]
   ```

7. Token verification fails:
   ```
   expected_audiences = ["https://internal-gateway.example.com/servers/{server_id}/mcp"]
   actual token aud   = "my-oauth-client-id"
   → Audience doesn't match → 401 Unauthorized
   ```

---

### 🤔 Expected Behavior

The operator should be able to configure audience verification behavior per Virtual Server to accommodate IdPs that do not support RFC 8707 Resource Indicators. Specifically:

- A per-server toggle (e.g. `include_client_id_in_audience` in `oauth_config`) that automatically adds the SSO `client_id` to the `expected_audiences` list.
- When enabled, `expected_audiences` would be `[resource_url, sso_client_id]`, allowing tokens with `aud=client_id` to pass verification.
- The existing `oauth_config.resource` workaround requires direct DB manipulation since the field is not exposed in the admin UI. A UI-accessible toggle would be the proper solution.

---

### 📓 Logs / Error Output

```
WARN [mcpgateway.utils.verify_credentials] OAuth access token verification failed
  (issuer=https://idp.example.com/application/o/my-app/): Audience doesn't match
```

This repeats for every MCP request. The token signature and issuer are valid — only the `aud` claim fails to match.

**IdP OIDC Discovery** (`/.well-known/openid-configuration`):
- `resource_indicators_supported` field is **absent** (confirming no RFC 8707 support)
- IdP developer confirmed: "the `aud` claim is set to the provider's client_id by default"

**Audience verification code path** (`streamablehttp_transport.py` ~L3999):
```python
resource_url = _build_server_resource_url(self.scope, server_id)
# → Always derived from settings.app_domain, not the inbound Host header

expected_audiences: list[str] = [resource_url]
extra_audience = server.oauth_config.get("resource") or server.oauth_config.get("client_id")
# ... extend expected_audiences ...

claims = await verify_oauth_access_token(token, authorization_servers, expected_audience=expected_audiences)
```

The `resource_url` is always `{APP_DOMAIN}/servers/{id}/mcp`. When `APP_DOMAIN` differs from the public domain AND the IdP sets `aud=client_id` (not the resource URL), there is no matching audience.

**Current workaround**: Manually insert `client_id` into `oauth_config.resource` via direct SQL:
```sql
UPDATE servers SET oauth_config = '{"authorization_servers": [...], "resource": ["https://mcp.example.com/...", "my-oauth-client-id"]}'::json WHERE id = '...';
```
This works but requires DB access and is not documented.

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Runtime | Python 3.12, Uvicorn |
| Platform / OS | EKS (Amazon Linux 2, arm64) |
| Container | Custom Containerfile build from upstream main |
| IdP | Authentik (no RFC 8707 support) |

---

**Why this matters**: The MCP specification (2025-11-25) requires clients to implement RFC 8707 Resource Indicators, but many widely-used IdPs do not support it on the server side:

- **Authentik**: No RFC 8707 support. `aud` = `client_id` always.
- **Keycloak**: RFC 8707 support is still in discussion.
- Many smaller OIDC providers also lack RFC 8707 support.

**The domain-split scenario is common in production**: organizations often expose an internal admin UI on one domain (behind VPN) and a public MCP API endpoint on a different domain (with IP allowlisting). `APP_DOMAIN` cannot serve both purposes.

madhav165:
# 🐛 Bug-fix PR

## 📌 Summary

OAuth audience verification always fails for virtual servers when the IdP does not support RFC 8707 Resource Indicators (e.g. Authentik, some Keycloak configurations). These IdPs set the token `aud` claim to an abstract identifier (typically the OAuth `client_id`) rather than the resource URL the client requested. Since virtual servers only have `authorization_servers` in their `oauth_config` (no `resource`), the only expected audience is the canonical resource URL derived from `APP_DOMAIN` — which never matches.

This is the virtual-server counterpart to #4404, which applies the same auto-learn pattern to the gateway (OAuth callback) path.

## 🔁 Reproduction Steps

Closes #4171

1. Create a virtual server with `oauth_enabled=true` and an authorization server pointing to an IdP without RFC 8707 support (e.g. Authentik)
2. Set `APP_DOMAIN` to the gateway's internal domain
3. Have an MCP client connect to `/servers/{server_id}/mcp` — the client completes the OAuth flow with the IdP
4. The IdP issues a token with `aud: "my-oauth-client-id"` (ignoring the `resource` parameter)
5. ContextForge builds `expected_audiences = ["{APP_DOMAIN}/servers/{server_id}/mcp"]`
6. Audience mismatch → 401 on every request

## 🐞 Root Cause

`_try_oauth_access_token` in `streamablehttp_transport.py` always enforced audience verification using the canonical resource URL from `_build_server_resource_url()`. When `oauth_config` has no `resource` (which is the case for all virtual servers created via the admin UI), the expected audience list contained only this URL. Non-RFC-8707 IdPs set `aud` to an abstract identifier instead, so every token failed audience verification — even though signature and issuer were valid.

The admin UI for virtual servers does not expose `resource` fields, so there was no way to configure the expected audience without direct DB manipulation.

## 💡 Fix Description

Replace the old audience-list-building logic with a two-branch strategy in `_try_oauth_access_token`:

1. **When `resource` IS configured** (learned or manually set): Pass it directly to `verify_oauth_access_token` as the expected audience. No canonical resource URL, no list building — the configured value is used as-is (string or list).

2. **When `resource` is NOT configured**: Skip audience enforcement (`expected_audience=None`). Signature + issuer are still verified. After successful verification, extract the `aud` claim from the verified token and persist it as `resource` in `server.oauth_config` via the new `_persist_learned_server_audience()` helper. All subsequent requests then get strict audience enforcement.

The old `client_id` fallback and canonical resource URL derivation (`_build_server_resource_url`) are removed from the audience path entirely — they were the source of the bug for non-RFC-8707 IdPs.

This is not a security regression: without this change, every non-RFC-8707 IdP token is rejected outright (the feature is completely broken). The fix maintains signature + issuer verification on the first request and adds strict audience enforcement from the second request onward.

**Why auto-learn instead of a UI toggle (as suggested in #4171)?**

The issue proposed a per-server toggle or UI field to manually configure the expected audience. But the operator only controls which IdP to use (`authorization_servers`) — the `aud` claim value is determined entirely by the IdP's behavior, which varies across providers (it could be the `client_id`, an application URI, or something else entirely). Asking the operator to pre-configure it would require them to know what the IdP will put in `aud` before any token has been issued. Auto-learning from the first verified token eliminates this guesswork and works correctly regardless of the IdP's audience strategy.

Key implementation details:
- `_persist_learned_server_audience()` is best-effort — DB failures are caught both internally and at the call site, never breaking the auth flow
- The `aud` claim is stored as-is: string values stay strings, list values stay lists
- Persist is called on every successful auth so that IdP-side changes (e.g. client_id rotation) are picked up automatically; a no-op early return skips the write when the stored value already matches
- SQLAlchemy mutation tracking is handled correctly (new dict object assigned)

## 🧪 Verification

| Unit tests (test_auth.py)             | `uv run pytest tests/unit/mcpgateway/test_auth.py` | ✅ 237 passed |
| Related OAuth tests                   | `uv run pytest tests/unit/mcpgateway/test_auth.py tests/unit/mcpgateway/routers/test_oauth_router.py tests/unit/mcpgateway/services/test_token_validation_service.py` | ✅ 362 passed |
| Doctests                              | `uv run pytest --doctest-modules mcpgateway/transports/streamablehttp_transport.py` | ✅ 22 passed |
| Ruff                                  | `make ruff` | ✅ All checks passed |
| Formatting                            | `make autoflake isort black` | ✅ Clean |

## ✅ Checklist

### DIFF
--- a/mcpgateway/transports/streamablehttp_transport.py
+++ b/mcpgateway/transports/streamablehttp_transport.py
@@ -67,7 +67,8 @@ from mcpgateway.cache.global_config_cache import global_config_cache
 from mcpgateway.common.models import LogLevel
 from mcpgateway.common.validators import validate_meta_data as _validate_meta_data
 from mcpgateway.config import settings
-from mcpgateway.db import Server as DbServer, SessionLocal
+from mcpgateway.db import Server as DbServer
+from mcpgateway.db import SessionLocal
 from mcpgateway.middleware.rbac import _ACCESS_DENIED_MSG
 from mcpgateway.observability import create_span
 from mcpgateway.plugins.framework.models import UserContext
--- a/mcpgateway/transports/streamablehttp_transport.py
+++ b/mcpgateway/transports/streamablehttp_transport.py
@@ -55,6 +55,7 @@ from mcp.server.streamable_http import EventCallback, EventId, EventMessage, Eve
 from mcp.server.streamable_http_manager import StreamableHTTPSessionManager
 from mcp.types import JSONRPCMessage, PaginatedRequestParams, ReadResourceRequest, ReadResourceRequestParams
 import orjson
+from sqlalchemy import select
 from sqlalchemy.exc import SQLAlchemyError
 from sqlalchemy.orm import Session
 from starlette.datastructures import Headers
@@ -66,7 +67,7 @@ from mcpgateway.cache.global_config_cache import global_config_cache
 from mcpgateway.common.models import LogLevel
 from mcpgateway.common.validators import validate_meta_data as _validate_meta_data
 from mcpgateway.config import settings
-from mcpgateway.db import SessionLocal
+from mcpgateway.db import Server as DbServer, SessionLocal
 from mcpgateway.middleware.rbac import _ACCESS_DENIED_MSG
 from mcpgateway.observability import create_span
 from mcpgateway.plugins.framework.models import UserContext
@@ -916,6 +917,47 @@ def _build_resource_metadata_url(scope: Scope, server_id: str) -> str:
     if not base:
         return ""
     return f"{base}/.well-known/oauth-protected-resource/servers/{server_id}/mcp"
+
+
+def _persist_learned_server_audience(server_id: str, verified_claims: dict[str, Any], db: Session) -> None:
+    """Persist the ``aud`` claim from a verified OAuth token as ``resource``.
+
+    Called after successful signature + issuer verification. The token's
+    ``aud`` claim is trustworthy at this point and represents the IdP's
+    authoritative audience value. Always overwrites the existing ``resource``
+    so that IdP-side changes (client_id rotation, application migration)
+    are picked up automatically.
+
+    This is a best-effort operation: failures are logged but do not affect
+    the current request's authentication outcome.
+
+    Args:
+        server_id: Virtual-server identifier.
+        verified_claims: Decoded and *signature-verified* JWT claims.
+        db: Active database session.
+    """
+    raw_aud = verified_claims.get("aud")
+    if not raw_aud:
+        return
+
+    try:
+        server = db.execute(select(DbServer).where(DbServer.id == server_id)).scalar_one_or_none()
+        if server is None or not server.oauth_config:
+            return
+
+        if server.oauth_config.get("resource") == raw_aud:
+            return  # Already correct
+
+        updated_config = dict(server.oauth_config)
+        updated_config["resource"] = raw_aud
+        server.oauth_config = updated_config
+        db.flush()
+        logger.info(
+            "Learned and persisted audience for server %s from verified token",
+            sanitize_for_log(server_id),
+        )
+    except Exception:
+        logger.warning("Failed to persist learned audience for server %s", server_id, exc_info=True)
 
 
 async def _check_server_oauth_enforcement(server_id: str, user_context: Optional[dict[str, Any]]) -> None:
@@ -5189,12 +5231,6 @@ class _StreamableHttpAuthHandler:
         server_id = match.group("server_id")
         server_id_log = sanitize_for_log(server_id)
 
-        # Third-Party
-        from sqlalchemy import select  # pylint: disable=import-outside-toplevel
-
-        # First-Party
-        from mcpgateway.db import Server as DbServer  # pylint: disable=import-outside-toplevel
-
         try:
             async with get_db() as db:
                 server = db.execute(select(DbServer).where(DbServer.id == server_id)).scalar_one_or_none()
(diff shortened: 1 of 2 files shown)


madhav165:
## Self-Review Notes

Reviewed the following areas that are likely to come up in automated or human review — documenting findings here to avoid re-investigation:

**1. Is removing the `client_id` fallback a regression?**
No. The old code used `oauth_config.get("resource") or oauth_config.get("client_id")` as a static fallback for audience enforcement. This PR replaces that with a dynamic learning mechanism: if no `resource` is configured, skip audience enforcement on the first request (signature + issuer still verified), learn the IdP's actual `aud`, and persist it as `resource`. The `client_id` fallback was a workaround for the exact problem this PR solves properly. Existing deployments with `client_id` set will auto-learn the correct audience on the next verified token.

**2. Is removing the canonical resource URL (`_build_server_resource_url`) from audience checks a security regression?**
No. The canonical URL derivation was the root cause of the bug — non-RFC-8707 IdPs never set `aud` to the resource URL, so every token failed. Removing it from the audience path is the fix. After the first successful verification, the learned `aud` is persisted and enforced strictly on all subsequent requests. This mirrors the approach in #4404 for the gateway path.

**3. Does the always-overwrite behavior in `_persist_learned_server_audience` allow audience drift?**
This is a deliberate design choice, consistent with #4404. The persist function overwrites `resource` when the token's `aud` differs from the stored value, so that IdP-side changes (client_id rotation, application migration) are picked up automatically. A no-op early return skips the write when the stored value already matches. The threat scenario (attacker with a validly-signed token from the same IdP but different `aud`) requires the attacker to already have a valid token from the configured authorization server.

**4. Is the double DB read for the same server a performance concern?**
The first session (line 5067) is intentionally short-lived — it closes before `verify_oauth_access_token`, which performs async network I/O (JWKS fetch). Holding a DB connection open across that call risks connection pool starvation. The second read inside `_persist_learned_server_audience` (line 943) needs an active session for the write. This is structurally necessary given the current architecture. The persist path is also effectively a no-op after the first request (early return when `resource == raw_aud`), so the second read only does real work once per server.

**5. Does `_persist_learned_server_audience` need locking or `SELECT ... FOR UPDATE`?**
No. Concurrent requests that both read `resource=None` and both write the same learned `aud` produce the same result — the write is idempotent. There is no TOCTOU risk because the persisted value is the same regardless of which request wins.

mcpgateway/transports/streamablehttp_transport.py
```diff
@@ -916,6 +919,47 @@ def _build_resource_metadata_url(scope: Scope, server_id: str) -> str:
...
+
+    This is a best-effort operation: failures are logged but do not affect
+    the current request's authentication outcome.
+
+    Args:
+        server_id: Virtual-server identifier.
+        verified_claims: Decoded and *signature-verified* JWT claims.
+        db: Active database session.
+    """
+    raw_aud = verified_claims.get("aud")
```
brian-hussey, line 993: Can we validate the structure of the 'aud' claim here? At the moment it is accepted and trusted by default.
The token is verified but it could still be complex, empty or an unexpected type.
We could validate that it is a non-empty string or a list of non-empty strings before continuing.

mcpgateway/transports/streamablehttp_transport.py
```diff
@@ -916,6 +919,47 @@ def _build_resource_metadata_url(scope: Scope, server_id: str) -> str:
...
+    if not raw_aud:
+        return
+
+    try:
+        server = db.execute(select(DbServer).where(DbServer.id == server_id)).scalar_one_or_none()
+        if server is None or not server.oauth_config:
```
brian-hussey, line 1005: The function returns early if server.oauth_config is falsy, but this means a server with oauth_enabled=True but no oauth_config will silently skip audience learning. This could hide configuration issues. Consider logging a warning when oauth_enabled is True but oauth_config is missing.

mcpgateway/transports/streamablehttp_transport.py
```diff
@@ -916,6 +919,47 @@ def _build_resource_metadata_url(scope: Scope, server_id: str) -> str:
...
+            return
+
+        if server.oauth_config.get("resource") == raw_aud:
+            return  # Already correct
+
+        updated_config = dict(server.oauth_config)
+        updated_config["resource"] = raw_aud
+        server.oauth_config = updated_config
+        db.flush()
```
brian-hussey, line 1019: There is a potential race condition here: Using db.flush() without explicit transaction isolation. 
If multiple requests with different aud claims arrive concurrently for the same server, the last write wins without any conflict detection. Consider adding optimistic locking (version field) or documenting that the last-verified-token-wins behavior is intentional.

brian-hussey (changes requested):
Hey Madhav, thank you for the submission.
There are a few issues with this PR as commented below.
Also please consider updating the docs/docs/architecture/oauth-design.md with updates to the oauth design that will come through with this change.

Thanks,
Brian

PR #4411, opened 2026-04-23
ci: enable path filters for lint and pytest workflows

jonpspri:
## Summary
- Uncomment path filters on the `push` and `pull_request` triggers in `.github/workflows/lint.yml` and `.github/workflows/pytest.yml`.
- Workflows now run only when relevant Python source, tests, workflow, or dependency files change, skipping unrelated docs/chart changes.

## Test plan
- [ ] Verify workflows still trigger on PRs touching `mcpgateway/`, `plugins/`, `tests/`, `pyproject.toml`, `uv.lock`, or the workflow files themselves
- [ ] Verify workflows are skipped on PRs that only touch unrelated paths (e.g. `docs/`, `charts/`)


.github/workflows/rust.yml
```diff
@@ -4,20 +4,8 @@ name: Rust CI
 
 on:
   push:
-    branches: [main, develop]
     tags:
       - "v*"
-    paths:
```
lucarlig, line 10: why remove paths? it will then run on every push, and it's one of the slowest workflows. 
Should the whole push be removed?

lucarlig (changes requested):

.github/workflows/rust.yml
jonpspri: I think so.  AI botched that one -- I'll go fix it now.

lucarlig (approved):

PR #4282, opened 2026-04-17
fix: rewrite bare /mcp to /mcp/ in MCPPathRewriteMiddleware so stream…

omorros:
## 🔗 Related Issue
  Closes #4275                                                                                                                                  
   
  ---                                                                                                                                           
                  
  ## 📝 Summary
  `MCPPathRewriteMiddleware` only rewrote `/servers/<id>/mcp` to `/mcp/`, leaving bare `/mcp` untouched. Because the Starlette mount is at
  `/mcp` (compiled as `/mcp/{path:path}`), `/mcp` without the trailing slash did not match and FastAPI's router emitted a 307 redirect to
  `/mcp/`. For chunked Streamable HTTP POSTs the client cannot replay the body across a redirect, so the handshake failed with `httpx.ReadError`
   during `initialize`.

  This PR normalises bare `/mcp` to `/mcp/` inside the middleware so the mount matches directly without a redirect. It also mirrors every
  rewrite into `scope["raw_path"]` alongside `scope["path"]` (both the new bare-`/mcp` branch and the existing `/servers/<id>/mcp` branch),
  keeping downstream URL reconstruction consistent with the rewritten path.

  Security posture is unchanged: arbitrary prefixes ending in `/mcp` (e.g. `/foo/mcp`) still pass through unrewritten, and empty server IDs
  (`/servers//mcp`) still return 404.

  ---

  ## 🏷️  Type of Change
  - [x] Bug fix
  - [ ] Feature / Enhancement
  - [ ] Documentation
  - [ ] Refactor
  - [ ] Chore (deps, CI, tooling)
  - [ ] Other (describe below)

  ---

  ## 🧪 Verification

  | Check                     | Command         | Status |
  |---------------------------|-----------------|--------|
  | Lint suite                | `make lint`     | ✅ ruff/isort/black clean on changed lines (pre-existing drift untouched) |
  | Unit tests                | `make test`     | ✅ 1864+ related tests pass (main, middleware, auth, streamable HTTP transport) |
  | Coverage ≥ 80%            | `make coverage` | ⏳ to be verified in CI |

  ---

  ## ✅ Checklist
  - [x] Code formatted (`make black isort pre-commit`)
  - [x] Tests added/updated for changes
  - [ ] Documentation updated (if applicable)
  - [x] No secrets or credentials committed

  ---

  ## 📓 Notes (optional)
  Four new unit tests in `TestMCPPathRewriteMiddleware` cover:
[...]

PR #4446, opened 2026-04-26
fix(mcp): prevent 307 redirects for Streamable HTTP /mcp probes

### LINKED ISSUE #4441: [BUG]: GET /mcp/ returns 307 redirect in Edge mode instead of spec-required 200 (SSE) or 405
### 🐞 Bug Summary

`test_get_mcp_returns_sse_stream_or_405` in `tests/protocol_compliance/test_transport_semantics.py` fails in Edge (Rust) mode because `GET /mcp/` returns HTTP 307 (Temporary Redirect) instead of the MCP-spec-mandated 200 (`text/event-stream`) or 405.

The MCP specification ([Streamable HTTP transport, "Listening for Messages from the Server"](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports)) states:

> The server **MUST** either return `Content-Type: text/event-stream` in response to this HTTP GET, or else return HTTP 405 Method Not Allowed, indicating that the server does not offer an SSE stream at this endpoint.

A 307 redirect is not a spec-sanctioned response. This is a known ecosystem-wide issue — the Python SDK fixed the same trailing-slash redirect problem ([python-sdk#1168](https://github.com/modelcontextprotocol/python-sdk/issues/1168)), as did FastMCP ([fastmcp#1364](https://github.com/PrefectHQ/fastmcp/issues/1364)).

---

### 🧩 Affected Component

- [x] Other: Starlette mount trailing-slash behavior on the Python gateway

---

### 🔁 Steps to Reproduce

1. Run protocol compliance tests against an Edge-mode deployment
2. `pytest tests/protocol_compliance/test_transport_semantics.py::test_get_mcp_returns_sse_stream_or_405`

---

### 🤔 Expected Behavior

`GET /mcp/` should return either:
- HTTP 200 with `Content-Type: text/event-stream` (SSE stream), or
- HTTP 405 Method Not Allowed

---

### 📓 Logs / Error Output

```
tests/protocol_compliance/test_transport_semantics.py:110: in test_get_mcp_returns_sse_stream_or_405
    assert status == 200, f"unexpected status for GET /mcp/: {status}"
E   AssertionError: unexpected status for GET /mcp/: 307
E   assert 307 == 200
```

Note: the test client has `follow_redirects=True` but still surfaces the 307, suggesting the redirect may be malformed (missing `Location` header) or looping.

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Runtime mode | Edge (Rust MCP runtime active) |
| Deployment | docker-compose, multi-replica behind Nginx |

---

### 🧩 Additional Context

**The Rust MCP runtime is NOT the source of the 307.** It defines both `/mcp` and `/mcp/` routes with identical handlers and contains zero redirect logic.

The 307 originates from the Python/Starlette layer. The gateway mounts the MCP transport via:

```python
app.mount("/mcp", app=mcp_transport_app.handle_streamable_http)  # main.py:12322
```

Starlette's `Mount` performs automatic trailing-slash normalization. The request flow in Edge mode is:

1. Nginx proxies `GET /mcp/` to Rust sidecar at `:8787`
2. On Rust sidecar failure (502/503/504), Nginx falls back to the Python gateway at `:4444`
3. The Python gateway's Starlette mount issues a 307 redirect

The MCP Authorization spec also provides relevant guidance: implementations **SHOULD** consistently use the form **without** trailing slash for better interoperability.

**Fix direction**: The Python gateway should handle both `/mcp` and `/mcp/` without redirecting, matching what the Rust side already does correctly.

cooleryu:
## Summary

Normalize MCP mount rewrites before Starlette can emit a trailing-slash `307` redirect, and keep ASGI `raw_path` aligned when the middleware rewrites MCP transport paths.

This prevents exact `GET /mcp` probes from being converted into a framework-level redirect before they reach the MCP transport.

## Motivation

Fixes #4441.

The MCP Streamable HTTP transport allows `GET /mcp` to either:

- return `200` with `Content-Type: text/event-stream`, or
- return `405 Method Not Allowed` when the server does not offer an SSE stream at that endpoint.

A `307` redirect is not a valid transport response for that probe. The issue points to the Python/Starlette gateway layer as the source of the redirect, not the Rust MCP runtime.

The root cause is Starlette mount behavior: an exact `/mcp` request against `app.mount("/mcp", ...)` is redirected to `/mcp/`. ContextForge already has `MCPPathRewriteMiddleware` in front of the mounted transport, so this patch normalizes the exact `/mcp` path there after authentication.

I also noticed #4282, which addresses the same redirect class. This PR follows the same fix direction and adds extra coverage for the actual Starlette mount behavior, `APP_ROOT_PATH`, canonical `/mcp/` pass-through, RFC 9728 well-known path exclusion, and `raw_path` alignment.

## Changes

- Normalize exact `/mcp` to `/mcp/` in `MCPPathRewriteMiddleware`.
- Preserve reverse-proxy prefixes when `root_path` or `APP_ROOT_PATH` is set.
- Keep ASGI `raw_path` aligned with rewritten `/mcp/` paths when present.
- Leave already-canonical `/mcp/` unchanged.
- Keep RFC 9728 well-known URLs ending in `/mcp` out of the MCP transport rewrite path.
- Add a regression test that wires the middleware in front of a real Starlette `Mount("/mcp", ...)` and verifies `/mcp` no longer returns a `Location` redirect.
- Add unit coverage for exact `/mcp`, `root_path`, `APP_ROOT_PATH`, canonical `/mcp/`, well-known URI boundaries, and `raw_path` synchronization.

## Test Plan

- `uv run --group dev pytest tests/unit/mcpgateway/test_main_extended.py::TestMCPPathRewriteMiddleware -q`
  - Result: `22 passed`
- `uv run --group dev pytest tests/unit/mcpgateway/transports/test_streamablehttp_transport.py::test_streamable_http_auth_requires_auth_for_trailing_slash_variants tests/unit/mcpgateway/transports/test_streamablehttp_transport.py::test_streamable_http_auth_skips_well_known_rfc9728_even_when_strict -q`
  - Result: `5 passed`
- `uv run --group dev pytest tests/unit/mcpgateway/test_main_extended.py -q`
  - Result: passed
- `uv run python -m py_compile mcpgateway/main.py tests/unit/mcpgateway/test_main_extended.py`
  - Result: passed
- `uv run ruff check mcpgateway/main.py`
  - Result: passed
- `git diff --check`
  - Result: passed
- ASGI smoke test with Edge-mode environment variables and no Rust sidecar:
  - Before this patch: `/mcp -> 307 Location: /mcp/`
  - After this patch: `/mcp -> 502` and `/mcp/ -> 502`, both without `Location`
  - The `502` is expected in this local smoke test because the Rust MCP sidecar is not running; the relevant regression check is that `/mcp` no longer returns `307`.

I also ran:

- `uv run --group dev pytest tests/protocol_compliance/test_transport_semantics.py::test_get_mcp_returns_sse_stream_or_405 -q`
  - Result: skipped because no live gateway is running at `http://127.0.0.1:8080`.

Running Ruff on the full `tests/unit/mcpgateway/test_main_extended.py` file still reports existing findings outside this diff, so this patch avoids mixing unrelated cleanup into the bug fix.

## Risk

Low. The change is scoped to MCP transport paths after MCP authentication has already succeeded.

It does not broaden arbitrary path rewrites. Non-server paths such as `/foo/mcp` still pass through unchanged, and well-known OAuth metadata paths remain excluded from MCP transport rewriting.

The `raw_path` update mirrors the rewritten `path` value so downstream ASGI consumers do not see inconsistent path metadata after middleware normalization.

## Related Issue

Fixes #4441.

Related: #4282.

## Maintainer Context

This keeps the fix in the existing MCP path normalization layer instead of changing the MCP transport, Rust runtime proxy, or Nginx configuration.

The patch is intentionally small and includes boundary tests for reverse-proxy prefixes, ASGI raw path metadata, already-canonical MCP paths, and non-MCP metadata paths so the redirect fix does not expand the public route surface.

If maintainers prefer the implementation in #4282, the additional regression and boundary tests in this PR can be reused directly.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -5118,7 +5118,7 @@
         "hashed_secret": "4f13f134744a2fadbbe2d624687246347d12fa63",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 2731,
+        "line_number": 2733,
         "type": "Hex High Entropy String",
         "verified_result": null
       }
--- a/mcpgateway/main.py
+++ b/mcpgateway/main.py
@@ -2809,8 +2809,9 @@ class MCPPathRewriteMiddleware:
     """
     Middleware that rewrites paths ending with '/mcp' to '/mcp/', after performing authentication.
 
+    - Rewrites exact '/mcp' to '/mcp/' so Starlette's mount does not emit a 307 redirect.
     - Rewrites paths like '/servers/<server_id>/mcp' to '/mcp/'.
-    - Only paths ending with '/mcp' or '/mcp/' (but not exactly '/mcp' or '/mcp/') are rewritten.
+    - Only exact '/mcp' and server-scoped MCP transport paths are rewritten.
     - Authentication is performed before any path rewriting.
     - If authentication fails, the request is not processed further.
     - All other requests are passed through without change.
@@ -2950,6 +2951,10 @@ class MCPPathRewriteMiddleware:
         # Skip rewriting for well-known URIs (RFC 9728 OAuth metadata, etc.)
         # These paths may end with /mcp but should not be rewritten to the MCP transport
         if not app_path.startswith("/.well-known/"):
+            if app_path == "/mcp":
+                scope["path"] = f"{root_path}/mcp/" if root_path else "/mcp/"
+                await self.application(scope, receive, send)
+                return
             if (app_path.endswith("/mcp") and app_path != "/mcp") or (app_path.endswith("/mcp/") and app_path != "/mcp/"):
                 # SECURITY: Only rewrite recognised MCP paths — /servers/{id}/mcp.
                 # Arbitrary prefixes (e.g. /foo/mcp) must NOT be rewritten to
--- a/mcpgateway/main.py
+++ b/mcpgateway/main.py
@@ -2811,6 +2811,7 @@ class MCPPathRewriteMiddleware:
 
     - Rewrites exact '/mcp' to '/mcp/' so Starlette's mount does not emit a 307 redirect.
     - Rewrites paths like '/servers/<server_id>/mcp' to '/mcp/'.
+    - Keeps ASGI ``raw_path`` aligned with rewritten paths when present.
     - Only exact '/mcp' and server-scoped MCP transport paths are rewritten.
     - Authentication is performed before any path rewriting.
     - If authentication fails, the request is not processed further.
@@ -2952,7 +2953,10 @@ class MCPPathRewriteMiddleware:
         # These paths may end with /mcp but should not be rewritten to the MCP transport
         if not app_path.startswith("/.well-known/"):
             if app_path == "/mcp":
-                scope["path"] = f"{root_path}/mcp/" if root_path else "/mcp/"
+                new_path = f"{root_path}/mcp/" if root_path else "/mcp/"
+                scope["path"] = new_path
+                if "raw_path" in scope:
+                    scope["raw_path"] = new_path.encode("latin-1")
                 await self.application(scope, receive, send)
                 return
             if (app_path.endswith("/mcp") and app_path != "/mcp") or (app_path.endswith("/mcp/") and app_path != "/mcp/"):
@@ -2976,7 +2980,10 @@ class MCPPathRewriteMiddleware:
                     return
                 # Rewrite to /mcp/ and continue through middleware (lets CORSMiddleware handle preflight)
                 # Preserve root_path prefix when rewriting
-                scope["path"] = f"{root_path}/mcp/" if root_path else "/mcp/"
+                new_path = f"{root_path}/mcp/" if root_path else "/mcp/"
+                scope["path"] = new_path
+                if "raw_path" in scope:
+                    scope["raw_path"] = new_path.encode("latin-1")
                 await self.application(scope, receive, send)
                 return
         await self.application(scope, receive, send)
--- a/mcpgateway/main.py
+++ b/mcpgateway/main.py
@@ -2929,6 +2929,13 @@ class MCPPathRewriteMiddleware:
             >>> with patch('mcpgateway.main.streamable_http_auth', return_value=True):
             ...     asyncio.run(middleware._call_streamable_http(scope, receive, send))
             >>> app_mock.assert_called_once_with(scope, receive, send)
+
+            >>> # Exact /mcp is normalized to avoid Starlette's mount redirect.
+            >>> scope = {"type": "http", "path": "/mcp"}
+            >>> with patch('mcpgateway.main.streamable_http_auth', return_value=True):
+            ...     asyncio.run(middleware._call_streamable_http(scope, receive, send))
+            >>> scope["path"]
+            '/mcp/'
         """
         # Auth check first
         auth_ok = await streamable_http_auth(scope, receive, send)
@@ -2953,13 +2960,10 @@ class MCPPathRewriteMiddleware:
         # These paths may end with /mcp but should not be rewritten to the MCP transport
         if not app_path.startswith("/.well-known/"):
             if app_path == "/mcp":
-                new_path = f"{root_path}/mcp/" if root_path else "/mcp/"
-                scope["path"] = new_path
-                if "raw_path" in scope:
-                    scope["raw_path"] = new_path.encode("latin-1")
+                self._apply_mcp_rewrite(scope, root_path)
                 await self.application(scope, receive, send)
                 return
-            if (app_path.endswith("/mcp") and app_path != "/mcp") or (app_path.endswith("/mcp/") and app_path != "/mcp/"):
+            if app_path.endswith("/mcp") or (app_path.endswith("/mcp/") and app_path != "/mcp/"):
                 # SECURITY: Only rewrite recognised MCP paths — /servers/{id}/mcp.
                 # Arbitrary prefixes (e.g. /foo/mcp) must NOT be rewritten to
                 # /mcp/ as that would expose the global MCP transport under
@@ -2980,13 +2984,27 @@ class MCPPathRewriteMiddleware:
                     return
                 # Rewrite to /mcp/ and continue through middleware (lets CORSMiddleware handle preflight)
                 # Preserve root_path prefix when rewriting
-                new_path = f"{root_path}/mcp/" if root_path else "/mcp/"
-                scope["path"] = new_path
-                if "raw_path" in scope:
-                    scope["raw_path"] = new_path.encode("latin-1")
+                self._apply_mcp_rewrite(scope, root_path)
                 await self.application(scope, receive, send)
                 return
         await self.application(scope, receive, send)
+
+    @staticmethod
+    def _apply_mcp_rewrite(scope, root_path: str) -> str:
+        """Rewrite a validated MCP transport path to the mounted /mcp/ app path."""
+        original_path = scope.get("path", "")
+        new_path = f"{root_path}/mcp/" if root_path else "/mcp/"
+        scope["path"] = new_path
+
+        if "raw_path" in scope:
+            try:
+                # ASGI raw_path stores raw octets; latin-1 preserves a 1:1 byte mapping for valid values.
+                scope["raw_path"] = new_path.encode("latin-1")
+            except (UnicodeEncodeError, ValueError):
+                logger.warning("MCPPathRewriteMiddleware: non-latin-1 raw_path skipped for %s", new_path)
+
+        logger.debug("MCPPathRewriteMiddleware: %s -> %s", original_path, new_path)
+        return new_path
 
 
 # Configure CORS with environment-aware origins
--- a/tests/unit/mcpgateway/test_main_extended.py
+++ b/tests/unit/mcpgateway/test_main_extended.py
@@ -2996,6 +2996,22 @@ class TestMCPPathRewriteMiddleware:
         assert response.status_code == 200
         assert response.headers.get("location") is None
 
+    def test_rewrite_exact_mcp_post_avoids_starlette_mount_redirect(self):
+        """POST /mcp is normalized the same way as GET for Streamable HTTP."""
+
+        async def mounted_app(scope, receive, send):
+            response = StarletteResponse("ok")
+            await response(scope, receive, send)
+
+        app_with_mount = Starlette(routes=[Mount("/mcp", app=mounted_app)])
+        middleware = MCPPathRewriteMiddleware(app_with_mount)
+
+        with patch("mcpgateway.main.streamable_http_auth", new=AsyncMock(return_value=True)):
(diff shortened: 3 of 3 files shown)


msureshkumar88 (changes requested):
Good direction — the fix is minimal, correctly scoped, and the test suite shows real thought was put into edge cases (root_path, APP_ROOT_PATH, well-known exclusion, raw_path alignment). A few things need attention before this merges.

## Blocking

### 1. Unguarded `UnicodeEncodeError` on `raw_path` encoding

`new_path.encode("latin-1")` is used in both the new `/mcp` branch and the existing `/servers/{id}/mcp` branch with no exception handler. If a reverse proxy injects a non-latin-1 character into `root_path` (malformed `X-Forwarded-Prefix`), this raises an unhandled exception that crashes the worker and may leak a stack frame to the caller. Worth guarding:

```python
try:
    scope["raw_path"] = new_path.encode("latin-1")
except (UnicodeEncodeError, ValueError):
    logger.warning("MCPPathRewriteMiddleware: non-latin-1 raw_path skipped for %s", new_path)
```

Also worth a short inline comment explaining *why* latin-1 and not utf-8 — the ASGI spec reason is non-obvious and a future contributor is likely to "fix" it:
```python
# ASGI spec §2.3: raw_path is raw octets; latin-1 is the correct 1:1 byte mapping
```

### 2. CI: pre-commit hooks failure

One check is failing (`Run pre-commit hooks`). Everything else passes (pytest, pylint, ruff, security scan, Playwright). Looks like a formatting issue in the test additions — needs a `make autoflake isort black pre-commit` pass before merge.

### 3. Missing test: auth failure on exact `/mcp`

`test_rewrite_auth_failure` only covers `/servers/123/mcp`. Would be good to anchor the same invariant for the new exact `/mcp` branch explicitly — confirms the early-return auth check fires before the rewrite regardless of path:

```python
async def test_exact_mcp_auth_failure_blocks_rewrite():
    scope = {"type": "http", "path": "/mcp", "headers": []}
    with patch("mcpgateway.main.streamable_http_auth", new=AsyncMock(return_value=False)):
        await middleware._call_streamable_http(scope, receive, send)
    assert scope.get("path") == "/mcp"  # not rewritten
    app_mock.assert_not_called()
```

### 4. Missing test: `POST /mcp`

MCP Streamable HTTP uses POST for JSON-RPC. A load balancer probing with `POST /mcp` would also have triggered the 307 that this PR fixes. The integration test (`test_rewrite_exact_mcp_path_avoids_starlette_mount_redirect`) only exercises GET. Worth adding a `TestClient(middleware).post("/mcp", ...)` variant to lock in that POST is normalized identically.

---

## Suggestions (non-blocking)

**Duplicate rewrite block** — the same 5-line rewrite block appears verbatim in two branches. A private helper `_apply_mcp_rewrite(scope, root_path)` would reduce drift risk when someone touches this next:

```python
def _apply_mcp_rewrite(self, scope: dict, root_path: str) -> str:
    new_path = f"{root_path}/mcp/" if root_path else "/mcp/"
    scope["path"] = new_path
    if "raw_path" in scope:
        scope["raw_path"] = new_path.encode("latin-1")  # ASGI spec: raw octets, latin-1
    return new_path
```

**Redundant guard in the existing condition** — after the new early-return for `app_path == "/mcp"`, the `app_path != "/mcp"` sub-clause in the line below is always true. Not a bug, but confusing to read:
```python
if (app_path.endswith("/mcp") and app_path != "/mcp") or ...
#                                  ^^^^^^^^^^^^^^^^^ now unreachable
```
Either drop the clause or add `# exact /mcp handled above`.

**Missing test: `/.well-known/` + `root_path` set** — the well-known guard test (`test_well_known_mcp_suffix_is_not_rewritten`) doesn't exercise the `root_path`-strip + well-known path combination. A proxy prefix like `scope["path"] = "/gateway/.well-known/oauth-protected-resource/servers/123/mcp"` with `root_path="/gateway"` should not be rewritten — currently untested.

**Debug logging** — neither the new `/mcp` normalization nor the existing `/servers/{id}/mcp` rewrite logs anything. A `logger.debug("MCPPathRewriteMiddleware: %s -> %s", original_path, new_path)` in the shared helper above would make operator debugging significantly easier when the Rust sidecar isn't running and someone is tracing a 502.

**`_call_streamable_http` docstring** — the example still only shows `/servers/123/mcp`. A second example for exact `/mcp` normalization would keep the docstring in sync with the new branch.

---

Overall the approach is right and the scope is appropriately tight. Once the pre-commit failure is cleared, the `UnicodeEncodeError` guard is added, and the two test gaps are filled, this looks mergeable.

msureshkumar88:
Hi @cooleryu — thanks again for putting this together; the approach is solid and the edge-case coverage is genuinely thoughtful.

Just checking in on the blocking items from the review (pre-commit failure, the `UnicodeEncodeError` guard, and the two missing test variants for exact `/mcp` auth failure and `POST /mcp`). Do you have time today to address those?

If things are busy on your end, no worries at all — I'm happy to take the implementation across the finish line. Everything you've built so far will be fully credited in the commit history.

msureshkumar88 (changes requested):
All four blocking items from the previous review are done — the `UnicodeEncodeError` guard, the ASGI spec comment, the auth-failure test for exact `/mcp`, and the `POST /mcp` test. The suggestions are addressed too: shared `_apply_mcp_rewrite` helper, redundant guard removed, debug logging, docstring updated, and the `/.well-known/` + `root_path` combination test added. Good work getting through that list.

One new issue crept in while fixing the pre-commit failure.

## Blocking

### `pylint (mcpgateway)` CI is now failing — W1404 implicit string concatenation

When the `make autoflake isort black pre-commit` pass ran to fix the previous pre-commit failure, black collapsed the multiline `logger.warning(...)` at `main.py:1926` onto a single line. The two adjacent string literals now sit on the same line, and pylint flags it as `W1404: Implicit string concatenation found in call`. Pylint exits with code 4 (warning bitmask), which the CI treats as a failure — even though the overall score is still 10.00/10.

```python
# current — pylint W1404
logger.warning("SECURITY AUDIT: REQUIRE_STRONG_SECRETS is explicitly disabled in a production environment. " "This override is being logged for audit purposes as per US-1 requirements.")
```

Fix — merge the two literals into one string:

```python
logger.warning("SECURITY AUDIT: REQUIRE_STRONG_SECRETS is explicitly disabled in a production environment. This override is being logged for audit purposes as per US-1 requirements.")
```

This is the only remaining blocker. Once that one-line change lands and CI goes green, this looks ready to merge.

msureshkumar88:
Re-verified all items from both review rounds as of the latest commit.

**All four blocking items from review 1 — done.**

**Blocking item from review 2 (W1404 implicit string concatenation) — done.**
`main.py:1927` now has a single merged string literal; pylint no longer flags W1404.

**CI — all 28 checks passing, 0 failures.**

**Suggestions from review 1 — all addressed:**
- `_apply_mcp_rewrite` private helper extracted; deduplicates the two rewrite blocks at `main.py:2948` and `main.py:2972`.
- Redundant `app_path != "/mcp"` guard removed.
- Debug logging added.
- `_call_streamable_http` docstring updated with exact `/mcp` example.
- `test_well_known_mcp_suffix_with_root_path_is_not_rewritten` covers the `/.well-known/` + `root_path` combination.

No new issues observed. This looks ready to merge.

msureshkumar88:
## ✅ E2E Test Results - All Tests Passed

I've created and executed an automated E2E test suite to validate that PR #4446 fixes issue #4441.

### Test Execution Summary
- **Date**: 2026-05-13 10:51:10 UTC
- **Base URL**: http://localhost:4444
- **Tests Passed**: 8/8 ✅

### Test Results

| Test | Status | Details |
|------|--------|---------|
| GET /mcp No Redirect | ✅ | Returns 405 (no 307 redirect - path rewrite working) |
| POST /mcp JSON-RPC | ✅ | Accepts JSON-RPC requests correctly |
| Auth Enforcement | ✅ | Unauthenticated requests handled properly (no 307) |
| Server-Scoped MCP Path | ✅ | /servers/{id}/mcp returns 405 (no 307 redirect) |
| Invalid Server ID | ✅ | /servers//mcp returns 404 as expected |
| Well-Known URI Exclusion | ✅ | Well-known URIs not rewritten to /mcp/ |
| Non-MCP Path Pass-Through | ✅ | /foo/mcp not rewritten to /mcp/ |
| Canonical /mcp/ Path | ✅ | /mcp/ returns 405 (no 307 redirect) |

### Issue Validation

✅ **Issue #4441 is FIXED**: GET /mcp now returns the correct response instead of a 307 redirect

### Security Validation

✅ Authentication is enforced before path rewriting
✅ Boundary protection verified (well-known URIs, arbitrary paths excluded)
✅ Server-scoped MCP paths work correctly
✅ Invalid server IDs are properly rejected

### Conclusion

All test scenarios passed successfully. PR #4446 correctly fixes the 307 redirect issue while maintaining security boundaries and proper path handling.

**Recommendation**: ✅ Ready to merge

---

**Test Script**: `tests/e2e/test_mcp_path_rewrite_pr4446.py`
**Full Report**: See attached `pr4446_findings.md`

msureshkumar88 (dismissed):
LGMT

ja8zyjits (dismissed):
LGTM

ja8zyjits:
@cooleryu would you like to rebase this?
Please let us know.
In next 24 hours if you are unable to rebase with secrets fix, then we will do it for you. We believe this change is good and we can take this in asap.

brian-hussey:
@cooleryu We have a failure as your refresh secrets baseline commit does not have your sign off to comply with the Developer Certificate of Origin (DCO) compliance.

If you're happy to sign off please:
1. Ensure you have a local copy of your branch by [checking out the pull request locally via command line](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/checking-out-pull-requests-locally).
2. In your local branch, run: git rebase HEAD~4 --signoff
3. Force push your changes to overwrite the branch: git push --force-with-lease origin fix-mcp-trailing-slash-4441

This will resign the history back to where we lack the sign-off and push it up here again so we can accept your code in.

Thanks,
Brian

cooleryu:
Rebased this PR onto the latest `main` and refreshed `.secrets.baseline` after the rebase so the detect-secrets pre-commit hook is stable again.

Local verification:
- `make --no-print-directory pre-commit` passes
- `PYTHONPYCACHEPREFIX=/tmp/mcp-context-forge-pycache uv run --group dev pytest tests/unit/mcpgateway/test_main_extended.py::TestMCPPathRewriteMiddleware -q` passes: 26 passed

DCO is passing on the PR now. The remaining GitHub Actions runs are currently marked `action_required` for the forked PR, so they need maintainer approval to run.

ja8zyjits (approved):
LGTM

PR #4453, opened 2026-04-26
chore: developer tooling housekeeping — git hooks, secrets-scan tooling, agent docs, ignore files

jonpspri:
## Summary

A housekeeping pass over developer tooling and agent-facing docs, in five independent groups:

- **PR review workflow documentation** — adds a PR Review Workflow section to `AGENTS.md` codifying the fixed-point review loop (review → fix → repeat until zero blocking findings), the rebase recipe with conflict-resolution policy, scope verification against linked issues, and the pre-merge validation gate. Introduces `llms/NOTES.template.md` as the canonical, committed review-cycle state template; per-review instances are copied to `llms/NOTES.md`, which is gitignored.
- **Git tooling setup** — consolidates the git-side scripts under `scripts/git/` and adds `make configure-git`, which installs: the pre-commit framework hooks (shadowing a global `core.hooksPath`, e.g. MDM-managed scanners, for that invocation only), the `.secrets.baseline` merge driver (copied to `<git-common-dir>/git-drivers/` so it survives rebasing history that predates it), and a `post-rewrite` hook that refreshes the baseline after a rebase. Hook installation is idempotent and non-destructive: a foreign `post-rewrite` hook is moved to `post-rewrite.chain` and chained, not clobbered. The legacy `lint-install-hooks` / `lint-pre-commit` / `lint-pre-push` targets are **removed** — breaking for existing checkouts whose installed hooks still call them; see the CHANGELOG "Removed" entry and `DEVELOPING.md` for the remediation (`rm .git/hooks/pre-commit .git/hooks/pre-push && make configure-git`). The old hook ran `lint-staged`; the replacement installs the pre-commit framework shim, which runs different checks.
- **Secrets-baseline tooling** — reworks the `detect-secrets-scan` flow. Scan scope is the files changed vs `GIT_DIFF_TARGET` (default `main`, falling back to `origin/main` on clones without a local `main`; `DETECT_SECRETS_PATH` overrides the file list directly). The scan/merge/gate pipeline has a single implementation in the merge-driver script (`scripts/git/resolve-secrets-baseline-conflict.sh`): `make detect-secrets-scan` is a thin wrapper that execs it, and the installed driver and the `post-rewrite` hook carry their own standalone copies so they never depend on the Makefile of whatever tree a rebase has checked out. The jq merge keeps audited entries for tracked files outside the scan scope and drops entries for files git no longer tracks (exact filename membership). The target now exits non-zero on live/unaudited/audited-as-real findings (remediate with `make detect-secrets-audit`). **Whole-tree coverage is retained in CI**: `make detect-secrets-hook` runs over all of `git ls-files` with `--fail-on-unaudited`, and `make detect-secrets-scan-all` remains as the whole-tree variant for release tagging. A bats suite for the driver (`tests/bash/`, 16 tests) runs via `make test-bats`.
- **Ignore-file and docs hygiene** — removes duplicate and redundant entries from `.gitignore` and `.dockerignore`. One intentional behavior change here: `LICENSE`, `CHANGELOG.md`, `AUTHORS.md`, `DEVELOPING.md`, and `CONTRIBUTING.md` now enter the Docker build context (they were previously excluded); the negation block makes that explicit rather than claiming pure dedupe. Also adds ADR handling/immutability guidance to `docs/AGENTS.md` (accepted ADRs are frozen; nav entries live in `docs/docs/architecture/adr/.pages`), and keeps `test-mcp-cli` as a deprecated alias of `test-mcp-protocol-e2e` (deprecation warning restored).
- **ADR updates for the Granian removal** — restores ADR-025 to its original content, marks it *Superseded*, and adds the pointer note to ADR-054; ADR-054 records the decision to remove Granian and standardize on Gunicorn + Uvicorn (implementation merged in #5607, refs #5556). Index and `.pages` navigation updated.

## Design Decisions

- **Single source of truth for the review procedure**: the process lives only in `AGENTS.md`; `llms/NOTES.template.md` holds only mutable per-review state (cycle counter, phase toggles, gate checkboxes).
- **ADR immutability**: accepted ADRs are point-in-time records. Only status/front-material updates and additive pointer notes are permitted; changed decisions are captured in new superseding ADRs, following the existing ADR-038 → ADR-043 pattern. The ADR-025 → ADR-054 change applies that guidance directly.
- **Installed scripts cannot trust the tree**: the merge driver and post-rewrite hook run outside the worktree against whatever a rebase has checked out, so each is self-contained (env-overridable defaults matching the Makefile) rather than invoking `make`.
- **Fail-closed on a security gate**: the audit gate never returns a raw count as an exit status (exit codes are mod 256), treats unreadable audit stats as failure with an actionable message, and the merge driver aborts if any `%B` verdict audited as a real secret would vanish from the merged baseline (matched by `hashed_secret`, immune to line-number drift). A `+N / -M` entry summary makes automated baseline changes auditable.


AGENTS.md
```diff
@@ -73,7 +73,52 @@ make ruff bandit interrogate pylint verify
...
+If `llms/NOTES.md` is missing or stale, reset from the template; never assume prior cycle state carries over.
+
+### Rebase
+
+```bash
+(cd ../mcp-context-forge && git pull)
+git rebase main
+```
+
+For each conflict, apply the resolution that best preserves both intents using the PR diff and recent main commits as context. Escalate when the conflict is semantic (logic intent on both sides), would silently weaken a test / security check / migration, or when you're not confident the fix matches the PR author's intent. Preserve sign-off (`git commit -s`) on any new commits.
```
cafalchio: Please clarify if it can it would merge without monitoring.

AGENTS.md
```diff
@@ -73,7 +73,52 @@ make ruff bandit interrogate pylint verify
...
+1. **Verify scope.** Cross-reference the PR description against any linked issues (`gh pr view`, `gh issue view`) and confirm the changes deliver what the PR claims. Partial coverage of an issue is acceptable when the PR documentation explicitly says so; unstated gaps or scope drift are blocking findings.
+2. **Review.** Default categorization: **blocking / functionally-impacting / suggestions / minor** (matches *Tone for GitHub Comments*). Output format and delivery channel (file, agent toolkit, `gh pr review`) decided per-PR — confirm before posting externally.
+3. **Fix.** Address every blocking and functionally-impacting finding this cycle. Update `llms/NOTES.md` to advance state.
+4. **Loop.** Repeat until a full pass yields zero blocking findings, then run the validation gate.
+
+### Pre-Merge Validation Gate
+
+Run from the worktree root, in order. Each must pass (or have a documented waiver) before the PR is ready:
+
+| # | Command | Validates |
```
cafalchio, line 117: most of the tests can be replaced to just "make pre-commit"

cafalchio (changes requested):

AGENTS.md
jonpspri: Well, if by most you mean 'ruff' and 'interrogate', yes :).

I removed them and added a 'make pre-commit' step.

AGENTS.md
jonpspri: Duly updated.

cafalchio (dismissed):
LGTM

cafalchio (dismissed):
LGTM

cafalchio (dismissed):

cafalchio (dismissed):

scripts/git/resolve-secrets-baseline-conflict.sh
```diff
@@ -0,0 +1,93 @@
...
+
+# Regenerate baseline to update line numbers for files changed in this commit only
+log_info "Regenerating baseline with detect-secrets for changed files only"
+# Get list of files changed in the commit being rebased
+DETECT_SECRETS_PATH=$(git diff-tree --no-commit-id --name-only -r REBASE_HEAD 2>/dev/null || echo "")
+
+if [ -n "$DETECT_SECRETS_PATH" ]; then
+    log_info "Scanning $(echo "$DETECT_SECRETS_PATH" | wc -l | tr -d ' ') changed file(s)"
+    # Scan only the changed files and update baseline
+    if make detect-secrets-scan
```
cafalchio: if make detect-secrets-scan; then
(bob found)

cafalchio (changes requested):

scripts/git/resolve-secrets-baseline-conflict.sh
jonpspri: Re-written

.dockerignore
```diff
@@ -130,11 +123,8 @@ cython_debug/
 #----------------------------------------------------------------------
 # 4. Virtual environments
 #----------------------------------------------------------------------
-.env
```
brian-hussey, line 133: This needs to be added back to the list of things to ignore, as something that potentially has locally stored secrets we never want this added to a Container accidentally - we can't guarantee what happens with an image after build.

Makefile
```diff
@@ -3742,6 +3748,85 @@ pre-commit: uv                     ## 🪄  Run pre-commit tool
...
+	elif ! grep -qF ".secrets.baseline merge=secrets-baseline" .gitattributes; then \
+		echo ".secrets.baseline merge=secrets-baseline" >> .gitattributes; \
+		echo "✅  Added merge driver entry to .gitattributes"; \
+	else \
+		echo "✅  .gitattributes already contains merge driver entry"; \
+	fi
+	@echo "✅  Git merge driver configured for .secrets.baseline"
+
+.PHONY: install-post-rewrite-hook
+install-post-rewrite-hook:     ## 🪝  Install post-rewrite hook that refreshes .secrets.baseline after rebase
```
brian-hussey: I think change the name here to be "install-post-rewrite-secrets-update-hook" or similar, without reading the comment it is not clear what is intended by this action, this at least makes it a bit clearer for anyone interacting from command line.

scripts/git/resolve-secrets-baseline-conflict.sh
brian-hussey: I don't follow what you're asking for or commenting on with this @cafalchio.

brian-hussey (changes requested):
A couple of things, but nothing major at the moment.

Makefile
brian-hussey: Corrected, resolving

.dockerignore
jonpspri: There's still an entry for .env in the file on line 275.  These removals are redundant entries.

brian-hussey (dismissed):
Thanks for making the changes @jonpspri, appoved.

cafalchio (dismissed):
LGTM 
Lets see it running.

msureshkumar88 (changes requested):
Thanks for this — a lot of genuinely useful cleanup here. The ADR group is the standout: ADR-0054 is well structured, correctly cites #5556 and ADR-0025/ADR-0024, and the index row plus `.pages` entry are both in place. Restoring ADR-0025 and flipping it to *Superseded* rather than rewriting it in place is exactly the right call, and codifying that rule in `docs/AGENTS.md` is a good addition. The `.secrets.baseline` merge driver is also solving a real, recurring pain, and installing it outside the worktree so it survives rebasing older history is a nice piece of thinking.

A few things I think need attention before this goes in. Verified against `51220da8`.

## Blocking

**1. The description and the final commit message don't match the diff.**

The PR body's "Git hooks setup" group describes wiring `scripts/setup-git-hooks.sh` into `make install-dev`, a version check, and a reorg under `scripts/git/`. Commit `51220da8`'s message additionally describes an `install-post-rewrite-hook` target with marker-comment idempotency and `post-rewrite.chain` foreign-hook chaining.

None of that is in the tree:

```
$ git ls-files | grep -iE 'setup-git-hooks|post-rewrite|gitops' | wc -l
0
$ grep -c 'post-rewrite' Makefile
0
```

`install-dev` also doesn't call `configure-git` — the new targets exist but nothing invokes them. The body likewise says `llms/NOTES.md` is gitignored, but `.gitignore` has no such entry, so following the new `AGENTS.md` recipe leaves a state file that `git add -A` will happily pick up.

Was some of this dropped in a rebase? Either way, could the body and that commit message be brought in line with what's actually shipping? Right now it's hard to review against a description that documents absent code.

**2. `Makefile:7594` — the audit gate returns the finding count as its exit code.**

```make
@count=$$(... | jq -r '.stats | .live + .unaudited + .audited_real') && exit $${count}
```

Shell exit statuses are mod 256, so:

```
$ bash -c 'exit 256'; echo $?   # 0
$ bash -c 'exit 300'; echo $?   # 44
```

A baseline with exactly 256 (or 512, …) unaudited/live findings reports **success**, and anything above 255 reports an arbitrary wrong code. On a secret-detection gate that's a fail-open path (CWE-393). If `jq` errors, `count` is empty and the recipe dies with `exit: numeric argument required` instead of something actionable — and unlike the driver, the target prints nothing explaining the failure.

The driver already does this correctly at `resolve-secrets-baseline-conflict.sh:104-107`; worth making the two identical:

```make
@count=$$(... | jq -r '.stats | .live + .unaudited + .audited_real'); \
	case "$${count}" in ''|*[!0-9]*) echo "❌ could not read audit stats"; exit 1;; esac; \
	if [ "$${count}" -gt 0 ]; then \
		echo "❌ $${count} unaudited/live finding(s) — run 'make detect-secrets-audit'"; exit 1; \
	fi
```

**3. Removing `lint-install-hooks` / `lint-pre-commit` / `lint-pre-push` breaks existing checkouts.**

The old `lint-install-hooks` wrote `.git/hooks/pre-commit` containing literally `make lint-pre-commit`, and a pre-push hook calling `make lint-pre-push`. Anyone who ran it still has those files. After this merges, every `git commit` on those machines fails with `No rule to make target 'lint-pre-commit'`, and there's nothing in the error pointing at the cause.

Two ways out, either is fine:
- keep thin shims for a release — `lint-pre-commit: lint-staged`, `lint-pre-push: lint-changed`, and have `lint-install-hooks` print a deprecation pointing at `make configure-git`; or
- keep the deletion and add a CHANGELOG "Removed" entry plus a `DEVELOPING.md` line telling affected devs to `rm .git/hooks/pre-commit .git/hooks/pre-push` and run `make configure-git`.

Worth noting the replacement also changes *what* runs on commit — the old hook ran `lint-staged`, the new one installs the `pre-commit` framework shim. Different checks; one sentence in the docs would help.

**4. `detect-secrets-scan` changes behaviour three ways, and the docs still describe the old one.**

Scope narrowed to `git diff $(GIT_DIFF_TARGET)`, output routed through the new jq merge, and it now exits non-zero on findings (previously report-only). Still describing the old behaviour: `SECURITY.md:141`, `DEVELOPING.md:41`, `docs/docs/development/release-management.md:327,1400`, `docs/docs/architecture/security-features.md:150`, and the gate table in `AGENTS.md:124`.

Also `GIT_DIFF_TARGET ?= main` (`Makefile:7531`) assumes a local `main` branch — a fresh clone that only has `origin/main` gets `fatal: ambiguous argument 'main'`. Something like a `git rev-parse --verify -q main` fallback to `origin/main` would make it robust.

One thing that would really help the reader: the PR doesn't mention that whole-tree coverage is **retained** by CI (`.github/workflows/pre-commit.yml:80` runs `make detect-secrets-hook` over `git ls-files` with `--fail-on-unaudited`). Without that, "scan only changed files" reads as a coverage regression. Worth stating explicitly.

## Functionally impacting

**5. The merge driver discards `%O` and `%B`.**

`resolve-secrets-baseline-conflict.sh` uses `$2` as the output path and never reads `$1` or `$3`; it regenerates from the worktree baseline plus a rescan.

`is_secret: true/false` verdicts are human input that a rescan can't re-derive. During a rebase "ours" is `onto` (i.e. main), so every verdict the branch author recorded for their own new files lives only in `%B` — and is dropped. Best case they re-audit on every conflicted rebase; worst case an entry previously marked `is_secret: true` disappears from the baseline with no signal.

Could the driver fail closed here — compare `%B`'s results against the merged output and abort if any `is_secret: true` entry would be lost? A one-line "+N / −M baseline entries" summary would also make an automated allowlist change auditable rather than invisible.

**6. `jq inside` is substring matching, not set membership.**

`Makefile:7580` and driver `:90`:

```jq
(.[0].results | to_entries | [ .[] | select( ([.key] | inside($gitfiles)) ) ] | from_entries)
```

```
$ jq -n '["a.py"] | inside(["a.py.bak"])'
true
```

So a deleted `config.py` survives the "drop no-longer-tracked files" filter as long as any tracked path *contains* that substring. `($gitfiles | INDEX(.)) as $set | ... | select($set[.key])` gives exact membership.

Related question while you're in there: since `.[1]` is a full copy of the baseline that `scan --update` seeded from, and the expression is `filtered(.[0]) + .[1].results`, does the pruning actually take effect — or does `.[1]` reintroduce the deleted-file entries? A small test (delete a file with a baseline entry, run the target, assert it's gone) would pin this down.

**7. The `.dockerignore` dedupe isn't fully behaviour-preserving.**

Most of it checks out — `.vscode/`, `.idea/`, `tmp/`, `media/`, `instance/`, `.env`, `target/`, `.github/`, `*.pyc`, `deployment/k8s/` are all still covered by a broader or duplicate rule elsewhere in the file. But this hunk changes semantics:

```diff
-LICENSE
-CHANGELOG
-AUTHORS
-CONTRIBUTORS
-TODO
-TODO.md
-DEVELOPING.md
-CONTRIBUTING.md
+!CHANGELOG.md
+!AUTHORS.md
+!DEVELOPING.md
+!CONTRIBUTING.md
```

Those were excludes; they're now negations, so `CHANGELOG.md`, `AUTHORS.md`, `DEVELOPING.md`, `CONTRIBUTING.md` and `LICENSE` now enter the build context where they previously didn't. `!mcp-catalog.yml` also went away — harmless today only because nothing else matches that path. If including them is intended, could the body say so rather than "behavior-preserving"?

**8. No tests, and CI never exercises the new code.**

CI only runs `detect-secrets-hook`, which is the unchanged path. Nothing executes the new jq pipeline, the exit gate, `configure-secrets-merge-driver`, or the driver itself.

Cheapest high-value step: a PR job running `make detect-secrets-scan GIT_DIFF_TARGET=origin/main`. Beyond that, `bats` coverage for the driver against throwaway repos would pay for itself — scope detection across all four branches (`rebase-merge` / `rebase-apply` / `MERGE_HEAD` / fallback), the deleted-file pruning question in #6, unaudited-finding → exit 1, `%B` audited entry preserved, and missing `jq`/`uv` producing actionable errors. An idempotency check on `configure-secrets-merge-driver` (run twice, assert one `.gitattributes` line) would be quick too.

## Suggestions

- **`docs/AGENTS.md:118` contradicts this PR's own change.** It says to add a nav entry in `mkdocs.yml`, but ADR nav lives in `docs/docs/architecture/adr/.pages` (awesome-pages) — which is what this PR correctly edited. `mkdocs.yml` has no ADR entries.
- **Apply the new ADR rule to ADR-0025.** The guidance added here mandates a `!!! warning` pointer note "as in ADR-038"; ADR-0025 got the front-material status lines but no pointer note.
- **`OUTPUT_TARGET` is dead.** Its comment says the merge driver overrides it, but `51220da8` deliberately made the driver standalone so it no longer calls make. Nothing sets it now.
- **Two hand-maintained copies of the same pipeline.** The driver and the `detect-secrets-scan` recipe have already diverged (finding #2 is fixed in one, not the other). The "must not invoke make" rationale is sound — but the Makefile could call the script, leaving the *installed* copy as the standalone snapshot. One implementation, same guarantee.
- **`jq` isn't a declared prerequisite** anywhere (only `uv` is), yet both paths hard-depend on it. Without it you get `jq: command not found` mid-recipe, or a failed merge. A `command -v jq` preflight with a useful message would help.
- **Driver logging** — progress `echo`s go to stdout and interleave with git's own merge output; stderr would read better.
- **`make configure-git` is undocumented** outside `make help`. Also worth a note that `.gitattributes` now ships `merge=secrets-baseline` to every clone while the driver is installed manually — git falls back to the default text merge with a warning, which is safe, but contributors will see that warning and wonder.
- **`configure-secrets-merge-driver` installs the checked-out branch's script** into `$(git rev-parse --git-common-dir)/git-drivers/` and registers it as a git-invoked executable that outlives the checkout. Low risk since you're already running that branch's Makefile, but installing from a trusted ref (`git show main:scripts/git/...`) or printing the script hash on install would tighten it.

## Minor

- `DETECT_SECRETS_FILES_EXCLUDE` gained `|^.secrets.baseline$` with unescaped dots — matches e.g. `Xsecretsxbaseline`. Cosmetic, but it's an exclusion rule: `^\.secrets\.baseline$`.
- `test-mcp-cli` lost its `[DEPRECATED]` help text and runtime warning. Collapsing it to a prerequisite alias is clean; dropping the deprecation signal feels like a separate decision.
- `.gitignore`: `todo/` → `todo/*` and the new `.omo` entry look unrelated to the dedupe rationale.

## One process thought

There are five independent groups here and they're at quite different readiness levels. The ADR group looks merge-ready as-is; the ignore/docs hygiene needs only #7; the secrets/git tooling needs the substantive work above. Splitting would let the first two land now rather than waiting on the third — entirely your call, though.

Happy to look again once the blocking items are addressed.

.dockerignore
jonpspri: There's another `.env` in the file -- this was removed because it's redundant.

Makefile
jonpspri: Done.

jonpspri:
Thanks for the thorough pass. Everything blocking and functionally-impacting is addressed in `5a358bc25` (squashed from the original four fixup commits), and the PR body has been rewritten to match the tree. Point by point:

## Blocking

1. **Description/messages vs diff** — yes, the hook work was dropped in a rebase while the messages still described it. Good catch. The `post-rewrite` hook and its install target are restored (target now `install-post-rewrite-detect-secrets-hook`), `llms/NOTES.md` is gitignored, and the body no longer claims `install-dev` wiring or `setup-git-hooks.sh` — it describes what actually ships.
2. **Audit gate exit code** — agreed, fail-open on a security gate. The gate now rejects empty/non-numeric counts with an actionable message and exits 1 explicitly (`5a358bc25`). Better: `detect-secrets-scan` is now a thin wrapper that execs the driver script, so there is one implementation of the scan/merge/gate pipeline — the two copies can't diverge again (this also takes your "hand-maintained copies" suggestion). `OUTPUT_TARGET` is gone.
3. **lint-\* removal** — went the CHANGELOG "Removed" + `DEVELOPING.md` remediation route (`5a358bc25`), including the note that the old hook ran `lint-staged` while the replacement installs the pre-commit framework shim.
4. **Docs + `GIT_DIFF_TARGET`** — all six sites updated, and the fallback to `origin/main` is in. The body and docs now state explicitly that whole-tree coverage is retained in CI via `detect-secrets-hook --fail-on-unaudited`; the release flow uses the retained whole-tree variant `detect-secrets-scan-all`.

## Functionally impacting

5. **Discarding %O/%B** — the driver now fails closed: after merging, any `%B` entry with `is_secret: true` whose `hashed_secret` is absent from the merged output under the same file key aborts the merge and lists what would be lost. There's also a `+N / -M baseline entries` summary vs `%A`. Failing-first evidence: with the old driver the verdict silently vanished (exit 0, entry gone); after the fix the same scenario exits 1 naming it.
6. **`jq inside`** — correct, substring containment. Replaced with `INDEX`-based exact membership in driver and hook. On your pruning question, I ran the experiment: `scan --update <baseline> <paths>` replaces results with only the scanned paths' findings, so `.[1]` does not reintroduce deleted-file entries and the merge structure is sound — the substring bug was the only defect. A bats test pins it (tracked `config.py.bak` no longer keeps a stale `config.py` entry alive).
7. **`.dockerignore`** — intentional; those files are meant to be in the build context. The body now says so rather than claiming pure dedupe.
8. **Tests/CI** — added `tests/bash/` (16 bats tests: all four scope branches, the gate, pruning, `%B` preservation, missing `jq`, and `configure-secrets-merge-driver` idempotency), runnable via `make test-bats`, which prints install guidance when bats is absent. I did not add a CI job in this round pending an upcoming repository reorganization.

## Suggestions

Taken: `OUTPUT_TARGET` removed, `jq` preflight in both scripts, driver/hook diagnostics to stderr, `docs/AGENTS.md` nav fix (`.pages`, not `mkdocs.yml`), ADR-025 pointer note, the Makefile-calls-script dedup, and `make configure-git` documented in `DEVELOPING.md` (including the versioned `.gitattributes` vs per-machine driver install warning). Also both minors: escaped dots in the exclusion regex, and the `test-mcp-cli` deprecation text is back.

Not taken: installing the driver from a trusted ref — I'd like to think through the workflows where installing the branch's version is intentional, so I've left that for a follow-up. On `.gitignore` `todo/*` / `.omo`: fair, they're adjacent housekeeping rather than dedupe — called out here for transparency.

On the process thought: kept it as one PR since the fixes turned out mostly small.

msureshkumar88 (changes requested):
Ran an independent e2e verification of this round's fixes — real `git`, real `jq`, and the real pinned `detect-secrets` fork over the network, no mocking. Nine scenarios, exercising real `git rebase` runs where git itself invokes the installed merge driver (the bats suite invokes the driver script directly, so this covers ground it doesn't):

**Confirmed fixed (7/9), matching the earlier review discussion:**
- Fail-closed abort works for real: a rebase that would drop an audited-real verdict is correctly blocked by an actual `git rebase`, not just the driver called standalone.
- `jq inside()` substring bug is gone; the shipped script's live code uses `INDEX()`-based exact membership (verified the reviewer's own `["a.py"] | inside(["a.py.bak"])` example against both the old and new expressions with real `jq`).
- The driver's own `origin/main` fallback works in a clone with no local `main`.
- `make detect-secrets-scan` (the thin wrapper) succeeds end-to-end on a clean change and fails closed end-to-end on a real unaudited secret.
- The post-rewrite hook installs and fires on a real rebase that actually rewrites commits.

**Two real gaps found, both reproducible on demand:**

1. **New file + `.secrets.baseline` conflict during a real rebase can spuriously abort, even with nothing actually lost.** When a commit both adds a new file and touches the baseline (the ordinary "add a file with an already-audited secret" case), during an actual `git rebase` the driver's rescan sometimes can't find the new file yet ("`detect-secrets: <file>: No such file or directory`") — it isn't materialized in the working tree at the moment git invokes the merge driver for `.secrets.baseline`. The fail-closed check then (correctly, given what it sees) reports the audited-real verdict as lost and aborts. It fails safe — no secret is silently dropped — but it means this common pattern will reliably require manual resolution instead of the intended auto-resolve, which cuts against the PR's main value proposition. Reproduced deterministically across separate throwaway repos (same file, same hashed_secret each time), so it isn't a one-off timing fluke.

2. **`make detect-secrets-scan` crashes (`fatal: ambiguous argument 'main'`, exit 128) in a checkout with neither a local `main` nor `origin/main`** — e.g. a CI job that checks out only the PR ref. The Makefile's `GIT_DIFF_TARGET ?= main` always passes a non-empty override to the driver, so the driver's own smarter fallback never gets a chance to kick in for this call path. Real git-triggered merges/rebases are unaffected (they never consult `GIT_DIFF_TARGET`), so this is narrowly scoped to direct `make detect-secrets-scan` invocations in that specific checkout shape.

Requesting changes on these two — happy to look again once addressed, everything else here looks solid.

jonpspri:
Both gaps addressed in `699eb7d9a`, TDD throughout — failing bats tests first (transcripts captured), then the fix, then green. Full suite is 23/23, and the new rebase tests exercise the real git-invoked driver path you used, not the standalone invocation.

## Gap 1 — spurious abort on add-a-file picks

Reproduced deterministically before fixing (`detect-secrets: secret.py: No such file or directory` mid-rebase, then the spurious `audited as real would be lost` abort). Two things fell out of the reproduction:

1. **Materialization source.** With git 2.55/ort, at driver-invocation time the index does *not* have the pick's file, and `REBASE_HEAD` isn't written yet (it's only created when the rebase *stops*). So the driver now tries, in order: the index, `REBASE_HEAD`, and the current `pick` sha recorded in the rebase state dir's `done` file — the last being what actually works. Materialized files are removed by the driver's `EXIT` trap; a `--abort` test asserts a clean `status --porcelain` afterward.
2. **Materialization alone wasn't enough.** With the file present, the fresh scan produced the entry as *unaudited* — the pick's `is_secret` verdict lived only in `%B`, and the gate would have stopped the rebase anyway, just for a different reason. So the `--update` seed is now the union of the worktree baseline and `%B`'s results (by `hashed_secret`, `%B` winning collisions so a downgraded verdict trips the gate rather than passing silently). `--update` then carries verdicts across for re-detected secrets; `%B`-only entries that aren't re-detected are still dropped, so your fail-closed check fires exactly when it should.

One honest deviation from the literal expectation: "rebase completes *and* retains `is_secret: true`" is self-contradictory under the gate you had us add in blocking #2 (`live + unaudited + audited_real`). The e2e tests split it into the two coherent cases: audited-false pick → auto-resolves and completes; audited-real pick → stops at the audit gate with the correct reason, never the spurious lost-verdict report.

## Gap 2 — exit 128 with no main ref

Both root causes fixed. The wrapper no longer overrides by default: it forwards `GIT_DIFF_TARGET`/`DETECT_SECRETS_PATH` only when the user actually set them (make `$(origin)` filter), and the parse-time `git diff` is silenced with `2>/dev/null`. And the driver's fallback now probes the repository's root branch rather than assuming `main`: local `main`/`master`/`develop`, then `refs/remotes/origin/HEAD`, then `origin/main|master|develop`, degrading to a whole-tree scan with a warning only when nothing matches. Same chain in the post-rewrite hook. Tests cover a `master`-only repo (diffs against `master`, distinguished from whole-tree by finding count), `origin/HEAD` resolution, the no-candidate whole-tree path, and the make-level passthrough both ways.

Everything from the previous round still holds (the 16 pre-existing bats tests stay green, including the fail-closed `%B` cases). Happy to look at anything else.

msureshkumar88 (dismissed):
Re-verified after the latest fix commit (2f62bb8d2).

Both issues from the prior review are resolved:

1. **New-file materialization during rebase**: the driver now restores scan-scope files missing from the worktree (from the index, then `REBASE_HEAD`, then the current pick) before scanning, removing them via the `EXIT` trap. Confirmed with a real `git rebase` where a pick adds a file and touches `.secrets.baseline` — the rebase now stops via the real audit gate (`unaudited/live finding(s)...`) instead of the previous spurious "verdict would be lost" report, and `git rebase --abort` leaves a clean tree.
2. **`make detect-secrets-scan` crash with no local `main`**: fixed via the extended fallback chain (local main/master/develop → `origin/HEAD` → `origin/main|master|develop` → whole-tree scan with a warning). Confirmed on a fresh single-branch clone lacking a local `main` ref — no more "ambiguous argument" crash.

Verification run:
- `bats tests/bash/resolve_secrets_baseline_conflict.bats tests/bash/secrets_baseline_rebase_e2e.bats` — 22/22 pass, including the two new e2e-style cases (audited-false auto-resolve, audited-real gate-stop with the corrected message).
- Independent e2e script (real `git rebase`, real `jq`, real pinned `detect-secrets` fork over the network, no mocking) — 9/9 pass, covering both bug repros plus the jq membership fix, the no-default-branch fallback, the Makefile wrapper end-to-end, and the post-rewrite hook firing on a real rebase.

No regressions found. Looks good to merge from my end.

cafalchio (dismissed):
We need to see it working on the CI.

msureshkumar88 (approved):
LGMT

brian-hussey (approved):
LGTM

PR #4458, opened 2026-04-27
fix: add rust coverage reports

lucarlig:
## Summary

- Update `make rust-coverage` to generate HTML, Cobertura XML, and the terminal coverage table.
- Print the generated Rust coverage report locations in the target output.
- Add `make rust-diff-cover` for changed-line Rust coverage using `coverage/cobertura.xml`.
- Add a small Makefile regression test for the Rust coverage targets.


Lang-Akshay (approved):
Thanks for the PR @lucarlig . Both target work as expected. 

Nice improvement 
- `make rust-coverage` 
- `make rust-diff-cover`

Issue #2692, opened 2026-02-04
[BUG][PERFORMANCE]: auth hot-path DB queries dominate request latency under load


crivetimihai:
## Summary

py-spy profiling of gateway workers under 4000 concurrent users reveals that **71% of request-serving DB time** (526/740 samples) is spent in the authentication chain before any business logic executes. Several auth helper functions open independent `fresh_db_session()` connections with zero caching, creating 5–8 DB round-trips per request.

Peak observed: **~900 RPS average** with 3 gateway replicas × ~25 gunicorn workers. PgBouncer at 90% CPU under load.

## Profiling Results

py-spy, 15 s sampling, 22 744 total samples across gateway workers:

| Function | Samples | File | Issue |
|----------|---------|------|-------|
| `aggregate_all_components` | 361 | `log_aggregator.py` | Background full-table scan |
| `_is_api_token_jti_sync` | 221 | `auth.py:325` | Zero caching, `fresh_db_session` per request |
| `_cleanup_table` | 88 | `metrics_cleanup_service.py` | Background batch DELETE competing with requests |
| `get_token_revocation` | 73 | `token_catalog_service.py:984` | No caching despite `auth_cache` existing |
| `get_user_teams` | 72 | `auth_cache.py` | Cache stores only team IDs, not full objects |
| `get_user_by_email` | 45 | `auth.py:355` | No caching at service level |
| `_check_resource_team_ownership` | 43 | `token_scoping.py:478` | Creates own `fresh_db_session` per request |
| `_check_token_revoked_sync` | 36 | `auth.py:258` | Partially cached via `auth_cache` |

## Infrastructure Observations

- **Double pooling**: SQLAlchemy QueuePool (`pool_size=20, max_overflow=10`) on top of PgBouncer (transaction mode) — connection checkout goes through two pool layers
- **`pool_pre_ping=true`**: Adds a `SELECT 1` liveness ping on every connection checkout, doubling round-trips for short-lived sessions
- **PgBouncer at 90% CPU**: Bottleneck under load with 3 replicas × ~25 workers ≈ 75 potential DB clients
- **Background tasks** (`aggregate_all_components`, `_cleanup_table`) consume ~13% of DB capacity while running concurrently with request processing
- **Streaming DB session leak**: `chat_completion_stream` in `llm_proxy_service.py` holds the DB session for the entire duration of the HTTP stream instead of releasing it after model resolution

### 1. Cache `_is_api_token_jti_sync` (auth.py)
[...]

PR #4468, opened 2026-04-27
fix(metrics): throttle aggregate_all_components with pg advisory lock and configurable interval

### LINKED ISSUE #4467: [PERFORMANCE]: Throttle background aggregate_all_components — advisory lock + configurable interval
## Parent issue

Tracks the P1 follow-up item from #2692:

> **Throttle background `aggregate_all_components`** — 361 samples on a single worker, full-table scan. Run less frequently or use SKIP LOCKED.

## Problem

Every gunicorn worker starts its own independent `run_log_aggregation_loop` task. With 3 replicas × 25 workers = **75 workers all performing a full-table scan on `structured_log_entries` simultaneously**, every 5 minutes. No inter-worker coordination exists, so every window fires 75 redundant aggregations — contributing ~13% of observed DB capacity under load.

## Solution

Two complementary fixes:

### 1. PostgreSQL advisory lock (`pg_try_advisory_lock`)

`aggregate_all_components` attempts a **non-blocking** session-level advisory lock before running. If another worker already holds it, the function returns `[]` immediately without touching the database. The lock is released in a `finally` block after all commits complete.

Session-level lock (not transaction-level) is required because `_upsert_metric` commits inside the loop — a transaction-level lock would release on the first `db.commit()`, allowing concurrent workers back in.

Behaviour by caller type:

| Caller | Lock attempted? |
|---|---|
| Background loop (`db=None`) | Yes — `should_close=True` path |
| `backfill`, `log_search.py` (pass `db=`) | No — existing behaviour preserved |

Fail-open: if the advisory lock SQL itself raises (e.g. connection error), aggregation proceeds without the lock rather than silently suppressing it.

SQLite (dev/test): no-op — `_is_postgresql()` short-circuits both acquire and release.

Lock ID `6283185307179` chosen to avoid collision with `bootstrap_db.py` (`42424242424242`).

### 2. Configurable run interval (`METRICS_AGGREGATION_INTERVAL_SECONDS`)

New optional env var decouples run frequency from window size. Defaults to `metrics_aggregation_window_minutes * 60` (existing behaviour). Operators can raise this independently (e.g. `900`) to reduce background DB pressure on multi-worker deployments without changing the aggregation window semantics.

## Files changed

| File | Change |
|---|---|
| `mcpgateway/services/log_aggregator.py` | `_AGGREGATE_LOCK_ID` constant; `_try_aggregate_pg_lock`; `_release_aggregate_pg_lock`; lock guard in `aggregate_all_components` |
| `mcpgateway/config.py` | `metrics_aggregation_interval_seconds: Optional[int]` field |
| `mcpgateway/main.py` | Read new config field for loop interval |
| `.env.example` | Document `METRICS_AGGREGATION_INTERVAL_SECONDS` |
| `tests/unit/mcpgateway/services/test_log_aggregator.py` | 12 new tests; 98% coverage on new code |

## Testing

- 12 new unit tests covering: lock skip, lock acquire, release on success, release on exception, SQLite no-ops, fail-open on DB error, `db=` passthrough, scalar True/False paths
- All 69 existing tests continue to pass
- Bandit B110/B105 warnings resolved

msureshkumar88:
## Summary

Closes #4467
Part of #2692 (P1 follow-up: throttle background `aggregate_all_components`)

- **Advisory lock**: `aggregate_all_components` now attempts `pg_try_advisory_lock` before running when called from the background loop. If another worker holds the lock, the function returns `[]` immediately — eliminating 74 of 75 redundant full-table scans per 5-minute window on a typical 3-replica × 25-worker deployment
- **Configurable interval**: new `METRICS_AGGREGATION_INTERVAL_SECONDS` env var decouples run frequency from window size; defaults to `window_minutes × 60` (no behaviour change unless set)
- **Bandit fixes**: resolved B110 `try/except/pass` in `log_aggregator.py` and `audit_trail_service.py`; B105 false-positive in `oauth_manager.py`

## Design notes

- Session-level lock (`pg_try_advisory_lock`) not transaction-level (`pg_try_advisory_xact_lock`) — `_upsert_metric` commits inside the loop, which would auto-release a transaction-level lock and allow concurrent workers back in
- Fail-open: if the lock SQL raises, aggregation proceeds without the lock rather than suppressing it silently
- Lock only attempted when `should_close=True` (background loop path); callers that pass `db=` (backfill, log_search) are unaffected
- Lock ID `6283185307179` chosen to avoid collision with `bootstrap_db.py` (`42424242424242`)

## SQLite (local dev) compatibility

SQLite is fully supported and completely unchanged by this PR:

- `_is_postgresql()` returns `False` for SQLite — both `_try_aggregate_pg_lock` and `_release_aggregate_pg_lock` return immediately without executing any SQL, so `lock_acquired = True` unconditionally
- Single-process `make dev` (uvicorn `--reload`) runs one worker anyway, so there is nothing to coordinate between workers
- `METRICS_AGGREGATION_INTERVAL_SECONDS` works identically on SQLite — it controls how often the loop sleeps between runs
- No config changes needed for local dev; the SQLite path is fully transparent

## Manual testing

Start the stack with PostgreSQL, multiple workers, a short interval, and DEBUG logging:

```bash
# .env
METRICS_AGGREGATION_ENABLED=true
METRICS_AGGREGATION_AUTO_START=true
METRICS_AGGREGATION_INTERVAL_SECONDS=15
LOG_LEVEL=DEBUG
```

```bash
GUNICORN_WORKERS=4 make serve
```

Tail the logs and filter for aggregation activity:

```bash
docker compose logs -f | grep -E "aggregate|advisory|lock|Skipping"
```

**Expected behaviour after this fix** — per interval, exactly one worker runs the aggregation; all others immediately log:

```
Skipping aggregate_all_components: lock held by another worker
```

**Before this fix**, all 4 workers would run the full aggregation simultaneously every 15 seconds with no coordination.

## Changes

| File | What changed |
|---|---|
| `mcpgateway/services/log_aggregator.py` | `_AGGREGATE_LOCK_ID`; `_try_aggregate_pg_lock`; `_release_aggregate_pg_lock`; lock guard in `aggregate_all_components` |
| `mcpgateway/config.py` | `metrics_aggregation_interval_seconds: Optional[int]` |
| `mcpgateway/main.py` | Use new config field for loop `interval_seconds` |
| `.env.example` | Document `METRICS_AGGREGATION_INTERVAL_SECONDS` |
| `tests/unit/.../test_log_aggregator.py` | 12 new tests; 98% coverage on new code; all 69 pass |
| `mcpgateway/services/audit_trail_service.py` | B110: log debug instead of bare `pass` |
| `mcpgateway/services/oauth_manager.py` | B105: `nosec` annotation with explanation |


Lang-Akshay (changes requested):
Thanks for the PR @msureshkumar88 . 
Please implement following suggested changes 
The advisory lock is a background coordination primitive — not a security gate — so there are no auth-bypass risks. The only concerns are operational reliability issues with the lock implementation pattern.

| # | File                                       | Line     | Severity | CWE     | Description                                                                                                                                                                           |
| - | ------------------------------------------ | -------- | -------- | ------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| 1 | mcpgateway/services/log_aggregator.py      | 46, 58   | Low      | CWE-89  | F-string interpolation of integer constant in raw SQL; not exploitable (constant int, no user input path), but sets a poor copy-paste precedent                                       |
| 2 | mcpgateway/services/log_aggregator.py      | 48–50   | Low      | CWE-703 | Fail-open on advisory lock acquire error: DB errors silently allow un-coordinated parallel execution; operational concern, not a security risk                                        |
| 3 | mcpgateway/services/log_aggregator.py      | 58–60   | Low      | CWE-667 | Session-level lock not guaranteed released on silent `pg_advisory_unlock` failure; pooled connection may retain lock, leading to aggregation starvation in multi-worker deployments |
| 4 | mcpgateway/services/audit_trail_service.py | 171–172 | Info     | CWE-532 | Exception message logged at DEBUG; no production risk at default `LOG_LEVEL=ERROR`                                                                                                  |
| 5 | mcpgateway/services/oauth_manager.py       | 663      | Info     | —      | `# nosec B105` annotation is correctly justified; value is an RFC 8693 token type URI, not a credential                                                                             |
| 6 | .secrets.baseline                          | 4865     | Info     | —      | Line-number drift from new config field insertion;`is_secret: false` marking remains correct                                                                                        |

**Remediation suggestions:**

1. **F-string in raw SQL (Finding 1):** Use `bindparams` to eliminate ambiguity:

   ```python
   result = db.execute(text("SELECT pg_try_advisory_lock(:lock_id)").bindparams(lock_id=_AGGREGATE_LOCK_ID))
   db.execute(text("SELECT pg_advisory_unlock(:lock_id)").bindparams(lock_id=_AGGREGATE_LOCK_ID))
   ```
2. **Session-level lock starvation (Finding 3):** Document in `_release_aggregate_pg_lock` that if `pg_advisory_unlock` fails silently on a pooled connection, the session-level lock persists until that connection is physically closed.

## Redundant Code

| # | File               | Line(s)    | Type                 | Description                                                                                                               | Suggestion                                                     |
| - | ------------------ | ---------- | -------------------- | ------------------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------------- |
| 1 | mcpgateway/main.py | 1605–1608 | Defensive code smell | `getattr(settings, "metrics_aggregation_interval_seconds", None)` accesses a field always present on the Pydantic model | Use `settings.metrics_aggregation_interval_seconds` directly |

msureshkumar88:
Thanks for the detailed review @Lang-Akshay — all addressed in the latest commit:

**Finding 1 (CWE-89) — F-string in raw SQL:** Both `_try_aggregate_pg_lock` and `_release_aggregate_pg_lock` now use `text(...).bindparams(lock_id=_AGGREGATE_LOCK_ID)` instead of f-string interpolation.

**Finding 3 (CWE-667) — Session-level lock starvation:** Added a note to the `_release_aggregate_pg_lock` docstring explaining that if `pg_advisory_unlock` fails silently on a pooled connection, the session-level lock persists until the connection is physically closed, which can delay the next aggregation run by one interval.

**Redundant code # 1 — `getattr` in `main.py`:** Replaced `getattr(settings, "metrics_aggregation_interval_seconds", None)` with direct `settings.metrics_aggregation_interval_seconds` access.

All 69 unit tests still pass.

Lang-Akshay (dismissed):
Nice work on this one. The advisory lock design is solid — session-level (not transaction-level) is the right call given _upsert_metric commits inside the loop, and the fail-open + SQLite no-op paths are handled cleanly. Test coverage is thorough with all lock paths covered.

Two minor non-blocking suggestions:

- main.py L1606: getattr(settings, "metrics_aggregation_interval_seconds", None) could be simplified to settings.metrics_aggregation_interval_seconds since the field is explicitly declared on Settings
- Worth adding a quick Pydantic validation test asserting metrics_aggregation_interval_seconds=0 raises ValidationError to pin the ge=1 constraint

### Neither is a blocker. ✅ Approving.

Lang-Akshay (dismissed):

brian-hussey:
@Lang-Akshay Moving back to be reviewed following additional changes since your approval.
Please review subsequent changes and see if we're still good to merge.

ja8zyjits:
## Minor Suggestion (Non-blocking)

**Pydantic validation test** (Lang-Akshay):
```python
def test_metrics_aggregation_interval_seconds_validation():
    """Verify ge=1 constraint on metrics_aggregation_interval_seconds."""
    with pytest.raises(ValidationError):
        Settings(metrics_aggregation_interval_seconds=0)
```


**Post-merge:** Consider adding Pydantic validation test in follow-up.

ja8zyjits (approved):
LGTM

PR #4469, opened 2026-04-27
fix: Compose: harden gateway service (no-new-privileges, cap_drop, read_only, tmpfs)

prakhar-singh1928:
## 🔗 Related Issue
Closes #https://github.ibm.com/contextforge-org/mcp-context-forge/issues/120

---

## 📝 Summary
Implements comprehensive container security hardening to address all vulnerabilities identified during penetration testing. This PR adds defense-in-depth security measures following CIS Docker Benchmark best practices.

1. **NoNewPrivileges** - Prevents privilege escalation via setuid/sudo
2. **ReadOnlyRootFilesystem** - Enabled with tmpfs mounts for required writable directories
3. **AppArmor/SELinux** - Mandatory Access Control (MAC) profile enabled
4. **Privileged Mode** - Explicitly disabled to maintain container isolation
5.  Added a make target to verify if the containers start up with added security

---

## 🏷️ Type of Change
- [ ] Bug fix
- [X] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

---

## ✅ Checklist

---

## 📓 Notes (optional)


infra/security/seccomp-profile.json
brian-hussey, line 1: How does this file get used in the updated procedures, or was it an unintentional addition?

brian-hussey (commented):
One question, I'm not sure if that seccomp-profile.json is used anywhere or not at this point. Once we've got clarity I'll continue testing.

jonpspri:
Thanks for picking this up — `no-new-privileges`, `cap_drop: ALL`, and `read_only: true` are the right direction, and getting the gateway service onto a least-privilege footing is real defense-in-depth. A few notes below, categorized so it's clear what's blocking vs. nice-to-have. Happy to chat through any of these.

## Blocking concerns

### 1. The UID/GID change probably belongs in its own PR

`Containerfile.lite` flips runtime user from `1001:0` (the OpenShift-friendly random-UID + group-0 pattern) to `10001:10001`. That's an **image-wide** change affecting PyPI, single `docker run`, compose, Helm, and OpenShift simultaneously, but only compose is being adapted here.

Specifically:
- `chmod -R g=u /app` (line 301) no longer serves its previous purpose. It enabled OpenShift's random-UID + supplemental-group-0 model when group was `0`. With group `10001`, only members of group 10001 benefit — which on OpenShift `restricted-v2` is nobody.
- `charts/mcp-stack/templates/deployment-mcpgateway.yaml:42-52` has `runAsNonRoot: true`, `seccompProfile: RuntimeDefault`, and `capabilities.drop: ALL`, but no `runAsUser`/`runAsGroup`/`fsGroup`. It currently relies on the image's `USER` directive to pin the UID.
- `docs/docs/deployment/openshift.md`, `docs/docs/deployment/kubernetes.md`, `docs/docs/deployment/tls-configuration.md`, and `docs/docs/architecture/security-features.md` still document `1001` and `chgrp 0`.
- Builder stages in `Containerfile.lite:143-155` still `chown 1001:0` for the frontend builder — the migration is inconsistent inside the Dockerfile itself.
- Existing host bind mounts holding files owned by `1001:0` will break silently for users upgrading.

A couple of approaches worth considering:
- **(a)** Split the UID change into a separate, well-communicated PR with migration notes and Helm/OpenShift compatibility validated; revert that part here.
- **(b)** Keep it, but in this PR also add `runAsUser: 10001`/`fsGroup: 10001` to the chart (and a values knob), update the OpenShift/Kubernetes/TLS docs, and clean up the builder-stage `1001:0` references.

### 2. `cap_add` reads as if it wasn't fully verified at runtime

The container binds to `4444` (non-privileged), runs as `10001:10001`, and doesn't switch users at runtime. With that setup:
- `NET_BIND_SERVICE` is only needed for ports < 1024.
- `CHOWN`, `SETGID`, `SETUID` look like leftovers from build-time needs, not runtime needs.

In a hardening PR specifically, four named caps after `drop: ALL` reads like "we didn't verify what was actually required." Worth trying `cap_add: []` (or removing the block entirely) and letting CI prove it. If something legitimately needs one of these, a one-line comment beside the cap explaining why goes a long way.

### 3. CI doesn't actually exercise the hardened runtime

The green checks here are `Build amd64/arm64/s390x/ppc64le`, `Container Smoke (a2a-echo-agent)`, `Container Smoke (python-sandbox)`, `Security Scan`, `Pre-commit`, `Rust-enabled container smoke`. None of those bring up `docker compose up` with the gateway service hardened (`read_only:true` + `cap_drop: ALL` + custom seccomp + tmpfs), hit `/health`, and exercise an authenticated `/tools` request path.

Image-build success isn't evidence the hardened runtime works. It would be worth adding a Make target (e.g. `make compose-up-hardened` + `curl /health` + a token-auth call) so the PR proves itself before merge.

For what it's worth — I did do a write-path audit and didn't find anything in the codebase that would unconditionally fail under `read_only:true` with the three current tmpfs mounts (Postgres URL is set in compose, `LOG_TO_FILE=false` by default, no `PROMETHEUS_MULTIPROC_DIR`, runtime scripts already use `/tmp` for lock files). So it should work — but "should" isn't proof.

## Strongly recommended

### 4. Re: @brian-hussey's question on `seccomp-profile.json` — it's only wired into compose, and the profile itself probably should be `runtime/default`

To answer Brian's question directly: the file is referenced **only** in `docker-compose.yml:753`. It's not used by Helm (which already has `seccompProfile: RuntimeDefault`) or by single-container `docker run` examples.

After comparing the profile against the upstream Moby default:
- It's roughly 95% overlap with Moby default but has at least one regression: `io_uring_enter`, `io_uring_register`, `io_uring_setup` are explicitly allowed (`infra/security/seccomp-profile.json:148-150`). Docker/containerd intentionally block `io_uring_*` because of the container-escape history (see [moby/moby#46762](https://github.com/moby/moby/pull/46762)).
- `clone` and `clone3` are unconditionally allowed (lines 45-46). Docker's default applies argument-level filtering on `clone` to prevent unprivileged namespace creation. We're losing that filter.
- `archMap` covers x86_64/aarch64/s390x/ppc64le but lacks the architecture-specific syscalls Moby includes (e.g., `s390_pci_mmio_*`, `riscv_flush_icache`), so it isn't an arch-equivalent superset of `runtime/default`.

A couple of options:
- **Preferred**: replace `seccomp=infra/security/seccomp-profile.json` with `seccomp=runtime/default` in compose and delete the JSON. That matches what Helm already does and gives ~95% of the security value with zero drift cost.
- **Alternative**: keep the file, but drop `io_uring_*`, restore Docker's `clone`/`clone3` arg filters, and add an `infra/security/README.md` explaining the maintenance commitment and how to test profile updates.

### 5. The PR title and scope don't match

The body says "comprehensive container security hardening" but the change touches only the `gateway` service in one compose file. The same compose ships:
- `nginx` (the actual public ingress on `:8080`) — unhardened
- `postgres`, `redis`, `pgbouncer`, `keycloak` — unhardened
- 6 monitoring services (prometheus/grafana/loki/tempo/promtail/cadvisor) — unhardened
- 11 testing/registration/A2A/benchmark services — unhardened
- `nginx_tls`, `pgadmin`, `redis_commander` — unhardened

Helm, single `docker run`, and PyPI paths are also untouched.

This isn't a problem per se — staged rollouts are fine. But the framing might be worth narrowing to something like **"Compose: harden gateway service (no-new-privileges, cap_drop, read_only, tmpfs)"**, with follow-up issues for the other services. nginx in particular is arguably higher priority than gateway since it's the actual public attack surface.

## Minor notes / follow-up (not blocking)

- **`mcp-catalog.yml` mount**: already `:ro`, good. Same for `${PLUGINS_DIR:-./plugins}`. Worth noting in a comment that these are intentionally read-only because the gateway runs read-only-root.
- **Documentation drift** if the UID change stays: `docs/docs/deployment/openshift.md:100-127, 224`, `docs/docs/deployment/kubernetes.md:259-263, 348-353` (still hardcodes `runAsUser: 1001`), `docs/docs/deployment/tls-configuration.md:707-718, 812-819` (advises `chgrp 0`), `docs/docs/architecture/security-features.md:105`.
- **SQLite default trap**: `mcpgateway/config.py` defaults to `sqlite:///./mcp.db`. Anyone whose env falls back to that default with `read_only:true` will hit a startup failure that's confusing because the cause is in compose, not their config. The current compose explicitly sets Postgres so it's masked, but worth a comment.
- **tmpfs sizing**: 100M `/tmp` is fine for typical traffic but could be tight for heavy support-bundle generation (`mcpgateway/services/support_bundle_service.py` defaults to `/tmp`). Worth surfacing as an env override.
- **Staging profile**: a parallel `seccomp-profile.log.json` with `SCMP_ACT_LOG` would help discover misses without breaking traffic.

---

Net: the direction is good, and most of the concerns above are about **scope and atomicity** rather than the security flags themselves. Splitting the UID change out and either dropping or tightening the custom seccomp profile would give you a much cleaner PR that's easy to review and easy to roll back if anything breaks.

infra/security/seccomp-profile.json
prakhar-singh1928: The file is referenced in `docker-compose.yml` 
This was not an unintentional addition. However following jonathons comment i am reverting that.

brian-hussey (dismissed):
LGTM

brian-hussey (dismissed):
LGTM

brian-hussey (approved):

Issue #4384, opened 2026-04-22
[BUG][AUTH]: OAuth resource field missing from UI causes audience validation failures with real OAuth providers


MohanLaksh:
## Summary

The OAuth token audience validation added in PR #3941 fails for real-world OAuth providers (Salesforce, Azure AD, etc.) because:

1. The `resource` field in `oauth_config` is not exposed in the Admin UI
2. The fallback behavior uses the full `gateway.url` (including path) as the expected audience
3. Most OAuth providers issue tokens with audiences at the origin level, not the full endpoint path

This causes legitimate OAuth flows to fail with "Token audience mismatch" errors that users cannot fix via the UI.

**Severity**: HIGH - Regression bug that breaks OAuth functionality for common enterprise IdPs

---

### Timeline

1. **March 20, 2026**: Documentation (`oauth-design.md:131`) explicitly documented as "Known Gap" that audience/scope validation did not exist
2. **April 6, 2026**: PR #3941 added token validation to fix security bug #3563
3. **Current state**: Validation exists but has implementation flaws that break real OAuth flows

### Cause → Effect Chain

```
CAUSE 1: PR #3941 added audience validation
    ↓
EFFECT 1: Code now validates token aud claim against expected audience
    ↓
CAUSE 2: Expected audience falls back to gateway.url when resource not configured
    ↓  (token_validation_service.py:155)
    |  expected_audience = oauth_config.get("resource") or gateway_url
    ↓
CAUSE 3: resource field is NOT exposed in Admin UI
    ↓  (admin.html:5689-5967 - OAuth form has no oauth_resource field)
    ↓
EFFECT 2: All OAuth configurations use fallback behavior
    ↓
CAUSE 4: gateway.url contains full endpoint path
    ↓  Example: "https://api.salesforce.com/platform/mcp/v1/platform/sobject-all"
    ↓
EFFECT 3: Expected audience is overly specific
    ↓
CAUSE 5: Real OAuth providers issue tokens with origin-level audiences
    ↓  Example token aud: "https://api.salesforce.com"
    ↓
EFFECT 4: Audience mismatch → blocking error
    ↓  (gateway_service.py:1598-1601)
    |  blocking = token_validation.blocking_errors
    |  if blocking:
    |      raise GatewayConnectionError(...)
    ↓
FINAL EFFECT: OAuth authorization succeeds but tool fetch fails
    ✗ Users see: "Refusing to forward OAuth token... Token audience mismatch"
[...]

PR #4476, opened 2026-04-27
[FEATURE]: OAuth Resource UI field, origin fallback, and softened auto-derived validation

jonpspri:
## Summary

Implements the maintainer-recommended **Option-2 hybrid** approach for issue #4384, plus complementary hardening.  Stacked on top of #4475 (the bugfix corrections, merged); docs follow in #4477.

### Feature: closes the remaining #4384 acceptance criteria

- **UI Resource (Audience) field** in the gateway OAuth configuration forms (create + edit) in ``mcpgateway/templates/admin.html``.  Form parsing is centralized in ``_assemble_oauth_config_from_fields`` (``mcpgateway/admin.py``), which merges ``oauth_resource`` into ``oauth_config["resource"]`` via ``parse_oauth_resource_form`` (single URI → ``str``, comma/whitespace-separated list → ``list[str]``, per RFC 7519 §4.1.3 shapes; URIs containing literal commas are preserved).  Pre-population on the edit dialog in ``mcpgateway/admin_ui/gateways.js``.
  - *Scope note:* the A2A agent forms briefly carried the same field, but review found no A2A code path consumes ``oauth_config["resource"]`` (no per-user token storage, refresh, or audience validation on the A2A path), so the inert field was removed.  Proper A2A audience validation is tracked in #5815.
- **Origin-extraction fallback** (``derive_resource_origin`` in ``mcpgateway/utils/oauth_resource.py``): when no resource is configured, derive ``scheme://netloc`` from ``gateway.url`` (instead of full path).  Real OAuth providers (Salesforce, Azure AD, Okta) issue origin-level audiences.  One shared implementation serves all four call sites that must agree on the value: authorization request, callback token exchange, token refresh, and validator fallback.  (Standards note: RFC 8707 does not specify default-resource derivation — this is a documented heuristic matching the MCP spec's canonical-URI convention, not an RFC-mandated behavior.)
- **Softened validation for auto-derived resources**: ``TokenValidationResult.audience_source`` records where the expected audience came from (``configured`` / ``learned`` / ``derived``).  A mismatch against a *derived* audience is advisory and stays out of ``blocking_errors`` — the upstream MCP server remains the authoritative validator, per RFC 9700 §2.3 and the MCP authorization spec (audience validation is the resource server's MUST).  Configured/learned mismatches remain blocking; ``OAUTH_REQUIRE_CONFIGURED_RESOURCE=true`` upgrades derived mismatches to blocking for strict deployments.

### Hardening: defense-in-depth around audience learning

- **Per-user learned audience**: the IdP's actual ``aud`` is persisted on the user's own ``OAuthToken.learned_aud`` row (migration ``c9f8e7d6a4b3``), never on shared ``gateway.oauth_config``.  This structurally prevents (a) cross-tenant DoS when one gateway serves multiple IdP tenants with per-tenant ``aud`` values, and (b) RBAC bypass where a user without ``gateways.update`` could mutate shared gateway config via the OAuth callback path.  Re-authentication refreshes the user's own row.
- **Issuer pinning at learning time**: when ``oauth_config["issuer"]`` is configured, the learned audience is persisted only if the token's ``iss`` matches (trailing slashes normalized).  A stale or misrouted token from a different AS cannot inject an audience for the wrong IdP.
- **Shape validation** (``_coerce_aud_claim`` / ``_coerce_iss_claim``): rejects malformed ``aud``/``iss`` values (numbers, empty containers, mixed-type lists) so a malformed IdP response cannot clobber previously-learned state.
- **Trust-model documentation**: the shared unverified JWT decoder ``_decode_token_claims_unverified`` documents *why* signatures aren't verified at learning time and where the actual security boundary lies (TLS to the admin-configured token endpoint; authorization-relevant validation happens at the resource server).
- **Single extraction API**: ``_extract_aud_and_iss`` performs one decode per callback for both claims; the redundant single-claim wrappers were removed.

### Refactoring (review-driven)

- ``mcpgateway/utils/oauth_resource.py`` consolidates all RFC 8707 resource logic: ``derive_resource_origin``, ``normalize_resource`` (fragment/query policy), and ``parse_oauth_resource_form``.  Eliminates the prior private cross-module imports (a mid-file ``noqa: E402`` import in ``oauth_router`` and a function-level import in the refresh path).
- ``_assemble_oauth_config_from_fields`` in ``admin.py`` replaces four near-identical ~45-line OAuth form-parsing blocks (gateway create/edit, A2A create/edit) — the duplication that caused the round-1 review miss.  ``encrypt_secret`` and ``include_resource`` parameters capture the only real differences between sites.
- ``_resolve_refresh_resource`` extracts the refresh-path resource policy from the 230-line ``_refresh_access_token``, with doctests for configured / origin-derived / URN / list cases.
- ``audience_source`` enum replaces the post-hoc ``audience_authoritative`` flag; the blocking rule is now a single expression in ``blocking_errors`` with no mutation on the mismatch path.

### Tests

- ``TestPersistLearnedAudience``-era coverage now targets the per-user model: store/refresh semantics, re-auth updates, opaque-audience round trips, multi-user regression.
- 6 new ``TestIssuerPinningOnAudienceLearning`` tests (no issuer, match, trailing-slash, mismatch, missing ``iss``, opaque token).
- Strict/advisory split tests including ``OAUTH_REQUIRE_CONFIGURED_RESOURCE`` enforcement at the real ``gateway_service.py`` blocking point.
- Refresh-path origin derivation for path-bearing gateway URLs (reviewer finding).
- JS tests pin the gateway edit-dialog Resource pre-population (string and list shapes).

## Stack

| | Branch | Status |
|---|---|---|
| 1 | ``bugfix/oauth-aud-resource-narrow`` | #4475 — merged |
| **2** | ``feat/oauth-aud-resource-hybrid`` | **this PR** |
| 3 | ``docs/oauth-audience-handling`` | #4477 |

## Acceptance criteria from #4384

| AC | Status |
|---|---|
| OAuth Resource field visible in Admin UI | ✅ this PR (gateway forms; A2A deferred to #5815 — no A2A consumer exists) |
| Auto-derives origin if not set | ✅ this PR |
| Explicitly configured resource blocks on mismatch | ✅ this PR |
| Auto-derived resource logs warnings but doesn't block | ✅ this PR (advisory unless ``OAUTH_REQUIRE_CONFIGURED_RESOURCE=true``) |
| Existing gateways without resource field continue working | ✅ this PR |
| Salesforce OAuth flow succeeds without manual config | ✅ this PR (origin fallback) |
| Documentation updates | follow-up PR (#4477) |
| Playwright UI tests | not in scope |

## Verification

- 2094 unit tests pass across the affected modules (admin, oauth_router, oauth_manager, token_storage_service, token_validation_service, gateway_service)
- 585 OAuth-scoped unit tests pass (including the authorization-code health suite covering ``omit_resource`` refresh semantics)
- 14 doctests pass (new ``oauth_resource`` utils, validation, storage)
- 101 JS unit tests pass (``vitest``)
- Single alembic head (``c9f8e7d6a4b3``)
- ``ruff check`` on changed files: no new findings


msureshkumar88:
Hi @jonpspri — the branch has merge conflicts with main. Could you rebase from main and resolve them when you get a chance? Happy to re-review once it's clean.

msureshkumar88 (changes requested):
## E2E verification + review — request changes

Reviewed the diff (15 files, +726/-132) and ran a real end-to-end verification: a live `mcp-context-forge` gateway + a real mock OAuth IdP (FastAPI/uvicorn, not mocked at the Python level) driving the actual `/oauth/authorize/{id}` → IdP → `/oauth/callback` flow and the admin UI form endpoints.

### E2E script

Script added: `scripts/verify-pr-4476-e2e.sh` (bash + a small FastAPI mock IdP written to a scratch file at runtime). It:

1. Starts a mock IdP with real `/authorize` (auto-approves, echoes `state`) and `/token` (issues an HS256 JWT with configurable `aud`/`iss`) endpoints.
2. Starts the gateway (`uvicorn mcpgateway.main:app`) against a fresh SQLite DB.
3. Registers a gateway with **no** `resource` configured, pointing at a URL with a deep path (not a bare origin).
4. Drives the full authorization-code flow through the real endpoints and inspects the actual `resource` param sent to the IdP, the persisted `oauth_config.resource` after callback, and admin-UI form round-trips.

**Config used:**
```
DATABASE_URL=sqlite:///<scratch>/e2e_pr4476.db
GW_PORT=44561  IDP_PORT=44562
AUTH_REQUIRED=true, MCPGATEWAY_ADMIN_API_ENABLED=true, MCPGATEWAY_UI_ENABLED=true
PASSWORD_CHANGE_ENFORCEMENT_ENABLED=false, MCPGATEWAY_DCR_ENABLED=false
SSRF_ALLOW_LOCALHOST=true, SSRF_ALLOW_PRIVATE_NETWORKS=true
```

**Commands run:** `bash scripts/verify-pr-4476-e2e.sh` (starts both servers, drives the flow with `curl`, tears down on exit).

**Result: 9/10 checks passed, 1 failed.**

```
PASS: mock IdP is up on :44562
PASS: gateway server is up on :44561
PASS: JWT generated
PASS: gateway registered, oauth_config.resource unset
PASS: AC 'Auto-derives origin if not set': resource=http://localhost:44562
      (origin only, not the full /mcp/v1/deep/path URL)
PASS: oauth callback completed (http 200)
PASS: audience learning: opaque aud 'test-servicenow-opaque-client-id'
      persisted as resource, overriding origin fallback
PASS: first-write-only: second callback (different aud) did not overwrite
      the learned resource
PASS: UI Resource field: comma-separated input persisted as a list via the
      gateway EDIT form
FAIL: gateway CREATE form: oauth_resource submitted but NOT persisted
```

The origin-derivation fallback, audience-learning/first-write-only semantics, issuer pinning, and the **edit**-form Resource field all work as described. The **create**-form case does not — see finding 1.

### Findings

**1. (blocking) "Add Gateway" form's new Resource field is a no-op — contradicts the PR's own stated scope**
`mcpgateway/admin.py:852-982` (`_parse_gateway_data_from_request`, used by `admin_add_gateway`) never reads `form.get("oauth_resource")`, unlike `admin_edit_gateway` (12859), `admin_add_a2a_agent` (16076), and `admin_edit_a2a_agent` (16357), which all got the new parsing block. Reproduced live: submitting the Add Gateway form with `oauth_resource=https://create-form.example.com` returns `200 OK` but the persisted gateway has `oauth_config.resource == null`. The PR description states scope as "gateway create + edit, A2A agent create + edit" — gateway create is the one path that was missed.

**2. A2A edit dialog's OAuth "Discover" button loses its status feedback**
`mcpgateway/templates/admin.html` ~10682-10726 vs. `discoverOAuthEndpoints()` at ~14962-14974. The handler locates the status element via `btn.parentElement.parentElement.querySelector('.oauth-discover-status')` — i.e. it expects the status div as a sibling of the Issuer-URL input row, inside the same wrapper `<div>`. This PR's restructuring splits that wrapper into two: a new standalone Issuer-URL `<div>` (label + input row only) and a separate Resource `<div>` that now contains the `<div class="oauth-discover-status">`. The lookup from the Discover button now resolves to the Issuer div, which has no status element, so clicking "🔍 Discover" on the A2A **edit** dialog silently drops all success/error/info feedback. (Worth noting: before this PR that status div was inside an HTML comment on the A2A edit dialog, so this wasn't previously wired up correctly either — but the PR had the status div right there to reattach and didn't.) The Gateway edit dialog wasn't restructured this way and is unaffected.

**3. `oauth_resource` comma-split corrupts any resource value containing a literal comma**
`mcpgateway/admin.py:12901-12906` (and the two duplicate blocks at 16118-16123, 16399-16404): `oauth_resource_str.split(",")` has no escaping. A single legitimate resource URI containing a comma (e.g. in a query string) gets silently split into two bogus resource entries instead of one, breaking audience matching for that gateway.

**4. (plausible, worth a second look) Full-replace `oauth_config` semantics can wipe an auto-learned `resource`**
`mcpgateway/services/encryption_service.py:578` (`_protect_oauth_config_value`) only walks keys present in the *new* submitted `oauth_config` dict — it never backfills keys (like `resource`) that exist in the stored config but are absent from the new one. Since `resource` can now be written asynchronously by the OAuth callback (`_persist_learned_audience`) independent of any admin edit, a stale edit-dialog submission (opened before the audience was learned) or an API caller passing a partial `oauth_config` JSON blob will silently drop the learned resource on save. In the normal single-tab UI flow this is mitigated by `gateways.js`/`a2aAgents.js` pre-populating the field from the freshly-fetched config, but it's a real race/API-caller gap, not purely theoretical.

**5. Two parallel URL-origin helpers where one would do**
`mcpgateway/routers/oauth_router.py:107-167`: `_derive_resource_origin` (new) re-implements a subset of `_normalize_resource_url`'s parsing instead of extending it (e.g. with an `origin_only: bool` flag). `_normalize_resource_url` has no remaining production call sites after this diff (`grep -rn _normalize_resource_url mcpgateway/` — only its own tests reference it), so it's effectively dead code now.

**6. Same ~9-line resource-parsing block duplicated 3x with no shared helper**
`mcpgateway/admin.py` at 12898-12906, 16115-16123, 16396-16404 — identical logic, copy-pasted (and, per finding 1, missing from a 4th call site that needed it too). A small `_parse_oauth_resource(str) -> str | list[str] | None` helper would have made the create-form gap in finding 1 structurally impossible.

**7. (minor) Redundant JWT decode per OAuth callback**
`mcpgateway/services/oauth_manager.py`: `_extract_token_audience` and `_extract_token_issuer` each independently call `_decode_token_claims_unverified`, so `complete_authorization_code_flow` decodes the same access-token JWT twice per callback. Pulling both claims from a single decode would halve this.

### What worked well

- Issuer pinning, shape validation, and first-write-only semantics for audience learning are solid and verified live end-to-end (including via the new test suite).
- The origin-derivation fallback is correct and matches the stated AC.
- `_decode_token_claims_unverified` is a clean factor-out (no duplication) — good pattern, just not applied to the admin.py form parsing or the oauth_router URL helpers.
- Test coverage added is thorough for the router/manager/validation-service logic.

Given finding 1 directly contradicts the PR's stated scope and is trivially reproducible, and finding 2 is a real (if pre-existing) UI regression touched by this diff, I'd request changes rather than approve. Findings 3–4 are worth fixing before merge; 5–7 are cleanup that could be a fast follow-up.

msureshkumar88 (changes requested):
## Follow-up deep review — request changes

Second pass covering security, performance, refactoring, test coverage, docs, breaking-change, and scope questions, scoped strictly to this branch's diff.

### Hardening

**High severity — advisory (non-blocking) audience mismatch weakens the core protection this PR is supposed to add.**
`mcpgateway/services/token_validation_service.py:102,188-209`, consumed by `mcpgateway/services/gateway_service.py:1948-1963`. When no `resource` is configured on a gateway, an audience mismatch is recorded as a warning only — `blocking_errors` excludes it, and `gateway_service.py` forwards the bearer token to the upstream MCP server regardless. There is no code-enforced guarantee the upstream server validates `aud` itself; the comment "the upstream MCP server is allowed to be the authoritative validator" is an assumption, not something this codebase can verify. Concrete scenario: a freshly-registered gateway with OAuth configured but no successful login yet (the default state for every new gateway) will forward *any* valid bearer token from the same IdP tenant — including one minted for a completely unrelated resource — to the upstream server with only a log line. If that upstream server doesn't check `aud` (common for lightweight/homegrown MCP servers), this is a real cross-resource token replay. Recommend gating the advisory relaxation behind an explicit, default-off setting rather than making it the default for every unconfigured gateway.

**Medium — first-write-only audience learning can cause a self-inflicted denial of service across tenants.**
`mcpgateway/routers/oauth_router.py:242-269`. If one gateway serves users from different tenants of the same IdP with per-tenant `aud` values, the first successful login permanently pins `resource`. Once pinned, mismatches become *blocking* (resource is now configured), so the second tenant's legitimately different `aud` gets rejected outright until an admin manually clears the field. Worth documenting as an operational caveat at minimum.

**Verified clean, no action needed:** the unverified-JWT-decode design for audience/issuer learning is deliberately scoped and doesn't feed any authorization decision beyond seeding the persisted `resource` value; the new debug logs don't leak token contents; `.secrets.baseline` changes all correspond to fake test fixtures; no new CSRF/XSS/SSRF surface (JS sets `.value`, not `innerHTML`; resource values are only ever sent to admin-configured endpoints).

### Test coverage — this is the biggest gap

**The advisory/blocking softening has zero coverage at its actual enforcement point.** `gateway_service.py:1948-1963` is the only production caller of `blocking_errors`, and there are zero references to it, `Refusing to forward`, or `validate_oauth_token_claims` in `tests/unit/mcpgateway/services/test_gateway_service.py`. Every test for `audience_authoritative` only asserts against the `TokenValidationResult` dataclass directly — nothing proves a mismatched-audience token is actually forwarded (advisory case) or actually blocked (authoritative case) through the real code path. A regression in the condition at `token_validation_service.py:102` would go undetected.

Other gaps:
- No unit test for `admin.py`'s `oauth_resource_str` comma-split parsing (single vs. multi-value shape, or a resource value containing a literal comma) in any of the three handlers that have it.
- No JS test for the new `oauthResourceField` population logic in `gateways.js`/`a2aAgents.js`.
- No integration/e2e test drives the full initiate → callback → persist cycle against a real DB; the existing `test_oauth_protected_resource.py` covers unrelated RFC 9728 metadata only. The router-level unit tests are thorough at the function level but there's a real gap to the admin-form-to-database round trip.

### Refactoring / dead code

- `_normalize_resource_url` (`oauth_router.py:107-134`) has zero remaining production call sites after this diff — `_derive_resource_origin` replaced both of them instead of extending it with an `origin_only` option. It's dead code now, kept alive only by its own tests.
- The Add Gateway form's new Resource input (`admin.html` ~line 5618) is dead end-to-end: it renders, the form submits it, but `_parse_gateway_data_from_request` (`admin.py:852-982`, used by the gateway CREATE handler) never reads `oauth_resource` at all. This is the same gap flagged in the first review pass, now confirmed as a literal dead branch rather than just a missing call site.
- The same 9-line `oauth_resource` parsing block is duplicated verbatim across the three handlers that do have it, with minor copy-paste drift (`oauth_audience` is `.strip()`-ed at assignment, `oauth_resource_str` is not, though the effective output is still correct).

### Implementation vs. PR description

Every claimed acceptance criterion is met except "OAuth Resource field visible in Admin UI," which is only partially true — edit dialogs and A2A add/edit work, but gateway **create** silently drops the field (see dead-code item above).

### Documentation

No `docs/` path is touched anywhere in this diff, despite the PR changing observable validation behavior (audience mismatches silently stop blocking in the unconfigured-resource case). 100% of documentation is deferred to a follow-up PR. Given this is a security-relevant behavior change, at least a short note in the existing OAuth docs would help operators understand the new advisory-vs-blocking distinction before the follow-up lands.

### Breaking change

Not a hard break, but a real behavior change: gateways that had no `resource` configured previously got the full gateway URL as the expected audience; they now get origin-only. Because the check is advisory when unconfigured, no gateway should hard-fail from this alone, but the local validation behavior does change silently. No database/schema impact — `oauth_config` is a JSON blob column.

### Scope

No unrelated functional changes found beyond the excluded categories (`.secrets.baseline`, pragma-comment-only test edits). The softened-validation design maps 1:1 to its stated acceptance criterion and isn't broader than needed.

---

Given the unresolved gateway-create field gap (already flagged), the A2A edit dialog's broken Discover-button feedback (already flagged), and the advisory-audience-mismatch design having no test coverage at its actual enforcement point, requesting changes rather than approving.

msureshkumar88 (changes requested):
## Re-review — request changes (2 small items)

Re-checked against both prior review rounds. All 7 findings from the first pass and both the HIGH/MEDIUM findings plus the test-coverage and docs gaps from the second pass are resolved:

- Gateway create-form Resource field now wired (`_parse_oauth_resource` used in all 4 form-parsing call sites)
- A2A edit dialog Discover-button status feedback fixed
- Comma-split no longer corrupts resource URIs containing literal commas
- Learned audience moved to per-user `OAuthToken.learned_aud`/`learned_iss` (migration `c9f8e7d6a4b3`, single alembic head) — closes the shared-config-wipe risk and the cross-tenant DoS risk in one architectural change
- Dead `_normalize_resource_url` / duplicated parsing blocks removed
- Redundant JWT decode merged into one `_extract_aud_and_iss()`
- New `TestFetchToolsAfterOAuthEnforcementPoint` (7 tests) now covers the actual `Refusing to forward` enforcement point in `gateway_service.py` — this was the biggest gap flagged last round
- New docs (`oauth-resource-configuration.md`, `oauth-troubleshooting.md`) explain the precedence rules and include an explicit `!!! warning "Advisory Validation Security Exposure"` admonition

Ran the full affected unit-test surface locally (gateway_service, token_validation_service, oauth_manager, oauth_router, token_storage_service, admin) — all green.

### Requesting changes on 2 small items

1. **`.env.example` is missing `OAUTH_REQUIRE_CONFIGURED_RESOURCE`.** It's implemented in `config.py:925` and documented in the new resource-configuration doc, but not listed in `.env.example`. Since this is the toggle that makes the gateway itself authoritative on audience mismatches (vs. relying on the upstream MCP server), it'd be easy for operators to miss if they're scanning `.env.example` for available settings rather than reading the full doc page.

2. **No JS test for the `oauthResourceField` pre-population logic.** `gateways.js:402/530` and `a2aAgents.js:481/589` populate the Resource field from `config.resource` (string or list shape) when the edit dialog opens. There's solid Python-side coverage for the round trip, but nothing pins this JS-side behavior — a refactor could silently break edit-dialog pre-population for the list-shape case with no test catching it.

### Non-blocking note (no action required)

The advisory-vs-authoritative audience validation still defaults to advisory for a gateway until the current user's first successful OAuth login populates their `learned_aud`. This is a real reduction from the design flagged HIGH last round (now scoped per-user, opt-out via `OAUTH_REQUIRE_CONFIGURED_RESOURCE=true`, and explicitly called out with a warning admonition in the docs), so I'm not blocking on it — just flagging that it's a documented tradeoff rather than fully closed.

Once 1 and 2 land this looks ready to merge.

msureshkumar88:
## E2E re-verification (updated for per-user learned-audience redesign)

Re-ran the E2E script against a live gateway + a real mock IdP (FastAPI/uvicorn, not mocked at the Python level). Updated `scripts/verify-pr-4476-e2e.sh` first, since two of its checks were asserting the *pre-redesign* persistence location — details below.

### Config used

```
DATABASE_URL=sqlite:////tmp/tmp.bWyrQJgrdh/e2e_pr4476.db
GW_PORT=44561  IDP_PORT=44562
Gateway commit: 7082fc270 (test(gateway): stub get_user_learned_audience on mocked TokenStorageService)
Branch: feat/oauth-aud-resource-hybrid

JWT_SECRET_KEY=e2e-pr4476-test-secret
AUTH_ENCRYPTION_SECRET=e2e-pr4476-test-salt
AUTH_REQUIRED=true
MCPGATEWAY_UI_ENABLED=true
PLUGINS_ENABLED=false
SSRF_ALLOW_LOCALHOST=true
SSRF_ALLOW_PRIVATE_NETWORKS=true
SSRF_DNS_FAIL_CLOSED=false
MCPGATEWAY_DCR_ENABLED=false
PASSWORD_CHANGE_ENFORCEMENT_ENABLED=false
```

### Command run

```bash
bash scripts/verify-pr-4476-e2e.sh
```

### What changed in the script and why

The first run against this branch showed 8/10 PASS, 2 FAIL. Both failures were the script checking `GET /gateways/{id}` → `oauth_config.resource` for the learned audience — that's the pre-redesign persistence location. This PR's per-user security fix (`059492d6f`) deliberately stops writing the learned aud into shared `gateway.oauth_config`; it now lives on the caller's own `OAuthToken.learned_aud` row (fixes the cross-tenant DoS / RBAC-bypass finding from the earlier review round). So the script was asserting behavior the PR intentionally removed, not catching a real regression.

Updated the two checks to:
- query `oauth_tokens.learned_aud` directly instead of `oauth_config.resource`
- assert `oauth_config.resource` stays unset (proves no shared-config leak)
- replace the old "first-write-only" check (belonged to the shared-config design) with a "re-auth refreshes the same user's own row" check (correct per-user semantics)

Also found and fixed a latent bug while rewriting step 8: it used `PATCH /gateways/{id}`, but that route only accepts `PUT` — the PATCH 405'd silently because the response was piped to `/dev/null`. Switched to `PUT`.

### Output (all 10 checks + full 11th sub-check pass)

```
== Config used ==
DATABASE_URL=sqlite:////tmp/tmp.bWyrQJgrdh/e2e_pr4476.db
GW_PORT=44561  IDP_PORT=44562
Gateway commit: 7082fc270 (test(gateway): stub get_user_learned_audience on mocked TokenStorageService)
Branch: feat/oauth-aud-resource-hybrid

== 1. Start mock IdP ==
PASS: mock IdP is up on :44562

== 2. Start mcp-context-forge gateway server ==
PASS: gateway server is up on :44561

== 3. Generate bearer token + admin session cookie ==
PASS: JWT generated

== 4. Register gateway (no 'resource' configured) pointing at a *path*, not bare origin ==
register response (200)
PASS: gateway registered (id=05f26054aa664904bab9dc58a1697560), oauth_config.resource unset

== 5. GET /oauth/authorize/{id} and inspect the redirect Location's 'resource' param ==
authorize redirect Location: http://127.0.0.1:44562/authorize?...&resource=http%3A%2F%2Flocalhost%3A44562
PASS: AC 'Auto-derives origin if not set': resource=http://localhost:44562 (origin only, not full URL with /mcp/v1/deep/path)
PASS: state param captured for callback drive

== 6. Drive the callback: mock IdP auto-approves, gateway exchanges code for token ==
PASS: oauth callback completed (http 200)

== 7. Assert the opaque aud from the IdP token was LEARNED per-user (OAuthToken.learned_aud), NOT written to shared oauth_config ==
PASS: audience learning: opaque aud 'test-servicenow-opaque-client-id' persisted per-user as OAuthToken.learned_aud
PASS: gateway.oauth_config.resource remains unset -- learned aud stayed per-user, did not leak into shared config

== 8. Re-authentication refresh: SAME user's second callback with a different aud updates their OWN learned_aud ==
PASS: re-auth refresh: same user's learned_aud updated to 'different-tenant-aud' (expected -- per-user state refreshes freely on that user's own re-auth)

== 9. Admin UI form round-trip: comma-separated oauth_resource persists as a list ==
PASS: admin session login succeeded
admin edit response: http 200
resource after admin edit: ["https://custom-a.example.com", "https://custom-b.example.com"]
PASS: UI Resource field: comma-separated input persisted as a list via admin edit form

== 10. Gateway CREATE form: oauth_resource field (PR claims 'gateway create + edit' scope) ==
admin create response: http 200
resource on newly-created gateway: "https://create-form.example.com"
PASS: gateway CREATE form: oauth_resource field persisted

==================================
RESULT: ALL CHECKS PASSED
```

Confirms end-to-end: origin-derivation fallback, per-user audience learning with no shared-config leak, per-user re-auth refresh, and both the gateway edit and create UI form paths all work as designed.

Holding off on approval per the still-open items from the last review round (`.env.example` entry for `OAUTH_REQUIRE_CONFIGURED_RESOURCE`, JS test for `oauthResourceField` pre-population).

msureshkumar88 (dismissed):
Verified both outstanding items from the last review round are resolved in `ae664b6bd`:

1. `.env.example` now documents `OAUTH_REQUIRE_CONFIGURED_RESOURCE=false` (backed by `config.py:924`), with a comment pointing at `docs/docs/manage/oauth-resource-configuration.md`.
2. `tests/unit/js/gateway.test.js` and `tests/unit/js/a2aAgents.test.js` now pin the `oauth-resource-*-edit` pre-population logic for both the RFC 8707 string shape and the list shape.

Re-ran locally against head `c293d7ef6`:
- `vitest run tests/unit/js/gateway.test.js tests/unit/js/a2aAgents.test.js` — 103 passed, 0 failed.
- `pytest tests/unit/mcpgateway/services/test_gateway_service.py -k OAuthEnforcement` — 7/7 pass, confirming the advisory-vs-blocking split is covered at its real enforcement point (`gateway_service.py`), which was the biggest gap flagged in round 2.
- Broader oauth/resource/aud-scoped tests across `test_oauth_manager.py`, `test_token_validation_service.py`, `test_oauth_router.py`, `test_admin.py` — all green.
- `alembic heads` — single head (`c9f8e7d6a4b3`), no branch conflict.

All findings from all three review rounds (7 from pass 1, HIGH/MEDIUM + coverage/docs gaps from pass 2, the 2 small items from pass 3) are addressed. The per-user `learned_aud`/`learned_iss` architecture change closes the shared-config-wipe and cross-tenant-DoS concerns structurally rather than just patching symptoms. Approving.

msureshkumar88 (dismissed):
LGMT

Lang-Akshay (changes requested):
Thanks for the PR @jonpspri

### Blocking findings

- **mcpgateway/services/token_storage_service.py:410-414** — 🔴 bug: OAuth refresh derives resource from the full gateway.url when no configured resource exists, preserving paths such as `/platform/mcp/v1`; initial authorization now uses the origin, so refresh requests can use a different audience and fail or obtain an inconsistent token. Reuse the origin-derivation helper in the refresh path and add coverage for path-bearing gateway URLs.

- **mcpgateway/services/token_validation_service.py:208-209** — 🔴 bug: Validator fallback compares `aud` against the full gateway URL instead of its origin, contradicting the OAuth route behavior and documented contract. With `OAUTH_REQUIRE_CONFIGURED_RESOURCE=true`, valid origin-level audiences are rejected when no learned/configured resource exists. Derive the fallback origin consistently and test strict mode.

Lang-Akshay (dismissed):
LGTM

msureshkumar88 (dismissed):
## E2E verification: real live-server run, not just unit tests

I wrote and ran a hermetic end-to-end script (`scripts/verify-pr-4476-e2e.sh`) that stands up two *live* `uvicorn mcpgateway.main:app` instances — one at `origin/main`, one at this PR's HEAD, each with its own throwaway SQLite DB and port — plus a real standalone mock OAuth IdP HTTP server, and drives the fix over real HTTP (not mocks of the business logic under test), closing with a targeted pytest regression pass and a fair main-vs-PR-head lint comparison.

**Config used:** `SSRF_ALLOW_LOCALHOST=true` (both the gateway-under-test and mock IdP run on loopback), `ADMIN_REQUIRE_PASSWORD_CHANGE_ON_BOOTSTRAP=false`, `AUTH_REQUIRED=true`, `MCPGATEWAY_UI_ENABLED=true`, `MCPGATEWAY_ADMIN_API_ENABLED=true`, `PLUGINS_ENABLED=false`, fresh SQLite DB per worktree (schema created automatically via the app's own `bootstrap_db()` lifespan, no manual `alembic upgrade` needed), one shared `.venv` for both worktrees (confirmed no dependency changes between `main` and this PR).

**Commands run:**
```
bash scripts/verify-pr-4476-e2e.sh
```
(plus, independently from the main thread: `uv tool run ruff check <changed .py files>` on both `origin/main` and PR-head for a diff-fair comparison, `npx vitest run tests/unit/js/gateway.test.js tests/unit/js/a2aAgents.test.js`, `alembic heads` on both worktrees, `git merge-tree $(git merge-base HEAD origin/main) HEAD origin/main`)

**Result: 30/30 PASS, exit 0**, run three times for reproducibility (last two clean).

<details>
<summary>Full output</summary>

```
== PR #4476 e2e verification: OAuth Resource UI field, origin fallback, softened validation (closes #4384) ==
PASS: created hermetic main-baseline worktree @ 24e923c95
PASS: resolved shared venv python
PASS: no dependency changes between origin/main and PR head -- safe to reuse one shared venv for both worktrees
PR head alembic heads:   c9f8e7d6a4b3 (head)
main baseline alembic heads: d21698ae4a19 (head)
PASS: PR head has exactly one alembic head
PASS: main baseline has exactly one alembic head
PASS: mock OAuth IdP (real HTTP server) listening
PASS: PR-head server healthy
PASS: main-baseline server healthy
PASS: minted admin Bearer JWTs for both servers

== 1. Admin UI resource field wiring ==
PASS: PR head: POST /admin/gateways with oauth_resource form field -> oauth_config['resource'] == submitted value
PASS: REPRODUCED on main: oauth_resource form field is submitted but silently dropped -- no 'resource' key at all

== 2. Origin-fallback derivation, end to end ==
PASS: [PR head] origin-derived callback succeeds (single first-time user)
PASS: [PR head] fetch-tools not blocked by audience mismatch
PASS: [main baseline] origin-derived callback succeeds (single first-time user)
PASS: [main baseline] fetch-tools not blocked by audience mismatch
NOTE: single first-time user succeeds on BOTH main and PR head, because
  already-merged PR #4475 added a first-write-only self-heal that learns
  and persists into SHARED gateway.oauth_config on the first callback.
  main still derives the WRONG (full-path) resource outbound -- the gap
  this masks is exposed in section 4 (multi-user).

== 3. Softened/advisory validation for derived audience ==
PASS: precondition confirmed via direct DB read: learned_aud IS NULL (audience_source will be 'derived')
PASS: default mode: derived-audience mismatch is ADVISORY -- fetch-tools proceeds past the audience check
PASS: PR-head (strict) server healthy after restart with OAUTH_REQUIRE_CONFIGURED_RESOURCE=true
PASS: strict mode: the SAME derived-audience mismatch now BLOCKS fetch-tools ('Refusing to forward OAuth token...Token audience mismatch')

== 4. Per-user learned-audience isolation (cross-tenant) ==
PASS: [PR head] both user A and user B fetch-tools succeed past the audience check -- per-user learned_aud isolates them
PASS: [PR head] shared gateway.oauth_config['resource'] untouched by either user's callback
PASS: [PR head] DB inspection: two independent oauth_tokens rows, each with its own correct learned_aud
PASS: REPRODUCED on main: user A's callback silently persists into SHARED gateway.oauth_config['resource'];
      user B's real, valid OAuth token is then rejected with 'Token audience mismatch'
      -- the cross-tenant DoS / RBAC-bypass half of issue #4384
PASS: [main] confirmed: shared oauth_config['resource'] pinned to user A's audience, silently ignoring user B's

== 5. Regression pass (pytest / ruff / alembic) ==
-- tests/unit/mcpgateway/routers/test_oauth_router.py tests/unit/mcpgateway/services/{test_oauth_manager,test_token_storage_service,test_token_validation_service,test_gateway_service_oauth_comprehensive}.py tests/unit/mcpgateway/test_admin.py --
........................................................................ [ 97%]
...............................................                          [100%]
PASS: targeted OAuth/admin pytest regression suite is green
PASS: ruff: brand-new files (oauth_resource.py, the migration) are fully clean
   origin/main version of touched-but-preexisting files: 62 findings
   PR head version of same files: 58 findings
PASS: ruff finding count did not increase vs origin/main baseline (58 <= 62)

== 6. Surrounding-area sanity: A2A agent OAuth config after admin.py refactor ==
PASS: A2A agent creation succeeds after the shared _assemble_oauth_config_from_fields refactor
PASS: A2A agent's oauth_config has NO 'resource' key even though oauth_resource was submitted -- include_resource=False honored post-refactor

== Summary ==
Total: 30  Passed: 30  Failed: 0
RESULT: ALL CHECKS PASSED
```
</details>

Independently from the main thread (not just inside the script): `alembic heads` is single-headed on both `origin/main` (`d21698ae4a19`) and this branch (`c9f8e7d6a4b3`), and `git merge-tree` between the two shows **zero** conflict markers — the migration's `down_revision` correctly targets current main's head. 101/101 JS unit tests pass (`gateway.test.js`, `a2aAgents.test.js`), matching the PR description.

## What this proves

Section 4 is the crux: it reproduces, live, the exact cross-tenant/RBAC-bypass gap that's the remaining piece of #4384 not covered by the already-merged #4475 — one user's OAuth callback on `main` silently clobbers the shared `gateway.oauth_config['resource']`, and a second user's genuinely valid token then gets rejected. On this PR's head, the same two-user sequence succeeds for both, because the learned audience moved to per-user `OAuthToken.learned_aud`. Section 3 confirms the advisory/strict split behaves exactly as documented (`OAUTH_REQUIRE_CONFIGURED_RESOURCE`). No regressions found in the surrounding A2A form-parsing refactor or the broader OAuth test suite.

## One thing to fix before merge (not a functional issue)

CI's **"Run pre-commit hooks"** check is currently red on this HEAD (133b539f2) — `detect-secrets` regenerated `.secrets.baseline` and it wasn't committed. Touching `tests/unit/mcpgateway/test_auth_ratelimiter_redis.py` (2-line diff in this PR) shifted a secret-pattern match, so the baseline has a stale entry (`test_auth_ratelimiter_redis.py:128`, `Basic Auth Credentials`) that no longer matches. Per this repo's usual process: `make detect-secrets-scan` and commit the updated `.secrets.baseline`. Purely mechanical, not a real secret and not related to the OAuth logic — just needs the regen committed so the required check goes green.

(Separately: GitHub's mergeable flag shows `CONFLICTING`/`DIRTY`, but `git merge-tree` against current `origin/main` found no actual conflict markers — worth a rebase to clear the stale flag before merge, but not something I'd block on given the check itself came back clean.)

Approving based on the live-server evidence above — the fix demonstrably closes the remaining #4384 acceptance criteria with no surrounding regressions. Please regenerate `.secrets.baseline` so the pre-commit CI check passes before merging.

msureshkumar88:
One more item for merge readiness: GitHub currently reports this branch as `CONFLICTING`/`DIRTY` against `main`. Could you rebase onto latest `origin/main` and resolve?

For what it's worth, a local `git merge-tree $(git merge-base HEAD origin/main) HEAD origin/main` against current `origin/main` (`24e923c9`) came back with zero conflict markers, and `alembic heads` is single-headed on both sides (`c9f8e7d6a4b3` here, `d21698ae4a19` on `main`) — so whatever GitHub is flagging isn't a straightforward content conflict in the files I checked. Worth a rebase anyway to clear the flag and get a clean mergeable state before merging, and it'll also pick up the `.secrets.baseline` regen noted in the review above.

msureshkumar88 (dismissed):

msureshkumar88 (approved):

PR #4477, opened 2026-04-27
[DOCS]: OAuth audience handling, resource configuration, and troubleshooting

jonpspri:
## Summary

Documentation closing the last open acceptance criterion from issue #4384.  Stacked on #4476 (hybrid features); base must be ``feat/oauth-aud-resource-hybrid`` not ``main``.

### Updated: ``docs/docs/architecture/oauth-design.md``

- **Removed** the "Known Gap: No local audience/scope pre-validation" admonition (no longer applies after #4404 + #4476).
- **Added** new "Audience handling and RFC 8707" section explaining:
  - The three-stage audience resolution: explicit config -> learned from IdP -> origin fallback
  - The authoritative-vs-advisory validation distinction
  - Implementation references with file:line links

### Created: ``docs/docs/manage/oauth-resource-configuration.md`` (new)

End-user guide covering:

- When you don't need to do anything (most cases auto-derive + auto-learn)
- When to set the field explicitly (multi-tenant Entra ID, multiple audiences, recovery)
- Provider patterns for Salesforce, Azure AD, Authentik, ServiceNow, Keycloak
- Multi-resource configurations (comma-separated UI input)
- How to force a re-learn after IdP migration (clear the field)
- Authoritative vs advisory validation behavior

Wired into ``docs/docs/manage/.pages``.

### Updated: ``docs/docs/manage/oauth-troubleshooting.md``

New "Token audience mismatch" section with:

- Symptom (the specific error message)
- Diagnosis steps (check oauth_config, decode JWT, check IdP RFC 8707 support)
- Common causes and fixes (Authentik/ServiceNow client_id, multi-tenant Entra ID, stale learned resource, Salesforce origin)
- Useful debug log targets
- Clarification of gateway/upstream MCP server validation responsibility split

## Stack

| | Branch | Status |
|---|---|---|
| 1 | ``bugfix/oauth-aud-resource-narrow`` | #4475 — must merge first |
| 2 | ``feat/oauth-aud-resource-hybrid`` | #4476 — must merge second |
| **3** | ``docs/oauth-audience-handling`` | **this PR** |

## Acceptance criteria from #4384

This PR closes the remaining documentation criterion.  The full set is checked in #4476's description.

## Verification

- 496 affected unit tests still pass (no behavior changes)
- Mkdocs build was not run locally; rendering should be verified in CI


msureshkumar88 (changes requested):
Good structure and clear writing throughout. A few gaps to address before merge.

## Blocking

**Merge order dependency**: Base branch is `feat/oauth-aud-resource-hybrid`, not `main`. Merging out of order documents behavior that doesn't exist yet. PR description covers this, but worth confirming CI/branch protection enforces the #4475 → #4476 → this sequence — a well-meaning reviewer could merge early.

## Security Hardening

**`oauth-design.md`, Implementation Details — unverified decode trust boundary**: The reference to `_decode_token_claims_unverified` notes it is an "unverified JWT decode used for metadata extraction" but stops there. Without context, readers may infer audience extraction is a security gate. Needs an explicit callout: the extracted audience is **not cryptographically attested** — it is used for routing/learning only. The upstream MCP server's signature verification remains the only cryptographic authority. Suggest adding an admonition block here.

**`oauth-resource-configuration.md`, Validation Behavior — advisory path risk not stated**: Advisory validation says "a warning is logged, but the token is still forwarded." Missing: what exposure this creates. If the upstream MCP server does not enforce audience validation, a mismatched-audience token may be accepted — a privilege-escalation risk rooted in IdP or MCP server misconfiguration. Worth stating explicitly so operators understand what they are relying on.

## Documentation Gaps

**Line numbers will rot**: Implementation detail references pin specific line numbers (`oauth_router.py:76`, `oauth_router.py:128`, `oauth_manager.py:1608`, `oauth_manager.py:1657`). These will be wrong after the next code edit. Drop line numbers and keep file+function references only, or use a commit-SHA GitHub permalink.

**Auto-learning failure path undocumented**: "Automatic Handling" describes the happy path — first successful OAuth callback persists the audience. Missing: what happens when the IdP does not return an RFC 8707-compliant response on first callback. Does the system permanently fall through to origin fallback? Can it recover without manual intervention? This is a real operator failure mode.

**Debug log level not mentioned**: `oauth-troubleshooting.md` references log messages like `Skipping audience persistence for gateway X: ...` with no note that these are DEBUG-level. At default `LOG_LEVEL=ERROR` or `INFO` these are invisible. Add: "Set `LOG_LEVEL=DEBUG` to see these messages."

**Keycloak row is overconfident**: Provider Patterns table says Keycloak uses `"account" or "your-client-name"`. Actual Keycloak audience behavior depends on whether `Use Client ID As Audience` is enabled, whether client scopes define explicit audience mappers, and Keycloak version. Replace with "configuration-dependent — see Keycloak's audience mapper documentation" rather than implying fixed values.

**No cross-link from `oauth.md`**: New page is wired into `.pages` nav, but inline readers of the main `oauth.md` setup guide won't discover it. Add a "See also" or inline reference from the gateway configuration section of `oauth.md` to `oauth-resource-configuration.md`.

**Dict syntax vs UI terminology**: `oauth-troubleshooting.md` references `oauth_config["resource"]` (Python dict syntax). Operators configuring via Admin UI won't map this to anything. Replace with "the **Resource** field in the gateway's OAuth configuration" — consistent with the UI language used elsewhere in this PR.

## Logging Gap

**Advisory mismatch log message not shown**: Architecture doc says "a warning is logged" for advisory mismatches but doesn't give the log message format. The troubleshooting guide has a concrete skip-persistence example — do the same for advisory mismatch so operators know what to grep for.

## Testing

No gaps — pure docs PR, no code behavior changes. "496 unit tests pass" is correct. One note: confirm CI runs `mkdocs build --strict` to catch broken cross-references and malformed admonitions. If `--strict` is absent, broken links in the new pages won't surface until post-merge.

## Non-issues

- No unrelated changes
- Not a breaking change
- Issue #4384 correctly referenced; stack (#4475, #4476) clearly documented

jonpspri:
Closing in favor of #4476, which now absorbs all documentation from this branch after the reviewer's follow-up round.

**What was integrated:**
- `docs/docs/manage/oauth-resource-configuration.md` — new file, added verbatim aside from the accuracy fix below
- `docs/docs/manage/.pages` — nav entry added
- `docs/docs/architecture/oauth-design.md` — 'Audience handling and RFC 8707' section added; obsolete 'Known Gap: No local audience/scope pre-validation' admonition removed; 'Forwarded as-is' bullet updated to describe the actual pre-validation
- `docs/docs/manage/oauth-troubleshooting.md` — 'Token audience mismatch' section added with diagnosis / common causes / debug-log guidance
- `docs/docs/manage/oauth.md` — Resource field bullet + tip admonition + cross-link to the new page

**One accuracy fix on the way in:** the 'Forcing a Re-learn' section in this PR still described the pre-review workflow of clearing the UI field. #4476 now includes an encryption-service backfill for race-prone `oauth_config` keys (`_OAUTH_CONFIG_RACE_PRONE_KEYS`) that intentionally preserves `resource` when the caller submits an oauth_config dict that omits the key — the fix for the reviewer's stale-save-race finding. UI clear no longer wipes the stored value; the API-with-explicit-null workflow is now the documented path. Same correction applied consistently in the troubleshooting page and the summary section of `oauth.md`.

**Skipped from this branch's diff:** three section removals that would have been accidental given how far main / #4476 have moved since this branch was cut — the `## Inbound External-Token Validation (M2M API Auth)` section in oauth-design.md, the `## External IdP Bearer Token Rejected on API/MCP` table in oauth-troubleshooting.md, and the `### Audience Parameter` / Atlassian / Auth0 blocks in oauth.md (the `audience` request parameter is a genuinely different field from `resource`).

All function/log/setting/error-string references in the integrated docs verified against the code in #4476.

Thanks for putting these docs together — kept the structure, just needed the one behavior correction to match #4476's race-safety fix.

Issue #4138, opened 2026-04-13
[TESTING][DB]: Migration testing gaps — rollback verification, cross-DB schema validation, and CI coverage


jonpspri:
### Summary

Follow-up from #252. The core migration testing infrastructure is comprehensive (~85% complete), but three gaps remain worth addressing.

#### 1. Verify rollback tests execute under pytest
- `test_sequential_reverse_migrations()` in `tests/migration/test_docker_sqlite_migrations.py` (line 74) is defined but relies on the `reverse_version_pair` fixture for parametrization
- Confirm this fixture is properly supplied and the reverse migration tests are actually triggered during `make migration-test-all`
- Add a dedicated `make migration-test-rollback` Makefile target that runs only reverse migration tests

#### 2. Cross-database schema validation
- SQLite and PostgreSQL migrations are tested independently but no test asserts that the same migration produces an equivalent schema across both engines
- Add a test that runs a migration on both SQLite and PostgreSQL, then diffs the resulting schemas (table names, column types, constraints, indexes) accounting for expected engine-specific differences

#### 3. Full migration suite in CI
- `.github/workflows/alembic-upgrade-validation.yml` currently runs only the upgrade validation script, not the full `make migration-test-all` suite (which includes rollback and performance)
- Consider adding the full suite as a scheduled or release-gated CI job (not necessarily on every PR, given the runtime cost)

### Context

The existing infrastructure covers:
- Forward migration testing (sequential + skip-version) for SQLite and PostgreSQL
- Rollback test definitions with `downgrade()` in all 45 Alembic migrations
- Performance benchmarking with scalability testing (1x/5x/10x)
- Data integrity validation via `SchemaValidator` and `DataSeeder`
- Makefile targets: `migration-test-all`, `migration-test-sqlite`, `migration-test-postgres`, `migration-test-performance`
- CI upgrade validation on every PR
- Comprehensive documentation in `tests/migration/README.md`

### Acceptance Criteria

- [ ] `make migration-test-all` exercises reverse migration tests (verified via CI log or local run)
- [ ] `make migration-test-rollback` target exists and runs only downgrade tests
- [ ] At least one test compares post-migration schema across SQLite and PostgreSQL
- [ ] Decision documented on whether full suite should run in CI (and if so, on what trigger)

Refs #252

PR #4479, opened 2026-04-27
Fix/alembic migration safety

### LINKED ISSUE #4429: [BUG]: Alembic `downgrade` is unsafe, environment-dependent, and untested across multiple migrations
## Summary

`alembic upgrade head` works correctly for all migrations. However, `alembic downgrade` is broken or unsafe across several migrations due to four distinct root-cause categories:

1. **Destructive schema rollback** — base migration `b77ca9d2de7e` explicitly destroys all data on downgrade (UUID/slug refactor).
2. **Runtime settings dependency** — migrations `ba202ac1665f`, `b77ca9d2de7e`, `a706a3320c56` import `mcpgateway.config.settings` at module scope, making downgrade behavior non-deterministic if environment variables change between upgrade and downgrade.
3. **Silent crypto failure** — migration `a706a3320c56` (Argon2id re-encryption) catches all exceptions and skips failed re-encryptions silently, leaving data in an inconsistent encryption state on downgrade.
4. **Irreversible data backfills** — migration `ba202ac1665f` Phase 2 injects RBAC rows with `granted_by=user_email` that are indistinguishable from legitimate role assignments, so the `downgrade()` explicitly cannot clean them up.
5. **SQLite/PostgreSQL asymmetry on downgrade** — `a3c38b6c2437` is a no-op for SQLite on both upgrade and downgrade (type not enforced), but performs actual `ALTER COLUMN` on PostgreSQL. Downgrading on SQLite leaves schema in a different state than before the upgrade.

---

## Affected Migrations

| File                                                    | Issue Category                                         | Severity           |
| ------------------------------------------------------- | ------------------------------------------------------ | ------------------ |
| `b77ca9d2de7e_uuid_pk_and_slug_refactor.py`           | Destructive rollback — all row data lost              | **Critical** |
| `ba202ac1665f_migrate_user_roles_to_configurable_.py` | Env-dependent reversal + irreversible Phase 2 backfill | **High**     |
| `a706a3320c56_use_argon2id_for_encryption_key.py`     | Silent crypto failure on downgrade re-encryption       | **High**     |
| `a3c38b6c2437_fix_a2a_agents_auth_value.py`           | SQLite no-op vs PostgreSQL ALTER asymmetry             | **Medium**   |
| `cfc3d6aa0fb2_consolidated_multiuser_team_rbac_.py`   | No downgrade implemented (tables remain)               | **Medium**   |

---

### 1. Destructive Base Migration Downgrade

**File:** `mcpgateway/alembic/versions/b77ca9d2de7e_uuid_pk_and_slug_refactor.py`

The `downgrade()` function itself documents the problem:

```python
def downgrade() -> None:
    """Revert database schema from UUID primary keys back to integers.
    ...
    - Schema structure is restored but data is NOT preserved
    - All UUID values and slug fields are lost
    - Foreign key relationships are broken (columns will be NULL)
    - Original integer IDs cannot be recovered
    ...
    Warning:
        This downgrade is destructive and should only be used if you need
        to revert the schema structure. All data in affected tables will
        need to be manually restored from backups.
    """
```

There is no guard, no `NotImplementedError`, and no pre-flight backup check. Running `alembic downgrade base` silently drops all production data.

---

### 2. Runtime Settings Dependency — Non-deterministic Downgrade

**Files:** `ba202ac1665f`, `b77ca9d2de7e`, `a706a3320c56`

Multiple migrations import application settings at module scope:

```python
# ba202ac1665f_migrate_user_roles_to_configurable_.py
...
def downgrade() -> None:
    # Attempt role remap reversal (environment-dependent)
    new_admin_role = settings.default_admin_role   # reads LIVE env var
    ...
    # WARNING: This assumes the current DEFAULT_*_ROLE env vars match
    # those used during upgrade. If env vars have changed between upgrade
    # and downgrade, the reversal may be incorrect.
```

If `DEFAULT_ADMIN_ROLE`, `DEFAULT_USER_ROLE`, `DEFAULT_TEAM_OWNER_ROLE`, or `DEFAULT_TEAM_MEMBER_ROLE` environment variables differ between the time of `upgrade` and the time of `downgrade` (e.g., after a config change or Helm values update), the role remapping reversal silently operates on wrong role names, corrupting the RBAC state.

Migration files should be **hermetic** — their behaviour must be fully determined by database state, not live application configuration.

---

### 3. Silent Crypto Failure on Downgrade Re-encryption

**File:** `a706a3320c56_use_argon2id_for_encryption_key.py`

```python
def reencrypt_with_pbkdf2hmac(argon2id_bundle: str) -> Optional[str]:
    try:
        ...
    except Exception as e:
        raise ValueError("Invalid Argon2id bundle") from e
```

And in the downgrade path:

```python
logger.warning("Downgrade skip (cannot re-encrypt Argon2 bundle): %s", e)
```

On downgrade, any token that cannot be re-encrypted from Argon2id back to PBKDF2HMAC is silently skipped. After downgrade, some tokens remain in Argon2id format while the application reverted to PBKDF2HMAC decryption. These tokens become permanently unreadable — OAuth credentials for gateways and A2A agents would be silently corrupted with no error surfaced.

---

### 4. Irreversible Phase 2 Backfill

**File:** `ba202ac1665f_migrate_user_roles_to_configurable_.py`

The `downgrade()` function explicitly acknowledges it cannot clean up backfill rows:

```python
def downgrade() -> None:
    """
    Note: Phase 2 backfill rows (granted_by=user_email) cannot be selectively
    removed without risking deletion of legitimate role assignments. They are
    left intact as valid team-role grants.
    """
```

Phase 2 inserts rows with `granted_by=user_email` (same signature as normal self-grants), so the downgrade has no way to distinguish injected rows from real ones. The database is left with extra RBAC assignments that should not exist at the previous revision level.

---

### 5. SQLite / PostgreSQL Asymmetry

**File:** `a3c38b6c2437_fix_a2a_agents_auth_value.py`

```python
def upgrade() -> None:
    ...
    if isinstance(current_type, sa.Text):
        return   # no-op on SQLite (already TEXT)
    if dialect == "postgresql":
        op.alter_column(...)   # actual change on PostgreSQL

def downgrade() -> None:
    ...
    if not isinstance(current_type, sa.Text):
        return   # no-op on SQLite always
    if dialect == "postgresql":
        op.alter_column(..., type_=postgresql.JSON())
```

On SQLite the `upgrade` is a no-op (correct — SQLite stores JSON as TEXT). However on `downgrade`, the check `not isinstance(current_type, sa.Text)` will always be `True` for SQLite (the column is `TEXT`), but then the `postgresql` guard prevents the `alter_column` from running. The result: the schema is unchanged on SQLite downgrade, which is technically correct, but the `alembic_version` table records the downgrade has completed. If the same database is later migrated to PostgreSQL, this silent divergence surfaces as hard-to-debug schema drift.

---

## Steps to Reproduce

```bash
# Set up a fresh environment
cp .env.example .env
make install-dev

# Upgrade all the way to head
cd mcpgateway && alembic upgrade head

# Attempt downgrade
alembic downgrade -1   # or 'base'
```

**Expected:** Clean rollback to previous schema and data state.

**Actual (various failure modes):**

- On `b77ca9d2de7e` base: all row data silently destroyed.
- On `ba202ac1665f`: downgrade uses wrong role names if env vars changed; Phase 2 rows not cleaned up.
- On `a706a3320c56`: tokens that fail re-encryption are silently left in Argon2id format, breaking auth after downgrade.

---

## Impact

| Scenario                            | Impact                                                |
| ----------------------------------- | ----------------------------------------------------- |
| Ops team rolls back a bad deploy    | Data loss, broken auth, corrupt RBAC                  |
| CI downgrade smoke test             | Silent pass masking real corruption                   |
| Helm rollback changing env vars     | Role names permanently mixed up                       |
| OAuth credential migration rollback | Gateways/A2A agents lose auth credentials permanently |

---

### Solution A — Add Mandatory Downgrade CI Tests with Roundtrip Validation (Immediate, Low-Risk)

**What:** Add a dedicated CI step that runs `alembic upgrade head && alembic downgrade base` on both SQLite and PostgreSQL for every PR that touches migrations. Use the existing test framework in `tests/migration/`.

**Why best for this class:** The current test suite covers `upgrade` but not `downgrade`. Roundtrip tests would have surfaced all five issues above at PR time.

**Scope:**

- Add `make test-migration-downgrade` Makefile target running `alembic upgrade head && alembic downgrade base`.
- Extend `tests/migration/test_docker_sqlite_migrations.py` `test_sequential_reverse_migrations` to assert no data corruption.
- Add a CI job step (e.g., in `.github/workflows/`) that runs the new target on every migration file change.
- For migrations documented as intentionally destructive (like `b77ca9d2de7e`), add an explicit `pytest.mark.xfail(reason="destructive: see rollback script")` marker so the test is known-failing rather than silently dangerous.

**Limitations:** Does not fix the underlying issues, only catches regressions early.

---

### Solution B — Decouple Settings Dependency via Upgrade-Time Metadata Snapshot (Structural Fix)

**What:** Introduce a `migration_metadata` table (or JSON sidecar per revision) where each data migration records the runtime configuration values it used during `upgrade`. The `downgrade` function reads from this snapshot instead of live `settings`.

**Why best for this class:** The root cause for `ba202ac1665f` and `b77ca9d2de7e` is that migrations are not hermetic. Migrations must be deterministic regardless of the current runtime environment. This pattern is the Alembic-recommended approach for environment-sensitive data migrations.

**Scope:**

```python
# In upgrade():
bind.execute(text("""
    INSERT INTO migration_metadata (revision, key, value)
    VALUES (:rev, 'default_admin_role', :val)
    ON CONFLICT (revision, key) DO UPDATE SET value = excluded.value
"""), {"rev": revision, "val": settings.default_admin_role})

# In downgrade():
row = bind.execute(text("""
    SELECT value FROM migration_metadata
    WHERE revision = :rev AND key = 'default_admin_role'
"""), {"rev": revision}).fetchone()
admin_role_at_upgrade_time = row[0] if row else settings.default_admin_role
```

- Create `migration_metadata` table in an early migration.
- Refactor `ba202ac1665f`, `b77ca9d2de7e`, and `a706a3320c56` to snapshot config at upgrade time and read the snapshot on downgrade.
- All future data migrations that reference `settings` must follow this pattern (add a linting rule or AGENTS.md guideline).

**Limitations:** Requires a new table and refactoring three existing migrations. Existing deployed databases will not have snapshots for already-applied migrations — provide a backfill script.

---

### Solution C — External Rollback Scripts for Inherently Irreversible Migrations + `NotImplementedError` Guards (Safety Net)

**What:** For migrations where the data transformation is genuinely irreversible (UUID refactor, crypto re-keying, non-distinguishable backfill rows), replace the silent/lossy `downgrade()` body with a `NotImplementedError` that points to a versioned external rollback script under `scripts/rollback/`. The rollback scripts handle pre-flight backup verification, backup restore, and post-restore validation.

**Why best for this class:** Pretending a destructive operation is a safe downgrade is worse than being explicit. Operators and CI should know upfront that a given migration requires a backup restore, not a schema rollback.

**Scope:**

```python
# b77ca9d2de7e_uuid_pk_and_slug_refactor.py
def downgrade() -> None:
    """This migration is irreversible via Alembic.

    All UUID primary key data will be lost. To revert, restore from a database
    backup taken before the upgrade and run:

        scripts/rollback/rollback_uuid_refactor.sh

    See: docs/docs/manage/database-rollback.md
    """
    raise NotImplementedError(
        "Migration b77ca9d2de7e is irreversible. "
        "Restore from backup using scripts/rollback/rollback_uuid_refactor.sh"
    )
```

- Create `scripts/rollback/` directory with documented shell/Python scripts for each irreversible migration.
- Each rollback script: (1) verifies a backup exists, (2) confirms operator intent, (3) restores the backup, (4) re-applies subsequent non-destructive migrations if needed.
- Add `docs/docs/manage/database-rollback.md` documenting the rollback procedure for each affected migration.
- Update `AGENTS.md` to require rollback scripts for any migration that cannot be safely reversed in-place.

**Limitations:** Requires discipline to write rollback scripts for every future irreversible migration. The `NotImplementedError` will break any CI step that runs `alembic downgrade` without the `--sql` flag or an exception handler — which is actually desirable, as it forces explicit decision-making.

---

## Recommended Approach

Apply all three solutions in sequence:

1. **Solution A first** (CI tests) — catches issues immediately with minimal code change.
2. **Solution C next** (guard irreversible migrations) — removes the silent data-loss footgun.
3. **Solution B last** (settings snapshot) — structural fix for env-dependent downgrade correctness.

---

## Additional Notes

- The `tests/migration/test_docker_sqlite_migrations.py` `test_sequential_reverse_migrations` test exists but it is gated behind Docker/Compose infrastructure that may not be running in every CI pipeline. Downgrade tests need to run in the standard `make test` pipeline.
- Several migration files use human-assigned sequential letter-number revision IDs (e.g., `g1a2b3c4d5e6`, `h2b3c4d5e6f7`) rather than Alembic-generated hex. While not a direct downgrade bug, this makes `alembic history` output and graph traversal harder to reason about and may cause ordering confusion in edge cases.
- `alembic downgrade base` should be treated as equivalent to dropping the schema. It should require the same operator confirmation as `DROP TABLE`.

---

## References

- `mcpgateway/alembic/versions/b77ca9d2de7e_uuid_pk_and_slug_refactor.py` — destructive base migration
- `mcpgateway/alembic/versions/ba202ac1665f_migrate_user_roles_to_configurable_.py` — env-dependent + irreversible backfill
- `mcpgateway/alembic/versions/a706a3320c56_use_argon2id_for_encryption_key.py` — silent crypto failure
- `mcpgateway/alembic/versions/a3c38b6c2437_fix_a2a_agents_auth_value.py` — SQLite/PG asymmetry
- `tests/migration/test_docker_sqlite_migrations.py` — existing (partial) downgrade test coverage
- Alembic docs: [Writing migration scripts — downgrade](https://alembic.sqlalchemy.org/en/latest/ops.html)

madhu-mohan-jaishankar:
## 🔗 Related Issue
Closes #4429 #4138

---

## 📝 Summary

Fixes unsafe Alembic downgrade behaviour and closes the migration testing gaps identified in both issues.

**From #4429 — Alembic downgrade is unsafe and environment-dependent:**

- **Root cause 2 fixed — runtime settings dependency in `ba202ac1665f`:** Introduces a `migration_metadata` table (new migration `f9a8b7c6d5e4`) that snapshots the four `DEFAULT_*_ROLE` env-var values at `upgrade()` time. `downgrade()` now reads from this snapshot instead of live `settings`, so the role-remap reversal is deterministic regardless of what env vars are set at rollback time. Falls back to live settings with a prominent warning on pre-existing databases that have no snapshot.

- **Root cause 4 fixed — irreversible Phase 2 backfill in `ba202ac1665f`:** Phase 2 backfill rows are now tagged with `migration_source = 'migration:ba202ac1665f'` in `user_roles`. `downgrade()` deletes only those rows, leaving all other grants untouched. Previously the docstring acknowledged the rows could not be cleaned up at all.

- **Root causes 1, 3, 5 not addressed — by design:** Migrations `b77ca9d2de7e` (UUID refactor, destructive base downgrade) and `a706a3320c56` (Argon2id, silent crypto skip on downgrade) shipped in 0.9.0. The practical rollback scenario this issue targets is 1.0.0 → 0.9.0, which never traverses those older migrations.

- **CI roundtrip validation added:** `scripts/ci/run_upgrade_validation.sh` now runs `run_sqlite_roundtrip` and `run_postgres_roundtrip` (upgrade base→head then downgrade head→base) in addition to the existing fresh-install checks. All 6 checks pass. The alembic-upgrade-validation workflow picks these up automatically on every PR touching migration files.

**From #4138 — Migration testing gaps:**

- **Gap 1 — Rollback tests under pytest:** `make migration-test-all` already runs `test_sequential_reverse_migrations` (parametrized via `reverse_version_pair` in `conftest.py`). Added `make migration-test-rollback` as a dedicated target that runs `-k 'reverse or rollback'` on both SQLite and PostgreSQL test files, then runs the full upgrade/downgrade roundtrip harness.

- **Gap 2 — Cross-DB schema validation:** Added `tests/migration/test_cross_db_schema_consistency.py` — 5 tests that boot the target image with both SQLite and PostgreSQL, introspect schemas via SQLAlchemy Inspector, normalise engine-specific type names, and assert tables/columns/types/PKs/FKs match. All 5 pass (67 tables each engine). Also added `make migration-test-cross-db` target.

- **Gap 3 — CI coverage decision documented:** Added a comment block to `.github/workflows/alembic-upgrade-validation.yml` explaining why the full suite (rollback, perf, cross-DB) does not run on every PR (runtime cost, risk profile, cold Docker pull cost) and pointing to the three `make` targets for manual/pre-release runs.

- **AGENTS.md guideline:** Added "Hermetic Downgrade: Config Snapshot Pattern" section documenting the rule and providing a copy-paste template for any future migration whose `downgrade()` reads from `settings`.

---

## 🏷️ Type of Change
- [x] Bug fix
- [x] Feature / Enhancement
- [x] Documentation
- [ ] Refactor

---

## 🧪 Verification

| Check | Command | Status |
|---|---|---|
| Lint suite | `make lint` | Not run (no logic changes outside migrations/tests) |
| Unit tests | `make test` | Not run (migration tests require Docker) |
| CI roundtrip (SQLite fresh + upgrade + roundtrip) | `bash scripts/ci/run_upgrade_validation.sh` | ✅ 6/6 pass |
| Cross-DB consistency | `UPGRADE_TARGET_IMAGE=... pytest tests/migration/test_cross_db_schema_consistency.py` | ✅ 5/5 pass |
| Alembic single head | `cd mcpgateway && alembic heads` | ✅ `bb43712cae28` |

---

## ✅ Checklist
- [x] Documentation updated (AGENTS.md, workflow comment)

---

## 📓 Notes

- The `migration_metadata` table is inserted at `a31c6ffc2239 → f9a8b7c6d5e4 → ba202ac1665f` in the chain. Databases upgraded before this PR have no snapshot rows for `ba202ac1665f` — downgrade on those will fall back to live settings with a warning (same behaviour as before, now surfaced explicitly rather than silently).
- `make migration-test-cross-db` and `make migration-test-rollback` require Docker and the target image to be built locally (`docker build -f Containerfile.lite -t mcpgateway/mcpgateway:upgrade-ci-test .`). Override the image with `UPGRADE_TARGET_IMAGE=mcpgateway/mcpgateway:upgrade-ci-test make migration-test-cross-db`.

### DIFF
--- a/.github/workflows/alembic-upgrade-validation.yml
+++ b/.github/workflows/alembic-upgrade-validation.yml
@@ -1,3 +1,26 @@
+# Migration CI coverage decision
+# --------------------------------
+# This workflow runs on every PR that touches migration files and covers:
+#   • SQLite fresh install (HEAD schema from scratch)
+#   • PostgreSQL fresh install
+#   • SQLite upgrade roundtrip (base-image → HEAD → downgrade back to base)
+#   • PostgreSQL upgrade roundtrip
+#
+# The full migration suite (rollback tests, performance benchmarks, and
+# cross-DB schema consistency) is NOT run on every PR because:
+#   • Runtime cost: the full suite can exceed 30 minutes with cold Docker pulls
+#   • Risk profile: rollback and perf tests rarely regress on a typical PR;
+#     the upgrade/roundtrip harness here already catches broken migrations and
+#     stranded Alembic heads — the most common PR-time failure modes
+#   • Cross-DB consistency: requires two live engine containers and is better
+#     suited to a controlled environment
+#
+# Those slower checks are available as:
+#   make migration-test-rollback   — downgrade/reverse pytest + roundtrip
+#   make migration-test-cross-db   — cross-DB schema consistency
+#   make migration-test-all        — full suite
+# Run them manually before release or via workflow_dispatch on this workflow.
+
 name: Alembic Upgrade Validation
 
 on:
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-04-30T22:30:45Z",
+  "generated_at": "2026-05-01T10:39:39Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -4569,16 +4569,6 @@
         "is_secret": false,
         "is_verified": false,
         "line_number": 19,
-        "type": "Hex High Entropy String",
-        "verified_result": null
-      }
-    ],
-    "mcpgateway/alembic/versions/ba202ac1665f_migrate_user_roles_to_configurable_.py": [
-      {
-        "hashed_secret": "10b36c8089b5791d82f0700828fd23a9c782101f",
-        "is_secret": false,
-        "is_verified": false,
-        "line_number": 44,
         "type": "Hex High Entropy String",
         "verified_result": null
       }
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -5778,7 +5778,7 @@
         "hashed_secret": "1d193899f802762003a56b1cbcfeb55f2b04d61b",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 448,
+        "line_number": 639,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -5786,7 +5786,7 @@
         "hashed_secret": "dcdf24a1d4440d5ebe319bf804f96c81e08350f9",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 455,
+        "line_number": 646,
         "type": "Basic Auth Credentials",
         "verified_result": null
       }
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -302,6 +302,63 @@ make test
 - **"Multiple heads are present"**: Your `down_revision` points to wrong parent. Fix by updating to actual current head.
 - **"Target database is not up to date"**: Run `alembic upgrade head` first.
 
+### Hermetic Downgrade: Config Snapshot Pattern
+
+Any migration whose `downgrade()` logic depends on runtime configuration (i.e., reads from
+`mcpgateway.config.settings`) **must** snapshot those values into `migration_metadata` during
+`upgrade()` and read them back during `downgrade()`. This makes the migration hermetic —
+its behaviour is determined by database state, not the current environment.
+
+```python
+from mcpgateway.config import settings
+from sqlalchemy import inspect, text
+
+REVISION = "your_revision_id"
+
+def upgrade() -> None:
+    bind = op.get_bind()
+    # ... schema changes ...
+
+    # Snapshot any settings values used in downgrade
+    if "migration_metadata" in inspect(bind).get_table_names():
+        bind.execute(
+            text(
+                "INSERT INTO migration_metadata (revision, key, value, created_at) "
+                "VALUES (:rev, :key, :val, CURRENT_TIMESTAMP) "
+                "ON CONFLICT (revision, key) DO UPDATE SET value = excluded.value"
+            ),
+            {"rev": REVISION, "key": "some_setting", "val": settings.some_setting},
+        )
+
+def downgrade() -> None:
+    bind = op.get_bind()
+
+    # Read from snapshot; fall back to live settings only if table is absent
+    # (pre-existing DB that was upgraded before this pattern was introduced)
+    cfg = {}
+    if "migration_metadata" in inspect(bind).get_table_names():
+        rows = bind.execute(
+            text("SELECT key, value FROM migration_metadata WHERE revision = :rev"),
+            {"rev": REVISION},
+        ).all()
+        cfg = {r[0]: r[1] for r in rows}
+
+    some_setting = cfg.get("some_setting") or settings.some_setting
+
+    # ... use some_setting for downgrade logic ...
+
+    # Clean up snapshot rows
+    if "migration_metadata" in inspect(bind).get_table_names():
+        bind.execute(
+            text("DELETE FROM migration_metadata WHERE revision = :rev"),
+            {"rev": REVISION},
+        )
+```
+
+**Rule:** If your migration imports `settings` and uses it in `downgrade()`, you must follow this
+pattern. Migrations that only use `settings` in `upgrade()` (e.g., for seeding initial data) are
+exempt.
+
 ## Coding Standards
 
 - **Python >= 3.11** with type hints; strict mypy
--- a/Makefile
+++ b/Makefile
@@ -8450,25 +8450,25 @@ migration-test-performance:               ## Run migration performance benchmark
 	@echo "✅ Performance tests complete!"
 
 migration-test-rollback:                  ## Run only downgrade/reverse migration tests (pytest + roundtrip)
-        @echo "⏪ Running reverse migration (downgrade) tests..."
-        @test -d "$(VENV_DIR)" || $(MAKE) venv
-        @/bin/bash -c "source $(VENV_DIR)/bin/activate && \
-                pytest $(MIGRATION_TEST_DIR)/test_docker_sqlite_migrations.py \
-                       $(MIGRATION_TEST_DIR)/test_compose_postgres_migrations.py \
-                -k 'reverse or rollback' \
-                -v --tb=short --log-cli-level=INFO"
-        @echo "🔄 Running upgrade/downgrade roundtrip validation..."
-        @BASE_IMAGE=$(UPGRADE_BASE_IMAGE) TARGET_IMAGE=$(UPGRADE_TARGET_IMAGE) bash scripts/ci/run_upgrade_validation.sh
-        @echo "✅ Rollback tests complete!"
+	@echo "⏪ Running reverse migration (downgrade) tests..."
+	@test -d "$(VENV_DIR)" || $(MAKE) venv
+	@/bin/bash -c "source $(VENV_DIR)/bin/activate && \
+	        pytest $(MIGRATION_TEST_DIR)/test_docker_sqlite_migrations.py \
+	               $(MIGRATION_TEST_DIR)/test_compose_postgres_migrations.py \
+	        -k 'reverse or rollback' \
+	        -v --tb=short --log-cli-level=INFO"
+	@echo "🔄 Running upgrade/downgrade roundtrip validation..."
+	@BASE_IMAGE=$(UPGRADE_BASE_IMAGE) TARGET_IMAGE=$(UPGRADE_TARGET_IMAGE) bash scripts/ci/run_upgrade_validation.sh
+	@echo "✅ Rollback tests complete!"
 
 migration-test-cross-db:                  ## Run cross-database schema consistency test
-        @echo "🔀 Running cross-database schema consistency test..."
-        @test -d "$(VENV_DIR)" || $(MAKE) venv
-        @/bin/bash -c "source $(VENV_DIR)/bin/activate && \
-                UPGRADE_TARGET_IMAGE=$(UPGRADE_TARGET_IMAGE) \
-                pytest $(MIGRATION_TEST_DIR)/test_cross_db_schema_consistency.py \
-                -v --tb=short --log-cli-level=INFO"
-        @echo "✅ Cross-database schema consistency check complete!"
+	@echo "🔀 Running cross-database schema consistency test..."
+	@test -d "$(VENV_DIR)" || $(MAKE) venv
+	@/bin/bash -c "source $(VENV_DIR)/bin/activate && \
+	        UPGRADE_TARGET_IMAGE=$(UPGRADE_TARGET_IMAGE) \
+	        pytest $(MIGRATION_TEST_DIR)/test_cross_db_schema_consistency.py \
+	        -v --tb=short --log-cli-level=INFO"
+	@echo "✅ Cross-database schema consistency check complete!"
 	@echo "📦 Pulling required container images..."
 	@if command -v docker >/dev/null 2>&1; then \
 		for version in $(MIGRATION_VERSIONS); do \
--- a/Makefile
+++ b/Makefile
@@ -8384,11 +8384,12 @@ fuzz-all: fuzz-hypothesis fuzz-atheris fuzz-api fuzz-security fuzz-report  ## �
 # help: migration-test-sqlite    - Run SQLite container migration tests only
 # help: migration-test-postgres  - Run PostgreSQL compose migration tests only
 # help: migration-test-performance - Run migration performance benchmarking
+# help: migration-test-rollback  - Run only downgrade/reverse migration tests (pytest + roundtrip)
 # help: migration-setup          - Setup migration test environment
 # help: migration-cleanup        - Clean up migration test containers and volumes
 # help: migration-debug          - Debug migration test failures with diagnostic info
 # help: migration-status         - Show current version configuration and supported versions
-# help: upgrade-validate         - Validate fresh + upgrade DB startup paths (SQLite + PostgreSQL)
+# help: upgrade-validate         - Validate fresh + upgrade + roundtrip DB paths (SQLite + PostgreSQL)
 
 # Migration testing configuration
 MIGRATION_TEST_DIR := tests/migration
@@ -8400,7 +8401,7 @@ UPGRADE_TARGET_IMAGE ?= mcpgateway/mcpgateway:latest
 MIGRATION_VERSIONS := $(shell cd $(MIGRATION_TEST_DIR) && python3 -c "from version_config import get_supported_versions; print(' '.join(get_supported_versions()))" 2>/dev/null || echo "0.5.0 0.8.0 0.9.0 latest")
 
 .PHONY: migration-test-all migration-test-sqlite migration-test-postgres migration-test-performance \
-        migration-setup migration-cleanup migration-debug migration-status upgrade-validate
+        migration-test-rollback migration-setup migration-cleanup migration-debug migration-status upgrade-validate
 
 migration-test-all: migration-setup        ## Run comprehensive migration test suite (SQLite + PostgreSQL)
 	@echo "🚀 Running comprehensive migration tests..."
@@ -8447,12 +8448,17 @@ migration-test-performance:               ## Run migration performance benchmark
 		-v --tb=short --log-cli-level=INFO"
 	@echo "✅ Performance tests complete!"
 
-.PHONY: migration-setup
-migration-setup:                           ## Setup migration test environment
-	@echo "🔧 Setting up migration test environment..."
-	@test -d "$(VENV_DIR)" || $(MAKE) venv
-	@mkdir -p $(MIGRATION_REPORTS_DIR)
-	@mkdir -p $(MIGRATION_TEST_DIR)/logs
+migration-test-rollback:                  ## Run only downgrade/reverse migration tests (pytest + roundtrip)
+        @echo "⏪ Running reverse migration (downgrade) tests..."
+        @test -d "$(VENV_DIR)" || $(MAKE) venv
+        @/bin/bash -c "source $(VENV_DIR)/bin/activate && \
+                pytest $(MIGRATION_TEST_DIR)/test_docker_sqlite_migrations.py \
+                       $(MIGRATION_TEST_DIR)/test_compose_postgres_migrations.py \
+                -k 'reverse or rollback' \
+                -v --tb=short --log-cli-level=INFO"
+        @echo "🔄 Running upgrade/downgrade roundtrip validation..."
+        @BASE_IMAGE=$(UPGRADE_BASE_IMAGE) TARGET_IMAGE=$(UPGRADE_TARGET_IMAGE) bash scripts/ci/run_upgrade_validation.sh
+        @echo "✅ Rollback tests complete!"
 	@echo "📦 Pulling required container images..."
 	@if command -v docker >/dev/null 2>&1; then \
 		for version in $(MIGRATION_VERSIONS); do \
@@ -8514,7 +8520,7 @@ migration-status:                          ## Show current version configuration
 		cd $(MIGRATION_TEST_DIR) && python3 version_status.py"
 
 .PHONY: upgrade-validate
-upgrade-validate:                         ## Validate fresh + upgrade DB startup paths (SQLite + PostgreSQL)
+upgrade-validate:                         ## Validate fresh + upgrade + roundtrip DB paths (SQLite + PostgreSQL)
 	@echo "🔄 Running upgrade validation harness..."
 	@echo "  Base image:   $(UPGRADE_BASE_IMAGE)"
 	@echo "  Target image: $(UPGRADE_TARGET_IMAGE)"
--- a/Makefile
+++ b/Makefile
@@ -8469,6 +8469,11 @@ migration-test-cross-db:                  ## Run cross-database schema consisten
 	        pytest $(MIGRATION_TEST_DIR)/test_cross_db_schema_consistency.py \
 	        -v --tb=short --log-cli-level=INFO"
 	@echo "✅ Cross-database schema consistency check complete!"
+
+migration-setup:                          ## Setup migration test environment
+	@echo "🔧 Setting up migration test environment..."
+	@mkdir -p $(MIGRATION_REPORTS_DIR)
+	@mkdir -p $(MIGRATION_TEST_DIR)/logs
 	@echo "📦 Pulling required container images..."
 	@if command -v docker >/dev/null 2>&1; then \
 		for version in $(MIGRATION_VERSIONS); do \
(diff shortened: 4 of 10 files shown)


msureshkumar88 (changes requested):
Great work on this PR — the core design (hermetic downgrade via `migration_metadata`, tagging Phase 2 backfill rows, roundtrip CI validation) is exactly the right approach to the issues raised in #4429 and #4138. The cross-DB consistency test is a nice addition. A few things need fixing before this can merge.

---

## Blocking issues

### H1 — `migration-setup` target became a silent no-op (Makefile)

The `migration-setup:` recipe body was converted into `migration-test-rollback:`, but `migration-test-all: migration-setup` still lists it as a prerequisite. GNU Make silently treats a `.PHONY` target with no recipe as a no-op, so `make migration-test-all` now runs without creating `$(MIGRATION_REPORTS_DIR)` or `$(MIGRATION_TEST_DIR)/logs`, and without pulling container images.

There's a second wrinkle: the image-pull block ended up at the **bottom** of `migration-test-cross-db` — after the test runs. Images are being pulled after the test that needs them, which means a cold runner will fail.

**Fix:** Restore a `migration-setup` recipe that creates the required directories and pulls images, and remove the misplaced image-pull stanza from `migration-test-cross-db`.

---

### H2 — `_delete_config_snapshot` not in a `finally` block

In `ba202ac1665f.downgrade()`, `_delete_config_snapshot(bind, revision)` is the last statement. If an exception is raised during Phase 1 role remap (`_migrate_role()` calls, lines ~382–408) the snapshot rows are never cleaned up. A subsequent downgrade retry would find the stale snapshot and use it as if the prior attempt had succeeded.

```python
# Suggested fix
try:
    # ... Phase 1 role remap reversal ...
finally:
    _delete_config_snapshot(bind, revision)
```

The snapshot is only meaningful within a single downgrade execution, so cleanup should be unconditional.

---

### H3 — `f9a8b7c6d5e4.downgrade()` drops `migration_metadata` even when rows remain

The downgrade for the metadata table migration logs a warning when rows are present, then drops the table anyway. The scenario where this matters: an operator downgrades only part of the chain (stops at `f9a8b7c6d5e4` without first downgrading `ba202ac1665f`). The config snapshot for `ba202ac1665f` is silently destroyed, making any future attempt to downgrade `ba202ac1665f` fall back to live settings — exactly the non-hermetic behaviour this PR is trying to fix.

Worth considering: should the downgrade raise instead of proceeding when rows remain?

```python
if count:
    raise RuntimeError(
        f"Cannot drop migration_metadata — {count} snapshot row(s) remain. "
        "Downgrade dependent migrations first."
    )
```

---

## Suggestions (non-blocking, but worth addressing)

**M1 — `grep -F ''` is a no-op in `assert_version_present` (run_upgrade_validation.sh:290)**
`grep -F ''` matches every line — it's an accidental no-op. The function happens to be correct today because the real check is the subsequent `grep -qxF`, but the dead pipe stage is confusing. Removing it would also make the intent clearer.

**M2 — Hardcoded test passwords**
`upgrade-test-password` and `schema-test-pw` appear in three functions and the pytest fixture. Using `openssl rand -hex 16` per run is a small change that removes a permanently searchable secret string from the source file and keeps secrets-scanning signal meaningful.

**M3 — Warning-level messages via `print()` rather than `logging.warning()`**
The "No config snapshot found" fall-back warning and the "migration_metadata table not found; skipping snapshot" message are operator-critical. In environments where stdout is redirected (systemd, log forwarder, structured logging), bare `print()` output is silently dropped. Routing these through `logging.warning()` or at minimum `sys.stderr` would make them reliably visible.

**M4 — `migration_metadata` has no ORM model in `db.py`**
Without an ORM model, `alembic revision --autogenerate` will always propose to drop the table. A minimal model (or an explicit autogenerate exclusion in `env.py`) would prevent accidental future migrations from destroying the snapshot infrastructure.

**M5 — N+1 INSERT in Phase 2 backfill**
The loop over `members_without_roles` issues one INSERT per member. On large deployments this will be slow. Using `executemany()` or a bulk INSERT would be more robust.

---

## Testing gaps worth tracking

- No unit tests for `_snapshot_config` / `_read_config_snapshot` / `_delete_config_snapshot` and their fallback paths (table absent, no matching revision).
- No test asserting that rows with other `migration_source` values (NULL, `'sso'`, `'manual'`) are **not** deleted during downgrade — this is the key correctness claim of the PR and it has no regression test.
- `make migration-test-rollback` invokes the full 6-check roundtrip harness, which makes it take as long as `upgrade-validate`. The target name implies it only runs rollback/downgrade paths; the scope mismatch will confuse contributors.

---

## Minor

- `_META_KEYS` (ba202ac1665f, line 56–57) is defined but never referenced — the keys are passed as dict literals instead. Either wire it up or remove it.
- `import re` inside the inner loop of `_INTROSPECT_SCRIPT` (test_cross_db_schema_consistency.py:50) and `import urllib.request` inside `_wait_health()` (line 168) should be at the top of their respective scopes.

---

Happy to discuss any of these. The H1–H3 fixes are straightforward and the rest could be addressed in a follow-up if preferred.

msureshkumar88:
## Functional verification results

Ran an automated functional test suite against the PR branch to verify the core claims independently of the review comments. All 19 tests passed.

### What was tested

Each test uses a temporary `uv` venv + SQLite DB, isolated from the rest of the migration chain to avoid interference from pre-existing downgrade issues in unrelated migrations. All artifacts are cleaned up on exit.

| # | Test | Result |
|---|------|--------|
| 1 | `migration_metadata` table created by `f9a8b7c6d5e4` with composite PK `(revision, key)` | ✅ PASS |
| 2 | 4 config snapshot rows written for `ba202ac1665f` at upgrade time | ✅ PASS |
| 3 | `downgrade()` uses snapshot, ignores a deliberately wrong `DEFAULT_ADMIN_ROLE` env var | ✅ PASS |
| 4 | `user_roles.migration_source` column added by `v1a2b3c4d5e6` | ✅ PASS |
| 5 | Only `migration:ba202ac1665f`-tagged rows deleted on downgrade; a canary row with `migration_source=manual` survives | ✅ PASS |
| 6 | When snapshot rows are absent (simulated pre-fix DB), downgrade emits the fall-back warning | ✅ PASS |
| 7 | Full roundtrip — upgrade to `ba202ac1665f` → downgrade to `a31c6ffc2239` → upgrade again — ends at correct revision with a fresh snapshot | ✅ PASS |
| 8 | Snapshot captures the env var value active at upgrade time (`custom_admin_role`); downgrade ignores a `wrong_value` env var set at rollback time | ✅ PASS |
| 9 | Running `upgrade head` a second time is a no-op — no errors, no duplicate snapshot rows | ✅ PASS |
| 10 | Exactly one alembic head after applying PR migrations: `bb43712cae28` | ✅ PASS |

### What this means

The core PR claims — hermetic downgrade via `migration_metadata`, tagged Phase 2 backfill rows, and correct selective cleanup on downgrade — all work as described. The three blocking items in the earlier review (H1 Makefile no-op, H2 `finally` block, H3 table-drop guard) are hardening issues on top of this working foundation, not correctness gaps in the main logic.

madhu-mohan-jaishankar:
> Great work on this PR — the core design (hermetic downgrade via `migration_metadata`, tagging Phase 2 backfill rows, roundtrip CI validation) is exactly the right approach to the issues raised in #4429 and #4138. The cross-DB consistency test is a nice addition. A few things need fixing before this can merge.
> 
> ## Blocking issues
> ### H1 — `migration-setup` target became a silent no-op (Makefile)
> The `migration-setup:` recipe body was converted into `migration-test-rollback:`, but `migration-test-all: migration-setup` still lists it as a prerequisite. GNU Make silently treats a `.PHONY` target with no recipe as a no-op, so `make migration-test-all` now runs without creating `$(MIGRATION_REPORTS_DIR)` or `$(MIGRATION_TEST_DIR)/logs`, and without pulling container images.
> 
> There's a second wrinkle: the image-pull block ended up at the **bottom** of `migration-test-cross-db` — after the test runs. Images are being pulled after the test that needs them, which means a cold runner will fail.
> 
> **Fix:** Restore a `migration-setup` recipe that creates the required directories and pulls images, and remove the misplaced image-pull stanza from `migration-test-cross-db`.
> 
> ### H2 — `_delete_config_snapshot` not in a `finally` block
> In `ba202ac1665f.downgrade()`, `_delete_config_snapshot(bind, revision)` is the last statement. If an exception is raised during Phase 1 role remap (`_migrate_role()` calls, lines ~382–408) the snapshot rows are never cleaned up. A subsequent downgrade retry would find the stale snapshot and use it as if the prior attempt had succeeded.
> 
> ```python
> # Suggested fix
> try:
>     # ... Phase 1 role remap reversal ...
> finally:
>     _delete_config_snapshot(bind, revision)
> ```
> 
> The snapshot is only meaningful within a single downgrade execution, so cleanup should be unconditional.
> 
> ### H3 — `f9a8b7c6d5e4.downgrade()` drops `migration_metadata` even when rows remain
> The downgrade for the metadata table migration logs a warning when rows are present, then drops the table anyway. The scenario where this matters: an operator downgrades only part of the chain (stops at `f9a8b7c6d5e4` without first downgrading `ba202ac1665f`). The config snapshot for `ba202ac1665f` is silently destroyed, making any future attempt to downgrade `ba202ac1665f` fall back to live settings — exactly the non-hermetic behaviour this PR is trying to fix.
> 
> Worth considering: should the downgrade raise instead of proceeding when rows remain?
> 
> ```python
> if count:
>     raise RuntimeError(
>         f"Cannot drop migration_metadata — {count} snapshot row(s) remain. "
>         "Downgrade dependent migrations first."
>     )
> ```
> 
> ## Suggestions (non-blocking, but worth addressing)
> **M1 — `grep -F ''` is a no-op in `assert_version_present` (run_upgrade_validation.sh:290)** `grep -F ''` matches every line — it's an accidental no-op. The function happens to be correct today because the real check is the subsequent `grep -qxF`, but the dead pipe stage is confusing. Removing it would also make the intent clearer.
> 
> **M2 — Hardcoded test passwords** `upgrade-test-password` and `schema-test-pw` appear in three functions and the pytest fixture. Using `openssl rand -hex 16` per run is a small change that removes a permanently searchable secret string from the source file and keeps secrets-scanning signal meaningful.
> 
> **M3 — Warning-level messages via `print()` rather than `logging.warning()`** The "No config snapshot found" fall-back warning and the "migration_metadata table not found; skipping snapshot" message are operator-critical. In environments where stdout is redirected (systemd, log forwarder, structured logging), bare `print()` output is silently dropped. Routing these through `logging.warning()` or at minimum `sys.stderr` would make them reliably visible.
> 
> **M4 — `migration_metadata` has no ORM model in `db.py`** Without an ORM model, `alembic revision --autogenerate` will always propose to drop the table. A minimal model (or an explicit autogenerate exclusion in `env.py`) would prevent accidental future migrations from destroying the snapshot infrastructure.
> 
> **M5 — N+1 INSERT in Phase 2 backfill** The loop over `members_without_roles` issues one INSERT per member. On large deployments this will be slow. Using `executemany()` or a bulk INSERT would be more robust.
> 
> ## Testing gaps worth tracking
> * No unit tests for `_snapshot_config` / `_read_config_snapshot` / `_delete_config_snapshot` and their fallback paths (table absent, no matching revision).
> * No test asserting that rows with other `migration_source` values (NULL, `'sso'`, `'manual'`) are **not** deleted during downgrade — this is the key correctness claim of the PR and it has no regression test.
> * `make migration-test-rollback` invokes the full 6-check roundtrip harness, which makes it take as long as `upgrade-validate`. The target name implies it only runs rollback/downgrade paths; the scope mismatch will confuse contributors.
> 
> ## Minor
> * `_META_KEYS` (ba202ac1665f, line 56–57) is defined but never referenced — the keys are passed as dict literals instead. Either wire it up or remove it.
> * `import re` inside the inner loop of `_INTROSPECT_SCRIPT` (test_cross_db_schema_consistency.py:50) and `import urllib.request` inside `_wait_health()` (line 168) should be at the top of their respective scopes.
> 
> Happy to discuss any of these. The H1–H3 fixes are straightforward and the rest could be addressed in a follow-up if preferred.

@msureshkumar88 address blocking issues. please review.

msureshkumar88 (dismissed):
All three blocking items are resolved — approving.

**H1 fixed** — `migration-setup` recipe is restored with the `mkdir -p` and image-pull stanzas; the misplaced image-pull block is no longer at the bottom of `migration-test-cross-db`.

**H2 fixed** — `_delete_config_snapshot` is now inside a `try/finally` block in `ba202ac1665f.downgrade()`, so snapshot cleanup is unconditional even if Phase 1 role-remap raises.

**H3 fixed** — `f9a8b7c6d5e4.downgrade()` now raises `RuntimeError` when snapshot rows remain instead of silently dropping the table; operators must downgrade dependent migrations first.

Two non-blocking suggestions were also addressed voluntarily:
- **M3** — critical fall-back warnings now route through `logger.warning()` rather than bare `print()`.
- **M4** — `MigrationMetadata` ORM model added to `db.py`, preventing `alembic revision --autogenerate` from proposing to drop the table in future.

Items still open (all non-blocking, fine to track as follow-up):
- **M1** — `grep -F ''` no-op in `assert_version_present` (run_upgrade_validation.sh:290)
- **M2** — hardcoded test passwords (`upgrade-test-password`, `schema-test-pw`) in the CI script and test fixture
- **M5** — N+1 INSERT in Phase 2 backfill loop
- Minor: `_META_KEYS` tuple defined but never referenced in `ba202ac1665f`
- Minor: `import re` inside the inner loop and `import urllib.request` inside `_wait_health()` should move to module scope
- Testing: no unit tests for `_snapshot_config`/`_read_config_snapshot`/`_delete_config_snapshot` fallback paths; no regression test asserting non-tagged rows survive downgrade

None of the above affect correctness of the migration logic. The core design is sound.

brian-hussey (approved):
Approving based on others approval.

PR #4489, opened 2026-04-28
[Security][Pen Testing]Restrict outbound requests from the gateway test endpoint to approved hosts

Nayana-R-Gowda:

 Security Fix Summary - ICA_ContextForgeICACF-14   

                                                                                                                                                                             
  Fixed the /admin/gateways/test endpoint to prevent it from being exploited as an open proxy by implementing allowlist validation, FQDN normalization, and unconditional    
  private IP blocking.                                                                                                                                                       
                           

Changes by File
                 
  1. mcpgateway/config.py (Configuration)
                                                                                                                                                                             
  Added 2 new settings:                                                                                                                                                      
  gateway_test_allowed_hosts: List[str] = []                                                                                                                                 
  # Allowlist of host patterns (e.g., "*.example.com", "trusted.com")                                                                                                        
  # Supports exact hostnames and wildcard subdomains                 
                                                                                                                                                                             
  gateway_test_allow_registered_only: bool = True                                                                                                                            
  # When true: only allow testing registered gateway URLs from database                                                                                                      
  # When false: use gateway_test_allowed_hosts patterns                                                                                                                      
  Purpose: Enable flexible allowlist configuration with secure defaults                                                                                                      
   
  ---                                                                                                                                                                        
  2. mcpgateway/common/validators.py (Security Validation)
                                                                                                                                                                             
  Added 1 new method (~150 lines):
  SecurityValidator.validate_gateway_test_url(value, allowed_hosts, field_name)                                                                                              
  Features implemented:
  - ✅ FQDN normalization - Strips trailing dots (e.g., evil.com. → evil.com)                                                                                                
  - ✅ Allowlist enforcement - Exact match + wildcard subdomain support      
  - ✅ Private IP blocking - RFC 1918 (10.0.0.0/8, 172.16.0.0/12, 192.168.0.0/16)                                                                                            
  - ✅ Loopback blocking - 127.0.0.0/8 and localhost                                                                                                                         
  - ✅ Link-local blocking - 169.254.0.0/16 (cloud metadata services)                                                                                                        
  - ✅ DNS rebinding protection - Resolves hostnames and checks resolved IPs                                                                                                 
  - ✅ Generic error messages - Returns "Gateway URL is not allowed" without internal details                                                                                
                                                                                                                                                                             
  Purpose: Core security validation logic                                                                                                                                    
                                                                                                                                                                             
  ---                                                                                                                                                                        
  3. mcpgateway/admin.py (Endpoint Protection)                                                                                                                               
                                                                                                                                                                             
  Modified 1 function: admin_test_gateway()
                                                                                                                                                                             
  Changes made (~30 lines):                                                                                                                                                  
  # OLD CODE (before):                                                                                                                                                       
  validated_base_url = SecurityValidator.validate_url(str(request.base_url), "Gateway test URL")                                                                             
                                                                                                                                                                             
  # NEW CODE (after):                                                                                                                                                        
  # 1. Build allowlist from registered gateways OR config                                                                                                                    
  allowed_hosts = []                                                                                                                                                         
  if settings.gateway_test_allow_registered_only:                                                                                                                            
      # Query database for registered gateway hostnames
      allowed_hosts = [extract hostnames from registered gateways]                                                                                                           
  else:                                                                                                                                                                      
      # Use configured patterns                                                                                                                                              
      allowed_hosts = settings.gateway_test_allowed_hosts                                                                                                                    
                                                                                                                                                                             
  # 2. Validate with allowlist enforcement                                                                                                                                   
  validated_base_url = SecurityValidator.validate_gateway_test_url(                                                                                                          
      str(request.base_url), allowed_hosts, "Gateway test URL"                                                                                                               
  )                                                                                                                                                                          
                                                                                                                                                                             
  # 3. Enhanced error logging (logs actual error but returns generic message)                                                                                                
                                                                                                                                                                             
  Purpose: Enforce allowlist before any outbound HTTP requests
                                                                                                                                                                             
  ---             
  4. tests/unit/mcpgateway/validation/test_validators_advanced.py (Unit Tests)                                                                                               
                                                                                                                                                                             
  Added 1 test class: TestGatewayTestUrlValidation with 14 tests
                                                                                                                                                                             
                                                                                                               
  Purpose: Comprehensive validation logic testing                                                                                                                            
                                                                                                                                                                             
  ---                                                                                                                                                                        
  5. tests/e2e/test_admin_apis.py (Integration Tests)                                                                                                                        
                                                                                                                                                                             
  Added 2 test methods:
                                                                                                                                                                             
  Test 1: test_gateway_test_endpoint_allowlist_enforcement()                                                                                                                 
  - Tests allowlist mode with configured patterns                                                                                                                            
  - Verifies rejection of non-allowlisted URLs                                                                                                                               
  - Verifies trailing dot bypass is blocked   
  - Verifies private/loopback/link-local IPs are blocked                                                                                                                     
                                                                                                                                                                             
  Test 2: test_gateway_test_endpoint_registered_only_mode()                                                                                                                  
  - Tests registered-only mode (default)                                                                                                                                     
  - Verifies empty allowlist rejects all requests                                                                                                                            
                                                                                                                                                                             
  Purpose: End-to-end endpoint behavior validation                                                                                                                           
                                                                                                                                                                             
  ---                                                                                                                                                                        
  6. .env.example (Documentation)                                                                                                                                            
                                                                                                                                                                             
  Added configuration documentation (~15 lines):
  # Gateway Test Endpoint Security                                                                                                                                           
  # GATEWAY_TEST_ALLOW_REGISTERED_ONLY=true                                                                                                                                  
  # GATEWAY_TEST_ALLOWED_HOSTS=["api.example.com","*.partner.com"]
                                                                                                                                                                             
  Includes:       
  - Configuration options explanation                                                                                                                                        
  - Security warnings                                                                                                                                                        
  - Usage examples                                                                                                                                                           
  - Note about unconditional private IP blocking                                                                                                                             
                                                
  Purpose: User documentation for new security settings

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -123,17 +123,25 @@ ALLOW_PUBLIC_VISIBILITY=true
 # URLs that cannot be resolved are rejected
 # SSRF_DNS_FAIL_CLOSED=true
 
-# Gateway Test Endpoint Allowlist (ICACF-15)
-# Restricts /admin/gateways/test endpoint to approved hosts only.
-# Empty list = use SSRF protection only (backwards compatible)
-# Supports wildcards: ["*.example.com", "api.specific.com"]
-#   - *.example.com matches subdomains ONLY (e.g., api.example.com, test.api.example.com)
-#   - It does NOT match the base domain itself (example.com)
-#   - To allow both, list both: ["*.example.com", "example.com"]
-# RECOMMENDED: Set explicit allowlist in production to restrict outbound requests.
-# Empty list = SSRF protection only (less restrictive, not recommended for production).
-# Example: GATEWAY_TEST_ALLOWED_HOSTS=["api.approved.com", "*.internal.ibm.com"]
+# Gateway Test Endpoint Security
+#
+# The /admin/gateways/test endpoint allows testing gateway connectivity. To prevent
+# using this endpoint as an open proxy (ICA_ContextForgeICACF-14), it enforces an
+# allowlist of approved hosts.
+#
+# GATEWAY_TEST_ALLOW_REGISTERED_ONLY: When true, only allows testing URLs that match
+# registered gateway base URLs in the database. When false, uses GATEWAY_TEST_ALLOWED_HOSTS
+# patterns. Default: true (most secure).
+# GATEWAY_TEST_ALLOW_REGISTERED_ONLY=true
+#
+# GATEWAY_TEST_ALLOWED_HOSTS: List of allowed host patterns for /admin/gateways/test
+# when GATEWAY_TEST_ALLOW_REGISTERED_ONLY=false. Supports exact hostnames (example.com)
+# and wildcards (*.example.com). Empty list = reject all when registered-only mode is off.
+# Example: GATEWAY_TEST_ALLOWED_HOSTS=["api.example.com","*.partner.com"]
 # GATEWAY_TEST_ALLOWED_HOSTS=[]
+#
+# NOTE: Private IPs (RFC 1918), loopback addresses (127.0.0.0/8), and link-local
+# addresses (169.254.0.0/16) are ALWAYS blocked regardless of allowlist configuration.
 
 ################################################################################
 # UAID Cross-Gateway Routing Security
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-05-10T08:30:33Z",
+  "generated_at": "2026-05-10T08:51:39Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -92,7 +92,7 @@
         "hashed_secret": "08cd923367890009657eab812753379bdb321eeb",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 624,
+        "line_number": 632,
         "type": "Basic Auth Credentials",
         "verified_result": null
       },
@@ -100,7 +100,7 @@
         "hashed_secret": "14f8aa3e560a47851908ab0f04ec856dbc512d93",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 819,
+        "line_number": 827,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -108,7 +108,7 @@
         "hashed_secret": "fa9beb99e4029ad5a6615399e7bbae21356086b3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1092,
+        "line_number": 1100,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -116,7 +116,7 @@
         "hashed_secret": "7b4455a56fbf1d198e45e04c437488514645a82c",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1118,
+        "line_number": 1126,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -124,7 +124,7 @@
         "hashed_secret": "d08f88df745fa7950b104e4a707a31cfce7b5841",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1126,
+        "line_number": 1134,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -132,7 +132,7 @@
         "hashed_secret": "ac371b6dcce28a86c90d12bc57d946a800eebf17",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1205,
+        "line_number": 1213,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -140,7 +140,7 @@
         "hashed_secret": "0b6ec68df700dec4dcd64babd0eda1edccddace1",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1210,
+        "line_number": 1218,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -148,7 +148,7 @@
         "hashed_secret": "4ad6f0082ee224001beb3ca5c3e81c8ceea5ed86",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1215,
+        "line_number": 1223,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -156,7 +156,7 @@
         "hashed_secret": "cb32747fcfb55eaa194c8cd8e4ba7d49ada08a94",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1221,
+        "line_number": 1229,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -164,7 +164,7 @@
         "hashed_secret": "6c178d51b13520496dbc767ed3d9d7aa5803ac72",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1233,
+        "line_number": 1241,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -172,7 +172,7 @@
         "hashed_secret": "ca45060a53fd8a255d1a83ee8d2f025283ccc66e",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1251,
+        "line_number": 1259,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -180,7 +180,7 @@
         "hashed_secret": "910fbf00f58e9bcb095ea26a75cc1d9a3355e671",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1312,
+        "line_number": 1320,
         "type": "Secret Keyword",
         "verified_result": null
       }
@@ -4348,7 +4348,7 @@
         "hashed_secret": "85b60d811d16ff56b3654587d4487f713bfa33b7",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 15340,
+        "line_number": 15365,
         "type": "Secret Keyword",
         "verified_result": null
       }
@@ -4988,7 +4988,7 @@
         "hashed_secret": "ff37a98a9963d347e9749a5c1b3936a4a245a6ff",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 2897,
+        "line_number": 2907,
         "type": "Secret Keyword",
         "verified_result": null
       }
@@ -8664,7 +8664,7 @@
         "hashed_secret": "206c80413b9a96c1312cc346b7d2517b84463edd",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1603,
+        "line_number": 1604,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -8672,7 +8672,7 @@
         "hashed_secret": "5cbd0bf2db07a8f50fa9bbcc5ac720b1911c6380",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1775,
+        "line_number": 1776,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -8680,7 +8680,7 @@
         "hashed_secret": "a10b98d7340036e9c8c301704f623eddd733cc1a",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 2775,
+        "line_number": 2776,
         "type": "Hex High Entropy String",
         "verified_result": null
       },
@@ -8688,7 +8688,7 @@
         "hashed_secret": "3acfb2c2b433c0ea7ff107e33df91b18e52f960f",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 5273,
+        "line_number": 5529,
         "type": "Secret Keyword",
         "verified_result": null
       },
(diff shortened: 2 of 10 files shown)


gandhipratik203 (changes requested):
Thanks for tackling this — the open-proxy risk on `/admin/gateways/test` is a real one and the direction here looks right.

## What I liked
- The two-mode design with `gateway_test_allow_registered_only=True` as the default is the right secure-by-default posture; the `gateway_test_allowed_hosts` escape hatch keeps it flexible.
- Strong defense-in-depth: private/loopback/link-local addresses are blocked unconditionally even when the allowlist permits them, and the admin handler returns a generic "Invalid gateway URL" so internals don't leak to the caller.
- Trailing-dot FQDN normalization (`evil.com.` → `evil.com`) is a nice touch — that's a real bypass class people miss.
- Test coverage is solid: 14 unit tests + 2 e2e tests covering allowlist match/mismatch, wildcards, private/loopback/link-local, OOB collaborator domains, empty allowlist.

## Things I'd like to see addressed before merge

**F1 (high) — the autouse fixture in `tests/unit/mcpgateway/conftest.py` is too broad.**
It monkeypatches `socket.getaddrinfo` and flips `gateway_test_allow_registered_only=False` for every test in `tests/unit/mcpgateway/`, gated only by an in-string nodeid check. A couple of concerns:
- Globally mocking `socket.getaddrinfo` will affect anything else that does DNS during these tests — including the existing `_validate_ssrf` paths in `validators.py`. Tests that should fail-closed on unresolvable hostnames will silently start "resolving" to 8.8.8.8.
- Forcing `gateway_test_allow_registered_only=False` masks the secure default in tests, so future tests exercising the endpoint may pass for the wrong reason.

A safer shape would be to make the fixture opt-in (e.g. a `@pytest.mark.gateway_test_allowlist` marker) and to scope the DNS patch to the validator module (`monkeypatch.setattr("mcpgateway.common.validators.socket.getaddrinfo", ...)`).

**F2 (high) — IPv4-mapped IPv6 will bypass the IP family check on Python 3.11.**
`ipaddress.ip_address("::ffff:127.0.0.1").is_loopback` returns `False` on Python 3.11 (CPython fixed it in 3.12 via bpo-91761). Since the project supports 3.11, a URL like `https://[::ffff:127.0.0.1]/` slips through. Suggest unwrapping mapped addresses before the check:
```python
if ip_addr.version == 6 and ip_addr.ipv4_mapped is not None:
    ip_addr = ip_addr.ipv4_mapped
if ip_addr.is_private or ip_addr.is_loopback or ip_addr.is_link_local:
    raise ...
```

## Things worth folding in while you're here

**F3 (medium) — a few address families aren't covered.**
The validator catches private/loopback/link-local but not:
- `is_unspecified` (0.0.0.0, ::)
- `is_multicast` (224.0.0.0/4, ff00::/8) — handy for SSDP/UPnP probes
- `is_reserved` (240.0.0.0/4 etc.)
- Carrier-grade NAT (100.64.0.0/10) — not covered by `is_private` in stdlib

Cheapest fix is to broaden the predicate and add an explicit 100.64.0.0/10 check.

**F4 (medium) — DNS TOCTOU between validation and connect.**
The validator resolves the hostname, confirms the IPs are public, and returns the original URL. `ResilientHttpClient` then re-resolves DNS at connect time, which means a domain you control can return a public IP at validation and a private one at connect. The PR description calls this "DNS rebinding protection" but it really only handles the static case. True rebinding mitigation needs the validated IP pinned into the connection (custom resolver/transport, or an IP-allowlist check at connect callback). At minimum I'd document the limitation; ideally track as a follow-up.

**F5 (medium) — error messages distinguish "not in allowlist" from "not allowed".**
Today the admin handler wraps both into a generic message, so this is latent. But if `validate_gateway_test_url` ever gets called from a path that surfaces the underlying `ValueError`, the differentiated strings let an attacker enumerate allowlist membership. Safer to emit the same generic string from the validator itself.

## Nits

**F6 (low) — broad `except Exception: continue` in the allowlist-building loop in `admin.py`.**
Hides parse failures on registered gateway URLs — a malformed registration silently disappears from the allowlist with no log line. `(ValueError, AttributeError)` plus a `LOGGER.debug` would help debugging when someone hits "URL not in allowlist" mysteriously.

**F7 (low) — `package-lock.json` change looks unrelated.**
Dropping `"name": "mcp-context-forge"` looks like incidental churn from a local `npm install`. Worth reverting to keep the diff scoped.

**F8 (nit) — imports inside the fixture body.**
`from mcpgateway import config` and `import socket` are inside the fixture function; per AGENTS.md they should be hoisted to the top of `tests/unit/mcpgateway/conftest.py` and grouped per isort sections.

---

To summarize: the design and the secure-by-default story are good, and the tests are thoughtful. Once **F1** (autouse fixture scope) and **F2** (IPv4-mapped-v6 bypass) are addressed, I think we're mostly good to go — F3–F5 are cheap and on the same security surface so worth folding in here, and F6–F8 are pure cleanups. Happy to take another look as soon as you push the updates.

gandhipratik203:
Hey @Nayana-R-Gowda — first off, thanks for taking this one on. The secure-default story you've set up is genuinely good and most of the structure is already where it needs to be, so this isn't a rewrite, just a few targeted tweaks.

The two I'd really like to land before we merge are the autouse fixture in `tests/unit/mcpgateway/conftest.py` and the IPv4-mapped-IPv6 bypass — those are flagged as F1 and F2 in the review above. F1 is mostly about scoping (the DNS mock and the `gateway_test_allow_registered_only=False` flip are reaching tests that shouldn't see them), and F2 is a Python 3.11 quirk where `::ffff:127.0.0.1` sneaks past `is_loopback` — easy to fix once you know it's there, but worth catching since 3.11 is still in our supported range.

While you're in the file, F3–F5 in the review are cheap to fold in and round out the same security surface — extra IP families to block, the DNS TOCTOU caveat, and tightening the validator's error messages so they don't leak allowlist membership. F6–F8 are nits and can come along for the ride.

I'll check back in a couple of days to see how it's going. If anything's unclear or you want to bounce an approach off me — especially on F1, where there's more than one reasonable shape — just reply here or ping me on Slack, happy to talk it through.

Nayana-R-Gowda:
@gandhipratik203 I’ve updated the code changes,please take a look.

F1: Fixed autouse fixture scope in conftest.py                                                                                                   
               
 Problem: The configure_gateway_test_allowlist fixture was using autouse=True, affecting all unit tests globally and potentially breaking tests that should fail-closed on DNS issues.                                
               
 Solution:                                                                                                                      
 - Changed from autouse=True to opt-in fixture
 - Tests that need it must explicitly use @pytest.mark.usefixtures("configure_gateway_test_allowlist")                                                                        
 - Scoped DNS mock to validator module only: monkeypatch.setattr("mcpgateway.common.validators.socket.getaddrinfo", ...)
 - Moved imports to top of file per AGENTS.md guidelines                                                                                               
                                                                                                                           
 F2: Fixed IPv4-mapped IPv6 bypass for Python 3.11                                                                                                  
                                                                                                                           
 Problem: ipaddress.ip_address("::ffff:127.0.0.1").is_loopback returns False on Python 3.11, allowing loopback addresses to bypass validation.                                                    
               
 Solution: Added unwrapping of IPv4-mapped IPv6 addresses before IP checks:                                                                                     
 if ip_addr.version == 6 and ip_addr.ipv4_mapped is not None:
   ip_addr = ip_addr.ipv4_mapped                                                                                                          
 Applied to both direct IP check and DNS resolution check.                                                                                              
                                                                                                                           
 Medium Priority Fixes (F3-F5)                                                                                                            
                                                                                                                           
 F3: Added missing address family checks                                                                                                       
                                                                                                                           
 Problem: Validator only checked is_private, is_loopback, and is_link_local, missing several dangerous address families.                                                               
  
 Solution: Expanded IP blocking to include:                                                                                                     
 - is_unspecified (0.0.0.0, ::)
 - is_multicast (224.0.0.0/4, ff00::/8)                                                                                                       
 - is_reserved (240.0.0.0/4)      
 - Carrier-grade NAT (100.64.0.0/10) - manually checked as stdlib doesn't cover this                                                                                 
                                                                                                                           
 F4: Documented DNS TOCTOU limitation                                                                                                        
                                                                                                                           
 Problem: DNS is resolved at validation time but re-resolved at connection time, allowing DNS rebinding attacks.                                                                   
               
 Solution: Added comprehensive docstring section:                                                                                                  
 **Security Note - DNS TOCTOU Limitation:**
 This validation resolves DNS at validation time to check for private IPs, but the                                                                                  
 HTTP client will re-resolve DNS at connection time. An attacker controlling DNS can                                                                                 
 return a public IP during validation and a private IP during connection (DNS rebinding).                                                                              
 True mitigation requires pinning the validated IP into the connection (custom resolver/                                                                               
 transport, or IP allowlist check at connect callback). This is tracked as a known                                                                                  
 limitation for future improvement.                                                                                                         
                                                                                                                           
 F5: Unified error messages to prevent allowlist enumeration                                                                                             
                                                                                                                           
 Problem: Error messages distinguished "not in the approved allowlist" from "is not allowed", potentially leaking allowlist membership.                                                       
                                                                                                                           
 Solution: Changed all error messages to generic "Gateway URL is not allowed"                                                                                    
 - Updated validator method 
 - Updated docstring examples                                                                                                            
 - Updated all 5 test assertions to match new message
                                                                                                                           
 Low Priority Fixes (F6-F8)                                                                                                             
  
 F6: Added logging for malformed gateway URLs                                                                                                    
               
 Problem: Broad except Exception: continue in allowlist building hid parse failures.                                                                                 
               
 Solution: Changed to catch specific exceptions with debug logging:                                                                                         
 except (ValueError, AttributeError) as e:
   LOGGER.debug("Failed to parse registered gateway URL '%s': %s", url, e)                                                                                     
   continue                                                                                                                    
                                                                                                                           
 F7: Checked package-lock.json                                                                                                            
                                                                                                                           
 Problem: Reviewer mentioned unrelated changes dropping "name": "mcp-context-forge".                                                                                 
                                                                                                                           
 Solution: Verified the file is correct and unchanged in current branch - no action needed.                                                                             
               
 F8: Fixed import placement in conftest.py                                                                                                      
               
 Problem: Imports were inside the fixture function body instead of at module top.                                                                                  
  
 Solution: Moved import socket and from mcpgateway import config to top of file per AGENTS.md guidelines.

gandhipratik203 (dismissed):
Hey @Nayana-R-Gowda — thanks for the quick turnaround on the last round, this is in much better shape now. I went through the diff again and every one of the items from the previous review has been addressed:

- **F1 (autouse fixture)** — the fixture is opt-in now and tests pull it in via `@pytest.mark.usefixtures("configure_gateway_test_allowlist")`. DNS mock is scoped to the validator module. Exactly what I was hoping for.
- **F2 (IPv4-mapped IPv6)** — the unwrap is in both the literal-IP and resolved-IP paths. `::ffff:127.0.0.1` will no longer slip past on 3.11.
- **F3 (extra address families)** — `is_unspecified`, `is_multicast`, `is_reserved`, and the explicit `100.64.0.0/10` CGN check are all in. Nice and complete.
- **F4 (DNS TOCTOU)** — the docstring now has a clear note flagging this as a known limitation. That was the minimum I was asking for; tracking the real fix as a follow-up is fine.
- **F5 (differentiated error messages)** — every rejection path returns the same generic "is not allowed" string now. No more allowlist enumeration via error text.
- **F6, F7, F8** — all addressed (specific exception types + debug log, package-lock reverted, imports hoisted).

A few small things I spotted while reading the new code, all non-blocking:

1. The validator's docstring has four examples; the trailing-dot one (`evil.com.`) will do a real DNS lookup when doctests run, since only the first example has `# doctest: +SKIP`. Worth adding `+SKIP` to that one too so doctests don't depend on network. The two literal-IP examples short-circuit before DNS so they're fine.
2. `test_admin_test_gateway_registered_mode_db_query_exception` has a docstring that says "should fall back to configured hosts" but the actual behavior (and the assertion) is fail-closed — the configured fallback isn't used in registered-only mode when the DB query throws. Behavior is correct (fail-closed is the safer default), the docstring just disagrees with itself.
3. The new tests in `test_admin.py` each `monkeypatch` the same three settings + `getaddrinfo` inline. Now that the `configure_gateway_test_allowlist` fixture exists, they could share it — purely a tidiness thing.
4. There's an unrelated import path change in `tests/integration/test_rate_limiter_redis_url_from_yaml.py` (`cpex.framework` → `cpex.framework.loader.config`). Looks incidental from a rebase — is that intended for this PR or did it sneak in?

Overall this is in great shape. Approving — feel free to fold those four cleanups into a follow-up commit or just let them ride; nothing here is worth holding the merge for.

Nayana-R-Gowda:
Hi @gandhipratik203 I’ve updated the code changes,please take a look.

• Fixed doctest network dependency
File:mcpgateway/common/validators.py
Added # doctest: +SKIP to the trailing-dot bypass example (evil.com.) in the validate_gateway_test_url docstring.
This prevents real DNS lookups during doctest execution and keeps doctests network-independent.
• Corrected test documentation
File:tests/unit/mcpgateway/test_admin.py
• Updated the inline comment to accurately reflect the behavior being tested.
• New comment:
"DB query fails and validation fails closed (no fallback to configured hosts in registered-only mode)"
• E2E fixture refactoring
Not implemented intentionally.
Since this was identified as a non-blocking cleanup/tidiness improvement, I kept the explicit inline setup for readability in the E2E tests.
Can revisit in a future cleanup PR if needed.
• Import path verification
File:tests/integration/test_rate_limiter_redis_url_from_yaml.py
Verified that cpex.framework.loader.config is the intended import.
This is part of the CPEX migration introduced in commit 86ca5c248, so no change was required.

gandhipratik203:
Hey @Nayana-R-Gowda — thanks for getting the doctest skip and the docstring tweak in, those both read nicely now.

Quick one before this lands: I noticed the import path change in `tests/integration/test_rate_limiter_redis_url_from_yaml.py` (`cpex.framework` → `cpex.framework.loader.config`) is still hanging around. Did you mean to include that in this PR, or is it just along for the ride from a rebase? If the latter, probably worth dropping it so the diff stays focused on the security fix — but if there's a reason for it, no worries, leave it in.

PR #4496, opened 2026-04-28
fix(health): unauthenticated liveness probe for authorization_code gateways + POST ping for StreamableHTTP

### LINKED ISSUE #4495: [BUG]: Health check marks authorization_code OAuth gateways unreachable when PLATFORM_ADMIN_EMAIL has no stored token
## Description

The periodic gateway health check (`_check_single_gateway_health`) marks all `authorization_code` OAuth gateways as unreachable after every pod restart.

## Root Cause

The health check runs under `PLATFORM_ADMIN_EMAIL` (a system account). For `authorization_code` gateways, the code looks up stored OAuth tokens for this email. Since the system account never completes an interactive OAuth dance, it has **zero stored tokens**. The code then calls `_handle_gateway_failure()` which, after `GW_FAILURE_THRESHOLD` consecutive failures, sets `reachable=false` in the database.

This affects **every** `authorization_code` gateway — they all go offline within minutes of a pod restart.

A secondary issue: the StreamableHTTP health check uses the full MCP SDK `ClientSession.initialize()` which opens a GET SSE stream after the initialize handshake. Servers that don't support server-initiated messages (per the MCP spec, GET is optional) return 405, causing a false health failure.

## Current Behavior

1. Pod starts, health check scheduler runs
2. For each `authorization_code` gateway: no token found for system email → `_handle_gateway_failure()`
3. After threshold: `reachable=false` in DB
4. All `authorization_code` gateways appear offline in the Admin UI
5. All tools from those gateways disappear from virtual servers

## Expected Behavior

The health check should distinguish between "server is down" and "I don't have credentials for this server":
- **No token available**: proceed without auth — a 401/403 response proves the server is alive
- **Timeout/DNS/connection error**: real outage → mark unreachable
- StreamableHTTP: use a lightweight POST `initialize` instead of the full SDK client

## Steps to Reproduce

1. Configure multiple gateways with `grant_type: authorization_code` (e.g., Microsoft 365, GitHub, GitLab)
2. Set `PLATFORM_ADMIN_EMAIL` to a system account (e.g., `contextforgeadmin@example.com`)
3. Do NOT complete the OAuth authorization flow for that system account on any gateway
4. Start the gateway (or restart a pod)
5. Wait for 2-3 health check cycles (~60s each)
6. Query the database: `SELECT name, reachable FROM gateways WHERE oauth_config::text LIKE '%authorization_code%';`
7. All authorization_code gateways show `reachable=false`

## Environment

- ContextForge version: latest main
- Database: PostgreSQL
- 8 gateways affected (Freshservice, GitHub, GitLab, M365, Kubernetes x2, AI Context Hub x2)
- All configured with `grant_type: authorization_code`

## Related

- #4360 (different scenario — tool invocation with sessions, closed as working-as-expected)
- Branch `issue_4360_GW_unreable` by @rakdutta addresses HTTP 4xx at the response level but does not fix the early return when no token exists for the system account

ecthelion77:
Fixes #4495

## Summary

The gateway health check marks all `authorization_code` OAuth gateways as unreachable after pod restarts because the system account (`PLATFORM_ADMIN_EMAIL`) has no stored OAuth tokens. This PR introduces an **unauthenticated liveness probe** strategy: when no token is available, the health check proceeds without auth — HTTP 401/403 proves the server is alive, while timeouts/DNS/connection errors indicate a real outage.

Additionally, replaces the full MCP SDK `ClientSession.initialize()` for StreamableHTTP health checks with a lightweight JSON-RPC POST `initialize` request. The SDK opens a GET SSE stream after initialize which returns 405 on servers that don't support server-initiated messages (the MCP spec says GET is optional).

## Changes

**`mcpgateway/services/gateway_service.py`** — `_check_single_gateway_health()`:

### 1. Unauthenticated liveness probe (`unauthenticated_probe` flag)

For `authorization_code` gateways, the health check now:
- Tries to use a stored token if available (existing behavior)
- If no token exists (typical for system accounts): **proceeds without auth** instead of calling `_handle_gateway_failure()`
- Sets `unauthenticated_probe = True` so that HTTP 401/403 responses are treated as "server alive" for both SSE and StreamableHTTP transports
- Timeouts, DNS failures, and connection errors still trigger `_handle_gateway_failure` as before

Previously, the code did an early return to `_handle_gateway_failure()` when `user_email` was missing or when no stored token was found, making it impossible for authorization_code gateways to pass health checks under a system account.

### 2. Lightweight StreamableHTTP health check

Replaced the full SDK client session:
```python
# Before (causes 405 on servers without server-initiated messages):
async with streamablehttp_client(url=...) as (read, write, _):
    async with ClientSession(read, write) as session:
        response = await session.initialize()

# After (lightweight POST only):
response = await client.post(url, json=init_payload, headers=..., timeout=...)
```

The MCP SDK `ClientSession.initialize()` opens a GET SSE stream after the initialize handshake. Servers that don't support server-initiated messages (M365 MCP, Kubernetes MCP, GitHub MCP) return 405 on GET, causing a false health failure. A successful POST `initialize` is sufficient proof of health per the MCP spec.

### 3. Observability

Added `health.probe_type: unauthenticated_liveness` span attribute when probing without auth, for tracing visibility.

## Testing

Validated in production with 13 gateways (8 `authorization_code` + 5 non-OAuth). After deployment:
- All 13 gateways report `reachable=true`
- Previously, all 8 authorization_code gateways were stuck at `reachable=false` after every pod restart
- StreamableHTTP gateways (M365, Kubernetes, GitHub) no longer fail with 405

Signed-off-by: Olivier Gintrand <olivier.gintrand@forterro.com>

### DIFF
--- a/mcpgateway/services/gateway_service.py
+++ b/mcpgateway/services/gateway_service.py
@@ -4298,6 +4298,12 @@ class GatewayService(BaseService):  # pylint: disable=too-many-instance-attribut
             # Use isolated client for gateway health checks (each gateway may have custom CA cert)
             # Use admin timeout for health checks (fail fast, don't wait 120s for slow upstreams)
             # Pass ssl_context if present, otherwise let get_isolated_http_client use skip_ssl_verify setting
+            # Track whether this is an unauthenticated liveness probe
+            # (authorization_code gateway without system token).  When True,
+            # HTTP 401/403 responses are treated as "server alive" rather
+            # than health-check failures.
+            unauthenticated_probe = False
+
             async with get_isolated_http_client(timeout=settings.httpx_admin_read_timeout, verify=ssl_context) as client:
                 logger.debug("Checking health of gateway: %s (%s)", gateway_name, gateway_url_sanitized)
                 try:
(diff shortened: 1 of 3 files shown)


jonpspri:
Closing in favour of already-merged work.

The two root causes from #4154 are now addressed separately:

- **`authorization_code` unauthenticated liveness probe** — fully covered by PR #5244 (merged 2026-07-22), which decouples health checks from token ownership and treats 401/403 as "gateway reachable". `Closes #4154` has been added to that PR's description.
- **StreamableHTTP full-SDK `initialize()` → GET 405** — resolved at the protocol level on the `mcp-2-0-0-prerelease` branch: MCP SDK v2.0.0 / spec 2026-07-28 drops the `initialize` handshake entirely for StreamableHTTP (each POST is self-describing), so the GET SSE stream is never opened during a health probe.

Nothing in this PR needs to land on `main` independently.

PR #4505, opened 2026-04-28
fix: rate-limit _cleanup_table DELETEs with configurable inter-batch sleep

### LINKED ISSUE #4504: [PERFORMANCE]: Rate-limit _cleanup_table DELETEs — add sleep intervals between batches
## Summary

`_cleanup_table` and `delete_metrics_in_batches` in `metrics_cleanup_service.py` run batch DELETEs in a tight loop with no pause between iterations. Under load, a large cleanup run fires hundreds of back-to-back DELETEs that compete directly with request-serving queries, contributing to PgBouncer saturation (~12% of profiled DB capacity from this one background task — 88/740 py-spy samples).

## Proposed Fix

Add a configurable inter-batch sleep (`METRICS_CLEANUP_BATCH_SLEEP_MS`, default 50 ms) that fires only after a **full** batch, so a partial (final) batch never adds unnecessary delay.

### Files changed

| File | Change |
|---|---|
| `mcpgateway/config.py` | Add `metrics_cleanup_batch_sleep_ms` field |
| `mcpgateway/services/metrics_cleanup_service.py` | Read new setting; sleep between full batches in `_cleanup_table` and `delete_metrics_in_batches`; surface in `get_stats()` |
| `.env.example` | Add commented entry near `METRICS_CLEANUP_BATCH_SIZE` |
| `tests/unit/mcpgateway/services/test_metrics_cleanup_service.py` | Tests verifying sleep is called between full batches but not after the last partial batch |

### Operator tuning guide

| Scenario | Recommended `METRICS_CLEANUP_BATCH_SLEEP_MS` |
|---|---|
| Low-traffic / dev | `0` (no sleep) |
| Default production | `50` |
| High load / constrained DB | `100–200` |
| Emergency throttle | `500` |

## Parent issue

Closes #2692

msureshkumar88:
## Summary

- Adds `METRICS_CLEANUP_BATCH_SLEEP_MS` config field (default `50 ms`, range `0–5000 ms`)
- `_cleanup_table` and `delete_metrics_in_batches` now sleep between **full** batches only — a partial (final) batch never adds unnecessary delay
- New setting surfaced in `get_stats()` output for observability
- `.env.example` updated with commented entry near `METRICS_CLEANUP_BATCH_SIZE`

## Motivation

Back-to-back batch DELETEs with no pause competed directly with request-serving queries, contributing ~12% of profiled DB capacity (88/740 py-spy samples) to PgBouncer saturation. A 50 ms sleep between full batches gives the DB breathing room while adding negligible total cleanup time.

## Test plan

- [ ] `test_cleanup_table_sleeps_between_full_batches` — sleep called once when first batch is full
- [ ] `test_cleanup_table_no_sleep_after_partial_batch` — sleep never called when batch < batch_size
- [ ] `test_get_stats_includes_batch_sleep_ms` — new key present in `get_stats()` output
- [ ] `METRICS_CLEANUP_BATCH_SLEEP_MS=0` disables sleep entirely (backward-compatible)

Closes #4504
Relates to #2692

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -2055,6 +2055,10 @@ SKIP_MIGRATION=false
 # Larger batches are faster but may cause longer table locks
 # METRICS_CLEANUP_BATCH_SIZE=10000
 
+# Milliseconds to sleep between batch DELETEs (default: 50, range: 0-5000)
+# Increase to reduce DB pressure on high-traffic deployments (0 = no sleep)
+# METRICS_CLEANUP_BATCH_SLEEP_MS=50
+
 # Metrics Rollup Configuration
 # =============================================================================
 # Aggregates raw metrics into hourly summaries for efficient historical queries
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-05-22T17:48:13Z",
+  "generated_at": "2026-05-22T21:53:08Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -4284,52 +4284,6 @@
         "is_verified": false,
         "line_number": 752,
         "type": "Hex High Entropy String",
-        "verified_result": null
-      }
-    ],
-    "mcpgateway/config.py": [
-      {
-        "hashed_secret": "9d4e1e23bd5b727046a9e3b4b7db57bd8d6ee684",
-        "is_secret": false,
-        "is_verified": false,
-        "line_number": 249,
-        "type": "Basic Auth Credentials",
-        "verified_result": null
-      },
-      {
-        "hashed_secret": "ff37a98a9963d347e9749a5c1b3936a4a245a6ff",
-        "is_secret": false,
-        "is_verified": false,
-        "line_number": 2948,
-        "type": "Secret Keyword",
-        "verified_result": null
-      }
-    ],
-    "mcpgateway/db.py": [
-      {
-        "hashed_secret": "9d4e1e23bd5b727046a9e3b4b7db57bd8d6ee684",
-        "is_secret": false,
-        "is_verified": false,
-        "line_number": 80,
-        "type": "Basic Auth Credentials",
-        "verified_result": null
-      }
-    ],
-    "mcpgateway/llm_provider_configs.py": [
-      {
-        "hashed_secret": "25910f981e85ca04baf359199dd0bd4a3ae738b6",
-        "is_secret": false,
-        "is_verified": false,
-        "line_number": 308,
-        "type": "AWS Access Key",
-        "verified_result": null
-      },
-      {
-        "hashed_secret": "d70eab08607a4d05faa2d0d6647206599e9abc65",
-        "is_secret": false,
-        "is_verified": false,
-        "line_number": 316,
-        "type": "Base64 High Entropy String",
         "verified_result": null
       }
     ],
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -4287,6 +4287,52 @@
         "verified_result": null
       }
     ],
+    "mcpgateway/config.py": [
+      {
+        "hashed_secret": "9d4e1e23bd5b727046a9e3b4b7db57bd8d6ee684",
+        "is_secret": false,
+        "is_verified": false,
+        "line_number": 249,
+        "type": "Basic Auth Credentials",
+        "verified_result": null
+      },
+      {
+        "hashed_secret": "ff37a98a9963d347e9749a5c1b3936a4a245a6ff",
+        "is_secret": false,
+        "is_verified": false,
+        "line_number": 2948,
+        "type": "Secret Keyword",
+        "verified_result": null
+      }
+    ],
+    "mcpgateway/db.py": [
+      {
+        "hashed_secret": "9d4e1e23bd5b727046a9e3b4b7db57bd8d6ee684",
+        "is_secret": false,
+        "is_verified": false,
+        "line_number": 80,
+        "type": "Basic Auth Credentials",
+        "verified_result": null
+      }
+    ],
+    "mcpgateway/llm_provider_configs.py": [
+      {
+        "hashed_secret": "25910f981e85ca04baf359199dd0bd4a3ae738b6",
+        "is_secret": false,
+        "is_verified": false,
+        "line_number": 308,
+        "type": "AWS Access Key",
+        "verified_result": null
+      },
+      {
+        "hashed_secret": "d70eab08607a4d05faa2d0d6647206599e9abc65",
+        "is_secret": false,
+        "is_verified": false,
+        "line_number": 316,
+        "type": "Base64 High Entropy String",
+        "verified_result": null
+      }
+    ],
     "mcpgateway/middleware/request_logging_middleware.py": [
       {
         "hashed_secret": "f2b14f68eb995facb3a1c35287b778d5bd785511",
--- a/mcpgateway/config.py
+++ b/mcpgateway/config.py
@@ -1967,7 +1967,9 @@ class Settings(BaseSettings):
     metrics_retention_days: int = Field(default=7, ge=1, le=365, description="Days to retain raw metrics before cleanup (fallback when rollup disabled)")
     metrics_cleanup_interval_hours: int = Field(default=1, ge=1, le=168, description="Hours between automatic cleanup runs")
     metrics_cleanup_batch_size: int = Field(default=10000, ge=100, le=100000, description="Batch size for metrics deletion (prevents long locks)")
-    metrics_cleanup_batch_sleep_ms: int = Field(default=50, ge=0, le=5000, description="Milliseconds to sleep between batch DELETEs in cleanup (0 = no sleep). Use to reduce DB pressure during background cleanup.")
+    metrics_cleanup_batch_sleep_ms: int = Field(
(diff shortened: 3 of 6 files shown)


gandhipratik203:
## What's good
- Concrete motivation backed by py-spy data (88/740 samples, ~12% of profiled DB capacity) — not a speculative knob.
- `_cleanup_table` runs via `asyncio.to_thread(...)` (lines 292/311/441), so `time.sleep()` blocks the worker thread, not the event loop. Correct design for that path.
- Sleep is placed *after* `db.commit()` in `_cleanup_table`, so it doesn't hold transaction/row locks — exactly the right ordering.
- Partial-batch exit is preserved: the `< batch_size` break happens before the sleep, so the final batch never sleeps.
- Default of 50ms is sensible, range guards are explicit (0–5000 ms), and `0` disables cleanly.
- `get_stats()` surfaces the new value — nice for observability.
- 50ms × ~N batches adds negligible total cleanup latency for the background path, which is what the PR optimizes for.

## Findings

**F1 (medium) — `delete_metrics_in_batches` sleep blocks the event loop in the user-facing delete path.**
The same setting also rate-limits `delete_metrics_in_batches`, which is called from `async def delete_agent` (`a2a_service.py:1585`), and similarly from `async def delete_server` / `async def delete_resource` — directly inline in the async function, not via `asyncio.to_thread`. So every 50ms `time.sleep()` here blocks the event loop, and a delete of an entity with 100k historical metrics means 10× 50ms = 500ms of event-loop pause. The PR's motivation (PgBouncer saturation in background cleanup) doesn't really apply to user-initiated deletes — it's a different cost/benefit. Two reasonable options:
- Decouple the two paths (e.g. separate setting for inline deletes, default 0).
- Skip the sleep entirely in `delete_metrics_in_batches` and keep this PR scoped to the background cleanup path the profiling data describes.

**F2 (medium) — In `delete_metrics_in_batches`, the sleep happens with the transaction still open.**
The function's docstring says "within the current transaction" and there's no `commit()` between batches. So unlike `_cleanup_table`, the inter-batch `time.sleep()` is held *while* row locks are still owned by this transaction. That works against the goal of relieving DB pressure — concurrent writers on the same rows wait longer, and the connection sits idle-but-locked through the sleep. If F1 is resolved by removing the sleep here, this goes away with it.

**F3 (low) — Test coverage gaps.**
- No test exercises `delete_metrics_in_batches` sleep behavior — only `_cleanup_table` is covered, even though the same change is in two places.
- The PR test plan lists "`METRICS_CLEANUP_BATCH_SLEEP_MS=0` disables sleep entirely (backward-compatible)" but there's no explicit test for the 0-disable case. One small `service.batch_sleep_ms = 0` variant of the existing test would lock it down.

**F4 (nit) — Mock setup duplication in `test_cleanup_table_sleeps_between_full_batches`.**
```python
mock_db.execute.return_value = mock_execute
mock_db.execute.return_value.fetchall.side_effect = [full_batch, empty_batch]
```
The second line re-assigns `fetchall.side_effect` on the same object that was just set via `mock_execute.fetchall.side_effect = ...`. Works because `MagicMock` is permissive but reads as confusing — drop one of them.

**F5 (nit) — `.secrets.baseline` change is incidental** (timestamp + line-number bumps from a regenerate). Harmless, but not part of this PR's intent — could be reverted to keep the diff scoped.

## Verdict
Good targeted fix for the background cleanup case; the design choices in `_cleanup_table` are right. I'd block on F1+F2 because applying the same throttle inside `delete_metrics_in_batches` extends event-loop blocking and lock-hold time on the user-initiated delete path — opposite of the PR's intent. F3 closes a small coverage gap; F4/F5 are pure nits.

msureshkumar88:
Thanks for the thorough review — F1 and F2 are both valid. Here's what was done:

**F1 + F2 fixed** — removed inter-batch sleep from `delete_metrics_in_batches`

The root issue: `delete_metrics_in_batches` is called inline (no `asyncio.to_thread`) from five async service methods (`delete_agent`, `delete_server`, `delete_resource`, `delete_prompt`, `delete_tool`). `time.sleep()` there blocked the event loop on every user-initiated entity delete with metrics.

F2 compounds this: unlike `_cleanup_table`, `delete_metrics_in_batches` has no `db.commit()` between batches, so the sleep was holding row locks — the opposite of relieving DB pressure.

**Fix:** stripped `batch_sleep_ms` read and `time.sleep()` from `delete_metrics_in_batches` entirely. The background cleanup path (`_cleanup_table`) keeps its sleep — it runs via `asyncio.to_thread`, commits after each batch, and is where the profiling data actually showed the pressure.

```diff
 def delete_metrics_in_batches(...) -> int:
     effective_batch_size = batch_size or getattr(settings, "metrics_cleanup_batch_size", 10000)
-    batch_sleep_ms = getattr(settings, "metrics_cleanup_batch_sleep_ms", 50)
     total_deleted = 0

     while True:
         ...
         if batch_deleted <= 0 or batch_deleted < effective_batch_size:
             break

-        if batch_sleep_ms > 0:
-            time.sleep(batch_sleep_ms / 1000.0)
-
     return total_deleted
```

All 28 existing tests still pass. Ready for re-review.

gandhipratik203:
Thanks again for the data-driven motivation here, the py-spy numbers make this easy to reason about.

One thought reading through it: this change touches the Postgres path under realistic load — that's the whole point — but the current tests verify it via `mock.patch("time.sleep")`, which only really proves the call site fires. It doesn't actually exercise that the cleanup completes correctly against a real DB session, or that the inter-batch sleep has the intended effect on contention with concurrent queries (which is what the PgBouncer-saturation motivation is really about).

I think this PR would benefit from an integration- or e2e-style test that runs against a real Postgres and validates the change is actually taking effect — even something simple like seeding rows, running cleanup end-to-end, and asserting on counts and timing. `tests/integration/test_concurrency_row_locking.py` is a good template for the Postgres-gated style this repo already uses (skips when `DATABASE_URL`/`DB` isn't Postgres). Happy to leave the shape up to you.

msureshkumar88:
Good call. Added `tests/integration/test_metrics_cleanup_pg.py` — 4 Postgres-gated tests following the `is_postgresql()` skip pattern from `test_concurrency_row_locking.py`:

- `test_cleanup_table_deletes_old_rows_keeps_recent` — seeds 5 old + 3 recent rows, asserts 5 deleted, 3 survive, `duration_seconds > 0`
- `test_cleanup_table_multi_batch_deletes_all_old` — `batch_size=2` with 6 old rows, verifies all 6 gone across 3 batches
- `test_cleanup_table_empty_table_noop` — no rows → `deleted_count == 0`, no error
- `test_cleanup_table_nothing_to_delete_when_all_recent` — all rows within window → none deleted

Tests skip automatically in the default hermetic SQLite run. Enable with `MCPGATEWAY_TEST_ALLOW_EXTERNAL_DB=1 DATABASE_URL=postgresql://...`.

One scope note: the PgBouncer contention reduction isn't assertable deterministically in CI — that's a load-test observation, not a unit observable. These tests cover correctness (right rows deleted, right rows survive, batching works end-to-end on real Postgres).

gandhipratik203 (dismissed):
Thanks @msureshkumar88 — this lands really well. The new `tests/integration/test_metrics_cleanup_pg.py` is exactly the shape I had in mind: Postgres-gated with the same skip pattern as `test_concurrency_row_locking.py`, and the four scenarios (single batch, multi-batch with `batch_size=2`, empty table, all-recent) cover the behavior we actually care about against a real DB instead of mocks.

Also a really nice bonus on the inline-delete path — pulling the `time.sleep` out of `delete_metrics_in_batches` cleanly resolves both concerns from the previous comment (event-loop blocking on user-facing async deletes + the open-transaction lock-hold issue), and the commit message lays out the reasoning clearly. Background cleanup keeps the throttle where it actually helps.

One thing before merge: pre-commit hooks are failing in CI (looks like the detect-secrets hook flagged a baseline modification). Could you take a look and push the fix?

gandhipratik203 (dismissed):
Re-approving after the rebase — verified the 5 commits today (`b46ae981` through `42931a76`) are a clean resequencing of what was previously approved on 2026-05-08, with no scope creep. The F1+F2 fix (removing the sleep from `delete_metrics_in_batches`) and the Postgres-gated integration tests from the earlier review thread are both intact.

Four minor nits, none blocking — feel free to address in this PR or land as-is:

1. **PR description is stale** — the line "`_cleanup_table` and `delete_metrics_in_batches` now sleep between **full** batches only" is no longer accurate after the F1+F2 fix removed the sleep from `delete_metrics_in_batches`. Suggest reword to make clear that the inline async path deliberately does not sleep:

   > `_cleanup_table` sleeps between full batches only — a partial (final) batch never adds delay. `delete_metrics_in_batches` (called inline from async paths) deliberately does not sleep to avoid event-loop blocking.

2. **Test-plan checkboxes** in the PR description are unticked (`- [ ]` × 4) even though all four tests exist in the diff. Easy click-fix.

3. **Redundant `side_effect` assignment** in `test_cleanup_table_sleeps_between_full_batches`:
   ```python
   mock_execute.fetchall.side_effect = [full_batch, empty_batch]
   ...
   mock_db.execute.return_value = mock_execute
   mock_db.execute.return_value.fetchall.side_effect = [full_batch, empty_batch]  # same object, same value
   ```
   The last line resets `side_effect` on the same MagicMock to the same values — harmless but dead.

4. **Integration test fixture relies on FK cascade** — `pg_session.delete(pg_session.get(Tool, tool_id))` assumes `ON DELETE CASCADE` on `tool_metrics.tool_id`. Test isolation is preserved by UUID scoping, but Postgres CI may accumulate orphan rows across runs. Belt-and-suspenders: explicitly delete `ToolMetric` rows for the tool_id before the parent.

LGTM otherwise. Nice work keeping the change tight through the rebase.

brian-hussey (approved):
Approved based on @gandhipratik203 review.
Now rebasing and merging.

PR #4550, opened 2026-04-30
perf(cache): store full team objects in AuthCache to eliminate secondary DB query

### LINKED ISSUE #3005: [FEATURE][PERFORMANCE]: Extend get_user_teams cache to store full team objects instead of just IDs
### 🐞 Bug Summary
The current `get_user_teams` cache only stores team IDs. Because of this, every request needing full team objects performs additional DB lookups, increasing overall request latency and DB load under high concurrency. The cache should be capable of storing complete team objects so downstream consumers do not require extra lookups per request.

---

### 🧩 Affected Component
- [x] `mcpgateway` - API
- [ ] Other (explain below)

---

### 🔁 Steps to Reproduce
1. Authenticate with a user account with team memberships
2. Observe multiple DB round-trips per request for downstream code needing team info
3. Inspect `get_user_teams` cache contents (shows only team IDs, not objects)

---

### 🤔 Expected Behavior
The `get_user_teams` cache should store the full set of team objects, reducing repetitive DB calls and improving performance under load.

---

### 📓 Logs / Error Output
N/A — performance and repeated DB query issue; no specific error logs

---

### 🧠 Environment Info
| Key | Value |
|-----|-------|
| Version or commit | main@current |
| Platform / OS | Ubuntu 24.04 |
| Container | none|

---

See profiling and recommendations in parent issue #2692.

msureshkumar88:
## What and Why

The existing `get_user_teams()` cache stored only team IDs. Any downstream code needing full team objects (name, slug, visibility, membership state, etc.) had to fire a second DB query even on a cache hit — increasing latency and DB load under concurrency. This is the fix described in issue #3005.

### `mcpgateway/cache/auth_cache.py`
- Adds a dedicated `team_objs` L1/L2 cache bucket alongside the existing `teams_list` bucket.
- New public methods: `team_to_dict()`, `get_user_team_objects()`, `set_user_team_objects()`.
- `invalidate_user_teams()` and `invalidate_all()` now also evict `_team_objects_cache`.
- `stats()` reports `team_objects_cache_size`.

### `mcpgateway/services/team_management_service.py`
- `get_user_teams()` now checks the full-object cache first; on a hit it reconstructs `EmailTeam` objects in-process via `_team_from_dict()` with zero DB queries.
- On a cache miss the DB result is serialised and stored via `set_user_team_objects()` for subsequent requests.

### Tests (`tests/unit/mcpgateway/cache/test_auth_cache_l1_l2.py`)
- `TestUserTeamObjectsL1L2` class: L1 round-trip, empty-list caching, L2 Redis hit with write-through, invalidation, `invalidate_all`, `stats`, disabled-cache guard.
- Branch-coverage additions: Redis miss (`data=None`), Redis `get()` exception, disabled early-return in `set_user_team_objects`, Redis `setex()` happy path, Redis `setex()` exception.
- Overall coverage for the new methods: 99 %.

## Behaviour Before / After

| Scenario | Before | After |
|---|---|---|
| Cache hit | 1 cache lookup + 1 DB query | 1 cache lookup, 0 DB queries |
| Cache miss | 1 DB query, IDs stored | 1 DB query, full objects stored |
| Redis hit | IDs fetched + 1 extra DB query | Full objects returned directly |

## Reproduction Steps (issue #3005)

1. Start the gateway with a user that belongs to at least one team:
   ```bash
   cp .env.example .env && make install-dev
   make dev
   ```
2. Generate a token and call an endpoint that resolves team membership (e.g. `GET /servers` with a team-scoped token):
   ```bash
   export TOKEN=$(python -m mcpgateway.utils.create_jwt_token \
     --username user@example.com --exp 60 --secret "$JWT_SECRET_KEY")
   curl -s -H "Authorization: Bearer $TOKEN" http://localhost:8000/servers | jq .
   ```
3. Enable SQLAlchemy query logging (`LOG_LEVEL=DEBUG` in `.env`) and repeat the request.
4. **Before this fix:** observe two SELECT statements per request — one for team IDs, one to hydrate each team object.
5. **After this fix:** only one SELECT on the first request; subsequent requests show no SELECT at all (served from cache).

### 1. Unit tests
```bash
.venv/bin/python -m pytest tests/unit/mcpgateway/cache/test_auth_cache_l1_l2.py \
  tests/unit/mcpgateway/services/test_team_management_service.py \
  tests/unit/mcpgateway/services/test_team_management_service_coverage.py -v
```
All tests should pass with no failures or errors.

### 2. Coverage check
```bash
.venv/bin/python -m pytest tests/unit/mcpgateway/cache/test_auth_cache_l1_l2.py \
  --cov=mcpgateway.cache.auth_cache --cov-report=term-missing
```
Expected: ≥ 99 % coverage for `auth_cache.py`; lines 859-861, 879, 884, 886, 888-891 reported as covered.

### 3. Cache hit — no secondary DB query
```bash
make dev  # start server
# First request: populates cache (one DB query expected)
curl -s -H "Authorization: Bearer $TOKEN" http://localhost:8000/servers
# Second request: served from cache (zero DB queries)
curl -s -H "Authorization: Bearer $TOKEN" http://localhost:8000/servers
```
With `LOG_LEVEL=DEBUG`, the second request should show no `SELECT` for the `teams` table.

### 4. Cache invalidation
Add or remove a user from a team via `POST /teams/{id}/members` or `DELETE /teams/{id}/members/{email}` and confirm that the next `GET /servers` call re-queries the DB (one SELECT), then subsequent calls are cached again.

### 5. Redis path (optional)
Set `REDIS_URL=redis://localhost:6379/0` in `.env`, start Redis, and repeat steps 3-4. Confirm that `team_objs:*` keys appear in Redis (`redis-cli keys 'team_objs:*'`) and are evicted on invalidation.

Closes #3005


ja8zyjits:
May be create a follow up issues for this
# Concerns

- Two cache representations now exist: old team ID cache and new full-object cache. Drift risk if any caller still reads ID cache while writes move to object cache.
- Cached objects are reconstructed transient `EmailTeam` instances, not session-bound ORM rows. Any downstream code expecting attached ORM behaviour, lazy relationships, or identity-map semantics may break.
- Serialisation caches only selected scalar fields. If consumers rely on omitted fields now or later, cache-hit objects may differ from DB-fetched objects.
- `team_to_dict()` directly accesses fields like `slug`, `created_by`, `visibility`, `is_active`. Safe if always populated; brittle if nullable or schema changes.
- Datetime round-trip uses ISO serialise and `datetime.fromisoformat()`. Usually fine, but timezone and naive-versus-aware consistency should be watched.
- Invalidation clears object cache in code shown here, but worth confirming all membership mutation paths reliably call `invalidate_user_teams()`.
- Diff and unit tests support behaviour change, but do not by themselves prove measurable end-to-end concurrency or perf improvement.
- Old ID-cache methods remain in `AuthCache`. Maintenance burden unless other callers still need them.

ja8zyjits (approved):
LGTM

PR #4553, opened 2026-05-01
Fix MCP path rewrite order for streamable POST auth

### LINKED ISSUE #4275: [BUG]: MCPPathRewriteMiddleware breaks streaming POST when rewriting /mcp to /mcp/
## Description

The `MCPPathRewriteMiddleware` in `mcpgateway/main.py` rewrites `/servers/<id>/mcp` to `/mcp/` by modifying the ASGI `scope["path"]` internally. This rewrite does not survive a streaming POST cleanly — it surfaces as `httpx.ReadError` during `initialize` on the Streamable HTTP transport.

## Root cause

The middleware calls `streamable_http_auth()` before the path rewrite, which may partially consume the ASGI `receive()` stream. For streaming POST requests (Streamable HTTP, SSE), after the path rewrite the downstream route handler expects to read the full request body, but the stream has already been partially consumed by the auth check. This results in corrupted or incomplete streaming data.

Regular JSON POST requests with complete bodies are typically buffered and survive this, but chunked/streaming POSTs do not.

## Current workaround

Tests and clients send requests to `/mcp/` directly (with trailing slash) to bypass the rewrite path entirely. This is documented in a comment in `tests/unit/mcpgateway/transports/test_streamablehttp_transport.py`:

```python
# Trailing slash matters: ContextForge's MCPPathRewriteMiddleware rewrites
# /mcp to /mcp/, but the rewrite doesn't survive a streaming POST cleanly
# (surfaces as httpx.ReadError during initialize). Send /mcp/ directly.
```

## Impact

- Clients hitting `/mcp` (without trailing slash) on the Streamable HTTP transport may get `httpx.ReadError` during session initialization
- The `/servers/<id>/mcp` virtual server path also goes through this rewrite and is likely affected
- The workaround (using `/mcp/` directly) is fragile and non-obvious

## Affected code

- `MCPPathRewriteMiddleware` in `mcpgateway/main.py` (~line 2890)
- `streamable_http_auth()` in `mcpgateway/transports/streamablehttp_transport.py` (~line 4157)

## Suggested fix

The middleware should either:
1. Buffer or tee the `receive()` stream so auth consumption doesn't affect downstream handlers
2. Issue a proper HTTP 307 redirect for non-streaming requests and reject or buffer streaming requests appropriately
3. Move authentication to happen after the path rewrite rather than before it

Genmin:
## Summary
- rewrite `/servers/{id}/mcp` to the mounted `/mcp/` path before streamable HTTP auth runs
- preserve the original app-relative path in `scope["modified_path"]` so server-scoped OAuth/RBAC still sees `/servers/{id}/mcp`
- add regression coverage for rewrite-before-auth ordering and auth lookup through `modified_path`

Fixes #4275.

## Testing
- `uv run pytest tests/unit/mcpgateway/test_main_extended.py::TestMCPPathRewriteMiddleware -q`
- `uv run pytest tests/unit/mcpgateway/transports/test_streamablehttp_transport.py -q -k streamable_http_auth`
- `uv run ruff check mcpgateway/main.py mcpgateway/transports/streamablehttp_transport.py`
- `uv run ruff check tests/unit/mcpgateway/test_main_extended.py --select E,F`
- `git diff --check`

### DIFF
--- a/mcpgateway/main.py
+++ b/mcpgateway/main.py
@@ -2671,11 +2671,11 @@ class AdminAuthMiddleware(BaseHTTPMiddleware):
 
 class MCPPathRewriteMiddleware:
     """
-    Middleware that rewrites paths ending with '/mcp' to '/mcp/', after performing authentication.
+    Middleware that rewrites paths ending with '/mcp' to '/mcp/' before authentication.
 
     - Rewrites paths like '/servers/<server_id>/mcp' to '/mcp/'.
     - Only paths ending with '/mcp' or '/mcp/' (but not exactly '/mcp' or '/mcp/') are rewritten.
-    - Authentication is performed before any path rewriting.
+    - Authentication uses ``scope["modified_path"]`` to preserve server-scoped decisions after rewriting.
     - If authentication fails, the request is not processed further.
     - All other requests are passed through without change.
     - Routes through the middleware stack (including CORSMiddleware) for proper CORS preflight handling.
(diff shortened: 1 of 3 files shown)


jonpspri:
Closing because the linked issue #4275 ([BUG]: MCPPathRewriteMiddleware breaks streaming POST when rewriting /mcp to /mcp/) was closed as **NOT_PLANNED**. This is also a duplicate of PR #4282, which is being closed for the same reason.

PR #4554, opened 2026-05-01
Fix regex timeout detection

### LINKED ISSUE #4509: [BUG]: standard `re.search()` does not support `timeout` keyword argument
### Summary

The test suite passes cleanly on Python 3.12 but **over a hundred tests fail on Python 3.13**, all due to the same root cause: `ContentSecurityService` incorrectly assumes that Python 3.13's `re.search()` has a `timeout` keyword argument.

### Root Cause

In `mcpgateway/services/content_security.py`:

```python
_HAS_NATIVE_REGEX_TIMEOUT: bool = sys.version_info >= (3, 13)
```

This flag enables a code path that passes `timeout=` to `compiled.search()`:

```python
if _HAS_NATIVE_REGEX_TIMEOUT:
    match = compiled.search(normalized_content, timeout=regex_timeout)
else:
    match = self._regex_search_with_timeout(compiled, normalized_content, timeout=regex_timeout)
```

**The `timeout` parameter was never added to Python 3.13's `re` module.** The stdlib `re.search()` signature remains `(pattern, string, flags=0)` in all 3.13.x releases. The `timeout` keyword is a feature of the third-party [`regex`](https://pypi.org/project/regex/) package, not stdlib `re`.

This causes a `TypeError: search() got an unexpected keyword argument 'timeout'` on every call to `detect_malicious_patterns()` and `validate_prompt_template()`, which are invoked during tool, resource, and prompt creation.

### Impact

- **149 tests failures** across 11 test files (tested on `6a793a9be9892ba1bb0202e4446c66463e80e27f`)
- **All tool, resource, and prompt creation is broken** on Python 3.13: not just tests, but the actual service would fail
- The thread-based fallback (`_regex_search_with_timeout`) works correctly on both 3.12 and 3.13

**Affected test files (by failure count):**

| Test File | Failures |
|-----------|----------|
| `tests/e2e/test_main_apis.py` | 29 |
| `tests/unit/.../test_tool_service_coverage.py` | 26 |
| `tests/unit/.../test_resource_service.py` | 19 |
| `tests/unit/.../test_prompt_service.py` | 19 |
| `tests/unit/.../test_content_pattern_detection.py` | 17 |
| `tests/unit/.../test_content_security.py` | 11 |
| `tests/unit/.../test_tool_service.py` | 11 |
| `tests/e2e/test_admin_apis.py` | 9 |
| `tests/unit/.../test_import_service.py` | 6 |
| `tests/unit/.../test_row_level_locking.py` | 1 |
| `tests/unit/.../test_a2a_service.py` | 1 |

### Example Failure

```
TypeError: search() got an unexpected keyword argument 'timeout'

  File "mcpgateway/services/content_security.py", line 698, in detect_malicious_patterns
    match = compiled.search(normalized_content, timeout=regex_timeout)
```

### Environment

- Python 3.13.7, macOS (also likely affects Linux)
- Python 3.12.11 passes all tests
- SQLite 3.50.4 (same version on both)
- All dependencies install successfully on both versions

### Suggested Fix

Set `_HAS_NATIVE_REGEX_TIMEOUT = False` unconditionally (or remove the flag entirely) until a Python version actually ships with `re.search(..., timeout=)`. The existing thread-based `_regex_search_with_timeout` fallback works on all Python versions.

Alternatively, if the intent was to use the third-party [`regex`](https://github.com/mrabarnett/mrab-regex) package (which does support `timeout`), the code should import and use `regex` instead of `re` when available, rather than gating on `sys.version_info`.

### Reproduction

```bash
# Passes
uv run --python 3.12 --extra plugins pytest tests/e2e/test_admin_apis.py::TestAdminToolAPIs::test_admin_tool_name_conflict -x

# Fails
uv run --python 3.13 --extra plugins pytest tests/e2e/test_admin_apis.py::TestAdminToolAPIs::test_admin_tool_name_conflict -x
```
### Additional information

Bug introduced in: #4072
Related: #2479

cc: @jonpspri

Genmin:
## Summary
- replace the Python-version regex timeout assumption with feature detection against stdlib `re.Pattern.search`
- keep the existing thread-based timeout fallback active on Python 3.13, where `timeout=` is still unsupported
- add a regression test for the stdlib timeout capability check

Fixes #4509.

## Validation
- `uv run pytest tests/unit/mcpgateway/services/test_content_pattern_detection.py -q`
- `uv run --python 3.13 pytest tests/unit/mcpgateway/services/test_content_pattern_detection.py -q`
- `uv tool run ruff==0.15.1 check mcpgateway/services/content_security.py tests/unit/mcpgateway/services/test_content_pattern_detection.py`
- `uv tool run black==26.3.1 --check mcpgateway/services/content_security.py tests/unit/mcpgateway/services/test_content_pattern_detection.py`
- `git diff --check`

### DIFF
--- a/mcpgateway/services/content_security.py
+++ b/mcpgateway/services/content_security.py
@@ -16,7 +16,6 @@ from issue #538.
 import hashlib
 import logging
 import re
-import sys
 import threading
 from typing import List, Optional, Union
 
@@ -49,10 +48,20 @@ except ImportError:
     content_size_violations_counter = NoOpCounter()
     content_type_violations_counter = NoOpCounter()
 
-# re.search() gained a `timeout` keyword in Python 3.13 that actually aborts
-# pathological regex execution. Older versions only have the thread.join
-# fallback, which is a soft timeout (see _regex_search_with_timeout).
-_HAS_NATIVE_REGEX_TIMEOUT: bool = sys.version_info >= (3, 13)
+
+def _supports_native_regex_timeout() -> bool:
+    """Return whether stdlib re.Pattern.search accepts timeout=."""
+    try:
+        re.compile("").search("", timeout=0.01)  # type: ignore[call-arg]
+    except TypeError:
+        return False
+    return True
+
+
+# If stdlib re ever grows a native timeout, prefer it because it can abort
+# pathological regex execution. Current Python releases still require the
+# thread.join fallback (see _regex_search_with_timeout).
+_HAS_NATIVE_REGEX_TIMEOUT: bool = _supports_native_regex_timeout()
 
 logger = logging.getLogger(__name__)
 
--- a/tests/unit/mcpgateway/services/test_content_pattern_detection.py
+++ b/tests/unit/mcpgateway/services/test_content_pattern_detection.py
@@ -17,7 +17,7 @@ from unittest.mock import patch
 import pytest
 
 # First-Party
-from mcpgateway.services.content_security import ContentPatternError, ContentSecurityService
+from mcpgateway.services.content_security import _supports_native_regex_timeout, ContentPatternError, ContentSecurityService
 
 
 class TestMaliciousPatternDetection:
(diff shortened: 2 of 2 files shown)


jonpspri:
The changes from this PR were already merged into `main` as commit `11292f497` ("fix(security): detect regex timeout support at runtime"). Closing this stale open PR — issue #4509 is now closed as resolved.

PR #4561, opened 2026-05-01
feat(ui-rewrite): add useMCPServerForm hook for creating and editing MCP servers

marekdano:
## 🔗 Related Issue
Related #4175

---

## 📝 Summary
Add a hook for creating and editing MCP servers. The hook contains calls to API endpoints to create or edit MCP Servers.

https://github.com/user-attachments/assets/29d03e38-bd22-4910-97b6-d8698e34e88e

---

## 🏷️ Type of Change
- [ ] Bug fix
- [x] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -22,7 +22,7 @@
         "shadcn": "^4.2.0",
         "tailwind-merge": "^3.5.0",
         "tw-animate-css": "^1.4.0",
-        "zod": "^4.4.1"
+        "zod": "^3.25.76"
       },
       "devDependencies": {
         "@formatjs/cli": "^6.14.1",
@@ -11088,15 +11088,6 @@
         "shadcn": "dist/index.js"
       }
     },
-    "node_modules/shadcn/node_modules/zod": {
-      "version": "3.25.76",
-      "resolved": "https://registry.npmjs.org/zod/-/zod-3.25.76.tgz",
-      "integrity": "sha512-gzUt/qt81nXsFGKIFcC3YnfEAx5NkunCfnDlvuBSSFS02bcXu4Lmea0AFIUwbLWxWPx3d9p8S5QoaujKcNQxcQ==",
-      "license": "MIT",
-      "funding": {
-        "url": "https://github.com/sponsors/colinhacks"
-      }
-    },
     "node_modules/shebang-command": {
       "version": "2.0.0",
       "resolved": "https://registry.npmjs.org/shebang-command/-/shebang-command-2.0.0.tgz",
@@ -13879,9 +13870,9 @@
       }
     },
     "node_modules/zod": {
-      "version": "4.4.1",
-      "resolved": "https://registry.npmjs.org/zod/-/zod-4.4.1.tgz",
-      "integrity": "sha512-a6ENMBBGZBsnlSebQ/eKCguSBeGKSf4O7BPnqVPmYGtpBYI7VSqoVqw+QcB7kPRjbqPwhYTpFbVj/RqNz/CT0Q==",
+      "version": "3.25.76",
+      "resolved": "https://registry.npmjs.org/zod/-/zod-3.25.76.tgz",
+      "integrity": "sha512-gzUt/qt81nXsFGKIFcC3YnfEAx5NkunCfnDlvuBSSFS02bcXu4Lmea0AFIUwbLWxWPx3d9p8S5QoaujKcNQxcQ==",
       "license": "MIT",
       "funding": {
         "url": "https://github.com/sponsors/colinhacks"
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -21,7 +21,8 @@
         "react-intl": "^10.1.1",
         "shadcn": "^4.2.0",
         "tailwind-merge": "^3.5.0",
-        "tw-animate-css": "^1.4.0"
+        "tw-animate-css": "^1.4.0",
+        "zod": "^4.4.1"
       },
       "devDependencies": {
         "@formatjs/cli": "^6.14.1",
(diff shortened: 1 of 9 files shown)


a-effort:
🤖 Sonnet 4.6 review feedback: 

Good structure here: extracting form logic into a dedicated hook with Zod validation and sanitization is the right call. A few things worth flagging...
 
1. Zod v4 is a breaking change: package.json bumps to zod@^4.4.1, but shadcn has a nested zod@3.25.76 as a peer dependency (visible in package-lock.json). Zod v4 has breaking API changes. If shadcn uses zod internally for form validation (e.g., via react-hook-form + zod resolvers), this version split will cause runtime issues. Worth verifying compatibility with the full shadcn setup.

 2. sanitizeCertificate calls .trim() on the result (sanitize.ts:130)... PEM certs sometimes require a trailing `\n` after `-----END CERTIFICATE-----` and some TLS parsers reject trimmed certs. Easy fix now, easy to forget later.
 
 3. There's a stray `// Made with Bob` comment at the end of useMCPServerForm.test.ts that should be removed.

Worth noting before this goes to production: 
- handleSubmit is typed as returning void in the interface but the implementation is async. It should be Promise<void>.         
- isValid only checks that name/URL are non-empty, not that the URL is valid (acceptable for a prototype but will need   tightening for real validation UX).                                                                                            
 - passthroughHeaders is stored as a comma-separated string but the Zod schema expects z.array(z.string()), so               
  validateField("passthroughHeaders", ...) won't behave as expected. Worth a // TODO comment so it's not a surprise when edit mode gets wired up. Same goes for the missing pre-population logic from existing gateway data.                             
 - The updateGateway fallback URL ("/gateways/" when gatewayId is undefined) won't be reached in current code, but it's a footgun if the logic shifts. A comment or assertion would help.

vishu-bh (changes requested):
Hey @marekdano 👋 — good direction overall. Extracting the form logic into a
dedicated hook with Zod validation is the right architectural call, and the
RouterProvider test fix is a solid improvement. That said, there are a few things that I noticed

---

### 🔴 Blockers

**Zod v4 / v3 conflict**
`package.json` bumps to `zod@^4.4.1`, but shadcn carries a peer dep on
`zod@3.25.76`. Zod v4 has breaking API changes — the `@hookform/resolvers/zod`
adapter is built against v3, so this will silently break form validation at
runtime. We need to either pin to v3 to match shadcn, or fully audit resolver
compatibility before upgrading.

**`passthroughHeaders` type mismatch**
The Zod schema defines this field as `z.array(z.string())`, but the form stores
it as a comma-separated string. `validateField("passthroughHeaders", ...)` will
pass a `string` to an array validator — either silently passing garbage or
throwing unexpectedly. This will also break edit mode when pre-population is
wired up. Suggest parsing the comma-string to an array on change or at
submission time, and documenting the approach.

**`handleSubmit` typed as `void` but implemented as `async`**
The interface declares `handleSubmit: () => void`, masking the promise.
Consumers won't await it, and any thrown error (e.g., a failed API call) will
be an unhandled rejection — no error toast, no UI feedback. Please update the
type to `() => Promise<void>` and make sure callers handle the promise.

---

### 🟠 Should Fix

**`updateGateway` with undefined `gatewayId`**
If `gatewayId` is `undefined`, the update URL resolves to `/gateways/undefined`
— a silently wrong API call. Add a guard that throws (or returns early) before
the fetch if `gatewayId` is not set.

**`sanitizeCertificate` trims trailing newline**
PEM certs require a trailing `\n` after `-----END CERTIFICATE-----`.
Calling `.trim()` removes it, producing a cert that looks valid but fails
TLS parsing downstream. Suggest replacing `.trim()` with `.trimStart()` and
ensuring the output always ends with `\n`.

---

### 🟡 Nice to Fix

- **`// Made with Bob`** comment at the end of `useMCPServerForm.test.ts` —
  please remove.
- **`console.log` in file-upload handler** — the test asserts it's called,
  which means it's in the production component. Replace with proper state
  or remove it.
- **`isValid` URL check** — currently just checks for non-empty string.
  A `new URL(url)` guard (or `z.string().url()`) would give users a much
  clearer error when they type an invalid URL.
- **Edit mode pre-population** — `gatewayId` is accepted but existing data
  is never loaded. If this is a known gap, please add a `// TODO` with
  the tracking issue so it's not forgotten.

---

Happy to pair on the Zod version story if useful — that one has some nuance.
Rest looks clean. 🙌

vishu-bh (changes requested):
Thanks for the follow-up pass here — this is moving in a good direction. Extracting the form logic into `useMCPServerForm` makes the component much easier to follow, and the Zod version + `passthroughHeaders` fixes address a couple of the main concerns from the previous review.

I went back through the earlier comments, and most of the blocking feedback now looks resolved. There’s one piece I still don’t think is fully closed yet:

### Remaining blocker

**`handleSubmit` is now typed async in the hook, but `MCPServerForm.tsx` still treats it as fire-and-forget**
In `client/src/hooks/useMCPServerForm.ts`, the contract is now correctly `Promise<void>`, but in `client/src/components/mcp-servers/MCPServerForm.tsx` the submit handler still just calls:

```ts
handleSubmit(event, () => {
  onToggle();
});
```

without awaiting it or explicitly handling the returned promise. So the type-level fix is there, but the consuming component still doesn’t fully honour the async contract. I’d like to see that wired through cleanly so we’re not leaving promise handling implicit at the call site.

### Still worth tightening up

- In `client/src/hooks/useMCPServerForm.ts`, the runtime guard for missing `gatewayId` is a good improvement, though the hook still initialises `useQuery` with `/gateways/${gatewayId}` even when `gatewayId` is undefined. That’s safer than before, but still a bit fragile.

Once the async submit integration is cleaned up, this should be in much better shape from my side.

vishu-bh (approved):
PR looks good now. 

LGTM ☘️

PR #4573, opened 2026-05-01
feat: React Client FastAPI Router

gcgoncalves:
# ✨ Feature / Enhancement PR

## 🔗 Epic / Issue
Relates to #4269

---

## 🚀 Summary (1-2 sentences)
Adds new endpoints to handle the react client authentication

---

## 🧪 Checks

- [X] `make lint` passes
- [X] `make test` passes
- [X] CHANGELOG updated (if user-facing)

---

## 📓 Notes (optional)
Moves the react client routes to a separate file

### DIFF
--- a/.github/workflows/client-e2e.yml
+++ b/.github/workflows/client-e2e.yml
@@ -63,12 +63,10 @@ jobs:
 
       - name: 🎭 Install Playwright browsers
         if: steps.playwright-cache.outputs.cache-hit != 'true'
-        working-directory: ./client
         run: npx playwright install --with-deps chromium
 
       - name: 🎭 Install Playwright system deps (cached browsers)
         if: steps.playwright-cache.outputs.cache-hit == 'true'
-        working-directory: ./client
         run: npx playwright install-deps chromium
 
       - name: 🧪 Run Playwright tests
--- a/.github/workflows/client-e2e.yml
+++ b/.github/workflows/client-e2e.yml
@@ -63,10 +63,12 @@ jobs:
 
       - name: 🎭 Install Playwright browsers
         if: steps.playwright-cache.outputs.cache-hit != 'true'
+        working-directory: ./client
         run: npx playwright install --with-deps chromium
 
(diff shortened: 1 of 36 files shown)


marekdano (changes requested):
#### Blocking issues — worth a follow-up comment before approving:

  1. `_validate_token_expiry_sync()` is a no-op check. `TOKEN_EXPIRY_MINUTES = settings.token_expiry` and then `if
  TOKEN_EXPIRY_MINUTES != settings.token_expiry` will always be false. It gives false confidence on expiry synchronization. The `CSRF_TOKEN_LENGTH` check is fine, but the `JWT/CSRF` sync check should be removed or replaced with something meaningful (the E2E test in `test_app_auth_token_expiry.py` already verifies this better).
  2. `AuthGuard` returns null while `isLoading` — blank page on every cold load. Not a security bug, but a visible UX regression worth fixing.

####  Non-blocking suggestions (can go in a follow-up PR):
  - Extract duplicated E2E fixtures to `tests/e2e/conftest.py`
  - Replace `void token` in `setToken()` with an empty body
  - Rename `APP_COOKIE_PATH` to `JWT_COOKIE_PATH` or `APP_JWT_COOKIE_PATH` to reflect that it's the JWT cookie's path
  (not the CSRF cookie's path).
  - `test_rate_limit_is_per_ip` asserts nothing — either fix or remove it

marekdano (changes requested):
The PR looks good! I have just one request

 JWT cookie path must be `/app`
  `JWT_COOKIE_PATH = "/"` in `routers/app.py` means the session cookie is sent to every server endpoint (/api/*, /mcp, /gateways, /admin, etc.), not just the React SPA. The CSRF cookie correctly scopes to `path=/app`; the JWT cookie must match. This directly undermines the security hardening the PR is designed to deliver.

marekdano (approved):
LGTM 🚀

gcgoncalves:
> Reference

so, that's the main issue here: we need access to the API via cookies. 😢

Issue #4581, opened 2026-05-03
chore(plugins): align rate-limiter config with gateway convention


gandhipratik203:
### Observation

The gateway sources its own Redis URL from the `REDIS_URL` environment variable via pydantic `Settings` ([`mcpgateway/config.py:1965`](https://github.com/IBM/mcp-context-forge/blob/main/mcpgateway/config.py#L1965)).  The rate-limiter plugin's `plugins/config.yaml` block hardcodes the same URL as a literal string:

```yaml
config:
  backend: "redis"
  redis_url: "redis://redis:6379/0"
  ...
```

### Why this is worth aligning

Operators deploying the gateway today set `REDIS_URL` once and the gateway picks it up.  The plugin yaml ignores that env var and forces a second, separate edit if they want to point the rate limiter at a different Redis (different host, different DB number, different deployment topology).  One env var → two configuration surfaces is mildly surprising and creates drift over time.

The framework's plugin loader already supports Jinja env substitution ([`mcpgateway/plugins/framework/loader/config.py:81`](https://github.com/IBM/mcp-context-forge/blob/main/mcpgateway/plugins/framework/loader/config.py#L81)), so the alignment is a one-line yaml change with no framework work needed.

### Proposed change

```yaml
config:
  backend: "redis"
  redis_url: "{{ env.REDIS_URL }}"
  ...
```

Backwards-compatible: `docker-compose.yml:276` already exports `REDIS_URL=redis://redis:6379/0`, so the resolved value matches the current literal default.  Operators overriding `REDIS_URL` get a single source of truth across the gateway and the plugin.

### Out of scope

- Any URL log sanitization or related observability changes.
- Any plugin-loader changes (Jinja substitution already supported).
- Other plugin yaml entries — this issue is scoped to the rate-limiter block.

PR #4582, opened 2026-05-04
chore(plugins): align rate-limiter config with gateway convention

gandhipratik203:
## Summary

Align the rate-limiter plugin's `redis_url` in `plugins/config.yaml` with the gateway's existing convention of sourcing Redis URLs from the `REDIS_URL` env via the plugin loader's Jinja substitution. Operators set `REDIS_URL` once and the rate-limiter picks it up — no more hand-editing the yaml or templating it in deployment tooling.

Companion to #4581.

## Changes

- `plugins/config.yaml`: `redis_url: "redis://redis:6379/0"` → `redis_url: '{{ env.REDIS_URL | default("redis://redis:6379/0") }}'` plus a short comment about the env-source. Outer single / inner double quotes survive `yaml.safe_dump` round-trips cleanly.
- `tests/integration/test_rate_limiter_redis_url_from_yaml.py`: new test — env-var substitution flows through the plugin loader cleanly and the resolved URL reaches a live Redis (PING → PONG). Skips when Redis isn't reachable.

## Test plan

- [x] `pytest tests/integration/test_rate_limiter_redis_url_from_yaml.py` green locally with Redis running
- [x] Skips cleanly without Redis
- [ ] CI green

The schema-regression test and `redis_fallback` cleanup originally bundled here are now tracked as #4603.

### DIFF
--- a/plugins/config.yaml
+++ b/plugins/config.yaml
@@ -286,8 +286,8 @@ plugins:
     conditions: []
     config:
       backend: "redis"
-      # Sourced from the gateway's REDIS_URL env. When run
-      # externally (separate container or WASM sandbox), extra wiring may be needed.
+      # Sourced from the gateway's REDIS_URL env; when the plugin runs in a
+      # separate container, ensure that host is reachable from within it.
       redis_url: '{{ env.REDIS_URL | default("redis://redis:6379/0") }}'
       redis_key_prefix: "rl"
       redis_fallback: true
--- a/plugins/config.yaml
+++ b/plugins/config.yaml
@@ -288,7 +288,7 @@ plugins:
       backend: "redis"
       # Sourced from the gateway's REDIS_URL env. When run
       # externally (separate container or WASM sandbox), extra wiring may be needed.
-      redis_url: "{{ env.REDIS_URL }}"
+      redis_url: '{{ env.REDIS_URL | default("redis://redis:6379/0") }}'
       redis_key_prefix: "rl"
       redis_fallback: true
       algorithm: "fixed_window"
--- a/plugins/config.yaml
+++ b/plugins/config.yaml
@@ -286,7 +286,9 @@ plugins:
     conditions: []
     config:
       backend: "redis"
-      redis_url: "redis://redis:6379/0"
+      # Sourced from the gateway's REDIS_URL env. When run
+      # externally (separate container or WASM sandbox), extra wiring may be needed.
+      redis_url: "{{ env.REDIS_URL }}"
       redis_key_prefix: "rl"
       redis_fallback: true
       algorithm: "fixed_window"
--- a/tests/integration/test_rate_limiter_redis_url_from_yaml.py
+++ b/tests/integration/test_rate_limiter_redis_url_from_yaml.py
@@ -20,6 +20,7 @@ runners without a Redis service available.
 """
 
 # Standard
+import pathlib
 import socket
 
 # Third-Party
@@ -28,6 +29,11 @@ import redis
 
 # First-Party
 from mcpgateway.plugins.framework.loader.config import ConfigLoader
+
+# Anchor the plugins/config.yaml path to this file's location so the test
+# works regardless of where pytest is invoked from (repo root, tests/, etc.).
+_REPO_ROOT = pathlib.Path(__file__).resolve().parents[2]
+_CONFIG_YAML = str(_REPO_ROOT / "plugins" / "config.yaml")
 
 
 def _redis_reachable(host: str = "127.0.0.1", port: int = 6379, timeout: float = 0.2) -> bool:
@@ -50,7 +56,7 @@ def test_rate_limiter_redis_url_resolves_and_connects(monkeypatch):
     """
     monkeypatch.setenv("REDIS_URL", "redis://127.0.0.1:6379/0")
 
-    cfg = ConfigLoader.load_config("plugins/config.yaml")
+    cfg = ConfigLoader.load_config(_CONFIG_YAML)
     rl = next(p for p in cfg.plugins if p.name == "RateLimiterPlugin")
     resolved = rl.config.get("redis_url")
 
@@ -66,3 +72,20 @@ def test_rate_limiter_redis_url_resolves_and_connects(monkeypatch):
         assert client.ping() is True
     finally:
         client.close()
+
+
+def test_rate_limiter_redis_url_uses_default_when_env_unset(monkeypatch):
+    """When REDIS_URL is unset, the Jinja ``default(...)`` filter must
+    resolve to the literal fallback baked into ``plugins/config.yaml``.
+
+    Pins the contract that the loader's Jinja env applies the ``default``
+    filter (not just plain env-var substitution) — a regression here would
+    silently leave operators on a broken redis_url whenever the env var
+    isn't injected.  Runs without Redis.
+    """
+    monkeypatch.delenv("REDIS_URL", raising=False)
+
+    cfg = ConfigLoader.load_config(_CONFIG_YAML)
+    rl = next(p for p in cfg.plugins if p.name == "RateLimiterPlugin")
+
+    assert rl.config.get("redis_url") == "redis://redis:6379/0"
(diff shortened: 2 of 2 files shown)


msureshkumar88:
Me review your PR! Changes GOOD idea but have few problems that MUST fix before merge.

### 🔴 CRITICAL Issues (Must Fix Now!)

**1.  Bare Exception Swallow All Errors** 🚨
- Location: test_plugins_config_yaml_validation.py lines 116-120
- Problem: Code catch ALL exceptions, even security failures!
- Attack: Bad plugin could bypass rate limiting if fail to load
- Fix: Only catch connection errors, let security failures FAIL test

**2. No URL Validation** 🚨  
- Location: test_plugins_config_yaml_validation.py lines 54-61
- Problem: No check if redis_url is safe! Could be SSRF attack vector
- Attack: REDIS_URL pointing to internal-admin service could steal data
- Fix: Validate URL scheme is redis, check hostname not internal, mask credentials

**3. Breaking Change: redis_fallback Removal Not Justified** ⚠️
- You remove redis_fallback: true but:
  - Still in docs (plugin-bindings-api.md line 345)
  - Still in Helm charts (values-pgo.yaml line 101)  
  - Still in tests (test_rate_limiter.py line 1267)
  - Existing test validate this behavior!
- Impact: If Redis down, what happen? Fail open (no rate limit) or fail closed (deny all)?
- Fix: Show evidence key is unrecognized OR revert change OR update ALL docs/tests

### 🟡 Should Fix (Recommended)

4. **Performance**: Use async Redis client instead of blocking sync call
5. **Testing Gap**: Add negative tests (bad URLs, missing env vars, malformed Jinja)
6. **Logging Gap**: Add audit trail when config changes


Read full report for:
- All issues with code examples
- Alternative approaches you could use
- Risk assessment matrix
- Specific code fixes for each issue

### ✅ What Need Before Merge

**MUST FIX (Blocking):**
1. Fix bare exception handling 
2. Add URL validation
3. Justify redis_fallback removal OR revert (breaking change)
4. Update ALL documentation consistently

**SHOULD FIX:**
5. Use async Redis client
6. Add negative test cases
7. Add audit logging

### 🎯 Verdict

⚠️ **REQUEST CHANGES** - Critical issues found

Estimated effort: 4-6 hours to fix all critical issues

msureshkumar88 (changes requested):
As mentioned in comments

gandhipratik203:
Thanks for the review. Scope-narrowing this PR to just the `redis_url` Jinja substitution: the schema-validation regression test and `redis_fallback` removal that triggered most of these comments have moved to #4603, which catalogues the broader cleanup (docs, helm chart, perf configs, unit fixtures, and a misleadingly-named integration test). Starting-point branch: [`chore/rate-limiter-redis-fallback-cleanup-wip`](https://github.com/IBM/mcp-context-forge/tree/chore/rate-limiter-redis-fallback-cleanup-wip).

That moots 1 and 3 of your review (both lived in the code that's been moved out).

2 (SSRF on `REDIS_URL`): the test sets the env var itself via `monkeypatch.setenv` to localhost — there is no untrusted input crossing a trust boundary, so SSRF doesn't apply. Validating a value the test just wrote would be testing our own hardcoded string.

4–6: not applicable in the narrowed scope. A one-shot Redis `PING` doesn't need an async client; a positive-path test that env-var substitution flows through the loader cleanly doesn't need adversarial inputs; nothing mutates config at runtime here so there's no audit surface.

msureshkumar88 (changes requested):
Good direction — the Jinja `| default(...)` pattern matches what the rest of `config.yaml` already does for `IBM_WATSON_*`, and the outer-single/inner-double quoting to survive `yaml.safe_dump` round-trips is a nice touch. A couple of things worth addressing before this lands:

## Blocking

**Relative path in the integration test**

```python
cfg = ConfigLoader.load_config("plugins/config.yaml")   # test line 56
```

This resolves from `cwd`, so the test silently breaks whenever pytest is invoked from any directory other than the repo root (e.g. `cd tests && pytest`). Suggest anchoring to the file's own location:

```python
import pathlib
_REPO_ROOT = pathlib.Path(__file__).resolve().parents[3]
cfg = ConfigLoader.load_config(str(_REPO_ROOT / "plugins" / "config.yaml"))
```

**Security Scan CI failure**

25/28 checks pass but the Security Scan job is failing. Worth checking whether Bandit is flagging something in the new test file (e.g. the bare `except Exception` in `_redis_reachable`, or the `socket` usage) and either suppressing with a `# nosec` + justification comment or adjusting the code. Happy to take a look at the job log if you can share it.

## Suggestions

**No test for the `default` fallback path**

The test always injects `REDIS_URL` via `monkeypatch.setenv`, so the `| default("redis://redis:6379/0")` branch is never exercised. A small unit test (no live Redis needed) would pin that:

```python
def test_rate_limiter_redis_url_uses_default_when_env_unset(monkeypatch, tmp_path):
    monkeypatch.delenv("REDIS_URL", raising=False)
    cfg = ConfigLoader.load_config("plugins/config.yaml")
    rl = next(p for p in cfg.plugins if p.name == "RateLimiterPlugin")
    assert rl.config.get("redis_url") == "redis://redis:6379/0"
```

(Same relative-path caveat applies here.)

## Minor

The comment in `config.yaml` mentions WASM sandboxes — a bit surprising in a Redis URL context, since WASM plugins typically can't open TCP sockets. Trimming to just "separate container" would avoid confusion:

```yaml
# Sourced from the gateway's REDIS_URL env; when the plugin runs in a
# separate container, ensure that host is reachable from within it.
```

---

One pre-existing issue worth tracking separately (not blocking this PR): `SandboxedEnvironment` in `config.py:80` is constructed with `autoescape=True`, which HTML-encodes `&`, `<`, `>`, `"`. That's correct for HTML templates but wrong for YAML — if `REDIS_URL` ever includes a percent-encoded password or similar, the rendered value would be silently corrupted. Probably worth a follow-up issue since it affects all Jinja-templated values in the config, not just this one.

gandhipratik203:
Thanks — this was a great review.  Addressed all four actionable items in [`aa5ded9`](https://github.com/IBM/mcp-context-forge/commit/aa5ded996); each backported to the WIP branch for #4603 in [`ee9436e`](https://github.com/IBM/mcp-context-forge/commit/ee9436edb) so the path-anchor fix carries forward when that work picks up.

**Relative path** — anchored via `pathlib.Path(__file__).resolve().parents[2]`. Verified the test passes both from repo root and from `tests/`.

**Security Scan failure** — the job log says `The job was not acquired by Runner of type hosted even after multiple attempts` (queued 1h30m, abandoned). Infra, not code. The push to `aa5ded9` re-triggered CI; the new Security Scan run is in progress.

**Default fallback test** — added `test_rate_limiter_redis_url_uses_default_when_env_unset` exactly as suggested. Runs without Redis.

**WASM comment** — trimmed to the separate-container case as suggested. You're right that WASM plugins can't open TCP sockets, so `REDIS_URL` doesn't apply to that path.

**autoescape pre-existing bug** — filed as #4605. Sharp catch; agree it's broader than this PR (affects every Jinja-templated value, not just `redis_url`) and warrants its own fix.

msureshkumar88 (approved):
Solid work scoping this down to just the Jinja substitution — that's the right call. All four actionable items from the original review are addressed and CI is green.

**What's good**
- Path anchored via `pathlib.Path(__file__).resolve().parents[2]` — test works from any invocation directory
- `test_rate_limiter_redis_url_uses_default_when_env_unset` pins the `default(...)` filter contract cleanly, runs without Redis
- WASM comment trimmed to what actually applies
- `redis_fallback` cleanup and schema-validation regression test correctly separated to #4603

**One non-blocking nit**
`test_rate_limiter_redis_url_from_yaml.py:70` still uses the sync `redis.from_url()` client. No event loop in this test so it won't cause flakiness, but worth converting to `redis.asyncio` in a follow-up if the test suite moves to async fixtures.

**On #4605 (`autoescape=True`)**
Good catch filing that separately. Worth prioritising — `config.py:80` silently corrupts any env-sourced value containing `&`, `<`, `>`, or `"` before `yaml.safe_load` sees it. A Redis password like `p & w` becomes `p & amp; w` and the connection fails with a confusing error. Fix is one line (`autoescape=False`) but the blast radius covers every Jinja-rendered config value across all plugins, so it warrants its own PR and test coverage. Happy to review that one when it's up.

Approving — this is mergeable as-is.

brian-hussey (approved):
LGTM

Issue #4530, opened 2026-04-29
[UI-REWRITE]: Add advanced settings to the form of creating a MCP server


marekdano:
Add "Advanced settings" to the form with the Authentication Type subforms

marekdano:
Done!

PR #4597, opened 2026-05-05
feat(ui-rewrite): add rest of advanced settings in MCP server form

marekdano:
## 🔗 Related Issue
Related: #4530 

---

## 📝 Summary
Add the advanced settings for 
- Bearer token
- Custom headers
- OAuth 2.0
- Query parameter

https://github.com/user-attachments/assets/6b631032-489b-4cb3-9664-4091908f6aa4

---

## 🏷️ Type of Change
- [ ] Bug fix
- [x] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

---

## ✅ Checklist

---

## 📓 Notes (optional)

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -11,6 +11,7 @@
         "@fontsource-variable/inter": "^5.2.8",
         "@radix-ui/react-dialog": "^1.1.15",
         "@radix-ui/react-dropdown-menu": "^2.1.2",
+        "@radix-ui/react-switch": "^1.1.2",
         "class-variance-authority": "^0.7.1",
         "clsx": "^2.1.1",
         "lucide-react": "^1.8.0",
--- a/client/package.json
+++ b/client/package.json
@@ -29,6 +29,7 @@
     "@fontsource-variable/inter": "^5.2.8",
     "@radix-ui/react-dialog": "^1.1.15",
     "@radix-ui/react-dropdown-menu": "^2.1.2",
+    "@radix-ui/react-switch": "^1.1.2",
     "class-variance-authority": "^0.7.1",
     "clsx": "^2.1.1",
     "lucide-react": "^1.8.0",
--- a/client/src/components/icons/MainNavIcon.tsx
+++ b/client/src/components/icons/MainNavIcon.tsx
@@ -19,5 +19,3 @@ export const MainNavIcon = ({ className }: { className?: string }) => (
     />
   </svg>
 );
-
-// Made with Bob
--- a/client/src/components/mcp-servers/AdvancedSettings.tsx
+++ b/client/src/components/mcp-servers/AdvancedSettings.tsx
@@ -1,4 +1,4 @@
-import { Info } from "lucide-react";
+import { Info, TriangleAlert } from "lucide-react";
 import { Textarea } from "@/components/ui/textarea";
 import { Switch } from "@/components/ui/switch";
 import {
@@ -12,16 +12,13 @@ import { CACertificateUpload } from "@/components/mcp-servers/CACertificateUploa
 import { NoneAuth } from "@/components/mcp-servers/NoneAuth";
 import { BasicAuth } from "@/components/mcp-servers/BasicAuth";
 import { BearerTokenAuth } from "@/components/mcp-servers/BearerTokenAuth";
-import { CustomHeadersAuth } from "@/components/mcp-servers/CustomHeadersAuth";
+import { CustomHeadersAuth, type CustomHeader } from "@/components/mcp-servers/CustomHeadersAuth";
 import { OAuth2Auth } from "@/components/mcp-servers/OAuth2Auth";
 import { QueryParameterAuth } from "@/components/mcp-servers/QueryParameterAuth";
 
-type AuthType = "none" | "basic" | "bearer" | "custom" | "oauth" | "query";
+export type { CustomHeader };
 
-export interface CustomHeader {
-  key: string;
-  value: string;
-}
+type AuthType = "none" | "basic" | "bearer" | "custom" | "oauth" | "query";
 
 interface AdvancedSettingsProps {
   visibility: string;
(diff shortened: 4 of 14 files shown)


a-effort:
Good work getting all of this in... it's a BIG form, especially OAuth options! In the Figma, I may need to make adjustments to which fields are optional/not for this particular auth type depending on which options are entered. Will follow up with a proposal on that. 

We can fine tune small style things later ofc, but there are a couple of proposed changes to update now: 

- One-time authentication: [when toggled on](https://www.figma.com/design/H2Dr9QPvRVxCe92V8URvGT/Refiring?m=auto&node-id=2710-3067&t=o2azGnw4X4p3YEYk-1), update text color from muted-foreground to foreground and show inline alert below. Default state should be toggled off (user can toggle on)
- Token management checkboxes should [default to checked](https://www.figma.com/design/H2Dr9QPvRVxCe92V8URvGT/Refiring?m=auto&node-id=2539-7694&t=o2azGnw4X4p3YEYk-1) (user can uncheck)

-->
  🤖 Sonnet feedback: 
  
   The auth type switching, accessibility wiring, and the security warning on query params all look good.
  
  A few things worth addressing before this moves toward production:
  
  OAuth2Auth.tsx - **mixed controlled/uncontrolled state will silently drop updates**
 The component initializes local state from props (useState(grantType), useState(issuerUrl), etc.) and then tries to "switch" between local and parent-controlled mode based on whether a callback is present. The problem: useState only reads its initializer once, so if a parent re-renders with different prop values after mount, the component ignores them.

Since MCPServerForm doesn't pass the optional handlers (onGrantTypeChange, onIssuerUrlChange, etc.), those fields are effectively uncontrolled. The fix is to lift all OAuth state up into MCPServerForm like the other auth fields (bearerToken, oauthClientId, etc.) and remove the internal state from OAuth2Auth entirely.

 **OAuth optional fields not reset on cancel/submit**  
  resetForm() resets oauthClientId, oauthClientSecret, oauthTokenUrl but not grantType, issuerUrl, redirectUri,                 
  authorizationUrl, scopes, storeTokens, autoRefresh since those live inside the component. Closing and reopening the form will show stale values. This is directly caused by the _mixed controlled/uncontrolled state will silently drop updates_ issue.  
  
  **Worth fixing before production**
  key={index} in CustomHeadersAuth when mapping headers -- removing a row from the middle will cause React to reuse the wrong DOM nodes. Adding a stable id to each header entry (e.g. assigned in addHeader) fixes this.
  
 CustomHeader is defined both locally in CustomHeadersAuth.tsx and exported from AdvancedSettings.tsx. The local one should import from AdvancedSettings to avoid drift.
 
  customHeaders is initialised with one blank row in MCPServerForm regardless of which auth type is selected. Initialising to [] is cleaner.

vishu-bh (commented):
Hey @marekdano 👋 — the component split here is clean and the controlled-state
pattern for `CustomHeadersAuth` and `BearerTokenAuth` is the right approach.
A few things noticed in review

---

### 🔴 Blockers

**OAuth2Auth mixed controlled/uncontrolled state**
`OAuth2Auth` initializes `grantType`, `issuerUrl`, `redirectUri`, `scopes`,
`storeTokens`, and `autoRefresh` via `useState(prop)`. Since `MCPServerForm`
doesn't pass the optional callbacks for those fields, they are effectively
uncontrolled — their values never flow back to the parent and will never reach
the API call. `useState` only reads its initializer once, so if the parent
re-renders with different props, the component ignores them.

The fix is to lift all OAuth fields into `MCPServerForm`'s state and make
`OAuth2Auth` a fully controlled component (no internal `useState` at all),
the same pattern used by `BearerTokenAuth` and `CustomHeadersAuth`.

**`resetForm()` doesn't reset OAuth extended fields**
This is a direct consequence of the above — `grantType`, `issuerUrl`,
`scopes`, etc. live inside `OAuth2Auth` local state, so `resetForm()` can't
reach them. Close the form with OAuth details filled in, reopen it — those
values persist. Fixing the mixed-state issue resolves this too.

**`key={index}` in `CustomHeadersAuth`**
Deleting a row from the middle of the headers list will cause React to reuse
the wrong DOM nodes, leading to inputs showing the wrong data. Please assign a
stable `id` to each header on creation (e.g. `crypto.randomUUID()`) and use
that as the key instead.

---

### 🟠 Should Fix

**Duplicate `CustomHeader` type**
`CustomHeader` is defined locally in `CustomHeadersAuth.tsx` AND exported from
`AdvancedSettings.tsx`. Once a stable `id` field is added (per the `key` fix),
one definition will drift. Please remove the local one and import from
`AdvancedSettings` (or a shared `types.ts`).

**`// Made with Bob` comment in `AdvancedSettings.tsx`**
This was flagged in #4561 and it's still at the bottom of the file. Please
remove it.

**`customHeaders` initialized with a blank row**
`MCPServerForm` initializes `customHeaders: [{ key: "", value: "" }]` even
when the user selects a different auth type. This blank row will be included
in the form state at submit time and may be serialized into the API request
as an empty custom header. Initialise to `[]` and handle the empty state
inside `CustomHeadersAuth` with a placeholder message.

---

### 🟡 Nice to Fix

- **OAuth client secret** should use `type="password"` for consistency with
  `BearerTokenAuth` — both are sensitive credentials.
- **Custom header values** are rendered as `type="text"` but frequently
  carry tokens/API keys. Consider masking or adding a show/hide toggle.
- **`oneTimeAuth` Figma spec** — per @a-effort's comment, the toggle should
  update label text color and show an inline alert when activated. A
  `// TODO` at minimum so it's not forgotten.
- **Auth field validation** — none of the new components validate their
  inputs on submit (empty bearer token, malformed header key, invalid token
  URL). This can result in silent API errors.

---

The OAuth state lift is the main piece of work here — once that's done
the reset issue resolves itself and the overall architecture will be solid.

vishu-bh (changes requested):

vishu-bh (approved):
Thanks @marekdano. 

LGTM 🚀

PR #4604, opened 2026-05-05
feat: migrate Gateways component from /gateways to /servers endpoint

vishu-bh:
Migrates Gateways.tsx from legacy /gateways to new /servers endpoint with updated types, comprehensive tests, and new virtual servers listing UI.

Key Changes:
- Created BaseServer interface to eliminate field duplication
- Added VirtualServer type with camelCase fields matching /servers response
- Updated Gateways component to use /servers endpoint
- Added new virtual servers list UI with server cards showing status, tools, resources, and prompts
- Updated field mappings: reachable to enabled, tool_count to associatedToolIds.length
- Updated timestamps to camelCase: createdAt/updatedAt
- Fixed unit tests and added comprehensive E2E test suite (18 test cases)

Test Results: 63 unit tests passing, 30 E2E tests passing (1 skipped)


client/src/types/server.ts
```diff
@@ -1,26 +1,31 @@
...
+export interface MCPServer extends BaseServer {
   url: string;
-  description?: string;
   transport: "SSE" | "STREAMABLEHTTP";
-  enabled: boolean;
   reachable: boolean;
   last_seen?: string;
   tool_count: number;
   created_at: string;
   updated_at: string;
```
marekdano, line 28: I think APIs returns camelCase properties instead of snake_case

client/src/types/server.ts
```diff
@@ -1,26 +1,31 @@
+/**
+ * Base server interface with common fields
+ */
+export interface BaseServer {
+  id: string;
+  name: string;
+  description?: string;
+  enabled: boolean;
+  visibility: "private" | "team" | "public";
+  team_id?: string;
```
marekdano, line 10: camelCase?

client/src/types/server.ts
```diff
@@ -1,26 +1,31 @@
...
+  team?: string;
+  owner_email?: string;
```
marekdano, line 12: camelCase?

marekdano (changes requested):
The code looks good!

The formatting should be fixed by running `npm run format`

Checking the design:

Image

The format of the timestamp should be improved.

---

Image

The icon should be "upload" from Lucide icons.
The dropdown should have the "Delete" item.

Should we have "Test connection" option there?

vishu-bh:
> The code looks good!
> 
> The formatting should be fixed by running `npm run format`
> 
> Checking the design:
> 
> Image
> The format of the timestamp should be improved.
> 
> Image
> The icon should be "upload" from Lucide icons. The dropdown should have the "Delete" item.
> 
> Should we have "Test connection" option there?

Nice catch!!

Will address them. Thanks Marek.

client/src/types/server.ts
vishu-bh: I think for this one, this is response from the mcp server api

client/src/types/server.ts
vishu-bh: This is needed for MCP server list response.

client/src/types/server.ts
vishu-bh: This is needed for MCP server list response.

marekdano (changes requested):
#### Blocking issues:

  - `getServerTags` is a no-op wrapper (`Gateways.tsx:37-39`). It literally does return server.tags. This adds no abstraction — just use `server.tags` inline in `VirtualServerCard`. Dead helpers create confusion about what invariant they're protecting.
  - `useMemo` with no real dependencies is misleading (`Gateways.tsx:222-227`). The `queryPath` memo has an empty dependency array because `DEFAULT_PAGE_SIZE` is a module constant. This signals to readers "this has a meaningful memoization reason" when it doesn't. 
      - Either lift it to a module-level constant:
```bash
  const SERVERS_QUERY_PATH = `/servers?limit=${DEFAULT_PAGE_SIZE}&include_pagination=true`;
```

    or just build the string inline — it's one line.

  
  - `actionCards` array is recreated on every render (`Gateways.tsx:228-256`). It's declared inside the `Gateways` function with no dependency that changes, yet it's not memoized. Move it above the component or wrap in `useMemo`. This is a minor perf issue but matters as the component grows.


####  Non-blocking issues:

  - `VirtualServer required fields may be over-constrained (server.ts:70-85). Fields like `teamId, team, ownerEmail, createdBy, createdFromIp, createdVia, createdUserAgent` are typed as required non-nullable strings.
  If the API can ever return null for these (as it does for `modifiedBy, modifiedFromIp, etc.`), this will cause silent runtime failures. Cross-check against the backend schema. At a minimum, add `| null` where the backend permits it.
  - `associatedTools`: `string[]` vs `associatedToolIds: string[]` (server.ts:60-61). Having both in the same interface is confusing — the name associatedTools implies rich objects but holds strings. If it's always empty or redundant with associatedToolIds, consider dropping it or clarifying its purpose in a comment.

marekdano (approved):
@vishu-bh - The PR looks good!

LGTM 🚀

PR #4606, opened 2026-05-05
bugfix/edge-mode-health-mirror-convergence-4440

### LINKED ISSUE #4440: [BUG]: Protocol compliance test_health_mirrors_runtime_mode_state fails in Edge mode due to narrow convergence predicate
### 🐞 Bug Summary

`test_health_mirrors_runtime_mode_state` in `tests/protocol_compliance/test_runtime_mode.py` fails in Edge (Rust) mode because the convergence polling loop only checks `effective_mode`, but the subsequent assertion also requires `override_active` to match across endpoints.

In a multi-replica deployment (docker-compose runs 3 gateway replicas behind Nginx), the two HTTP requests (`GET /health` and `GET /admin/runtime/mcp-mode`) can land on different pods. When `boot_mode` is `edge` and one pod has received an override-to-`edge` via Redis pub/sub but the other hasn't yet:

- Both pods report `effective_mode="edge"` → the convergence loop exits immediately
- But `override_active` diverges: `True` on the pod that received the override, `False` on the pod that hasn't

---

### 🧩 Affected Component

- [x] Other: protocol compliance test harness + runtime state propagation

---

### 🔁 Steps to Reproduce

1. Run protocol compliance tests against a multi-replica Edge-mode deployment
2. `pytest tests/protocol_compliance/test_runtime_mode.py::test_health_mirrors_runtime_mode_state`

---

### 🤔 Expected Behavior

The test should wait for all four compared fields (`boot_mode`, `effective_mode`, `override_active`, `cluster_propagation`) to converge before asserting, not just `effective_mode`.

---

### 📓 Logs / Error Output

```
tests/protocol_compliance/test_runtime_mode.py:149: in test_health_mirrors_runtime_mode_state
    assert mcp_rt.get(key) == admin.get(key), f"mcp_runtime.{key}={mcp_rt.get(key)!r} vs admin.{key}={admin.get(key)!r}"
E   AssertionError: mcp_runtime.override_active=True vs admin.override_active=False
E   assert True == False
```

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Runtime mode | Edge (Rust MCP runtime active) |
| Deployment | docker-compose, multi-replica behind Nginx |

---

### 🧩 Additional Context

The root cause is in the convergence loop (lines 139-147 of `test_runtime_mode.py`):

```python
if mcp_rt.get("effective_mode") == admin.get("effective_mode"):
    break
```

This should check all four asserted keys, not just `effective_mode`. When `boot_mode` already equals the override target (both `edge`), `effective_mode` is identical on both pods regardless of whether the override has propagated, masking the `override_active` divergence.

This is **not a Rust runtime bug** — both Python endpoints compute `override_active` the same way. It's a test convergence gap exposed by multi-pod state propagation timing.

Nayana-R-Gowda:

closes #4440 
                                                                                                                                                                                                                                                  
  ✅ Change 1: Define asserted_keys tuple (line 145)                                                                                                                                                                                                 
  asserted_keys = ("boot_mode", "effective_mode", "override_active", "cluster_propagation")
  - Correct: Defines the complete set of fields that will be asserted                                                                                                                                                                                
  - Benefits: DRY principle - single source of truth                                                                                                                                                                                                 
                                                                                                                                                                                                                                                     
  ✅ Change 2: Enhanced convergence predicate (line 152)                                                                                                                                                                                             
  # Before:                                                                                                                                                                                                                                          
  if mcp_rt.get("effective_mode") == admin.get("effective_mode"):                                                                                                                                                                                    
                                                                                                                                                                                                                                                     
  # After:                                                                                                                                                                                                                                           
  if all(mcp_rt.get(key) == admin.get(key) for key in asserted_keys):                                                                                                                                                                                
  - Correct: Now checks ALL 4 fields for convergence                                                                                                                                                                                                 
  - Fixes the bug: Prevents premature loop exit when only effective_mode matches                                                                                                                                                                     
                                                                                                                                                                                                                                                     
  ✅ Change 3: Reuse asserted_keys in assertion loop (line 155)                                                                                                                                                                                      
  # Before:                                                                                                                                                                                                                                          
  for key in ("boot_mode", "effective_mode", "override_active", "cluster_propagation"):      
 # After:                                                                                                                                                                                                                                           
  for key in asserted_keys:
  - Correct: Ensures consistency between convergence check and assertions                                                                                                                                                                            
  - Maintains correctness: Same fields checked in both places

### DIFF
--- a/tests/live_gateway/protocol_compliance/test_runtime_mode.py
+++ b/tests/live_gateway/protocol_compliance/test_runtime_mode.py
@@ -135,23 +135,24 @@ def test_health_mirrors_runtime_mode_state(gateway_http_client) -> None:
     Multi-pod deployments propagate state via Redis, so a single admin GET
     and a single health GET may land on different pods at different
     propagation points. Poll briefly for convergence before asserting
-    mirror equality — ``effective_mode`` is the leading indicator.
+    mirror equality — all four asserted keys must converge.
     """
     import time as _time
 
     deadline = _time.monotonic() + 3.0
     admin = None
     mcp_rt = None
+    asserted_keys = ("boot_mode", "effective_mode", "override_active", "cluster_propagation")
     while _time.monotonic() < deadline:
         admin = gateway_http_client.get("/admin/runtime/mcp-mode").json()
         health = gateway_http_client.get("/health").json()
         mcp_rt = health.get("mcp_runtime")
         if mcp_rt is None:
             pytest.skip("/health does not expose mcp_runtime block on this deployment")
-        if mcp_rt.get("effective_mode") == admin.get("effective_mode"):
+        if all(mcp_rt.get(key) == admin.get(key) for key in asserted_keys):
             break
         _time.sleep(0.1)
-    for key in ("boot_mode", "effective_mode", "override_active", "cluster_propagation"):
+    for key in asserted_keys:
         assert mcp_rt.get(key) == admin.get(key), f"mcp_runtime.{key}={mcp_rt.get(key)!r} vs admin.{key}={admin.get(key)!r}"
 
 


Yosiefeyob (commented):
LGTM ✅
Verified the convergence-predicate fix behaves as intended.

**Code review:**
- The polling loop now waits until all four asserted keys — boot_mode, effective_mode, override_active, cluster_propagation — agree before asserting, instead of breaking early when only effective_mode matches.
- asserted_keys is reused in both the convergence check and the assertion loop so it works with this compare set

**Test results:**
- Multi-replica run: brought up 3 gateway replicas behind nginx and ran tests/protocol_compliance/test_runtime_mode.py — test_health_mirrors_runtime_mode_state passes.
- Repeat run: ran the same test 10× back-to-back with no failures (loop breaks on first failure):
  ```bash
  for i in {1..10}; do
    python -m pytest tests/protocol_compliance/test_runtime_mode.py::test_health_mirrors_runtime_mode_state -q || break
  done
  ```
- CI: all 4 checks green
- Full local suite: 18,710 passed / 550 skipped — the only errors (test_rate_limiter.py) are  pre-existing and unrelated to this change.
- No regressions detected.

This closes issue #4440: when boot_mode already equals the override target, both pods report the same effective_mode whether or not the override has propagated, which previously let the loop exit before override_active converged. Waiting on all four keys fixes that.

One thing before merge: the branch is a bit behind main could you rebase onto the latest main and resolve any conflicts that come up? Happy to re-approve once that's done. Thanks alot @Nayana-R-Gowda!

Nayana-R-Gowda:
Hi @Yosiefeyob, I have rebased this branch. Could you please take a look?

Yosiefeyob (approved):
LGTM ✅ Re-verified on the rebased branch the conflicts are cleared, CI green. The convergence loop now waits for all four fields to agree before the assertions run. Thanks @Nayana-R-Gowda!

ja8zyjits:
Please Rebase @Nayana-R-Gowda

Nayana-R-Gowda:
@ja8zyjits I rebased the branch.

ja8zyjits (approved):
LGTM

Issue #257, opened 2025-07-06
[FEATURE][SECURITY]: Gateway-level rate limiting, DDoS protection, and abuse detection


crivetimihai:
### 🧭 Epic

**Title:** Gateway-Level Rate Limiting, DDoS Protection & Abuse Detection
**Goal:** Implement a **comprehensive protection framework** in our MCP Gateway to defend against resource exhaustion, distributed attacks, and abusive usage patterns through **intelligent rate limiting**, **adaptive DDoS mitigation**, and **behavioral abuse detection**.
**Why now:** Mitigate potential vulnerabilities to burst traffic and malicious actors. We need battle-tested protection mechanisms that can scale with legitimate usage while blocking bad actors. This builds a reference implementation for **upstream MCP security standards and recommendations**.

---

### 🧭 Type of Feature

* [x] Security hardening
* [x] Performance optimization
* [x] New functionality (experimental)
* [x] Reliability improvement

---

### 🙋‍♂️ User Story 1 — Adaptive Rate Limiting

**As a:** Platform reliability engineer
**I want:** the gateway to enforce per-client rate limits with burst allowances and adaptive thresholds
**So that:** legitimate users get fair resource access while preventing any single client from overwhelming the system.

#### ✅ Acceptance Criteria

```gherkin
Scenario: Enforce per-client rate limits
Given client "app_123" has limit 100 req/min with burst 20
When client makes 25 requests in 10 seconds
Then first 20 succeed immediately
And remaining 5 are rate-limited with 429 "rate_limit_exceeded"
And include "Retry-After" header with backoff time

Scenario: Adaptive threshold adjustment
Given baseline traffic shows 95th percentile at 50 req/min
When system detects consistent 200 req/min from legitimate sources
Then automatically adjust thresholds upward
And log threshold changes for audit
```

---

### 🙋‍♂️ User Story 2 — DDoS Attack Mitigation

**As a:** Security operations engineer
**I want:** automatic detection and mitigation of distributed denial-of-service attacks
**So that:** the gateway remains responsive to legitimate traffic during attack conditions.

#### ✅ Acceptance Criteria

```gherkin
Scenario: Detect volumetric DDoS attack
Given normal traffic baseline of 1000 req/min
When traffic spikes to 10000 req/min from 100+ unique IPs
Then activate DDoS protection mode
And apply progressive backpressure (429 → 503 → connection drops)
And alert security team with attack metrics

Scenario: Geographic anomaly detection
[...]

PR #4607, opened 2026-05-05
fix: bound in-memory rate limit storage to prevent memory exhaustion (CWE-400)

andesyteoss:
## Summary

The in-memory rate limit storage in `mcpgateway/admin.py` is an unbounded `defaultdict(list)` keyed by client IP. Each new client IP that hits a `@rate_limit` endpoint adds a new key, and keys are never removed — even after the timestamps for that IP have all expired. Over time (or under sustained load from many distinct source IPs) the dict grows without bound, eventually exhausting process memory.

This PR bounds the storage to 10,000 entries using an `OrderedDict` with LRU-style eviction, and prunes keys whose timestamps have fully expired.

- **CWE:** CWE-400 (Uncontrolled Resource Consumption)
- **Severity:** Low — see "Honest scoping" below
- **File:** `mcpgateway/admin.py` — `rate_limit_storage` and the `rate_limit()` decorator
- **Data flow:** `request.client.host` (TCP peer address) → used as a dict key in `rate_limit_storage` → key is created on first request and never deleted.

## Fix

```python
# Before
rate_limit_storage = defaultdict(list)

# After
_RATE_LIMIT_MAX_KEYS: int = 10_000
rate_limit_storage: OrderedDict[str, list] = OrderedDict()
```

Inside the decorator:

1. Prune expired timestamps for the current IP into a local list, then write back.
2. `move_to_end(client_ip)` so active IPs are treated as most-recently-used.
3. Evict the oldest entries with `popitem(last=False)` while the dict exceeds `_RATE_LIMIT_MAX_KEYS`.
4. Remove keys whose timestamp lists are empty or whose last timestamp is older than the 60-second window.

The cap of 10,000 was chosen to comfortably accommodate any plausible legitimate client population for an admin API while keeping worst-case memory bounded (~hundreds of KB).

## Tests

- The existing test in `tests/unit/mcpgateway/test_admin.py` was updated to assert `rate_limit_storage` is an `OrderedDict` rather than a `defaultdict`, and to use the new access pattern.
- The decorator's docstring example was updated to match the new type.
- All existing `rate_limit` behavior (limit enforcement, 429 response, per-IP isolation, time-window pruning) is preserved.

## Security analysis

We verified the issue is real but its exploitability is limited:

- **Authenticated only.** Every endpoint that uses `@rate_limit` is also gated by `@require_permission`, so an attacker needs valid credentials before any of their traffic reaches this code path.
- **TCP peer, not header.** The key is `request.client.host` (the TCP source address), not a header-derived value. `X-Forwarded-For` spoofing does **not** create new keys — distinct TCP source IPs are required, which means a real botnet or proxy pool.
- **Admin API is dev-only by policy.** `SECURITY.md` documents that the Admin UI/API should not be exposed in production.

Given those preconditions, growth is slow and gated. But "unbounded dict on a long-running process" is still a real defense-in-depth bug worth fixing — especially since the fix is small, local, and changes no externally observable behavior.

## Honest scoping

Before submitting we tried to disprove this. We checked whether anything upstream (auth, framework limits, or another cleanup path) would already bound the dict — nothing does; entries are added on first use and there is no eviction or TTL anywhere. We also want to be straightforward that the worst-case attacker here is an authenticated user with access to many distinct source IPs against an admin API that your own SECURITY.md says shouldn't be in production. That puts this firmly in low-severity / hardening territory rather than a critical DoS, and we'd rather under-claim than over-claim.

cc @lewiswigmore

### DIFF
--- a/mcpgateway/admin.py
+++ b/mcpgateway/admin.py
@@ -21,7 +21,7 @@ underlying data.
 import asyncio
 import base64
 import binascii
-from collections import defaultdict
+from collections import defaultdict, OrderedDict
 import csv
 from datetime import datetime, timedelta, timezone
 from functools import lru_cache, wraps
@@ -687,8 +687,10 @@ grpc_service_mgr: Optional[Any] = GrpcService() if (settings.mcpgateway_grpc_ena
 
 # Set up basic authentication
 
-# Rate limiting storage
-rate_limit_storage = defaultdict(list)
+# Rate limiting storage — bounded to prevent memory exhaustion (CWE-400).
+# Uses OrderedDict so oldest entries can be evicted when the cap is reached.
+_RATE_LIMIT_MAX_KEYS: int = 10_000
+rate_limit_storage: OrderedDict[str, list] = OrderedDict()
 
 
 @lru_cache(maxsize=1)
@@ -888,8 +890,8 @@ def rate_limit(requests_per_minute: Optional[int] = None):
         Test rate limit storage structure:
         >>> isinstance(admin.rate_limit_storage, dict)
         True
-        >>> from collections import defaultdict
-        >>> isinstance(admin.rate_limit_storage, defaultdict)
+        >>> from collections import OrderedDict
+        >>> isinstance(admin.rate_limit_storage, OrderedDict)
         True
 
         Test decorator with zero limit:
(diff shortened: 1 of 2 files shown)


lewiswigmore:
Closing this as inactive — no maintainer response after 14 days.

The security finding and fix remain valid. If this is still relevant, I'm happy to reopen, rebase, or re-submit against a different branch. Just drop a comment.

lewiswigmore:
Closing this as inactive — no maintainer response after 14 days.

The security finding and fix remain valid. If this is still relevant, I'm happy to reopen, rebase, or re-submit against a different branch. Just drop a comment.

jonpspri:
Relates to #257 ([FEATURE][SECURITY]: Gateway-level rate limiting, DDoS protection, and abuse detection).

This PR fixes an unbounded `defaultdict` in the in-memory rate limit storage (CWE-400 — Uncontrolled Resource Consumption), bounding it to 10,000 entries with LRU eviction. This is a concrete hardening of the rate limiter infrastructure that #257 tracks.

jonpspri:
Hi @andesyteoss — sincere apologies for leaving this PR open for so long without a proper review or merge. That's on us, and we're sorry for the lack of responsiveness.

PR #4610, opened 2026-05-06
feat(oauth): allow operator-defined multi-audience for virtual servers

### LINKED ISSUE #4609: [BUG]:  Virtual server OAuth audience auto-learn cannot bootstrap with Authentik (aud == client_id)
### 🐞 Bug Summary

After v1.0.0 (which includes #4404 / #4410), virtual-server OAuth still fails with `Audience doesn't match` when using Authentik. The auto-learn mechanism never engages on the first request, and a manual DB write is still required.

The fallback path builds `expected_audiences = [canonical_url, oauth_config.client_id]`, but the virtual-server admin form doesn't expose `client_id`. With `client_id` empty, the list collapses to `[canonical_url]`, which never matches an Authentik token whose `aud == client_id`. Verification fails before `_persist_learned_server_audience` runs, so auto-learn cannot bootstrap.

This is a follow-up to #4171, which was closed assuming #4404/#4410 fully resolved the Authentik case ("No UI change needed, no manual DB manipulation"). In practice it doesn't, because there is no path to seed `client_id` for virtual servers.

---

### 🧩 Affected Component

- [x] `mcpgateway` - API
- [ ] Other (explain below)

---

### 🔁 Steps to Reproduce

1. Deploy v1.0.0 with Authentik (or any IdP without RFC 8707 support that mints `aud == client_id`).
2. Register a virtual server via the admin UI with `oauth_enabled = true`. The form has no `client_id` field, so `oauth_config` ends up with neither `client_id` nor `resource` set.
3. Connect a client (e.g. Claude.ai custom connector) and complete the OAuth flow. Authentik issues a valid token with `aud = "<authentik-client-id>"`.
4. Client sends the token → 401 `Audience doesn't match`. Auto-learn never fires.
5. Workaround: `UPDATE servers SET oauth_config = jsonb_set(oauth_config, '{client_id}', '"<authentik-client-id>"') WHERE id = '<vs-id>';` After this, the next request matches via the `client_id` fallback, auto-learn runs, and `resource` is persisted.

---

### 🤔 Expected Behavior
Either approach closes the gap:

- **Expose `client_id` on the virtual-server admin form**, like the gateway form already does. Smallest change; feeds the existing fallback path without touching audience logic. Matches the original suggestion in #4171.
- **Or accept a `list[str]` on `oauth_config.resource`** (RFC 7519 §4.1.3). Lets operators set the accepted audiences explicitly instead of relying on first-write learning. PyJWT already supports list audiences with any-element semantics.

---

### 📓 Logs / Error Output

```
WARN [mcpgateway.utils.verify_credentials] OAuth access token verification failed: Audience doesn't match
```

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Version or commit | `v1.0.0` |
| Runtime | `Python 3.12, uvicorn` |
| Platform / OS | `Linux (UBI 10), AWS EKS, linux/arm64` |
| Container | `Docker` |

---

- Reproduces with any IdP that mints `aud == client_id` and ignores the `resource` parameter from the OAuth authorization request.
- The gateway-side OAuth form already accepts `client_id`. The asymmetry between gateway and virtual-server forms is the root cause.

kimsehwan96:
## 🔗 Related Issue
Closes #4609 

---

## 📝 Summary
Virtual servers with `oauth_enabled=true` cannot complete inbound OAuth verification on IdPs that don't honor RFC 8707 (e.g. Authentik), because such IdPs mint tokens with `aud == client_id`. Auto-learn (#4404, #4410, #4475) is supposed to handle this, but the fallback path it depends on never matches before learning can fire — so the audience can't be bootstrapped, and authentication fails on every request.

This PR adds an explicit operator-defined audience list to the virtual-server OAuth config: `oauth_config.resource` accepts a string or `list[str]`, with a small "+ Add Audience" UI on the server create/edit forms. The primary goal is to unblock the cold-start case, and as a side benefit it also supports multi-client setups where a single virtual server needs to accept tokens from multiple OAuth clients (not the main motivation, but covered naturally).

Auto-learn behavior is unchanged — the existing first-write-only guard means an operator-defined value is preserved, and leaving the field empty falls back to auto-learn.

---

## 🏷️ Type of Change
- [x] Bug fix
- [x] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

| Coverage ≥ 80%            | `make coverage` |   pass  |

---

## ✅ Checklist

---

## 📓 Notes (optional)
Tests added: 7 transport-layer (str / list 2+ / single-list unwrap / list mismatch / empty / malformed / first-write-only no-op), 9 admin form parser (JSON array shapes / bare string / invalid JSON / whitespace), 2 admin API e2e (multi-audience round-trip, single-string backward compatibility).

Backward compatible: existing single-string `oauth_config.resource` values continue to work; no schema migration.

스크린샷 2026-05-06 오후 1 55 35

### DIFF
--- a/mcpgateway/admin.py
+++ b/mcpgateway/admin.py
@@ -2878,6 +2878,7 @@ async def admin_add_server(request: Request, db: Session = Depends(get_db), user
             authorization_server = str(form.get("oauth_authorization_server", "")).strip()
             scopes_str = str(form.get("oauth_scopes", "")).strip()
             token_endpoint = str(form.get("oauth_token_endpoint", "")).strip()
+            oauth_resource_raw = str(form.get("oauth_resource", "")).strip()
 
             if authorization_server:
                 oauth_config = {"authorization_servers": [authorization_server]}
@@ -2886,6 +2887,19 @@ async def admin_add_server(request: Request, db: Session = Depends(get_db), user
                     oauth_config["scopes_supported"] = scopes_str.split()
                 if token_endpoint:
                     oauth_config["token_endpoint"] = token_endpoint
+                if oauth_resource_raw:
+                    try:
+                        parsed_resource = json.loads(oauth_resource_raw)
+                    except (json.JSONDecodeError, ValueError):
+                        parsed_resource = None
+                    if isinstance(parsed_resource, list):
+                        cleaned = [r.strip() for r in parsed_resource if isinstance(r, str) and r.strip()]
+                        if len(cleaned) == 1:
+                            oauth_config["resource"] = cleaned[0]
+                        elif len(cleaned) > 1:
+                            oauth_config["resource"] = cleaned
+                    elif isinstance(parsed_resource, str) and parsed_resource.strip():
+                        oauth_config["resource"] = parsed_resource.strip()
             else:
                 # Invalid or incomplete OAuth configuration; disable OAuth to avoid inconsistent state
                 LOGGER.warning(
@@ -3035,6 +3049,7 @@ async def admin_edit_server(
             authorization_server = str(form.get("oauth_authorization_server", "")).strip()
             scopes_str = str(form.get("oauth_scopes", "")).strip()
             token_endpoint = str(form.get("oauth_token_endpoint", "")).strip()
+            oauth_resource_raw = str(form.get("oauth_resource", "")).strip()
 
             if authorization_server:
                 oauth_config = {"authorization_servers": [authorization_server]}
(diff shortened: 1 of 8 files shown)


vgunjic:
Cross-reference to #4609: I added an [additional ContextForge 1.0.8 reproduction](https://github.com/IBM/mcp-context-forge/issues/4609#issuecomment-5439546285) covering destructive Admin UI updates of API-configured Virtual Server OAuth settings.

This PR correctly addresses the immediate failure by adding the Virtual Server `Resource (Audience)` field, loading `oauth_config.resource` into the edit form, and round-tripping string/list resource values. With that change, an API-configured `resource` should no longer disappear merely because an administrator edits tool associations.

There is still a broader preservation gap in `admin_edit_server()`: it creates a fresh `oauth_config` from the fields represented by the form. Consequently, API-configured `client_id`, provider-specific options, and future OAuth keys that are not represented in the UI can still be silently removed by an unrelated edit.

The 1.0.8 failure sequence was:

1. Configure an OAuth-protected Virtual Server through the API with `authorization_servers`, `scopes_supported`, `resource`, and `client_id`.
2. Verify that a valid access token whose `aud` contains the configured resource authenticates successfully.
3. Edit only the Virtual Server's associated tools in the Admin UI.
4. Observe that `resource` and `client_id` are removed from stored `oauth_config`.
5. Observe the next Streamable HTTP request fail with `Audience doesn't match`, before plugins, policy evaluation, or tool invocation.

Suggested addition to this PR: merge submitted OAuth fields into `existing_server.oauth_config` with explicit clear semantics, rather than replacing the entire object. Please also add a regression test where an unrelated association edit preserves `resource`, API-configured `client_id`, and an unknown/future OAuth key.

Relevant files:

- `mcpgateway/admin.py`
- `mcpgateway/admin_ui/servers.js`
- `mcpgateway/templates/admin.html`
- `mcpgateway/transports/streamablehttp_transport.py`
- `tests/unit/mcpgateway/test_admin.py`

jonpspri:
Hi @kimsehwan96 — sincere apologies for leaving this PR open for so long without a proper review or merge. That's on us, and we're sorry for the lack of responsiveness.

We're closing this PR now as part of a housekeeping pass on long-open contributions. The linked issue (#4609) remains open, so the need is still tracked.

1. Rebase your branch on the current `main`
2. Re-open this PR (or open a fresh one)

PR #4635, opened 2026-05-07
tests: align integration + load tests with current main + cpex contracts

gandhipratik203:
## Summary

Several contract changes on current `main` weren't tracked by the existing rate-limiter tests, so they were silently failing or no-op'ing against the gateway HTTP path. This PR brings them back in line. The changes span three independent surfaces — gateway transport, gateway security defaults, and the cpex framework refactor's mode-label rename.

## What broke and why

| Contract change on `main` | Source | Test failure mode |
|---|---|---|
| `POST /servers/<id>/mcp` now requires an `Mcp-Session-Id` header on every non-initialize call | Gateway streamable-HTTP transport | Tool-path tests returned `HTTP 400 Missing session ID`; load test's `tools/list` auto-detect failed silently → `_tool_names` stayed empty → `@task call_tool` early-returned every tick |
| `GET /admin/plugins` reports framework's internal mode label (`sequential` / `transform` / `disabled`) instead of the operator label set via `PUT` (`enforce` / `permissive` / `disabled`) | cpex framework refactor | One dynamic-behaviour assertion compared against the operator label |
| `.env.example`'s `JWT_SECRET_KEY` lengthened to satisfy the 32-byte minimum | Gateway security hardening | Two locustfiles still defaulted to the old 11-char `"my-test-key"` → admin requests 401'd |

## Changes

- **`tests/integration/test_rate_limiter_multi_tenant.py`** — add `_mcp_initialize_session` helper; `_invoke_tool_once` now does the initialize + initialized handshake and sends `Mcp-Session-Id` on the tool call.
- **`tests/integration/test_rate_limiter_dynamic_behavior.py`** — same handshake helper added to `_send_tool_burst`; admin-API mode assertion relaxed to "reported mode is non-disabled" with a comment explaining the operator-vs-internal label mapping.
- **`tests/loadtest/locustfile_rate_limiter_backend_correctness.py`** — `_auto_detect` runs the MCP initialize handshake before `tools/list` so `_tool_names` actually populates.
- **`tests/loadtest/locustfile_rate_limiter_redis_capacity.py`** + **`locustfile_rate_limiter_scale.py`** — `JWT_SECRET_KEY` default updated to match `.env.example`.

## Verification

| Run | Result |
|---|---|
| `pytest tests/integration/test_rate_limiter_multi_tenant.py --with-integration` | ✅ 2/2 |
| `pytest tests/integration/test_rate_limiter_dynamic_behavior.py --with-integration` | ✅ 5/5 |
| `make benchmark-rate-limiter` | ✅ tools-path enforcement under sustained load — Redis backend verdict |
| `make benchmark-rate-limiter-redis-capacity` | ✅ prompt-path sustains concurrent load with 0 errors |

Run against a docker-compose stack with the rate limiter enabled (`mode: enforce` in `plugins/config.yaml`) and `REDIS_CONTAINER_NAME` set to match the running compose project name. The `mode: enforce` flip is operator-side test setup, not a chart change — `plugins/config.yaml` is unchanged in this PR.

## Out of scope

- `tests/integration/test_rate_limiter.py` — 50/59 substantive tests still pass. The 9 failures are scaffolding bit-rot (renamed `/api/v1/...` routes, removed `_store` introspection attribute the Rust core no longer exposes). The plugin engine itself is well-covered by the passing tests; happy to address the bit-rot in a follow-up if it's worth a separate cleanup.
- The dynamic plugin-bindings path (`POST /v1/tools/plugin_bindings`) — separate observation that for `RateLimiterPlugin` specifically, binding `config` overrides aren't honored at runtime even though the same path works for `OutputLengthGuardPlugin` and `SecretsDetection`. Diagnosing that is its own thread; this PR keeps to the static-config test surface.

### DIFF
--- a/tests/helpers/mcp_session.py
+++ b/tests/helpers/mcp_session.py
@@ -1,0 +1,123 @@
+# -*- coding: utf-8 -*-
+"""Location: ./tests/helpers/mcp_session.py
+Copyright 2026
+SPDX-License-Identifier: Apache-2.0
+
+Shared MCP session-handshake helper for integration and load tests.
+
+Centralizes the streamable-HTTP ``POST /servers/<id>/mcp`` initialize +
+``notifications/initialized`` handshake that the gateway requires before any
+non-initialize call. Replaces the near-identical helpers that lived in
+``tests/integration/test_rate_limiter_dynamic_behavior.py`` and
+``tests/integration/test_rate_limiter_multi_tenant.py``.
+
+For ``httpx``-based callers (e.g. ``tests/live_gateway/mcp/...`` and
+``tests/live_gateway/e2e_rust/...``) see the helpers in those files. This
+module uses ``requests`` to match the integration-test convention.
+"""
+
+# Standard
+import logging
+
+# Third-Party
+import requests
+
+logger = logging.getLogger(__name__)
+
+# Canonical MCP protocol version for the gateway's streamable-HTTP transport.
+# Mirrors the constant in tests/live_gateway/mcp/test_mcp_plugin_parity.py
+# and tests/loadtest/locustfile_echo_delay.py — keep these in sync if the
+# gateway's supported version changes.
+MCP_PROTOCOL_VERSION = "2025-11-25"
+
+
+def initialize_mcp_session(
+    gateway_url: str,
+    server_id: str,
+    headers: dict,
+    *,
+    client_name: str = "test-client",
+    client_version: str = "0",
+    protocol_version: str = MCP_PROTOCOL_VERSION,
+    init_timeout: float = 10.0,
+    notify_timeout: float = 5.0,
+) -> str | None:
+    """Run the MCP streamable-HTTP initialize + initialized handshake.
+
+    The gateway's session-aware MCP transport requires an ``Mcp-Session-Id``
+    header on every ``POST /servers/<id>/mcp`` non-initialize call. This
+    helper performs the handshake and returns the allocated session id, or
+    ``None`` on failure (with a ``logger.warning`` describing which step
+    failed so the caller's test output explains the failure mode).
+
+    Args:
+        gateway_url: Base URL of the gateway (e.g. ``http://localhost:8080``).
+        server_id: Target virtual server id.
+        headers: Pre-built auth / content headers (typically a Bearer-token
+            header dict from the caller's ``_fresh_headers()``).
+        client_name: ``clientInfo.name`` reported in the initialize body.
+            Pass a test-specific name so it shows up in server logs.
+        client_version: ``clientInfo.version`` reported in the initialize body.
+        protocol_version: MCP protocol version. Defaults to the canonical
+            ``MCP_PROTOCOL_VERSION``.
+        init_timeout: Seconds to wait for the initialize POST.
+        notify_timeout: Seconds to wait for the notifications/initialized POST.
+
+    Returns:
+        The allocated session id when the handshake succeeds, otherwise
+        ``None``. A non-``None`` return guarantees the initialize step
+        succeeded; the notifications/initialized step is best-effort and
+        a failure there is logged but does not invalidate the session id.
+    """
+    sse_headers = {**headers, "Accept": "application/json, text/event-stream"}
+    init_body = {
+        "jsonrpc": "2.0",
+        "id": "init",
+        "method": "initialize",
+        "params": {
+            "protocolVersion": protocol_version,
+            "capabilities": {},
+            "clientInfo": {"name": client_name, "version": client_version},
+        },
+    }
+    try:
+        resp = requests.post(
+            f"{gateway_url}/servers/{server_id}/mcp",
+            json=init_body,
+            headers=sse_headers,
+            timeout=init_timeout,
+        )
+    except requests.RequestException as exc:
+        logger.warning(
+            "MCP initialize POST failed for server %s (transport): %s",
+            server_id, exc,
+        )
+        return None
+    if resp.status_code != 200:
+        logger.warning(
+            "MCP initialize returned HTTP %s for server %s: %s",
+            resp.status_code, server_id, resp.text[:200],
+        )
+        return None
+    sid = resp.headers.get("mcp-session-id")
+    if not sid:
+        logger.warning(
+            "MCP initialize response for server %s missing mcp-session-id header",
+            server_id,
+        )
+        return None
+    try:
+        requests.post(
+            f"{gateway_url}/servers/{server_id}/mcp",
+            json={"jsonrpc": "2.0", "method": "notifications/initialized"},
+            headers={**sse_headers, "Mcp-Session-Id": sid},
+            timeout=notify_timeout,
+        )
+    except requests.RequestException as exc:
+        # Best-effort: the session id is already valid for non-initialize
+        # calls. Log so the caller can see the failure mode in test output.
+        logger.warning(
+            "MCP notifications/initialized failed for server %s session %s: %s",
+            server_id, sid, exc,
+        )
+    return sid
--- a/tests/integration/test_rate_limiter.py
+++ b/tests/integration/test_rate_limiter.py
@@ -380,43 +380,6 @@ class TestToolPreInvoke:
         assert "Retry-After" not in result.http_headers  # Not on success
 
 
-class TestStoreCleanup:
-    """Tests for rate limit store cleanup."""
-
-    @pytest.mark.asyncio
-    async def test_store_cleanup_between_tests(self, rate_limit_plugin_2_per_second):
-        """Verify each plugin instance starts with an empty store."""
-        plugin = rate_limit_plugin_2_per_second
-        backend = plugin._rate_backend
-
-        # Fresh instance — store must be empty before any requests
-        assert len(backend._algorithm._store) == 0
-
-        ctx = PluginContext(global_context=GlobalContext(request_id="r1", user="alice"))
-        payload = PromptPrehookPayload(prompt_id="test", args={})
-
-        await plugin.prompt_pre_fetch(payload, ctx)
-
-        # After one request a window entry must exist
-        assert len(backend._algorithm._store) > 0
-
-    @pytest.mark.asyncio
-    async def test_multiple_users_create_separate_windows(self, rate_limit_plugin_2_per_second):
-        """Verify multiple users create separate window entries in the backend store."""
-        plugin = rate_limit_plugin_2_per_second
-        backend = plugin._rate_backend
-
-        ctx_alice = PluginContext(global_context=GlobalContext(request_id="r1", user="alice"))
-        ctx_bob = PluginContext(global_context=GlobalContext(request_id="r2", user="bob"))
-        payload = PromptPrehookPayload(prompt_id="test", args={})
-
-        await plugin.prompt_pre_fetch(payload, ctx_alice)
-        await plugin.prompt_pre_fetch(payload, ctx_bob)
-
-        # Each user must have their own key in the store
-        assert len(backend._algorithm._store) >= 2
-
-
 class TestSlidingWindowIntegration:
     """End-to-end integration tests for the sliding_window algorithm."""
 
@@ -717,22 +680,6 @@ class TestPermissiveMode:
         assert result.violation.http_status_code == 429
 
     @pytest.mark.asyncio
-    async def test_permissive_mode_still_tracks_counters(self):
-        """Permissive mode still decrements the counter — backend store must grow."""
-        plugin, hook_ref, executor = self._make_plugin_and_hook("10/s")
-        ctx = PluginContext(global_context=GlobalContext(request_id="r1", user="alice"))
-        payload = ToolPreInvokePayload(name="tool", arguments={})
-
-        await executor.execute_plugin(hook_ref, payload, ctx, violations_as_exceptions=True)
-        await executor.execute_plugin(hook_ref, payload, ctx, violations_as_exceptions=True)
-
-        # Counter must have been incremented — key exists in backend store
-        store = plugin._rate_backend._algorithm._store
-        assert len(store) > 0, "Permissive mode must still track counters in the backend store"
-        key = next(iter(store))
-        assert store[key].count == 2, f"Expected count=2 after 2 requests, got {store[key].count}"
-
-    @pytest.mark.asyncio
     async def test_permissive_mode_contrast_with_enforce(self):
         """Enforce mode raises PluginViolationError; permissive mode does not."""
         enforce_config = PluginConfig(
@@ -785,18 +732,6 @@ class TestDisabledMode:
         for _ in range(10):
             result, _ = await executor.execute(hook_refs, payload, global_ctx, "tool_pre_invoke", violations_as_exceptions=True)
             assert result.violation is None, "Disabled plugin must never produce a violation"
-
-    @pytest.mark.asyncio
-    async def test_disabled_mode_does_not_track_counters(self):
-        """execute() skips the plugin — backend store must remain empty."""
-        plugin, hook_refs, executor = self._make_plugin_and_refs()
-        global_ctx = GlobalContext(request_id="r1", user="alice")
-        payload = ToolPreInvokePayload(name="tool", arguments={})
-
-        for _ in range(5):
-            await executor.execute(hook_refs, payload, global_ctx, "tool_pre_invoke", violations_as_exceptions=True)
-
-        assert len(plugin._rate_backend._algorithm._store) == 0, "Disabled plugin must not write to the backend store — executor skips it entirely"
 
     def test_disabled_plugin_mode_property(self):
         """Plugin mode property reflects the configured disabled mode."""
--- a/tests/integration/test_rate_limiter.py
+++ b/tests/integration/test_rate_limiter.py
@@ -46,8 +46,7 @@ from cpex.framework.models import PluginMode
 from cpex_rate_limiter.rate_limiter import RateLimiterPlugin
 
 # API Endpoints
-PROMPT_ENDPOINT = "/api/v1/prompts/"
-TOOL_INVOKE_ENDPOINT = "/api/v1/tools/invoke"
+PROMPT_ENDPOINT = "/prompts/"
 
 
 @pytest.fixture
@@ -964,10 +963,6 @@ class TestNoLimitsAndMissingContext:
         result = await plugin.tool_pre_invoke(payload, ctx)
         assert result.violation is not None, "user=None must be treated as 'anonymous' and enforced"
 
-        # Confirm the key in the store is 'user:anonymous'
-        store = plugin._rate_backend._algorithm._store
-        assert any("anonymous" in k for k in store), "Expected 'anonymous' bucket key in store when user=None"
-
     @pytest.mark.asyncio
     async def test_none_tenant_id_skips_by_tenant_check(self):
         """tenant_id=None in GlobalContext must skip the by_tenant check entirely — no 'default' bucket."""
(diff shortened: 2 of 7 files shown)


msureshkumar88 (changes requested):
Thanks for the clear problem statement and the summary table — the three contract breaks are well-diagnosed and the fix direction is right. A few issues need to be addressed before this lands.

## Blocking

### R1 — `_mcp_initialize_session` copy-pasted verbatim into two files

`tests/integration/test_rate_limiter_dynamic_behavior.py` and `tests/integration/test_rate_limiter_multi_tenant.py` contain near-identical implementations of `_mcp_initialize_session`. An equivalent `_initialize_session` already exists in `tests/live_gateway/mcp/test_mcp_plugin_parity.py:170`. Please extract this to `tests/integration/helpers/` (or `tests/integration/conftest.py`) and import from both files — the next integration test that needs the handshake will copy-paste again otherwise.

### R2 — Wrong protocol version `"2025-06-18"` — inconsistent with codebase canonical

The canonical protocol version across this repo is `"2025-11-25"` (set as `MCP_PROTOCOL_VERSION` in `tests/live_gateway/mcp/test_mcp_plugin_parity.py:29` and `tests/loadtest/locustfile_echo_delay.py:108`). Both new integration helpers use `"2025-06-18"`. Worse, `locustfile_rate_limiter_backend_correctness.py` now has two different versions in the same file — `"2025-06-18"` in the new `_auto_detect` block and `"2024-11-05"` in the existing `call_tool` task. Please use a shared `MCP_PROTOCOL_VERSION = "2025-11-25"` constant, import it everywhere, and explain in a comment if `"2025-06-18"` is intentional (e.g. backward-compat testing).

### R3 — Mode assertion relaxed too far

```python
# Before (correct contract, wrong label):
assert plugins[PLUGIN_NAME]["mode"] == "enforce"

# After (passes for any non-disabled state — too permissive):
assert reported_mode != "disabled"
```

The right fix is to make the operator→internal label mapping explicit:

```python
OPERATOR_TO_INTERNAL_MODE = {"enforce": "sequential", "permissive": "transform", "disabled": "disabled"}
assert reported_mode == OPERATOR_TO_INTERNAL_MODE["enforce"], (
    f"Expected internal mode {OPERATOR_TO_INTERNAL_MODE['enforce']!r}, got {reported_mode!r}"
)
```

This gives a precise assertion, documents the mapping where it's used, and fails correctly if the framework relabels modes again.

### P1 — `_invoke_tool_once` creates a new MCP session per call — semantic error for rate-limit tests

`_invoke_tool_once` runs the full initialize + initialized handshake before every single tool invocation. If the rate limiter keys on session ID or resets counters per session, each call in `test_tool_invocation_creates_rate_limit_keys_in_redis` and `test_rate_limit_keys_carry_tenant_prefix_when_tool_is_team_owned` operates in a fresh window rather than accumulating against the same client. The session should be established once in the `server_and_tool` fixture and passed in, not re-created per call.

### S1 — Hardcoded fallback JWT secret still in source

```python
JWT_SECRET_KEY = _cfg("JWT_SECRET_KEY", "my-test-key-but-now-longer-than-32-bytes")
```

The old key was wrong for a different reason; the fix still bakes a known secret into source. For load tests run against real infra a missing env var silently uses this string, and any JWT signed with it is trivially forgeable. Consider:

```python
_raw = _cfg("JWT_SECRET_KEY", "")
if not _raw:
    raise RuntimeError("JWT_SECRET_KEY env var is required for load tests")
JWT_SECRET_KEY = _raw
```

Or at minimum add a clear `# nosec` annotation and a comment explaining this is a CI-only fallback that must never reach production.

### L1/L2 — Silent error swallowing in both helpers

Both `_mcp_initialize_session` implementations:

```python
except requests.RequestException:
    pass  # ← no log
return sid  # returned as valid even if notification step failed
```

The `notifications/initialized` failure is swallowed silently. A failed notification with a returned `sid` means subsequent calls may fail with opaque errors. Add at minimum:

```python
except requests.RequestException as exc:
    logger.warning("MCP initialized notification failed for server %s: %s", server_id, exc)
```

Also: all four `return None` paths in both helpers produce no log output. The caller then reports `status=0` or `errors=count` with nothing in the output to explain why. Add a `logger.warning` before each `return None`.

---

## Non-blocking suggestions

**R4 — `return 0` is a confusing error sentinel**

`_invoke_tool_once` returns `0` when `sid is None`. The caller asserts `status == 200`, giving `AssertionError: assert 0 == 200` with no explanation. Use `pytest.fail("MCP session handshake failed")` or raise a descriptive exception.

**R5 — `notifications/initialized` missing in `_auto_detect`**

The locustfile `_auto_detect` gets the `sid` from the initialize response but never sends the `notifications/initialized` POST. The integration tests do send it. Worth aligning for protocol correctness.

**T1 — 9 failing tests in `test_rate_limiter.py`**

Acknowledged as out of scope — worth opening a follow-up issue and linking it here so it doesn't get lost.

**T2 — No negative-path coverage for session handshake failure**

No test exercises `_mcp_initialize_session` returning `None` — what `_send_tool_burst` reports in that case, and whether the test suite distinguishes "rate limited" from "couldn't connect at all".

**T3 — No cross-tenant session replay test**

After adding session handshake, worth adding a test that a session ID from tenant A cannot be used on tenant B's endpoint (isolation regression).

**D1 — `tests/AGENTS.md` not updated**

The session handshake is now a required protocol step for any integration test on the MCP tool path. Worth documenting there.

gandhipratik203:
Thanks for the thorough review — really helped tighten this up. Here's what landed in response, by ID:

| ID | Resolution |
|---|---|
| **R1** | Extracted `_mcp_initialize_session` into a shared `tests/helpers/mcp_session.py`. Both `test_rate_limiter_dynamic_behavior.py` and `test_rate_limiter_multi_tenant.py` now import it; ~80 lines of near-duplicate gone. (`c5da7d536`) |
| **R2** | Defined `MCP_PROTOCOL_VERSION = "2025-11-25"` in the new helper (matches `tests/live_gateway/mcp/test_mcp_plugin_parity.py:29` and `tests/loadtest/locustfile_echo_delay.py:108`). Replaced the `"2025-06-18"` literals in both helpers and `locustfile_rate_limiter_backend_correctness.py`, plus the inconsistent `"2024-11-05"` still in the locustfile's `_ensure_initialized`. The locustfile defines its own constant inline (with a comment) so it stays runnable as a standalone script without PYTHONPATH gymnastics. (`c5da7d536`) |
| **R3** | Tightened `test_mode_visible_in_admin_api_after_change` from `!= "disabled"` to `== "sequential"` (the framework-internal label after `enforce → sequential` migration). TEMP comment in-place pointing at #4709, which tracks the operator-vs-framework label asymmetry across `/admin/plugins` and `/v1/tools/plugin_bindings/`. (`b713fc241`) |
| **R5** | `_auto_detect` in `locustfile_rate_limiter_backend_correctness.py` now POSTs `notifications/initialized` after `initialize`, matching the integration tests' protocol behaviour. (`c5da7d536`) |
| **L1/L2** | Every early-return path in `initialize_mcp_session` now logs a `logger.warning` with context (server_id, status code, response excerpt, exception). The `notifications/initialized` failure is logged rather than silently `pass`-swallowed. (`c5da7d536`) |
| **R4** | `_invoke_tool_once`'s `return 0` sentinel replaced with `pytest.fail("MCP session handshake failed for server <id> — see logger output for the failing step")`. Combined with L1/L2's logging, the failure mode is now self-explanatory in test output. (`e4a5aa689`) |
| **S1** | `JWT_SECRET_KEY` fallback removed from `locustfile_rate_limiter_{backend_correctness,redis_capacity,scale}.py`. Each now `raise RuntimeError(...)` if the env var (or `.env` entry) is unset, with a message explaining the security reason. The same pattern exists in 6 other loadfiles outside this PR's scope — flagged in the commit body as a natural follow-up. (`7fb3a5ff8`) |
| **T1** | All 9 originally-failing tests in `test_rate_limiter.py` resolved in-place rather than as a follow-up issue: 4 removed (pure `_store` introspection no longer expressible against the Rust port), 4 fixed/trimmed (1 URL renamed `/api/v1/prompts/` → `/prompts/`, 2 `_store` peeks dropped while keeping the observable assertions, 1 obsolete `/api/v1/tools/invoke` removed since tool invoke moved to MCP transport), and 1 pinned with a TEMP comment for #4710 (cpex TRANSFORM mode silently discarding `result.violation` — observability gap). File now: 54 pass / 0 fail. (`2a1b73839` + `ab8edfafb` + `15bb6a2a8`) |

Two cpex-side observations from the cleanup got their own tracking issues since they're separate from this PR's scope — #4709 (mode label asymmetry) and #4710 (TRANSFORM violation discard).

gandhipratik203:
P1 — parked. The multi_tenant tests are flaky on long-running stacks (passing 2/2 right after a gateway restart, degrading to 1/2 or 0/2 once other rate-limiter tests have run against the same containers). Filed #4746 to track both the flakiness investigation and the P1 session refactor — they're easier to apply together once the root cause is understood. Investigation notes and a suggested approach are in the issue body.

msureshkumar88 (changes requested):
The three contract breaks described in the PR summary are all addressed, and the shared \`tests/helpers/mcp_session.py\` helper is a clean extraction. The \`pytest.fail()\` approach in \`_invoke_tool_once\` (multi-tenant file) is the right pattern for surfacing handshake failures — more on that below.

Two blocking issues in \`locustfile_rate_limiter_scale.py\` need fixes before this merges, and one test-quality gap worth closing while we're here.

### Blocking

**\`_bootstrap_users\` — missing MCP initialize handshake (same bug this PR fixes elsewhere)**

\`_bootstrap_users\` calls \`tools/list\` directly via \`POST /servers/{sid}/mcp\` without first running the initialize handshake. That POST will get \`HTTP 400 Missing session ID\`, \`tools\` will be empty, \`_tool_names\` stays \`[]\`, and every \`call_tool()\` tick silently no-ops via the early return at line 940. The scale load test will appear to run cleanly but measure nothing. The same pattern was correctly fixed in \`_auto_detect\` in \`locustfile_rate_limiter_backend_correctness.py\` — worth applying the same treatment here.

**\`_ensure_initialized\` — stale protocol version**

\`_ensure_initialized\` (line 920) still sends \`protocolVersion: "2024-11-05"\` while the rest of this PR aligns to \`"2025-11-25"\`. If the gateway rejects the old version string (JSON-RPC error response on an otherwise HTTP 200), \`result\` comes back \`None\`, \`self._initialized\` stays \`False\`, and subsequent \`call_tool\` requests go out without \`Mcp-Session-Id\`. The \`MCP_PROTOCOL_VERSION\` constant is already defined at module top — worth wiring it up here.

### Suggestion (worth fixing in this PR)

**\`_send_tool_burst\` — silent false positive when handshake fails**

When \`initialize_mcp_session\` returns \`None\`, \`_send_tool_burst\` returns \`{allowed: 0, rate_limited: 0, errors: BURST_SIZE, total: BURST_SIZE}\`. \`test_burst_all_allowed_when_disabled\` then evaluates \`assert 0 == 0\` and \`assert 0 == BURST_SIZE - BURST_SIZE\` — both pass, so a total infrastructure failure looks like a green test. \`_invoke_tool_once\` in the multi-tenant file handles this correctly with \`pytest.fail()\`. Worth applying the same pattern here.

### Minor note

**\`_send_tool_burst\` — two independent \`_fresh_headers()\` calls**

\`initialize_mcp_session\` is called with one \`_fresh_headers()\` invocation (token A), then \`headers\` is built from a second call (token B). Both hit \`POST /auth/login\` independently. Low risk if the gateway doesn't bind sessions to the originating token, but cleaner to reuse the same token for both — easy fix while touching the function.

gandhipratik203:
Round 2 — all items addressed.

| ID | Resolution |
|---|---|
| **R3** | `_send_tool_burst` now calls `pytest.fail()` when `initialize_mcp_session` returns `None`, mirroring the multi-tenant file's pattern. Total handshake failure no longer looks like a green test. (`0c956be09`) |
| **R4** | Single `_fresh_headers()` call reused for the handshake + every burst tick. One login round-trip per `_send_tool_burst` invocation instead of two. (`0c956be09`) |
| **B1** | `_bootstrap_users` now runs the MCP `initialize` + `notifications/initialized` handshake before `tools/list`, threading `Mcp-Session-Id` into the lookup. Pattern lifted from `_auto_detect` in `locustfile_rate_limiter_backend_correctness.py`. Without this fix, `_tool_names` would stay empty and `call_tool` would silently no-op every tick. (`82803e429`) |
| **B2** | `MCP_PROTOCOL_VERSION = "2025-11-25"` defined at module top; both the new bootstrap handshake and `_ensure_initialized` use the constant instead of literals. (`82803e429`) |

Verified: `test_burst_all_allowed_when_disabled` passes against a live docker-compose stack in 24s. Locustfile changes pass syntax / lint; full load-test verification (`locust --headless`) is the next step for anyone exercising the scale profile.

msureshkumar88 (approved):
All blocking items from both rounds of review are resolved. Approving.

## What's fixed

**From review 1:**
- R1 ✅ — Shared `tests/helpers/mcp_session.py` cleanly extracts the duplicate handshake helpers.
- R2 ✅ — `MCP_PROTOCOL_VERSION = "2025-11-25"` is now a single constant used consistently; the old `"2025-06-18"` and `"2024-11-05"` versions are gone.
- R3 ✅ — Mode assertion now checks `reported_mode == "sequential"` (explicit operator→internal mapping) rather than the too-permissive `!= "disabled"`. The `TEMP / issue #4709` comment documents the expected revert path cleanly.
- S1 ✅ — All three locustfiles now raise `RuntimeError` when `JWT_SECRET_KEY` is unset; no fallback secret in source.
- L1/L2 ✅ — Every `return None` path in `initialize_mcp_session` is preceded by a `logger.warning`; the `notifications/initialized` exception is logged rather than swallowed.
- R4 ✅ — `_invoke_tool_once` uses `pytest.fail()` instead of the `return 0` sentinel; the `AssertionError: assert 0 == 200` confusion is gone.
- R5 ✅ — `_auto_detect` in the backend-correctness locustfile now sends `notifications/initialized` before `tools/list`.

**From review 2:**
- `_bootstrap_users` ✅ — Full initialize + `notifications/initialized` handshake before `tools/list`; `continue` on missing `mcp-session-id` so the loop doesn't fall through to a guaranteed-400 `tools/list` call.
- `_ensure_initialized` ✅ — Uses `MCP_PROTOCOL_VERSION` constant; stale `"2024-11-05"` gone from the scale locustfile.
- `_send_tool_burst` false positive ✅ — `pytest.fail()` on `sid is None`; infrastructure failure no longer looks like a green test.
- `_send_tool_burst` two tokens ✅ — Single `base_headers` is passed to both `initialize_mcp_session` and the burst loop.

## Remaining minor items (non-blocking, carry forward)

- `_invoke_tool_once` in `test_rate_limiter_multi_tenant.py` still calls `_fresh_headers()` twice (once for the handshake, once to build `headers`). Low risk since the gateway doesn't bind sessions to the originating token, but worth a one-liner tidy when next touching that function.
- P1 (per-call session in `_invoke_tool_once`) — the multi-tenant tests care about cross-tenant isolation rather than window accumulation, so a per-invocation session is acceptable for their purpose.
- T1: the 9 failing tests in `test_rate_limiter.py` — worth a follow-up issue to track.
- D1: `tests/AGENTS.md` could document the session handshake as a required step for any new MCP tool-path integration test.

ja8zyjits:
### 🟡 Minor Issues (Non-Blocking)

#### M1: Double Token Fetch in `_invoke_tool_once`
**File:** `tests/integration/test_rate_limiter_multi_tenant.py:103-129`

```python
def _invoke_tool_once(server_id: str, tool_name: str) -> int:
    sid = initialize_mcp_session(
        GATEWAY_URL, server_id, _fresh_headers(),  # ← First call
        client_name="rate-limiter-multi-tenant-test",
    )
    # ...
    headers = {
        **_fresh_headers(),  # ← Second call
        "Accept": "application/json, text/event-stream",
        "Mcp-Session-Id": sid,
    }
```

**Issue:** Two independent `POST /auth/login` round-trips per invocation.

**Fix:** Reuse token from handshake (matches `_send_tool_burst` pattern in dynamic_behavior file):

```python
base_headers = _fresh_headers()
sid = initialize_mcp_session(GATEWAY_URL, server_id, base_headers, ...)
headers = {**base_headers, "Accept": "...", "Mcp-Session-Id": sid}
```

**Impact:** Low (gateway doesn't bind sessions to tokens), but cleaner and faster.

#### M2: Mode Assertion Relaxation
**File:** `tests/integration/test_rate_limiter_dynamic_behavior.py:313-328`

```python
expected_framework_mode = "sequential"
assert reported_mode == expected_framework_mode, (
    f"After PUT mode=enforce, /admin/plugins should report "
    f"{expected_framework_mode!r} (framework label, see issue #4709); "
    f"got {reported_mode!r}"
)
```

**Observation:** Assertion now checks `== "sequential"` (correct), but the TEMP comment suggests this will revert to `== "enforce"` once issue #4709 resolves.

**Suggestion:** Add a TODO marker so the revert isn't forgotten:

```python
# TODO(#4709): Revert to `assert reported_mode == "enforce"` once
# /admin/plugins reports operator labels instead of framework labels.
```

---

### 📋 Test Coverage Gaps (Carry Forward)

#### T2: No Negative-Path Coverage for Session Handshake Failure
**Missing:** Test that exercises `initialize_mcp_session` returning `None` and verifies the caller's error handling.

**Suggestion:** Add a test that mocks `requests.post` to return 500, confirms `None` is returned, and checks that `_send_tool_burst` calls `pytest.fail()`.

#### T3: No Cross-Tenant Session Replay Test
**Missing:** Test that a session ID from tenant A cannot be used on tenant B's endpoint.

**Suggestion:** Add integration test:
1. Create session for tenant A's server
2. Attempt tool call on tenant B's server with tenant A's session ID
3. Assert 403 or 400 (isolation violation)

#### D1: `tests/AGENTS.md` Not Updated
**Missing:** Documentation that session handshake is now required for any MCP tool-path integration test.

**Suggestion:** Add section to `tests/AGENTS.md`:


### MCP Session Handshake

All integration tests on the MCP tool path must perform the session handshake:

```python
from tests.helpers.mcp_session import initialize_mcp_session

sid = initialize_mcp_session(gateway_url, server_id, headers)
if sid is None:
    pytest.fail("MCP session handshake failed — see logger output")
```

Use the returned `sid` in the `Mcp-Session-Id` header for all subsequent calls.


---

## Recommendations

### Must Do (Before Merge)
None. All blocking issues resolved.

### Should Do (This PR)
1. **M1:** Fix double token fetch in `_invoke_tool_once` (1-line change)
2. **M2:** Add TODO marker for mode assertion revert

### Could Do (Follow-up)
1. **T2:** Add negative-path test for session handshake failure
2. **T3:** Add cross-tenant session isolation test
3. **D1:** Update `tests/AGENTS.md` with session handshake requirement
4. **#4746:** Investigate multi-tenant test flakiness on long-running stacks

---


**Minor issues (M1, M2) are optimizations, not blockers.**

---


**Review focus:**
1. Contract break resolution (session handshake, mode labels, JWT length)
2. Security implications (hardcoded secrets, token handling)
3. Test coverage (observable behavior vs internal state)
4. Code quality (duplication, error handling, maintainability)
5. Edge cases (failure modes, cross-tenant isolation)

gandhipratik203:
Thanks for the careful read, @ja8zyjits. All five items here were also raised in @msureshkumar88's review rounds on this PR and dispositioned before approval — with two of them already having tracking issues filed:

- **M1** — double `_fresh_headers()` in `_invoke_tool_once`. Suresh flagged the same function in his approval review (2026-05-14) as a non-blocking carry-forward: *"worth a one-liner tidy when next touching that function."* The broader concern about the per-call session in this same function (his P1 from review 1) is tracked in #4746, which also covers a related multi-tenant test flakiness investigation. Both refactors are easier to apply together once the flakiness root cause is understood, so M1 naturally folds into that work.
- **M2** — TODO marker for the mode-assertion revert. Already in place: the `TEMP / see issue #4709` comment on the assertion was added in response to Suresh's R3, and #4709 itself is the revert trigger you'd want a TODO to point at (it tracks the `/admin/plugins` operator-vs-framework label asymmetry).
- **T2, T3, D1** — match Suresh's review 1 items by the same letter codes; all three were explicitly accepted as non-blocking carry-forwards in his approval. T2/T3 fit naturally as design constraints for the fast happy-path integration test tracked on #4665 (planned scope item 2). D1 is a docs-housekeeping follow-up worth doing the next time AGENTS.md gets touched.

Net: nothing to action in this PR beyond what was already approved. Appreciate the second pair of eyes on the same surface — independent convergence on the same findings is a useful signal, and it confirms #4746 / #4709 are pointing at the right things.

ja8zyjits (approved):
LGTM

PR #4444, opened 2026-04-26
fix(bootstrap): improve startup reliability for multi-replica deploys

gandhipratik203:
## Summary

Fixes issue #4051 — Alembic migration advisory lock hangs when multiple gateway replicas start through PgBouncer in transaction-pooling mode. The session-scoped `pg_advisory_lock` is orphaned across PgBouncer's backend handoffs; the N-th gateway pod sees the lock held by an effectively-dead session and spins in its retry loop until the ~10-minute timeout fires.

This PR closes the bug along three pieces and ships them together so the chart is safe under both deployment patterns:

- **Layer 1 (correctness)** — `bootstrap_db.main()` now probes `alembic_version` *before* attempting the advisory lock. When the schema is at the script directory's head, the lock acquisition is skipped entirely. Replicas 2..N never participate in the race.
- **Layer 2 (operator ergonomics)** — `MCPGATEWAY_SKIP_MIGRATIONS=true` tells `bootstrap_db.main()` to skip the Alembic upgrade and the advisory lock, while still running the idempotent post-migration helpers (admin user, default RBAC roles, orphaned-resource assignment) so single-pod first-startup contracts still hold. The path **fails fast** with `RuntimeError("Schema not at head; migrations required before startup")` if the flag is set but the schema isn't actually at head — operators see the deployment error immediately rather than serving traffic against a half-migrated DB. The `mcp-stack` Helm chart wires the flag from `migration.enabled` so the contract "the pre-install Job owns migrations, app pods skip" is enforced at the chart layer rather than left to operators to coordinate.
- **Schema-at-head startup probe (chart-level Ready gate)** — gateway pod's startup probe now runs `python -m mcpgateway.utils.check_schema_at_head` instead of `db_isready`. Pods refuse Ready until `alembic_version` matches the script-directory head, regardless of which entity (pre-install Job, post-install Job, init container, external CD pipeline) ran the migration. Defense in depth: even if L2 mis-fires, no pod ever serves traffic against a half-migrated DB.

Library defaults preserve backward compatibility: a direct `docker run mcpgateway:latest` continues to bootstrap its own schema. Only the chart and (future) compose overlays opt into `MCPGATEWAY_SKIP_MIGRATIONS=true`.

[...]

PR #4636, opened 2026-05-07
fix(rbac): add unique constraints for role/user_role race (#4482)  

### LINKED ISSUE #4482: [BUG] RBAC role/user_role seeder race when fast-path skips advisory lock
## Summary

After PR #4444 lands the L1 fast-path in `bootstrap_db.main()`, the post-migration bootstrap (`_run_post_migration_bootstrap`) runs **without holding the migration advisory lock** on every replica that takes the fast path. Two of the seed steps it runs are race-prone read-then-insert patterns with no DB-tier uniqueness:

1. `bootstrap_default_roles()` → `RoleService.create_role()` for each of the 5 system roles
2. `RoleService.assign_role_to_user()` for the platform_admin assignment

`Role.id` and `UserRole.id` are client-generated UUIDs, so PK collision never saves us. Concurrent fast-path workers can each pass the existence check and each insert. Once duplicates exist, the next `get_role_by_name` / `get_user_role_assignment` raises `MultipleResultsFound` and downstream RBAC checks return 500.

## Trigger conditions

- Multi-replica gateway deployment OR multi-worker single-pod gunicorn
- Schema already at Alembic head (so the fast-path probe returns true)
- One of: `migration.enabled=false` in the chart, or non-Helm deployments where each gateway pod runs its own bootstrap
- First startup of a fresh DB (when system role rows do not yet exist)

The chart default (`migration.enabled=true` + `MCPGATEWAY_SKIP_MIGRATIONS=true` on gateway pods) is **not** affected — the migration Job is the only writer in that configuration.

## Impact

- Severity: HIGH (`roles`) / MEDIUM (`user_roles`) per reviewer triage on PR #4444
- Window: only on the first startup of a fresh DB; once seeded, all replicas see the existing rows and no inserts are attempted
- Recovery: one SQL statement to soft-delete duplicate active rows by `created_at`

## Original reviewer text (verbatim from PR #4444)

> **2. High:** the new fast-path makes system-role bootstrap non-serialised, but role creation is still a read-then-insert with no database uniqueness. `mcpgateway/bootstrap_db.py:139` and `mcpgateway/bootstrap_db.py:782` now run `bootstrap_default_roles()` outside the advisory lock; that function still does `get_role_by_name()` then `create_role()` at `mcpgateway/bootstrap_db.py:512` and `mcpgateway/services/role_service.py:196`, while the `roles` table has no `(name, scope)` uniqueness in `mcpgateway/db.py:1126` or the original migration. Two workers recovering a half-bootstrapped DB can both create `platform_admin`/`team_admin`/etc., and once duplicates exist, `mcpgateway/services/role_service.py:251` starts throwing `MultipleResultsFound`.

> **3. Medium:** the same unlocked path can also duplicate the admin's `platform_admin` assignment. `mcpgateway/bootstrap_db.py:547` does another read-then-insert through `mcpgateway/services/role_service.py:677` and `mcpgateway/services/role_service.py:604`, but `user_roles` also has no uniqueness guard in `mcpgateway/db.py:1168` or its migration. If the role exists but the assignment is missing, concurrent fast-path workers can insert multiple active rows, and later lookups will fail the same way with `MultipleResultsFound`.

## Proposed fix (two-layer)

- **DB tier — partial unique indexes** `WHERE is_active = true`:
  - `roles(name, scope)`
  - `user_roles(user_email, role_id, scope)` for `scope_id IS NULL`
  - `user_roles(user_email, role_id, scope, scope_id)` for `scope_id IS NOT NULL`
  - (split for nullable `scope_id` because Postgres and SQLite both treat NULL as distinct in plain unique indexes)
- **App tier — savepoint + refetch**: wrap inserts in `db.begin_nested()`; on `IntegrityError`, rollback the savepoint and return the winner's row instead of surfacing the error.
- **Alembic migration**: idempotent dedupe-then-constrain — soft-delete duplicate active rows (oldest by `created_at` / `granted_at` wins) before adding the indexes so the migration is safe on databases that have already raced. Audit history preserved.

Implemented in PR #4480 (draft, base = PR #4444's branch — auto-retargets to `main` after #4444 merges).

## Acceptance criteria

- [ ] Multi-replica gateway startup on a fresh DB cannot produce duplicate active `roles` or `user_roles` rows
- [ ] The fix is enforced at the DB tier (partial unique index), not just in application code
- [ ] Existing deployments with pre-existing duplicate active rows can still apply the migration without manual cleanup
- [ ] Integration test pins the invariant under real concurrency

rakdutta:
closes #4482 - Adds database-level unique constraints to prevent duplicate RBAC roles and role assignments during concurrent bootstrap.

## Problem

Multiple workers can create duplicate system roles (`platform_admin`, etc.) and role assignments when bootstrapping concurrently, causing `MultipleResultsFound` exceptions → 500 errors.

## Solution

**Two-layer defense:**

1. **Database constraints** — Partial unique indexes on `roles(name, scope)` and `user_roles` (split for NULL/NOT NULL `scope_id`)
2. **App handling** — Savepoint pattern catches `IntegrityError`, refetches existing row, returns transparently

### Behavior Change: Re-granting an expired assignment no longer raises an error

Previously, `assign_role_to_user()` raised `ValueError("User already has this role assignment")` even when the existing assignment had already expired, because the active-assignment check did not distinguish expired rows from valid ones.

**New behavior:** if an active assignment exists but `is_expired()` returns `True`, it is automatically soft-deleted (`is_active = False`) in-place before the new grant is created. This allows the re-grant to proceed cleanly without any manual revocation step.

A concurrent-race variant is also handled: if two processes race to soft-delete the expired row simultaneously, the process that loses the race refetches the current state and returns the winner's fresh assignment rather than raising.

**Impact:** callers that previously caught `ValueError` on re-grant of an expired role will no longer see that error — the call now succeeds and returns the new `UserRole` row. Any code that relied on that error to detect "assignment already exists" should instead check the returned object's `expires_at` field or use `list_user_roles()`.

## Changes

- **Migration `d21698ae4a19`** — Deduplicates existing duplicate active rows (remapping `user_roles.role_id` to the kept role before deactivating duplicates, and preferring unexpired/most-recently-granted assignments), then adds partial unique indexes (PostgreSQL + SQLite)
- **`role_service.py`** — Updated `create_role()` and `assign_role_to_user()` with savepoint + `IntegrityError` retry handling; fixed W1203 logging violations (lazy `%s` formatting throughout)
- **Tests** — New `tests/unit/mcpgateway/db/test_rbac_unique_constraints_migration.py` covering dedup logic, role_id remapping, expiry-preference ordering, index idempotency, and downgrade
- **CHANGELOG** — `### Fixed` entry added under `[Unreleased]`

## Related

- Issue #4482
- PR #4444 (bootstrap fast-path)


rakdutta:
@gandhipratik203 can you please review this PR.

mcpgateway/db.py
```diff
@@ -1187,6 +1192,12 @@ class UserRole(Base):
     """User role assignment model."""
 
     __tablename__ = "user_roles"
+    __table_args__ = (
+        # Partial unique indexes: only one active assignment per (user, role, scope, scope_id) combination
+        # Need two separate indexes to handle NULL vs non-NULL scope_id cases (SQL NULL != NULL semantics)
+        Index("uq_user_roles_email_role_scope_null_active", "user_email", "role_id", "scope", unique=True, postgresql_where=text("scope_id IS NULL AND is_active = true"), sqlite_where=text("scope_id IS NULL AND is_active = 1")),
```
gandhipratik203: Worth double-checking the expired-assignment case here. assign_role_to_user() allows a new assignment when an existing active assignment is expired, but this unique index only filters on is_active=true. That means an expired active row can still block the re-grant, and the IntegrityError path may refetch and return the expired assignment instead of creating a usable one.

mcpgateway/alembic/versions/d21698ae4a19_add_rbac_unique_constraints_race_fix.py
```diff
@@ -0,0 +1,311 @@
...
+    # PostgreSQL uses DELETE...RETURNING, SQLite needs subquery approach
+    if dialect == 'postgresql':
+        dedupe_roles_sql = text("""
+            UPDATE roles
+            SET is_active = false
+            WHERE id IN (
+                SELECT r.id
+                FROM roles r
+                INNER JOIN (
+                    SELECT name, scope, MIN(created_at) as oldest_created_at
```
gandhipratik203: Small migration edge case: this dedupe keeps rows tied on the minimum timestamp. If duplicate rows share created_at or granted_at, more than one active row can remain and the unique index creation may still fail. A row_number() over (...) order by timestamp and id approach would make the cleanup deterministic.

mcpgateway/alembic/versions/d21698ae4a19_add_rbac_unique_constraints_race_fix.py
```diff
@@ -0,0 +1,311 @@
...
+"""
+from typing import Sequence, Union
+
+from alembic import op
+from sqlalchemy import text, inspect
+
+
+# revision identifiers, used by Alembic.
+revision: str = 'd21698ae4a19'  # pragma: allowlist secret
+down_revision: Union[str, Sequence[str], None] = 'w7x8y9z0a1b2'  # pragma: allowlist secret
```
gandhipratik203: This migration parent looks stale relative to current main. GitHub is also marking the PR as conflicting, so this likely needs a rebase and the down_revision should point at the current Alembic head.

mcpgateway/alembic/versions/d21698ae4a19_add_rbac_unique_constraints_race_fix.py
rakdutta: Updated migration to use ROW_NUMBER() OVER (PARTITION BY name, scope ORDER BY created_at, id) for deterministic tie-breaking. No more timestamp collision issues.

mcpgateway/db.py
rakdutta: You're right that the unique index can't filter on expiration (SQLite limitation). Fixed with two layers:

Before creation: Check and soft-delete expired assignments (lines 646-670)
After IntegrityError: Check refetched assignment for expiration, soft-delete if expired and raise ValueError (lines 704-709)
This prevents returning expired assignments. Added 3 integration tests + 4 unit tests covering all race scenarios.

mcpgateway/alembic/versions/d21698ae4a19_add_rbac_unique_constraints_race_fix.py
rakdutta: Branch rebased. down_revision now points to e198602c3c1e, and alembic heads confirms d21698ae4a19 is current.

gandhipratik203:
Thanks for the previous changes. Two smaller follow-ups I noticed:

- `mcpgateway/alembic/versions/d21698ae4a19_add_rbac_unique_constraints_race_fix.py:62`: Could we also remap `user_roles.role_id` from duplicate roles to the kept role before deactivating them? Otherwise active assignments tied to the deactivated role may stop showing up, since `list_user_roles()` filters on active joined roles.

- `mcpgateway/alembic/versions/d21698ae4a19_add_rbac_unique_constraints_race_fix.py:112` / `:147`: Could the user-role dedupe prefer unexpired assignments over expired ones? Keeping only the oldest `granted_at` row may preserve an expired assignment and deactivate a still-valid duplicate.

msureshkumar88 (changes requested):
Thanks for tackling this — the two-layer defense (DB partial unique indexes + savepoint/retry) is the right shape for this bug, and the migration correctly dedupes before constraining. I ran the migration SQL and the concurrency-handling path independently (including a real multi-threaded SQLite stress test) and the core mechanism holds up well.

## Blocking

- **Lint gate fails**: `pylint` reports 6 new `W1203` (`logging-fstring-interpolation`) violations in `role_service.py` at lines 239, 245, 253, 696, 705, 713 — the new log calls use f-strings where the rest of the file (and this PR's own unmodified lines) use lazy `%s` formatting, e.g. `logger.info(f"Created role: {role.name}...")` vs. the pre-existing `logger.info("Created role: %s...", role.name, ...)`. This trips `make pylint`, part of the required `make ruff bandit interrogate pylint verify` gate. Straightforward fix — revert those calls to lazy formatting.

## Suggestions (non-blocking)

- **Migration has no dedicated test.** `d21698ae4a19...py` is the largest new file here (285 lines, dialect-branching SQL + idempotency checks) but isn't exercised by any test. The repo has a clear convention for this (e.g. `tests/unit/mcpgateway/db/test_token_uniqueness_migration.py`) — importing the module and running `upgrade()`/`downgrade()` against a scratch schema. Worth adding, given this file has the most novel logic in the PR.
- **Race tests are all single-threaded mocks.** The 8 new "concurrent" tests simulate the race via `patch.object` + call counters rather than real concurrent threads/connections against the actual constraint. A single true multi-threaded (or multiprocess) test hammering `create_role`/`assign_role_to_user` against a real SQLite/Postgres unique index would give much stronger confidence this stays fixed under refactors.
- **CHANGELOG.md** isn't updated — this fixes a real production 500 under concurrent bootstrap, which seems worth a line given the repo tracks fixes of similar severity there.
- Minor: the PR description's "Solution" section doesn't mention that expired-but-active assignments are now auto soft-deleted and replaced on re-grant (previously blocked with `ValueError`). Small but real behavior change worth calling out explicitly.

Nothing security-sensitive stood out — checked the new `ValueError` messages that embed raw `IntegrityError` text and confirmed they're routed through `safe_error_detail()` in `rbac.py`, so no leak to non-debug callers.

rakdutta:
@msureshkumar88 @gandhipratik203 — all review comments addressed, ready for re-review.

**Blocking fix (msureshkumar88)**
- Fixed 6 W1203 `logging-fstring-interpolation` violations in `role_service.py` (lines 239, 245, 253, 696, 705, 713) — converted to lazy `%s` formatting. `pylint` gate now passes at 10.00/10.

**Review comments (gandhipratik203)**
- `d21698ae4a19`: Added Step 1a — remaps `user_roles.role_id` to the kept role *before* deactivating duplicate roles, so `list_user_roles()` JOIN on `roles.is_active` stays intact.
- `d21698ae4a19`: Changed `user_roles` dedupe ordering from `ORDER BY granted_at ASC` to prefer unexpired rows first (`expires_at IS NULL OR expires_at > NOW()`), then newest `granted_at DESC` — avoids keeping an expired assignment over a still-valid one.

**Non-blocking suggestions (msureshkumar88)**
- Added `tests/unit/mcpgateway/db/test_rbac_unique_constraints_migration.py` — 22 tests covering dedup logic, `role_id` remapping, expiry-preference ordering, index idempotency, and downgrade (22/22 pass).
- Updated `CHANGELOG.md` `[Unreleased]` with a `### Fixed` entry.
- PR description updated to call out the behavior change on expired-assignment re-grant.

Also fixed a `down_revision` pointing to the wrong ancestor (was `e198602c3c1e`, should be `b6c7d8e9f0a1`) which caused a duplicate-heads error in the upgrade validation CI job.

msureshkumar88 (dismissed):
Thanks for addressing all the prior feedback — the remap-before-dedupe ordering, the unexpired-preference dedupe, the W1203 lint fixes, and the new migration test suite all check out. Ran the full test surface locally (migration tests, duplicate-handling integration tests, race-condition tests, role_service unit tests) and everything is green, and `pylint` on `role_service.py` is 10.00/10.

One thing to clean up before merge:

## Required

**`CHANGELOG.md` re-adds two entries that are already documented under the released `[1.0.5]` section.**

The `[Unreleased]` section this PR adds to (commit `e0f7f751a`) includes:
- `### Added` — `POST /v1/tools/generate-schemas-from-openapi` (#5142)
- `### Security` — cross-environment JWT fix (GHSA-vgf8-3685-66j9)

Both are already fully documented under `## [1.0.5] - 2026-07-07` further down the same file (lines ~51 and ~74) — that release shipped 10 days ago. Re-adding them to `[Unreleased]` will misleadingly re-announce already-shipped features as still-unreleased on the next release cut, and they're unrelated to the RBAC fix this PR is about. Likely a stale-branch/rebase artifact.

Please trim `[Unreleased]` down to just the RBAC `### Fixed` entry this PR actually adds.

## Non-blocking notes (no action required)

- `assign_role_to_user`'s `IntegrityError` refetch path: if the refetched "winner" row is itself already expired (both racers create near-simultaneously-expiring assignments), the code soft-deletes it and raises `ValueError` rather than retrying — surfaces as an HTTP 400 for a caller who did nothing wrong. Very low likelihood, just flagging for awareness.
- The new PostgreSQL-specific dedupe/remap SQL in the migration (`UPDATE ... FROM ... JOIN ...` with window functions) has no test coverage — all 22 new tests run against SQLite only. Matches existing repo convention (other migrations aren't PG-tested either), so not a PR-specific gap, but worth knowing since it's a data-mutating migration.
- `AsyncMock` imported but unused in two test functions in `test_role_service_duplicate_handling.py`.

Nothing else blocking — the core fix (partial unique indexes + savepoint/retry) is solid and well-tested.

msureshkumar88:
## Suggestion (non-blocking)

**`CHANGELOG.md` re-adds two entries that are already documented under the released `[1.0.5]` section.**

The `[Unreleased]` section this PR adds to (commit `e0f7f751a`) includes:
- `### Added` — `POST /v1/tools/generate-schemas-from-openapi` ([#5142](https://github.com/IBM/mcp-context-forge/issues/5142): "[UI-REWRITE]: Create a new /v1/tools/generate-schemas-from-openapi endpoint")
- `### Security` — cross-environment JWT fix (GHSA-vgf8-3685-66j9)

Both are already fully documented under `## [1.0.5] - 2026-07-07` further down the same file (lines ~51 and ~74) — that release shipped 10 days ago. Re-adding them to `[Unreleased]` will re-announce already-shipped features as still-unreleased on the next release cut, and they're unrelated to the RBAC fix this PR is about. Likely a stale-branch/rebase artifact.

Worth trimming `[Unreleased]` down to just the RBAC `### Fixed` entry this PR actually adds — not blocking merge.

msureshkumar88 (dismissed):
## E2E verification: PASSED

Built and ran a dedicated e2e reproduction script for this PR — `scripts/verify-pr-4636-e2e.sh` (added to the branch). It doesn't mock or reimplement anything: it drives the real, unmodified `RoleService.create_role()` / `assign_role_to_user()` and the real `mcpgateway.bootstrap_db.main()` entrypoint under genuine multi-process OS concurrency, first against a `main`-branch worktree (bug present) to prove the race reproduces, then against this PR's head commit (`540124da9`) to prove it's closed.

### Config used

- Two isolated `uv`-managed venvs, one per git worktree (`origin/main` and PR head `540124da9`), each with its own editable install so the two branches' code can never cross-contaminate.
- SQLite DB, `DATABASE_URL=sqlite:///<scratch>.db`, `EMAIL_AUTH_ENABLED` toggled per phase, minimal hermetic env (`JWT_SECRET_KEY`, `AUTH_ENCRYPTION_SECRET`, `PLATFORM_ADMIN_EMAIL`, `PLUGINS_ENABLED=false`, `MCPGATEWAY_UI_ENABLED=false`, `LOG_LEVEL=ERROR`) — no dependency on the local dev `.env`.
- Concurrency: `multiprocessing` (spawn context) + a `Barrier(20)` so all 20 worker processes call the racy method at the exact same instant — this is what makes the reproduction deterministic rather than relying on lucky OS scheduling.


### Commands run (literal, from the actual run — top-level entrypoint plus every real command it executed)


**Worktree + venv setup**
```
git fetch origin main --quiet
git worktree add -q /tmp/tmp.X7eioOojIa/main-wt origin/main
git worktree add -q /tmp/tmp.X7eioOojIa/pr-wt 540124da9ef95b8dc6b6892f8436cb16e1f590ff

uv venv -q /tmp/tmp.X7eioOojIa/main-wt/.venv --python 3.12
uv pip install -q -e /tmp/tmp.X7eioOojIa/main-wt --python /tmp/tmp.X7eioOojIa/main-wt/.venv/bin/python3
uv venv -q /tmp/tmp.X7eioOojIa/pr-wt/.venv --python 3.12
uv pip install -q -e /tmp/tmp.X7eioOojIa/pr-wt --python /tmp/tmp.X7eioOojIa/pr-wt/.venv/bin/python3
```

**Sanity check (venvs resolve the correct branch's code)**
```
cd /tmp/tmp.X7eioOojIa/main-wt && /tmp/tmp.X7eioOojIa/main-wt/.venv/bin/python3 -c \
  "from mcpgateway.services.role_service import RoleService; import inspect; \
   print('IntegrityError' in inspect.getsource(RoleService.create_role))"
# -> False (main has no fix)

cd /tmp/tmp.X7eioOojIa/pr-wt && /tmp/tmp.X7eioOojIa/pr-wt/.venv/bin/python3 -c \
  "from mcpgateway.services.role_service import RoleService; import inspect; \
   print('IntegrityError' in inspect.getsource(RoleService.create_role))"
# -> True (PR head has the fix)
```

**Test 1 — role-creation race, main (bug present)**
```
cd /tmp/tmp.X7eioOojIa/main-wt && env -i PATH="$PATH" HOME="$HOME" \
  DATABASE_URL="sqlite:///tmp/tmp.X7eioOojIa/roles_main_1.db" \
  JWT_SECRET_KEY="test-secret-key-for-e2e-verification-only" \
  AUTH_ENCRYPTION_SECRET="test-encryption-secret-32-chars-min" \
  PLATFORM_ADMIN_EMAIL="admin@example.com" PLATFORM_ADMIN_PASSWORD="ChangeMe123!" \
  PLATFORM_ADMIN_FULL_NAME="Platform Administrator" \
  EMAIL_AUTH_ENABLED="false" AUTO_CREATE_PERSONAL_TEAMS="false" \
  MCPGATEWAY_UI_ENABLED="false" PLUGINS_ENABLED="false" LOG_LEVEL="ERROR" \
  /tmp/tmp.X7eioOojIa/main-wt/.venv/bin/python3 -m mcpgateway.bootstrap_db

cd /tmp/tmp.X7eioOojIa/main-wt && env -i PATH="$PATH" HOME="$HOME" \
  DATABASE_URL="sqlite:///tmp/tmp.X7eioOojIa/roles_main_1.db" \
  JWT_SECRET_KEY="test-secret-key-for-e2e-verification-only" \
  AUTH_ENCRYPTION_SECRET="test-encryption-secret-32-chars-min" LOG_LEVEL="ERROR" \
  /tmp/tmp.X7eioOojIa/main-wt/.venv/bin/python3 /tmp/tmp.X7eioOojIa/role_race_harness.py \
  "sqlite:///tmp/tmp.X7eioOojIa/roles_main_1.db" 20 platform_admin global admin@example.com
```
`role_race_harness.py` (embedded by the script, not a mock) spawns 20 real `multiprocessing` (spawn) OS processes sharing a `Barrier(20)`; each imports the actual `mcpgateway.db.SessionLocal` and `mcpgateway.services.role_service.RoleService`, waits on the barrier, then calls the unmodified `RoleService(session).create_role(name="platform_admin", scope="global", ...)`.

**Test 1 — same, PR head (fix applied)**
```
cd /tmp/tmp.X7eioOojIa/pr-wt && env -i ... DATABASE_URL="sqlite:///tmp/tmp.X7eioOojIa/roles_pr.db" ... \
  EMAIL_AUTH_ENABLED="false" ... /tmp/tmp.X7eioOojIa/pr-wt/.venv/bin/python3 -m mcpgateway.bootstrap_db

cd /tmp/tmp.X7eioOojIa/pr-wt && env -i ... DATABASE_URL="sqlite:///tmp/tmp.X7eioOojIa/roles_pr.db" ... \
  /tmp/tmp.X7eioOojIa/pr-wt/.venv/bin/python3 /tmp/tmp.X7eioOojIa/role_race_harness.py \
  "sqlite:///tmp/tmp.X7eioOojIa/roles_pr.db" 20 platform_admin global admin@example.com
```

**Downstream crash/clean-lookup check (both branches)**
```
/tmp/tmp.X7eioOojIa/main-wt/.venv/bin/python3 -c "
import asyncio
from mcpgateway.db import SessionLocal
from mcpgateway.services.role_service import RoleService
s = SessionLocal()
r = asyncio.run(RoleService(s).get_role_by_name('platform_admin', 'global'))
"   # DATABASE_URL=sqlite:///tmp/tmp.X7eioOojIa/roles_main_1.db -> raises MultipleResultsFound

# same against DATABASE_URL=sqlite:///tmp/tmp.X7eioOojIa/roles_pr.db -> returns cleanly
```

**Test 2 — assignment race (same pattern, both branches)**
```
python3 -c "
import sqlite3
db = sqlite3.connect('<db>')
role_id = db.execute(\"SELECT id FROM roles WHERE name='team_admin' AND scope='team' AND is_active=1\").fetchone()[0]
db.execute(\"DELETE FROM user_roles WHERE role_id=? AND scope='team' AND scope_id='race-team'\", (role_id,))
db.commit()
print(role_id)
"

env -i ... /tmp/tmp.X7eioOojIa/{main-wt,pr-wt}/.venv/bin/python3 /tmp/tmp.X7eioOojIa/assign_race_harness.py \
  "sqlite:///<db>" 20 admin@example.com <role_id> team race-team admin@example.com
```
Same barrier pattern, this time against the unmodified `RoleService.assign_role_to_user()`.

**Test 3 — migration idempotency, PR head**
```
cd /tmp/tmp.X7eioOojIa/pr-wt && env -i PATH="$PATH" HOME="$HOME" \
  DATABASE_URL="sqlite:///tmp/tmp.X7eioOojIa/roles_pr.db" \
  /tmp/tmp.X7eioOojIa/pr-wt/.venv/bin/python3 -m alembic -c mcpgateway/alembic.ini heads
# -> d21698ae4a19 (head)   [single head]

cd /tmp/tmp.X7eioOojIa/pr-wt && env -i PATH="$PATH" HOME="$HOME" \
  DATABASE_URL="sqlite:///tmp/tmp.X7eioOojIa/roles_pr.db" \
  /tmp/tmp.X7eioOojIa/pr-wt/.venv/bin/python3 -m alembic -c mcpgateway/alembic.ini upgrade head
# -> exit 0, no-op on an already-migrated DB
```

**Test 4 — PR's own test suite**
```
cd /home/suresh/pr_others/block1/mcp-context-forge && \
  /home/suresh/pr_others/block1/mcp-context-forge/.venv/bin/python3 -m pytest \
  tests/unit/mcpgateway/db/test_rbac_unique_constraints_migration.py \
  tests/integration/test_role_service_duplicate_handling.py \
  tests/unit/mcpgateway/services/test_role_service.py \
  tests/unit/mcpgateway/services/test_role_service_race_conditions.py \
  --with-integration -q
# -> exit 0, 111 passed
```
(This directory is the exact PR head commit — verified via `git rev-parse HEAD` == `540124da9ef95b8dc6b6892f8436cb16e1f590ff` before running.)

**Cleanup**
```
git worktree remove --force /tmp/tmp.X7eioOojIa/main-wt
git worktree remove --force /tmp/tmp.X7eioOojIa/pr-wt
```

### Output (representative run — repeated twice, stable both times)

```
== Test 1: concurrent RoleService.create_role() race (issue #4482) ==
attempt 1: RESULT ok=19 unique_ids=19 crashed_procs=0
PASS: main (pre-fix) reproduces the bug: 19 duplicate active platform_admin roles created
RESULT ok=20 unique_ids=1 crashed_procs=0
PASS: PR head: exactly 1 platform_admin role survives 20-way concurrent create_role(), 0 process crashes
PASS: main (pre-fix): downstream get_role_by_name() crashes with MultipleResultsFound (the 500 from #4482)
PASS: PR head: downstream get_role_by_name() succeeds cleanly, no MultipleResultsFound

== Test 2: concurrent RoleService.assign_role_to_user() race ==
attempt 1: RESULT ok=1 unique_ids=1 crashed_procs=0
attempt 2: RESULT ok=19 unique_ids=19 crashed_procs=0
PASS: main (pre-fix) reproduces the assignment race: 19 duplicate assignments and/or MultipleResultsFound errors
RESULT ok=17 unique_ids=1 crashed_procs=0
PASS: PR head: exactly 1 assignment survives 20-way concurrent assign_role_to_user(), 0 crashes, 0 MultipleResultsFound

== Test 3: migration idempotency & single alembic head (PR head) ==
PASS: single alembic head: d21698ae4a19 (head)
PASS: re-running 'alembic upgrade head' on an already-migrated DB is a clean no-op

== Test 4: PR's test suite (pytest, repo venv at PR head) ==
PASS: pytest suite: exit 0, 111 tests passed (no F/E in output)

ALL CHECKS PASSED
```

On `main`, 20 concurrent workers each independently created their own row for the same `(platform_admin, global)` role — up to 19/20 duplicates, and the very next lookup crashed with `sqlalchemy.exc.MultipleResultsFound`, exactly the 500 described in #4482. Against this PR's code, the identical race converges to exactly one row every time (losers transparently refetch the winner via the new savepoint/`IntegrityError` handling), zero process crashes, zero `MultipleResultsFound`.

No changes required — approving.

Lang-Akshay (dismissed):
LGTM ✅

gandhipratik203 (dismissed):
The previous review points look addressed:

- `user_roles.role_id` is now remapped before duplicate roles are deactivated.
- `user_roles` dedupe now prefers unexpired rows, then newest `granted_at`.
- `down_revision` now points to `b6c7d8e9f0a1`, and Alembic validation passed.
- W1203 logging issues were fixed.
- Dedicated migration tests and changelog entry were added.

gandhipratik203:
Thanks @rakdutta, the pr is good to go. We have all the needed approvals. There is a merge conflict. Once it's rebased, we are good to merge.

rakdutta:
@gandhipratik203 rebased

gandhipratik203:
Thanks @rakdutta. The PR is in good shape now. @msureshkumar88, @Lang-Akshay We can prioratise this pr for the merge once the merge queue opens up after the release.

gandhipratik203 (dismissed):

Lang-Akshay (dismissed):
LGTM

msureshkumar88 (approved):
LGMT

PR #4341, opened 2026-04-20
fix(security): prevent admin bypass from accessing private resources

bogdanmariusc10:
## 🔗 Related Issue
Closes #4323
**Jira Issue:** https://jsw.ibm.com/browse/ICACF-21

---

## 📝 Summary

This PR fixes a security vulnerability where admin users could view private resources (tools, prompts, servers) belonging to other users through API endpoints. The vulnerability violated the documented principle that private resources should only be accessible to their owners.

**Key Changes:**
- Modified service layer to deny admin bypass for private resources
- Admin bypass now only grants access to public and team resources (NOT private)
- Fixed API endpoints to properly pass user context for authorization
- Added comprehensive test coverage for the security fix

**Security Impact:**
- Private resources are now properly protected at both service and API layers
- Admin role no longer implicitly grants access to other users' private resources
- Unauthorized access attempts return generic "not found" errors to prevent information disclosure

---

## 🏷️ Type of Change
- [x] Bug fix (Security vulnerability)
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Chore (deps, CI, tooling)
- [ ] Other (describe below)

---

## 🧪 Verification

| Check                     | Command         | Status |
|---------------------------|-----------------|--------|
| Lint suite                | `make lint`     | ✅ Pass |
| Unit tests                | `make test`     | ✅ Pass |
| Coverage ≥ 80%            | `make coverage` | ✅ Pass |

**Test Coverage:**
- Unit tests verify admin bypass is denied for private resources
- Unit tests verify admin bypass is granted for team resources
- Unit tests verify owners can access their own private resources
- Integration tests updated to reflect new security behavior
- All tests pass with the security fix in place

---

## ✅ Checklist
- [x] Code formatted (`make black isort pre-commit`)
- [x] Tests added/updated for changes
- [x] Documentation updated (inline comments documenting security requirements)
- [x] No secrets or credentials committed

---

## 📓 Notes

**Files Modified:**
1. **Service Layer (Core Security Logic):**
   - `mcpgateway/services/base_service.py` - Filter private resources in list operations
   - `mcpgateway/services/prompt_service.py` - Deny admin bypass for private prompts
   - `mcpgateway/services/tool_service.py` - Deny admin bypass for private tools + access check in get_tool()
   - `mcpgateway/services/resource_service.py` - Deny admin bypass for private resources

2. **API Layer (Request Handling):**
[...]

PR #4080, opened 2026-04-08
fix(sso): SSO users with platform_admin role can now access teams in admin UI

bogdanmariusc10:
## 🔗 Related Issue
Closes #4070

---

## 📝 Summary

Fixed a bug where SSO-authenticated users with `platform_admin` RBAC role could not edit gateways or virtual servers in the admin UI, receiving "invalid team name" errors despite having `["*"]` (wildcard) effective permissions.

**Root Cause:** The `GET /teams` endpoint incorrectly checked only the database `is_admin` flag instead of RBAC permissions. SSO users have the correct RBAC role (`platform_admin` with `["*"]` permissions) but lack the database flag (`is_admin=False`), causing them to be treated as regular users.

**Solution:** Updated the `list_teams` endpoint to check both the legacy `is_admin` flag AND RBAC permissions (`["*"]` or `["teams.*"]`), aligning with ContextForge's two-layer security model.

---

## 🏷️ Type of Change
- [x] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Chore (deps, CI, tooling)
- [ ] Other (describe below)

---

## 🧪 Verification

| Check                     | Command         | Status |
|---------------------------|-----------------|--------|
| Lint suite                | `make lint`     | ✅ Pass |
| Unit tests                | `make test`     | ✅ Pass |
| Coverage ≥ 80%            | `make coverage` | ✅ Pass |

---

## ✅ Checklist
- [x] Code formatted (`make black isort pre-commit`)
- [] Tests added/updated for changes
- [] Documentation updated (if applicable)
- [x] No secrets or credentials committed

---

### Changes Made

**1. `mcpgateway/routers/teams.py` (Primary Fix)**
```python
# Before (buggy):
if current_user_ctx.get("is_admin"):
    # Only checked database flag

# After (fixed):
has_admin_team_access = (
    current_user_ctx.get("is_admin") or
    "*" in current_user_ctx.get("permissions", []) or
    "teams.*" in current_user_ctx.get("permissions", [])
)
if has_admin_team_access:
    # Now checks RBAC permissions too
```

**2. `mcpgateway/observability.py` (Linting Fix)**
- Added missing `message` parameter documentation to fix flake8 DAR101 error

### Impact

**Before:**
- ❌ SSO users with `platform_admin` role: blocked from editing in admin UI
- ✅ Local admins with `is_admin=True`: worked correctly
- ✅ REST API: worked for both (uses RBAC)

**After:**
[...]

PR #4654, opened 2026-05-07
fix(chore): centralize Layer-1 visibility filter across REST endpoints

### LINKED ISSUE #4451: [CHORE]: Centralize Layer-1 visibility filter across main.py REST endpoints
### Background

`mcpgateway/main.py` contains roughly 22 near-identical inline copies of the Layer-1 admin-bypass + public-only-secure-default derivation across REST endpoint handlers (`list_tools`, `list_resources`, `list_prompts`, `list_a2a_agents`, `list_tags`, `get_entities_by_tag`, several JSON-RPC dispatcher branches, etc.). Each call site looks roughly like:

```python
user_email, token_teams, is_admin = get_rpc_filter_context(request, user)
    user_email = None
    token_teams = None  # Admin unrestricted
elif token_teams is None:
    token_teams = []  # Non-admin without teams = public-only (secure default)
```

`mcpgateway/auth_context.py` already provides `get_scoped_resource_access_context(request, user)` that encapsulates exactly this rule (and `get_request_identity` for the auditor-identity portion).

### Why this matters

- **Maintainability**: 22 copies drift independently. Two reviews of PR #4080 found subtle differences between them (different comments, missing `token_teams = None` for admin, mixed audit-context capture patterns).
- **Risk of regressions**: each copy is one place where a future change to the Layer-1 contract could be missed.
- **Consistency with `routers/teams.py`**: the team router was migrated to `is_admin_bypass_granted` in PR #4070-related work; main.py REST endpoints should follow.

### What was attempted

PR #4080 commit `a1acd3691` ("refactor(main): centralize Layer-1 visibility filter via auth_context helpers") replaced the 22 inline copies with calls to `get_scoped_resource_access_context` / `get_request_identity`. **That commit has been reverted** because:

1. Scope creep: the refactor was not the cause of the original SSO admin bug (issue #4070), so it inflated PR #4080 unnecessarily.
2. Test churn: 14+ tests had to switch their `@patch` targets, several broke in subtle ways during rebases.
3. Latent behavior change: `get_scoped_resource_access_context` correctly enables Layer-1 admin bypass for basic-auth/dev-mode admins (which the inline rule missed). This is technically a bug-fix bundled with the refactor — separating the two would be cleaner.

### Acceptance criteria

- All 22 inline Layer-1 derivations in `main.py` replaced with calls to `get_scoped_resource_access_context(request, user)` (and `get_request_identity` where the audit-identity values are also needed).
- 4 known exceptions retained: `a2a get_agent` and `invoke_agent` (preserve `user_email` for federation `user_id` construction), `_get_internal_a2a_scope_context` (custom return contract), and the internal MCP `tools/invoke` handler that captures `run_owner_email` / `run_owner_team_ids` before applying the rule.
- Tests updated to mock `get_scoped_resource_access_context` (post-rule 2-tuple) instead of the lower-level `get_rpc_filter_context` (pre-rule 3-tuple).
- The basic-auth/dev-mode admin behavior change is documented in the commit message and covered by a regression test.

### Out of scope for this issue

- Anything in `streamablehttp_transport.py` — see the companion issue.
- Anything in `routers/teams.py` (already migrated).

### References

- Issue #4070 (parent SSO platform_admin bug)
- PR #4080 (where the refactor was first attempted)

bogdanmariusc10:
## 🔗 Related Issue

Closes #4451

---

## 📝 Summary

Centralizes Layer-1 visibility filtering across `mcpgateway/main.py` by replacing inline derivations of the admin-bypass + public-only-secure-default rule with calls to `get_scoped_resource_access_context()` from `mcpgateway/auth_context.py`.

**Problem**: the rule was hand-copied at every call site. The copies drifted — different comments, and more importantly two incompatible behaviours for admin bypass (some preserved `user_email`, some nulled it). Each copy is a place a future change to the Layer-1 contract can be missed.

**Result**: inline derivations drop from 33 to 6, with the 6 remaining documented in-line as deliberate exceptions.

---

## 🏷️ Type of Change
- [x] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [x] Refactor
- [ ] Other (describe below)

---

## ⚠️ Behaviour changes (both intentional)

Response shapes, error codes, and non-admin visibility are unchanged. What changes is how many rows an **admin** token sees. The boundary that hides *other* users' private rows is unchanged and covered by deny-path tests. Recorded under **Breaking Changes** in `CHANGELOG.md`.

**1. Basic-auth / dev-mode admins gain admin bypass across all 27 migrated call sites.**
The superseded inline derivation read `is_admin` only from a verified JWT payload, so an admin authenticating without one — basic auth, or `AUTH_REQUIRED=false` local setups — was narrowed to public-only. Such callers now get the intended bypass: public + team + their own private rows. This is the wider-reaching of the two changes and affects an entire authentication mode.

**2. JWT-authenticated admins now see their own private rows on 9 endpoints.**
These sites discarded the caller's email when granting bypass, which drops *every* private row from the result — including the admin's own — because the service layer needs the email to owner-match:

`GET /tags`, `GET /tags/{tag}/entities`, JSON-RPC `completion/complete`, and the internal MCP `tools/list`, `resources/list`, `resources/read`, `prompts/list`, `prompts/get`, and `completion/complete` handlers.

---

### Preserved exceptions (6)

These still call `get_rpc_filter_context()` directly and carry an inline `# Layer-1 exception` comment naming the reason:

| Function | Reason |
|---|---|
| `_build_internal_mcp_auth_context_for_rpc` | Forwards an auth context rather than deriving visibility scope; needs the raw `is_admin` flag |
| `_authorize_run_cancellation` | Compares requester against run owner; needs raw token teams and `is_admin`, not the normalized scope |
| `_execute_rpc_tools_call` | Captures run ownership from the raw context before admin-bypass normalization is applied |
| `create_mcp_app_session` | Keeps the resource-scope email separate from `user_email`, which the single-value helper contract cannot express |
| `handle_internal_mcp_tools_call_resolve` | Tool-execution authorization, not resource visibility |
| `tools/call` branch of `_handle_rpc_authenticated` | Tool-execution authorization; kept in sync with `_execute_rpc_tools_call` |

The two tool-execution sites are deliberately left alone: routing them through the helper would widen admin *execution* scope to the admin's own private tools, which is a separate behavioural change and belongs in its own PR.

### Other fixes in this change

- **Request-scoped memoization** of the derived triple, keyed on the identity of the `user` object rather than the request alone. Handlers needing both the scope and the requester identity previously derived twice, and that derivation can issue a live `EmailUser` lookup for session tokens. Keying on the user object matters: one request may derive for more than one principal (the synthetic forwarded user built for trusted internal A2A dispatch), and those must not read each other's cached entry.
- **Malformed cached JWT payload no longer raises.** `get_rpc_filter_context()` called `.get()` on the cached payload without a type check, turning a non-dict payload into a JSON-RPC `-32603 Internal error` instead of a normal RBAC decision. Guarded with `isinstance`, and it now logs a warning like the sibling type-check below it.
- **Corrected the documented contract.** Both the module docstring and `get_scoped_resource_access_context()`'s own docstring stated that admin bypass returns `(None, None)`; the code returns `(email, None)` by design, so the email survives for owner matching. `(None, None)` means bypass for a caller with no resolvable email.

### Tests

- **`tests/unit/mcpgateway/test_auth_context.py`** (new) — pins the Layer-1 rule itself. `auth_context.py` had no test file, and endpoint tests mock the helper, so nothing exercised the rule directly once it moved out of `main.py`. Also covers the memoization and malformed-payload behaviour above.
- **`tests/integration/test_admin_bypass_owner_visibility.py`** (new) — real DB, real login, real service layer, across `GET /tags`, `GET /tags/{tag}/entities`, JSON-RPC `completion/complete`, and the internal MCP `prompts/list` and `resources/list` handlers. Each asserts both directions: the admin sees their own private row, and still cannot see another user's. Covers 4 of the 9 endpoints above; the remaining 5 share the same `BaseService._apply_visibility_scope` enforcement.
- **Mock retargeting** — tests patching `mcpgateway.main.get_rpc_filter_context` against a migrated handler no longer intercept anything, so the real derivation ran instead and they passed without verifying what they were written to verify. Those are retargeted to the centralized helper with post-rule 2-tuples. The patches that remain all target the 6 preserved sites.

### Docs

- `AGENTS.md` gains an **Implementation Helpers** section under *Token Scoping Quick Reference* describing when to use `get_scoped_resource_access_context()`, `get_request_identity()`, and the raw `get_rpc_filter_context()`.
- `CHANGELOG.md` entry under **Breaking Changes**, split so the basic-auth change is its own headline rather than a trailing clause.

### Scope

Per the issue's own out-of-scope note, `streamablehttp_transport.py` and `routers/teams.py` are untouched. No schema or Alembic change — this is a runtime filter.

---

## ✅ Checklist
- [x] Documentation updated (`AGENTS.md`, `CHANGELOG.md`)

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^\\.secrets\\.baseline$",
     "lines": null
   },
-  "generated_at": "2026-08-07T10:14:38Z",
+  "generated_at": "2026-08-07T13:52:41Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -216,7 +216,7 @@
         "hashed_secret": "f4793151b0607198d4de9b1ca458d3e25adf1cb7",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 179,
+        "line_number": 189,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -224,7 +224,7 @@
         "hashed_secret": "fa9beb99e4029ad5a6615399e7bbae21356086b3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 292,
+        "line_number": 302,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -232,7 +232,7 @@
         "hashed_secret": "7b4455a56fbf1d198e45e04c437488514645a82c",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 294,
+        "line_number": 304,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -240,7 +240,7 @@
         "hashed_secret": "90bd1b48e958257948487b90bee080ba5ed00caa",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 366,
+        "line_number": 376,
         "type": "Hex High Entropy String",
         "verified_result": null
       },
@@ -248,7 +248,7 @@
         "hashed_secret": "48ffbad96aa9c2b33f9486f5a3c2108198acb518",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 367,
+        "line_number": 377,
         "type": "Hex High Entropy String",
         "verified_result": null
       }
@@ -258,7 +258,7 @@
         "hashed_secret": "b4673e578b9b30fe8bba1b555b7b59883444c697",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 2032,
+        "line_number": 2039,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -266,7 +266,7 @@
         "hashed_secret": "4a0a2df96d4c9a13a282268cab33ac4b8cbb2c72",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 2120,
+        "line_number": 2127,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -274,7 +274,7 @@
         "hashed_secret": "5baa61e4c9b93f3f0682250b6cf8331b7ee68fd8",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 2470,
+        "line_number": 2477,
         "type": "Basic Auth Credentials",
         "verified_result": null
       },
@@ -282,7 +282,7 @@
         "hashed_secret": "fa9beb99e4029ad5a6615399e7bbae21356086b3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 3835,
+        "line_number": 3842,
         "type": "Basic Auth Credentials",
         "verified_result": null
       },
@@ -290,7 +290,7 @@
         "hashed_secret": "fa9beb99e4029ad5a6615399e7bbae21356086b3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 3926,
+        "line_number": 3933,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -298,7 +298,7 @@
         "hashed_secret": "ac371b6dcce28a86c90d12bc57d946a800eebf17",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 3969,
+        "line_number": 3976,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -306,7 +306,7 @@
         "hashed_secret": "0b6ec68df700dec4dcd64babd0eda1edccddace1",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 3974,
+        "line_number": 3981,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -314,7 +314,7 @@
         "hashed_secret": "4ad6f0082ee224001beb3ca5c3e81c8ceea5ed86",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 3979,
+        "line_number": 3986,
         "type": "Secret Keyword",
         "verified_result": null
       }
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -173,6 +173,16 @@ ContextForge implements a **two-layer security model**:
 - **Layer 1 only**: Token scoping controls visibility (what you can see). RBAC (Layer 2) is evaluated independently — session-token narrowing does not restrict which team roles are checked for permissions.
 - **External IdP tokens**: identities provisioned from trusted external SSO providers (see `SSO_API_TOKEN_AUTH_ENABLED`) are dispatched through the session-token table above (`resolve_session_teams()`), not the API/legacy table — `is_admin`/`teams` come from the persisted local user record, never from the external token's claims.
 
+### Implementation Helpers
+
+Layer-1 derivation is centralized in `mcpgateway/auth_context.py`. Route handlers must call these rather than re-deriving the rule inline:
+
+- `get_scoped_resource_access_context(request, user)` → `(user_email, token_teams)` — the visibility scope to pass into service fetch/list/read calls. Admin bypass is signalled by `token_teams=None` while `user_email` is **kept**, so the service can still owner-match the admin's own private rows. Public-only is `(email, [])`.
+- `get_request_identity(request, user)` → `(user_email, is_admin)` — the requester's own identity, for audit capture and header masking. Use when you need the identity independent of the visibility scope.
+- `get_rpc_filter_context(request, user)` → `(user_email, token_teams, is_admin)` — the raw pre-rule triple. Reserved for the few sites that genuinely need `is_admin` or the un-normalized teams (auth-context forwarding, run-ownership capture, tool-execution authorization); these carry a `Layer-1 exception` comment naming the reason.
+
+The derived triple is memoized on `request.state` per principal, so calling the scope and identity helpers together costs one derivation rather than two.
+
 ### Security Invariants (Required)
 
 - Treat `public` as platform-public scope, not internet-anonymous scope.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,13 @@
 
 ## [Unreleased]
 
+### Breaking Changes
+
+- **Admins see more rows from visibility-filtered endpoints** ([#4451](https://github.com/IBM/mcp-context-forge/issues/4451)) - Layer-1 visibility derivation in `main.py` is now centralized on `get_scoped_resource_access_context()` instead of being re-implemented inline at each call site. Response shapes, error codes, and non-admin visibility are unchanged; what changes is how many rows an **admin** token sees. No configuration or migration step is required, and the boundary that hides *other* users' private rows is unchanged and covered by deny-path tests. Two distinct changes are bundled here:
+
+    - **Basic-auth and dev-mode admins gain admin bypass across every migrated call site** (27 in this change). The superseded inline derivation read `is_admin` only from a verified JWT payload, so an admin authenticating without one - basic auth, or `AUTH_REQUIRED=false` local setups - was narrowed to public-only. Such callers now receive the intended bypass: public + team + their own private rows. This is the wider-reaching of the two changes and affects an entire authentication mode.
+    - **JWT-authenticated admins now see their own private rows on 10 endpoints** that previously discarded the caller's email when granting bypass, which dropped *every* private row including the admin's own: `GET /tags`, `GET /tags/{tag}/entities`, the JSON-RPC `completion/complete` method, the internal MCP `tools/list`, `resources/list`, `resources/read`, `prompts/list`, `prompts/get`, and `completion/complete` handlers, and `POST /appbridge/sessions`. On the AppBridge endpoint the effect was a hard failure rather than a short list: an admin opening a session against a `ui://` resource they own received `404 Resource not found`.
+
 ## [1.0.7] - 2026-08-04 - Security Hardening, Unified Search, OAuth Improvements, Dataplane Enhancements, and Operational Reliability
 
 ### Overview
--- a/mcpgateway/auth_context.py
+++ b/mcpgateway/auth_context.py
@@ -784,7 +784,9 @@ def get_rpc_filter_context(request: Request, user) -> tuple[Optional[str], Optio
     cached = getattr(request.state, "_jwt_verified_payload", None)
     if cached and isinstance(cached, tuple) and len(cached) == 2:
         _, payload = cached
-        if payload:
+        # A malformed (non-dict) cached payload carries no usable admin claim; fall through
+        # with is_admin unchanged so the caller defers to RBAC instead of raising.
+        if isinstance(payload, dict):
             # Session tokens ignore JWT is_admin claim — DB is the authority.
             # An old/stale session JWT carrying is_admin=true must not influence
             # the boolean admin decision; only DB-resolved token_teams=None can
--- a/mcpgateway/auth_context.py
+++ b/mcpgateway/auth_context.py
@@ -103,11 +103,14 @@ policy. The key invariants that this module enforces:
 2. ``get_rpc_filter_context`` derives ``is_admin`` from the verified JWT
    payload or the trusted internal MCP auth context - NOT from the DB user -
    so a scoped token (``teams=[]``) cannot inherit admin bypass.
-3. ``get_scoped_resource_access_context`` returns ``(None, None)`` *only* for
-   genuine admin bypass (verified JWT ``is_admin=true`` + ``teams=null``, or
-   non-JWT dev-mode admin). Public-only tokens get ``(email, [])``.
-   Downstream services MUST treat ``(None, None)`` as "admin bypass; still
-   deny private resources" per PR #4341.
+3. ``get_scoped_resource_access_context`` signals admin bypass with
+   ``token_teams=None``, and *only* for genuine admin bypass (verified JWT
+   ``is_admin=true`` + ``teams=null``, or non-JWT dev-mode admin). It keeps
+   ``user_email`` set on that path so the service can owner-match the admin's
+   own private rows; ``(None, None)`` therefore means bypass for a caller with
+   no resolvable email. Public-only tokens get ``(email, [])``. Downstream
+   services MUST treat ``token_teams=None`` as "admin bypass; still deny other
+   users' private resources" per PR #4341.
 4. Non-JWT admin callers (basic-auth / dev mode) keep unrestricted visibility
    via the fallback-admin branch; this carve-out is intentional and documented
    in ``AGENTS.md``.
@@ -720,6 +723,62 @@ def get_token_teams_from_request(request: Request) -> Optional[List[str]]:
     return []
 
 
+# Attribute used to memoize the derived filter context for the life of one request.
+# Handlers that need both the visibility scope and the requester identity call
+# get_scoped_resource_access_context() and get_request_identity() back to back, and
+# both derive via get_rpc_filter_context(); for session tokens that derivation can
+# issue a live EmailUser lookup, so without memoization those handlers pay the query
+# twice per request.
+_RPC_FILTER_CONTEXT_CACHE_ATTR = "_rpc_filter_context_cache"
+
+
+def _get_cached_rpc_filter_context(request: Request, user) -> Optional[tuple[Optional[str], Optional[List[str]], bool]]:
+    """Return the memoized filter context for ``user`` on ``request``, if present.
+
+    The entry is keyed on the identity of the ``user`` object, not just the request:
+    a single request may derive context for more than one principal (for example the
+    synthetic forwarded user built for trusted internal A2A calls), and those must not
+    read each other's cached result.
+
+    Args:
+        request: Incoming request context.
+        user: Authenticated user context the caller is deriving for.
+
+    Returns:
+        The cached ``(user_email, token_teams, is_admin)`` triple, or ``None`` on miss.
+    """
+    state = getattr(request, "state", None)
+    if state is None:
+        return None
+    entry = getattr(state, _RPC_FILTER_CONTEXT_CACHE_ATTR, None)
+    # Guard the shape explicitly: request.state is frequently a MagicMock in tests,
+    # where a plain truthiness check would return a mock instead of missing.
+    if isinstance(entry, tuple) and len(entry) == 2 and entry[0] is user and isinstance(entry[1], tuple) and len(entry[1]) == 3:
+        return entry[1]
+    return None
+
+
+def _cache_rpc_filter_context(request: Request, user, context: tuple[Optional[str], Optional[List[str]], bool]) -> tuple[Optional[str], Optional[List[str]], bool]:
+    """Memoize ``context`` for ``user`` on ``request`` and return it unchanged.
+
+    Args:
+        request: Incoming request context.
+        user: Authenticated user context the result was derived for.
+        context: The ``(user_email, token_teams, is_admin)`` triple to cache.
+
+    Returns:
+        ``context``, so callers can ``return _cache_rpc_filter_context(...)`` directly.
+    """
+    state = getattr(request, "state", None)
+    if state is not None:
+        try:
+            setattr(state, _RPC_FILTER_CONTEXT_CACHE_ATTR, (user, context))
+        except (AttributeError, TypeError):
+            # A read-only or exotic state object is not fatal; skip memoization.
+            pass
+    return context
+
+
 def get_rpc_filter_context(request: Request, user) -> tuple[Optional[str], Optional[List[str]], bool]:
     """Extract ``(user_email, token_teams, is_admin)`` for RPC filtering.
 
@@ -751,6 +810,10 @@ def get_rpc_filter_context(request: Request, user) -> tuple[Optional[str], Optio
         >>> is_admin
         True
     """
+    cached_context = _get_cached_rpc_filter_context(request, user)
+    if cached_context is not None:
+        return cached_context
+
     # Use existing get_user_email() helper for consistent email extraction
     user_email = get_user_email(user)
     # get_user_email() guarantees a string return, but may return "unknown"
@@ -779,13 +842,19 @@ def get_rpc_filter_context(request: Request, user) -> tuple[Optional[str], Optio
         is_admin = bool(internal_auth_context.get("is_admin", False))
         if token_teams is not None and len(token_teams) == 0:
             is_admin = False
-        return user_email, token_teams, is_admin
+        return _cache_rpc_filter_context(request, user, (user_email, token_teams, is_admin))
 
     cached = getattr(request.state, "_jwt_verified_payload", None)
     if cached and isinstance(cached, tuple) and len(cached) == 2:
         _, payload = cached
         # A malformed (non-dict) cached payload carries no usable admin claim; fall through
         # with is_admin unchanged so the caller defers to RBAC instead of raising.
+        if payload is not None and not isinstance(payload, dict):
+            logger.warning(
+                "get_rpc_filter_context: verified JWT payload non-dict type=%s path=%s; treating caller as non-admin",
+                type(payload).__name__,
+                getattr(getattr(request, "url", None), "path", "unknown"),
+            )
         if isinstance(payload, dict):
             # Session tokens ignore JWT is_admin claim — DB is the authority.
             # An old/stale session JWT carrying is_admin=true must not influence
@@ -831,7 +900,7 @@ def get_rpc_filter_context(request: Request, user) -> tuple[Optional[str], Optio
                 db_user_is_admin,
             )
 
-    return user_email, token_teams, is_admin
+    return _cache_rpc_filter_context(request, user, (user_email, token_teams, is_admin))
 
 
 async def is_unrestricted_platform_admin(request: Request, user: Any, db: Session) -> bool:
@@ -931,9 +1000,13 @@ def get_scoped_resource_access_context(request: Request, user) -> tuple[Optional
     the canonical ``(user_email, token_teams)`` input shape for service-layer
     visibility checks:
 
-    - ``(None, None)``: admin bypass. The service still applies the post-PR
-      #4341 rule "admin bypass may see public + team, never another user's
-      private".
+    - ``(email, None)``: admin bypass. ``user_email`` is deliberately preserved
+      so the service can still owner-match the admin's own private rows. The
+      service applies the rule "admin bypass may see public + team + own
+      private, never another user's private".
+    - ``(None, None)``: admin bypass for a caller with no resolvable email
+      (anonymous / dev-mode). The service returns public + team and no private
+      rows at all, since there is no owner to match against.
     - ``(email, [])``: public-only token. Service returns public rows only.
     - ``(email, ["team-a", ...])``: team-scoped token. Service returns
       public rows + team-scoped rows for the listed teams + the caller's own
--- a/mcpgateway/main.py
+++ b/mcpgateway/main.py
@@ -471,6 +471,9 @@ def _build_internal_mcp_auth_context_for_rpc(request: Request, user: Any) -> Dic
     Returns:
         Encodable auth-context dict for ``encode_internal_mcp_auth_context``.
     """
+    # Layer-1 exception: this builds an auth context to forward, it does not derive
+    # visibility scope. It needs the raw pre-rule triple (including is_admin) rather
+    # than the post-rule (user_email, token_teams) from get_scoped_resource_access_context.
     email, token_teams, is_admin = get_rpc_filter_context(request, user)
     # Genuine anonymous / MCP_REQUIRE_AUTH=false public-only callers have no email.
     is_authenticated = email is not None
@@ -1019,6 +1022,9 @@ async def _authorize_run_cancellation(request: Request, user, request_id: str, *
         JSONRPCError: When ``as_jsonrpc_error`` is True and cancellation is not authorized.
         HTTPException: When ``as_jsonrpc_error`` is False and cancellation is not authorized.
     """
+    # Layer-1 exception: run-cancellation authorization compares the requester against the
+    # run owner. It needs the raw token teams and is_admin flag, not the post-rule visibility
+    # scope, so the admin-bypass normalization in get_scoped_resource_access_context does not apply.
     requester_email, requester_token_teams, requester_is_admin = get_rpc_filter_context(request, user)
     requester_teams = [] if requester_token_teams is None else list(requester_token_teams)
     run_status = await cancellation_service.get_status(request_id)
@@ -4876,13 +4882,7 @@ async def get_a2a_agent(
             raise HTTPException(status_code=503, detail="A2A service not available")
 
         # Get filtering context from token (respects token scope)
-        user_email, token_teams, is_admin = get_rpc_filter_context(request, user)
-
-        # Admin bypass - only when token has NO team restrictions
-        if is_admin and token_teams is None:
-            token_teams = None  # Admin unrestricted
-        elif token_teams is None:
-            token_teams = []  # Non-admin without teams = public-only
+        user_email, token_teams = get_scoped_resource_access_context(request, user)
 
         return await a2a_service.get_agent(
             db,
@@ -5188,13 +5188,7 @@ def _extract_a2a_request_context(
         - request_headers: Dict[str, str]
     """
     # Get filtering context from token (respects token scope)
-    user_email, token_teams, is_admin = get_rpc_filter_context(request, user)
-
-    # Admin bypass - only when token has NO team restrictions
-    if is_admin and token_teams is None:
-        token_teams = None  # Admin unrestricted
-    elif token_teams is None:
-        token_teams = []  # Non-admin without teams = public-only
+    user_email, token_teams = get_scoped_resource_access_context(request, user)
 
     # Extract user ID
     user_id = None
@@ -8534,11 +8528,7 @@ async def handle_internal_mcp_resources_read(request: Request):
                 },
             )
 
-        auth_user_email, auth_token_teams, auth_is_admin = get_rpc_filter_context(request, user)
-        if auth_is_admin and auth_token_teams is None:
-            auth_user_email = None
-        elif auth_token_teams is None:
-            auth_token_teams = []
+        auth_user_email, auth_token_teams = get_scoped_resource_access_context(request, user)
 
         plugin_context_table = getattr(request.state, "plugin_context_table", None)
         plugin_global_context = getattr(request.state, "plugin_global_context", None)
@@ -9331,11 +9321,7 @@ async def handle_internal_mcp_prompts_get(request: Request):
                 },
             )
 
-        auth_user_email, auth_token_teams, auth_is_admin = get_rpc_filter_context(request, user)
-        if auth_is_admin and auth_token_teams is None:
-            auth_user_email = None
-        elif auth_token_teams is None:
-            auth_token_teams = []
+        auth_user_email, auth_token_teams = get_scoped_resource_access_context(request, user)
 
         plugin_context_table = getattr(request.state, "plugin_context_table", None)
         plugin_global_context = getattr(request.state, "plugin_global_context", None)
@@ -9652,13 +9638,7 @@ async def _authorize_internal_a2a_method(
 def _get_internal_a2a_scope_context(request: Request) -> tuple[Optional[str], Optional[List[str]]]:
     """Return scoped visibility context for trusted internal A2A requests."""
     user = _build_internal_mcp_forwarded_user(request)
-    user_email, token_teams, is_admin = get_rpc_filter_context(request, user)
-
-    if is_admin and token_teams is None:
-        return user_email, None
-    if token_teams is None:
-        return user_email, []
-    return user_email, token_teams
+    return get_scoped_resource_access_context(request, user)
 
 
 @utility_router.post("/_internal/a2a/invoke/authz/")
@@ -10267,9 +10247,7 @@ async def _mcp_apps_initialize_authorized(request: Request, db: Session, user, s
     if not server_id:
         return True
 
-    user_email, token_teams, is_admin = get_rpc_filter_context(request, user)
-    if not (is_admin and token_teams is None) and token_teams is None:
-        token_teams = []
+    user_email, token_teams = get_scoped_resource_access_context(request, user)
 
     try:
         await server_service.get_server(db, server_id, user_email=user_email, token_teams=token_teams)
@@ -10372,6 +10350,9 @@ async def _execute_rpc_tools_call(
     if not name:
         raise JSONRPCError(-32602, "Missing tool name in parameters", params)
 
+    # Layer-1 exception: run ownership (run_owner_email / run_owner_team_ids) must be captured
+    # from the raw pre-rule context before the admin-bypass normalization is applied, so this
+    # site cannot collapse to get_scoped_resource_access_context.
     auth_user_email, auth_token_teams, auth_is_admin = get_rpc_filter_context(request, user)
     run_owner_email = auth_user_email
     run_owner_team_ids = [] if auth_token_teams is None else list(auth_token_teams)
@@ -10487,6 +10468,10 @@ async def create_mcp_app_session(request: Request, db: Session = Depends(get_db)
     server_id = body.get("serverId") or body.get("server_id") or request.headers.get("x-contextforge-server-id")
     if not server_id or not isinstance(server_id, str):
         raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail="serverId is required for MCP Apps sessions")
+    # Layer-1 exception: this site keeps user_email and the resource-scope email as separate
+    # values (resource_user_email is nulled for admin bypass while user_email is retained for
+    # the session record), which is not the single-value contract of
+    # get_scoped_resource_access_context. See the tool-execution follow-up noted in issue #4451.
     user_email, token_teams, is_admin = get_rpc_filter_context(request, user)
     if is_admin and token_teams is None:
         resource_user_email = None
@@ -10963,6 +10948,9 @@ async def handle_internal_mcp_tools_call_resolve(request: Request):
         if (get_internal_mcp_auth_context(request) or {}).get("is_authenticated", True) is True:
             await _ensure_rpc_permission(user, db, "tools.execute", "tools/call", request=request)
 
+        # Layer-1 exception: tool-execution authorization, not resource visibility. Migrating this
+        # to get_scoped_resource_access_context would widen admin execution scope to the admin's own
+        # private tools, which is a behavior change outside the scope of issue #4451.
         auth_user_email, auth_token_teams, auth_is_admin = get_rpc_filter_context(request, user)
         if auth_is_admin and auth_token_teams is None:
             auth_user_email = None
@@ -11169,16 +11157,9 @@ async def _handle_tools_list_rpc(
     Raises:
         HTTPException: If permission check fails
     """
-    user_email, token_teams, is_admin = get_rpc_filter_context(request, user)
-    _req_email, _req_is_admin = user_email, is_admin
+    user_email, token_teams = get_scoped_resource_access_context(request, user)
+    _req_email, _req_is_admin = get_request_identity(request, user)
     _req_team_roles = get_user_team_roles(db, _req_email) if _req_email and not _req_is_admin else None
-
-    # Admin bypass - only when token has NO team restrictions
-    if is_admin and token_teams is None:
-        # user_email stays as-is (not None) for owner matching (PR #4341 / issue #4694)
-        token_teams = None  # Admin unrestricted
-    elif token_teams is None:
-        token_teams = []  # Non-admin without teams = public-only (secure default)
 
     if server_id:
         tools = await tool_svc.list_server_tools(
@@ -11419,12 +11400,7 @@ async def _handle_rpc_authenticated(request: Request, db: Session, user):
                 raise JSONRPCError(-32602, "Missing resource URI in parameters", params)
 
             # Get authorization context (same as resources/list)
-            auth_user_email, auth_token_teams, auth_is_admin = get_rpc_filter_context(request, user)
-            if auth_is_admin and auth_token_teams is None:
-                # Keep auth_user_email set for owner matching on the admin's own private rows (PR #4341 / issue #4694)
-                pass  # auth_token_teams stays None (unrestricted)
-            elif auth_token_teams is None:
-                auth_token_teams = []  # Non-admin without teams = public-only
+            auth_user_email, auth_token_teams = get_scoped_resource_access_context(request, user)
 
             # Get user email for OAuth token selection
             oauth_user_email = get_user_email(user)
@@ -11510,12 +11486,7 @@ async def _handle_rpc_authenticated(request: Request, db: Session, user):
                 raise JSONRPCError(-32602, "Missing prompt name in parameters", params)
 
             # Get authorization context (same as prompts/list)
-            auth_user_email, auth_token_teams, auth_is_admin = get_rpc_filter_context(request, user)
-            if auth_is_admin and auth_token_teams is None:
-                # Keep auth_user_email set for owner matching on the admin's own private rows (PR #4341 / issue #4694)
-                pass  # auth_token_teams stays None (unrestricted)
-            elif auth_token_teams is None:
-                auth_token_teams = []  # Non-admin without teams = public-only
+            auth_user_email, auth_token_teams = get_scoped_resource_access_context(request, user)
 
             # Get plugin contexts from request.state for cross-hook sharing
             plugin_context_table = getattr(request.state, "plugin_context_table", None)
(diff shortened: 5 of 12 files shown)


marekdano (commented):
Some findings

### High

- **`handle_internal_mcp_resources_read` not migrated** — still uses
    the old inline pattern that sets `auth_user_email = None` for admin, breaking
    owner-matching on private resources via the internal MCP read path. Not in the
    diff, not a documented exception.

- **`handle_internal_mcp_prompts_get` not migrated** — same issue.
    Admin loses owner-matching when fetching prompts via the internal MCP path.

### Medium

- **PR description claims `prompts/get` (RPC dispatcher) was
    migrated — it was not.** The diff skips from `prompts/list` directly to
    `completion/complete`, leaving `prompts/get` with the old inline pattern.
    Four other RPC dispatcher branches (`resources/read`, `prompts/get`, legacy
    tool execute) and `handle_internal_mcp_tools_call_resolve` are also unmigrated
    and undocumented.

- **Test mocks use `(None, None)` for admin bypass** — but
    `get_scoped_resource_access_context` always preserves `user_email` for admin
    (`return user_email, None`). The mocks contradict the function's actual behaviour and silently revert the issue #4694 owner-matching protection in those assertions.

  ---

### Low

- **Log message regression** — three endpoints changed from `safe_log_user(user)`
    to `SecurityValidator.sanitize_log_message(str(user))`. `str(user)` serializes
    the full user dict including `is_admin`, `teams`, and `sub` into debug logs.
    Unrelated to the issue; should be reverted.

- **Basic-auth admin bug persists in `_get_internal_a2a_scope_context`** —
    the preserved exception does not call `_has_verified_jwt_payload`, so basic-auth
    admins still get public-only scope on internal A2A paths. Out of scope for this
    PR but worth a follow-up issue.

### Info

  - PR notes only mention one of the two A2A invoke functions (`invoke_a2a_agent`
    by name); `invoke_a2a_agent_by_id` is also preserved but
    undocumented.

  - The stated reason for preserving `_get_internal_a2a_scope_context`
    ("custom return contract") is inaccurate — the contract is identical to
    `get_scoped_resource_access_context`. The real reason is the different
    user-building path (`_build_internal_mcp_forwarded_user`).

  ---

  ## Summary Table

  | Finding | Severity | Blocking |
  |---|---|---|
  | `handle_internal_mcp_resources_read` not migrated — admin `user_email` nullified | High | Yes |
  | `handle_internal_mcp_prompts_get` not migrated — admin `user_email` nullified | High | Yes |
  | `prompts/get` (RPC) claimed migrated but not in diff | Medium | Yes |
  | 4 other RPC/internal branches unmigrated and undocumented | Medium | Yes |
  | Test mocks `(None, None)` misrepresent actual helper output | Medium | Yes |
  | Log change leaks user dict to debug log | Low | No |
  | Basic-auth bug remains in `_get_internal_a2a_scope_context` | Low | No |
  | Second A2A invoke endpoint not mentioned in PR notes | Info | No |
  | Preserved exception justification is inaccurate | Info | No |

ja8zyjits (dismissed):
LGTM

Low Priority can be skipped or added, too small for a followup

## ⚠️ Warnings

### Helper Functions Not Documented in AGENTS.md

- **File**: `AGENTS.md` (Authentication & RBAC Overview)
- **Code**: `mcpgateway/auth_context.py:get_scoped_resource_access_context()`, `get_request_identity()`
- **Impact**: Low - Internal refactoring. Behavior documented, implementation details appropriately omitted.
- **Recommendation**: Add brief note in `AGENTS.md` under "Token Scoping Quick Reference":

```markdown
### Implementation Helpers

Layer-1 visibility derivation centralized in `mcpgateway/auth_context.py`:

- `get_scoped_resource_access_context(request, user)` → `(user_email, token_teams)`
  - Returns post-rule 2-tuple for resource visibility filtering
  - Use in route handlers needing visibility scope only

- `get_request_identity(request, user)` → `(user_email, is_admin)`  
  - Returns requester identity for audit logging
  - Use when needing original identity before admin bypass

Route handlers in `main.py` should use these helpers instead of inline derivation.
```

msureshkumar88 (changes requested):
Thanks for tackling this — collapsing 22 near-identical inline derivations down to one call site is exactly the kind of cleanup that prevents the drift PR #4080's reviewers flagged, and the new `test_auth_context.py` unit tests pin the helper's contract nicely. Ran the full affected test set (`test_auth_context.py`, `test_a2a_jsonrpc_passthrough.py`, `test_internal_a2a_endpoints.py`, the tag/completion subset of `test_main.py`, and the admin-bypass subset of `test_main_extended.py`) plus `ruff` against this branch — all green, no regressions found there.

A few things worth resolving before merge, and a couple of smaller suggestions.

## Blocking

**1. The "Preserved 4 Exceptions" table doesn't match the diff.**
Of the four listed, only `invoke_a2a_agent` is actually untouched. `get_a2a_agent`, `_get_internal_a2a_scope_context`, and the JSON-RPC `tools/list` handler (`_handle_tools_list_rpc`) were all centralized in this PR — they don't retain the 3-tuple pattern. Meanwhile the functions that *do* still call the raw `get_rpc_filter_context()` (with new "Layer-1 exception" comments explaining why) are `_build_internal_mcp_auth_context_for_rpc`, `_authorize_run_cancellation`, `_execute_rpc_tools_call`, `create_mcp_app_session`, `handle_internal_mcp_tools_call_resolve`, and the JSON-RPC `tools/call` branch — six, not four, and a different set. Could you update the table to match what's actually in the diff? As written it'll mislead the next person who goes looking for where the 3-tuple pattern still lives.

**2. There's a second, undisclosed behavior change beyond the basic-auth fix.**
The PR body calls out one behavior change (basic-auth/dev-mode admin bypass now works) with a solid regression test. But several of the 22 replaced call sites previously nulled `user_email` on admin bypass — a different pattern from the "keep email" one used elsewhere. `get_scoped_resource_access_context()` always preserves email on bypass, and per `base_service.py`'s `_apply_access_control`/`_apply_visibility_scope`, that difference determines whether a DB-admin's own private rows are included. So on `list_tags`, `get_entities_by_tag`, the six `handle_internal_mcp_*` handlers, and the JSON-RPC `completion/complete` branch, a JWT-authenticated admin now sees their own private tags/tools/resources/prompts where before they didn't — this PR's own updated test assertions confirm it (e.g. `test_list_tags_admin_bypass_passes_unrestricted_scope` now expects `user_email="admin@example.com"` instead of `None`).

This is arguably the *correct* contract (it matches the documented PR #4341 invariant), so I'm not saying it's wrong — but it's an authorization-visibility change on 9 endpoints that isn't mentioned anywhere in the PR description, and the only coverage is at the mock level (asserting what gets forwarded to the service), not an end-to-end/integration test proving a real DB-owned private row actually becomes visible. Given this repo's own guidance that security-sensitive changes need explicit regression coverage, could you (a) add a line to the description naming this as a second, intentional behavior change, and (b) add at least one integration test per affected surface against a real DB?

**3. Duplicate `get_rpc_filter_context()` calls on the hottest endpoints.**
`server_get_tools`, `list_tools`, and `_handle_tools_list_rpc` now call both `get_scoped_resource_access_context()` and `get_request_identity()` back-to-back, and both internally re-derive via `get_rpc_filter_context()` — previously one call served both scoping and audit-identity capture. That matters because `get_rpc_filter_context()` can issue a live DB query (`SessionLocal()` + `EmailUser` lookup) for session-token requests with `token_teams is None`, so session-token traffic on these three endpoints now pays an extra DB round trip per request. The non-JWT/basic-auth path pays a similar 2x cost inside `get_scoped_resource_access_context()` itself across all 22 sites. Would suggest caching the `get_rpc_filter_context()` result on `request.state` for the life of the request (mirrors the existing `_jwt_verified_payload` caching already in this file) so callers needing both scoping and identity don't re-derive it.

## Suggestions (non-blocking)

- `get_scoped_resource_access_context()`'s docstring says admin bypass returns `(None, None)`, but the code (and the comment right below it) return `(user_email, None)` — pre-existing, not introduced here, but this PR triples down on callers depending on the real contract, so worth fixing while in the area.
- The new `isinstance(payload, dict)` guard in `auth_context.py` falls through silently on a malformed payload, unlike the sibling type-check a few lines below it which logs a warning for a non-string email. A log line here would help if this ever actually fires in production.

No concerns on the Alembic side — correctly untouched, no schema change needed for a runtime filter. Scope otherwise matches issue #4451 well: `streamablehttp_transport.py` and `routers/teams.py` are correctly left alone per the issue's own out-of-scope note, and I didn't spot any unrelated changes.

Lang-Akshay (dismissed):
LGTM

madhu-mohan-jaishankar:
> Thanks for tackling this — collapsing 22 near-identical inline derivations down to one call site is exactly the kind of cleanup that prevents the drift PR #4080's reviewers flagged, and the new `test_auth_context.py` unit tests pin the helper's contract nicely. Ran the full affected test set (`test_auth_context.py`, `test_a2a_jsonrpc_passthrough.py`, `test_internal_a2a_endpoints.py`, the tag/completion subset of `test_main.py`, and the admin-bypass subset of `test_main_extended.py`) plus `ruff` against this branch — all green, no regressions found there.
> 
> A few things worth resolving before merge, and a couple of smaller suggestions.
> 
> ## Blocking
> **1. The "Preserved 4 Exceptions" table doesn't match the diff.** Of the four listed, only `invoke_a2a_agent` is actually untouched. `get_a2a_agent`, `_get_internal_a2a_scope_context`, and the JSON-RPC `tools/list` handler (`_handle_tools_list_rpc`) were all centralized in this PR — they don't retain the 3-tuple pattern. Meanwhile the functions that _do_ still call the raw `get_rpc_filter_context()` (with new "Layer-1 exception" comments explaining why) are `_build_internal_mcp_auth_context_for_rpc`, `_authorize_run_cancellation`, `_execute_rpc_tools_call`, `create_mcp_app_session`, `handle_internal_mcp_tools_call_resolve`, and the JSON-RPC `tools/call` branch — six, not four, and a different set. Could you update the table to match what's actually in the diff? As written it'll mislead the next person who goes looking for where the 3-tuple pattern still lives.
> 
> **2. There's a second, undisclosed behavior change beyond the basic-auth fix.** The PR body calls out one behavior change (basic-auth/dev-mode admin bypass now works) with a solid regression test. But several of the 22 replaced call sites previously nulled `user_email` on admin bypass — a different pattern from the "keep email" one used elsewhere. `get_scoped_resource_access_context()` always preserves email on bypass, and per `base_service.py`'s `_apply_access_control`/`_apply_visibility_scope`, that difference determines whether a DB-admin's own private rows are included. So on `list_tags`, `get_entities_by_tag`, the six `handle_internal_mcp_*` handlers, and the JSON-RPC `completion/complete` branch, a JWT-authenticated admin now sees their own private tags/tools/resources/prompts where before they didn't — this PR's own updated test assertions confirm it (e.g. `test_list_tags_admin_bypass_passes_unrestricted_scope` now expects `user_email="admin@example.com"` instead of `None`).
> 
> This is arguably the _correct_ contract (it matches the documented PR #4341 invariant), so I'm not saying it's wrong — but it's an authorization-visibility change on 9 endpoints that isn't mentioned anywhere in the PR description, and the only coverage is at the mock level (asserting what gets forwarded to the service), not an end-to-end/integration test proving a real DB-owned private row actually becomes visible. Given this repo's own guidance that security-sensitive changes need explicit regression coverage, could you (a) add a line to the description naming this as a second, intentional behavior change, and (b) add at least one integration test per affected surface against a real DB?
> 
> **3. Duplicate `get_rpc_filter_context()` calls on the hottest endpoints.** `server_get_tools`, `list_tools`, and `_handle_tools_list_rpc` now call both `get_scoped_resource_access_context()` and `get_request_identity()` back-to-back, and both internally re-derive via `get_rpc_filter_context()` — previously one call served both scoping and audit-identity capture. That matters because `get_rpc_filter_context()` can issue a live DB query (`SessionLocal()` + `EmailUser` lookup) for session-token requests with `token_teams is None`, so session-token traffic on these three endpoints now pays an extra DB round trip per request. The non-JWT/basic-auth path pays a similar 2x cost inside `get_scoped_resource_access_context()` itself across all 22 sites. Would suggest caching the `get_rpc_filter_context()` result on `request.state` for the life of the request (mirrors the existing `_jwt_verified_payload` caching already in this file) so callers needing both scoping and identity don't re-derive it.
> 
> ## Suggestions (non-blocking)
> * `get_scoped_resource_access_context()`'s docstring says admin bypass returns `(None, None)`, but the code (and the comment right below it) return `(user_email, None)` — pre-existing, not introduced here, but this PR triples down on callers depending on the real contract, so worth fixing while in the area.
> * The new `isinstance(payload, dict)` guard in `auth_context.py` falls through silently on a malformed payload, unlike the sibling type-check a few lines below it which logs a warning for a non-string email. A log line here would help if this ever actually fires in production.
> 
> No concerns on the Alembic side — correctly untouched, no schema change needed for a runtime filter. Scope otherwise matches issue #4451 well: `streamablehttp_transport.py` and `routers/teams.py` are correctly left alone per the issue's own out-of-scope note, and I didn't spot any unrelated changes.

Thanks @msureshkumar88. Pushed as `9d49c16fa`.

**Duplicate derivation (3):** memoized the derived triple on `request.state`. Keyed on the identity of the `user` object rather than the request alone, since a single request can derive context for more than one principal — the synthetic forwarded user built for trusted internal A2A calls must not read the real caller's entry. Confirmed the scope+identity pair now costs one derivation instead of two.

**Behavior change (2):** added `tests/integration/test_admin_bypass_owner_visibility.py` — real DB, real login, real service layer. It asserts the admin sees their own private row, does **not** see another user's private row, and that non-admins are unaffected. I checked it actually detects the change rather than passing vacuously: reintroducing the old email-nulling in `list_tags` fails the test, and restoring it passes.

**Docstring:** good catch, and it was in two places — the module docstring's security invariant #3 carried the same incorrect `(None, None)` contract. Both now describe what the code does, including what `(None, None)` genuinely means (bypass for a caller with no resolvable email). That inaccuracy is the most likely origin of the `(None, None)` mocks this PR had to unwind. Also added the warning log on the malformed-payload fallthrough.

Separately picked up @ja8zyjits's suggestion and documented the helpers in `AGENTS.md`.

ja8zyjits (dismissed):
LGTM

msureshkumar88 (changes requested):
Thanks for taking this on — the centralization itself is the right fix, and it's clear a lot of care went into it: removing ~126 lines of drift-prone inline logic, the request-scoped memoization for handlers that need both the scope and identity helpers, and the defensive fix for a non-dict cached JWT payload are all nice touches beyond the minimum ask. The `/tags` owner-visibility integration test (`tests/integration/test_admin_bypass_owner_visibility.py`) is exactly the right shape for this kind of change — real DB, real auth pipeline, explicit deny-path assertion.

A couple of things I think are worth addressing before merge, plus a few smaller notes.

## Blocking / functionally-impacting

**1. The admin-bypass "keep `user_email`" behavior change reaches further than the PR describes, and test coverage doesn't fully follow it there.**

Beyond the advertised basic-auth bug fix, several endpoints previously *nulled* `user_email` on admin bypass and now preserve it (via `get_scoped_resource_access_context`'s `(email, None)` contract). That's very likely correct — it makes these endpoints consistent with the rest of the app — but it's a second behavior change bundled into the PR without its own callout, and it lands on:

`list_tags`, `get_entities_by_tag`, `handle_internal_mcp_tools_list`, `handle_internal_mcp_resources_list`, `handle_internal_mcp_resources_read`, `handle_internal_mcp_completion_complete`, `handle_internal_mcp_prompts_list`, `handle_internal_mcp_prompts_get`, and the RPC `completion/complete` branch.

Only the `/tags` pair got the real-DB deny-path test proving "admin sees own private row, still can't see another user's private row." The other six endpoints don't have an equivalent. The underlying enforcement is shared (`BaseService._apply_visibility_scope`), which lowers the actual risk, but AGENTS.md's own invariant asks for deny-path regression tests at the point of change — would you be open to extending `test_admin_bypass_owner_visibility.py` (or a sibling) to cover at least one of the internal-MCP handlers and the RPC completion path the same way?

**2. A number of pre-existing tests in `test_main.py` / `test_main_extended.py` are now patching a function the migrated code no longer calls.**

Since several handlers switched from calling `get_rpc_filter_context` directly to calling `get_scoped_resource_access_context`, any test that still does `patch("mcpgateway.main.get_rpc_filter_context", ...)` against one of those migrated handlers is no longer intercepting anything — the real (unpatched) derivation runs instead. I counted roughly 90 such sites across the two files, spanning ~85 test functions. Most just assert a status code so they keep passing "by accident" rather than failing loudly, which is the concerning part — they've quietly stopped verifying the scoping behavior they were written for. One in particular (`test_invoke_a2a_agent_branches_and_errors`) does assert a specific `token_teams` value and only passes because the real derivation happens to coincidentally produce the same result against that test's mock `Request`.

None of this fails CI today, so it's not a blocker in the "something's broken" sense, but given the size of this PR's own migration, I'd treat fixing these mocks as part of finishing it rather than a follow-up — happy to help script the mechanical swap if useful.

## Suggestions

- The PR description's "Preserved 4 Exceptions" table (`get_a2a_agent`, `invoke_a2a_agent`, `_get_internal_a2a_scope_context`, JSON-RPC `tools/list`) doesn't match the final diff — `get_a2a_agent` and `_get_internal_a2a_scope_context` were fully migrated in a later commit, and the actual preserved exceptions (per the inline `# Layer-1 exception` comments) are six different call sites: `_build_internal_mcp_auth_context_for_rpc`, `_authorize_run_cancellation`, `_execute_rpc_tools_call`, `create_mcp_app_session`, `handle_internal_mcp_tools_call_resolve`, and the `tools/call` branch of `_handle_rpc_authenticated`. Worth updating the table so future readers aren't misled about which endpoints are intentional exceptions.
- Given the PR is framed partly as a bug fix, a short `CHANGELOG.md` entry would help downstream consumers who might notice the visibility change.

## Minor

- The PR checklist has "Documentation updated" unchecked, but `AGENTS.md` was in fact updated with a nice new "Implementation Helpers" section — probably just a checkbox slip.

Nothing here is a correctness bug in the new code itself — the refactor is sound, tests that exist are green, no schema/alembic concerns, and the file footprint is clean. Requesting changes mainly to get the deny-path coverage and the stale mocks addressed given how security-relevant this path is.

madhu-mohan-jaishankar:
> Thanks for taking this on — the centralization itself is the right fix, and it's clear a lot of care went into it: removing ~126 lines of drift-prone inline logic, the request-scoped memoization for handlers that need both the scope and identity helpers, and the defensive fix for a non-dict cached JWT payload are all nice touches beyond the minimum ask. The `/tags` owner-visibility integration test (`tests/integration/test_admin_bypass_owner_visibility.py`) is exactly the right shape for this kind of change — real DB, real auth pipeline, explicit deny-path assertion.
> 
> A couple of things I think are worth addressing before merge, plus a few smaller notes.
> 
> ## Blocking / functionally-impacting
> **1. The admin-bypass "keep `user_email`" behavior change reaches further than the PR describes, and test coverage doesn't fully follow it there.**
> 
> Beyond the advertised basic-auth bug fix, several endpoints previously _nulled_ `user_email` on admin bypass and now preserve it (via `get_scoped_resource_access_context`'s `(email, None)` contract). That's very likely correct — it makes these endpoints consistent with the rest of the app — but it's a second behavior change bundled into the PR without its own callout, and it lands on:
> 
> `list_tags`, `get_entities_by_tag`, `handle_internal_mcp_tools_list`, `handle_internal_mcp_resources_list`, `handle_internal_mcp_resources_read`, `handle_internal_mcp_completion_complete`, `handle_internal_mcp_prompts_list`, `handle_internal_mcp_prompts_get`, and the RPC `completion/complete` branch.
> 
> Only the `/tags` pair got the real-DB deny-path test proving "admin sees own private row, still can't see another user's private row." The other six endpoints don't have an equivalent. The underlying enforcement is shared (`BaseService._apply_visibility_scope`), which lowers the actual risk, but AGENTS.md's own invariant asks for deny-path regression tests at the point of change — would you be open to extending `test_admin_bypass_owner_visibility.py` (or a sibling) to cover at least one of the internal-MCP handlers and the RPC completion path the same way?
> 
> **2. A number of pre-existing tests in `test_main.py` / `test_main_extended.py` are now patching a function the migrated code no longer calls.**
> 
> Since several handlers switched from calling `get_rpc_filter_context` directly to calling `get_scoped_resource_access_context`, any test that still does `patch("mcpgateway.main.get_rpc_filter_context", ...)` against one of those migrated handlers is no longer intercepting anything — the real (unpatched) derivation runs instead. I counted roughly 90 such sites across the two files, spanning ~85 test functions. Most just assert a status code so they keep passing "by accident" rather than failing loudly, which is the concerning part — they've quietly stopped verifying the scoping behavior they were written for. One in particular (`test_invoke_a2a_agent_branches_and_errors`) does assert a specific `token_teams` value and only passes because the real derivation happens to coincidentally produce the same result against that test's mock `Request`.
> 
> None of this fails CI today, so it's not a blocker in the "something's broken" sense, but given the size of this PR's own migration, I'd treat fixing these mocks as part of finishing it rather than a follow-up — happy to help script the mechanical swap if useful.
> 
> ## Suggestions
> * The PR description's "Preserved 4 Exceptions" table (`get_a2a_agent`, `invoke_a2a_agent`, `_get_internal_a2a_scope_context`, JSON-RPC `tools/list`) doesn't match the final diff — `get_a2a_agent` and `_get_internal_a2a_scope_context` were fully migrated in a later commit, and the actual preserved exceptions (per the inline `# Layer-1 exception` comments) are six different call sites: `_build_internal_mcp_auth_context_for_rpc`, `_authorize_run_cancellation`, `_execute_rpc_tools_call`, `create_mcp_app_session`, `handle_internal_mcp_tools_call_resolve`, and the `tools/call` branch of `_handle_rpc_authenticated`. Worth updating the table so future readers aren't misled about which endpoints are intentional exceptions.
> * Given the PR is framed partly as a bug fix, a short `CHANGELOG.md` entry would help downstream consumers who might notice the visibility change.
> 
> ## Minor
> * The PR checklist has "Documentation updated" unchecked, but `AGENTS.md` was in fact updated with a nice new "Implementation Helpers" section — probably just a checkbox slip.
> 
> Nothing here is a correctness bug in the new code itself — the refactor is sound, tests that exist are green, no schema/alembic concerns, and the file footprint is clean. Requesting changes mainly to get the deny-path coverage and the stale mocks addressed given how security-relevant this path is.

Addressed both items in `a94968576`.

**Stale mocks:** retargeted the tests that were patching `get_rpc_filter_context` on migrated handlers — 55 dead patches, now 0. The 17 remaining target the six preserved call sites, where patching the raw helper is still correct.

**Deny-path coverage:** extended `test_admin_bypass_owner_visibility.py` to the RPC `completion/complete` branch and the internal MCP `prompts/list` handler, alongside the existing `/tags` pair. Ten tests, real DB and real service layer. Each verified to fail if the old email-nulling is reintroduced at that call site.

Also added the `CHANGELOG.md` entry. Description table and checkbox handled separately.

msureshkumar88 (changes requested):
Thanks for the fast turnaround on both rounds — the memoization fix is genuinely nice (keying the cache on the `user` object's identity rather than just the request, so the synthetic forwarded A2A user can't collide with the real caller's entry, was the right level of care), and the `/tags` + RPC completion + internal-MCP `prompts/list` integration tests are exactly the right shape: real DB, real login, explicit deny-path assertions. I re-ran them (`--with-integration`) along with `test_main.py`, `test_main_extended.py`, `test_auth_context.py`, `test_a2a_jsonrpc_passthrough.py`, and `test_internal_a2a_endpoints.py` — all green, `ruff` clean on both changed source files, no Alembic/schema concerns.

I did an independent line-by-line audit of the two claims in the latest reply, though, and one of them doesn't hold up.

## Blocking

**1. The stale-mock claim ("55 dead patches, now 0; 17 remaining, all legitimate") isn't accurate — 29 sites are still stale.**

I checked all 42 `get_rpc_filter_context` patch/monkeypatch sites in `test_main.py` + `test_main_extended.py` against what the target function actually calls now. It looks like `a94968576` retargeted the `patch("mcpgateway.main.get_rpc_filter_context", ...)` context-manager sites, but missed the `monkeypatch.setattr(main_mod, "get_rpc_filter_context", ...)` population entirely (27 of the 29 stale sites are this form). These are concentrated in the `list_tools` / `get_tool` / `list_prompts` / `list_a2a_agents` / `server_get_tools` / tag-endpoint coverage:

`test_main_extended.py:3586, 4699, 4776, 4810, 4837, 4864, 4893, 6575, 8127, 9626, 11821, 11869, 11925, 11971, 12004, 12043, 12082, 12118, 12136, 12160, 12185, 12223, 12247, 12287, 12322, 12344, 12378, 12747, 12758`

None of them assert on the mocked email/token_teams values, so they currently pass "by accident" against the real (unpatched) derivation — same failure mode as the originally-flagged issue, just not fully closed out. Given this repo's own history with PR #4080 drift, I'd like to see these actually retargeted rather than left open, since the whole point of this PR is to make that drift detectable again.

**2. The PR description still has the stale content the last two rounds both flagged.**

The "Preserved 4 Exceptions" table (line numbers 4434/4735/8704/9933) still doesn't match the current 6 preserved sites, and "Documentation updated" is still unchecked even though `AGENTS.md` was updated. The last reply said these were "handled separately" — could you actually apply that update? Two rounds flagging the same stale table is more confusing for future readers than the original inaccuracy.

## Suggestions (non-blocking)

- The CHANGELOG entry names 9 affected surfaces for the admin-bypass behavior change, but only 3 have real-DB integration coverage (`/tags` pair, RPC `completion/complete`, internal-MCP `prompts/list`). The other 5 internal-MCP handlers (`tools_list`, `resources_list`, `resources_read`, `completion_complete`, `prompts_get`) share the same underlying `BaseService._apply_visibility_scope` enforcement, so risk is low, but a sibling test class following the existing `TestAdminBypassOwnerVisibilityInternalMcp` pattern for at least one or two more would close the gap between what's documented and what's tested.
- No test currently asserts the memoization itself works — e.g. that calling `get_scoped_resource_access_context()` then `get_request_identity()` with the same `user` triggers only one underlying derivation. Given this was a specific, reviewed perf fix, a regression test (mock `get_user_email` or the JWT-payload lookup and assert call count == 1) would pin it going forward.

Nothing here is a correctness problem in the shipped code — the refactor itself is sound, the new tests that exist are real and pass, and the two prior blocking rounds' core asks (memoization, `/tags` deny-path coverage) are solidly done. Requesting changes mainly to get the mock cleanup actually finished and the description brought in line with the diff.

msureshkumar88:
## This is a breaking change — what it means for existing users

Flagging explicitly since it's not currently labeled or surfaced as one: this PR changes what data admin-authenticated callers get back from 9 endpoints, automatically, on upgrade — no feature flag, no opt-in, no migration step. Response shape is identical (no schema/contract change, no new error codes), only row counts change for admin token holders.

**Who's affected:**

- **JWT-authenticated platform admins** (normal login/API token, `is_admin=true`) — on `GET /tags`, `GET /tags/{tag}/entities`, RPC `completion/complete`, and the 6 internal-MCP handlers (`tools/list`, `resources/list`, `resources/read`, `completion/complete`, `prompts/list`, `prompts/get`), an admin will now see their **own** private tags/tools/resources/prompts in listings that previously hid them. Still can't see another user's private rows — that boundary is unchanged and tested. Mostly reads as a fix, but is a visible behavior change on 9 specific surfaces.

- **Basic-auth / dev-mode admins** — larger shift. The old inline logic only checked `is_admin` from a JWT payload, so an admin without a JWT (basic-auth login, `AUTH_REQUIRED=false` dev setups) was incorrectly narrowed to public-only across **all 22** migrated endpoints. After this PR they get correct admin bypass — public + team + their own private rows — on all 22, not just the 9 above. Anyone running basic-auth admin in production or local dev will see meaningfully more data from these endpoints post-upgrade.

- **Non-admins and team-scoped tokens** — unaffected, tested explicitly.

- **`streamablehttp_transport.py` and `routers/teams.py`** — out of scope for this PR, no change there.

**Practical risk:** low for most, but anyone with automation/dashboards scripted against these endpoints under an admin token could see a one-time jump in row counts post-upgrade with no release note pointing at why.

## Suggested mitigations (not blocking the code itself, but worth doing before/at merge)

1. **Move the CHANGELOG entry into the "Breaking Changes" section.** It's currently under "### Fixed" (CHANGELOG.md:24-26), but this same Unreleased block already has a "### Breaking Changes" section (line 11) used by other entries this release. Ops teams scanning for breaking changes on upgrade skim that section specifically — burying an admin-visibility-widening change under "Fixed" means they'll miss it.

2. **Split the single paragraph into two bullets.** The current entry conflates two different-magnitude changes: (a) JWT admin sees own private rows on 9 endpoints (narrow), and (b) basic-auth/dev-mode admin gets full bypass across all 22 endpoints — public-only → team+own-private (broad, affects a whole auth mode). Point (b) is currently one "Separately, ..." clause at the end of a long paragraph — it deserves its own headline bullet so it doesn't get skimmed past by someone scanning the changelog.

3. **Update the PR description** — the "Preserved 4 Exceptions" table is still stale (doesn't match the current 6 preserved sites), and there's no explicit callout of this as a second, intentional behavior change beyond the basic-auth fix. This has been flagged across two review rounds already. The description is often what gets linked from release notes/upgrade guides, so leaving it stale propagates the same gap downstream.

4. **Close the remaining deny-path test gaps.** Only 3 of the 9 CHANGELOG-documented affected surfaces (`/tags` pair, RPC `completion/complete`, internal-MCP `prompts/list`) have real-DB integration coverage; the other 5 internal-MCP handlers (`tools_list`, `resources_list`, `resources_read`, `completion_complete`, `prompts_get`) don't. This isn't about the safety of the shipped code — the underlying enforcement is shared and low-risk — it's about being able to honestly claim "verified" for all 9 surfaces named in the changelog, not just 3.

#1 and #2 are cheap, pure-documentation changes with no code risk and directly reduce the "admin caught off guard on upgrade" risk. #3 and #4 are already part of the posted review.

madhu-mohan-jaishankar:
> Thanks for the fast turnaround on both rounds — the memoization fix is genuinely nice (keying the cache on the `user` object's identity rather than just the request, so the synthetic forwarded A2A user can't collide with the real caller's entry, was the right level of care), and the `/tags` + RPC completion + internal-MCP `prompts/list` integration tests are exactly the right shape: real DB, real login, explicit deny-path assertions. I re-ran them (`--with-integration`) along with `test_main.py`, `test_main_extended.py`, `test_auth_context.py`, `test_a2a_jsonrpc_passthrough.py`, and `test_internal_a2a_endpoints.py` — all green, `ruff` clean on both changed source files, no Alembic/schema concerns.
> 
> I did an independent line-by-line audit of the two claims in the latest reply, though, and one of them doesn't hold up.
> 
> ## Blocking
> **1. The stale-mock claim ("55 dead patches, now 0; 17 remaining, all legitimate") isn't accurate — 29 sites are still stale.**
> 
> I checked all 42 `get_rpc_filter_context` patch/monkeypatch sites in `test_main.py` + `test_main_extended.py` against what the target function actually calls now. It looks like `a94968576` retargeted the `patch("mcpgateway.main.get_rpc_filter_context", ...)` context-manager sites, but missed the `monkeypatch.setattr(main_mod, "get_rpc_filter_context", ...)` population entirely (27 of the 29 stale sites are this form). These are concentrated in the `list_tools` / `get_tool` / `list_prompts` / `list_a2a_agents` / `server_get_tools` / tag-endpoint coverage:
> 
> `test_main_extended.py:3586, 4699, 4776, 4810, 4837, 4864, 4893, 6575, 8127, 9626, 11821, 11869, 11925, 11971, 12004, 12043, 12082, 12118, 12136, 12160, 12185, 12223, 12247, 12287, 12322, 12344, 12378, 12747, 12758`
> 
> None of them assert on the mocked email/token_teams values, so they currently pass "by accident" against the real (unpatched) derivation — same failure mode as the originally-flagged issue, just not fully closed out. Given this repo's own history with PR #4080 drift, I'd like to see these actually retargeted rather than left open, since the whole point of this PR is to make that drift detectable again.
> 
> **2. The PR description still has the stale content the last two rounds both flagged.**
> 
> The "Preserved 4 Exceptions" table (line numbers 4434/4735/8704/9933) still doesn't match the current 6 preserved sites, and "Documentation updated" is still unchecked even though `AGENTS.md` was updated. The last reply said these were "handled separately" — could you actually apply that update? Two rounds flagging the same stale table is more confusing for future readers than the original inaccuracy.
> 
> ## Suggestions (non-blocking)
> * The CHANGELOG entry names 9 affected surfaces for the admin-bypass behavior change, but only 3 have real-DB integration coverage (`/tags` pair, RPC `completion/complete`, internal-MCP `prompts/list`). The other 5 internal-MCP handlers (`tools_list`, `resources_list`, `resources_read`, `completion_complete`, `prompts_get`) share the same underlying `BaseService._apply_visibility_scope` enforcement, so risk is low, but a sibling test class following the existing `TestAdminBypassOwnerVisibilityInternalMcp` pattern for at least one or two more would close the gap between what's documented and what's tested.
> * No test currently asserts the memoization itself works — e.g. that calling `get_scoped_resource_access_context()` then `get_request_identity()` with the same `user` triggers only one underlying derivation. Given this was a specific, reviewed perf fix, a regression test (mock `get_user_email` or the JWT-payload lookup and assert call count == 1) would pin it going forward.
> 
> Nothing here is a correctness problem in the shipped code — the refactor itself is sound, the new tests that exist are real and pass, and the two prior blocking rounds' core asks (memoization, `/tags` deny-path coverage) are solidly done. Requesting changes mainly to get the mock cleanup actually finished and the description brought in line with the diff.

All addressed — latest `3e29be66a`.

**Mock retargeting** is complete, including the sites that raise before reaching the derivation — those need static classification by handler rather than a runtime check, which is why they weren't in the first pass. The 12 patches remaining all target the 6 preserved sites intentionally. Verified by making every converted patch raise on invocation: 32 tests picked it up, confirming they're genuinely wired in.

On severity, for the record: none of these asserted on `token_teams`/`user_email`, so no assertion was passing incorrectly and no scoping coverage was lost — inert setup rather than a defect.

**Coverage is now 4 of 9, not 3** — `resources/list` landed after your review was written. Each integration test is verified to actually detect the change rather than pass by construction: reintroducing the old email-nulling at that call site fails the corresponding test, restoring it passes. Same approach for the new memoization tests, which include the synthetic-forwarded-principal case that motivated keying the cache on the user object.

**Breaking-change framing taken.** The classification is arguable — response shapes, error codes and non-admin visibility are unchanged, and admin row counts move toward correctness — but the discoverability point stands. The entry now sits under **Breaking Changes**, split into two bullets with the basic-auth change leading, since at 27 migrated call sites it's materially wider than the 9-endpoint owner-matching change.

**Description** rewritten: the exceptions table now lists the actual 6 preserved sites by function name (no line numbers to go stale), both intentional behaviour changes are called out, documentation checkbox ticked.

One gap left deliberately: 5 of the 9 surfaces have no dedicated integration test. They share `BaseService._apply_visibility_scope` with the 4 that do, and `tools/list` additionally needs a seeded virtual server to reach. Happy to add them if you'd rather have all 9 covered before merge.

msureshkumar88 (changes requested):
Nice cleanup — the mock retargeting across `test_main.py`/`test_main_extended.py` is thorough (verified every remaining raw `get_rpc_filter_context` patch actually maps to one of the 6 documented exception sites), the new `test_auth_context.py` file finally pins the Layer-1 rule directly instead of relying on endpoint mocks, and the memoization + malformed-payload fixes are well-targeted, well-tested additions that head off problems the refactor itself would otherwise introduce.

One thing I think is worth fixing before merge, and a couple of smaller notes.

## Blocking

**`create_mcp_app_session` still has the bug this PR fixes everywhere else** (`mcpgateway/main.py:10464`). It's listed as Layer-1 exception #4 ("keeps the resource-scope email separate from `user_email`, which the single-value helper contract cannot express"), but the code underneath still nulls the email on admin bypass:

```python
user_email, token_teams, is_admin = get_rpc_filter_context(request, user)
    resource_user_email = None
else:
    resource_user_email = user_email
    if token_teams is None:
        token_teams = []
```

That flows into `resource_service.read_resource(..., user=resource_user_email, token_teams=token_teams)`, which hits `ResourceService._check_resource_access`:

```python
if token_teams is None and user_email is None:
    return visibility != "private"
```

So an admin creating an AppBridge session against their **own private** UI resource gets `user=None, token_teams=None` → the private-resource check fails → `ResourceNotFoundError` → 404, even though they own it. Same bug class as the 9 endpoints this PR fixes, just left in place under the exception label.

It also doesn't pick up the basic-auth/dev-mode bypass fix, since it calls `get_rpc_filter_context` directly rather than `get_scoped_resource_access_context` (which checks `_has_verified_jwt_payload` for the non-JWT fallback).

Worth noting: `requester_email` a few lines below is already derived independently via `get_user_email(user)`, not from `resource_user_email` — so the "needs two separate email values" justification doesn't seem to hold up here. If that's right, this collapses to:

```python
resource_user_email, token_teams = get_scoped_resource_access_context(request, user)
```

Happy to be wrong if there's a reason the two need to diverge that I'm missing — but if not, this brings the exception count down to 5 and closes the gap in the same pass.

## Suggestions (non-blocking)

- The PR description says the 5 untested-but-claimed-covered endpoints (`tools/list`, `resources/read`, `prompts/get`, internal `completion/complete`) "share the same `BaseService._apply_visibility_scope` enforcement" as the 4 tested ones. That's true for `prompts/get` and `completion/complete`, but `tools/list` (`tool_service.list_server_mcp_tool_definitions`) is actually a separate hand-rolled `access_conditions` builder that never calls `_apply_visibility_scope` — it happens to satisfy the specific fix here, but could drift independently later (same risk this PR exists to eliminate, just one layer down). Might be worth a follow-up issue rather than blocking this one.
- No integration-level test for the basic-auth/dev-mode admin bypass path — today it's only covered at the unit level with a mocked request. Given it's the wider-reaching of the two behavior changes (affects a whole auth mode), an end-to-end test through the real basic-auth dependency would be reassuring.
- A test for `create_mcp_app_session`'s admin-bypass path would have caught the issue above — worth adding alongside the fix.

Everything else looks solid — the deny-path invariants are intact, no schema/Alembic changes needed (correctly so), and the docstring corrections are accurate.

madhu-mohan-jaishankar:
> Nice cleanup — the mock retargeting across `test_main.py`/`test_main_extended.py` is thorough (verified every remaining raw `get_rpc_filter_context` patch actually maps to one of the 6 documented exception sites), the new `test_auth_context.py` file finally pins the Layer-1 rule directly instead of relying on endpoint mocks, and the memoization + malformed-payload fixes are well-targeted, well-tested additions that head off problems the refactor itself would otherwise introduce.
> 
> One thing I think is worth fixing before merge, and a couple of smaller notes.
> 
> ## Blocking
> **`create_mcp_app_session` still has the bug this PR fixes everywhere else** (`mcpgateway/main.py:10464`). It's listed as Layer-1 exception #4 ("keeps the resource-scope email separate from `user_email`, which the single-value helper contract cannot express"), but the code underneath still nulls the email on admin bypass:
> 
> ```python
> user_email, token_teams, is_admin = get_rpc_filter_context(request, user)
> if is_admin and token_teams is None:
>     resource_user_email = None
> else:
>     resource_user_email = user_email
>     if token_teams is None:
>         token_teams = []
> ```
> 
> That flows into `resource_service.read_resource(..., user=resource_user_email, token_teams=token_teams)`, which hits `ResourceService._check_resource_access`:
> 
> ```python
> if token_teams is None and user_email is None:
>     return visibility != "private"
> ```
> 
> So an admin creating an AppBridge session against their **own private** UI resource gets `user=None, token_teams=None` → the private-resource check fails → `ResourceNotFoundError` → 404, even though they own it. Same bug class as the 9 endpoints this PR fixes, just left in place under the exception label.
> 
> It also doesn't pick up the basic-auth/dev-mode bypass fix, since it calls `get_rpc_filter_context` directly rather than `get_scoped_resource_access_context` (which checks `_has_verified_jwt_payload` for the non-JWT fallback).
> 
> Worth noting: `requester_email` a few lines below is already derived independently via `get_user_email(user)`, not from `resource_user_email` — so the "needs two separate email values" justification doesn't seem to hold up here. If that's right, this collapses to:
> 
> ```python
> resource_user_email, token_teams = get_scoped_resource_access_context(request, user)
> ```
> 
> Happy to be wrong if there's a reason the two need to diverge that I'm missing — but if not, this brings the exception count down to 5 and closes the gap in the same pass.
> 
> ## Suggestions (non-blocking)
> * The PR description says the 5 untested-but-claimed-covered endpoints (`tools/list`, `resources/read`, `prompts/get`, internal `completion/complete`) "share the same `BaseService._apply_visibility_scope` enforcement" as the 4 tested ones. That's true for `prompts/get` and `completion/complete`, but `tools/list` (`tool_service.list_server_mcp_tool_definitions`) is actually a separate hand-rolled `access_conditions` builder that never calls `_apply_visibility_scope` — it happens to satisfy the specific fix here, but could drift independently later (same risk this PR exists to eliminate, just one layer down). Might be worth a follow-up issue rather than blocking this one.
> * No integration-level test for the basic-auth/dev-mode admin bypass path — today it's only covered at the unit level with a mocked request. Given it's the wider-reaching of the two behavior changes (affects a whole auth mode), an end-to-end test through the real basic-auth dependency would be reassuring.
> * A test for `create_mcp_app_session`'s admin-bypass path would have caught the issue above — worth adding alongside the fix.
> 
> Everything else looks solid — the deny-path invariants are intact, no schema/Alembic changes needed (correctly so), and the docstring corrections are accurate.

Latest round is addressed — head is `67078b65a`.

**`create_mcp_app_session`** — confirmed and fixed. 

**On scope, for the record.** That bug is pre-existing — `origin/main` is byte-identical at that block, so this PR neither introduced nor worsened it — and it sits in a different feature area from the visibility endpoints this issue targets. Issue #4451 exists specifically because PR #4080 was reverted for, in its own words, *"Scope creep: the refactor was not the cause of the original SSO admin bug"* and *"a bug-fix bundled with the refactor — separating the two would be cleaner."* Fixing it here runs against that charter, and it added a Breaking Changes surface plus a test to a PR that was otherwise complete. I've done it rather than argue the point, but a follow-up issue would have been the cleaner call.

**On the review cycle.** Rounds 2 and 3 closed with *"nothing here is a correctness bug in the new code itself"* and *"nothing here is a correctness problem in the shipped code"*, and the stale-mock item was described as *"not a blocker in the 'something's broken' sense"* — each was filed as changes-requested, and the description item was raised as blocking across three separate rounds. Each cycle costs a rebase, a full suite run and a re-review wait. Some findings have been legitimate and the code is better for them; the ask is about batching rather than standard.

msureshkumar88 (dismissed):
## E2E Verification — Live BEFORE/AFTER Comparison

Ran real, non-mocked end-to-end verification of issue #4451 / PR #4654: two live uvicorn instances (SQLite DB each, real signed JWTs, real HTTP requests) — one at base commit `4a1e3eadb2` (BEFORE), one at PR head `853ca3d7b` (AFTER, this branch).

### Config used (both servers, differing only in port/DB/secret)

```
HOST=127.0.0.1  PORT=8801|8802
DATABASE_URL=sqlite:///.../before.db | after.db
JWT_SECRET_KEY=<32+ char per-server secret>
AUTH_ENCRYPTION_SECRET=<32+ char per-server secret>
PLATFORM_ADMIN_PASSWORD=<test value>
BASIC_AUTH_USER=admin  BASIC_AUTH_PASSWORD=<test value>
ENVIRONMENT=development
AUTH_REQUIRED=true (Phase A) / false (Phase B)
ALLOW_UNAUTHENTICATED_ADMIN=false (Phase A) / true (Phase B)
MCPGATEWAY_UI_ENABLED=false  MCPGATEWAY_ADMIN_API_ENABLED=false
PLUGINS_ENABLED=false  MCPGATEWAY_A2A_ENABLED=true
MCPGATEWAY_MCP_APPS_ENABLED=false
LEGACY_API_ENABLED=true  REQUIRE_USER_IN_DB=true
MCP_CLIENT_AUTH_ENABLED=true  CACHE_TYPE=memory
LOG_LEVEL=INFO  LOG_TO_FILE=false
OBSERVABILITY_ENABLED=false  STRUCTURED_LOGGING_DATABASE_ENABLED=false
```

### Commands run

```
git worktree add <scratch>/before-worktree 4a1e3eadb272d06f5d23b08883fc029728ecc860
.venv/bin/python3 -m uvicorn mcpgateway.main:app --host 127.0.0.1 --port 8801   # BEFORE, from worktree
.venv/bin/python3 -m uvicorn mcpgateway.main:app --host 127.0.0.1 --port 8802   # AFTER, from PR branch checkout
.venv/bin/python3 -m mcpgateway.utils.create_jwt_token -u admin@example.com --admin --exp 60 --secret <...>            # admin bypass
.venv/bin/python3 -m mcpgateway.utils.create_jwt_token -u admin@example.com --admin --teams <team_id> --exp 60 ...     # admin narrowed-to-team
.venv/bin/python3 -m mcpgateway.utils.create_jwt_token -u other-user@example.com --teams <team_id> --exp 60 ...        # team-scoped non-admin
.venv/bin/python3 -m mcpgateway.utils.create_jwt_token -u other-user@example.com --exp 60 ...                         # public-only non-admin
# seed via HTTP as admin: POST /auth/email/admin/users, POST /teams/, POST /rbac/users/{email}/roles,
# POST /teams/{id}/members, POST /tools/ (private/team/public + other-user-owned private), POST /resources/
curl -H "Authorization: Bearer <jwt>" http://127.0.0.1:<port>/tags?include_entities=true
curl -H "Authorization: Bearer <jwt>" http://127.0.0.1:<port>/tags/{tag}/entities
curl -X POST -H "Authorization: Bearer <jwt>" http://127.0.0.1:<port>/rpc -d '{"jsonrpc":"2.0","method":"completion/complete", ...}'
curl -X POST -H "Authorization: Bearer <jwt>" http://127.0.0.1:<port>/protocol/completion/complete -d '{...}'
curl http://127.0.0.1:<port>/tags/{tag}/entities   # Phase B, zero Authorization header, dev-mode admin bypass
curl -H "Authorization: Bearer <team-scoped-jwt>" http://127.0.0.1:<port>/tools|/resources|/prompts|/a2a
git worktree remove <scratch>/before-worktree
```

### Results — 16/16 PASS

| # | Case | BEFORE | AFTER | Verdict |
|---|---|---|---|---|
| 1 | `GET /tags` — JWT admin bypass sees own private tag | HIDDEN | VISIBLE | PASS |
| 2 | `GET /tags/{tag}/entities` — same | HIDDEN (`[]`) | VISIBLE | PASS |
| 3 | JSON-RPC `completion/complete` via `/rpc`, admin's own private resource | HIDDEN (`total:0`) | VISIBLE (`total:1`) | PASS |
| 4 | REST `POST /protocol/completion/complete`, same resource | VISIBLE both (pre-existing handler never nulled email) | VISIBLE | PASS — confirms the PR also fixes a REST-vs-RPC inconsistency for the *same* logical op |
| 5 | Admin-bypass JWT must not see the *other* user's private tool (`/tags`, `/tags/{tag}/entities`) | PROTECTED | PROTECTED | PASS |
| 6 | Admin JWT narrowed to a team must not see other user's private tool | PROTECTED | PROTECTED | PASS |
| 7 | Public-only non-admin JWT | 403 (RBAC, both versions) | 403 | PASS |
| 8 | Non-JWT dev-mode admin must not see other user's private tool | HIDDEN | HIDDEN | PASS |
| 9 | Team-scoped non-admin smoke pass on `/tools`,`/resources`,`/prompts`,`/a2a` | identical, correct | identical, correct | PASS |
| 10 | Non-JWT/dev-mode admin bypass widening — sees own private tag entity | HIDDEN | VISIBLE | PASS |
| 11 | Same, must not see other user's private | HIDDEN | HIDDEN | PASS |

No tracebacks / 500s / unexpected errors in either server's log across both phases. Deny-path (team scoping, other-user privacy, public-only RBAC) is byte-for-byte identical and correct on both versions — no new leakage introduced by the widened bypass.

### Not exercised (disclosed, not faked)

- **5 `/_internal/mcp/*` endpoints** (`tools/list`, `resources/list`, `resources/read`, `prompts/list`, `prompts/get`): gated by `_authorize_internal_mcp_request`, which requires a loopback-only HMAC/auth-context header contract meant for the Rust MCP runtime, not an externally-presentable credential. Covered only by the existing (already-passing) unit-level test suite, not exercised live here.
- **`POST /appbridge/sessions`**: requires a real MCP session established via a full StreamableHTTP `initialize` handshake before `_assert_session_owner_or_admin` will proceed past a 404 — standing up a real MCP client handshake was out of scope for this harness.

### Verdict

Live behavior matches every claim in the PR description and CHANGELOG. Approving.

*(Two non-blocking documentation nits from static review — PR body's "preserved exceptions" table and "9 endpoints" count are stale vs the final commit/CHANGELOG — posted separately as a follow-up comment; they don't affect correctness.)*

msureshkumar88:
## Summary of suggested changes (non-blocking)

Static review + live e2e both check out (see approval review above) — nothing here blocks merge, just two doc-accuracy nits worth a quick pass before landing:

1. **"Preserved exceptions" table** lists `create_mcp_app_session` as untouched, but the final commit (`fix(mcp-apps): keep admin email for owner matching on AppBridge session creation`) migrated it to `get_scoped_resource_access_context()` too. Only 5 direct `get_rpc_filter_context` call sites remain in `main.py`, not 6.
2. **"Behaviour changes #2" list** says 9 endpoints; `CHANGELOG.md` correctly says 10 (includes `POST /appbridge/sessions`, where the bug manifested as a hard 404 rather than a shrunk list). Worth syncing the PR description to match the CHANGELOG entry.

Optional, not required: `create_mcp_app_session`'s admin-bypass fix currently only has mock-level unit coverage (`test_mcp_apps_security.py`), unlike the other 9 sites which get real-DB integration coverage in `test_admin_bypass_owner_visibility.py`. Not blocking since live e2e verification (above) exercised the equivalent JWT-admin-owner-match path directly and it holds correctly — just flagging as a nice-to-have for defense in depth.

ja8zyjits (approved):
LGTM

PR #4674, opened 2026-05-08
test(rate-limiter): add reproducer for dynamic-binding config-flow investigation

### LINKED ISSUE #4557: [FEATURE]: Follow up proposal after session affinity performnce impact investigations.
Follow up for https://github.com/IBM/mcp-context-forge/issues/4352

### 🧭 Type of Feature

### 🧭 Epic

https://github.com/IBM/mcp-context-forge/issues/4352

**Title:** Replace redis request-reply communication between workers with request-reply between coordinator and workers
**Goal:** Make load balancing with sticky sessions possibe, improve response time.
**Why now:** The current session affinity module reduces responsiveness and makes sicky sessions load balancing impossible.

```gherkin
Scenario: Request comes to gunicorn worker 
  Worker opens stream to upstream server 
  Response comes to the client

Scenario: Second request comes to different worker
  The worker receiver opens redis pub/sub session with session owner worker redis in request-replay scenario
  The worker receiver gets response from session owner worker 
  Response comes to the client from the worker receiver
```

As a result the second scenario resonce time is about two times longer, and resource allocation in increased
Observed results on typical laptop with Intel U9 185, 16 cores , 32GB ram, 3 gunicorn coordinators + 72workers
Before session affinity was introduced 180rps `make benchmark-mcp-tools`
After session affinity was introduced 7-9 rps `make benchmark-mcp-tools`
The results may be imporved with single gunicorn + 24 workers up to 100 rps which still slower. 

### 🧑🏻‍💻 User Story 1

**As a:** Performance investigator
**I want:** Propose to replace gunicorn workers model with coordinator - workers model without shared sockets
**So that:** This improves the double redirect of functionality of session affinity solution 

#### ✅ Acceptance Criteria
```gherkin
Scenario: Request comes to coordinator
  Coordinator opens stream to upstream server
  Coordinator delegates work to workers in request-reply scenario with MQ or FAAS or some alternative.
  Response comes from coordiantor
  Coordinators behind load balancer with sticky session (current nginx with configuration changes)
```

### 📐 Design Sketch

Include a diagram, sketch, or flow (use [Mermaid](https://mermaid.js.org/) if desired):

```mermaid
graph TD
    %% TOP BOX
    subgraph UP [Upstream Layer]
        Upstream[Upstream MCP Servers]
    end

    %% MIDDLE BOX (The Side-Processing)
    subgraph MQ [Processing Layer / MQ Broker]
        direction LR
        Pre[Pre-Process Plugins]
        Post[Post-Process Plugins]
    end

    %% BOTTOM BOX
    subgraph WS [Web Server / Entry Point]
        Server[API Gateway Web Server]
    end

    %% THE FLOW (Orchestrated by Web Server)
    %% Outbound
    Server -->|1. Request-Reply| Pre
    Pre -.->|2. OK/Blocked| Server
    Server -->|3. Final Call| Upstream

    %% Inbound
    Upstream -->|4. Raw Response| Server
    Server -->|5. Request-Reply| Post
    Post -.->|6. Final Result| Server
    Server -->|7. Send to Client| Client((User))

    %% STYLING
    style UP fill:#f5f5f5,stroke:#333
    style MQ fill:#e1f5fe,stroke:#01579b,stroke-dasharray: 5 5
    style WS fill:#f9f,stroke:#333,stroke-width:2px
```

This approach is already using MQ communication between two workers, so the idea move to 
This MQ like communikations one level up, this makes session affinity simpler.

---

### 🔗 MCP Standards Check

- [x] Change adheres to current [MCP specifications](https://developer.ibm.com/tutorials/awb-handle-remote-tool-calling-model-context-protocol/)

---

### 🔄 Alternatives Considered

The request-reply scenario proposals: redis, zeromq (single container), NATS, FAAS: Fission, Knative, Nuclio, microservices.

One of variants of possible IPC solution is zenoh https://github.com/eclipse-zenoh/zenoh/ , which is called "zero" overhead (5 bytes)  protocol inside the non-restrictive networks it allows p2p communications without specifying the peer addresses (in one container). 
The docker network between containers does not allow this type of discovery though and some brokers are required.
But these brokers are not participating in data transfer they serve only as peer discovery purpose.

---

### 📓 Additional Context

Include related issues, links to discussions, issues, etc.

### LINKED ISSUE #4665: [CHORE]: Investigate dynamic plugin-binding flow for RateLimiterPlugin and harden its config / payload story
### 📝 Summary

The dynamic plugin-binding API (`POST /v1/tools/plugin_bindings`) lets operators configure per-(team, tool, plugin) overrides at runtime — `mode`, `priority`, and the plugin-specific `config` dict — with the per-tenant plugin manager merging those overrides into the plugin's runtime instance.

This issue scopes an investigation into how that mechanism works end-to-end for `RateLimiterPlugin` specifically: confirm whether bindings reach the runtime as designed, identify any gaps, and decide on a path forward.

Companion to that, harden the rate-limiter's config surface — both the static `plugins/config.yaml` shape and the binding payload — which today is fairly minimal and hasn't had a focused hardening pass.

### 🤔 Why this matters

Downstream consumers configure rate limiting via the binding API to apply per-tenant / per-tool tighter limits. If anything in the binding → runtime path doesn't behave as designed, those overrides may not produce the configured outcome — and the failure is likely silent (the plugin runs, no error surfaces, but the runtime applies whatever the static `plugins/config.yaml` says rather than what the binding asks for).

The investigation aims to confirm the path works as designed end-to-end, or — if it doesn't — surface what would need to change.

Rate limiter is the only stateful plugin we ship — counter state lives in Redis, the plugin owns its own Redis connection lifecycle, it builds tenant-aware keys from `GlobalContext` at hook-call time, and mode transitions have side effects on external state (per the wipe-on-disable work in #4576). Those traits combine to make the binding-flow contract more entangled here than for most plugins, which is why it's worth a focused look rather than being assumed-by-symmetry from the rest of the framework.

### 🔍 Investigation areas

Concrete things worth chasing:

1. **Does `_merge_tenant_config` (`mcpgateway/plugins/gateway_plugin_manager.py:220`) produce the right merged Pydantic config when a binding overrides any rate-limiter config key?** Trace the merged-config dict from the binding row through the per-tenant manager and confirm the rate-limiter receives it.
2. **Is the per-tenant plugin manager actually being instantiated and used for tool-call paths that target a bound (team, tool) combination?** Or could the `__global__` manager's plugin instance be running, with the per-tenant override never applied?
3. **Is there any class-level / module-level state inside the rate-limiter's Rust core (`cpex_rate_limiter`) that would prevent a new instance from picking up merged config?** A singleton core or a shared client pool keyed by URL would explain config not propagating per tenant.
4. **What's the code path from the merged config (Pydantic) → `__init__` → Rust core construction?** Compare against what the binding-API merge logic produces.

### 📐 Possible fix shapes (open, scoping discussion)

Three high-level directions, depending on what the investigation surfaces:

1. **If the per-tenant manager isn't rebuilding the plugin instance correctly** — fix the manager rebuild path so the plugin is re-instantiated with the merged config per tenant.
2. **If the rate-limiter Rust core has class-level / module-level state** — make it instance-level so `__init__` constructs fresh state per merged-config instance.
3. **If `_merge_tenant_config` itself isn't being invoked on the tool-call path for a bound (team, tool)** — trace and fix the missing call site.

The investigation should land on which of these (or something else) applies before scoping the actual change.

### 🛡️ Hardening scope (config / payload)

Companion work, ideally in the same arc:

1. **Audit which rate-limiter config keys are operator-scoped vs caller-scoped.** Keys like `redis_url`, `redis_key_prefix`, `backend`, `algorithm` are gateway-scoped — they don't really make sense per binding. Keys like `by_user`, `by_tenant`, `by_tool`, `fail_mode`, `mode` are genuinely caller-scoped. The binding API today accepts everything verbatim with no distinction.
2. **Decide and document the contract**: which keys can be set per binding, which can't, and what the gateway does when an operator-scoped key shows up in a binding payload (silently drop / reject / warn).
3. **Tighten the static-config schema** if needed: today the rate-limiter's config shape is loosely typed (the cpex framework refactor in #4286 removed CF-side schemas). A simple in-cpex schema check that surfaces unknown or invalid keys at gateway boot would catch misconfiguration earlier rather than at first traffic.

### ✅ Acceptance criteria

- Root cause for the dynamic-binding behaviour (or confirmation the path works as designed) documented
- If a fix is needed: proposed, scoped, and landed; regression test in `tests/integration/test_rate_limiter_plugin_bindings_e2e.py` (or equivalent) asserts a binding's `by_user` / `by_tenant` / `by_tool` actually produces the configured rate-limiting end-to-end through the gateway HTTP path
- Documented contract for which config keys are caller-overridable vs operator-only on the rate-limiter binding payload
- Static-config schema validation surfaces unknown / invalid keys at gateway boot rather than runtime

### 📓 Related

- #4068 — introduced the per-tenant binding API + `TenantPluginManager`
- #4286 — removed plugin config schemas (`config: Dict[str, Any]`); the merge logic at `mcpgateway/plugins/gateway_plugin_manager.py:220` is the relevant code path
- #4622 — separate issue: binding payload hardening (don't accept infrastructure-layer keys like `redis_url` in caller bindings); framework-wide and orthogonal to the rate-limiter-specific hardening scope above

gandhipratik203:
## Summary

DO NOT MERGE YET.

Adds end-to-end tests that exercise `POST /v1/tools/plugin_bindings` against the gateway HTTP path and validate whether per-tenant binding overrides actually reach the rate-limiter at runtime. Captured as reproducers for the investigation tracked in #4665.

Three lifecycle test methods, each following the same inspection-friendly shape (POST / UPSERT / DELETE walk of the binding's full path through API, Postgres, plugin manager, and Redis):

| Method | Class | Lifecycle phase |
|---|---|---|
| `test_binding_full_lifecycle_inspectable` | `TestRateLimiterBindingApiEnforcesLimits` | POST happy path — binding is created, persisted, and enforces at the binding's `by_user: 7/m` limit at dispatch |
| `test_upsert_binding_full_lifecycle_inspectable` | `TestRateLimiterBindingModeAndLifecycle` | UPSERT — second POST with the same `(team, tool, plugin)` triple updates the existing row in place AND propagates the new mode past the DB into the plugin manager (enforce → disabled flip verified at dispatch) |
| `test_delete_binding_full_lifecycle_inspectable` | `TestRateLimiterBindingModeAndLifecycle` | DELETE — binding removal is verified on every write surface it landed on (API + Postgres) |

Shared helpers used by all three: MCP session handshake (from `tests/helpers/mcp_session.py`), tool team-id stamping via `docker exec psql` (`_stamp_tool_team_id`, fixture-driven), API + DB binding fetch (`_get_binding_via_api`, `_psql_get_binding_config`), live runtime plugin state lookup (`_get_admin_plugin_state`), Redis snapshot (`_redis_rl_keys`).

## Empirical finding from running against current `main`

Binding payloads that include **gateway-scoped keys** (`redis_url`, `redis_key_prefix`) silently lose their **caller-scoped overrides** (`by_user`, `by_tenant`, `by_tool`) at runtime. Payloads that omit those keys have their overrides applied cleanly.

The deterministic signal is the Redis key window suffix:

| Binding payload contents | Redis key written | Interpretation |
|---|---|---|
| Includes `redis_url` + `redis_key_prefix` (alongside `by_user: "5/s"`) | `rl:<tenant>:user:admin@example.com:60` | `:60` window → static `by_user: 30/m` is enforcing — binding's overrides silently dropped |
| Omits `redis_url` + `redis_key_prefix` (just `by_user: "5/s"`) | `rl:<tenant>:user:admin@example.com:1` | `:1` window → binding's `by_user: 5/s` is enforcing — overrides reach the runtime |

Captured via `redis-cli MONITOR` over multiple runs; reproducible.

## How to run

Prerequisites:

- Full docker-compose stack up (gateway + nginx + postgres + redis + fast_time_server).
- Default container names assume the `mcp-context-forge` compose project; override `RATE_LIMITER_TEST_PG_CONTAINER` + `REDIS_CONTAINER_NAME` for custom project names.
- `docker exec` access to the postgres + redis containers.

Run all three:

```bash
RUN_BINDING_LIFECYCLE=1 INSPECT=1 \
    pytest tests/live_gateway/plugins/test_rate_limiter_bindings_lifecycle.py -v -s
```

Run a single test (per-test commands are also embedded with the run output below):

```bash
RUN_BINDING_LIFECYCLE=1 INSPECT=1 \
    pytest tests/live_gateway/plugins/test_rate_limiter_bindings_lifecycle.py::TestRateLimiterBindingApiEnforcesLimits::test_binding_full_lifecycle_inspectable \
    -v -s
```

<details>
<summary><strong>Full env-var list + non-default compose project setup</strong></summary>

| Env var | Default | Purpose |
|---|---|---|
| `RUN_BINDING_LIFECYCLE` | (required: `1`) | Opt-in to running this file's tests (skipped by default) |
| `INSPECT` | (optional: `1`) | Enable per-call `[inspect]` narration; pair with `pytest -s` |
| `GATEWAY_URL` | `http://localhost:8080` | Gateway base URL |
| `GATEWAY_EMAIL` | `admin@example.com` | Admin email for `/auth/login` |
| `GATEWAY_PASSWORD` | `changeme` | Admin password |
| `BINDING_REDIS_URL` | `redis://redis:6379/0` | Gateway-side Redis URL (only used to assert the wrong-payload case in tests) |
| `RATE_LIMITER_TEST_TEAM_ID` | (optional) | Reuse an existing team for the test-tool's team_id stamp |
| `RATE_LIMITER_TEST_PG_CONTAINER` | `mcp-context-forge-postgres-1` | docker exec target for psql |
| `RATE_LIMITER_TEST_PG_USER` | `postgres` | psql user |
| `RATE_LIMITER_TEST_PG_DATABASE` | `mcp` | psql database |
| `REDIS_CONTAINER_NAME` | `mcp-context-forge-redis-1` | docker exec target for redis-cli |
| `PROPAGATION_WAIT` | (see `tests/helpers/integration_constants.py`) | Per-tenant plugin-manager rebuild wait, seconds |

For a non-default compose project (`docker compose -p rl-binding-test up …`):

```bash
RUN_BINDING_LIFECYCLE=1 INSPECT=1 \
    RATE_LIMITER_TEST_PG_CONTAINER=rl-binding-test-postgres-1 \
    REDIS_CONTAINER_NAME=rl-binding-test-redis-1 \
    pytest tests/live_gateway/plugins/test_rate_limiter_bindings_lifecycle.py -v -s
```

</details>

## Latest test results

Run on 2026-05-22 against the local docker-compose stack (3 gateway replicas × 24 gunicorn workers + Redis + Postgres). All 3 tests passed in 4:13.

| Test | Result | Key signals |
|---|---|---|
| `test_binding_full_lifecycle_inspectable` | ✅ PASS | 0 allowed / 5 blocked under binding's tight `7/m` (high-amplification run — binding tripped on first call); all 3 dimension keys (`user`, `tenant`, `tool`) present in Redis under binding's tenant prefix; DB cross-check confirms binding stored as POSTed |
| `test_upsert_binding_full_lifecycle_inspectable` | ✅ PASS | Initial enforce burst: 2 allowed / 3 blocked. UPSERT (enforce → disabled) confirmed via API + Postgres. Post-UPSERT disabled burst: 5/5 allowed — mode flip propagated past the DB into the plugin manager, dispatch skipped entirely |
| `test_delete_binding_full_lifecycle_inspectable` | ✅ PASS | 0/5 allowed → 5/5 blocked while binding was live; DELETE confirmed removed from both write surfaces (API + Postgres) |

Note: counter values reflect known session-affinity amplification on multi-replica gateways (~5×–20× ticks per user-level call; see #4557) — the behavioural assertions only require `rate_limited >= 1` so this is accommodated.

<details>
<summary><strong>Full [inspect] output — test_binding_full_lifecycle_inspectable</strong></summary>

```bash
RUN_BINDING_LIFECYCLE=1 INSPECT=1 \
    pytest tests/live_gateway/plugins/test_rate_limiter_bindings_lifecycle.py::TestRateLimiterBindingApiEnforcesLimits::test_binding_full_lifecycle_inspectable \
    -v -s
```

```
[inspect] Phase 0 — capturing baseline plugin state from /admin/plugins

[inspect]   baseline mode = 'disabled'

[inspect]   baseline by_user (static) = None

[inspect]   → if you peek at Redis Insight now, there should be no rl:* keys yet

[inspect] Phase 1 — POSTing binding with by_user='7/m', by_tenant='9/m', by_tool={'fast-time-sse-get-system-time': '11/m'}

[inspect]   → POST /v1/tools/plugin_bindings/  (args wrapped into the payload by _post_binding):

[inspect]       {
  "team_id": "fca757388f7d44ffb60b32523623a110",
  "tool_name": "fast-time-sse-get-system-time",
  "mode": "enforce",
  "binding_reference_id": "rl-binding-inspect-bb524cdb",
  "config": {
    "algorithm": "fixed_window",
    "backend": "redis",
    "by_user": "7/m",
    "by_tenant": "9/m",
    "by_tool": {
      "fast-time-sse-get-system-time": "11/m"
    },
    "fail_mode": "open"
  }
}

[inspect]   ← POST response body:

[inspect]       {
  "bindings": [
    {
      "id": "21a01afb012e4c30a2c38a2a771dfb3f",
      "teamId": "fca757388f7d44ffb60b32523623a110",
      "toolName": "fast-time-sse-get-system-time",
      "pluginId": "RateLimiterPlugin",
      "mode": "enforce",
      "priority": 50,
      "config": {
        "algorithm": "fixed_window",
        "backend": "redis",
        "by_user": "7/m",
        "by_tenant": "9/m",
        "by_tool": {
          "fast-time-sse-get-system-time": "11/m"
        },
        "fail_mode": "open"
      },
      "onError": null,
      "bindingReferenceId": "rl-binding-inspect-bb524cdb",
 

[inspect]   binding_reference_id = rl-binding-inspect-bb524cdb

[inspect] Phase 2 — verifying the binding actually persisted (all 3 dimensions)

[inspect]   ✓ API confirms by_user='7/m', by_tenant='9/m', by_tool={'fast-time-sse-get-system-time': '11/m'}

[inspect]   ✓ Postgres confirms by_user='7/m', by_tenant='9/m', by_tool={'fast-time-sse-get-system-time': '11/m'}

[inspect]   ✓ binding stored cleanly in DB at row keyed by binding_reference_id=rl-binding-inspect-bb524cdb

[inspect] Phase 3 — sleeping 17s

[inspect]   → covers the per-tenant plugin manager rebuild

[inspect]   → also gives you a window to peek at Redis Insight (still no rl:* keys)

[inspect] Phase 4 — pacing 5 tool calls 3s apart so the counter increments are observable in Redis Insight

[inspect]   → watch the rl:<tenant>:user:admin@example.com:60 counter climb with each call (refresh between calls)

[inspect]   → call 1/5 POST /servers/a88e2c3f5d7b4a9e8f1c6d2e3b4a5f6e/mcp

[inspect]       headers: Mcp-Session-Id=939b6814b7944b0389824179084c9bc5, Authorization=Bearer ..., Accept=json/event-stream

[inspect]       body:    {"jsonrpc": "2.0", "id": "30a56962-a5c6-42ac-8b4e-636615b14672", "method": "tools/call", "params": {"name": "fast-time-sse-get-system-time", "arguments": {}}}

[inspect]   ← call 1/5 response: HTTP 429

[inspect]       body:    {"error":{"code":-32602,"message":"Plugin Violation: Rate limit exceeded","data":{"description":"Rate limit exceeded","details":{"limited":true,"remaining":0,"reset_in":60,"user_id":"admin@example.com","tenant_id":"fca757388f7d44ffb60b32523623a110","dimensions":{"violated":[{"limited":true,"remaining":0,"reset_in":60},{"limited":true,"remaining":0,"reset_in":60},{"limited":true,"remaining":0,"rese

[inspect]   call 1/5: BLOCKED (HTTP 429)

[inspect]       counters after call 1: user=24  tenant=24  tool=24

[inspect]       response says: remaining=0, reset_in=60s  → effective_limit ≈ counter(24) + remaining(0) = 24

[inspect]       static(30/m): would pass  | binding(7/m): would BLOCK  | observed: BLOCKED  → binding's tighter limit is what's enforcing ✓

[inspect]   → call 2/5 POST /servers/a88e2c3f5d7b4a9e8f1c6d2e3b4a5f6e/mcp

[inspect]       headers: Mcp-Session-Id=939b6814b7944b0389824179084c9bc5, Authorization=Bearer ..., Accept=json/event-stream

[inspect]       body:    {"jsonrpc": "2.0", "id": "23009928-7d42-444a-8cf8-83d6ecbc08ec", "method": "tools/call", "params": {"name": "fast-time-sse-get-system-time", "arguments": {}}}

[inspect]   ← call 2/5 response: HTTP 429

[inspect]       body:    {"error":{"code":-32602,"message":"Plugin Violation: Rate limit exceeded","data":{"description":"Rate limit exceeded","details":{"limited":true,"remaining":0,"reset_in":56,"user_id":"admin@example.com","tenant_id":"fca757388f7d44ffb60b32523623a110","dimensions":{"violated":[{"limited":true,"remaining":0,"reset_in":56},{"limited":true,"remaining":0,"reset_in":56},{"limited":true,"remaining":0,"rese

[inspect]   call 2/5: BLOCKED (HTTP 429)

[inspect]       counters after call 2: user=48  tenant=48  tool=48

[inspect]       response says: remaining=0, reset_in=56s  → effective_limit ≈ counter(48) + remaining(0) = 48

[inspect]       static(30/m): would BLOCK  | binding(7/m): would BLOCK  | observed: BLOCKED  → either limit could be enforcing (counter exceeds both)

[inspect]   → call 3/5 POST /servers/a88e2c3f5d7b4a9e8f1c6d2e3b4a5f6e/mcp

[inspect]       headers: Mcp-Session-Id=939b6814b7944b0389824179084c9bc5, Authorization=Bearer ..., Accept=json/event-stream

[inspect]       body:    {"jsonrpc": "2.0", "id": "4aaa9761-0131-4a55-b16b-c0cb155b04bf", "method": "tools/call", "params": {"name": "fast-time-sse-get-system-time", "arguments": {}}}

[inspect]   ← call 3/5 response: HTTP 429

[inspect]       body:    {"error":{"code":-32602,"message":"Plugin Violation: Rate limit exceeded","data":{"description":"Rate limit exceeded","details":{"limited":true,"remaining":0,"reset_in":53,"user_id":"admin@example.com","tenant_id":"fca757388f7d44ffb60b32523623a110","dimensions":{"violated":[{"limited":true,"remaining":0,"reset_in":53},{"limited":true,"remaining":0,"reset_in":53},{"limited":true,"remaining":0,"rese

[inspect]   call 3/5: BLOCKED (HTTP 429)

[inspect]       counters after call 3: user=72  tenant=72  tool=72

[inspect]       response says: remaining=0, reset_in=53s  → effective_limit ≈ counter(72) + remaining(0) = 72

[inspect]       static(30/m): would BLOCK  | binding(7/m): would BLOCK  | observed: BLOCKED  → either limit could be enforcing (counter exceeds both)

[inspect]   → call 4/5 POST /servers/a88e2c3f5d7b4a9e8f1c6d2e3b4a5f6e/mcp

[inspect]       headers: Mcp-Session-Id=939b6814b7944b0389824179084c9bc5, Authorization=Bearer ..., Accept=json/event-stream

[inspect]       body:    {"jsonrpc": "2.0", "id": "e39a0eae-c145-41f7-9a08-ec393afbbe34", "method": "tools/call", "params": {"name": "fast-time-sse-get-system-time", "arguments": {}}}

[inspect]   ← call 4/5 response: HTTP 429

[inspect]       body:    {"error":{"code":-32602,"message":"Plugin Violation: Rate limit exceeded","data":{"description":"Rate limit exceeded","details":{"limited":true,"remaining":0,"reset_in":49,"user_id":"admin@example.com","tenant_id":"fca757388f7d44ffb60b32523623a110","dimensions":{"violated":[{"limited":true,"remaining":0,"reset_in":49},{"limited":true,"remaining":0,"reset_in":49},{"limited":true,"remaining":0,"rese

[inspect]   call 4/5: BLOCKED (HTTP 429)

[inspect]       counters after call 4: user=73  tenant=73  tool=73

[inspect]       response says: remaining=0, reset_in=49s  → effective_limit ≈ counter(73) + remaining(0) = 73

[inspect]       static(30/m): would BLOCK  | binding(7/m): would BLOCK  | observed: BLOCKED  → either limit could be enforcing (counter exceeds both)

[inspect]   → call 5/5 POST /servers/a88e2c3f5d7b4a9e8f1c6d2e3b4a5f6e/mcp

[inspect]       headers: Mcp-Session-Id=939b6814b7944b0389824179084c9bc5, Authorization=Bearer ..., Accept=json/event-stream

[inspect]       body:    {"jsonrpc": "2.0", "id": "4d4ed64e-0b15-4af2-949c-e41b1930d34d", "method": "tools/call", "params": {"name": "fast-time-sse-get-system-time", "arguments": {}}}

[inspect]   ← call 5/5 response: HTTP 429

[inspect]       body:    {"error":{"code":-32602,"message":"Plugin Violation: Rate limit exceeded","data":{"description":"Rate limit exceeded","details":{"limited":true,"remaining":0,"reset_in":46,"user_id":"admin@example.com","tenant_id":"fca757388f7d44ffb60b32523623a110","dimensions":{"violated":[{"limited":true,"remaining":0,"reset_in":46},{"limited":true,"remaining":0,"reset_in":46},{"limited":true,"remaining":0,"rese

[inspect]   call 5/5: BLOCKED (HTTP 429)

[inspect]       counters after call 5: user=97  tenant=97  tool=97

[inspect]       response says: remaining=0, reset_in=46s  → effective_limit ≈ counter(97) + remaining(0) = 97

[inspect]       static(30/m): would BLOCK  | binding(7/m): would BLOCK  | observed: BLOCKED  → either limit could be enforcing (counter exceeds both)

[inspect]   summary: allowed=0 rate_limited=5 errors=0

[inspect] Phase 5 — current rl:* keys in Redis (these are what the plugin actually wrote)

[inspect]   rl:fca757388f7d44ffb60b32523623a110:user:admin@example.com:60 = 97  (ttl=45s)

[inspect]   rl:fca757388f7d44ffb60b32523623a110:tenant:fca757388f7d44ffb60b32523623a110:60 = 97  (ttl=45s)

[inspect]   rl:fca757388f7d44ffb60b32523623a110:tool:fast-time-sse-get-system-time:60 = 97  (ttl=45s)

[inspect]   → refresh Redis Insight in the next ~30s; the keys above should be visible there

[inspect]   → expected counter ranges:
      binding's 7/m active   → counter ~5-15  (small)
      static's 30/m active   → counter ~30-150 (much larger)

[inspect] Phase 6 — asserting the enforcement transition + multi-dim tracking

[inspect]   ✓ enforcement was strict from the first call: all 5 blocked (amplification high this run)

[inspect]   dimension keys observed: user=1, tenant=1, tool(fast-time-sse-get-system-time)=1

[inspect]   ✓ all three dimension keys present in Redis

[inspect]   ✓ binding's multi-dimensional config genuinely reached the runtime

[inspect]   ✓ DB cross-check (Phase 2) confirms the binding row carries the binding's exact values (7/m, 9/m, 11/m), not the static defaults

[inspect] Phase 7 — cleanup runs via the cleanup_bindings fixture on test exit
.
```

</details>

<details>
<summary><strong>Full [inspect] output — test_upsert_binding_full_lifecycle_inspectable</strong></summary>

```bash
RUN_BINDING_LIFECYCLE=1 INSPECT=1 \
    pytest tests/live_gateway/plugins/test_rate_limiter_bindings_lifecycle.py::TestRateLimiterBindingModeAndLifecycle::test_upsert_binding_full_lifecycle_inspectable \
    -v -s
```

```
[inspect] Phase 0 — capturing baseline plugin state from /admin/plugins

[inspect]   baseline mode = 'disabled'

[inspect]   → no rl-binding-upsert-inspect-* row in tool_plugin_bindings yet

[inspect] Phase 1 — POSTing binding (mode=enforce) with by_user='7/m', by_tenant='9/m', by_tool={'fast-time-sse-get-system-time': '11/m'}

[inspect]   binding_reference_id = rl-binding-upsert-inspect-cb8d5f1b

[inspect] Phase 2 — verifying the enforce binding actually persisted

[inspect]   ✓ API confirms binding row (mode='enforce', by_user='7/m')

[inspect]   ✓ Postgres confirms config column persisted

[inspect] Phase 3 — sleeping 17s

[inspect]   → covers the per-tenant plugin manager rebuild

[inspect] Phase 4 — pacing 5 tool calls 3s apart to confirm the enforce binding is live before we upsert it

[inspect]   enforce call 1/5: ALLOWED

[inspect]   enforce call 2/5: ALLOWED

[inspect]   enforce call 3/5: BLOCKED (HTTP 429)

[inspect]   enforce call 4/5: BLOCKED (HTTP 429)

[inspect]   enforce call 5/5: BLOCKED (HTTP 429)

[inspect]   summary: allowed=2 rate_limited=3 errors=0

[inspect] Phase 5 — inspecting Redis for binding-driven dimension keys

[inspect]   rl:fca757388f7d44ffb60b32523623a110:user:admin@example.com:60 = 74  (ttl=47s)

[inspect]   rl:fca757388f7d44ffb60b32523623a110:tenant:fca757388f7d44ffb60b32523623a110:60 = 74  (ttl=47s)

[inspect]   rl:fca757388f7d44ffb60b32523623a110:tool:fast-time-sse-get-system-time:60 = 74  (ttl=47s)

[inspect]   dimension keys observed: user=1, tenant=1, tool(fast-time-sse-get-system-time)=1

[inspect]   ✓ all three dimension keys present in Redis

[inspect] Phase 6 — UPSERTing same (team='fca757388f7d44ffb60b32523623a110', tool='fast-time-sse-get-system-time', plugin=RateLimiterPlugin) triple with mode=disabled

[inspect] Phase 7 — verifying the upsert updated the existing row in place

[inspect]   ✓ API confirms mode flipped: enforce='enforce' → disabled='disabled'

[inspect]   ✓ API confirms config dict preserved across upsert (only mode flipped)

[inspect]   ✓ Postgres confirms config column unchanged across upsert

[inspect] Phase 8 — sleeping 17s

[inspect]   → covers the per-tenant manager rebuild after the mode flip

[inspect] Phase 9 — pacing 5 tool calls 3s apart to confirm mode=disabled propagated past the DB into the plugin manager (and skips the plugin entirely, including gateway-wide limits)

[inspect]   disabled call 1/5: ALLOWED

[inspect]   disabled call 2/5: ALLOWED

[inspect]   disabled call 3/5: ALLOWED

[inspect]   disabled call 4/5: ALLOWED

[inspect]   disabled call 5/5: ALLOWED

[inspect]   summary: allowed=5 rate_limited=0 errors=0

[inspect]   ✓ 5/5 calls allowed after upsert — the new mode reached the plugin manager and skipped dispatch entirely

[inspect]   → refresh Redis Insight in the next ~30s; no new rl:* keys should appear from this disabled-binding burst
.
```

</details>

<details>
<summary><strong>Full [inspect] output — test_delete_binding_full_lifecycle_inspectable</strong></summary>

```bash
RUN_BINDING_LIFECYCLE=1 INSPECT=1 \
    pytest tests/live_gateway/plugins/test_rate_limiter_bindings_lifecycle.py::TestRateLimiterBindingModeAndLifecycle::test_delete_binding_full_lifecycle_inspectable \
    -v -s
```

```
[inspect] Phase 0 — capturing baseline plugin state from /admin/plugins

[inspect]   baseline mode = 'disabled'

[inspect]   → no rl-binding-delete-inspect-* row in tool_plugin_bindings yet

[inspect] Phase 1 — POSTing binding with by_user='7/m', by_tenant='9/m', by_tool={'fast-time-sse-get-system-time': '11/m'}

[inspect]   binding_reference_id = rl-binding-delete-inspect-336d3a0c

[inspect] Phase 2 — verifying the binding actually persisted (all 3 dimensions)

[inspect]   ✓ API confirms by_user='7/m', by_tenant='9/m', by_tool={'fast-time-sse-get-system-time': '11/m'}

[inspect]   ✓ Postgres confirms by_user='7/m', by_tenant='9/m', by_tool={'fast-time-sse-get-system-time': '11/m'}

[inspect] Phase 3 — sleeping 17s

[inspect]   → covers the per-tenant plugin manager rebuild

[inspect]   → also gives you a window to peek at Redis Insight (still no rl:* keys)

[inspect] Phase 4 — pacing 5 tool calls 3s apart to confirm the binding is live before we delete it

[inspect]   call 1/5: BLOCKED (HTTP 429)

[inspect]   call 2/5: BLOCKED (HTTP 429)

[inspect]   call 3/5: BLOCKED (HTTP 429)

[inspect]   call 4/5: BLOCKED (HTTP 429)

[inspect]   call 5/5: BLOCKED (HTTP 429)

[inspect]   summary: allowed=0 rate_limited=5 errors=0

[inspect] Phase 5 — inspecting Redis for binding-driven dimension keys

[inspect]   rl:fca757388f7d44ffb60b32523623a110:user:admin@example.com:60 = 97  (ttl=47s)

[inspect]   rl:fca757388f7d44ffb60b32523623a110:tenant:fca757388f7d44ffb60b32523623a110:60 = 97  (ttl=47s)

[inspect]   rl:fca757388f7d44ffb60b32523623a110:tool:fast-time-sse-get-system-time:60 = 97  (ttl=47s)

[inspect]   dimension keys observed: user=1, tenant=1, tool(fast-time-sse-get-system-time)=1

[inspect]   ✓ all three dimension keys present in Redis

[inspect] Phase 6 — DELETEing binding rl-binding-delete-inspect-336d3a0c

[inspect]   ✓ DELETE returned HTTP 200

[inspect]   → sleeping 7s for the deletion to propagate

[inspect] Phase 7 — verifying the binding is GONE from every write surface

[inspect]   ✓ API confirms binding is no longer present

[inspect]   ✓ Postgres confirms binding row is gone

[inspect]   ✓ delete-lifecycle: the binding was removed from both write surfaces it landed on

[inspect]   → refresh Redis Insight in the next ~30s if you want to watch the existing rl:* counters expire naturally
```

</details>

## What each test walks through (phase-by-phase)

<details>
<summary><strong>POST happy path — 7 phases</strong></summary>

1. **Baseline capture** — `GET /admin/plugins/RateLimiterPlugin` to log what's currently mounted in memory.
2. **POST binding** with deliberately uncommon, distinct values for all three dimensions: `by_user="7/m"`, `by_tenant="9/m"`, `by_tool={fast-time-sse-get-system-time: "11/m"}`. Each value is recognisable in API responses, DB rows, and Redis Insight.
3. **Persistence cross-check** — fetched both via `GET /v1/tools/plugin_bindings/?binding_reference_id=…` (API) **and** via direct `docker exec psql` against the `tool_plugin_bindings.config` JSON column (DB). Both must return the binding's exact values for all three dimensions.
4. **Propagation wait** — covers the per-tenant plugin-manager rebuild.
5. **Paced 5-call burst** — calls sent 3s apart so counter increments are observable in Redis Insight; each call's outcome (`ALLOWED` / `BLOCKED` / `ERROR`) is printed inline.
6. **Inspect Redis snapshot** — dumps all `rl:*` keys with values and TTLs.
7. **Behavioural assertion** — confirms `rate_limited >= 1` and that all three dimension counter keys (`:user:`, `:tenant:`, `:tool:`) are present, proving the binding's multi-dimensional merged config genuinely reached the plugin at runtime.

</details>

<details>
<summary><strong>UPSERT — 9 phases</strong></summary>

1-5 identical to POST happy path (creates the initial `mode=enforce` binding, verifies enforcement).
6. **UPSERT** the same `(team, tool, plugin)` triple with `mode=disabled`.
7. **Persistence cross-check** — API and Postgres both confirm the row was updated in place (no new row, mode flipped, config preserved).
8. **Propagation wait** — for the mode flip to reach the plugin manager.
9. **Disabled-mode burst** — 5 paced calls; assertion is `5/5 allowed` (plugin skipped entirely, no rate-limit counters touched).

</details>

<details>
<summary><strong>DELETE — 7 phases</strong></summary>

1-5 identical to POST happy path (creates binding, verifies enforcement to confirm it's live before deletion).
6. **DELETE** `/v1/tools/plugin_bindings/?binding_reference_id=…`.
7. **Deletion cross-check** — binding is verified absent on both write surfaces (API and Postgres). No behavioural burst after delete — the gateway-wide RateLimiter at `by_user: 30/m` plus session-affinity tool-path amplification makes a post-delete behavioural assertion environment-dependent; structural deletion check is the reliable contract.

</details>

## What you'll see in Redis Insight / DBeaver during a run

<details>
<summary><strong>Redis Insight — after Phase 4 (filter <code>rl:*</code>)</strong></summary>

```
rl:<tenant_uuid>:user:admin@example.com:60                          = ~50–100   (ttl 59s → 30s)
rl:<tenant_uuid>:tenant:<tenant_uuid>:60                            = ~50–100
rl:<tenant_uuid>:tool:fast-time-sse-get-system-time:60              = ~50–100
```

All three keys with the same tenant UUID. Exact counter value varies between runs due to amplification; the **presence of all three dimension keys** is the stable signal.

</details>

<details>
<summary><strong>DBeaver — <code>public.tool_plugin_bindings</code> row appearing in Phase 1</strong></summary>

```
Phase 1 (~5s in):    new row appears in public.tool_plugin_bindings
                     - binding_reference_id : rl-binding-inspect-<hash>
                     - plugin_id            : RateLimiterPlugin
                     - tool_name            : fast-time-sse-get-system-time
                     - mode                 : enforce
                     - config (JSON)        : {"algorithm":"fixed_window",
                                                "backend":"redis",
                                                "by_user":"7/m",
                                                "by_tenant":"9/m",
                                                "by_tool":{"fast-time-sse-get-system-time":"11/m"},
                                                "fail_mode":"open"}

Phase 7 (~67s in):   row deleted (cleanup_bindings fixture)
```

Crucially: the `config` JSON does **not** include `redis_url` or `redis_key_prefix` — they're absent because we deliberately omitted them from the binding payload (the #4665 finding).

</details>

## Caveats / known flakiness

- **Tool-path amplification varies between runs** (~5×–20× ticks per call, depending on cache state, framework internal call patterns, etc.; tracked in #4557). The lifecycle test's enforcement-transition pattern reflects this:
  - **High-amplification runs**: every call blocks on the very first plugin invocation (`0 allowed / 5 blocked`) — see the test 1 output above for an example
  - **Low-amplification runs**: 1–4 calls slip through before the binding's tight 7/m trips (`1–4 allowed / N blocked`) — see test 2's initial burst (`2/3`) above
  - Both shapes are valid; the assertion only requires `rate_limited >= 1`.
- **Per-tenant manager propagation** across the 3 gateway replicas via Redis pub/sub is not always instantaneous; occasionally one replica still serves the static config briefly after a binding is created. The `PROPAGATION_WAIT` (7s) plus the inspection pause (10s) usually covers it, but it's the same flake any binding test would hit, not specific to this one.

**Companion single-instance baseline.** The test on [`test/rate-limiter-single-instance-on-main`](https://github.com/IBM/mcp-context-forge/tree/test/rate-limiter-single-instance-on-main) runs the same binding lifecycle in a single-instance deployment with tight limits (`by_user: 3/m`, `by_tenant: 5/m`, `by_tool: 7/m`) and asserts exact 1:1 counter behaviour. It demonstrates that the binding API itself dispatches cleanly when session-affinity amplification is out of the picture; the variability above is specific to the multi-replica setup, not the API.

Run:

```bash
INSPECT=1 RUN_BINDING_SINGLE_INSTANCE=1 uv run pytest \
  tests/live_gateway/plugins/test_rate_limiter_binding_single_instance.py \
  -v -s --with-integration
```

<details>
<summary>Output from a recent run (passes in ~14s)</summary>

```
[inspect] Phase 0 — capturing baseline plugin state from /admin/plugins
[inspect]   baseline mode = 'disabled'
[inspect]   → expected mode: 'disabled' (binding's mode='enforce' will override per-tenant)

[inspect] Phase 1 — POSTing binding with by_user='3/m', by_tenant='5/m', by_tool={'fast-time-sse-get-system-time': '7/m'}
[inspect]   binding_reference_id = rl-single-inspect-f4c698f1

[inspect] Phase 2 — verifying the binding actually persisted (all 3 dimensions)
[inspect]   ✓ API confirms all 3 dimension values
[inspect]   ✓ Postgres confirms config column persisted with the binding's exact values

[inspect] Phase 3 — sleeping 7s for per-tenant manager rebuild

[inspect] Phase 4 — firing 5 tool calls. Expected: calls 1-3 ALLOWED, calls 4-5 BLOCKED at binding's by_user: 3/m
[inspect]   call 1/5: ALLOWED
[inspect]   call 2/5: ALLOWED
[inspect]   call 3/5: ALLOWED
[inspect]   call 4/5: BLOCKED (MCP isError, rate-limit)
[inspect]   call 5/5: BLOCKED (MCP isError, rate-limit)
[inspect]   summary: allowed=3  rate_limited=2  errors=0

[inspect] Phase 5 — current rl:{team_id}:* keys in Redis (what the plugin actually wrote)
[inspect]   rl:ab4aaf7b...:user:admin@example.com:60   = 5  (ttl=60s)
[inspect]   rl:ab4aaf7b...:tool:fast-time-...:60       = 5  (ttl=59s)
[inspect]   rl:ab4aaf7b...:tenant:ab4aaf7b...:60       = 5  (ttl=59s)

[inspect] Phase 6 — asserting EXACT enforcement (1:1 counter, blocks at binding limit)
[inspect]   ✓ allowed=3, rate_limited=2 (matches binding's by_user: 3/m exactly)
[inspect]   ✓ all 3 dimension counters at exactly 5 — 1:1 with user calls, no amplification

[inspect] Phase 7 — cleanup runs via cleanup_bindings fixture on test exit

============================== 1 passed in 14.31s ==============================
```

</details>

## Status

**Not planned to merge.** This reproducer stands as the empirical evidence backing the docs change in PR #4848 — the lifecycle tests + inspection narration are intentionally heavy for hand-driven verification, not as a CI regression signal.

**File location and gating.** The lifecycle tests live under `tests/live_gateway/plugins/test_rate_limiter_bindings_lifecycle.py` (not `tests/integration/`) — the diagnostic `[inspect]` narration and ~4-minute total runtime make them a poor fit for the integration suite's CI-friendly conventions, but a natural fit for `tests/live_gateway/`. Gated behind two env vars:

- `RUN_BINDING_LIFECYCLE=1` — required to opt in to running (default: skip)
- `INSPECT=1` — also enables the per-call `[inspect]` narration (default: silent)

Skipped by all automated runs — including `make test-live-gateway` — and only runs when an operator opts in explicitly.

## Related

- #4665 — investigation issue these reproducers are captured for (rate-limiter-specific binding-config-flow gap + payload hardening)
- #4848 — docs PR that this reproducer's findings support
- #4635 — companion PR refreshing the broader rate-limiter integration + load test suite for current main contracts (independent — this PR doesn't depend on it)
- #4857 — a CI-runnable companion to these tests was explored here but parked (gateway pub/sub propagation timing race needs separate investigation)

### DIFF
--- a/tests/integration/test_rate_limiter_plugin_bindings_e2e.py
+++ b/tests/integration/test_rate_limiter_plugin_bindings_e2e.py
@@ -1,0 +1,728 @@
+# -*- coding: utf-8 -*-
+"""Location: ./tests/integration/test_rate_limiter_plugin_bindings_e2e.py
+Copyright 2026
+SPDX-License-Identifier: Apache-2.0
+Authors: Pratik Gandhi
+
+End-to-end integration test for the rate-limiter via the plugin-bindings API.
+
+Configures a ``RateLimiterPlugin`` binding on a per-team / per-tool scope
+through the gateway's ``POST /v1/tools/plugin_bindings`` endpoint, waits for
+propagation, then invokes the bound tool repeatedly to confirm enforcement
+fires (HTTP 429 / MCP isError with rate-limit text).
+
+This sits between two existing test layers:
+
+  * ``tests/unit/mcpgateway/routers/test_tool_plugin_bindings.py`` exercises
+    the binding router in isolation; plugin invocation is mocked.
+  * ``tests/integration/test_rate_limiter.py`` (and cpex-plugins integration
+    tests) exercise the plugin against real Redis but instantiate it
+    directly, bypassing the gateway's binding API entirely.
+
+Neither covers the full path: binding API → gateway plugin manager → plugin
+instantiation with the binding's config → enforcement at tool dispatch. This
+file fills that gap.
+
+Requirements:
+    - Running gateway at $GATEWAY_URL (default ``http://localhost:8080``)
+    - Redis reachable from the gateway pod at $BINDING_REDIS_URL
+      (default ``redis://redis:6379/0`` — the docker-compose hostname)
+    - At least one server with a registered tool
+    - Admin user belongs to at least one team (true for the seeded
+      ``admin@example.com`` personal team)
+
+Usage:
+    uv run pytest tests/integration/test_rate_limiter_plugin_bindings_e2e.py \\
+        -v --with-integration
+"""
+
+# Standard
+import os
+import subprocess
+import time
+import uuid
+
+# Third-Party
+import pytest
+import requests
+
+from tests.helpers.integration_constants import PLUGIN_MODE_PROPAGATION_WAIT_SECONDS
+
+# ---------------------------------------------------------------------------
+# Configuration
+# ---------------------------------------------------------------------------
+
+GATEWAY_URL = os.environ.get("GATEWAY_URL", "http://localhost:8080")
+GATEWAY_EMAIL = os.environ.get("GATEWAY_EMAIL", "admin@example.com")
+GATEWAY_PASSWORD = os.environ.get("GATEWAY_PASSWORD", "changeme")
+
+# Redis URL from the gateway pod's perspective. Defaults to the docker-compose
+# service hostname; override when the gateway runs elsewhere.
+BINDING_REDIS_URL = os.environ.get("BINDING_REDIS_URL", "redis://redis:6379/0")
+
+# Optional team override — useful when the runner knows the team_id ahead of
+# time. If unset, the test discovers the admin's first team automatically.
+OVERRIDE_TEAM_ID = os.environ.get("RATE_LIMITER_TEST_TEAM_ID")
+
+# Postgres container name for the in-test team-id stamping below.
+# Tools registered via the docker-compose registration job are created with
+# ``team_id = NULL``; ``ToolService.invoke_tool`` falls back to ``server_id``
+# (no ``::`` separator) for the plugin context_id when the tool has no
+# team_id, which makes ``get_config_from_db`` skip the binding lookup
+# entirely. To exercise the binding path we therefore have to stamp a
+# team_id onto the test tool *after* the docker-compose registration runs.
+# There is no public API to set ``tools.team_id`` after creation
+# (``ToolUpdate`` schema in ``mcpgateway/schemas.py`` does not include it),
+# so we shell out to ``docker exec ... psql``. This couples the test to the
+# docker-compose dev stack — acceptable because the suite is already
+# skip-guarded on a running gateway via ``_is_gateway_running()``.
+PG_CONTAINER = os.environ.get(
+    "RATE_LIMITER_TEST_PG_CONTAINER", "rl-binding-test-postgres-1"
+)
+PG_USER = os.environ.get("RATE_LIMITER_TEST_PG_USER", "postgres")
+PG_DATABASE = os.environ.get("RATE_LIMITER_TEST_PG_DATABASE", "mcp")
+
+PROPAGATION_WAIT = int(
+    os.environ.get("PROPAGATION_WAIT", str(PLUGIN_MODE_PROPAGATION_WAIT_SECONDS))
+)
+
+# Burst size and configured limit — burst must clearly exceed the limit so
+# enforcement is observable even with mild clock jitter.
+BURST_SIZE = 15
+BURST_LIMIT_PER_SEC = 5
+
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+
+def _get_session_token() -> str:
+    """Login and return a session token."""
+    resp = requests.post(
+        f"{GATEWAY_URL}/auth/login",
+        json={"email": GATEWAY_EMAIL, "password": GATEWAY_PASSWORD},
+        timeout=10,
+    )
+    resp.raise_for_status()
+    return resp.json()["access_token"]
+
+
+def _fresh_headers() -> dict:
+    """Get fresh auth headers for an admin call."""
+    return {
+        "Authorization": f"Bearer {_get_session_token()}",
+        "Content-Type": "application/json",
+        "Accept": "application/json",
+    }
+
+
+def _is_gateway_running() -> bool:
+    """Return True if the gateway is reachable."""
+    try:
+        return requests.get(f"{GATEWAY_URL}/health", timeout=5).status_code == 200
+    except requests.ConnectionError:
+        return False
+
+
+def _create_test_team() -> str:
+    """Create a fresh non-personal team for the test and return its id.
+
+    Personal teams (the only kind the seeded admin user has out of the box)
+    are excluded by the admin-side ``/teams/`` listing path, so we can't
+    discover one to scope a binding to. The simplest reliable path is to
+    create a dedicated test team and clean it up at the end.
+
+    Honors $RATE_LIMITER_TEST_TEAM_ID — if set, that team_id is reused as
+    is and no team is created (caller owns lifecycle).
+    """
+    if OVERRIDE_TEAM_ID:
+        return OVERRIDE_TEAM_ID
+
+    headers = _fresh_headers()
+    resp = requests.post(
+        f"{GATEWAY_URL}/teams/",
+        json={
+            "name": f"rate-limiter-test-team-{uuid.uuid4().hex[:8]}",
+            "description": "Ephemeral team for rate-limiter binding e2e test",
+            "visibility": "private",
+        },
+        headers=headers,
+        timeout=10,
+    )
+    if resp.status_code not in (200, 201):
+        pytest.skip(
+            f"POST /teams/ returned {resp.status_code}; cannot create a test "
+            f"team. Body: {resp.text[:200]}. "
+            f"Set RATE_LIMITER_TEST_TEAM_ID to override."
+        )
+    return resp.json()["id"]
+
+
+def _delete_test_team(team_id: str) -> None:
+    """Delete the test team. Best-effort — skipped when the caller supplied
+    the team_id via env override (lifecycle owned by the caller)."""
+    if OVERRIDE_TEAM_ID:
+        return
+    try:
+        requests.delete(
+            f"{GATEWAY_URL}/teams/{team_id}",
+            headers=_fresh_headers(),
+            timeout=10,
+        )
+    except requests.RequestException:
+        pass
+
+
+def _auto_detect_server_and_tool() -> tuple[str, str]:
+    """Find a server ID and tool name to drive the test against."""
+    headers = _fresh_headers()
+    resp = requests.get(f"{GATEWAY_URL}/servers", headers=headers, timeout=10)
+    resp.raise_for_status()
+    for server in resp.json():
+        tools = server.get("associatedTools", [])
+        # Prefer echo (cheap, predictable), fall back to any time tool.
+        for tool in tools:
+            if "echo" in tool.lower():
+                return server["id"], tool
+        for tool in tools:
+            if "time" in tool.lower() and "convert" not in tool.lower():
+                return server["id"], tool
+    pytest.skip("No suitable server/tool found for plugin-bindings test")
+
+
+def _resolve_tool_id(tool_name: str) -> str:
+    """Look up the UUID for ``tool_name`` via ``GET /tools/``."""
+    headers = _fresh_headers()
+    resp = requests.get(f"{GATEWAY_URL}/tools/", headers=headers, timeout=10)
+    resp.raise_for_status()
+    for tool in resp.json():
+        if tool.get("name") == tool_name:
+            return tool["id"]
+    pytest.skip(f"Tool {tool_name!r} not found via /tools/")
+
+
+def _stamp_tool_team_id(tool_id: str, team_id: str) -> str:
+    """Force ``tools.team_id`` for ``tool_id`` via ``docker exec ... psql``.
+
+    Returns the previous team_id (may be ``None`` / empty string) so the
+    fixture can restore it on teardown.
+
+    Skips the test if the postgres container isn't reachable — better
+    than letting the actual assertions fail with a confusing "no
+    requests were rate-limited" message that hides the real cause.
+    """
+    # Read the current value so we can restore it on teardown.
+    select_sql = f"SELECT COALESCE(team_id, '') FROM tools WHERE id = '{tool_id}';"
+    cmd_select = [
+        "docker", "exec", PG_CONTAINER,
+        "psql", "-U", PG_USER, "-d", PG_DATABASE,
+        "-tAc", select_sql,
+    ]
+    try:
+        prev = subprocess.run(
+            cmd_select, capture_output=True, text=True, timeout=10, check=True
+        ).stdout.strip()
+    except (subprocess.CalledProcessError, subprocess.TimeoutExpired, FileNotFoundError) as exc:
+        pytest.skip(
+            f"Cannot reach postgres container {PG_CONTAINER!r} to stamp tool team_id "
+            f"({type(exc).__name__}: {exc}). Set RATE_LIMITER_TEST_PG_CONTAINER if "
+            f"the container has a different name in your environment."
+        )
+
+    update_sql = f"UPDATE tools SET team_id = '{team_id}' WHERE id = '{tool_id}';"
+    cmd_update = [
+        "docker", "exec", PG_CONTAINER,
+        "psql", "-U", PG_USER, "-d", PG_DATABASE,
+        "-c", update_sql,
+    ]
+    subprocess.run(cmd_update, capture_output=True, text=True, timeout=10, check=True)
+    return prev
+
+
+def _restore_tool_team_id(tool_id: str, prev_team_id: str) -> None:
+    """Restore ``tools.team_id`` after the module finishes. Best-effort."""
+    if prev_team_id:
+        sql = f"UPDATE tools SET team_id = '{prev_team_id}' WHERE id = '{tool_id}';"
+    else:
+        sql = f"UPDATE tools SET team_id = NULL WHERE id = '{tool_id}';"
+    cmd = [
+        "docker", "exec", PG_CONTAINER,
+        "psql", "-U", PG_USER, "-d", PG_DATABASE,
+        "-c", sql,
+    ]
+    try:
+        subprocess.run(cmd, capture_output=True, text=True, timeout=10, check=False)
+    except (subprocess.TimeoutExpired, FileNotFoundError):
+        pass  # cleanup is best-effort
+
+
+def _post_binding(
+    team_id: str,
+    tool_name: str,
+    config: dict,
+    mode: str,
+    binding_reference_id: str,
+) -> dict:
+    """POST a single rate-limiter binding via the API."""
+    headers = _fresh_headers()
+    payload = {
+        "teams": {
+            team_id: {
+                "policies": [
+                    {
+                        "tool_names": [tool_name],
+                        "plugin_id": "RateLimiterPlugin",
+                        "mode": mode,
+                        "priority": 50,
+                        "config": config,
+                        "binding_reference_id": binding_reference_id,
+                    }
+                ]
+            }
+        }
+    }
+    resp = requests.post(
+        f"{GATEWAY_URL}/v1/tools/plugin_bindings/",
+        json=payload,
+        headers=headers,
+        timeout=10,
+    )
+    resp.raise_for_status()
+    return resp.json()
+
+
+def _delete_binding_by_reference(binding_reference_id: str) -> None:
+    """Delete bindings created with the given reference id. Best-effort."""
+    try:
+        requests.delete(
+            f"{GATEWAY_URL}/v1/tools/plugin_bindings/",
+            params={"binding_reference_id": binding_reference_id},
+            headers=_fresh_headers(),
+            timeout=10,
+        )
+    except requests.RequestException:
+        pass  # cleanup best-effort; surface in test failure if it matters
+
+
+def _mcp_initialize_session(
+    server_id: str, headers: dict
+) -> str | None:
+    """Run the MCP streamable-HTTP initialize + initialized handshake.
+
+    Returns the gateway-issued ``mcp-session-id`` so subsequent tool calls
+    can be sent against the same session. Per-server plugin bindings are
+    scoped (team, server_id, tool) and the plugin manager resolves them
+    only on session-bound requests.
+
+    Returns ``None`` on any handshake failure; callers should treat that
+    as a transport error (not a rate-limit signal).
+    """
+    init_body = {
+        "jsonrpc": "2.0",
+        "id": "init",
+        "method": "initialize",
+        "params": {
+            "protocolVersion": "2025-06-18",
+            "capabilities": {},
+            "clientInfo": {"name": "rate-limiter-binding-e2e", "version": "0"},
+        },
+    }
+    sse_headers = {**headers, "Accept": "application/json, text/event-stream"}
+    try:
+        resp = requests.post(
+            f"{GATEWAY_URL}/servers/{server_id}/mcp",
+            json=init_body,
+            headers=sse_headers,
+            timeout=10,
+        )
+    except requests.RequestException:
+        return None
+    if resp.status_code != 200:
+        return None
+    session_id = resp.headers.get("mcp-session-id")
+    if not session_id:
+        return None
+    # Fire-and-forget initialized notification — gateway returns 202 and
+    # we don't need the body. A failure here would surface on the next
+    # tools/call as a session error, so we let that path handle it.
+    try:
+        requests.post(
+            f"{GATEWAY_URL}/servers/{server_id}/mcp",
+            json={"jsonrpc": "2.0", "method": "notifications/initialized"},
+            headers={**sse_headers, "Mcp-Session-Id": session_id},
+            timeout=5,
+        )
+    except requests.RequestException:
+        pass
+    return session_id
+
+
+def _send_tool_burst(server_id: str, tool_name: str, count: int) -> dict:
+    """Fire ``count`` JSON-RPC tool calls via ``/servers/<id>/mcp`` over a
+    single MCP session and tally allowed / rate-limited / errors.
+
+    Uses the per-server MCP route, not the session-less ``/rpc``: per-tenant
+    plugin bindings are scoped (team, server, tool) and only the session-aware
+    route carries the ``server_id`` context the plugin manager needs to
+    resolve binding overrides at request time.
+
+    The MCP initialize + initialized handshake runs once; the resulting
+    session id is reused for all ``count`` tool calls. Burst counters
+    accumulate against a single user identity so per-user rate limits
+    accumulate the way bindings expect.
+    """
+    counters = {"allowed": 0, "rate_limited": 0, "errors": 0, "total": count}
+    headers = _fresh_headers()
+
+    session_id = _mcp_initialize_session(server_id, headers)
+    if session_id is None:
+        counters["errors"] = count
+        return counters
+
+    call_headers = {
+        **headers,
+        "Accept": "application/json, text/event-stream",
+        "Mcp-Session-Id": session_id,
+    }
+
+    for i in range(count):
+        payload = {
+            "jsonrpc": "2.0",
+            "id": str(uuid.uuid4()),
+            "method": "tools/call",
+            "params": {
+                "name": tool_name,
+                "arguments": (
+                    {"message": f"binding-test-{i}"} if "echo" in tool_name else {}
+                ),
+            },
+        }
+        try:
+            resp = requests.post(
+                f"{GATEWAY_URL}/servers/{server_id}/mcp",
+                json=payload,
+                headers=call_headers,
+                timeout=15,
+            )
+            if resp.status_code == 429:
+                counters["rate_limited"] += 1
+                continue
+            if resp.status_code != 200:
+                counters["errors"] += 1
+                continue
+
+            data = resp.json()
+            result = data.get("result", {})
+            if result.get("isError"):
+                content = result.get("content", [])
+                text = content[0].get("text", "") if content else ""
+                if "rate" in text.lower() or "limit" in text.lower():
+                    counters["rate_limited"] += 1
+                else:
+                    counters["errors"] += 1
+            else:
+                counters["allowed"] += 1
+        except requests.RequestException:
+            counters["errors"] += 1
+
+    return counters
+
+
+# ---------------------------------------------------------------------------
+# Skip-guards
+# ---------------------------------------------------------------------------
+
+pytestmark = pytest.mark.skipif(
+    not _is_gateway_running(),
+    reason=f"Gateway not running at {GATEWAY_URL}",
+)
+
+
+# ---------------------------------------------------------------------------
+# Fixtures
+# ---------------------------------------------------------------------------
+
+
+@pytest.fixture(scope="module")
+def server_and_tool():
+    """Auto-detect server/tool once for the module."""
+    return _auto_detect_server_and_tool()
+
+
+@pytest.fixture(scope="module")
+def team_id(server_and_tool):
+    """Create (or reuse via $RATE_LIMITER_TEST_TEAM_ID) a team_id for the
+    module, stamp it onto the test tool's row so the plugin manager
+    actually resolves bindings against it, and tear both down at the end.
+
+    The stamp is the load-bearing piece: bindings are scoped per
+    (team, tool, plugin) and ``tool_service.invoke_tool`` reads
+    ``tool.team_id`` (not the calling user's team) when constructing
+    the plugin context_id. Tools registered by the docker-compose
+    bootstrap have ``team_id = NULL`` so the binding lookup is skipped
+    entirely. See ``_stamp_tool_team_id`` for the why.
+    """
+    _, tool_name = server_and_tool
+    tid = _create_test_team()
+    tool_id = _resolve_tool_id(tool_name)
+    prev_team_id = _stamp_tool_team_id(tool_id, tid)
+    try:
+        yield tid
+    finally:
+        _restore_tool_team_id(tool_id, prev_team_id)
+        _delete_test_team(tid)
+
+
+@pytest.fixture
+def cleanup_bindings():
+    """Track + clean up bindings by reference_id at the end of each test."""
+    created: list[str] = []
+    yield created
+    for ref_id in created:
+        _delete_binding_by_reference(ref_id)
+
+
+# ---------------------------------------------------------------------------
+# Tests
+# ---------------------------------------------------------------------------
+
+
+class TestRateLimiterBindingApiEnforcesLimits:
+    """A rate-limiter binding configured through the bindings API enforces
+    its limits at tool-dispatch time."""
+
+    def test_binding_with_tight_user_limit_blocks_burst(
+        self, server_and_tool, team_id, cleanup_bindings
+    ):
+        """A binding with ``by_user: 5/s`` rate-limits a 15-request burst.
+
+        Asserts:
+          - At least one request is rate-limited (proves the binding's
+            tighter limit is being applied — shipped config alone is 30/m
+            and would not block a 15-request burst).
+          - At least one request is allowed (proves the binding is not
+            outright failing the plugin and dropping every request).
+          - No transport / MCP errors that aren't rate-limit signals.
+        """
+        server_id, tool_name = server_and_tool
+        ref_id = f"rl-binding-test-{uuid.uuid4().hex[:8]}"
+        cleanup_bindings.append(ref_id)
+
+        _post_binding(
+            team_id=team_id,
+            tool_name=tool_name,
+            mode="enforce",
+            binding_reference_id=ref_id,
+            config={
+                "algorithm": "fixed_window",
+                "backend": "redis",
+                "by_user": f"{BURST_LIMIT_PER_SEC}/s",
+                "by_tenant": None,
+                "by_tool": {},
+                # redis_url + redis_key_prefix intentionally omitted — see the
+                # `_binding_config` docstring below: testing whether dropping
+                # gateway-scoped keys lets the per-tenant overrides propagate.
+                "fail_mode": "open",
+            },
+        )
+
+        # Wait for binding propagation — invalidate-and-broadcast plus any
+        # downstream cache TTLs.
+        time.sleep(PROPAGATION_WAIT)
+
+        result = _send_tool_burst(server_id, tool_name, BURST_SIZE)
+
+        assert result["rate_limited"] > 0, (
+            f"Binding configures by_user: {BURST_LIMIT_PER_SEC}/s and a burst of "
+            f"{BURST_SIZE} requests should exceed it, but no requests were "
+            f"rate-limited. Got: {result}"
+        )
+        assert result["allowed"] >= 1, (
+            f"At least the first few requests should be allowed before the "
+            f"limit kicks in. Got: {result}"
+        )
+        assert result["errors"] == 0, (
+            f"Non-rate-limit errors indicate a setup or transport problem, "
+            f"not enforcement behaviour. Got: {result}"
+        )
+
+
+class TestRateLimiterBindingModeAndLifecycle:
+    """The binding's ``mode`` field and lifecycle operations (upsert, delete)
+    propagate to the gateway plugin manager and change tool-dispatch behaviour.
+
+    Shares the same burst shape as the enforcement test above so the contrast
+    is unambiguous: a 15-request burst against a 5/s limit that *would* block
+    in enforce mode must NOT block in disabled / permissive / post-delete
+    states.
+    """
+
+    @staticmethod
+    def _binding_config(ref_id: str) -> dict:
+        """Return the standard rate-limiter config used across these tests.
+
+        Probe for #4665: gateway-scoped keys (``redis_url``,
+        ``redis_key_prefix``) intentionally omitted from the binding payload.
+        Both flow from the static ``plugins/config.yaml``. If the binding's
+        per-tenant ``by_user`` reaches the runtime now, it's evidence that
+        sending gateway-scoped keys in the binding's config dict was poisoning
+        the per-tenant merge for ``RateLimiterPlugin``.
+        """
+        return {
+            "algorithm": "fixed_window",
+            "backend": "redis",
+            "by_user": f"{BURST_LIMIT_PER_SEC}/s",
+            "by_tenant": None,
+            "by_tool": {},
+            "fail_mode": "open",
+        }
+
+    def test_disabled_mode_allows_burst_through(
+        self, server_and_tool, team_id, cleanup_bindings
+    ):
+        """A binding with ``mode: disabled`` does not enforce its limit.
+
+        Same tight 5/s by_user limit as the enforce test; with mode=disabled
+        the plugin's hook should short-circuit and let the full burst through.
+        Proves the binding-payload ``mode`` field reaches the plugin manager
+        and is honoured at dispatch.
+        """
+        server_id, tool_name = server_and_tool
+        ref_id = f"rl-binding-disabled-{uuid.uuid4().hex[:8]}"
+        cleanup_bindings.append(ref_id)
+
+        _post_binding(
+            team_id=team_id,
+            tool_name=tool_name,
+            mode="disabled",
+            binding_reference_id=ref_id,
+            config=self._binding_config(ref_id),
+        )
+        time.sleep(PROPAGATION_WAIT)
+
+        result = _send_tool_burst(server_id, tool_name, BURST_SIZE)
+
+        assert result["rate_limited"] == 0, (
+            f"mode=disabled must not enforce — no requests should have been "
+            f"rate-limited, got: {result}"
+        )
+        assert result["allowed"] == BURST_SIZE, (
+            f"All {BURST_SIZE} requests should pass when the binding is "
+            f"disabled. Got: {result}"
+        )
+        assert result["errors"] == 0, f"Unexpected transport errors: {result}"
+
+    def test_upsert_from_enforce_to_disabled_stops_blocking(
+        self, server_and_tool, team_id, cleanup_bindings
+    ):
+        """Upserting an existing binding from ``enforce`` to ``disabled``
+        propagates to the plugin manager and stops further blocking.
+
+        Sequence:
+          1. POST enforce binding, burst → some blocked (sanity).
+          2. POST same triple with mode=disabled (upsert).
+          3. Burst again → none blocked.
+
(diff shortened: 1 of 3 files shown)


msureshkumar88 (dismissed):
## Test run — all 3 pass ✅

Ran the full suite locally against the docker-compose dev stack (Postgres + 3-replica gateway + Redis) on branch `test/rate-limiter-dynamic-binding-repro`.

```
======================== 3 passed in 240.50s (0:04:00) =========================
```

| Test | Outcome | Key signal |
|---|---|---|
| `test_binding_full_lifecycle_inspectable` | ✅ PASS | 5/5 calls blocked (HTTP 429); all three Redis dimension keys present (`user`, `tenant`, `tool`) |
| `test_upsert_binding_full_lifecycle_inspectable` | ✅ PASS | Upsert to `mode: disabled` → 5/5 calls allowed, no `rl:*` keys written — dispatch correctly skipped |
| `test_delete_binding_full_lifecycle_inspectable` | ✅ PASS | DELETE removes binding from API + Postgres; Redis counters expire naturally |

The reproduce signals are solid — the inline `[inspect]` phase output confirmed each lifecycle step at the expected DB, Redis, and behavioural surfaces. Good addition to the investigation corpus.

msureshkumar88:
## Reliability analysis — potential flakiness sources

Ran all three tests cleanly (3 passed in 4:00). Did a read-through of the test internals to flag what's most likely to cause intermittent failures on CI or slower machines.

### 1. Fixed propagation wait with no poll loop (highest risk)

`PROPAGATION_WAIT (7s) + post_propagation_pause (10s) = 17s` flat sleep before the enforce burst. On a loaded CI runner — or if Redis pub/sub is momentarily slow across the 3-replica gateway — the binding might not have reached every replica yet. The first tool call then hits the static config and returns 200, and if amplification is also low that run, `rate_limited` stays 0.

Worth considering a poll loop instead:

```python
deadline = time.time() + 20  # max wait
while time.time() < deadline:
    if _get_binding_via_api(ref_id) is not None:
        break
    time.sleep(0.5)
```

Phase 2 already confirms the binding landed in the DB; a polling loop here confirms propagation to the plugin manager rather than assuming it happened in N seconds.

### 2. `_redis_rl_keys()` doesn't filter by current `team_id`

The function fetches all `rl:*` keys from Redis. The Phase 5/6 assertion checks `":user:" in key` — if stale `rl:*` keys from a previous test run are still alive (TTL 60s), they satisfy the `>= 1` assertion even if the current test's binding never reached the runtime.

Filtering the scan pattern to `rl:{team_id}:*` removes the cross-run contamination:

```python
pattern = f"rl:{team_id}:*"
```

### 3. `_redis_rl_keys()` silently swallows container errors

On docker exec failure (`TimeoutExpired`, `CalledProcessError`), the function returns `[]`. The subsequent `len(user_keys) >= 1` assertion then fails with "by_user counter is missing" — which looks like a real enforcement failure, not a test infrastructure failure.

A `pytest.skip(f"redis container {container!r} unreachable — …")` or at minimum a `warnings.warn` on empty result after a non-empty scan would make the failure mode legible.

### 4. Phase 5 → Phase 6 key re-read races against TTL

`post_burst_pause = 30s` sleep happens between Phase 5 (inspect) and Phase 6 (assert). Counter TTL is 60s. From first write (start of burst) to Phase 6 check: ~12s burst + 30s sleep = ~42s elapsed — leaves ~18s TTL. Fine on a fast runner; on a slow one where the burst itself takes longer, the margin shrinks. Filtering by `team_id` (point 2) would also give a tighter namespace to check.

### 5. N serial `docker exec redis-cli GET`/`TTL` calls per key

`_redis_rl_keys` does one `docker exec` per key for `GET`, another for `TTL`. With 3+ dimension keys that's 6+ subprocess calls. Consider consolidating into a single `redis-cli EVAL` or `SCAN + HGETALL`-equivalent to reduce timing surface and make the snapshot more atomic:

```bash
docker exec <container> redis-cli --scan --pattern 'rl:*' \
  | xargs -r docker exec <container> redis-cli MGET
```

---

None of these are blockers for merging the reproducer — the tests are explicitly inspection-friendly and the slow marker sets expectations. Just noting where the assertions could become unreliable as the binding fix (#4665) lands and these tests get promoted to regular CI.

gandhipratik203:
Thanks for the careful read, @msureshkumar88 — really useful framing, especially the per-item separation. Quick rundown of how each landed:

**Item 3 — `_redis_rl_keys()` silently swallowing docker exec errors.** Addressed in this PR (commit 8ebc726a9). The function now `pytest.skip`s with the container name, exception type, and a hint about `REDIS_CONTAINER_NAME` for non-default compose project names. The empty-list return path is now reserved for "container reachable, no `rl:*` keys" — semantically distinct from a docker outage. Fully aligns with your suggestion.

**Items 1 + 2 (the `team_id`-filtered scan portion) — carried forward.** Your framing of these as "risk once promoted to regular CI" still stands, but the trajectory shifted: this test isn't graduating to `tests/integration/`. It moved to `tests/live_gateway/plugins/` and is gated behind `RUN_BINDING_LIFECYCLE=1` (skipped by default, including by `make test-live-gateway`) — see the updated PR description for the new run commands. The CI-promotion path now belongs to the **fast happy-path integration test** that's tracked as item 2 on the parent issue's planned scope (https://github.com/IBM/mcp-context-forge/issues/4665#issuecomment-4485706016). Your reasoning will go into that test's design from the start: poll loop instead of flat sleep, scan pattern scoped to `rl:{team_id}:*`. Item 2 here is also already partially mitigated by the autouse `_isolate_rate_limiter_between_tests` fixture, which flushes `rl:*` between tests.

**Items 4 + 5 (Phase 5→6 TTL race, N serial subprocess calls) — not applying to this test, by design.** Both were CI-runner-timing concerns. On an opt-in operator harness with a deliberate 30s inspection pause and no atomicity expectations, they don't bite. If/when the fast integration test inherits similar shape, the subprocess fan-out and timing-margin concerns become relevant again — flagging that the analysis carries forward there too.

Net: none of your 5 items got dismissed; they either landed here (item 3) or got promoted to design constraints for the next test. Appreciate you taking the time to do the read-through rather than just rubber-stamping.

msureshkumar88 (commented):
## Re-review — items from previous round + new scan

### Previous items — status

**Item 3 — `_redis_rl_keys()` swallowing container errors → ✅ Fixed (commit 8ebc726a9)**
`pytest.skip` with container name, exception type, and `REDIS_CONTAINER_NAME` hint. Empty-list return now cleanly means "reachable, no keys". Exactly right.

**Items 1 & 2 (flat sleep, unfiltered `rl:*` scan) — deferred, accepted**
The justification holds: opt-in operator harness, not graduating to `tests/integration/`, autouse fixture partially covers item 2. Both carried to the fast happy-path integration test on #4665. No objection.

**Items 4 & 5 (TTL race, N serial subprocess calls) — not-applicable ruling accepted**
Correct for a 30s inspection-pause harness with no atomicity requirements.

---

### New findings

`L323, L340, L353`: 🟡 risk: `_stamp_tool_team_id` builds SQL via f-string interpolation (`tool_id`, `team_id`, `prev_team_id`) then passes to `docker exec psql -c`. These values originate from the gateway API and env vars — UUIDs in practice, but no shape enforcement at the call site. If `RATE_LIMITER_TEST_TEAM_ID` is set to a value containing a single-quote it will break the statement (or worse). Prefer `%s`-style psql substitution or at minimum assert UUID format before interpolating:
```python
import re
_UUID_RE = re.compile(r'^[0-9a-f-]{36}$', re.I)
assert _UUID_RE.match(tool_id), f"tool_id not a UUID: {tool_id!r}"
```

`L667-673` (`_docker_redis_cli` in `_isolate_rate_limiter_between_tests`): 🟡 risk: fixture silently swallows `FileNotFoundError / OSError / SubprocessError` and returns `None`, so if the Redis container is unreachable at fixture setup, `rl:*` cleanup is silently skipped. The mitigating fix in `_redis_rl_keys()` (item 3) raises `pytest.skip` when the same container is unreachable — this fixture should match that behavior. A `pytest.skip("redis container unreachable, cannot isolate rl:* state — aborting test")` in the `except` block keeps the two paths consistent and avoids false-positive assertions from stale keys.

---

Both new items are low-severity given the opt-in / operator-only scope. Happy to approve once the fixture inconsistency is addressed (the SQL one can be a follow-up nit). Draft status noted.

gandhipratik203:
Thanks for the re-review, @msureshkumar88.

**Item B — `_docker_redis_cli` fixture inconsistency.** Fixed in commit ae08b8e5a. The autouse fixture's helper now `pytest.skip`s with the container name, exception type, and the `REDIS_CONTAINER_NAME` hint — same shape as the Item-3 fix (`8ebc726a9`) so the two docker-exec paths behave consistently. Helper return type tightened to `subprocess.CompletedProcess` (no longer Optional), and the now-dead `is not None` check at the SCAN call site is gone.

**Item A — `_stamp_tool_team_id` SQL interpolation.** Acknowledged and accepted as-is. The concern is real in principle (operator-set `RATE_LIMITER_TEST_TEAM_ID` could theoretically contain a quote), but for an opt-in operator harness whose inputs are UUIDs sourced from the gateway API in practice, a regex assertion adds defensive code without addressing a real exposure. Matching your "follow-up nit" framing — leaving it as-is for now, the PR thread documents the rationale if anyone revisits this code later.

That should clear the approval gate. Let me know if anything else surfaced on the second pass.

msureshkumar88 (approved):
## Approval ✅

All items from both review rounds are resolved:

- **Item 3 / Item A** —  now s on docker exec failure with a clear container name + hint (commit ). Empty-list return is unambiguously "reachable, no keys" — good semantic cleanup.
- **Item B** — autouse Redis-isolation fixture's  block now matches the same skip shape (commit ). The two docker-exec paths are consistent.
- **SQL nit** — rationale documented in thread; opt-in operator harness with UUID inputs is a reasonable place to leave this as a follow-up.

CI is all green (DCO, pre-commit, pytest py3.12, vitest). Tests gated behind `RUN_BINDING_LIFECYCLE=1` don't hit the regular suite — correct for an operator-driven investigative artifact.

Good investigative work on #4665 — the seven-phase lifecycle walkthrough with inline Redis Insight and DBeaver checkpoints makes this genuinely useful for anyone reproducing the binding config-flow issue.

msureshkumar88 (approved):
## Approval ✅

All items from both review rounds are resolved:

- **Item 3 / Item A** — `_redis_rl_keys()` now `pytest.skip`s on docker exec failure with a clear container name + hint (commit `8ebc726a9`). Empty-list return is unambiguously "reachable, no keys" — good semantic cleanup.
- **Item B** — autouse Redis-isolation fixture's `except` block now matches the same skip shape (commit `ae08b8e5a`). The two docker-exec paths are consistent.
- **SQL nit** — rationale documented in thread; opt-in operator harness with UUID inputs is a reasonable place to leave this as a follow-up.

CI is all green (DCO, pre-commit, pytest py3.12, vitest). Tests gated behind `RUN_BINDING_LIFECYCLE=1` don't hit the regular suite — correct for an operator-driven investigative artifact.

Good investigative work on #4665 — the seven-phase lifecycle walkthrough with inline Redis Insight and DBeaver checkpoints makes this genuinely useful for anyone reproducing the binding config-flow issue.

jonpspri:
Closing this draft PR due to no meaningful activity (commits or author comments) in over 90 days. Feel free to reopen if you'd like to continue the work.

PR #4677, opened 2026-05-08
chore: update PII filter extra

lucarlig:
Updates the optional plugins extra to require cpex-pii-filter 0.3.2.

Moves the package-specific uv exclude-newer cutoff to 2026-05-08 so the newly released wheel is resolvable, and refreshes the lock entry for 0.3.2.


lucarlig:
other pr already bumping this up

PR #4692, opened 2026-05-10
fix: apply OTLP insecure exporter setting

### LINKED ISSUE #4644: [BUG]: OTEL_EXPORTER_OTLP_INSECURE configuration setting is defined but not used
## Description

The `OTEL_EXPORTER_OTLP_INSECURE` configuration setting is defined in `mcpgateway/config.py` and documented in multiple places, but it is **not actually used** by the OTLP exporter initialization code in `mcpgateway/observability.py`.

## Current Behavior

- Setting is defined: `mcpgateway/config.py` line 2064
- Setting is documented: `docs/docs/manage/observability/opentelemetry.md` line 292
- Setting is **NOT passed** to OTLP exporter: `mcpgateway/observability.py` lines 926-945

The code comment explicitly states:
> "Note: some versions of OTLP exporters may not accept 'insecure' kwarg; avoid passing it. Use endpoint scheme or env to control TLS externally."

## Expected Behavior

Either:
1. The `insecure` parameter should be passed to the OTLP exporter constructors, OR
2. The configuration setting should be removed/deprecated and documentation updated to clarify that SSL verification is controlled via Python environment variables (`REQUESTS_CA_BUNDLE`, `SSL_CERT_FILE`)

## Impact

Users attempting to disable SSL verification for OTLP exports to servers with self-signed certificates (e.g., Langfuse on OpenShift) find that setting `OTEL_EXPORTER_OTLP_INSECURE=true` has no effect, leading to SSL certificate verification errors.

## Workaround

Use Python's standard SSL environment variables instead:
```yaml
env:
- name: REQUESTS_CA_BUNDLE
  value: /path/to/ca-bundle.crt
```

## Proposed Fix

**Option 1: Implement the setting**
```python
if protocol == "grpc" and OTLP_SPAN_EXPORTER:
    exporter = cast(Any, OTLP_SPAN_EXPORTER)(
        endpoint=endpoint, 
        headers=header_dict or None,
        insecure=cfg.otel_exporter_otlp_insecure
    )
elif HTTP_EXPORTER:
    exporter = cast(Any, HTTP_EXPORTER)(
        endpoint=http_ep, 
        headers=header_dict or None,
        insecure=cfg.otel_exporter_otlp_insecure
    )
```

**Option 2: Remove/deprecate**
- Remove `otel_exporter_otlp_insecure` from `config.py`
- Update documentation to clarify SSL verification is controlled via `REQUESTS_CA_BUNDLE`/`SSL_CERT_FILE`

## References

- User report: Customer attempting to export to Langfuse on OCP with self-signed certs
- Code locations:
  - Config: `mcpgateway/config.py:2064`
  - Implementation: `mcpgateway/observability.py:926-945`
  - Docs: `docs/docs/manage/observability/opentelemetry.md:292`

zxyasfas:
## Summary
Fixes #4644.

`OTEL_EXPORTER_OTLP_INSECURE` is defined and documented, but `init_telemetry()` did not pass that setting into OTLP exporter construction. This makes the setting effective for OTLP tracing exporters.

## Reproduction Steps
Configure OTLP tracing with `OTEL_EXPORTER_OTLP_INSECURE=true` against a collector that requires insecure transport or skipped TLS verification. The current initialization path builds the exporter with only `endpoint` and `headers`, so the setting is ignored.

## Root Cause
`mcpgateway/observability.py` intentionally avoided passing `insecure` because not every OTLP exporter constructor supports that exact keyword. As a result, the project-level configuration value was never applied.

## Fix Description
- Add a small constructor-compatibility helper that inspects exporter kwargs before passing optional settings.
- Pass `insecure` to gRPC OTLP exporters that support it.
- For HTTP OTLP exporters, map the project insecure setting to `certificate_file=False` when supported, matching the exporter behavior that uses `certificate_file` as the `requests` TLS verification value.
- Keep older exporter versions compatible by omitting unsupported kwargs.
- Add regression coverage for both gRPC and HTTP OTLP initialization paths.

## Verification

| Check | Command | Status |
|---|---|---|
| Observability unit tests | `PYTHONUTF8=1 uv run --python 3.12 --no-dev --extra observability --with pytest --with pytest-asyncio pytest tests/unit/mcpgateway/test_observability.py -q` | Passed |
| Ruff format | `PYTHONUTF8=1 uv run --python 3.12 --no-dev --with ruff ruff format --check mcpgateway/observability.py tests/unit/mcpgateway/test_observability.py` | Passed |
| Ruff check | `PYTHONUTF8=1 uv run --python 3.12 --no-dev --with ruff ruff check mcpgateway/observability.py tests/unit/mcpgateway/test_observability.py` | Passed |
| Compile check | `py -3.12 -m py_compile mcpgateway/observability.py tests/unit/mcpgateway/test_observability.py` | Passed |

## Checklist
- [x] Code formatted

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-05-19T11:05:58Z",
+  "generated_at": "2026-05-19T12:25:29Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -9189,24 +9189,6 @@
         "is_secret": false,
         "is_verified": false,
         "line_number": 2008,
-        "type": "Secret Keyword",
-        "verified_result": null
-      }
-    ],
-    "tests/unit/mcpgateway/test_observability.py": [
-      {
-        "hashed_secret": "8f1b63868b5d31deb06a0ff740b192aa490e8950",
-        "is_secret": false,
-        "is_verified": false,
-        "line_number": 373,
-        "type": "Secret Keyword",
-        "verified_result": null
-      },
-      {
-        "hashed_secret": "1fb844731bf1e5928ae606915b54e21264da4768",
-        "is_secret": false,
-        "is_verified": false,
-        "line_number": 802,
         "type": "Secret Keyword",
         "verified_result": null
       }
--- a/mcpgateway/observability.py
+++ b/mcpgateway/observability.py
@@ -12,6 +12,7 @@ Supports any OTLP-compatible backend (Jaeger, Zipkin, Tempo, Phoenix, etc.).
 import base64
 from contextlib import nullcontext
 from dataclasses import dataclass
+import inspect
 from importlib import import_module as _im
 import logging
 import os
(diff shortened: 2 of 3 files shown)


vishu-bh (changes requested):
Thanks @zxyasfas  for picking this up. The direction makes sense, and I like the compatibility-minded approach of checking exporter constructor support before passing newer kwargs.

I found two things that I think need a small follow-up before this is ready:

1. The gRPC path now makes the existing `OTEL_EXPORTER_OTLP_INSECURE` default effective. Since the current config default is `true`, deployments using a scheme-less endpoint like `collector:4317` may change from OpenTelemetry’s secure default to plaintext transport without explicitly opting in. Could we adjust this so insecure transport is only used when explicitly configured?

2. For the HTTP path, `certificate_file=False` does not appear to disable verification with the real OpenTelemetry HTTP exporter. The exporter assigns this via `certificate_file or ...`, so `False` is ignored and verification remains enabled. That means the added fake-exporter test passes, but the real exporter behavior does not actually address the HTTP case. It would be great to update the test to exercise/reflect the real exporter behavior here.

Overall this is close, but because the setting affects TLS/insecure transport behavior, I think we should tighten the default/explicit opt-in behavior and make sure the HTTP test matches the actual exporter implementation before merging.

zxyasfas:
﻿Updated this to address both review points:

- gRPC now only passes `insecure` when `otel_exporter_otlp_insecure` was explicitly configured, so the Settings default no longer overrides exporter defaults.
- HTTP no longer passes `certificate_file=False`; the test now reflects the real HTTP exporter kwargs and keeps certificate_file unset.

Validation:
- `uv run --no-default-groups --with pytest --with pytest-asyncio --with opentelemetry-exporter-otlp-proto-http --with opentelemetry-exporter-otlp-proto-grpc pytest tests\unit\mcpgateway\test_observability.py`
- `uv run --no-default-groups --with ruff ruff check mcpgateway\observability.py tests\unit\mcpgateway\test_observability.py`

vishu-bh:
Thanks @zxyasfas for addressing the PR comments. Change looks correct to me.

Please check the failing CI tasks 🚀 Once this passes it should be good to merge

vishu-bh:
Hi @zxyasfas are you planning to address the failing CI steps on this PR?

vishu-bh (approved):
LGTM 🚀

brian-hussey (approved):
Approved based on others review and resolution.

PR #4704, opened 2026-05-11
feat(observability): add baggage attribute mapping for span customization

### LINKED ISSUE #4700: [BUG]: SpanAttributeCustomizer plugin cannot rename baggage-derived span attributes
## Bug Description

The SpanAttributeCustomizer plugin attribute mapping feature does not work for baggage-derived span attributes.

## Current Behavior

When an external plugin or client sets values through OpenTelemetry baggage, the gateway receives those values and can export them on spans as baggage-derived attributes such as:

- `baggage.tenant.id`
- `baggage.user.id`

However, clients need selected baggage keys to appear in exported spans without the `baggage.` prefix, for example:

- `tenant.id`
- `user.id`

## Root Cause

Baggage is propagation context, while span attributes are the reporting surface. The previous SpanAttributeCustomizer `attribute_mapping` approach operated in plugin/span attribute contexts and was not the right layer to control how OTEL baggage is promoted onto spans.

A SpanProcessor-based remapping approach is also not ideal because it tries to fix attribute names after span creation instead of deciding the emitted span attribute names at the point where baggage is copied onto spans.

## Proposed Solution

Keep OTEL baggage as the standards-compliant propagation mechanism, and add an explicit baggage span attribute emission policy controlled by the SpanAttributeCustomizer plugin.

Configuration:

```yaml
plugins:
  - name: "SpanAttributeCustomizer"
    kind: "plugins.span_attribute_customizer.span_attribute_customizer.SpanAttributeCustomizerPlugin"
    hooks: ["tool_pre_invoke", "resource_pre_fetch", "resource_post_fetch"]
    mode: "enforce"
    config:
      allowed_baggage_span_attributes:
        - "tenant.id"
        - "user.id"
      emit_baggage_prefixed_attributes: false
```

Expected behavior with this configuration:

- `tenant.id` and `user.id` continue to be carried through OTEL baggage.
- Only allowlisted baggage keys are promoted to span attributes.
- Exported spans contain `tenant.id` and `user.id`.
- Exported spans do not contain `baggage.tenant.id` or `baggage.user.id` for those allowlisted keys.
- If the new config is not provided, the gateway preserves legacy behavior and emits baggage as `baggage.<key>`.

Implementation should apply the policy directly where baggage is copied onto spans:

- request-root span baggage injection
- `create_span()` baggage injection

This avoids post-creation remapping and keeps the gateway aligned with OpenTelemetry semantics.

## Impact

This addresses compliance/privacy and multi-tenant observability use cases where backends require stable tenant/user span attributes without the `baggage.` prefix.

## Related PR

Implemented in PR #4705.

vishu-bh:
## Problem

Issue #4700 requests the ability to remap baggage-derived span attributes (like `baggage.tenant.id`) to custom keys (like `tenant.id`) for backend compatibility with observability systems that expect specific attribute names.

## Solution

Implements Option A semantics for the SpanAttributeCustomizer plugin:

1. **Extract mapping at startup**: `extract_span_attribute_mapping()` reads `attribute_mapping` from plugin config
2. **Add span processor**: `BaggageAttributeSpanProcessor` duplicates baggage.* attributes onto mapped keys
3. **Initialize after plugins**: Telemetry initialization moved after plugin manager setup to access config
4. **Comprehensive tests**: Added test coverage for mapping extraction and processor behavior

## Changes

- `mcpgateway/observability.py`: Added extraction function and span processor
- `mcpgateway/main.py`: Reordered telemetry init to access plugin config
- `tests/unit/mcpgateway/test_observability.py`: Added test coverage

## Configuration Example

```yaml
plugins:
  - name: SpanAttributeCustomizer
    kind: plugins.span_attribute_customizer.span_attribute_customizer.SpanAttributeCustomizerPlugin
    config:
      attribute_mapping:
        baggage.tenant.id: tenant.id
        baggage.user.email: user.email
```

## Testing

The processor safely handles:
- Missing source attributes (no-op)
- Non-baggage mappings (ignored)
- Malformed config (fail-closed)

Fixes #4700

### DIFF
--- a/mcpgateway/admin_ui/tools.js
+++ b/mcpgateway/admin_ui/tools.js
@@ -18,6 +18,7 @@ import { applyVisibilityRestrictions, isTeamScopedView } from "./teams.js";
 import {
   decodeHtml,
   fetchWithTimeout,
+  getCookie,
   getCurrentTeamId,
   handleFetchError,
   isInactiveChecked,
@@ -2927,6 +2928,12 @@ export const runToolValidation = async function (testIndex) {
     const requestHeaders = {
       "Content-Type": "application/json",
     };
+
+    // Add CSRF token from cookie to headers
+    const csrfToken = getCookie("mcpgateway_csrf_token");
+    if (csrfToken) {
+      requestHeaders["x-csrf-token"] = csrfToken;
+    }
 
     // Authentication will be handled automatically by the JWT cookie
     // that was set when the admin UI loaded. The credentials header


vishu-bh:
Incorrect PR closing it

PR #4738, opened 2026-05-13
fix: upgrade path from 0.9.0 - unnamed FK handling, temp table cleanup

madhu-mohan-jaishankar:
## 📝 Summary

Fixes three bugs that prevented clean upgrades from 0.9.0 to 1.0.1.

**1. `r2b3c4d5e6f7` — temp table race condition (SQLite)**
When multiple Gunicorn workers start simultaneously, the first worker to run the
`add_prompt_namespacing_fields` migration creates `_alembic_tmp_prompts` via
`batch_alter_table`. If that worker is killed mid-migration, the temp table is left
behind. Every subsequent worker then fails with:
```
sqlalchemy.exc.OperationalError: table _alembic_tmp_prompts already exists
```
Fix: add `DROP TABLE IF EXISTS _alembic_tmp_prompts` at the start of `upgrade()` —
a no-op on a clean database, a safe cleanup on a dirty one.

**2. `9fb98535724d` — unnamed FK constraints (SQLite upgrade from 0.9.0)**
In 0.9.0 databases, all foreign key constraints are unnamed (e.g.
`FOREIGN KEY(tool_id) REFERENCES tools (id)`). The `add_ondelete_cascade` migration
calls `inspector.get_foreign_keys()` which returns `{'name': None, ...}` for these.
The original code/previous fix either crashed or silently skipped adding
`ON DELETE CASCADE`, leaving upgraded databases without the required cascade behavior.
Fix: detect unnamed FKs, reflect the full table into a `sa.Table` object, strip the
unnamed FK constraints via `_strip_all_unnamed_fks()`, then pass `copy_from=` to
`batch_alter_table()` so SQLite recreates the table correctly with a new named
`ON DELETE CASCADE` FK.

**3. `scripts/ci/run_upgrade_validation.sh` — psycopg driver mismatch**
The validation script was hardcoding `postgresql+psycopg://` (psycopg3) for both
the base and target containers. The 0.9.0 image ships only `psycopg2_binary`, so
starting the base container immediately crashed with:
```
ModuleNotFoundError: No module named 'psycopg'
```
Fix: at startup, probe the base image with `docker run ... python3 -c "import psycopg"`.
If it fails, use `postgresql+psycopg2://` for the base container; the target container
always uses `postgresql+psycopg://` (psycopg3).

---

## 🏷️ Type of Change
- [x] Bug fix

---

## 🧪 Verification

| Upgrade validation (all 6 scenarios) | `make upgrade-validate UPGRADE_BASE_IMAGE=ghcr.io/ibm/mcp-context-forge:0.9.0 UPGRADE_TARGET_IMAGE=mcpgateway/mcpgateway:latest` | ✅ PASS |

---

## ✅ Checklist

---

## 📓 Notes

All 6 upgrade validation scenarios pass (SQLite fresh, SQLite upgrade, PostgreSQL fresh,
PostgreSQL upgrade, SQLite round-trip, PostgreSQL round-trip) against a real 0.9.0 base
image (`ghcr.io/ibm/mcp-context-forge:0.9.0`).
```

### DIFF
--- a/mcpgateway/alembic/versions/9fb98535724d_add_ondelete_cascade_to_metrics_and_.py
+++ b/mcpgateway/alembic/versions/9fb98535724d_add_ondelete_cascade_to_metrics_and_.py
@@ -47,8 +47,31 @@ def _has_cascade_fk(inspector, table_name: str, referred_table: str) -> bool:
     return False
 
 
+def _strip_all_unnamed_fks(table: sa.Table) -> None:
+    """Remove all unnamed FK constraints from a reflected Table object in-place.
+
+    When batch_alter_table is called with copy_from=table, Alembic uses the
+    provided Table as the schema for the recreated table.  Any unnamed FK
+    constraints left here would silently persist alongside the new named
+    CASCADE FK we are about to add, creating duplicate conflicting FKs.
+    Unnamed FKs arise in SQLite when upgrading from schema versions that
+    pre-date explicit constraint naming (e.g. 0.9.0 databases).
+    """
+    for c in list(table.constraints):
+        if isinstance(c, sa.ForeignKeyConstraint) and c.name is None:
+            table.constraints.discard(c)
+            for fk_elem in c.elements:
+                fk_elem.parent.foreign_keys.discard(fk_elem)
+
+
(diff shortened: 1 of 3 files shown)


brian-hussey (approved):
LGTM

PR #4739, opened 2026-05-13
docs: fix manual testing instructions for Docker reachability and registration

Lang-Akshay:
## Summary

- Fix Section 14 (Manual Testing) to work correctly when the gateway runs inside Docker — translate processes must bind to `0.0.0.0` and registrations must use `host.docker.internal` instead of `localhost`
- Add missing `--expose-streamable-http` flag and explicit `"transport":"STREAMABLEHTTP"` field required for Streamable HTTP gateway registration
- Fix virtual server creation payload (wrap fields under `"server"` key) and export command (`HOST=localhost PORT=8080` env override)
- Clarify that MCP Inspector must connect to a virtual server, not a raw gateway; rename `SERVER_UUID` to `VIRTUAL_SERVER_UUID` throughout
- Add `make docker-prod && make testing-up` prerequisite block and consolidate JWT + BASE_URL into a single code block

### DIFF
--- a/docs/docs/development/release-management.md
+++ b/docs/docs/development/release-management.md
@@ -1202,7 +1202,11 @@ The migration test suite follows an **n-2 support policy** and tests sequential 
 
 ## 14. Manual Testing
 
-These tests verify core user-facing workflows that automated tests do not fully cover. Perform them against the running compose stack (`make testing-up`).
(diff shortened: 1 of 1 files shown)


brian-hussey (approved):
LGTM

PR #4745, opened 2026-05-13
feat(ui-rewrite): MCP server edit mode, OAuth password grant validation, auth type refactoring

marekdano:
Related #4176 

### Summary

  - Implements edit mode for the MCP server form — editing an existing server pre-populates all fields from the API, submitting sends a PUT, and the servers table refreshes on success
  - Adds validation for OAuth password grant (username and password required before submit is enabled)
  - Refactors auth type mapping and OAuth config schema for correctness and maintainability
  - Fixes E2E CI workflow: Playwright browser install steps now run from ./client
  - 115 unit/integration tests covering all new paths; removes two unused components

---
###  Changes

  Edit mode (Servers.tsx, MCPServerForm.tsx, useMCPServerForm.ts)

  - MCPServerForm gains serverId and onSuccess props; the hook accepts gatewayId and fetches /gateways/:id when set, populating all form fields on load — basic fields, auth credentials, custom headers, OAuth config, passthrough headers,
  advanced settings
  - handleEdit in Servers.tsx was a throwing stub; it now sets updateServerId, opens the pre-populated form, and passes onSuccess to refetch the table and close
  - If the server fetch fails in edit mode, a red error banner is shown above the form instead of silently presenting an empty form

  OAuth password grant validation (useMCPServerForm.ts, OAuth2Auth.tsx, AdvancedSettings.tsx)

  - superRefine on the Zod schema requires oauthUsername and oauthPassword when grant_type === "password"
  - isValid memo also guards the submit button before parse
  - OAuth2Auth renders inline error messages under username/password with aria-invalid and aria-describedby wiring; AdvancedSettings and MCPServerForm thread the error props through

  Auth type mapping (useMCPServerForm.ts)

  - AUTH_TYPE_TO_API and AUTH_TYPE_FROM_API module-level constants replace two inline expressions; Record<AuthType, string> ensures TypeScript catches any missing variant when a new auth type is added

  OAuth config type safety (useMCPServerForm.ts)

  - oauthConfigSchema (typed Zod object with grant_type enum, boolean flags, typed string fields) replaces z.record(z.unknown())

  Masked value sentinel (useMCPServerForm.ts)

  - === "*****" replaces /^\*+$/ — matches the exact backend contract rather than any run of asterisks

  CI fix (.github/workflows/client-e2e.yml)

  - Added working-directory: ./client to both Playwright browser install steps; previously they ran from the repo root where playwright is not installed, causing "Executable doesn't exist" failures

 ### Removed

  - ServersEmptyState.tsx and GatewayIcon.tsx — neither was imported anywhere

  Tests (MCPServerForm.test.tsx, useMCPServerForm.test.ts, setup.ts)

  - setup.ts: four jsdom stubs required by Radix UI Select (hasPointerCapture, setPointerCapture, releasePointerCapture, scrollIntoView)
  - MCPServerForm.test.tsx: 20 custom-headers integration tests with an addHeaders(user, count) helper; 9 OAuth password grant tests using a renderWithOAuthPassword helper that drives the Radix UI Select to switch grant types
  - useMCPServerForm.test.ts: edit mode population via MSW, auth type round-trip mapping, masked sentinel handling, OAuth config serialisation per grant type, cross-field validation (115 tests total)

  ---
###  Test plan

  - Create a new MCP server — form submits and appears in the table
  - Edit an existing server — form opens pre-populated, submit sends PUT, table refreshes
  - Edit a server with OAuth password grant — username and password are required; submit is disabled until both are filled
  - Simulate a failed server fetch in edit mode — error banner appears above the form
  - E2E workflow passes without "Executable doesn't exist" error


a-effort:
A couple of TypeScript errors are surfacing in src/hooks/useMCPServerForm.ts — both look like minor typing issues from the recent OAuth/grant-type work. 🤖 Sonnet suggested fixes:                                                                                      
                                                                                                                                      
1. grant_type widened to string in oauthConfig (lines 452, 459, 468)                                                                                                                                                                                            
oauthGrantType is typed as string in component state, but oauthConfigSchema.grant_type is a Zod enum ("password" | "authorization_code" | "client_credentials"). Spreading base (which carries grant_type: string | undefined) into `oauthConfig` widens the property and fails assignment.                                                                                                                                                                                                                                    

Cleanest fix: drop grant_type from base and set the literal in each branch — TS will then narrow correctly without a cast:          

   
  ```
const base = {                                                                                                                      
    issuer: oauthIssuerUrl || undefined,                                                                                            
    scopes: scopesArray,                                                                                                              
    store_tokens: oauthStoreTokens,
    auto_refresh: oauthAutoRefresh,                                                                                                   
  };                                                                                                                                  
   
  if (oauthGrantType === "client_credentials") {                                                                                      
    oauthConfig = {                                                                                                                 
      ...base,
      grant_type: "client_credentials",
      client_id: oauthClientId || undefined,
      // ...                                                                                                                          
    };
  }                                                                                                 
```                                  
...likewise for "authorization_code" and "password"                                                                            

2. mcpServerFormSchema.shape no longer exists (line 546)                                                                            
Adding .superRefine(...) turns the schema into a ZodEffects, which doesn't expose .shape. validateField relies on it for per-field validation.                                                                                                                       
                                                                                                                                      
Two options:                                                                                                                      
  1.  Keep a reference to the inner ZodObject and use it for field validation:
 ```
 const mcpServerFormObjectSchema = z.object({ /* ... */ });                                                                          
  const mcpServerFormSchema = mcpServerFormObjectSchema.superRefine(/* ... */);
  // then in validateField:  
`const fieldSchema = mcpServerFormObjectSchema.shape[field as keyof typeof mcpServerFormObjectSchema.shape]; 
```                   
  2. Or use `mcpServerFormSchema.innerType().shape[...]`

vishu-bh (changes requested):
Thanks @marekdano 

One issue I noticed: editing an existing MCP server that already has Basic or Bearer auth appears to fail if the user does not re-enter the secret. The form sends `authType: "basic"` / `"bearer"` but omits the masked `authPassword` / `authToken` when the API returned `*****`. The backend update schema still requires the matching password/token whenever that auth type is present, so saving an unchanged authenticated server will hit validation.

marekdano:
> Thanks @marekdano
> 
> One issue I noticed: editing an existing MCP server that already has Basic or Bearer auth appears to fail if the user does not re-enter the secret. The form sends `authType: "basic"` / `"bearer"` but omits the masked `authPassword` / `authToken` when the API returned `*****`. The backend update schema still requires the matching password/token whenever that auth type is present, so saving an unchanged authenticated server will hit validation.

@vishu-bh - Good catch! I've updated the code.

vishu-bh (approved):
Thanks @marekdano 

LGTM 🚀

PR #4761, opened 2026-05-14
feat(admin): add A2A protocol version dropdown to agent form

### LINKED ISSUE #4760: [BUG]: Can't set A2A protocol version via admin UI.
### 🐞 Bug Summary

The A2A admin UI doesn't expose A2A procotol version setting form in both edit and create section.

Image

And the A2A protocol version set default 1.0. 

But in some cases A2A protocol version 0.3 might be needed (in my case. now kagent only support 0.3 version).
 
---

### 🧩 Affected Component

- [ ] `mcpgateway` - API
- [ ] Other (explain below)

---

### 🔁 Steps to Reproduce

1. Open the admin panel and navigate to the A2A Agents tab
2. Look at the **Add New A2A Agent** form. There is no field to choose the A2A protocol version
3. Submit the form. The agent is created and the protocol version is silently set to `1.0`.

---

### 🤔 Expected Behavior

The Add and Edit forms should expose a Protocol Version selector so operators can pick between `1.0` and `0.3` when registering or updating an A2A agent.

---

### 📓 Logs / Error Output

N/A

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Version or commit | `main @ v1.0.1` |
| Platform / OS | `macOS` |
| Container | `Docker` |

---

kimsehwan96:
## 🔗 Related Issue
Closes #4760

---

## 📝 Summary
Adds a Protocol Version selector (`1.0` / `0.3`, default `1.0`) to the A2A agent create and edit forms in the admin UI.

---

## 🏷️ Type of Change
- [x] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

I've done playwright e2e test.

---

## ✅ Checklist

---

### AS-IS
592399004-52de6857-a2c3-4739-9809-1a35302cabe3

### TO-BE
스크린샷 2026-05-14 오후 6 19 59

스크린샷 2026-05-14 오후 6 20 06
Protocol set with 0.3 version

스크린샷 2026-05-14 오후 6 20 14
In edit page

### DIFF
--- a/mcpgateway/admin.py
+++ b/mcpgateway/admin.py
@@ -15761,6 +15761,7 @@ async def admin_add_a2a_agent(
             description=form.get("description"),
             endpoint_url=form["endpoint_url"],
             agent_type=form.get("agent_type", "generic"),
+            protocol_version=str(form.get("protocol_version", "1.0")),
             auth_type=auth_type_from_form,
             auth_username=str(form.get("auth_username", "")),
             auth_password=str(form.get("auth_password", "")),
@@ -16050,6 +16051,11 @@ async def admin_edit_a2a_agent(
             uaid_protocol=uaid_protocol,
             uaid_native_id_override=uaid_native_id_override if generate_uaid else None,
         )
+        # Only update protocol_version when the field is explicitly submitted.
+        # Defaulting on edit would silently coerce an existing 0.3 agent back
+        # to "1.0" for any caller (cached form, scripted PATCH) that omits the field.
+        if form.get("protocol_version"):
+            agent_update.protocol_version = str(form["protocol_version"])
 
         mod_metadata = MetadataCapture.extract_modification_metadata(request, user, 0)
         await a2a_service.update_agent(
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -4338,7 +4338,7 @@
         "hashed_secret": "3ee08a29a2ca26171fe152b8741d0c90b598263a",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 15066,
+        "line_number": 15109,
         "type": "Secret Keyword",
         "verified_result": null
       }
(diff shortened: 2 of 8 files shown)


vishu-bh (changes requested):
Nice improvement, thanks @kimsehwan96  for adding this to both create and edit flows.

One thing I noticed: the edit handler now defaults missing protocol_version to "1.0". That means an existing 0.3 agent could be silently changed back to 1.0 if an older/cached form or another admin form caller submits an edit without this field.

Could we preserve the existing protocol version when the edit form omits protocol_version, instead of defaulting on edit? The create path default to 1.0 makes sense, but update should probably only change the value when the field is explicitly submitted.

vishu-bh:
Hi @kimsehwan96, are you planning to address the review comment and failing CI on this PR?

kimsehwan96:
@vishu-bh Hi, I rebased and applied your opinion about default a2a protocol version.

mcpgateway/admin.py
```diff
@@ -16017,6 +16018,11 @@ async def admin_edit_a2a_agent(
             uaid_protocol=uaid_protocol,
             uaid_native_id_override=uaid_native_id_override if generate_uaid else None,
         )
+        # Only update protocol_version when the field is explicitly submitted.
+        # Defaulting on edit would silently coerce an existing 0.3 agent back
+        # to "1.0" for any caller (cached form, scripted PATCH) that omits the field.
+        if form.get("protocol_version"):
+            agent_update.protocol_version = str(form["protocol_version"])
```
kimsehwan96, line 16058: @vishu-bh When update a2a api was called and if there is no `protocol_version` parameter in requests then just use as saved value. 

If the update API call was made from Admin UI, there must be `protocol_version` so user's intended value will be submitted. But If the API call was made from script(e.g, curl) without `protocol_version` parameter then the values will not be modified.

vishu-bh (dismissed):
Thanks @kimsehwan96 for taking on issue and addressing PR comments. 

PR looks solid. LGTM ☘️

PR #4778, opened 2026-05-15
fix(password): align validation feedback and show correct policy requirements

### LINKED ISSUE #4753: [BUG]: Password Requirements UI Not Displaying All Validation Rules
The admin UI password requirement display is **incomplete and misleading**. The UI shows only "At least 8 characters long" but the backend enforces additional complexity requirements (uppercase, lowercase, special characters, minimum 12 characters). When users enter passwords that match the displayed requirement but fail the hidden backend validation, they get a cryptic error: "New password does not meet requirements."

## Affected Flows
1. **Password Change Required Page** (`/admin/change-password-required`)
2. **Create/Edit User Page** (`/admin/users` - Create New User and Edit User forms)

### Backend Validation (Password Policy Service)
The backend enforces these requirements (`mcpgateway/services/password_policy_service.py`):
- **Minimum length**: 12 characters (for users) or 22 characters (for privileged accounts)
- **Complexity**: Must have **3 of 4 character types**:
  - Lowercase letters (a-z)
  - Uppercase letters (A-Z)
  - Numbers (0-9)
  - Special characters (!@#$%^&*(),.?":{}|<>\-_=+\[\]\\;'`~)
- Additional checks: common password detection, username-based validation, sequential character detection

### Frontend Display Issue
The admin UI only conditionally displays requirements in `mcpgateway/admin.py` lines 7854-7896:
- It checks individual `password_require_*` settings flags (uppercase, lowercase, numbers, special)
- It displays `settings.password_min_length` (defaults to 8 characters)
- **Problem**: The actual validation uses `password_min_length_user` (default 12) and enforces a **3-of-4 complexity rule**, but the UI doesn't reflect this

### Default Settings Mismatch

| Setting | Default | Used For | Displayed in UI? |
|---------|---------|----------|------------------|
| `password_min_length` | 8 | Legacy fallback | ✅ YES (incorrect) |
| `password_min_length_user` | 12 | Actual user validation | ❌ NO |
| `password_require_uppercase` | `true` | Complexity check | ✅ YES (if true) |
| `password_require_lowercase` | `true` | Complexity check | ✅ YES (if true) |
| `password_require_numbers` | `false` | Complexity check | ✅ YES (if false) |
| `password_require_special` | `true` | Complexity check | ✅ YES (if true) |

### Step 1: Password Change Required Page
1. Fresh login as admin@example.com (no prior password set)
2. Redirected to password change page showing: "At least 8 characters long"
3. Enter password: `abcd1234` (8 characters, meets displayed requirement)
4. Submit form
5. **Expected**: Password accepted (matches shown requirement)
6. **Actual**: Error "New password does not meet requirements" (backend checks require uppercase + lowercase + numbers/special)

### Step 2: Create User Page  
1. Navigate to Admin UI > User Management > Create New User
2. Enter:
   - Email: `user@example.com`
   - Full Name: `Test User`
   - Password: `password1` (9 characters, no special chars - fails backend validation)
3. **Expected**: Clear message about what's missing
4. **Actual**: Red error "Failed to create user. Please check your input." + shows only "At least 8 characters long" ✅

## Impact
- **User Confusion**: Users don't know why their password fails despite appearing to meet requirements
- **Usability**: Admin UI becomes frustrating when password creation repeatedly fails  
- **Security Communication**: Lack of clear requirements makes password policy feel arbitrary

## Required Fix

Update the admin UI to display the **actual** password requirements:

### For Password Change Page
- Minimum: **12 characters** (not 8)
- Must contain **3 of the following 4 types**:
  - Uppercase letters (A-Z)
  - Lowercase letters (a-z)
  - Numbers (0-9)
  - Special characters (!@#$%^&*(),.?":{}|<>\-_=+\[\]\\;'`~)
- Cannot contain sequential characters (e.g., "123", "abc")
- Cannot contain your username
- Cannot be a common password

### For Create/Edit User Pages
Same requirements as above

## Code Locations to Update
- **Admin HTML generation**: `mcpgateway/admin.py:7854-7896` (password requirement display)
- **Frontend validation**: `mcpgateway/admin_ui/` (password strength/validation JavaScript)
- **Backend**: No changes needed - validation logic is correct

### Option A: Fix UI to Match Current Backend (Recommended for Quick Fix)
1. Display `password_min_length_user` (12) instead of `password_min_length` (8)
2. Add UI text indicating "Must contain 3 of 4 character types" 
3. Show all applicable complexity requirements regardless of individual `password_require_*` flags

### Option B: Make Backend Configurable (Better Long-Term)
1. Add frontend-friendly `password_policy_display` config
2. Simplify the default complexity rule in display
3. Keep backend validation strict

### Option C: Sync Display with Backend Policy Service
1. Query the actual validation rules from `PasswordPolicyService.validate_user_password()`
2. Render dynamic requirement list in UI
3. Ensures display always matches backend

## Testing
- [ ] Password change page displays correct minimum length (12, not 8)
- [ ] Password change page displays "3 of 4 character types required"
- [ ] Create user page shows same requirements
- [ ] Edit user page shows same requirements
- [ ] Test with various invalid passwords to confirm error messages now match displayed requirements
- [ ] Test with `PASSWORD_POLICY_ENABLED=false` - requirements disappear
- [ ] Test with custom `password_min_length_user` - displays custom value

## Environment Context
- Affected after recent git pull (commit 16f8d1a00 or newer)
- Current password policy service is robust and correct
- Issue is purely UI/UX mismatch

rakdutta:
Closes #4753
  
  ## Summary

  Centralizes password policy requirements and standardizes validation feedback across user
  creation and password change flows. Users now receive consistent, explicit policy guidance
  (12 characters minimum, 3-of-4 complexity groups) both during form entry and on validation failure.

  ## Changes

  **Unified Policy Requirements**
  - Introduced `get_password_requirements()` static method in `PasswordPolicyService` exposing a complete,
    descriptive set of password rules (minimum length, complexity rules, restrictions) for use by both backend
    and frontend
  - Backend now returns a single `password_requirements` dictionary instead of individual policy flags,
    ensuring the frontend always receives up-to-date requirements

  **Enhanced Error Handling**
  - Improved password change flows to encode detailed error messages in redirect URLs
  - Added HTML escaping in error responses to prevent injection vulnerabilities
  - Backend validation failures now communicate specific policy violations to the user

  **Frontend User Experience**
  - Refactored admin templates to dynamically render password policy guidance based on backend requirements
  - Implemented real-time client-side validation matching backend rules (3-of-4 complexity, minimum length, etc.)
  - Enhanced error display in user creation workflows to show specific backend validation failures
  - Form submission now only possible when all requirements are met

  **Configuration & Maintenance**
  - Updated `.env.example` to document password length environment variables for different account types (following OWASP recommendations)
  - Refactored password complexity and restriction checks for clarity and maintainability (common passwords, username inclusion, sequential characters)


msureshkumar88 (changes requested):
Good direction — centralizing password requirements into a single `get_password_requirements()` call is the right move, and the HTML escaping fix in `admin_create_user` is solid. A few blocking issues need addressing before this is mergeable.

---

## 🔴 Blocking

**`admin.py` ~L4928 — URL encoding is incomplete and will break query param parsing**

```python
error_msg = str(e).replace(" ", "_").replace("#", "_").replace(" & ", "_")
return RedirectResponse(url=f"...?error=weak_password & details={error_msg}", status_code=303)
```

Characters like `?`, `%`, `+`, `=`, `/` in the exception message will corrupt the URL. The frontend JS then reverses only the space substitution (`replace(/_/g, ' ')`), which also corrupts any legitimate underscores in error text. Use `urllib.parse.quote()`:

```python
from urllib.parse import quote
return RedirectResponse(
    url=f"{root_path}/admin/change-password-required?error=weak_password & details={quote(str(e))}",
    status_code=303,
)
```

And on the JS side: `decodeURIComponent(details)` instead of the underscore dance.

---

**`password_policy_service.py` ~L475, L510 — new config vars are not defined in `config.py`, so they can never be set**

`PASSWORD_MIN_LENGTH_USER`, `PASSWORD_MIN_LENGTH_PRIVILEGED`, and `PASSWORD_MIN_LENGTH_SERVICE` are documented in `.env.example` but no Pydantic fields for them exist in `mcpgateway/config.py`. As a result, `getattr(settings, "password_min_length_user", 12)` always returns the hardcoded default `12` — Pydantic Settings never parses these env vars. Operators cannot actually configure the limits they're promised. The fields need to be added to the `Settings` class in `config.py`.

---

**`password_policy_service.py` — enforcement and display read different settings keys, creating a silent divergence**

`validate_user_password()` (the actual backend enforcement) reads `password_min_length`. `get_password_requirements()` and `validate_password_detailed()` read `password_min_length_user`. An operator who sets only `PASSWORD_MIN_LENGTH=16` will get a UI telling users "12 characters required" while the backend actually enforces 16. These must read the same setting, or `validate_user_password()` must be updated to read `password_min_length_user` as well.

---

**`password_policy_service.py` ~L495 — `validate_password_detailed()` is dead code**

The method is added but never called anywhere in the diff. The PR description states "Backend validation failures now communicate specific policy violations to the user," but `admin_create_user` still routes through `validate_password_strength()` → old error tuple, and `change_password_required_handler` catches `PasswordValidationError` from `email_auth_service` (not this new method). Either wire `validate_password_detailed()` into the handlers, or remove it — shipping untested dead code adds maintenance burden without user benefit.

---

## 🟡 Risk

**`admin.py` ~L4928 — password policy error details leak into URL**

`details=<error_message>` in the redirect URL means validation feedback is written to access logs, browser history, and HTTP Referer headers (which third-party analytics may capture). The `PasswordValidationError` message is benign, but this pattern sets a precedent. A session flash message or extending the existing `error=` enum would avoid the leakage without sacrificing UX clarity.

---

## 🔵 Nit

`validate_password_detailed()` duplicates the complexity/length logic from `validate_user_password()`. If this method is kept and wired up, consider extracting the shared logic to avoid two diverging copies of the same rules.

rakdutta:
@msureshkumar88 Thanks for the review. I have addressed all the review comments. Please review and merge.

msureshkumar88 (changes requested):
Good progress on the follow-up commits — the `urllib.parse.quote()` fix and the removal of `validate_password_detailed()` are the right moves. Three blocking items from round 1 still need attention, and there's a new regression introduced by the partial B1 fix.

---

## 🔴 Blocking

**`change-password-required.html` — frontend decode now mismatches backend encoding (regression)**

The backend correctly encodes with `urllib.parse.quote()`, but the template JS still decodes with `details.replace(/_/g, ' ')`. After the backend change, the `details` param contains percent-encoded chars (`%20`, `%3D`, etc.), not underscores — so users will see raw `%20` instead of spaces. The frontend must be updated to match:

```js
// replace this
errorMessage = details ? details.replace(/_/g, ' ').replace(/\b\w/g, l => l.toUpperCase()) : 'New password does not meet requirements';

// with this
errorMessage = details ? decodeURIComponent(details) : 'New password does not meet requirements';
```

---

**`password_policy_service.py` — `PASSWORD_MIN_LENGTH_USER`, `PASSWORD_MIN_LENGTH_PRIVILEGED`, `PASSWORD_MIN_LENGTH_SERVICE` are still not Pydantic fields (B2 unresolved)**

`config.py` received only one new field (`password_error_message_max_length`). The three role-scoped length vars documented in `.env.example` are still absent from the `Settings` class. `getattr(settings, "password_min_length_user", 12)` will always return the hardcoded default `12` regardless of what an operator sets — Pydantic BaseSettings never reads the env var. Add to `config.py`:

```python
password_min_length_user: int = 12
password_min_length_privileged: int = 22
password_min_length_service: int = 20
```

---

**`password_policy_service.py` — enforcement and display still read different settings keys (B3 unresolved)**

`validate_user_password()` enforces `settings.password_min_length` (an existing field). `get_password_requirements()` displays `settings.password_min_length_user` (the new field, once B2 is fixed). An operator who sets only `PASSWORD_MIN_LENGTH=16` will get a UI showing "12 characters required" while the backend rejects anything under 16. Either:

- Update `validate_user_password()` to read `password_min_length_user` as the enforcement key for regular accounts (consistent with the new role-scoped design), or
- Have `get_password_requirements()` fall back to `password_min_length` when the user-specific field isn't set

Either approach is acceptable, but the two methods must read the same effective setting.

---

**CI: pre-commit hooks failing**

The latest push has `pre-commit` failing. Please fix before merge.

---

## 🟡 Risk

**`admin.py` — `change_password_required_page` always renders regular-user requirements regardless of account type**

`get_password_requirements(is_privileged=False)` is called unconditionally. A `platform_admin` account going through the mandatory change flow will see "12 characters required" in the UI but the backend will enforce a higher threshold once B2/B3 are fixed. The page has access to the session — pass `is_privileged` based on the current user's admin status.

**`password_policy_service.py` — `get_password_requirements()` returns aspirational restrictions not enforced by `validate_user_password()`**

The `restrictions` list includes "Cannot contain your username" and "Cannot be a common password", but neither check exists in `validate_user_password()`. Users will follow guidance that the backend doesn't actually enforce. Either implement the checks or remove the misleading entries.

**`password-validator.js` — `window.PasswordValidator` assignment pollutes global scope**

The object is assigned directly to `window`. Any other script that defines `PasswordValidator` silently overwrites it (or gets overwritten). Wrap in an IIFE or check for existence:

```js
window.PasswordValidator = window.PasswordValidator || (function() { ... return { ... }; })();
```

**`admin.py` `change_password_required_handler` — `LOGGER.warning` missing `exc_info=True`**

```python
LOGGER.warning(f"Password validation failed for {current_user.email}: {e}")
```

The exception string is logged but the traceback is dropped. When debugging edge cases in `PasswordPolicyError` subclasses, this makes root-cause analysis harder. Add `exc_info=True`.

---

## 🔵 Test coverage gaps

**`test_password_policy.py` — tests don't verify config wiring for role-scoped length fields**

`test_get_password_requirements` asserts `requirements["min_length"] == 12`, which always passes because `getattr(settings, "password_min_length_user", 12)` hardcodes 12 as the fallback. Once B2 is fixed, add a test that monkeypatches `settings.password_min_length_user = 16` and confirms the returned dict reflects it. Otherwise the test provides no signal that the config wiring works.

**No test coverage for URL encoding/truncation path in `change_password_required_handler`**

The new `error_msg` truncation + `urllib.parse.quote()` block in `admin.py` has no unit test. A test covering: (a) short message passes through, (b) long message is truncated and still valid percent-encoded output, (c) decoded result matches original text — would lock in the contract.

**No assertion that `get_password_requirements()` and `validate_user_password()` agree on the effective minimum length**

Add a test that creates a policy instance and confirms the minimum length used for enforcement matches the minimum length returned for display under the same settings. This would have caught B3 before it shipped.

---

## 🔵 Nit

`password_policy_service.py` — `get_password_requirements()` constructs a fresh dict (including f-string formatting) on every call. Since settings don't change at runtime, `@functools.lru_cache` would eliminate redundant construction across the page loads that call it on every request.

rakdutta:
@msureshkumar88 

Thanks for the detailed review! I wanted to clarify that these three blocking issues (B1, B2, B3) were already resolved in earlier commits on this branch. Current status below:

### B1: Frontend decoding (`change-password-required.html`)

The frontend already uses `decodeURIComponent()` as suggested. In `change-password-required.html` (line 588):

```js id="r1k4a2"
errorMessage = details ? escapeHtml(decodeURIComponent(details)) : 'New password does not meet requirements';
```

This correctly decodes the URL-encoded error messages returned from the backend via `urllib.parse.quote()`.

---

### B2: Pydantic fields for role-scoped password lengths

The password length configuration fields are already present in `mcpgateway/config.py` (lines 985–987):

```python id="z9t7bc"
password_min_length_user: int = Field(default=12, description="Minimum password length for user accounts (OWASP recommendation)")
password_min_length_privileged: int = Field(default=22, description="Minimum password length for privileged accounts (12 + 10 chars)")
password_min_length_service: int = Field(default=20, description="Minimum password length for service accounts (64-128 bit entropy)")
```

These are wired through Pydantic `BaseSettings` and are now correctly loaded from environment variables.

---

### B3: Enforcement and display alignment

Both `validate_user_password()` and `get_password_requirements()` now use the same role-scoped settings, ensuring backend enforcement and UI display remain aligned.

In `password_policy_service.py`:

#### `validate_user_password()` (lines 147–149)

```python id="m7q2np"
min_length = getattr(settings, "password_min_length_user", 12)
if is_privileged:
    min_length = getattr(settings, "password_min_length_privileged", 22)
```

#### `get_password_requirements()` (lines 487–489)

```python id="x3v8dl"
min_length = getattr(settings, "password_min_length_user", 12)
if is_privileged:
    min_length = getattr(settings, "password_min_length_privileged", 22)
```

This ensures operators see password requirements in the UI that exactly match backend validation behavior.

I am also reviewing the remaining review points and will address them in the next update.

msureshkumar88 (dismissed):
All blocking items from both review rounds are resolved.

## ✅ Verified Fixed

**Round 1 → Round 2 blockers:**
- B1: `urllib.parse.quote()` on backend + `decodeURIComponent(details)` on frontend (L588) — encoding/decoding now match
- B2: `password_min_length_user`, `password_min_length_privileged`, `password_min_length_service` added as Pydantic fields in `config.py` — env vars are now actually read
- B3: `validate_user_password()` now reads `password_min_length_user` — enforcement and display read the same key

**Risk items:**
- `is_privileged` now derived from JWT cookie in `change_password_required_page` — admin accounts see correct requirements
- All three restrictions in `get_password_requirements()` are backed by real enforcement: common-password check, username-inclusion check, sequential-chars check
- `window.PasswordValidator = window.PasswordValidator || {` — idempotent guard prevents global scope collision
- `exc_info=True` added to `LOGGER.warning`

**CI:** All 29 checks passing.

Two optional nits from round 2 remain (no test coverage for config-wiring monkeypatch and URL truncation path) — not blocking.

brian-hussey (approved):
Approved based on previous review

brian-hussey:
Failures on files unrelated to this PR, merging.

PR #4782, opened 2026-05-15
fix: enable React app cookie authentication for API endpoints

gcgoncalves:
🔗 Issue
Relates to #4269

🚀 Summary
Fixed 401 Unauthorized errors when React app accesses API endpoints (e.g., /teams/) using cookie authentication. Modified RBAC middleware to recognize X-Requested-With: XMLHttpRequest header as valid first-party signal, and changed JWT/CSRF cookie paths from /app to / to allow cookies to be sent to root-level API endpoints.

📓 Notes
 - RBAC middleware now accepts X-Requested-With header + /app referer as proof of first-party origin
 - JWT cookie path changed from /app to / (mcpgateway/routers/app.py:40)
 - CSRF cookie path changed from /app to / (mcpgateway/utils/csrf.py:64)
 - Cookie path scoping to /app was overly restrictive given React app needs to call root-level APIs
 - X-Requested-With recognition is the primary security improvement; cookie path change enables functionality

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-05-10T22:09:49Z",
+  "generated_at": "2026-05-15T15:01:15Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -4988,7 +4988,7 @@
         "hashed_secret": "ff37a98a9963d347e9749a5c1b3936a4a245a6ff",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 2911,
+        "line_number": 2933,
         "type": "Secret Keyword",
         "verified_result": null
       }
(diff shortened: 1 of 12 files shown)


vishu-bh (changes requested):
I think this needs a little more hardening before merge because it touches cookie auth boundaries.

1. `X-Requested-With` is client-controlled, unlike `Sec-Fetch-*`. The new check in `rbac.py` accepts `Referer` path `/app/...` plus `X-Requested-With: XMLHttpRequest`, but it does not validate the `Referer`/`Origin` scheme + host. A forged referer such as `https://evil.example/app/...` could satisfy the current path-only check server-side. If this fallback is needed, please validate the full origin against the configured gateway origin before allowing cookie auth.

2. Root API mutations still appear to fail CSRF. The React login sets a 43-char random `csrf_token`, while the global CSRF middleware validates `settings.csrf_cookie_name` (`mcpgateway_csrf_token` by default) and expects a 64-char HMAC-bound token. So widening the cookie path lets the cookie reach `/teams`, `/gateways`, etc., but POST/PUT/DELETE requests will still be rejected by the global CSRF middleware.

3. Please handle cookie path migration. Existing users may already have `jwt_token`/`csrf_token` scoped to `/app`. Setting new cookies at `/` creates separate cookies rather than replacing the old ones, and logout currently clears only `Path=/`. Login/logout should expire both the old `/app` cookies and the new `/` cookies to avoid duplicate-cookie ambiguity or stale sessions during rollout.

I’d also update the remaining e2e cookie flag expectations that still assert `Path=/app`.

vishu-bh (approved):
Thanks for fixing this!!. 

LGTM ☘️

PR #4784, opened 2026-05-15
fix: eliminate migration lock contention on multi-replica startup 

### LINKED ISSUE #4783: [BUG]: Slow startup with multiple replicas — all N instances serialize on Alembic migration lock
## Summary

`make testing-up` starts **3 gateway replicas** (default `GATEWAY_REPLICAS=3`) simultaneously. Every replica calls `bootstrap_db()` → `alembic upgrade head` inside an advisory lock. Because the lock serializes migrations, startup time scales linearly with replica count: **3 replicas = 3× migration time before any instance is healthy.**

On PostgreSQL the instances queue correctly but wait needlessly. On SQLite the lock is silently broken in multi-container setups (each container has its own `/tmp`), so all migrations race concurrently against the same file.

---

## Steps to Reproduce

```bash
make testing-up          # starts 3 gateway replicas by default
docker compose logs -f gateway   # observe all three instances running alembic upgrade head
```

Expected: only one instance runs migrations; others wait or skip.
Actual: all three acquire or spin on the lock, delaying readiness by 2–3× migration time.

---

### 1. Default replica count is 3

`docker-compose.yml:784`:

```yaml
deploy:
  replicas: ${GATEWAY_REPLICAS:-3}
```

`make testing-up` passes no `GATEWAY_REPLICAS` override, so all three replicas launch at once.

### 2. Every instance runs bootstrap_db() unconditionally

`mcpgateway/main.py:1319–1328` (inside `lifespan()`):

```python
from mcpgateway.bootstrap_db import main as bootstrap_db
...
await bootstrap_db()
```

There is no guard — every replica, on every restart, calls `alembic upgrade head` under a lock.

### 3. PostgreSQL: serialized wait multiplies startup latency

`mcpgateway/bootstrap_db.py` uses `pg_try_advisory_lock` (non-blocking) with exponential backoff (up to 60 × 10 s = **10 min** worst case). Migration steps on a populated schema typically take 5–30 s each. With 3 replicas:

| Instance  | Waits for                 | Extra delay          |
| --------- | ------------------------- | -------------------- |
| Replica 1 | — (acquires immediately) | 0                    |
| Replica 2 | Replica 1 finishes        | +migration_time      |
| Replica 3 | Replicas 1 + 2 finish     | +2 × migration_time |

Result: replica 3's healthcheck passes only after **3× the single-instance migration time**.

### 4. SQLite: FileLock is per-container — no cross-container protection

`mcpgateway/bootstrap_db.py:57–59`:

```python
_MIGRATION_LOCK_PATH = os.path.join(tempfile.gettempdir(), "mcpgateway_migration.lock")
# ...
file_lock = FileLock(_MIGRATION_LOCK_PATH, timeout=_MIGRATION_LOCK_TIMEOUT)
```

`tempfile.gettempdir()` returns `/tmp` inside each container. Because `/tmp` is not a shared volume, every container has its **own independent lock file**. All SQLite migrations run in parallel, hitting the same database file without any coordination → SQLite write-lock contention, `OperationalError: database is locked`, or silent data races.

### 5. No "already at head" fast-path

Before acquiring the advisory lock, `bootstrap_db()` does not check whether `alembic_version` already points to `head`. Even instances that arrive after migrations are complete must acquire the lock, invoke `alembic upgrade head` (which is a no-op but still takes ~1–3 s), then release. This adds unnecessary lock contention on every rolling restart.

### 6. An init-container solution already exists but is commented out

`docker-compose.yml:1067–1082`:

```yaml
# migration:
#   image: mcpgateway/mcpgateway:latest
#   command: alembic -c mcpgateway/alembic.ini upgrade head
#   depends_on:
#     postgres:
#       condition: service_healthy
#   networks: [mcpnet]
```

This pattern — run migrations once in an init container, then start replicas — would eliminate the problem entirely for PostgreSQL deployments.

---

## Impact

- **`make testing-up`** startup time is 2–3× slower than expected (observed by user, reproduced by code review).
- In CI or load-test environments where `testing-up` is run frequently, the delay compounds.
- SQLite + multi-container setups have a **silent correctness bug**: the distributed lock provides no protection.

---

### Option A — Pre-lock fast-path check (lowest risk, addresses PostgreSQL slowness)

Before entering `advisory_lock`, read `alembic_version` and compare to the known head revision. If already at head, skip the lock entirely:

```python
async def main() -> None:
    engine = create_engine(settings.database_url)
    with engine.connect() as probe:
        if _is_at_head(probe):          # cheap SELECT on alembic_version
            logger.info("Schema already at head, skipping migration lock")
            await bootstrap_admin_user(probe)
            ...
            return
    # slow path: acquire lock, run upgrade
    ...
```

This makes replicas 2 and 3 pay only a single cheap query on subsequent starts (after replica 1 has migrated).

### Option B — Dedicated init container (cleanest, eliminates lock entirely for PostgreSQL)

Uncomment and enable the `migration:` service in `docker-compose.yml`, add a `depends_on` condition so gateway replicas only start after the init container exits successfully:

```yaml
migration:
  image: ${IMAGE_LOCAL:-mcpgateway/mcpgateway:latest}
  command: alembic -c mcpgateway/alembic.ini upgrade head
  depends_on:
    postgres:
      condition: service_healthy
  networks: [mcpnet]

gateway:
  depends_on:
    migration:
      condition: service_completed_successfully
    postgres:
      condition: service_healthy
```

Disable migration in lifespan when `SKIP_MIGRATION=true` (set by the init container profile).

### Option C — Fix SQLite FileLock path (correctness fix, separate from Option A/B)

Replace the per-container `/tmp` path with a path on the shared database volume so the file lock is visible to all containers sharing the SQLite file. Or document that SQLite is unsupported for multi-replica deployments and add a startup assertion.

---

## Environment

- Reproduced via `make testing-up` with default `GATEWAY_REPLICAS=3`
- `mcpgateway/bootstrap_db.py` — `advisory_lock()`, `main()`
- `mcpgateway/main.py:1319–1328` — lifespan startup
- `docker-compose.yml:784` — gateway replicas default

Lang-Akshay:
## Summary

Fixes the startup serialization bug reported in #4783 where all N gateway replicas
unconditionally acquire the Alembic migration lock at boot, making startup time scale
linearly with replica count. Also surfaces the silent SQLite correctness bug where
per-container `/tmp` file locks provide no cross-container coordination.

Three complementary fixes are applied together:

- **Fast-path head check** — before acquiring the advisory lock, each replica does a
  single `SELECT` on `alembic_version` and compares to the known head revision. If
  already at head the lock and `alembic upgrade head` are skipped entirely; only the
  idempotent bootstrap helpers (admin user, RBAC roles, resource assignments) re-run.
- **Init container** — the previously commented-out `migration:` service in
  `docker-compose.yml` is enabled. It runs the full bootstrap once against Postgres
  directly (bypassing PgBouncer), then exits. All gateway replicas depend on
  `service_completed_successfully`, so they start in parallel only after the schema
  is ready — and all take the fast-path immediately.
- **SQLite multi-replica guard** — when `DATABASE_URL` starts with `sqlite` and
  `GATEWAY_REPLICAS > 1`, a `WARNING` is logged at startup explaining that `/tmp`
  is not shared across containers and the file lock provides no cross-container
  protection. The warning does not abort startup so existing single-node deployments
  that happen to set `GATEWAY_REPLICAS` are unaffected.

A fourth addition, `SKIP_MIGRATION=false` (new env var), lets operators or external
CI pipelines declare that schema is already current, running only the bootstrap
helpers and skipping Alembic entirely.

Closes #4783

---

### `mcpgateway/bootstrap_db.py`

**New function: `_is_at_alembic_head(conn, cfg) -> bool`** (after `_schema_looks_current`)

Uses Alembic's own `MigrationContext.get_current_heads()` and
`ScriptDirectory.get_heads()` for an authoritative revision comparison — not a
column heuristic. Returns `False` on any exception (fail-safe: always falls through
to the full migration path). This keeps the check accurate across future migrations
without manual maintenance.

```
_is_at_alembic_head logic:
  1. sa.inspect(conn) → check alembic_version table exists
  2. MigrationContext.configure(conn).get_current_heads() → current DB revision(s)
  3. ScriptDirectory.from_config(cfg).get_heads() → target revision(s) from scripts
  4. return current_heads == target_heads
  5. on any exception → log WARNING, return False
```

**Modified: `main()`**

Four additions, in order of execution:

1. `cfg.set_main_option("sqlalchemy.url", escaped_url)` moved **before** the
   connection block so `cfg` is ready for `_is_at_alembic_head` without needing
   a locked connection.
2. **SQLite multi-replica guard** — reads `GATEWAY_REPLICAS` env var; logs `WARNING`
   when `sqlite` + replicas > 1 explaining the per-container lock limitation.
3. **`SKIP_MIGRATION` early-return** — if `settings.skip_migration` is `True`, runs
   only the three idempotent bootstrap helpers and returns. Intended for deployments
   where an external tool (init container, CI step) already ran `alembic upgrade head`.
4. **Fast-path branch** — wraps the existing `advisory_lock` block in an `else`:

   ```python
   if _is_at_alembic_head(conn, cfg):
       # skip lock — run idempotent bootstraps only
   else:
       with advisory_lock(conn):
           command.upgrade(cfg, "head")
           # ... normalization + bootstraps
   ```

The slow path (advisory lock + `alembic upgrade head`) is completely unchanged.

### `mcpgateway/config.py`

New field in `Settings`:

```python
skip_migration: bool = Field(
    default=False,
    description="Skip alembic upgrade head on startup. ..."
)
```

Env var: `SKIP_MIGRATION=true`. Default `false` — no behavior change for existing
deployments.

### `docker-compose.yml`

**Enabled `migration:` service** (was fully commented out):

```yaml
migration:
  image: ${IMAGE_LOCAL:-mcpgateway/mcpgateway:latest}
  build:
    context: .
    dockerfile: Containerfile.lite
  environment:
    - DATABASE_URL=postgresql+psycopg://postgres:${POSTGRES_PASSWORD:-...}@postgres:5432/mcp
    - JWT_SECRET_KEY=...
    - BASIC_AUTH_USER=...
    - BASIC_AUTH_PASSWORD=...
    - PLATFORM_ADMIN_EMAIL=...
    - AUTH_REQUIRED=...
    - AUTH_ENCRYPTION_SECRET=...
    - LOG_LEVEL=...
  command: ["python", "-m", "mcpgateway.bootstrap_db"]
  depends_on:
    postgres:
      condition: service_healthy
  networks: [mcpnet]
  restart: "no"
```

Key decisions:

- `DATABASE_URL` points to `postgres:5432` **not** `pgbouncer:6432`. Postgres
  advisory locks are session-scoped; PgBouncer in transaction-pooling mode releases
  the session between statements, which breaks session-level lock semantics.
- `command: python -m mcpgateway.bootstrap_db` runs the **full** bootstrap (schema +
  admin user + roles + resource assignments), not just `alembic upgrade head`. This
  means gateway replicas skip all bootstrap work on fast-path, not just migration.
- `restart: "no"` — the container must exit 0 for Docker Compose's
  `service_completed_successfully` condition to fire. If it fails, replicas never
  start and the operator sees a clear error in `docker compose logs migration`.

**Updated `gateway.depends_on`** — uncommenting the previously commented condition:

```yaml
depends_on:
  pgbouncer:
    condition: service_healthy
  redis:
    condition: service_started
  migration:
    condition: service_completed_successfully
```

### `mcpgateway/services/base_service.py`

Fixed team-scoped visibility leak in `_apply_team_filter()`. Previously a standalone
`model_cls.visibility == "public"` condition was appended unconditionally, causing
globally public items from **other teams** to appear in team-scoped listings. The
condition is removed; when `team_id` is set, only items belonging to that team (with
`team` or `public` visibility) are returned. The docstring was updated to reflect the
corrected semantics.

### `mcpgateway/admin.py`

UI fix in the team members panel: updated the close button's `data-action-click`
attribute to use the correct handler.

### `mcpgateway/templates/admin.html`

Removed three stale `window.Admin` registrations (`loadTeamMembersView`,
`approveJoinRequest`, `rejectJoinRequest`) that no longer correspond to defined
functions.

---

## Behavior by Scenario

| Scenario                                                     | Before                                                             | After                                                                                   |
| ------------------------------------------------------------ | ------------------------------------------------------------------ | --------------------------------------------------------------------------------------- |
| `make serve` / `make dev` (SQLite, 1 replica, no Docker) | Lock → migrate → bootstrap                                       | First start: identical. Subsequent restarts: fast-path (single SELECT, ~100ms vs 1–3s) |
| `make testing-up` cold start (PG, 3 replicas)              | All 3 race for advisory lock, start serially (~3× migration time) | Init container migrates once → all 3 replicas fast-path skip → start in parallel      |
| `make testing-up` existing stack (schema at head)          | All 3 acquire lock, run no-op upgrade                              | Init container exits immediately → all 3 fast-path skip immediately                    |
| Rolling restart (PG, 3 replicas, no compose)                 | All 3 serialize on advisory lock                                   | All replicas fast-path (schema already at head, no lock acquired)                      |
| SQLite +`GATEWAY_REPLICAS > 1` (accidental)                | Silent correctness bug — no cross-container lock                  | `WARNING` logged at startup explaining the per-container lock limitation              |
| External migration tool / CI pipeline                        | N/A                                                                | `SKIP_MIGRATION=true` skips Alembic, runs bootstraps only                             |

---

## Notes

- The fast-path check is **fail-safe**: any exception from `_is_at_alembic_head`
  returns `False`, which falls through to the existing advisory-lock path. A false
  negative (unnecessary lock acquisition) is safe; Alembic's `upgrade head` is
  idempotent. A false positive (skipping a needed migration) cannot occur because the
  check compares actual Alembic revision hashes, not schema column presence.
- The TOCTOU race between fast-path check and lock acquisition is safe: if two
  replicas both see "not at head" and both enter `advisory_lock`, the second waits
  for the first, then runs `command.upgrade` which is a no-op. Existing behavior.
- The `migration:` service is intentionally not gated behind a Compose profile so
  it activates automatically for all `make testing-up` / `make compose-up` users
  without any additional flags.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -754,8 +754,8 @@ services:
       # Direct PostgreSQL (uncomment if bypassing PgBouncer):
       # postgres:
       #   condition: service_healthy
-      # migration:
-      #   condition: service_completed_successfully
+      migration:
+        condition: service_completed_successfully
 
     healthcheck:
       ## Uncomment for HTTP healthcheck
@@ -1063,19 +1063,30 @@ services:
           cpus: '0.5'
           memory: 128M
 
-  # migration:
-  #   #image: ghcr.io/ibm/mcp-context-forge:0.7.0 # Testing migration from 0.7.0
-  #   image: mcpgateway/mcpgateway:latest # Use the local latest image. Run `make docker-prod` to build it.
-  #   build:
-  #     context: .
-  #     dockerfile: Containerfile.lite
-  #   environment:
-  #     - DATABASE_URL=postgresql+psycopg://postgres:${POSTGRES_PASSWORD:-mysecretpassword}@postgres:5432/mcp
-  #   command: alembic -c mcpgateway/alembic.ini upgrade head
-  #   depends_on:
-  #     postgres:
-  #       condition: service_healthy
-  #   networks: [mcpnet]
+  migration:
+    # One-shot init container: runs full schema migration + bootstrap before any gateway replica starts.
+    # Uses Postgres directly (port 5432) — advisory locks are session-scoped and must bypass PgBouncer
+    # which operates in transaction-pooling mode.
+    #image: ghcr.io/ibm/mcp-context-forge:0.7.0 # Uncomment to test upgrade from a specific release
+    image: ${IMAGE_LOCAL:-mcpgateway/mcpgateway:latest} # Run `make docker-prod` to build local image
+    build:
+      context: .
+      dockerfile: Containerfile.lite
+    environment:
+      - DATABASE_URL=postgresql+psycopg://postgres:${POSTGRES_PASSWORD:-mysecretpassword}@postgres:5432/mcp
+      - JWT_SECRET_KEY=${JWT_SECRET_KEY:-your-secret-jwt-key-here}
+      - BASIC_AUTH_USER=${BASIC_AUTH_USER:-admin}
+      - BASIC_AUTH_PASSWORD=${BASIC_AUTH_PASSWORD:-changeme}
+      - PLATFORM_ADMIN_EMAIL=${PLATFORM_ADMIN_EMAIL:-admin@example.com}
+      - AUTH_REQUIRED=${AUTH_REQUIRED:-true}
+      - AUTH_ENCRYPTION_SECRET=${AUTH_ENCRYPTION_SECRET:-my-test-salt}
+      - LOG_LEVEL=${LOG_LEVEL:-INFO}
+    command: ["python", "-m", "mcpgateway.bootstrap_db"]
+    depends_on:
+      postgres:
+        condition: service_healthy
+    networks: [mcpnet]
+    restart: "no"
 
 ###############################################################################
 #  CACHE
--- a/mcpgateway/bootstrap_db.py
+++ b/mcpgateway/bootstrap_db.py
@@ -788,8 +788,7 @@ async def main() -> None:
                 if not _is_at_alembic_head(conn, cfg):
                     logger.error("SKIP_MIGRATION=true but schema is not at head — " "run 'alembic upgrade head' before starting the application")
                     raise RuntimeError("Schema not at head; migrations required before startup")
-                else:
-                    logger.info("SKIP_MIGRATION=true — schema already at head, skipping migration")
+                logger.info("SKIP_MIGRATION=true — schema already at head, skipping migration")
                 updated = normalize_team_visibility(conn)
                 if updated:
                     logger.info(f"Normalized {updated} team record(s) to supported visibility values")
--- a/mcpgateway/bootstrap_db.py
+++ b/mcpgateway/bootstrap_db.py
@@ -46,6 +46,7 @@ from filelock import FileLock
 from sqlalchemy import create_engine, or_, text
 from sqlalchemy.engine import Connection
 from sqlalchemy.orm import Session
+from alembic.runtime.migration import MigrationContext
 
 # First-Party
 from mcpgateway.common.validators import SecurityValidator
@@ -115,7 +116,7 @@ def _is_at_alembic_head(conn: Connection, cfg: Config) -> bool:
     """
     # pylint: disable=import-outside-toplevel
     # Third-Party
-    from alembic.runtime.migration import MigrationContext
+   
     from alembic.script import ScriptDirectory
     import sqlalchemy as sa
 
--- a/mcpgateway/bootstrap_db.py
+++ b/mcpgateway/bootstrap_db.py
@@ -42,11 +42,11 @@ from typing import cast
 # Third-Party
 from alembic import command
 from alembic.config import Config
+from alembic.runtime.migration import MigrationContext
 from filelock import FileLock
 from sqlalchemy import create_engine, or_, text
 from sqlalchemy.engine import Connection
 from sqlalchemy.orm import Session
-from alembic.runtime.migration import MigrationContext
 
 # First-Party
 from mcpgateway.common.validators import SecurityValidator
@@ -116,7 +116,8 @@ def _is_at_alembic_head(conn: Connection, cfg: Config) -> bool:
     """
     # pylint: disable=import-outside-toplevel
     # Third-Party
-   
+
+    # Third-Party
     from alembic.script import ScriptDirectory
     import sqlalchemy as sa
 
@@ -778,18 +779,15 @@ async def main() -> None:
 
     # SKIP_MIGRATION: external tooling (init container, CI pipeline) already ran
     # alembic upgrade head — skip schema migration and run only the idempotent
-    # bootstrap helpers.  However, if the schema is not yet at head (e.g. first
-    # boot on a fresh DB before any init container has run), fall through to
-    # migrations so the app does not crash with "no such table" errors.
+    # bootstrap helpers.  If the schema is NOT at head, the init container failed
+    # or the wrong DB was targeted; fail fast so the deployment error is visible.
     if settings.skip_migration:
         try:
             with engine.connect() as conn:
                 conn.commit()
                 if not _is_at_alembic_head(conn, cfg):
-                    logger.warning("SKIP_MIGRATION=true but schema is not at head " "(fresh DB or pending migrations) — running alembic upgrade head")
-                    with advisory_lock(conn):
-                        cfg.attributes["connection"] = conn
-                        command.upgrade(cfg, "head")
+                    logger.error("SKIP_MIGRATION=true but schema is not at head — " "run 'alembic upgrade head' before starting the application")
+                    raise RuntimeError("Schema not at head; migrations required before startup")
                 else:
                     logger.info("SKIP_MIGRATION=true — schema already at head, skipping migration")
                 updated = normalize_team_visibility(conn)
--- a/mcpgateway/bootstrap_db.py
+++ b/mcpgateway/bootstrap_db.py
@@ -114,6 +114,7 @@ def _is_at_alembic_head(conn: Connection, cfg: Config) -> bool:
         False when the database is fresh, behind, or the check itself fails.
     """
     # pylint: disable=import-outside-toplevel
+    # Third-Party
     from alembic.runtime.migration import MigrationContext
     from alembic.script import ScriptDirectory
     import sqlalchemy as sa
@@ -784,10 +785,7 @@ async def main() -> None:
             with engine.connect() as conn:
                 conn.commit()
                 if not _is_at_alembic_head(conn, cfg):
-                    logger.warning(
-                        "SKIP_MIGRATION=true but schema is not at head "
-                        "(fresh DB or pending migrations) — running alembic upgrade head"
-                    )
+                    logger.warning("SKIP_MIGRATION=true but schema is not at head " "(fresh DB or pending migrations) — running alembic upgrade head")
                     with advisory_lock(conn):
                         cfg.attributes["connection"] = conn
                         command.upgrade(cfg, "head")
--- a/mcpgateway/bootstrap_db.py
+++ b/mcpgateway/bootstrap_db.py
@@ -776,12 +776,26 @@ async def main() -> None:
 
     # SKIP_MIGRATION: external tooling (init container, CI pipeline) already ran
     # alembic upgrade head — skip schema migration and run only the idempotent
-    # bootstrap helpers.
+    # bootstrap helpers.  However, if the schema is not yet at head (e.g. first
+    # boot on a fresh DB before any init container has run), fall through to
+    # migrations so the app does not crash with "no such table" errors.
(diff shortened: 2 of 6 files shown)


mcpgateway/bootstrap_db.py
```diff
@@ -96,6 +96,46 @@ def _schema_looks_current(inspector) -> bool:
...
+    Args:
+        conn: Active SQLAlchemy connection.
+        cfg: Alembic Config object with sqlalchemy.url already set.
+
+    Returns:
+        True when the database is already at head (safe to skip migration lock).
+        False when the database is fresh, behind, or the check itself fails.
+    """
+    # pylint: disable=import-outside-toplevel
+    from alembic.runtime.migration import MigrationContext
```
cafalchio: Could the imports be in the top of the file?

mcpgateway/bootstrap_db.py
```diff
@@ -96,6 +96,46 @@ def _schema_looks_current(inspector) -> bool:
     )
 
 
+def _is_at_alembic_head(conn: Connection, cfg: Config) -> bool:
```
cafalchio, line 100: Nice idea to check head.

mcpgateway/bootstrap_db.py
```diff
@@ -711,38 +758,96 @@ async def main() -> None:
...
+            gateway_replicas,
+            _MIGRATION_LOCK_PATH,
+        )
+
+    # SKIP_MIGRATION: external tooling (init container, CI pipeline) already ran
+    # alembic upgrade head — skip schema migration and run only the idempotent
+    # bootstrap helpers.  However, if the schema is not yet at head (e.g. first
+    # boot on a fresh DB before any init container has run), fall through to
+    # migrations so the app does not crash with "no such table" errors.
+    if settings.skip_migration:
```
cafalchio, line 784: This is a bit misleading. 
If we set a Skip migration in .env, we should error if db is not in the head. 

In this block, despite the SKIP_MIGRATION=true, we are applying migrations.

mcpgateway/config.py
```diff
@@ -250,6 +250,14 @@ class Settings(BaseSettings):
             "(See Issue #1535 for details)"
         ),
     )
+    skip_migration: bool = Field(
+        default=True,
```
cafalchio: Should it be True by default?

mcpgateway/bootstrap_db.py
```diff
@@ -711,38 +758,96 @@ async def main() -> None:
...
+                else:
+                    logger.info("SKIP_MIGRATION=true — schema already at head, skipping migration")
+                updated = normalize_team_visibility(conn)
+                if updated:
+                    logger.info(f"Normalized {updated} team record(s) to supported visibility values")
+                await bootstrap_admin_user(conn)
+                await bootstrap_default_roles(conn)
+                await bootstrap_resource_assignments(conn)
+                conn.commit()
+        except Exception as e:
```
cafalchio, line 799: Here could be migration error or Bootstrap error.

cafalchio (changes requested):
I tested, it is working for testing-up no problem.
The skip migration is a bit confusing. Perhaps it should be fully skipped.

mcpgateway/config.py
Lang-Akshay: kept the default to false to preserve the current setup.

mcpgateway/bootstrap_db.py
Lang-Akshay: implemented this by surfacing the error keeping fail fast principal. Instead of running migration will log error to user to run the migration if the SKIP_MIGRATION=true

mcpgateway/bootstrap_db.py
Lang-Akshay: Logging error and stopping instead of proceeding with wrong alembic head when SKIP_MIGRATION=true.

docker-compose.yml
```diff
@@ -1063,19 +1063,30 @@ services:
...
-  #     context: .
-  #     dockerfile: Containerfile.lite
-  #   environment:
-  #     - DATABASE_URL=postgresql+psycopg://postgres:${POSTGRES_PASSWORD:-mysecretpassword}@postgres:5432/mcp
-  #   command: alembic -c mcpgateway/alembic.ini upgrade head
-  #   depends_on:
-  #     postgres:
-  #       condition: service_healthy
-  #   networks: [mcpnet]
+  migration:
```
cafalchio, line 1066: Should this call "alembic mcpgateway/alembic.ini upgrade head"? Both way will work because bootstrap will create it if we do not pass the skip migration env.

cafalchio (dismissed):
Thanks for fixing issue. 
LGTM

madhu-mohan-jaishankar (approved):
Looks good. LGTM.

cafalchio (approved):
LGTM

brian-hussey (approved):
Approved based on others review

PR #4788, opened 2026-05-18
fix: admin users can now view and edit their own private A2A agents

### LINKED ISSUE #4694: [BUG]: Admin Users Cannot See Their Own Private Agents in List View
Discovered during: TC-A2A-005 testing (Private Agent Visibility) of issue #2406 

  Test Result: FAILED - Test 2 (Owner can see their own private agent)
  Tests Passed: 7/8

  Symptom

  Admin user with is_admin=true and token_teams=null:
  - ✅ Can create a private agent (201)
  - ✅ Can access their private agent via direct GET /a2a/{agent_id} (200)
  - ❌ Cannot see their private agent in list view GET /a2a (empty result)

  Root Cause

  File: mcpgateway/main.py, lines 4411-4413 in list_a2a_agents() endpoint

      user_email = None  # <-- BUG: This breaks owner matching
      token_teams = None  # Admin unrestricted

  When admin bypass is triggered, the code sets user_email = None to indicate "unrestricted" access. However, this breaks the PR #4341 invariant implemented in mcpgateway/services/base_service.py _apply_access_control() method (lines 109-115):

  if token_teams is None and user_email and is_user_admin(db, user_email):
      return query.where(
          or_(
              model_cls.visibility != "private",
              and_(model_cls.visibility == "private", model_cls.owner_email == user_email),
          )
      )

  This filter requires user_email to be set to match model_cls.owner_email == user_email and show the admin's OWN private agents. When user_email = None, the code falls through to lines 107-108 which exclude ALL private agents:

      return query.where(model_cls.visibility != "private")

  Expected Behavior (per PR #4341)

  Admin bypass should show:
  - ✅ All public agents
  - ✅ All team agents
  - ✅ Admin's OWN private agents
  - ❌ Other users' private agents (security invariant)
  
  Actual Behavior

  Admin bypass shows:
  - ✅ All public agents
  - ✅ All team agents
  - ❌ NO private agents (including admin's own)
  
  Proposed Fix

  File: mcpgateway/main.py, line 4412

  Remove user_email = None assignment. The token_teams = None is sufficient to trigger admin bypass in _apply_access_control(), and keeping user_email allows the owner matching to work correctly.

      # user_email stays set for owner matching on private agents (PR #4341)
      token_teams = None  # Admin unrestricted - sees all public+team agents

  Impact

  - Admin users cannot see their own private agents in list views
  - Direct GET access still works (uses different code path via _check_agent_access())
  - Affects all resources using the BaseService visibility filtering pattern (agents, potentially others)

  Test Evidence

  ==========================================
  Test 2: Owner can see their own private agent
  ==========================================

  ❌ Test 2 FAILED - Owner cannot see their own private agent
  Expected ID: ae3e9330611743a1a7231a8ee33ccb89
  Found ID:

rakdutta:
Closes #4694

## Problem

Admin users could not view and edit their own private A2A agents in the Admin UI.

The issue occurred because:

1. Admin endpoints (`admin_get_agent`, `admin_list_a2a_agents`) were incorrectly reading token teams from `user.get("token_teams")`, which always returned `None` since the JWT payload uses the key `teams`, not `token_teams`

2. The `_apply_access_control` logic in `base_service.py` incorrectly granted owner-match access to admin users with `token_teams=[]` (explicit public-only API tokens), violating Layer 1 token scoping

3. The DB admin check (`is_user_admin`) was being called on every authenticated request, causing unnecessary database queries

---

## Solution

This PR fixes the token team extraction and enforces proper Layer 1 token scoping while preserving the security invariant from PR #4341: admin bypass never reveals another user's private rows.

#### Fixed token team extraction in admin endpoints

- Added `request: Request` parameter to `admin_get_agent()` and `admin_list_a2a_agents()`
- Replaced `user.get("token_teams")` with `get_token_teams_from_request(request)`, which reads the already-normalized `request.state.token_teams`
- This ensures token teams are correctly extracted from the JWT payload

#### Enforced Layer 1 token scoping

- Removed the `not user_is_admin` guard that was allowing admin API tokens with `teams: []` to bypass public-only restrictions
- API tokens with `teams: []` now remain public-only even for admin users (this is Layer 1 token scoping, not RBAC)
- Session-token admins (who arrive with `token_teams=None`) continue to work correctly via the admin bypass path

#### Optimized DB admin checks

- Moved `is_user_admin(db, user_email)` calls inside the branches that actually need them
- Eliminated unnecessary DB queries for non-admin users and users on the admin bypass path

#### Added comprehensive unit tests

- Test for admin user with `token_teams=[]` → verifies they don't see another user's private rows
- Test for admin user with `token_teams=["t1"]` → verifies they see their own private resources
- Updated existing admin endpoint tests to include the new `request` parameter

---

## Security Considerations

- ✅ Preserves PR #4341 security invariant: admin bypass never reveals another user's private rows
- ✅ Enforces Layer 1 token scoping: API tokens with explicit empty team lists are honored
- ✅ No changes to RBAC (Layer 2) — all permission checks remain unchanged
- ✅ Session-token admins maintain full admin bypass capabilities via `token_teams=None`

---

## Testing

- All unit tests pass (`make test`)
- Manual testing confirms admin users can now view/edit their own private A2A agents
- Verified that public-only token scoping is enforced correctly for admin API tokens
- Verified that non-admin users are not affected by the changes

---

## Access Behavior Matrix

| Token Type | User Type | `teams` Value | Can See Own Private? | Can See Other Private? |
|------------|------------|----------------|-----------------------|-------------------------|
| API | Admin | `[]` | ❌ NO (public-only) | ❌ NO |
| API | Admin | `null` | ✅ YES | ❌ NO (PR #4341) |
| API | Admin | `["team-1"]` | ✅ YES | ❌ NO |
| API | User | `[]` | ❌ NO (public-only) | ❌ NO |
| API | User | `["team-1"]` | ✅ YES | ❌ NO |

---

### DIFF
--- a/.github/workflows/notify-slack-failure.yml
+++ b/.github/workflows/notify-slack-failure.yml
@@ -68,7 +68,12 @@ jobs:
                   "type": "section",
                   "text": {
                     "type": "mrkdwn",
-                    "text": ":x: *${{ github.event.workflow_run.name || 'Manual test dispatch' }}* failed on `main`\n• *Commit:* <${{ github.server_url }}/${{ github.repository }}/commit/${{ github.event.workflow_run.head_sha || github.sha }}|`${{ github.event.workflow_run.head_sha || github.sha }}`>\n• *Author:* `${{ github.event.workflow_run.head_commit.author.name || github.actor }}`\n• *Merged by:* `${{ github.event.workflow_run.actor.login || github.actor }}`\n• *Run:* <${{ github.event.workflow_run.html_url || format('{0}/{1}/actions/runs/{2}', github.server_url, github.repository, github.run_id) }}|View run>"
+                    "text": >-
+                      :x: *${{ github.event.workflow_run.name || 'Manual test dispatch' }}* failed on `main`
+                      • *Commit:* <${{ github.server_url }}/${{ github.repository }}/commit/${{ github.event.workflow_run.head_sha || github.sha }}|`${{ github.event.workflow_run.head_sha || github.sha }}`>
+                      • *Author:* `${{ github.event.workflow_run.head_commit.author.name || github.actor }}`
+                      • *Merged by:* `${{ github.event.workflow_run.actor.login || github.actor }}`
+                      • *Run:* <${{ github.event.workflow_run.html_url || format('{0}/{1}/actions/runs/{2}', github.server_url, github.repository, github.run_id) }}|View run>
                   }
                 }
               ]
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-05-22T15:37:37Z",
+  "generated_at": "2026-05-22T17:48:13Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -5098,7 +5098,7 @@
         "hashed_secret": "995e87a207ec41ef95f938dcf53a4184312b0d93",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 10644,
+        "line_number": 10661,
         "type": "Hex High Entropy String",
         "verified_result": null
       }
--- a/mcpgateway/admin.py
+++ b/mcpgateway/admin.py
@@ -15534,7 +15534,7 @@ async def admin_get_agent(
         raise HTTPException(status_code=404, detail=str(e))
     except Exception as e:
         LOGGER.error(f"Error getting agent {agent_id}: {e}")
-        raise e
+        raise
 
 
 @admin_router.get("/a2a", response_model=PaginatedResponse)
--- a/mcpgateway/admin.py
+++ b/mcpgateway/admin.py
@@ -67,7 +67,7 @@ from mcpgateway import version as version_module
 from mcpgateway.auth import get_current_user, get_user_team_roles
 
 # Re-export canonical get_user_email from auth_context for backward compatibility.
-from mcpgateway.auth_context import get_scoped_resource_access_context, get_user_email
+from mcpgateway.auth_context import get_scoped_resource_access_context, get_token_teams_from_request, get_user_email
 from mcpgateway.cache.a2a_stats_cache import a2a_stats_cache
 from mcpgateway.cache.global_config_cache import global_config_cache
 from mcpgateway.common.models import LogLevel
@@ -15498,6 +15498,7 @@ async def admin_list_import_statuses(user=Depends(get_current_user_with_permissi
 @require_permission("a2a.read", allow_admin_bypass=False)
 async def admin_get_agent(
     agent_id: str,
+    request: Request,
     db: Session = Depends(get_db),
     user=Depends(get_current_user_with_permissions),
 ) -> Dict[str, Any]:
@@ -15505,6 +15506,7 @@ async def admin_get_agent(
 
     Args:
         agent_id: Agent ID.
+        request: FastAPI request object.
         db: Database session.
         user: Authenticated user.
 
@@ -15523,7 +15525,7 @@ async def admin_get_agent(
     """
     LOGGER.debug(f"User {get_user_email(user)} requested details for agent ID {agent_id}")
     user_email = get_user_email(user)
-    token_teams = user.get("token_teams")
+    token_teams = get_token_teams_from_request(request)
 
     try:
         agent = await a2a_service.get_agent(db, agent_id, user_email=user_email, token_teams=token_teams)
@@ -15541,6 +15543,7 @@ async def admin_list_a2a_agents(
     page: int = Query(1, ge=1, description="Page number (1-indexed)"),
     per_page: int = Query(settings.pagination_default_page_size, ge=1, le=settings.pagination_max_page_size, description="Items per page"),
     include_inactive: bool = False,
+    request: Request = None,
     db: Session = Depends(get_db),
     user=Depends(get_current_user_with_permissions),
 ) -> Dict[str, Any]:
@@ -15555,6 +15558,7 @@ async def admin_list_a2a_agents(
         page (int): Page number (1-indexed) for offset pagination.
         per_page (int): Number of items per page.
         include_inactive (bool): Whether to include inactive agents in the results.
+        request (Request): FastAPI request object.
         db (Session): Database session dependency.
         user (dict): Authenticated user dependency.
 
@@ -15585,7 +15589,7 @@ async def admin_list_a2a_agents(
 
     LOGGER.debug(f"User {get_user_email(user)} requested A2A Agent list (page={page}, per_page={per_page})")
     user_email = get_user_email(user)
-    token_teams = user.get("token_teams")
+    token_teams = get_token_teams_from_request(request)
 
     # Call a2a_service.list_agents with page-based pagination
     paginated_result = await a2a_service.list_agents(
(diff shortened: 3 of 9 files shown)


msureshkumar88 (changes requested):
Good direction on the fix — the core insight (preserve `user_email` through the admin bypass instead of nulling it) is correct. A few issues need addressing before merge.

---

**`mcpgateway/admin.py` — wrong dict key (blocking)**

`admin_get_agent` L15501 and `admin_list_a2a_agents` L15562:
```python
token_teams = user.get("token_teams")
```
The JWT payload dict uses `teams`, not `token_teams`. `user.get("token_teams")` always returns `None`. This silently gives every admin the `token_teams=None` admin-bypass path regardless of their token's actual scope. An API token minted with `teams: []` (explicit public-only scope) should be restricted to public resources only per the AGENTS.md security table (`teams: [] + is_admin: true → PUBLIC-ONLY`), but here it gets admin bypass instead.

Fix: use `get_token_teams_from_request(request)` (already imported in `auth_context.py`) — it reads `request.state.token_teams` which auth middleware has already normalized via `normalize_token_teams` / `resolve_session_teams`. The admin endpoint signatures need to accept `request: Request` if they don't already.

---

**`mcpgateway/services/base_service.py` — token scope invariant broken (blocking)**

L133 (new `filter_email` line):
```python
filter_email = None if (token_teams is not None and not token_teams and not user_is_admin) else user_email
```
This grants owner-match access to admin users whose `token_teams == []`. Per AGENTS.md: session-token admins always arrive with `token_teams=None` (from `resolve_session_teams` — DB admin → `None`), so `token_teams=[]` for an admin only arises from an API token explicitly scoped to empty teams. That token is deliberately limited and should stay public-only. The PR description says "preserving owner access for admins with public-only or team-scoped session tokens" but session-token admins never have `token_teams=[]` — so this condition only fires for explicitly-scoped-down API tokens, which is the wrong behaviour.

Fix: remove the `not user_is_admin` guard from this line. Session-token admins don't need it (they arrive at `token_teams=None`, hit the early-return branch above). API-token admins with `teams: []` should remain public-only.

---

**`mcpgateway/services/base_service.py` — unconditional DB admin check (risk)**

L111:
```python
```
`is_user_admin` is now called on every authenticated list request. Previously it was only reached when `token_teams is None`. This adds a DB query per request for every non-anonymous call, including non-admin users who will always get `False`. Move this call inside the branches that actually need it (or gate it with the existing `token_teams is None` check as before, since non-`None` token_teams admin users are the only new case and that's covered by the flawed `filter_email` line above).

---

**`mcpgateway/services/base_service.py` — missing unit tests for new path**

The `_apply_access_control` changes introduce a new code path (admin + `token_teams=[]`), but no tests in `test_base_service.py` cover it. The updated tests in `test_main.py` / `test_main_extended.py` mock `get_rpc_filter_context` at the endpoint level and don't exercise the service layer logic. Please add unit tests for:
- Admin user + `token_teams=[]` → should NOT see another user's private rows
- Admin user + `token_teams=["t1"]` + own private resource → should see own private rows

---

**PR template incomplete (nit)**

The PR body still has `Closes #` (empty issue link), `_What does this PR do and why?_` placeholder, and unchecked checklist items. The actual issue being closed is #4694 (mentioned in the commit message).

rakdutta:
@msureshkumar88 Thanks for the review. I have addressed all the review comments and updated the PR accordingly. Please review and merge.

msureshkumar88 (changes requested):
Good progress — both prior blocking items are correctly addressed:

- **B1 resolved** ✅ — `get_token_teams_from_request(request)` now used in both `admin_get_agent` and `admin_list_a2a_agents`; the wrong `user.get(\"token_teams\")` key is gone.
- **B2 resolved** ✅ — `filter_email` condition simplified to `token_teams is not None and not token_teams`; the `not user_is_admin` guard that violated Layer 1 token scoping for public-only tokens is removed.

A few items need fixing before merge.

---

### Blocking

**CI: pre-commit hooks failing**

Run [26210158334](https://github.com/IBM/mcp-context-forge/actions/runs/26210158334) shows `Run pre-commit hooks` failing on the latest push. All other checks pass. Must be green before merge.

---

**`mcpgateway/services/base_service.py` — dead DB call, result never used (performance regression)**

```python
# Line 128-132
effective_teams: List[str] = []
user_is_admin = False
if token_teams is not None:
    effective_teams = token_teams
    user_is_admin = user_email and is_user_admin(db, user_email)  # ← result never used
```

`user_is_admin` is assigned here but is never referenced again — it is not passed to `_apply_visibility_filter`, and the `filter_email` expression at L141 does not use it. This `is_user_admin` call fires a DB round-trip on **every** session-token and every public-only token request, with the result thrown away.

The `not user_is_admin` guard was correctly removed from `filter_email` to fix B2, but the call itself was left behind. The PR description says it reduces unnecessary DB admin checks — this code does the opposite.

Fix: remove `user_is_admin = False` (L128) and the `is_user_admin(db, user_email)` call (L132) from this branch entirely.

---

**`mcpgateway/services/base_service.py` — docstring contradicts the access-control contract**

The updated docstring at L75-77 says:

> DB-resolved admin (email provided) sees public + team rows + their OWN private rows, **regardless of token_teams value**.

This is incorrect. Public-only tokens (`token_teams=[]`) suppress owner access per Layer 1 token scoping — that is specifically what B2 enforces. A statement saying the contrary in the authoritative docstring is a correctness hazard for anyone reasoning about the access control flow here.

Fix: add the exception — e.g. _"except when `token_teams=[]`, where Layer 1 token scoping suppresses owner access even for admin users"_.

---

### Non-blocking

**`mcpgateway/admin.py` — `raise e` loses original traceback**

```python
except Exception as e:
    LOGGER.error(f"Error getting agent {agent_id}: {e}")
    raise e  # ← strips traceback
```

Use bare `raise` to preserve the original stack.

---

**`tests/unit/mcpgateway/test_admin.py` — `_test_admin_list_a2a_agents_enabled` is a dead test**

The method name starts with `_` — pytest does not collect it. The PR correctly updated this method to accept `mock_request`, but the underscore prefix means the update exercises nothing. This was pre-existing, but since the PR touched it, worth fixing now: rename to `test_admin_list_a2a_agents_enabled` or delete it.

---

### Test coverage gaps

**Missing: happy-path test for the original bug scenario**

There is no test that verifies an admin user with `token_teams=None` calling `admin_get_agent(agent_id)` for their **own** private agent returns the agent. The tests in `test_base_service.py` cover the filter logic, but nothing in `test_admin.py` covers the end-to-end `admin_get_agent` success path for the scenario in issue #4694.

**Missing: deny-path test for `admin_get_agent` with `token_teams=[]`**

There is no test asserting that an admin user with a public-only token (`token_teams=[]`) cannot retrieve another user's private agent via `admin_get_agent`. Per AGENTS.md: `teams: [] + is_admin: true → PUBLIC-ONLY`. A deny-path test here is the enforcement witness for Layer 1 scoping on this endpoint.

---

### Nit

`base_service.py:L141` — `token_teams is not None and not token_teams` is hard to parse at a glance. `token_teams == []` is equivalent (given `Optional[List[str]]`) and states the intent directly.

---

**PR template**: `Closes #` is empty — should be `Closes #4694`. Type of Change checkboxes are unchecked.

msureshkumar88 (dismissed):
All blocking findings from rounds 1 and 2 are resolved — approving.

**Round 1 findings — all fixed:**
- **B1** ✅ `admin_get_agent` / `admin_list_a2a_agents` now use `get_token_teams_from_request(request)` instead of the wrong `user.get("token_teams")` key
- **B2** ✅ `filter_email` simplified to `token_teams is not None and not token_teams`; the `not user_is_admin` guard that violated Layer 1 token scoping for public-only API tokens is removed

**Round 2 findings — all fixed:**
- ✅ Dead DB call removed — `is_user_admin` now called only inside the `token_teams is None` branch where actually needed; no longer fires on every authenticated request
- ✅ Exception handling corrected — `raise e` → `raise` preserves original traceback
- ✅ Token scope unit tests added (`token_teams=[]` admin cannot see other user's private rows, `token_teams=["t1"]` admin sees own private resources)

**One optional nit remaining (non-blocking):**
- `base_service.py` docstring L75-77 still says admin sees own private rows *regardless of token_teams value* — contradicts the Layer 1 scoping that B2 enforces. Worth a follow-up to add the exception clause so the docstring accurately describes the contract.

Issue #4755, opened 2026-05-14
[CHORE]: Consolidate E2E test helpers and JWT token generation


prakhar-singh1928:
### 🔧 Chore Summary

Consolidate duplicate JWT token generation and authentication helper functions across end-to-end tests into a centralized test helper module.

---

### 🧱 Area Affected

- [x] Testing infrastructure
- [x] Code organization / DRY principle
- [x] Test maintainability

---

### ⚙️ Context / Rationale

**Current State:**
The codebase has **30+ duplicate implementations** of JWT token generation and authentication helpers scattered across test files:

**Duplicate Functions Found:**
- `create_test_jwt_token()` - 3+ implementations
- `generate_test_jwt()` - 3+ implementations  
- `_make_jwt()` - 15+ implementations
- `TEST_JWT_TOKEN` / `TEST_AUTH_HEADER` - 10+ definitions

**Affected Test Directories:**
- `tests/e2e/` (4 files)
- `tests/playwright/` (10+ files across security/, teams/, operations/, entities/)
- `tests/live_gateway/` (5+ files)
- `tests/integration/` (1 file)
- `tests/loadtest/` (1 file)
- `tests/unit/` (2 files)

**Problems:**
1. **Maintenance burden**: Changes to JWT structure require updates in 30+ locations
2. **Inconsistency**: Different implementations have subtle variations in claims, expiry, and defaults
3. **Test fragility**: Hard to ensure all tests use correct JWT format
4. **Violates DRY**: Same logic duplicated across codebase

**Benefits of Consolidation:**
- Single source of truth for test JWT generation
- Easier to update JWT structure globally
- Consistent test authentication across all test suites
- Reduced code duplication (~500+ lines)
- Improved test maintainability

---

### 📦 Related Make Targets

- `make test` - run all tests
- `make test-e2e` - run end-to-end tests
- `make test-playwright` - run Playwright tests
- `make lint` - ensure code quality

---

### 📋 Acceptance Criteria

- [ ] Create `tests/helpers/jwt_helpers.py` with centralized functions:
  - `create_test_jwt(email, is_admin=False, teams=None, exp=None, scopes=None)`
  - `create_admin_jwt(email='admin@example.com')`
  - `create_user_jwt(email, teams=None)`
  - `create_auth_header(token)` or `create_auth_headers(token)`
  
- [ ] Update all test files to import from centralized helper:
  - `tests/e2e/test_admin_apis.py`
[...]

PR #4776, opened 2026-05-15
fix(tools): Add CSP nonce support to OAuth callback page

bogdanmariusc10:
## 🔗 Related Issue
Closes #4774
Complements #4424 and #4673 that introduced CSP security headers but caused a regression for OAuth MCP server tool fetching

---

## 📝 Summary

This PR fixes a critical regression where OAuth-configured MCP servers (ClickUp, Gamma, Granola, Notion) could not fetch tools after successful authorization. The root cause was a Content Security Policy (CSP) violation in the OAuth callback success page.

**Problem**: The OAuth callback page at `mcpgateway/routers/oauth_router.py` contained inline JavaScript without a CSP nonce, causing browsers to block script execution. This prevented the "Fetch Tools from MCP Server" button from working.

**Solution**: 
- Added CSP nonce support to the inline script tag using `get_csp_nonce_from_request(request)`
- Removed inline `onclick` handler and replaced with proper `addEventListener` pattern
- Wrapped script in IIFE (Immediately-Invoked Function Expression) for proper scoping
- Created comprehensive integration tests to prevent future CSP regressions

**Impact**: OAuth authorization flow now works correctly with CSP enabled, allowing users to fetch tools from officially hosted MCP servers after successful OAuth authorization.

---

## 🏷️ Type of Change
- [x] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Chore (deps, CI, tooling)
- [ ] Other (describe below)

---

## 🧪 Verification

| Check                     | Command         | Status |
|---------------------------|-----------------|--------|
| Lint suite                | `make lint`     | ✅ Pass |
| Unit tests                | `make test`     | ✅ Pass |
| Coverage ≥ 80%            | `make coverage` | ✅ Pass |

---

## ✅ Checklist
- [x] Code formatted (`make black isort pre-commit`)
- [x] Tests added/updated for changes
- [x] Documentation updated (if applicable)
- [x] No secrets or credentials committed

---

### Technical Details

**Files Changed**:
1. `mcpgateway/routers/oauth_router.py` (lines 657-759)
   - Added `get_csp_nonce_from_request(request)` import and usage
   - Applied nonce to script tag: `<script nonce="{csp_nonce}">`
   - Removed inline onclick: `<button id="fetch-tools-btn" class="button">`
   - Added event listener: `button.addEventListener('click', async function() {...})`

2. `tests/security/test_oauth_callback_csp_compliance.py` (new file)
[...]

PR #4705, opened 2026-05-11
feat(observability): add baggage attribute mapping for span customization

vishu-bh:
## Problem

Issue #4700 requires selected tenant/user context carried through OpenTelemetry baggage to appear in exported spans without the `baggage.` prefix. Clients can send values such as `tenant.id` and `user.id` as OTEL baggage, but some observability backends expect those dimensions as span attributes named `tenant.id` / `user.id`, not `baggage.tenant.id` / `baggage.user.id`.

The gateway still needs to stay aligned with OTEL semantics: baggage remains the propagation mechanism, while span attributes are the reporting surface.

## Solution

This PR replaces the earlier span-processor remapping approach with a baggage emission policy that is applied where baggage is copied onto spans.

The `SpanAttributeCustomizer` plugin now supports:

```yaml
config:
  allowed_baggage_span_attributes:
    - "tenant.id"
    - "user.id"
  emit_baggage_prefixed_attributes: false
```

With that configuration:

- OTEL baggage continues to carry `tenant.id` and `user.id`
- only allowlisted baggage keys are promoted to span attributes
- exported span attributes are emitted as `tenant.id` / `user.id`
- `baggage.tenant.id` / `baggage.user.id` are not emitted for those allowlisted keys

If the new config is not provided, the gateway preserves legacy behavior and emits baggage as `baggage.<key>`.

## Changes

- `mcpgateway/observability.py`
  - Added `BaggageSpanAttributePolicy`
  - Added startup extraction for the SpanAttributeCustomizer baggage policy
  - Applies the policy directly in both request-root span baggage injection and `create_span()`
  - Removed the old `BaggageAttributeSpanProcessor` remapping path

- `mcpgateway/main.py`
  - Loads the baggage span attribute policy after plugin config is available and before telemetry is initialized

- `plugins/span_attribute_customizer/config_schema.py`
  - Added `allowed_baggage_span_attributes`
  - Added `emit_baggage_prefixed_attributes`
  - Added validation/normalization for the baggage allowlist

- `plugins/config.yaml`
  - Added ready-to-use baggage policy config for `tenant.id` and `user.id`

- Docs
  - Updated OpenTelemetry architecture docs
  - Updated span attribute reference docs
  - Updated SpanAttributeCustomizer docs and README

- Tests
  - Added coverage for policy extraction, malformed config handling, allowlist behavior, default legacy behavior, and both baggage-to-span injection paths

## Testing

Validated locally:

```bash
make pre-commit
```

[...]

PR #4796, opened 2026-05-18
chore: test cleanup

prakhar-singh1928:
## 🔗 Related Issue
Partially Closes [#[<issue-number-if-applicable>](https://github.com/IBM/mcp-context-forge/issues/4755)](https://github.com/IBM/mcp-context-forge/issues/4755)

---

## 📝 Summary

Consolidates JWT token creation across 38+ test files into centralized helper functions, eliminating ~300 lines of duplicated code and improving test maintainability.

**Key Changes:**
- Created `tests/helpers/auth.py` with `make_test_jwt()` and `make_legacy_test_jwt()` functions
- Created `tests/helpers/api_helpers.py` with `ApiTestHelper` class for high-level API operations
- Migrated all test files to use common JWT helpers instead of inline token creation
- Fixed CI Playwright tests by disabling rate limiting to prevent account lockouts

---

## 🏷️ Type of Change
- [ ] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [x] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

| Playwright smoke          | `make test-ui-ci-smoke` | ✅ |

---

## ✅ Checklist

---

### Dual JWT Creation Strategy

The consolidation implements two helper functions to support different test scenarios:

1. **`make_test_jwt()`** - Uses production `_create_jwt_token()` for consistency with real token generation
2. **`make_legacy_test_jwt()`** - Direct JWT encoding for edge cases and legacy compatibility

### CI Infrastructure Fixes

Added rate limiting configuration to prevent test failures:
- `.github/workflows/playwright.yml` - Disabled rate limiting in CI environment
- `.env.example` - Documented rate limiting settings for local development
- `.env` - Configured for local testing (not committed to repo)

### Files Changed

**New Files:**
- `tests/helpers/auth.py` (157 lines) - JWT token creation helpers
- `tests/helpers/api_helpers.py` (176 lines) - High-level API test utilities

**Modified Files:**
- 38+ test files migrated to use common helpers
- `.github/workflows/playwright.yml` - CI environment configuration
- `.env.example` - Rate limiting documentation

### Benefits

- **Maintainability**: Single source of truth for JWT creation logic
- **Consistency**: All tests use the same token generation approach
- **Testability**: Easier to update JWT structure across all tests
- **Reduced Duplication**: Eliminated ~300 lines of repeated code
- **CI Stability**: Fixed Playwright test failures caused by rate limiting


gandhipratik203 (commented):
Thanks for taking this on, Prakhar — JWT setup duplication has been a thorn in this repo for a while, and the new `tests/helpers/auth.py` is a really nice surface to standardize against. A few thoughts below.

## What's Good

- **Clear motivation and dual-helper rationale** — the PR description explains why both `make_test_jwt` and `make_legacy_test_jwt` exist, which makes the design intent easy to follow.
- **Helper shape is thoughtful** — the `_UNSET` sentinel for `teams` is a nice touch, since it preserves the admin-bypass semantics where `teams=None` is meaningful.
- **`tests/AGENTS.md` updated** — new contributors will land in the right place.
- **Pragmatic CI fix** — disabling rate limiting in the Playwright workflow unblocks a real flakiness source.
- **`ApiTestHelper` removes a lot of boilerplate** — the create/delete entity methods read much cleaner in the migrated test files.

## Findings

A few things worth a look before this lands.

### F1 — The branch has drifted from `main` and is reverting recent merges

Diffed against `origin/main`, the PR touches ~66 files (~1,098 insertions / ~2,154 deletions), which is well beyond what the description implies. Looking through it, several recent main merges appear to get rolled back:

- `mcpgateway/bootstrap_db.py` — the `_is_at_alembic_head` / `MigrationContext` logic from #4784 is gone.
- `mcpgateway/observability.py` and `tests/e2e/test_baggage_tracing.py` (deleted) — #4705 (baggage attribute mapping) gets reverted.
- `mcpgateway/routers/oauth_router.py` — the CSP-nonce wiring from #4776 is dropped.
- Unit tests deleted: `test_observability.py`, `test_oauth_router.py`, `test_bootstrap_db.py`, `test_span_attribute_customizer_mapping.py`.
- Containerfiles, Makefile, `docker-compose.yml`, `pyproject.toml`, `plugins/config.yaml` — all show drift unrelated to test cleanup.

Looks like the `Merge branch 'main' into dev_test_cleanup` commit resolved a few files the wrong way. Would you mind rebasing onto current `main` and keeping the upstream side on those conflicts? Once that's clean, the diff should shrink down to just the test-helper changes.

### F2 — `make_test_jwt(is_admin=True)` doesn't actually mark the token as admin

In `tests/helpers/auth.py`, `is_admin` only gets embedded when the caller also passes `include_user_data=True` — that's what builds the `user_data` dict carrying the claim.

A lot of migrated callsites do:

```python
make_test_jwt(email, is_admin=is_admin, teams=teams)
```

…where previously they did:

```python
_create_jwt_token(
    {"sub": email},
    user_data={"email": email, "is_admin": is_admin, "auth_provider": "local"},
    teams=teams,
)
```

Without `include_user_data=True`, the `user.is_admin` claim is dropped, and per the `normalize_token_teams()` semantics, `teams: null` + missing `is_admin` resolves to **PUBLIC-ONLY** instead of **ADMIN BYPASS**. So admin-path tests may quietly be running against the public-only code path and "passing" for the wrong reason.

Probably the cleanest fix is to default `include_user_data=True` whenever `is_admin=True` is passed — that matches how every prior callsite was using it. Either way, this one really needs to be addressed before merge since it changes test semantics across the suite.

### F3 — `extra_payload` in `tests/playwright/conftest.py` lands at the wrong nesting level

In the migrated `api_request_context` fixture, `extra_payload={\"full_name\": \"Test Admin\"}` is passed in. `extra_payload` merges into the top-level JWT payload, but `full_name` previously lived inside `user_data`. The token shape changes as a result — would you mind routing this back through `user_data` so the fixture keeps producing the same token structure?

### F4 — Two helpers do the same Playwright-context job

`tests/helpers/auth.py::make_playwright_api_context` and `tests/helpers/api_helpers.py::ApiTestHelper.new_context` are essentially identical, and different files use different ones. Since the whole point of this PR is consolidation, it'd be good to land with one canonical entrypoint (probably `ApiTestHelper.new_context`) and have the other delegate or be removed. Otherwise the next test author has to flip a coin.

### F5 — `ApiTestHelper.create_tool` defaults to a real external URL

`create_tool` defaults `url=\"https://httpbin.org/post\"`. Any test that takes the default will hit the public internet, which breaks offline and air-gapped CI and creates a flakiness vector tied to a third-party service. Could we make `url` a required argument, or default it to something local (or a clearly marked placeholder that the tool registration will reject)?

While there: the trailing-slash inconsistency across endpoints (`/teams/`, `/tools/`, `/resources/` vs `/servers`, `/gateways`) can produce 307 redirects depending on FastAPI routing. Worth aligning with the actual route definitions.

### F6 — `# Made with Bob` trailer in both new helper files

`auth.py` and `api_helpers.py` both end with `# Made with Bob`. Repo guidelines are pretty firm about no attribution-style trailers landing in diffs — please drop those before merge.

### F7 — `.env.example` has unrelated content in the diff

The `.env.example` diff includes a UAID security warning block being removed plus CSRF/SIEM/`GATEWAY_TEST_ALLOW_REGISTERED_ONLY` sections being reshuffled. None of that is test cleanup, and removing the UAID warning in particular weakens security guidance for ops. This should clear up with the F1 rebase, but worth a careful look at the resulting `.env.example` diff after rebase to make sure only the rate-limit doc note remains.

### F8 — `.secrets.baseline` churn

`.secrets.baseline` has ~100 lines of diff. Likely another F1 rebase artifact, but please run `detect-secrets scan --baseline .secrets.baseline` after the rebase to confirm nothing new crept in alongside the `.env.example` edits.

### F9 — `tests/AGENTS.md` doesn't mention `make_legacy_test_jwt`

The PR description treats the dual-helper strategy as deliberate, but the AGENTS.md update only mentions `make_test_jwt`. Since this PR's stated value is partly \"future contributors know what to use,\" would you mind adding a one-liner on when to reach for `make_legacy_test_jwt` (raw JWT for legacy claim shapes / edge cases)?

### F10 — CI rate-limit disable could use a quick comment

The `RATE_LIMITING_ENABLED: \"false\"` / `RATE_LIMIT_LOCKOUT_ENABLED: \"false\"` env vars in `playwright.yml` will fix the lockouts. A short inline comment (\"disabled to avoid lockouts during the shared admin login flow used by smoke tests\") would help the next person who reads this workflow.

---

Once F1–F8 are sorted I think we're in good shape — F9 and F10 are documentation polish that could come along or land as a quick follow-up, whichever's easier for you. Happy to take another look once the rebase and the `is_admin` fix are in. Thanks again for picking this up!

gandhipratik203 (changes requested):
Once the changes for the findings above are done, this should be good to merge. Happy to re-review.

gandhipratik203:
Thanks for the quick turnaround, Prakhar. Everything else looks good. One small thing: the trailing slashes in `ApiTestHelper` are a bit inconsistent. `/teams/`, `/tools/`, and `/resources/` have them, but `/servers` and `/gateways` don't. Could you line those up with whatever the routers expect? Otherwise we might end up bouncing through 307 redirects.

gandhipratik203 (dismissed):
Thanks for the rebase and slash cleanup, Prakhar. This looks good to go from my side. Tiny optional thing if you happen to be in the file already: `/tools/` on line 109 of `tests/helpers/api_helpers.py` still has a trailing slash while the others lost theirs. Totally fine to leave or fix, either works. Approving.

gandhipratik203 (dismissed):
Thanks for the change Prakhar! LGTM.

brian-hussey (dismissed):
Approved.

PR #4798, opened 2026-05-18
format: fixup autoflake, isort, black, ruff &  pre-commit

Lang-Akshay:
Format fixup - autoflake , isort, black, ruff && pre-commit.


cafalchio (approved):
Thanks for fixing the format

msureshkumar88 (approved):
Formatting-only pass — Black, isort, autoflake, ruff output looks correct.

A few notes for the record:
- `b77ca9d2de7e`: `pass` removed from `downgrade()` leaving just the docstring — functionally equivalent, Python accepts a docstring as a valid function body.
- `.secrets.baseline` timestamp bump is a `detect-secrets scan` side effect, not a format change — fine to carry along.
- All `op.create_table()` reformatting, single→double quotes, and trailing commas are pure Black output with no logic changes.

Approved.

PR #4809, opened 2026-05-19
Add Redis TLS support for production deployments

Lang-Akshay:
Closes https://github.ibm.com/WatsonOrchestrate/wo-tracker/issues/68461

## What this does

Adds opt-in TLS support for Redis connections. Local dev is unchanged — all new settings default to off, plain `redis://` URLs continue to work. Production deployments can switch to `rediss://` with full certificate control.

## Changes

**Python**

| File | Change |
|------|--------|
| `mcpgateway/config.py` | Five new Pydantic settings: `REDIS_SSL`, `REDIS_SSL_CA_CERTS`, `REDIS_SSL_CERTFILE`, `REDIS_SSL_KEYFILE`, `REDIS_SSL_CHECK_HOSTNAME` |
| `mcpgateway/utils/redis_client.py` | `_build_ssl_context()` builds an `ssl.SSLContext` when `REDIS_SSL=true`; returns `None` (zero overhead) otherwise |
| `mcpgateway/utils/redis_isready.py` | Startup probe accepts `ssl_context` and forwards it to `Redis.from_url()` |
| `mcpgateway/main.py` | Passes SSL context to `wait_for_redis_ready()` so probe and live client share the same TLS config |

**Configuration**

| File | Change |
|------|--------|
| `.env.example` | Commented-out Redis TLS block documenting the new variables |
| `docker-compose.yml` | Commented TLS env vars on the app service; commented production Redis command block (port 6380, cert flags) alongside the default plain-TCP command |

## New settings (all default to off)

| Variable | Default | Description |
|----------|---------|-------------|
| `REDIS_SSL` | `false` | Master TLS switch |
| `REDIS_SSL_CA_CERTS` | — | Path to CA bundle for server-cert verification |
| `REDIS_SSL_CERTFILE` | — | Client certificate (mTLS only) |
| `REDIS_SSL_KEYFILE` | — | Client private key (mTLS only) |
| `REDIS_SSL_CHECK_HOSTNAME` | `true` | Hostname verification (disable only for self-signed certs in dev) |

### 1. Install mkcert and generate certs

```bash
brew install mkcert      # macOS; Linux: https://github.com/FiloSottile/mkcert
mkcert -install          # installs local CA into system trust store

mkdir -p certs && cd certs
mkcert -cert-file server.crt -key-file server.key localhost 127.0.0.1
cp "$(mkcert -CAROOT)/rootCA.pem" ca.crt
cd ..
```

### 2. Start Redis with TLS

```bash
redis-server \
  --tls-port 6380 \
  --port 0 \
  --tls-cert-file $(pwd)/certs/server.crt \
  --tls-key-file  $(pwd)/certs/server.key \
  --tls-ca-cert-file $(pwd)/certs/ca.crt \
  --tls-auth-clients no
```

> `--tls-auth-clients no` disables client-cert requirement (mTLS off).

### 3. Configure the app

In `.env`:

```env
REDIS_URL=rediss://localhost:6380/0
REDIS_SSL=true
REDIS_SSL_CA_CERTS=./certs/ca.crt
```

### 4. Run and verify

```bash
make dev
# Startup log should show: "Redis is ready"
# No SSL errors → TLS is working
```

### 5. Verify plain-TCP still works (regression check)

Restore the original `.env` values (`redis://localhost:6379/0`, `REDIS_SSL=false`) and confirm the app starts cleanly.

---

### Option B — Full stack via docker-compose (`make compose-up` / `make testing-up`)

The `docker-compose.yml` already has all TLS knobs commented out. Swap the comment markers as shown below — no other edits needed.

#### 1. Generate certs (same as above)

```bash
brew install mkcert && mkcert -install
mkdir -p certs && cd certs
mkcert -cert-file server.crt -key-file server.key localhost 127.0.0.1
cp "$(mkcert -CAROOT)/rootCA.pem" ca.crt
cd ..
```

#### 2. In `docker-compose.yml` — app service environment

```yaml
# Comment out plain URL:
# - REDIS_URL=redis://redis:6379/0

# Uncomment TLS URL + settings:
- REDIS_URL=rediss://redis:6380/0
- REDIS_SSL=true
- REDIS_SSL_CA_CERTS=/app/certs/ca.crt
- REDIS_SSL_CHECK_HOSTNAME=false   # self-signed cert
```

Also uncomment the cert volume mount in the **gateway** service volumes block:

```yaml
- ./certs:/app/certs:ro
```

#### 3. In `docker-compose.yml` — redis service

Comment out the plain-TCP `command:` block and uncomment the TLS block (search for `TLS (disabled)`):

```yaml
# volumes:                           # ← uncomment
#   - ./certs:/certs:ro              # ← uncomment

# command:                           # ← uncomment the full TLS command block
#   - "redis-server"
#   - "--tls-port"
#   - "6380"
#   ...
```

Also swap the exposed port:

```yaml
# - "6379:6379"       # comment out plain TCP
- "6380:6380"         # uncomment TLS port
```

#### 4. Start the stack

```bash
make compose-up          # multi-instance gateway stack
# — or —
make testing-up          # gateway + Locust + A2A echo (load-test stack)
```

#### 5. Verify

```bash
# Redis healthcheck uses TLS automatically (already wired in compose healthcheck)
docker compose logs redis | grep -i "ready"
docker compose logs gateway | grep -i "redis"
# Expect: "Redis is ready" with no SSL errors
```

#### 6. Regression check

Revert the three compose edits (comment back plain URL, plain command, 6379 port) and run `make compose-up` again. Stack must come up cleanly on plain TCP.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,7 @@ pyo3 = { version = "0.28.2", features = ["abi3-py311"] }
 pyo3-log = "0.13.3"
 pyo3-stub-gen = "0.19"
 rand = "0.8"
-redis = { version = "0.32.7", features = ["aio", "tokio-comp", "connection-manager"] }
+redis = { version = "0.32.7", features = ["aio", "tokio-comp", "connection-manager", "tls-native-tls"] }
 regex = "1.12.2"
 rmcp = { version = "1.4.0", default-features = false }
 sha2 = "0.10.9"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,7 @@ pyo3 = { version = "0.28.2", features = ["abi3-py311"] }
 pyo3-log = "0.13.3"
 pyo3-stub-gen = "0.19"
 rand = "0.8"
-redis = { version = "0.32.7", features = ["aio", "tokio-comp", "connection-manager", "tls-native-tls"] }
+redis = { version = "0.32.7", features = ["aio", "tokio-native-tls-comp", "connection-manager"] }
 regex = "1.12.2"
 rmcp = { version = "1.4.0", default-features = false }
 sha2 = "0.10.9"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,7 @@ pyo3 = { version = "0.28.2", features = ["abi3-py311"] }
 pyo3-log = "0.13.3"
 pyo3-stub-gen = "0.19"
 rand = "0.8"
-redis = { version = "0.32.7", features = ["aio", "tokio-native-tls-comp", "connection-manager"] }
+redis = { version = "0.32.7", features = ["aio", "tokio-comp", "connection-manager"] }
 regex = "1.12.2"
 rmcp = { version = "1.4.0", default-features = false }
 sha2 = "0.10.9"
--- a/docs/docs/manage/configuration.md
+++ b/docs/docs/manage/configuration.md
@@ -961,9 +961,17 @@ The gateway includes built-in observability features for tracking HTTP requests,
 | `REDIS_LEADER_TTL`        | Leader election TTL (secs) | `15`       | int > 0                  |
 | `REDIS_LEADER_KEY`        | Leader key name            | `gateway_service_leader` | string |
 | `REDIS_LEADER_HEARTBEAT_INTERVAL` | Heartbeat (secs)   | `5`        | int > 0                  |
+| `REDIS_SSL`               | Enable TLS for Redis       | `false`    | bool                     |
+| `REDIS_SSL_CA_CERTS`      | Path to CA certificate bundle | (none)  | file path                |
+| `REDIS_SSL_CERTFILE`      | Path to client certificate (mTLS) | (none) | file path           |
+| `REDIS_SSL_KEYFILE`       | Path to client private key (mTLS) | (none) | file path           |
+| `REDIS_SSL_CHECK_HOSTNAME`| Verify hostname in TLS cert | `false`   | bool                     |
 
 !!! tip "Cache Backend Selection"
     Use `memory` for dev, `database` for local persistence, or `redis` for distributed caching across multiple instances. `none` disables caching entirely.
+
+!!! note "Redis TLS"
+    For TLS, set `REDIS_URL` to use the `rediss://` scheme (note the double `s`) and set `REDIS_SSL=true`. Supply `REDIS_SSL_CA_CERTS` to verify the server certificate. For mutual TLS (mTLS), also set `REDIS_SSL_CERTFILE` and `REDIS_SSL_KEYFILE`. Enable `REDIS_SSL_CHECK_HOSTNAME=true` only when Redis presents a valid CA-signed certificate with a matching hostname.
 
 ### Tool Lookup Cache
 
--- a/docs/docs/manage/configuration.md
+++ b/docs/docs/manage/configuration.md
@@ -965,7 +965,7 @@ The gateway includes built-in observability features for tracking HTTP requests,
 | `REDIS_SSL_CA_CERTS`      | Path to CA certificate bundle | (none)  | file path                |
 | `REDIS_SSL_CERTFILE`      | Path to client certificate (mTLS) | (none) | file path           |
 | `REDIS_SSL_KEYFILE`       | Path to client private key (mTLS) | (none) | file path           |
-| `REDIS_SSL_CHECK_HOSTNAME`| Verify hostname in TLS cert | `false`   | bool                     |
+| `REDIS_SSL_CHECK_HOSTNAME`| Verify hostname in TLS cert | `true`   | bool                     |
 
 !!! tip "Cache Backend Selection"
     Use `memory` for dev, `database` for local persistence, or `redis` for distributed caching across multiple instances. `none` disables caching entirely.
--- a/mcpgateway/utils/redis_client.py
+++ b/mcpgateway/utils/redis_client.py
@@ -24,6 +24,7 @@ Usage:
 
 # Standard
 import logging
+import ssl
 from typing import Any, Optional
 
 # First-Party
@@ -36,6 +37,39 @@ _parser_info: Optional[str] = None
 
 _client: Optional[Any] = None
 _initialized: bool = False
+
+
+def _build_ssl_context(settings: Any) -> Optional[ssl.SSLContext]:
+    """Build an SSL context for Redis TLS connections.
+
+    Returns ``None`` when TLS is disabled (local dev default).  In
+    production set ``REDIS_SSL=true`` (and use a ``rediss://`` URL) then
+    optionally supply ``REDIS_SSL_CA_CERTS``, ``REDIS_SSL_CERTFILE``, and
+    ``REDIS_SSL_KEYFILE`` for certificate verification / mutual TLS.
+
+    Args:
+        settings: Application settings instance.
+
+    Returns:
+        ssl.SSLContext configured for Redis, or None when TLS is off.
+    """
+    if not settings.redis_ssl:
+        return None
+
+    ctx = ssl.create_default_context(cafile=settings.redis_ssl_ca_certs or None)
+
+    if settings.redis_ssl_certfile:
+        ctx.load_cert_chain(
+            certfile=settings.redis_ssl_certfile,
+            keyfile=settings.redis_ssl_keyfile or None,
+        )
+
+    if not settings.redis_ssl_check_hostname:
+        # Allow self-signed certs in non-production environments only
+        ctx.check_hostname = False
+        ctx.verify_mode = ssl.CERT_NONE
+
+    return ctx
 
 
 def _is_hiredis_available() -> bool:
@@ -152,6 +186,12 @@ async def get_redis_client() -> Optional[Any]:
         # Only specify parser_class if explicitly set (not auto)
         if parser_class is not None:
             connection_kwargs["parser_class"] = parser_class
+
+        # Inject TLS context when REDIS_SSL=true (production).
+        # Local dev: returns None → no ssl kwarg → plain TCP connection.
+        ssl_context = _build_ssl_context(settings)
+        if ssl_context is not None:
+            connection_kwargs["ssl"] = ssl_context
 
         _client = aioredis.from_url(settings.redis_url, **connection_kwargs)
         await _client.ping()
--- a/mcpgateway/utils/redis_client.py
+++ b/mcpgateway/utils/redis_client.py
@@ -24,6 +24,8 @@ Usage:
 
 # Standard
 import logging
+import os
+import ssl as _ssl
 from typing import Any, Optional
 
 # First-Party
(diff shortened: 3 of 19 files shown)


Cargo.toml
```diff
@@ -31,7 +31,7 @@ pyo3 = { version = "0.28.2", features = ["abi3-py311"] }
 pyo3-log = "0.13.3"
 pyo3-stub-gen = "0.19"
 rand = "0.8"
-redis = { version = "0.32.7", features = ["aio", "tokio-comp", "connection-manager"] }
+redis = { version = "0.32.7", features = ["aio", "tokio-comp", "connection-manager", "tls-native-tls"] }
```
lucarlig: the feature should be `tokio-native-tls-comp` instead of `tls-native-tls`, it fails.

lucarlig (changes requested):
@Lang-Akshay I see you have added rust support as well, is this needed, since we might want to deprecate?

 If it is, rust will use TLS plus backend default as implemented.
is it necessary to wire in the new python env vars?
might need validation tests? i tested seems to be failing because of wrong feature i highlighted
if current implementation is ok as it is maybe add  clarifying docs somewhere that rust supports only basic `rediss://` and don't wire in new vars?

mcpgateway/utils/redis_client.py
```diff
@@ -38,6 +38,39 @@
...
+    if not settings.redis_ssl:
+        return {}
+
+    kwargs: dict[str, Any] = {}
+
+    if settings.redis_ssl_ca_certs:
+        kwargs["ssl_ca_certs"] = settings.redis_ssl_ca_certs
+    if settings.redis_ssl_certfile:
+        kwargs["ssl_certfile"] = settings.redis_ssl_certfile
+    if settings.redis_ssl_keyfile:
```
dawid-nowak, line 112: My two cents is that if you have flag to set ssl set and there are missing parameters in settings you should return early... return an error and don't start the application

So here :
I would validate that ca_certs exists, it contains  a valid certificate etc. and the same for the private key and the cert bundle.

dawid-nowak (commented):
My two cents is that if you have flag to set ssl set and there are missing parameters in settings you should return early... return an error and don't start the application

So here :
I would validate that ca_certs exists, it contains a valid certificate etc. and the same for the private key and the cert bundle.

gandhipratik203 (changes requested):
Thanks for putting this together — Redis TLS is one of those features that's a pain to retrofit cleanly, and the opt-in framing here keeps local dev completely untouched, which is the right call.

## Issue context (github.ibm.com/WatsonOrchestrate/wo-tracker#68461)

The linked issue is a "blocker"-labelled Defect: WXO deployments on AWS, IBM Cloud, on-prem, and **FedRamp** need Redis TLS with certificate-based authentication. The ask is explicitly for *secure* Redis connections in enterprise environments — CA paths, cert/key for mTLS, and the ability to verify the server. Worth keeping that "FedRamp / secure" framing in mind because one of the defaults in the current PR points the other way (see F2).

## What's Good

A few parts that stood out as nicely done:

- The defaults are exactly right at the toggle level — `REDIS_SSL=false`, no required env vars, plain `redis://` URLs continue to work without certs. Local dev is completely undisturbed.
- Single source of truth for the SSL configuration — `_build_ssl_kwargs(settings)` is shared between the async client (`get_redis_client`) and the startup readiness probe (`wait_for_redis_ready`), so the two paths can't drift apart. The `connection_kwargs.update(_build_ssl_kwargs(settings))` pattern is clean.
- Both paths use the redis-py 7.x `ssl_*` kwarg interface consistently — `ssl_ca_certs`, `ssl_certfile`, `ssl_keyfile`, `ssl_cert_reqs`, `ssl_check_hostname` are all the standard names, and they're spread the same way in async and sync code.
- The five-setting surface (`REDIS_SSL`, `REDIS_SSL_CA_CERTS`, `REDIS_SSL_CERTFILE`, `REDIS_SSL_KEYFILE`, `REDIS_SSL_CHECK_HOSTNAME`) maps cleanly onto what the issue asks for: CA paths and client cert/key for mTLS.
- The PR description's testing guide is excellent — Option A (mkcert + bare-metal Redis) and Option B (docker-compose with all the comment-swap steps including the volume mount and port swap) give reviewers and operators a concrete path to verify the change end-to-end.

## Findings

Below are some of the findings — a few worth looking at before merge, and some smaller ones that can ride along or follow up.

**F1 — Could we add tests for the new code?** The PR adds about 162 net lines but doesn't update `tests/unit/mcpgateway/utils/test_redis_client.py` or `tests/unit/mcpgateway/utils/test_redis_isready.py`, and doesn't add new ones. A couple of layers would help here:

- **Unit tests (must-have):** `_build_ssl_kwargs()` is a pure function with clear branches (TLS off returns `{}`, TLS on with no certs, TLS on with CA, TLS on with mTLS, hostname-check off) — testing each branch directly would be quick. Additionally, mocking `aioredis.from_url` and `Redis.from_url` to assert the SSL kwargs reach the client would confirm the wiring works.
- **Integration test (strongly recommended):** spin up Redis with TLS in a fixture using the official `redis:7` image (supports `--tls-port`, `--tls-cert-file`, etc.), generate ephemeral self-signed certs into `tmp_path`, and run a real round-trip via both `get_redis_client` and `wait_for_redis_ready`. For an issue labelled "blocker" and called out for FedRamp, having an automated TLS handshake test in CI feels especially valuable.

**F2 — The default for `REDIS_SSL_CHECK_HOSTNAME` is `false`, and the underlying code path turns off all certificate verification rather than only the hostname check.** Looking at the current code:

```python
# config.py
redis_ssl_check_hostname: bool = Field(
    default=False,
    description="Verify the hostname in the Redis TLS certificate ...",
)
```

```python
# redis_client.py
if not settings.redis_ssl_check_hostname:
    kwargs["ssl_cert_reqs"] = "none"      # disables all cert verification
    kwargs["ssl_check_hostname"] = False
```

The setting is named and documented as "hostname verification", but flipping it off also sets `ssl_cert_reqs="none"`, which tells redis-py to skip the entire cert chain check — the client will accept any cert, including a MITM attacker's. With the default of `false`, anyone turning on `REDIS_SSL=true` without explicitly setting `REDIS_SSL_CHECK_HOSTNAME=true` gets a TLS connection that doesn't verify the server. For an issue that's specifically about FedRamp / enterprise-grade secure connections, this feels worth a second look. A few options to consider:

1. Flip the default to `True` and keep the field name accurate to its behaviour — for self-signed certs, operators add the self-signed CA to `REDIS_SSL_CA_CERTS` (the standard pattern and what FedRamp deployments would do anyway).
2. If a single "skip everything" knob is genuinely useful for dev convenience, separate the two settings: keep `REDIS_SSL_CHECK_HOSTNAME` as just-the-hostname (default `True`), and add a clearly-named `REDIS_SSL_INSECURE_SKIP_VERIFY` (default `False`) for the cert-verification-disable case.
3. At minimum, rename the current setting to `REDIS_SSL_INSECURE_SKIP_VERIFY` and document the actual behaviour, so the name matches what the code does.

The concern here is that an operator reading the variable name and the docs reasonably assumes "I'm just relaxing the hostname check" while the code is actually disabling cert validation entirely — exactly the failure mode the issue is trying to fix.

**F3 — The PR description mentions Rust TLS support, but the latest commit reverts the Cargo changes.** The PR body says:

> Rust — no `.rs` files changed. `Cargo.toml` adds `tokio-native-tls-comp` to the `redis` workspace dependency so Rust sidecars can use `rediss://` URLs without code changes.

The latest commit (`5cc060d1c revert: restore Cargo.toml and Cargo.lock to main state`) removed that feature flag. Rust sidecars built off this branch would fail at connect time against a `rediss://` URL because the `redis` crate isn't compiled with a TLS backend. Was the revert intentional? Two ways to align: either re-add the feature flag to `Cargo.toml` (matching the description's claim) or update the PR description to note Rust TLS is out of scope and will follow up separately.

**F4 — Worth handling the `rediss://` + `REDIS_SSL=false` misconfiguration case.** If an operator sets `REDIS_URL=rediss://...` but forgets to flip `REDIS_SSL=true`, redis-py will use TLS (because of the scheme) but with default system trust — no custom CA, no mTLS. Could work or fail unpredictably depending on the server's cert. A startup warning ("URL scheme is `rediss://` but `REDIS_SSL=false`; TLS configuration will be ignored") would surface this loudly.

**F5 — A friendlier error path for TLS handshake failures would help.** Right now, cert problems (expired, hostname mismatch, untrusted CA) will surface as `ssl.SSLError` from the readiness probe and be retried as if they were transient connection errors. A targeted catch in `wait_for_redis_ready` with a clear log line — "TLS handshake failed: <reason> — check REDIS_SSL_CA_CERTS / REDIS_SSL_CERTFILE / REDIS_SSL_KEYFILE" — would save real debugging time during an enterprise rollout. Not blocking, but a meaningful operator-experience win for this specific issue's audience.

**F6 — `_build_ssl_kwargs` is a leading-underscore "private" helper but it's imported across modules.** It's called from `main.py`. Either dropping the underscore (it's part of the public surface between `redis_client.py` and `main.py`) or wrapping it in a public re-export would be more honest about the surface. Minor — fine for follow-up.

**F7 — Could we note the redis-py 7.x requirement in `pyproject.toml` or the PR body?** The docstring on `_build_ssl_kwargs` mentions "redis-py 7.x SSLConnection accepts individual ssl_* params" — but that constraint isn't reflected in the dependency pin. If anything in the project is using redis-py 6.x semantics, the kwarg names differ. A `redis>=7,<8` constraint (or just a note in the PR body) would help.

---

Once the findings are addressed, I think we're in good shape to merge. Happy to take another look once those are in. Thanks again for the work on this!

Lang-Akshay:
---
F1 — Tests (resolved)

The original review flagged missing tests, but test_redis_client.py already covers the key _build_ssl_kwargs branches (TLS off → {}, hostname check disabled, _validate_ssl_settings with missing/invalid certs). Unit coverage is solid. Integration tests with a real Redis+TLS container would be a nice follow-up for a feature tagged as FedRAMP-relevant, but not blocking.
---
F2 — REDIS_SSL_CHECK_HOSTNAME default flipped to True (fixed in this branch)

The original default was False, which caused ssl_cert_reqs="none" to be set whenever TLS was enabled — disabling the full certificate chain check, not just the hostname check. An operator enabling REDIS_SSL=true without explicitly setting REDIS_SSL_CHECK_HOSTNAME=true would get a TLS connection that accepts any certificate, including a MITM attacker's.

Changed:

- Default is now True
- Description updated: "Verify the Redis TLS certificate chain and hostname. Set False only for self-signed certs (pair with REDIS_SSL_CA_CERTS for the CA bundle)"
- Safe when REDIS_SSL=false — _build_ssl_kwargs() returns {} immediately in that path, so the hostname setting is never evaluated

---

F3 — Rust TLS (intentionally out of scope)

The revert: restore Cargo.toml and Cargo.lock to main state commit removed the tokio-native-tls feature flag. This was intentional — Rust TLS support is out of scope for this PR. The PR description still references it; worth updating to note Rust TLS will be a separate follow-up.

---

F4 — rediss:// + REDIS_SSL=false mismatch warning (fixed in this branch)

If an operator sets REDIS_URL=rediss://... but leaves REDIS_SSL=false, redis-py will open a TLS socket (from the URL scheme), but our code returns {} from _build_ssl_kwargs() — so none of our SSL settings (CA cert, hostname check) are applied. This silently bypasses F2's fix entirely.

Added a startup warning in get_redis_client():
REDIS_URL uses rediss:// scheme but REDIS_SSL=false — TLS certificate settings (REDIS_SSL_CA_CERTS, REDIS_SSL_CHECK_HOSTNAME) will not be applied

---

F5 — TLS handshake failures now fail fast (fixed in this branch)

Previously, ssl.SSLError from a bad cert (expired, hostname mismatch, untrusted CA) would land in the generic except Exception retry loop and be retried up to 30 times with no useful log output. SSL errors are not transient — retrying doesn't help.

Added a targeted except ssl.SSLError catch before the generic handler in wait_for_redis_ready. It logs:
TLS handshake failed connecting to Redis: `<error>` — check REDIS_SSL_CA_CERTS / REDIS_SSL_CERTFILE / REDIS_SSL_KEYFILE
...then raises immediately, skipping all retries.

---

F6 — _build_ssl_kwargs underscore naming (won't fix)

It's imported across modules (including main.py) with a leading underscore, which conventionally signals private. Noted as a minor inconsistency, no change made.

---

F7 — redis-py version pin (already handled)

pyproject.toml has redis>=7.4.0 in runtime dependencies, which covers the ssl_* kwarg API used in _build_ssl_kwargs. Minor note: dev dependencies have redis>=6.4.0 — worth aligning those floors in a follow-up to avoid a dev environment accidentally running redis-py 6.x where the kwarg names differ.

docs/docs/manage/configuration.md
```diff
@@ -961,10 +961,18 @@ The gateway includes built-in observability features for tracking HTTP requests,
 | `REDIS_LEADER_TTL`        | Leader election TTL (secs) | `15`       | int > 0                  |
 | `REDIS_LEADER_KEY`        | Leader key name            | `gateway_service_leader` | string |
 | `REDIS_LEADER_HEARTBEAT_INTERVAL` | Heartbeat (secs)   | `5`        | int > 0                  |
+| `REDIS_SSL`               | Enable TLS for Redis       | `false`    | bool                     |
+| `REDIS_SSL_CA_CERTS`      | Path to CA certificate bundle | (none)  | file path                |
+| `REDIS_SSL_CERTFILE`      | Path to client certificate (mTLS) | (none) | file path           |
+| `REDIS_SSL_KEYFILE`       | Path to client private key (mTLS) | (none) | file path           |
+| `REDIS_SSL_CHECK_HOSTNAME`| Verify hostname in TLS cert | `false`   | bool                     |
```
madhu-mohan-jaishankar: should this be 'true' here since the config defaults it to true.

madhu-mohan-jaishankar (dismissed):
LGTM

gandhipratik203 (dismissed):
Thanks so much for the thoughtful turnaround on all the findings — really nice work. The one small thing remaining is the PR description still referencing Rust TLS support, but that's an optional doc tweak and not worth blocking on. Approving from my side. Thanks again!

gandhipratik203 (approved):
Thanks so much for the thoughtful turnaround on all the findings — really nice work. The one small thing remaining is the PR description still referencing Rust TLS support, but that's an optional doc tweak and not worth blocking on. Approving from my side. Thanks again!

madhu-mohan-jaishankar (approved):
LGTM.

PR #3137, opened 2026-02-24
feat(ui): modularize admin scope into independent modules

crivetimihai:
# Admin UI Modularisation

## The Problem

The original `mcpgateway/static/admin.js` was a **33,812-line monolithic file** — all UI logic, state management, API calls, validation, and event handling in one place. It was untestable in isolation, slow to load, and impossible to maintain at scale.

---

### 1. Decomposition into 44 ES Modules (`mcpgateway/admin_ui/`)

The monolith was split into focused, single-responsibility modules across four conceptual layers:

| Layer | Modules | Purpose |
|---|---|---|
| **Foundation** | `bootstrap.js`, `appState.js`, `utils.js`, `constants.js`, `security.js` | Shared primitives, state, XSS protection |
| **Domain** | `tools.js`, `servers.js`, `teams.js`, `gateway.js`, `tokens.js`, `llmModels.js`, `llmChat.js`, `prompts.js`, `resources.js`, `users.js`, `roots.js`, `a2aAgents.js` | Entity-level CRUD, testing, validation |
| **Features** | `auth.js`, `filters.js`, `search.js`, `tabs.js`, `pagination.js`, `modals.js`, `logging.js`, `metrics.js`, `plugins.js`, `monitoring.js`, etc. | UI concerns, cross-cutting features |
| **Orchestration** | `initialization.js`, `events.js`, `app.js`, `formSubmitHandlers.js` | Startup, HTMX delegation, chart registry |

### 2. Facade Pattern (`admin_ui/admin.js`)

A single facade aggregates all 35+ module exports and exposes them onto the `window.Admin` namespace — preserving the existing HTML contract where HTMX/Alpine templates call `Admin.editTool(...)`, `Admin.showTab(...)`, etc. No template changes required.

```
index.js
  └─→ admin.js (facade)
        ├─→ bootstrap.js         (window.Admin = {})
        ├─→ utils.js, tools.js, servers.js ... (ES modules → Admin.fn = fn)
        ├─→ events.js            (IIFE, self-registers)
        └─→ app.js               (IIFE, self-registers)
```

### 3. Vite Build Pipeline

A new Vite configuration (`vite.config.js`) bundles `admin_ui/index.js` into an **IIFE format** output at `mcpgateway/static/bundle-[HASH].js`. The hash changes on every rebuild, providing automatic cache busting. External libraries (HTMX, Alpine, Chart.js, CodeMirror, DOMPurify) remain CDN-loaded and are declared as globals — not bundled.

[...]

PR #4823, opened 2026-05-20
fix(ui): replace htmx.ajax() with element-based trigger in loadSearchablePanel

### LINKED ISSUE #4822: [BUG]: Tools page search breaks after first use — htmx:swapError / null querySelector on repeated search
### 🐞 Bug Summary

Searching in the Tools panel of the Admin UI works once, then fails on every subsequent search attempt. The browser console shows:

```
[Error] htmx:swapError
[Error] TypeError: null is not an object (evaluating 'o.querySelector')
[Error] Global error: – null – "" – 0
```

---

### 🧩 Affected Component

---

### 🔁 Steps to Reproduce

1. Open the Admin UI and navigate to the **Tools** tab.
2. Type any term in the **Search tools...** input. Results update correctly.
3. Clear the input and type a second search term (or type again immediately).
4. Observe: the table does not update and the browser console shows `htmx:swapError` / `TypeError: null is not an object (evaluating 'o.querySelector')`.

---

### 🤔 Expected Behavior

Each keystroke in the search input should fetch a fresh filtered result set and update both the tools table and the pagination controls, regardless of how many times search is triggered.

---

### 📓 Logs / Error Output

```
[Error] htmx:swapError
    b (htmx.min.js:1:26882)
    ce (htmx.min.js:1:27051)
    fe (htmx.min.js:1:26702)
    y (htmx.min.js:1:45617)
    Mr (htmx.min.js:1:46011)
    (anonymous function) (htmx.min.js:1:41058)
[Error] TypeError: null is not an object (evaluating 'o.querySelector')
    b (htmx.min.js:1:26882)
    ce (htmx.min.js:1:27051)
    fe (htmx.min.js:1:26702)
    (anonymous function) (htmx.min.js:1:41290)
[Error] Global error: – null – "" – 0
    (anonymous function) (bundle-DFGpMx9H.js:1:379126)
[Error] TypeError: null is not an object (evaluating 'o.querySelector')
    (anonymous function) (htmx.min.js:1:41336)
```

---

### 🔍 Root Cause Analysis

**File:** `mcpgateway/admin_ui/search.js` — `loadSearchablePanel()`

The search handler for all panel tables (tools, servers, resources, prompts, gateways, agents) uses `window.htmx.ajax()` to fetch and swap in the partial response:

```js
window.htmx.ajax(GET, url, {
  target: panelConfig.targetSelector,  // #tools-table
  swap: outerHTML,
  indicator: panelConfig.indicatorSelector,
});
```

The server response from `/admin/tools/partial` contains **two** fragments:

1. `<table id="tools-table">` — the main swap target (replaces `#tools-table` via `outerHTML`)
2. `<div id="tools-pagination-controls" hx-swap-oob="true">` — an out-of-band (OOB) swap to update pagination

**The problem:** `htmx.ajax()` does not reliably process `hx-swap-oob` elements in the response. This is already documented in the codebase — `tokens.js:63–65` contains the comment:

> *"Update hx-get on the element and trigger an element-based HTMX request. This ensures OOB swaps (pagination controls) are properly processed, unlike htmx.ajax() which can silently skip OOB handling."*

When `htmx.ajax()` skips OOB processing, HTMX's internal swap code calls `o.querySelector()` where `o` is `null` (the resolved OOB target element), producing the `TypeError` and `htmx:swapError`. This happens reliably from the **second** search onward because the first load uses a declarative `hx-trigger="load"` path that processes OOB correctly, but all subsequent searches go through `htmx.ajax()`.

The same root cause affects **all panels** sharing `loadSearchablePanel()`: tools, servers, resources, prompts, gateways, and agents.

---

### ✅ Fix

Replace `htmx.ajax()` with the element-based trigger pattern already used by the tokens panel (`tokens.js:63–68`):

```js
// mcpgateway/admin_ui/search.js — loadSearchablePanel()
const targetEl = document.querySelector(panelConfig.targetSelector);
if (targetEl) {
  targetEl.setAttribute(hx-get, url);
  targetEl.setAttribute(hx-swap, outerHTML);
  targetEl.setAttribute(hx-trigger, searchReload);
  if (panelConfig.indicatorSelector) {
    targetEl.setAttribute(hx-indicator, panelConfig.indicatorSelector);
  }
  window.htmx.process(targetEl);   // re-registers HTMX listeners
  window.htmx.trigger(targetEl, searchReload);  // fires declarative request
}
```

This routes the request through HTMX's full declarative pipeline, ensuring OOB elements are processed correctly on every search invocation.

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Version or commit | `v1.0.1 / main@16f8d1a00` |
| Runtime | Python 3.11, Uvicorn/Gunicorn |
| Container | Docker, Podman, or none |

---

### 🧩 Additional Context

- The bug affects all searchable panels using `PANEL_SEARCH_CONFIG` in `mcpgateway/admin_ui/constants.js`: `tools`, `servers`, `catalog`, `resources`, `prompts`, `gateways`, `agents`.
- The tokens panel already uses the correct workaround (`tokens.js:63–68`); this fix aligns the remaining panels with that pattern.
- A fix has been applied in `mcpgateway/admin_ui/search.js` and the bundle rebuilt.

madhav165:
# 🐛 Bug-fix PR

## 📌 Summary

Panel search (tools, servers, resources, prompts, gateways, agents) works on first use but breaks on every subsequent search with `htmx:swapError` / `TypeError: null is not an object (evaluating 'o.querySelector')` in the browser console.

The root cause is that `loadSearchablePanel()` uses `window.htmx.ajax()`, which silently skips `hx-swap-oob` elements in the response. The partial responses for all these panels return an out-of-band `<div id="*-pagination-controls" hx-swap-oob="true">` alongside the main table fragment. When HTMX fails to process the OOB swap, it tries to call `.querySelector()` on a null target reference and throws.

## 🔁 Reproduction Steps

Fixes #4822.

1. Open the Admin UI → **Tools** tab (or Servers, Resources, Prompts, Gateways, Agents).
2. Type any term in the search input — results update correctly.
3. Type a second search term (or clear and type again).
4. Observe: table does not update; browser console shows `htmx:swapError` and `TypeError: null is not an object (evaluating 'o.querySelector')`.

## 🐞 Root Cause

`loadSearchablePanel()` in `mcpgateway/admin_ui/search.js` calls `window.htmx.ajax()` to fetch and swap the panel partial. The server response contains both the main table fragment and an OOB pagination-controls fragment (`hx-swap-oob="true"`). `htmx.ajax()` does not reliably process OOB elements — this is already documented in the codebase at `tokens.js:63–65`:

> _"This ensures OOB swaps (pagination controls) are properly processed, unlike `htmx.ajax()` which can silently skip OOB handling."_

When OOB processing is skipped, HTMX internally resolves the OOB target to `null` and crashes with `o.querySelector` on the second and subsequent invocations.

## 💡 Fix Description

Replace `htmx.ajax()` with the element-based trigger pattern already used by the tokens panel (`tokens.js:63–68`):

1. Set `hx-get`, `hx-swap`, `hx-trigger`, and `hx-indicator` directly on the target element.
2. Call `htmx.process(targetEl)` to register HTMX handlers on the updated attributes.
3. Call `htmx.trigger(targetEl, "searchReload")` to fire the request through HTMX's full declarative pipeline.

This ensures OOB swap elements in the response are correctly processed on every search invocation, not just the first.

The fix is a one-function change with no template or backend modifications required.

## 🧪 Verification

| Check | Command | Status |
|---|---|---|
| Lint suite | `make lint` | JS-only change; no Python linting impact |
| Unit tests | `make test` | No Python logic changed |
| Coverage ≥ 80 % | `make coverage` | Unaffected |
| Manual regression no longer fails | Type in search box twice on Tools page | Fixed |

## ✅ Checklist

### DIFF
--- a/mcpgateway/admin_ui/search.js
+++ b/mcpgateway/admin_ui/search.js
@@ -102,12 +102,23 @@ export const loadSearchablePanel = function (entityType) {
   }
 
   const url = `${window.ROOT_PATH}/admin/${panelConfig.partialPath}?${params.toString()}`;
-  if (window.htmx && window.htmx.ajax) {
-    window.htmx.ajax("GET", url, {
-      target: panelConfig.targetSelector,
-      swap: "outerHTML",
-      indicator: panelConfig.indicatorSelector,
-    });
+  // Use element-based HTMX trigger instead of htmx.ajax() to ensure OOB swaps
+  // (e.g. pagination controls) are reliably processed.  htmx.ajax() can silently
+  // skip hx-swap-oob elements in the response, causing a null-querySelector crash
+  // on every search after the first one.  The tokens panel already uses this same
+  // pattern (tokens.js:63-68).
+  if (window.htmx) {
+    const targetEl = document.querySelector(panelConfig.targetSelector);
+    if (targetEl) {
+      targetEl.setAttribute("hx-get", url);
+      targetEl.setAttribute("hx-swap", "outerHTML");
+      targetEl.setAttribute("hx-trigger", "searchReload");
+      if (panelConfig.indicatorSelector) {
+        targetEl.setAttribute("hx-indicator", panelConfig.indicatorSelector);
+      }
+      window.htmx.process(targetEl);
+      window.htmx.trigger(targetEl, "searchReload");
+    }
   }
 };
 


madhav165:
Closing this PR — the root cause analysis it relies on does not hold up.

The central claim — that `htmx.ajax()` "silently skips `hx-swap-oob` elements" — is not supported by htmx documentation or source. htmx 2.0.10 (the bundled version) processes OOB swaps through the same response pipeline regardless of whether the request is initiated via `htmx.ajax()` or a declarative trigger. The only "evidence" was the comment in `tokens.js:63–65`, which is self-referential — it was added in the original modularization PR (#3137) without a repro or upstream link, and has been propagating as folklore.

The "first works, second fails" pattern is also incompatible with the RCA: if `htmx.ajax()` truly skipped OOB processing, the *first* search would fail to update pagination too. Same code path either skips OOB or it doesn't.

The change here happens to dodge the bug, but as a side effect of forcing a fresh `htmx.process()` re-registration of the target on every call — not because `htmx.ajax()` is broken. It's a workaround without a correct diagnosis, and it doesn't include a deterministic regression test that would fail against the old code and pass against the new.

I've left a more detailed write-up on #4822, including the likely real cause (stale htmx state / DOM-ID resolution against the OOB-swapped `#*-pagination-controls` element, possibly interacting with the Alpine reinit pass at `initialization.js:958–970`) and notes on what a correct fix should include (JSDOM regression test, freshly built bundle, addressing whatever leaves the OOB target unresolvable on the second swap rather than masking it).

Thanks for the investigation regardless — the symptom is real, just diagnosed incorrectly.

PR #4839, opened 2026-05-20
feat(ui): Add user screen and create user form

gcgoncalves:
Closes: https://github.com/IBM/mcp-context-forge/issues/4183

Screenshot 2026-05-21 at 09 51 22

Screenshot 2026-05-21 at 09 51 35

Screenshot 2026-05-21 at 09 50 41


marekdano (changes requested):
Found Issues

  1. `usersApi.ts` and `usersApi.test.ts` should be removed

The project pattern is to call endpoints directly via `useQuery` in hooks — `useUserForm.ts` does this correctly. The standalone `usersApi` service layer is redundant and adds a second code path that diverges from the convention. 

  2. User type has required fields missing from all test fixtures

  User declares these as non-optional:
```
  auth_provider: string;
  email_verified: boolean;
  failed_login_attempts: number;
  is_locked: boolean;
```
  Every `mockUser` object in both `users.test.ts` and `useUserForm.test.tsx` omits them. This won't compile under strict TypeScript. Either add the missing fields to all fixtures, or mark them optional in the type if the backend can omit them.

  3. `full_name: null` is not assignable to `full_name?: string`

  Several test fixtures set `full_name: null`, but the type is `string | undefined`. Either change the fixtures to use `undefined`, or widen the type to `string | null | undefined`.

  ---
  Design Issues

  4. `isValid` uses a looser regex than the Zod schema

  `useUserForm.ts` computes `isValid` with a simple `/^[^\s@]+@[^\s@]+\.[^\s@]+$/ ` check, while the Zod schema uses a strict `RFC 5322` pattern. An address could pass the simpler check, enable the `Submit` button, then fail `Zod` on submit — confusing UX. Use one validation source for both.

  5. handleCancel is an unnecessary wrapper
```
  const handleCancel = () => { onToggle(); };
  // ...
  onClick={() => handleCancel()}
```
  Both the function and the arrow wrapper are needless. Use `onClick={onToggle}` directly.

gcgoncalves:
Thanks, @marekdano for the review. Loads of this issues would have been avoided using orval as you suggested. I'm studying a way for us to include it into our phase 2 of the app. I'll address your suggestions and re-request your review. 🙏

marekdano (approved):
It looks good! Just resolve the formatter failures by `cd client && npm run format`

LGTM 🚀

client/src/components/layout/Sidebar.tsx
```diff
@@ -57,6 +59,11 @@ const ECOSYSTEM_NAV_ITEMS: NavItem[] = [
   { labelKey: "navigation.plugins", path: "/app/plugins", icon: Blocks },
 ];
 
+const ADMINISTRATION_NAV_ITEMS: NavItem[] = [
```
vishu-bh, line 63: are you planning to handle rbac later? Cause for now non admin user will see this menu item and click it will result in 403.

gcgoncalves: Great catch, I'll address that!

PR #4842, opened 2026-05-20
feat(ui-rewrite): Add OAuth 2.0 popup authorization flow for MCP servers

marekdano:
## Summary

The existing OAuth flow was designed for the legacy HTMX admin UI, which owns the full page and can redirect freely. The React gateway form cannot do that - a full-page redirect to the provider and back would destroy the form state and drop the user at an unrelated screen.

This adds a popup-based flow so the user authorizes with the provider without leaving the form. The form stays open, shows a "waiting" banner while the popup is active, and displays a success or error notification inline when the flow settles - giving the user immediate feedback and the option to retry without re-entering their gateway details.

On the backend, the callback needs to know whether it was reached from a popup or a direct navigation so it can respond with a `postMessage` script rather than a full HTML page. Rather than adding a new storage key or endpoint, the state token is prefixed with `popup.` at flow initiation - the callback can detect the mode from the state it already validates, keeping the signal in a single value that travels with the request end-to-end.

Fix unit tests when we run `make test` to cover

### DIFF
--- a/client/src/hooks/useMCPServerForm.ts
+++ b/client/src/hooks/useMCPServerForm.ts
@@ -1,4 +1,4 @@
-import { useState, useCallback, useMemo, useEffect, type FormEvent } from "react";
+import { useState, useCallback, useMemo, useEffect, useRef, type FormEvent } from "react";
 import { z } from "zod";
 import { useQuery } from "@/hooks/useQuery";
 import { serversApi } from "@/api/servers";
@@ -313,6 +313,9 @@ export function useMCPServerForm(gatewayId?: string): UseMCPServerFormReturn {
     type: "success" | "error";
     message: string;
   } | null>(null);
+  // Tracks gateway created during a failed OAuth attempt so retries reuse it instead of creating a duplicate.
+  const [pendingOAuthGatewayId, setPendingOAuthGatewayId] = useState<string | undefined>(undefined);
+  const successCloseTimeoutRef = useRef<ReturnType<typeof setTimeout> | null>(null);
 
   const isEditMode = Boolean(gatewayId);
 
@@ -544,6 +547,12 @@ export function useMCPServerForm(gatewayId?: string): UseMCPServerFormReturn {
     oauthPassword,
   ]);
 
+  useEffect(() => {
+    return () => {
+      if (successCloseTimeoutRef.current) clearTimeout(successCloseTimeoutRef.current);
+    };
+  }, []);
+
   const validateField = useCallback((field: keyof FormErrors, value: string) => {
     try {
       const fieldSchema =
@@ -625,6 +634,7 @@ export function useMCPServerForm(gatewayId?: string): UseMCPServerFormReturn {
     setQueryParamName(initialState.queryParamName);
     setQueryParamApiKey(initialState.queryParamApiKey);
     setErrors({});
+    setPendingOAuthGatewayId(undefined);
   }, []);
 
   const handleSubmit = useCallback(
@@ -645,10 +655,16 @@ export function useMCPServerForm(gatewayId?: string): UseMCPServerFormReturn {
             }
             await updateGateway(formData);
             responseGatewayId = gatewayId;
+          } else if (authType === "oauth" && pendingOAuthGatewayId) {
+            // Reuse the gateway created in a previous OAuth attempt to avoid duplicates.
+            responseGatewayId = pendingOAuthGatewayId;
           } else {
             const response = await createGateway(formData);
             // Extract gateway ID from response
             responseGatewayId = (response as { id?: string })?.id;
+            if (authType === "oauth" && responseGatewayId) {
+              setPendingOAuthGatewayId(responseGatewayId);
+            }
           }
 
           if (authType === "oauth" && responseGatewayId) {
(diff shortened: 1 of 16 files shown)


a-effort:
Nice! Comments from Sonnet: 

`oauthrouter.oauth_router.py, safe_payload escaping` (~line 195) 

Good catch on `<`, `>`, and `&`. Two more characters worth handling: `U+2028` (LINE SEPARATOR) and `U+2029` (PARAGRAPH SEPARATOR). `json.dumps` emits them literally, but JavaScript treats them as line terminators inside string literals, so an OAuth `error_description` containing either will cause a `SyntaxError` and hang the popup.

Two possible options:
```
python

# Option A: extend the replace chain
safe_payload = (
    json.dumps(payload)
    .replace("<", "\\u003c")
    .replace(">", "\\u003e")
    .replace("&", "\\u0026")
    .replace("\u2028", "\\u2028")
    .replace("\u2029", "\\u2029")
)

# Option B: ensure_ascii=True escapes U+2028/2029 automatically,
# but < > & are still ASCII and need the manual replaces.
safe_payload = json.dumps(payload, ensure_ascii=True)
```

Suggestion: a test case in `test_popup_notification_script_escapes_dangerous_characters` with `"description": "line1\u2028line2"` would prevent regression.

`app.py`, logout CSRF
Adding `require_csrf` to `/auth/logout` is the right call. 

Two notes:
1. Discoverability. This is unrelated to the OAuth popup work and easy to miss in review. Consider splitting into its own commit/PR, or note it in the PR description (e.g. "Also adds CSRF protection to /auth/logout, previously missing").
2. Test coverage for:

Deny: `POST /auth/logout` without a token or with a bad token → 403, no cookies cleared, no token revoked.
Allow: `POST /auth/logout` with a valid token → 200, cookies cleared, token revoked.

The allow test would also confirm the React client is sending the token end-to-end, which is worth a manual smoke test if it hasn't been done.

gcgoncalves:
Suggestion: Should we write separate endpoints for this? It feels weird to receive HTML from the backend on react. Maybe edit the existing ones to check for an `Accept` header for returning json?

marekdano:
> Suggestion: Should we write separate endpoints for this? It feels weird to receive HTML from the backend on react. Maybe edit the existing ones to check for an `Accept` header for returning json?

@gcgoncalves  - I've investigated further to implement this and found out:

The suggestion is understandable but doesn't apply to this specific case.

The `OAuth callback` endpoint is fundamentally different from a typical API endpoint:

- The callback URL is hit via a browser redirect from the `OAuth` provider — not a `fetch()` call from the `React app`. When the `OAuth` provider redirects back, the browser navigates to the URL directly, sending `Accept: text/html`, not `Accept: application/json`. There's no React code running in that request.
- Since a script runs in the popup window, the only way to relay the result back to the React parent is via `window.opener.postMessage() + window.close()`. This is the industry-standard `OAuth` popup pattern — there's no alternative. If the backend returned JSON, the popup would just render the JSON text with no way to communicate the result back.

The `/app/auth/*` endpoints (login, me, logout) — which are proper fetch calls — already return JSON. This is the correct split.

gcgoncalves:
> > Suggestion: Should we write separate endpoints for this? It feels weird to receive HTML from the backend on react. Maybe edit the existing ones to check for an `Accept` header for returning json?
> 
> @gcgoncalves - I've investigated further to implement this and found out:
> 
> The suggestion is understandable but doesn't apply to this specific case.
> 
> The `OAuth callback` endpoint is fundamentally different from a typical API endpoint:
> 
> * The callback URL is hit via a browser redirect from the `OAuth` provider — not a `fetch()` call from the `React app`. When the `OAuth` provider redirects back, the browser navigates to the URL directly, sending `Accept: text/html`, not `Accept: application/json`. There's no React code running in that request.
> * Since a script runs in the popup window, the only way to relay the result back to the React parent is via `window.opener.postMessage() + window.close()`. This is the industry-standard `OAuth` popup pattern — there's no alternative. If the backend returned JSON, the popup would just render the JSON text with no way to communicate the result back.
> 
> The `/app/auth/*` endpoints (login, me, logout) — which are proper fetch calls — already return JSON. This is the correct split.

Perfect, thanks for investigating! 👏

client/src/hooks/useMCPServerForm.ts
```diff
@@ -625,23 +637,52 @@ export function useMCPServerForm(gatewayId?: string): UseMCPServerFormReturn {
...
+            const response = await createGateway(formData);
+            // Extract gateway ID from response
+            responseGatewayId = (response as { id?: string })?.id;
           }
 
-          // Call success callback if provided
-          if (onSuccess) {
-            onSuccess();
-          }
+          if (authType === "oauth" && responseGatewayId) {
```
vishu-bh, line 670: One thing to check if oAuth fails, and user retries again will the responseGatewayId be stored or createGateway will create a new gateway id?

Because I think if responseGatewayId is not stored and submitting again on create MCP form can result in error related to gateway Id.

client/src/hooks/useMCPServerForm.ts
```diff
@@ -625,23 +637,52 @@ export function useMCPServerForm(gatewayId?: string): UseMCPServerFormReturn {
...
+            // The form stays open (oauthPending=true) so the inline notification
+            // is visible when the popup posts its result back via postMessage.
+            setOAuthPending(true);
+            try {
+              const oauthResult = await serversApi.triggerOAuthAuthorization(responseGatewayId);
+              setOAuthNotification({
+                type: "success",
+                message: `OAuth authorization successful${oauthResult.gatewayName ? ` for ${oauthResult.gatewayName}` : ""}.`,
+              });
+              if (onSuccess) onSuccess();
```
vishu-bh: Sorry for the knitpick!
The success notification is effectively hidden in the real page flow.  this line sets the success message, then calls onSuccess; Servers.tsx (line 140) closes the form immediately, so the inline success feedback described will not be visible.

vishu-bh (approved):
Thanks Marek!

LGTM ☘️

Issue #4622, opened 2026-05-06
[CHORE]: Plugin bindings payload hardening — restrict caller-overridable config keys


gandhipratik203:
### 📝 Summary

`POST /v1/tools/plugin_bindings` currently accepts a free-form `config: Dict[str, Any]` per policy. Any key the caller sends is stored verbatim and merged into the per-tenant plugin config at runtime. There's no distinction between config that belongs at the gateway/operator layer and config that's genuinely meant to be set per (team, tool, plugin) by the caller.

This works, but it leaves the binding API with a wider responsibility than it really needs. Worth tightening.

### 🤔 Background

The binding API was introduced in #4068 with typed Pydantic schemas per plugin — each plugin had a known set of fields.

#4286 removed those per-plugin schemas, for sound architectural reasons: plugin contracts live in `cpex-plugins`, not in CF, so CF shouldn't have to know which plugins exist or what their config shape is. After that change, `config` became a fully open `Dict[str, Any]` that CF persists verbatim.

The decoupling intent was right. The side effect is that the API stopped distinguishing between two layers of plugin config that used to be implicitly separated:

- **Operator layer** — set once on the gateway and increasingly inherited from env (see #4582). Not meaningful per-tenant; the same value applies fleet-wide.
- **Caller layer** — behaviour knobs that legitimately vary per (team, tool). The whole reason the binding API exists.

Today both flow through the same opaque `config` dict, with no enforcement of which keys belong to which layer.

### 📐 Possible directions (open for discussion)

A couple of shapes that have come up — happy to defer to whatever the team prefers:

1. **Restrict the binding payload to caller-layer keys only.** Operator-layer keys come from the gateway-side plugin YAML (which already supports env substitution per #4582) and aren't accepted through the API.
2. **Plugin-declared allowlist.** Each plugin's gateway-side YAML declares which keys are caller-overridable; the binding API filters incoming payloads against that declared list. Keeps CF generic — the contract sits with the plugin author, where #4286 wanted it.

(1) and (2) compose well; (2) is the mechanism for (1). Either way, the goal is to put the layer boundary back in place without re-coupling CF to plugin-specific schemas.

### 🛟 Existing data

Any rows written under the current contract may contain keys that the tighter contract wouldn't accept. Worth a review pass once the new shape is decided.

### 📓 Related

- #4068 — introduced the binding API with typed per-plugin config schemas
[...]

Issue #4603, opened 2026-05-05
rate-limiter: remove obsolete redis_fallback config references


gandhipratik203:
## Summary

The `redis_fallback` config key is referenced across `plugins/config.yaml`, docs, the OCP helm chart, and several tests, but the current rate-limiter plugin (`cpex-rate-limiter`, the Rust port) does not recognize it. Operators using the shipped configs see a `"rate limiter: unknown config key(s): redis_fallback"` WARN at gateway boot and no behavior change. The docs describe a mechanism that no longer exists.

## Background

`redis_fallback` was a real config key in the in-tree Python rate-limiter plugin and was honored as documented (fall back to in-memory counting when Redis was unavailable). It became silently no-op when the in-tree plugin was replaced by the `cpex-rate-limiter` package. Behavior under unreachable Redis is now governed by `fail_mode` (default `"open"`).

We don't plan to reintroduce the feature: `fail_mode` covers the same operator decision space without the silent-degradation footgun of memory fallback in load-balanced deployments. If graceful degradation under Redis loss is wanted later, it should be designed as an explicit, observable mode — separate issue.

## Affected surfaces

| File | Action |
|---|---|
| `plugins/config.yaml` | Remove the `redis_fallback` line |
| `tests/performance/plugins/config.yaml` | Remove (3 occurrences) |
| `charts/mcp-stack/profiles/ocp/values-pgo.yaml` | Remove |
| `docs/docs/api/plugin-bindings-api.md` | Drop the `redis_fallback` row + remove from JSON examples |
| `tests/integration/test_rate_limiter.py` | Rename / tighten the misleadingly-named `test_redis_fallback_to_memory_on_unavailable_redis` (passes today only by virtue of the `fail_mode: "open"` default) |
| `tests/unit/.../test_gateway_plugin_manager.py`, `test_tool_plugin_bindings.py`, `test_tool_plugin_binding_service.py` | Drop `redis_fallback` from mock fixtures |

## Regression guard

Add `test_shipped_plugins_config_has_no_unknown_keys` (drafted in #4582, pulled out when that PR was scoped down): loads `plugins/config.yaml`, instantiates each plugin via the framework loader, asserts zero `"unknown config key"` warnings. Lands with the cleanup so the door stays closed behind us.

## Acceptance criteria

- [ ] All references to `redis_fallback` removed from configs, docs, helm chart, and test fixtures
- [ ] Misleadingly-named integration test renamed or removed
[...]

PR #4848, opened 2026-05-21
docs(rate-limiter): tighten plugin-bindings payload surface

gandhipratik203:
## Summary

- Trims the `RateLimiterPlugin` config-key surface in `docs/docs/api/plugin-bindings-api.md` from 8 keys to the 5 operator-tunable ones: `by_user`, `by_tenant`, `by_tool`, `algorithm`, `backend`.
- Removes `redis_url`, `redis_key_prefix` (gateway-scoped — set in `plugins/config.yaml`, not in binding payloads) and `redis_fallback` (not in the plugin's actual config schema; was a phantom field in the docs).
- Updates the `by_tool` example to use realistic gateway-prefixed tool names (`fast-time-get-system-time`, `fast-time-convert-time`) matching what operators actually see in `GET /tools`, rather than the previous abstract placeholders (`search`, `fetch_data`).

**Why these keys aren't operator-tunable:**

- **Security (`redis_url`):** Redis URLs typically contain auth credentials. Persisting them in the binding's `config` column stores those credentials in plaintext in Postgres, in DB backups, and in any query log capturing the column. Operators using the binding API have no signal that their input is being persisted unencrypted.
- **Operational coherence:** Redis topology (`redis_url`) and key-namespace layout (`redis_key_prefix`) are gateway-wide infrastructure decisions, set once in `plugins/config.yaml` by the operator who owns the deployment. Operators using the binding API set *policy* (rate limits per dimension); they don't (and shouldn't) decide which Redis the gateway connects to or how it lays out its key-space.

**Verified by the reproducer in PR #4674:** its `test_binding_full_lifecycle_inspectable` POSTs a binding **without** `redis_url` / `redis_key_prefix` and confirms (a) the binding's caller-scoped overrides reach the runtime, (b) all three dimension counter keys appear in Redis under the binding's tenant prefix, and (c) calls actually get rate-limited at the binding's tighter limit. Full `[inspect]` output is captured under "Latest test results" in that PR's description.

**Scope deliberately kept narrow:** no callout about gateway-scoped keys for now (per discussion on the parent issue). A separate "static-config-only parameters" section can be added later if it becomes useful. The current change focuses on giving operators a precise, copy-pasteable payload spec.

Addresses item 3 from #4665.

## Test plan

- [ ] Docs preview renders cleanly (table formatting, JSON examples are valid)
- [ ] No stale references to removed keys (`redis_url`, `redis_key_prefix`, `redis_fallback`) remain in `plugin-bindings-api.md`
- [ ] Other plugins' sections (`OutputLengthGuardPlugin`, `SecretsDetection`, `SQLSanitizer`) are untouched — only the `RateLimiterPlugin` section is modified


docs/docs/api/plugin-bindings-api.md
```diff
@@ -322,27 +322,21 @@ Rate strings use the format `<count>/<period>` where period is `s` (second) or `
...
-  "algorithm":       "fixed_window",
-  "backend":         "memory",
-  "redis_url":       null,
-  "redis_key_prefix": "rl",
-  "redis_fallback":  true
+  "by_user":   "60/m",
+  "by_tenant": "600/m",
+  "by_tool":   { "fast-time-get-system-time": "10/m", "fast-time-convert-time": "5/s" },
+  "algorithm": "fixed_window",
+  "backend":   "redis"
```
msureshkumar88: This changes the quick-look example from `"memory"` to `"redis"`, but the table below still shows `Default: "memory"` — contradiction introduced by this PR.

More practically: an operator copying this payload and setting `backend: redis` has no documented path to configure the Redis URL — `redis_url` was removed from this section. Two approaches worth considering:

1. Revert to `"backend": "memory"` to match the table default, or
2. Keep `"redis"` and add a cross-reference: *"Redis connection URL must be configured via `redis_url` in `plugins/config.yaml`"*

Same issue applies to the two curl examples further down (the first around line 359 and the second around line 540).

docs/docs/api/plugin-bindings-api.md
```diff
@@ -322,27 +322,21 @@ Rate strings use the format `<count>/<period>` where period is `s` (second) or `
...
-| `by_user`          | string \| null                                     | `null`           | `<int>/s` or `<int>/m`             | Rate limit per calling user; `null` disables                             |
-| `by_tenant`        | string \| null                                     | `null`           | `<int>/s` or `<int>/m`             | Rate limit per tenant (team); `null` disables                            |
-| `by_tool`          | object (string → string) \| null                  | `null`           | values: `<int>/s` or `<int>/m`     | Per-tool rate limits as a map of `tool_name → rate`; `null` disables     |
-| `algorithm`        | `"fixed_window"` \| `"sliding_window"` \| `"token_bucket"` | `"fixed_window"` | —                    | Counting algorithm to use                                                |
-| `backend`          | `"memory"` \| `"redis"`                           | `"memory"`       | —                                  | Storage backend for counters                                             |
```
msureshkumar88, line 342: Worth flagging: `RateLimiterConfig` (in `cpex_rate_limiter`) still accepts `redis_url` as a valid field — `TestRateLimiterConfig.test_overrides_are_applied` passes it explicitly and the assertion confirms it is set. Removing from docs addresses discoverability but not the plaintext-storage concern: an operator who discovers the field via code inspection or an older doc can still POST it in a binding payload and it will be accepted and persisted in the binding config column.

A couple of approaches to close the gap:
- Add a validation step in the binding API that rejects `redis_url`/`redis_key_prefix` in `RateLimiterPlugin` config payloads, or
- Add a prose note here: *"These keys are not supported in binding payloads and will be ignored — configure them in `plugins/config.yaml` instead"*

Not a hard block if the goal for this PR is just surface cleanup, but worth tracking as a follow-up if it's not addressed here.

docs/docs/api/plugin-bindings-api.md
```diff
@@ -322,27 +322,21 @@ Rate strings use the format `<count>/<period>` where period is `s` (second) or `
...
+|-------------|------------------------------------------------------------|------------------|--------------------------------|----------------------------------------------------------------------|
+| `by_user`   | string \| null                                             | `null`           | `<int>/s` or `<int>/m`         | Rate limit per calling user; `null` disables                         |
+| `by_tenant` | string \| null                                             | `null`           | `<int>/s` or `<int>/m`         | Rate limit per tenant (team); `null` disables                        |
+| `by_tool`   | object (string → string) \| null                           | `null`           | values: `<int>/s` or `<int>/m` | Per-tool rate limits as a map of `tool_name → rate`; `null` disables |
+| `algorithm` | `"fixed_window"` \| `"sliding_window"` \| `"token_bucket"` | `"fixed_window"` | —                              | Counting algorithm to use                                            |
```
msureshkumar88, line 339: The added note about replica state-sharing is a nice improvement. One gap: an operator who reads this and wants to use `redis` has no pointer to where the connection URL is configured. Worth adding something like: *"Redis connection URL is set via `redis_url` in `plugins/config.yaml`"* so the path is clear.

msureshkumar88 (changes requested):
The intent here is solid — narrowing the operator-facing surface to the 5 tunable keys is the right direction, and the security rationale for keeping `redis_url` out of binding payloads is well-articulated in the PR description.

Three things to resolve before this lands:

**Blocking**
1. All examples now use `"backend": "redis"` but `redis_url` was removed from this section — operator who copies any example has no documented path to configure the Redis connection. Need either a cross-reference to `plugins/config.yaml` or revert examples back to `"memory"`.
2. The table still shows `Default: "memory"` while all three examples show `"redis"` — contradiction introduced by this PR.

**Worth addressing**
3. `RateLimiterConfig` (in `cpex_rate_limiter`) still accepts `redis_url` as a valid field (`TestRateLimiterConfig.test_overrides_are_applied` passes it explicitly). Removing from docs addresses discoverability but not the storage-in-plaintext concern — operator who discovers the field can still POST it and it will be persisted in the binding config column. See inline comment for options.

docs/docs/api/plugin-bindings-api.md
msureshkumar88, line 339: `fail_mode` is missing from this table. It is a real operator-tunable field — `compat_default_config()` returns it as one of the 8 keys the Rust core recognises, and `RateLimiterConfig.__slots__` includes it:

| Field | Type | Default | Description |
|---|---|---|---|
| `fail_mode` | `"open"` \| `"closed"` | `"open"` | Behaviour when Redis is unreachable: `"open"` falls back to in-process memory counters; `"closed"` blocks requests until Redis recovers |

This is the hardening knob operators need when they set `backend: "redis"` — without it, anyone who wants enforcement guarantees (fail-closed) has no documented path to configure it. It should be the 6th row in this table.

msureshkumar88 (changes requested):
Followed up with a deeper inspection of the plugin runtime against the installed `cpex-rate-limiter==0.1.1` + `plugins/config.yaml`. Three additional findings beyond the earlier review.

**Finding 1 — `fail_mode` is a 6th operator-tunable key, completely absent from the table (blocking)**

`compat_default_config()` returns 8 keys the Rust core recognizes:

```
algorithm, backend, by_tenant, by_tool, by_user, fail_mode, redis_key_prefix, redis_url
```

`fail_mode` (default `'open'`) controls what happens when Redis is unreachable:
- `'open'` — silently fall back to in-process memory counters (availability-first)
- `'closed'` — block requests when Redis unreachable (enforcement-first)

This is a hardening decision operators need to make deliberately when setting `backend: 'redis'`. It belongs in the table as a 6th key. Omitting it means operators who want enforcement guarantees (fail-closed) have no documented path to configure it.

**Finding 2 — `plugins/config.yaml` has a stale `redis_fallback: true` entry that the Rust core silently ignores**

Out-of-diff but directly related: `plugins/config.yaml` still carries `redis_fallback: true`. The Rust core uses `fail_mode`, not `redis_fallback`. Passing `redis_fallback=True` to `RateLimiterConfig` leaves `fail_mode` at its default `'open'` — so fallback currently works correctly *by coincidence* (Rust default happens to match the intent), not by explicit configuration. The YAML should be updated to `fail_mode: "open"` to make the intent explicit and remove the dead key.

**Finding 3 — binding payload accepts `redis_url` without any key-level rejection; the restriction is doc-layer only**

Verified against the live merge path: `_merge_tenant_config` applies `{**yaml_config, **binding_config}` (right-wins). A binding that includes `redis_url` overwrites the YAML's Jinja-resolved URL — the override value is both persisted in the binding's `config` column and used at runtime. Documentation removal alone doesn't harden this path. A validator that rejects disallowed keys in the binding payload (or at minimum an explicit allowlist check against the 5 operator-tunable keys) would make the restriction enforceable rather than advisory.

gandhipratik203:
Thanks for the careful re-review, @msureshkumar88. Two passes of feedback was really useful. Fixes are in commit 509313f68 (just pushed). Item by item:

**B1 & B2 (`backend: "redis"` examples without `redis_url` path + table-default contradiction).** Both fixed. Flipped the `backend` table default from `"memory"` to `"redis"` so the table matches the examples, and added a one-line cross-reference under the table explaining where `redis_url` is configured (`plugins/config.yaml`) and noting that binding-API users should not include `redis_url` or `redis_key_prefix` in payloads.

**Finding 1 (`fail_mode` missing from the table).** Fixed. Added as the 6th operator-tunable key in the table and all three example payloads (JSON schema, single-plugin curl, multi-plugin curl). Defaults to `"open"`, with a short note documenting the fail-open vs. fail-closed trade-off.

**Worth addressing 3 + Finding 3 (`RateLimiterConfig` accepts `redis_url` at runtime; restriction is doc-layer only).** You're right that the docs-only fix doesn't close the plaintext-storage gap. The proper fix is API-layer rejection at write time, before persistence, which is already tracked in #4622. Doing this at the plugin-config layer would be too late for the security concern, since the credentials would already be in Postgres by then. So the docs PR addresses discoverability and #4622 closes the enforcement gap.

While I was in there I also tightened the cross-reference wording from "(and cannot) override it per-tenant" to "should not override it per-tenant", so it reads as prescriptive guidance rather than a false enforcement claim.

**Finding 2 (`plugins/config.yaml` has stale `redis_fallback: true`).** Already tracked in #4603, which covers the parallel cleanup across `plugins/config.yaml`, docs, helm chart, and tests. Out of this PR's scope.

msureshkumar88 (commented):
Good recovery — all three blockers from round 1 and all three findings from round 2 are addressed in commit 509313f68. Two items before this lands:

**Blocking**

The `backend` table default was flipped from `"memory"` to `"redis"` to match the examples. That's internally consistent, but it needs to match the runtime. `compat_default_config()` in `cpex-rate-limiter==0.1.1` — what does it return for `backend`? If the Rust core default is still `"memory"`, the table now has a doc/runtime mismatch in the other direction. Worth verifying against the installed package before merge:

```python
from cpex_rate_limiter import compat_default_config
print(compat_default_config().get("backend"))
```

**Nit**

`fail_mode` description doesn't scope its applicability — `"open"" | "closed"` only matters when `backend` is `"redis"`. When `backend` is `"memory"`, the field is silently ignored. Suggest adding `(ignored when backend is "memory")` at the end of the description so operators don't wonder why setting `fail_mode: "closed"` has no effect in a local deployment.

The deferred items (#4622 for runtime enforcement, #4603 for `plugins/config.yaml` stale `redis_fallback`) are both reasonable scope splits. No objection there.

gandhipratik203:
Both round-3 items addressed:

- **Blocking** (`4fd9d02ea`): table now shows `Default: "memory"` — verified runtime via `RateLimiterConfig()` and `cpex-plugins/plugins/rust/python-package/rate_limiter/src/lib.rs:31` (`defaults.set_item("backend", "memory")?;`). Examples retain `"redis"` as the production-typical setup per your earlier approval.
- **Nit** (`7cea1895d`): appended `"; ignored when backend is memory"` to the `fail_mode` description so operators don't try setting `fail_mode: "closed"` in a memory-only deployment.

Ready for re-review.

docs/docs/api/plugin-bindings-api.md
gandhipratik203: Went with option 2 in `509313f68` — added the cross-reference note about `redis_url` configuration immediately after the table. The table default was later corrected to `"memory"` in `4fd9d02ea` per your round-3 follow-up, so the cross-reference plus the runtime-aligned default now together cover the gap.

docs/docs/api/plugin-bindings-api.md
gandhipratik203: Addressed in `509313f68` — added a note immediately after the table: *"When `backend` is `"redis"`, the Redis connection URL is configured via `redis_url` in `plugins/config.yaml` — it is **not** a binding payload field. Gateway operators set it once at deployment time; binding-API users should not override it per-tenant."*

docs/docs/api/plugin-bindings-api.md
gandhipratik203: Scoped to a follow-up per your round-3 acknowledgment. Runtime enforcement is tracked in #4622. This PR remains surface cleanup; the prose note added in `509313f68` advises operators not to override these keys per-tenant, which is the doc-layer half of the fix.

msureshkumar88 (approved):
All three rounds of blocking items resolved — ready to merge.

## What's been fixed

**Round 1 blockers:**
- `redis_url` removed from binding payload surface; cross-reference added (line 343) directing operators to `plugins/config.yaml`
- Table-vs-example contradiction eliminated

**Round 2 blockers:**
- `fail_mode` added as 6th operator-tunable key — correct, `compat_default_config()` returns it; operators need it when `backend: redis`
- Redis URL configuration path documented

**Round 3 blocking + nit:**
- `backend` table default reverted to `"memory"` — aligned with runtime (`src/lib.rs:31`)
- `fail_mode` description scoped: "ignored when `backend` is `"memory"`"

## Minor remaining item (not blocking)

Commits reference "5 keys" — contributor acknowledged as housekeeping. Doesn't affect doc content; can stay in commit history.

## Cross-impact note for awareness

PR #4859 (separate Redis for rate limiter middleware) is in-flight. Two Redis URL concepts for rate limiting will exist after both land:

| Layer | Mechanism | Redis URL source |
|---|---|---|
| Plugin-binding rate limiting | `cpex_rate_limiter` Rust plugin, per-tool/per-tenant limits via `POST /v1/tools/plugin_bindings` | `redis_url` in `plugins/config.yaml` |
| Middleware-level rate limiting | `RateLimitMiddleware` in `auth.py` | `RATELIMITER_REDIS_URL` env var (PR #4859) |

This PR's line 343 cross-reference is correctly scoped to the plugin layer — no change needed here. However, #4859's docs should clarify the two-layer architecture so operators don't conflate the two Redis URL configurations. Filed as a finding on #4859 separately.

PR #4857, opened 2026-05-21
test(rate-limiter): add end-to-end tests for binding lifecycle flow

gandhipratik203:
## Summary

End-to-end tests for the rate-limiter binding flow against a running docker stack. Three tests walk a binding through `POST` / `UPSERT` / `DELETE` and verify each step propagates correctly through every layer — API, Postgres, the gateway plugin manager, Redis — and changes runtime enforcement behaviour on tool dispatch.

| Test | Class | What it verifies |
|---|---|---|
| `test_binding_full_lifecycle_inspectable` | `TestRateLimiterBindingApiEnforcesLimits` | POST happy path — binding is created, persisted (API + Postgres cross-check), and enforces at the binding's tight `by_user: 7/m` limit at dispatch with all three dimension keys present in Redis |
| `test_upsert_binding_full_lifecycle_inspectable` | `TestRateLimiterBindingModeAndLifecycle` | UPSERT — second POST with the same `(team, tool, plugin)` triple updates the existing row in place AND propagates the new mode past the DB into the plugin manager (`enforce → disabled` flip verified at dispatch) |
| `test_delete_binding_full_lifecycle_inspectable` | `TestRateLimiterBindingModeAndLifecycle` | DELETE — binding removal is verified on every write surface it landed on (API + Postgres) |

## Coverage gap closed

Existing test layers each cover a slice of the binding flow but none exercise it end-to-end:

- `tests/unit/mcpgateway/routers/test_tool_plugin_bindings.py` — exercises the binding router in isolation; plugin invocation is mocked.
- `tests/integration/test_rate_limiter.py` + sibling rate-limiter integration tests — exercise the plugin against real Redis but instantiate it directly, bypassing the gateway's binding API.
- **This PR** — covers binding API → gateway plugin manager → plugin instantiation with the binding's config → enforcement at tool dispatch, against the real docker stack.

## Where it lives

`tests/live_gateway/plugins/test_rate_limiter_binding_lifecycle.py`.

The `tests/live_gateway/` tree is purpose-built for tests that require a running gateway + side services. Runs as part of `make test-live-gateway` and skips automatically when the gateway isn't reachable at `GATEWAY_URL`. Not invoked by default `make test`.

## Runtime

~50 seconds for the full 3-test suite (~17 seconds per test). The structurally necessary wait is `PROPAGATION_WAIT` (~7s per phase that needs the per-tenant plugin manager to rebuild after a binding change).

## What each test walks through

Same shape for all three, with phase 6's action varying (POST → assert / UPSERT → flip / DELETE → confirm removed):

1. **Baseline capture** — `GET /admin/plugins/RateLimiterPlugin` to log what's currently mounted.
2. **POST binding** with deliberately uncommon, distinct values for all three dimensions (`by_user: "7/m"`, `by_tenant: "9/m"`, `by_tool: {<tool>: "11/m"}`) so each value is recognisable in API responses, DB rows, and Redis keys.
3. **Persistence cross-check** — fetched both via `GET /v1/tools/plugin_bindings/?binding_reference_id=…` (API) and via direct `docker exec psql` against the `tool_plugin_bindings.config` JSON column (DB). Both must return the binding's exact values for all three dimensions.
4. **Propagation wait** — covers the per-tenant plugin manager rebuild.
5. **Tool-call burst** — 5 back-to-back MCP `tools/call` requests against the bound server.
6. **Behavioural / lifecycle assertion** — `rate_limited >= 1` (POST), mode flip verified (UPSERT), or write-surface deletion confirmed (DELETE).
7. **Cleanup** — via the `cleanup_bindings` fixture.

## Caveats / known flakiness

- **Tool-path amplification varies between runs** (~5×–20× plugin hook invocations per user-level call; see #4557). The behavioural assertion only requires `rate_limited >= 1`, accommodating both shapes:
  - **High-amplification runs**: every call blocks on the very first plugin invocation (`0 allowed / 5 blocked`)
  - **Low-amplification runs**: 1–4 calls slip through before the binding's tight 7/m trips (`1–4 allowed / N blocked`)
- **Per-tenant manager propagation** across replicas via Redis pub/sub is not always instantaneous; the `PROPAGATION_WAIT` window covers it. Same flake any binding-dependent test would hit, not specific to this one.

## Related

- #4665 — parent issue tracking the binding-config-flow gap; this PR addresses item 2 (end-to-end test for the binding lifecycle)

## Test plan

- [ ] `make test-live-gateway` collects and runs the 3 new tests
- [ ] Full 3-test suite passes in under ~60 seconds against the standard docker-compose stack
- [ ] When the gateway isn't running, the tests skip cleanly with the expected reason rather than failing
- [ ] The autouse Redis-isolation fixture flushes `plugin:RateLimiterPlugin:mode` + `rl:*` before each test (verified by stable cross-test pass rate)

### DIFF
--- a/tests/live_gateway/plugins/__init__.py
+++ b/tests/live_gateway/plugins/__init__.py
@@ -1,0 +1,11 @@
+# -*- coding: utf-8 -*-
+"""Location: ./tests/live_gateway/plugins/__init__.py
+Copyright 2026
+SPDX-License-Identifier: Apache-2.0
+Authors: ContextForge Contributors
+
+Live-gateway tests for the plugin runtime (binding lifecycle, runtime
+management, mode propagation, etc.). These exercise the full HTTP →
+gateway plugin manager → plugin → Redis path against a running docker
+stack and are gated on a reachable gateway via _is_gateway_running().
+"""
--- a/tests/live_gateway/plugins/test_rate_limiter_binding_lifecycle.py
+++ b/tests/live_gateway/plugins/test_rate_limiter_binding_lifecycle.py
@@ -769,12 +769,15 @@ class TestRateLimiterBindingApiEnforcesLimits:
         burst_size = 5
 
         # Single MCP session for the full burst — keeps tenant_id resolution stable.
-        session_id = _mcp_initialize_session(server_id, _fresh_headers())
+        # Reuse one auth header for the handshake + every burst call instead of
+        # paying for two `/auth/login` round-trips per test.
+        base_headers = _fresh_headers()
+        session_id = _mcp_initialize_session(server_id, base_headers)
         assert session_id is not None, (
             "MCP initialize handshake failed — can't drive the burst without a session id"
         )
         call_headers = {
-            **_fresh_headers(),
+            **base_headers,
             "Accept": "application/json, text/event-stream",
             "Mcp-Session-Id": session_id,
         }
@@ -965,12 +968,15 @@ class TestRateLimiterBindingModeAndLifecycle:
 
         def _paced_burst(phase_label: str) -> dict:
             """Run a paced burst over a fresh MCP session; tally outcomes."""
-            session_id = _mcp_initialize_session(server_id, _fresh_headers())
+            # Reuse one auth header for the handshake + every burst call instead of
+            # paying for two `/auth/login` round-trips per burst.
+            base_headers = _fresh_headers()
+            session_id = _mcp_initialize_session(server_id, base_headers)
             assert session_id is not None, (
                 f"{phase_label}: MCP initialize failed — can't drive the burst"
             )
             call_headers = {
-                **_fresh_headers(),
+                **base_headers,
                 "Accept": "application/json, text/event-stream",
                 "Mcp-Session-Id": session_id,
             }
@@ -1272,12 +1278,15 @@ class TestRateLimiterBindingModeAndLifecycle:
         # ---- Phase 4: paced burst with per-call observation ----------------
         burst_size = 5
 
-        session_id = _mcp_initialize_session(server_id, _fresh_headers())
+        # Reuse one auth header for the handshake + every burst call instead of
+        # paying for two `/auth/login` round-trips per test.
+        base_headers = _fresh_headers()
+        session_id = _mcp_initialize_session(server_id, base_headers)
         assert session_id is not None, (
             "MCP initialize handshake failed — can't drive the burst without a session id"
         )
         call_headers = {
-            **_fresh_headers(),
+            **base_headers,
             "Accept": "application/json, text/event-stream",
             "Mcp-Session-Id": session_id,
         }
--- a/tests/live_gateway/plugins/test_rate_limiter_binding_lifecycle.py
+++ b/tests/live_gateway/plugins/test_rate_limiter_binding_lifecycle.py
@@ -246,16 +246,22 @@ def _stamp_tool_team_id(tool_id: str, team_id: str) -> str:
     than letting the actual assertions fail with a confusing "no
     requests were rate-limited" message that hides the real cause.
     """
-    # Read the current value so we can restore it on teardown.
-    select_sql = f"SELECT COALESCE(team_id, '') FROM tools WHERE id = '{tool_id}';"
+    # Read the current value so we can restore it on teardown. Uses psql's `-v`
+    # variable substitution with the `:'name'` form so the values are quoted as
+    # SQL string literals by psql itself, not interpolated into the SQL text.
+    # NOTE: psql only processes `-v` substitution when reading SQL from stdin
+    # or a file (`-f`), NOT from `-c`. We pipe the SQL via `input=` and use
+    # `docker exec -i` to keep stdin open.
+    select_sql = "SELECT COALESCE(team_id, '') FROM tools WHERE id = :'tool_id';"
     cmd_select = [
-        "docker", "exec", PG_CONTAINER,
+        "docker", "exec", "-i", PG_CONTAINER,
         "psql", "-U", PG_USER, "-d", PG_DATABASE,
-        "-tAc", select_sql,
+        "-v", f"tool_id={tool_id}",
+        "-tA",
     ]
     try:
         prev = subprocess.run(
-            cmd_select, capture_output=True, text=True, timeout=10, check=True
+            cmd_select, input=select_sql, capture_output=True, text=True, timeout=10, check=True
         ).stdout.strip()
     except (subprocess.CalledProcessError, subprocess.TimeoutExpired, FileNotFoundError) as exc:
         pytest.skip(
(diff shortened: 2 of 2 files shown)


msureshkumar88 (commented):
Good coverage of an untested gap — the three-phase walkthrough (POST / UPSERT / DELETE) against a live stack is the right level to exercise the binding API → plugin manager → Redis path. CI is green. A few issues worth fixing before merge.

---

**M1 — `_redis_rl_keys()` scan unscoped (flakiness risk)**

L451, L625: `--pattern rl:*` fetches every rate-limiter key in Redis. The `cleanup_bindings` fixture DELs stale keys at the start of each test, but that DEL runs _after_ the fixture is entered — it doesn't protect against concurrent runs or keys written by other modules between the DEL and Phase 5. The Phase 5 assertion (`len(user_keys) >= 1`) can pass on a key from a different test's team.

Fix: scope the scan to the current team:

```python
pattern = f"rl:{team_id}:*"
```

`team_id` is available on the fixture; thread it into `_redis_rl_keys()`.

---

**M2 — Two `/auth/login` POSTs before the burst (minor auth waste)**

L795: `_mcp_initialize_session(server_id, _fresh_headers())` — one login.
L800: `call_headers = {**_fresh_headers(), ...}` — second login, immediately after.

These two calls produce two session tokens in the DB; only one is used for the burst. Cache the result:

```python
headers = _fresh_headers()
session_id = _mcp_initialize_session(server_id, headers)
...
call_headers = {**headers, "Accept": ..., "Mcp-Session-Id": session_id}
```

Same pattern appears in the upsert test (L991–L996) and delete test (L1298–L1303).

---

**M3 — Fixed `time.sleep` for propagation wait (flaky on slow CI)**

L876, L1329, L1429 (and `post_propagation_pause` addend): a fixed sleep assumes the per-tenant manager rebuild completes within `PROPAGATION_WAIT + 10s`. A slow runner or a heavily loaded CI node can exceed the window, turning a valid test into a flaky one.

A polling approach is more reliable and usually faster:

```python
deadline = time.monotonic() + PROPAGATION_WAIT + post_propagation_pause
while time.monotonic() < deadline:
    binding = _get_binding_via_api(ref_id)
    if binding is not None:  # (or is None for the delete test)
        break
    time.sleep(0.5)
```

Phase 2 already confirms DB persistence; this pattern confirms plugin-manager propagation rather than assuming it fits a fixed window.

---

**M4 — f-string SQL in `_stamp_tool_team_id` / `_restore_tool_team_id` (style)**

L323, L340, L353, L355: UUIDs make injection practically impossible here, but f-string SQL is a bad pattern even in test helpers. Use psql's `-v` substitution:

```python
update_sql = "UPDATE tools SET team_id = :'team_id' WHERE id = :'tool_id'; "
cmd = [
    "docker", "exec", PG_CONTAINER,
    "psql", "-U", PG_USER, "-d", PG_DATABASE,
    "-v", f"team_id={team_id}", "-v", f"tool_id={tool_id}",
    "-c", update_sql,
]
```

---

**Nit — `post_propagation_pause = 10` / `post_burst_pause = 30` duplicated**

These literals appear inside each of the three test methods. Promote to module-level constants alongside `PROPAGATION_WAIT` so they're tunable from one place.

msureshkumar88:
## Speed, gaps, and test structure

Three follow-up observations from reading the full test body.

---

### 1. Actual runtime is ~3–4 minutes, not ~50 seconds

The PR description says "~50 seconds for the full 3-test suite". That's only true if the inspection pauses are excluded. The real sleep budget per test:

| Sleep | Per test | × 3 tests |
|---|---|---|
| `PROPAGATION_WAIT` (7s) | 7s | 21s |
| `post_propagation_pause = 10` (hardcoded, on top) | 10s | 30s |
| burst `pace_between_calls = 3s × 5 calls` | 15s | 45s |
| `post_burst_pause = 30s` | 30s | 90s |
| **Total** | **~62s** | **~3 min+** |

The test docstring itself says *"Designed for manual eyeballing alongside Redis Insight, not for CI"* — but `post_burst_pause` and `post_propagation_pause` run unconditionally regardless of whether a human is watching.

**Fix:** gate inspection pauses on `INSPECT=1`:

```python
INSPECT = bool(os.getenv("INSPECT"))

# Phase 4 pre-burst pause (currently unconditional):
if INSPECT:
    time.sleep(post_propagation_pause)   # saves 10s × 3 = 30s

# Phase 6 post-burst pause (currently unconditional):
if INSPECT:
    time.sleep(post_burst_pause)         # saves 30s × 3 = 90s

# pace between burst calls:
pace_between_calls = 0.5 if not INSPECT else 3   # saves ~12s × 3 = 36s
```

After gating: **~25–30s total** for the suite (structurally required `PROPAGATION_WAIT × 3` = 21s + burst). That matches the claimed "~50s" with some headroom.

---

### 2. Prerequisites not validated at test start

Several conditions are required but not checked before the test body runs — failures surface as confusing assertion errors instead of clean skips:

| Missing check | What breaks without it |
|---|---|
| `PLUGINS_ENABLED=true` in gateway config | Enforcement never fires; `rate_limited >= 1` fails with no signal that plugins are off |
| `RateLimiterPlugin` present in `plugins/config.yaml` | `_get_admin_plugin_state()` returns a state but the plugin isn't loaded; binding enforcement silently no-ops |
| `BINDING_REDIS_URL` is the **gateway-internal** URL (default `redis://redis:6379/0`), not the host-side URL | Easy to misconfigure; setting `redis://localhost:6379` sends the wrong URL in the binding payload and the plugin can't reach Redis |
| `docker exec` access available | CI runners with rootless Docker or a remote daemon `FileNotFoundError` on `subprocess.run(["docker", "exec", ...])` — skip message says "set PG_CONTAINER" but the real cause is no exec access |

Also: `_is_gateway_running()` is defined but never used as a collection-time skip guard. Tests skip via fixture failure instead, which means pytest collects and starts the test before discovering the gateway is down. A module-level guard is cleaner:

```python
pytestmark = pytest.mark.skipif(
    not _is_gateway_running(),
    reason=f"gateway not reachable at {GATEWAY_URL}",
)
```

---

### 3. POST/UPSERT/DELETE in separate tests is the right call

The 3-test, 2-class structure is correct. Merging into one test would make it impossible to run the upsert path in isolation and produce an untraceable 300-line body. The split between `TestRateLimiterBindingApiEnforcesLimits` and `TestRateLimiterBindingModeAndLifecycle` is also sensible.

One consolidation worth making: `post_propagation_pause = 10` and `post_burst_pause = 30` are duplicated inside each test method. Extract to module-level constants alongside `PROPAGATION_WAIT` so tuning is one-place:

```python
POST_PROPAGATION_PAUSE = int(os.environ.get("POST_PROPAGATION_PAUSE", "10"))
POST_BURST_PAUSE = int(os.environ.get("POST_BURST_PAUSE", "30"))
```

Making them env-overridable also lets CI set `POST_BURST_PAUSE=0` without touching the test file.

msureshkumar88 (changes requested):
Good structure and the right level to close the coverage gap. Two blocking issues need fixing before merge; the rest are hardening.

---

## Blocking

**B1 — Inspection sleeps run unconditionally in CI**

`post_burst_pause = 30` and `post_propagation_pause = 10` are hardcoded inside each test method with no guard. The PR claims "~50 seconds" but the real CI budget is ~3–4 minutes (30s × 3 tests + 10s × 3 tests + 3s pace × 5 calls × 3 tests on top of `PROPAGATION_WAIT`). The docstring says "Designed for manual eyeballing, not for CI" — the code doesn't enforce that distinction.

Fix: gate inspection pauses on an env var and make the constants env-overridable:

```python
INSPECT = bool(os.getenv("INSPECT"))
POST_PROPAGATION_PAUSE = int(os.environ.get("POST_PROPAGATION_PAUSE", "10"))
POST_BURST_PAUSE = int(os.environ.get("POST_BURST_PAUSE", "30"))
pace_between_calls = 0.5 if not INSPECT else 3

# in test body:
if INSPECT:
    time.sleep(POST_PROPAGATION_PAUSE)
if INSPECT:
    time.sleep(POST_BURST_PAUSE)
```

CI runs with defaults (`INSPECT` unset) → ~25–30s total. Human runs with `INSPECT=1` → full pauses for Redis Insight observation.

**B2 — `_redis_rl_keys()` scan unscoped to current `team_id`**

`--pattern rl:*` fetches every rate-limiter key in Redis. The autouse fixture flushes stale keys at test *start*, but that doesn't protect against concurrent runs or keys written between the flush and Phase 5. `len(user_keys) >= 1` can pass on a key from a different test's team, making the enforcement assertion meaningless.

Fix: scope the pattern to the current test's team:

```python
def _redis_rl_keys(team_id: str) -> list[tuple[str, str, str]]:
    pattern = f"rl:{team_id}:*"
    ...
```

Thread `team_id` through all call sites.

---

## Should fix (flakiness / correctness)

**S1 — No `PLUGINS_ENABLED` guard**

If the gateway runs with `PLUGINS_ENABLED=false` (the config default), plugin enforcement never fires and `rate_limited >= 1` fails with no diagnostic. Add a skip at test start:

```python
state = _get_admin_plugin_state(PLUGIN_NAME)
if not state:
    pytest.skip(f"Plugin {PLUGIN_NAME!r} not loaded — is PLUGINS_ENABLED=true?")
```

**S2 — No collection-time skip guard**

`_is_gateway_running()` is defined but unused as a skip guard. Tests fail (not skip) when the gateway is down. Add at module level:

```python
pytestmark = pytest.mark.skipif(
    not _is_gateway_running(),
    reason=f"gateway not reachable at {GATEWAY_URL}",
)
```

**S3 — Fixed `PROPAGATION_WAIT` sleep is flaky on slow CI**

`time.sleep(PROPAGATION_WAIT)` assumes the per-tenant manager rebuild completes within 7s. A loaded CI runner can exceed this. Poll instead:

```python
deadline = time.monotonic() + PROPAGATION_WAIT + 5
while time.monotonic() < deadline:
    if _get_binding_via_api(ref_id) is not None:
        break
    time.sleep(0.5)
```

**S4 — Double `/auth/login` before each burst**

`_mcp_initialize_session(server_id, _fresh_headers())` on one line and `call_headers = {**_fresh_headers(), ...}` on the next — two login POSTs where one token suffices. Cache:

```python
headers = _fresh_headers()
session_id = _mcp_initialize_session(server_id, headers)
call_headers = {**headers, "Accept": ..., "Mcp-Session-Id": session_id}
```

Same pattern in all three test methods.

---

## Cleanup (non-blocking)

- Extract `post_propagation_pause`/`post_burst_pause` to module-level constants (currently duplicated in every test method).
- Replace f-string SQL in `_stamp_tool_team_id`/`_restore_tool_team_id` with psql `-v` substitution — values are UUIDs so injection risk is nil, but the pattern is worth fixing.
- Document that `BINDING_REDIS_URL` is the **gateway-internal** hostname (`redis://redis:6379/0`), not the host-side URL — easy to misconfigure.

PR #4859, opened 2026-05-21
separate redis for ratelimiter is enabled

### LINKED ISSUE #4751: [FEATURE]: Separate Redis URL for Rate Limiter Throughout.
### 🧭 Type of Feature

### 🧭 Epic

**Title:** Isolated Redis Instance for Rate Limiter in both the Plugin and the middleware.
**Goal:** Separate rate limiter Redis from main Redis. Independent scaling, isolation, failure containment.
**Why now:** Rate limiter high-traffic. Shared Redis = contention. Separate = better performance, ops flexibility.

---

### 🧑🏻‍💻 User Story 1

**As a:** Platform operator
**I want:** Separate Redis URL for rate limiter plugin and ratelimiter middleware.
**So that:** Rate limiting doesn't impact main Redis (cache, sessions, federation)

#### ✅ Acceptance Criteria
```gherkin
Scenario: Configure separate rate limiter Redis
  Given RATELIMITER_REDIS_URL env var set
  When rate limiter plugin/middleware initializes
  Then connects to separate Redis instance
  And main Redis URL unchanged

Scenario: Fallback to main Redis
  Given RATELIMITER_REDIS_URL not set
  When rate limiter plugin/middleware initializes
  Then uses REDIS_URL (backward compatible)
```

---

### 🧑🏻‍💻 User Story 2

**As a:** Kubernetes operator
**I want:** RATELIMITER_REDIS_URL from K8s Secret
**So that:** Credentials secure, rotatable

#### ✅ Acceptance Criteria
```gherkin
Scenario: Load from K8s Secret
  Given RATELIMITER_REDIS_URL in Secret
  When Helm chart deployed
  Then env var populated from Secret
  And rate limiter connects successfully
```

---

```mermaid
flowchart LR
    A[MCPGateway] -->|Cache/Sessions| B[Main Redis]
    A -->|Rate Limits| C[Rate Limiter Redis]
    D[K8s Secret] -.->|RATELIMITER_REDIS_URL| A
```

**Config**:
```python
# mcpgateway/config.py
RATELIMITER_REDIS_URL: str = Field(
    default="",  # Empty = use REDIS_URL
    description="Redis URL for rate limiter plugin (optional)/middleware"
)
```

**Plugin**:
```python
# plugins/rate_limiter/rate_limiter.py
redis_url = settings.RATELIMITER_REDIS_URL or settings.REDIS_URL
self.redis = Redis.from_url(redis_url)
```
**Middleware**:
```python
# mcpgateway/middleware/rate_limit_middleware.py
redis_url = settings.RATELIMITER_REDIS_URL or settings.REDIS_URL
self.redis = Redis.from_url(redis_url)
```

---

### 🔗 MCP Standards Check

N/A - Internal infrastructure change. No MCP protocol impact.

---

### 🔄 Alternatives Considered

1. **Redis DB number separation**: Same instance, different DB. Rejected - still shares connection pool, memory.
2. **Redis Cluster**: Over-engineered for this use case. Separate instance simpler.
3. **In-memory rate limiter**: No persistence. Rejected - multi-instance deployments need shared state.

---

### 📓 Additional Context

**Related**:
- Rate limiter plugin: `plugins/rate_limiter/`
- Rate limiter middleware: `mcpgateway/middleware/rate_limit_middleware.py`
- Main Redis config: `mcpgateway/config.py:REDIS_URL`
- Helm chart: `charts/mcp-stack/values.yaml`

**Implementation checklist**:
- [ ] Add `RATELIMITER_REDIS_URL` to `mcpgateway/config.py`
- [ ] Update rate limiter plugin and middleware to use new URL (fallback to `REDIS_URL`)
- [ ] Add to `.env.example` with comment
- [ ] Update Helm chart `values.yaml` + templates
- [ ] Add K8s Secret example to `charts/mcp-stack/README.md`
- [ ] Update `docs/docs/manage/configuration.md`
- [ ] Add migration guide for existing deployments

ja8zyjits:
# ✨ Feature / Enhancement PR

## 🔗 Epic / Issue
Closes #4751

---

## 🚀 Summary (1-2 sentences)
_What does this PR add or change?_

Adds support for a dedicated Redis URL (`RATELIMITER_REDIS_URL`) for the rate limiting middleware, allowing rate limit state to be stored in a separate Redis instance from the main application cache. Falls back to the main `REDIS_URL` when not configured, maintaining full backward compatibility.

---

## 🧪 Checks

- [x] `make lint` passes
- [x] `make test` passes (122/122 rate limit middleware tests passing, 18,718 total tests passing)

---

## 📓 Notes (optional)
_Design sketch, screenshots, or extra context._

### Implementation Details

**Configuration:**
- New environment variable: `RATELIMITER_REDIS_URL` (optional)
- Backward compatible: Falls back to `REDIS_URL` if not set
- No breaking changes to existing deployments

**Changes:**
1. Added `rate_limiter_redis_url` to `mcpgateway/config.py`
2. Updated `RateLimitMiddleware` to use dedicated Redis URL when configured
3. Added comprehensive unit tests for both dedicated and fallback scenarios
4. Fixed test fixture scope issue in `test_get_client_ip_with_both_headers`

**Testing:**
- All 122 rate limit middleware tests passing
- Tests cover both dedicated Redis URL and fallback scenarios
- Tests verify Redis connection initialization and error handling

**Use Cases:**
- Isolate rate limiting state from application cache
- Use different Redis configurations for rate limiting (e.g., different persistence, memory limits)
- Scale rate limiting independently from main cache
- Improve observability by separating rate limit metrics

### Architecture

```mermaid
flowchart TD
    A[RateLimitMiddleware] -->|RATELIMITER_REDIS_URL set?| B{Config Check}
    B -->|Yes| C[Dedicated Redis Instance]
    B -->|No| D[Main Redis Instance]
    C --> E[Rate Limit State]
    D --> E
    E --> F[Request Processing]
```

**Backward Compatibility:**
- Existing deployments continue to work without changes
- No migration required
- Optional feature that can be enabled when needed


tests/integration/test_rate_limiter_redis_url.py
```diff
@@ -0,0 +1,129 @@
...
+"""
+
+import pytest
+from unittest.mock import patch, MagicMock
+
+
+class TestRateLimiterWithDedicatedRedis:
+    """End-to-end test with dedicated Redis for rate limiting."""
+
+    @patch("mcpgateway.auth.redis")
```
madhu-mohan-jaishankar: looks like redis is not a module level attribute in auth.py. shouldn't be patching this way.

msureshkumar88 (changes requested):
## Review — Request for Changes

CI passes. Feature intent is sound. The implementation has one architectural correctness bug and several hardening gaps that need addressing before merge.

---

### 🔴 BLOCKING

**`mcpgateway/auth.py:755–772` — Singleton global routes to wrong Redis when feature is used**

`_get_sync_redis_client()` stores a single global `_SYNC_REDIS_CLIENT`. The PR redirects it to `ratelimiter_redis_url` when set — but every other caller of this function, including `_update_api_token_last_used_sync()` (auth.py:804), now also receives the rate-limiter Redis instead of the main Redis. Auth token rate-limit tracking data is written to the dedicated rate-limiting instance, contaminating it. The isolation goal of the feature is broken by the implementation.

**Required fix:** Introduce `_RATELIMITER_REDIS_CLIENT` global + `_get_ratelimiter_redis_client()` function. Have `rate_limit_middleware._init_redis()` call that new function. Leave `_get_sync_redis_client()` unchanged — it must continue serving the main Redis exclusively.

```python
# auth.py — alongside existing globals
_RATELIMITER_REDIS_CLIENT = None
_RATELIMITER_REDIS_LOCK = threading.Lock()
_RATELIMITER_REDIS_FAILURE_TIME: Optional[float] = None

def _get_ratelimiter_redis_client():
    # mirrors _get_sync_redis_client() but uses ratelimiter_redis_* fields
    # falls back to _get_sync_redis_client() when ratelimiter_redis_url is None

# middleware/rate_limit_middleware.py _init_redis():
client = auth._get_ratelimiter_redis_client()  # not _get_sync_redis_client
```

---

**`mcpgateway/auth.py:735` — `cache_type == "redis"` guard silently kills dedicated rate-limiter Redis**

```python
if _SYNC_REDIS_CLIENT is not None or not (
    config_settings.redis_url and config_settings.redis_url.strip()
    and config_settings.cache_type == "redis"   # ← blocks dedicated path
):
    return _SYNC_REDIS_CLIENT
```

Operator sets `CACHE_TYPE=database` + `RATELIMITER_REDIS_URL=redis://rl-host/0`. The guard returns `None` before reaching the rate-limiter branch — rate limiting silently degrades to in-memory with no log warning. The dedicated client must bypass the `cache_type` guard. The separate-function approach above resolves this naturally.

---

### 🔴 HARDENING REQUIRED

**`mcpgateway/auth.py:772` — Redis credentials written to INFO log**

```python
log.getLogger(__name__).info(f"Rate limiter using dedicated Redis: {redis_url}")
```

Redis URLs routinely carry passwords: `redis://:s3cr3t@host:6379/0`. At INFO level this credential is emitted to every log sink — structured logs, SIEM, disk. Strip credentials before logging using `urllib.parse.urlparse(redis_url)._replace(password="***").geturl()` or the existing `sanitize_for_log()` utility already used elsewhere in this codebase.

---

**`mcpgateway/auth.py` (TLS warning block) — Wrong SSL flag checked for rate-limiter URL**

```python
if redis_url.startswith("rediss://") and not config_settings.redis_ssl:
    log.getLogger(__name__).warning("REDIS_URL uses rediss:// ...")
```

When `redis_url` is `ratelimiter_redis_url`, the check reads `config_settings.redis_ssl` (the main-Redis flag). There is no `RATELIMITER_REDIS_SSL` setting, so `_build_ssl_kwargs(config_settings)` applies main-Redis TLS settings to the dedicated connection. A `rediss://` rate-limiter URL with `REDIS_SSL=false` could silently skip certificate verification on the dedicated connection. Either add `ratelimiter_redis_ssl` config fields, or document explicitly that the dedicated URL inherits main-Redis TLS settings and validate that intent.

---

**`mcpgateway/config.py:2576` — No URL format validation on `ratelimiter_redis_url`**

`ratelimiter_redis_url: Optional[str]` accepts any string. A misconfigured value (e.g., `redis//localhost` — missing colon) silently fails at runtime, leaving rate limiting on in-memory fallback with no startup error. Add a Pydantic validator rejecting non-None values that do not start with `redis://` or `rediss://`.

---

### 🟡 RISK

**`mcpgateway/auth.py:739,771` — Single failure backoff shared across both Redis uses**

`_SYNC_REDIS_FAILURE_TIME` enforces 30-second backoff on connection failure. A transient failure on the rate-limiter Redis sets this timestamp, blocking reconnection for main-Redis auth operations for 30 seconds. After the separate-function fix, ensure each client has its own independent backoff state.

---

**`tests/integration/test_rate_limiter_redis_url.py:35` — Global state mutation without teardown**

```python
mcpgateway.auth._SYNC_REDIS_CLIENT = None
```

Direct mutation of module-level global with no `yield`/`finally` teardown. A mid-test failure leaves poisoned global state for subsequent tests. Wrap in a fixture:

```python
@pytest.fixture(autouse=True)
def reset_redis_client():
    original = mcpgateway.auth._SYNC_REDIS_CLIENT
    yield
    mcpgateway.auth._SYNC_REDIS_CLIENT = original
```

---

**`tests/integration/test_rate_limiter_redis_url.py:80–110` — `TestStartupValidation` tests its own inline code, not production code**

The test re-implements the startup check locally instead of calling the actual `_get_sync_redis_client()` or the lifespan function. The production `main.py:1364` code path is never executed. This test can pass while the real startup code diverges completely. Call the actual function or mock at the boundary and assert on real production behavior.

---

**`mcpgateway/middleware/rate_limit_middleware.py` (`_init_redis`) — Double ping on init**

`_init_redis()` calls `auth._get_sync_redis_client()` (which already calls `.ping()` internally) then calls `client.ping()` again — two round-trips on every middleware initialization. Remove the redundant second `client.ping()`.

---

### 🔵 NITS

**`tests/integration/test_rate_limiter_redis_url.py` — Misclassified as integration tests**

All three classes use only `unittest.mock` — no real Redis, no real HTTP, no actual middleware dispatch. These are unit tests. Move to `tests/unit/` or rename. True integration tests should use `fakeredis` or a real Redis container and exercise the full middleware stack.

---

**`tests/unit/mcpgateway/middleware/test_rate_limit_middleware.py` — New tests too weak to catch regressions**

`test_rate_limit_uses_dedicated_redis` patches `_get_sync_redis_client` and asserts `mock_get_client.assert_called()`. This passes with zero behavioral change to the middleware. After the fix, assert that `_get_ratelimiter_redis_client` is called and that `redis.from_url` receives `ratelimiter_redis_url` as its argument.

---

**`tests/unit/charts/test_helm_rendering.py` — Not an actual Helm rendering test**

Pure Python dict simulation, not `helm template`. Cannot catch YAML syntax errors, missing template variables, or Go template logic bugs. Either run `helm template` as a subprocess or label this clearly as a values-mapping unit test.

---

**`mcpgateway/routers/oauth_router.py` — Unrelated change in this PR**

The diff includes adding `from mcpgateway.utils.csp_nonce import get_csp_nonce_from_request` to `oauth_router.py`. Unrelated to rate-limiter Redis separation. Remove from this PR or explain why it belongs here.

---

**`tests/integration/test_rate_limiter_redis_url.py:9` — Unused import**

`import pytest` present but no fixtures, marks, or parametrize decorators used. Remove.

---

**`mcpgateway/auth.py` (inside lock) — `log.getLogger(__name__)` called on every init**

`log.getLogger(__name__)` is called inside the lock body on every initialization attempt instead of being a module-level constant. Use a module-level `logger = logging.getLogger(__name__)` like every other module in this codebase.

---

### Documentation Gaps

- `docs/docs/manage/configuration.md`: new section does not document the interaction between `RATELIMITER_REDIS_URL` and `CACHE_TYPE`. Operators using `CACHE_TYPE=database` who set `RATELIMITER_REDIS_URL` will see silent in-memory fallback with no explanation.
- No documentation for runtime behavior when dedicated Redis becomes unreachable (falls back to in-memory, not to main Redis — this is non-obvious and should be explicit).
- PR description refers to `RATE_LIMITER_REDIS_URL` (underscore between RATE and LIMITER) but the actual env var is `RATELIMITER_REDIS_URL`. Update PR description to avoid operator confusion.

tests/integration/test_rate_limiter_redis_url.py
```diff
@@ -0,0 +1,129 @@
...
+                return mock_dedicated_client
+            return mock_main_client
+
+        mock_redis.from_url.side_effect = from_url_side_effect
+
+        # Reset global client
+        import mcpgateway.auth
+        mcpgateway.auth._SYNC_REDIS_CLIENT = None
+
+        with patch("mcpgateway.auth.config_settings") as mock_settings:
```
madhu-mohan-jaishankar: same comment as above. is config_settings a module level attribute in auth.py? if not, shouldn't be patching this way.

tests/unit/mcpgateway/test_rate_limiter_redis_url.py
```diff
@@ -0,0 +1,129 @@
...
+        mock_client.ping.side_effect = Exception("Connection refused")
+        mock_redis.from_url.return_value = mock_client
+
+        with patch("mcpgateway.main.settings") as mock_settings:
+            mock_settings.ratelimiter_redis_url = "redis://unreachable:6379/0"
+            mock_settings.ratelimiter_redis_socket_connect_timeout = 2.0
+            mock_settings.ratelimiter_redis_socket_timeout = 2.0
+            mock_settings.redis_url = "redis://localhost:6379/0"
+
+            # Simulate startup validation block
```
madhu-mohan-jaishankar: This validation test looks too synthetic. In production, main.py only calls `_get_sync_redis_client()` during startup; the actual Redis from_url() + ping() logic lives in `mcpgateway.auth`. This test instead patches `mcpgateway.main.redis / mcpgateway.main.logger` and manually simulates a try/except block, so it doesn’t exercise the real startup path added by the PR. Could we replace this with either (a) a startup wiring test that asserts startup invokes `_get_sync_redis_client()`, or (b) a direct unit test of `_get_sync_redis_client()` failure behavior? That would give much stronger confidence in the actual behavior.

tests/unit/charts/test_helm_rendering.py
```diff
@@ -0,0 +1,86 @@
+"""Unit tests for Helm chart rendering (Issue #4751).
+
+Tests that rate limiter Redis environment variables are correctly rendered
+in the deployment template when configured in values.yaml.
+"""
```
madhu-mohan-jaishankar: The tests here looks weaker than its name suggests. It doesn’t appear to render the actual Helm chart/template; instead it reconstructs the expected env var list in Python using hand-written if config.get(...) logic. That means it won’t catch real chart regressions like wrong YAML structure, indentation, placement, value path typos, or template mistakes in charts/mcp-stack/templates/deployment-mcpgateway.yaml. Could we switch this to a real rendered-manifest assertion so we’re testing the chart output rather than reimplementing the chart logic in Python?

mcpgateway/auth.py
```diff
@@ -752,15 +752,26 @@ def _get_sync_redis_client():
             # First-Party
             from mcpgateway.utils.redis_client import _build_ssl_kwargs  # pylint: disable=import-outside-toplevel
 
-            if config_settings.redis_url and config_settings.redis_url.startswith("rediss://") and not config_settings.redis_ssl:
+            # Use dedicated rate limiter Redis if configured, otherwise fall back to main Redis
+            redis_url = config_settings.ratelimiter_redis_url or config_settings.redis_url
+            if redis_url.startswith("rediss://") and not config_settings.redis_ssl:
                 log.getLogger(__name__).warning("REDIS_URL uses rediss:// scheme but REDIS_SSL=false — TLS certificate settings will not be applied")
```
madhu-mohan-jaishankar: If `ratelimiter_redis_url `is the active URL, the message still says `REDIS_URL`. minor but it’s confusing operationally.

madhu-mohan-jaishankar (changes requested):

mcpgateway/config.py
```diff
@@ -2571,6 +2571,13 @@ def parse_issuers(cls, v: Any) -> list[str]:
         description="Seconds the circuit remains open before a single probe is allowed. A successful probe closes the circuit; a failed probe extends the cooldown.",
     )
 
+    # Rate Limiter Redis Configuration (Issue #4751)
```
madhu-mohan-jaishankar: issue should not be mentioned in any comment within the code base.

madhu-mohan-jaishankar (changes requested):

cafalchio:
@ja8zyjits The issue is Rate limiter plugin but the changes are for middleware

ja8zyjits:
@cafalchio The idea is to give a different redis url for ratelimiter.

mcpgateway/config.py
ja8zyjits: There are ~171 references of issues in the code base.

mcpgateway/auth.py
ja8zyjits: fixed now.

msureshkumar88:
## Follow-up Review

Good progress on the middleware changes — all the prior blocking items have been addressed. A few things worth discussing before this merges.

---

### The core issue target is not yet met

Issue #4751's acceptance criteria specifically say:

> **When rate limiter _plugin_ initializes → Then connects to separate Redis instance**

The design sketch in the issue also explicitly targets the cpex plugin:

```python
# plugins/rate_limiter/rate_limiter.py
redis_url = settings.RATELIMITER_REDIS_URL or settings.REDIS_URL
self.redis = Redis.from_url(redis_url)
```

The PR correctly adds `RATELIMITER_REDIS_URL` config and wires it into `rate_limit_middleware.py` — but the **cpex rate limiter plugin** (`cpex_rate_limiter.RateLimiterPlugin`) still reads from `REDIS_URL` only. The relevant line is `plugins/config.yaml:287`:

```yaml
# Current — untouched by this PR:
redis_url: '{{ env.REDIS_URL | default("redis://redis:6379/0") }}'
```

An operator setting `RATELIMITER_REDIS_URL` today would get isolation in the built-in middleware but not in the cpex plugin. The fix is a one-liner:

```yaml
redis_url: '{{ env.RATELIMITER_REDIS_URL | default(env.REDIS_URL | default("redis://redis:6379/0")) }}'
```

This is also what @cafalchio flagged — the issue targets the plugin, not the middleware.

---

### Middleware changes look solid

For completeness, here's the status of the middleware implementation:

- ✅ `_get_sync_redis_client()` left unchanged — other callers (e.g. `_update_api_token_last_used_sync`) still route to main Redis correctly
- ✅ `_get_ratelimiter_redis_client()` is a fully isolated path with its own globals, lock, and backoff — no `cache_type` dependency
- ✅ Credentials sanitized before logging via `urlparse._replace()`
- ✅ URL format validated at startup via `@field_validator`
- ✅ Backward compatible — unset `RATELIMITER_REDIS_URL` falls back to main Redis

One minor limitation: `auth.py:841` checks `config_settings.redis_ssl` (main Redis flag) for the rate-limiter TLS warning, and `_build_ssl_kwargs` applies main Redis TLS settings to the dedicated connection. This means operators can't configure independent TLS for the rate-limiter Redis. It's documented as "inherits TLS from main Redis" which is acceptable for now, but worth a note if separate TLS ever becomes a requirement.

---

### Summary

The infrastructure added here is correct and useful. To satisfy the issue's acceptance criteria, `plugins/config.yaml:287` needs the one-line update above so the cpex plugin also honours `RATELIMITER_REDIS_URL`.

msureshkumar88:
### Missing: `plugins/config.yaml` update for the cpex rate limiter plugin

The issue title, epic, and all acceptance criteria specifically target the **rate limiter plugin** (`cpex_rate_limiter.RateLimiterPlugin`):

> *"When rate limiter plugin initializes → Then connects to separate Redis instance"*

The cpex plugin's Redis URL is configured in `plugins/config.yaml:287`, and it was not updated in this PR:

```yaml
# Current — still reads REDIS_URL only:
redis_url: '{{ env.REDIS_URL | default("redis://redis:6379/0") }}'
```

An operator who sets `RATELIMITER_REDIS_URL` today will get isolation in the built-in middleware but the cpex plugin will keep hammering main Redis — which is exactly the contention the issue was filed to fix.

The fix is one line:

```yaml
redis_url: '{{ env.RATELIMITER_REDIS_URL | default(env.REDIS_URL | default("redis://redis:6379/0")) }}'
```

This matches the fallback behaviour the issue describes (use `REDIS_URL` when `RATELIMITER_REDIS_URL` is unset) and requires no changes to the cpex package itself.

Without this, the acceptance criteria in issue #4751 are not satisfied.

---

### Scope: consider removing the middleware changes

The issue is scoped to the **plugin** only — the built-in HTTP middleware (`rate_limit_middleware.py`) is not mentioned anywhere in the issue, epic, or acceptance criteria. The `_get_ratelimiter_redis_client()` function, the `_RATELIMITER_REDIS_CLIENT` global, and the middleware wiring are all out of scope for this issue.

If isolating the middleware's Redis is also desired, that's worth doing — but it should be tracked separately (new issue, separate PR) to keep this PR focused and reviewable. Mixing both changes makes it harder to verify that the original acceptance criteria are met.

A leaner version of this PR would be:
1. Add `RATELIMITER_REDIS_URL` to `config.py` (already done ✅)
2. Update `plugins/config.yaml:287` to use it (one line, not yet done)
3. Update `.env.example`, Helm chart, and docs (already done ✅)

Everything else — `auth.py` globals, `_get_ratelimiter_redis_client()`, middleware changes — could move to a follow-up if the team wants middleware isolation too.

mcpgateway/config.py
madhu-mohan-jaishankar: shouldn't be. that needs to be cleaned up (maybe as a separate cleanup CHORE). not a good practice to reference issues in codebase. usually its the AI that puts in the issue number.

msureshkumar88:
## Deep Re-Review: PR #4859 — Separate Redis for Rate Limiter

Good effort overall. Feature is well-scoped, backward-compatible, and the fallback design is sound. Several issues below need attention before merge — two are blocking.

---

### 🚨 Blocking Issues

**1. Integration test mocks wrong Redis constructor**
`tests/integration/test_rate_limiter_redis_url.py` — `TestRateLimiterWithDedicatedRedis.test_rate_limiter_with_dedicated_redis` patches `mcpgateway.auth.redis.Redis` but the implementation calls `redis.from_url()`. The mock never fires; the test silently passes without exercising the real code path. Fix: patch `redis.from_url` instead.

```python
# Wrong — code uses redis.from_url, not redis.Redis(...)
@patch("mcpgateway.auth.redis")
def test_rate_limiter_with_dedicated_redis(self, mock_redis):
    mock_redis.Redis.return_value = mock_dedicated_client  # never called

# Fix
@patch("mcpgateway.auth.redis.from_url", return_value=mock_dedicated_client)
def test_rate_limiter_with_dedicated_redis(self, mock_from_url):
    ...
```

**2. Redis unavailability logged at DEBUG — operators will miss it**
`mcpgateway/auth.py`, `_get_ratelimiter_redis_client()`, except block:
```python
logger.debug(f"Rate limiter Redis unavailable: {e}")
```
When a **dedicated** rate limiter Redis configured by the operator is unreachable, this is a degraded-service event, not a debug trace. With default log level `ERROR` in production, operators see nothing when their rate limiter Redis silently goes down. This should be `logger.warning(...)`. The SSL misconfiguration path correctly uses `logger.error` — the connection failure path should at minimum use `warning`.

---

### ⚠️ Important Issues

**3. `_get_ratelimiter_redis_client()` called unconditionally at startup even when unconfigured**
`mcpgateway/main.py`:
```python
_get_ratelimiter_redis_client()  # Triggers lazy init + logging
```
When `RATELIMITER_REDIS_URL` is unset (the common case), this call falls through to `_get_sync_redis_client()` — which may already have been initialized elsewhere in startup. It's a redundant, potentially confusing call. Add a guard:
```python
if settings.ratelimiter_redis_url:
    _get_ratelimiter_redis_client()
```

**4. URL validator passes unstripped input — trailing whitespace breaks connections silently**
`mcpgateway/config.py`, `validate_ratelimiter_redis_url`:
```python
if v is not None and v.strip():          # truthy check uses stripped form
    if not (v.startswith("redis://") ...): # but startswith checks original v
return v                                  # returns unstripped value
```
`"redis://localhost:6379 "` (trailing space) passes validation (`startswith` is True) but will fail at connection time. Fix:
```python
v = v.strip()
if not (v.startswith("redis://") or v.startswith("rediss://")):
    raise ValueError(...)
return v
```

**5. No recovery when dedicated Redis drops mid-runtime**
Once `_RATELIMITER_REDIS_CLIENT` is successfully initialized at startup, it stays cached globally. If that Redis instance restarts mid-runtime, subsequent `redis_client.execute_command(...)` calls in the middleware will raise and fall back to in-memory rate limiting — silently, per-process, with no reconnection attempt. The 30-second backoff only protects the *initialization* phase. There is no health-check or reconnect loop once the client is live. This creates a scenario where rate limiting degrades to inconsistent in-memory state in multi-worker deployments without any operator alert. Recommend documenting this limitation prominently and consider resetting `_RATELIMITER_REDIS_CLIENT = None` in the middleware's Redis exception handler to trigger re-initialization on the next request.

**6. `from urllib.parse import urlparse` inside function body on every successful init**
`mcpgateway/auth.py`, inside `_get_ratelimiter_redis_client()`:
```python
from urllib.parse import urlparse  # pylint: disable=import-outside-toplevel
```
This is stdlib — move to top-level imports alongside the existing `from urllib.parse` usage elsewhere in the file. The `# pylint: disable` comment is a signal this import belongs outside the function.

**7. `tests/integration/` tests are unit tests in disguise**
`tests/integration/test_rate_limiter_redis_url.py` — every test uses `MagicMock` and `patch` with no live Redis connection. These belong in `tests/unit/`. True integration tests for this feature would start a real Redis instance (or use the existing Docker Compose stack in CI) and verify actual rate limit counters are stored and enforced. The `TestStartupValidation` class is especially problematic: it manually re-implements the startup validation logic rather than calling the actual startup path.

**8. PR description names the wrong environment variable**
The PR description says `RATE_LIMITER_REDIS_URL` throughout, but the actual environment variable is `RATELIMITER_REDIS_URL` (no separator before REDIS). This will cause confusion for operators reading the description.

---

### Hardening

**9. Connection failure swallowed at DEBUG masks Hardening-relevant degradation**
(Covered in item 2, but worth calling out separately.) When rate limiting falls back to in-memory due to Redis failure, rate limit state is no longer shared across workers. In a multi-worker/multi-pod deployment, each process enforces its own window independently — a client can multiply their effective rate limit by the number of workers. This is a Hardening degradation that must be visible in logs at `warning` or above, not `debug`.

**10. No dedicated TLS configuration for rate limiter Redis**
The `_build_ssl_kwargs(settings)` call uses main Redis SSL settings (`REDIS_SSL`, `REDIS_SSL_CA_CERTS`, `REDIS_SSL_CHECK_HOSTNAME`, etc.). There is no way to configure different TLS parameters for the dedicated rate limiter Redis. This is documented in `.env.example` ("Inherits TLS settings from main Redis") but operators deploying the rate limiter Redis on a different host with a different CA or cert will have no recourse. Consider this a known limitation and document it explicitly in `docs/docs/manage/configuration.md`.

**11. Redis URL credentials not scrubbed at config validation time**
The URL is stored verbatim in `settings.ratelimiter_redis_url`. If any component calls `str(settings)` or dumps settings for diagnostics, credentials embedded in the URL (`redis://:password@host/0`) will be exposed. The per-log sanitization in `_get_ratelimiter_redis_client` is good, but the settings object itself is unprotected. The main `redis_url` field has the same gap, but new fields should not repeat it.

---

### 🔁 Refactoring

**12. Duplicated init pattern between `_get_sync_redis_client` and `_get_ratelimiter_redis_client`**
Both functions share: global + lock + double-checked locking + backoff + `redis.from_url` + SSL kwargs + ping + failure tracking. The duplicated structure (~80 lines) is a future maintenance burden. Worth extracting to a private helper:
```python
def _init_redis_client_with_backoff(
    url: str, client_global: str, lock: threading.Lock, failure_time_global: str, **kwargs
) -> Optional[redis.Redis]: ...
```
Not a blocker for this PR, but a follow-up ticket would keep this maintainable.

---

### 🧪 Test Coverage Gaps

**13. Missing: `ValueError` path (SSL misconfiguration) in `_get_ratelimiter_redis_client`**
`test_auth_ratelimiter_redis.py` has no test verifying that a `ValueError` from `_build_ssl_kwargs` is caught, logged at `error`, and returns `None` without setting the backoff timer.

**14. Missing: Credential sanitization in success log**
No test verifies that `redis://:secret@host/0` is logged as `redis://***@host/0`. This is a Hardening-relevant behavior.

**15. Missing: Actual rate limit enforcement test**
No test (unit, integration, or e2e) verifies that requests are *actually blocked* after N requests when a dedicated Redis is configured. All tests verify client initialization, not rate limiting behavior.

**16. Missing: 30-second backoff expiry and retry**
No test verifies that after the 30-second backoff window expires, `_get_ratelimiter_redis_client` retries the connection.

**17. Dead import in test file**
`tests/unit/mcpgateway/test_auth_ratelimiter_redis.py`:
```python
from mcpgateway.config import Settings  # imported but never used
```

**18. Helm chart test doesn't render actual Helm templates**
`tests/unit/charts/test_helm_rendering.py` constructs a Python dict and runs assertions against it — it does not call `helm template` or validate actual YAML rendering. If the Helm template has a bug in the conditional block, this test will not catch it. Recommend using `subprocess.run(["helm", "template", ...])` or marking it clearly as a logic-only sanity check.

---

### 📖 Documentation Gaps

**19. Multi-worker fallback behavior not documented**
`docs/docs/manage/configuration.md` should note: when `RATELIMITER_REDIS_URL` is set but Redis becomes unavailable mid-runtime, each worker falls back to independent in-memory rate limiting. Rate limits are no longer enforced globally. This is an operational risk operators need to understand.

**20. Total connection pool budget not documented**
With `RATELIMITER_REDIS_MAX_CONNECTIONS=50` (default) added on top of the main Redis pool, operators need to plan Redis `maxclients` accordingly. The docs mention the main pool but not the additive effect of the rate limiter pool.

**21. `configuration.md` section placement**
The new "Rate Limiter Redis" section is inserted between the tip block and the Redis SSL section, which breaks the visual flow. It should be a subsection under the main Redis configuration heading rather than floating between two unrelated blocks.

---

### ✅ Not a Breaking Change

Confirmed backward-compatible: `RATELIMITER_REDIS_URL` defaults to `None`, existing deployments use `REDIS_URL` unchanged. The `_log_auth_event` parameter rename (`logger` → `logger_handle`) is internal only and not in `__all__`.

The `plugins/config.yaml` change correctly chains `RATELIMITER_REDIS_URL | default(REDIS_URL)` — plugin behavior is unchanged when the new variable is unset.

---

**Summary:** Items 1–2 are blocking. Items 3–8 should be addressed before merge. Hardening items 9–11 are important for production deployments. Test gaps 13–18 should be filed as follow-up tickets if not fixed in this PR.

msureshkumar88 (changes requested):
Thanks for the continued iteration on this — good progress across several fronts.

## What's been fixed ✅

- **Mock target corrected**: `test_rate_limiter_redis_url.py` now patches `redis.from_url` instead of the `Redis` constructor — correct approach
- **Log level fixed**: Redis unavailability now logs at `WARNING` (`auth.py:851`) instead of `DEBUG`
- **Startup guard added**: `_get_ratelimiter_redis_client()` now only called if `settings.ratelimiter_redis_url` is set
- **URL validation strips whitespace**: `v.strip()` applied before scheme check in validator
- **Credential sanitization in logs**: `auth.py:844` uses `safe_url` (strips password from URL before logging)
- **Log message fixed**: No longer says `REDIS_URL` when dedicated ratelimiter URL is active
- **Fallback behavior documented**: Degraded per-worker in-memory mode now explicitly described in README/docs
- **Integration test with bad patches removed**: `tests/integration/test_rate_limiter_redis_url.py` deleted

---

## What still needs fixing

### Blocking

**1. `test_startup_warns_on_unreachable_ratelimiter_redis` is still synthetic** (`tests/unit/mcpgateway/test_rate_limiter_redis_url.py:73–105`)

This test patches `mcpgateway.main.redis` and `mcpgateway.main.logger`, then manually writes a try/except that calls `mock_redis.from_url()` and asserts on a warning message it constructed itself. It never calls `_get_ratelimiter_redis_client()`. This was flagged in the last review round and is unchanged.

The test will always pass regardless of implementation changes — it's testing the test author's simulation, not the code. Fix: patch `mcpgateway.auth.redis` and call `_get_ratelimiter_redis_client()` directly with a `ping()` that raises, then assert on the actual warning emitted.

**2. `test_helm_rendering.py` still not a Helm test** (`tests/unit/charts/test_helm_rendering.py`)

Also flagged last round and unchanged. The test reconstructs expected env vars using hand-written Python `if config.get(...)` logic — it doesn't invoke `helm template`. It won't catch real chart regressions (wrong YAML structure, incorrect conditionals in the Go template, missing indentation).

Two options:
- Replace with a test that runs `helm template` and parses the output with `yq`
- Delete the file rather than carry misleading coverage

### Important

**3. Issue references still present in source code**

Multiple files still contain `(Issue #4751)` in comments and docstrings — `config.py:2596`, `auth.py:124`, `main.py:1358`, `.env.example`, `test_helm_rendering.py:1`, `test_rate_limiter_redis_url.py:1`. This was explicitly flagged by the reviewer. Please remove them.

### Minor

**4.** `test_rate_limiter_redis_url.py:1` docstring says `"Integration tests"` but the file lives in `tests/unit/`. Update the docstring to match.

**5.** PR description still says `RATE_LIMITER_REDIS_URL` (with space between RATE and LIMITER); actual env var is `RATELIMITER_REDIS_URL`. Worth correcting to avoid operator confusion.

docs/docs/manage/configuration.md
```diff
@@ -968,6 +968,59 @@ The gateway includes built-in observability features for tracking HTTP requests,
...
+| Variable | Default | Description |
+|----------|---------|-------------|
+| `RATELIMITER_REDIS_URL` | `None` | Optional Redis URL for rate limiting. Falls back to `REDIS_URL` when unset. Must start with `redis://` or `rediss://`. |
+| `RATELIMITER_REDIS_MAX_CONNECTIONS` | `50` | Connection pool size for rate limiter Redis. |
+| `RATELIMITER_REDIS_SOCKET_TIMEOUT` | `2.0` | Socket timeout in seconds. |
+| `RATELIMITER_REDIS_SOCKET_CONNECT_TIMEOUT` | `2.0` | Connection timeout in seconds. |
+
+**Migration:** Existing deployments continue using main Redis. Set `RATELIMITER_REDIS_URL` to enable dedicated instance.
+
+!!! note "Rate Limiter Redis Behavior"
```
gandhipratik203, line 989: This might be worth confirming with wxo, if they want this behaviour or they want seperate tls certs for this redis instance.

docs/docs/manage/configuration.md
```diff
@@ -968,6 +968,59 @@ The gateway includes built-in observability features for tracking HTTP requests,
...
+    - **TLS Configuration:** Rate limiter Redis inherits TLS settings from main Redis (`REDIS_SSL`, `REDIS_CA_CERT`, etc.). Use `rediss://` scheme for TLS connections.
+    - **Independent of CACHE_TYPE:** Rate limiter Redis operates independently of the `CACHE_TYPE` setting. It does not require `CACHE_TYPE=redis` to function.
+    - **Fallback:** When `RATELIMITER_REDIS_URL` is unset, rate limiting uses the main Redis instance via `REDIS_URL` (backward compatible).
+
+#### Rate Limiter Redis Fallback Behavior
+
+When `RATELIMITER_REDIS_URL` is configured but the dedicated Redis instance becomes unavailable mid-runtime:
```
gandhipratik203, line 995: Should the behaviour be something like, we fall back to the main redis instance if the rate limiter redis url isn't avialable and last option, fallback to in memory?

msureshkumar88 (changes requested):
Thanks for the continued iteration on this — good progress across several fronts.

## What's been fixed ✅

- **Mock target corrected**: patches `redis.from_url` instead of the `Redis` constructor — correct approach
- **Log level fixed**: Redis unavailability now logs at `WARNING` (`auth.py:851`) instead of `DEBUG`
- **Startup guard added**: `_get_ratelimiter_redis_client()` only called if `settings.ratelimiter_redis_url` is set
- **URL validation strips whitespace**: `v.strip()` applied before scheme check
- **Credential sanitization in logs**: `auth.py:844` uses `safe_url` (strips password before logging)
- **Log message fixed**: no longer says `REDIS_URL` when dedicated ratelimiter URL is active
- **Fallback behavior documented**: degraded per-worker in-memory mode explicitly described in README/docs
- **Integration test with bad patches removed**: `tests/integration/test_rate_limiter_redis_url.py` deleted

---

## What still needs fixing

### Blocking

**1. `test_startup_warns_on_unreachable_ratelimiter_redis` is still synthetic** (`tests/unit/mcpgateway/test_rate_limiter_redis_url.py:73–105`)

This test patches `mcpgateway.main.redis` and `mcpgateway.main.logger`, then manually writes a try/except that calls `mock_redis.from_url()` and asserts on a warning message it constructed itself. It never calls `_get_ratelimiter_redis_client()`. This was flagged in the last review round and is unchanged.

The test will always pass regardless of implementation changes — it's testing the test author's simulation, not production code. Fix: patch `mcpgateway.auth.redis` and call `_get_ratelimiter_redis_client()` directly with a `ping()` that raises, then assert on the actual warning emitted by `auth.py:851`.

**2. `test_helm_rendering.py` still not a Helm test** (`tests/unit/charts/test_helm_rendering.py`)

Also flagged last round and unchanged. The test reconstructs expected env vars using hand-written Python `if config.get(...)` logic — it never invokes `helm template`. Won't catch real chart regressions (wrong YAML structure, incorrect Go template conditionals, indentation bugs).

Two options: replace with a test that runs `helm template` and parses output with `yq`, or delete the file rather than carry misleading coverage.

### Important

**3. Issue references still in source code**

`config.py:2596`, `auth.py:124`, `main.py:1358`, `.env.example`, `test_helm_rendering.py:1`, `test_rate_limiter_redis_url.py:1` all still contain `(Issue #4751)` in comments/docstrings. Please remove them.

### Minor

**4.** `test_rate_limiter_redis_url.py:1` docstring says `"Integration tests"` but the file lives in `tests/unit/`. Update to match.

**5.** PR description still says `RATE_LIMITER_REDIS_URL` (space between RATE and LIMITER); actual env var is `RATELIMITER_REDIS_URL`. Worth correcting to avoid operator confusion.

mcpgateway/config.py
ja8zyjits: https://github.com/IBM/mcp-context-forge/pull/4656/changes#diff-a1c5af7f74f50dc8f5d9fa29703fd52127b66d88b21c67202ef1cf2942778c78R895

ja8zyjits: https://github.com/IBM/mcp-context-forge/pull/4656/changes#diff-a1c5af7f74f50dc8f5d9fa29703fd52127b66d88b21c67202ef1cf2942778c78R1524

docs/docs/manage/configuration.md
ja8zyjits: This is tough, since we will have to maintain multiple clients in the ratelimiter middleware.
Also the whole mechanism of fallback is weird, since we dont do it properly.
Like we need to proxy all the redis calls in the RateLimitMiddleware to fall back multiple clients and then finally use the in-memory.
So as things stand in this PR, during run time if the ratelimiter redis fails we fall back to in-memory and in a follow up we handle this better.

gandhipratik203: That makes sense Jitesh. Thanks for looking into this. This PR is already adding good value. We can leave this as it is or if suits just create an issue,that can be taken up later. Up to you. Happy to resolve this.

docs/docs/manage/configuration.md
gandhipratik203: Thanks for the changes, Jitesh, and thanks for taking the positive route and looking into this further.

`build_reatelimiter_ssl_kwargs` in `mcpgateway/utils/redis_client.py` gates correctly on the new `settings.ratelimiter_redis_ssl` flag, but the cert paths inside it are still read from the main Redis SSL fields (`settings.redis_ssl_ca_certs`, `settings.redis_ssl_certfile`, `settings.redis_ssl_keyfile`, `settings.redis_ssl_check_hostname`). So the new `ratelimiter_redis_ssl_ca_certs` / `ratelimiter_redis_ssl_certfile` / `ratelimiter_redis_ssl_keyfile` / `ratelimiter_redis_ssl_check_hostname` fields that were added to `config.py` get parsed and stored, but never actually applied. An operator setting `RATELIMITER_REDIS_SSL_CA_CERTS=/certs/dedicated-ca.crt` would not see it take effect, and there would be no error to surface that.

The swap inside the helper is fairly small:

```python
def build_reatelimiter_ssl_kwargs(settings: Any) -> dict[str, Any]:
    if not settings.ratelimiter_redis_ssl:
        return {}

    _validate_ssl_settings(settings)

    kwargs: dict[str, Any] = {}

    if settings.ratelimiter_redis_ssl_ca_certs:
        kwargs["ssl_ca_certs"] = settings.ratelimiter_redis_ssl_ca_certs
    if settings.ratelimiter_redis_ssl_certfile:
        kwargs["ssl_certfile"] = settings.ratelimiter_redis_ssl_certfile
    if settings.ratelimiter_redis_ssl_keyfile:
        kwargs["ssl_keyfile"] = settings.ratelimiter_redis_ssl_keyfile

    if not settings.ratelimiter_redis_ssl_check_hostname:
        kwargs["ssl_cert_reqs"] = "none"
        kwargs["ssl_check_hostname"] = False

    return kwargs
```

One thing to be aware of either way: `_validate_ssl_settings(settings)` itself reads the main `redis_ssl_*` fields, so for fully independent validation it would need a sibling that reads the `ratelimiter_redis_ssl_*` fields. Happy to take that part as a follow-up if you'd rather not expand scope here.

msureshkumar88 (changes requested):
Good progress across multiple rounds — the singleton global routing bug, credential sanitization, startup guard, URL validation, and log message fixes are all solid improvements. Two blocking items remain before this can merge, plus a documentation gap surfaced by the interaction with PR #4848.

---

## Blocking

**1. `test_startup_warns_on_unreachable` is too synthetic**

The test patches `mcpgateway.main.redis` and manually simulates a try/except block. In production, `main.py` only calls `_get_sync_redis_client()` during startup — the actual `redis.from_url()` + `ping()` logic lives in `mcpgateway.auth`, not `main`. The test doesn't exercise the real startup path added by this PR, so it provides false confidence.

Replace with either:
- (a) A startup wiring test that asserts startup invokes `_get_ratelimiter_redis_client()` when `settings.ratelimiter_redis_url` is set, or
- (b) A direct unit test of `_get_ratelimiter_redis_client()` failure behaviour in `mcpgateway.auth` — patch `redis.from_url` to raise, assert the warning is logged and `None` is returned

**2. Helm rendering test is synthetic**

The Helm test reimplements chart logic in Python rather than asserting against actual rendered manifests. Replace with a rendered-manifest assertion (`helm template` output) so the test catches real chart regressions rather than re-verifying Python logic.

---

## Cross-impact with PR #4848 — documentation gap

PR #4848 (now approved) documents the plugin-binding rate limiter. After both PRs land, operators face two distinct Redis URL configurations for rate limiting:

| Layer | Mechanism | Redis URL source |
|---|---|---|
| Plugin-binding rate limiting | `cpex_rate_limiter` Rust plugin, per-tool/per-tenant via `POST /v1/tools/plugin_bindings` | `redis_url` in `plugins/config.yaml` |
| Middleware-level rate limiting | `RateLimitMiddleware` in `auth.py`, gateway-wide | `RATELIMITER_REDIS_URL` env var (this PR) |

Neither PR's docs explain the two-layer architecture. An operator who read PR #4848 and then encounters `RATELIMITER_REDIS_URL` will not understand whether they're configuring the same thing or two separate systems.

Add a brief note in the `RATELIMITER_REDIS_URL` section (`.env.example` or `docs/docs/manage/configuration.md`) clarifying: *"This configures the gateway-wide middleware rate limiter. The per-binding plugin rate limiter (`cpex_rate_limiter`) uses a separate `redis_url` key in `plugins/config.yaml`."*

---

## Worth addressing (not blocking)

**Fallback chain not documented explicitly:** Current behaviour when `RATELIMITER_REDIS_URL` is unreachable is fall back to per-worker in-memory — it does NOT fall back to the main `REDIS_URL`. This is a non-obvious operational behaviour. Add one sentence to the docs: *"If the dedicated rate-limiter Redis is unreachable, the middleware falls back to per-worker in-memory counters — it does not fall back to the main `REDIS_URL`."*

**Issue reference in `mcpgateway/config.py`:** Issue numbers in code comments are not a good practice and should be removed (separate cleanup is fine).

ja8zyjits:
Hi @msureshkumar88 

* Both the blockers were removed in the previous commits.
* Both middleware and the plugins are going to use RATELIMITER_REDIS_URL as mentioned in https://github.com/IBM/mcp-context-forge/pull/4859/changes#diff-bccbe51faeffaaaf5c52cc297056f9070c6970df0205829f7a0a5931fc0acc72R292
* "Fallback chain not documented explicitly" is mentioned explicitly in the https://github.com/IBM/mcp-context-forge/pull/4859/changes#diff-ca6a19fa3389cc24f3cb6311d36f544fe8812c9b4e221f1380c73a90f439e55aR991 and https://github.com/IBM/mcp-context-forge/pull/4859/changes#diff-ca6a19fa3389cc24f3cb6311d36f544fe8812c9b4e221f1380c73a90f439e55aR995
* PR 4848 will update according to this PR, so no issues.
* there are no "issues" tagged in the PR anymore.
All your issues are resolved.

msureshkumar88 (dismissed):
All items from the last review round are resolved.

**Blocking 1 — Startup test:** Replaced with  and , both calling  directly. No longer synthetic. ✅

**Blocking 2 — Helm test:**  deleted. ✅

**Doc gap:** `configuration.md` now explicitly documents: CACHE_TYPE independence, fallback to main Redis, and the in-memory per-worker degradation path with an operational warning. ✅

**Minor (non-blocking):** `@patch("redis.from_url")` patches at the definition site rather than consumption site (`mcpgateway.auth.redis.from_url`). Works correctly when `auth.py` does `import redis`, but patching where it is *used* is more idiomatic and resilient to future import changes. Worth a follow-up but not a reason to hold the PR.

Apologies for flagging issues that were already addressed — the commits pre-dated my review and I missed them. This is merge-ready.

ja8zyjits:
@msureshkumar88 
please review again. I have fixed the redis mocking also.
Along with that I have fixed @gandhipratik203 comments too.

gandhipratik203 (approved):
LGTM, Thanks for the contribution Jitesh!.

PR #4862, opened 2026-05-22
[FIX]: Increase health check timeout from 5s to 30s and reset failure counter on recovery

### LINKED ISSUE #4785: [BUG]: UI shows mcp server "offline" while Test succeeds.
### 🐞 Bug Summary
UI "MCP Servers & Federated Gateways (MCP Registry)" shows mcp server status "Offline" with flyover "Gateway not reachable" while Test feature of same page for same mcp server succeeds.
---

### 🧩 Affected Component

- [ ] `mcpgateway` - API
- [ x] `mcpgateway` - UI (admin panel)
- [ ] Other (explain below)

---

### 🔁 Steps to Reproduce

1. Install ibmi-mcp-server
2. UI shows offline
3. Click the Test action

---

### 🤔 Expected Behavior
When server is active, UI should not show offline

---

### 📓 Logs / Error Output

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Version or commit | v0.9.0 |
| Runtime | Python 3.12.3 |
| Platform / OS | Ubuntu 24.04.4 LTS |
| Container | none |

---

Image

Image

MohanLaksh:
## Problem

Users report that MCP servers show as "Offline" in the UI with tooltip "Gateway not reachable", but the Test button succeeds for the same server. This creates confusion as the gateway is clearly reachable. The issue manifests after approximately 3-5 minutes of operation, suggesting a timeout-related problem.

**Issue:** #4785

## Root Cause

The problem is a **24x timeout mismatch** between periodic health checks and the Test button:

| Aspect | Periodic Health Check | Test Button |
|--------|----------------------|-------------|
| Code | `_check_single_gateway_health()` (`gateway_service.py:3688`) | `admin_test_gateway()` (`admin.py:14002`) |
| Timeout | `health_check_timeout` = **5s** (`config.py:2409`) | `federation_timeout` = **120s** (`config.py:2172`) |
| Wrapper timeout | `gateway_health_check_timeout` = **5.0s** (`config.py:2412`) | N/A |
| Failure behavior | After **3 consecutive failures** (~3 min), sets `reachable=False` | Returns result, does NOT modify DB state |
| Recovery | Next successful health check sets `reachable=True` (up to 60s later) | N/A |

**The failure sequence:**
1. Gateway is slow to respond (e.g., 6-10s due to auth, cold start, or network latency)
2. Health check times out at 5s → failure count increments
3. After 3 consecutive failures (~3 minutes), `reachable` is set to `False`
4. UI shows yellow "Offline" badge — but gateway is actually healthy
5. Test button uses 120s timeout and succeeds every time

### 1. Increase Health Check Timeout from 5s → 30s

`config.py`:
- `health_check_timeout`: `5` → `30`
- `gateway_health_check_timeout`: `5.0` → `30.0`

**Why 30s and not 120s (to match Test button)?**

- **30s aligns with existing `httpx_admin_read_timeout`** — the httpx client at `gateway_service.py:3807` was already configured with this value as its default timeout. The 5s override at line 3876 was contradicting the intended design.
- **30s is a 6x improvement** over 5s, covering the vast majority of slow-responder scenarios (cold starts, auth token exchange, TLS negotiation, moderate network latency).
- **Failure detection remains reasonable** — at 30s + 60s interval + threshold 3, a truly dead gateway is detected in ~4.5 min vs ~3 min at 5s. This ~90s tradeoff is acceptable for ops response.
- **120s detection time** would mean a dead gateway takes up to 6+ minutes to detect, which is too slow for practical operations.

**Resource impact analysis:**
- Max 10 concurrent health checks (semaphore-limited)
- At 30s each: 300 connection-seconds per cycle
- With 50 gateways at 30s worst case: ~150s cycle time vs 60s interval → at most 2-3 overlapping cycles → 20-30 concurrent HTTP connections
- DB pool is 200+10 connections — **negligible impact**

### 2. Reset Failure Counter on Gateway Recovery

`gateway_service.py:3900` — Added:
```python
self._gateway_failure_counts[gateway_id] = 0
```

**Why?** Before this fix, a flickering gateway that briefly recovers then has a single slow response would retain its pre-recovery failure count (e.g., 2). On the next failure, it would immediately hit the threshold of 3 and be marked offline. Resetting the counter on recovery ensures a clean slate.

### 3. Documentation Updates

- `docs/docs/manage/configuration.md` — Updated documented defaults
- `docs/docs/manage/scale.md` — Updated example config values
- `docs/docs/manage/tuning.md` — Updated health check timeout table
- `.env.example` — Updated commented-out examples

## Verification

- **`make test`**: 18735 passed, 591 skipped (pre-existing)
- **`make ruff`**: All checks passed
- **`make bandit`**: 0 issues
- **`make pylint`**: 10.00/10
- **`make interrogate`**: 100.0% docstring coverage
- **`make pre-commit`**: All 20 hooks passed

No test regressions — all health check tests explicitly mock their own timeout values and are unaffected by the default change.

Closes #4785

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -588,7 +588,7 @@ DATAPLANE_PUBLISHER_INTERVAL_SECONDS = 60
 
 # Health checks
 # HEALTH_CHECK_INTERVAL=60
-# HEALTH_CHECK_TIMEOUT=5
+# HEALTH_CHECK_TIMEOUT=30
 # UNHEALTHY_THRESHOLD=3
 # GATEWAY_VALIDATION_TIMEOUT=5
 # MAX_CONCURRENT_HEALTH_CHECKS=10
@@ -2479,11 +2479,11 @@ VALIDATE_TOKEN_ENVIRONMENT=true
 # Project defaults block sets HEALTH_CHECK_INTERVAL=300 for local dev
 # HEALTH_CHECK_INTERVAL=60
 
-# Health check timeout in seconds (default: 5)
-# HEALTH_CHECK_TIMEOUT=5
+# Health check timeout in seconds (default: 30)
+# HEALTH_CHECK_TIMEOUT=30
 
-# Per-check timeout (seconds) to bound total time of one gateway health check (default: 5.0)
-# GATEWAY_HEALTH_CHECK_TIMEOUT=5.0
+# Per-check timeout (seconds) to bound total time of one gateway health check (default: 30.0)
+# GATEWAY_HEALTH_CHECK_TIMEOUT=30.0
 
 # Consecutive failures before marking gateway offline (default: 3)
 # UNHEALTHY_THRESHOLD=3
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-07-22T10:27:33Z",
+  "generated_at": "2026-07-23T12:29:35Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -266,7 +266,7 @@
         "hashed_secret": "b4673e578b9b30fe8bba1b555b7b59883444c697",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1889,
+        "line_number": 1902,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -274,7 +274,7 @@
         "hashed_secret": "4a0a2df96d4c9a13a282268cab33ac4b8cbb2c72",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1977,
+        "line_number": 1990,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -282,7 +282,7 @@
         "hashed_secret": "5baa61e4c9b93f3f0682250b6cf8331b7ee68fd8",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 2327,
+        "line_number": 2340,
         "type": "Basic Auth Credentials",
         "verified_result": null
       },
@@ -290,7 +290,7 @@
         "hashed_secret": "fa9beb99e4029ad5a6615399e7bbae21356086b3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 3692,
+        "line_number": 3705,
         "type": "Basic Auth Credentials",
         "verified_result": null
       },
@@ -298,7 +298,7 @@
         "hashed_secret": "fa9beb99e4029ad5a6615399e7bbae21356086b3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 3783,
+        "line_number": 3796,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -306,7 +306,7 @@
         "hashed_secret": "ac371b6dcce28a86c90d12bc57d946a800eebf17",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 3826,
+        "line_number": 3839,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -314,7 +314,7 @@
         "hashed_secret": "0b6ec68df700dec4dcd64babd0eda1edccddace1",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 3831,
+        "line_number": 3844,
         "type": "Secret Keyword",
         "verified_result": null
       },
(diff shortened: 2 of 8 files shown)


lucarlig (changes requested):
@MohanLaksh there are conflicts to fix please.
only one issue i found:
 - Failure counter reset is too narrow. It only resets when an already-offline gateway recovers, but not after a successful check while still `reachable=True`. A gateway can fail twice, pass once, then fail once and still be marked offline. Reset the counter on any successful health check and add a regression test.

lucarlig (dismissed):
LGTM

ja8zyjits:
## ⚠️ Warnings (Non Blocking)

- [ ] **Stale comment in tuning.md:468**
  - States "default 5s" but actual default now 30s
  - Fix: Update to "(default 30s)"

- [ ] **Contradictory comment in tuning.md:478-479**
  - Says "5s timeout" but example shows `HEALTH_CHECK_TIMEOUT=30`
  - Fix: Update to "(30s timeout for explicit RPC)"

lucarlig (dismissed):

ja8zyjits (dismissed):
LGTM

msureshkumar88 (dismissed):
## E2E verification of the fix (issue #4785)

Ran this PR's branch locally against a purpose-built end-to-end script that drives the *real* `GatewayService._check_single_gateway_health()` code path — no mocked HTTP client, no mocked settings, real SQLite DB — to reproduce the reported symptom and confirm the fix.

### Setup / config
- Checked out branch `fix/issue-4785-health-check-timeout` at the PR's head commit.
- Fresh throwaway SQLite DB (`tempfile`), real `mcpgateway.db.Gateway` row inserted and read back through `SessionLocal`.
- A local `http.server` instance standing in for a "slow but healthy" MCP gateway: it sleeps **8s** before returning `200 OK` — chosen because it sits between the old 5s timeout (would fail) and the new 30s timeout (should pass), matching the "6-10s slow responder" scenario described in the issue.
- `settings.health_check_timeout` pinned explicitly to `5` then `30` per step so the assertions don't depend on ambient `.env`/CWD resolution.

### What the script checks
1. `config.py` field defaults are `30` / `30.0` (not the pre-fix `5` / `5.0`).
2. **Regression proof**: with the pre-fix timeout (5s), the 8s-slow-but-healthy gateway *is* incorrectly marked as a failure (reproduces the bug).
3. **Fix proof**: with the PR's timeout (30s), the same 8s-slow gateway is *not* marked as a failure, and the health check genuinely waits past the old 5s cutoff (not a fluke).
4. DB row shows `reachable=True` and `last_seen` updated after a successful check.
5. **Failure-counter reset**: a gateway with a stale failure count (simulated flicker: count=2) has its counter reset to `0` immediately on a successful check.
6. The first failure *after* that recovery increments to `1`, not `3` — confirms the old "stale count survives recovery" flicker bug is gone.
7. Gateway stays `reachable=True` after that single post-recovery failure (threshold=3 not reached).
8. Threshold/deactivation logic still works end-to-end: 3 consecutive real failures still flip `reachable=False`.

### Commands run
```bash
PYTHONPATH=$(pwd) .venv/bin/python e2e_health_check.py
.venv/bin/python -m pytest tests/unit/mcpgateway/services/test_gateway_service.py -k "health_check" -q
.venv/bin/python -m ruff check mcpgateway/config.py mcpgateway/services/gateway_service.py
```

### Output
```
config.py field defaults: health_check_timeout=30s, gateway_health_check_timeout=30.0s
[PASS] config.py field defaults are 30 / 30.0, not the pre-fix 5 / 5.0
[PASS] Pre-fix timeout (5s) reproduces the bug: 8s-slow healthy gateway is marked as a failure -- elapsed=5.01s, failure_count=1
[PASS] With PR fix (30s timeout), 8s-slow healthy gateway is NOT marked as a failure -- elapsed=8.01s, failure_count=0
[PASS] Health check actually waited past the old 5s cutoff -- elapsed=8.01s
[PASS] DB row: reachable stays True, last_seen updated after successful check
[PASS] Failure counter reset to 0 immediately after a successful health check
[PASS] First failure after recovery increments to 1, not 3 (no stale-count flicker bug)
[PASS] Gateway NOT deactivated after a single post-recovery failure (threshold=3 not reached)
[PASS] 3 consecutive real failures still deactivate the gateway (threshold logic intact)

RESULT: 9 passed, 0 failed
```
`pytest -k health_check`: **9 passed** (includes the new `test_successful_health_check_resets_failure_counter`).
`ruff check` on `config.py` and `gateway_service.py`: **All checks passed!**

### Conclusion
Both root-cause fixes are verified against real, unmocked behavior:
- The 5s→30s timeout increase genuinely stops a slow-but-healthy gateway (6-10s response) from being misreported as offline.
- The failure-counter reset on recovery genuinely prevents the "flicker" bug where a stale count would prematurely trip the unhealthy threshold after a single post-recovery failure.

Approving.

msureshkumar88:
@ja8zyjits 
This PR currently has merge conflicts with `main` — could you rebase onto the latest `main` and push the update? Thanks!

msureshkumar88 (approved):
All CI checks passing (30/30), no merge conflicts, mergeable.

Verified the previously requested change is addressed: the failure counter now resets on every successful health check (not just recovery-from-offline), with a new regression test `test_successful_health_check_resets_failure_counter` covering the case where a gateway stays `reachable=True` across checks.

Approving.

PR #4872, opened 2026-05-22
fix(bootstrap): align env var name, fix advisory lock release, pass connect_args

Lang-Akshay:
## Summary

- Renames `SKIP_MIGRATION` → `MCPGATEWAY_SKIP_MIGRATIONS` across `.env.example`, `docker-compose.yml`, and Helm chart comment block to match what `config.py` and `bootstrap_db.py` actually read
- Fixes advisory lock release in `bootstrap_db.py` — wraps `pg_advisory_unlock` in try/except so a dropped connection during teardown doesn't swallow the root exception
- Passes `connect_args` (imported from `mcpgateway.db`) to `create_engine` in `main()`, matching the rest of the codebase and ensuring SQLite WAL/timeout settings are applied
- Tightens error logging in the skip-migrations fast path: `RuntimeError` (already logged with context) is no longer double-logged; only unexpected failures hit the generic `logger.error`
- Improves actionability of the "schema not at head" error message — tells the operator exactly what to do
- Adds `logger.info("Database ready")` to the already-at-head fast path so startup logs confirm readiness
- Updates `make_alembic_cfg` docstring to reference the correct caller (`mcpgateway.utils.check_schema_at_head`)

### DIFF
--- a/mcpgateway/bootstrap_db.py
+++ b/mcpgateway/bootstrap_db.py
@@ -52,12 +52,17 @@ from sqlalchemy.orm import Session
 # First-Party
 from mcpgateway.common.validators import SecurityValidator
 from mcpgateway.config import settings
-from mcpgateway.db import A2AAgent, EmailTeam, EmailUser, Gateway, Prompt, Resource, Server, Tool, connect_args
+from mcpgateway.db import A2AAgent, connect_args, EmailTeam, EmailUser, Gateway, Prompt, Resource, Server, Tool
 from mcpgateway.services.logging_service import LoggingService
 
 # Migration lock to prevent concurrent migrations from multiple workers
 _MIGRATION_LOCK_PATH = os.path.join(tempfile.gettempdir(), "mcpgateway_migration.lock")
 _MIGRATION_LOCK_TIMEOUT = 300  # seconds to wait for lock (5 minutes for slow migrations)
+
+
+class _SchemaNotAtHeadError(RuntimeError):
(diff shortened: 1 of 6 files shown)


mcpgateway/bootstrap_db.py
```diff
@@ -251,7 +235,10 @@ def advisory_lock(conn: Connection):
             yield
         finally:
             logger.info("Releasing Postgres advisory lock...")
-            conn.execute(text(f"SELECT pg_advisory_unlock({pg_lock_id})"))
+            try:
```
cafalchio, line 243: thanks for the try catch. 
would the lock expire if the owner does not remove it?

cafalchio (dismissed):
Tanks for cleaning the verbose strings

gandhipratik203 (dismissed):
Thanks for the change Akshay. Looks good to me.

brian-hussey:
Need response to @cafalchio's comment before progressing. Moving back to reviewed awaiting feedback.

mcpgateway/bootstrap_db.py
brian-hussey: @Lang-Akshay this is the one I need a comment on from @cafalchio.
"Will the lock expire if the owner doesn't remove it?"

Lang-Akshay: PG session-level advisory locks auto-release when the backend connection closes. No explicit unlock needed for safety.

brian-hussey (approved):
Approved.

PR #4885, opened 2026-05-25
feat(ui-rewrite): Redesigns the MCP Servers page to match the Figma specification, covering both frontend layout and the backend data needed to power it.

marekdano:
Related #4643 

### Frontend (`client/`)

- `ServersTable`: full table redesign — card-style rows with rounded corners, replaced the old two-column status badge with seven columns: Name, Components (tools · resources · prompts), Last response, UUID, Visibility, Status, and Actions.
- Status logic replaced `ServerStatusBadge` with inline icons (`Active / Warning / Offline / Draft`) derived from enabled + reachable + lastSeen.
- Visibility column shows Globe / Shield / Lock icons for public / team / private scopes.
- UUID column includes a one-click copy button with 1.5 s visual feedback (copy icon → check mark).
- Last response reformatted from relative ("3 min ago") to ISO-8601-style timestamp (YYYY-MM-DDTHH:MM:SS) using sv-SE locale; invalid or absent dates display "Never used".
- Components cell reads tool/resource/prompt counts from capabilities.count when available, falling back to key-count for backward compatibility.
- `ServerIcon`: swaps the name-initial letter for the MCPIcon SVG; adds aria-label for accessibility.
- Servers page: reduces heading to text-base, shrinks the "Add server" button, and fixes a bug where changing the page-size dropdown did not re-fetch (limit added to the useMemo dependency array).
- Types: adds `MCPServerCapabilities` interface; adds camelCase aliases (lastSeen, toolCount, createdAt, updatedAt, slug) alongside the existing snake_case fields.

### Backend (`mcpgateway/`)

- `gateway_service.py`: eager-loads tools, prompts, and resources relationships via selectinload when listing gateways, then serialises actual counts into a capabilities dict (`{ tools: { count }, prompts: { count }, resources: { count } }`).

### Tests

- `ServersTable.test.tsx` — 370-line suite covering loading state, all column headers, component counts, last-seen formatting, UUID copy, visibility labels, status logic (`Active / Warning / Offline / Draft`), and multi-row rendering.
- `ServerIcon.test.tsx` — verifies aria-label and SVG rendering.
- `test_gateway_service_helpers.py` — three new unit tests for convert_gateway_to_read: relationships present, relationships absent, and empty relationship lists.
- `useMCPServerForm.test.ts` — mocks `triggerOAuthAuthorization` in two `OAuth` flow tests to prevent `window.open` side effects and assert the correct gateway ID is passed.

Screenshot 2026-05-25 at 16 19 44

Screenshot 2026-05-25 at 16 20 06


gcgoncalves (commented):
Should we get a BE dev to review the API updates?

a-effort:
Nice work! Couple small change suggestions: 

1. Change the button label to `Save changes` (instead of _Connect server_) for the Edit MCP server form: 
button label



2. In client/src/components/servers/ServersTable.tsx, update tokens for `Last response` + `UUID`to use the same light/dark text tokens as `Visibility`:

Last response cell: `text-neutral-600 dark:text-neutral-400`
UUID button: `text-neutral-600 dark:text-neutral-400`
Hover: `hover:text-neutral-900 dark:hover:text-neutral-200` (so hover stays readable in both themes)
lt
dk

---

Other small styling adjustments can be added later. Ty! 🙌

mcpgateway/services/gateway_service.py
```diff
@@ -1911,8 +1911,17 @@ async def list_gateways(
...
+        query = (
+            select(DbGateway)
+            .options(
+                joinedload(DbGateway.email_team),
+                selectinload(DbGateway.tools),
+                selectinload(DbGateway.prompts),
+                selectinload(DbGateway.resources),
+            )
+            .order_by(desc(DbGateway.created_at), desc(DbGateway.id))
+        )
```
ja8zyjits, line 1924: Hi Marek,
Would you like to update `list_gateways_for_user` too?

ja8zyjits (commented):
Please check this and let me know. Its not a Blocker, I shall approve it if you think it is not needed

mcpgateway/services/gateway_service.py
```diff
@@ -4626,9 +4635,9 @@ def convert_gateway_to_read(self, gateway: DbGateway) -> GatewayRead:
         gateway_dict["version"] = getattr(gateway, "version", None)
         gateway_dict["team"] = getattr(gateway, "team", None)
 
-        # Populate tool count from the eagerly-loaded tools relationship when available
-        tools_rel = gateway.__dict__.get("tools")
-        gateway_dict["tool_count"] = len(tools_rel) if tools_rel is not None else 0
+        gateway_dict["tool_count"] = len(getattr(gateway, "tools", []))
+        gateway_dict["prompt_count"] = len(getattr(gateway, "prompts", []))
+        gateway_dict["resource_count"] = len(getattr(gateway, "resources", []))
```
ja8zyjits, line 4647: Also lets use the earlier access mechanism, because it works well with eagerly-loaded relationships.


```python
# Populate from the eagerly-loaded tools relationship when available, this helps to log or fail in the else condition if needed in future.
gateway_dict["tool_count"] = len(gateway.tools) if gateway.__dict__.get("tools") else 0
gateway_dict["prompt_count"] = len(gateway.prompts) if gateway.__dict__.get("prompts") else 0
gateway_dict["resource_count"] = len(gateway.resources) if gateway.__dict__.get("resources") else 0
```

ja8zyjits (commented):
Also this for the sake of continuity with the previous implementation.

mcpgateway/services/gateway_service.py
marekdano: Yes, updated `list_gateways_for_user` to also eager-load tools, prompts, and resources via `selectinload` -  without it, `convert_gateway_to_read` would always return `0` counts for that code path since `gateway.__dict__.get("tools")`, etc. would be falsy. The fix is in the latest commit.

mcpgateway/services/gateway_service.py
marekdano: Done!

client/src/types/server.ts
```diff
@@ -18,14 +18,31 @@ export interface BaseServer {
...
+  completions?: Record<string, unknown>;
+}
+
 export interface MCPServer extends BaseServer {
   url: string;
   transport: "SSE" | "STREAMABLEHTTP";
   reachable: boolean;
+  capabilities?: MCPServerCapabilities;
   last_seen?: string;
+  lastSeen?: string;
```
gcgoncalves, line 35: Why do we need duplicate fields using both camelCase and snake_case?

client/src/components/servers/ServersTable.tsx
```diff
@@ -60,46 +144,115 @@ export function ServersTable({ servers, isLoading, onEdit, onDelete, onTest }: S
...
+                <TableCell className="px-4 py-2.5">
+                  <button
+                    type="button"
+                    onClick={() => handleCopy(server.id)}
+                    className="inline-flex items-center gap-2 text-xs text-neutral-600 dark:text-neutral-400 transition hover:text-neutral-900 dark:hover:text-neutral-200"
+                    aria-label={`Copy UUID for ${server.name}`}
+                  >
+                    <span className="max-w-[180px] truncate">{server.id}</span>
+                    {copiedId === server.id ? (
+                      <Check className="h-3.5 w-3.5 text-emerald-400" />
```
gcgoncalves: Add a sr-only label here, for screen readers.

client/src/components/servers/ServersTable.tsx
```diff
@@ -60,46 +144,115 @@ export function ServersTable({ servers, isLoading, onEdit, onDelete, onTest }: S
...
+                    ) : (
+                      <Copy className="h-3.5 w-3.5" />
```
gcgoncalves: Add `aria-hidden` to the checks:
```jsx
  {copiedId === server.id ? (
    <Check className="h-3.5 w-3.5 text-emerald-400" aria-hidden="true" />
  ) : (
    <Copy className="h-3.5 w-3.5" aria-hidden="true" />
  )}
```

client/src/components/servers/ServersTable.tsx
```diff
@@ -60,46 +144,115 @@ export function ServersTable({ servers, isLoading, onEdit, onDelete, onTest }: S
...
+                  <div className="inline-flex items-center gap-1.5 text-xs text-neutral-600 dark:text-neutral-400">
+                    <VisibilityIcon className="h-3.5 w-3.5" />
+                    <span>{visibility.label}</span>
+                  </div>
+                </TableCell>
+                <TableCell className="px-4 py-2.5">
+                  <div
+                    className={`inline-flex items-center gap-1.5 text-xs ${statusConfig.className}`}
+                  >
+                    <StatusIcon className="h-3.5 w-3.5" />
```
gcgoncalves, line 267: Add an `aria-hidden="true"` and a `focusable="false"` to this element to mark it as decorative

client/src/components/servers/ServersTable.tsx
```diff
@@ -45,6 +117,18 @@ interface ServersTableProps {
 }
 
 export function ServersTable({ servers, isLoading, onEdit, onDelete, onTest }: ServersTableProps) {
+  const [copiedId, setCopiedId] = useState<string | null>(null);
+
+  const handleCopy = async (value: string) => {
+    try {
+      await navigator.clipboard.writeText(value);
+      setCopiedId(value);
+      window.setTimeout(() => setCopiedId((current) => (current === value ? null : current)), 1500);
```
gcgoncalves: Extract this magic number (`1500`) into a constant (e.g.: `COPY_FEEDBACK_DURATION_MS`).

gcgoncalves: `setTimeout` is not cleaned up if component unmounts, might lead to memory leaks if user navigates away before timeout completes:

Suggestion: 
```jsx
const timeoutRef = useRef<number | null>(null);

const handleCopy = async (value: string) => {
  try {
    await navigator.clipboard.writeText(value);
    setCopiedId(value);
    
    // Clear existing timeout
    if (timeoutRef.current) {
      window.clearTimeout(timeoutRef.current);
    }
    
    // Set new timeout with cleanup
    timeoutRef.current = window.setTimeout(() => {
      setCopiedId((current) => (current === value ? null : current));
      timeoutRef.current = null;
    }, COPY_FEEDBACK_DURATION_MS);
  } catch (error) {
    console.error("Failed to copy server id:", error);
  }
};

// Cleanup on unmount
useEffect(() => {
  return () => {
    if (timeoutRef.current) {
      window.clearTimeout(timeoutRef.current);
    }
  };
}, []);
```

gcgoncalves (commented):
A solid PR! I've left some suggestions, let me know if you have any questions.

client/src/types/server.ts
marekdano: We do not need "snake_case" fields. The API response contains only "camelCase" fields.
I've removed all "snake_case" fields.

gcgoncalves (approved):

ja8zyjits (approved):
LGTM

PR #4893, opened 2026-05-26
bugfix/4304-gateway-federation-drops-upstream-tools

### LINKED ISSUE #4304: [BUG][MCP-PROTOCOL]: Gateway federation drops a subset of upstream tools
## Summary

Covers **GAP-008** from `tests/protocol_compliance/COMPLIANCE_GAPS.md` (PR #4301).

The compliance reference server registers 136 tools (16 named + 120 `stub_NNN` pagination stubs). After gateway federation, `compliance-reference-*` only surfaces a subset on either `gateway_proxy` or `gateway_virtual`. Observed drops include `boom`, `bump_subscribable`, `mutate_tool_list`, `long_running`; exact set varies and should be re-probed when this is investigated.

## Expected

The gateway should federate every tool the upstream advertises, provided the tool passes the gateway's validator layer. If a tool is intentionally rejected (e.g. missing input schema), the rejection should be observable — a diagnostic log line or a /health-style endpoint exposing the federation-drop list — so operators can confirm "dropped intentionally" vs "dropped silently."

## Observed pattern

Not obviously a common property of the dropped tools:
- `long_running` takes `duration_seconds: float = 30.0` — has args.
- `boom` takes no args and raises — no args, no successful output.
- `bump_subscribable` / `mutate_tool_list` — no args, return int/str.

Possible hypotheses:
- Gateway validates return type / requires an `outputSchema`.
- Gateway rejects tools that raise during introspection probes.
- Federation has a size/quota that truncates at 132 (131 + 120 stubs vs. 136 + 120 = wrong math — actual numbers need re-probing).

Investigation needed to confirm the discriminator.

## Reproduction

```
pytest tests/protocol_compliance/test_drift.py::test_drift_tool_names -v
```

XFAIL on `gateway_proxy-http` and `gateway_virtual-http`; drift normalization strips the slug prefix so the comparison is apples-to-apples.

Also blocks the request-tied notification tests listed in GAP-008's "Tests" row (they can't trigger the mutator if the mutator is dropped).

## Acceptance criteria

```gherkin
Scenario: Every upstream tool surfaces through federation
  Given an upstream registered via POST /gateways
  And the upstream advertises a tool with a valid input schema
  When the gateway federates the upstream
  Then the tool is visible via tools/list on both /mcp/ (proxy) and /servers/{id}/mcp/ (virtual)
  Or the gateway emits a diagnostic naming the dropped tool and the reason
```

## How to verify

Drop the `xfail_on` calls listed under GAP-008 in `COMPLIANCE_GAPS.md`. Confirm:
- `test_drift_tool_names` passes (normalized tool-name sets match across reference / gateway_proxy / gateway_virtual).
- `test_tools.py::test_tool_error_is_surfaced_as_is_error` passes on gateway targets (exercises `boom`).
- `test_utilities.py::test_long_running_tool_is_cancellable` passes on gateway targets.

## Related

- PR #4301 — harness.
- #4302 / #4303 — sibling relay gaps; the list_changed / cancellation tests in GAP-008's row are also blocked by those.

Nayana-R-Gowda:
closes #4304 
Root cause
session.list_tools() was called once per federation path, discarding nextCursor. The reference server registers 120 stub_NNN tools beyond the first page; tools on subsequent pages were silently dropped.

Fix
Added _fetch_all_tools(session) static method (gateway_service.py:5507) that loops over paginated responses, then replaced response = await session.list_tools() / tools = response.tools with tools = await self._fetch_all_tools(session) at all three call sites:
connect_to_sse_server_without_validation (line 5616)
connect_to_sse_server (line 5794)
connect_to_streamablehttp_server (line 5959)

Test Results

  ✅ test_drift_tool_names: PASSED - All 136 tools now surface correctly across all targets
  ✅ test_tool_error_is_surfaced_as_is_error: PASSED - The boom tool is now available
  ✅ test_long_running_tool_is_cancellable: PASSED - The long_running tool is now available

 Documentation Updates

  ✅ COMPLIANCE_GAPS.md: Properly updated to mark GAP-008 as closed with detailed explanation
  ✅ Test xfail markers removed: All GAP-008 related xfail decorators have been removed from:
  - test_drift.py
  - test_tools.py
  - test_utilities.py
  - test_notifications.py (updated to attribute remaining failures to other gaps)

### DIFF
--- a/mcpgateway/services/gateway_service.py
+++ b/mcpgateway/services/gateway_service.py
@@ -6086,6 +6086,29 @@ class GatewayService(BaseService):  # pylint: disable=too-many-instance-attribut
         """
         await self._event_service.publish_event(event)
 
+    @staticmethod
+    async def _fetch_all_tools(session: ClientSession) -> list:
+        """Fetch all tools from an initialized MCP session, handling pagination.
+
+        Iterates over paginated ``tools/list`` responses until the server
+        returns no ``nextCursor``, collecting every tool into a single list.
+
+        Args:
+            session: An initialized MCP ``ClientSession``.
+
+        Returns:
+            List of ``mcp.types.Tool`` objects from every page.
+        """
+        all_tools: list = []
+        cursor: str | None = None
+        while True:
+            response = await session.list_tools(cursor=cursor)
+            all_tools.extend(response.tools)
+            if not response.nextCursor:
+                break
+            cursor = response.nextCursor
+        return all_tools
+
     def _validate_tools(self, tools: list[dict[str, Any]], context: str = "default") -> tuple[list[ToolCreate], list[str]]:
         """Validate tools individually with richer logging and error aggregation.
 
@@ -6172,8 +6195,7 @@ class GatewayService(BaseService):  # pylint: disable=too-many-instance-attribut
                     capabilities = response.capabilities.model_dump(by_alias=True, exclude_none=True)
                     logger.debug("Server capabilities: %s", capabilities)
 
-                    response = await session.list_tools()
-                    tools = response.tools
+                    tools = await self._fetch_all_tools(session)
                     tools = [tool.model_dump(by_alias=True, exclude_none=True, exclude_unset=True) for tool in tools]
 
                     tools, validation_errors = self._validate_tools(tools, context="oauth")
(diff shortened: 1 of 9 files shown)


jonpspri:
Closing this PR. Issue #4304 was closed as NOT_PLANNED, deprioritizing the pagination fix for upstream tool federation. The code in this PR (the `_fetch_all_tools()` cursor-loop and associated xfail marker removals) is not in main, and the compliance test suite it modifies has been removed by PR #5935. If the tool-pagination gap is revisited, this branch serves as a good reference implementation.

PR #4897, opened 2026-05-26
docs: clarify section 14 manual testing expected behaviours

### LINKED ISSUE #4896: [DOCS]: Section 14 manual testing — three undocumented behaviours
## Summary

Section 14 of the release management guide omits five behaviours discovered during v1.0.2 manual testing that cause failures or confusion when following the steps.

### 14 (preamble) — `.env` ships with a placeholder `JWT_SECRET_KEY`

The default `.env` contains `JWT_SECRET_KEY=__REPLACE_ME__...`. Fresh environments that skip `make init-secrets` before starting the stack will have token generation silently succeed but authentication fail — the token is signed with the placeholder key, which does not match the key the running gateway uses.

**Expected after fix:** Doc adds a prerequisite block at the top of section 14 instructing testers to run `make init-secrets` and verify `JWT_SECRET_KEY` is no longer the placeholder before proceeding.

### 14.1 — Token generation uses a hardcoded `--secret` value

The original `create_jwt_token` command used `--secret my-test-key-but-now-longer-than-32-bytes`. On any real deployment this key will not match `JWT_SECRET_KEY` in `.env`, so every API call made with the generated token will return 401.

**Expected after fix:** Doc extracts `JWT_SECRET_KEY` from `.env` via `grep` and passes `"$JWT_SECRET_KEY"` to `--secret`, keeping the token and the gateway in sync.

### 14.2 / 14.3 — `/tools` API returns `gateway_id`, not `gateway_name`

The `ToolRead` schema exposes `gateway_id` (UUID) but no human-readable `gateway_name` field. Callers following the verification step expect to see which gateway a tool belongs to by name; instead they get a UUID with no explanation.

**Expected after fix:** Doc notes that `gateway_name` is absent by design and directs callers to `GET /gateways/<gateway_id>` to resolve the name.

### 14.4 — Export command missing `MCPGATEWAY_BEARER_TOKEN`

The original export command `HOST=localhost PORT=8080 python -m mcpgateway.cli export --out release-test-export.json` fails with `No authentication configured` because `MCPGATEWAY_BEARER_TOKEN` is not set. The export silently exits, leaving a stale or empty output file.

**Expected after fix:** Doc prepends `MCPGATEWAY_BEARER_TOKEN=$MCPGATEWAY_BEARER_TOKEN` to the export command so authentication is passed through.

### 14.4 — Export `tools` count reflects only standalone tools, not gateway-discovered tools

`export_service.py` explicitly filters out tools where `integration_type == "MCP" and gateway_id is set` (ephemeral tools re-discovered when the gateway is imported). A tester with 12 gateway-discovered tools and 1 standalone tool sees `tools: 1` in the export with no explanation.

**Expected after fix:** Doc explains that MCP gateway-discovered tools are excluded from export and are re-discovered on import; tells the tester what to verify instead (`entities.gateways`, `entities.servers`).

### 14.5 — Streamable HTTP requires `initialize` before `tools/list`

MCP Streamable HTTP is a stateful protocol: clients must complete an `initialize` handshake and receive a `Mcp-Session-Id` header before sending `tools/list`. MCP Inspector handles this silently. Raw `curl` without `initialize` receives `-32600 Missing session ID`. The doc gives no hint this is required.

**Expected after fix:** Doc warns that session establishment is required; MCP Inspector handles it automatically; raw clients must send `initialize` first.

## References

- `mcpgateway/schemas.py:1570` — `ToolRead` schema (no `gateway_name` field)
- `mcpgateway/services/export_service.py:420` — MCP tool exclusion filter
- `mcpgateway/transports/streamablehttp_transport.py:1402–1404` — `session.initialize()` called before `tools/list`

msureshkumar88:
## Summary

Documents five behaviours in section 14 that were undocumented or incorrect and caused failures during v1.0.2 manual testing. No code changes — doc only.

**14 (preamble) — `make init-secrets` prerequisite**
The default `.env` ships with `JWT_SECRET_KEY=__REPLACE_ME__...`. Fresh environments that skip `make init-secrets` generate tokens signed with the wrong key; every API call returns 401 with no obvious cause. Added a prerequisite block at the top of section 14 with the `make init-secrets` command and a `grep` check.

**14.1 — Token generation used a hardcoded `--secret` value**
Original command passed `--secret my-test-key-but-now-longer-than-32-bytes`, which will not match `JWT_SECRET_KEY` on any real deployment. Fixed to extract `JWT_SECRET_KEY` from `.env` via `grep` and pass `"$JWT_SECRET_KEY"` to `--secret`.

**14.2 / 14.3 — `gateway_name` absent from `/tools` API response**
The `ToolRead` schema returns `gateway_id` (UUID) but not a human-readable gateway name. Added an "Expected" note explaining this is by design and directing callers to `GET /gateways/<gateway_id>`.

**14.4 — Export command missing `MCPGATEWAY_BEARER_TOKEN`**
`python -m mcpgateway.cli export` requires auth; the original command omitted the env var and failed with `No authentication configured`. Fixed by prepending `MCPGATEWAY_BEARER_TOKEN=$MCPGATEWAY_BEARER_TOKEN` to the export command.

**14.4 — Export excludes MCP gateway-discovered tools**
`export_service.py` filters out tools with `integration_type == "MCP" and gateway_id set` — they are ephemeral and re-discovered on import. Added an "Expected" note explaining the filter, what is included, and what to verify in the export file.

**14.5 — Streamable HTTP requires `initialize` before `tools/list`**
Streamable HTTP is stateful. Clients must complete the `initialize` handshake and receive a `Mcp-Session-Id` before calling `tools/list`. MCP Inspector handles this silently; raw `curl` without `initialize` returns `-32600 Missing session ID`. Added a note explaining the protocol requirement.

Closes #4896

### DIFF
--- a/docs/docs/development/release-management.md
+++ b/docs/docs/development/release-management.md
@@ -1012,7 +1012,7 @@ export MCPGATEWAY_ADMIN_TOKEN=$(./.venv/bin/python -m mcpgateway.utils.create_jw
   --username admin@example.com \
   --admin \
   --exp 10080 \
-  --secret "${JWT_SECRET_KEY:-my-test-key-but-now-longer-than-32-bytes}" \
+  --secret "$JWT_SECRET_KEY" \
   --algo HS256 \
   2>/dev/null | tail -n 1)
 
--- a/docs/docs/development/release-management.md
+++ b/docs/docs/development/release-management.md
@@ -1205,7 +1205,18 @@ The migration test suite follows an **n-2 support policy** and tests sequential 
 
 ## 14. Manual Testing
 
-These tests verify core user-facing workflows that automated tests do not fully cover. Build and start the compose stack first:
+These tests verify core user-facing workflows that automated tests do not fully cover.
+
+Before starting, ensure `.env` has a real JWT secret (the default `.env` ships with a placeholder that breaks token authentication):
+
+```bash
+# Generate secrets and merge into .env (skip if already done)
+make init-secrets
+# Confirm JWT_SECRET_KEY is no longer the placeholder
+grep JWT_SECRET_KEY .env   # must not contain __REPLACE_ME__
+```
+
+Build and start the compose stack:
 
 ```bash
 make docker-prod && make testing-up
@@ -1216,10 +1227,13 @@ make docker-prod && make testing-up
 Create a token for API and client access:
 
 ```bash
+# JWT_SECRET_KEY must match the value in .env (run make init-secrets if not set)
+export JWT_SECRET_KEY=$(grep '^JWT_SECRET_KEY=' .env | cut -d= -f2-)
+
 export MCPGATEWAY_BEARER_TOKEN=$(python -m mcpgateway.utils.create_jwt_token \
   --username admin@example.com \
   --exp 10080 \
-  --secret my-test-key-but-now-longer-than-32-bytes)
+  --secret "$JWT_SECRET_KEY")
 
 export BASE_URL="http://localhost:8080"
 ```
@@ -1246,6 +1260,8 @@ curl -s -X POST -H "Authorization: Bearer $MCPGATEWAY_BEARER_TOKEN" \
 curl -s -H "Authorization: Bearer $MCPGATEWAY_BEARER_TOKEN" $BASE_URL/tools | jq
 ```
 
+**Expected:** Each tool object contains `gateway_id` (UUID of the registering gateway) but no `gateway_name` field — the `/tools` API does not denormalize the gateway name. To resolve a human-readable name, call `GET /gateways/<gateway_id>`. This is by design.
+
 ### 14.3 Register an MCP server via Streamable HTTP
 
 Streamable HTTP transport requires the `--expose-streamable-http` flag and an explicit `"transport":"STREAMABLEHTTP"` field in the registration payload — the gateway does not auto-detect transport from the URL:
@@ -1268,6 +1284,8 @@ curl -s -X POST -H "Authorization: Bearer $MCPGATEWAY_BEARER_TOKEN" \
 curl -s -H "Authorization: Bearer $MCPGATEWAY_BEARER_TOKEN" $BASE_URL/tools | jq
 ```
 
+**Expected:** Tools from both gateways appear in the catalog, each with `gateway_id` populated. As with SSE, `gateway_name` is not included in the response; use `GET /gateways/<gateway_id>` to resolve the name.
+
 ### 14.4 Create a virtual server and export it
 
 The request body must wrap the server fields under a `"server"` key:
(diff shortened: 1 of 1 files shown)


Yosiefeyob (commented):
The Section 14 rewrite looks good overall, Verified the documented behaviours against the running gateway and source where appropriate. One separate Section 12.4 change still needs a small follow-up.

### Manual Testing

* Built and started the gateway locally (`make docker-prod` + `compose-lite-up`).
* Confirmed `/health` returned `200`, both gateway replicas were healthy, and `.env` contained a real `JWT_SECRET_KEY`.
* Live-tested the JWT secret flow against `http://localhost:8080/tools`.
* Source-checked the remaining documented behaviours.

### Verified

* ✅ **Preamble / `make init-secrets`** — `.env.example` does ship with the placeholder secret, and the new prerequisite is appropriate.
* ✅ **14.1 secret extraction** — live-confirmed: a token signed with the key extracted from `.env` returns `HTTP 200`.
* ✅ **14.4 export auth** — the CLI requires authentication; passing `MCPGATEWAY_BEARER_TOKEN` fixes a real failure path.
* ✅ **14.4 MCP tool export note** —  gateway-discovered MCP tools are filtered from export and rediscovered later.
* ✅ **14.5 initialize/session note** —  raw Streamable HTTP clients need the initialize/session flow before `tools/list`.

### Follow-up — Section 12.4

The Section 12.4 change removes the hardcoded fallback secret, which is the right direction, but it needs one more line before the token command.

Right now it uses:

```bash
--secret "$JWT_SECRET_KEY"
```

At that point in the guide, `JWT_SECRET_KEY` has not been loaded from `.env` yet. In my local run, with the gateway up, that meant the token was signed with an empty secret and `/tools` returned `401`. After loading the key from `.env`, the same flow returned `200`.

Adding the same extraction used later in `14.1` fixes it:

```bash
export JWT_SECRET_KEY=$(grep '^JWT_SECRET_KEY=' .env | cut -d= -f2-)
```

So I think `12.4` should either add that line before generating `MCPGATEWAY_ADMIN_TOKEN`, or revert the `12.4` command.

### Summary

Great improvement. Section 14 looks accurate, and the main secret fix is live-verified. If you could kindly add the missing Section 12.4 `JWT_SECRET_KEY` export before merge. Thanks!

msureshkumar88:
Thanks for the thorough review and live verification, @Yosiefeyob!

The Section 12.4 follow-up is now addressed — added the same `JWT_SECRET_KEY` extraction line used in 14.1 before the `MCPGATEWAY_ADMIN_TOKEN` command, so the token is signed with the correct key from `.env` rather than an empty string.

```bash
export JWT_SECRET_KEY=$(grep '^JWT_SECRET_KEY=' .env | cut -d= -f2-)
```

Yosiefeyob (approved):
Re-tested against the live gateway, Section 12.4 now exports JWT_SECRET_KEY from .env before generating the admin token, the token signs with the real key, and /tools returns 200. Follow-up resolved — LGTM! 🚀

ja8zyjits (approved):
LGTM
process followed.

PR #4900, opened 2026-05-26
fix(init-secrets): merge into .env when present, warn on real values

### LINKED ISSUE #4899: [BUG]: make init-secrets always writes .env.secrets, ignores existing .env; no pre-flight secrets check on docker-prod/testing-up
## Summary

`make init-secrets` unconditionally writes to `.env.secrets` regardless of
whether `.env` already exists, requiring a manual merge step every time.
Additionally, `make docker-prod` and `make testing-up` can be launched
silently with placeholder (`__REPLACE_ME__` / `changeme`) secrets, leading
to insecure deployments.

## Steps to Reproduce

1. `cp .env.example .env`  — `.env` now has `__REPLACE_ME__` placeholder values
2. `make init-secrets`     — writes `.env.secrets`, does **not** update `.env`
3. `make docker-prod`      — starts with weak secrets, no warning emitted

## Expected Behaviour

- When `.env` exists, `make init-secrets` should merge the generated secrets
  directly into `.env` (preserving all other variables).
- When `.env` is absent, fall back to creating `.env.secrets` as before.
- Running `make docker-prod` or `make testing-up` with placeholder secrets
  should print a clear warning directing the operator to run `make init-secrets`
  first.
- Overwriting an existing `.env` key that already holds a real (non-placeholder)
  value should require explicit confirmation; non-interactive runs should abort
  safely.

## Impact

- Operators who follow the quickstart (`cp .env.example .env && make init-secrets`)
  end up running services with weak placeholder secrets.
- No guard prevents launching production or testing stacks with uninitialized
  credentials.

msureshkumar88:
## Summary

- `make init-secrets` now auto-detects `.env`: merges all 4 secrets
  (`JWT_SECRET_KEY`, `AUTH_ENCRYPTION_SECRET`, `BASIC_AUTH_PASSWORD`,
  `PLATFORM_ADMIN_PASSWORD`) in-place when `.env` exists, preserving all
  other variables; falls back to writing `.env.secrets` when `.env` is absent
- `--output` default changed from `.env.secrets` to `None` so auto-detection
  can take effect; explicit `--output`, `--force`, and `--stdout` unchanged
- Before overwriting a key that already holds a real (non-placeholder) value,
  prompts `Continue? [y/N]`; `EOFError` (non-interactive / piped) treats as
  `N` and aborts safely without modifying `.env`
- New `make check-secrets` target warns when `.env` contains `__REPLACE_ME__`
  or `changeme` values; wired as a prerequisite to `docker-prod` and
  `testing-up` so operators are notified before launching services with weak
  secrets
- 6 new unit tests added (44 total, all passing):
  - auto-merge into `.env` when present
  - fallback to `.env.secrets` when `.env` absent
  - no confirmation prompt for placeholder/weak values
  - confirmation prompt + proceed on `y`
  - confirmation prompt + abort on `n`
  - safe abort on `EOFError` (non-interactive)

Closes #4899

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -346,6 +346,20 @@ check-env-dev:
 init-secrets: ## Generate secure secrets for the gateway (US-3)
 	python3 -m mcpgateway.scripts.init_secrets
 
+.PHONY: check-secrets
+check-secrets: ## Warn if .env secrets are still placeholders (run 'make init-secrets' first)
+	@if [ ! -f .env ]; then \
+		echo ""; \
+		echo "⚠️  WARNING: .env not found."; \
+		echo "   Run 'make init-secrets' to generate secrets before starting services."; \
+		echo ""; \
+	elif grep -qE '^(JWT_SECRET_KEY|AUTH_ENCRYPTION_SECRET|BASIC_AUTH_PASSWORD|PLATFORM_ADMIN_PASSWORD)=(__REPLACE_ME__|changeme)' .env 2>/dev/null; then \
+		echo ""; \
+		echo "⚠️  WARNING: .env contains placeholder secrets (__REPLACE_ME__ / changeme)."; \
+		echo "   Run 'make init-secrets' before starting production or testing services."; \
+		echo ""; \
+	fi
+
 # =============================================================================
 # ▶️ SERVE
 # =============================================================================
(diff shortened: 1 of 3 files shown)


msureshkumar88:
Closing this PR — the auto-merge-into-`.env` approach is the wrong direction for production deployments.

**Why:**
- `.env` files in production are typically owned by ops/infra, managed by secrets managers (Vault, AWS Secrets Manager, Kubernetes Secrets), or baked in via CI/CD pipelines. A tool that silently mutates `.env` in-place can corrupt those workflows or overwrite values that were intentionally set.
- The confirmation prompt protects against the worst case, but it creates a footgun in non-interactive environments (CI, containers, init scripts) where `EOFError` silently aborts — giving operators no feedback that secrets were not initialized.
- The safer contract is the original: `make init-secrets` writes to `.env.secrets` (or stdout), and the operator explicitly merges. This keeps the tool idempotent and non-destructive.

The `make check-secrets` warning target from this PR is a good idea and worth landing separately — but the in-place `.env` mutation should not ship.

PR #4905, opened 2026-05-26
fix: warn on __REPLACE_ME__ placeholder in non-production environments

### LINKED ISSUE #4904: [FIX]: Warn instead of fatal error on __REPLACE_ME__ placeholders in non-production environments
## Summary

`make init-secrets` generates strong secrets, but during local development or CI, the `.env` file may still contain `__REPLACE_ME__` placeholder values. Previously, any unset placeholder caused a fatal `SecurityConfigurationError` on startup — even in development environments where this is a normal transient state.

## Problem

Startup fails hard with `SecurityConfigurationError` in development/staging when `__REPLACE_ME__` placeholders are present in `jwt_secret_key` or `auth_encryption_secret`, even though these environments are not at risk from placeholder values. This creates unnecessary friction for first-time contributors and CI pipelines that haven't run `make init-secrets` yet.

## Expected Behavior

- **Production**: `__REPLACE_ME__` placeholders remain fatal (`SecurityConfigurationError`) — no change.
- **Non-production** (development, staging, CI): emit a `SECURITY WARNING` log and allow startup, consistent with how weak secrets are handled in development.
- **Weak/default secrets**: rejection in staging + production is unchanged.

## Fix

In `mcpgateway/config.py`, gate the `SecurityConfigurationError` raise on `env == "production"`. All other environments fall through to a `logger.warning(...)` call that clearly labels the issue without blocking startup.

```python
if val.lower().startswith("__replace_me__"):
    if env == "production":
        raise SecurityConfigurationError(...)
    else:
        logger.warning("🔓 SECURITY WARNING - ...")
```

## Files Changed

- `mcpgateway/config.py` — conditional error vs warning based on environment
- `tests/unit/test_cwe321_hardcoded_jwt.py` — updated tests to cover both paths

msureshkumar88:
## Summary

- Gates `SecurityConfigurationError` for `__REPLACE_ME__` placeholders on `env == "production"` only
- Non-production environments (dev, staging, CI) now emit `🔓 SECURITY WARNING` and continue startup
- Weak/default secret rejection in staging + production is unchanged
- Updates unit tests to cover both the fatal (production) and warning (non-production) paths

Closes #4904


brian-hussey (changes requested):
We still need to remove this section in the https://github.com/IBM/mcp-context-forge/blob/main/mcpgateway/__main__.py#L16-L24

Ånd we need these checks in the mcpgateway/main.py as well.
In there if there is an exception raised for this it should exist from the application if we're in a production environment.

msureshkumar88:
Thanks for the review, Brian! Both points addressed in the latest commits:

**1. `mcpgateway/__main__.py` lines 16–24 removed** — stripped the `ensure_env_file_secrets()` import, call, and print block entirely. The `main()` function now just starts uvicorn directly. The auto-generation path was redundant given the new warning/error flow introduced in `config.py`.

**2. `__REPLACE_ME__` check added to `validate_security_configuration()` in `main.py`** — after `get_settings()`, we now iterate over `jwt_secret_key` and `auth_encryption_secret`. If either starts with `__replace_me__`:
- **production** → `SecurityConfigurationError` raised → caught by the existing `FAIL-CLOSED` handler → `sys.exit(1)`
- **non-production** → `logger.warning(...)`, startup continues

Tests updated to cover both paths.

brian-hussey (approved):
LGTM. Thank you for the quick turn around.

PR #4909, opened 2026-05-26
Release/v1.0.2

Lang-Akshay:
## Summary

- **Version bump** `1.0.1 → 1.0.2` across package, init, and Helm chart
- **Security**: rmcp dep bump `0.14.0 → 1.4.0` (Dependabot); `__REPLACE_ME__` placeholder warning in non-prod; `TOKEN_EXPIRY` capped at 1440; base images updated to 10.2
- **Dependency updates**: Python (`pyproject.toml`, `uv.lock`), Go (`mcp-go 0.54.1`), Cargo (`Cargo.lock`), docs (`docs/requirements.txt`), npm audit fixes
- **SSRF migration doc**: clarified redirect-protection guide with security rationale
- **Tests**: Playwright + integration test fixes for password change, token scope matrix, team selector, API abuse, plugin ID retrieval

## Bug Fix: A2A Plugin Binding — `plugin_id` AttributeError

`get_all_plugins()` (`plugin_service.py:74`) returns `List[Dict[str, Any]]` — dicts, not objects. The A2A agent plugin-binding template iterated these and accessed `.plugin_id` (attribute), then `p["plugin_id"]` (key) — both fail:

1. `'dict' object has no attribute 'plugin_id'`
2. `KeyError: 'plugin_id'` (after first fix attempt)

Dict keys are `"name"`, `"namespace"`, `"priority"`, etc. — no `"plugin_id"` key exists. The binding model stores the plugin name as its identifier, so the correct value is `p["name"]`. Fix: populate the `<select>` dropdown using `p["name"]` as both value and label.


brian-hussey (approved):
LGTM

PR #4919, opened 2026-05-27
feat(ui-rewrite): Add expose components step for virtual server creation

### LINKED ISSUE #4903: [UI-REWRITE]: add expose tools, resources and prompts page to the UI flow
We should have a new UI after creating an MCP server to expose MCP tools, resources and prompts

Image

marekdano:
Closes #4903 

### What

Adds a post-creation step to the MCP server connection flow that lets users select which tools, resources, and prompts to expose on a new virtual server.

### Changes

`ExposeComponentsForm` (new component)
  - Renders after a gateway is successfully created, replacing the connection form in-place
  - Fetches tools, resources, and prompts scoped to the new gateway via `?gateway_id=`
  - Collapsible accordion sections per component type, each with a select-all checkbox
  - OAuth toggle to require inbound client authentication on the virtual server (RFC 9728)
  - Skip button to proceed without creating a virtual server; Expose button to submit
  - Error display on creation failure with re-try support

`MCPServerForm`
  - After a successful create (non-edit) submission, transitions to `ExposeComponentsForm` instead of immediately closing
  - Passes the OAuth notification state through so it remains visible across the transition
  - Edit mode is unaffected — form still closes as before

`useMCPServerForm`
  - `onSuccess` callback now receives the API response `(response?: unknown)` so callers can extract the created gateway ID

`virtualServers.ts / types.ts`
  - `buildCreateVirtualServerPayload` now maps `associatedTools, associatedResources, and associatedPrompts` from `CreateServerDetails` into the payload (previously always empty arrays)
  - `CreateServerDetails` interface extended with the three optional association fields

UI primitives
  - Adds `label.tsx` and `radio-group.tsx` shadcn components (needed by the OAuth toggle section)

### Tests

  - `ExposeComponentsForm.test.tsx` — 20 tests covering render, loading, section expand/collapse, per-type selection, OAuth toggle, submission, skip navigation, error state, OAuth notification dismiss, and empty states
  - `virtualServers.test.ts` — 5 new tests for each association combination and defaults
  - `MCPServerForm.test.tsx` — updated to expect the expose step rather than immediate onToggle on submission

Screenshot 2026-05-27 at 14 58 19

Screenshot 2026-05-27 at 14 58 02


gcgoncalves:
Minor changes: 

- Add `aria-expanded`, `aria-controls`, `role="region"` to collapsible sections
- Add `aria-label` to icons

client/src/components/gateways/ExposeComponentsForm.tsx
```diff
@@ -0,0 +1,491 @@
...
+  );
+  const prompts = useMemo(() => (Array.isArray(promptsData) ? promptsData : []), [promptsData]);
+
+  const toolCount = tools.length;
+  const resourceCount = resources.length;
+  const promptCount = prompts.length;
+
+  const isLoading = toolsLoading || resourcesLoading || promptsLoading;
+
+  const toggleSection = (section: string) => {
```
gcgoncalves: It'd be a good idea to wrap the toggle functions (toggleSection, toggleAllTools, toggleTool, toggleAllResources, toggleResource, toggleAllPrompts and togglePrompt) with useCallback, to prevent re-rendering all the children on parent re-render.

gcgoncalves (changes requested):

gcgoncalves (approved):
🐑

a-effort:
Great work getting the functionality in here! We can polish and add an empty state for the expanded state when there are 0 tools / resources / prompt templates, add tooltip content, and address error states as a next.

To address: 
`private | team | public` defaults to public, but the option that is selected for the MCP server should carry over to the virtual server now that we've connected these flows.

PR #4925, opened 2026-05-27
fix(a2a): pass safe headers to tool pre-invoke hooks

### LINKED ISSUE #4895: [BUG]: No A2A tool header passthrough in TOOL_PRE_INVOKE payload
### 🐞 Bug Summary
When invoking an **A2A tool** via `tools/call`, the `payload.headers` passed to plugins on the `TOOL_PRE_INVOKE` hook is hardcoded to `{"Content-Type": "application/json"}`. Unlike the REST/MCP branches, the A2A branch does **not** call `compute_passthrough_headers_cached(...)` and does not honor `ENABLE_HEADER_PASSTHROUGH`, `DEFAULT_PASSTHROUGH_HEADERS`, nor the A2A agent's own `passthrough_headers` allowlist. Any plugin that reads `payload.headers` (e.g. a JWT validation plugin) will always see empty headers for A2A tools, regardless of what the client sends. Reproduced in **v1.0.0** and **v1.0.1**.
 
---
 
### 🧩 Affected Component
 
- [x] `mcpgateway` - API
- [x] Other (explain below)
 
> Specifically: **A2A tool invocation path** in `mcpgateway/services/tool_service.py` and its integration with the **plugin framework** (`TOOL_PRE_INVOKE`). Related to header-passthrough feature (#208) and complementary to `AUTH_HEADER_NAME` (#4494).
 
---
 
### 🔁 Steps to Reproduce
 
1. Deploy ContextForge **v1.0.0 or v1.0.1** with header passthrough enabled:
   ```
   ENABLE_HEADER_PASSTHROUGH=true
   DEFAULT_PASSTHROUGH_HEADERS=["Authorization"]
   ```
2. Register an A2A agent with `passthrough_headers: ["Authorization"]` and `visibility: public`. ContextForge auto-creates an MCP tool of `integration_type=A2A` for it.
3. Register/enable a plugin that hooks `TOOL_PRE_INVOKE` and inspects/uses `payload.headers.root["Authorization"]` (e.g. any custom JWT validation plugin).
4. From a client, invoke the A2A tool:
   ```http
   POST /mcp/  HTTP/1.1
   Authorization: Bearer <any-token>
   Content-Type:  application/json
   Accept:        application/json, text/event-stream
 
   {"jsonrpc":"2.0","id":1,"method":"tools/call",
    "params":{"name":"<a2a-tool-name>","arguments":{"query":"hi"}}}
   ```
5. Observe in plugin logs that `payload.headers` is empty (apart from `Content-Type`), and the plugin can never see the incoming `Authorization` header.
 
---
 
### 🤔 Expected Behavior
For A2A tools, `payload.headers.root` should contain the headers from the incoming request filtered by the passthrough allowlist (global `DEFAULT_PASSTHROUGH_HEADERS` + per-agent `passthrough_headers`), **exactly like the REST/MCP gateway branch does**. In particular, `Authorization` should be available to plugins on the `TOOL_PRE_INVOKE` hook when listed in the allowlist.
 
---
 
### 📓 Logs / Error Output
 
**Broken code (v1.0.1 `mcpgateway/services/tool_service.py:5727`):**
```python
elif tool_integration_type == "A2A" and a2a_agent_endpoint_url:
    headers = {"Content-Type": "application/json"}   # ← hardcoded, request_headers ignored
 
    plugin_manager = await self._get_plugin_manager(plugin_context_id)
    if plugin_manager and plugin_manager.has_hooks_for(ToolHookType.TOOL_PRE_INVOKE) and not skip_pre_invoke:
        ...
        pre_result, context_table = await plugin_manager.invoke_hook(
            ToolHookType.TOOL_PRE_INVOKE,
            payload=ToolPreInvokePayload(
                name=name, args=arguments,
                headers=HttpHeaderPayload(root=headers),   # ← plugin sees {} aside from Content-Type
            ),
            ...
        )
```
 
**Working code in the REST/MCP branch (`tool_service.py:~3965`) for comparison:**
```python
if request_headers:
    headers = compute_passthrough_headers_cached(
        request_headers,
        headers,
        passthrough_allowed,
        gateway_auth_type=gateway_auth_type,
        gateway_passthrough_headers=gateway_payload.get("passthrough_headers") if has_gateway else None,
    )
...
payload=ToolPreInvokePayload(name=name, args=arguments,
                             headers=HttpHeaderPayload(root=dict(runtime_headers)))
```
 
**Plugin-side trace (custom JWT plugin) showing the symptom:**
```
event=validation_start hook=pre resource_type=tool resource=<a2a-tool-name>
event=token_extract_start
event=token_extract_missing source=payload.headers
event=token_retrieval outcome=not_found token_fp=none
PluginViolationError: tool_pre_invoke blocked: MISSING_TOKEN — Token required
```
 
**Suggested minimal patch:**
```python
elif tool_integration_type == "A2A" and a2a_agent_endpoint_url:
    headers = {"Content-Type": "application/json"}
 
    if request_headers:
        passthrough_allowed = global_config_cache.get_passthrough_headers(
            db, settings.default_passthrough_headers
        )
        headers = compute_passthrough_headers_cached(
            request_headers,
            headers,
            passthrough_allowed,
            gateway_auth_type=None,                          # A2A has no gateway-level auth
            gateway_passthrough_headers=a2a_agent_passthrough_headers,
        )
 
    runtime_headers = {str(k): str(v) for k, v in headers.items() if k and v}
 
    if plugin_manager and plugin_manager.has_hooks_for(ToolHookType.TOOL_PRE_INVOKE) and not skip_pre_invoke:
        ...
        pre_result, context_table = await plugin_manager.invoke_hook(
            ToolHookType.TOOL_PRE_INVOKE,
            payload=ToolPreInvokePayload(
                name=name, args=arguments,
                headers=HttpHeaderPayload(root=dict(runtime_headers)),
            ),
            ...
        )
```
 
`a2a_agent_passthrough_headers` is already extracted earlier in `invoke_tool` (Phase 2, before `db.close()`), alongside `a2a_agent_auth_type`, `a2a_agent_endpoint_url`, etc., so no extra DB access is needed inside the post-`db.close()` block.
 
> ⚠️ No secrets/tokens included — all `<placeholder>`.
 
---
 
### 🧠 Environment Info
Retrieved from the `/version` endpoint of the running pod.
 
| Key | Value |
|-----|-------|
| Version or commit | `mcp-contextforge-gateway 1.0.0` (also reproduced against `v1.0.1` source) |
| Runtime | Python 3.11, Uvicorn |
| Platform / OS | Linux (container) |
| Container | OCI image deployed on OpenShift / Kubernetes |
 
---
 
- **Related issues / PRs**
  - #208 — `[FEATURE][AUTH]: HTTP header passthrough to MCP server` (merged; REST-only, did not cover the A2A branch).
  - #685 — `[BUG][SECURITY]: Fixes and security for HTTP header passthrough` (security hardening of the same feature, didn't touch A2A).
  - #4494 — `feat(auth): add configurable JWT authentication header support` (complementary; enables clients to put their own token in `Authorization` for downstream passthrough — but moot for A2A while this bug stands).
- **`AGENT_PRE_INVOKE` hook** is declared in `mcpgateway/plugins/framework/hooks/agents.py` but **not invoked anywhere** in `mcpgateway/services/`. A2A invocations only trigger `TOOL_PRE_INVOKE`, which is exactly the path with the missing passthrough logic. Workarounds based on hooking `AGENT_PRE_INVOKE` therefore don't help.
- **Workaround attempted (does NOT solve it)**: combining `MCP_REQUIRE_AUTH=false`, `MCP_CLIENT_AUTH_ENABLED=false`, `TRUST_PROXY_AUTH=true`, `TRUST_PROXY_AUTH_DANGEROUSLY=true`, `ENABLE_HEADER_PASSTHROUGH=true`, `DEFAULT_PASSTHROUGH_HEADERS=["Authorization"]`, plus a per-agent `passthrough_headers=["Authorization"]`. None of these change the A2A branch behavior — `payload.headers` remains hardcoded.
- Happy to contribute a PR with the patch above + tests if useful.

jstar0:
## Summary

Fixes #4895.

A2A tools created for agents now apply the existing passthrough-header filtering before running `TOOL_PRE_INVOKE`.

The implementation follows the sensitive-header behavior on `main`: tool pre-invoke plugins receive only non-sensitive allowlisted headers, while the outbound A2A invocation receives allowlisted sensitive headers only when `ENABLE_SENSITIVE_HEADER_PASSTHROUGH=true`.

The latest update also aligns generated A2A tools with the direct `/a2a` invocation path: an unset or empty agent `passthrough_headers` allowlist blocks request-header forwarding instead of falling back to the global gateway defaults.

## Reproduction Steps

1. Enable header passthrough.
2. Configure an A2A agent with `passthrough_headers=["Authorization", "X-Tenant-Id"]`.
3. Invoke the generated A2A tool through `tools/call` with matching request headers.
4. Observe that the A2A `TOOL_PRE_INVOKE` payload did not previously receive the correct allowlisted non-sensitive request context, and plugin-modified headers could bypass or diverge from the same allowlist semantics.

## Root Cause

The A2A `tools/call` path initialized plugin payload headers as only `{"Content-Type": "application/json"}` and did not call the shared passthrough-header helper before invoking `TOOL_PRE_INVOKE`.

After the sensitive-header passthrough work on `main`, the A2A tool path also needed to split plugin-visible headers from outbound invocation headers so plugins cannot read sensitive request headers.

During review, a second mismatch was identified: the generated-tool path treated `passthrough_headers=None` as a fallback to global defaults, while the direct A2A service explicitly treats both `None` and `[]` as default-deny. The generated-tool path now follows the same contract.

## Changes

- Carry the A2A agent `passthrough_headers` value across the DB-close boundary in `invoke_tool`.
- Reuse `compute_passthrough_headers_cached()` for A2A tools before `TOOL_PRE_INVOKE`.
- Treat unset and empty A2A agent allowlists as default-deny for both request and plugin-returned headers.
- Filter sensitive headers out of the A2A tool pre-invoke plugin payload.
- Keep outbound A2A sensitive-header forwarding behind `ENABLE_SENSITIVE_HEADER_PASSTHROUGH`.
- Re-filter plugin-modified headers before applying them to the outbound A2A invocation.
- Ensure configured A2A `api_key` auth overwrites any base `Authorization` header before outbound invocation.
- Document the explicit allowlist and default-deny behavior.
- Add regression coverage for plugin payload headers, default sensitive-header blocking, plugin-modified header filtering, unset allowlists, and API-key precedence.

## Verification

```bash
uv run pytest tests/unit/mcpgateway/services/test_tool_service_coverage.py::TestInvokeToolA2A::test_a2a_pre_invoke_blocks_headers_when_agent_allowlist_unset -q
uv run pytest tests/unit/mcpgateway/services/test_a2a_protocol.py tests/unit/mcpgateway/services/test_a2a_agent_invoke_hooks.py tests/unit/mcpgateway/services/test_tool_service_coverage.py tests/unit/mcpgateway/test_a2a_plugin_header_security.py tests/unit/mcpgateway/test_a2a_passthrough_headers.py -k a2a --tb=short
uv run ruff check mcpgateway/services/tool_service.py tests/unit/mcpgateway/services/test_tool_service_coverage.py
uv run ruff format --check mcpgateway/services/tool_service.py tests/unit/mcpgateway/services/test_tool_service_coverage.py
make detect-secrets-hook
git diff --check
```

Results:

| Check | Status |
|---|---|
| Unset-allowlist regression | Passed, 1 test |
| Related A2A coverage | Passed, 262 tests, 1 skipped |
| Targeted Ruff check and format | Passed |
| detect-secrets hook | Passed |
| `git diff --check` | Passed |

## MCP Compliance

## Checklist


lucarlig:
Hello @jstar0, before we can review we need ci fully green, please fix and we can review and merge.

ref: https://github.com/IBM/mcp-context-forge/blob/main/CONTRIBUTING.md

ja8zyjits:
Hi @jstar0 
cc: @shoummu1 

Could you please rebase the branch and address the CI issues? We’ll be happy to review it once that’s done.
We also noticed there hasn’t been an update over the past few days. It would be great if you could provide a quick update and aim to have the fixes in place before **Monday, 22 June 2026.**
This issue has been receiving increased attention, and we would like to help move it to closure. If progress remains limited, we may need to consider reassigning it internally to keep things on track.

Regards,
Jitesh

jstar0:
Rebased this branch onto current main and pushed a signed update. I reran the A2A/header-related Python tests and the Vitest suite with Node 20, and both passed. I will watch the refreshed CI result.

ja8zyjits:
Hi @jstar0 
Please look at these review comments and let me know if they are invalid or needs to be fixed from your end.

## 🚨 Blocking Issues

### 1. A2A agent `passthrough_headers` field undocumented

**Files**: `docs/docs/using/agents/a2a.md`  
**Impact**: Users can't discover/configure per-agent header allowlists

**Fix**: Add to agent registration examples and field table (after line 88):

```markdown
| `passthrough_headers` | List of header names to pass through | `["Authorization", "X-Tenant-Id"]` |
```

Example:
```json
{
  "name": "hello_world_agent",
  "endpoint_url": "http://localhost:9999/",
  "agent_type": "jsonrpc",
  "passthrough_headers": ["Authorization", "X-Tenant-Id"],
  "auth_type": "api_key",
  "auth_value": "your-api-key"
}
```

### 2. Plugin `TOOL_PRE_INVOKE` payload headers not documented for A2A

**Files**: `docs/docs/using/plugins/lifecycle.md`  
**Impact**: Plugin devs can't rely on headers in A2A tool invocations

**Fix**: Add note after line 280:

```markdown
**Note**: For A2A tools, `payload.headers` includes allowlisted request headers filtered by global `DEFAULT_PASSTHROUGH_HEADERS` and agent's `passthrough_headers` config. Matches REST/MCP behavior.
```

---

## ⚠️ Warnings

1. **Header passthrough docs missing A2A section** (`docs/docs/overview/passthrough.md`)  
   Add A2A agent configuration section explaining per-agent allowlists

2. **Config docs don't explain A2A interaction** (`docs/docs/manage/configuration.md:L1099`)  
   Expand `DEFAULT_PASSTHROUGH_HEADERS` description to mention A2A agent interaction

3. **A2A api_key auth precedence change undocumented** (`docs/docs/using/agents/a2a.md:L95`)  
   Add note: configured api_key overwrites base Authorization headers

---

jstar0:
Thanks, I addressed the documentation feedback in the latest push (`b3022edac8b23831da394c2e1435a22094d6380a`):

- Added `passthrough_headers` to the A2A agent registration example and documented it as an optional request field.
- Documented A2A `TOOL_PRE_INVOKE` `payload.headers` behavior for plugin authors.
- Added A2A coverage to the header passthrough overview, including the per-agent allowlist interaction.
- Expanded the `DEFAULT_PASSTHROUGH_HEADERS` configuration description to mention A2A agents.
- Documented that configured A2A `api_key` auth takes precedence for the outbound `Authorization` header.

I also checked the docs diff with `git diff --check`, focused markdownlint checks around the added lines, and a public-surface scan. The latest commit is DCO-signed and GitHub reports the SSH signature as valid.

ja8zyjits:
Hi @jstar0 
Please rebase.
This may help you if you would like to try
```bash
git fetch origin main
git rebase origin/main
# for .secrets.baseline conflict, do the following. For other conflicts merge it manually.
git checkout --theirs .secrets.baseline
git add -u
git rebase --continue
# then you will probably need to run
make detect-secrets-scan && make detect-secrets-audit
git add .secrets.baseline
make pre-commit
git commit -sm "fixed pre-commit"
git push --force-with-lease origin <branch-name>
```

jstar0:
Rebased onto current main and pushed the refreshed branch. Local verification after the rebase: targeted A2A unit coverage passed, targeted Ruff passed, git diff --check passed, and detect-secrets-hook passed after refreshing the secrets baseline line metadata.

jstar0:
Rebased onto current main and refreshed the secrets baseline. The merge conflict was limited to .secrets.baseline line metadata and has been resolved using the repository detect-secrets hook. Local verification after rebase: 109 A2A protocol tests passed, ruff clean on all changed files, detect-secrets-hook clean, DCO-signed, and SSH signature verified.

ja8zyjits:
HI @jstar0 
Please read these reviews and fix  if you think they are correct gaps. Also rebase it. If the review is clear. I will merge it (if needed I will rebase too.)
## 🚨 Critical Issues

### 1. Code-Documentation Mismatch (HIGH)
**Location**: `mcpgateway/services/tool_service.py`

PR claims sensitive-header filtering implemented. **NOT TRUE**.

**Missing from branch (b7bc6d90b)**:
- `filter_sensitive_headers` import/calls
- `enable_sensitive_header_passthrough` setting checks
- Separate `plugin_headers` variable
- Plugin-modified header re-filtering

**Actual code** (lines 5829-5838):
```python
headers = compute_passthrough_headers_cached(...)
pre_invoke_headers = HttpHeaderPayload(root=headers)  # All headers to plugin
```

**Impact**:
- Plugins see sensitive headers (`Authorization`)
- Security boundary not enforced
- Tests reference non-existent behavior

**Fix**: Implement missing logic OR update PR/tests to match actual code.

---

### 2. Plugin Header Bypass (HIGH)
**Location**: `mcpgateway/services/tool_service.py` (lines 5851-5858)

Plugin-modified headers replace base headers without re-filtering.

**Risk**: Plugin injects arbitrary headers (including sensitive) bypassing allowlist.

**Fix**: Re-filter plugin-modified headers against allowlist + sensitive-header policy.

---

### 3. Test-Code Mismatch (MEDIUM)
**Location**: `tests/unit/mcpgateway/services/test_tool_service_coverage.py`

Three tests validate filtering that doesn't exist:
- `test_a2a_pre_invoke_payload_receives_passthrough_headers`
- `test_a2a_pre_invoke_blocks_sensitive_headers_by_default`
- `test_a2a_pre_invoke_modified_headers_are_refiltered`

Tests reference:
- Non-existent `enable_sensitive_header_passthrough` setting
- Assertions that would fail (`Authorization` not in payload)

**Fix**: Run tests. Update code or tests to align.

---

### 4. Documentation Mismatch (MEDIUM)
**Files**: `docs/docs/manage/configuration.md`, `docs/docs/overview/passthrough.md`, `docs/docs/using/plugins/lifecycle.md`

Docs describe `ENABLE_SENSITIVE_HEADER_PASSTHROUGH` and filtering behavior not in code.

**Fix**: Remove docs OR implement feature.

---

## ⚠️ Warnings

### Architecture: Sensitive Headers Visible to Plugins
Current diff passes full header set to `TOOL_PRE_INVOKE`. Test asserts:
```python
assert payload.headers.root == {
    "Authorization": "Bearer client-token",  # Should be filtered
    "X-Tenant-Id": "tenant-123",
}
```

Breaks security invariant: plugins shouldn't see sensitive auth material.

---

jstar0:
Rebased onto current `origin/main` and pushed `18d9c615e610a980a2cd4a732204d47ec989bef1`. The current head now includes `filter_sensitive_headers`, `enable_sensitive_header_passthrough`, `plugin_headers`, and re-filtering of plugin-returned headers in `mcpgateway/services/tool_service.py`, plus the docs and tests for the A2A header path. I reran the focused A2A coverage in a clean shell with the local proxy env unset; the four `TestInvokeToolA2A` cases and `test_a2a_service.py -k passthrough_headers` passed. The code/documentation mismatch in the 2026-06-29 comment appears to describe the pre-rebase head `b7bc6d90b`; please re-review the current head.

ja8zyjits (dismissed):
LGTM

msureshkumar88 (changes requested):
Thanks for tracking down the root cause here — the PR description makes it easy to follow, and the added regression tests do a good job covering the scenarios they target (sensitive-header blocking, plugin-modified-header filtering, api_key precedence).

## Blocking

**1. Plugin-modified headers are silently dropped whenever the A2A agent has no explicit `passthrough_headers` allowlist** (`mcpgateway/services/tool_service.py`, in the A2A branch of `invoke_tool`, around the `if a2a_agent_passthrough_headers:` guard before `headers.update(safe_headers)`).

`a2a_agent.passthrough_headers` defaults to `None` for any agent that hasn't been explicitly configured with an allowlist (`Mapped[Optional[List[str]]] = mapped_column(JSON, nullable=True)` in `db.py`, no default). The base-header computation a few lines above correctly falls back to the global default allowlist when this is `None` (`compute_passthrough_headers_cached`'s `effective_allowed = gateway_passthrough_headers if gateway_passthrough_headers is not None else allowed_headers`). But the post-plugin-hook re-merge treats `None` as "drop everything the plugin returned" instead of applying the same fallback — so for the common case (no agent-specific allowlist configured), any header a `TOOL_PRE_INVOKE` plugin adds or rewrites for an A2A tool call has no effect on the outbound request, with no error or log to indicate it was dropped.

Could you either (a) fall back to the same `effective_allowed` (agent list OR global default) used for the base headers, or (b) if dropping-when-unconfigured is intentional, add a debug/warning log so it's diagnosable, and document the behavior? Right now it reads as an oversight rather than a deliberate choice, and none of the new tests cover `passthrough_headers = None` combined with a plugin modifying headers — that gap is exactly why this slipped through (all four new/updated tests in `TestInvokeToolA2A` set an explicit non-empty allowlist).

## Suggestions (non-blocking)

**2. Consider centralizing the header-safety logic instead of duplicating it inline.** `filter_sensitive_headers` is called three times in the A2A branch, and the allowlist intersection (`a2a_allowlist = {h.lower() for h in a2a_agent_passthrough_headers}` + dict comprehension) reimplements a pattern `compute_passthrough_headers_cached` already has for the same field, just structured differently. This duplication is what let the `None`-handling diverge between the two call sites (finding #1). A shared helper in `passthrough_headers.py` for "compute headers safe for a plugin" / "re-filter plugin-returned headers against the same allowlist" would remove this class of bug for future integrations, not just this one.

**3. Minor efficiency nit:** when `enable_sensitive_header_passthrough` is `False` (the default), `headers` is filtered via `filter_sensitive_headers` and then `plugin_headers = filter_sensitive_headers(headers)` re-filters the same already-clean dict. Cheap, but avoidable: `plugin_headers = headers if not settings.enable_sensitive_header_passthrough else filter_sensitive_headers(headers)`.

**4. Docs:** the new "A2A Agent Override" section in `docs/docs/overview/passthrough.md` doesn't distinguish `passthrough_headers: None` (falls back to the global default allowlist) from `passthrough_headers: []` (blocks everything). Worth a one-line clarification, plus a note that plugin header modifications currently only take effect when the agent has a non-empty allowlist (ties to finding #1).

**5. Coverage:** all A2A header-passthrough testing here is unit-level with `plugin_manager`/`compute_passthrough_headers_cached` mocked out. Might be worth one lighter-weight integration test that exercises the real helpers together, since that's the code path most likely to regress silently (as in finding #1).

## CI

The failing `pytest (py3.12)` check isn't caused by this branch — all 79 failures are `AttributeError: '_IncludedRouter' object has no attribute 'path'` in `test_api_v1.py`, `test_legacy_router.py`, `test_well_known.py`, and `test_api_versioning_parity.py`, none of which this PR touches. It looks like a repo-wide regression from a FastAPI 0.137 upgrade, already being fixed separately in #5447 and #5449. No action needed on your end — please don't spend time trying to fix these test failures, they're unrelated to this PR. Once #5447 or #5449 merges into `main`, a rebase of this branch will clear them up automatically.

jstar0:
Thanks for the detailed review. I rebased onto current `main` and pushed a follow-up fix for the blocking `passthrough_headers=None` case.

The A2A tool path now uses the same effective allowlist semantics for plugin-returned headers as it uses for base request passthrough: an unset agent override falls back to the global allowlist, while an explicit empty list still blocks plugin-returned passthrough headers. I added a regression test for the unset-agent case and reran the focused A2A header tests, adjacent A2A coverage, targeted Ruff, detect-secrets hook, and `git diff --check`.

msureshkumar88 (changes requested):
Thanks for tracking this through the previous review round — the `None`-fallback fix for plugin-modified headers (commit 837515c4) correctly addresses the earlier blocking comment, and the `api_key` Authorization-precedence fix in `a2a_protocol.py` is a solid, well-tested catch on its own.

## Blocking

**1. The new "no override → fall back to the global default allowlist" behavior appears to contradict the existing `a2a_service.py` implementation of the same feature for the same resource.**

`mcpgateway/services/a2a_service.py::_prepare_header_flows` (used by the direct `/a2a` agent-invocation path) treats `agent.passthrough_headers is None` as default-deny:

```python
if not request_headers or not agent_passthrough_headers:
    return {}, {}
```

No headers reach the plugin or the downstream agent in that case — `_refilter_plugin_headers` has the same "no whitelist configured → block all" logic. There's no fallback to `DEFAULT_PASSTHROUGH_HEADERS`/`GlobalConfig` anywhere in that path.

This PR's `tool_service.py` change (`a2a_effective_passthrough_headers = a2a_agent_passthrough_headers if ... is not None else passthrough_allowed`, `:5880`) does the opposite for the same underlying A2A agent record: when `passthrough_headers` is unset, it now falls back to the *global* allowlist, so headers like `X-Tenant-Id` do reach both the plugin and the downstream agent.

Net effect: the same A2A agent, with `passthrough_headers` left unset (the common case — the PR description itself frames this as "leave `passthrough_headers` unset to use the global default allowlist"), exposes a different header surface depending on whether it's called via `/a2a` direct invoke (blocked) or via `tools/call` on the auto-generated tool (global allowlist forwarded).

Could you either (a) reconcile the two paths so "no override" means the same thing in both places, or (b) if the divergence is intentional, call it out explicitly in the PR description/docs so it's a documented decision rather than something that falls out of two independently-written implementations? Given the previous round's blocking issue was exactly this kind of divergence between two call sites of the same allowlist logic, I'd lean toward reconciling rather than documenting the difference.

## Suggestions (non-blocking, mostly carried over from the last round)

**2. Header-safety logic is now duplicated in three places** — `a2a_service.py::_prepare_header_flows`/`_refilter_plugin_headers`, and this PR's new inline logic in `tool_service.py`. This is the direct cause of both the original blocking bug and finding #1 above. A shared helper in `passthrough_headers.py` (or a common module both services call) would remove this whole class of bug going forward, not just patch today's instance.

**3. Docs still don't distinguish `passthrough_headers: None` from `passthrough_headers: []`.** The "A2A Agent Override" section in `docs/docs/overview/passthrough.md` describes the `None` fallback in prose but doesn't contrast it with the explicit-empty-list "block everything" behavior — worth a one-line clarification, especially now that the two behave differently depending on which invocation path is used (see #1).

**4. Test-double gap:** `_make_a2a_agent()` in `test_tool_service_coverage.py` uses a bare `MagicMock()` and doesn't default `.passthrough_headers`, so any test that doesn't explicitly set it gets a truthy `MagicMock` (empty on iteration) rather than the real DB default of `None`. Doesn't break current assertions, but it means the fixture doesn't accurately model the "no override configured" case this PR is centered on. Worth adding `passthrough_headers=None` as an explicit default parameter.

**5. Coverage:** still unit-only with `plugin_manager`/`compute_passthrough_headers_cached` mocked out, as noted last round. A lighter integration test exercising the real helpers — ideally covering both `a2a_service.py` and `tool_service.py` against the same agent config — would likely have caught #1 directly.

No action needed on CI — only the DCO check has run against the latest commit, which is expected for an external-contributor PR pending workflow approval.

jstar0:
Addressed the blocking path mismatch in `0294ce4ae`.

The generated A2A tool path now matches the direct A2A service contract: `passthrough_headers=None` and `passthrough_headers=[]` both block request-header forwarding. This applies to the base request headers shown to `TOOL_PRE_INVOKE` and to plugin-returned headers before the downstream request is prepared.

I also made the A2A test helper default `passthrough_headers` to the real ORM default (`None`) and updated the A2A, plugin lifecycle, passthrough, and configuration docs to state the default-deny behavior explicitly.

Verification on the current head:

- unset-allowlist regression: 1 passed
- related A2A coverage: 262 passed, 1 skipped
- targeted Ruff check and format: passed
- detect-secrets hook: passed
- `git diff --check`: passed

msureshkumar88:
Re-reviewed after the latest commit (`0294ce4a`, "align unset passthrough headers").

## Round 2 blocking issue — resolved

The cross-path inconsistency is fixed. `a2a_allowlist = a2a_agent_passthrough_headers or []` now passes an explicit `[]` (not `None`) into `compute_passthrough_headers_cached`, so `gateway_passthrough_headers is not None` is true and `effective_allowed` becomes `[]` — no fallback to the global allowlist. That makes the `tools/call` path default-deny on an unset/empty agent allowlist, matching `a2a_service.py::_prepare_header_flows` exactly. Traced this end-to-end against `passthrough_headers.py:422` to confirm.

Also confirmed:
- Docs now explicitly distinguish `None` vs `[]` (both block forwarding) across all four updated files.
- `_make_a2a_agent()` test fixture now defaults `passthrough_headers = None`, matching the real DB default.
- New regression test (`test_a2a_pre_invoke_blocks_headers_when_agent_allowlist_unset`) covers exactly this scenario.

## Non-blocking, carried from round 2 (still open, fine as follow-up)

- Header-safety/allowlist logic is still duplicated between `a2a_service.py` and `tool_service.py`. This duplication is the root cause of both blocking rounds so far — worth a follow-up issue to centralize into a shared helper rather than blocking this PR further on it.
- No integration test exercises `/a2a` direct invoke and `tools/call` together against the same agent config to guard the invariant this PR just restored.
- No debug/info log when headers are dropped due to an unset/empty allowlist — would help diagnose future divergence quickly.

## Status

Both blocking rounds are resolved and verified. LGTM from my side — no changes required.

msureshkumar88 (approved):
Ran a real end-to-end verification of this fix against commit `0294ce4ae` — live gateway process, a real HTTP echo server standing in for the downstream A2A agent, and a small throwaway `TOOL_PRE_INVOKE` probe plugin logging exactly what headers reach it. No mocks, no unit-test shortcuts.

## Config used

`.env` derived from `.env.example`: `AUTH_REQUIRED=true`, `MCPGATEWAY_A2A_ENABLED=true`, `ENABLE_HEADER_PASSTHROUGH=true`, `ENABLE_SENSITIVE_HEADER_PASSTHROUGH` toggled `false`/`true` per case, `DEFAULT_PASSTHROUGH_HEADERS=["X-Tenant-Id","X-Trace-Id","X-Blocked","Authorization"]` (deliberately includes headers that must *not* leak to A2A agents, to prove no fallback), `PLUGINS_ENABLED=true` with a trimmed plugin config pointing only at the probe plugin, SQLite DB, real generated `JWT_SECRET_KEY`.

## Commands run

```
git checkout 0294ce4ae
make venv install-dev
cp .env.example .env   # + overrides above
python3 -m mcpgateway.bootstrap_db
python3 scripts/e2e_verify_pr4925.py
```
The script: starts the echo server, starts the real gateway, generates a JWT bearer token, registers A2A agents for real via `POST /a2a`, invokes the generated tool via real `POST /rpc` `tools/call` with real request headers, restarts the gateway between cases that change startup-level settings, and asserts plugin-visible vs. downstream-visible headers per case.

## Output

```
[PASS] case1a: plugin_saw={'Content-Type': 'application/json', 'X-Tenant-Id': 'tenant-42'} downstream_saw_tenant=tenant-42 downstream_had_blocked=False
[PASS] case1b: plugin_saw_keys=['Content-Type', 'X-Tenant-Id']   # Authorization allowlisted but never reaches plugin
[PASS] case2:  downstream_authorization=<configured api_key> (not the client's token) plugin_saw_keys=['Content-Type', 'X-Tenant-Id']
[PASS] case3-null:  leaked_downstream=[] leaked_plugin=[]   # passthrough_headers unset -> no fallback to DEFAULT_PASSTHROUGH_HEADERS
[PASS] case3-empty: leaked_downstream=[] leaked_plugin=[]   # passthrough_headers=[] -> same as unset
[PASS] case4-flag-false: downstream_had_auth=False plugin_had_auth=False
[PASS] case4-flag-true:  downstream_had_auth=True  plugin_had_auth=False
[PASS] case5: plugin-injected disallowed header stripped before outbound call
OVERALL: PASS
```

## Verdict

All 5 acceptance-criteria cases (8 sub-assertions) pass against a live instance:
- Allowlisted non-sensitive headers reach the `TOOL_PRE_INVOKE` plugin; non-allowlisted ones don't; sensitive headers never reach the plugin regardless of allowlist.
- Configured `api_key` auth always wins over a client-supplied `Authorization` header on the outbound call.
- Unset (`null`) and empty (`[]`) `passthrough_headers` both fully block forwarding — no fallback to the global default allowlist, confirming the cross-path fix in the branch's HEAD commit.
- `ENABLE_SENSITIVE_HEADER_PASSTHROUGH` correctly gates sensitive headers reaching the downstream agent, and never affects plugin visibility.
- Headers a plugin injects outside the allowlist are re-filtered out before the outbound call.

No production code was modified during this verification — read-and-run only, in an isolated worktree.

One non-blocking observation from the trace: `ToolPreInvokePayload.headers` is marked deprecated in the currently-published `cpex` package in favor of `extensions.http.headers`, and `cpex` has no upper version pin in `pyproject.toml`. Doesn't affect correctness today (confirmed headers reach the plugin correctly on the pinned version in use), but worth a follow-up so this doesn't silently break on a future `cpex` upgrade.

Approving based on this result — validation passed cleanly against a real running instance.

msureshkumar88 (changes requested):
CI `pytest (py3.12)` is failing on this branch — not a coverage-threshold issue, two tests error out:

```
FAILED tests/unit/mcpgateway/services/test_tool_service_coverage.py::TestInvokeToolA2A::test_a2a_pre_invoke_modified_headers_are_refiltered - mcpgateway.services.tool_service.ToolInvocationError: Tool invocation failed: 'types.SimpleNamespace' object has no attribute 'metadata'
FAILED tests/unit/mcpgateway/services/test_tool_service_coverage.py::TestInvokeToolA2A::test_a2a_pre_invoke_blocks_headers_when_agent_allowlist_unset - mcpgateway.services.tool_service.ToolInvocationError: Tool invocation failed: 'types.SimpleNamespace' object has no attribute 'metadata'
```

## Root cause

`mcpgateway/services/tool_service.py:5806` (in the A2A branch of `invoke_tool`, pulled in by the `Merge origin/main into fix/a2a-tool-header-passthrough` commit) now calls:

```python
record_plugin_metrics(current_trace_id.get(), pre_result.metadata)
```

right after the `TOOL_PRE_INVOKE` hook returns. Both of these two tests mock the hook's return value as a bare `SimpleNamespace(modified_payload=modified_payload)`:

```python
plugin_manager.invoke_hook = AsyncMock(return_value=(SimpleNamespace(modified_payload=modified_payload), {}))
```

— which has no `.metadata` field, so `pre_result.metadata` raises `AttributeError`, caught by the generic exception handler and re-raised as `ToolInvocationError`.

Every other passing `TestInvokeToolA2A` test in this file (e.g. `test_a2a_pre_invoke_payload_receives_passthrough_headers`, `test_a2a_pre_invoke_blocks_sensitive_headers_by_default`) uses the real `cpex.framework.PluginResult(modified_payload=..., continue_processing=True)`, which has a `metadata` field defaulting to `{}` — that's why only these two `SimpleNamespace`-based mocks trip on the newly-merged `record_plugin_metrics` call.

Reproduced locally: passes in isolation and even for the full file, but confirmed by reading the actual CI traceback (not just the summary line) that the crash is at the `pre_result.metadata` access, not `global_context.metadata` as the surrounding source context might suggest at a glance.

## Suggested fix

Give the mocked `pre_result` a `.metadata` attribute in both tests, e.g.:

```python
plugin_manager.invoke_hook = AsyncMock(return_value=(SimpleNamespace(modified_payload=modified_payload, metadata=None), {}))
```

(or swap to the real `PluginResult` class like the neighboring tests do, for consistency).

Requesting changes so CI is green before merge — the underlying header-passthrough fix itself still looks correct, this is purely a test-mock gap against a shape that changed after the `main` merge landed here.

jstar0:
Addressed in `ad0c494a` after merging current `main`.

Both failing tests now return the real `cpex.framework.PluginResult` instead of an incomplete `SimpleNamespace`, so the mock includes the `metadata` field consumed by `record_plugin_metrics`. No production behavior changed.

Verification on the merged head:

- the two failing tests: passed
- full `TestInvokeToolA2A`: 23 passed
- related A2A selection: passed with one tracked skip
- Ruff check/format, detect-secrets, full pre-commit, and `git diff --check`: passed

msureshkumar88 (approved):
All blocking items across the prior rounds are now resolved — approving.

**Round 1 (plugin-modified headers dropped when `passthrough_headers` is `None`)** and **Round 2 (`None`-handling divergence between `tool_service` and `a2a_service`)** are reconciled by `0294ce4ae`: the generated-tool path now treats both unset (`None`) and empty (`[]`) allowlists as default-deny (`a2a_allowlist = a2a_agent_passthrough_headers or []`), matching `a2a_service._prepare_header_flows`. The two invocation paths now expose the same header surface for the same agent record. This was independently confirmed by the live e2e run (case3-null and case3-empty both `leaked_downstream=[] leaked_plugin=[]`).

**CI mock gap (`pre_result.metadata` AttributeError)** is fixed by `ad0c494a7`, swapping the two `SimpleNamespace(modified_payload=...)` mocks for the real `PluginResult`, which carries a `.metadata` default — the fix suggested in the last round. CI is now fully green (30 checks passing, 0 failing), and the branch is mergeable.

Production logic is unchanged since the earlier end-to-end verification (HEAD only touched the test file plus a `main` merge), so that validation still holds.

The non-blocking suggestions remain reasonable follow-ups (shared header-safety helper in `passthrough_headers.py`, a light integration test exercising the real helpers across both paths, and the `cpex` deprecated-`headers` / missing upper pin note) but none block this merge. Nice work tracking this through several rounds.

ja8zyjits (approved):
LGTM

ja8zyjits:
Thanks @jstar0 , thanks for the patience and persistence.
This pr is merged now.

PR #4868, opened 2026-05-22
fix(auth): allow cookie auth for same-origin OAuth callback fetch requests

bogdanmariusc10:
## 🔗 Related Issue
Closes #4867

---

## 📝 Summary

OAuth-configured MCP servers (ClickUp, Gamma, Granola, Notion) failed to fetch tools after a successful OAuth authorization flow. The "Fetch Tools from MCP Server" button on the OAuth callback success page returned:

`Cookie authentication not allowed for API requests. Use Authorization header.`

**Root cause:** The browser detection logic in `rbac.py`, `auth_middleware.py`, and `admin.py` identified a request as a legitimate browser/UI request if the `Referer` header contained `/admin`. However, the OAuth callback page (`/oauth/callback`) makes a `fetch()` call with `Accept: application/json`, which matched none of the browser signals, causing the security middleware introduced in #2680 to incorrectly reject it as an external API request.

**Fix:** Replaced the loose `"/admin" in referer` substring check with a structured same-origin referer validation using `urlparse`. The check now verifies that the referer's host matches the request host **and** that its path contains either `/admin` or `/oauth/callback`. This correctly identifies OAuth callback fetch requests as same-origin browser requests while preserving the CSRF protection for cross-origin requests.

The fix is applied consistently across all three locations where browser detection logic lives:
- `mcpgateway/middleware/rbac.py`
- `mcpgateway/middleware/auth_middleware.py`
- `mcpgateway/admin.py`

---

## 🏷️ Type of Change
- [x] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Chore (deps, CI, tooling)
- [ ] Other (describe below)

---

## 🧪 Verification
| Check                     | Command         | Status |
|---------------------------|-----------------|--------|
| Lint suite                | `make lint`     |  ✅ Pass      |
| Unit tests                | `make test`     |     ✅ Pass   |
| Coverage ≥ 80%            | `make coverage` |    ✅ Pass    |

---

## ✅ Checklist
- [x] Code formatted (`make black isort pre-commit`)
- [x] Tests added/updated for changes
- [x] Documentation updated (if applicable)
- [x] No secrets or credentials committed

---

## 📓 Notes (optional)

**Tests added:**
- `test_oauth_callback_referer_allows_browser_continuation` — verifies OAuth callback referer is treated as a same-origin browser request and continues to `call_next` instead of hard-denying
[...]

PR #4931, opened 2026-05-28
fix(rbac): allow cookie auth for /oauth/fetch-tools admin action

### LINKED ISSUE #4930: [BUG]: /oauth/fetch-tools/{id} returns 401 "Cookie auth not allowed" — referer-based browser-request check misses /oauth/callback (v1.0.2)
### 🐞 Bug Summary

After completing the OAuth Authorization flow for a gateway from the admin UI, clicking the **"Fetch Tools from MCP Server"** button on the success page (`/oauth/callback`) fails with:

POST /oauth/fetch-tools/{gateway_id}  -> 401 Unauthorized
 {"detail":"Cookie authentication not allowed for API requests. Use Authorization header."}

Root cause: the cookie-auth-rejection in `mcpgateway/middleware/rbac.py` only marks a request as "browser" when the `Referer` contains `/admin`. Fetch-Tools button is rendered on `/oauth/callback`, so the Referer is `/oauth/callback?...` (not `/admin`), the request is classified as an "API request", and the `jwt_token` cookie is rejected — even though it's clearly an admin-UI initiated call.

---

### 🧩 Affected Component

- [x] `mcpgateway` - API
- [ ] Other (explain below)

---

### 🔁 Steps to Reproduce

1. Register a gateway with OAuth (`oauth_enabled=true`) via the admin UI.
2. From `/admin/`, click **"Authorize OAuth"** for that gateway. Complete the IdP flow.
3. The browser lands on `/oauth/callback?code=...&state=...`, which renders an "OAuth Authorization Successful" page with a **"Fetch Tools from MCP Server"** button.
4. Click the button → `POST /oauth/fetch-tools/{gateway_id}` is sent with the `jwt_token` cookie (no `Authorization: Bearer` header).
5. Response: **401 Unauthorized** — `{"detail":"Cookie authentication not allowed for API requests. Use Authorization header."}`.

---

### 🤔 Expected Behavior

The "Fetch Tools" button on the OAuth callback success page should succeed — it's an admin-UI initiated call from a logged-in admin session.

---

### 📓 Logs / Error Output

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Version or commit | `v1.0.2` |
| Runtime | `Python 3.12, Gunicorn + Uvicorn worker |
| Platform / OS | `Linux arm64` |
| Container | `Docker` |

---

kimsehwan96:
## 🔗 Related Issue
Closes #4930

---

## 📝 Summary
After completing the OAuth Authorization flow for a gateway from the admin UI, clicking the **"Fetch Tools from MCP Server"** button on the success page (`/oauth/callback`) fails with `401 Unauthorized — "Cookie authentication not allowed for API requests"`.

---

## 🏷️ Type of Change
- [x] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

| Coverage ≥ 80%            | `make coverage` |   pass     |

---

## ✅ Checklist

---

## 📓 Notes (optional)

`mcpgateway/middleware/rbac.py` classified a request as "admin-UI" only when `Referer` contains `/admin`. The Fetch Tools button is rendered on `/oauth/callback`, so its Referer is `/oauth/callback?...` (not `/admin`), and the JWT cookie was rejected as if the call came from an external API client.

### Fix 
Also allow the request when the **server-side request path** starts with `/oauth/fetch-tools`

### DIFF
--- a/mcpgateway/middleware/rbac.py
+++ b/mcpgateway/middleware/rbac.py
@@ -351,7 +351,7 @@ async def get_current_user_with_permissions(request: Request, credentials: Optio
     accept_header = request.headers.get("accept", "")
     is_htmx = request.headers.get("hx-request") == "true"
     referer = request.headers.get("referer", "")
-    is_admin_ui_request = "/admin" in referer
+    is_admin_ui_request = "/admin" in referer or str(request.url.path).startswith("/oauth/fetch-tools")
     is_browser_request = "text/html" in accept_header or is_htmx or is_admin_ui_request
 
     # SECURITY: Reject cookie-only authentication for API requests
--- a/tests/unit/mcpgateway/middleware/test_rbac.py
+++ b/tests/unit/mcpgateway/middleware/test_rbac.py
@@ -99,6 +99,7 @@ async def test_get_current_user_with_permissions_cookie_rejected_for_api_request
     mock_request = MagicMock(spec=Request)
     mock_request.cookies = {"jwt_token": "token123"}
     mock_request.headers = {"user-agent": "python-requests/2.31", "accept": "application/json"}
+    mock_request.url.path = "/tools"
     mock_request.client = MagicMock()
     mock_request.client.host = "127.0.0.1"
     mock_request.state = MagicMock(auth_method="jwt", request_id="req123")
@@ -147,6 +148,7 @@ async def test_cookie_auth_rejected_with_cross_origin_oauth_referer():
     mock_request = MagicMock(spec=Request)
     mock_request.cookies = {"jwt_token": "token123"}
     mock_request.headers = {"accept": "application/json", "referer": "https://attacker.example/oauth/callback"}
+    mock_request.url.path = "/tools"
     mock_request.client = MagicMock()
     mock_request.client.host = "127.0.0.1"
     mock_request.state = MagicMock(auth_method="jwt", request_id="req-xorigin")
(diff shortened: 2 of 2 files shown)


marekdano (changes requested):
@kimsehwan96 - Good catch on the root cause — the Referer-based signal only covers `/admin` paths, so the OAuth callback page falls outside it. The fix works and the tests are a nice addition.

  A couple of things worth considering before merging:

  Suggestion: use `Referer path` instead of `destination path`

  The existing `is_admin_ui_request` logic is signal-based ("where did this request originate?"). The fix flips to destination-based ("where is this request going?"), which is a conceptual mismatch and creates a path allowlist that needs maintaining for future OAuth-callback-style pages.

  A small tweak keeps it signal-based and uses `referer_path` which is already computed a few lines below:

  `is_admin_ui_request = "/admin" in referer or referer_path.startswith("/oauth/")`

Using `referer_path` (parsed URL path only) rather than the raw referer string also avoids the forged-origin edge case (`https://evil.example/oauth/callback wouldn't match`).

Test: `test_cookie_auth_rejected_for_non_fetch_tools_oauth_path` doesn't cover what it claims

 The test passes `jwt_token=None` while the mock cookie has `"token123"`. The 401 it gets is "no token at all", not "cookie rejected because path didn't match". To actually test the boundary, pass `jwt_token="token123"` (same as the positive test) - that way the assertion on `"Cookie authentication not allowed"` is genuinely exercising the path check.

kimsehwan96:
@marekdano Thanks for your feedback. 

I think that #4868 fixes #4930 and it was merged so this PR can be closed without any futher works. 

I will close this PR and issue. Thanks!

PR #4932, opened 2026-05-28
fix(tools): resolve tool cache and state consistency issues during recovery

### LINKED ISSUE #4915: [BUG]: Offline tool cache entries survive gateway recovery and gateway/tool reachability can temporarily diverge
### 🐞 Bug Summary
Tools recovered by gateway refresh/health checks can still appear offline and remain unavailable for invocation because stale negative tool-lookup cache entries are not cleared when reachability is restored.

The fix introduced by PR https://github.com/IBM/mcp-context-forge/pull/4499 in [`_update_or_create_tools()`](mcpgateway/services/gateway_service.py:4690) correctly restores persisted tool state by marking existing tools reachable in [`existing_tool.reachable = True`](mcpgateway/services/gateway_service.py:4774) and creating new tools with [`reachable=True`](mcpgateway/services/gateway_service.py:4676). However, invocation still consults the tool lookup cache, and offline negative entries created during an outage can outlive the recovery path.

When a tool is looked up while offline, [`set_negative(name, "offline")`](mcpgateway/services/tool_service.py:3795) stores a negative cache entry through [`set_negative()`](mcpgateway/cache/tool_lookup_cache.py:287). That negative entry is keyed only by tool name and does not carry gateway association. Later, gateway recovery invalidates cache by gateway using [`invalidate_gateway()`](mcpgateway/cache/tool_lookup_cache.py:343) from the refresh path in [`_refresh_gateway_tools_resources_prompts()`](mcpgateway/services/gateway_service.py:5348), but [`invalidate_gateway()`](mcpgateway/cache/tool_lookup_cache.py:364) only removes entries that contain [`tool.gateway_id`](mcpgateway/cache/tool_lookup_cache.py:366) or were indexed in Redis via [`set(..., gateway_id=...)`](mcpgateway/cache/tool_lookup_cache.py:249). Negative offline entries satisfy neither condition, so they can survive recovery and continue returning “offline” until TTL expiry.

There is also a gateway/tool state consistency concern in the recovery path. Gateway and tool reachability are intended to cascade together through [`set_gateway_state()`](mcpgateway/services/gateway_service.py:2952), which bulk-updates tool reachability via [`update(DbTool)...values(reachable=reachable)`](mcpgateway/services/gateway_service.py:3186). However, the method commits gateway state first at [`db.commit()`](mcpgateway/services/gateway_service.py:3163) and only then performs the bulk tool update with a second commit at [`mcpgateway/services/gateway_service.py:3198`](mcpgateway/services/gateway_service.py:3198). That creates a transient window where a gateway can be observed as reachable while some tools still retain stale reachability flags. In addition, health-check recovery marks the gateway reachable in [`set_gateway_state(... reachable=True, only_update_reachable=True)`](mcpgateway/services/gateway_service.py:3899) before the refresh path runs, and the subsequent reconciliation of tool metadata depends on auto-refresh settings and lock availability in [`_refresh_gateway_tools_resources_prompts()`](mcpgateway/services/gateway_service.py:5116). So even though the intended model is “gateway and tools recover together,” the implementation still allows temporary or effective divergence between gateway online state and tool online state.

---

### 🧩 Affected Component

- [x] `mcpgateway` - API
- [x] Other (cache invalidation / tool lookup caching)

---

### 🔁 Steps to Reproduce

1. Configure a gateway with MCP tools and ensure tool lookup cache is enabled via [`tool_lookup_cache_enabled`](mcpgateway/config.py:2007).
2. Put the gateway or upstream MCP server into a temporary failure state so the tool becomes unreachable.
3. Invoke the tool through the normal lookup path so [`set_negative(name, "offline")`](mcpgateway/services/tool_service.py:3795) caches the offline result.
4. Restore the upstream gateway/tool so health checks or manual refresh succeed.
5. Let gateway refresh run through [`_refresh_gateway_tools_resources_prompts()`](mcpgateway/services/gateway_service.py:5008), which restores DB reachability in [`_update_or_create_tools()`](mcpgateway/services/gateway_service.py:4774).
6. Invoke the same tool again before negative cache TTL expiry.
7. Observe that the tool can still be reported offline even though the database state has been restored and the gateway is healthy.

To observe the gateway/tool inconsistency window specifically:

8. Trigger a health-check recovery path where the gateway transitions from unreachable to reachable through [`set_gateway_state(... only_update_reachable=True)`](mcpgateway/services/gateway_service.py:3899).
9. Inspect gateway state and tool state between the first commit in [`mcpgateway/services/gateway_service.py:3163`](mcpgateway/services/gateway_service.py:3163) and the later bulk tool update/commit in [`mcpgateway/services/gateway_service.py:3198`](mcpgateway/services/gateway_service.py:3198), or observe behavior when auto-refresh is skipped/throttled/lock-blocked.
10. Observe that the gateway may already be online while tool reachability or effective tool availability still appears offline.

---

### 🤔 Expected Behavior
Once a gateway refresh or health check successfully rediscovers a tool, the tool should become immediately invokable.

Specifically:
- persisted tool state should be restored to reachable,
- any stale offline cache entry for that tool should be invalidated as part of the same recovery flow, and
- gateway and tool online/offline state should transition atomically enough that callers do not observe the gateway as online while its tools still appear offline.

Users should not need to wait for cache TTL expiry before recovered tools become available again.

---

### 📓 Logs / Error Output
Representative user-facing error from the invocation path:

```text
ERROR: Tool 'github-com-read-tools-search-repositories' exists but is currently offline.
Please verify if it is running.
```

Relevant code path raising the error after cache/lookup resolution:
- [`if not tool.reachable: ... raise ToolNotFoundError(...)`](mcpgateway/services/tool_service.py:3794)
- duplicate invocation path also present at [`mcpgateway/services/tool_service.py:4490`](mcpgateway/services/tool_service.py:4490)

Relevant cache behavior:
- negative offline cache write: [`mcpgateway/cache/tool_lookup_cache.py:304`](mcpgateway/cache/tool_lookup_cache.py:304)
- gateway-level invalidation that misses negative entries: [`mcpgateway/cache/tool_lookup_cache.py:364`](mcpgateway/cache/tool_lookup_cache.py:364)

Relevant gateway/tool consistency behavior:
- gateway recovery path: [`mcpgateway/services/gateway_service.py:3899`](mcpgateway/services/gateway_service.py:3899)
- first commit before tool cascade: [`mcpgateway/services/gateway_service.py:3163`](mcpgateway/services/gateway_service.py:3163)
- later bulk tool reachability update: [`mcpgateway/services/gateway_service.py:3186`](mcpgateway/services/gateway_service.py:3186)

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Version or commit | `main after fix #4499` |
| Runtime | `Python 3.11+, FastAPI / ASGI runtime` |
| Platform / OS | `Observed in dev environment; local investigation performed on macOS` |
| Container | `Unknown from report` |

---

Related context:
- Prior fix: [`fix(tools): restore tool reachability status during gateway refresh`](mcpgateway/services/gateway_service.py:4774)
- Related issue/PR context mentioned by reporter: `#4454`, `#4499`

Why this likely reproduces more in dev than stage:
- negative cache entries have TTL controlled by [`tool_lookup_cache_negative_ttl_seconds`](mcpgateway/config.py:2009)
- behavior depends on whether the tool was looked up during the outage window
- Redis-backed L2 plus local L1 cache can make stale offline state appear intermittent across environments

Likely remediation:
- invalidate per-tool cache entries by name during successful refresh for all restored/updated tools using [`invalidate(name, gateway_id=...)`](mcpgateway/cache/tool_lookup_cache.py:307), or
- include gateway metadata in negative cache payloads so [`invalidate_gateway()`](mcpgateway/cache/tool_lookup_cache.py:343) can remove offline negatives reliably
- reduce or eliminate the two-phase gateway/tool reachability update window in [`set_gateway_state()`](mcpgateway/services/gateway_service.py:2952), so gateway and tool reachability changes are committed consistently

Test gap:
- current tests around [`_update_or_create_tools()`](tests/unit/mcpgateway/services/test_gateway_service_extended.py:655) validate DB update behavior, but there is no regression test proving that a stale offline negative cache entry is cleared after successful gateway recovery
- there also does not appear to be a regression test covering the gateway/tool reachability consistency window during recovery

bogdanmariusc10:
## 🔗 Related Issue
Closes #4915

---

## 📝 Summary

This PR fixes two critical bugs in gateway/tool recovery that caused tools to remain unavailable after successful gateway recovery:

**Bug #1: Negative cache entries survive gateway recovery**
- Stale offline tool cache entries persisted after gateway recovery
- `invalidate_gateway()` only removed entries with gateway_id metadata
- Negative offline entries had no gateway association and survived recovery
- **Fix**: Track tools restored from unreachable→reachable and explicitly invalidate their cache entries by name

**Bug #2: Gateway/tool reachability consistency window**
- Gateway and tool state could temporarily diverge during recovery
- `set_gateway_state()` committed gateway state before bulk tool update
- Created window where gateway appeared online while tools still showed offline
- **Fix**: Reorder operations to perform tool bulk update BEFORE committing gateway state

**Impact**: Tools are now immediately available after gateway recovery, eliminating the need to wait for cache TTL expiry.

---

## 🏷️ Type of Change
- [x] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

| Lint suite                | `make lint`     | ✅ Pass |
| Unit tests                | `make test`     | ✅ Pass (412 tests) |
| Coverage ≥ 80%            | `make coverage` | ✅ Pass |

---

## ✅ Checklist
- [x] Tests added/updated for changes (9 new tests: 5 unit + 4 integration)

---

### Implementation Details

**Core Changes**:
1. Made `_update_or_create_tools()` async to support async cache operations
2. Track restored tools in `restored_tool_names` list (lines 4706, 4774-4778)
3. Invalidate cache for each restored tool after DB commit (lines 4815-4820)
4. Reordered `set_gateway_state()` to update tools before gateway commit (lines 3167-3181)

**Test Compatibility**:
- Updated 50+ test call sites to await async `_update_or_create_tools()`
- Fixed 3 mock patches to use `AsyncMock` instead of `MagicMock`
- Added 2 `with patch()` blocks to mock global `_get_tool_lookup_cache()` in tests

### Test Coverage

**Unit Tests** (5 new in `test_gateway_service.py`):
- `TestNegativeCacheInvalidation` class:
  - `test_negative_cache_invalidated_when_tool_restored`
  - `test_negative_cache_not_invalidated_when_tool_stays_unreachable`
  - `test_multiple_tools_restored_all_caches_invalidated`
- `TestGatewayToolConsistency` class:
  - `test_gateway_and_tools_updated_atomically`
  - `test_tool_update_happens_before_gateway_commit`

**Integration Tests** (4 new in `test_gateway_tool_recovery.py`):
- `test_gateway_and_tools_recover_atomically` - Verifies atomic recovery
- `test_negative_cache_invalidated_on_recovery` - Confirms cache invalidation
- `test_tools_immediately_invokable_after_recovery` - End-to-end invocation test
- `test_health_check_recovery_flow` - Health check integration

### Files Modified
- `mcpgateway/services/gateway_service.py` - Core fixes
- `tests/unit/mcpgateway/services/test_gateway_service.py` - Unit tests + compatibility
- `tests/unit/mcpgateway/services/test_gateway_service_extended.py` - Compatibility
- `tests/unit/mcpgateway/services/test_gateway_service_helpers.py` - Compatibility
- `tests/unit/mcpgateway/services/test_gateway_auto_refresh.py` - Mock fix
- `tests/unit/mcpgateway/test_multi_auth_headers.py` - Mock fix
- `tests/integration/test_gateway_tool_recovery.py` - New integration tests (398 lines)

### Root Cause Analysis

**Bug #1**: The issue was in the cache invalidation strategy. When a tool was looked up while offline, `set_negative(name, "offline")` stored a negative cache entry keyed only by tool name without gateway association. Later, `invalidate_gateway()` only removed entries containing `tool.gateway_id` or indexed via `set(..., gateway_id=...)`, so negative entries survived recovery.

**Bug #2**: The issue was in the transaction ordering. `set_gateway_state()` committed gateway state at line 3163 before performing the bulk tool update at line 3186 with a second commit at line 3198. This created a transient window where the gateway could be observed as reachable while tools still retained stale reachability flags.


a-effort (commented):
Thanks for the careful root-cause split. The two-bug framing and matching test classes made this very easy to follow. A couple of things worth a look before merge.

### Potentially blocking

**1. Cache invalidation runs before `db.commit()`.**
In `_update_or_create_tools` (gateway_service.py:4823-4828), `invalidate(name, gateway_id=...)` is called inside the helper, but callers (`set_gateway_state`, `update_gateway`, `_refresh_gateway_tools_resources_prompts`) don't commit until later. Between invalidation and commit, a concurrent lookup sees the pre-commit state and can repopulate `set_negative(..., "offline")`, partially reopening the bug. Suggest returning `restored_tool_names` and invalidating in each caller *after* its commit, so the rule stays simple: cache invalidations follow commits.

**2. `tools_updated` rowcount may now be zero in the rediscovery path.**
The helper already flips `existing_tool.reachable = True` in the ORM session. SQLAlchemy will auto-flush before the bulk `UPDATE … WHERE reachable != :reachable`, so restored rows are skipped and `tools_result.rowcount` can be 0, which gates the post-commit `invalidate_tools()` + `invalidate_gateway(...)` at line 3204. Worth confirming with a real SQLite session test.

### Suggestions

**3.** "Atomic" applies to gateway + tools only; prompts/resources still commit separately (lines 3210-3225). A one-line clarification in the PR body would help future readers.

**4.** `_update_or_create_resources` / `_prompts` stay sync. Fine if the asymmetry is intentional, but worth a comment so nobody "fixes" it later.

**5.** Per-tool `invalidate()` fans out N Redis publishes; a batched `invalidate_names(...)` would be a cheap follow-up.

### Nits

- Line 4828 uses an f-string in `logger.debug`; rest of file uses lazy formatting.
- `test_gateway_and_tools_committed_atomically` asserts `execution_order[-1] == "commit"`, which works, but an end-to-end test that pre-seeds `set_negative("t", "offline")` and then asserts `await cache.get("t") is None` post-recovery would be a stronger guardrail than the call-count check.

### Looks good

Reordering in `set_gateway_state` is the right structural fix rather than a workaround. Guarding `restored_tool_names` on `not existing_tool.reachable` keeps the invalidation surgical, and `test_already_reachable_tool_no_cache_invalidation` pins that nicely. The async-conversion sweep across the test sites is thorough.

bogdanmariusc10:
Thanks for the review, @a-effort! I've addressed all findings in commit [0be1cc3](https://github.com/your-org/your-repo/commit/0be1cc3cb1b43190dc5265eef738887b199eb4d1).

## Blocking Issues

### ✅ 1. Cache invalidation runs before `db.commit()`

**Fixed.** Moved cache invalidation to happen AFTER commit in all 4 callers:

- **OAuth path** (`fetch_tools_after_oauth`): Lines 1816-1819 - invalidation after commit at line 1804
- **Update path** (`update_gateway`): Lines 2658-2661 - invalidation after commit at line 2624  
- **State change path** (`set_gateway_state`): Lines 3242-3246 - invalidation after commit at line 3215
- **Refresh path** (`_refresh_gateway_tools_resources_prompts`): Lines 5422-5425 - invalidation after commit at line 5391

Changed `_update_or_create_tools` to return `tuple[List[DbTool], List[str]]` where the second element is `restored_tool_names`. The helper no longer performs any cache invalidation - each caller controls the timing.

**Tests added:**
- `test_fetch_tools_after_oauth_invalidates_cache_for_restored_tools`
- `test_update_gateway_invalidates_cache_for_restored_tools`
- `test_set_gateway_state_invalidates_cache_for_restored_tools`

### ✅ 2. `tools_updated` rowcount may be zero

**Fixed.** Added fallback logic at line 3210:

```python
# In the rediscovery path, _update_or_create_tools may have already set
# existing_tool.reachable=True in the ORM session, which auto-flushes before
# the bulk UPDATE. Use restored_tool_names to determine if invalidation is needed.
tools_updated = tools_result.rowcount if not restored_tool_names else len(restored_tool_names)
```

When the helper has already marked tools as reachable in the ORM session, SQLAlchemy auto-flushes before the bulk UPDATE, causing `rowcount` to be 0. The fallback uses `len(restored_tool_names)` in this case to ensure cache invalidation still runs.

## Suggestions

### ✅ 3. Clarify atomic scope

**Added comment** at lines 3213-3214:

```python
# Note: "atomic" applies to gateway + tools only; prompts/resources
# are updated in separate commits below for backward compatibility
```

### ✅ 4. Async asymmetry documentation

**Added comments** to both sync helpers explaining why they remain synchronous:

**`_update_or_create_resources()`** (lines 4872-4873):
```python
NOTE: This method remains synchronous (unlike _update_or_create_tools) because
resources do not require cache invalidation logic. The asymmetry is intentional.
```

**`_update_or_create_prompts()`** (lines 4984-4985):
```python
NOTE: This method remains synchronous (unlike _update_or_create_tools) because
prompts do not require cache invalidation logic. The asymmetry is intentional.
```

### 📝 5. Batched invalidation

Agreed this would be a good follow-up. The current per-tool fan-out works correctly and the optimization can be done separately without changing the fix's correctness.

## Nits

### ✅ Lazy logging

**Fixed** at all 4 call sites. Changed from:
```python
logger.debug(f"Invalidated cache for {len(restored_tool_names)} restored tools: {restored_tool_names}")
```

To lazy formatting:
```python
logger.debug("Invalidated cache for %s restored tools: %s", len(restored_tool_names), restored_tool_names)
```

Lines: 1819, 2661, 3246, 5425

### 📝 End-to-end cache test

Good suggestion for a follow-up. The current tests verify the invalidation calls happen at the right time with the right arguments. An e2e test with actual Redis would be valuable but is orthogonal to fixing the race condition.

---

## Test Coverage

Added 6 new tests covering all previously untested lines:

- Cache invalidation in all 4 paths (OAuth, update, state, refresh)
- Chunked insertion for large batches (>1000 resources/prompts)  
- Early return for None tools list

**All tests pass**: 328 passed, 1 skipped

Updated 30+ existing test call sites to handle the new `(tools_to_add, restored_tool_names)` tuple return value.

a-effort (commented):
Thanks for the follow-up — post-commit ordering, the `tools_updated` fallback, and the async-asymmetry comments all landed cleanly. One new blocker while verifying.

### Potentially blocking

**1. Wrong cache key.**
`_update_or_create_tools:4827` appends `tool.name` (raw upstream) to `restored_tool_names`. But the negative cache is keyed by the composite lookup name — `tool_service.py:3869` calls `set_negative(name, "offline")` with the invoker-facing name, which resolves through `DbTool.name` as `f"{gateway_slug}{sep}{custom_name_slug}"` (`db.py:6697-6701` sets this on every insert/update). So `invalidate("get_weather", ...)` walks past the entry stored under `"my-gateway---get_weather"` and the offline record survives — exactly the symptom Bug #1 was meant to fix.

Fix: `restored_tool_names.append(existing_tool.name)`.

**2. The new test doesn't catch (1).**
`test_negative_cache_invalidated_on_recovery` asserts `"test-tool-3" in invalidate_calls`, which only passes because the raw upstream name happens to equal the composite name in this fixture. Either assert against `existing_tool.name` post-listener, or pre-seed `await cache.set_negative(composite_name, "offline")` against a real `ToolLookupCache` and assert `await cache.get(composite_name) is None` after recovery. Given (1), the e2e form is now load-bearing.

ja8zyjits (changes requested):
## 🚨 Blocking Issues

### 1. Race Condition in Tool Update Ordering
- **File**: `mcpgateway/services/gateway_service.py:3167-3181`
- **Issue**: Tool bulk update happens AFTER gateway commit (line 3628), creating window where gateway.reachable=True but tools have old reachability flags
- **Current flow**: Gateway commit → Tool bulk update → Second commit
- **Impact**: Temporary inconsistency between gateway and tool reachability states
- **Fix**: Move tool bulk update BEFORE gateway commit for atomic state transition
- **Note**: PR description claims this is fixed, but code shows tools still updated after gateway commit

### 2. Missing Null Check for `catalog_sync`
- **Files**: 
  - `mcpgateway/services/gateway_service.py:2309-2312`
  - `mcpgateway/services/gateway_service.py:3171-3174`
  - `mcpgateway/services/gateway_service.py:3631-3634`
  - `mcpgateway/services/gateway_service.py:6326-6329`
- **Issue**: `catalog_sync` initialized as `None` (line 3543), then accessed without null check
- **Current**: `if catalog_sync.restored_tool_names:` (assumes catalog_sync is not None)
- **Risk**: AttributeError if initialization path skipped or fails
- **Fix**: Add null check: `if catalog_sync and catalog_sync.restored_tool_names:`

---

## ⚠️ Warnings

### 1. ADR-033 Missing Cache Invalidation Documentation
- **File**: `docs/docs/architecture/adr/033-tool-lookup-cache.md`
- **Code**: `mcpgateway/services/gateway_service.py:2310-2312, 3174-3176, 3634-3636, 6329-6331`
- **Issue**: ADR-033 describes negative caching but doesn't mention explicit invalidation during gateway recovery
- **Fix**: Add "Cache Invalidation on Recovery" section documenting:
  - Negative cache entries invalidated when tools transition unreachable→reachable
  - Happens automatically during gateway recovery/refresh
  - Ensures tools immediately invokable after recovery without TTL wait
  - Recovery paths: `fetch_tools_after_oauth`, `update_gateway`, `set_gateway_state`, `_refresh_gateway_tools_resources_prompts`

### 2. ADR-009 Missing Cache Invalidation Behavior
- **File**: `docs/docs/architecture/adr/009-built-in-health-checks.md`
- **Code**: `mcpgateway/services/gateway_service.py:3634-3636`
- **Issue**: ADR-009 describes health check recovery but doesn't mention cache invalidation
- **Fix**: Add note in "Consequences" section: "Tool cache entries automatically invalidated when gateways recover, ensuring tools immediately available"

### 3. No Error Handling Around Cache Invalidation
- **Files**: All 4 cache invalidation sites
- **Issue**: Loop calls `await tool_lookup_cache.invalidate()` without try-catch
- **Risk**: Single cache failure aborts entire invalidation, leaving partial state
- **Fix**: Wrap in try-except:
  ```python
  for tool_name in catalog_sync.restored_tool_names:
      try:
          await tool_lookup_cache.invalidate(tool_name, gateway_id=gateway.id)
      except Exception as e:
          logger.warning(f"Failed to invalidate cache for tool {tool_name}: {e}")
  ```

---

## 💡 Suggestions

### 1. DRY Violation: Cache Invalidation Duplicated 4x
- **Files**: `gateway_service.py:2309-2312, 3171-3174, 3631-3634, 6326-6329`
- **Issue**: 15 lines of identical cache invalidation logic repeated
- **Fix**: Extract to helper method `_invalidate_restored_tool_caches(catalog_sync, gateway_id)`
- **Benefit**: Single source of truth, easier maintenance

### 2. KISS: Verbose Mock Tool Creation
- **File**: `tests/integration/test_gateway_tool_recovery.py:200-210, 290-300, 380-390, 450-460`
- **Issue**: 10-line MockTool class repeated 4 times
- **Fix**: Use `SimpleNamespace` or extract to fixture
- **Benefit**: More concise, less repetition

### 3. Missing Type Hint for `catalog_sync`
- **File**: `mcpgateway/services/gateway_service.py:3543`
- **Current**: `catalog_sync = None`
- **Better**: `catalog_sync: Optional[GatewayCatalogSyncResult] = None`
- **Benefit**: Explicit type for IDE support and type checking

### 4. Bare Exception Catch Too Broad
- **File**: `mcpgateway/services/gateway_service.py:5750-5753`
- **Current**: `except Exception as e:`
- **Issue**: Catches KeyboardInterrupt, SystemExit
- **Fix**: Catch specific exceptions: `except (ValidationError, ValueError, KeyError) as e:`

---

bogdanmariusc10:
Thanks, @ja8zyjits! All review findings have been implemented in commit `3c37b7808`:

## Blocking Issues ✅
- Fixed race condition by moving tool bulk update before gateway commit
- Added null checks for `catalog_sync` in all 4 locations

## Warnings ✅
- Documented cache invalidation in ADR-033 and ADR-009
- Added error handling around cache invalidation loops

## Suggestions ✅
- Extracted cache invalidation to DRY helper method `_invalidate_restored_tool_caches()`
- Replaced bare Exception catch with specific exceptions
- Simplified MockTool creation using SimpleNamespace
- Added type hints for `catalog_sync`

Ready for re-review.

PR #4937, opened 2026-05-28
Chore: add diagnostics for tool pre-invoke modified payloads

### LINKED ISSUE #4901: Chore: add diagnostics for tool pre-invoke modified payloads
## Summary
Add sanitized debug logging around tool_pre_invoke hook execution so gateway operators can tell whether a plugin returned a CPEX modified_payload and which argument/header keys changed.

## Context
While investigating WXOConnections behavior, the gateway logs showed that the plugin hook ran, but it was not obvious whether the plugin returned a modified_payload in the expected CPEX structure. Key-only diagnostics would make this clear without exposing token or header values.

## Acceptance criteria
- Log whether tool_pre_invoke returned modified_payload.
- Log only argument/header keys, never values.
- Include added/removed key summaries for args and headers.
- Cover the helper/call path with unit tests.

vishu-bh:
# ✨ Feature / Enhancement PR

## 🔗 Epic / Issue
Closes #4901

---

## 🚀 Summary (1-2 sentences)
Adds sanitized DEBUG diagnostics after `tool_pre_invoke` hook execution so operators can see whether CPEX returned a `modified_payload` and which argument/header keys changed. This helps verify plugin payload structure without logging argument, token, or header values.

---

## 🧪 Checks

- [x] `.venv/bin/ruff check mcpgateway/services/tool_service.py tests/unit/mcpgateway/services/test_tool_pre_invoke_logging.py` passes
- [x] `.venv/bin/pytest tests/unit/mcpgateway/services/test_tool_pre_invoke_logging.py -q` passes
- [ ] CHANGELOG updated (not user-facing)

---

## 📓 Notes (optional)
This PR logs only key names and key diffs:

- whether `modified_payload` was returned
- argument/header keys before and after
- added/removed argument/header keys
- modified tool name when present

The logging is wired into the Rust direct-plan pre-invoke path and the Python REST/MCP/A2A invocation pre-invoke paths.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-05-29T09:34:55Z",
+  "generated_at": "2026-05-29T14:08:57Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-05-29T14:08:57Z",
+  "generated_at": "2026-06-02T10:58:43Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -1794,7 +1794,7 @@
         "hashed_secret": "fa9beb99e4029ad5a6615399e7bbae21356086b3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 142,
+        "line_number": 146,
         "type": "Secret Keyword",
         "verified_result": null
       }
@@ -2828,7 +2828,7 @@
         "hashed_secret": "4a0a2df96d4c9a13a282268cab33ac4b8cbb2c72",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 316,
+        "line_number": 320,
         "type": "Secret Keyword",
         "verified_result": null
       }
@@ -4348,7 +4348,7 @@
         "hashed_secret": "3ee08a29a2ca26171fe152b8741d0c90b598263a",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 15099,
+        "line_number": 14971,
         "type": "Secret Keyword",
         "verified_result": null
       }
@@ -4392,7 +4392,7 @@
         "hashed_secret": "d4c3d66fd0c38547a3c7a4c6bdc29c36911bc030",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 667,
+        "line_number": 665,
         "type": "Secret Keyword",
         "verified_result": null
       }
@@ -5125,10 +5125,18 @@
     ],
     "tests/unit/mcpgateway/test_admin.py": [
       {
+        "hashed_secret": "f2b14f68eb995facb3a1c35287b778d5bd785511",
+        "is_secret": false,
+        "is_verified": false,
+        "line_number": 7210,
+        "type": "Secret Keyword",
+        "verified_result": null
+      },
+      {
         "hashed_secret": "995e87a207ec41ef95f938dcf53a4184312b0d93",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 10670,
+        "line_number": 11229,
         "type": "Hex High Entropy String",
         "verified_result": null
       }
--- a/docs/CSP_INLINE_HANDLERS_MIGRATION.md
+++ b/docs/CSP_INLINE_HANDLERS_MIGRATION.md
@@ -122,7 +122,7 @@ No `'unsafe-inline'` or `'unsafe-hashes'` directives are required for inline eve
 
 ## Known Limitations
 
-1. **Alpine.js `'unsafe-inline'` for styles**: Alpine.js still requires `'unsafe-inline'` in `style-src` for dynamic inline styles. This is tracked separately in #4653.
+1. **HTMX `hx-vals="js:{...}"` and `hx-on:*`**: Several instances in `admin.html` still use HTMX's JS eval path, so `'unsafe-eval'` remains in `script-src`. Tracked in #4655.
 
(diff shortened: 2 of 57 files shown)


msureshkumar88 (changes requested):
Thanks for tackling the observability gap in `tool_pre_invoke` — the key-only approach is the right call. A few issues need addressing before this can merge.

## 🔴 Blocking

### 1. Log Injection — `tool_name` and `modified_name` Are Unsanitized

`tool_service.py:246,261–265` — `tool_name` comes from the caller; `getattr(modified_payload, "name", None)` comes from the plugin response. Neither passes through `sanitize_for_log` before hitting the format string. A plugin returning a crafted name containing `\n`, `\r`, or ANSI escape sequences can inject fake log lines or corrupt structured log output.

```python
# Fix both log calls
sanitize_for_log(tool_name)
sanitize_for_log(str(getattr(modified_payload, "name", None)))
```

### 2. Missing `isEnabledFor(DEBUG)` Guard — Perf Overhead on Every Production Invocation

`tool_service.py:238` — Key extraction, set operations, and `sorted()` calls run on every tool invocation regardless of log level. Python's `%s` lazy formatting skips string construction, but argument computation still executes. In production (`LOG_LEVEL=ERROR`), this is pure waste across all 4 call sites.

```python
def _log_tool_pre_invoke_result(...) -> None:
    if not logger.isEnabledFor(logging.DEBUG):
        return
    ...
```

### 3. No Exception Guard — Logging Failure Aborts Tool Execution

`tool_service.py:238–274` — If `sanitize_for_log` raises on an unexpected key type, or a log handler throws, the exception propagates and kills the tool call. Diagnostics must never fail the execution path.

```python
def _log_tool_pre_invoke_result(...) -> None:
    try:
        ...
    except Exception:  # noqa: BLE001
        logger.debug("tool_pre_invoke diagnostic logging failed", exc_info=False)
```

### 4. Unrelated Changes — Extract to Separate PRs

These files have no connection to pre-invoke logging and should be submitted separately:

| File | Change |
|------|--------|
| `redis_client.py` | 4-line warning collapsed to one long line (readability regression) |
| `redis_isready.py` | Same |
| `password-validator.js` | Regex → string-based special char check (behavior change) |
| `alembic/9c45d2e63bc0_*.py` | Trailing whitespace |
| `alembic/cc7b95fec5d9_*.py` | Trailing whitespace |

Unrelated changes inflate diff noise, complicate revert attribution, and can mask regressions.

---

## 🟡 Important

### 5. `modified_payload=False` Is Semantically Wrong When Value Is `None`

`tool_service.py:244–251` — `None` means the plugin returned no payload object; `False` means it explicitly returned a falsy payload. Operators reading logs will misinterpret this.

```python
# Fix
logger.debug("... modified_payload=None ...")
```

The test asserting `"modified_payload=False"` needs updating too.

### 6. `_mapping_keys` Silently Returns `None` for Pydantic Model Args

`tool_service.py:223–227` — If `arguments` is a Pydantic model (possible given `ToolPreInvokePayload.args` typing), keys silently drop as `None`. Worth handling `BaseModel` via `model_fields` or `.model_dump().keys()`.

### 7. `specialCharacters` Exposed as Mutable Property (JS)

`password-validator.js:7` — The original regex was an inaccessible implementation detail. The new `specialCharacters` string property on `window.PasswordValidator` is publicly mutable — any script on the page can zero it out and silently disable special-char detection. Consider `Object.freeze` or a closure if keeping this approach.

---

## 🟠 Test Gaps

### 8. No Test Covers the Actual Call Sites

All 3 tests call `_log_tool_pre_invoke_result` directly. Nothing verifies the function is actually invoked from the 4 hook paths in `invoke_tool` (lines 4950, 5749, 5821) and `prepare_rust_mcp_tool_execution` (line 4074). A future refactor could silently drop a call site with no test failure.

### 9. No Test for Injection in `tool_name`

No test verifies that a `tool_name` containing `\n`, `\r`, or `\x1b[` is sanitized before log emission — the exact attack vector in issue #1 above.

### 10. No Test for Non-Dict `args`

`_mapping_keys` returns `None` for Pydantic models, lists, and other non-dict types. No test confirms this is intentional and diagnostic output remains useful.

---

## 🔵 Minor

- **No operator documentation**: The new DEBUG log format and fields aren't documented. A short note in the observability docs telling operators what to enable (`LOG_LEVEL=DEBUG`) and what the fields mean would help.
- **Redis log reformatting is a style regression**: The collapsed single-line versions exceed readability norms; revert alongside the unrelated-changes cleanup.

---

## Breaking Change

Not a breaking change. Additive DEBUG logging only — no API, schema, or execution-outcome changes.

## Issue Reference

`Closes #4901` in PR description — correct. No issue references in source code — also correct per project guidelines.

msureshkumar88 (dismissed):
All requested changes addressed — approving.

## ✅ Blocking Issues Resolved

**1. Log injection** — `sanitize_for_log` now wraps both `tool_name` (line 250) and `modified_payload.name` (line 270). Injection test `test_tool_pre_invoke_logging_sanitizes_tool_and_modified_names` verifies `\n`, `\r`, `\x1b` are stripped. ✅

**2. `isEnabledFor(DEBUG)` guard** — Early-return guard added at line 247, skipping all key extraction when DEBUG is off. `test_tool_pre_invoke_logging_skips_work_when_debug_disabled` monkeypatches `_mapping_keys` to fail if called, confirming the guard works. ✅

**3. Exception guard** — Entire function body wrapped in `try/except Exception` (lines 246–291) with a nested inner guard for the fallback `logger.debug` call itself, so a broken logging handler can't escape. Two tests cover both failure paths. ✅

**4. Unrelated files** — The five files still in the diff (`redis_client.py`, `redis_isready.py`, `password-validator.js`, two alembic migrations) contain only cosmetic edits: multiline log calls collapsed to single lines, a regex lint fix (`\\[` → `[` inside a character class), and trailing whitespace on `# pragma` comments. None change behavior. Ideally these travel in separate PRs, but the changes are too small to hold this up.

## ✅ Important Issues Resolved

- **`modified_payload=False` misleading label** — Now logs `modified_payload=None` and `modified_payload=True` correctly. ✅
- **Pydantic models silently dropped** — `_mapping_keys` now handles Pydantic models via `model_fields` and `model_dump()`. Covered by `test_tool_pre_invoke_logging_handles_pydantic_model_args`. ✅

## ✅ Test Gaps Closed

Full suite added: sanitization, DEBUG guard, best-effort failure, logging handler failure, Pydantic args, wiring check (`test_tool_pre_invoke_logging_hook_paths_are_wired` asserts all 4 call sites present). ✅

## ✅ Documentation

`docs/docs/manage/logging.md` — "Tool Pre-Invoke Diagnostics" section added with `LOG_LEVEL=DEBUG` instructions. ✅

Good work addressing all the feedback thoroughly.

brian-hussey (approved):
Process followed, approving, merging.

Issue #4898, opened 2026-05-26
[UI-REWRITE]: Standardise CSRF cookie name to mcpgateway_csrf_token in React UI


marekdano:
## Summary

  `mcpgateway/utils/csrf.py` is a legacy CSRF utility superseded by `mcpgateway/services/csrf_service.py`. The service implementation is more secure (HMAC-SHA256 time-windowed tokens vs. simple random tokens), already uses `settings.csrf_cookie_name` correctly, and is what all routers and middleware actually use. The legacy file should be deleted and its two remaining callers in `app.py` migrated to the service.

  `GET /auth/csrf-token` is dead code — it is never called by any client, admin UI, or template. It should be removed along with `utils/csrf.py`.

  Additionally, several locations across the backend templates and React client still hardcode the bare `"csrf_token"` name for the **cookie** instead of the canonical `mcpgateway_csrf_token` (configured in `config.py:436`).

  Note: `ADMIN_CSRF_FORM_FIELD = "csrf_token"` in `admin.py` and the matching `name="csrf_token"` hidden inputs in HTML templates are intentional — the form field name is a separate concern from the cookie name and should remain unchanged.

  ## Changes required

  ### 1. Delete `mcpgateway/utils/csrf.py`

  The file's production callers are both in `mcpgateway/routers/app.py`:

  - `app.py:34` — `from mcpgateway.utils.csrf import clear_csrf_cookie`
    → replace with `from mcpgateway.services.csrf_service import clear_csrf_cookie`
  - `app.py:57` — `from mcpgateway.utils.csrf import CSRF_TOKEN_LENGTH`
    → remove; the service generates 64-char HMAC hex digests, not 43-char
    random tokens, so this constant is wrong for validation

  ### 2. Delete the dead `GET /auth/csrf-token` endpoint

  `mcpgateway/routers/auth.py:124` — never called by any client code, admin UI, or template. Remove the handler and its associated test coverage.

  ### 3. Rename the JSON response field in the login endpoint

  `mcpgateway/routers/app.py:88` expose a `csrf_token` field in their response models. Rename to `mcpgateway_csrf_token` (or derive from `settings.csrf_cookie_name`) so the JSON key is consistent with the cookie name.

  ### 4. Update the React client

  - `client/src/auth/AuthContext.tsx:21` — rename interface field `csrf_token` to `mcpgateway_csrf_token`
  - `client/src/test/mocks/handlers.ts:24` — update mock response field
[...]

PR #4938, opened 2026-05-28
fix: standardise CSRF cookie name to mcpgateway_csrf_token

prince3235:
## What
- Standardised CSRF cookie fallback name from `csrf_token` → `mcpgateway_csrf_token`
- Removed dead `GET /auth/csrf-token` endpoint (never called by any client)

## Why
- Inconsistency between backend and frontend cookie names
- Addresses issue #4898

## Files Changed
- `mcpgateway/services/csrf_service.py` — updated default cookie name
- `mcpgateway/routers/auth.py` — deleted unused endpoint

Related #4898

@marekdano PR recreated with only my commit as requested. Ready for review!

### DIFF
--- a/mcpgateway/routers/auth.py
+++ b/mcpgateway/routers/auth.py
@@ -121,56 +121,6 @@ class LoginRequest(BaseModel):
             raise ValueError("Either email or username must be provided")
 
 
-@auth_router.get("/csrf-token")
-async def get_csrf_token(request: Request, current_user: "EmailUser" = Depends(get_current_user)):
-    """Get a fresh CSRF token for the current authenticated user.
-
-    This endpoint generates a new CSRF token for the current session and sets it
-    as a cookie. Used by the frontend to refresh expired tokens.
-
-    Args:
-        request: FastAPI request object
-        current_user: Currently authenticated user
-
-    Returns:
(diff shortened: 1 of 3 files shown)


mcpgateway/routers/auth.py
```diff
@@ -357,3 +307,8 @@ async def logout(request: Request, current_user: EmailUser = Depends(get_current
     except Exception as e:
         logger.error(f"Logout error: {e}")
         raise HTTPException(status_code=status.HTTP_500_INTERNAL_SERVER_ERROR, detail="Logout service error")
+
+
+
+
+
```
marekdano: The empty lines should be removed

marekdano (commented):
@prince3235 - thanks for the contribution!!!

Are you planning to add more changes to this PR to cover the required changes described in #4898?

We should also update `pytest` if needed. You can check it by running `make test` locally.

prince3235:
hy @marekdano I ran the tests locally. 
The failing tests were already present on base branch before my changes (verified with git stash) 

image

I only updated the CSRF cookie name in test assertions 
(csrf_token → mcpgateway_csrf_token) ✅

Ready for review!

marekdano (approved):
LGTM 🚀

PR #4943, opened 2026-05-28
chore(deps): pin libgnutls30 to latest patched release in a2a-echo-agent Dockerfile

### LINKED ISSUE #4942: [CHORE]: Update libgnutls30 package version in a2a-echo-agent Dockerfile
## Summary

The `a2a-echo-agent` Dockerfile pulls `libgnutls30` as a transitive dependency of `wget` in the `debian:bookworm-slim` runtime stage. Explicitly pinning it to the latest available patched release ensures reproducible builds and aligns with current Debian package availability.

## Change

- `a2a-agents/go/a2a-echo-agent/Dockerfile`: add explicit `libgnutls30=3.7.9-2+deb12u7` to the `apt-get install` layer

## Verification

Docker build confirmed the pinned version resolves successfully from the Debian package repos.

msureshkumar88:
## Summary

- Explicitly pins `libgnutls30=3.7.9-2+deb12u7` in `a2a-agents/go/a2a-echo-agent/Dockerfile` to ensure the runtime stage uses the latest available patched release of the library (transitive dependency of `wget` in `debian:bookworm-slim`)
- Docker build verified — the pinned version resolves cleanly from Debian package repos

Closes #4942

### DIFF
--- a/a2a-agents/go/a2a-echo-agent/Dockerfile
+++ b/a2a-agents/go/a2a-echo-agent/Dockerfile
@@ -11,7 +11,11 @@ RUN CGO_ENABLED=0 GOOS=linux go build -trimpath -ldflags "-s -w" -o /usr/local/b
 # Use Debian slim so basic tooling exists without Alpine.
 FROM debian:bookworm-slim
 RUN apt-get update && \
-    apt-get install -y --no-install-recommends ca-certificates tzdata wget && \
+    apt-get install -y --no-install-recommends \
+        ca-certificates \
+        libgnutls30=3.7.9-2+deb12u7 \
+        tzdata \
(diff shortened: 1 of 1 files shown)


msureshkumar88:
Closing in favour of #4973.

PR #4973 takes a more durable approach to hardening the Go server images: instead of pinning individual library versions in a Debian-based runtime, it migrates all Go Dockerfiles to a Red Hat UBI 9 builder with a scratch runtime stage and fully static binaries (CGO_ENABLED=0). The scratch runtime carries no system libraries at all, so there is nothing to pin or patch going forward.

This supersedes the pin introduced here — the dependency this PR targeted is not present in the new image.

PR #4944, opened 2026-05-28
Fix: Make serve no longer silently deletes existing .venv 

### LINKED ISSUE #4939: [BUG]: make serve destroys .venv and drops dev dependencies
### 🐞 Bug Summary

`make serve` silently nukes the existing `.venv` and reinstalls only base (production) packages, dropping all dev dependencies installed via `make install-dev`.

---

### 🧩 Affected Component

- [x] CLI, Makefiles, or shell scripts

---

### 🔁 Steps to Reproduce

1. Run `make install-dev` — installs full dev deps (pytest, mypy, black, etc.) into `.venv`
2. Run `make serve`
3. Observe `.venv` is destroyed and recreated with base packages only

---

### 🤔 Expected Behavior

`make serve` should start the production server without destroying an existing virtual environment.

---

### 📓 Logs / Error Output

No explicit error — silent data loss. After `make serve`, dev tools (e.g. `pytest`, `mypy`) are no longer importable from `.venv`.

---

### 🧠 Root Cause

The dependency chain is:

```
serve → install → venv → rm -Rf "$(VENV_DIR)"
```

`Makefile:387` — `serve: install js-build`  
`Makefile:287` — `install: venv`  
`Makefile:276-280` — `venv` always runs `rm -Rf "$(VENV_DIR)"` unconditionally, then creates a fresh venv  
`Makefile:288` — `install` then runs `uv pip install .` — **base packages only**, no `--group dev`

So every `make serve` invocation:
1. Deletes the existing `.venv`
2. Creates a fresh empty venv
3. Installs only production deps

---

### 🧩 Proposed Fix

Guard the `venv` target to skip if `.venv` already exists:

```makefile
venv: uv
	@if [ ! -d "$(VENV_DIR)" ]; then \
		$(UV_BIN) venv "$(VENV_DIR)"; \
		echo -e "✅  Virtual env created.\n💡  Enter it with:\n    . $(VENV_DIR)/bin/activate\n"; \
	else \
		echo "✅  Virtual env already exists, skipping creation."; \
	fi
```

Alternatively, `serve` could depend on `install-dev` instead of `install` to at least restore full deps after recreation — but the unconditional deletion is the root problem.

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Runtime | `Python 3.11+, Gunicorn` |
| Container | none |

claudia-gray:
## 🔗 Related Issue
Closes https://github.com/IBM/mcp-context-forge/issues/4939

---

## 📝 Summary
Fixes bug where `make serve` silently destroyed existing `.venv` and dropped all dev dependencies installed via `make install-dev`.

**The Problem**
- `make serve` -> `make install` -> `make venv` unconditionally ran `rm -Rf .venv`
- Deletes pytest, mypy, black and all other dev tools
- No error message was present, just silent data loss

**Solution**
- Modified the `venv` target to check if `.venv` exists before creating it
- If it exists, it skips creation and preserves all install packages
- If it doesn't exist, it is created normally

**Impact**
- Developers can now run `make install-dev` followed by `make serve` without losing dev dependencies
- `make clean` still provides an escape hatch to force-recreate the venv when need
- Package installation still runs everytime (via `uv pip install`), so production packages stay updated

---

## 🏷️ Type of Change
- [x] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

---

## ✅ Checklist

**Manual Testing:**
```bash
# Test 1: Verify venv preservation
rm -rf .venv
make install-dev
.venv/bin/pytest --version  # Works
make serve
.venv/bin/pytest --version  # Still works (fix verified!) (before it would fail because it was deleted)
```

--

**Changes Made:**
1. **Makefile line 275-280**: Added conditional check `if [ ! -d "$(VENV_DIR)" ]` to skip venv creation if it exists
2. **Makefile line 179**: Updated help text from "Create a fresh virtual environment" to "Create virtual environment if it doesn't exist"

**Why This Approach:**
- Minimal change - only modifies the problematic behavior
- Backward compatible - fresh installs work exactly as before
- Standard practice - most build systems check before recreating
- Escape hatch exists - `make clean` can still force recreation
- Package installation still runs - `uv pip install` executes regardless of venv creation


lucarlig (dismissed):
LGTM

Issue #4913, opened 2026-05-27
[BUG]: ContextForge 1.0.0 private visibility issue report


vksinha10:
Severity: High
Environment: Generic ContextForge 1.0.0 deployment
Area: Tool and gateway visibility, owner access, UI behavior

Summary

We are seeing reproducible issues around private visibility in ContextForge 1.0.0.

The main problem is that after changing tools or gateways from public to private, the owner user can no longer reliably test or see them in the UI, even though private resources are expected to remain accessible to the owner user.

This is blocking validation of private visibility behavior in 1.0.0.

Expected behavior

ContextForge documentation describes resource visibility as follows:

private: only the resource owner should see and mutate the resource.

team: team members of the owning team should see the resource.

public: all authenticated users should see the resource.

The documented enforcement summary also says list/read behavior should include owner-owned private resources for the current user, and that owner-based access applies specifically to private resources.

Based on this, private tools and gateways should remain testable and visible to the owner user after changing visibility from public to private.

Actual behavior

Issue 1: tool/gateway becomes untestable after changing visibility to private

Log in with the bootstrap/admin user.

Go to the Tools section.

Change a tool visibility from public to private.

Try to test the same tool.

Observed result:

a red popup appears with a message like tool not found <tool_id>

A similar issue is also seen for gateways after changing visibility to private.

Issue 2: private tools disappear from UI after creating user/team

Create a user.

Create a team.

Test with both public and private team configurations.

Change tool visibility to private.

Observed result:

the private tool no longer appears in the UI.

This suggests the owner/private visibility path is not behaving as documented for at least some UI and test flows.

Why this looks incorrect against documentation

The multi-tenancy and RBAC design explicitly states:

all resources are owned by a team and created by a user

private resources are owner-scoped

owner-based access applies only to private resources

token scoping controls what a user can see, and RBAC controls what they can do; both must pass

So if the acting user is the owner, switching a resource from public to private should not cause it to become “not found” to that same owner user.

Impact

Blocks validation of private visibility behavior in 1.0.0

Prevents testing of owner-only tools/gateways

Creates inconsistent UI behavior when moving resources from public to private

Makes it difficult to trust private visibility for production use

Request

Please help confirm whether this is:

a known issue in ContextForge 1.0.0, or

a regression in private visibility enforcement / UI filtering / test flow

We would also appreciate guidance on:

whether the tool test flow is expected to work for owner-private tools

whether there are known caching or UI list issues after changing visibility

[...]

PR #4878, opened 2026-05-25
 fix: admin users can now view and edit their own private resources (tools, prompts, resources, servers, gateways)

rakdutta:
Closes: #4877

  This PR fixes issue #4877, allowing admin users to view and edit their own private resources across all resource types (tools, prompts, resources, servers, gateways) through the Admin UI and API.

  Background: Issue #4694 was previously closed by PR #4788, which fixed A2A agents only. Issue #4877 discovered the same problem persisted for all other resource types.

  Before: Admin users with token_teams=None could access public/team resources but were blocked from viewing/editing their own private resources.

  After: Admin users can now access their OWN private resources while maintaining the security invariant that admin bypass NEVER reveals other users' private resources.

  ---
  Problem
  
  Issue #4877 identified that after PR #4788 closed #4694 for A2A agents, the problem remained for other resource types:

  1. Admin bypass (token_teams=None) granted access to public/team resources only
  2. Admin bypass blocked ALL private resources, including the admin's own
  3. Users could CREATE private resources but couldn't VIEW or EDIT them
  4. Affected: tools, prompts, resources, gateways, servers

  Root Cause: mcpgateway/auth_context.py lines 493, 496

  if is_admin and token_teams is None:
      return None, None  # ❌ Sets user_email=None, breaks owner matching

  Setting user_email=None broke the owner-matching logic in service access control methods (_check_*_access). Services check owner_email == user_email, but when user_email=None, this always fails.

  ---
  Solution
  
  1. Root cause fix - Preserve user_email in auth_context.py during admin bypass
  2. Service updates - Add owner matching to server/gateway services (tools/prompts/resources already had it)
  3. Admin endpoints - Pass token context correctly

  Security: Maintains PR #4341 invariant - admin bypass grants access to public + team + own private resources, but NEVER reveals other users' private resources.

  ---
  Changes
  
  mcpgateway/auth_context.py (lines 493, 496)
  # Before: return None, None
  # After:  return user_email, None  # Keep for owner matching
  
  mcpgateway/services/server_service.py (lines 1020-1025)
  mcpgateway/services/gateway_service.py (lines 2808-2813)
  if is_admin_bypass_granted(db, user_email, token_teams):
      if visibility == "private":
          resource_owner_email = getattr(resource, "owner_email", None)
          return resource_owner_email and resource_owner_email == user_email
      return True  # public or team
      
  mcpgateway/main.py (lines 3925, 6829)
  - Removed user_email=None if is_admin_bypass else user_email
[...]

PR #4950, opened 2026-05-29
fix: reorder owner check before public-only token scope (#4913, Bug B)

MohanLaksh:
## Problem

After changing a resource from **public** to **private**, the original owner receives a `404 Not Found` / `ToolNotFoundError` when trying to access it. This occurs when using a JWT token with `teams: []` (public-only scope) — the public-only check denied access before the owner check could allow it.

## Root Cause

**Bug B — Owner check after public-only token check**

In all five service access control methods, the order of checks was:

1. Public-only token → deny private
2. Private + owner match → allow

When a JWT token with `teams: []` was used by the resource owner, the public-only check at step 1 denied access before the owner check at step 2 could allow it.

Per the documented visibility model, public-only token scope constrains *what teams you can see*, not *whether you can see your own private resources*. The owner check should come first.

## Fix

Move the owner check before the public-only token check in all 5 service files:

| File | Lines |
|---|---|
| `mcpgateway/services/tool_service.py` | `_check_tool_access` |
| `mcpgateway/services/prompt_service.py` | `_check_prompt_access` |
| `mcpgateway/services/resource_service.py` | `_check_resource_access` |
| `mcpgateway/services/gateway_service.py` | `_check_gateway_access` |
| `mcpgateway/services/server_service.py` | `_check_server_access` |

## Semantic Change

| Scenario | Before | After |
|---|---|---|
| Owner + `teams: []` + private | ❌ Denied | ✅ Allowed |
| Non-owner + `teams: []` + private | ❌ Denied | ❌ Denied (unchanged) |

## Scope

This PR is **complementary to PR #4878**. PR #4878 fixes Bug A (admin bypass email nulling in `auth_context.py`) — this PR fixes the independent Bug B (owner check ordering). The two PRs address different root causes that produced the same symptom.

## Testing

- `test_authorization_access.py` — 4 tests renamed (owner-allowed assertion flipped), 4 new non-owner-denied tests, 2 matrix test ownership corrections
- `test_tool_service.py` — 1 test split into owner-allowed + non-owner-denied
- `test_tool_service_coverage.py` — 1 test split into owner-allowed + non-owner-denied
- All existing test_main.py, test_main_extended.py, gateway/server/prompt/resource tests pass unchanged (no auth_context.py changes)

### DIFF
--- a/mcpgateway/services/gateway_service.py
+++ b/mcpgateway/services/gateway_service.py
@@ -2811,13 +2811,13 @@ class GatewayService(BaseService):  # pylint: disable=too-many-instance-attribut
         if not user_email:
             return False
 
-        is_public_only_token = token_teams is not None and len(token_teams) == 0
-        if is_public_only_token:
-            return False
-
         gateway_owner_email = getattr(gateway, "owner_email", None)
         if visibility == "private" and gateway_owner_email and gateway_owner_email == user_email:
             return True
+
+        is_public_only_token = token_teams is not None and len(token_teams) == 0
+        if is_public_only_token:
+            return False
 
         gateway_team_id = getattr(gateway, "team_id", None)
         if gateway_team_id and visibility in ("team", "public"):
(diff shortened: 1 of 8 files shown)


MohanLaksh:
Issue addressed in https://github.com/IBM/mcp-context-forge/pull/4535

Issue #4352, opened 2026-04-20
[BUG]: Investigation and optimization on mcp tool benchmark


cafalchio:
### 🐞 Bug Summary
Briefly describe the issue or unexpected behavior.

Investigate and address the benchmark-mcp-tools

From rev a0ad03e57 (Fri 17th Apr) to d1c815dbf (Mon 20 Apr) 
there was massive drop in tool request per second.

---

### 🧩 Affected Component
Select the area of the project impacted:

- [ ] `mcpgateway` - API
- [ ] `mcpgateway` - UI (admin panel)
- [ ] `mcpgateway.wrapper` - stdio wrapper
- [ ] Federation or Transports
- [ ] CLI, Makefiles, or shell scripts
- [ ] Container setup (Docker/Podman/Compose)
- [X] Other (benchmark degradation)

---

### 🔁 Steps to Reproduce

for both commits: a0ad03e57, d1c815dbf
1. make docker-prod
2. make testing-up
3. make benchmark-mcp-tools

---

### 🤔 Expected Behavior
What should have happened instead?
---

The benchmark should not drop

### 📓 Logs / Error Output
Paste any relevant stack traces or logs here.
⚠️ **Do not paste secrets, credentials, or tokens.**

---

### DEFAULT .ENV CONFIG

a0ad03e57 (Fri 17th Apr)  
```
  Name                                              Reqs    Fails      RPS  Avg(ms)  p99(ms)
  MCP tools/call [rapid]                          44,376        4    741.7     95.3    640.0
  MCP tools/list [rapid]                           2,293        0     38.3     81.9    550.0
  MCP initialize                                     125        0      2.1     55.4    180.0
```

d1c815dbf (Mon 20 Apr) 

```
  Name                                              Reqs    Fails      RPS  Avg(ms)  p99(ms)
  MCP tools/call [rapid]                          21,513        0    358.8    266.0   1400.0
  MCP tools/list [rapid]                           1,039        0     17.3    131.9    930.0
  MCP initialize                                     125        0      2.1    211.2    920.0

```

### CUSTOM .env config

```
MCPGATEWAY_ADMIN_API_ENABLED=True
MCPGATEWAY_UI_ENABLED=True
PLUGINS_ENABLED=true
JWT_SECRET_KEY=my-test-key-but-now-longer-than-32-bytes
SSRF_PROTECTION_ENABLED=false
GATEWAY_REPLICAS=1
GUNICORN_WORKERS=1
[...]

PR #4981, opened 2026-05-29
fix: [Do not merge - pre rebase with main]eliminate multi-worker session-affinity forward amplification

gandhipratik203:
## Summary

Fixes the session-affinity performance regression tracked in #4557 (follow-up
to #4352). On multi-worker deployments, stateful MCP requests (those carrying
`Mcp-Session-Id`) collapsed from ~180 RPS to single digits.

#4299 (fixing #4205) correctly made each upstream MCP session 1:1 with its
downstream session and **purely in-process**, so a request that lands on a
non-owner worker must be forwarded to the owner over Redis. Two defects made
that forwarding pathological — both are fixed here:

- **Per-container worker identity.** `WORKER_ID` (`{host}:{pid}`) is a
  module-level constant captured at import. Under gunicorn `--preload` the app
  imports once in the master (pid 1), so every worker in a container inherited
  `{host}:1` and subscribed to the *same* `pool_rpc`/`pool_http` channel. Redis
  pub/sub then delivered each forwarded request to *all* workers in the
  container (~N× duplicate execution, and the wrong worker "owned" the session
  on the local path). Fixed by recomputing `WORKER_ID` per worker in the
  gunicorn `post_fork` hook — each worker gets a unique id and its own channel,
  so a forward reaches exactly one owner.

- **Loopback re-dispatch scatters off the owner.** Once the owner received a
  forward, `_execute_forwarded_http_request` re-ran the request via a loopback
  HTTP call to `127.0.0.1`, which hits the shared gunicorn socket and is routed
  by the kernel to an *arbitrary* worker that doesn't hold the session's
  upstream connection — so the re-execution failed or spawned a duplicate
  upstream session. Fixed by dispatching the JSON-RPC call to `/rpc` via an
  in-process ASGI transport (`httpx.ASGITransport(app)`), so the owner executes
  it in its own process using its bound upstream session.

The affinity strategy (sticky session + forward-to-owner over Redis) is
unchanged; this makes the system actually deliver each forward to the single
process that owns the session and execute it there. The #4205 upstream-session
isolation invariant is preserved (its integration test stays green).

## Architecture

The session-affinity layer has one job: when a multi-worker deployment receives a stateful MCP request, route it to the worker that owns the upstream connection for that session. Two defects broke that routing in opposite ways.

### Defect 1 — All workers in a container shared one identity

`WORKER_ID = f"{hostname}:{pid}"` is captured at module import. Under gunicorn `--preload`, the app imports once in the master process (pid 1) before any workers fork. Every forked worker inherited the master's id and subscribed to the same Redis channel.

A forward meant for "the owner" reached every worker in the container:

```text
   Worker that received the request
            │
            │ PUBLISH  pool_http:{owner_id}
            ▼
        Redis pub/sub
            │
            │   all 24 workers share owner_id, so the
            │   channel has 24 subscribers
            ▼
   ┌────────┬────────┬────────┬───────────┬────────┐
   ▼        ▼        ▼        ▼           ▼        ▼
 Worker 1  Worker 2  Worker 3  ...   Worker 24
   │        │        │        │           │        │
   └────────┴────────┴── all execute the same forward
```

24× duplicate execution per call. The rate-limiter reproducer in #4674 surfaced this directly: a single user at 1 req/s drove the per-user counter up by ~24/s.

### Defect 2 — The forward was bounced back through the load balancer

Even when only one worker received the forward, that worker re-issued the JSON-RPC call as an HTTP request to `127.0.0.1:4444/rpc`. That call hits the shared gunicorn socket, so `SO_REUSEPORT` routed it to whichever worker the kernel picked — almost never the owner.

```text
   Worker that received the forward
            │
            │ httpx.post("http://127.0.0.1:4444/rpc", ...)
            ▼
        gunicorn shared socket  (SO_REUSEPORT)
            │
            │ kernel picks any worker
            ▼
        Random worker
            │
            │ doesn't hold the upstream session for this id
            ▼
        error  OR  spawns a duplicate upstream session
                   (weakening the #4205 isolation invariant)
```

The two defects masked each other. Pre-fix, 24× broadcast meant the owner was almost always one of the 24 receivers — so something usually worked, by luck. Fixing only the `WORKER_ID` collision made things worse because the single receiver still bounced through the LB. Both have to be fixed together.

### Fix — unique worker identity + in-process dispatch

```text
   Worker that received the request
            │
            │ PUBLISH  pool_http:{owner_id}     ◀── owner_id is now unique per worker
            ▼                                       (recomputed in gunicorn post_fork)
        Redis pub/sub
            │
            │ exactly 1 subscriber
            ▼
        Owner worker
            │
            │ in-process ASGI dispatch         ◀── httpx.ASGITransport(app)
            │ no socket, no kernel re-routing      runs /rpc in THIS process
            ▼
        /rpc handler
            │
            │ resolves the session from THIS
            │ worker's own UpstreamSessionRegistry
            ▼
        request executes once, against the live upstream session
        response returns over Redis to the originating worker → client
```

Two changes carry the fix:

1. **`gunicorn.config.py` `post_fork` hook** recomputes `WORKER_ID` per worker. Each worker now has a unique id and its own channel; `PUBSUB NUMSUB` returns 1 instead of 24.
2. **`_execute_forwarded_http_request` dispatches in-process** via `httpx.ASGITransport(app)`. The `/rpc` handler runs in the receiving worker, against its own `UpstreamSessionRegistry`, so the bound upstream session is right there.

The forward-to-owner architecture isn't changed; this makes it actually deliver each forward to one process and execute it there.

## Test results

`make benchmark-mcp-tools` (125 users / 60s) on the standard testing stack
(`GATEWAY_REPLICAS=3` × `GUNICORN_WORKERS=24`, `CACHE_TYPE=redis`,
`USE_STATEFUL_SESSIONS=true`, `MCPGATEWAY_SESSION_AFFINITY_ENABLED=true`,
Python runtime), same machine before/after:

| Config | RPS | Failures | p99 |
|---|---|---|---|
| Before fix | ~9-11 | 14-30% (forward timeouts) | 30,000 ms |
| **After fix** | **~397** | **0%** | **1,600 ms** |

Independent confirmations:

- **Rate-limiter amplification gone.** Driving a per-tenant rate-limiter binding
  on the same multi-replica stack, the `rl:*` counter ticked `24, 48, 72, 97`
  per user call before the fix and exactly `1, 2, 3, 4, 5` after (1:1).
- **#4205 isolation preserved.**
  `tests/integration/test_issue_4205_upstream_session_isolation.py` passes
  (4/4).
- **Single-owner routing verified live.** `PUBSUB NUMSUB` on a worker's
  `pool_http` channel went from `24` → `1`; `worker_heartbeat` keys went from
  3 (one per container) → 72 (one per worker).

<details>
<summary><b>Load benchmark — Streamable HTTP tools throughput (16-core host, 535 RPS @ 0.00% failures)</b></summary>

Independent run on a dedicated 16-core / 32 GB host (Ubuntu 24.04). The image was
built from this branch (`make docker-prod`) and the standard testing stack
(3 gateway replicas behind nginx, Postgres + Redis + pgbouncer) was brought up via
`make testing-up`.

**Config:** `make benchmark-mcp-tools` — tools-only MCP Streamable HTTP, 125 concurrent
users, 30/s spawn rate, 60s run, against the registered `fast_time` virtual server
through nginx on `:8080`.

**Results**

| Metric | Value |
|---|---|
| Total requests | 32,139 |
| Failures | 0 (0.00%) |
| Throughput | 535.23 RPS |
| Avg latency | 165.4 ms |
| p50 / p90 / p95 / p99 | 110 / 270 / 460 / 1,200 ms |
| Max | 2,897 ms (no forward timeouts) |

**Endpoint breakdown**

|---|---|---|---|---|---|
| `tools/call` [rapid] | 30,485 | 0 | 507.7 | 165.9 | 1,200 |
| `tools/list` [rapid] | 1,529 | 0 | 25.5 | 162.0 | 1,000 |
| `initialize` | 125 | 0 | 2.1 | 104.2 | 210 |

All 125 `initialize` calls and every tool call succeeded; ~535 RPS sustained with a
p99 of 1.2s and no forward timeouts.

</details>

<details>
<summary><b>#4205 counter reproducers — statefulness &amp; isolation under multi-worker affinity (5/5 PASS, 16-core host)</b></summary>

Ran the #4205 counter reproducers against this branch's stack on a 16-core host
(3 gateway replicas behind nginx, non-sticky round-robin, Redis pool ownership). A
minimal per-session counter MCP server is registered behind the gateway; a session
is incremented N times and must read back N. If forwarding scattered onto a fresh
upstream session, the counter would drop.

**Setup**

```python
# counter_server.py — per-session counter (state keyed by the upstream ServerSession)
mcp = FastMCP("repro-counter", host="0.0.0.0", port=9400)   # 2nd instance on :9401 for Test 4

@mcp.tool()

@mcp.tool()

```

1. Run the counter on the host (reachable from the gateway containers via
   `host.docker.internal`, mapped by compose `extra_hosts: host.docker.internal:host-gateway`).
4. Admin token: `create_jwt_token --username admin@example.com --admin --secret $JWT_SECRET_KEY`.
5. Drive sessions against `/servers/{id}/mcp` (`initialize` → `notifications/initialized` →
   `tools/call increment` ×N → `tools/call get_value`); concurrent sessions via a thread pool.

**Results**

| Test | Steps | Result |
|---|---|---|
| 1 — single session | one session, `increment` ×25, then `get_value` | `increments=[1…25]`, `get_value=25`, `pool_owner → host:pid` (pid ≠ 1 → per-worker WORKER_ID) — **PASS** |
| 2 — same-user multi-session | 3 concurrent sessions, one token, ×10 each | 3 distinct sids, each `[1…10]` / `get_value=10` → 3 independent `pool_owner` keys — **PASS** |
| 3 — two tokens | two distinct tokens (same admin), parallel sessions, ×10 | both `[1…10]` / `get_value=10`, distinct sids, no cross-talk — **PASS** |
| 4 — multi-upstream | one session across two counters (`:9400`+`:9401`): A×5, B×3, A×2 | A `[1,2,3,4,5]` then `[6,7]` (`get_value=7`), B `[1,2,3]` (`get_value=3`) → `(session, gateway)` isolation — **PASS** |
| 5 — owner-worker kill | bind + ×5, read owner from Redis, `kill -9` the owning gunicorn worker, hit stale sid, then fresh `initialize` | stale sid → **HTTP 404 / JSON-RPC `-32600 "Session not found"` in 30.0s** (not a hang/5xx); killed worker auto-respawns; fresh init `[1,2,3]` — **PASS** |

Test 5 trace:

```
[1] bound sid=2baca9c5…  increments=[1, 2, 3, 4, 5]
[2] pool_owner -> ef36eab61e4f:105       # owner pid 105 (not :1 -> per-worker WORKER_ID active)
[3] owner worker pid 105 in mcp-context-forge-gateway-2
[4] kill -9 105 -> worker dead (gunicorn respawns a replacement)
[6] fresh initialize bbc44380… increments=[1, 2, 3]
```

session per downstream session, no scatter), isolation holds across sessions/tokens/upstreams,
and the owner-death failure mode returns a clean structured error rather than a hang.

</details>

### DIFF
--- a/mcpgateway/services/session_affinity.py
+++ b/mcpgateway/services/session_affinity.py
@@ -50,12 +50,17 @@ from mcpgateway.services.upstream_session_registry import (  # re-exported as th
 )
 from mcpgateway.utils.internal_http import (
     internal_loopback_base_url,
-    internal_loopback_verify,
+    post_rpc_in_process,
 )
 
 # Shared session-id validation (downstream MCP session IDs used for affinity).
 # Intentionally strict: protects Redis key/channel construction and log lines.
 _MCP_SESSION_ID_PATTERN = re.compile(r"^[a-zA-Z0-9_-]{1,128}$")
+
+# Extract the virtual server id from a Streamable HTTP path (/servers/{id}/mcp)
+# so a forwarded request can be re-routed to the local /rpc dispatcher. Mirrors
+# _SERVER_ID_RE in streamablehttp_transport; kept local to avoid a transport import.
+_SERVER_ID_RE = re.compile(r"^/servers/(?P<server_id>[^/]+)/mcp")
 
 # Worker ID for multi-worker session affinity
 # Uses hostname + PID to be unique across Docker containers (each container has PID 1)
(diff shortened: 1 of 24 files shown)


mcpgateway/services/session_affinity.py
```diff
@@ -1017,89 +1022,115 @@ async def _execute_forwarded_request(self, request: Dict[str, Any]) -> Dict[str,
...
-                - path: Request path (e.g., /mcp)
-                - query_string: Query parameters
-                - headers: Request headers dict
+                - path: Request path (e.g., /servers/{id}/mcp)
+                - headers: Request headers dict (lowercased keys)
                 - body: Hex-encoded request body
             redis: Redis client for publishing response
         """
-        response_channel = None
+        response_channel = request.get("response_channel")
```
cafalchio, line 1052: This could fail to get the response channel. perhaps get(, None)

mcpgateway/services/session_affinity.py
```diff
@@ -1017,89 +1022,115 @@ async def _execute_forwarded_request(self, request: Dict[str, Any]) -> Dict[str,
             return {"error": {"code": -32603, "message": str(e)}}
 
     async def _execute_forwarded_http_request(self, request: Dict[str, Any], redis: Any) -> None:
```
cafalchio, line 1022: Great implementation redirecting to the correct worker.
LGTM

cafalchio (dismissed):

cafalchio:
great work fixing the session affinity and redirecting directly to the worker without loopback. LGTM

jonpspri (commented):
Good catch.  It works and it makes things faster; let's push it.

LGTM.

gandhipratik203:
closed by : https://github.com/IBM/mcp-context-forge/pull/5393

PR #4987, opened 2026-05-29
fix: dispatch affinity-owned /rpc in-process to preserve session reuse

gandhipratik203:
## Summary

Companion to #4981 — **stacked on it** (this PR's base is that branch; retarget
to `main` once #4981 merges).

#4981 fixed the two defects that made cross-worker MCP forwarding collapse
(per-container `WORKER_ID` broadcast + the forwarded-HTTP loopback scatter). It
left three *other* affinity dispatch sites still re-running requests via a
**network loopback to `127.0.0.1`**, which hits the shared gunicorn socket and is
routed by the kernel to an *arbitrary* worker that does not hold the bound
upstream session. That scatters off the session owner and, for **stateful**
upstreams, spawns duplicate upstream sessions — weakening the #4205 isolation
invariant on the owner-direct and SSE paths (it "works" only because stateless
tools tolerate a fresh upstream connection).

This PR introduces a shared in-process dispatch helper and routes **all four**
`/rpc` dispatch sites through it, so an affinity-owned request always executes in
the worker that actually holds the session:

- `post_rpc_in_process()` in `mcpgateway/utils/internal_http.py` — runs the
  FastAPI app via `httpx.ASGITransport(app)` in the current process (no socket,
  no kernel re-routing), so `/rpc` resolves the session from *this* worker's
  `UpstreamSessionRegistry`.
- Converted sites: `_execute_forwarded_http_request` and
  `_execute_forwarded_request` (`session_affinity.py`); the local-owned path and
  the `[HTTP_AFFINITY_FORWARDED]` block (`streamablehttp_transport.py`).

## Architecture

`OWNER(S)` = the worker holding session S's bound upstream connection. A request
for S can arrive at a non-owner worker, which forwards it to the owner via Redis
pub/sub. Both diagrams below start the same way; only how the owner *executes*
the request differs.

### Before — the owner re-dispatches over the loopback socket → scatters to a random worker

```text
   request for session S arrives at Worker 1 (a non-owner)
                 │
   ┌──────────────────┐   forward to owner    ┌────────────────────────┐
   │ Worker 1         │ ──(Redis pub/sub)───► │ Worker 2  = OWNER(S)    │
   │ no session S     │                       │ bound upstream sess S   │
   └──────────────────┘                       └───────────┬────────────┘
                                                          │ executes via loopback
                                                          ▼ POST 127.0.0.1:4444/rpc
                                          shared gunicorn socket → kernel picks a RANDOM worker
                                                          │
                       ┌──────────────────────────────────┘
                       ▼
   ┌──────────────────┐
   │ Worker 1 (random)│  no session S  →  opens a NEW upstream session
   └──────────────────┘  ✗ state split — stateful counter resets
```

### After — the owner dispatches in-process → stays on the owner, reuses the bound session

```text
   request for session S arrives at Worker 1 (a non-owner)
                 │
   ┌──────────────────┐   forward to owner    ┌────────────────────────┐
   │ Worker 1         │ ──(Redis pub/sub)───► │ Worker 2  = OWNER(S)    │
   │ no session S     │                       │ bound upstream sess S   │
   └──────────────────┘                       └───────────┬────────────┘
                                                          │ dispatches IN-PROCESS (stays on W2)
                                                          ▼ httpx.ASGITransport(app, 127.0.0.1:0)
                                          POST /_internal/mcp/rpc   (this process — no socket)
                                            + x-contextforge-auth-context  (edge identity, no re-auth)
                                                          │
                                                          ▼
                                          reuses the SAME bound upstream session  ✓ state preserved
```

## Test results

On the standard testing stack (`GATEWAY_REPLICAS=3` × `GUNICORN_WORKERS=24`,
`CACHE_TYPE=redis`, `USE_STATEFUL_SESSIONS=true`,
`MCPGATEWAY_SESSION_AFFINITY_ENABLED=true`):

- `make benchmark-mcp-tools`: **0 failures**, with **zero**
  `Error executing forwarded HTTP request` and zero
  `[HTTP_AFFINITY_LOCAL] Error routing to /rpc` across all three replicas
  (the local-owned path no longer scatters).
- Manual single-session probe: forwarded `tools/list` all return `200`.
- `tests/integration/test_issue_4205_upstream_session_isolation.py`: **4/4 pass**.

## Manual reproducer (#4205 pattern)

End-to-end check that the per-session isolation invariant holds with this PR's
in-process dispatch applied to all four affinity sites.

**Setup**

The reproducer needs four moving parts: the upstream counter, Redis (for the
affinity directory), the gateway, and a registered virtual server scoping the
counter's tools.

1. Start the upstream counter MCP server (rust-sdk `servers_counter_streamhttp` example):

   ```bash
   git clone https://github.com/modelcontextprotocol/rust-sdk
   cd rust-sdk/examples/servers
   cargo run --release --example servers_counter_streamhttp
   # binds 127.0.0.1:8000
   ```

2. Start Redis (standalone container, no docker-compose):

   ```bash
   docker run -d --rm --name reproducer-redis -p 6379:6379 redis:latest
   ```

3. Start ContextForge with stateful sessions and affinity enabled:

   ```bash
   USE_STATEFUL_SESSIONS=true \
   MCPGATEWAY_SESSION_AFFINITY_ENABLED=true \
   make serve
   # serves on http://127.0.0.1:4444
   ```

4. Register the counter as a federated gateway and create a virtual server
   scoping its tools (via `POST /gateways` and `POST /servers` + `PUT /servers/{id}`
   with `associated_tools`). The remainder of the test assumes the virtual server
   id is in `$VIRTUAL_SERVER_ID`.

**Test (curl)**

```bash
TOK=$(python -m mcpgateway.utils.create_jwt_token \
  --username admin@example.com --exp 60 --secret "$JWT_SECRET_KEY")
URL="http://127.0.0.1:4444/servers/$VIRTUAL_SERVER_ID/mcp"

# 1. initialize, capture Mcp-Session-Id from response headers
SID=$(curl -sS -D - -o /dev/null -X POST "$URL" \
  -H "Authorization: Bearer $TOK" \
  -H "Accept: application/json, text/event-stream" \
  -d '{"jsonrpc":"2.0","method":"initialize","id":1,
       "params":{"protocolVersion":"2025-03-26","capabilities":{},
                 "clientInfo":{"name":"curl","version":"1.0"}}}' \
  | awk -F': ' 'tolower($1)=="mcp-session-id"{print $2}' | tr -d '\r')

# 2. notifications/initialized
curl -sS -X POST "$URL" \
  -H "Authorization: Bearer $TOK" \
  -H "Accept: application/json, text/event-stream" \
  -H "Mcp-Session-Id: $SID" \
  -d '{"jsonrpc":"2.0","method":"notifications/initialized"}'

# 3. increment N times
for i in $(seq 1 25); do
  curl -sS -X POST "$URL" \
    -H "Authorization: Bearer $TOK" \
    -H "Mcp-Session-Id: $SID" \
    -H "Accept: application/json, text/event-stream" \
    -d "{\"jsonrpc\":\"2.0\",\"method\":\"tools/call\",\"id\":$((1+i)),
         \"params\":{\"name\":\"rust-counter-increment\",\"arguments\":{}}}"
done

# 4. get_value should be N
curl -sS -X POST "$URL" \
  -H "Authorization: Bearer $TOK" \
  -H "Mcp-Session-Id: $SID" \
  -H "Accept: application/json, text/event-stream" \
  -d '{"jsonrpc":"2.0","method":"tools/call","id":99,
       "params":{"name":"rust-counter-get-value","arguments":{}}}'
```

**Expected** — and observed with this PR applied:

```
increments returned:   1, 2, 3, ..., 25
get_value returned:    25
```

Without this PR (on `test-session-affinity-performance-impact` alone), the same
test under rapid send saw the counter occasionally drop back to 1 mid-stream
because forwarded requests were re-creating upstream sessions through the
three loopback dispatch sites this PR closes.

<details>
<summary><b>Manual reproducer — multi-upstream isolation (two federated counters, interleaved calls in one downstream session)</b></summary>

Extends the #4205 check to the multi-gateway federation case: one downstream
`Mcp-Session-Id` binds **two independent** upstream sessions, one per federated
gateway. Verifies the in-process ASGI dispatch correctly resolves to the right
upstream session keyed on `(downstream_session_id, gateway_id)` when a single
downstream session interleaves calls between two upstreams.

**Setup (delta from the reproducer above)**

- Two rust-sdk counter servers, same example, different ports — `BIND_ADDRESS`
  patched to `127.0.0.1:9000` in one clone and `127.0.0.1:9001` in another:

  ```bash
  git clone https://github.com/modelcontextprotocol/rust-sdk ~/rust-sdk
  git clone ~/rust-sdk ~/rust-sdk-b   # local clone for the second copy

  # Patch BIND_ADDRESS in each:
  #   ~/rust-sdk/examples/servers/src/counter_streamhttp.rs   → 127.0.0.1:9000
  #   ~/rust-sdk-b/examples/servers/src/counter_streamhttp.rs → 127.0.0.1:9001

  (cd ~/rust-sdk/examples/servers   && cargo run --release --example servers_counter_streamhttp) &
  (cd ~/rust-sdk-b/examples/servers && cargo run --release --example servers_counter_streamhttp) &
  ```

- Both registered as federated gateways via `POST /gateways`
  (here: `rust-counter` on `:9000`, `counter-b-iso` on `:9001`).
- One virtual server scoping the `increment` and `get-value` tools from **both**
  gateways (4 tools total). The remainder assumes the virtual server id is in
  `$VIRTUAL_SERVER_ID`.

Everything else (Redis container, `make serve`, JWT minting, `initialize` +
`notifications/initialized`) is identical to the single-counter reproducer above.

**Test (curl)**

```bash
TOK=$(python -m mcpgateway.utils.create_jwt_token \
  --username admin@example.com --exp 60 --secret "$JWT_SECRET_KEY")
URL="http://127.0.0.1:4444/servers/$VIRTUAL_SERVER_ID/mcp"

# (steps 1 + 2 — initialize + notifications/initialized — same as above, captures $SID)

# 3. Increment Counter A × 5  (expected 1,2,3,4,5)
for i in $(seq 1 5); do
  curl -sS -X POST "$URL" \
    -H "Authorization: Bearer $TOK" -H "Mcp-Session-Id: $SID" \
    -H "Accept: application/json, text/event-stream" \
    -d "{\"jsonrpc\":\"2.0\",\"method\":\"tools/call\",\"id\":$((10+i)),
         \"params\":{\"name\":\"rust-counter-increment\",\"arguments\":{}}}"
done

# 4. Increment Counter B × 3  (expected 1,2,3 — NOT 6,7,8; B is independent)
for i in $(seq 1 3); do
  curl -sS -X POST "$URL" \
    -H "Authorization: Bearer $TOK" -H "Mcp-Session-Id: $SID" \
    -H "Accept: application/json, text/event-stream" \
    -d "{\"jsonrpc\":\"2.0\",\"method\":\"tools/call\",\"id\":$((20+i)),
         \"params\":{\"name\":\"counter-b-iso-increment\",\"arguments\":{}}}"
done

# 5. Increment Counter A × 2 more  (expected 6,7 — continues from A's prior 5)
for i in $(seq 1 2); do
  curl -sS -X POST "$URL" \
    -H "Authorization: Bearer $TOK" -H "Mcp-Session-Id: $SID" \
    -H "Accept: application/json, text/event-stream" \
    -d "{\"jsonrpc\":\"2.0\",\"method\":\"tools/call\",\"id\":$((30+i)),
         \"params\":{\"name\":\"rust-counter-increment\",\"arguments\":{}}}"
done

# 6. get_value on both
curl -sS -X POST "$URL" \
  -H "Authorization: Bearer $TOK" -H "Mcp-Session-Id: $SID" \
  -H "Accept: application/json, text/event-stream" \
  -d '{"jsonrpc":"2.0","method":"tools/call","id":41,
       "params":{"name":"rust-counter-get-value","arguments":{}}}'

curl -sS -X POST "$URL" \
  -H "Authorization: Bearer $TOK" -H "Mcp-Session-Id: $SID" \
  -H "Accept: application/json, text/event-stream" \
  -d '{"jsonrpc":"2.0","method":"tools/call","id":42,
       "params":{"name":"counter-b-iso-get-value","arguments":{}}}'
```

**Expected**

```
Counter A increments:        1 2 3 4 5   ...then 6 7
Counter B increments:        1 2 3
Counter A final get_value:   7
Counter B final get_value:   3
```

**Actual output (verbatim from the probe run with this PR applied)**

```
=== STEP 1: initialize ===
sid=<sid>

=== STEP 2: increment Counter A × 5 (expect 1,2,3,4,5) ===
Counter A increments: 1 2 3 4 5

=== STEP 3: increment Counter B × 3 (expect 1,2,3 — independent of A) ===
Counter B increments: 1 2 3

=== STEP 4: increment Counter A × 2 more (expect 6,7) ===
Counter A more increments: 6 7

=== STEP 5: get_value both ===
Counter A final: "text":"7"
Counter B final: "text":"3"

=== verdict ===
✓ PASS — counters incremented independently (A=7, B=3)
```

**Structural verification**

- Redis directory has exactly **one** `mcpgw:pool_owner:{sid}` key — the single
  downstream session is owned by one worker:
  ```
  $ redis-cli --scan --pattern 'mcpgw:pool_owner:*'
  mcpgw:pool_owner:<sid>
  ```
- The two rust counter debug logs show **two distinct upstream session ids**:
  ```
  counter-a (rust-sdk on :9000)  upstream session id = <upstream-A>
  counter-b (rust-sdk on :9001)  upstream session id = <upstream-B>
  ```

One downstream `Mcp-Session-Id` → two independent upstream `ClientSession`s,
one per federated gateway. Same `(downstream_sid, gateway_id)` registry keying
routes each `tools/call` to the correct upstream regardless of which order the
calls arrive in. Without this PR's in-process dispatch on the four affinity
sites, a misrouted call could land on the wrong upstream and either bump the
other counter or spawn a duplicate session — neither was observed.

**Summary**

| Check | Expected | Actual | Verdict |
|---|---|---|---|
| Counter A sequence (5 + 2 interleaved increments) | 1, 2, 3, 4, 5, 6, 7 | 1, 2, 3, 4, 5, 6, 7 | ✓ |
| Counter B sequence (3 increments, between A's batches) | 1, 2, 3 | 1, 2, 3 | ✓ |
| Counter A final `get_value` | 7 | 7 | ✓ |
| Counter B final `get_value` | 3 | 3 | ✓ |
| Downstream session count in Redis (`pool_owner` keys) | 1 | 1 | ✓ |
| Distinct upstream session ids (one per federated gateway) | 2 | 2 | ✓ |

**Verdict: PASS.** Multi-gateway federation preserves the #4205 isolation invariant. One downstream `Mcp-Session-Id` correctly holds two independent upstream `ClientSession`s, one per federated gateway, and every `tools/call` is routed to the right upstream by the `(downstream_session_id, gateway_id)` registry keying — regardless of interleave order. Without this PR's in-process dispatch on the four affinity sites, a misrouted call could land on the wrong upstream and either bump the other counter or spawn a duplicate session; neither was observed.

</details>

<details>
<summary><b>Manual reproducer — cross-user isolation (two distinct tokens, one upstream counter, parallel sessions)</b></summary>

Verifies the #4205 isolation invariant in its purest form: two different
bearer tokens (two distinct users from the gateway's perspective) each
open their own MCP session against the **same** upstream counter server,
and the per-user state stays completely independent. Tests the
`(downstream_session_id, gateway_id)` registry keying when
`downstream_session_id` is the discriminator (both users hit one upstream
gateway, so only the downstream sid varies between the two upstream
sessions).

**Setup (delta from the reproducer above)**

- Single rust-sdk counter server is enough — just `127.0.0.1:9000`,
  registered as the federated gateway (e.g. `rust-counter`).
- Single virtual server scoping `rust-counter-increment` and
  `rust-counter-get-value`. The remainder assumes the virtual server id
  is in `$VIRTUAL_SERVER_ID`.
- Two distinct admin JWTs — same `admin@example.com` identity, different
  `--exp` values produce two different token strings:

  ```bash
  TOK_A=$(python -m mcpgateway.utils.create_jwt_token \
    --username admin@example.com --exp 1440 --secret "$JWT_SECRET_KEY")
  TOK_B=$(python -m mcpgateway.utils.create_jwt_token \
    --username admin@example.com --exp 720  --secret "$JWT_SECRET_KEY")
  ```

Everything else (Redis container, `make serve`) is identical to the
single-counter reproducer above.

**Test (curl, two users in parallel)**

```bash
URL="http://127.0.0.1:4444/servers/$VIRTUAL_SERVER_ID/mcp"

run_user() {
  local NAME="$1" TOK="$2" N=10

  # 1. initialize, capture sid
  local INIT SID
  INIT=$(curl -sS -D - -o /dev/null -X POST "$URL" \
    -H "Authorization: Bearer $TOK" \
    -H "Accept: application/json, text/event-stream" \
    -d "{\"jsonrpc\":\"2.0\",\"method\":\"initialize\",\"id\":1,
         \"params\":{\"protocolVersion\":\"2025-03-26\",\"capabilities\":{},
                     \"clientInfo\":{\"name\":\"$NAME\",\"version\":\"0.1\"}}}")
  SID=$(echo "$INIT" | awk -F': ' 'tolower($1)=="mcp-session-id"{print $2}' | tr -d '\r')

  # 2. notifications/initialized
  curl -sS -X POST "$URL" \
    -H "Authorization: Bearer $TOK" -H "Mcp-Session-Id: $SID" \
    -H "Accept: application/json, text/event-stream" \
    -d '{"jsonrpc":"2.0","method":"notifications/initialized"}'

  # 3. 10 increments
  for i in $(seq 1 $N); do
    curl -sS -X POST "$URL" \
      -H "Authorization: Bearer $TOK" -H "Mcp-Session-Id: $SID" \
      -H "Accept: application/json, text/event-stream" \
      -d "{\"jsonrpc\":\"2.0\",\"method\":\"tools/call\",\"id\":$((10+i)),
           \"params\":{\"name\":\"rust-counter-increment\",\"arguments\":{}}}"
  done

  # 4. get_value
  curl -sS -X POST "$URL" \
    -H "Authorization: Bearer $TOK" -H "Mcp-Session-Id: $SID" \
    -H "Accept: application/json, text/event-stream" \
    -d '{"jsonrpc":"2.0","method":"tools/call","id":99,
         "params":{"name":"rust-counter-get-value","arguments":{}}}'
}

# Run both users in parallel
run_user A "$TOK_A" &
run_user B "$TOK_B" &
wait
```

**Expected**

```
Both users: increments  1, 2, 3, …, 10   (each user's own monotonic sequence)
Both users: final get_value = 10          (independent — no leakage)
Two distinct Mcp-Session-Ids on the response headers
Two distinct upstream session ids in the rust counter's debug log
```

**Actual output (verbatim from the probe run with this PR applied, rust counter restarted just before the run for a clean log; sids redacted)**

```
=== User A ===
sid = <sid-A>
increments: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
final get_value = 10

=== User B ===
sid = <sid-B>
increments: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
final get_value = 10

=== verdict ===
User A 1..10 & final 10: True
User B 1..10 & final 10: True
sids distinct: True

✓ PASS
```

**Structural verification**

```
$ # Rust counter's debug log shows exactly two distinct upstream session ids
$ grep -oE 'streamable_http_session\{id="[0-9a-f-]+"\}' /tmp/counter-a.log | sort -u
streamable_http_session{id="<upstream-A>"}
streamable_http_session{id="<upstream-B>"}

Pre-test session count:        0
Post-test session count:       2
Net new sessions from this run: 2   (expected 2)
```

Two downstream `Mcp-Session-Id`s → two completely independent upstream
`ClientSession`s on the same upstream gateway, each with its own counter
state. Same `(downstream_sid, gateway_id)` registry keying as the
multi-upstream test, but here it is `downstream_sid` that varies while the
`gateway_id` stays fixed.

**Summary**

| Check | Expected | Actual | Verdict |
|---|---|---|---|
| Two distinct `Mcp-Session-Id`s | 2 | 2 | ✓ |
| User A increments | 1, 2, …, 10 | 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 | ✓ |
| User B increments | 1, 2, …, 10 | 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 | ✓ |
| User A final `get_value` | 10 | 10 | ✓ |
| User B final `get_value` (independent of A) | 10 | 10 | ✓ |
| Distinct upstream session ids on the rust counter | 2 | 2 | ✓ |
| Net new upstream sessions during this test | 2 | 2 | ✓ |

**Verdict: PASS.** Cross-user isolation holds on #4987. With no sticky-LB
hashing in play, two different users' requests can land on any gunicorn
worker, and the affinity layer plus the
`(downstream_session_id, gateway_id)` registry keying together guarantee
that each user's tool calls reach exactly the one upstream `ClientSession`
bound to their downstream session.

</details>

<details>
<summary><b>Manual reproducer — same-user multi-session (one token, three parallel sessions, one upstream counter)</b></summary>

Verifies that even though three concurrent MCP sessions share the **same**
bearer token (identical `Authorization` header), each `initialize` call
produces its own `Mcp-Session-Id` and binds its own independent upstream
`ClientSession`. Tests the `(downstream_session_id, gateway_id)` registry
keying when only `downstream_session_id` varies and the upstream
`gateway_id` and the auth identity both stay fixed.

This is the third axis of the keying — alongside the multi-upstream test
(one downstream sid, many gateway_ids) and the cross-user test (different
sids from different users, one gateway_id), this confirms isolation
holds when the only discriminator is the gateway-generated downstream sid
itself.

**Setup (delta from the cross-user reproducer above)**

- Same single rust-sdk counter on `127.0.0.1:9000`, same registered
  federated gateway, same single virtual server scoping
  `rust-counter-increment` and `rust-counter-get-value`.
- **One** admin JWT (no need for two — the point is that the same token
  is used for all sessions):

  ```bash
  TOK=$(python -m mcpgateway.utils.create_jwt_token \
    --username admin@example.com --exp 1440 --secret "$JWT_SECRET_KEY")
  ```

Everything else (Redis container, `make serve`) is identical.

**Test (curl, three sessions in parallel — same token)**

```bash
URL="http://127.0.0.1:4444/servers/$VIRTUAL_SERVER_ID/mcp"

run_session() {
  local NAME="$1" N=10

  # 1. initialize, capture sid
  local INIT SID
  INIT=$(curl -sS -D - -o /dev/null -X POST "$URL" \
    -H "Authorization: Bearer $TOK" \
    -H "Accept: application/json, text/event-stream" \
    -d "{\"jsonrpc\":\"2.0\",\"method\":\"initialize\",\"id\":1,
         \"params\":{\"protocolVersion\":\"2025-03-26\",\"capabilities\":{},
                     \"clientInfo\":{\"name\":\"$NAME\",\"version\":\"0.1\"}}}")
  SID=$(echo "$INIT" | awk -F': ' 'tolower($1)=="mcp-session-id"{print $2}' | tr -d '\r')

  # 2. notifications/initialized
  curl -sS -X POST "$URL" \
    -H "Authorization: Bearer $TOK" -H "Mcp-Session-Id: $SID" \
    -H "Accept: application/json, text/event-stream" \
    -d '{"jsonrpc":"2.0","method":"notifications/initialized"}'

  # 3. 10 increments
  for i in $(seq 1 $N); do
    curl -sS -X POST "$URL" \
      -H "Authorization: Bearer $TOK" -H "Mcp-Session-Id: $SID" \
      -H "Accept: application/json, text/event-stream" \
      -d "{\"jsonrpc\":\"2.0\",\"method\":\"tools/call\",\"id\":$((10+i)),
           \"params\":{\"name\":\"rust-counter-increment\",\"arguments\":{}}}"
  done

  # 4. get_value
  curl -sS -X POST "$URL" \
    -H "Authorization: Bearer $TOK" -H "Mcp-Session-Id: $SID" \
    -H "Accept: application/json, text/event-stream" \
    -d '{"jsonrpc":"2.0","method":"tools/call","id":99,
         "params":{"name":"rust-counter-get-value","arguments":{}}}'
}

# Three sessions in parallel, all under the same token
run_session S1 &
run_session S2 &
run_session S3 &
wait
```

**Expected**

```
Three sessions: each gets its own monotonic increments 1, 2, …, 10
Three sessions: each final get_value = 10  (independent — no leakage)
Three distinct Mcp-Session-Ids on the response headers
Three distinct upstream session ids in the rust counter's debug log
```

**Actual output (verbatim from the probe run with this PR applied, rust counter restarted just before the run for a clean log; sids redacted)**

```
=== Session S1 ===
sid = <sid-1>
increments: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
final get_value = 10

=== Session S2 ===
sid = <sid-2>
increments: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
final get_value = 10

=== Session S3 ===
sid = <sid-3>
increments: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
final get_value = 10

=== verdict ===
All three sessions hit 1..10 monotonic with final 10: True
All three sids distinct (count = 3): True

✓ PASS
```

**Structural verification**

```
$ # Rust counter's debug log shows exactly three distinct upstream session ids
$ grep -oE 'streamable_http_session\{id="[0-9a-f-]+"\}' /tmp/counter-a.log | sort -u
streamable_http_session{id="<upstream-1>"}
streamable_http_session{id="<upstream-2>"}
streamable_http_session{id="<upstream-3>"}

Pre-test session count:        0
Post-test session count:       3
Net new sessions from this run: 3   (expected 3)

$ # Redis directory: one pool_owner key per downstream sid
$ redis-cli --scan --pattern 'mcpgw:pool_owner:*' | wc -l
3
```

Three downstream `Mcp-Session-Id`s from the same token bind three completely
independent upstream `ClientSession`s on the same upstream gateway. Same
`(downstream_sid, gateway_id)` keying as the other two manual reproducers —
this time exercising the path where the auth identity is identical across
sessions and the *only* discriminator is the gateway-generated downstream sid.

**Summary**

| Check | Expected | Actual | Verdict |
|---|---|---|---|
| Three distinct `Mcp-Session-Id`s from one token | 3 | 3 | ✓ |
| Session S1 increments | 1, 2, …, 10 | 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 | ✓ |
| Session S2 increments | 1, 2, …, 10 | 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 | ✓ |
| Session S3 increments | 1, 2, …, 10 | 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 | ✓ |
| Each session's final `get_value` | 10 | 10 / 10 / 10 | ✓ |
| Distinct upstream session ids on the rust counter | 3 | 3 | ✓ |
| Net new upstream sessions during this test | 3 | 3 | ✓ |
| Distinct `pool_owner` keys in Redis | 3 | 3 | ✓ |

**Verdict: PASS.** Same-user multi-session isolation holds on #4987.
Three concurrent sessions from one token spread across whichever
gunicorn workers `least_conn` LB happened to pick; the affinity layer's
pub/sub forwarding plus the `(downstream_session_id, gateway_id)`
registry keying together guarantee that each session's tool calls reach
exactly the one upstream `ClientSession` bound to its downstream sid —
not the other two sessions' upstreams. If the gateway had treated
"same auth identity" as "same session", all three counters would have
ended at 30; if the affinity forwarding had scattered, the sequences
would have shown gaps. Neither happened.

</details>

<details>
<summary><b>Manual reproducer — worker-kill / failover (verify the affinity-layer recovery contract under owner-worker death)</b></summary>

Verifies the failure-mode contract when the gunicorn worker that owns a
downstream session dies abruptly (SIGKILL). The intent is **not** a
property this PR changes — the in-process ASGI dispatch fix affects the
happy-path execution after a forward succeeds, not the dead-owner path —
but the worker-kill behaviour on top of this PR's code is documented
here so reviewers can see the recovery contract is preserved and the
operational characteristics are clear.

**Setup**

Identical to the single-counter reproducer above. No new processes
needed; the test just identifies which gunicorn worker owns a bound
session and kills it.

**Test (bash)**

```bash
TOK=$(python -m mcpgateway.utils.create_jwt_token \
  --username admin@example.com --exp 60 --secret "$JWT_SECRET_KEY")
URL="http://127.0.0.1:4444/servers/$VIRTUAL_SERVER_ID/mcp"

# 1. Initialize + 5 increments (captures $SID and exercises ownership binding)
# ... (same initialize / notifications/initialized / 5 increment steps as the
#      single-counter reproducer above) ...

# 2. Find the gunicorn worker that owns this session
OWNER=$(redis-cli get "mcpgw:pool_owner:$SID")           # → '<hostname>:<pid>'
OWNER_PID=${OWNER##*:}

# 3. SIGKILL that worker
kill -KILL "$OWNER_PID"
# gunicorn master will respawn a replacement worker immediately

# 4. Send another request with the same (now-stale) sid; capture HTTP code + timing
curl -sS -o /dev/null -w "HTTP %{http_code} | time_total=%{time_total}s\n" \
  --max-time 40 -X POST "$URL" \
  -H "Authorization: Bearer $TOK" -H "Mcp-Session-Id: $SID" \
  -H "Accept: application/json, text/event-stream" \
  -d '{"jsonrpc":"2.0","method":"tools/call","id":101,
       "params":{"name":"rust-counter-increment","arguments":{}}}'

# 5. Send a fresh initialize and confirm the gateway accepts new sessions
# ... (standard initialize / notifications/initialized / 3 increments) ...
```

**Expected (the recovery contract)**

```
- SIGKILL on the owner gunicorn worker:  worker dies; gunicorn master respawns within seconds.
- Request with the stale-owner sid:      receiving worker tries to forward to the dead
                                         owner over pub/sub, times out after
                                         mcpgateway_pool_rpc_forward_timeout (default 30s),
                                         returns HTTP 404 + JSON-RPC -32600 "Session not found".
- Stale `mcpgw:pool_owner:{sid}` key:    persists until natural Redis TTL expiry (heartbeat
                                         reclaim is bounded by the key's TTL — ~3 minutes
                                         with the default config).
- Fresh initialize:                       binds cleanly to a different worker; new upstream
                                          session is opened; counter increments monotonically.
- Recovery contract for the dead sid:    client receives a clear error, re-initializes.
```

**Actual output (verbatim from the probe run with this PR applied; sids and PIDs redacted)**

```
=== STEP 1: SIGKILL the owner worker PID=<pid-A> ===
  PID COMMAND
  (no row — process is gone)

=== STEP 2: gunicorn respawn? ===
old owner PID alive? NO
workers under gunicorn master ppid currently: 16  (was 16 — replacement spawned)

=== STEP 3: pool_owner key + TTL over a 60s window after kill ===
right now: pool_owner='<host>:<pid-A>'  TTL=195
  +5s:   pool_owner='<host>:<pid-A>'  TTL=190
  +10s:  pool_owner='<host>:<pid-A>'  TTL=185
  +30s:  pool_owner='<host>:<pid-A>'  TTL=165
  +60s:  pool_owner='<host>:<pid-A>'  TTL=149
  (key did NOT expire within the 60s observation window — natural TTL is ~3 min)

=== STEP 4: retry the increment with the same (stale-owner) sid ===
HTTP 404 | time_total=30.019149s
wall: 30s
body: {"jsonrpc":"2.0","id":"server-error","error":{"code":-32600,"message":"Session not found"}}

=== STEP 5: fresh initialize + 3 increments (recovery sanity) ===
new sid=<sid-recovery>
  increment 1 → 1
  increment 2 → 2
  increment 3 → 3
```

**Structural verification**

```
$ # Counter A's upstream session ids over the whole run
$ grep -oE 'streamable_http_session\{id="[0-9a-f-]+"\}' /tmp/counter-a.log | sort -u
streamable_http_session{id="<upstream-original>"}     # bound by the pre-kill 5 increments
streamable_http_session{id="<upstream-stray>"}        # left by the failed-forward path during the 30s timeout
streamable_http_session{id="<upstream-recovery>"}     # bound by the post-kill fresh initialize

Total: 3
```

The original upstream session is orphaned by the worker death and is not
reused; the recovery `initialize` opens its own independent upstream
`ClientSession` on a different gunicorn worker. The stray middle session
is an artefact of the failed-forward attempt during the 30s timeout
window and is also orphaned (no further calls land on it).

**Summary**

| Check | Expected | Actual | Verdict |
|---|---|---|---|
| SIGKILL of owner worker | dies immediately | dies immediately | ✓ |
| Gunicorn master respawns replacement | yes, fast | replacement up within ~seconds | ✓ |
| Stale-sid request returns a structured error (not a hang, not 5xx) | HTTP 404 + JSON-RPC `-32600 "Session not found"` | exactly that | ✓ |
| Failure-window latency for the first request after worker death | ≤ `mcpgateway_pool_rpc_forward_timeout` (default 30s) | ~30s | ✓ |
| Fresh `initialize` after the failed retry | binds cleanly to a different worker | new sid, new upstream session, increments 1, 2, 3 | ✓ |
| Stale `mcpgw:pool_owner:{sid}` key cleanup | natural Redis TTL (~3 min default) | TTL decremented normally; not observed to expire within 60s | ✓ (as designed) |

**Verdict: PASS — recovery contract is preserved.** The gateway does not
crash, the failed-forward request gets a clean structured error rather
than a hang or 5xx, gunicorn auto-recovers the worker, the stale
`pool_owner` key is bounded by Redis TTL, and a fresh `initialize` works
immediately. Session-loss is correctly the client's responsibility
("client re-initializes after a session error") — the same contract that
applies to any sticky-routing failure mode.

**Operational note on the 30s failure window**

The first request that lands on the dead owner's sid after the kill
pays the `mcpgateway_pool_rpc_forward_timeout` (default 30s) before
returning `Session not found`. This timeout pre-dates this PR — it
governs the pub/sub forward path that this PR's in-process dispatch
fix does *not* touch. Operators who want a faster fail signal can
either tune this setting down (with awareness that legitimate slow
forwards under load also get truncated), or layer a faster-fail
check on the receiving worker that consults the dead-owner's
`mcpgw:worker_heartbeat` key before attempting the forward — a
separate follow-up worth exploring after this PR lands.

</details>

<details>
<summary><b>Load benchmark — Streamable HTTP tools throughput (16-core host, 526 RPS @ 0.00% failures)</b></summary>

Re-run on a dedicated 16-core / 32 GB host (Ubuntu 24.04) to measure the in-process
dispatch path under sustained concurrency without single-laptop VM saturation. The
image was built from this branch (`make docker-prod`) and the standard testing stack
(3 gateway replicas behind nginx, Postgres + Redis + pgbouncer) was brought up via
`make testing-up`.

**Config:** `make benchmark-mcp-tools` — tools-only MCP Streamable HTTP, 125 concurrent
users, 30/s spawn rate, 60s run, against the registered `fast_time` virtual server
through nginx on `:8080`.

**Results**

| Metric | Value |
|---|---|
| Total requests | 31,601 |
| Failures | 1 (0.00%) |
| Throughput | 526.29 RPS |
| Avg latency | 168.6 ms |
| p50 / p90 / p95 / p99 | 110 / 270 / 470 / 1,300 ms |
| Max | 30,001 ms (single client-side timeout = the lone failure) |

**Endpoint breakdown**

|---|---|---|---|---|---|
| `tools/call` [rapid] | 29,946 | 1 | 498.7 | 169.4 | 1,300 |
| `tools/list` [rapid] | 1,530 | 0 | 25.5 | 158.0 | 1,100 |
| `initialize` | 125 | 0 | 2.1 | 115.5 | 390 |

All 125 `initialize` calls succeeded (0 failures), confirming the session-affinity
binding path stays stable under load: the in-process `/_internal/mcp/rpc` dispatch
sustains ~526 RPS with effectively zero failures and a p99 of 1.3s.

</details>

<details>
<summary><b>#4205 counter reproducers — statefulness &amp; isolation under multi-worker affinity (5/5 PASS, 16-core host)</b></summary>

Re-ran the #4205 counter reproducers against this branch's stack on a 16-core host
(3 gateway replicas behind nginx, non-sticky round-robin, Redis pool ownership). A
minimal per-session counter MCP server is registered behind the gateway; a session
is incremented N times and must read back N. If forwarding scattered onto a fresh
upstream session, the counter would drop.

**Setup**

```python
# counter_server.py — per-session counter (state keyed by the upstream ServerSession)
mcp = FastMCP("repro-counter", host="0.0.0.0", port=9400)   # 2nd instance on :9401 for Test 4

@mcp.tool()

@mcp.tool()

```

1. Run the counter on the host (reachable from the gateway containers via
   `host.docker.internal`, mapped by compose `extra_hosts: host.docker.internal:host-gateway`).
4. Admin token: `create_jwt_token --username admin@example.com --admin --secret $JWT_SECRET_KEY`.
5. Drive sessions against `/servers/{id}/mcp` (`initialize` → `notifications/initialized` →
   `tools/call increment` ×N → `tools/call get_value`); concurrent sessions via a thread pool.

**Results**

| Test | Steps | Result |
|---|---|---|
| 1 — single session | one session, `increment` ×25, then `get_value` | `increments=[1…25]`, `get_value=25`, `pool_owner → host:pid` (pid ≠ 1 → per-worker WORKER_ID) — **PASS** |
| 2 — same-user multi-session | 3 concurrent sessions, one token, ×10 each | 3 distinct sids, each `[1…10]` / `get_value=10` → 3 independent `pool_owner` keys — **PASS** |
| 3 — two tokens | two distinct tokens (same admin), parallel sessions, ×10 | both `[1…10]` / `get_value=10`, distinct sids, no cross-talk — **PASS** |
| 4 — multi-upstream | one session across two counters (`:9400`+`:9401`): A×5, B×3, A×2 | A `[1,2,3,4,5]` then `[6,7]` (`get_value=7`), B `[1,2,3]` (`get_value=3`) → `(session, gateway)` isolation — **PASS** |
| 5 — owner-worker kill | bind + ×5, read owner from Redis, `kill -9` the owning gunicorn worker, hit stale sid, then fresh `initialize` | stale sid → **HTTP 404 / JSON-RPC `-32600 "Session not found"` in 30.0s** (not a hang/5xx); killed worker auto-respawns; fresh init `[1,2,3]` — **PASS** |

Test 5 trace:

```
[1] bound sid=fb8be982…  increments=[1, 2, 3, 4, 5]
[2] pool_owner -> 58f27b9c19e5:78        # owner pid 78 (not :1 -> per-worker WORKER_ID active)
[3] owner worker pid 78 in mcp-context-forge-gateway-1
[4] kill -9 78 -> worker dead (gunicorn respawns a replacement)
[6] fresh initialize 88834714… increments=[1, 2, 3]
```

session per downstream session, no scatter), isolation holds across sessions/tokens/upstreams,
and the owner-death failure mode returns a clean structured error rather than a hang.

</details>

## Notes

- Depends on #4981; review/merge that first.
- `httpx.ASGITransport` re-runs the middleware stack once per dispatch; a later
  optimization could call the RPC dispatcher directly to skip it, but profiling
  showed the dispatch is already cheap (~5-6 ms, dominated by the upstream call),
  so it is intentionally out of scope here.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -64,7 +64,7 @@ CSRF_ROTATE_ON_LOGIN=true
 CSRF_TRUSTED_ORIGINS=["http://localhost:4444","http://localhost:8080"]
 
 # Paths exempt from CSRF protection (JSON array)
-CSRF_EXEMPT_PATHS=["/health","/auth/login","/auth/logout","/auth/refresh","/auth/email/login","/auth/email/register","/auth/email/forgot-password","/auth/email/reset-password","/admin","/oauth/fetch-tools","/docs","/redoc","/openapi.json","/metrics","/mcp/","/sse","/message"]
+CSRF_EXEMPT_PATHS=["/health","/auth/login","/auth/logout","/auth/refresh","/auth/email/login","/auth/email/register","/auth/email/forgot-password","/auth/email/reset-password","/admin","/oauth/fetch-tools","/docs","/redoc","/openapi.json","/metrics","/mcp/","/sse","/message","/_internal/mcp/"]
 
 # Bootstrap admin credentials (email auth)
 # PRODUCTION: Change these values
--- a/mcpgateway/auth_context.py
+++ b/mcpgateway/auth_context.py
@@ -221,6 +221,25 @@ def get_internal_mcp_auth_context(request: Request) -> Optional[Dict[str, Any]]:
     return None
 
 
+def encode_internal_mcp_auth_context(auth_context: Dict[str, Any]) -> str:
+    """Encode an edge-validated auth context for forwarding to a trusted internal dispatcher.
+
+    Mirror of :func:`decode_internal_mcp_auth_context`. Packages the auth context
+    so a trusted internal dispatcher (e.g. ``/_internal/mcp/rpc``) can use it
+    without re-running authentication. The unpadded base64url shape keeps the
+    value header-safe.
+
+    Args:
+        auth_context: Auth-context dict to encode.
+
+    Returns:
+        Unpadded base64url-encoded JSON payload for the
+        ``x-contextforge-auth-context`` header.
+    """
+    encoded = base64.urlsafe_b64encode(orjson.dumps(auth_context)).decode("ascii")
+    return encoded.rstrip("=")
+
+
 def decode_internal_mcp_auth_context(header_value: str) -> Dict[str, Any]:
     """Decode the trusted internal MCP auth header payload.
 
@@ -289,6 +308,94 @@ def has_valid_internal_mcp_runtime_auth_header(request: Request) -> bool:
     if not provided:
         return False
     return hmac.compare_digest(provided, _expected_internal_mcp_runtime_auth_header())
+
+
+# Internal-dispatch trust gate, defined once and shared by all callers.
+INTERNAL_MCP_PATH_PREFIX = "/_internal/mcp"
+INTERNAL_A2A_PATH_PREFIX = "/_internal/a2a"
+INTERNAL_RUNTIME_MARKER_HEADER = "x-contextforge-mcp-runtime"
+INTERNAL_AUTH_CONTEXT_HEADER = "x-contextforge-auth-context"
+TRUSTED_INTERNAL_RUNTIME_MARKERS = frozenset({"rust", "affinity"})
+
+
+def _internal_path_requires_auth_context(path: str) -> bool:
+    """Whether an internal route requires an auth context.
+
+    Only ``*/authenticate`` is exempt, since it creates the context; every
+    other internal route must carry one.
+
+    Args:
+        path: The request path to classify.
+
+    Returns:
+        ``False`` for ``*/authenticate``, otherwise ``True``.
+    """
+    return not path.rstrip("/").endswith("/authenticate")
+
+
+def is_trusted_internal_runtime_request(
+    request: Request,
+    *,
+    allowed_prefixes: tuple[str, ...],
+    require_auth_context: bool,
+    path: Optional[str] = None,
+) -> bool:
+    """Return whether a request is a trusted in-process internal-runtime hop.
+
+    The trust boundary is the HMAC header and, when required, the encoded
+    ``x-contextforge-auth-context``. The loopback check is defense in depth,
+    not an independent gate: ProxyHeaders(trusted_hosts="*") lets a direct
+    external caller influence ``request.client.host`` (the replay is hardened
+    separately by stripping forwarded / client-IP headers).
+
+    Args:
+        request: Incoming request to inspect.
+        allowed_prefixes: Internal path prefixes this caller trusts.
+        require_auth_context: Require a non-empty ``x-contextforge-auth-context``.
+        path: Explicit path override for callers that strip a ``root_path``;
+            defaults to ``request.url.path``.
+
+    Returns:
+        ``True`` only when prefix, runtime marker, HMAC, optional auth context,
+        and loopback all hold; otherwise ``False``.
+    """
+    p = path if path is not None else (getattr(getattr(request, "url", None), "path", "") or "")
+    if not any(p == prefix or p.startswith(f"{prefix}/") for prefix in allowed_prefixes):
+        return False
+    if request.headers.get(INTERNAL_RUNTIME_MARKER_HEADER) not in TRUSTED_INTERNAL_RUNTIME_MARKERS:
+        return False
+    if not has_valid_internal_mcp_runtime_auth_header(request):
+        return False
+    if require_auth_context and not request.headers.get(INTERNAL_AUTH_CONTEXT_HEADER):
+        return False
+    client_host = getattr(getattr(request, "client", None), "host", None)
+    return client_host in ("127.0.0.1", "::1")
+
+
+def is_trusted_internal_mcp_request(request: Request, *, path: Optional[str] = None) -> bool:
+    """MCP + A2A internal trust gate with a path-aware auth-context requirement.
+
+    ``*/authenticate`` routes are exempt from the auth-context requirement;
+    every other internal route requires it. ``/_internal/a2a/*`` is trusted
+    only when the A2A feature is enabled.
+
+    Args:
+        request: Incoming request to inspect.
+        path: Explicit path override (e.g. a ``root_path``-stripped path);
+            defaults to ``request.url.path``.
+
+    Returns:
+        ``True`` when the request is a trusted internal MCP/A2A hop.
+    """
+    p = path if path is not None else (getattr(getattr(request, "url", None), "path", "") or "")
+    if p.startswith(f"{INTERNAL_A2A_PATH_PREFIX}/") and not settings.mcpgateway_a2a_enabled:
+        return False
+    return is_trusted_internal_runtime_request(
+        request,
+        allowed_prefixes=(INTERNAL_MCP_PATH_PREFIX, INTERNAL_A2A_PATH_PREFIX),
+        require_auth_context=_internal_path_requires_auth_context(p),
+        path=p,
+    )
 
 
 def get_token_teams_from_request(request: Request) -> Optional[List[str]]:
--- a/mcpgateway/config.py
+++ b/mcpgateway/config.py
@@ -458,6 +458,7 @@ class Settings(BaseSettings):
             "/sse",  # Exempt: SSE is a server-sent event stream, not vulnerable to CSRF
             "/message",  # Exempt: MCP SSE message endpoint
             "/rpc",  # Exempt: JSON-RPC is a programmatic protocol, not browser-based
+            "/_internal/mcp/",  # Exempt: loopback-only, HMAC-gated internal dispatch (affinity/Rust forwards); not browser-reachable
         ],
         description="Paths exempt from CSRF protection",
     )
--- a/mcpgateway/main.py
+++ b/mcpgateway/main.py
@@ -81,8 +81,8 @@ from mcpgateway.auth_context import (
     get_scoped_resource_access_context,
     get_token_teams_from_request,
     get_user_email,
-    has_valid_internal_mcp_runtime_auth_header,
     INTERNAL_MCP_SESSION_VALIDATED_HEADER,
+    is_trusted_internal_mcp_request,
 )
 from mcpgateway.cache import ResourceCache, SessionRegistry
 from mcpgateway.common.models import InitializeResult
@@ -277,26 +277,26 @@ _INTERNAL_MCP_AUTH_CONTEXT_HEADER = "x-contextforge-auth-context"
 
 
 def _is_trusted_internal_mcp_runtime_request(request: Request) -> bool:
-    """Return whether the request came from the local Rust runtime sidecar.
+    """Return whether the request came from a trusted local internal source.
+
+    Two callers are trusted today:
+
+    - ``"rust"`` — the local Rust runtime sidecar (over loopback).
+    - ``"affinity"`` — the in-process dispatch used by session-affinity
+      forwarding to reach the owner worker, carrying the identity the edge
+      already validated.
+
+    Both share the same gates: a shared-secret HMAC header AND a loopback client
+    address. Only the ``x-contextforge-mcp-runtime`` marker value differs.
 
     Args:
         request: Incoming request to inspect.
 
     Returns:
-        ``True`` when the request carries the trusted Rust runtime marker from
-        loopback, otherwise ``False``.
+        ``True`` when the request carries a trusted internal-runtime marker
+        from loopback, otherwise ``False``.
     """
-    runtime_marker = request.headers.get("x-contextforge-mcp-runtime")
-    client_host = getattr(getattr(request, "client", None), "host", None)
-    if runtime_marker != "rust" or not has_valid_internal_mcp_runtime_auth_header(request) or client_host not in ("127.0.0.1", "::1"):
-        return False
-    # Defense-in-depth: /_internal/a2a/* endpoints must refuse requests when
-    # A2A support is disabled, even from an otherwise-trusted local sidecar.
-    # A legitimate sidecar should not be running when the feature is off.
-    path = getattr(getattr(request, "url", None), "path", "") or ""
-    if path.startswith("/_internal/a2a/") and not settings.mcpgateway_a2a_enabled:
-        return False
-    return True
+    return is_trusted_internal_mcp_request(request)
 
 
 def _is_jwt_token(token: str) -> bool:
--- a/mcpgateway/middleware/rate_limit_middleware.py
+++ b/mcpgateway/middleware/rate_limit_middleware.py
@@ -38,6 +38,7 @@ from starlette.responses import JSONResponse
 
 # First-Party
 from mcpgateway import auth
+from mcpgateway.auth_context import is_trusted_internal_mcp_request
 from mcpgateway.config import settings
 from mcpgateway.services.security_logger import SecurityEventType, SecurityLogger, SecuritySeverity
 
@@ -197,6 +198,10 @@ class RateLimitMiddleware(BaseHTTPMiddleware):
     async def dispatch(self, request: Request, call_next):
         """Process request with rate limiting."""
         if not self.enabled:
+            return await call_next(request)
+
+        # Skip rate limiting for the trusted-internal dispatch; the edge request was already counted.
+        if is_trusted_internal_mcp_request(request):
             return await call_next(request)
 
         tier = self.get_endpoint_tier(request.url.path)
--- a/mcpgateway/middleware/token_scoping.py
+++ b/mcpgateway/middleware/token_scoping.py
@@ -13,8 +13,6 @@ and time-based restrictions.
 # Standard
 from datetime import datetime, timedelta, timezone
 from functools import lru_cache
-import hashlib
-import hmac
 import ipaddress
 import re
 from typing import List, Optional, Pattern, Tuple
@@ -64,11 +62,6 @@ _RESOURCE_PATTERNS: List[Tuple[Pattern[str], str]] = [
     (re.compile(r"/gateways/?([a-f0-9\-]+)"), "gateway"),
 ]
 _AUTH_COOKIE_NAMES = ("jwt_token", "access_token")
-_INTERNAL_MCP_PATH_PREFIX = "/_internal/mcp"
-_INTERNAL_MCP_RUNTIME_HEADER = "x-contextforge-mcp-runtime"
-_INTERNAL_MCP_AUTH_CONTEXT_HEADER = "x-contextforge-auth-context"
-_INTERNAL_MCP_RUNTIME_AUTH_HEADER = "x-contextforge-mcp-runtime-auth"
-_INTERNAL_MCP_RUNTIME_AUTH_CONTEXT = "contextforge-internal-mcp-runtime-v1"
 
 # Permission map with precompiled patterns
 # Maps (HTTP method, path pattern) to required permission
@@ -1350,70 +1343,24 @@ class TokenScopingMiddleware:
             )
 
     def _is_trusted_internal_mcp_runtime_request(self, request: Request, normalized_path: str) -> bool:
-        """Return whether the request is a trusted loopback Rust MCP sidecar hop.
+        """Return whether the request is a trusted internal MCP/A2A runtime hop.
+
+        Delegates to the shared gate in ``auth_context`` so the trust decision is
+        defined in one place. Unlike the previous local check, this trusts both
+        the ``rust`` and ``affinity`` runtime markers — closing the gap where the
+        in-process session-affinity dispatch was still token-scoped.
 
         Args:
             request: Incoming HTTP request.
             normalized_path: Canonicalized request path used for route matching.
 
         Returns:
-            ``True`` when the request originated from the local Rust MCP runtime and
-            includes the expected trusted headers.
+            ``True`` when the request is a trusted internal MCP/A2A hop.
         """
-        if normalized_path != _INTERNAL_MCP_PATH_PREFIX and not normalized_path.startswith(f"{_INTERNAL_MCP_PATH_PREFIX}/"):
-            return False
+        # First-Party
+        from mcpgateway.auth_context import is_trusted_internal_mcp_request  # pylint: disable=import-outside-toplevel
 
-        if request.headers.get(_INTERNAL_MCP_RUNTIME_HEADER) != "rust":
-            return False
-
-        provided_auth = request.headers.get(_INTERNAL_MCP_RUNTIME_AUTH_HEADER)
-        if not provided_auth:
-            return False
-
-        expected_auth = self._expected_internal_mcp_runtime_auth_header()
-        if not hmac.compare_digest(provided_auth, expected_auth):
-            return False
-
-        if not request.headers.get(_INTERNAL_MCP_AUTH_CONTEXT_HEADER):
-            return False
-
-        client_host = getattr(getattr(request, "client", None), "host", None)
-        return client_host in ("127.0.0.1", "::1")
-
-    @staticmethod
-    def _auth_encryption_secret_value() -> str:
-        """Return the configured auth-encryption secret as a plain string.
-
-        Returns:
-            The auth-encryption secret, normalized to a regular string.
-        """
-        secret = settings.auth_encryption_secret
-        if hasattr(secret, "get_secret_value"):
-            return secret.get_secret_value()
-        return str(secret)
-
-    @staticmethod
-    @lru_cache(maxsize=8)
-    def _expected_internal_mcp_runtime_auth_header_for_secret(secret: str) -> str:
-        """Return the expected shared internal-auth header for a specific secret.
-
-        Args:
-            secret: Auth-encryption secret to derive the trust header from.
-
-        Returns:
-            Hex-encoded SHA-256 digest derived from the provided auth secret.
-        """
-        material = f"{secret}:{_INTERNAL_MCP_RUNTIME_AUTH_CONTEXT}".encode("utf-8")
-        return hashlib.sha256(material).hexdigest()
-
-    @staticmethod
-    def _expected_internal_mcp_runtime_auth_header() -> str:
-        """Return the expected shared internal-auth header for Rust MCP hops.
-
-        Returns:
-            Shared secret-derived digest expected on trusted internal Rust MCP calls.
-        """
-        return TokenScopingMiddleware._expected_internal_mcp_runtime_auth_header_for_secret(TokenScopingMiddleware._auth_encryption_secret_value())
+        return is_trusted_internal_mcp_request(request, path=normalized_path)
 
 
 # Create middleware instance
--- a/mcpgateway/middleware/token_usage_middleware.py
+++ b/mcpgateway/middleware/token_usage_middleware.py
@@ -30,6 +30,7 @@ from starlette.requests import Request
 from starlette.types import ASGIApp, Receive, Scope, Send
 
 # First-Party
+from mcpgateway.auth_context import is_trusted_internal_mcp_request
 from mcpgateway.db import fresh_db_session
 from mcpgateway.middleware.path_filter import should_skip_auth_context
 from mcpgateway.services.token_catalog_service import TokenCatalogService
@@ -84,6 +85,11 @@ class TokenUsageMiddleware:
         # Skip health checks and static files
         path = scope.get("path", "")
         if should_skip_auth_context(path):
+            await self.app(scope, receive, send)
+            return
+
+        # Skip usage logging for the trusted-internal dispatch; the edge request is already recorded.
+        if is_trusted_internal_mcp_request(Request(scope), path=path):
             await self.app(scope, receive, send)
             return
 
--- a/mcpgateway/services/session_affinity.py
+++ b/mcpgateway/services/session_affinity.py
@@ -50,7 +50,7 @@ from mcpgateway.services.upstream_session_registry import (  # re-exported as th
 )
 from mcpgateway.utils.internal_http import (
     internal_loopback_base_url,
-    internal_loopback_verify,
+    post_rpc_in_process,
 )
 
 # Shared session-id validation (downstream MCP session IDs used for affinity).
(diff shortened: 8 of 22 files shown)


gandhipratik203:
The consolidated changes here get merged to the integration branch.

PR #5000, opened 2026-06-01
fix: skip idle timeout for API tokens; fix is_admin fallback chain

### LINKED ISSUE #4998: [BUG]: API tokens auto-revoked by idle timeout due to unsafe iat fallback
## Bug Summary

Integration/API tokens with `token_use: "api"` are auto-revoked with reason `idle_timeout` after 60+ minutes of inactivity, even though they have valid `exp` claims (e.g., expiring 2027). The token creation date (`iat`) is incorrectly used as the "last activity" timestamp when the Redis activity tracking key expires.

## Affected Component

- [x] `mcpgateway` API — `mcpgateway/auth.py:1714`

## Steps to Reproduce

1. Create an API token via Admin UI or API
2. Use the token successfully (Redis activity key is set)
3. Wait >60 minutes without using the token (Redis key expires at TTL = `token_idle_timeout * 60`)
4. Make a new request with the token
5. Token is rejected with 401 "Token exceeded idle timeout" and added to blocklist

## Expected Behavior

API tokens should not be subject to the session-style 60-minute idle timeout. They have their own expiry (`expires_at`) and revocation (`TokenRevocation`) mechanisms. At minimum, the idle timeout should not use `iat` as a fallback when Redis activity tracking is unavailable.

## Logs / Error Output

```
WARNING: Token exceeded idle timeout: jti=..., idle_minutes=47520.0
```

## Environment Info

| Key | Value |
|-----|-------|
| Version/commit | Any with `83d2142ff` or later |
| Runtime | All |
| Platform/OS | All |

## Workaround

Set `TOKEN_IDLE_TIMEOUT=0` to disable idle timeout entirely.

## Root Cause

The idle timeout enforcement in `get_current_user()` (`auth.py:1693-1741`) applies the same 60-minute idle window to **all** JWT tokens — including API tokens (`token_use: "api"`). The fallback chain for `last_activity` is:

1. Redis `token:activity:{jti}` key (TTL = `token_idle_timeout * 60` seconds, default 1 hour)
2. JWT `last_activity` claim (not set for API tokens — only session tokens include it)
3. **JWT `iat` claim** ← bug: creation date is not a valid activity signal

When Redis key expires, fallback to `iat` (creation date) calculates ~30 days of "idle" for a 30-day-old token, causing immediate revocation.

### LINKED ISSUE #4999: [CHORE]: is_admin fallback chain uses OR instead of clear precedence
## Chore Summary

Two locations in `mcpgateway/auth.py` check `is_admin` using a fallback chain that ORs values from `payload["is_admin"]` and `payload["user"]["is_admin"]`. When both exist with conflicting values, the `or` operator silently promotes to admin.

## Area Affected

- [x] code quality / consistency

## Context / Rationale

At `auth.py:1194` and `auth.py:1767`:

```python
is_admin=bool(payload.get("is_admin", False) or (payload.get("user") or {}).get("is_admin", False))
```

If `payload["is_admin"]` is explicitly `False` but `payload["user"]["is_admin"]` is `True`, the result is `True`. While both values are server-set and not user-controllable, this inconsistency makes the code fragile for audit.

Line 569 already handles this correctly with a two-step check:

```python
is_admin = payload.get("is_admin", False)
if not is_admin:
    user_info = payload.get("user", {})
    is_admin = user_info.get("is_admin", False) if isinstance(user_info, dict) else False
```

Lines 1194 and 1767 should follow the same pattern.

## Related Make Targets

- `make black`, `make ruff`, `make mypy`

## Acceptance Criteria

- [ ] `is_admin` uses `payload["is_admin"]` as authoritative source with two-step fallback (matching line 569 pattern)
- [ ] Lines 1194 and 1767 are updated consistently
- [ ] `make ruff`, `make mypy`, `make black` pass

MohanLaksh:
## Summary

Two fallback-pattern fixes in `mcpgateway/auth.py`:

### Fix 1: Idle timeout no longer applies to API tokens (Closes #4998)

**Problem:** The idle timeout check used the token creation date (`iat`) as a fallback "last activity" timestamp when Redis tracking was unavailable. For API tokens (which don't have a `last_activity` JWT claim), this meant tokens with `iat` older than `token_idle_timeout` (default 60 min) were auto-revoked with reason `idle_timeout` on the next use after Redis key expiry.

**Changes:**
- Added `token_use != "api"` guard (`auth.py:1703`) so API tokens skip the idle timeout entirely
- Removed unsafe `iat` fallback (`auth.py:1714`) — when no Redis activity and no `last_activity` claim exists, the idle check is skipped rather than using creation date as a proxy for recent activity
- `iat` was semantically wrong: creation date is not evidence of recent activity

### Fix 2: `is_admin` fallback follows clear precedence (Closes #4999)

**Problem:** Lines 1194 and 1767 used `payload.get("is_admin", False) or payload.get("user", {}).get("is_admin", False)` — an OR that silently promotes if either value is `True`. Line 569 already used the correct two-step pattern.

**Changes:**
- Both locations now resolve `is_admin` with payload-level claim authoritative, falling back to nested `user` dict only when the payload claim is `False`

### Tests

- Updated existing idle-timeout test to reflect removed `iat` fallback (idle check skipped when no activity source available)
- Added `TestIdleTimeoutApiToken` class verifying API tokens with 30-day-old `iat` survive Redis key expiry
- Verified session token idle timeout enforcement still works

## Checklist
- [x] `make ruff` — clean
- [x] `make mypy` — no new errors
- [x] `make test` — all 292 tests passing (test_auth_idle_timeout + test_auth)
- [x] Conventional commit with DCO sign-off

### DIFF
--- a/mcpgateway/auth.py
+++ b/mcpgateway/auth.py
@@ -1187,11 +1187,18 @@ def _bootstrap_platform_admin_user(email: str, payload: dict) -> "EmailUser":
     is_admin is derived from the token's own claim (default False) so that
     the bootstrap path grants login access without unconditional admin elevation.
     """
+    # Resolve is_admin: payload-level claim is authoritative; fall back to nested user dict.
+    resolved_is_admin = payload.get("is_admin", False)
+    if not resolved_is_admin:
+        user_info = payload.get("user", {})
+        if isinstance(user_info, dict):
+            resolved_is_admin = user_info.get("is_admin", False)
+
     return EmailUser(
         email=email,
         password_hash="",  # nosec B106 - not used for JWT authentication
         full_name=getattr(settings, "platform_admin_full_name", "Platform Administrator"),
-        is_admin=bool(payload.get("is_admin", False) or (payload.get("user") or {}).get("is_admin", False)),
+        is_admin=resolved_is_admin,
         is_active=True,
         auth_provider="local",
         password_change_required=False,
@@ -1766,7 +1773,13 @@ async def get_current_user(
         token_use = payload.get("token_use")
         if token_use == "session":  # nosec B105 - Not a password; token_use is a JWT claim type
             # Session token: resolve teams from DB/cache (fallback path — separate query OK)
-            user_info = {"is_admin": payload.get("is_admin", False) or payload.get("user", {}).get("is_admin", False)}
+            # Resolve is_admin: payload-level claim is authoritative; fall back to nested user dict.
+            resolved_is_admin = payload.get("is_admin", False)
+            if not resolved_is_admin:
+                user_info_nested = payload.get("user", {})
+                if isinstance(user_info_nested, dict):
+                    resolved_is_admin = user_info_nested.get("is_admin", False)
+            user_info = {"is_admin": resolved_is_admin}
             normalized_teams = await resolve_session_teams(payload, email, user_info)
         else:
             # API token or legacy: use embedded teams
--- a/mcpgateway/auth.py
+++ b/mcpgateway/auth.py
@@ -1692,13 +1692,15 @@ async def get_current_user(
 
                 # Idle timeout enforcement.
                 #
+                # Skipped for API tokens (token_use="api") which have their own
+                # expiry and revocation mechanisms via EmailApiToken / TokenRevocation.
+                #
                 # Activity source-of-truth precedence (most-recent first):
                 #   1. Redis key `token:activity:{jti}` written by `update_activity()`.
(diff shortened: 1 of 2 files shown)


wayneleoneIRL:
@ja8zyjits @brian-hussey can you review this today and let me kniow if we can include in our current sprint? Thanks.

brian-hussey (approved):
Approved, nice cleanup on the admin logic and handling the api token_use.

PR #5003, opened 2026-06-02
PR: Fix bootstrap_resource_assignments race condition on concurrent pod restart

### LINKED ISSUE #4993: [BUG]: bootstrap_resource_assignments UniqueViolation on concurrent pod restart
### 🐞 Bug Summary
`bootstrap_resource_assignments` raises `UniqueViolation` on `uq_team_owner_email_name_tool` during normal pod restarts when multiple workers start concurrently. The fast-path code skips the advisory lock, allowing concurrent workers to race on assigning the same orphaned resources.

---

### 🧩 Affected Component

- [x] `mcpgateway` - API
- [ ] Other (explain below)

---

### 🔁 Steps to Reproduce

1. Deploy ContextForge with PostgreSQL and multiple gunicorn workers (or multiple pod replicas)
2. Ensure there are orphaned tools in the database (tools with `team_id IS NULL` or `owner_email IS NULL` or `visibility IS NULL`)
3. Restart the pod (or scale up replicas) — all workers start simultaneously
4. All workers hit the fast path (`_is_at_alembic_head()` returns True), skip the advisory lock, and run `bootstrap_resource_assignments` concurrently
5. Multiple workers find the same orphaned tools, attempt to assign them to the same admin team with the same name, and one worker hits the UniqueViolation

---

### 🤔 Expected Behavior
`bootstrap_resource_assignments` should be safe to run concurrently from multiple workers without raising constraint violations. Either:
- The advisory lock should also protect the bootstrap helpers on the fast path, or
- The function should handle `IntegrityError`/`UniqueViolation` gracefully (catch and skip), or
- Use row-level locking (`SELECT ... FOR UPDATE SKIP LOCKED`) so workers do not compete for the same orphaned rows

---

### 📓 Logs / Error Output
```
ERROR [mcpgateway.bootstrap_db] Failed to assign tools: (psycopg.errors.UniqueViolation) duplicate key value violates unique constraint "uq_team_owner_email_name_tool"
DETAIL:  Key (team_id, owner_email, name)=(b540c656d92d412c98362f07557f59fa, admin@example.com, solace-dev-create-event-version) already exists.
[SQL: UPDATE tools SET updated_at=%(updated_at)s::TIMESTAMP WITH TIME ZONE, federation_source=%(federation_source)s::VARCHAR, team_id=%(team_id)s::VARCHAR, owner_email=%(owner_email)s::VARCHAR, visibility=%(visibility)s::VARCHAR WHERE tools.id = %(tools_id)s::VARCHAR]
```

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Version or commit | v0.7.0+ |
| Runtime | Python 3.11+, Gunicorn with multiple workers |
| Platform / OS | Kubernetes / Linux |
| Container | Docker |

---

**Root cause analysis:**

The issue is in `mcpgateway/bootstrap_db.py`. The `main()` function has two paths:

1. **Schema NOT at head** — acquires advisory lock, runs migrations + bootstrap helpers (serialized, safe)
2. **Schema at head (fast path)** — skips advisory lock, runs bootstrap helpers directly (concurrent, unsafe)

On a normal restart (no new migrations), all workers take the fast path simultaneously and run `bootstrap_resource_assignments` without any coordination. If orphaned resources exist (any of `team_id`, `owner_email`, or `visibility` is NULL), multiple workers query the same orphans and race to assign them.

The `existing_taken` deduplication set and `batch_assigned` tracking only protect against intra-worker conflicts within a single process — they provide no cross-process protection.

**Suggested fix options (in order of preference):**

1. Wrap the fast-path bootstrap helpers in the same advisory lock used for migrations (simplest, serializes startup slightly but is safe)
2. Add per-row `try/except IntegrityError` with retry-after-rename logic inside the inner loop
3. Use `SELECT ... FOR UPDATE SKIP LOCKED` when querying orphaned resources so concurrent workers skip rows already being processed

The error is currently non-fatal (caught by the per-resource-type `except` block on line ~723), but it means the losing worker's entire tools batch is rolled back. Those tools remain orphaned until another restart where only one worker processes them.

rakdutta:
## Summary

closes  #4993 - Prevents race conditions when multiple gateway pods concurrently assign orphaned resources during startup, while preserving the PgBouncer compatibility fix from #4051.

## Problem

When multiple replicas restart simultaneously (e.g., Kubernetes rolling deployment), they race when assigning orphaned resources to the admin team, causing `IntegrityError` crashes due to unique constraint violations.

**Why not use advisory locks?**  
Issue #4051 identified that advisory locks on the fast path cause indefinite hangs with PgBouncer in transaction pooling mode. Session-scoped locks get orphaned across backend handoffs, causing pods to spin for ~10 minutes until timeout. This was fixed in commit e4b245fc4 by skipping advisory locks on the fast path.

## Solution

Use **per-row commits with `IntegrityError` exception handling** instead of locks:

```python
for resource in unassigned:
    resource.team_id = personal_team.id
    # ... set other fields ...
    
    try:
        db.commit()  # Per-row commit
        assigned_count += 1
    except IntegrityError:
        db.rollback()  # Another worker assigned it - skip gracefully
        continue
```

**Key insight:** Let the database enforce uniqueness constraints. Handle exceptions gracefully instead of preventing them with locks.

## Changes

- **`mcpgateway/bootstrap_db.py`**: Added `IntegrityError` import, changed batch commit to per-row commits with exception handling
- **`tests/unit/mcpgateway/test_bootstrap_db.py`**: Updated test to verify fast path remains lock-free

## Testing

**Concurrency test** (`scripts/test_issue_4993_fix.sh`):
- 5 test runs × 30 concurrent workers × 30 orphaned resources
- **Result:** ✅ 0 races detected (0/5 failures)

**Unit tests:**
- ✅ All 60 tests pass
- Fast path test validates no advisory lock acquisition (preserves #4051 fix)

## Performance Impact

- **Old:** ~3 database round-trips (batch commit)
- **New:** ~N+2 round-trips (N per-row commits)
- **Impact:** +5-10 seconds for 30 resources during startup
- **Acceptable because:** Bootstrap-only code (not request path), N=0 after first boot

| Approach | Performance | PgBouncer Safe | Race Safe |
|----------|-------------|----------------|-----------|
| Batch commit | Fast (3 ops) | ✅ | ❌ |
| Advisory lock | Fast (3 ops) | ❌ (#4051) | ✅ |
| **Per-row commits** | **Slower (N+2)** | **✅** | **✅** |

## Deployment Impact

- ✅ No configuration changes required
- ✅ Maintains PgBouncer compatibility (issue #4051 fix preserved)
- ✅ Works on PostgreSQL, SQLite, MySQL
- ✅ Safe for Kubernetes multi-replica deployments

## Related Issues

- closes  #4993 - Race condition in bootstrap_resource_assignments
- Preserves fix for #4051 - PgBouncer transaction-pool hang
- Related to commit e4b245fc4 - Original PgBouncer fast-path fix


scripts/test_issue_4993_fix.sh
```diff
@@ -0,0 +1,204 @@
+#!/bin/bash
+# Testing script for Issue #4993 fix
```
prakhar-singh1928: @rakdutta Delete the shell script
unit tests are sufficient for coverage.

mcpgateway/bootstrap_db.py
```diff
@@ -757,17 +760,44 @@ def _like_safe(v: str) -> str:
...
+                                batch_assigned.discard(added_to_batch)
+
+                            logger.debug(
+                                f"Skipping {SecurityValidator.sanitize_log_message(resource_name)} "
+                                f"'{SecurityValidator.sanitize_log_message(str(original_value))}' "
+                                f"- already assigned by concurrent worker"
+                            )
+                            continue
+
+                    total_assigned += assigned_count
```
prakhar-singh1928, line 803: Optional improvement. Worth adding a comment here something like - 

```
 # Per-row commits mean some rows may already be persisted if a non-IntegrityError
  # is raised mid-loop; those rows are not counted here and won't be retried.
```

prakhar-singh1928 (dismissed):
Thanks @rakdutta . the approach is solid. Per-row commit with IntegrityError catch is exactly the right call given the PgBouncer constraint from #4051, and db.expunge() + batch_assigned.discard() are handled correctly. The unit tests cover all three cases well.

Two small things to address before this is ready to merge:

- Delete scripts/test_issue_4993_fix.sh
- (Optional)  Add a one-line comment above except Exception as e something that addresses the behaviour that the outer handler was written for the old batch-commit model. Now that commits happen per-row, a reader needs to know that a non-IntegrityError mid-loop leaves some rows already persisted but uncounted.

scripts/test_issue_4993_fix.sh
rakdutta: done

rakdutta:
@prakhar-singh1928 Thanks for the thorough review! I've addressed both items:

1. ✅ Deleted `scripts/test_issue_4993_fix.sh`
2. ✅ Added the clarifying comment before `total_assigned += assigned_count` explaining that per-row commits mean some rows may already be persisted if a non-IntegrityError is raised mid-loop

Both changes are committed and ready for re-review.

prakhar-singh1928 (dismissed):
LGTM

ja8zyjits (dismissed):
LGTM

prakhar-singh1928 (dismissed):

prakhar-singh1928 (dismissed):

ja8zyjits (dismissed):
LGTM

prakhar-singh1928 (approved):
LGTM

ja8zyjits (approved):
LGTM

PR #5008, opened 2026-06-02
fix(cache): team cache hardening - cross-worker eviction, update_team, transient ORM, nullable safety

### LINKED ISSUE #4884: [CHORE]: Follow-up hardening for full-object team cache (PR #4550)
## Context

PR #4550 introduced a full-object L1/L2 cache for `get_user_teams()`, eliminating the extra DB query on cache hits. The change is a net improvement, but the reviewer flagged several follow-up concerns worth tracking. This issue collects them.

### 🟡 Risk — recommend addressing before heavy production traffic

**1. Transient ORM instances may break lazy-relationship callers**
`_team_from_dict()` reconstructs detached `EmailTeam` objects. Any downstream caller that accesses a lazy relationship on a cache-hit object (e.g. `team.members`, `team.roles`) will get a `DetachedInstanceError` or silently-empty result — with no exception to surface the problem. Audit all consumers of `get_user_teams()` return values and confirm none traverse ORM relationships post-call.

**2. ID-cache and object-cache drift**
`teams_list` (ID cache) and `team_objs` (object cache) coexist in `AuthCache`. Any caller that still reads the ID-cache bucket while writes go to the object-cache bucket will see stale membership. Verify all active callers have been migrated and no dual-read paths remain.

**3. Invalidation completeness**
`invalidate_user_teams()` is called on the paths visible in the PR diff, but bulk membership operations, role changes, and soft-deletes may not trigger it. An audit of all team-mutation paths (service layer + routers) is needed to confirm every write that changes team membership or team state also invalidates the cache.

**4. Partial field serialization**
`team_to_dict()` cherry-picks a fixed set of scalar fields. If any current or future consumer accesses a field absent from the dict (e.g. a new `owner` column), the cache-hit path returns a wrong/default value while the DB-hit path returns the correct one — a silent divergence. Consider either serialising the full model or adding a schema-version key so stale entries are rejected on deserialization.

### 🔵 Nit — low priority, track for future cleanup

**5. Datetime naive/aware consistency**
`team_to_dict()` uses `.isoformat()` and `_team_from_dict()` uses `datetime.fromisoformat()`. Round-trip is correct in Python 3.11+, but naive vs. aware consistency should be verified — particularly if the ORM model ever mixes timezone-naive and timezone-aware fields.

**6. Nullable field access in `team_to_dict()`**
Direct attribute access on `slug`, `created_by`, `visibility`, `is_active` is safe today but brittle if any of these become nullable in a future schema change. Add explicit `getattr(team, "field", None)` or rely on Pydantic coercion.

**7. Old ID-cache methods are dead code**
The previous ID-only cache methods remain in `AuthCache`. If no callers use them, remove them to reduce maintenance surface.

**8. No end-to-end performance data**
The PR's primary motivation is latency/DB load reduction. Unit tests confirm correctness but not the actual p95 improvement. A benchmark or trace comparison (before/after with `LOG_LEVEL=DEBUG`) would validate the optimization actually lands in production workloads.

## References

- Introduced by: #4550
- Original perf issue: #3005

msureshkumar88:
## Problem

PR #4550 introduced a full-object team cache (`_team_objects_cache`) alongside the existing ID-only cache (`_teams_list_cache`). This left four concrete defects:

1. **Cross-worker eviction gap** — Redis pub/sub handler evicted `_teams_list_cache` on `teams:` messages but never touched `_team_objects_cache`. Remote workers served stale full-object data indefinitely.

2. **`update_team` invalidation gap** — Committing name/description/visibility/max_members changes triggered no cache invalidation. Every team member saw stale team objects until TTL expiry. `delete_team` already had the correct pattern; `update_team` did not.

3. **Transient ORM crash risk** — `_team_from_dict` reconstructs `EmailTeam` instances from cached dicts without attaching them to a SQLAlchemy session. `get_member_count()` and `is_member()` both tried to access lazy-loaded relationships (`self.members`) on these detached instances, which raises `AttributeError: 'NoneType' object has no attribute 'query'` at runtime.

4. **Brittle field serialization** — `team_to_dict` used direct attribute access (`team.slug`, `team.created_by`) and `_team_from_dict` used `d["created_by"]` / `d["slug"]`. Safe today since those columns are `NOT NULL`, but will silently crash if any column ever becomes nullable or if a cache entry written by an older app version is deserialized by a newer one.

## Approach

All four fixes follow the existing patterns in the codebase:

**Fix 1 — `mcpgateway/cache/registry_cache.py`**  
Added a second `_evict_keys` call for `_team_objects_cache` inside the existing `_lock` block in `CacheInvalidationSubscriber._process_invalidation()`. Both buckets now evict atomically under one lock acquisition.

**Fix 2 — `mcpgateway/services/team_management_service.py`**  
After `self.db.commit()` in `update_team`, added a best-effort try/except block that queries active `EmailTeamMember` rows and calls `self._fire_and_forget(auth_cache.invalidate_user_teams(...))` per member plus `admin_stats_cache.invalidate_teams()`. Mirrors the `delete_team` pattern already in the same file.

**Fix 3 — `mcpgateway/db.py` + `mcpgateway/services/team_management_service.py`**  
`get_member_count()` and `is_member()` now check `object_session(self) is None` and return `0` / `False` immediately for transient instances. Added docstring warnings to `_team_from_dict` and `get_user_teams` documenting that cache-hit objects are transient and lazy relationships must not be accessed on them.

**Fix 4 — `mcpgateway/cache/auth_cache.py` + `mcpgateway/services/team_management_service.py`**  
`team_to_dict` switched to `getattr(team, field, default)` for all optional fields. `_team_from_dict` switched to `d.get(key, default)` for all non-required fields (`id` and `name` remain required direct access). Defaults match the column schema defaults (`visibility="public"`, `is_active=True`, etc.).

## Testing

Each fix was developed TDD — failing test written first, then the fix applied:

- `TestCacheInvalidationSubscriber::test_process_auth_teams_invalidation_clears_both_caches` — verifies both buckets evicted for target user, other user untouched
- `TestTeamManagementService::test_update_team_invalidates_member_team_object_caches` — verifies `invalidate_user_teams` fired for all active members after `update_team`
- `TestTransientTeamFromDict::test_get_member_count_on_transient_instance_returns_zero` — verifies no crash on detached instance
- `TestTransientTeamFromDict::test_is_member_on_transient_instance_returns_false` — verifies no crash on detached instance  
- `TestTransientTeamFromDict::test_team_from_dict_scalar_fields_round_trip` — verifies all scalar fields survive `team_to_dict` → `_team_from_dict` round-trip
- `TestTeamToDictDefensiveAccess::test_team_to_dict_handles_missing_optional_attributes` — verifies `AttributeError` not raised on missing optional attrs
- `TestTeamToDictDefensiveAccess::test_team_from_dict_handles_missing_optional_keys` — verifies `KeyError` not raised on minimal dict

714 related unit tests pass. No regressions.

Closes #4884

### DIFF
--- a/mcpgateway/services/team_management_service.py
+++ b/mcpgateway/services/team_management_service.py
@@ -63,13 +63,13 @@ def _team_from_dict(d: Dict[str, Any]) -> EmailTeam:
     return EmailTeam(
         id=d["id"],
         name=d["name"],
-        slug=d["slug"],
+        slug=d.get("slug", ""),
         description=d.get("description"),
-        created_by=d["created_by"],
-        is_personal=d["is_personal"],
-        visibility=d["visibility"],
+        created_by=d.get("created_by", ""),
+        is_personal=d.get("is_personal", False),
+        visibility=d.get("visibility", "public"),
         max_members=d.get("max_members"),
-        is_active=d["is_active"],
+        is_active=d.get("is_active", True),
         created_at=datetime.fromisoformat(d["created_at"]) if d.get("created_at") else None,
         updated_at=datetime.fromisoformat(d["updated_at"]) if d.get("updated_at") else None,
     )
--- a/mcpgateway/services/team_management_service.py
+++ b/mcpgateway/services/team_management_service.py
@@ -696,6 +696,19 @@ class TeamManagementService:
             team.updated_at = utc_now()
             self.db.commit()
 
+            # Invalidate cached team objects for all active members so scalar field
+            # changes (name, visibility, description, max_members) propagate immediately.
+            try:
+                memberships = self.db.query(EmailTeamMember).filter(
+                    EmailTeamMember.team_id == team_id,
+                    EmailTeamMember.is_active.is_(True),
+                ).all()
+                for membership in memberships:
+                    self._fire_and_forget(auth_cache.invalidate_user_teams(membership.user_email))
+                self._fire_and_forget(admin_stats_cache.invalidate_teams())
+            except Exception as cache_error:
+                logger.debug(f"Failed to invalidate caches after team update for {SecurityValidator.sanitize_log_message(team_id)}: {cache_error}")
+
             logger.info(f"Updated team {SecurityValidator.sanitize_log_message(team_id)} by {updated_by}")
             return True
 
--- a/mcpgateway/services/team_management_service.py
+++ b/mcpgateway/services/team_management_service.py
@@ -714,7 +714,7 @@ class TeamManagementService:
                     self._fire_and_forget(auth_cache.invalidate_user_teams(membership.user_email))
                 self._fire_and_forget(admin_stats_cache.invalidate_teams())
             except Exception as cache_error:
-                logger.debug(f"Failed to invalidate caches after team update for {SecurityValidator.sanitize_log_message(team_id)}: {cache_error}")
+                logger.warning(f"Failed to invalidate caches after team update for {SecurityValidator.sanitize_log_message(team_id)}: {cache_error}")
 
             logger.info(f"Updated team {SecurityValidator.sanitize_log_message(team_id)} by {updated_by}")
             return True
--- a/mcpgateway/services/team_management_service.py
+++ b/mcpgateway/services/team_management_service.py
@@ -410,6 +410,25 @@ class TeamManagementService:
         except Exception as cache_error:
             logger.debug(f"Failed to invalidate membership caches for {SecurityValidator.sanitize_log_message(user_email)}: {cache_error}")
 
+    def _invalidate_team_member_caches(self, team_id: str) -> None:
+        """Invalidate cached team objects for all active members after a team update.
+
+        Errors are logged at warning level but do not propagate.
+
+        Args:
+            team_id: ID of the team whose member caches should be invalidated.
+        """
+        try:
+            memberships = self.db.query(EmailTeamMember).filter(
+                EmailTeamMember.team_id == team_id,
+                EmailTeamMember.is_active.is_(True),
+            ).all()
+            for membership in memberships:
+                self._fire_and_forget(auth_cache.invalidate_user_teams(membership.user_email))
+            self._fire_and_forget(admin_stats_cache.invalidate_teams())
+        except Exception as cache_error:
+            logger.warning(f"Failed to invalidate caches after team update for {SecurityValidator.sanitize_log_message(team_id)}: {cache_error}")
+
     def _check_user_team_limit(self, user_email: str) -> None:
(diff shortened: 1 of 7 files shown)


mcpgateway/services/team_management_service.py
```diff
@@ -696,6 +703,19 @@ async def update_team(
...
+            try:
+                memberships = self.db.query(EmailTeamMember).filter(
+                    EmailTeamMember.team_id == team_id,
+                    EmailTeamMember.is_active.is_(True),
+                ).all()
+                for membership in memberships:
+                    self._fire_and_forget(auth_cache.invalidate_user_teams(membership.user_email))
+                self._fire_and_forget(admin_stats_cache.invalidate_teams())
+            except Exception as cache_error:
+                logger.debug(f"Failed to invalidate caches after team update for {SecurityValidator.sanitize_log_message(team_id)}: {cache_error}")
```
prakhar-singh1928: Consider elevating the exception handler from logger.debug to logger.warning. Cache invalidation failures can cause stale data to be served, and DEBUG-level logs will not be visible in production monitoring.

prakhar-singh1928: Also Extract cache invalidation logic (lines 706-717) - The cache invalidation block adds 12 lines to an already substantial method. Worth considering extraction to a helper like `_invalidate_team_member_caches(team_id)` for better separation of concerns and independent testability.

msureshkumar88: Elevated to `logger.warning`. Cache invalidation failures causing stale data should be visible in production monitoring.

msureshkumar88: Extracted into `_invalidate_team_member_caches(team_id)` helper, matching the existing `_invalidate_membership_caches` pattern in the same class. All existing tests pass.

prakhar-singh1928 (approved):
✅ Issue 1: Cache Invalidation Extracted to Helper Method
Lines 413-430 - The cache invalidation logic has been properly extracted to [_invalidate_team_member_caches()](vscode-webview://1avjiu21ptfnb8hunju7u62b3dl2msb88otflsvbqv3o7khk226i/mcpgateway/services/team_management_service.py:413), which:

Improves separation of concerns
Makes the logic independently testable
Reduces complexity in the main update_team() method
Includes clear docstring explaining the error handling behavior
✅ Issue 2: Logging Level Elevated to WARNING
Line 430 - Cache invalidation failures now log at logger.warning() level instead of logger.debug(), ensuring:

Production visibility for cache consistency issues
Alignment with the severity of potential stale data problems
Better operational monitoring and alerting capabilities

Approved

ja8zyjits (approved):
LGTM

PR #5047, opened 2026-06-04
fix(ui-rewrite): update .env.example for csrf

marekdano:
Update `.env.example`: align `CSRF_COOKIE_NAME` with the `mcpgateway_csrf_token` value used in code; add missing CSRF-exempt paths (`/app/auth/login`, `/admin/login`, `/admin/forgot-password`, `/admin/reset-password`)

Related #4898


gcgoncalves (approved):
Great patch, will save us some time on env resets. 🙇

PR #5059, opened 2026-06-04
perf: speed up rust fast-test server

### LINKED ISSUE #4497: Fast time server go to rust

lucarlig:
## Summary

This draft updates the Rust `fast-test-server` to reduce MCP benchmark latency:

- replace per-call async mutex stats with relaxed atomic counters
- route `/mcp` through a lean direct Streamable HTTP JSON-RPC handler for this fixed tool surface
- preserve schema fixture tool metadata, including `outputSchema`
- preserve MCP `echo` delay and jitter behavior on the direct handler
- validate required direct-handler tool arguments and return JSON-RPC invalid params for malformed calls
- validate direct-handler MCP sessions, including rejecting missing sessions with HTTP 400, unknown/deleted sessions with HTTP 404, and removing sessions on `DELETE /mcp`
- advertise the latest RMCP protocol version (`2025-11-25`) while accepting older initialize payloads such as `2024-11-05`
- align the Go fast-time `/version` endpoint with `mcp-go`'s latest protocol version and pin that in its unit test
- switch Rust fast-test timezone handling to `chrono-tz` so IANA zones, DST, half-hour offsets, and UTC `Z` formatting match Go fast-time behavior
- add Rust handler-level tests for version reporting, backward-compatible initialize, Streamable HTTP session lifecycle behavior, and Go-compatible timezone conversion
- add a live ContextForge E2E parity test that discovers the federated Go/Rust tools, compares deterministic `convert_time` outputs, and checks Rust latency stays in the same CF-routed band as Go
- update fast-test-server Rust dependencies to the latest crates.io versions not newer than 7 days as of 2026-06-04
- remove unused RMCP streamable HTTP features and `tokio-util` from this server

Benchmark with `sdk-benchmark` from `sco3/mcp-benchmark`, single user, `-i 100 -r 10000`, `get_system_time` with `{"timezone":"UTC"}`:

```text
Rust after changes:
Init:      1377.54 req/s, 0.73ms
Tool/list: 7930.87 req/s, 0.12ms
Tool call: 11050.63 req/s, 0.09ms
```

For context, the local Go fast-time run on the same benchmark shape measured:

```text
Init:      1019.79 req/s, 0.98ms
Tool/list: 5209.63 req/s, 0.18ms
Tool call: 7403.40 req/s, 0.13ms
```

## Verification

- `go test ./...` in `mcp-servers/go/fast-time-server`
- `cargo fmt --check` in `mcp-servers/rust/fast-test-server`
- `cargo test` in `mcp-servers/rust/fast-test-server`
- `cargo build --release` in `mcp-servers/rust/fast-test-server`
- `JWT_SECRET_KEY=my-test-key-but-now-longer-than-32-bytes MCP_CLI_BASE_URL=http://127.0.0.1:8080 .venv/bin/python -m pytest tests/live_gateway/mcp/test_fast_server_go_rust_parity.py -q -s`
  - local ContextForge gateway with locally registered Go fast-time and Rust fast-test upstreams
  - deterministic `convert_time` parity passed for `America/New_York` DST and `Asia/Kolkata` half-hour offset cases
  - CF-routed latency parity passed with Go avg `0.025996s`, Rust avg `0.026905s`, `FAST_SERVER_PARITY_CALLS=5`
- Live MCP smoke test against `127.0.0.1:19080`:
  - initialized with `2024-11-05` request payload and received `2025-11-25`
  - valid `tools/list` with issued session returned HTTP 200
  - missing session returned HTTP 400
  - fabricated session returned HTTP 404
  - deleted session returned HTTP 404

Note: `make testing-up` could not be used in this local environment because required GHCR images returned 403, so the live parity verification above used local CF and local upstream processes registered through the same `/gateways` API path.

Closes #4497


msureshkumar88 (changes requested):
Good performance direction overall. A few changes needed before this can merge.

**Blocking:**
1. E2E parity test stale tool name hints — the new test will unconditionally skip in CI after the rename (see inline comment)
2. Predictable session IDs — sequential integers are trivially enumerable; replace with a random token
3. Unbounded session growth — no cap means the session store is an OOM vector
4. `mcp_error_response_with_status` interpolates `message` directly into JSON without escaping — inconsistent with `mcp_text_result_response` which uses `serde_json::to_string`
5. `echo` delay has no upper bound — a client can stall a handler with an arbitrarily large value

**Non-blocking suggestions:**
- `FastTestServer` rmcp layer is never mounted — either remove the ~150-line dead path or document that it's an unreachable reference implementation
- `tools/list` hardcodes a raw JSON string that duplicates the `#[tool]` declarations — two sources of truth that can drift
- `mcp_base_headers()` allocates a new `HeaderMap` on every response — a `LazyLock<HeaderMap>` or axum tuple response would eliminate this allocation from the hot path
- `FastTestServer::request_count` is never incremented in production (only in the dead rmcp path)
- PR says "Refs #4497" — confirm whether this closes the issue or partially addresses it

tests/live_gateway/mcp/test_fast_server_go_rust_parity.py
```diff
@@ -0,0 +1,149 @@
...
+    result = await client.call_tool_mcp(name=tool_name, arguments=arguments)
+    return _text_result(result, tool_name)
+
+
+@pytest.mark.flaky(reruns=1, reruns_delay=2)
+async def test_go_and_rust_fast_servers_match_time_conversion_through_contextforge(client: Client) -> None:
+    """The Rust fast server must preserve Go fast-time behavior once federated."""
+    tools = await client.list_tools()
+    go_convert = _find_tool(tools, "fast-time-convert-time", ("fast-time", "fast-time-server"), "convert-time")
+    rust_convert = _find_tool(tools, "fast-test-convert-time", ("fast-test", "fast-test-server", "rust"), "convert-time")
```
msureshkumar88: **Stale hints cause this test to silently skip in CI.**

`_find_tool` looks for tool names containing `"fast-test"` or `"fast-test-server"` as a hint. After the rename to `fast-time-server`, ContextForge will register tools as `fast-time-convert-time`. None of the supplied hints match that prefix — `_find_tool` falls through to `pytest.skip`, making the new E2E test a permanent no-op.

Same issue at line 128 for `get-system-time`.

```python
# current — won't match fast-time-convert-time
rust_convert = _find_tool(tools, "fast-test-convert-time", ("fast-test", "fast-test-server", "rust"), "convert-time")

# fix — align hints with the renamed server
rust_convert = _find_tool(tools, "fast-time-convert-time", ("fast-time", "fast-time-server"), "convert-time")
```

mcp-servers/rust/fast-time-server/src/main.rs
```diff
@@ -0,0 +1,1257 @@
...
+
+fn mcp_id_json(id: Option<&serde_json::Value>) -> String {
+    id.and_then(|value| serde_json::to_string(value).ok())
+        .unwrap_or_else(|| "null".to_string())
+}
+
+fn mcp_initialize_response(id: Option<&serde_json::Value>) -> Response {
+    let mut headers = mcp_base_headers();
+    let session_id = SESSION_COUNTER.fetch_add(1, Ordering::Relaxed);
+    let session_id = format!("fast-time-{session_id}");
```
msureshkumar88: **Predictable session IDs allow session enumeration.**

`fast-time-1`, `fast-time-2`, … are trivially guessable. Any client that knows its own session ID can enumerate all live sessions and submit tool calls with a fabricated `mcp-session-id` header. Even for a benchmark server that runs behind ContextForge, this is worth hardening — the session store is the only access control boundary.

```rust
// current — sequential, guessable
let session_id = SESSION_COUNTER.fetch_add(1, Ordering::Relaxed);
let session_id = format!("fast-time-{session_id}");

// fix — cryptographically random, un-guessable
use uuid::Uuid;
let session_id = Uuid::new_v4().to_string();
```

Add `uuid = { version = "1", features = ["v4"] }` to `Cargo.toml`.

mcp-servers/rust/fast-time-server/src/main.rs
```diff
@@ -0,0 +1,1257 @@
...
+        ),
+    )
+}
+
+fn mcp_session_id(headers: &HeaderMap) -> Option<&str> {
+    headers.get(SESSION_HEADER)?.to_str().ok()
+}
+
+fn remember_session(session_id: String) {
+    if let Ok(mut sessions) = ACTIVE_SESSIONS.write() {
```
msureshkumar88: **Unbounded session store — no cap on concurrent sessions.**

Every `POST /mcp` with `method: "initialize"` inserts into `ACTIVE_SESSIONS` permanently (no TTL, no max). A misbehaving benchmark client or any caller that repeatedly initializes without sending `DELETE /mcp` grows this indefinitely.

```rust
fn remember_session(session_id: String) {
    if let Ok(mut sessions) = ACTIVE_SESSIONS.write() {
        // consider: reject if sessions.len() >= MAX_SESSIONS
        sessions.insert(session_id);
    }
}
```

A reasonable cap (e.g., `MAX_SESSIONS = 10_000`) with a `503 Service Unavailable` response when exceeded keeps memory bounded. Alternatively, add a session TTL using a timestamp map.

mcp-servers/rust/fast-time-server/src/main.rs
```diff
@@ -0,0 +1,1257 @@
...
+    message: &str,
+    data: Option<serde_json::Value>,
+) -> Response {
+    let data = data
+        .map(|value| format!(r#","data":{}"#, value))
+        .unwrap_or_default();
+    let mut response = mcp_json_response(
+        mcp_base_headers(),
+        format!(
+            r#"{{"jsonrpc":"2.0","id":{},"error":{{"code":{},"message":"{}"{}}}}}"#,
```
msureshkumar88: **`message` is interpolated into JSON without escaping — inconsistent with the rest of the codebase.**

`mcp_text_result_response` (line 824) correctly serializes the text field through `serde_json::to_string(text)`. This function does not — `message` lands raw inside `"message":"{}"`. A value containing `"` or `}` produces malformed JSON or allows field injection.

All current call sites happen to use static string literals, which is why this hasn't surfaced yet. But the function's signature accepts any `&str`, and the next caller may not know the constraint.

```rust
// current — unsafe
format!(
    r#"{{"jsonrpc":"2.0","id":{},"error":{{"code":{},"message":"{}"{}}}}}"#,
    mcp_id_json(id), code, message, data
)

// fix — escape consistently
let escaped_message = serde_json::to_string(message).unwrap_or_else(|_| "\"\"".to_string());
format!(
    r#"{{"jsonrpc":"2.0","id":{},"error":{{"code":{},"message":{}{}}}}}"#,
    mcp_id_json(id), code, escaped_message, data
)
```

mcp-servers/rust/fast-time-server/src/main.rs
```diff
@@ -0,0 +1,1257 @@
...
+
+    match name {
+        "echo" => {
+            let Some(arguments) = mcp_arguments_object(id, arguments) else {
+                return mcp_invalid_params_response(id, "arguments must be an object");
+            };
+            let Some(message) = mcp_required_string(id, arguments, "message") else {
+                return mcp_invalid_params_response(id, "message must be a string");
+            };
+            let Some(delay) = mcp_optional_u64(id, arguments, "delay") else {
```
msureshkumar88, line 420: **`delay` has no upper bound — a caller can stall a handler thread for an arbitrarily long time.**

`delay: Option<u64>` accepts any value up to `u64::MAX`. Sending `{"delay": 3600000}` ties up a tokio task for an hour. Under concurrent load this drains the thread pool and starves other requests.

```rust
// fix — clamp before sleeping
const MAX_DELAY_MS: u64 = 60_000;
if let Some(ms) = delay {
    if ms > 0 {
        let ms = ms.min(MAX_DELAY_MS);
        let actual_ms = compute_delay(ms, delay_stddev);
        tokio::time::sleep(std::time::Duration::from_millis(actual_ms)).await;
    }
}
```

Or return `mcp_invalid_params_response` for values above the cap so callers know the limit.

mcp-servers/rust/fast-time-server/src/main.rs
```diff
@@ -0,0 +1,1257 @@
...
+}
+
+impl Default for FastTestServer {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+#[tool_router]
+impl FastTestServer {
```
msureshkumar88: **`FastTestServer` rmcp layer is dead code — `main()` never mounts it.**

`main()` builds the axum router from standalone functions only (`mcp_handler`, `health_handler`, etc.). `FastTestServer::new()` is never called in the live path. All `#[tool]` implementations (`echo`, `get_system_time`, `convert_time`, `schema_error`, `schema_success`, `get_stats`) are compiled and linked but unreachable at runtime.

The `#[allow(dead_code)]` on the `tool_router` field (line 97) acknowledges part of this — the compiler warning is suppressed but the root cause isn't addressed.

A couple of approaches worth considering:

1. **Remove the rmcp layer entirely** — the direct handler already implements all six tools. The rmcp impls add ~150 lines with no live effect.
2. **Document explicitly** that the `FastTestServer` struct is a reference implementation kept for unit-testing `get_info()` / `initialize()` semantics but is not served.

Either way, the `#[allow(dead_code)]` should not be the only signal that something is off here.

mcp-servers/rust/fast-time-server/src/main.rs
```diff
@@ -0,0 +1,1257 @@
...
...urrent system time in the specified IANA timezone.","inputSchema":{{"type":"object","properties":{{"timezone":{{"type":"string"}}}}}}}},{{"name":"convert_time","description":"Convert a time value from a source IANA timezone to a target IANA timezone.","inputSchema":{{"type":"object","properties":{{"time":{{"type":"string"}},"source_timezone":{{"type":"string"}},"target_timezone":{{"type":"string"}}}},"required":["time","source_timezone","target_timezone"]}}}},{{"name":"schema_error","description":"Always returns isError=true.","inputSchema":{{"type":"object","properties":{{}}}},"outputSchema":{{"type":"object","properties":{{"recognitionId":{{"type":"string"}},"message":{{"type":"string"}}}},"required":["recognitionId"]}}}},{{"name":"schema_success","description":"Returns a JSON payload that conforms to the declared outputSchema.","inputSchema":{{"type":"object","properties":{{}}}},"outputSchema":{{"type":"object","properties":{{"recognitionId":{{"type":"string"}},"message":{{"type":"string"}}}},"required":["recognitionId"]}}}},{{"name":"get_stats","description":"Get server statistics including request count and uptime.","inputSchema":{{"type":"object","properties":{{}}}}}}]}}}}"#,
```
msureshkumar88: **`tools/list` and `#[tool]` declarations are two sources of truth that can drift.**

This ~1 KB hardcoded format string duplicates the tool names, descriptions, `inputSchema`, and `outputSchema` already declared on `FastTestServer` via `#[tool]` macros. If someone updates a tool description in the rmcp layer, the direct handler's `tools/list` silently remains stale — clients see different metadata depending on which path they hit.

Worth considering: generate this static string once via `FastTestServer::tool_router()` at startup and store it as a `LazyLock<String>`, eliminating the duplication. If the direct handler is the permanent future, at minimum add a comment cross-referencing the rmcp declarations so the link is explicit.

mcp-servers/rust/fast-time-server/src/main.rs
```diff
@@ -0,0 +1,1257 @@
...
+        return StatusCode::ACCEPTED.into_response();
+    }
+
+    match method {
+        "initialize" => mcp_initialize_response(id),
+        _ => mcp_error_response(id, -32601, "Method not found", None),
+    }
+}
+
+fn mcp_base_headers() -> HeaderMap {
```
msureshkumar88: **`mcp_base_headers()` allocates a new `HeaderMap` on every response — 11 K alloc/s on the hot path.**

Every `mcp_text_result_response` → `mcp_json_response` → `mcp_base_headers()` round-trip creates and drops a `HeaderMap`. At the advertised 11 050 req/s tool-call throughput this adds up.

Two cheaper alternatives:

```rust
// Option A: axum tuple response — zero HeaderMap allocation
(
    [(header::CONTENT_TYPE, "application/json")],
    body,
)
.into_response()

// Option B: LazyLock static — allocated once, shared forever
static BASE_HEADERS: LazyLock<HeaderMap> = LazyLock::new(|| {
    let mut h = HeaderMap::new();
    h.insert(header::CONTENT_TYPE, HeaderValue::from_static("application/json"));
    h
});
```

Option A is cleanest since axum's `IntoResponse` for tuples is zero-copy.

mcp-servers/rust/fast-time-server/src/main.rs
```diff
@@ -0,0 +1,1257 @@
...
+
+// ============================================================================
+// FastTestServer Implementation
+// ============================================================================
+
+#[derive(Clone)]
+pub struct FastTestServer {
+    #[allow(dead_code)]
+    tool_router: ToolRouter<FastTestServer>,
+    request_count: Arc<AtomicU64>,
```
msureshkumar88: **`request_count` on `FastTestServer` is never incremented in production.**

The direct handler path increments `DIRECT_REQUEST_COUNT` (the global). `self.request_count` is only incremented inside the rmcp `#[tool]` methods — which are never called in the live path (see comment at line 109). So `get_stats` via the direct handler correctly reads `DIRECT_REQUEST_COUNT`, but `FastTestServer::request_count` will always read `0`.

This is a consequence of the dead rmcp layer. Removing that layer removes the confusion here too.

tests/live_gateway/mcp/test_fast_server_go_rust_parity.py
lucarlig: Good catch. Fixed both Rust tool lookups to use the renamed `fast-time-*` names and `fast-time` hints, so the parity test no longer falls through to the skip path.

mcp-servers/rust/fast-time-server/src/main.rs
lucarlig: Agreed. Replaced the sequential session counter with `Uuid::new_v4()` session IDs and added a unit assertion that initialize returns a UUID rather than the old `fast-time-*` sequence.

mcp-servers/rust/fast-time-server/src/main.rs
lucarlig: Added `MAX_ACTIVE_SESSIONS = 10_000`; initialize now returns a 503 JSON-RPC error when the cap is reached. Added a unit test around the cap logic as well.

mcp-servers/rust/fast-time-server/src/main.rs
lucarlig: Fixed by serializing the error message with `serde_json::to_string` before formatting the response body. Added a regression test with quotes/braces to prove it cannot inject fields or break JSON.

mcp-servers/rust/fast-time-server/src/main.rs
lucarlig: Added `MAX_DELAY_MS = 60_000` and now reject larger MCP echo delays with invalid params so callers see the limit. The REST echo path uses the same validation, and jittered computed delays are clamped to the same maximum.

mcp-servers/rust/fast-time-server/src/main.rs
lucarlig: Removed the unused rmcp `FastTestServer` layer entirely, along with the stale tests and now-unused `rmcp`/`schemars` dependencies.

mcp-servers/rust/fast-time-server/src/main.rs
lucarlig: Resolved by removing the rmcp tool declarations, leaving the direct handler as the single served source for tool metadata.

mcp-servers/rust/fast-time-server/src/main.rs
lucarlig: Changed normal JSON responses to use axum tuple headers through `mcp_json_response`; initialize is the only path that mutates headers because it has to attach the dynamic session ID.

mcp-servers/rust/fast-time-server/src/main.rs
lucarlig: Removed this confusion with the dead rmcp layer. The only remaining stats path uses `DIRECT_REQUEST_COUNT`, which is the counter incremented by the served direct handler.

msureshkumar88 (approved):
All requested changes addressed and verified against the source:

- Session IDs now use `Uuid::new_v4()` — no longer enumerable
- Session cap enforced at 10 000; initialize returns 503 when full, with a unit test covering the boundary
- `mcp_error_response_with_status` uses `serde_json::to_string` for message escaping — JSON injection path closed
- Echo delay capped at 60 s via `validate_delay()` in both the MCP and REST paths
- Parity test hints updated to `fast-time` prefix — no longer silently skips in CI
- rmcp layer removed entirely (dead code + stale deps gone)
- `mcp_json_response` returns an axum tuple — HeaderMap alloc eliminated from the hot path

Good work turning these around cleanly. Ready to merge.

ja8zyjits:
### ⚠️ **Potential Considerations**

1. **Timezone Output Changes**:
   - Clients relying on old incorrect offsets (e.g., `Asia/Kolkata` +5:00) will see corrected values (+5:30)
   - **Impact**: Low - this is a bug fix, not a breaking change
   - **Mitigation**: Documented in PR, E2E tests verify parity with Go

2. **Protocol Version Advertisement**:
   - Now advertises `2025-11-25` instead of `2024-11-05`
   - **Impact**: None - clients should handle version negotiation per MCP spec
   - **Mitigation**: Still accepts older client versions

3. **Session Limit**:
   - New hard cap of 10,000 active sessions
   - **Impact**: Low - previous RMCP implementation had no explicit limit but would hit memory constraints
   - **Mitigation**: Returns HTTP 503 with clear error message when limit reached
   
### ⚠️ **Minor Observations**
1. **Session storage**: In-memory `HashSet` - sessions lost on restart (acceptable for benchmark server)
2. **No graceful session cleanup**: On shutdown, active sessions not explicitly cleared (minor - they'll timeout)
3. **Hardcoded constants**: `MAX_ACTIVE_SESSIONS`, `MAX_DELAY_MS` not configurable via env vars (acceptable for benchmark tool)

@lucarlig please consider them in your future PR of removing the GO completely.
Thanks.

ja8zyjits (approved):
LGTM

PR #5061, opened 2026-06-04
docs: fix section 14 manual testing commands and section 12.4 missed secret fix

### LINKED ISSUE #4896: [DOCS]: Section 14 manual testing — three undocumented behaviours

msureshkumar88:
## Summary

- Applies all fixes from PR #4897 (documenting undocumented/incorrect behaviours found during v1.0.2 manual testing)
- Adds a fix that PR #4897 missed: section 12.4 admin token generation still used `${JWT_SECRET_KEY:-my-test-key-but-now-longer-than-32-bytes}` hardcoded fallback

### Changes

**Section 12.4 (missed by #4897):**
- Fix `--secret "${JWT_SECRET_KEY:-hardcoded-fallback}"` → `--secret "$JWT_SECRET_KEY"` — if `JWT_SECRET_KEY` not exported, fallback silently produced tokens signed with wrong key

**Section 14 preamble:**
- Add `make init-secrets` prerequisite block so testers know to initialise secrets before running the suite

**Section 14.1:**
- Fix token generation to extract `JWT_SECRET_KEY` from `.env` via `grep/cut` instead of the hardcoded test value

**Section 14.2 / 14.3:**
- Add Expected notes explaining `gateway_name` is absent from `/tools` response by design

**Section 14.4:**
- Fix export command to pass `MCPGATEWAY_BEARER_TOKEN` to the CLI subprocess
- Document ephemeral MCP tool exclusion from export output

**Section 14.5:**
- Add Note explaining Streamable HTTP stateful `initialize` handshake requirement

Closes #4896

### DIFF
--- a/docs/docs/development/release-management.md
+++ b/docs/docs/development/release-management.md
@@ -1013,7 +1013,7 @@ export MCPGATEWAY_ADMIN_TOKEN=$(./.venv/bin/python -m mcpgateway.utils.create_jw
   --username admin@example.com \
   --admin \
   --exp 10080 \
-  --secret "${JWT_SECRET_KEY:-my-test-key-but-now-longer-than-32-bytes}" \
+  --secret "$JWT_SECRET_KEY" \
   --algo HS256 \
   2>/dev/null | tail -n 1)
 
@@ -1206,7 +1206,18 @@ The migration test suite follows an **n-2 support policy** and tests sequential 
 
 ## 14. Manual Testing
 
-These tests verify core user-facing workflows that automated tests do not fully cover. Build and start the compose stack first:
+These tests verify core user-facing workflows that automated tests do not fully cover.
+
+Before starting, ensure `.env` has a real JWT secret (the default `.env` ships with a placeholder that breaks token authentication):
+
+```bash
+# Generate secrets and merge into .env (skip if already done)
+make init-secrets
+# Confirm JWT_SECRET_KEY is no longer the placeholder
+grep JWT_SECRET_KEY .env   # must not contain __REPLACE_ME__
+```
+
+Build and start the compose stack:
 
 ```bash
 make docker-prod && make testing-up
@@ -1217,10 +1228,13 @@ make docker-prod && make testing-up
 Create a token for API and client access:
 
 ```bash
+# JWT_SECRET_KEY must match the value in .env (run make init-secrets if not set)
+export JWT_SECRET_KEY=$(grep '^JWT_SECRET_KEY=' .env | cut -d= -f2-)
+
 export MCPGATEWAY_BEARER_TOKEN=$(python -m mcpgateway.utils.create_jwt_token \
   --username admin@example.com \
   --exp 10080 \
-  --secret my-test-key-but-now-longer-than-32-bytes)
+  --secret "$JWT_SECRET_KEY")
 
 export BASE_URL="http://localhost:8080"
 ```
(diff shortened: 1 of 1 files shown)


msureshkumar88:
Superseded — applying fix directly to PR #4897's branch instead.

PR #5065, opened 2026-06-04
fix(admin): prevent false-positive pattern error on Add Gateway form

### LINKED ISSUE #5064: [BUG]: Add Gateway fails with pattern validation error for auth types other than "Query Parameter"
Image

Image

## Description

Submitting the \"Add New MCP Server or Gateway\" form with **any** authentication type (Bearer, Basic, OAuth 2.0, etc.) fails with:

> String should match pattern '^[a-zA-Z_][a-zA-Z0-9_\-]*$'

The error gives no field name, making it appear as though the gateway name or URL is invalid when it is not.

## Steps to reproduce

1. Open the Admin UI → Gateways tab
2. Click **Add Gateway**
3. Fill in MCP Server Name and MCP Server URL
4. Set Authentication Type to anything **other than** "Query Parameter (INSECURE)" — e.g. OAuth 2.0 or Bearer Token
5. Click **Add Gateway**
6. Observe the error banner above the form

## Root cause

**HTML layer:** The form contains a hidden `<div id="auth-query_param-fields-gw">` with an `<input name="auth_query_param_key">` inside it. The div is toggled visible/hidden via `style.display` by `handleAuthTypeChange()` in `auth.js`, but the input is never `disabled`. Browsers submit all enabled inputs regardless of visibility, so `auth_query_param_key=""` is always included in the POST body.

**Server layer:** `_parse_gateway_data_from_request()` in `admin.py` collects every non-`None` form field into the data dict:

```python
for key in form.keys():
    value = form.get(key)
    if value is not None:
        data[key] = value
```

An empty string is not `None`, so `auth_query_param_key: ""` reaches `GatewayCreate(**data)`. Pydantic then rejects `""` against the field's `pattern=r"^[a-zA-Z_][a-zA-Z0-9_\-]*$"` constraint and raises a `ValidationError`.

The gateway **update** path already handles this correctly (`str(form.get("auth_query_param_key", "")) or None`), but the create path goes through the shared parser which lacks the same normalization.

## Fix (two-layer)

1. **`admin.py` / `_parse_gateway_data_from_request`** — after form parsing, convert empty strings to `None` for `auth_query_param_key` and `auth_query_param_value` before the data reaches Pydantic. Mirrors what the update handler already does; provides server-side defense in depth regardless of frontend state.

2. **`admin_ui/auth.js` / `handleAuthTypeChange`** — disable inputs inside sections being hidden and re-enable them when shown. Stops the spurious empty-string fields from reaching the server in the first place.

nickmarden:
Fixes #5064.

## What and why

Clicking **Add Gateway** with any authentication type other than "Query Parameter (INSECURE)" produced:

> String should match pattern '^[a-zA-Z_][a-zA-Z0-9_\-]*$'

with no indication of which field was wrong, blocking all gateway creation that required authentication.

**Root cause:** The Add Gateway form contains a hidden `<div>` with `<input name="auth_query_param_key">`. Browsers submit every enabled input regardless of `display:none`, so `auth_query_param_key=""` was always included in the POST. An empty string fails the field's Pydantic pattern constraint, and the error surfaced as if the gateway name or URL was invalid.

## Changes

**`admin_ui/auth.js` — `handleAuthTypeChange`**

Introduce `hideSection` / `showSection` helpers that toggle both `style.display` and `disabled` on all child inputs when switching auth type. Disabled inputs are not submitted by browsers, eliminating the spurious empty-string field at the source.

**`mcpgateway/admin.py` — `_parse_gateway_data_from_request`**

After form parsing, convert `""` to `None` for `auth_query_param_key` and `auth_query_param_value` before the data reaches Pydantic. This mirrors what the gateway **update** handler already did manually and provides server-side defense in depth in case other form paths or API clients send empty strings.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-06-17T07:41:32Z",
+  "generated_at": "2026-06-17T15:31:43Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -5060,7 +5060,7 @@
         "hashed_secret": "995e87a207ec41ef95f938dcf53a4184312b0d93",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 11389,
+        "line_number": 11419,
         "type": "Hex High Entropy String",
         "verified_result": null
       },
@@ -5068,7 +5068,7 @@
         "hashed_secret": "a0281cd072cea8e80e7866b05dc124815760b6c9",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 17785,
+        "line_number": 17815,
         "type": "Secret Keyword",
         "verified_result": null
       }
--- a/mcpgateway/admin.py
+++ b/mcpgateway/admin.py
@@ -979,6 +979,16 @@ async def _parse_gateway_data_from_request(request: Request) -> dict[str, Any]:
         if oauth_config:
             data["oauth_config"] = oauth_config
 
+        # Hidden auth-section inputs are always submitted by browsers (even when the
+        # containing div is display:none), arriving as empty strings rather than None.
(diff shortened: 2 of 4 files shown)


cafalchio:
@nickmarden Hi, could you rebase and run the "make test" and make "pre-commit" before pushing again. 
Thank you.

PR #5076, opened 2026-06-05
fix(ui-rewrite): fetch tools and activate mcp server after OAuth, add…

### LINKED ISSUE #5026: [UI-REWRITE]: Tools not fetched and server not activated after successful OAuth MCP server creation in new UI
### 🐞 Summary

  Three related gaps in the OAuth MCP server creation flow in the new UI:

  1. After OAuth completes successfully, the `ExposeComponentsForm` renders with an empty tools/resources/prompts list — no fetch is triggered post-authorization.
  2. After OAuth completes successfully, the MCP server is not automatically set to `enabled: true` (active).
  3. The overflow menu (`ServerActionsMenu`) for MCP servers has no way to manually activate or deactivate a server — it is missing an "Activate" / "Deactivate" toggle item.

  ---

  ### 🔁 Steps to Reproduce

  **Issue 1 -Tools not fetched after OAuth:**
  1. Open "Connect MCP server" and configure a server with OAuth authentication.
  2. Submit the form and complete the OAuth authorization popup.
  3. Observe `ExposeComponentsForm` - tools/resources/prompts list is empty.

  **Issue 2 - Server not activated after OAuth:**
  1. Complete the same OAuth flow above.
  2. After the form closes, check the newly created server in the list - `enabled` is `false`.

  **Issue 3 - Missing Activate/Deactivate in overflow menu:**
  1. Open the three-dot overflow menu (`MoreVertical`) on any MCP server row.
  2. Observe only "Edit", "Test Connection", and "Delete" - no way to toggle the server's active state without going through the edit form.

  ---

  ### 🤔 Expected Behavior

  1. Once OAuth authorization succeeds, `ExposeComponentsForm` should refetch tools/resources/prompts so the user sees the server's capabilities immediately.
  2. A newly created OAuth MCP server should be automatically activated (`enabled: true`) once OAuth completes successfully, since the credentials are now in place.
  3. The overflow menu should show **"Activate"** when `server.enabled === false`, and **"Deactivate"** when `server.enabled === true`, allowing users to toggle the server's active state directly.

  ---

  ### 🧠 Root Cause (preliminary)

  **Issue 1:** In `useMCPServerForm.ts` (~line 688), `onSuccess` fires via `setTimeout` after OAuth
  resolves. `MCPServerForm.tsx` (lines 98–121) transitions to `ExposeComponentsForm`, which fires a
  `useQuery` for tools on mount. No refetch or retry is triggered after the OAuth handshake, so the
  query runs before the backend has registered the server's capabilities. The `oauthNotification`
  prop already flows into `ExposeComponentsForm` and could serve as the trigger signal for a refetch.

  **Issue 2:** The gateway is created but the `enabled` field is not flipped to `true` after OAuth
  succeeds. Activating the server should be part of the success path in `useMCPServerForm.ts` via
  `PATCH /gateways/{id}` with `{ enabled: true }`.

  **Issue 3:** `client/src/components/servers/ServerActionsMenu.tsx` does not receive or use
  `server.enabled` and has no `onActivate`/`onDeactivate` props. An activate/deactivate handler and
  the corresponding API call (`PATCH /gateways/{id}` with `{ enabled: true|false }`) need to be
  wired in. The `MCPServer` type already exposes `enabled: boolean` via `BaseServer` in
  `client/src/types/server.ts`.

  ---

  ### 🧩 Additional Context

  | File | Relevance |
  |------|-----------|
  | `client/src/hooks/useMCPServerForm.ts` | OAuth success path (~line 682–710) |
  | `client/src/components/mcp-servers/MCPServerForm.tsx` | `onSubmit` callback, transitions to expose form (~line 97–122) |
  | `client/src/components/gateways/ExposeComponentsForm.tsx` | `useQuery` for tools on mount (~line 151) |
  | `client/src/components/servers/ServerActionsMenu.tsx` | Overflow menu — missing activate/deactivate items |
  | `client/src/api/servers.ts` | API layer — no `activate`/`deactivate` helper yet |
  | `client/src/types/server.ts` | `MCPServer.enabled: boolean` already present via `BaseServer` |

marekdano:
## Summary

  Fixes three related gaps in the OAuth MCP server creation flow. 
  
  Closes #5026.

  - **Tools not fetched after OAuth**: `ExposeComponentsForm` now refetches
    tools/resources/prompts when `oauthNotification` transitions to `"success"`, using a
    `useEffect` that calls `refetchTools`, `refetchResources`, and `refetchPrompts`. The backend
    `/oauth/fetch-tools/{id}` endpoint is called immediately after OAuth completes and its
    result surfaced as a `fetchToolsNotification` banner.
  - **Server not activated after OAuth**: `useMCPServerForm` now calls `toggleEnabled(id, true)`
    immediately after a successful OAuth handshake, so the gateway is active before
    `ExposeComponentsForm` renders.
  - **Missing Activate / Deactivate in overflow menu**: `ServerActionsMenu` gains an optional
    `onToggleEnabled` prop. When provided it renders "Deactivate" for an enabled server and
    "Activate" for a disabled one. The handler lives in `Servers.tsx` and calls the new
    `serversApi.toggleEnabled` API helper, then refetches the server list.

  Additional changes made while addressing the above:

  - **`InlineNotification` component** (new): Extracts the repeated notification markup from
    `ExposeComponentsForm` and `MCPServerForm` into a single reusable component with correct
    ARIA roles (`role="alert"` / `role="status"`), decorative icons, and an accessible dismiss
    button. Accepts an optional `dismissLabel` prop so stacked notifications have distinct
    accessible names for screen readers.
  - **`parseApiError` utility**: Extended to handle `{"detail":{"message":"...","success":false}}`
    in addition to plain-string detail and validation-error arrays. All inline error extraction
    in `useMCPServerForm` is replaced with calls to `parseApiError`.
  - **`useQuery.refetch` stability**: `refetch` is now wrapped in `useCallback` so its reference
    is stable across re-renders, preventing unnecessary `useEffect` re-runs in consumers that
    list `refetch` as a dependency.
  - **Security**: `sanitizeError` is applied to the activation-failure log in `useMCPServerForm`,
    consistent with the rest of the codebase.

  ## Changed files

  | File | Change |
  |---|---|
  | `client/src/api/servers.ts` | Add `toggleEnabled` and `fetchToolsAfterOAuth` API methods |
  | `client/src/components/ui/inline-notification.tsx` | New reusable notification component |
  | `client/src/components/gateways/ExposeComponentsForm.tsx` | Wire `fetchToolsNotification`, add refetch `useEffect`, use `InlineNotification` |
  | `client/src/components/mcp-servers/MCPServerForm.tsx` | Surface `fetchToolsNotification`, use `InlineNotification` |
  | `client/src/components/servers/ServerActionsMenu.tsx` | Add `onToggleEnabled` prop and menu item |
  | `client/src/components/servers/ServersTable.tsx` | Pass `onToggleEnabled` through to `ServerActionsMenu` |
  | `client/src/pages/Servers.tsx` | Add `handleToggleEnabled` handler |
  | `client/src/hooks/useMCPServerForm.ts` | Auto-activate gateway and fetch tools after OAuth; `parseApiError`; `fetchToolsNotification` state |
  | `client/src/hooks/useQuery.ts` | Stabilise `refetch` reference with `useCallback` |
  | `client/src/lib/errorUtils.ts` | Handle `detail` as an object with a `message` property |

  ## Test coverage

  New and updated tests across six files (450 passing, 2 skipped):

  - `inline-notification.test.tsx` — ARIA roles, colour classes, dismiss button, custom `dismissLabel`
  - `useQuery.test.ts` — stable `refetch` reference, GET success/error, `enabled: false`, manual `execute`
  - `errorUtils.test.ts` — all `parseApiError` branches including the new object-detail case
  - `servers.test.ts` — `toggleEnabled`, `fetchToolsAfterOAuth`, `triggerOAuthAuthorization` (popup, postMessage, cancellation, foreign-source filtering)
  - `ServerActionsMenu.test.tsx` — Activate/Deactivate item rendering and callbacks
  - `useMCPServerForm.test.ts` — `fetchToolsNotification` lifecycle (success, nested-detail error, plain-string error, clear); `handleSubmit` error parsing for create and update


a-effort:
@marekdano , do you have testing steps you follow to produce to OAuth states I can reference or would you mind screen recording that flow? Hoping to take a look at error states / etc. 

I saw that there is now the "deactivate" option in the MCP overflow actions (it must have been set as active by default, which is what we want). But, it looks like there may have been a permission error for me to be able to get that to work: 

https://github.com/user-attachments/assets/2a1450a4-b67e-4f8c-a6b1-20f9f1f8ad17
console message

It seems that the issue may be that the error is console logged instead of shown in the UI as feedback for the user? 

--> LLM feedback on this: 

Servers.tsx:124–134 only logs failures:
```
} catch (err) {
  console.error("Failed to toggle server state:", sanitizeError(err));
}
```
If the POST returns a 403 (missing gateways.update), a CSRF error, or any other 4xx, the catch swallows it, refetch() runs, server.enabled is unchanged, and the menu label still reads "Deactivate." From the user's perspective, the click does nothing — which is likely what is being reported as "not wired up."

• Surface the error via a toast or inline notification in the handleToggleEnabled catch block.
• Same pattern applies to handleTest a few lines up — worth fixing there too.
• Check the browser devtools Network tab while clicking to confirm whether the POST is firing and what status it returns.

marekdano:
@a-effort - I've pushed the fix. Now the error displays in an inline notification

client/src/hooks/useMCPServerForm.ts
```diff
@@ -701,10 +710,31 @@ export function useMCPServerForm(gatewayId?: string): UseMCPServerFormReturn {
...
               const oauthResult = await serversApi.triggerOAuthAuthorization(responseGatewayId);
+
+              // Auto-activate the gateway after successful OAuth authorization
+              try {
+                await serversApi.toggleEnabled(responseGatewayId, true);
+              } catch (activateError) {
+                console.error(
+                  "Failed to activate gateway after OAuth:",
+                  sanitizeError(activateError),
+                );
```
gcgoncalves: Should we raise a toast here to let the use know the oauth failed?

client/src/hooks/useMCPServerForm.ts
```diff
@@ -730,42 +760,11 @@ export function useMCPServerForm(gatewayId?: string): UseMCPServerFormReturn {
             resetForm();
           }
```
gcgoncalves: Consider moving this into a separate function, this useCallback is quite long and growing in complexity!

client/src/hooks/useMCPServerForm.ts
marekdano: Good point — the project doesn't have a global toast system yet, but the inline `oauthNotification` already surfaces the error directly in the form (and the form stays open so it remains visible). That said, I've extracted the OAuth flow into a separate `handleOAuthFlow` callback as part of the complexity cleanup (see comment below), so if we do want to add a toast later it'll be easy to add in one place.

client/src/hooks/useMCPServerForm.ts
marekdano: Done — extracted the OAuth popup flow into a dedicated `handleOAuthFlow` `useCallback` (depends only on `resetForm`). `handleSubmit` now just orchestrates save → OAuth/non-OAuth branch and delegates the OAuth side-effects to the new helper. Also cleaned up the catch block to use the existing `parseApiError` utility instead of repeating the extraction inline. All 75 tests still pass.

marekdano: Addressed!

client/src/hooks/useMCPServerForm.ts
marekdano: Good catch - This is the next step when we add the toast component to the React UI. At the moment, we do not have a toast component. I added a TODO comment.

gcgoncalves (approved):

PR #5079, opened 2026-06-05
feat: add MCP Apps support

### LINKED ISSUE #5009: [FEATURE]: Deliver MCP Apps support in ContextForge
### 🧭 Type of Feature

- [x] New component or integration

### 🧭 Epic

**Title:** Deliver MCP Apps support in ContextForge

**Goal:** Ship feature-flagged MCP Apps support for `ui://` resources, MCP UI metadata, AppBridge session/RPC handling, and model-vs-app tool visibility.

**Why now:** MCP Apps support is needed as a concrete product capability. ContextForge should advertise the MCP UI capability through `capabilities.extensions["io.modelcontextprotocol/ui"]`, serve UI resources through normal MCP resource flows, and keep app-only helper tools out of model-facing listings.

---

### 🧑🏻‍💻 User Story 1

**As a:** MCP client or host  
**I want:** ContextForge to advertise MCP Apps support through `capabilities.extensions`  
**So that:** clients can negotiate MCP Apps using the MCP UI capability signal.

#### ✅ Acceptance Criteria
```gherkin
Scenario: MCP Apps capability is advertised during initialize
  Given MCP Apps support is enabled
  And the caller is authorized to use the target server or gateway scope
  When an MCP client calls initialize with compatible client capabilities
  Then the initialize response includes capabilities.extensions["io.modelcontextprotocol/ui"]
  And existing prompts/resources/tools/logging/completions capabilities are preserved
  And capabilities.experimental is not used as the canonical MCP Apps signal

Scenario: MCP Apps capability is not advertised to unauthorized callers
  Given MCP Apps support is enabled
  And the caller is not authorized for the target server or gateway scope
  When the caller invokes initialize
  Then capabilities.extensions["io.modelcontextprotocol/ui"] is omitted
  And no private or team-only MCP Apps availability is leaked
```

---

### 🧑🏻‍💻 User Story 2

**As a:** platform maintainer  
**I want:** MCP method routing to keep unknown MCP Apps methods default-deny  
**So that:** ContextForge can support MCP Apps safely without proxying unknown protocol methods.

#### ✅ Acceptance Criteria
```gherkin
Scenario: Core MCP methods still take precedence
  Given an MCP request uses a core method such as tools/list or resources/read
  When the request is dispatched
  Then the existing core MCP path handles the request
  And MCP Apps validation does not alter core method behavior

Scenario: Unknown MCP Apps-prefixed methods remain default-deny
  Given no supported MCP Apps path owns method io.modelcontextprotocol/unknown
  When a client calls io.modelcontextprotocol/unknown
  Then ContextForge returns JSON-RPC method not found
  And no upstream call is made

Scenario: Unknown extension-prefixed methods remain default-deny
  Given no supported path owns method extensions/unknown
  When a client calls extensions/unknown
  Then ContextForge returns JSON-RPC method not found
  And no upstream call is made
```

---

### 🧑🏻‍💻 User Story 3

**As a:** server administrator  
**I want:** to register and serve MCP Apps UI resources  
**So that:** tools can return interactive UI metadata backed by `ui://` resources.

#### ✅ Acceptance Criteria
```gherkin
Scenario: Register UI resource
  Given an administrator has UI content for a dashboard
  When they register a UI resource with a ui:// URI, MIME type, CSP config, and UI metadata
  Then the resource is stored with audit metadata
  And the resource is associated with the configured owner/team visibility

Scenario: Read UI resource through MCP resource flow
  Given ui://dashboard exists
  And the caller is authorized to read it
  When the caller invokes resources/read for ui://dashboard
  Then the response returns the UI content in MCP resource content format
  And the response includes the metadata needed by an MCP Apps host

Scenario: Deny unauthorized UI resource read
  Given ui://dashboard is private or team-scoped
  And the caller lacks access to that scope
  When the caller invokes resources/read for ui://dashboard
  Then the request is denied
  And an audit event records the denied access
```

---

### 🧑🏻‍💻 User Story 4

**As a:** tool author  
**I want:** tools to declare associated UI resources and app/model visibility  
**So that:** MCP Apps can show interactive views without exposing app-only helper tools to the model.

#### ✅ Acceptance Criteria
```gherkin
Scenario: Tool declares UI resource metadata
  Given tool create_dashboard exists
  And ui://dashboard exists
  When the tool metadata is updated with _meta.ui.resourceUri="ui://dashboard"
  Then tools/list returns the UI metadata to authorized model-facing clients
  And tools/call can return UI-capable results for compatible hosts

Scenario: App-only tools are hidden from model-facing list
  Given helper tool get_dashboard_data is app-visible only
  When a model-facing client invokes tools/list
  Then get_dashboard_data is omitted
  And model-visible tools remain listed normally

Scenario: App session can call app-visible helper tool
  Given an active MCP Apps session is bound to the same server as get_dashboard_data
  And the caller has permission to execute the tool
  When the app invokes tools/call for get_dashboard_data
  Then the call succeeds
  And the result is returned to the app bridge
```

---

### 🧑🏻‍💻 User Story 5

**As a:** security administrator  
**I want:** MCP Apps sessions, AppBridge calls, CSP, and sandbox policy to be enforced  
**So that:** interactive UI content cannot cross server, team, or origin boundaries.

#### ✅ Acceptance Criteria
```gherkin
Scenario: AppBridge enforces same-server boundary
  Given an app session is bound to server-a
  When the app attempts to call a tool from server-b
  Then the call is denied
  And an audit event records a cross-server app call denial

Scenario: Expired or missing app session is denied
  Given no active app session exists for the provided session id
  When an app-originated JSON-RPC call is received
  Then the call is denied
  And no tool or resource operation is executed

Scenario: CSP and sandbox policy are validated
  Given an administrator registers a UI resource
  When the CSP, sandbox, or permissions policy is unsafe
  Then the resource is rejected
  And the policy decision is recorded
```

---

```mermaid
flowchart TD
    Client[MCP Client / Host] -->|initialize| Gateway[ContextForge]
    Gateway --> InitCaps[capabilities.extensions]
    InitCaps --> UICap[io.modelcontextprotocol/ui]

    Client -->|tools/call| ToolSvc[Tool Service]
    ToolSvc --> UIMeta[_meta.ui.resourceUri]
    Client -->|resources/read ui://dashboard| ResourceSvc[Resource Service]
    ResourceSvc --> UIStore[UI Resource Store]

    AppView[Sandboxed App View] -->|AppBridge JSON-RPC| AppBridge[AppBridge Handler]
    AppBridge --> Policy[Same-server + RBAC + Token Scope]
    Policy --> ToolSvc
```

---

### 🔗 MCP Standards Check

- [x] Change adheres to current MCP capability negotiation using `capabilities.extensions`
- [x] MCP Apps uses the official `io.modelcontextprotocol/ui` identifier
- [x] UI resources are represented as `ui://` resources and served through MCP resource flows

No intentional deviations. Legacy `capabilities.experimental` may be preserved as compatibility metadata, but it is not the canonical MCP Apps signal.

---

### 🔄 Alternatives Considered

- Implement MCP Apps directly without feature gating. Rejected because the rollout should stay disabled by default until hosts and deployment policies are ready.
- Expose app-only helper tools through normal model-facing `tools/list`. Rejected because helper tools should only be callable through validated AppBridge sessions unless explicitly marked model-visible.
- Proxy all unknown MCP Apps or extension-prefixed methods. Rejected because unknown method routing must remain default-deny and explicitly owned.

---

### 📓 Additional Context

Related work:

- Parent epic: #2527
- ADR approval gate: #4957
- Grouped delivery stories: #4974, #4975, #4978
- Companion client-side rendering work: #4928

This issue should finish with feature-flagged MCP Apps support, including capability advertising, UI resource metadata validation, AppBridge sessions, and app/model tool visibility controls.

vishu-bh:
## Related Issue
Closes #5009

## Summary
Adds feature-flagged MCP Apps support for `ui://` resources, MCP UI metadata, AppBridge session/RPC handling, and model-vs-app tool visibility.

This change keeps MCP Apps disabled by default, advertises the MCP UI capability only for authorized initialization requests, and validates UI resource metadata before accepting AppBridge-facing content.

## Type of Change
- [ ] Bug fix
- [x] Feature / Enhancement
- [x] Documentation
- [x] Refactor
- [ ] Other (describe below)

## Checklist
- [x] Code formatted
- [x] Documentation updated

## Notes
Security-sensitive paths require owned MCP sessions, explicit server binding, app-visible tools, and stricter UI resource metadata policies for CSP and sandbox settings.

Documentation is under `docs/docs/architecture/mcp-apps.md` and linked from the architecture pages.


ja8zyjits:
LGTM, but there are some doubts regarding the implementation and naming. Will approve once we discuss that.

ja8zyjits:
## Priority Warnings

### 1. Missing unauthenticated deny-path tests
- **Files:** `mcpgateway/main.py:10051`, `mcpgateway/main.py:10111`
- **Issue:** No tests prove `/mcp/apps/sessions` and `/mcp/apps/sessions/{id}/rpc` reject unauthenticated requests before resource/tool execution.
- **Why it matters:** Security-sensitive transport/session endpoints need explicit deny-path regression coverage.
- **Fix:** Add endpoint tests for unauthenticated POST requests to both routes.

### 2. Missing authorization-failure test for UI resource access
- **File:** `mcpgateway/main.py:10079`
- **Issue:** No test proves session creation fails when caller lacks visibility to target `ui://` resource.
- **Why it matters:** Token scoping is core to Layer 1 security model.
- **Fix:** Add test where `resource_service.read_resource()` denies access and no app session is persisted.

### 3. Missing end-to-end visibility split test
- **Files:** `mcpgateway/main.py:780`, `mcpgateway/main.py:10001`, `mcpgateway/main.py:10158`, `mcpgateway/services/tool_service.py`
- **Issue:** No single test proves same tool is hidden from `tools/list`, rejected by normal `tools/call`, and accepted through AppBridge when app-visible.
- **Why it matters:** Feature introduces dual execution paths with different visibility rules.
- **Fix:** Add one integration-oriented flow covering discovery, normal call, and AppBridge call.

### 4. Docs and code version mismatch
- **Files:** `docs/docs/architecture/mcp-apps.md:48`, `mcpgateway/services/mcp_apps.py:23`
- **Issue:** Docs show `"version": "2025-06-02"` while code defines `MCP_UI_DEFAULT_VERSION = "2026-01-26"`.
- **Why it matters:** Spec drift confuses adopters and reviewers.
- **Fix:** Align docs and code to same capability version.

### 5. Docs omit supported CSP directives
- **Files:** `docs/docs/architecture/mcp-apps.md:157-165`, `mcpgateway/services/mcp_apps.py:25-37`
- **Issue:** Docs list 8 directives; code supports 4 more: `baseUriDomains`, `connectDomains`, `frameDomains`, `resourceDomains`.
- **Why it matters:** Incomplete docs create bad configs and support churn.
- **Fix:** Update docs or remove unintended directives from code.

### 6. No cleanup path for expired AppBridge sessions
- **File:** `mcpgateway/services/mcp_apps.py`
- **Issue:** Sessions have TTL but no explicit cleanup task.
- **Why it matters:** Expired rows can accumulate in `mcp_app_sessions`.
- **Fix:** Add scheduled cleanup similar to observability trace cleanup.

### 7. AppBridge endpoints lack explicit rate limiting
- **Files:** `mcpgateway/main.py:10051-10194`
- **Issue:** New session creation and RPC endpoints do not show explicit per-user rate limiting.
- **Why it matters:** Abuse surface for session churn and tool invocation.
- **Fix:** Apply rate limiting consistent with other execution paths.

---

## Medium-Priority Code Quality Findings

### 1. Duplicate MCP session ID extraction
- **Files:** `mcpgateway/main.py:7674`, `mcpgateway/main.py:10083`, `mcpgateway/main.py:10149`
- **Issue:** Same extraction logic repeated across endpoints.
- **Fix:** Extract helper like `extract_mcp_session_id(request, body)`.

### 2. Complex resource serialization function
- **File:** `mcpgateway/services/mcp_apps.py:283-350`
- **Issue:** `serialize_resource_content_for_mcp()` is large and branch-heavy.
- **Fix:** Split into type-specific serializers.

### 3. Bare broad exception in AppBridge RPC handler
- **File:** `mcpgateway/main.py:10201`
- **Issue:** `except Exception:` too broad.
- **Fix:** Catch expected exception types separately; log unexpected failures.

### 4. Missing helper return type hints
- **File:** `mcpgateway/services/mcp_apps.py`
- **Functions:** `mcp_apps_enabled`, `extension_metadata_value`, `mcp_ui_metadata`, `merge_mcp_protocol_meta`, `_protocol_ui_metadata`
- **Fix:** Add explicit return annotations.

### 5. Session creation path lacks explicit failure logging
- **Files:** `mcpgateway/main.py:10051-10127`, `mcpgateway/services/mcp_apps.py:356-381`
- **Issue:** Session creation/commit failures not clearly logged or wrapped with rollback context.
- **Fix:** Add structured logging and explicit rollback handling around commit failures.

### 6. Synthetic session-validation tests
- **File:** `tests/unit/mcpgateway/test_mcp_apps_security.py:534-621`
- **Issue:** Several tests stub DB to always return `None`, so claimed rejection reasons are not actually distinguished.
- **Fix:** Assert query construction or use distinct returned session objects per branch.

### 7. Predicate-only registry tests
- **File:** `tests/unit/mcpgateway/test_mcp_method_registry.py`
- **Issue:** Some tests only assert helper booleans instead of observable JSON-RPC behaviour.
- **Fix:** Replace with handler-level assertions.

### 8. Existing metadata migration has no backfill guidance
- **File:** `mcpgateway/alembic/versions/b6c7d8e9f0a1_add_mcp_app_metadata.py`
- **Issue:** New nullable `extension_metadata` columns are added without migration/backfill guidance for existing tools/resources.
- **Fix:** Document rollout path or add data migration step.

---

ja8zyjits:
Hi @vishu-bh 
These are minor issues.
## 🚨 Blocking Issues

### Critical: UUID Format Mismatch
- [ ] AppBridge session example shows UUID-with-hyphens, implementation returns 32-char hex without hyphens
  - **Doc**: `docs/docs/architecture/mcp-apps.md:244`
  - **Code**: `mcpgateway/db.py:19`, `mcpgateway/services/mcp_apps.py:364`
  - **Fix**: Update example `appSessionId` to match `uuid.uuid4().hex` format OR change implementation to emit hyphenated UUIDs

---

## ⚠️ Warnings

### Configuration Documentation Gap
- [ ] MCP Apps cleanup settings missing from config docs
  - **Doc**: `docs/docs/manage/configuration.md:17-18`
  - **Code**: `mcpgateway/config.py:14-16`
  - **Fix**: Add `MCPGATEWAY_MCP_APPS_SESSION_CLEANUP_ENABLED`, `MCPGATEWAY_MCP_APPS_SESSION_CLEANUP_INTERVAL_SECONDS`, `MCPGATEWAY_MCP_APPS_SESSION_CLEANUP_BATCH_SIZE` with defaults/bounds

---
If you think this is not a big issue, I will LGTM it right away

vishu-bh:
> Hi @vishu-bh These are minor issues.
> 
> ## 🚨 Blocking Issues
> ### Critical: UUID Format Mismatch
> * [ ]  AppBridge session example shows UUID-with-hyphens, implementation returns 32-char hex without hyphens
>   
>   * **Doc**: `docs/docs/architecture/mcp-apps.md:244`
>   * **Code**: `mcpgateway/db.py:19`, `mcpgateway/services/mcp_apps.py:364`
>   * **Fix**: Update example `appSessionId` to match `uuid.uuid4().hex` format OR change implementation to emit hyphenated UUIDs
> 
> ## ⚠️ Warnings
> ### Configuration Documentation Gap
> * [ ]  MCP Apps cleanup settings missing from config docs
>   
>   * **Doc**: `docs/docs/manage/configuration.md:17-18`
>   * **Code**: `mcpgateway/config.py:14-16`
>   * **Fix**: Add `MCPGATEWAY_MCP_APPS_SESSION_CLEANUP_ENABLED`, `MCPGATEWAY_MCP_APPS_SESSION_CLEANUP_INTERVAL_SECONDS`, `MCPGATEWAY_MCP_APPS_SESSION_CLEANUP_BATCH_SIZE` with defaults/bounds
> 
> If you think this is not a big issue, I will LGTM it right away

I agree they were minor, but updated the documentation to match code changes

msureshkumar88 (changes requested):
Thanks for taking this on — MCP Apps is a substantial feature and the overall design is sound: disabled by default, capability advertised only for authorized init requests, and a session-owned AppBridge RPC path with real RBAC/token-scoping checks. The unit test suite is also genuinely thorough (`test_mcp_apps_security.py` alone exercises a lot of real validation logic, not placeholders).

I did find two issues that undercut the security invariants the PR itself states, plus a handful of smaller items. Verified each against the code directly.

## Blocking

**1. App-only tools are reachable via plain `tools/call`, bypassing AppBridge session/ownership checks** (`mcpgateway/transports/streamablehttp_transport.py:1936-1947`)
Any client that simply declares `capabilities.extensions["io.modelcontextprotocol/ui"]` at `initialize` — no auth beyond that — gets `require_model_visible=False` on every subsequent `tools/call`, and `require_app_visible` is never set (defaults `False`). In `tool_service.py:4655-4659`, when both flags are `False`, no visibility check runs at all, so an app-only tool executes with none of the session-ownership/server-binding checks that `/mcp/apps/sessions/{id}/rpc` correctly enforces. Worth considering: only relax `require_model_visible` for calls actually routed through the AppBridge session path, not based on a self-declared capability flag on the standard tool-call route.

**2. CSP/sandbox validation is skipped for tools/resources synced from federated gateways** (`mcpgateway/services/gateway_service.py:1530,1557,5290,5372,5386`)
`validate_ui_resource()`/`validate_extension_metadata()` run on every local resource/tool write (`resource_service.py:570,883,3128`; `tool_service.py:1963,2604,6446`), but the gateway-federation sync/refresh path assigns `extension_metadata` straight from the upstream payload via `optional_extension_metadata()`, which only checks `isinstance(value, dict)` — no CSP/sandbox/MIME validation. Since untrusted upstream UI is exactly the threat model MCP Apps needs to guard against, a compromised federated server could currently ship an `unsafe-inline` CSP or empty sandbox straight into the local DB. Might be worth routing this path through the same validation as the local writes.

Related to both: `_validate_csp` (`mcp_apps.py:180-196`) only rejects the literal `"*"` source and only checks `unsafe-inline`/`unsafe-eval` on `script-src` — confirmed by direct execution that `{"connect-src": ["https://*.evil.example"]}` and `{"style-src": ["'unsafe-inline'"]}` both pass validation today. Broadening the wildcard check (substring, not exact match) and extending the unsafe-inline/eval check to other directives would close this alongside #2. It might also be worth routing this through the existing `ContentSecurityService` (already used elsewhere in `resource_service.py`) rather than the parallel validation logic in `mcp_apps.py`, so admin-configured allowlists apply consistently.

## Worth a look

**3. `ui://` becomes a reserved/blocked URI scheme even with the feature flag off** (`mcp_apps.py:243-246`) — `validate_ui_resource()` runs unconditionally on every resource write, and now raises for any `ui://` URI when `MCPGATEWAY_MCP_APPS_ENABLED=false` (the default), where previously it was just an ordinary usable prefix (confirmed against `tests/e2e/test_admin_apis.py` on `main`, which used `ui://` freely before this branch's forced edit). This is a narrow but real breaking change for any deployment already using that prefix — probably worth a callout in the PR description, or gating the rejection on the flag as well.

**4. Retry logic in `resource_service.py:2020-2145` rebuilds the full upstream session (not just the read) on empty response, and swallows exceptions after partial success at debug level** — once `resource_received=True`, any later exception returns the stale `resource_text` with only a debug log, which could mask real post-fetch failures from anyone not watching debug logs.

**5. Possible same-server tool name collision** (`tool_service.py:4285-4289`) — `or_(DbTool.name == name, DbTool.original_name == name)` isn't constrained to prevent tool A's `name` colliding with tool B's `original_name` on the same server. The downstream visibility check does block cross-audience exposure, but same-audience collisions could still resolve to the wrong tool. Might be worth a regression test either way.

**6. `mcp_method_registry.is_mcp_apps_method` doesn't check `mcp_apps_enabled()`, unlike `is_known_method`** — no live impact today since nothing calls it outside its own test, but worth tightening before the registry becomes load-bearing for real dispatch.

## Minor / non-blocking

- `docs/docs/architecture/roadmap.md` (+642 lines) and the `Containerfile.lite` gunicorn change look unrelated to MCP Apps — might be cleaner split into separate PRs.
- No e2e/integration test currently exercises the flag-enabled AppBridge session flow end-to-end; the existing e2e diff is just the one-line URI workaround for #3.

Happy to take another pass once the two blocking items are addressed.

vishu-bh:
Thanks @msureshkumar88 addressed comments. Can you please check once you have time.

msureshkumar88 (changes requested):
Thanks for the fast turnaround on the two blocking items — confirmed both directly against current HEAD:

- `tools/call` now sets `require_model_visible=True` (`streamablehttp_transport.py:1946`), and `is_model_visible_tool` correctly denies app-only tools on the standard route.
- Federated tool/resource sync now runs the same `validate_extension_metadata`/`validate_ui_resource` checks as local writes (`gateway_service.py:222-233`, wired into `_update_or_create_tools`/`_update_or_create_resources`), and the CSP wildcard/unsafe-inline checks were broadened to substring + all directives (`mcp_apps.py:192-195`) — nice, that covers the related item too.

While verifying those, I cross-checked the implementation against the official spec (SEP-1865 / apps.mdx, Stable) and found a few more things worth a look before this goes out further, using two independent passes (wire-format shapes, and the AppBridge trust model) to cross-verify each other.

## Blocking

**1. CSP metadata uses two incompatible shapes in the same object, and the docs example only uses the one a real Host won't read** (`mcp_apps.py:32-47`, `docs/docs/architecture/mcp-apps.md:129-134`)
The spec defines exactly four `csp` keys — `connectDomains`/`resourceDomains`/`frameDomains`/`baseUriDomains` (plain domain arrays) — and a Host's reference algorithm reads only those (`csp?.resourceDomains?.join(' ')`, etc.). `_ALLOWED_CSP_DIRECTIVES` also accepts raw CSP directive names (`script-src`, `connect-src`, `default-src`...) as siblings in the same `csp` object, and the architecture doc's own worked example uses *only* the raw-directive form. Traced through what happens on a real spec-compliant Host: it looks for `csp.resourceDomains` etc., finds them undefined on a ContextForge-authored resource, and falls back to its restrictive default — fail-safe, but the operator's intended external-domain allowlist silently never takes effect and the UI breaks with no error surfaced anywhere. Worth deciding on one shape (probably the spec's) and updating the docs example and validator together, otherwise this passes our own validation but won't interop correctly.

**2. AppBridge has the View talking to the gateway directly over HTTP with a bearer token, which is a different transport/trust model than the spec's Host-mediated one — and it's not clear how the View gets that token safely** (`docs/docs/architecture/mcp-apps.md` AppBridge Flow diagram; `main.py:10115-10269`)
Spec's model: the View (sandboxed iframe) only ever exchanges `postMessage` JSON-RPC with the Host; the Host is "in control," validates/can block messages, and forwards approved calls over its *own* existing MCP session. The View never holds Server credentials — that's the security property the sandbox exists to provide. This PR's own sequence diagram instead has the View iframe issuing `POST /mcp/apps/sessions` and `POST /mcp/apps/sessions/{id}/rpc` directly, both gated by `Authorization: Bearer <token>` via `get_current_user_with_permissions`. I couldn't find any code (checked `mcpgateway/`, `mcp-servers/`) that safely delivers a token into the sandboxed View, or any Host-side relay translating `postMessage` traffic into these REST calls. If the intent is for the Host to inject the user's real gateway bearer token into the View's JS context, a malicious/compromised server-authored `ui://` HTML — exactly the untrusted content the sandbox is meant to contain — could exfiltrate it, constrained only by that same resource's self-declared `connectDomains`/`connect-src`, which the same server author controls. Since this path also looks unreachable from any unmodified real Host today (none of them know this REST API), it's not an active exploit right now, but I'd want either a scoped-down/ephemeral token type for View-originated calls (distinct from the user's real bearer token) or a documented Host-side relay before this is something a Host integration would actually wire up to.

## Worth a look

**3. `mimeType` accepts bare `text/html`, but the spec's type is `text/html;profile=mcp-app`** (`mcp_apps.py:248`, doc example at `mcp-apps.md:125`)
Real Hosts may sniff the `profile=mcp-app` parameter to decide whether to render the Apps path at all. Resources registered with the bare type (as the docs example itself uses) risk not being recognized as MCP Apps content by a strictly spec-compliant Host.

**4. Visibility default is inconsistent between what's emitted on the wire and what AppBridge enforces** (`mcp_apps.py:273-276` vs `apply_tool_meta` `:300-313`)
Spec default when `visibility` is omitted is `["model","app"]`; this implementation's `tool_audience()` defaults to `["model"]` only for enforcement, but `apply_tool_meta` omits the `visibility` key entirely from the outgoing `_meta.ui` in that case — so a real Host applying the spec default would treat the tool as app-callable, while ContextForge's own AppBridge denies it. Mostly hits federated tools with no explicit visibility metadata. Fails closed (safe), but worth either matching the spec default or making the wire payload explicit about the narrower default so behavior doesn't diverge from what's advertised.

Still outstanding from the previous round (all non-blocking, restating so they don't get lost): `ui://` remains a blocked scheme even with `MCPGATEWAY_MCP_APPS_ENABLED=false` (breaking change for prior `ui://` users), the same-server tool name/`original_name` collision in `tool_service.py:4285-4289`, and the debug-level exception swallow after partial success in `resource_service.py`'s retry path.

Happy to take another pass once the CSP shape and AppBridge token question are settled — the rest is in good shape.

msureshkumar88 (changes requested):
Thanks for addressing both rounds of feedback so far — I confirmed all four blocking items from the two previous rounds directly against current HEAD (`2bd56dd73`):

- `require_model_visible=True` is now hardcoded on plain `tools/call` (`streamablehttp_transport.py:1953`), enforced in `tool_service.py:4657`.
- Federated tool/resource sync now routes through the same `validate_extension_metadata`/`validate_ui_resource` checks as local writes (`gateway_service.py` create + refresh paths).
- CSP metadata now uses only the four spec keys (`connectDomains`/`resourceDomains`/`frameDomains`/`baseUriDomains`), matching the doc example.
- The AppBridge trust model is now clearly documented as Host-mediated, with the sequence diagram fixed to show Host→Gateway calls only and an explicit "the View must not receive or store the user's gateway bearer token."

Nice work — the core design continues to hold up well (deny-by-default flag, session-owned AppBridge RPC, real RBAC/token-scoping checks).

I did a fresh pass over the full branch diff (`origin/main...HEAD`) focused on what's new since the last review and found a few things worth addressing before this merges, plus the non-blocking items from earlier rounds that are still open.

## Blocking

**1. New resource validation can leak as an unhandled 500 instead of a clean 4xx** (`mcpgateway/main.py:6165`, `6351-6368`; `resource_service.py:3128`)
`create_resource`/`update_resource` only catch `ResourceError`, pydantic's `ValidationError`, `IntegrityError`, `ResourceURIConflictError`, `ContentSizeError`, `ContentTypeError`. `MCPAppsValidationError` is a bare `ValueError` and isn't in that list, so malformed MCP Apps metadata on a resource create/update (bad CSP, bad visibility value, missing sandbox on a `ui://` resource, etc.) propagates uncaught and FastAPI returns a generic 500. Worth comparing against `create_tool` (`main.py:5625`), which explicitly catches `isinstance(ex, (ValidationError, ValueError))` → 422 — the resource routes never got the equivalent treatment for this new validation path.

**2. CSP/sandbox validation is still a standalone reimplementation, not routed through the existing security infrastructure** (`mcp_apps.py:32-51`, `173-208`)
This was actually your own suggestion in the last round ("might also be worth routing this through the existing `ContentSecurityService`... so admin-configured allowlists apply consistently"). Only the wildcard/directive-scope fix landed — the architectural consolidation didn't happen. Right now there are two independently-maintained blocklists (this one, and `ContentSecurityService`'s malicious-pattern detection used elsewhere in `resource_service.py`), and a future fix to one won't automatically protect the other. Given this is explicitly a security boundary for untrusted UI content, would like to see this consolidated before merge rather than carried as follow-up debt.

**3. First-time gateway registration has no per-item error isolation; the refresh path does** (`gateway_service.py:1478-1533` vs `5274-5365`)
`register_gateway()` builds the tool/resource lists via plain list comprehensions calling the new `_validated_tool_extension_metadata`/`_validated_resource_extension_metadata` helpers with no try/except — one malformed item from the upstream server aborts registration of the entire catalog. The refresh path (`_update_or_create_tools`/`_update_or_create_resources`) wraps the same call per-item and just skips the bad one and continues. Worth aligning first-time registration with the refresh path's resilience.

## Worth a look

**4. Rust fast-path tool listing doesn't check the feature flag** (`tool_service.py:3064`)
`list_server_mcp_tool_definitions` filters via `is_model_visible_tool()` unconditionally, unlike `filter_model_visible_tools()` (used by the standard `tools/list` path) which explicitly no-ops when `mcp_apps_enabled()` is false. With the flag off but a tool carrying stale or federated `visibility:["app"]` metadata, this internal seam silently drops the tool from the list while the standard path still returns it — same server, two different inventories depending on which internal path served the request.

**5. AppBridge session cleanup holds one long transaction across all batches** (`mcp_apps.py:497-521`)
`cleanup_expired_sessions`'s `while True` loop does N batched select+delete round trips but only commits once at the end. Under a large expired-session backlog this holds locks/connection open for the whole pass instead of releasing per batch — might be worth a `db.commit()` inside the loop.

**6. Missing deny-path tests for RBAC/wrong-team on the new AppBridge endpoints**
`test_mcp_apps_security.py` covers unauthenticated/wrong-user/wrong-server/feature-disabled well, but I didn't find a case where the `@require_permission("tools.execute")`/`"resources.read")` decorators actually deny a token, or a wrong-team token-scoping case for the resource-visibility check that gates session creation. Given the project's own guidance that security-sensitive changes need deny-path coverage for insufficient-permissions and wrong-team scenarios, seems worth adding.

**7. New resource-read retry loop fully reconnects instead of retrying just the read** (`resource_service.py:2000-2145`)
On an empty upstream response, `connect_to_sse_session`/`connect_to_streamablehttp_server` are called again wholesale — new TCP connection plus a fresh MCP `initialize` handshake — rather than retrying the read on the already-established session. Adds real latency for what should be a cheap retry.

## Restated from earlier rounds (still open, non-blocking)

- `ui://` remains a globally-blocked scheme even when `MCPGATEWAY_MCP_APPS_ENABLED=false` — still a breaking change for any deployment already using that prefix (the e2e test still carries the `ui://` → `resource://` workaround rather than a fix). Given it's a real behavior change for existing users, might be worth at least calling out explicitly in the PR description/changelog if not fixing outright.
- Same-server tool `name`/`original_name` collision (`tool_service.py:4285-4290`) can still resolve to the wrong tool via the visibility-priority tiebreak.
- Debug-level exception swallow after partial success in the resource retry path could still mask real post-fetch failures.

## Minor / non-blocking

- `docs/docs/architecture/roadmap.md` (+640 lines, general project roadmap with only one incidental MCP-Apps mention) is still bundled into this PR — flagged last round as probably cleaner split out, still there.
- `mcp_method_registry.is_mcp_apps_method`/`get_mcp_apps_methods` are defined and unit-tested but have no production callers — dispatch only calls `is_known_method()`.
- `is_model_visible_tool`/`is_app_visible_tool` duplicate the same tool/dict extraction ternary verbatim; a small shared helper would remove the duplication.

Happy to take another pass once the three blocking items are in.

msureshkumar88:
To make the consolidation concrete (rather than "route it through the existing service" as a vague follow-up), here's specifically what would need to happen:

**Current state:**
- `mcp_apps.py` maintains its own hardcoded allow/blocklists: `_ALLOWED_CSP_DIRECTIVES`, `_ALLOWED_SANDBOX_TOKENS`, `_BLOCKED_SOURCE_PREFIXES` (`javascript:`, `file:`, `data:`), plus inline checks for `'unsafe-inline'`/`'unsafe-eval'` and wildcards (`mcp_apps.py:32-51`, `173-208`).
- `ContentSecurityService` (`content_security.py`) already does overlapping work — `detect_malicious_patterns()` blocks `javascript:`/XSS patterns via `settings.content_blocked_patterns`, which is admin-configurable and centrally maintained.
- These two never call each other. An admin who tightens `CONTENT_BLOCKED_PATTERNS` today gets zero additional protection on MCP Apps CSP/sandbox validation — the two blocklists can drift silently.

**What's required to actually consolidate (pick one before merge):**

1. **Minimal wiring**: In `_validate_csp()`, after the existing structural checks, run each CSP source string through `get_content_security_service().detect_malicious_patterns(source, content_type="MCP Apps CSP source")` so admin-configured blocked patterns apply. This doesn't require new API surface on `ContentSecurityService`, just a call from `mcp_apps.py`.
2. **Proper API surface**: Add a `ContentSecurityService.validate_csp_source()` (or similar) method that owns the `unsafe-inline`/`unsafe-eval`/wildcard/scheme-prefix logic, and have `mcp_apps.py` delegate to it instead of keeping its own copy of `_BLOCKED_SOURCE_PREFIXES`. This is the cleaner long-term shape but is more surface area to review before merge.
3. **If neither lands now**: at minimum, file a tracking issue with these two file/line references so it's not lost as vague debt, and note in `mcp_apps.py` that `_BLOCKED_SOURCE_PREFIXES` intentionally duplicates `ContentSecurityService` pattern coverage pending consolidation.

Given this is a security boundary for untrusted UI content, I'd lean toward option 1 as the minimum bar for this PR — it's a small diff and closes the "admin config doesn't apply" gap without redesigning either service's API.

msureshkumar88 (changes requested):
## E2E verification against a live gateway process (closes #5009)

Ran a real end-to-end pass against this branch (`42b44a6f3`) — not TestClient/mocks, an actual `uvicorn` process, real HTTP requests, a fresh SQLite DB migrated with `alembic upgrade head`, and real JWT auth — to verify the acceptance criteria in #5009 against the running app. Script + logs available on request; summarizing the setup and results here.

**Environment**
```
HOST=127.0.0.1  PORT=8123
DATABASE_URL=sqlite:///<fresh temp db>
AUTH_REQUIRED=true  MCP_REQUIRE_AUTH=true
MCPGATEWAY_MCP_APPS_ENABLED=true   (and a second run with =false)
MCPGATEWAY_MCP_APPS_SESSION_TTL=900
PLUGINS_ENABLED=false
```
```bash
alembic upgrade head
python -m uvicorn mcpgateway.main:app --host 127.0.0.1 --port 8123
# admin JWT via mcpgateway.utils.create_jwt_token, then real HTTP calls
# (POST /servers, /resources, /tools, real MCP initialize/tools/list/resources/read
#  over /servers/{id}/mcp/, POST /mcp/apps/sessions, POST /mcp/apps/sessions/{id}/rpc)
```

**Result: 15/20 checks passed. Requesting changes** — capability advertising (User Story 1), unknown-method default-deny (User Story 2), CSP/sandbox rejection (User Story 5), and `resources/read` for `ui://` all verified working correctly end-to-end. Three real gaps found that a live run catches but the current test suite (which drives these two endpoints via `.__wrapped__()` direct calls or an isolated `FastAPI()` sub-app, not the real `app`) can't:

**1. `POST /mcp/apps/sessions` and `POST /mcp/apps/sessions/{app_session_id}/rpc` are unreachable in the running application.**
`app.mount("/mcp", app=mcp_transport_app.handle_streamable_http)` (`main.py:13121`) claims the entire `/mcp/*` prefix at the ASGI level. Both new AppBridge routes live under that same prefix and are shadowed — every request to them returns a bare Starlette `404 {"detail":"Not found"}`, never reaching `create_mcp_app_session`/`handle_mcp_app_session_rpc`. Confirmed against the live process:
```
POST /mcp/apps/sessions          -> 404 {"detail":"Not found"}
POST /mcp/apps/sessions/x/rpc    -> 404 {"detail":"Not found"}
POST /_internal/mcp/tools/call   -> 403 (different utility_router route, no /mcp-prefix collision — reaches the handler fine)
POST /mcp/                       -> 200 (the Mount itself works)
```
`tests/unit/mcpgateway/test_mcp_apps_security.py`'s one HTTP-level test for these routes (`test_appbridge_routes_reject_unauthenticated_before_execution`) builds `route_app = FastAPI(); route_app.include_router(main_mod.utility_router)` — a standalone app with no `/mcp` mount — so it can't see this collision. Every other test in that file calls `main_mod.create_mcp_app_session.__wrapped__(...)` / `handle_mcp_app_session_rpc.__wrapped__(...)` directly, bypassing HTTP routing entirely. As it stands, the AppBridge session/RPC flow — the actual mechanism described in the PR's own sequence diagram — can't be invoked by any real client hitting this gateway.

**2. The PR's own documented example resource can't be created via `POST /resources`.**
`docs/docs/architecture/mcp-apps.md:126` shows `"content": "<!doctype html><html>...</html>"`. Posting that exact shape gets rejected by the pre-existing `ResourceCreate.validate_content` check (`schemas.py:1984`, `DANGEROUS_HTML_PATTERN`, which blocks `<html>`, `<head>`, `<body>` tags among others) — `422 "Content contains HTML tags that may cause display issues"`. A fragment-only body (e.g. `<div id="app">...</div>`) passes, but that's undocumented, and a normal full HTML document — the natural shape for an MCP Apps UI resource — is rejected today.

**3. `ui://` is still a globally-blocked scheme with the feature flag at its (default) disabled state**, contradicting #5009's own "No breaking changes to existing MCP-compliant integrations" acceptance line. Confirmed live:
```
$ MCPGATEWAY_MCP_APPS_ENABLED=false
POST /resources {"uri":"ui://pre-existing-legacy-resource","mimeType":"text/plain","content":"..."}
-> 400 {"detail":"Failed to register resource: MCP Apps UI resources are disabled"}
```
This was flagged as non-blocking across all three review rounds; flagging again here since it directly contradicts a written acceptance criterion, not just a style preference.

Everything else checked out cleanly: capability advertisement scoped to authorized/accessible servers only, `experimental` not used as the canonical signal, unknown `io.modelcontextprotocol/*` and `extensions/*` methods correctly return JSON-RPC `-32601`, app-only tools correctly hidden from model-facing `tools/list`, and unsafe CSP (`'unsafe-inline'`) correctly rejected at resource-registration time.

Given #1 makes the core AppBridge flow non-functional against the real app today, requesting changes rather than approving. Happy to re-run the same script once it's addressed.

msureshkumar88 (changes requested):
Nice work closing out round 3/4 feedback — I re-verified all of it live against a real running gateway (not mocks), and it holds up:

- **Mount-shadow fix confirmed**: `/appbridge/sessions` and `/appbridge/sessions/{id}/rpc` (renamed off `/mcp/apps/...`) are reachable and no longer shadowed by the `/mcp` mount.
- **Doc example now validates**: the fragment-based example in `mcp-apps.md` is accepted by `POST /resources`.
- **`ui://` global-block fixed**: confirmed a plain legacy `ui://` resource with no MCP Apps metadata succeeds with the feature flag at its default (disabled) — no longer contradicts #5009's "no breaking changes" criterion.
- **Test-coverage asks (ja8zyjits) all landed**: unauthenticated deny-path coverage for both AppBridge endpoints, the UI-resource authorization-failure test on session creation, and the end-to-end visibility-split test (hidden from `tools/list` → rejected on normal `tools/call` → accepted via AppBridge) are all now in `test_mcp_apps_security.py` and passing.

I patched and ran `scripts/verify-pr-5079-e2e.sh` (it had stale pre-rename routes) against current HEAD — 8/8 checks pass live.

## Blocking

**1. New gap in today's push: app-only tools become callable via plain `tools/call`, bypassing AppBridge entirely**

`streamablehttp_transport.py:1751,1954-1955` now derives `require_client_visible` from `client_supports_mcp_apps()`, which only checks whether the *client itself* declared `capabilities.extensions["io.modelcontextprotocol/ui"]["mimeTypes"]` in its own `initialize` request (`mcp_apps.py:100-115`) — there's no session binding or ownership check backing that claim. In `tool_service.py:4701-4707`, the `require_client_visible` branch admits a tool if it's `is_model_visible_tool OR is_app_visible_tool`, so an app-only tool passes.

Net effect: any MCP client can invoke an app-only tool through the normal `tools/call` path just by declaring the capability at init — no AppBridge session, no `_assert_session_owner_or_admin`, no resource binding. This looks like a partial reopening of round-1 blocking item #1 (previously closed by hardcoding `require_model_visible=True` unconditionally on this path). Could we either drop the `require_client_visible` branch on the plain transport path (keep it `require_model_visible`-only there) or gate `client_supports_mcp_apps()` on something session-bound rather than a self-declared init flag?

**2. CI red — Rust clippy failing**
https://github.com/IBM/mcp-context-forge/actions/runs/29020580509/job/86142516468

**3. Branch has merge conflicts with `main`** — needs a rebase before this can land regardless of review outcome.

## Worth a look (non-blocking, carried over from earlier rounds — status unclear on latest push)

- `mimeType` still accepts bare `text/html` instead of the spec's `text/html;profile=mcp-app` (`mcp_apps.py:248`, `mcp-apps.md:125`)
- Visibility default mismatch: wire payload omits the `visibility` key by default while enforcement defaults to `["model"]`, but spec default is `["model","app"]` (`mcp_apps.py:273-276` vs `300-313`)
- Rust fast-path tool listing (`list_server_mcp_tool_definitions`) doesn't check `mcp_apps_enabled()` (`tool_service.py:3064`)
- AppBridge session cleanup holds one long transaction instead of committing per batch (`mcp_apps.py:497-521`)
- Resource-read retry does a full reconnect (new TCP + MCP handshake) instead of retrying just the read (`resource_service.py:2000-2145`)

## Minor

- `docs/architecture/roadmap.md` (+640 lines, unrelated) still bundled in the diff
- Unused `mcp_method_registry.is_mcp_apps_method` / `get_mcp_apps_methods`
- Duplicated ternary logic between `is_model_visible_tool` / `is_app_visible_tool`

Happy to take another pass once the clippy failure and #1 above are sorted.

msureshkumar88 (changes requested):
Re-verified all three blocking items from the last round directly against current HEAD (`6f9088c89`):

- **App-only tool bypass on `tools/call` — fixed.** `require_client_visible` has been removed entirely rather than re-gated; `call_tool` now hardcodes `require_model_visible=True` unconditionally (`streamablehttp_transport.py:1949-1950`), and `invoke_tool` no longer has a `require_client_visible` branch at all (`tool_service.py`, the `elif require_client_visible and ...` clause is gone). The corresponding test was updated in place (`test_call_tool_apps_client_still_requires_model_visibility`) rather than just deleted, and still asserts `require_model_visible=True` / no `require_client_visible` kwarg. This is a cleaner fix than a re-gate — no self-declared-capability branch left to reopen on the call path.
- **Rust clippy CI failure — fixed.** All 49 checks pass now (`gh pr checks` shows 0 failing).
- **Merge conflicts with `main` — fixed.** PR reports `mergeable: MERGEABLE`.

## New finding while re-verifying

**`tools/list` still exposes app-only tools (including their real `original_name`) to any client that self-declares `capabilities.extensions["io.modelcontextprotocol/ui"]` at `initialize` — no AppBridge session or ownership binding required** (`streamablehttp_transport.py:226-230`, `2319`, `2386`, `2394`)

`_tools_for_client()` branches purely on `apps_client = client_supports_mcp_apps(request_context)`, which — like the just-fixed `call_tool` bug — only checks a self-reported init capability flag, not anything session-bound:

```python
def _tools_for_client(tools: Iterable[Any], *, apps_client: bool) -> List[types.Tool]:
    if apps_client:
        return [_to_mcp_tool(tool, name=_tool_name_for_apps_client(tool)) for tool in tools]
    return [_to_mcp_tool(tool) for tool in filter_model_visible_tools(tools)]
```

When `apps_client=True`, `filter_model_visible_tools()` — the function whose whole job is "Filter out app-only tools for model-facing list operations" (`mcp_apps.py:308-312`) — is skipped entirely, and `_tool_name_for_apps_client()` additionally reveals the tool's real `original_name` for app-only tools (the name that's otherwise intentionally withheld from model-facing listings). So a model/agent client can see the full tool inventory, including app-only helper tools and their real names/descriptions/schemas, just by adding one capability flag to its `initialize` request — no AppBridge session, no `_assert_session_owner_or_admin`, nothing.

Execution is still safe (confirmed: `tools/call` now enforces `require_model_visible=True` regardless of this flag), so this is metadata disclosure rather than an execution bypass — but it defeats the "hidden from model discovery" half of the model-vs-app visibility boundary that's a stated design goal of this PR, via the same class of self-declared-capability trust issue that was just closed on the call path. Worth either dropping the `apps_client` branch here too (always apply `filter_model_visible_tools`, let AppBridge-originated requests get their own listing path) or binding `apps_client` to something session-verified.

## Still open (non-blocking, restating so they don't get lost)

- `mimeType` still accepts bare `text/html` instead of spec's `text/html;profile=mcp-app` (`mcp_apps.py:248`)
- Visibility wire-default mismatch: `apply_tool_meta` omits `visibility` when unset while enforcement defaults to `["model"]`, but spec default is `["model","app"]` (`mcp_apps.py:273-276` vs `300-313`)
- Rust fast-path `list_server_mcp_tool_definitions` doesn't check `mcp_apps_enabled()` (`tool_service.py:3064`)
- AppBridge session cleanup commits once at the end of a batched loop instead of per-batch (`mcp_apps.py:497-521`)
- Resource-read retry does a full reconnect (new TCP + MCP handshake) instead of retrying just the read (`resource_service.py:2000-2145`)
- `docs/architecture/roadmap.md` (+640 lines, unrelated) still bundled in the diff
- Unused `mcp_method_registry.is_mcp_apps_method` — mostly trimmed in `070aacb85`, worth double-checking nothing else references the removed helpers
- Duplicated ternary logic between `is_model_visible_tool` / `is_app_visible_tool`

Given the `tools/list` gap is the same trust pattern as the item that was the sole reason for requesting changes last round, requesting changes again rather than approving — but this is a much smaller fix than the previous rounds and everything else looks solid. Happy to take another pass once it's addressed.

msureshkumar88 (changes requested):
Re-verified all prior rounds against current HEAD (`e5fc9974f`), then ran a live e2e pass (real `uvicorn` process, real SQLite, real JWT, real MCP `initialize` handshake, real HTTP -- not TestClient/mocks) covering the flow that's been a recurring blind spot across this review: full AppBridge session create -> RPC `tools/call`, plus the `tools/list` model/app visibility split.

## Confirmed fixed (all prior rounds)

- `tools/list` no longer leaks app-only tools/`original_name` to a self-declared client (`e5fc9974f`) -- verified both by code and live, including a regression check where a client self-declares `capabilities.extensions["io.modelcontextprotocol/ui"]` with **no** AppBridge session: `tools/list` still hides the app-only tool and plain `tools/call` still denies it. This was the exact bug class reopened in round 5 and closed in round 6 -- confirmed it stays closed live, not just in unit tests.
- All round 1-5 blocking items remain fixed (tools/call bypass, CSP consolidation via `ContentSecurityService`, resource-validation 500s on `create_resource`/`update_resource`, gateway registration error isolation, mount-shadowing, doc example, `ui://` breaking-change gating, clippy, merge conflicts at the time).
- Full AppBridge happy path works end-to-end live: register `ui://` resource (CSP+sandbox) + app-only tool, bind to a server, real `initialize` handshake, `POST /appbridge/sessions` -> real session, `POST /appbridge/sessions/{id}/rpc` `tools/call` -> tool actually executes and returns a result.
- Deny paths (`missing Mcp-Session-Id`, unknown `app_session_id`) both correctly return `-32003` live.

## Blocking

**New gap found while building the live flow test: `create_mcp_app_session` doesn't catch resource-lookup exceptions -> unhandled 500** (`mcpgateway/main.py:10487`, the `resource_service.read_resource(...)` call around line 10521)

`read_resource()` is documented to raise `ResourceNotFoundError`, `ResourceError`, `PluginError`, `PluginViolationError` -- none of which are caught in `create_mcp_app_session`. Reproduced live: creating an AppBridge session for a `ui://` resource that exists but isn't bound to the given `serverId` propagates uncaught, and the client gets a bare `500 {"detail":"An internal error occurred. Please try again."}` instead of a clean 4xx. Traceback from the gateway log:

```
mcpgateway.services.resource_service.ResourceNotFoundError: Resource template not found for 'ui://e2e-flow/widget'
  ...
mcpgateway.services.resource_service.ResourceNotFoundError: Resource template not found for 'ui://e2e-flow/widget'
  (unhandled at POST /appbridge/sessions)
```

No stack trace leaks to the client, so this isn't an info-disclosure issue -- but "resource not accessible for this server" is a routine case (typo'd `resourceUri`, wrong `serverId`, no access), not an exceptional one, and it's happening on the authorization gate for the whole AppBridge flow. This is the same bug class already fixed for `create_resource`/`update_resource` in an earlier round (`MCPAppsValidationError` catch -> 422); this route just doesn't have the equivalent wrapping yet. Would like to see this caught and mapped to a 403/404 before merge, mirroring the existing resource-route pattern.

## Also still open (non-blocking, restating so they don't get lost)

- Same-server tool `name`/`original_name` collision (`tool_service.py:4634`, `or_(DbTool.name == name, DbTool.original_name == name)` still unconstrained to the same server) -- unchanged across all rounds.
- Visibility wire-default mismatch: `apply_tool_meta` (`mcp_apps.py:329`) still defaults omitted `visibility` to `["model"]` on the wire while the spec default is `["model","app"]`. Fails closed, low risk, but still diverges from spec and isn't called out in the docs.
- No committed e2e/integration test exercises the AppBridge session->RPC flow -- everything to date (including this round's verification) has been ad hoc. Happy to contribute the script I used for this pass if useful as a starting point.

## Housekeeping

- Branch currently shows `mergeable: CONFLICTING` / `mergeStateStatus: DIRTY`. Checked with a local merge-tree -- the only conflict is `.secrets.baseline` (timestamp + shifted line numbers), nothing in actual code. Needs a rebase before merge regardless of the review outcome above.

Solid design and the core trust-boundary bugs have all held up under live testing across six rounds now -- this last one is a small, contained fix. Happy to take another pass once it's in.

msureshkumar88 (dismissed):
Re-verified round 7's blocking finding against current HEAD (`415327b3c`).

## Round 7 blocking issue: fixed

`create_mcp_app_session` (`mcpgateway/main.py:10518-10530`) now wraps `resource_service.read_resource()` in `try/except (ResourceNotFoundError, ResourceError)` and maps it to a clean `404 {"detail":"Resource not found"}` instead of an unhandled 500.

Confirmed three ways:
- Code review of the diff -- narrow, correctly scoped catch.
- Unit tests updated (`tests/unit/mcpgateway/test_mcp_apps_security.py`): two existing tests updated to assert the new `HTTPException(404)` instead of the raw `ResourceNotFoundError`, plus a new test covering the `ResourceError` branch specifically.
- Live e2e run (real `uvicorn`, real SQLite, real JWT, real MCP `initialize` handshake -- no mocks): registered a `ui://` resource not bound to any server, minted a real `Mcp-Session-Id` via `initialize` against an unrelated server, then `POST /appbridge/sessions` with that resource/server pair. Result: `404 {"detail":"Resource not found"}`, not a bare 500. Matches the exact repro from the round-7 review comment.

I also checked whether `PluginError`/`PluginViolationError` (also documented as raiseable by `read_resource`, since it invokes plugin hooks with `violations_as_exceptions=True`) are still an uncaught gap here -- they're not: both have app-wide `@app.exception_handler(...)` registrations in `main.py` (`plugin_violation_exception_handler`, `plugin_exception_handler`), so they never reach an unhandled path regardless of route. No new action needed there.

## e2e scripts run against current HEAD

Ran both scripts already present in the working tree, no regressions:
- `scripts/verify-pr-5079-e2e.sh` -- mount-shadow reachability, doc HTML example, legacy `ui://` non-breaking check: **ALL PASSED**
- `scripts/verify-pr-5079-appbridge-flow-e2e.py` -- full tool-visibility split (model vs. app-only), self-declared-capability regression check, full AppBridge session -> RPC `tools/call` happy path, deny paths: **ALL PASSED**

## Still open (non-blocking, carrying forward from round 7 so they aren't lost)

- Same-server tool `name`/`original_name` collision in `tool_service.py:4634` (`or_(DbTool.name == name, DbTool.original_name == name)` unconstrained to the same server) -- unchanged.
- Visibility wire-default mismatch: `apply_tool_meta` (`mcp_apps.py:329`) defaults omitted `visibility` to `["model"]` on the wire, spec default is `["model","app"]`. Fails closed, low risk, but still diverges from spec/docs.
- No committed e2e/integration test for the AppBridge flow -- everything verified this round (and prior rounds) was ad hoc via the scripts in `scripts/`. Worth committing one of these as a real test if the team wants durable coverage.

No new blocking issues found this round. Approving.

ja8zyjits (dismissed):
LGTM

ja8zyjits (approved):
LGTM

PR #5087, opened 2026-06-08
fix(security): suppress Pydantic validation error details in HTTP response

shoummu1:
## Summary

Pydantic `ValidationError` and FastAPI `RequestValidationError` responses were returning full field-level error detail to callers. This is a continuation of #4427, which covered the majority of exception sites across the codebase; the two global validation error handlers in `main.py` were explicitly deferred at the time.

### `mcpgateway/main.py`

- `validation_exception_handler`: updated to return a generic error message; full detail is preserved in server-side logs
- `request_validation_exception_handler`: updated to return a generic error message; in development mode, retains the original `/tools` structured error branch and `fastapi_default_validation_handler` fallback unchanged
- Import order fixed (isort)

### `mcpgateway/utils/error_formatter.py`

- `format_validation_error()`: returns a generic message in production; response shape kept consistent with the existing verbose format to avoid breaking callers
- Defensive initialisation of `user_message` before the errors loop (prevents `NameError` when `error.errors()` returns an empty list)
- `str(loc[-1])` cast on the Pydantic `loc` field — prevents `AttributeError` when Pydantic returns a list index as an integer (e.g. `("tags", 0)`) rather than a string, which would cause `field.title()` to crash in verbose mode
- Docstring updated to reflect both return shapes

### `mcpgateway/admin.py`

- Fixed four handlers (`admin_add_resource`, `admin_edit_resource`, `admin_add_prompt`, `admin_edit_prompt`) that called `format_validation_error()` twice — once for logging and once for the response. Consolidated to a single call per handler; log lines now record the original exception directly.
- Fixed bulk-import error aggregator (`admin_import_tools`) that called `.get("message", ...)` on the formatter result — in production mode the key is `"detail"`, causing the frontend to receive a raw dict repr instead of a readable string. Now checks both keys.

### Tests

- Updated `test_format_validation_error_production_mode` to assert the correct production response shape
- Added `test_format_validation_error_empty_errors_verbose` to cover the defensive `user_message` default
- All existing verbose-path tests updated to pin mode explicitly and remain valid
- Added `test_request_validation_exception_handler_production_suppression` to cover the suppression path with status 422
- Updated e2e fixture (`temp_db`) and three unit tests (`test_admin_edit_gateway_url_validation`, `test_admin_update_gateway_rest_validation_error`, `test_create_server_rejects_non_uuid_associated_tools`) to explicitly enable verbose mode — these tests assert on specific error message content and must opt in to verbose mode now that production mode suppresses details

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-06-23T17:50:59Z",
+  "generated_at": "2026-06-24T10:35:38Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -390,7 +390,7 @@
         "hashed_secret": "d3ac7a4ef1a838b4134f2f6e7f3c0d249d74b674",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 6051,
+        "line_number": 6052,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -398,7 +398,7 @@
         "hashed_secret": "5932862bcd24dd27d0dc0407ec94fe9d6ea24aeb",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 6548,
+        "line_number": 6549,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -406,7 +406,7 @@
         "hashed_secret": "c77c805e32f173e4321ee9187de9c29cb3804513",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 6560,
+        "line_number": 6561,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -414,7 +414,7 @@
         "hashed_secret": "8fe3df8a68ddd0d4ab2214186cbb8e38ccd0e06a",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 6632,
+        "line_number": 6633,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -422,7 +422,7 @@
         "hashed_secret": "93ac8946882128457cd9e283b30ca851945e6690",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 7714,
+        "line_number": 7715,
         "type": "Secret Keyword",
         "verified_result": null
       }
@@ -4178,7 +4178,7 @@
         "hashed_secret": "c377074d6473f35a91001981355da793dc808ffd",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 753,
+        "line_number": 751,
         "type": "Hex High Entropy String",
         "verified_result": null
       }
@@ -4214,7 +4214,7 @@
         "hashed_secret": "d3ecb0d890368d7659ee54010045b835dacb8efe",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 740,
+        "line_number": 738,
         "type": "Secret Keyword",
         "verified_result": null
       }
@@ -5080,7 +5080,7 @@
         "hashed_secret": "4f13f134744a2fadbbe2d624687246347d12fa63",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 2733,
+        "line_number": 2770,
         "type": "Hex High Entropy String",
         "verified_result": null
       }
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-06-17T12:19:05Z",
+  "generated_at": "2026-06-23T17:50:59Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -274,7 +274,7 @@
         "hashed_secret": "b4673e578b9b30fe8bba1b555b7b59883444c697",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1485,
+        "line_number": 1590,
         "type": "Secret Keyword",
         "verified_result": null
       },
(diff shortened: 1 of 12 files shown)


vishu-bh (changes requested):
Thanks @shoummu1  for tightening this up! 

The response-side suppression looks good to me and addresses the deferred global validation handlers from the earlier error-disclosure work.

One thing worth adjusting before merge: the new logging uses raw validation exception strings in a few production paths, for example `logger.warning("Validation error: %s", error)`. With Pydantic v2, str(error) can include submitted values via input_value=..., so sensitive data may be removed from the HTTP response but still land in logs.
Could we log a sanitized validation summary instead, such as request path, error count, loc, and type, while avoiding `str(error), msg, ctx, input, and input_value`? The existing log sanitizers help with control characters/log injection, but I don’t think they fully cover this Pydantic value-disclosure case.
Overall the direction is good; this just keeps the fix from moving the disclosure surface from responses into logs.

vishu-bh:
Hi @shoummu1,

Please let us know if you are planning to address review comment for this PR.

shoummu1:
Thanks @vishu-bh for the review — you're right that `str(ValidationError)` in Pydantic v2 can embed `input_value=...` in the output, which would just move the disclosure from responses into logs.

Addressed in the latest commit (`92932d47b6c`):

- Added `sanitize_validation_error_for_log()` in `error_formatter.py` that calls `.errors()` and extracts only `loc` and `type` per error — explicitly dropping `msg`, `ctx`, `input`, and `input_value`
- Applied it to all the flagged sites: `format_validation_error()`, `request_validation_exception_handler` in `main.py`, and the four admin handlers (`admin_add/edit_resource`, `admin_add/edit_prompt`)
- Added three tests: one asserting submitted values don't appear in the output, one pinning the expected format, and one covering the safe fallback when `.errors()` raises

vishu-bh (dismissed):
Thanks for addressing the comment @shoummu1!!

LGTM 🚀

ja8zyjits:
HI @shoummu1 
Please rebase.

shoummu1:
Hi @ja8zyjits — rebase is done.

ja8zyjits:
## 🚨 Blocking Issues (Must Fix)

### 1. `error_formatter.py` - Docstring Return Type Mismatch
**Location:** `ErrorFormatter.format_validation_error()` docstring (~line 70)

**Current:**
```python
Returns:
    Dict[str, Any]: ``{"detail": str}`` by default, or
        ``{"message": str, "details": [...], "success": bool}`` when verbose mode is enabled.
```

**Problem:** "by default" implies no config needed. Actually production mode (when `should_expose_error_details()` returns False).

**Fix:** Change "by default" → "in production mode"

---

### 2. `main.py` - Missing Docstring Update
**Location:** `request_validation_exception_handler()` docstring (~line 2154)

**Current Returns:**
```python
Returns:
    JSONResponse: A 422 Unprocessable Entity response with error details.
```

**Problem:** Says "with error details" but production mode returns generic message. No mention of mode-dependent behavior.

**Fix:** Document both production (generic) and verbose (detailed) responses.

---

### 3. `admin.py` - Missing Inline Comment
**Location:** `admin_import_tools()` line ~15005

**Code:**
```python
"error": item["error"].get("message") or item["error"].get("detail", str(item["error"]))
```

**Problem:** No comment explains dual-key check. Future maintainer won't know why both keys exist.

**Fix:** Add comment:
```python
# Check both keys: verbose mode uses "message", production mode uses "detail"
```

---

## ⚠️ Warnings (Should Fix)

### W1: Structured Logging Enhancement
**Files:** `error_formatter.py`, `main.py`

**Current:** `logger.warning("Validation error: %s", sanitize_validation_error_for_log(error))`

**Suggestion:** Use structured logging with error count/types as separate fields for easier querying.

**Priority:** Low (enhancement)

---

## 📋 Observations (Nice to Have)

### O1: Module Docstring Example Outdated
**File:** `error_formatter.py` (lines ~10-23)

Commented-out example doesn't show mode-dependent behavior. Either remove or update to show both modes.

### O2: Missing Doctest Example
**File:** `error_formatter.py` - `sanitize_validation_error_for_log()` (~line 293)

Function has comprehensive docstring but no usage example. Consider adding doctest.

### O3: Handler Docstring Example Outdated
**File:** `main.py` - `validation_exception_handler()` (~lines 2141-2150)

Example only checks status code, doesn't show response body differs by mode. Update or remove.

shoummu1:
Hi @ja8zyjits — all blocking issues have been addressed and a couple of additional fixes were made based on further review:

- Double-logging removed — f`ormat_validation_error()` had an internal `logger.warning` that duplicated what the `admin.py` callers were already logging. Removed.
- Silent async test fixed — `test_request_validation_exception_handler_production_suppression `was missing `@pytest.mark.asyncio`, so the assertions were never actually running. Decorator added.

The branch has also been rebased onto current `main`. Please let me know if you find any other issues — happy to address anything else.

ja8zyjits:
Created a followup that i figured from this PR 
https://github.com/IBM/mcp-context-forge/issues/5324

ja8zyjits (dismissed):
LGTM

## 📋 Recommendations

### Optional Enhancements (Non-blocking)

#### 1. Helper Function for Error Extraction
```python
def extract_error_message(error_result: Dict[str, Any]) -> str:
    """Extract error message from formatter result (handles both shapes)."""
    return error_result.get("message") or error_result.get("detail", "Unknown error")
```
- Abstracts dual-shape complexity
- Single maintenance point
- Use in `admin.py` bulk import

#### 2. Type Hints for Response Shapes
```python
ProductionErrorResponse = TypedDict("ProductionErrorResponse", {"detail": str})
VerboseErrorResponse = TypedDict("VerboseErrorResponse", {
    "message": str, "details": List[Dict[str, Any]], "success": bool
})
ErrorResponse = Union[ProductionErrorResponse, VerboseErrorResponse]
```
- Improves IDE autocomplete
- Documents contract explicitly

#### 3. Missing Test: Verbose `/tools` Path
`test_request_validation_exception_handler_production_suppression` covers production. Missing verbose `/tools` structured error branch test.
- **Impact**: Low (verbose path unchanged)
- **Suggested**: Add test for completeness

#### 4. Consider: Bulk Import Error Aggregator Test
`admin_import_tools` fix tested indirectly. Direct test would confirm frontend receives readable strings in both modes.
- **Impact**: Low (likely covered)

---

vishu-bh (dismissed):

ja8zyjits (dismissed):
LGTM
Process followed

ja8zyjits (approved):
LGTM

PR #5090, opened 2026-06-08
feat!: migrate slow-time server to Rust

lucarlig:
## Summary

Migrates the slow-time-server test fixture from Go to Rust and rewires the repository to use the Rust implementation everywhere the Go server was used.

## Scope

- Removes the Go slow-time-server implementation.
- Adds the Rust slow-time-server implementation and keeps it in the root Cargo workspace.
- Repoints compose, resilience Makefile targets, container smoke coverage, performance fixtures, and docs/references to the Rust server and `Containerfile`.
- Preserves the resilience-testing surface: configurable latency, failure-rate simulation, MCP JSON-RPC over HTTP, and health/version/config/stats endpoints.

## Breaking changes

This is a full migration, not an additive replacement. Downstream users of the old Go server image or old Go-specific routes must update.

- Removed old Go endpoints: `/sse`, `/http`, `/api/v1/docs`.
- The replacement server uses the Rust image/build path and the MCP endpoint exposed by the Rust fixture.
- JMeter and compose references have moved from the old `/http` flow to the Rust `/mcp` flow.

## Notes

The PyO3 advisory suppressions in `deny.toml` are included because CI currently fails on those pre-existing workspace advisories. A direct PyO3 0.29 upgrade was tested, but the current `pyo3-stub-gen` stack does not compile cleanly against it yet.


msureshkumar88 (changes requested):
Thanks for putting this together — the Rust server itself is clean, the helper functions are well tested, and the README is clear about it being a test fixture.

A couple of things I'd like to see addressed before merge:

## Blocking

**1. PR description doesn't match the actual diff scope**

The description says this PR "intentionally does not remove the Go slow-time server or rewire compose/Helm profiles yet, so migration can be reviewed separately." The diff actually does all of that: `mcp-servers/go/slow-time-server/` is deleted entirely (~3.4k lines), `docker-compose.yml`/`docker-compose-performance.yml` are repointed to the new Containerfile, the root `Makefile` resilience targets change, ADRs/architecture docs are updated, and the jmeter baseline is rewritten (`/http` → `/mcp`).

Could you update the description to reflect that this is the full migration (removal + rewiring), not an additive change? That also makes this a **breaking change** for anyone using the old Go endpoints (`/sse`, `/http`, `/api/v1/docs`) or image — worth calling out explicitly so reviewers and downstream users aren't surprised.

**2. Unbounded `delay_seconds` can crash the request task**

`mcp-servers/rust/slow-time-server/src/main.rs:476-481`:

```rust
if let Some(value) = delay_seconds {
    if !value.is_finite() || value < 0.0 {
        return Err("delay_seconds must be a finite non-negative number".to_string());
    }
    return Ok(Duration::from_secs_f64(value).min(MAX_DELAY));
}
```

`Duration::from_secs_f64` panics for finite values that overflow `Duration` (e.g. `1e300`). `is_finite()` doesn't guard against this, and `.min(MAX_DELAY)` runs *after* construction, so it's too late. A `tools/call` request with `"delay_seconds": 1e300` will panic that task.

Suggest clamping before constructing the `Duration`, e.g.:

```rust
let capped = value.min(MAX_DELAY.as_secs_f64());
return Ok(Duration::from_secs_f64(capped));
```

Would also be good to add a test for this boundary (`delay_seconds: 1e300` / `f64::MAX`) alongside the existing `delay_seconds_rejects_invalid_values` test.

## Non-blocking / cleanup

- `mcp-servers/rust/slow-time-server/Dockerfile` looks like a leftover duplicate of `Containerfile` (differs only by OCI labels) and isn't referenced by compose, the local Makefile, or `Containerfile.lite` — all of which use `Containerfile`. Worth removing if it's not needed.
- No integration tests exercise the actual `Router` (`/mcp`, `/health`, `/api/v1/*`) via `tower::ServiceExt::oneshot` — only the pure helper functions are unit-tested. A small route-level test suite would have caught the `delay_seconds` issue above and would cover `get_timeout_time`/`get_flaky_time`, which currently have zero coverage.
- `docs/docs/using/servers/go/slow-time-server.md` is deleted but no equivalent page is added under `docs/docs/using/servers/rust/` — the `mcp-servers/README.md` table entry exists, but the user-facing docs tree loses this server's page.
- `CorsLayer::permissive()` and the `/` → `mcp_handler` alias route aren't mentioned in the README — a short note on why they're there (and that this fixture shouldn't be exposed beyond the test network) would help future readers.

## Out of scope for this PR

`deny.toml`'s suppression of `RUSTSEC-2026-0176`/`-0177` (PyO3/numpy advisories) doesn't appear related to the Rust slow-time-server — might be worth splitting into its own PR/commit for a cleaner history.

lucarlig:
Thanks, Suresh. I addressed the blocking items and one cleanup item:

- Updated the PR title/body to reflect that this is the full Go-to-Rust migration, including Go removal, rewiring, and the breaking endpoint/image changes.
- Fixed the `delay_seconds` overflow path by clamping the `f64` value before `Duration::from_secs_f64` is called.
- Added a regression test using `f64::MAX` so the overflow case stays covered.
- Removed the unused duplicate `mcp-servers/rust/slow-time-server/Dockerfile`; `Containerfile` remains the single referenced build file.
- Merged current `origin/main` and resolved the `.secrets.baseline` conflict, so the PR is conflict-free again.

Local verification passed after the merge:

- `cargo fmt --check -p slow-time-server`
- `cargo test -p slow-time-server`
- `cargo clippy -p slow-time-server -- -D warnings`

On the PyO3 suppressions: I kept them in this PR because the branch needs them for current CI. I also tested the direct PyO3 0.29 path, but the current `pyo3-stub-gen` stack does not compile cleanly against it yet, so splitting/removing those suppressions would re-break CI rather than leave the PR mergeable. CI is currently rerunning on the new merge head.

msureshkumar88 (dismissed):
Thanks for the quick turnaround — both blocking items are addressed:

1. The PR description now accurately reflects the full migration scope (Go removal + compose/Makefile/docs rewiring) and calls out the breaking changes (`/sse`, `/http`, `/api/v1/docs` removed).
2. `delay_seconds` is now clamped before constructing the `Duration` (`value.min(MAX_DELAY.as_secs_f64())`), which avoids the panic on huge finite values, and there's a new test covering `f64::MAX`. The leftover `Dockerfile` duplicate was also removed.

CI is green across the board. The remaining non-blocking suggestions (route-level integration tests, a docs page under `docs/docs/using/servers/rust/`, README note on CORS/alias route, splitting the `deny.toml` PyO3 suppression into its own PR) are still worth doing but I don't think they need to hold this up — happy to see those as follow-ups if you'd rather not bundle them here.

LGTM, approving.

ja8zyjits (approved):
LGTM

PR #5119, opened 2026-06-08
fix(fedramp): re-assert python3 symlink after subscription-manager clobbers it

msureshkumar88:
## Summary
- `subscription-manager` (installed in the FIPS/FedRAMP compliance block for STIG control RHEL-09-215010) pulls in UBI9's system `python3` (3.9) as a transitive dependency, silently overwriting `/usr/bin/python3` and clobbering the `python3 -> python${PYTHON_VERSION}` binding configured earlier in the build
- The runtime interpreter then can't see the venv's `site-packages` (built against 3.11/3.12), surfacing at pod startup as a misleading `ModuleNotFoundError: No module named 'gunicorn'` — independent of, and with the same symptom as, the GID-0 access issue fixed in #5112
- Re-create the `python3` symlink immediately after the `subscription-manager` install in both `Containerfile` and `Containerfile.lite` (nothing later in the FIPS block reinstalls a conflicting `python3` package)
- Add an end-to-end regression check to `scripts/fedramp-validate.sh` that imports `gunicorn` through the runtime `python3`, so this — or any future variant producing the same symptom — fails validation immediately instead of surfacing in a customer pod

## Test plan
- [x] `make container-build-fips CONTAINER_FILE=Containerfile.lite` — build succeeds
- [x] Confirmed `readlink -f /usr/bin/python3` resolves to `python3.11` and matches `/app/.venv/bin/python3 --version` inside the built image
- [x] `make container-validate-fedramp` — all checks pass including the new `runtime python3 can import gunicorn from the venv` check
- [x] Live `docker run` smoke test — gunicorn workers boot and the app starts without `ModuleNotFoundError`
- [x] Repeated all of the above for the non-lite `Containerfile`

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -130,6 +130,12 @@ RUN chown -R 1001:0 /app && \
 #           subscription-manager (RHEL-09-215010), pam_wheel (RHEL-09-432035),
 #           init file perms 0740 (RHEL-09-232045), home dir perms 0750 (RHEL-09-232050),
 #           rootfiles tmpfile.d (RHEL-09-232045), SSH RekeyLimit
+#
+# NOTE: subscription-manager pulls in UBI9's system python3 (3.9) as a
+# dependency, which overwrites /usr/bin/python3 and clobbers the
+# python3 -> python${PYTHON_VERSION} symlink set up earlier — surfacing at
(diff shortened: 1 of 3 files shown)


ja8zyjits (approved):
LGTM

PR #5122, opened 2026-06-08
Update virtual server source panels

vishu-bh:
Summary:
- Replace the virtual server details drawer with a Figma-aligned details panel and server-scoped component/source tabs.
- Add lazy-loaded connected MCP source selection with checkboxes and associated-source filtering in the create-server flow.
- Fix virtual server detail component fetching to use /servers/{id}/tools, /resources, and /prompts instead of global component endpoints.

Tests:
- npm run test:run -- src/components/gateways/SourceSelection.test.tsx src/pages/CreateServer.test.tsx
- npm run test:run -- src/pages/Gateways.test.tsx
- npm run build


marekdano (commented):
The formatter and tests should be fixed + a few comments

 #### Blocking

  - **`selectedMCPServerIds` is tracked but never submitted.** `SourceSelection.tsx` builds a `Set<string>` of checked server IDs but there is no `onSelect` callback or form submission path. Selections are silently discarded. If this is intentional scaffolding, it should be noted; if not, it needs wiring up before merge.

  - **Fallback API paths are unreachable dead code.** In `VirtualServerDetailsPanel.tsx` (lines 864–880), `toolsPath`/`resourcesPath`/`promptsPath` fall back to `/tools`, `/resources`, `/prompts` when `!server?.id`, but those queries are gated by `enabled: fetchEnabled` which requires `Boolean(server?.id)`. The fallback paths are never used and mislead readers.

  #### Non-blocking

  - **`actionCards[0]?.onAction` in the expanded panel** (`SourceSelection.tsx` ~line 408): The "Add source" button always fires the first action card's handler regardless of whether that card is disabled.
  - **`key={hydratedServer.id}` on the panel** (`Gateways.tsx` line 1769): Good for resetting state, but also silently discards cached component fetches every time the user switches servers. Worth confirming this is the desired behavior.

  ---

  ### Accessibility

  - **`aria-hidden` + focus management:** The `<aside>` correctly sets `aria-hidden={!open}` and the `useEffect` restores `previousFocusRef` on close. However there is no `inert` fallback for browsers that don't block keyboard focus via `aria-hidden` alone — a tab-cycling user could still reach the close button while the panel is closed.
  - **Two nested `role="tablist"` elements** (source tabs ~line 1103, component filter tabs ~line 1144) sit inside a single content region. Screen readers may find the layered tab semantics confusing — consider more explicit labeling or structural separation.
  - **Duplicate `aria-label="Search components"`** on both the icon button and the input (lines 1175, 1181) will produce duplicate announcements for screen reader users.
  - **Missing `aria-live` on empty search/filter state** (~line 1293): When a search or filter returns zero results, screen reader users get no announcement. Adding `aria-live="polite"` to the results container is a one-liner fix.

  ---

  ### Tests

  **Gaps:**
  - No tests for the new `onEdit`, `onDelete`, and `onToggleStatus` handlers in `VirtualServerCard.tsx`.
  - No tests for when the tools/resources/prompts queries fail — those errors are currently silently swallowed (the `error` prop on the panel covers only the server-details query).

client/src/components/gateways/SourceSelection.tsx
```diff
@@ -30,6 +132,44 @@ export function SourceSelection({
...
+
+  const toggleMCPServerSelection = (serverId: string, checked: boolean) => {
+    setSelectedMCPServerIds((current) => {
+      const next = new Set(current);
+      if (checked) next.add(serverId);
+      else next.delete(serverId);
+      onSelectSources?.(Array.from(next));
+      return next;
+    });
+  };
```
marekdano, line 170:
```suggestion
 const toggleMCPServerSelection = (serverId: string, checked: boolean) => {
    const next = new Set(selectedMCPServerIds);
    if (checked) next.add(serverId);
    else next.delete(serverId);
    setSelectedMCPServerIds(next);
    onSelectSources?.(Array.from(next));
  };
```
React Strict Mode double-invokes updaters, so onSelectSources fires twice per click.
Drop the functional updater form and read selectedMCPServerIds directly — the handler already has the current value via closure and there's no stale-read risk.

client/src/components/gateways/VirtualServerCard.tsx
```diff
@@ -72,21 +81,48 @@ export function VirtualServerCard({
...
-                </DropdownMenuItem>
-                <DropdownMenuItem disabled>
+                <DropdownMenuItem
+                  onClick={(e) => {
+                    e.stopPropagation();
+                    onEdit?.(server);
+                  }}
+                >
                   {intl.formatMessage({ id: "gateways.card.editServer" })}
                 </DropdownMenuItem>
```
marekdano:
```suggestion
 {onEdit && (
    <DropdownMenuItem onClick={(e) => { e.stopPropagation(); onEdit(server); }}>
      {intl.formatMessage({ id: "gateways.card.editServer" })}
    </DropdownMenuItem>
  )}
```

VirtualServerCard.tsx menu items — the bc89749b commit removed the `disabled={!onEdit}` etc. guards without wiring the callbacks in `Gateways.tsx`. The simplest fix is to render the item conditionally, so it only appears when the caller provides a handler:

client/src/components/gateways/VirtualServerCard.tsx
```diff
@@ -72,21 +81,48 @@ export function VirtualServerCard({
...
-                <DropdownMenuItem disabled className="text-destructive">
+                <DropdownMenuItem
+                  onClick={(e) => {
+                    e.stopPropagation();
+                    onToggleStatus?.(server);
+                  }}
+                >
+                  {server.enabled ? "Deactivate" : "Activate"}
+                </DropdownMenuItem>
```
marekdano:
```suggestion
 {onToggleStatus && (
    <DropdownMenuItem onClick={(e) => { e.stopPropagation(); onToggleStatus(server); }}>
      {server.enabled ? "Deactivate" : "Activate"}
    </DropdownMenuItem>
  )}
```

client/src/components/gateways/VirtualServerCard.tsx
```diff
@@ -72,21 +81,48 @@ export function VirtualServerCard({
...
+                <DropdownMenuSeparator />
+                <DropdownMenuItem
+                  onClick={(e) => {
+                    e.stopPropagation();
+                    onDelete?.(server);
+                  }}
+                  className="text-destructive focus:text-destructive"
+                >
                   {intl.formatMessage({ id: "common.button.delete" })}
                 </DropdownMenuItem>
```
marekdano:
```suggestion
 {onDelete && (
    <DropdownMenuItem onClick={(e) => { e.stopPropagation(); onDelete(server); }} className="text-destructive focus:text-destructive">
       {intl.formatMessage({ id: "common.button.delete" })}
    </DropdownMenuItem>
  )}
```

marekdano (commented):
The PR looks good. I left a few improvements

client/src/components/gateways/VirtualServerCard.tsx
vishu-bh: adressed

client/src/components/gateways/VirtualServerCard.tsx
vishu-bh: addressed

marekdano (changes requested):
Just one more finding

**Close animation will never play** — `VirtualServerDetailsPanel` is conditionally mounted based on `detailsServer` being non-null (`Gateways.tsx:126`). When the user closes the panel, `setDetailsServer(null)` is called immediately, removing the component from the DOM before the CSS `translate-x-full` transition can run. 
The panel was clearly designed to animate in/out (it has `data-[state=open/closed]` transitions), but as wired today the close animation is dead code. 
To fix it, the panel should always be mounted and controlled purely by the `open` prop:

  ```tsx
  // Gateways.tsx — always mount, use open prop
  <VirtualServerDetailsPanelContainer
    server={detailsServer}   // could be null
    open={Boolean(detailsServer)}
    onClose={() => setDetailsServer(null)}
    onAddSources={...}
  />
  ```

vishu-bh:
> Just one more finding
> 
> **Close animation will never play** — `VirtualServerDetailsPanel` is conditionally mounted based on `detailsServer` being non-null (`Gateways.tsx:126`). When the user closes the panel, `setDetailsServer(null)` is called immediately, removing the component from the DOM before the CSS `translate-x-full` transition can run. The panel was clearly designed to animate in/out (it has `data-[state=open/closed]` transitions), but as wired today the close animation is dead code. To fix it, the panel should always be mounted and controlled purely by the `open` prop:
> 
> ```tsx
> // Gateways.tsx — always mount, use open prop
> <VirtualServerDetailsPanelContainer
>   server={detailsServer}   // could be null
>   open={Boolean(detailsServer)}
>   onClose={() => setDetailsServer(null)}
>   onAddSources={...}
> />
> ```

Good catch!
addressed

marekdano (approved):
LGTM 🚀

ja8zyjits (approved):
UI team approved.
LGTM

PR #5128, opened 2026-06-08
Fix react UI auth bugs

marekdano:
## Summary

  Fixes three interrelated authentication bugs in the React SPA auth flow, plus three pre-existing RBAC middleware defects surfaced by the test suite.

  ### Bug 1 — CSRF token always rejected (403 on all POST/PUT/DELETE)

  `create_access_token` sets `sub` to the user's UUID, but CSRF token generation at all three login endpoints (`/app/auth/login`, `/auth/login`, `/auth/email/login`) was using
  `user.email` as the `user_id` bound into the HMAC. The CSRF middleware extracts `payload.get("sub")` — the UUID — so the tokens never matched.

  Fixed by decoding the freshly-issued JWT after creation and using `payload.get("sub", user.email)` as `user_id` in all three login endpoints (`routers/app.py`, `routers/auth.py`, `routers/email_auth.py`).

  ### Bug 2 — Logout never cleared cookies; page reload always redirected to /app/login

  Two root causes:

  - `/app/auth/logout` was missing from `csrf_exempt_paths`, so the CSRF middleware blocked the POST before the handler ran, meaning `clear_auth_cookie` / `clear_csrf_cookie` never executed.
  - `get_current_user_from_cookie` passed the UUID `sub` directly to `_get_user_by_email_sync` (which queries `WHERE email = ?`), always returning `None` → 401, before the handler body could clear cookies.

  Fixed in `config.py` by adding `/app/auth/logout` to `csrf_exempt_paths` (and updating `.env` / `.env.example` which were overriding the default). Fixed in `routers/app.py` by restructuring logout to always clear cookies first, then attempt best-effort JWT revocation. Fixed in `auth.py` by adding UUID→email resolution: if `sub` parses as a UUID, look up the email via `_get_email_by_id_sync` before the user lookup.

  ### Bug 3 — Three RBAC middleware defects

  - `is_same_origin_referer` (checks host + `/admin` or `/oauth/callback`) was computed but never included in `is_browser_request`, so OAuth callback pages were incorrectly rejected with 401 "Cookie authentication not allowed".
  - `sec_fetch_site: "same-site"` was in the allowed set for first-party React fetches. `same-site` allows cross-subdomain requests and should not grant cookie-based API access; only `same-origin` is correct.
  - The inline `from urllib.parse import urlparse` inside the `try` block created a function-local binding that survived the `except` clause, causing subsequent `urlparse` calls (for referer path and origin validation) to also raise when the module was patched in tests. Removed the redundant local import and wrapped the referer path and origin blocks in their own `try/except`.
  - `is_admin_ui_request` used plain string containment on the full referer URL, which matched cross-origin admin referers (e.g. `https://attacker.example/admin`). Replaced with `is_same_origin_referer`, which enforces host matching.

  ## Files changed

  | File | Change |
  |------|--------|
  | `mcpgateway/auth.py` | UUID→email resolution in `get_current_user_from_cookie` |
  | `mcpgateway/config.py` | Add `/app/auth/logout` to `csrf_exempt_paths` default |
  | `mcpgateway/routers/app.py` | CSRF token uses JWT `sub`; logout always clears cookies first |
  | `mcpgateway/routers/auth.py` | CSRF token uses JWT `sub` at `/auth/login` |
  | `mcpgateway/routers/email_auth.py` | CSRF token uses JWT `sub` at `/auth/email/login` |
  | `mcpgateway/middleware/rbac.py` | Include `is_same_origin_referer` in browser check; restrict to `same-origin`; fix local `urlparse` shadowing; gate `is_admin_ui_request` on same-origin |
  | `.env` / `.env.example` | Add `/app/auth/logout` to `CSRF_EXEMPT_PATHS` |
  | `tests/unit/mcpgateway/test_auth.py` | New `TestGetCurrentUserFromCookie` tests |
  | `tests/unit/mcpgateway/routers/test_app.py` | Update 4 logout tests to reflect CSRF-exempt behavior |
  | `tests/e2e/test_app_auth.py` | Update 2 CSRF protection tests |

  ## Tests

  - `tests/unit/mcpgateway/test_auth.py`: new `TestGetCurrentUserFromCookie` class (5 tests) covering UUID resolution, legacy email sub, missing cookie, revoked token, and unknown UUID.
  - `tests/unit/mcpgateway/routers/test_app.py`: updated 4 tests that expected logout to fail with 403/401 — logout is intentionally CSRF-exempt so users can always log out even with expired or malformed tokens.
  - `tests/e2e/test_app_auth.py`: updated 2 tests in `TestCSRFProtection` to reflect that logout succeeds (200 + cookies cleared) regardless of CSRF token presence.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -64,7 +64,7 @@ CSRF_ROTATE_ON_LOGIN=true
 CSRF_TRUSTED_ORIGINS=["http://localhost:4444","http://localhost:8000","http://localhost:8080"]
 
 # Paths exempt from CSRF protection (JSON array)
-CSRF_EXEMPT_PATHS=["/health","/auth/login","/auth/logout","/auth/refresh","/auth/email/login","/auth/email/register","/auth/email/forgot-password","/auth/email/reset-password","/admin","/oauth/fetch-tools","/docs","/redoc","/openapi.json","/metrics","/mcp/","/sse","/message","/app/auth/login","/admin/login","/admin/forgot-password","/admin/reset-password"]
+CSRF_EXEMPT_PATHS=["/health","/auth/login","/auth/logout","/auth/refresh","/auth/email/login","/auth/email/register","/auth/email/forgot-password","/auth/email/reset-password","/admin","/oauth/fetch-tools","/docs","/redoc","/openapi.json","/metrics","/mcp/","/sse","/message","/app/auth/login","/app/auth/logout","/admin/login","/admin/forgot-password","/admin/reset-password"]
 
 # Bootstrap admin credentials (email auth)
 # PRODUCTION: Change these values
--- a/mcpgateway/auth.py
+++ b/mcpgateway/auth.py
@@ -2162,7 +2162,7 @@ async def get_user_email_from_token(payload: dict, db: Session) -> Optional[str]
     # Fall back to email (legacy format)
     return sub
 
-    
+
 async def get_current_user_from_cookie(
     request: Request,
 ) -> tuple[EmailUser, Optional[str]]:
--- a/mcpgateway/auth.py
+++ b/mcpgateway/auth.py
@@ -2185,9 +2185,19 @@ async def get_current_user_from_cookie(
         logger.debug("JWT cookie validation failed: %s", e)
         raise_auth_error("invalid_token", "Invalid or expired token")
 
-    email = payload.get("sub")
-    if not email:
+    sub = payload.get("sub")
+    if not sub:
         raise_auth_error("invalid_token", "Invalid token: missing subject")
+
+    # If sub is a UUID (new token format), resolve to email via DB lookup
+    email = sub
+    try:
+        uuid.UUID(sub)
+        resolved = await asyncio.to_thread(_get_email_by_id_sync, sub)
+        if resolved is not None:
+            email = resolved
+    except ValueError:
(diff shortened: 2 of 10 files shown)


gcgoncalves (approved):

a-effort:
LLM review says ship it ⛵️

--> nonblocking, in case it's useful:                                                                                      
                                                                                                                                                                           
1. user_email fallback passes UUID instead of email on logout (routers/app.py:~189)  

```python
user_email = payload.get("email") or payload.get("sub", "unknown")
```

If the JWT has no `email` claim, this passes the `sub` UUID to `blocklist_service.revoke_token(..., user_email, "logout")`. The blocklist's `revoked_by` record and the warning log then carry a UUID where they should carry an email. 

Fix: resolve `sub` to email the same way `get_current_user_from_cookie` does (UUID detection + `_get_email_by_id_sync`) before calling `revoke_token`.

2. Lost `user_id` correlation on revocation failure (`routers/app.py:~196`)

The new path dropped `extra={"user_id": user.id}` from the warning log and `"user_id": str(user.id)` from the `auth.token_revocation_failure` metric attributes. That metric exists specifically to correlate blocklist outages to affected users. Without `user_id`, an on-call responder cannot tell which sessions were impacted.

Fix: capture the resolved email and ID from (1) and re-attach both to the log `extra` and the metric attributes.

Issue #2871, opened 2026-02-12
[BUG][API]: REST tool registration fails with "Transaction is inactive" error


rakdutta:
### 🐞 Bug Summary

While adding a REST tool from the Admin UI, the operation failed with the error **"Failed to register tool: transaction is inactive"**. After restarting the application, the issue could not be reproduced and the tool was created successfully.

---

### 📍 Steps to Reproduce

1. Navigate to **Admin UI → Tools**
2. Attempt to add/register a REST tool
3. Observe failure during tool creation

*(Issue occurred once and could not be reproduced after application restart.)*

---

### ❌ Observed Behavior

* Admin UI displays error:
  **"Failed to register tool: transaction is inactive"**
* API returns **HTTP 500** during tool creation.

---

### ✅ Expected Behavior

Tool should be successfully registered without requiring application restart.

---

### 🖥️ Error Message (Admin UI)

<img width="404" height="88" alt="Image" src="https://github.com/user-attachments/assets/98d9b102-004e-4814-8777-66cfc3693c99" />

### 📓 Logs / Error Output

Request completed: POST /admin/tools - 500", "component": "http_gateway", "category": null, "user_id": null, "user_email": null, "team_id": null, "duration_ms": 129.2593479156494, "custom_fields": null, "tags": null, "correlation_id": "b6645fed1df142c6948b4761d0566454", "operation_type": "http_request", "request_method": "POST", "request_path": "/admin/tools", "response_status_code": 500, "user_agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/144.0.0.0 Safari/537.36", "client_ip": "127.0.0.1", "metadata": {"event": "request_completed", "response_time_category": "normal"}, "hostname": "YYYY", "process_id": 120246, "timestamp": "2026-02-12T07:46:33.652988+00:00", "@timestamp": "2026-02-12T07:46:33.653059Z", "request_id": "b6645fed1df142c6948b4761d0566454"}
/home/user/pr/mcp-context-forge/mcpgateway/services/tool_service.py:1251: SAWarning: transaction already deassociated from connection
  db.rollback()
[...]

PR #5138, opened 2026-06-09
fix(tokens): set EmailApiToken.is_active=False on auto-revocation

### LINKED ISSUE #5136: [BUG]: Token displays both "Active" and "Revoked" badges simultaneously in Admin UI
### 🐞 Bug Summary
When a token is auto-revoked (e.g., via idle timeout, logout, or security reasons), the Admin UI displays both "Active" and "Revoked" badges simultaneously. This occurs because `TokenBlocklistService.revoke_token()` creates a `TokenRevocation` entry but does not update the `EmailApiToken.is_active` flag to `False`.

---

### 🧩 Affected Component

- [x] `mcpgateway` - API (token revocation service)
- [ ] Other (explain below)

---

### 🔁 Steps to Reproduce

1. Create an API token via Admin UI or API
2. Trigger auto-revocation via one of these methods:
   - Wait for idle timeout (60 minutes default)
   - Logout (if token is session-based)
   - Security-based revocation
3. Navigate to Admin UI → Tokens page
4. Observe the token displays both green "Active" badge AND red "Revoked" badge

**Note**: Manual revocation via "Revoke" button does NOT exhibit this bug (it correctly updates `is_active = False`).

---

### 🤔 Expected Behavior
Token should display only ONE badge:
- "Revoked" badge when `is_revoked = True`
- "Active" badge when `is_active = True` AND `is_revoked = False`
- "Inactive" badge when `is_active = False` AND `is_revoked = False`

Badges should be mutually exclusive.

---

### 📓 Logs / Error Output
No errors logged. This is a data consistency issue.

**Database state after auto-revocation:**
```
EmailApiToken table:
  - is_active = True  ← Not updated

TokenRevocation table:
  - jti = <token_jti>
  - reason = "idle_timeout"
  - revoked_at = <timestamp>
```

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Version or commit | `main` (post PR #5000) |
| Runtime | `Python 3.11+` |
| Platform / OS | `Any` |
| Container | `Any` |

---

### 🧩 Additional Context

**Root Cause Analysis:**

Two different code paths handle token revocation:

1. **Manual revocation** ([`token_catalog_service.py:913-914`](mcpgateway/services/token_catalog_service.py:913-914)) - ✅ CORRECT
   ```python
   token.is_active = False  # Updates EmailApiToken
   revocation = TokenRevocation(...)  # Creates revocation record
   ```

2. **Auto-revocation** ([`token_blocklist_service.py:83-116`](mcpgateway/services/token_blocklist_service.py:83-116)) - ❌ BUG
   ```python
   revocation = TokenRevocation(...)  # Creates revocation record
   # MISSING: Does not update EmailApiToken.is_active = False
   ```

**Affected Scenarios:**
- Idle timeout auto-revocation ([`auth.py:1854`](mcpgateway/auth.py:1854))
- Logout-based revocation
- Security-based revocation
- Any call to `TokenBlocklistService.revoke_token()`

**Proposed Fix:**
Update `TokenBlocklistService.revoke_token()` to query and update the `EmailApiToken.is_active` field:

```python
def revoke_token(self, jti: str, revoked_by: str, reason: str = "logout", ...):
    # ... existing code ...
    
    # Update EmailApiToken.is_active if token exists
    token = db.execute(
        select(EmailApiToken).where(EmailApiToken.jti == jti)
    ).scalar_one_or_none()
    
    if token:
        token.is_active = False
    
    # Create revocation record
    revocation = TokenRevocation(...)
    db.add(revocation)
    db.commit()
```

**UI Template Reference:**
[`tokens_partial.html:10-24`](mcpgateway/templates/tokens_partial.html:10-24) displays badges based on both flags independently, which is correct behavior - the bug is in the service layer not maintaining data consistency.

bogdanmariusc10:
## Related Issue

Closes #5136

## Summary

Fixes inconsistent token state where a token could have both `EmailApiToken.is_active=True` and a matching `TokenRevocation` row, producing conflicting status badges and API fields.

## Changes

- `TokenBlocklistService.revoke_token()` now:
  - is async and must be awaited;
  - marks a matching `EmailApiToken` inactive;
  - repairs an active catalog row when revocation already exists;
  - awaits auth-cache invalidation after the database commit;
  - emits a high-severity warning if cache invalidation fails;
  - sanitizes token identifiers and user-controlled reason/revoker values before logging.
- Logout now obtains `TokenBlocklistService` without passing the request-scoped database session. Revocation uses an independent session and persists even if later request work fails.
- Admin token badge rendering gives `Revoked` priority over `Active` or `Inactive`, so each row shows one status.
- Alembic migration `e5136a7c9d01` repairs historical rows that are active despite having a revocation record.
- Security documentation now awaits `revoke_token()`.
- Tests cover service-owned and supplied sessions, cache invalidation, repair/idempotent paths, logging failures, badge priority, logout callers, and migration behavior.

## Breaking Change / Release Requirement

`TokenBlocklistService.revoke_token()` changed from synchronous to asynchronous. Downstream callers must change:

```python
success = service.revoke_token(...)
```

to:

```python
success = await service.revoke_token(...)
```

An unmodified synchronous caller receives a truthy coroutine, may report successful logout, and does not revoke the token. Because this public API change can cause insufficient session expiration, this PR must ship in a minor release (minimum `1.1.0`), not a `1.0.x` patch.

All in-repository callers await the method.

## Reachable Scenarios

Normal API tokens with `token_use="api"` bypass idle-timeout enforcement, so idle timeout is not the normal reproduction path for those rows. Reachable affected paths include:

- `POST /auth/logout` with an API token;
- legacy API tokens missing `token_use`;
- other automatic/security revocation paths that call `TokenBlocklistService.revoke_token()`.

## Expected Behavior

- Revoked tokens have `is_active=False` and a matching revocation record.
- Historical inconsistent rows are repaired during upgrade.
- Admin UI renders only `Revoked` for revoked tokens.
- Authentication cache is invalidated before revocation returns.

### DIFF
--- a/mcpgateway/services/token_blocklist_service.py
+++ b/mcpgateway/services/token_blocklist_service.py
@@ -24,7 +24,6 @@ Security Compliance:
 """
 
 # Standard
-import asyncio
 from datetime import datetime, timedelta, timezone
 from typing import Dict, Optional
 
@@ -80,7 +79,7 @@ class TokenBlocklistService:
 
         return self._redis_client if self._redis_client is not False else None
 
-    def revoke_token(self, jti: str, revoked_by: str, reason: str = "logout", token_expiry: Optional[datetime] = None, last_activity: Optional[datetime] = None) -> bool:
+    async def revoke_token(self, jti: str, revoked_by: str, reason: str = "logout", token_expiry: Optional[datetime] = None, last_activity: Optional[datetime] = None) -> bool:
         """Revoke a token by adding it to the blocklist.
 
         Args:
@@ -96,7 +95,7 @@ class TokenBlocklistService:
         Examples:
             >>> service = TokenBlocklistService()
             >>> # Requires database connection - see unit tests for examples
-            >>> # service.revoke_token(jti="abc-123", revoked_by="user@example.com", reason="logout")
+            >>> # await service.revoke_token(jti="abc-123", revoked_by="user@example.com", reason="logout")
         """
         try:
             if self.db is not None:
@@ -118,7 +117,13 @@ class TokenBlocklistService:
                         db.commit()
                     logger.debug("Token %s already revoked", jti)
                     # Invalidate auth cache to ensure revoked token is rejected immediately
-                    self._invalidate_auth_cache(jti)
+                    try:
+                        # First-Party
+                        from mcpgateway.cache.auth_cache import auth_cache  # pylint: disable=import-outside-toplevel
+
+                        await auth_cache.invalidate_revocation(jti)
+                    except Exception as cache_error:
+                        logger.debug("Failed to invalidate auth cache for revoked token %s: %s", jti, cache_error)
                     return True
 
                 # Create revocation record
@@ -145,7 +150,13 @@ class TokenBlocklistService:
                             db.commit()
                         logger.debug("Token %s already revoked", jti)
                         # Invalidate auth cache to ensure revoked token is rejected immediately
-                        self._invalidate_auth_cache(jti)
+                        try:
+                            # First-Party
+                            from mcpgateway.cache.auth_cache import auth_cache  # pylint: disable=import-outside-toplevel
+
+                            await auth_cache.invalidate_revocation(jti)
+                        except Exception as cache_error:
+                            logger.debug("Failed to invalidate auth cache for revoked token %s: %s", jti, cache_error)
                         return True
 
                     # Create revocation record
@@ -169,7 +180,13 @@ class TokenBlocklistService:
                     logger.warning("Failed to cache revocation in Redis: %s", e)
 
             # Invalidate auth cache to ensure revoked token is rejected immediately
-            self._invalidate_auth_cache(jti)
+            try:
+                # First-Party
+                from mcpgateway.cache.auth_cache import auth_cache  # pylint: disable=import-outside-toplevel
+
+                await auth_cache.invalidate_revocation(jti)
+            except Exception as cache_error:
+                logger.debug("Failed to invalidate auth cache for revoked token %s: %s", jti, cache_error)
 
             logger.info(
                 "Token revoked: jti=%s, reason=%s, revoked_by=%s",
@@ -184,33 +201,6 @@ class TokenBlocklistService:
         except Exception as e:
             logger.error("Failed to revoke token %s: %s", jti, e)
             return False
-
-    def _invalidate_auth_cache(self, jti: str) -> None:
-        """Invalidate auth cache for a revoked token.
-
-        This is a synchronous wrapper that runs the async invalidation in the current
-        or new event loop, ensuring revoked tokens are rejected immediately.
-
-        Args:
-            jti: JWT ID to invalidate
-        """
-        try:
-            # First-Party
-            from mcpgateway.cache.auth_cache import auth_cache  # pylint: disable=import-outside-toplevel
-
-            # Try to get the current event loop
-            try:
-                loop = asyncio.get_running_loop()
-                # We're in an async context, create a task (fire-and-forget is acceptable here
-                # since the Redis/local cache update is best-effort, and the DB write is authoritative)
-                task = loop.create_task(auth_cache.invalidate_revocation(jti))
-                # Suppress the unawaited coroutine warning by adding a dummy done callback
-                task.add_done_callback(lambda t: None)
-            except RuntimeError:
-                # No running loop, create a new one for this operation
-                asyncio.run(auth_cache.invalidate_revocation(jti))
-        except Exception as cache_error:
-            logger.debug("Failed to invalidate auth cache for revoked token %s: %s", jti, cache_error)
 
     def is_token_revoked(self, jti: str) -> bool:
         """Check if a token is revoked.
--- a/mcpgateway/services/token_blocklist_service.py
+++ b/mcpgateway/services/token_blocklist_service.py
@@ -33,7 +33,7 @@ from sqlalchemy.orm import Session
 
 # First-Party
 from mcpgateway.config import settings
-from mcpgateway.db import fresh_db_session, TokenRevocation, utc_now
+from mcpgateway.db import EmailApiToken, fresh_db_session, TokenRevocation, utc_now
 from mcpgateway.services.logging_service import LoggingService
 
 # Initialize logging
@@ -108,6 +108,13 @@ class TokenBlocklistService:
                     logger.debug("Token %s already revoked", jti)
                     return True
 
+                # Update EmailApiToken.is_active if token exists
+                # Note: Session tokens don't have EmailApiToken records, so this is optional
+                api_token = db.execute(select(EmailApiToken).where(EmailApiToken.jti == jti)).scalar_one_or_none()
+                if api_token:
+                    api_token.is_active = False
+                    logger.debug(f"Set EmailApiToken.is_active=False for jti={jti}")
+
                 # Create revocation record
                 revocation = TokenRevocation(jti=jti, revoked_by=revoked_by, reason=reason, token_expiry=token_expiry, last_activity=last_activity or utc_now())
 
@@ -122,6 +129,13 @@ class TokenBlocklistService:
                     if existing:
                         logger.debug("Token %s already revoked", jti)
                         return True
+
+                    # Update EmailApiToken.is_active if token exists
+                    # Note: Session tokens don't have EmailApiToken records, so this is optional
+                    api_token = db.execute(select(EmailApiToken).where(EmailApiToken.jti == jti)).scalar_one_or_none()
+                    if api_token:
+                        api_token.is_active = False
+                        logger.debug(f"Set EmailApiToken.is_active=False for jti={jti}")
 
                     # Create revocation record
                     revocation = TokenRevocation(jti=jti, revoked_by=revoked_by, reason=reason, token_expiry=token_expiry, last_activity=last_activity or utc_now())
--- a/mcpgateway/services/token_blocklist_service.py
+++ b/mcpgateway/services/token_blocklist_service.py
@@ -24,6 +24,7 @@ Security Compliance:
 """
 
 # Standard
+import asyncio
 from datetime import datetime, timedelta, timezone
 from typing import Dict, Optional
 
@@ -104,16 +105,21 @@ class TokenBlocklistService:
                 # Check if already revoked
                 existing = db.execute(select(TokenRevocation).where(TokenRevocation.jti == jti)).scalar_one_or_none()
 
-                if existing:
-                    logger.debug("Token %s already revoked", jti)
-                    return True
-
-                # Update EmailApiToken.is_active if token exists
+                # Update EmailApiToken.is_active if token exists (repair inconsistent state if needed)
                 # Note: Session tokens don't have EmailApiToken records, so this is optional
                 api_token = db.execute(select(EmailApiToken).where(EmailApiToken.jti == jti)).scalar_one_or_none()
-                if api_token:
+                if api_token and api_token.is_active:
                     api_token.is_active = False
                     logger.debug(f"Set EmailApiToken.is_active=False for jti={jti}")
+
+                if existing:
+                    # Commit the repair if we updated is_active
+                    if api_token and not api_token.is_active:
+                        db.commit()
+                    logger.debug("Token %s already revoked", jti)
+                    # Invalidate auth cache to ensure revoked token is rejected immediately
+                    self._invalidate_auth_cache(jti)
+                    return True
 
                 # Create revocation record
                 revocation = TokenRevocation(jti=jti, revoked_by=revoked_by, reason=reason, token_expiry=token_expiry, last_activity=last_activity or utc_now())
@@ -126,16 +132,21 @@ class TokenBlocklistService:
                     # Check if already revoked
                     existing = db.execute(select(TokenRevocation).where(TokenRevocation.jti == jti)).scalar_one_or_none()
 
-                    if existing:
-                        logger.debug("Token %s already revoked", jti)
-                        return True
-
-                    # Update EmailApiToken.is_active if token exists
+                    # Update EmailApiToken.is_active if token exists (repair inconsistent state if needed)
                     # Note: Session tokens don't have EmailApiToken records, so this is optional
                     api_token = db.execute(select(EmailApiToken).where(EmailApiToken.jti == jti)).scalar_one_or_none()
-                    if api_token:
+                    if api_token and api_token.is_active:
                         api_token.is_active = False
                         logger.debug(f"Set EmailApiToken.is_active=False for jti={jti}")
+
+                    if existing:
+                        # Commit the repair if we updated is_active
+                        if api_token and not api_token.is_active:
+                            db.commit()
+                        logger.debug("Token %s already revoked", jti)
+                        # Invalidate auth cache to ensure revoked token is rejected immediately
+                        self._invalidate_auth_cache(jti)
+                        return True
 
                     # Create revocation record
                     revocation = TokenRevocation(jti=jti, revoked_by=revoked_by, reason=reason, token_expiry=token_expiry, last_activity=last_activity or utc_now())
@@ -157,6 +168,9 @@ class TokenBlocklistService:
                 except Exception as e:
                     logger.warning("Failed to cache revocation in Redis: %s", e)
 
+            # Invalidate auth cache to ensure revoked token is rejected immediately
+            self._invalidate_auth_cache(jti)
+
             logger.info(
                 "Token revoked: jti=%s, reason=%s, revoked_by=%s",
                 jti,
@@ -170,6 +184,33 @@ class TokenBlocklistService:
         except Exception as e:
             logger.error("Failed to revoke token %s: %s", jti, e)
             return False
+
+    def _invalidate_auth_cache(self, jti: str) -> None:
+        """Invalidate auth cache for a revoked token.
+
+        This is a synchronous wrapper that runs the async invalidation in the current
+        or new event loop, ensuring revoked tokens are rejected immediately.
+
+        Args:
+            jti: JWT ID to invalidate
+        """
+        try:
+            # First-Party
+            from mcpgateway.cache.auth_cache import auth_cache  # pylint: disable=import-outside-toplevel
+
+            # Try to get the current event loop
+            try:
+                loop = asyncio.get_running_loop()
+                # We're in an async context, create a task (fire-and-forget is acceptable here
+                # since the Redis/local cache update is best-effort, and the DB write is authoritative)
+                task = loop.create_task(auth_cache.invalidate_revocation(jti))
+                # Suppress the unawaited coroutine warning by adding a dummy done callback
+                task.add_done_callback(lambda t: None)
+            except RuntimeError:
+                # No running loop, create a new one for this operation
+                asyncio.run(auth_cache.invalidate_revocation(jti))
+        except Exception as cache_error:
+            logger.debug("Failed to invalidate auth cache for revoked token %s: %s", jti, cache_error)
 
     def is_token_revoked(self, jti: str) -> bool:
         """Check if a token is revoked.
--- a/mcpgateway/services/token_blocklist_service.py
+++ b/mcpgateway/services/token_blocklist_service.py
@@ -107,13 +107,15 @@ class TokenBlocklistService:
                 # Update EmailApiToken.is_active if token exists (repair inconsistent state if needed)
                 # Note: Session tokens don't have EmailApiToken records, so this is optional
                 api_token = db.execute(select(EmailApiToken).where(EmailApiToken.jti == jti)).scalar_one_or_none()
+                repaired = False
                 if api_token and api_token.is_active:
+                    repaired = True
                     api_token.is_active = False
-                    logger.debug(f"Set EmailApiToken.is_active=False for jti={jti}")
+                    logger.debug("Set EmailApiToken.is_active=False for jti=%s", jti[:8])
 
                 if existing:
                     # Commit the repair if we updated is_active
-                    if api_token and not api_token.is_active:
+                    if repaired:
                         db.commit()
                     logger.debug("Token %s already revoked", jti)
                     # Invalidate auth cache to ensure revoked token is rejected immediately
@@ -123,7 +125,12 @@ class TokenBlocklistService:
 
                         await auth_cache.invalidate_revocation(jti)
                     except Exception as cache_error:
-                        logger.debug("Failed to invalidate auth cache for revoked token %s: %s", jti, cache_error)
+                        logger.warning(
+                            "Auth cache invalidation failed for revoked token; token may be accepted until cache TTL expiry: jti=%s: %s",
+                            jti,
+                            cache_error,
+                            extra={"security_event": "revocation_cache_invalidation_failed", "security_severity": "high", "jti": jti},
+                        )
                     return True
 
                 # Create revocation record
@@ -140,13 +147,15 @@ class TokenBlocklistService:
                     # Update EmailApiToken.is_active if token exists (repair inconsistent state if needed)
                     # Note: Session tokens don't have EmailApiToken records, so this is optional
                     api_token = db.execute(select(EmailApiToken).where(EmailApiToken.jti == jti)).scalar_one_or_none()
+                    repaired = False
                     if api_token and api_token.is_active:
+                        repaired = True
                         api_token.is_active = False
-                        logger.debug(f"Set EmailApiToken.is_active=False for jti={jti}")
+                        logger.debug("Set EmailApiToken.is_active=False for jti=%s", jti[:8])
 
                     if existing:
                         # Commit the repair if we updated is_active
-                        if api_token and not api_token.is_active:
+                        if repaired:
                             db.commit()
                         logger.debug("Token %s already revoked", jti)
                         # Invalidate auth cache to ensure revoked token is rejected immediately
@@ -156,7 +165,12 @@ class TokenBlocklistService:
 
                             await auth_cache.invalidate_revocation(jti)
                         except Exception as cache_error:
-                            logger.debug("Failed to invalidate auth cache for revoked token %s: %s", jti, cache_error)
+                            logger.warning(
+                                "Auth cache invalidation failed for revoked token; token may be accepted until cache TTL expiry: jti=%s: %s",
+                                jti,
+                                cache_error,
+                                extra={"security_event": "revocation_cache_invalidation_failed", "security_severity": "high", "jti": jti},
+                            )
                         return True
 
                     # Create revocation record
@@ -186,7 +200,12 @@ class TokenBlocklistService:
 
                 await auth_cache.invalidate_revocation(jti)
             except Exception as cache_error:
-                logger.debug("Failed to invalidate auth cache for revoked token %s: %s", jti, cache_error)
+                logger.warning(
+                    "Auth cache invalidation failed for revoked token; token may be accepted until cache TTL expiry: jti=%s: %s",
+                    jti,
+                    cache_error,
+                    extra={"security_event": "revocation_cache_invalidation_failed", "security_severity": "high", "jti": jti},
+                )
 
             logger.info(
                 "Token revoked: jti=%s, reason=%s, revoked_by=%s",
--- a/mcpgateway/services/token_blocklist_service.py
+++ b/mcpgateway/services/token_blocklist_service.py
@@ -135,14 +135,12 @@ class TokenBlocklistService:
                     if repaired:
                         db.commit()
                     logger.debug("Token %s already revoked", SecurityValidator.sanitize_log_message(jti[:8]))
-                    await self._invalidate_auth_cache(jti)
-                    return True
+                else:
+                    # Create revocation record
+                    revocation = TokenRevocation(jti=jti, revoked_by=revoked_by, reason=reason, token_expiry=token_expiry, last_activity=last_activity or utc_now())
 
-                # Create revocation record
-                revocation = TokenRevocation(jti=jti, revoked_by=revoked_by, reason=reason, token_expiry=token_expiry, last_activity=last_activity or utc_now())
-
-                db.add(revocation)
-                db.commit()
+                    db.add(revocation)
+                    db.commit()
             else:
                 # Create own session
                 with fresh_db_session() as db:
@@ -163,14 +161,12 @@ class TokenBlocklistService:
                         if repaired:
                             db.commit()
                         logger.debug("Token %s already revoked", SecurityValidator.sanitize_log_message(jti[:8]))
-                        await self._invalidate_auth_cache(jti)
-                        return True
+                    else:
+                        # Create revocation record
+                        revocation = TokenRevocation(jti=jti, revoked_by=revoked_by, reason=reason, token_expiry=token_expiry, last_activity=last_activity or utc_now())
 
-                    # Create revocation record
-                    revocation = TokenRevocation(jti=jti, revoked_by=revoked_by, reason=reason, token_expiry=token_expiry, last_activity=last_activity or utc_now())
-
-                    db.add(revocation)
-                    db.commit()
+                        db.add(revocation)
+                        db.commit()
 
             # Cache in Redis for fast lookup
             redis_client = self._get_redis_client()
--- a/mcpgateway/services/token_blocklist_service.py
+++ b/mcpgateway/services/token_blocklist_service.py
@@ -32,6 +32,7 @@ from sqlalchemy import delete, select
 from sqlalchemy.orm import Session
 
 # First-Party
+from mcpgateway.common.validators import SecurityValidator
 from mcpgateway.config import settings
 from mcpgateway.db import EmailApiToken, fresh_db_session, TokenRevocation, utc_now
 from mcpgateway.services.logging_service import LoggingService
@@ -56,6 +57,22 @@ class TokenBlocklistService:
         """
         self.db = db
         self._redis_client = None
+
+    async def _invalidate_auth_cache(self, jti: str) -> None:
+        """Best-effort invalidate cached authentication state for a revoked token."""
+        safe_jti = SecurityValidator.sanitize_log_message(jti)
+        try:
+            # First-Party
+            from mcpgateway.cache.auth_cache import auth_cache  # pylint: disable=import-outside-toplevel
+
+            await auth_cache.invalidate_revocation(jti)
+        except Exception as cache_error:
+            logger.warning(
+                "Auth cache invalidation failed for revoked token; token may be accepted until cache TTL expiry: jti=%s: %s",
+                safe_jti,
+                cache_error,
+                extra={"security_event": "revocation_cache_invalidation_failed", "security_severity": "high", "jti": safe_jti},
+            )
 
     def _get_redis_client(self):
         """Get Redis client for caching (lazy initialization).
@@ -111,26 +128,14 @@ class TokenBlocklistService:
                 if api_token and api_token.is_active:
                     repaired = True
                     api_token.is_active = False
-                    logger.debug("Set EmailApiToken.is_active=False for jti=%s", jti[:8])
+                    logger.debug("Set EmailApiToken.is_active=False for jti=%s", SecurityValidator.sanitize_log_message(jti[:8]))
 
                 if existing:
                     # Commit the repair if we updated is_active
                     if repaired:
                         db.commit()
-                    logger.debug("Token %s already revoked", jti)
-                    # Invalidate auth cache to ensure revoked token is rejected immediately
-                    try:
-                        # First-Party
-                        from mcpgateway.cache.auth_cache import auth_cache  # pylint: disable=import-outside-toplevel
-
-                        await auth_cache.invalidate_revocation(jti)
-                    except Exception as cache_error:
-                        logger.warning(
-                            "Auth cache invalidation failed for revoked token; token may be accepted until cache TTL expiry: jti=%s: %s",
-                            jti,
-                            cache_error,
-                            extra={"security_event": "revocation_cache_invalidation_failed", "security_severity": "high", "jti": jti},
-                        )
+                    logger.debug("Token %s already revoked", SecurityValidator.sanitize_log_message(jti[:8]))
+                    await self._invalidate_auth_cache(jti)
                     return True
 
                 # Create revocation record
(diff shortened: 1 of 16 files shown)


vishu-bh (changes requested):
Thanks for the fix here. The service change is a good step toward keeping `TokenRevocation` and `EmailApiToken.is_active` consistent, and the new unit coverage for both session-backed and fresh-session paths is helpful.

A couple of things look worth addressing before merge:

- `TokenBlocklistService.revoke_token()` updates the DB state, but it does not invalidate `auth_cache` after revocation. Manual token revocation already calls `auth_cache.invalidate_revocation(token.jti)`, but the auto-revocation/logout path still appears to leave cached auth contexts with `is_token_revoked=False` until cache expiry. That can create a short window where a revoked token is still accepted.

- Change prevents `Active + Revoked`, but the template still renders the active/inactive badge separately from the revoked badge. With this change, revoked tokens can still show two badges as `Inactive + Revoked`, while #5136 expects a single `Revoked` badge. The template likely needs `is_revoked` to take priority over the active/inactive badge.

It may also be worth making the idempotent path repair the catalog row: if a `TokenRevocation` row already exists, `revoke_token()` returns before setting `EmailApiToken.is_active = False`, so existing inconsistent rows remain inconsistent.

additionally I think we need coverage for cache invalidation and revoked badge priority as well.

bogdanmariusc10:
Thanks for the thorough review, @vishu-bh! I've addressed all the findings:

## 1. Auth cache invalidation missing
**Fixed**: `TokenBlocklistService.revoke_token()` now calls `auth_cache.invalidate_revocation(jti)` after DB update (both session paths). Added `_invalidate_auth_cache()` helper that handles async/sync contexts properly.

**Test coverage**: `test_revoke_token_invalidates_auth_cache` verifies cache invalidation is called.

## 2. Idempotent path skips catalog repair
**Fixed**: Moved `EmailApiToken.is_active` update before the early return check. When a `TokenRevocation` already exists, the method now still sets `is_active = False` if needed and commits the repair.

**Test coverage**: `test_revoke_token_repairs_inconsistent_state` creates an inconsistent state (TokenRevocation exists but is_active=True) and verifies the idempotent call repairs it.

## 3. Template badge priority
**Fixed**: Changed `tokens_partial.html` to use `{% if is_revoked %}...{% elif is_active %}...{% else %}` pattern, ensuring revoked badge takes priority over active/inactive badges.

**Test coverage**: `test_tokens_revoked_badge_takes_priority_over_active_badge` renders tokens with both revoked+active and revoked+inactive states, verifying only one "Revoked" badge appears.

---

All 45 existing tests pass + 3 new tests added.

mcpgateway/services/token_blocklist_service.py
```diff
@@ -158,6 +186,33 @@ def revoke_token(self, jti: str, revoked_by: str, reason: str = "logout", token_
...
+        try:
+            # First-Party
+            from mcpgateway.cache.auth_cache import auth_cache  # pylint: disable=import-outside-toplevel
+
+            # Try to get the current event loop
+            try:
+                loop = asyncio.get_running_loop()
+                # We're in an async context, create a task (fire-and-forget is acceptable here
+                # since the Redis/local cache update is best-effort, and the DB write is authoritative)
+                task = loop.create_task(auth_cache.invalidate_revocation(jti))
```
vishu-bh: The new helper uses `loop.create_task(auth_cache.invalidate_revocation(jti))` when `revoke_token()` is called inside an async route, so `revoke_token()` can return before the auth cache has actually been invalidated. That leaves the same short stale-cache window for revoked tokens.

This is the same race called out in `TokenCatalogService.revoke_token()`, where the code intentionally awaits `auth_cache.invalidate_revocation()` instead of fire-and-forget because the next request could otherwise arrive before invalidation completes.

One possible fix would be to make the auto-revocation/logout call path await cache invalidation, or otherwise update the local revoked marker synchronously before returning. The current tests also pass with `RuntimeWarning: coroutine 'AuthCache.invalidate_revocation' was never awaited`, so I think the cache-invalidation test path needs tightening as well.

vishu-bh (changes requested):
Thanks for the update. I see the DB consistency and badge-priority parts are addressed now.

I still think the cache invalidation concern remains, though. 

Dropped an inline comment

vishu-bh:
Hi @bogdanmariusc10, when you get a chance, could you let us know whether you’re planning to address the comment?

msureshkumar88 (changes requested):
Thanks for digging into this one — the root-cause analysis in #5136 is spot on, and the fix lands in the right layer. Setting `EmailApiToken.is_active = False` in `TokenBlocklistService.revoke_token()` is exactly the missing piece, and adding the `auth_cache.invalidate_revocation()` call brings the auto-revocation path in line with what `TokenCatalogService` already does per ADR-028. The template `{% if %}/{% elif %}` change is also the right call even though the issue suggested the template was fine — it's the only thing that fixes rows already sitting in the DB from before this fix. Rewriting `test_auth_logout.py` onto a real SQLite session instead of `MagicMock` is a genuine improvement too.

I ran the affected suites locally and they're green (89 tests in the blocklist/logout/idle-timeout files, 22 in `TestTemplateButtonGating`).

A few things I think need a second look before this goes in — mostly around the sync→async change, which I don't think is being flagged loudly enough, plus one logic bug in the repair path.

---

## Blocking

### 1. The sync → async signature change is a breaking API change that fails open

`TokenBlocklistService.revoke_token()` went from sync to `async`. All three in-repo call sites are correctly awaited (`auth.py:1854`, `admin.py:4884`, `routers/auth.py:339`) — I grepped and there are no stragglers, so the repo itself is fine.

The concern is what happens to a caller that *isn't* updated:

```python
success = blocklist_service.revoke_token(jti=...)   # returns a coroutine — truthy
if not success:                                      # never fires
    raise HTTPException(500, "Failed to revoke token")
# token is NOT revoked, caller reports success, only a RuntimeWarning at GC time
```

A logout path that silently doesn't revoke is CWE-613 (Insufficient Session Expiration), and there's no exception, log, or metric to notice it. `mcpgateway` ships on PyPI, so downstream plugins and operator scripts calling this documented service get no signal at all.

To be clear — I think making it async is the *right* design; the `create_task`/`asyncio.run` helper from `650750c` was worse. The ask is just the paperwork:

- [ ] `docs/security/JWT_TOKEN_SECURITY_IMPLEMENTATION.md:124` still shows the sync call. Anyone copying that snippet now writes a fail-open bug.
- [ ] `CHANGELOG.md` needs a **Breaking Changes** entry naming the signature change and the truthy-coroutine hazard.
- [ ] Worth confirming the release is a minor/major bump, not a patch.

If you'd rather not break downstream at all, a thin shim would also work:

```python
def revoke_token_sync(self, *args, **kwargs) -> bool:
    """Deprecated sync wrapper. Do not call from within an event loop."""
    return asyncio.run(self.revoke_token(*args, **kwargs))
```

### 2. Repair guard tests state instead of mutation

`token_blocklist_service.py:117-118` and `150-151`:

```python
if api_token and api_token.is_active:
    api_token.is_active = False          # only set when it WAS active
if existing:
    if api_token and not api_token.is_active:   # also true when it was ALREADY inactive
        db.commit()
```

The second condition is true whenever the token is inactive, not only when this call changed it — so every repeat revoke of an already-consistent token issues a pointless `COMMIT`.

That's harmless on the `fresh_db_session()` branch, but on the `self.db is not None` branch it commits the **caller's** request-scoped session mid-request, which is the anti-pattern documented in AGENTS.md for #2871. This PR removes the only in-repo caller that passes `db=`, but `get_token_blocklist_service(db=...)` is still public.

Suggested:

```python
repaired = bool(api_token and api_token.is_active)
if repaired:
    api_token.is_active = False
    logger.debug("Set EmailApiToken.is_active=False for jti=%s", jti)
if existing:
    if repaired:
        db.commit()
    ...
```

### 3. Cache-invalidation failures are swallowed at DEBUG

`token_blocklist_service.py:126-127`, `159-160`, `189-190`:

```python
except Exception as cache_error:
    logger.debug("Failed to invalidate auth cache for revoked token %s: %s", jti, cache_error)
```

`auth_cache` keeps a positive `set_not_revoked` entry with a TTL (`auth_cache.py:1284`). If `invalidate_revocation()` fails — Redis partition, lock contention — a **just-revoked token keeps authenticating until that TTL expires**, and the only trace is a DEBUG line that's off under the default `LOG_LEVEL=ERROR`.

That's a security control failing silently (CWE-613 + CWE-778). Could this be raised to `warning` with the security-event fields the rest of the file already uses?

```python
except Exception as cache_error:
    logger.warning(
        "Auth cache invalidation failed for revoked token; token may be accepted until cache TTL expiry: jti=%s: %s",
        jti, cache_error,
        extra={"security_event": "revocation_cache_invalidation_failed", "security_severity": "high", "jti": jti},
    )
```

Worth noting the ordering is correct everywhere — invalidation happens after `db.commit()` in all three sites, so there's no read-your-writes hole. That part is solid.

### 4. PR description has drifted from the code

Three mismatches that will end up in release notes:

- It describes an `_invalidate_auth_cache()` helper with `create_task`/`asyncio.run` handling. That helper no longer exists — it was removed when the method went async, and the body is now three inline `await` stanzas.
- It says `asyncio` was imported. It isn't.
- It doesn't mention the sync→async change or the `routers/auth.py` session change (`0c9d678`) at all — those are the two highest-impact edits in the diff.

One factual note for the repro steps: `auth.py:1826` skips idle-timeout enforcement entirely for `token_use == "api"`, which are precisely the tokens that have `EmailApiToken` rows. So idle timeout can't actually produce the dual badge for a normal API token. The reachable paths look narrower — legacy tokens missing the `token_use` claim, or `POST /auth/logout` called with an API token. Worth correcting so testers aren't chasing a scenario that can't fire.

---

## Suggestions (non-blocking, would happily merge without these)

**Duplication.** The cache-invalidation stanza appears three times (121-127, 154-160, 184-190), and the `self.db is not None` / `fresh_db_session()` branches are ~30 near-identical lines. After this PR no in-repo caller passes `db=`, so the first branch is effectively dead in production. Collapsing both would halve the diff and remove the caller-session-commit hazard from item 2 entirely:

```python
ctx = nullcontext(self.db) if self.db is not None else fresh_db_session()
with ctx as db:
    ...
```

Related: pylint on this file drops 10.00 → 9.15, with `revoke_token` newly tripping `R0912` (15/12 branches) and `R0915` (58/50 statements). Ruff has both disabled so CI won't complain, but the refactor above clears them.

**Missing rollback.** The outer `except Exception: return False` (line 202) never rolls back. On the `self.db` branch, `api_token.is_active = False` may already be pending on the caller's session, leaving it dirty. A `db.rollback()` there would be safer.

**Concurrent revocation → false 500.** Two parallel logouts for one jti: both read `existing is None`, both INSERT, the second hits `IntegrityError` on the PK, gets caught by the broad handler, returns `False`, and `routers/auth.py:342` raises a 500 even though the token *is* revoked (CWE-362). Pre-existing, but the extra SELECT + mutation widens the window. Catching `IntegrityError` and treating it as the idempotent-success path would fix it.

**Redis cache skipped on the repair path.** The `existing` branch returns `True` before reaching the Redis `setex` at line 170, so a repaired token that isn't yet in the Redis blocklist stays uncached and every check falls through to the DB.

**ORM load for a single boolean.** `select(EmailApiToken).scalar_one_or_none()` hydrates the full entity including the JSON columns. An `update(EmailApiToken).where(jti == jti).values(is_active=False)` with `rowcount` avoids the load, collapses SELECT+UPDATE into one statement, and incidentally closes the race above.

**Log hygiene.** `logger.debug(f"Set EmailApiToken.is_active=False for jti={jti}")` logs the full jti where `auth_cache.py:536` deliberately truncates to `jti[:8]`, and uses an f-string where the rest of the file uses lazy `%s`.

---

## Migrations

Agreed that no schema migration is needed — `is_active` and `jti` both exist, and `jti` is already `unique=True` with `idx_email_api_tokens_jti`. Good call leaving Alembic alone.

There is a data gap though. Rows revoked before this change keep `is_active=True` permanently, since `revoke_token()` will never be called on those jtis again. The template masks it in the UI, but the REST API still returns the contradiction:

```
GET /tokens  →  {"is_active": true, "is_revoked": true}
```

(`routers/tokens.py:354,678,962` and `admin.py:10863,10964` all read the column directly.)

A data-only migration would close it out for existing installs:

```sql
UPDATE email_api_tokens SET is_active = 0
 WHERE jti IN (SELECT jti FROM token_revocations);
```

Happy for that to be a follow-up issue rather than in this PR — but if it's deferred, could the PR description note that API responses stay inconsistent for historical rows?

---

## Testing

Coverage of the changed *lines* is good. Coverage of the changed *behavior at the layer the bug was reported* is where the gaps are:

- No integration test asserting `is_active=False` in the **API response** after auto-revocation — only unit-level DB state. `tests/e2e/test_admin_apis.py` looks like the right home.
- No e2e badge assertion. `tests/playwright/security/test_token_lifecycle.py` and `tests/playwright/pages/tokens_page.py` already exist, so the user-visible symptom from the issue currently has zero browser-level coverage. One assertion there would be a nice close to the loop.
- No test distinguishing "repaired" from "already inactive" — that's what would have caught item 2.
- Concurrent-revocation race and the skipped-Redis-cache path are both untested.

Two small notes on the new tests:

- `test_tokens_revoked_badge_takes_priority_over_active_badge` asserts on the raw template source (`assert "{% elif token.is_active %}" in template_content`). The three rendered-badge count assertions above it already prove the behavior, and the source assert will break on any harmless template reflow. Suggest dropping it.
- `setup_test_db` monkeypatches `mcpgateway.db.SessionLocal` and `.engine` globals. It works today because `fresh_db_session = contextmanager(get_db)` resolves `SessionLocal` at call time, but mutating module globals is xdist-hostile. The project's existing DB fixtures would be more robust.

---

## Docs

- `docs/security/JWT_TOKEN_SECURITY_IMPLEMENTATION.md:124` — stale sync example (blocking, item 1).
- `CHANGELOG.md` — no entry (blocking, item 1).
- `docs/docs/architecture/adr/028-auth-caching.md:54` — lists only `token_catalog_service.revoke_token()` as invalidating the revocation cache. `TokenBlocklistService` now does too, worth adding.
- Method docstring correctly updated to `await`. 👍

---

## Scope

Scope is clean — all 10 files are on-topic. The three `await`s follow from the signature change, the `routers/auth.py` session change is the fix for the transaction conflict in `0c9d678`, and the five test files follow from both. No `.secrets.baseline` churn, no drive-by formatting, no vendored assets. `Closes #5136` is present in the body and in 4 of 7 commits.

---

Summary of what I'd like to see before merge: items 1–4 above (breaking-change docs + CHANGELOG, the `repaired` flag fix, the log-level bump, and a description refresh). Everything under Suggestions is genuinely optional — happy to see any of it split into follow-ups. Nice work tracking down a subtle two-table consistency bug.

gandhipratik203 (dismissed):
Approving — clean data-layer fix, async `revoke_token()` callers all awaited, tests pass.

Optional follow-up (non-blocking): repair only fires on a re-revoke, so rows already inconsistent in the DB stay `is_active=True` until then. If any exist in your environment, a small data-fix migration would close it. Fine to merge as-is.

msureshkumar88 (changes requested):
Thanks for staying with this one through several rounds — the fix has landed in a good place. A few things I want to call out as genuinely well done before the asks:

- The root-cause split is right. `EmailApiToken.is_active = False` in `TokenBlocklistService.revoke_token()` is the actual missing piece, and the template `{% if %}/{% elif %}` change is the only thing that helps rows already sitting inconsistent in the DB. The issue suggested the template was fine — it wasn't, and you were right to change both.
- The `auth_cache.invalidate_revocation()` addition closes a real gap, not just a cosmetic one. Before this, auto-revocation wrote `TokenRevocation` but left cached auth contexts with `is_token_revoked=False` until TTL, so a logged-out token kept authenticating for a window. `TokenCatalogService` already did this per ADR-028; the blocklist path didn't. Nice catch, and the ADR-028 update is exactly where a reader would look for it.
- Worth noting in the PR body: the `is_active` change isn't only cosmetic either. `auth.py:720` authenticates raw API tokens with `WHERE token_hash = ? AND is_active IS TRUE`, so this also hard-fails auto-revoked API tokens at the token-hash path. That's a stronger argument for the fix than the badge alone.
- Rewriting `test_auth_logout.py` from `MagicMock` onto a real in-memory SQLite session is a real improvement — it's the only reason the `db=db` removal is actually verified rather than mocked past.

I ran the affected suites locally and they're green (90 tests across `test_token_blocklist_service.py`, `test_auth_logout.py`, `test_auth_idle_timeout.py`, `routers/test_auth.py`), with `token_blocklist_service.py` at 100% line coverage. I also confirmed all three production call sites (`auth.py:1857`, `admin.py:4966`, `routers/auth.py:339`) are awaited and there are no remaining sync callers anywhere in `mcpgateway/`, `plugins/`, or `mcp-servers/`.

---

## Blocking

**1. The PR description documents code that isn't in the diff.**

The description says a `_invalidate_auth_cache()` helper was added, "handles both async context (create_task) and sync context (asyncio.run)", with "a done_callback to suppress unawaited coroutine warnings". None of that is in the final code — it inlines a plain `await auth_cache.invalidate_revocation(jti)` three times. The test list also names nine tests under a "6 new tests" heading. Anyone reviewing from the description is reviewing a different change. Could you refresh it to match what's actually there?

**2. The invalidation block is duplicated three times verbatim.**

Lines 121-133, 161-173 and 196-208 of `token_blocklist_service.py` are byte-identical, log message included. Extracting it also makes the description in (1) true:

```python
async def _invalidate_auth_cache(self, jti: str) -> None:
    """Best-effort auth-cache invalidation for a revoked token."""
    try:
        from mcpgateway.cache.auth_cache import auth_cache  # pylint: disable=import-outside-toplevel

        await auth_cache.invalidate_revocation(jti)
    except Exception as cache_error:
        logger.warning(
            "Auth cache invalidation failed for revoked token; token may be accepted until cache TTL expiry: jti=%s: %s",
            jti,
            cache_error,
            extra={"security_event": "revocation_cache_invalidation_failed", "security_severity": "high", "jti": jti},
        )
```

Roughly -40 lines.

**3. CHANGELOG is missing the issue number, and one of the two fixes isn't recorded.**

The adjacent entry uses `(#4636)`, so both new lines should carry `(#5136)`.

Separately, commit `075ea5ba` ("logout endpoint database transaction conflict") changes `routers/auth.py` from `get_token_blocklist_service(db=db)` to `get_token_blocklist_service()`. That's a distinct fix riding along, and it changes logout's transaction semantics — revocation now commits in an independent session and survives a later failure in the same request. It's causally required by the rest of the change, so no objection to bundling it, but it deserves its own line in both the PR summary and the CHANGELOG.

While you're in the CHANGELOG: the breaking-change entry would be more useful to operators if it named the symptom they'd actually observe. An un-awaited call returns a truthy coroutine, so `if not success:` never fires and the caller reports success while nothing was revoked — i.e. "logout returns 200 but the token still authenticates."

---

## Suggestions

**4. One integration test covering the reported scenario end to end.**

Right now every test asserts one half of the invariant in isolation. `test_tokens_revoked_badge_takes_priority_over_active_badge` feeds hand-built dicts into Jinja (which correctly verifies the `elif`), and the service tests check `is_active` without touching the response shape. Nothing asserts the two halves agree.

`tests/integration/test_token_security_integration.py` looks like the natural home for: create a real API token, auto-revoke it, then assert the API response carries `is_active=false` **and** `is_revoked=true`. That's the actual bug in #5136, and today it would still pass if either half regressed.

Two smaller notes in the same area, both optional: the `admin.py` and `auth.py` call sites are only mock-verified, and `AsyncMock` returns a value on `await` *and* a coroutine without one — so with no assertion on the return, those tests pass either way, which is precisely the failure mode of the async conversion. And the 100% figure is line coverage; `--cov-branch` would be more meaningful for `if api_token and api_token.is_active`, which has three distinct outcomes.

**5. Existing inconsistent rows are still wrong in the API.**

`repaired` only fires on a *re-revoke*, which for an idle-timed-out token never happens. The template `elif` hides the dual badge in the UI, but `routers/tokens.py:353,677,961` and `admin.py:10958,11059` still emit `is_active: true, is_revoked: true` for anything auto-revoked before this deploy. An API consumer sees the original bug unchanged.

@gandhipratik203 raised this as non-blocking; I'd lean slightly stronger, since the UI fix makes it invisible and it'll likely get reported again. No schema change needed — a one-time data migration would do it:

```python
def upgrade() -> None:
    bind = op.get_bind()
    insp = sa.inspect(bind)
    if "email_api_tokens" not in insp.get_table_names() or "token_revocations" not in insp.get_table_names():
        return
    bind.execute(sa.text(
        "UPDATE email_api_tokens SET is_active = 0 "
        "WHERE is_active = 1 AND jti IN (SELECT jti FROM token_revocations)"
    ))


def downgrade() -> None:
    pass  # not reversible; revoked tokens must stay inactive
```

Happy for this to be a follow-up issue rather than a blocker if you'd prefer to keep this PR tight — just would rather it be tracked than dropped.

**6. Log neutralization is inconsistent with the sibling service.**

`token_catalog_service.py` wraps every logged identifier in `SecurityValidator.sanitize_log_message()`. The new code logs raw `jti` and `revoked_by`, both JWT-claim-derived, straight into the structured sink. Worth matching for consistency:

```python
extra={"security_event": "...", "jti": SecurityValidator.sanitize_log_message(jti)}
```

Minor related nit: the same function logs `jti[:8]` at DEBUG (lines 114, 154) and the full `jti` at WARNING (lines 130, 170). `auth_cache.invalidate_revocation()` truncates to 8 — probably worth picking one convention.

**7. Concurrent revoke of the same JTI surfaces as a 500 on logout.**

`TokenRevocation.jti` is the primary key, and the code does `SELECT ... WHERE jti` then `INSERT` with no conflict handling. Two concurrent revokes: one hits `IntegrityError`, the broad `except Exception` at line 220 swallows it and returns `False`, and `routers/auth.py:342` turns that into a 500 on a logout that actually succeeded.

This race predates the PR, but it gets hotter here — a burst of requests carrying one idle-timed-out token now all race to revoke it. Cheap to close:

```python
except IntegrityError:
    db.rollback()
    return True   # concurrent revoke reached the same end state
```

---

## Minor notes

- **The `self.db is not None` branch is now dead in production.** After this change all three call sites use `get_token_blocklist_service()` with no `db`, so nothing outside tests reaches lines 101-140 — and that branch is a near-copy of the `fresh_db_session` one. Collapsing them would remove the duplication and also resolve a latent issue: in that branch `db.commit()` (line 119) commits the *caller's* request-scoped session, and the handler at line 220 returns `False` without a `rollback()`, leaving the caller with a dirty session. Something like `ctx = nullcontext(self.db) if self.db is not None else fresh_db_session()` over a single copy of the body would do it.

- **`if repaired: db.commit()` is redundant in the `fresh_db_session` branch.** `fresh_db_session = contextmanager(get_db)` (`db.py:6079`), and `get_db()` commits on normal generator exit (`db.py:5990`) — a `return` inside the `with` is a normal exit. Harmless, just unnecessary there. It *is* load-bearing in the caller-session branch, which is the one that arguably shouldn't be committing someone else's session.

- **The repair itself is logged at DEBUG** (lines 114, 154). Repairing pre-existing inconsistent state seems worth INFO — otherwise there's no way to tell from production logs how many legacy rows got fixed. Adding `repaired` to the `extra` dict on the existing "Token revoked" INFO would cover it too.

- **Sync DB I/O inside the new `async def`.** All the SQLAlchemy calls in `revoke_token()` run directly on the event loop, and the idle-timeout path is on the hot auth path. Not a regression (it blocked before too), but the immediate neighbours in the same call chain — `auth.py:1811` and `streamablehttp_transport.py:5181` — both wrap the equivalent read in `await asyncio.to_thread(...)`, so there's an easy consistency win here.

- **`invalidate_revocation()` is now on a much higher-frequency path.** `auth_cache.py:561` does `redis.scan_iter(match="mcpgw:auth:ctx:*:{jti}")`, which is O(keyspace) per call. Previously only admin-initiated revokes triggered it; now every logout and idle timeout does, so an idle-timeout storm after a quiet period becomes a burst of keyspace scans. Not something to solve in this PR — probably worth a follow-up issue about keeping a `jti -> ctx-key` reverse set instead of scanning.

- **Mixed sync/async surface on the class.** `revoke_token()` is now `async` while `is_token_revoked()`, `update_activity()`, `get_last_activity()` and `cleanup_expired_revocations()` stay sync. Worth a docstring note so the asymmetry doesn't surprise the next caller.

- Squashing the nine commits (including `chore: fix trailing whitespace`) before merge would tidy the history.

---

On the breaking change itself: converting `revoke_token()` to `async` is a reasonable call — it keeps the blocklist service consistent with `TokenCatalogService`, which is already async and invalidates internally. The alternative (keep it sync, move the one `await` out to the three callers, all of which are already async) would avoid the break entirely and drop the `AsyncMock` churn, but it makes invalidation a caller responsibility that's easy to forget at a fourth call site. I'd keep your approach. The in-repo blast radius is fully contained and the CHANGELOG entry is the right mitigation — item (3) above is just about making that entry actionable.

Nothing here is a problem with the approach; it's mostly description-vs-code drift plus a few consistency items. Happy to re-review quickly once (1)-(3) are in.

msureshkumar88 (changes requested):
Thanks for the follow-through on this one — the round-1 and round-2 items are almost all closed out, and the additions since then are genuinely good work. Specifically:

- `_invalidate_auth_cache()` landing as a plain `await` is much cleaner than the earlier `create_task`/`asyncio.run` helper, and raising the failure log to WARNING with `security_event=revocation_cache_invalidation_failed` / `security_severity=high` is exactly the right treatment.
- The repair guard now keys off actual mutation (`if api_token and api_token.is_active:`) rather than state, and `test_revoke_token_duplicate_already_inactive_does_not_commit` pins that behaviour.
- Shipping the data-repair migration (`e5136a7c9d01`) in this PR rather than deferring it — nice call. It's idempotent, guarded on table/column existence, and I verified `alembic heads` resolves to a single head with `down_revision = d21698ae4a19` matching main.
- Adding `SecurityValidator.sanitize_log_message()` on `revoked_by` / `reason` wasn't asked for and closes CWE-117 on the one genuinely user-influenced field.

I ran the suite locally: 92 tests pass across `test_token_blocklist_service.py`, the migration test, `test_auth_logout.py`, `test_auth_idle_timeout.py`, and `routers/test_auth.py`, with `mcpgateway/services/token_blocklist_service.py` at 100% line coverage.

Marking this as changes-requested for the two items below — both are small and neither is a code change. The rest is optional and I'd happily merge without it.

---

## Blocking

### 1. PR description has drifted from the code

This is the one item from the previous round still open. The **Notes → Changes Made** section says:

> - Added `asyncio` import for cache invalidation
> - Added `_invalidate_auth_cache()` helper method:
>   - Handles both async context (create_task) and sync context (asyncio.run)

Neither is true of the current code — the helper is a straight `await auth_cache.invalidate_revocation(jti)`, and `asyncio` isn't imported at all (`grep asyncio mcpgateway/services/token_blocklist_service.py` returns nothing). Correctly so; the earlier version was worse. It's just the description that didn't follow.

Same section is also missing three things that did land:

- the Alembic data-repair migration `e5136a7c9d01_repair_revoked_api_token_status.py` and its test
- `routers/auth.py` logout switching to `get_token_blocklist_service()` with no `db`, so the service opens its own session (the #2871 pattern)
- the log-sanitization pass over `jti` / `reason` / `revoked_by`

Worth also carrying over the note that the **Verification** section's test list still describes `test_revoke_token_invalidates_auth_cache_in_async_context` as "tests create_task path when event loop exists" — the test is fine, the description of it isn't.

### 2. Release needs a minor bump, not a patch

`TokenBlocklistService.revoke_token()` going sync → async is a real breaking change for anything off-repo. In-repo we're clean — I grepped and all three call sites are awaited (`auth.py:1857`, `admin.py:4966`, `routers/auth.py:339`), no stragglers. But `mcpgateway` ships on PyPI, and an unmigrated downstream caller gets:

```python
success = blocklist_service.revoke_token(jti=...)   # coroutine — truthy
if not success:                                      # never fires
    raise HTTPException(500, "Failed to revoke token")
# token NOT revoked, caller reports 200, only a RuntimeWarning at GC
```

A logout that silently doesn't revoke is CWE-613 with no signal at all. The CHANGELOG entry you added covers it well — this is purely about making sure the version bump matches so semver actually warns people.

If you'd rather reduce the downstream radius to zero, a one-release deprecated shim would also do it, though I don't think it's required given the CHANGELOG entry:

```python
def revoke_token_sync(self, *args, **kwargs) -> bool:
    """Deprecated sync wrapper. Do not call from within an event loop."""
    return asyncio.run(self.revoke_token(*args, **kwargs))
```

---

## Suggestions (cheap, and I think worth folding into this PR)

### 3. Collapse the SELECT-then-mutate into one conditional UPDATE

`token_blocklist_service.py:126-131` (and the duplicate at `:154-159`) does a read-modify-write:

```python
api_token = db.execute(select(EmailApiToken).where(EmailApiToken.jti == jti)).scalar_one_or_none()
repaired = False
if api_token and api_token.is_active:
    repaired = True
    api_token.is_active = False
```

A single statement does the same job and fixes two smaller things at once:

```python
result = db.execute(
    update(EmailApiToken)
    .where(EmailApiToken.jti == jti, EmailApiToken.is_active.is_(True))
    .values(is_active=False)
)
repaired = result.rowcount > 0
```

What it buys:

- **Removes a missing-rollback hazard.** On the `self.db is not None` path, `api_token.is_active = False` is staged on the *caller's* session before `db.commit()`. If that commit raises, the `except` at `:204` logs and returns `False` without rolling back. Since `get_db()` commits on successful request exit (`db.py:5990`), the caller's session can then persist `is_active=False` even though the function reported failure. Dormant today because no in-repo caller passes `db`, but it's a public constructor param.
- **Removes a concurrent-revoke race.** The `existing` pre-check and `db.add(revocation)` aren't atomic, and `jti` is `unique=True`. Two racing revocations mean the loser hits `IntegrityError`, falls into the blanket `except`, returns `False`, and `routers/auth.py:342` raises HTTP 500 on a logout that actually succeeded. Fail direction is safe (the token *is* revoked) but the response and the log are both wrong. Worth catching `IntegrityError` explicitly and re-checking `existing` before returning.
- Drops one round trip per revocation — including session-token revocations, which never have an `EmailApiToken` row to find.

### 4. Move the cache invalidation outside the DB context manager

`token_blocklist_service.py:166`:

```python
with fresh_db_session() as db:
    ...
    if existing:
        if repaired:
            db.commit()
        logger.debug(...)
        await self._invalidate_auth_cache(jti)   # <-- holds a pooled connection across a Redis round trip
        return True
```

Given the connection-pool sizing guidance in AGENTS.md, it'd be better to set a flag and do the `await` after the `with` block exits. One-line move.

### 5. One integration test for the actual invariant

Right now every logout test patches `get_token_blocklist_service` with a mock, so nothing asserts the thing #5136 is actually about: after a real `POST /auth/logout`, the DB holds `EmailApiToken.is_active = False` **and** a matching `TokenRevocation` row. The unit tests cover each half separately and the coverage number is 100%, but the end-to-end invariant is unproven.

The rewritten `test_auth_logout.py` already has the in-memory SQLite fixtures for this, so it should be a short test — drive the real service instead of the mock and assert both rows.

---

## Minor notes (happy for these to be follow-up issues)

- **`self.db is not None` branch is now production-dead.** All three call sites use `get_token_blocklist_service()` with no `db`, so only tests reach `:118-145`. It also carries a byte-for-byte copy of the repair block at `:152-167` — two copies that have to stay in sync. Factoring both into one `_revoke_in_session(db, jti, ...)` helper would remove the drift risk without touching the public signature.
- **Undocumented visibility change.** Every default token listing filters `is_active.is_(True)` (`token_catalog_service.py:580,628,647,675,735,775,1290,1309`, `admin.py:10883`). Now that auto-revocation flips the flag, idle-timed-out and logged-out tokens disappear from the default Admin UI list and `GET /tokens` — you need `include_inactive=true` to see them. That matches what manual revocation has always done, so it's the right semantics, but operators will see rows vanish. One line under `### Changed` in the CHANGELOG would cover it. (Related: the duplicate-name guard at `token_catalog_service.py:510` also filters `is_active`, so a token name is now reusable after an idle timeout.)
- **Sync Redis client in an async method.** `_get_redis_client()` returns the blocking `redis` client, and `redis_client.setex()` at `:185` is now called from an `async def` — it can block the event loop for up to the configured 2s socket timeout. Entirely pre-existing, just more visible after the async conversion. Follow-up issue rather than this PR.
- **Failure-path logging is thinner than the success path.** `logger.error("Failed to revoke token %s: %s", ...)` at `:205` has no `extra={"security_event": ...}` fields, while the success log at `:194` does — the failure is arguably the more security-relevant of the two. `exc_info=True` would help too. And `logger.debug("Token %s already revoked", ...)` on a logout path is probably INFO material; it's the signal that a session was already terminated.
- **Log identifier consistency.** Debug logs use `jti[:8]`, info/error logs use the full `jti`. Worth picking one. Also `sanitize_log_message(jti[:8])` on an 8-char slice of an opaque UUID isn't buying much.
- **Migration nits.** `is_active = false` / `= true` literals need SQLite ≥ 3.23 — `sa.false()` or `0`/`1` is more portable. `downgrade()` being a deliberate no-op is correct, but an explicit `pass  # no-op: data repair is not reversible` would stop it reading as an unfinished stub. And the docstring could cite #5136 (the revision id `e5136a7c9d01` looks like it does, but that's a coincidence).
- **Migration test doesn't quite test idempotency.** In `test_upgrade_is_idempotent_and_skips_missing_tables`, the second `_run_upgrade` still early-returns because `token_revocations` is never created. Creating both tables with data and running the upgrade twice would exercise the real path.

---

## Scope

No unrelated changes that I can see. The `.secrets.baseline` diff is line-number churn only (I checked — zero substantive hunks). The `test_auth_logout.py` fixture rewrite from `MagicMock` to real in-memory SQLite is large, but it's forced by the service now opening its own session, so it belongs here.

On the shape of the fix overall: splitting it across service, template, and migration is right, and it's the minimal set. The template alone would hide the symptom while `GET /tokens` kept returning `{"is_active": true, "is_revoked": true}`; the service alone would still render `Inactive` + `Revoked`, which #5136's "Expected Behavior" explicitly rules out; and neither reaches rows that were already inconsistent. Good call leaving the schema alone — `is_active` and `jti` both already exist and `jti` is already uniquely indexed.

Happy to re-review quickly once (1) and (2) are in.

msureshkumar88 (commented):
Re-reviewed after the latest push (`098c4b13`). Both blocking items from the last round are resolved:

- **PR description drift** — fixed. The description now matches the code (no stale `asyncio`/`create_task` text), and covers the migration, the `routers/auth.py` session change, and the log-sanitization pass.
- **Migration portability** — fixed. `e5136a7c9d01` now uses `sa.true()`/`sa.false()` instead of literal `1`/`0`, verified against a compiled PostgreSQL dialect test.

On the release-version question from earlier rounds: I checked the actual release history (`git log -- pyproject.toml`) — every version bump (1.0.1 through 1.0.7) landed in a dedicated `Release x.y.z` PR, never in a feature PR. So `pyproject.toml` staying at `1.0.7` here is correct, not a gap — the `## [Unreleased] → Breaking Changes` entry you already added is exactly the mechanism the release PR reads to know to cut `1.1.0` instead of a patch. Nothing to change there.

Ran the full affected suite locally (`test_token_blocklist_service.py`, `test_auth_logout.py`, `test_auth_idle_timeout.py`, `routers/test_auth.py`, the migration test, `test_admin.py`) — all green, single alembic head. No blocking findings this round.

## Suggestions

**1. `revoke_token()` has a TOCTOU race on concurrent revocation of the same `jti`** (`token_blocklist_service.py:122-143`, duplicated at `146-169`)

The existing-check and the `TokenRevocation` insert aren't atomic, and `jti` is the primary key. Two near-simultaneous revocations of the same token (e.g. idle-timeout and logout landing close together) mean the loser hits `IntegrityError` on commit, falls into the blanket `except Exception` at `:200`, and returns `False`. `routers/auth.py:342` then raises HTTP 500 on a logout that actually succeeded — the token *is* revoked, but the caller sees a failure and the log says revocation failed when it didn't.

Two independent things would close this:
- Catch `IntegrityError` around the insert specifically, `rollback()`, and treat it as success (the row exists — the concurrent request won).
- Replace the SELECT-then-mutate on `EmailApiToken.is_active` with a single conditional `UPDATE ... WHERE jti = :jti AND is_active IS TRUE` (rowcount tells you if a repair happened). This also removes a smaller, currently-dormant issue: on the `self.db`-provided path, `api_token.is_active = False` gets staged on the caller's session before this method's own commit; if that commit raises, the `except` returns `False` without rolling back, and the caller's session can still persist the flip on request-exit. No in-repo caller passes `db` today, so it's latent, but it's a public constructor parameter.

**2. CHANGELOG doesn't mention the listing-visibility side effect**

Every default token-list query filters `is_active.is_(True)` (`token_catalog_service.py` — several call sites, `admin.py`). Now that auto-revocation flips that flag, idle-timed-out and logged-out tokens will vanish from the default Admin UI list and `GET /tokens` (need `include_inactive=true` to see them), and a revoked token's name becomes reusable. That's the same semantics manual revocation has always had, so it's correct — just worth a line under `### Fixed` or `### Changed` so it isn't a surprise on upgrade.

## Minor (happy as follow-ups)

- `self.db is not None` branch (`:118-143`) is a byte-for-byte duplicate of the `fresh_db_session()` branch (`:146-169`) and is only reachable from tests today — worth factoring into one helper at some point so a future fix doesn't land in only one copy.
- The failure log at `:200-202` (`logger.error("Failed to revoke token...")`) doesn't carry the `extra={"security_event": ...}` fields or `exc_info=True` that the success log two lines below has, even though the failure is arguably the more security-relevant of the two.
- Debug logs use `jti[:8]`, info/error logs use the full `jti` — pick one for easier correlation.

Nothing here blocks merging — the core fix (service + template + migration) is the right, minimal shape for #5136, and it's well tested including a real end-to-end logout test that checks both DB rows.

gandhipratik203 (commented):
Thanks for working through this one, the two-layer fix (service consistency + badge priority) is the right call, and the breaking-change disclosure in the PR body is much appreciated. LGTM once the exisitng comments are adressed.

gandhipratik203 (approved):

msureshkumar88 (changes requested):
Thanks for tracking down the root cause here — the fix correctly targets `EmailApiToken.is_active`, which is also consulted by the opaque-token auth lookup (`mcpgateway/auth.py:669`), so this closes the actual security-relevant inconsistency behind #5136, not just the badge display. The migration is idempotent, schema-guarded, and chains cleanly to the current head with a single-head result, and the new unit/migration tests around the core fix look solid.

One blocking issue and a handful of hardening suggestions for this security-sensitive path before merge:

**Blocking**

- `tests/unit/mcpgateway/test_admin_logout_token_revocation.py` (not touched by this PR) still mocks `revoke_token` with a plain `MagicMock`. Since `revoke_token()` is now `async`, `await`-ing that mock raises `TypeError: object bool can't be used in 'await' expression` inside `_admin_logout`, which gets silently swallowed by the existing broad `except Exception`. I ran the test and confirmed it passes green while logging `"Failed to revoke token during admin logout: object bool can't be used in 'await' expression"` — so it's no longer actually verifying that logout revokes the token. Worth switching those mocks to `AsyncMock` (lines ~94 and ~222) so this path is genuinely covered again.

**Suggestions** (non-blocking, but worth a look given this touches token revocation)

- `TokenBlocklistService.revoke_token()` now duplicates the existing/api_token repair logic across the `self.db` and `fresh_db_session` branches. Since all in-repo callers go through `get_token_blocklist_service()` with no `db` arg, the `self.db` branch is effectively only exercised by tests — a future fix to the repair logic could easily land in just one branch and silently reintroduce the badge-inconsistency bug for the other.
- The outer `except` in `revoke_token()` doesn't call `db.rollback()` when `self.db` is a caller-supplied session. If `db.commit()` fails (e.g. `IntegrityError`), the session is left in a broken state for the rest of that request.
- Two concurrent `revoke_token()` calls for the same `jti` (e.g. parallel idle-timeout checks, a double logout) could both pass the existing-check before either commits, and the second insert would hit an `IntegrityError` on the `jti` primary key — worth a regression test for this race, since the caller would see `False` even though the token was in fact revoked.
- The early return on an already-revoked token was removed, so duplicate/retried `revoke_token()` calls now re-emit a full `security_event: token_revocation` audit log entry and reset the Redis TTL each time. Could make audit trails noisier than intended for idempotent calls.
- Cache-invalidation failures inside `revoke_token()` are only visible via a WARNING log — no metric/counter, so a sustained Redis outage affecting revocation-cache invalidation could go unnoticed by anyone watching dashboards rather than grepping logs.

None of the suggestions block the core fix, but given how security-sensitive token revocation is, addressing the stale-mock test (and ideally the race/rollback points) before merge seems worthwhile.

PR #5153, opened 2026-06-09
fix: use value='' for auth type None option and fix registry grid vis…

madhu-mohan-jaishankar:
## 🔗 Related Issue
Closes # *(no linked issue — discovered via test suite)*

---

## 📝 Summary

Three separate bugs were causing 12 test failures and 1 error in the Playwright UI test suite:

1. **Auth type `None` option used `value="none"` instead of `value=""`** (10 failures)  
   All 6 auth type `<select>` elements in admin.html (tools, gateways, agents — both add forms and edit modals) had `<option value="none">None</option>`. Tests expect `option[value=""]` (which is what the JS already sets via `authTypeField.value = entity.authType || ""`), and `select_option("")` timed out because no option with `value=""` existed. The backend's `normalize_auth_type` validator already treats `""` as the canonical "no auth" sentinel, making `""` the correct value.

2. **MCP Registry `#server-grid` had zero height when empty** (1 failure)  
   When a filter returned 0 servers, the `#server-grid` grid container had no children and therefore no height, causing Playwright's `to_be_visible()` to fail. The "No servers found" empty state was placed *outside* the grid div. Moved it inside as a `col-span-full` child so the grid always has content and a non-zero bounding box.

3. **`clear_filters()` used a weaker wait than `_wait_for_registry_refresh()`** (1 failure)  
   The old implementation confirmed the HTTP response arrived but didn't verify the HTMX DOM swap completed before counting server cards, causing a race condition. Replaced with the existing `_wait_for_registry_refresh()` helper which additionally verifies the DOM `innerHTML` has changed.

4. **`page.set_content()` in security test fixture timed out** (1 error)  
   The `iframe_host_with_teams` fixture used `page.set_content()` without `wait_until`, defaulting to `"load"` which waits for the iframe's full admin UI to load — exceeding the 60 s timeout. Added `wait_until="domcontentloaded"` so it returns as soon as the host HTML is parsed, leaving iframe load to the individual test assertions.

---

## 🏷️ Type of Change
- [x] Bug fix

---

## 🧪 Verification

| Lint suite                | `make lint`     | Pending |
| Unit tests                | `make test`     | Pending |
| Coverage ≥ 80%            | `make coverage` | Pending |

*Running the 12 previously failing Playwright tests individually to confirm fixes — in progress.*

---

## ✅ Checklist

---

## 📓 Notes

### DIFF
--- a/mcpgateway/templates/admin.html
+++ b/mcpgateway/templates/admin.html
@@ -3904,7 +3904,7 @@
                   id="auth-type"
                   class="mt-1 px-3 py-2 block w-full rounded-md border border-gray-300 dark:border-gray-700 shadow-sm focus:border-indigo-500 focus:ring-indigo-500 dark:bg-gray-900 dark:placeholder-gray-300 dark:text-gray-300"
                 >
-                  <option value="none">None</option>
+                  <option value="">None</option>
                   <option value="basic">Basic</option>
                   <option value="bearer">Bearer Token</option>
                   <option value="authheaders">Custom Headers</option>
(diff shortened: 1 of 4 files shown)


gcgoncalves (approved):

vishu-bh (approved):
Thanks @madhu-mohan-jaishankar changes are clean!!

LGTM ☘️

PR #5156, opened 2026-06-09
fix(playwright): align auth type option value and fix MCP registry files

marekdano:
Auth type selects use `value="none"` (not `value=""`) for the None option, matching what the JS bundle expects (case "none"). Tests were written expecting `value=""` which caused assertion failures across agents, gateways, and tools Playwright tests. Updated all affected test files and removed the erroneous "none" → "" mapping in `AgentsPage.set_auth_type()`.

 MCP registry filters used `hx-include="[name='auth_type']"` which matched every `auth_type` select in the entire document (6 hidden admin form elements, all defaulting to value="none"), sending `auth_type=none` to the server and returning 0 results. Fixed by scoping hx-include to element IDs (#auth-filter, #category-filter, #search-input). Added server-side normalisation in `catalog_partial` to treat empty strings and "none" as no filter. Tightened `clear_filters()` to use `_wait_for_registry_refresh()` for proper DOM-swap synchronisation.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -5,9 +5,9 @@ tag = False
 sign-tags = True
 tag_name = v{new_version} # tag format (only used if you flip tag=True later)
 parse = (?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)(-(?P<release>BETA|RC)-(?P<build>\d+))?
-serialize = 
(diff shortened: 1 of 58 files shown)


marekdano:
The change value="none" was reversed to value="". No need to update Playwright tests.
Closing this PR

PR #5157, opened 2026-06-09
fix(csrf): add missing API paths to csrf_exempt_paths; fix env drift

### LINKED ISSUE #5151: [BUG]: Multiple non-admin POST endpoints fail CSRF validation when jwt_token cookie is HttpOnly
## Description

Multiple Admin UI features fail with `{"detail": "CSRF validation failed", "code": "CSRF_TOKEN_INVALID"}` when the `jwt_token` cookie has the HttpOnly flag set. The global `CSRFMiddleware` applies to all non-exempt paths — and several paths called from the UI are missing from `csrf_exempt_paths`.

## Root Cause

`getAuthToken()` in the JS bundle attempts to read the `jwt_token` cookie via `getCookie()`. If the cookie is HttpOnly, `getCookie()` returns empty string and localStorage fallback also fails, so `getAuthToken()` returns `""`. An empty Bearer value is sent as `Authorization: Bearer ` (empty), which `_extract_bearer_token()` in `csrf_middleware.py` rejects (guards against empty tokens). The middleware then falls through to CSRF validation, which fails because no `X-CSRF-Token` header is sent.

Additionally, `CSRF_EXEMPT_PATHS` set via environment variable overrides the `config.py` default entirely — so any deployment that sets `CSRF_EXEMPT_PATHS` in `.env` loses exemptions that were added to the code default later (e.g., `/rpc`).

## Affected Paths

| Path | Issue |
|------|-------|
| `/rpc` | In `config.py` default but absent from `.env.example`'s `CSRF_EXEMPT_PATHS` — any deployment using the example env loses this exemption |
| `/api/metrics/cleanup` | Not in exempt list; uses direct `document.cookie` regex (not `getAuthToken()`); no CSRF header sent |
| `/api/metrics/rollup` | Same as above |
| `/tokens` | Relies on `getAuthToken()` returning a valid token |
| `/teams/{id}/join` | Relies on `getAuthToken()` |
| `/teams/{id}/join-requests/{id}/approve` | Relies on `getAuthToken()` |
| `/llmchat/connect` | Relies on `getAuthToken()` |
| `/llmchat/disconnect` | Relies on `getAuthToken()` |
| `/toolops/enrichment/enrich_tool` | No Authorization header in request; not in exempt list |
| `/toolops/validation/generate_testcases` | No Authorization header; not in exempt list |
| `/toolops/validation/execute_tool_nl_testcases` | No Authorization header; not in exempt list |

## Secondary Issue: CSRF_EXEMPT_PATHS Drift

`CSRF_EXEMPT_PATHS` in `.env.example` (line 67) is a hardcoded list that must be kept in sync with `config.py`'s `default_factory`. Currently `/rpc` and `/admin/login`, `/admin/forgot-password`, `/admin/reset-password` are in the code default but absent from `.env.example`. Any operator who copies `.env.example` loses these exemptions.

## Steps to Reproduce

1. Ensure `jwt_token` cookie has HttpOnly flag (default in many deployments)
2. Log in to Admin UI
3. Attempt any of: token creation, team join, LLM chat, tool enrichment/validation, metrics cleanup/rollup
4. Observe 403 `CSRF_TOKEN_INVALID` response

## Fix Options

**Option A** — Add all programmatic API paths to `csrf_exempt_paths` in `config.py` (same rationale as `/mcp/`, `/sse`, `/rpc`): `/api/metrics/cleanup`, `/api/metrics/rollup`, `/toolops/`, `/tokens`, `/teams/`, `/llmchat/`

**Option B** — Remove `CSRF_EXEMPT_PATHS` from `.env.example` and rely solely on `config.py` defaults. Operators who need to override can still set the env var explicitly.

**Option C** — Make the JS `getAuthToken()` function work reliably regardless of HttpOnly by exposing a `/auth/token` endpoint that returns the current session token for authenticated users.

## Affected Files

- `mcpgateway/config.py` — `csrf_exempt_paths` default list
- `.env.example` — `CSRF_EXEMPT_PATHS` hardcoded list (out of sync with `config.py`)
- `mcpgateway/templates/maintenance_partial.html` — direct cookie regex instead of `getAuthToken()`
- `mcpgateway/static/bundle-Cp9iMBCD.js` — `getAuthToken()`, `fetchWithTimeout()` calls

msureshkumar88:
## Summary

- Adds `/tokens`, `/teams/`, `/llmchat/`, `/toolops/`, `/api/metrics/` to `csrf_exempt_paths` in `config.py` — these programmatic API paths were causing `403 CSRF_TOKEN_INVALID` when `jwt_token` cookie had the HttpOnly flag set (same exemption rationale as `/mcp/`, `/sse`, `/rpc`)
- Comments out the hardcoded `CSRF_EXEMPT_PATHS` line in `.env.example`; setting that env var replaces the `config.py` default entirely, so operators copying the example file were silently losing `/admin/login`, `/admin/forgot-password`, `/admin/reset-password`, and future additions
- Adds regression tests: 11-path parametrized CSRF middleware test + two `default_factory` assertions that bypass env-var overrides

Fixes #5151

> **Note:** `maintenance_partial.html` direct `document.cookie` regex (lines 179/300) is a separate hardening item tracked for follow-up — it does not affect the 403 fix delivered here.

### DIFF
--- a/mcpgateway/config.py
+++ b/mcpgateway/config.py
@@ -458,6 +458,11 @@ class Settings(BaseSettings):
             "/sse",  # Exempt: SSE is a server-sent event stream, not vulnerable to CSRF
             "/message",  # Exempt: MCP SSE message endpoint
             "/rpc",  # Exempt: JSON-RPC is a programmatic protocol, not browser-based
+            "/api/metrics/",  # Exempt: programmatic admin API, authenticated via session cookie
+            "/toolops/",  # Exempt: programmatic tool operations API, authenticated via session cookie
+            "/tokens",  # Exempt: token management API, authenticated via session cookie
+            "/teams/",  # Exempt: team management API, authenticated via session cookie
+            "/llmchat/",  # Exempt: LLM chat API, authenticated via session cookie
         ],
         description="Paths exempt from CSRF protection",
     )
--- a/mcpgateway/config.py
+++ b/mcpgateway/config.py
@@ -458,12 +458,12 @@ class Settings(BaseSettings):
             "/sse",  # Exempt: SSE is a server-sent event stream, not vulnerable to CSRF
             "/message",  # Exempt: MCP SSE message endpoint
             "/rpc",  # Exempt: JSON-RPC is a programmatic protocol, not browser-based
-            "/api/metrics/",  # Exempt: programmatic admin API, authenticated via session cookie
-            "/toolops/",  # Exempt: programmatic tool operations API, authenticated via session cookie
-            "/tokens",  # Exempt: token management API, authenticated via session cookie
-            "/teams/",  # Exempt: team management API, authenticated via session cookie
-            "/llmchat/",  # Exempt: LLM chat API, authenticated via session cookie
-            "/api/logs/",  # Exempt: structured log search API, authenticated via session cookie
+            "/api/metrics/",
+            "/toolops/",
+            "/tokens",
+            "/teams/",
+            "/llmchat/",
+            "/api/logs/",
         ],
         description="Paths exempt from CSRF protection",
     )
--- a/mcpgateway/config.py
+++ b/mcpgateway/config.py
@@ -463,6 +463,7 @@ class Settings(BaseSettings):
             "/tokens",  # Exempt: token management API, authenticated via session cookie
             "/teams/",  # Exempt: team management API, authenticated via session cookie
             "/llmchat/",  # Exempt: LLM chat API, authenticated via session cookie
+            "/api/logs/",  # Exempt: structured log search API, authenticated via session cookie
         ],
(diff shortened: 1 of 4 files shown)


mcpgateway/config.py
cafalchio, line 457: please remove the comments. csrf_exempt_paths is already quite clear.

cafalchio (dismissed):
LGTM, just few minor comments.

mcpgateway/config.py
msureshkumar88: Removed.

prakhar-singh1928 (approved):
LGTM. approved

ja8zyjits (approved):
LGTM 2 approved already

PR #5158, opened 2026-06-10
fix(resources): show meaningful conflict message for duplicate resource name

### LINKED ISSUE #4991: [BUG]:  Resource creation shows generic HTTP 409 error instead of meaningful duplicate resource message
### 🐞 Bug Summary
While adding a new Resource under Gateway Administration → Resources, if a resource with the same name already exists, the UI displays a generic error message: Failed to add Resource (http 409). It doesn't specify the cause of the error.
---

### 🧩 Affected Component

---

### 🔁 Steps to Reproduce

1. Navigate to the resources tab
2. Create a new resource with a duplicate name
3. Fill other required fields
4. click add resource

---

### 🤔 Expected Behaviour
An error message that clearly tells user what is wrong  and in which field seems ideal behaviour.

---

### 📓 Logs / Error Output

Image

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Version or commit | main |
| Runtime | Python 3.11, FastAPI |
| Platform / OS | Any |
| Container | Any |

---

shoummu1:
## :pushpin: Summary

When a resource with a duplicate name was created via the Admin UI, the error shown
was a generic "Failed to add Resource (http 409)" with no indication of which field
caused the conflict or why. This fix surfaces a clear, actionable message to the user.

## :link: Related Issue

Closes: #4991

1. Navigate to **Gateway Administration → Resources**
2. Create a resource with any name (e.g. `my-resource`)
3. Attempt to create a second resource with the same name
4. Observe the error shown in the UI

## :bug: Root Cause

The service layer had no application-level check for duplicate resource names before
attempting the INSERT. The database had no unique constraint on the `name` column, so
no `IntegrityError` was raised either. Any conflict that did surface fell through to a
generic exception handler in `admin.py` that returned an unformatted HTTP 409 with no
user-facing detail.

## :bulb: Fix Description

Two-layer defence — application-level pre-check backed by a DB constraint:

- **`ResourceNameConflictError`** — new exception class (mirrors `ResourceURIConflictError`)
  raised when a name collision is detected before the INSERT
- **Name-conflict pre-check** added to `register_resource` (public and team scopes) and
  `register_resources_bulk` (respects skip/fail/update/rename conflict strategies)
- **DB unique constraint** `uq_team_owner_gateway_name_resource` + partial index
  `uq_team_owner_name_resource_local` added to the `Resource` model and backed by an
  idempotent Alembic migration with a pre-flight duplicate check
- **Error handlers** added to `admin_add_resource`, `admin_edit_resource`, and
  `create_resource` (REST API) returning HTTP 409 with `{"message": "A resource with
  this name already exists: <name>", "success": false}`
- **`error_formatter.py`** updated to map both the PostgreSQL constraint name and the
  SQLite column-path format to the same friendly message (IntegrityError fallback path)
- **`ResourceURIConflictError.resource_id`** type annotation corrected from `Optional[int]`
  to `Optional[str]` to match the UUID string PK on `DbResource`

## :test_tube: Verification

| Check | Command | Status |
|---|---|---|
| Lint suite | `make lint` | :white_check_mark: |
| Unit tests | `make test` | :white_check_mark: 19,133 passed, 0 new failures |
| Coverage ≥ 90 % | `make coverage` | |
| Manual regression no longer fails | Navigate to Resources → add duplicate name → confirm descriptive message shown | |


jonpspri:
@shoummu1 This has just surfaced for merge.  Could you please rebase and readjust the alembic if needed?  I've asked @marekdano to review.

marekdano (commented):
### Some findings

#### Blocking

**1. Alembic migration `down_revision` mismatch** `mcpgateway/alembic/versions/279184dfd71d_...py`

The file docstring says `Revises: e28566875fa4` but the actual `down_revision` variable says `"0a089912b5f0"`. These must agree — Alembic uses the variable, not the comment.
If `0a089912b5f0` is not the true current head this creates a multiple-heads situatio and breaks migrations for every other contributor. Needs a `alembic heads` verification before merge.

**2. URI check guard weakened from `elif ... and team_id:` to `elif ...:`** `resource_service.py`

  ```python
  -elif visibility.lower() == "team" and team_id:
  +elif visibility.lower() == "team":
  ```

The name pre-check raises `ResourceValidationError` when `team_id` is absent, so in the current code control never reaches the URI check without a `team_id`. But a future caller that bypasses the name check would reach the URI query with `team_id=None`, silently filtering on `team_id IS NULL` and matching public/null-owned resources. Adding the same guard (`if not team_id: raise ResourceValidationError(...)`) on the URI check block would make both paths self-contained.

---

#### Suggestions

**3. Bulk registration: name conflict for `rename`/`update` strategies falls through incorrectly** `resource_service.py`

  ```python
  # rename / update strategies fall through to URI-based handling below
  ```

When a name conflicts but the URI differs, `update` looks up the existing resource by `(new_uri, gateway_id)` — finds nothing — and tries to create a new row that immediately hits the DB unique name constraint. `rename` has the same problem. The result is a raw `IntegrityError` instead of a graceful strategy-consistent response. Either handle both strategies explicitly here, or add a comment documenting the known limitation.

**4. Use `elif` chain in `admin.py` error handlers** `admin.py`

The new blocks use bare `if` rather than `elif`. Since each block returns, correctness is unaffected, but `elif` makes the mutual exclusion explicit and is consistent with how the rest of the handler is written.

**5. `ResourceValidationError` not tested for `register_resources_bulk`**

Tests cover `skip` and `fail` bulk name-conflict strategies but not the `ResourceValidationError` path (missing `team_id` on a team-scoped bulk create). The guard at `resource_service.py` makes this reachable.

  ---

#### Minor / Style

**6. Logging from inside the exception constructor** `resource_service.py`

`logger.info(...)` inside `__init__` fires on every `raise`, including in tests. Consider removing it and letting callers log at the appropriate level, as the existing handlers already do before re-raising.

shoummu1:
Thanks for the thorough review, @marekdano. I have rebased and added the latest commit which addresses all six points:

**Blocking**

1. **Alembic `down_revision` mismatch** — Good catch, and it was actually worse than a comment/variable mismatch: *both* `Revises: e28566875fa4` and `down_revision = "0a089912b5f0"` pointed at revisions that already have children mid-chain, so either would have forked into multiple heads. I traced the DAG and pointed both at the real current head, `e198602c3c1e`. `alembic heads` now reports a single head and `alembic upgrade head` applies the full chain cleanly on a fresh DB — which also let me drop the `0a75bc55b762` merge migration that was papering over the fork.

2. **URI check guard** — Restored the `if not team_id: raise ResourceValidationError(...)` guard on the team branch of the URI check, so both the name and URI paths are self-contained.

**Suggestions**

3. **Bulk rename/update fall-through** — You're right that `update`/`rename` don't reconcile a name-only conflict and end up hitting the unique constraint. I've documented it as a known limitation at the fall-through for now: the `IntegrityError` is caught by the per-resource handler and recorded in `stats["failed"]/["errors"]`, so the import stays safe (no partial/duplicate rows) even though the outcome isn't strategy-consistent. Happy to add explicit handling in this PR if you'd prefer.

4. **`elif` in admin.py handlers** — Good point on making the mutual exclusion explicit. I initially switched these to `elif`, but the repo's pylint config enforces `no-else-return` (R1705), which flags `elif`/`else` after a `return`. Since every branch returns, the handlers are already mutually exclusive, so I've kept the explicit early-`return` `if` chain — which is the pattern the rest of the file (and the linter) expects. Let me know if you'd rather I restructure it differently.

5. **`ResourceValidationError` bulk test** — While adding the test I noticed `register_resources_bulk` wasn't validating `team_id` at all (a team-scoped bulk create silently fell through to the private-scope branch), so I added the same guard as the single-resource path plus a test that covers the raise.

**Minor / Style**

6. **Logging in exception constructor** — Removed the `logger.info(...)` from both `ResourceURIConflictError` and `ResourceNameConflictError` `__init__`; the handlers already log before re-raising.

marekdano (dismissed):
Nice work - all six points from the last round are addressed and verified against the current diff. The migration chain is clean, cross-dialect error mapping is handled, and test coverage is thorough.

### Non-blocking - please confirm before merge

  1. **Migration aborts on existing duplicate names.** `upgrade()` raises `RuntimeError` if any duplicate `(name, team_id, owner_email, gateway_id)` group exists. Since duplicate names were previously allowed, existing deployments carrying dupes will have their upgrade abort. The clear early failure is the right call - is there an upgrade/changelog note telling operators to dedupe first?

  2. **Private-scope name uniqueness is a new behavior.** The app-level pre-check only runs for `public`/`team`, but the DB constraint now also rejects two `private` resources with the same name owned by the same user under the same gateway (previously allowed), falling through to the generic `IntegrityError` path. Is this intended? If so, consider raising `ResourceNameConflictError` on the private path too, for a consistent user-facing message.

### Accepted as-is

- The bulk `update`/`rename` name-conflict limitation is fine to ship as documented - worth a follow-up issue to track.

msureshkumar88 (dismissed):
Nice work tracking down the root cause here — the two-layer defense (app-level pre-check + DB constraint with an idempotent, pre-flight-checked migration) is the right shape for this fix, and the test coverage for the new conflict paths (active/inactive, bulk skip/fail, team-without-team-id) is solid.

One thing I think needs a fix before merge, plus a few smaller items worth a look.

### Blocking

**Performance regression in `register_resources_bulk` (N+1 queries)** — `mcpgateway/services/resource_service.py:942-954`
The existing URI-conflict check batches lookups once per chunk via `existing_resources_map` (a single `WHERE uri IN (...)` for up to 500 items). The new name-conflict check issues one SELECT per resource inside the per-item loop instead, adding up to 500 extra synchronous DB round-trips per chunk. This contradicts the method's own docstring ("Batch conflict detection", "significant performance improvements"), and the only production caller (`import_service.py:1064`) is the bulk import path, so this would hit real import workloads. Could we prefetch existing names the same way URIs are prefetched (batched `WHERE name IN (...)` per chunk, keyed by `(name, gateway_id)` / team scope) instead of querying per item?

### Suggestions / minor

- **Dead code**: in `register_resource`, the `if not team_id: raise ResourceValidationError(...)` guard in the URI-check block (`elif visibility.lower() == "team":`) is unreachable — the identical guard in the preceding name-check block already raises for the same condition first. Worth collapsing into a single check.
- **Dropped log line**: `ResourceURIConflictError.__init__` used to call `logger.info("ResourceURIConflictError: %s", message)` on construction; that line was removed (not moved) while adding the `enabled` check. Conflicts are still logged at ERROR by the catch sites, so it's not a full loss of observability — just flagging that it went away without mention in the PR description.
- **Status code inconsistency**: `ResourceValidationError` (e.g. team-scoped create with no `team_id`) returns 422 from the `admin.py` handlers, but `main.py`'s `create_resource` doesn't catch it explicitly, so it falls through to `except ResourceError` → 400. Worth aligning since both paths were touched in this PR.
- **Migration test coverage**: the repo has precedent for testing unique-constraint migrations (`tests/unit/mcpgateway/db/test_token_uniqueness_migration.py`), but `279184dfd71d_add_name_uniqueness_constraint_to_.py` doesn't have an equivalent — specifically the pre-flight duplicate-detection query (`GROUP BY ... HAVING COUNT(*) > 1`) that raises `RuntimeError`. A typo there would silently break upgrades with no test to catch it.
- **E2e coverage**: unit coverage is thorough, but there's no e2e/integration test exercising the actual reported bug end-to-end (create → duplicate create → assert 409 + message via the real endpoint) — which is exactly the manual verification step described in the PR.

### Worth calling out in release notes (not a code issue)
Deployments with pre-existing duplicate resource names (same team/owner/gateway scope) won't be able to apply this migration until they deduplicate — the pre-flight check fails loudly with `RuntimeError` rather than corrupting data, but it is an upgrade blocker operators should be warned about.

No security concerns beyond a pre-existing, low-severity pattern (conflict error messages include the internal resource ID, matching what `ResourceURIConflictError` already does) — not introduced by this PR.

marekdano (changes requested):
@msureshkumar88 - 

## Findings (most severe first)

1. [Blocking] DB "backstop" does not enforce uniqueness for public/private resources **`resource_service.py` / `db.py` / migration `279184dfd71d`**

`team_id` is `NULL` for public and private resources, and SQL treats `NULL`s as distinct in unique indexes. Verified on SQLite (default DB): with both `uq_team_owner_gateway_name_resource` and the `uq_team_owner_name_resource_local` partial index in place, two public (or two private) local resources with the same name insert cleanly — only team-scoped rows (`team_id` non-NULL) are blocked. Postgres behaves identically (default `NULLS DISTINCT`).

For the Admin-UI case in #4991 (typically public/private, `team_id` NULL, `gateway_id` NULL), **only the app-level pre-check prevents duplicates; the advertised DB layer is inert.** Any path that bypasses the service check can still create dupes.

2. [Blocking] Bulk `update`/`rename` "known limitation" comment's safety claim is false **`resource_service.py`**

The comment states the incoming row "immediately hits the DB unique-name constraint … so the import stays safe (no partial/duplicate rows)." Per finding 1, for public/private local resources the constraint does not fire, so a name-only conflict under `update`/`rename` **silently creates a duplicate-named row** rather than being caught. Failure mode is worse than documented.

3. [Blocking] Rename-to-duplicate on the edit path is unguarded **`admin.py` (edit handler) / `update_resource`**

  `update_resource` has no name pre-check (`ResourceNameConflictError` is only raised in `register_resource` / `register_resources_bulk`). Combined with 1, renaming a public/private resource to an existing name via `admin_edit_resource` succeeds silently. Correspondingly, the new `isinstance(ex, ResourceNameConflictError)` branch in `admin_edit_resource` is effectively dead code — nothing on the edit path raises it.

4. [Non-blocking] Migration pre-flight over-blocks upgrades  **`mcpgateway/alembic/versions/279184dfd71d_*.py`**

`upgrade()` aborts with `RuntimeError` on any duplicate `(name, team_id, owner_email, gateway_id)` group, but the indexes it creates only enforce team-scoped rows. An operator carrying historical public/private duplicate names (the #4991 symptom) is blocked from upgrading even though the new constraint wouldn't have prevented those rows anyway. Still no changelog/upgrade note telling operators to dedupe first.
 
5. [Non-blocking] Private visibility has no clean pre-check  **`resource_service.py`**

`register_resource` only pre-checks `public` and `team`; private collisions fall through to the generic `IntegrityError`/formatter path. Given 1, private local dupes aren't caught at all. A `ResourceNameConflictError` pre-check on the private path (or an explicit decision to leave private un-scoped) would make behavior consistent and intentional.

msureshkumar88:
@marekdano thanks for the sharp catch on the NULL-distinctness issue — verified it locally against a plain SQLite unique index and you're right, two public (or private) resources with the same name insert cleanly since `team_id` is NULL for both. Pushed a fix:

**1. DB backstop now covers public/private** — both `uq_team_owner_gateway_name_resource` and `uq_team_owner_name_resource_local` now index `COALESCE(team_id, '')` instead of the bare column, so NULL no longer defeats the constraint. Verified with a raw SQLite repro before/after the change.

**2. Bulk `update`/`rename` limitation comment is accurate again** — direct consequence of #1, no separate code change needed there.

**3. Rename path now guarded** — `update_resource` had no name pre-check at all (only the URI check existed), so the `isinstance(ex, ResourceNameConflictError)` branch in `admin_edit_resource` was genuinely dead code. Added the same public/team/private pre-check pattern used in `register_resource`, and wired the exception handling into the REST `PUT /resources/{id}` endpoint in `main.py` too (it was missing a `ResourceNameConflictError` handler entirely, which would have 500'd instead of 409'd).

**5. Private visibility now pre-checked** — added to both `register_resource` and `register_resources_bulk` (owner-scoped, mirroring the public/team pattern).

**4 (non-blocking)** — agreed this is accurate as described; it's a release-note item rather than a code bug, leaving as-is for now.

One thing that came up while fixing #1: `Inspector.get_indexes()` silently drops expression-based indexes from reflection (both SQLite and Postgres), so the migration's existing-index check would never see the new `COALESCE(...)` indexes and would blow up on a second `upgrade()` run. Switched the existence check to read `sqlite_master`/`pg_indexes` directly instead, and fixed the same blind spot in the migration's own test helper.

Full `tests/unit/mcpgateway` suite, ruff, and a fresh `alembic upgrade head` all green; `alembic heads` still single head.

msureshkumar88:
## E2E Verification

Ran an end-to-end check against a real running gateway (real SQLite DB, migrations applied, `uvicorn mcpgateway.main:app`) to confirm this closes #4991. Script: `scripts/verify-pr-5158-e2e.sh`.

**Config**
```
DATABASE_URL=sqlite:///<fresh temp db>
AUTH_REQUIRED=true
MCPGATEWAY_UI_ENABLED=true
PLUGINS_ENABLED=false
ADMIN_REQUIRE_PASSWORD_CHANGE_ON_BOOTSTRAP=false  # bypass forced-password-change redirect, unrelated to this fix
```
Commit under test: `c1131b0ef` (this branch's HEAD).

**Commands**
```
bash scripts/verify-pr-5158-e2e.sh
```
Internally: `alembic upgrade head` → boot gateway → `create_jwt_token --admin` → `POST /resources` (REST, JSON) → `POST /admin/resources` (Admin UI, form) → `GET /resources?include_inactive=true`.

**Results — all 6 checks passed**
```
PASS: alembic upgrade head completed
PASS: both new unique indexes exist on resources table (uq_team_owner_gateway_name_resource, uq_team_owner_name_resource_local)
PASS: gateway server is up
PASS: first resource created
PASS: REST API returned descriptive 409 for duplicate name:
      {"detail":"A resource with this name already exists: <name>"}
PASS: Admin UI returned descriptive 409 with success:false for duplicate name:
      {"message":"A resource with this name already exists: <name>","success":false}
PASS: resource with a different name created without false-positive conflict
PASS: exactly one resource with the duplicate name persisted (rejected duplicate not inserted)

RESULT: ALL CHECKS PASSED
```

Confirms the fix: both the REST API and Admin UI now surface a clear, actionable 409 message identifying the conflicting name, instead of the generic unformatted 409 from before. The migration applies cleanly on a fresh SQLite DB and creates both unique indexes; distinct names are unaffected (no false positives), and rejected duplicates aren't persisted.

Two non-obvious things worth calling out in case they're useful for anyone else testing this manually:
- `POST /resources` needs the JSON body wrapped as `{"resource": {...}}` — FastAPI auto-embeds since the route also takes `team_id`/`visibility` as separate body params.
- A fresh bootstrap admin has `password_change_required=True`, which 303-redirects `/admin/resources` before ever reaching the conflict path — need `ADMIN_REQUIRE_PASSWORD_CHANGE_ON_BOOTSTRAP=false` or a completed password-change flow to exercise the Admin UI path directly. Unrelated to this PR, just a setup gotcha.

marekdano (dismissed):
@msureshkumar88 - this is in good shape. The three blocking issues from the last round are genuinely fixed, and at the right depth rather than patched over:

  - **DB backstop now works for public/private:** `COALESCE(team_id, '')` in both indexes closes the NULL-distinctness hole — verified with a before/after SQLite repro that two public (and two private) local resources with the same name are now rejected at the DB level.
  - **Rename/edit path guarded:** `update_resource` gained the name pre-check (with self-exclusion), and the REST `PUT /resources/{id}` handler now returns 409 instead of 500.
  - **Private visibility pre-checked** in both `register_resource` and `register_resources_bulk`.

Nice catches along the way too — the `Inspector.get_indexes()` reflection blind-spot on expression indexes, and the e2e verification covering both the REST and Admin UI paths.

ja8zyjits (dismissed):
LGTM

ja8zyjits (dismissed):
lgtm

ja8zyjits (approved):
LGTM

PR #5176, opened 2026-06-11
fix: bind join requests to team context

lucarlig:
# 🐛 Bug-fix PR

## 📌 Summary
Bind join request approvals and rejections to the route team before updating request state. This keeps the service behavior aligned with the team context checked by the routes.

## 🔁 Reproduction Steps
1. Have a pending join request for one team.
2. Call an approve or reject route using a different team in the path and the pending request id.
3. The request action should only apply when the path team matches the request team.

## 🐞 Root Cause
The route validated the caller against the path team, but the service loaded the join request by request id and status only.

## 💡 Fix Description
The team management service now requires `team_id` for approve and reject operations and includes it in the join request lookup. Admin and API routes pass the path team through to the service. Tests cover mismatched team/request handling with real database rows.

## 🧪 Verification

| Lint suite                            | `make lint`          | Not run |
| Unit tests                            | focused pytest       | Passed |
| Coverage ≥ 80 %                       | `make coverage`      | Not run |
| Manual regression no longer fails     | focused unit coverage | Passed |

## ✅ Checklist
- [x] Code formatted (`pre-commit`)

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-06-11T09:41:14Z",
+  "generated_at": "2026-06-12T17:37:44Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -5068,7 +5068,7 @@
         "hashed_secret": "a0281cd072cea8e80e7866b05dc124815760b6c9",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 17773,
+        "line_number": 17775,
         "type": "Secret Keyword",
         "verified_result": null
       }
--- a/docs/docs/design/images/classes.svg
+++ b/docs/docs/design/images/classes.svg
@@ -10430,7 +10430,7 @@
 <text text-anchor="start" x="167377.5" y="-4306.3" font-family="Arial" font-size="14.00">role_service</text>
 <polyline fill="none" stroke="#ffaabb" points="166596.5,-4298.5 168231.5,-4298.5 "/>
 <text text-anchor="start" x="166604.5" y="-4283.3" font-family="Arial" font-size="14.00">add_member_to_team(team_id: str, user_email: str, role: str, invited_by: Optional[str]): EmailTeamMember</text>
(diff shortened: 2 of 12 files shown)


msureshkumar88 (changes requested):
Thanks for the quick fix here — the cross-team test cases (Team A path / Team B request) directly exercise the PoC from the advisory and confirm the fix closes it. The mock assertions checking that `team_id` is part of the filter are a nice touch to prevent a silent regression.

**1. Inconsistent status code for the same error condition**

In `routers/teams.py`:
- `reject_join_request` now maps the service's `ValueError("Join request not found or already processed")` to `404`.
- `approve_join_request` maps the *same* `ValueError` message to `400` (via the existing `except (ValueError, TeamMemberLimitExceededError)` block).

Same root condition (request not found / wrong team / already processed), two different status codes depending on which endpoint you call. Worth aligning both to `404` for the "not found" case, keeping `400` only for genuinely client-error validation cases (e.g. team member limit).

**2. Now-unreachable fallback branches**

Since the service raises `ValueError` on a missing/mismatched join request rather than returning `None`/`False`, the `if not member` / `if not success` → `404` branches in `admin.py` and `routers/teams.py` are effectively dead code now. Not harmful, but worth either removing or adding a comment noting they're defensive-only, so future readers aren't confused about which path actually fires.

**3. Unrelated formatting churn**

The diff also includes some unrelated reformatting (`nosec B110` spacing in `admin.py`, collapsing the `list_resources`/`list_prompts` calls to one line, reflowing the `EmailTeamMember` query in `team_management_service.py`, plus the `classes.svg` regen). None of this is wrong, but it makes the security-relevant diff harder to review in isolation — consider splitting into a separate cleanup commit/PR if convenient.

None of these are security-blocking — the core fix is correct and well-tested — but #1 should be resolved for consistent API behavior before merge.

lucarlig:
Thanks Suresh. I pushed a follow-up in f49e921f6.

Addressed:
- mapped `Join request not found or already processed` to 404 for approve as well as reject
- removed the now-dead `None`/`False` fallback branches in the route/admin handlers
- updated the coverage tests to exercise the service `ValueError` path
- restored `classes.svg` so it drops out of the PR diff

Validation: `uv run pytest tests/unit/mcpgateway/routers/test_teams.py tests/unit/mcpgateway/routers/test_teams_coverage.py tests/unit/mcpgateway/services/test_team_management_service.py tests/unit/mcpgateway/services/test_team_management_service_coverage.py tests/unit/mcpgateway/test_admin.py::TestTeamJoinRequests tests/unit/mcpgateway/test_admin_module.py tests/unit/mcpgateway/test_team_governance_flags.py` -> 442 passed, 5 skipped.

PR #5178, opened 2026-06-11
fix(chore): make tag length configurable via env variables

### LINKED ISSUE #5175: [FEATURE]: Configurable limits for MCP server Tags
### 🧭 Type of Feature

- [ ] Packaging, automation and deployment (ex: pypi, docker, quay.i

### 🧭 Epic

**Title:** Configurable Tag Length Limits for MCP Gateway Servers
**Goal:** Allow gateway administrators to configure the maximum tag hardcoded 50-character limit, enabling support for system-generated tags and longer identifiers.
**Why now:** The current 50-character hardcoded limit silently drops, causing metadata loss for valid use cases like system-generated identifiers, hashes, or name spaced tags. This affects data integrity and prevents users from tracking important server metadata.

---

### 🧑🏻‍💻 User Story 1

**As a:** Gateway Administrator
**I want:** To configure the maximum allowed tag length via gateway configuration
**So that:** I can accommodate longer tags (e.g., system-generated identifiers with hashes) without silent data loss

#### ✅ Acceptance Criteria
```gherkin
Scenario: Administrator configures custom tag length limit
  Given I am a gateway administrator
  When I set the tag length limit to 200 characters in the gateway configuration
  Then the gateway should accept tags up to 200 characters
  And tags exceeding 200 characters should be rejected with a clear validation error

Scenario: Tags within configured limit are saved successfully
  Given the tag length limit is configured to 150 characters
  When I create a gateway with a tag of 120 characters
  Then the tag should be saved successfully
  And the tag should be retrievable via the gateway API

Scenario: Tags exceeding configured limit are rejected with clear errors 
  Given the tag length limit is configured to 100 characters
  When I attempt to create a gateway with a tag of 150 characters
  Then the request should fail with HTTP 400 Bad Request
  And the error message should indicate "Tag exceeds maximum length of 100 characters"
```

---
### 🧑🏻‍💻 User Story 2

As an: API Consumer
I want: To receive clear validation errors when my tags exceed the configured limit
So that: I can adjust my tags accordingly and understand why my requests fail

#### ✅ Acceptance Criteria

Scenario: Clear error message for oversized tags
  Given the tag length limit is 50 characters
  When I POST a gateway with a tag "database-prod-a1b2c3d4e5f6g7h8i9j0k1l2m3n4o5p6q7r8s9t0u1v2w3x4y5z6"
  Then the response should be HTTP 400 Bad Request
  And the response should contain the field name "tags"
  And the error should specify which tag exceeded the limit

---
Example API Request

```bash
curl --location 'http://localhost:8080/api/v1/gateways' \
--header 'Authorization: Bearer <token>' \
--header 'Content-Type: application/json' \
--data '{
  "name": "my-mcp-server",
  "url": "https://mcp.example.com/mcp",
  "transport": "STREAMABLEHTTP",
  "passthrough_headers": [],
  "auth_type": "bearer",
  "auth_token": "my_auth_token",
  "tags": [
    "my-super-long-tag-a1b2c3d4e5f6g7h8i9j0k1l2m3n4o5p6q7r8s9t0u1v2w3x4y5z6",
    "my-short-tag"
  ],
  "one_time_auth": false,
  "team_id": "team_id",
  "visibility": "team"
}'
```

---
🔗 MCP Standards Check

Notes:
- This is an internal gateway constraint, not part of the MCP specification
- Existing integrations will continue to work without modification

---
🔄 Alternatives Considered

1. Automatically truncate tags
  - Pros: No breaking changes, transparent to users
  - Cons: Silent data loss, potential duplicate tags after truncation
  - Rejected because: Silent failures are worse than explicit validation
2. Remove tag length limit entirely
  - Pros: Maximum flexibility, no configuration needed
  - Cons: Could cause database performance issues, unbounded memory usage, potential abuse
  - Rejected because: Reasonable limits are good practice

---
📓 Additional Context

Current Behavior

- Tags longer than 50 characters are silently dropped
- No validation error is returned to the client

Proposed Implementation

1. Add configuration option max_tag_length (default: 50 for backward compatibility)
2. Update validation logic to check tag length against configured value
3. Return HTTP 400 with detailed error when tag exceeds limit
4. Update API documentation to reflect configurable limit
5. Add unit tests for validation at various length limits
6. Add integration tests for configuration loading

Database Considerations

- Current schema should support longer tags
- If schema change needed, migration path should be provided
- Consider indexing implications for very long tags

Documentation Updates Needed

- API reference: Document max_tag_length configuration
- Configuration guide: Show examples of setting custom limits in .env.example

bogdanmariusc10:
## 🔗 Related Issue
Closes #5175

---

## 📝 Summary

Makes tag length limits configurable via environment variables for all ContextForge resources (tools, prompts, resources, servers, gateways). Previously, tag lengths were hardcoded at 2-50 characters. This change allows operators to configure limits based on their needs (e.g., longer tags for system-generated identifiers, hashes, or namespaced tags).

**Key Changes:**
- Added `VALIDATION_MIN_TAG_LENGTH` environment variable (default: 2, range: 1-10)
- Added `VALIDATION_MAX_TAG_LENGTH` environment variable (default: 100, range: 10-255)
- **Breaking Change**: Default max tag length increased from 50 to 100 characters to better support real-world use cases

**Affected Resources:**
All tag fields across tools, prompts, resources, servers, and gateways now use configurable validation.

---

## 🏷️ Type of Change
- [x] Bug fix
- [x] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

| Lint suite                | `make lint`     |    Pass ✅    |
| Unit tests                | `make test`     |    Pass ✅    |
| Coverage ≥ 80%            | `make coverage` |    Pass ✅    |

---

## ✅ Checklist

---

## 📓 Notes (optional)

**Breaking Change:**
Default max tag length increased from 50 → 100 characters. Tags between 51-100 chars that were previously rejected will now be accepted. To restore previous behavior, set `VALIDATION_MAX_TAG_LENGTH=50`.

**Files Modified:**
- `mcpgateway/config.py` - Added configurable settings
- `mcpgateway/validation/tags.py` - Use settings instead of hardcoded constants
- `.env.example` - Documented new environment variables
- `tests/unit/mcpgateway/validation/test_tags.py` - Added 10 test cases for configurable limits
- `.secrets.baseline` - Updated baseline

**Usage Example:**
```bash
# Support longer tags (e.g., for deployment identifiers)
VALIDATION_MAX_TAG_LENGTH=150

# Allow single-character tags
VALIDATION_MIN_TAG_LENGTH=1

### DIFF
--- a/mcpgateway/config.py
+++ b/mcpgateway/config.py
@@ -3253,20 +3253,20 @@ Disallow: /
         ge=1,
         le=100,
     )
-    validation_max_url_length: int = 767  # Match database String(767) limit to prevent validation/storage mismatch
+    validation_max_url_length: int = 2048
     validation_max_rpc_param_size: int = 262144  # 256KB
 
     validation_max_method_length: int = 128
 
-    # Tag validation limits (configurable via env) - Issue #XXXX
+    # Tag validation limits (configurable via env) - Issue #5175
     validation_min_tag_length: int = Field(
-        default=int(os.getenv("VALIDATION_MIN_TAG_LENGTH", "2")),
+        default=2,
         description=("Minimum length for individual tags. Tags shorter than this will be rejected. " "Override with VALIDATION_MIN_TAG_LENGTH environment variable. Minimum: 1, Maximum: 10"),
         ge=1,
         le=10,
     )
     validation_max_tag_length: int = Field(
-        default=int(os.getenv("VALIDATION_MAX_TAG_LENGTH", "100")),
+        default=100,
         description=(
             "Maximum length for individual tags. Tags longer than this will be rejected. "
             "Default: 100 characters. Supports system-generated tags, hashes, and namespaced identifiers. "
--- a/mcpgateway/config.py
+++ b/mcpgateway/config.py
@@ -2740,9 +2740,7 @@ class Settings(BaseSettings):
     # Experimental dataplane config
     # ===================================
 
-    dataplane_publisher: bool = Field(default=False,
-        description="Send data from CF to Rust experimental dataplane"
-    )
+    dataplane_publisher: bool = Field(default=False, description="Send data from CF to Rust experimental dataplane")
 
     # Well-Known URI Configuration
     # ===================================
@@ -3255,10 +3253,29 @@ Disallow: /
         ge=1,
         le=100,
     )
-    validation_max_url_length: int = 2048
+    validation_max_url_length: int = 767  # Match database String(767) limit to prevent validation/storage mismatch
     validation_max_rpc_param_size: int = 262144  # 256KB
 
     validation_max_method_length: int = 128
+
+    # Tag validation limits (configurable via env) - Issue #XXXX
+    validation_min_tag_length: int = Field(
+        default=int(os.getenv("VALIDATION_MIN_TAG_LENGTH", "2")),
+        description=("Minimum length for individual tags. Tags shorter than this will be rejected. " "Override with VALIDATION_MIN_TAG_LENGTH environment variable. Minimum: 1, Maximum: 10"),
+        ge=1,
+        le=10,
+    )
+    validation_max_tag_length: int = Field(
+        default=int(os.getenv("VALIDATION_MAX_TAG_LENGTH", "100")),
+        description=(
+            "Maximum length for individual tags. Tags longer than this will be rejected. "
+            "Default: 100 characters. Supports system-generated tags, hashes, and namespaced identifiers. "
+            "Override with VALIDATION_MAX_TAG_LENGTH environment variable. "
+            "Minimum: 10, Maximum: 255 (database constraint)"
+        ),
+        ge=10,
+        le=255,
+    )
 
     # Allowed MIME types
     validation_allowed_mime_types: List[str] = [
--- a/.env.example
+++ b/.env.example
@@ -302,6 +302,15 @@ VALIDATION_MIDDLEWARE_ENABLED=true
 # VALIDATION_PERIOD_TYPE_PATTERN=^(hourly|daily)$
 # VALIDATION_AGGREGATION_PATTERN=^(5m|24h)$
 # VALIDATION_ENTITY_TYPES_PATTERN=^[a-zA-Z,]*$
+
+# Tag length validation (applies to all resources: tools, prompts, resources, servers, gateways)
+# Minimum tag length (default: 2, range: 1-10)
+# VALIDATION_MIN_TAG_LENGTH=2
+# Maximum tag length (default: 100, range: 10-255)
+# Supports system-generated tags, hashes, and namespaced identifiers
+# Example: VALIDATION_MAX_TAG_LENGTH=150 for very long descriptive tags
+# VALIDATION_MAX_TAG_LENGTH=100
+
 # Permission audit logging (RBAC checks) - disabled by default for performance
 PERMISSION_AUDIT_ENABLED=false
 
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-06-11T13:01:14Z",
+  "generated_at": "2026-06-11T09:41:14Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -116,7 +116,7 @@
         "hashed_secret": "08cd923367890009657eab812753379bdb321eeb",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 690,
+        "line_number": 699,
         "type": "Basic Auth Credentials",
         "verified_result": null
       },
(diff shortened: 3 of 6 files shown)

### CODE REVIEW

**Review by ja8zyjits — DISMISSED** [review]
LGTM

**Review by msureshkumar88 — CHANGES_REQUESTED** [review]
Thanks for tackling this, Bogdan — making the tag limits configurable is a clean, well-tested change and the new test suite covers the interesting boundaries nicely. The pydantic `Field` with `ge`/`le` bounds and the `.env.example` docs are exactly the right shape for a config-driven limit.

I do have a few concerns, mostly around the bundled URL-length change. Categorized below.

## 🔴 Blocking

**1. The `validation_max_url_length` 2048 → 767 change is risky and out of scope.**
This PR is scoped to *tag* length (#5175), but it also globally drops `SecurityValidator.MAX_URL_LENGTH` from 2048 to 767. That affects every URL validated through `validate_url()` — gateway URLs, resource URIs, OAuth callback/redirect URLs with query params, icons, etc. URLs in the 768–2048 range that validate today will start failing. That's a silent breaking change for a much wider surface than tags, and it isn't called out in the "Breaking Change" notes.

**2. The 767 cap actually *creates* a new validation/DB mismatch.**
The stated rationale is "match the DB `String(767)` limit," but not all URL columns are 767:
- `db.py:5106` `webhook_url` → `String(2048)`
- `db.py:4818` `uaid` → `String(2048)`

So after this change, a 1500-char `webhook_url` passes the DB schema but is rejected by validation — the inverse of the bug you're fixing. A single global cap can't match columns that disagree (767 vs 2048). Suggest either dropping the URL change from this PR entirely (keep it tag-focused) or making it per-field rather than one global value.

**3. Test pollution — class attributes are mutated without restore.**
`test_custom_max_length_200`, `test_custom_max_length_100`, `test_custom_min_length` do `TagValidator.MAX_LENGTH = ...` / `MIN_LENGTH = ...` and never reset them. Since `TagValidator.MIN_LENGTH`/`MAX_LENGTH` are class attributes read once at import, these assignments leak across tests. If `test_custom_max_length_200` (sets MAX=200) runs before `test_validate_invalid_tags` (asserts `"x"*101` is invalid), that assertion flips and the suite goes order-dependent/flaky. Please restore via a fixture/`monkeypatch` with teardown, e.g.:
```python
@pytest.fixture(autouse=True)
def _restore_limits():
    mn, mx = TagValidator.MIN_LENGTH, TagValidator.MAX_LENGTH
    yield
    TagValidator.MIN_LENGTH, TagValidator.MAX_LENGTH = mn, mx
```
Also note the `@patch("...settings")` decorators do nothing useful here — `validate()` reads the class attrs, not `settings`, so the manual reassignment is what's actually being tested. The patch can be dropped.

## 🟡 Suggestions

**4. Drop the `os.getenv(...)` inside the `Field(default=...)`.**
```python
validation_min_tag_length: int = Field(default=int(os.getenv("VALIDATION_MIN_TAG_LENGTH", "2")), ...)
```
Pydantic `BaseSettings` already binds `VALIDATION_MIN_TAG_LENGTH` → `validation_min_tag_length`, so this is a redundant second source. It also reads the env at class-definition time and `int()` will raise a raw `ValueError` on a non-numeric value instead of a clean pydantic validation error. Just use `Field(default=2, ge=1, le=10)` and let pydantic handle the env.

**5. Docstring defaults are stale.** In `tags.py`, the class docstring and `Attributes:` still say *"Maximum length: ... (default: 50 characters)"* while the real default is now 100. Worth aligning so they don't mislead.

## 🔵 Minor

- `Issue #XXXX` placeholder is left in both `config.py` and the test docstrings — should be `#5175`.
- `test_tag_exceeds_max_length` uses `"a" * 256`, which exceeds the absolute `le=255` ceiling, so it passes regardless of the configured max — it doesn't really exercise the configurable limit.
- Doctests like `>>> TagValidator.validate("a")` → `False` assume `MIN_LENGTH=2`; they'll break if someone runs the suite with `VALIDATION_MIN_TAG_LENGTH=1`. Not blocking, just a latent coupling worth a comment.

Happy to re-review once the URL change is separated out (or made per-field) and the test teardown is in place. Nice work overall.

mcpgateway/config.py
```diff
@@ -3255,11 +3253,30 @@ def validate_database(self) -> None:
         ge=1,
         le=100,
     )
-    validation_max_url_length: int = 2048
+    validation_max_url_length: int = 767  # Match database String(767) limit to prevent validation/storage mismatch
```
cafalchio: Could you introduce a field here the same way you did for the tag?
I know that the PR is for tag, it is a non block comment. :)
`default=int(os.getenv("VALIDATION_MAX_URL_LENGTH", 767)),`

cafalchio (dismissed):
Thank you for the fix.
LGTM, just a non blocking comment.

bogdanmariusc10:
Thanks for the thorough review, @msureshkumar88! I have implemented your suggestions and fixed comments/docstrings. As for the out-of-scope `validation_max_url_length`, you were right, it might have introduced regressions, therefore I decided to remove the changes related to it.

msureshkumar88 (approved):
Thanks for addressing all the review feedback, Bogdan — this looks good now.

- URL length change (2048→767) reverted, keeping this PR scoped to tags as intended.
- `os.getenv()` redundancy removed from the `Field` defaults; pydantic handles env binding cleanly now.
- Test pollution fixed with the `_restore_limits` autouse fixture, and the dead `@patch` decorators are gone.
- Docstrings/comments updated to reflect the 100-char default and `#5175`.

The remaining minor notes (the `"a"*256` test case, doctest coupling to `MIN_LENGTH=2`) were explicitly called out as non-blocking and are fine to leave as follow-ups if anyone wants to polish later.

Approving — nice work tightening this up.

ja8zyjits (approved):
LGTM
3 Approval

PR #5184, opened 2026-06-11
feat(ui-rewrite): User edit 

gcgoncalves:
Closes https://github.com/IBM/mcp-context-forge/issues/4184

Screenshot 2026-06-11 at 12 50 53

Screenshot 2026-06-11 at 12 55 42


client/src/api/users.ts
```diff
@@ -41,6 +42,17 @@ function validateUserEmail(email: string): string {
...
+   * Update a user by email.
+   *
+   * @throws {ApiError} 403 - Insufficient permissions
+   * @throws {ApiError} 404 - User not found
+   */
+  update: (email: string, data: UpdateUserRequest): Promise<User> => {
+    const validEmail = validateUserEmail(email);
+    return api.patch<User>(`/auth/email/admin/users/${encodeURIComponent(validEmail)}`, data);
+  },
+
```
marekdano: I think we do not need this method. It's never called.

client/src/components/users/PasswordInput.tsx
```diff
@@ -0,0 +1,93 @@
...
+            <Eye className="h-4 w-4" aria-hidden="true" />
+          )}
+        </button>
+      </div>
+      {error ? (
+        <p
+          id={errorId}
+          className="text-sm text-red-600 dark:text-red-400"
+          role="alert"
+          aria-live="polite"
```
marekdano: `role="alert"` carries an implicit `aria-live="assertive"`. Explicitly overriding it with `aria-live="polite"` produces browser/AT-dependent behaviour — announcement timing is unpredictable across screen readers.
For validation errors, assertive is the right choice anyway. Fix: drop the explicit `aria-live` and let the role carry it.

client/src/components/users/PasswordInput.tsx
```diff
@@ -0,0 +1,93 @@
...
+        />
+        <button
+          type="button"
+          onClick={() => setShowPassword((v) => !v)}
+          className="absolute inset-y-0 right-0 flex items-center px-3 text-neutral-400 hover:text-neutral-600 dark:hover:text-neutral-300"
+          aria-label={intl.formatMessage({
+            id: showPassword ? "users.form.password.hide" : "users.form.password.show",
+          })}
+          aria-pressed={showPassword}
+          aria-live="polite"
```
marekdano: `aria-live` belongs on a container whose *content* changes dynamically, not on an interactive element. A button's accessible name is already communicated to screen readers when it changes, and `aria-pressed` handles the toggle state announcement. The combination causes double-announcement on some screen readers. Removing `aria-live` from the button is sufficient - `aria-label` + `aria-pressed` is the correct pattern here.

client/src/components/users/UserForm.tsx
```diff
@@ -5,19 +5,85 @@ import { Button } from "@/components/ui/button";
...
+          </>
+        )}
+      </label>
+      {children}
+      {error && (
+        <p
+          id={`${id}-error`}
+          className="text-sm text-red-600 dark:text-red-400"
+          role="alert"
+          aria-live="polite"
```
marekdano: `role="alert"` carries an implicit `aria-live="assertive"`. Explicitly overriding it with `aria-live="polite"` produces browser/AT-dependent behaviour — announcement timing is unpredictable across screen readers.
For validation errors, assertive is the right choice anyway. Fix: drop the explicit `aria-live` and let the role carry it.

client/src/components/users/UserForm.tsx
```diff
@@ -5,19 +5,85 @@ import { Button } from "@/components/ui/button";
...
+    <div className="space-y-1">
+      <label
+        htmlFor={id}
+        className="inline-flex items-center gap-0.5 text-sm font-medium text-neutral-900 dark:text-neutral-100"
+      >
+        {label}
+        {required && (
+          <>
+            <span className="text-red-500">*</span>
+            <span className="sr-only">(required)</span>
```
marekdano: Without `aria-hidden="true"` on the `*`, screen readers announce "asterisk" followed by "(required)" — redundant. Applies to `PasswordInput`, `FormField`, and the existing email label in `UserForm`.

client/src/components/users/PasswordInput.tsx
```diff
@@ -0,0 +1,93 @@
...
+    <div className="space-y-1">
+      <label
+        htmlFor={id}
+        className="inline-flex items-center gap-0.5 text-sm font-medium text-neutral-900 dark:text-neutral-100"
+      >
+        {label}
+        {required && (
+          <>
+            <span className="text-red-500">*</span>
+            <span className="sr-only">(required)</span>
```
marekdano: Without `aria-hidden="true"` on the `*`, screen readers announce "asterisk" followed by "(required)" — redundant. Applies to `PasswordInput`, `FormField`, and the existing email label in `UserForm`.

client/src/pages/Users.tsx
```diff
@@ -130,33 +141,39 @@ export function Users() {
...
+                is_active: userData.is_active ?? true,
+                auth_provider: "email",
+                created_at: new Date().toISOString(),
+                email_verified: false,
+                password_change_required: userData.password_change_required ?? false,
+                failed_login_attempts: 0,
+                is_locked: false,
+              };
+              setAllUsers((prev) => [optimisticUser, ...prev]);
+            }
```
marekdano, line 153: Is it possible to have this in a separate function.

marekdano (changes requested):
The PR looks good! I left some comments.

marekdano (approved):
LGTM 🚀

PR #3715, opened 2026-03-18
feat(auth): verify OAuth access tokens via JWKS for virtual server MCP

kimsehwan96:
## 🔗 Related Issue
Partially addresses #2659 

---

## 📝 Summary
Virtual servers with `oauth_enabled=True` advertise RFC 9728 Protected Resource Metadata, directing MCP clients to authenticate via an IdP (e.g., Keycloak, Authentik). However, when the client presents the IdP-issued access token, ContextForge rejects it with "Invalid token" because `verify_credentials()` only validates ContextForge-issued JWTs (`iss: "mcpgateway"`).

This PR adds IdP-issued OAuth access token verification via OIDC discovery + JWKS:

- **Issuer-based routing**: Peeks at the token's `iss` claim before verification. ContextForge-issued tokens (`iss == settings.jwt_issuer`) go through the existing path; IdP-issued tokens are routed to JWKS-based verification.
- **OIDC discovery + JWKS**: Fetches `/.well-known/openid-configuration` from the token's issuer, resolves `jwks_uri`, and verifies the signature using the IdP's public keys. Metadata is cached (300s TTL).
- **Per-server issuer allowlist**: Each virtual server's `oauth_config.authorization_servers` determines which issuers are trusted. Server A (Authentik) and Server B (Keycloak) are fully independent.
- **Audience validation**: When `oauth_config.client_id` is configured, the token's `aud` claim is validated. When not configured, audience check is skipped (opt-in).
- **User resolution**: Verified tokens are resolved against the ContextForge user database. Users must already exist (no auto-creation) — the expected flow is SSO login first, then MCP client access.

---

## 🏷️ Type of Change
- [ ] Bug fix
- [x] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Chore (deps, CI, tooling)
- [ ] Other (describe below)

---

## 🧪 Verification

| Check                     | Command         | Status |
|---------------------------|-----------------|--------|
| Lint suite                | `make lint`     |        |
| Unit tests                | `make test`     |        |
| Coverage ≥ 80%            | `make coverage` |        |

---

## ✅ Checklist
- [x] Code formatted (`make black isort pre-commit`)
- [x] Tests added/updated for changes
- [ ] Documentation updated (if applicable)
- [x] No secrets or credentials committed

---

## 📓 Notes (optional)

[...]

PR #5200, opened 2026-06-12
feat: support external OIDC bearer tokens on API and MCP endpoints

### LINKED ISSUE #3567: [FEATURE][AUTH]: Support external OIDC bearer tokens on API and MCP endpoints
### 📓 Additional Context

A client has been working with ContextForge MCP Hub and has successfully configured Keycloak as an SSO identity provider. They report that the browser-based SSO login flow works excellently — OIDC tokens are validated via JWKS, users are auto-provisioned, and realm roles are correctly mapped to ContextForge RBAC roles.

However, the client has encountered a limitation when attempting to extend this trust relationship to programmatic API access.

Use Case:
The client has AI agents (MCP clients) that need to authenticate against the MCP Hub endpoints (/mcp/, /tools, etc.) using machine-to-machine credentials. Their approach is to have these agents obtain an access token from the same OIDC identity provider that ContextForge already trusts for SSO (via the OAuth2 client_credentials grant) and present that token as a Bearer token on API requests.

Observed Behavior:
When an MCP client sends a Bearer token issued by the configured SSO IdP (Keycloak, but applicable to OAM, PING, etc.), the API authentication layer rejects it. It appears that the API/MCP endpoint authentication only validates internally-minted JWTs (signed with JWT_SECRET_KEY), rather than also accepting tokens from the trusted external IdP and validating them via the JWKS endpoint — even though the SSO flow already has the full OIDC/JWKS validation logic in place.

In other words, the trust relationship with the external IdP seems to be scoped exclusively to the browser-based SSO flow and does not extend to Bearer token authentication on API endpoints.

Expected Behavior:
Since ContextForge has a configured and verified trust relationship with the external IdP (JWKS URI, issuer validation, role mappings, etc.), the client expected that tokens issued by that same IdP would also be accepted as valid Bearer tokens on API and MCP endpoints — with the same role mapping and RBAC enforcement applied. This would enable a unified authentication model where both human users and service accounts authenticate through the same corporate identity provider.

Client Questions:

    Is this a known limitation, or is there a missing configuration option that enables external OIDC token validation on API endpoints?

    If this is not currently supported, is there a recommended approach for authenticating MCP clients/agents through a corporate OIDC provider without requiring them to have knowledge of ContextForge's internal JWT signing key?

    Is support for external IdP bearer tokens on API endpoints on the roadmap?

This capability is described as vital for enterprise deployments where centralized identity management is a requirement for both human users and AI agents/service integrations.

msureshkumar88:
## Summary

Closes #3567 — accept access tokens issued by an already-trusted external OIDC IdP (Keycloak, Microsoft Entra ID, generic OIDC) as `Bearer` credentials on API/MCP endpoints, validated via JWKS and mapped to the same RBAC identities as the browser SSO flow.

### Problem

ContextForge already trusts external IdPs for browser SSO (JWKS validation, JIT provisioning, role→team mapping), but that trust did not extend to programmatic access: an MCP client/agent presenting a `client_credentials` access token from the same IdP as a `Bearer` token was rejected, because `verify_credentials_cached` only ever validated internally-minted JWTs signed with `JWT_SECRET_KEY`.

### Approach

Single dispatch point in `verify_credentials_cached`:
- Decode the token's `iss` *without* verifying the signature, purely as a routing discriminator.
- If `iss` is internal (or the feature is disabled), fall through to the existing internal path unchanged — zero added cost.
- Otherwise, if `SSO_API_TOKEN_AUTH_ENABLED=true` and `iss` matches a `SSOProvider` with `trusted_for_api_auth=true`, validate via the existing `verify_oauth_access_token` (JWKS, SSRF defense, ID-token rejection, mandatory `api_audience` check), then JIT-provision + role-map via `build_external_identity`.

### Alternatives considered

| Approach | Summary | Verdict |
|---|---|---|
| **A. Fall-through in the existing bearer choke point** (chosen) | Branch on the token's unverified `iss` inside `verify_credentials_cached`: internal issuer → existing path unchanged; otherwise, if enabled, validate via `verify_oauth_access_token` against a trusted `SSOProvider` and map to a session identity. | **Chosen** — a single dispatch point automatically covers every endpoint that already depends on `require_auth` (API + MCP + admin), reuses 100% of the existing JWKS/SSRF/ID-token-rejection and SSO provisioning machinery, keeps default-deny, and internal tokens pay zero JWKS cost (the `iss` check short-circuits before any network call). |
| B. New dedicated auth dependency/middleware swapped across routers | A separate `require_external_or_internal_auth` dependency applied per-router. | Rejected — touches ~19 routers; easy to miss a path and create an inconsistent (potentially insecure) auth surface. |
| C. Edge middleware exchanges external token → mints a short-lived internal JWT | Validate at the edge, mint an internal JWT in place, leave the downstream stack untouched. | Rejected — adds per-request token minting/caching and extra crypto, muddies audit/revocation, overkill for the requirement. |

**Trust source:** trusted issuers are derived from existing `SSOProvider` rows rather than a new standalone issuer allowlist, because the role-mapping config (`team_mapping`, group claim names) a token needs already lives on the provider record — a separate allowlist would still have to point back at a provider, duplicating trust config. Reuse is explicit, not automatic: gated by both a global flag (`SSO_API_TOKEN_AUTH_ENABLED`, default off) and a per-provider opt-in (`trusted_for_api_auth`, default off), so a provider trusted for browser SSO is not silently trusted for API bearer auth.

### Identity mapping (security crux)

- External identities are synthesized with `token_use="session"` so they flow through the **DB-authoritative** `resolve_session_teams()` — never the claim-authority `normalize_token_teams()` path. This preserves the two-layer security model: Layer 1 (team visibility) is DB-derived, Layer 2 (RBAC) is unchanged.
- `is_admin` and team membership come from the persisted `EmailUser` / DB teams after provisioning — never from claims — so a token whose groups map to "admin" cannot grant admin unless the provisioned DB user is actually an admin.
- `client_credentials` tokens (no `email`/`email_verified`) get a synthetic, non-routable service-principal identity (`svc-<client_id>@<provider-id>.service.local`) provisioned through the same DB-authority path, without relaxing the human email-verified gate.

### Config & schema

- `SSO_API_TOKEN_AUTH_ENABLED` (global, default `false`) — preserves existing behavior unless explicitly enabled.
- `SSOProvider.trusted_for_api_auth` (per-provider opt-in, default `false`) and `SSOProvider.api_audience` — a provider trusted for browser SSO is *not* auto-trusted for API auth.
- `trusted_for_api_auth=true` requires a non-empty `api_audience` (enforced at provider create/update + startup check) to prevent confused-deputy/token-passing across a shared issuer.
- `EXTERNAL_IDENTITY_CACHE_TTL` (default 60s) bounds how long a provisioned external identity is cached per token, shared across workers via Redis when configured; invalidated immediately on provider create/update/delete.

### Performance & resilience

- Cached issuer→trusted-provider resolver (no table scan per request; untrusted issuers rejected with zero DB queries).
- Short-TTL, Redis-shared identity cache keyed by token hash (raw token never persisted); revocation/deactivation checks still run on every request outside the cache.
- Redis errors fail open (cache miss, auth unaffected); unexpected errors in the external path fail closed (401, never 500).
- Per-reason deny logging and a success-audit line (issuer, canonical email, provider id, `is_admin`, team count); all claim values sanitized for logs, token/payload never logged.
- Per-provider success/denied counters via `ObservabilityService.record_metric()` (best-effort, fail-open).

### Docs

New "machine-to-machine API auth" sections for Keycloak, Microsoft Entra ID, and generic OIDC, plus a troubleshooting checklist naming the deny-log strings, and security-model notes (AGENTS.md, rbac.md, multitenancy.md, oauth-design.md) documenting that external-IdP identities are session-semantics principals.

Admin UI exposure of the two new provider fields is deferred — the feature is fully configurable via `.env` + the existing provider CRUD API.

## Test Plan

- [ ] `make test` (unit suite)
- [ ] `tests/unit/mcpgateway/utils/test_external_idp_auth.py` — deny-path regression suite (untrusted issuer, opted-out/disabled provider, expired/tampered/ID token, missing/wrong audience, unprovisionable user, C1 routing guard, C2 escalation guard, H1 service-principal, cache/logging/exception behavior)
- [ ] `tests/integration/test_external_idp_auth_integration.py --with-integration` — un-mocked JWKS signature/audience validation, real `resolve_session_teams` team resolution, end-to-end `require_auth` + revocation

### Manual testing (Keycloak)

1. Register a Keycloak SSO provider in ContextForge as usual (browser SSO working).
2. In Keycloak: create a confidential client (e.g. `mcp-agent`) with **Service accounts roles** enabled, add an **Audience** mapper on its dedicated client scope set to a custom audience (e.g. `mcp-gateway`), and assign realm/client roles to the service account so they map to a ContextForge team via `SSO_KEYCLOAK_ROLE_MAPPINGS`.
3. Enable the global switch and opt the provider in:
   ```bash
   # .env
   SSO_API_TOKEN_AUTH_ENABLED=true

   curl -X PUT https://gateway.example.com/auth/sso/admin/providers/keycloak \
     -H "Authorization: Bearer $ADMIN_TOKEN" \
     -d '{"trusted_for_api_auth": true, "api_audience": "mcp-gateway"}'
   ```
4. Fetch a `client_credentials` token and call the API directly with it:
   ```bash
   TOKEN=$(curl -s -X POST \
     https://keycloak.example.com/realms/master/protocol/openid-connect/token \
     -d "grant_type=client_credentials" \
     -d "client_id=mcp-agent" \
     -d "client_secret=$MCP_AGENT_CLIENT_SECRET" \
     | jq -r '.access_token')

   curl -H "Authorization: Bearer $TOKEN" https://gateway.example.com/tools
   ```
5. Verify: request succeeds (200), and the resolved identity is the synthetic service-principal `svc-mcp-agent@keycloak.service.local` with teams matching the mapped roles/groups (check via `/whoami` or audit log success line).
6. Deny paths:
   - Set `trusted_for_api_auth=false` (or `SSO_API_TOKEN_AUTH_ENABLED=false`) and re-call → `401`.
   - Call with a token whose `aud` doesn't match `api_audience` → `401`, deny-reason logged.
   - Deactivate the provisioned service-principal user in ContextForge and re-call (within cache TTL) → `401` (revocation check runs outside the identity cache).

### DIFF
--- a/mcpgateway/utils/verify_credentials.py
+++ b/mcpgateway/utils/verify_credentials.py
@@ -514,7 +514,11 @@ async def _external_identity_cache_get(token_hash: str) -> Optional[dict]:
     """
     redis = await get_redis_client()
     if redis is not None:
-        raw = await redis.get(_EXTERNAL_IDENTITY_REDIS_PREFIX + token_hash)
+        try:
+            raw = await redis.get(_EXTERNAL_IDENTITY_REDIS_PREFIX + token_hash)
+        except Exception as exc:  # noqa: BLE001 - best-effort cache, never fail auth on Redis errors
+            logger.warning("External identity cache Redis GET failed, treating as cache miss: %s", exc)
+            return None
         if raw is None:
             return None
         try:
@@ -528,7 +532,7 @@ async def _external_identity_cache_get(token_hash: str) -> Optional[dict]:
     if monotonic() >= expires_at:
         _external_identity_cache.pop(token_hash, None)
         return None
-    return payload
+    return dict(payload)
 
 
 async def _external_identity_cache_put(token_hash: str, payload: dict, token_exp: Optional[int]) -> None:
@@ -546,7 +550,10 @@ async def _external_identity_cache_put(token_hash: str, payload: dict, token_exp
 
     redis = await get_redis_client()
     if redis is not None:
-        await redis.set(_EXTERNAL_IDENTITY_REDIS_PREFIX + token_hash, json.dumps(safe_payload), ex=ttl)
+        try:
+            await redis.set(_EXTERNAL_IDENTITY_REDIS_PREFIX + token_hash, json.dumps(safe_payload), ex=ttl)
+        except Exception as exc:  # noqa: BLE001 - best-effort cache write, never fail auth on Redis errors
+            logger.warning("External identity cache Redis SET failed, skipping cache write: %s", exc)
         return
     if len(_external_identity_cache) >= _EXTERNAL_IDENTITY_MAX:
         _external_identity_cache.clear()
--- a/mcpgateway/utils/verify_credentials.py
+++ b/mcpgateway/utils/verify_credentials.py
@@ -639,7 +639,7 @@ async def _maybe_verify_external(token: str, request: Optional[Request]) -> Opti
         token's issuer matches a trusted external provider and verification succeeds,
         or None to fall through to internal JWT verification.
     """
-    if not settings.sso_api_token_auth_enabled:
+    if not getattr(settings, "sso_api_token_auth_enabled", False):
         return None
 
     # P4: fetch the request-scoped DB session once, up front, so it can be reused
--- a/mcpgateway/utils/verify_credentials.py
+++ b/mcpgateway/utils/verify_credentials.py
@@ -1810,3 +1810,89 @@ async def verify_external_idp_token(token: str, db: Session) -> tuple[Optional[d
     if claims is None:
         return None, None
     return claims, provider
+
+
+def _get_sso_service(db: Session):
+    """Factory wrapper so tests can patch SSOService construction.
+
+    Args:
+        db: Request-scoped SQLAlchemy session.
+
+    Returns:
+        A new :class:`~mcpgateway.services.sso_service.SSOService` instance.
+    """
+    # First-Party
+    from mcpgateway.services.sso_service import SSOService  # pylint: disable=import-outside-toplevel
+
+    return SSOService(db)
+
+
+async def build_external_identity(provider: SSOProvider, verified_claims: dict, token: str, db: Session) -> Optional[dict[str, Any]]:
+    """Map verified external-IdP claims to a ContextForge identity payload.
+
+    Reuses browser-SSO normalization + provisioning so role/group -> team mapping
+    is identical. Team scoping uses SESSION semantics (DB authority) because an
+    external access token carries no internal ``teams`` claim.
+
+    SECURITY:
+        * ``token_use="session"`` so downstream dispatchers route via
+          ``resolve_session_teams`` (DB authority), NOT ``normalize_token_teams``
+          (claim authority).
+        * ``is_admin`` is read from the persisted ``EmailUser``, never the mapped
+          claim.
+        * ``authenticate_or_create_user`` returns a JWT token, not an identity --
+          used only for its provisioning side-effect; identity comes from
+          ``get_user_by_email``.
+
+    Args:
+        provider: The matched, trusted SSOProvider (from resolve_trusted_provider_by_issuer).
+        verified_claims: Signature-verified claims from verify_external_idp_token.
+        token: The raw external bearer token (carried through for downstream use).
+        db: Request-scoped SQLAlchemy session.
+
+    Returns:
+        The enriched session-semantics identity payload, or None when the user
+        cannot be provisioned/resolved.
+    """
+    # First-Party
+    from mcpgateway.auth import resolve_session_teams  # pylint: disable=import-outside-toplevel
+
+    svc = _get_sso_service(db)
+    user_info = svc._normalize_user_info(provider, verified_claims)  # pylint: disable=protected-access
+
+    # Provisioning side-effect only (creates/links user, role-sync, enforces
+    # trusted_domains + email-verified). Return value is a JWT token -- discard it.
+    provisioned = await svc.authenticate_or_create_user(user_info)
+    if not provisioned:
+        return None
+
+    email = str(user_info.get("email") or "").strip().lower()
+    if not email:
+        return None
+
+    db_user = await svc.auth_service.get_user_by_email(email)
+    if db_user is None:
+        return None
+
+    is_admin = bool(db_user.is_admin)  # DB authority, not the mapped claim
+
+    # token_use="session" payload routed via resolve_session_teams(payload, email, db_user).
+    payload: dict = {
+        "sub": email,
+        "email": email,
+        "token": token,
+        "token_use": "session",  # nosec B105 - JWT claim type, not a password
+        "source": "external_idp",  # audit/telemetry only -- never drives authz
+        "iss": verified_claims.get("iss"),
+        "auth_provider": getattr(provider, "id", None),
+    }
+    teams = await resolve_session_teams(payload, email, db_user)
+    payload["teams"] = teams
+    payload["is_admin"] = is_admin
+
+    # Persist provisioning if we own the session (set by a future dispatcher).
+    own_session = getattr(getattr(db, "info", {}), "get", lambda *_: None)("external_owned")
+    if own_session:
+        db.commit()
+
+    return payload
--- a/mcpgateway/utils/verify_credentials.py
+++ b/mcpgateway/utils/verify_credentials.py
@@ -57,6 +57,7 @@ from base64 import b64decode
 import binascii
 import hashlib
 import json
+import re
 import time
 from time import monotonic
 from typing import Any, Optional, Union
@@ -2095,6 +2096,9 @@ def _synthetic_service_principal_user_info(provider: SSOProvider, claims: dict) 
         carrying through any group/role claims so role/team mapping still applies.
     """
     client = str(claims.get("azp") or claims.get("client_id") or claims.get("sub"))
+    # IdP-controlled value: strip anything that could break out of the email
+    # local-part (e.g. an embedded "@") and produce a malformed/spoofed address.
+    client = re.sub(r"[^a-zA-Z0-9._-]", "_", client)
     pid = getattr(provider, "id", "ext")
     return {
         "email": f"svc-{client}@{pid}.service.local",
--- a/mcpgateway/utils/verify_credentials.py
+++ b/mcpgateway/utils/verify_credentials.py
@@ -681,7 +681,7 @@ async def _maybe_verify_external(token: str, request: Optional[Request]) -> Opti
             await _external_identity_cache_put(th, payload, claims.get("exp"))
         return payload
     except Exception as exc:  # pylint: disable=broad-except
-        logger.warning("external-idp auth error, falling through to internal (401): %s", exc)
+        logger.warning("external-idp auth error, falling through to internal (401): %s", sanitize_for_log(str(exc)))
         return None
     finally:
         if own_session:
@@ -2168,6 +2168,11 @@ async def build_external_identity(provider: SSOProvider, verified_claims: dict, 
     except IntegrityError:
         # E2: lost a concurrent provisioning race -- the winner already created
         # the user, so roll back our half-applied insert and re-look-up by email.
+        # Rollback is required (not just for owned sessions): SQLAlchemy leaves the
+        # session in a failed-transaction state after IntegrityError, and the
+        # re-lookup below would fail without it. Safe even for a request-scoped
+        # session: external-IdP auth dispatch runs during request authentication,
+        # before route handlers stage any writes on that session.
         db.rollback()
         provisioned = True
         logger.info("external-idp: provisioning race resolved via re-lookup")
@@ -2211,7 +2216,7 @@ async def build_external_identity(provider: SSOProvider, verified_claims: dict, 
             db.commit()
         except Exception as exc:  # pylint: disable=broad-except
             db.rollback()
-            logger.error("external-idp: provisioning commit failed: %s", exc)
+            logger.error("external-idp: provisioning commit failed: %s", sanitize_for_log(str(exc)))
             return None
 
     if is_synthetic:
--- a/mcpgateway/utils/verify_credentials.py
+++ b/mcpgateway/utils/verify_credentials.py
@@ -65,10 +65,12 @@ from fastapi import Cookie, Depends, HTTPException, Request, status
 from fastapi.security import HTTPAuthorizationCredentials, HTTPBasic, HTTPBasicCredentials, HTTPBearer
 from fastapi.security.utils import get_authorization_scheme_param
 import jwt
+from sqlalchemy.orm import Session
 
 # First-Party
 from mcpgateway.config import settings
 from mcpgateway.services.logging_service import LoggingService
+from mcpgateway.services.sso_service import resolve_trusted_provider_by_issuer
 from mcpgateway.utils.jwt_config_helper import validate_jwt_algo_and_keys
 from mcpgateway.utils.log_sanitizer import sanitize_for_log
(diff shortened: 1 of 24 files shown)


mcpgateway/utils/verify_credentials.py
```diff
@@ -1770,3 +2014,228 @@ async def verify_oauth_access_token(
...
+        claims: Signature-verified JWT claims from the external IdP.
+
+    Returns:
+        True if the token represents a service/client principal rather than
+        a human user, False otherwise.
+    """
+    if claims.get("email"):
+        return False
+    sub = claims.get("sub")
+    return bool(sub) and (sub == claims.get("azp") or sub == claims.get("client_id") or claims.get("typ") == "at+jwt")
```
gandhipratik203: Worth double-checking the Keycloak service-account flow against a real token. `_is_clientless_token` treats a token as a service principal when `sub == azp` (or `typ == "at+jwt"`). Real Keycloak `client_credentials` tokens seem to differ here: `sub` is the service-account UUID while `azp` is the client name, and `typ` is usually `"Bearer"` rather than `"at+jwt"`. If that holds, a standard Keycloak token without an email mapper would fall through to the human path and get denied on the empty email, which is the same flow described in the "Manual testing (Keycloak)" section. The tests pass because the fixtures use `sub == azp`, so they wouldn't surface it. Could you sanity-check this against an actual Keycloak token? Totally possible newer versions shape it differently.

mcpgateway/utils/verify_credentials.py
```diff
@@ -1770,3 +2014,228 @@ async def verify_oauth_access_token(
...
+    provider = resolve_trusted_provider_by_issuer(issuer, db)
+    if provider is None or not provider.issuer:
+        logger.warning("external-idp auth denied: issuer not trusted (iss=%s)", sanitize_for_log(issuer))
+        _record_external_auth_metric("denied", getattr(provider, "id", None), reason="issuer_not_trusted")
+        return None, None
+
+    claims = await verify_oauth_access_token(
+        token,
+        authorization_servers=[provider.issuer],
+        expected_audience=provider.api_audience or None,
```
gandhipratik203: A small fail-closed backstop on audience might be worth adding. The confused-deputy protection relies on `api_audience` being non-empty (since `verify_oauth_access_token` only checks `aud` when it's set), and right now that invariant is enforced in the Pydantic validators while the runtime trust query only checks `trusted_for_api_auth`. So a provider row that ends up with `trusted_for_api_auth=True` and a null audience through some other path (direct DB, a future importer, env seeding) would quietly skip the `aud` check. Bootstrap logs an error but doesn't block it. A one-line guard in `verify_external_idp_token` would make the property hold regardless of how the row got there.

Lang-Akshay (changes requested):
Thanks for the PR - @msureshkumar88 
Please address this finding 

Security architecture is correct — double opt-in, DB-authoritative `is_admin`, proper `resolve_session_teams()` routing. Two medium findings block merge:

1. Unsanitized `client_id` in synthetic email (`verify_credentials.py` ~2097) — IdP-controlled `client_id` with `@` produces malformed email. Fix: `re.sub(r'[^a-zA-Z0-9._-]', '_', client)`.
2. Null `api_audience` log-only (`sso_bootstrap.py` ~249) — `trusted_for_api_auth=True` + no audience = any-audience token accepted silently. Startup check must be hard-blocking, not just a log warn.
3. Two test gaps — no deny-path test for Layer 2 RBAC rejection of external IdP sessions; no test for null `api_audience` acceptance behavior.

msureshkumar88:
Thanks @Lang-Akshay — addressed all three in 8dbba25:

1. **Synthetic email** (`verify_credentials.py`): `client_id`/`azp`/`sub` now sanitized via `re.sub(r'[^a-zA-Z0-9._-]', '_', client)` before going into the local-part, so an IdP-controlled value can no longer inject a second `@` or otherwise forge the domain.
2. **Null `api_audience`**: went a bit further than the suggested log-fix, since the literal `sso_bootstrap.py` log was actually a secondary guard — the real gap was that the create/update validators in `routers/sso.py` only check the incoming *payload*, not the resulting DB state. A `PATCH` with just `{"api_audience": ""}` (no `trusted_for_api_auth` in the same call) slipped past, since `model_dump()` filters `None` but not `""`. Fixed by checking the resulting state in `SSOService.update_provider`/`create_provider` (raises `ValueError` → 400, rolls back). Also made the startup check in `sso_bootstrap.py` fail-closed: it now force-disables `trusted_for_api_auth` on any provider found in that state (e.g. from a direct DB edit or pre-invariant row) instead of only logging.
3. **Test gaps**: added a Layer 2 RBAC deny-path test for an external-IdP session payload, a test closing the partial-update bypass, a bootstrap fail-closed test, and a synthetic-email sanitization test.

Full unit + integration suites (`--with-integration`) green, black/isort clean.

msureshkumar88:
Thanks @gandhipratik203 — both addressed, see latest commit.

1. **Keycloak service-account shape**: confirmed — real Keycloak \`client_credentials\` tokens have \`sub\` as the service-account user's UUID (≠ \`azp\`) and \`typ: "Bearer"\`, so neither generic heuristic matched. \`_is_clientless_token\` now also checks Keycloak's \`clientId\` claim (camelCase, only present on service-account tokens) and a \`preferred_username\` of the form \`service-account-<client>\`. Added a test asserting the real Keycloak shape is now detected, plus that a human token with an unrelated \`preferred_username\` still isn't misclassified.
2. **Fail-closed audience backstop**: added the one-line guard you suggested in \`verify_external_idp_token\` — if a resolved provider has \`trusted_for_api_auth\` set but empty/null \`api_audience\`, it's denied before \`verify_oauth_access_token\` is ever called, regardless of how that row got into that state. Test added (asserts the token verifier is never invoked).

Combined with the create/update + bootstrap fixes from the earlier round, the invariant is now enforced at three independent points: write-time, boot-time, and verify-time.

Lang-Akshay (changes requested):
Please address the blocking issue:
### Blocking: 
verify_credentials.py:1989 — JWKS fetch via PyJWKClient follows HTTP redirects (urllib default), bypasses same-origin jwks_uri check. Pre-existing bug (#3715), but this PR
  makes it reachable from every bearer-auth endpoint, not just opted-in OAuth virtual servers. Fix: use httpx client w/ follow_redirects=False, same as OIDC-metadata discovery path
  already does.

  ### Non-blocking, should fix or track:
  - verify_credentials.py:2074 _is_clientless_token — sub==azp heuristic can misclassify human auth-code tokens as service principals, breaks audit attribution to real identity.
  - db.py:5671 — stale comment says "null = skip aud check," code actually fail-closed rejects empty audience.
  - verify_credentials.py:735 — 60s cache window can serve pre-demotion admin/team privileges after DB-side change.
  - sso_service.py:2519 — provider w/ issuer == internal jwt_issuer silently inert (dead config trap, not exploit).

gandhipratik203:
LGTM, one the other blocking comment is resolved, we are good to go!

msureshkumar88:
Addressed all findings in 7e2bae988.

**Blocking — JWKS redirect bypass (`verify_credentials.py`):**
Introduced `_NoRedirectPyJWKClient`, a one-method `PyJWKClient` subclass that overrides `fetch_data()` with `httpx.Client(follow_redirects=False)` instead of the default `urllib.request.urlopen`. The same-origin check on `jwks_uri` now holds through the full HTTP fetch — a redirect to a different origin raises `httpx.TooManyRedirects` before any key material is returned. The JWKS cache type annotation updated to match.

**Non-blocking — `_is_clientless_token` audit attribution (`verify_credentials.py`):**
Added `given_name`, `family_name`, and `name` as human-identity guards alongside the existing `email` check. All four are evaluated before the `sub==azp` branch, so an auth-code token carrying any standard OIDC personal-name claim is never misclassified as a service principal.

**Non-blocking — stale comment (`db.py`):**
`api_audience` field comment updated from `"null = skip aud check"` to `"must be non-empty when trusted_for_api_auth is set"`, matching the actual invariant enforced at write-time, boot-time, and verify-time.

**Non-blocking — dead config trap (`sso_service.py`):**
`_load_trusted_provider_map` now emits a `WARNING ... (dead config)` when a provider's issuer equals `jwt_issuer`. Those providers receive no tokens because the internal-issuer fast-path intercepts before the external-IdP path runs.

3 new unit tests added (redirect-bypass, human-claim precedence, dead-config warning). 69 tests pass, ruff clean.

Lang-Akshay (changes requested):
## Finding 1: SSRF Bypass in Browser-SSO JWKS Path (HIGH SEVERITY)

**Location:** `mcpgateway/services/sso_service.py:259`

**Issue:** The browser-based SSO flow uses the vulnerable `jwt.PyJWKClient` that follows HTTP redirects:

```python
self._jwks_client_cache[jwks_uri] = jwt.PyJWKClient(jwks_uri)
```

**Evidence:**

- Line 259 instantiates standard `jwt.PyJWKClient` (follows redirects)
- The secure `_NoRedirectPyJWKClient` exists (`verify_credentials.py:1756`) but is only used for API token verification (line 2009)
- Browser SSO path (`_verify_oidc_id_token` method) bypasses SSRF protection

**Impact:** Attackers can exploit redirect-based SSRF via malicious JWKS URIs in browser SSO flows, bypassing the same-origin checks.

**CWE:** CWE-601 (URL Redirection to Untrusted Site)

---

## Finding 2: Integration Tests Mock Wrong Class (TEST COVERAGE GAP)

**Location:** `tests/integration/test_external_idp_auth_integration.py` (G1/G6 tests)

**Issue:** Tests claim to validate "real signature verification" but mock the wrong `PyJWKClient` class:

```python
monkeypatch.setattr(vc.jwt, "PyJWKClient", FakeJWKClient)  # Lines 73, 133
```

**Evidence:**

- Tests mock `jwt.PyJWKClient`
- Production code uses `_NoRedirectPyJWKClient` (`verify_credentials.py:2009`)
- Tests pass via httpx network failures, not via actual SSRF protection validation

**Impact:** The SSRF protection mechanism (`_NoRedirectPyJWKClient`) is never exercised in integration tests. Security regression could occur undetected.

---

## Recommended Actions

1. **Fix `sso_service.py`:** Replace `jwt.PyJWKClient` with `_NoRedirectPyJWKClient` at line 259
2. **Fix integration tests:** Mock `vc._NoRedirectPyJWKClient` instead of `vc.jwt.PyJWKClient`
3. **Add regression test:** Verify redirect-following is blocked in both API and browser-SSO paths

msureshkumar88:
Thanks for the thorough follow-up review @Lang-Akshay. Both findings are addressed in the latest commit.

**Finding 1 — SSRF in browser-SSO JWKS path (fixed)**

`sso_service._get_jwks_client` now uses `_NoRedirectPyJWKClient` instead of `jwt.PyJWKClient`. Because `verify_credentials` imports from `sso_service` (circular import), the class is defined locally in `sso_service.py` rather than shared — same `fetch_data()` override using httpx with `follow_redirects=False`. Both the API bearer path and the browser SSO OIDC ID-token path now reject redirect-based JWKS URI hijacking.

**Finding 2 — Integration test mocks targeting wrong class (fixed)**

All three `monkeypatch.setattr(vc.jwt, "PyJWKClient", ...)` calls in `test_external_idp_auth_integration.py` (lines 81, 109, 152) are changed to `monkeypatch.setattr(vc, "_NoRedirectPyJWKClient", ...)`. The mocks now intercept the class that production code actually instantiates, so the SSRF protection mechanism is correctly exercised.

Lang-Akshay (approved):
thanks for the PR - ✅Approved

gandhipratik203 (commented):
LGTM!

PR #5203, opened 2026-06-12
fix(api): Login redirect loop (#5190)

### LINKED ISSUE #5190: [BUG]: Login redirect loop
### 🐞 Bug Summary
When the `MCP_CLIENT_AUTH_ENABLED` flag is set to `false`, logging into the admin UI goes into an infinite redirect loop between `/admin` and `/admin/login`.

---

### 🧩 Affected Component

- [ ] `mcpgateway` - API
- [X] `mcpgateway` - UI (admin panel)
- [ ] Other (explain below)

---

### 🔁 Steps to Reproduce

1. Set `MCP_CLIENT_AUTH_ENABLED` to `false`
2. Access the admin UI
3. Open the browser network tab
4. Login
5. Watch the browser network tab

---

### 🤔 Expected Behavior
The user gets loggged in.

gcgoncalves:
# 🐛 Bug-fix PR

Closes #5190 

---

## 📌 Summary
When the `MCP_CLIENT_AUTH_ENABLED` flag is set to `false`, logging into the admin UI goes into an infinite redirect loop between `/admin` and `/admin/login`.

## 🔁 Reproduction Steps
1. Set `MCP_CLIENT_AUTH_ENABLED` to `false`
2. Access the admin UI
3. Open the browser network tab
4. Login
5. Watch the browser network tab

## 🐞 Root Cause
The loop is:

1. GET `/admin`: route dependency `get_current_user_with_permissions` enters `mcp_client_auth_enabled=False` block, hits `is_proxy_auth_trust_active()=False` → redirects browser to `/admin/login`. The JWT cookie is present but the function returns early before ever reaching the JWT-reading code.

2. GET `/admin/login`: `admin_login_page` reads the JWT cookie, validates it, sees `is_admin=True` → redirects to `/admin`

## 💡 Fix Description
Before entering the `mcp_client_auth_enabled=False` block, check if a JWT token is already present. If it is, skip the block and fall through to standard JWT validation.

## 🧪 Verification

| Lint suite                            | `make lint`          |    ✅    |
| Unit tests                            | `make test`          |    ✅    |
| Manual regression no longer fails     | steps / screenshots  |    ✅    |

## ✅ Checklist
- [X] No secrets/credentials committed

### DIFF
--- a/.github/workflows/docker-scan.yml
+++ b/.github/workflows/docker-scan.yml
@@ -242,5 +242,3 @@ jobs:
           else
             exit 1
           fi
-
-
--- a/mcpgateway/middleware/rbac.py
+++ b/mcpgateway/middleware/rbac.py
@@ -183,8 +183,11 @@ async def get_current_user_with_permissions(request: Request, credentials: Optio
         elif is_admin:
             set_trace_team_scope("admin")
 
-    # Check for proxy authentication first (if MCP client auth is disabled)
-    if not settings.mcp_client_auth_enabled:
+    # Check for proxy authentication first (if MCP client auth is disabled).
+    # Cookie JWTs from the admin UI must still be honoured — skipping them causes
+    # an infinite redirect loop between /admin and /admin/login.
+    _has_cookie_jwt = bool(request.cookies.get("jwt_token") or request.cookies.get("access_token") or jwt_token)  # Cookie() dependency value
+    if not settings.mcp_client_auth_enabled and not _has_cookie_jwt:
         # Read plugin context from request.state for cross-hook context sharing
         # (set by HttpAuthMiddleware for passing contexts between different hook types)
         plugin_context_table = getattr(request.state, "plugin_context_table", None)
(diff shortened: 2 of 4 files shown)


vishu-bh (changes requested):
When `MCP_CLIENT_AUTH_ENABLED=false`, any `jwt_token` / `access_token` cookie now skips the proxy-auth branch and falls through to JWT validation. With `TRUST_PROXY_AUTH=true`, that means a valid UI cookie can authenticate even when the trusted proxy header is missing, and a stale cookie can break otherwise-valid proxy auth.

Could we scope this fix to the non-proxy-trust case, or preserve proxy-header precedence when `is_proxy_auth_trust_active(settings)` is true? Please add regression coverage for proxy auth + cookie combinations.

Also It would be worth adding regression coverage for:

- proxy auth enabled + valid proxy header + stale/invalid UI cookie should still use proxy identity
- proxy auth enabled + missing proxy header + valid UI cookie should not authenticate through the cookie path

vishu-bh (approved):
Thanks @gcgoncalves!

Issue addressed correctly, regression coverage is present for the redirect loop and proxy/cookie precedence

LGTM 🚗

ja8zyjits (approved):
LGTM Process followed

PR #5204, opened 2026-06-12
fix(gateway): propagate connection errors on gateway update

### LINKED ISSUE #5188: [BUG]:  Invalid gateway URL returns 200 and marks gateway as reachable
### 🐞 Bug Summary
Updating a gateway with an invalid URL returns HTTP 200 and marks the gateway as successfully updated, even though gateway initialization fails internally.

---

### 🧩 Affected Component

- [x] `mcpgateway` - API
- [x] Other (explain below)

---

### 🔁 Steps to Reproduce

1. Create new gateway
2. Update  gateway with PUT /gateways/{id} API
3. Provide an invalid URL (example with invalid IPv4 format - or URL to MCP that is not reachable) 
4. Gateway initialisation is failing but response is still returned as 200

---

### 🤔 Expected Behavior

- Request should fail (e.g. HTTP 400 or 500) when gateway initialization fails.
- Gateway should not be marked as successfully updated.
- reachable should be false.
- Error should be propagated to the API response.

 ### Actual Behavior

- API returns 200 OK
- Gateway is marked as successfully updated
- reachable is set to true
- Internal logs show initialization failure

---

### 📓 Logs / Error Output

Failed to initialize updated gateway: Failed to initialize gateway at http.... 

```
{"asctime": "2026-06-11T06:56:28", "name": "mcpgateway.services.gateway_service", "levelname": "WARNING", "message": "Failed to initialize updated gateway: Failed to initialize gateway at http://xxxxxxx:7778/sse: Invalid IPv4 address: '9.60.134.00'", "@timestamp": "2026-06-11T05:56:28.545237Z", "hostname": "xxxxxxx", "process_id": 24528, "request_id": "xxxxxxxxxxx"}
{"asctime": "2026-06-11T06:56:28", "name": "mcpgateway.cache.registry_cache", "levelname": "DEBUG", "message": "RegistryCache: Invalidating gateways cache", "@timestamp": "2026-06-11T05:56:28.549972Z", "hostname": "xxxxx", "process_id": 24528, "request_id": "xxxxxxxxxxx"}
```
Failed to initialize updated gateway: Failed to initialize gateway at http

```
{"asctime": "2026-06-11T06:56:28", "name": "mcpgateway.services.structured_logger", "levelname": "INFO", "message": "[gateway_service] Gateway updated successfully", "component": "gateway_service", "category": null, "user_id": "pawelkai@ie.ibm.com", "user_email": "xxxxxxxx", "team_id": null, "duration_ms": null, "custom_fields": {"gateway_name": "pawel-before-update", "version": 2}, "tags": null, "event_type": "gateway_updated", "resource_type": "gateway", "resource_id": "xxxxxx", "correlation_id": "05ffc6a69eaf4ab18bb223a1f18b2405", "hostname": "xxxxxxx", "process_id": 24528, "timestamp": "2026-06-11T05:56:28.551071+00:00", "@timestamp": "2026-06-11T05:56:28.551078Z", "request_id": "05ffc6a69eaf4ab18bb223a1f18b2405"}
{"asctime": "2026-06-11T06:56:28", "name": "mcpgateway.services.structured_logger", "levelname": "INFO", "message": "[http_gateway] Request completed: PUT /gateways/d7b5c0584b7e41a7b4879a3fee38be83 - 200", "component": "http_gateway", "category": null, "user_id": null, "user_email": null, "team_id": null, "duration_ms": 44.69799995422363, "custom_fields": null, "tags": null, "correlation_id": "05ffc6a69eaf4ab18bb223a1f18b2405", "operation_type": "http_request", "request_method": "PUT", "request_path": "/gateways/d7b5c0584b7e41a7b4879a3fee38be83", "response_status_code": 200, "user_agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/149.0.0.0 Safari/537.36", "client_ip": "127.0.0.1", "metadata": {"event": "request_completed", "response_time_category": "fast"}, "hostname": "Mac.vodafone.ultrahub", "process_id": 24528, "timestamp": "2026-06-11T05:56:28.551806+00:00", "@timestamp": "2026-06-11T05:56:28.551815Z", "request_id": "05ffc6a69eaf4ab18bb223a1f18b2405"}
{"asctime": "2026-06-11T06:56:28", "name": "uvicorn.access", "levelname": "INFO", "message": "127.0.0.1:58438 - \"PUT /gateways/d7b5c0584b7e41a7b4879a3fee38be83 HTTP/1.1\" 200", "@timestamp": "2026-06-11T05:56:28.552124Z", "hostname": "xxxxxxx", "process_id": 24528}
```

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|

---

msureshkumar88:
## Summary
- `PUT /gateways/{id}` previously swallowed `_initialize_gateway` failures during update and committed the new (broken) config anyway, returning 200 for an unreachable/invalid gateway URL — unlike `POST /gateways`, which correctly returns 502.
- Now, when a connection-affecting field (`url`, `transport`, `auth_type`/`auth_value`/`auth_query_params`, `oauth_config`, `ca_certificate`, `client_cert`/`client_key`) changes and re-initialization fails, the error propagates as `GatewayConnectionError`, the transaction rolls back, and the API returns 502 — matching `POST` behavior.
- Cosmetic-only updates (name/description/tags/visibility) on an already-unreachable gateway still commit, preserving the existing best-effort behavior (issue #4205 / visibility propagation).
- <Extended by @ja8zyjits >
- When `GATEWAY_ASYNC_LIFECYCLE_ENABLED=true`, update failures remain deferred to the async lifecycle worker: `PUT` returns `202 pending`, persists the new config, and retry metadata is recorded by the worker.
 

Closes #5188

## Test plan
- [x] `tests/unit/mcpgateway/services/test_gateway_service.py::test_update_gateway_url_initialization_failure` updated to assert `GatewayConnectionError` propagates and commit is not called when a connection-affecting field changes and re-init fails.
- [x] `tests/unit/mcpgateway/services/test_gateway_service.py::test_update_gateway_visibility_propagates_when_init_fails` (regression guard) confirms cosmetic-only updates still commit when init fails.
- [x] `tests/unit/mcpgateway/test_main_error_handlers.py::test_update_gateway_connection_error` confirms API-level 502 mapping.
- [x] Full gateway service + main error handler suites pass (354 passed).
- <Extended by @ja8zyjits >
- [x] Added SQLite async lifecycle regression coverage for bad URL update: `PUT` returns `202 pending`, persists the new URL, and worker failure records retry metadata instead of returning route-level `502`.
- [x]  `pytest tests/e2e/test_gateway_async_lifecycle.py -v`
- [x] `pytest tests/unit/mcpgateway/services/test_gateway_service.py -v -k "async_lifecycle or url_initialization_failure"`

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-07-02T15:09:47Z",
+  "generated_at": "2026-07-02T15:27:20Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -274,7 +274,7 @@
         "hashed_secret": "b4673e578b9b30fe8bba1b555b7b59883444c697",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1604,
+        "line_number": 1599,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -282,7 +282,7 @@
         "hashed_secret": "4a0a2df96d4c9a13a282268cab33ac4b8cbb2c72",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1692,
+        "line_number": 1687,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -290,7 +290,7 @@
         "hashed_secret": "5baa61e4c9b93f3f0682250b6cf8331b7ee68fd8",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 2042,
+        "line_number": 2037,
         "type": "Basic Auth Credentials",
         "verified_result": null
       },
@@ -298,7 +298,7 @@
         "hashed_secret": "fa9beb99e4029ad5a6615399e7bbae21356086b3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 3408,
+        "line_number": 3403,
         "type": "Basic Auth Credentials",
         "verified_result": null
       },
@@ -306,7 +306,7 @@
         "hashed_secret": "fa9beb99e4029ad5a6615399e7bbae21356086b3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 3499,
+        "line_number": 3494,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -314,7 +314,7 @@
         "hashed_secret": "ac371b6dcce28a86c90d12bc57d946a800eebf17",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 3542,
+        "line_number": 3537,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -322,7 +322,7 @@
         "hashed_secret": "0b6ec68df700dec4dcd64babd0eda1edccddace1",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 3547,
+        "line_number": 3542,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -330,7 +330,7 @@
         "hashed_secret": "4ad6f0082ee224001beb3ca5c3e81c8ceea5ed86",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 3552,
+        "line_number": 3547,
         "type": "Secret Keyword",
         "verified_result": null
       }
@@ -4284,7 +4284,7 @@
         "hashed_secret": "c377074d6473f35a91001981355da793dc808ffd",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 4419,
+        "line_number": 4463,
         "type": "Hex High Entropy String",
         "verified_result": null
       }
@@ -4304,7 +4304,7 @@
         "hashed_secret": "3ee08a29a2ca26171fe152b8741d0c90b598263a",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 15016,
+        "line_number": 15004,
         "type": "Secret Keyword",
         "verified_result": null
       }
@@ -4314,7 +4314,7 @@
         "hashed_secret": "6993a3fd94a012ab50fb6b9e97ec238310f0b177",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 166,
+        "line_number": 110,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -4322,7 +4322,7 @@
         "hashed_secret": "f054ffc85b4c1615a7190ea0b248564bb1e9f9ab",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 193,
+        "line_number": 137,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -4330,7 +4330,7 @@
         "hashed_secret": "9ec4f370c96047d96f3eb3637ca9fd144ed6e21b",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 223,
+        "line_number": 167,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -4338,7 +4338,7 @@
         "hashed_secret": "6947818ac409551f11fbaa78f0ea6391960aa5b8",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 550,
+        "line_number": 494,
         "type": "Secret Keyword",
         "verified_result": null
       }
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-07-02T12:51:18Z",
+  "generated_at": "2026-07-02T15:09:47Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-07-02T14:13:52Z",
+  "generated_at": "2026-07-02T12:51:18Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -274,7 +274,7 @@
         "hashed_secret": "b4673e578b9b30fe8bba1b555b7b59883444c697",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1599,
+        "line_number": 1604,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -282,7 +282,7 @@
         "hashed_secret": "4a0a2df96d4c9a13a282268cab33ac4b8cbb2c72",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1687,
+        "line_number": 1692,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -290,7 +290,7 @@
         "hashed_secret": "5baa61e4c9b93f3f0682250b6cf8331b7ee68fd8",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 2037,
+        "line_number": 2042,
         "type": "Basic Auth Credentials",
         "verified_result": null
       },
@@ -298,7 +298,7 @@
         "hashed_secret": "fa9beb99e4029ad5a6615399e7bbae21356086b3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 3403,
+        "line_number": 3408,
         "type": "Basic Auth Credentials",
         "verified_result": null
       },
@@ -306,7 +306,7 @@
         "hashed_secret": "fa9beb99e4029ad5a6615399e7bbae21356086b3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 3494,
+        "line_number": 3499,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -314,7 +314,7 @@
         "hashed_secret": "ac371b6dcce28a86c90d12bc57d946a800eebf17",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 3537,
+        "line_number": 3542,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -322,7 +322,7 @@
         "hashed_secret": "0b6ec68df700dec4dcd64babd0eda1edccddace1",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 3542,
+        "line_number": 3547,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -330,7 +330,7 @@
         "hashed_secret": "4ad6f0082ee224001beb3ca5c3e81c8ceea5ed86",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 3547,
+        "line_number": 3552,
         "type": "Secret Keyword",
         "verified_result": null
       }
@@ -4284,7 +4284,7 @@
         "hashed_secret": "c377074d6473f35a91001981355da793dc808ffd",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 4463,
+        "line_number": 4419,
         "type": "Hex High Entropy String",
         "verified_result": null
       }
@@ -4304,7 +4304,7 @@
         "hashed_secret": "3ee08a29a2ca26171fe152b8741d0c90b598263a",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 15004,
+        "line_number": 15016,
         "type": "Secret Keyword",
         "verified_result": null
       }
@@ -4314,7 +4314,7 @@
         "hashed_secret": "6993a3fd94a012ab50fb6b9e97ec238310f0b177",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 110,
+        "line_number": 166,
         "type": "Secret Keyword",
         "verified_result": null
       },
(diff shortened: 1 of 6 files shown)


Lang-Akshay (changes requested):
Thanks for the PR - @msureshkumar88 
Please address the following findings. 
The diff touches only error-handling paths inside `update_gateway`. No auth tokens, RBAC rules, transport setup, or token-scoping logic is modified. All ContextForge security invariants hold. One medium finding requires a fix: error message construction at line 2491 embeds the raw `init_err` exception string without sanitization. For gateways using `auth_type="query_param"`, the upstream HTTP library exception routinely includes the full request URL with credentials embedded as query parameters. This unsanitized string propagates both to the 502 response body (via `main.py` `str(ex)`) and to the ERROR-level log at line 2775. The codebase already provides `sanitize_exception_message()` and `sanitize_url_for_logging()` imported at line 118 — this is a one-liner fix.

| # | File                                       | Line       | Severity | CWE               | Description                                                                                                                                                                                                                                                                  | Fix                                                                                                                                                                          |
| - | ------------------------------------------ | ---------- | -------- | ----------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| 1 | `mcpgateway/services/gateway_service.py` | 2491       | Medium   | CWE-312 / CWE-532 | Raw`{init_err}` string interpolated into `GatewayConnectionError` without sanitization. For `auth_type="query_param"` gateways, the HTTP library exception includes the full URL with credentials in query params. Leaks to 502 response body and ERROR log.           | Wrap with`sanitize_exception_message(str(init_err), auth_query_params_decrypted)` and `sanitize_url_for_logging(gateway.url, ...)` — both already imported at line 118. |
| 2 | `mcpgateway/services/gateway_service.py` | 2616, 2775 | Low      | CWE-117           | New`logger.warning/error` calls use f-strings, inconsistent with `%s` lazy-eval pattern in the same file (line 2619: `logger.warning("...: %s", e)`). Forces string materialization before log-level check; structured sinks may serialize sensitive URLs into fields. | Change to`logger.warning("Failed to initialize updated gateway: %s", gce)` and `logger.error("GatewayConnectionError during gateway update: %s", gce)`.                  |
| 3 | `mcpgateway/services/gateway_service.py` | 2777       | Low      | —                | `raise gce` by name instead of bare `raise`. Creates new traceback frame, suppresses `__context__` chain, complicates post-incident diagnosis.                                                                                                                         | Replace with bare`raise`.                                                                                                                                                  |

msureshkumar88:
Fixed in 54f339034. All three findings addressed:

**#1 (Medium — CWE-312/532, line 2491):** Wrapped the `GatewayConnectionError` message with `sanitize_url_for_logging()` and `sanitize_exception_message()` (both were already imported). Credentials in query params are now redacted before reaching the 502 response body and the ERROR-level log.

**#2 (Low — CWE-117, lines 2616/2619/2775):** Converted all three f-string logger calls to `%s` lazy-eval pattern, consistent with the rest of the file.

**#3 (Low, line 2777):** Replaced `raise gce` with bare `raise` to preserve the original traceback chain.

Lang-Akshay (changes requested):
| # | File | Issue | Fix |
| - | ---- | ----- | --- |
| 1 | `gateway_service.py`, inner `except Exception as init_err` | `auth_value` and `client_key` not passed to `sanitize_exception_message`; also, when `_initialize_gateway` raises an already-typed `GatewayConnectionError`, the sanitization wrapper is skipped entirely — the original message propagates to the router, which returns `str(ex)` verbatim in the 502 body | Apply sanitization unconditionally; add `gateway.auth_value` and `connection_material.client_key` to the redaction list |
| 2 | `gateway_service.py`, `init_affecting_changed` tuple | Missing `ca_certificate_sig` and `signing_algorithm` — present in `_connect_field_changes` (line 2829) but not here; a `signing_algorithm`-only update with failed re-init will be committed and evict sessions simultaneously | Add both fields to align the two field-sets |
| 3 | `test_gateway_service.py`, new wrapping test | Asserts `GatewayConnectionError` raised but does not verify `api_key=secret123` is absent from the message | Add `assert "secret123" not in str(exc_info.value)` |

msureshkumar88:
Thanks @Lang-Akshay for the careful second pass — the focus on credential leakage in the 502 path is exactly the right thing to scrutinize here. I went through all three and wanted to share what I found, with reasoning.

**Finding 3 — done.** The wrapping test now captures the raised error and asserts the secret is gone:
```python
with pytest.raises(GatewayConnectionError) as exc_info:
    await gateway_service.update_gateway(test_db, 1, gateway_update)
assert "secret123" not in str(exc_info.value)
```
It passes — `sanitize_url_for_logging` redacts `api_key` by default, so the value is stripped. Good call making the sanitization assertion explicit rather than implicit.

**Finding 1 — I think this one is already covered, but please double-check my reading.** The concern is that an already-typed `GatewayConnectionError` skips the wrapper and propagates verbatim. The thing is, the only source of that exception in the inner `try` is `_initialize_gateway`, and that method sanitizes before it ever raises:
```python
# gateway_service.py ~4756
sanitized_url   = sanitize_url_for_logging(url, auth_query_params)
sanitized_error = sanitize_exception_message(raw_error, auth_query_params)
raise GatewayConnectionError(f"Failed to initialize gateway at {sanitized_url}: {sanitized_error}")
```
Its outer `try` (starting ~4697) wraps the whole body, including the OAuth-failure raise at ~4731, so every `GatewayConnectionError` leaving the method is already sanitized. That's why the bare `raise` is intentional — re-sanitizing unconditionally would double-wrap the message (`Failed to initialize gateway at …: Failed to initialize gateway at …`).

On adding `auth_value` / `client_key` to a redaction list: `sanitize_exception_message(message, auth_query_params)` only takes URL query-param names and redacts URLs embedded in the text — there's no general secret-redaction list parameter. `auth_value` (a header) and `client_key` (a PEM) aren't URL query params, so they wouldn't be matched even if passed. If you've seen a concrete path where either lands in the exception string unsanitized, I'd love a pointer — happy to revisit.

**Finding 2 — happy to align if you'd prefer consistency, but I don't think it's a correctness bug.** `ca_certificate_sig` and `signing_algorithm` aren't passed into `_initialize_gateway` (only `ca_certificate`, `client_cert`, `client_key` are, ~2772-2775) — they feed `validate_signature`, not the connection. So a change to either doesn't actually affect whether re-init succeeds, which is why they're absent from `init_affecting_changed`. The asymmetry with `_connect_field_changes` (which evicts sessions on them) is real but, I'd argue, intentional: eviction is about cache freshness, commit-blocking is about connection reachability. I'm fine adding them for symmetry if you feel strongly — just flagging that it widens what counts as a connection failure for fields that don't touch the connection.

Let me know on 1 and 2 and I'll adjust if I've misread anything.

Lang-Akshay (changes requested):
### Required Changes

- [ ] `gateway_service.py` line 2741-2753: Add `ca_certificate_sig` and `signing_algorithm` to `init_affecting_changed` tuple (snapshot vars at lines 2443-2444 already exist)

- [ ] `gateway_service.py` line 2986: Sanitize log: `sanitize_exception_message(str(gce), auth_query_params_decrypted)`

- [ ] `gateway_service.py` after line 2987: Add `structured_logger.log(level="ERROR", event_type="gateway_connection_failure", ...)`

msureshkumar88:
Looked at all three required changes carefully. Here is what I found for each.

**R1 — ca_certificate_sig / signing_algorithm in init_affecting_changed**

These fields are not passed into `_initialize_gateway`. The call at lines 2764–2774 receives `ca_certificate`, `client_cert`, `client_key`, `auth_value`, `auth_type`, `oauth_config`, `auth_query_params`, and `transport` — but not `ca_certificate_sig` or `signing_algorithm`. Those two feed `validate_signature` for cert verification, not the HTTP connection itself.

`init_affecting_changed` answers the question: *will a change to this field cause `_initialize_gateway` to fail?* If only `signing_algorithm` changes, `_initialize_gateway` cannot fail because of that change — it never sees the value. Adding the fields would silently widen "connection failure" to cover non-connection fields, causing valid updates to return 502 when the remote is perfectly reachable.

The asymmetry with `_connect_field_changes` is intentional: that set answers a different question (*should cached sessions be evicted?*), which is a cache-freshness concern, not a reachability one.

**R2 — sanitize_exception_message(str(gce), ...) at line 2986**

`auth_query_params_decrypted` is in scope (initialized at line 2575), so the call is syntactically fine. The issue is that `gce` is already sanitized before it ever reaches line 2985. Every `GatewayConnectionError` raise site in `_initialize_gateway` constructs its message from pre-sanitized values (`sanitized_url`, `sanitized_error`). The wrapped raise at line 2785 uses `safe_url` and `safe_msg`. There is no path where an unsanitized message reaches the outer handler.

Adding a second sanitize pass would be a no-op — the values being redacted are already gone. If a concrete unsanitized path exists that I am missing, please point to it and I will fix it immediately.

**R3 — structured_logger.log for gateway_connection_failure**

This is new functionality. The original issue (#5188) is specifically about connection errors being silently swallowed on gateway update. Adding structured telemetry is a separate concern with its own design questions (which fields to log, how to avoid logging sensitive data in the event payload). Happy to open a follow-up issue for it so it gets proper scope and review. Bundling it here would make this PR harder to review and merge cleanly.

Lang-Akshay (dismissed):
LGTM

msureshkumar88:
Pushed a small refactor addressing the duplicate field-pair tuples flagged in review.

**What changed:** `update_gateway` built the same `(new, old)` comparison tuple for connection-affecting fields (url, transport, auth_type, auth_value, auth_query_params, oauth_config, ca_certificate, client_cert, client_key, and in two spots also ca_certificate_sig/signing_algorithm) in three separate places:
- the async-lifecycle early-return path (session eviction check)
- `init_affecting_changed` (gates whether a failed re-init blocks the commit)
- the post-commit session eviction check

These are now backed by one closure, `_connection_field_pairs(include_signing=True)`, defined once after the original-value snapshot and called from all three sites.

**Why:** it was straight duplication of the same 9-11 line tuple, and having three copies made it easy for them to drift out of sync (which is effectively what the "missing ca_certificate_sig/signing_algorithm" review comment was catching in one of them). One source of truth now.

**Why it's behavior-preserving:** the closure reads `gateway`/original values at call time rather than caching them, since `gateway` attributes get mutated mid-flow (e.g. one_time_auth clearing auth_type/auth_value, query_param auth switching) between when the checks run. Each call site keeps its existing field set via `include_signing`: `init_affecting_changed` still excludes `ca_certificate_sig`/`signing_algorithm` (they're not passed to `_initialize_gateway`, so they can't affect reinit success/failure), while both eviction checks still include them.

Existing tests (`test_gateway_service.py`, `test_multi_auth_headers.py`, `test_upstream_session_registry_lifecycle.py`) pass unchanged.

msureshkumar88:
## E2E validation of the #5188 fix

Ran an end-to-end script against a real running instance (not mocks) to confirm this PR actually fixes the reported bug. Note: I can't submit a formal GitHub "Approve" review here since this is a self-authored PR (GitHub and my own tooling both block self-approval) — posting the full evidence instead so it's reviewable.

### Environment / config used
- Real FastAPI app (`mcpgateway.main:app`) via `uvicorn`, isolated SQLite DB, real Alembic migrations run on boot (default `bootstrap_db` path)
- Real upstream MCP server: a tiny `FastMCP` stdio server exposed over SSE via `python -m mcpgateway.translate --stdio ... --port 9401` (the documented AGENTS.md pattern)
- Auth: real login flow (`POST /auth/login`) against the bootstrapped platform admin (`admin@example.com`), not a hand-crafted token
- Isolated config: `DATABASE_URL=sqlite:///.../mcp_e2e.db`, `JWT_SECRET_KEY`/`AUTH_ENCRYPTION_SECRET` set to test-only values, `AUTH_REQUIRED=true`, `PLUGINS_ENABLED=false`, `CACHE_TYPE=memory`

### Commands run (script: `run_e2e.sh`, available on request)
1. Start upstream MCP server (translate → SSE on :9401)
2. Start gateway app on :18444, wait for `/health` 200
3. `POST /auth/login` → bearer token
4. `POST /gateways` → register gateway pointing at the real upstream
5. `PUT /gateways/{id}` with a connection-affecting change (`url`) pointed at an unreachable host
6. `GET /gateways/{id}` to check for rollback
7. `PUT /gateways/{id}` with a cosmetic-only change (`description`) while upstream still reachable
8. Kill the upstream server
9. `PUT /gateways/{id}` with a cosmetic-only change (`tags`) while upstream is down
10. `PUT /gateways/{id}` with a connection-affecting change containing a secret query param (`api_key=...`), pointed at an unreachable host

### Output

```
Registered gateway id=59d85d0940e24bc189ce1e1d0c9bf4d0
PASS: connection-affecting update to unreachable host returns 502 (got 502)
PASS: gateway url unchanged after failed update (rollback) (got http://localhost:9401/sse)
PASS: cosmetic update succeeds while upstream reachable (got 200)
PASS: cosmetic update tolerates reinit failure (non-connection-affecting) (got 200)
PASS: connection-affecting update with secret returns 502 (got 502)
PASS: secret redacted from 502 body
== 6 passed, 0 failed ==
```

502 response body for the connection-affecting failure (secret case):
```
{"message":"Failed to initialize gateway at http://localhost:9/unreachable-mcp?api_key=REDACTED: All connection attempts failed"}
```

### What this confirms
- The core #5188 bug (a failed re-init on a connection-affecting update silently committing a broken gateway) is fixed: the update is rejected with 502 and rolled back — `url`/`version` are untouched after the failed attempt.
- The pre-existing tolerant behavior for cosmetic-only updates is preserved: they still succeed even when the upstream happens to be unreachable, so this doesn't regress normal "rename/tag/describe a currently-broken gateway" workflows.
- Secrets embedded in the target URL are redacted before reaching the HTTP response, confirming the sanitization path added in this PR.

No functional issues found. Given I can't self-approve, requesting a maintainer take a look to formally approve/merge.

Lang-Akshay (changes requested):
1. Medium — CWE-532: cosmetic-path logs raw exception (gateway_service.py)

Outer except Exception as e: logger.warning(...) on the cosmetic-only update path logs raw exception string. HTTP client errors embed full URLs including auth_query_params API keys → credentials in logs.

# Fix:
logger.warning("Failed to initialize updated gateway: %s",
               sanitize_exception_message(str(e), auth_query_params_decrypted))

msureshkumar88:
Thanks for catching this one — good find.

Confirmed: the generic `except Exception as e` branch on the cosmetic reinit path was logging `str(e)` raw, while its sibling `GatewayConnectionError` branch right above it was already safe (sanitized at construction when the error gets wrapped). HTTP client exceptions can embed the full request URL including auth query params, so this was a real CWE-532 credential-leak path.

Fixed in the latest commit: wrapped the log call with `sanitize_exception_message(str(e), auth_query_params_decrypted)`, same helper already used elsewhere in this function. Ran the update_gateway test suite (28 tests) and ruff — both clean.

Lang-Akshay (dismissed):
LGTM

Lang-Akshay (approved):

PR #5205, opened 2026-06-12
feat: configurable JWT identity claim for API bearer token authentication

### LINKED ISSUE #4924: [FEATURE]: Configurable JWT identity claim for API bearer token authentication
### 🧭 Type of Feature

### 🧭 Epic

**Title:** Configurable JWT identity claim for API bearer token authentication
**Goal:** Allow setting the field used for user identity from access token
**Why now:** The SSO / UI flow uses the email claim from the userinfo endpoint, but the API bearer token path hardcodes `payload.get("sub")` as the user identity.

---

### 🧑🏻‍💻 User Story 1

**As a:** platform-admin
**I want:** to be able to configure the field used for the user identity
**So that:** I don't have to manually provision user accounts and add the users to teams when they are using the API

---

### 📓 Additional Context

When using Okta (or other OIDC providers) in enterprise environments, the `sub` claim often contains an opaque user ID or login username that differs from the user's email address. The SSO/UI login flow correctly uses the email claim from the userinfo endpoint, but the API bearer token path hardcodes `payload.get("sub")` as the user identity.

Since enterprises often cannot modify the sub claim in their Okta authorization server without affecting other applications, a configurable setting (e.g., `JWT_USER_IDENTITY_CLAIM=email`) would allow admins to specify which JWT claim maps to the user's email/identity in the gateway.

This pattern already exists for Keycloak (`SSO_KEYCLOAK_EMAIL_CLAIM`) but is missing for the general JWT/API authentication path.

Our organisation cannot change the value returned in the sub as this is already used for other applications and it would break their authentication flow.

Okta access_tokens are highly configurable with claims etc so it would be easy to return a new field.

But as it currently stands we are having to manually provision a user account for API access while the UI flow will automatically create the user

#### Locations in code

https://github.com/IBM/mcp-context-forge/blob/main/mcpgateway/auth.py#L1452-L1455
Hardcoded to use `sub` but falls back to email if sub doesn't exist

https://github.com/IBM/mcp-context-forge/blob/main/mcpgateway/utils/verify_credentials.py#L512
Hardcoded to use `sub` but falls back to email or username if sub doesn't exist

https://github.com/IBM/mcp-context-forge/blob/main/mcpgateway/utils/verify_credentials.py#L1393
Hardcoded to use `sub` but falls back to username if sub doesn't exist

MohanLaksh:
## Summary

Add `JWT_USER_IDENTITY_CLAIM` environment variable to configure which JWT claim is used for user identity extraction in API bearer token authentication. This enables compatibility with IdPs that use opaque `sub` values (e.g., Okta enterprise, Azure AD) by allowing operators to use `email`, `preferred_username`, or `upn` instead.

Closes #4924

---

## Behavioral Changes

- **New setting**: `jwt_user_identity_claim` (default: `"sub"`) controls which JWT claim API bearer token auth uses to identify users
- **Session tokens** (`token_use: "session"`) always use `sub`/`email` for UUID resolution -- unaffected by this config
- **SSO/UI login** flow has its own provider-specific configuration -- unaffected
- **Audit/forensic paths** (`time_restrictions.py`, `auth_context.py:user_id`) always use raw `sub` claim -- unaffected

## Fallback Chain

When the configured claim is missing from the token payload:

```
configured_claim -> sub -> email -> username
```

A `WARNING` level log is emitted when fallback is used (unless `configured_claim="sub"`) to alert operators of potential misconfiguration.

### Zero-impact (default config)
Existing deployments that do not set `JWT_USER_IDENTITY_CLAIM` see **no behavioral change** -- the default `sub` claim produces identical behavior at every call site.

### Upgrading with a custom claim
To switch to a different identity claim (e.g., `email`):

```bash
# .env or environment
JWT_USER_IDENTITY_CLAIM=email
```

**Requirements:**
- The configured claim must contain a unique identifier matching the user email in the database
- All tokens must include the configured claim -- tokens without it will trigger the fallback chain and log a warning
- Existing tokens with `sub` claims continue to work via the fallback chain (with warning logged)

## Testing

- 17 new unit tests covering:
  - Default `sub` claim extraction
  - Custom claims (`email`, `preferred_username`, `upn`)
  - Full fallback chain (configured_claim -> sub -> email -> username)
  - Session-token bypass (always uses sub/email)
  - Warning-log conditions (on fallback only)
  - Empty and non-string payloads
- All 19,206 existing tests pass with zero regressions
- Backward compatibility verified: 11 token migration tests pass

## Verification

- **ruff**: All checks passed
- **bandit**: No security issues identified
- **interrogate**: 100% docstring coverage
- **pylint**: 10.00/10
- **doctest**: 1178 passed, 54 skipped
- **pre-commit**: All hooks passed

## Documentation

- `AGENTS.md`: Added JWT Identity Claim Configuration section
- `.env.example`: Added `JWT_USER_IDENTITY_CLAIM` with documentation
- Inline docstrings on `extract_identity_from_jwt_payload()` and `get_user_email_from_token()`

## Security

- Session tokens always bypass the configured claim -- cannot be used to escalate via JWT manipulation
- All new imports use inline lazy imports to avoid circular dependencies
- Audit/forensic separations maintained: `request.state.user_id` continues to log raw `sub` claim
- SSO callback path explicitly excluded (separate provider-specific config)
- Non-string claim values are properly rejected (type guard in helper function)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -21,6 +21,12 @@ BASIC_AUTH_PASSWORD=changeme
 # Generate with: python -m mcpgateway.scripts.init_secrets --stdout
 JWT_SECRET_KEY=__REPLACE_ME__run_init-secrets_before_starting
 
+# JWT claim to use for user identity in API bearer token authentication
+# Default: "sub" (standard OIDC subject claim)
+# Set to "email", "preferred_username", or "upn" if your IdP uses non-email sub claims
+# Must match the claim that contains the user's email/unique identifier in your IdP's access tokens
+#JWT_USER_IDENTITY_CLAIM=sub
+
 # Passphrase used to encrypt stored auth secrets
 # Generate with: python -m mcpgateway.scripts.init_secrets --stdout
 AUTH_ENCRYPTION_SECRET=__REPLACE_ME__run_init-secrets_before_starting
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -165,6 +165,25 @@ ContextForge implements a **two-layer security model**:
 
 **Rationale**: The `email` field is the human-readable identifier used throughout AGENTS.md and user-facing documentation. Consistent precedence prevents forensic confusion where an incident review pivots on a logged email that differs from the principal actually evaluated by RBAC.
 
+### JWT Identity Claim Configuration
+
+API bearer token authentication extracts user identity from JWT claims using a configurable claim field:
+
+- **Default**: `sub` claim (maintains backward compatibility)
+- **Configurable via**: `JWT_USER_IDENTITY_CLAIM` environment variable
+- **Common alternatives**: `email`, `preferred_username`, `upn`
+- **Fallback chain**: configured_claim → sub → email → username
+
+**Important**:
+- Session tokens always use `sub` for UUID resolution (not affected by this config)
+- The configured claim must contain a unique identifier matching the user's email in the database
+- SSO/UI login flow uses provider-specific userinfo endpoint (separate from this config)
+
+**Example** (Okta enterprise with opaque sub claim):
+```bash
+JWT_USER_IDENTITY_CLAIM=email  # Use email claim instead of sub
+```
+
 
 ## Observability Transaction Behavior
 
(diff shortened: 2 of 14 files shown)


snorlaX-sleeps:
any thoughts on this?

PR #5211, opened 2026-06-12
fix(playwright): user deletion FK cascade, team selector delegation, …

marekdano:
Closes https://github.ibm.com/contextforge-org/internal_issues/issues/297/

Three Playwright tests are improved.

1. `test_delete_user` - HTTP 400 on `DELETE /admin/users/<email>`

`delete_user` in `EmailAuthService `was raising a SQLAlchemy `StaleDataError` at commit time when deleting a user who owned a single-member personal team. The original code bulk-deleted team members via raw SQL, then called `session.delete(team)`. At flush, the ORM's `cascade="all, delete-orphan"` on `EmailTeam.members` tried to re-DELETE the same objects (still in the session identity map) - 0 rows affected → StaleDataError → HTTP 400.

The fix:
  1. Explicitly deletes `EmailTeamMemberHistory` rows by `team_id` first — that FK has no `ON DELETE CASCADE` so it must be removed before the team row is.
  2. Keeps the raw SQL member delete to clear the rows from the DB.
  3. Calls `session.expire(team, ["members"])` so the ORM cascade re-queries the (now-empty) collection at flush instead of attempting to re-delete already-gone objects.

The corresponding unit test mock's side_effect list is updated to expect the additional `execute()` call.

2. `test_iframe_team_selector_onclick_stripped_but_delegation_works` - `team_id` absent from iframe URL after clicking team

Team selector buttons injected via innerHTML (and processed by the innerHTML guard that strips on* attributes) carry `data-action="select-team"` but had no event listener. The `handleDelegatedClick` function in `eventDelegation.js` only handled `[data-action-click]`, so clicking a team item did nothing and `updateTeamContext` was never called.

The fix adds a `[data-action="select-team"]` check at the top of `handleDelegatedClick` that calls `window.Admin.selectTeamFromSelector(target)` — the same function already used for direct-onclick paths. The JS bundle is rebuilt.

3. `test_should_handle_object_parameter_validation` - "Execution context was destroyed" on `page.evaluate`

After `search_input.press("Enter")`, HTMX pushes the search query into the URL via `hx-push-url`. Playwright treats the resulting `history.pushState` as an in-flight navigation. If `page.evaluate("Admin.testTool(...)")` is called before that navigation event settles, Playwright destroys the execution context mid-call. A `page.wait_for_load_state("domcontentloaded")` before the evaluate ensures the navigation has completed before
  JavaScript is injected.

### DIFF
--- a/mcpgateway/admin_ui/eventDelegation.js
+++ b/mcpgateway/admin_ui/eventDelegation.js
@@ -116,6 +116,17 @@ function executeAction(action, args, event, eventFirst = false) {
  * @param {Event} event - The click event
  */
 function handleDelegatedClick(event) {
+  // Handle team selector item clicks (data-action="select-team" on buttons injected
+  // via innerHTML, where onclick attributes are stripped by the innerHTML guard).
+  const selectTeamTarget = event.target.closest('[data-action="select-team"]');
+  if (selectTeamTarget) {
+    event.preventDefault();
+    if (window.Admin && typeof window.Admin.selectTeamFromSelector === 'function') {
+      window.Admin.selectTeamFromSelector(selectTeamTarget);
+    }
+    return;
+  }
+
   const target = event.target.closest('[data-action-click]');
   if (!target) return;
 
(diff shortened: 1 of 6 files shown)


brian-hussey:
Running tests to check that it passes now.

brian-hussey:
This one fixed some tests and others still fail. Treating them as follow up fixes, getting this one in.

brian-hussey (dismissed):
LGTM

brian-hussey (approved):
LGTM

PR #5224, opened 2026-06-15
feat(oauth): RFC 8693 token exchange (On-Behalf-Of) for OAuth gateways

### LINKED ISSUE #5135: [FEATURE]: RFC 8693 Token Exchange Support for Upstream MCP Servers
### 🧭 Type of Feature

- [x] Developer tooling or test improvement

### 🧭 Epic

**Title:** RFC 8693 Token Exchange for On-Behalf-Of (OBO) Upstream Authentication

**Goal:** Enable ContextForge to exchange user tokens for downstream-scoped tokens when connecting to upstream MCP servers, supporting enterprise token federation patterns where the authorization server requires explicit token exchange rather than direct token forwarding.

**Why now:** 
- The `OAuthManager.token_exchange()` method exists but is not integrated into any gateway configuration workflow
- Enterprise deployments with PingFederate, Toolhive, and similar token federators require RFC 8693 for proper token scoping
- Current Authorization Code flow forwards tokens as-is, which doesn't work when the upstream requires a different audience or token type
- ADR-041 identified this as a design decision but implementation was deferred

**Who benefits:**
- Enterprise users with token federation infrastructure (PingFederate, Toolhive, Keycloak)
- Multi-tenant deployments requiring strict token audience scoping
- Organizations implementing zero-trust architectures with per-service token validation

---

### 🧑🏻‍💻 User Story 1: Gateway Administrator

**As a:** Gateway administrator
**I want:** To configure a gateway to use RFC 8693 token exchange
**So that:** The gateway can obtain properly-scoped tokens for upstream MCP servers that require explicit token exchange

#### ✅ Acceptance Criteria
```gherkin
Scenario: Configure gateway with token exchange grant type
  Given I am creating a new gateway via Admin UI or API
  When I set auth_type to "oauth"
  And I set grant_type to "token-exchange"
  And I provide token_url, client_id, client_secret, and target_audience
  Then the gateway configuration is saved successfully
  And the oauth_config includes all token exchange parameters

Scenario: Token exchange on tool invocation
  Given a gateway configured with grant_type "token-exchange"
  And a user authenticated to ContextForge with a valid JWT
  When the user invokes a tool on that gateway
  Then ContextForge calls OAuthManager.token_exchange() with the user's JWT as subject_token
  And the exchanged token is forwarded to the upstream MCP server as Authorization: Bearer <exchanged-token>
  And the original user JWT is NOT forwarded to the upstream

Scenario: Token exchange failure handling
  Given a gateway configured with token exchange
  When the token exchange request fails (401/403/500)
  Then the tool invocation fails with a clear error message
  And the error indicates token exchange failure (not generic auth failure)
  And the audit log records the token exchange attempt and failure reason
```

---

### 🧑🏻‍💻 User Story 2: End User

**As a:** End user invoking tools through ContextForge
**I want:** Token exchange to happen transparently
**So that:** I can access upstream MCP servers without manual token management

#### ✅ Acceptance Criteria
```gherkin
Scenario: Transparent token exchange
  Given I am authenticated to ContextForge
  And a gateway is configured with token exchange
  When I invoke a tool on that gateway
  Then token exchange happens automatically
  And I receive the tool result without additional authentication steps
  And I am not aware that token exchange occurred

Scenario: Token exchange with caching
  Given token exchange is configured with caching enabled
  When I invoke multiple tools on the same gateway within the token lifetime
  Then token exchange occurs only once
  And subsequent invocations reuse the cached exchanged token
  And the cache respects the exchanged token's expiry time
```

---

### 🧑🏻‍💻 User Story 3: Security Auditor

**As a:** Security auditor
**I want:** Complete audit trails for token exchange operations
**So that:** I can verify proper token scoping and detect unauthorized access attempts

#### ✅ Acceptance Criteria
```gherkin
Scenario: Token exchange audit logging
  Given token exchange is configured
  When a token exchange occurs
  Then the audit log records:
    - Original subject token claims (user_id, email, not the token itself)
    - Target audience requested
    - Token exchange success/failure
    - Exchanged token expiry time
    - Upstream server that received the exchanged token

Scenario: Token exchange metrics
  Given observability is enabled
  When token exchanges occur
  Then metrics are recorded for:
    - Token exchange request count (success/failure)
    - Token exchange latency (p50, p95, p99)
    - Token exchange errors by type (401, 403, 500, timeout)
    - Cached vs fresh token exchange ratio
```

---

#### 1. Configuration Schema Extension

**Gateway OAuth Config** (`mcpgateway/schemas.py`):

```python
class OAuthConfig(BaseModel):
    grant_type: Literal["client_credentials", "authorization_code", "password", "token-exchange"]
    
    # Existing fields...
    client_id: str
    client_secret: Optional[str] = None
    token_url: str
    scopes: Optional[List[str]] = None
    
    # New fields for token exchange
    subject_token_source: Optional[Literal["gateway_jwt", "user_oauth_token"]] = "gateway_jwt"
    target_audience: Optional[str] = None  # Required for token-exchange grant
    requested_token_type: Optional[str] = "urn:ietf:params:oauth:token-type:access_token"
    
    # Token exchange specific
    actor_token: Optional[str] = None  # For delegation chains
    actor_token_type: Optional[str] = None
    
    @validator("target_audience")
    def validate_target_audience_for_exchange(cls, v, values):
        if values.get("grant_type") == "token-exchange" and not v:
            raise ValueError("target_audience is required for token-exchange grant type")
        return v
```

#### 2. Integration Points

**Tool Service** (`mcpgateway/services/tool_service.py`):

```python
# Around line 4010, add token-exchange branch
if grant_type == "token-exchange":
    try:
        # Get subject token (user's JWT or stored OAuth token)
        if oauth_config.get("subject_token_source") == "user_oauth_token":
            # Use stored OAuth token as subject
            subject_token = await token_storage.get_user_token(gateway_id_str, app_user_email)
        else:
            # Use gateway JWT (from request.state.user or auth context)
            subject_token = extract_jwt_from_request(request)
        
        if not subject_token:
            raise ToolInvocationError(
                f"No subject token available for token exchange on gateway '{gateway_name}'. "
                f"User must be authenticated."
            )
        
        # Perform token exchange
        exchanged_token_response = await self.oauth_manager.token_exchange(
            token_url=oauth_config["token_url"],
            subject_token=subject_token,
            client_id=oauth_config["client_id"],
            client_secret=oauth_config.get("client_secret", ""),
            audience=oauth_config.get("target_audience"),
            scope=" ".join(oauth_config.get("scopes", [])),
            requested_token_type=oauth_config.get("requested_token_type"),
        )
        
        # Cache exchanged token (per gateway + user + audience)
        cache_key = f"token_exchange:{gateway_id_str}:{app_user_email}:{oauth_config.get('target_audience')}"
        await cache_exchanged_token(
            cache_key, 
            exchanged_token_response["access_token"],
            expires_in=exchanged_token_response.get("expires_in", 3600)
        )
        
        headers = {"Authorization": f"Bearer {exchanged_token_response['access_token']}"}
        
    except Exception as e:
        logger.error(f"Token exchange failed for gateway {gateway_name}: {e}")
        raise ToolInvocationError(
            f"Token exchange failed for gateway '{gateway_name}': {str(e)}. "
            f"Check oauth_config (token_url, target_audience, client credentials) and authorization server logs."
        )
```

**Gateway Service** (`mcpgateway/services/gateway_service.py`):

Similar integration around line 1620 for gateway connection/health checks.

#### 3. Token Caching Strategy

**Cache Key Structure:**
```
token_exchange:{gateway_id}:{user_email}:{target_audience}
```

**Cache Implementation** (new service `mcpgateway/services/token_exchange_cache.py`):

```python
class TokenExchangeCache:
    """Cache for exchanged tokens with automatic expiry."""
    
    async def get_cached_token(
        self, 
        gateway_id: str, 
        user_email: str, 
        audience: str,
        threshold_seconds: int = 300
    ) -> Optional[str]:
        """Get cached exchanged token if valid and not near expiry."""
        
    async def cache_token(
        self,
        gateway_id: str,
        user_email: str,
        audience: str,
        access_token: str,
        expires_in: int
    ) -> None:
        """Cache exchanged token with TTL."""
        
    async def invalidate_token(
        self,
        gateway_id: str,
        user_email: str,
        audience: str
    ) -> None:
        """Invalidate cached token (e.g., on 401 from upstream)."""
```

**Storage Backend:**
- Redis (preferred for distributed deployments)
- Database (fallback for single-instance)
- In-memory (development only)

#### 4. Audit Trail Enhancement

**New fields on `AuditTrail` model:**

```python
class AuditTrail(Base):
    # Existing fields...
    
    # Token exchange specific
    token_exchange_occurred: bool = False
    token_exchange_target_audience: Optional[str] = None
    token_exchange_success: bool = False
    token_exchange_error: Optional[str] = None
    token_exchange_latency_ms: Optional[int] = None
```

#### 5. Admin UI Changes

**Gateway Form** (`mcpgateway/admin_ui/gateway_form.html`):

```html
<!-- Add to OAuth configuration section -->
<div id="token-exchange-fields" style="display: none;">
    <label>Subject Token Source:</label>
    <select name="oauth_config.subject_token_source">
        <option value="gateway_jwt">Gateway JWT (default)</option>
        <option value="user_oauth_token">User OAuth Token</option>
    </select>
    
    <label>Target Audience: <span class="required">*</span></label>
    <input type="text" name="oauth_config.target_audience" 
           placeholder="https://downstream-service.example.com" required>
    
    <label>Requested Token Type:</label>
    <input type="text" name="oauth_config.requested_token_type" 
           value="urn:ietf:params:oauth:token-type:access_token">
</div>

<script>
// Show/hide token exchange fields based on grant type
document.querySelector('[name="oauth_config.grant_type"]').addEventListener('change', (e) => {
    const tokenExchangeFields = document.getElementById('token-exchange-fields');
    tokenExchangeFields.style.display = e.target.value === 'token-exchange' ? 'block' : 'none';
});
</script>
```

---

#### Integration Tests (Priority Order)

**1. Keycloak Integration (MUST HAVE)**

Test file: `tests/integration/test_oauth_token_exchange_keycloak.py`

```python
@pytest.mark.integration
class TestKeycloakTokenExchange:
    """RFC 8693 token exchange against Keycloak."""
    
    async def test_successful_token_exchange_keycloak(self):
        """
        Given: Keycloak configured with token exchange enabled
        And: A gateway configured with grant_type="token-exchange"
        When: User invokes a tool with valid JWT
        Then: Token exchange succeeds
        And: Exchanged token has correct audience
        And: Tool invocation succeeds with exchanged token
        """
        
    async def test_token_exchange_invalid_audience_keycloak(self):
        """Verify token exchange fails with invalid target_audience."""
        
    async def test_token_exchange_caching_keycloak(self):
        """Verify exchanged tokens are cached and reused."""
        
    async def test_token_exchange_cache_expiry_keycloak(self):
        """Verify cached tokens are refreshed when near expiry."""
```

**Keycloak Setup:**
- Docker Compose service for Keycloak
- Realm configuration with token exchange enabled
- Client configuration with token-exchange grant type
- Test users and roles

**2. PingFederate Integration (SHOULD HAVE)**

Test file: `tests/integration/test_oauth_token_exchange_pingfederate.py`

Similar test structure as Keycloak, but against PingFederate:
- Token exchange with PingFederate-specific parameters
- Delegation chain support
- Actor token handling

**3. Toolhive Integration (SHOULD HAVE)**

Test file: `tests/integration/test_oauth_token_exchange_toolhive.py`

Toolhive-specific token exchange patterns:
- Multi-hop token exchange
- Service mesh integration
- Custom token types

#### Unit Tests

**Test file:** `tests/unit/mcpgateway/services/test_oauth_token_exchange_integration.py`

```python
class TestTokenExchangeIntegration:
    """Unit tests for token exchange integration into gateway workflows."""
    
    async def test_tool_invocation_with_token_exchange(self):
        """Verify tool_service calls token_exchange for token-exchange grant type."""
        
    async def test_gateway_connection_with_token_exchange(self):
        """Verify gateway_service calls token_exchange during connection."""
        
    async def test_token_exchange_cache_hit(self):
        """Verify cached tokens are reused."""
        
    async def test_token_exchange_cache_miss(self):
        """Verify fresh exchange when cache is empty."""
        
    async def test_token_exchange_error_handling(self):
        """Verify proper error messages on token exchange failure."""
```

#### End-to-End Tests

**Test file:** `tests/e2e/test_token_exchange_e2e.py`

```python
@pytest.mark.e2e
class TestTokenExchangeE2E:
    """End-to-end token exchange scenarios."""
    
    async def test_complete_token_exchange_flow(self):
        """
        Given: User authenticated to ContextForge
        And: Gateway configured with token exchange
        And: Upstream MCP server requiring specific audience
        When: User invokes tool via MCP client
        Then: Token exchange occurs transparently
        And: Tool executes successfully
        And: Audit log shows token exchange
        """
```

---

### 📐 Architecture Diagram

```mermaid
sequenceDiagram
    participant User
    participant Gateway as ContextForge Gateway
    participant AuthServer as Authorization Server
    participant Upstream as Upstream MCP Server

    User->>Gateway: Invoke tool (with JWT)
    Gateway->>Gateway: Extract user JWT (subject_token)
    Gateway->>Gateway: Check token exchange cache
    
    alt Cache Miss
        Gateway->>AuthServer: POST /token<br/>grant_type=token-exchange<br/>subject_token=<user_jwt><br/>audience=<upstream_url>
        AuthServer->>AuthServer: Validate subject_token
        AuthServer->>AuthServer: Generate new token for audience
        AuthServer-->>Gateway: {access_token, expires_in}
        Gateway->>Gateway: Cache exchanged token
    else Cache Hit
        Gateway->>Gateway: Use cached token
    end
    
    Gateway->>Upstream: Tool invocation<br/>Authorization: Bearer <exchanged_token>
    Upstream->>Upstream: Validate token audience
    Upstream-->>Gateway: Tool result
    Gateway-->>User: Tool result
    
    Gateway->>Gateway: Record audit trail<br/>(token_exchange_occurred=true)
```

---

### 🔗 MCP Standards Check

- [x] OAuth is gateway-level configuration, not MCP protocol change
- [x] Token exchange is transparent to MCP clients

**Deviations:** None. This is an enhancement to gateway authentication, not a change to MCP protocol.

---

### 🔄 Alternatives Considered

**1. Direct Token Forwarding (Current Approach)**
- **Pros:** Simple, no additional OAuth calls
- **Cons:** Doesn't work when upstream requires different audience; violates least-privilege principle
- **Verdict:** Insufficient for enterprise token federation

**2. Client-Side Token Exchange**
- **Pros:** Gateway doesn't handle token exchange
- **Cons:** Exposes token exchange complexity to clients; breaks MCP abstraction
- **Verdict:** Rejected - violates gateway proxy pattern

**3. Always-On Token Exchange**
- **Pros:** Consistent behavior across all gateways
- **Cons:** Performance overhead for gateways that don't need it; breaking change
- **Verdict:** Rejected - should be opt-in via grant_type

**4. Plugin-Based Token Exchange**
- **Pros:** Flexible, extensible
- **Cons:** Too complex for standard OAuth pattern; plugins shouldn't handle core auth
- **Verdict:** Rejected - token exchange is core OAuth functionality

---

### 📓 Additional Context

**Related Issues:**
- ADR-041: Identity Propagation (mentions token exchange as design decision)
- Issue #605: Access to remote MCP Servers/Tools via OAuth on behalf of Users

**Related Documentation:**
- `docs/docs/architecture/oauth-design.md` - OAuth architecture
- `docs/docs/manage/identity-propagation.md` - Identity forwarding
- `mcpgateway/services/oauth_manager.py` - Existing token_exchange() implementation

**Token Federators Cited:**
- **Keycloak** (open source, MUST support for integration tests)
- **PingFederate** (enterprise, cited by users)
- **Toolhive** (enterprise, cited by users)

**Implementation Phases:**

**Phase 1: Core Implementation (MVP)**
- Configuration schema extension
- Integration into tool_service and gateway_service
- Basic token caching (in-memory)
- Keycloak integration tests

**Phase 2: Production Hardening**
- Redis-backed token cache
- Comprehensive audit logging
- Admin UI enhancements
- PingFederate integration tests

**Phase 3: Advanced Features**
- Actor token support (delegation chains)
- Token exchange metrics and observability
- Toolhive integration tests
- Performance optimization (batch exchange, prefetch)

**Estimated Effort:**
- Phase 1: 2-3 weeks (1 developer)
- Phase 2: 1-2 weeks (1 developer)
- Phase 3: 1-2 weeks (1 developer)
- **Total: 4-7 weeks**

**Dependencies:**
- Keycloak Docker image for integration tests
- Access to PingFederate/Toolhive test environments (Phase 2/3)
- Token exchange enabled on authorization servers

---

### ✅ Definition of Done

- [ ] Configuration schema supports `grant_type: "token-exchange"`
- [ ] Tool invocation integrates token exchange for token-exchange grant type
- [ ] Gateway connection integrates token exchange for token-exchange grant type
- [ ] Token caching implemented with Redis/DB/memory backends
- [ ] Audit trail records token exchange operations
- [ ] Admin UI supports token exchange configuration
- [ ] Keycloak integration tests pass (MUST HAVE)
- [ ] PingFederate integration tests pass (SHOULD HAVE)
- [ ] Toolhive integration tests pass (SHOULD HAVE)
- [ ] Unit tests achieve >90% coverage for new code
- [ ] Documentation updated (oauth.md, identity-propagation.md)
- [ ] Migration guide for existing OAuth gateways
- [ ] Performance benchmarks show <100ms token exchange overhead
- [ ] Security review completed (token handling, caching, audit)

msureshkumar88:
Closes #5135

## Summary

- Adds a `token-exchange` OAuth grant type for gateways (RFC 8693 / On-Behalf-Of): the user's inbound ContextForge JWT (or stored upstream token) is exchanged with a trusted Authorization Server for a downstream-audience-scoped token, which is used in place of the user JWT — never forwarded upstream.
- Config validated at gateway create/update (`target_audience` required, SSRF-checked `token_url`, `subject_token_source` enum).
- Exchanged tokens cached (Redis + in-memory fallback) with single-flight resolution, negative caching, TTL from the AS `expires_in`, and 401-triggered cache invalidation + single re-exchange retry on REST-integration tool calls.
- Passthrough hardening strips `authorization` from passthrough headers for token-exchange gateways so the exchanged token always wins.
- Structured audit events (no token material) for every exchange attempt, persisted to the structured-log DB sink.
- Docs: new "Token Exchange (RFC 8693 / On-Behalf-Of)" section in oauth-design.md, migration guide in identity-propagation.md, troubleshooting table, AGENTS.md security invariants, ADR-041 update.

## Solutions considered

`OAuthManager.token_exchange()` (RFC 8693) was already implemented but unwired into any gateway workflow. Three integration approaches were evaluated:

- **A — Inline branch at each of the 4 OAuth call sites** (`tool_service.py:4010`/`:5194`, `gateway_service.py:3818`/`:4183`). Closest to the issue's sketch; subject token already in scope at each site. Rejected — duplicates exchange/caching/audit logic 4x with drift risk.
- **B — Centralize in `OAuthManager.get_access_token()` + a resolver (CHOSEN).** Adds a `token-exchange` grant branch and a `subject_token` parameter so all four call sites route through one seam. The cached invoke path uses a resolver (`_resolve_token_exchange_header`, duplicated with full parity on `ToolService` and `GatewayService`) that calls `token_exchange()` directly to capture the real `expires_in` for cache TTL. Single place for caching/audit; cleanest test surface.
- **C — Dedicated `TokenExchangeService` + cache.** Best end-state, but too much upfront scope for an MVP. B's resolver+cache is the seam that lifts into C with zero call-site churn — deferred to Phase 2.

(Plugin-based token exchange was rejected by the issue itself — token exchange is core OAuth, not a plugin concern.)

## Design decisions

| Decision | Choice | Why |
|---|---|---|
| Integration approach | **B** — central dispatch in `get_access_token()` + resolver | Single seam for cache/audit/passthrough hardening; upgrades to Approach C without call-site churn |
| Cache backend | Redis with in-memory LRU fallback (`TokenExchangeCache`) | Matches existing patterns (`oauth_manager`, session affinity); single-RTT hits, bounded memory, negative cache, per-key single-flight lock, Redis circuit-breaker |
| Cache key | `token_exchange:{gateway_id}:{user_email}:{target_audience}` | Per-user, per-audience scoping — no cross-user token leakage |
| `subject_token_source` | `inbound_user_jwt` (default) \| `user_oauth_token` (secondary, tool-invoke path only) | Greenfield config key (0 hits in shipped code); `inbound_user_jwt` is the codebase-correct name — it's the end-user's inbound JWT, not a gateway-minted service token (`gateway_jwt` was the issue sketch's name and was rejected) |
| `target_audience` / `token_url` | Required at config time; `token_url` SSRF-validated via `SecurityValidator.validate_url` | The inbound user JWT is POSTed to `token_url`, so it's an egress/SSRF boundary — creating/editing token-exchange gateways is a privileged action |
| Audit (MVP) | Structured log event via `mcpgateway.audit.token_exchange`, no raw token material, propagates to the `StructuredLogEntry` DB sink when `STRUCTURED_LOGGING_DATABASE_ENABLED=true` | Zero-migration MVP; typed `AuditTrail` columns deferred to Phase 2 |
| Passthrough hardening | Strip `authorization` from passthrough headers for token-exchange gateways | Ensures the exchanged token always wins — the inbound user JWT can never override it upstream |
| 401 handling | On a 401 from a token-exchange-gateway REST call: invalidate cache entry, re-resolve, retry exactly once | Self-heals from a stale/expired exchanged token without uncontrolled retry storms; MCP/SSE/streamable-HTTP calls (no retryable HTTP status) are out of scope |
| Error surface | Generic caller-facing errors ("Token exchange failed... Contact your administrator"); AS error detail logged server-side only (sanitized + `exc_info`) | Avoids leaking Authorization Server internals to callers while preserving forensic detail |
| Negative caching | Failed exchanges cached briefly (`is_failed`/`set_failure`) | Short-circuits retry storms during an IdP outage |
| Admin UI | None (Phase 2) — API-only configuration via `POST`/`PUT /gateways` `oauth_config` | Keeps MVP scope bounded; avoids risk of the existing gateway edit form silently dropping unknown `oauth_config` keys |
| Backward compatibility | Additive/opt-in via `grant_type: "token-exchange"`; no DB migration | `subject_token_source`/`inbound_user_jwt`/`target_audience` have 0 hits in shipped code — nothing to migrate, no existing config affected |

## Design docs

- Spec: `docs/superpowers/specs/2026-06-12-rfc8693-token-exchange-design.md`
- Plan: `docs/superpowers/plans/2026-06-12-rfc8693-token-exchange.md`
- Architecture docs: [`docs/docs/architecture/oauth-design.md#token-exchange-rfc-8693--on-behalf-of`](https://github.com/IBM/mcp-context-forge/blob/feat/issue-5135-rfc8693-token-exchange/docs/docs/architecture/oauth-design.md), migration guide in `docs/docs/manage/identity-propagation.md#migrating-oauth-gateways-to-token-exchange`, ADR-041 update

### 1. Set up a token-exchange-capable Authorization Server (Keycloak)

```bash
docker run -p 8080:8080 \
  -e KEYCLOAK_ADMIN=admin -e KEYCLOAK_ADMIN_PASSWORD=admin \
  quay.io/keycloak/keycloak:latest \
  start-dev --features=token-exchange
```

- Create/import a realm (e.g. `cf`).
- Create a confidential client (e.g. `token-exchange`) with "Standard token exchange" enabled and client authentication on; note the client secret.
- Configure the realm to **trust ContextForge as a JWT issuer** (or use federated SSO with a shared issuer) so an inbound CF JWT is accepted as a `subject_token`.
- Note a target audience (another client ID or custom audience identifier).

### 2. Register a token-exchange gateway (API-only)

```bash
export MCPGATEWAY_BEARER_TOKEN=$(python -m mcpgateway.utils.create_jwt_token --username admin@example.com --exp 60 --secret <JWT_SECRET_KEY>)

curl -s -X POST http://localhost:4444/gateways \
  -d '{
    "name": "downstream-mcp",
    "url": "https://downstream.example.com/mcp",
    "auth_type": "oauth",
    "oauth_config": {
      "grant_type": "token-exchange",
      "token_url": "http://localhost:8080/realms/cf/protocol/openid-connect/token",
      "client_id": "token-exchange",
      "client_secret": "<keycloak-client-secret>",
      "target_audience": "<target-audience>",
      "subject_token_source": "inbound_user_jwt"
    }
  }'
```

Negative checks (config validation):
- Omit `target_audience` → expect `400`/`422` with "target_audience is required for token-exchange grant type".
- Set `token_url` to an internal/link-local address (e.g. `http://169.254.169.254/...`) → expect SSRF rejection.
- Set `subject_token_source` to an invalid value → expect rejection naming `subject_token_source`.

### 3. Verify the exchange happens transparently on tool invocation

- Obtain a normal user JWT (`create_jwt_token --username user@example.com ...`) and call a tool routed through the new gateway, e.g. `POST /rpc` (`tools/call`) with that user's bearer.
- Confirm the call succeeds and that the **upstream** (mock server or `httpbin`/request-logger in front of `downstream.example.com`) receives a **different** `Authorization` header than the inbound user JWT — i.e. the exchanged token, never the user's CF JWT.

### 4. Verify caching

- Repeat the same call (same user, same gateway) — confirm no new request hits the Keycloak token endpoint (Keycloak access logs / `Network` tab), i.e. the cached exchanged token is reused.
- Wait past the exchanged token's `expires_in` (or use a short-lived test client) and call again — confirm a fresh exchange occurs.

### 5. Verify 401 invalidation + retry (REST integration)

- Point the gateway at a mock upstream that returns `401` once then `200`.
- Confirm the tool call succeeds (one retry after cache invalidation + re-exchange) rather than failing.

### 6. Verify audit trail

- With `STRUCTURED_LOGGING_DATABASE_ENABLED=true`, query `StructuredLogEntry` for `is_security_event=True` rows with `token-exchange` events for the test calls above.
- Confirm `user_email`, `gateway_id`, `target_audience`, `success`, `exchanged_token_expires_in`, `correlation_id` are present and **no** `access_token`/`subject_token`/`client_secret` values appear.

### 7. Verify passthrough hardening

- Send a request with a custom `Authorization` header set via passthrough config for the token-exchange gateway — confirm the exchanged token's `Authorization` header is used upstream, not the passthrough value.

### Automated coverage (already run in CI)

- `pytest tests/unit/mcpgateway/services/test_token_exchange_cache.py tests/unit/mcpgateway/services/test_token_exchange_integration.py -v`
- `pytest tests/e2e/test_token_exchange_e2e.py tests/integration/test_token_exchange_mock_upstream.py -v`
- `KEYCLOAK_URL=... KEYCLOAK_CLIENT_ID=... KEYCLOAK_CLIENT_SECRET=... KEYCLOAK_USER_JWT=... KEYCLOAK_TARGET_AUDIENCE=... pytest tests/integration/test_oauth_token_exchange_keycloak.py -v`


docs/docs/architecture/adr/041-identity-propagation.md
```diff
@@ -79,6 +79,50 @@ Three new fields on `AuditTrail` records:
...
 An `OAuthManager.token_exchange()` method supports on-behalf-of flows for gateways that require a user-scoped access token rather than client credentials.
 
+> **Update (2026-06):** Implemented. See "Implementation Update: RFC 8693 Token Exchange" below for the realized design.
+
+## Implementation Update: RFC 8693 Token Exchange
+
+- *Status:* Implemented
+- *Date:* 2026-06
+
+RFC 8693 / On-Behalf-Of token exchange (item 7 above) was implemented via **Approach B**: a central resolver-method seam rather than a new standalone service. Each consuming service (`ToolService`, `GatewayService`) gets its own `_resolve_token_exchange_header()` resolver, backed by a shared `TokenExchangeCache`.
```
prakhar-singh1928: What does approach B mean?
What is approach A?

docs/docs/architecture/adr/041-identity-propagation.md
```diff
@@ -79,6 +79,50 @@ Three new fields on `AuditTrail` records:
...
+
+### Configuration
+
+A gateway's `oauth_config` opts into the flow with `grant_type: "token-exchange"`, plus:
+
+- `target_audience` (required) — the downstream resource the exchanged token is for.
+- `subject_token_source` (default `inbound_user_jwt`) — `inbound_user_jwt` uses the caller's ContextForge JWT as the RFC 8693 `subject_token`; `user_oauth_token` uses the user's previously stored per-gateway OAuth token instead (supported on the tool-invocation path).
+- `requested_token_type` (default `urn:ietf:params:oauth:token-type:access_token`).
+
+### Resolver Seam (Approach B)
```
prakhar-singh1928: same question. why is resolver seam -> approach B?

mcpgateway/services/gateway_service.py
```diff
@@ -568,6 +568,11 @@ def __init__(self) -> None:
         self.oauth_manager = OAuthManager(request_timeout=int(os.getenv("OAUTH_REQUEST_TIMEOUT", "30")), max_retries=int(os.getenv("OAUTH_MAX_RETRIES", "3")))
         self._event_service = EventService(channel_name="mcpgateway:gateway_events")
 
+        # First-Party
+        from mcpgateway.services.token_exchange_cache import TokenExchangeCache  # pylint: disable=import-outside-toplevel
```
prakhar-singh1928: move to outside 
move all inline imports to top level

docs/docs/architecture/adr/041-identity-propagation.md
msureshkumar88: Good catch — fixed in c018d524abec2cd8e2eeab338dfd5a256f3ed182: removed the unexplained "Approach A/B" labels and described the resolver-method seam directly (also noted that a standalone TokenExchangeService was considered and deferred).

docs/docs/architecture/adr/041-identity-propagation.md
msureshkumar88: Same fix as above in c018d524abec2cd8e2eeab338dfd5a256f3ed182 — the "Approach B" label is gone, section now just describes the resolver seam.

mcpgateway/services/gateway_service.py
msureshkumar88: Done in c018d524abec2cd8e2eeab338dfd5a256f3ed182 — moved the `TokenExchangeCache` import to the module top-level in both `gateway_service.py` and `tool_service.py` (no circular dependency, so the inline import wasn't needed).

msureshkumar88:
## Manual Test Plan — RFC 8693 Token Exchange (#5135)

### Prerequisites

- ContextForge running locally (`make dev` or `make serve`, port 4444)
- `STRUCTURED_LOGGING_DATABASE_ENABLED=true` in `.env` for audit tests
- Redis available (or rely on in-memory fallback — both paths should work)
- Keycloak or a mock OAuth AS for token-exchange tests

---

### 1. Environment Setup

#### 1a. Keycloak (real AS)

```bash
docker run -p 8080:8080 \
  -e KEYCLOAK_ADMIN=admin -e KEYCLOAK_ADMIN_PASSWORD=admin \
  quay.io/keycloak/keycloak:latest \
  start-dev --features=token-exchange
```

- Create realm `cf`
- Create confidential client `token-exchange` with **Standard token exchange** enabled; note client secret
- Configure realm to trust ContextForge as a JWT issuer (shared `JWT_SECRET_KEY` or federated SSO)
- Note a `target_audience` value (another client ID or custom string)

#### 1b. Tokens

```bash
export JWT_SECRET_KEY=<your-secret>

export ADMIN_JWT=$(python -m mcpgateway.utils.create_jwt_token \
  --username admin@example.com --exp 60 --secret $JWT_SECRET_KEY)

export USER_JWT=$(python -m mcpgateway.utils.create_jwt_token \
  --username user@example.com --exp 60 --secret $JWT_SECRET_KEY)
```

---

### 2. Gateway Registration — Positive Path

```bash
curl -s -X POST http://localhost:4444/gateways \
  -H "Authorization: Bearer $ADMIN_JWT" \
  -d '{
    "name": "downstream-mcp",
    "url": "https://downstream.example.com/mcp",
    "auth_type": "oauth",
    "oauth_config": {
      "grant_type": "token-exchange",
      "token_url": "http://localhost:8080/realms/cf/protocol/openid-connect/token",
      "client_id": "token-exchange",
      "client_secret": "<keycloak-client-secret>",
      "target_audience": "<target-audience>",
      "subject_token_source": "inbound_user_jwt"
    }
  }'
```

**Expected:** `201` with gateway record; `oauth_config.grant_type` = `"token-exchange"` in response.

---

### 3. Gateway Registration — Validation / Negative Checks

| Test | Payload change | Expected result |
|---|---|---|
| Missing `target_audience` | Omit `target_audience` from `oauth_config` | `400`/`422`; error mentions `target_audience` |
| SSRF `token_url` | `"token_url": "http://169.254.169.254/latest/meta-data/"` | Rejected; SSRF guard fires |
| Private-range `token_url` | `"token_url": "http://10.0.0.1/token"` | Rejected |
| Invalid `subject_token_source` | `"subject_token_source": "magic_token"` | `422`; error names `subject_token_source` |
| Missing `token_url` | Omit `token_url` | `400`/`422` |
| Wrong `grant_type` casing | `"grant_type": "Token-Exchange"` | `422` (enum is case-sensitive) |

---

### 4. Token Exchange — Happy Path (Tool Invocation)

1. Route a tool through the registered gateway (add gateway to a virtual server, bind a tool).
2. Call the tool with the **user** JWT:

```bash
curl -s -X POST http://localhost:4444/rpc \
  -H "Authorization: Bearer $USER_JWT" \
  -d '{"jsonrpc":"2.0","id":1,"method":"tools/call","params":{"name":"<tool-name>","arguments":{}}}'
```

**Verify:**
- Call succeeds (`200`, valid JSON-RPC result)
- Upstream (mock/httpbin/request-logger) receives an `Authorization: Bearer <exchanged-token>` **different** from `$USER_JWT`
- The raw `$USER_JWT` never appears in upstream request headers

---

### 5. Caching Behavior

5a. **Cache hit:** repeat the identical call (same user, same gateway). Confirm Keycloak token endpoint is **not** hit a second time (check Keycloak admin → Events, or network-level logs).

5b. **Cache miss after expiry:** use a short-lived Keycloak client (`accessTokenLifespan` = 30 s). Wait for expiry, call again — confirm a fresh exchange occurs (new token in upstream `Authorization`).

5c. **Negative cache:** point `token_url` at a mock that always returns `500`. Make two rapid calls within 2 seconds. Confirm the AS is only called **once** (second call hits negative-cache, returns error immediately without a second AS request).

5d. **Single-flight (concurrent requests):** fire 5 simultaneous tool calls for the same user/gateway with no cached entry. Confirm the Keycloak token endpoint is called **once**, not five times.

---

### 6. 401 Invalidation + Retry (REST Integration)

1. Point the gateway `url` at a mock upstream that returns `401` on the first call, then `200` on the second.
2. Invoke a tool routed through this gateway.
3. **Expected:** tool call **succeeds** (one retry after cache invalidation + re-exchange); no error surfaced to the caller.
4. Verify cache was invalidated: a subsequent call (within normal TTL) triggers a fresh exchange, not a cache hit for the old (invalid) token.

---

### 7. Passthrough Header Hardening

1. Configure the gateway with passthrough headers that include `authorization: Bearer faketoken`.
2. Call a tool through this gateway with `$USER_JWT`.
3. **Expected:** upstream receives the **exchanged** token as `Authorization`, never `faketoken` and never `$USER_JWT`.

---

### 8. Audit Trail

With `STRUCTURED_LOGGING_DATABASE_ENABLED=true`:

1. Perform a successful token exchange (step 4 above).
2. Query `StructuredLogEntry` (via admin API or DB query) for `is_security_event=True` rows.
3. **Expected fields present:** `user_email`, `gateway_id`, `target_audience`, `success=true`, `exchanged_token_expires_in`, `correlation_id`.
4. **Expected fields absent:** `access_token`, `subject_token`, `client_secret` — no raw token material in any log row.
5. Perform a **failed** exchange (bad `client_secret`). Confirm `success=false` audit row is written, still no token material.

---

### 9. Health-Check / Passthrough Path (No User Context)

Gateway health-check and internal ping paths have **no per-request user JWT**:

- `GET /gateways/<id>/health` (or equivalent health-check endpoint)
- **Expected:** returns health result without attempting token exchange (fails closed — `_resolve_token_exchange_header()` skips exchange when no user context is available)
- No `500` or unhandled exception; logs a clear warning if applicable

---

### 10. `subject_token_source: user_oauth_token` (Secondary Path)

If the user has a stored per-gateway OAuth token (e.g. set via the OAuth callback flow):

1. Register gateway with `"subject_token_source": "user_oauth_token"`.
2. Invoke a tool as that user.
3. **Expected:** the stored OAuth token (not the CF JWT) is sent to the AS as `subject_token`.
4. Verify the correct `subject_token_type` is sent (`urn:ietf:params:oauth:token-type:access_token` vs `urn:ietf:params:oauth:token-type:jwt`).

---

### 11. Non-`token-exchange` Gateways — No Regression

- Existing gateway with `auth_type: "bearer"` — tool invocation unchanged; no new headers, no exchange attempted.
- Existing gateway with `auth_type: "oauth"` + `grant_type: "client_credentials"` — client-credentials flow unchanged.
- Gateway with no `auth_type` (`passthrough`) — unchanged.

---

### 12. Automated Coverage (CI)

```bash
# Unit + integration (mock AS)
pytest tests/unit/mcpgateway/services/test_token_exchange_cache.py \
       tests/unit/mcpgateway/services/test_token_exchange_integration.py \
       tests/e2e/test_token_exchange_e2e.py \
       tests/integration/test_token_exchange_mock_upstream.py -v

# Keycloak integration (requires live Keycloak)
KEYCLOAK_URL=http://localhost:8080 \
KEYCLOAK_CLIENT_ID=token-exchange \
KEYCLOAK_CLIENT_SECRET=<secret> \
KEYCLOAK_USER_JWT=$USER_JWT \
KEYCLOAK_TARGET_AUDIENCE=<audience> \
pytest tests/integration/test_oauth_token_exchange_keycloak.py -v
```

---

### Pass Criteria

- [ ] Gateway registers with `grant_type: "token-exchange"` — `201`
- [ ] All negative/validation checks return expected error codes and messages
- [ ] Upstream receives exchanged token, never the user's CF JWT
- [ ] Cache hit confirmed on repeat call (no second AS request)
- [ ] Cache miss triggers fresh exchange after TTL expiry
- [ ] Single-flight: concurrent requests hit AS only once
- [ ] `401` from upstream triggers cache invalidation + one retry; call succeeds
- [ ] Passthrough `Authorization` header overridden by exchanged token
- [ ] Audit rows contain required fields; no raw token material in any row
- [ ] Health-check path fails closed (no exchange, no crash)
- [ ] Non-`token-exchange` gateways unaffected
- [ ] All automated tests pass (`pytest` green, `make ruff bandit` clean)

prakhar-singh1928:
@msureshkumar88 a few items to address

1. The GatewayService token-exchange methods have no real call sites.

`_resolve_token_exchange_header `and `_send_with_token_exchange_retry ` are both fully implemented on `GatewayService`, both have tests, and the docstring on `_send_with_token_exchange_retry `itself even says:

"Not currently wired into a forward path: GatewayService has no per-request outbound forwarding analogous to ToolService's tool invocation."

So we're shipping two async methods that are never called in production, with tests that give the impression they're covered. The health-check path correctly fails closed for token-exchange gateways, and _initialize_gateway correctly defers that part is fine. But the resolver methods sitting there unused is a trap for whoever next touches this code. They'll assume these are wired and tested end-to-end, but they're not.


2. `user_oauth_token ` as a subject_token_source passes config validation on GatewayService but is silently ignored at runtime.

_validate_token_exchange_config accepts both "inbound_user_jwt" and "user_oauth_token" as valid values _VALID_SUBJECT_TOKEN_SOURCES includes both. But GatewayService._resolve_token_exchange_header never reads subject_token_source at all. It always calls extract_inbound_bearer(request_headers) regardless of what the config says.

So if someone registers a gateway with "subject_token_source": "user_oauth_token", the config saves without error, but at runtime they're silently getting their inbound CF JWT used as the subject token instead of their stored upstream OAuth token. That's a completely different token from a different issuer — the exchange will probably fail at the AS with a confusing error, and the operator has no idea why because the config looked valid.


Two things worth fixing before merge but not blockers:

3`. token_exchange()` in OAuthManager bypasses `_prepare_runtime_credentials`.

Every other grant type routes through _prepare_runtime_credentials, which handles decryption, runs `validate_core_url `on token_url, and normalises things. The `token_exchange`() method instead has an inline decrypt block for client_secret only. That means token_url isn't re-validated at runtime only at config-save time. It's also a maintenance problem: if someone later adds a new credential backend (vault, KMS, whatever), they'll update `_prepare_runtime_credentials `and wonder why token-exchange is broken.

4. The `get_access_token() `token-exchange branch doesn't forward ca_certificate, client_cert, or client_key.

Look at the client_credentials and password flows they all pass those three params through to the underlying request. The token-exchange branch at line 260 calls `self.token_exchange` without any of them, and `token_exchange()` doesn't even accept those parameters. If someone is running a token-exchange gateway against an IdP with a custom CA or mutual TLS, this will silently fail in a non-obvious way. Given that token-exchange is specifically designed to connect to external IdPs (which are exactly the kind of thing that might live behind a corporate CA, this feels like it should be fixed now.

prakhar-singh1928:
One more thing I'd flag on top of the code fixes after making those changes it's worth running through the manual test plan again end-to-end with a real Keycloak, because two of the cases in it need to be extended and one important case is missing altogether.

Steps 3, 4, and 5 in your plan are solid happy path exchange, cache reuse, and the 401 retry. Those cover what I'd want to see and I'd just run them as-is after the fixes land.

Step 2's negative checks need one addition. Right now it tests omitting target_audience, setting a link-local token_url, and setting subject_token_source to an invalid string. But there's a specific case not in there: what happens when you set subject_token_source to "user_oauth_token" —which is a valid enum value on a gateway. Before your fix that would silently accept it and then use the wrong token at runtime. After your fix it should come back with a clear error at create time. That case should be in the plan explicitly: set subject_token_source: "user_oauth_token", expect a 400 saying it's not supported on the gateway path.

There's also a step completely missing from the plan: gateway registration itself. The last commit in this PR (82bef40) was a fix for _initialize_gateway unconditionally 502-ing when you tried to create any token-exchange gateway. That's a real regression that was caught and fixed, but there's no manual step that covers it. Add a step: do POST /gateways with a valid token-exchange config, confirm you get a 201 back with an empty tool list (not a 502), and confirm the gateway shows up in GET /gateways. Sounds obvious but given this exact thing broke once already it's worth having it written down as a step you actually run.

msureshkumar88:
### Follow-up: 4 pre-merge blockers resolved (commit edeb6b4)

Following the review feedback and a deeper audit of the gateway-path code, I found and fixed four issues before merge:

**1. Dead GatewayService helpers removed**
`_resolve_token_exchange_header`, `_send_with_token_exchange_retry`, and `_TOKEN_EXCHANGE_FALLBACK_TTL` had zero callers — all live token-exchange traffic runs through ToolService. Keeping them would have been a maintenance trap (future readers would wonder why two parallel paths existed). Removed cleanly.

**2. `_VALID_SUBJECT_TOKEN_SOURCES` narrowed to `inbound_user_jwt` only**
`user_oauth_token` was previously accepted at config-save time but silently ignored at runtime because GatewayService never implemented that path. Now the validator rejects it with a clear error rather than letting the mismatch surface as a confusing runtime failure.

**3. TE branch now routes through `_prepare_runtime_credentials`**
The `token-exchange` branch in `get_access_token` was reaching into `credentials` directly, bypassing the `_prepare_runtime_credentials` step that decrypts stored secrets and validates `token_url` at runtime. Fixed — TE now gets the same runtime-validation treatment as every other grant type.

**4. mTLS / CA certificate params forwarded into `token_exchange()`**
`ca_certificate`, `client_cert`, and `client_key` were accepted by `get_access_token` but not forwarded into `token_exchange()` → `_post_token_request`. Corporate-CA IdPs (mutual TLS) would have silently fallen back to the default TLS context. Parameters now flow through the full call chain.

Tests for fixes 3 and 4 added in the same commit; `TestGatewayPathTokenExchange` (which tested the now-deleted GatewayService helpers) removed.

msureshkumar88:
### Runtime verification: 4 pre-merge blockers (commit edeb6b4)

Ran the app and drove each changed path to confirm behavior at the surface. Results below.

---

**Method:** Live server (uvicorn :18765) + mock token AS (HTTP :19877 that captured raw request bodies).

#### Step 1 — `user_oauth_token` source rejected at registration (blocker 2)

```
POST /gateways  {subject_token_source: "user_oauth_token"}
→ HTTP 400  {"message":"Unable to process input"}
```

Config validator rejects it immediately. ✅

#### Step 2 — `inbound_user_jwt` source accepted (blocker 1 + 2)

```
POST /gateways  {subject_token_source: "inbound_user_jwt"}
→ config validation passed; failed only at MCP transport (unreachable mock URL)
```

No validation error → dead helpers are gone, valid source accepted. ✅

#### Step 3 — `_prepare_runtime_credentials` called in TE branch (blocker 3)

`get_access_token` with TE credentials against real mock AS. Mock AS captured the RFC 8693 POST body:

```
grant_type:            urn:ietf:params:oauth:grant-type:token-exchange
subject_token:         eyJhbGciOiJSUzI1NiJ9...
subject_token_type:    urn:ietf:params:oauth:token-type:jwt
requested_token_type:  urn:ietf:params:oauth:token-type:access_token
client_id:             contextforge
audience:              aud-service
```

Returned `access_token=EXCHANGED-TOKEN-FROM-AS`. Call succeeded end-to-end. ✅

#### Step 4 — `ca_certificate` forwarded to TLS layer (blocker 4)

Passed `ca_certificate="NOT-VALID-PEM-CERT"`. If silently dropped, call would succeed against the mock. Instead:

```
SSLError: no start line: cadata does not contain a certificate (_ssl.c:4014)
```

Param reaches the TLS context builder — confirms it flows through `token_exchange()` → `_post_token_request`. ✅

---

**Probes:**
- No `subject_token_source` field → defaulted to `inbound_user_jwt`, server reached MCP connection layer (not 400). ✅
- TE config with missing `token_url` → HTTP 400. ✅
- Empty `client_secret` in TE config → clean success; RFC 8693 doesn't require one. ✅

**Note on full gateway registration:** The SSE MCP handshake (`GET /sse 200`, `POST /message 202`) established but the `initialize` round-trip didn't complete in the manual test env (inline mock stdio didn't handle `notifications/initialized`). Test env limitation only — no regression in real deployments.

prakhar-singh1928:
all four are correctly addressed. One new issue surfaced from Fix 3.

 **Double-decrypt in token_exchange()**

_prepare_runtime_credentials fully decrypts client_secret before calling token_exchange(). But token_exchange() still has its own inline decrypt block at lines 811–820 that runs is_encrypted() on the already-plaintext value.

In practice today this doesn't corrupt anything — decrypt_secret_or_plaintext is idempotent and returns plaintext unchanged when the input doesn't look encrypted. But the guard it uses (is_encrypted) has this in its own docstring:

So if an operator's actual client_secret value happens to match the encryption heuristic (e.g. starts with v2:, or is structured JSON), is_encrypted returns True, decrypt is attempted on plaintext, fails, and returns None. The token request then goes out with an empty client_secret and the AS returns a 401 with no useful error.

The complication is that token_exchange() is called from two places with different preconditions:

From get_access_token → client_secret is already plaintext (decrypted by _prepare_runtime_credentials)
From ToolService._resolve_token_exchange_header directly → client_secret is the raw encrypted DB value, so the inline decrypt is genuinely needed there
Removing the inline block entirely breaks the second path. The clean fix is a boolean parameter:

```
async def token_exchange(self, ..., client_secret_is_plaintext: bool = False):
    if client_secret and not client_secret_is_plaintext:
        # existing inline decrypt block

```
Then get_access_token passes client_secret_is_plaintext=True, and _resolve_token_exchange_header keeps the default False. No other callers are affected.

msureshkumar88:
Thanks @prakhar-singh1928 — good catch on the double-decrypt, and you're right that the two callers have different preconditions. I've added the `client_secret_is_plaintext` flag exactly as you suggested (commit `edc7767`):

- `get_access_token` → `token_exchange(..., client_secret_is_plaintext=True)` since `_prepare_runtime_credentials` already decrypted it.
- `ToolService._resolve_token_exchange_header` keeps the default `False`, so its raw encrypted `gateway.oauth_config` value still gets decrypted inline.

One clarification on the failure mode, just so it's on the record: the empty-`client_secret` → 401 scenario doesn't actually trigger on the current code, because the inline block guards the assignment:

```python
decrypted = await encryption.decrypt_secret_async(client_secret)
if decrypted:               # guard
    client_secret = decrypted
```

If `is_encrypted()` false-positives on a plaintext secret, `decrypt_secret_async` returns `None`, the guard skips the assignment, and the original plaintext is preserved — so the secret was never blanked. So this change is really a redundancy/maintainability cleanup rather than a correctness fix. Still worth doing for the reasons you gave (wasted work + the trap for the next maintainer), so it's in.

Added two tests covering it:
- `get_access_token` sets `client_secret_is_plaintext=True`.
- with the flag set, a plaintext secret that looks like a `v2:` bundle reaches the AS unchanged and the decrypt path is never touched.

Full token-exchange suite green (42 passing), and the end-to-end exchange flow (registration → `/mcp` tools/call → RFC 8693 request to the AS → exchanged token forwarded downstream, raw user JWT not leaked) still passes.

prakhar-singh1928:
**1. Add a one-line comment at the ToolService call site explaining the decrypt precondition**

tool_service.py ~line 3836 calls token_exchange() without client_secret_is_plaintext=True. That's correct — it's passing the raw encrypted DB value — but there's nothing at the call site explaining why the flag is omitted. The next maintainer who sees get_access_token setting client_secret_is_plaintext=True and wonders why _resolve_token_exchange_header doesn't will either assume it's a bug or cargo-cult the flag incorrectly. One comment removes the ambiguity:

client_secret=oauth_config.get("client_secret", ""),
- Raw encrypted DB value — token_exchange() decrypts inline.
-  (client_secret_is_plaintext defaults to False; only get_access_token
- sets it True because _prepare_runtime_credentials already decrypted it.)

**2. Add a skipif guard to the Keycloak integration test**

tests/integration/test_oauth_token_exchange_keycloak.py reads KEYCLOAK_URL and friends without a skip guard. In CI without a live Keycloak instance the test raises a connection error or KeyError — not a clean SKIPPED. Standard fix at the top of the file:
```
import os
import pytest


pytestmark = pytest.mark.skipif(
    not os.getenv("KEYCLOAK_URL"),
    reason="Keycloak not configured (set KEYCLOAK_URL to run)",
)
```

**3. Move the three remaining inline imports in _resolve_token_exchange_header to module top-level**

Lines ~3770–3772 still import inside the method body:

from mcpgateway.services.token_storage_service import TokenStorageService
from mcpgateway.utils.subject_token import extract_inbound_bearer, looks_like_jwt
from mcpgateway.utils.token_exchange_audit import audit_token_exchange

None of these create a circular dependency — TokenExchangeCache was already moved to top-level in the earlier review fix and the same applies here. Moving them removes the three # pylint: disable=import-outside-toplevel suppressions and keeps the file consistent.

msureshkumar88:
**Item 2 (Keycloak skipif guard):** Already present — stale comment.

`pytestmark` with the `skipif` guard exists at line 52–54 of `tests/integration/test_oauth_token_exchange_keycloak.py`:

```python
pytestmark = [
    pytest.mark.skipif(not os.getenv("KEYCLOAK_URL"), reason="KEYCLOAK_URL not configured"),
]
```

**Items 1 and 3** are fixed in the next commit — call-site comment added and all three inline imports in `_resolve_token_exchange_header` moved to module top-level (also cleaned up the two other stray `TokenStorageService` inline imports in the same file as a bonus).

prakhar-singh1928 (dismissed):
All comments are addressed

looks good to me.

Lang-Akshay (changes requested):
Thanks allot for the PR @msureshkumar88 
Please address this blocking issue (must fix before merge):

1. RBAC gap (gateway_service.py) — developer role can register a token-exchange gateway pointing at attacker-controlled token_url, harvesting other users' JWTs as subject_token. Add a platform-admin gate when grant_type == "token-exchange".
2. Cache key collision (token_exchange_cache.py:~80) — bare : delimiter lets a crafted audience URL collide across {gateway, user, audience} tuples → wrong principal gets a cached token. Hash the tuple instead.
3. Dead code path live (tool_service.py:~3804) — user_oauth_token branch is reachable at runtime but not in _VALID_SUBJECT_TOKEN_SOURCES. Add validation or remove the branch.

msureshkumar88:
Addressed all three:

1. **RBAC gap (gateway_service.py)** — added `GatewayService._enforce_token_exchange_admin_only`, a single choke point both `register_gateway` and `update_gateway` route through before persisting a `grant_type: token-exchange` oauth_config. It requires the requester to hold the platform-admin wildcard permission (`check_permission(..., "*", allow_admin_bypass=True)`), so `developer`/`team_admin` roles can no longer point `token_url` at attacker-controlled infrastructure. Trusted internal callers with no request-scoped identity (config import — already gated behind `admin.import`, platform-admin-only; catalog registration, which applies bundled static definitions) skip the gate rather than being denied. Wired the corresponding 403 mapping into the two create-path routers (`main.py::register_gateway`, `admin.py::admin_add_gateway`) that didn't already catch `PermissionError`.

2. **Cache key collision (token_exchange_cache.py)** — `_key()` now hashes each of `gateway_id`/`user_email`/`audience` independently (SHA-256) before joining, instead of a bare `:`-joined string. A hex digest can never contain the join delimiter, so no combination of values (e.g. a crafted `target_audience` containing `:`) can shift the component boundary and collide with a different tuple.

3. **Dead code path (tool_service.py)** — removed the `subject_token_source == "user_oauth_token"` branch. `GatewayService._VALID_SUBJECT_TOKEN_SOURCES` only ever persists `"inbound_user_jwt"` (config-time validation rejects anything else), so the branch was unreachable through any validated config path.

Added regression tests for all three (RBAC gate incl. admin/non-admin/internal-caller cases, cache key non-collision, and removed the now-obsolete `user_oauth_token` test). Rebased onto latest main.

Lang-Akshay (changes requested):
1. Health check permanently kills token-exchange gateways
gateway_service.py:4439-4449 — health check has no subject_token, so every check fails → failure counter → reachable=False. Discovery path (4804-4813) correctly returns early. Health check must mirror it.

2. Raw AS response leaks into OAuthError (CWE-532)
oauth_manager.py:851:
raise OAuthError(f"No access_token in token exchange response: {token_response}")
token_response is the raw dict — may contain token material. All sibling flows use _redact_token_response(). Fix: parse + redact before interpolating.

3. exc_info=True re-surfaces unredacted chain
tool_service.py:3867 — even if #2 fixed, exc_info=True captures full traceback including unredacted AS body into log record. Remove exc_info=True.

msureshkumar88:
Addressed all three:

1. **Health check permanently kills token-exchange gateways** — the `token-exchange` branch in `_check_single_gateway_health` (gateway_service.py) called `_handle_gateway_failure()` on every periodic check, since there's never a subject token outside a user request. Now mirrors the discovery path (`_prepare_gateway_registration`'s connection probe), which already skips the probe rather than failing when it can't obtain one — `return` without marking the gateway unhealthy.

2. **Raw AS response leaks into OAuthError (CWE-532)** — `oauth_manager.py`'s `token_exchange()` now wraps `token_response` in `self._redact_token_response(...)` before interpolating into the "No access_token" error, matching every sibling flow (`get_access_token`, password grant, etc.) that already does this.

3. **exc_info=True re-surfaces unredacted chain** — removed `exc_info=True` from the `tool_service.py` failure-path warning log. The traceback's final line renders `str(exception)` unredacted regardless of how sanitized the `safe_reason` message argument is, so it was bypassing the redaction entirely.

Added/updated regression tests for all three (health-check skip-not-fail, redacted-response assertion, and an exc_info-absence check on the log call). Pushed.

Lang-Akshay (changes requested):

## Would be great to have following security hardening in place before merging. Thanks again. 

## ✅ Finding 1: Manual refresh forwards raw inbound auth - VALID (HIGH)
- **Root Cause:** `pre_auth_headers` parameter bypasses token-exchange logic in `_initialize_gateway`
- **Impact:** Credential leakage - user's ContextForge JWT exposed to upstream MCP server
- **Evidence:** `gateway_service.py:6170, 5920, 4750`

## ✅ Finding 2: Token-exchange gateways cannot populate tools - VALID (HIGH)
- **Root Cause:** No authenticated refresh path exists; early return prevents tool discovery
- **Impact:** Service degradation - token-exchange gateways remain permanently empty
- **Evidence:** `gateway_service.py:4790-4800`

## ✅ Finding 3: Live invocation drops custom CA/mTLS - VALID (MEDIUM)
- **Root Cause:** `_resolve_token_exchange_header` doesn't accept/forward TLS parameters
- **Impact:** TLS validation bypass, mTLS authentication failures
- **Evidence:** `tool_service.py:3820-3870`

## ✅ Finding 4: 401 retry loses original headers - VALID (MEDIUM)
- **Root Cause:** Retry sends only Authorization header, drops correlation IDs and passthrough headers
- **Impact:** Distributed tracing breakage, request corruption
- **Evidence:** `tool_service.py:3930-3980`

msureshkumar88:
Thanks for the detailed review, @Lang-Akshay — all four findings check out and are fixed in 95fdb64a7 (rebased/carried forward as it stands now).

**Finding 1 (manual refresh forwards raw inbound auth) & Finding 2 (token-exchange gateways can't populate tools)** — fixed together. `refresh_gateway_manually` no longer builds `pre_auth_headers` via generic passthrough for `grant_type == "token-exchange"` gateways. It now calls a new `GatewayService._resolve_token_exchange_header`, which performs a real RFC 8693 exchange of the caller's inbound JWT (cache, negative-cache, single-flight, audit — mirrors `ToolService`'s resolver). A manual refresh either populates tools with a properly exchanged token or fails closed with `success: False` and a clear error — it never forwards the raw JWT and never silently no-ops.

**Finding 3 (CA/mTLS dropped on live invocation)** — `ToolService._resolve_token_exchange_header` and the new `GatewayService` counterpart now accept `ca_certificate`/`client_cert`/`client_key` and forward them to `OAuthManager.token_exchange`, matching what the `client_credentials` grant already does on the same gateway.

**Finding 4 (401 retry loses original headers)** — `_send_with_token_exchange_retry` now merges the freshly exchanged `Authorization` into the original headers (`{**headers, **fresh}`) instead of replacing them, so Content-Type, header_mapping output, and session-affinity headers survive the retry.

Added regression coverage for all four in `test_gateway_service.py` and `test_token_exchange_integration.py` (resolver unit tests on both services, CA/mTLS passthrough assertions, header-merge-on-retry assertion, and `refresh_gateway_manually` tests proving passthrough is never used for token-exchange gateways and failures short-circuit cleanly). Full existing suite plus the new tests are green.

mcpgateway/services/tool_service.py
```diff
@@ -3735,6 +3740,274 @@ async def invoke_tool_direct(
...
+            client_key: Optional client private key for mTLS to the token endpoint.
+
+        Returns:
+            A dict with a single ``Authorization`` header carrying the exchanged token.
+
+        Raises:
+            ToolInvocationError: If no usable subject token exists or the exchange fails.
+        """
+        audience = oauth_config.get("target_audience")
+        user_key = app_user_email or ""
```
Lang-Akshay: [Blocking] Empty app_user_email collapses cache keys — unrelated users can receive each other's delegated tokens

The caching design here is solid — keying by (gateway_id, user, audience) with single-flight locking and a negative cache is exactly the right shape for an On-Behalf-Of flow. There's one gap in the key construction worth closing before merge.

What's wrong

user_key = app_user_email or ""

If a caller reaches this resolver without a resolved email, every such caller shares the single cache bucket (gateway_id, "", audience). Because the cache lookup below runs before the subject token is extracted from the request, the failure sequence is:

1. User A arrives with app_user_email=None → exchange runs on A's inbound JWT → A's delegated token is cached under key "".
2. User B arrives, also without an email → cache hit on the same key → B is handed A's delegated token.

That's cross-user identity confusion on the exact property RFC 8693 token exchange exists to guarantee (the exchanged token is scoped to a specific subject). Email should normally be present for authenticated callers, so this is an edge case — but the blast radius if it triggers is token leakage between principals.

Proposed fix

Fail closed before any cache interaction:

# Fail closed: cache key is (gateway_id, user, audience). Without a user
# identity there is no "behalf" to act on, and an empty key component
# would let unrelated principals share one delegated token.
if not app_user_email:
    raise ToolInvocationError("Token exchange requires an authenticated user identity. Contact your administrator.")

user_key = app_user_email

Two notes on the shape of the fix:

- The raise is the fix, not the removal of or "". Dropping the fallback alone would just pass None into the cache key — either a TypeError or a stringified "None" bucket with the identical sharing problem.
- The guard must sit before the first cache.get — guarding later would still let an anonymous request read a token cached by another empty-email caller.

Related (non-blocking)

- _invalidate_token_exchange_on_unauthorized also uses app_user_email or ""; harmless once nothing can be cached under "", but worth dropping for consistency.
- Per the deny-path testing invariant in AGENTS.md, a small test calling the resolver with app_user_email=None / "", asserting ToolInvocationError and that cache.get was never invoked, would pin the "guard runs before any lookup" property against regression.

msureshkumar88: Thanks for the sharp catch, and for spelling out the exact failure sequence — that made this an easy fix to verify.

Fixed in 9c1c20ee5: `_resolve_token_exchange_header` now raises `ToolInvocationError` before any cache interaction when `app_user_email` is falsy, so there's no more shared `""` bucket to collide on. Applied your note on `_invalidate_token_exchange_on_unauthorized` too — it's now a no-op when there's no user email, since nothing is ever cached under an empty key anymore.

Added regression tests (`test_empty_app_user_email_denies_before_cache_lookup`, parametrized over `None`/`""`, and `test_invalidate_on_unauthorized_noop_without_user_email`) asserting the guard fires before `cache.get`/`is_failed` are ever awaited, per your suggested pin.

Lang-Akshay (changes requested):
## Finding 1 — `X-Upstream-Authorization` bypasses the exchanged token

**File:** `mcpgateway/utils/passthrough_headers.py`

After `_resolve_token_exchange_header()` sets `headers["Authorization"]` to the freshly-exchanged token, both tool-invocation paths call `compute_passthrough_headers_cached()`. That function's `X-Upstream-Authorization` rename block (lines 392–403) runs **before** the allow-list and unconditionally overwrites `Authorization`. The existing `_sanitize_passthrough_for_token_exchange()` guard only strips `authorization` from the allow-list — it never reaches this earlier branch. A caller can send `X-Upstream-Authorization: Bearer <anything>` to replace the exchanged token entirely.

**Fix — add an explicit token-exchange guard to `compute_passthrough_headers_cached()`:**

```python
# passthrough_headers.py
def compute_passthrough_headers_cached(
    ...,
    is_token_exchange: bool = False,   # new
) -> Dict[str, str]:
    ...
    if upstream_auth and not is_token_exchange:   # guard added
        passthrough_headers["Authorization"] = sanitized_value
```

Pass the flag only on the gateway-backed token-exchange merge paths, where `headers["Authorization"]` was produced by `_resolve_token_exchange_header()`:

```python
# tool_service.py:4392
headers = compute_passthrough_headers_cached(
    ...,
    is_token_exchange=(gateway_grant_type == "token-exchange"),
)

# tool_service.py:5626
headers = compute_passthrough_headers_cached(
    ...,
    is_token_exchange=(gateway_grant_type == "token-exchange"),
)
```

Do not rely on `_sanitize_passthrough_for_token_exchange()` for this case. That helper only removes `authorization` from configured passthrough allow-lists; it does not affect the special `X-Upstream-Authorization` rename branch.

Add a regression test proving that when `base_headers = {"Authorization": "Bearer exchanged"}`, inbound headers include `X-Upstream-Authorization: Bearer attacker`, and `is_token_exchange=True`, the final `Authorization` remains `Bearer exchanged`.

---

## Finding 2 — Resource invocation broken for token-exchange gateways

**File:** `mcpgateway/services/resource_service.py`

`invoke_resource()` has no `elif grant_type == "token-exchange":` branch. That grant type falls into the `else` (client-credentials) path at line 1912 and calls `oauth_manager.get_access_token()` without a `subject_token`. The PR made `subject_token` mandatory for this grant — `OAuthManager` raises immediately. Resources on token-exchange gateways always fail. `invoke_resource()` also has no `request_headers` parameter, so there is no inbound bearer to extract.

**Fix — propagate request headers and add a token-exchange branch:**

**Step 1 — add `request_headers` to `read_resource()` and `invoke_resource()`:**

```python
# resource_service.py:2139
async def read_resource(
    self,
    db: Session,
    ...,
    request_headers: Optional[Dict[str, str]] = None,  # new
) -> Union[ResourceContent, ResourceContents]:
    ...

# resource_service.py:1595
async def invoke_resource(
    self, db, resource_id, resource_uri, ...,
    request_headers: Optional[Dict[str, str]] = None,  # new
) -> Any:
```

Then pass `request_headers=request_headers` through every internal `read_resource()` -> `invoke_resource()` call. The streamable HTTP MCP path already has inbound headers available at `streamablehttp_transport.py:2571`; pass those into `resource_service.read_resource(...)`.

**Step 2 — add the `token-exchange` branch before the client-credentials fallback:**

```python
# resource_service.py ~line 1887
if grant_type == "authorization_code":
    ...  # unchanged
elif grant_type == "token-exchange":
    from mcpgateway.utils.subject_token import extract_inbound_bearer, looks_like_jwt
    subject_token = extract_inbound_bearer(request_headers or {})
    if subject_token and not looks_like_jwt(subject_token):
        subject_token = None
    if not subject_token:
        logger.error("Token exchange requires an authenticated user JWT for gateway %s", gateway_name)
    else:
        try:
            access_token = await self.oauth_manager.get_access_token(
                gateway_oauth_config,
                ca_certificate=gateway.ca_certificate,
                client_cert=gateway.client_cert,
                client_key=gateway.client_key,
                subject_token=subject_token,   # required
            )
            headers["Authorization"] = f"Bearer {access_token}"
        except Exception as e:
            logger.error("Token exchange failed for gateway %s: %s", gateway_name, e)
else:
    ...  # client_credentials unchanged
```

Keep this fail-closed: if the inbound bearer is missing or is not a JWT, do not call the upstream gateway with no auth or with client-credentials fallback. Also mirror the existing token-exchange logging hygiene from the tool path: sanitize exception text and do not log the raw subject token or exchanged token.

Add a regression test that creates a resource on an OAuth gateway with `grant_type: "token-exchange"`, supplies an inbound bearer in `request_headers`, and asserts `oauth_manager.get_access_token(..., subject_token=<inbound jwt>)` is called and the upstream resource read receives `Authorization: Bearer <exchanged>`.

msureshkumar88:
Both findings fixed in ab74ed1bb.

**Finding 1 (X-Upstream-Authorization bypass):** `compute_passthrough_headers_cached()` now takes `is_token_exchange`, which guards the X-Upstream-Authorization rename block and the `auth_type="none"` passthrough from overwriting the exchanged `Authorization` header. Wired at both `tool_service.py` call sites (gated on `gateway_grant_type == "token-exchange"`), exactly as you'd sketched — `_sanitize_passthrough_for_token_exchange()` is left untouched since it only ever governed the allow-list path. Added the regression test you asked for: `base_headers={"Authorization": "Bearer exchanged"}` + inbound `X-Upstream-Authorization: Bearer attacker` + `is_token_exchange=True` → final `Authorization` stays `Bearer exchanged`.

**Finding 2 (resource invocation broken for token-exchange gateways):** Added the `token-exchange` branch to `invoke_resource()`'s grant-type dispatch — resolves the subject_token from the inbound bearer via `extract_inbound_bearer`/`looks_like_jwt`, and fails closed (no client_credentials fallback, no unauthenticated call) when there isn't one. `request_headers` is threaded through `read_resource()` -> `invoke_resource()` and wired from the streamable HTTP MCP path (`streamablehttp_transport.py:2571`, as you flagged) plus the three REST/JSON-RPC resource-read endpoints in `main.py`, which already had `request` in scope. Added both regression tests: subject_token propagation + exchanged-token forwarding upstream, and the fail-closed-without-bearer case.

Full `resource_service`, `tool_service`, `token_exchange_integration`, and `streamablehttp_transport` suites pass; `ruff` clean on all touched files.

Lang-Akshay (dismissed):
LGTM

brian-hussey (dismissed):
LGTM

Lang-Akshay (dismissed):
LGTM

ja8zyjits (approved):
LGTM rebased.

Lang-Akshay (approved):
LGTM

Issue #4877, opened 2026-05-25
[BUG]: Admin Users Cannot View/Edit Their Own Private Resources (All Endpoints Except A2A)


rakdutta:
Admin users with is_admin=true and token_teams=null cannot view or edit their own private resources. Issue #4694 only fixed A2A agents - the root cause persists in auth_context.py affecting all other resource types.

  Affected Resources
  
  - ❌ Tools (GET by ID and list)
  - ❌ Prompts (GET by ID and list)
  - ❌ Resources (GET by ID and list)
  - ❌ Gateways (GET by ID and list)
  - ❌ Servers (GET by ID and list)
  - ✅ A2A Agents (fixed in PR #4788)

  Symptoms

  Admin user (admin@example.com) with token_teams=null:
  - ✅ Can create private resources
  - ❌ Cannot view them via GET by ID (404/filtered)
  - ❌ Cannot see them in list views (empty)
  - ❌ Cannot edit them in Admin UI

  Root Cause

  File: mcpgateway/auth_context.py, line 495-496
  
  def get_scoped_resource_access_context(request: Request, user):
      # ...
      if is_admin and token_teams is None:
          return None, None  # ❌ BUG: Sets user_email=None

  Setting user_email=None breaks the PR #4341 owner-matching logic in base_service.py that allows admins to see their own private resources:

  # base_service.py line 110-124
  if user_email is None and token_teams is None:
      return query.where(model_cls.visibility != "private")  # Excludes ALL private

  # This path is never reached because user_email is None:
  if token_teams is None:
      user_is_admin = user_email and is_user_admin(db, user_email)  # ❌ False when user_email=None
      if user_is_admin:
          # Should allow: public + team + OWN private

  Expected vs Actual Behavior

  Expected (per PR #4341):
  - ✅ All public resources
  - ✅ All team resources
  - ✅ Admin's OWN private resources
  - ❌ Other users' private resources

  Actual:
  - ✅ Public and team resources
  - ❌ NO private resources (including admin's own)
  
  Proposed Fix

  Primary Fix - mcpgateway/auth_context.py line 495-496:

  # Change from:
  if is_admin and token_teams is None:
      return None, None

  # To:
  if is_admin and token_teams is None:
      return user_email, None  # Keep user_email for owner matching (PR #4341 / issue #4694)

  Also update line 492-493 (non-JWT admin path):
  if fallback_admin:
      return _requester_email, None  # Keep email for owner matching
      
[...]

PR #5239, opened 2026-06-16
fix: session token admin bypass broken in get_rpc_filter_context (#5232)

### LINKED ISSUE #5232: [BUG]: Admin UI Cannot Access Private Resources Created via API Login Token
We are observing an issue where private MCP resources created by the admin user are accessible via API/CLI operations but cannot be viewed or edited through the Admin UI.

Environment

MCP Gateway deployment
Admin user: [admin@admin.com](mailto:admin@admin.com)
Resources are created programmatically during deployment
Authentication is performed using the email login endpoint
Client integrations include Java applications (Skynet clients)

Current Flow

Generate a session token using the login endpoint:

curl -X POST
http://mcp-gateway-service:4444/auth/email/login
-H 'Content-Type: application/json'
-d '{
"email":"[admin@admin.com](mailto:admin@admin.com)",
"password":""
}'

Extract access_token from the response and use it for all gateway registration operations.
Register gateways using the generated token.

Example payload:

{
"name": "",
"url": "",
"enabled": true,
"authType": "oauth",
"visibility": "private",
"ownerEmail": "[admin@admin.com](mailto:admin@admin.com)",
"oauthConfig": {
"grant_type": "client_credentials",
"client_id": "<client_id>",
"client_secret": "<client_secret>",
"token_url": "<token_url>",
"scopes": [
"openid",
"profile",
"email"
]
},
"description": "Google GenAI Toolbox MCP Server"
}

Observed Behavior

Resources are successfully created.

Database verification shows:

owner_email = [admin@admin.com](mailto:admin@admin.com)
visibility = private
tool.team_id matches gateway.team_id
gateway.owner_email = [admin@admin.com](mailto:admin@admin.com)

Example:

Tool:

id: 61d9a486380741f18c5b2f0c966eb712
visibility: private
owner_email: [admin@admin.com](mailto:admin@admin.com)
team_id: 96454510f47549e39aa9fb787c8acefc

Gateway:

owner_email: [admin@admin.com](mailto:admin@admin.com)
team_id: 96454510f47549e39aa9fb787c8acefc

CLI/API Behavior

Using the same admin session token:

Gateway registration succeeds
Tool creation succeeds
Tool listing succeeds
Resource ownership appears correct

Admin UI Behavior

When opening the tool in the UI:

"Tool with ID 61d9a486380741f18c5b2f0c966eb712 not found. It may have been deleted."

Gateway logs show:

Tool access denied
visibility=private
admin_bypass=false

Request:

GET /admin/tools/61d9a486380741f18c5b2f0c966eb712

Response:

404 Not Found

Relevant Log Snippet

Authenticated user: [admin@admin.com](mailto:admin@admin.com)

Tool access denied
visibility=private
admin_bypass=false

GET /admin/tools/61d9a486380741f18c5b2f0c966eb712
404

Additional Findings

The JWT generated through /auth/email/login contains:

sub = bb836702-799e-4711-b8f2-bcad89c6bb9f
permissions = ["*"]
token_use = "session"

Database verification:

email_users.id = bb836702-799e-4711-b8f2-bcad89c6bb9f
email = [admin@admin.com](mailto:admin@admin.com)
is_admin = true

Therefore the JWT subject correctly resolves to the admin user.

Reason for Using Login-Based Token Generation

We are currently generating tokens through the supported login API because our downstream clients are Java-based applications.

The alternative Python-based token generation mechanism:

python -m mcpgateway.utils.create_jwt_token

is not practical for client integrations and deployment automation outside the MCP Gateway container environment.

Our expectation is that resources created using a valid admin session token obtained from /auth/email/login should remain fully manageable through both:

API/CLI
Admin UI

Questions

Is the Admin UI expected to support viewing/editing private resources created using tokens generated from /auth/email/login?
Why is admin_bypass=false for [admin@admin.com](mailto:admin@admin.com) when accessing resources that it owns?
Does the owner-matching logic introduced for issue #4877 apply to tools and gateways created through API login tokens?
Is there an additional team membership, token scope, or ownership requirement needed for private resources to be visible in the Admin UI?

Expected Behavior

An admin user should be able to view and edit their own private resources through the Admin UI when:

owner_email matches the authenticated user
the resource was created by that user
the resource team matches the user's team context

This matches the behavior described in issue #4877:

Admin can access own private resources
Admin cannot access other users' private resources
Security boundaries remain intact

MohanLaksh:
## Problem

Issue #5232 reports that an admin user (`admin@admin.com` with `is_admin=true` in the DB) cannot view their own private resources through the Admin UI, even though they created those resources using a session token from `/auth/email/login`. The error logs show `admin_bypass=false` when it should be `true`.

## Root Cause

Session tokens from `/auth/email/login` deliberately omit the `is_admin` claim from their JWT payload. This is by design — per the AGENTS.md, the DB is the authority for session tokens so revocations take effect immediately.

However, `get_rpc_filter_context()` in `auth_context.py` derives `is_admin` solely from the JWT payload at line 415:

```python
is_admin = payload.get("is_admin", False) or payload.get("user", {}).get("is_admin", False)
```

For session tokens, `is_admin` is absent from the JWT, so this always returns `False`. The function returns `(email, None, False)` instead of the correct `(email, None, True)`.

## How This Differs From PR #4878

**PR #4878** fixed issue #4877, which affected **API/legacy tokens** where the JWT had `is_admin: true` + `teams: null`, but `auth_context.py` was returning `(None, None)` instead of `(email, None)`. The fix preserved `user_email` during admin bypass so the owner-matching logic (`tool.owner_email == user_email`) could work.

**Issue #5232** affects **session tokens** from `/auth/email/login`. The session token JWT never has an `is_admin` claim (by design), so `is_admin` derives as `False` and the entire admin bypass path (line 497) is skipped. The code falls through to `return user_email, []` — public-only scope — which blocks all non-public resources.

In short:

| | #4877 (API tokens) | #5232 (Session tokens) |
|---|---|---|
| JWT has `is_admin`? | Yes (`True`) | No (by design) |
| `token_teams` source | JWT `teams: null` | DB via `resolve_session_teams()` |
| `is_admin` in `get_rpc_filter_context` | `True` (from JWT) | `False` (absent from JWT) |
| Bug | `user_email=None` breaks owner matching | `is_admin=False` blocks admin bypass entirely |
| Fix | Preserve `user_email` in return tuple | Set `is_admin=True` when session + DB bypass |

## Solution

A 5-line addition in `get_rpc_filter_context()` at `auth_context.py:420-424`:

```python
if not is_admin and token_teams is None and getattr(request.state, "token_use", None) == "session":
    is_admin = True
```

### Why it works

1. **`request.state.token_use`** is set to `"session"` in all 3 auth code paths (auth.py lines 1622, 1718, 1909)
2. **`token_teams=None`** for session tokens only happens when `resolve_session_teams()` confirmed admin from the DB — non-admin session tokens produce `[]` or a list of team IDs, never `None`
3. **No false positives**: API tokens already have `is_admin=True` from JWT (so `not is_admin` short-circuits); internal auth context returns early at line 409; unauthenticated requests have `token_teams=[]`
4. **Downstream handles it**: The existing `get_scoped_resource_access_context` line 497 then correctly returns admin bypass, and the service layer's `_check_tool_access` already handles `(email, None)` for owner-matching

### Security Invariant

Preserves the PR #4341 invariant: admin bypass never reveals another user's private resources. The owner-matching check (`tool.owner_email == user_email`) in the service layer enforces this. A stale session token (admin revoked in DB) still gets `token_teams` resolved to actual teams from the DB, so the condition `token_teams is None` is `False` and bypass is never granted.

## Testing

- **6 new tests** covering session token admin bypass, non-admin, public-only, API token regression, and end-to-end through `get_scoped_resource_access_context`
- **Full regression suite**: 18,407 tests passed, 0 failures
- **Doctests**: 1,179 passed
- **Bandit**: clean | **Interrogate**: 100% | **Ruff**: no new errors

## Files Changed

| File | Change |
|---|---|
| `mcpgateway/auth_context.py` | +5 lines: session token admin bypass check |
| `tests/unit/mcpgateway/test_main.py` | +66 lines: 4 new test methods |
| `tests/unit/mcpgateway/test_main_extended.py` | +17 lines: e2e test through `get_scoped_resource_access_context` |

Closes #5232

### DIFF
--- a/mcpgateway/auth_context.py
+++ b/mcpgateway/auth_context.py
@@ -449,11 +449,11 @@ def get_rpc_filter_context(request: Request, user) -> tuple[Optional[str], Optio
                 db_user_is_admin = bool(_db_user and _db_user.is_admin) if _db_user else None
             finally:
                 _db.close()
-        # Grant bypass when DB confirms admin, OR when DB is unavailable/db_user_is_admin is None:
-        # fall back to the cached token_teams=None signal (set by auth.py resolve_session_teams).
-        # The DB check is defense-in-depth against cache staleness, not a gate — if the DB
-        # session fails or the user row is missing, trust the pipeline that set request.state.
-        if db_user_is_admin is not False:
+        # Grant bypass only when the fresh DB check positively confirms admin.
+        # db_user_is_admin is None (user missing from DB or query error) must
+        # fail-closed — a deleted or missing user should not inherit bypass
+        # even if the cached token_teams=None signal persists.
+        if db_user_is_admin is True:
             is_admin = True
             logger.debug(
                 "Session admin bypass: token_use=%s, email=%s path=%s (db_check=%s)",
--- a/mcpgateway/auth_context.py
+++ b/mcpgateway/auth_context.py
@@ -341,6 +341,14 @@ def get_token_teams_from_request(request: Request) -> Optional[List[str]]:
     if cached and isinstance(cached, tuple) and len(cached) == 2:
         _, payload = cached
         if payload:
+            # KNOWN LIMITATION: This fallback path uses normalize_token_teams()
+            # which doesn't understand session semantics. A session token
+            # reaching here would never produce token_teams=None, so the
+            # session admin bypass in get_rpc_filter_context() wouldn't fire.
+            # This is fail-secure (denies bypass rather than wrongly granting
+            # it) but is uncovered and undocumented as a design constraint.
+            # If auth.py ever stops setting request.state.token_teams before
+            # this code runs, session admin bypass would silently stop working.
             return normalize_token_teams(payload)
 
     # No JWT payload - return [] for public-only (secure default).
@@ -420,8 +428,14 @@ def get_rpc_filter_context(request: Request, user) -> tuple[Optional[str], Optio
     # Session token admin bypass: resolve_session_teams() confirmed admin from DB,
     # but JWT payload lacks is_admin claim (by design — DB is the authority for
     # session tokens so revocations take effect immediately).
-    if not is_admin and token_teams is None and getattr(request.state, "token_use", None) == "session":
+    if not is_admin and token_teams is None and getattr(request.state, "token_use", None) == "session":  # nosec B105 - Not a password; token_use is a JWT claim type
         is_admin = True
+        logger.debug(
+            "Session admin bypass: token_use=%s, email=%s path=%s",
+            getattr(request.state, "token_use", None),
+            user_email,
+            getattr(getattr(request, "url", None), "path", "unknown"),
+        )
 
     return user_email, token_teams, is_admin
 
--- a/mcpgateway/auth_context.py
+++ b/mcpgateway/auth_context.py
@@ -417,6 +417,12 @@ def get_rpc_filter_context(request: Request, user) -> tuple[Optional[str], Optio
     if token_teams is not None and len(token_teams) == 0:
         is_admin = False
 
+    # Session token admin bypass: resolve_session_teams() confirmed admin from DB,
+    # but JWT payload lacks is_admin claim (by design — DB is the authority for
+    # session tokens so revocations take effect immediately).
+    if not is_admin and token_teams is None and getattr(request.state, "token_use", None) == "session":
+        is_admin = True
+
     return user_email, token_teams, is_admin
 
 
--- a/mcpgateway/auth_context.py
+++ b/mcpgateway/auth_context.py
@@ -420,7 +420,12 @@ def get_rpc_filter_context(request: Request, user) -> tuple[Optional[str], Optio
     if cached and isinstance(cached, tuple) and len(cached) == 2:
         _, payload = cached
         if payload:
-            is_admin = payload.get("is_admin", False) or payload.get("user", {}).get("is_admin", False)
+            # Session tokens ignore JWT is_admin claim — DB is the authority.
+            # An old/stale session JWT carrying is_admin=true must not influence
+            # the boolean admin decision; only DB-resolved token_teams=None can
+            # produce admin bypass below.
+            if getattr(request.state, "token_use", None) != "session":
+                is_admin = payload.get("is_admin", False) or payload.get("user", {}).get("is_admin", False)
 
     if token_teams is not None and len(token_teams) == 0:
         is_admin = False
@@ -428,14 +433,35 @@ def get_rpc_filter_context(request: Request, user) -> tuple[Optional[str], Optio
     # Session token admin bypass: resolve_session_teams() confirmed admin from DB,
     # but JWT payload lacks is_admin claim (by design — DB is the authority for
     # session tokens so revocations take effect immediately).
+    #
+    # Fresh DB check: auth caching can produce token_teams=None from stale
+    # cached_ctx.user["is_admin"] without re-reading the DB.  Verify admin
+    # status from DB before granting bypass so demotions take effect immediately
+    # even when the auth cache has not yet expired.
     if not is_admin and token_teams is None and getattr(request.state, "token_use", None) == "session":  # nosec B105 - Not a password; token_use is a JWT claim type
-        is_admin = True
-        logger.debug(
-            "Session admin bypass: token_use=%s, email=%s path=%s",
-            getattr(request.state, "token_use", None),
-            user_email,
-            getattr(getattr(request, "url", None), "path", "unknown"),
-        )
+        db_user_is_admin = None
+        if user_email:
+            from mcpgateway.db import EmailUser, SessionLocal  # lazy import — avoids cycle
+
+            _db = SessionLocal()
+            try:
+                _db_user = _db.query(EmailUser).filter(EmailUser.email == user_email).first()
+                db_user_is_admin = bool(_db_user and _db_user.is_admin) if _db_user else None
+            finally:
+                _db.close()
+        # Grant bypass when DB confirms admin, OR when DB is unavailable/db_user_is_admin is None:
+        # fall back to the cached token_teams=None signal (set by auth.py resolve_session_teams).
+        # The DB check is defense-in-depth against cache staleness, not a gate — if the DB
+        # session fails or the user row is missing, trust the pipeline that set request.state.
+        if db_user_is_admin is not False:
+            is_admin = True
+            logger.debug(
+                "Session admin bypass: token_use=%s, email=%s path=%s (db_check=%s)",
+                getattr(request.state, "token_use", None),
+                user_email,
+                getattr(getattr(request, "url", None), "path", "unknown"),
+                db_user_is_admin,
+            )
 
     return user_email, token_teams, is_admin
 
--- a/tests/integration/test_session_admin_bypass.py
+++ b/tests/integration/test_session_admin_bypass.py
@@ -28,11 +28,7 @@ from sqlalchemy.pool import StaticPool
 import mcpgateway.db as db_mod
 import mcpgateway.main as main_mod
 from mcpgateway.auth import get_current_user
-from mcpgateway.auth_context import (
-    get_rpc_filter_context,
-    get_scoped_resource_access_context,
-    get_user_email,
-)
+from mcpgateway.auth_context import get_rpc_filter_context
 from mcpgateway.config import settings
 from mcpgateway.db import Base, EmailUser
 from mcpgateway.middleware.rbac import get_current_user_with_permissions
--- a/tests/integration/test_session_admin_bypass.py
+++ b/tests/integration/test_session_admin_bypass.py
@@ -7,7 +7,7 @@ Authors: Mihai Criveti
 Integration tests for session token admin bypass (PR #5232, issue #5232).
 
 Exercises the real auth pipeline end-to-end: session login via /auth/email/login
-followed by a request that triggers get_rpc_filter_context. Catches regressions
+followed by requests that trigger get_rpc_filter_context. Catches regressions
 where auth.py stops setting request.state.token_use, which unit-level tests
 (which mock request.state directly) would not catch.
 """
@@ -28,7 +28,11 @@ from sqlalchemy.pool import StaticPool
 import mcpgateway.db as db_mod
 import mcpgateway.main as main_mod
 from mcpgateway.auth import get_current_user
-from mcpgateway.auth_context import get_rpc_filter_context
+from mcpgateway.auth_context import (
+    get_rpc_filter_context,
+    get_scoped_resource_access_context,
+    get_user_email,
+)
 from mcpgateway.config import settings
 from mcpgateway.db import Base, EmailUser
 from mcpgateway.middleware.rbac import get_current_user_with_permissions
@@ -41,6 +45,10 @@ def app_and_client():
     Uses the real auth pipeline (require_auth, get_current_user) but overrides
     get_current_user_with_permissions to bypass complex RBAC while preserving
     request.state setup from the real auth chain.
+
+    The app registers a synthetic admin bypass check endpoint
+    (_test_session_admin_bypass_check) that returns get_rpc_filter_context
+    results and can look up a private tool by ID.
     """
     from _pytest.monkeypatch import MonkeyPatch
 
@@ -85,9 +93,11 @@ def app_and_client():
         finally:
             db.close()
 
+    from mcpgateway.db import get_db as mcpgateway_get_db
     from mcpgateway.routers.auth import get_db as auth_get_db
     from mcpgateway.middleware.rbac import get_db as rbac_get_db
 
+    main_mod.app.dependency_overrides[mcpgateway_get_db] = override_get_db
     main_mod.app.dependency_overrides[auth_get_db] = override_get_db
     main_mod.app.dependency_overrides[rbac_get_db] = override_get_db
 
@@ -105,17 +115,19 @@ def app_and_client():
 
     main_mod.app.dependency_overrides[get_current_user_with_permissions] = mock_user_with_permissions
 
-    # Seed an admin user with a valid UUID for the id field
+    # Seed users and a private tool
     import uuid
 
     from mcpgateway.services.argon2_service import Argon2PasswordService
 
     argon2 = Argon2PasswordService()
     db = TestSessionLocal()
+
+    # Admin user
     admin_user = EmailUser(
         id=str(uuid.uuid4()),
         email="admin-bypass@example.com",
-        password_hash=argon2.hash_password("TestPass123!"),
+        password_hash=argon2.hash_password("TestPass123!"),  # pragma: allowlist secret
         full_name="Admin Bypass Test",
         is_admin=True,
         is_active=True,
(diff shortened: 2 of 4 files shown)


vksinha10:
@MohanLaksh @Lang-Akshay @brian-hussey @crivetimihai @ja8zyjits - Any chances of merging this PR. We are running this in production and we need this as an urgent fix

msureshkumar88 (changes requested):
Thanks for the clear root-cause writeup and the comparison table against #4878 — that made this much easier to verify. I traced the logic end-to-end (all 3 `request.state.token_use` write sites in `auth.py`, and `resolve_session_teams`/`_narrow_by_jwt_teams` in `auth.py`) and the core claim holds: `token_teams is None` for a session token only happens when the DB confirmed admin, so the new bypass condition is sound and doesn't introduce a privilege-escalation path. The owner-matching check downstream still enforces the #4341 invariant.

A couple of things worth addressing before merge, plus a few non-blocking suggestions:

**Requesting changes on:**
- `mcpgateway/auth_context.py:423` — the other three `token_use == "session"` comparisons in `auth.py` (lines 1624, 1697, 1890) all carry `# nosec B105` since bandit's hardcoded-password heuristic can match on the string literal. This new comparison doesn't have it. Worth confirming bandit was actually run against this line (the PR says "Bandit: clean") and adding the suppression comment for consistency if it applies — otherwise a future CI run could flag it.

**Suggestions (non-blocking):**
- This branch grants admin bypass silently — no log line. Every other `is_admin` derivation in this file is also silent, so it's consistent with existing style, but an admin-bypass grant feels like exactly the kind of event worth a `logger.debug` for the audit trail (token_use + masked email).
- Test coverage is solid at the unit level (5 tests, not 6 as the PR description states — minor mismatch, the commit message correctly says 5), but all of them mock `request.state` directly. There's no test that exercises the real pipeline: session login via `/auth/email/login` → RPC call → confirms the admin sees their own private resource. That would catch a future regression where `auth.py` stops setting `request.state.token_use`, which the current mocked tests wouldn't catch.
- Minor edge case, not a vulnerability: `get_token_teams_from_request`'s fallback path (when `request.state.token_teams` isn't set and it falls back to `normalize_token_teams(payload)`) doesn't know session semantics, so a session token hitting that fallback would never produce `token_teams=None` and the new bypass wouldn't fire. Fail-secure (denies bypass rather than wrongly granting it), but uncovered and undocumented as a known limitation.

Everything else checks out — scope is tight (3 files, no unrelated changes), no breaking change, issue is correctly linked (`Closes #5232`), and the AGENTS.md table already documented the intended behavior so no doc update was needed.

msureshkumar88 (dismissed):
Blocking item is resolved — `mcpgateway/auth_context.py:431` now carries `# nosec B105` matching the three existing `token_use == "session"` comparisons in `auth.py`. The non-blocking suggestions were also picked up: a `logger.debug` fires on session admin bypass (token_use/email/path), the `get_token_teams_from_request` fallback limitation is now documented inline, and `tests/integration/test_session_admin_bypass.py` exercises the real `/auth/login` → RPC pipeline rather than mocking `request.state`.

I also ran a manual end-to-end check against a live gateway (real SQLite DB, no mocks) to validate the fix against the original bug report:

1. Logged in as the bootstrap admin via `/auth/email/login` (session token, `token_use=session`, no `is_admin` JWT claim).
2. Created a `visibility=private` tool owned by that admin via `POST /tools`.
3. Fetched it back via `GET /admin/tools/{id}` with the same session token.

**Pre-fix** (parent commit `2dcb92a7a`, before this PR): step 3 returned **HTTP 404** — reproduces issue #5232 exactly (`admin_bypass=false` blocking the admin from their own private resource).

**Post-fix** (this branch, HEAD `dac44b2b3`): step 3 returns **HTTP 200** — admin bypass works.

**Negative control**: a freshly registered non-admin user hitting the same `GET /admin/tools/{id}` gets **HTTP 403** — confirms the #4341 invariant (admin bypass never leaks another user's private resources) still holds.

Approving.

msureshkumar88:
For reproducibility, here's the environment and commands used for the manual e2e check:

**Environment**
- Python venv: project `.venv` (already has `mcp_contextforge_gateway` installed editable, FastAPI/uvicorn/SQLAlchemy from `pyproject.toml`)
- DB: fresh SQLite file per run (`sqlite:///<tmpdir>/test.db`), bootstrapped via the app's normal `alembic upgrade head` + `bootstrap_db.py` startup path — no fixtures, no mocks
- Server: `uvicorn mcpgateway.main:app` on `127.0.0.1`, single process

Key env vars (test-only relaxations, not security-relevant to the fix itself):
```bash
DATABASE_URL=sqlite:///<tmpdir>/test.db
JWT_SECRET_KEY=e2e-test-secret-key-5239
AUTH_ENCRYPTION_SECRET=e2e-test-salt-5239
AUTH_REQUIRED=true
MCPGATEWAY_UI_ENABLED=true
PLATFORM_ADMIN_PASSWORD=changeme123
DETECT_DEFAULT_PASSWORD_ON_LOGIN=false        # skip unrelated forced-password-change UX
PASSWORD_CHANGE_ENFORCEMENT_ENABLED=false
EXPOSE_ERROR_DETAILS=true                     # see real validation errors instead of generic 422
PUBLIC_REGISTRATION_ENABLED=true              # needed only for the non-admin negative control
```

**Commands**

Start server:
```bash
.venv/bin/python -m uvicorn mcpgateway.main:app --host 127.0.0.1 --port 18443
```

1. Admin session-token login (no `is_admin` claim in JWT — by design):
```bash
curl -s -X POST http://127.0.0.1:18443/auth/email/login \
  -H 'Content-Type: application/json' \
  -d '{"email":"admin@example.com","password":"changeme123"}'
```

2. Create a private tool with that session token:
```bash
curl -s -X POST http://127.0.0.1:18443/tools \
  -H "Authorization: Bearer $TOKEN" -H 'Content-Type: application/json' \
  -d '{"tool":{"name":"e2e_tool","url":"http://example.com/e2e_tool","integration_type":"REST","request_type":"GET","visibility":"private"}}'
```

3. The actual repro step — fetch it back via the Admin UI JSON endpoint with the same session token:
```bash
curl -s -w '\n%{http_code}' http://127.0.0.1:18443/admin/tools/$TOOL_ID \
  -H "Authorization: Bearer $TOKEN"
```
Pre-fix (parent commit `2dcb92a7a`) → `404`. Post-fix (this branch) → `200`.

4. Negative control — register + log in as a second, non-admin user, repeat step 3 with their token → `403` on both commits (security boundary unaffected by the fix).

To diff the two commits without touching this checkout, I used a separate `git worktree add <tmp-dir> 2dcb92a7a`, then bypassed the editable-install metapath finder so `.venv`'s interpreter would import `mcpgateway` from that worktree instead of the main checkout:
```bash
.venv/bin/python -c "
import sys
sys.path.insert(0, '<worktree-dir>')
sys.meta_path = [m for m in sys.meta_path if 'editable' not in type(m).__module__.lower()]
import uvicorn
uvicorn.run('mcpgateway.main:app', host='127.0.0.1', port=18443)
"
```
Worktree was removed after the comparison (`git worktree remove --force`).

mcpgateway/auth_context.py
vishu-bh, line 419: Worth tightening while we are in this code: `get_rpc_filter_context()` still reads `payload["is_admin"]` for all JWTs, including `token_use="session"`.

For session tokens, the PR rationale says the DB is the authority and the JWT intentionally omits `is_admin`. To preserve that invariant, session tokens should ignore any JWT `is_admin`/`user.is_admin` claim entirely; only DB-resolved `token_teams is None` should produce session admin bypass. Otherwise an older/custom/stale session JWT carrying `is_admin=true` can still influence boolean admin decisions.

mcpgateway/auth_context.py
```diff
@@ -417,6 +425,18 @@ def get_rpc_filter_context(request: Request, user) -> tuple[Optional[str], Optio
     if token_teams is not None and len(token_teams) == 0:
         is_admin = False
 
+    # Session token admin bypass: resolve_session_teams() confirmed admin from DB,
+    # but JWT payload lacks is_admin claim (by design — DB is the authority for
+    # session tokens so revocations take effect immediately).
+    if not is_admin and token_teams is None and getattr(request.state, "token_use", None) == "session":  # nosec B105 - Not a password; token_use is a JWT claim type
```
vishu-bh, line 441: This grants session admin bypass whenever `request.state.token_teams is None`, but one auth path can derive that `None` from cached user state rather than a fresh DB admin check.

In the cached auth path, `auth.py` calls `resolve_session_teams(payload, email, cached_ctx.user)`. `_resolve_teams_from_db()` then trusts `cached_ctx.user["is_admin"]` and returns `None` without re-reading the DB. Since auth caching is enabled by default and invalidation is best-effort, a recently demoted admin can retain cached `token_teams=None` for the cache TTL; this new condition then turns that stale cached value into `is_admin=True`.

That conflicts with the session-token invariant that the DB is authoritative so admin revocation takes effect immediately. Could we force a fresh DB admin check before granting session admin bypass, or otherwise make admin demotion/cache invalidation fail closed?

tests/integration/test_session_admin_bypass.py
```diff
@@ -0,0 +1,180 @@
...
+        )
+        assert payload.get("token_use") == "session", "Expected session token"
+
+        # Step 3: Use the session JWT to call /servers/ endpoint
+        # This exercises the real auth pipeline:
+        #   require_auth → get_current_user → _inject_userinfo_instate
+        #   → request.state.token_use = "session"
+        #   → get_rpc_filter_context → admin bypass fires
+        headers = {"Authorization": f"Bearer {token}"}
+        resp = client.get("/servers/", headers=headers)
```
vishu-bh: This integration test checks that the real login token can call `/servers/`, but because the DB has no servers and the assertion is `200` + `[]`, a public-only/non-admin session could also pass.

The manual e2e described in the review comment is stronger than the committed test. Could we seed a private tool/resource owned by the admin and assert `GET /admin/tools/{id}` succeeds with the admin session token, plus a non-admin token cannot read it? That would lock down the actual #5232 regression path.

vishu-bh (changes requested):
Thanks for the focused fix here. The fresh-session path does appear to address #5232: a /auth/email/login session token without an is_admin JWT claim can now flow through as admin bypass when resolve_session_teams() has produced token_teams=None, and the downstream owner check still protects other users’ private resources.

Spotted potential issues, worth looking in.

MohanLaksh:
| Comment | Code | Approach |
|---------|------|----------|
| **1** — JWT `is_admin` trusted for session tokens | `auth_context.py:430` | Skip JWT `is_admin` read when `token_use == "session"` so only DB-resolved `token_teams=None` can produce bypass |
| **2** — Cache staleness for demoted admins | `auth_context.py:441-466` | Fresh DB admin check inside bypass condition; falls back to cached `token_teams=None` signal when user row is missing (fail-soft) |
| **3** — Integration test too weak | `test_session_admin_bypass.py` | Seeds admin, non-admin, and a private tool. Tests via synthetic endpoint that exercises real auth pipeline + `get_rpc_filter_context` directly. Asserts admin gets `(email, None, True)`, non-admin gets `(email, teams, False)` |

mcpgateway/auth_context.py
```diff
@@ -412,11 +420,49 @@ def get_rpc_filter_context(request: Request, user) -> tuple[Optional[str], Optio
...
+            try:
+                _db_user = _db.query(EmailUser).filter(EmailUser.email == user_email).first()
+                db_user_is_admin = bool(_db_user and _db_user.is_admin) if _db_user else None
+            finally:
+                _db.close()
+        # Grant bypass when DB confirms admin, OR when DB is unavailable/db_user_is_admin is None:
+        # fall back to the cached token_teams=None signal (set by auth.py resolve_session_teams).
+        # The DB check is defense-in-depth against cache staleness, not a gate — if the DB
+        # session fails or the user row is missing, trust the pipeline that set request.state.
+        if db_user_is_admin is not False:
```
vishu-bh: Small security concern: I think this should be `if db_user_is_admin is True` rather than `if db_user_is_admin is not False`.

If no DB user is found, `db_user_is_admin` remains `None`, and `None is not False` still grants `is_admin=True`. Since this is session-token admin bypass and the DB is supposed to be the authority, we should only grant bypass when the fresh DB check positively confirms admin status. Missing/unknown DB state should fail closed.

vishu-bh (changes requested):
Thanks @MohanLaksh for addressing the comments. 

All looks good except one behavior related change.

vishu-bh (approved):
Thanks @MohanLaksh, this PR fixes the session-token admin bypass issue for the Admin UI path by making `get_rpc_filter_context()` handle `token_use="session"` correctly. Session JWTs now ignore any JWT `is_admin` claim, and admin bypass is granted only when `token_teams=None` is paired with a fresh DB check confirming the user is still an admin. The latest change also correctly fails closed when the DB user is missing.

The added tests cover the happy path, non-admin/public-only behavior, API token regression, missing DB user, and a real login/session integration path. This addresses #5232 while preserving the owner-matching/private-resource security boundary.

LGTM 🚀

msureshkumar88 (approved):
All blocking items resolved:
- `auth_context.py:431` carries `# nosec B105` matching the three existing comparisons in `auth.py`
- Session tokens now correctly ignore JWT `is_admin` claim (DB is sole authority)
- Fresh DB check added before granting session admin bypass (stale cache can't escalate)
- Missing DB user now fails closed (`is True` guard, commit 58011bdf)
- Non-blocking suggestions also addressed: `logger.debug` on bypass, fallback limitation documented inline, integration test exercises real login pipeline

LGTM.

ja8zyjits (approved):
LGTM 
Process followed

vksinha10:
After upgrading to the latest image, the public/private tool list issue was resolved, but RPC /tools/call requests started failing with "tool not found" errors for admin users trying to access their own private tools.

Root Cause: The RPC handler in main.py (function _execute_rpc_tools_call, lines 10337-10340) was setting auth_user_email = None when admin bypass was active (auth_is_admin and auth_token_teams is None). This broke the owner-matching logic in the tool service layer, preventing admins from accessing their own private resources via RPC calls.

Affected Code: File: main.py Function: _execute_rpc_tools_call (line 10298) Lines: 10337-10340

Before (broken):

python
auth_user_email, auth_token_teams, auth_is_admin = get_rpc_filter_context(request, user)
run_owner_email = auth_user_email
run_owner_team_ids = [] if auth_token_teams is None else list(auth_token_teams)
if auth_is_admin and auth_token_teams is None:
auth_user_email = None # ❌ Breaks owner matching
elif auth_token_teams is None:
auth_token_teams = []

Impact:

Admin users with token_teams=None (admin bypass) could not invoke their own private tools via /rpc calls
Error: "Tool not found: {tool_name}"
Affects all RPC tool invocations for private resources when using admin bypass tokens
Related Issues:

This is the same bug pattern fixed in PR https://github.com/IBM/mcp-context-forge/issues/4877 for other resource types (tools, prompts, resources, gateways, servers)
PR https://github.com/IBM/mcp-context-forge/issues/4877 reference: "Admin users can now access their OWN private resources while maintaining the security invariant that admin bypass NEVER reveals other users' private resources"

PR #5241, opened 2026-06-16
bugfix/5213-admin-session-team-visibility-404

### LINKED ISSUE #5213: [BUG]: Admin session-cookie users get 404 on team-visibility MCP servers/gateways
### 🧩 Affected Component

- [x] `mcpgateway` - API
- [ ] Other (explain below)

---

## Describe the bug

When logged into the `/admin` UI using cookie-based session authentication, an admin user receives a 404 error when attempting to view or edit an MCP server (gateway) whose visibility is set to `Team`. Public-visibility resources are unaffected.

## Steps to reproduce

  1. Create an MCP gateway/server and set its visibility to `Team` with a team assigned.
  2. Log in as a platform admin via the `/admin` UI (cookie/session auth).
  3. Navigate to the MCP Servers list — the server appears in the list.
  4. Click **View** or **Edit** on the team-visibility server.
  5. Receive a 404 error.

## Expected behavior

Admin users should be able to view and edit all resources regardless of visibility (`public`, `team`, or own `private`), consistent with the admin bypass documented in `docs/docs/manage/rbac.md`.

## Root cause

  The bug is in `mcpgateway/auth_context.py:get_scoped_resource_access_context`.

Session tokens (`token_use: "session"`, issued on cookie login) do not embed `is_admin` in the JWT payload — the DB is the authority for session tokens, per the two-layer security model. During request processing, `get_current_user` calls `resolve_session_teams()` → `_resolve_teams_from_db()`, which confirms the user is a DB admin and correctly sets `request.state.token_teams = None` (the admin bypass sentinel).

However, `get_scoped_resource_access_context` reads `is_admin` from the JWT payload (via `get_rpc_filter_context`), gets `False` because session tokens have no such claim, and then hits:

  ```python
  if is_admin and token_teams is None:   # False — is_admin is False for session tokens
      return user_email, None

  if token_teams is None:                # True — but incorrectly treated as "unknown"
      return user_email, []              # ← converts admin bypass to public-only
  ```

This transforms the correctly DB-resolved `token_teams=None` (admin bypass) into `[]` (public-only). Downstream, `_check_gateway_access` / `_check_server_access` receives `token_teams=[]`, `is_admin_bypass_granted` returns `False`, and the team-visibility resource is denied with a 404.

## The fix

  In `mcpgateway/auth_context.py`, after the existing `is_admin and token_teams is None` check, add a session-token-specific guard that trusts the DB-backed `token_teams=None`:

  ```python
  # Session tokens do not embed is_admin in the JWT payload; instead,
  # resolve_session_teams() queries the DB and stores token_teams=None when
  # the user is confirmed as admin. Trust that DB-backed resolution here so
  # that admin session-cookie users can see team-visibility resources.
  if token_teams is None and getattr(request.state, "token_use", None) == "session":
      return user_email, None
  ```

  `token_teams=None` is set for session tokens **only** when `resolve_session_teams` → `_resolve_teams_from_db` confirms `is_admin=True` in the DB. It is never set to `None` for non-admin session token users. The guard is therefore safe.

 ## Affected paths

  - `GET /admin/gateways/{id}` — Admin UI view
  - `POST /admin/gateways/{id}/edit` — Admin UI edit
  - `GET /admin/servers/{id}` — same issue for virtual servers

Nayana-R-Gowda:
closes #5213 

**Two changes in mcpgateway/auth_context.py:**

**Change 1** — Root cause fix in get_rpc_filter_context() (line 420-424): Session tokens don't embed is_admin in the JWT payload, so payload.get("is_admin") returns False even for DB admin users. Added a guard that checks: if is_admin is still False but token_teams is None (the admin bypass sentinel, only set by resolve_session_teams() for DB-confirmed admins) and token_use is "session", then set is_admin = True. This fixes all ~20+ callers in main.py that use get_rpc_filter_context directly, plus _assert_session_owner_or_admin via get_request_identity.

**Change 2** — Defense-in-depth in get_scoped_resource_access_context() (line 499-504): Additional session-token guard at the scope resolution layer. Even if get_rpc_filter_context returns is_admin=False, a session token with token_teams=None is treated as admin bypass. This is the fix described in the issue.

### DIFF
--- a/mcpgateway/auth_context.py
+++ b/mcpgateway/auth_context.py
@@ -417,6 +417,12 @@ def get_rpc_filter_context(request: Request, user) -> tuple[Optional[str], Optio
     if token_teams is not None and len(token_teams) == 0:
         is_admin = False
 
+    # Session tokens do not embed is_admin in the JWT payload; resolve_session_teams()
+    # queries the DB and sets token_teams=None only when the user is a confirmed DB
+    # admin. Trust that DB-backed resolution to set is_admin correctly here.
+    if not is_admin and token_teams is None and getattr(request.state, "token_use", None) == "session":
+        is_admin = True
+
     return user_email, token_teams, is_admin
 
 
@@ -496,6 +502,12 @@ def get_scoped_resource_access_context(request: Request, user) -> tuple[Optional
 
     if is_admin and token_teams is None:
         return user_email, None  # Keep user_email for owner matching (PR #4341 / issue #4694)
+    # Session tokens do not embed is_admin in the JWT payload; instead,
+    # resolve_session_teams() queries the DB and stores token_teams=None when
+    # the user is confirmed as admin. Trust that DB-backed resolution here so
+    # that admin session-cookie users can see team-visibility resources.
+    if token_teams is None and getattr(request.state, "token_use", None) == "session":
+        return user_email, None
     if token_teams is None:
         return user_email, []
     return user_email, token_teams
--- a/tests/unit/mcpgateway/test_main.py
+++ b/tests/unit/mcpgateway/test_main.py
@@ -2371,7 +2371,6 @@ class TestRootEndpoints:
         assert response.status_code == 500
         assert "Internal error" in response.json()["detail"]
 
-
     @patch("mcpgateway.main.root_service.subscribe_changes")
     def test_subscribe_root_changes(self, mock_subscribe, test_client, auth_headers):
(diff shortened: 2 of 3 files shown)


jonpspri:
Closing as superseded. The linked issue #5213 ([BUG]: Admin session-cookie users get 404 on team-visibility MCP servers/gateways) was closed as COMPLETED — the fix was already merged. This PR is no longer needed.

PR #5244, opened 2026-06-16
fix(gateways): resolve authorization_code OAuth gateway offline issues

### LINKED ISSUE #4154: [BUG][TRANSPORT]: StreamableHTTP health checks return 405 — full SDK initialize() incompatible with health probing

### LINKED ISSUE #5237: [BUG]: authorization_code OAuth gateways become permanently offline due to health-check token assumptions and destructive refresh-token handling
### 🐞 Bug Summary

Two independent defects in the OAuth gateway implementation cause authorization_code gateways to become permanently unavailable:

Health checks incorrectly depend on a single platform-admin token, causing healthy gateways to be marked offline within ~60 seconds.
Refresh-token handling can permanently delete valid user tokens, causing gateways to fail again after token expiration (~1 hour).

Together, these issues create a self-reinforcing failure loop with no automatic recovery.

---

### 🧩 Affected Component

mcpgateway/services/gateway_service.py
initialize() (around line 748)
_check_single_gateway_health() (around lines 3831–3847)
register_gateway() (around line 1076)
mcpgateway/services/token_storage_service.py
_refresh_access_token() (around lines 296–302, 352–354, 420–425)
mcpgateway/config.py
platform_admin_email
---

**### 🔁 Steps to Reproduce**

**Issue 1: Health Check Uses a Single Synthetic Identity**

**Root Cause**

The gateway health-check process always attempts to retrieve a token using:

user_email = settings.platform_admin_email

For authorization_code gateways, tokens are stored per-user. In many deployments the platform admin account is a synthetic identity (e.g. admin@example.com) that never completes OAuth authorization.

Health checks therefore fail even when valid user tokens exist:

access_token = await token_storage.get_user_token(gateway_id, user_email)

if access_token:
    headers["Authorization"] = f"Bearer {access_token}"
else:
    await self._handle_gateway_failure(gateway)
    return
Impact
Gateway starts as reachable=true
Users successfully authorize and can initially use tools
Within ~60 seconds, health check finds no admin token
Gateway is marked reachable=false
Tool execution is blocked for all users despite valid user tokens
Additional Concern

Using a real user account as PLATFORM_ADMIN_EMAIL is not a scalable workaround:

Only gateways authorized by that user remain healthy
Tokens belonging to other users are ignored
Creates a single point of failure for all OAuth gateways
Configuration only supports one health-check identity

**Expected Behavior**

Health checks should verify service reachability, not token ownership.

**Suggested behavior:**

Missing token → skip authenticated probe
401/403 responses → gateway considered reachable
Only connection failures, timeouts, or genuine server errors should mark a gateway offline

---

**Issue 2: Refresh Logic Can Permanently Delete Valid Tokens**
**2a. Inconsistent client_secret Handling**

API registration stores client_secret differently than UI registration.

UI path:

Secret is encrypted before storage

API path:

Secret is stored as provided

Refresh logic always attempts decryption:

try:
    oauth_config["client_secret"] = await self.encryption.decrypt_secret_async(
        oauth_config["client_secret"]
    )
except Exception:
    pass

**Problems:**

Decryption failures are silently ignored
Ciphertext may be used as a credential
Impossible to distinguish encrypted vs plaintext secrets
Refresh requests can fail with invalid_client
Expected Behavior
Encrypt secrets consistently for both UI and API registration paths
Fail refresh explicitly if decryption fails
Do not silently continue with an unknown credential value
**2b. omit_resource: true Not Fully Respected**

Refresh logic injects a resource value derived from the gateway URL:

elif gateway.url:
    oauth_config["resource"] = normalize_resource(gateway.url)

This occurs even when:

omit_resource: true
Impact
Configuration is modified contrary to explicit user intent
Creates future compatibility risk for Azure AD v2 scope-based flows
Obscures the meaning of omit_resource
Expected Behavior

Skip resource derivation entirely when omit_resource is enabled.

**2c. Broad Exception Matching Deletes Tokens**

Refresh failures trigger token deletion when exception text contains:

"invalid"

or

"expired"
except Exception as e:
    if "invalid" in str(e).lower() or "expired" in str(e).lower():
        self.db.delete(token_record)
        self.db.commit()
Impact

Valid tokens can be permanently deleted due to:

invalid_client
invalid_request
Temporary provider-side failures
Any exception message containing those keywords

Once deleted:

Health check no longer finds a token
Gateway is marked offline
User must manually re-authorize
Expected Behavior

Delete tokens only after confirming a permanent OAuth failure, such as:

{
  "error": "invalid_grant"
}

Transient refresh failures should preserve the token and allow retry.

**Reproduction**
Create an OAuth gateway using:
grant_type=authorization_code
omit_resource=true
Authorize as a normal user.
Verify tools work initially.
Wait for the health check (~60 seconds).
Gateway becomes reachable=false because no token exists for platform_admin_email.
Configure a real user as PLATFORM_ADMIN_EMAIL and authorize.
Wait until token refresh is attempted (~55–60 minutes).
Refresh fails (invalid_client, invalid_request, etc.).
Token is deleted.
Next health check again marks the gateway offline.

**Expected Result**
Gateway health should be independent of any single user's OAuth token.
Missing tokens should not mark a gateway unreachable.
Refresh-token failures should not automatically delete stored tokens.
client_secret handling should be consistent across API and UI registration paths.
omit_resource=true should prevent all resource injection behavior.

**Actual Result**

OAuth authorization_code gateways eventually transition to an unrecoverable offline state and require repeated manual re-authorization, despite the gateway service itself remaining reachable and functional.

bogdanmariusc10:
## 🔗 Related Issue
Closes #5237
Closes #4154

---

## 📝 Summary

This PR fixes critical issues where OAuth `authorization_code` gateways would become permanently offline due to health check assumptions and destructive token handling behavior.

**Problem:** 
`authorization_code `gateways (where users authorize individually) were failing health checks when no platform admin token existed, causing cascading failures:
1. Health checks incorrectly failed when no token was available (expected state for user-specific auth)
2. 401/403 responses were treated as connectivity failures instead of auth challenges
3. Token refresh failures would permanently delete valid tokens due to overly broad error matching
4. client_secret decryption failures fell back silently to using ciphertext as credentials
5. omit_resource flag was ignored during token refresh, breaking IdPs that reject the resource parameter

**Solution:**
- Decouple health checks from token ownership - missing tokens don't fail health checks
- Treat 401/403 responses as "gateway reachable" (auth issue, not connectivity failure)
- Reactivate previously-unreachable gateways when they respond with 401/403
- Update last_seen timestamp even on auth failures to reflect gateway responsiveness
- Make client_secret decryption failures explicit (raise OAuthError) instead of silent fallback
- Respect omit_resource flag to prevent resource parameter injection when disabled
- Only delete tokens on RFC 6749 §5.2 `invalid_grant` errors (permanent failure), not transient errors

---

## 🏷️ Type of Change
- [x] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

| Lint suite                | `make lint`     | ✅ PASS |
| Unit tests                | `make test`     | ✅ PASS  |
| Coverage ≥ 80%            | `make coverage` | ✅ PASS |

**Test Results:**
```
✅ test_oauth_authorization_code_health.py: 8/8 passed
✅ test_gateway_service_401_reactivation.py: 5/5 passed  
✅ test_gateway_service_health_oauth.py: 15/15 passed  
✅ test_token_storage_service.py: 52/52 passed
✅ test_oauth_manager.py: 131/131 passed
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
✅ TOTAL: 172 OAuth-related tests passing
```

---

## ✅ Checklist
- [x] Documentation updated

---

### Files Changed

**Core Implementation (2 files):**
- `mcpgateway/services/gateway_service.py` - Health check logic for authorization_code flow
  - Lines 3824-3850: Skip authentication when no token available
  - Lines 3972-4018: Distinguish auth failures (401/403) from connectivity failures
  - Lines 3996-3998: Reactivate previously-unreachable gateways on 401/403
  - Lines 4005-4008: Update last_seen even on auth failures + exception handling

- `mcpgateway/services/token_storage_service.py` - Token refresh improvements
  - Lines 298-313: Explicit client_secret decryption with error handling
  - Lines 350-373: omit_resource flag enforcement
  - Lines 439-472: Selective token deletion (only invalid_grant)

**Test Updates (4 files):**
- `tests/unit/mcpgateway/services/test_gateway_service.py` - Added httpx import, updated expectations
- `tests/unit/mcpgateway/services/test_gateway_service_health_oauth.py` - OAuth health check test updates
- `tests/unit/mcpgateway/services/test_token_storage_service.py` - Token refresh test updates
- `tests/unit/mcpgateway/test_oauth_manager.py` - Changed to use OAuthError with invalid_grant

**New Test Files (2 files):**
- `tests/unit/mcpgateway/services/test_oauth_authorization_code_health.py` (8 tests)
  - Test health check behavior documentation
  - Test client_secret decryption failure handling
  - Test omit_resource flag prevention
  - Test resource injection when omit_resource=false
  - Test invalid_grant token deletion
  - Test invalid_client token preservation
  - Test transient error token preservation

- `tests/unit/mcpgateway/services/test_gateway_service_401_reactivation.py` (5 tests)
  - Test 401 reactivates previously unreachable gateway (coverage lines 3996-3998)
  - Test 403 reactivates previously unreachable gateway
  - Test 401 last_seen update failure handling (coverage lines 4007-4008)
  - Test 401 on already-reachable gateway (skip reactivation, update last_seen)
  - Test 401 on disabled gateway (skip reactivation, update last_seen)

### Technical Details

**Health Check Behavior:**
- authorization_code gateways no longer depend on platform admin token availability
- 401/403 responses now treated as "gateway is reachable but unauthorized" (healthy state)
- Previously-unreachable gateways are reactivated when they respond with auth challenges
- Gateway last_seen timestamp updated on all responses (including auth failures)

**Token Deletion Logic (RFC 6749 Compliance):**
```python
# OLD (bug): Any error with "invalid" or "expired" deleted tokens
if "invalid" in error_str or "expired" in error_str:
    delete_token()

# NEW (fixed): Only RFC 6749 §5.2 invalid_grant deletes tokens
except OAuthError as e:
    if "invalid_grant" in str(e).lower():
        delete_token()  # Permanent failure - refresh token revoked
    else:
        preserve_token()  # Configuration error or transient failure
except Exception:
    preserve_token()  # Network errors, not token issues
```

**client_secret Handling:**
```python
# OLD (bug): Silent fallback on decryption failure
try:
    decrypt(client_secret)
except:
    pass  # Uses encrypted value as credential!

# NEW (fixed): Explicit error on decryption failure
if is_encrypted(client_secret):
    try:
        decrypt(client_secret)
    except Exception as e:
        raise OAuthError(f"Failed to decrypt client_secret: {e}")
```

**omit_resource Flag:**
```python
# NEW: Check flag BEFORE resource injection
omit_resource = oauth_config.get("omit_resource", False)
if omit_resource:
    oauth_config.pop("resource", None)
    logger.debug("Omitting resource parameter per omit_resource=true")
else:
    # Only inject resource when flag is false/absent
    if gateway.url:
        oauth_config["resource"] = normalize_resource(gateway.url)
```

### Impact

**Before (broken):**
- authorization_code gateways marked unhealthy when no platform admin token exists
- 401/403 responses trigger _handle_gateway_failure() → gateway goes offline
- Any OAuth error matching "invalid" or "expired" permanently deletes tokens
- Decryption failures use ciphertext as client_secret → authentication always fails
- omit_resource flag ignored → breaks IdPs that reject resource parameter

**After (fixed):**
- Health checks verify service reachability independent of token ownership
- 401/403 responses treated as "gateway reachable" → stays online
- Only invalid_grant deletes tokens (per RFC 6749) → transient errors don't lose tokens
- Decryption failures raise explicit errors → clear diagnosis of configuration issues
- omit_resource flag respected → works with all IdP configurations

### Testing Strategy

1. **Unit tests** for individual components (token refresh, decryption, resource handling)
2. **Integration-style tests** for health check paths (401 reactivation, last_seen updates)
3. **Regression tests** to ensure existing OAuth functionality still works

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-07-20T10:15:31Z",
+  "generated_at": "2026-07-17T16:05:28Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -266,7 +266,7 @@
         "hashed_secret": "b4673e578b9b30fe8bba1b555b7b59883444c697",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1889,
+        "line_number": 1746,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -274,7 +274,7 @@
         "hashed_secret": "4a0a2df96d4c9a13a282268cab33ac4b8cbb2c72",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1977,
+        "line_number": 1834,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -282,7 +282,7 @@
         "hashed_secret": "5baa61e4c9b93f3f0682250b6cf8331b7ee68fd8",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 2327,
+        "line_number": 2184,
         "type": "Basic Auth Credentials",
         "verified_result": null
       },
@@ -290,7 +290,7 @@
         "hashed_secret": "fa9beb99e4029ad5a6615399e7bbae21356086b3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 3692,
+        "line_number": 3549,
         "type": "Basic Auth Credentials",
         "verified_result": null
       },
@@ -298,7 +298,7 @@
         "hashed_secret": "fa9beb99e4029ad5a6615399e7bbae21356086b3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 3783,
+        "line_number": 3640,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -306,7 +306,7 @@
         "hashed_secret": "ac371b6dcce28a86c90d12bc57d946a800eebf17",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 3826,
+        "line_number": 3683,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -314,7 +314,7 @@
         "hashed_secret": "0b6ec68df700dec4dcd64babd0eda1edccddace1",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 3831,
+        "line_number": 3688,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -322,7 +322,7 @@
         "hashed_secret": "4ad6f0082ee224001beb3ca5c3e81c8ceea5ed86",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 3836,
+        "line_number": 3693,
         "type": "Secret Keyword",
         "verified_result": null
       }
@@ -382,7 +382,7 @@
         "hashed_secret": "d3ac7a4ef1a838b4134f2f6e7f3c0d249d74b674",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 5966,
+        "line_number": 5963,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -390,7 +390,7 @@
         "hashed_secret": "5932862bcd24dd27d0dc0407ec94fe9d6ea24aeb",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 6463,
+        "line_number": 6460,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -398,7 +398,7 @@
         "hashed_secret": "c77c805e32f173e4321ee9187de9c29cb3804513",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 6475,
+        "line_number": 6472,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -406,7 +406,7 @@
         "hashed_secret": "8fe3df8a68ddd0d4ab2214186cbb8e38ccd0e06a",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 6547,
+        "line_number": 6544,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -414,7 +414,7 @@
         "hashed_secret": "93ac8946882128457cd9e283b30ca851945e6690",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 7614,
+        "line_number": 7611,
         "type": "Secret Keyword",
         "verified_result": null
       }
@@ -580,7 +580,7 @@
         "hashed_secret": "c58a69e6abb14e59f5e5761ac85a933c671bda09",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 306,
+        "line_number": 286,
         "type": "Secret Keyword",
         "verified_result": null
       }
@@ -590,7 +590,7 @@
         "hashed_secret": "cff0d14e4337fa8bdb68dfa906f04b0df6fad72f",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 61,
+        "line_number": 56,
         "type": "Secret Keyword",
         "verified_result": null
       }
@@ -642,7 +642,7 @@
         "hashed_secret": "d08f88df745fa7950b104e4a707a31cfce7b5841",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1104,
+        "line_number": 1103,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -650,7 +650,7 @@
         "hashed_secret": "7288edd0fc3ffcbe93a0cf06e3568e28521687bc",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1107,
+        "line_number": 1106,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -658,7 +658,7 @@
         "hashed_secret": "8674c9b302d20800e4ab3808f139704d8641a6e3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1274,
+        "line_number": 1272,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -666,7 +666,7 @@
         "hashed_secret": "cff0d14e4337fa8bdb68dfa906f04b0df6fad72f",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1313,
+        "line_number": 1311,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -674,7 +674,7 @@
         "hashed_secret": "f865b53623b121fd34ee5426c792e5c33af8c227",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1367,
+        "line_number": 1365,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -682,7 +682,7 @@
         "hashed_secret": "acde39840735314af1300688b6c2324ea89770a3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1463,
+        "line_number": 1460,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -690,7 +690,7 @@
         "hashed_secret": "fa9beb99e4029ad5a6615399e7bbae21356086b3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1834,
+        "line_number": 1831,
         "type": "Secret Keyword",
         "verified_result": null
       }
@@ -1191,6 +1191,216 @@
         "is_secret": false,
         "is_verified": false,
         "line_number": 184,
+        "type": "Secret Keyword",
+        "verified_result": null
+      }
+    ],
+    "docs/docs/architecture/roadmap.md": [
+      {
+        "hashed_secret": "ff2ccd73154c1c71ef9a2982fe16e3c529e7a1ae",
+        "is_secret": false,
+        "is_verified": false,
+        "line_number": 106,
+        "type": "Secret Keyword",
+        "verified_result": null
+      },
+      {
+        "hashed_secret": "74f82b992f8a92b849c2be77447f98a510338333",
+        "is_secret": false,
+        "is_verified": false,
+        "line_number": 107,
+        "type": "Secret Keyword",
+        "verified_result": null
+      },
+      {
+        "hashed_secret": "64814a3b7fd8444a56ad3641fd3451c6deaf0757",
+        "is_secret": false,
+        "is_verified": false,
+        "line_number": 115,
+        "type": "Secret Keyword",
+        "verified_result": null
+      },
+      {
+        "hashed_secret": "af84d91fde168566c7dc18f3121ea2fbe651af1f",
+        "is_secret": false,
+        "is_verified": false,
+        "line_number": 144,
+        "type": "Secret Keyword",
+        "verified_result": null
+      },
+      {
+        "hashed_secret": "d1081e351fbebe0deb0c2867d5f731c8f9cc3fd8",
+        "is_secret": false,
+        "is_verified": false,
+        "line_number": 397,
+        "type": "Secret Keyword",
+        "verified_result": null
+      },
+      {
+        "hashed_secret": "b7e9a6e2e04ed3edbf8b4e21def4beccbb6eb6b1",
+        "is_secret": false,
+        "is_verified": false,
+        "line_number": 398,
+        "type": "Secret Keyword",
+        "verified_result": null
+      },
+      {
+        "hashed_secret": "6d244f58364223afcc2322f696137b01ece78d9d",
+        "is_secret": false,
+        "is_verified": false,
+        "line_number": 399,
+        "type": "Secret Keyword",
+        "verified_result": null
+      },
+      {
+        "hashed_secret": "8bf177a72c93155020ebca9ee7f3c6e0fbdee946",
+        "is_secret": false,
+        "is_verified": false,
+        "line_number": 417,
+        "type": "Secret Keyword",
+        "verified_result": null
+      },
+      {
+        "hashed_secret": "c41e6cd4245e404f07635fdbac351aad4d54a213",
+        "is_secret": false,
+        "is_verified": false,
+        "line_number": 516,
+        "type": "Secret Keyword",
+        "verified_result": null
+      },
+      {
+        "hashed_secret": "6568a9761e26d16a111247f279b6fcabff1c8c86",
+        "is_secret": false,
+        "is_verified": false,
+        "line_number": 551,
+        "type": "Secret Keyword",
+        "verified_result": null
+      },
+      {
+        "hashed_secret": "314fc7fd580f8c510be196143946bbe5ea753443",
+        "is_secret": false,
+        "is_verified": false,
+        "line_number": 552,
+        "type": "Secret Keyword",
+        "verified_result": null
+      },
+      {
+        "hashed_secret": "afb9d1dcf212fa33d079a070ab90f0bef03c324b",
+        "is_secret": false,
+        "is_verified": false,
+        "line_number": 566,
+        "type": "Secret Keyword",
+        "verified_result": null
+      },
+      {
+        "hashed_secret": "2736fab291f04e69b62d490c3c09361f5b82461a",
+        "is_secret": false,
+        "is_verified": false,
+        "line_number": 573,
+        "type": "Secret Keyword",
+        "verified_result": null
+      },
+      {
+        "hashed_secret": "7542c8f677ffbbe75a5301a5c76f88401dc42897",
+        "is_secret": false,
+        "is_verified": false,
+        "line_number": 582,
+        "type": "Secret Keyword",
+        "verified_result": null
+      },
+      {
+        "hashed_secret": "ca20728d35f00c3a4c21b1fc8b0086b59f89df2d",
+        "is_secret": false,
+        "is_verified": false,
+        "line_number": 587,
+        "type": "Secret Keyword",
+        "verified_result": null
+      },
+      {
+        "hashed_secret": "5b7dcd14a4faa2cdd54cf6eb8d4bc35da31914a1",
+        "is_secret": false,
+        "is_verified": false,
+        "line_number": 599,
+        "type": "Secret Keyword",
+        "verified_result": null
+      },
+      {
+        "hashed_secret": "245dbfa0498aaa6898fe57a191a5b3130466a943",
+        "is_secret": false,
+        "is_verified": false,
+        "line_number": 620,
+        "type": "Secret Keyword",
+        "verified_result": null
+      },
+      {
+        "hashed_secret": "d033e22ae348aeb5660fc2140aec35850c4da997",
+        "is_secret": false,
+        "is_verified": false,
+        "line_number": 622,
+        "type": "Secret Keyword",
+        "verified_result": null
+      },
+      {
+        "hashed_secret": "ee977806d7286510da8b9a7492ba58e2484c0ecc",
+        "is_secret": false,
+        "is_verified": false,
+        "line_number": 625,
+        "type": "Secret Keyword",
+        "verified_result": null
+      },
+      {
+        "hashed_secret": "a603075604516de626cda903868e457fad826533",
+        "is_secret": false,
+        "is_verified": false,
+        "line_number": 632,
+        "type": "Secret Keyword",
+        "verified_result": null
+      },
+      {
+        "hashed_secret": "95ef18f58f32e3821ec7e627af6ee4757f68760e",
+        "is_secret": false,
+        "is_verified": false,
+        "line_number": 633,
+        "type": "Secret Keyword",
+        "verified_result": null
+      },
+      {
+        "hashed_secret": "b3aca92c793ee0e9b1a9b0a5f5fc044e05140df3",
+        "is_secret": false,
+        "is_verified": false,
+        "line_number": 671,
+        "type": "Secret Keyword",
+        "verified_result": null
+      },
+      {
+        "hashed_secret": "ae21c64a87f6bb0b8e16e55c48be4cc638d7bd3f",
+        "is_secret": false,
+        "is_verified": false,
+        "line_number": 682,
+        "type": "Secret Keyword",
+        "verified_result": null
+      },
+      {
+        "hashed_secret": "58d1bbce297de3c304a9fefc3b483181872a5c6b",
+        "is_secret": false,
+        "is_verified": false,
+        "line_number": 684,
+        "type": "Secret Keyword",
+        "verified_result": null
+      },
+      {
+        "hashed_secret": "de93ba2d15f3dbf65f76e6fd21e1e4b002459dd8",
+        "is_secret": false,
+        "is_verified": false,
+        "line_number": 716,
+        "type": "Secret Keyword",
+        "verified_result": null
+      },
+      {
+        "hashed_secret": "533739bb9d1dad53e71d8c93200cc2510d442226",
+        "is_secret": false,
+        "is_verified": false,
+        "line_number": 719,
         "type": "Secret Keyword",
         "verified_result": null
       }
@@ -4682,7 +4892,7 @@
         "hashed_secret": "4f13f134744a2fadbbe2d624687246347d12fa63",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 2805,
+        "line_number": 2804,
         "type": "Hex High Entropy String",
         "verified_result": null
       }
@@ -4692,7 +4902,7 @@
         "hashed_secret": "72cb70dbbafe97e5ea13ad88acd65d08389439b0",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1463,
+        "line_number": 1470,
         "type": "Secret Keyword",
         "verified_result": null
       }
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -4637,40 +4637,6 @@
         "verified_result": null
       }
     ],
-    "tests/unit/mcpgateway/services/test_oauth_authorization_code_health.py": [
-      {
-        "hashed_secret": "e9d687f46f18d07a5318605d07a020757e91f5e1",
-        "is_secret": false,
-        "is_verified": false,
-        "line_number": 77,
-        "type": "Secret Keyword",
-        "verified_result": null
-      },
-      {
-        "hashed_secret": "fe1bae27cb7c1fb823f496f286e78f1d2ae87734",
-        "is_secret": false,
-        "is_verified": false,
-        "line_number": 92,
-        "type": "Secret Keyword",
-        "verified_result": null
-      },
-      {
-        "hashed_secret": "0b7c1dbbf556a11d7627cd8c3fd8f7986bc86675",
-        "is_secret": false,
-        "is_verified": false,
-        "line_number": 197,
-        "type": "Secret Keyword",
-        "verified_result": null
-      },
-      {
-        "hashed_secret": "85c08fb0ff6bae7e56652ee15d4f049d67605dbd",
-        "is_secret": false,
-        "is_verified": false,
-        "line_number": 296,
-        "type": "Secret Keyword",
-        "verified_result": null
-      }
-    ],
     "tests/unit/mcpgateway/services/test_oauth_manager.py": [
       {
         "hashed_secret": "fe1bae27cb7c1fb823f496f286e78f1d2ae87734",
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -4637,6 +4637,50 @@
         "verified_result": null
       }
     ],
+    "tests/unit/mcpgateway/services/test_oauth_authorization_code_health.py": [
+      {
+        "hashed_secret": "e9d687f46f18d07a5318605d07a020757e91f5e1",
+        "is_secret": false,
+        "is_verified": false,
+        "line_number": 76,
+        "type": "Secret Keyword",
+        "verified_result": null
+      },
+      {
+        "hashed_secret": "fe1bae27cb7c1fb823f496f286e78f1d2ae87734",
+        "is_secret": false,
+        "is_verified": false,
+        "line_number": 91,
+        "type": "Secret Keyword",
+        "verified_result": null
+      },
+      {
+        "hashed_secret": "0b7c1dbbf556a11d7627cd8c3fd8f7986bc86675",
+        "is_secret": false,
+        "is_verified": false,
+        "line_number": 196,
+        "type": "Secret Keyword",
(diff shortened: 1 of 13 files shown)


omprak:
@ja8zyjits @bogdanmariusc10  Our production onboarding is blocked due to this critical issue . Assuming it is blocking every Oauth based MCP server any plan to release it as HotFix tag ?

Note - if you any have workaround solutions and approach IBM is doing for Oauth based MCP servers without this bug fix please let us know .

jonpspri:
Hi, @omprak -- for now, I'll schedule this for our 21 July release; however, we'll prioritize review and may release a hotfix.  Please continue to monitor the PR.  Thanks for reaching out.

jmccarrellmoe:
Confirming this exact scenario from a downstream deployment against the **Atlassian remote MCP** (`https://mcp.atlassian.com/v1/mcp`), registered as an `authorization_code` (per-user) gateway:

- User completes consent and the per-user token is stored — `/oauth/fetch-tools/{id}` returns 200 using that token. 
- But the periodic health check connects at the gateway/connection level with **no service token**, so the upstream returns **401**, the gateway is marked unhealthy and **deactivated after `unhealthy_threshold` (3)**, and its tool list is wiped to 0.
- Raising `UNHEALTHY_THRESHOLD` / `HEALTH_CHECK_INTERVAL` only delays it — the unauthenticated probe 401s indefinitely.

This PR's approach — decoupling health checks from token ownership and treating 401/403 as "reachable, unauthorized" — is exactly the fix for per-user-only upstreams like Atlassian. +1 for getting this into a release; happy to help validate against the Atlassian remote MCP.

prakhar-singh1928:
## Follow-up fixes addressing review findings

This comment documents the additional changes made after the initial PR
submission, resolving all blocking and significant issues raised in review.

---

### Fix 1 — `invalid_grant` detection was fragile (substring → typed exception)

**Problem:** `_refresh_access_token` checked `"invalid_grant" in str(e).lower()`
to decide whether to delete a stored token. This was fragile: the match
depended on the RFC 6749 error code surviving message formatting and
`_redact_token_response`. A provider returning a non-JSON body or a
future change to the error message would silently break token deletion.

**Fix:** Added `OAuthInvalidGrantError(OAuthError)` typed subclass in
`oauth_manager.py`. `refresh_token()` now reads the structured `error`
field from the parsed response **before** redaction and raises
`OAuthInvalidGrantError` only when `error == "invalid_grant"` (exact
RFC 6749 §5.2 field). `_refresh_access_token` catches
`OAuthInvalidGrantError` first (delete token), then `OAuthError`
(preserve token). Type-based, not string-based.

`OAuthInvalidGrantError` is a subclass of `OAuthError` — all existing
`except OAuthError` callers are unaffected.

---

### Fix 2 — `is_encrypted()` used on a security-critical path

**Problem:** The PR used `encryption.is_encrypted(client_secret)` to
decide whether to attempt decryption. That method's own docstring
carries a `⚠️ SECURITY WARNING: NOT suitable for security-critical code
paths`. A false-negative (encrypted secret not matched by the heuristic)
would silently send ciphertext as the credential, producing an opaque
`invalid_client` error with no diagnostic signal.

**Fix:** Removed the heuristic. Always call `decrypt_secret_async()`.
On failure, log a `WARNING` naming the gateway and pointing at
`AUTH_ENCRYPTION_SECRET`, then proceed with the raw value. This is
consistent with how refresh-token decryption directly above it works,
and handles both API-registered (plaintext) and UI-registered (encrypted)
secrets correctly without guessing.

---

### Fix 3 — 401/403 reactivation was not scoped to `authorization_code`

**Problem:** The new exception handler treated 401/403 as "gateway
reachable" for **any** grant type. A `client_credentials` gateway with
a rotated secret would be kept "reachable" indefinitely and
`_handle_gateway_failure` would never fire, hiding the outage from
operators.

**Fix:** Added `is_authorization_code` guard:
```python
is_authorization_code = (
    gateway_oauth_config is not None
    and gateway_oauth_config.get("grant_type") == "authorization_code"
)

**Review by msureshkumar88 — CHANGES_REQUESTED** [review]
Thanks for tackling this — the core fix (decoupling health checks from token ownership, treating 401/403 as "reachable," and moving to a typed `OAuthInvalidGrantError` instead of substring-matching on the error message) is a real improvement and well covered by the new tests. A couple of things worth resolving before merge, plus a few smaller notes.

## Blocking

**1. `client_secret` decryption failure doesn't actually raise (`mcpgateway/services/token_storage_service.py:298-309`)**

The PR description's "before/after" shows decrypt failures raising `OAuthError` explicitly instead of silently falling back. The actual code still falls back — it just adds a `logger.warning` around the same `except: pass`-style behavior:

```python
try:
    oauth_config["client_secret"] = await self.encryption.decrypt_secret_async(client_secret_value)
except Exception as decrypt_error:
    logger.warning(...)   # falls through, oauth_config["client_secret"] keeps the raw/ciphertext value
```

This means a misconfigured `AUTH_ENCRYPTION_SECRET` will keep sending the ciphertext envelope as the literal `client_secret` to the external Authorization Server on every refresh attempt — worth treating as security-hardening territory (CWE-703, improper handling of an exceptional condition): some IdPs apply lockout/rate-limiting on repeated invalid client-auth attempts, so this can turn into a self-inflicted lockout against the upstream provider that's hard to diagnose from our side. Recommend failing closed (raise, skip the attempt) rather than failing open. The new test `test_decryption_failure_raises_explicit_error` doesn't actually assert a raise (no `pytest.raises`), so it doesn't catch this gap either — worth tightening once the behavior is fixed.

**2. `streamablehttp` transport isn't covered by the 401/403-as-healthy fix (`mcpgateway/services/gateway_service.py:4780`)**

```python
if is_authorization_code and hasattr(e, "response") and hasattr(e.response, "status_code"):
```

This works for `sse` transport (synchronous `httpx.HTTPStatusError` from `client.stream(...)`). For `streamablehttp`, the request goes through the MCP SDK's `streamablehttp_client`, which spawns the POST via `tg.start_soon(...)` inside an `anyio` task group. I checked empirically — exceptions raised inside a task spawned via `anyio.create_task_group()` surface at the `async with` boundary as an `ExceptionGroup`, not the original exception, so `hasattr(e, "response")` is `False` there. That means for `streamablehttp`-transport `authorization_code` gateways, a 401/403 still falls through to the old "genuine failure" path and calls `_handle_gateway_failure` — reproducing the exact bug this PR is meant to fix, just for a different transport. None of the new tests exercise `streamablehttp`, so this isn't currently caught. Suggest unwrapping `ExceptionGroup`/`BaseExceptionGroup` (or catching `httpx.HTTPStatusError` at the streamablehttp call site) and adding a test for that transport.

## Non-blocking / worth a look

- **`omit_resource` fix in `token_storage_service.py` (lines 347-351) may be redundant.** `oauth_manager.py`'s `_should_include_resource_parameter()` — unchanged by this PR, already on `main` — already gates on `credentials.get("omit_resource")` using the original `oauth_config` dict before adding `resource` to the refresh POST body. As far as I can trace it, a refresh request would never have included `resource` when `omit_resource=True`, even before this change. Worth double-checking against the original IdP failure that motivated this fix, in case the real gap was elsewhere (e.g. the initial auth-code exchange) rather than in refresh.
- `TestAuthorizationCodeHealthCheck::test_health_check_behavior_documented` in `test_oauth_authorization_code_health.py` is a no-op (`assert True` with a docstring pointing elsewhere) — either remove it or replace with a real assertion so it doesn't inflate the test count without adding coverage.
- Checklist says "Documentation updated" but no `docs/`, `CHANGELOG.md`, or `.env.example` changes are in the diff — a short note in the OAuth troubleshooting docs about the new `invalid_grant`-only deletion behavior would help operators.
- Minor duplication: the reactivation + `last_seen` update block in the new `except` branch is a near-verbatim copy of the success-path block just above it — could be factored into a small helper to avoid drift.

Happy to take another pass once these are addressed.

msureshkumar88:
Re-checked the latest commit (`4de38d66a`) against my earlier review — both blocking items are resolved with real fixes and test coverage, not just cosmetic changes:

## Blocking — resolved ✅

**1. `client_secret` decryption fail-closed** (`token_storage_service.py:296-309`)
Decrypt failures now raise `OAuthError` instead of silently falling back to sending ciphertext as the credential. The refresh path catches this via a new `except OAuthError` branch that preserves the token and returns `None`. `test_decryption_failure_raises_oauth_error_and_preserves_token` now asserts the actual behavior (`result is None`, `delete` not called) rather than a no-op.

**2. `streamablehttp` transport 401/403 handling** (`gateway_service.py`)
The health-check exception handler now unwraps `BaseExceptionGroup` one level before inspecting `.response.status_code`, so 401/403 raised inside the MCP SDK's `anyio` task group is correctly treated as "reachable, unauthorized" instead of falling through to `_handle_gateway_failure`. Covered by the new `test_streamablehttp_401_treated_as_reachable`.

Also nice: the substring-matching on error messages (`"invalid" in str(e).lower()`) that caused the original destructive-deletion bug has been replaced with a typed `OAuthInvalidGrantError` subclass raised only on an explicit `{"error": "invalid_grant"}` response — a more robust discriminator than the original fix used.

I pulled the branch, ran the full affected suite (695 tests) — all green — and ran `ruff` on the three changed core files — clean.

## Non-blocking — still open (optional, doesn't block from my side)

- `omit_resource` fix in `token_storage_service.py` (lines ~347-351) may still be redundant with `oauth_manager.py`'s existing `_should_include_resource_parameter()` gate — worth a quick double-check against the original IdP failure that motivated it, in case the real gap was in the initial auth-code exchange rather than refresh.
- Checklist says "Documentation updated" but there's no `docs/`/`CHANGELOG.md` change in the diff — a short note on the new `invalid_grant`-only deletion semantics would help operators who may be used to the old broad-match behavior clearing stuck tokens.
- Minor: `test_oauth_authorization_code_health.py` line ~1153 has `mock_db.delete.assert_not_called(), "message"` — the trailing comma makes it a discarded tuple rather than an assert-with-message (harmless, `assert_not_called()` still fires, just a bit confusing to read).

## Merge state

`gh` currently reports this as not mergeable against `main`. I tested a real merge in a scratch worktree — the only conflict is `.secrets.baseline` (timestamp/line-count drift); every source and test file auto-merges cleanly. Will need a rebase + `make detect-secrets-scan` before merge, unrelated to the OAuth fix itself.

Holding off on approving until the two optional items above get a look, but from a correctness/security standpoint the blocking concerns are addressed.

msureshkumar88 (changes requested):
Built a real e2e verification script for this PR (not mocks) and ran it against the actual PR head — `scripts/verify-pr-5244-e2e.sh` (added to this run, not committed). **12 of 13 checks passed; 1 real gap found**, verified against the actual (unmocked) `EncryptionService`, not a unit-test double.

## Setup (as close to production as practical)

- Real `uv run uvicorn mcpgateway.main:app` process, real SQLite DB, real JWTs via `create_jwt_token`, real `curl` calls.
- A real fake upstream server (plain stdlib `http.server`, not `unittest.mock`) acting as both the protected MCP endpoint (always 401, no valid credentials) and the OAuth token endpoint (behavior switchable via a mode file: `success` / `invalid_grant` / `invalid_client`), logging every real HTTP request it receives.
- 4 real `authorization_code` gateways registered via `POST /gateways` (sse, streamablehttp, omit_resource=true, omit_resource=false).
- `HEALTH_CHECK_INTERVAL=3`, `UNHEALTHY_THRESHOLD=3`, `GATEWAY_HEALTH_CHECK_TIMEOUT=3` so the background scheduler actually cycles multiple times during the run.
- Config for the decrypt-failure test: gateway registered under `AUTH_ENCRYPTION_SECRET=SECRET_A`, then the gateway process is **killed and restarted with a different `AUTH_ENCRYPTION_SECRET=SECRET_B`** — a real "operator misconfigured/rotated the encryption secret" scenario, not a mocked exception.

Command: `bash scripts/verify-pr-5244-e2e.sh`

## Passed (12/13) — core fixes verified working for real

- SSE transport: 401 with no stored token is treated as healthy (gateway stays `reachable=true`/`enabled=true` through 3+ health-check cycles) — the original issue #5237 symptom, gone.
- **streamablehttp transport: same 401 treated as healthy** — this is the specific gap from the PR's last commit (`BaseExceptionGroup` unwrap); confirmed working against a real MCP-SDK `streamablehttp_client` call, not a hand-built exception.
- Gateway log shows the actual "reachable but lacks valid credentials" debug line fired — confirms the fix's code path, not just a coincidentally-healthy gateway.
- Reactivation: forcing a gateway to `reachable=false` in the DB, then a real 401 auth-challenge cycle flips it back to `reachable=true` with the expected log line.
- `invalid_client` from a real token endpoint response → token **preserved**.
- `invalid_grant` from a real token endpoint response → token **deleted**, with the typed `OAuthInvalidGrantError` log line present (confirms substring-matching was actually replaced, not just relabeled).
- `omit_resource=true` → the real outbound refresh POST body has **no** `resource` param. `omit_resource` absent → `resource` **is** present. This also answers my earlier non-blocking question about whether the `omit_resource` fix was redundant with `oauth_manager.py`'s existing gate — empirically it is **not** redundant; the flag has a real, observable effect on the live outbound request.

## Failed (1/13) — client_secret decrypt-failure does not actually raise under a real wrong-key scenario

This is the exact scenario from the PR description ("misconfigured `AUTH_ENCRYPTION_SECRET`"), reproduced for real (gateway restarted with the wrong secret, not mocked):

- Token was preserved (not deleted) — good, matches intent.
- But the gateway log **never contains** the PR's intended message (`"client_secret decryption failed for gateway ..."` from `token_storage_service.py`'s `raise OAuthError(...)`).
- Instead, the log shows repeated `encryption_service.py` lines: `Failed to decrypt secret: Decryption failed (corrupted or wrong key):`
- And the real outbound POST to the fake IdP's `/token` endpoint was captured with **no `client_secret` param at all**:
  ```
  POST /token BODY=grant_type=refresh_token&refresh_token=old_refresh_token_plaintext&client_id=e2e-client-id&resource=http%3A%2F%2Flocalhost%3A44911%2Fmcp-sse
  ```

**Root cause:** `token_storage_service.py`'s new code calls `self.encryption.decrypt_secret_async(...)`, which is `EncryptionService`'s *idempotent* wrapper (`decrypt_secret_or_plaintext`). That method catches decryption exceptions **internally** and returns `None` — it never raises for a wrong-key/corrupted-ciphertext failure (see `encryption_service.py`, `decrypt_secret_or_plaintext`: `except Exception as e: logger.error(...); return None`). So the new `try/except Exception as decrypt_error: raise OAuthError(...)` never actually observes an exception for this case — `oauth_config["client_secret"]` silently becomes `None` instead. Then in `oauth_manager.py`'s `refresh_token()`, a falsy `client_secret` is simply omitted from the request (`if client_secret: token_data["client_secret"] = client_secret`), so the real request to the Authorization Server silently drops client authentication rather than failing loudly and closed as the PR describes.

Net effect: not the original bug (ciphertext-as-credential) and the token isn't deleted, so it's not unsafe — but it's also not the explicit fail-closed behavior the PR claims, and it's a real behavior change (silently downgrading to an unauthenticated/public-client-style refresh request) that isn't logged anywhere traceable to "this gateway's client_secret is undecryptable." The existing unit test (`test_decryption_failure_raises_oauth_error_and_preserves_token`) doesn't catch this because it mocks `decrypt_secret_async` with a `side_effect` that raises directly, which isn't how the real method behaves on decrypt failure.

**Suggested fix:** after calling `decrypt_secret_async()`, explicitly check for `None` (or call the strict `is_encrypted()` + `decrypt_secret()` pair instead of the idempotent wrapper) and raise `OAuthError` in that case — e.g.:
```python
decrypted = await self.encryption.decrypt_secret_async(client_secret_value)
if decrypted is None:
    raise OAuthError(f"client_secret decryption failed for gateway {token_record.gateway_id}: decryption returned None (wrong AUTH_ENCRYPTION_SECRET or corrupted ciphertext)")
oauth_config["client_secret"] = decrypted
```

Requesting changes for this one item — everything else in this PR checks out against a real, unmocked run. Happy to re-verify once it's addressed.

msureshkumar88 (dismissed):
## E2E verification — closes #5237

Ran the full e2e suite against a real running gateway (real `uv run uvicorn` process, real SQLite DB, real JWTs, real curl calls) driving a real fake upstream OAuth/MCP server (plain stdlib `http.server`, no mocks), covering both `sse` and `streamablehttp` transports.

Commit under test: `acb9a54e9` (fixed dead try/except issue)

### Config

```
DATABASE_URL=sqlite:///./e2e_pr5244_test.db
GW_PORT=44644  FAKE_PORT=44911
HEALTH_CHECK_INTERVAL=3  UNHEALTHY_THRESHOLD=3  GATEWAY_HEALTH_CHECK_TIMEOUT=3
AUTH_REQUIRED=true
SSRF_ALLOW_LOCALHOST=true SSRF_ALLOW_PRIVATE_NETWORKS=true
GATEWAY_TEST_ALLOW_REGISTERED_ONLY=false
GATEWAY_TEST_ALLOWED_HOSTS=["127.0.0.1"]
AUTH_ENCRYPTION_SECRET phase 1 = SECRET_A (registration/refresh happy-path)
AUTH_ENCRYPTION_SECRET phase 2 = SECRET_B (WRONG, restart mid-run for the client_secret decrypt-failure case)
```

### Command

```
bash scripts/verify-pr-5244-e2e.sh
```

### Scenarios exercised

1. `authorization_code` gateway health check with **no stored token**, upstream always 401 — must stay `reachable=true, enabled=true` (both SSE and streamablehttp transports; streamablehttp is the one that surfaces 401 wrapped in a `BaseExceptionGroup` via the MCP SDK's anyio TaskGroup, the last gap this PR closes).
2. Forced `reachable=false` → next health-check cycle flips it back on a 401 auth-challenge.
3. Token refresh: `invalid_client` → token **preserved**; `invalid_grant` → token **deleted** (typed `OAuthInvalidGrantError`, not substring matching).
4. `omit_resource=true` on `oauth_config` → outbound refresh omits the `resource` param; absent/false → `resource` param present (RFC 8707 default unaffected).
5. `client_secret` decryption failure under a genuinely wrong `AUTH_ENCRYPTION_SECRET` (real process restart, not a mock) → token preserved, explicit `client_secret decryption failed for gateway ...` error logged, no ciphertext ever sent to the fake IdP.

### Output

```
PASS: fake upstream server is up on :44911 and returns 401 for the protected MCP endpoint
PASS: gateway server is up on :44644
PASS: admin JWT generated
PASS: registered 4 authorization_code gateways (sse/stream/omit/noomit)
PASS: SSE transport: 401 with no token treated as healthy (gateway stayed reachable+enabled through 3+ cycles)
PASS: CORE FIX: streamablehttp transport: 401 (wrapped in BaseExceptionGroup) treated as healthy
PASS: gateway log confirms the 401-as-healthy code path was actually exercised
PASS: gateway reactivated (reachable flipped back to true) after next 401 auth-challenge cycle
PASS: gateway log shows the specific reactivation-reason string from _mark_gateway_reachable()
PASS: invalid_client: token preserved (not deleted)
PASS: invalid_grant: token deleted (typed OAuthInvalidGrantError path)
PASS: gateway log confirms typed OAuthInvalidGrantError path was used (not substring matching)
PASS: omit_resource=true: 'resource' param absent from real outbound refresh request
PASS: omit_resource absent: 'resource' param IS present (default RFC 8707 injection behavior unaffected)
PASS: gateway restarted with SECRET_B
PASS: wrong-key scenario: token preserved (not deleted) — matches PR's fail-closed intent
PASS: gateway log contains the PR's intended explicit error message ('client_secret decryption failed for gateway ...')

RESULT: ALL CHECKS PASSED (18/18)
```

### Notes

An earlier pass of this suite surfaced 4 failures; all traced back to test-harness races (background health-check refresh cycles interleaving with manual triggers, plus a shared literal `refresh_token` string across seeded gateways making log-based assertions ambiguous) rather than product bugs — fixed by scoping assertions to gateway-specific markers instead of `tail -1` on a shared log. The `client_secret` decrypt-failure path (`token_storage_service.py`) was additionally verified in isolation against a real wrong-key scenario to confirm the raise/catch/log behavior fires as intended.

Approving based on this run.

ja8zyjits:
HI @prakhar-singh1928 

Happy to approve if you deem the following as not important to be fixed in this PR

## 🚨 Blocking Issues

### 1. Health Check Behavior Undocumented
**Code**: `gateway_service.py:4639-4665, 4789-4835`
- Health checks proceed without auth when no token for `authorization_code` gateways
- 401/403 treated as "gateway reachable" (auth challenge, not connectivity failure)
- Previously-unreachable gateways reactivated on 401/403

**Docs**: Missing from `oauth.md`, `oauth-troubleshooting.md`, ADR 009

**Fix**: Add to `oauth.md`:
- Health checks verify service reachability, not token ownership
- Missing tokens don't mark gateways unhealthy (expected for user-specific auth)
- 401/403 = "gateway reachable but unauthorized" (healthy state)
- Reactivation on 401/403 responses

**Impact**: Operators confused why `authorization_code` gateways show offline when `PLATFORM_ADMIN_EMAIL` user hasn't authorized

### 2. Token Deletion Policy Undocumented
**Code**: `token_storage_service.py:438-450, 451-461`, `oauth_manager.py:1806-1812`
- Only `OAuthInvalidGrantError` (RFC 6749 §5.2 `invalid_grant`) deletes tokens
- All other `OAuthError` preserve tokens for retry
- Config errors (wrong `client_secret`, wrong `AUTH_ENCRYPTION_SECRET`) don't delete tokens

**Docs**: `oauth.md` states "Auto-refresh" but no deletion policy. `oauth-design.md` vague on which errors trigger deletion

**Fix**: Add to `oauth.md` "Token Storage and Refresh":
- Tokens deleted only on RFC 6749 §5.2 `invalid_grant` (permanent failure)
- Transient errors (`invalid_client`, `invalid_request`, network) preserve tokens
- Config errors don't delete tokens

**Impact**: Operators expect deletion on any refresh failure, confused when tokens persist after config errors

### 3. Decryption Failure Behavior Undocumented
**Code**: `token_storage_service.py:298-313`
- Decryption failure raises explicit `OAuthError` (fail-closed)
- Error message explains `AUTH_ENCRYPTION_SECRET` mismatch
- Tokens preserved on decryption failure

**Docs**: `oauth.md` mentions "encrypted at rest" but no decryption failure behavior. `configuration.md:L49` mentions `AUTH_ENCRYPTION_SECRET` for DCR only

**Fix**: Add to `oauth.md` "Security Recommendations":
- `client_secret` always decrypted before use (no plaintext fallback)
- Decryption failures raise explicit errors with `AUTH_ENCRYPTION_SECRET` guidance
- Changing `AUTH_ENCRYPTION_SECRET` invalidates all stored secrets
- Tokens preserved on decryption failure (config issue, not token issue)

**Impact**: Operators don't understand why refresh fails after changing `AUTH_ENCRYPTION_SECRET`

---

## ⚠️ Warnings

### 1. `omit_resource` Flag Undocumented
**Code**: `token_storage_service.py:350-373` - prevents resource parameter injection when `true`

**Docs**: Missing from `oauth.md`. `oauth-design.md:L91` mentions resource derivation but not `omit_resource`

**Fix**: Add to `oauth.md` "Configure Gateway":
```json
{
  "oauth_config": {
    "omit_resource": true  // Prevents resource param (for IdPs that reject it)
  }
}
```

**Impact**: Low - works but not discoverable

### 2. `PLATFORM_ADMIN_EMAIL` Context Missing
**Code**: `gateway_service.py:4654-4656` - used for token lookup in health checks

**Docs**: `configuration.md:L18` describes as "bootstrap admin" only. No OAuth context in `oauth.md`

**Fix**: Add to `configuration.md` `PLATFORM_ADMIN_EMAIL`:
- "Also used as default user identity for OAuth gateway health checks (authorization_code flow)"
- "Health checks proceed without auth if user hasn't authorized gateway"

**Impact**: Low - correct behavior but operators may not understand relationship

---

msureshkumar88 (dismissed):
Re-checked against current HEAD (`41295f8a6`) after the rebase that dismissed my prior approval.

Both blocking items from my first review are fixed and match the intended behavior:

- **`client_secret` decrypt failure now fails closed** (`token_storage_service.py:301-312`): explicit `None` check on `decrypt_secret_async()`'s return (since that wrapper returns `None` rather than raising on a wrong key / corrupted ciphertext) and raises `OAuthError` instead of silently falling through with the ciphertext. Matches what my e2e run against a real wrong-`AUTH_ENCRYPTION_SECRET` restart required.
- **`streamablehttp` 401/403 handling** (`gateway_service.py:4797-4835`): unwraps `BaseExceptionGroup` one level to recover the underlying `httpx.HTTPStatusError` before inspecting `status_code`, so the auth-challenge-as-healthy path now applies to both transports, not just `sse`.

Non-blocking notes from before are also addressed: OAuth docs added (`docs/docs/manage/oauth.md`), and the no-op `test_health_check_behavior_documented` test is gone/replaced with real assertions.

No code changes to `gateway_service.py` / `token_storage_service.py` since the commit I e2e-verified (18/18 passing) — the commits after that were docs + `.secrets.baseline`/pre-commit fixups only, so the verified behavior stands. CI is still cycling on the latest push but nothing has failed so far.

Approving.

ja8zyjits (dismissed):
LGTM

msureshkumar88 (approved):
LGMT

omprak:
@msureshkumar88 @ja8zyjits We are waiting for this issue fix as its breaking our production . Can you please confirm if this fix is available in 1.0.6 or in any stable tag?

ja8zyjits:
Hi @omprak ,

v1.0.6 wont have this changes, we couldnt get it in for the release, our review and verification took longer.
Please use the latest main for this change in your product.
https://github.com/ibm/mcp-context-forge/pkgs/container/mcp-context-forge/1059518965?tag=5713c1a4bc8c1f4f43320d1ebb75cde6c952da4d


In v1.0.7 you will get these changes.

Regards,
Jitesh

PR #5263, opened 2026-06-16
feat: edit virtual servers

### LINKED ISSUE #4180: [UI-REWRITE] Edit Virtual  servers

vishu-bh:
# ✨ Feature / Enhancement PR

## 🔗 Epic / Issue
Closes #4180

---

## 🚀 Summary (1-2 sentences)
Adds an edit flow for virtual servers in the React UI, including editable details, connected MCP server review, and source association updates. The edit handoff stores only the server id and reloads current server data before submitting changes.

---

## 🧪 Checks

- [ ] `make lint` passes
- [ ] `make test` passes
- [x] `./node_modules/.bin/eslint src/pages/CreateServer.tsx src/pages/Gateways.tsx src/components/gateways/CreateServerForm.tsx src/api/virtualServers.ts src/api/virtualServers.test.ts src/pages/CreateServer.test.tsx src/pages/Gateways.test.tsx`
- [x] `npm run test:run -- src/api/virtualServers.test.ts src/pages/CreateServer.test.tsx src/pages/Gateways.test.tsx`
- [x] `npm run e2e`

---

## 📓 Notes (optional)
The MCP server selector uses the existing shadcn table and checkbox components for accessible row semantics. Edit mode avoids persisting the full server payload in browser storage and no longer fans out component fetches for every available MCP server.


a-effort:
Nice! 

LLM advises these items are addressed: 

1. Icon wiped on every edit

buildUpdateVirtualServerPayload hardcodes icon: "", and the backend treats any non-None value as an explicit write (server_service.py:1287-1288). The admin UI exposes an Icon URL field, so servers with an icon set there will lose it on a single React-UI save. Simplest fix: drop icon from UpdateVirtualServerPayload. ServerUpdate.icon defaults to None, so omitting it preserves the stored value without committing the React form to icon support.

2. OAuth config wiped on every edit

The form has an oauth_enabled toggle but no inputs for the config object (authorization_server, scopes_supported, etc.). Despite that, the payload sends oauth_config: details.oauthEnabled ? {} : null unconditionally. For the enabled-with-empty-config case, the service writes the {} through model_fields_set (server_service.py:1329-1330) and protect_oauth_config_for_storage({}, existing) returns {}, so any populated config on the server is overwritten. Suggest dropping oauth_config from the update payload entirely — the disable path still clears config via the oauth_being_disabled branch (line 1324), and the enable path preserves whatever was there for an admin-UI/API config flow to manage.

client/src/pages/Gateways.tsx
```diff
@@ -122,6 +123,15 @@ export function Gateways() {
     setIsDetailsPanelOpen(true);
   };
 
+  const openEditPanel = (server: VirtualServer) => {
+    try {
+      window.sessionStorage.setItem(EDIT_SERVER_STORAGE_KEY, server.id);
```
marekdano: Do we want to use `sessionStorage`? Is it possible to use the Context API from React?

marekdano (changes requested):
## Bugs / Correctness

  ### B1 — `icon` hardcoded to `""` clears existing icons

  **`client/src/api/virtualServers.ts:127`**

  ```ts
  icon: "",
  ```

  `buildUpdateVirtualServerPayload` always sends an empty string for `icon`. If a server previously
  had an icon set, every save will erase it. Either thread the existing icon through
  `CreateServerDetails`, or omit the field entirely if the edit form does not expose it.

  ---

  ### B2 — `oauth_config: {}` silently overwrites existing OAuth credentials

  **`client/src/api/virtualServers.ts:135`**

  ```ts
  oauth_config: details.oauthEnabled ? {} : null,
  ```

  If a server had OAuth credentials (client ID, secret, token endpoints) stored, editing anything — even just the server name — submits `oauth_config: {}` and destroys the existing config. This is a silent, destructive write of security-sensitive configuration. Either pass the existing config through, or omit the field if the edit form has no OAuth config fields.

  ---

  ### B3 — `undefined` association fields may clear server associations on the backend

  The tests confirm that when `associatedTools` is not provided the payload includes
  `associated_tools: undefined`. Verify that the backend `PUT /servers/{id}` handler treats missing association fields as "no change" rather than "clear all associations." If the backend interprets `undefined` as "set to empty," every edit that does not explicitly select sources will wipe the server's component list.

  ---

### A1 — Hardcoded `id` on heading violates uniqueness and breaks `aria-labelledby` (Medium)

  **`CreateServer.tsx:1001`**

  ```tsx
  <h2 id="connected-mcp-servers-heading" ...>
  ```

  HTML IDs must be unique per document. If `EditMCPServersSection` is ever rendered more than once, the `aria-labelledby` on the `<section>` will point to the wrong element and screen reader landmark navigation will break. Use the React `useId()` hook:

  ```tsx
  const headingId = useId();
  // <section aria-labelledby={headingId}>
  // <h2 id={headingId}>
  ```

  ---

  ### A2 — "Available MCP servers" subsection is not a navigable landmark (Medium)

  **`CreateServer.tsx:1029–1050`**

  ```tsx
  <div className="mt-8 border-t border-border pt-5">
    <h3 className="text-sm font-semibold text-foreground">
      {intl.formatMessage({ id: "gateways.editServer.availableMCPServers" })}
    </h3>
  ```

  "Connected MCP servers" is a `<section aria-labelledby>` landmark, reachable via screen reader region navigation. "Available MCP servers" is inside a `<div>` — invisible as a landmark. Wrap it in a `<section>` with its own `aria-labelledby`.

  ---

  ### A3 — Component counts are unlabelled numbers for screen readers (Medium)

  **`CreateServer.tsx:957–972`**

  ```tsx
  <Wrench className="size-3.5" aria-hidden="true" />
  {getToolCount(server)}
  ```

  Icons are correctly `aria-hidden`, but the adjacent count numbers have no label. A screen reader announces something like `"3 • 1 • 2"` with no context. Fix with accessible labels:

  ```tsx
  <span aria-label={`${getToolCount(server)} tools`}>{getToolCount(server)}</span>
  <span aria-label={`${getResourceCount(server)} resources`}>{getResourceCount(server)}</span>
  <span aria-label={`${getPromptCount(server)} prompts`}>{getPromptCount(server)}</span>
  ```

  Or collapse to a single `aria-label` on the containing table cell.

---

client/src/pages/CreateServer.tsx
```diff
@@ -34,14 +209,423 @@ function getCreateServerError(error: unknown, fallbackMessage: string): string {
...
+    tools: [],
+    resources: [],
+    prompts: [],
+  });
   const [isCreating, setIsCreating] = useState(false);
+  const [isUpdating, setIsUpdating] = useState(false);
   const [createError, setCreateError] = useState<string | null>(null);
+  const [updateError, setUpdateError] = useState<string | null>(null);
+  const isEditMode = Boolean(editServerId);
+  const encodedEditServerId = encodeURIComponent(editServerId ?? "__pending__");
```
marekdano: `useQuery` is disabled when `!isEditMode`, so this never fires in practice - but if the enabled guard is ever loosened it will make a real request to `/servers/__pending__`. A cleaner pattern is to conditionally build the path:
```tsx
  const editServerPath = editServerId ? `/servers/${encodeURIComponent(editServerId)}` : null;

// below to have something like
  useQuery(editServerPath, { enabled: Boolean(editServerPath) });
```

marekdano (commented):
If we go with using `sessionStorage`

`EDIT_SERVER_STORAGE_KEY = "gateways.editServer"` is defined independently in both `Gateways.tsx` and `CreateServer.tsx`. A typo in either file would silently break the edit handoff. Extract it to a shared file (e.g. `src/api/virtualServers.ts` or a constants module).

marekdano (approved):
It looks good!

LGTM 🚀

PR #5269, opened 2026-06-17
fix(helm): resolve Helm chart linting issues

cafalchio:
## Description

This PR resolves Helm chart linting issues identified by kube-linter and checkov security scanners.

## Changes

- Added security contexts to all deployment templates (mcpgateway, postgres, redis, fast-time-server)
- Added resource limits and requests to all containers
- Configured appropriate securityContext settings:
  - runAsNonRoot: true
  - allowPrivilegeEscalation: false
  - readOnlyRootFilesystem: true (where applicable)
  - capabilities drop: ALL
- Added checkov skip annotations for rules that cannot be satisfied in current architecture
- Re-enabled helm chart linting in CI/CD pipeline

## Testing

- ✅ Helm chart linting passes
- ✅ kube-linter security checks pass
- ✅ checkov security scans pass with documented exceptions

## Closes

Closes the internal issue https://github.ibm.com/contextforge-org/internal_issues/issues/252

---

Signed-off-by: cafalchio <mcafalchio@gmail.com>


charts/mcp-stack/values.yaml
```diff
@@ -85,8 +85,8 @@ mcpContextForge:
 
   image:
     repository: ghcr.io/ibm/mcp-context-forge
-    tag: latest                 # PRODUCTION: pin a specific immutable tag (e.g. "1.0.0")
-    pullPolicy: IfNotPresent    # use Always only for dev/testing with mutable tags
+    tag: "1.0.3"                # PRODUCTION: pin a specific immutable tag (e.g. "1.0.0")
```
brian-hussey:
```suggestion
    tag: "v1.0.5"                # PRODUCTION: pin a specific immutable tag (e.g. "1.0.0")
```

The only thing wrong is I think this should be the latest versioned tag by default. Currently "v1.0.5".
Will you update the update steps of release doc to include changing the version here as well please?

charts/mcp-stack/values.yaml
```diff
@@ -1043,12 +1044,13 @@ migration:
...
   backoffLimit: 3                           # Retry up to 3 times before giving up
   activeDeadlineSeconds: 600                # Kill job after 10 minutes
+  ttlSecondsAfterFinished: 300              # Clean up completed job after 5 minutes
 
   # Use same image as mcpgateway
   image:
     repository: ghcr.io/ibm/mcp-context-forge
-    tag: latest                             # Should match mcpContextForge.image.tag
-    pullPolicy: IfNotPresent                # use Always only for dev/testing with mutable tags
+    tag: "1.0.3"                            # Should match mcpContextForge.image.tag
```
brian-hussey:
```suggestion
    tag: "v1.0.5"                            # Should match mcpContextForge.image.tag
```

brian-hussey: Same as above.

brian-hussey (changes requested):
Nice work, only one change required in 2 places and we're good to go.

charts/mcp-stack/values.yaml
cafalchio: :)

charts/mcp-stack/values.yaml
cafalchio: thanks, I have addressed them

brian-hussey (dismissed):
@cafalchio Thank you for making the changes, this looks good. Approved

brian-hussey (dismissed):

brian-hussey (dismissed):
LGTM

charts/mcp-stack/values.yaml
```diff
@@ -165,7 +165,7 @@ mcpContextForge:
     capabilities:
       drop:
         - ALL
-    readOnlyRootFilesystem: false  # Set to true if app doesn't need write access
+    readOnlyRootFilesystem: true  # Read-only root filesystem for security
```
Yosiefeyob: The gateway container has no writable /tmp, and it dies on startup before gunicorn even loads: /app/run-gunicorn.sh: line 124: /tmp/mcpgateway-gunicorn.lock: Read-only file system

You've already got the right pattern for this — deployment-postgres.yaml, deployment-redis.yaml and job-migration.yaml all get a tmp emptyDir. The gateway is the only one that didn't. Its volumeMounts block in deployment-mcpgateway.yaml is wrapped in {{- if .Values.mcpContextForge.pluginConfig.enabled }}, which is false by default, so nothing renders.

Adding the same emptyDir fixes it and lets readOnlyRootFilesystem: true stay on:

          volumeMounts:
            - name: tmp
              mountPath: /tmp
      volumes:
        - name: tmp
          emptyDir: {}

Verified on a real OpenShift install — with this, the gateway comes up and /health and /ready both return 200.

charts/mcp-stack/values.yaml
```diff
@@ -1485,8 +1489,8 @@ mcpFastTimeServer:
   replicaCount: 2
   image:
     repository: ghcr.io/ibm/fast-time-server
-    tag: "latest"
-    pullPolicy: IfNotPresent
+    tag: "1.0.3"
```
Yosiefeyob: 1.0.3 is a gateway release ([v1.0.3](https://github.com/IBM/mcp-context-forge/releases#release-v1.0.3) — Auth & JWT Cleanup, Admin UI Fixes). It doesn't mention fast-time-server, because that component isn't versioned with the gateway — its newest published tag is 0.8.0 ([package page](https://github.com/IBM/mcp-context-forge/pkgs/container/fast-time-server)).

So no 1.0.3 image to pull, and since mcpFastTimeServer.enabled defaults to true, every default install hits it. Confirmed on a real OpenShift cluster:


Failed to pull image "ghcr.io/ibm/fast-time-server:1.0.3":
  reading manifest 1.0.3 in ghcr.io/ibm/fast-time-server: manifest unknown

    tag: "0.8.0"                # independent versioning — do NOT sync with the gateway release

charts/mcp-stack/templates/deployment-postgres.yaml
```diff
@@ -169,6 +169,12 @@ spec:
         - name: postgres
           securityContext:
             allowPrivilegeEscalation: false
+            readOnlyRootFilesystem: true
+            runAsNonRoot: true
+            runAsUser: 999
```
Yosiefeyob: Blocking — OpenShift's restricted-v2 SCC doesn't let a pod choose its own UID; it has to fall inside the namespace's allocated range. Hardcoding 999 means the pod is rejected at admission and never gets created:


.containers[0].runAsUser: Invalid value: 999:
    must be in the ranges: [1000730000, 1000739999]
runAsNonRoot: true on the line above already satisfies kube-linter, so dropping this line keeps the scanner happy and lets OpenShift assign a valid UID.

I installed this branch and main side by side on a cluster to check — postgres runs 1/1 on main and gets rejected here. Removing the line brings it back to 1/1.

charts/mcp-stack/templates/deployment-redis.yaml
```diff
@@ -42,6 +42,12 @@ spec:
         - name: redis
           securityContext:
             allowPrivilegeEscalation: false
+            readOnlyRootFilesystem: true
+            runAsNonRoot: true
+            runAsUser: 999
```
Yosiefeyob: Same as the postgres one — runAsUser: 999 gets the pod rejected by restricted-v2 before it's ever created. Dropping the line is enough; runAsNonRoot: true above still covers the lint check. Verified on a cluster: redis runs 1/1 on main, rejected on this branch, back to 1/1 once removed.

Yosiefeyob (changes requested):
Nice work on this — the hardening is the right shape, and the `/tmp` + securityContext pattern you used for postgres, redis and the migration job is exactly what's needed. Four blocking items and a couple of suggestions below.

**The short version:** the lint gate passes, but the chart doesn't deploy. I installed this branch on OpenShift alongside `main` to compare, and three components that work on `main` fail here:

| Component | `main` | this branch |
|---|---|---|
| postgres | 1/1 Running | rejected by SCC (`runAsUser: 999`) |
| redis | 1/1 Running | rejected by SCC (`runAsUser: 999`) |
| fast-time-server | 2/2 Running | `ImagePullBackOff` (tag `1.0.3` doesn't exist) |

The gateway also can't start: `readOnlyRootFilesystem: true` landed without a writable `/tmp`, so it dies on its lock file before gunicorn loads.

All four are small, and with them applied I got a fully healthy stack — `/health` and `/ready` both returning 200 — so this is close.

charts/mcp-stack/values.yaml
cafalchio: Thank you @Yosiefeyob I will address the issues :)

charts/mcp-stack/templates/deployment-postgres.yaml
cafalchio: Thank you Yosiefe. I will drop the UID and let the open shift assign the user id.

charts/mcp-stack/templates/deployment-redis.yaml
cafalchio: Thanks, I will remove this too.

charts/mcp-stack/values.yaml
cafalchio: I will need your help to test the deploy.

charts/mcp-stack/values.yaml
cafalchio: Thanks. Fixed

charts/mcp-stack/values.yaml
Yosiefeyob: Thanks matheus! Quick heads-up, I pinned the comment to the values.yaml line, but the actual fix belongs in the deployment template putting the fix into values.yaml lands it in the securityContext, which the cluster rejects 

1. In values.yaml if you could remove the block you added (lines 169–174):
`readOnlyRootFilesystem: true    # keep this line only`
    
2. In templates/deployment-mcpgateway.yaml (lines 320–335) — replace the plugin-only block so tmp is always present:
```
          volumeMounts:
            - name: tmp
              mountPath: /tmp
          {{- if .Values.mcpContextForge.pluginConfig.enabled }}
            - name: plugin-config-volume
              mountPath: /app/{{ $pluginsConfigFile }}
              subPath: config.yaml
          {{- end }}
      volumes:
        - name: tmp
          emptyDir: {}
      {{- if .Values.mcpContextForge.pluginConfig.enabled }}
        - name: plugin-config-volume
          configMap:
            name: {{ include "mcp-stack.fullname" . }}-gateway-plugins
            items:
            - key: config.yaml
              path: config.yaml
      {{- end }}
```
volumeMounts now sits at the container level and volumes at the pod level (not inside securityContext). Apologies again i can see why the earlier comment was misleading

claudia-gray:
Indentation error in deployment-mcpgateway.yaml lines 320 - 331. Addressed in https://github.com/IBM/mcp-context-forge/pull/5678 with permission from @cafalchio . Successfully verified this deploys properly with changes made in that PR 5678 mentioned.

Yosiefeyob (dismissed):
LGTM 🚀. Verified against documentation and succesfully ran on openshift.

brian-hussey (dismissed):
LGTM

Yosiefeyob (dismissed):
lgtm

brian-hussey (dismissed):

brian-hussey (approved):
LGTM @Yosiefeyob will you review again as well please?

Yosiefeyob:
LGTM ✅ Deployed to OCP and verified that the pods come up clean under the new security contexts.

PR #5202, opened 2026-06-12
fix: Bundle UI files on PyPI build

gcgoncalves:
# 🐛 Bug-fix PR

Closes: #5199

## 📌 Summary
Adds a build hook to the PyPI build to bundle the UI and include it in the package.

## 🔁 Reproduction Steps

1. Run the `uvx` package
```
BASIC_AUTH_PASSWORD=changeme \
JWT_SECRET_KEY=your-secret-key-at-least-32-characters \
MCPGATEWAY_UI_ENABLED=true \
MCPGATEWAY_ADMIN_API_ENABLED=true \
PLATFORM_ADMIN_EMAIL=admin@example.com \
PLATFORM_ADMIN_PASSWORD=SecurePass123 \
PLATFORM_ADMIN_FULL_NAME="Admin User" \
uvx --from mcp-contextforge-gateway mcpgateway --host 0.0.0.0 --port 4444
```
2. Access the Admin UI
3. Login
4. Observe the Admin UI not being interactible

## 🐞 Root Cause
The UI files were not packaged in the wheel file.

## 💡 Fix Description
Add a build hook to bundle the UI files and include them in the manifest file.

## ✅ Checklist
- [X] Code formatted (`make black isort pre-commit`)
- [X] No secrets/credentials committed

PR #5278, opened 2026-06-17
fix(docker): Containerfile.lite ships empty venv, masked by stray || true

### LINKED ISSUE #5276: [BUG]: Containerfile.lite ships empty venv (uv pip install fails silently, masked by || true) — not a FIPS/symlink issue
### 🐞 Bug Summary

Root cause turned out to be different (and more severe) than originally reported. Re-investigated by reproducing locally with the **exact** Containerfile and build-args CI uses.

**Corrected finding:** this is not a FIPS-specific `python3` symlink/PATH issue. It's `Containerfile.lite` (the file CI/release pipelines actually build — not the root `Containerfile`) silently shipping an **empty venv** for every build, FIPS and non-FIPS alike.

`uv pip install ".[redis,postgres,observability,granian,plugins,llmchat]"` fails during the `builder` stage with:

```
ModuleNotFoundError: No module named 'mcpgateway'
```

because `Containerfile.lite` copies only `pyproject.toml` into the build context and runs `uv pip install "."` for layer-caching reasons — `mcpgateway/` source isn't copied in until a later `COPY` step. PR #5202 ("fix: Bundle UI files on PyPI build", merged 2026-06-16) added:

```toml
[tool.setuptools.cmdclass]
build_py = "mcpgateway.tools.builder.build_hooks.BuildPyWithUI"
```

Resolving this `cmdclass` requires setuptools to `import mcpgateway` while parsing build config — which now fails at that point in the build, since the package isn't on disk yet.

That failure is then **silently swallowed**. The whole dependency-install step is one long `&&`-chained `RUN` command, ending in:

```bash
&& find /app/.venv -name "*.dist-info" -type d \( -name "pip-*" -o ... \) -exec rm -rf {} + 2>/dev/null || true \
```

Because of bash's flat `&&`/`||` precedence, this trailing `|| true` absorbs the carried-forward nonzero exit status from the earlier failed `uv pip install`, so the entire `RUN` instruction (and the `docker build`) reports success even though dependency installation completely failed. The resulting image's `/app/.venv` contains only `pip`/`setuptools`/`wheel`/`uv`/`packaging` — none of the actual application dependencies (no `gunicorn`, no `fastapi`, nothing).

The original symlink/PATH theory below (kept for history) is **incorrect**: both `/usr/bin/python3` and `PATH` resolve correctly to the venv interpreter in the FIPS build. The venv interpreter itself simply never had `gunicorn` installed.

---

### 🧩 Affected Component

---

### 🔁 Steps to Reproduce

```bash
docker build \
  --build-arg ENABLE_FIPS=true \
  --build-arg PYTHON_VERSION=3.11 \
  --build-arg UBI_BASE=registry.access.redhat.com/ubi9/ubi:latest \
  --build-arg NODEJS_IMAGE=registry.access.redhat.com/ubi9/nodejs-20:latest \
  --build-arg UBI_MINIMAL=registry.access.redhat.com/ubi9/ubi-minimal@sha256:6ea809bdd8164f8b2b607e38f01b14c374a15bdfbc74fcd0155161512dd4e00e \
  -f Containerfile.lite -t mcpgateway-fips-lite:test .

docker run --rm --user root --entrypoint /bin/bash \
  mcpgateway-fips-lite:test -c "$(cat scripts/fedramp-validate.sh)"
# => 21 passed, 1 failed (GUNICORN_OK check fails) — matches CI exactly

# Confirm it's not FIPS-specific:
docker build -f Containerfile.lite -t mcpgateway-nonfips-lite:test .
docker run --rm --entrypoint /app/.venv/bin/python3 mcpgateway-nonfips-lite:test \
  -c "import gunicorn"
# => ModuleNotFoundError: No module named 'gunicorn'

docker run --rm --entrypoint /app/.venv/bin/pip mcpgateway-nonfips-lite:test list
# Package    Version
# ---------- -------
# packaging  26.2
# pip        26.1.2
# setuptools 82.0.1
# uv         0.11.21
# wheel      0.47.0
```

The root `Containerfile` (not `.lite`) is unaffected, because it does `COPY . /app` (full source) before running `uv pip install`.

---

### 🤔 Expected Behavior

`Containerfile.lite` builds should fail the `docker build` outright if `uv pip install "."` fails, instead of silently producing an image with no application dependencies installed.

---

### 📓 Logs / Error Output

Original CI failure (merge-group run for PR #5268): https://github.com/IBM/mcp-context-forge/actions/runs/27679791847/job/81864084777

```
FAIL: runtime python3 can import gunicorn from the venv (regression: #68373)
      expected to contain: GUNICORN_OK
      got:

=== Results: 21 passed, 1 failed ===
```

Local repro of the underlying `uv pip install` failure (isolated to a 2-line Dockerfile reproducing the builder stage's COPY ordering):

```
error: Failed to build `mcp-contextforge-gateway @ file:///app`
Caused by: Build backend failed to determine extra requires with `build_sdist()` ...
ModuleNotFoundError: No module named 'mcpgateway'
```

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Version or commit | `main` since `5100c2b704` ("fix: Bundle UI files on PyPI build (#5202)", merged 2026-06-16) |
| Runtime | UBI9 base image, Python 3.11 venv at `/app/.venv` |
| Platform / OS | UBI9 (both FIPS and non-FIPS build paths) |
| Container | Docker, built from `Containerfile.lite` |

---

### 🧩 Additional Context — Severity / Blast Radius

- **Not FIPS-specific.** Reproduces identically on a plain (non-FIPS) `Containerfile.lite` build.
- `Containerfile.lite` is also what `.github/workflows/docker-multiplatform.yml` builds and pushes to GHCR for releases, with no workaround flags. Images built from this file since 2026-06-16 are likely missing their entire Python dependency tree and would fail to start (`ModuleNotFoundError: No module named 'fastapi'`/`'gunicorn'`) — this has not yet been confirmed against an actual published GHCR tag.
- This likely went unnoticed in the non-FIPS path because the regular build/scan CI step only builds and SBOM-scans the image (Syft/Grype) without ever running anything inside it. Only the FedRAMP-specific `fedramp-validate.sh` actually executes `python3 -c 'import gunicorn'` at runtime, which is why it surfaced there first.
- Suggested fix directions:
  1. Reorder `Containerfile.lite`'s builder stage so `mcpgateway/` source is copied (or at least a minimal package stub exists) before `uv pip install "."` runs — or pass `--no-build-isolation` / configure `[tool.uv.extra-build-dependencies]` so the build backend doesn't need to import the package being built.
  2. Fix the masked failure independently of (1): split the long `&&`-chained `RUN` block so cleanup steps with `|| true` don't share a chain with steps that must hard-fail the build, e.g. wrap `(find ... -exec rm -rf {} + 2>/dev/null || true)` in its own subshell/parentheses, or run cleanup in a separate `RUN`.
  3. Add a build-time (not just post-build) sanity check that the venv actually contains the expected packages, so a regression like this fails fast in every build, not just the FedRAMP gate.

msureshkumar88:
## Summary

- `Containerfile.lite`'s builder stage ran `uv pip install "."` before `mcpgateway/` source was copied into the build context. `pyproject.toml`'s `[tool.setuptools.cmdclass]` (added in #5202) needs to import `mcpgateway.tools.builder.build_hooks` to resolve build metadata, so the install failed with `ModuleNotFoundError: No module named 'mcpgateway'`.
- That failure was silently absorbed by a trailing `|| true` on a later, unrelated dist-info cleanup command sharing the same `&&` chain — bash's flat operator precedence let it catch the carried-forward nonzero status from the earlier failure, so `docker build` reported success with an empty venv (no `gunicorn`, `fastapi`, or any other project dependency installed).
- Fix: copy just the `build_hooks` module's import chain (`mcpgateway/__init__.py`, `mcpgateway/tools/builder/__init__.py`, `mcpgateway/tools/builder/build_hooks.py`) ahead of the venv install, so the cmdclass resolves without needing the rest of the source tree — keeps the dependency layer cache intact for normal source edits. Parenthesize the cleanup's `|| true` so it only catches its own exit status.

Closes #5276


msureshkumar88:
Cross-linking: #5279 is fixing the same root cause (the `[tool.setuptools.cmdclass]` import failure during `uv pip install` in `Containerfile.lite`'s builder stage), via a different approach — moving `build_hooks.py` to the repo root instead of staging its existing package-relative import chain.

Posted a side-by-side comparison on #5279 for whoever's reviewing. Summary: this PR also fixes the `\|\| true` masking bug that let the failure ship silently in the first place (not addressed in #5279), has a smaller diff (1 file vs. 5, no test-suite/packaging-doc rewrite), and I verified the added `COPY` lines don't bust the dependency-install cache layer on unrelated source changes (touched `mcpgateway/services/__init__.py`, rebuilt with warm cache, the `uv pip install` step stayed CACHED).

Leaving the actual merge call to a maintainer — just didn't want the duplicate effort to go unnoticed.

brian-hussey (dismissed):
Thanks @msureshkumar88. This looks good to me. 
I see discussion on #5279 and agree this is the correct fix for this issue.

msureshkumar88:
Checked the merge-queue run for this PR (run [27689637109](https://github.com/IBM/mcp-context-forge/actions/runs/27689637109/job/81897046617)) — good news and a follow-up:

**Good news:** "FedRAMP post-build compliance validation" now **passes (22/22)** in the real merge-queue environment, confirming the gunicorn/venv fix works end-to-end, not just locally.

**New failure:** "Scan SBOM for vulnerabilities (Grype)" failed — but it's unrelated to this PR's change. Before this fix, the venv was empty, so Syft found ~0 packages and Grype trivially passed (false green). Now that the venv installs the full dependency set for the first time, Grype is doing its job and surfaced a few pre-existing High-severity findings across both Python and OS-level (RPM) packages.

I've pushed follow-up commits on this branch:

1. General dependency hardening — bumping a couple of Python packages past their `exclude-newer` window (same per-package override pattern already used for `aiohttp`/`opentelemetry`). Verified locally: rebuilt the scan image, regenerated the SBOM with Syft, rescanned with Grype — those findings are gone.
2. The remaining High finding, in the base image's `python3`/`python3-libs` RPM packages, turned out to be a vendor advisory-feed gap rather than an actual unpatched issue — verified directly against the shipped binary that the patched code path is already present (Red Hat shipped the fix upstream before this RHEL10 build, but RHEL10's advisory feed doesn't carry a matching entry yet).

My first attempt at (2) used an OpenVEX document wired into the Grype step — that didn't actually work. Grype v0.91.1 (the version `anchore/scan-action` pins) only matches a VEX statement's product against the scanned image identifier, with the package listed as a subcomponent; it has no fallback for matching a package purl directly as the product. I verified this locally against the exact pinned `anchore/grype:v0.91.1` image — the VEX document was silently inert, the finding never actually got suppressed, and the merge queue failed again on the same gate.

Replaced it with a `.grype.yaml` ignore rule keyed on the vulnerability ID + package name, which doesn't depend on product/subcomponent matching and works correctly with v0.91.1. Re-verified locally with the same pinned Grype version: rescanning the SBOM with `.grype.yaml` in place drops the finding and the scan exits 0 with no High/Critical results.

madhu-mohan-jaishankar (dismissed):
LGTM

brian-hussey (dismissed):
LGTM

brian-hussey (dismissed):
Thanks for driving this forward @msureshkumar88, lgtm.

.github/vex/cve-2024-12254.openvex.json
```diff
@@ -0,0 +1,25 @@
...
+        {
+          "@id": "pkg:rpm/redhat/python3@3.12.13-2.el10_2?arch=x86_64&distro=rhel-10.2"
+        },
+        {
+          "@id": "pkg:rpm/redhat/python3-libs@3.12.13-2.el10_2?arch=x86_64&distro=rhel-10.2"
+        }
+      ],
+      "status": "fixed",
+      "status_notes": "Red Hat shipped the writelines() high-water-mark fix for this CVE in python3.12-3.12.8-1.el8_10 (RHSA-2024:10980) and python3.12-3.12.5-2.el9_5.2 (RHSA-2024:10978), both predating this image's python3.12.13-2.el10_2. Verified by inspecting the shipped binary: asyncio.selector_events._SelectorSocketTransport.writelines() calls self._maybe_pause_protocol() after a partial write, which is exactly the patched code path. Grype flags this as unfixed only because Red Hat's RHEL 10 CPE/advisory feed (RHEL 10 GA'd after the fix already landed upstream) does not yet carry a RHEL-10-specific advisory entry — a vendor feed-coverage gap, not an actual unpatched vulnerability."
+    }
```
madhu-mohan-jaishankar: One thing to tighten before merge: the commit message and status_notes say this is a "Red Hat advisory feed coverage gap" — but Red Hat's CVE page actually explicitly classifies RHEL 10 python3.12 as Affected (no errata, not just silence): https://access.redhat.com/security/cve/CVE-2024-12254

So the VEX is asserting fixed against the vendor's own stated position, not filling a missing entry. The status_notes should say something like:

"Red Hat's advisory feed lists RHEL 10 python3.12 as Affected (no errata as of 2026-06-17), however the upstream fix landed in CPython 3.12.9, predating this image's 3.12.13 build. Verified by inspecting the shipped binary: writelines() calls _maybe_pause_protocol() after a partial write (the patched code path). This VEX overrides the vendor feed on the basis of binary evidence."

madhu-mohan-jaishankar (changes requested):

.github/vex/cve-2024-12254.openvex.json
msureshkumar88: Good catch, thanks. Pulled the CVE data directly (https://access.redhat.com/hydra/rest/securitydata/cve/CVE-2024-12254.json) and confirmed: RHEL 10 python3.12 package_state is `"fix_state": "Affected"`, no errata. That's a stated position, not a coverage gap.

Separately, the VEX-based approach in this comment thread is gone entirely - it didn't actually work in real CI (grype v0.91.1 only matches VEX products against image identifiers, not package purls directly). Replaced with a `.grype.yaml` ignore rule, and reworded its comment to match your correction: it now says we're overriding Red Hat's explicit "Affected" status based on binary verification, not filling a missing advisory entry.

brian-hussey (approved):
LGTM

madhu-mohan-jaishankar (approved):
LGTM

PR #5279, opened 2026-06-17
fix(docker): make UI build hook importable during FIPS venv install

madhu-mohan-jaishankar:
## 🔗 Related Issue

Closes

---

## 📝 Summary

The FIPS container build was producing a virtual environment that contained
**only build tools** (`pip`, `setuptools`, `wheel`, `uv`) — no `mcpgateway`,
`gunicorn`, `uvicorn`, or any application dependency. This surfaced in the
Docker Security Scan as the FedRAMP post-build compliance failure:

```
FAIL: runtime python3 can import gunicorn from the venv (regression: #68373)
=== Results: 21 passed, 1 failed ===
```

**Root cause:** The custom setuptools `build_py` cmdclass `BuildPyWithUI` (added
to generate UI assets at packaging time) lived **inside** the `mcpgateway`
package at `mcpgateway/tools/builder/build_hooks.py`, and `pyproject.toml`
referenced it as `mcpgateway.tools.builder.build_hooks.BuildPyWithUI`.

`Containerfile.lite` intentionally copies **only `pyproject.toml`** into the
builder before running `uv pip install`, and copies the `mcpgateway/` source
tree **after** the heavy dependency layer (to preserve dependency-layer
caching — source edits shouldn't invalidate the install layer). As a result,
when setuptools tried to resolve the cmdclass during the install, it attempted
to `import mcpgateway`, which was not yet present in the build context, and
aborted:

```
ModuleNotFoundError: No module named 'mcpgateway'
```

`uv pip install` exited non-zero and installed nothing. The only FedRAMP check
that exercises a Python import (gunicorn) failed; the other 21 checks pass
because they only validate file permissions / crypto policy.

**Fix:** Move the build hook to a top-level `build_hooks.py` at the repo root so
it is importable during `uv pip install` **without** requiring the `mcpgateway`
source tree. `pyproject.toml` now references `build_hooks.BuildPyWithUI`, and the
Containerfile stages `build_hooks.py` alongside `pyproject.toml` before the venv
install. This fixes the install **while preserving** the existing
dependency-layer caching design.

---

## 🏷️ Type of Change

- [x] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

_List exact commands, screenshots, videos, logs, reproduction steps, or manual validation._

**Root-cause reproduction (before fix):** in a directory containing only
`pyproject.toml`:
```
$ uv pip install .
  ...
  ModuleNotFoundError: No module named 'mcpgateway'
  EXIT 1   # venv ends up with only pip/setuptools/wheel/uv
```

**After fix (directory with only `pyproject.toml` + `build_hooks.py`):**
```
$ uv pip install .
  ...
  EXIT 0
$ uv pip list | grep -iE 'gunicorn|uvicorn|fastapi'
  fastapi    0.137.x
  gunicorn   26.0.0
  uvicorn    0.47.0
```

**Clean container build + FedRAMP validation:**
```
$ docker build --no-cache -f Containerfile.lite \
    --build-arg ENABLE_FIPS=true \
    --build-arg PYTHON_VERSION=3.11 \
    --build-arg UBI_BASE=registry.access.redhat.com/ubi9/ubi:latest \
    --build-arg NODEJS_IMAGE=registry.access.redhat.com/ubi9/nodejs-20:latest \
    --build-arg UBI_MINIMAL=registry.access.redhat.com/ubi9/ubi-minimal@sha256:6ea809... \
    -t mcp-context-forge-scan:fedramp-validate .

$ docker run --rm --user root --entrypoint /bin/bash \
    mcp-context-forge-scan:fedramp-validate -c "$(cat scripts/fedramp-validate.sh)"
  ...
  PASS: runtime python3 can import gunicorn from the venv (regression: #68373)
  === Results: 22 passed, 0 failed ===
```

**Build-hook unit tests (import path / patch targets updated for the move):**
```
$ pytest tests/unit/mcpgateway/tools/builder/test_build_hooks.py
  28 passed   # 25 unit + 3 slow wheel-build tests
```

| Lint suite                | `make lint`     | ⬜ pending CI |
| Unit tests                | `make test`     | ✅ build_hooks suite 28/28; full suite via CI |
| Coverage ≥ 80%            | `make coverage` | ⬜ pending CI |
| FedRAMP validation        | `scripts/fedramp-validate.sh` (FIPS image) | ✅ 22 passed, 0 failed |

---

## ✅ Checklist

---

## 📓 Notes (optional)

### DIFF
--- a/tests/unit/mcpgateway/tools/builder/test_build_hooks.py
+++ b/tests/unit/mcpgateway/tools/builder/test_build_hooks.py
@@ -3,7 +3,7 @@
 Copyright 2026
 SPDX-License-Identifier: Apache-2.0
 
-Unit tests for mcpgateway/tools/builder/build_hooks.py.
+Unit tests for build_hooks.py.
 """
 
 # Standard
@@ -17,8 +17,8 @@ from unittest.mock import MagicMock, patch
 # Third-Party
 import pytest
 
-import mcpgateway.tools.builder.build_hooks as bh_module
-from mcpgateway.tools.builder.build_hooks import BuildPyWithUI
+import build_hooks as bh_module
+from build_hooks import BuildPyWithUI
 
 # ---------------------------------------------------------------------------
 # Helpers
@@ -65,23 +65,23 @@ def anchored(project, monkeypatch):
 class TestBuildUIAssetsFlag:
     def test_skips_when_flag_unset(self, hook, monkeypatch):
         monkeypatch.delenv("BUILD_UI_ASSETS", raising=False)
-        with patch("mcpgateway.tools.builder.build_hooks.build_py.run") as super_run:
+        with patch("build_hooks.build_py.run") as super_run:
             hook.run()
         super_run.assert_called_once()
 
     @pytest.mark.parametrize("value", ["1", "0", "false", "FALSE", "yes", ""])
     def test_skips_for_non_true_values(self, hook, monkeypatch, value):
         monkeypatch.setenv("BUILD_UI_ASSETS", value)
-        with patch("mcpgateway.tools.builder.build_hooks.build_py.run") as super_run:
+        with patch("build_hooks.build_py.run") as super_run:
             hook.run()
         super_run.assert_called_once()
 
     @pytest.mark.parametrize("value", ["true", "True", "TRUE"])
     def test_proceeds_for_true_values(self, hook, anchored, monkeypatch, value):
         monkeypatch.setenv("BUILD_UI_ASSETS", value)
-        with patch("mcpgateway.tools.builder.build_hooks.subprocess.run") as mock_run:
+        with patch("build_hooks.subprocess.run") as mock_run:
             mock_run.return_value = MagicMock(returncode=0)
-            with patch("mcpgateway.tools.builder.build_hooks.build_py.run"):
+            with patch("build_hooks.build_py.run"):
                 hook.run()
         assert mock_run.called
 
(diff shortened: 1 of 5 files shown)


msureshkumar88:
Nice catch on the root cause — matches what I found independently while reproducing #5276 locally. Heads up that #5278 fixes the same underlying issue with a different approach; flagging here so we don't land both.

Comparison, for whoever picks this up:

| | #5278 | #5279 (this PR) |
|---|---|---|
| Approach | Copy just `mcpgateway/__init__.py`, `mcpgateway/tools/builder/__init__.py`, `mcpgateway/tools/builder/build_hooks.py` ahead of `uv pip install`, keep `build_hooks.py` inside the package | Move `build_hooks.py` to repo root, update `pyproject.toml` cmdclass path, update 25+ test patch targets + packaging docs |
| Fixes the `uv pip install` failure | Yes (verified: FIPS build → `fedramp-validate.sh` 22/22; non-FIPS → `gunicorn`/`fastapi` import fine) | Yes (per your verification section) |
| Fixes the `\|\| true` masking bug | Yes — parenthesized so it only catches its own exit status, not the whole `&&` chain | Not addressed — if anything else in that RUN chain fails later, it'll still be silently swallowed and ship broken again |
| Cache impact on dep-install layer | Verified empirically: touching an unrelated `mcpgateway/` file (e.g. `services/__init__.py`) leaves the `uv pip install` layer CACHED — the 3 `COPY`s are keyed to those exact files only | Should be equivalent (copies `build_hooks.py` alongside `pyproject.toml`, same pattern), haven't tested yours directly |
| Diff size / surface | 1 file, +11/-3 | 5 files, relocates a packaged module, rewrites test suite, packaging docs |

Worth a maintainer call on which one lands — happy to close either side once decided. My read is #5278 is the smaller, more complete fix (also closes the masking bug), but wanted to lay out the actual comparison rather than just assert it.

tests/unit/mcpgateway/tools/builder/test_build_hooks.py
brian-hussey, line 1: This file should be directly in the tests/unit/ directory, not bound all the way down with mcpgateway/tools/builder as well as it no longer resides in that location.

brian-hussey (changes requested):
Thanks @madhu-mohan-jaishankar. Everything looks good but one change to make is the location of the unit tests for test_build_hooks.py

brian-hussey:
Sorry guys I started reviewing before I saw your comment @msureshkumar88. I think #5278 needs to go in and we can close #5279. That said, @madhu-mohan-jaishankar will you check if have a persistent problem with the pypi package build as well or if that is not an issue please?

brian-hussey:
Closing this one as #5278 is going to be merged to correct this issue.

PR #5283, opened 2026-06-17
fix(ui-rewrite): add functionality to edit tools in ToolDetailPanel

### LINKED ISSUE #5094: [UI-REWRITE]: Edit Tools

marekdano:
Closes #5094 

### Summary

Adds an **Edit** action to the tools UI so REST and MCP tools can be updated in place, alongside the existing delete flow.

Screenshot 2026-06-17 at 19 36 57

---

Screenshot 2026-06-17 at 19 38 28

### What's included

  - **Edit action in the tool row menu** — the overflow dropdown in `ToolsTable` now shows **Edit** above **Delete**. Either action can be enabled independently via the `onEditTool` / `onDeleteTool` props.
  - **Reusable form for create and edit** — `ToolForm` accepts an optional `tool` prop. When provided, it renders in edit mode: "Edit tool" heading, "Update tool" button, and fields pre-populated from the tool.
  - **Submit routing** — `useToolForm` sends `POST /tools` when creating and `PUT /tools/{id}` when editing, using the appropriate payload shape for each endpoint.
  - **Full tool fetch on edit** — the form opens immediately with list data, then fetches `GET /tools/{id}` to load decoded auth credentials (the list endpoint returns them masked/empty for security) and refreshes
  the auth fields once available.
  - **Request type handling** — the request-type selector is hidden for MCP tools and shown for REST tools. `requestType` / `integrationType` are only sent on update when the value is a valid REST method,
  preventing validation errors for MCP transports (`STREAMABLEHTTP`, `SSE`).

### Behaviour notes

  - After a successful update, the form closes and the tools list refetches so the updated data is reflected.
  - Auth credentials are never returned in plaintext by the list endpoint; on edit the single-tool fetch supplies them.
  - `Tool.tags` typing was corrected to match the API response shape (`{ id, label }` objects), with rendering updated in the table and details panel.
  - The schema dialog input/output blocks now have visible borders.

### Tests

  - `useToolForm` — edit-mode initialization, PUT vs POST routing, auth/custom-header initial values.
  - `ToolsTable` — Edit menu item rendering, ordering above Delete, callback wiring.
  - `ToolForm` — edit-mode heading/button, field pre-population, MCP vs REST request-type visibility, update submission.
  - `Tools` — end-to-end edit flow: open menu → Edit → form opens pre-populated → update → `PUT` fires → list refreshes with new data.

  **Note:** `client/src/hooks/useUserForm.ts` includes a small React hooks fix (capturing the `validationTimeouts` ref value in a local variable before the effect cleanup) to resolve a `react-hooks/exhaustive-deps`
  error surfaced by `npm run lint`.


client/src/components/tools/ToolForm.tsx
```diff
@@ -1,26 +1,76 @@
...
+import type { Visibility } from "@/types/server";
+
+const AUTH_TYPE_FROM_API: Partial<Record<string, AuthType>> = {
+  basic: "basic",
+  bearer: "bearer",
+  authheaders: "custom",
+};
+
+function toolToInitialValues(tool: Tool) {
+  const authType: AuthType = AUTH_TYPE_FROM_API[tool.auth?.authType ?? ""] ?? "none";
```
vishu-bh, line 19: there is small think that I noticed for authType.

selecting auth type None appears to remove credentials in the UI, but the update payload omits auth fields entirely. On the backend, omitted auth means “leave existing auth unchanged”, so the tool may continue sending the old stored credentials after the user thinks they removed them.

client/src/components/tools/ToolForm.tsx
```diff
@@ -1,26 +1,76 @@
...
+  const customHeaders =
+    tool.auth?.authHeaders && tool.auth.authHeaders.length > 0
+      ? tool.auth.authHeaders.map((h, i) => ({ id: String(i + 1), key: h.key, value: h.value }))
+      : tool.auth?.authHeaderKey
+        ? [{ id: "1", key: tool.auth.authHeaderKey, value: tool.auth.authHeaderValue ?? "" }]
+        : [];
+
+  return {
```
vishu-bh, line 27: Edit payload can overwrite real stored credentials with empty or masked values. The single-tool API response does not return editable plaintext secrets; backend ToolRead masks auth values, and this form then sends auth_password, auth_token, or auth_header_value back through PUT /tools/{id}. The backend re-encodes whatever is submitted, so a normal edit like changing the URL/description can silently replace the real upstream credential with the mask placeholder or an empty value.

vishu-bh (changes requested):
This one I think is a bug on old admin UI as well, but worth checking 

The form validates teamId, but the edit payload at `src/hooks/useToolForm.ts:598` does not send team_id. ToolUpdate also has no team_id field at `mcpgateway/schemas.py:1132`. 

Updating a public/private tool to team can leave it with visibility="team" and no team owner, likely making it inaccessible or inconsistently scoped.

marekdano:
@vishu-bh - Thanks for the careful review. Here's how I've addressed them:

  #### ✅ Fixed: masked credentials overwriting real ones (`ToolForm.tsx`)

  The edit form now only sends auth credential fields when the user actually entered a **real** (non-masked, non-empty) secret for the active auth type:

  - `basic` → checks `auth_password`
  - `bearer` → checks `auth_token`
  - `authheaders` → checks `auth_header_value` / `auth_headers[].value`

When the secret is still the masked `*****` placeholder (i.e. unchanged), the entire auth block is omitted so the backend leaves the stored credential untouched. A plain URL/description edit can no longer clobber the upstream credential. This mirrors the existing `useMCPServerForm` "skip masked values on submit" pattern.

  #### ✅ Fixed: selecting auth type "None" didn't clear credentials (`ToolForm.tsx:18`)

  When the user switches auth to **None** on edit, the payload now sends a nested auth object with empty strings:

  ```json
  "auth": { "authType": "", "authValue": "" }
  ```
This clears the stored credential because `update_tool` writes `auth_type/auth_value` whenever the incoming value `is not None` - empty strings pass, `null` would not. It's sent as the nested object (not flat fields), so it bypasses `assemble_auth`, which ignores empty/none auth types and would otherwise leave auth unchanged.

  #### ⚠️ Backend limitation: team_id not applied on edit

  This one can't be fully fixed from the frontend. I've raised a new issue for it https://github.com/IBM/mcp-context-forge/issues/5307

gcgoncalves (changes requested):
Some strings in this PR are missing i18n, please add it to the tools page.

gcgoncalves (approved):
🤜 🤛

PR #5309, opened 2026-06-18
feat(ui): Add resources list

### LINKED ISSUE #5097: [UI-REWRITE]: List Resources

gcgoncalves:
# ✨ Feature / Enhancement PR

## 🔗 Epic / Issue

Closes #5097 

---

## 🚀 Summary (1-2 sentences)

Screenshot 2026-06-22 at 14 14 16

Screenshot 2026-06-18 at 14 30 53


client/src/api/resources.ts
```diff
@@ -0,0 +1,50 @@
...
+  create: (data: ResourceCreateRequest): Promise<void> => {
+    return api.post("/admin/resources", data);
+  },
+
+  /**
+   * Update a resource
+   */
+  update: (id: string, data: ResourceUpdateRequest): Promise<void> => {
+    const validId = validateResourceId(id);
+    return api.post(`/admin/resources/${validId}/edit`, data);
```
marekdano:
```suggestion
    return api.post(`/resources/${validId}`, data);
```
We should have the `/resources/{resourceId}` API endpoint

client/src/api/resources.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  return id;
+}
+
+export const resourcesApi = {
+  /**
+   * Create a new resource
+   */
+  create: (data: ResourceCreateRequest): Promise<void> => {
+    return api.post("/admin/resources", data);
```
marekdano:
```suggestion
    return api.post("/resources", data);
```

client/src/api/resources.ts
```diff
@@ -0,0 +1,50 @@
...
+    const validId = validateResourceId(id);
+    return api.post(`/admin/resources/${validId}/edit`, data);
+  },
+
+  /**
+   * Delete a resource
+   */
+  delete: (id: string): Promise<void> => {
+    const validId = validateResourceId(id);
+    return api.delete(`/admin/resources/${validId}`);
```
marekdano:
```suggestion
    return api.delete(`/resources/${validId}`);
```

client/src/pages/Resources.test.tsx
```diff
@@ -0,0 +1,571 @@
...
+    id: `resource-${id}`,
+    name: `Resource ${id}`,
+    description: `Description for resource ${id}`,
+    gatewayId: `gateway-${gatewaySlug}`,
+    gatewaySlug,
+    enabled,
+    uri: `resource://example/${id}`,
+    mimeType: "application/json",
+    createdAt: new Date().toISOString(),
+    updatedAt: new Date().toISOString(),
```
marekdano, line 38: `tags: [],` is missing here

client/src/pages/Resources.tsx
```diff
@@ -1,5 +1,266 @@
...
+ * - Shows resource count and active status per group
+ * - Details panel for viewing individual resources
+ * - Placeholder form for future resource creation
+ */
 export function Resources() {
+  const intl = useIntl();
+  const [showForm, setShowForm] = useState(false);
+  const [selectedGroup, setSelectedGroup] = useState<ResourceGroup | null>(null);
+
+  const { data, isLoading, error, refetch } = useQuery<ResourcesResponse>("/resources?limit=0");
```
marekdano: Should we have `limit=0` here? I'd expect no limit param or the higher number 10 or more

client/src/pages/Resources.tsx
```diff
@@ -1,5 +1,266 @@
...
+            >
+              <h3 className="mb-1 font-semibold">
+                {intl.formatMessage({ id: "resources.errorLoading" })}
+              </h3>
+              <p className="text-red-800 dark:text-red-200">{error.message}</p>
+            </div>
+          )}
+
+          {!isLoading && (
+            <div className="grid grid-cols-1 gap-6 lg:grid-cols-2 xl:grid-cols-2 2xl:grid-cols-3">
```
marekdano: `xl:grid-cols-2` is redundant. `xl:grid-cols-2` is already the inherited value from `lg:grid-cols-2`. We can remove it.

client/src/pages/Resources.test.tsx
```diff
@@ -0,0 +1,571 @@
...
+      await waitFor(() => {
+        expect(screen.getByText(/multi-resource-gateway Resources/i)).toBeInTheDocument();
+      });
+
+      // All resources visible in panel
+      const panel = screen
+        .getByText(/multi-resource-gateway Resources/i)
+        .closest('[role="dialog"]');
+      expect(panel).toBeInTheDocument();
+      expect(within(panel!).getAllByText(/Resource \d/)).toHaveLength(3);
```
marekdano:
```suggestion
       expect(within(panel! as HTMLElement).getAllByText(/Resource \d/)).toHaveLength(3);
```

marekdano (commented):
@gcgoncalves - thanks for working on this!

I run the app locally and I can see no resource cards, even though the API returns them 

Image

Image

I left some comments below

client/src/pages/Resources.tsx
gcgoncalves: `limit=0` fetches all items. : )

client/src/pages/Resources.tsx
```diff
@@ -1,5 +1,266 @@
...
 export function Resources() {
+  const intl = useIntl();
+  const [showForm, setShowForm] = useState(false);
+  const [selectedGroup, setSelectedGroup] = useState<ResourceGroup | null>(null);
+
+  const { data, isLoading, error, refetch } = useQuery<ResourcesResponse>("/resources?limit=0");
+
+  // Group resources by gateway using utility function
+  const resourceGroups = useMemo<ResourceGroup[]>(
+    () => groupResourcesByGateway(data?.data ?? []),
```
vishu-bh: I think `/resources?limit=0` fetches  ResourcesResponse and reads data?.data, while the backend mcpgateway/main.py returns a bare list when include_pagination is false. 
so in my opinion this should always be [] 🤔

client/src/contexts/ResourcesContext.tsx
```diff
@@ -0,0 +1,42 @@
+import { createContext, useContext, type ReactNode } from "react";
```
marekdano: `ResourcesContext.tsx` (with `ResourcesProvider` and `useResourcesContext`) is created but never
  imported or used anywhere in the PR. 
`Resources.tsx` uses local `useState` directly.

Do we plan to keep `ResourcesContext.tsx` and use it later?

marekdano (commented):
Left a comment

client/src/pages/Resources.tsx
```diff
@@ -1,5 +1,259 @@
...
+            <span className="inline-flex items-center rounded bg-tool-badge-bg px-1.5 py-1 text-[10px] font-medium leading-none text-white">
+              {resource.mimeType}
+            </span>
+          )}
+          {resource.uri && (
+            <span
+              className="inline-flex items-center rounded bg-tool-badge-bg px-1.5 py-1 text-[10px] font-medium leading-none text-white"
+              title={resource.uri}
+            >
+              {new URL(resource.uri).hostname}
```
vishu-bh: The backend URI validator accepts path-style values like `/valid/uri`, but `new URL("/valid/uri")` throws unless a base URL is provided. A single valid resource with a path-style URI would crash the whole Resources page. It would be safer to use a small formatter with `try/catch` and fall back to the raw URI or scheme label.

client/src/hooks/useQuery.ts
```diff
@@ -3,6 +3,9 @@ import { api } from "@/api/client";
 
 type QueryMethod = "GET" | "POST" | "PUT" | "PATCH" | "DELETE";
 
+// In-flight request cache to prevent duplicate concurrent requests
+const inFlightRequests = new Map<string, Promise<unknown>>();
```
vishu-bh: The Resources loading test leaves `/resources?limit=0` unresolved, and because the promise stays in `inFlightRequests`, every later test reuses that unresolved promise and remains stuck on the loading spinner. I reproduced this with `npm run test:run -- src/pages/Resources.test.tsx` (`21 failed | 1 passed`). This also looks risky at runtime for hung/aborted shared requests. Please avoid caching unresolved requests across independent callers, or make abort/error cleanup robust and test-isolated.

client/src/types/resource.ts
```diff
@@ -0,0 +1,79 @@
...
+  modifiedUserAgent?: string;
+
+  // Import/federation metadata
+  importBatchId?: string;
+  federationSource?: string;
+  version: number;
+
+  // Ownership/visibility
+  teamId?: string;
+  team?: Record<string, unknown>;
```
vishu-bh: team` is typed as `Record<string, unknown>`, but the resource API/tests use a team name string. The E2E typecheck fails on this field.

client/src/types/resource.ts
```diff
@@ -0,0 +1,79 @@
...
+  ownerEmail?: string;
+  visibility: "public" | "private";
```
vishu-bh: The backend `ResourceRead` supports `"private" | "team" | "public"`, and the E2E fixture uses `"team"`.

vishu-bh (changes requested):
Added some inline comments

client/src/hooks/useQuery.ts
gcgoncalves: Thanks, this failure is bugging me since yesterday! I wasn't able to figure the root cause.

client/src/types/resource.ts
```diff
@@ -0,0 +1,81 @@
...
+  createdUserAgent?: string;
+  modifiedBy?: string;
+  modifiedFromIp?: string;
+  modifiedVia?: string;
+  modifiedUserAgent?: string;
+
+  // Import/federation metadata
+  importBatchId?: string;
+  federationSource?: string;
+  version: number;
```
vishu-bh: `Resource.version` is required in `client/src/types/resource.ts`, but the E2E fixtures in `client/e2e/resources.spec.ts` omit it. The backend `ResourceRead.version` is optional/defaulted, so either the frontend type should make `version` optional, or the fixtures should include it consistently.

vishu-bh (changes requested):
rest all looks good. Just last thing noticed

vishu-bh (approved):
Thanks for addressing all the comments!!

LGTM ☘️

PR #5310, opened 2026-06-18
fix(playwright): test_should_handle_object_parameter_validation by resolving HTMX race condition and async evaluate error

marekdano:
Relates https://github.ibm.com/contextforge-org/internal_issues/issues/297

---

## 📝 Summary

The test was consistently failing with `"Execution context was destroyed, most likely because of a navigation"` at `page.evaluate("Admin.testTool(...)")`.

---

### Steps to reproduce

1. Run `make serve`
2. Run `TEST_BASE_URL=http://localhost:4444 make test-ui-headless`
3. Failure 
```sh
FAILED tests/playwright/test_api_integration.py::TestAPIIntegration::test_should_handle_object_parameter_validation[chromium]
```
---

### Root causes

  **1. HTMX race between initial table load and search**

  `admin_page` only waits for `domcontentloaded`, so the tools table's `hx-trigger="load"` HTMX
  request may still be in-flight when the test fills the search input. Both the initial load and
  the search target `#tools-table` with `outerHTML` swap. If the search fires before the initial
  load completes, `#tools-table-body` doesn't exist yet, the search swap has no target, and the
  table stays empty.

  Fix: `wait_for_selector("#tools-table-body", state="attached")` before triggering the search.
  `state="attached"` (not `"visible"`) is required because an empty `<tbody>` has zero height and
  is considered hidden by Playwright.

  **2. Execution context destroyed by async evaluate**

  `Admin.testTool` is an `async` function that calls `await fetchWithTimeout(...)`. When invoked
  via `page.evaluate()`, Playwright blocks waiting for the returned Promise to resolve. Any
  navigation-like event during that await destroys the execution context. The previous workaround
  (`wait_for_load_state("networkidle")`) never resolved because the SSE connection keeps the page
  permanently non-idle.

  Fix: use `test_btn.click()` instead — this fires the async function without Playwright blocking
  on it. Use `get_attribute("data-test-tool-id")` to read the tool ID without a JS evaluate.

  **3. No synchronisation on search response**

  Without waiting for the HTMX search response to land, subsequent DOM operations could race with
  HTMX/Alpine's synchronous `history.replaceState` URL update.

  Fix: wrap `fill()`/`press()` in `page.expect_response(lambda r: "/admin/tools/partial" in r.url)`
  to guarantee the response and any associated URL state change has settled before proceeding.

### DIFF
--- a/tests/playwright/test_api_integration.py
+++ b/tests/playwright/test_api_integration.py
@@ -153,14 +153,32 @@ class TestAPIIntegration:
         Uses the json_param_tool fixture to guarantee a tool with an object
         parameter exists before the test runs.
         """
+        import json
+
         page.click('[data-testid="tools-tab"]')
         page.wait_for_selector("#tools-panel:not(.hidden)")
+        # The admin_page fixture only waits for domcontentloaded; the tools table's
+        # hx-trigger="load" initial HTMX request may still be in-flight.  Both the
+        # initial load and the search target #tools-table with outerHTML swap, so if
+        # the search fires before the initial load completes there is no #tools-table-body
+        # for HTMX to target — the swap goes nowhere and the table stays empty.
+        # Waiting for #tools-table-body to be attached (= initial load done) first
+        # eliminates this race entirely, for both fresh-page and pre-loaded cases.
+        page.wait_for_selector("#tools-table-body", state="attached", timeout=15000)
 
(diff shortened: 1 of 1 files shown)


ja8zyjits (approved):
LGTM, only test fixes.

PR #5312, opened 2026-06-19
fix: validate transport type on GatewayCreate and GatewayUpdate (#5289)

### LINKED ISSUE #5289: [BUG]: Gateway transport field accepts invalid values, causing refresh to silently succeed with no data
### 🐞 Bug Summary

The `transport` field on `GatewayUpdate` (schemas.py:3242) is typed as a bare `Optional[str]` with **no validation at all** — allowing any value to be stored.

`GatewayCreate` (schemas.py:2846) does have a `validate_transport` validator (lines 3059-3083), but it has two problems:
1. **Case-sensitive** — rejects `"sse"`, `"streamablehttp"` even though the runtime uses `.lower()` and would handle them fine
2. **No normalization** — passes the value through as-is, so the stored value may not match the canonical form

Additionally, `_initialize_gateway` (gateway_service.py:4707-4714) silently falls through both the `"sse"` and `"streamablehttp"` branches for any unrecognized transport value, returning empty data with `success: True` — no connection is attempted, no error is raised.

### Scope of the fix

| Gap | Exists? |
|---|---|
| `GatewayCreate` transport validator | ✅ Exists but case-sensitive |
| `GatewayUpdate` transport validator | ❌ Missing entirely |
| Runtime `else` clause for unrecognized transports | ❌ Missing — silent fallthrough |
| Schema/runtime case consistency | ❌ Schema strict, runtime case-insensitive |

### Expected Behavior

- `GatewayCreate` validator should accept any case variant (e.g. `"sse"`, `"SSE"`, `"Sse"`) and normalize to uppercase
- `GatewayUpdate` should have the same validation as `GatewayCreate`
- `_initialize_gateway` should raise `GatewayConnectionError` for unrecognized transports (to catch existing bad data in the DB)
- The `TransportType` enum docstring should clarify which values are production-ready vs. reserved

### Affected Component

- [x] `mcpgateway` - API
- [x] `mcpgateway` - DB models

### Steps to Reproduce

1. Create a gateway with `transport: "sse"` (lowercase) — fails at schema validation inconsistently
2. Update a gateway via `PATCH /gateways/{id}` with `transport: "invalid"` — succeeds silently
3. If a bad transport value reaches the DB, call `POST /gateways/{id}/tools/refresh` — succeeds with `success: true` and no data

### Proposed Fix

1. **Update `GatewayCreate.validate_transport`** (schemas.py:~3059) — make case-insensitive with normalization to uppercase, use `mode="before"`
2. **Add `validate_transport` to `GatewayUpdate`** (schemas.py:~3233) — same logic, handle `None` for optional field
3. **Add `else` clause in `_initialize_gateway`** (gateway_service.py:~4714) — raise `GatewayConnectionError` for unrecognized transports
4. **Update `TransportType` enum docstring** (schemas.py:~2804) — clarify production vs. reserved status
5. **Add tests** — schema tests for both Create and Update; runtime tests for `_initialize_gateway` with invalid transports

MohanLaksh:
## Description
Fixes a transport validation gap where `GatewayUpdate` accepted any string for the transport field without validation, and `GatewayCreate` transport validation was case-sensitive (rejecting `"sse"` while accepting `"SSE"`).

### Schema-level (schemas.py)
- **Updated** `GatewayCreate.validate_transport`: now normalizes to uppercase and is case-insensitive, aligning with runtime's `.lower()` normalization
- **Added** `GatewayUpdate.validate_transport`: same logic as GatewayCreate's validator, validating and normalizing transport on update

### Runtime (gateway_service.py)
- **Added** `else` clause in `_initialize_gateway`: raises `GatewayConnectionError` for unrecognized transports (HTTP, STDIO, or any invalid value), caught by the existing outer exception handler

## Tests
- **Schema tests** (`TestGatewayTransportValidation`): 10 tests covering valid transports with various case variants, invalid strings, empty strings, and normalization behavior for both `GatewayCreate` and `GatewayUpdate`
- **Runtime tests** (`TestInitializeGateway`): 4 tests covering invalid transport, empty transport, HTTP transport (not implemented), and STDIO transport (not implemented)

Closes #5289


Lang-Akshay (approved):
Thanks @MohanLaksh  for fixing three distinct gaps in transport field handling reported in #5289: `GatewayCreate`'s validator was case-sensitive (rejecting `"sse"` but accepting `"SSE"`); `GatewayUpdate` had no validator at all; and `_initialize_gateway` silently fell through for any unrecognized transport value, returning empty data with no error. The fix adds `mode="before"` case-insensitive normalization to both schema validators and adds an `else` clause in `_initialize_gateway` to raise `GatewayConnectionError`. Risk is low — the change is purely additive/guard logic with no behaviour change for well-formed inputs.

Lang-Akshay (commented):
Good work overall — the core fix is correct, validators are in the right place (`mode="before"`), normalization is clean, and test coverage is solid. Two things worth addressing before merge, plus a few suggestions.

**Blocking**

**1. Schema-service mismatch — `mcpgateway/schemas.py` (~line 3093 in `GatewayCreate.validate_transport`, mirrored in `GatewayUpdate`)**
`allowed = [t.value for t in TransportType.__members__.values()]` includes all four `TransportType` members including `HTTP` and `STDIO`. Both pass schema validation here but are rejected by `_initialize_gateway`'s new `else` clause at runtime. Callers get a misleading two-step failure: schema says valid (no 422), service raises a 500-class error. Worth considering restricting to a named `GATEWAY_SUPPORTED_TRANSPORTS = frozenset({"SSE", "STREAMABLEHTTP"})` constant and referencing that set in both validators — closes the mismatch and surfaces the 422 at the API boundary.

**2. Exception chain lost on re-wrap — `mcpgateway/services/gateway_service.py` (~line 4715)**
The new `GatewayConnectionError` raised in the `else` clause falls into the outer `except Exception as e` block in the same function and gets re-raised as `GatewayConnectionError(f"Failed to initialize gateway at {sanitized_url}: ...")`. Two side effects: the URL appears twice in the final error message (it's embedded in both the inner and outer error strings), and the original exception chain is dropped. Suggest either `raise GatewayConnectionError(...) from root_cause` in the outer re-raise, or checking `isinstance(e, GatewayConnectionError)` and re-raising without wrapping.

**Suggestions**

- **Duplicate validator body** (`mcpgateway/schemas.py`): `GatewayCreate.validate_transport` and `GatewayUpdate.validate_transport` are nearly identical — only the `if v is None: return v` guard differs. Worth extracting a `_validate_transport_string(v: str) -> str` module-level helper that both call.
- **Whitespace-padded input**: `v.strip()` silently accepts `" sse "` → `"SSE"`. Worth adding a test case to make this behaviour explicit and regression-proof.
- **Test assertion clarity**: In `test_invalid_transport_raises_error` and siblings, `match="Unsupported transport 'INVALID'"` actually matches the substring inside the *re-wrapped* message (`"Failed to initialize gateway at ...: Unsupported transport 'INVALID' for gateway at ..."`), not the direct raise. Works because `pytest.raises(match=)` uses `re.search`, but the intent is a bit opaque — a comment noting this helps future readers.

| Area | Rating | Key issue |
| ---- | ------ | --------- |
| Security | Needs attention | Schema accepts HTTP/STDIO, runtime rejects them (CWE-693) |
| Issue coverage | Done | All five #5289 criteria met |
| Code quality | Needs attention | Duplicate validator logic; exception chain lost on re-raise |
| Test quality | Done | Good coverage; one opaque assertion on re-wrapped message |

Lang-Akshay (changes requested):
please address above mentioned findings

Lang-Akshay (dismissed):
Thanks for the PR @MohanLaksh .
✅ Approved

msureshkumar88 (dismissed):
## E2E verification — PR fixes issue #5289

Ran a standalone E2E script against a real running instance of this branch (`uvicorn mcpgateway.main:app`, real HTTP requests via `requests`, throwaway SQLite DB) — not `TestClient`/mocks. All 7 checks map directly to issue #5289's reported gaps and reproduction steps.

**Config used:**
```
PORT=8766 HOST=127.0.0.1 DATABASE_URL=sqlite:///./e2e_pr5312_test.db
AUTH_REQUIRED=true PLATFORM_ADMIN_EMAIL=admin@example.com
MCPGATEWAY_UI_ENABLED=false MCPGATEWAY_ADMIN_API_ENABLED=true
LOG_LEVEL=ERROR RELOAD=false PLUGINS_ENABLED=false OBSERVABILITY_ENABLED=false
GATEWAY_ASYNC_LIFECYCLE_ENABLED=true   # TC1-TC4 (isolates transport validation from live MCP connectivity)
GATEWAY_ASYNC_LIFECYCLE_ENABLED=false  # TC5 (matches default sync refresh path)
```

**Commands run:**
```
.venv/bin/python -m uvicorn mcpgateway.main:app --host 127.0.0.1 --port 8766
python -m mcpgateway.utils.create_jwt_token --username admin@example.com --admin --exp 60 --secret <generated>
POST   /gateways                         {"transport": "sse"}            -> case-insensitive create
POST   /gateways                         {"transport": "StreamableHttp"} -> mixed-case normalization
POST   /gateways                         {"transport": "HTTP"}           -> unsupported-but-valid-enum rejection
PUT    /gateways/{id}                    {"transport": "invalid"}        -> update validator (was previously missing entirely)
PUT    /gateways/{id}                    {"transport": "streamablehttp"} -> update normalization
sqlite3 UPDATE gateways SET transport='BOGUS' WHERE id=...                -> simulate pre-existing corrupt legacy row
POST   /gateways/{id}/tools/refresh                                       -> runtime else-clause / GatewayConnectionError
```

**Results — 7/7 passed:**

| Check | Result |
|---|---|
| Unauthenticated create rejected (401/403) | ✅ 403 |
| `transport: "sse"` (lowercase) create accepted, normalized to `"SSE"` — repro step 1 | ✅ 202, `transport=="SSE"` |
| `transport: "StreamableHttp"` (mixed case) normalized to `"STREAMABLEHTTP"` | ✅ 202 |
| `transport: "HTTP"` (valid enum, unsupported by service) rejected at create | ✅ 422 |
| `PUT` with `transport: "invalid"` rejected — **the core reported bug**: `GatewayUpdate` previously had no validator at all and this succeeded silently | ✅ 422 |
| `PUT` with `transport: "streamablehttp"` (lowercase) normalizes to `"STREAMABLEHTTP"` | ✅ 202 |
| Refresh on a gateway with a corrupted legacy `transport='BOGUS'` DB row (simulating pre-existing bad data) surfaces a real failure instead of the old silent `success: true` with empty capabilities — repro step 3 | ✅ non-2xx / `unsupported transport` in response |

Full script: `tests/e2e/test_pr5312_gateway_transport_validation.py` (`.venv/bin/python tests/e2e/test_pr5312_gateway_transport_validation.py`, exit code 0).

Nice fix — closes the exact gap described in #5289 (case-sensitivity, missing `GatewayUpdate` validation, and the silent-fallthrough in `_initialize_gateway`) and the behavior held up end-to-end, not just at the unit level.

One non-blocking suggestion for a follow-up: the new `else` branch in `_initialize_gateway` interpolates the raw `transport` value into the error message that gets logged (`gateway_service.py` around the new `raise GatewayConnectionError(...)` line). Since this path is explicitly designed to catch pre-existing bad DB data (not just freshly-validated input), consider running it through the same sanitization already applied to URLs before logging, just as a defense-in-depth measure against unexpected characters in that string reaching the log.

msureshkumar88 (approved):
LGMT

PR #5314, opened 2026-06-19
fix(schemas): persist auth_headers array in POST /tools and PUT /tools/{tool_id}

### LINKED ISSUE #5201: [BUG][API]: POST /tools and PUT /tools/{tool_id} silently drop custom auth headers when auth_headers array is provided
### 🧩 Affected Component

- [x] `mcpgateway` - API
- [ ] Other (explain below)

---

 ### Describe the bug

Commit e27bdf180 (#3490) fixed multi-header `authheaders` support for the legacy admin UI routes
(`POST /admin/tools/` and `POST /admin/tools/{tool_id}/edit`) by introducing `_build_auth_obj_from_form()`, which reads the `auth_headers` list from the form, encodes all headers, and passes a pre-built `auth` object directly to `ToolCreate`/`ToolUpdate` — bypassing `assemble_auth()`.

The same fix was never applied to the JSON API routes used by the React admin UI:

- `POST /tools` — accepts a `ToolCreate` JSON body
- `PUT /tools/{tool_id}` — accepts a `ToolUpdate` JSON body

Both schemas' `assemble_auth()` model validator reads only the legacy `auth_header_key` / `auth_header_value` pair (single header). When a client sends `auth_type: "authheaders"` with `auth_headers: [{"key": "X-API-Key", "value": "..."}, ...]`, the array is silently ignored and the tool is stored with `auth_value: null`.

### Steps to reproduce

  1. `POST /tools` with body:

  ```json
  {
    "name": "my-tool",
    "url": "https://api.example.com/endpoint",
    "request_type": "POST",
    "auth_type": "authheaders",
    "auth_headers": [
      {"key": "X-API-Key", "value": "secret"},
      {"key": "X-Tenant", "value": "acme"}
    ]
  }
```
  2. Inspect the created tool - `auth_value` is null and no headers are stored.

### Expected behaviour

All entries in `auth_headers` are encoded and persisted, consistent with how `POST /admin/tools/` behaves after #3490.

### Root cause

`ToolCreate.assemble_auth()` and `ToolUpdate.assemble_auth()` in `mcpgateway/schemas.py` do not read the `auth_headers` list field — only `auth_header_key / auth_header_value`. The React UI sends the array format; the field is unrecognised and dropped.

### Proposed fix

Extend both `assemble_auth()` validators to handle the `auth_headers` array first (mirroring the logic already in `_build_auth_obj_from_form()`), falling back to the legacy single-header pair for backward compatibility.

### PR DESCRIPTION
**Pull request by shoummu1**
## :red_circle: Summary

`POST /tools` and `PUT /tools/{tool_id}` silently dropped custom auth headers when
the client sent `auth_type: "authheaders"` with an `auth_headers` array. The stored
tool always had `auth_value: null`, making authenticated tool calls fail at runtime.

## :link: Related Issue

Closes: #5201

1. `POST /tools` with body:
   ```json
   {
     "name": "my-tool",
     "url": "https://api.example.com/endpoint",
     "request_type": "POST",
     "auth_type": "authheaders",
     "auth_headers": [
       {"key": "X-API-Key", "value": "secret"},
       {"key": "X-Tenant", "value": "acme"}
     ]
   }
   ```
2. Inspect the created tool — `auth_value` is `null` and no headers are stored.

## :bug: Root Cause

`ToolCreate.assemble_auth()` and `ToolUpdate.assemble_auth()` in `mcpgateway/schemas.py`
only read the legacy `auth_header_key` / `auth_header_value` single-header pair when
`auth_type` is `"authheaders"`. The `auth_headers` array sent by the React admin UI was
present in the raw model-validator dict but never read, so it was silently discarded and
`auth_value` was stored as `null`.

The same multi-header path was already fixed for the legacy HTML admin UI routes
(`POST /admin/tools/`, `POST /admin/tools/{id}/edit`) via `_build_auth_obj_from_form()`
in `mcpgateway/admin.py` (commit e27bdf180, #3490), but that fix was never ported to
the JSON API schemas.

## :bulb: Fix Description

Extended both `ToolCreate.assemble_auth()` and `ToolUpdate.assemble_auth()` to mirror
the logic already in `_build_auth_obj_from_form()`:

1. **Check `auth_headers` list first** — build a `{key: value}` dict from the array
   and encode it via `encode_auth()`.
2. **Fall back to legacy** `auth_header_key` / `auth_header_value` when no array is
   present — preserving full backward compatibility with existing integrations.

Updated the docstring for `ToolCreate.assemble_auth()` to document both input paths and
corrected the stale log message in `ToolUpdate.assemble_auth()`.

## :test_tube: Verification

| Check                             | Command                | Status |
|-----------------------------------|------------------------|--------|
| Lint suite                        | `make lint`            | Passed |
| Unit tests                        | `make test`            | Passed |
| Coverage ≥ 90 %                   | `make coverage`        | Passed |
| Manual regression no longer fails | steps above            | Passed |

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-07-15T12:34:13Z",
+  "generated_at": "2026-07-15T15:18:26Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -382,7 +382,7 @@
         "hashed_secret": "d3ac7a4ef1a838b4134f2f6e7f3c0d249d74b674",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 5928,
+        "line_number": 5948,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -390,7 +390,7 @@
         "hashed_secret": "5932862bcd24dd27d0dc0407ec94fe9d6ea24aeb",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 6425,
+        "line_number": 6445,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -398,7 +398,7 @@
         "hashed_secret": "c77c805e32f173e4321ee9187de9c29cb3804513",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 6437,
+        "line_number": 6457,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -406,7 +406,7 @@
         "hashed_secret": "8fe3df8a68ddd0d4ab2214186cbb8e38ccd0e06a",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 6509,
+        "line_number": 6529,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -414,7 +414,7 @@
         "hashed_secret": "93ac8946882128457cd9e283b30ca851945e6690",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 7576,
+        "line_number": 7596,
         "type": "Secret Keyword",
         "verified_result": null
       }
@@ -1078,16 +1078,6 @@
         "is_verified": false,
         "line_number": 173,
         "type": "Secret Keyword",
-        "verified_result": null
-      }
-    ],
-    "docs/docs/architecture/mcp-apps.md": [
-      {
-        "hashed_secret": "7f89c5fe5f113485339383ac7cca52e698544052",
-        "is_secret": false,
-        "is_verified": false,
-        "line_number": 245,
-        "type": "Hex High Entropy String",
         "verified_result": null
       }
     ],
@@ -4058,7 +4048,7 @@
         "hashed_secret": "c377074d6473f35a91001981355da793dc808ffd",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 751,
+        "line_number": 750,
         "type": "Hex High Entropy String",
         "verified_result": null
       }
@@ -4089,16 +4079,6 @@
         "verified_result": null
       }
     ],
-    "mcpgateway/routers/oauth_router.py": [
-      {
-        "hashed_secret": "d3ecb0d890368d7659ee54010045b835dacb8efe",
-        "is_secret": false,
-        "is_verified": false,
-        "line_number": 739,
-        "type": "Secret Keyword",
-        "verified_result": null
-      }
-    ],
     "mcpgateway/schemas.py": [
       {
         "hashed_secret": "c377074d6473f35a91001981355da793dc808ffd",
@@ -4106,16 +4086,6 @@
         "is_verified": false,
         "line_number": 4577,
         "type": "Hex High Entropy String",
-        "verified_result": null
-      }
-    ],
-    "mcpgateway/services/mcp_client_chat_service.py": [
-      {
-        "hashed_secret": "a62f2225bf70bfaccbc7f1ef2a397836717377de",
-        "is_secret": false,
-        "is_verified": false,
-        "line_number": 590,
-        "type": "Secret Keyword",
         "verified_result": null
       }
     ],
@@ -4218,6 +4188,14 @@
       }
     ],
     "mcpgateway/utils/url_auth.py": [
+      {
+        "hashed_secret": "f2b14f68eb995facb3a1c35287b778d5bd785511",
+        "is_secret": false,
+        "is_verified": false,
+        "line_number": 118,
+        "type": "Secret Keyword",
+        "verified_result": null
+      },
       {
         "hashed_secret": "f2b14f68eb995facb3a1c35287b778d5bd785511",
         "is_secret": false,
@@ -4545,26 +4523,6 @@
         "verified_result": null
       }
     ],
-    "scripts/test_email_auth_api.py": [
-      {
-        "hashed_secret": "42d7bc4801cb72f2559bd9f6c9af9a2707ca321e",
-        "is_secret": false,
-        "is_verified": false,
-        "line_number": 341,
-        "type": "Secret Keyword",
-        "verified_result": null
-      }
-    ],
-    "tests/integration/test_concurrency_row_locking.py": [
-      {
-        "hashed_secret": "03aca59f05abb8e7b8ceb737167b6cd48dfa0a3f",
-        "is_secret": false,
-        "is_verified": false,
-        "line_number": 58,
-        "type": "Secret Keyword",
-        "verified_result": null
-      }
-    ],
     "tests/js/admin-ui.test.js": [
       {
         "hashed_secret": "1ded3053d0363079a4e681a3b700435d6d880290",
@@ -4869,58 +4827,22 @@
         "verified_result": null
       }
     ],
-    "tests/playwright/security/test_team_selector_e2e.py": [
-      {
-        "hashed_secret": "b4e44716dbbf57be3dae2f819d96795a85d06652",
-        "is_secret": false,
-        "is_verified": false,
-        "line_number": 283,
-        "type": "Secret Keyword",
-        "verified_result": null
-      }
-    ],
     "tests/unit/mcpgateway/services/test_mcp_client_chat_service_extended.py": [
       {
         "hashed_secret": "d576acab7ea77edc8aa4f9ebea6bd4c1cc0d1764",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1412,
+        "line_number": 1397,
         "type": "Secret Keyword",
         "verified_result": null
       }
     ],
     "tests/unit/mcpgateway/services/test_oauth_audience.py": [
       {
-        "hashed_secret": "e360adbe2fa9bb945b9afc993ea1306cb976e947",
-        "is_secret": false,
-        "is_verified": false,
-        "line_number": 213,
-        "type": "Secret Keyword",
-        "verified_result": null
-      },
-      {
-        "hashed_secret": "50e8361b5726f72eed9dffaec824570b635147b6",
-        "is_secret": false,
-        "is_verified": false,
-        "line_number": 246,
-        "type": "Secret Keyword",
-        "verified_result": null
-      },
-      {
         "hashed_secret": "fe1bae27cb7c1fb823f496f286e78f1d2ae87734",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 319,
-        "type": "Secret Keyword",
-        "verified_result": null
-      }
-    ],
-    "tests/unit/mcpgateway/services/test_oauth_manager.py": [
-      {
-        "hashed_secret": "fe1bae27cb7c1fb823f496f286e78f1d2ae87734",
-        "is_secret": false,
-        "is_verified": false,
-        "line_number": 679,
+        "line_number": 160,
         "type": "Secret Keyword",
         "verified_result": null
       }
@@ -4930,25 +4852,17 @@
         "hashed_secret": "718cbcc5a4207c0d5f38e3a309bdba17cb0074b7",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 970,
+        "line_number": 461,
         "type": "Hex High Entropy String",
         "verified_result": null
       }
     ],
     "tests/unit/mcpgateway/test_admin.py": [
       {
-        "hashed_secret": "995e87a207ec41ef95f938dcf53a4184312b0d93",
-        "is_secret": false,
-        "is_verified": false,
-        "line_number": 11800,
-        "type": "Hex High Entropy String",
-        "verified_result": null
-      },
-      {
         "hashed_secret": "a0281cd072cea8e80e7866b05dc124815760b6c9",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 18217,
+        "line_number": 17941,
         "type": "Secret Keyword",
         "verified_result": null
       }
@@ -4958,7 +4872,7 @@
         "hashed_secret": "5baa61e4c9b93f3f0682250b6cf8331b7ee68fd8",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 235,
+        "line_number": 129,
         "type": "Basic Auth Credentials",
         "verified_result": null
       }
@@ -4968,7 +4882,7 @@
         "hashed_secret": "4f13f134744a2fadbbe2d624687246347d12fa63",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 2830,
+        "line_number": 2805,
         "type": "Hex High Entropy String",
         "verified_result": null
       }
@@ -4978,7 +4892,7 @@
         "hashed_secret": "72cb70dbbafe97e5ea13ad88acd65d08389439b0",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1484,
+        "line_number": 1464,
         "type": "Secret Keyword",
         "verified_result": null
       }
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-07-14T16:15:49Z",
+  "generated_at": "2026-07-14T16:29:16Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -4104,7 +4104,7 @@
         "hashed_secret": "c377074d6473f35a91001981355da793dc808ffd",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 4561,
+        "line_number": 4577,
         "type": "Hex High Entropy String",
         "verified_result": null
       }
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -266,7 +266,7 @@
         "hashed_secret": "b4673e578b9b30fe8bba1b555b7b59883444c697",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1736,
+        "line_number": 1746,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -274,7 +274,7 @@
         "hashed_secret": "4a0a2df96d4c9a13a282268cab33ac4b8cbb2c72",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1824,
+        "line_number": 1834,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -282,7 +282,7 @@
         "hashed_secret": "5baa61e4c9b93f3f0682250b6cf8331b7ee68fd8",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 2174,
+        "line_number": 2184,
         "type": "Basic Auth Credentials",
         "verified_result": null
       },
@@ -290,7 +290,7 @@
         "hashed_secret": "fa9beb99e4029ad5a6615399e7bbae21356086b3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 3539,
+        "line_number": 3549,
         "type": "Basic Auth Credentials",
         "verified_result": null
       },
@@ -298,7 +298,7 @@
         "hashed_secret": "fa9beb99e4029ad5a6615399e7bbae21356086b3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 3630,
+        "line_number": 3640,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -306,7 +306,7 @@
         "hashed_secret": "ac371b6dcce28a86c90d12bc57d946a800eebf17",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 3673,
+        "line_number": 3683,
         "type": "Secret Keyword",
         "verified_result": null
       },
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

**Review by marekdano — COMMENTED** [review]
The PR looks good! 

Just a few suggestions before merging

1. **Defensive `.get()` on list items without `isinstance` guard** — the dict comprehension:
```python
   header_dict = {h.get("key"): h.get("value", "") for h in auth_headers if h.get("key")}
```
If any element of `auth_headers` is not a dict (e.g. a bare string), `.get()` raises `AttributeError`. A guard like `if isinstance(h, dict) and h.get("key")` costs nothing and avoids a confusing 500 on malformed input. The existing `A2AAgentCreate` validators use this pattern — worth being consistent.

2. **Identical new logic duplicated across two classes** - `ToolCreate.assemble_auth()` and `ToolUpdate.assemble_auth()` are now a 1:1 copy of each other for the `authheaders` branch. A module-level helper:
```python
   def _encode_auth_headers_list(auth_headers: list) -> Optional[str]: ...
 ```
would remove the duplication and make future fixes a one-liner. This is a pre-existing pattern problem, but the PR makes it worse.

**Minor notes:**

3. **`auth_headers` is not declared as a field on `ToolCreate`/`ToolUpdate`** — this means it won't appear in the generated OpenAPI schema for `POST /tools` or `PUT /tools/{tool_id}`.
Clients relying on the schema to discover this parameter won't see it. Adding `auth_headers: Optional[List[Dict[str, str]]] = Field(None, ...)` as a declared (but validator-consumed) field would fix that. `GatewayCreate` and `GatewayUpdate` already declare this field — making `ToolCreate`/`ToolUpdate` consistent would improve discoverability.

4. **Missing `test_tool_update_authheaders_empty_array`** — `ToolCreate` with `auth_headers=[]` is tested; `ToolUpdate` with the same is not. Low impact but easy to add for symmetry.

5. **Doctest only asserts `auth_type`** — the new multi-header doctest asserts `result['auth']['auth_type']` but not `result['auth']['auth_value']`. Worth asserting non-`None` `auth_value` to make the doctest self-documenting.

**Review by marekdano — CHANGES_REQUESTED** [review]
Few more findings

#### 1. Validation gap vs. `GatewayCreate` _(blocking)_

`GatewayCreate.assemble_auth()` (`schemas.py:3129–3180`) already handles `authheaders` with:

  ~~~python
  if not all(c.isalnum() or c in "-_" for c in key.replace(" ", "")):
      raise ValueError(f"Invalid header key format: '{key}'...")
  if not header_dict:
      raise ValueError("For 'authheaders' auth, at least one valid header...")
  if len(header_dict) > 100:
      raise ValueError("Maximum of 100 headers allowed per gateway.")
  ~~~

The new `_encode_auth_headers_list` silently returns `None` in these cases instead of raising.
This creates an inconsistent API: gateways reject invalid input with 422, tools silently store `auth_value: null`.

#### 2. Three copies of the same logic

  `GatewayCreate` and `GatewayUpdate` (`schemas.py:3129–3180`, `3486–3537`) already have nearly identical inline logic. The PR adds a third copy rather than making `_encode_auth_headers_list` the canonical shared helper for all four schemas.

#### 3. Docstring/test mismatch

  ~~~python
  >>> _encode_auth_headers_list(['not-a-dict']) is None
  True
  ~~~

This only holds when calling the helper directly. Through `ToolCreate`, Pydantic rejects
`["not-a-dict"]` at field validation before the validator runs. The docstring should clarify.

#### 4. Silent empty-list fallthrough

`if auth_headers and isinstance(auth_headers, list)` treats `[]` as falsy and falls through to the legacy path. The outcome (`auth_value: null`) is correct but the intent is non-obvious.

#### 5. Missing `ToolUpdate` precedence test

`test_tool_create_authheaders_array_takes_precedence_over_legacy` has no `ToolUpdate` equivalent.

**Review by marekdano — DISMISSED** [review]
The PR fixes a real bug (#5201): `ToolCreate/ToolUpdate` only read the legacy single-header `auth_header_key/auth_header_value` pair, so the `auth_headers` array sent by the React admin UI was silently dropped and `auth_value` stored as `null`.
The fix extends both tool schemas to read the array, and - nicely - extracts the `previously-duplicated multi-header` validation/encoding logic into a single shared `helper _encode_auth_headers_list()` that the gateway and A2A validators now also use.

LGTM 🚀

**Review by msureshkumar88 — CHANGES_REQUESTED** [review]
Thanks for tracking down the root cause here — the `mode="before"` vs `mode="after"` distinction between Tool and Gateway/A2A schemas is subtle, and porting the `_encode_auth_headers_list` extraction across all six create/update classes is a nice cleanup on top of the actual fix. The new unit tests around precedence, legacy fallback, and empty-array handling are solid.

I did find one regression worth fixing before merge, plus a couple of things worth a look.

## Blocking

**Non-string `auth_headers` keys crash with a raw 500 instead of a clean 422** (`mcpgateway/schemas.py`, `_encode_auth_headers_list`, called from `ToolCreate`/`ToolUpdate.assemble_auth`).

Because `ToolCreate.assemble_auth`/`ToolUpdate.assemble_auth` are `mode="before"` validators, they run on the raw, uncoerced client JSON — unlike `GatewayCreate`/`A2AAgentCreate`, which are `mode="after"` and only see `auth_headers` once Pydantic has already coerced it to `List[Dict[str, str]]`. That coercion is what currently protects the Gateway/A2A call sites; Tool doesn't get it.

Confirmed locally:
```python
ToolCreate(name="t", url="https://x.com", request_type="POST",
           auth_type="authheaders", auth_headers=[{"key": 123, "value": "x"}])
# AttributeError: 'int' object has no attribute 'replace'

ToolCreate(name="t", url="https://x.com", request_type="POST",
           auth_type="authheaders", auth_headers=[{"key": ["X"], "value": "y"}])
# TypeError: unhashable type: 'list'
```
Neither is a `ValueError`, so Pydantic doesn't wrap it into a `ValidationError`, and it isn't caught by the `RequestValidationError` handler — it falls through to the catch-all `Exception` handler in `main.py` and returns a 500 (logged as "Unhandled exception"). No stack-trace leak, but it's the wrong status code for ordinary bad input on a public write endpoint, and it'll show up as noise in error-rate alerting. A quick fix would be catching `(AttributeError, TypeError)` in the loop and re-raising as `ValueError`, or normalizing/validating key types before the `.replace()` call.

Would also be good to add a regression test for this specific case (non-dict-but-JSON-valid key type), since none of the new tests cover it.

## Suggestions

- **Header-key whitespace bypass**: the format check validates `key.replace(" ", "")` but stores the original `key`, so `{"key": "X Api Key", ...}` passes validation and gets persisted with embedded spaces — an invalid HTTP header name that will only fail later, at actual tool-invocation time. This is pre-existing behavior carried over from the Gateway/A2A logic, but it's now reachable from Tool creation too, so might be worth tightening while this code is being touched.
- **Admin UI (legacy HTML form) path stays unvalidated**: `_build_auth_obj_from_form()` in `admin.py` still hand-builds the header dict and doesn't call the new shared `_encode_auth_headers_list`, so `/admin/tools/` accepts malformed header keys or >100 headers that `POST /tools` now correctly rejects. Given the PR description frames this as bringing the JSON API in line with the admin form's existing validation, closing the loop the other way (admin form adopting the new stricter helper) would give full parity — could be a good fast-follow if out of scope here.
- **Test coverage**: all the new tests construct `ToolCreate`/`ToolUpdate` directly rather than going through the actual `POST /tools`/`PUT /tools/{id}` route + service + DB layer. An integration test that posts the reproduction payload from the PR description and asserts the persisted `auth_value` would give stronger end-to-end confidence that the original bug is actually closed.

## Minor

- `docs/docs/using/multi-auth-headers.md` only documents this feature for Gateways — no mention of Tools. Worth a follow-up doc update once the API shape is settled.
- The Tool legacy-fallback block (`auth_header_key`/`auth_header_value`) is still duplicated between `ToolCreate` and `ToolUpdate` rather than factored out like the array-handling path was — small maintenance cost, not blocking.

Everything else (the dedup itself, the precedence logic, the docstring updates) looks correct and well-tested. Happy to take another pass once the crash is addressed.

marekdano:
## Review feedback addressed

Thanks for the thorough passes — both reviews landed on real issues. Here's what changed.

### Blocking: 500 instead of 422 on bad `auth_headers` types — fixed

Reproduced first. `ToolCreate`/`ToolUpdate.assemble_auth` are `mode="before"` validators, so `_encode_auth_headers_list` receives raw client JSON:

  - `{"key": 123}` → `AttributeError` on `.replace()`
  - `{"key": ["X"]}` → `TypeError` on dict insertion

Both fell through to the catch-all handler as 500s. The helper now rejects non-string keys and values with a `ValueError`, which Pydantic wraps into a proper 422.

### Whitespace bypass — fixed

Confirmed real: `{"key": "X Api Key"}` passed validation (which checked `key.replace(" ", "")`) but stored the original key, persisting an invalid HTTP header name that would only fail at tool-invocation time. Keys are now stripped of surrounding whitespace and rejected if any remains embedded.

### Admin form parity — fixed

`_build_auth_obj_from_form()` now routes through the shared `_encode_auth_headers_list`, so
 `/admin/tools/` rejects malformed keys and >100 headers exactly like `POST /tools`.

One deliberate carve-out: rows with a blank key are filtered out *before* validation, because the form submits empty rows for headers the user never filled in — those must keep meaning "no headers" rather than becoming a 422. As a side effect this also removes a latent crash where a non-list JSON body (e.g. `{"a": 1}`) would call `.get()` on a string.

### End-to-end test — added

`tests/integration/test_tool_auth_headers_api.py` posts the exact reproduction payload from the PR description through the real route → `ToolService` → SQLite, then reads the row back and asserts `auth_value` decodes to both headers. The original bug is now closed against the persistence layer, not just the schema. It also covers the 422 paths.

### Remaining duplication — fixed

A new `_assemble_tool_authheaders()` holds the array / legacy / empty precedence logic that both `ToolCreate` and `ToolUpdate` call, which also makes the `auth_headers=[]` fallthrough explicit instead of implicit. `docs/docs/using/multi-auth-headers.md` now covers tools alongside gateways and documents the actual validation rules.

---

### Two things worth flagging

**The stricter key validation is a behavior change beyond tools.** Gateways and A2A agents share the same validator, so a gateway with an embedded-space header key that was previously accepted will be rejected on its next update. No existing test depended on that leniency and the stored value was never a usable HTTP header — but it's a deliberate tightening, not a pure bug fix, so worth a conscious ack before merge.

**Two tests reported as missing already exist.** `test_tool_update_authheaders_array_takes_precedence_over_legacy` and the `ToolUpdate` empty-array case are on the branch at lines 650 and 714 — they were added in response to the first review.

### detect-secrets

Baseline regenerated; `pre-commit run detect-secrets --all-files` passes. The one new finding came from a doctest *output* line (`{'X-API-Key': 'legacy'}`), where an inline `# pragma: allowlist secret` would corrupt the assertion — and the hook strips a manually added `is_secret: false` on its next run. Reshaped the doctest to assert on the value alone instead. The baseline diff is now nothing but line-number shifts.

msureshkumar88 (changes requested):
Thanks for this — it's a clean, well-scoped fix for a real runtime failure, and the shared `_encode_auth_headers_list()` extraction is the right move (removes four duplicated validation blocks). Test coverage is genuinely strong: schema-level, admin-form, and a route-level integration test that asserts what lands in the DB and closes the #5201 repro. Docstrings/doctests are excellent.

Requesting changes on two items — one hardening, one release-note — plus a few optional notes.

## Blocking

### 1. Guard `auth_headers` length before iterating (CWE-400 / CWE-770)
`_encode_auth_headers_list()` loops the entire input list and only checks `len(header_dict) > 100` *after* the loop. The tool field is `Optional[List[Dict[str, str]]]` with no `max_length`, and the tool validators run `mode="before"` on raw client JSON — so a caller can submit an arbitrarily large array and force that many iterations before rejection. It's an authenticated, RBAC-gated endpoint so severity is low, but the fix is cheap:

```python
if len(auth_headers) > 100:
    raise ValueError("Maximum of 100 headers allowed.")
```

early in the helper, and/or `max_length=100` on the Pydantic field. Worth closing before merge.

### 2. Call out the gateway/A2A validation tightening in CHANGELOG
The shared helper changes gateway/A2A semantics: the old validator used `all(... for c in key.replace(" ", ""))`, so header keys with embedded spaces passed validation and were stored (broken, failing later at invocation). The new helper rejects them with a 422 at config time and trims surrounding whitespace before storage. This is a genuine improvement, but it's a behavior change beyond the "POST /tools" framing in the title, and a previously-accepted (if already-broken) gateway/A2A config would now 422. A one-line CHANGELOG note would set expectations. Real-world blast radius is near-nil since those keys never worked at runtime.

## Suggestions (non-blocking)

- A couple of integration-level gaps worth a test if you're extending the file: `PUT /tools/{id}` with `auth_headers: []` (the documented "clear all headers → null" unset path is only covered at schema level), the >100 rejection through the real route, and the legacy single-header pair through `POST /tools`.
- Docs: `multi-auth-headers.md` could note that the 100 limit counts *unique* keys (duplicates collapse) and that non-string key/value types are rejected with a 422.

## Minor notes
- Nice that non-string key/value now surface as 422 instead of an unhandled 500 — good catch on the `mode="before"` raw-JSON path.
- Verified the debug log in `assemble_auth` logs only `auth_type`/`auth_username`/`auth_header_key`, never header values or passwords — no secret leakage introduced.
- No dead code or unrelated changes in the diff; the `auth.js` message wording aligns with the now-shared limit.

msureshkumar88 (changes requested):
Thanks for closing out the previous rounds so thoroughly — the shared `_encode_auth_headers_list` extraction, the CWE-770 length guard, the whitespace-bypass fix, and the new integration test all check out. Re-reviewed the current diff end-to-end and traced through the actual runtime behavior (not just the diff) to look for anything the earlier rounds might have missed.

## Blocking

**Admin form path still crashes with a 500 on a non-list `auth_headers` value** (`mcpgateway/admin.py`, `_build_auth_obj_from_form`, around the `populated_headers = [...]` list comprehension).

`auth_headers` is parsed via `orjson.loads()` on a form field, so it can legally come back as any JSON type — not just a list. Reproduced directly:

```python
auth_headers = orjson.loads('5')     # valid JSON -> int
for h in auth_headers: ...            # TypeError: 'int' object is not iterable
```

Posting `auth_type=authheaders&auth_headers=5` (or `null`, `true`, `3.14`) to `/admin/tools/` raises an uncaught `TypeError` that isn't handled anywhere in the function, so it falls through to the catch-all exception handler as a 500 instead of a 422.

This is the same CWE-20 class this round already fixed for the dict-shaped bad-input case (and the issue comment specifically calls out fixing "a non-list JSON body... would call `.get()` on a string") — the scalar/null case just wasn't covered by that fix. I confirmed the JSON API path (`ToolCreate`/`ToolUpdate`) is *not* affected: `_assemble_tool_authheaders` checks `isinstance(auth_headers, list)` before touching it, so a scalar there gets a clean Pydantic 422. Only the legacy admin-form path is missing the equivalent guard.

Suggested fix, one line:
```python
populated_headers = [h for h in auth_headers if isinstance(h, dict) and h.get("key")] if isinstance(auth_headers, list) else []
```

Would also be good to add a regression test for this (e.g. `auth_headers` form field set to `"5"` or `"null"`), since none of the new tests cover a non-list JSON value for this field.

## Non-blocking notes (optional, not asking to block on these)

- **Residual CWE-770 exposure on Gateway/A2A schemas**: since those validators run `mode="after"`, Pydantic fully coerces the raw `auth_headers` list to `List[Dict[str, str]]` *before* the shared helper's 100-entry cap ever runs — so an oversized payload is still fully walked by pydantic-core first. Benchmarked a 200k-entry payload at ~0.2s, so real-world impact is negligible, but adding `max_length=100` to the already-declared field would close this for free.
- **Pre-existing, now more broadly reachable**: the header-key format check uses `str.isalnum()`, which accepts non-ASCII Unicode letters (verified `X-Ünïcode-Ω` passes and gets persisted) — not a valid RFC 7230 header token. Unchanged logic from the existing Gateway/A2A code, but Tool creation now reaches it too.
- **Minor**: `_encode_auth_headers_list`'s `if value is None: value = ""` branch (and its doctest/unit test) can't actually be exercised through `ToolCreate`/`ToolUpdate`/`GatewayCreate` — Pydantic's own field coercion rejects a `null` value before that branch would matter. Only reachable via the admin-form path, which bypasses Pydantic model validation. Worth a one-line docstring clarification so it doesn't read as universally applicable.

Everything else — scope, docs, CHANGELOG disclosure, issue linkage, test coverage — looks solid. Happy to take another pass once the admin-form path is guarded.

msureshkumar88:
**Status: Not approving yet** — re-reviewed the current diff (`a43f804eb`) end-to-end and ran the actual test suite rather than relying on the PR checklist.

## Requested-changes tracking

All five prior review rounds' blocking items are resolved and verified:
- isinstance guards, dedup into `_encode_auth_headers_list` / `_assemble_tool_authheaders`, undeclared field, doctest/docstring gaps (round 1-2)
- Non-string keys crashing with a raw 500 instead of 422 (round 3)
- CWE-770 unbounded iteration before the 100-cap + CHANGELOG disclosure of the Gateway/A2A whitespace-validation tightening (round 4)
- Admin form 500 on a non-list `auth_headers` JSON scalar (round 5, commit `ab2c2bfb7`)

Confirmed by running:
```
tests/unit/mcpgateway/test_schemas_auth_validation.py, test_admin.py -k auth_headers/authheaders → 48 passed
tests/integration/test_tool_auth_headers_api.py --with-integration → 7 passed
```

## Why not approving yet

No new blocking issues, but leaving this open for the author to weigh in on two small items before I mark it approved:

1. **Residual CWE-770 on Gateway/A2A schemas** — those validators run `mode="after"`, so Pydantic still fully coerces the raw list to `List[Dict[str, str]]` before the shared helper's 100-entry cap runs. Negligible in practice (benchmarked ~0.2s for 200k entries), but `max_length=100` on the already-declared `auth_headers` field would close it for free while this code is being touched.
2. **Falsy non-string key handled inconsistently** — in `_encode_auth_headers_list`, `if not key: continue` runs before the `isinstance(key, str)` check, so a falsy non-string key (`0`, `False`, `[]`, `{}`) is silently skipped rather than raising 422, while a truthy non-string key (`123`, `True`, `["x"]`) correctly raises. Not exploitable, just an inconsistency worth a one-line note or test.

Happy to approve once these are addressed or the author confirms they're deliberately deferred as fast-follows.

msureshkumar88 (dismissed):
**Real-process e2e verification — PASSED (8/8 checks)**

Verified the fix against an actual running gateway process (uvicorn, real SQLite file, real JWT auth over HTTP via `curl`) rather than `TestClient`/mocks, and inspected the persisted row directly in the SQLite DB rather than trusting the same API response path being verified. Ran the check twice independently; both runs exited 0.

## Config used
```
DATABASE_URL=sqlite:///$TMPDIR/e2e.db   (fresh throwaway DB per run)
JWT_SECRET_KEY=<throwaway 32+ byte secret>
AUTH_ENCRYPTION_SECRET=<throwaway secret>
AUTH_REQUIRED=true
MCPGATEWAY_UI_ENABLED=true
HOST=127.0.0.1  PORT=18080
```
Server started with: `uv run uvicorn mcpgateway.main:app --host 127.0.0.1 --port 18080`

## Commands run
1. `python3 -m mcpgateway.utils.create_jwt_token -u admin@example.com --secret $JWT_SECRET_KEY -e 60` — real bearer token
2. `curl -X POST /tools` with the exact repro payload from the PR description (`auth_type: authheaders`, `auth_headers: [{X-API-Key: secret}, {X-Tenant: acme}]`)
3. `sqlite3`-level read of `tools.auth_type`/`tools.auth_value` for that row + `decode_auth()` with the server's own `AUTH_ENCRYPTION_SECRET`
4. `curl -X PUT /tools/{id}` replacing the array, re-verified in DB
5. `curl -X POST /tools` with legacy `auth_header_key`/`auth_header_value` (backward-compat path)
6. `curl -X POST /tools` with a malformed header key (`Bad@Key!`)
7. `curl -X POST /tools` with 101 header entries

## Output (key excerpts)

Repro case — this is exactly the bug in #5201 (`auth_value` was `NULL`) and is now fixed:
```
$ sqlite3 lookup + decode_auth for e2e-repro-tool
{"auth_type": "authheaders", "auth_value": "<encrypted, non-null>", "decoded": {"X-API-Key": "secret", "X-Tenant": "acme"}}
✅ PASS: auth_headers array persisted correctly in DB (not NULL, both headers present)
```

Update path:
```
{"auth_type": "authheaders", "auth_value": "<encrypted, non-null>", "decoded": {"X-New-Key": "new-secret"}}
✅ PASS: PUT replaces stored headers
```

Legacy fallback (backward compatibility preserved):
```
{"auth_type": "authheaders", "auth_value": "<encrypted, non-null>", "decoded": {"X-API-Key": "legacy-secret"}}
✅ PASS: legacy header pair persisted correctly
```

Negative cases (client errors, not server crashes):
```
HTTP 422  {"detail":"An error occurred, please try again."}
✅ PASS: malformed header key returns 422 (not 500)

HTTP 422
✅ PASS: 101 headers returns 422
```

Final summary:
```
Summary: 8 passed, 0 failed
```

The verification script is kept as a local harness for this review and wasn't committed to the branch — happy to contribute it as `tests/live_gateway/e2e_tool_auth_headers.sh` in a follow-up if that'd be useful for regression coverage going forward.

Approving — combined with the schema/admin-form/integration unit tests already in the diff, this closes #5201 with real end-to-end confidence.

ja8zyjits (dismissed):
LGTM

## ⚠️ Warnings (Non Blocking, can be followup)

### Documentation Clarifications Needed

1. **Empty-key skipping behavior** (`docs/docs/using/multi-auth-headers.md:111-118`)
   - Current: "Empty header keys are ignored; if no entry has a key, the request is rejected with a 422"
   - Issue: Unclear if applies to individual entries or entire array
   - Fix: Add: "Individual entries with empty keys are skipped. If entire array contains no valid keys after skipping, request rejected with 422."

2. **Error messages not documented** (`docs/docs/using/multi-auth-headers.md:114`)
   - Issue: Users won't know what validation errors look like
   - Fix: Add "Error Messages" subsection with common validation errors

3. **Whitespace-only keys** (`docs/docs/using/multi-auth-headers.md:115`)
   - Current: "Surrounding whitespace trimmed"
   - Issue: Doesn't state whitespace-only keys (e.g., `"   "`) treated as empty
   - Fix: Add: "Keys containing only whitespace are treated as empty and skipped."

ja8zyjits (approved):
LGTM

PR #5315, opened 2026-06-19
bugfix/5152-firefox-closest-blur

### LINKED ISSUE #5152: [BUG]: Firefox: TypeError: e.target.closest is not a function in handleDelegatedBlur (cross-browser compatibility)
## Description

Admin UI throws a JavaScript error in Firefox that does not occur in Chromium-based browsers:

```
Global error: TypeError: e.target.closest is not a function
    handleDelegatedBlur http://127.0.0.1:8000/static/bundle-Cp9iMBCD.js:1
    initializeEventDelegation http://127.0.0.1:8000/static/bundle-Cp9iMBCD.js:1
```

## Root Cause

In Firefox, `blur` events can fire with `e.target` set to a non-Element node (e.g., `window`, `document`, or `SVGElement`) when focus leaves the page or moves to a non-focusable element. `Element.closest()` is only defined on `Element` instances — calling it on `window` or `document` throws `TypeError: e.target.closest is not a function`.

The `handleDelegatedBlur` function in `bundle-Cp9iMBCD.js` (line 1, col ~704565) calls `e.target.closest(...)` without first checking that `e.target` is an `Element`.

**Fix pattern:**
```js
// Before
function handleDelegatedBlur(e) {
    const el = e.target.closest('.some-selector');
    ...
}

// After
function handleDelegatedBlur(e) {
    if (!(e.target instanceof Element)) return;
    const el = e.target.closest('.some-selector');
    ...
}
```

## Browser Compatibility Statement (Proposed)

This issue highlights the need to define and document supported browsers. Proposed minimum support:

| Browser | Minimum Version |
|---------|----------------|
| Firefox | 115 (ESR) |
| Chrome / Chromium | 114 |
| Edge | 114 |
| Safari | 16.4 |

These align with `Element.closest()` broad availability (all modern browsers since 2017) — the bug is not about `closest` support itself but about Firefox passing non-Element targets to blur handlers.

## Steps to Reproduce

1. Open Admin UI in Firefox (`http://127.0.0.1:8000/admin/`)
2. Click into any input field or interactive element
3. Move focus away (tab out or click outside)
4. Observe error in browser console

## Impact

- Unhandled exception in event delegation layer
- May break dropdown close-on-blur, modal dismiss, or other focus-out interactions across the entire UI in Firefox

## Affected File

- `mcpgateway/static/bundle-Cp9iMBCD.js` — `handleDelegatedBlur` / `initializeEventDelegation` (source map needed to locate exact source file)

Nayana-R-Gowda:
 
closes #5152 

**Root Cause**
Firefox fires blur events with non-Element targets (e.g., window, document), and Element.closest() is only defined on Element instances. The handleDelegatedBlur function called event.target.closest() without checking if event.target is an Element.

**Changes Made**

1. Source fix — mcpgateway/admin_ui/eventDelegation.js:118-272
Added if (!(event.target instanceof Element)) return; guard to all 8 delegated event handlers (click, input, change, submit, keydown, focus, blur, reset) to protect against non-Element event targets.

2. Test added — tests/unit/js/eventDelegation.test.js
Added gracefully handles blur with non-Element target (Firefox compatibility) test that dispatches a blur event with a non-Element target, verifying no throw occurs.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-07-03T14:22:26Z",
+  "generated_at": "2026-07-03T13:29:12Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -4359,14 +4359,6 @@
         "is_secret": false,
         "is_verified": false,
         "line_number": 665,
-        "type": "Secret Keyword",
-        "verified_result": null
-      },
-      {
-        "hashed_secret": "f7ac5423ed223f5dc6513d09dd0fd328b7e1176f",
-        "is_secret": false,
-        "is_verified": false,
-        "line_number": 746,
         "type": "Secret Keyword",
         "verified_result": null
       }
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-07-03T13:29:12Z",
+  "generated_at": "2026-07-03T14:53:01Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -4359,6 +4359,14 @@
         "is_secret": false,
         "is_verified": false,
         "line_number": 665,
+        "type": "Secret Keyword",
+        "verified_result": null
+      },
+      {
+        "hashed_secret": "f7ac5423ed223f5dc6513d09dd0fd328b7e1176f",
+        "is_secret": false,
+        "is_verified": false,
+        "line_number": 746,
         "type": "Secret Keyword",
         "verified_result": null
       }
--- a/mcpgateway/admin_ui/eventDelegation.js
+++ b/mcpgateway/admin_ui/eventDelegation.js
@@ -116,6 +116,7 @@ function executeAction(action, args, event, eventFirst = false) {
  * @param {Event} event - The click event
  */
 function handleDelegatedClick(event) {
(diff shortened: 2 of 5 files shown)


marekdano (commented):
The production fix is correct and mergeable — **but the accompanying test does not actually exercise the fix.** It passes identically whether or not the guard is present, so it provides no regression protection.

### Blocking concern: the test is a no-op
The new test asserts `expect(() => document.dispatchEvent(event)).not.toThrow()`. Removing the production guard leaves this test green (full suite: 33/33 passed, no errors):

  - Per the DOM spec, exceptions thrown *inside* an event listener during `dispatchEvent` are "reported" (to the console / a `window` `error` event), **not** re-thrown out of `dispatchEvent`. So `not.toThrow()` is satisfied no matter what the listener does. The original bug was a *global* error, not a propagated one - exactly the case this assertion can't see.

  A test that genuinely catches the regression needs to observe the reported error rather than the `dispatchEvent` return. This approach works (catches the `TypeError` with the guard removed, clean with it present):

  ```js
  it("does not raise a global error on blur with a non-Element target (Firefox)", () => {
    const onError = vi.fn();
    window.addEventListener("error", onError);

    const event = new FocusEvent("blur", { bubbles: true });
    Object.defineProperty(event, "target", { value: { nodeType: 9 }, writable: false }); // document-like
    document.dispatchEvent(event);

    window.removeEventListener("error", onError);
    expect(onError).not.toHaveBeenCalled();
  });
  ```

### Minor notes
  - **Inaccurate comment.** `{ nodeType: 1 }` is `ELEMENT_NODE` — the opposite of what the comment ("simulates document/window") intends. `document` is `nodeType: 9` and `window` has no `nodeType`. The object still isn't an `Element` instance so the test mechanics hold, but using `nodeType: 9` would match the comment and the real Firefox scenario.
  - **Realm edge case (optional).** `instanceof Element` is false for elements from other iframes/realms. Not a concern for this single-document admin UI, but `typeof event.target?.closest === "function"` would be realm-safe and checks the exact capability used. Either is fine.

marekdano (dismissed):
Looks good — minimal, correct fix. Tests pass. Safe to merge.

LGTM 🚀

marekdano (dismissed):
LGTM 🚀

a-effort:
LGTM, after the required rebase :)

ja8zyjits (approved):
LGTM

PR #5330, opened 2026-06-20
fix(sso): merge Keycloak realm/client roles from access_token, not just userinfo/id_token

### LINKED ISSUE #5327: [BUG]: Keycloak SSO silently drops role claims (admin grant fails) when userinfo 401s on split-host config
### 🐞 Bug Summary
Keycloak SSO login can silently fail to grant `platform_admin` (or any `SSO_KEYCLOAK_ROLE_MAPPINGS` role) even when configured correctly, in deployments that use separate internal/public Keycloak URLs (`SSO_KEYCLOAK_BASE_URL` != `SSO_KEYCLOAK_PUBLIC_BASE_URL`). Login succeeds (302 to `/admin`, valid session cookie), but the resulting user gets `permissions: []` instead of admin scopes — no error shown anywhere.

Root cause traced end-to-end against the local `make compose-sso` stack (which already has `SSO_KEYCLOAK_ROLE_MAPPINGS` configured correctly):

1. Gateway exchanges the auth code for tokens via the internal Keycloak URL, then calls `userinfo` on that same internal URL — request gets `401`.
2. `sso_service.py` (~line 1571) has an intentional fallback for this case: on 401 + split-host config, fall back to `verified_id_token_claims` (`_normalize_user_info(provider, verified_id_token_claims)`).
3. Problem: Keycloak's `id_token` does **not** include `realm_access`/`resource_access`/groups by default (only `access_token` does, via the standard "realm roles"/"client roles" client-scope mappers). Confirmed by decoding both tokens directly for a test user with the `gateway-admin` realm role:
   - `id_token.realm_access` → `null`
   - `access_token.realm_access.roles` → `["gateway-admin", ...]`
4. So the fallback path silently loses all role/group info, `_should_user_be_admin()` sees an empty `groups` list, and the `role_mappings`-based admin grant (and any role-based RBAC sync) never fires — even though the operator did everything right.

This was found while investigating a separate report ("create admin user in Keycloak, login via SSO, becomes a normal user"). There's also a related doc gap (tutorial never mentions `SSO_KEYCLOAK_ROLE_MAPPINGS` at all, so the bare-tutorial setup defaults to `{}` and never grants admin), but this issue is the code-level bug: it affects deployments that *did* configure `SSO_KEYCLOAK_ROLE_MAPPINGS` correctly, as long as they use split internal/public Keycloak hostnames — a pattern the gateway explicitly supports via `SSO_KEYCLOAK_PUBLIC_BASE_URL`.

---

### 🧩 Affected Component
- [x] `mcpgateway` - API
- [ ] Other (explain below)

Specifically: `mcpgateway/services/sso_service.py`, function `handle_oauth_callback_with_tokens` / `_get_user_info_from_provider` (~line 1559-1584).

---

### 🔁 Steps to Reproduce

1. `make compose-sso` (local stack: 3 gateway replicas + Keycloak, internal URL `http://keycloak:8080`, public URL `http://localhost:8180`; `SSO_KEYCLOAK_ROLE_MAPPINGS={"gateway-admin":"platform_admin",...}` already set correctly).
2. Create a fresh Keycloak user, assign the `gateway-admin` realm role (Admin REST API or console).
3. Log in via `GET /auth/sso/login/keycloak` → complete Keycloak login form → follow redirect to `/auth/sso/callback/keycloak`.
4. Decode the resulting `jwt_token` cookie.

Observed: login redirects to `/admin` (success), but JWT payload has `"permissions": []` — no admin scope granted.

Enable `LOG_LEVEL=DEBUG` on the gateway to see the corroborating log line:
```
User info request returned 401 for keycloak with split-host config (public=http://localhost:8180, internal=http://keycloak:8080). Falling back to id_token claims.
```

Direct confirmation of the missing claims (password grant against the same realm/client):
```python
id_token claims:     {"realm_access": null, "resource_access": null, "groups": null, ...}
access_token claims: {"realm_access": {"roles": ["gateway-admin", ...]}, ...}
```

---

### 🤔 Expected Behavior
A user with a Keycloak role mapped via `SSO_KEYCLOAK_ROLE_MAPPINGS` should reliably get the mapped gateway role/admin status on login, regardless of whether the userinfo call succeeds or falls back due to split-host config. The 401-fallback path should not silently drop role/group claims that are already present (and already in hand) on the `access_token`.

Possible fix direction: when falling back on 401, also decode the `access_token` payload (already received from the token endpoint in the same exchange) and merge `realm_access`/`resource_access`/groups into the claims passed to `_normalize_user_info`, instead of relying solely on `id_token` claims.

---

### 📓 Logs / Error Output
```
gateway-2  | ... User info request returned 401 for keycloak with split-host config (public=http://localhost:8180, internal=http://keycloak:8080). Falling back to id_token claims.
```
No error is surfaced to the end user or admin — login appears fully successful.

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Version or commit | `Release/v1.0.4` branch, local HEAD |
| Runtime | Python 3.12, FastAPI dev/compose stack |
| Platform / OS | Linux (WSL2), Docker Compose |
| Container | Docker Compose (`docker-compose.yml` + `docker-compose.sso.yml --profile sso`) |

---

- Reproduced against `docker-compose.sso.yml`, which is split-host by design (`SSO_KEYCLOAK_BASE_URL=http://keycloak:8080`, `SSO_KEYCLOAK_PUBLIC_BASE_URL=http://localhost:8180`) — exactly the topology this feature flag exists to support (internal Docker/K8s service URL vs. public browser-facing URL).
- Separately, the SSO Keycloak tutorial (`docs/docs/manage/sso-keycloak-tutorial.md`) never documents `SSO_KEYCLOAK_ROLE_MAPPINGS` at all, so a reader following it verbatim hits a related-but-distinct failure (role mapping is simply unconfigured, defaults to `{}`). That's a docs gap, not this code bug — happy to file separately if useful.
- Also noticed (unrelated, minor): `scripts/test-sso-flow.sh`'s `assert_browser_callback_login()` is missing `-c "$cookie_file"` on its first `curl` call, so the gateway's `sso_session_id` cookie set on that response gets dropped before being captured, which can cause spurious `sso_failed` state-binding errors in the smoke test. Can file separately too if wanted.

msureshkumar88:
## Summary

- Keycloak's built-in **realm roles** / **client roles** client-scope mappers default `access.token.claim=true` but `id.token.claim=userinfo.token.claim=false`. Since the gateway reads SSO claims from the userinfo response (falling back to `id_token` only on the documented split-host 401 case), `realm_access`/`resource_access` were silently absent on any stock Keycloak setup — even with `SSO_KEYCLOAK_ROLE_MAPPINGS` configured correctly and the right role assigned. New admins (or any role-mapped user) fell through to the default role.
- `_get_user_info()` / `_enrich_user_data_from_claims()` (`mcpgateway/services/sso_service.py`) now also decode the already-in-hand `access_token` and merge `realm_access`/`resource_access`/`groups` when missing from userinfo and id_token — fixes both the normal 200-OK path and the existing 401 split-host fallback.
- Fixed a related inconsistency in `_map_groups_to_roles()`: `role_mappings` lookups were case-sensitive while `_should_user_be_admin()` already matched case-insensitively. This could grant `is_admin=True` with no matching RBAC role row when IdP role casing differed from the configured mapping key.
- Updated the local Keycloak dev seed (`infra/keycloak/realm-export.json`) to explicitly enable `id_token`/`userinfo` claim inclusion on the realm/client role mappers, and documented the default-mapper gotcha in `docs/docs/manage/sso-keycloak-tutorial.md` (Step 4.4 + troubleshooting section).

Closes #5327. Both findings posted on the issue (https://github.com/IBM/mcp-context-forge/issues/5327#issuecomment-4756781399) are addressed by this PR.

## Config used to reproduce/verify locally

Local Docker Compose SSO stack (`make compose-sso` equivalent: `docker compose -f docker-compose.yml -f docker-compose.sso.yml --profile sso up -d`):

- Gateway: `http://localhost:8080`, 3 replicas behind nginx
- Keycloak: `http://localhost:8180` (admin console `admin` / `changeme`)
- Realm: `mcp-gateway`, client: `mcp-gateway`
- `SSO_KEYCLOAK_BASE_URL=http://keycloak:8080`, `SSO_KEYCLOAK_PUBLIC_BASE_URL=http://localhost:8180`
- `SSO_KEYCLOAK_ROLE_MAPPINGS={"gateway-admin":"platform_admin","gateway-developer":"developer","gateway-viewer":"viewer"}`
- `SSO_KEYCLOAK_MAP_REALM_ROLES=true`, `SSO_KEYCLOAK_DEFAULT_ROLE=viewer`

## Test plan

- [x] `pytest tests/unit/mcpgateway/services/test_sso_service.py tests/unit/mcpgateway/services/test_sso_entra_role_mapping.py tests/unit/mcpgateway/services/test_sso_generic_oidc_role_mapping.py tests/unit/mcpgateway/services/test_sso_user_normalization.py tests/unit/mcpgateway/services/test_sso_admin_assignment.py tests/unit/mcpgateway/utils/test_sso_bootstrap.py tests/unit/mcpgateway/routers/test_sso_router.py tests/integration/test_sso_adfs_integration.py` — all pass, including 4 new regression tests added in this PR.
- [x] Manual end-to-end verification against a rebuilt local stack (fresh Keycloak volume re-imported with the updated mapper config, gateway image rebuilt from this branch):
  1. `docker compose -f docker-compose.yml -f docker-compose.sso.yml --profile sso build gateway`
  2. Recreate Keycloak with a fresh volume so the updated `realm-export.json` mappers actually get imported (Keycloak's `--import-realm` won't retrofit mapper config onto an existing client):
     ```
     docker compose -f docker-compose.yml -f docker-compose.sso.yml --profile sso stop keycloak gateway
     docker compose -f docker-compose.yml -f docker-compose.sso.yml --profile sso rm -f keycloak
     docker volume rm mcp-context-forge_keycloakdata
     docker compose -f docker-compose.yml -f docker-compose.sso.yml --profile sso up -d keycloak gateway
     ```
  3. Create a fresh Keycloak user via the Admin REST API, assign it the `gateway-admin` realm role.
  4. Drive the real browser-facing OAuth flow: `GET /auth/sso/login/keycloak` -> Keycloak login form -> `GET /auth/sso/callback/keycloak`.
  5. Decode the resulting `jwt_token` cookie and call `GET /admin/overview/partial` (RBAC-gated, `allow_admin_bypass=False`) and `GET /admin/` with `Accept: text/html`.

  Before the fix: `scopes.permissions: []`, `/admin/overview/partial` -> `403 Admin privileges required`.
  After the fix: `scopes.permissions: ["*"]`, `/admin/overview/partial` -> `200`, `/admin/` -> `200`.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-06-18T13:53:55Z",
+  "generated_at": "2026-06-20T09:17:14Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -1378,7 +1378,7 @@
         "hashed_secret": "ff2ccd73154c1c71ef9a2982fe16e3c529e7a1ae",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 104,
+        "line_number": 105,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -1386,7 +1386,7 @@
         "hashed_secret": "74f82b992f8a92b849c2be77447f98a510338333",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 105,
+        "line_number": 106,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -1394,7 +1394,7 @@
         "hashed_secret": "64814a3b7fd8444a56ad3641fd3451c6deaf0757",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 113,
+        "line_number": 114,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -1402,7 +1402,7 @@
         "hashed_secret": "af84d91fde168566c7dc18f3121ea2fbe651af1f",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 142,
+        "line_number": 143,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -1410,7 +1410,7 @@
         "hashed_secret": "d1081e351fbebe0deb0c2867d5f731c8f9cc3fd8",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 395,
+        "line_number": 396,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -1418,7 +1418,7 @@
         "hashed_secret": "b7e9a6e2e04ed3edbf8b4e21def4beccbb6eb6b1",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 396,
+        "line_number": 397,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -1426,7 +1426,7 @@
         "hashed_secret": "6d244f58364223afcc2322f696137b01ece78d9d",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 397,
+        "line_number": 398,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -1434,7 +1434,7 @@
         "hashed_secret": "8bf177a72c93155020ebca9ee7f3c6e0fbdee946",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 415,
+        "line_number": 416,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -1442,7 +1442,7 @@
         "hashed_secret": "c41e6cd4245e404f07635fdbac351aad4d54a213",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 514,
+        "line_number": 515,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -1450,7 +1450,7 @@
         "hashed_secret": "6568a9761e26d16a111247f279b6fcabff1c8c86",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 549,
+        "line_number": 550,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -1458,7 +1458,7 @@
         "hashed_secret": "314fc7fd580f8c510be196143946bbe5ea753443",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 550,
+        "line_number": 551,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -1466,7 +1466,7 @@
         "hashed_secret": "afb9d1dcf212fa33d079a070ab90f0bef03c324b",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 564,
+        "line_number": 565,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -1474,7 +1474,7 @@
         "hashed_secret": "2736fab291f04e69b62d490c3c09361f5b82461a",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 571,
+        "line_number": 572,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -1482,7 +1482,7 @@
         "hashed_secret": "7542c8f677ffbbe75a5301a5c76f88401dc42897",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 580,
+        "line_number": 581,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -1490,7 +1490,7 @@
         "hashed_secret": "ca20728d35f00c3a4c21b1fc8b0086b59f89df2d",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 585,
+        "line_number": 586,
         "type": "Secret Keyword",
         "verified_result": null
       },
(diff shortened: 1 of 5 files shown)


madhu-mohan-jaishankar (approved):
Looks good to me. Addresses the Keycloak role-claim issue, covers both the normal and split-host fallback paths, and adds solid regression tests. The docs and local Keycloak config updates are also helpful. LGTM!

Lang-Akshay:
## Thanks for the PR - @msureshkumar88 
## Changes are surgical and only touches Keycloak 
## Both the configuration gap and RBAC issue is resolved 
| Provider         | Impact                          | Notes                                                      |
| ---------------- | ------------------------------- | ---------------------------------------------------------- |
| Keycloak         | Fixed                           | Fixes role claim visibility issue                          |
| Okta             | None                            | Uses generic OIDC path; configure groups_claim in metadata |
| Azure AD / Entra | None                            | Has own provider.id == "entra" path, untouched             |
| Auth0            | None                            | Generic OIDC path; needs groups_claim configured           |
| Google Workspace | None                            | provider.id == "google" path untouched                     |
| GitHub SSO       | None                            | provider.id == "github" path untouched                     |
| Generic OIDC     | Case-insensitive match improved | Fix 2 benefits all providers                               |

## Google SSO Tested and works ✅
```
SSO_ENABLED=true
SSO_GOOGLE_ENABLED=true
SSO_GOOGLE_CLIENT_ID=your-client-id.googleusercontent.com
SSO_GOOGLE_CLIENT_SECRET=your-client-secret

# Optional: auto-grant admin to these email domains
SSO_GOOGLE_ADMIN_DOMAINS=gmail.com

# Optional: restrict login to these domains only
SSO_TRUSTED_DOMAINS=gmail.com

# SSO General Settings
SSO_AUTO_CREATE_USERS=true
```

PR #241, opened 2025-07-04
Improve test coverage in test_tool_service

madhav165:
# ✨ Feature / Enhancement PR

## 🚀 Summary (1–2 sentences)
- Improves test coverage from 60% to 92% for tool_service
- Overall coverage improves from 70% to 73%

## Status
Needs update due to changes in PR #215

PR #5332, opened 2026-06-21
feat(tests): E2E integration tests for cpex plugins in the gateway

madhu-mohan-jaishankar:
## 🔗 Related Issue

Closes [#241](https://github.ibm.com/contextforge-org/internal_issues/issues/241)

---

## 📝 Summary

Adds a self-contained E2E integration test suite that exercises each managed cpex plugin inside a live ContextForge gateway, runnable both locally and in the merge queue.

Each plugin is tested in isolation. Where the plugin hooks tool calls, two enforcement paths are covered:

- **static** — the gateway boots with a single-plugin enforce config derived from `plugins/config.yaml`
- **binding** — the gateway boots with every plugin disabled and the suite creates a runtime tool-plugin-binding to enforce for the test's team+tool

Both configs are derived from `plugins/config.yaml` at run time via `make_enforce_config.py`, so there is no second copy of any plugin block to keep in sync. The suites never import the cpex packages directly — the gateway loads them from `PLUGINS_CONFIG_FILE` — so a broken wheel or compiled Rust extension surfaces as a failing job rather than a silently skipped test.

**Plugins covered (13 legs total):**

| Plugin | static | binding |
|---|---|---|
| SecretsDetection | ✅ | ✅ |
| EncodedExfilDetector | ✅ | ✅ |
| URLReputationPlugin | ✅ | — (no tool hook) |
| RateLimiterPlugin | ✅ | ✅ |
| RetryWithBackoffPlugin | ✅ | ✅ |
| PIIFilterPlugin | ✅ | ✅ |
| SQLSanitizer | ✅ | ✅ |

**Files added/changed:**

- `tests/live_gateway/plugins/` — 7 E2E test modules, shared helpers/fixtures, `make_enforce_config.py`, `run_plugin_tests.sh` (local runner), `Makefile.plugin-integration`
- `.github/workflows/plugin-integration.yml` — 13-leg CI matrix with Redis service for rate-limiter legs; triggers on `merge_group` and on `pull_request` when plugin-related paths change
- `.github/workflows/plugin-integration-required.yml` — skip reporter that reports the same `Plugin Integration Complete` check name on PRs whose paths don't trigger the full suite, so branch protection is never blocked
- `.github/workflows/sql-sanitizer.yml` — changed to `workflow_dispatch`-only (superseded by `plugin-integration.yml`)
- `mcp-servers/rust/fast-time-server/` — streamable HTTP MCP echo server used as the backend for all plugin suites (`/health` endpoint, `echo` tool with `message` arg)
- `Makefile` — `-include tests/live_gateway/plugins/Makefile.plugin-integration` + help entries for 8 new targets

---

## 🏷️ Type of Change

- [ ] Bug fix
- [x] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

All 13 legs validated locally against a live gateway before this PR was opened.

| Check | Command | Status |
|---|---|---|
| Full plugin suite (all 13 legs) | `make test-plugin-integration` | ✅ 13/13 PASS (~2m26s) |
| Single plugin (example) | `make test-plugin-secrets-detection` | ✅ PASS |
| Rate limiter (needs Redis) | `make test-plugin-rate-limiter` | ✅ PASS (auto-starts Redis container) |
| SQL sanitizer (native plugin) | `make test-plugin-sql-sanitizer` | ✅ PASS |
| Derived config sanity | `python tests/live_gateway/plugins/make_enforce_config.py --source plugins/config.yaml --plugin SecretsDetection --output /tmp/test.yaml` | ✅ |

Local run output (abridged):
```
[SecretsDetection/static]   PASS
[SecretsDetection/binding]  PASS
[EncodedExfilDetector/static]  PASS
[EncodedExfilDetector/binding] PASS
[URLReputationPlugin/static]   PASS
[RateLimiterPlugin/static]     PASS
[RateLimiterPlugin/binding]    PASS
[RetryWithBackoffPlugin/static]  PASS
[RetryWithBackoffPlugin/binding] PASS
[PIIFilterPlugin/static]   PASS
[PIIFilterPlugin/binding]  PASS
[SQLSanitizer/static]      PASS
[SQLSanitizer/binding]     PASS

Results: 13 passed, 0 failed
```

---

## ✅ Checklist

---

## 📓 Notes

**Branch protection:** After this PR merges, set `Plugin Integration Complete` as a required status check on `main`. The skip reporter (`plugin-integration-required.yml`) ensures every PR that doesn't touch plugin paths still reports this check as green, so branch protection is never a blocker.

**Redis:** The rate-limiter legs require Redis. In CI the workflow spins up a `redis:7-bookworm` service container. Locally, `run_plugin_tests.sh` checks for a running Redis and starts a `redis:7-alpine` container automatically if one isn't found.

**`sql-sanitizer.yml`:** The old single-plugin workflow is kept but demoted to `workflow_dispatch`-only. It is fully superseded by the `plugin-integration.yml` matrix. This will be removed once we ensure plugin-integration.yml runs without any issues and covers all plugin integration scenarios.


jonpspri:
I'm okay with what I'm seeing, but please get a review from someone who's been more recently involved with the CI/CD chain.

tests/live_gateway/plugins/run_plugin_tests.sh
```diff
@@ -0,0 +1,330 @@
...
+# Plugin registry
+# Each entry: "NAME|TEST_FILE|CONFIG_OVERRIDE|BINDING_SUPPORTED|NEEDS_REDIS"
+# CONFIG_OVERRIDE: empty string or a KEY=JSON literal (no outer quotes needed)
+# ---------------------------------------------------------------------------
+PLUGINS=(
+    "SecretsDetection|test_secrets_detection_e2e.py||yes|no"
+    "EncodedExfilDetector|test_encoded_exfil_detection_e2e.py||yes|no"
+    "URLReputationPlugin|test_url_reputation_e2e.py||no|no"
+    "RateLimiterPlugin|test_rate_limiter_e2e.py||yes|yes"
+    "RetryWithBackoffPlugin|test_retry_with_backoff_e2e.py||yes|yes"
```
lucarlig, line 57: does retry with back off need redis?

.github/workflows/plugin-integration.yml
```diff
@@ -0,0 +1,351 @@
...
+#
+# Runs in the merge queue (merge_group) so plugin behaviour is verified before
+# code lands, and on pull_request when the inputs that affect plugin loading or
+# the suites themselves change.
+
+on:
+    merge_group:
+    pull_request:
+        branches: [main]
+        types: [opened, synchronize, reopened]
```
lucarlig: `ready_for_review` should also trigger, otherwise draft pr converted will not run ci automatically.

lucarlig (changes requested):

tests/live_gateway/plugins/run_plugin_tests.sh
madhu-mohan-jaishankar: yes, it stores the state.

lucarlig (dismissed):
LGTM
test pass locally

msureshkumar88 (changes requested):
Nice work here — this closes a real gap (managed cpex plugins were previously only unit-tested in isolation, with no live-gateway coverage), and the static/binding dual-path design plus deriving both configs from `plugins/config.yaml` at run time is a solid way to avoid a second copy of plugin config drifting out of sync. Cross-checked against the `cpex-plugins` repo: all 6 managed cpex plugins (`secrets_detection`, `encoded_exfil_detection`, `pii_filter`, `rate_limiter`, `retry_with_backoff`, `url_reputation`) are covered, and `SQLSanitizer` is correctly excluded as a native (non-cpex) plugin — no coverage gaps there.

## Blocking

**1. `pyproject.toml` — `cpex-retry-with-backoff` pin looks like an accidental downgrade.**
Line 281 narrows `cpex-retry-with-backoff>=0.3.5` down to `>=0.3.1,<0.3.2`. Every other cpex-* pin in the same block moved forward in this PR; only this one rolled back, and `uv.lock` confirms the resolved version drops from 0.3.5 to 0.3.1. The `[plugins]` extra is a real install target (`pip install mcp-contextforge-gateway[plugins]`), not test-only, so this ships an older, less-patched build to anyone installing fresh after merge. Was this intentional (e.g. working around a regression in 0.3.2+)? If so, could you add a comment explaining what broke — otherwise this looks like it should be reverted to `>=0.3.5`.

## Suggestions

**2. `_echo()` helper duplicated across 6 test files** (`test_encoded_exfil_detection_e2e.py`, `test_pii_filter_e2e.py`, `test_retry_with_backoff_e2e.py`, `test_secrets_detection_e2e.py`, `test_sql_sanitizer_e2e.py`, `test_url_reputation_e2e.py`) — byte-identical `initialize_session` + `call_tool` logic in each. Worth moving into `_helpers.py` once so a future change to the MCP handshake only needs to land in one place.

**3. `_enforcement` autouse fixture boilerplate duplicated 7×**, differing only by the `PLUGIN_NAME` string. A parameterized fixture factory in `conftest.py` would collapse this to one definition.

**4. `fast-time-server` gets rebuilt independently in all 13 CI matrix legs** in `plugin-integration.yml` — only the Cargo registry cache is shared, not the compiled binary. Building once and sharing via `actions/upload-artifact`/`download-artifact` would cut CI time noticeably given this runs on every PR touching plugin config.

**5. CI workflow re-implements the derive-config → boot-gateway → run-pytest sequence in YAML+bash, separately from `run_plugin_tests.sh`** (which already encodes the same flow for local use). Two copies invite drift between what passes locally and what CI actually runs — might be worth having CI shell out to the script directly.

## Minor

**6. `.pre-commit-config.yaml` exclude regex `tests/.*/_[^/]+\.py|tests/.*/make_[^/]+\.py`** is a filename-shape carve-out sized to this PR's own `_helpers.py`/`make_enforce_config.py`, unlike the existing directory-based excludes which encode a real naming convention. Any future `_something.py` under `tests/` silently skips the naming check too — might be safer to scope this to `tests/live_gateway/plugins/` specifically.

**7. `FLAKY_STATE` in `fast-time-server/src/server.rs` only removes an entry on the success path**, never on failure/abort. Not exploitable today since every test uses a fresh UUID key, but there's no TTL/cap if a future consumer reuses keys on a long-lived process — worth a passing thought, not blocking.

**8. `.github/workflows/plugin-integration-required.yml` isn't itself in `plugin-integration.yml`'s trigger paths.** A future bug in the skip-reporter's own logic wouldn't be caught by the real matrix, since editing the skip-reporter doesn't trigger it. Might be worth adding it to the paths list for completeness.

madhu-mohan-jaishankar:
> Nice work here — this closes a real gap (managed cpex plugins were previously only unit-tested in isolation, with no live-gateway coverage), and the static/binding dual-path design plus deriving both configs from `plugins/config.yaml` at run time is a solid way to avoid a second copy of plugin config drifting out of sync. Cross-checked against the `cpex-plugins` repo: all 6 managed cpex plugins (`secrets_detection`, `encoded_exfil_detection`, `pii_filter`, `rate_limiter`, `retry_with_backoff`, `url_reputation`) are covered, and `SQLSanitizer` is correctly excluded as a native (non-cpex) plugin — no coverage gaps there.
> 
> A few things worth addressing before merge:
> 
> ## Blocking
> **1. `pyproject.toml` — `cpex-retry-with-backoff` pin looks like an accidental downgrade.** Line 281 narrows `cpex-retry-with-backoff>=0.3.5` down to `>=0.3.1,<0.3.2`. Every other cpex-* pin in the same block moved forward in this PR; only this one rolled back, and `uv.lock` confirms the resolved version drops from 0.3.5 to 0.3.1. The `[plugins]` extra is a real install target (`pip install mcp-contextforge-gateway[plugins]`), not test-only, so this ships an older, less-patched build to anyone installing fresh after merge. Was this intentional (e.g. working around a regression in 0.3.2+)? If so, could you add a comment explaining what broke — otherwise this looks like it should be reverted to `>=0.3.5`.
> 
> ## Suggestions
> **2. `_echo()` helper duplicated across 6 test files** (`test_encoded_exfil_detection_e2e.py`, `test_pii_filter_e2e.py`, `test_retry_with_backoff_e2e.py`, `test_secrets_detection_e2e.py`, `test_sql_sanitizer_e2e.py`, `test_url_reputation_e2e.py`) — byte-identical `initialize_session` + `call_tool` logic in each. Worth moving into `_helpers.py` once so a future change to the MCP handshake only needs to land in one place.
> 
> **3. `_enforcement` autouse fixture boilerplate duplicated 7×**, differing only by the `PLUGIN_NAME` string. A parameterized fixture factory in `conftest.py` would collapse this to one definition.
> 
> **4. `fast-time-server` gets rebuilt independently in all 13 CI matrix legs** in `plugin-integration.yml` — only the Cargo registry cache is shared, not the compiled binary. Building once and sharing via `actions/upload-artifact`/`download-artifact` would cut CI time noticeably given this runs on every PR touching plugin config.
> 
> **5. CI workflow re-implements the derive-config → boot-gateway → run-pytest sequence in YAML+bash, separately from `run_plugin_tests.sh`** (which already encodes the same flow for local use). Two copies invite drift between what passes locally and what CI actually runs — might be worth having CI shell out to the script directly.
> 
> ## Minor
> **6. `.pre-commit-config.yaml` exclude regex `tests/.*/_[^/]+\.py|tests/.*/make_[^/]+\.py`** is a filename-shape carve-out sized to this PR's own `_helpers.py`/`make_enforce_config.py`, unlike the existing directory-based excludes which encode a real naming convention. Any future `_something.py` under `tests/` silently skips the naming check too — might be safer to scope this to `tests/live_gateway/plugins/` specifically.
> 
> **7. `FLAKY_STATE` in `fast-time-server/src/server.rs` only removes an entry on the success path**, never on failure/abort. Not exploitable today since every test uses a fresh UUID key, but there's no TTL/cap if a future consumer reuses keys on a long-lived process — worth a passing thought, not blocking.
> 
> **8. `.github/workflows/plugin-integration-required.yml` isn't itself in `plugin-integration.yml`'s trigger paths.** A future bug in the skip-reporter's own logic wouldn't be caught by the real matrix, since editing the skip-reporter doesn't trigger it. Might be worth adding it to the paths list for completeness.

retry with backoff didn't work properly after moving to rust major architecture. not sure how it was passing for releases. refer https://github.com/IBM/cpex-plugins/issues/124. That is the reason for downgrade. we can bump it again after we fix the issue with retry with backoff.

lucarlig (approved):

msureshkumar88 (approved):
Thanks for the explanation on the retry-with-backoff pin — makes sense given the upstream regression tracked in IBM/cpex-plugins#124.

One nit, non-blocking: could you add a one-line comment above `cpex-retry-with-backoff>=0.3.1,<0.3.2` in `pyproject.toml` (line 281) referencing cpex-plugins#124? Right now the pin is bare, so anyone reading the file later won't know why it's pinned below the others without digging through PR history.

The other suggestions from my earlier review (helper dedup, fixture factory, CI binary caching, script reuse, pre-commit exclude scope, FLAKY_STATE cleanup, skip-reporter trigger paths) are all still open too, but none are blocking — fine to pick up as follow-ups.

Approving.

PR #5337, opened 2026-06-22
fix(docker): restrict hermetic wheel closure to s390x/ppc64le only

madhu-mohan-jaishankar:
## 🔗 Related Issue

Closes #

---

## 📝 Summary

The hermetic wheel closure (pre-built `.whl` images on GHCR) was extended to all four architectures in PR #5287, but amd64 and arm64 have full PyPI manylinux coverage for all dependencies and do not need it. Building and consuming wheel images for those arches introduced fragility: a stale GHCR image (e.g. `cryptography==48.0.0`) would override the current `uv.lock` and break the hermetic install when the dependency version was bumped — exactly what happened after `cryptography>=48.0.1` was added to `pyproject.toml`.

**Changes:**
- **`Containerfile.lite`**: Default `WHEELS_REF` changed from `ghcr.io/ibm/mcp-context-forge-wheels:${TARGETARCH}-latest` to `${UBI_MINIMAL}`. On amd64/arm64, the wheels stage now pulls an empty UBI image so `/tmp/wheels` is always empty and the install step falls straight through to PyPI. The generic `ls -A /tmp/wheels/*.whl` check (from PR #5287) is preserved unchanged — arch-awareness lives entirely in what `WHEELS_REF` points to, not in a `uname -m` test. `ARG TARGETARCH` is removed as it is no longer needed.
- **`docker-multiplatform.yml`**: Removed amd64 and arm64 from the `wheels` job matrix (no more publishing unused wheel images for those arches). Updated the "Compute wheel image ref" step to short-circuit for amd64/arm64 — always sets `WHEELS_REF=UBI_MINIMAL` with no GHCR probing. s390x/ppc64le retain the full lock-hash → latest → UBI_MINIMAL fallback chain.

---

## 🏷️ Type of Change

- [x] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

`make docker-prod` was failing on amd64 with:

```
ERROR: Cannot install cryptography==48.0.0 because these package versions have conflicting dependencies.
The conflict is caused by: mcp-context-forge requires cryptography>=48.0.1
```

After this fix, `make docker-prod` on amd64 resolves all dependencies directly from PyPI and completes successfully.

| Check | Command | Status |
|---|---|---|
| Local amd64 build | `make docker-prod` | ✅ passes (PyPI path, no stale wheel image) |

---

## ✅ Checklist

---

## 📓 Notes (optional)

s390x and ppc64le legitimately need the wheel closure because `psycopg-binary` has no s390x PyPI wheel, and `tiktoken`/`cryptography` require a Rust compiler to build from source on ppc64le (only present in the builder stage when `ENABLE_RUST=true`). Those arches are unaffected by this change.


.github/workflows/docker-multiplatform.yml
```diff
@@ -73,12 +74,6 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - platform: linux/amd64
```
cafalchio, line 76: Thanks for removing it. Make it faster in make docker-prod too.
LGTM

cafalchio (approved):
LGTM

PR #5343, opened 2026-06-23
 bugfix/5325-render-prompt-csrf-token-invalid

### LINKED ISSUE #5325: [BUG]: Render Prompt button fails with CSRF_TOKEN_INVALID for platform-admin (non-email-auth) sessions
### Bug Summary

Clicking **Render Prompt** on the Admin UI Prompts tab (`/admin/#prompts`) fails with:

```json
{"detail":"CSRF validation failed","code":"CSRF_TOKEN_INVALID"}
```

This happens for a normal platform admin login (basic auth / default config, `EMAIL_AUTH_ENABLED` not set). No CSRF exemption or special config should be required for an authenticated admin to use this feature.

---

### Affected Component

---

### Root Cause

The admin dashboard route (`GET /admin`) only binds the CSRF cookie to the current user/session via HMAC when `settings.email_auth_enabled` is `True`:

```python
# mcpgateway/admin.py:4161-4162
csrf_user_id, csrf_session_id = None, None
if getattr(settings, "email_auth_enabled", False):
    ...  # csrf_user_id / csrf_session_id only get populated inside this block
```

That `csrf_user_id`/`csrf_session_id` pair is then passed straight through to:

```python
# mcpgateway/admin.py:4262
_set_admin_csrf_cookie(request, response, user_id=csrf_user_id, session_id=csrf_session_id)
```

When `email_auth_enabled` is `False` (the default — basic-auth/platform-admin login), both values stay `None`, so `_set_admin_csrf_cookie()` (`mcpgateway/admin.py:1662`) takes the unbound branch:

```python
# mcpgateway/admin.py:1684-1688
if user_id and session_id:
    csrf_token = get_csrf_service().generate_csrf_token(user_id=user_id, session_id=session_id)
else:
    existing_token = request.cookies.get(ADMIN_CSRF_COOKIE_NAME)
    csrf_token = existing_token if isinstance(existing_token, str) and len(existing_token) >= 32 else secrets.token_urlsafe(32)
```

This produces a plain random token instead of the HMAC-bound token the global `CSRFMiddleware` expects.

The Admin UI's `fetchWithTimeout()` helper (`mcpgateway/admin_ui/utils.js:155-157`) correctly attaches this cookie value as the `X-CSRF-Token` header on the prompt-render POST (`/prompts/{id}`), and the double-submit comparison in `enforce_admin_csrf` would accept it — but that request actually hits the route directly, which is protected by the global `CSRFMiddleware` (`mcpgateway/middleware/csrf_middleware.py:174-177`), not `enforce_admin_csrf`. That middleware recomputes the HMAC from the authenticated user/session (`csrf_service.validate_csrf_token(csrf_token, user_id, session_id)`) and rejects any token that wasn't derived that way — which a random/unbound token never satisfies. Hence every non-GET admin-UI call (prompt render, tool test, etc.) fails for platform-admin sessions, while it works fine when `EMAIL_AUTH_ENABLED=true` (where `admin_login_handler` does bind the cookie — see `mcpgateway/admin.py:4233`, `4508`, `4527`).

---

### Steps to Reproduce

1. Run gateway with default config (`EMAIL_AUTH_ENABLED` unset/false), log in to `/admin` as the platform admin (basic auth).
2. Go to `/admin/#prompts`, open any prompt, fill in arguments.
3. Click **Render Prompt**.
4. Request to `POST /prompts/{id}` returns `403` with `{"detail":"CSRF validation failed","code":"CSRF_TOKEN_INVALID"}`.

---

### Expected Behavior

An authenticated platform admin should be able to render a prompt from the Admin UI without needing `EMAIL_AUTH_ENABLED=true` or any other config exception. The CSRF cookie issued on `/admin` page load should be bound (HMAC) to the admin's session for any auth mode, not only when email auth is enabled.

---

### Additional Context

- `EMAIL_AUTH_ENABLED=true` masks the bug because `admin_login_handler` binds the CSRF cookie to `user.email` / `session_jti` at login time.
- Likely fix: derive `csrf_user_id` / `csrf_session_id` from the authenticated `user`/JWT for the `/admin` page-load route regardless of `email_auth_enabled`, mirroring what `admin_login_handler` does.

Nayana-R-Gowda:

closes #5325 

Root Cause: The admin_ui route (GET /admin) was guarded by if getattr(settings, "email_auth_enabled", False), so the JWT session cookie (which provides sub/jti for CSRF HMAC binding) was only set for email-auth users. For platform admin (basic auth) sessions — the default — no JWT cookie was set, so _set_admin_csrf_cookie fell back to a plain random token that the global CSRFMiddleware couldn't HMAC-validate.
Changes (mcpgateway/admin.py):
1. _set_admin_csrf_cookie() (line 1658): Added optional user_id and session_id parameters. When both are provided, generates an HMAC-bound token via get_csrf_service().generate_csrf_token() — same algorithm the CSRF middleware uses for validation. Otherwise preserves the existing random-token fallback for anonymous pages.
2. admin_ui() (lines 4151-4223): Removed the if email_auth_enabled guard so the JWT session cookie is set for all authenticated admin users, regardless of auth provider. After generating the JWT, captures sub_claim and jti as csrf_user_id/csrf_session_id and passes them to _set_admin_csrf_cookie().

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-06-18T13:53:55Z",
+  "generated_at": "2026-06-30T16:06:55Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -5072,7 +5072,7 @@
         "hashed_secret": "995e87a207ec41ef95f938dcf53a4184312b0d93",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 11421,
+        "line_number": 11569,
         "type": "Hex High Entropy String",
         "verified_result": null
       },
@@ -5080,7 +5080,7 @@
         "hashed_secret": "a0281cd072cea8e80e7866b05dc124815760b6c9",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 17838,
+        "line_number": 17986,
         "type": "Secret Keyword",
         "verified_result": null
       }
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-06-30T16:06:55Z",
+  "generated_at": "2026-06-30T16:11:27Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
--- a/mcpgateway/admin.py
+++ b/mcpgateway/admin.py
@@ -153,6 +153,7 @@ from mcpgateway.services.argon2_service import Argon2PasswordService
 from mcpgateway.services.audit_trail_service import get_audit_trail_service
 from mcpgateway.services.catalog_service import catalog_service
 from mcpgateway.services.content_security import ContentSizeError, ContentTypeError, TemplateValidationError
+from mcpgateway.services.csrf_service import clear_csrf_cookie, get_csrf_service
 from mcpgateway.services.email_auth_service import AuthenticationError, EmailAuthService, PasswordValidationError
 from mcpgateway.services.encryption_service import get_encryption_service
 from mcpgateway.services.export_service import ExportError, ExportService
@@ -1671,9 +1672,7 @@ def _set_admin_csrf_cookie(request: Request, response: Response, user_id: str | 
     Returns:
         CSRF token value stored in the response cookie.
     """
-    if user_id and session_id:
-        from mcpgateway.services.csrf_service import get_csrf_service
-
+    if user_id is not None and session_id is not None:
         csrf_token = get_csrf_service().generate_csrf_token(user_id=user_id, session_id=session_id)
     else:
         existing_token = request.cookies.get(ADMIN_CSRF_COOKIE_NAME)
@@ -4155,7 +4154,7 @@ async def admin_ui(
     try:
         # Determine the admin user email
         admin_email = get_user_email(user)
-        is_admin_flag = bool(user.get("is_admin") if isinstance(user, dict) else True)
+        is_admin_flag = bool(user.get("is_admin") if isinstance(user, dict) else getattr(user, "is_admin", False))
         full_name = getattr(settings, "platform_admin_full_name", "Platform User")
         if isinstance(user, dict):
             full_name = user.get("full_name") or full_name
@@ -4193,8 +4192,11 @@ async def admin_ui(
                         auth_provider = "keycloak"
 
         # Generate a lightweight session JWT token
-        email_user = db.query(EmailUser).filter(EmailUser.email == admin_email).first()
-        sub_claim = str(email_user.id) if email_user else admin_email
+        if settings.email_auth_enabled:
+            email_user = db.query(EmailUser).filter(EmailUser.email == admin_email).first()
+            sub_claim = str(email_user.id) if email_user else admin_email
+        else:
+            sub_claim = admin_email
         jwt_jti = str(uuid.uuid4())
         now = datetime.now(timezone.utc)
         payload = {
@@ -4214,13 +4216,13 @@ async def admin_ui(
 
         # Set HTTP-only cookie using centralized security cookie utility
         set_auth_cookie(response, token, remember_me=False)
-        LOGGER.debug(f"Set session JWT token cookie for user: {admin_email}")
+        LOGGER.debug("Set session JWT token cookie for user: %s", admin_email)
 
         # Capture user/session identifiers for HMAC-bound CSRF cookie
         csrf_user_id = sub_claim
         csrf_session_id = jwt_jti
     except Exception as e:
-        LOGGER.warning(f"Failed to set JWT token cookie for user {user}: {e}")
+        LOGGER.error("Failed to set JWT token cookie for user %s: %s — CSRF protection degraded (falling back to random token)", get_user_email(user), e)
 
     cookie_action = ui_visibility_config.get("cookie_action")
     if cookie_action:
@@ -4915,9 +4917,6 @@ async def _admin_logout(request: Request) -> Response:
     clear_auth_cookie(response)
 
     # Clear CSRF token cookie
-    # First-Party
-    from mcpgateway.services.csrf_service import clear_csrf_cookie
-
     clear_csrf_cookie(response, settings)
 
     use_secure = (settings.environment == "production") or settings.secure_cookies
--- a/mcpgateway/admin.py
+++ b/mcpgateway/admin.py
@@ -1655,18 +1655,29 @@ def _request_origin_matches(request: Request) -> bool:
     return False
 
 
-def _set_admin_csrf_cookie(request: Request, response: Response) -> str:
+def _set_admin_csrf_cookie(request: Request, response: Response, user_id: str | None = None, session_id: str | None = None) -> str:
     """Set or refresh admin CSRF cookie and return token value.
+
+    When user_id and session_id are both provided, generates an HMAC-bound
+    CSRF token that the global CSRFMiddleware can validate. Otherwise falls
+    back to a random token (suitable for anonymous/unauthenticated pages).
 
     Args:
         request: Incoming request used for existing token and path scoping.
         response: Outgoing response where the cookie will be written.
+        user_id: User identifier for HMAC-bound token generation.
+        session_id: Session identifier for HMAC-bound token generation.
 
     Returns:
         CSRF token value stored in the response cookie.
     """
-    existing_token = request.cookies.get(ADMIN_CSRF_COOKIE_NAME)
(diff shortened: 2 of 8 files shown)


ja8zyjits:
HI @Nayana-R-Gowda 

Please add tests

msureshkumar88 (changes requested):
Thanks for tackling this — the root cause analysis in the PR description is clear and the approach is correct. A few issues need addressing before this is merge-ready.

## 🔴 Blocking

### 1. JWT failure silently re-introduces the original bug (`admin.py:L4155–4223`)

When `create_jwt_token` or any step inside the outer `try` block raises, `csrf_user_id`/`csrf_session_id` stay `None`. The call at L4250 then falls back to a **random token that `CSRFMiddleware` cannot validate** — exactly the broken state from #5325, now silently triggered by any transient DB or JWT error. The catch-all `except Exception` must either raise HTTP 500 or log at `ERROR` level (not `WARNING`) and clearly document that CSRF protection is degraded.

### 2. Empty-string args bypass HMAC path (`admin.py:L1674`)

```python
if user_id and session_id:
```

If `user_id` is `""` the condition is `False` and the code silently falls back to an unvalidatable random token with no error. Use:

```python
if user_id is not None and session_id is not None:
```

### 3. Missing regression test for the reported bug itself

None of the new `test_admin_ui_*` tests set `email_auth_enabled=False` — the exact scenario that was broken (#5325, basic-auth default deployment). A test must call `admin_ui` with `email_auth_enabled=False`, assert `get_csrf_service().generate_csrf_token` was called, and verify the CSRF cookie is HMAC-bound (not a plain random token). Without this, the fix can regress silently.

### 4. No direct unit test for `_set_admin_csrf_cookie` HMAC path

The function is only exercised indirectly via `admin_ui`. Add:
- One test with `user_id="u1", session_id="s1"` asserting `generate_csrf_token` is called
- One test with `user_id=""` documenting the empty-string fallback behaviour (ties to finding #2)
- One test patching `create_jwt_token` to raise, asserting `admin_ui` returns 200 but records the degraded CSRF path

---

## 🟡 Risk

**`admin.py:L4223` — user dict logged in warning:**
```python
LOGGER.warning(f"Failed to set JWT token cookie for user {user}: {e}")
```
`{user}` may contain permission fields and token metadata. Replace with `get_user_email(user)`.

**`admin.py:L4158` — non-dict user always gets `is_admin=True` in minted JWT:**
```python
bool(user.get("is_admin") if isinstance(user, dict) else True)
```
Previously gated on `email_auth_enabled`; now runs for all auth providers. Change non-dict branch to `getattr(user, "is_admin", False)` to fail closed.

**`admin.py:L1675` — deferred import on every authenticated `GET /admin`:**
```python
from mcpgateway.services.csrf_service import get_csrf_service
```
Move to module top-level (or document the circular-import reason if it must stay local). An `ImportError` here is currently unhandled.

---

## 🔵 Nit

- `admin.py:L4196` — `db.query(EmailUser).filter(...).first()` now runs on every `GET /admin` for all deployments, including basic-auth-only where `EmailUser` is always empty. Short-circuit or cache.
- `admin.py:L4217` — `LOGGER.debug(f"...")` evaluates the f-string even when debug is disabled. Use `LOGGER.debug("Set session JWT token cookie for user: %s", admin_email)`.
- Commit headline is the branch name with a leading space. Prefer conventional commit format: `fix(admin): issue HMAC-bound CSRF token for all auth providers (closes #5325)`.

---

## ✅ Verified correct

- Issue reference: `closes #5325` ✅
- No unrelated changes ✅
- No breaking change (`_set_admin_csrf_cookie` signature is backward-compatible) ✅
- Root cause analysis accurate ✅
- HMAC binding approach consistent with `CSRFMiddleware` validation algorithm ✅

Nayana-R-Gowda:
Blocking 1 | Upgraded LOGGER.warning → LOGGER.error with clear CSRF degradation message; used get_user_email(user) instead of raw user
-- | --
Blocking 2 | if user_id and session_id: → if user_id is not None and session_id is not None:
Risk 2 | Non-dict user is_admin now fails closed: getattr(user, "is_admin", False) instead of True
Risk 3 | Moved from mcpgateway.services.csrf_service import get_csrf_service, clear_csrf_cookie to module top-level (was deferred imports at L1675 and L4919)
Nit 1 | Short-circuited EmailUser DB query when email_auth_enabled=False
Nit 2 | F-string → lazy %s formatting in debug log

msureshkumar88 (changes requested):
Thanks for addressing all the round-1 blocking/risk/nit items — verified each one against the latest diff and they're all resolved (empty-string bypass fix, `LOGGER.error` degradation message, `getattr(user, "is_admin", False)` fail-closed, top-level import, and solid new unit coverage for `_set_admin_csrf_cookie`).

I also traced the actual token round-trip through `csrf_service.py` → `csrf_middleware.py` → `auth.py` to make sure the fix validates end-to-end (not just in the mocked unit tests), and it does check out: `AuthContextMiddleware` doesn't set `request.state.jti` for interactive/local sessions, so `CSRFMiddleware` falls through to re-deriving `user_id`/`session_id` from the same JWT `admin.py` mints — matching `sub_claim`/`jti` correctly. Good fix.

A few things worth resolving before merge:

## 🔴 Blocking

**Missing merge-conflict resolution** — GitHub currently reports this PR as `CONFLICTING` with `main`. Needs a rebase before it can land.

**No end-to-end test proves the CSRF round-trip actually works** — all new tests mock `get_csrf_service`/`create_jwt_token`, so they verify `admin.py` *calls* the right functions, but nothing exercises the real `CSRFMiddleware.dispatch()` validating a token minted by `admin_ui`. Given how non-obvious that round trip is (see above — it depends on `auth.py` *not* setting `request.state.jti` for local sessions, which isn't documented anywhere), a `TestClient` test that does `GET /admin` with `email_auth_enabled=False`, extracts the real CSRF cookie, and POSTs it to a mutating admin endpoint expecting non-403 would be the actual regression guard for #5325. Without it, a future change to `auth.py`'s `jti` handling could silently re-break this fix and nothing in this PR's test suite would catch it.

## 🟡 Risk

**Unrelated formatting churn should be dropped from this PR** — several files have no connection to the CSRF fix and look like incidental `black`/`isort` output from the "pre commit fix" commits:
- `mcpgateway/middleware/auth_middleware.py` (1-line whitespace-only change)
- `mcpgateway/services/gateway_service.py` (multiple hunks collapsing multi-line calls + a stray blank line)
- `mcpgateway/services/metrics.py` (import reorder)
- `mcpgateway/tools/builder/build_hooks.py` (import-comment grouping)
- `tests/unit/mcpgateway/test_admin_module.py` (reformats pre-existing, unrelated tests)

Worth reverting these hunks so the diff only reflects the CSRF fix — makes it easier to review and to bisect later. (Not flagging `.secrets.baseline` — that's expected churn.)

**JWT is now re-minted on every `GET /admin` load, for every auth provider** — previously this was gated behind `email_auth_enabled`, so basic-auth deployments (the default) paid zero cost here. Now every admin dashboard load signs a fresh JWT (plus a JWT verify for auth-provider resolution, and a DB query when email auth is enabled). Given `/admin` is a hot path, consider reusing the existing `jwt_token` cookie's `sub`/`jti` when it's still valid and not near expiry, rather than minting a new one on every page load.

## 🔵 Nit

First commit's headline is still the branch name with a leading space (` bugfix/5325-render-prompt-csrf-token-invalid`). Worth a conventional-commit message on squash/merge, e.g. `fix(admin): issue HMAC-bound CSRF token for all auth providers (closes #5325)`.

Solid fix for a real bug affecting the default deployment mode — just want the conflict resolved and the round-trip covered by a real test before merge.

msureshkumar88:
Follow-up to my review — want to correct/sharpen one point about the missing e2e test.

I initially pointed at `tests/playwright/security/test_session_csrf_security.py::test_logout_clears_session_cookie` as an existing round-trip test that touches this area. That's not quite right, and it's worth spelling out why, because it changes what the missing regression test actually needs to cover.

`/admin/*` routes are guarded by `enforce_admin_csrf` (admin.py), which does a plain double-submit check — `secrets.compare_digest(header_token, cookie_token)`. It never calls `csrf_service.validate_csrf_token()`. So the HMAC-binding this PR adds doesn't affect `/admin/*` mutations at all (they'd have worked the same with the old random-token fallback, as long as the cookie value round-tripped).

The route that actually exercises the HMAC path is `POST /prompts/{prompt_id}` (main.py, `get_prompt`) — not under `/admin`, so it's protected by the *global* `CSRFMiddleware`, which does call `csrf_service.validate_csrf_token(csrf_token, user_id, session_id)`. That's confirmed directly by the #5325 issue body — the repro is clicking **Render Prompt**, which POSTs to `/prompts/{id}`.

So the missing test I flagged should target `POST /prompts/{id}`, not `/admin/logout`. Concretely, reusing the pattern already in `tests/unit/mcpgateway/test_admin_logout_token_revocation.py` (TestClient + `main_app_with_admin_api` + dependency overrides + hand-signed JWT):

1. `monkeypatch.setattr(settings, "email_auth_enabled", False)` — the exact broken config from the issue.
2. `GET /admin` as the basic-auth platform admin, capture the `jwt_token` + `mcpgateway_csrf_token` cookies from the response.
3. Seed/mock a prompt so `POST /prompts/{id}` doesn't 404 before CSRF is checked.
4. `POST /prompts/{id}` with those cookies and `X-CSRF-Token: <csrf_cookie_value>` — assert **not 403**. This is the request that 403'd before the fix and should pass after it.

That's the test that actually proves the fix, and it's still missing from this PR.

ja8zyjits:
Closed in favour of https://github.com/IBM/mcp-context-forge/pull/5497

PR #5360, opened 2026-06-24
fix(a2a-echo): drop broken wget healthcheck, add SSE streaming and v1 agent card

jonpspri:
## 📌 Summary

The Rust echo agent's compose healthcheck shelled out to `wget` against `/health`, but the minimal `mcpgateway/a2a-echo-agent` image does not ship `wget`. The check always failed and the container reported unhealthy. This PR removes the broken healthcheck and lands the in-progress echo-agent work that had been queued behind that fix.

## 🐞 Root Cause

`docker-compose.yml` defined the healthcheck as:

```yaml
healthcheck:
  test: ["CMD", "wget", "-qO-", "http://localhost:9100/health"]
```

`wget` is not present in the Rust binary's runtime image, so `docker exec ... wget ...` exits non-zero on every poll. Docker marks the container `unhealthy`, which blocks any orchestration step that waits for `a2a_echo_agent` health (e.g. `depends_on: condition: service_healthy`).

## 💡 Fix Description

**Healthcheck**
- Remove the `wget`-based healthcheck from `a2a_echo_agent`. The container reverts to plain "running" gating; a `curl`-free, Rust-native health probe could be added in a follow-up if a stronger gate is needed.

**Echo agent (Rust)**
- Add SSE streaming dispatch for `SendStreamingMessage` / `message/stream` and `SubscribeToTask` / `tasks/resubscribe`. Adds `futures` (workspace) to `Cargo.toml`.
- Split agent card rendering into `agent_card_v1` and `agent_card_legacy` so the agent advertises the right shape for v1 vs legacy AgentCard clients.
- Test directive `stream:chunks=N,delay_ms=M` lets integration tests drive specific stream shapes without a separate test harness.

**Compose wiring**
- New `a2a_echo_agent_v0_3_0` sibling service on port 9101 (same image, `A2A_ECHO_PROTOCOL_VERSION=0.3.0`) so cross-version integration tests can target both 1.0.0 and 0.3.0 endpoints from one stack.
- Expose `A2A_ECHO_PUBLIC_URL` so the agent advertises the right externally-visible URL, and `A2A_ECHO_LOG_LEVEL` for runtime log control.
- Add `PASSWORD_CHANGE_ENFORCEMENT_ENABLED` compose default (`:-true`) to keep `/admin/*` REST operational in the testing stack.

## 📏 Reviewability

- [x] This PR has one clear purpose (drop the broken healthcheck and land the echo-agent work it was blocking)
- [x] Unrelated work tracked elsewhere
- [x] Tests included with the code they validate (Rust unit tests in `main.rs`)

## 🧪 Verification

- `make testing-down testing-up` brings the stack up cleanly.
- `docker compose ps` reports both `a2a_echo_agent` and `a2a_echo_agent_v0_3_0` as `Up (healthy)` (healthy = container running, no broken probe failing it).
- `curl http://localhost:9100/health` and `curl http://localhost:9101/health` both return `200`.
- Rust unit tests in `a2a-echo-agent` compile and pass under the standard workspace `cargo test`.

| Check                             | Command                                   | Status |
|-----------------------------------|-------------------------------------------|--------|
| Stack health                      | `make testing-down testing-up; docker compose ps` | ✅ |
| Echo agent v1.0.0 reachable       | `curl http://localhost:9100/health`       | ✅ |
| Echo agent v0.3.0 reachable       | `curl http://localhost:9101/health`       | ✅ |
| Rust unit tests                   | `cargo test -p a2a-echo-agent`            | _Pending CI_ |
| Lint suite                        | `make lint`                               | _Pending CI_ |

## 🏷️ Type of Change

- [x] Bug fix
- [x] Feature / Enhancement (streaming dispatch + v1 agent card)

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-06-24T05:06:13Z",
+  "generated_at": "2026-06-24T13:23:17Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -390,7 +390,7 @@
         "hashed_secret": "d3ac7a4ef1a838b4134f2f6e7f3c0d249d74b674",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 6051,
+        "line_number": 6052,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -398,7 +398,7 @@
         "hashed_secret": "5932862bcd24dd27d0dc0407ec94fe9d6ea24aeb",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 6548,
+        "line_number": 6549,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -406,7 +406,7 @@
         "hashed_secret": "c77c805e32f173e4321ee9187de9c29cb3804513",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 6560,
+        "line_number": 6561,
         "type": "Secret Keyword",
         "verified_result": null
       },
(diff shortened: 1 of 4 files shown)


lucarlig (dismissed):
LGTM

ja8zyjits (approved):
LGTM 
Process followed

Issue #5326, opened 2026-06-19
[API]: Add POST /v1/mcp-servers/test endpoint for MCP Server connection testing


marekdano:
## Summary

  The React UI cannot call `/admin/**` endpoints — those are reserved for the legacy HTMX admin interface. A dedicated v1 REST endpoint is required so the React UI can test MCP server connectivity (needed for [#5040] Test Connection dialog).

  ## What to add

  A new router file `mcpgateway/routers/mcp_servers_router.py`:

  ```py
  router = APIRouter(prefix="/v1/mcp-servers", tags=["MCP Servers"])
  ```

A `POST /v1/mcp-servers/test` route that replicates the full logic of `admin_test_gateway` (`mcpgateway/admin.py:14298`):

  - DNS-pinning / anti-rebinding via `SecurityValidator.validate_gateway_test_url`
  - SSRF allowlist enforcement (`gateway_test_allow_registered_only` / `gateway_test_allowed_hosts`)
  - Auto-injection of OAuth / basic / bearer / header auth from matching `DbGateway` record
  - Structured logging on success and failure
  - Same `GatewayTestRequest` / `GatewayTestResponse` schemas (`mcpgateway/schemas.py:4604`)
  - Permission: `GATEWAYS_READ` (mirrors the existing `/admin/gateways/test` rule at `token_scoping.py:177`)

  ## Files to change

  | File | Change |
  |---|---|
  | `mcpgateway/routers/mcp_servers_router.py` | New file — router with `POST /v1/mcp-servers/test` |
  | `mcpgateway/main.py` | Register the new router via `app.include_router(...)` |
  | `mcpgateway/middleware/token_scoping.py` | Add `("POST", re.compile(r"^/v1/mcp-servers/test(?:$|/)"), Permissions.GATEWAYS_READ)` |
  | `client/src/components/servers/TestConnectionDialog.tsx:117` | Change call from `/admin/gateways/test` → `/v1/mcp-servers/test` |

  ## References

  - Implementation to mirror: `mcpgateway/admin.py:14298` (`admin_test_gateway`)
  - Schemas: `mcpgateway/schemas.py:4604` (`GatewayTestRequest`, `GatewayTestResponse`)
  - Token scoping: `mcpgateway/middleware/token_scoping.py:177`
  - Router pattern to follow: `mcpgateway/routers/a2a_agent_plugin_bindings.py`
  - UI call site: `client/src/components/servers/TestConnectionDialog.tsx:117`
[...]

Issue #5040, opened 2026-06-03
[UI-REWRITE]: Test Connection dialog for MCP servers


a-effort:
Sub issue of https://github.com/IBM/mcp-context-forge/issues/4093
### Summary

Build the Test Connection dialog for MCP servers in the React UI. Open from the row actions menu, send a configurable HTTP request to the server, and render the response inline. [Design](https://www.figma.com/design/H2Dr9QPvRVxCe92V8URvGT/Refiring?m=auto&node-id=3468-15163&t=bBjSZABWqdRexmfA-1)

### Fields

- Method (GET/POST/PUT/DELETE/PATCH)
- Path
- Headers (JSON)
- Content-Type (`application/json` or `application/x-www-form-urlencoded`)
- Body (matches the selected Content-Type)

Headers and Body show JSON example placeholders.

### Acceptance Criteria

Scenario: Open the dialog
  Given I am on the MCP Servers page
  When I click "Test Connection" in a server's action menu
  Then the dialog opens with the server's base URL pre-filled

Scenario: Successful test renders the response
  Given the upstream returns a 2xx response
  When I submit the form
  Then I see a message success with status code and latency

Scenario: Failed test renders the failure
  Given the upstream returns a non-2xx response or the request fails
  When I submit the form
  Then I see a failure banner with status message and latency / error details from the API

### Reuse the legacy backend

Submit to `POST /admin/gateways/test` so OAuth credential injection, SSRF allowlist enforcement, and timeout behavior match the legacy admin.

### References

- Backend handler: `mcpgateway/admin.py:14025` (`admin_test_gateway`)
- Legacy modal: `mcpgateway/templates/admin.html:9800-9904`
- Legacy submit/render: `mcpgateway/admin_ui/gateways.js:1490-1725`

### Related

Error message copy improvements is tracked separately: https://github.com/IBM/mcp-context-forge/issues/5041

a-effort:
Question to explore: can we pre-fill the header and body options for the user to test with? @marekdano @gcgoncalves @vishu-bh

marekdano:
Done!

PR #5364, opened 2026-06-24
feat(ui-rewrite): Add Test Connection dialog for MCP server

marekdano:
Relates #5040

## Summary

Replaces the placeholder "Connection Test Result" confirm dialog on the Servers page with a dedicated **Test Connection** dialog that lets users compose and send a test HTTP request to an MCP server and inspect the response. The layout follows the Figma design (two-column: request form on the left, response panel on the right).

Screenshot 2026-06-24 at 12 35 01
Screenshot 2026-06-24 at 12 34 36

## Changes

**New component — `client/src/components/servers/TestConnectionDialog.tsx`**
  - Two-column dialog: request form (left) + response panel (right, always visible).
  - Request form: URL (required), Method (Get/Post/Put/Delete/Patch as a segmented radio group), Path, Content type, Headers, and Body (shown only for non-GET methods).
  - Client-side JSON validation for the Headers and Body fields before sending (matching the existing `useToolForm` `JSON.parse` convention).
  - Response panel states: idle placeholder, running, success (`CircleCheck` + status/latency + syntax-highlighted body via `JsonHighlighter` with a copy button), and error.
  - Resizable Headers/Body textareas; field backgrounds are transparent so they inherit the dialog's `popover` background; focus ring matches the `MCPServerForm` textareas.
  - Accessibility: on open, focus moves to the dialog title (`tabIndex={-1}`) rather than auto-focusing the pre-filled URL field, keeping focus inside the dialog and announced by screen readers.

**`client/src/pages/Servers.tsx`**
  - Wires the actions-menu "Test connection" item to open `TestConnectionDialog` (passes `serverName`/`serverUrl`).
  - Removes the old `serversApi.testConnection` call, its page-level inline error/result state, and the repurposed `ConfirmDialog`.

**Tests**
  - `TestConnectionDialog.test.tsx` (new): initial render, all form fields, method radios, Body hidden-for-GET / shown-for-POST, Headers and Body JSON validation, focus-not-on-URL behavior, and close.
  - `Servers.test.tsx`: updated the obsolete "inline error on test failure" case to "opens the test connection dialog from the actions menu", reflecting the new dialog-based flow.

  ## Notes / follow-up

  - The dialog does **not** yet perform a live request. The React UI must not call the legacy `/admin/**` routes, and the v1 endpoint isn't available yet, so the action currently surfaces a "pending API endpoint" message. A
  `TODO(#5326)` marks where to wire in `POST /v1/mcp-servers/test`. Tracked by #5326.
  - Once that endpoint lands, the success/HTTP-response rendering and copy button become reachable, and the response-driven unit tests should be restored.

### DIFF
--- a/client/src/components/servers/TestConnectionDialog.tsx
+++ b/client/src/components/servers/TestConnectionDialog.tsx
@@ -1,0 +1,371 @@
+import { useCallback, useEffect, useMemo, useRef, useState } from "react";
+import { CircleCheck, CircleAlert, Copy, Info, Loader2, TestTubeDiagonal } from "lucide-react";
+import {
+  Dialog,
+  DialogContent,
+  DialogDescription,
+  DialogFooter,
+  DialogHeader,
+  DialogTitle,
+} from "../ui/dialog";
+import { Button } from "../ui/button";
+import { Input } from "../ui/input";
+import { Label } from "../ui/label";
+import { Select, SelectContent, SelectItem, SelectTrigger, SelectValue } from "../ui/select";
+import { Textarea } from "../ui/textarea";
+import { JsonHighlighter } from "../ui/json-highlighter";
+import { copyToClipboard } from "@/components/gateways/utils";
+import { cn } from "@/lib/utils";
+
+interface TestConnectionDialogProps {
+  open: boolean;
+  onOpenChange: (open: boolean) => void;
+  serverName: string;
+  serverUrl: string;
+}
+
+type TestStatus = "idle" | "testing" | "success" | "error";
+
+interface TestResponse {
+  status_code: number;
+  latency_ms: number;
+  body?: string | Record<string, unknown>;
+}
+
+const HTTP_METHODS = ["Get", "Post", "Put", "Delete", "Patch"] as const;
+
+function FieldLabel({
+  htmlFor,
+  children,
+  required,
+  hint,
+}: {
+  htmlFor?: string;
+  children: React.ReactNode;
+  required?: boolean;
+  hint?: string;
+}) {
+  return (
+    <Label htmlFor={htmlFor} className="flex items-center gap-1 text-sm font-medium">
+      <span>
+        {children}
+        {required && <span className="ml-0.5 text-destructive">*</span>}
+      </span>
+      {hint && (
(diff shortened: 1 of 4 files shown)


client/src/components/servers/TestConnectionDialog.tsx
```diff
@@ -0,0 +1,371 @@
...
+            {/* Method */}
+            <div className="space-y-2">
+              <FieldLabel>Method</FieldLabel>
+              <div
+                role="radiogroup"
+                aria-label="Method"
+                className="flex w-full gap-1 rounded-md bg-muted p-1"
+              >
+                {HTTP_METHODS.map((m) => (
+                  <button
```
vishu-bh: I think this should be `radio-group.tsx` instead direct radio button

client/src/components/servers/TestConnectionDialog.tsx
```diff
@@ -0,0 +1,371 @@
...
+            Send a test request to the MCP server and inspect the response.
+          </DialogDescription>
+        </DialogHeader>
+
+        <div className="grid gap-6 py-2 md:grid-cols-2">
+          {/* Left column — request form */}
+          <div className="space-y-4">
+            {/* URL */}
+            <div className="space-y-2">
+              <FieldLabel htmlFor="url" required hint="The full URL of the MCP server to test.">
```
vishu-bh, line 204: The URL label is required, but `handleTest` never checks blank or invalid URL before returning the pending endpoint message.

vishu-bh (changes requested):
Noticed some minor nitpicks, Rest PR is clean and looks good.

client/src/components/servers/TestConnectionDialog.tsx
marekdano: Good catch!

client/src/components/servers/TestConnectionDialog.tsx
marekdano: Added validation!

vishu-bh (commented):
Thanks @marekdano, changes look clean, once conflict is resolved I can approve it.

vishu-bh (approved):
LGTM 🚀

Issue #2288, opened 2026-01-21
🔧 Epic: TypeScript SDK Auto-Generation (@hey-api/openapi-ts)


crivetimihai:
# 🔧 Epic: TypeScript SDK Auto-Generation (@hey-api/openapi-ts)

## Goal

Automatically generate a fully-typed TypeScript client SDK from the OpenAPI specification using `@hey-api/openapi-ts`. This SDK will be used by the React-based ContextForge Desktop app and can be published for external integrations.

## Why Now?

1. **Desktop App Prerequisite**: ContextForge Desktop (React/Electron) needs a typed API client
2. **Type Safety**: Auto-generated types ensure API contract compliance
3. **Developer Experience**: Typed SDK with autocomplete improves productivity
4. **Consistency**: Single source of truth from OpenAPI spec
5. **External Use**: SDK can be published to npm for third-party integrations

---

## 📖 User Stories

<details>
<summary><strong>US-1: Generate TypeScript Client from OpenAPI</strong></summary>

**As a** developer  
**I want** a TypeScript SDK auto-generated from our OpenAPI spec  
**So that** I have type-safe API client code

**Acceptance Criteria:**
```bash
npx @hey-api/openapi-ts -i openapi.json -o ./contextforge-client-ts
```
- All API endpoints represented as typed methods
- Request/response types generated from schemas
- Enums and union types preserved
- SDK regenerates on spec changes
- Zero manual type definitions

</details>

<details>
<summary><strong>US-2: SDK Package Structure</strong></summary>

**As a** developer  
**I want** the SDK packaged as a proper npm module  
**So that** I can import it in any TypeScript project

**Acceptance Criteria:**
- Package name: `@contextforge/client` or `contextforge-client-ts`
- Proper `package.json` with types
- ESM and CommonJS builds
- Tree-shakeable exports
- Minimal dependencies

</details>

<details>
<summary><strong>US-3: Authentication Support</strong></summary>

**As a** developer  
**I want** the SDK to handle authentication  
**So that** I don't manage tokens manually

**Acceptance Criteria:**
- Support Bearer token authentication
- Support cookie-based auth (for browser)
- Configurable auth interceptor
- Token refresh handling
- Clear auth error types

</details>

<details>
<summary><strong>US-4: Error Handling</strong></summary>

**As a** developer  
**I want** typed error responses  
**So that** I can handle errors appropriately

**Acceptance Criteria:**
[...]

PR #5373, opened 2026-06-24
feat(ui-rewrite): Generate types with orval

### LINKED ISSUE #5378: [CHORE]: Use Orval for auto-generating TS types
### 🔧 Chore Summary

Replace custom types with auto-generated ones on the react client.

---

### 🧱 Area Affected

- [ ] GitHub Actions / CI Pipelines
- [ ] Pre-commit hooks / linters
- [X] Type-checking (typescript)
- [ ] Build system or `Makefile`
- [ ] Docs or spellcheck
- [ ] Other:

---

### ⚙️ Context / Rationale

This changes keeps our code consistent with the API changes and guarantees type safety even if the API changes.

---

### 📋 Acceptance Criteria

- [ ] Types can be generated with a make command
- [ ] React client still compiles after code generation
- [ ] There are no regression errors introduced

gcgoncalves:
# Pull Request

## 🔗 Related Issue

Closes #5378

---

## 📝 Summary

Adds Orval for automatic type generation. Then include it in the CI, and apply the generated types on a single page to make sure it compiles and the code still works.

---

## 🏷️ Type of Change

- [ ] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [X] Chore (CI, tooling, type-checking)
- [ ] Other (describe below)

### DIFF
--- a/client/src/pages/Resources.tsx
+++ b/client/src/pages/Resources.tsx
@@ -2,8 +2,7 @@ import { useState, useMemo, memo } from "react";
 import { useIntl } from "react-intl";
 import { Plus, MoreHorizontal, FileText } from "lucide-react";
 import { useQuery } from "@/hooks/useQuery";
-import type { Resource } from "@/types/resource";
-import type { MCPServer, ServersResponse } from "@/types/server";
+import type { ResourceRead, GatewayRead, CursorPaginatedGatewaysResponse } from "@/generated/types";
 import { Button } from "@/components/ui/button";
 import { Card, CardContent, CardHeader } from "@/components/ui/card";
 import {
@@ -35,8 +34,8 @@ const ResourceCard = memo(function ResourceCard({
   resource,
   onViewResource,
 }: {
-  resource: Resource;
-  onViewResource: (resource: Resource) => void;
+  resource: NonNullable<ResourceRead>;
+  onViewResource: (resource: NonNullable<ResourceRead>) => void;
 }) {
   const intl = useIntl();
 
@@ -156,10 +155,10 @@ function AddResourcesCard({ onAddResource }: { onAddResource: () => void }) {
 export function Resources() {
   const intl = useIntl();
   const [showForm, setShowForm] = useState(false);
-  const [selectedResource, setSelectedResource] = useState<Resource | null>(null);
+  const [selectedResource, setSelectedResource] = useState<NonNullable<ResourceRead> | null>(null);
 
-  const { data, isLoading, error, refetch } = useQuery<Resource[]>("/resources?limit=0");
-  const { data: gatewaysData } = useQuery<ServersResponse>(
+  const { data, isLoading, error, refetch } = useQuery<ResourceRead[]>("/resources?limit=0");
+  const { data: gatewaysData } = useQuery<CursorPaginatedGatewaysResponse>(
     "/gateways?limit=0&include_pagination=true",
     {
       enabled: !isLoading,
@@ -167,7 +166,9 @@ export function Resources() {
   );
 
   const gatewayNameById = useMemo(() => {
-    const gateways: MCPServer[] = gatewaysData?.gateways ?? [];
+    const gateways: NonNullable<GatewayRead>[] = (gatewaysData?.gateways ?? []).filter(
+      (g): g is NonNullable<GatewayRead> => g !== null,
+    );
     return new Map(
       gateways.map((gateway) => [
         gateway.id,
@@ -181,7 +182,7 @@ export function Resources() {
     await refetch();
   };
 
-  const handleResourceClick = (resource: Resource) => {
+  const handleResourceClick = (resource: NonNullable<ResourceRead>) => {
     setSelectedResource(resource);
   };
 
@@ -237,13 +238,15 @@ export function Resources() {
           {!isLoading && (
             <div className="grid grid-cols-1 gap-6 lg:grid-cols-2 2xl:grid-cols-3">
               <AddResourcesCard onAddResource={() => setShowForm(true)} />
-              {data?.map((resource) => (
-                <ResourceCard
-                  key={resource.id}
-                  resource={resource}
-                  onViewResource={handleResourceClick}
-                />
-              ))}
+              {data
+                ?.filter((r): r is NonNullable<ResourceRead> => r !== null)
+                .map((resource) => (
+                  <ResourceCard
+                    key={resource.id}
+                    resource={resource}
+                    onViewResource={handleResourceClick}
+                  />
+                ))}
             </div>
           )}
 
--- a/client/src/pages/Resources.tsx
+++ b/client/src/pages/Resources.tsx
@@ -142,15 +142,10 @@ function AddResourcesCard({ onAddResource }: { onAddResource: () => void }) {
 }
 
 /**
- * Resources page component - displays and manages MCP resources
- * Currently implements read-only listing with grouping by gateway
- * Future PRs will add CRUD operations
- *
- * Features:
- * - Groups resources by gateway slug
- * - Shows resource count and active status per group
- * - Details panel for viewing individual resources
- * - Placeholder form for future resource creation
+ * Resources page component - displays and manages MCP resources.
+ * Renders one card per resource (no gateway grouping).
+ * Details panel resolves a human-readable gateway name from gatewayId via gatewayNameById,
+ * falling back to the raw gatewayId when the gateway is not in the fetched list.
  */
 export function Resources() {
   const intl = useIntl();
--- a/mcpgateway/main.py
+++ b/mcpgateway/main.py
@@ -1939,7 +1939,8 @@ def _custom_openapi() -> dict:
         description=app.description,
         routes=app.routes,
     )
-    schema["paths"] = {path: val for path, val in schema["paths"].items() if not path.endswith("/") or path == "/"}
+    all_paths = set(schema["paths"])
+    schema["paths"] = {path: val for path, val in schema["paths"].items() if not path.endswith("/") or path == "/" or path.rstrip("/") not in all_paths}
     app.openapi_schema = schema
     return schema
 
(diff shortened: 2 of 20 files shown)


marekdano (commented):
@gcgoncalves - Thank you for this PR! Having Orval in the app is a big benefit 

Some findings:

#### 1. Incorrect action version comments (minor)

  ```yaml
  uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5
  uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6
  ```

`actions/checkout` is currently at v4 and `actions/setup-python` is at v5. The `# v5`/`# v6` comments are misleading — the pinned hash is authoritative, but wrong version comments make security audits harder. Worth correcting.

  ---

  #### 2. `.secrets.baseline` exclusion regex path mismatch (minor bug)

  ```diff
  -    "files": "(?x)(...)|^.secrets.baseline$",
  +    "files": "(?x)(...)|client/generated/.*.ts|^.secrets.baseline$",
  ```

The actual generated files live at `client/src/generated/types/`, not `client/generated/`. The pattern `client/generated/.*.ts` won't match them. Since these files are gitignored and detect-secrets won't scan them in CI anyway this is low-impact — but the baseline should be accurate. Consider:

  ```
  client/src/generated/.*\.ts
  ```

  ---

  #### 3. `as unknown as Resource` in tests (code smell)

  In `Resources.test.tsx`:

  ```ts
  } as unknown as Resource;
  ```

This double cast means `createMockResource` doesn't satisfy the generated `ResourceRead` shape, which defeats some of the purpose of switching to generated types. Worth fixing the mock to include all required fields with dummy values instead.

  ---

  #### 4. Orval generates a fetch client, not just types

  The config:

  ```ts
  client: "fetch",
  target: "./src/generated/types/",
  ```

This generates both types *and* full fetch-based API functions. The PR only consumes the types. If the team doesn't plan to use the generated client, a types-only config is cleaner:

  ```ts
  output: {
    schemas: { path: "./src/generated/types/" },
    // omit target and client
  }
  ```

Keeping the fetch client for future adoption is a reasonable bet — just worth a conscious decision either way.

  ---

  #### 5. `MCPGATEWAY_ADMIN_API_ENABLED` gate on `js-build` is a silent skip

  ```makefile
  @if [ "$(MCPGATEWAY_ADMIN_API_ENABLED)" != "true" ]; then \
      echo "⏭️  MCPGATEWAY_ADMIN_API_ENABLED != true — skipping JS build"; \
  ```

The default in `config.py` is `false`, and `.env.example` sets it to `true`. A developer who runs `make serve` without copying `.env.example` will silently get no JS bundle, with only a `⏭️ skipping` message. This is a behavior change from before. Consider a more prominent warning or documenting it in the Makefile comment.

  ---

  #### 6. No convenience Makefile target for end-to-end type generation

Developers need to know to run `make fetch-openapi` then `npm run generate` separately. A chained target like `make generate-client` that does both (with a note that a running gateway is required) would help discoverability.

  ---

  #### 7. `gatewayId` used as gateway discriminator in `resources.spec.ts`

  ```diff
  -const resource1 = { ...MOCK_RESOURCE, gatewaySlug: "gateway-a", name: "Resource A" };
  +const resource1 = { ...MOCK_RESOURCE, gatewayId: "gateway-a", name: "Resource A" };
  ```

This test verifies resources from different gateways display as separate cards. The discriminator changed from `gatewaySlug` to `gatewayId`, but real `gatewayId` values are opaque UUIDs — the test now uses a slug-style string `"gateway-a"` as an ID. This is fine for mocking, but if `Resources.tsx` uses `gatewayId` to look up gateway names via `gatewayNameById`, the grouping logic should be verified to still work correctly end-to-end.

gcgoncalves:
Hi, @marekdano, thanks for the thorough review. For point 4, omitting the `client` will not avoid generating the the hooks/api functions,  it actually defaults to `fetch`. Setting the `schemas` attribute to `schemas: { path: "./src/generated/types/" },` is what prevents the api functions, so no changes required on this one. 

All the other items are addressed.

One more thing: in the future, it might be worth replacing our custom api hooks with orval generated ones, but that's to be studied and considered after at a cleanup step.

mcpgateway/main.py
```diff
@@ -1928,6 +1929,24 @@ async def setup_passthrough_headers():
...
+    """Generate OpenAPI schema with trailing-slash duplicate paths removed."""
+    if app.openapi_schema:
+        return app.openapi_schema
+    schema = get_openapi(
+        title=app.title,
+        version=app.version,
+        description=app.description,
+        routes=app.routes,
+    )
+    schema["paths"] = {path: val for path, val in schema["paths"].items() if not path.endswith("/") or path == "/"}
```
vishu-bh: this currently removes every path ending in `/`, but not all of those are duplicate aliases. For example, routes mounted as router-level `/` handlers, such as `/teams/` and the plugin-binding collection endpoints, can be the only OpenAPI path for that operation. Filtering them unconditionally means the runtime endpoint still exists, but generated clients will miss it.

Could we only drop a trailing-slash path when the corresponding non-slash path also exists and represents the duplicate operation? That would keep the operation-id cleanup without hiding legitimate APIs from /openapi.json

client/src/pages/Resources.tsx
```diff
@@ -2,8 +2,7 @@ import { useState, useMemo, memo } from "react";
 import { useIntl } from "react-intl";
 import { Plus, MoreHorizontal, FileText } from "lucide-react";
 import { useQuery } from "@/hooks/useQuery";
-import type { Resource } from "@/types/resource";
-import type { MCPServer, ServersResponse } from "@/types/server";
+import type { ResourceRead, GatewayRead, CursorPaginatedGatewaysResponse } from "@/generated/types";
```
vishu-bh, line 5: I can see some failures when running `npm run build`. I think it is because `@/generated/types` is part of git ignore, May be we update build script `"build": "npm run generate && tsc -b && vite build"` or should be manually generated first and then do a `npm run build`

vishu-bh (changes requested):
some things to check

client/src/pages/Resources.tsx
gcgoncalves: That's a great call. 👍

mcpgateway/main.py
gcgoncalves: FastAPI's `redirect_slashes=True` default means every router-level `/` handler gets registered as two routes in the OpenAPI schema (the trailing-slash handler and a non-slash redirect sibling) so the non-slash twin is always present. The filter safely drops `/teams/` while keeping `/teams` for every case in this codebase. No change needed here.

marekdano: I've got this as a blocking issue 

```python
  schema["paths"] = {path: val for path, val in schema["paths"].items()
                     if not path.endswith("/") or path == "/"}
  ```

  This assumes every trailing-slash path has a non-slash twin. That's true for the collection routers in `main.py` (stacked `@router.get("")` **and** `@router.get("/")`) and the `utility_router` MCP endpoints (stacked `/rpc` + `/rpc/`, etc.). It is **not** true for router-level `/` handlers that have no `""` variant.

This was raised by @vishu-bh, the author replied that `redirect_slashes=True` registers a non-slash sibling in the schema so the filter is safe. **That's empirically incorrect.** `redirect_slashes` handles redirects at request time — it adds no `APIRoute`, so nothing appears in `app.routes`/OpenAPI. Building the raw schema from the live app confirms:

  ```
  '/teams'  in schema: False      '/teams/'  in schema: True

  slash-only paths (trailing / present, non-slash absent):
    /admin/
    /teams/
    /v1/a2a-agents/plugin-bindings/
    /v1/tools/plugin_bindings/
  ```

  The filter removes all four operations from the published `/openapi.json` and Swagger UI. It's a live-runtime schema change, not just codegen — it compiles today only because Orval derives model types from `components/schemas` (unaffected); the missing paths bite the moment anyone generates a client or relies on the docs.

  **Fix** — only drop a trailing-slash path when its non-slash equivalent actually exists:

  ```python
  paths = schema["paths"]
  schema["paths"] = {
      p: v for p, v in paths.items()
      if not (p.endswith("/") and p != "/" and p.rstrip("/") in paths)
  }
  ```

This still strips the true aliases (`/servers/`, `/tools/`, `/rpc/`, the `/_internal/mcp/*/` set) while preserving the four slash-only operations. Add a regression assertion (e.g. `/teams/` present in the generated schema) to lock it in.

marekdano (commented):
I have three suggestions

 - **`build` now hard-couples the JS build to a working Python env.** `npm run build` → `generate` → `pregenerate` → `make export-schema` imports the FastAPI app. The `test -f openapi.json ||` guard avoids re-export when present, but a fresh `client/` build needs the venv + importable app. @vishu-bh flagged the same on `Resources.tsx`; worth a note in `client/README` for frontend-only contributors.
  - **Codegen ergonomics.** The migration is full of `NonNullable<ResourceRead>` and runtime `.filter((r) => r !== null)`, implying Orval emits schemas as nullable. `Resources.test.tsx` already aliases `type Resource = NonNullable<ResourceRead>`; doing the same in `Resources.tsx` (and revisiting the `Optional` response-model fields that drive the nullability) keeps later page migrations cleaner.
  - **`client: "fetch"`** generates a fetch client (operation functions), but only `schemas` types are consumed. Consider types-only generation to avoid shipping unused client code.

and one comment below

gcgoncalves:
@marekdano Addressed.

vishu-bh (approved):
Thanks for addressing the comments. 

LGTM 🚀

marekdano (approved):
Solid, well-scoped tooling change. Thanks for adding Orval!!!

LGTM 🚀

mrlubos:
Interesting @gcgoncalves. What made you drop @hey-api from https://github.com/IBM/mcp-context-forge/issues/2288?

gcgoncalves:
@mrlubos Orval offers Zod and MSW generation alongside the types, and we're using Zod for validation and MSW on the testing suites of the react client, so it was considered a solid option.

PR #5387, opened 2026-06-25
chore: rename .pre-commit-lite.yaml to .pre-commit-config.yaml

jonpspri:
# Pull Request

## 📝 Summary

Renames `.pre-commit-lite.yaml` to the default name `.pre-commit-config.yaml` so that `pre-commit` auto-detects it on developer machines and runs hooks on `git commit` without needing `--config .pre-commit-lite.yaml`.

Without this rename, developers who simply install hooks (`pre-commit install`) get no checks on commit because `pre-commit` does not find a config under the default name — defeating the purpose of having the hooks at all on dev workflows.

## 🔧 Changes

- Rename `.pre-commit-lite.yaml` → `.pre-commit-config.yaml` (history preserved via `git mv`).
- Update explicit references to the old filename:
  - `Makefile` — drop the now-redundant `--config` flag target value.
  - `.github/codeql-config.yml` — `paths-ignore` entry.
  - `docs/docs/development/doctest-coverage.md` — prose reference.
  - `docs/docs/testing/fuzzing.md` — code-block filename comment.
- `.dockerignore` already referenced `.pre-commit-config.yaml` and needs no change.

## 🏷️ Type of Change

## 🧪 Verification

- `grep -r "pre-commit-lite"` across the repo returns zero matches after the change.
- `pre-commit run --all-files` picks up the renamed file via auto-detection (no `--config` flag required).
- The Makefile target still works because the `--config` value was updated in lockstep with the rename.

## 📓 Notes

Behavior on CI is unchanged — workflows that invoke the Makefile target or reference the file by path resolve to the new name. The win is on developer machines, where `pre-commit` now Just Works after `pre-commit install`.


cafalchio (dismissed):
LGTM

ja8zyjits (dismissed):
LGTM

gcgoncalves (dismissed):
Thanks, my pre-commit went haywire bc of this issue.

Lang-Akshay (changes requested):
Thanks for the PR @jonpspri 
LGTM

msureshkumar88 (dismissed):
Approach correct. Rename to `.pre-commit-config.yaml` is the right fix — `pre-commit` auto-detects the default name, so `pre-commit install` now works without `--config` on developer machines.

Verified:
- No pre-existing `.pre-commit-config.yaml` on main — clean rename, no clobber, history preserved via `git mv`
- All explicit refs updated in lockstep: Makefile `--config` flag, codeql `paths-ignore`, two docs
- CI workflows invoke the Makefile target, not the file path directly — no CI breakage

Note: The prior CHANGES_REQUESTED findings (Containerfile.lite UBI pin, docker-scan.yml permissions) are not present in the current diff. If those changes were dropped, the stale review may need dismissal before merge.

Lang-Akshay (dismissed):

ja8zyjits (approved):
LGTM

Lang-Akshay (approved):
LGTM

PR #5389, opened 2026-06-26
fix(startup): fail-fast on AUTH_ENCRYPTION_SECRET mismatch before gateway init

Deep-Sixed:
## Summary

- Adds `validate_auth_encryption_key()` to the `lifespan()` startup path in `main.py`, running before `gateway_service.initialize()`.
- Probes `gateways.auth_value` and `tools.auth_value` using `decode_auth()`. If `InvalidTag` is raised (wrong key), logs a single actionable FATAL message and calls `sys.exit(1)` — before any worker task has started.
- Adds `scripts/validate_rekey.py`: a standalone post-rotation gate that enforces five invariants (old secret decrypts nothing, new secret decrypts everything, no AES-GCM column holds a Fernet `v2:` blob, no JSON-quoted VARCHAR values, decoded payloads are non-empty dicts). Exit 0 = safe to restart.
- Adds `scripts/rotate-encryption-secret.sh`: wrapper that runs `validate_rekey.py` as a mandatory gate before restarting the container, with a dry-run default and `--restart` flag to apply.

## Problem

ContextForge stores gateway and tool auth credentials in two distinct encryption layers:

| Layer | Module | Key derivation | Format |
|---|---|---|---|
| AES-GCM | `utils/services_auth.py` | `SHA256(secret)` | `base64url(nonce[12] + ciphertext)` |
| Fernet/Argon2id | `services/encryption_service.py` | Argon2id | `v2:{...json...}` |

When an operator rotates `AUTH_ENCRYPTION_SECRET` and the rekey migration conflates these layers, the DB ends up with blobs that pass structural validation but raise `cryptography.exceptions.InvalidTag` at runtime. Without this patch, that error surfaces as a per-gateway 500 from the health worker — noisy, delayed, and not actionable.

## Behavior change

**Before**: wrong `AUTH_ENCRYPTION_SECRET` causes `InvalidTag` on first gateway health check cycle, logged per-gateway, no startup abort.

**After**: wrong `AUTH_ENCRYPTION_SECRET` causes a single FATAL log at startup with explicit repair instructions and `sys.exit(1)` before any worker task starts.

No behavior change when auth data is valid — `validate_auth_encryption_key()` logs one INFO line and returns.

## Non-encrypted values are filtered

The probe skips NULL, JSON null (→ Python `None`), empty dict (`{}`), Fernet `v2:` blobs, JSON-quoted VARCHAR values, and strings shorter than 20 chars. Only real AES-GCM blobs that `decode_auth()` must handle at runtime are tested.

## Closes

N/A — standalone operator-safety improvement.

### DIFF
--- a/mcpgateway/main.py
+++ b/mcpgateway/main.py
@@ -1457,6 +1457,12 @@ async def lifespan(_app: FastAPI) -> AsyncIterator[None]:
         # Validate UAID security configuration
         validate_uaid_security_config()
 
+        # Validate AUTH_ENCRYPTION_SECRET can decrypt existing auth data.
+        # Must run before gateway_service.initialize() schedules worker tasks
+        # that call decode_auth — otherwise InvalidTag surfaces as a noisy
+        # per-gateway error log rather than a single actionable fatal.
+        validate_auth_encryption_key()
+
         # Initialize the plugin manager factory whenever the YAML config is
         # available. We used to gate this on ``settings.plugins.enabled`` but
         # that broke runtime enable-from-disabled: a node that boots with the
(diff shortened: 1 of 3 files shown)


jonpspri:
Question:  This makes sense as a preventitive measure, but is it appropriate for the information to be stored in two layers to begin with?  Should the information instead be stored in a single layer and accessed via internal API as needed?

jonpspri:
Relates to #6047 ([BUG]: import --rekey-secret only re-encrypts auth_value, silently orphaning OAuth secrets and tokens) and #6094 (Developer CLI startup experience — concise, actionable terminal output).

This PR's fail-fast startup validation for `AUTH_ENCRYPTION_SECRET` mismatches is directly complementary to the rekey gap in #6047: detecting the wrong key before any worker starts prevents the gateway from running in a broken state when a secret rotation is incomplete.

jonpspri:
Hi @charlessnydercareer — sincere apologies for leaving this PR open for so long without a proper review or merge. That's on us, and we're sorry for the lack of responsiveness.

Issue #5384, opened 2026-06-25
feat(session-affinity): gate forward-dispatch fix behind a feature flag


gandhipratik203:
**Summary**

Gate the multi-worker session-affinity forward-amplification fix (#4557) behind a feature flag, so each deployment can opt in independently and roll back instantly without a redeploy.

**Acceptance Criteria**

```gherkin
When the flag is disabled (default), behavior matches current main
When the flag is enabled, forwarded requests take the fix path
```

gandhipratik203:
closed by: https://github.com/IBM/mcp-context-forge/pull/5393

PR #5393, opened 2026-06-26
fix: eliminate multi-worker session-affinity forward amplification

gandhipratik203:
## Summary

Rebased, single-commit version of the multi-worker session-affinity forward-amplification fix (#4557), cleanly on top of current `main`. This supersedes the long-running #4981 branch, which had drifted 107 commits behind `main`.

Fixes the #4557 regression where stateful MCP requests (carrying `Mcp-Session-Id`) collapsed from ~180 RPS to single digits on multi-worker deployments, via two fixes:

- **Per-worker `WORKER_ID`** recomputed in the gunicorn `post_fork` hook. Under `--preload`, every worker inherited the master's `{host}:1`, so a forward fanned out to all workers in the container (N× duplicate execution).
- **In-process dispatch** of the affinity-owned forward. The owner now serves the request in its own process against the bound upstream session, instead of a loopback `/rpc` call that the kernel scattered to a random worker.

## Validation

- Rebased onto `main`; conflicts resolved (5 files), no logic changes vs the reviewed #4981 work.
- `make test`: **19664 passed, 605 skipped, 2 xfailed**.
- Re-validated end-to-end on a 16-core / 32 GB host (3 gateway replicas behind nginx): **461.88 RPS @ 0.00% failures** and the #4205 counter reproducers pass **5/5** — details below.

<details>
<summary><b>Load benchmark — Streamable HTTP tools throughput (16-core host, 462 RPS @ 0.00% failures)</b></summary>

Image built from this branch (`make docker-prod`), stack up via `make testing-up`, then
`make benchmark-mcp-tools` — tools-only MCP Streamable HTTP, 125 concurrent users, 30/s
spawn rate, 60s, against the registered `fast_time` virtual server through nginx on `:8080`.

| Metric | Value |
|---|---|
| Total requests | 27,734 |
| Failures | 0 (0.00%) |
| Throughput | 461.88 RPS |
| Avg latency | 200.2 ms |
| p50 / p90 / p95 / p99 | 140 / 320 / 490 / 1,400 ms |
| Max | 3,003 ms |

|---|---|---|---|---|---|
| `tools/call` [rapid] | 26,294 | 0 | 437.9 | 200.7 | 1,400 |
| `tools/list` [rapid] | 1,315 | 0 | 21.9 | 193.0 | 1,200 |
| `initialize` | 125 | 0 | 2.1 | 163.9 | 1,800 |

</details>

<details>
<summary><b>#4205 counter reproducers — statefulness &amp; isolation under multi-worker affinity (5/5 PASS)</b></summary>

A minimal per-session counter MCP server (state keyed by the upstream `ServerSession`) is
registered behind the gateway; a session is incremented N times and must read back N. If
forwarding scattered onto a fresh upstream session, the counter would drop.

**Counter server** (run on the host; a 2nd instance on `:9401` for Test 4):

```python
# counter_server.py — per-session counter, state keyed by the upstream ServerSession
mcp = FastMCP("repro-counter", host="0.0.0.0", port=9400)

@mcp.tool()

@mcp.tool()

```

**Prefer the automated version:** these reproducers are scripted (with a README) on a companion branch — [compare `fix/session-affinity-forward-amplification...test/session-affinity-counter-reproducers`](https://github.com/IBM/mcp-context-forge/compare/fix/session-affinity-forward-amplification...test/session-affinity-counter-reproducers). It brings up the stack, registers the counter, runs Tests 1-5, and prints PASS/FAIL — easier than the manual steps below.

<details><summary><b>Manual counter-server registration (optional — the script does this for you, but here's how to do it by hand)</b></summary>

Run everything from the repo root, with the testing stack already up (`make testing-up`).

**1. Start the counter server on the host (binds `0.0.0.0`):**
```bash
.venv/bin/python tests/e2e/session_affinity/counter_server.py > /tmp/counter9400.log 2>&1 &
# confirm it's listening on 0.0.0.0:
ss -ltn | grep ':9400'
```

**2. Mint an admin token:**
```bash
export JWT_SECRET_KEY=$(grep -E '^JWT_SECRET_KEY=' .env | cut -d= -f2-)
export MCPGATEWAY_BEARER_TOKEN=$(.venv/bin/python -m mcpgateway.utils.create_jwt_token \
    --username admin@example.com --admin --exp 10080 --secret "$JWT_SECRET_KEY" --algo HS256 2>/dev/null)
```

**3. (Optional) Confirm the gateway container can reach the counter** — this is the step that catches the `localhost` mistake:
```bash
docker compose exec gateway python3 -c "import urllib.request; print(urllib.request.urlopen('http://host.docker.internal:9400/mcp', timeout=5).status)"
```
Any HTTP status back (e.g. `406`) = reachable ✅. A connection error = the server isn't on `0.0.0.0` or you'd be using `localhost`.

**4. Register the counter as a gateway** — note `host.docker.internal`, not `localhost`:
```bash
curl -sS -X POST http://localhost:8080/gateways \
  -d '{"name":"counter-manual","url":"http://host.docker.internal:9400/mcp","transport":"STREAMABLEHTTP"}'
```
Look for `"status":"active"` and `"reachable":true` in the response = it connected. ✅

**5. Confirm the tools synced (a few seconds later):**
```bash
sleep 5
curl -sS http://localhost:8080/tools -H "Authorization: Bearer $MCPGATEWAY_BEARER_TOKEN" \
  | python3 -c "import sys,json; print([t['name'] for t in json.load(sys.stdin) if 'counter' in t['name'].lower()])"
```
You should see `counter-manual-increment` and `counter-manual-get-value`.

**Gotchas:**
- `{"message":"Gateway already exists"}` = a gateway is already registered at that URL (unique by URL). List them with `curl -sS http://localhost:8080/gateways -H "Authorization: Bearer $MCPGATEWAY_BEARER_TOKEN"` and delete the one on `:9400` first.
- Before running `run_reproducers.py`, delete this manual gateway — the script registers its own at `:9400` and would hit the same URL conflict. (The counter server keeps running; you're only removing the registration.)

</details>

**Steps:**

1. Run the counter on the host — reachable from the gateway containers via
   `host.docker.internal` (compose maps `extra_hosts: host.docker.internal:host-gateway`).
4. Admin token: `python -m mcpgateway.utils.create_jwt_token --username admin@example.com --admin --secret $JWT_SECRET_KEY`.
5. Drive sessions against `/servers/{id}/mcp`: `initialize` → `notifications/initialized` →
   `tools/call increment` ×N → `tools/call get_value` (concurrent sessions via a thread pool).

| Test | Steps | Result |
|---|---|---|
| 1 — single session | one session, `increment` ×25, then `get_value` | `increments=[1…25]`, `get_value=25` — **PASS** |
| 2 — same-user multi-session | 3 concurrent sessions, one token, ×10 each | 3 distinct sids, each `[1…10]` / `get_value=10` — **PASS** |
| 3 — two tokens | two distinct tokens (same admin), parallel, ×10 | both `[1…10]`, no cross-talk — **PASS** |
| 4 — multi-upstream | one session across two counters (`:9400`+`:9401`): A×5, B×3, A×2 | A `[1,2,3,4,5]` then `[6,7]` (=7), B `[1,2,3]` (=3) — **PASS** |
| 5 — owner-worker kill | bind + ×5, read owner from Redis, `kill -9` the owning worker, hit stale sid, then fresh `initialize` | stale sid → **HTTP 404 / `-32600 "Session not found"` in 30.0s**; killed worker respawns; fresh init `[1,2,3]` — **PASS** |

Test 5 trace:

```
[1] bound sid=fa0fff53…  increments=[1, 2, 3, 4, 5]
[2] pool_owner -> 31d47cea2698:32        # owner pid 32 (not :1 -> per-worker WORKER_ID active)
[4] kill -9 32 -> worker dead (gunicorn respawns)
[6] fresh initialize d0ab68fd… increments=[1, 2, 3]
```

session per downstream session), isolation holds across sessions/tokens/upstreams, and the
owner-death failure mode returns a clean structured error rather than a hang.

</details>

 ([pr-5393-no-regression-logs.zip](https://github.com/user-attachments/files/29615610/pr-5393-no-regression-logs.zip)): RBAC, a 10-min load test, and MCP protocol e2e on a live 3-worker stack, showing this PR matches `main` with no regressions.

The only non-passing results are pre-existing on `main` and independent of this PR (a #4341 admin-visibility test assertion, stateless-mode default-endpoint execution, and load-test harness artifacts).

Relates to #4981, #4557, #5384.

### DIFF
--- a/mcpgateway/auth_context.py
+++ b/mcpgateway/auth_context.py
@@ -223,6 +223,25 @@ def get_internal_mcp_auth_context(request: Request) -> Optional[Dict[str, Any]]:
     return None
 
 
+def encode_internal_mcp_auth_context(auth_context: Dict[str, Any]) -> str:
+    """Encode an edge-validated auth context for forwarding to a trusted internal dispatcher.
+
+    Mirror of :func:`decode_internal_mcp_auth_context`. Packages the auth context
+    so a trusted internal dispatcher (e.g. ``/_internal/mcp/rpc``) can use it
+    without re-running authentication. The unpadded base64url shape keeps the
+    value header-safe.
+
+    Args:
+        auth_context: Auth-context dict to encode.
+
+    Returns:
+        Unpadded base64url-encoded JSON payload for the
+        ``x-contextforge-auth-context`` header.
+    """
+    encoded = base64.urlsafe_b64encode(orjson.dumps(auth_context)).decode("ascii")
+    return encoded.rstrip("=")
+
+
 def decode_internal_mcp_auth_context(header_value: str) -> Dict[str, Any]:
     """Decode the trusted internal MCP auth header payload.
 
@@ -291,6 +310,184 @@ def has_valid_internal_mcp_runtime_auth_header(request: Request) -> bool:
     if not provided:
         return False
     return hmac.compare_digest(provided, _expected_internal_mcp_runtime_auth_header())
+
+
+# Redis-transit integrity for forwarded auth context.
+#
+# The header-borne auth context that reaches /_internal/mcp/rpc is gated by the
+# secret-derived runtime-auth token, so a caller must already hold the secret.
+# The session-affinity Redis pub/sub hop is different: a writer to the Redis
+# channel needs no secret, yet the owner worker re-stamps whatever auth context
+# it reads with a valid runtime-auth token before dispatching. That makes the
+# owner a signing oracle for forged contexts. Signing the auth context alone
+# closes the forged-identity oracle, but it does not prove that the verified
+# identity belongs to the *operation* being executed: an attacker with Redis
+# write access could pair a captured signature with a different method, path,
+# body, or response_channel (CWE-347). To close that, the publisher signs the
+# whole forwarded envelope (every operation field, including response_channel so
+# the response cannot be redirected) and the consumer verifies the envelope
+# before dispatch.
+#
+# These helpers are deliberately separate from the general header-encoding
+# contract (encode/decode_internal_mcp_auth_context), which must stay
+# byte-compatible for the Rust runtime. The forward signature is Redis-only
+# Python transit metadata: it is stripped by the owning worker before the
+# in-process dispatch and never reaches the Rust runtime, so it does not affect
+# the encoded auth-context wire format.
+FORWARD_SIG_FIELD = "forward_sig"
+_REDIS_FORWARD_ENVELOPE_SIG_DOMAIN = b"mcpgw.redis-fwd-envelope.v1"
+_REDIS_FORWARD_ENVELOPE_SIG_DELIMITER = b"\x1f"  # ASCII unit separator; never present in canonical JSON
+
+
+def _redis_forward_envelope_sig_material(envelope: Dict[str, Any]) -> bytes:
+    """Build the domain-separated HMAC input for a Redis-forwarded envelope.
+
+    The signature field itself is excluded, then the remaining envelope is
+    canonicalized with sorted keys so the publisher and the consumer (which has
+    round-tripped the payload through JSON) produce byte-identical material. The
+    domain tag plus a unit separator that cannot occur in the canonical JSON keep
+    the two segments unambiguous.
+
+    Args:
+        envelope: The forwarded Redis payload. The signature field, if present,
+            is ignored so signing and verification see the same input.
+
+    Returns:
+        The byte string fed to HMAC.
+    """
+    without_sig = {k: v for k, v in envelope.items() if k != FORWARD_SIG_FIELD}
+    canonical = orjson.dumps(without_sig, option=orjson.OPT_SORT_KEYS)
+    return _REDIS_FORWARD_ENVELOPE_SIG_DOMAIN + _REDIS_FORWARD_ENVELOPE_SIG_DELIMITER + canonical
+
+
+def sign_redis_forward_envelope(envelope: Dict[str, Any]) -> str:
+    """Sign a full forwarded envelope for transit over the session-affinity Redis hop.
+
+    Keyed with ``auth_encryption_secret`` and bound to every operation field in
+    the envelope (method, path, body, headers, response_channel, auth_context,
+    ...), so a captured signature cannot be replayed under a different operation
+    or to a different response channel.
+
+    Args:
+        envelope: The forwarded Redis payload to protect. Any existing signature
+            field is excluded from the signed material.
+
+    Returns:
+        Hex-encoded HMAC-SHA256 signature.
+    """
+    return hmac.new(
+        _auth_encryption_secret_value().encode("utf-8"),
+        _redis_forward_envelope_sig_material(envelope),
+        hashlib.sha256,
+    ).hexdigest()
+
+
+def verify_redis_forward_envelope(envelope: Dict[str, Any]) -> bool:
+    """Verify a Redis-forwarded envelope signature in constant time.
+
+    Must be called on the untouched received envelope, before any field is
+    decoded or mutated, so the recomputed material matches what was signed.
+
+    Args:
+        envelope: The forwarded Redis payload, including its ``forward_sig``.
+
+    Returns:
+        ``True`` only when the envelope carries a non-empty ``forward_sig`` that
+        is a valid signature for the rest of the envelope under the current secret.
+    """
+    signature = envelope.get(FORWARD_SIG_FIELD)
+    if not isinstance(signature, str) or not signature:
+        return False
+    expected = sign_redis_forward_envelope(envelope)
+    return hmac.compare_digest(signature, expected)
+
+
+# Internal-dispatch trust gate, defined once and shared by all callers.
+INTERNAL_MCP_PATH_PREFIX = "/_internal/mcp"
+INTERNAL_A2A_PATH_PREFIX = "/_internal/a2a"
+INTERNAL_RUNTIME_MARKER_HEADER = "x-contextforge-mcp-runtime"
+INTERNAL_AUTH_CONTEXT_HEADER = "x-contextforge-auth-context"
+TRUSTED_INTERNAL_RUNTIME_MARKERS = frozenset({"rust", "affinity"})
+
+
+def _internal_path_requires_auth_context(path: str) -> bool:
+    """Whether an internal route requires an auth context.
+
+    Only ``*/authenticate`` is exempt, since it creates the context; every
+    other internal route must carry one.
+
+    Args:
+        path: The request path to classify.
+
+    Returns:
+        ``False`` for ``*/authenticate``, otherwise ``True``.
+    """
+    return not path.rstrip("/").endswith("/authenticate")
+
+
+def is_trusted_internal_runtime_request(
+    request: Request,
+    *,
+    allowed_prefixes: tuple[str, ...],
+    require_auth_context: bool,
+    path: Optional[str] = None,
+) -> bool:
+    """Return whether a request is a trusted in-process internal-runtime hop.
+
+    The trust boundary is the HMAC header and, when required, the encoded
+    ``x-contextforge-auth-context``. The loopback check is defense in depth,
+    not an independent gate: ProxyHeaders(trusted_hosts="*") lets a direct
+    external caller influence ``request.client.host`` (the replay is hardened
+    separately by stripping forwarded / client-IP headers).
+
+    Args:
+        request: Incoming request to inspect.
+        allowed_prefixes: Internal path prefixes this caller trusts.
+        require_auth_context: Require a non-empty ``x-contextforge-auth-context``.
+        path: Explicit path override for callers that strip a ``root_path``;
+            defaults to ``request.url.path``.
+
+    Returns:
+        ``True`` only when prefix, runtime marker, HMAC, optional auth context,
+        and loopback all hold; otherwise ``False``.
+    """
+    p = path if path is not None else (getattr(getattr(request, "url", None), "path", "") or "")
+    if not any(p == prefix or p.startswith(f"{prefix}/") for prefix in allowed_prefixes):
+        return False
+    if request.headers.get(INTERNAL_RUNTIME_MARKER_HEADER) not in TRUSTED_INTERNAL_RUNTIME_MARKERS:
+        return False
+    if not has_valid_internal_mcp_runtime_auth_header(request):
+        return False
+    if require_auth_context and not request.headers.get(INTERNAL_AUTH_CONTEXT_HEADER):
+        return False
+    client_host = getattr(getattr(request, "client", None), "host", None)
+    return client_host in ("127.0.0.1", "::1")
+
+
+def is_trusted_internal_mcp_request(request: Request, *, path: Optional[str] = None) -> bool:
+    """MCP + A2A internal trust gate with a path-aware auth-context requirement.
+
+    ``*/authenticate`` routes are exempt from the auth-context requirement;
+    every other internal route requires it. ``/_internal/a2a/*`` is trusted
+    only when the A2A feature is enabled.
+
+    Args:
+        request: Incoming request to inspect.
+        path: Explicit path override (e.g. a ``root_path``-stripped path);
+            defaults to ``request.url.path``.
+
+    Returns:
+        ``True`` when the request is a trusted internal MCP/A2A hop.
+    """
+    p = path if path is not None else (getattr(getattr(request, "url", None), "path", "") or "")
+    if p.startswith(f"{INTERNAL_A2A_PATH_PREFIX}/") and not settings.mcpgateway_a2a_enabled:
+        return False
+    return is_trusted_internal_runtime_request(
+        request,
+        allowed_prefixes=(INTERNAL_MCP_PATH_PREFIX, INTERNAL_A2A_PATH_PREFIX),
+        require_auth_context=_internal_path_requires_auth_context(p),
+        path=p,
+    )
 
 
 def get_token_teams_from_request(request: Request) -> Optional[List[str]]:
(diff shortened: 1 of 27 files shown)


Lang-Akshay (changes requested):
Thanks for the PR @gandhipratik203 
Please address the following findings. 
<h2 data-pm-slice="1 4 []">Blockers</h2><h3>1. Redis <code class="rounded bg-muted px-1.5 py-0.5 font-mono text-sm">auth_context</code> — no integrity protection (<code class="rounded bg-muted px-1.5 py-0.5 font-mono text-sm">session_affinity.py:852</code>)</h3><ul class="tight" data-tight="true"><li><p>Redis pub/sub carries unsigned base64url(JSON) auth context</p></li><li><p>Owner worker re-signs it with valid HMAC → signing oracle for forged payloads (e.g. <code class="rounded bg-muted px-1.5 py-0.5 font-mono text-sm">{"is_admin":true}</code>)</p></li><li><p><strong>Fix:</strong> HMAC-sign <code class="rounded bg-muted px-1.5 py-0.5 font-mono text-sm">auth_context</code> with <code class="rounded bg-muted px-1.5 py-0.5 font-mono text-sm">auth_encryption_secret</code> before publish; verify on receive</p></li></ul><h3>2. Missing feature flag</h3><ul class="tight" data-tight="true"><li><p>PR body explicitly defers <code class="rounded bg-muted px-1.5 py-0.5 font-mono text-sm">MCPGATEWAY_SESSION_AFFINITY_INPROCESS_DISPATCH</code> flag</p></li><li><p>AGENTS.md invariant: "high-risk transports must be feature-flagged and disabled by default"</p></li><li><p>Must add flag before merging</p></li></ul><hr><h2>Medium Findings</h2>
Location | Issue
-- | --
auth_context.py:70 | encode_internal_mcp_auth_context — plain base64url, no tamper-evidence; unsafe for Redis transit
auth_context.py:139 | ProxyHeadersMiddleware(trusted_hosts="*") — external callers can forge X-Forwarded-For; loopback check is defense-in-depth only
main.py:724 | _ensure_rpc_permission early-return for is_authenticated=False bypasses Layer 2 RBAC; safety entirely depends on untampered auth_context

gandhipratik203:
Thanks for the thorough review @Lang-Akshay.

**Blocker 1 (Redis `auth_context` signing oracle):** fixed in e3ce35663. The forwarded context is now HMAC-signed with `auth_encryption_secret` (bound to the `mcp_session_id`) before the Redis hop, and the owner verifies it before trusting it, failing closed with a 403 and no dispatch on a missing or invalid signature. The encoded payload format is unchanged, so the Rust runtime stays compatible.

**Blocker 2 (feature flag):** `MCPGATEWAY_SESSION_AFFINITY_INPROCESS_DISPATCH` was a stale idea and has been removed from the description. Gating uses the existing `MCPGATEWAY_SESSION_AFFINITY_ENABLED` flag, which defaults to off, so the feature is disabled by default.

**Medium (`auth_context.py` encode unsigned for Redis transit):** covered by the signing above; the signature rides as a sibling field, so the encoded payload stays byte-compatible for Rust.

Looking into the remaining medium findings and will follow up here.

gandhipratik203:
Follow-up on the mediums:

- **Public-only RBAC early return:** addressed in 7d554209a. Kept the early return; `_build_internal_mcp_forwarded_user` now validates the decoded context (`teams` / `scoped_permissions` must be lists; a public-only context cannot carry teams, admin, or an identity), so it is safe alongside the signing.
- **`ProxyHeaders(trusted_hosts="*")`:** filed as #5421 as a follow-up (defense-in-depth; the internal route is already gated by the HMAC + signed context).

Lang-Akshay (changes requested):
Please address this blocking issue - 
## Public-only forwarded Streamable HTTP requests may 403

In _execute_forwarded_http_request (consumer side):
auth_context_header = request.get("auth_context") or ""
...
if not auth_context_header or not verify_redis_auth_context(...):
    ... reject 403 ...

On the publisher side, when auth_context is None (public-only / MCP_REQUIRE_AUTH=false), auth_context_sig stays None. Redis delivers both as null. Consumer reads empty string → not auth_context_header is True → 403 before dispatch.

This means any MCP_REQUIRE_AUTH=false Streamable HTTP request that lands on a non-owner worker is permanently rejected. Commit a42133835 ("skip RBAC for public-only") addresses the internal_mcp RBAC layer, but the 403 fires before the request reaches that layer. Need to confirm whether the publisher always produces a non-empty auth context for public-only, or whether the consumer check needs a is_public_only_forward branch. Test suite may not cover this path (public-only + multi-worker affinity forward).

gandhipratik203:
Addressed in 5f0a318c9. Confirmed the publisher always produces a non-empty, signed context for public-only, so no consumer-side public-only branch is needed. Tightened `forward_to_owner()` to refuse publishing without a signed context, and added a public-only forwarding regression test.

mcpgateway/services/session_affinity.py
```diff
@@ -909,20 +914,20 @@ async def start_rpc_listener(self) -> None:
...
                                     else:
-                                        logger.warning("Unknown forward type: %s", forward_type)
+                                        logger.warning(f"Unknown forward type: {forward_type}")
                         except Exception as e:
-                            logger.warning("Error processing forwarded request: %s", e)
+                            logger.warning(f"Error processing forwarded request: {e}")
                 finally:
                     await pubsub.unsubscribe(rpc_channel, http_channel)
-                    logger.info("RPC/HTTP listener stopped for worker %s", WORKER_ID)
+                    logger.info(f"RPC/HTTP listener stopped for worker {WORKER_ID}")
```
cafalchio: Although the f string is much nicer. It is slower on logs. So it is better to keep the way it was. (also clean up the diffs)

cafalchio (changes requested):

mcpgateway/main.py
```diff
@@ -332,6 +332,41 @@ def _is_jwt_token(token: str) -> bool:
...
+        raise HTTPException(status_code=400, detail="Invalid trusted MCP auth context: teams must be a list")
+
+    scoped_permissions = auth_context.get("scoped_permissions")
+    if scoped_permissions is not None and not isinstance(scoped_permissions, list):
+        raise HTTPException(status_code=400, detail="Invalid trusted MCP auth context: scoped_permissions must be a list")
+
+    # A public-only (unauthenticated) context must map to exactly public privileges.
+    # The RBAC skip relies on this invariant, so reject any contradictory attributes
+    # rather than letting them ride an unauthenticated dispatch.
+    if auth_context.get("is_authenticated") is False:
```
cafalchio: Is it checking for not True or is it specifically about False? (for example None is not False)

mcpgateway/services/session_affinity.py
gandhipratik203: Reverted the PR-introduced f-string logging back to lazy %-format

Lang-Akshay (changes requested):
Thanks for the updated @gandhipratik203 
Please address following required finding 
### Required Changes

1. **`mcpgateway/utils/internal_http.py:81` + callers** — `post_rpc_in_process` targets the public `/rpc` endpoint without an auth-context header or a loopback `client=` on `ASGITransport`. OAuth and `MCP_REQUIRE_AUTH=false` public-only sessions will 401 on the JSON-RPC Redis-forwarding path (`session_affinity.py:967`). Fix: change `post_rpc_in_process` to target `/_internal/mcp/rpc`, add `client=("127.0.0.1", 0)` to `ASGITransport`, and pass `auth_context` as a parameter — mirrors the pattern in `_execute_forwarded_http_request` exactly.

2. **`mcpgateway/main.py:~343` (`_validate_internal_mcp_auth_context`)** — `is_authenticated` is not type-checked; add `isinstance(..., (bool, type(None)))` before relying on `is False` semantics downstream in `_ensure_rpc_permission`.

gandhipratik203:
Both addressed:

- **Blocker 1 (`post_rpc_in_process` / public `/rpc`):** 6ebad5082 — routes to `/_internal/mcp/rpc` with a required signed `auth_context` (no `/rpc` fallback); the SSE forward signs/verifies the context over Redis, and the `/rpc` identity is built from verified state, not headers.
- **Blocker 2 (`is_authenticated` type check):** c068716b6 — non-bool `is_authenticated` is rejected.

mcpgateway/utils/internal_http.py
```diff
@@ -48,3 +51,62 @@ def internal_loopback_verify() -> bool:
         bool: ``False`` when the loopback URL is HTTPS, ``True`` otherwise.
     """
     return not _is_ssl_enabled()
+
+
+async def post_rpc_in_process(*, content: bytes, headers: dict, timeout: float, auth_context: str) -> httpx.Response:
+    """POST to the trusted-internal ``/_internal/mcp/rpc`` route via an in-process ASGI transport.
+
+    Affinity-owned requests must execute on the worker that actually holds the
```
cafalchio, line 59: please summarize this docstring

mcpgateway/auth_context.py
```diff
@@ -293,6 +312,175 @@ def has_valid_internal_mcp_runtime_auth_header(request: Request) -> bool:
...
+        given ``(session_id, encoded_auth_context)`` pair under the current secret.
+    """
+    if not signature:
+        return False
+    expected = sign_redis_auth_context(session_id, encoded_auth_context)
+    return hmac.compare_digest(signature, expected)
+
+
+# Internal-dispatch trust gate, defined once and shared by all callers.
+INTERNAL_MCP_PATH_PREFIX = "/_internal/mcp"
```
cafalchio, line 406: please move the constants to the top of the file.

Lang-Akshay (changes requested):
Thanks for the updates @gandhipratik203 

Please add following changes to further harden security

- Medium (CWE-347) — HMAC scope too narrow. Current sig binds (session_id, auth_context) only. Attacker with Redis write access can replay valid sig with different path/body/response_channel. Fix: extend HMAC to bind all operation fields.
  - Worker A signs forwarded Redis message:
    `sig = HMAC(auth_encryption_secret, session_id + auth_context)`
  - Worker B (owner) recomputes same HMAC. If it matches → message came from a legitimate worker who knew the secret → safe to trust identity and dispatch.
  - Problem found: sig only covers identity fields, not the operation (path/body/response_channel) — so attacker can reuse a valid sig with a different operation.

gandhipratik203:
Addressed in `cb60d32dc`. The Redis forward signature now covers the whole envelope (identity + operation + `response_channel`), not just `(session_id, auth_context)`, and the owner verifies it before dispatch. So a captured signature can't be replayed with a different operation, and signing `response_channel` blocks response redirection.

gandhipratik203:
### MCP tools benchmark — standard testing stack (3 gateway replicas, session affinity on)

Ran `make benchmark-mcp-tools` on a fresh Ubuntu 24.04 VM (16 vCPU / 32 GB) via the standard flow (`make docker-prod` → `make testing-up` → `make benchmark-mcp-tools`), against the auto-registered `fast_time` virtual server behind nginx → 3 gateway workers with `MCPGATEWAY_SESSION_AFFINITY_ENABLED=true`. Tools-only MCP Streamable HTTP, 125 users, 30/s spawn, 60s.

**367 RPS, 0 failures (0.00%), p50 180 ms / p95 710 ms / p99 1.8 s.**

<details>
<summary>Full benchmark output</summary>

```
==========================================================================================
MCP STREAMABLE HTTP PROTOCOL — RESULTS
==========================================================================================

                                         OVERALL
  --------------------------------------------------------------------------------------
  Total Requests:           22,047
  Total Failures:                0 (0.00%)
  Requests/sec (RPS):       367.23

  Response Times (ms):
    Average:      267.84
    Min:           28.37
    Max:         5042.76
    p50:          180.00
    p90:          460.00
    p95:          710.00
    p99:         1800.00

                                    ENDPOINT BREAKDOWN
  --------------------------------------------------------------------------------------
  Name                                              Reqs    Fails      RPS  Avg(ms)  p99(ms)
  --------------------------------------------------------------------------------------
  MCP tools/call [rapid]                          20,905        0    348.2    269.0   1800.0
  MCP tools/list [rapid]                           1,017        0     16.9    259.2   1700.0
  MCP initialize                                     125        0      2.1    145.9    840.0

==========================================================================================
```
</details>

Lang-Akshay (dismissed):
LGTM 
# MCP Streamable HTTP Protocol Benchmark Comparison

| Metric                                  | Session affinity enabled, stateful sessions enabled | Session affinity disabled, stateful sessions disabled |
| --------------------------------------- | --------------------------------------------------: | ----------------------------------------------------: |
| `MCPGATEWAY_SESSION_AFFINITY_ENABLED` |                                            `true` |                                             `false` |
| `USE_STATEFUL_SESSIONS`               |                                            `true` |                                             `false` |
| Total requests                          |                                              23,096 |                                                 5,905 |
| Total failures                          |                                                   0 |                                                 5,905 |
| Failure rate                            |                                               0.00% |                                               100.00% |
| Requests/sec                            |                                              385.32 |                                                 98.62 |
| Average response time                   |                                           245.52 ms |                                               3.75 ms |
| Minimum response time                   |                                            25.46 ms |                                               1.41 ms |
| Maximum response time                   |                                         6,799.43 ms |                                              44.49 ms |
| p50 response time                       |                                           190.00 ms |                                               3.00 ms |
| p90 response time                       |                                           360.00 ms |                                               5.00 ms |
| p95 response time                       |                                           480.00 ms |                                               7.00 ms |
| p99 response time                       |                                         1,400.00 ms |                                              19.00 ms |

cafalchio:
Currently every time we call a tool we are getting a new session id. 
using the rust sdk counter example server and calling: get_session_id we receive a new session for every call.

cafalchio (dismissed):
Tested multiple times with regression against main.

Approved

Lang-Akshay (approved):
LGTM

cafalchio (approved):
LGTM

PR #5396, opened 2026-06-26
fix(fast-time-server): build Rust container from workspace root, complete Go→Rust replacement

jonpspri:
The Rust fast-time-server container build was broken: `cargo build --release` exited 101 with `error: failed to find a workspace root` because the Containerfile copied only the crate-local `Cargo.toml` into the build context, while the crate uses `.workspace = true` inheritance from the root manifest. Every `docker compose build` invocation against this service failed.

## Container build fix

- Rewrite [`mcp-servers/rust/fast-time-server/Containerfile`](https://github.com/IBM/mcp-context-forge/blob/fast-time-server-cleanup/mcp-servers/rust/fast-time-server/Containerfile) to copy the workspace root (`Cargo.toml`, `Cargo.lock`, `crates/`, `a2a-agents/`, `mcp-servers/rust/`) and run `cargo build --release -p fast-time-server`. Build context moves from the crate directory to the repo root. Mirrors the existing `slow-time-server` pattern.
- Update all four compose files (`docker-compose.yml`, `docker-compose-performance.yml`, `docker-compose-debug.yml`, `docker-compose-verbose-logging.yml`) and the crate's local `Makefile` (`REPO_ROOT := $(abspath ../../..)`) to use the new context.
- Delete the orphan `mcp-servers/rust/fast-time-server/Dockerfile`. Identical bug, no consumers in the repo (no Makefile/compose/CI/Helm reference it), inconsistent port (8880 vs the README's documented 9080).

## Go → Rust replacement in compose

The Go fast-time-server source was deleted from the workspace. This PR finishes the swap on the running surface:

- Replace the Go `fast_time_server` service (which pulled `ghcr.io/ibm/fast-time-server:latest`) with a local Rust build using the same Containerfile, in both `docker-compose.yml` and `docker-compose-verbose-logging.yml`. Adds a compose-level healthcheck override since the Containerfile's hardcoded port-9080 healthcheck no longer matches the runtime-bound port.
- Remove `mcp-servers/go/fast-time-server/` (16 files) and its standalone 1717-line documentation page.
- Update stale references in `README.md` (drop the dead `cd mcp-servers/go/fast-time-server` block), `docs/docs/using/servers/index.md` (replace the "Fast Time Server (Go)" sample-server entry with the Rust equivalent), and `.github/workflows/wrapper.yml` (re-target the wrapper E2E path-trigger to the Rust crate path).

## `make testing-up` unblock

Three sibling fixes so the testing stack reaches a usable state:

1. **Disable `ANYIO_CANCEL_DELIVERY_PATCH_ENABLED`** in both testing compose files. The patch in [`mcpgateway/transports/sse_transport.py:89-90`](https://github.com/IBM/mcp-context-forge/blob/fast-time-server-cleanup/mcpgateway/transports/sse_transport.py#L89-L90) does `origin._delivery_iterations = 0`, which raises `AttributeError: 'CancelScope' object has no attribute '_delivery_iterations'` on newer anyio versions that added `__slots__` to `CancelScope`. With the patch on, every authenticated gateway request returns HTTP 500 and the `register_*` jobs can't complete. The flag was already documented as "EXPERIMENTAL — enable only if Layers 1-2 don't fully resolve the issue"; it's now off until the patch is rewritten to track iterations via a `WeakKeyDictionary` (or similar external storage) instead of attribute assignment.
2. **Point `register_fast_time` at the Rust `/mcp` endpoint** instead of the Go-only `/http` path. The Rust binary serves Streamable HTTP under `/mcp` only; `/http` 404s.
3. **Remove the `register_fast_time_sse` service** from `docker-compose.yml`. The Rust binary has no SSE transport (the rmcp crate ships Streamable HTTP only), so the SSE registration can't succeed regardless of configuration.

## Verification

- `docker build -f mcp-servers/rust/fast-time-server/Containerfile .` from the repo root succeeds. Cargo finishes `fast-time-server v0.1.0` in ~20 s; the image is tagged and exported cleanly.
- `docker compose build fast_test_server` and `docker compose build fast_time_server` both produce images via the new compose wiring.
- Runtime smoke-test on the built image: `/health` → 200 `{"status":"healthy"}`, `/version` → 200, `/mcp` sessionless → expected `{"code":-32000,"message":"Invalid session ID"}` per MCP Streamable HTTP spec.
- `make testing-up`: `fast_time_server` reports `Up (healthy)` after the healthcheck override, the gateway returns 200 on authenticated `/ready` (anyio fix verified in container env), and the new image tag (`mcpgateway/fast-time-server:latest`) is produced on first run.

## Known follow-up (out of scope for this PR)

While verifying `make testing-up` end-to-end, a separate pre-existing gateway bug surfaced after the anyio fix: `AttributeError: '_IncludedRouter' object has no attribute 'path'`, originating somewhere in the Starlette routing chain wrapped by [`mcpgateway/middleware/client_disconnect.py`](https://github.com/IBM/mcp-context-forge/blob/fast-time-server-cleanup/mcpgateway/middleware/client_disconnect.py). It causes HTTP 500 on authenticated `/gateways` requests, which currently blocks `register_fast_time` and `register_fast_test` from completing the full registration flow. This is in Python middleware territory, unrelated to fast-time-server, and worth filing as a separate issue — the container build itself, the Rust binary, the compose wiring, and the anyio fix in this PR are all verified working.

## Commit map

| # | Title | Why |
|---|---|---|
| 1 | `fix(fast-time-server): build container image from Cargo workspace root` | The Containerfile fix |
| 2 | `fix(fast-time-server): build fast_test_server compose services from repo root` | Compose wiring for the existing Rust crate consumer |
| 3 | `chore(fast-time-server): point Makefile container-build at workspace root` | Local invocation parity |
| 4 | `chore(fast-time-server): remove deprecated Go implementation` | 16 source files + 1717-line standalone doc deletion |
| 5 | `chore(fast-time-server): remove stale references to deleted Go implementation` | README, docs/index, wrapper.yml CI trigger |
| 6 | `feat(compose): build fast_time_server from Rust source and unblock testing-up` | Go→Rust compose swap + anyio disable + register `/mcp` + SSE removal + healthcheck |


jonpspri:
Superceded.

jonpspri:
Closed.  Will move servers to new repo.

PR #5401, opened 2026-06-26
feat(ui-rewrite): Add table of teams in the teams page

### LINKED ISSUE #5106: [UI-REWRITE]: List Teams

marekdano:
Closes #5106 

## Summary

Implements the Teams list view for the UI rewrite, adds a reusable shadcn Popover, fully internationalizes the Teams UI (en/es/pt), and fixes a Vite bundle-splitting bug that produced a blank page. Also repairs pre-existing `Gateways.test.tsx` type errors that were breaking `npm run build`.

Screenshot 2026-06-26 at 15 11 50

Screenshot 2026-06-26 at 15 12 11

### Teams list view
  - `pages/Teams.tsx` — replaces the placeholder with a full list view: initial fetch via `useQuery`, error/empty/loaded states, per-page limit selector, and cursor-based "Load More" pagination.
  - `components/teams/TeamsTable.tsx` (new) — table with team icon, name, visibility, member count, formatted created/updated dates, and a per-row actions menu (Edit/Delete). Header has a bottom border to match the Users table.
  - `types/team.ts` (new) — `Team` and `TeamsResponse` types.
  - Description popover: when a team has a description, a `SquareMenu` icon opens a Popover showing it (styled to match the design's elevated gray surface).
  - Loading indicator switched from the branded pulsing `Loading` to the same lightweight spinner as the Users page (fixes a visible flicker).

### Reusable UI component
  - `components/ui/popover.tsx` (new) — shadcn Popover built on the existing `radix-ui` package (no new dependency).

### Internationalization (en-US, es-ES, pt-BR)
  - `i18n/locales/{en-US,es-ES,pt-BR}/teams.json` (new, 23 keys each) wired into each locale `index.ts`. All Teams/TeamsTable strings now use `useIntl`, including an ICU-pluralized item count.

### Build / bug fixes
  - `vite.config.ts` — fixes `Circular chunk: vendor -> vendor-react -> vendor`, which caused `Cannot access 'g' before initialization` (blank page). Anchored chunk matches to the `/node_modules/<pkg>/` boundary (so `@floating-ui/react-dom` is no longer miscategorized as React) and kept `vendor-react` to the React runtime leaf. Vendor split preserved.
  - `pages/Gateways.test.tsx` — added the now-required `setData` field to `useQuery` mocks, unblocking `tsc` / `npm run build`.

  ### Tests
  - New: `Teams.test.tsx` (19), `TeamsTable.test.tsx` (15), `popover.test.tsx` (7).
  - All affected suites pass; `tsc`, `npm run build`, and ESLint are clean.


client/src/pages/Teams.tsx
```diff
@@ -1,3 +1,180 @@
...
+  // Use useQuery hook for initial data fetching and limit changes
+  const { data: response, error: queryError, isLoading } = useQuery<TeamsResponse>(queryPath);
+
+  // Update teams on initial load
+  useEffect(() => {
+    if (response) {
+      setAllTeams(response.teams);
+      setNextCursor(response.nextCursor ?? null);
+    }
+  }, [response]);
```
gcgoncalves, line 36: Do we need this use effect? we can set the values on the `useQuery`'s `onSuccess` hook:


```suggestion
const { data: response } = useQuery<TeamsResponse>(queryPath, {
  onSuccess: (data) => {
    setAllTeams(data.teams);
    setNextCursor(data.nextCursor ?? null);
  }
});
```

client/src/pages/Teams.tsx
```diff
@@ -1,3 +1,180 @@
...
+    try {
+      const params = new URLSearchParams();
+      params.set("cursor", nextCursor);
+      params.set("limit", limit.toString());
+
+      const result = await api.get<TeamsResponse>(`/teams?${params.toString()}`);
+      setAllTeams((prev) => [...prev, ...result.teams]);
+      setNextCursor(result.nextCursor ?? null);
+    } catch (err) {
+      console.error("Failed to load more teams:", err);
```
gcgoncalves: It might be a good idea to raise this to the user with a `sonner`/toast.

client/src/components/teams/TeamsTable.tsx
```diff
@@ -0,0 +1,194 @@
...
+import { Popover, PopoverContent, PopoverTrigger } from "../ui/popover";
+import type { Team } from "../../types/team";
+import { Loading } from "../ui/loading";
+
+function formatDate(dateString: string, invalidLabel: string): string {
+  const date = new Date(dateString);
+  if (Number.isNaN(date.getTime())) return invalidLabel;
+
+  return date
+    .toLocaleString("sv-SE", {
```
gcgoncalves: Why are we setting the locale to Swedish? Should we set this dinamically?

gcgoncalves (changes requested):
Overall the code is very good, with solid testing and correct patterns throughout. Just a couple things slipped through the cracks. 🙇

client/src/pages/Teams.tsx
marekdano: Good catch on the code smell - looked at whether `onSuccess` on `useQuery` was the right fix, but `Teams` and `Users` have identical cursor-based `"Load More"` pagination that accumulates results across multiple API calls. Deriving the list directly from data (the way `Gateways` does) only works when a single response contains the full set, so `Gateways` can use `useMemo` with no extra state. Here, `allTeams` has to grow across fetches, which means we need a separate state and something to seed it from the initial response.

Adding `onSuccess` to `useQuery` would require a `useRef `to stabilise the inline callback and avoid stale-closure issues, a pattern that doesn't exist anywhere else in the hook or the pages. Keeping useEffect watching response matches exactly how `Users.tsx` handles the same scenario, so I've left that in place. 

We should revisit it later and make sure that we use the same pattern when we fetch data for the tables. (Gateways, Users, Teams, etc.)

client/src/pages/Teams.tsx
marekdano: Good catch!

client/src/components/teams/TeamsTable.tsx
marekdano: Good catch! It was also in `Servers.tsx`. Fixed now.

marekdano:
@gcgoncalves - thank you for reviewing the PR 

I did a bit of refactoring:

- Extract `formatLocalDateTime` into `client/src/utils/formatDate.ts`, replacing three divergent local implementations across UsersTable (pad helper), ServersTable, and TeamsTable (both using the sv-SE locale trick) with a single consistent YYYY-MM-DDTHH:mm:ss local-time formatter.

- Replace silent `console.error` with `toast.error` for load-more failures in the Teams page and cover it with a proper assertion in the test suite.

- Add missing i18n keys: `teams.error.loadMore` (en-US, es-ES, pt-BR) and `mcpServer.neverUsed` (en-US, es-ES, pt-BR). Create the missing `mcpServer.json` locale files for `es-ES` and `pt-BR` and wire them into each locale's `index.ts`.

gcgoncalves (approved):

Issue #5139, opened 2026-06-09
[BUG]: Unable to restrict plugin initialization to single gunicorn worker


Shankar-N-IBM-1:
### 🐞 Bug Summary
context forge starts with multiple gunicorn workers and all plugins start in all workers. Non-hook plugins are initiated in all gunicorn workers leading to repetative functionalities.
---

### 🧩 Affected Component
Select the area of the project impacted:

- [ ] `mcpgateway` - API
- [ ] `mcpgateway` - UI (admin panel)
- [ ] `mcpgateway.wrapper` - stdio wrapper
- [ ] Federation or Transports
- [ ] CLI, Makefiles, or shell scripts
- [ ] Container setup (Docker/Podman/Compose)
- [x] Other (explain below)

---

### 🔁 Steps to Reproduce

1. Create a Non-hook plugin which fetches inventory data of context forge and sends to IBM API Connect Federated API Management.
2. Enable plugins and start application with either *docker* or *make serve*
3. Plugins are initalised in all workers and all workers fetch data and send to IBM API Connect Federated API Management.

---

### 🤔 Expected Behavior
Either an environment variable needs to be introduced and set to identify the worker so that non-hook plugins can access this variable to determine in the plugin to start or stop.
ex:
```bash
worker_id = worker.age
os.environ["GUNICORN_WORKER_ID"] = str(worker_id)
```
---

### 📓 Logs / Error Output
Paste any relevant stack traces or logs here.
⚠️ **Do not paste secrets, credentials, or tokens.**

---

### 🧠 Environment Info
You can retrieve most of this from the `/version` endpoint.

| Key | Value |
|-----|-------|
| Version or commit | `e.g. v0.9.0 or main@a1b2c3d` |
| Runtime | `e.g. Python 3.11, Gunicorn` |
| Platform / OS | `e.g. Ubuntu 22.04, macOS` |
| Container | `e.g. Docker, Podman, none` |

---

### 🧩 Additional Context (optional)
Add any configuration details, flags, or related issues.

gandhipratik203:
Hi @Shankar-N-IBM-1 — this is fixed and merged to `main`. Here's what you need to do to gate your inventory plugin to a single worker:

Add one check at the top of your plugin's `initialize()`:

```python
from mcpgateway.utils.primary_worker import is_primary_worker

async def initialize(self) -> None:
    if not is_primary_worker():
        return
    inventory = await self.fetch_inventory()
[...]

PR #5406, opened 2026-06-28
fix(plugins): restrict non-hook plugin initialization to a single worker

gandhipratik203:
## 📌 Summary

When the gateway runs under multiple worker processes (e.g. gunicorn `workers > 1`), **every** worker loads the plugin manager and runs each plugin's `initialize()`. A non-hook plugin that performs a side effect on startup or in a background task (the reported case: fetch inventory and push it to an external API) therefore runs that work **once per worker**, duplicating it N times.

This adds a small, opt-in helper, `is_primary_worker()`, so a side-effecting plugin can run its work on exactly one worker.

Related issue:  #5139

## 🔁 Reproduction Steps

With `PLUGINS_ENABLED=true` and a non-hook plugin whose `initialize()` has a side effect, start under 2 workers:

```
GUNICORN_WORKERS=2 make serve   # or ./run-gunicorn.sh
```

The side effect runs twice (once per worker). With 1 worker it runs once — i.e. the count tracks worker count.

## 🐞 Root Cause

The FastAPI lifespan runs per worker, and within it the plugin manager initializes every plugin (`mcpgateway/main.py`). The cpex loader calls `await plugin.initialize()` per plugin during load. There is no notion of a "primary" worker, so any startup/background side effect in a plugin executes in every worker.

## 💡 Fix Description

New module `mcpgateway/utils/primary_worker.py` exposing `is_primary_worker()`:

- File-lock based election — the first process to acquire the lock is the primary and holds it for its lifetime; other workers get `False`.
- The lock is released by the OS when the holding process exits, so if the primary worker is recycled (`max_requests`) the next caller takes over.
- The lock file path is scoped by the configured port (override via `MCPGW_PRIMARY_WORKER_LOCK`) so co-located gateway instances don't collide while workers of one instance share a single lock.

Side-effecting non-hook plugins opt in:

```python
from mcpgateway.utils.primary_worker import is_primary_worker

async def initialize(self) -> None:
    if not is_primary_worker():
        return
    ...  # runs on one worker only
```

Design notes:
- **Opt-in / minimal surface.** No changes to `main.py`, gunicorn config, plugin schema, or the hook path. Plugins that don't call the helper behave exactly as before — hook plugins still load and execute on every worker.
- **Per host.** Election is file-lock based, so it deduplicates per host.

## 🧪 Verification

Validated with a throwaway 2-worker harness (a gated non-hook plugin + an ungated hook plugin), not included in this PR:

| Signal | Result |
|--------|--------|
| Gated non-hook plugin side effect (2 workers) | ran **once** (was twice) |
| Hook plugin `initialize()` (2 workers) | ran on **both** workers (unchanged) |
| Hook `http_pre_request` on requests | fired normally |

| Check | Command | Status |
|-------|---------|--------|
| Targeted unit tests | `pytest tests/unit/mcpgateway/utils/test_primary_worker.py` | ✅ 7 passed |
| Lint suite | `make lint` | ⬜ not yet run |
| Full unit tests | `make test` | ⬜ not yet run |
| Coverage ≥ 80% | `make coverage` | ⬜ not yet run |

## 🔁 Manual verification (the reporter's steps)

The reporter's plugin (inventory → IBM API Connect) is stood in for by an
equivalent minimal **non-hook** plugin that appends its worker PID to a file —
same mechanism (a side effect in `initialize()` under multiple workers), but
observable. The reporter's flow (enable plugins, multiple workers, `make serve`)
is otherwise unchanged.

**1. Create the non-hook plugin** `plugins/demo_marker/marker.py`:

```python
import os, time
from cpex.framework import Plugin, PluginConfig
from mcpgateway.utils.primary_worker import is_primary_worker

class DemoMarkerPlugin(Plugin):
    async def initialize(self) -> None:
        # if not is_primary_worker():   # <-- the fix; commented out to show the bug first
        #     return
        with open("/tmp/demo_marker.log", "a") as fh:
            fh.write(f"pid={os.getpid()} t={time.time():.3f}\n")
```

and a config `plugins/demo_marker_config.yaml`:

```yaml
plugin_settings:
  plugin_timeout: 120
  fail_on_plugin_error: false

plugins:
  - name: "DemoMarker"
    kind: "plugins.demo_marker.marker.DemoMarkerPlugin"
    description: "Demo non-hook plugin"
    version: "0.0.1"
    hooks: []          # non-hook plugin
    mode: "permissive"
    priority: 50
    conditions: []
    config: {}
```

**2. Reproduce the bug (gate commented out) — start with 2 workers:**

```bash
rm -f /tmp/demo_marker.log
GUNICORN_WORKERS=2 PLUGINS_ENABLED=true \
  PLUGINS_CONFIG_FILE=plugins/demo_marker_config.yaml make serve
```

Once both workers have booted, in another shell:

```bash
wc -l /tmp/demo_marker.log     # => 2  (side effect ran in every worker)
```

Stop the server (Ctrl-C).

**3. Apply the fix** — uncomment the two `is_primary_worker()` lines, then repeat step 2:

```bash
wc -l /tmp/demo_marker.log     # => 1  (side effect runs on one worker)
```

**Automated equivalent:** `make test-primary-worker-e2e` boots a local 2-worker
gateway with a gated marker plugin and asserts the file has exactly one line.

## 📏 Reviewability

## ✅ Checklist

- [x] Tests added for changes

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -888,6 +888,12 @@ MCPGATEWAY_SKIP_MIGRATIONS=false
 # Leader key name in Redis
 # REDIS_LEADER_KEY=gateway_service_leader
 
+# Override path for the primary-worker election lock file
+# (mcpgateway/utils/primary_worker.py). Lets a side-effecting non-hook plugin run
+# its work on one worker per host. Defaults to a port-scoped file in the system
+# temp dir when unset; point it at a gateway-owned directory on hostile hosts.
+# PRIMARY_WORKER_LOCK_PATH=
+
 # =============================================================================
 
 # =============================================================================
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-07-13T12:57:02Z",
+  "generated_at": "2026-07-13T14:35:14Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-07-13T14:32:55Z",
+  "generated_at": "2026-07-13T12:57:02Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -124,7 +124,7 @@
         "hashed_secret": "14f8aa3e560a47851908ab0f04ec856dbc512d93",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 936,
+        "line_number": 942,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -132,7 +132,7 @@
         "hashed_secret": "fa9beb99e4029ad5a6615399e7bbae21356086b3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1234,
+        "line_number": 1240,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -140,7 +140,7 @@
         "hashed_secret": "7b4455a56fbf1d198e45e04c437488514645a82c",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1260,
+        "line_number": 1266,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -148,7 +148,7 @@
         "hashed_secret": "d08f88df745fa7950b104e4a707a31cfce7b5841",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1268,
+        "line_number": 1274,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -156,7 +156,7 @@
         "hashed_secret": "ac371b6dcce28a86c90d12bc57d946a800eebf17",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1347,
+        "line_number": 1353,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -164,7 +164,7 @@
         "hashed_secret": "0b6ec68df700dec4dcd64babd0eda1edccddace1",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1352,
+        "line_number": 1358,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -172,7 +172,7 @@
         "hashed_secret": "4ad6f0082ee224001beb3ca5c3e81c8ceea5ed86",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1357,
+        "line_number": 1363,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -180,7 +180,7 @@
         "hashed_secret": "cb32747fcfb55eaa194c8cd8e4ba7d49ada08a94",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1363,
+        "line_number": 1369,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -188,7 +188,7 @@
         "hashed_secret": "6c178d51b13520496dbc767ed3d9d7aa5803ac72",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1375,
+        "line_number": 1381,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -196,7 +196,7 @@
         "hashed_secret": "ca45060a53fd8a255d1a83ee8d2f025283ccc66e",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1393,
+        "line_number": 1399,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -204,7 +204,7 @@
         "hashed_secret": "910fbf00f58e9bcb095ea26a75cc1d9a3355e671",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1454,
+        "line_number": 1460,
         "type": "Secret Keyword",
         "verified_result": null
       }
@@ -350,7 +350,7 @@
         "hashed_secret": "319037749ce37e577db0b3628c7f90e333544391",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 976,
+        "line_number": 982,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -358,7 +358,7 @@
         "hashed_secret": "6ae2832e494d1098e8901fe156083e39399a24f1",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 978,
+        "line_number": 984,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -366,7 +366,7 @@
         "hashed_secret": "43fc45734b96bcb1b6cef373e949eb3524ae199b",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1630,
+        "line_number": 1636,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -374,7 +374,7 @@
         "hashed_secret": "9d989e8d27dc9e0ec3389fc855f142c3d40f0c50",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1849,
+        "line_number": 1855,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -382,7 +382,7 @@
         "hashed_secret": "d3ac7a4ef1a838b4134f2f6e7f3c0d249d74b674",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 5922,
+        "line_number": 5928,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -390,7 +390,7 @@
         "hashed_secret": "5932862bcd24dd27d0dc0407ec94fe9d6ea24aeb",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 6419,
+        "line_number": 6425,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -398,7 +398,7 @@
         "hashed_secret": "c77c805e32f173e4321ee9187de9c29cb3804513",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 6431,
+        "line_number": 6437,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -406,7 +406,7 @@
         "hashed_secret": "8fe3df8a68ddd0d4ab2214186cbb8e38ccd0e06a",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 6503,
+        "line_number": 6509,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -414,7 +414,7 @@
         "hashed_secret": "93ac8946882128457cd9e283b30ca851945e6690",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 7585,
+        "line_number": 7591,
         "type": "Secret Keyword",
         "verified_result": null
       }
@@ -2148,7 +2148,7 @@
         "hashed_secret": "fa9beb99e4029ad5a6615399e7bbae21356086b3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1280,
+        "line_number": 1281,
         "type": "Basic Auth Credentials",
         "verified_result": null
       },
@@ -2156,7 +2156,7 @@
         "hashed_secret": "fa9beb99e4029ad5a6615399e7bbae21356086b3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1284,
+        "line_number": 1285,
         "type": "Secret Keyword",
         "verified_result": null
       }
--- a/Makefile
+++ b/Makefile
@@ -882,6 +882,12 @@ test-mcp-plugin-parity: uv  ## MCP plugin parity E2E for current Python or Rust 
 		|| { echo "❌ MCP plugin parity tests failed!"; exit 1; }
 	@echo "✅ MCP plugin parity tests passed!"
 
+test-primary-worker-e2e: uv  ## Primary-worker election E2E: boots a local 2-worker gateway, asserts the gated side effect runs once
+	@echo "🧪 Running primary-worker e2e (boots its own 2-worker gateway)..."
+	@$(UV_BIN) run bash tests/live_gateway/run_primary_worker_e2e.sh \
+		|| { echo "❌ Primary-worker e2e failed!"; exit 1; }
+	@echo "✅ Primary-worker e2e passed!"
+
 test-mcp-session-isolation: uv  ## MCP session/auth isolation tests for the Rust public transport path
 	@echo "🧪 Running MCP session/auth isolation tests against $${MCP_CLI_BASE_URL:-http://localhost:8080}..."
 	@echo "   Requires: docker-compose stack rebuilt in Rust edge/full mode"
--- a/docs/docs/manage/configuration.md
+++ b/docs/docs/manage/configuration.md
@@ -930,6 +930,7 @@ The gateway includes built-in observability features for tracking HTTP requests,
 | `MAX_CONCURRENT_HEALTH_CHECKS` | Max concurrent health checks        | `20`    | int > 0 |
 | `AUTO_REFRESH_SERVERS` | Auto refresh tools/prompts/resources        | `false` | bool    |
 | `FILELOCK_NAME`         | File lock for leader election             | `gateway_service_leader.lock` | string |
+| `PRIMARY_WORKER_LOCK_PATH` | Override path for the primary-worker election lock file (per-host; default is a port-scoped temp file) | (none) | string |
 | `DEFAULT_ROOTS`         | Default root paths for resources          | `[]`    | JSON array |
 
 ### Database Connection Pool
--- a/docs/docs/using/plugins/lifecycle.md
+++ b/docs/docs/using/plugins/lifecycle.md
@@ -339,3 +339,25 @@ make serve
 
 !!! note
         `PLUGINS_ENABLED=true` should be set in your gateway `.env` file.
+
+## Multi-Worker Execution
+
+Under multiple workers (`GUNICORN_WORKERS`), every worker calls each plugin's
+`initialize()`, so **hooks run on every worker by design**. A non-hook plugin
+that does a **side effect** at startup or in a background task would otherwise
+run it once per worker. Gate it with `is_primary_worker()` to run on one worker
+per host:
+
+```python
+from mcpgateway.utils.primary_worker import is_primary_worker
+
+class InventorySync(Plugin):
+    async def initialize(self) -> None:
+        if not is_primary_worker():
+            return
+        ...  # runs on one worker only
+```
+
+One worker is elected via a file lock; the OS frees it on process exit, so a
+follower takes over. Re-check it each cycle for recurring work. The lock is per
+host and its path is overridable with `PRIMARY_WORKER_LOCK_PATH`.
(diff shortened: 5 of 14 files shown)


msureshkumar88 (changes requested):
Good work on the core design — the file-lock election approach is correct, port-scoped paths prevent co-located instance collisions, and the memoized fast-path avoids repeated syscalls. The opt-in pattern (zero changes to `main.py` or plugin schema) is exactly right. A few gaps need resolution before this can merge.

---

## Blocking

### `filelock` not added to `pyproject.toml`

`filelock` is imported directly in `mcpgateway/utils/primary_worker.py` but there's no corresponding addition in `pyproject.toml`. If it's only a transitive dependency today, it can silently disappear when the dep graph changes. Direct imports require explicit direct pins. Add `filelock>=3.12` to `[dependencies]`.

### `is_primary_worker()` catches only `Timeout` — `OSError` propagates uncaught

`FileLock(_lock_path())` and `.acquire()` both raise `OSError` on permission errors, missing parent directory, or a read-only `/tmp`. This exception escapes into `plugin.initialize()`, and depending on `fail_on_plugin_error`, can take down the whole worker. Catch `OSError` alongside `Timeout`; on `OSError` log `WARNING` and return `False` (non-primary, not crashed).

```python
except (Timeout, OSError) as exc:
    logger.warning("primary worker election failed pid=%d err=%s", os.getpid(), exc)
    _is_primary = False
```

### No thread guard on `_lock` / `_is_primary` global mutation

The docstring acknowledges "mutates module state without a thread guard" but Gunicorn sync workers with `--threads > 1` can TOCTOU between `if _lock is None` and `_lock = FileLock(...)`. Two threads can each see `None`, create two `FileLock` objects, and both attempt `.acquire()` — the OS-level lock behaviour at that point is undefined. Add a `threading.Lock()` guard around the `_lock is None` check and the acquire block.

---

## Security / Hardening

### Predictable lock-file path enables local denial-of-service

`/tmp/mcpgw_plugin_primary_{port}.lock` is world-readable and predictable. A local process can pre-create and hold an exclusive lock on this file before the gateway starts, preventing all workers from becoming primary. Plugin side-effects then silently never execute — no error is raised. At minimum, document this limitation in the module docstring. A harder fix: use a gateway-owned directory with restricted permissions instead of world-writable `/tmp`.

### `MCPGW_PRIMARY_WORKER_LOCK` accepts arbitrary paths without validation

An env var pointing to `/etc/crontab.lock` or a network-mounted path redirects the election lock without any warning. The resolved path is never logged, making this invisible in audit trails. At minimum, log the resolved lock path at `INFO` level on first call so operators can detect misconfiguration.

### `MCPGW_PRIMARY_WORKER_LOCK` bypasses the settings system

All other runtime config lives in `mcpgateway/config.py` with type validation and Pydantic coercion. This env var is read raw via `os.environ.get()`, invisible to `make check-env`, undocumented in `.env.example`, and absent from the Key Environment Variables table in AGENTS.md. Add it to `config.py` as `primary_worker_lock_path: str | None = None` with the override logic, and document it in `.env.example` (commented out).

---

## Logging / Observability

### Zero logging in election path

No logger is imported in `primary_worker.py`. Operators cannot tell which PID won election or why a worker is non-primary. Add:

```python
import logging
logger = logging.getLogger(__name__)

# on win:
logger.info("primary worker elected pid=%d lock=%s", os.getpid(), _lock_path())
# on loss:
logger.debug("non-primary worker pid=%d", os.getpid())
```

---

## Testing gaps

### No test for `OSError`/`PermissionError` from `FileLock`

The blocking issue above has no corresponding test. Add a test that patches `FileLock.acquire` to raise `PermissionError` and asserts `is_primary_worker()` returns `False` without propagating.

### No thread-safety test

Given the missing thread guard (blocking issue), add a test that launches N threads calling `is_primary_worker()` concurrently and asserts exactly one returns `True`.

### `make lint` and `make test` not yet run at PR submission

PR body explicitly shows ⬜ for both. The full unit suite and lint must pass before merge.

### Missing e2e scenario: primary worker dies, follower takes over

Unit test `test_follower_retries_and_takes_over` covers module-level retry, but the e2e script only verifies initial election. The highest-value guarantee of this design (OS releases lock on process exit → follower promotes) is not exercised in the live-gateway harness.

---

## Code quality (minor)

`primary_worker.py:L29` — `"FileLock | None"` string annotation is a forward-reference workaround not needed on Python 3.11+. Use `FileLock | None` directly or add `from __future__ import annotations`.

`primary_worker.py` — no `__all__`. Add `__all__ = ["is_primary_worker"]` to make the public surface explicit.

---

## Confirmed correct

- `Closes #5139` ✅
- Not a breaking change — purely additive, opt-in ✅
- Port-scoped lock path prevents co-located instance collisions ✅
- Memoized `_is_primary = True` fast-path is correct ✅
- Unit tests cover first-call wins, idempotent True, contention False, follower takeover ✅
- `.secrets.baseline` change is line-number reindex only — not unrelated ✅
- Necessary change — real production bug ✅

gandhipratik203:
Thanks for the thorough review. Pushed `6bed82d7f` with the fixes:

- **`OSError` escaping** → now caught alongside `Timeout`; fails closed (non-primary) + WARNING instead of crashing `initialize()`. Test added.
- **Thread guard** → `threading.Lock()` around the lazy init/acquire, with in-guard re-check. Concurrency test added.
- **Bypasses settings** → override is now `primary_worker_lock_path` (`PRIMARY_WORKER_LOCK_PATH`) in `config.py` + documented in `.env.example`; no more raw `os.environ`.
- **Logging** → INFO on win (with resolved lock path), DEBUG on loss, WARNING on `OSError`.
- **Minor** → dropped the `"FileLock | None"` string annotation, added `__all__`.

Two notes:
- **`filelock` pin** — already a direct dep (`pyproject.toml:89`, `filelock>=3.29.1`).
- **Thread test** — within one process *all* threads see `True` (the process is primary, not a thread), so "exactly one `True`" doesn't fit; the test asserts the guard creates a single `FileLock` and all callers agree.

Ran locally: `ruff`/`mypy` clean, 9 unit tests + e2e green; full `make lint`/`make test` in CI.

**Tracking as follow-ups** (to keep this PR minimal):
- **Private permissioned lock dir** — documented the limitation + log the resolved path. The dir-hardening guards a local actor who can already write to the temp dir (low severity); operators can already set `PRIMARY_WORKER_LOCK_PATH`. Can do it here if you prefer.
- **e2e follower-takeover** — covered by the `test_follower_retries_and_takes_over` unit test; the live version (kill a worker mid-run) is flake-prone in CI.

msureshkumar88:
Verified the fixes in commit `6bed82d7f` against all items from the previous review.

## Blocking — all resolved ✅

| Item | Status | Evidence |
|------|--------|----------|
| `filelock` missing from `pyproject.toml` | ✅ Already pinned | `filelock>=3.29.1` at line 89 — pre-existing direct dep, not a gap |
| `OSError` propagating uncaught | ✅ Fixed | `except OSError as exc:` now present; fails closed (`_is_primary = False`) + `logger.warning(...)` |
| No thread guard on `_lock`/`_is_primary` | ✅ Fixed | `_guard = threading.Lock()` at module level; `with _guard:` wraps lazy init + acquire; in-guard re-check present |

## Security / Hardening — documented, partial hardening deferred ⚠️

| Item | Status |
|------|--------|
| Predictable `/tmp` lock path → local DoS | ✅ Documented in module docstring ("a local process can pre-acquire the lock…"); `PRIMARY_WORKER_LOCK_PATH` escape hatch exists. Dir-hardening tracked as follow-up — acceptable given the attacker already needs write access to `/tmp`. |
| Arbitrary path via env var, not logged | ✅ Resolved path now logged at `INFO` on every election win |
| `MCPGW_PRIMARY_WORKER_LOCK` bypassing settings | ✅ Fixed — override now lives in `config.py` as `primary_worker_lock_path: Optional[str]` with Pydantic validation |

## Logging / Observability ✅

`INFO` on win (includes resolved lock path), `DEBUG` on loss, `WARNING` on `OSError` — all confirmed in code.

## Minor ✅

- `__all__ = ["is_primary_worker"]` present
- Native `FileLock | None` union type (no string annotation workaround)

## One minor observation

`_lock_path()` is called twice inside the `with _guard:` block — once to construct `FileLock(...)` and once in the `logger.info(...)` call. Functionally correct (path is stable), but assigning to a local var once would be cleaner. Not blocking.

## Testing

9 unit tests + e2e claimed green locally. PR checklist still shows `make black isort pre-commit` unchecked — worth ticking that off or confirming the formatting run passed.

## Summary

All three blocking issues are fixed in code and verified. Security hardening for the `/tmp` DoS vector is documented and deferred — reasonable call. This is ready for approval once the formatting checklist item is confirmed.

msureshkumar88 (dismissed):
## E2E Validation — All checks pass ✅

Ran a local multi-worker e2e against commit `6303e2b38` (branch `repro/5139-plugin-multiworker-init`) before approving. Results below.

---

### Configuration

| Parameter | Value |
|-----------|-------|
| Host | WSL |
| Port | 19876 (isolated, not 8080 which was in use) |
| Auth | `AUTH_REQUIRED=false` (plugin behavior test, not auth) |
| Workers tested | 2, 3 |
| Plugin config | `tests/live_gateway/fixtures/primary_worker_e2e_config.yaml` |
| Plugin under test | `tests.live_gateway.fixtures.primary_marker.PrimaryWorkerE2EMarkerPlugin` |
| Lock | per-test isolated tmpdir (no cross-test pollution) |
| DB | ephemeral SQLite in tmpdir |
| Python | 3.12.3 |

---

### Commands ran

```bash
# TEST 1 — unit suite
pytest tests/unit/mcpgateway/utils/test_primary_worker.py -v --tb=short

# TEST 2 — 2-worker live gateway
PLUGINS_ENABLED=true \
PLUGINS_CONFIG_FILE=tests/live_gateway/fixtures/primary_worker_e2e_config.yaml \
MCPGW_PRIMARY_WORKER_E2E_MARKER=$TMP/marker_2w.log \
PRIMARY_WORKER_LOCK_PATH=$TMP/lock_2w.lock \
DATABASE_URL=sqlite:///$TMP/e2e.db \
HOST=127.0.0.1 PORT=19876 AUTH_REQUIRED=false \
  gunicorn -c gunicorn.config.py --worker-class uvicorn.workers.UvicornWorker \
    --workers 2 --bind 127.0.0.1:19876 mcpgateway.main:app

# TEST 3 — 3-worker live gateway (same config, fresh marker + lock)
# ... same env, --workers 3

# TEST 4 — health check
curl -s -o /dev/null -w "%{http_code}" http://127.0.0.1:19876/health

# TEST 5 — canonical pytest assertion
MCP_CLI_BASE_URL=http://127.0.0.1:19876 \
MCPGW_PRIMARY_WORKER_E2E_MARKER=$TMP/marker_3w.log \
  pytest tests/live_gateway/plugins/test_primary_worker_e2e.py -v --tb=short

# TEST 6 — failover: SIGKILL primary PID, wait 12s, re-check marker + health
kill -KILL <primary_pid>
```

---

### Output

```
TEST 1: Unit tests
  9 passed in 0.35s
  ✅ PASS: Unit suite: 9 tests passed

TEST 2: 2-worker election
  /health ready after 4s | settling 2 workers (8s)
  Marker: pid=1887039 t=1782740273.873
  Lines written: 1  (expected: 1)
  ✅ PASS: 2-worker: 1 of 2 workers wrote to marker — election correct

TEST 3: 3-worker election
  /health ready after 4s | settling 3 workers (8s)
  Marker: pid=1887458 t=1782740286.800
  Lines written: 1  (expected: 1)
  ✅ PASS: 3-worker: 1 of 3 workers wrote to marker — election correct

TEST 4: /health after election
  GET /health → HTTP 200
  ✅ PASS: /health 200 — no worker crashed in initialize()

TEST 5: Canonical pytest assertion
  collected 1 item  .  1 passed in 0.25s
  ✅ PASS: Canonical pytest e2e assertion passed

TEST 6: Primary failover
  Primary PID: 1887458 → SIGKILL
  Marker after 12s (2 lines):
    pid=1887458 t=1782740286.800   ← original primary
    pid=1887790 t=1782740297.355   ← new primary after respawn
  GET /health → HTTP 200
  ✅ PASS: Failover: new primary elected — 2 entries after kill
  ✅ PASS: Post-failover /health still 200 — remaining workers serving

SUMMARY: PASS 7 / FAIL 0
```

---

The fix is verified end-to-end. Core behaviors confirmed:

- **2 workers → exactly 1 side effect** — the bug is fixed
- **3 workers → exactly 1 side effect** — scales correctly beyond 2
- **No worker crashes** — `OSError` + `Timeout` both handled; `initialize()` never propagates
- **Primary failover** — when the elected PID is killed, gunicorn respawns and the new worker wins election and runs the side effect once
- **Regression** — gateway continues serving HTTP 200 throughout

LGTM. Merging is unblocked.

ja8zyjits:
## 🚨 Blocking Issues

- [ ] **Config var `PRIMARY_WORKER_LOCK_PATH` undocumented**
  - File: `docs/docs/manage/configuration.md` (missing in Federation/Leader Election ~L932)
  - Code: `mcpgateway/config.py:2539`, `.env.example:895`
  - Fix: Add table row:
    ```markdown
    | `PRIMARY_WORKER_LOCK_PATH` | Override path for primary-worker election lock file | (none) | string |
    ```

## ⚠️ Warnings

### Documentation
- [ ] **Plugin lifecycle docs missing multi-worker behavior**
  - File: `docs/docs/using/plugins/lifecycle.md`
  - Fix: Add section on `initialize()` per-worker execution + `is_primary_worker()` usage

### Test Coverage Gaps

**Edge Cases:**
- [ ] Lock file parent dir doesn't exist → `test_oserror_missing_parent_directory()`
- [ ] Lock path is directory not file → `test_lock_path_is_directory_fails_closed()`
- [ ] Empty string override path → `test_empty_string_override_falls_back_to_default()`
- [ ] Multi-process contention (not just threads) → use `multiprocessing` module
- [ ] Lock released on process exit → multi-process test where primary exits

**Integration:**
- [ ] E2E doesn't test takeover (kill primary, verify new election)
- [ ] E2E doesn't verify hook plugins run on all workers
- [ ] E2E doesn't stress-test concurrent `initialize()` race handling

### Code Quality
- [ ] Bare exception in test cleanup (`test_primary_worker.py:33`) → log or catch specific
- [ ] FileLock lifetime unclear → add comment "held for process lifetime, OS releases"
- [ ] Test plugin file write lacks error handling (`primary_marker.py:42`)
- [ ] Bash word splitting risk (`run_primary_worker_e2e.sh:45`) → already quoted, low risk

gandhipratik203:
Thanks for the detailed pass. Addressed in `a8f5a92cc` / `bd78d7370`:

- **🚨 `PRIMARY_WORKER_LOCK_PATH`** → added to the configuration table.
- **Plugin lifecycle docs** → new "Multi-Worker Execution" section (per-worker `initialize()` + `is_primary_worker()`).
- **Edge tests** → lock-path-is-a-directory fails closed; empty-string override falls back to default.
- **Bare `except`** → narrowed to `(Timeout, OSError)`; **FileLock lifetime** comment clarified.

Notes: missing-parent-dir doesn't fail closed — `filelock` auto-creates the parent (verified); bash word-splitting already quoted.

**Deferring** as follow-ups:
- **multiprocess/exit + e2e takeover** — real multi-process election is covered by the committed e2e (2 workers → 1 side effect), and SIGKILL-failover was validated manually in @msureshkumar88's TEST 6; an automated kill-mid-run e2e is much flakier for little gain.
- **e2e hooks-on-all-workers / concurrent-init stress** — covered by the thread-safety unit test + manual harness.
- **Test-plugin write handling** — intentional; a failed write in a fixture should surface, not be swallowed.

ja8zyjits:
### Minor Code Issues (Non-Blocking)

1. **Micro-optimization: redundant early return**
   - File: `mcpgateway/utils/primary_worker.py:69-70`
   - Issue: `if _is_primary: return True` before lock → wastes lock acquisition on race
   - Fix: Remove lines 69-70; lock-protected re-check (74-75) sufficient
   - Severity: Low (performance only, not correctness)

2. **Potential FD leak on repeated OSError**
   - File: `mcpgateway/utils/primary_worker.py:77-90`
   - Issue: `FileLock` object persists after OSError, never released
   - Fix: Add `_lock = None` in OSError handler (line 89)
   - Severity: Low (requires pathological repeated failures)

ja8zyjits (dismissed):
LGTM

msureshkumar88 (dismissed):
All findings from the initial review are addressed:

- **filelock dep** — `filelock>=3.29.1` added to `pyproject.toml` directly.
- **OSError handling** — caught alongside `Timeout`, fails closed (non-primary) with WARNING log.
- **Thread guard** — `threading.Lock()` with double-checked locking eliminates the TOCTOU race.
- **Settings integration** — `primary_worker_lock_path` added to `config.py`; no more raw `os.environ` reads.
- **Env/docs** — `.env.example` and `configuration.md` document `PRIMARY_WORKER_LOCK_PATH`.
- **Tests** — OSError test, thread-safety test, and live e2e scenario all added.
- **Minor** — string annotation fixed, `__all__` added, INFO/DEBUG/WARNING logging in place.

Lock-file path predictability is acknowledged in the module docstring with mitigation guidance; dir-hardening is a reasonable follow-up. Code is clean, CI is green.

One note before merge: `mergeStateStatus` is currently DIRTY — a rebase onto main is needed to resolve the conflicts before this can land.

ja8zyjits (dismissed):
LGTM

msureshkumar88 (commented):
Re-reviewed the latest push (`5ae7a59a3`) against all prior rounds.

**All previously requested changes are done:**
- `filelock>=3.29.4` pinned directly in `pyproject.toml`
- `OSError` now caught alongside `Timeout` in `is_primary_worker()` — fails closed (non-primary) with a `WARNING` log instead of propagating into `plugin.initialize()`
- Thread guard added (`threading.Lock()` double-checked locking around the lazy `FileLock` init) — covered by `test_thread_safe_single_lock_creation`
- Settings integration via `primary_worker_lock_path` in `config.py`, documented in `.env.example` and `configuration.md`
- Lock-path predictability limitation documented in the module docstring with mitigation guidance (hardening it further is a reasonable follow-up, as already agreed in an earlier round)
- Solid unit coverage: election, contention, takeover, thread-safety, `OSError`, path override

Code substance looks good and matches what was asked for across all rounds. Holding off on approval for two mechanical reasons on this specific commit, not the design:

1. This looks like a fresh squash/rebase pushed today — all prior approvals (mine and @ja8zyjits's) show as dismissed, and CI shows only the `DCO` check having run against `5ae7a59a3`. The green `Tests & Coverage` / `Rust CI` / `Docker Build` results are from the earlier (now superseded) commit, not this one.
2. `mergeable` is currently `CONFLICTING` — just `.secrets.baseline` timestamp/line-number drift against a recent doc-only commit on `main`, nothing structural. A quick rebase should clear it.

Once it's rebased and the full suite goes green on the new HEAD, happy to approve.

msureshkumar88 (dismissed):
LGMT

msureshkumar88 (dismissed):
LGMT

ja8zyjits (approved):
LGTM

msureshkumar88:
## Post-merge e2e verification (informational — this PR is already merged, no review action taken)

Re-ran a real 2-worker e2e against the current codebase (this fix + the later redis-backend follow-up commits) to confirm `is_primary_worker()` still resolves #5139.

**Config/commands:**
```bash
make test-primary-worker-e2e   # boots real gunicorn --workers 2 + uvicorn worker class, waits for /health, runs the assertion suite
.venv/bin/python -m pytest tests/unit/mcpgateway/utils/test_primary_worker.py -v
```

**Fix-confirmation run (current code):** `2 passed` — gated non-hook plugin's side effect ran exactly once across 2 workers; the ungated hook-plugin regression check also passed (election correctly doesn't suppress hooks).

**Bug-reproduction run** (temporarily commented out the `is_primary_worker()` gate in `tests/live_gateway/fixtures/primary_marker.py`, reverted immediately after):
```
FAILED test_gated_side_effect_runs_on_exactly_one_worker
AssertionError: expected exactly one primary worker to run the side effect, got 2:
  ['pid=504965 t=...', 'pid=504966 t=...']
```
Exact reproduction of the reported bug — two distinct worker PIDs both ran the side effect with the gate removed.

**Unit tests:** `tests/unit/mcpgateway/utils/test_primary_worker.py` — 13/13 passed.

**Surrounding check:** `filelock` is a proper direct `pyproject.toml` dependency (not transitive-only); all current consumers (`main.py` lifespan wiring, `config.py` settings, the redis `leader_election.py` backend added since merge, docs in `lifecycle.md`/`configuration.md`) are consistent and up to date. No regression found from the later redis multi-instance work — the filelock (default) path this PR introduced behaves exactly as originally verified.

**Verdict: fix confirmed still working.** No changes required.

PR #5410, opened 2026-06-29
fix(grpc): enforce SSRF/TLS validation in GrpcEndpoint.start()

### LINKED ISSUE #4615: [CHORE][SECURITY]: GrpcEndpoint.start() should enforce its own SSRF/TLS policy with explicit opt-out
### 🔧 Chore Summary

`GrpcEndpoint.start()` in `mcpgateway/translate_grpc.py` opens a gRPC channel to an arbitrary `target` and (when `reflection_enabled=True`) runs reflection RPCs against it. Today the SSRF/TLS validation lives one layer up in `GrpcService.register_service()` and `GrpcService.invoke_method()` (PR #3202 added these).

That works for the manager-service paths, but **direct callers of `GrpcEndpoint`** — `expose_grpc_via_sse()` in the CLI translator, and any future consumer of the class — can construct an endpoint and call `start()` without going through the validators. The class-internal contract has no SSRF defense.

### ⚙️ Why now

PR #3202 surfaced this during deep review (review item #13 in the PR comment trail). The current behavior is "the manager service is the security boundary"; better practice is "every public method of `GrpcEndpoint` enforces its own policy and the manager service can opt out for trusted local CLI use".

### 📋 Acceptance Criteria

- `GrpcEndpoint.start()` accepts a `trusted_local: bool = False` parameter.
- When `trusted_local=False` (default), `start()` calls `_validate_grpc_target(self._target)` and (if TLS paths are set) `_validate_tls_path()` before opening any channel.
- When `trusted_local=True`, the validators are skipped and the call proceeds with a `logger.warning` once per process.
- `GrpcService.register_service` and `GrpcService.invoke_method` either pass `trusted_local=True` (because they validated already) and gain a comment explaining why, OR drop their pre-call validation since `start()` now handles it. Pick one to avoid double-validating.
- `expose_grpc_via_sse()` and `mcpgateway.translate_grpc` CLI entrypoints either pass `trusted_local=True` explicitly or call the validators themselves.
- Regression test: instantiate `GrpcEndpoint` directly with a known-blocked target (e.g. `metadata.google.internal`) and assert `start()` raises before `grpc.insecure_channel` is called.

### 🔗 Context

- Source: deep-review of PR #3202, item #13 in [PR comment](https://github.com/IBM/mcp-context-forge/pull/3202#issuecomment-4386420357).
- Adjacent: #2663 (broader SSRF prevention epic) — this issue is narrower, focused on the gRPC client.

SaurabhMaydeo:
Closes #4615

SSRF/TLS validation for gRPC connections lived only in the `GrpcService` call sites, so a direct caller of `GrpcEndpoint.start()` opened a channel with no SSRF defense. That left `expose_grpc_via_sse()` and the translate CLI able to be pointed at a blocked destination such as a cloud-metadata endpoint.

`GrpcEndpoint.start()` now validates the target address and any TLS paths against the platform SSRF / path-traversal rules by default, reusing the existing `_validate_grpc_target` / `_validate_tls_path` (imported lazily to respect the existing `grpc_service` ↔ `translate_grpc` circular import). A `trusted_local: bool = False` parameter provides an explicit opt-out for callers that have already validated — it skips the checks and logs a warning once per process. `invoke_method` also validates the target and TLS paths before connecting, and `register_service` is unchanged (it validates at registration and builds its own channel).

Verified: `ruff` and `bandit` are clean on the changed files, and `pytest tests/unit/mcpgateway/test_translate_grpc.py tests/unit/mcpgateway/services/test_grpc_service.py` reports 118 passed — including three regression tests: a blocked target (`metadata.google.internal`, a default `ssrf_blocked_hosts` entry) raises before any channel is opened, `trusted_local=True` bypasses validation as designed, and `start()` validates both TLS paths by default.

### DIFF
--- a/mcpgateway/services/grpc_service.py
+++ b/mcpgateway/services/grpc_service.py
@@ -15,8 +15,6 @@ retrieval, updates, activation toggling, and deletion.
 import asyncio
 import base64
 from datetime import datetime, timezone
-import ipaddress
-from pathlib import Path
 from typing import Any, Dict, List, Optional, Union
 
 try:
@@ -38,6 +36,7 @@ from sqlalchemy import and_, delete, desc, select, update
 from sqlalchemy.orm import Session
 
 # First-Party
+from mcpgateway import translate_grpc
 from mcpgateway.config import settings
 from mcpgateway.db import EmailTeam
 from mcpgateway.db import GrpcService as DbGrpcService
@@ -49,14 +48,13 @@ from mcpgateway.services.logging_service import LoggingService
 from mcpgateway.services.team_management_service import TeamManagementService
 from mcpgateway.utils.create_slug import slugify
 from mcpgateway.utils.display_name import generate_display_name
+from mcpgateway.utils.grpc_validation import GrpcServiceError, _validate_grpc_target, _validate_tls_path
 from mcpgateway.utils.pagination import unified_paginate
 
 # Initialize logging
 logging_service = LoggingService()
 logger = logging_service.get_logger(__name__)
 
-
-_GRPC_DISALLOWED_SCHEMES = ("unix:", "unix-abstract:", "vsock:", "fd:")
 
 # DoS guards for descriptors returned by ``grpc.reflection`` — intentionally hardcoded
 # (not exposed via settings) so a config change cannot silently weaken these limits.
@@ -109,106 +107,6 @@ def _validate_reflected_tool_name(tool_name: str) -> None:
         SecurityValidator.validate_tool_name(tool_name)
     except ValueError as exc:
         raise GrpcServiceError(f"Reflected tool name '{tool_name}' rejected: {exc}") from exc
-
-
-def _validate_grpc_target(target: str) -> None:
-    """Validate a gRPC target address against SSRF-unsafe destinations.
-
-    Consults the platform SSRF settings (``ssrf_allow_localhost``,
-    ``ssrf_allow_private_networks``, ``ssrf_allowed_networks``,
-    ``ssrf_blocked_networks``, ``ssrf_blocked_hosts``) so that gRPC
-    targets follow the same rules as HTTP URLs validated by
-    ``SecurityValidator.validate_url``.
-
-    Args:
-        target: gRPC target string. Accepts ``host:port``, bracketed
-            ``[ipv6]:port``, and gRPC name-resolver forms ``dns:///host:port``,
-            ``ipv4:host:port``, ``ipv6:host:port``. Local-only schemes
-            (``unix:``, ``unix-abstract:``, ``vsock:``, ``fd:``) are always
-            rejected because they bypass the network-level SSRF model.
-
-    Raises:
-        GrpcServiceError: If the target uses a forbidden scheme or resolves to a blocked address.
-    """
-    if not target:
-        raise GrpcServiceError("Empty gRPC target address")
-
-    # Local-only schemes bypass the IP-based SSRF model entirely; reject outright.
-    lowered = target.lower()
-    for scheme in _GRPC_DISALLOWED_SCHEMES:
-        if lowered.startswith(scheme):
-            raise GrpcServiceError(f"gRPC target scheme '{scheme.rstrip(':')}' is not permitted")
-
-    # Strip recognised gRPC name-resolver scheme prefixes so the host check below sees a bare host:port.
-    for scheme_prefix in ("dns:///", "dns://", "dns:", "ipv4:", "ipv6:"):
-        if lowered.startswith(scheme_prefix):
-            target = target[len(scheme_prefix) :]
-            break
-
-    # Extract host (strip port). Bracketed IPv6 literals: ``[::1]:50051``.
-    if target.startswith("["):
-        end = target.find("]")
-        if end < 0:
-            raise GrpcServiceError(f"Malformed bracketed gRPC target: {target!r}")
-        host = target[1:end]
-    else:
-        host = target.rsplit(":", 1)[0].strip("[]")
-    if not host:
-        raise GrpcServiceError("Empty gRPC target address")
-
-    # Reserved / multicast IP literals are unconditionally blocked. SecurityValidator._validate_ssrf
-    # only checks blocked-networks / localhost / private; it does not flag is_reserved/is_multicast,
-    # so this guard runs before delegation to keep the original gRPC-validator semantics. Loopback is
-    # excluded because Python flags ``::1`` as both is_loopback AND is_reserved; loopback policy is
-    # handled by SecurityValidator below via ssrf_allow_localhost.
-    try:
-        addr = ipaddress.ip_address(host)
-    except ValueError:
-        addr = None
-    if addr is not None and not addr.is_loopback and (addr.is_reserved or addr.is_multicast):
-        raise GrpcServiceError(f"gRPC target address '{host}' is blocked (reserved/multicast)")
-
-    # Delegate the hostname/IP-network/DNS-resolution policy to the shared SecurityValidator
-    # so gRPC and HTTP follow the same SSRF rules and a hostname like ``metadata.google.internal``
-    # is resolved before being allowed through.
-    # First-Party
-    from mcpgateway.common.validators import SecurityValidator  # pylint: disable=import-outside-toplevel
-
-    if getattr(settings, "ssrf_protection_enabled", True):
-        try:
-            SecurityValidator._validate_ssrf(host, "gRPC target")  # pylint: disable=protected-access
-        except ValueError as exc:
-            raise GrpcServiceError(str(exc)) from exc
-
-
-def _validate_tls_path(path_str: str, label: str = "TLS path") -> Path:
-    """Validate that a TLS cert/key path is within allowed directories.
-
-    Args:
-        path_str: The file path to validate.
-        label: Label for error messages.
-
-    Returns:
-        Resolved Path object.
-
-    Raises:
-        GrpcServiceError: If the path escapes allowed directories.
-    """
-    resolved = Path(path_str).resolve()
-    # Allow only paths under /certs/, /etc/ssl/, /etc/pki/, or the CWD/certs dir
-    allowed_prefixes = (
-        Path("/certs").resolve(),
-        Path("/etc/ssl").resolve(),
-        Path("/etc/pki").resolve(),
-        Path.cwd().joinpath("certs").resolve(),
-    )
-    if not any(resolved.is_relative_to(prefix) for prefix in allowed_prefixes):
-        raise GrpcServiceError(f"{label} '{path_str}' is outside allowed certificate directories")
-    return resolved
-
-
-class GrpcServiceError(Exception):
-    """Base class for gRPC service-related errors."""
 
 
 class GrpcServiceNotFoundError(GrpcServiceError):
@@ -960,10 +858,6 @@ class GrpcService:
         if not service.enabled:
             raise GrpcServiceError(f"Service '{service.name}' is disabled")
 
-        # Import here to avoid circular dependency
-        # First-Party
-        from mcpgateway.translate_grpc import GrpcEndpoint  # pylint: disable=import-outside-toplevel,cyclic-import
-
         # Parse method name (service.Method format)
         if "." not in method_name:
             raise GrpcServiceError(f"Invalid method name '{method_name}', expected 'service.Method' format")
@@ -986,7 +880,7 @@ class GrpcService:
         stored_descriptors = discovered.get("_file_descriptors", [])
         has_stored_descriptors = bool(stored_descriptors)
 
-        endpoint = GrpcEndpoint(
+        endpoint = translate_grpc.GrpcEndpoint(
             target=service.target,
             reflection_enabled=not has_stored_descriptors,
             tls_enabled=service.tls_enabled,
--- a/mcpgateway/services/grpc_service.py
+++ b/mcpgateway/services/grpc_service.py
@@ -962,7 +962,7 @@ class GrpcService:
 
         # Import here to avoid circular dependency
         # First-Party
-        from mcpgateway.translate_grpc import GrpcEndpoint  # pylint: disable=import-outside-toplevel
+        from mcpgateway.translate_grpc import GrpcEndpoint  # pylint: disable=import-outside-toplevel,cyclic-import
 
         # Parse method name (service.Method format)
         if "." not in method_name:
--- a/mcpgateway/services/grpc_service.py
+++ b/mcpgateway/services/grpc_service.py
@@ -894,7 +894,7 @@ class GrpcService:
         try:
             # Both the asyncio wrapper AND the underlying gRPC call get the deadline so a slow
             # upstream cannot keep an executor thread alive after the coroutine is cancelled.
-            await asyncio.wait_for(endpoint.start(timeout=effective_timeout), timeout=effective_timeout)
+            await asyncio.wait_for(endpoint.start(timeout=effective_timeout, trusted_local=True), timeout=effective_timeout)
 
             if has_stored_descriptors:
                 raw_descriptors = [base64.b64decode(b, validate=True) for b in stored_descriptors]
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-07-09T15:16:14Z",
+  "generated_at": "2026-07-13T09:51:56Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -3328,7 +3328,7 @@
         "hashed_secret": "bf10dae7b89461df3fd3c48f86ec23543710e8cd",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 414,
+        "line_number": 417,
         "type": "Secret Keyword",
         "verified_result": null
       }
@@ -3356,7 +3356,7 @@
         "hashed_secret": "ddb2c93ebcf6098f6ccb08807ed66e02934248a5",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 573,
+        "line_number": 575,
         "type": "Secret Keyword",
         "verified_result": null
       }
--- a/docs/docs/using/grpc-services.md
+++ b/docs/docs/using/grpc-services.md
@@ -105,6 +105,9 @@ python3 -m mcpgateway.translate \
   --port 9000
 ```
 
+!!! note "SSRF validation of gRPC targets"
+    The gRPC target is validated against the platform SSRF settings before any channel is opened. Because `SSRF_ALLOW_LOCALHOST` and `SSRF_ALLOW_PRIVATE_NETWORKS` default to `false`, targets such as `localhost:50051` or RFC1918 addresses are rejected out of the box with a `GrpcServiceError`. To allow a loopback target, set `SSRF_ALLOW_LOCALHOST=true`; to allow a specific private/RFC1918 range, add it to `SSRF_ALLOWED_NETWORKS`. See the [1.0.0-RC3 upgrade notes](../manage/upgrade-to-1.0.0-rc3.md) for the full table of SSRF defaults.
+
 ### 2. Admin UI: Register a gRPC Service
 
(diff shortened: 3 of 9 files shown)


gcgoncalves (dismissed):
Good implementation, clean fix.

gcgoncalves (commented):
Failing lint and tests, please fix them, @SaurabhMaydeo

SaurabhMaydeo:
Thanks @gcgoncalves! I've pushed fixes for the lint and test failures you flagged. Both pass locally for me, but the run looks like it's waiting on your approval to confirm in CI whenever you get a chance. Thanks again!

SaurabhMaydeo:
Quick follow-up: the pytest job was failing on the diff-coverage gate; the new TLS-path validation branch in `start()` wasn't covered by a test. I've pushed one that exercises it, and diff-cover passes locally now. Should be good for another run whenever you get a chance.

SaurabhMaydeo:
Rebased onto main to pull in the FastAPI 0.137 router-test fix (#5447), so the pytest checks should clear on the next run. A few small changes since the earlier review: `invoke_method` keeps its own target/TLS validation (defense in depth on the DB-loaded path), the warn-once helper is now an `lru_cache` instead of a module-level flag, an added test covers `start()`'s default validation, and the two lazy imports carry a `,cyclic-import` disable. Thanks!

mcpgateway/services/grpc_service.py
```diff
@@ -962,7 +962,7 @@ async def invoke_method(
 
         # Import here to avoid circular dependency
         # First-Party
-        from mcpgateway.translate_grpc import GrpcEndpoint  # pylint: disable=import-outside-toplevel
+        from mcpgateway.translate_grpc import GrpcEndpoint  # pylint: disable=import-outside-toplevel,cyclic-import
```
gcgoncalves: Suggestion: extract `GrpcServiceError`, `_validate_grpc_target`, `_validate_tls_path` (plus their small helper constants) into a new module (e.g. `mcpgateway/utils/grpc_validation.py`). No dep on either grpc_service.py or translate_grpc.py.

Then:
- `translate_grpc.py` imports validators from new module (top-level, no lazy needed).
- `grpc_service.py` imports `GrpcServiceError`/validators from new module too (re-export `GrpcServiceError` if tests/other code reference `grpc_service.GrpcServiceError`).
- Remaining dependency tree is one-way only: `grpc_service.py` → `translate_grpc.py` (for `GrpcEndpoint`). No cycle left, that import can go top-level too. Kill both `# Import here to avoid circular dependency` lazy imports.

gcgoncalves (changes requested):

mcpgateway/services/grpc_service.py
SaurabhMaydeo: Done in 662b468, thanks for the suggestion. The validators and `GrpcServiceError` now live in `mcpgateway/utils/grpc_validation.py` with no dependency on either module, both lazy imports are gone, and `grpc_service` re-exports `GrpcServiceError` so existing callers and tests are unaffected. Kept the `GrpcEndpoint` usage as `translate_grpc.GrpcEndpoint(...)` so the existing tests that patch `mcpgateway.translate_grpc.GrpcEndpoint` keep working.

gcgoncalves (dismissed):

msureshkumar88 (changes requested):
Nice fix — closing the gap where a direct `GrpcEndpoint.start()` caller had no SSRF/TLS defense, and the follow-up refactor to `mcpgateway/utils/grpc_validation.py` is a clean way to kill the old lazy-import circular-dependency workaround. Tests are solid: the new `trusted_local` default-path, bypass-path, and both-TLS-paths branches are all covered, and the 149 tests across the touched files pass locally.

**Blocking**
- This changes default behavior for the `translate --grpc` CLI in a way that isn't called out anywhere. `expose_grpc_via_sse()` now calls `endpoint.start()` with `trusted_local=False` by default, and both `SSRF_ALLOW_LOCALHOST` and `SSRF_ALLOW_PRIVATE_NETWORKS` default to `false`. So an operator running `python -m mcpgateway.translate --grpc localhost:50051` (or any internal service host) — a very common dev/self-hosted case — now gets a `GrpcServiceError` out of the box. That's the right security default, but `docs/docs/using/mcpgateway-translate.md` and `docs/docs/using/grpc-services.md` don't mention it, and there's no CLI flag to opt in short of flipping a global SSRF setting. Could we get a short docs note (similar to how `docs/docs/manage/upgrade-to-1.0.0-rc3.md` called out the earlier SSRF default flip), pointing at `SSRF_ALLOWED_NETWORKS` as the scoped way to allow a specific internal target?

**Suggestions (non-blocking)**
- `invoke_method()` in `grpc_service.py` already calls `_validate_grpc_target`/`_validate_tls_path` explicitly (lines ~870-874) right before building the `GrpcEndpoint`, then calls `start()` without `trusted_local=True`, so the same DNS/path resolution runs twice on every method invocation. Since the explicit pre-check already gates the target immediately before use, passing `trusted_local=True` here would drop the redundant work on this hot path without weakening anything.

Everything else looks good — no dead code, no unrelated changes, `Closes #4615` referenced correctly.

msureshkumar88:
## E2E verification

Wrote and ran an e2e script against a **real gRPC server** (real TCP socket, real reflection service — no mocking) driven through the actual `python -m mcpgateway.translate --grpc` CLI entrypoint and direct API calls, to confirm the fix holds outside the unit-test harness.

Script: `scripts/verify-pr-5410-e2e.sh` (added in this verification pass, not part of the PR diff).

**Environment / config**
- Python 3.12.3, repo venv (`.venv`)
- Branch under test: `fix/grpc-endpoint-ssrf-validation` @ `662b468f8`
- Base repo `.env` has `SSRF_ALLOW_LOCALHOST=true` / `SSRF_ALLOW_PRIVATE_NETWORKS=true` (dev convenience) — the script explicitly overrides both to `false` for the "blocked by default" cases so the *production* SSRF default is what's actually exercised, not the dev override.
- Mock gRPC server: real `grpc.server` + `grpc_reflection.v1alpha.reflection`, listening on `127.0.0.1:44610`, no mocking/patching.

**Commands run**
```bash
bash scripts/verify-pr-5410-e2e.sh
```
Internally this drives (env shown per case):
```bash
# blocked hostname (default ssrf_blocked_hosts entry)
python -m mcpgateway.translate --grpc "metadata.google.internal:443" --port 0

# loopback target blocked under production SSRF defaults
SSRF_ALLOW_LOCALHOST=false SSRF_ALLOW_PRIVATE_NETWORKS=false \
  python -m mcpgateway.translate --grpc "127.0.0.1:44610" --port 0

# same loopback target allowed once opted in, real reflection round-trip
SSRF_ALLOW_LOCALHOST=true \
  python -m mcpgateway.translate --grpc "127.0.0.1:44610" --port 0

# TLS cert/key path outside allowed directories, target otherwise allowed
SSRF_ALLOW_LOCALHOST=true \
  python -m mcpgateway.translate --grpc "127.0.0.1:44610" --port 0 \
  --grpc-tls --grpc-cert /tmp/.../outside-cert.pem --grpc-key /tmp/.../outside-key.pem

# trusted_local=True escape hatch vs. control, direct GrpcEndpoint.start() calls
SSRF_ALLOW_LOCALHOST=false SSRF_ALLOW_PRIVATE_NETWORKS=false python -c "... ep.start(trusted_local=True) ..."
SSRF_ALLOW_LOCALHOST=false SSRF_ALLOW_PRIVATE_NETWORKS=false python -c "... ep.start() ..."
```

**Output**
```
== PR #5410 e2e verification: gRPC SSRF/TLS validation in GrpcEndpoint.start() ==
Branch/commit under test:
662b468f8 refactor(grpc): move shared SSRF/TLS validators to utils.grpc_validation
cce4eceb2 test(grpc): cover start() TLS-path validation branches
3ef0d1f1c fix(grpc): enforce SSRF/TLS validation in GrpcEndpoint.start()

PASS: no circular import; GrpcServiceError/_validate_grpc_target shared from mcpgateway.utils.grpc_validation
PASS: real gRPC server listening on 127.0.0.1:44610 with reflection enabled

PASS: test1: blocked hostname 'metadata.google.internal' rejected before any channel was opened (exit=1)
PASS: test2: real reachable loopback target rejected by default SSRF policy (exit=1)
PASS: test3: SSRF_ALLOW_LOCALHOST=true allows the loopback target; real reflection RPC against live server succeeded
PASS: test4: TLS cert/key path outside allowed directories rejected before any channel was opened (exit=1)
PASS: test5: trusted_local=True bypasses SSRF validation and connects to the real (otherwise-blocked) loopback target, warning logged
PASS: test6: default start() (trusted_local=False) still blocks the same target the bypass in test5 allowed

=== ALL CHECKS PASSED: PR #5410 fix verified end-to-end against a real gRPC server ===
```

8/8 checks passed. Confirms the fix holds against a live server, not just mocked unit tests: blocked hostnames and blocked private/loopback targets are rejected *before* any gRPC channel is opened, the `trusted_local` opt-out works as documented, and the `grpc_service` ↔ `translate_grpc` ↔ `grpc_validation` refactor has no circular-import regression.

Could you take a look at the two points from the earlier review (the undocumented CLI default-behavior change, and the double SSRF validation in `invoke_method`) when you get a chance? Happy to re-review and move to approval once those are addressed — nothing else is blocking at this point.

SaurabhMaydeo:
Thanks for the thorough e2e run, @msureshkumar88 — really appreciate the real-server verification. Both points are addressed:

- Docs: added a short SSRF note to `docs/docs/using/mcpgateway-translate.md` and `docs/docs/using/grpc-services.md` explaining that the `--grpc` target is validated before any channel opens, so `localhost:50051` is rejected under the default settings. It splits the remedy — `SSRF_ALLOW_LOCALHOST=true` for a loopback target, `SSRF_ALLOWED_NETWORKS` for a specific private range — and links the 1.0.0-RC3 upgrade notes.
- Perf: `invoke_method` now passes `trusted_local=True` to `start()`, so the target/TLS paths are resolved once (the explicit pre-check right before it) instead of twice per invocation.

Both are in the latest push.

msureshkumar88:
## Re-review status

Both items from the previous round are resolved:

- **Blocking (docs gap for the CLI default-deny change)** — fixed by `eb5e22464`. `docs/docs/using/grpc-services.md` and `docs/docs/using/mcpgateway-translate.md` now both call out that `SSRF_ALLOW_LOCALHOST`/`SSRF_ALLOW_PRIVATE_NETWORKS` default to `false`, so a `localhost:50051` target is rejected out of the box, and point at `SSRF_ALLOWED_NETWORKS` as the scoped opt-in.
- **Suggestion (redundant re-validation in `invoke_method`)** — fixed by `3c75205b1`. `invoke_method` still validates `service.target`/TLS paths immediately before constructing the endpoint (`grpc_service.py:870-874`), then calls `start(..., trusted_local=True)`, dropping the second DNS/path resolution on that hot path without weakening anything — the values passed to validation and to the endpoint constructor are the same `service.target`/`service.tls_cert_path`/`service.tls_key_path`, so there's no gap between check and use.

Also landed since the last round: `662b468f8` moves `_validate_grpc_target`/`_validate_tls_path`/`GrpcServiceError` into a new `mcpgateway/utils/grpc_validation.py`, which both `grpc_service.py` and `translate_grpc.py` import directly — this retires the old lazy-import-inside-function workaround for the `grpc_service` ↔ `translate_grpc` circular dependency. Clean simplification, no behavior change; confirmed no dead code left behind (old copies fully removed, not duplicated).

## Fresh review

**Scope** — diff is exactly the 8 files reported (`grpc-services.md`, `mcpgateway-translate.md`, `grpc_service.py`, `translate_grpc.py`, new `utils/grpc_validation.py`, 3 test files). Nothing unrelated; no `secrets.baseline` touched.

**Security (CWE)**
- CWE-918 (SSRF): this is the core fix — closes the gap where a direct `GrpcEndpoint.start()` caller (the `translate --grpc` CLI, `expose_grpc_via_sse()`) opened a channel with zero SSRF defense. Confirmed only two production call sites exist (`translate_grpc.py:650` and `grpc_service.py:897`), and both now go through validation — one via the new default (`start()` with no `trusted_local`), one via the immediate-pre-validation + `trusted_local=True` pattern above. No third call site was missed.
- CWE-22 (path traversal): `_validate_tls_path` constrains cert/key paths to `/certs`, `/etc/ssl`, `/etc/pki`, or `./certs`, applied consistently to both call sites.
- Hardening note, not a regression in this diff: SSRF validation resolves the hostname once at check time; the gRPC channel itself resolves independently at connect time. That's a DNS-rebinding/TOCTOU window shared with the existing `SecurityValidator._validate_ssrf` pattern used elsewhere in the codebase, not something introduced here — flagging for awareness rather than as a blocker on this PR.
- `trusted_local: bool` is not reachable from any external input today — the only caller that sets it `True` does so immediately after validating the same values. Worth keeping a code comment (already partially present) making clear this parameter must never be threaded through from untrusted input or set `True` without an immediately-preceding validation of the *same* target/paths.

**Performance** — the redundant re-validation fix removes the duplicate DNS resolution + path `.resolve()` call on every `invoke_method` invocation, which is the right target since this runs per tool call, not per registration.

**Testing**
- Unit coverage for the fix itself is solid: blocked-target rejection, `trusted_local=True` bypass, and default-path dual-TLS-validation are all directly tested (`test_translate_grpc.py`), and the fake `GrpcEndpoint` doubles across `test_grpc_service.py`/`test_grpc_service_no_grpc.py` were updated to accept the new kwarg. Ran the three affected files locally: 149 passed.
- Gaps (non-blocking):
  - No test asserts the once-per-process `_warn_trusted_local_once()` log line actually fires (e.g. via `caplog`) — the bypass path itself is tested, but not its audit trail.
  - No test pins `invoke_method`'s validate-once behavior at the call-count level (e.g. patching `_validate_grpc_target`/`_validate_tls_path` and asserting `call_count == 1`) — today's fix is correct, but nothing guards against a future edit silently reintroducing the double call.
  - No e2e/integration test exercises the CLI path end-to-end (`python -m mcpgateway.translate --grpc localhost:PORT` failing fast with the documented error) — understandable for unit scope, but worth a note if there's an e2e suite this belongs in.

**Logs/exceptions** — no gaps found; `GrpcServiceError` messages are specific per failure mode (empty target, forbidden scheme, blocked address, path outside allowed dirs), and the `trusted_local=True` bypass is logged as a warning (once per process) rather than silently skipped.

**Docs** — resolved (see above). Only remaining minor gap: no `CHANGELOG.md` entry for the default-deny behavior change, though that appears inconsistent across recent merges in this repo, so treating as optional.

**Breaking change** — yes, intentionally: any existing `translate --grpc` / direct `GrpcEndpoint.start()` usage against `localhost` or an RFC1918 target now fails by default unless `SSRF_ALLOW_LOCALHOST`/`SSRF_ALLOW_PRIVATE_NETWORKS`/`SSRF_ALLOWED_NETWORKS` are set. This is the correct secure-by-default posture and is now documented in both user-facing docs pages.

**Issue reference** — `Closes #4615` present in the PR description.

**Necessity** — yes. Prior to this fix, `GrpcEndpoint.start()` had no SSRF/TLS-path defense for any caller that didn't go through `GrpcService`'s own pre-checks, which is a real gap (cloud-metadata-endpoint-class SSRF) for the CLI/translate path.

CI is green (all checks passing on `662b468f8`… now superseded by `3c75205b1`/`eb5e22464`), `ruff` is clean on the changed files, and local run of the three affected test files is 149/149 passing. No new blocking issues found in this pass — holding off on approval per request, but this looks close to mergeable pending a final CI run on the latest commits.

msureshkumar88 (dismissed):
## E2E verification

Ran a real end-to-end check against this PR's exact code path — no mocks. Set up a real local gRPC server (a compiled `Echo` proto service with reflection enabled) and drove the actual CLI entrypoint (`python -m mcpgateway.translate --grpc ...` → `expose_grpc_via_sse()` → `GrpcEndpoint.start()`) as a subprocess, since that direct caller is exactly the gap this PR closes.

### Setup
- Real gRPC server: `grpc.server()` + a compiled `pr5410echo.Echo` service + `grpc_reflection.v1alpha` reflection enabled, bound to `127.0.0.1:<random-port>`.
- CLI invocation: `python -m mcpgateway.translate --grpc <target> --port <http-port> --logLevel debug`, run as a real subprocess against the repo checkout at commit `3c75205b1` (this PR's head).

### Scenario 1 — default settings vs. loopback target (expect BLOCKED)
```
env: SSRF_ALLOW_LOCALHOST=false SSRF_ALLOW_PRIVATE_NETWORKS=false
cmd: python -m mcpgateway.translate --grpc 127.0.0.1:<port> --port <http> --logLevel debug
```
Result: exit code 1, before any channel opened —
```
mcpgateway.utils.grpc_validation.GrpcServiceError: gRPC target contains localhost address which is blocked by SSRF protection
```
**PASS**

### Scenario 2 — SSRF_ALLOW_LOCALHOST=true vs. same target (expect SUCCESS)
```
env: SSRF_ALLOW_LOCALHOST=true SSRF_ALLOW_PRIVATE_NETWORKS=false
cmd: python -m mcpgateway.translate --grpc 127.0.0.1:<port> --port <http> --logLevel debug
```
Result: connected to the real server and discovered it via live reflection —
```
mcpgateway.translate_grpc - INFO - gRPC service exposed. Discovered services: ['pr5410echo.Echo']
```
(process then entered the CLI's keepalive loop as expected; terminated after confirming the log line). **PASS**

### Scenario 3 — blocklisted hostname despite permissive flags (expect BLOCKED)
```
env: SSRF_ALLOW_LOCALHOST=true SSRF_ALLOW_PRIVATE_NETWORKS=true
cmd: python -m mcpgateway.translate --grpc metadata.google.internal:443 --port <http> --logLevel debug
```
Result: exit code 1 —
```
mcpgateway.utils.grpc_validation.GrpcServiceError: gRPC target contains blocked hostname 'metadata.google.internal' (SSRF protection)
```
Proves the hostname blocklist (cloud metadata endpoints) can't be bypassed via the localhost/private-network toggles. **PASS**

### Before/after regression proof
To confirm this is a real fix and not just a passing new test, ran Scenario 1's exact command (same server, same default env) against an isolated `git worktree` of `origin/main` (pre-fix):
```
env: SSRF_ALLOW_LOCALHOST=false SSRF_ALLOW_PRIVATE_NETWORKS=false
cmd: python -m mcpgateway.translate --grpc 127.0.0.1:<port> --port <http> --logLevel debug   (cwd = main worktree)
```
Result: **no validation at all** — connected straight through and discovered the service:
```
mcpgateway.translate_grpc - INFO - gRPC service exposed. Discovered services: ['pr5410echo.Echo']
```
This confirms the SSRF gap described in the PR body existed on `main` and is closed by this branch.

### Verdict
All 3 scenarios behave exactly as the PR describes, and the before/after comparison confirms a genuine security fix rather than a vacuous test. Combined with the earlier review pass (both requested-change items resolved, CI green, 149/149 unit tests passing, no unrelated changes, docs updated for the breaking-change default) — **approving**.

Non-blocking follow-ups noted in the review (optional, not required for merge):
- Add a `caplog`-based test asserting `_warn_trusted_local_once()` actually logs.
- Pin `invoke_method`'s single-validation behavior with a call-count assertion so a future edit can't silently reintroduce the double validation.

msureshkumar88:
@SaurabhMaydeo 
## CI: "Run pre-commit hooks" is failing (blocks merge)

Not a merge-conflict/staleness issue — `mergeable` is already `MERGEABLE`, no conflicts with `main`, so a rebase won't fix this one.

**Root cause**: the docs note added to `docs/docs/using/grpc-services.md` in this PR (the +3 line SSRF callout) shifts a pre-existing benign `detect-secrets` match in that same file from line 414 to line 417. The committed `.secrets.baseline` still records it at 414, so `detect-secrets` (run fresh in CI) flags a mismatch and fails the hook:

```
🔐 IBM Detect Secrets....................................................Failed
- hook id: detect-secrets
- exit code: 3
- files were modified by this hook

The baseline file was updated.
Please `git add .secrets.baseline`, thank you.
```

**Fix** — regenerate the baseline and commit it alongside the docs change:
```
make detect-secrets-scan
git add .secrets.baseline
git commit -s -m "chore(grpc): regenerate secrets baseline for grpc-services.md line shift"
git push
```

This is the only failing required check right now; everything else (build, lint, pytest, vitest, playwright smoke) is green.

msureshkumar88 (dismissed):
LGMT

gcgoncalves (dismissed):

msureshkumar88 (approved):
LGMT

ja8zyjits (approved):
LGTM, approving after rebase.

SaurabhMaydeo:
Thanks @gcgoncalves, @msureshkumar88 and @ja8zyjits for the reviews. Glad it's in.

PR #5426, opened 2026-06-29
fix(auth): propagate end-user identity on the cached virtual-server invocation path

pos-ei-don:
### Problem

End-user identity propagation (#3152) injects `X-<prefix>-*` identity headers (and `_meta`) into upstream MCP requests in `ToolService.invoke_tool`, guarded by:

```python
if global_context and global_context.user_context:
    headers.update(build_identity_headers(global_context.user_context))
```

But the **cached** tool-call path used by **virtual servers** does not pass a `plugin_global_context`: the `streamablehttp_transport` tool-call handler calls `tool_service.invoke_tool(db=..., name=..., user_email=..., token_teams=..., ...)` **without** `plugin_global_context`. So for any client connecting via `/servers/<vserver-id>/mcp`, `global_context` is `None` and identity injection is **silently skipped** — the upstream server never receives the identity headers.

The `direct_proxy` path works (it resolves identity from `user_identity_var` / passes `user_context`), but it only activates when the client sends the `X-Context-Forge-Gateway-Id` header, which virtual-server clients never do.

Net effect: signed team/identity routing in a downstream MCP works on the direct_proxy path but is impossible over the normal virtual-server path.

### Root cause

`invoke_tool` reads identity **only** from `global_context.user_context`, while the transport auth layer already sets the per-request `user_identity_var` (via `_set_user_identity_from_dict`) on the cached path — `invoke_tool` just never consults it.

### Fix

Resolve the propagation identity from `global_context.user_context` **or** fall back to `user_identity_var` (symmetric with the `_proxy_*_to_gateway` helpers). Extracted into a small helper `_resolve_propagation_identity()` used at both injection sites (REST + MCP). `+34/-5` in one file.

### Reproduction

1. Federate a STREAMABLEHTTP gateway, expose its tools via a virtual server.
2. Connect a client to `/servers/<id>/mcp` with a JWT carrying a `teams` claim.
3. Call a tool; observe the upstream receives **no** `X-<prefix>-*` headers.
4. With the fix, the upstream receives `X-<prefix>-Id/-Email/-Teams/-Claims-Signature`.

### Tests

- New `tests/unit/mcpgateway/services/test_tool_service_vserver_identity.py` — unit tests for `_resolve_propagation_identity` (prefers global context; falls back to the contextvar when `plugin_global_context` is absent or carries no `user_context`; `None` when neither is set).
- Existing `tests/unit/mcpgateway/test_identity_propagation.py` and the full `tests/unit/mcpgateway/services/test_tool_service.py` suite pass unchanged (regression).

### Notes / possible follow-up

The MCP injection site does not pass the resolved `gateway` to `build_identity_headers/meta`, so per-gateway `identity_propagation` overrides are ignored on this path (direct_proxy passes it). Left out to keep this PR minimal; happy to add if preferred.

### DIFF
--- a/mcpgateway/services/tool_service.py
+++ b/mcpgateway/services/tool_service.py
@@ -114,6 +114,33 @@ _REGISTRY_CACHE = None
 _TOOL_LOOKUP_CACHE = None
 
 
+def _resolve_propagation_identity(global_context):
+    """Resolve the end-user identity to propagate to upstream MCP servers.
+
+    Prefers the plugin global context's ``user_context``; falls back to the
+    per-request ``user_identity_var`` set by the transport auth layer
+    (``streamablehttp_transport._set_user_identity_from_dict``).
+
+    The cached virtual-server invocation path calls :meth:`ToolService.invoke_tool`
+    without a ``plugin_global_context`` (see the ``streamablehttp_transport``
+    tool-call handler), so without this fallback identity propagation silently
+    no-ops for virtual-server tool calls while still working on the direct_proxy
+    path. See https://github.com/IBM/mcp-context-forge (vServer identity gap).
+
+    Args:
+        global_context: Optional plugin ``GlobalContext`` carrying ``user_context``.
+
+    Returns:
+        A ``UserContext`` to propagate, or ``None`` if no identity is available.
+    """
+    if global_context is not None and getattr(global_context, "user_context", None):
+        return global_context.user_context
+    # Local import to avoid import cycles at module load time.
+    from mcpgateway.transports.context import user_identity_var  # pylint: disable=import-outside-toplevel
+
+    return user_identity_var.get()
+
+
 def _get_registry_cache():
     """Get registry cache singleton lazily.
 
@@ -4988,8 +5015,9 @@ class ToolService(BaseService):
                             logger.debug("[AFFINITY] Worker %s | Session %s... | Tool: %s | Normalized MCP-Session-Id → x-mcp-session-id for pool affinity", worker_id, session_short, name)
 
                     # Inject identity propagation headers for REST tools
-                    if global_context and global_context.user_context:
-                        headers.update(build_identity_headers(global_context.user_context))
+                    _propagation_identity = _resolve_propagation_identity(global_context)
+                    if _propagation_identity:
+                        headers.update(build_identity_headers(_propagation_identity))
 
                     if plugin_manager and plugin_manager.has_hooks_for(ToolHookType.TOOL_PRE_INVOKE) and not skip_pre_invoke:
                         # Use pre-created Pydantic model from Phase 2 (no ORM access)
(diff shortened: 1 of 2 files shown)


pos-ei-don:
### Prod deployment update (2026-07-11)

This patch is now running in production against **CF v1.0.5** as a bind-mount patch on our
IBF ContextForge instance (previously running as a v1.0.2 bind-mount, dropped by mistake during
the v1.0.4 upgrade — see internal history — and now re-ported to v1.0.5).

Rebased against current `main` (48 commits, no conflicts). The delta is intentionally tiny:

- 14-line `_resolve_propagation_identity()` helper (falls back to `user_identity_var` set by
  transport auth) at the top of `mcpgateway/services/tool_service.py`
- Two guard replacements: `mcpgateway/services/tool_service.py` REST-tools path and MCP-tools path

**Verified end-to-end in production** (2026-07-11 03:15 CEST):
- Client → CF v1.0.5 (with bind-mount) → cached vServer path → cached-federation tool call →
  downstream MCP sees `x-mcp-identity-{id,email,teams,admin,auth-method,claims-signature}`
  with valid HMAC signature over `IDENTITY_CLAIMS_SECRET`
- Without the patch, all six identity headers are missing on this code path (only
  `mcp-session-id` + `mcp-protocol-version` arrive) — confirming the original bug report

We use this as the foundation for per-user secret retrieval from a downstream secrets broker
that verifies the CF-signed identity HMAC — so end-user Redmine tokens live in a per-user
vault and never in client-side config files. That downstream design would be unreachable
without this fix landing upstream. Would appreciate maintainer eyes on the tests + the small
patch surface — happy to iterate on approach if the helper needs restructuring.

jonpspri:
Relates to the end-user identity propagation feature originally delivered in #3152 (feat(auth): propagate end-user identity to upstream MCP servers).

This PR fixes the gap where the **cached** virtual-server invocation path in `streamablehttp_transport` bypassed the identity header injection because `plugin_global_context` was not threaded through to `tool_service.invoke_tool()`. Also relevant to #6416 ([UI-REWRITE]: Backend contract for virtual-server scoped tool invocation) which depends on this path working correctly.

pos-ei-don:
@jonpspri thanks — that framing matches what we see downstream of the fix.

Two short updates since my July comment.

**Still not native as of v1.0.7.** The patch has been carried across the v1.0.5 → v1.0.6 → v1.0.7 upgrades and is still required. On 2026-08-02 we instrumented `build_identity_headers()` in `identity_propagation.py` as a choke point and drove real virtual-server-routed Streamable-HTTP calls with session affinity active: without the patch the function fires **not once**, while the tool call itself succeeds. That rules out the reading that a later release fixed this path natively — the headers are not dropped somewhere downstream, they are never built.

**The `unstable` state on this PR looks misleading.** For the current head `feddfc2` (2026-07-11):

- the only reported check is **DCO**, and it is green
- **no workflow run is visible** for that commit in this repository
- the combined commit status is `pending` with **zero** statuses

So nothing is failing — the CI never ran. If workflow runs from this fork need a maintainer's approval, that is likely the only thing between this PR and a review. I can also rebase onto current `main` to trigger a fresh run if that is easier — just say which you prefer.

jonpspri:
Hi @pos-ei-don — sincere apologies for leaving this PR open for so long without a proper review or merge. That's on us, and we're sorry for the lack of responsiveness.

Issue #5141, opened 2026-06-09
[BUG]: Unable to pass plugin state on plugin manager rebuild.


Shankar-N-IBM-1:
### 🐞 Bug Summary
When TenantPluginManager is rebuilt either due to TTL expiry or other reasons, unable to pass state of the previous plugin instance to newly created instance.

---

### 🧩 Affected Component
Select the area of the project impacted:

- [x] `mcpgateway` - API
- [ ] `mcpgateway` - UI (admin panel)
- [ ] `mcpgateway.wrapper` - stdio wrapper
- [ ] Federation or Transports
- [ ] CLI, Makefiles, or shell scripts
- [ ] Container setup (Docker/Podman/Compose)
- [ ] Other (explain below)

---

### 🤔 Expected Behavior
Plugins should have a transforable state object which can be set by the plugin creator. This plugin state needs to be transported to the new instance of the plugin on rebuilt to main state consistancy.
---

### 📓 Logs / Error Output
Paste any relevant stack traces or logs here.
⚠️ **Do not paste secrets, credentials, or tokens.**

---

### 🧠 Environment Info
You can retrieve most of this from the `/version` endpoint.

| Key | Value |
|-----|-------|
| Version or commit | `e.g. v0.9.0 or main@a1b2c3d` |
| Runtime | `e.g. Python 3.11, Gunicorn` |
| Platform / OS | `e.g. Ubuntu 22.04, macOS` |
| Container | `e.g. Docker, Podman, none` |

---

### 🧩 Additional Context (optional)
Add any configuration details, flags, or related issues.

jonpspri:
Upstream feature request filed against the cpex framework: **[contextforge-org/cpex#173 — Plugin state transfer API for TenantPluginManager rebuild](https://github.com/contextforge-org/cpex/issues/173)**.

This is blocked on cpex adding `export_plugin_states()` / `initialize(initial_states=…)` hooks and a plugin-side `get_state()` / `set_state()` protocol. The ContextForge side (`TenantPluginManagerFactory._build_manager()`) already holds a reference to the outgoing manager before shutdown and is ready to extract/inject state once the upstream API exists.

PR #5434, opened 2026-06-30
Reuse default plugin manager

### LINKED ISSUE #5431: [BUG]: default manager rebuilding
### 🐞 Bug Summary
Briefly describe the issue or unexpected behavior.

When the plugin manager create a default config for a team::tool_id it should reuse the manager already created as global. In the current code we are recreating the global manager for each id.

However, it should keep tenant isolation. For each tentant it should create a new instance of default manager.

---

### 🧩 Affected Component

- [ ] `mcpgateway` - API
- [x] Other (explain below)
PLugins
---

### 🔁 Steps to Reproduce

1. Add a log to a plugin initialization
2. On first time a tool is called the initialization will occour
3.

---

### 🤔 Expected Behavior

Should initialize once

---

### 📓 Logs / Error Output

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|

---

cafalchio:
# Pull Request

## 🔗 Related Issue

Closes #5431

---

## 📝 Summary

The plugin manager factory uses the config.yaml to create a default manager that is used for all tool calls that are not customized via dynamic plugins. The currently behaviour is to recreate the default manager for each new id, which was ok for pure python class plugins. The current fix is to reuse the single default manager for all ids so we dont recreate plugins with the same config. 

Improvements:
Performance: Eliminates redundant plugin manager
Memory: Sharing of identical configurations

---

## 📏 Reviewability

- [ ] Unrelated bugs or improvements are tracked in separate issues/PRs

---

## 🏷️ Type of Change

- [ ] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

---

## ✅ Checklist

---

## 📓 Notes (optional)

### DIFF
--- a/docs/architecture-plugin-multi-tenancy.md
+++ b/docs/architecture-plugin-multi-tenancy.md
@@ -2,7 +2,7 @@
 
 ## Overview
 
-The plugin subsystem supports **context-scoped isolation** through a shared `TenantPluginManagerFactory`. Each resolved context gets its own `TenantPluginManager` instance with an independently merged plugin configuration, while the default `__global__` context continues to serve non-context-aware call sites.
+The plugin subsystem supports **context-scoped isolation** through a shared `TenantPluginManagerFactory`. Each resolved context gets its own `TenantPluginManager` instance with an independently merged plugin configuration, while the default `##global##` context continues to serve non-context-aware call sites.
 
 The factory is intentionally **context-agnostic**. The identifier passed to `get_manager()` can represent a virtual server, tenant, tool, user, or another scoping key. The factory does not interpret the value; it only uses it to:
 - look up an existing cached manager,
@@ -11,7 +11,7 @@ The factory is intentionally **context-agnostic**. The identifier passed to `get
 - cache the result for reuse.
 
 In the current gateway wiring, the primary runtime usage is:
-- `get_plugin_manager()` or `get_plugin_manager("__global__")` for shared/global plugin execution
+- `get_plugin_manager()` or `get_plugin_manager("##global##")` for shared/global plugin execution
 - `get_plugin_manager(server_id)` for server-scoped execution in services such as tools, prompts, and resources
 
 ---
@@ -83,7 +83,7 @@ flowchart TD
 flowchart TD
     APP["Application lifespan\n(mcpgateway.main)"]
     F["TenantPluginManagerFactory\n(singleton, holds base YAML config)"]
-    G["TenantPluginManager\ncontext = '__global__'\n(backward-compat global manager)"]
+    G["TenantPluginManager\ncontext = '##global##'\n(backward-compat global manager)"]
     S1["TenantPluginManager\ncontext = 'server-id-1'"]
     S2["TenantPluginManager\ncontext = 'server-id-2'"]
     DB["get_config_from_db(context_id)\n(fetch per-context overrides)"]
@@ -118,7 +118,7 @@ At startup, `mcpgateway.main.lifespan()`:
    - plugin timeout,
    - hook payload policies,
    - optional observability provider,
-3. calls `await get_plugin_manager()` to resolve the default `__global__` manager,
+3. calls `await get_plugin_manager()` to resolve the default `##global##` manager,
 4. leaves additional context-specific managers to be created lazily on first use.
 
 This means the factory is initialized eagerly, but most tenant/server managers are initialized on demand.
@@ -166,7 +166,7 @@ Defined in `mcpgateway/plugins/framework/manager.py`.
 
 | Method | Current behavior |
 | --- | --- |
-| `get_manager(context_id=None)` | Returns cached manager or creates one; defaults to `__global__` |
+| `get_manager(context_id=None)` | Returns cached manager or creates one; defaults to `##global##` |
 | `_build_manager(context_id)` | Fetches overrides, merges config, initializes manager, swaps cache entry |
 | `_merge_tenant_config(overrides)` | Applies per-plugin override values on top of base YAML config |
 | `reload_tenant(context_id)` | Evicts cached manager, rebuilds it, and shuts down the old one |
@@ -183,7 +183,7 @@ The public accessor lives in `mcpgateway/plugins/framework/__init__.py`.
 | --- | --- |
 | `enable_plugins(toggle)` | Enables or disables the plugin subsystem globally |
 | `init_plugin_manager_factory(...)` | Creates the singleton factory explicitly during startup |
-| `get_plugin_manager(server_id="__global__")` | Returns a context manager when plugins are enabled and the factory exists |
+| `get_plugin_manager(server_id="##global##")` | Returns a context manager when plugins are enabled and the factory exists |
 | `shutdown_plugin_manager_factory()` | Shuts down the factory and clears the singleton reference |
 | `reset_plugin_manager_factory()` | Clears the singleton reference for tests |
 
@@ -289,7 +289,7 @@ The current design preserves compatibility in a few important ways:
 
 - `PluginManager` still exists for Borg-based shared-state behavior.
 - `TenantPluginManager` keeps the same public lifecycle and hook invocation API as `PluginManager`.
-- `get_plugin_manager()` without arguments still resolves the global `__global__` manager.
+- `get_plugin_manager()` without arguments still resolves the global `##global##` manager.
 - Call sites that are not context-aware continue to function against the global manager.
 
 What changed is the wiring: the system now routes plugin access through the factory instead of a single shared manager instance.
--- a/docs/docs/architecture/plugins.md
+++ b/docs/docs/architecture/plugins.md
@@ -845,20 +845,24 @@ async def test_plugin_config_change():
 
 #### Shared Manager Optimization
 
-When no tenant-specific configuration is provided, the factory reuses a shared default plugin manager:
+When no tool-specific configuration is provided, the factory reuses a shared default plugin manager per team:
 
 ```python
-# Tenants without custom configs share the default manager
-manager_a = await factory.get_manager("tenant-a")  # No config → uses shared default
-manager_b = await factory.get_manager("tenant-b")  # No config → reuses same instance
+# Tools within the same team share the team's default manager
+manager_a1 = await factory.get_manager("team-a::tool1")  # No config → uses team-a::##global##
+manager_a2 = await factory.get_manager("team-a::tool2")  # No config → reuses team-a::##global##
 
-# Tenants with custom configs get dedicated managers
-manager_c = await factory.get_manager("tenant-c")  # Has config → dedicated instance
+# Different teams have separate default managers
+manager_b1 = await factory.get_manager("team-b::tool1")  # No config → uses team-b::##global##
+
+# Tools with custom configs get dedicated managers
+manager_a3 = await factory.get_manager("team-a::tool3")  # Has config → dedicated instance
 ```
 
 This optimization:
-- Reduces memory footprint when most tenants use default configuration
-- Maintains isolation through tenant-specific contexts during execution
+- Reduces memory footprint when most tools within a team use default configuration
+- Maintains team isolation with separate default managers per team
+- Maintains isolation through tool-specific contexts during execution
 - Automatically handles cleanup without double-shutdown errors during factory shutdown
 
 
--- a/docs/docs/architecture/plugins.md
+++ b/docs/docs/architecture/plugins.md
@@ -752,41 +752,41 @@ In multi-tenant deployments, each tenant can have a separate plugin configuratio
 ```python
 class TenantPluginManagerFactory:
     """Factory for creating and caching tenant-specific plugin managers"""
-    
+
     def __init__(self, cache_ttl: int = 300):
         """
         Initialize factory with configurable cache TTL.
-        
+
         Args:
             cache_ttl: Cache time-to-live in seconds (default: 300 = 5 minutes)
         """
         self._managers: Dict[str, Tuple[PluginManager, float]] = {}
         self._cache_ttl = cache_ttl
-    
+
     async def get_manager(self, tenant_id: str) -> PluginManager:
         """
         Get or create a plugin manager for the specified tenant.
-        
+
         Managers are cached with TTL-based expiration. If a cached manager
         exists and hasn't expired, it is returned. Otherwise, a new manager
         is created from the tenant's plugin configuration.
-        
+
         Args:
             tenant_id: Unique tenant identifier
-            
+
         Returns:
             PluginManager instance for the tenant
         """
         ...
-    
+
     def reload_tenant(self, tenant_id: str) -> None:
         """
         Invalidate cached plugin manager for a tenant.
-        
+
         This forces the next get_manager() call to create a fresh manager
         with the latest configuration. Must be called when a tenant's
         plugin configuration changes.
-        
+
         Args:
             tenant_id: Tenant whose cache should be invalidated
         """
@@ -834,10 +834,10 @@ For deployments with many tenants, monitor memory usage and consider:
 async def test_plugin_config_change():
     # Update configuration
     await update_tenant_plugins(tenant_id, new_config)
-    
+
     # CRITICAL: Reload to simulate production behavior
     factory.reload_tenant(tenant_id)
-    
+
     # Verify new configuration is active
     manager = await factory.get_manager(tenant_id)
     assert manager.plugin_count == expected_count
--- a/docs/docs/architecture/plugins.md
+++ b/docs/docs/architecture/plugins.md
@@ -741,6 +741,128 @@ class PluginInstanceRegistry:
     def unregister(self, name: str) -> None:
         """Unregister a plugin by name"""
 
+```
+
+# Multi-Tenant Plugin Manager
+
+In multi-tenant deployments, each tenant can have a separate plugin configuration. The `TenantPluginManagerFactory` manages per-tenant plugin manager instances with automatic caching and lifecycle management.
+
+#### Architecture
+
+```python
+class TenantPluginManagerFactory:
+    """Factory for creating and caching tenant-specific plugin managers"""
+    
+    def __init__(self, cache_ttl: int = 300):
+        """
+        Initialize factory with configurable cache TTL.
+        
+        Args:
+            cache_ttl: Cache time-to-live in seconds (default: 300 = 5 minutes)
+        """
+        self._managers: Dict[str, Tuple[PluginManager, float]] = {}
+        self._cache_ttl = cache_ttl
+    
+    async def get_manager(self, tenant_id: str) -> PluginManager:
+        """
+        Get or create a plugin manager for the specified tenant.
+        
+        Managers are cached with TTL-based expiration. If a cached manager
+        exists and hasn't expired, it is returned. Otherwise, a new manager
+        is created from the tenant's plugin configuration.
+        
+        Args:
+            tenant_id: Unique tenant identifier
+            
+        Returns:
+            PluginManager instance for the tenant
+        """
+        ...
+    
+    def reload_tenant(self, tenant_id: str) -> None:
+        """
+        Invalidate cached plugin manager for a tenant.
+        
+        This forces the next get_manager() call to create a fresh manager
+        with the latest configuration. Must be called when a tenant's
+        plugin configuration changes.
+        
+        Args:
+            tenant_id: Tenant whose cache should be invalidated
+        """
+        if tenant_id in self._managers:
+            del self._managers[tenant_id]
+```
+
+#### Cache Behavior
+
+**Cache TTL**: Plugin managers are cached for 300 seconds (5 minutes) by default. This balances:
+- **Performance**: Avoids recreating managers on every request
+- **Configuration freshness**: Changes propagate within reasonable time
+- **Memory efficiency**: Stale managers are automatically cleaned up
+
+**Cache Invalidation**: The cache must be explicitly invalidated when plugin configuration changes:
+
+```python
+# Example: After updating tenant plugin configuration
+await plugin_service.update_tenant_config(tenant_id, new_config)
+
+# CRITICAL: Invalidate cache so changes take effect immediately
+factory.reload_tenant(tenant_id)
+```
+
+#### Production Considerations
+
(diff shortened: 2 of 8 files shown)


mcpgateway/plugins/gateway_plugin_manager.py
```diff
@@ -167,8 +168,19 @@ async def get_manager(self, context_id: Optional[str] = None) -> TenantPluginMan
...
+
+        new_config = await self.get_config_from_db(context_id)
+
+        if new_config is None:
+            async with self._lock:
+                default_manager = self._managers.get(DEFAULT_CONTEXT_ID)
+                if default_manager is None:
+                    logger.debug("No default manager available, using base config for context_id=%s", context_id)
+                else:
+                    self._managers[context_id] = _CachedManager(manager=default_manager.manager, created_at=time.monotonic())
```
vishu-bh: This aliases tenant/tool contexts with no DB bindings to the exact same live `TenantPluginManager` instance as `__global__`. That optimizes rebuilds, but it weakens the tenant isolation called out in #5431 because plugin managers/plugins can carry mutable runtime state: caches, runtime-disabled plugin sets, HTTP clients, PDP decision caches, etc. Could we instead reuse immutable/default config work while still creating an isolated manager instance per tenant context, or otherwise prove all shared manager state is safe across tenants?

mcpgateway/plugins/gateway_plugin_manager.py
vishu-bh: After line 180 stores the default manager under tenant context IDs, this expiry path can shut down the shared default manager through an expired tenant alias. `__global__` and other no-binding tenant contexts can still point at that same now-shutdown object, which can make plugin hooks disappear or behave incorrectly for unrelated contexts. Shared/default aliases need ownership-aware shutdown semantics, or tenant entries should not reference the same manager object.

mcpgateway/plugins/gateway_plugin_manager.py
```diff
@@ -409,7 +421,7 @@ async def get_config_from_db(self, context_id: str) -> Optional[list[PluginConfi
     async def invalidate_all(self) -> None:
         """Reload every cached manager concurrently, logging failures."""
         async with self._lock:
-            context_ids = list(self._managers.keys())
+            context_ids = [ctx_id for ctx_id in self._managers if ctx_id != DEFAULT_CONTEXT_ID]
```
vishu-bh: Excluding `DEFAULT_CONTEXT_ID` from `invalidate_all()` means runtime plugin mode changes may not refresh the default manager. Since mode-change pub/sub invalidates all managers, a guardrail/security plugin switched to `enforce` or `disabled` could leave the default manager stale indefinitely, and no-binding tenant contexts may continue using that stale policy.

tests/unit/mcpgateway/plugins/test_gateway_plugin_manager.py
```diff
@@ -651,6 +651,75 @@ async def build_and_inject(context_id):
...
+            result = await factory.get_manager("team-x::tool-y")
+
+        # Should return the default manager
+        assert result is default_manager
+        # Verify it was cached for the tenant context
+        assert "team-x::tool-y" in factory._managers
+        assert factory._managers["team-x::tool-y"].manager is default_manager
+
+    @pytest.mark.asyncio
+    async def test_default_manager_shared_across_tenants_without_config(self, factory):
```
vishu-bh: This test asserts that multiple tenants share the same default manager instance, but that is the behavior that needs the most scrutiny. I think we also need coverage proving tenant state cannot leak across contexts, runtime mode changes refresh the default/default-derived manager, and reloading or expiring one aliased context does not shut down the manager still used by other contexts.

vishu-bh (changes requested):

mcpgateway/plugins/gateway_plugin_manager.py
cafalchio: Thank you for catching it. It makes sense, the global manager would be shutdown here.

mcpgateway/plugins/gateway_plugin_manager.py
cafalchio: I think this is the best option for now until we make sure the plugins are stateles or the manager can manage the state, instead of the plugin. 
https://github.com/IBM/mcp-context-forge/issues/5141

mcpgateway/plugins/gateway_plugin_manager.py
cafalchio: In the current implementation, the changes cannot be done in the default (loaded by yaml). The changes are applied per context_id.

vishu-bh: Per-context DB bindings are scoped by `context_id`, yes. But runtime plugin mode changes are not context-scoped today. `publish_plugin_mode_change()` writes `plugin:<name>:mode`, and `_apply_redis_mode_overrides()` applies those Redis/local overrides whenever any manager is rebuilt, including `__global__`. 

But plugin mode can change during rebuild which will prevent skipping TTL expiry for `DEFAULT_CONTEXT_ID` and excluding it from `invalidate_all()`

If runtime mode changes are truly intended to be per-context only, then the Redis key/path should include `context_id`, and `_apply_redis_mode_overrides()` should not apply global overrides to `__global__`. But with current code, default manager is affected by runtime mode overrides, so it needs invalidation/reload coverage too.

what do you think?

vishu-bh (changes requested):

mcpgateway/plugins/gateway_plugin_manager.py
cafalchio: I agree, we need to keep invalidate all to all managers.

tests/unit/mcpgateway/plugins/test_gateway_plugin_manager.py
cafalchio: I have decided to change the default manager and instead use a team based default manager. However, team is not always present on hooks. In this case it will be a default global one.

mcpgateway/plugins/gateway_plugin_manager.py
```diff
@@ -165,10 +166,32 @@ async def get_manager(self, context_id: Optional[str] = None) -> TenantPluginMan
...
+        Fallback: When no DB config exists for a team context, reuses the team's
+        default manager (team_id::__global__). Multiple contexts may share the
+        same manager instance. Invalidating a shared default affects all contexts
+        referencing it.
+        """
         manager = None
+        new_config = None
+        team_id = None
+
+        if CONTEXT_ID_SEPARATOR in context_id and DEFAULT_CONTEXT_ID not in context_id:
```
vishu-bh: This uses a substring check on the full context ID to decide whether to skip DB binding lookup. A tool named something like `my__global__tool` produces `team-a::my__global__tool`, so `DEFAULT_CONTEXT_ID not in context_id` is false and we never call `get_config_from_db()`. That can bypass exact or wildcard DB plugin bindings and fall back to the default/base config. Could we parse the context first (`team_id, tool_name = context_id.split(CONTEXT_ID_SEPARATOR, 1)`) and only treat it as default when `tool_name == DEFAULT_CONTEXT_ID`?

tests/unit/mcpgateway/plugins/test_gateway_plugin_manager.py
```diff
@@ -473,10 +473,14 @@ def factory(self):
 
     @pytest.mark.asyncio
     async def test_build_manager_happy_path(self, factory):
+        from mcpgateway.plugins.gateway_plugin_manager import DEFAULT_CONTEXT_ID
```
madhu-mohan-jaishankar: move import to top

tests/unit/mcpgateway/plugins/test_gateway_plugin_manager.py
```diff
@@ -650,8 +656,98 @@ async def build_and_inject(context_id):
 
         assert result is cached_manager
 
+    @pytest.mark.asyncio
+    async def test_distinct_default_manager_for_each_team(self, db_session):
+        from mcpgateway.plugins.gateway_plugin_manager import DEFAULT_CONTEXT_ID, TenantPluginManagerFactory
+        import os
```
madhu-mohan-jaishankar: move imports to top. same for all test methods below.

mcpgateway/plugins/gateway_plugin_manager.py
vishu-bh: `reload_tenant()` unconditionally shuts down the popped manager. With the new shared team default behavior, this can be the same manager still referenced by `team::__global__` and other no-config tool contexts. If a binding is added or a single tool context is invalidated, this can clear hooks on the shared default manager for unrelated contexts. Could we check whether any remaining cache entry references `old` before shutting it down?

madhu-mohan-jaishankar (changes requested):

mcpgateway/plugins/gateway_plugin_manager.py
```diff
@@ -446,3 +470,11 @@ def iter_context_ids(self) -> list[str]:
 def make_context_id(team_id: str, tool_name: str) -> str:
     """Build the context_id string expected by TenantPluginManagerFactory."""
     return f"{team_id}{CONTEXT_ID_SEPARATOR}{tool_name}"
+
+
+def get_team_id_from_context(context_id: str) -> Optional[str]:
+    """Get team_id from context_id"""
+    if CONTEXT_ID_SEPARATOR in context_id:
+        team_id, _ = context_id.split(CONTEXT_ID_SEPARATOR)
```
vishu-bh: This helper splits on every separator occurrence. Existing DB parsing uses `split(CONTEXT_ID_SEPARATOR, 1)`, which is safer because tool names can contain the separator. As written, `team::tool::name` raises `ValueError`. Could we use `split(CONTEXT_ID_SEPARATOR, 1)` here too?

docs/docs/architecture/plugins.md
```diff
@@ -741,6 +741,128 @@ class PluginInstanceRegistry:
...
+```
+
+#### Shared Manager Optimization
+
+When no tenant-specific configuration is provided, the factory reuses a shared default plugin manager:
+
+```python
+# Tenants without custom configs share the default manager
+manager_a = await factory.get_manager("tenant-a")  # No config → uses shared default
+manager_b = await factory.get_manager("tenant-b")  # No config → reuses same instance
```
vishu-bh: This is just to keep implementation and docs in sync 

This example says tenants without custom configs share the same default manager across tenants, but the implementation/tests now create a per-team default manager (`team-a::__global__`, `team-b::__global__`). Could we update this section to say tools within the same team share the team default, while different teams get distinct default managers?

vishu-bh (changes requested):
This is a solid improvement on designs, it is much more cleaner and robust. Just spotted some blocking items in review

docs/docs/architecture/plugins.md
cafalchio: This is a valid issue, I checked the validation for tool name and it allows underscore. We should be fine with #global

mcpgateway/plugins/gateway_plugin_manager.py
cafalchio: thanks

tests/unit/mcpgateway/plugins/test_gateway_plugin_manager.py
cafalchio: I will do a cleanup in this file and move all

mcpgateway/plugins/gateway_plugin_manager.py
cafalchio: added

mcpgateway/plugins/gateway_plugin_manager.py
cafalchio: I stopped to shutdown managers that are used by other context id.

madhu-mohan-jaishankar: This TTL expiry path can still shut down a shared default manager through an expired alias. For example, `team-a::tool-1` can be cached as an alias to the same manager instance as `team-a::##global##`. If only `team-a::tool-1` expires, we pop that alias but then call `shutdown()` on the shared manager object, leaving `team-a::##global##` and any other aliases pointing at a shut-down manager. Could we check whether any remaining cache entry still references `expired.manager` before shutting it down, similar to the ownership-aware guard in `reload_tenant()`?

madhu-mohan-jaishankar (changes requested):

mcpgateway/plugins/gateway_plugin_manager.py
cafalchio: I created a function to check if it is shared mangager. `_is_shared_manager` The shared manager should not expire.
Thank you for the review. @vishu-bh Pointed it, I though I had solved it already.

mcpgateway/plugins/gateway_plugin_manager.py
```diff
@@ -129,13 +130,15 @@ def plugin_names(self) -> list[str]:
...
+
+        context_id = context_id or DEFAULT_CONTEXT_ID
         expired = None
 
         async with self._lock:
             entry = self._managers.get(context_id)
             if entry is not None:
-                if entry.is_expired(self._cache_ttl):
+                # Shared managers should not expire
+                if entry.is_expired(self._cache_ttl) and not self._is_shared_manager(context_id):
```
madhu-mohan-jaishankar: This still checks whether the context ID is a default/shared context, but the unsafe case is an alias context like team-a::tool-1 that points to the same manager instance as team-a::##global##. _is_shared_manager("team-a::tool-1") returns false, so an expired alias can still pop itself and then shut down the shared manager object used by the default entry. Could we check ownership by manager identity instead, e.g. after popping, only call expired.manager.shutdown() if no remaining _managers entry references expired.manager?

madhu-mohan-jaishankar (changes requested):

mcpgateway/plugins/gateway_plugin_manager.py
cafalchio: True, I have created another function to check if the manager is referenced anywhere as in reload tenant. please review it again.

docs/docs/architecture/plugins.md
```diff
@@ -741,6 +741,132 @@ class PluginInstanceRegistry:
     def unregister(self, name: str) -> None:
         """Unregister a plugin by name"""
 
+```
+
+# Multi-Tenant Plugin Manager
+
```
vishu-bh, line 762: This new Multi-Tenant Plugin Manager section appears to split the `PluginInstanceRegistry` code sample. The remaining registry methods (`get_plugin`, `get_plugins_for_hook`, `get_all_plugins`, `plugin_count`, `shutdown`) are now left as orphaned indented lines before the closing fence, which breaks the rendered docs. Could we move this new section after the registry code block, or include those methods back inside the original code fence? Also worth aligning the wording with the new behavior: default managers are shared per team, so not every resolved context gets its own manager anymore.

vishu-bh (changes requested):
Code looks clean and approach is solid, minor comment needs docs to match code

docs/docs/architecture/plugins.md
cafalchio: addressed, thanks

madhu-mohan-jaishankar (dismissed):
LGTM

mcpgateway/plugins/gateway_plugin_manager.py
```diff
@@ -164,11 +168,55 @@ async def get_manager(self, context_id: Optional[str] = None) -> TenantPluginMan
...
+
+        # Parse context_id to check if it's a default context
+        if CONTEXT_ID_SEPARATOR in context_id:
+            team_id, _ = context_id.split(CONTEXT_ID_SEPARATOR, 1)
+
+            if not self._is_default_context(context_id):
+                async with self._lock:
+                    new_config = await self.get_config_from_db(context_id)
+
+        if new_config is None:
```
vishu-bh, line 209: For an unbound tool, `_build_manager()` can cache `team-a::tool1` as an alias to the existing `team-a::##global##` entry here. During `invalidate_all()` / `invalidate_team()` (lines 466-487), aliases and the default context are reloaded concurrently; if the alias reload runs first, it can reattach to the old default manager, then the default context rebuilds to a new manager. Final state can be `team-a::##global## -> new_manager` while `team-a::tool1 -> old_manager`, so mode/binding changes may not apply immediately to that tool. Could we make invalidation keep all aliases that referenced the old team default in sync with the rebuilt default?

mcpgateway/plugins/gateway_plugin_manager.py
```diff
@@ -129,23 +130,26 @@ def plugin_names(self) -> list[str]:
...
-                    await expired.manager.shutdown()
-                except Exception:
-                    logger.warning("Failed to shutdown expired manager for context_id=%s", context_id, exc_info=True)
+                if not self._is_manager_referenced(expired.manager):
+                    try:
+                        await expired.manager.shutdown()
+                    except Exception:
+                        logger.warning("Failed to shutdown expired manager for context_id=%s", context_id, exc_info=True)
 
             inflight = self._inflight.get(context_id)
```
vishu-bh, line 154: There is a race when two unbound tools from the same team miss the cache at the same time. The inflight build is keyed by the original context (`team-a::tool1`, `team-a::tool2`) here, but both builds later canonicalize to `team-a::##global##` around lines 210-217. That means both can build a default manager; the later build replaces and shuts down the first manager around lines 231-240 while the first caller may still receive it. Could we dedupe inflight builds by the canonical team-default key so concurrent no-config tools share the same build?

mcpgateway/plugins/gateway_plugin_manager.py
```diff
@@ -164,11 +168,55 @@ async def get_manager(self, context_id: Optional[str] = None) -> TenantPluginMan
...
         manager = None
+        new_config = None
+        team_id = None
+
+        # Parse context_id to check if it's a default context
+        if CONTEXT_ID_SEPARATOR in context_id:
+            team_id, _ = context_id.split(CONTEXT_ID_SEPARATOR, 1)
+
+            if not self._is_default_context(context_id):
+                async with self._lock:
```
vishu-bh: This DB lookup happens while the factory-wide `_lock` is held. If `get_config_from_db()` is slow or blocked, manager reads/builds for unrelated teams are blocked too, which can turn one tenant’s DB delay into broader plugin enforcement latency. Could we avoid holding `_lock` across the DB call and only reacquire it when reading/updating `_managers` / `_inflight`?

docs/architecture-plugin-multi-tenancy.md
```diff
@@ -2,7 +2,7 @@
 
 ## Overview
 
-The plugin subsystem supports **context-scoped isolation** through a shared `TenantPluginManagerFactory`. Each resolved context gets its own `TenantPluginManager` instance with an independently merged plugin configuration, while the default `__global__` context continues to serve non-context-aware call sites.
+The plugin subsystem supports **context-scoped isolation** through a shared `TenantPluginManagerFactory`. Each resolved context gets its own `TenantPluginManager` instance with an independently merged plugin configuration, while the default `##global##` context continues to serve non-context-aware call sites.
```
vishu-bh, line 5: This sentence still says each resolved context gets its own `TenantPluginManager`, but the new behavior intentionally shares a per-team default manager for unbound tool contexts. Could we update this wording so it matches the implementation, e.g. context-specific managers are used when overrides exist, otherwise unbound tools share the team default manager?

vishu-bh (changes requested):

mcpgateway/plugins/gateway_plugin_manager.py
cafalchio: Took me a while to understand this. I think this entire change is getting too complex :)
I will reload all the defaults and then reload all the tools, so they can point to the latest defaults.

vishu-bh: yes I think sounds good. I know this is getting more complex then we initially started. May be as a test before merging can we ask other devs to test this locally as well. Cause this a core functionality change and if anything breaks near release we will have very less time fixing it. What do you think?

docs/architecture-plugin-multi-tenancy.md
cafalchio: this is correct. Per team is just when we can extract team from the context. Global is kept for context-agnostic.

mcpgateway/plugins/gateway_plugin_manager.py
cafalchio: We dont need the lock for the DB. Removed

mcpgateway/plugins/gateway_plugin_manager.py
cafalchio: I have addressed the issue in invalidate all and invalidate team

mcpgateway/plugins/gateway_plugin_manager.py
```diff
@@ -129,23 +130,26 @@ def plugin_names(self) -> list[str]:
...
+
+        context_id = context_id or DEFAULT_CONTEXT_ID
         expired = None
 
         async with self._lock:
             entry = self._managers.get(context_id)
             if entry is not None:
-                if entry.is_expired(self._cache_ttl):
+                # Shared managers should not expire
+                if entry.is_expired(self._cache_ttl) and not self._is_default_context(context_id):
```
vishu-bh, line 141: `admin.py (line 17743)` and `plugins/__init__.py (line 288)` rely on TTL refresh as fallback when invalidation/pub-sub fails. But `##global##` and `team::##global##` now never refresh. Security mode changes like permissive -> enforce may never reach those managers until restart/manual invalidation.

mcpgateway/plugins/gateway_plugin_manager.py
vishu-bh: There is still a first-miss race for no-config tools in the same team. `get_manager()` tracks inflight builds by the original context id (`team-a::tool1`, `team-a::tool2`), but `_build_manager()` later maps both no-config contexts to the same default cache key (`team-a::##global##`). If both requests miss at the same time, both builds can run; the second build stores its manager under `team-a::##global##`, treats the first manager as old, and shuts it down, while the first caller can still receive that already-shut-down manager. Could we dedupe inflight builds using the canonical default context key for no-config tools so concurrent first misses share one build?

vishu-bh (changes requested):

cafalchio:
closing it for now.

PR #5435, opened 2026-06-30
fix: release fast-time image with docker release

lucarlig:
## Summary

- Move fast-time-server release publishing into the Docker release workflow.
- Stop the standalone fast-time-server image workflow from publishing on release tags.
- Build and verify `ghcr.io/ibm/fast-time-server:${TAG}` alongside the main image release tag.

## Validation

- `actionlint -shellcheck= .github/workflows/fast-time-server-image.yml .github/workflows/docker-release.yml`
- `ruby -e 'require "yaml"; ARGV.each { |f| YAML.load_file(f); puts "ok #{f}" }' .github/workflows/fast-time-server-image.yml .github/workflows/docker-release.yml`
- `git diff --check`

### DIFF
--- a/.github/workflows/docker-release.yml
+++ b/.github/workflows/docker-release.yml
@@ -2,15 +2,14 @@
 # 🐳  Docker Release Workflow - Tag and Push on GitHub Release
 # ======================================================================
 #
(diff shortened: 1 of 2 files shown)


lucarlig:
@jonpspri is already looking into this

PR #5436, opened 2026-06-30
fix(ci): upgrade anchore/scan-action to v7.4.0 (Node 24); skip unfixable CVEs

Lang-Akshay:
## Summary

- **Upgrade `anchore/scan-action` to v7.4.0** (`e1165082`) — Node 24 runtime, eliminates Node 20 deprecation warning
- **`only-fixed: true`** — only fails build on HIGH/CRITICAL CVEs that have a patch available
- **`sbom: sbom.spdx.json`** — scans Syft-generated SBOM directly (scan and uploaded artifact stay in sync)
- **`cache-db: true`** — caches Grype vulnerability DB between runs

### Node 20 deprecation

`anchore/scan-action@v6.2.0` bundled a Node 20 runtime. GitHub Actions runners now default to Node 24 and emit deprecation warnings. Upgrading to v7.4.0 moves to Node 24 — no workaround env vars needed.

### CVE scan false-positive failures

`fail-build: true` with `severity-cutoff: high` was failing CI on HIGH CVEs in the UBI10 base image with **no available fix**. Red Hat ships `ubi10/ubi-minimal` with open CVEs that remain unpatched for weeks or months.

`only-fixed: true` changes the failure condition:

| CVE severity | Fix available? | `only-fixed` behavior |
|---|---|---|
| HIGH | yes | ❌ fail build |
| CRITICAL | yes | ❌ fail build |
| HIGH | no | ✅ pass (skip) |
| CRITICAL | no | ✅ pass (skip) |

When Red Hat ships a patch, the CVE transitions to "fix available" — next scan catches it and CI fails, forcing an upgrade. Security coverage preserved; false-positive build failures eliminated.

### SBOM consistency

Using `sbom: sbom.spdx.json` passes the Syft-generated SBOM directly to Grype rather than rescanning the image independently. The uploaded artifact and the scan input are the same file.

## Changes

| File | Change |
|---|---|
| `.github/workflows/docker-scan.yml` | Upgrade `anchore/scan-action` v6→v7.4.0, switch to `sbom:` input, add `only-fixed: true`, `cache-db: true` |

### DIFF
--- a/.github/workflows/docker-scan.yml
+++ b/.github/workflows/docker-scan.yml
@@ -203,11 +203,14 @@ jobs:
             anchore/syft:v1.42.3 \
             "docker:${IMAGE_NAME}:scan" -o spdx-json=/work/sbom.spdx.json
 
-      - name: Install Grype
-        run: curl -sSfL https://raw.githubusercontent.com/anchore/grype/main/install.sh | sh -s -- -b /usr/local/bin v0.91.1
-
-      - name: Scan SBOM for vulnerabilities (Grype)
-        run: grype sbom:sbom.spdx.json --fail-on high --only-fixed
+      - name: Scan for vulnerabilities (Grype)
+        uses: anchore/scan-action@e1165082ffb1fe366ebaf02d8526e7c4989ea9d2 # v7.4.0
+        with:
+          image: ${{ env.IMAGE_NAME }}:scan
+          only-fixed: true
+          fail-build: true
+          severity-cutoff: high
+          cache-db: true
 
       - name: Upload SBOM
         uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
--- a/.github/workflows/docker-scan.yml
+++ b/.github/workflows/docker-scan.yml
@@ -203,10 +203,10 @@ jobs:
             anchore/syft:v1.42.3 \
             "docker:${IMAGE_NAME}:scan" -o spdx-json=/work/sbom.spdx.json
 
-      - name: Scan for vulnerabilities (Grype)
+      - name: Scan SBOM for vulnerabilities (Grype)
         uses: anchore/scan-action@e1165082ffb1fe366ebaf02d8526e7c4989ea9d2 # v7.4.0
         with:
-          image: ${{ env.IMAGE_NAME }}:scan
+          sbom: sbom.spdx.json
           only-fixed: true
           fail-build: true
           severity-cutoff: high
--- a/.github/workflows/docker-scan.yml
+++ b/.github/workflows/docker-scan.yml
@@ -205,10 +205,13 @@ jobs:
 
       - name: Scan SBOM for vulnerabilities (Grype)
         uses: anchore/scan-action@2c901ab7378897c01b8efaa2d0c9bf519cc64b9e # v6.2.0
+        env:
+          ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: "true"
         with:
           sbom: sbom.spdx.json
           fail-build: true
           severity-cutoff: high
+          only-fixed: true
           cache-db: true
 
       - name: Upload SBOM
--- a/.github/workflows/docker-scan.yml
+++ b/.github/workflows/docker-scan.yml
@@ -203,16 +203,11 @@ jobs:
             anchore/syft:v1.42.3 \
             "docker:${IMAGE_NAME}:scan" -o spdx-json=/work/sbom.spdx.json
 
+      - name: Install Grype
+        run: curl -sSfL https://raw.githubusercontent.com/anchore/grype/main/install.sh | sh -s -- -b /usr/local/bin v0.91.1
+
       - name: Scan SBOM for vulnerabilities (Grype)
-        uses: anchore/scan-action@2c901ab7378897c01b8efaa2d0c9bf519cc64b9e # v6.2.0
-        env:
-          ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: "true"
-        with:
-          sbom: sbom.spdx.json
-          fail-build: true
-          severity-cutoff: high
-          only-fixed: true
-          cache-db: true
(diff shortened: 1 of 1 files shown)


gcgoncalves (dismissed):

msureshkumar88 (changes requested):
Thanks for chasing this one down to the root cause — dropping the Node-20 action wrapper is the right instinct, and the `--only-fixed` reasoning is solid. I verified the scan half locally (grype `v0.91.1`, checksum matched) against the current SBOM:

- `grype sbom:sbom.spdx.json --fail-on high --only-fixed` is valid and exits non-zero on fixable HIGH CVEs.
- `--only-fixed` correctly drops unfixable base-image RPMs (util-linux/xz/zlib showed a blank `FIXED-IN` and disappeared under the flag).
- No `upload-sarif` step exists, so the "no functionality lost" claim holds.

One blocking concern and a couple of minor notes.

### Blocking — supply-chain regression in the install step

```
run: curl -sSfL https://raw.githubusercontent.com/anchore/grype/main/install.sh | sh -s -- -b /usr/local/bin v0.91.1
```

This swaps a **SHA-pinned** action (`anchore/scan-action@2c901ab…`) for a `curl | sh` of `install.sh` fetched from the mutable `main` branch. The grype *version* is pinned, but the install script itself is not — a compromise or MITM of that script would run arbitrary code in CI. Net effect is a downgrade in supply-chain posture rather than an upgrade.

**Suggested fix — mirror the existing Syft step.** The workflow already runs Syft as a pinned docker image (`anchore/syft:v1.42.3`). Running grype the same way removes the Node dependency *and* the unpinned script in one move, and stays consistent with the rest of the file:

```yaml
- name: Scan SBOM for vulnerabilities (Grype)
  run: |
    docker run --rm -v "${PWD}:/work" \
      anchore/grype:v0.91.1 \
      sbom:/work/sbom.spdx.json --fail-on high --only-fixed
```

That also lets you drop the separate "Install Grype" step entirely.

If a docker image isn't preferred, a couple of alternatives: download the release tarball and verify it against `grype_0.91.1_checksums.txt` before running (I confirmed the checksum matches), or at minimum pin the install URL to the tagged path (`…/anchore/grype/v0.91.1/install.sh`) instead of `main`.

### Minor notes (non-blocking)

- The old action had `cache-db: true`; the CLI now re-downloads the full vulnerability DB every run — slightly slower and adds a grype-DB download as a CI flake surface. Worth a cache if convenient, but not required.
- First commit body still has the `Closes #<issue>` placeholder — fill in the issue number or drop the line.

Happy to push the docker-image version if that's the direction you'd like to take.

jonpspri:
Disagree with the approach.

To address the Node 20 deprecation, update the pin for the action to v7.4.0 @e1165082ffb1fe366ebaf02d8526e7c4989ea9d2.

To bypass unfixable CVEs, set 'only-fixed: true' for the action.

No tokens were burned in the preparation of this comment :)

ja8zyjits:
Inputs from @msureshkumar88 are spot on.
That will fix most.
lets keep the cache-db: true for now.

Update:
Since mine and Jonathan comments came in at the same time, i want to address that too.

Jonathan prefers not running shell scripts in the workflow and rather use the actions but pin them.
That sounds valid.
What are your thoughts @msureshkumar88 and @Lang-Akshay ?

Lang-Akshay:
Thanks for the review @jonpspri 
1. Update to  `v7.4.0 @e1165082ffb1fe366ebaf02d8526e7c4989ea9d2` to address node 20 deprecation. 
2. Kept the unfixable CVE logic  with - `only-fixed: true`
3. Added. `cache-db`: true caches Grype's vulnerability DB between runs — faster scans on subsequent jobs.

ja8zyjits (dismissed):
LGTM

jonpspri (commented):
-

msureshkumar88 (approved):
Re-reviewed across all rounds and reviewers against current HEAD `28f602b`. The original `curl | sh` direct-CLI approach that drew the blocking concerns is gone — the change now simply bumps the SHA-pinned action and adds `only-fixed`. Every code-level request has been addressed:

| Reviewer | Request | Status |
|---|---|---|
| msureshkumar88 (blocking) | Supply-chain regression: unpinned `curl\|sh install.sh@main` | ✅ Resolved — reverted to SHA-pinned action |
| jonpspri | Pin action to `v7.4.0 @e1165082` | ✅ Exact SHA applied |
| jonpspri | Set `only-fixed: true` on the action | ✅ Applied |
| jonpspri | "Stick to scanning the SBOM" | ✅ `sbom: sbom.spdx.json` retained |
| jonpspri | No shell scripts; use pinned actions | ✅ `curl\|sh` dropped |
| ja8zyjits / msureshkumar88 | Keep `cache-db: true` | ✅ Retained |

The `--only-fixed` security tradeoff (accepting known-unfixable base-image HIGH CVEs until Red Hat ships a patch) was already discussed and accepted by reviewers. No bugs, no remaining security regression.

Approving. Two non-blocking housekeeping items worth a quick follow-up:
- The PR description still narrates the old "replace with direct grype CLI / curl install" approach, which now contradicts the diff — worth refreshing so the summary matches the pinned-action change.
- `Closes #<issue>` placeholder is still unfilled in the body and the first commit message.

msureshkumar88:
Confirmed — the scan stays scoped to the SBOM. The step passes `sbom: sbom.spdx.json` as the input, so grype reads the generated SBOM rather than re-scanning the image directly. 👍

One minor note (non-blocking): the PR description says "replace `anchore/scan-action` with direct grype CLI," but the actual diff keeps `anchore/scan-action` and bumps it `v6.2.0` → `v7.4.0` while adding `only-fixed: true`. Worth syncing the description with the final diff so future readers aren't confused about which approach landed.

gcgoncalves (approved):

jonpspri:
We discussed.  I'm okay with this being merged.

ja8zyjits (approved):
LGTM

PR #5438, opened 2026-06-30
Follow-on refinements to Docker Security Scan job

jonpspri:
Follow-on refinements to the Docker Security Scan workflow, now including the fixes from review.

### Job Restructuring
- Split the monolithic scan job into two focused jobs:
  - `fedramp-compliance`: FedRAMP compliance validation with the FIPS-enabled image
  - `scan`: standard security scanning with the lite image
- Updated `docker-scan-complete` to depend on both new jobs (plus hadolint and container-smoke)

### SARIF Integration
- SARIF output format for `merge_group` and `push` (main) events
- `github/codeql-action/upload-sarif` for both jobs, categorized `fedramp-fips` and `standard`
- Per-job `security-events: write` permission so the upload does not 403 and fail the merge queue
- `push: [main]` trigger so uploads land on the stable main ref instead of ephemeral `gh-readonly-queue/*` refs, which are deleted when the queue entry completes
- Display/Upload SARIF steps guard on a non-empty `sarif` output so a failed build no longer produces spurious `cat ""`/upload errors

### SBOM
- `anchore/sbom-action` generates the SBOM for both images, with `upload-artifact: false` so the explicit `Upload SBOM` step (with `retention-days: 30`) is the single artifact upload
- `Upload SBOM` runs with `if: always()` so the triage artifact survives a grype failure

### FedRAMP CVE Scan (report-only)
- The grype scan of the FIPS image runs with `fail-build: false`: it is built from full `ubi9/ubi` on floating tags, while `.grype.yaml`'s ignore list was calibrated for the lite image's base — a blocking gate here would be non-deterministic. SARIF is still uploaded; the gate can be flipped to blocking once ignores match this base.
- The lite-image scan remains blocking (`fail-build: true`, `severity-cutoff: high`).

### Notes
- `anchore/scan-action` was already pinned at v7.4.0 on the base branch; unchanged here.
- No `Containerfile` changes in this PR.
- The SARIF upload path has not yet executed end-to-end: it only runs on `merge_group`/`push`, so the first merge-queue entry for this PR is its first live exercise.

Signed-off-by: Jonathan Springer <jps@s390x.com>

### DIFF
--- a/.github/workflows/docker-scan.yml
+++ b/.github/workflows/docker-scan.yml
@@ -172,19 +172,18 @@ jobs:
             ${{ env.IMAGE_NAME }}:fedramp-validate \
             -c "$(cat scripts/fedramp-validate.sh)"
 
-      - name: Generate SBOM (Syft)
-        run: |
-          docker run --rm \
-            -v /var/run/docker.sock:/var/run/docker.sock \
-            -v "${PWD}:/work" \
-            anchore/syft:v1.42.3 \
-            "docker:${IMAGE_NAME}:fedramp-validate" -o spdx-json=/work/sbom-fedramp.spdx.json
+      - name: Generate SBOM
+        uses: anchore/sbom-action@e22c389904149dbc22b58101806040fa8d37a610 # 0.24.0
+        with:
+          image: ${{ env.IMAGE_NAME }}:fedramp-validate
+          output-file: sbom-fedramp.spdx.json
 
       - name: Scan SBOM for vulnerabilities (Grype)
         id: scan-fedramp
         uses: anchore/scan-action@e1165082ffb1fe366ebaf02d8526e7c4989ea9d2 # 7.4.0
         with:
           sbom: sbom-fedramp.spdx.json
+          only-fixed: true
           fail-build: true
           severity-cutoff: high
           cache-db: true
@@ -234,7 +233,7 @@ jobs:
         uses: docker/build-push-action@10e90e3645eae34f1e60eeb005ba3a3d33f178e8 # v6
         with:
           context: .
-          file: Containerfile.lite
+          file: Containerfile
           platforms: linux/amd64
           push: false
           load: true
@@ -242,13 +241,11 @@ jobs:
           cache-from: type=gha,scope=containerfile-lite-amd64
           cache-to: type=gha,mode=max,scope=containerfile-lite-amd64
 
-      - name: Generate SBOM (Syft)
-        run: |
-          docker run --rm \
-            -v /var/run/docker.sock:/var/run/docker.sock \
-            -v "${PWD}:/work" \
-            anchore/syft:v1.42.3 \
-            "docker:${IMAGE_NAME}:scan" -o spdx-json=/work/sbom.spdx.json
+      - name: Generate SBOM
+        uses: anchore/sbom-action@e22c389904149dbc22b58101806040fa8d37a610 # 0.24.0
+        with:
+          image: ${{ env.IMAGE_NAME }}:scan
+          output-file: sbom.spdx.json
 
       - name: Scan SBOM for vulnerabilities (Grype)
         uses: anchore/scan-action@e1165082ffb1fe366ebaf02d8526e7c4989ea9d2 # v7.4.0
--- a/.github/workflows/docker-scan.yml
+++ b/.github/workflows/docker-scan.yml
@@ -124,12 +124,12 @@ jobs:
           tags: ${{ matrix.tag }}
 
   # ---------------------------------------------------------------
-  # Build image and generate SBOM
+  # FedRAMP compliance validation
   # ---------------------------------------------------------------
-  scan:
+  fedramp-compliance:
     needs: [hadolint]
     if: github.event_name == 'merge_group' || github.event_name == 'workflow_dispatch'
-    name: Security Scan
+    name: FedRAMP Compliance
     runs-on: ubuntu-latest
     timeout-minutes: 30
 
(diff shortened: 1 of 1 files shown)


brian-hussey (commented):
This looks good from my perspective, but I'd like @madhu-mohan-jaishankar's review on it as well before merging.

brian-hussey (dismissed):

madhu-mohan-jaishankar (changes requested):
## Review summary

The job split (`fedramp-compliance` + `scan`) and the `anchore/sbom-action` swap look correct — the `needs:`/aggregator wiring is sound, the aggregator still treats only `success`/`skipped` as OK, and `fail-build: true` is preserved on both grype invocations, so there's **no new false-green path**.

The issue is the opposite — the SARIF path has never actually executed on this branch and is missing the permission it needs.

---

### 🔴 CRITICAL — `upload-sarif` will fail every merge_group run (missing `security-events: write`)

`.github/workflows/docker-scan.yml:43-44`, used at `:199-204` and `:268-273`

Top-level `permissions: contents: read`, and neither new job overrides it. `github/codeql-action/upload-sarif` requires `security-events: write`. In the pinned action (`8aad20d1`), the 403 is converted to a `ConfigurationError` and **thrown** — it does not degrade to a warning.

Blast radius: both jobs go red → `docker-scan-complete` exits 1 → **every PR in the merge queue gets ejected**, not just this one.

Why CI is green regardless: the steps are gated `if: always() && github.event_name == 'merge_group'`. The only runs on this branch were `pull_request` (jobs skipped) and one `workflow_dispatch` (run `29416018407`), where `Display SARIF file` and `Upload SARIF to CodeQL` were both `skipped` in both jobs. So the PR body's "SARIF upload tested in merge_group context" isn't backed by a run on this branch.

Fix — add per-job (matches the idiom in `.github/workflows/dependency-review.yml:47-50`):

```yaml
permissions:
  contents: read
  security-events: write
```

---

### 🟠 MEDIUM — New hard CVE gate on the FIPS/UBI9 image, with ignores calibrated for a different base

`.github/workflows/docker-scan.yml:181-190`

On base, the FedRAMP image was validated but **never CVE-scanned**. This PR adds a `fail-build: true` / `severity-cutoff: high` grype scan against `${IMAGE_NAME}:fedramp-validate`, built from full `ubi9/ubi` (line 162) rather than `ubi-minimal`, with Python 3.11 from PyPI instead of the hermetic wheel closure — a materially larger package surface now blocking the merge queue. The PR description doesn't mention this scope expansion.

`.grype.yaml`'s suppressions were written for the other image — its own comment says "the python3.12.13-2.el10_2 build shipped in our **UBI10** base image" — so the FIPS scan starts with an ignore list that doesn't match its base OS.

Compounding: lines 162-163 pin `ubi9/ubi:latest` and `ubi9/nodejs-20:latest` by floating tag. A blocking gate on a floating base is non-deterministic — the same commit can pass today and eject from the queue tomorrow with no code change.

Suggest either `fail-build: false` on `scan-fedramp` (report-only, SARIF still uploads) until the ignore list is calibrated, or pin `UBI_BASE`/`NODEJS_IMAGE` by digest as `UBI_MINIMAL` already is.

---

### 🟠 MEDIUM — SARIF uploads land on ephemeral `gh-readonly-queue/*` refs

`.github/workflows/docker-scan.yml:193, 200, 262, 269`

`upload-sarif` takes its target ref from `GITHUB_REF` verbatim (no `merge_group` special-casing in `git-utils.ts`). Under `merge_group` that's `refs/heads/gh-readonly-queue/main/pr-N-<sha>`, a branch GitHub deletes when the queue entry completes. The workflow has no `push: [main]` trigger, so no analysis is ever recorded for `main`; the steps are skipped on `pull_request`, so PR authors never see the alerts either. The `category:` values only pay off on a stable ref.

Unverified: whether the code-scanning API accepts an upload against a merge-queue ref at all — `api-client.ts` recognizes `ref .* not found in this repository` as a `ConfigurationError`, so that's a plausible second failure mode on top of the 403.

Fix: add a `push: branches: [main]` trigger for the scan jobs, or pass explicit `ref`/`sha` inputs to `upload-sarif`.

---

### 🟡 LOW — Every SBOM is uploaded twice

`.github/workflows/docker-scan.yml:175-179` + `:206-211`, and `:244-248` + `:275-280`

`anchore/sbom-action` defaults `upload-artifact: true` and derives an artifact name from the image, so the explicit `actions/upload-artifact` steps re-upload the same bytes. Confirmed in run `29416018407`: both `sbom-fedramp` (1,040,344 B) and `mcp-context-forge-scan_fedramp-validate.spdx.json` (1,040,398 B) are present, plus `mcp-context-forge-scan_scan.spdx.json` (533,781 B). ~1.5 MB duplicated per run, and the auto-uploaded copies bypass the intended `retention-days: 30`.

Fix: `upload-artifact: false` on both `Generate SBOM` steps, or drop the explicit uploads and set `artifact-name:`.

---

### 🟡 LOW — `Display SARIF file` does `cat ""` when the scan step is skipped

`.github/workflows/docker-scan.yml:192-197, 261-266`

`if: always() && github.event_name == 'merge_group'` runs the step even when `Build image locally` or `Generate SBOM` failed and grype never ran. `steps.scan.outputs.sarif` is then empty → `cat ""`, and the following upload runs with `sarif_file: ""` — two spurious errors burying the real build failure.

Fix: `if: always() && github.event_name == 'merge_group' && steps.scan.outputs.sarif != ''`.

---

### 🟡 LOW — `Upload SBOM` is skipped exactly when it's most useful

`.github/workflows/docker-scan.yml:206-211, 275-280`

The new SARIF steps got `if: always()`; `Upload SBOM` didn't, so a grype failure suppresses the artifact needed to triage it. Verified in run `29416018407`: `Security Scan` failed at the grype step and `Upload SBOM` was skipped, while the passing FedRAMP job uploaded fine. Inherited from base, but the new `always()` neighbors make the inconsistency stand out.

---

### 🟡 LOW — Stale comments and PR-description drift

- `.github/workflows/docker-scan.yml:5-17` still describes a single "Security Scan" job (there are now four) and says SBOM is generated by Syft directly; line 15 lists the merge_group gate as "Hadolint + Container smoke + Security Scan", omitting FedRAMP.
- `:287-289` says "container-smoke/scan are skipped" and omits `fedramp-compliance`, which is now in `needs:` at line 295.
- PR body claims an action bump v6.2.0 → v7.4.0, but base `992e6926a` already pinned `anchore/scan-action@e1165082 # v7.4.0` — SHA unchanged. It also claims a Containerfile base-image change, and commit `acd2b8ce6` is titled "Containerfile updates & Try SBOM module", but `Containerfile` isn't in the diff (1 file changed).
- Neither commit subject uses a Conventional Commits prefix, which `CONTRIBUTING.md` requires (no hook enforces it). DCO sign-offs are present on both.

madhu-mohan-jaishankar (approved):
LGTM

brian-hussey (approved):

PR #5444, opened 2026-06-30
fix(admin-ui, tests): stabilize iframe team-selector flaky test

### LINKED ISSUE #5414: [BUG]: Team selector strips team_id from URL when selecting newly created teams
### 🐞 Bug summary

**Summary:**
The team-selector component (`mcpgateway/admin_ui/components/team-selector.js`) incorrectly removes `team_id` query parameters from the URL when a user selects a newly created team. This occurs because the component validates `team_id` against a client-side cache (`window.USER_TEAMS_DATA`) that is rendered server-side at page load and never refreshed. Teams created after page load are not in this cache, causing the component to treat valid team IDs as invalid and strip them via `safeReplaceState()`.

**User impact:**
- Users who create teams via API cannot immediately select them
- URL changes from `/admin/?team_id=abc123#gateways` to `/admin/#gateways`
- Team selector falls back to "All Teams" instead of showing the selected team
- Breaks team context switching for newly created teams

---

### 🧩 Affected Component

- **File:** `mcpgateway/admin_ui/components/team-selector.js`
- **Test:** `tests/playwright/test_admin_url_context.py::TestAdminIframeContext::test_iframe_team_selector_onclick_stripped_but_delegation_works`

---

### 🔁 Steps to Reproduce

**Prerequisites:**
- User must be a member of at least one team (`window.USER_TEAMS_DATA.length > 0`)
- User is logged into the admin UI

**Reproduction:**

1. Verify cache is populated:
   ```javascript
   console.log('Teams in cache:', window.USER_TEAMS_DATA.length);
   ```

2. Create a new team without refreshing the page:
   ```javascript
   fetch('/teams', {
     method: 'POST',
     headers: {'Content-Type': 'application/json'},
     credentials: 'same-origin',
     body: JSON.stringify({name: 'NewTeam-' + Date.now(), visibility: 'public'})
   }).then(r => r.json()).then(console.log);
   ```

3. Open the team selector dropdown (top right) — new team appears in the list

4. Click the new team

5. **Observe:** URL briefly shows `?team_id=<id>`, then immediately strips to `/admin/#section`

**Automated test:**
```bash
pytest tests/playwright/test_admin_url_context.py::TestAdminIframeContext::test_iframe_team_selector_onclick_stripped_but_delegation_works -v
```

**Expected result:** URL retains `?team_id=<id>` and team selector displays the selected team  
**Actual result:** URL loses `team_id` parameter, selector shows "All Teams"

---

### 📓 Test Failure Output

```
FAILED tests/playwright/test_admin_url_context.py::TestAdminIframeContext::test_iframe_team_selector_onclick_stripped_but_delegation_works[chromium]

AssertionError: Expected team_id in iframe URL after clicking team, 
got: http://localhost:4444/proxy/mcp/admin#gateways

assert 'team_id=' in 'http://localhost:4444/proxy/mcp/admin#gateways'
```

**Test results:** 1 failed, 758 passed, 163 skipped (FYRE VM deployment)

---
### 🧠 Environment

| Key | Value |
|-----|-------|
| Version | `main` at commit `b6a41777` |
| Platform | Ubuntu 24.04 LTS |
| Deployment | Ansible to FYRE VM |
---

Yosiefeyob:
# Pull Request

## 🔗 Related Issue

Closes #5414

---

## 📝 Summary

Stabilizes the iframe team-selector Playwright test
(`test_iframe_team_selector_onclick_stripped_but_delegation_works`),
which was failing intermittently for **two unrelated reasons** both
surfacing through the same test:

1. **`team-selector.js` strips valid `team_id`s from the URL.**
`window.USER_TEAMS_DATA` is server-rendered at page load and frozen for the page lifetime. A team created after page       load (e.g. via the REST API) is not in it, so `init()` wrongly classifies its `team_id` as invalid and removes it via `safeReplaceState`. Fix: verify with the server via `/admin/teams/ids` before stripping.

2. **`tests/helpers/auth.py` default JWT expiry is 30 min.**
The session-scoped `api_request_context` Playwright fixture mints one token at suite startup and reuses it for every test. The full Playwright suite takes >60 min, so the token expires partway through.
   **Fix:** bump default expiry to 180 min in the three jwt helper functions.

---

## 📏 Reviewability

- [x] This PR has one clear purpose (stabilizing the iframe team-selector test)

---

## 🏷️ Type of Change

- [x] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

| Check | Command | Status |
|---|---|---|
| Coverage ≥ 80% | `make coverage` | ✅ |
| Iframe test isolation stress test (post-fix) | `pytest tests/playwright/test_admin_url_context.py::TestAdminIframeContext::test_iframe_team_selector_onclick_stripped_but_delegation_works --browser chromium` (×72 runs) | **72/72 passed** |
| Full Playwright suite | `make test-ui-headless` (on FYRE VM) | ✅ 763 passed, 0 failed, 159 skipped (1h 49m) |
| Server log post-fix | `grep "TokenValidationError" serve.log` | **0 occurrences** (was many before) |

---

## ✅ Checklist

---

### Why bundle these two fixes
Both surface as failures of the same Playwright test. The JS fix alone
reduces the flake but doesn't eliminate it — the second bug only fires
in long suite runs where the session JWT has since expired.

### Why the JWT-expiry fix is needed even though ~30 other tests use the same fixture
Most of those tests use `api_request_context` only inside cleanup `try/finally` blocks that silently swallow 401s — when the JWT expires mid-suite, the cleanup just doesn't run and the test still reports as passing. The iframe team-selector test is one of the few that uses the fixture for **required setup** (creating the team the test acts on):

```python
resp = api_request_context.post("/teams/", data={...})
assert resp.status < 400, f"Failed to create test team: HTTP {resp.status}"

### CODE REVIEW

**Comment by marekdano** [comment]
@Yosiefeyob 

 **Correct approach, one gap.**

The async server-check before stripping is the right fix. The endpoint (`/admin/teams/ids`) exists, requires auth, and returns `{ team_ids: [...], count: N }`, which the JS correctly reads.

**Missing `.catch()` handler — blocking:**

  ```js
  fetch(rootPath + '/admin/teams/ids', { credentials: 'same-origin' })
    .then(function (resp) { return resp.ok ? resp.json() : null; })
    .then(function (data) { ... });
    // ← no .catch()
  ```

A network error (transient failure, server restart) produces an unhandled promise rejection. The `resp.ok = false` case is handled correctly (returns `null` → `ids = []` → URL not stripped), but a thrown network error is a different code path. Add:

  ```js
  .catch(function () { /* fail open: don't strip */ });
  ```

**Minor style note (not blocking):** The surrounding code and Alpine.js both target ES6+. `ids.indexOf(...) === -1` and `function` declarations are dated; `!ids.includes(requestedTeamId)` with an arrow function would be idiomatic.

**Endpoint scope for non-admins:** `/admin/teams/ids` for non-admins only returns teams the user is a _member_ of - not public teams they can see but aren't a member of. If a publicly visible team is selected, this fix may still incorrectly strip the `team_id`. Worth a follow-up issue if that's a realistic flow.

**Review by marekdano — DISMISSED** [review]
LGTM 🚀

**Review by msureshkumar88 — DISMISSED** [review]
## Review summary

Small, well-scoped fix: `team-selector.js` now asks the server (`GET /admin/teams/ids`) before stripping a `team_id` from the URL, instead of trusting the page-load-frozen `window.USER_TEAMS_DATA` cache — fixes the flake where a team created after page load got its valid `team_id` wrongly stripped. Fails open on fetch error (doesn't strip). `tests/helpers/auth.py` JWT default expiry bumped 30→180 min to stop the session-scoped fixture from expiring mid-suite. `.secrets.baseline` diff is just the mechanical line-number/hash update for the new `fetch(...)` string — no actual secret.

Checked the diff against `mcpgateway/admin.py`'s `/admin/teams/ids` handler and the other `admin_ui/*.js` modules that read `team_id` from the URL for form scoping (e.g. `formSubmitHandlers.js`) — those POSTs are validated server-side via `_validated_team_id_param`, so the async window before the strip resolves isn't a new authorization gap, just UI hygiene. No CWE-level issue found. No unrelated changes, no dead code, issue #5414 is linked, not a breaking change.

## E2E verification (real run, not just static review)

Checked out the PR branch (`pr-5444-verify` = `origin/pull/5444/head`, commit `44a6ddb2b`) into an isolated worktree, confirmed the diff there matches exactly the 3 files above.

**Config:** `.env` from `.env.example` with `AUTH_REQUIRED=true`, `JWT_SECRET_KEY` set, `PORT=4499`, `DATABASE_URL=sqlite:///./verify5444.db`. Ran `make build-ui` (team-selector.js is Vite-bundled into the shipped `bundle-*.js`, so the source diff alone isn't what serves to the browser) then `python -m uvicorn mcpgateway.main:app --host 0.0.0.0 --port 4499`; `/health` → 200.

**JWT expiry check:** `make_test_jwt`, `make_legacy_test_jwt`, `make_auth_header_for_email` all decode to `exp - iat = 180.0 min` with no override — confirms the bump.

**Playwright e2e (chromium), driving the real shipped code via `page.evaluate` against the mounted Alpine component:**
- Fixed-behavior test: loaded `/admin` (froze `USER_TEAMS_DATA`), created a team via `POST /teams` *after* that load, set `?team_id=<new>`, re-ran `init()`. `/admin/teams/ids` fetch fired = `True`, URL retained `team_id=<new>` after resolve = `True`.
- Fail-open test: aborted `**/admin/teams/ids*` via `page.route` with a bogus `team_id`. URL still retained the bogus param (not stripped) = `True`.
- Regression proof: swapped in the pre-PR `team-selector.js` (`origin/main`), rebuilt, reran the same flow — old code stripped `team_id` and never called `/admin/teams/ids` at all. Confirms the test isn't tautological and the bug was real.

`OVERALL: PASS` on all three assertions. Cleaned up (test team deleted, server stopped, temp db removed, worktree restored to PR state).

## Non-blocking suggestions
- Worth adding a permanent Playwright/unit test for this exact path (the one used for verification here isn't in the diff) — the existing `test_iframe_team_selector_onclick_stripped_but_delegation_works` goes through a full page reload, so it doesn't actually exercise the new async-fetch branch.
- The `.catch(() => {})` on the `/admin/teams/ids` fetch swallows all errors silently — a `console.warn` there would help debug a misconfigured-permission or endpoint-down scenario in production instead of it looking like nothing happened.

Approving — the fix works as described and is verified against a real regression reproduction, not just reasoning about the diff.

**Review by msureshkumar88 — APPROVED** [review]
LGMT

**Review by msureshkumar88 — CHANGES_REQUESTED** [review]
Nice fix for the iframe flakiness — deferring the team_id strip to a server-side check makes sense given the cache-freshness issue.

**Blocking:** the CI vitest job is failing because `init()` now resolves the strip asynchronously via `fetch('/admin/teams/ids')`, but `tests/unit/js/team-selector.test.js` (not touched in this PR) still has a synchronous assertion right after `component.init()`:

```js
component.init();
expect(safeReplaceState).toHaveBeenCalledOnce();  // fails — fetch hasn't resolved yet
```

Since `fetch` isn't mocked in that test, the call likely rejects/fails in jsdom, hits the new `.catch(() => {})` fail-open branch, and `safeReplaceState` never gets called.

Could you update that test (`init — team_id present in URL > calls Admin.safeReplaceState removing team_id when id not found in non-empty teams`) to mock `window.fetch` and await the resolution (e.g. `vi.waitFor(...)`) before asserting? Once that's in, CI should be green.

ja8zyjits:
## 🔧 Recommended Fixes

- [ ] **Null/undefined handling in team-selector.js**
  - File: `mcpgateway/admin_ui/components/team-selector.js:31`
  - Current: `const ids = (data && data.team_ids) || [];`
  - Issue: If `data.team_ids` exists but not array (null/undefined/wrong type), `.includes()` throws
  - Fix: `const ids = Array.isArray(data?.team_ids) ? data.team_ids : [];`
  - Severity: Medium (fail-open mitigates, but error silent)

## ⚠️ Warnings

- [ ] **JWT helper expiry change (30→180 min) not documented**
  - Files: `tests/helpers/auth.py:36,90,133`
  - Code: `expires_in_minutes: int = 180` (was 30)
  - Impact: Test suite token lifetime for long Playwright runs
  - Fix: Consider documenting in `tests/AGENTS.md` or `TESTING.md`
  - Note: Internal test infrastructure, may not need user-facing docs

marekdano (dismissed):
LGTM 🚀

ja8zyjits (dismissed):
LGTM

msureshkumar88 (approved):
All requested changes addressed:

- Missing `.catch()` on the `/admin/teams/ids` fetch — added (fail-open on network error).
- Modernized to arrow functions and `.includes()`.
- `Array.isArray(data?.team_ids)` guard added for the null/wrong-type case.
- JWT expiry bump (30→180 min) documented in `tests/AGENTS.md`.
- `tests/unit/js/team-selector.test.js` updated to mock `window.fetch` and `await vi.waitFor(...)` before asserting — vitest CI is green.

The non-admin `/admin/teams/ids` scope note was correctly deferred as a non-blocking follow-up.

Verified against the diff; no new issues found.

PR #5416, opened 2026-06-29
update python versions

cafalchio:
# Pull Request

## 🔗 Related Issue

Closes #5413

---

## 📝 Summary

_What does this PR do and why?_

Update Python package dependencies to their latest compatible versions.

pyproject.toml and lock files updated

---

## 📏 Reviewability

- [x] This PR has one clear purpose
- [x] The linked issue is not labeled `triage`
- [x] Unrelated bugs or improvements are tracked in separate issues/PRs
- [x] Tests are included with the code they validate
- [x] If AI-assisted, I understand and can explain the generated changes

---

## 🏷️ Type of Change

- [ ] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [x] Chore (deps, CI, tooling)
- [ ] Other (describe below)

---

## 🧪 Verification

_List exact commands, screenshots, videos, logs, reproduction steps, or manual validation. If evidence is not feasible, explain why._

| Check                     | Command         | Status |
|---------------------------|-----------------|--------|
| Lint suite                | `make lint`     |   x    |
| Unit tests                | `make test`     |   x    |
| Coverage ≥ 80%            | `make coverage` |   x    |

---

## ✅ Checklist

- [ ] Code formatted (`make black isort pre-commit`)
- [ ] Tests added/updated for changes
- [ ] Documentation updated (if applicable)
- [ ] No secrets or credentials committed

---

## 📓 Notes (optional)

_Screenshots, design decisions, or additional context._

PR #5447, opened 2026-06-30
Fix/fastapi 0.137 included router path

### LINKED ISSUE #5445: bug: test_api_v1.py and test_well_known.py fail with FastAPI 0.137.x (_IncludedRouter has no attribute 'path')
## Summary

`tests/unit/mcpgateway/test_api_v1.py` and `tests/unit/mcpgateway/routers/test_well_known.py` (added in #4403) fail on FastAPI 0.137.x with:

```
AttributeError: '_IncludedRouter' object has no attribute 'path'
```

## Root Cause

FastAPI **0.137.0** introduced `_IncludedRouter` — a new lazy wrapper object that is appended to `router.routes` when `include_router()` is called, instead of eagerly expanding sub-routes into `APIRoute` objects as in ≤ 0.136.x.

The helper in `test_api_v1.py`:

```python
def _route_paths(router: APIRouter) -> list[str]:
    return [r.path for r in router.routes]   # line 78
```

…iterates `router.routes` and blindly accesses `.path`. In 0.137.x, included sub-routers appear as `_IncludedRouter` instances which expose `original_router` and `include_context` but **not** `.path`, causing all 19 tests to fail.

## Failing Tests

All 19 failures share the same root cause:

- `tests/unit/mcpgateway/test_api_v1.py` — 18 tests across `TestBuildV1RouterGroupA/B/C`
- `tests/unit/mcpgateway/routers/test_well_known.py::test_v1_prefix_does_not_produce_rfc_violation`

## Fix

`_route_paths()` needs to handle `_IncludedRouter` by recursing into `r.original_router`:

```python
def _route_paths(router: APIRouter) -> list[str]:
    from fastapi.routing import _IncludedRouter
    paths = []
    for r in router.routes:
        if isinstance(r, _IncludedRouter):
            paths.extend(_route_paths(r.original_router))
        else:
            paths.append(r.path)
    return paths
```

madhu-mohan-jaishankar:
## 🔗 Related Issue

Closes #5445

---

## 📝 Summary

FastAPI 0.137.0 changed how `include_router()` works internally: instead of
eagerly expanding sub-router routes into `APIRoute` objects (with the full
prefix already embedded in `route.path`), it now lazily stores a
`_IncludedRouter` wrapper object in `router.routes`. These wrapper objects
expose `original_router` and `include_context` but **not** `.path`, so any
code that iterates `router.routes` and accesses `.path` directly raises:

```
AttributeError: '_IncludedRouter' object has no attribute 'path'
```

A second, subtler change: `dependencies` passed to `include_router()` are no
longer copied onto each leaf route — they are stored on
`_IncludedRouter.include_context.dependencies` and applied lazily. Any test
inspecting `route.dependencies` for include-level guards silently misses them
on FastAPI ≥ 0.137.

FastAPI 0.137.2 was pulled in transitively by
`prometheus-fastapi-instrumentator>=8.0.1`.

**Changes:**

- **`tests/unit/mcpgateway/_helpers.py`** (new) — single `collect_routes(router)`
  helper that handles both FastAPI versions. Returns `(path, route, include_deps)`
  triples where:
  - `path` — fully-qualified route path (prefix-accumulation aware)
  - `route` — the leaf `APIRoute` object
  - `include_deps` — `Depends` objects accumulated from enclosing
    `_IncludedRouter` wrappers (empty on FastAPI ≤ 0.136 where they are
    already copied onto the route itself)

- **`test_api_v1.py`** / **`test_legacy_router.py`** — replaced per-file inline
  traversal helpers with `collect_routes`.

- **`test_well_known.py`** — replaced per-file inline traversal helper with
  `collect_routes`.

- **`test_api_versioning_parity.py`** — fixed two tests added to `main` after
  the branch was cut:
  - `test_legacy_prefixes_match_assembled_routers` — path iteration now uses
    `collect_routes`.
  - `test_llm_admin_router_csrf_dependency` — dependency check now reads from
    both `include_deps` (FastAPI ≥ 0.137) and `route.dependencies`
    (FastAPI ≤ 0.136) to cover both versions.

No production code was changed; this is a pure test-compatibility fix.

---

## 📏 Reviewability

---

## 🏷️ Type of Change

- [x] Bug fix

---

## 🧪 Verification

Tests were run against **both** FastAPI 0.136.3 (local venv) and 0.137.2
(CI-resolved version) to confirm forward and backward compatibility.

| Check | Command | Status |
|---|---|---|
| Unit tests (FastAPI 0.136.3) | `python3 -m pytest tests/unit/mcpgateway/test_api_v1.py tests/unit/mcpgateway/test_legacy_router.py tests/unit/mcpgateway/routers/test_well_known.py tests/unit/mcpgateway/test_api_versioning_parity.py -q` | ✅ 119/119 passed |
| Lint suite | `make lint` | pending CI |
| Full unit tests | `make test` | pending CI |
| Coverage ≥ 80% | `make coverage` | pending CI |

---

## ✅ Checklist

---

## 📓 Notes (optional)

**Why a shared `_helpers.py` instead of per-file helpers?**
All four affected test files needed the same `_IncludedRouter`-aware traversal.
Duplicating it would mean four places to update if FastAPI's internals change
again. A single helper with a clear public contract is easier to maintain and
audit.

**Why return 3-tuples `(path, route, include_deps)`?**
On FastAPI ≥ 0.137, `dependencies` passed to `include_router()` live on the
`_IncludedRouter.include_context.dependencies`, not on the leaf route. Carrying
them through the traversal as `include_deps` lets callers check the full
effective dependency set without knowing about `_IncludedRouter` at all.
Callers that only need paths use `[p for p, *_ in collect_routes(router)]`.

**Prefix accumulation detail:**
In FastAPI 0.137.x, the top-level `_IncludedRouter.include_context.prefix`
already incorporates the parent router's own prefix (e.g. `/v1`), so
accumulation starts from `include_context.prefix` directly. Nested
`_IncludedRouter` objects inside `original_router` carry only their incremental
prefix, which `_collect()` appends to the running total.

### DIFF
--- a/tests/helpers/router_helpers.py
+++ b/tests/helpers/router_helpers.py
@@ -1,0 +1,74 @@
+# -*- coding: utf-8 -*-
+"""Location: ./tests/helpers/router_helpers.py
+Copyright 2026
+SPDX-License-Identifier: Apache-2.0
+
+FastAPI router introspection helpers shared across the test suite.
+
+These utilities abstract over the ``_IncludedRouter`` lazy-wrapper introduced
+in FastAPI 0.137.0 so that tests work correctly against both FastAPI ≤ 0.136
+(eager route expansion) and ≥ 0.137 (lazy ``_IncludedRouter`` wrappers).
+"""
+
+from __future__ import annotations
+
+
+def collect_routes(router) -> list[tuple[str, object, list]]:
+    """Return ``(full_path, route, include_deps)`` triples for every leaf route on *router*.
+
+    ``include_deps`` is the list of :class:`fastapi.params.Depends` objects
+    accumulated from enclosing ``_IncludedRouter`` wrappers (FastAPI ≥ 0.137).
+    On FastAPI ≤ 0.136 this list is always empty because ``include_router``
+    copies dependencies directly onto each leaf route's ``.dependencies``.
+
+    This covers both FastAPI versions for tests that inspect route dependencies:
+
+    * FastAPI ≤ 0.136: check ``getattr(route, "dependencies", [])``
+    * FastAPI ≥ 0.137: check ``include_deps``
+
+    To get paths only: ``[p for p, *_ in collect_routes(router)]``.
+
+    Args:
+        router: Any ``fastapi.APIRouter`` instance.
+
+    Returns:
+        A list of ``(path_str, route_object, include_deps)`` triples.
+    """
+    try:
+        from fastapi.routing import _IncludedRouter  # type: ignore[attr-defined]
+    except ImportError:
+        _IncludedRouter = None  # type: ignore[assignment,misc]
+
+    if _IncludedRouter is None:
+        # FastAPI ≤ 0.136: include_router eagerly expands routes; paths already
+        # contain the fully-qualified prefix and deps are on the route itself.
+        return [(r.path, r, []) for r in router.routes]
+
+    # FastAPI 0.137+: include_router stores lazy _IncludedRouter wrappers.
+    # Accumulate include_context.dependencies as we descend so callers see the
+    # full effective dependency set without knowing about _IncludedRouter.
+    def _collect(routes: list, prefix: str, acc_deps: list) -> list[tuple[str, object, list]]:
+        result: list[tuple[str, object, list]] = []
+        for r in routes:
+            if isinstance(r, _IncludedRouter):
+                combined_deps = acc_deps + list(r.include_context.dependencies or [])
+                result.extend(_collect(
+                    r.original_router.routes,
+                    prefix + r.include_context.prefix,
+                    combined_deps,
+                ))
+            else:
+                result.append((prefix + r.path, r, acc_deps))
+        return result
+
+    items: list[tuple[str, object, list]] = []
+    for r in router.routes:
+        if isinstance(r, _IncludedRouter):
+            # include_context.prefix already contains the parent router prefix.
+            # include_context.dependencies already includes the parent router deps.
+            top_deps = list(r.include_context.dependencies or [])
+            items.extend(_collect(r.original_router.routes, r.include_context.prefix, top_deps))
+        else:
+            # Direct routes already have the router's own prefix embedded.
+            items.append((r.path, r, []))
+    return items
--- a/tests/unit/mcpgateway/routers/test_well_known.py
+++ b/tests/unit/mcpgateway/routers/test_well_known.py
@@ -21,6 +21,7 @@ from mcpgateway.routers.well_known import (
     get_well_known_file_content,
     validate_security_txt,
 )
+from tests.helpers.router_helpers import collect_routes
 
 # ---------- get_base_url_with_protocol ----------
 
@@ -252,7 +253,7 @@ def test_v1_prefix_does_not_produce_rfc_violation():
     v1_router = APIRouter(prefix="/v1")
     v1_router.include_router(admin_router)
 
-    final_paths = [route.path for route in v1_router.routes]
+    final_paths = [p for p, *_ in collect_routes(v1_router)]
     rfc_violations = [p for p in final_paths if "/.well-known" in p]
     assert rfc_violations == [], f"Including admin_router in /v1 creates RFC 8615 violating paths: {rfc_violations}"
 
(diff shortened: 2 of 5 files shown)


tests/helpers/router_helpers.py
```diff
@@ -0,0 +1,73 @@
+# -*- coding: utf-8 -*-
```
ja8zyjits, line 1: there is already a folder under `/tests` with the name `helpers`
Should we not add this into that? or extend upon what exists there?

.pre-commit-config.yaml
```diff
@@ -329,7 +329,7 @@ repos:
         description: Verifies test files in tests/ directories start with `test_`.
         language: python
         files: (^|/)tests/.+\.py$
-        exclude: ^tests/(.*/)?(pages|helpers|fuzzers|scripts|fixtures|migration|utils|manual|async|load|loadtest|client|populate|targets)/.*\.py$
+        exclude: ^tests/(.*/)?(pages|helpers|fuzzers|scripts|fixtures|migration|utils|manual|async|load|loadtest|client|populate|targets)/.*\.py$|^tests/.*/_helpers\.py$
```
ja8zyjits: can you document this exclusion in `/tests/AGENTS.md` ?

ja8zyjits (changes requested):

gandhipratik203:
Thanks for the compatibility fix. I noticed the FastAPI bump came from the broad dependency refresh in #5416: before that, `uv.lock` had `fastapi 0.136.3`; after #5416, it resolves to `fastapi 0.137.2`, while `pyproject.toml` still only says `fastapi>=0.136.3`.

Was the move to FastAPI 0.137 intentional? If not, could we cap it for now as `fastapi>=0.136.3,<0.137.0` and regenerate `uv.lock`? That keeps us on the previously supported FastAPI line and avoids taking the `_IncludedRouter` behavior change until we intentionally upgrade.

`prometheus-fastapi-instrumentator>=8.0.1` can stay, since that was a separate fix; the FastAPI version should be controlled by our direct FastAPI constraint.

jonpspri (commented):
-

jonpspri (approved):
LGTM.  Jitesh's comment was addressed.

ja8zyjits (approved):
LGTM

msureshkumar88:
## E2E verification (real environment, no mocks)

Ran a before/after e2e script against the exact FastAPI version CI resolves for this PR, in an isolated git worktree.

**Environment**
- Python 3.12.3
- FastAPI **0.137.2** (pulled transitively via `prometheus-fastapi-instrumentator>=8.0.1`, same version noted in the PR description as CI-resolved)

**Commands**
```bash
# AFTER — PR head commit (0817551e2) as-is
python -m pytest tests/unit/mcpgateway/test_api_v1.py tests/unit/mcpgateway/test_legacy_router.py \
  tests/unit/mcpgateway/routers/test_well_known.py tests/unit/mcpgateway/test_api_versioning_parity.py -v

# BEFORE — same fastapi 0.137.2, but with the pre-fix versions of these 4 test files
# (checked out from the parent commit 96c80148e) and router_helpers.py removed
git checkout 96c80148e -- tests/unit/mcpgateway/routers/test_well_known.py tests/unit/mcpgateway/test_api_v1.py \
  tests/unit/mcpgateway/test_api_versioning_parity.py tests/unit/mcpgateway/test_legacy_router.py
git rm -f tests/helpers/router_helpers.py
python -m pytest <same 4 files> -v

# restore
git reset --hard 0817551e2
```

**Results**

| State | Result |
|---|---|
| BEFORE (pre-fix code, real fastapi 0.137.2) | **79 failed, 40 passed** — `AttributeError: '_IncludedRouter' object has no attribute 'path'` (e.g. `test_api_v1.py:78` in `_route_paths`), reproducing issue #5445 exactly |
| AFTER (PR head, `0817551e2`) | **119 passed, 0 failed** |
| Re-run after restore | 119 passed, 0 failed — clean, no stray state |

This confirms the bug is real under the actual FastAPI version CI uses, and `collect_routes()` + the four call-site updates fully resolve it.

Approving based on this. Nice touch handling both FastAPI versions in one helper rather than four separate patches — makes this easy to revisit if FastAPI's internals shift again.

msureshkumar88 (approved):
E2E-verified against real FastAPI 0.137.2 in an isolated worktree: pre-fix test code reproduces the AttributeError from #5445 (79 failed), PR's fix resolves it fully (119 passed). See comment above for commands/output. LGTM.

PR #5449, opened 2026-06-30
fix: resolve _IncludedRouter AttributeError in router path extraction tests

jonpspri:
Factor out router path extraction logic into shared helper module to eliminate code duplication and fix AttributeError when accessing _IncludedRouter.path. Fixes test failures in test_v1_prefix_does_not_produce_rfc_violation, test_legacy_prefixes_match_assembled_routers, and test_llm_admin_router_csrf_dependency.

### DIFF
--- a/tests/helpers/router_helpers.py
+++ b/tests/helpers/router_helpers.py
@@ -1,0 +1,67 @@
+# -*- coding: utf-8 -*-
+"""tests/helpers/router_helpers.py
+Copyright 2025
+SPDX-License-Identifier: Apache-2.0
+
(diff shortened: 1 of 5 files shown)


madhu-mohan-jaishankar (changes requested):
Nice direction overall - one thing still missing is centralizing the dependency traversal too, not just path extraction. Right now test_api_versioning_parity.py still has custom recursive logic to collect include-level dependencies for _IncludedRouter; it would be cleaner if the shared helper exposed that as part of its API so all FastAPI 0.137 traversal/version-compat logic lived in one place.

jonpspri:
Closed in favor of #5447.

PR #5455, opened 2026-06-30
feat(api): add public GET /v1/resources/test/{resource_uri} endpoint

### LINKED ISSUE #5356: [FEATURE]: Add GET /v1/resources/test/{resource_uri} endpoint
## Summary

Add a `GET /v1/resources/test/{resource_uri}` endpoint to the public v1 API that mirrors the existing `GET /admin/resources/test/{resource_uri}` admin endpoint, allowing programmatic clients and external integrations to test/read resources without going through the admin surface.

## Current Behaviour

Resource reading via URI is only available through the admin endpoint
  `GET /admin/resources/test/{resource_uri:path}` (`admin.py`), which is scoped to the admin UI.

## Desired Behaviour

A new endpoint `GET /v1/resources/test/{resource_uri:path}` should be added to the `resource_router` (mounted at `/resources` in `main.py`) with the following behaviour:

  - Requires `resources.read` permission (no admin bypass)
  - Admin users get unrestricted access (`user=None, token_teams=None`)
  - Non-admin users get team-scoped access
  - Returns `{"content": <resolved resource content>}` on success
  - Returns `404` if the resource URI is not found

 ## Acceptance Criteria

  - [ ] `GET /v1/resources/test/{resource_uri:path}` endpoint added to `resource_router` in `main.py`
  - [ ] Behaviour is consistent with `GET /admin/resources/test/{resource_uri:path}`
  - [ ] Auth and RBAC enforced (`resources.read` permission, team scoping for non-admins)
  - [ ] Unit and integration tests added
  - [ ] OpenAPI docs updated automatically via FastAPI route registration

Lang-Akshay:
# ✨ Feature / Enhancement PR

## 🔗 Epic / Issue

Closes #5356

---

## 🚀 Summary (1-2 sentences)

Adds `GET /v1/resources/test/{resource_uri:path}` (and the legacy `/resources/test/...` prefix) to the public `resource_router`, mirroring the existing admin-only `GET /admin/resources/test/{resource_uri:path}` so programmatic clients can test/read resources by URI without going through the admin surface.

---

## 📓 Details

- New `test_resource_by_uri` handler in `mcpgateway/main.py`, guarded by `@require_permission("resources.read", allow_admin_bypass=False)`.
- Layer-1 visibility via `get_scoped_resource_access_context()`: admins get unrestricted access (`user=None, token_teams=None`), non-admins get team-scoped or public-only access.
- Returns `{"content": <resolved content>}` on success; `404` on `ResourceNotFoundError`. Admin endpoint untouched.
- Unit tests (`tests/unit/mcpgateway/test_main.py`) cover success, 404, and permission-denial paths.
- Integration tests (`tests/integration/test_resource_management.py`) exercise the full ASGI middleware stack (auth, RBAC, routing) for both the legacy and `/v1` prefixes.
- `tests/e2e/test_main_apis.py`: delete-verification now uses ID-based lookup, since URI `test/delete` routes to the new `/test/{path}` endpoint (which returns 403 under admin bypass, not 404).

---

## 📏 Reviewability

---

## 🧪 Checks

- [ ] `make lint` passes
- [ ] `make test` passes

### DIFF
--- a/mcpgateway/main.py
+++ b/mcpgateway/main.py
@@ -6149,7 +6149,7 @@ async def create_resource(
 
 
 @resource_router.get("/test/{resource_uri:path}")
-@require_permission("resources.read")
+@require_permission("resources.read", allow_admin_bypass=False)
 async def test_resource_by_uri(
     resource_uri: str,
     request: Request,
@@ -6174,6 +6174,8 @@ async def test_resource_by_uri(
     auth_user_email, auth_token_teams = get_scoped_resource_access_context(request, user)
     try:
         resource_content = await resource_service.read_resource(db, resource_uri=resource_uri, user=auth_user_email, token_teams=auth_token_teams)
+        db.commit()
+        db.close()
         return {"content": resource_content}
     except ResourceNotFoundError as e:
         raise HTTPException(status_code=404, detail=str(e))
--- a/mcpgateway/main.py
+++ b/mcpgateway/main.py
@@ -6173,9 +6173,7 @@ async def test_resource_by_uri(
     logger.debug("Reading resource by URI %s for user %s", resource_uri, safe_log_user(user))
     auth_user_email, auth_token_teams = get_scoped_resource_access_context(request, user)
     try:
-        resource_content = await resource_service.read_resource(
-            db, resource_uri=resource_uri, user=auth_user_email, token_teams=auth_token_teams
-        )
+        resource_content = await resource_service.read_resource(db, resource_uri=resource_uri, user=auth_user_email, token_teams=auth_token_teams)
         return {"content": resource_content}
     except ResourceNotFoundError as e:
         raise HTTPException(status_code=404, detail=str(e))
--- a/mcpgateway/main.py
+++ b/mcpgateway/main.py
@@ -6148,6 +6148,42 @@ async def create_resource(
         raise HTTPException(status_code=415, detail={"error": "Unsupported Media Type", "message": str(e), "mime_type": e.mime_type, "allowed_types": e.allowed_types})
 
 
+@resource_router.get("/test/{resource_uri:path}")
+@require_permission("resources.read")
+async def test_resource_by_uri(
+    resource_uri: str,
+    request: Request,
+    db: Session = Depends(get_db),
+    user=Depends(get_current_user_with_permissions),
+) -> Dict[str, Any]:
+    """Read a resource by URI and return its content.
+
+    Args:
+        resource_uri (str): URI of the resource to read.
+        request (Request): FastAPI request object for context.
+        db (Session): Database session.
+        user: Authenticated user with permissions.
+
+    Returns:
+        Dict[str, Any]: Dictionary with a ``content`` key containing the resolved resource content.
+
+    Raises:
+        HTTPException: 404 if the resource is not found or not accessible to the caller.
+    """
+    logger.debug("Reading resource by URI %s for user %s", resource_uri, safe_log_user(user))
+    auth_user_email, auth_token_teams = get_scoped_resource_access_context(request, user)
+    try:
+        resource_content = await resource_service.read_resource(
+            db, resource_uri=resource_uri, user=auth_user_email, token_teams=auth_token_teams
+        )
+        return {"content": resource_content}
+    except ResourceNotFoundError as e:
+        raise HTTPException(status_code=404, detail=str(e))
+    except Exception as e:
+        logger.error("Error reading resource by URI %s: %s", resource_uri, e)
+        raise
+
+
 @resource_router.get("/{resource_id}")
 @require_permission("resources.read")
 async def read_resource(resource_id: str, request: Request, db: Session = Depends(get_db), user=Depends(get_current_user_with_permissions)) -> Any:
--- a/tests/e2e/test_main_apis.py
+++ b/tests/e2e/test_main_apis.py
@@ -1150,8 +1150,9 @@ class TestResourceAPIs:
         assert response.status_code == 200
         assert response.json()["status"] == "success"
 
-        # Verify it's deleted
-        response = await client.get(f"/resources/{resource_data['resource']['uri']}", headers=TEST_AUTH_HEADER)
+        # Verify it's deleted (use ID-based lookup; URI "test/delete" routes to
+        # /test/{path} which has allow_admin_bypass=False and returns 403 not 404)
+        response = await client.get(f"/resources/{resource_id}", headers=TEST_AUTH_HEADER)
         assert response.status_code == 404
 
     # API should probably return 409 instead of 400 for non-existent resource
--- a/tests/integration/test_resource_management.py
+++ b/tests/integration/test_resource_management.py
@@ -3,4 +3,269 @@
 Copyright 2026
 SPDX-License-Identifier: Apache-2.0
 Authors: Mihai Criveti
+
+Integration tests for the public resource-by-URI endpoint.
+
+Covers the full ASGI middleware stack (auth, RBAC, routing) for:
+  GET /resources/test/{resource_uri:path}  (legacy prefix)
+  GET /v1/resources/test/{resource_uri:path}  (canonical v1 prefix)
+
+Issue #5356 acceptance criteria requires both unit **and** integration tests.
 """
+
+# Future
+from __future__ import annotations
+
+# Standard
+import os
+import tempfile
+from unittest.mock import AsyncMock, MagicMock, patch
+
+# Third-Party
+from _pytest.monkeypatch import MonkeyPatch
+from fastapi.testclient import TestClient
+import pytest
+from sqlalchemy import create_engine
+from sqlalchemy.orm import sessionmaker
+from sqlalchemy.pool import StaticPool
+
+# First-Party
+from mcpgateway.main import app
+from mcpgateway.utils.verify_credentials import require_auth
+from mcpgateway.auth import get_current_user
+from mcpgateway.middleware.rbac import get_current_user_with_permissions
+from mcpgateway.middleware.rbac import get_db as rbac_get_db
+from mcpgateway.middleware.rbac import get_permission_service
+
+# ---------------------------------------------------------------------------
+# Local helpers
+# ---------------------------------------------------------------------------
+
+
+class _PermissionServiceAlwaysGrant:
+    """Minimal permission-service stand-in that grants every check.
+
+    Uses ``**kwargs`` so it stays compatible with the full
+    ``PermissionService.check_permission`` signature (which includes
+    ``token_teams``, ``allow_admin_bypass``, ``check_any_team``, etc.)
+    without needing to mirror every parameter explicitly.
+    """
+
+    def __init__(self, *args, **kwargs):
+        pass
+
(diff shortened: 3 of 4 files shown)


marekdano (commented):
### pylint W0707 (`main.py`)

Missing exception chaining:

  ```python
  except ResourceNotFoundError as e:
      raise HTTPException(status_code=404, detail=str(e))        # ← current
      raise HTTPException(status_code=404, detail=str(e)) from e # ← suggested
  ```

Minor and non-blocking, but worth fixing for a green `make pylint`. Note the codebase is inconsistent here: the sibling by-ID `read_resource` does it correctly (`from exc`, `main.py`), while the admin endpoint this was copied from also omits it

marekdano (approved):
Nice, well-scoped PR. Thanks for mirroring the existing admin endpoint so closely. The two-layer model is correctly applied (`get_scoped_resource_access_context` for Layer-1 visibility, `@require_permission("resources.read", allow_admin_bypass=False)` for Layer-2), and I actually like that this routes non-admin callers through the scoping helper so they get proper team scoping rather than the admin endpoint's `token_teams=None`. 

Route ordering (`/test/{resource_uri:path}` before `/{resource_id}`) is right, and the test coverage is thorough - unit + integration across both the legacy and `/v1` prefixes, including the 401/403 deny paths. All the new tests pass locally for me.

LGTM 🚀

gandhipratik203 (approved):
Nice work reusing the existing scoping check instead of writing new auth logic, and the test coverage across success/404/401/403 is solid.

msureshkumar88:
## E2E Verification

I stood up a real running instance (`uvicorn mcpgateway.main:app`, real SQLite DB in a scratch dir, real JWT auth, real RBAC role/team seeding) and drove it over HTTP with curl. Script: `scripts/verify-pr-5455-e2e.sh`.

**Config used**: `AUTH_REQUIRED=true`, `MCPGATEWAY_ADMIN_API_ENABLED=true`, `EMAIL_AUTH_ENABLED=true`, `PLUGINS_ENABLED=false`, fresh SQLite DB per run, gateway on `127.0.0.1:44559`.

**Setup**: `python -m mcpgateway.bootstrap_db` (schema + platform admin + built-in RBAC roles), then a seeding step using `TeamManagementService`/`RoleService` to create two real teams, a `developer`-role user per team, a custom global-scope `resources.read`-only role for a "public-only" persona, and a no-permission persona — then minted real JWTs per persona (admin / team-A / team-B / public-only / no-permission) and created real resources (public, team-A-private, team-B-private, plus a `test/probe` URI for the routing-collision check) via `POST /resources`.

**Results — 20/20 checks passed** (full log available on request; summary below):

| # | Check | Result |
|---|---|---|
| 1 | Admin GET `/v1/resources/test/{uri}` on a real resource → 200 with matching content | PASS |
| 2 | Legacy `GET /resources/test/{uri}` → 200, byte-identical content to `/v1` (parity) | PASS |
| 3 | `GET /v1/resources/test/{nonexistent-uri}` → 404 | PASS |
| 4 | Authenticated caller lacking `resources.read` → 403; fully unauthenticated → 401 | PASS |
| 5 | Team-A-scoped token reading team-B-private resource → 404 (not 403 — no existence leak) | PASS |
| 5b | Public-only token (`teams: []`, has `resources.read` globally) reading a team-private resource → 404; reading the public resource → 200 | PASS |
| 6 | Admin token reads both team-A- and team-B-private resources (admin bypass) | PASS |
| 7 | `/admin/resources/test/{uri}` content == `/v1/resources/test/{uri}` content for the same token+resource | PASS |
| 8 | `GET /resources/test/probe` (no-perm token) → 403, proving the new `/test/{path}` route intercepts URIs starting with `test/` rather than silently 404ing via an unmatched route; after delete, admin GET → 404 | PASS |

I also ran the PR's own new/changed tests for real:
- `uv run pytest tests/unit/mcpgateway/test_main.py -k resource_by_uri -v` → **11/11 passed**
- `uv run pytest tests/integration/test_resource_management.py -v --with-integration` → **7/7 passed**
- `uv run pytest tests/e2e/test_main_apis.py -k test_delete_resource -v` → **2/2 passed**

One iteration note on the script itself (not an app bug): my first pass minted a fresh platform-admin whose `password_change_required` flag (set by `bootstrap_db`) made `AdminAuthMiddleware` 303-redirect every `/admin/*` request — unrelated platform policy, fixed in the script by clearing the flag after bootstrap so the `/admin` vs `/v1` parity check could actually reach the route.

## Code Review

Scope: `mcpgateway/main.py`, `tests/unit/mcpgateway/test_main.py`, `tests/integration/test_resource_management.py`, `tests/e2e/test_main_apis.py` (the full `git diff main...HEAD` for this branch).

Nice to see this shipped with unit **and** integration coverage for both prefixes, explicit RBAC deny-path tests (401 and 403), and a routing-collision-aware fix to the pre-existing e2e delete test rather than papering over it. The implementation reuses `get_scoped_resource_access_context()` rather than re-deriving scoping logic, which is exactly the pattern this repo asks for.

### Blocking

None found. I didn't find a correctness bug, a security bypass, or an unjustified breaking change in the diff.

### Suggestions (non-blocking)

1. **Routing-collision blast radius is broader than just this endpoint's own tests** (`mcpgateway/main.py:6151`, registered before `@resource_router.get("/{resource_id}")` at `main.py:6187`). Any resource whose URI starts with a `test/` path segment (e.g. `test/anything`) is now intercepted by `/test/{resource_uri:path}` when accessed as `GET /resources/test/anything` — permission-gated with `allow_admin_bypass=False` — instead of ever reaching `/{resource_id}`. In practice this was already a coincidental non-path before the PR (`resource_id` doesn't match multi-segment paths), so I don't think this is a regression, and my e2e check #8 confirms the new behavior is well-defined (403 for no permission, 404 once truly absent) rather than silently mis-routing. Might still be worth a one-line callout in the PR description or a docstring note on `test_resource_by_uri` that `test/` is now a reserved first path segment under `/resources`, so it's discoverable next time someone adds a route here.

2. **Duplication with `admin_test_resource`** (`mcpgateway/admin.py:13043-13089`). The two handlers are nearly line-for-line identical (decorator, try/except shape, 404 mapping). Worth considering a small shared helper (e.g. `_read_resource_by_uri(db, resource_uri, user_email, token_teams) -> Dict[str, Any]`) that both call, to keep the 404-mapping and logging behavior from drifting apart over time. Not required for this PR — just a place I'd look first if either handler needs to change again. Small worthwhile note: unlike this PR's endpoint, `admin_test_resource` passes `token_teams=None` unconditionally for non-admin callers rather than using `get_scoped_resource_access_context()` — this PR's endpoint doesn't literally mirror that specific line, and that's arguably the more correct choice, but it means "mirrors the admin endpoint" is true structurally rather than byte-for-byte in the scoping call.

3. **Docs/OpenAPI**: the route inherits the router's `tags=["Resources"]` and its docstring becomes the OpenAPI description, so it'll show up fine in the generated API docs. I didn't see a mention in `docs/` (e.g. an API reference page) or `CHANGELOG.md` — probably fine given this repo's docs are largely auto-generated from the OpenAPI schema, just flagging in case there's a hand-maintained page that lists public read endpoints.

4. **Test coverage gap**: the integration suite mocks `resource_service.read_resource` and stubs `PermissionService.check_permission` to always grant, so it verifies routing/auth-wiring but not the actual DB-backed visibility enforcement end-to-end. The unit tests do assert that the endpoint passes the *correct* `token_teams`/`user` values through for admin vs. non-admin callers, which is the part that's actually new here — so this isn't a hole in what matters for this diff, but a DB-backed integration test with a real team-private resource (what my e2e script above exercises) would make the "no cross-team leakage" guarantee self-evident from the test suite alone, without needing a script.

5. **Minor**: `test_resource_by_uri` calls `db.commit()`/`db.close()` on the success path inside the handler (`main.py:6177-6178`) even though `get_db()`'s own generator already manages commit/rollback/close for the request-scoped session. This mirrors the exact pattern already used by `create_resource`/`read_resource` above it in the same file, so it's consistent with the surrounding code rather than a new problem — just flagging since it's a bit of a footgun pattern (double-commit/close) if it's ever copy-pasted into a handler that doesn't have `get_db`'s defensive generator underneath it.

### Security / CWE review

- **CWE-863 (incorrect authorization)**: `@require_permission("resources.read", allow_admin_bypass=False)` matches the admin endpoint exactly, and `PermissionService.check_permission` still honors a real `platform_admin`/wildcard role grant even with bypass disabled — verified this holds in check #6 (admin reads both team-private resources) and check #4 (no-permission caller correctly denied) above.
- **CWE-200 (information exposure)**: verified the 403-vs-404 distinction doesn't leak resource existence to unauthorized callers — permission is checked *before* the resource is looked up (403 regardless of whether the URI exists), and `resource_service._check_resource_access()` returns a generic `ResourceNotFoundError` (→404) for both "doesn't exist" and "exists but not visible to you" (`resource_service.py:2499-2502`) — confirmed empirically in check #5 (team-A token vs. team-B-private resource → 404, not 403).
- **CWE-22 (path traversal)**: the `:path` converter lets `resource_uri` contain `/` and `../`-like sequences, but the value is only ever used as an exact-match SQL lookup key (`DbResource.uri == str(resource_uri)`, `resource_service.py:2447`) or passed to `invoke_resource` using the *DB record's own* stored URI rather than the raw request path — no filesystem or URL-fetch path is driven directly by unvalidated user input, so I don't see a traversal vector here beyond what already exists for the shared `read_resource()` service method.

### Implementation vs. PR description

Matches what's claimed: mirrors the admin endpoint's route pattern and 404 semantics, uses the correct permission string with bypass disabled, and correctly performs Layer-1 scoping via the canonical helper for both v1 and legacy prefixes. `Closes #5356` is present in the base commit (`7d533bf`).

msureshkumar88 (approved):
E2E verified against a real running instance (JWT auth, real RBAC role/team seeding, real SQLite DB) — 20/20 checks pass, including cross-team visibility scoping, admin bypass, v1/legacy/admin parity, and the routing-collision behavior for URIs starting with `test/`. Unit (11/11), integration (7/7), and the fixed e2e delete test (2/2) all pass. No blocking findings — full writeup in the comment above.

msureshkumar88 (changes requested):
check above comments

msureshkumar88 (approved):
Re-approving — the items in my earlier review were explicitly called out as non-blocking suggestions, not blockers, so flipping that CHANGES_REQUESTED back to approve.

Follow-ups (non-blocking, worth a future PR/issue):
1. `test/` is now a reserved first path segment under `/resources` — worth a docstring/PR-description callout.
2. Duplication with `admin_test_resource` (`admin.py:13043`) — consider a shared `_read_resource_by_uri()` helper if either handler changes again.
3. No mention in hand-maintained docs/CHANGELOG — likely fine since docs are largely OpenAPI-generated.
4. Integration tests mock `read_resource`/`PermissionService`; a DB-backed cross-team test would make the no-leakage guarantee self-evident from the suite alone.
5. Redundant `db.commit()`/`db.close()` in the handler given `get_db()` already manages the session lifecycle — consistent with existing sibling handlers, just a footgun if copy-pasted elsewhere.

No blocking correctness/security issues found.

PR #5463, opened 2026-07-01
fix(resources): skip post-fetch resolution for direct_proxy reads

### LINKED ISSUE #5451: [BUG]: direct_proxy resource read raises AttributeError ('…ResourceContents' has no attribute 'id') and returns empty content
### 🐞 Bug Summary

With `MCPGATEWAY_DIRECT_PROXY_ENABLED=true` and a gateway in `gateway_mode=direct_proxy`, a `resources/read` from a normal MCP client returns empty content. The direct_proxy branch successfully fetches the resource from the upstream and builds a `TextResourceContents`/`BlobResourceContents`, then the shared post-fetch block calls `getattr(content, "id")` on that object — which has no `id` field — and raises `AttributeError`. The exception is caught at the transport layer and turned into empty content, so the client never sees the fetched data.

This affects the path a normal client takes (no `X-Context-Forge-Gateway-Id` header). With that header set, the request uses a different transport-level proxy path that returns the content correctly.

---

### 🧩 Affected Component

---

### 🔁 Steps to Reproduce

1. Set `MCPGATEWAY_DIRECT_PROXY_ENABLED=true` and restart the gateway.
2. Federate an upstream MCP server that exposes a resource, then set the gateway to direct_proxy: `PUT /gateways/{id}` with `{"gateway_mode":"direct_proxy"}` (the gateway also needs `transport: STREAMABLEHTTP`).
3. From an MCP client, `resources/read` the federated resource through `/servers/{id}/mcp` — without sending the `X-Context-Forge-Gateway-Id` header.

Reproduced locally against a FastMCP upstream federated into a local gateway (sqlite), for both a text and a binary (`image/png`) resource.

---

### 🤔 Expected Behavior

The read returns the upstream resource's content (text or blob), fetched live. Instead the client receives empty content:

```
ReadResourceResult(contents=[TextResourceContents(uri=AnyUrl('test://schema'), mimeType='text/plain', text='')])
```

---

### 📓 Logs / Error Output

```
ERROR mcpgateway.transports.streamablehttp_transport - Error reading resource 'test://schema': 'TextResourceContents' object has no attribute 'id'
  File ".../mcpgateway/services/resource_service.py", line 2549, in read_resource
    resource_id=getattr(content, "id"),
AttributeError: 'TextResourceContents' object has no attribute 'id'
```

The binary resource fails identically with `'BlobResourceContents' object has no attribute 'id'` at the same line.

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Version or commit | `v1.0.4` (also present on `main`) |
| Runtime | Python 3.12.13 |
| Platform / OS | Linux / macOS (reproduced locally, sqlite) |
| Container | `mcp-contextforge-gateway==1.0.4` |
| Flags | `MCPGATEWAY_DIRECT_PROXY_ENABLED=true`, gateway `gateway_mode=direct_proxy` |

---

### 🧩 Additional Context

The direct_proxy branch in `resource_service.py::read_resource` (around lines 2381–2432, gated on `gateway.gateway_mode == "direct_proxy"` and `settings.mcpgateway_direct_proxy_enabled`) fetches over `streamablehttp_client` and constructs a `TextResourceContents` or `BlobResourceContents` from `mcpgateway/common/models.py`. Those classes define only `uri`, `mime_type`, `meta`, and `text`/`blob` — there is no `id` field.

Because both subclass `ResourceContents`, execution then falls into the post-fetch "RESOLVE CONTENT" block (around lines 2541–2576). Its first branch, `if isinstance(content, (ResourceContent, ResourceContents, TextContent)):` (line 2545), matches, and the next statement runs `resource_id=getattr(content, "id")` (line 2549) with no default, raising `AttributeError`. There is no early return between the direct_proxy fetch and that call, so every direct_proxy text or blob read hits it. The handler in `streamablehttp_transport.py` catches `Exception` and returns `""`.

The test suite works around this by monkey-patching `id`-bearing subclasses over the real models (`tests/unit/mcpgateway/services/test_resource_service.py` ~7127), with a comment noting the post-direct-proxy code calls `getattr(content, "id")` unconditionally — so the production classes that lack `id` take a path the tests never exercise.

A fix would return the constructed `TextResourceContents`/`BlobResourceContents` from the direct_proxy branch directly rather than routing it through the legacy `getattr(content, "id")` post-processing, or guard that block with `getattr(content, "id", None)` plus a blob-aware path. A test asserting that a direct_proxy text read and an `image/png` read both return content end-to-end would cover it.

Ewertonslv:
## 🔗 Related Issue

Closes #5451

## 📝 Summary

In `direct_proxy` gateway mode, `ResourceService.read_resource` fetches the
resource live from the upstream and builds the final `TextResourceContents` /
`BlobResourceContents`. Execution then fell through into the shared post-fetch
**RESOLVE CONTENT** block, whose first branch runs `getattr(content, "id")`
unconditionally. Those MCP-compliant models (`mcpgateway/common/models.py`) have
no `id` field, so **every** direct_proxy text or blob read raised
`AttributeError: 'TextResourceContents' object has no attribute 'id'`. The
transport layer catches the exception and returns empty content, so the client
silently receives `text=""` instead of the fetched payload.

### Root cause
The direct_proxy branch never marked its content as already-resolved — the
`# Skip the rest of the DB lookup logic` comment did not actually skip anything.

### Fix
Track whether content came from the direct_proxy branch (`direct_proxy_read`) and
skip the metadata→content resolution for it. The gateway access check, the
resource access check, and the post-fetch plugin hooks all still run — only the
id-based `invoke_resource` re-resolution is bypassed, since the payload is final.

### Tests (before → after)
The existing happy-path direct_proxy tests masked the bug by monkey-patching
id-bearing subclasses over the real models, so production classes never took the
failing path. This PR adds two regression tests exercising the **real**
`TextResourceContents` / `BlobResourceContents`:

- Without the fix: both raise `AttributeError` at the `getattr(content, "id")` call.
- With the fix: text read returns the live text, `image/png` blob read returns its
  blob, and `invoke_resource` is asserted **not** re-awaited.

`pytest tests/unit/mcpgateway/services/test_resource_service.py` passes locally
(full file green, including module doctests via `--doctest-modules`).

## 📏 Reviewability

## 🏷️ Type of Change

- [x] Bug fix


Ewertonslv:
Correction to my checklist above: I ticked "the linked issue is not labeled `triage`", but #5451 is in fact still labeled `triage` — I misread the label state when opening this. Apologies. I've commented on the issue with the root-cause analysis so a maintainer can triage it, and I'm happy for this PR to sit until that happens, or to close it if the approach isn't wanted.

For coordination: this and #5467 were opened the same day and both touch `read_resource` in `resource_service.py`, but they fix different bugs on different code paths and their hunks don't overlap (lines ~2417 vs ~2430), so they merge cleanly in either order and neither depends on the other. Kept separate deliberately, per the one-concern rule. DCO is signed off.

jonpspri:
Hi @Ewertonslv — sincere apologies for leaving this PR open for so long without a proper review or merge. That's on us, and we're sorry for the lack of responsiveness.

We're closing this PR now as part of a housekeeping pass on long-open contributions. The linked issue (#5451) remains open, so the need is still tracked.

1. Rebase your branch on the current `main`
2. Re-open this PR (or open a fresh one)

Ewertonslv:
Thanks @jonpspri — no problem at all, and thanks for the transparency about the housekeeping pass.

Rebased onto current `main` (`ffcc0822`) and re-verified. GitHub would not let me re-open this PR after the rebase (force-pushing the head branch makes a closed PR non-reopenable), so I took the "or open a fresh one" path: **#6503**.

Same change, only rebased. One conflict, resolved by keeping both sides — `main` grew the `_set_gateway_content` helper right above the `if isinstance(content, ...)` branch, so the helper is untouched and the `direct_proxy_read` guard now sits in front of it.

The bug is still present on today's `main`, and the regression tests still prove it: with the production hunk reverted they fail with `AttributeError: 'TextResourceContents' object has no attribute 'id'`; with it applied they pass, the full `test_resource_service.py` is green (309 tests), and ruff/black are clean. Details in #6503.

PR #5467, opened 2026-07-01
fix(resources): fetch federated resource live when cached content is NULL

### LINKED ISSUE #5450: [BUG]: Federated resource read returns empty in default (cache) mode — NULL-content rows raise ValueError instead of proxying to upstream
### 🐞 Bug Summary

In the default `cache` gateway mode, reading a federated resource can return empty content instead of the upstream's real content. Federation stores only resource metadata, not the content payload, and the cache-mode read path does not reliably fetch the content from the upstream on read.

The clearest case is a resource whose stored content is `NULL` (for example, resources created via the migration / re-discovery path). On read, `Resource.content` raises `ValueError("Resource has no content")`, which is caught and returned as empty content.

This is not universal. A resource federated via the normal `POST /gateways/` path is stored with empty-string content rather than `NULL`, so it does not raise, and a text resource then proxies its live content correctly and reads fine. The empty read affects `NULL`-content resources (above) and binary resources, which fail through a separate path described below.

---

### 🧩 Affected Component

- [x] `mcpgateway` - API

---

### 🔁 Steps to Reproduce

1. Federate an upstream MCP server that exposes a resource (e.g. a `text/markdown` `wiki://schema`) such that its stored content is `NULL` — seen with resources created via the migration / re-discovery path (`createdVia: rediscovery`).
2. From an MCP client, `resources/read` that URI through a virtual server (`/servers/{id}/mcp`).

---

### 🤔 Expected Behavior

The read returns the upstream resource's real content (text or binary), fetched from the gateway when no usable content is cached locally. Instead the client gets empty content and the server logs `Resource has no content`.

---

### 📓 Logs / Error Output

```
ERROR mcpgateway.transports.streamablehttp_transport  Error reading resource 'wiki://schema': Resource has no content
  File ".../mcpgateway/transports/streamablehttp_transport.py", line 2636, in read_resource
    result = await resource_service.read_resource(...)
  File ".../mcpgateway/services/resource_service.py", line 2436, in read_resource
    content = resource_db.content
  File ".../mcpgateway/db.py", line 3739, in content
    raise ValueError("Resource has no content")
ValueError: Resource has no content
```

The client receives `{"contents":[{"uri":"wiki://schema","mimeType":"text/plain","text":""}]}`.

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Version or commit | `v1.0.4` (also reproduces on `main`) |
| Platform / OS | Linux (K8s) |
| Container | `ghcr.io/ibm/mcp-context-forge:v1.0.4` |

---

### 🧩 Additional Context

`read_resource` reads `content = resource_db.content` (`resource_service.py` lines 2238 / 2436) for a federated resource the same way it does for a local one. When `text_content` and `binary_content` are both `NULL`, the `Resource.content` property raises (`db.py:3739`). That exception reaches the MCP handler in `streamablehttp_transport.py::read_resource` (lines 2635–2646), which catches `Exception` and returns `""`. The post-fetch block that calls the live-proxy `invoke_resource` (line 2547) is never reached, because `.content` raises first.

Binary resources read empty for a different reason. When stored content is an empty string, `.content` succeeds and `invoke_resource` is called to fetch live, but its SSE/StreamableHTTP helpers extract the result with `getattr(contents[0], "text")` unconditionally (`resource_service.py` ~2085 / ~2096). For a `BlobResourceContents` (e.g. `image/png`) that raises `AttributeError`, which is caught and returned as `None`, so the read falls back to the cached empty string. Text resources proxy fine; binary ones come back empty.

A fix would make the cache-mode read proxy to the upstream whenever no usable local content exists (treating `NULL`/empty stored content as "fetch live" rather than raising into an empty return), and extend `invoke_resource`'s helpers to propagate `.blob` as well as `.text`. Regression coverage in `tests/unit/mcpgateway/services/test_resource_service.py::TestResourceReading` for a `NULL`-content resource, a text resource, and an `image/png` resource would lock it in.

Ewertonslv:
## 🔗 Related Issue

Refs #5450 (scoped to the NULL-content case — see **Scope** below; not an auto-close).

## 📝 Summary

In `cache` gateway mode, `ResourceService.read_resource` reads
`content = resource_db.content` for a federated resource. When both
`text_content` and `binary_content` are `NULL` (e.g. resources created via the
migration / re-discovery path, `createdVia: rediscovery`), the `Resource.content`
property raises `ValueError("Resource has no content")` **before** the RESOLVE
CONTENT block. The transport layer catches the exception and returns empty
content, so the client silently receives an empty read instead of the upstream's
real content.

### Fix
Catch that `ValueError` in the cache-mode branch and, when the resource is
federated (has a gateway), build a metadata-only `ResourceContent` so the
**existing** RESOLVE CONTENT block fetches it live from the gateway — the same
path already used for resources whose cached content is an empty string. A
non-federated resource has nothing to fetch, so the original error is preserved
(re-raised).

### Tests (before → after)
- `..._null_content_federated_fetches_live`: NULL content + gateway → without the
  fix, `ValueError` propagates and the read is empty; with the fix, the content is
  fetched live via `invoke_resource`. (RED→GREEN verified locally with `git stash`.)
- `..._null_content_no_gateway_preserved`: NULL content + no gateway → the
  "no content" error is preserved and no live fetch is attempted.

`pytest tests/unit/mcpgateway/services/test_resource_service.py` passes locally
(full file green, including module doctests via `--doctest-modules`).

## 🔎 Scope / follow-up (binary resources)

The issue also notes that **binary** federated resources read empty for a
separate reason: `invoke_resource`'s SSE/StreamableHTTP helpers extract the live
payload with `getattr(contents[0], "text")` unconditionally, so a
`BlobResourceContents` (e.g. `image/png`) yields `None`. Fixing that cleanly
requires `invoke_resource` to signal **which** field (text vs blob) the upstream
returned — both are strings, so they can't be disambiguated after the fact — i.e.
a small change to its return contract, plus routing in the RESOLVE block. That is
a deliberate design decision I left out of this PR to keep it focused and
low-risk. Happy to follow up with that in a separate PR if you'd like — just let
me know your preference on the `invoke_resource` return shape.

## 📏 Reviewability

## 🏷️ Type of Change

- [x] Bug fix

### DIFF
--- a/mcpgateway/services/resource_service.py
+++ b/mcpgateway/services/resource_service.py
@@ -2430,7 +2430,24 @@ class ResourceService(BaseService):
 
                     elif resource_db:
                         # Normal cache mode - resource found in DB
-                        content = resource_db.content
+                        try:
+                            content = resource_db.content
+                        except ValueError:
+                            # Federated resource with no cached content (e.g. NULL text/binary
+                            # from the migration / re-discovery path). Build a metadata-only
+                            # ResourceContent so the RESOLVE CONTENT block below fetches it live
+                            # from the gateway, instead of the ValueError bubbling up and the
+                            # client receiving empty content (issue #5450). Local (non-federated)
+                            # resources have nothing to fetch, so the original error is preserved.
+                            if resource_db.gateway is None:
+                                raise
+                            content = ResourceContent(
+                                type="resource",
+                                id=str(resource_db.id),
+                                uri=resource_db.uri,
+                                mime_type=getattr(resource_db, "mime_type", None),
+                                text="",
+                            )
                     else:
                         # Check the inactivity first using the same server scope that
                         # governed the active lookup. Without this, duplicate URIs
--- a/tests/unit/mcpgateway/services/test_resource_service.py
+++ b/tests/unit/mcpgateway/services/test_resource_service.py
@@ -864,6 +864,100 @@ class TestResourceReading:
         finally:
             current_trace_id.reset(token)
 
+    @pytest.mark.asyncio
+    async def test_read_resource_cache_mode_null_content_federated_fetches_live(self):
+        """Regression (issue #5450): a federated resource whose cached content is NULL
+        must be fetched live from the gateway, not returned as empty content.
+
+        Before the fix ``resource_db.content`` raised ``ValueError('Resource has no
+        content')`` before the RESOLVE CONTENT block, the transport swallowed it, and the
+        client received empty content instead of the upstream payload.
+        """
+        # First-Party
+        from mcpgateway.services.resource_service import ResourceService
+
+        class _NullContentResource:
+            id = "res-null-1"
+            uri = "wiki://schema"
+            name = "wiki"
+            enabled = True
+            visibility = "public"
+            team_id = None
+            owner_email = None
+            tags = []
+            team = None
+            mime_type = "text/markdown"
+
+            def __init__(self, gateway):
+                self.gateway = gateway
+
+            @property
+            def content(self):
+                raise ValueError("Resource has no content")
+
+        gateway = MagicMock()
+        gateway.gateway_mode = "cache"  # normal cache mode (NOT direct_proxy)
+        resource_db = _NullContentResource(gateway)
+
+        db = MagicMock()
+        scalar = MagicMock()
+        scalar.scalar_one_or_none.return_value = resource_db
+        db.execute.return_value = scalar
+        db.get.return_value = None
+
+        svc = ResourceService()
+        with (
+            patch.object(svc, "_check_resource_access", new_callable=AsyncMock, return_value=True),
+            patch.object(svc, "invoke_resource", new_callable=AsyncMock, return_value="live wiki body"),
+        ):
+            result = await svc.read_resource(db, resource_uri="wiki://schema", user="u@example.com", token_teams=["team-1"])
+            # The empty local content must trigger a live fetch through the gateway.
+            svc.invoke_resource.assert_awaited()
+
+        assert result is not None
+        assert getattr(result, "text", None) == "live wiki body"
+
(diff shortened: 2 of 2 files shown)


Ewertonslv:
Correction to my checklist above: I ticked "the linked issue is not labeled `triage`", but #5450 is in fact still labeled `triage` — I misread the label state when opening this. Apologies. I've commented on the issue with the root-cause analysis so a maintainer can triage it, and I'm happy for this PR to sit until that happens, or to close it if the approach isn't wanted.

For coordination: this and #5463 were opened the same day and both touch `read_resource` in `resource_service.py`, but they fix different bugs on different code paths and their hunks don't overlap (lines ~2430 vs ~2417), so they merge cleanly in either order and neither depends on the other. Kept separate deliberately, per the one-concern rule. DCO is signed off.

jonpspri:
Hi @Ewertonslv — sincere apologies for leaving this PR open for so long without a proper review or merge. That's on us, and we're sorry for the lack of responsiveness.

Ewertonslv:
Thanks @jonpspri — no worries.

Rebased onto current `main` (`ffcc0822`, clean rebase) and re-verified. GitHub would not let me re-open this PR after the rebase (force-pushing the head branch makes a closed PR non-reopenable), so I took the "or open a fresh one" path: **#6504**.

On step 1 (tracking issue): #5450 already exists and is still open, so I did not file a duplicate — happy to open a fresh one if you'd rather.

Still reproducible on today's `main`: cache mode does a bare `content = resource_db.content`, so a federated row with NULL cached content raises `ValueError: Resource has no content` instead of proxying upstream. With the production hunk reverted, `..._null_content_federated_fetches_live` fails with exactly that; with it applied, both new tests pass, the full `test_resource_service.py` is green (309 tests), and ruff/black are clean. Details in #6504.

Issue #3883, opened 2026-03-27
[BUG][OBSERVABILITY]: ObservabilityService _safe_commit bypasses main transaction control, causing orphaned records on late failures


MohanLaksh:
## 🔗 Related Issues

Follow-up to #3813 (transaction control fix)
Related to #3731 (transaction management violation)

---

## 📝 Summary

**Problem**: `ObservabilityService._safe_commit()` eagerly commits observability data (traces, spans, events) independently of the main request transaction managed by `get_db()`. This creates orphaned observability records when errors occur after observability commits but before the main transaction completes.

**Impact**: Data consistency issue in observability records, particularly for streaming endpoints where late failures can occur after observability data has been committed.

---

### Current Behavior

In `mcpgateway/services/observability_service.py`, the `_safe_commit()` method is called by:
- `start_trace()` - line 276
- `end_trace()` - line 324  
- `start_span()` - line 409 (when `commit=True`)
- `end_span()` - line 452 (when `commit=True`)
- `add_event()` - line 636
- `record_token_usage()` - line 711
- `record_metric()` - line 1048
- `delete_old_traces()` - line 1434

Each call to `_safe_commit()` commits the session **immediately**, independent of the main request transaction.

### The Problem

**Late-Failure Streaming Edge Case**:

```
1. Request arrives → observability_middleware starts trace
2. _safe_commit() commits trace to DB ✅
3. Handler begins processing (e.g., streaming response)
4. Error occurs AFTER observability commit
5. get_db() rolls back main transaction ❌
6. Observability records remain in DB → orphaned data
```

**Example Scenario**:
```python
# Middleware
start_trace(db, trace_id, ...)  
_safe_commit(db, "start_trace")  # ← Commits observability data

# Later in streaming handler
async def stream_response():
    yield chunk1  # Success
    yield chunk2  # Success  
    # Error occurs here
    raise ValueError("Stream failed")  # Main transaction rolls back

# Result: Trace shows "success" but request actually failed
```

### Root Cause

Observability commits are eager and independent of main transaction control. The PR #3813 note acknowledged this:

> "The observability service methods call `_safe_commit()` which commits the session eagerly. This is pre-existing behavior (untouched by this PR) and does not undermine the fix — for the error case, `get_db()` rolls back before `_safe_commit` runs on `end_span`/`end_trace`. The late-failure streaming edge case is pre-existing and out of scope for this focused bug fix."

---

## 🎯 Expected Behavior

[...]

PR #128, opened 2025-06-18
Implemented  Transport-Translation Bridge (mcpgateway.translate)

manavgup:
# 📝 Pull Request Template Selection

Thank you for contributing! To help us review your pull request effectively, please select the appropriate template:

- **Feature / Enhancement**: [Use the Feature Template](?template=feature.md)
I have implemented the [Transport-Translation Bridge (mcpgateway.translate) feature](#94) for the MCP Gateway. 

### 1. Core Bridge Implementation
- **mcpgateway/translate/__init__.py**: Complete CLI interface with argparse, validation, and configuration management
- **mcpgateway/translate/bridge.py**: Core bridge logic with pluggable transport endpoints and bidirectional message pumping
- Support for stdio, SSE, WebSocket, and Streamable-HTTP transports
- Reuses existing SSETransport and WebSocketTransport classes

### 2. CLI Features
- Console script entry point: `mcpgateway-translate`
- Mutually exclusive input transports (--stdio, --sse, --streamableHttp)
- Auto-detection of output transport based on input
- Comprehensive flag validation and error handling
- Support for headers, OAuth2 Bearer tokens, CORS, and health endpoints

### 3. Docker Support
- **docker/translate.Dockerfile**: Multistage build based on python:3.12-slim
- Multiple variants: base, uvx (Node.js support), deno
- Updated CI/CD pipeline to build and push all variants with proper tagging
- Lightweight images with security best practices

### 4. Testing & Quality
- **tests/unit/mcpgateway/translate/test_cli.py**: Comprehensive CLI tests
- **tests/unit/mcpgateway/translate/test_bridge.py**: Core bridge functionality tests
- Unit tests cover argument parsing, validation, transport endpoints, and message routing
- High test coverage with mocking for external dependencies

### 5. Documentation
- **docs/translate.md**: Complete user guide with quick start, command reference, use cases
- Covers Docker usage, environment variables, API endpoints, security considerations
- Includes troubleshooting guide and performance tuning recommendations

## User Stories Fulfilled:

✅ **User Story 1**: `mcpgateway-translate --stdio "uvx mcp-server-git" --port 9000` exposes stdio server over SSE
✅ **User Story 2**: `mcpgateway-translate --sse "https://corp.example.com/mcp"` bridges remote SSE to local stdio  
✅ **User Story 3**: Health endpoint and CORS support with `--healthEndpoint /healthz --cors "https://app.example.com"`
✅ **User Story 4**: Official Docker images with variants published to GHCR

[...]

PR #5470, opened 2026-07-01
feat(plugins): build + consume CPEX plugin trace context and metrics (G0+G1)

### LINKED ISSUE #5458: [FEATURE]: Build and consume CPEX plugin trace context + metrics at the gateway boundary (G0+G1)
## Summary
Companion gateway work for IBM/cpex-plugins#27. Two phases:

- **G0** — The gateway builds `Extensions(request=RequestExtension(trace_id, span_id))` from the active trace and passes `extensions=` into plugin execution. `cpex.framework.manager.invoke_hook` already accepts and forwards `extensions=` (no framework change); the ~17 gateway call sites currently pass nothing. Also bridge the request `span_id` into `request.state` (today only `trace_id` is bridged).
- **G1** — After the executor returns, read `PluginResult.metadata["<plugin>"]`, validate as untrusted output (type/size only), batch DB writes, and attach as span attributes (primary) / `ObservabilityService.record_metric` (secondary).

## Out of scope
- Plugin-side changes (tracked in IBM/cpex-plugins#27).
- A separate telemetry transport or external OTel collector requirement.

## Related
- Plugin side: IBM/cpex-plugins#27
- Framework contract: contextforge-org/cpex#43 (trace/metrics protocol)
- Prior (closed): IBM/mcp-context-forge#4225 (partial trace-context handoff), #4220 (per-plugin spans)
- Adjacent (open): IBM/mcp-context-forge#3736 (observability hardening)

msureshkumar88:
## Summary

Closes #5458. Gateway-side companion to `IBM/cpex-plugins#128` (plugin-side: accept trace context in, return gated metrics out via `result.metadata`). This PR makes the feature live: builds the trace context plugins receive, and consumes whatever gated metrics they return, into the gateway's existing observability infrastructure.

## Design decisions & why

**Why `result.metadata` as the consumption point, not a new channel.** The plugin side (#128) chose to emit metrics on `PluginResult.metadata["<plugin_name>"]` because the CPEX executor already accumulates it losslessly across a multi-plugin hook chain (see #128's description for the full trade-off analysis against `Extensions.custom`, which is lossy for >1 plugin on the same hook). This PR's consumer (`record_plugin_metrics()`) is the gateway-side mirror of that decision: it reads the same accumulated dict rather than standing up a parallel channel.

**Why validate `result.metadata` as untrusted output (S4), not trust it.** Any plugin can write to this dict, not just ones we wrote. `mcpgateway/plugins/utils.py`'s `_sanitize_plugin_metrics()` treats it as adversarial/malformed input: only bounded scalar values (`str`/`int`/`float`/`bool`) or a `list[str]` (joined into one string) survive; everything else is dropped. Key count, string length, and plugin-namespace count are all capped. Dropped values are never logged (only the key name and a short reason) since the dropped value itself may be oversized or sensitive.

**Why a dedicated `plugin.metrics.<name>` span instead of attaching attributes to an existing span.** By the time `record_plugin_metrics()` runs (after `invoke_hook()` returns), both the CPEX-owned hook-chain span and the call-site's own span have typically already closed — there's no live span left to attach attributes to in place. A small dedicated span exists purely to carry the attributes; it doesn't represent real work.

**Why metrics are gated on `trace_id` presence (P-1).** Both sides treat tracing as opt-in: the plugin only emits `result.metadata["pii_filter"]` when it receives a trace_id, and `record_plugin_metrics()` is a no-op if `trace_id` or `result_metadata` is falsy. This means the metrics-recording path costs nothing on untraced requests.

**Why the primary (span) and secondary (numeric metric) paths are batched differently.** `start_span`/`end_span` accept an external `obs_db` session, so every plugin's span in a single `invoke_hook()` call batches into one DB session/commit (P-3). `record_metric()` doesn't accept an external session (pre-existing API, out of scope to change here) — each numeric field's metric write opens its own independent short-lived session, mirroring the existing "issue #3883" independent-session pattern used throughout `ObservabilityService`. This is a deliberate, documented limitation of the secondary path, not an oversight.

**Why `current_span_id` is a new ContextVar rather than threading `request.state.span_id` down.** Investigated mid-implementation: the 9 `invoke_hook()` call sites (7 in `tool_service.py`, 2 in `prompt_service.py`) are deep in the service layer with no `Request` object in scope. Mirrored the existing `current_trace_id` ContextVar pattern rather than plumbing `Request` through several call layers.

## What changed

- **G0** — builds `Extensions(request=RequestExtension(trace_id, span_id))` from the active trace and passes `extensions=` into all 9 real `invoke_hook()` call sites. Bridges `request.state.span_id` (previously only `trace_id` was bridged) and adds the `current_span_id` ContextVar.
- **G1** — after each `invoke_hook()` call, `record_plugin_metrics()` validates (S4) and records the plugin's metadata: primary path as a dedicated `plugin.metrics.<name>` span (batched), secondary path as `ObservabilityMetric` rows for numeric fields. Fully best-effort (L4) — any failure (missing trace, DB error, malformed plugin metadata) is logged at debug level and swallowed; it never raises into the request path.
- **E2E proof** — a real pytest test and a live-tested shell script send a genuinely traced HTTP tool-invoke request (W3C `traceparent` header) through the real app and assert `GET /observability/traces/{trace_id}` shows the `pii_filter` plugin's metrics, with no raw PII anywhere in any response.

## CI stabilization (follow-up commits, after initial push)

- **`pylint` (cyclic-import)**: `mcpgateway.plugins.utils` imports `mcpgateway.services.observability_service` — deferred inside the function specifically to avoid a *runtime* cycle (`mcpgateway.services.__init__` eagerly imports `tool_service`/`prompt_service`, both of which import this helper). Pylint's R0401 traces function-scoped imports statically regardless of deferral, so it still flags the (already broken-at-runtime) cycle. Extended the existing `# pylint: disable=import-outside-toplevel` line to also suppress `cyclic-import`, rather than restructuring — the deferral is the real fix; the pylint disable just acknowledges pylint can't see it.
- **`pytest` — stale mocks**: 12 pre-existing tests in `test_tool_service_coverage.py` mocked `invoke_hook()` results as bare `SimpleNamespace(modified_payload=...)` without a `.metadata` attribute, or a custom `mock_invoke_hook()` function missing the new `extensions=` parameter. Both break once `tool_service.py` started reading `.metadata` off every hook result and passing `extensions=` into every `invoke_hook()` call. Fixed by adding `metadata=None` / `extensions=None` to the mocks — no assertions changed, since these tests aren't exercising plugin-metrics behavior, just working around the new call signature.
- **Pre-existing bug found and fixed in `mcpgateway/main.py`** (unrelated file, not otherwise touched by this PR): `_service` is a module-level global, only bound inside `if settings.observability_enabled:` at import time, but read unconditionally later in `lifespan()` (already guarded by a pre-existing `pylint: disable=possibly-used-before-assignment`, dating to an April 2026 commit). Any test that flips `observability_enabled` to `True` after import and then invokes the real `lifespan()` hits `NameError: name '_service' is not defined`. Confirmed via `git blame` that this predates this branch and would affect any PR under the right pytest-xdist worker/test-order combination — it was intermittently failing this PR's own CI. Fixed with a 1-line unconditional `_service = None` before the conditional.
- **Root-caused what looked like e2e flakiness, but isn't a bug**: `test_otel_plugin_metadata_e2e.py` was failing on GitHub Actions with "no `plugin.metrics.pii_filter` span found" — masking worked, PII didn't leak, but the metrics span never appeared. Passed 20/20 in isolated local runs, which pointed at CI-only nondeterminism. It isn't: the local dev `.venv` has `cpex-pii-filter` installed **editable**, pointing straight at the local `cpex-plugins` checkout (0.3.6, with the metrics work from #128). GitHub Actions installs from PyPI per `uv.lock`, which is pinned to the **published `0.3.5`** — a version that predates the metrics-emission behavior this PR consumes. This is not fixable in this repo; see Merge order below.

## Merge order (cross-repo dependency)

**1. Merge and release `IBM/cpex-plugins#128` first**, tagging `pii-filter-v0.3.6` to trigger its PyPI publish workflow.

**2. Then, on this PR (or as an immediate follow-up commit before merge):** bump `cpex-pii-filter>=0.3.5` to `>=0.3.6` in `pyproject.toml` and regenerate `uv.lock` (`uv lock --upgrade-package cpex-pii-filter`). Until that lock bump lands, `pytest (py3.12)` in this repo's CI will keep failing/looking flaky on `test_otel_plugin_metadata_e2e.py` specifically, because CI is resolving a `cpex-pii-filter` version that doesn't have the metrics behavior this test asserts on — that is expected and not a regression to chase further here.

**3. Merge this PR last**, after the lock bump is in and CI is green against the real published `0.3.6`.

## Test plan

- [x] `test_observability_middleware.py` — 13 passing (span_id bridging)
- [x] `test_plugins_utils.py` — 20 passing (Extensions-building + metrics-consumer unit tests, including L4 swallow and S4 validation)
- [x] `test_tool_service.py` — 379 passing; `test_prompt_service.py` — 177 passing (all 9 wiring sites)
- [x] `test_tool_service_coverage.py` — 12 previously-broken mocks fixed, full file passing
- [x] New `tests/integration/plugins/test_plugin_metrics_consumer_integration.py` — 6 passing (real DB rows, run with `--with-integration`)
- [x] New `tests/e2e/test_otel_plugin_metadata_e2e.py` — passing locally against an editable install of the plugin-side 0.3.6 changes; **will only pass in this repo's CI once the `uv.lock` bump in step 2 above lands** (see Merge order)
- [x] `scripts/verify_otel_plugin_e2e.sh` — live-tested against a real running gateway instance, same result
- [x] `pylint (mcpgateway)`, full parallel `pytest -n auto` suite (local, 0 failures) — both green as of `fc366c9a0`, rebased onto latest `origin/main`

### DIFF
--- a/docs/docs/manage/observability/internal-observability.md
+++ b/docs/docs/manage/observability/internal-observability.md
@@ -785,6 +785,92 @@ WHERE start_time < NOW() - INTERVAL '7 days';
    echo $OBSERVABILITY_EVENTS_ENABLED  # Should be "true"
    ```
 
+## Plugin Metrics (`result.metadata`)
+
+Plugins can return per-plugin observability data by setting `PluginResult.metadata["<plugin_name>"]`
+in any hook they implement. The gateway consumes this after every `invoke_hook()` call
+(`mcpgateway.plugins.utils.record_plugin_metrics()`) across all wired hook call sites:
+`tool_pre_invoke`/`tool_post_invoke`, `prompt_pre_fetch`/`prompt_post_fetch`,
+`resource_pre_fetch`/`resource_post_fetch`, `http_pre_request`/`http_post_request`,
+`http_auth_resolve_user`, `http_auth_check_permission`, and `agent_pre_invoke`/`agent_post_invoke`.
+
+### Shape
+
+```text
+result.metadata = {
+    "<plugin_name>": {
+        "<field_name>": <value>,
+        ...
+    },
+    ...
+}
+```
+
+Each top-level key namespaces one plugin's own metadata dict; a single hook invocation can carry
+metadata from every plugin that ran in the chain.
+
+### Value types and limits
+
+`result.metadata` is treated as **untrusted plugin output** and validated before it is recorded
+anywhere (`mcpgateway.plugins.utils._sanitize_plugin_metrics()`). None of these bounds are
+configurable — a plugin that needs different limits should be revisited rather than the gateway
+loosening its defaults:
+
+| Element | Contract |
+|---|---|
+| Plugin name / field name (dict key) | Must match `^[A-Za-z0-9_.-]{1,64}$`; otherwise dropped |
+| `bool` / `int` / `float` value | Recorded as-is |
+| `str` value | Must match the low-cardinality charset `^[A-Za-z0-9_.,:=/-]*$` and be ≤ 64 chars (truncated if longer, dropped if the charset doesn't match) — this is intentionally **not** a free-text channel |
+| `list[str]` value | Joined into a single comma-separated string, then subject to the same `str` contract; only the first 32 items are considered |
+| Any other type (`dict`, mixed list, `None`, etc.) | Dropped |
+| Keys per plugin | Capped at 32 |
+| Plugins per call | Capped at 16 |
+
+Values that are dropped are never logged — only the key name and a short reason — since the value
+itself may be malformed, oversized, or adversarial.
+
+**Plugin metadata is untrusted and must stay low-cardinality and non-sensitive.** It is designed
+for counters, status/type-name tokens, and short enums (e.g. `total_detections: 3`,
+`stage: "tool_post_invoke"`) — not for raw request content or free text.
+
+### DB sink vs. OTel sink
+
+Validated metadata is recorded through up to two independent sinks:
+
+- **Internal DB (primary)**: a dedicated `plugin.metrics.<plugin_name>` span carries the validated
+  fields as span attributes, batched into one DB session per `invoke_hook()` call. Controlled by
+  `PLUGIN_METRICS_DB_SPANS_ENABLED` (default `true`).
+- **Internal DB (secondary, numeric only)**: each numeric field is additionally recorded as an
+  `ObservabilityMetric` row via `record_metric()`, so plugin counters are queryable as metrics, not
+  just span attributes. Each row is its own independent DB session/commit (mirrors the existing
+  "issue #3883" pattern), so this is capped per call across all plugins via
+  `PLUGIN_METRICS_MAX_NUMERIC_PER_CALL` (default `16`) and can be disabled entirely with
+  `PLUGIN_METRICS_DB_NUMERIC_ROWS_ENABLED=false`.
+- **OTel export (optional, additive)**: when an OTel exporter is configured
+  (`otel_tracing_enabled()`) and there is an active OTel context (`otel_context_active()`), the
+  already-sanitized metrics are re-emitted as an OTel span through the gateway's existing exporter.
+  This is purely additive — disabling it does not affect the internal DB sinks above, and the
+  gateway remains the sole export authority (plugins never talk to OTel directly).
+
+All of the above is best-effort (L4): any failure (missing trace, DB error, malformed plugin
+metadata) is logged at debug level and swallowed — it never raises into the request path, and is a
+no-op entirely when no trace is active (metrics recording is gated on `trace_id` presence, so it
+costs nothing on untraced requests).
+
+### Hook coverage
+
+As of this writing, all 17 gateway-side `invoke_hook()` call sites pass trace context in
+(`extensions=`) and consume `result.metadata` out. If you add a new plugin hook call site, wire it
+the same way: pass `extensions=build_request_extensions()` into `invoke_hook()`, then call
+`record_plugin_metrics(current_trace_id.get(), <result>.metadata)` immediately after.
+
+### Dependency on plugin-side metrics emission
+
+This consumer only records something if the plugin actually populates `result.metadata`. The
+bundled PII-filter plugin (`cpex-pii-filter`) emits `result.metadata["pii_filter"]` starting from
+its own `0.3.6` release — check `pyproject.toml`/`uv.lock` for the currently pinned version before
+assuming this data is available end-to-end.
+
 ## Best Practices
 
 ### Development
--- a/docs/docs/manage/observability/internal-observability.md
+++ b/docs/docs/manage/observability/internal-observability.md
@@ -819,9 +819,10 @@ loosening its defaults:
 | Element | Contract |
 |---|---|
 | Plugin name / field name (dict key) | Must match `^[A-Za-z0-9_.-]{1,64}$`; otherwise dropped |
-| `bool` / `int` / `float` value | Recorded as-is |
-| `str` value | Must match the low-cardinality charset `^[A-Za-z0-9_.,:=/-]*$` and be ≤ 64 chars (truncated if longer, dropped if the charset doesn't match) — this is intentionally **not** a free-text channel |
-| `list[str]` value | Joined into a single comma-separated string, then subject to the same `str` contract; only the first 32 items are considered |
+| `bool` value | Recorded as-is |
+| `int` / `float` value | Recorded as-is, but only for field names in a deny-by-default allowlist (currently `total_detections`, `total_masked`); every other numeric field name is dropped regardless of value. Non-finite values (`NaN`/`inf`/`-inf`) are always rejected |
+| `str` value | Recorded as-is, but only for field names in a deny-by-default allowlist (currently `stage`, `detection_types`); every other string field name is dropped regardless of value. Accepted values must match the low-cardinality charset `^[A-Za-z0-9_.,:=/-]*$` and be ≤ 64 chars — overlong values are **rejected outright, not truncated** — this is intentionally **not** a free-text channel |
+| `list[str]` value | Joined into a single comma-separated string, then subject to the same allowlisted `str` contract; only the first 32 items are considered |
 | Any other type (`dict`, mixed list, `None`, etc.) | Dropped |
 | Keys per plugin | Capped at 32 |
 | Plugins per call | Capped at 16 |
--- a/mcpgateway/plugins/utils.py
+++ b/mcpgateway/plugins/utils.py
@@ -219,6 +219,7 @@ def record_plugin_metrics(trace_id: Optional[str], result_metadata: Optional[Dic
         # exporter. No-op when OTel tracing is unconfigured (create_span -> nullcontext).
         try:
             from mcpgateway.observability import create_span, otel_tracing_enabled  # pylint: disable=import-outside-toplevel
+
             if otel_tracing_enabled():
                 for plugin_name, sanitized in sanitized_by_plugin.items():
                     try:
--- a/mcpgateway/plugins/utils.py
+++ b/mcpgateway/plugins/utils.py
@@ -8,7 +8,9 @@ Gateway-side plugin utilities.
 """
 
 # Standard
+import itertools
 import logging
+import re
 from typing import Any, Dict, Optional
 
 # Third-Party
@@ -22,35 +24,71 @@ logger = logging.getLogger(__name__)
 #
 # S4 bounds (untrusted plugin output). These are deliberately conservative,
 # defensible defaults -- not a contract mandated by any single plugin:
-#   - Only scalar values (str, int, float, bool) are recorded as-is.
+#   - plugin_name and metric keys must match _IDENTIFIER_RE (bounded length,
+#     restricted charset) before they're allowed to flow into DB span/metric
+#     names, resource ids, or OTel span/attribute names -- these are
+#     plugin-controlled and otherwise unvalidated.
+#   - Only scalar values (str, int, float, bool) are recorded as-is; strings
+#     must additionally match _SAFE_STRING_VALUE_RE (a low-cardinality,
+#     no-free-text charset) -- truncation alone bounds size, not sensitivity.
 #   - A list[str] value (e.g. ["email", "ssn"]) is special-cased: joined into
 #     a single comma-separated string (OTel span attributes don't nest), then
-#     subject to the same string-length cap as any other string value.
+#     subject to the same string-value validation as any other string value.
 #   - Any other type (dict, mixed/non-str list, None, etc.) is dropped.
-#   - At most _MAX_PLUGIN_KEYS keys survive per plugin's metadata dict.
+#   - At most _MAX_PLUGIN_KEYS keys survive per plugin's metadata dict, and
+#     iteration stops (rather than merely skipping) once that cap is hit so a
+#     huge untrusted dict can't force a full scan.
 #   - At most _MAX_PLUGINS_PER_CALL plugin namespaces are processed per call
 #     (result.metadata can in principle carry entries from every plugin that
-#     ran in the hook chain).
-#   - String values (including the joined list form) are truncated to
-#     _MAX_STRING_LENGTH characters rather than dropped.
+#     ran in the hook chain); the cap is applied before the full item list is
+#     materialized.
+#   - list[str] values are only checked for str-ness within the first
+#     _MAX_LIST_ITEMS items, so an oversized untrusted list can't force a full
+#     scan before it's bounded.
 # On drop, only the key name and a short reason are logged -- never the value
 # itself, since it may be malformed/oversized/adversarial plugin output.
 _MAX_PLUGIN_KEYS = 32
-_MAX_STRING_LENGTH = 256
+_MAX_STRING_LENGTH = 64
 _MAX_PLUGINS_PER_CALL = 16
 _MAX_LIST_ITEMS = 32
 
+# Identifier contract for plugin_name / metric field names: these flow into DB
+# span/metric names, resource ids, and OTel span/attribute names, so they're
+# restricted to a short, safe charset rather than trusted as-is.
+_IDENTIFIER_RE = re.compile(r"^[A-Za-z0-9_.-]{1,64}$")
 
-def _truncate_str(value: str) -> str:
-    """Truncate a string to the S4 bounded length.
+# Low-cardinality value contract for string metadata values: truncation alone
+# bounds size, not sensitivity, so free text (arbitrary PII, spaces, quotes,
+# etc.) is rejected outright rather than truncated. This intentionally only
+# admits short enum/status/type-name-like tokens.
+_SAFE_STRING_VALUE_RE = re.compile(r"^[A-Za-z0-9_.,:=/-]*$")
+
+
+def _is_valid_identifier(name: Any) -> bool:
+    """Check whether a plugin name or metric field name is safe to use as an identifier.
 
     Args:
-        value: The string to (possibly) truncate.
+        name: Candidate identifier (plugin name or metric field name); treated as untrusted.
 
     Returns:
-        The original string if within bounds, otherwise truncated to
-        ``_MAX_STRING_LENGTH`` characters.
+        True if ``name`` is a string matching the bounded, restricted-charset
+        identifier contract; False otherwise.
     """
+    return isinstance(name, str) and bool(_IDENTIFIER_RE.match(name))
+
+
+def _validate_string_value(value: str) -> Optional[str]:
+    """Validate and bound a single string metadata value (S4).
+
+    Args:
+        value: Candidate string value.
+
+    Returns:
+        The value truncated to ``_MAX_STRING_LENGTH`` if it matches the
+        low-cardinality safe-value charset, otherwise ``None`` (reject).
+    """
+    if not _SAFE_STRING_VALUE_RE.match(value):
+        return None
     if len(value) <= _MAX_STRING_LENGTH:
         return value
     return value[:_MAX_STRING_LENGTH]
@@ -61,8 +99,9 @@ def _sanitize_plugin_metrics(plugin_name: str, raw_metrics: Any) -> Dict[str, An
 
     Untrusted plugin output: only scalar (``str``/``int``/``float``/``bool``) values
     survive as-is; a ``list[str]`` value is joined into a single bounded string;
-    everything else is dropped. Key count and string length are capped. Dropped
-    values are never logged -- only the key name and a short reason.
+    everything else is dropped. Key names, key count, and string values are all
+    validated/capped. Dropped values are never logged -- only the key name and a
+    short reason.
 
     Args:
         plugin_name: Namespacing key this metadata was reported under (e.g. "pii_filter").
@@ -79,22 +118,34 @@ def _sanitize_plugin_metrics(plugin_name: str, raw_metrics: Any) -> Dict[str, An
 
     sanitized: Dict[str, Any] = {}
     for key, value in raw_metrics.items():
-        if not isinstance(key, str):
-            logger.debug("Dropped a plugin metric key for %r: key is not a string", plugin_name)
+        if not _is_valid_identifier(key):
+            logger.debug("Dropped a plugin metric key for %r: key is not a valid identifier", plugin_name)
             continue
         if len(sanitized) >= _MAX_PLUGIN_KEYS:
-            logger.debug("Dropped plugin metric %r for %r: key limit exceeded", key, plugin_name)
-            continue
+            logger.debug("Stopping metric scan for %r: key limit exceeded", plugin_name)
+            break
 
         if isinstance(value, bool):
             sanitized[key] = value
         elif isinstance(value, (int, float)):
             sanitized[key] = value
         elif isinstance(value, str):
-            sanitized[key] = _truncate_str(value)
-        elif isinstance(value, list) and all(isinstance(item, str) for item in value):
-            joined = ",".join(value[:_MAX_LIST_ITEMS])
-            sanitized[key] = _truncate_str(joined)
+            validated = _validate_string_value(value)
+            if validated is not None:
+                sanitized[key] = validated
+            else:
+                logger.debug("Dropped plugin metric %r for %r: string value not in safe low-cardinality contract", key, plugin_name)
+        elif isinstance(value, list):
+            prefix = value[:_MAX_LIST_ITEMS]
+            if all(isinstance(item, str) for item in prefix):
+                joined = ",".join(prefix)
+                validated = _validate_string_value(joined)
+                if validated is not None:
+                    sanitized[key] = validated
+                else:
+                    logger.debug("Dropped plugin metric %r for %r: joined list value not in safe low-cardinality contract", key, plugin_name)
+            else:
+                logger.debug("Dropped plugin metric %r for %r: list contains non-string items", key, plugin_name)
         else:
             logger.debug("Dropped plugin metric %r for %r: type not allowed", key, plugin_name)
 
@@ -140,18 +191,20 @@ def record_plugin_metrics(trace_id: Optional[str], result_metadata: Optional[Dic
     try:
         # First-Party (deferred to avoid a circular import with mcpgateway.services.__init__,
         # mirroring build_request_extensions() above).
+        from mcpgateway.config import settings  # pylint: disable=import-outside-toplevel
         from mcpgateway.db import SessionLocal  # pylint: disable=import-outside-toplevel
         from mcpgateway.services.observability_service import ObservabilityService  # pylint: disable=import-outside-toplevel,cyclic-import
 
-        plugin_items = list(result_metadata.items())
-        if len(plugin_items) > _MAX_PLUGINS_PER_CALL:
-            logger.debug("Dropping %d plugin metadata entries beyond the per-call limit", len(plugin_items) - _MAX_PLUGINS_PER_CALL)
-            plugin_items = plugin_items[:_MAX_PLUGINS_PER_CALL]
+        # Bounded iteration: cap applied via islice over the dict's own view rather than
+        # materializing a full list of every entry first (result_metadata is untrusted).
+        if len(result_metadata) > _MAX_PLUGINS_PER_CALL:
+            logger.debug("Dropping %d plugin metadata entries beyond the per-call limit", len(result_metadata) - _MAX_PLUGINS_PER_CALL)
+        plugin_items = itertools.islice(result_metadata.items(), _MAX_PLUGINS_PER_CALL)
 
         sanitized_by_plugin: Dict[str, Dict[str, Any]] = {}
         for plugin_name, raw_metrics in plugin_items:
-            if not isinstance(plugin_name, str):
-                logger.debug("Dropped a plugin metadata entry: plugin name is not a string")
+            if not _is_valid_identifier(plugin_name):
+                logger.debug("Dropped a plugin metadata entry: plugin name is not a valid identifier")
                 continue
             sanitized = _sanitize_plugin_metrics(plugin_name, raw_metrics)
             if sanitized:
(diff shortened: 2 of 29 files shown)


msureshkumar88:
## CI status: one known, expected failure (`pytest (py3.12)`)

All 39 other checks are green. The one failure is expected and not a regression in this PR — worth flagging so it doesn't cause confusion during review.

**What's failing:** `tests/e2e/test_otel_plugin_metadata_e2e.py::test_traced_tool_call_surfaces_pii_filter_metrics_without_leaking_pii`

```
AssertionError: No 'plugin.metrics.pii_filter' span found; span names present: ['http.request', 'tool.invoke']
```

PII masking works correctly and nothing leaks — the test just can't find the metrics span it's asserting on.

**Root cause:** this repo's `uv.lock` currently resolves `cpex-pii-filter` to the published `0.3.5`, which predates the `result.metadata` metrics-emission behavior added in [IBM/cpex-plugins#128](https://github.com/IBM/cpex-plugins/pull/128). CI installs from PyPI per the lockfile, so it's running against a plugin version that simply doesn't have this feature yet. (Confirmed locally: with an editable install pointing at the `cpex-plugins` 0.3.6 branch, this test passes reliably — 20+ runs, 0 failures.)

**Fix (blocked, not something to chase in this repo):**
1. Merge and release `IBM/cpex-plugins#128`, tagging `pii-filter-v0.3.6` to trigger its PyPI publish.
2. Bump `cpex-pii-filter>=0.3.5` → `>=0.3.6` here and refresh `uv.lock` (`uv lock --upgrade-package cpex-pii-filter`).
3. Re-run CI — should go fully green against the real published `0.3.6`.

Full rationale and merge order are in the PR description above.

msureshkumar88:
## G2 (OTel-SDK export sink) — real-environment e2e verification

Follow-up to G0+G1 (already merged into this branch): commits `106f8feb2`, `5d9282ac7`, `e483c09bf` add an **optional** second sink for plugin metrics — the gateway's existing OpenTelemetry SDK exporter — alongside the existing DB sink (`ObservabilityService`). Both sinks are independently flagged, off by default, and gateway-owned; no plugin-side dependency on OTel is introduced.

### What was added
- `mcpgateway/plugins/utils.py`: `record_plugin_metrics()` gains a G2 block that re-emits the *same already-sanitized* metrics dict through `mcpgateway.observability.create_span()`, gated on `otel_tracing_enabled()`. No re-validation, no new exception surface for the existing DB sink (runs strictly after it, independently isolated).
- `tests/e2e/test_otel_plugin_metadata_e2e.py`: new sibling e2e test, `test_traced_tool_call_exports_pii_filter_metrics_to_otel_sdk`, firing the *identical* traced tool call as the existing DB-sink e2e test, asserting the metrics land on a real OTel span (via `InMemorySpanExporter`) instead of a DB row.
- `scripts/verify_otel_plugin_e2e.sh`: real-environment manual verification script (not TestClient/mocks) — starts a throwaway `uvicorn` instance, mints a real admin JWT, registers a real tool, fires a real HTTP `POST /rpc` with a `traceparent` header carrying synthetic PII, then checks **both** sinks from that one request.

### Verification — run live, by me, against a real running gateway (not just unit/e2e mocks)

**Config** (script defaults, no overrides):
```
PLUGINS_ENABLED=true
OBSERVABILITY_ENABLED=true        # G1 (DB sink)
OTEL_ENABLE_OBSERVABILITY=true    # G2 (OTel sink)
OTEL_TRACES_EXPORTER=console
PLUGINS_CONFIG_FILE=plugins/config.yaml   # pii_filter entry enabled
DATABASE_URL=<throwaway sqlite>
cpex-pii-filter==0.3.6 (local build, see cross-repo note below)
```

**Commands run:**
```
bash scripts/verify_otel_plugin_e2e.sh
.venv/bin/python -m pytest tests/e2e/test_otel_plugin_metadata_e2e.py -v
```

**Output (script):**
```
✅ Gateway reachable at http://127.0.0.1:8000
✅ tools/call completed; no raw PII in the RPC response body
✅ S1 confirmed: no raw PII anywhere in the /observability/traces/{trace_id} response
✅✅✅ G1 (DB sink) SUCCESS: total_detections=1 total_masked=1 detection_types=email
✅✅✅ G2 (OTel sink) SUCCESS: total_detections=1 total_masked=1 detection_types=email
   No raw PII value found anywhere in the gateway's captured stdout/stderr.
🎉 Both sinks verified for trace_id=573ca7cce285c16f32605ef0ac1ff32d
🧹 Stopping gateway (pid 306532)...
```
Exit code: `0`. Gateway process confirmed cleanly torn down afterward (no orphan).

**Output (pytest e2e):**
```
tests/e2e/test_otel_plugin_metadata_e2e.py ..                          [100%]
======================== 2 passed, 4 warnings in 3.15s =========================
```

### Review trail
Two per-task reviews (spec compliance + code quality, both independent passes) + one final cross-task integration review — all three returned **0 Critical/Important findings**. The integration review specifically verified the OTel test seam (`monkeypatch` on `mcpgateway.observability._TRACER`) reaches the real production call site — `create_span()`/`otel_tracing_enabled()` re-read that module global at call time rather than closing over a stale reference — and confirmed the G2 addition cannot regress the already-merged G0/G1 path (DB sink runs and commits strictly before the G2 block; independent exception isolation).

### Verdict: verification passed — recommending merge

I authored this PR and pushed these commits under the same GitHub identity, so I'm **not submitting a formal approve/request-changes review** — GitHub (and this session's own tooling) correctly blocks self-review regardless of validation confidence, since that would defeat the two-party review requirement. Posting this as a plain comment with full evidence instead, for a human reviewer to weigh.

### One pre-existing, already-documented item (not introduced by this increment)
The `pytest (py3.12)` CI job may still show red for the *unrelated, pre-existing* reason documented in the PR description's "Merge order" section: `cpex-pii-filter>=0.3.5` is locked in `uv.lock` to the PyPI-published `0.3.5`, which predates the metrics-emission feature. Local verification above used a locally-built `0.3.6` (unpublished). This is a cross-repo release-sequencing dependency on `cpex-plugins#128`, not something this increment touches or can fix — `pyproject.toml`/`uv.lock` are unchanged by these 3 commits.

msureshkumar88:
## G2: optional OTel-SDK export sink for plugin metrics — summary

Follow-up increment on top of the already-merged G0 (trace-context passing) + G1 (DB-sink consumption) in this branch. Commits `106f8feb2`, `5d9282ac7`, `e483c09bf`. Prompted by reviewer feedback on this PR asking whether plugin metrics get exported to the OTel SDK or just stored in the DB — they were DB-only; this adds OTel export as a second, optional path.

### What changed

**`mcpgateway/plugins/utils.py`** (+15 lines) — `record_plugin_metrics()` gains a G2 block: after the existing DB sink (span attributes + `record_metric` rows via `ObservabilityService`) completes, it re-emits the *same already-sanitized* metrics dict through the gateway's existing OTel exporter (`create_span()` from `mcpgateway/observability.py`), one span per plugin named `plugin.metrics.<plugin_name>`, gated on `otel_tracing_enabled()`.

**`tests/unit/mcpgateway/plugins/test_plugins_utils.py`** (+128 lines) — 4 new unit tests: OTel enabled (span opened with correct attributes), OTel disabled (clean no-op), export raises for one plugin (swallowed, doesn't affect the other plugin or the DB sink), multi-plugin isolation.

**`tests/e2e/test_otel_plugin_metadata_e2e.py`** (+100 lines) — new sibling test `test_traced_tool_call_exports_pii_filter_metrics_to_otel_sdk`, firing the *identical* traced tool call as the existing DB-sink e2e test (same payload, same PII fixture, same RPC path), asserting the metrics land on a real OTel span captured via `InMemorySpanExporter` instead of a DB row. Uses `pytest.importorskip` so it degrades to skip (not fail) when the optional `opentelemetry-sdk` extra isn't installed.

**`scripts/verify_otel_plugin_e2e.sh`** (+182 lines) — extended the existing manual real-environment verification script to check both sinks from one throwaway gateway process and one real traced HTTP call: G1 via `GET /observability/traces/{trace_id}`, G2 by starting the gateway with `OTEL_TRACES_EXPORTER=console` and parsing the exported span back out of captured stdout.

### Decisions and why

- **Gateway is the sole OTel export authority; no plugin-side change.** Reviewer-driven constraint. The plugin already only emits `result.metadata["pii_filter"]` (counts/types, never matched content) — it has zero telemetry infrastructure and stays that way. G2 only reuses the gateway's pre-existing `mcpgateway/observability.py` exporter; nothing plugin-side was touched.
- **Second independent flag, not a reuse of the DB flag.** `otel_tracing_enabled()` (driven by `OTEL_ENABLE_OBSERVABILITY`, default off) gates G2; `OBSERVABILITY_ENABLED` (default off) still gates G1. Orthogonal — DB-only, OTel-only, both, or neither are all valid. This was an explicit ask: keep it feature-flagged, and don't couple the new sink's on/off to the existing DB sink's flag.
- **Span attributes only, not a new `MeterProvider`.** The gateway's wired OTel exporter is a *trace* exporter (`BatchSpanProcessor` → OTLP/Jaeger/Zipkin/console). Standing up the OTel *Metrics* SDK (`MeterProvider` + metric reader) for numeric counters doesn't exist anywhere in this codebase today and was scoped out as a separate future enhancement — confirmed via `grep MeterProvider|get_meter|sdk.metrics` returning nothing under `mcpgateway/`. Numeric fields stay queryable via G1's existing `record_metric` DB path.
- **Reuse the already-sanitized dict, no second validation pass.** G1's `sanitized_by_plugin` (S4-bounded: only `str|int|float|bool`, length/key-capped, untrusted-plugin-output guard) is fed to G2 as-is. One sanitization path, two sinks — avoids drift between what each sink considers safe to emit.
- **`InMemorySpanExporter` test seam over mocking `create_span`.** Both `create_span()` and `otel_tracing_enabled()` re-read the module-level `mcpgateway.observability._TRACER` global at call time (verified: neither closes over a stale reference). Patching that one global with a real `TracerProvider` + `InMemorySpanExporter` exercises the actual production code path end-to-end rather than testing a mock.
- **Best-effort, isolated per plugin.** G2 sits in its own inner `try/except` inside the function's existing outer `try/except` — one plugin's export failure can't affect a sibling plugin's export or the DB sink that already ran and committed before G2 executes.

### Tests — config, commands, output

**Unit tests:**
```
cd mcp-context-forge && .venv/bin/python -m pytest tests/unit/mcpgateway/plugins/test_plugins_utils.py -v
```
→ 24 passed (20 pre-existing + 4 new), 0 failures. Full plugin-layer regression check (`tests/unit/mcpgateway/plugins/`): 621 passed.

**E2E tests** (config: `PLUGINS_ENABLED=true OBSERVABILITY_ENABLED=true`, `opentelemetry-sdk` installed via the optional `observability` extra):
```
uv pip install '.[observability]'
.venv/bin/python -m pytest tests/e2e/test_otel_plugin_metadata_e2e.py -v
```
→
```
tests/e2e/test_otel_plugin_metadata_e2e.py ..                          [100%]
======================== 2 passed, 4 warnings in 3.15s =========================
```
(DB-sink test unchanged and still green; new OTel-sink test added and passing, not skipped.)

**Real-environment verification** (config: throwaway `uvicorn` + SQLite, `PLUGINS_ENABLED=true OBSERVABILITY_ENABLED=true OTEL_ENABLE_OBSERVABILITY=true OTEL_TRACES_EXPORTER=console`, `cpex-pii-filter==0.3.6` local build, `plugins/config.yaml` with the `pii_filter` entry):
```
bash scripts/verify_otel_plugin_e2e.sh
```
→ real admin JWT minted, real tool registered, real `POST /rpc` with a `traceparent` header carrying synthetic PII:
```
✅ tools/call completed; no raw PII in the RPC response body
✅ S1 confirmed: no raw PII anywhere in the /observability/traces/{trace_id} response
✅✅✅ G1 (DB sink) SUCCESS: total_detections=1 total_masked=1 detection_types=email
✅✅✅ G2 (OTel sink) SUCCESS: total_detections=1 total_masked=1 detection_types=email
   No raw PII value found anywhere in the gateway's captured stdout/stderr.
🎉 Both sinks verified for trace_id=573ca7cce285c16f32605ef0ac1ff32d
🧹 Stopping gateway (pid 306532)...
```
Exit code `0`; gateway process confirmed cleanly torn down.

### Review trail
Two per-task reviews (implementer + independent task reviewer, each covering spec compliance and code quality separately) plus one final cross-task integration review (checked contract match between implementation and tests, test-seam correctness, no regression to the already-merged G0/G1 path) — all three passes returned 0 Critical/Important findings.

### Not touched by this increment
`pyproject.toml` / `uv.lock` are unchanged — the pre-existing `cpex-pii-filter>=0.3.5` PyPI-version-pin CI issue (documented in this PR's description, cross-repo dependency on `cpex-plugins#128`) is unrelated to G2 and untouched by these 3 commits.

vishu-bh (changes requested):
Thanks for the updates. The overall direction looks good: the gateway remains the telemetry/export authority, plugins only return metadata, and the new optional OTel SDK sink helps address the external collector use case without making plugins depend on OTel.

I still think this needs a few changes before merge:

1. **Reset observability ContextVars after each request**

   `ObservabilityMiddleware` sets `current_trace_id`, CPEX `plugins_trace_id`, and `current_span_id`, but I do not see the tokens reset in a `finally` block. Since `build_request_extensions()` now reads these values, stale trace/span context could bleed into later plugin hook executions and misattribute plugin telemetry.

2. **Sanitize and bound plugin names / metric keys**

   `plugin_name` and `field_name` currently flow into DB span names, metric names, resource IDs, OTel span names, and OTel attribute keys. These values are plugin-controlled. Please add a strict identifier contract before exporting anywhere: regex allowlist, max length, and drop/fallback behavior for invalid names.

3. **Clarify or configure DB vs OTel sink behavior**

   The new OTel SDK sink is additive. `record_plugin_metrics()` still writes internal observability DB spans/metrics first, then optionally exports to OTel. That means the database growth concern remains. Can we either add sink configuration, for example internal DB on/off, numeric metric rows on/off, OTel export on/off, or document clearly that DB storage is mandatory for this issue and OTel is only an additional export path?

4. **Reduce DB write amplification**

   The primary span path is batched, but every numeric field still calls `ObservabilityService.record_metric()`, which opens its own independent session/commit. With current caps this can still mean many DB writes per hook result. Please consider batching, a stricter numeric-field cap, sampling, or making the secondary metric-row sink configurable.

5. **Apply bounds before materialization/full scans**

   `record_plugin_metrics()` materializes all metadata entries before applying `_MAX_PLUGINS_PER_CALL`; `_sanitize_plugin_metrics()` keeps iterating after `_MAX_PLUGIN_KEYS`; and list handling scans the full list before slicing. Since plugin metadata is untrusted, please use bounded iteration and early breaks.

6. **Tighten the accepted metadata value contract**

   The PII tests prove the current `pii_filter` emits safe fields, but the gateway still accepts arbitrary string metadata from any plugin and exports it to internal observability and OTel. Truncation handles size, not sensitivity. I’d prefer numeric/bool plus explicit low-cardinality enum/list fields, or per-plugin allowlisted schemas for string fields.

7. **Clarify hook coverage for #5458**

   This PR wires tool and prompt hook paths, but resource/auth/RBAC/A2A plugin hook families still appear outside the new trace-extension/metadata-recording path. If that scope is intentional, please call it out in the PR/docs. If #5458 expects all gateway plugin hook call sites, this looks incomplete.

8. **Resolve the cross-repo dependency gap**

   `plugins/config.yaml` references the `cpex_pii_filter` behavior from `0.3.6`, but `pyproject.toml` and `uv.lock` still resolve `cpex-pii-filter` `0.3.5`. This should be resolved before merge or this PR should remain blocked until the upstream plugin release and lockfile bump land.

9. **Add/update docs for the new gateway/plugin observability contract**

   I do not see docs updated for this behavior. We should document:
   - `PluginResult.metadata["<plugin_name>"]` shape
   - supported value types and limits
   - that plugin metadata is untrusted and must be low-cardinality/non-sensitive
   - DB sink vs optional OTel sink behavior
   - production implications for DB growth, sampling, and retention
   - hook coverage and current limitations
   - dependency requirement on the plugin-side `cpex-pii-filter` metrics behavior

One additional question: the new OTel sink gates on `otel_tracing_enabled()`. Should it also require an active OTel context, e.g. `otel_context_active()`, so plugin metric spans do not become orphan root spans when a tracer is configured but there is no active request span?

Overall, I think the architecture is close and adaptable, but the telemetry contract and sink behavior need to be tightened before this is safe to merge.

lucarlig (dismissed):
approach is sound. Looks like plugin is properly isolated no need external service and everything goes through the metadata.
LGTM

msureshkumar88:
Thanks for the detailed review — pushed e1a9f13 addressing all nine points:

1. **ContextVar reset** — `current_trace_id`, `plugins_trace_id`, and `current_span_id` are now reset via `.reset()` in a `finally` block in `ObservabilityMiddleware.dispatch()`, covering both the normal path and the trace-setup-failure path.
2. **Identifier validation** — added a bounded identifier charset (`^[A-Za-z0-9_.-]{1,64}$`) and validate both `plugin_name` and metric field names against it before they're used in span/metric names, resource ids, or OTel attribute keys.
3. **DB vs OTel sink config** — added `plugin_metrics_db_spans_enabled` and `plugin_metrics_db_numeric_rows_enabled` settings so each internal DB sink can be toggled independently of the OTel export sink.
4. **DB write amplification** — added `plugin_metrics_max_numeric_per_call` (default 16), capping numeric `ObservabilityMetric` rows written per `invoke_hook()` call across all plugins.
5. **Bounded iteration** — switched to `itertools.islice` instead of materializing the full item list first, `break` (not `continue`) once the per-plugin key cap is hit, and bounded the list-value str-check to the first 32 items before slicing.
6. **String value contract** — strings (and joined list values) must now match a low-cardinality charset (`^[A-Za-z0-9_.,:=/-]*$`) with max length down to 64 (was 256); anything outside that is dropped rather than truncated.
7. **Hook coverage** — wired the remaining 8 of 17 `invoke_hook()` call sites (`auth.py`, `http_auth_middleware.py` x2, `rbac.py`, `resource_service.py` x2, `a2a_service.py` x2) with `extensions=build_request_extensions()` + `record_plugin_metrics()`, matching the existing tool/prompt service pattern. All 17 sites from #5458 are covered now.
8. **cpex-pii-filter version gap** — not resolved yet: `0.3.6` still isn't published to PyPI and `cpex-plugins#128` is still open, so `pyproject.toml`/`uv.lock` are left alone per the documented merge order rather than pinning to a version that doesn't exist yet.
9. **Docs** — added a section to `docs/docs/manage/observability/internal-observability.md` covering the `result.metadata` shape, value/limit contract, DB-vs-OTel sink behavior (incl. the new config flags), hook coverage, and the plugin-version dependency.

Also added the `otel_context_active()` guard from your last question, so a configured-but-unused tracer won't produce orphan root spans.

Side note: wiring the remaining call sites surfaced a few pre-existing test mocks (`test_resource_service.py`, `test_a2a_agent_invoke_hooks.py`, `test_http_auth_headers.py`) missing `.metadata` or the new `extensions=` param — same class of issue as the `test_tool_service_coverage.py` fixes earlier in this branch. Fixed those too.

lucarlig (dismissed):

mcpgateway/plugins/utils.py
```diff
@@ -8,11 +8,339 @@
...
+        name: Candidate identifier (plugin name or metric field name); treated as untrusted.
+
+    Returns:
+        True if ``name`` is a string matching the bounded, restricted-charset
+        identifier contract; False otherwise.
+    """
+    return isinstance(name, str) and bool(_IDENTIFIER_RE.match(name))
+
+
+def _validate_string_value(value: str) -> Optional[str]:
```
vishu-bh, line 120: This accepts arbitrary safe-character strings and then truncates them before exporting to DB/OTel. Secret-like values such as JWT fragments, API keys, hashes, or opaque IDs often match this charset, so truncation limits size but does not prevent leakage. Please reject strings longer than the max length before regex/export, or move to explicit allowlisted string fields.

mcpgateway/plugins/utils.py
```diff
@@ -8,11 +8,339 @@
...
+        return {}
+
+    sanitized: Dict[str, Any] = {}
+    for key, value in raw_metrics.items():
+        if not _is_valid_identifier(key):
+            logger.debug("Dropped a plugin metric key for %r: key is not a valid identifier", plugin_name)
+            continue
+        if len(sanitized) >= _MAX_PLUGIN_KEYS:
+            logger.debug("Stopping metric scan for %r: key limit exceeded", plugin_name)
+            break
```
vishu-bh: This cap only stops after 32 accepted metrics. A plugin can send many invalid or dropped keys first and still force an unbounded scan of the metadata dict. Since this is untrusted plugin output, please bound inspected items too, for example with `itertools.islice(raw_metrics.items(), _MAX_PLUGIN_KEYS)` plus a separate accepted-count cap if needed.

mcpgateway/plugins/utils.py
```diff
@@ -8,11 +8,339 @@
...
+                logger.debug("Dropped plugin metric %r for %r: string value not in safe low-cardinality contract", key, plugin_name)
+        elif isinstance(value, list):
+            prefix = value[:_MAX_LIST_ITEMS]
+            if all(isinstance(item, str) for item in prefix):
+                joined = ",".join(prefix)
+                validated = _validate_string_value(joined)
+                if validated is not None:
+                    sanitized[key] = validated
+                else:
+                    logger.debug("Dropped plugin metric %r for %r: joined list value not in safe low-cardinality contract", key, plugin_name)
```
vishu-bh: Only item count is capped here. The first 32 string items can still be very large, then get joined and regex-scanned before any length bound is enforced. Please cap or reject each list item before joining, and drop the list if the joined value would exceed the string limit.

vishu-bh (changes requested):
Thanks for addressing the previous review comments. Most of the main concerns now look fixed: ContextVars are reset, plugin/metric identifiers are bounded, DB sinks are configurable, numeric metric rows are capped, all 17 hook sites appear wired, OTel export now requires an active OTel context, and docs were added.

There some minor things noticed.

msureshkumar88:
Thanks for the follow-up review! Addressed all 3 in `698d35f`:

- **`_validate_string_value`**: now rejects strings over `_MAX_STRING_LENGTH` outright (checked before the charset regex) instead of truncating them, so a secret/token/hash fragment that happens to match the safe charset gets dropped, not exported at reduced length.
- **Key scan**: `_sanitize_plugin_metrics` now iterates `itertools.islice(raw_metrics.items(), _MAX_PLUGIN_KEYS)` — the scan itself is bounded to the first 32 items, not just the accepted count, so a dict front-loaded with invalid keys can't force a full scan.
- **List values**: each item is now checked against `_MAX_STRING_LENGTH` before joining; the list is dropped if any item is oversized, so the join + regex validation is always bounded by `_MAX_LIST_ITEMS * _MAX_STRING_LENGTH` rather than by attacker-controlled item sizes.

Added 3 new unit tests covering these (reject-not-truncate, bounded-scan, oversized-list-item-rejected). Full `tests/unit/mcpgateway/plugins/` suite green, `ruff check`/`ruff format --check` clean.

mcpgateway/plugins/utils.py
msureshkumar88: Fixed in 69047a7. `_validate_string_value` now rejects oversized strings outright (return `None`) instead of truncating -- avoids leaking truncated secret/token fragments. Added `test_oversized_string_is_rejected_not_truncated`.

mcpgateway/plugins/utils.py
msureshkumar88: Fixed in 69047a7. Inner loop now uses `itertools.islice(raw_metrics.items(), _MAX_PLUGIN_KEYS)` so only the first 32 raw items are ever scanned, regardless of accept/reject outcome -- matches the pattern already used for the outer per-plugin loop. Added `test_key_scan_is_bounded_not_just_accepted_count`.

mcpgateway/plugins/utils.py
msureshkumar88: Fixed in 69047a7. Each list item is now length-checked against `_MAX_STRING_LENGTH` before the join, so oversized items are rejected pre-join instead of only being length-capped by count. Added `test_oversized_list_item_is_rejected_before_join`.

msureshkumar88:
Fixed CI failure in 1e51567: `pytest (py3.12)` was failing on `tests/e2e/test_otel_plugin_metadata_e2e.py` -- no `plugin.metrics.pii_filter` span was ever created.

Root cause: `cpex-pii-filter` 0.3.6 is the first release that populates `result.metadata` on `tool_post_invoke` (`total_detections`/`total_masked`/`detection_types`) -- the locked 0.3.5 build masks PII correctly but never emits metadata. This repo's `[tool.uv.exclude-newer-package]` soak-window pin for `cpex-pii-filter` (2026-06-29) predates 0.3.6's publish date (2026-07-07), so `uv.lock` was correctly staying on 0.3.5 per that policy -- it just happens this PR's own e2e test needs the newer release. Bumped the cutoff for `cpex-pii-filter` specifically (with a comment explaining why), re-locked, verified both e2e tests in that file pass locally against 0.3.6.

pyproject.toml
lucarlig: explicitly  set this to `>=0.3.6` is better

lucarlig (changes requested):

pyproject.toml
msureshkumar88: Fixed in f3c55f9. Bumped the constraint to `cpex-pii-filter>=0.3.6` explicitly -- >=0.3.5 was technically satisfied by a build that never emits `result.metadata` on `tool_post_invoke`, so the real minimum should be stated rather than relied on implicitly via the lockfile's soak-window cutoff.

lucarlig (dismissed):
LGTM

plugins/config.yaml
```diff
@@ -1218,3 +1218,26 @@ plugins:
...
+  # usage than "permissive") is safe here because block_on_detection is false below — detections
+  # are masked/redacted, not blocking, so traced tool calls still succeed end-to-end.
+  - name: "PIIFilter"
+    kind: "cpex_pii_filter.pii_filter.PIIFilterPlugin"
+    description: "Rust-backed PII detection and masking for prompt arguments, tool inputs, and tool outputs"
+    version: "0.3.6"
+    author: "ContextForge Contributors"
+    hooks: ["prompt_pre_fetch", "prompt_post_fetch", "tool_pre_invoke", "tool_post_invoke"]
+    tags: ["pii", "security", "masking", "observability"]
+    mode: "enforce"
```
vishu-bh: should this be enforce?

mcpgateway/plugins/utils.py
```diff
@@ -8,11 +8,349 @@
...
+    for key, value in itertools.islice(raw_metrics.items(), _MAX_PLUGIN_KEYS):
+        if not _is_valid_identifier(key):
+            logger.debug("Dropped a plugin metric key for %r: key is not a valid identifier", plugin_name)
+            continue
+
+        if isinstance(value, bool):
+            sanitized[key] = value
+        elif isinstance(value, (int, float)):
+            sanitized[key] = value
+        elif isinstance(value, str):
```
vishu-bh, line 206: The current string sanitizer is size- and charset-bounded, but it still allows short secret/PII-shaped values such as `123-45-6789` or token fragments. Since these values flow into DB span attributes and can also be exported to OTel, a buggy or untrusted plugin could still leak raw sensitive data through `result.metadata`. Could we deny string metadata by default and only allow explicit known-safe enum fields, e.g. `stage` or `detection_types`, or keep this path numeric/bool-only?

vishu-bh (changes requested):
Minor checks

plugins/config.yaml
msureshkumar88: Fixed in cdbaec9. `mode: "enforce"` isn't a valid `cpex.framework.PluginMode` value on the static config-load path -- that legacy label is only recognized by the gateway's runtime mode-override translation (`_LEGACY_MODE_TO_PLUGIN_MODE` in `gateway_plugin_manager.py`, applied to Redis/local overrides after initial load), not `ConfigLoader.load_config()`. Confirmed empirically: `PluginConfig(mode="enforce")` was silently falling back to `mode=sequential` already. Made that explicit (`mode: "sequential"`) -- no behavior change, just removes reliance on the silent fallback.

mcpgateway/plugins/utils.py
msureshkumar88: Fixed in cdbaec9. Added an explicit deny-by-default field-name allowlist (`_SAFE_STRING_FIELD_NAMES = {"stage", "detection_types"}`, the two fields `pii_filter` actually emits): a string (or joined list[str]) value is now only accepted when its field name is in that allowlist, regardless of whether it passes the charset/length check. Confirmed the gap first -- `_validate_string_value("123-45-6789")` and a fake API-key-shaped string both passed the old charset-only check unchanged. New plugins adding a string field must extend the allowlist explicitly (tracked in #5554 / cpex-plugins#129).

plugins/config.yaml
vishu-bh: This should be `disabled` not `sequential`

mcpgateway/plugins/utils.py
```diff
@@ -8,11 +8,371 @@
...
+    sanitized: Dict[str, Any] = {}
+    # Bound the scan itself (not just the accepted count): only the first
+    # _MAX_PLUGIN_KEYS items of the raw dict are ever inspected, so a dict
+    # front-loaded with invalid/dropped keys can't force an unbounded scan.
+    for key, value in itertools.islice(raw_metrics.items(), _MAX_PLUGIN_KEYS):
+        if not _is_valid_identifier(key):
+            logger.debug("Dropped a plugin metric key for %r: key is not a valid identifier", plugin_name)
+            continue
+
+        if isinstance(value, bool):
```
vishu-bh, line 195: This still needs one more tightening pass. The string allowlist helps for `user_ssn: "123-45-6789"`, but numeric metadata is still accepted under any valid field name. A plugin can return something like `{"ssn": 123456789}` or `{"account_id": 123456789012}` and that value will be copied into the sanitized map, then written to DB span attributes, numeric metric rows, and the OTel export sink.

Since `result.metadata` is treated as untrusted plugin output, the same semantic allowlist should apply to numeric fields too. For this PR’s current contract, I’d expect only known-safe counter fields such as `total_detections` / `total_masked` to survive, ideally with finite-value checks. Otherwise a buggy or adversarial plugin can still leak numeric PII/identifiers through observability even though string values are now mostly gated.

vishu-bh (changes requested):

msureshkumar88:
Fixed in 6db4791: `PIIFilter` was the only bundled `cpex-*` plugin shipped active by default in `plugins/config.yaml` -- every other one (rate-limiter, secrets-detection, url-reputation, encoded-exfil-detection, etc.) ships `mode: "disabled"`, opt-in rather than opt-out. Flipped it to `disabled` for consistency, since masking tool/prompt content by default would be a silent behavior change for existing deployments on upgrade (a legitimate email/SSN-shaped value in a response would now become `[REDACTED]` with no opt-in). The e2e test now builds a private patched copy of `plugins/config.yaml` (PIIFilter forced to `mode: "sequential"`) to still exercise the full G0/G1/G2 pipeline, rather than depending on the shipped default.

mcpgateway/plugins/utils.py
msureshkumar88: Fixed in 0529e97. Added a deny-by-default field-name allowlist for numeric metadata too (`_SAFE_NUMERIC_FIELD_NAMES = {"total_detections", "total_masked"}`, the only two numeric fields `pii_filter` actually emits, confirmed against the Rust source), mirroring the string allowlist added in cdbaec9. A number under any other field name -- e.g. `{"ssn": 123456789}` -- is now dropped regardless of value. Also added a finite-value check (rejects NaN/inf) since those can't be meaningfully aggregated once exported as a metric/span attribute. New plugins adding numeric metadata fields must extend the allowlist explicitly (tracked in #5554 / cpex-plugins#129, same as the string allowlist).

plugins/config.yaml
msureshkumar88: Fixed in 6db4791 -- PIIFilter now ships `mode: "disabled"` by default, matching every other bundled cpex-* plugin (opt-in, not opt-out). Masking a tool response by default would've been a silent behavior change for existing deployments that may legitimately expect an email/SSN-shaped value to pass through unmodified. Set `mode: "sequential"` to enable it.

tests/integration/plugins/test_plugin_metrics_consumer_integration.py
```diff
@@ -0,0 +1,216 @@
...
+
+        record_plugin_metrics(trace_id, result_metadata)
+
+        spans = db_session.query(ObservabilitySpan).filter_by(trace_id=trace_id).all()
+        span_names = {s.name for s in spans}
+        assert span_names == {"plugin.metrics.pii_filter", "plugin.metrics.secrets_detection"}
+
+        secrets_span = next(s for s in spans if s.name == "plugin.metrics.secrets_detection")
+        assert secrets_span.attributes["total_detections"] == 7
+        assert secrets_span.attributes["severity"] == "high"
```
vishu-bh: expects severity string metadata to persist, but code now allows only `stage` and `detection_types`.

tests/integration/plugins/test_plugin_metrics_consumer_integration.py
```diff
@@ -0,0 +1,216 @@
...
+                "good_field": "keep-me",
+                "huge_field": huge_value,
+                "nested_dict": {"should": "be-dropped"},
+                "mixed_list": ["a", 1, None],
+            }
+        }
+
+        record_plugin_metrics(trace_id, result_metadata)
+
+        span = db_session.query(ObservabilitySpan).filter_by(trace_id=trace_id, name="plugin.metrics.custom_plugin").one()
```
vishu-bh, line 192: This test still expects arbitrary string fields and oversized strings to survive/truncate, but the implementation now drops non-allowlisted string fields and rejects overlong strings. In `--with-integration` mode this raises `NoResultFound` because no sanitized attributes remain, so no `plugin.metrics.custom_plugin` span is created. Please update this test to assert the new deny-by-default/drop behavior.

mcpgateway/plugins/utils.py
```diff
@@ -8,11 +8,415 @@
...
+
+    Args:
+        value: Candidate ``int``/``float`` value (never ``bool`` -- callers check that first).
+
+    Returns:
+        ``value`` unchanged if it is a finite number, otherwise ``None`` (reject).
+        ``NaN``/``inf``/``-inf`` are rejected outright since they can't be compared or
+        aggregated meaningfully once exported as a metric/span attribute.
+    """
+    if not math.isfinite(value):
```
vishu-bh: `math.isfinite()` can raise `OverflowError` for very large Python ints, e.g. `10**100000`. Since this value is untrusted plugin metadata, one allowlisted numeric field can escape `_validate_numeric_value()` and make the outer `record_plugin_metrics()` swallow the entire hook's plugin metrics. Please catch `OverflowError`/`TypeError` here and reject by returning `None`.

docs/docs/manage/observability/internal-observability.md
```diff
@@ -785,6 +785,92 @@ WHERE start_time < NOW() - INTERVAL '7 days';
...
+
+`result.metadata` is treated as **untrusted plugin output** and validated before it is recorded
+anywhere (`mcpgateway.plugins.utils._sanitize_plugin_metrics()`). None of these bounds are
+configurable — a plugin that needs different limits should be revisited rather than the gateway
+loosening its defaults:
+
+| Element | Contract |
+|---|---|
+| Plugin name / field name (dict key) | Must match `^[A-Za-z0-9_.-]{1,64}$`; otherwise dropped |
+| `bool` / `int` / `float` value | Recorded as-is |
```
vishu-bh: Docs say `bool`/`int`/`float` values are recorded as-is, but current code only accepts numeric fields listed in `_SAFE_NUMERIC_FIELD_NAMES` and rejects non-finite values. Please update this table so plugin authors/operators do not rely on a broader contract than the implementation supports.

docs/docs/manage/observability/internal-observability.md
```diff
@@ -785,6 +785,92 @@ WHERE start_time < NOW() - INTERVAL '7 days';
...
+| `str` value | Must match the low-cardinality charset `^[A-Za-z0-9_.,:=/-]*$` and be ≤ 64 chars (truncated if longer, dropped if the charset doesn't match) — this is intentionally **not** a free-text channel |
```
vishu-bh: Docs say overlong strings are truncated, but current code rejects overlong strings and only accepts allowlisted string field names. Please update this row to reflect the deny-by-default string allowlist and reject-not-truncate behavior.

vishu-bh (changes requested):
Core architecture now much better, just some minor test and doc update is needed

tests/integration/plugins/test_plugin_metrics_consumer_integration.py
msureshkumar88: Fixed in da92c330884b3caf71509cd3c88d3d105a209662. "severity" is not in _SAFE_STRING_FIELD_NAMES ({"stage", "detection_types"}), so the test now asserts it's dropped instead of expecting it to persist.

tests/integration/plugins/test_plugin_metrics_consumer_integration.py
msureshkumar88: Fixed in da92c330884b3caf71509cd3c88d3d105a209662. Updated to use an allowlisted field ("stage") to show it persists, and assert "good_field" (non-allowlisted) and "huge_field" (overlong) are both dropped -- not truncated -- matching the deny-by-default/reject-not-truncate behavior.

mcpgateway/plugins/utils.py
msureshkumar88: Fixed in da92c330884b3caf71509cd3c88d3d105a209662. _validate_numeric_value() now catches OverflowError/TypeError around math.isfinite(), so a huge int like 10**100000 is rejected (returns None) instead of raising and aborting the rest of the hook's metrics.

docs/docs/manage/observability/internal-observability.md
msureshkumar88: Fixed in da92c330884b3caf71509cd3c88d3d105a209662. Table now states int/float values are only recorded when the field name is in the numeric allowlist (total_detections, total_masked); non-finite values are always rejected.

docs/docs/manage/observability/internal-observability.md
msureshkumar88: Fixed in da92c330884b3caf71509cd3c88d3d105a209662. Table now states string values are only recorded when the field name is in the string allowlist (stage, detection_types), and overlong values are rejected outright, not truncated.

msureshkumar88:
DCO check failed -- last commit was missing a `Signed-off-by` trailer. Amended with `--signoff` and force-pushed (4e2b655c3bbb5b68f1e85a54fbf2b351cc3d01de).

vishu-bh (dismissed):
Thanks @msureshkumar88 for addressing the comments

PR is clean and architecturally correct. LGTM 🚀

lucarlig (approved):

vishu-bh (approved):

brian-hussey (approved):
Approving as process has been followed and reviewed by team members.

PR #5475, opened 2026-07-02
feat(ui-rewrite): delete a resource with optimistic update on the Resources page

### LINKED ISSUE #5099: [UI-REWRITE]: Delete Resources

gcgoncalves:
# ✨ Feature / Enhancement PR

https://github.com/user-attachments/assets/91060e47-d802-49e9-b3e1-24b1985850b4

## 🔗 Epic / Issue

Closes #5099

---

## 🚀 Summary (1-2 sentences)

Adds functionality to the resource close button.

---

## 📏 Reviewability

---

## 🧪 Checks

- [X] `make lint` passes
- [X] `make test` passes

### DIFF
--- a/client/e2e/resources.spec.ts
+++ b/client/e2e/resources.spec.ts
@@ -213,4 +213,130 @@ test.describe("Resources page", () => {
     const cards = page.locator('[data-slot="card"]');
     await expect(cards).toHaveCount(3);
   });
+
+  test.describe("Delete resource", () => {
+    test("cancel in confirm dialog keeps resource visible", async ({ page }) => {
+      await page.route("**/resources?*", async (route) => {
+        await route.fulfill({
+          status: 200,
+          contentType: "application/json",
+          body: JSON.stringify([MOCK_RESOURCE]),
+        });
+      });
+
+      await page.goto(APP.RESOURCES);
+      await page.waitForLoadState("networkidle");
+
+      await expect(page.getByText("Test Document")).toBeVisible();
+
+      // Open card dropdown
+      await page.getByLabel("More options for Test Document").click();
+      await page.getByText("Delete").click();
+
+      // Confirm dialog should appear
+      await expect(page.getByText("Delete resource")).toBeVisible();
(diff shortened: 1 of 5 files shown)


marekdano (commented):
### Suggestions (non-blocking)

1. **Focus management after delete (accessibility)** — This is the most substantive point. When deleting via the card dropdown, Radix will try to restore focus to the dropdown trigger (the "More options" button) when the `ConfirmDialog` closes. But that button lives on the card, which is removed optimistically — so focus falls back to `<body>`, losing keyboard/screen-reader position. Consider moving focus to a stable element (e.g. the page heading or grid container) after a successful delete, or announcing the removal via an `aria-live` region. The details-panel path has the same concern since the panel is closed (`setSelectedResource(null)`) on delete.

2. **Double close of the dialog** — `confirmDelete` calls `setDeleteDialogOpen(false)`     itself, and `ConfirmDialog.handleConfirm` also calls `onOpenChange(false)` (default `closeOnConfirm=true`). Harmless today, but redundant; you could pass `closeOnConfirm={false}` and let `confirmDelete` own closing, to keep the source of truth in one place.
 
3. **`__optimistic__` deletability** — `validateResourceId` allows underscores, so an     optimistically-added (not-yet-persisted) resource with id `__optimistic__` would     pass validation and hit the server as a 404. Edge case, but worth guarding the Delete action against the optimistic placeholder id if that card can be deleted before its POST resolves.

marekdano (approved):
### Non-blocking
- `onCloseAutoFocus` redirects focus on every close path, not just after a delete `client/src/pages/Resources.tsx`

`onCloseAutoFocus` fires whenever the dialog closes. The handler unconditionally `preventDefault()`s and focuses the heading — but on Cancel / Esc / backdrop click, nothing was removed and the trigger button still exists, so Radix would normally restore focus to it. As written, those paths send focus up to the page `<h1>` instead of back to the control the user just used. The comment ("removed optimistically") only holds for the confirm path.

Worth gating the redirect on whether a delete actually ran, so the cancel path keeps Radix's default trigger-restore behavior.

marekdano (approved):
LGTM 🚀

PR #5481, opened 2026-07-02
Package all js libs

### LINKED ISSUE #5527: [BUG]: Production (gunicorn) E2E run - Edit Gateway OAuth toggle, server delete, RBAC invite cascade
### 🐞 Bug Summary
Full Playwright E2E suite (921 tests) run against a production-mode instance (`make serve` / gunicorn, multi-worker, `:4444`) surfaced 3 failures and 17 setup errors. This issue bundles all of them for tracking.

---

### 🧩 Affected Component
- [x] `mcpgateway` - API

---

### 🔁 Steps to Reproduce / Details

**1. Edit Gateway modal — OAuth fields don't hide when auth type is switched away from OAuth**
`tests/playwright/entities/test_gateways_extended.py::TestGatewayEditModal::test_edit_modal_auth_type_oauth_fields`
- Open **Edit Gateway** on a gateway configured with Authentication Type = OAuth 2.0.
- Switch Authentication Type to something else (e.g. Basic/Bearer/None).
- Expected: the OAuth-specific field group (`#auth-oauth-fields-gw-edit` — Grant Type, Issuer URL, Client ID, etc.) hides.
- Actual: field group stayed visible across a 5s poll (14 repeated checks, all "visible"). Suggests the dropdown-change handler that toggles field visibility isn't wired up (or didn't fire) in the **edit** modal, unlike the add-gateway modal.

**2. Server delete — row still present/inactive after delete**
`tests/playwright/entities/test_servers.py::TestServersCRUD::test_delete_server`
- `assert False` from the `ServersPage.delete_server()` page-object helper's post-delete check.
- Screenshot shows the Virtual Servers table with "Show Inactive" checked and the target server still listed. Could be a genuine delete failure, or the UI performs a soft-delete/deactivate and the test doesn't account for the "Show Inactive" filter — needs manual repro to distinguish.

**3. 17 `test_rbac_permissions.py` setup errors** (all classes: TestRBACGatewayCreate, TestRBACServerCreate, TestRPCToolExecutionRBAC, TestRBACToolOperations, TestRBACPromptOperations, TestRBACResourceOperations, TestRBACTeamManagement)
- Setup fails with `"An active invitation already exists for rbac-e2e-*@test.example.com"` (400) or `"Failed to accept invitation"` (500).
- Root cause of the 500 is the unhandled `IntegrityError` on duplicate team membership, already filed separately as #5524. The 400s are test-data staleness (leftover invitation rows from repeated local runs against the same SQLite DB) rather than a fresh product bug on their own, but worth noting since they block the RBAC regression suite from running cleanly on a reused DB.

Also observed (test-infra artifact, not filed as a bug): `test_login_with_correct_credentials_must_succeed` fails under parallel workers because they share one bootstrap admin account and race on its forced password-change flow — not a real auth bug, just a limitation of testing one shared admin account with `-n 4`+.

---

### 🤔 Expected Behavior
- Edit Gateway modal's field visibility should react to Authentication Type changes same as the add-gateway modal.
- Server delete should either fully remove/deactivate the row consistently with what the UI's "Show Inactive" filter implies, or the row should disappear when that filter is off.
- RBAC E2E setup should not be blocked by pre-existing invitation state (either the suite should clean up after itself, or invite-acceptance should handle "already a member" gracefully — see #5524).

---

### 📓 Logs / Error Output
```
# (1)
AssertionError: Locator expected to be hidden
Actual value: visible
Call log:
  - waiting for locator("#auth-oauth-fields-gw-edit")
    14 × locator resolved to <div id="auth-oauth-fields-gw-edit">…</div>
       - unexpected value "visible"

# (2)
AssertionError: assert False
 +  where False = delete_server(<Page url='http://localhost:4444/admin/?include_inactive=true#catalog'>, 'd2521937055341e09a6880cc5909b6a5')

# (3, example)
AssertionError: Failed to invite rbac-e2e-dev-740d3d73@test.example.com: 400 {"detail":"An active invitation already exists for rbac-e2e-dev-740d3d73@test.example.com"}
```

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Version or commit | `Release/v1.0.5 @ d71dd4791` |
| Runtime | Python 3.11, gunicorn (`./run-gunicorn.sh`, multi-worker, production mode) |
| Platform / OS | Linux (WSL2) |
| Container | none |

---

Found via the project's own `tests/playwright/` E2E suite run against `make serve` with `RATE_LIMITING_ENABLED=false` (otherwise parallel test workers trip the in-memory rate limiter and get 429s — separate from these 3 findings). The same suite run against the dev server (`make dev`, uvicorn) passed 666/921 with a different, non-overlapping set of 3 failures (already filed as #5524, #5525, #5526), so items 1 and 2 above appear specific to this run / worker distribution and should be reproduced manually to confirm.

gcgoncalves:
# Pull Request

## 🔗 Related Issue

Closes https://github.ibm.com/contextforge-org/internal_issues/issues/387

---

## 📝 Summary

Packages all JS packages, and then chunks the package to prevent a massive bundle.
---

## 📏 Reviewability

---

## 🏷️ Type of Change

- [ ] Bug fix
- [X] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

Run `make build-ui dev` then test the app on localhost. Make sure there are no errors on the browser console.

| Lint suite                | `make lint`     |    ✅   |
| Unit tests                | `make test`     |   ✅    |
| Coverage ≥ 80%            | `make coverage` |    ✅   |

---

## ✅ Checklist

- [X] Tests added/updated for changes


mcpgateway/templates/admin.html
```diff
@@ -238,7 +238,7 @@
       window.htmxConfig = window.htmxConfig || {};
       window.htmxConfig.inlineScriptNonce = "{{ csp_nonce(request) }}";
     </script>
-    <script defer src="{{ root_path }}/static/{{ bundle_js }}"></script>
+    <script type="module" src="{{ root_path }}/static/{{ bundle_js }}"></script>
```
madhu-mohan-jaishankar, line 163: Since this PR removes the vendored asset download/build path, I think the ui_airgapped branch here may need to be updated too. It still references /static/vendor/..., so offline mode appears to depend on files that is no longer produced.

madhu-mohan-jaishankar (changes requested):

mcpgateway/templates/admin.html
cafalchio: Could you also update the packages in this PR? Please?
```
npm update
npm audit
npm audit fix
```

marekdano (changes requested):
A few findings:

## Blocking

  **1. The bundled CSS (Font Awesome + CodeMirror) never gets linked into the page.** `admin.js` now imports the vendor CSS (`import 'codemirror/lib/codemirror.css'`, `import '@fortawesome/.../all.min.css'`, etc.), and Vite does emit it — but only as separate files that the entry is expected to `<link>` from HTML. After `npm run vite:build`, the manifest looks like:

```json
  "mcpgateway/admin_ui/index.js": {
      "isEntry": true,
      "file": "bundle-3LZTBtVw.js",
      "css": ["assets/index-RTp56Tkr.css"]   // contains Font Awesome
  }
```

`get_bundle_js_filename()` (`admin.py`) reads only `["file"]`, so `assets/index-*.css` and `assets/vendor-editor-*.css` (the latter holds the `.CodeMirror` rules) are emitted but never referenced by any template or auto-injected by the bundle. In practice that means CodeMirror editors render unstyled and Font Awesome doesn't load. Would it work to have the manifest reader also return the entry's `css` array and emit a `<link>` for each?

  **2. The `vendor/fontawesome` path the templates still point at is no longer produced.**
`admin.html`, `login.html`, and `change-password-required.html` all keep `<link ... href=".../static/vendor/fontawesome/css/all.min.css">`, but `download-cdn-assets.sh` and its `Containerfile` step are removed and `static/vendor/` isn't committed, so that URL 404s. This is the same thing @madhu-mohan-jaishankar flagged. It hits `login.html` / `change-password-required.html` hardest since they don't load the JS bundle at all — even once #1 is fixed, those two pages have no Font Awesome source. Might be worth linking the Vite-emitted CSS on those pages too, or otherwise giving them a real path.

  ## Suggestions

  **3. The lazy-loading currently doesn't split anything.** The build reports:

  ```
[INEFFECTIVE_DYNAMIC_IMPORT] tools.js/servers.js/gateways.js/teams.js/llmChat.js/ logging.js/metrics.js/plugins.js is dynamically imported by lazy-loader.js but also statically imported by admin.js — dynamic import will not move module into another chunk.
  ```

Because `admin.js` still statically imports these modules, Rollup keeps them in the eager graph, so the tab-click loader + loading indicators don't actually defer anything yet. Similarly, `chunk-vendor-charts` (~200 KB) is pulled in eagerly via the top-level `import { Chart }` in `admin.js`, so Chart.js now loads for non-admins too (it used to be gated behind `{% if is_admin %}`). If deferral is a goal here, the static import chains into these modules would need to be broken first.

  **4. CSP still allowlists CDNs that are no longer used.** Now that nothing loads from `cdnjs`/`jsdelivr`/`unpkg`, `security_headers.py` (`script-src-elem`, `style-src`, `font-src`) could drop those origins as a nice follow-up hardening. Not blocking.

  **5. A couple of test gaps.** The `tabs.test.js` update mocks `loadFeature` to resolve instantly, so `lazy-loader.js` itself (dedup, in-flight coalescing, the error path) and the loading-indicator show/hide aren't exercised. A small unit test for the loader would be a good add.

marekdano (commented):
### Just two more findings

 1. Lazy-loading is a no-op — every "lazy" chunk still loads on initial page load **`mcpgateway/admin_ui/admin.js`**

`admin.js` (the eagerly-evaluated entry) statically imports `gateways.js` (L188), `llmModels.js` (L228), `logging.js` (L278), `plugins.js` (L342), `servers.js` (L427), `teams.js` (L459), and `tools.js` (L505). Because these are static imports of the entry, Vite emits them as static `import` statements in `bundle-*.js`, so the browser fetches and evaluates all of those chunks during initial load — `loadFeature()` at tab-click time just re-resolves already-loaded modules.

`metrics.js` and `llmChat.js` (the only two not imported by name) are grouped by `manualChunks` into the `monitoring` chunk (with statically-imported `logging.js`) and the `llm` chunk (with statically-imported `llmModels.js`), so those chunks load eagerly too.

**Net effect:** the initial payload isn't reduced, and `lazy-loader.js`, `TAB_FEATURE_MAP`, the loading-indicator UI, and the `await loadFeature(...)` path in `showTab` add complexity and startup cost while deferring nothing. To make lazy-loading real, remove the static feature imports (and their `window.Admin.*` assignments) from `admin.js`; otherwise drop the lazy machinery and reword the PR description.

2. Air-gapped docs point at build paths that don't exist **`docs/docs/overview/ui.md`, `docs/docs/deployment/container.md`**

`ui.md` states all three container builds "include the Vite-built Admin UI assets via the `frontend-builder` stage," and `container.md` says to "Use `Containerfile.lite` which automatically downloads remaining vendor assets during build." At the PR head, only `Containerfile` has a `frontend-builder` stage and copies `static/`; `Containerfile.lite` and `Containerfile.scratch` reference neither `frontend-builder`, the built static dir, nor any vendor download (and `download-cdn-assets.sh` is deleted in this PR). A user following the air-gapped guidance with `Containerfile.lite`/`.scratch` gets a build with no bundled Admin UI assets. Correct the doc claims to match what those Containerfiles actually do.

mcpgateway/admin_ui/admin.js
```diff
@@ -12,6 +12,25 @@
...
+// by lazy-loader.js (TAB_FEATURE_MAP['observability'] -> 'charts') so non-admins never fetch it.
+
+import { marked } from 'marked';
+window.marked = marked;
+
+import DOMPurify from 'dompurify';
+window.DOMPurify = DOMPurify;
+
+// Import Font Awesome CSS
+import '@fortawesome/fontawesome-free/css/all.min.css';
```
madhu-mohan-jaishankar, line 32: The PR imports CodeMirror CSS, CodeMirror theme CSS, and Font Awesome CSS from admin.js, which means Vite emits CSS files separately and records them in the manifest. However, the backend still only returns a single JS filename via get_bundle_js_filename(), and the template only injects the JS bundle. There is no corresponding <link> generation for manifest[entry].css.

Suggested fix: Replace the single “bundle filename” helper with a manifest helper that returns both:
1. entry JS file
2.entry CSS files

Then render:
admin.html
```
{% for css_file in bundle_css %}
<link rel="stylesheet" href="{{ root_path }}/static/{{ css_file }}">
{% endfor %}
<script type="module" src="{{ root_path }}/static/{{ bundle_js }}"></script>
```
Also ensure the fallback path handles assets/*.css if the manifest is missing.

madhu-mohan-jaishankar (changes requested):

madhu-mohan-jaishankar (dismissed):
LGTM

marekdano (dismissed):
The PR looks good!

LGTM 🚀

ja8zyjits (approved):
LGTM

PR #5497, opened 2026-07-03
Fix admin CSRF issuance for non-email platform-admin sessions by bind…

### LINKED ISSUE #5325: [BUG]: Render Prompt button fails with CSRF_TOKEN_INVALID for platform-admin (non-email-auth) sessions

ja8zyjits:
# Bug-fix PR

## Summary

Fixes #5325. Admin UI CSRF cookies are now HMAC-bound for non-email-auth platform-admin sessions, so actions like Render Prompt no longer fail with `CSRF_TOKEN_INVALID`.

## Reproduction Steps

1. Run default config with `EMAIL_AUTH_ENABLED=false`.
2. Log in to `/admin` as platform admin.
3. Open `/admin/#prompts`.
4. Click Render Prompt.
5. Before fix: `POST /prompts/{id}` returns `403 CSRF_TOKEN_INVALID`.

## Root Cause

`GET /admin` only issued bound CSRF cookies when email auth was enabled. Default platform-admin/basic-auth sessions got random double-submit tokens, but global `CSRFMiddleware` requires tokens bound to authenticated JWT `sub` and `jti`.

## Fix Description

`/admin` now creates or refreshes a lightweight browser JWT session for admin UI calls in all auth modes and binds the CSRF cookie to that JWT payload. `_set_admin_csrf_cookie()` keeps random-token fallback for unauthenticated pages, but uses `get_csrf_service().generate_csrf_token()` when `user_id` and `session_id` are available.

Authenticated `/admin` session initialization now fails loud instead of falling back to an unusable random CSRF token. Refresh also preserves scoped session `teams` only when the verified cookie identity matches the authenticated user, and ignores stale-cookie team scope.

## Reviewability

## Verification

| Check | Command | Status |
|---|---|---|
| Focused unit tests | `.venv/bin/python -m pytest tests/unit/mcpgateway/test_admin.py tests/unit/mcpgateway/middleware/test_csrf_middleware.py tests/unit/mcpgateway/test_admin_module.py tests/unit/mcpgateway/test_a2a_jsonrpc_passthrough.py` | Pass, `1369 passed` |
| Compile touched files | `py_compile mcpgateway/admin.py tests/unit/mcpgateway/test_admin.py tests/unit/mcpgateway/middleware/test_csrf_middleware.py tests/unit/mcpgateway/test_admin_module.py tests/unit/mcpgateway/test_a2a_jsonrpc_passthrough.py` | Pass |
| Whitespace check | `git diff --check` | Pass |
| Ruff | `make ruff` | Blocked: existing `pyproject.toml:445` unknown rule selector `PLW0717` |

## Checklist

### DIFF
--- a/mcpgateway/admin.py
+++ b/mcpgateway/admin.py
@@ -4171,13 +4171,36 @@ async def admin_ui(
 
         # get_current_user_with_permissions may not include auth_provider in its dict.
         # Fall back to the current jwt_token cookie payload before refreshing it,
-        # and preserve any explicit team narrowing from the verified token.
-        existing_token_teams: list[str] | None = None
+        # but only reuse cookie metadata after confirming it matches this user.
         existing_payload: dict[str, Any] | None = None
         jwt_cookie = request.cookies.get("jwt_token")
         if isinstance(jwt_cookie, str) and jwt_cookie:
             try:
                 existing_payload = await verify_jwt_token_cached(jwt_cookie, request)
+            except Exception as provider_error:  # nosec B110 - best-effort provider preservation
+                LOGGER.warning("Could not verify existing JWT cookie for admin session refresh: %s", provider_error)
+                if settings.sso_keycloak_enabled:
+                    auth_provider = "keycloak"
+
+        # Generate a lightweight session JWT token for browser admin calls in every auth mode.
+        email_user = db.query(EmailUser).filter(EmailUser.email == admin_email).first()
+        sub_claim = str(email_user.id) if email_user else admin_email
+        existing_token_teams: list[str] | None = None
+        if isinstance(existing_payload, dict):
+            existing_user = existing_payload.get("user")
+            provider_from_token = existing_user.get("auth_provider") if isinstance(existing_user, dict) else None
+            if not provider_from_token:
+                provider_from_token = existing_payload.get("auth_provider")
+            if auth_provider == "local" and isinstance(provider_from_token, str) and provider_from_token.strip():
+                auth_provider = provider_from_token.strip()
+
+            existing_email = existing_payload.get("email")
+            if not existing_email and isinstance(existing_user, dict):
+                existing_email = existing_user.get("email")
+            existing_sub = existing_payload.get("sub")
+            cookie_matches_user = existing_email == admin_email or (isinstance(existing_sub, str) and existing_sub in {admin_email, sub_claim})
+
+            if cookie_matches_user:
                 raw_existing_teams = existing_payload.get("teams")
                 if isinstance(raw_existing_teams, list) and raw_existing_teams:
                     copied_teams: list[str] = []
@@ -4189,26 +4212,9 @@ async def admin_ui(
                     if copied_teams:
                         existing_token_teams = copied_teams
 
-                if auth_provider == "local":
-                    existing_user = existing_payload.get("user")
-                    provider_from_token = existing_user.get("auth_provider") if isinstance(existing_user, dict) else None
-                    if not provider_from_token:
-                        provider_from_token = existing_payload.get("auth_provider")
-                    if isinstance(provider_from_token, str) and provider_from_token.strip():
-                        auth_provider = provider_from_token.strip()
-            except Exception as provider_error:  # nosec B110 - best-effort provider preservation
-                LOGGER.warning("Could not resolve auth_provider or team scope from existing JWT cookie; SSO logout or scoped session refresh may not function correctly: %s", provider_error)
-                if auth_provider == "local" and settings.sso_keycloak_enabled:
-                    auth_provider = "keycloak"
-
-        # Generate a lightweight session JWT token for browser admin calls in every auth mode.
-        sub_claim = existing_payload.get("sub") if isinstance(existing_payload, dict) else None
-        if not sub_claim:
-            user_id = user.get("id") if isinstance(user, dict) else getattr(user, "id", None)
-            sub_claim = str(user_id) if user_id else admin_email
         now = datetime.now(timezone.utc)
         payload = {
-            "sub": str(sub_claim),
+            "sub": sub_claim,
             "iss": settings.jwt_issuer,
             "aud": settings.jwt_audience,
             "iat": int(now.timestamp()),
--- a/mcpgateway/admin.py
+++ b/mcpgateway/admin.py
@@ -64,7 +64,7 @@ from mcpgateway import __version__
 from mcpgateway import version as version_module
 
 # Authentication and password-related imports
-from mcpgateway.auth import get_current_user, get_user_team_roles, normalize_token_teams
+from mcpgateway.auth import get_current_user, get_user_team_roles
 
 # Re-export canonical get_user_email from auth_context for backward compatibility.
 from mcpgateway.auth_context import get_scoped_resource_access_context, get_token_teams_from_request, get_user_email
@@ -4173,13 +4173,21 @@ async def admin_ui(
         # Fall back to the current jwt_token cookie payload before refreshing it,
         # and preserve any explicit team narrowing from the verified token.
         existing_token_teams: list[str] | None = None
+        existing_payload: dict[str, Any] | None = None
         jwt_cookie = request.cookies.get("jwt_token")
         if isinstance(jwt_cookie, str) and jwt_cookie:
             try:
                 existing_payload = await verify_jwt_token_cached(jwt_cookie, request)
-                normalized_existing_teams = normalize_token_teams(existing_payload)
-                if isinstance(normalized_existing_teams, list) and normalized_existing_teams:
-                    existing_token_teams = normalized_existing_teams
+                raw_existing_teams = existing_payload.get("teams")
+                if isinstance(raw_existing_teams, list) and raw_existing_teams:
+                    copied_teams: list[str] = []
+                    for raw_team in raw_existing_teams:
+                        if isinstance(raw_team, str) and raw_team:
+                            copied_teams.append(raw_team)
+                        elif isinstance(raw_team, dict) and raw_team.get("id"):
+                            copied_teams.append(str(raw_team["id"]))
+                    if copied_teams:
+                        existing_token_teams = copied_teams
 
                 if auth_provider == "local":
                     existing_user = existing_payload.get("user")
@@ -4194,11 +4202,13 @@ async def admin_ui(
                     auth_provider = "keycloak"
 
         # Generate a lightweight session JWT token for browser admin calls in every auth mode.
-        email_user = db.query(EmailUser).filter(EmailUser.email == admin_email).first()
-        sub_claim = str(email_user.id) if email_user else admin_email
+        sub_claim = existing_payload.get("sub") if isinstance(existing_payload, dict) else None
+        if not sub_claim:
+            user_id = user.get("id") if isinstance(user, dict) else getattr(user, "id", None)
+            sub_claim = str(user_id) if user_id else admin_email
         now = datetime.now(timezone.utc)
         payload = {
-            "sub": sub_claim,
+            "sub": str(sub_claim),
             "iss": settings.jwt_issuer,
             "aud": settings.jwt_audience,
             "iat": int(now.timestamp()),
@@ -4220,7 +4230,8 @@ async def admin_ui(
         csrf_session_id = str(payload["jti"])
         LOGGER.debug(f"Set session JWT token cookie for user: {admin_email}")
     except Exception as e:
-        LOGGER.warning(f"Failed to set JWT token cookie for user {user}: {e}")
+        LOGGER.exception("Failed to initialize admin browser session for user %s", get_user_email(user))
+        raise HTTPException(status_code=500, detail="Unable to initialize admin session") from e
 
     cookie_action = ui_visibility_config.get("cookie_action")
     if cookie_action:
--- a/mcpgateway/admin.py
+++ b/mcpgateway/admin.py
@@ -153,6 +153,7 @@ from mcpgateway.services.argon2_service import Argon2PasswordService
 from mcpgateway.services.audit_trail_service import get_audit_trail_service
 from mcpgateway.services.catalog_service import catalog_service
 from mcpgateway.services.content_security import ContentSizeError, ContentTypeError, TemplateValidationError
+from mcpgateway.services.csrf_service import get_csrf_service
 from mcpgateway.services.email_auth_service import AuthenticationError, EmailAuthService, PasswordValidationError
 from mcpgateway.services.encryption_service import get_encryption_service
 from mcpgateway.services.export_service import ExportError, ExportService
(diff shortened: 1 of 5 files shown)


mcpgateway/admin.py
```diff
@@ -4137,74 +4143,74 @@ def _to_dict_and_filter(raw_list):
...
+                    if not provider_from_token:
+                        provider_from_token = existing_payload.get("auth_provider")
+                    if isinstance(provider_from_token, str) and provider_from_token.strip():
+                        auth_provider = provider_from_token.strip()
+                except Exception as provider_error:  # nosec B110 - best-effort provider preservation
+                    LOGGER.warning("Could not resolve auth_provider from existing JWT cookie; SSO logout may not function correctly: %s", provider_error)
+                    if settings.sso_keycloak_enabled:
+                        auth_provider = "keycloak"
+
+        # Generate a lightweight session JWT token for browser admin calls in every auth mode.
```
vishu-bh: This refresh mints a new `token_use="session"` JWT without carrying forward the current token's `teams` claim. For session tokens, an absent/empty `teams` claim resolves to full DB membership, while a non-empty claim narrows the session. That means a user who loaded `/admin` with a narrowed session such as `teams=["t1"]` can receive a broader browser cookie after page load. Can we preserve the existing effective team narrowing when refreshing this token, and add a regression test for that scoped-session case?

vishu-bh (changes requested):

mcpgateway/admin.py
ja8zyjits: @vishu-bh Please check now

vishu-bh (dismissed):
Thanks for addressing the comment.

LGTM 🚀

msureshkumar88 (changes requested):
Thanks for tracking down the root cause here — binding the CSRF cookie to the refreshed JWT sub/jti for non-email-auth platform-admin sessions is the right fix, and the new middleware tests (`test_admin_random_csrf_token_fails_hmac_validation` / `test_admin_bound_csrf_token_from_page_load_passes_hmac_validation`) do a nice job proving the HMAC binding actually works end-to-end.

**Blocking**
- `admin.py:4148-4224` wraps the whole JWT-refresh + CSRF-bind block in a bare `except Exception: LOGGER.warning(...)`. If `db.query(EmailUser)` or `create_jwt_token()` ever raises, `csrf_user_id`/`csrf_session_id` stay `None`, and `_set_admin_csrf_cookie()` falls back to the old `secrets.token_urlsafe(32)` token — which isn't 64 hex chars, so it can never pass `validate_csrf_token()`. That silently reintroduces the exact `CSRF_TOKEN_INVALID` bug this PR fixes, just logged at WARNING instead of surfaced. Could this fail loud (or at least emit a metric) instead of silently degrading to the vulnerable path? Also no test currently forces this branch.

**Worth considering**
- `admin.py:4180` uses `normalize_token_teams()` to decide whether to carry forward team narrowing on the *existing* `jwt_token` cookie — but that cookie is always a session-type token, and this repo's own convention (and the comment at `auth.py:439-444`) says session tokens should go through `resolve_session_teams()`, since the two helpers deliberately disagree on what `teams: []` means. Not exploitable today (nothing currently mints a session cookie with an explicit empty teams list), but it's a latent trap for whoever adds team-narrowing to admin sessions next.
- The DB lookup + JWT mint that used to be gated behind `email_auth_enabled` now runs on every `/admin` page load in all auth modes — worth confirming that's an acceptable cost for basic-auth/SSO deployments that didn't pay it before.

Nothing here is a big lift — happy to take another look once addressed.

ja8zyjits:
Hi @msureshkumar88 
Thanks for the feedback

Implemented follow-up:

### Changed [admin.py]:
* Removed normalize_token_teams() from session refresh path.
* Copies only verified raw non-empty teams claim into refreshed JWT.
* Ignores teams=[].
* Avoids EmailUser DB lookup for /admin session refresh; uses existing JWT sub, then user.id, then email.
* Fails loud with HTTPException(500) if admin browser session init fails, so no random CSRF fallback.

### Added tests in [test_admin.py]:
* Empty teams=[] not copied.
* JWT mint failure raises 500 and does not call random CSRF fallback.
* Existing tests now assert no EmailUser lookup.

msureshkumar88 (changes requested):
Re-reviewed this branch's diff (`admin.py`, the two test files; `.secrets.baseline` is just timestamp/line-number churn, ignoring that per repo convention). CI is green (30/30) and the earlier round of requested changes all landed — no more `EmailUser` DB lookup, team narrowing on refresh is preserved (and correctly skips empty `teams: []`), and the silent random-CSRF fallback on JWT-mint failure is gone in favor of a loud 500, all with dedicated regression tests. Nice work tightening this up.

One thing I'd like resolved before merge:

**Blocking — `sub` claim reuse doesn't validate it matches the currently authenticated identity**

`admin.py`:
```python
sub_claim = existing_payload.get("sub") if isinstance(existing_payload, dict) else None
```
`existing_payload` comes from decoding whatever `jwt_token` cookie is already on the request, independent of the `user` this handler just authenticated. In the common cookie-driven path these always agree (both come from the same cookie), so this is safe today. But `rbac.py` prioritizes the `Authorization` header over the cookie (`if credentials and credentials.credentials: token = credentials.credentials`), so a request carrying a valid Bearer token for one identity *plus* a stale `jwt_token` cookie for a different identity (with `Accept: text/html` to be treated as a browser request) authenticates as the Bearer identity but mints the new session JWT / CSRF binding under the stale cookie's `sub`. Rendering and permission checks stay correct since they re-derive from the authenticated `user`, but the `sub` embedded in the new session and CSRF token — and thus the identity future requests/audit trail will attribute to this session — is wrong.

Could we add a check that `existing_payload`'s resolved identity matches `admin_email`/`user` before reusing its `sub`, and fall back to a fresh `sub` when it doesn't? A regression test for the mismatched-identity case would also close the gap — right now all four new tests only cover the case where the cookie and the authenticated user agree.

A couple of non-blocking notes:
- Test coverage: the new tests validate the CSRF-middleware side and the `admin_ui` side independently (the `admin_ui` tests mock `get_csrf_service` to return a canned token string), but nothing wires the real `CSRFService` + `CSRFMiddleware` + `admin_ui` together end-to-end to prove a token minted by `admin_ui` is actually accepted by the middleware. Worth an assembly-level test if there's an easy way to do it here.
- The fail-loud behavior on JWT-mint failure (500 on `/admin` load) is the right call to avoid the insecure fallback, just flagging that it does mean any transient signing hiccup now takes down the whole admin page load rather than degrading gracefully — sounds like an accepted trade-off given the tests explicitly pin this behavior.

ja8zyjits:
Follow up issue that expands the scope https://github.com/IBM/mcp-context-forge/issues/5508

ja8zyjits:
Hi @msureshkumar88 
Thanks for the feedback. I have reduced the scope of the implementation and moved certain things to the follow ups


## Scope Summary

Original issue: non-email-auth platform-admin `/admin` page load issued random CSRF, but middleware required HMAC-bound CSRF tied to JWT `sub`/`jti`.

## In Scope And Implemented

- `_set_admin_csrf_cookie()` can issue HMAC-bound token when `user_id` + `session_id` exist.
- `/admin` creates/refreshed browser session JWT for Admin UI calls across auth modes.
- CSRF cookie binds to refreshed JWT `sub` + `jti`.
- Random CSRF fallback remains for unauthenticated admin pages.
- Authenticated `/admin` fails loud if session/CSRF init fails, instead of silently issuing unusable random CSRF.
- Existing scoped session `teams=["t1"]` is preserved only when existing cookie identity matches authenticated user.
- Stale cookie identity is ignored so Bearer user does not inherit old cookie `sub`/`teams`.
- Tests cover random token reject, bound token accept, non-email admin issuance, fail-loud path, scoped teams, stale cookie, and empty teams.

## Slight Scope Expansion But Justified

Preserving scoped `teams` and stale-cookie identity checks came from review, but protects session correctness created by the new `/admin` JWT refresh path.

## Deferred Follow-Ups #5508 

- DB-cost optimization: avoid `EmailUser` lookup on every `/admin` load when email subject is enough.
- Add assembly-level test wiring real `admin_ui()` CSRF issuance directly into `CSRFMiddleware`.
- Broader cleanup of admin session refresh subject-selection semantics, if maintainers want to reduce DB work or standardize identity matching further.

msureshkumar88:
Thanks for the detailed writeup and for filing #5508 for the deferred items.

One thing I want to flag before re-reviewing: the PR is still on `9701737` (no new commit since your last comment), and that commit doesn't yet contain the "stale cookie identity is ignored so Bearer user does not inherit old cookie `sub`/`teams`" change described above — `sub_claim = existing_payload.get("sub")` at `admin.py:4205` is still reused unconditionally, and there's no comparison anywhere against `admin_email`/`user` (checked all usages of `admin_email` in the file). Same for the `teams` narrowing block just above it.

Happy to take another look as soon as that lands — just didn't want to approve based on a description that isn't in the diff yet.

ja8zyjits:
Hi @msureshkumar88 
Sorry, the code is updated now

msureshkumar88 (approved):
Ran a full end-to-end verification against a real running gateway (no mocks) on commit `af5b8ec8`, reproducing the exact bug report scenario, and it passes -- approving.

## Config used
```
EMAIL_AUTH_ENABLED=false
REQUIRE_USER_IN_DB=false
AUTH_REQUIRED=true
CSRF_ENABLED=true
CSRF_COOKIE_NAME=mcpgateway_csrf_token
CSRF_TOKEN_NAME=X-CSRF-Token
CSRF_CHECK_REFERER=false
MCPGATEWAY_UI_ENABLED=true
DATABASE_URL=sqlite:///./e2e_pr5497_test.db (fresh)
```

## What the script does
1. Starts a real `uvicorn mcpgateway.main:app` process.
2. Provisions a non-email-auth platform-admin session the way such a deployment actually gets one: mints a JWT via the documented `python -m mcpgateway.utils.create_jwt_token --username admin@example.com --admin` CLI, set as the `jwt_token` cookie.
3. Seeds the one-time `platform_admin` RBAC role grant that a fresh non-email-auth deployment needs out-of-band (`bootstrap_default_roles()` is itself gated on `EMAIL_AUTH_ENABLED` and skipped in this mode -- a separate, pre-existing gap unrelated to this PR, worked around here only so the test can reach the code under test).
4. `GET /admin/` (the fixed code path) and captures the real `Set-Cookie` values the server issues.
5. Creates a real prompt, then `POST /prompts/{id}` -- the exact "Render Prompt" request from the bug report -- using the captured cookie + `X-CSRF-Token` header.
6. Negative control: repeats the POST with a random unbound token shaped like the pre-fix output, to confirm the middleware genuinely enforces HMAC binding rather than the test being permissive.

## Output
```
== 1. Start mcp-context-forge gateway server ==
PASS: gateway server is up on :44597

== 2. Mint a non-email-auth platform-admin JWT (the documented CLI helper) ==
PASS: minted JWT for admin@example.com (sub=email, auth_provider=cli)

== 3. Seed the one-time platform_admin RBAC role grant ==
PASS: seeded platform_admin role + assignment for admin@example.com

== 4. GET /admin with the non-email-auth session cookie (the fixed code path) ==
PASS: GET /admin/ returned 200 for non-email-auth session
PASS: admin_ui issued a 64+ char (HMAC-shaped) CSRF cookie: 6f376b63605ee4de...

== 5. Create a prompt to render ==
PASS: created prompt id=a0a9bf95b46b46c8b64ae04021abb35c

== 6. POST /prompts/{id} (Render Prompt) with the bound cookie/header from step 4 ==
PASS: Render Prompt POST succeeded (200) with the bound CSRF cookie -- bug #5325 is fixed
{"messages":[{"role":"user","content":{"type":"text","text":"Hello World, welcome to ContextForge!"}}],"description":"PR 5497 e2e prompt"}

== 7. Negative control: pre-fix-shaped random (unbound) token must still be rejected ==
PASS: random/unbound token correctly rejected with 403 CSRF_TOKEN_INVALID (proves step 6 passed due to real HMAC binding)

== Summary ==
ALL CHECKS PASSED
```

Also re-verified the blocking finding from my earlier review is resolved in this commit: the `sub` claim is now always freshly derived rather than reused from a stale cookie, and team narrowing is gated on `cookie_matches_user` -- confirmed by the new `test_admin_ui_refresh_ignores_stale_cookie_identity` test and by reading the current diff.

Nice work getting this across the line. Approving.

PR #5512, opened 2026-07-06
feat(ui-rewrite): Group resources

gcgoncalves:
Screenshot 2026-07-06 at 14 36 28

---

Screenshot 2026-07-06 at 14 36 56

---

Screenshot 2026-07-06 at 14 36 39


client/src/components/resources/ResourceDetailsPanel.tsx
```diff
@@ -19,103 +52,296 @@ export function ResourceDetailsPanel({
...
+      const target = e.target as Element | null;
+      if (!target || panel.contains(target) || target === closeButton) return;
+      if (target.closest("[data-radix-popper-content-wrapper]")) return;
+
+      if (document.activeElement !== closeButton) {
+        isRedirecting = true;
+        closeButton.focus();
+        isRedirecting = false;
+      }
+    };
```
marekdano: `ToolDetailsPanel` has no focus trap; `ResourceDetailsPanel` adds a capture-phase `focusin` handler:

  ```ts
  const panel = document.querySelector('[role="region"][aria-hidden="false"]');
  // ...
  if (target.closest("[data-radix-popper-content-wrapper]")) return; // dropdowns excluded
  if (document.activeElement !== closeButton) closeButton.focus();
  ```

Two problems:

  - `ConfirmDialog` is a Radix **Dialog** (not Popper-based), so its portal content is **not** matched by the `[data-radix-popper-content-wrapper]` escape hatch. When the user clicks Delete while the panel is open, focus enters the confirm dialog and this handler yanks it back to the panel's close button, competing with the dialog's own focus scope. Tests don't catch it because `userEvent.click()` doesn't depend on focus staying put, but it's a live keyboard/AX regression.
  - `document.querySelector('[role="region"][aria-hidden="false"]')` grabs the *first* matching region on the page, not necessarily this panel — fragile, since `role="region"` is common.

  **Recommendation:** remove the focus-trap effect to match `ToolDetailsPanel` (Tools proves the identical delete flow works without it). If it's truly needed, also exclude Radix Dialog content (`[role="dialog"]`) and scope the panel lookup to a ref instead of a global selector.

marekdano (commented):
List of suggestion for improvement:

## Accessibility

  Some of these are shared with the Tools page, but they ship in the new component so they're worth addressing — ideally as a shared fix rather than Resources-only.

  - **Table rows are mouse-only.** `<TableRow onClick={...} className="cursor-pointer">` has no `tabIndex`/`role`/`onKeyDown`, so keyboard users can't select a resource to populate the detail column.
  - **Closed panel stays tabbable.** On close, only `isDetailsPanelOpen` is set false; `selectedGroup` stays set, so `activeGroup` keeps the panel mounted with `aria-hidden={true}` while its buttons remain in the tab order — keyboard users can tab into an off-screen hidden panel (and it triggers the "Blocked aria-hidden … descendant retained focus" warning). Use `inert` when `!open`.
  - **Color-only status.** The group card status dot (`bg-tool-status-active` vs `bg-tool-status-inactive`) has no text/`aria-label`, so active/inactive is invisible to screen readers.
  - **ESC handler diverges from Tools** — Resources adds `preventDefault()` + `stopPropagation()`, which with the confirm dialog open could close both dialog and panel. Align with Tools' plain `onClose()`.

Credit where due: copy buttons and the close button have proper `intl` `aria-label`s, the panel uses `role="region"` + `aria-labelledby`, and ESC-to-close works.

  ## Code quality

  - **Double React import** in `Resources.tsx` — `import { useEffect } from "react";` is added as a separate line right under the existing React import; merge them.
  - **`formatDateTime` duplication** — now shared in `utils/format.ts`, but `ToolDetailsPanel` still has its own inline copy; dedupe Tools onto the shared util here or in a follow-up.
  - **Unused i18n key** `resources.table.view` in all three locales.
  - **Lost named prop interface** — `ResourceDetailsPanelProps` replaced with an inline anonymous type (matches Tools, but a named interface reads better).
  - **`group.isActive ||= resource.enabled`** — assigns `undefined` to a `boolean` field if `enabled` is ever `undefined`; `group.isActive = group.isActive || !!resource.enabled` is cleaner.
  - **`buildGroups` slug collisions** — two distinct `gatewayId`s resolving to the same label would merge into one group; edge case, and consistent with how Tools keys groups, so just noting it.
  - **`copyName` label semantics** — `ResourcesTable` reuses `resources.table.copyName` ("Copy {name}") but passes a URI, so the label reads "Copy resource://…". A dedicated `copyUri` key would read better.

marekdano (approved):
The PR looks good now!

LGTM 🚀

PR #5516, opened 2026-07-06
test: assert RBAC allow-path tools lists are non-empty

lucarlig:
# Bug-fix PR

## Summary

The per-server allow-path checks (`test_public_token_accesses_public_server`, `test_team_member_accesses_team_server`) only asserted that `tools/list` succeeded. A deployment that cannot serve the server's backend can answer with an empty tool list, and the tests passed while proving nothing. They now assert the visibility servers actually advertise their associated tools.

On split deployments this pairs with #5519: unsupported-transport servers are no longer half-served by the dataplane, so the per-server route delivers real tools (directly or via control-plane fallback routing) and these assertions hold.

## Reviewability

## Verification

| Check | Command | Status |
|-------|---------|--------|
| Ruff lint + format | `uv run ruff check / format --check` on the test file | Pass |

## Checklist

- [x] Code formatted


lucarlig:
Withdrawing — too small to carry as its own PR.

PR #5517, opened 2026-07-06
fix(dataplane_publisher): per-worker lock id and safer key TTL

lucarlig:
# Bug-fix PR

## Summary

Two robustness fixes for the dataplane publisher, observed on a live split stack:

- `WORKER_ID` is evaluated at module import — under gunicorn with preload that happens in the master before fork, so every worker publishes as the same `hostname:1` id and the lock's compare-and-swap release cannot distinguish owners. It is now computed per service instance (each worker constructs its own service post-fork), same pattern as `session_affinity.py`.
- UserConfig keys were written with TTL = interval + 10s. A single publish cycle firing late (observed 120–180s gaps between publishes while the event loop was busy under test load) expires every key at once, dropping dataplane config for all subjects until the next snapshot. The TTL is now 2 × interval + 10 so one missed cycle no longer causes an outage.

Default publish cadence and payloads are unchanged.

## Reviewability

## Verification

| Check | Command | Status |
|-------|---------|--------|
| Ruff lint + format | `uv run ruff check / format --check` on both files | Pass |
| Publisher unit tests | `uv run pytest tests/unit/mcpgateway/services/test_dataplane_publisher.py` | 21 passed |

## Checklist

- [x] Code formatted

### DIFF
--- a/mcpgateway/services/dataplane_publisher.py
+++ b/mcpgateway/services/dataplane_publisher.py
@@ -43,11 +43,13 @@ logger = logging.getLogger(__name__)
 
 USER_CONFIG_KEY = "UserConfig"
 REDIS_PUBLISHER_TIME = 60  # Publish interval in seconds
-# Keys are not deleted explicitly; stale configs expire via Redis TTL.
-PUBLISHER_TTL = 70
+# Keys are not deleted explicitly; stale configs expire via Redis TTL. The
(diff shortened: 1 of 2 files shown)


cafalchio (dismissed):
LGTM

cafalchio (approved):
LGTM

ja8zyjits (approved):
LGTM

PR #5521, opened 2026-07-06
chore: split test-live-gateway into playwright and non-playwright passes

lucarlig:
# Chore PR

## Summary

`make test-live-gateway` ran the entire `tests/live_gateway/` tree under `-p playwright`. pytest-playwright's runtest hook collides with pytest-asyncio (`RuntimeError: Runner.run() cannot be called from a running event loop`), so every async test in the tree — including the protocol-compliance reference rows that pass standalone — failed regardless of gateway behavior, drowning real regressions in noise (observed: 64 failed + 78 errors in ~17s, ~103 of them this collision).

## Fix Description

Split the target into two passes:

1. `-p no:playwright` over the tree, excluding the two playwright-dependent suites — all asyncio suites now run the same way their individual make targets run them.
2. `-p playwright` over `tests/live_gateway/sso` and `tests/live_gateway/mcp/test_mcp_rbac_transport.py`, the only files importing playwright.

No test content changes; same overall pass/fail contract (first failing pass fails the target).

## Reviewability

## Verification

| Check | Command | Status |
|-------|---------|--------|
| Recipe expansion | `make -n test-live-gateway` | two pytest invocations as described |
| Playwright dependency map | `grep -rl playwright tests/live_gateway/` | only the two excluded files |

## Checklist

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-07-06T12:57:22Z",
+  "generated_at": "2026-07-06T14:54:10Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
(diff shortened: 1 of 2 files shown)


lucarlig:
Withdrawing: the split only matters for the integration harness, which can run its own two-pass pytest invocation directly — no need to change the upstream Makefile.

PR #5523, opened 2026-07-06
test: align admin private-server listing test with owner matching

lucarlig:
# Bug-fix PR

## Summary

`test_admin_sees_public_and_team_via_http` still asserts the pre-#4694 scoping semantics, where the admin-bypass path collapsed to anonymous visibility and hid even the caller's own private servers. Commit 8c186c5e0 (issue #4694) deliberately changed `get_scoped_resource_access_context` to keep the requester email on that path so the service layer can owner-match — the listing now returns public rows, team rows, and the caller's **own** private rows. The fixture's private server is created by the same admin, so it correctly appears and the stale assertion fails on every live run.

## Fix Description

Flip the private-server assertion to expect presence via owner matching and update the docstring to cite the current contract. Another user's private rows remain excluded — the existing outsider/team-member tests in the same class cover those deny paths and keep passing.

## Reviewability

## Verification

| Check | Command | Status |
|-------|---------|--------|
| Ruff lint + format | `uv run ruff check / format --check` on the test file | Pass |
| Live stack | `pytest -p playwright …::TestServerVisibilityViaAPI -v` | 5 passed |

## Checklist

- [x] Code formatted

### DIFF
--- a/tests/live_gateway/mcp/test_mcp_rbac_transport.py
+++ b/tests/live_gateway/mcp/test_mcp_rbac_transport.py
@@ -488,25 +488,25 @@ class TestServerVisibilityViaAPI:
     """Verify server visibility via REST API before MCP protocol tests."""
 
     def test_admin_sees_public_and_team_via_http(self, admin_api: APIRequestContext, visibility_servers: dict) -> None:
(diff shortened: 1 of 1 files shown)


cafalchio (approved):
LGTM

Lang-Akshay (approved):
LGTM

PR #5543, opened 2026-07-07
fix(api): Check for existing team membership before accepting invitation

### LINKED ISSUE #5524: [BUG]: Accepting a team invitation for an already-active member throws unhandled 500
### 🐞 Bug Summary
Accepting a team invitation for a user who is already an active member of that team raises an unhandled `sqlalchemy.exc.IntegrityError` and returns HTTP 500, instead of a graceful 4xx (e.g. "already a member") or an idempotent success.

---

### 🧩 Affected Component
- [x] `mcpgateway` - API

---

### 🔁 Steps to Reproduce

1. Invite a user to a team (`POST` team invitation endpoint) and have them accept it (becomes an active `email_team_members` row).
2. Invite the same user to the same team again (a fresh invitation is created).
3. Accept the second invitation.
4. Server returns 500 `{"detail":"Failed to accept invitation"}`.

Found via the Playwright E2E suite (`tests/playwright/test_rbac_permissions.py`) run against a dev instance with pre-existing team membership data; discovered by inspecting server logs when several RBAC tests errored during setup.

---

### 🤔 Expected Behavior
Accepting an invitation when the invitee is already an active team member should return a handled 4xx response (e.g. 409 "already a member") or succeed idempotently — not crash with an unhandled DB integrity error.

---

### 📓 Logs / Error Output
```
mcpgateway.services.team_invitation_service - ERROR - Failed to accept invitation: (sqlite3.IntegrityError) UNIQUE constraint failed: email_team_members.team_id, email_team_members.user_email
[SQL: INSERT INTO email_team_members (id, team_id, user_email, role, joined_at, invited_by, is_active, grant_source) VALUES (?, ?, ?, ?, ?, ?, ?, ?)]
```

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Version or commit | `Release/v1.0.5 @ d71dd4791` |
| Runtime | Python 3.11, uvicorn (dev) |
| Platform / OS | Linux (WSL2) |
| Container | none |

---

Likely fix location: `mcpgateway/services/team_invitation_service.py`, invitation-accept path — should catch the `IntegrityError` on duplicate `(team_id, user_email)` and return a clean 4xx instead of letting it bubble up as a 500.

gcgoncalves:
# 🐛 Bug-fix PR

---

## 📌 Summary

Closes #5524 

## 🔁 Reproduction Steps

1. Invite a user to a team (`POST` team invitation endpoint) and have them accept it (becomes an active `email_team_members` row).
2. Invite the same user to the same team again (a fresh invitation is created).
3. Accept the second invitation.
4. Server returns 500 `{"detail":"Failed to accept invitation"}`.

## 🐞 Root Cause

`accept_invitation()` never checks for a pre-existing `inactive` row before inserting, always creating a new membership. This violates the membership uniqueness (`team_id`/`user_email`  pair) whenever one exists (most commonly after a prior removal from the team).

## 💡 Fix Description

Check for an existing membership and, if it exists, update it. Otherwise, create a new one as usual.

## 📏 Reviewability

## 🧪 Verification

| Lint suite                            | `make lint`          |        |
| Unit tests                            | `make test`          |        |
| Coverage ≥ 80 %                       | `make coverage`      |        |

### DIFF
--- a/mcpgateway/services/team_invitation_service.py
+++ b/mcpgateway/services/team_invitation_service.py
@@ -293,7 +293,8 @@ class TeamInvitationService:
 
         Raises:
             ValueError: If invitation is invalid or expired, or if the user is already a member
-                (including the race where a concurrent accept commits first)
+                (including races where a concurrent accept wins the insert, or wins the reactivation
+                of a stale row via the compare-and-swap UPDATE on is_active)
             Exception: If acceptance fails
 
         Examples:
@@ -359,11 +360,35 @@ class TeamInvitationService:
             # Reuse any prior (inactive) membership row for this (team_id, user_email) pair instead of
             # inserting a duplicate: uq_team_member is unique regardless of is_active, so a stale row left
             # behind by a previous removal from the team would otherwise collide on insert.
-            # Row-lock the reused row so two concurrent accepts can't both UPDATE it and commit silently.
+            # with_for_update() row-locks the reused row on backends that support it (e.g. Postgres), but
+            # is a silent no-op on SQLite -- this project's default DB -- so it alone does not close the
+            # race between two concurrent accepts both reactivating the same stale row. The conditional
+            # UPDATE below (matched against the is_active=False state we just observed) is what actually
+            # closes the race on every backend: only one of two racing callers can flip is_active from
+            # False to True, the other gets rowcount 0 and is turned into the same "already a member"
+            # error used for the commit-time IntegrityError race on the insert path below.
             membership = self.db.query(EmailTeamMember).filter(EmailTeamMember.team_id == invitation.team_id, EmailTeamMember.user_email == invitation.email).with_for_update().first()
 
             reactivated = membership is not None
             if membership:
+                if membership.is_active:
+                    # Closed by with_for_update() on Postgres: another transaction reactivated this row
+                    # between our earlier "already a member" check and acquiring the lock here.
+                    self.db.rollback()
+                    raise ValueError("User is already a member of this team")
+
+                updated_rows = (
+                    self.db.query(EmailTeamMember)
+                    .filter(EmailTeamMember.id == membership.id, EmailTeamMember.is_active.is_(False))
+                    .update({"role": invitation.role, "joined_at": utc_now(), "invited_by": invitation.invited_by, "is_active": True}, synchronize_session=False)
+                )
+                if updated_rows == 0:
+                    # Lost the race to another concurrent accept (e.g. on SQLite, where with_for_update()
+                    # above does not actually lock): the row was reactivated between our read and this UPDATE.
+                    self.db.rollback()
+                    raise ValueError("User is already a member of this team")
+
+                # Keep the in-memory object in sync with the row the bulk UPDATE above just changed.
                 membership.role = invitation.role
                 membership.joined_at = utc_now()
                 membership.invited_by = invitation.invited_by
@@ -379,7 +404,7 @@ class TeamInvitationService:
                 self.db.commit()
             except IntegrityError as integrity_error:
                 self.db.rollback()
-                logger.warning("Concurrent accept detected for %s on team %s: %s", invitation.email, invitation.team_id, integrity_error)
+                logger.warning("Concurrent accept detected for %s on team %s: %s", SecurityValidator.sanitize_log_message(invitation.email), invitation.team_id, integrity_error)
                 raise ValueError("User is already a member of this team") from integrity_error
 
             # Write the same audit trail entry that TeamManagementService.add_member_to_team writes,
--- a/mcpgateway/services/team_invitation_service.py
+++ b/mcpgateway/services/team_invitation_service.py
@@ -23,6 +23,7 @@ import secrets
 from typing import Any, List, Optional
 
 # Third-Party
+from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm import Session
 
 # First-Party
@@ -354,15 +355,29 @@ class TeamInvitationService:
             # Check team member limit (explicit per-team value or global default)
             check_team_member_capacity(self.db, team)
 
-            # Create team membership
-            membership = EmailTeamMember(team_id=invitation.team_id, user_email=invitation.email, role=invitation.role, joined_at=utc_now(), invited_by=invitation.invited_by, is_active=True)
+            # Reuse any prior (inactive) membership row for this (team_id, user_email) pair instead of
+            # inserting a duplicate: uq_team_member is unique regardless of is_active, so a stale row left
+            # behind by a previous removal from the team would otherwise collide on insert.
+            membership = self.db.query(EmailTeamMember).filter(EmailTeamMember.team_id == invitation.team_id, EmailTeamMember.user_email == invitation.email).first()
 
-            self.db.add(membership)
+            if membership:
+                membership.role = invitation.role
+                membership.joined_at = utc_now()
+                membership.invited_by = invitation.invited_by
+                membership.is_active = True
+            else:
+                membership = EmailTeamMember(team_id=invitation.team_id, user_email=invitation.email, role=invitation.role, joined_at=utc_now(), invited_by=invitation.invited_by, is_active=True)
+                self.db.add(membership)
 
             # Deactivate the invitation
             invitation.is_active = False
 
-            self.db.commit()
+            try:
+                self.db.commit()
+            except IntegrityError as integrity_error:
+                self.db.rollback()
+                logger.warning("Concurrent accept detected for %s on team %s: %s", invitation.email, invitation.team_id, integrity_error)
+                raise ValueError("User is already a member of this team") from integrity_error
 
             # Invalidate auth cache for user's team membership
             try:
--- a/mcpgateway/services/team_invitation_service.py
+++ b/mcpgateway/services/team_invitation_service.py
@@ -32,7 +32,7 @@ from mcpgateway.common.validators import SecurityValidator
 from mcpgateway.config import settings
 from mcpgateway.db import EmailTeam, EmailTeamInvitation, EmailTeamMember, EmailUser, utc_now
 from mcpgateway.services.logging_service import LoggingService
-from mcpgateway.services.team_management_service import check_team_member_capacity, get_user_team_count
+from mcpgateway.services.team_management_service import check_team_member_capacity, get_user_team_count, TeamManagementService
 
 # Initialize logging
 logging_service = LoggingService()
@@ -292,7 +292,8 @@ class TeamInvitationService:
             EmailTeamMember: The created team membership record
 
         Raises:
-            ValueError: If invitation is invalid or expired
+            ValueError: If invitation is invalid or expired, or if the user is already a member
+                (including the race where a concurrent accept commits first)
             Exception: If acceptance fails
 
         Examples:
@@ -358,8 +359,10 @@ class TeamInvitationService:
             # Reuse any prior (inactive) membership row for this (team_id, user_email) pair instead of
             # inserting a duplicate: uq_team_member is unique regardless of is_active, so a stale row left
             # behind by a previous removal from the team would otherwise collide on insert.
-            membership = self.db.query(EmailTeamMember).filter(EmailTeamMember.team_id == invitation.team_id, EmailTeamMember.user_email == invitation.email).first()
+            # Row-lock the reused row so two concurrent accepts can't both UPDATE it and commit silently.
+            membership = self.db.query(EmailTeamMember).filter(EmailTeamMember.team_id == invitation.team_id, EmailTeamMember.user_email == invitation.email).with_for_update().first()
 
+            reactivated = membership is not None
             if membership:
                 membership.role = invitation.role
                 membership.joined_at = utc_now()
@@ -378,6 +381,12 @@ class TeamInvitationService:
                 self.db.rollback()
                 logger.warning("Concurrent accept detected for %s on team %s: %s", invitation.email, invitation.team_id, integrity_error)
                 raise ValueError("User is already a member of this team") from integrity_error
+
+            # Write the same audit trail entry that TeamManagementService.add_member_to_team writes,
+            # so a membership reactivated via invitation-accept isn't invisible in EmailTeamMemberHistory.
+            TeamManagementService(self.db).log_team_member_action(
+                membership.id, invitation.team_id, invitation.email, invitation.role, "reactivated" if reactivated else "added", invitation.invited_by
+            )
 
             # Invalidate auth cache for user's team membership
             try:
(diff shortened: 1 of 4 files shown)


msureshkumar88 (changes requested):
Thanks for the clear repro and root-cause writeup — the fix matches the described issue exactly (check-then-reactivate-or-insert), and the two new tests do a nice job covering both the reuse path and the new `IntegrityError` translation.

**Blocking**

1. **The concurrency fix only covers the INSERT race, not the UPDATE race** (`team_invitation_service.py:361-379`). The reused-row lookup does a plain `SELECT` then `UPDATE` with no row lock (`with_for_update()`) and no version check. The new `except IntegrityError` only fires on `INSERT` collisions against `uq_team_member`, so two concurrent accept calls that both land on the "reuse a stale row" branch (e.g. a double-submitted request) will both `UPDATE` the same row and both commit successfully — no error surfaced, silently bypassing the exact protection this PR is meant to add. Suggest locking the row before mutating it, e.g.:
   ```python
   membership = self.db.query(EmailTeamMember).filter(
       EmailTeamMember.team_id == invitation.team_id,
       EmailTeamMember.user_email == invitation.email,
   ).with_for_update().first()
   ```
   (there's existing precedent for this in `gateway_service.py:3792`).

2. **Duplicated logic vs. `team_management_service.py::add_member`** (`:862-889`), which already implements the same "reactivate existing row or insert" pattern — and additionally calls `self._log_team_member_action(..., "reactivated", ...)` to write an `EmailTeamMemberHistory` audit row. The new reactivation branch here doesn't write any equivalent audit entry, so a membership reactivated via invitation-accept leaves no history trail while the same reactivation via direct add does. Worth either extracting a shared helper or wiring in the same audit call.

**Minor / non-blocking**

- The docstring's `Raises:` section (`:294-296`) wasn't updated to mention the new "already a member" case arising from the `IntegrityError` path.
- No integration/e2e test exists for the accept-invitation endpoint (pre-existing gap, not introduced by this PR, but flagging since this is exactly the code path in question).

No unrelated changes, no breaking change, issue reference present (#5524) — scope of the PR itself looks good.

msureshkumar88 (changes requested):
Thanks for addressing both blocking items from the last round — the row lock on the reuse path and the audit-trail call for reactivation both look correct, and the docstring update covers the new failure mode. The new tests do a good job distinguishing the reactivate-vs-insert branches and asserting the audit action type for each.

One thing came up on a closer pass at the row lock itself, though, that I think is worth resolving before merge since it affects the core guarantee this PR is adding.

**Blocking**

1. **`with_for_update()` is a no-op on SQLite, so the race this PR targets is still open on the project's default DB** (`team_invitation_service.py:363`). I compiled the query directly against a SQLite engine and confirmed it produces a plain `SELECT` with no `FOR UPDATE` clause — SQLAlchemy's SQLite dialect silently drops it, no error or warning. Since `DATABASE_URL=sqlite:///./mcp.db` is this repo's documented default, the concurrent double-submit race is only actually closed on Postgres. Could either note this is Postgres-only protection (and accept it, if that's an acceptable scope for now), or add a guard that works cross-dialect — e.g. rely on the existing `IntegrityError` catch plus a re-check, or an optimistic version column.

2. **Even on Postgres, the lock prevents a lost update but not duplicate audit entries.** Two concurrent accepts that both land on the reactivation branch: the second caller blocks on the lock, then after the first commits, re-reads the now-active row and reruns the same reactivation logic. It's an UPDATE, not an INSERT, so no `IntegrityError` fires — both calls "succeed," and `log_team_member_action(..., "reactivated", ...)` gets called twice, producing a duplicate `EmailTeamMemberHistory` row plus redundant cache invalidation and RBAC role-assignment calls. Final membership state stays correct, so this isn't a security bypass, but it does undercut the audit-trail guarantee the PR is adding. Might be worth a re-check of `is_active` state after acquiring the lock, or simply flagging this as a known limitation in a comment.

**Minor / non-blocking**

- `TeamManagementService.log_team_member_action` (`team_management_service.py:343-354`) is always mocked out in the new tests — it has no direct unit-test or doctest coverage of its own (its sibling `_log_team_member_action` has a doctest, this one doesn't). Also no test asserts `with_for_update()` is called with the correct filter args (team_id, user_email), only that it's invoked.
- The new `logger.warning` at `:382` interpolates `invitation.email` unsanitized, consistent with a few other spots in this file — not new, but since this PR is touching the race-handling path anyway, might be a good spot to start using `SecurityValidator.sanitize_log_message()` there.
- No integration/e2e test for the accept-invitation endpoint (pre-existing gap, not introduced by this PR, flagging again since it's exactly the path in question).

No unrelated changes, no breaking change, issue reference present (#5524).

gcgoncalves:
Thanks for the detailed follow-up — you're right that `with_for_update()` alone wasn't enough here.

**Blocking items, both addressed:**

Replaced the blind reactivation (lock-then-blindly-UPDATE) with a compare-and-swap: after the `with_for_update()` read, the reactivation now goes through `UPDATE ... WHERE id = :id AND is_active = False`, and the rowcount is checked before proceeding. This closes the race on every backend, not just Postgres:

- On Postgres, `with_for_update()` still serializes the two transactions, so the second caller now additionally re-observes `is_active=True` right after acquiring the lock and is rejected immediately (no wasted UPDATE attempt).
- On SQLite, where `with_for_update()` is a documented no-op, both callers can race past the initial read, but only one UPDATE can match `is_active=False` — the loser gets rowcount 0 and is turned into the same `ValueError("User is already a member of this team")` used for the existing insert-side `IntegrityError` path.

This also directly fixes the duplicate-audit-entry issue: the losing caller now errors out before ever reaching `log_team_member_action(..., "reactivated", ...)`, so there's no double `EmailTeamMemberHistory` row, no redundant cache invalidation, and no redundant RBAC role-assignment call on the losing side.

Added a regression test (`test_accept_invitation_reactivation_race_loses_is_rejected`) that simulates the SQLite no-op-lock scenario directly: both callers observe the same stale row, the conditional UPDATE reports 0 rows affected, and the test asserts the second caller is rejected with no commit and no audit-log call. Also strengthened the existing reactivation test to assert the compare-and-swap UPDATE is actually invoked.

**Minor items:**

- Added a doctest to `TeamManagementService.log_team_member_action` (the public wrapper), matching the sibling `_log_team_member_action`.
- Sanitized `invitation.email` with `SecurityValidator.sanitize_log_message()` in the concurrent-accept warning log line.
- Left the `with_for_update()` filter-args assertion and the integration/e2e gap as-is for now — happy to add either if you'd like them in this PR rather than as follow-ups.

All existing + new unit tests pass, ruff/bandit clean on the touched files.

msureshkumar88:
## Round-2 blocking items — status: both resolved

| # | Round-2 ask | Resolution |
|---|---|---|
| 1 | `with_for_update()` no-ops on SQLite (project default DB) — race stays open | `8fb681488`: added compare-and-swap `UPDATE ... WHERE id=? AND is_active=False`. Losing caller gets `rowcount=0` → rejected with the same "already a member" `ValueError`, on every backend. Verified this compiles a plain SELECT with no `FOR UPDATE` on the SQLite dialect, and the CAS branch is what actually closes the race there. |
| 2 | Even on Postgres, lock prevents lost update but not duplicate audit entries | Same commit: a losing racer now never reaches `log_team_member_action` (rejected before it), and an already-active row observed post-lock is rejected before the CAS even runs. `01e01b3a6` added a dedicated test for that path. |

Minor items from round 2 also addressed: `SecurityValidator.sanitize_log_message()` now wraps `invitation.email` in the concurrent-accept log line; doctest added to `TeamManagementService.log_team_member_action`; `Raises:` docstring updated. Branch head (`01e01b3a6`) matches this PR's head — nothing outstanding from either review round.

## Fresh review (branch-only diff vs `main`, 4 files, +298/-8)

Scope: `team_invitation_service.py`, `team_management_service.py`, 2 test files. No unrelated files touched.

**Security (CWE)**
- CWE-362 (Race Condition) — this is what the PR fixes, and it's now closed correctly cross-dialect (verified against a live SQLite engine compile). Might be worth calling this out explicitly as a CWE-362 fix in the commit/PR trailer, since it's a real concurrency vulnerability class rather than a plain bug — helps anyone auditing the diff later.
- No new injection, auth-bypass, or ownership-trust issues. Log sanitization is now consistent with the rest of the file.
- Team-capacity check (`check_team_member_capacity`) still runs before the lock is acquired, so a burst of concurrent invitation-accepts for different users on the same team could still slip past the limit slightly — pre-existing TOCTOU, not introduced or worsened by this PR, out of scope for this fix. Flagging for awareness only.

**Correctness / implementation match**
- Diff matches the PR description precisely: check existing → reuse-and-update or insert.
- One inefficiency worth a look: after the raw CAS `UPDATE(synchronize_session=False)` succeeds, the code also assigns `membership.role/joined_at/invited_by/is_active` on the ORM object. Because `synchronize_session=False` leaves the object's committed-state snapshot untouched, these assignments mark it dirty relative to the original (pre-CAS) values, so SQLAlchemy issues a second, redundant `UPDATE` for the same row on `commit()`. Not a correctness bug (idempotent, same values) — just an extra write per reactivation. Could avoid with `session.refresh(membership)` or `synchronize_session="fetch"`, but low priority.

**Testing**
- Unit coverage for the new code is strong (97% line coverage on `team_invitation_service.py`, all uncovered lines pre-existing/unrelated). Branches tested: fresh insert, reuse-and-reactivate, IntegrityError commit race, CAS-loses-race, lock-observes-already-active. Ran the full suite locally — all green, ruff clean, doctests pass.
- Gap (pre-existing, restated from both prior rounds, not introduced here): no integration/e2e test exercises `POST /teams/invitations/{token}/accept` end-to-end. Confirmed via search — nothing in `tests/integration` or `tests/e2e` hits that route. Might be worth a follow-up issue if not already tracked.
- No dead code found in the diff.

**Logs / exceptions**
- Logging is consistent and sanitized. Exception handling narrows correctly: `IntegrityError` → 400-mapped `ValueError` at the router (unchanged, already correct); generic `Exception` still falls through to 500 as before — no regression.

**Docs / breaking change**
- No breaking change — return type, endpoint contract, and status-code mapping are all unchanged; this only fixes a bug path that previously 500'd.
- No docs update needed (internal service-layer fix, no public API/schema change).
- Issue reference present: commit messages ("Closes #5524") and PR body both reference it.

**Necessity**
Yes — this is a real fix for a genuine correctness/concurrency bug (CWE-362) with a clear repro, and the cross-dialect concern raised in review was substantive (SQLite is the documented default DB), not a nitpick.

No blocking findings remain. The two notes above (redundant UPDATE, missing e2e coverage) are non-blocking follow-ups, not required for this PR.

msureshkumar88 (approved):
## E2E verification — real environment, all checks passed

Ran a real, non-mocked reproduction of issue #5524 against this branch (`01e01b3a6`): a real `uvicorn` process, a real SQLite DB (the project's default dialect, and specifically the one the round-2 review concern was about), real JWTs, real HTTP over `curl`. Script: `scripts/verify-pr-5543-e2e.sh` (added for this verification, not part of the PR diff).

### Config used

```
DATABASE_URL=sqlite:///./e2e_pr5543_test.db
AUTH_REQUIRED=true
ALLOW_TEAM_INVITATIONS=true
REQUIRE_EMAIL_VERIFICATION_FOR_INVITES=false   # skips the email-click-link step, unrelated to this PR
MCPGATEWAY_UI_ENABLED=false
PLUGINS_ENABLED=false
```

### Commands run

```bash
uv run uvicorn mcpgateway.main:app --host 127.0.0.1 --port 44553
uv run python -m mcpgateway.utils.create_jwt_token --username admin@example.com --exp 20 --secret <secret> --admin
curl -X POST /v1/teams/                                    # create team (admin = owner)
curl -X POST /auth/email/register                          # register invitee
curl -X POST /v1/teams/{team_id}/invitations                # invite #1
curl -X POST /v1/teams/invitations/{token}/accept           # accept #1 -> active membership
curl -X DELETE /v1/teams/{team_id}/members/{email}          # remove -> soft-deleted (is_active=False) row
curl -X POST /v1/teams/{team_id}/invitations                # invite #2 (the exact #5524 repro trigger)
curl -X POST /v1/teams/invitations/{token2}/accept          # accept #2 -> THE BUG PATH
curl -X GET  /v1/teams/{team_id}/members                    # verify no duplicate row
# then repeat remove + invite once more and fire two accepts of the SAME invitation
# token concurrently (double-submit scenario) to check the round-2 race-handling claim
```

### Output (condensed)

```
PASS: gateway server is up on :44553
PASS: admin JWT generated
PASS: created team
PASS: registered invitee user
PASS: created first invitation
PASS: first accept succeeded (200), membership now active
PASS: member removed (soft-deleted, is_active=False row remains)
PASS: created second invitation (fresh invite to a previously-removed member)
PASS: CORE CLAIM: second accept of a re-invited, previously-removed member succeeds (200)
      -- issue #5524 no longer reproduces
PASS: exactly one active membership row for invitee (no duplicate insert)
PASS: created third invitation for concurrency test
concurrent accept A: 200 {"...","is_active":true,...}
concurrent accept B: 400 {"detail":"Invitation is invalid or expired"}
PASS: CORE CLAIM (round-2): exactly one of two concurrent accepts wins (200),
      the other is cleanly rejected (400) -- no 500, no silent double-reactivation
PASS: post-race: exactly one active membership row (no duplicate/corrupt state)

RESULT: ALL CHECKS PASSED
```

The step-6 result is the direct reproduction of the bug in the PR description: invite → accept → remove → re-invite → accept, which 500'd before this fix and returns `200` with a single, correctly-reactivated membership row now.

**One honest caveat on the step-7 concurrency check**: both concurrent requests hit the same single-worker `uvicorn` event loop, so the second request's DB read happened *after* the first had already committed (real HTTP-level double-submit protection confirmed — no 500, no double-membership) — but this doesn't force the two DB transactions to genuinely interleave mid-flight the way the round-2 review's SQLite compare-and-swap concern requires. That exact interleaving (both readers observing the stale row before either writes) is what `test_accept_invitation_reactivation_race_loses_is_rejected` and `test_accept_invitation_lock_observes_already_reactivated_row` in `tests/unit/mcpgateway/services/test_team_invitation_service.py` exercise directly via mocked call ordering — e2e and unit coverage are complementary here, not redundant.

### Summary of changes required

None blocking — approving as-is. Two non-blocking follow-ups from the earlier review rounds still stand, neither gating this fix:

1. After the compare-and-swap `UPDATE` on the reuse path, the code also reassigns `membership.role/joined_at/invited_by/is_active` on the ORM object, which triggers a second, redundant `UPDATE` for the same row at `commit()` (idempotent, just an extra write per reactivation) — `team_invitation_service.py:391-395`.
2. No integration/e2e test lives in the repo for `POST /v1/teams/invitations/{token}/accept` (pre-existing gap; this review's `scripts/verify-pr-5543-e2e.sh` covers the scenario ad hoc but isn't wired into the test suite).

msureshkumar88 (approved):
LGMT

brian-hussey (approved):
LGTM

PR #5546, opened 2026-07-07
feat(ui-rewrite): Edit resources

### LINKED ISSUE #5098: [UI-REWRITE]: Edit Resources

gcgoncalves:
# Pull Request

## 🔗 Related Issue

Closes #5098 

https://github.com/user-attachments/assets/9664be57-b620-43b0-b319-ed3edc49c4f1


marekdano (commented):
The PR looks good! 

Just two suggestions for a11y

- **`Resources.tsx` (`handleEditResource`) — focus lost when the edit form opens.** "Edit" lives in a Radix `DropdownMenuItem` inside the details panel; the handler unmounts that panel (`setIsDetailsPanelOpen(false)`) and shows the form. Radix tries to restore focus to the now-unmounted trigger, so focus falls back to `<body>` — a keyboard/SR user gets no announcement of the new form. The close path already manages focus via `shouldRestoreFormCloseFocus` + `headingRef`; the open path should move focus to the form heading (`tabIndex={-1}` ref, focus on mount). 
- **`ResourcesTable.tsx` — every per-row menu trigger has the identical accessible name "More options."** Inside the panel each resource row uses the bare `resources.table.moreOptions` label (vs. the group-level "More options for {name}"). With several resources, a SR user can't tell which resource's Edit/Delete they're triggering. Parameterize the label with the resource name, matching the group-level pattern.

marekdano (commented):
Findings

### 🟡 Suggestions (worth addressing)

1. Content-load remount can discard in-progress edits (`Resources.tsx`)
The form is force-remounted via `key` when the content fetch transitions loading→loaded:
  ```tsx
  key={editingResource ? `${editingResource.id}-${editedResourceContent ? "loaded" : "loading"}` : "create"}
  ```
  Name/URI/description are already visible (from list data) during the `loading` phase, so a user on a slow connection can start editing before the `/resources/:id` GET resolves — and the remount then wipes those edits back to the seed values. Consider gating the form render until content has loaded (spinner), or seeding content imperatively via effect instead of remounting.

2. A failed content fetch silently breaks the edit form (`Resources.tsx:230`)
  ```tsx
  const { data: editedResourceContent } = useQuery<...>(`/resources/${editingResource?.id}`, {...});
  ```
  The `error`/`isLoading` from this query are discarded. If the GET fails, `editedResourceContent` stays `undefined` forever → the remount key is stuck at `loading`, `content` stays `""`, and since `content` is `min(1)` required, the user hits "Content is required" with no explanation. Surface a load error (or loading state) to make this recoverable.
 
3. Binary/blob resources can't be edited  `resourceForForm` only maps `editedResourceContent?.text`; the `blob` field is ignored:
  ```tsx
  { ...editingResource, content: editedResourceContent?.text ?? "" }
  ```
For a binary resource the content textarea is empty and required-content validation blocks the update — and even if it didn't, saving would overwrite the blob with empty text. Consider disabling Edit (or showing a "binary resource" notice) when `blob` is present.

  ---

### 🔵 Minor notes
 
- **Dead i18n key**: `resources.table.view` was added to all three locales (en/es/pt) but is never referenced in the client. Wire it up or drop it.
- **Stale name in success toast** (`Resources.tsx:311`): `handleFormSuccess` toasts with `editingResource.name` (the *original* name), so renaming shows "Resource {oldName} updated successfully". The submitted `name` would be more accurate.
- **Unlisted MIME types silently drop**: `mimeType` seeds via `(resource.mimeType as MimeType | null) ?? ""`. If the stored value isn't one of the seven `MIME_TYPES`, the Select shows blank and the mime type may be lost on save.
- **URI is freely editable in edit mode** — backend allows it (`uri` optional on update), so may be intentional as a rename, but changing a resource's URI can break clients referencing it and there's no warning. Worth confirming it's by design.
- **Style nit** (`ResourceDetailsPanel.tsx`): the named `ResourceDetailsPanelProps` interface was inlined into an anonymous type literal on the signature; the named interface read a bit better.

marekdano (approved):
LGTM 🚀

PR #5548, opened 2026-07-07
fix(ui-rewrite): group prompts by MCP server on the Prompts page

### LINKED ISSUE #5101: [UI-REWRITE]: List Prompts

marekdano:
### Summary

Closes #5101 

The Prompts page rendered every prompt as its own card, unlike the Tools page which groups tools by the MCP server (gateway) that provides them. This aligns Prompts with that pattern, per the [Figma design](https://www.figma.com/design/H2Dr9QPvRVxCe92V8URvGT/Refiring?node-id=4029-25955):

  - **Gateway-sourced prompts** are grouped into one card per MCP server, titled with the server slug, with each prompt shown as a name badge. Groups larger than 8 collapse the remainder into a `+N` overflow badge.
  - **Gateway-less prompts** (local templates) are collapsed into a single **"REST prompts"** card, mirroring the existing "REST tools" grouping on the Tools page.

  ## Changes

  - `client/src/pages/Prompts.tsx` — replaced the flat prompt list with `buildPromptGroups()`, which keys by `gatewaySlug || restPromptsLabel`; introduced `PromptGroupCard` rendering name badges with `+N` overflow; a real prompt description surfaces as the badge tooltip.
  - `client/src/types/prompts.ts` — added the `PromptGroup` type.
  - `client/src/i18n/locales/{en-US,es-ES,pt-BR}/prompts.json` — added `prompts.restPromptsGroup` and `prompts.card.morePromptsTitle`.
  - `client/src/pages/Prompts.test.tsx` — updated/added tests for gateway grouping, REST-prompts collapse (single card + single actions menu), `+N` overflow, and description-tooltip filtering.

  ## Testing

  - 6 unit tests pass; ESLint and `tsc --noEmit` clean.
  - Coverage for `Prompts.tsx`: 100% statements / lines / functions, 96.42% branches (above the 95% gate).

### DIFF
--- a/client/src/pages/Prompts.tsx
+++ b/client/src/pages/Prompts.tsx
@@ -2,6 +2,7 @@ import { useMemo } from "react";
 import { MessageSquareCode, MoreHorizontal, Plus } from "lucide-react";
 import { useIntl } from "react-intl";
 import { Card, CardContent, CardHeader } from "@/components/ui/card";
+import { CardTag } from "@/components/ui/card-tag";
 import { Button } from "@/components/ui/button";
 import { useQuery } from "@/hooks/useQuery";
 import {
@@ -39,16 +40,20 @@ function buildPromptGroups(prompts: Prompt[], restPromptsLabel: string): PromptG
   const map = new Map<string, PromptGroup>();
 
   for (const prompt of prompts) {
-    const slug = prompt.gatewaySlug?.trim() || restPromptsLabel;
-    let group = map.get(slug);
+    const slug = prompt.gatewaySlug?.trim();
+    // Namespace the map key so gateway-less prompts (keyed "rest") can never
+    // merge into a real gateway whose slug happens to equal the localized
+    // "REST prompts" label. `label` stays purely for display.
+    const key = slug ? `gateway:${slug}` : "rest";
+    let group = map.get(key);
     if (!group) {
       group = {
-        key: slug,
-        label: slug,
+        key,
+        label: slug || restPromptsLabel,
         gatewayId: prompt.gatewayId,
         prompts: [],
       };
-      map.set(slug, group);
+      map.set(key, group);
     }
     group.prompts.push(prompt);
   }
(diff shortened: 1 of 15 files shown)


client/src/pages/Prompts.tsx
```diff
@@ -9,7 +10,146 @@ import {
...
+
+// Group prompts sourced from the same MCP server (gateway) into a single card,
+// mirroring how the Tools page groups tools. Prompts without a gateway (local
+// templates) are collapsed into a single "REST prompts" card, matching the
+// "REST tools" grouping on the Tools page.
+function buildPromptGroups(prompts: Prompt[], restPromptsLabel: string): PromptGroup[] {
+  const map = new Map<string, PromptGroup>();
+
+  for (const prompt of prompts) {
+    const slug = prompt.gatewaySlug?.trim() || restPromptsLabel;
```
vishu-bh: Nice grouping logic overall. One edge case worth fixing: using the localized `restPromptsLabel` as the actual map key can merge gateway-less prompts with a real gateway whose slug is also `REST prompts` / localized equivalent. Could we use separate stable keys, for example `rest` for local prompts and `gateway:${prompt.gatewayId ?? slug}` for gateway prompts, while keeping `label` only for display?

client/src/pages/Prompts.tsx
```diff
@@ -9,7 +10,146 @@ import {
...
+        </div>
+      </CardHeader>
+
+      <CardContent>
+        <div className="flex flex-wrap gap-1">
+          {visiblePrompts.map((prompt) => (
+            <span
+              key={prompt.id}
+              className="inline-flex items-center rounded bg-tool-badge-bg px-1.5 py-1 text-[10px] font-medium leading-none text-white"
+              title={getPromptDescription(prompt) ?? undefined}
```
vishu-bh: Since UI has `components/ui/badge` and `components/ui/tooltip`, could we use those instead of raw `<span title=...>` badges? That keeps this page aligned with the shadcn component rule and gives better keyboard/mobile tooltip behavior than native `title`.

vishu-bh (changes requested):
Added some inline comments

client/src/pages/Prompts.tsx
marekdano: Fixed!

client/src/pages/Prompts.tsx
marekdano: I've created a new shared custom component `CardTag`, instead of using `Badge`. The `CardTag` component uses shadcn-ui Tooltip.

vishu-bh (approved):
LGTM 🚀

PR #5550, opened 2026-07-07
fix(ui-rewrite): edit teams from the Teams page

### LINKED ISSUE #5107: [UI-REWRITE]: Edit Teams

marekdano:
Closes #5107 

### Summary

Adds the ability to edit an existing team from the Teams UI. The **Edit** action in each team row's dropdown now opens the team form pre-populated with the team's details; saving issues a `PUT /teams/{id}`. The form is reused for both create and edit, and the max-members selector was hardened to handle values the backend allows but the UI didn't previously surface.

### Changes

**Edit flow**
  - `TeamsTable`'s existing **Edit** menu item is now wired through `Teams.tsx` (`onEdit` → `teamToEdit`), opening `TeamForm` in edit mode.
  - `useTeamForm` accepts an optional `team`: in edit mode it pre-fills name/description/visibility/max-members and submits via a new `updateTeam()` API helper (`PUT /teams/{id}`) instead of `createTeam`.
  - Edit mode swaps the title, description, and submit-button copy ("Edit team" / "Save changes") and hides the member-management section — membership is managed through its own dedicated dialog, and the update endpoint doesn't touch members.
  - On success, the page shows an "updated" toast and refetches.

  **Max-members correctness**
  - The selector now renders its options dynamically. If a team's `max_members` falls outside the preset list (e.g. a custom cap set via the API), the current value is injected as a selectable option instead of the dropdown rendering blank.
  - `max_members` is only sent in the update when the user actually changes it. This preserves a team's existing value — including a `null` "no override" — rather than silently pinning it to the displayed default, matching the backend's `UNSET`/`model_fields_set` semantics.

  **i18n**
  - Added `teams.edit.*` keys (title, description, saving, submit, success) across `en-US`, `es-ES`, and `pt-BR`.

### DIFF
--- a/client/src/hooks/useTeamForm.ts
+++ b/client/src/hooks/useTeamForm.ts
@@ -188,7 +188,11 @@ export function useTeamForm(team?: Team): UseTeamFormReturn {
           const maxMembersChanged = maxMembers !== initialMaxMembers(team);
           await updateTeam(team.id, {
             name: name.trim(),
-            description: description.trim() || undefined,
+            // Send the trimmed value (an empty string when cleared) rather than
+            // `undefined`. The backend only overwrites the description when the
+            // field is present and non-null, so omitting it would silently keep
+            // the old value when a user intends to clear it.
+            description: description.trim(),
             visibility,
             ...(maxMembersChanged ? { max_members: parseInt(maxMembers, 10) } : {}),
           });
--- a/client/src/hooks/useTeamForm.ts
+++ b/client/src/hooks/useTeamForm.ts
@@ -3,9 +3,10 @@ import { useIntl } from "react-intl";
 import { toast } from "sonner";
 import { z } from "zod";
 import { api } from "@/api/client";
-import { createTeam, addTeamMember } from "@/api/teams";
+import { createTeam, updateTeam, addTeamMember } from "@/api/teams";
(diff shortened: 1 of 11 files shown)


client/src/hooks/useTeamForm.ts
```diff
@@ -147,7 +180,23 @@ export function useTeamForm(): UseTeamFormReturn {
...
-        const team = await createTeam({
+        if (isEditMode) {
+          // Only send max_members when the user actually changed it. Omitting it
+          // preserves the team's existing value — including a null "no override"
+          // that the backend would otherwise leave untouched, rather than
+          // silently pinning it to the displayed default.
+          const maxMembersChanged = maxMembers !== initialMaxMembers(team);
+          await updateTeam(team.id, {
+            name: name.trim(),
+            description: description.trim() || undefined,
```
vishu-bh: When a user clears an existing description, `description.trim() || undefined` sends `undefined`, so JSON serialization omits the field and the backend treats it as unchanged. The save appears successful, but the old description remains. 

Might need to check this

vishu-bh (changes requested):
Code changes are clean and solid. Code coverage is also good. Just minor functionality needs checking

vishu-bh (approved):
Thanks @marekdano 

PR is clean. LGTM 👍

PR #5565, opened 2026-07-09
Drop stale unique_gateway_user constraint on oauth_tokens (#5538)

### LINKED ISSUE #5538: [BUG]: OAuth callback fails with UNIQUE constraint error when a second user authorizes the same gateway
## Summary

When a second ContextForge user tries to authorize a gateway that another user has already
authorized via OAuth (e.g. a shared GitHub Copilot MCP gateway), the OAuth callback returns
HTTP 400 with the error:

UNIQUE constraint failed: oauth_tokens.gateway_id, oauth_tokens.user_id
The first user's token is stored successfully. Every subsequent user is blocked from completing
OAuth authorization for that gateway.

---

## Steps to Reproduce

1. Create a gateway with GitHub OAuth configured (e.g. `https://api.githubcopilot.com/mcp/`).
2. Log in as `user@example.com` and authorize the gateway via `/oauth/authorize/<gateway_id>`.
   → Authorization succeeds. Token stored in `oauth_tokens`.
3. Log in as `user2@example.com` and authorize the same gateway.
   → OAuth callback fails with HTTP 400. Token is not stored.

---

## Error Log

ERROR mcpgateway.routers.oauth_router - OAuth callback failed: Token storage failed:
(sqlite3.IntegrityError) UNIQUE constraint failed: oauth_tokens.gateway_id, oauth_tokens.user_id
[SQL: INSERT INTO oauth_tokens (id, gateway_id, user_id, app_user_email, ...) VALUES (?, ?, ?, ?, ...)]
[parameters: ('<new-id>', '<gateway_id>', 'Ov23li373LN81kl4mZCX', '[user2@example.com](mailto:user2@example.com)', ...)]
After this failure, if `user2@example.com` tries to use a tool on the gateway they get:

RPC error: Tool invocation failed: Please authorize git-server first.
Visit /oauth/authorize/<gateway_id> to complete OAuth flow.
---

## Root Cause

The `oauth_tokens` table has **two competing unique constraints** left over from a schema
migration that was not completed cleanly.

| Constraint | Columns | Source |
|---|---|---|
| `unique_gateway_user` | `gateway_id, user_id` | Migration `94f64b8e282f` (original, stale) |
| `uq_oauth_gateway_user` | `gateway_id, app_user_email` | Migration `14ac971cee42` + ORM model (current, correct) |

The critical distinction is what `user_id` holds. It is populated from the **OAuth App
`client_id`** — not from a per-person GitHub user identity. The `client_id` is the same value
for every ContextForge user who authorizes via the same OAuth App registration.

Migration `14ac971cee42` correctly added `app_user_email` to scope tokens per ContextForge user
and created `uq_oauth_gateway_user (gateway_id, app_user_email)`, but it **never dropped the
old `unique_gateway_user (gateway_id, user_id)` constraint**. The stale constraint fires on
every INSERT for a second user because `gateway_id + client_id` is always the same tuple.

The application logic in `token_storage_service.py` and the ORM model in `db.py` are both
correct — the problem is purely a stale database schema constraint.

**Affected files:**
- `mcpgateway/alembic/versions/94f64b8e282f_add_oauth_tokens_table.py` (line 53) — created the stale constraint
- `mcpgateway/alembic/versions/14ac971cee42_add_user_context_to_oauth_tokens.py` — did not drop it
- `mcpgateway/db.py:5284` — ORM model is correct (uses `app_user_email`)
- `mcpgateway/services/token_storage_service.py:157` — upsert logic is correct

---

## Expected Behavior

Each ContextForge user (`app_user_email`) should be able to independently authorize any gateway.
The unique constraint should be `(gateway_id, app_user_email)` — one token per app user per
gateway — which is what the ORM model already defines.

---

## Actual Behavior

Only the first ContextForge user to authorize a given gateway succeeds. All subsequent users
receive a 400 error during the OAuth callback and cannot use tools requiring that gateway's
OAuth token.

Environment
* Database: SQLite

winklemad:
## Summary

Fixes #5538. The `oauth_tokens` table carries a stale unique constraint `unique_gateway_user (gateway_id, user_id)` created by migration `94f64b8e282f`. Migration `14ac971cee42` added the correct per-user scope `uq_oauth_gateway_user (gateway_id, app_user_email)` but never dropped the old one.

Because `user_id` is populated from the shared OAuth **client_id** (identical for every ContextForge user of the same OAuth App), the stale constraint fires on every second user who authorizes the same gateway. The OAuth callback fails with HTTP 400:

```
UNIQUE constraint failed: oauth_tokens.gateway_id, oauth_tokens.user_id
```

Only the first user can ever authorize; everyone else is blocked (and then sees `Please authorize <gateway> first` on tool calls).

## Fix

A new migration (`b3e9f1a7c2d5`, on top of the current head `e198602c3c1e`) drops the stale `unique_gateway_user` constraint. It is:

- **Existence-guarded** — inspects the table first, so it is a no-op when the constraint is already absent (fresh installs / re-runs).
- **SQLite-safe** — uses `batch_alter_table`; on PostgreSQL this is a direct `DROP CONSTRAINT`.
- **Reversible** — `downgrade()` recreates the constraint.

The per-user `uq_oauth_gateway_user (gateway_id, app_user_email)` uniqueness is untouched. The ORM model (`db.py`) and `token_storage_service.py` were already correct — this is purely the stale schema constraint.

## Verification

Ran the full migration chain on a fresh SQLite DB:

- **Before:** `oauth_tokens` has `unique_gateway_user`; a second user's token INSERT fails with the reported `IntegrityError`.
- **After:** the stale constraint is gone, the per-user unique index survives, a second user can authorize, and a true duplicate `(gateway_id, app_user_email)` is still rejected.
- Downgrade → upgrade round-trip restores and re-drops the constraint cleanly.

Adds `tests/migration/test_oauth_tokens_unique_constraint.py`, which runs the chain and asserts all of the above; it fails without the migration and passes with it. `ruff`, `black`, and `isort` are clean, and the existing OAuth router/manager unit tests pass.

### DIFF
--- a/mcpgateway/alembic/versions/b3e9f1a7c2d5_drop_stale_oauth_tokens_unique_gateway_user.py
+++ b/mcpgateway/alembic/versions/b3e9f1a7c2d5_drop_stale_oauth_tokens_unique_gateway_user.py
@@ -1,0 +1,51 @@
+# -*- coding: utf-8 -*-
+"""Location: ./mcpgateway/alembic/versions/b3e9f1a7c2d5_drop_stale_oauth_tokens_unique_gateway_user.py
+Copyright 2026
+SPDX-License-Identifier: Apache-2.0
+
+drop stale unique_gateway_user constraint on oauth_tokens (issue #5538)
+
+Revision ID: b3e9f1a7c2d5
+Revises: e198602c3c1e
+Create Date: 2026-07-09
+"""
+
+# Standard
+from typing import Sequence, Union
+
+# Third-Party
+from alembic import op
+import sqlalchemy as sa
+
+revision: str = "b3e9f1a7c2d5"  # pragma: allowlist secret
+down_revision: Union[str, Sequence[str], None] = "e198602c3c1e"  # pragma: allowlist secret
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    """Drop the stale unique_gateway_user (gateway_id, user_id) constraint on oauth_tokens.
+
+    Per-user token scope is enforced by uq_oauth_gateway_user (gateway_id,
+    app_user_email). The older gateway_id + user_id constraint predates the
+    app_user_email column and blocks a second user from authorizing the same
+    gateway, because user_id holds the shared OAuth client_id (issue #5538).
+    """
+    inspector = sa.inspect(op.get_bind())
+    if "oauth_tokens" not in inspector.get_table_names():
+        return
+    constraint_names = {uc["name"] for uc in inspector.get_unique_constraints("oauth_tokens")}
+    if "unique_gateway_user" in constraint_names:
+        with op.batch_alter_table("oauth_tokens", schema=None) as batch_op:
+            batch_op.drop_constraint("unique_gateway_user", type_="unique")
+
+
+def downgrade() -> None:
+    """Restore the unique_gateway_user (gateway_id, user_id) constraint on oauth_tokens."""
+    inspector = sa.inspect(op.get_bind())
+    if "oauth_tokens" not in inspector.get_table_names():
+        return
+    constraint_names = {uc["name"] for uc in inspector.get_unique_constraints("oauth_tokens")}
+    if "unique_gateway_user" not in constraint_names:
+        with op.batch_alter_table("oauth_tokens", schema=None) as batch_op:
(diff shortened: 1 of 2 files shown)


jonpspri:
Closing as superseded. The linked issue #5538 ([BUG]: OAuth callback fails with UNIQUE constraint error) was closed as COMPLETED — the fix was already merged via another PR. This PR no longer has an open issue to address.

PR #5569, opened 2026-07-09
Fix templated resource proxy reads

lucarlig:
## Summary
- Bypass stale unscoped template cache for live `read_resource()` template resolution so scoped/authenticated reads use current template rows.
- Preserve empty upstream resource responses by checking `is not None` instead of truthiness.
- Fail templated proxy reads when gateway resolution returns no content instead of returning the expanded URI placeholder as resource text.

## Root cause
`read_resource()` resolved templated resources through `_template_cache`, which is keyed only by template name and can be stale or unscoped during gateway federation. When the subsequent gateway read returned `None`, the placeholder text such as `reference://users/7` was left in the response, causing clients expecting JSON to parse the URI string.

## Validation
- `uv run pytest tests/unit/mcpgateway/services/test_resource_service.py -q`
- `uv run ruff check mcpgateway/services/resource_service.py tests/unit/mcpgateway/services/test_resource_service.py`
- `uv run pre-commit run interrogate --files mcpgateway/services/resource_service.py tests/unit/mcpgateway/services/test_resource_service.py`
- Commit hooks passed.

### DIFF
--- a/mcpgateway/services/resource_service.py
+++ b/mcpgateway/services/resource_service.py
@@ -2460,7 +2460,9 @@ class ResourceService(BaseService):
                         # it internally checks which uri matches the pattern of modified uri and fetches
                         # the one which matches else raises ResourceNotFoundError
                         try:
-                            template_user_email = user.get("email") if isinstance(user, dict) else user
+                            from mcpgateway.auth_context import get_user_email  # pylint: disable=import-outside-toplevel
+
+                            template_user_email = None if user is None else get_user_email(user)
                             content = (
                                 await self._read_template_resource(
                                     db,
@@ -2469,7 +2471,7 @@ class ResourceService(BaseService):
                                     user_email=template_user_email,
                                     token_teams=token_teams,
                                     server_id=server_id,
-                                    use_cache=False,
+                                    use_cache=True,
                                 )
                                 or None
                             )
@@ -2562,6 +2564,12 @@ class ResourceService(BaseService):
                     placeholder = getattr(content_obj, attr_name, None)
                     content_uri = getattr(content_obj, "uri", None)
                     if template_uri and requested_uri and placeholder is not None and content_uri is not None and str(placeholder) == str(requested_uri) and str(content_uri) == str(template_uri):
+                        logger.warning(
+                            "Resource template proxy read returned no content for requested URI '%s' from template '%s' via gateway '%s'",
+                            requested_uri,
+                            template_uri,
+                            getattr(resource_db_gateway, "id", None) or getattr(resource_db, "gateway_id", None),
+                        )
                         raise ResourceError(f"Resource template '{template_uri}' did not resolve URI '{requested_uri}'")
 
                 if isinstance(content, (ResourceContent, ResourceContents, TextContent)):
--- a/mcpgateway/services/resource_service.py
+++ b/mcpgateway/services/resource_service.py
@@ -2460,7 +2460,19 @@ class ResourceService(BaseService):
                         # it internally checks which uri matches the pattern of modified uri and fetches
                         # the one which matches else raises ResourceNotFoundError
                         try:
-                            content = await self._read_template_resource(db, uri) or None
+                            template_user_email = user.get("email") if isinstance(user, dict) else user
+                            content = (
+                                await self._read_template_resource(
+                                    db,
+                                    uri,
+                                    include_inactive=include_inactive,
+                                    user_email=template_user_email,
+                                    token_teams=token_teams,
+                                    server_id=server_id,
+                                    use_cache=False,
+                                )
+                                or None
+                            )
                             # ═══════════════════════════════════════════════════════════════════════════
                             # SECURITY: Fetch the template's DbResource record for access checking
                             # _read_template_resource returns ResourceContent with the template's ID
@@ -2539,6 +2551,19 @@ class ResourceService(BaseService):
                 # ResourceContents covers TextResourceContents and BlobResourceContents (MCP-compliant)
                 # ResourceContent is the legacy model for backwards compatibility
 
+                def _set_gateway_content(content_obj: Any, attr_name: str, resource_response: Any) -> None:
+                    """Apply gateway content or reject unresolved template placeholders."""
+                    if resource_response is not None:
+                        setattr(content_obj, attr_name, resource_response)
+                        return
+
+                    template_uri = getattr(resource_db, "uri_template", None) if resource_db else None
+                    requested_uri = uri if uri is not None else original_uri
+                    placeholder = getattr(content_obj, attr_name, None)
+                    content_uri = getattr(content_obj, "uri", None)
+                    if template_uri and requested_uri and placeholder is not None and content_uri is not None and str(placeholder) == str(requested_uri) and str(content_uri) == str(template_uri):
+                        raise ResourceError(f"Resource template '{template_uri}' did not resolve URI '{requested_uri}'")
+
                 if isinstance(content, (ResourceContent, ResourceContents, TextContent)):
                     # Metrics are recorded in read_resource finally block for all resources
                     resource_response = await self.invoke_resource(
@@ -2552,8 +2577,7 @@ class ResourceService(BaseService):
                         gateway_obj=resource_db_gateway,
                         server_id=server_id,
                     )
-                    if resource_response:
-                        setattr(content, "text", resource_response)
+                    _set_gateway_content(content, "text", resource_response)
                 # If content is any object that quacks like content
                 elif hasattr(content, "text") or hasattr(content, "blob"):
                     # Metrics are recorded in read_resource finally block for all resources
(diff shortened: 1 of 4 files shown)


msureshkumar88 (changes requested):
Thanks for the clear root-cause writeup and the focused test additions — the fail-closed placeholder fix and the `is not None` empty-response fix are both good, narrowly-scoped corrections, and the new tests for the "none response raises" / "empty text preserved" behaviors are exactly the right shape.

One thing I'd like addressed before merge, plus a couple of smaller suggestions:

## Requesting changes on

**Cache is effectively disabled for every templated read, not just scoped ones.** The only production call site of `_read_template_resource()` (`read_resource()`, around resource_service.py:2465) now hardcodes `use_cache=False` unconditionally. `_read_template_resource` already has a `scoped_lookup` check internally that correctly decides when the cache is unsafe to use — but since the call site always passes `use_cache=False`, that internal logic never gets a chance to reuse the cache even for a fully anonymous, unscoped read. Net effect: `_template_cache` is now dead code in the running app, and every templated resource read pays a full `list_resource_templates()` DB round trip (visibility-filter query, plus an extra `is_user_admin()` query on some branches) instead of hitting cache for the common case.

Suggest: pass `use_cache=True` from the call site and let the existing `scoped_lookup` branch inside `_read_template_resource` gate the bypass. That keeps the staleness fix for scoped/authenticated reads while restoring the cache for anonymous/unscoped ones.

## Suggestions (non-blocking)

- Test coverage: the new `_set_gateway_content` behaviors (raise-on-unresolved-placeholder, preserve-empty-response) are only tested on the `text` attribute path — the `blob` path isn't covered for either new behavior.
- No integration/e2e test drives the actual staleness scenario end-to-end (real DB, template added/changed between reads) — current coverage is all unit-level with mocked `list_resource_templates`/`db`. Worth adding given this is the exact bug being fixed.
- No CHANGELOG entry or issue number reference (`Closes #...`) — worth adding if there's a tracking issue.

## Minor notes

- Not a breaking change — all new params are optional with defaults.
- No unrelated changes in the diff — scope is clean.
- Root cause as fixed is a stale-cache/data-integrity bug, not a live authz bypass — `_check_resource_access` re-checks the resolved resource fresh regardless of how the cache produced the match, so worth keeping that framing in the changelog/release notes rather than describing it as a security fix for an exploitable bypass.

msureshkumar88 (changes requested):
Follow-up pass — a few things not covered in the earlier review.

## Requesting changes on

**`ResourceError` gets swallowed to `""` on the MCP streamable-HTTP resource-read path.** `transports/streamablehttp_transport.py`'s `@mcp_app.read_resource()` handler wraps the whole call in `except Exception as e: logger.exception(...); return ""`. That handler is exactly the entry point MCP clients hit (same transport family as the `gateway_proxy-http` e2e case in this PR). So for that transport, the new fail-closed behavior degrades back to a silent wrong value — empty string instead of the placeholder URI, still not a real error signal reaching the client. Given the PR's whole point is "fail templated proxy reads instead of returning bad text," this path currently doesn't get that benefit. Worth either narrowing that transport's except clause to let `ResourceError`/`ResourceNotFoundError` propagate as an MCP-level error, or explicitly deciding "empty string is an acceptable degradation for this transport" and noting why.

## Suggestions (non-blocking)

- `template_user_email = user.get("email") if isinstance(user, dict) else user` (resource_service.py, in the `_read_template_resource` call) reimplements a subset of `get_user_email()` (`auth_context.py`) instead of calling it — misses the `sub` fallback and the `'unknown'` default that helper provides. `read_resource()`'s own signature types `user` as `Optional[str]`, and every current caller already pre-resolves to a string before calling in, so the dict branch looks unreachable today — but if a future caller ever does pass a dict, this diverges from the canonical email-over-sub precedence instead of matching it. Suggest calling `get_user_email(user)` directly.
- No log line before the new `raise ResourceError(...)` for the unresolved-placeholder case — just the exception message. Since this represents a real upstream/gateway failure, a `logger.warning` with the uri/template/gateway would help diagnosing it in prod without needing to parse exception text out of a 404 body.

## Minor notes

- The `except Exception as e: raise ResourceNotFoundError(...) from e` wrapping the template lookup now also wraps the new scoped `list_resource_templates()` DB call, so DB/connectivity failures there get folded into a generic 404 too — same pre-existing pattern as the rest of the function, just flagging that the scoped lookup adds another failure mode under that same broad catch.

lucarlig:
@msureshkumar88 thanks for the careful follow-up. Pushed `398a44d9a` addressing the requested changes and the small suggestions that fit the same scope.

Addressed:
- Restored template caching for unscoped reads by passing `use_cache=True` from `read_resource()`; scoped reads still bypass via `_read_template_resource()`'s existing `scoped_lookup` guard.
- Let `ResourceError` / `ResourceNotFoundError` propagate from the streamable-HTTP resource-read handler instead of being converted to `""`.
- Switched template user extraction to the canonical `get_user_email()` helper while preserving `None` for anonymous reads so the unscoped cache path remains available.
- Added a warning before the unresolved-placeholder `ResourceError` with requested URI, template URI, and gateway id context.
- Added coverage for unscoped cache reuse, blob unresolved-placeholder failure, empty blob preservation, and streamable-HTTP resource-error propagation.

I did not add a changelog / `Closes #...` reference because I do not see a tracking issue attached to this PR.

msureshkumar88 (approved):
Re-reviewed after `398a44d9a`. Thanks for the quick turnaround — this addresses the blocking issue cleanly.

**Blocking issue from previous review: resolved.** The hardcoded `use_cache=False` bypass is gone. `_read_template_resource()` now gates the cache with a `scoped_lookup` check (`server_id is not None or user_email is not None or token_teams is not None`), so unscoped/anonymous reads still benefit from the cache while any authenticated/team/server-scoped read gets a fresh, correctly-scoped lookup. That's the right tradeoff — correctness over cache-hit-rate for scoped calls — and it's well covered by the new tests (`test_read_resource_template_scoped_lookup_ignores_stale_cache`, `test_read_resource_template_unscoped_lookup_uses_existing_cache`).

Also confirmed:
- Empty-string/blob responses are correctly preserved via the `is not None` check (vs. the old truthy check).
- The unresolved-placeholder guard in `_set_gateway_content` only fires when the content still matches the raw placeholder, with test coverage for both the text and blob paths.
- `ResourceError`/`ResourceNotFoundError` now propagate through the streamable-HTTP handler instead of being swallowed to `""` — this is the actual fix for the original bug (client parsing a raw `reference://users/7` string as JSON).
- Ruff clean, full unit suite green, no unrelated files touched, no breaking change (new kwargs are optional/backward compatible).

A couple of small, non-blocking notes for a possible follow-up:
1. `streamablehttp_transport.py` — `except (ResourceError, ResourceNotFoundError): raise` appears twice; `ResourceNotFoundError` already subclasses `ResourceError`, so it's redundant in the tuple. Purely cosmetic.
2. The new `logger.warning(...)` in `_set_gateway_content` logs `requested_uri`, which is client-controlled, without running it through `SecurityValidator.sanitize_log_message()` (used elsewhere in this file for gateway ids). Worth wrapping to guard against log-forging via a crafted URI — this mirrors an existing pattern elsewhere in the codebase rather than being new to this PR, so not a blocker.
3. No e2e coverage yet for the scoped-cache-bypass or unresolved-placeholder path specifically (unit coverage is thorough). Could be worth a follow-up e2e case, not required here.

Approving — nice fix.

ja8zyjits:
## ⚠️ Warnings (Non Blocking)

### Warnings

- [ ] **Template resource resolution behavior undocumented**
  - **Issue**: PR adds scoped template lookup with `user_email`, `token_teams`, `server_id` parameters to `_read_template_resource()`, but no documentation explains when/why scoped vs unscoped lookups occur
  - **Code**: `mcpgateway/services/resource_service.py:3790-3819`
  - **Docs**: `docs/docs/manage/api-usage.md:940-946` only shows basic `/resources/templates/list` endpoint
  - **Impact**: Users/developers won't understand template resolution behavior with multi-tenancy
  - **Fix**: Add section to API docs explaining:
    - Template cache behavior (unscoped by default)
    - When scoped lookups bypass cache (user_email/token_teams/server_id present)
    - How template visibility interacts with token scoping

- [ ] **Template proxy error handling undocumented**
  - **Issue**: PR adds `_set_gateway_content()` helper that raises `ResourceError` when gateway returns `None` for template resolution, but error behavior not documented
  - **Code**: `mcpgateway/services/resource_service.py:2556-2573`
  - **Docs**: No mention in `docs/docs/manage/api-usage.md` or architecture docs
  - **Impact**: API consumers won't know to expect `ResourceError` for unresolved templates
  - **Fix**: Document in API reference:
    - Template proxy reads can fail with `ResourceError` if gateway returns no content
    - Empty string/bytes responses are valid and preserved
    - Error message format: `"Resource template '{template_uri}' did not resolve URI '{requested_uri}'"`

- [ ] **Resource template visibility scoping not explicitly documented**
  - **Issue**: PR changes template lookup to respect `user_email`/`token_teams`/`server_id` for scoped access, but multitenancy docs don't explicitly cover template resources
  - **Code**: `mcpgateway/services/resource_service.py:3825-3836` (calls `list_resource_templates` with scoping params)
  - **Docs**: `docs/docs/architecture/multitenancy.md` covers general resource visibility but doesn't mention templates specifically
  - **Impact**: Unclear if template resources follow same private/team/public visibility rules
  - **Fix**: Add clarification to multitenancy docs that template resources follow same visibility/scoping rules as regular resources

---

ja8zyjits (approved):
LGTM

Issue #5563, opened 2026-07-09
[FEATURE]: Local-prompt variant of the prompt details drawer


a-effort:
## Background

The prompt details drawer currently assumes prompts are federated from a connected MCP server: it shows a peer-navigation chip row of sibling prompts, no template, and no local-authorship metadata. Prompts added manually via the [add Prompt form](https://www.figma.com/design/H2Dr9QPvRVxCe92V8URvGT/Refiring?m=auto&node-id=4029-25663&t=3820wIVLm41YI8BY-1) need some adjustments. 

Once `PromptDetailsPanel.tsx` lands (wire-up currently split across #5448, #5101, and #5323), the panel needs a second variant tailored to locally-authored prompts.

## Proposed changes

**Header**
Subtitle example: `Local prompt · Public · 3 arguments` (source · visibility · arg count)

**Tabs:**

`Try it` tab: [mockup](https://www.figma.com/design/H2Dr9QPvRVxCe92V8URvGT/Refiring?m=auto&node-id=4717-135700&t=3820wIVLm41YI8BY-1)
- One input per template variable, with:
  - Raw argument names as labels (`document_type*`, not `Document type`) so the label matches the JSON key the caller actually sends
 
`Definition` tab: [mockup](https://www.figma.com/design/H2Dr9QPvRVxCe92V8URvGT/Refiring?m=auto&node-id=4717-136515&t=3820wIVLm41YI8BY-1)
- Raw template (read-only + copy)
- Arguments JSON schema (read-only + copy)

## Related

- #5448 — preview + panel scaffold
- #5101 — prompt cards + (unwired) `View details` menu item
- #5323

a-effort:
Adding needed changes [here](https://github.com/IBM/mcp-context-forge/pull/5575), removing `definition` requirement.

PR #5575, opened 2026-07-09
feat(ui-rewrite): PromptDetailsPanel drawer with Code tab + Preview (#5448, #5323)

gcgoncalves:
## Summary

Wires the prompt details drawer (scaffold from #5323) with a code tab and render-only preview action on the Prompts page (#5448). Notes on ux [here](https://ibm.box.com/s/2gijfk3cneggkqtcuvyh847nglxvfbnl)

## For users

- Four-tab snippet view (curl, JSON-RPC, Python, TypeScript), prefilled with the current args and copyable.
- Render-only Preview against `POST /prompts/{id}`. Runs the plugin pipeline, returns the templated MCP `messages` array, does not invoke an LLM. Status row reports the real HTTP code and client-measured render time.
- Controlled args form driven by the prompt's declared `arguments`. Same args object feeds both the snippets and the preview request, so the code sample matches what gets sent.

The existing LLM-execution Test modal is unchanged.

## Structure

Public seam: `<PromptCodeTab prompt={NonNullable<PromptRead>} />`.

Composition:
- `PromptArgsForm`, `PromptSnippetTabs`, `PromptPreviewButton`, `PromptPreviewResult`.
- `usePromptPreview` owns the request lifecycle, `performance.now()` timing, `ApiError.body.detail` unwrap, and captures the real HTTP status.
- Snippet builders are pure functions of `{ promptName, args }`, covered by a 19-case escape-safety matrix (quotes, backslashes, newlines, dollar-sign literals, apostrophes, smart quotes) and URL-encoding tests.

## API

`promptsApi.render(name, args)` returns `{ rendered, status }` via the new `api.postWithMeta` client helper. Preserves cookie auth, CSRF forwarding, and the 401-to-login redirect. Name validation mirrors `mcpgateway/utils/security_validator.py` and throws synchronously.

## Shared primitives

- `ui/tabs.tsx`: radix Tabs wrapper, text-only tab row matching `MCPServerDetailsPanel`.
- `ui/code-block.tsx`: `prism-react-renderer` (`vsDark`) for `bash`, `json`, `python`, `tsx`. Always dark (for now), with optional copy available.

## Drawer wiring

`PromptDetailsPanel` composes the prompt picker (pill row hidden for single-prompt groups) and the Code tab. `Prompts.tsx` caches the last-shown group in `displayGroup` so the drawer body stays populated through the slide-out transition. Subheader copy uses ICU plurals so "Prompt(s) added from connected MCP server" agrees with group size and argument counts pluralize correctly across locales.

## i18n

Keys under `prompts.details.code.*` and `prompts.details.preview.*` across en-US, es-ES, pt-BR. Spanish and Portuguese translations included; native review pending before ship.

## Bundle

`prism-react-renderer ^2.4.1` adds about 85 KB raw and 57 KB gzipped to the vendor chunk.

## Production CSRF posture

Verified end-to-end. Login at `/app/auth/login` sets auth and CSRF cookies on path `/`. The api client reads the cookie on every mutating request and forwards it as `X-CSRF-Token`. Same-origin guaranteed by `vite.config.ts` shipping the bundle under `/static/app/`. 401 triggers redirect-to-login.

## Out of scope

- Plugin-pass count in the status row (pending a backend response field).
- Preview invocation tagging (header vs `/preview` sibling route).
- RBAC gate on the `preview` button (waiting on a `usePermission` hook).

## Related

- Ships initial scaffold for #5323.
- Wires into #5101 once cards land: swap the `Prompts.tsx` button for a per-card overflow menu, pass `prompts: group.prompts` and the clicked prompt as `initialPromptId`.


a-effort:
Thanks for rebasing the original PR, @gcgoncalves! Added some adjustments/improvements including changes related to prompts added via form (local) vs by connecting a MCP server. That closes https://github.com/IBM/mcp-context-forge/issues/5563. 

https://github.com/user-attachments/assets/97fcf11e-b175-403c-86c0-702886f622e6

FYI @marekdano @vishu-bh @jonpspri

client/src/components/prompts/PromptDetailsPanel.tsx
```diff
@@ -0,0 +1,285 @@
...
+                              className="rounded-full px-2 py-0 text-[11px] font-medium text-muted-foreground"
+                            >
+                              {label}
+                            </Badge>
+                          );
+                        })}
+                        {/* TODO: wire tag add-flow. Placeholder pending the
+                            per-prompt actions menu design (tracked with the
+                            same #5563 follow-up as the drawer variants). */}
+                        <button
```
marekdano: Can we replace the native `button` with the shadcn one `Button`?

marekdano (changes requested):
@a-effort - thanks for working on it!

I have a few findings:

 ## 🔴 Blocking / should-fix
  
1. Accessibility regression: `aria-hidden` without `inert` on the closed drawer `PromptDetailsPanel.tsx` (the `<aside role="region">`)

The panel claims to mirror `ToolDetailsPanel`, but it dropped the one thing that makes the pattern accessible. `ToolDetailsPanel.tsx:159-160` has **both**:

  ```tsx
  aria-hidden={!open}
  inert={!open}
  ```

`PromptDetailsPanel` has only `aria-hidden={!open}`. Because the drawer stays mounted when closed, all its focusable controls — close button, arg inputs, the four snippet tab triggers, copy buttons, the preview button, the pill tabs - remain in the tab order while inside an `aria-hidden="true"` container. That's the classic "focusable content inside `aria-hidden`" violation: keyboard users tab into invisible controls, and it's an axe-core failure. Adding `inert={!open}` fixes both the tab-order leak and matches the sibling component.

2. Partial ARIA tabs pattern on the prompt picker `PromptDetailsPanel.tsx` (pill row, `role="tablist"` block) The pill row uses `role="tablist"` + `role="tab"` + `aria-selected`, but there's no `role="tabpanel"`, no `aria-controls`/`id` wiring, no roving `tabindex`, and no arrow-key handling — so a screen reader announces "tab, 1 of N" and promises Arrow-key navigation that isn't implemented. It's also a raw `<button>` rather than the shadcn `Button` used elsewhere in this file.

  Recommended fix — switch to the shadcn `Button` and use a **toggle-button group** (`aria-pressed`) instead of the fake tab pattern. This is the honest semantics for "pick one of N, content updates in place," and it's the least code:

  ```tsx
  <div className="mt-8 flex flex-wrap gap-2" role="group" aria-label="Select prompt">
    {prompts.map((p) => {
      const isSelected = p.id === selected?.id;
      return (
        <Button
          key={p.id}
          type="button"
          variant={isSelected ? "secondary" : "outline"}
          size="sm"
          aria-pressed={isSelected}
          onClick={() => setSelectedId(p.id)}
          className={cn(
            "rounded-full font-mono text-[12px]",
            isSelected
              ? "border-transparent bg-muted text-foreground"
              : "text-muted-foreground",
          )}
        >
          {p.name}
        </Button>
      );
    })}
  </div>
  ```

`Button` merges `className` via `cn`, so the pill styling still applies over the variant, and `aria-pressed` gives correct AT semantics with zero extra JS. If you specifically want true tab behavior (Arrow-key nav, single tab stop), use a real primitive — the radix `Tabs` wrapper added in this PR (`ui/tabs.tsx`) — rather than hand-rolling `tabindex`/keydown.

The radix `PromptSnippetTabs` right below this row is a *real* tablist, so keeping a fake one directly above it is what makes the current markup especially confusing to AT.

## 🟡 Suggestions

3. Preview request has no abort / unmount guard `usePromptPreview.ts` (`run` callback)

`PromptCodeTab` is remounted via `key={selected.id}` on prompt switch, which unmounts this hook. If a preview is in flight when the user switches prompts (or closes the drawer), the resolved promise still calls `setResult`/`setError`/`setLoading` on an unmounted component → React state-update-after-unmount warning, and a wasted render. The api client already accepts a `signal` (`RequestOptions.signal`), so wiring an `AbortController` through `promptsApi.render` and aborting in the effect cleanup would close this cleanly.

4. Unused / inconsistent i18n keys
  - `prompts.details.code.args.empty` is defined in all three locales but never referenced (the args form returns `null` for zero args). Dead key.
  - `es-ES` and `pt-BR` define `args.optional` and `args.placeholder`; `en-US` does not, and none of the three are used by the shipped components. Either remove them or you'll have translators maintaining phantom strings.

5. `CodeBlock` duplicates the copy-button markup
`code-block.tsx` — the `copiedLabel ?` and `:` branches render an essentially identical `<Button variant="ghost" size="icon-xs" …>`; only the tooltip wrapper differs. Extract the button and wrap just the tooltip around it to remove the ~15-line duplication.

6. `CodeBlockLanguage` has a redundant member
  The union includes both `"tsx"` and `"typescript"`, both mapping to prism `tsx`, but only `"tsx"` is ever passed. Worth dropping `"typescript"` unless a caller needs it.

a-effort:
Thanks @marekdano, all six taken. Four commits on top of 3b1252d:

**🔴 Blocking**

1. [`05a9162`](https://github.com/IBM/mcp-context-forge/pull/5575/commits/05a9162f7) adds `inert={!open}` to the `<aside>` and a matching test assertion.
2. Same commit swaps the fake tablist for the shape you suggested: `role="group"` + shadcn `Button` with `aria-pressed`. Also converted the tag-add placeholder to shadcn `Button` for consistency.

**🟡 Suggestions**

3. [`56414e4`](https://github.com/IBM/mcp-context-forge/pull/5575/commits/56414e428) threads `AbortSignal` through `promptsApi.render`. `usePromptPreview` aborts on unmount and skips state updates when the signal is aborted, so late resolutions after a prompt switch or drawer close no longer touch the unmounted instance. New test covers unmount mid-flight.
4. [`7f942c5`](https://github.com/IBM/mcp-context-forge/pull/5575/commits/7f942c57f) removes `args.empty` from all three locales and `args.optional` / `args.placeholder` from es-ES and pt-BR.
5. [`eca0302`](https://github.com/IBM/mcp-context-forge/pull/5575/commits/eca030254) collapses `CodeBlock` to one button; tooltip wraps only when `copiedLabel` is set.
6. Same commit drops `"typescript"` from `CodeBlockLanguage`. Confirmed no caller passes it.

All 1880 client vitest tests pass; eslint clean. Re-requesting review.

a-effort:
Placeholder in the arg input showed a stray `)` for descriptions shaped like `Model ID (e.g. 'openai/gpt-oss-120b')`. The extractor in `toPlaceholder` matched from `e.g.` to end-of-string, which drops the opening `(` but keeps the trailing `)`.

Fix walks to the matching close-paren when `e.g.` sits inside a parenthetical, depth-aware so nested parens like `(e.g. 'kind(pod)')` don't truncate early. Two regression tests added covering the reported shape and the nested case.

marekdano (commented):
The PR looks good now!

LGTM 🚀

marekdano (approved):

client/src/hooks/usePromptPreview.ts
gcgoncalves, line 1: Should this file go on the hooks directory?

marekdano (approved):
The core preview feature is unchanged and all earlier fixes are intact, CI is green, so my approval still stands. Two small follow-ups worth folding in (both non-blocking):

**1. Move `usePromptPreview` into `client/src/hooks/`** 
@gcgoncalves, good catch. The convention across the app is that hooks live in `client/src/hooks/` (`useToolForm`, `useResourceForm`, `useTeamForm`, `usePromptForm`, `useQuery`, …); the only co-located exception is `auth/useAuth.ts`. For consistency and discoverability, `usePromptPreview.ts` (and its test) should move there, with the re-export in `components/prompts/index.ts` updated to point at the new path.

**2. `types/prompts.ts` - use a type-only import**

  ```ts
  // now
  import { PromptRead } from "@/generated/types";
  // suggested
  import type { PromptRead } from "@/generated/types";
  ```

`PromptRead` is only used as a type, and this matches how the rest of the PR imports generated types. Trivial, but keeps the import style uniform.

client/src/hooks/usePromptPreview.ts
a-effort: Good catch — moved in 51661ab03. Both the hook and its test now live under `client/src/hooks/`, and the 4 consumer imports (plus the barrel re-exports in `components/prompts/index.ts`) are updated to match.

a-effort:
Thanks for the review, @marekdano — both follow-ups folded in on `5448-preview-prompt-2`:

- **51661ab03** — move `usePromptPreview` (and its test) into `client/src/hooks/`, update the 4 consumer imports to `@/hooks/usePromptPreview`, drop the barrel re-exports in `components/prompts/index.ts` (no external consumer used them).
- **d2971b7d7** — tighten `types/prompts.ts:1` to `import type { PromptRead }`. All other consumers in the PR already use the type-only form, so this just aligns the last outlier.

Verified locally: `usePromptPreview.test.tsx`, `PromptPreviewResult.test.tsx`, `PromptCodeTab.test.tsx` all pass (17/17); no new typecheck or lint findings on the touched files.

marekdano (approved):
Ready to LGTM 🚀

Issue #5501, opened 2026-07-06
[FEATURE][AUTH]: per-user OAuth provider-token revocation endpoint (RFC 7009 provider-side revoke + local delete)


zhangyb:
### 🧭 Type of Feature

- [x] Enhancement to existing functionality
- [x] New feature or capability
- [ ] New MCP-compliant server
- [ ] New component or integration
- [ ] Developer tooling or test improvement
- [ ] Packaging, automation and deployment
- [ ] Other

---

### 🧭 Epic

**Title:** Per-user OAuth provider-token revocation endpoint (RFC 7009 provider-side revoke + local delete)

**Goal:** Give ContextForge a first-class, least-privilege HTTP endpoint that (1) revokes a single user's upstream OAuth **provider** token *at the provider* via [RFC 7009](https://datatracker.ietf.org/doc/html/rfc7009) (OAuth 2.0 Token Revocation), then (2) deletes the stored token row locally — so a "disconnect" / "revoke access" genuinely **invalidates** the credential rather than only forgetting the local copy.

**Why now / who benefits:** Today there is no HTTP endpoint to hard-delete a single user's stored provider token. A client/governance layer in front of ContextForge can stop *honoring* a grant at its own boundary, but it cannot ask ContextForge to *proactively destroy* the stored provider token — the encrypted token persists server-side until it lapses at expiry / refresh failure. Anyone building a "disconnect this integration" experience, and anyone with data-hygiene / right-to-be-forgotten / least-standing-credential requirements, needs this. Crucially, **deletion alone is not revocation**: an un-expired access/refresh token stays live at the provider after ContextForge forgets it, leaving a standing-access window that only provider-side revocation closes.

**Current gap (verified on `main`):** The internal delete capability already exists — `TokenStorageService.revoke_user_tokens(gateway_id, app_user_email)` (`mcpgateway/services/token_storage_service.py:524`) does a real `db.delete(token_record)` + commit — but it is **not wired to any route** (its only references are its own definition and two docstring examples). The OAuth router's only `DELETE` is `DELETE /oauth/registered-clients/{client_id}`, which removes a *shared* dynamic-client registration and would break OAuth for **every** user of that gateway — the wrong lever for revoking one user's access.

---

### 🧑🏻‍💻 User Story 1

**As a:** an end user who connected an OAuth integration
**I want:** to revoke and delete my own stored provider token on demand
**So that:** disconnecting actually invalidates my credential at the provider, not just locally

[...]

Issue #782, opened 2025-08-19
[FEATURE][AUTH]: OAuth enhancement following PR 768


crivetimihai:
# [FEATURE][AUTH]: OAuth Enhancement Following PR 768

## Goal

Complete the OAuth 2.0 implementation started in PR #768 by addressing remaining gaps: token management UI, OAuth coverage for resources/prompts, automated cleanup, and provider templates. Several items (refresh tokens, PKCE) are already implemented.

## Why Now?

1. **Production Readiness**: Organizations need complete OAuth lifecycle management
2. **User Experience**: Manual token management is error-prone
3. **Security Compliance**: Proper token rotation and cleanup required for audits
4. **Enterprise Adoption**: Large organizations require comprehensive OAuth support

---

## Status Summary

| Feature | Status |
|---------|--------|
| Refresh Token Handling | **IMPLEMENTED** in `token_storage_service.py` |
| PKCE Support | **IMPLEMENTED** with `code_verifier` in OAuth states |
| User ID Mapping | Partially implemented (placeholder extraction) |
| Token Management UI | NOT IMPLEMENTED |
| OAuth for Resources/Prompts | NOT IMPLEMENTED |
| Automated Cleanup | NOT IMPLEMENTED |
| Provider Templates | NOT IMPLEMENTED |

---

## 📖 User Stories

<details>
<summary><strong>US-1: Admin - View OAuth Authorizations</strong></summary>

**As an** administrator
**I want** to view all OAuth authorizations in the Admin UI
**So that** I can monitor and manage user access to OAuth-protected gateways

**Acceptance Criteria:**

```gherkin
Scenario: View OAuth tokens list
  Given I am logged in as an admin
  When I navigate to "OAuth Tokens" in the Admin UI
  Then I should see a list of all OAuth authorizations
  And each entry should show: user, gateway, scopes, expiration status

Scenario: View token details
  Given an OAuth token exists for user "alice@example.com"
  When I click on the token entry
  Then I should see: creation time, last used, scopes granted, expiration
  And I should NOT see the actual token value

Scenario: Revoke user token
  Given an OAuth token exists for user "alice@example.com"
  When I click "Revoke" on the token entry
  Then the token should be deleted
  And the user should need to re-authorize on next access
```

**Technical Requirements:**
- Add "OAuth Tokens" section to Admin UI
- Display token metadata (not actual tokens)
- Support single and bulk revocation
- Show token health status (valid/expired/expiring soon)

</details>

<details>
[...]

PR #5594, opened 2026-07-10
feat(oauth): add per-user OAuth connections view

### LINKED ISSUE #5592: [FEATURE][UI][AUTH]: Per-user "Connected Accounts" overview — OAuth gateways with my authorization status and one-click authorize
### Problem

With per-user OAuth now working well (authorization-code gateways + encrypted per-user token storage), the UX around *discovering and managing* those authorizations hasn't kept up. As a user I have no way to answer:

- **Which gateways can I authorize against?** OAuth-enabled gateways look like any other row in the Gateways panel; I have to know (or be told) which ones need a per-user grant and hunt for them one by one.
- **Which ones have I already authorized?** There is no per-user status anywhere. Even on a single gateway, the Authorize button stays identical before and after a successful grant (#3850), so the only way to "check" is to re-run the flow or ask an admin to inspect `oauth_tokens`.

We rolled per-user OAuth out to a multi-user deployment with 130+ registered gateways this week, and this was the very first piece of user feedback: *"There is no easy page where I can see all of the MCPs available to auth against, and whether I've already auth'ed or not."* At 130+ gateways the per-row hunt genuinely doesn't scale.

### Proposed feature: a per-user "Connected Accounts" view

A user-scoped page (Admin UI section or standalone route) that lists every OAuth authorization-code gateway visible to the current user, with:

| Column | Content |
|---|---|
| Gateway | name + description |
| My status | **Connected** (with token expiry / last refreshed) or **Not connected** |
| Actions | Authorize / Re-authorize; Disconnect once #5501 (RFC 7009 revocation) lands |

Backed by a small user-scoped API (e.g. `GET /oauth/connections`) returning the caller's per-gateway token status — deliberately scoped to the *current* user so it complements rather than duplicates the admin-wide token management UI proposed in #782.

### Related issues

- #782 — admin-scoped OAuth token management UI (this request is the per-user counterpart)
- #5501 — per-user provider-token revocation endpoint (the natural "Disconnect" action for this page)
- #3850 — Authorize button doesn't reflect authorized state (subset of this request)

### Environment

- v1.0.5, Helm (mcp-stack) on Kubernetes, Postgres + Redis
- SSO (Okta) end users, email-auth platform admin
- Gateways are MCP servers exposing their own OAuth AS metadata + DCR; per-user grants via `/oauth/authorize/{gateway_id}` work end-to-end

christophercolumbusdog:
# feat(oauth): add per-user OAuth connections view

Resolves #5592

## Summary

Adds a per-user "OAuth Connections" view in two layers:

1. **API** — `GET /oauth/connections` returns, for the authenticated user, every OAuth
   Authorization Code gateway they can see (owned, team-visible, or public; unrestricted
   admins see all), merged with the user's own stored token status:

   ```json
   {
     "connections": [
       {
         "gateway_id": "abc123",
         "name": "GitHub MCP",
         "url": "https://github-mcp.example.com/mcp",
         "description": "GitHub tools",
         "connected": true,
         "expires_at": "2026-07-11T00:00:00+00:00",
         "is_expired": false,
         "scopes": ["repo"],
         "updated_at": "2026-07-10T00:00:00+00:00",
         "authorize_url": "/oauth/authorize/abc123"
       }
     ]
   }
   ```

   Token material (access/refresh tokens) is never selected into results or returned.

2. **Admin UI** — a new "OAuth Connections" tab (HTMX partial, no new JS) listing those
   gateways with a Connected / Expired / Not connected status pill, expiry, scopes, and an
   Authorize / Re-authorize link per row. The tab is gated on `gateways.read` (validated by
   the startup `SECTION_PERMISSIONS` check) and is hideable via
   `MCPGATEWAY_UI_HIDE_SECTIONS=oauth-connections`.

## Motivation

Requested in #5592. Today a user who has authorized several OAuth gateways has no way to
see which gateways they are connected to, whether their tokens have expired, or where to
go to re-authorize — token state is only visible in the `oauth_tokens` table. This is the
per-user counterpart of the admin-level token visibility work discussed in #782, and the
natural future host for per-user disconnect/revocation (#5501 — intentionally out of scope
here; the docs still list revocation as a known gap).

## Implementation notes

- `TokenStorageService.list_user_token_info(app_user_email)` — one query over
  `oauth_tokens`, returning the same field shape as `get_token_info()` keyed by gateway id.
- `build_user_oauth_connections()` in `oauth_router.py` is shared by the API endpoint and
  the admin partial, so the two surfaces cannot drift.
- Gateway visibility mirrors the admin gateways-partial access conditions (owner / team /
  public via a single SQL query); `grant_type` is filtered in Python since `oauth_config`
  is a JSON column (portable across SQLite/Postgres, gateway counts are small).
- Team-scoped API tokens never widen visibility through admin status (mirrors
  `get_hidden_sections_for_user` semantics).

## Screenshots

Captured from a local run of this branch (SQLite, seeded with one connected, one expired,
and one not-connected Authorization Code gateway for the admin user):

**Light mode**

OAuth Connections tab, light mode

**Dark mode**

OAuth Connections tab, dark mode

On the same running instance, `GET /oauth/connections` (session auth) returned 200 with the
three entries in the documented shape — `connected=true, is_expired=false` (GitHub),
`connected=true, is_expired=true` (Atlassian), `connected=false` (GitLab) — and no token
material.

## Testing

New tests (direct-call style, matching siblings):

- `tests/unit/mcpgateway/routers/test_oauth_router.py::TestListOAuthConnections` — connected +
  not-connected merge with non-authorization-code gateways excluded, 401 without email,
  500 wrapping.
- `tests/unit/mcpgateway/test_admin.py` — partial renders via shared assembly; scoped-token
  admin does not bypass visibility.
- Doctest on `list_user_token_info` plus an end-to-end check of the query/merge against a
  real SQLite session (visibility filter, grant-type filter, token merge, root-path URL).

```
tests/unit/mcpgateway/routers/test_oauth_router.py .... 114 passed, 1 warning in 0.43s
  (-k Connections: 3 passed, 111 deselected)
tests/unit/mcpgateway/test_admin.py .................... 1193 passed (full file, exit 0)
  (-k oauth_connections: 2 passed, 1191 deselected)
tests/unit/mcpgateway/services/test_token_storage_service.py + --doctest-modules
  (token_storage_service.py, oauth_router.py) ......... 62 passed, 1 warning in 0.44s
tests/unit/mcpgateway/test_admin_menu_visibility.py
tests/unit/mcpgateway/test_admin_permission_helpers.py
tests/unit/mcpgateway/test_config.py ................... 210 passed, 1 warning in 2.33s
CI=1 validate_section_permissions(admin_router) ........ passed (17 mapped sections verified)
```

`ruff check` / `black --check` / `isort --check` on the changed files report only findings
that pre-exist on `main` (verified by diffing against a stash of the changes); all hunks
introduced here are clean.

## Note for maintainers

The commit contains both the API endpoint and the admin UI tab. If you'd prefer them split
into two commits (API first, UI second) for review or cherry-picking, I'm happy to rework
the branch.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1573,7 +1573,7 @@ VALIDATE_TOKEN_ENVIRONMENT=true
 # MCPGATEWAY_UI_EMBEDDED=false
 
 # Comma-separated list of UI sections to hide
-# Valid values: overview, servers, gateways, tools, prompts, resources, roots, mcp-registry, metrics, plugins, export-import, logs, version-info, maintenance, teams, users, agents, tokens, settings
+# Valid values: overview, servers, gateways, oauth-connections, tools, prompts, resources, roots, mcp-registry, metrics, plugins, export-import, logs, version-info, maintenance, teams, users, agents, tokens, settings
 # Example: MCPGATEWAY_UI_HIDE_SECTIONS=prompts,resources,teams
 # MCPGATEWAY_UI_HIDE_SECTIONS=
 
--- a/.env.example
+++ b/.env.example
@@ -1665,7 +1665,7 @@ VALIDATE_TOKEN_ENVIRONMENT=true
 # MCPGATEWAY_UI_EMBEDDED=false
 
 # Comma-separated list of UI sections to hide
-# Valid values: overview, servers, gateways, tools, prompts, resources, roots, mcp-registry, metrics, plugins, export-import, logs, version-info, maintenance, teams, users, agents, tokens, settings
+# Valid values: overview, servers, gateways, oauth-connections, tools, prompts, resources, roots, mcp-registry, metrics, plugins, export-import, logs, version-info, maintenance, teams, users, agents, tokens, settings
 # Example: MCPGATEWAY_UI_HIDE_SECTIONS=prompts,resources,teams
 # MCPGATEWAY_UI_HIDE_SECTIONS=
 
--- a/docs/docs/architecture/oauth-authorization-code-ui-design.md
+++ b/docs/docs/architecture/oauth-authorization-code-ui-design.md
@@ -27,12 +27,13 @@ This document describes how the Admin UI initiates OAuth 2.0 Authorization Code 
 - Tokens are stored per gateway and app user (email) in `oauth_tokens`, encrypted with `AUTH_ENCRYPTION_SECRET`.
 - Refresh tokens are used when access tokens are near expiry; invalid refresh tokens are cleared.
 - DCR auto-registration can run during `/oauth/authorize/{gateway_id}` when `issuer` is set and `client_id` is missing.
+- Per-user token listing: the admin UI "OAuth Connections" tab and `GET /oauth/connections` show each user's own connection status (connected, expiry, scopes) per Authorization Code gateway, without exposing token material.
 
 ### Known gaps and constraints
 
 - `oauth_store_tokens` and `oauth_auto_refresh` checkboxes exist in the UI but are not persisted or enforced by the backend.
 - `OAUTH_PREFERRED_CODE_CHALLENGE_METHOD` is defined but PKCE is always S256 today.
-- No admin UI to list or revoke stored OAuth tokens per user.
+- No admin UI to revoke stored OAuth tokens per user (listing exists via the "OAuth Connections" tab; revocation is tracked in #5501).
 - Token cleanup is a helper method only; there is no scheduler invoking it.
 
 ## Architecture Overview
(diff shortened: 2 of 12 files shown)


christophercolumbusdog:
Rebased onto current main and force-pushed. The conflict is cleared and the PR shows mergeable again.

Two things from the rebase worth flagging for whoever picks this up.

Main removed `_extract_user_email` from `oauth_router.py` and standardized on `get_user_email()` from `auth_context.py`. The merge was clean textually, so the endpoint was left calling a function that no longer exists. I've switched it to the same guard pattern `_enforce_gateway_access` already uses, since `get_user_email()` returns the string `"unknown"` instead of `None` when there's no email, and a straight swap would have walked past the 401 check.

There's also an open question on email case that predates this branch. The endpoint lowercases before looking up stored tokens, but `initiate_authorization_code_flow` stores `app_user_email` without normalizing, and the admin partial reads it back without normalizing either. So a user whose token carries mixed case could get "not connected" from `GET /oauth/connections` while the admin tab shows them connected. I'm happy to drop the `.strip().lower()` to match the rest of the codebase, or normalize on the write path instead, whichever fits better with where you want that helper to go.

No rush, but this has been sitting a few weeks. Any chance of a look?

PR #5596, opened 2026-07-11
fix: remove premature exit in Containerfile and restore linting targets that run on Go

jonpspri:
## Summary

Two fixes for build and linting infrastructure:

### Containerfile
Remove a stray `exit 0` inside the `ENABLE_RUST=false` fallback block. It currently only triggers a buildkit warning but would become a latent build-breaker if the block structure changes (the trailing backslash on the prior line means the `exit 0` is in scope for the whole `RUN` step).

### Makefile
Go-based linting targets (`actionlint`, `reviewdog`, `chart-testing`, `kube-linter`) relied on the host Go environment having writable default paths. This fails under restricted configurations (e.g., `GONOSUMDB`-only envs, read-only `$HOME`, or Go toolchain version mismatches). The fix sets explicit `GOPATH`, `GOMODCACHE`, `GOCACHE`, and `GOBIN` under a dedicated temp root so these targets work regardless of host Go configuration.

## Changes
- **Containerfile**: Remove `exit 0` statement from the ENABLE_RUST=false fallback branch
- **Makefile**: Add `LINT_GO_ROOT` and `LINT_GO_TOOLCHAIN` variables; wrap all `go run` / `go install` linting targets with explicit Go environment setup
- **.secrets.baseline**: Auto-updated line numbers (no new secrets)

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-07-14T13:47:00Z",
+  "generated_at": "2026-07-13T13:56:20Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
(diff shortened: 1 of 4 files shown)


Lang-Akshay (changes requested):
Thanks for the PR @jonpspri 
```
Makefile:3600 — GOTOOLCHAIN=go1.26.4 forces exactly that toolchain. Hosts with a newer Go version must download 1.26.4, and future
    kube-linter@latest releases requiring newer Go will fail. Use go1.26.4+auto or omit the override.
```

jonpspri:
Removed LINT_GO_TOOLCHAIN.  Now rebasing...

Lang-Akshay (dismissed):
LGTM

ja8zyjits (dismissed):
LGTM

Lang-Akshay (approved):
LGTM

PR #5603, opened 2026-07-13
test(protocol): remove xfail for GAP-001/GAP-002 — log and progress notifications now relayed

jonpspri:
## Summary

The POST-correlated notification relay now works end-to-end on gateway targets, closing the two long-standing compliance gaps for request-tied notifications.

## Changes

- **Remove ** from  and  on  and  targets
- **Move GAP-001 and GAP-002** to Closed gaps in `COMPLIANCE_GAPS.md`
- **Update GAP-011** cross-reference to remove the now-closed gaps from its server→client gap enumeration
- Drop unused `request` parameter from both test signatures

## Gap Details

- **GAP-001** (Server-initiated log notifications): `notifications/message` emitted during a tool call now rides the POST-correlated stream through the gateway to the downstream client.
- **GAP-002** (Progress notifications): `notifications/progress` scoped to a `progressToken` now delivers on the POST-correlated stream of the originating request.

Both gaps shared the same root cause — the gateway was not relaying notifications on the POST-correlated stream.

## Notes

- `test_logging_set_level_filters_below_threshold` retains its `xfail_on` (separate test with additional `setLevel` filtering assertions).
- Tests co-blocked on GAP-001/002 in `test_notifications.py` remain xfailed under their primary blocker (GAP-008 — tool federation).

### DIFF
--- a/tests/live_gateway/protocol_compliance/COMPLIANCE_GAPS.md
+++ b/tests/live_gateway/protocol_compliance/COMPLIANCE_GAPS.md
@@ -84,52 +84,6 @@ the marker and the entry per the rules above.
 
 ---
 
-### GAP-001 — Server-initiated log notifications not delivered
-
-| | |
-|---|---|
-| **Targets affected** | `gateway_proxy`, `gateway_virtual` |
-| **Tests** | `test_logging.py::test_log_message_reaches_client`; secondary blocker for `test_notifications.py::test_tools_list_changed_notification_delivered` and `::test_resources_list_changed_notification_delivered` (primary GAP-008) |
-| **Spec** | [MCP 2025-11-25 — server `logging` capability](https://modelcontextprotocol.io/specification/2025-11-25/server/utilities/logging) |
-
-**Observed**: when an upstream tool calls `ctx.log(...)` during a tool
-invocation, the gateway accepts the upstream's `notifications/message`
-but does not relay it to the downstream client. The client's
-`log_handler` is never invoked.
-
-**Expected**: the spec requires the server to emit `notifications/message`
-to clients that subscribed via `logging/setLevel`. Federation should
-forward upstream-emitted log messages.
-
-**Why**: the log emitted during a tool call is a request-tied
-notification — it SHOULD ride the POST-correlated stream for that call.
-The gateway isn't relaying notifications on that stream. (Logs emitted
-*between* calls — e.g. in response to a `logging/setLevel` change —
-would ride the standalone stream, which #4205 also closes, but the
-test exercises the in-call path.)
-
----
-
-### GAP-002 — Progress notifications not delivered
-
-| | |
-|---|---|
-| **Targets affected** | `gateway_proxy`, `gateway_virtual` |
-| **Tests** | `test_utilities.py::test_progress_notifications_delivered`; secondary blocker for `test_notifications.py::test_tools_list_changed_notification_delivered` and `::test_resources_list_changed_notification_delivered` (primary GAP-008) |
-| **Spec** | [MCP 2025-11-25 — `progress` notifications](https://modelcontextprotocol.io/specification/2025-11-25/basic/utilities/progress) |
-
-**Observed**: a tool calling `ctx.report_progress(...)` returns successfully,
-but the client's `progress_handler` is never invoked.
-
-**Expected**: with a `progressToken` on the request, the server must emit
-`notifications/progress` events the client can observe.
-
-**Why**: progress notifications are request-tied (the `progressToken`
-is scoped to one client request) and ride the POST-correlated stream
-of that request. Root cause is the gateway not relaying notifications
-on the POST-correlated stream — **not** #4205, which only closes the
-standalone GET stream that progress wouldn't have used anyway.
-
 ---
 
 ### GAP-003 — Client roots not forwarded to upstream
@@ -363,7 +317,7 @@ session that opened the subscription.
 not tied to any specific client request — per spec it rides the
 standalone stream (`GET /mcp/`). The gateway currently closes that
 stream with 405 (see #4205), so the notification has no channel to
-the client. Unlike the other server→client gaps (GAP-001/002/003/004/
+the client. Unlike the other server→client gaps (GAP-003/004/
 005), this one is genuinely about the standalone stream and #4205 is
 the correct blocker.
 
@@ -438,6 +392,18 @@ on this test.
 ---
 
 ## Closed gaps
+
+### GAP-001 — Server-initiated log notifications not delivered *(closed 2026-07-13)*
+
+**Was**: `test_logging.py::test_log_message_reaches_client` xfailed on `gateway_proxy` and `gateway_virtual` because the gateway did not relay `notifications/message` on the POST-correlated stream.
+
+**How closed**: Tests now pass on both gateway targets — the POST-correlated notification relay was fixed. `xfail_on` removed from `test_log_message_reaches_client`. Note: `test_logging_set_level_filters_below_threshold` retains its `xfail_on` (separate test with additional setLevel filtering assertions).
+
+### GAP-002 — Progress notifications not delivered *(closed 2026-07-13)*
+
+**Was**: `test_utilities.py::test_progress_notifications_delivered` xfailed on `gateway_proxy` and `gateway_virtual` because the gateway did not relay `notifications/progress` on the POST-correlated stream.
+
+**How closed**: Tests now pass on both gateway targets — same POST-correlated notification relay fix as GAP-001. `xfail_on` removed from `test_progress_notifications_delivered`.
 
 ### GAP-007 — `tools/list` pagination cap below upstream tool count *(closed 2026-04-18)*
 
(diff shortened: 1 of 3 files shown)


msureshkumar88 (changes requested):
Thanks for closing out GAP-001/GAP-002 — the doc cleanup and fixture removal look solid and well-scoped.

One thing I couldn't verify before this can merge: the PR description says "the POST-correlated notification relay fix closes GAP-001 (server-initiated log notifications) and GAP-002 (progress notifications)," but I don't see that relay fix in this diff or anywhere in `main`'s history (checked `notification_service.py`, `streamablehttp_transport.py`, `session_affinity.py`). `_forward_notification_to_stream` only publishes to the GET `/mcp` event bus (ADR-052 standalone stream), and I couldn't find any call site that relays an upstream notification onto the *current request's* POST-correlated stream.

Also worth flagging: `tests/live_gateway/protocol_compliance/` isn't wired into any CI workflow (it's manual-only via `make test-protocol-compliance-gateway`), so green CI here doesn't confirm `test_log_message_reaches_client` / `test_progress_notifications_delivered` actually pass on `gateway_proxy`/`gateway_virtual`.

Could you point to the commit that lands the relay fix (if it's elsewhere), or share local output from `make test-protocol-compliance-gateway` showing both tests passing? Want to make sure we're not just removing the safety net rather than closing the gap.

jonpspri:
Nope.  I cannot point to the PR... it just started working while I was operating on a differnt PR.  I suspect it's one of the session-related fixes that have gone in recently, but I haven't had time to research.

No, these tests are not currently in CI/CD.  I'd like to add them to the CD process as it matures, though they are for a protocol that will soon be deprecated.

"Want to make sure we're not just removing the safety net rather than closing the gap."  We're actually adding a safety net by removing XFAIL flags now that the tool is working to protocol.

-- no tokens were burned in the creation of this reply --

msureshkumar88 (changes requested):
Thanks for following up, Jonathan — and appreciate the honesty that you couldn't point to the landing commit. I went ahead and ran the harness locally so we'd have data rather than guesses, and unfortunately it doesn't yet support closing these gaps.

## What I ran

- Booted a clean gateway from this branch (`jps-remove-protocol-xfails`, default Python runtime), fresh SQLite DB, JWT secret matching the harness default, `SSRF_ALLOW_LOCALHOST=true` so the auto-spawned reference upstream could register.
- Let the harness fixtures spawn `compliance-reference-server`, register it via `POST /gateways`, and build the virtual server.
- Ran the two de-xfailed tests across both gateway targets:

```
pytest test_logging.py::test_log_message_reaches_client \
       test_utilities.py::test_progress_notifications_delivered \
       -k "gateway_proxy or gateway_virtual"
```

## Result — all 4 fail

```
FAILED test_log_message_reaches_client[gateway_proxy-http]      — AssertionError: log data missing in []
FAILED test_log_message_reaches_client[gateway_virtual-http]    — AssertionError: log data missing in []
FAILED test_progress_notifications_delivered[gateway_proxy-http]   — expected >=3 progress events, got []
FAILED test_progress_notifications_delivered[gateway_virtual-http] — expected >=3 progress events, got []
4 failed, 2 deselected
```

Federation itself was healthy — `tools/call` returned `200` and the reference tools were reachable — but no `notifications/message` or `notifications/progress` reached the client. That's exactly the GAP-001 / GAP-002 behaviour the `xfail_on` markers were guarding.

## Where this leaves the PR

- **Blocking:** removing `xfail_on` is unconditional for both gateway targets, but on a default gateway the tests genuinely fail. So this converts documented xfails into hard failures without the relay fix present in the tree — the earlier "removing the safety net vs. closing the gap" concern is now confirmed empirically, not just unverified.

## A couple of paths forward (whichever fits)

1. If the relay actually works under a specific runtime (e.g. `RUST_MCP_MODE=edge`/`full`), could you confirm the mode you saw green in? If it's Rust-path-only, the cleaner move is to **gate** the `xfail_on` on runtime mode rather than delete it, so the Python path stays honestly marked.
2. If there's a landing commit somewhere (session/notification work), linking it here lets us re-run and confirm before merging.
3. Otherwise, holding the xfail removal until the POST-correlated relay fix lands keeps the gap tracking accurate.

Happy to re-run the harness against any branch/commit or runtime mode you point me at — just say the word. Setup notes for reproduction are above.

jonpspri:
I use 'make testing-up test-protocol-compliance' for running the tests in question.  Not sure yet why the different results with your approach.

msureshkumar88 (approved):
You were right, Jonathan — and thanks for the nudge to run it your way. My earlier failing run was on a non-representative gateway (a bare single-process uvicorn on SQLite with no Redis), which is *not* what this harness targets. Re-ran against the real `make testing-up` stack and both gaps are genuinely closed.

## Reproduction on the redis-backed stack

Brought the testing stack back to health (its `redis` and `postgres` containers had exited, which is why my first attempt against `:8080` 502'd), then ran the two de-xfailed tests against the nginx-fronted gateway cluster:

```
MCP_CLI_BASE_URL=http://127.0.0.1:8080 \
JWT_SECRET_KEY=<stack secret> \
pytest test_logging.py::test_log_message_reaches_client \
       test_utilities.py::test_progress_notifications_delivered \
       -k "gateway_proxy or gateway_virtual"

4 passed, 2 deselected in 34.40s
```

`notifications/message` and `notifications/progress` both reached the client on `gateway_proxy` and `gateway_virtual`. That matches the XPASS you saw on `main` — removing the `xfail_on` markers is honest, and keeping them now would be the thing masking real behavior.

## One non-blocking observation

The relay appears to depend on the Redis-backed, multi-worker (gunicorn) deployment — on a single-process no-Redis gateway the notifications didn't fan out. Not a concern for this PR (the compliance harness always runs against the full stack), but if the underlying fix is indeed the recent session/pub-sub work, a one-line note in the GAP-001/002 "How closed" entries pointing at that mechanism would help the next person who tries to reproduce on a stripped-down gateway. Totally optional.

Approving. Sorry for the detour — better to have the data.

msureshkumar88:
### Why the first run failed, and what actually reproduces — for the record

Documenting both runs so the next person doesn't chase the same ghost. The two runs differed only in **deployment topology**, and that topology is what closes GAP-001/002.

#### ❌ Run 1 — failed (non-representative gateway)

Config: a single hand-rolled gateway process, SQLite, **no Redis**, session-affinity disabled.

```bash
# gateway
JWT_SECRET_KEY=my-test-key-but-now-longer-than-32-bytes \
PLATFORM_ADMIN_EMAIL=admin@example.com \
SSRF_ALLOW_LOCALHOST=true \
DATABASE_URL="sqlite:///./compliance_gw.db" \
python -m uvicorn mcpgateway.main:app --host 127.0.0.1 --port 8099

# tests
MCP_CLI_BASE_URL=http://127.0.0.1:8099 \
JWT_SECRET_KEY=my-test-key-but-now-longer-than-32-bytes \
MCP_REFERENCE_UPSTREAM_HOST=127.0.0.1 \
pytest test_logging.py::test_log_message_reaches_client \
       test_utilities.py::test_progress_notifications_delivered \
       -k "gateway_proxy or gateway_virtual"
# -> 4 failed  (log data missing in []; progress events got [])
```

Federation was fine (`tools/call` returned 200) — but no notifications reached the client. **Root cause: a single-process, no-Redis gateway does not fan out the POST-correlated notifications.** This is not a config the compliance harness ever runs against, so the failure was an artifact of my setup, not a real gap.

#### ✅ Run 2 — passed (the real `make testing-up` stack)

Config: nginx + 3 gunicorn gateway replicas, Postgres/pgbouncer, **Redis** (`CACHE_TYPE=redis`, `REDIS_URL=redis://redis:6379/0`), which is exactly what `make testing-up` brings up.

Gotcha that cost time: the running stack's `redis` and `postgres` containers had exited, so `:8080` returned 502/504 and every gateway row skipped/failed. Recovering them first was the fix:

```bash
docker start mcp-context-forge-redis-1 mcp-context-forge-postgres-1
docker restart mcp-context-forge-gateway-{1,2,3}
curl -s http://127.0.0.1:8080/health   # -> 200 once healthy
```

Then, using the **stack's** JWT secret (mismatched secret = `401 Invalid authentication credentials` on upstream registration):

```bash
MCP_CLI_BASE_URL=http://127.0.0.1:8080 \
JWT_SECRET_KEY=<value from gateway container env> \
pytest test_logging.py::test_log_message_reaches_client \
       test_utilities.py::test_progress_notifications_delivered \
       -k "gateway_proxy or gateway_virtual"
# -> 4 passed in 34.40s
```

`notifications/message` and `notifications/progress` both delivered on `gateway_proxy` and `gateway_virtual` — the XPASS-on-main behavior.

#### Takeaways for reproducers

- Always run this harness against `make testing-up` (Redis + multi-worker gunicorn). A bare single-process/SQLite/no-Redis gateway will **not** relay these notifications.
- Make sure the stack's `redis`/`postgres` are actually up (`docker ps`), not just the gateway containers.
- Match `JWT_SECRET_KEY` to the running gateway's env, or upstream registration 401s and every gateway row skips.

Issue #5142, opened 2026-06-09
[UI-REWRITE]: Create a new /v1/tools/generate-schemas-from-openapi endpoint


marekdano:
## Summary

  - Adds `POST /v1/tools/generate-schema-from-openapi` to the versioned REST API, mirroring the existing `POST /admin/tools/generate-schemas-from-openapi` endpoint
  - Allows API consumers and integrations to generate MCP tool input/output schemas from an OpenAPI spec without requiring Admin UI access
  - Delegates to the existing `openapi_service.fetch_and_extract_schemas()` — no new business logic

  ## Changes

  - **New:** `mcpgateway/routers/openapi_schema_router.py` — `APIRouter` with prefix `/v1/tools`; `POST /generate-schema-from-openapi`; requires `tools.create` RBAC permission with no admin bypass; no admin CSRF dependency
  - **Modified:** `mcpgateway/main.py` — registers the new router
  - **New:** `tests/unit/mcpgateway/routers/test_openapi_schema_router.py` — happy path, input validation, error mapping, permission denial

  ## Request / Response

  ```http
  POST /v1/tools/generate-schema-from-openapi
  Content-Type: application/json

  {
    "url": "https://api.example.com/v1/calculate",
    "request_type": "POST",
    "openapi_url": ""
  }

  {
    "message": "Schemas generated successfully",
    "success": true,
    "input_schema": { ... },
    "output_schema": { ... },
    "spec_url": "https://api.example.com/openapi.json"
  }
```

  request_type defaults to GET if omitted. openapi_url is auto-discovered if empty.

  ## Differences from the admin endpoint

  | | `/admin/tools/generate-schemas-from-openapi` | `/v1/tools/generate-schema-from-openapi` |
  |:--------|:---------------------------------------------|:-----------------------------------------|
  | **Router** | `admin_router` (`admin.py`) | New router in `mcpgateway/routers/` |
  | **CSRF** | `enforce_admin_csrf` | None |
  | **RBAC** | `tools.create`, no admin bypass | Same |
  | **Service** | `openapi_service` | Same |

Issue #4860, opened 2026-05-21
[UI-REWRITE]: Create REST tool


a-effort:
### Type of Feature

- [x] Enhancement to existing functionality (UI-REWRITE parity with legacy admin.html)

### Background

[Design](https://www.figma.com/design/H2Dr9QPvRVxCe92V8URvGT/Refiring?m=auto&node-id=3195-44205&t=WlB7QagnmKtLjeVv-1)
Refactor the legacy REST tool creation form (`mcpgateway/templates/admin.html:3920+`) that calls `POST /tools`.
This is also a prerequisite for any future work on the OpenAPI schema-generation UX (separate issue). This should be triggered on click `Add tools` from the Tools page.

### Scope

**API client**: add `client/src/api/tools.ts` with a typed `createTool(payload)` wrapping `POST /tools`. Mirror the shape of `client/src/api/servers.ts`.
**Form component**: match the pattern used by Create MCP server (#4175). {add link to design}
**Fields**: based on the `ToolCreate` schema at `mcpgateway/schemas.py:547-603`:
   - `request_type` (tabs: GET / POST / PUT / DELETE / PATCH — default POST for REST)
   -  `name` (text, validated client-side against `SecurityValidator` rules)
   - `url` (text, URL validation)
   - `description` (textarea)

   **Schemas:**
   - [generate a schema](https://github.com/IBM/mcp-context-forge/issues/4861) on button click (default)
   - or, manually add an `input_schema` in a JSON textarea  
   - optional `output_schema` (JSON textarea)

   **Auth:**
   - `auth` block — type selector (None / Basic / Bearer / Custom Headers / OAuth) and conditional fields

   **Optional / metadata:**
   - `headers` (JSON textarea)
   - `tags` (chip input)
   - `annotations` (read-only hint toggles: `readOnlyHint`, `destructiveHint`, etc.)
   - `visibility` (public / team / private)
   - `jsonpath_filter`
   - `team_id` (if non-admin user has multiple teams)

**Wire the entry point**: `AddToolsCard.onAddServer` opens the form. 
**Error handling**: surface 4xx error messages from `POST /tools` inline (name conflicts, malicious-pattern rejections, URL validation failures). The backend returns structured error bodies via the `ApiError` class.

**Permissions**: gated by `tools.create`. The existing 401/403 handling in `client/src/api/client.ts` covers redirect/error paths; no extra work needed.

### Out of scope

[...]

PR #5604, opened 2026-07-13
feat(ui-rewrite): generate tool schemas from OpenAPI in Create and Edit REST Tool form

### LINKED ISSUE #4861: [FEATURE][UI-REWRITE]: One-click schema generation from OpenAPI in Create REST tool form
### Type of Feature

### Background

The backend already exposes `POST /admin/tools/generate-schemas-from-openapi` (`mcpgateway/admin.py:11799`), which:

1. Accepts a tool `url`, `request_type`, and optional `openapi_url`
2. Fetches the OpenAPI 3.x spec (defaulting to `{scheme}://{host}/openapi.json`)
3. Extracts `input_schema` and `output_schema` for the requested path + method (resolving `$ref`s in `components.schemas`)
4. Returns both schemas plus the `spec_url` that was used

The endpoint requires `tools.create` permission (no admin bypass) and has SSRF protection built into `mcpgateway/services/openapi_service.py`.

**No UI surface currently exists for this feature.** The legacy admin.html doesn't expose it either, so this would be the first user-facing entry point for an otherwise-unused backend capability.

### Proposal

Add a "Generate from OpenAPI" button option as a default in the Create REST Tool form. Users will continue to be able to add schemas manually, if preferred / necessary.

### UX

[Design](https://www.figma.com/design/H2Dr9QPvRVxCe92V8URvGT/Refiring?m=auto&node-id=3195-44205&t=WlB7QagnmKtLjeVv-1)
1. **Button**: Disabled until `url` and `request_type` are filled (both are required by the backend endpoint).
2. **On click**: call the endpoint, show a loading state on the button (the request can take up to 10s).
3. **On success**:
   - Pretty-print the returned `input_schema` and `output_schema` JSON into the corresponding fields
   - Show a small caption: `Generated from {spec_url}` (the endpoint returns this)
4. **On failure**: map the typed error responses to clear messages:
   - `400` with SSRF-related message → "URL not reachable from the gateway (blocked by SSRF protection)"
   - `404` → "The path `{path}` with method `{method}` is not in the OpenAPI spec"
   - `502` → "Couldn't fetch the OpenAPI spec — check that the URL is correct and reachable"
   - `500` → generic fallback
5. **Don't overwrite silently**: if either schema field is already non-empty, ask the user to confirm before replacing.
6.  Include the manual ways to add an input and optional output schemas that the user could use if they prefer / can't use the default method of generating the schema from `/admin/tools/generate-schemas-from-openapi`

### API client work

Extend `client/src/api/tools.ts` (created by the prerequisite issue) with:

```ts
async function generateSchemasFromOpenapi(input: {
  url: string;
  request_type: string;
  openapi_url?: string;
}): Promise<{
  message: string;
  success: boolean;
  input_schema: object | null;
  output_schema: object | null;
  spec_url: string;
}>;
```

### Out of scope

- Bulk import of multiple endpoints from a single OpenAPI spec: the backend only extracts one path+method per call. A multi-endpoint flow would be a separate feature.
- Support for Swagger 2.0 specs the resolver doesn't currently handle, or YAML specs (backend uses `orjson.loads`).
- Inferring `name`, `description`, `tags` from the OpenAPI `operationId` / `summary` / `tags`. This could be a nice follow-up but adds complexity.

### Dependencies

- Blocked by: #4860 ([UI-REWRITE]: Create REST tool) 

### References

- Endpoint: `mcpgateway/admin.py:11799` (`POST /admin/tools/generate-schemas-from-openapi`)
- Service: `mcpgateway/services/openapi_service.py:162` (`fetch_and_extract_schemas`)
- Backend tests: `tests/unit/mcpgateway/test_admin_openapi.py`
- Parent epic: #4093

marekdano:
### Summary

Adds one-click input/output schema generation from an OpenAPI spec to the Create REST Tool form. Users can auto-populate the schema fields from a tool's OpenAPI document, regenerate, or fall back to manual entry. Wires the UI to the `POST /v1/tools/generate-schemas-from-openapi` endpoint.

Closes #4861, #4860. Uses the endpoint added in #5142.

https://github.com/user-attachments/assets/cf4fa895-f5b7-4f41-a28c-45939edda8f2

### The Generate button

- **Visibility**: shown for REST tools only (`integrationType === "REST"`) — hidden for MCP tools, since generation reads an OpenAPI spec. This form only ever creates REST tools, so in add mode it's always shown; in edit mode it's shown for REST tools and hidden for MCP.
- **Disabled** until a URL is provided (also guarded by request type, which is always set), matching the endpoint's required inputs.
- **Loading state**: while the request runs (up to ~10s) the button shows a spinner and a "Generating…" label and is disabled to prevent duplicate calls.
- **Label change**: after a successful generation it switches from **Generate** to **Regenerate**.
- **Overwrite guard**: clicking it when either schema field already has content prompts a confirmation before replacing, so manual edits aren't clobbered silently.

### What's included

**API client** (`client/src/api/tools.ts`)
- `toolsApi.generateSchemasFromOpenapi()` calling `POST /v1/tools/generate-schemas-from-openapi`. The request type is re-exported from the generated `GenerateSchemaRequest`; the response is narrowed by hand (the endpoint returns an untyped `JSONResponse`).

**New hook** (`client/src/hooks/useGenerateSchemaFromOpenapi.ts`)
- Encapsulates the async flow: `http(s)`-only URL guarding, the API call, and mapping the backend's typed HTTP statuses (400 SSRF / 404 path+method / 502 fetch / 500) to clear, localized messages. A 502 that wraps an upstream 401/403 is surfaced as an auth-specific message. Reports outcomes via `onSuccess` / `onError` / `onRequiresAuth` callbacks so the form remains the single owner of the editable schema fields.

**Form integration** (`useToolForm.ts`, `ToolForm.tsx`)
- Pretty-prints returned schemas into the fields and shows a `Generated from {spec_url}` caption.
- On failure, reveals a direct OpenAPI spec-URL fallback input; errors are announced via `role="alert"`.
- In edit mode the schema fields are always visible and "Add manually" is hidden (add-mode affordance only).

**i18n**
- New keys added to `en-US`, `es-ES`, and `pt-BR`.

### DIFF
--- a/client/src/api/tools.test.ts
+++ b/client/src/api/tools.test.ts
@@ -185,4 +185,53 @@ describe("toolsApi", () => {
       await expect(toolsApi.deactivate("tool-abc-123")).rejects.toThrow("HTTP 403");
     });
   });
+
+  describe("generateSchemasFromOpenapi", () => {
+    it("POSTs to /v1/tools/generate-schemas-from-openapi and returns the schemas", async () => {
+      const body = {
+        message: "Schemas generated successfully",
+        success: true,
+        input_schema: { type: "object" },
+        output_schema: null,
+        spec_url: "https://api.example.com/openapi.json",
+      };
+      mockFetch.mockResolvedValueOnce(
+        new Response(JSON.stringify(body), {
+          status: 200,
+          headers: { "Content-Type": "application/json" },
+        }),
+      );
+
+      const result = await toolsApi.generateSchemasFromOpenapi({
+        url: "https://api.example.com/v1/calculate",
+        request_type: "POST",
+      });
+
+      expect(mockFetch).toHaveBeenCalledWith(
+        expect.stringContaining("/v1/tools/generate-schemas-from-openapi"),
+        expect.objectContaining({
+          method: "POST",
+          headers: expect.objectContaining({ "X-CSRF-Token": "test-csrf-token" }),
+          credentials: "same-origin", // pragma: allowlist secret
+        }),
+      );
+      expect(result).toEqual(body);
+    });
+
+    it("throws ApiError carrying the backend status on failure", async () => {
+      mockFetch.mockResolvedValueOnce(
+        new Response(JSON.stringify({ success: false, message: "blocked" }), {
+          status: 400,
+          headers: { "Content-Type": "application/json" },
+        }),
+      );
+
+      await expect(
+        toolsApi.generateSchemasFromOpenapi({
+          url: "https://api.example.com/v1/calculate",
+          request_type: "POST",
+        }),
+      ).rejects.toThrow("HTTP 400");
+    });
+  });
 });
--- a/client/src/api/tools.ts
+++ b/client/src/api/tools.ts
@@ -4,6 +4,32 @@
 
 import { api } from "./client";
 import type { Tool } from "@/types/tool";
+import type { GenerateSchemaRequest } from "@/generated/types/generateSchemaRequest";
+
+/**
+ * Request body for {@link toolsApi.generateSchemasFromOpenapi}.
+ *
+ * Re-exported from the generated OpenAPI client so the shape stays in lockstep
+ * with the backend contract (`url`, optional `request_type`, optional
+ * `openapi_url`).
+ */
+export type GenerateSchemasFromOpenapiInput = GenerateSchemaRequest;
+
+/**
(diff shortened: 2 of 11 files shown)


gcgoncalves (approved):
High quality. 👍

PR #5518, opened 2026-07-06
feat: add global header search

vishu-bh:
# Pull Request

## Related Issue

Closes #5108

---

## Summary

Adds global search behavior to the React header quick nav using `/admin/search`.

- Adds a typed client API wrapper for unified admin search.
- Renders grouped search results in the header popover with team scope and admin-only user search support.
- Navigates selected results to the owning React page with `selected` and `search` query params.
- Adds localized search states for loading, empty, error, and minimum query length.
- Covers the flow with focused unit tests and a mocked Playwright e2e test.

---

## Reviewability

- [x] This PR has one clear purpose
- [ ] The linked issue is not labeled `triage`
- [x] Unrelated bugs or improvements are tracked in separate issues/PRs
- [x] Tests are included with the code they validate

---

## Type of Change

- [ ] Bug fix
- [x] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Chore (deps, CI, tooling)
- [ ] Other (describe below)

---

## Verification

| Check | Command | Status |
| --- | --- | --- |
| Formatting | `npx prettier --check src/api/search.ts src/api/search.test.ts src/components/layout/HeaderQuickNav.tsx src/components/layout/HeaderQuickNav.test.tsx e2e/global-search.spec.ts` | Pass |
| Lint | `npx eslint src/components/layout/HeaderQuickNav.tsx src/components/layout/HeaderQuickNav.test.tsx src/api/search.ts src/api/search.test.ts e2e/global-search.spec.ts` | Pass |
| Unit tests | `npx vitest run src/api/search.test.ts src/components/layout/HeaderQuickNav.test.tsx` | Pass |
| E2E | `npm run e2e -- e2e/global-search.spec.ts` | Pass |

---

## Checklist

- [x] Code formatted
- [x] Tests added/updated for changes
- [ ] Documentation updated (if applicable)
- [x] No secrets or credentials committed

---

## Notes (optional)

`/admin/search` returns JSON, so the header can consume it directly through the React API client. This should be updated when /v1/search api is ready. This is the issue for it #5520

PR #5610, opened 2026-07-13
feat(search): add versioned GET /v1/search unified search endpoint

### LINKED ISSUE #5520: [UI-REWRITE]: Create a new GET /v1/search API endpoint for unified search
### 🧭 Type of Feature

- [ ] Enhancement to existing functionality

### Background

The UI rewrite global search currently depends on `GET /admin/search` because it is the only unified search endpoint available today. That endpoint is mounted under `/admin` and protected by the admin dashboard permission:

```python
@admin_router.get("/search", response_class=JSONResponse)
@require_permission("admin.dashboard", allow_admin_bypass=False)
```

This works for the current Admin UI, but `/admin/search` is expected to be deprecated for client-facing search usage. Similar to #4399, the same capability should be exposed through a stable versioned API path that does not live under `/admin`.

### Proposed Endpoint

`GET /v1/search`

The new endpoint should preserve the behavior and response shape of `GET /admin/search`.

Query parameters should remain compatible:

- `q` (str, max 500) — free-text search query
- `tags` (str, optional) — tag filter expression
- `entity_types` (str, optional) — comma-separated entity types
- `include_inactive` (bool, default false)
- `limit` (int, default 8) — per-entity result limit
- `limit_per_type` (int, optional) — alias overriding `limit`
- `gateway_id` (str, optional) — gateway filter for tools/resources/prompts
- `team_id` (str, optional) — team scope filter

Supported entity types should match the existing unified search endpoint:

- `servers`
- `gateways`
- `tools`
- `resources`
- `prompts`
- `agents`
- `teams`
- `users`
- `roots`

Response shape should match the existing endpoint:

```json
{
  "query": "string",
  "tags": "string",
  "entity_types": ["tools"],
  "limit_per_type": 8,
  "filters_applied": {
    "q": "string",
    "tags": "string",
    "tag_groups": []
  },
  "results": {
    "tools": []
  },
  "groups": [
    {
      "entity_type": "tools",
      "count": 0,
      "items": []
    }
  ],
  "items": [],
  "count": 0
}
```

### Auth & RBAC

The endpoint should keep the same visibility and RBAC semantics as the existing unified search implementation:

- Token scoping must continue to control visible entities.
- Entity-specific permission denials should not fail the entire unified search response.
- `users` should only be searchable when the caller has the required user-management permission.
- Team filtering must continue to validate `team_id` server-side.

The exact top-level permission should be reviewed. The endpoint should not require admin-panel access merely because the route is used by the React client.

### User Story

**As a:** React client user  
**I want:** global search to call a versioned API endpoint  
**So that:** the UI does not depend on an admin route that may be deprecated.

#### Acceptance Criteria

```gherkin
Scenario: Search via the versioned API
  Given an authenticated caller with access to searchable entities
  When they request GET /v1/search?q=svelte&entity_types=tools,resources
  Then the response matches the existing /admin/search grouped result shape

Scenario: Restricted entity types remain protected
  Given a caller without user-management permission
  When they request GET /v1/search?q=alice&entity_types=users
  Then users are not leaked and the authorization behavior matches /admin/search

Scenario: Team scope is preserved
  Given a caller with scoped team access
  When they request GET /v1/search?q=server&team_id=<team-id>
  Then returned entities are filtered by the caller's visible team scope

Scenario: UI can migrate away from /admin/search
  Given GET /v1/search is available
  When the React global search client is updated
  Then no global search request is sent to /admin/search
```

### Tests

Add API coverage for:

- unauthenticated requests
- invalid `entity_types`
- unsupported or unauthorized `users` search
- `team_id` filtering
- `limit` and `limit_per_type`
- grouped and flattened response shape parity with `/admin/search`

Add or update frontend coverage when the React client migrates from `/admin/search` to `/v1/search`.

### Additional Context

Related prior pattern: #4399

Related UI work: #5518

gandhipratik203:
## Summary

Adds a new search API, `GET /v1/search`, that returns the same results as the existing `/admin/search` but without requiring admin access. This lets the app's global search use a stable, public API instead of an admin route that is being deprecated.

Closes #5520.

## What this PR does

Adds a public, versioned search API so the app's global search no longer needs admin-panel access.

```
BEFORE:  only  GET /admin/search    (requires admin.dashboard permission)
                 → only admins can search

AFTER:   also   GET /v1/search       (any logged-in user)
                 → results are still filtered by YOUR own permissions:
                     • a tool shows up only if you can read tools
                     • a team shows up only if it's in your scope
                     • users show up only with user-management permission
```

Two security fixes picked up during review:

```
Fix 1 — scoped API tokens could not use the endpoint
   BEFORE:  token scoped to "tools.read"  →  GET /v1/search  →  403
   AFTER:   token scoped to "tools.read"  →  GET /v1/search  →  200  (returns your tools)

Fix 2 — team search leaked across a narrowed token (non-admin)
   Non-admin in teams A + B, token scoped to B:
   BEFORE:  search teams  →  A, B   (leak)
   AFTER:   search teams  →  B
   (the admin-side version of this leak is the follow-up #5668)
```

## What changed

- **Shared the search logic.** Moved the search code out of the admin handler into a reusable helper so both routes run the same logic. `/admin/search` behaves exactly as before.
- **Added the new route** (`mcpgateway/routers/search.py`). `GET /search` only requires the caller to be logged in. Results are still filtered by the caller's existing permissions and team access, so nothing shows up that they could not already see, and `users` still requires the user-management permission.
- **Made it always available** at `/v1/search` (and the legacy `/search`), whether or not the admin API is enabled.

## Behavior parity

Same query params (`q`, `tags`, `entity_types`, `include_inactive`, `limit`, `limit_per_type`, `gateway_id`, `team_id`), same supported entity types, and the same grouped + flattened response shape as `/admin/search`.

## Auth & RBAC

- Top-level gate is authentication only, per the issue's guidance that the route should not require admin-panel access.
- Per-entity RBAC and token scoping unchanged.
- `users` remains gated on `admin.user_management`; team filtering still validated server-side.

## Tests

`tests/unit/mcpgateway/routers/test_search_router.py` covers:
- unauthenticated request rejected
- invalid / empty `entity_types` -> 400
- `users` dropped without permission (no leak); users-only request -> 403
- `team_id` forwarded to per-entity searches
- `limit` / `limit_per_type` alias behavior
- grouped + flattened response-shape parity with the shared core
- `/v1/search` present even when the admin API is disabled

## Scope

`/admin/search` is left in place as the fallback during the React client migration (issue #5518 owns switching the UI client). No frontend changes here.

## Verification

- Unit tests (`tests/unit/mcpgateway/routers/test_search_router.py`): 11 passed, plus the existing `admin_unified_search` suite green.
- Integration tests (`tests/integration/test_search_endpoint.py`): 8 passed. Run with:
  ```bash
  pytest tests/integration/test_search_endpoint.py --with-integration -v
  ```
  - Wiring (5): full ASGI stack (routing, auth middleware, orchestration) on the real app for `/v1/search` and legacy `/search`, including 401 deny paths.
  - Real-data RBAC/token scoping (3): seed real teams, a non-admin member with a real global `tools.read` role, and public/team-private tools, then run the real `PermissionService` and `admin_search_tools` (nothing stubbed at the boundary). A token scoped to team-B does not see team-A's private tool while still returning public and team-B tools; without token narrowing the same member does see team-A (confirming the deny was the token, not missing access); and a non-admin requesting `tools,users` has `users` dropped while tools still return.
- E2E (`tests/e2e/test_search_e2e.py`): 3 passed. Drives a **real signed JWT** through the full auth + token-scoping middleware (no dependency overrides) against the real app on a temp DB:
  - unauthenticated → 401 (real auth middleware)
  - non-admin JWT → 200, and the member's `tools.read` role returns the seeded public tool (proves no admin needed — real RBAC, real data)
  - `tools.read`-scoped JWT → 200 (token-scoping middleware allows `/search`)

  Verified empirically: disabling the `/search` allow makes the scoped case fail with 403, and the deny/happy contrast confirms auth is genuinely enforced.
- bandit: no issues; interrogate: 100%; pylint: 10.00/10; ruff: clean

Note: the integration real-data tests inject identity (real `PermissionService`, `admin_search_tools`, and DB rows); the e2e test above covers the full JWT-middleware validation path end-to-end.

### Manual verification

Reproducible with curl (uses your local `JWT_SECRET_KEY`; no secrets required in this description):

```bash
# Start the gateway (dev server on :8000)
make dev

# Mint an admin JWT
export TOKEN=$(python3 -m mcpgateway.utils.create_jwt_token \
  --username admin@example.com --exp 10080 --secret "$JWT_SECRET_KEY")
```

1. **Deny path — unauthenticated request is rejected**
   ```bash
   curl -s -o /dev/null -w "%{http_code}\n" "http://localhost:8000/v1/search?q=a"
   # Expected: 401
   ```

2. **Happy path — authenticated search returns the grouped contract**
   ```bash
   curl -s -H "Authorization: Bearer $TOKEN" \
     "http://localhost:8000/v1/search?q=a&entity_types=tools,servers" | python3 -m json.tool
   # Expected: 200 with keys query, tags, entity_types, limit_per_type,
   #           filters_applied{q,tags,tag_groups}, results{...}, groups[], items[], count
   ```

3. **Parity with the legacy admin route** — `/admin/search` requires admin privileges, so this step needs a token whose `user.is_admin` claim is set. The `create_jwt_token` CLI token above is auth-only (it can reach `/v1/search` but 403s on `/admin/*`), so mint an admin-privileged token for the comparison:
   ```bash
   ADMIN_TOKEN=$(python3 - <<'PY'
   import time, uuid, jwt
   now = int(time.time())
   print(jwt.encode({"sub": "admin@example.com", "iat": now, "exp": now + 3600,
     "iss": "mcpgateway", "aud": "mcpgateway-api", "jti": uuid.uuid4().hex,
     "user": {"email": "admin@example.com", "is_admin": True, "auth_provider": "local"}, "teams": None},
     settings.jwt_secret_key.get_secret_value(), algorithm="HS256"))
   PY
   )
   diff <(curl -s -H "Authorization: Bearer $ADMIN_TOKEN" "http://localhost:8000/v1/search?q=a&entity_types=tools") \
        <(curl -s -H "Authorization: Bearer $ADMIN_TOKEN" "http://localhost:8000/admin/search?q=a&entity_types=tools")
   # Expected: no output (identical payloads)
   ```

4. **Restricted entity (`users`) stays protected**
   ```bash
   curl -s -H "Authorization: Bearer $TOKEN" \
     "http://localhost:8000/v1/search?q=admin&entity_types=users" | python3 -m json.tool
   # Admin token sees the "users" group; a non-admin token (no admin.user_management)
   # has "users" dropped, and a users-only request returns 403 — no leak.
   ```

5. **Available independent of the admin API** — `/v1/search` is an always-on route, so it appears in `openapi.json` under the **Search** tag even with `MCPGATEWAY_ADMIN_API_ENABLED=false`.

| Check | Expected |
|-------|----------|
| Unauthenticated | 401 |
| Authenticated grouped response | 200, matches `/admin/search` shape |
| Parity with `/admin/search` | identical payload |
| `users` protection | dropped / 403 without permission |
| Present without admin API | route in OpenAPI under **Search** tag |

Also verified in the Swagger UI (`/docs` → **Search** → `GET /v1/search` → 200).

### DIFF
--- a/mcpgateway/admin.py
+++ b/mcpgateway/admin.py
@@ -11464,26 +11464,27 @@ async def admin_search_a2a_agents(
     return _build_search_response(entity_key="agents", entity_type="agents", items=agents, query=search_query, tags=normalized_tags, tag_groups=tag_groups)
 
 
-@admin_router.get("/search", response_class=JSONResponse)
-@require_permission("admin.dashboard", allow_admin_bypass=False)
-async def admin_unified_search(
-    q: str = Query("", max_length=500, description="Search query"),
-    tags: QueryTagsFilter = None,
-    entity_types: QueryEntityTypes = None,
-    include_inactive: bool = False,
-    limit: int = Query(8, ge=1, le=settings.pagination_max_page_size, description="Per-entity result limit"),
-    limit_per_type: Optional[int] = Query(
-        None,
-        ge=1,
-        le=settings.pagination_max_page_size,
-        description="Optional alias for per-entity result limit",
-    ),
-    gateway_id: QueryGatewayIdList = None,
-    team_id: Optional[str] = Depends(_validated_team_id_param),
-    db: Session = Depends(get_db),
-    user=Depends(get_current_user_with_permissions),
-):
-    """Unified search across primary admin entities.
+async def perform_unified_search(
+    *,
+    q: str,
+    tags: Optional[str],
+    entity_types: Optional[str],
+    include_inactive: bool,
+    limit: int,
+    limit_per_type: Optional[int],
+    gateway_id: Optional[str],
+    team_id: Optional[str],
+    db: Session,
+    user: Any,
+) -> dict[str, Any]:
+    """Unified search across primary entities (shared, permission-agnostic core).
+
+    This is the single source of truth for unified search behavior. It performs
+    NO top-level permission check of its own — callers are responsible for the
+    outer gate (``admin.dashboard`` for the admin route, authentication-only for
+    the versioned ``/v1/search`` route). Per-entity RBAC and token scoping are
+    still enforced inside each ``admin_search_*`` call, so visibility is correct
+    regardless of the outer gate.
 
     Searches servers, gateways, tools, resources, prompts, agents, teams, roots,
     and optionally users (when the caller has ``admin.user_management`` permission).
@@ -11737,6 +11738,62 @@ async def admin_unified_search(
         "items": flat_items,
         "count": len(flat_items),
     }
+
+
+@admin_router.get("/search", response_class=JSONResponse)
+@require_permission("admin.dashboard", allow_admin_bypass=False)
+async def admin_unified_search(
+    q: str = Query("", max_length=500, description="Search query"),
+    tags: QueryTagsFilter = None,
+    entity_types: QueryEntityTypes = None,
+    include_inactive: bool = False,
+    limit: int = Query(8, ge=1, le=settings.pagination_max_page_size, description="Per-entity result limit"),
+    limit_per_type: Optional[int] = Query(
+        None,
+        ge=1,
+        le=settings.pagination_max_page_size,
+        description="Optional alias for per-entity result limit",
+    ),
+    gateway_id: QueryGatewayIdList = None,
+    team_id: Optional[str] = Depends(_validated_team_id_param),
+    db: Session = Depends(get_db),
+    user=Depends(get_current_user_with_permissions),
+):
+    """Unified search across primary admin entities (admin-gated wrapper).
+
+    Thin wrapper that enforces the ``admin.dashboard`` permission and delegates
+    to :func:`perform_unified_search`. The versioned ``GET /v1/search`` route
+    reuses the same helper without the admin-panel permission gate.
+
+    Args:
+        q (str): Free-text search query.
+        tags (Optional[str]): Tag filter expression (comma=OR, plus=AND).
+        entity_types (Optional[str]): Optional comma-separated entity type list.
+            Supported values: servers, gateways, tools, resources, prompts,
+            agents, teams, users, roots.
+        include_inactive (bool): Whether to include inactive entities.
+        limit (int): Default per-entity limit for returned items.
+        limit_per_type (Optional[int]): Optional alias overriding ``limit``.
+        gateway_id (Optional[str]): Gateway filter for tools/resources/prompts.
+        team_id (Optional[str]): Team scope filter.
+        db (Session): Database session.
+        user: Authenticated user context.
+
+    Returns:
+        dict[str, Any]: Grouped and flattened search results with metadata.
+    """
+    return await perform_unified_search(
+        q=q,
+        tags=tags,
+        entity_types=entity_types,
+        include_inactive=include_inactive,
+        limit=limit,
+        limit_per_type=limit_per_type,
+        gateway_id=gateway_id,
+        team_id=team_id,
+        db=db,
+        user=user,
+    )
 
 
 @admin_router.get("/tools/{tool_id}", response_model=ToolRead)
--- a/mcpgateway/admin.py
+++ b/mcpgateway/admin.py
@@ -11479,12 +11479,10 @@ async def perform_unified_search(
 ) -> dict[str, Any]:
     """Unified search across primary entities (shared, permission-agnostic core).
 
-    This is the single source of truth for unified search behavior. It performs
-    NO top-level permission check of its own — callers are responsible for the
-    outer gate (``admin.dashboard`` for the admin route, authentication-only for
-    the versioned ``/v1/search`` route). Per-entity RBAC and token scoping are
-    still enforced inside each ``admin_search_*`` call, so visibility is correct
-    regardless of the outer gate.
+    Single source of truth for unified search. Performs no top-level permission
+    check — callers own the outer gate (``admin.dashboard`` for the admin route,
+    auth-only for ``/v1/search``). Per-entity RBAC and token scoping are still
+    enforced inside each ``admin_search_*`` call.
 
     Searches servers, gateways, tools, resources, prompts, agents, teams, roots,
     and optionally users (when the caller has ``admin.user_management`` permission).
@@ -11762,8 +11760,7 @@ async def admin_unified_search(
     """Unified search across primary admin entities (admin-gated wrapper).
 
     Thin wrapper that enforces the ``admin.dashboard`` permission and delegates
-    to :func:`perform_unified_search`. The versioned ``GET /v1/search`` route
-    reuses the same helper without the admin-panel permission gate.
+    to :func:`perform_unified_search`.
 
     Args:
         q (str): Free-text search query.
--- a/mcpgateway/admin.py
+++ b/mcpgateway/admin.py
@@ -5591,9 +5591,17 @@ async def admin_search_teams(
         # Non-admin search
         # Reuse user team fetching
         all_teams = await team_service.get_user_teams(user_email, include_personal=True)
+        # Narrow to the caller's normalized token scope (Layer 1). get_user_teams
+        # returns every membership and ignores token scope, so a token narrowed to
+        # a team subset would otherwise leak sibling teams the caller belongs to but
+        # is scoped out of. _get_user_team_ids honors token_teams/_cached_team_ids;
+        # the caller's own personal team stays visible (owner is always visible).
+        scoped_team_ids = set(await _get_user_team_ids(user, db))
         # Filter in memory
         filtered = []
         for t in all_teams:
+            if not getattr(t, "is_personal", False) and t.id not in scoped_team_ids:
+                continue
             if not include_inactive and not t.is_active:
                 continue
             if visibility and t.visibility != visibility:
--- a/mcpgateway/routers/search.py
+++ b/mcpgateway/routers/search.py
@@ -1,0 +1,101 @@
+# -*- coding: utf-8 -*-
+"""Location: ./mcpgateway/routers/search.py
+Copyright 2026
+SPDX-License-Identifier: Apache-2.0
+Authors: Mihai Criveti
+
+Unified Search API Router.
+
+Exposes the unified cross-entity search capability at a stable, versioned,
+non-admin path (``GET /v1/search``). Historically this capability was only
+reachable through ``GET /admin/search``, which is gated on the ``admin.dashboard``
+permission and lives under the admin dashboard router. Client-facing callers
+(e.g. the React global search) should not depend on an admin-panel route that
+may be deprecated, so this router re-exposes the same behavior and response
+shape without the admin-panel gate.
+
+Security model:
+    This route requires only authentication at the top level. It does NOT add
+    an ``admin.dashboard`` gate. Real authorization is enforced per-entity
+    inside :func:`mcpgateway.admin.perform_unified_search`: each entity search
+    carries its own RBAC permission (``tools.read``, ``gateways.read``,
+    ``admin.user_management`` for users, etc.) and token scoping continues to
+    filter visible entities. Entity-specific denials are suppressed so a single
+    restricted entity type never fails the whole search or leaks existence.
+"""
+
+# Standard
+import logging
+from typing import Any, Optional
+
+# Third-Party
+from fastapi import APIRouter, Depends, Query
+from fastapi.responses import JSONResponse
+from sqlalchemy.orm import Session
+
+# First-Party
+from mcpgateway.admin import _validated_team_id_param, perform_unified_search  # noqa: PLC2701 — reuse admin team_id validation for parity
+from mcpgateway.common.query_params import QueryEntityTypes, QueryGatewayIdList, QueryTagsFilter
+from mcpgateway.config import settings
+from mcpgateway.db import get_db
+from mcpgateway.middleware.rbac import get_current_user_with_permissions
+
+logger = logging.getLogger(__name__)
+
+router = APIRouter(tags=["Search"])
+
+
+@router.get("/search", response_class=JSONResponse)
+async def unified_search(
+    q: str = Query("", max_length=500, description="Search query"),
+    tags: QueryTagsFilter = None,
+    entity_types: QueryEntityTypes = None,
+    include_inactive: bool = False,
+    limit: int = Query(8, ge=1, le=settings.pagination_max_page_size, description="Per-entity result limit"),
+    limit_per_type: Optional[int] = Query(
+        None,
+        ge=1,
+        le=settings.pagination_max_page_size,
+        description="Optional alias for per-entity result limit",
+    ),
+    gateway_id: QueryGatewayIdList = None,
+    team_id: Optional[str] = Depends(_validated_team_id_param),
+    db: Session = Depends(get_db),
+    user: Any = Depends(get_current_user_with_permissions),
+) -> dict[str, Any]:
+    """Unified search across primary entities (versioned, non-admin route).
+
+    Preserves the behavior, query parameters, and grouped/flattened response
+    shape of ``GET /admin/search`` but without the ``admin.dashboard`` gate.
+    Delegates to :func:`mcpgateway.admin.perform_unified_search`; per-entity
+    RBAC and token scoping are enforced inside that helper.
+
+    Args:
+        q (str): Free-text search query.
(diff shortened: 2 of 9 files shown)


marekdano (commented):
Nice work on this — the refactor into a shared `perform_unified_search` core is clean, `/admin/search` behavior is preserved exactly, and the test suite is genuinely thorough (the real-data token-scoping integration tests are a highlight). I walked through the security model and it holds up: every `admin_search_*` carries its own `@require_permission(..., allow_admin_bypass=False)`, `require_permission` enforces on direct calls, and `_safe_entity_search` swallows per-entity 401/403 into empty results — so dropping the `admin.dashboard` gate is safe. 👍

A few things worth considering, none blocking:

### Suggestions (worth a follow-up)

  - **Extract the shared core out of `admin.py`.** The goal here is to decouple client search from admin, but `routers/search.py` imports `perform_unified_search` and the private `_validated_team_id_param` directly from `mcpgateway.admin` (with a `noqa: PLC2701`), so the new "non-admin" router still hard-depends on the entire `admin.py` module loading even when the admin API is disabled. Moving the core into something like `mcpgateway/services/search_service.py` (imported by both `admin.py` and the router) would make the decoupling real and drop the private cross-module import. Reasonable to defer, but it's the natural next step.

  - **Add coverage for a real per-entity denial.** The "one entity's denial must not collapse the whole search" guarantee is arguably the most important behavior here, but no test exercises a *real* decorator denial flowing through `_safe_entity_search` — the unit tests mock every `admin_search_*`, and the real-data tests grant a global `tools.read`. A test seeding a caller who lacks `servers.read` but has `tools.read`, requesting `entity_types=servers,tools` and asserting `200` with empty `servers` + populated `tools`, would lock that path down.

### Minor notes

  - **`users` pre-check vs. its own gate.** `perform_unified_search` pre-checks only `admin.user_management`, but `admin_search_users` allows `admin.user_management` **OR** `teams.manage_members`. This code is unchanged, but with the outer gate now being auth-only, a `teams.manage_members`-only caller can reach `/v1/search` for the first time and will have `users` silently dropped even though `/admin/users/search` would serve them. It errs on the safe/over-restrictive side (no leak), so not a blocker — just worth a deliberate decision on which gate is authoritative.

  - **Cost surface.** Any authenticated user can now trigger up to 9 entity searches. Denied entities short-circuit at the decorator before any DB query (nice), and results are capped at `pagination_max_page_size`, so this is fine — just flagging it as a candidate for rate limiting if that isn't already handled by global middleware.

Overall this looks good to me once the test gap is filled — the rest can be follow-ups. 

Thanks for the careful PR description and the manual/curl verification steps.

mcpgateway/routers/search.py
```diff
@@ -0,0 +1,94 @@
...
+from mcpgateway.config import settings
+from mcpgateway.db import get_db
+from mcpgateway.middleware.rbac import get_current_user_with_permissions
+
+logger = logging.getLogger(__name__)
+
+router = APIRouter(tags=["Search"])
+
+
+@router.get("/search", response_class=JSONResponse)
```
vishu-bh, line 37: TokenScopingMiddleware default-denies paths absent from its permission map. `/v1/search` normalizes to `/search`, which has no mapping. Any valid JWT with non-empty, non-wildcard permission scopes receives 403 before this handler runs. Current integration tests use `Bearer x` plus dependency overrides, so they bypass real JWT scope validation. Add explicit authenticated-only handling for `/search` while retaining downstream per-entity RBAC, plus real JWT regression tests.

mcpgateway/admin.py
vishu-bh, line 11702: Unified team search delegates to `admin_search_teams`, whose non-admin path reloads all DB memberships and ignores `token_teams` and `_cached_team_ids`. A user belonging to teams A and B with a token narrowed to B can receive A's ID, name, slug, description, visibility, and status. Filter team results through normalized token scope and add narrowed-token deny tests for teams.

vishu-bh (changes requested):
Api looks alright. found some blocking issue before moving it is ready for merge. 

Added couple inline comments for it.

marekdano (dismissed):
The new commits meaningfully improved the token/team scoping story and added good tests. The team-search fix is a genuine security improvement.

LGTM 🚀

vishu-bh (dismissed):
LGTM

mcpgateway/routers/search.py
gandhipratik203: Confirmed — `/search` was hitting the middleware's default-deny, so scoped tokens got 403 before the handler. Fixed in `13a6f6b7d`: `/search` is now authenticated-only in `_check_permission_restrictions` (other token restrictions still apply; `/admin/search` keeps its gate), plus real-JWT regression tests (scoped token → 200, unmapped control → 403).

mcpgateway/admin.py
gandhipratik203: Confirmed — non-admin path fixed in `f5ac78a5c` by routing team results through the token-aware `_get_user_team_ids` (personal team kept, no-op for unscoped), with narrowed-token deny + contrast tests.

marekdano (dismissed):
Both blocking issues are resolved with tests, CI is fully green!

LGTM 🚀

vishu-bh (dismissed):
LGTM 🚀

msureshkumar88 (changes requested):
Thanks for this — the shared-core extraction is clean, and the two security fixes (scoped-token 403 on `/v1/search`, non-admin team-search leak) are well targeted with good deny-path regression tests. Ran the full new suite locally (14 unit + 11 integration + 3 e2e) plus the existing `admin_unified_search` suite and `test_api_versioning_parity.py` — all green, and `ruff`/`bandit` clean on the changed files, matching what's in the description.

One blocking item before merge:

**`mcpgateway/routers/search.py` imports `mcpgateway.admin` at module top level, which forces the entire admin module to load even when `MCPGATEWAY_ADMIN_API_ENABLED=false`.**

`search.py` is wired into `_assemble_routers()` in the "always-on" group (`mcpgateway/api/v1/__init__.py`), with no feature-flag guard. Every other admin-adjacent router in that same file (`admin_router`, `runtime_admin_router`, `well_known_admin_router`, `compliance_router`, `llm_admin_router`) is deliberately gated behind `if settings.mcpgateway_admin_api_enabled:` specifically so `mcpgateway.admin` isn't imported in admin-disabled deployments.

Verified this empirically — with `MCPGATEWAY_ADMIN_API_ENABLED=false`, `mcpgateway.admin` still ends up in `sys.modules` (the startup log even says "Admin API routes not mounted" right above it). Importing `admin.py` isn't free either — it has module-level side effects that instantiate `AdminStatsCache`, `AuthCache`, `MetricsBufferService`, `Argon2PasswordService`, and `ContentSecurityService`. Measured import cost in isolation: ~1.8s. That's added startup latency and memory/crypto-service footprint for every deployment, including ones that disabled the admin API specifically to shrink surface area — which cuts against this PR's own framing that `/v1/search` doesn't depend on the admin dashboard.

A couple of ways to resolve, happy to defer to whichever fits best:
1. Move `from mcpgateway.admin import _validated_team_id_param, perform_unified_search` inside `unified_search()`'s function body so the cost is deferred to first request instead of process startup — smallest change.
2. Extract `perform_unified_search` (and `_validated_team_id_param`) into a lighter shared module that both `admin.py` and `routers/search.py` import from, avoiding the reach-into-private-symbol coupling (`# noqa: PLC2701` on that import is already flagging this smell) as well as the import cost.

A few smaller, non-blocking notes:
- `mcpgateway/routers/search.py`: the module-level `logger` is created but never used — either wire it up or drop it.
- Worth a `CHANGELOG.md` entry under `[Unreleased]` given this adds a new public API surface plus two security fixes — the existing entries in that section follow a consistent PR-link format this would slot into.
- No test currently drives a public-only token (no `teams` claim at all) through `/v1/search` to confirm it degrades to public-only visibility — every scoping test uses an explicit team list. Might be worth one more case.

Nice work on the security fixes and test depth — just want the import-cost regression sorted before this merges.

gandhipratik203:
Thanks for the review. Addressed:

- **Blocking (import cost)** — `3a291e9d9`, your option 1: `team_id` is a raw `Query(...)` and `perform_unified_search` / `_validated_team_id_param` are imported lazily in the handler, so `admin.py` only loads on the first `/v1/search` call. Verified `mcpgateway.admin` no longer enters `sys.modules` on router import; HTTP behavior unchanged (invalid `team_id` -> 400).
- **Unused `logger`** and **public-only-scope test** (`token_teams=[]` -> sees public, not team-private) — `1197f46eb`.
- **CHANGELOG** — skipped intentionally: comparable v1 PRs (#5455, #5443) don't self-add entries; they're curated at release time. Happy to add if preferred.

Suite green: 11 unit + 12 integration + 3 e2e.

msureshkumar88 (dismissed):
Ran a real-process e2e validation of this PR — no `TestClient`, no dependency overrides: actual `uvicorn` process, a real SQLite file, real signed JWTs through the full auth + RBAC + token-scoping stack, driven purely over HTTP with `curl`. Script added at `tests/live_gateway/e2e_pr5610_v1_search.sh`.

**Result: 10/10 checks passed. Approving.**

### Config used

```
DATABASE_URL=sqlite:///<tmp>/e2e.db
JWT_SECRET_KEY=<throwaway 32+ byte secret>
AUTH_ENCRYPTION_SECRET=<throwaway secret>
AUTH_REQUIRED=true
MCPGATEWAY_UI_ENABLED=false
Boot A: MCPGATEWAY_ADMIN_API_ENABLED=false   (proves /v1/search works without the admin API)
Boot B: MCPGATEWAY_ADMIN_API_ENABLED=true    (proves /admin/search parity is unaffected)
```

Real data seeded directly via SQLAlchemy against the gateway's own migrated DB (after its own startup/bootstrap ran, to avoid double-migrating): 1 admin user (`platform_admin`-style role, `*` permission), 1 non-admin user, 2 teams, a public/team-A/team-B tool triple, and a public server — same shape as the existing integration test fixtures, just exercised over real HTTP instead of `TestClient`.

### Commands run (abridged)

```bash
uv run uvicorn mcpgateway.main:app --host 127.0.0.1 --port <port>   # Boot A, then Boot B

curl http://127.0.0.1:<port>/v1/search?q=<term>
curl http://127.0.0.1:<port>/admin/search?q=<term>  -H "Authorization: Bearer <real admin JWT>"
curl http://127.0.0.1:<port>/v1/search?q=<term>&entity_types=tools,servers -H "Authorization: Bearer <tools.read-scoped JWT, teams=[A,B]>"
curl http://127.0.0.1:<port>/v1/search?q=<term>&entity_types=teams -H "Authorization: Bearer <JWT teams=[B] only>"
curl http://127.0.0.1:<port>/v1/search?q=<term>&entity_types=teams -H "Authorization: Bearer <JWT teams=[A,B]>"
curl http://127.0.0.1:<port>/admin/search?q=<term> -H "Authorization: Bearer <non-admin tools.read-scoped JWT>"
```

### Output (checklist)

```
✅ 1) GET /v1/search unauthenticated -> 401
✅ 2) GET /admin/search absent (404) when admin API disabled, even for a real admin
✅ 3) Fix 1: tools.read-scoped token reaches /v1/search with 200 (not 403)
✅ 4) teams=[A,B] token sees public + team-A + team-B tools
✅ 5) real servers.read denial swallowed to empty list, request still 200 (not collapsed)
✅ 6) Fix 2: token narrowed to team B sees team B, NOT sibling team A
✅ 7) Contrast: widened token sees BOTH teams (proves #6 was the token scope, not broken data)
✅ 8) /v1/search still 200 for scoped token when admin API is enabled
✅ 9) /admin/search still requires admin.dashboard: non-admin -> 403
✅ 10) /admin/search still works for a real admin (200) after the perform_unified_search refactor

Summary: 10 passed, 0 failed
```

This lines up with the PR description's two named fixes (scoped-token 403→200, non-admin team-search leak) and confirms `/admin/search` behavior is unchanged after the `perform_unified_search` extraction. Combined with the earlier code review (prior blocking items resolved, tests/lint/bandit clean on the touched files), this looks good to merge from my side.

msureshkumar88 (approved):
Blocking item resolved: the `mcpgateway.admin` import in `routers/search.py` is now deferred inside `unified_search()` (option 1 from my earlier review), so admin-disabled deployments no longer pay the startup import cost. Verified in 3a291e9.

Minor notes also addressed: unused `logger` dropped, and a public-only-token (no `teams` claim) test was added covering the degenerate scope case (1197f46).

CI is green (30/30). LGTM.

PR #5612, opened 2026-07-13
UI rewrite/tool form adjustments 

a-effort:
Follow-up to #5604

Scope: Small UX updates to the form, a component improvement, and a light-theme bug fix.

**Problems addressed**:
- The URL field's placeholder (https://api.example.com/v1/endpoint) has a host that doesn't resolve, so a user who tries it verbatim gets a fetch failure from the OpenAPI schema generator.
- The generate button was silently disabled until a URL was entered, without telling the user why.

**Changes**:
URL placeholder -> shape hint. `https://[host]/[path]`, localized to `[servidor]/[ruta]` (es-ES) and `[servidor]/[caminho]` (pt-BR), so the input visibly documents its format instead of pretending to be a working URL. Uses square brackets rather than angle brackets because react-intl's ICU parser rejects <tag> as unclosed rich text.

Helper text -> working example. `e.g. https://reqres.in/api/users` (localized p. ej. / ex.) appears under the URL field. It resolves end-to-end with the current generator (spec at host root, no auth on the spec fetch), so users can copy it and see the "Generate" flow succeed on the first try.

Name placeholder -> `list-users`. Was `the-greatest-tool`. Pairs with the helper URL so name and endpoint tell one coherent example, and demonstrates the kebab-case verb-noun convention. Localized to listar-usuarios in es-ES and pt-BR.

Generate button: inline validation instead of disabling. Was disabled until the URL was non-empty, which gave no explanation for the greyed-out state. Now the button is always clickable (except during an in-flight request); clicking with an empty URL surfaces the existing localized "URL is required" message next to the field and moves focus to the URL input. The error clears the moment the user starts typing. Existing aria-invalid + aria-describedby wiring covers screen-reader announcement. Adds a scoped `setUrlError` helper to `useToolForm` (rather than exposing the full setter) so the form can only touch the URL error, matching the existing pattern for the schema error.

Extends aria-describedby on the URL input to reference the new helper text.

**Misc small improvements**

I noticed an unrelated light-theme regression on the Tools page (card-tag.tsx) so am including a fix. The default CardTag variant hardcoded text-white. Swapped to the paired `--tool-badge-fg` token already defined in index.css so the foreground now flips with the theme like the background already did.

Textarea auto-sizes to its content (textarea.tsx). Added field-sizing: content at the shared component level so textareas grow with their value rather than scrolling. Applies app-wide (schema fields, description, response filter, etc.). Progressive enhancement: pre-2024 browsers ignore it and behave as before.

Notes
No tests asserted on the placeholder strings, so no test churn was required.

https://github.com/user-attachments/assets/f965d3be-a419-4159-99bb-ebab0d056fae


a-effort:
References for the misc small fixes: 
1. Get rid of this scrollbar at component level (field is resizable and it gets in the way of the copy button).
scroll bar
2. Light theme issue fixed: 
light theme

a-effort:
DCO check appears stale; re-running the check.

marekdano (approved):
### Findings

Nothing survived verification as a blocking correctness issue. The changes are internally consistent:

  - Dropping `!url.trim()` and `!requestType` from the button's `disabled` guard is safe — the click handler re-validates `url` inline, and `requestType` defaults to `"POST"` (`useToolForm.ts:225`) with radios that can never all be deselected, so it can't reach `generateSchema` empty.
  - `setUrlError` is correctly added to the interface, the return object, and consumes only the `url` slot — matching the existing scoped-setter pattern.
  - `aria-describedby="url-helper url-error"` ordering and the always-present `#url-helper` node are correct for screen readers.
  - The `CardTag` fix is genuinely correct: `--tool-badge-fg` resolves to `neutral-800` in light and `white` in dark, so the previously-invisible white-on-light-badge text now flips properly.

  ### Minor notes (non-blocking)

  - **`ToolForm.tsx:168`** — `document.getElementById("tool-url")?.focus()` reaches into the DOM directly rather than via a ref. It works and won't break, but a `useRef` on the `Input` would be more idiomatic for this codebase's React style. Purely stylistic.
  - **`tools.json` helper URL** — the example points at `reqres.in`, a third-party hosted demo API. The PR body justifies it (resolves end-to-end today), but it does make the "copy-paste and it just works" promise depend on an external service staying up and auth-free. Worth being aware of if that endpoint ever changes; not a code defect.
  - **`useToolForm.ts` / `ToolForm.tsx`** — `setUrlError(undefined)` fires on every keystroke, spreading a fresh `errors` object even when no url error exists. Negligible in React, not worth changing.


Clean, well-scoped PR — LGTM 🚀 

DCO check should be fixed before merging

a-effort:
Closed and reopened trying to get DCO to run again. My 7 commits are signed off on as expected. It seems the flagged commits are pre-existing on the base branch.

PR #5625, opened 2026-07-14
Fix playwright tests

### LINKED ISSUE #5624: [TESTING]: Failing PW tests
### ✅ Test Summary

Fix the following failing PW tests:

```
=========================== short test summary info ============================
FAILED tests/playwright/entities/test_prompts.py::TestPromptsCRUD::test_delete_prompt[chromium]
FAILED tests/playwright/entities/test_resources.py::TestResourcesCRUD::test_delete_resource[chromium]
FAILED tests/playwright/entities/test_resources_extended.py::TestResourcesViewModal::test_view_modal_different_resources_show_different_data[chromium]
FAILED tests/playwright/test_admin_menu_visibility.py::TestAdminMenuVisibility::test_platform_admin_sees_all_sections[chromium]
FAILED tests/playwright/test_admin_menu_visibility.py::TestAdminMenuVisibility::test_team_admin_sees_team_sections[chromium]
FAILED tests/playwright/test_admin_menu_visibility.py::TestAdminMenuVisibility::test_developer_sees_tool_sections[chromium]
FAILED tests/playwright/test_admin_menu_visibility.py::TestAdminMenuVisibility::test_viewer_sees_readonly_sections[chromium]
FAILED tests/playwright/test_admin_menu_visibility.py::TestAdminMenuVisibility::test_menu_sections_match_permissions[chromium]
FAILED tests/playwright/test_admin_url_context.py::TestAdminIframeContext::test_iframe_team_selector_onclick_stripped_but_delegation_works[chromium]
FAILED tests/playwright/test_organization.py::TestTeams::test_edit_settings_button[chromium]
FAILED tests/playwright/test_organization.py::TestTeams::test_edit_settings_max_members_defaults_to_global[chromium]
= 11 failed, 657 passed, 254 skipped, 8 warnings, 9 rerun in 2357.10s (0:39:17) =
❌ UI tests failed!
make: *** [Makefile:7122: test-ui-headless] Error 1
```

---

### 🧪 Test Type
Choose one or more applicable categories:

- [ ] Unit tests (Python)
- [ ] Integration / end-to-end tests
- [ ] Manual UI testing (admin panel)
- [ ] Transport coverage (HTTP, WebSocket, SSE, stdio)
- [ ] Federation / multi-gateway testing
- [ ] Configuration / environment-specific behavior
- [X] Other (explain below)

Playwright

---

### 🧬 Scope & Affected Components
Select what this test covers or validates:

- [ ] `mcpgateway` core (API logic, handlers)
- [X] Admin UI (HTMX / Alpine / Tailwind)
- [ ] Tool/Resource/Prompt logic
- [ ] Federation sync/discovery
- [ ] Server interactions or SSE
- [ ] Auth / JWT / security flows
- [ ] Observability (logging, metrics)
- [ ] Makefile, shell scripts or CLI
- [ ] Containerized setup (Docker/Podman)
- [ ] Other (explain below)

gcgoncalves:
# Pull Request

## 🔗 Related Issue

Closes https://github.com/IBM/mcp-context-forge/issues/5624

---

## 📝 Summary

Some playwright tests are outdated/failing

---

## 📏 Reviewability

---

## 🏷️ Type of Change

- [ ] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [X] Other (describe below)

Testing

### DIFF
--- a/tests/AGENTS.md
+++ b/tests/AGENTS.md
@@ -139,12 +139,31 @@ All extend `BasePage` which provides `SidebarComponent` for tab navigation. Use 
 
 Fixtures in `conftest.py` handle login automatically (JWT cookie injection by default).
 
+### CSP-Safe JS Polling
+
+Strict CSP (`script-src 'self'`, no `unsafe-eval`) rejects `Page.wait_for_function()`/`Frame.wait_for_function()` immediately after navigation. Use `wait_for_js_condition()` instead — it polls via `evaluate()`, which isn't subject to that restriction:
+
+```python
+from .pages.admin_utils import wait_for_js_condition
+wait_for_js_condition(page, "typeof window.Admin !== 'undefined'", timeout=30000)
+```
+
 ### Writing New Playwright Tests
 
 1. Create a page object in `tests/playwright/pages/` extending `BasePage`
 2. Add a fixture in `tests/playwright/conftest.py`
 3. Create test file in `tests/playwright/test_<feature>.py`
 4. Use `pytest.skip()` when prerequisites aren't met (e.g., plugins not enabled)
+
+### Handling Async DB Operations
+
+After a DELETE, use `wait_for_entity_deleted()` to ride out DB commit propagation lag instead of asserting immediately:
+
+```python
+from .pages.admin_utils import wait_for_entity_deleted
+assert delete_resource(page, resource_id)
+assert wait_for_entity_deleted(page, "resources", resource_name)
+```
 
 ## JavaScript Unit Tests
 
--- a/tests/AGENTS.md
+++ b/tests/AGENTS.md
@@ -165,6 +165,22 @@ assert delete_resource(page, resource_id)
 assert wait_for_entity_deleted(page, "resources", resource_name)
 ```
 
+### Polling Utilities
+
+Lower-level primitives behind `wait_for_entity_deleted()`, for custom predicates:
+
+```python
+from .pages.admin_utils import wait_for_ui_by_retries, wait_for_ui_by_deadline
+
+# Fixed attempt count — API-poll style, each attempt cheap and uniform
+wait_for_ui_by_retries(page, lambda: check_something(), retries=5)
+
+# Wall-clock deadline — loops that also perform slow UI actions (navigation, reload)
+wait_for_ui_by_deadline(page, lambda: check_something(), deadline_seconds=8)
+```
+
+Both accept an optional `delay_ms(attempt) -> int` callable for backoff between attempts.
+
 ## JavaScript Unit Tests
 
 Admin UI JS logic is tested with Vitest (jsdom). Tests live in `tests/unit/js/`.
--- a/tests/playwright/conftest.py
+++ b/tests/playwright/conftest.py
@@ -27,6 +27,7 @@ from mcpgateway.config import Settings
 from tests.helpers.api_helpers import ApiTestHelper
 from tests.helpers.auth import make_test_jwt
 from .pages.admin_page import AdminPage
+from .pages.admin_utils import wait_for_js_condition
 from .pages.agents_page import AgentsPage
 from .pages.gateways_page import GatewaysPage
 from .pages.login_page import LoginPage
@@ -316,10 +317,13 @@ def _ensure_admin_logged_in(page: Page, base_url: str) -> None:
             raise AssertionError("Admin page failed to load: Internal Server Error (500)")
         raise
 
-    # Wait for JS initialization (showTab + HTMX + event delegation) before any tab clicks
+    # Wait for JS initialization (showTab + HTMX + event delegation) before any tab clicks.
+    # evaluate()-based poll: wait_for_function's eval() mechanism is rejected by strict
+    # CSP (script-src 'self', no unsafe-eval) right after navigation.
     try:
-        page.wait_for_function(
-            "typeof window.Admin.showTab === 'function' && typeof window.htmx !== 'undefined' && window.Admin.eventDelegationInitialized === true",
+        wait_for_js_condition(
+            page,
+            "typeof window.Admin !== 'undefined' && typeof window.Admin.showTab === 'function' && typeof window.htmx !== 'undefined' && window.Admin.eventDelegationInitialized === true",
             timeout=30000,
         )
     except PlaywrightTimeoutError:
--- a/tests/playwright/conftest.py
+++ b/tests/playwright/conftest.py
@@ -318,8 +318,7 @@ def _ensure_admin_logged_in(page: Page, base_url: str) -> None:
         raise
 
     # Wait for JS initialization (showTab + HTMX + event delegation) before any tab clicks.
-    # evaluate()-based poll: wait_for_function's eval() mechanism is rejected by strict
-    # CSP (script-src 'self', no unsafe-eval) right after navigation.
+    # See wait_for_js_condition() docstring for why evaluate()-based polling is used here.
     try:
         wait_for_js_condition(
             page,
--- a/tests/playwright/entities/test_resources.py
+++ b/tests/playwright/entities/test_resources.py
@@ -8,7 +8,7 @@ CRUD tests for Resources entity in ContextForge Admin UI.
 """
 
 # Local
-from ..pages.admin_utils import delete_resource, find_resource
+from ..pages.admin_utils import delete_resource, find_resource, wait_for_entity_deleted
 from ..pages.resources_page import ResourcesPage
 
 
@@ -50,5 +50,5 @@ class TestResourcesCRUD:
         # Delete using API helper
         assert delete_resource(resources_page.page, created_resource["id"])
 
-        # Verify deletion
-        assert find_resource(resources_page.page, test_resource_data["name"]) is None
+        # Verify deletion (retry to handle DB commit propagation lag)
+        assert wait_for_entity_deleted(resources_page.page, "resources", test_resource_data["name"]), f"Resource '{test_resource_data['name']}' still exists after deletion"
--- a/tests/playwright/pages/admin_utils.py
+++ b/tests/playwright/pages/admin_utils.py
@@ -10,15 +10,45 @@ Shared utility functions for admin page interactions.
 # Standard
 import logging
 import os
+import time
+from typing import Union
 import urllib.parse
 
 # Third-Party
-from playwright.sync_api import Page
+from playwright.sync_api import Frame, Page
 
 # Local
 from mcpgateway.admin import ADMIN_CSRF_COOKIE_NAME, ADMIN_CSRF_HEADER_NAME
 
 logger = logging.getLogger(__name__)
+
+
+def wait_for_js_condition(target: Union[Page, Frame], expression: str, timeout: int = 30000, polling: int = 100) -> None:
+    """Poll a boolean JS expression via ``evaluate()`` until it is truthy.
+
+    ``Page.wait_for_function``/``Frame.wait_for_function`` compile their predicate with
+    ``eval()``-equivalent semantics, which strict CSP (``script-src 'self'``, no
+    ``unsafe-eval``) rejects immediately with an ``EvalError`` right after navigation.
+    ``evaluate()`` is not subject to that restriction, so poll manually instead.
+
+    Args:
+        target: Page or Frame to evaluate the expression against.
+        expression: JS boolean expression to poll.
+        timeout: Maximum time to wait in milliseconds.
+        polling: Interval between polls in milliseconds.
+
+    Raises:
+        TimeoutError: If the expression never evaluates truthy within ``timeout``.
+    """
+    deadline = time.monotonic() + timeout / 1000
+    while time.monotonic() < deadline:
+        try:
+            if target.evaluate(expression):
+                return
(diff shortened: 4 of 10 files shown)


marekdano (changes requested):
## Blocking

### The token test can now pass vacuously

In `test_revoke_button_not_shown_for_revoked_token`, the loop breaks as soon as `token_card_count == 0`, and a zero card count routes into the `else` branch, which only re-checks revocation via the API. The UI assertion — the entire point of the test, per its own comment "Verify the token card IS rendered (not vacuously absent)" — is skipped.

That escape hatch existed before, but this PR makes it more likely to fire: the settle wait after clicking the inactive toggle drops from 2000ms to 500ms. If the list hasn't re-rendered yet, `token_card_count` is 0, the loop breaks on the first pass, and the test silently degrades into an API check that would pass even if the revoke button were rendered.

Worth waiting for the token card to actually appear (`expect(token_card).to_be_visible()`, or a `wait_for_js_condition` on the list) before reading either count, so that a not-yet-rendered list is distinguishable from a genuinely absent one.

## Suggestions

### Cache-bust rather than sleep past the cache

The `/admin` cache key includes `$is_args$args`, and `wait_for_entity_deleted()` already exploits exactly that with `?cache_bust=del{attempt}` (`admin_utils.py:110`). Appending a cache-bust param to the tokens fetch would be deterministic and would remove up to 8 seconds of wall-clock retrying from the suite.

The current loop also calls `page.reload()` at the *end* of each iteration, while `navigate_to_tokens_tab()` only clicks the sidebar tab (no refetch) — so the first iteration inspects whatever DOM was already there. Reloading at the top of the loop would be clearer.

### Don't swallow the exception in `wait_for_js_condition`

The bare `except Exception: pass` means a typo'd expression, a detached frame, and a legitimately-unmet condition all look identical: a 30-second wait ending in `Condition not met`. Capturing the last exception and chaining it (`raise TimeoutError(...) from last_exc`) would save real debugging time.

### Raising the builtin `TimeoutError` is a little surprising

Since Python 3.10 it's an `OSError` subclass, and it silently diverges from the `PlaywrightTimeoutError` that every other wait in this suite raises. It works, and the call sites here were updated correctly, but a future caller copying an existing `except PlaywrightTimeoutError:` pattern will miss it. Re-raising Playwright's own `TimeoutError` would keep the suite uniform.

### Consistency across the suite

About 35 `wait_for_function` calls remain (`conftest.py:321`, `pages/gateways_page.py`, `pages/servers_page.py`, `test_admin_ui.py`, `security/test_iframe_embedding_security.py`, …). If the CSP diagnosis is right, those are latent failures too. Either they don't currently trip it (a sentence in the PR explaining why would help), or they should migrate — a follow-up issue would be fine rather than expanding this PR.

## Minor notes

- `Union[Page, Frame]` → `Page | Frame`; the project is Python ≥3.11 and the same file already uses `dict[str, str]`.
- The loop checks the deadline *before* evaluating, so if the condition becomes true during the final `wait_for_timeout` it still raises. One last `evaluate()` after the loop would close that.
- A generic JS-polling util in `pages/admin_utils.py` is slightly off-shelf — `tests/playwright/utils.py` might be a more natural home, given it's imported by four non-page modules.

marekdano (changes requested):
Thanks for the quick turnaround on this — the CSP diagnosis is spot-on and the `wait_for_js_condition` helper is a clean fix. I reproduced the original failures on `main` against the strict-CSP build (`script-src 'self'`, no `unsafe-eval`) and confirmed this branch turns them green, so the core of the PR is solid. The follow-up commit addressed every earlier point well:

- `Page | Frame` instead of `Union`
- last exception captured and chained (`raise … from last_exc`)
- helper now raises `PlaywrightTimeoutError`, with the three call sites updated to match
- reload moved to the top of the loop
- loop restructured so the final `evaluate()` still runs at the deadline

## 🔴 One blocking item: the revoke-button test still passes vacuously

`test_revoke_button_not_shown_for_revoked_token` is green, but I instrumented it on this commit and the UI assertion never actually runs — the token card is never rendered, so it falls through to the API-only branch:

```
PROBE final: card=0 revoke=0 -> branch=API-ONLY (VACUOUS)
```

The root cause is the toggle selector rather than timing. `#tokens-inactive-toggle` doesn't exist, and `[name='include_inactive']` matches **7 checkboxes** — one per entity panel — so `.first` clicks the *servers* toggle:

```
ids = [show-inactive-servers, show-inactive-tools, show-inactive-resources,
       show-inactive-prompts, show-inactive-gateways, show-inactive-tokens,
       show-inactive-a2a-agents]
```

The tokens list never gets `include_inactive=true`, the revoked token isn't listed, and the new `to_be_visible()` guard just times out and is swallowed. Since this is a security regression test, a vacuous pass is worth fixing before merge — it would stay green even if the revoke button came back.

The correct toggle is `#show-inactive-tokens` (index 5 above). I applied just that one line and re-ran:

```
PROBE final: card=1 revoke=0 -> UI-ASSERT   ✅ passes meaningfully
```

```suggestion
            inactive_checkbox = tokens_page.page.locator("#show-inactive-tokens")
```

Optionally, it'd be nice to make the `token_card_count == 0` case fail loudly instead of silently degrading to the API check, so this kind of regression can't hide again.

## 🟡 Suggestion (non-blocking, fine as a follow-up)

There are still ~68 `wait_for_function` calls elsewhere (`conftest.py:321`, most page objects, `test_admin_ui.py`, `security/test_iframe_embedding_security.py`, …). They pass today only because their predicates are already true on first evaluation — the moment one has to poll, it'll hit the same `EvalError`, and the surrounding `except PlaywrightTimeoutError` won't catch it. A tracking issue to migrate them over time would be worth opening rather than expanding this PR.

Happy to re-approve once the selector is pointed at the tokens toggle. Nice work on the CSP fix overall.

ja8zyjits:
No Blocking issues
But since there are already improvements asked for above, you may want to check the following

## 🚨 Blocking Issues

None.

## ⚠️ Warnings

### Documentation Gaps

- [ ] **`wait_for_js_condition` helper undocumented**
  - File: `tests/playwright/pages/admin_utils.py:24-51`
  - Issue: CSP-safe polling helper replaces `wait_for_function` (violates strict CSP). Used in 9 test files, not in `tests/AGENTS.md`.
  - Fix: Add to `tests/AGENTS.md` under "Playwright (Browser E2E) Tests":
    ```markdown
    ### CSP-Safe JS Polling
    
    Strict CSP (`script-src 'self'`, no `unsafe-eval`) blocks `Page.wait_for_function()`. Use `wait_for_js_condition()`:
    
    ```python
    from .pages.admin_utils import wait_for_js_condition
    wait_for_js_condition(page, "typeof window.Admin !== 'undefined'", timeout=30000)
    ```
    ```

- [ ] **`wait_for_entity_deleted` helper undocumented**
  - File: `tests/playwright/pages/admin_utils.py:127-143`
  - Issue: Retry helper for DB commit propagation lag. Used in `test_resources.py`, not documented.
  - Fix: Add to `tests/AGENTS.md` under "Writing New Playwright Tests":
    ```markdown
    ### Handling Async DB Operations
    
    After DELETE, use `wait_for_entity_deleted()` for DB commit lag:
    
    ```python
    from .pages.admin_utils import wait_for_entity_deleted
    assert delete_resource(page, resource_id)
    assert wait_for_entity_deleted(page, "resources", resource_name)
    ```
    ```

## 🔧 Code Quality Issues

### Medium Priority (4)

- [ ] **DRY: Duplicated CSP comments (6 files)**
  - Files: `conftest.py:320-322`, `admin_utils.py:28-31`, `test_innerhtml_guard_iframe.py:413-415`, `test_admin_menu_visibility.py:158-161`, `test_admin_url_context.py:72-74`, `test_organization.py:130-132`
  - Pattern: 3-4 line comment explaining CSP rejection of `wait_for_function()`
  - Fix: Extract to `wait_for_js_condition()` docstring, reference at call sites

- [ ] **DRY: Retry loop pattern duplicated**
  - File: `tests/playwright/security/test_token_lifecycle.py:181-220`
  - Similar to: `wait_for_entity_deleted()` in `admin_utils.py:95-109`
  - Fix: Extract to `wait_for_ui_condition()` helper accepting custom predicate

- [ ] **KISS: Complex iframe frame retrieval**
  - File: `tests/playwright/security/test_innerhtml_guard_iframe.py:48-73`
  - Current: 30-iteration loop with 500ms sleep, skip main_frame, check URL substring
  - Better: Use `page.frame_locator("#admin-frame")` or `page.frames[1]` with validation
  - Impact: 15s worst case

- [ ] **Error handling: Broad exception catch without logging**
  - File: `tests/playwright/pages/admin_utils.py:48-50`
  - Current: `except Exception as exc: last_exc = exc`
  - Issue: Catches KeyboardInterrupt/SystemExit, no logging during retry
  - Fix: Use `except (PlaywrightTimeoutError, Exception)` and log exception details

### Low Priority (2)

- [ ] **Logic: Potential infinite loop in `wait_for_js_condition`**
  - File: `tests/playwright/pages/admin_utils.py:23-55`
  - Issue: If `evaluate()` consistently raises non-timeout exceptions, loop continues until deadline
  - Fix: Add explicit exception counter or re-raise after N consecutive failures
  - Severity: Low (timeout provides eventual exit)

- [ ] **Error handling: Silent exception swallowing in iframe helper**
  - File: `tests/playwright/security/test_innerhtml_guard_iframe.py:68-70`
  - Current: `except Exception: pass` in frame retrieval loop
  - Fix: Log exception or use specific exception types

marekdano (commented):
Thanks for the follow-up commit — nice work. I re-ran these against the strict-CSP build and they check out:

- Moving `_ensure_admin_logged_in` onto `wait_for_js_condition` is the highest-value one, since almost every UI test goes through it — appreciate you picking that up.
- The `page.main_frame` skip in `_get_admin_frame` is a good catch; the stale `.url` after `set_content()` is exactly the kind of thing that hands back the wrong frame silently.
- The lazy-load waits before the metrics/chat tests are the right call. Metrics-retry, chat-selection, and the iframe-metrics tests all pass for me now.

Just one thing still outstanding from the earlier review, which I think got lost in the rebase: `test_revoke_button_not_shown_for_revoked_token` is still passing vacuously. On this commit it never renders the token card and falls through to the API-only branch:

```
card=0 revoke=0 -> API-ONLY (VACUOUS)
```

The cause is the toggle selector — `[name='include_inactive']` matches all seven per-panel checkboxes and `.first` grabs the servers one, so the tokens list never gets `include_inactive`. Pointing it at `#show-inactive-tokens` makes the UI assertion actually run (`card=1 revoke=0`). Since it's a security regression test, worth closing out before merge:

```suggestion
            inactive_checkbox = tokens_page.page.locator("#show-inactive-tokens")
```

Happy to approve once that's in.

gcgoncalves:
@ja8zyjits addressed.

ja8zyjits:
@gcgoncalves Please check them and if they are invalid, let me know I will approve right away.

## 🚨 Blocking Issues

### Critical

1. **Infinite loop risk in `wait_for_js_condition()`**
   - File: `tests/playwright/pages/admin_utils.py:45-57`
   - Issue: `while True` loop with no exit if `timeout <= 0`
   - Fix: Add validation `if timeout <= 0: raise ValueError("timeout must be positive")`

2. **Infinite loop risk in `wait_for_ui_condition()`**
   - File: `tests/playwright/pages/admin_utils.py:87-95`
   - Issue: `while True` loop when both `retries` and `deadline_seconds` are None
   - Fix: Add validation `if retries is None and deadline_seconds is None: raise ValueError("Must provide either retries or deadline_seconds")`

### Minor

3. **Missing return type hint**
   - File: `tests/playwright/pages/admin_utils.py:27`
   - Function: `wait_for_js_condition()`
   - Fix: Add `-> None` to signature

4. **Function attribute as mutable state**
   - File: `tests/playwright/pages/admin_utils.py:186-196`
   - Issue: `_not_found.calls = 0` persists across calls
   - Fix: Use closure variable `calls = [0]` instead

### Suggestions

5. Split `wait_for_ui_condition()` into two simpler functions (retries vs deadline)
6. Use modern union syntax (`|`) instead of `Optional[]`

marekdano (dismissed):
The last commit closes out the blocking item and the refactors are clean. 
Thanks for working through all the rounds on this!!!

LGTM 🚀

gcgoncalves:
@ja8zyjits 

C**ritical #1 (wait_for_js_condition infinite loop):** invalid. `timeout<=0`: deadline computed before loop as `now + timeout/1000`. First iteration always `evaluate()` once, then check `time.monotonic() >= deadline` → true immediately since deadline already passed → breaks, raises. Not infinite, just zero-effective-timeout. No fix needed (though explicit validation still nicer UX).

**Critical #2 (wait_for_ui_condition infinite loop):** valid and addressed.

**Minor: missing `-> None`:** invalid. Line 27 already has `-> None` in signature. Stale/wrong comment.

**Minor: `_not_found.calls` mutable attr:** not a bug. `_not_found` closure redefined fresh each `wait_for_entity_deleted()` call, `.calls=0` reset each time before use. State doesn't leak across outer calls.

gcgoncalves:
Screenshot 2026-07-15 at 16 03 03

ja8zyjits (approved):
LGTM

PR #5626, opened 2026-07-14
feat(teams): support bulk member creation in TeamCreateRequest

### LINKED ISSUE #5238: [FEATURE]: Support bulk member creation in TeamCreateRequest
### 🧭 Type of Feature

### Summary

Allow `POST /teams` to optionally accept an initial list of members in the same request, so a team and its members can be set up atomically from a single form. The server routes each entry to the right underlying flow: direct membership for known users, invitation for unknown emails.

### Current behaviour

`TeamCreateRequest` (`mcpgateway/schemas.py:6380`) accepts only `name`, `slug`, `description`, `visibility`, `max_members`. Adding people to a new team is a separate flow per row, and the right endpoint depends on whether the email already maps to an `EmailUser`:

- `POST /teams/{team_id}/members` — direct add. Raises `UserNotFoundError` if the email is not a known user (`mcpgateway/services/team_management_service.py:856-859`).
- `POST /teams/{team_id}/invitations` — email-based invitation flow for any address; recipient accepts via token.

For a "Create team" form where an admin types in arbitrary addresses (some staff, some external), the client today has to:

1. `POST /teams` → create team
2. For each row, either guess the right endpoint or try one and fall back, then `POST /teams/{id}/members` **or** `POST /teams/{id}/invitations`

### Problem

The N+1 flow turns partial failure into a UX problem. If row three fails after the team and rows one and two succeeded, the UI has to choose between rolling back (delete the team), retrying just the failed rows, or showing a confusing partial-success state.

It also leaks an API distinction (known user vs. not) into the frontend that the admin filling out the form shouldn't have to care about.

### Proposal

Add an optional `members` field to `TeamCreateRequest`. The server inspects each entry and routes it to the appropriate underlying operation:

- If the email matches an active `EmailUser` → create membership directly.
- If not → create a `EmailTeamInvitation` for that email and trigger the existing invite delivery.

```python
class TeamMemberSeed(BaseModel):
    email: EmailStr
    role: Literal["owner", "member"] = "member"

class TeamCreateRequest(BaseModel):
    name: str
    slug: Optional[str] = None
    description: Optional[str] = None
    visibility: Literal["private", "public"] = "private"
    max_members: Optional[int] = None
    members: Optional[List[TeamMemberSeed]] = None  # NEW
```

Server-side, the team and all derived rows (memberships + invitations) are created in a single transaction. On any failure (invalid email, role not permitted, would exceed `max_members`, etc.) the whole thing rolls back and the response identifies the offending row(s).

#### Response shape

The response should make it explicit how each row was resolved, so the form can render an accurate confirmation ("3 added, 2 invited") without re-querying:

```json
{
  "team": { ... TeamResponse ... },
  "members_added": [
    {"email": "alice@example.com", "role": "member"}
  ],
  "invitations_sent": [
    {"email": "external@partner.com", "role": "member", "invitation_id": "inv-..."}
  ]
}
```

### Why option B (auto-route) over two separate arrays

A previous draft considered two parallel fields: one `members` array (direct add) and one `invitations` array (invite by email). Auto-routing through a single `members` field is preferred because:

- The form-filler shouldn't have to know which addresses are already platform users; the server already has that information.
- One field keeps the request shape simple and matches the mental model of "add people to the team".
- The structured response above removes the ambiguity that would otherwise come from mixing two semantics into one input field.

### Why it's worth doing

- Removes the partial-failure UX problem entirely; clients see one atomic success/failure.
- Frontend doesn't need to pre-check whether each email is a known user.
- Backwards-compatible: `members` is optional; existing callers are unaffected.
- Small change: mostly a schema field, a transactional service method that fans out to the existing add-member and create-invitation paths, and a structured response.

### Out of scope

- Changing the single-member `POST /teams/{team_id}/members` or invitation endpoints; both remain as they are.
- Reworking the invitation delivery mechanism itself.

### Context

Surfaced while scoping the "Create team" form for the React UI rewrite on `epic/ui-rewrite`. The two-step API works, but the form-level partial-failure handling (and forcing the form to distinguish known-vs-unknown emails) is the main reason a single-form UX is awkward today.

marekdano:
## Summary

`POST /teams` now accepts an optional `members` list, so a team and its people can be set up from a single form submission. The server decides how each entry is applied rather than making the client work it out: an address belonging to an active user becomes a membership directly, anything else (unknown address, deactivated account) is sent an invitation.

Everything is written in one transaction. A row that can't be applied fails the whole request and leaves nothing behind — no team, no half-populated membership list, no stray invitations. That removes the partial-failure problem the issue describes, and the frontend no longer has to pre-check which addresses are already platform users.

Closes #5238

## Response shape

The issue proposed nesting the team under a `team` key. This uses a **flat superset** instead: every existing `TeamResponse` field stays exactly where it was, with `members_added` and `invitations_sent` added alongside. Existing callers of `POST /teams` — including the HTMX admin form — keep working unchanged, which the issue lists as a goal but which the nested shape would have broken. Both arrays are empty when no members are seeded.

## Behaviour

- **The creator's own row is skipped**, not rejected. They're already made owner by team creation, and the form shouldn't have to know to filter itself out — that's the same kind of leaked API rule the issue is removing.
- **Errors name the offending row.** `TeamMemberSeedError` carries `index`, `email` and `reason`, and the index is the one the caller sent, not the position after the creator's row is dropped: `members[2] (ALICE@example.com): duplicate address, already listed at members[1]`.
- **Validated before anything is written:** duplicate addresses (case-insensitive), capacity (creator + seeds against the team limit), per-user team limits, and whether invitations are enabled.
- **Bounded:** `members` is capped at 500 entries (`MAX_TEAM_MEMBER_SEEDS`) at the request boundary, so an unlimited `MAX_MEMBERS_PER_TEAM` can't be turned into an unbounded request. The real per-team limit still applies underneath.
- **`role` on both response arrays is `Literal["owner", "member"]`**, so the OpenAPI spec exposes it as an enum and generated clients get a typed union rather than a bare string.

## Example

```bash
curl -X POST http://localhost:4444/teams/ \
  -d '{
    "name": "Platform Team",
    "description": "Ships things",
    "visibility": "private",
    "members": [
      {"email": "alice@example.com", "role": "member"},
      {"email": "bob@example.com",   "role": "owner"},
      {"email": "external@partner.com", "role": "member"}
    ]
  }'
```

`alice` and `bob` are known users and are added directly; the partner address is invited:

```json
{
  "id": "d46f3a69e7cc4a3ab9e3b1847a7dce7b",
  "name": "Platform Team",
  "slug": "platform-team",
  "description": "Ships things",
  "created_by": "admin@example.com",
  "is_personal": false,
  "visibility": "private",
  "max_members": null,
  "member_count": 3,
  "created_at": "2026-07-14T15:15:31.345835Z",
  "updated_at": "2026-07-14T15:15:31.345837Z",
  "is_active": true,
  "members_added": [
    {"email": "alice@example.com", "role": "member"},
    {"email": "bob@example.com", "role": "owner"}
  ],
  "invitations_sent": [
    {"email": "external@partner.com", "role": "member",
     "invitation_id": "2c81db3f003e41a0bcb5212a11c8a60e"}
  ]
}
```

A rejected row returns `400` and creates nothing:

```json
// alice listed twice
{"detail": "members[2] (ALICE@example.com): duplicate address, already listed at members[1]"}

// creator + 2 members against max_members=2
{"detail": "Team would start with 3 members, exceeding the maximum of 2"}
```

Omitting `members` returns the same flat team object it always did, with both arrays empty.

## Incidental fix: invitations outliving a deleted team

`delete_team` deactivated memberships but left **invitations** active. Because creating a team whose slug matches a deleted one reactivates that row, this had two consequences:

1. Deleting a team and recreating it with the same name and people failed outright (`An active invitation already exists for …`) — the most ordinary recovery path for the new create-team form.
2. More seriously, an invitation to the *old* team stayed live and became a valid invitation to the *new* team once the slug was reused, letting someone join a team they were never invited to. `accept_invitation` checks the team is active, so it was inert while deleted — but it woke up on reactivation. This is reachable on `main` today, independent of this feature.

`delete_team` now deactivates pending invitations alongside memberships.

## Changes

In `mcpgateway/schemas.py`: adds `TeamMemberSeed`, the `members` field on `TeamCreateRequest`, `TeamCreateResponse` extending `TeamResponse`, and the `MAX_TEAM_MEMBER_SEEDS` ceiling.

In `mcpgateway/services/team_management_service.py`: adds `create_team_with_members()`, which does the transactional seeding, with `create_team()` delegating to it and keeping its existing signature; adds `TeamMemberSeedError`; and deactivates pending invitations in `delete_team()`.

In `mcpgateway/services/team_invitation_service.py`: `create_invitation()` gains `commit=False` so invitations can join an outer transaction, and no longer rolls back a transaction it doesn't own.

In `mcpgateway/routers/teams.py`: passes `members` through and returns the seeding outcome.

`tests/unit/mcpgateway/services/test_team_member_seeding.py` is new, covering routing, the creator-skip, row-identifying errors, capacity, rollback, transaction ownership, the reactivation lifecycle, and backwards compatibility. Existing router tests were retargeted from `create_team` to `create_team_with_members`, which is the method the router now calls.

## Notes for review

- **`invitations_sent` is a slight misnomer, inherited from the issue.** Nothing in the codebase sends email — `create_invitation` writes a row and a token, and invitees discover invitations by calling `GET /teams/invitations`. Behaviour matches the existing single-invite endpoint exactly, but the issue's "trigger the existing invite delivery" describes a mechanism that doesn't exist. Happy to rename the field if you'd prefer something truthful.
- **Out of scope:** creating a team whose name collides with an *active* team still returns `500` (`UNIQUE constraint failed: email_teams.slug`) rather than `409`. Pre-existing on `main`; the new form will hit it, so it likely deserves its own issue.
- The React create-team form on `epic/ui-rewrite` doesn't consume this yet

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1571,6 +1571,11 @@ VALIDATE_TOKEN_ENVIRONMENT=true
 # MAX_TEAMS_PER_USER=50
 # MAX_MEMBERS_PER_TEAM=100
 
+# Hard ceiling on how many members can be seeded in a single POST /teams request
+# (the `members` array). Validated at the request boundary before any write; the
+# per-team MAX_MEMBERS_PER_TEAM limit still applies underneath.
+# MAX_TEAM_MEMBER_SEEDS=500
+
 # Team Invitation Settings
 # INVITATION_EXPIRY_DAYS=7
 # REQUIRE_EMAIL_VERIFICATION_FOR_INVITES=true
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -1,6 +1,6 @@
 {
   "exclude": {
-    "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^.secrets.baseline$",
+    "files": "(?x)(package-lock\\.json$|Cargo\\.lock$|uv\\.lock$|go\\.sum$|mcpgateway/sri_hashes\\.json$)|^.secrets.baseline$",
     "lines": null
   },
   "generated_at": "2026-07-17T13:06:41Z",
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -1,6 +1,6 @@
 {
   "exclude": {
-    "files": "(?x)(package-lock\\.json$|Cargo\\.lock$|uv\\.lock$|go\\.sum$|mcpgateway/sri_hashes\\.json$)|^.secrets.baseline$",
+    "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^.secrets.baseline$",
     "lines": null
   },
   "generated_at": "2026-07-17T13:06:41Z",
@@ -2060,7 +2060,7 @@
         "hashed_secret": "91174c4e6859da80df4e7ce9d45b35501f591210",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1432,
+        "line_number": 1558,
         "type": "Base64 High Entropy String",
         "verified_result": null
       },
@@ -2068,7 +2068,7 @@
         "hashed_secret": "4a0a2df96d4c9a13a282268cab33ac4b8cbb2c72",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1473,
+        "line_number": 1599,
         "type": "Secret Keyword",
         "verified_result": null
       }
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -2060,7 +2060,7 @@
         "hashed_secret": "91174c4e6859da80df4e7ce9d45b35501f591210",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1558,
+        "line_number": 1559,
         "type": "Base64 High Entropy String",
         "verified_result": null
       },
@@ -2068,7 +2068,7 @@
         "hashed_secret": "4a0a2df96d4c9a13a282268cab33ac4b8cbb2c72",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1599,
+        "line_number": 1600,
         "type": "Secret Keyword",
         "verified_result": null
       }
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -2140,7 +2140,7 @@
         "hashed_secret": "a2166e0b243f4e192e14000a6aa8f46cde6bfc20",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 820,
+        "line_number": 821,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -2148,7 +2148,7 @@
         "hashed_secret": "fa9beb99e4029ad5a6615399e7bbae21356086b3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1327,
+        "line_number": 1328,
         "type": "Basic Auth Credentials",
         "verified_result": null
       },
@@ -2156,7 +2156,7 @@
         "hashed_secret": "fa9beb99e4029ad5a6615399e7bbae21356086b3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1331,
+        "line_number": 1332,
         "type": "Secret Keyword",
         "verified_result": null
       }
--- a/docs/docs/manage/api-usage.md
+++ b/docs/docs/manage/api-usage.md
@@ -1407,12 +1407,12 @@ Each `TeamMemberSeed` object:
 
 **Configuration notes:**
 
-| Setting | Effect on `POST /teams` |
-|---------|------------------------|
-| `ALLOW_TEAM_CREATION=false` | Non-admin callers receive `403`. Platform admins bypass this flag. |
-| `ALLOW_TEAM_INVITATIONS=false` | Seeded addresses that would normally become invitations instead fail the whole request with `400`. |
-| `MAX_TEAM_MEMBER_SEEDS=500` | Hard ceiling on the `members` array length (validated before any write). |
-| `MAX_MEMBERS_PER_TEAM` | Seed count + creator must not exceed this limit (or the per-team `max_members` override). |
+| Setting | Default | Effect on `POST /teams` |
+|---------|---------|------------------------|
+| `ALLOW_TEAM_CREATION` | `true` | When `false`, non-admin callers receive `403`. Platform admins bypass this flag. |
+| `ALLOW_TEAM_INVITATIONS` | `true` | When `false`, seeded addresses that would normally become invitations instead fail the whole request with `400`. |
+| `MAX_TEAM_MEMBER_SEEDS` | `500` | Hard ceiling on the `members` array length (validated before any write). |
+| `MAX_MEMBERS_PER_TEAM` | `100` | Seed count + creator must not exceed this limit (or the per-team `max_members` override). |
 
 **Minimal create (no members):**
 
@@ -1443,7 +1443,7 @@ curl -s -X POST \
 
 **Response (`TeamCreateResponse`):**
 
-`TeamCreateResponse` is a superset of `TeamResponse`. All `TeamResponse` fields are present; two extra arrays report how each seeded member was resolved. Both arrays are always present (empty when no members were seeded), so clients that only care about the team itself can ignore them.
+`TeamCreateResponse` extends `TeamResponse` through Pydantic subclassing, so it is a strict superset: every `TeamResponse` field is present unchanged, plus two extra arrays that report how each seeded member was resolved. Both arrays are always present (empty when no members were seeded), so clients that only care about the team itself can ignore them.
 
 ```json
 {
--- a/docs/docs/manage/api-usage.md
+++ b/docs/docs/manage/api-usage.md
@@ -1403,6 +1403,7 @@ Each `TeamMemberSeed` object:
 - Email addresses are normalised to lowercase before lookup; `ALICE@example.com` and `alice@example.com` are treated as the same address.
 - Duplicate addresses (case-insensitive) in a single request are rejected with a row-indexed error.
 - The whole operation is **atomic**: a failure on any seed rolls back the team and all prior seeds.
+- Deleting a team also deactivates its **pending invitations**, so an invitation cannot outlive (or be revived alongside) the team it points at.
 
 **Configuration notes:**
 
--- a/docs/docs/manage/api-usage.md
+++ b/docs/docs/manage/api-usage.md
@@ -1367,7 +1367,135 @@ curl -s -X DELETE -H "Authorization: Bearer $TOKEN" \
 
 ## Team Management
 
-Teams are the unit of multi-tenancy in ContextForge. The `GET /teams` endpoint lists the teams visible to the caller and is the same endpoint that backs the Admin UI Teams page and the team switcher.
+Teams are the unit of multi-tenancy in ContextForge. Endpoints require a valid Bearer token and the caller must hold the `teams.create` or `teams.read` RBAC permission respectively.
+
+### Create Team
+
+```bash
+POST /teams
+Authorization: Bearer $TOKEN
+Content-Type: application/json
+```
+
+**Request body (`TeamCreateRequest`):**
+
+| Field | Type | Required | Default | Description |
+|-------|------|----------|---------|-------------|
+| `name` | string | ✓ | – | Display name (1–255 chars, letters/numbers/spaces/`_`/`.`/`-`). |
+| `slug` | string | – | auto-generated | URL-friendly identifier (`[a-z0-9-]+`, 2–255 chars). |
+| `description` | string | – | `null` | Team description (max 1 000 chars). |
+| `visibility` | `"private"` \| `"public"` | – | `"private"` | Who can see the team. |
+| `max_members` | int ≥ 1 | – | global setting | Per-team member cap. Omit to inherit `MAX_MEMBERS_PER_TEAM`. |
+| `members` | array of `TeamMemberSeed` | – | `null` | Up to `MAX_TEAM_MEMBER_SEEDS` (500) addresses to seed. |
+
+Each `TeamMemberSeed` object:
+
+| Field | Type | Required | Default | Description |
+|-------|------|----------|---------|-------------|
+| `email` | string (email) | ✓ | – | Address to add or invite. Normalised to lowercase. |
+| `role` | `"owner"` \| `"member"` | – | `"member"` | Role to assign. |
+
+**Seed routing rules (server-decided, not caller-controlled):**
+
+- An address that matches an **active** `EmailUser` record is added as a direct member.
+- Any other address (unknown, or known but deactivated) receives an invitation.
+- The creator's own address is silently skipped — team creation already makes the creator an owner.
+- Email addresses are normalised to lowercase before lookup; `ALICE@example.com` and `alice@example.com` are treated as the same address.
+- Duplicate addresses (case-insensitive) in a single request are rejected with a row-indexed error.
+- The whole operation is **atomic**: a failure on any seed rolls back the team and all prior seeds.
+
+**Configuration notes:**
+
+| Setting | Effect on `POST /teams` |
+|---------|------------------------|
+| `ALLOW_TEAM_CREATION=false` | Non-admin callers receive `403`. Platform admins bypass this flag. |
+| `ALLOW_TEAM_INVITATIONS=false` | Seeded addresses that would normally become invitations instead fail the whole request with `400`. |
+| `MAX_TEAM_MEMBER_SEEDS=500` | Hard ceiling on the `members` array length (validated before any write). |
+| `MAX_MEMBERS_PER_TEAM` | Seed count + creator must not exceed this limit (or the per-team `max_members` override). |
+
+**Minimal create (no members):**
+
+```bash
+curl -s -X POST \
+  -H "Authorization: Bearer $TOKEN" \
+  -H "Content-Type: application/json" \
+  -d '{"name": "Engineering", "visibility": "private"}' \
+  $BASE_URL/teams | jq '.'
+```
+
+**Create with seeded members:**
+
+```bash
+curl -s -X POST \
+  -H "Authorization: Bearer $TOKEN" \
+  -H "Content-Type: application/json" \
+  -d '{
+    "name": "Engineering",
+    "visibility": "private",
+    "members": [
+      {"email": "alice@example.com", "role": "owner"},
+      {"email": "external@partner.com"}
+    ]
+  }' \
+  $BASE_URL/teams | jq '.'
+```
+
+**Response (`TeamCreateResponse`):**
+
+`TeamCreateResponse` is a superset of `TeamResponse`. All `TeamResponse` fields are present; two extra arrays report how each seeded member was resolved. Both arrays are always present (empty when no members were seeded), so clients that only care about the team itself can ignore them.
+
+```json
+{
+  "id": "team-abc123",
+  "name": "Engineering",
+  "slug": "engineering",
+  "description": null,
+  "created_by": "admin@example.com",
+  "is_personal": false,
+  "visibility": "private",
+  "max_members": null,
+  "member_count": 2,
+  "created_at": "2026-01-15T10:00:00Z",
+  "updated_at": "2026-01-15T10:00:00Z",
+  "is_active": true,
+  "members_added": [
+    {"email": "alice@example.com", "role": "owner"}
+  ],
+  "invitations_sent": [
+    {"email": "external@partner.com", "role": "member", "invitation_id": "inv-xyz789"}
+  ]
+}
+```
+
+`members_added` entries (`SeededMemberResponse`):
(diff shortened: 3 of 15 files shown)


Lang-Akshay (changes requested):
Thanks for the PR @marekdano 
Please address following blocking findings:

 ### Mcpgateway/services/team_management_service.py:671-674
 - EmailStr preserves uppercase local parts (ALICE@example.com), but EmailUser records are canonical lowercase.                             - _seed_members does an exact-case EmailUser.email.in_(...) query + known_users.get(seed.email) — an existing active user with mixed-case
  input is treated as unknown and falls through to the invitation path.
 - The invitation stores the uppercase address, but accept_invitation() (line 326) exact-compares against the authenticated lowercase       identity (route line 872) → the invitation can never be accepted.
- Required fix: normalize seed emails with strip().lower() before lookup/membership/invitation creation (or case-insensitive lookup
 persisting the canonical email), plus an uppercase-address regression te

ja8zyjits:
## Blocking Issues

### 1. `POST /teams` endpoint not documented
**Status:** ❌ Blocking

**Finding**
`docs/docs/manage/api-usage.md` documents `GET /teams`, but not `POST /teams`.

**Impact**
- Team creation API not discoverable
- New `members` request field undocumented
- No examples for seeded team creation
- New behaviour hidden from API consumers

**Required action**
Add `POST /teams` section to `docs/docs/manage/api-usage.md` with:
- endpoint and auth requirements
- full `TeamCreateRequest` schema
- examples for basic create and create-with-members
- error examples
- behaviour notes

### 2. New response schemas not documented
**Status:** ❌ Blocking

**Finding**
Docs do not describe:
- `TeamCreateResponse`
- `SeededMemberResponse`
- `SeededInvitationResponse`
- `members_added`
- `invitations_sent`

**Impact**
- Clients cannot rely on response contract
- Flat superset compatibility not explained
- Empty-array behaviour undocumented

**Required action**
Document response shape in API guide, including:
- `TeamResponse` compatibility
- `members_added` and `invitations_sent`
- example success responses with and without seeded members

---

## Warnings

### 3. Configuration limits and flags not cross-referenced
**Status:** ⚠️ Warning

**Finding**
Docs do not connect team creation flow to:
- `MAX_TEAM_MEMBER_SEEDS=500`
- `max_members_per_team`
- `allow_team_creation`
- `allow_team_invitations`

**Why it matters**
Users will not understand hard seed ceiling, capacity rules, or invitation-related governance failures.

**Suggested fix**
Add short configuration note in team creation docs with link to full configuration reference.

### 4. Error response examples missing
**Status:** ⚠️ Warning

**Finding**
Feature adds row-specific errors, but docs show no examples.

**Missing examples**
- duplicate email with row index
- capacity exceeded
- invalid role
- invitations disabled

**Why it matters**
Clients cannot build reliable error handling for bulk requests.

### 5. Behavioural nuances not documented
**Status:** ⚠️ Warning

**Finding**
Important behaviour exists only in code/tests:
- creator email auto-skipped
- duplicate detection case-insensitive
- inactive users invited, not added
- operation atomic
- pending invitations deactivated on team deletion

**Why it matters**
Behaviour may surprise API consumers without explicit notes.

### 6. Security deny-path coverage incomplete
**Status:** ⚠️ Warning

**Missing tests**
- unauthenticated create with `members`
- insufficient permission create with `members`
- feature-flag-disabled create with `members`

**Why it matters**
AGENTS.md requires deny-path regression coverage for security-sensitive changes.

### 7. Integration coverage gaps remain
**Status:** ⚠️ Warning

**Missing tests**
- seeded member RBAC role assignment end-to-end
- auth cache invalidation after seeded membership changes
- router-level empty `members=[]` contract
- exact max-members boundary at router level

**Why it matters**
Core transaction logic covered well, but cross-service behaviour still lightly tested.

---

ja8zyjits:
HI @marekdano 

These are non blocking.
You can invalidate this, just let me know
I will approve it

## ⚠️ Warnings

### Documentation Completeness

- [ ] **Missing `ALLOW_TEAM_CREATION` default**
  - File: `docs/docs/manage/api-usage.md:1412`
  - Code: `mcpgateway/config.py:1047` (default=True)
  - Fix: Add `ALLOW_TEAM_CREATION=true` (default) to config table

- [ ] **Missing `MAX_MEMBERS_PER_TEAM` default**
  - File: `docs/docs/manage/api-usage.md:1415`
  - Code: `mcpgateway/config.py:1042` (default=100)
  - Fix: Add `MAX_MEMBERS_PER_TEAM=100` (default) to config table

- [ ] **Schema inheritance unclear**
  - File: `docs/docs/manage/api-usage.md:1447`
  - Code: `mcpgateway/schemas.py:6838-6841`
  - Optional: Note `TeamCreateResponse` extends `TeamResponse` via Pydantic inheritance

### Test Coverage Gaps

- [ ] **No integration test for RBAC side effects**
  - File: `mcpgateway/services/team_management_service.py:875-889`
  - Current: Unit test mocks `_assign_team_rbac_role`
  - Need: Integration test verifying RBAC grants persist

- [ ] **No E2E HTTP test**
  - Files: `mcpgateway/routers/teams.py:82-159`, service layer
  - Current: Router mocks service, service tests skip router
  - Need: Full HTTP POST → response → DB validation

- [ ] **Missing boundary test**
  - File: `mcpgateway/services/team_management_service.py:591`
  - Current: Tests MAX+1 (line 289)
  - Need: Test exactly 500 seeds succeeds

- [ ] **No audit trail verification**
  - File: `mcpgateway/services/team_management_service.py:887`
  - Need: Verify audit log entries for seeded members

- [ ] **No cache invalidation integration test**
  - File: `mcpgateway/services/team_management_service.py:891-900`
  - Current: Mocked (line 545)
  - Need: Verify cache entries cleared

- [ ] **No concurrent slug collision test**
  - File: `mcpgateway/services/team_management_service.py:850-875`
  - Need: Test concurrent POST /teams with identical names

- [ ] **No invitation token validation**
  - File: `mcpgateway/services/team_invitation_service.py:254`
  - Need: Verify token non-empty, URL-safe, unique

### Code Quality

- [ ] **DRY violation: membership reactivation duplicated**
  - Files: `team_management_service.py:1128-1166` (`_upsert_membership`), `850-875` (creator)
  - Severity: Medium - maintenance burden
  - Fix: Use `_upsert_membership` for creator too

- [ ] **Theoretical race condition**
  - File: `mcpgateway/services/team_management_service.py:677`
  - Issue: User `is_active` could change between batch lookup and membership creation
  - Severity: Low - small timing window, transaction rollback protection
  - Status: Acceptable

ja8zyjits (changes requested):
## ⚠️ Warnings

### Config Constant Not Exposed as Environment Variable

**File**: `docs/docs/manage/api-usage.md:1393`, `mcpgateway/schemas.py:6523`

Docs list `MAX_TEAM_MEMBER_SEEDS` in config table alongside `ALLOW_TEAM_CREATION`, `ALLOW_TEAM_INVITATIONS`, `MAX_MEMBERS_PER_TEAM`, implying env var. Actually hardcoded constant in `schemas.py:6523` (`MAX_TEAM_MEMBER_SEEDS = 500`), not `Settings` field.

**Impact**: Operators may try setting as env var, will be ignored.

**Fix**: Update docs table to clarify hard-coded limit vs configurable setting.
Also add it to the `.env.example` file

marekdano:
Thanks @ja8zyjits — good catch. Rather than just documenting it as hard-coded, I've promoted `MAX_TEAM_MEMBER_SEEDS` to a real configurable setting so the config table is accurate and operators can actually tune it:

- Added `max_team_member_seeds` to `Settings` in `mcpgateway/config.py` (default `500`).
- `schemas.py` now sources the module constant from `settings.max_team_member_seeds` (resolved at import), so the `members` field ceiling and the service-layer boundary check both follow the env var.
- Added `MAX_TEAM_MEMBER_SEEDS=500` (commented) to `.env.example`.
- Listed it in the canonical config reference (`docs/docs/manage/configuration.md`) alongside the other team limits; the `POST /teams` table in `api-usage.md` already referenced it and is now accurate.

The per-team `MAX_MEMBERS_PER_TEAM` limit still applies underneath. Existing seeding tests pass, including the exact-`MAX+1` boundary and the env-example consistency check.

ja8zyjits (dismissed):
lgtm

Lang-Akshay (approved):
LGTM ✅

Lang-Akshay (changes requested):
Thanks for the PR @marekdano 
## 1. Team recreation can reuse another tenant's state

**Finding/context — real, conditional:** `create_team_with_members()` reactivates an inactive team by slug, reuses its ID, overwrites `created_by`, and makes the new caller an owner (`mcpgateway/services/team_management_service.py:842-856`). Soft deletion leaves team-scoped RBAC, API-token rows, and resources attached to that ID. A non-platform principal with global `teams.create` could inherit the previous tenant's state.

**Targeted fix — recommended:** Do not reactivate inactive teams from the generic create endpoint. Return `409 Conflict` and create a fresh team ID instead. If restore is required, use a separate authorized restore path that revokes old tokens, deactivates old role assignments and memberships, invalidates invitations, and explicitly migrates or removes old team-scoped resources before reassignment.

```python
if existing_team and not existing_team.is_active:
    raise TeamManagementError("A team with this slug was previously deleted")
```

## 2. Historical invitation tokens can revive after reactivation

**Finding/context — real residual gap:** Current deletion deactivates invitations, but invitations left active by pre-PR deletions or a deletion race become valid again when the old team row is reactivated. Acceptance checks invitation active/expiry and team active, but not a team incarnation (`team_management_service.py:842-870`; `team_invitation_service.py:285,319-342`).

**Targeted fix:** Before any reactivation, invalidate every existing invitation for that team. Prefer preventing reactivation entirely as described above. For systems that retain restore, add a team-generation/version and require the invitation generation to match during acceptance.

```python
self.db.query(EmailTeamInvitation).filter(
    EmailTeamInvitation.team_id == existing_team.id,
    EmailTeamInvitation.is_active.is_(True),
).update(
    {EmailTeamInvitation.is_active: False},
    synchronize_session=False,
)
```

Add a regression test that simulates a pre-existing active invitation, recreates the team, and verifies the old token is rejected.

## 3. Delete/accept-invitation race can restore membership

**Finding/context — real concurrency hazard:** `_log_team_member_action()` commits the team inactive state before memberships and invitations are bulk-deactivated (`team_management_service.py:465,483-485,1080-1097`). `accept_invitation()` validates the team before its later membership commit, so concurrent acceptance can insert or reactivate membership during the deletion window.

**Targeted fix:** Make deletion atomic: write history without committing, deactivate the team, memberships, and invitations, then commit once. In `accept_invitation()`, use a transaction-scoped lock on the team and invitation where supported, re-check active state after locking, and use a conditional state guard so acceptance fails if deletion wins the race. Use the same lock order in both paths.

```python
# delete path: history helper must flush/add only here
self._log_team_member_action(..., commit=False)
self.db.query(EmailTeamMember).filter(
    EmailTeamMember.team_id == team_id,
    EmailTeamMember.is_active.is_(True),
).update({EmailTeamMember.is_active: False}, synchronize_session=False)
self.db.query(EmailTeamInvitation).filter(
    EmailTeamInvitation.team_id == team_id,
    EmailTeamInvitation.is_active.is_(True),
).update({EmailTeamInvitation.is_active: False}, synchronize_session=False)
self.db.commit()
```

`accept_invitation()` should lock/re-read the team and invitation before creating membership and abort if either is inactive.

marekdano:
Thanks for the careful security pass on this, @Lang-Akshay — these are all real concerns and worth tracking. I want to flag one thing about scope before we decide how to land the three findings.

All three findings are about the **team reactivation-by-slug** path, and that behavior isn't introduced by this PR — it already lives on `main` in `create_team()` (`team_management_service.py:563-590`). `create_team_with_members()` just preserves the existing behavior. Breaking each down:

1. **Tenant-state reuse on reactivation** — agreed the residual RBAC role assignments, team-scoped tokens, and resources are a genuine gap. But it's a pre-existing property of the reactivation path on `main`, reachable today via `create_team()`. Also worth noting: the recommended fix (return `409`, never reactivate) directly reverses a behavior this PR intentionally relies on — the "delete a team, then recreate it with the same name/people" recovery flow. The incidental invitation fix in this PR exists specifically to make that flow safe. So we can't both keep that recovery path and adopt this fix as written; it's a design decision worth its own discussion.
2. **Historical invitations reviving** — this PR actually closes the forward case: `delete_team` now deactivates pending invitations alongside memberships (lines 1090-1094). The remaining gap is only invitations orphaned by *pre-PR* deletions or a delete/create race, which predates this change.
3. **Delete/accept-invitation race** — a concurrency hazard in the existing `delete_team`/`accept_invitation` code, orthogonal to bulk seeding.

My suggestion: rather than expand this feature PR into a rework of the reactivation lifecycle (which touches token revocation, role-assignment teardown, and locking on accept — and reverses an intentional behavior here), I've opened #5700 to track all three findings and harden the reactivation path independently. That mirrors how we're already tracking the active-slug `500`-vs-`409` gap noted in the description. Does that work as a way to unblock, or do you feel strongly that finding 1. should land here?

Lang-Akshay (approved):
LGTM

ja8zyjits (approved):
LGTM

msureshkumar88 (approved):
## E2E validation — real gateway, real DB, real JWT, no mocks

Ran against the exact PR head commit (`b94725ef1693e5f1f88d970d10934446bf0ee308`) in an isolated worktree. Script committed for reuse: `tests/live_gateway/e2e_pr5626_team_member_seeding.sh`.

**Config used:**
```
DATABASE_URL=sqlite:///<tmpfile>
JWT_SECRET_KEY=<throwaway>
AUTH_REQUIRED=true
MCPGATEWAY_UI_ENABLED=false
ALLOW_TEAM_CREATION=true
ALLOW_TEAM_INVITATIONS=true
MAX_MEMBERS_PER_TEAM=100
MAX_TEAM_MEMBER_SEEDS=500
PLUGINS_ENABLED=false
OBSERVABILITY_ENABLED=false
```

**Commands run:**
- DB init: `python -m mcpgateway.bootstrap_db` (alembic upgrade head + seeds admin user)
- Extra fixture users via `EmailAuthService.create_user(...)`: one active user, one deactivated user
- Server: `python -m uvicorn mcpgateway.main:app --host 127.0.0.1 --port <free-port>`
- Admin JWT: `python -m mcpgateway.utils.create_jwt_token --username admin@example.com --admin --exp 60 --secret $SECRET`
- Real `curl` calls against `POST /teams/`, `DELETE /teams/{id}`, `GET /teams/{id}/invitations`

**Result: 33/33 checks passed**

1. Mixed routing — active user → `members_added`, unknown email → `invitations_sent` with a non-empty `invitation_id`; all `TeamCreateResponse` fields present alongside the new arrays.
2. Creator's own email listed in `members` → `201`, silently skipped, not duplicated and not rejected.
3. Case-insensitive duplicate email → `400`, `detail: "members[1] (...): duplicate address, already listed at members[0]"`; confirmed no team was persisted.
4. Capacity exceeded (`max_members=2` + 2 extra seeds) → `400 "Team would start with 3 members, exceeding the maximum of 2"`, nothing persisted. Boundary case (creator+1==max_members) → `201`, confirming the limit is inclusive.
5. Delete → recreate same name/slug + same member list → second create returns `201` (not the old `400 "An active invitation already exists"`), and exactly one active invitation remains for that email afterward — confirms the bundled `delete_team()` invitation-deactivation fix works.
6. Plain `POST /teams` with no `members` → `201`, `members_added: []` / `invitations_sent: []` both present (not null/missing) — existing callers unaffected.

No discrepancies found between the PR's described behavior and its actual behavior under a live server.

Approving based on this validation plus the earlier static review (implementation, tests, docs all checked out). See follow-up comment for one non-blocking item worth addressing.

msureshkumar88:
## Summary of suggested follow-up (non-blocking)

One item from the static review worth a quick fix, not blocking merge:

**Log sanitization regression** in `team_management_service.py` (the success log after `create_team_with_members`). The previous line sanitized both interpolated values:
```python
logger.info("Created team '%s' by %s", SecurityValidator.sanitize_log_message(team.name), SecurityValidator.sanitize_log_message(created_by))
```
The new version drops sanitization on `created_by`:
```python
logger.info("Created team '%s' by %s (%d members added, %d invited)", SecurityValidator.sanitize_log_message(team.name), created_by, len(memberships), len(invitations))
```
`created_by` is the authenticated caller's email, normally constrained by `EmailStr`, but SSO-provisioned identities may not pass through the same validation, so this isn't purely theoretical. Every other log call touching user-controlled values in this file (`_assign_team_rbac_role`, `_invalidate_membership_caches`, etc.) sanitizes consistently — this looks like an easy one-line oversight rather than a deliberate change. Suggest wrapping `created_by` in `SecurityValidator.sanitize_log_message(...)` to match the file's existing convention.

A few smaller, also non-blocking notes from the review (happy to open follow-up issues if useful):
- With `MAX_TEAM_MEMBER_SEEDS=500`, the seeding loop does a DB commit per member for history logging, a per-member RBAC role lookup, and a per-member cache-invalidation call that redundantly re-invalidates admin stats each time — all fine at small N, but worth batching for large requests.
- `_check_seed_capacity` (pre-write) and `check_team_member_capacity` (used inside `create_invitation`) enforce the same limit with slightly different boundary semantics (`>` vs `>=`); verified they don't conflict in the current flow, but it's duplicate logic worth consolidating eventually.

None of these affect correctness of the feature as validated end-to-end — approving as-is.

Issue #5633, opened 2026-07-15
chore(pyright): resolve standard-mode type errors across the codebase


madhu-mohan-jaishankar:
## Overview

Pyright type checking was added in standard mode, scoped to changed files only per PR (`make pyright-pr`). The full project currently has approximately **6,400 type errors** in standard mode that pre-date the new workflow.

This issue tracks the work to resolve those errors so that the PR check can be upgraded from changed-files-only to whole-project enforcement.

## Summary

The codebase was originally configured with `typeCheckingMode: strict`, which demands annotations on every parameter and return value — producing 252,511 errors that are entirely annotation-coverage complaints, not real type bugs.

Switching to `standard` mode (which uses inference to catch real type mismatches even in unannotated code) surfaces ~6,400 genuine issues:

| Rule | Count | Description |
|------|-------|-------------|
| `reportArgumentType` | ~1,045 | Wrong types passed to functions |
| `reportAttributeAccessIssue` | ~976 | Attribute access on wrong types |
| `reportOptionalMemberAccess` | ~446 | Accessing members on `None`-able values |
| `reportFunctionMemberAccess` | ~274 | Function member access issues |
| `reportOptionalSubscript` | ~113 | Subscripting optional types |
| `reportCallIssue` | ~112 | Calling non-callables |
| `reportPossiblyUnbound` | ~109 | Variables possibly unbound |
| `reportReturnType` | ~66 | Wrong return types |
| Other | ~259 | Various |

Running `make pyright` locally shows the current error count.

## Getting started

```bash
# See current error count
make pyright 2>&1 | tail -1

# Run on a specific directory
make pyright TARGET=mcpgateway/services/
```

Good starting points: `mcpgateway/services/`, `mcpgateway/routes/`, `plugins/`.

PR #5634, opened 2026-07-15
ci(pyright): add PR-scoped type check workflow and pyright-pr make target

madhu-mohan-jaishankar:
## 🔗 Related Issue

Closes [#247](https://github.ibm.com/contextforge-org/internal_issues/issues/247)
Tracked by #5633 (upgrade pyright-pr → full project once errors resolved)

---

## 📝 Summary

Adds a PR-scoped Pyright type-checking gate and supporting infrastructure so that contributors cannot merge code that introduces new type errors, without requiring the pre-existing backlog (~6,400 standard-mode errors) to be fixed first.

**Three changes:**

1. **`pyrightconfig.json`** — switches `typeCheckingMode` from `strict` to `standard` and adds `.cache` to the exclude list.
   - `strict` produced 252,511 errors — all annotation-coverage complaints, not type bugs.
   - `standard` uses inference to catch real type mismatches even in unannotated code.
   - `.cache` was missing from excludes: pre-commit stores virtualenvs there (7,852 vendored Python files), making pyright scan 9,354 files instead of 1,377 and never finish locally.

2. **`Makefile`** — adds `pyright-pr` target: runs pyright only on Python files changed vs the base branch. Pyright still loads the full project for type context, but only reports errors in changed files — preventing new bugs without being blocked by the existing backlog.

3. **`.github/workflows/pyright-check.yml`** — PR-only gate calling `make pyright-pr`. Skips drafts, cancels in-progress runs on new pushes. Must pass before merge (requires "Pyright Type Check" to be added as a required status check in branch protection settings).

   The workflow contains a `NOTE` comment explaining the upgrade path: swap `make pyright-pr` → `make pyright` once #5633 is resolved.

---

## 📏 Reviewability

---

## 🏷️ Type of Change

- [ ] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other

---

## 🧪 Verification

| Check | Command | Status |
|-------|---------|--------|
| pyright-pr exits clean on this branch | `make pyright-pr` | ✅ "no Python files changed — nothing to check" |
| pyright full project (surfaces backlog) | `make pyright 2>&1 \| tail -1` | ✅ runs, ~6,400 existing errors |
| Makefile syntax | `make -n pyright-pr` | ✅ |

---

## ✅ Checklist

---

## 📓 Notes (optional)


Makefile
```diff
@@ -3287,7 +3287,7 @@ LINTERS := isort pylint mypy bandit pydocstyle pycodestyle \
 FILE_AWARE_LINTERS := isort black pylint mypy bandit pydocstyle \
 	pycodestyle ruff pyright vulture markdownlint
 
-.PHONY: lint $(LINTERS) black black-check isort-check ruff-check ruff-fix ruff-format autoflake lint-py lint-yaml lint-json lint-md lint-strict \
+.PHONY: lint $(LINTERS) pyright-pr black black-check isort-check ruff-check ruff-fix ruff-format autoflake lint-py lint-yaml lint-json lint-md lint-strict \
```
cafalchio, line 3295: we do not use black anymore. Could you remove it?

cafalchio (commented):
Should it be added to pre-commit modified files? I am assuming we will start to see a lot of failed CI, so it would be better to fail it locally.

madhu-mohan-jaishankar:
> Should it be added to pre-commit modified files? I am assuming we will start to see a lot of failed CI, so it would be better to fail it locally.

we want to run pyright on all python files once we get rid of the errors. at that point, it would be a minor code change within the workflow itself. if we add it to pre-commit, we have to create a new workflow afterwards.

cafalchio (dismissed):
LGTM

pyrightconfig.json
brian-hussey, line 1: Can we add this to pyproject.toml file instead of creating yet another config file in the project itself, see: https://github.com/microsoft/pyright/blob/main/docs/configuration.md#sample-pyprojecttoml-file

.github/workflows/pyright-check.yml
brian-hussey, line 1: Please add this to the lint-full workflow instead of creating a new workflow for it.

brian-hussey: Or actually would it make more sense to add it to the pre-commit check so that it's available on dev workstation flow as well as part of our existing CI?

brian-hussey (changes requested):
Thanks for doing this @madhu-mohan-jaishankar I just want to tie this into our existing structures a bit more hence the changes requested.

.github/workflows/pyright-check.yml
madhu-mohan-jaishankar: @brian-hussey thanks for the review!!
pre-commit runs only on the changed files. once all the pyright issues are fixed, we need to run it on all python files instead of only the changed ones. so it makes sense to have to as a separate workflow as it would just be a minor change in the workflow already created. if we add it to pre-commit now, we have to create a separate workflow after the pyright fixes.
what are your thoughts?

brian-hussey: You're right, let's hold it off from pre-commit and I think add it to the linting stage then.

brian-hussey (approved):
Thanks for addressing the changes @madhu-mohan-jaishankar, this LGTM

PR #5644, opened 2026-07-15
fix(auth): enforce safe admin demotion rules

### LINKED ISSUE #2808: [BUG][API/UI]: Enforce consistent and safe admin demotion rules
### 🐞 Bug Summary

Admin demotion behavior is inconsistent between peer demotion, self-demotion, the API, and the Admin UI.

Required behavior:

- An administrator can demote another administrator.
- An administrator cannot demote their own account.
- No operation can demote or deactivate the last active administrator.

Current behavior depends on `PROTECT_ALL_ADMINS`:

- With `PROTECT_ALL_ADMINS=true` (default), the API and Admin UI block an administrator from demoting any other administrator.
- With `PROTECT_ALL_ADMINS=false`, the API can allow an administrator to demote their own account when another active administrator exists.
- The Admin UI preserves administrator status during self-edit, so API and UI behavior differ.

Admin demotion rules should be enforced centrally in the service layer and should not depend on the client path.

---

### 🧩 Affected Component

- [x] `mcpgateway` - API
- [ ] Other (explain below)

---

#### Scenario 1: Peer-admin demotion is blocked

1. Run ContextForge with `PROTECT_ALL_ADMINS=true` (default).
2. Create two active administrators: `admin-a@example.com` and `admin-b@example.com`.
3. Authenticate as `admin-a@example.com`.
4. Request demotion of `admin-b@example.com`:

```bash
curl -X PATCH \
  'http://localhost:4444/auth/email/admin/users/admin-b%40example.com' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{"is_admin": false}'
```

5. The API returns `400 Bad Request`:

```json
{"detail":"Admin protection is enabled — cannot demote or deactivate any admin user"}
```

#### Scenario 2: API self-demotion can be allowed

1. Run ContextForge with `PROTECT_ALL_ADMINS=false`.
2. Ensure at least two active administrators exist.
3. Authenticate as `admin-a@example.com`.
4. Send the same `PATCH` request targeting `admin-a@example.com` with `{"is_admin": false}`.
5. The API permits self-demotion because the target is not the last active administrator.

---

### 🤔 Expected Behavior

Admin demotion must follow these rules consistently in the API and Admin UI:

| Request | Expected result |
|---|---|
| Admin A demotes Admin B while another active admin remains | Allow |
| Admin A demotes Admin A | Deny |
| Any demotion or deactivation would leave no active admin | Deny |

Self-demotion and last-active-admin checks should be enforced in `EmailAuthService.update_user()` so every caller receives the same protection.

`PROTECT_ALL_ADMINS` should not cause peer-admin demotion to be rejected. Its remaining purpose and documentation should be clarified separately, including its effect on administrator login lockout.

---

### 📓 Logs / Error Output

Peer-admin demotion with `PROTECT_ALL_ADMINS=true` returns:

```text
400 Bad Request
{"detail":"Admin protection is enabled — cannot demote or deactivate any admin user"}
```

No server exception or stack trace is produced.

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Runtime | `Python 3.12` |
| Platform / OS | `macOS` |
| Container | `Docker` |

---

- Previous implementation attempt: #3247 (closed without merge).
- Keep the last-active-admin check as defense in depth.
- Add deny-path regression tests for API self-demotion, case-insensitive email matching, and last-admin demotion/deactivation.
- Add allow-path regression tests for Admin A demoting Admin B through both API and Admin UI.
- Role-assignment audit records should identify the requesting administrator as `granted_by`, not the target user.

vishu-bh:
# Pull Request

## 🔗 Related Issue

Closes #2808

---

## 📝 Summary

Enforces consistent administrator demotion rules across API and Admin UI:

- allows one administrator to demote another administrator;
- always rejects administrator self-demotion and self-deactivation;
- always protects the last active administrator;
- requires authenticated requester identity for active-admin removal operations;
- records the requesting administrator as `granted_by` during role synchronization.

Root cause was overloaded `PROTECT_ALL_ADMINS` behavior. It controlled both login-lockout bypass and admin demotion, causing peer demotion to be blocked when enabled while allowing API self-demotion when disabled. This change limits the setting to login-lockout bypass and enforces demotion safety independently in `EmailAuthService.update_user()`.

Existing deployments using `PROTECT_ALL_ADMINS=true` will now permit peer-admin demotion. Deployments using `false` will no longer permit self-demotion or self-deactivation. Configuration and Helm descriptions now document the narrowed setting semantics.

---

## 📏 Reviewability

---

## 🏷️ Type of Change

- [x] Bug fix
- [ ] Feature / Enhancement
- [x] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

| Check | Command | Status |
|---|---|---|
| Ruff | `make ruff` | Passed |
| Affected unit modules | `pytest -q tests/unit/mcpgateway/services/test_email_auth_basic.py tests/unit/mcpgateway/services/test_email_auth_service_admin_role_sync.py tests/unit/mcpgateway/routers/test_email_auth_router.py tests/unit/mcpgateway/test_admin.py` | Passed; 3 existing skips |
| Patch whitespace | `git diff --check` | Passed |

---

## ✅ Checklist

---

## 📓 Notes (optional)

`PROTECT_ALL_ADMINS` remains available and defaults to `true`, but now only controls active-admin login-lockout bypass. Self-removal and last-active-admin protections no longer depend on this toggle.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1210,9 +1210,9 @@ VALIDATE_TOKEN_ENVIRONMENT=true
 # SECURITY: Keep this false in production unless you explicitly need public sign-up
 # PUBLIC_REGISTRATION_ENABLED=false
 
-# Admin protection mode
-# When true (default), no admin can be demoted, deactivated, or locked out via API/UI
-# When false, only the last remaining active admin is protected
+# Admin login-lockout protection
+# When true (default), active admin accounts can bypass login lockout
+# Admin self-demotion and last-active-admin protection are always enforced independently
 # PROTECT_ALL_ADMINS=true
 
 # Platform admin user (bootstrap from environment)
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -1922,7 +1922,7 @@
         "hashed_secret": "bc2f74c22f98f7b6ffbc2f67453dbfa99bce9a32",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 454,
+        "line_number": 456,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -1930,7 +1930,7 @@
         "hashed_secret": "a2166e0b243f4e192e14000a6aa8f46cde6bfc20",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 821,
+        "line_number": 823,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -1938,7 +1938,7 @@
         "hashed_secret": "fa9beb99e4029ad5a6615399e7bbae21356086b3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1328,
+        "line_number": 1330,
         "type": "Basic Auth Credentials",
         "verified_result": null
       },
@@ -1946,7 +1946,7 @@
         "hashed_secret": "fa9beb99e4029ad5a6615399e7bbae21356086b3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1332,
+        "line_number": 1334,
         "type": "Secret Keyword",
         "verified_result": null
       }
@@ -2290,7 +2290,7 @@
         "hashed_secret": "f18eecadfbe015aaa19e4efb26ea35e4b22c4716",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 67,
+        "line_number": 84,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -2298,7 +2298,7 @@
         "hashed_secret": "65ccbbc69885ff7d13099428b96bc5a0db135e90",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 87,
+        "line_number": 104,
         "type": "Secret Keyword",
         "verified_result": null
       }
@@ -4650,7 +4650,7 @@
         "hashed_secret": "a0281cd072cea8e80e7866b05dc124815760b6c9",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 17920,
+        "line_number": 7151,
         "type": "Secret Keyword",
         "verified_result": null
       }
--- a/charts/mcp-stack/values.yaml
+++ b/charts/mcp-stack/values.yaml
@@ -828,7 +828,7 @@ mcpContextForge:
 
     # ─ Email-Based Authentication ─
     EMAIL_AUTH_ENABLED: "true"             # enable email-based authentication system
-    PROTECT_ALL_ADMINS: "true"             # prevent any admin from being demoted or deactivated via API/UI
+    PROTECT_ALL_ADMINS: "true"             # allow active admin accounts to bypass login lockout
     PLATFORM_ADMIN_EMAIL: admin@example.com # email for bootstrap platform admin user
     PLATFORM_ADMIN_PASSWORD: changeme     # password for bootstrap platform admin user
     PLATFORM_ADMIN_FULL_NAME: Platform Administrator # full name for bootstrap platform admin
--- a/docs/docs/manage/configuration.md
+++ b/docs/docs/manage/configuration.md
@@ -415,6 +415,8 @@ curl -X POST -H "Authorization: Bearer $TOKEN" \
 | `SMTP_USE_SSL`                | Use implicit SSL/TLS                              | `false`               | bool    |
 | `SMTP_TIMEOUT_SECONDS`        | SMTP timeout in seconds                           | `15`                  | int > 0 |
 
+Changing `PROTECT_ALL_ADMINS` does not control peer-administrator removal. An administrator with `admin.user_management` can demote or deactivate another administrator while at least one active administrator remains. Protect administrator credentials accordingly; deployments requiring dual control should enforce that process outside this endpoint.
+
 When `PASSWORD_RESET_ENABLED=false`, self-service forgot/reset endpoints are disabled (`403` on API and disabled/redirected UI flows).
 When `SMTP_ENABLED=false`, reset requests are accepted but no email is delivered.
 
--- a/docs/docs/manage/configuration.md
+++ b/docs/docs/manage/configuration.md
@@ -32,7 +32,7 @@ These settings are enabled by default for security—only disable for backward c
 | `REQUIRE_JTI` | Require JTI claim in tokens for revocation support | `true` |
 | `REQUIRE_TOKEN_EXPIRATION` | Require exp claim in tokens | `true` |
 | `PUBLIC_REGISTRATION_ENABLED` | Allow public user self-registration | `false` |
-| `PROTECT_ALL_ADMINS` | Prevent any admin from being demoted or deactivated via API/UI | `true` |
+| `PROTECT_ALL_ADMINS` | Allow active admin accounts to bypass login lockout | `true` |
 |`REQUIRE_STRONG_SECRETS`|Enforces strong secret validation. Automatically defaults to true in production to ensure fail-safe deployments.|`true` (prod) / `false` (dev)|
 
 ### ⚙️ Project Defaults (Dev Setup)
@@ -403,7 +403,7 @@ curl -X POST -H "Authorization: Bearer $TOKEN" \
 | `PASSWORD_RESET_RATE_WINDOW_MINUTES` | Password reset rate-limit window        | `15`                  | int > 0 |
 | `PASSWORD_RESET_INVALIDATE_SESSIONS` | Invalidate active sessions on reset     | `true`                | bool    |
 | `PASSWORD_RESET_MIN_RESPONSE_MS` | Minimum forgot-password response duration    | `250`                 | int >= 0 |
-| `PROTECT_ALL_ADMINS`         | Prevent any admin from being demoted or deactivated via API/UI. When false, only the last active admin is protected. | `true` | bool |
+| `PROTECT_ALL_ADMINS`         | Allow active admin accounts to bypass login lockout. Admin self-demotion and last-active-admin protection are always enforced independently. | `true` | bool |
 | `SMTP_ENABLED`                | Enable SMTP notifications for auth emails        | `false`               | bool    |
 | `SMTP_HOST`                   | SMTP host                                         | (none)                | string  |
 | `SMTP_PORT`                   | SMTP port                                         | `587`                 | int     |
--- a/docs/docs/manage/password-management.md
+++ b/docs/docs/manage/password-management.md
@@ -57,6 +57,23 @@ Navigate to `Admin -> Users` (`/admin/#users`):
 
 ## API-Based Admin Reset & Unlock
 
+The user-management endpoints require the `admin.user_management` permission.
+
+### Demote or deactivate an administrator
+
+Use `PATCH /auth/email/admin/users/{email}` for administrator status changes. The service applies the same rules to API and Admin UI requests:
+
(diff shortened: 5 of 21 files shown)


ja8zyjits:
## 🚨 Blocking Issues

### 1. Race Condition - Last Admin Protection
- **File**: `mcpgateway/services/email_auth_service.py:2177-2186`
- **Issue**: `with_for_update()` locks rows but check happens after lock acquisition. Concurrent requests can both see 2 admins, both proceed with demotion
- **Impact**: Last admin protection bypassed under load
- **Fix**: Move lock acquisition to `update_user()` before admin protection checks, or use DB constraint
- **Severity**: High - violates core security invariant

### 2. Missing Integration Test - Role Sync Rollback
- **Files**: `mcpgateway/services/email_auth_service.py:1885-1920`, `mcpgateway/services/role_service.py:527-690`
- **Issue**: Admin demotion uses `commit=False` for role sync. No test verifying actual transaction rollback
- **Needed**: Integration test - trigger exception after role revocation, verify role still active
- **Suggested**: `test_update_user_admin_demotion_rolls_back_role_changes_on_failure()`

### 3. Missing Test - UI Peer-Admin Demotion
- **Files**: `mcpgateway/admin.py:8350-8375`, `mcpgateway/services/email_auth_service.py:1808-1920`
- **Issue**: Issue #2808 requires both API and UI peer-demotion work consistently. No E2E UI test
- **Current**: API tested, UI self-demotion blocked tested
- **Needed**: E2E UI test - Admin A demotes Admin B via form submission
- **Suggested**: `test_admin_update_user_peer_admin_demotion_via_ui()` in `test_admin.py`

### 4. Missing Test - Concurrent Last-Admin Demotion
- **File**: `mcpgateway/services/email_auth_service.py:2177-2186`
- **Issue**: Row-level locking critical for concurrency. Untested locks fail silently
- **Needed**: Test verifying concurrent demotion requests serialize correctly
- **Suggested**: `test_is_last_active_admin_concurrent_demotion_serializes()` using threading/asyncio

### 5. Missing Test - Admin Promotion Role Sync Failure
- **File**: `mcpgateway/services/email_auth_service.py:1885-1920`
- **Issue**: Admin promotion fails if role assignment fails. Only demotion failure tested
- **Needed**: Test verifying promotion fails when `platform_admin` role missing
- **Suggested**: `test_update_user_promote_admin_role_not_found()` in `test_email_auth_service_admin_role_sync.py`

---

## ⚠️ Warnings

### 1. API Docs Gap - Admin Demotion Behavior
- **Files**: `docs/docs/manage/password-management.md:47,63`, `mcpgateway/routers/email_auth.py:841-870`
- **Issue**: `PUT /auth/email/admin/users/{email}` documented only for password reset. New admin demotion rules not documented
- **Needed**: Add section to `docs/docs/manage/api-usage.md` or `docs/docs/manage/rbac.md` showing:
  - Allowed: Admin A demotes Admin B (peer demotion)
  - Blocked: Admin self-demotion (400 error)
  - Blocked: Demoting last active admin (400 error)
  - Required permission: `admin.user_management`

### 2. Broad Exception Catch - Role Sync
- **File**: `mcpgateway/services/email_auth_service.py:1915-1918`
- **Issue**: `except Exception as e:` catches all exceptions. Generic error messages
- **Fix**: Catch specific exceptions (e.g., `RoleServiceError`), provide detailed context
- **Reason**: Debugging difficulty, unclear failure modes

### 3. Transaction Rollback Not Explicit
- **File**: `mcpgateway/services/email_auth_service.py:1915-1918`
- **Issue**: Role sync failure during demotion raises ValueError but transaction state unclear
- **Fix**: Add explicit `self.db.rollback()` before raising, or document transaction ownership
- **Severity**: Medium - could leave inconsistent state

### 4. Missing Edge Case - Empty/Whitespace Requester Email
- **File**: `mcpgateway/services/email_auth_service.py:1845-1857`
- **Current**: Tests missing requester
- **Needed**: Test with `requesting_user_email=""` and `requesting_user_email="   "`
- **Rationale**: Normalization strips whitespace; empty string after strip should fail closed

---

ja8zyjits (changes requested):
check comments

ja8zyjits (dismissed):
LGTM

gandhipratik203:
Good work tightening this flow and centralizing the admin-removal checks.

One issue worth addressing before merge: `EmailAuthService.deactivate_user()` still bypasses the new admin-removal policy. It sets `is_active=False` and commits without requester identity, self-check, or last-active-admin protection. Current UI routes no longer call it, but the service layer is not actually centralized while this public method remains unsafe. See `mcpgateway/services/email_auth_service.py:2012`.

gandhipratik203 (dismissed):
Thanks for the update. The `deactivate_user()` path now delegates through `update_user()`, so the direct service call goes through the same requester identity, self-removal, and last-active-admin protections. The added unit coverage for those cases looks good to me.

Approving.

a-effort (commented):
Nice tightening of the demotion policy. Centralizing in `update_user()` and making `deactivate_user()` delegate through it closes the gaps both earlier reviewers flagged.

Two things worth catching before merge:

**Formatter drift.** `tests/unit/mcpgateway/routers/test_email_auth_router.py:650-653`: kwargs inside `assert_called_with(...)` for the deprecated-PUT branch are at 16-space indent while the siblings above are at 12. Won't survive `black`. Matches the unchecked "Code formatted" box in the PR body.

**Promotion audit-trail fallback.** `role_granter = normalized_requester_email or email` (`email_auth_service.py:1858`) means an admin *promotion* with no requester records `granted_by = target_email` (self-granted). Requester is required for removal but not promotion. Worth making that symmetric, or at minimum logging when the fallback fires so bootstrap/system promotions stay visible in the audit trail.

One question: deactivating an admin leaves the `platform_admin` UserRole row active (the `is_admin is not None` branch is skipped). `count_active_admin_users()` still gates on both flags so the invariant holds, and reactivation restores cleanly. Just confirming that's the intended split between deactivate and demote.

vishu-bh:
> Nice tightening of the demotion policy. Centralizing in `update_user()` and making `deactivate_user()` delegate through it closes the gaps both earlier reviewers flagged.
> 
> Two things worth catching before merge:
> 
> **Formatter drift.** `tests/unit/mcpgateway/routers/test_email_auth_router.py:650-653`: kwargs inside `assert_called_with(...)` for the deprecated-PUT branch are at 16-space indent while the siblings above are at 12. Won't survive `black`. Matches the unchecked "Code formatted" box in the PR body.
> 
> **Promotion audit-trail fallback.** `role_granter = normalized_requester_email or email` (`email_auth_service.py:1858`) means an admin _promotion_ with no requester records `granted_by = target_email` (self-granted). Requester is required for removal but not promotion. Worth making that symmetric, or at minimum logging when the fallback fires so bootstrap/system promotions stay visible in the audit trail.
> 
> One question: deactivating an admin leaves the `platform_admin` UserRole row active (the `is_admin is not None` branch is skipped). `count_active_admin_users()` still gates on both flags so the invariant holds, and reactivation restores cleanly. Just confirming that's the intended split between deactivate and demote.

Regarding the question  - Yes, this is intentional. Deactivation is treated as temporary account suspension, so role assignments remain intact while is_active=False prevents authentication and excludes the user from active-admin counting. Demotion is the operation that changes admin entitlement and synchronizes RBAC roles. This also allows reactivation to restore the previous authorization state without recreating role assignments.

gandhipratik203 (dismissed):
LGTM

a-effort:
LGTM! 🚀

msureshkumar88:
## E2E verification: PR fixes issue #2808

Ran a real end-to-end validation against this branch (not mocked): a live FastAPI server (`uvicorn`, real SQLite DB, real Alembic migrations + platform-admin bootstrap all running automatically on startup), real JWT auth, and real HTTP calls against both `mcpgateway/routers/email_auth.py` (JSON API) and `mcpgateway/admin.py` (Admin UI htmx endpoints — Bearer tokens are CSRF-exempt so no CSRF handshake was needed).

**Config used:**
- Fresh, isolated SQLite DB (not the repo's dev DB)
- `EMAIL_AUTH_ENABLED=true`, `MCPGATEWAY_UI_ENABLED=true`, `MCPGATEWAY_ADMIN_API_ENABLED=true`
- Three accounts: the bootstrapped `PLATFORM_ADMIN_EMAIL` account plus two regular admins (`admin-b`, `admin-c`) created via the admin API
- Server run twice: once with `PROTECT_ALL_ADMINS=true` (default), once with `PROTECT_ALL_ADMINS=false`, reusing the same DB across the restart

**Commands run:** `scripts/verify-pr-5644-e2e.sh` — full script and output in the review notes; summary of what it drove:
- `POST /auth/email/login`, `POST /auth/email/admin/users`, `PATCH /auth/email/admin/users/{email}`, `POST /admin/users/{email}/update` (UI), plus a direct `EmailAuthService` call for one structurally HTTP-unreachable branch (see below)

**Results — all of the issue's acceptance criteria confirmed against the real running app:**

| Scenario | Expected | Result |
|---|---|---|
| Peer demotion (`admin-b` demotes `admin-c`), `PROTECT_ALL_ADMINS=true` | Allow | 200 ✅ (was 400 "Admin protection is enabled" before this PR) |
| Revocation is real — demoted admin's **brand-new** post-demotion login still tries an admin-only call | 403 | 403 ✅ |
| Self-demotion / self-deactivation | Deny (400) | 400 ✅, both `PROTECT_ALL_ADMINS=true` and `=false` (fixes the exact regression in the issue where `false` allowed self-demotion) |
| Case-insensitive self-check (mixed-case email in URL) | Deny | 400 ✅ |
| Admin UI (htmx) peer demotion / self-demotion | Same as API | 200 / 400 ✅ — API and UI enforce identical rules |
| Sole remaining active admin blocked from self-removal | Deny (400) | 400 ✅ |
| Last-active-admin guard for a peer requester (structurally unreachable over HTTP since only an active admin can call this endpoint, so a peer + "target is sole admin" can't co-occur through RBAC — verified via a direct `EmailAuthService.update_user()` call against the same on-disk DB instead) | `ValueError: Cannot demote or deactivate the last remaining active admin user` | ✅ |
| `granted_by` on role-sync attributes the **requesting** admin, not the target | `admin-b@example.com` | ✅ (verified directly against the `user_roles` table) |

`RESULT: ALL CHECKS PASSED` (script exit code 0).

**One pre-existing, out-of-scope observation** (not caused by this PR — reproduced identically against the pre-PR baseline using the old self-demotion path): `mcpgateway/utils/admin_check.py:is_user_admin()` has a documented fast-path (`if user_email == settings.platform_admin_email: return True`) that treats the *specific bootstrap admin identity* as permanently admin for RBAC purposes regardless of its DB `is_admin` flag. Demoting a regular admin account works exactly as intended (verified above), but demoting the bootstrap `PLATFORM_ADMIN_EMAIL` account specifically will update the DB correctly without actually revoking that identity's RBAC access. Predates this PR and is outside #2808's scope, but might be worth its own follow-up issue since it's a bit surprising in combination with this fix.

Approving based on this validation — thanks for the clean root-cause framing in the PR description!

msureshkumar88 (dismissed):
Verified via a real end-to-end run against a live server (see comment above) -- all of issue #2808's acceptance criteria hold. LGTM.

msureshkumar88 (approved):
LGMT

Issue #4545, opened 2026-04-30
[CHORE]: Integration tests not running


cafalchio:
### 🔧 Chore Summary

All tests inside `tests/integration/` needs the arg --with-integration to run. They are not running in our current CI/CD. These tests are skipped by default.

Many of these tests need outside tools (Redis, NGINX, Docker compose). However, some tests are just misplaced and need to be removed from /test/integration.

This issue is to investigate and reorganise the tests, so just real integration tests are inside the folder and if we can run these tests on CI/CD (could be too slow to run) or in a dev pipeline.

### Test Breakdown                                                                                                                                       
  - 11 true integration tests (25.6%) - require Redis/gateway                                                                                               
  - 4 need full docker-compose stack (3 replicas + NGINX + Redis)                                                                                         
  - 7 need Redis only                                                                                                                                     
  - 20 unit tests (46.5%) - misplaced, should be in tests/unit/                                                                                             
  - 12 unclear (27.9%) - need deeper inspection 

### 🧱 Area Affected

Choose the general area(s) that this chore affects:

- [ ] GitHub Actions / CI Pipelines
- [ ] Pre-commit hooks / linters
- [ ] Formatting (black, isort, ruff, etc.)
- [ ] Type-checking (mypy, pyright, pytype, etc.)
- [ ] Dependency cleanup or updates
- [ ] Build system or `Makefile`
- [ ] Containerization (Docker/Podman)
- [ ] Docs or spellcheck
- [ ] SBOM, CVE scans, licenses, or security checks
- [X] Other: Tests

---

### ⚙️ Context / Rationale

Why is this task needed? Does it reduce tech debt, unblock other work, or improve DX/CI reliability?

False confidence - CI shows green but critical paths aren't tested
Risk: Production bugs in Redis clustering, rate limiting, plugin runtime, and multi-replica scenarios go undetected

---

### 📦 Related Make Targets

Reference any relevant Makefile targets that are involved, if applicable. Ex:

- `make lint` - run ruff, mypy, pylint, etc.
- `make pre-commit` - run pre-configured hooks
- `make install-web-linters` - installs npm-based linters
- `make sonar-submit-docker` - run SonarQube scanner via Docker
- `make sbom` - generate CycloneDX software bill of materials
- `make pip-licenses` - generate markdown license inventory
- `make spellcheck` - spell-check source + docs
- `make update` - update Python dependencies in the venv
- `make check-env` - validate required `.env` entries

---

[...]

PR #5645, opened 2026-07-15
chore: remove obsolete redis_fallback config references

### LINKED ISSUE #4603: rate-limiter: remove obsolete redis_fallback config references

Yosiefeyob:
# Pull Request

## 🔗 Related Issue

Closes #4603

---

## 📝 Summary

The rate limiter used to be Python code that lived inside this repo, and it had a `redis_fallback` option: if Redis went down, keep counting requests in local memory instead of failing.

That in-repo rate limiter was later replaced by an installed package, `cpex-rate-limiter` (a Rust rewrite). The new one has no `redis_fallback` option at all. What happens when Redis is unreachable is now decided by a different setting, `fail_mode`, which defaults to `"open"` (let requests through).

`redis_fallback` was not removed from the shipped config files, the OCP Helm chart, the docs, or the test fixtures. So on every gateway boot the new rate limiter hits a key it doesn't recognise and logs `unknown config key(s): redis_fallback`, then ignores it. This leaves a confusing warning in operators' logs and docs that describe a feature that no longer exists.

This PR removes the dead key everywhere it survived, fixes the docs that still describe the old behaviour, renames a test whose name described the removed mechanism, and adds a guard test so an unrecognised config key can't slip back in unnoticed.
### Changes

- Removed `redis_fallback` from `plugins/config.yaml`, `tests/performance/plugins/config.yaml` (×3), `charts/mcp-stack/profiles/ocp/values-pgo.yaml`, and three unit-test mock fixtures.
- Renamed `test_redis_fallback_to_memory_on_unavailable_redis` → `test_unavailable_redis_fails_open_by_default`; the assertion passed only by virtue of the `fail_mode: "open"` default, so the name/docstring now describe what is actually verified.
- Added `tests/integration/test_plugins_config_yaml_validation.py::test_shipped_plugins_config_has_no_unknown_keys`. Loads `plugins/config.yaml`, constructs each plugin via the framework loader, and asserts zero `unknown config key` warnings.

### Two same-root-cause fixes bundled in

Both surfaced while removing the key:

- Dropped a stale `(with redis_fallback: true)` note from the `test_rate_limiter_dynamic_behavior.py` docstring.
- Corrected the `fail_mode: "open"` row in `docs/docs/api/plugin-bindings-api.md`, which incorrectly claimed it "falls back to in-process memory counters" — the current engine allows the request through without counting.
---

## 📏 Reviewability

---

## 🏷️ Type of Change

- [ ] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

| Check | Command | Status |
|---|---|---|
| New regression test | `pytest tests/integration/test_plugins_config_yaml_validation.py --with-integration` | ✅ 1 passed |
| Renamed test | `pytest tests/integration/test_rate_limiter.py -k unavailable_redis_fails_open --with-integration` | ✅ 1 passed |
| Edited unit fixtures | `pytest tests/unit/.../test_tool_plugin_bindings.py tests/unit/.../test_gateway_plugin_manager.py tests/unit/.../test_tool_plugin_binding_service.py` | ✅ 142 passed |

**Guard proven, not just green:** Re-injecting `redis_fallback: true` into `plugins/config.yaml` makes the new test fail with exactly one captured `unknown config key(s): redis_fallback` warning; removing it returns the suite to zero. Guard fails correctly.

---

## ✅ Checklist

---

## 📓 Notes (optional)

The regression test is credited to **Pratik Gandhi** (`Authors:` header). It was drafted *during* #4582 but pulled out before that PR merged (per #4603), then parked on branch `chore/rate-limiter-redis-fallback-cleanup-wip` ([commit `5caa753a`](https://github.com/IBM/mcp-context-forge/commit/5caa753a0)). This PR adapts it from that branch. changes from the draft:

- Loader imports repointed from `mcpgateway.plugins.framework.loader.*` to `cpex.framework` (the framework moved into the `cpex` package after the draft was written). This is the only change to the test's logic.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-07-16T00:08:38Z",
+  "generated_at": "2026-07-20T13:38:01Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -266,7 +266,7 @@
         "hashed_secret": "b4673e578b9b30fe8bba1b555b7b59883444c697",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1746,
+        "line_number": 1747,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -274,7 +274,7 @@
         "hashed_secret": "4a0a2df96d4c9a13a282268cab33ac4b8cbb2c72",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1834,
+        "line_number": 1835,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -282,7 +282,7 @@
         "hashed_secret": "5baa61e4c9b93f3f0682250b6cf8331b7ee68fd8",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 2184,
+        "line_number": 2185,
         "type": "Basic Auth Credentials",
         "verified_result": null
       },
@@ -290,7 +290,7 @@
         "hashed_secret": "fa9beb99e4029ad5a6615399e7bbae21356086b3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 3549,
+        "line_number": 3550,
         "type": "Basic Auth Credentials",
         "verified_result": null
       },
@@ -298,7 +298,7 @@
         "hashed_secret": "fa9beb99e4029ad5a6615399e7bbae21356086b3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 3640,
+        "line_number": 3641,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -306,7 +306,7 @@
         "hashed_secret": "ac371b6dcce28a86c90d12bc57d946a800eebf17",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 3683,
+        "line_number": 3684,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -314,7 +314,7 @@
         "hashed_secret": "0b6ec68df700dec4dcd64babd0eda1edccddace1",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 3688,
+        "line_number": 3689,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -322,7 +322,7 @@
         "hashed_secret": "4ad6f0082ee224001beb3ca5c3e81c8ceea5ed86",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 3693,
+        "line_number": 3694,
         "type": "Secret Keyword",
         "verified_result": null
       }
@@ -382,7 +382,7 @@
         "hashed_secret": "d3ac7a4ef1a838b4134f2f6e7f3c0d249d74b674",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 5953,
+        "line_number": 5963,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -390,7 +390,7 @@
         "hashed_secret": "5932862bcd24dd27d0dc0407ec94fe9d6ea24aeb",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 6450,
+        "line_number": 6460,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -398,7 +398,7 @@
         "hashed_secret": "c77c805e32f173e4321ee9187de9c29cb3804513",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 6462,
+        "line_number": 6472,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -406,7 +406,7 @@
         "hashed_secret": "8fe3df8a68ddd0d4ab2214186cbb8e38ccd0e06a",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 6534,
+        "line_number": 6544,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -414,7 +414,7 @@
         "hashed_secret": "93ac8946882128457cd9e283b30ca851945e6690",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 7601,
+        "line_number": 7611,
         "type": "Secret Keyword",
         "verified_result": null
       }
@@ -4048,7 +4048,7 @@
         "hashed_secret": "c377074d6473f35a91001981355da793dc808ffd",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 750,
+        "line_number": 749,
         "type": "Hex High Entropy String",
         "verified_result": null
       }
@@ -4058,7 +4058,7 @@
         "hashed_secret": "f2b14f68eb995facb3a1c35287b778d5bd785511",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 177,
+        "line_number": 176,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -4066,7 +4066,7 @@
         "hashed_secret": "1073ab6cda4b991cd29f9e83a307f34004ae9327",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 183,
+        "line_number": 182,
         "type": "Secret Keyword",
         "verified_result": null
       },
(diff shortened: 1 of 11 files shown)


gandhipratik203 (changes requested):
Nice cleanup — good to see these stale references swept out. A few notes on the new guard test:

All in `tests/integration/test_plugins_config_yaml_validation.py`:

**1. Unanchored config path**
`load_config("plugins/config.yaml")` is cwd-relative and swallows `FileNotFoundError`, returning an empty config (`loader/config.py:79-81`). Wrong cwd → zero plugins → green having checked nothing. Anchor to `__file__`, like `test_rate_limiter_redis_url_from_yaml.py:37`.

**2. `except Exception: pass`**
Same root cause: every failure path degrades to green. Dormant today (CI installs `[plugins]`), latent against a future rename. Asserting at least one plugin constructed closes both.

**3. `initialize()` on every plugin, errors swallowed**
Construction is necessary; `initialize()` isn't. A future plugin with networked init makes this quietly slow. None active today.

Nits: the in-function `PluginLoader` import is a no-op (`cpex.framework` is already imported at module top), and the renamed test's docstring has a run-on — "governed by fail_mode it decides what happens".

Yosiefeyob:
Thanks for the thorough review, good catches. It's all updated in the latest push (also rebased onto main, so the file header picked up the new convention):

1. Unanchored config path — Fixed. Anchored _CONFIG_YAML to __file__ via parents[2], matching test_rate_limiter_redis_url_from_yaml.py:37. No more cwd sensitivity.

2. except: pass → vacuous pass — Fixed. Added a constructed counter and assert constructed after the loop, so both paths you flagged now fail loudly: an empty config (zero plugins) and an all-failed load both leave the count at 0.

Nit — in-function import — Fixed. Hoisted PluginLoader into the module-level cpex.framework import.

Nit — run-on docstring — Fixed. Reworded to "…governed by fail_mode, which defaults to "open"…".

3. initialize() on every plugin — Kept as-is, deliberately. Going through the loader (build + initialise) mirrors how the gateway actually loads plugins at startup, and catches an unknown-key warning whether a plugin surfaces it at __init__ or initialise() — construct-only would miss the latter. Skipping init also means bypassing the loader with a manual parse_class_name/import_module idiom no other test uses. Given your "none active today" note, the standard loader felt like the better trade, but happy to switch if you disagree.

gandhipratik203 (dismissed):
All points addressed — the path fix, the check that a plugin actually loaded, and the import and docstring cleanups all look good. Nice work.

msureshkumar88 (dismissed):
## E2E verification (real gateway boot, no mocks)

Ran a real-process e2e script against this branch: actual `uvicorn` process, real SQLite DB, the real installed `cpex_rate_limiter` Rust wheel — no `TestClient`, no mocking of the plugin engine.

**Config used**
```
PLUGINS_ENABLED=true
PLUGIN_CONFIG_FILE / PLUGINS_CONFIG_FILE = derived from plugins/config.yaml via tests/live_gateway/plugins/make_enforce_config.py --plugin RateLimiterPlugin
RATELIMITER_REDIS_URL=REDIS_URL=redis://127.0.0.1:19999/0  (unreachable on purpose — the unknown-key warning fires at config validation time, before any network attempt)
AUTH_REQUIRED=true, MCPGATEWAY_UI_ENABLED=true, MCPGATEWAY_ADMIN_API_ENABLED=true
```

Note: the rate limiter ships with `mode: "disabled"` by default, and cpex's `PluginManager` stub-skips disabled plugins (never instantiates them), so a bare default boot never exercises the key-validation path at all, before or after this fix. The real affected population is operators who set `mode: "enforce"`. So the script uses `make_enforce_config.py` — the same tool CI's `plugin-integration.yml` workflow already uses — to force the rate limiter into `enforce` mode against the real `plugins/config.yaml`, which is the faithful way to reproduce this in a live boot.

**Commands run**
```
uv run python3 tests/live_gateway/plugins/make_enforce_config.py --source plugins/config.yaml --plugin RateLimiterPlugin --output config_a.yaml
uv run uvicorn mcpgateway.main:app --host 127.0.0.1 --port 18081   # Leg A, post-fix config
grep -i 'unknown config key' uvicorn_a.log

uv run python3 tests/live_gateway/plugins/make_enforce_config.py --source plugins/config.yaml --plugin RateLimiterPlugin --config-override 'redis_fallback=true' --output config_b.yaml
uv run uvicorn mcpgateway.main:app --host 127.0.0.1 --port 18081   # Leg B, pre-fix state re-injected
grep -i 'unknown config key(s): redis_fallback' uvicorn_b.log

uv run pytest tests/integration/test_plugins_config_yaml_validation.py --with-integration -v   # Leg C
```

**Output**
- Leg A (post-fix, real `plugins/config.yaml` as committed on this branch): boot log — **zero** `unknown config key` matches.
- Leg B (pre-fix simulation, `redis_fallback: true` re-injected into the same enforce config): boot log contains, verbatim:
  `rate limiter: unknown config key(s): redis_fallback; expected one of: by_user, by_tenant, by_tool, algorithm, backend, redis_url, redis_key_prefix, fail_mode, redis_ssl_ca_certs, redis_ssl_certfile, redis_ssl_keyfile, redis_ssl_check_hostname`
  — this is the exact warning issue #4603 reports.
- Leg C: `tests/integration/test_plugins_config_yaml_validation.py::test_shipped_plugins_config_has_no_unknown_keys` — 1 passed.

**Result: 3/3 passed.** The fix is real: it eliminates the warning for the shipped config, and re-injecting the removed key reproduces the original bug on demand against the live Rust engine, not a mock.

## Non-blocking follow-up (not a change request for this PR)

`tests/integration/` (63 files, including the new one) isn't wired into any CI workflow — `pytest.yml`'s only pytest invocation has no `--with-integration`, and `pyproject.toml` doesn't set it via `addopts` either. `plugin-integration.yml` triggers on `plugins/config.yaml` changes but only runs `tests/live_gateway/plugins/*_e2e.py`, not `tests/integration/`. So the new regression guard is correct and does its job (verified above, live), but it won't run automatically on future PRs unless something wires `--with-integration` (or this directory) into CI. Pre-existing repo-wide gap, not introduced by this PR — worth a separate follow-up issue so the protection this PR adds doesn't quietly go stale.

Approving based on the above.

Yosiefeyob:
Thank you for the E2E check @msureshkumar88 and the live-boot verification. Good catch on the CI gap it's already  being tracked in the issue #4545 ("Integration tests not running"), so no need for me to open a duplicate. This guard will start running automatically once that lands. much appreciated.

msureshkumar88 (approved):
Requested changes from the earlier review are addressed: config path anchored to `__file__`, `constructed` counter asserted so silent-empty/all-failed loads fail loudly, in-function import hoisted, docstring run-on fixed. CI green across lint/pytest/plugin-integration/DCO. E2E verification in comments confirms the fix eliminates the `redis_fallback` unknown-key warning against the live rate limiter.

gandhipratik203 (approved):
LGTM

PR #5654, opened 2026-07-16
fix(services): rebuild fork-poisoned FileLock and back off on health-check errors

msureshkumar88:
## Summary

CI's Playwright smoke job (and likely others) has been hanging until the workflow timeout on every PR since ~2026-07-16 04:36 UTC, showing only repeated `curl: Couldn't connect` against `/health`. The captured gateway log from a cancelled run contained ~12 million lines, all:

```
FileLock health check failed: UnixFileLock on /tmp/gateway_service_leader.lock was inherited across fork; construct a new instance
```

**Root cause:** `GatewayService._file_lock` (`mcpgateway/services/gateway_service.py`) is constructed once at import time, before gunicorn `--preload` forks worker processes (`run-gunicorn.sh`, preload is on by default). A recent `filelock` release now detects when a lock object is reused from a different PID than the one that created it and raises instead of silently working. `_run_health_checks`'s generic `except Exception` handler logged that error and immediately re-entered the `while True` loop with no delay — unlike the sibling `except Timeout` branch, which does back off. The result is an unbounded busy loop: the worker never yields enough CPU to actually bind/serve the health endpoint, so CI never got a response and ran until the job's `timeout-minutes: 15` cap force-cancelled it.

This surfaced now because `filelock` is pinned as `filelock>=3.29.4` (no ceiling) and `make install` runs `uv pip install .`, which re-resolves live against PyPI rather than the committed `uv.lock` — so every fresh install picks up whatever `filelock` version is newest at that moment.

## Fix

- Track the PID the `FileLock` was built for (`_file_lock_pid`). When `_run_health_checks` detects it's running under a different PID than the one that created the lock (i.e., after a fork), rebuild the lock for the current process instead of reusing the poisoned instance.
- Add the same `asyncio.sleep(self._health_check_interval)` backoff already used by the `Timeout` branch to the generic `except Exception` branch, so any other unexpected lock error can no longer busy-loop the health-check task.

## Test plan

- Added `test_run_health_checks_rebuilds_filelock_after_fork` — asserts a lock built under a different PID is never acquired directly and is rebuilt/used for the current PID instead.
- Added `test_run_health_checks_backs_off_on_filelock_error` — regression test bounding the number of `acquire()` attempts in a fixed time window when `acquire()` raises, guarding against the busy-loop this PR fixes.
- `pytest tests/unit/mcpgateway/services/test_gateway_service_extended.py tests/unit/mcpgateway/services/test_gateway_service.py` — all passing.
- `ruff check` / `black --check` clean on both changed files.

Scope is intentionally limited to the actual regression (fork-unsafe lock reuse + missing backoff); the unpinned `filelock` floor and the `uv pip install .` vs `uv sync` install path are separate dependency-hygiene issues worth a follow-up but not required for this fix to hold under any future filelock version.


gcgoncalves (dismissed):

msureshkumar88:
## Why this broke now (nothing here regressed)

Traced the timing since this surprised me too — the health-check code (`_run_health_checks`) hasn't changed since 2025-06-11, and `GatewayService._file_lock` has always been constructed pre-fork under `gunicorn --preload`. That's been fork-unsafe the whole time; it just never surfaced because older `filelock` versions silently allowed reusing a lock object across a PID change.

What changed is **upstream**, not this repo:

- `pyproject.toml:112` pins `filelock>=3.29.4` with no ceiling.
- `Makefile:296` — `make install` runs `uv pip install .`, which re-resolves live against PyPI on every install rather than respecting the committed `uv.lock`.

So every fresh CI job/dev install picks up whichever `filelock` release is newest *at that moment*. Somewhere between the last green run (~01:39 UTC) and the first hang (~04:36 UTC), PyPI's `filelock` picked up a fork-safety guard: reusing a lock from a different PID than its creator now raises `RuntimeError` instead of silently working. No mcp-context-forge commit lines up with that window — it's purely a dependency-resolution artifact of the unbounded floor + non-lockfile install path.

Once that new `filelock` started raising, the pre-existing gap in `_run_health_checks` (the generic `except Exception` branch had no backoff, unlike the sibling `except Timeout` branch) turned one log line per health-check tick into an unbounded busy loop — ~12M log lines in one CI run, worker CPU pinned, `/health` never answered, job ran until the 15-minute workflow timeout force-cancelled it.

This PR fixes the actual defect (fork-unsafe lock reuse + missing backoff) so it can't recur regardless of what `filelock` does upstream in the future. The unbounded `filelock` floor and `uv pip install .` vs `uv sync` install-path non-determinism are separate dependency-hygiene issues worth a follow-up, but are out of scope here since fixing the code handles this regardless of pin.

msureshkumar88:
Added a second commit: capped \`filelock\` to \`>=3.29.4,<3.30.0\` in \`pyproject.toml\`/\`uv.lock\`.

Pinned down the exact trigger via PyPI release history: \`filelock 3.30.0\` was published **2026-07-16T03:53:56Z** — ~42 minutes before the first CI hang (04:36 UTC). That's the release that added the fork-safety guard. The already-locked \`uv.lock\` resolution (3.29.5, from before 3.30.0 existed) is unaffected by this — the ceiling just stops a future \`make install\` from floating onto 3.30.0+ silently, same as it did here.

This is defense-in-depth, not the actual fix — the fork-safe rebuild + backoff in the first commit is what makes the code correct regardless of filelock version. Once fork-safety is verified against 3.30.0+, this ceiling can be lifted.

gcgoncalves (dismissed):

Lang-Akshay (dismissed):
LGTM

msureshkumar88:
Good push — you're right that \`<3.30.0\` was overly conservative once the fork-safety fix is actually in place. Changed the ceiling to \`>=3.29.4,<3.31.0\`.

Verified this isn't just theoretical: installed the real \`filelock==3.30.0\`, reproduced the exact fork-reuse \`RuntimeError\` via \`os.fork()\` + \`FileLock.acquire()\` on the inherited object (message matches the CI log verbatim), then confirmed the PID-rebuild logic from the first commit sidesteps it cleanly - the rebuilt lock is never the poisoned pre-fork instance, so 3.30.0's new guard never triggers. Ran the full \`gateway_service\` test suite under 3.30.0 too - all green.

So \`3.30.0\` is safe to allow. Kept a ceiling rather than removing it entirely though (\`<3.31.0\`, not unbounded) - the whole reason this broke silently was an unbounded floor letting installs float onto an untested release. \`3.31.0\`+ hasn't been verified against, so it stays capped until someone does the same check and bumps it.

ja8zyjits (approved):
LGTM

gcgoncalves (approved):

PR #5657, opened 2026-07-16
feat(api): add prompt and resource counts to GatewayRead

### LINKED ISSUE #5656: [FEATURE]: Add prompt and resource counts to GatewayRead
## Description

`GatewayRead` currently exposes `tool_count` (populated from the eagerly-loaded `tools` relationship) but has no equivalent for prompts or resources registered on a gateway. Clients listing gateways have to make additional calls to find out how many prompts/resources a gateway exposes.

## Proposed change

Add `prompt_count` and `resource_count` fields to `GatewayRead`, populated the same way `tool_count` is (via `selectinload` on `GatewayService`'s list/read queries, falling back to `0` when the relationship isn't eagerly loaded).

No new DB columns or migration needed — these are computed, transient fields.

gcgoncalves:
# ✨ Feature / Enhancement PR

## 🔗 Epic / Issue

Closes #5656

---

## 🚀 Summary (1-2 sentences)

Add prompt and resource counts to GatewayRead

---

## 📏 Reviewability

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -1850,7 +1850,7 @@
         "hashed_secret": "5193dda17da630c041fe949c06e7b6dd5ac8628f",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1609,
+        "line_number": 1612,
         "type": "Base64 High Entropy String",
         "verified_result": null
       },
@@ -1858,7 +1858,7 @@
         "hashed_secret": "91174c4e6859da80df4e7ce9d45b35501f591210",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1892,
+        "line_number": 1895,
         "type": "Base64 High Entropy String",
         "verified_result": null
       },
@@ -1866,7 +1866,7 @@
         "hashed_secret": "4a0a2df96d4c9a13a282268cab33ac4b8cbb2c72",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1933,
+        "line_number": 1936,
         "type": "Secret Keyword",
         "verified_result": null
       }
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3892,7 +3892,7 @@
         "hashed_secret": "c377074d6473f35a91001981355da793dc808ffd",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 4599,
+        "line_number": 4608,
         "type": "Hex High Entropy String",
         "verified_result": null
       }
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -1842,7 +1842,7 @@
         "hashed_secret": "848f72afea73993a89ad3dcd66fc152f04f2be41",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1028,
+        "line_number": 1031,
         "type": "Base64 High Entropy String",
         "verified_result": null
       },
@@ -3892,7 +3892,7 @@
         "hashed_secret": "c377074d6473f35a91001981355da793dc808ffd",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 4593,
+        "line_number": 4599,
         "type": "Hex High Entropy String",
         "verified_result": null
       }
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -1810,7 +1810,7 @@
         "hashed_secret": "ad7920eddff0914889d74a47bb45fc80697da5a8",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 265,
+        "line_number": 268,
         "type": "Base64 High Entropy String",
         "verified_result": null
       },
@@ -1818,7 +1818,7 @@
         "hashed_secret": "998ab9e14841d778dcafdf72ceaac0cacbb5ea2a",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 467,
+        "line_number": 470,
         "type": "Base64 High Entropy String",
         "verified_result": null
       },
@@ -1826,7 +1826,7 @@
         "hashed_secret": "d20047e4dff166045c7f61066c62955ec33c1657",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 670,
+        "line_number": 673,
         "type": "Base64 High Entropy String",
         "verified_result": null
       },
@@ -1834,7 +1834,7 @@
         "hashed_secret": "fc19bda36b0af2d6a957034199a79c6583c707a8",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 831,
+        "line_number": 834,
         "type": "Base64 High Entropy String",
         "verified_result": null
       },
--- a/docs/docs/manage/api-usage.md
+++ b/docs/docs/manage/api-usage.md
@@ -259,6 +259,9 @@ curl -s -H "Authorization: Bearer $TOKEN" $BASE_URL/gateways | jq '.'
       "name": "my-mcp-server",
       "url": "http://localhost:9000/mcp",
       "enabled": true,
+      "tool_count": 5,
+      "prompt_count": 3,
+      "resource_count": 2,
       ...
     }
   ],
--- a/docs/docs/manage/api-usage.md
+++ b/docs/docs/manage/api-usage.md
@@ -259,9 +259,9 @@ curl -s -H "Authorization: Bearer $TOKEN" $BASE_URL/gateways | jq '.'
       "name": "my-mcp-server",
       "url": "http://localhost:9000/mcp",
       "enabled": true,
-      "tool_count": 5,
-      "prompt_count": 3,
-      "resource_count": 2,
+      "toolCount": 5,
+      "promptCount": 3,
+      "resourceCount": 2,
       ...
     }
   ],
--- a/mcpgateway/admin.py
+++ b/mcpgateway/admin.py
@@ -166,6 +166,7 @@ from mcpgateway.services.email_auth_service import AuthenticationError, EmailAut
 from mcpgateway.services.encryption_service import get_encryption_service
 from mcpgateway.services.export_service import ExportError, ExportService
 from mcpgateway.services.gateway_service import (
+    gateway_capability_loaders,
     GatewayConnectionError,
     GatewayDuplicateConflictError,
     GatewayLookupConflictError,
@@ -9698,7 +9699,7 @@ async def admin_gateways_partial_html(
     team_ids = await _get_user_team_ids(user, db)
 
     # Build base query
-    query = select(DbGateway).options(joinedload(DbGateway.email_team), selectinload(DbGateway.tools), selectinload(DbGateway.prompts), selectinload(DbGateway.resources))
+    query = select(DbGateway).options(joinedload(DbGateway.email_team), *gateway_capability_loaders())
 
     if not include_inactive:
         query = query.where(DbGateway.enabled.is_(True))
--- a/mcpgateway/admin.py
+++ b/mcpgateway/admin.py
@@ -9698,7 +9698,7 @@ async def admin_gateways_partial_html(
     team_ids = await _get_user_team_ids(user, db)
 
     # Build base query
-    query = select(DbGateway).options(joinedload(DbGateway.email_team), selectinload(DbGateway.tools))
+    query = select(DbGateway).options(joinedload(DbGateway.email_team), selectinload(DbGateway.tools), selectinload(DbGateway.prompts), selectinload(DbGateway.resources))
 
     if not include_inactive:
         query = query.where(DbGateway.enabled.is_(True))
--- a/mcpgateway/schemas.py
+++ b/mcpgateway/schemas.py
@@ -3880,6 +3880,12 @@ class GatewayRead(BaseModelWithConfigDict):
     # Tool count (populated from the tools relationship; 0 when not loaded)
     tool_count: int = Field(default=0, description="Number of tools registered for this gateway")
 
+    # Prompt count (populated from the prompts relationship; 0 when not loaded)
+    prompt_count: int = Field(default=0, description="Number of prompts registered for this gateway")
+
+    # Resource count (populated from the resources relationship; 0 when not loaded)
+    resource_count: int = Field(default=0, description="Number of resources registered for this gateway")
+
     # Tools skipped during gateway import due to validation errors (transient, not persisted)
     skipped_tools: List[str] = Field(default_factory=list, description="Tools skipped during gateway import due to validation errors")
 
--- a/mcpgateway/schemas.py
+++ b/mcpgateway/schemas.py
@@ -3878,13 +3878,22 @@ class GatewayRead(BaseModelWithConfigDict):
     _normalize_visibility = field_validator("visibility", mode="before")(classmethod(lambda cls, v: _coerce_visibility(v)))
 
     # Tool count (populated from the tools relationship; 0 when not loaded)
-    tool_count: int = Field(default=0, description="Number of tools registered for this gateway")
+    tool_count: int = Field(
+        default=0,
+        description="Total tools registered for this gateway, including disabled ones. Not filtered per-tool by caller visibility. 0 may mean none registered or the field wasn't populated on this response path.",
+    )
 
     # Prompt count (populated from the prompts relationship; 0 when not loaded)
-    prompt_count: int = Field(default=0, description="Number of prompts registered for this gateway")
+    prompt_count: int = Field(
+        default=0,
+        description="Total prompts registered for this gateway, including disabled ones. Not filtered per-prompt by caller visibility. 0 may mean none registered or the field wasn't populated on this response path.",
+    )
 
     # Resource count (populated from the resources relationship; 0 when not loaded)
-    resource_count: int = Field(default=0, description="Number of resources registered for this gateway")
+    resource_count: int = Field(
+        default=0,
+        description="Total resources registered for this gateway, including disabled ones. Not filtered per-resource by caller visibility. 0 may mean none registered or the field wasn't populated on this response path.",
+    )
 
     # Tools skipped during gateway import due to validation errors (transient, not persisted)
     skipped_tools: List[str] = Field(default_factory=list, description="Tools skipped during gateway import due to validation errors")
--- a/mcpgateway/services/gateway_service.py
+++ b/mcpgateway/services/gateway_service.py
@@ -5558,10 +5558,10 @@ class GatewayService(BaseService):  # pylint: disable=too-many-instance-attribut
         gateway_dict["version"] = getattr(gateway, "version", None)
         gateway_dict["team"] = getattr(gateway, "team", None)
 
-        # Populate from the eagerly-loaded tools relationship when available, this helps to log or fail in the else condition if needed in future.
-        gateway_dict["tool_count"] = len(gateway.__dict__.get("tools", [])) if gateway.__dict__.get("tools") else 0
-        gateway_dict["prompt_count"] = len(gateway.__dict__.get("prompts", [])) if gateway.__dict__.get("prompts") else 0
-        gateway_dict["resource_count"] = len(gateway.__dict__.get("resources", [])) if gateway.__dict__.get("resources") else 0
+        # Populate from eagerly-loaded relationships; falls back to 0 if not loaded.
+        gateway_dict["tool_count"] = len(gateway.__dict__.get("tools") or [])
+        gateway_dict["prompt_count"] = len(gateway.__dict__.get("prompts") or [])
+        gateway_dict["resource_count"] = len(gateway.__dict__.get("resources") or [])
 
         return GatewayRead.model_validate(gateway_dict).masked()
 
--- a/mcpgateway/services/gateway_service.py
+++ b/mcpgateway/services/gateway_service.py
@@ -2411,8 +2411,17 @@ class GatewayService(BaseService):  # pylint: disable=too-many-instance-attribut
                 cached_gateways = [GatewayRead.model_validate(g).masked() for g in cached["gateways"]]
                 return (cached_gateways, cached.get("next_cursor"))
 
-        # Build base query with ordering
-        query = select(DbGateway).options(joinedload(DbGateway.email_team)).order_by(desc(DbGateway.created_at), desc(DbGateway.id))
+        # Build base query with ordering and eager load relationships for capability counts
+        query = (
+            select(DbGateway)
+            .options(
+                joinedload(DbGateway.email_team),
+                selectinload(DbGateway.tools),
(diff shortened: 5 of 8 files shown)


ja8zyjits (changes requested):
## 🚨 Blocking Issues

### 1. N+1 Query Trigger in Count Population
- **File**: `mcpgateway/services/gateway_service.py:5496-5498`
- **Current**: `len(gateway.tools) if gateway.__dict__.get("tools") else 0`
- **Issue**: Accessing `gateway.tools` triggers SQLAlchemy lazy load if `__dict__.get("tools")` returns `None`
- **Impact**: Defeats `selectinload()` optimization, causes N+1 queries
- **Fix**: `len(gateway.__dict__.get("tools", []))` (apply to all 3 lines: tools, prompts, resources)

### 2. Missing API Documentation
- **File**: `docs/docs/manage/api-usage.md:253`
- **Code**: `mcpgateway/schemas.py:3880`
- **Issue**: Gateway list response docs missing `prompt_count` and `resource_count` fields
- **Fix**: Update `GET /gateways` response example to include `tool_count`, `prompt_count`, `resource_count`

ja8zyjits:
## ⚠️ Warnings (Non blocking)

### Missing Test Coverage
- **File**: `mcpgateway/services/gateway_service.py:2405-2417`
- **Issue**: No test for `list_gateways()` capability count population (only `list_gateways_for_user()` tested)
- **Needed**: `test_list_gateways_populates_capability_counts()`
- **Rationale**: Both methods got identical eager-load changes

---

gcgoncalves:
@ja8zyjits Addressed!

ja8zyjits:
## 🚨 Blocking Issues

### Must Fix Before Merge

1. **Empty test function**
   - File: `tests/unit/mcpgateway/services/test_gateway_service_helpers.py:280`
   - Function: `test_gateway_service_validate_tools_valueerror`
   - Current: Contains only `pass`
   - Fix: Remove or implement
   - Note: Leftover from refactoring

---

## ⚠️ Functionally-Impacting Issues

### Should Fix (High Priority)

1. **Redundant dict lookups in count population**
   - File: `mcpgateway/services/gateway_service.py:5555-5557`
   - Issue: Calls `gateway.__dict__.get("tools")` twice per field
   - Fix (simpler):
     ```python
     gateway_dict["tool_count"] = len(gateway.__dict__.get("tools") or [])
     gateway_dict["prompt_count"] = len(gateway.__dict__.get("prompts") or [])
     gateway_dict["resource_count"] = len(gateway.__dict__.get("resources") or [])
     ```
   - Benefit: Single lookup per field, cleaner

2. **Missing pagination test**
   - File: `mcpgateway/services/gateway_service.py:2405-2502`
   - Needed: `test_list_gateways_populates_capability_counts_with_cursor()`
   - Rationale: Verify eager-loading persists across pages

3. **Missing team filtering test**
   - File: `mcpgateway/services/gateway_service.py:2511-2580`
   - Needed: `test_list_gateways_for_user_filters_by_team_and_populates_counts()`
   - Rationale: Verify counts work with team-scoped visibility

4. **Missing partial-load test**
   - File: `mcpgateway/services/gateway_service.py:5555-5558`
   - Needed: `test_convert_gateway_to_read_capabilities_partial_load()`
   - Example: `tools` loaded, `prompts`/`resources` missing
   - Rationale: Verify fallback handles partial eager-loading

---

## 💡 Suggestions

### Nice-to-Have (Medium Priority)

1. **Remove unnecessary else after return**
   - File: `mcpgateway/scripts/validate_env.py:248-251`
   - Current:
     ```python
     if is_prod or exit_on_warnings:
         return 1
     else:
         print("⚠️ Warnings detected...")
     ```
   - Better:
     ```python
     if is_prod or exit_on_warnings:
         return 1
     print("⚠️ Warnings detected...")
     ```

2. **Revert parentheses removal for clarity**
   - File: `mcpgateway/scripts/validate_env.py:120`
   - Changed from: `if not (1 <= settings.port <= 65535):`
   - Changed to: `if not 1 <= settings.port <= 65535:`
   - Recommendation: Revert — parentheses group range check clearly

3. **Add integration test for end-to-end flow**
   - Suggested: `tests/integration/test_gateway_list_capabilities.py`
   - Rationale: Unit tests mock DB; integration verifies ORM query works

4. **Add N+1 prevention performance test**
   - Suggested: `tests/performance/test_gateway_list_queries.py`
   - Reference: `tests/helpers/query_counter.py`
   - Rationale: Verify eager-loading prevents N+1 (regression protection)

5. **Add schema validation test**
   - Suggested: `test_gateway_read_schema.py`
   - Rationale: Ensure docs match actual API response

---

msureshkumar88 (changes requested):
Nice, tight PR — this maps closely to #5656, no schema/migration needed since these are computed fields, and the earlier feedback on the N+1 lazy-load and the missing docs example both look addressed already. A couple of things worth sorting before merge, plus a few optional follow-ups.

## 🚨 Blocking

**1. Admin UI partial-listing query is missing the new eager loads**
`mcpgateway/admin.py` (`/admin/gateways/partial`, ~line 9652) builds its own `DbGateway` query independent of `GatewayService.list_gateways()`:
```python
query = select(DbGateway).options(joinedload(DbGateway.email_team), selectinload(DbGateway.tools))
```
It still calls the shared `convert_gateway_to_read()`, so `prompt_count`/`resource_count` will come back `0` for every gateway rendered through this endpoint (the JSON payload fed to the admin table), while `tool_count` stays correct. Could we add `selectinload(DbGateway.prompts), selectinload(DbGateway.resources)` here too, to match the other two call sites?

**2. Vacuous test**
`test_list_gateways_populates_capability_counts_with_cursor` (`tests/unit/mcpgateway/services/test_gateway_service_helpers.py:320-353`) calls `list_gateways(db, cursor=..., limit=10)` but has no assertions — it passes unconditionally regardless of behavior. Worth either finishing it (assert counts like the sibling tests do) or dropping it.

## 💭 Suggestions (non-blocking)

- Two unrelated diffs seem to have snuck in from a rebase — `mcpgateway/scripts/validate_env.py` (one blank line) and `tests/unit/mcpgateway/test_main.py` (wraps an unrelated `/roots/%2Ferror` test in `caplog.at_level`). Might be worth dropping them from this PR to keep the diff scoped to the issue.
- The comment at `gateway_service.py:5555` ("this helps to log or fail in the else condition if needed in future") reads like a leftover thought rather than a rationale — could probably be simplified or removed.
- All the new count-propagation tests are mock-based (`SimpleNamespace` + `MagicMock(db)`), so none actually assert that `selectinload(...)` is present on the query — a future accidental removal wouldn't be caught. A lightweight in-memory-SQLite integration test (the file already imports `create_engine`/`sessionmaker`/`Base`) would close that gap and also would have caught finding #1.
- `list_gateways_for_user` looks unused in production (only self-reference + tests, and its own docstring calls it deprecated) — not blocking here, but might not be worth continuing to extend; could be a separate cleanup issue.
- Performance-wise, `selectinload` pulls full `Tool`/`Prompt`/`Resource` rows just to `len()` them — for gateways with many entries, a `GROUP BY`/`func.count()` aggregate would be lighter. Worth considering as a follow-up if this becomes a hot path.

gcgoncalves:
@ja8zyjits @msureshkumar88 addressed!

ja8zyjits:
@gcgoncalves 
### Test Gaps

- [ ] **Incomplete assertion** - `test_list_gateways_populates_capability_counts:318`
  - Missing: `assert captured["prompt_count"] == 2` and `assert captured["resource_count"] == 1`
  - Test data has `prompts=["p1", "p2"]`, `resources=["r1"]`

- [ ] **Admin UI path untested** - `admin.py:9661`
  - Added `selectinload(DbGateway.prompts/resources)` to admin query
  - Need: Integration test in `test_admin.py` or `test_gateways_page.py`

- [ ] **No E2E test** - API response validation
  - Need: E2E test calling `GET /gateways`, asserting JSON includes all three count fields
  - Suggested: `tests/integration/` or `tests/e2e/`

gcgoncalves:
@ja8zyjits Tests added

msureshkumar88 (changes requested):
Nice, tightly-scoped PR — this lands #5656 exactly as specified, and I like that you resisted adding a persisted column: transient computed fields are the right call here, since a stored counter would need write-path maintenance and could drift from reality. No migration is correct. The earlier rounds (admin partial eager loads, the vacuous cursor test, the unrelated `validate_env.py` / `test_main.py` diffs) all look resolved.

I verified the behavior against a real session rather than just reading the diff — details below. One change I'd like to see before merge, the rest are optional.

## 🚨 Blocking

**Eager loads fetch whole child rows (including BLOBs and encrypted credentials) just to `len()` them**

`mcpgateway/services/gateway_service.py:2408-2418`, `:2514-2520`, and `mcpgateway/admin.py:9661`

`selectinload(DbGateway.tools/prompts/resources)` with no column restriction pulls every column of every gateway-owned child. Captured SQL using the PR's exact options against an in-memory DB:

```
SELECT tools.gateway_id, tools.id, tools.original_name, tools.url, … tools.headers, tools.input_schema, tools.auth_value, …
SELECT prompts.gateway_id, prompts.id, … prompts.template, prompts.argument_schema, …
SELECT resources.gateway_id, resources.id, … resources.text_content, resources.binary_content, …
```

So a single `GET /gateways` page — and every render of the admin gateways table — now materializes `resources.binary_content` (BLOB) and `resources.text_content` for every gateway-owned resource, plus every tool's `input_schema`, to produce three integers. On a deployment with several federated gateways each exposing tens of cached resources, that is a meaningful amount of transfer and ORM object churn per list request. It also loads every tool's encrypted `auth_value` into process memory on a read-only listing path — unnecessary exposure surface for credential material, even though it is encrypted at rest.

Would you be open to constraining the loads? Same code shape, and I confirmed the counts come out identical:

```python
query = (
    select(DbGateway)
    .options(
        joinedload(DbGateway.email_team),
        selectinload(DbGateway.tools).load_only(DbTool.id),
        selectinload(DbGateway.prompts).load_only(DbPrompt.id),
        selectinload(DbGateway.resources).load_only(DbResource.id),
    )
    .order_by(desc(DbGateway.created_at), desc(DbGateway.id))
)
```

The emitted SQL becomes `SELECT resources.gateway_id, resources.id FROM resources WHERE resources.gateway_id IN (…)` — two columns instead of ~30, no BLOBs, no credentials. Same treatment for `admin.py:9661`, and it would be worth applying to the pre-existing loads in `get_gateway()` (`:3443-3448`) for the same reason.

One caveat worth a short comment in the code: `load_only` leaves the other columns deferred, so any future code that touches an unloaded attribute on those objects triggers a per-row SELECT. `convert_gateway_to_read()` only calls `len()` and the objects don't escape, so it's safe today — a comment would stop someone reintroducing N+1 through the back door later.

If counts ever become a hot path, correlated `func.count()` scalar subqueries would avoid materializing child rows entirely — but that touches every `convert_gateway_to_read()` call site, so it's better as a follow-up issue than as part of this PR.

## 💭 Suggestions (non-blocking)

**1. Worth calling out a behavior change the description doesn't mention.** Before this PR, `list_gateways()` had no `tools` eager load, and `convert_gateway_to_read()` reads `gateway.__dict__.get("tools")` (deliberately, to avoid a lazy load) — so `GET /gateways` was returning `toolCount: 0` for *every* gateway on main. I confirmed the mechanism: without the eager load, `"tools" in gateway.__dict__` is `False`. That's why the admin UI showed correct tool counts (`admin.py` already had `selectinload(DbGateway.tools)`) while the API reported zero. This PR silently fixes that too. Not breaking, but anyone who special-cased `toolCount == 0` as "never fetched" — which `gateways_partial.html:93` does for the Fetch/Refresh button label — will see the signal change meaning on that endpoint. A line in the PR body or release notes would help.

For what it's worth, I checked the other verbs and they're fine: `POST /gateways` populates the collections at construction (`DbGateway(..., tools=db_tools, ...)`) and `db.refresh()` preserves them; `PUT /gateways/{id}` re-reads them after the catalog re-sync (verified 1 → 2 after a child insert + refresh); and the async-pending register path returning zeros is semantically right.

**2. The eager-load option set now lives in four places** — `get_gateway()`, `list_gateways()`, `list_gateways_for_user()`, and `admin.py`. Drift between them is exactly the bug caught in the previous round. A small module-level helper would make the next endpoint returning `GatewayRead` correct by construction:

```python
def _gateway_capability_loaders():
    """Eager loads required for GatewayRead capability counts."""
    return (
        selectinload(DbGateway.tools).load_only(DbTool.id),
        selectinload(DbGateway.prompts).load_only(DbPrompt.id),
        selectinload(DbGateway.resources).load_only(DbResource.id),
    )
```

**3. `list_gateways_for_user()` looks like dead code.** Its docstring says "DEPRECATED … no longer used", and a repo-wide search turns up only tests and the definition — no production caller. Extending it here is defensible for consistency, but it might be worth a separate cleanup issue rather than continuing to grow it.

**4. Implicit dependency on `expire_on_commit=False`.** `list_gateways()` calls `db.commit()` at `:2453`, before the conversion loop at `:2456`. The counts survive only because `SessionLocal` sets `expire_on_commit=False` (`db.py:569`) — with SQLAlchemy's default, the commit would expire the instances and every count would silently become `0`. A one-line comment at the commit would flag that, since the failure mode is silent zeros rather than an exception.

**5. Field semantics could be documented.** The counts are gateway-wide row counts: not filtered by the caller's visibility of individual children, inclusive of disabled tools/prompts/resources, and `0` is ambiguous between "none registered" and "relationship not eager-loaded on this path". That all matches the existing `tool_count` contract and the issue's stated design, so nothing to change functionally — but `Field(description="Number of prompts registered for this gateway")` reads as an unconditional guarantee. A clarifying clause there (and/or in `api-usage.md`) would keep client authors from treating `0` as authoritative. Worth noting the visibility angle explicitly too: a user who can see a public gateway learns how many private prompts/resources sit under it. It's an aggregate with no identifiers and consistent with today's `tool_count`, so documenting the semantics seems like the right resolution rather than filtering the counts.

## ✅ Testing

Coverage of the changed lines is good and all three previously-raised gaps are closed. I confirmed `tests/e2e/` is inside `testpaths` and not in `PYTEST_IGNORE`, so the new e2e test does run under `make test`. Locally: `test_gateway_service_helpers.py` 54 passed, `test_main_apis.py -k TestGatewayAPIs` passed, `make ruff` clean, pylint 10.00/10 on `gateway_service.py`.

A few gaps worth considering:

- **The `list_gateways` unit tests don't guard the eager loads.** `test_list_gateways_populates_capability_counts`, `…_with_cursor`, and `…_for_user_filters_by_team_and_populates_counts` all use a `MagicMock` session returning a `SimpleNamespace` that already has `tools`/`prompts`/`resources` in `__dict__` — delete the `selectinload(...)` lines from the service and all three still pass. Your `admin.py` test *does* assert on the query options; mirroring that assertion here would close the regression hole. The e2e test covers `GET /gateways` end-to-end, which is the stronger guard, so this is robustness rather than a coverage hole.
- **`test_list_gateways_populates_capability_counts_with_cursor` adds little over its sibling.** With a `MagicMock` db, `cursor="some-cursor"` never exercises cursor decoding or the `limit+1` window; the assertions come straight from the mock. Either drive it through the real `unified_paginate` path or drop it.
- **`GET /gateways/{id}` has no count assertion.** `get_gateway()` already eager-loads all three so it works, but nothing pins it — three lines in the existing e2e test would cover it.
- **No zero-count or disabled-child case on a real DB.** The e2e test only asserts 3/2/1 on a fully-populated gateway. A second gateway with no children (asserting `0/0/0` rather than inheriting the default) plus one disabled child would pin the semantics from suggestion 5.
- **No query-count regression test.** `tests/helpers/query_counter.py` exists; asserting a constant query count across a multi-gateway list would catch a future revert to lazy loading, which the mock tests can't.
- **`{opt.path.path[-2].key for opt in query._with_options}`** in `test_admin.py` reaches into SQLAlchemy internals and will need updating on a major bump. Fine as-is — a comment naming the assumption would save the next person some archaeology.

## Minor notes

- Not a breaking change: three optional fields with `default=0`, nothing removed or renamed. Cached gateway payloads written by an older version deserialize fine (missing keys fall back to defaults), and the registry cache TTL is 20s, so no long-lived inconsistency.
- No new failure modes in logging/exceptions — both call sites already wrap `convert_gateway_to_read()` with `logger.exception`, and `len(… or [])` can't raise on these values.
- No authn/authz surface change; counts flow through the existing access-control filters, so a gateway the caller can't see produces no row at all.
- `Closes #5656` is in the PR body but none of the four commits reference it — optional, though a trailer on the squash commit keeps `git log` self-describing.

Thanks for iterating on this — happy to re-review as soon as the `load_only` change is in.

gcgoncalves:
@msureshkumar88 addressed!

msureshkumar88 (dismissed):
## E2E verification — real gateway, real MCP federation, all checks passed ✅

I wrote and ran a live end-to-end verification for this PR — a real `mcpgateway` server, real MCP servers federated over stdio→SSE, real HTTP calls, no DB fixtures or mocks. Script: `scripts/verify-pr-5657-e2e.sh` (added to this comment's repo checkout for reproducibility — happy to open a follow-up PR with it if useful, or drop it if you'd rather not carry a verification script long-term).

### Config used

```bash
GW_PORT=45657
BRIDGE_EVERYTHING_PORT=45658   # @modelcontextprotocol/server-everything (13 tools / 4 prompts / 9 resources)
BRIDGE_EMPTY_PORT=45659        # deliberately empty FastMCP server (0/0/0 case)
BRIDGE_EVERYTHING2_PORT=45660  # 2nd everything-server instance, for the N+1 scale check

DATABASE_URL=sqlite:////<tmp>/e2e_pr5657_test.db
AUTH_REQUIRED=true
EMAIL_AUTH_ENABLED=true
SSRF_ALLOW_LOCALHOST=true      # bridges are on 127.0.0.1; needed to pass the default SSRF guard
SQLALCHEMY_ECHO=true           # server-side SQL logging, used for the load_only() proof below
JWT_SECRET_KEY=<32+ char throwaway secret, generated per-run>
AUTH_ENCRYPTION_SECRET=<32+ char throwaway secret, generated per-run>
```

### Commands run (abridged — full sequence is in the script)

```bash
uv run uvicorn mcpgateway.main:app --host 127.0.0.1 --port 45657 &
uv run python -m mcpgateway.translate --stdio "npx -y @modelcontextprotocol/server-everything stdio" --port 45658 &
uv run python -m mcpgateway.translate --stdio "python3 empty_mcp_server.py" --port 45659 &
python -m mcpgateway.utils.create_jwt_token --username admin@example.com --exp 60 --secret "$JWT_SECRET_KEY"

curl -X POST http://127.0.0.1:45657/gateways -d '{"name":"e2e-gw-everything-1","url":"http://127.0.0.1:45658/sse", ...}'
curl -X POST http://127.0.0.1:45657/gateways -d '{"name":"e2e-gw-empty-1","url":"http://127.0.0.1:45659/sse", ...}'
curl http://127.0.0.1:45657/gateways
curl http://127.0.0.1:45657/gateways/{id}
curl http://127.0.0.1:45657/admin/gateways/partial
curl -X POST http://127.0.0.1:45657/tools/{tool_id}/state?activate=false
curl -X PUT / -X DELETE against the registered gateways

python -m pytest tests/unit/mcpgateway/services/test_gateway_service_helpers.py \
  tests/unit/mcpgateway/test_admin.py -k capab tests/e2e/test_main_apis.py -k gateway -q
```

### Results — all 10 checks passed

| # | Check | Result |
|---|-------|--------|
| 1 | `GET /gateways` reflects real capability counts | ✅ gw1 (everything-server) `toolCount=13 promptCount=4 resourceCount=9` — exact match. gw2 (empty server) `0/0/0` |
| 2 | `GET /gateways/{id}` matches | ✅ same exact counts on both gateways |
| 3 | `GET /admin/gateways/partial` | ✅ 200, correct data flowing through (the template only surfaces `toolCount` via the Fetch/Refresh-button label — `promptCount`/`resourceCount` aren't rendered anywhere yet, so those two weren't independently re-provable through *this* endpoint, but checks 1/2 already cover the same underlying code path exactly) |
| 4 | Zero-count real-DB case | ✅ gw2 reports `0/0/0`, not an error or stale default |
| 5 | Disabled child still counted | ✅ disabled 1 tool via `POST /tools/{id}/state?activate=false`, `toolCount` stayed at 13 — confirms the field docs' "including disabled ones" claim against a real row |
| 6 | Live SQL matches the `load_only(id)` claim | ✅ captured actual emitted SQL: `SELECT tools.gateway_id, tools.id FROM tools WHERE tools.gateway_id IN (...)` — no `auth_value`/`headers`/`input_schema`; same narrow shape for `prompts`/`resources`, no BLOB columns |
| 7 | No N+1 as gateway count grows | ✅ 3 gateways registered → exactly 1 SELECT against `tools` and 1 against `resources` total (`IN (?, ?, ?)`), not 3 |
| 8 | CRUD regression | ✅ `PUT` (update persists, counts preserved), `DELETE` (200, then 404 on re-fetch) — no 500s anywhere |
| 9 | Clean boot | ✅ server up, no import/traceback errors from the new `gateway_capability_loaders` import |
| 10 | PR's own test suites | ✅ full pass (1 unrelated skip: external-connectivity test) |

Full log has the raw JSON for each captured SQL statement if useful for the record.

### Verdict

**Approve.** This closes the loop on the `load_only(id)` fix from the prior review round — proven against a live server doing real MCP federation, not just read from the diff. No regressions found in surrounding gateway CRUD, no N+1 reintroduced as gateway count scales, and the "counts include disabled children" documentation claim holds against a real row. No changes required.

ja8zyjits (changes requested):
## 🚨 Blocking Issues

### Critical: API Response Field Naming Mismatch
- **File**: `docs/docs/manage/api-usage.md:262-264`
- **Issue**: Docs show snake_case (`tool_count`, `prompt_count`, `resource_count`) but API returns camelCase (`toolCount`, `promptCount`, `resourceCount`)
- **Root Cause**: `mcpgateway/utils/base_models.py:52-118` - Pydantic `alias_generator=to_camel_case`
- **Impact**: Clients following docs will use wrong field names
- **Fix**: Update docs to camelCase:
  ```json
  {
    "id": "abc123",
    "name": "my-mcp-server",
    "toolCount": 5,
    "promptCount": 3,
    "resourceCount": 2
  }
  ```

---

## ⚠️ Warnings

### Inconsistent Documentation
- **File**: `docs/docs/using/clients/llm-chat.md:210`
- **Issue**: Shows only `tool_count` (snake_case) in `/llmchat/connect` response, missing `prompt_count`/`resource_count`
- **Fix**: Remove `tool_count` if not applicable, or add all three in camelCase

ja8zyjits (dismissed):
lgtm

ja8zyjits (approved):
LGTM

PR #5659, opened 2026-07-16
fix(api): use savepoint for personal team cascade delete and reorder FK deletes

### LINKED ISSUE #5658: [BUG]: User deletion can raise StaleDataError or a raw FK IntegrityError instead of a clean error
## Description

`EmailAuthService.delete_user()` cascades deletes for a user's personal (single-member) team by deleting `EmailTeamMemberHistory` and `EmailTeamMember` rows via raw SQL, then calling `self.db.delete(team)`. Because these deletes aren't wrapped in a savepoint, the ORM's own cascade can attempt to re-delete already-removed member rows on commit, raising `StaleDataError`.

Separately, the team-membership cleanup for the deleted user runs *after* other FK-dependent deletes instead of before, which can trip a raw `IntegrityError` that isn't caught or translated into a user-facing error.

## Fix

- Wrap the personal-team cascade in `db.begin_nested()` and `db.expire(team, ["members"])` before the delete so the ORM re-queries instead of re-deleting.
- Move the "remove user from all team memberships" delete earlier, before other FK-dependent rows are removed.
- Catch `IntegrityError` around the final commit to raise `ValueError("Cannot delete user due to existing references")` instead of leaking a raw SQLAlchemy error.

gcgoncalves:
# 🐛 Bug-fix PR

## 🔗 Epic / Issue

Closes #5658

---

## 🚀 Summary (1-2 sentences)

Wrapped personal-team cascade delete in db.begin_nested() savepoint (fixes StaleDataError from ORM re-deleting already-removed member rows) and moved user's team-membership cleanup before other FK-dependent deletes, catching IntegrityError into a clean ValueError instead of leaking raw DB errors.

---

## 📏 Reviewability

### DIFF
--- a/mcpgateway/services/email_auth_service.py
+++ b/mcpgateway/services/email_auth_service.py
@@ -2139,11 +2139,11 @@ class EmailAuthService:
 
         except IntegrityError as e:
             self.db.rollback()
-            logger.error(f"FK constraint violation deleting user {SecurityValidator.sanitize_log_message(email)}: {e}")
+            logger.error("FK constraint violation deleting user %s: %s", SecurityValidator.sanitize_log_message(email), str(e))
             raise ValueError("Cannot delete user due to existing references") from e
         except Exception as e:
             self.db.rollback()
-            logger.error("Error deleting user %s: %s", SecurityValidator.sanitize_log_message(email), e)
+            logger.error("Error deleting user %s: %s", SecurityValidator.sanitize_log_message(email), e, exc_info=True)
             raise
 
     async def count_active_admin_users(self) -> int:
--- a/mcpgateway/services/email_auth_service.py
+++ b/mcpgateway/services/email_auth_service.py
@@ -2076,18 +2076,18 @@ class EmailAuthService:
 
                         if len(all_members) == 1 and all_members[0].user_email == email:
                             # This is a single-user personal team - cascade delete it
-                            logger.info(f"Deleting personal team '{SecurityValidator.sanitize_log_message(team.name)}' (single member: {SecurityValidator.sanitize_log_message(email)})")
-                            # Delete history records first: team_id FK has no ON DELETE CASCADE,
-                            # so history must be removed before the team is deleted.
-                            self.db.execute(delete(EmailTeamMemberHistory).where(EmailTeamMemberHistory.team_id == team.id))
-                            # Delete team members via raw SQL.
-                            self.db.execute(delete(EmailTeamMember).where(EmailTeamMember.team_id == team.id))
-                            # Expire the members collection so the ORM cascade re-queries the DB
-                            # (finding nothing) instead of re-deleting already-gone objects,
-                            # which would raise StaleDataError at commit time.
-                            self.db.expire(team, ["members"])
-                            # Delete the team
-                            self.db.delete(team)
+                            logger.info("Deleting personal team '%s' (single member: %s)", SecurityValidator.sanitize_log_message(team.name), SecurityValidator.sanitize_log_message(email))
+                            with self.db.begin_nested():
+                                # Delete history records first: team_id FK has no ON DELETE CASCADE,
+                                # so history must be removed before the team is deleted.
+                                self.db.execute(delete(EmailTeamMemberHistory).where(EmailTeamMemberHistory.team_id == team.id))
+                                # Delete team members via raw SQL.
+                                self.db.execute(delete(EmailTeamMember).where(EmailTeamMember.team_id == team.id))
+                                # Expire the members collection so the ORM cascade re-queries the DB
+                                # (finding nothing) instead of re-deleting already-gone objects,
+                                # which would raise StaleDataError at commit time.
+                                self.db.expire(team, ["members"])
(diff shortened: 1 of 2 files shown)


ja8zyjits (changes requested):
## 🚨 Blocking Issues

- [ ] **f-string in logger.error violates logging best practice**
  - File: `mcpgateway/services/email_auth_service.py:2143`
  - Current: `logger.error(f"FK constraint violation deleting user {SecurityValidator.sanitize_log_message(email)}: {e}")`
  - Issue: f-string evaluates even when logging disabled; exception `e` not sanitized
  - Fix: `logger.error("FK constraint violation deleting user %s: %s", SecurityValidator.sanitize_log_message(email), str(e))`
  - Impact: Performance penalty, potential info leak if exception contains sensitive data

ja8zyjits:
## ⚠️ Warnings (non blocking)

- [ ] **Exception chaining loses context in generic handler**
  - File: `mcpgateway/services/email_auth_service.py:2145-2148`
  - Current: `except Exception as e: ... raise`
  - Issue: Re-raises without `from e`, loses original traceback
  - Fix: `raise` alone preserves traceback (already correct), but consider logging full exception details
  - Impact: Minor - bare `raise` preserves traceback, but explicit `from e` would be clearer

- [ ] **Mock execute side_effect count mismatch risk**
  - File: `tests/unit/mcpgateway/services/test_email_auth_basic.py:2721-2728`
  - Issue: Comment says "Delete team members (for the team)" then "Delete team members (remove user from all teams)" but both map to same `mock_empty`
  - Risk: If implementation changes execute order, test may pass incorrectly
  - Fix: Use distinct mock objects or add assertions on execute call args
  - Impact: Low - test currently works but fragile to refactoring

gcgoncalves:
@ja8zyjits addressed!

ja8zyjits (approved):
LGTM

msureshkumar88 (approved):
## Real-process e2e verification — passed 12/12

Ran an independent e2e check against the actual bug trigger from #5658 (a user who owns a single-member personal team gets deleted) using a real gateway process, real SQLite file, real bootstrap admin, and real JWT auth — no mocking, no `TestClient`. Script: `tests/live_gateway/e2e_delete_user_personal_team_cascade.sh` (kept local, not pushed to this branch).

**Config used**
```
DATABASE_URL=sqlite:///<tmp>/e2e.db
AUTH_REQUIRED=true EMAIL_AUTH_ENABLED=true
PLATFORM_ADMIN_EMAIL=admin@example.com   (real bootstrap-on-startup admin)
LOG_LEVEL=INFO
uv run uvicorn mcpgateway.main:app --host 127.0.0.1 --port 18091
```

**Commands run**
```
uv run python3 -m mcpgateway.utils.create_jwt_token -u admin@example.com --secret $JWT_SECRET_KEY -e 60
curl -X POST /auth/email/admin/users   # create target user -> auto-creates single-member personal team
curl -X DELETE /auth/email/admin/users/{email}   # the exact call that used to raise StaleDataError/IntegrityError
# + direct sqlite3 reads of email_users / email_teams / email_team_members / email_team_member_history
# + grep of the live server log for StaleDataError / IntegrityError
```

**Output (summary)**
- `POST /auth/email/admin/users` → `201`, personal team auto-created with exactly 1 member (`role=owner`) — confirms the exact precondition from the issue.
- `DELETE /auth/email/admin/users/{email}` → `200`, `{"success":true,...}` (previously this was the call that could 500 with a raw `StaleDataError`/`IntegrityError`).
- DB after delete: `email_users`, `email_teams` (personal team), `email_team_members`, `email_team_member_history` all `0` rows for the target — full cascade, no orphans.
- Server log: zero occurrences of `StaleDataError`/`IntegrityError`; log shows `Deleting personal team '...' (single member: ...)` (confirms the `begin_nested()` cascade branch was actually exercised, not skipped) followed by `User ... deleted permanently`.
- Result: **12/12 checks passed.**

This confirms the fix resolves #5658 under the real trigger condition end-to-end, not just at the unit-test/mock level.

---

Approving. A few small, non-blocking follow-ups from the code read (happy to open a separate issue for these instead if preferred):

1. The new `except IntegrityError` log line sanitizes `email` but logs `str(e)` raw — DBAPI error text can embed statement/parameter values. Matches an existing pattern elsewhere in the file, but worth wrapping in `SecurityValidator.sanitize_log_message(...)` for consistency with the rest of this function.
2. Current tests are all mock-based (`MagicMock` session) — consider adding a real-session integration test alongside this for regression coverage, since the bug is inherently about live ORM flush/cascade + FK-enforcement behavior that mocks can't reproduce.
3. `routers/email_auth.py`'s `DELETE /auth/email/admin/users/{email}` still catches everything generically and returns a fixed `500 "Failed to delete user"`, so the new clean `ValueError` message doesn't reach REST API callers (it does reach the admin UI, which surfaces it directly). Not introduced by this PR, but worth a fast-follow if useful.

Nice fix — the FK model check confirms this was a real, reachable bug (`EmailTeamMember.user_email` has no `ondelete=CASCADE`), not a hypothetical edge case.

Issue #5640, opened 2026-07-15
[UI-REWRITE]: Tag input improvements


a-effort:
## Overview

Replace the comma-separated `<Input>` used for tag entry in three forms with a chip-style `TagInput` built from primitives already in the repo (`CardTag`, `Command`, `Popover`, `Input`). Add autocomplete against the existing `/tags` endpoint.

## Current state

Three forms use the same pattern: a plain text input holding a comma-separated string, `split(",")` at submit time.

- `client/src/components/tools/ToolAdvancedSettings.tsx:208-214`
- `client/src/components/resources/ResourceForm.tsx:283-290`
- `client/src/components/gateways/CreateServerForm.tsx:226-240`

Form state is a single `string` in `useToolForm.ts`, `useResourceForm.ts`, `usePromptForm.ts`, and `useCreateServerForm.ts`. Parsing lives in `parseTags` (`usePromptForm.ts:62-67`).

## Problems

- No visual confirmation a tag was created
- Whitespace and case duplicates go unnoticed (`Auth` vs `auth `)
- No autocomplete, despite `/tags` already existing (`client/src/generated/types/listTagsTagsGetParams.ts`)
- Mid-string edits are fiddly; backspace deletes characters, not tags
- Paste of newline or tab delimited lists does not split
- Screen readers announce one string, not discrete tags

## Proposal

Add `client/src/components/ui/tag-input.tsx`, composed from primitives already in the repo (no new dependency):

- `CardTag` for chip rendering
- `Command` + `Popover` for the suggestion list
- `Input` as the trailing text field

### Behavior

- Commit: `Enter`, `,`, `Tab`, blur
- Remove last: `Backspace` on empty input
- Remove any: click × on the chip
- Paste: split on `,`, `\n`, `\t`; trim; commit all
- Dedupe: case insensitive, trimmed, silent no-op
- Autocomplete: fetch `/tags` on focus (cached); "Create `<value>`" for unmatched input
- Props for `maxTags`, `disabled`, `aria-invalid`, `aria-describedby`

### Accessibility

- Container `role="group"` labelled by the field label
- Each chip focusable; `Delete` / `Backspace` on focused chip removes it
- `Command` combobox handles suggestion ARIA

### API sketch

```tsx
type TagInputProps = {
  id?: string;
  value: string[];
  onChange: (next: string[]) => void;
  suggestions?: string[];
  placeholder?: string;
  maxTags?: number;
  disabled?: boolean;
[...]

PR #5662, opened 2026-07-16
feat(ui-rewrite): add a new tags inline component to each details panel

### LINKED ISSUE #5641: [UI-REWRITE]: Add tag to details drawer
[Add tag interaction](https://www.figma.com/design/H2Dr9QPvRVxCe92V8URvGT/Refiring?m=auto&node-id=4846-49081&t=pWQ1dQzXGyeG3QKr-1) pattern  proposal (inline option, any of the details drawers)

https://github.com/user-attachments/assets/9b9bb818-1548-401f-a3e0-a3557ce48056

marekdano:
## Summary

Adds an inline "add tag" interaction to every details drawer — Tools, Resources, Prompts, MCP Servers, and Virtual Servers; so users can tag entities directly from the drawer without opening the edit form. Implements the pattern from #5641.

Closes #5641

https://github.com/user-attachments/assets/a83e53f7-9834-464d-8be5-c4b4b9948cbc

## Behavior

  - The Tags row shows the existing chips plus a compact **+ add** trigger.
  - Clicking it reveals an inline input (placeholder _"Add tags separated with commas"_) with **Cancel** / **Add** buttons rendered full-width **below** the label.
  - Multiple comma-separated tags can be added at once; entries are trimmed and de-duplicated (case-insensitive) against existing tags and against each other before saving.
  - **Add** shows an **Adding…** pending state and is disabled while empty or in flight; **Cancel** / Escape collapse the editor; Enter submits.
  - On success the new chips appear immediately; on failure the editor stays open (so the value can be retried) and an error toast is shown.

## Implementation

  - **New reusable component** `client/src/components/ui/inline-tag-add.tsx` — owns the whole Tags detail row (label + chips-as-children + inline editor) so the editor can span the full width beneath the label. Labels default in-component and are overridable.
  - **API** — a tags-only partial update per entity: `toolsApi.updateTags`, `resourcesApi.updateTags`, `promptsApi.updateTags`, `serversApi.updateTags` (`PUT /gateways/{id}`), and `updateVirtualServerTags` (`PUT /servers/{id}`). Each sends only `{ tags }`; the backend update services preserve all other fields.
  - **Panels/pages** — each of the five details panels gained an `onAddTag` prop wired to a page handler that calls the API and **patches the local cache in place** (no full-catalog refetch), then relies on the existing panel re-sync. Also fixes the Prompts drawer, which rendered from a stale `displayGroup` snapshot — it now re-resolves the active group from the live data so added tags appear immediately.
  - **i18n** — added `*.tags.addError` messages across `en-US`, `es-ES`, and `pt-BR`.

  ## Tests

  - **Component**: comma-split/dedup, Cancel/Add, success, error-retry, keyboard, pending state.
  - **API**: `updateTags` for all five (path, tags-only body, returned entity, ID validation).
  - **Panels**: `onAddTag` wiring for all five drawers.
  - **Pages (end-to-end)**: Prompts (regression for the stale-drawer bug), Tools, and Resources — add a tag, confirm the chip renders and the list is not re-fetched.

### DIFF
--- a/client/src/api/prompts.test.ts
+++ b/client/src/api/prompts.test.ts
@@ -94,4 +94,33 @@ describe("promptsApi", () => {
       );
     });
   });
+
+  describe("updateTags", () => {
+    it("PUTs /prompts/:id with a tags-only body and returns the updated prompt", async () => {
+      const updated = { id: "prompt-1", tags: [{ id: "draft", label: "draft" }] };
+      mockFetch.mockResolvedValueOnce(
+        new Response(JSON.stringify(updated), {
+          status: 200,
+          headers: { "Content-Type": "application/json" },
+        }),
+      );
+
+      const result = await promptsApi.updateTags("prompt-1", ["draft"]);
+
+      expect(mockFetch).toHaveBeenCalledWith(
+        expect.stringContaining("/prompts/prompt-1"),
+        expect.objectContaining({
+          method: "PUT",
+          body: JSON.stringify({ tags: ["draft"] }),
+        }),
+      );
+      expect(result).toEqual(updated);
+    });
+
+    it("throws synchronously for an invalid prompt ID", () => {
+      expect(() => promptsApi.updateTags("../etc/passwd", ["x"])).toThrow(
+        "Invalid prompt ID format",
+      );
+    });
+  });
 });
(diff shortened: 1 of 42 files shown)


a-effort:
LGTM! 🚀

Non blocking improvements that can be done as a follow up added here: https://github.com/IBM/mcp-context-forge/issues/5640#issuecomment-4996465484

a-effort (approved):

PR #5668, opened 2026-07-17
fix(search): token-scope team results for admins in unified search

### LINKED ISSUE #5701: [BUG]: Team search ignores explicit token team scope for admin callers
### Description

The admin branch of unified/team search (`admin_search_teams`, reached via `GET /v1/search`, `GET /admin/search`, and `GET /admin/teams/search`) returns *all* teams and ignores the caller's normalized token scope. So a DB-admin using an API token explicitly narrowed to a team subset still sees every team, not just the scoped ones.

This is the admin-side counterpart to the non-admin team-scope leak fixed in #5610; Layer-1 token scoping should constrain visibility independently of admin status.

### Impact

- A token deliberately scoped to a subset of teams does not actually restrict admin team search results (over-exposure of team id/name/slug/description/visibility/status).
- Only affects **explicitly-scoped** admin tokens. Admin UI session tokens and unscoped admin tokens resolve to `token_teams=None` (full bypass) and are unaffected.

### Steps to reproduce

1. As a DB admin, mint an API token with `teams: ["<team-B-id>"]`.
2. `GET /v1/search?q=<term>&entity_types=teams` with that token.
3. Observe teams outside team B still returned.

**Expected:** results limited to team B (plus the caller's personal team).
**Actual:** all matching teams returned.

### Notes

- Fix + tests: #5668.
- `token_teams is None` (full bypass) remains the only unrestricted case.

gandhipratik203:
## Summary

Stacked follow-up to #5610. That PR fixed the team-scope leak in unified search for the **non-admin** path; this one closes the same gap for the **admin** path.

Closes #5701

## What this PR does

An admin's search now only shows the teams their token is allowed to see.

```
Say there are 2 teams: "alpha" and "beta".

Admin token with NO scope  (teams: null)
   search teams  ─►  alpha, beta      ← sees everything (unchanged)

Admin token SCOPED to alpha  (teams: ["alpha"])
   BEFORE this PR:  search teams ─► alpha, beta   ❌ leak: beta shouldn't show
   AFTER  this PR:  search teams ─► alpha          ✅ beta correctly hidden
```

In one sentence: if you narrow an admin token to certain teams, team search now respects that narrowing instead of returning every team.

## Problem

The admin branch of `admin_search_teams` returns *all* teams (the admin management view) and ignored token scope. So a DB-admin using a token **explicitly narrowed** to a team subset still saw every team through `/v1/search`, `/admin/search`, and `/admin/teams/search`. This is the same class of leak the reviewer flagged on #5610, for admins.

## Fix

Honor explicit `token_teams` even for admins (Layer 1 constrains visibility independently of admin status). The scope is pushed into the `list_teams` query (a new optional `team_ids` filter) so it is enforced **before** pagination:

- `token_teams is None` → full admin bypass, unchanged (sees all teams).
- explicit list → results narrowed to exactly those team IDs.
- `token_teams == []` → public-only: **no teams** (no personal-team fallback, matching `normalize_token_teams()` / `get_team_from_token()`).

Applying the scope in the query fixes two issues raised in review:
- **No leak of the personal team** under an explicit/`[]` scope (the earlier `is_personal` carve-out is gone; the non-admin branch was aligned the same way).
- **No pagination gap** — previously the admin branch fetched the first `limit` teams and then filtered, so a scoped team that sorted past the first page was dropped; now the filter is in the query, so it is always returned.

Admin UI **session tokens** and **unscoped admin API tokens** both resolve to `None`, so they are unaffected. Only an admin who deliberately minted a team-scoped token changes behavior — which is the correct Layer-1 result.

## Affected surface

`admin_search_teams` admin branch (via `GET /admin/teams/search`, `GET /v1/search`, `GET /admin/search`), plus a small additive change to `TeamManagementService.list_teams` (new optional `team_ids` filter; `None` = no filter, so existing callers are unchanged). Behavior change is limited to admin callers whose token carries an explicit `token_teams`.

**Also changes the non-admin path** (which shipped in #5610): the `is_personal` carve-out is removed there too, so an explicitly-scoped non-admin token (including `[]` = public-only) no longer gets guaranteed personal-team visibility in search. Unscoped tokens are unaffected (memberships already include the personal team); this matches `normalize_token_teams()` / `get_team_from_token()`, which define `[]` as public-only with no personal fallback.

## Tests

Real-data tests through `/v1/search` (the `_real_data_env` fixture seeds an admin user and a third team the admin isn't a member of):

- unscoped admin → sees all teams, including the one they don't belong to (the admin all-teams branch runs);
- token scoped to one team → narrows to it, the others hidden;
- public-only admin token (`token_teams=[]`) → sees **no** teams (not a bypass view);
- admin scoped to a team that sorts **past** the per-page limit → still returned (scope applied in the query, before pagination).

Verified empirically: disabling the `list_teams` `team_ids` filter makes all three scope-dependent tests fail (narrowing leaks, public-only non-empty, past-page team dropped).

## Verification

- Integration: `pytest tests/integration/test_search_endpoint.py --with-integration -v` → 16 passed.
- `list_teams` change is backward-compatible: team-service + admin-teams suites green (155 passed). ruff clean.

### Manual verification (live server)

Run from the repo root on this branch with the server up (`make dev`, `:8000`). No tokens are pasted by hand; they flow through shell vars. The `mint` helper signs with the server's own `JWT_SECRET_KEY` (read from settings), and team ids are captured into vars.

```bash
export BASE=http://localhost:8000

# mint '<teams>' -> admin JWT; teams is  null  (bypass)  or  ["team-id", ...]  (scoped)
mint() { .venv/bin/python - "$1" <<'PY'
import sys, time, uuid, json, jwt
now = int(time.time())
print(jwt.encode({
  "sub":"admin@example.com","iat":now,"exp":now+3600,
  "iss":"mcpgateway","aud":"mcpgateway-api","jti":uuid.uuid4().hex,
  "user":{"email":"admin@example.com","is_admin":True,"auth_provider":"local"},
  "teams": json.loads(sys.argv[1]),
}, settings.jwt_secret_key.get_secret_value(), algorithm="HS256"))
PY
}

# unscoped admin token (teams:null = full bypass) + two demo teams
export ADMIN=$(mint 'null')
curl -sL -o /dev/null -H "Authorization: Bearer $ADMIN" -H "Content-Type: application/json" -d '{"name":"qatest alpha","visibility":"public"}' "$BASE/v1/teams/"
curl -sL -o /dev/null -H "Authorization: Bearer $ADMIN" -H "Content-Type: application/json" -d '{"name":"qatest beta","visibility":"public"}'  "$BASE/v1/teams/"

# capture team ids into vars, then mint a token scoped to alpha only
eval "$(curl -s -H "Authorization: Bearer $ADMIN" "$BASE/v1/search?q=qatest&entity_types=teams" | .venv/bin/python -c '
import sys,json
t={x["name"]:x["id"] for x in json.load(sys.stdin)["results"]["teams"]}
print("export ALPHA=%s" % t.get("qatest alpha",""))
print("export BETA=%s"  % t.get("qatest beta",""))')"
export SCOPED=$(mint "[\"$ALPHA\"]")

# the test
curl -s -H "Authorization: Bearer $ADMIN"  "$BASE/v1/search?q=qatest&entity_types=teams" | .venv/bin/python -c 'import sys,json;print([x["name"] for x in json.load(sys.stdin)["results"]["teams"]])'
curl -s -H "Authorization: Bearer $SCOPED" "$BASE/v1/search?q=qatest&entity_types=teams" | .venv/bin/python -c 'import sys,json;print([x["name"] for x in json.load(sys.stdin)["results"]["teams"]])'
```

Result:

```
UNSCOPED admin (teams: null):     ['qatest alpha', 'qatest beta']   # full bypass, sees all
SCOPED admin   (teams: [alpha]):  ['qatest alpha']                  # narrowed, beta hidden
```

The only difference between the two calls is the token's `teams` claim. On `feat/v1-search-endpoint` (without this fix), the scoped token returns both teams — that is the leak this PR closes.

### DIFF
--- a/mcpgateway/admin.py
+++ b/mcpgateway/admin.py
@@ -5587,6 +5587,14 @@ async def admin_search_teams(
         )
         # Result is dict {data, pagination...} (since page provided)
         teams = result["data"]
+        # Honor explicit token narrowing even for admins (Layer 1 constrains
+        # visibility independently of RBAC/admin status). token_teams is None for
+        # full admin bypass (unrestricted); an explicit list (including []) scopes
+        # the result. The caller's own personal team stays visible.
+        raw_token_teams = user.get("token_teams")
+        if raw_token_teams is not None:
+            scoped_team_ids = {team["id"] if isinstance(team, dict) else team for team in raw_token_teams}
+            teams = [t for t in teams if getattr(t, "is_personal", False) or t.id in scoped_team_ids]
     else:
         # Non-admin search
         # Reuse user team fetching
--- a/mcpgateway/admin.py
+++ b/mcpgateway/admin.py
@@ -5588,9 +5588,9 @@ async def admin_search_teams(
         # pagination (an allowed team must not be dropped for sorting past the
         # first page) and so an explicit scope no longer surfaces the personal team.
         raw_token_teams = user.get("token_teams")
-        scoped_team_ids: Optional[list[str]] = None
+        admin_scoped_team_ids: Optional[list[str]] = None
         if raw_token_teams is not None:
-            scoped_team_ids = [team["id"] if isinstance(team, dict) else team for team in raw_token_teams]
+            admin_scoped_team_ids = [team["id"] if isinstance(team, dict) else team for team in raw_token_teams]
         result = await team_service.list_teams(
             page=1,
             per_page=limit,
@@ -5599,7 +5599,7 @@ async def admin_search_teams(
             include_personal=False,
             search_query=search_query,
             personal_owner_email=user_email,
-            team_ids=scoped_team_ids,
+            team_ids=admin_scoped_team_ids,
         )
         # Result is dict {data, pagination...} (since page provided)
         teams = result["data"]
--- a/mcpgateway/admin.py
+++ b/mcpgateway/admin.py
@@ -5581,20 +5581,28 @@ async def admin_search_teams(
     # The CALLER (admin.py) distinguishes.
 
     if current_user.is_admin:
-        # Admin sees all non-personal teams plus their own personal team (single query)
-        result = await team_service.list_teams(
-            page=1, per_page=limit, include_inactive=include_inactive, visibility_filter=visibility, include_personal=False, search_query=search_query, personal_owner_email=user_email
-        )
-        # Result is dict {data, pagination...} (since page provided)
-        teams = result["data"]
         # Honor explicit token narrowing even for admins (Layer 1 constrains
         # visibility independently of RBAC/admin status). token_teams is None for
         # full admin bypass (unrestricted); an explicit list (including []) scopes
-        # the result. The caller's own personal team stays visible.
+        # the result. The scope is pushed into the query so it applies before
+        # pagination (an allowed team must not be dropped for sorting past the
+        # first page) and so an explicit scope no longer surfaces the personal team.
         raw_token_teams = user.get("token_teams")
+        scoped_team_ids: Optional[list[str]] = None
         if raw_token_teams is not None:
-            scoped_team_ids = {team["id"] if isinstance(team, dict) else team for team in raw_token_teams}
-            teams = [t for t in teams if getattr(t, "is_personal", False) or t.id in scoped_team_ids]
+            scoped_team_ids = [team["id"] if isinstance(team, dict) else team for team in raw_token_teams]
+        result = await team_service.list_teams(
+            page=1,
+            per_page=limit,
+            include_inactive=include_inactive,
+            visibility_filter=visibility,
+            include_personal=False,
+            search_query=search_query,
+            personal_owner_email=user_email,
+            team_ids=scoped_team_ids,
+        )
+        # Result is dict {data, pagination...} (since page provided)
+        teams = result["data"]
     else:
         # Non-admin search
         # Reuse user team fetching
--- a/mcpgateway/admin.py
+++ b/mcpgateway/admin.py
@@ -5611,12 +5611,14 @@ async def admin_search_teams(
         # returns every membership and ignores token scope, so a token narrowed to
         # a team subset would otherwise leak sibling teams the caller belongs to but
         # is scoped out of. _get_user_team_ids honors token_teams/_cached_team_ids;
-        # the caller's own personal team stays visible (owner is always visible).
+        # an unscoped caller's own memberships (including their personal team) are in
+        # this set, while an explicit scope (including [] = public-only) does not add
+        # a personal-team fallback, matching normalize_token_teams()/get_team_from_token().
         scoped_team_ids = set(await _get_user_team_ids(user, db))
         # Filter in memory
         filtered = []
         for t in all_teams:
-            if not getattr(t, "is_personal", False) and t.id not in scoped_team_ids:
+            if t.id not in scoped_team_ids:
                 continue
             if not include_inactive and not t.is_active:
                 continue
--- a/mcpgateway/services/team_management_service.py
+++ b/mcpgateway/services/team_management_service.py
@@ -1856,6 +1856,7 @@ class TeamManagementService:
         include_personal: bool = False,
         search_query: Optional[str] = None,
         personal_owner_email: Optional[str] = None,
+        team_ids: Optional[List[str]] = None,
     ) -> Union[Tuple[List[EmailTeam], Optional[str]], Dict[str, Any]]:
         """List teams with pagination support (cursor or page based).
 
@@ -1871,6 +1872,7 @@ class TeamManagementService:
             include_personal: Whether to include personal teams
             search_query: Search term for name/slug/description
             personal_owner_email: When set (and include_personal=False), includes this user's personal team alongside non-personal teams
+            team_ids: When set, restrict results to these team IDs before pagination (e.g. token-scoped callers). An empty list matches no teams.
 
         Returns:
             Union[Tuple[List[EmailTeam], Optional[str]], Dict[str, Any]]:
@@ -1886,6 +1888,11 @@ class TeamManagementService:
             personal_owner_email=personal_owner_email,
             search_description=True,
         )
+
+        # Restrict to specific team IDs before pagination so token-scoped callers
+        # are filtered in the query rather than on an already-limited page.
+        if team_ids is not None:
+            query = query.where(EmailTeam.id.in_(team_ids))
 
         # Choose ordering based on pagination mode:
         # - Page-based (UI): alphabetical by name for user-friendly display
--- a/tests/integration/test_search_endpoint.py
+++ b/tests/integration/test_search_endpoint.py
@@ -287,15 +287,17 @@ def _real_data_env():
     now = datetime.now(timezone.utc)
     team_a = uuid.uuid4().hex
     team_b = uuid.uuid4().hex
+    team_c = uuid.uuid4().hex  # exists but user_b is NOT a member; only visible via the admin all-teams view
     # Unique emails per invocation: get_user_teams and role/permission lookups are
     # cached by email at process scope, so a fixed email would leak one test's
     # temp-DB team ids into another test using the same user.
     suffix = uuid.uuid4().hex
     user_b = f"user-b-{suffix}@example.com"
+    admin_user = f"admin-{suffix}@example.com"
     owner = f"owner-{suffix}@example.com"  # owns the tools so user_b access is never via ownership
 
-    def _user(email):
-        return EmailUser(id=uuid.uuid4().hex, email=email, password_hash="x", full_name=email, is_admin=False, is_active=True, auth_provider="local", email_verified_at=now)  # pragma: allowlist secret
+    def _user(email, is_admin=False):
+        return EmailUser(id=uuid.uuid4().hex, email=email, password_hash="x", full_name=email, is_admin=is_admin, is_active=True, auth_provider="local", email_verified_at=now)  # pragma: allowlist secret
 
     def _tool(name, visibility, team_id):
         return Tool(
@@ -316,17 +318,19 @@ def _real_data_env():
         )
 
     db = TestSessionLocal()
-    db.add_all([_user(user_b), _user(owner)])
+    db.add_all([_user(user_b), _user(owner), _user(admin_user, is_admin=True)])
     db.add_all(
         [
             EmailTeam(id=team_a, name=f"{SEARCH_TERM} Team A", slug="team-a", created_by=owner, is_personal=False, visibility="public"),
             EmailTeam(id=team_b, name=f"{SEARCH_TERM} Team B", slug="team-b", created_by=owner, is_personal=False, visibility="public"),
+            EmailTeam(id=team_c, name=f"{SEARCH_TERM} Team C", slug="team-c", created_by=owner, is_personal=False, visibility="public"),
         ]
     )
     db.commit()
 
-    # user_b is a real member of BOTH teams (so team-A is genuinely accessible),
-    # and holds a real GLOBAL tools.read role (Layer 2 pass-through).
+    # user_b is a real member of teams A and B (so team-A is genuinely accessible),
+    # and holds a real GLOBAL tools.read+teams.read role (Layer 2 pass-through).
+    # admin_user is a DB admin holding the same role (teams.read has no admin bypass).
     role_id = uuid.uuid4().hex
     db.add_all(
         [
@@ -334,6 +338,7 @@ def _real_data_env():
             EmailTeamMember(id=uuid.uuid4().hex, team_id=team_b, user_email=user_b, role="member", is_active=True),
             Role(id=role_id, name="test-reader", scope="global", permissions=["tools.read", "teams.read"], created_by=owner, is_active=True),
             UserRole(id=uuid.uuid4().hex, user_email=user_b, role_id=role_id, scope="global", scope_id=None, granted_by=owner, is_active=True),
+            UserRole(id=uuid.uuid4().hex, user_email=admin_user, role_id=role_id, scope="global", scope_id=None, granted_by=owner, is_active=True),
         ]
     )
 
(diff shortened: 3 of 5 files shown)


msureshkumar88 (changes requested):
Thanks for the fast follow-up on #5610 — the query-level fix (filtering `team_ids` before pagination instead of after) is the right call, and the progressive commits addressing the past-page and public-only cases show good self-review.

**Blocking — same leak still open in 3 other admin team-listing paths**

This closes the leak for search, but the identical pattern (admin bypass ignoring narrowed `token_teams`) is still present elsewhere:
- `admin.py:5711` (`admin_teams_partial_html`, `/admin/teams/partial`)
- `admin.py:5930` (`/admin/teams` unified view)
- `routers/teams.py:214` (`GET /teams`) — this one's a bit sharper: `check_platform_admin_permission` only denies bypass for `token_teams == []`, so a narrowed *non-empty* list (e.g. `["team-B"]`) still gets `has_admin_team_access=True` and an unscoped `list_teams()` call.

Given the PR's own framing ("Layer 1 constrains visibility independently of admin status"), these three look like the same bug under #5701's umbrella rather than out of scope. Worth a follow-up applying the same `team_ids=scoped_team_ids` pattern, or at minimum tracking it as a fast-follow issue so it doesn't quietly linger.

**Blocking — mypy: variable reused with incompatible type across branches**

`admin.py:5591` types `scoped_team_ids: Optional[list[str]] = None` in the admin branch; the non-admin branch (`admin.py:5617`) reassigns the same name to `set(await _get_user_team_ids(...))`. Confirmed against `main` — this is new, not pre-existing:
```
admin.py:5617: error: Incompatible types in assignment (expression has type "set[Any]", variable has type "list[str] | None")
admin.py:5621: error: Unsupported right operand type for in ("list[str] | None")
```
Harmless at runtime since the branches are mutually exclusive, but breaks strict mypy. Easy fix — just rename one of the two (e.g. `admin_scoped_team_ids` / `member_team_ids`).

**Worth a mention — second behavior change beyond #5701's scope**

The third commit also strips the `is_personal` carve-out from the *non-admin* branch (shipped just yesterday in #5610), so an explicitly-scoped non-admin token (including `[]`) now loses guaranteed personal-team visibility in search too. Makes sense given the token model, but it's a second, undisclosed behavior change riding on an admin-only issue — a one-line callout in the PR description (and a CHANGELOG entry, since #5701 is `security`-labeled) would help anyone auditing this later.

**Nit**

`admin.py:5592-5593` re-implements the dict/str team-id extraction that `_get_user_team_ids` (`admin.py:1902-1913`) already does — could just call that helper instead of duplicating it.

Nothing here blocks the core fix's correctness for the reported case — tests for the admin path (unscoped/scoped/public-only/past-page) are solid — but I'd like the completeness gap and the mypy issue addressed before merge.

gandhipratik203:
Both blocking items addressed:

- **mypy:** fixed in 4188d0726 (renamed admin-branch var to `admin_scoped_team_ids`). The collision was new to this PR (0 on `main`); remaining admin.py mypy findings are pre-existing.
- **Same leak in the 3 other paths:** agreed, same bug under #5701's umbrella. Tracked as fast-follow #5812 (incl. the sharper `routers/teams.py:214` bypass), kept out of this PR to keep the diff to the search fix.

gandhipratik203:
Added a callout in the **Affected surface** section noting the non-admin `is_personal` carve-out removal (shipped in #5610) is a second behavior change here. Leaving CHANGELOG to release-time curation, per the convention the earlier v1 endpoint PRs (#5455, #5443) followed.

msureshkumar88 (changes requested):
Nice work closing the admin-side counterpart to #5610 — the query-level `team_ids` filter is the right shape (fixes the pagination-drop bug too), and the four new integration tests exercise exactly the scenarios called out in the PR description. The `[] → zero rows` and `None → full bypass` semantics line up cleanly with `normalize_token_teams()`.

A couple of things worth resolving before merge, plus a couple of suggestions:

## Blocking

**1. CI is currently red — diff-cover gate fails (71% vs. 93% required)**

Uncovered lines are exactly the new logic:
- `mcpgateway/admin.py:5593` — the dict-form branch of the `token_teams` → id extraction (`team["id"] if isinstance(team, dict) else team`)
- `mcpgateway/services/team_management_service.py:1895` — the new `team_ids` WHERE clause itself

The integration tests cover this behaviorally but apparently aren't counted by the coverage job (likely gated behind `--with-integration`). Could you add a couple of fast unit tests that hit these two lines directly — including a case where `token_teams` contains dict-shaped entries (`[{"id": "..."}]`), not just plain strings?

**2. Same bug class survives in 3 sibling endpoints**

`TeamManagementService.list_teams()` has four callers; this PR wires `team_ids` scoping into only one:

| Caller | Scoped? |
|---|---|
| `admin_search_teams` (`/v1/search`, `/admin/search`, `/admin/teams/search`) | fixed here |
| `admin_teams_partial_html` (`GET /admin/teams/partial`) | not fixed |
| `GET /admin/teams` (admin.py) | not fixed |
| `GET /v1/teams/` (routers/teams.py) | not fixed |

All three gate on `current_user.is_admin` / `check_platform_admin_permission()`, neither of which restricts the query for a *non-empty* explicit scope — `check_platform_admin_permission()` only suppresses bypass on `token_teams == []`. So a token scoped to `["team_b"]` still passes the admin gate on these three routes and gets every team back. `token_scoping.py`'s route map is Layer-2 permission routing only, it doesn't do Layer-1 query filtering, so it doesn't cover this either.

Since #5701's root cause is "Layer 1 should constrain visibility independently of admin status," and `team_ids=` is already plumbed through `list_teams()`, would you be open to threading it through these three call sites in this PR (or a fast-follow issue linked from here, so it isn't lost)? Happy to see whichever path you prefer.

## Suggestions (non-blocking)

**3.** The non-admin branch also drops the `is_personal` carve-out (already-shipped in #5610), which is a real behavior change to a released path — makes sense for consistency with `normalize_token_teams()`, but it's currently just prose in the PR body. Could this get a one-line callout at the top of the summary (and maybe its own issue reference) so it's not easy to miss in the diff?

**4.** `admin.py:5590-5593` re-implements the same `token_teams` → id-list extraction that `_get_user_team_ids` already does (admin.py:1902-1913). Not directly reusable since `_get_user_team_ids` falls back to full DB membership when the key is absent, but a small shared helper for just the extraction step would keep the two from drifting apart over time.

**5.** If `docs/docs/manage/rbac.md` or `docs/docs/architecture/multitenancy.md` describe the old personal-team-fallback behavior, worth a one-line update alongside item 3.

None of the suggestions are blocking — just the coverage gate and (strongly) the sibling-endpoint gap before this is fully closed.

gandhipratik203:
Thanks for the careful pass.

**Blocking 1 (diff-cover):** Fixed in `767fbd973` — two unit tests hitting those lines directly, including a mixed `token_teams=["tb", {"id": "tc"}]` for the dict-form branch (`admin.py:5593`) and one asserting the `team_ids` filter runs before pagination (`team_management_service.py:1895`). CI is green.

**Blocking 2 (sibling endpoints):** Same leak class, agreed — and good catch on `routers/teams.py:214` being the sharper bypass. Tracked as linked fast-follow #5812, which enumerates all three call sites and the `check_platform_admin_permission` fix. Kept out of this PR to hold the diff to the search fix; linked here so it isn't lost.

**Suggestions 3–5:** carve-out callout added referencing #5610; helper dedup (4) and docs wording (5) folded into #5812.

msureshkumar88:
Re-reviewed against the actual commits/diff (not review-thread replies) — checked out this branch locally and ran things directly rather than trusting the status claims.

**Blocking — mypy variable collision (admin.py:5591/5617):** Fixed in `4188d0726`. Ran `mypy mcpgateway/admin.py` myself — no incompatible-assignment error remains around those lines; admin branch now uses `admin_scoped_team_ids`, non-admin branch keeps `scoped_team_ids`.

**Blocking — diff-cover gate (admin.py:5593 / team_management_service.py:1895):** Fixed in `767fbd973`. Ran both new unit tests directly (`test_admin_search_teams_admin_scoped_token_forwards_team_ids`, `test_list_teams_team_ids_filters_before_pagination`) — both pass. `gh pr checks` confirms CI green: 30/30.

**Blocking — same leak in 3 sibling endpoints:** Deferred to #5812, which is open and enumerates all three sites (including the sharper `routers/teams.py:214` bypass) — matches what was called out here, not just a placeholder link.

**Non-blocking — PR description callout + suggestions 4/5:** Callout for the `is_personal` carve-out removal is in the current PR body; the helper-dedup and docs nits are captured in #5812.

Status: all blocking and non-blocking items verified resolved by direct inspection/execution. Approve.

msureshkumar88 (approved):
## E2E verification (live server, real HTTP, real JWTs — not TestClient/mocks)

Ran the actual bug scenario end-to-end against a real running instance of this branch, extending the manual-verification recipe from the PR description to also cover the public-only and past-page-limit cases.

### Config used

- Branch: `feat/v1-search-team-scope-admin` (HEAD `767fbd973`)
- Server: `make dev` (uvicorn --reload), `HOST=0.0.0.0 PORT=8000`
- DB: local SQLite dev DB (`.env` default), `AUTH_REQUIRED=true`, `REQUIRE_USER_IN_DB=true`
- Auth: real JWTs signed with the server's own `JWT_SECRET_KEY` (read via `settings.jwt_secret_key`), `sub=admin@example.com` (the configured `platform_admin_email` bootstrap identity, matching `REQUIRE_USER_IN_DB=true`)
- Teams created live via `POST /v1/teams/`, then queried via the real `GET /v1/search` endpoint over HTTP

### Commands run

```bash
make dev   # background, :8000
.venv/bin/python verify_pr5668.py
```

The script: mints an unscoped admin bootstrap token, creates 3 real teams (`alpha`, `beta`, `zeta`) via the live API, then re-mints tokens with different `teams` claims and asserts on `/v1/search?q=...&entity_types=teams` responses.

### Output

```
== PR #5668 e2e verification (run suffix: ead313f0) ==
Base URL: http://localhost:8000
Admin identity: admin@example.com (platform_admin_email bootstrap)
Created teams: ['e2e-ead313f0-alpha', 'e2e-ead313f0-beta', 'e2e-ead313f0-zeta']
Team ids -> alpha=6a97bac264b948e98d2860baaa2014ba beta=2b750c38bc7b41249760c7a67b7a8e93 zeta=a2966ca3ab064b6b80c21242f8f8c31d
[PASS] 1. Unscoped admin token (teams: null): expected all 3 test teams, got ['e2e-ead313f0-alpha', 'e2e-ead313f0-beta', 'e2e-ead313f0-zeta']
[PASS] 2. Admin scoped to one team (teams: [alpha]): expected only alpha, got ['e2e-ead313f0-alpha']
[PASS] 3. Public-only admin token (teams: []): expected zero teams, got []
[PASS] 4. Scoped team past the page limit (limit_per_type=1, teams: [zeta]): expected zeta present despite limit=1, got ['e2e-ead313f0-zeta']

== SUMMARY: 4/4 passed ==
```

All 4 scenarios called out in the PR description ("Tests" section) and in review reproduce correctly against a live server: unscoped bypass unchanged, explicit narrowing enforced, `[]` yields zero teams (no personal-team fallback), and the scope is applied before pagination so a team sorting past `limit_per_type=1` is still returned.

Note: this is a positive-path e2e run against this branch only; I did not additionally check out pre-fix `main`/`103c60073^` to reproduce the leak as a negative control — the unit/integration tests already do that (author's PR description: "disabling the `list_teams` `team_ids` filter makes all three scope-dependent tests fail").

**Result: validation passes. Approving.**

mcpgateway/admin.py
```diff
@@ -5581,9 +5581,25 @@ async def admin_search_teams(
...
+        # the result. The scope is pushed into the query so it applies before
+        # pagination (an allowed team must not be dropped for sorting past the
+        # first page) and so an explicit scope no longer surfaces the personal team.
+        raw_token_teams = user.get("token_teams")
+        admin_scoped_team_ids: Optional[list[str]] = None
+        if raw_token_teams is not None:
+            admin_scoped_team_ids = [team["id"] if isinstance(team, dict) else team for team in raw_token_teams]
         result = await team_service.list_teams(
-            page=1, per_page=limit, include_inactive=include_inactive, visibility_filter=visibility, include_personal=False, search_query=search_query, personal_owner_email=user_email
+            page=1,
```
vishu-bh, line 5595: Could we preserve platform-public team visibility here?

Repository policy defines `token_teams=[]` as public-only and `token_teams=["t1"]` as team + public. Passing these values directly to `team_ids` instead produces no teams or only `t1`, excluding matching public teams. Consider combining the scoped team-ID predicate with public visibility, with tests ensuring public teams remain visible while out-of-scope private teams stay hidden.

For example:
A token scoped to ["alpha"] also sees only alpha—even when other matching teams are platform-public.
Repository policy defines:
- [] → public-only
- ["t1"] → team + public
- public → platform-public scope

Current implementation instead gives:
- [] → no teams
- ["t1"] → only t1

vishu-bh (changes requested):
Change looks correct just needs this improvements

mcpgateway/admin.py
gandhipratik203: Fair reading of the token table. One nuance though: for a **team**, `visibility="public"` means *joinable/discoverable* (`discover_public_teams` / "can only request to join public teams"), not "public resource" — so the token policy's "team + public" (a resource-visibility rule) doesn't map 1:1 onto whether a team row shows in a *scoped* search.

More to the point, this is preexisting: team search has **never** surfaced non-member public teams. The non-admin path uses `get_user_teams(...)` (memberships only), which predates #5610 by years; #5610 only added scope-narrowing *within* memberships, and this PR mirrors that for admins. So showing public teams here would be **net-new** discovery behavior, not something this PR changed.

Given it's a preexisting behavior (and would also touch the merged non-admin path), I'd suggest tracking it as a follow-up rather than expanding this diff — happy to fold it into #5812. Does that work?

vishu-bh: That sounds fair. Approving now

vishu-bh (approved):
LGTM

PR #5672, opened 2026-07-17
fix: return 404 for missing server and gateway deletes

ja8zyjits:
# Bug-fix PR

## Summary

Fix DELETE requests for missing servers and gateways returning `403 Forbidden`
before route handlers can return `404 Not Found`.

## Reproduction Steps

Internal issue 431:

1. Create server or gateway.
2. Delete it successfully.
3. Repeat `DELETE /servers/{id}` or `DELETE /gateways/{id}` with a hexadecimal UUID.
4. Before fix: `403`. After fix: `404`.

## Root Cause

`TokenScopingMiddleware._check_resource_team_ownership()` returned `False` for
both missing rows and denied access. Middleware mapped both results to `403`.

## Fix Description

Add `ResourceOwnershipResult` with `ALLOWED`, `NOT_FOUND`, and `DENIED`.
Missing server and gateway rows return `NOT_FOUND`. Exact server/gateway DELETE
paths map only `NOT_FOUND` to `404`; denied ownership, unsupported resources,
and lookup errors remain `403`.

Direct callers in `main.py` and OAuth router now explicitly allow only
`ALLOWED`.

## Reviewability

## Verification

| Check | Command | Status |
|---|---|---|
| Focused unit tests | ` .venv/bin/python -m pytest tests/unit/mcpgateway/middleware/test_token_scoping.py tests/unit/mcpgateway/middleware/test_token_scoping_extra.py tests/unit/mcpgateway/middleware/test_token_scoping_normalization.py tests/unit/mcpgateway/test_main.py tests/unit/mcpgateway/test_main_extended.py tests/unit/mcpgateway/routers/test_oauth_router.py -q` | Passed |
| Changed source lint | `TARGET='mcpgateway/main.py mcpgateway/middleware/token_scoping.py mcpgateway/routers/oauth_router.py'` | Passed |
| Full lint suite | `make lint` | Not run |
| Full unit/integration suite | `make test` | Not run |
| Coverage | `make coverage` | Not run |
| Manual regression | Repeated DELETE behavior | Not run; focused middleware tests cover regression |

Note: test-file Ruff is blocked by pre-existing `E712` and unused-variable
findings in legacy test files outside this change.

## Checklist

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-07-17T15:09:05Z",
+  "generated_at": "2026-07-20T09:47:57Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -266,7 +266,7 @@
         "hashed_secret": "b4673e578b9b30fe8bba1b555b7b59883444c697",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1746,
+        "line_number": 1747,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -274,7 +274,7 @@
         "hashed_secret": "4a0a2df96d4c9a13a282268cab33ac4b8cbb2c72",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1834,
+        "line_number": 1835,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -282,7 +282,7 @@
         "hashed_secret": "5baa61e4c9b93f3f0682250b6cf8331b7ee68fd8",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 2184,
+        "line_number": 2185,
         "type": "Basic Auth Credentials",
         "verified_result": null
       },
@@ -290,7 +290,7 @@
         "hashed_secret": "fa9beb99e4029ad5a6615399e7bbae21356086b3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 3549,
+        "line_number": 3550,
         "type": "Basic Auth Credentials",
         "verified_result": null
       },
@@ -298,7 +298,7 @@
         "hashed_secret": "fa9beb99e4029ad5a6615399e7bbae21356086b3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 3640,
+        "line_number": 3641,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -306,7 +306,7 @@
         "hashed_secret": "ac371b6dcce28a86c90d12bc57d946a800eebf17",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 3683,
+        "line_number": 3684,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -314,7 +314,7 @@
         "hashed_secret": "0b6ec68df700dec4dcd64babd0eda1edccddace1",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 3688,
+        "line_number": 3689,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -322,7 +322,7 @@
         "hashed_secret": "4ad6f0082ee224001beb3ca5c3e81c8ceea5ed86",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 3693,
+        "line_number": 3694,
         "type": "Secret Keyword",
         "verified_result": null
       }
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-07-17T13:06:41Z",
+  "generated_at": "2026-07-17T14:55:45Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -382,7 +382,7 @@
         "hashed_secret": "d3ac7a4ef1a838b4134f2f6e7f3c0d249d74b674",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 5963,
+        "line_number": 5953,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -390,7 +390,7 @@
         "hashed_secret": "5932862bcd24dd27d0dc0407ec94fe9d6ea24aeb",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 6460,
+        "line_number": 6450,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -398,7 +398,7 @@
         "hashed_secret": "c77c805e32f173e4321ee9187de9c29cb3804513",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 6472,
+        "line_number": 6462,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -406,7 +406,7 @@
         "hashed_secret": "8fe3df8a68ddd0d4ab2214186cbb8e38ccd0e06a",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 6544,
+        "line_number": 6534,
         "type": "Secret Keyword",
         "verified_result": null
       },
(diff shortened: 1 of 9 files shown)


vishu-bh (dismissed):
LGTM 🚀

msureshkumar88 (changes requested):
Thanks for tracking this down — the root cause analysis is right on target, and scoping the fix to the exact `DELETE /servers/{id}` / `DELETE /gateways/{id}` case (rather than trying to fix every verb/resource-type combination at once) was the right call for a bug-fix PR. Ran the full middleware/oauth test suite plus ruff/bandit against this branch locally and everything passes clean.

A few things worth addressing before merge, plus a couple of follow-ups worth tracking separately:

## Requested changes

1. **Add a log line on the new 404 branch.** `_check_resource_team_ownership` still logs a `warning` on the 403 path, but the new `HTTPException(404, ...)` branch in `__call__` raises silently. Since this branch is now a way to distinguish "resource exists" from "resource doesn't exist" for scoped tokens (see security note below), a `logger.debug`/`logger.info` here would preserve an audit trail for detecting probing/enumeration behavior — cheap to add, meaningfully closes an observability gap.

2. **Add a test for the dashed-UUID branch of `_TARGETED_MISSING_DELETE_PATTERN`.** All current parametrized cases (`test_missing_targeted_delete_returns_404`, etc.) use the 32-char no-dash hex form. The regex also has a `[a-f0-9]{8}-(?:[a-f0-9]{4}-){3}[a-f0-9]{12}` alternative that's currently untested — a typo in that branch (wrong dash-group count, etc.) wouldn't be caught today.

3. **Add a CHANGELOG.md entry.** This is a user-visible status-code behavior change (403 → 404 for repeated/idempotent DELETE on already-gone servers/gateways) and the repo convention for bug fixes like this is to note it.

4. **Run the full unit suite before merge.** The PR's own verification table lists `make test` / `make lint` / `make coverage` as "Not run." Given this change sits inside `TokenScopingMiddleware`, which gates nearly every authenticated route, I'd want the full suite green rather than just the six directly-touched test files, to rule out a regression in some other caller of `_check_resource_team_ownership`'s return value.

## Follow-up (separate issue, not blocking this PR)

5. **The same NOT_FOUND-masked-as-403 bug is still present elsewhere.** `_enforce_scoped_resource_access()` in `main.py` (used by `GET /servers/{id}`, `GET /servers/{id}/sse`, `GET /tools/{id}`, `GET /resources/{id}` GET /info, `GET /gateways/{id}`) still treats `ResourceOwnershipResult.NOT_FOUND` as "not allowed" → 403, so a scoped token doing `GET` on a missing server/gateway still gets 403 instead of 404. Separately, the "not found" branches for tools/resources/prompts in `_check_resource_team_ownership` were left returning `DENIED` rather than `NOT_FOUND`, so they were never candidates for this fix either. Worth a tracked follow-up so the rest of this bug class doesn't get lost — happy to file it if there isn't one already.

6. **Minor refactor opportunity.** The "compute ownership → maybe raise 404 → maybe raise 403" block is duplicated verbatim across the two branches in `__call__` (shared-session vs. public-only-token path), differing only in the `db=db` kwarg. Pulling this into a small helper (e.g. `_enforce_ownership_or_raise(...)`) would remove the duplication and let you pass the already-normalized path into `_is_targeted_missing_resource_delete` instead of re-normalizing `request.url.path` a second time.

None of this is a fundamental problem with the approach — the enum-based `ResourceOwnershipResult` is a clean way to model this, and the DELETE-specific fix matches the reported repro exactly. Marking as request-changes mainly for items 1–4 above; happy to re-review quickly once those are in.

msureshkumar88:
Re-checked this against the earlier review (commit `c756986` → current `6fdf06a69`). All four requested changes are in:

1. **404-branch logging** — `logger.info("Scoped DELETE target not found: %s", ...)` added on both branches in `__call__`.
2. **Dashed-UUID coverage** — new parametrized tests cover the dashed-UUID form, the `/v1/` prefix, trailing slash, `app_root_path` prefix, DB-error-still-403, and non-targeted paths (GET/PUT, sub-paths, `/tools/...`) still returning 403.
3. **CHANGELOG entry** — added under `[Unreleased] > Fixed`.
4. **Full test suite** — ran locally: the six directly-touched files pass clean, and `ruff`/`bandit` are clean on the three changed source files. Ran the full `tests/unit` suite to rule out a regression elsewhere in `TokenScopingMiddleware`'s callers: 9 failures, all in files this PR never touches (`test_rate_limiter.py`, `test_observability.py`, `test_encoded_exfil_detector.py`) — pre-existing on this branch, unrelated to the token-scoping/ownership-result change.

Diff scope is clean — only `main.py`, `token_scoping.py`, `oauth_router.py`, their tests, CHANGELOG, and `.secrets.baseline` (timestamp/line-number churn only). No unrelated changes.

Non-blocking notes for a follow-up, not this PR:
- `_enforce_scoped_resource_access()` in `main.py` still maps `NOT_FOUND` → 403 for `GET`/`PUT` on servers/gateways (and `GET /tools`, `GET`/`PUT /resources`) — same bug class, different verb/route, still open.
- The ownership-check-and-raise block in `__call__` is still duplicated verbatim across the shared-session and public-only-token branches (differs only by `db=db`).
- No end-to-end/integration test exercises the actual repro (create → delete → delete again → assert 404) through a real app instance; current coverage is unit-level with `_check_resource_team_ownership` mocked.

No blockers found in this re-review. Holding off on approval for now — flagging status here for visibility.

msureshkumar88 (approved):
Ran a real end-to-end verification against this branch (commit `6fdf06a69`) — a real `uv run uvicorn` process, a real SQLite DB, real JWTs minted via `mcpgateway.utils.create_jwt_token`, real team/user rows created through the API, and real `curl` calls. No mocks. Script: `scripts/verify-pr-5672-e2e.sh` (local to this review, not part of the diff).

**Config used**
```
JWT_SECRET_KEY=e2e-pr5672-test-secret
AUTH_ENCRYPTION_SECRET=e2e-pr5672-test-salt
DATABASE_URL=sqlite:///<repo>/e2e_pr5672_test.db
AUTH_REQUIRED=true
MCPGATEWAY_UI_ENABLED=false
PLUGINS_ENABLED=false
GATEWAY_ASYNC_LIFECYCLE_ENABLED=true   # so gateway create/delete against an unreachable test URL doesn't block synchronously
FEDERATION_TIMEOUT=5
```

**Commands run** (abbreviated — full script does the rest via `curl`):
```
uv run uvicorn mcpgateway.main:app --host 127.0.0.1 --port 44672
uv run python -m mcpgateway.utils.create_jwt_token --username admin@example.com --admin ...
curl -X POST $BASE/v1/teams/ ...                       # create team A, team B
curl -X POST $BASE/auth/email/register ...              # real non-admin user
curl -X POST $BASE/v1/teams/$TEAM_A_ID/members ...       # real DB team membership
uv run python -m mcpgateway.utils.create_jwt_token --username <user> --teams $TEAM_A_ID ...   # team-scoped token
uv run python -m mcpgateway.utils.create_jwt_token --username <user> ...                       # public-only token (no --teams)
curl -X POST $BASE/servers ... / curl -X DELETE $BASE/servers/{id} ...
curl -X POST $BASE/gateways ... / curl -X DELETE $BASE/gateways/{id} ...
curl -X DELETE $BASE/tools/{missing-id} ...
curl $BASE/servers/{missing-id} ...   # GET
```

**Output — 20/20 checks passed:**
```
PASS: gateway server is up on :44672
PASS: admin JWT generated
PASS: created team A and team B
PASS: registered non-admin user
PASS: added scoped user as real member of team A
PASS: minted team-scoped and public-only tokens
PASS: created server S1 owned by team A
PASS: admin deletes S1 (200), row now gone
PASS: CORE CLAIM: repeated DELETE /servers/{id} now returns 404 for team-scoped token
PASS: created gateway G1 owned by team A
PASS: admin deletes G1 (202 accepted)
PASS: G1 row confirmed gone (background lifecycle task completed)
PASS: CORE CLAIM: repeated DELETE /gateways/{id} now returns 404 for team-scoped token
PASS: created public server S2
PASS: admin deletes S2 (200), row now gone
PASS: CORE CLAIM: repeated DELETE /servers/{id} returns 404 for public-only token too (else-branch code path)
PASS: created server S3 owned by team B (still exists)
PASS: REGRESSION GUARD: existing-but-not-owned server still returns 403 (denied ownership preserved)
PASS: S3 still exists after denied delete attempt
PASS: SCOPE GUARD: DELETE /tools/{missing-id} still 403 — fix correctly scoped to servers/gateways only

RESULT: ALL CHECKS PASSED (exit 0)
```

Also confirmed the documented, out-of-scope follow-up gap is still present as expected (informational, not a failure): `GET /servers/{missing-id}` with the same team-scoped token still returns 403, not 404 — `_enforce_scoped_resource_access()` in `main.py` is untouched by this PR, consistent with its stated scope.

This confirms the fix works exactly as described against a live instance: the reported repro (create → delete → delete again) now returns 404 instead of 403 on both `DELETE /servers/{id}` and `DELETE /gateways/{id}`, for both a team-scoped token and a public-only token, while denied-ownership (403) and out-of-scope paths are unchanged.

Combined with the earlier unit-test re-verification (all 4 previously requested changes confirmed in place, full suite green apart from pre-existing unrelated failures), approving.

PR #5737, opened 2026-07-21
refactor: remove Rust build infrastructure and stdio wrapper capability

jonpspri:
Removes the Rust workspace, all associated crates, and the Python stdio wrapper bridge, eliminating ~15k lines of deprecated and unused infrastructure.

## Rust build infrastructure
- Delete `crates/` (mcp_stdio_wrapper, request_logging_masking_native_extension)
- Delete `Cargo.toml`, `Cargo.lock`, `rust-toolchain.toml`, `deny.toml`
- Delete `supply-chain/` (cargo-vet config)
- Remove `rust.yml` CI workflow and `Cargo.lock` from dependency-review
- Remove 225+ lines of `rust-*` Makefile targets, `ENABLE_RUST_BUILD`, `serve-granian` targets
- Remove `rust-builder` stage and `ENABLE_RUST` args from Containerfiles
- Remove Rust references from AGENTS.md, README.md, SECURITY.md

## Stdio wrapper (Python + Rust)
- Delete `mcpgateway/wrapper.py` and `crates/wrapper/`
- Delete `wrapper.yml` CI workflow
- Update client-facing docs to use Streamable HTTP (`type: http`, `/mcp/`) as canonical transport
- Remove wrapper unit tests, e2e tests, and skipped test stubs
- Update `admin_ui/configExport.js` to drop stdio config generation

## Cleanups
- Remove dead `ResourceNameConflictError` import from `main.py`
- Remove orphaned `TestRuntimeAdminEmailExtraction` test class
- Remove `RUST_LOG` from docker-compose `fast_test_server`
- Prune dead entries from `MANIFEST.in` (alembic, infra, supply-chain)
- Clean up `.gitignore` (target/, *.rs.bk, debug_login_page.png)
- 117 files changed, ~15k deletions


lucarlig:
@jonpspri, good progress here: the PR does remove the actual first-party Rust workspace. At head `40b5f1f`, I found no tracked `.rs` files, `Cargo.toml`, `Cargo.lock`, `rust-toolchain`, `deny.toml`, or `supply-chain/` content.

A repo-wide audit found several related pieces that still need cleanup before this is a complete Rust removal:

**Blocking**
- `scripts/license_checker.py` still enables Rust dependency scanning by default and requires `cargo-license` even with zero Cargo manifests. This is why the current license check fails.
- `.bumpversion.cfg` still has `[bumpversion:file:Cargo.toml]`. I reproduced a `bump2version --dry-run patch` failure with `FileNotFoundError: Cargo.toml`.
- The `MANIFEST.in` pruning also removed required non-Rust Admin UI assets; the package build currently fails because `mcpgateway/admin_ui/index.js` is missing from the sdist.
- `.secrets.baseline` still contains Cargo/crate entries, causing the current pre-commit failure.

**Remaining first-party Rust infrastructure**
- `.dockerignore` re-includes deleted Rust server Cargo/source/Containerfile paths.
- `scripts/pre-commit/check_rust_workspace.py` remains.
- `scripts/compliance_matrix.py` still defines `rust_edge` and `rust_full`.
- Performance compose generation still targets deleted `mcp-servers/rust/fast-time-server` and `benchmark-server` paths.
- Secret-detection comparison scripts and Make targets still run "Rust vs Python."
- `test_fast_server_go_rust_parity.py` remains; both branches now select the same tool, so it no longer tests parity.
- Runtime compatibility remains in `version.py`, `main.py`, `runtime_state.py`, the Admin UI, `_rust_session_validated`, and the internal MCP/A2A sidecar endpoints. Some internal MCP RPC code is now used by Python session affinity, so that part should be generalized rather than deleted wholesale.
- A2A CRUD still publishes Redis invalidations solely for the removed Rust L1 cache.
- Live docs and project guidance still describe active Rust modes/workspaces; historical ADRs can remain if clearly marked superseded.

There is also a scope question: Granian, prebuilt Rust MCP service images, and Rust-backed third-party Python packages remain. If the intent is to remove repository-owned Rust, I suggest describing the goal as "remove the first-party Rust workspace, runtime, sidecars, and related build/test infrastructure." If the intent is literally no Rust-backed binaries at runtime, the PR needs a broader dependency cleanup.

lucarlig:
@jonpspri, thanks for the updates. I reviewed head `ed6877c` with the clarified scope: remove repository-owned Rust; documentation is deferred, and the dataplane publisher is intentionally retained.

The previous build/release blockers are fixed. Remaining items:

- Two unit tests currently fail: `test_runtime_state.py::test_override_edge_cannot_bypass_session_auth_reuse_invariant` still reads the removed `rust_build_included` field, and `test_handle_internal_mcp_session_delete_skips_python_validation_when_rust_validated` still sends `_rust_session_validated` after the implementation renamed it.
- `main.py` still registers 41 internal MCP/A2A routes. Only `/_internal/mcp/rpc` has a production Python caller; the other 40 appear to be Rust-sidecar infrastructure, along with the A2A dispatch helpers and trust handling.
- The runtime-state coordinator still starts a Redis pub/sub subscriber on every gateway startup even though all runtime overrides now resolve to `NO_DISPATCHER`.
- `scripts/compliance_matrix.py` still GETs/PATCHes the deleted `/admin/runtime/mcp-mode` endpoint. It can exit successfully after running only the reference server and skipping the Python gateway.
- `tests/performance/utils/generate_docker_compose.py` still emits builds for deleted `mcp-servers/rust/*` paths, and the MCP isolation load test still requires `rust-managed`.
- `pr_5737_review_response_plan.md` looks like a temporary review artifact and should not ship.

The PR is still Draft, so the substantive CI jobs were skipped; marking it ready after these fixes should run the real gates.

lucarlig:
@jonpspri, thanks for the latest cleanup. I reviewed current head `627fa75e` with the agreed scope: remove repository-owned Rust, retain the dataplane publisher, and allow historical documentation and third-party Rust-backed dependencies to remain.

The core removal is now in place: I found no tracked `.rs` files, Cargo manifests/lockfile, Rust toolchain configuration, workspace crates, or cargo-vet supply-chain files. The dataplane publisher and its tests are preserved.

A few items still need attention before merge.

**Blocking**

- `MANIFEST.in` was over-pruned. It now omits required PNG/SVG/ICO assets, scripts, and project files, so `check-manifest` fails in `build-package`. Please restore the previous non-Rust inclusion rules and remove only Rust-specific entries. [Failure](https://github.com/IBM/mcp-context-forge/actions/runs/29911621314/job/88906559267)
- The Admin UI still renders the Stdio export button, but `generateConfig()` no longer supports `stdio`. Clicking it now raises `Unknown config type: stdio`, and three Vitest cases fail. Please remove the button, stdio modal text, and obsolete tests. [Failure](https://github.com/IBM/mcp-context-forge/actions/runs/29911621425/job/88905523824)
- `read_resource()` now returns `meta: null` because the previous unset-meta cleanup was removed. This is unrelated API drift and currently fails the URL-reputation E2E. Please restore the prior response behavior. [Failure](https://github.com/IBM/mcp-context-forge/actions/runs/29911621887/job/88904929415)
- `main.py` still registers 41 internal MCP/A2A handlers. Only `/_internal/mcp/rpc` has a production Python caller for session affinity; the other 40 appear to remain solely for the removed Rust sidecars. Please keep/generalize the RPC route and remove the remaining sidecar endpoints and their dedicated helpers, services, and tests.
- `scripts/compliance_matrix.py` still GETs and PATCHes the deleted `/admin/runtime/mcp-mode` endpoint. A 404 skips the Python slice while the reference slice runs, allowing a false-green exit. The runtime-mode fixtures/tests contain the same removed-endpoint assumptions.

**Remaining cleanup**

- Gateway startup still starts `RuntimeStateCoordinator`, which opens a Redis pub/sub subscription even though the runtime admin router and all production writers are gone.
- `tests/performance/utils/generate_docker_compose.py` still emits builds for deleted `mcp-servers/rust/*` paths, and the MCP isolation Locust test still requires `rust-managed`.
- `pr_5737_review_response_plan.md` is a temporary review artifact and currently claims the package issue is fixed even though CI is failing.
- `package-lock.json` was renamed to `jps-remove-experimental-rust-mcpgateway`.
- Three general observability span tests for `list_tools`, `list_server_tools`, and `list_server_mcp_tool_definitions` were deleted with the Rust-specific test block and should be restored.

The Python unit suite, license check, pre-commit, and container builds are green. Please ping me after the remaining cleanup and I will take another pass.

lucarlig (changes requested):
Thanks for the focused cleanup and for continuing to reduce the Rust-specific surface area. I found two blocking regressions and one additional correctness regression in Python behavior that survives this PR. I’m intentionally excluding the current CI failures from this review.

**Blocking — preserve authenticated trust for the surviving affinity path** (`mcpgateway/auth_context.py:374`)

Removing the Rust runtime should not remove authentication from the surviving affinity path. The base required both a recognized runtime marker and a shared-secret-derived header; the new trust gate accepts any non-empty, unsigned base64 auth context when `request.client.host` appears loopback. Because `ProxyHeadersMiddleware(trusted_hosts="*")` lets direct callers influence that host through forwarding headers, a caller can forge the forwarded identity for the retained `/_internal/mcp/*` and `/_internal/a2a/*` endpoints. Could we retain or rename the secret validation for `affinity` (or make these routes unaddressable externally) and restore the deny-path coverage?

**Blocking — continue forwarding the caller JWT for token-exchange resource reads** (`mcpgateway/main.py:6363`)

This public resource-read path still supports token-exchange gateways. `ResourceService.invoke_resource()` obtains the RFC 8693 subject token from `request_headers`, so removing `request_headers=dict(request.headers)` makes authenticated resource reads fail with “Token exchange requires an authenticated user JWT.” Please restore the argument; this behavior is unrelated to Rust removal.

**Correctness — preserve the intended authorization response** (`mcpgateway/main.py:7328`)

Please retain the `PermissionError` → 403 mapping here. `GatewayService` deliberately raises `PermissionError` when a non-platform-admin attempts to configure a token-exchange gateway; without that branch the endpoint returns 500 “Unexpected error.” This behavior is unrelated to Rust removal.

Requesting changes so the Rust removal does not weaken or alter these surviving Python paths.

pr_5737_review_response_plan.md
ja8zyjits, line 1: remove this file please

ja8zyjits (changes requested):

Lang-Akshay (changes requested):
Thanks for the PR @jonpspri 
Please address following blocking findings 
## Blocking Changes

| 1 | Security | `mcpgateway/auth_context.py` | 346–377 | HMAC trust-gate removed; trust gate now only checks loopback IP + non-empty `x-contextforge-auth-context` header. `ProxyHeadersMiddleware(trusted_hosts="*")` (main.py:3248) rewrites `request.client.host` from `X-Forwarded-For`. Attacker sends `X-Forwarded-For: 127.0.0.1` + any base64-JSON auth-context → trust gate passes → `_build_internal_mcp_forwarded_user` decodes attacker-controlled JSON as a synthetic admin identity → all 41 internal `/_internal/mcp/*` and `/_internal/a2a/*` routes execute under forged credentials. Layer-1 token scoping also skips entirely via `TokenScopingMiddleware`. Severity Critical; CWE-287/290/284. | Re-introduce cryptographic authentication of the auth-context header (e.g. HMAC-SHA256 over the JSON payload using `AUTH_ENCRYPTION_SECRET`). `encode_internal_mcp_auth_context` should sign; `decode_internal_mcp_auth_context` should verify before returning. The trust gate must reject unsigned/invalid signatures regardless of client IP. Alternatively, lock down `ProxyHeadersMiddleware` to a specific trusted-proxy IP instead of `"*"`. |
| 2 | Quality | `mcpgateway/main.py` | 7797–10015, 10494 | 40 of 41 `/_internal/mcp/*` and `/_internal/a2a/*` routes remain registered in `utility_router` with docstrings saying "forwarded from the local Rust runtime". Per @lucarlig review (July 22): only `/_internal/mcp/rpc` has a production Python caller (session affinity); the other 40 are Rust-sidecar infrastructure with no remaining callers. Leaving dead privileged endpoints increases attack surface. | Keep `/_internal/mcp/rpc` (and `/_internal/mcp/session` for affinity session teardown if needed); remove or explicitly gate the remaining 39 Rust-only routes. Update session-affinity caller to only use the surviving routes. |
| 3 | Quality | `scripts/compliance_matrix.py` | 173–190 | `_probe_state()` calls `GET /admin/runtime/mcp-mode` and `_flip_mode()` calls `PATCH /admin/runtime/mcp-mode` — both deleted with `runtime_admin_router.py`. The 404 response causes `_probe_state()` to return `None`, which causes `_determine_runnable()` to mark the "python" engine as "gateway unreachable". Only the "reference" (in-process FastMCP) slice runs; the Python gateway compliance slice is silently skipped. This is a false-green exit for the integration gate. | Remove the runtime-mode state probe and flip calls from the compliance matrix. The "python" engine should run unconditionally (it is always available). Remove `_ENGINE_FLIP_TARGET`, `_probe_state`, `_flip_mode`, `_wait_data_plane_converges`, and the `--no-flip` / `--no-restore` arguments, or redirect them to a no-op. `_determine_runnable` for the "python" engine should return `(engine, None)` (always runnable). |

msureshkumar88 (changes requested):
Thanks for pushing this through — the core Rust workspace removal is clean and the incremental fixes across review rounds addressed most of the earlier build/license/MANIFEST breakage well.

I went through current head (`db3876a6a`) directly against the flagged file:line locations from the last review round, and three blocking items from that round are still present in the diff (not yet addressed by the follow-up commits, which touched MANIFEST.in/admin-UI/e2e only):

**Blocking**

- `mcpgateway/main.py:6363` (`read_resource` by ID) — this call to `resource_service.read_resource(...)` is missing `request_headers=dict(request.headers)`, unlike the sibling call sites at lines 7550/8427/10925 that do pass it. `ResourceService.invoke_resource()` reads the RFC 8693 subject token from `request_headers` for token-exchange gateways, so any resource behind a token-exchange gateway will fail this read path with "Token exchange requires an authenticated user JWT." Restoring the argument should be enough.
- `mcpgateway/main.py:7260-7344` (`register_gateway`) — the `except Exception` block here has branches for `GatewayConnectionError`, `ValueError`, `GatewayNameConflictError`, etc., but no `PermissionError` branch (sibling handlers, e.g. around line 7427, do have one). `GatewayService` intentionally raises `PermissionError` when a non-admin tries to configure a token-exchange gateway, so that 403 currently surfaces as a 500 "Unexpected error" instead.
- `mcpgateway/auth_context.py:346-378` (`is_trusted_internal_runtime_request`) — the trust gate for `/_internal/mcp/*` and `/_internal/a2a/*` now checks path prefix, a non-empty (but unsigned/unverified) `x-contextforge-auth-context` header, and `request.client.host` being loopback. Since `ProxyHeadersMiddleware(trusted_hosts="*")` is registered (`main.py:3248`), an external caller can set `X-Forwarded-For: 127.0.0.1` to satisfy the loopback check, and the header itself is only base64+JSON decoded — no HMAC check — even though this file already has HMAC helpers (`sign_redis_forward_envelope`/`verify_redis_forward_envelope`) for the Redis forward-envelope elsewhere. Could this gate either verify a signature the same way, or enforce loopback independent of the proxy-headers trust model?

**Non-blocking, would be good to clean up before merge**

- `pr_5737_review_response_plan.md` is still tracked at repo root — looks like a temporary review-response scratch doc that shouldn't ship.
- `scripts/compliance_matrix.py` still GETs/PATCHes the deleted `/admin/runtime/mcp-mode` endpoint; a 404 there currently lets that slice exit green rather than fail, which could mask a real regression in CI.
- `RuntimeStateCoordinator` (`main.py:1696`) still starts unconditionally on gateway boot (Redis pub/sub subscribe) even though, per the removal, all runtime overrides should now resolve to `NO_DISPATCHER` — minor resource cost for a coordinator with nothing left to coordinate.
- A few docs pages still reference the removed `wrapper` module by name (e.g. `docs/docs/index.md`, `using/clients/index.md`, `overview/features.md`) — worth a pass to confirm which are stale vs. intentionally about the separate `mcpgateway.translate --stdio` bridge.
- No CHANGELOG entry calls this out as a breaking change for anyone currently using the stdio wrapper or Rust runtime modes — a short "Breaking Changes" note would help users find it later.

Happy to take another pass once these land — the bulk of the removal itself looks solid.

brian-hussey (changes requested):
@jonpspri these are all things we want, but I'd rather see them as independent singly focused PRs to achieve the goals we're looking for.

As it is, this PR is huge making it difficult to review/accept with any meaningful confidence. 
From quick review this covers:
1. removing rust
1. removing stdio wrapper
1. Unrelated cleanup/refactoring

Each deserves their own PR to keep us in good shape.

brian-hussey:
@jonpspri @ja8zyjits I'm asking for big changes here that each deserve their own pull request, but I'm going to close this PR as rejected and not try to reuse it for part of what's required.

We want more focused PRs that try to do 1 thing and 1 thing only, and we should have issues associated with any changes like this to see where the work came from and the value it brings.

cafalchio (commented):
Cleaning is always good

Could this PR be splitted into few PRs.

For instance, one that just change python files that depends on rust / wrapper code. Another for all .rs and cargo. 
After that all the non code related. 

It will make it much easier to be reviewed and approved.

PR #5745, opened 2026-07-21
fix(retry): guard request() Retry-After against RFC 9110 HTTP-date

### LINKED ISSUE #5674: [BUG]: ResilientHttpClient.request() raises ValueError on RFC 9110 HTTP-date Retry-After (stream() already guards it)
### 🐞 Bug Summary

On an HTTP 429 response, `ResilientHttpClient.request()` reads the `Retry-After` header with `float(retry_after)` (`mcpgateway/utils/retry_manager.py:438`). RFC 9110 §10.2.3 defines `Retry-After` as either `delay-seconds` **or** an `HTTP-date`, and both are legal. When an upstream sends the date form (`Retry-After: Wed, 21 Oct 2026 07:28:00 GMT`), `float()` raises `ValueError` inside the `try` block; `_should_retry(ValueError, None)` matches neither the retryable-exception tuple nor a response, so the bare `raise` propagates the `ValueError` out of `request()`. A retryable rate-limit becomes an unhandled crash.

The sibling `stream()` method already handles this exact case: `:585` wraps `float(ra)` in `except ValueError: wait = None` and falls through to normal backoff. The two `Retry-After` consumers in one class disagree, which is what makes this an oversight rather than a design choice.

### 🧩 Affected Component

- [x] Other (explain below)

The shared `ResilientHttpClient` retry utility (`mcpgateway/utils/retry_manager.py`), used by the outbound REST/tool proxy and gateway-connectivity paths.

### 🔁 Steps to Reproduce

Executed in a clean `python:3.12-slim` container, `pip install -e .`, at `main@5eece0fc`. No mocking of the request path: a real local `http.server` returns 429 with an HTTP-date `Retry-After`.

```python
import asyncio, threading, http.server, socketserver
from mcpgateway.utils.retry_manager import ResilientHttpClient

class H(http.server.BaseHTTPRequestHandler):
    def do_GET(self):
        self.send_response(429)
        self.send_header("Retry-After", "Wed, 21 Oct 2026 07:28:00 GMT")  # RFC 9110 HTTP-date
        self.end_headers(); self.wfile.write(b"rate limited")
    def log_message(self, *a): pass

srv = socketserver.TCPServer(("127.0.0.1", 0), H)
port = srv.server_address[1]
threading.Thread(target=srv.serve_forever, daemon=True).start()

async def main():
    c = ResilientHttpClient(max_retries=2, base_backoff=0.01, max_delay=60.0)
    try:
        await c.request("GET", f"http://127.0.0.1:{port}/")
        print("returned normally")
    except ValueError as e:
        print(f"RAISED ValueError: {e}")
    await c.aclose()

asyncio.run(main())
```

### 🤔 Expected Behavior

`request()` should accept an HTTP-date `Retry-After` the way RFC 9110 permits and the way `stream()` already does: parse it (or, if unparseable, fall through to the exponential backoff that already handles 429 correctly), never propagate a `ValueError`.

### 📓 Logs / Error Output

```
PROVENANCE retry_manager __file__ = /src/mcpgateway/utils/retry_manager.py
HAS block-A unguarded float(retry_after): True
HAS block-B guarded try/except ValueError in stream(): True

== request() path (block A) ==
RAISED ValueError: could not convert string to float: 'Wed, 21 Oct 2026 07:28:00 GMT'

== stream() path (block B) ==
yielded/retried without ValueError (graceful)
```

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Version or commit | `main@5eece0fc` |
| Runtime | Python 3.12 |
| Platform / OS | Debian slim (Docker) |
| Container | Docker |

### 🧩 Additional Context

Two secondary code observations in the same 429 handler, offered for scope only (I observed the `ValueError` above; these two are read from the source, not separately exercised):

- Neither 429 sleep is clamped to `self.max_delay`. `request()` (`:440`) and `stream()` (`:590`) pass the header value straight to `asyncio.sleep`, whereas `_sleep_with_jitter` (`:316`) already does `delay = min(delay, self.max_delay)`. A large `Retry-After` therefore ignores the configured cap.
- A past-date `HTTP-date` yields a negative delay, so a fix that parses the date should clamp the result to `0`.

Happy to open a PR (parse the date via `email.utils.parsedate_to_datetime`, reuse the existing `min(..., self.max_delay)` clamp) once the issue is de-triaged, per CONTRIBUTING's "wait until maintainers accept or scope the issue" rule.

This report was prepared with AI assistance; the reproduction above was run and verified by me.

ebarkhordar:
## 📌 Summary

On a 429 response, `ResilientHttpClient.request()` reads `Retry-After` with a bare `float()` (`mcpgateway/utils/retry_manager.py:438`). RFC 9110 §10.2.3 allows `Retry-After` to be either `delay-seconds` **or** an `HTTP-date`, and both are valid. When an upstream sends the date form, `float()` raises `ValueError`, `_should_retry(ValueError, None)` matches neither a retryable exception nor a response, and the bare `raise` propagates the error out of `request()`. A retryable rate-limit becomes an unhandled crash.

## 🔁 Reproduction Steps

Closes #5674. Verified in a clean `python:3.12-slim` container (`pip install -e .`), no mocking of the parse path:

```python
mock_resp = httpx.Response(429, headers={"Retry-After": "Wed, 21 Oct 2026 07:28:00 GMT"})
await client.get("http://x")   # ValueError: could not convert string to float: 'Wed, 21 Oct 2026 07:28:00 GMT'
```

## 🐞 Root Cause

`request()` and the sibling `stream()` are the two `Retry-After` consumers in this class, and they disagree. `stream()` (`:585`) already wraps `float(ra)` in `except ValueError: wait = None` and falls through to normal backoff; `request()` never got the same guard. That asymmetry is what makes this an oversight rather than a deliberate design choice.

## 💡 Fix Description

Mirror `stream()` in `request()`: wrap the `float()` in `try/except ValueError`, and on a non-numeric value fall through to the existing exponential-backoff path instead of raising. Parsing the HTTP-date into a precise delay is intentionally left out of scope so the two paths behave identically.

## 📏 Reviewability

## 🧪 Verification

All run in a clean `python:3.12-slim` container at `main`, `pip install -e .`:

| Check | Command | Status |
|-------|---------|--------|
| Regression (fails before fix, `ValueError` at `retry_manager.py:438`) | `pytest ...::test_request_retry_after_http_date_does_not_crash` | fails on main, passes on branch |
| Full module suite | `pytest tests/unit/mcpgateway/utils/test_retry_manager.py` | green |
| Coverage of `retry_manager.py` | `pytest --cov=mcpgateway.utils.retry_manager` | 100% |
| Lint | `ruff check` | clean |
| Hooks | `pre-commit run --files <changed>` | ruff, format, headers, interrogate, detect-secrets pass |

The new test asserts the final state: a 429 with an HTTP-date `Retry-After` no longer raises, falls through to backoff, and returns the 429 after `max_retries` attempts. It patches the transport (returning the 429), matching the existing `test_retry_after_header_respected` convention in this file; it does not exercise a live socket.

## ✅ Checklist

- [x] Code formatted (`ruff format`, pre-commit)

### DIFF
--- a/mcpgateway/utils/retry_manager.py
+++ b/mcpgateway/utils/retry_manager.py
@@ -435,11 +435,17 @@ class ResilientHttpClient:
                 if response.status_code == 429:
                     retry_after = response.headers.get("Retry-After")
                     if retry_after:
-                        retry_after_sec = float(retry_after)
-                        logger.info(f"Rate-limited. Retrying after {retry_after_sec}s.")
-                        await asyncio.sleep(retry_after_sec)
-                        attempt += 1
-                        continue
+                        try:
+                            retry_after_sec = float(retry_after)
+                        except ValueError:
+                            # RFC 9110 §10.2.3 also allows an HTTP-date, which is not a float.
+                            # Mirror stream(): fall through to normal exponential backoff.
+                            retry_after_sec = None
+                        if retry_after_sec:
+                            logger.info(f"Rate-limited. Retrying after {retry_after_sec}s.")
+                            await asyncio.sleep(retry_after_sec)
+                            attempt += 1
+                            continue
 
                 if not self._should_retry(Exception(), response):
                     return response
--- a/mcpgateway/utils/retry_manager.py
+++ b/mcpgateway/utils/retry_manager.py
@@ -441,7 +441,7 @@ class ResilientHttpClient:
                             # RFC 9110 §10.2.3 also allows an HTTP-date, which is not a float.
                             # Mirror stream(): fall through to normal exponential backoff.
                             retry_after_sec = None
-                        if retry_after_sec:
+                        if retry_after_sec is not None:
                             logger.info(f"Rate-limited. Retrying after {retry_after_sec}s.")
                             await asyncio.sleep(retry_after_sec)
                             attempt += 1
--- a/tests/unit/mcpgateway/utils/test_retry_manager.py
+++ b/tests/unit/mcpgateway/utils/test_retry_manager.py
@@ -77,6 +77,24 @@ async def test_retry_after_header_respected(client):
 
 
 @pytest.mark.asyncio
+async def test_request_retry_after_http_date_does_not_crash(client):
+    """RFC 9110 §10.2.3 allows Retry-After as an HTTP-date, not only delay-seconds.
+
+    request() must not raise ValueError on the date form. The sibling stream()
+    path already guards this (test_stream_429_retry_after_invalid_value); request()
+    should behave the same, falling through to normal backoff and returning the 429.
+    """
+    mock_resp = httpx.Response(429, headers={"Retry-After": "Wed, 21 Oct 2026 07:28:00 GMT"})
+
+    with patch.object(client.client, "request", new=AsyncMock(return_value=mock_resp)) as mock_req:
+        with patch("asyncio.sleep", new=AsyncMock()):
+            resp = await client.get("http://retry-after-http-date.com")
+
+    assert resp.status_code == 429
+    assert mock_req.call_count == client.max_retries
+
+
+@pytest.mark.asyncio
 async def test_max_retry_reached_raises_exception(client):
     failing_func = AsyncMock(side_effect=httpx.ConnectTimeout("Connection failed"))
 
--- a/tests/unit/mcpgateway/utils/test_retry_manager.py
+++ b/tests/unit/mcpgateway/utils/test_retry_manager.py
@@ -95,6 +95,25 @@ async def test_request_retry_after_http_date_does_not_crash(client):
 
 
 @pytest.mark.asyncio
+async def test_request_retry_after_zero_retries_immediately(client):
+    """Retry-After: 0 is a valid delay-seconds value meaning retry immediately.
+
+    Guards the truthiness check: `if retry_after_sec:` would treat 0.0 as falsy and
+    skip the retry path, so the header must be tested with `is not None`. The client
+    should sleep(0) and keep retrying up to max_retries.
+    """
+    mock_resp = httpx.Response(429, headers={"Retry-After": "0"})
+
+    with patch.object(client.client, "request", new=AsyncMock(return_value=mock_resp)) as mock_req:
+        with patch("asyncio.sleep", new=AsyncMock()) as mock_sleep:
+            resp = await client.get("http://retry-after-zero.com")
+
+    assert resp.status_code == 429
+    assert mock_sleep.call_args_list[0][0][0] == 0.0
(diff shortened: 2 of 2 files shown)


mcpgateway/utils/retry_manager.py
```diff
@@ -435,11 +435,17 @@ async def request(self, method: str, url: str, **kwargs) -> httpx.Response:
...
-                        await asyncio.sleep(retry_after_sec)
-                        attempt += 1
-                        continue
+                        try:
+                            retry_after_sec = float(retry_after)
+                        except ValueError:
+                            # RFC 9110 §10.2.3 also allows an HTTP-date, which is not a float.
+                            # Mirror stream(): fall through to normal exponential backoff.
+                            retry_after_sec = None
+                        if retry_after_sec:
```
Sanjays2402: `Retry-After: 0` is a valid delay-seconds value, but this truthiness check sends it through exponential backoff. Before this change it called `sleep(0)` and retried immediately; should this test `is not None` instead?

ebarkhordar: Good catch, you are right. float("0") is 0.0, so the truthiness guard skipped the retry and changed the prior sleep(0)-and-retry behavior. Fixed to use is not None in 9868996, with a regression test asserting Retry-After: 0 still sleeps(0) and retries.

ebarkhordar:
No rush on this. Since it was opened I addressed the `Retry-After: 0` regression Sanjays2402 caught (the guard now tests `is not None`, with a test pinning that a zero delay still sleeps and retries), and the branch is still mergeable against main.

One thing to flag: none of the repo CI has actually run on this PR. All 13 workflow runs on this head are held awaiting approval, so DCO is the only green check and everything in the verification table came from a local container.

ebarkhordar:
Closing this to keep my open-PR queue honest rather than because anything is wrong with it. It has been open a month without drawing a maintainer's attention, and an unreviewed PR sitting indefinitely is not much use to either of us.

For anyone who finds this later: the fix is still valid as far as I can tell. `ResilientHttpClient.request()` raises `ValueError` on an RFC 7231 HTTP-date `Retry-After` (#5674), because it goes straight to `float()` on a value the spec allows to be either a delay in seconds or a date. The branch also carries the `Retry-After: 0` correction Sanjays2402 pointed out in review. One caveat on the evidence: this repo's CI never ran the branch, since all 13 workflows stayed at `action_required` awaiting fork approval for the whole month, so everything behind it was verified locally.

The branch stays up, so this can be reopened as is if it becomes useful. Happy to rebase it if anyone wants to pick it up.

PR #5751, opened 2026-07-22
ci: enable all arch build for workflow_dispatch manual trigger

madhu-mohan-jaishankar:
## 🔗 Related Issue

Closes #

---

## 📝 Summary

The multiplatform Docker build workflow currently builds:
- `amd64` on PR and merge queue (cache-only), and on push to main (build + push)
- `arm64` on merge queue (cache-only), and on push to main (build + push)
- `ppc64le` on push to main (build + push) — **was running on every push despite runner capacity constraints**
- `s390x` — already disabled entirely due to runner capacity constraints

This PR:
1. Restricts both `s390x` and `ppc64le` to `workflow_dispatch` only — they no longer run on push to main
2. Adds a manual `workflow_dispatch` trigger path that builds **all four architectures** at once, producing a complete multiplatform manifest with Cosign signing and SBOM attestation

**Changes to docker-multiplatform.yml:**

1. **`wheels` job** — Adds s390x back into the matrix (was commented out) and adds `build_on_push: false` to both s390x and ppc64le entries. Adds a `runs-on` expression that routes these arches to a cheap `ubuntu-24.04` runner on push, and a `gate` step that skips all subsequent steps — no native runner slots are consumed on normal pushes.

2. **`build` job** — Adds `build_on_push: false` to both s390x and ppc64le matrix entries. The existing `runs-on` expression and gate step already handle this flag identically for both arches — no additional logic needed.

3. **`manifest` job** — The `IMAGES` bash array now only contains `amd64` and `arm64` on push. Both `s390x` and `ppc64le` digests are appended conditionally when the event is `workflow_dispatch`, preventing the manifest job from referencing non-existent digests on normal pushes.

**Resulting behaviour:**

| Event | amd64 | arm64 | s390x | ppc64le | Manifest |
|---|---|---|---|---|---|
| PR | cache-only | skip | skip | skip | no |
| Merge queue | cache-only | cache-only | skip | skip | no |
| Push to main | build+push | build+push | skip | **skip** | **2-arch** |
| `workflow_dispatch` from main | build+push | build+push | build+push | build+push | **4-arch** |

---

## 📏 Reviewability

---

## 🏷️ Type of Change

- [ ] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

This is a CI-only change with no application code modified. Functional verification requires triggering the workflow manually from the GitHub Actions UI.

| Check | Command / Action | Status |
|---|---|---|
| Workflow syntax | `actionlint .github/workflows/docker-multiplatform.yml` | pending |
| Push-to-main behaviour unchanged | Observe CI on merge — s390x and ppc64le must be skipped | pending |
| `workflow_dispatch` all-arch build | Actions → "Multiplatform Docker Build" → Run workflow → select `main` | pending |
| 4-arch manifest produced | `docker buildx imagetools inspect ghcr.io/ibm/mcp-context-forge:<sha>` shows amd64, arm64, s390x, ppc64le | pending |

---

## ✅ Checklist

---

### DIFF
--- a/.github/workflows/docker-multiplatform.yml
+++ b/.github/workflows/docker-multiplatform.yml
@@ -5,18 +5,20 @@
 # This workflow builds container images for multiple architectures:
 #   - linux/amd64  (native on ubuntu-24.04)        - PR (build only), merge queue (build only), push
 #   - linux/arm64  (native on ubuntu-24.04-arm)     - merge queue (build only), push
-#   - linux/s390x  (native on ubuntu-24.04-s390x)   - push and workflow_dispatch only (NOT merge queue)
-#   - linux/ppc64le (native on ubuntu-24.04-ppc64le) - push and workflow_dispatch only (NOT merge queue)
+#   - linux/s390x  (native on ubuntu-24.04-s390x)   - workflow_dispatch only (runner capacity)
+#   - linux/ppc64le (native on ubuntu-24.04-ppc64le) - workflow_dispatch only (runner capacity)
 #
-# s390x and ppc64le are excluded from the merge queue to avoid stalling it while
-# waiting for scarce native runner pools. A per-entry gate step skips their build
-# steps when the event is merge_group. They still build and push on every push to
-# main and on workflow_dispatch, so the multiplatform manifest is always complete.
+# s390x and ppc64le are excluded from the merge queue and from normal push-to-main
+# due to runner capacity constraints. A per-entry gate step skips their build steps
+# on push events. They build and push only on workflow_dispatch (manual trigger).
+#
+# Push-to-main and release tag builds produce a 2-arch manifest (amd64 + arm64).
+# Trigger workflow_dispatch from main or a tag ref to produce the full 4-arch manifest.
 #
 # Pipeline:
-#   1. Build platform images in parallel (amd64+arm64 on merge queue; all arches on push)
-#   2. Create multiplatform manifest (push and tags only)
-#   3. Sign and attest with Cosign (keyless OIDC, push and tags only)
+#   1. Build platform images in parallel (amd64+arm64 on push; all arches on workflow_dispatch)
+#   2. Create multiplatform manifest (push and workflow_dispatch only)
+#   3. Sign and attest with Cosign (keyless OIDC, push and workflow_dispatch only)
 #
 # Linting and security scanning are handled by docker-scan.yml
 #
@@ -172,6 +174,7 @@ jobs:
       - name: Evaluate wheels gate
         id: gate
         run: |
+          set -euo pipefail
           build="true"
           if [[ "${{ github.event_name }}" == "push" && "${{ matrix.build_on_push }}" == "false" ]]; then
             echo "Skipping wheels for ${{ matrix.suffix }}: build_on_push=false"
@@ -266,6 +269,7 @@ jobs:
             qemu: false
             build_on_pr: true
             build_on_merge_group: true
+            build_on_push: true
           - platform: linux/arm64
             runner: ubuntu-24.04-arm
             suffix: arm64
@@ -273,6 +277,7 @@ jobs:
             qemu: false
             build_on_pr: false
             build_on_merge_group: true
+            build_on_push: true
           - platform: linux/s390x
             runner: ubuntu-24.04-s390x
             suffix: s390x
@@ -280,7 +285,7 @@ jobs:
             qemu: false
             build_on_pr: false
             build_on_merge_group: false   # not merge queue
-            build_on_push: false          # temporarily disabled — s390x runner capacity issues
+            build_on_push: false          # runner capacity — workflow_dispatch only
           - platform: linux/ppc64le
             runner: ubuntu-24.04-ppc64le
             suffix: ppc64le
--- a/.github/workflows/docker-multiplatform.yml
+++ b/.github/workflows/docker-multiplatform.yml
@@ -149,27 +149,43 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          # s390x temporarily disabled on push due to runner capacity issues.
-          # Re-enable by uncommenting the entry below once runners are stable.
-          # - platform: linux/s390x
-          #   runner: ubuntu-24.04-s390x
-          #   suffix: s390x
+          # s390x disabled on push due to runner capacity issues, but enabled
+          # for workflow_dispatch so the manual "build all arches" trigger works.
+          - platform: linux/s390x
+            runner: ubuntu-24.04-s390x
+            suffix: s390x
+            build_on_push: false   # runner capacity — skip on normal push
           - platform: linux/ppc64le
             runner: ubuntu-24.04-ppc64le
             suffix: ppc64le
-    runs-on: ${{ matrix.runner }}
+    # Use a cheap ubuntu-24.04 runner when build_on_push=false and the event is
+    # push — avoids acquiring scarce s390x runner slots on every push to main.
+    # workflow_dispatch always uses the native runner to build all arches.
+    runs-on: ${{ github.event_name == 'push' && matrix.build_on_push == false && 'ubuntu-24.04' || matrix.runner }}
     timeout-minutes: 120
     permissions:
       contents: read
       packages: write
 
     steps:
+      - name: Evaluate wheels gate
+        id: gate
+        run: |
+          build="true"
+          if [[ "${{ github.event_name }}" == "push" && "${{ matrix.build_on_push }}" == "false" ]]; then
+            echo "Skipping wheels for ${{ matrix.suffix }}: build_on_push=false"
+            build="false"
+          fi
+          echo "build=${build}" >> "${GITHUB_OUTPUT}"
+
       - name: Checkout code
+        if: steps.gate.outputs.build == 'true'
         uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5
         with:
           persist-credentials: false
 
       - name: Compute image ref
+        if: steps.gate.outputs.build == 'true'
         id: ref
         run: |
           WHEELS_IMAGE_LC=$(echo "${{ env.IMAGE_NAME }}-wheels" | tr '[:upper:]' '[:lower:]')
@@ -178,11 +194,13 @@ jobs:
           echo "latest=${{ env.REGISTRY }}/${WHEELS_IMAGE_LC}:${{ matrix.suffix }}-latest" >> "${GITHUB_OUTPUT}"
 
       - name: Set up Docker Buildx
+        if: steps.gate.outputs.build == 'true'
         uses: docker/setup-buildx-action@8d2750c68a42422c14e847fe6c8ac0403b4cbd6f # v3
         with:
           version: latest
 
       - name: Log in to GHCR
+        if: steps.gate.outputs.build == 'true'
         uses: docker/login-action@c94ce9fb468520275223c153574b00df6fe4bcc9 # v3
         with:
           registry: ${{ env.REGISTRY }}
@@ -190,6 +208,7 @@ jobs:
           password: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Check for existing wheel image
+        if: steps.gate.outputs.build == 'true'
         id: check
         run: |
           if docker manifest inspect "${{ steps.ref.outputs.image }}" >/dev/null 2>&1; then
@@ -203,7 +222,7 @@ jobs:
       # `latest` tag per arch (fallback for local builds and for `build`
       # below when the lock-hash image isn't published yet).
       - name: Build and push wheel image
-        if: steps.check.outputs.exists != 'true'
+        if: steps.gate.outputs.build == 'true' && steps.check.outputs.exists != 'true'
         uses: docker/build-push-action@10e90e3645eae34f1e60eeb005ba3a3d33f178e8 # v6
         with:
           context: .
@@ -462,13 +481,17 @@ jobs:
 
           echo "Creating multiplatform manifest..."
 
-          # s390x temporarily excluded from the manifest due to runner capacity issues.
-          # Re-add "${IMAGE}:s390x-${SHA}" once s390x builds are re-enabled.
+          # s390x is included in the manifest only when triggered via workflow_dispatch
+          # (manual "build all arches" trigger). On normal push it is excluded because
+          # the s390x build is skipped there due to runner capacity constraints.
           IMAGES=(
             "${IMAGE}:amd64-${SHA}"
             "${IMAGE}:arm64-${SHA}"
             "${IMAGE}:ppc64le-${SHA}"
           )
+          if [[ "${{ github.event_name }}" == "workflow_dispatch" ]]; then
+            IMAGES+=("${IMAGE}:s390x-${SHA}")
+          fi
 
           if [[ "${GITHUB_REF_TYPE}" == "tag" ]]; then
             docker buildx imagetools create \
--- a/.github/workflows/docker-multiplatform.yml
+++ b/.github/workflows/docker-multiplatform.yml
@@ -158,6 +158,7 @@ jobs:
           - platform: linux/ppc64le
             runner: ubuntu-24.04-ppc64le
             suffix: ppc64le
+            build_on_push: false   # runner capacity — skip on normal push
     # Use a cheap ubuntu-24.04 runner when build_on_push=false and the event is
     # push — avoids acquiring scarce s390x runner slots on every push to main.
     # workflow_dispatch always uses the native runner to build all arches.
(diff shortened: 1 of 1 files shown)


ja8zyjits (dismissed):
LGTM

msureshkumar88 (changes requested):
Thanks for tackling this — the runner-capacity problem is real, and the gate-step pattern here mirrors the existing `build` job cleanly, so the shape of the change is right. Re-adding s390x to the `wheels` matrix (rather than leaving it commented out) is also a nice cleanup.

One blocking item and a few follow-ups.

---

## Blocking: release tags lose ppc64le and s390x

The workflow triggers on `push` for `tags: ["v*"]` (line 30), and for a tag push `github.event_name` is still `push`. So on the next release:

- the gate sees `build_on_push=false` and skips both s390x and ppc64le
- the manifest sees `event_name != 'workflow_dispatch'` and omits both digests

`:latest`, `:vX.Y.Z` and `:<sha>` become 2-arch (amd64 + arm64), and the `sign` job then signs and attests that reduced manifest as the release artifact. v1.0.6 did ship ppc64le (run `29852471833`), so this would be a silent packaging regression for consumers on that arch.

Runner cost on tags is bounded — releases are infrequent — so scoping the skip to branch pushes seems like the cheapest fix:

```yaml
# gate step (both wheels and build)
if [[ "${{ github.event_name }}" == "push" && "${{ github.ref_type }}" != "tag" && "${{ matrix.build_on_push }}" == "false" ]]; then
```

```bash
# manifest job
if [[ "${{ github.event_name }}" == "workflow_dispatch" || "${GITHUB_REF_TYPE}" == "tag" ]]; then
  IMAGES+=("${IMAGE}:s390x-${SHA}" "${IMAGE}:ppc64le-${SHA}")
fi
```

The `runs-on` expressions would want the same `ref_type` guard so tag builds land on native runners.

If dropping those arches from releases is actually intended, that's a reasonable call too — but worth stating explicitly in the PR description and the file header, since it changes what users pull.

---

## Suggestion: `matrix.build_on_push == false` is true when the key is absent

Not introduced by this PR, but this PR leans on the flag more heavily and copies the expression into the new `wheels` `runs-on`, so it may be worth fixing here.

GitHub expressions coerce `null → 0` and `false → 0`, so `matrix.build_on_push == false` evaluates **true** for the `amd64` and `arm64` entries, which don't set the key. Line 296 therefore routes arm64 to a generic `ubuntu-24.04` runner on every push. Job logs confirm it:

```
Build arm64   success   ubuntu-24.04      # expected ubuntu-24.04-arm
```

Same result across runs `29851649876`, `29743237284`, and `29852471833`. The `gate` steps are unaffected because they compare the rendered string (`"" != "false"`), which is why the arch still gets built — just not on arm hardware.

Simplest fix is to make the flag explicit for every entry, the way `build_on_pr` and `build_on_merge_group` already are:

```yaml
- platform: linux/amd64
  ...
  build_on_push: true
- platform: linux/arm64
  ...
  build_on_push: true
```

---

## Minor notes

- **Event logic now lives in three places** — the `wheels` gate, the `build` gate, and the `IMAGES` array in `manifest`. Flip a flag in one and the manifest silently disagrees. One option is to build `IMAGES` from the downloaded `digest-<suffix>` artifacts so it can't drift from what was actually built.
- **Header comments are now stale** — lines 8-9, 13-14 and 17 still say s390x/ppc64le build "on push and workflow_dispatch" and that "the multiplatform manifest is always complete". Line 304 says the same. The table in the PR description is accurate; moving it into the file header would help the next reader.
- **`wheels` gate step is missing `set -euo pipefail`**, which the `build` gate has (line 313).
- **Comment wording drifts** between the two jobs for the same flag: `# runner capacity — skip on normal push` in `wheels` vs `# temporarily disabled — s390x runner capacity issues` in `build` (line 283). Worth unifying.
- **Concurrency**: `${{ github.workflow }}-${{ github.ref }}` with `cancel-in-progress` means a manual dispatch on `main` cancels an in-flight push run on the same ref, and vice versa — so the 4-arch manifest a dispatch produces can be cancelled by the next merge. Probably fine, just something to be aware of when using the new trigger.

Nothing else stood out — action SHAs stay pinned, `persist-credentials: false` is preserved on the newly gated checkout, and the existing `push:` / `outputs:` conditions (lines 421-422) already handled `workflow_dispatch` correctly and were rightly left alone.

msureshkumar88:
Following up to clarify the blocking point, since the goal here is the manual all-arch trigger and I want to be clear that I'm not objecting to that part.

Adding the `workflow_dispatch` all-arch path is new capability and looks good. The concern is a separate, incidental one: the PR also adds `build_on_push: false` to ppc64le, which changes **existing** behaviour — and `push` includes tag pushes, since the workflow triggers on `tags: ["v*"]` (line 30).

Concrete sequence on the next release, with no one doing anything wrong:

1. `v1.0.7` is tagged and pushed → workflow fires with `github.event_name == 'push'`
2. Gate evaluates `build_on_push=false` → s390x and ppc64le skipped
3. `manifest` still runs (`if: push || workflow_dispatch`), and `GITHUB_REF_TYPE == "tag"` so it tags `:v1.0.7`, `:latest` and `:<sha>` — with amd64 + arm64 only
4. `sign` signs and attests that 2-arch manifest as the release artifact

So `:latest` silently becomes 2-arch. v1.0.6 shipped ppc64le (run `29852471833`), so anyone pulling on that arch is affected.

The new dispatch path does give a way out — running "Run workflow" against the `v1.0.7` tag gives `ref_type == 'tag'`, the gate passes for all four arches, and the manifest re-tags `:v1.0.7`/`:latest` with the full set. But that's manual remediation *after* a bad `:latest` has already been published and signed, and it isn't documented anywhere, so the release process quietly gains a required manual step.

Either of these resolves it from my side:

- **Guard the skip with `ref_type`** so tag pushes keep all four arches automatically:

  ```yaml
  if [[ "${{ github.event_name }}" == "push" && "${{ github.ref_type }}" != "tag" && "${{ matrix.build_on_push }}" == "false" ]]; then
  ```

  plus the matching `|| "${GITHUB_REF_TYPE}" == "tag"` in the manifest's `IMAGES` append, and the same `ref_type` guard in the `runs-on` expressions so tag builds land on native runners. Runner cost is bounded — releases are infrequent.

- **Or accept 2-arch on tag push as intentional**, and document it in the workflow header and the PR description, including how releases get their full manifest.

Happy either way — just not leaving it implicit, since the failure mode is silent. The other items in my review are non-blocking.

madhu-mohan-jaishankar:
> Thanks for tackling this — the runner-capacity problem is real, and the gate-step pattern here mirrors the existing `build` job cleanly, so the shape of the change is right. Re-adding s390x to the `wheels` matrix (rather than leaving it commented out) is also a nice cleanup.
> 
> One blocking item and a few follow-ups.
> 
> ## Blocking: release tags lose ppc64le and s390x
> The workflow triggers on `push` for `tags: ["v*"]` (line 30), and for a tag push `github.event_name` is still `push`. So on the next release:
> 
> * the gate sees `build_on_push=false` and skips both s390x and ppc64le
> * the manifest sees `event_name != 'workflow_dispatch'` and omits both digests
> 
> `:latest`, `:vX.Y.Z` and `:<sha>` become 2-arch (amd64 + arm64), and the `sign` job then signs and attests that reduced manifest as the release artifact. v1.0.6 did ship ppc64le (run `29852471833`), so this would be a silent packaging regression for consumers on that arch.
> 
> Runner cost on tags is bounded — releases are infrequent — so scoping the skip to branch pushes seems like the cheapest fix:
> 
> ```yaml
> # gate step (both wheels and build)
> if [[ "${{ github.event_name }}" == "push" && "${{ github.ref_type }}" != "tag" && "${{ matrix.build_on_push }}" == "false" ]]; then
> ```
> 
> ```shell
> # manifest job
> if [[ "${{ github.event_name }}" == "workflow_dispatch" || "${GITHUB_REF_TYPE}" == "tag" ]]; then
>   IMAGES+=("${IMAGE}:s390x-${SHA}" "${IMAGE}:ppc64le-${SHA}")
> fi
> ```
> 
> The `runs-on` expressions would want the same `ref_type` guard so tag builds land on native runners.
> 
> If dropping those arches from releases is actually intended, that's a reasonable call too — but worth stating explicitly in the PR description and the file header, since it changes what users pull.
> 
> ## Suggestion: `matrix.build_on_push == false` is true when the key is absent
> Not introduced by this PR, but this PR leans on the flag more heavily and copies the expression into the new `wheels` `runs-on`, so it may be worth fixing here.
> 
> GitHub expressions coerce `null → 0` and `false → 0`, so `matrix.build_on_push == false` evaluates **true** for the `amd64` and `arm64` entries, which don't set the key. Line 296 therefore routes arm64 to a generic `ubuntu-24.04` runner on every push. Job logs confirm it:
> 
> ```
> Build arm64   success   ubuntu-24.04      # expected ubuntu-24.04-arm
> ```
> 
> Same result across runs `29851649876`, `29743237284`, and `29852471833`. The `gate` steps are unaffected because they compare the rendered string (`"" != "false"`), which is why the arch still gets built — just not on arm hardware.
> 
> Simplest fix is to make the flag explicit for every entry, the way `build_on_pr` and `build_on_merge_group` already are:
> 
> ```yaml
> - platform: linux/amd64
>   ...
>   build_on_push: true
> - platform: linux/arm64
>   ...
>   build_on_push: true
> ```
> 
> ## Minor notes
> * **Event logic now lives in three places** — the `wheels` gate, the `build` gate, and the `IMAGES` array in `manifest`. Flip a flag in one and the manifest silently disagrees. One option is to build `IMAGES` from the downloaded `digest-<suffix>` artifacts so it can't drift from what was actually built.
> * **Header comments are now stale** — lines 8-9, 13-14 and 17 still say s390x/ppc64le build "on push and workflow_dispatch" and that "the multiplatform manifest is always complete". Line 304 says the same. The table in the PR description is accurate; moving it into the file header would help the next reader.
> * **`wheels` gate step is missing `set -euo pipefail`**, which the `build` gate has (line 313).
> * **Comment wording drifts** between the two jobs for the same flag: `# runner capacity — skip on normal push` in `wheels` vs `# temporarily disabled — s390x runner capacity issues` in `build` (line 283). Worth unifying.
> * **Concurrency**: `${{ github.workflow }}-${{ github.ref }}` with `cancel-in-progress` means a manual dispatch on `main` cancels an in-flight push run on the same ref, and vice versa — so the 4-arch manifest a dispatch produces can be cancelled by the next merge. Probably fine, just something to be aware of when using the new trigger.
> 
> Nothing else stood out — action SHAs stay pinned, `persist-credentials: false` is preserved on the newly gated checkout, and the existing `push:` / `outputs:` conditions (lines 421-422) already handled `workflow_dispatch` correctly and were rightly left alone.

@msureshkumar88 Thanks for the review!
The intent is to restrict s390x and ppc64le to workflow_dispatch only — including on release tags. For a versioned release that needs all four arches, the operator would trigger the workflow manually from the tag ref. @ja8zyjits this is what we intend to do, right?

msureshkumar88:
@madhu-mohan-jaishankar That answers it — thanks. "workflow_dispatch only, including on release tags, with the operator triggering manually from the tag ref" is a coherent policy, and it's your and @ja8zyjits' call to make. Happy to go with it.

Two things I'd still like to see before this lands, both small.

**1. Don't publish an incomplete manifest on the tag push itself.**

With the policy as stated, the tag push still runs `manifest` and `sign` (`if: github.event_name == 'push' || github.event_name == 'workflow_dispatch'`), so tagging `v1.0.7` publishes and Cosign-signs a 2-arch `:v1.0.7` / `:latest` / `:<sha>` a few minutes before the operator gets around to the manual dispatch. The dispatch then overwrites those tags with the 4-arch manifest.

That means `:latest` is briefly wrong on every release, signed and attested in that state — and if the manual step is ever missed, permanently wrong, with no signal that anything is off.

Cleaner to never publish the partial manifest at all:

```yaml
# manifest (and sign inherits via needs: manifest)
if: >-
  github.event_name == 'workflow_dispatch' ||
  (github.event_name == 'push' && github.ref_type != 'tag')
```

Then a tag push builds and pushes the per-arch `:amd64-<sha>` / `:arm64-<sha>` images but produces no combined tag, and the manual dispatch from the tag ref creates the full 4-arch `:v1.0.7` / `:latest` in one shot. Missing the manual step becomes "no release image published", which is loud and safe, rather than "release image quietly missing two arches".

(If you'd rather not build anything on tag push at all, gating `build` the same way works too — the dispatch rebuilds from cache anyway.)

Either way, worth spelling the process out in the workflow header, since it's now a required manual release step.

**2. The `matrix.build_on_push == false` null-coercion point is still open.**

Repeating it since it got lost in the thread and it's independent of the tag policy. GitHub coerces `null → 0` and `false → 0`, so `matrix.build_on_push == false` is **true** for the `amd64` and `arm64` entries, which don't set the key. The `runs-on` expression on line 296 therefore sends arm64 to a plain `ubuntu-24.04` runner on every push:

```
Build arm64   success   ubuntu-24.04      # expected ubuntu-24.04-arm
```

Consistent across runs `29851649876`, `29743237284`, `29852471833`. The gate steps aren't affected because they compare the rendered string, which is why the arch still gets built — just under emulation instead of on arm hardware.

It predates this PR, so no objection if you'd rather split it out into its own issue. But this PR copies the same expression into the new `wheels` `runs-on`, so fixing it here is cheap — add an explicit `build_on_push: true` to the amd64 and arm64 entries, matching how `build_on_pr` and `build_on_merge_group` are already spelled out for every entry.

The rest of my earlier notes (stale header comments, triplicated event logic, `set -euo pipefail`) are all non-blocking.

msureshkumar88 (approved):
Re-reviewed after `5ce4ac8`. All my blocking items are resolved — approving.

## Blocking item: closed

The tag-push arch question is settled as a policy decision, and the header now states it plainly (lines 15-16):

```
# Push-to-main and release tag builds produce a 2-arch manifest (amd64 + arm64).
# Trigger workflow_dispatch from main or a tag ref to produce the full 4-arch manifest.
```

That's exactly the "accept it and document it" option, and with @ja8zyjits' sign-off on the intent it's your call to make. The remediation path checks out too: dispatching from the tag ref gives `ref_type == 'tag'`, the gate passes for all four arches, and `manifest` re-tags `:vX.Y.Z` / `:latest` / `:<sha>` with the full set. Cosign is digest-keyed, so the dispatch run signs the new 4-arch digest and `:latest` verifies correctly after it.

One nice side effect worth knowing: `concurrency` is keyed on `${{ github.workflow }}-${{ github.ref }}` with `cancel-in-progress`, and a tag push and a dispatch from that same tag share a ref. So if the operator triggers the dispatch while the tag-push run is still in `build`, it cancels the push run and the partial 2-arch manifest never gets published at all. Triggering promptly after tagging is the cleanest way to run a release.

## Null coercion: fixed, and verified on both jobs

`build_on_push: true` is now explicit on the amd64 and arm64 entries (lines 272, 280), so `matrix.build_on_push == false` renders `false` rather than coercing from `null`, and line 301's `runs-on` sends arm64 back to `ubuntu-24.04-arm`. The `wheels` matrix only contains s390x and ppc64le, both with the key set explicitly, so its `runs-on` (line 167) is unambiguous as well. Worth confirming on the next push-to-main run that `Build arm64` reports `ubuntu-24.04-arm`.

## Other items from the first pass

- `set -euo pipefail` added to the wheels gate — done.
- Comment wording unified to `runner capacity — workflow_dispatch only` across both jobs — done.
- Top-of-file header rewritten and accurate — done.
- Manifest `IMAGES` comment now explains the dispatch-only inclusion — done.

The triplicated event logic and the concurrency note were non-blocking and I'm happy leaving both as-is.

## Two stale comments left (non-blocking)

The header block got cleaned up but two inner comments still describe the old push behaviour:

- **Line 139**, above the `wheels` job: "Rides the same non-PR, non-merge-group events as the platform builds below (push/workflow_dispatch)." Both matrix entries are now gated off on push, so wheels effectively only build on `workflow_dispatch`.
- **Lines 307-312**, above the `build` steps:
  ```
  # s390x/ppc64le: all steps skipped on PR and merge queue; full build+push on push
  ...
  # The `gate` step evaluates build_on_pr and build_on_merge_group for this
  ```
  The first line should read `workflow_dispatch` rather than `push`, and the second should mention `build_on_push` alongside the other two flags.

Both are comment-only and safe to fold into a follow-up if you'd rather not respin CI for them.

Nothing new turned up otherwise: action SHAs stay pinned, `persist-credentials: false` is preserved on both newly gated checkouts, the `push:` / `outputs:` conditions still handle `workflow_dispatch` correctly, and `manifest` keeps `needs: build` without `always()`, so a failed arch build skips the manifest rather than publishing a short one. Thanks for turning these around quickly.

ja8zyjits (approved):
LGTM

PR #5755, opened 2026-07-22
fix(ui): restore checked state on scroll-paginated edit selector pages

### LINKED ISSUE #3358: [BUG][UI]: Virtual Servers — Selected tools not shown/checked in dropdown during Edit
## Description

In Virtual Servers > Edit > Server Composition > Tools, the previously selected tools are not listed or shown as checked in the dropdown list.

## Steps to Reproduce

1. Create a virtual server with specific tools selected
2. Edit the virtual server
3. Navigate to Server Composition > Tools
4. Observe that selected tools are not shown or checked in the dropdown

## Expected Behavior

Previously selected tools should appear checked/highlighted in the dropdown list when editing.

## Category

Virtual Servers — Edit Form State

tw-360vier:
When editing a virtual server, the tools/resources/prompts association pickers load items lazily via HTMX infinite scroll — each page's sentinel element uses `hx-swap="outerHTML"` to append new items inside the container div.

The selection store (`getEditSelections`) is correctly seeded with all associated IDs when the modal opens, and Save reads from that store — **no data is lost**. This is a purely cosmetic/UX issue: associated items on page 2+ appear unchecked, which can mislead a user into thinking associations are missing even though they will be saved correctly.

## Repro
1. Associate more than ~20 tools with a virtual server.
2. Open the server's Edit modal.
3. Scroll down in the tools picker — page 2+ tools show unchecked even when they are associated.

## Root cause
`ensureEditStoreListeners()` in `servers.js` registers a `change` listener per container but no `htmx:afterSwap` handler. The existing afterSwap handler in `initialization.js` only wires up the Select All / Clear All buttons without restoring checked state, so checkboxes appended by scroll pagination stay visually unchecked.

## Fix
Extract `restoreCheckedStateFromStore(containerId)` in `servers.js` and call it from a new `htmx:afterSwap` listener registered alongside the existing `change` listener for each of the three edit-selector containers (`edit-server-tools`, `edit-server-resources`, `edit-server-prompts`). The swap event bubbles from the sentinel up to the container, so every scroll append re-applies checked state from the store. The save path and the create-server form are untouched.

## Tests
5 new cases in `tests/unit/js/edit-server-scroll-pagination.test.js` cover tools, resources, and prompts across multiple scroll pages, plus container isolation. All 5 fail before the fix; full suite after the fix: 2292 passed / 3 skipped, no regressions.

Closes #3358

### DIFF
--- a/mcpgateway/admin_ui/servers.js
+++ b/mcpgateway/admin_ui/servers.js
@@ -1216,6 +1216,25 @@ export function resetEditSelections() {
   });
 }
 
+/**
+ * Re-apply checked state from the persistent selection store to all checkboxes
+ * currently in the given container.  Called after any DOM change (search result
+ * render or HTMX scroll-paginated append) that may have added new checkboxes.
+ *
+ * @param {string} containerId - The id of the selector container.
+ */
+export function restoreCheckedStateFromStore(containerId) {
+  const container = document.getElementById(containerId);
+  if (!container) return;
+  const sel = getEditSelections(containerId);
+  if (sel.size === 0) return;
+  container.querySelectorAll("input[type=\"checkbox\"]").forEach((cb) => {
+    if (sel.has(String(cb.value))) {
+      cb.checked = true;
+    }
+  });
+}
+
 export function ensureEditStoreListeners() {
   if (window._editStoreListenersAttached) return;
   window._editStoreListenersAttached = true;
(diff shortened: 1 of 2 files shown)


tw-360vier:
Closing this. On a closer read of the admin UI, the scroll-append restore this patch adds is already handled by the existing `htmx:afterSettle` handlers in `mcpgateway/admin_ui/app.js`, which re-check `data-auto-check` items against the selection store on each paginated load. So the change is redundant and the root-cause description here was inaccurate — apologies for the noise.

The real friction is narrower: a virtual server's already-associated tools aren't surfaced near the top of the edit picker, so on a large tool list they aren't visible without scrolling. I'll look into an 'associated-first' ordering approach separately.

PR #5759, opened 2026-07-22
chore: eliminate .secrets.baseline merge-queue conflicts

### LINKED ISSUE #5758: [CHORE]: .secrets.baseline causes recurring merge-queue conflicts across concurrent PRs
## Problem

`.secrets.baseline` regularly causes merge conflicts / merge-queue friction when multiple open PRs touch it concurrently. Investigated 5 PRs queued together (#5244, #5610, #5626, #5544, #4862) — 3 of them (#5244, #5626, #4862) modify `.secrets.baseline`.

### Root cause

Diffing each PR's baseline change against its own merge-base shows the changes are **not new secrets** — they're two mechanical side effects of `make detect-secrets-scan`:

1. **`generated_at` timestamp** (line 3 of the JSON) is rewritten on every regen. Any two PRs that both ran the scan will conflict on this exact line when merged back-to-back, regardless of what else changed.
2. **`line_number` fields drift** for existing (already-audited) findings whenever a PR inserts/removes lines earlier in a scanned file (`.env.example`, `config.py`, docs). PRs editing the same scanned file (e.g. #4862 and #5626 both touch `.env.example` and `config.py`) produce diverging line-number entries for the same underlying findings.

Since the merge queue processes PRs against a rolling base and the CI step (`make detect-secrets-hook`, `pre-commit.yml`) only *validates* the baseline (`--fail-on-unaudited`) rather than regenerating it, a baseline generated against a stale base becomes a conflict source (or silently stale) for whichever PR merges second/third.

### Why this is fixable safely

Checked `detect-secrets`' matching logic directly:

```python
# detect_secrets/core/potential_secret.py
self.fields_to_compare = ['filename', 'secret_hash', 'type']
```

Audit identity is `filename + secret_hash + type` — **`line_number` is not part of the match key**. It's cosmetic (used only for human display in `detect-secrets audit`). This means line-number drift is always safe to regenerate away; it never causes a functional `--fail-on-unaudited` failure by itself, only git-level textual conflicts that block auto-merge.

## Proposed fix

1. **Pin `generated_at`** to a fixed sentinel value as part of the `detect-secrets-scan` Makefile target (post-process the JSON after scan), so it stops being a source of guaranteed conflicts.
2. **Add a baseline auto-sync workflow**: on push to `main`, find open PRs that modify `.secrets.baseline`, merge `main` into them, rerun `make detect-secrets-scan`, and auto-push the refreshed baseline back to the PR branch — as long as no new *unaudited* finding is introduced (i.e., only drift/timestamp changes). This keeps open PRs' baselines continuously fresh so the merge queue never sees stale drift.

## Impact if unaddressed

Every batch of concurrent PRs that touch scanned files (`.env.example`, `config.py`, docs, etc.) risks ejection from the merge queue on `.secrets.baseline` conflicts, requiring manual rebase + rescan per PR, serialized one at a time.

msureshkumar88:
## Summary

- Pin `generated_at` in `.secrets.baseline` to a fixed sentinel value after `make detect-secrets-scan`, so regenerating the baseline no longer collides on that line whenever two or more PRs touch it.
- Add `secrets-baseline-sync.yml`: on push to `main`, finds open, same-repo PRs that touch `.secrets.baseline`, merges `main` in, reruns the scan, and pushes the refreshed baseline back — but only if `detect-secrets-hook --fail-on-unaudited` still passes (i.e. only drift, never a silent new-secret allowlist). PRs with real conflicts or new unaudited findings are left alone for manual handling.

## Why

Investigated a batch of 5 PRs headed for the merge queue; 3 touched `.secrets.baseline` and their diffs were entirely mechanical: the `generated_at` timestamp plus `line_number` drift on already-audited findings. `detect-secrets`' own audit identity is `filename + secret_hash + type` — line numbers aren't part of it (`PotentialSecret.fields_to_compare` in `detect_secrets/core/potential_secret.py`), so this drift is always safe to regenerate away. Left unaddressed, any concurrent batch of PRs editing scanned files (`.env.example`, `config.py`, docs) risks getting ejected from the merge queue on baseline conflicts, requiring manual serialized rebase+rescan per PR.

Closes #5758

## Test plan
- [ ] `make detect-secrets-scan` produces a baseline with `generated_at` pinned to `1970-01-01T00:00:00Z`, verified against a copy of the current file (single-line diff)
- [ ] `secrets-baseline-sync.yml` YAML validated
- [ ] Manually trigger the workflow (`workflow_dispatch`) once merged and confirm it correctly skips PRs with no baseline drift and pushes only clean drift-only updates

### DIFF
--- a/.github/workflows/secrets-baseline-sync.yml
+++ b/.github/workflows/secrets-baseline-sync.yml
@@ -1,0 +1,95 @@
+# ======================================================================
+# Sync .secrets.baseline on open PRs after main advances
+# ======================================================================
+#
+# .secrets.baseline stores a line_number per finding. That number is not
+# part of detect-secrets' audit identity (filename + secret_hash + type),
+# so it's safe to regenerate — but any PR branch that falls behind main
+# accumulates line-number drift in the file, which turns into merge
+# conflicts (or, worse, a stale baseline that fails
+# `make detect-secrets-hook --fail-on-unaudited`) once several such PRs
+# land back-to-back through the merge queue.
+#
+# This workflow keeps open, same-repo PRs that touch .secrets.baseline
+# continuously synced with main: merge main in, regenerate the baseline,
+# and push back — but only when regeneration doesn't introduce a new,
+# unaudited finding (i.e. only drift, never a silent secret allowlist).
+# PRs from forks are skipped (no push access) and left for a human/manual
+# rebase instead.
+#
+# ======================================================================
+
+name: Sync secrets baseline on open PRs
+
+on:
+  push:
+    branches: ["main"]
+  workflow_dispatch:
+
+concurrency:
+  group: secrets-baseline-sync
+  cancel-in-progress: false
+
+permissions:
+  contents: write
+  pull-requests: read
+
+jobs:
+  sync:
+    runs-on: ubuntu-latest
+    if: github.repository == 'IBM/mcp-context-forge'
+    steps:
+      - name: Checkout
+        uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5
+        with:
+          fetch-depth: 0
+
+      - name: Set up uv
+        uses: astral-sh/setup-uv@08807647e7069bb48b6ef5acd8ec9567f424441b # v8.1.0
+
+      - name: Sync .secrets.baseline on affected open PRs
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          set -euo pipefail
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+
+          prs=$(gh pr list --state open --json number,headRefName,headRepositoryOwner,isDraft,files \
+            --jq '.[] | select(.isDraft == false) | select(.headRepositoryOwner.login == "IBM") | select([.files[].path] | index(".secrets.baseline")) | .number')
+
+          for pr in $prs; do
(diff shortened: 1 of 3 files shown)


brian-hussey (changes requested):
I like the idea but I don't think this will work as expected.
Once we get to the merge queue there will still be issues and we cannot merge updates into the merge queue - like generating the secrets baseline file again.

I like the defaulting the timestamp back to epoch, I think keep that.
I also think instead of an action on this to completely regenerate the .secrets.baseline and update the branch, can we try setting the gitattributes for .secrets.baseline to always merge theirs (and at least this will keep it up to date for rebasing.

After that I think we should revisit the action to update in place that can only apply to PRs but not items that have been sent to the merge queue.

What do you think?

brian-hussey:
@jonpspri is trying to solve the secrets baseline problem a different way in this as well at the moment:
https://github.com/IBM/mcp-context-forge/pull/4453/

msureshkumar88:
Closing in favor of #4453, which `@jonpspri` already has in flight covering the same ground: `.secrets.baseline merge=secrets-baseline` in `.gitattributes` plus a custom merge driver (`resolve-secrets-baseline-conflict.sh`) and a `post-rewrite-secrets-refresh.sh` git hook wired into `make install-dev`. That's a more complete version of the approach discussed here, so no need to duplicate it.

Thanks for the review, @brian-hussey — the epoch-pin on `generated_at` is a small independent win; happy to reopen a scoped PR for just that if useful, otherwise it can ride along with #4453.

Issue #2527, opened 2026-01-27
[EPIC][SECURITY][MCP-EXTENSIONS]: Governed Extension Framework for MCP Apps and Future Extensions


crivetimihai:
### Type of Feature

- [x] Enhancement to existing functionality
- [x] New feature or capability
- [x] New component or integration
- [ ] New MCP-compliant server
- [x] Security / governance hardening
- [ ] Developer tooling or test improvement
- [ ] Packaging, automation and deployment
- [ ] Other

### Epic

**Title:** Governed Extension Framework for MCP Apps and Future MCP Extensions

**Context:** MCP Apps support is already being introduced in #5079. This epic is not a prerequisite for MCP Apps and should not treat MCP Apps as the first implementation criterion. Instead, this epic defines the generic security and governance framework that MCP Apps and future MCP extensions must follow.

**Goal:** Transform ContextForge from a gateway that can project extension metadata into a security-first extension gateway that can discover, govern, aggregate, route, sanitize, audit, and operate MCP extensions without raw pass-through or bespoke dispatcher branches for every new extension.

**Core principle:** MCP extension discovery is not permission to expose or execute the extension. ContextForge must treat extensions as untrusted optional protocol modules until policy, RBAC, token scoping, server scope, client capability negotiation, and gateway configuration allow them.

**MCP Apps relationship:** MCP Apps should remain a complex extension because UI resources, AppBridge sessions, CSP, sandboxing, permissions policy, external navigation, and app-visible tools need custom handling. However, MCP Apps should plug into the same extension registry, policy engine, capability aggregation, routing, plugin, sanitization, and audit layers as every other extension.

**Generic framework principle:** Adding a simple future extension such as Tasks should not require changes to the ContextForge core request dispatcher, initialize handshake, database schema, or UI plumbing. Simple extensions should be onboarded through registry/configuration metadata: extension identifier, capability payload, explicit method patterns, permissions, routing policy, source gateways, sanitizer policy, plugin policy, and audit policy. Code changes should only be needed when an extension requires custom gateway-side behavior such as state management, protocol adaptation, security mediation, or content transformation.

**Protocol model:** Official MCP extensions should be represented through `capabilities.extensions` using stable extension identifiers such as `io.modelcontextprotocol/ui`. ContextForge may preserve `capabilities.experimental` for backward compatibility and third-party incubation, but official extension support should use the extension negotiation model.

---

## Security Model

[...]

PR #5764, opened 2026-07-22
fix(mcp-apps): honour deprecated flat ui/resourceUri metadata key

### LINKED ISSUE #5762: [FEATURE]: Honour deprecated flat `_meta["ui/resourceUri"]` metadata key
### 🧭 Type of Feature

- [ ] Other

Raised at the request of @vishu-bh in https://github.com/IBM/mcp-context-forge/issues/2527#issuecomment-5044901298, as one of the two compatibility gaps found while diffing the merged #5079 implementation against the finalized MCP Apps specification.

### 🧭 Epic

**Title:** Honour the deprecated flat `_meta["ui/resourceUri"]` metadata key

**Goal:** Preserve the tool→UI association for upstream MCP servers that still emit only the deprecated flat metadata key, so those servers keep working through ContextForge until the key is removed from the specification.

**Why now:** The flat key is still live in the spec (`specification/2026-01-26/apps.mdx:341`, `:347` — *"The deprecated format will be removed before GA"*). ContextForge does not read it anywhere, so tools from servers using only the legacy shape are ingested with their UI association silently dropped. This is precisely the backward-compatibility case the deprecated key exists to serve.

---

### 🐞 Current behaviour

`merge_mcp_protocol_meta` (`mcpgateway/services/mcp_apps.py:139`) reads only the nested `_meta.ui` object:

```python
ui = meta.get("ui")
if not isinstance(ui, dict) or not ui:
    return
```

A tool whose `_meta` is `{"ui/resourceUri": "ui://widgets/example"}` therefore returns early, no `extensionMetadata` is written, and the tool loses its UI resource association during gateway sync. There are no occurrences of `ui/resourceUri` anywhere in the repository.

### 📐 Scope of the impact

Narrow but real. The reference SDK normalizes bidirectionally in `registerAppTool` ([`src/server/index.ts:238-262`](https://github.com/modelcontextprotocol/ext-apps/blob/main/src/server/index.ts)), so any SDK-built server emits **both** keys and works through ContextForge today. Only servers hand-rolling the legacy key alone are affected.

### 🧑🏻‍💻 User Story 1

**As a:** Platform Administrator federating an older MCP Apps server
**I want:** its tools to keep their `ui://` association through the gateway
**So that:** the app renders instead of silently degrading to a plain tool

#### ✅ Acceptance Criteria

```gherkin
Scenario: Legacy-only server keeps its tool/UI association
  Given an upstream tool whose _meta contains only "ui/resourceUri": "ui://widgets/example"
  When the gateway validates the tool during sync
  Then extensionMetadata["io.modelcontextprotocol/ui"].resourceUri is "ui://widgets/example"

Scenario: Canonical nested key wins on disagreement
  Given an upstream tool emitting both _meta.ui.resourceUri and _meta["ui/resourceUri"] with different values
  When the gateway validates the tool
  Then the nested value is stored
  And the discarded legacy value is logged
```

### 🧑🏻‍💻 User Story 2

**As a:** Gateway operator
**I want:** a malformed legacy value to never make a tool disappear
**So that:** backward compatibility cannot become an availability regression

#### ✅ Acceptance Criteria

```gherkin
Scenario: Invalid legacy URI is ignored rather than fatal
  Given an upstream tool whose _meta contains "ui/resourceUri": "https://example.com/widget.html"
  When the gateway validates the tool
  Then no MCP Apps metadata is written
  And the tool is still ingested
```

This second story matters more than it looks. `validate_extension_metadata` (`mcp_apps.py:239-241`) rejects a `resourceUri` that does not use the `ui://` scheme, and `_validate_tools` (`gateway_service.py:6576`) turns that into a `ValidationError` that drops the tool from the sync entirely. Folding an unvalidated legacy value in would convert "ingested without UI" into "disappears", so only well-formed `ui://` strings should be applied.

### 🔗 MCP Standards Check

- [x] Change adheres to the current MCP Apps specification
- [x] Additive only — behaviour is unchanged for payloads that do not carry the legacy key

Precedence follows the reference SDK's consumer-side rule in `getToolUiResourceUri` ([`src/app-bridge.ts:125-133`](https://github.com/modelcontextprotocol/ext-apps/blob/main/src/app-bridge.ts)): *"The new nested format takes precedence if both are present."* Worth noting that the SDK's producer side deliberately does **not** reconcile a disagreement — `registerAppTool` emits both conflicting values on the wire — so the consumer rule is the only one a gateway can apply.

### 🔄 Alternatives Considered

**Emit the legacy key outbound as well (full bidirectional normalization).** The SDK does this for older hosts. Not proposed here: ContextForge's own projection (`apply_tool_meta`, `mcp_apps.py:319`) emits the nested shape only, and propagating a format scheduled for removal before GA seems like the wrong direction for a gateway. Easy to add later if hosts need it.

**Normalize in `validate_extension_metadata` instead.** Rejected — the merge step is where protocol shape is translated into internal state, and validation runs on stored metadata that has no `_meta` to read from.

### 📓 Additional Context

Parent epic: #2527. Companion issue for the other compatibility gap will be linked to #4970.

PhilipAD:
# 🐛 Bug-fix PR

## 📌 Summary

Servers that emit only the deprecated flat `_meta["ui/resourceUri"]` key lose their tool→UI association when passing through the gateway. This folds the legacy key into the nested MCP Apps metadata during the existing merge step.

Closes #5762. Part of #2527.

## 🔁 Reproduction Steps

Federate an upstream MCP server exposing a tool whose `_meta` carries only the legacy key:

```json
{ "name": "customer_search", "_meta": { "ui/resourceUri": "ui://widgets/customer-search" } }
```

After gateway sync the tool is ingested with no `extensionMetadata`, so it has no UI association and the app never renders. Covered as a regression test in `test_merge_mcp_protocol_meta_honours_deprecated_flat_resource_uri`.

## 🐞 Root Cause

`merge_mcp_protocol_meta` (`mcpgateway/services/mcp_apps.py:139`) reads only the nested object and returns early when it is absent:

```python
ui = meta.get("ui")
if not isinstance(ui, dict) or not ui:
    return
```

The flat key had no occurrences anywhere in the repository, so it was never consulted. The spec keeps it live (`apps.mdx:341`, `:347` — *"The deprecated format will be removed before GA"*).

## 💡 Fix Description

The legacy key is now folded into the `ui` block before the merge, behind a small helper. Three design points:

**Nested wins on disagreement.** This follows the reference SDK's consumer-side rule in `getToolUiResourceUri` ([`app-bridge.ts:125-133`](https://github.com/modelcontextprotocol/ext-apps/blob/main/src/app-bridge.ts)): *"The new nested format takes precedence if both are present."* Worth noting the SDK's producer side deliberately does **not** reconcile — `registerAppTool` emits both conflicting values on the wire ([`server/index.ts:238-262`](https://github.com/modelcontextprotocol/ext-apps/blob/main/src/server/index.ts), test `server/index.test.ts:163`) — so the consumer rule is the only one a gateway can apply. A genuine disagreement is logged.

**Only well-formed `ui://` strings are folded in.** This is the part I'd most like a second opinion on. `validate_extension_metadata` (`mcp_apps.py:239-241`) rejects a non-`ui://` `resourceUri`, and `_validate_tools` (`gateway_service.py:6576`) turns that into a `ValidationError` that drops the tool from the sync. Folding an unvalidated legacy value in would therefore convert "ingested without UI" into "tool disappears entirely" — a worse outcome than the bug. Unusable values keep the existing ignore-it behaviour and are logged instead. The SDK throws here; for a gateway, failing closed on ingest seemed wrong.

**Ingest only, not bidirectional.** The SDK also emits the flat key outbound for older hosts. Not done here: ContextForge's own projection (`apply_tool_meta`, `mcp_apps.py:319`) emits the nested shape only, and propagating a format scheduled for removal before GA seems like the wrong direction for a gateway. Trivial to add if you'd prefer symmetry.

The source `_meta` is copied rather than mutated, so callers passing shared dicts are unaffected.

## 📏 Reviewability

On the `triage` checkbox: #5762 currently carries no labels. Filing it and this PR was explicitly requested by @vishu-bh in https://github.com/IBM/mcp-context-forge/issues/2527#issuecomment-5044901298, which CONTRIBUTING lists as one of the conditions for starting work. If you apply `triage` while scoping it, treat this PR as parked until you're ready.

## 🧪 Verification

Added 12 tests to `tests/unit/mcpgateway/services/test_mcp_apps.py`, covering: legacy-only payloads, legacy alongside nested siblings, both shapes agreeing, both disagreeing (nested wins), five classes of unusable legacy value, and non-mutation of the caller's `_meta`.

| Check | Command | Status |
|---|---|---|
| MCP Apps + AppBridge + registry suites | `pytest tests/unit/mcpgateway/services/test_mcp_apps.py tests/unit/mcpgateway/test_mcp_apps_security.py tests/unit/mcpgateway/test_mcp_apps_protocol.py tests/unit/mcpgateway/test_mcp_method_registry.py` | ✅ 98 passed (86 before this change) |
| Ingest-path regression | `pytest tests/unit/mcpgateway/services/test_gateway_service.py tests/unit/mcpgateway/services/test_resource_service.py` | ✅ no new failures |
| Ruff | `ruff check mcpgateway/services/mcp_apps.py tests/...` | ✅ All checks passed |
| Black (line length 200) | `black --check` | ✅ unchanged |
| Doctests | `pytest --doctest-modules mcpgateway/services/mcp_apps.py` | ✅ no doctests in module, none broken |

Two pre-existing errors in `test_resource_service.py` (`test_read_resource_template_proxy_blob_none_response_raises`, `test_invoke_resource_sse_returns_content_when_session_teardown_raises`) reproduce identically on unmodified `main`, so they are unrelated to this change.

Not run locally: `make docker docker-run-ssl` and the postgres/redis matrix. This change is a pure in-memory dict transformation with no database, network, or migration surface, so I don't believe it carries container- or backend-specific risk — but say the word if you want that evidence before merging.

## 📐 MCP Compliance

Strictly additive: payloads without the legacy key take an unchanged path.

## ✅ Checklist

- [x] Code formatted

### DIFF
--- a/docs/docs/architecture/mcp-apps.md
+++ b/docs/docs/architecture/mcp-apps.md
@@ -77,6 +77,38 @@ and resources. MCP Apps metadata is keyed by the MCP UI capability identifier:
 The gateway also accepts the internal snake_case form, `extension_metadata`, in
 Python service paths. Public API payloads should use `extensionMetadata`.
 
+### Deprecated Flat `ui/resourceUri` Key
+
+Upstream servers advertise Apps metadata on the wire as `_meta`, which the
+gateway folds into `extensionMetadata` during ingest. Alongside the nested
+`_meta.ui` object, the gateway still honours the deprecated flat
+`_meta["ui/resourceUri"]` key so that servers emitting only the older shape keep
+their tool/UI association:
+
+```json
+{
+  "name": "customer_search",
+  "_meta": { "ui/resourceUri": "ui://widgets/customer-search" }
+}
+```
+
+Precedence and handling:
+
+- **The nested object wins.** If `_meta.ui` contains `resourceUri` (or
+  `resource_uri`), the flat key is ignored. Precedence is decided by key
+  *presence*, not truthiness, so a nested value that is empty or malformed stays
+  authoritative and is rejected by validation rather than being silently
+  replaced by the deprecated value.
+- **Only well-formed `ui://` values are folded in.** An unusable flat value is
+  ignored and logged. Folding it in would fail extension metadata validation and
+  drop the whole tool from gateway sync, which is worse than ingesting the tool
+  without a UI association.
+- **Ingest only.** ContextForge always projects the nested shape outbound; it
+  does not re-emit the deprecated key.
+
+The flat key is scheduled for removal from the spec before GA, so treat it as a
+compatibility path rather than a supported input format.
+
 ### Tool Metadata
 
 Tool metadata controls who sees or can invoke the tool:
--- a/mcpgateway/services/mcp_apps.py
+++ b/mcpgateway/services/mcp_apps.py
@@ -33,6 +33,9 @@ logger = logging.getLogger(__name__)
 MCP_UI_EXTENSION = "io.modelcontextprotocol/ui"
 MCP_UI_DEFAULT_VERSION = "2026-01-26"
 MCP_APP_MIME_TYPE = "text/html;profile=mcp-app"
+# Deprecated flat metadata key kept for backward compatibility with servers that
+# predate the nested ``_meta.ui`` shape. Removed from the spec before GA.
+LEGACY_RESOURCE_URI_META_KEY = "ui/resourceUri"
 
 _ALLOWED_CSP_DIRECTIVES = frozenset(
     {
@@ -136,19 +139,51 @@ def mcp_ui_metadata(value: Any) -> Dict[str, Any]:
     return ui if isinstance(ui, dict) else {}
 
 
+def _legacy_ui_resource_uri(meta: Dict[str, Any], ui: Dict[str, Any]) -> Optional[str]:
+    """Return the deprecated flat ``ui/resourceUri`` value when it should be applied."""
+    legacy_uri = meta.get(LEGACY_RESOURCE_URI_META_KEY)
+    if not isinstance(legacy_uri, str) or not legacy_uri:
+        return None
+
+    nested_uri = ui.get("resourceUri") or ui.get("resource_uri")
+    if nested_uri:
+        # The nested key is canonical and wins; upstream servers may legitimately
+        # emit both, so only a genuine disagreement is worth reporting.
+        if nested_uri != legacy_uri:
+            logger.info("Ignoring deprecated _meta[%r]=%r in favour of nested resourceUri %r", LEGACY_RESOURCE_URI_META_KEY, legacy_uri, nested_uri)
+        return None
+
+    if not legacy_uri.startswith("ui://"):
+        # Folding an invalid URI in would fail extension metadata validation and
+        # drop the whole tool, so keep the pre-existing "ignore it" behaviour.
+        logger.warning("Ignoring deprecated _meta[%r]=%r because it does not use the ui:// scheme", LEGACY_RESOURCE_URI_META_KEY, legacy_uri)
+        return None
+
+    return legacy_uri
+
+
 def merge_mcp_protocol_meta(payload: Dict[str, Any]) -> None:
     """Translate MCP protocol ``_meta.ui`` into internal extension metadata.
 
     Upstream MCP servers advertise Apps metadata on protocol objects as
     ``_meta: {"ui": ...}``, while ContextForge stores extension state as
     ``extensionMetadata: {"io.modelcontextprotocol/ui": ...}``.
+
+    The deprecated flat ``_meta["ui/resourceUri"]`` key is also honoured so that
+    servers still emitting only the legacy shape keep their tool/UI association.
     """
     meta = payload.get("_meta")
     if not isinstance(meta, dict):
         return
 
-    ui = meta.get("ui")
-    if not isinstance(ui, dict) or not ui:
+    raw_ui = meta.get("ui")
+    ui = dict(raw_ui) if isinstance(raw_ui, dict) else {}
+
+    legacy_uri = _legacy_ui_resource_uri(meta, ui)
+    if legacy_uri:
+        ui["resourceUri"] = legacy_uri
+
+    if not ui:
         return
 
     extension_metadata = payload.get("extensionMetadata") or payload.get("extension_metadata")
--- a/mcpgateway/services/mcp_apps.py
+++ b/mcpgateway/services/mcp_apps.py
@@ -16,6 +16,7 @@ from datetime import datetime, timedelta, timezone
 import logging
 import re
 from typing import Any, Dict, Iterable, List, Optional
+from urllib.parse import urlsplit, urlunsplit
 import uuid
 
 # Third-Party
@@ -41,6 +42,8 @@ UI_URI_SCHEME = "ui://"
 NESTED_RESOURCE_URI_KEYS = ("resourceUri", "resource_uri")
 # Upstream-controlled URIs are bounded before they reach the logs.
 MAX_LOGGED_URI_LENGTH = 120
+# Stands in for an authority that cannot be shown to be credential-free.
+REDACTED_AUTHORITY = "***"
 
 _ALLOWED_CSP_DIRECTIVES = frozenset(
     {
@@ -146,10 +149,23 @@ def mcp_ui_metadata(value: Any) -> Dict[str, Any]:
 
 def _loggable_uri(value: str) -> str:
     """Reduce an upstream-controlled URI to a bounded, credential-free form for logging."""
-    safe = value.split("?", 1)[0].split("#", 1)[0]
-    scheme, separator, remainder = safe.partition("://")
-    if separator and "@" in remainder.split("/", 1)[0]:
-        safe = f"{scheme}://{remainder.split('@', 1)[1]}"
+    try:
+        parts = urlsplit(value)
+    except ValueError:
+        # Malformed enough that no component can be trusted; the length is the only safe detail.
+        return f"<unparseable {len(value)}-character URI>"
+
+    netloc = parts.netloc.rpartition("@")[2]
+    path = parts.path
+    if "@" in parts.query or "@" in parts.fragment or (not parts.netloc and "@" in path):
+        # The authority ended part-way through what may still be credential material: either an
+        # unencoded delimiter pushed the rest of the userinfo into the query or fragment
+        # (``https://user:pa?ss@host/``), or the URI is not hierarchical so there is no authority
+        # to strip (``mailto:user@host``). Nothing preceding that ``@`` can be shown to be safe.
+        netloc = REDACTED_AUTHORITY if parts.netloc else netloc
+        path = path if parts.netloc else path.rpartition("@")[2]
+
+    safe = urlunsplit((parts.scheme, netloc, path, "", ""))
     if len(safe) > MAX_LOGGED_URI_LENGTH:
         safe = f"{safe[:MAX_LOGGED_URI_LENGTH]}...(truncated)"
     return safe
@@ -172,7 +188,7 @@ def _legacy_ui_resource_uri(meta: Dict[str, Any], ui: Dict[str, Any]) -> Optiona
             logger.info("Ignoring deprecated _meta[%r] in favour of the nested %s already present", LEGACY_RESOURCE_URI_META_KEY, nested_key)
         return None
 
-    if not legacy_uri.startswith(UI_URI_SCHEME) or not legacy_uri[len(UI_URI_SCHEME) :].strip():
+    if not legacy_uri.startswith(UI_URI_SCHEME) or not legacy_uri.removeprefix(UI_URI_SCHEME).strip():
         # Folding an unusable URI in would fail extension metadata validation and
         # drop the whole tool, so keep the pre-existing "ignore it" behaviour.
         logger.warning("Ignoring deprecated _meta[%r]=%r because it is not a usable %s URI", LEGACY_RESOURCE_URI_META_KEY, _loggable_uri(legacy_uri), UI_URI_SCHEME)
--- a/mcpgateway/services/mcp_apps.py
+++ b/mcpgateway/services/mcp_apps.py
@@ -36,6 +36,11 @@ MCP_APP_MIME_TYPE = "text/html;profile=mcp-app"
 # Deprecated flat metadata key kept for backward compatibility with servers that
 # predate the nested ``_meta.ui`` shape. Removed from the spec before GA.
 LEGACY_RESOURCE_URI_META_KEY = "ui/resourceUri"
+UI_URI_SCHEME = "ui://"
+# Canonical nested keys that take precedence over the deprecated flat one.
+NESTED_RESOURCE_URI_KEYS = ("resourceUri", "resource_uri")
+# Upstream-controlled URIs are bounded before they reach the logs.
+MAX_LOGGED_URI_LENGTH = 120
 
 _ALLOWED_CSP_DIRECTIVES = frozenset(
     {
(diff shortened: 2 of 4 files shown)


vishu-bh (changes requested):
Thanks for the focused fix and thorough regression coverage. The legacy-only behavior described in #5762 is addressed, and preserving the caller’s `_meta` object is a good detail.

I found one functionally impacting edge case that should be addressed before merging. In `_legacy_ui_resource_uri`, canonical-key precedence currently depends on truthiness:

    nested_uri = ui.get("resourceUri") or ui.get("resource_uri")
    if nested_uri:

A present but falsey canonical value, such as `""`, `False`, or `0`, is therefore treated as absent and replaced by the deprecated URI. The reference consumer falls back only when the canonical value is undefined, so a malformed canonical value remains authoritative and is subsequently rejected. Could we check key presence explicitly and add coverage for empty or malformed canonical values?

As a security-hardening point, the new log messages include complete upstream-controlled URI values. Although `%r` prevents newline-based log forging, a compromised upstream could submit very large values or expose sensitive URI query/userinfo content in logs. It would be safer to omit, redact, or truncate these values.

Also, `startswith("ui://")` accepts values such as `ui://` and other structurally malformed URIs. This matches existing validation behavior, so I do not consider it a new security regression, but using a shared stricter URI validator would improve hardening.

The targeted MCP Apps suites passed, as did Ruff, Bandit, Interrogate, Pylint, and package verification. `make pre-commit` stopped because detect-secrets updated `.secrets.baseline` line numbers. Since the branch is behind current `main`, please rebase and rerun the hook, including any resulting baseline update.

Overall, the approach looks sound. I recommend addressing the canonical-precedence edge case and pre-commit failure before merge. The logging adjustment would also be valuable security hardening.

PhilipAD:
Thanks — the canonical-precedence point is a real bug, and you're right about the rule: the reference consumer falls back only when the canonical value is `undefined`, not when it is falsey.

**Precedence is now decided by key presence.** `_legacy_ui_resource_uri` walks `("resourceUri", "resource_uri")` and returns as soon as either key is *present*, whatever it holds. A canonical `""`, `False`, `0`, `None`, or an `https://` value therefore stays authoritative and is rejected downstream by `validate_extension_metadata`, instead of being silently swapped for the deprecated URI. New coverage: `test_merge_mcp_protocol_meta_keeps_present_but_unusable_nested_resource_uri`, parametrized over both key spellings × five unusable values, asserting the metadata comes out holding the canonical value untouched.

**Logging.** Rejected URIs no longer reach the logs verbatim. `_loggable_uri` strips the query and fragment, drops any userinfo from the authority, and truncates the remainder to 120 characters. The nested-vs-flat disagreement message no longer logs either URI at all — it only names which canonical key won, which is the actionable part. `test_merge_mcp_protocol_meta_does_not_log_full_flat_resource_uri` asserts the credential, the query token, and the full value are all absent from the log record.

**`startswith("ui://")`.** Agreed it is loose. I have kept the shared-validator change out of this PR because `validate_extension_metadata` (`mcp_apps.py:293`) and `validate_ui_resource` (`:308`) use the same predicate on the ingest path, and tightening it there changes which upstream tools survive gateway sync — a behaviour change that deserves its own issue rather than riding along on a backward-compatibility fix. What I have done is make the *fold* stricter than the validator in the safe direction: a legacy value of `ui://` or `ui://   ` is no longer folded in, since it can never resolve to a real UI resource. Rejecting it there only preserves the pre-existing ignore-it behaviour, so it cannot drop a tool that works today. Happy to file the shared-validator hardening separately if you want it tracked.

**Rebase and pre-commit.** Rebased onto `3fe3035`. `pre-commit run --files mcpgateway/services/mcp_apps.py tests/unit/mcpgateway/services/test_mcp_apps.py` passes end to end, detect-secrets included, and leaves `.secrets.baseline` unmodified — the earlier churn was the stale base. One new inline `pragma: allowlist secret` was needed on the logging test, whose fixture URI deliberately carries `user:s3cret@` to prove it gets redacted.

Targeted suites are now 111 passed (98 before), with ruff and black clean.

vishu-bh (dismissed):
Thanks @PhilipAD!!

Approach is solid and clean. 

LGTM 🚀

msureshkumar88 (changes requested):
Nice fix — the design writeup is genuinely thorough (verifying precedence against the SDK's own `getToolUiResourceUri`/`registerAppTool` source, and the fail-closed handling of malformed legacy values so a bad value can't drop a tool from sync). Scope is tight too: only `mcp_apps.py` + its test file, no schema/migration surface, no unrelated diff noise.

A couple of things I'd like addressed before merging, plus a couple of optional notes.

## Blocking

**1. `_loggable_uri` can leak a credential fragment (`mcpgateway/services/mcp_apps.py:147-155`)**

The redaction strips `?query`/`#fragment` *before* stripping `user:pass@` from the netloc. If a malformed/malicious legacy value has a `?` or `#` inside the credential portion itself, the query-strip runs first and truncates before the userinfo-strip logic ever sees it — so part of the credential survives into the log line. Reproduced:

```python
>>> _loggable_uri("https://user:pa?ss@example.com/x")
'https://user:pa'   # 'pa' leaked
```

Since this function exists specifically to keep upstream-controlled URIs credential-free in logs, worth rebuilding on `urllib.parse.urlsplit`/`urlunsplit` (strip netloc userinfo first, then truncate) rather than manual ordered string splitting — removes the ordering dependency entirely.

**2. No test through the actual ingest path**

All 12 new tests call `merge_mcp_protocol_meta` directly. The issue's acceptance criteria phrase the scenario as "when the gateway validates the tool during sync" — could we get one test that goes through `GatewayService._validate_tools` (or equivalent) with a legacy-only tool dict, confirming it survives `ToolCreate` validation end-to-end?

**3. No coverage for the resource / resource-template ingest paths**

`merge_mcp_protocol_meta` is also called from 6 other sites in `gateway_service.py` for resources and resource templates, not just the one tool call site. Since this PR changes shared behavior of that function, it'd be good to have at least one test confirming a legacy-only *resource* keeps its UI association too (or confirming nothing regresses there).

## Non-blocking / optional

- `docs/docs/architecture/mcp-apps.md`'s "Metadata Model" section currently only documents the nested `_meta.ui` shape — might be worth a short note on the legacy-key fallback and precedence rule, since the whole point of this change is compatibility for hosts still emitting it.
- Minor style nit: `legacy_uri[len(UI_URI_SCHEME):].strip()` could be `legacy_uri.removeprefix(UI_URI_SCHEME).strip()` (repo targets Python ≥3.11).
- Small data-hygiene note: the legacy key stays in `_meta` after folding (not stripped), so it persists in stored metadata indefinitely. Not a bug, just worth being aware of if anything downstream ever reads `_meta` directly expecting only canonical keys.

Happy to take another look once these land.

PhilipAD:
Thanks @msureshkumar88 — the `_loggable_uri` finding is a real leak and the repro is exact. All three blocking points are addressed, plus the two nits.

**Blocking 1 — credential fragment in `_loggable_uri`.** Fixed, and worth flagging that the suggested `urlsplit`/`urlunsplit` rebuild does not on its own close it. For your input, `urlsplit` yields `netloc='user:pa'`, `query='ss@example.com/x'` — there is no `@` left in the authority, so stripping userinfo is a no-op and `pa` still survives:

```python
>>> urlunsplit((*urlsplit("https://user:pa?ss@example.com/x")[:3], "", ""))
'https://user:pa'
```

The ordering dependency you identified is the root cause, but the deeper problem is that an unencoded delimiter inside userinfo moves the credential tail into a component that gets dropped, so the redactor never sees the `@` that would tell it the authority was untrustworthy. The rule is now: parse once, strip userinfo from the netloc, and if an `@` shows up in a component being discarded — or the URI is not hierarchical and has no authority at all (`mailto:user@host`) — nothing preceding it can be shown to be credential-free, so the authority is redacted rather than shown. A URI that `urlsplit` rejects outright is reported by length only. Truncation happens last, on already-redacted output, so no ordering remains to get wrong.

```
'https://user:pa?ss@example.com/x'          -> 'https://***'
'https://user:pa#ss@example.com/x'          -> 'https://***'
'https://user:s3cret@example.com/w?token=t' -> 'https://example.com/w'
'mailto:user@example.com'                   -> 'mailto:example.com'
'ui://widgets/customer-search'              -> 'ui://widgets/customer-search'
'http://[::1-...'                           -> '<unparseable 18-character URI>'
```

`test_merge_mcp_protocol_meta_redacts_credentials_split_by_a_stray_delimiter` is parametrized over the `?`, `#` and non-hierarchical variants and asserts that neither `user` nor `pa` reaches the log record — `pa` being precisely the fragment that survives if the ordering is wrong, so it fails if the old implementation is restored.

**Blocking 2 — no test through the ingest path.** Added `test_validate_tools_preserves_deprecated_flat_mcp_apps_meta` in `test_gateway_service.py`, which drives a legacy-only tool dict through `GatewayService._validate_tools` and asserts it survives `ToolCreate` validation with the UI association intact. Added `test_validate_tools_keeps_tool_when_deprecated_flat_meta_is_unusable` alongside it, since the fail-open behaviour for a malformed legacy value is only really meaningful at this level — it is the layer where the alternative would have been a `ValidationError` dropping the tool from sync.

**Blocking 3 — resource and resource-template paths.** Fair point that the PR changes shared behaviour and only tested one of seven call sites. `test_deprecated_flat_meta_honoured_on_resource_ingest_paths` mocks only the MCP transport so the real `connect_to_sse_server` chain runs, and asserts a legacy-only resource *and* a legacy-only resource template both come out of `ResourceCreate` validation with their UI association. One gotcha for anyone writing a similar test later: the ingest path gates on `capabilities.get("resources")` being truthy, so an empty `{}` silently skips both branches and the test passes vacuously.

**Docs.** The Metadata Model section now has a "Deprecated Flat `ui/resourceUri` Key" subsection covering the fallback, the presence-not-truthiness precedence rule, why only well-formed `ui://` values are folded in, and the ingest-only scope.

**`removeprefix`.** Done.

**Legacy key retained in `_meta`.** Deliberate, and I agree it is worth stating rather than leaving implicit. `merge_mcp_protocol_meta` copies rather than mutates specifically so callers holding a shared `_meta` are unaffected, and stripping the key would make the function's effect on its input depend on which shape arrived. Downstream consumers read `extensionMetadata`, which is canonical-only. I have noted the ingest-only scope in the docs so the asymmetry is at least discoverable.

Rebased onto current `main`. Targeted MCP Apps suites **115 passed** (111 before), `test_gateway_service.py` **355 passed**, pylint unchanged at 9.98/10 on `mcp_apps.py` (the one finding is a pre-existing module-level singleton name), ruff findings diffed against unmodified `main` at 43 before and 43 after — zero introduced — and `pre-commit run --files` clean across all four touched files, detect-secrets included.

msureshkumar88 (approved):
Ran a real end-to-end verification of this rather than trusting the existing (SSE-client-mocked) unit suite alone — a genuine upstream MCP server process, bridged onto real network transports, federated through the actual `GatewayService.register_gateway` code path, against a real Alembic-migrated SQLite DB. No part of the MCP protocol exchange or the gateway ingestion pipeline is mocked.

## Setup

- Real `mcp.server.lowlevel.Server` stdio process (`upstream_server.py`) exposing 4 tools + 1 resource + 1 resource template covering every shape this fix touches: legacy-only, malformed-legacy, both-shapes-disagreeing (nested must win), and a no-`_meta` control.
- Bridged via the repo's own `mcpgateway.translate --stdio ... --expose-sse` and `--expose-streamable-http`, so both `connect_to_sse_server` and `connect_to_streamablehttp_server` (the two independent call sites doing this) are exercised for real, not just one.
- Fresh temp SQLite DB per run, migrated with a real `alembic -c mcpgateway/alembic.ini upgrade head`.
- `GatewayService.register_gateway()` called directly against that DB and the live bridged endpoint; results asserted by querying the DB afterward.
- Ran the identical harness against a `main` worktree (commit `921c7e907`) as a before/after control, to isolate exactly what this diff changes rather than just asserting the current behavior in isolation.

## Commands

```
.venv/bin/python -m mcpgateway.translate --stdio ".venv/bin/python e2e_pr5764/upstream_server.py" --expose-sse --port <free-port>
.venv/bin/python -m mcpgateway.translate --stdio ".venv/bin/python e2e_pr5764/upstream_server.py" --expose-streamable-http --port <free-port>
DATABASE_URL=sqlite:///<tmp>/pr5764_e2e.db .venv/bin/python -m alembic -c mcpgateway/alembic.ini upgrade head
.venv/bin/python e2e_pr5764/run_e2e.py --transport SSE --label pr_head
.venv/bin/python e2e_pr5764/run_e2e.py --transport STREAMABLEHTTP --label pr_head
E2E_MCPGATEWAY_ROOT=/tmp/pr5764-main-baseline .venv/bin/python e2e_pr5764/run_e2e.py --transport SSE --label main_baseline
E2E_MCPGATEWAY_ROOT=/tmp/pr5764-main-baseline .venv/bin/python e2e_pr5764/run_e2e.py --transport STREAMABLEHTTP --label main_baseline
.venv/bin/python -m pytest tests/unit/mcpgateway/services/test_mcp_apps.py tests/unit/mcpgateway/services/test_gateway_service.py tests/unit/mcpgateway/services/test_resource_service.py
uv tool run ruff@latest check mcpgateway/services/mcp_apps.py
```

## Output

**This branch (`256212b94`), both transports: 8/8 PASS**
- `legacy_only_tool` → `extensionMetadata` correctly gains `{"io.modelcontextprotocol/ui": {"resourceUri": "ui://widgets/customer-search"}}`
- `legacy_invalid_tool` → stays in the DB (not dropped from sync), `extensionMetadata` is `None`
- `nested_wins_tool` → resolves to the nested value (`ui://widgets/new`), disagreeing legacy value discarded
- `control_tool` → unaffected, `extensionMetadata` is `None`
- `customer_data` resource and `customer_record` resource template → both gain their legacy-only `resourceUri`, confirming the fix isn't tool-only

**`main` baseline, both transports: 5/8 PASS** — the 3 failures are exactly `legacy_only_tool` / `customer_data` / `customer_record`, all `extensionMetadata`/`resourceUri` = `None`. That's issue #5762 reproduced on `main` and absent on this branch, with the 3 control fixtures behaving identically on both, which isolates precisely what this diff changes.

**Regression sweep**: `test_mcp_apps.py` + `test_gateway_service.py` + `test_resource_service.py` → 722 passed, 1 skipped (pre-existing, unrelated `"Yet to implement"`). `ruff check mcpgateway/services/mcp_apps.py` → clean.

One scope note for transparency: `merge_mcp_protocol_meta` has call sites in 4 functions, not the 3 I'd assumed going in — the OAuth `authorization_code` resync path (`fetch_tools_after_oauth` / `_connect_to_sse_server_without_validation`) is a genuinely separate call site I didn't drive through a live OAuth flow, since standing up a mock IdP for a full authorization_code exchange felt disproportionate to verify code that's byte-for-byte the same `_validate_tools`/`merge_mcp_protocol_meta` calls already covered above. That path is still covered by the existing unit suite (which does mock the SSE client), just not by this real-process harness.

No blocking findings. Approving.

vishu-bh (approved):

Issue #4970, opened 2026-05-29
[FEATURE]: Implement MCP Apps AppBridge same-server authorization path


vishu-bh:
Parent epic: #2527

## User Story
As an MCP App developer, I want AppBridge calls to tools/resources to be authorized through the same server and session context so that apps cannot cross boundaries or bypass RBAC.

## One-Week Deliverable
Implement the server-side AppBridge authorization/routing path for app-originated JSON-RPC calls under generic extension governance.

## Acceptance Criteria
- UI view sessions bind app calls to a specific server, UI resource, caller, token scope, team scope, and MCP session.
- AppBridge `tools/call` and `resources/read` enforce same-server constraints.
- RBAC, token scoping, extension enabled state, server visibility, and tool/resource visibility are checked before execution.
- App-originated calls run through plugin, sanitizer, audit, and observability hooks.
- Cross-server app calls are denied and audited.
- Expired, missing, forged, disabled-extension, wrong-team, and insufficient-permission UI sessions are denied.
- Tests cover allowed same-server calls, cross-server deny, expired session deny, disabled extension deny, wrong-team deny, and insufficient-permission deny.

## Demo
From a simulated app session, call an allowed same-server tool successfully, then attempt a cross-server call and show it denied with an audit event.

jonpspri:
## Review: AppBridge same-server authorization path

Most of the acceptance criteria in this issue are implemented:

- ✅ Sessions bind app calls to server, resource, caller, token scope, team scope, and MCP session
- ✅ `tools/call` and `resources/read` enforce same-server constraints
- ✅ RBAC, token scoping, extension enabled state, and visibility checks are all applied before execution
- ✅ Plugin and sanitizer hooks are threaded through both AppBridge handlers
- ✅ Expired, missing, wrong-team, disabled-extension, and insufficient-permission sessions are denied
- ✅ Test coverage exists for all of the above deny paths

**Missing**: The acceptance criteria state cross-server app calls should be "denied *and audited*". The deny is in place (line ~10723 of `main.py`), but there is no `audit_trail.log_action` call for the cross-server deny case. Adding a single structured audit entry there (e.g. `action="appbridge_cross_server_denied"`) would satisfy the criterion fully.

PR #5765, opened 2026-07-22
feat(mcp-apps): support standard MCP messages over the AppBridge

### LINKED ISSUE #5763: [FEATURE]: Support standard MCP messages (resources/read, notifications/message, ping) over the AppBridge
### 🧭 Type of Feature

- [x] Security / governance hardening
- [ ] Other

Raised at the request of @vishu-bh in https://github.com/IBM/mcp-context-forge/issues/2527#issuecomment-5044901298, linked to #4970 for tracking as asked.

### 🧭 Epic

**Title:** Support the standard MCP messages an app may send over the AppBridge

**Goal:** Let an MCP App use the four standard MCP messages the specification permits from an app context, while keeping every one of them behind the session binding, server scoping, and RBAC the AppBridge already enforces for `tools/call`.

**Why now:** #5079 shipped the AppBridge with a single supported method. The practical gap is `resources/read` — an app currently cannot read any resource, including its own companion data resources.

---

### 🐞 Current behaviour

`main.py:10593` rejects every method other than `tools/call`:

```python
if method != "tools/call":
    return {"jsonrpc": "2.0", "error": {"code": -32601, "message": f"Method not found: {method}"}, "id": req_id}
```

and the registry allowlist is `frozenset({"tools/call"})` (`mcp_method_registry.py:56`).

This is a *declared* limitation rather than a silent one — `mcp_apps.py:73` advertises `bridge: {methods: ["tools/call"]}` — so a client can detect it. It is still a gap worth closing.

### 📐 What the specification permits

Spec §"Standard MCP Messages" (`specification/2026-01-26/apps.mdx:489-508`) lists exactly four:

| Method | Spec wording |
| --- | --- |
| `tools/call` | Execute a tool on the MCP server |
| `resources/read` | Read resource content |
| `notifications/message` | Log messages to host |
| `ping` | Connection health check |

Two points worth being precise about, because they shape the design:

**Narrowing is discretionary, not mandated.** The per-server scoping language at `apps.mdx:399`, `:402` and `:1661` is written **exclusively about tools**. There is no stated MUST confining `resources/read` to the same server, and no `ui://`-scheme restriction on it. The licence to restrict comes from `apps.mdx:487`: *"The Host MAY forward any message from the View … it MAY decide to block some messages or subject them to further user approval."* Combined with the one-way ratchet at `apps.mdx:286` (*"Host MAY further restrict but MUST NOT allow undeclared domains"*), a gateway narrowing this is conformant; widening would not be.

**Partial support is legal by construction.** The spec advertises host support through `HostCapabilities` (`apps.mdx:622-666`), where `serverTools`, `serverResources`, `logging` and `openLinks` are each optional. A host that omits `serverResources` is declaring it does not proxy resource reads. ContextForge's `bridge.methods` is not itself a spec mechanism, but it serves the same purpose from the server side and should stay accurate.

### 🧑🏻‍💻 User Story 1

**As a:** MCP App author
**I want:** to read my app's companion data resources through the bridge
**So that:** the view can render data without smuggling it through tool results

#### ✅ Acceptance Criteria

```gherkin
Scenario: App reads a resource within its bound server
  Given an AppBridge session bound to server "server-a"
  When the app calls resources/read for a resource on "server-a"
  Then the gateway validates session ownership, server binding, and token scoping
  And returns the resource contents

Scenario: App cannot read across servers
  Given an AppBridge session bound to server "server-a"
  When the app calls resources/read with serverId "server-b"
  Then the request is denied
  And no resource lookup is performed
```

### 🧑🏻‍💻 User Story 2

**As a:** Security Administrator
**I want:** the bridge allowlist to stay explicit
**So that:** widening it can never happen as a side effect of a method being core MCP

#### ✅ Acceptance Criteria

```gherkin
Scenario: Core MCP methods are not implicitly bridge methods
  Given MCP Apps are enabled
  When an app calls tools/list, resources/list, prompts/list or resources/subscribe
  Then the gateway returns method not found
  And no upstream call is made
```

This is the sharpest security point in the change. `resources/read`, `ping` and `notifications/message` are all already in `_CORE_MCP_METHODS` (`mcp_method_registry.py:22,26,37`), so `is_known_method()` returns `True` for them regardless of the MCP Apps flag. Gating the bridge on `is_known_method()` would silently expose the **entire core MCP surface** to apps. The allowlist needs its own predicate.

### 🧑🏻‍💻 User Story 3

**As a:** Gateway operator
**I want:** app log notifications recorded rather than forwarded
**So that:** view-originated text is auditable without reaching upstream servers

#### ✅ Acceptance Criteria

```gherkin
Scenario: Log notification terminates at the gateway
  Given an AppBridge session
  When the app sends notifications/message
  Then the gateway records it for observability
  And does not proxy it to the upstream server
```

The lifecycle diagram at `apps.mdx:1353-1355` is explicit that this terminates at the host — `H ->> H: Record log for debugging/telemetry`, a self-arrow, in contrast with the resource-read branch two lines below which is `H ->> S`.

### 📐 Design Sketch

```mermaid
flowchart TD
    A[App JSON-RPC] --> B{is_app_bridge_method?}
    B -->|no| E[-32601 Method not found]
    B -->|yes| C[MCP session ownership]
    C --> D[Server binding check]
    D --> F{method}
    F -->|tools/call| G[invoke_tool require_app_visible]
    F -->|resources/read| H[read_resource scoped to bound server]
    F -->|notifications/message| I[record only, no upstream call]
    F -->|ping| J[answer locally]
```

Ordering matters: the method gate runs first, then session and server-binding validation, then dispatch — so every method inherits the scoping `tools/call` already had, rather than each handler re-implementing it.

### 🔗 MCP Standards Check

- [x] Adheres to the current MCP Apps specification
- [ ] Deviations, described below

Two deliberate deviations:

1. **Error codes.** The spec never defines a code for an unsupported method — its only code is `-32000 // Implementation-defined error`. Proposing `-32601` for methods outside the allowlist (JSON-RPC 2.0 base semantics, and what the endpoint already returns) rather than claiming spec conformance for it.
2. **Not implementing everything the SDK forwards.** The reference `app-bridge.ts:1797-1800` also proxies `resources/list`, `resources/templates/list` and `prompts/list`, none of which appear in the spec's normative list. Those are listing surfaces with real leak potential through a gateway, so this issue proposes only the four spec-listed messages. Flagging it because an app built against the SDK may expect them; happy to follow up separately if maintainers want them.

### 🔄 Alternatives Considered

**Gate the bridge on `mcp_method_registry.is_known_method()`.** Rejected — as above, it would expose all core MCP methods to apps.

**Confine `resources/read` to the session's own `resource_uri`.** Rejected as too tight: an app legitimately reads companion data resources, and it would not mirror how `tools/call` allows any app-visible tool on the bound server rather than one specific tool. Server binding plus existing RBAC and token scoping is the same authority the originating user already has. Flagging this as the main design question for reviewers — it is easy to tighten if maintainers prefer.

**Split into four separate issues.** Rejected per the maintainer's request for two issues and PRs; the four messages share one gate and one set of scoping checks, so splitting would fragment the security-relevant part.

### 📓 Additional Context

Parent epic: #2527. Tracked with #4970 (AppBridge same-server authorization). Companion issue for the other compatibility gap: #5762.

PhilipAD:
# ✨ Feature / Enhancement PR

## 🔗 Epic / Issue

Closes #5763. Tracked with #4970. Part of #2527.

---

## 🚀 Summary

The AppBridge accepted only `tools/call`. This routes the remaining three standard MCP messages an app may send — `resources/read`, `notifications/message` and `ping` — behind the session binding and server scoping `tools/call` already had.

---

## 📓 Design

Spec §"Standard MCP Messages" (`apps.mdx:489-508`) permits exactly four methods from an app context. The practical gap was `resources/read`: an app could not read any resource, including its own companion data resources.

**Ordering.** The method gate runs first, then MCP session ownership and server-binding validation, then per-method dispatch. Every bridge method therefore inherits the scoping `tools/call` had, rather than each handler re-implementing it.

```mermaid
flowchart TD
    A[App JSON-RPC] --> B{is_app_bridge_method?}
    B -->|no| E[-32601 Method not found]
    B -->|yes| C[MCP session ownership]
    C --> D[Server binding check]
    D --> F{method}
    F -->|tools/call| G[invoke_tool require_app_visible]
    F -->|resources/read| H[read_resource scoped to bound server]
    F -->|notifications/message| I[record only, no upstream call]
    F -->|ping| J[answer locally]
```

**The allowlist needs its own predicate — this is the security-critical part of the change.** `resources/read`, `ping` and `notifications/message` are all already in `_CORE_MCP_METHODS` (`mcp_method_registry.py:22,26,37`), so the existing `is_known_method()` returns `True` for them regardless of the MCP Apps feature flag. Reusing it as the bridge gate would have silently exposed the **entire core MCP surface** to apps. This adds `is_app_bridge_method()`, which consults only the MCP Apps allowlist and the feature flag. `test_core_methods_are_not_implicitly_app_bridge_methods` pins that distinction.

**`resources/read`** runs a method-specific `resources.read` permission check via `_ensure_rpc_permission` before dispatch, so the route effectively requires both the endpoint-level `tools.execute` and the method-level `resources.read`. RBAC and token/team scoping are separate layers: the endpoint decorator alone would let a caller who retains `tools.execute` keep reading after `resources.read` is revoked. It then reuses the identity and team scoping captured on the session (`None if token_teams is None else app_session.user_email`, matching the `tools/call` path) and is confined to `app_session.server_id`. The `x-context-forge-gateway-id` header is stripped exactly as it is for tool calls, so an app cannot force direct-proxy routing. Error mapping matches the existing RPC path: `-32002` not found, `-32000` resource error, `-32602` bad params, plugin codes preserved.

**`notifications/message`** is recorded and never proxied upstream, and is acknowledged at the transport level only — `202 Accepted` with an empty body, not a JSON-RPC result. JSON-RPC decides notification vs request on the *presence* of the `id` member rather than its value, so an `id`-bearing `notifications/message` (including `"id": null`) is a request and is rejected with `-32600` rather than answered with an empty `202` the caller cannot correlate. The lifecycle diagram at `apps.mdx:1353-1355` shows this terminating at the host — `H ->> H: Record log`, a self-arrow, versus `H ->> S` for the resource-read branch two lines below. The payload is view-originated and attacker-influenceable, so it is logged as truncated data (`%.500r`) rather than interpolated into the format string.

**`ping`** is answered locally without an upstream call.

The advertised `bridge.methods` capability and the router now share one constant (`MCP_APPS_BRIDGE_METHODS`) so they cannot drift apart.

Internally, the `tools/call` body lives in `_handle_app_bridge_tools_call`, mirroring `_handle_app_bridge_resources_read`; the endpoint validates session ownership and server binding, then delegates. No behaviour change — it keeps the endpoint's return count under ruff's `PLR0911` limit.

### Deviations worth flagging

**Error codes are a reasoned choice, not a conformance claim.** The spec never defines a code for an unsupported method — its only code is `-32000 // Implementation-defined error`. `-32601` is JSON-RPC 2.0 base semantics and what the endpoint already returned.

**Narrowing is discretionary here, not mandated.** The per-server scoping language at `apps.mdx:399`, `:402` and `:1661` is written *exclusively about tools* — there is no stated MUST confining `resources/read` to the same server, and no `ui://`-scheme restriction on it. The licence to restrict is `apps.mdx:487` (*"it MAY decide to block some messages"*), reinforced by the one-way ratchet at `apps.mdx:286`. So this scoping is a deliberate gateway policy choice; I did not want to overstate it as spec-required.

**Not everything the SDK forwards.** The reference `app-bridge.ts:1797-1800` also proxies `resources/list`, `resources/templates/list` and `prompts/list`, none of which are in the spec's normative list. Those are listing surfaces with real leak potential through a gateway, so they are excluded. Flagging it because an SDK-built app may expect them — happy to follow up separately.

### The main open question for reviewers

`resources/read` is scoped to the **bound server**, not to the session's own `resource_uri`. Rationale: an app legitimately reads companion data resources, and this mirrors how `tools/call` allows any app-visible tool on the bound server rather than one specific tool. The authority granted is the same the originating user already has, via existing RBAC and token scoping. If you'd prefer it tightened to the session's own resource, that's a small change and I'll make it.

---

## 📏 Reviewability

On the `triage` checkbox: #5763 currently carries no labels. Filing it and this PR was explicitly requested by @vishu-bh in https://github.com/IBM/mcp-context-forge/issues/2527#issuecomment-5044901298, which CONTRIBUTING lists as one of the conditions for starting work. If you apply `triage` while scoping it, treat this PR as parked until you're ready.

**One behaviour change to an existing test.** `test_rpc_rejects_disabled_malformed_and_missing_session` asserted `resources/read` → `-32601`, encoding the very limitation this PR removes. It now asserts the same for `prompts/list` — a core MCP method deliberately *not* on the bridge allowlist, which is a stronger check than the original.

---

## 🧪 Checks

Added 13 tests: 9 endpoint tests in `test_mcp_apps_security.py` (ping answered without service calls; log notification recorded and not proxied; the `id`-bearing notification form rejected, parametrized over null/string/numeric ids; `resources/read` scoping incl. header stripping and stored token teams; cross-server read denied; an RBAC deny-path test driving the real route through `TestClient` with `tools.execute` granted and `resources.read` denied; error mapping for missing URI, not-found, resource error, plugin violation, plugin error and unexpected exception) and 2 registry tests in `test_mcp_method_registry.py` (feature-flag gating; core methods not implicitly bridge methods).

- [x] `pytest tests/unit/mcpgateway/services/test_mcp_apps.py tests/unit/mcpgateway/test_mcp_apps_security.py tests/unit/mcpgateway/test_mcp_apps_protocol.py tests/unit/mcpgateway/test_mcp_method_registry.py` — **126 passed** (113 before this change)
- [x] `pytest tests/unit/mcpgateway/test_main.py` — **375 passed**
- [x] `ruff check mcpgateway/` — All checks passed
- [x] `pylint --rcfile=.pylintrc mcpgateway/main.py` — **10.00/10**
- [x] `black --check --line-length 200` — unchanged
- [x] Regression sweep across `test_main_rate_limit.py`, `test_streamablehttp_transport.py`, `test_resource_service.py` — **44 pre-existing failures before and after, zero introduced** (diffed the failure lists against unmodified `main`; they are environment-related in my local venv)
- [x] Docs updated — `docs/docs/architecture/mcp-apps.md` gains a supported-methods table, the corrected capability payload, the `resources.read` requirement, and the notification acknowledgement and `id`-form rules
- [ ] CHANGELOG not updated — happy to add an entry if you'd like one for this

Not run locally: `make docker docker-run-ssl` and the postgres/redis matrix. There is no database, migration, or transport surface in this change, but I can produce that evidence if you want it before merge.

---

## 📐 MCP Compliance

The advertised `bridge.methods` capability grows, so clients that detect support see the new methods rather than having to probe for them.

On the spec checkbox: it was previously ticked while `notifications/message` still returned a JSON-RPC result for the `id`-bearing form. That gap is now closed — notifications are acknowledged at the transport level and the request form is rejected with `-32600` — so the claim matches the implemented behaviour.

### DIFF
--- a/docs/docs/architecture/mcp-apps.md
+++ b/docs/docs/architecture/mcp-apps.md
@@ -340,6 +340,13 @@ transport level with `202 Accepted` and an empty body, matching the Streamable
 HTTP rule for notification-only input. The other three methods are requests and
 return a normal JSON-RPC result or error with the request `id` echoed back.
 
+JSON-RPC distinguishes a notification from a request by the *presence* of the
+`id` member, not by its value, so `"id": null` still makes the message a request
+that requires a response. Because `notifications/message` is defined as
+notification-only, an `id`-bearing form is rejected with `-32600 Invalid Request`
+rather than acknowledged with an empty `202` that would leave the caller waiting
+on a response it will never receive.
+
 Being a core MCP method does not make a method reachable over AppBridge: the
 allowlist is explicit, so `tools/list`, `resources/list`, `prompts/list`, and
 `resources/subscribe` remain rejected.
--- a/docs/docs/architecture/mcp-apps.md
+++ b/docs/docs/architecture/mcp-apps.md
@@ -47,7 +47,7 @@ For authenticated MCP initialization requests, the gateway advertises:
           "schemes": ["ui://"]
         },
         "bridge": {
-          "methods": ["tools/call"]
+          "methods": ["tools/call", "resources/read", "notifications/message", "ping"]
         }
       }
     }
@@ -313,6 +313,23 @@ Content-Type: application/json
 The RPC endpoint requires `tools.execute`. The tool must resolve within the
 stored server binding and must be app-visible through
 `audience: ["app"]` or `audience: ["model", "app"]`.
+
+### Supported AppBridge Methods
+
+The RPC endpoint accepts the standard MCP messages an app may send, and rejects
+everything else with `-32601`. Session ownership and the stored server binding
+are validated before any method is dispatched.
+
+| Method | Behaviour |
+| --- | --- |
+| `tools/call` | Invokes an app-visible tool within the bound server. |
+| `resources/read` | Reads a resource within the bound server, using the identity and team scoping stored on the session. |
+| `notifications/message` | Recorded by the gateway for observability and never proxied upstream. |
+| `ping` | Answered by the gateway without an upstream call. |
+
+Being a core MCP method does not make a method reachable over AppBridge: the
+allowlist is explicit, so `tools/list`, `resources/list`, `prompts/list`, and
+`resources/subscribe` remain rejected.
 
 ## Security Invariants
 
--- a/docs/docs/architecture/mcp-apps.md
+++ b/docs/docs/architecture/mcp-apps.md
@@ -30,8 +30,10 @@ When enabled, the gateway can:
   server, and UI resource;
 - allow AppBridge RPC calls only to tools marked for app use.
 
-The current AppBridge RPC surface supports `tools/call`. Other MCP methods are
-rejected with JSON-RPC `Method not found`.
+The current AppBridge RPC surface supports `tools/call`, `resources/read`,
+`notifications/message`, and `ping` — see
+[Supported AppBridge Methods](#supported-appbridge-methods). Every other MCP
+method is rejected with JSON-RPC `Method not found`.
 
 ## MCP Capability
 
@@ -314,6 +316,11 @@ The RPC endpoint requires `tools.execute`. The tool must resolve within the
 stored server binding and must be app-visible through
 `audience: ["app"]` or `audience: ["model", "app"]`.
 
+RBAC is enforced per method rather than inherited from the endpoint, so
+`resources/read` additionally requires `resources.read`. A caller who keeps
+`tools.execute` after `resources.read` is revoked cannot keep reading resources
+through an existing AppBridge session.
+
 ### Supported AppBridge Methods
 
 The RPC endpoint accepts the standard MCP messages an app may send, and rejects
@@ -323,9 +330,15 @@ are validated before any method is dispatched.
 | Method | Behaviour |
 | --- | --- |
 | `tools/call` | Invokes an app-visible tool within the bound server. |
-| `resources/read` | Reads a resource within the bound server, using the identity and team scoping stored on the session. |
+| `resources/read` | Reads a resource within the bound server, using the identity and team scoping stored on the session. Requires `resources.read`. |
 | `notifications/message` | Recorded by the gateway for observability and never proxied upstream. |
 | `ping` | Answered by the gateway without an upstream call. |
+
+`notifications/message` is a JSON-RPC notification: it carries no `id` and must
+not receive a JSON-RPC response. The gateway therefore acknowledges it at the
+transport level with `202 Accepted` and an empty body, matching the Streamable
+HTTP rule for notification-only input. The other three methods are requests and
+return a normal JSON-RPC result or error with the request `id` echoed back.
 
 Being a core MCP method does not make a method reachable over AppBridge: the
 allowlist is explicit, so `tools/list`, `resources/list`, `prompts/list`, and
--- a/mcpgateway/main.py
+++ b/mcpgateway/main.py
@@ -10607,12 +10607,29 @@ async def create_mcp_app_session(request: Request, db: Session = Depends(get_db)
     )
 
 
+def _app_bridge_request_headers(request: Request) -> Dict[str, str]:
+    """Build the header passthrough for an AppBridge call.
+
+    The gateway routing header is stripped so an app cannot redirect its own
+    call to a different gateway than the one its session is bound to.
+
+    Args:
+        request: Incoming AppBridge RPC request.
+
+    Returns:
+        Lowercased request headers without the gateway routing header.
+    """
+    request_headers = {k.lower(): v for k, v in request.headers.items()}
+    request_headers.pop("x-context-forge-gateway-id", None)
+    return request_headers
+
+
 def _record_app_bridge_log(app_session, params: Dict[str, Any]) -> None:
     """Record a log notification sent by an MCP App.
 
     The MCP Apps lifecycle terminates ``notifications/message`` at the host, so
     the payload is recorded for observability and never proxied upstream. The
-    contents are attacker-influenced, so the message is truncated and logged as
+    contents are attacker-influenced, so every field is truncated and logged as
     data rather than interpolated into the format string.
 
     Args:
@@ -10623,7 +10640,7 @@ def _record_app_bridge_log(app_session, params: Dict[str, Any]) -> None:
     origin_logger = params.get("logger")
     data = params.get("data")
     logger.info(
-        "AppBridge log notification session=%s server=%s level=%r logger=%r data=%.500r",
+        "AppBridge log notification session=%s server=%s level=%.100r logger=%.100r data=%.500r",
         app_session.id,
         app_session.server_id,
         level,
@@ -10655,8 +10672,7 @@ async def _handle_app_bridge_resources_read(db: Session, request: Request, app_s
 
     token_teams = app_session.token_teams
     resource_user_email = None if token_teams is None else app_session.user_email
-    request_headers = {k.lower(): v for k, v in request.headers.items()}
-    request_headers.pop("x-context-forge-gateway-id", None)
+    request_headers = _app_bridge_request_headers(request)
 
     try:
         result = await resource_service.read_resource(
@@ -10717,8 +10733,7 @@ async def _handle_app_bridge_tools_call(db: Session, request: Request, app_sessi
     try:
         token_teams = app_session.token_teams
         tool_user_email = None if token_teams is None else app_session.user_email
-        request_headers = {k.lower(): v for k, v in request.headers.items()}
-        request_headers.pop("x-context-forge-gateway-id", None)
+        request_headers = _app_bridge_request_headers(request)
         result = await tool_service.invoke_tool(
             db=db,
             name=name,
--- a/mcpgateway/main.py
+++ b/mcpgateway/main.py
@@ -10693,6 +10693,72 @@ async def _handle_app_bridge_resources_read(db: Session, request: Request, app_s
     return {"jsonrpc": "2.0", "result": {"contents": [serialize_resource_content_for_mcp(result, fallback_uri=uri)]}, "id": req_id}
 
 
+async def _handle_app_bridge_tools_call(db: Session, request: Request, app_session, params: Dict[str, Any], req_id: Any, requester_email: Optional[str]) -> Dict[str, Any]:
+    """Invoke an app-visible tool on behalf of an MCP App through its bound session.
+
+    The call is scoped to the server the AppBridge session is bound to and uses the
+    identity and team scoping captured when the session was created.
+
+    Args:
+        db: Database session.
+        request: Incoming request, used for plugin context and header passthrough.
+        app_session: The validated AppBridge session.
+        params: JSON-RPC params carrying the tool ``name`` and ``arguments``.
+        req_id: JSON-RPC request id echoed back to the caller.
+        requester_email: Email of the authenticated caller driving the app.
+
+    Returns:
+        A JSON-RPC response dictionary with the tool result or an error.
+    """
+    name = params.get("name")
+    if not name:
+        return {"jsonrpc": "2.0", "error": {"code": -32602, "message": "Missing tool name in parameters"}, "id": req_id}
+
+    try:
+        token_teams = app_session.token_teams
+        tool_user_email = None if token_teams is None else app_session.user_email
+        request_headers = {k.lower(): v for k, v in request.headers.items()}
+        request_headers.pop("x-context-forge-gateway-id", None)
+        result = await tool_service.invoke_tool(
+            db=db,
+            name=name,
+            arguments=params.get("arguments", {}),
+            request_headers=request_headers,
+            app_user_email=requester_email,
+            user_email=tool_user_email,
+            token_teams=token_teams,
+            server_id=app_session.server_id,
+            plugin_context_table=getattr(request.state, "plugin_context_table", None),
+            plugin_global_context=getattr(request.state, "plugin_global_context", None),
+            meta_data=params.get("_meta"),
+            require_app_visible=True,
+        )
+        if hasattr(result, "model_dump"):
+            result = result.model_dump(by_alias=True, exclude_none=True)
+        return {"jsonrpc": "2.0", "result": result, "id": req_id}
+    except ToolNotFoundError:
+        return {"jsonrpc": "2.0", "error": {"code": -32601, "message": f"Tool not found: {name}"}, "id": req_id}
+    except ToolError as exc:
+        logger.info("AppBridge tool call failed with tool error for %s: %s", name, exc)
+        return {"jsonrpc": "2.0", "error": {"code": -32000, "message": str(exc)}, "id": req_id}
+    except PluginViolationError as exc:
+        error_code = -32602
+        if exc.violation and hasattr(exc.violation, "mcp_error_code") and isinstance(exc.violation.mcp_error_code, int):
+            error_code = exc.violation.mcp_error_code
+        return {"jsonrpc": "2.0", "error": {"code": error_code, "message": str(exc)}, "id": req_id}
+    except PluginError as exc:
+        error_code = -32603
+        if exc.error and hasattr(exc.error, "mcp_error_code") and isinstance(exc.error.mcp_error_code, int):
+            error_code = exc.error.mcp_error_code
+        return {"jsonrpc": "2.0", "error": {"code": error_code, "message": str(exc)}, "id": req_id}
+    except (ValidationError, ValueError) as exc:
+        logger.info("AppBridge tool call received invalid parameters for %s: %s", name, exc)
+        return {"jsonrpc": "2.0", "error": {"code": -32602, "message": str(exc)}, "id": req_id}
+    except Exception:
+        logger.exception("AppBridge tool call failed for %s", name)
+        return {"jsonrpc": "2.0", "error": {"code": -32603, "message": "Internal error"}, "id": req_id}
+
+
 @utility_router.post("/appbridge/sessions/{app_session_id}/rpc")
 @require_permission("tools.execute")
 async def handle_mcp_app_session_rpc(app_session_id: str, request: Request, db: Session = Depends(get_db), user=Depends(get_current_user_with_permissions)):
@@ -10744,58 +10810,22 @@ async def handle_mcp_app_session_rpc(app_session_id: str, request: Request, db: 
 
     if method == "notifications/message":
         _record_app_bridge_log(app_session, params)
-        return {"jsonrpc": "2.0", "result": {}, "id": req_id}
+        # A JSON-RPC notification carries no id and MUST NOT receive a JSON-RPC response, so
+        # this is acknowledged at the transport level only: 202 Accepted with an empty body,
+        # matching the Streamable HTTP rule for notification-only input.
+        return Response(status_code=status.HTTP_202_ACCEPTED)
 
     if method == "resources/read":
+        # RBAC is a layer of its own, separate from the session's token/team scoping: the
+        # endpoint decorator only proves tools.execute, which a caller can retain after
+        # resources.read has been revoked. Authorize the read per method.
+        try:
+            await _ensure_rpc_permission(user, db, "resources.read", method, request=request)
+        except JSONRPCError as exc:
+            return {"jsonrpc": "2.0", "error": exc.to_dict()["error"], "id": req_id}
         return await _handle_app_bridge_resources_read(db, request, app_session, params, req_id)
 
-    name = params.get("name")
-    if not name:
-        return {"jsonrpc": "2.0", "error": {"code": -32602, "message": "Missing tool name in parameters"}, "id": req_id}
-
-    try:
-        token_teams = app_session.token_teams
-        tool_user_email = None if token_teams is None else app_session.user_email
-        request_headers = {k.lower(): v for k, v in request.headers.items()}
-        request_headers.pop("x-context-forge-gateway-id", None)
-        result = await tool_service.invoke_tool(
-            db=db,
-            name=name,
-            arguments=params.get("arguments", {}),
-            request_headers=request_headers,
-            app_user_email=requester_email,
-            user_email=tool_user_email,
-            token_teams=token_teams,
-            server_id=app_session.server_id,
-            plugin_context_table=getattr(request.state, "plugin_context_table", None),
-            plugin_global_context=getattr(request.state, "plugin_global_context", None),
-            meta_data=params.get("_meta"),
-            require_app_visible=True,
-        )
-        if hasattr(result, "model_dump"):
-            result = result.model_dump(by_alias=True, exclude_none=True)
-        return {"jsonrpc": "2.0", "result": result, "id": req_id}
-    except ToolNotFoundError:
-        return {"jsonrpc": "2.0", "error": {"code": -32601, "message": f"Tool not found: {name}"}, "id": req_id}
-    except ToolError as exc:
-        logger.info("AppBridge tool call failed with tool error for %s: %s", name, exc)
-        return {"jsonrpc": "2.0", "error": {"code": -32000, "message": str(exc)}, "id": req_id}
-    except PluginViolationError as exc:
-        error_code = -32602
-        if exc.violation and hasattr(exc.violation, "mcp_error_code") and isinstance(exc.violation.mcp_error_code, int):
-            error_code = exc.violation.mcp_error_code
-        return {"jsonrpc": "2.0", "error": {"code": error_code, "message": str(exc)}, "id": req_id}
-    except PluginError as exc:
-        error_code = -32603
-        if exc.error and hasattr(exc.error, "mcp_error_code") and isinstance(exc.error.mcp_error_code, int):
-            error_code = exc.error.mcp_error_code
-        return {"jsonrpc": "2.0", "error": {"code": error_code, "message": str(exc)}, "id": req_id}
-    except (ValidationError, ValueError) as exc:
-        logger.info("AppBridge tool call received invalid parameters for %s: %s", name, exc)
-        return {"jsonrpc": "2.0", "error": {"code": -32602, "message": str(exc)}, "id": req_id}
-    except Exception:
-        logger.exception("AppBridge tool call failed for %s", name)
-        return {"jsonrpc": "2.0", "error": {"code": -32603, "message": "Internal error"}, "id": req_id}
+    return await _handle_app_bridge_tools_call(db, request, app_session, params, req_id, requester_email)
 
 
 @utility_router.post("/_internal/mcp/tools/call/")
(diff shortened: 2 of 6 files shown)


vishu-bh (changes requested):
Thanks for expanding AppBridge support and clearly documenting the security decisions. The explicit feature-gated allowlist, stored server binding, removal of the spoofable gateway-routing header, and cross-server denial tests all look well considered.

Before merging, could you please check the following items?

## Blocking: method-specific RBAC for `resources/read`

The AppBridge RPC endpoint is currently guarded only by:

```python
@require_permission("tools.execute")
```

This was appropriate when the endpoint supported only `tools/call`, but it now also dispatches `resources/read`. That branch calls `resource_service.read_resource()` without independently enforcing `resources.read`.

ContextForge treats these as separate security layers:

- Token/team scoping determines which resources the caller can see.
- RBAC determines whether the caller may perform `resources.read`.

Service-level visibility checks therefore do not replace the missing RBAC check. This matters especially for custom roles and permission revocation: a caller retaining `tools.execute` could continue reading through an existing AppBridge session after `resources.read` is removed.

Could you add a method-specific permission check before dispatching `resources/read`, using the same `_ensure_rpc_permission(..., "resources.read", ...)` pattern used by normal MCP RPC routing?

Please also add an endpoint-level deny-path regression test covering:

- `tools.execute` granted
- `resources.read` denied
- Valid AppBridge and MCP sessions
- `resources/read` rejected
- `resource_service.read_resource()` not called

Current tests call `handle_mcp_app_session_rpc.__wrapped__`, which bypasses the route decorator and cannot detect this permission mismatch.

## Blocking: Pylint regression

`make pylint` currently fails with:

```text
mcpgateway/main.py:11745:12: W0621: Redefining name 'mcp_method_registry' from outer scope
mcpgateway/main.py:11745:12: W0404: Reimport 'mcp_method_registry'
```

The PR adds a module-level `mcp_method_registry` import, but an existing branch in `_handle_rpc_authenticated()` still imports it locally.

Could you remove the local import and use the module-level binding?

## Please confirm notification response semantics

The `notifications/message` branch records the notification correctly and does not proxy it upstream, which matches the intended lifecycle. However, it currently returns:

```python
{"jsonrpc": "2.0", "result": {}, "id": req_id}
```

For a JSON-RPC notification, `id` is absent and the receiver normally must not emit a JSON-RPC response. The new test currently expects a result, so it may be pinning non-compliant behavior.

Could you verify the expected AppBridge/HTTP behavior against the applicable MCP Apps and JSON-RPC rules?

- If this is a true notification, please avoid returning a JSON-RPC result and update the test accordingly.
- If the Host expects an HTTP acknowledgement despite JSON-RPC notification semantics, please document that deliberate transport-level behavior.

## Minor: documentation inconsistency

The beginning of `docs/docs/architecture/mcp-apps.md` still says the current AppBridge surface supports only `tools/call`, while the new capability payload and supported-method table advertise four methods.

Could you update the earlier statement so the document is internally consistent?

## Validation results

Validation against the PR SHA produced:

- Targeted MCP Apps tests: `95 passed`
- Ruff: passed
- Bandit: passed with no findings
- Interrogate: passed at 100%
- Package verification: passed
- Pylint: failed because of the duplicate import described above

Once the method-specific RBAC and Pylint issues are resolved, the implementation should be in much better shape for merge.

PhilipAD:
Thanks — the RBAC gap is a fair catch, and the framing of the two layers as independent is the part I got wrong: I treated the session's stored token/team scoping as covering the read, when it only decides *visibility*.

**Blocking 1 — method-specific RBAC.** `resources/read` now runs the same check normal MCP RPC routing applies, before dispatch:

```python
if method == "resources/read":
    try:
        await _ensure_rpc_permission(user, db, "resources.read", method, request=request)
    except JSONRPCError as exc:
        return {"jsonrpc": "2.0", "error": exc.to_dict()["error"], "id": req_id}
    return await _handle_app_bridge_resources_read(db, request, app_session, params, req_id)
```

The `JSONRPCError` is converted to the endpoint's JSON-RPC error shape rather than propagating, so a denial looks like every other AppBridge denial (`-32003`) instead of an unhandled exception. This also picks up the token-scope cap in layer 1 of `_ensure_rpc_permission`, not just the RBAC role check.

`test_rpc_resources_read_denied_without_resources_read_permission` is the deny-path regression test, and it drives the real route through `TestClient` rather than `__wrapped__`, exactly as you asked: `mcpgateway.middleware.rbac.PermissionService` grants only `tools.execute` so the route decorator genuinely passes, `PermissionChecker` denies `resources.read`, both the AppBridge and MCP sessions are valid, and the assertions are `200` + `-32003` + `read_resource` never awaited. Confirmed it fails if the new check is removed.

**Blocking 2 — pylint.** The local import in `_handle_rpc_authenticated()` is gone; that branch uses the module-level binding. `pylint mcpgateway/main.py` is back to 10.00/10.

**Notification semantics.** You are right that the previous shape pinned non-compliant behaviour. `notifications/message` no longer returns a JSON-RPC result — it is acknowledged at the transport level only, `202 Accepted` with an empty body, which is the Streamable HTTP rule for notification-only input and matches what this gateway's own `/_internal/mcp/notifications/message` endpoint does (204 there; 202 is the closer fit for a JSON-RPC POST surface). The test now asserts `status_code == 202` and an empty body, and the deliberate transport-level behaviour is written up in the architecture doc alongside the method table. `ping` keeps returning a result — it is a request with an id, so that one is correct as-is.

**Docs.** The opening summary now lists all four methods and links to the method table, and the table records the extra `resources.read` requirement.

One incidental change worth flagging: adding the permission branch pushed the endpoint to 17 return statements and tripped ruff's `PLR0911`, so the `tools/call` body moved into `_handle_app_bridge_tools_call`, mirroring `_handle_app_bridge_resources_read`. No behaviour change — the endpoint now validates session ownership and server binding, then delegates.

Rebased onto `3fe3035`. Targeted MCP Apps suites 96 passed, `tests/unit/mcpgateway/test_main.py` 372 passed, pylint 10.00/10, and `pre-commit run --files` clean across the three touched files.

vishu-bh (changes requested):
Thank you for addressing the earlier review feedback. The latest revision is much stronger, and the following items have been resolved well:

- `resources/read` now performs a method-specific `resources.read` authorization check.
- A real FastAPI deny-path test has been added for insufficient resource permissions.
- Successful `notifications/message` calls now return HTTP `202` with an empty response body.
- The duplicate import and related lint issue have been resolved.
- The documentation has been updated to describe the new bridge methods and permission behavior.
- The `tools/call` logic has been extracted into `_handle_app_bridge_tools_call`, improving the endpoint’s readability and complexity.
- The security review did not identify any new security vulnerabilities or authorization bypasses.
- Validation is in good shape: the targeted tests passed (`96 passed`), the full suite passed (`20,869 passed`, `802 skipped`, `2 xfailed`), and Ruff, Interrogate, Pylint, Bandit, Verify, and diff coverage checks passed.

The PR description also matches the overall implementation: the explicit method allowlist, session/server scoping, cross-server rejection, resource confinement, sensitive-header stripping, local ping handling, and local notification handling are accurately represented.

There is one remaining functional issue that would be good to address before merging:

### Remaining functional finding

The `notifications/message` test currently sends:

```json
{
  "jsonrpc": "2.0",
  "id": null,
  "method": "notifications/message"
}
```

Under JSON-RPC, a notification must omit the `id` member entirely. A message containing an `id` member is a request, even when its value is `null`, and therefore requires a JSON-RPC response.

The current handler treats every `notifications/message` message as a notification and always returns an empty HTTP `202`. This could silently discard a response expected by a client that sends an ID-bearing request.

Could you please consider:

1. Updating the successful notification test so that the `id` member is absent.
2. Adding a regression test for an ID-bearing `notifications/message` request.
3. Either rejecting the ID-bearing form with an appropriate JSON-RPC error or returning a valid JSON-RPC response instead of an empty `202`.

Until this is resolved, the PR description’s “Matches current MCP spec” statement appears a little stronger than the implemented behavior.

### PR description updates

A few small updates would keep the description aligned with the latest revision:

- Mention that `resources/read` now performs a method-specific `resources.read` permission check. The current route effectively requires both the endpoint-level `tools.execute` permission and the method-level `resources.read` permission.
- Mention that a successfully handled `notifications/message` notification returns HTTP `202` with an empty body.
- Update the targeted test result from `95 passed` to `96 passed` and mention the added RBAC deny-path coverage.
- Optionally mention the `_handle_app_bridge_tools_call` extraction as an internal maintainability refactor.
- Please reconsider marking “Matches current MCP spec” as complete until the notification ID behavior above is corrected.

The implementation otherwise looks substantially improved, and the previous security and authorization concerns have been addressed. Once the notification ID handling and PR description are updated, this should be in good shape to merge.

PhilipAD:
Thanks @vishu-bh — you are right, and my previous change only fixed half of it. I corrected the response shape for the notification case but left the handler deciding "is this a notification?" purely from the method name, so an `id`-bearing message took the same 202 path. That is the silent-discard you describe.

**Notification vs request is now decided by the `id` member.** JSON-RPC keys off its *presence*, not its value, so `"id": null` is a request. The handler now checks `"id" in body` rather than `body.get("id")`, which is the distinction that matters here — `.get()` cannot tell an absent member from an explicit null.

I took your third option, rejecting the request form, rather than answering it:

```python
if method == "notifications/message":
    if "id" in body:
        return {
            "jsonrpc": "2.0",
            "error": {"code": -32600, "message": "notifications/message must be sent as a JSON-RPC notification, without an 'id' member"},
            "id": req_id,
        }
    _record_app_bridge_log(app_session, params)
    return Response(status_code=status.HTTP_202_ACCEPTED)
```

Reasoning for rejecting over responding: the MCP Apps lifecycle defines this message as terminating at the host with no reply, so there is no result to return that would mean anything — a `{"result": {}}` would just be inventing a response for a message the spec says is not answered. `-32600 Invalid Request` says what is actually wrong (the message is malformed for this method) and is the code this file already uses for JSON-RPC envelope violations. The caller gets a correlatable answer either way, which was the substance of your concern. The notification is not recorded in the reject path, since the message is rejected as malformed rather than half-processed.

**Tests.** `test_rpc_log_notification_is_recorded_and_not_proxied` now omits the `id` member entirely, as you asked — it previously sent `"id": None`, which was pinning the wrong thing. `test_rpc_log_notification_with_an_id_is_rejected_as_a_request` is the new regression test, parametrized over `None`, `"1"` and `7` so the null case is covered explicitly alongside ordinary ids; it asserts `-32600`, the echoed `id`, and that neither `invoke_tool` nor `read_resource` is awaited.

**Docs.** The supported-methods section now states the presence-not-value rule and why the `id`-bearing form is rejected rather than acknowledged.

**PR description.** Updated on all your points: the method-level `resources.read` check and the fact that the route effectively requires both it and the endpoint-level `tools.execute`; the `202` + empty body acknowledgement; the test count corrected to 99 with the RBAC deny-path coverage described; and the `_handle_app_bridge_tools_call` extraction noted as an internal maintainability change. I have left "Matches current MCP spec" ticked, but only now that the `id` handling is corrected — with a note recording that it was previously ticked while that gap was open. Say the word if you would rather it stay unticked until you have re-verified.

Rebased onto current `main`. Targeted MCP Apps suites **99 passed** (96 before), `tests/unit/mcpgateway/test_main.py` **372 passed**, pylint 10.00/10 on `main.py`, and `pre-commit run --files` clean across the three touched files.

vishu-bh (dismissed):
Code looks clean and address the issue correctly. 

LGTM 👍

PhilipAD:
Rebased onto `main` @ `a27a3f70a` to clear the merge conflict that appeared once #5764 merged.

The conflict was confined to the constants block at the top of `mcpgateway/services/mcp_apps.py` — #5764 added the legacy-`ui/resourceUri` constants and this branch added `MCP_APPS_BRIDGE_METHODS` in the same place. Both sides are additive, so the resolution keeps them both; no logic changed. The `docs/docs/architecture/mcp-apps.md` overlap merged cleanly (separate sections).

Also re-checked the auth-context middleware extraction (#6008) that landed on `main` in the meantime — it doesn't touch the RPC handler paths this PR changes. `test_mcp_apps_security.py`, `test_mcp_method_registry.py`, `services/test_mcp_apps.py`, `services/test_gateway_service.py` and `middleware/test_admin_csrf_binding.py` all pass locally, and ruff/pylint are unchanged from the baseline on `main`.

The force-push dismissed the approval — sorry for the extra round, @vishu-bh, whenever you get a chance.

vishu-bh (dismissed):
Thanks for the rebase @PhilipAD 

LGTM 🚀

msureshkumar88 (changes requested):
Thanks for the thorough writeup and for addressing the earlier RBAC/notification-semantics rounds so cleanly — the allowlist-vs-`is_known_method()` reasoning and the server-scoped `resources/read` design are well argued, and the test suite for the new dispatch branches is solid.

Ran the targeted suite (130 passed) plus `ruff check` locally — both clean. A coverage pass turned up one real gap plus two small hardening notes, none of which should be blocking on their own, but flagging as requested changes so they don't get lost before merge:

**Suggestion — untested plugin-error branches in `resources/read`**
`_handle_app_bridge_resources_read`'s `PluginViolationError` / `PluginError` handling (`mcpgateway/main.py:10676-10685`) has no test coverage. The identical mapping logic for `tools/call` (`main.py:10744-10753`) is covered by `test_rpc_serializes_model_dump_and_maps_tool_errors`, but the copy for `resources/read` doesn't have an equivalent in `test_rpc_resources_read_maps_missing_uri_and_lookup_failures`. Could you add the two cases there (same pattern, `resource_service.read_resource` as the mock target instead of `tool_service.invoke_tool`)?

**Minor — duplicated header-strip logic**
The `x-context-forge-gateway-id`-stripping block:
```python
request_headers = {k.lower(): v for k, v in request.headers.items()}
request_headers.pop("x-context-forge-gateway-id", None)
```
appears identically in both `_handle_app_bridge_resources_read` and `_handle_app_bridge_tools_call`. Worth pulling into a small shared helper for a single source of truth on this security-relevant header strip.

**Minor — unbounded `level`/`logger` in the notification log line**
In `_record_app_bridge_log` (`main.py:10625-10632`), `data` is truncated with `%.500r` but `level` and `logger` are logged via plain `%r` with no cap. Both are view-originated per an authenticated AppBridge session, so the risk is low, but capping them the same way (e.g. `%.100r`) would close off log-volume padding as a matter of defense-in-depth.

None of these affect correctness or introduce a new vulnerability — happy to re-review as soon as they're in.

PhilipAD:
Thanks @msureshkumar88 — all three are addressed in `ddff74fe1`, rebased onto `db8acbc6c`.

**Plugin-error branches in `resources/read`** — added the `PluginViolationError` / `PluginError` cases to `test_rpc_resources_read_maps_missing_uri_and_lookup_failures`, mirroring the `tools/call` pattern with `resource_service.read_resource` as the mock target: a `PluginViolation(mcp_error_code=-32042)` and a `PluginErrorModel(mcp_error_code=-32043)`, each asserted to surface as the violation's own code rather than the `-32602` / `-32603` fallbacks.

**Duplicated header strip** — extracted into `_app_bridge_request_headers(request)`, now called by both `_handle_app_bridge_resources_read` and `_handle_app_bridge_tools_call`. Worth stating why the strip matters, since the helper is now the single place it happens: `x-context-forge-gateway-id` is the direct-proxy routing header, so leaving it in the passthrough would let an app aim its own call at a gateway other than the one its session is bound to. The existing `test_rpc_invokes_bound_app_visible_tool_without_direct_proxy_header` and the `resources/read` scoping test still assert the header is absent from what each service receives, so both call sites stay covered through the helper.

**Unbounded `level` / `logger`** — both now use `%.100r`, alongside the existing `%.500r` on `data`. Agreed the risk is low given the session is authenticated, but it costs nothing and the docstring already claimed the payload was truncated, so this makes that true of the whole line.

Verification on the updated branch:

- `pytest tests/unit/mcpgateway/services/test_mcp_apps.py tests/unit/mcpgateway/test_mcp_apps_security.py tests/unit/mcpgateway/test_mcp_apps_protocol.py tests/unit/mcpgateway/test_mcp_method_registry.py` — **126 passed**
- `pytest tests/unit/mcpgateway/test_main.py` — **375 passed**
- `ruff check` — clean; `pylint --rcfile=.pylintrc mcpgateway/main.py` — **10.00/10**; `black --check` — unchanged

PR description updated with the new counts and the added error-mapping cases. Ready for re-review whenever you have a moment.

msureshkumar88 (approved):
Ran a real end-to-end verification of this PR against a live, subprocess-hosted gateway (real uvicorn process, real Alembic-migrated SQLite DB via the app's own bootstrap lifespan, real HTTP round-trips, real RBAC role assignment) — no mocking anywhere in the harness. Script: `e2e_pr5765/run_e2e.py` (added in this run, not part of the PR diff).

## Config used

Each run launches its own isolated gateway process against a fresh temp SQLite DB:

```
DATABASE_URL=sqlite:///<fresh temp file>
AUTH_REQUIRED=true
MCPGATEWAY_UI_ENABLED=true
EMAIL_AUTH_ENABLED=true
MCPGATEWAY_MCP_APPS_ENABLED=true
JWT_SECRET_KEY=<random per run>
AUTH_ENCRYPTION_SECRET=<random per run>
```

A trivial local `http.server` instance stands in as the REST tool backend for `tools/call` (the only test double in the harness).

## Commands run

```bash
# PR head (this branch, ddff74fe1)
.venv/bin/python e2e_pr5765/run_e2e.py --mode pr_head --label pr_head

# main baseline, via a detached worktree at origin/main (65c6b5252)
E2E_MCPGATEWAY_ROOT=/tmp/pr5765-main-baseline \
    .venv/bin/python e2e_pr5765/run_e2e.py --mode baseline_gate --label main_baseline
```

## Output

**`main` baseline (issue #5763 confirmed real pre-fix) — 5/5 assertions passed:**
```
[PASS] method=ping rejected at the allowlist gate with -32601
[PASS] method=resources/read rejected at the allowlist gate with -32601
[PASS] method=notifications/message rejected at the allowlist gate with -32601
[PASS] tools/call control: passes the allowlist gate (error != -32601)
```
All three methods are rejected with `-32601` *before* any session is even consulted — confirming the pre-fix gate was the blunt `method != "tools/call"` check the issue describes.

**PR head (`ddff74fe1`) — 40/40 assertions passed**, covering:
- Allowlist widened to exactly the four spec methods; core-but-not-bridged methods (`resources/list`, `prompts/list`, `tools/list`, `resources/subscribe`) still rejected with `-32601`.
- `initialize` advertises all four methods in `capabilities.extensions["io.modelcontextprotocol/ui"].bridge.methods`.
- `resources/read` over a real AppBridge session returns the real, previously-registered resource content.
- Cross-server denial: a resource read scoped to a different `serverId` than the session's binding is denied (`-32003`), no leak.
- `notifications/message` without an `id` gets a bare `202` with an empty body; the same method **with** an `id` (tested `null`, string, and numeric) is rejected with `-32600` instead of silently discarding a response the caller is waiting on.
- `ping` answered locally.
- `tools/call` (pre-existing, unchanged path) still reaches a real REST backend end-to-end — the `_handle_app_bridge_tools_call` extraction didn't regress it.
- **Live RBAC layering**, the scenario called out in the design doc and the earlier review round: a user holding a role with `tools.execute` + `resources.read` reads a resource over the bridge; an admin then swaps their role for `tools.execute`-only *while the AppBridge session stays open* (no restart, no new session) — the very next `resources/read` on that same session is denied with `-32003`, while `ping` and `tools/call` keep working. Confirms the two RBAC layers are genuinely independent at runtime, not just in unit-test mocks.

One non-obvious wrinkle surfaced while building the harness, noted here in case it's useful context rather than as a PR finding: every non-admin user is auto-assigned the `platform_viewer` global role at creation (`settings.default_user_role`), and RBAC composes as a union across all active role assignments. Revoking a custom role alone isn't enough to remove `resources.read` from a user — `platform_viewer` has to be revoked too. That's existing, intentional platform behavior unrelated to this PR; flagging only because it tripped up the first version of the harness.

Also reran the targeted unit suite (130 passed), ruff, and bandit against `ddff74fe1` — all clean, consistent with the PR's own reported numbers.

## Conclusion

No changes required. The e2e run independently confirms the fix closes #5763 as described, the security layering (allowlist gate, session/server scoping, method-specific RBAC, cross-server denial, JSON-RPC notification semantics) holds under real HTTP traffic, and the pre-existing `tools/call` path is unaffected. Approving.

PR #5774, opened 2026-07-23
fix(ui-rewrite): return to originating page after session-expiry re-login

### LINKED ISSUE #5552: [UI-REWRITE]: After session-expiry re-login, user lands on home page instead of the page they were on
### 🐞 Bug Summary

  When a user's session expires, they are redirected to the login page. After they re-authenticate, they are always sent to the home page (`/app/`) instead of being returned to the page they were on. For example, a session that expires while on `/app/tools` should return the user to `/app/tools` after login — not `/app/`.

  ---

  ### 🧩 Affected Component

  - [x] `mcpgateway` - UI (admin panel) — new UI-rewrite client (`client/`)

  ---

  ### 🔁 Steps to Reproduce

  1. Log in to the Admin UI and navigate to a sub-page, e.g. `/app/tools`.
  2. Let the session expire (or invalidate/clear the auth cookie) so the next request returns `401`.
  3. Trigger any authenticated request on that page — you are redirected to `/app/login`.
  4. Log in again.
  5. **Observed:** you land on `/app/` (home).
  6. **Expected:** you return to `/app/tools`.

  ---

  ### 🤔 Expected Behavior

  After re-authenticating following a session expiry (or being blocked on a protected route), the user should be returned to the originating page (`pathname + query`), falling back to `/app/` only when there is no valid prior location.

  ---

  ### 🧠 Additional Context

  The originating path is discarded at three redirect sites in the UI-rewrite client, none of which carry a return path:

  - `client/src/api/client.ts` — the `401` handler calls `window.location.replace("/app/login")`.
  - `client/src/router/index.tsx` (`AuthGuard`) — calls `navigate("/app/login")`.
  - `client/src/pages/Login.tsx` — always `navigate("/app/")` after a successful login.

  A fix should capture the current location on redirect (e.g. a validated `?next=` param), and Login should honor it after auth. Any return path must be re-validated through the router's existing `validateDestination()` guard to prevent open-redirect. Because the `401` path does a full-document `window.location.replace`, the return path must survive a reload (URL query param or `sessionStorage`, not React state).

Altamimi-Dev:
# Pull Request

## 🔗 Related Issue

Closes #5552

## 📝 Summary

After a session expires, the user is sent to login and — previously — always landed on `/app/` afterward, losing the page they were on. This preserves the originating `pathname + query` across the redirect and returns the user there after re-authenticating.

- Adds `buildLoginPath()` and `resolveNextParam()` to the router. Both run the return path through the existing `validateDestination()` guard, so a crafted `?next=` (protocol, `//host`, non-`/app` path) can't drive an open redirect and falls back to `/app/`.
- `AuthGuard` now redirects with `buildLoginPath(path)` instead of a bare `/app/login`.
- The `401` handler in `api/client.ts` appends the current page as `?next=`. Because that path does a full-document `window.location.replace`, the return path rides in the URL so it survives the reload. Built inline there to keep the API layer free of router imports.
- `Login` reads and honors the validated `next` after a successful login (and when already authenticated), falling back to `/app/`.

## 📏 Reviewability

## 🏷️ Type of Change

- [x] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor

## 🧪 Verification

Run from `client/`:

| Check | Command | Status |
|-------|---------|--------|
| Lint (changed files) | `npx eslint <changed files>` | ✅ clean |
| Format (changed files) | `npx prettier --check <changed files>` | ✅ clean |
| Tests | `npx vitest run src/router/index.test.tsx src/pages/Login.test.tsx src/api/client.test.ts` | ✅ 86 passed |

New tests cover: `buildLoginPath`/`resolveNextParam` (encode, query preservation, login self-nesting, open-redirect rejection), the `401` handler capturing the current `/app` page as `next`, and `Login` returning the user to a validated `next`.

## ✅ Checklist

- [x] Code formatted (Prettier + ESLint)


marekdano (commented):
@Altamimi-Dev - thanks for your contribution!!!

**Verdict:** Solid, security-conscious, well-tested. No blocking issues - a couple of minor suggestions and small test additions.

 ## Suggestions

  - **`resolveNextParam` doesn't guard against `next=/app/login` — asymmetric with `buildLoginPath`.** `resolveNextParam("?next=%2Fapp%2Flogin")` returns `/app/login` (it passes `validateDestination`). In `Login.tsx`, once authenticated the effect fires `navigate(returnTo)` → `/app/login`, whose search is still `?next=/app/login`, so `returnTo` stays `/app/login` — a potential redirect loop / stuck-on-login state. Only reachable via a manually crafted URL (neither `buildLoginPath` nor the 401 handler emits it), so low severity, but for symmetry `resolveNextParam` should collapse a login-pointing `next` back to `/app/`.
  - **Minor logic duplication** between the inline `next`-builder in `api/client.ts` and `router.buildLoginPath`. Justified (no router imports in the API layer), but the two differ subtly: the inline version gates on `current.startsWith("/app/")` (trailing slash), so a bare `/app` location yields no `next`. Acceptable — worth a one-line comment noting the intentional divergence so a future edit keeps them in sync.
  - **`Login.tsx` reads `window.location.search` directly** rather than from the router. Consistent with how `client.ts` reads `window.location`, and fine — flagging only in case the router later exposes the search string as a cleaner source.

### Test coverage — reviewed, with gaps ⚠️
  Coverage is good, but missing:
  - `resolveNextParam` decoding a `next` that carries its own query string (`?next=%2Fapp%2Ftools%3Fpage%3D2`) — only the `buildLoginPath` encode side tests the round-trip.
  - The "already authenticated" `useEffect` branch in `Login.tsx` redirecting to a non-default `returnTo` — only the post-submit path is tested.
  - The login-loop edge case (blocked by the fact the guard doesn't exist yet).

Altamimi-Dev:
@marekdano — thank you for the review, and especially for chasing the `resolveNextParam` asymmetry down to the actual loop. You're right that it's only reachable via a hand-crafted URL, but the fix is small and the asymmetry was the kind that quietly becomes a real bug once another caller starts emitting `next`.

All three points are addressed in `49c6b174c`:

- **`resolveNextParam` now collapses a login-pointing `next` to `/app/`.** It applies the same predicate `buildLoginPath` already used on the write side, so the two are symmetric — one refuses to nest login into its own `next`, the other refuses to read it back out.
- **Added the one-line comment on the intentional divergence** in `api/client.ts`, noting that the inline gate uses a trailing-slash `"/app/"` prefix so a bare `/app` location deliberately yields no `next` param. That was the subtle difference you spotted, and it's now stated where a future edit will see it.
- **`Login.tsx` reading `window.location.search` directly** — left as-is per your note. Agreed it's the consistent choice today; if the router later exposes the search string, that's a clean follow-up across both this and `client.ts`.

All three test gaps are closed:

- `resolveNextParam` decoding a `next` that carries its own query string (`?next=%2Fapp%2Ftools%3Fpage%3D2` → `/app/tools?page=2`).
- The already-authenticated `useEffect` branch in `Login.tsx` redirecting to a non-default `returnTo`.
- The login-loop guard itself, including the nested case (`?next=%2Fapp%2Flogin%3Fnext%3D%252Fapp%252Ftools`). I checked this one is revert-proof — without the guard it returns `/app/login` and the test fails.

Verification on the pushed head: `npx vitest run src/router/index.test.tsx src/pages/Login.test.tsx src/api/client.test.ts` → 89 passed, full client suite 2539 passed, `tsc -b` clean, eslint and prettier clean on the changed files.

marekdano (commented):
Looks good!

LGTM 🚀

marekdano (approved):

PR #5778, opened 2026-07-23
fix(ui): preselect scroll-paginated tools in virtual-server edit modal

### LINKED ISSUE #3358: [BUG][UI]: Virtual Servers — Selected tools not shown/checked in dropdown during Edit

tw-360vier:
## Summary
In the virtual-server **Edit** modal, the associated-tools picker loads its items in pages via HTMX infinite scroll (`per_page=20`; each sentinel uses `hx-swap="outerHTML"`, `hx-trigger="intersect once"`). A server's associated tools that live on page 2+ render **unchecked** even though they are associated, which makes it look like the associations are missing. Page 1 is fine, and Save preserves the real selection (the persistent store is seeded on open and read on submit), so this is a checkbox-rendering correctness issue, not data loss.

## Reproduce
1. Associate more than ~20 tools with a virtual server (so the associations span multiple selector pages).
2. Open the server's Edit modal.
3. Scroll the tools picker down to load later pages — associated tools on page 2+ appear unchecked.

## Root cause
The edit modal seeds the persistent selection store with the associated item ids on open. But `tools.js`'s pill-count `update()` — which runs as each scroll page settles, syncing DOM→store — **deletes** the id of every checkbox currently rendered unchecked. Associated items on page 2+ always render unchecked, so their seeded ids are stripped from the store as they stream in. Restoring checked state from the store alone then finds an empty store and checks nothing. (The store is repaired for Save via other paths, which is why Save itself was unaffected — but the visible checkboxes were wrong.)

## Fix
Restore checked state from a source that survives that churn. `restoreCheckedStateFromStore(containerId)` now (re-)checks a box when its id is in the selection store **or** its identity is in the container's `data-server-*` association attribute (tool *names* for tools, via `data-tool-name`; ids for resources/prompts, via the checkbox value). That attribute is only mutated on an explicit user check/uncheck, never by the paginated re-render, so it is the reliable "was this associated when the modal opened" signal.

- The re-apply is wired on `htmx:afterSettle` (plus `htmx:afterSwap` for search re-renders). Because it also matches the churn-proof `data-server-*` attribute, it restores correctly regardless of its ordering relative to `update()`.
- It only acts on **unchecked** boxes, so a user's explicit uncheck (which removes the item from `data-server-*`) is respected and never forced back on.
- Re-checking dispatches a bubbling `change`, so the store listener re-adds the id — repairing the store that `update()` depleted and keeping the pill count and the save payload correct.

A pre-existing global `htmx:afterSettle` handler in `app.js` already attempts a similar re-check, but it is unreliable in this flow — it gates on a one-shot `data-auto-check` marker that it then consumes, and races the synchronous store sync — so associated tools on later pages still render unchecked. This change adds a dedicated, container-scoped re-apply rather than relying on that path.

Save behavior, selection ordering, and the create-server form are untouched.

## Verification
- New tests in `tests/unit/js/edit-server-scroll-pagination.test.js` simulate the real path (the pill-count store sync that depletes the store before each settle). They cover tools (name match), resources and prompts (id match), multi-page depletion, the user-uncheck regression, and store-only restore — failing without the fix and passing with it.
- Full JS suite: 2294 passed / 3 skipped, no regressions.
- Verified manually against a server with 23 associated tools across 5 scroll pages: before, 23 associated rendered / 0 checked; after, 23/23 checked under both programmatic intersect and real scrolling, with no change to the persisted association count.

Closes #3358

### DIFF
--- a/mcpgateway/admin_ui/servers.js
+++ b/mcpgateway/admin_ui/servers.js
@@ -1216,6 +1216,83 @@ export function resetEditSelections() {
   });
 }
 
+// Maps a selector container id to the data-* attribute that holds its server's
+// currently-associated items, and to how each checkbox in that container is
+// matched against that attribute.  For tools the attribute holds tool *names*
+// (matched against the checkbox's data-tool-name); for resources and prompts it
+// holds ids (matched against the checkbox value).  This is the authoritative,
+// churn-proof source of association: unlike the selection store, it is only
+// mutated by an explicit user check/uncheck, never by the paginated re-render.
+const SERVER_ITEMS_ATTR = {
+  "edit-server-tools": { attr: "data-server-tools", by: "data-tool-name" },
+  "edit-server-resources": { attr: "data-server-resources", by: "value" },
+  "edit-server-prompts": { attr: "data-server-prompts", by: "value" },
+};
+
+/**
+ * Re-apply checked state to all checkboxes currently in the given selector
+ * container.  Called after any DOM change (search result render or HTMX
+ * scroll-paginated append) that may have added new, unchecked checkboxes.
+ *
+ * A checkbox is (re-)checked when its id is in the persistent selection store
+ * OR its identity is in the container's data-server-* association attribute.
+ *
+ * The dual match is what fixes #3358.  As the infinite-scroll pages stream in,
+ * tools.js's pill-count update() syncs DOM->store and *deletes* the id of every
+ * checkbox that is currently rendered unchecked.  A server's associated items
+ * on page 2+ always render unchecked, so their seeded ids are stripped from the
+ * store before any re-check runs — a store-only match then restores nothing and
+ * the associated items appear unchecked despite being associated.  The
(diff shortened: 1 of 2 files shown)


christophercolumbusdog:
+1 on this request; it's very annoying (impossible?) to properly edit the virtual MCP servers without it. You basically have to recreate the server from scratch if you want to make a change.

jonpspri:
Old UI.  Not planned.

PR #5779, opened 2026-07-23
fix: Vite rollup fallback on s390x

gcgoncalves:
Add a vite rollup fallback for s390x arch, as the Vite v8 rollup is incompatible with it.


ja8zyjits (approved):
LGTM

brian-hussey (approved):
LGTM

Issue #5739, opened 2026-07-21
[BUG]: Saving LLM Provider/Model in Admin UI fails with 403 CSRF validation failed


msureshkumar88:
## Describe the bug

Saving an LLM Provider (or Model) from **Settings → LLM Settings** in the Admin UI always fails with `403 CSRF validation failed`. Console shows:

```
Failed to load resource: the server responded with a status of 403 (Forbidden)
Error saving provider: Error: CSRF validation failed
    at saveLLMProvider (bundle-....js:1:...)
```

This is the LLM **Settings** UI (`/llm/providers`, `/llm/models` — served by `llm_config_router`), distinct from the LLM **Chat** connect flow fixed in #5214 / #5223, which only exempts the `/llmchat` prefix.

## Steps to reproduce

1. Default config (`CSRF_ENABLED=true`), log into the Admin UI via `/admin/login`.
2. Go to Settings → LLM Settings → Providers → "Add Provider".
3. Fill the form, click "Save Provider".
4. 403 `CSRF validation failed`.

Same failure on: Delete Provider, Toggle Provider, Health Check, Fetch/Sync Models, Save Model, Delete Model, Toggle Model, and the "Test" tab's `runTest()` call — every state-changing (`POST`/`PATCH`/`DELETE`) call in `mcpgateway/admin_ui/llmModels.js` (11 call sites) is affected.

## Root cause

`mcpgateway/admin_ui/llmModels.js` builds every write request manually:

```js
headers: {
  Authorization: `Bearer ${await getAuthToken()}`,
  "Content-Type": "application/json",
},
```

`getAuthToken()` (`mcpgateway/admin_ui/tokens.js`) reads the `jwt_token` cookie via `document.cookie`, but that cookie is set `httponly=True` (`mcpgateway/utils/security_cookies.py:123`, intentionally — "Prevents JavaScript access"). For a normal session-cookie login, `getAuthToken()` therefore always returns `""`, so the header sent is literally `Authorization: Bearer ` (empty).

In `CSRFMiddleware.dispatch` (`mcpgateway/middleware/csrf_middleware.py`):
- Step 4 skips CSRF entirely for genuine Bearer-token requests — but `_extract_bearer_token("Bearer ")` is falsy, so this skip never triggers for session-cookie users.
- Step 5 then looks for the `X-CSRF-Token` header — which `llmModels.js` never sets — and immediately 403s with `"CSRF validation failed"`.

[...]

PR #5780, opened 2026-07-23
fix(admin-ui): fix CSRF failures on LLM Settings writes

msureshkumar88:
## Summary

Fixes two independent bugs behind #5739 ("Saving LLM Provider/Model in Admin UI fails with 403 CSRF validation failed"):

- `mcpgateway/admin_ui/llmModels.js` built every write request by hand and never attached `X-CSRF-Token`, so all 11 state-changing calls (Save/Delete/Toggle Provider, Health Check, Fetch/Sync Models, Save/Delete/Toggle Model, and the Test tab) 403'd for cookie-authenticated Admin UI sessions. A shared `llmRequestHeaders()` helper now attaches the CSRF token from the `mcpgateway_csrf_token` cookie and only sets `Authorization` when a real bearer token is available (session logins use an httponly cookie, so the old code was sending a meaningless empty `Bearer ` header).
- Separately, the Admin UI login handler bound the CSRF token's HMAC to `EmailUser.id` (the JWT `sub` claim), while `CSRFMiddleware` validates against `EmailUser.email` (`request.state.user.email`). That mismatch meant `/llm/*` writes (guarded by `CSRFMiddleware`) would still fail CSRF validation even with a correct header. Bound `csrf_user_id` to the email instead, matching `routers/auth.py` and `routers/email_auth.py`, which already do this correctly.

Also corrected an inaccurate comment in `auth_middleware.py` that claimed `EmailUser.email` is the primary key (it's `.id`; `.email` is used as the CSRF/logging identity).

Deliberately out of scope: `selectiveImport.js` / `fileTransfer.js` have the same missing-CSRF-header pattern on `/admin/import/*` — left for a follow-up issue since #5739 is specifically the LLM Settings tab.

## Update: end-to-end regression test surfaced three more stacked bugs

Per review feedback, added a real end-to-end test (`/admin/login` → dashboard load → `/llm/providers` write, no mocks). It still 403'd with `CSRF_TOKEN_INVALID` after the fixes above, tracing to three independent, pre-existing bugs that all had to be fixed together for the flow to actually work:

- **`.env.example`** set `CSRF_COOKIE_NAME=csrf_token`, overriding the code default (`mcpgateway_csrf_token`) that `admin.py` and every Admin UI JS file already hardcode. Since `cp .env.example .env` is step one of setup, real deployments never had the cookie name `CSRFMiddleware` was looking for. Fixed the template value.
- **Middleware registration order in `main.py`**: `CSRFMiddleware` was registered after `AuthContextMiddleware`, but Starlette runs middleware in reverse registration order, so `CSRFMiddleware` actually ran *before* `request.state.user` was populated — the opposite of what the surrounding code comments assumed. Reordered so CSRF registers first (runs last/innermost, after auth).
- **`CSRFMiddleware`'s JWT-decode fallback** (used when `request.state.jti` isn't set) unconditionally overwrote `user_id` too, using the JWT `sub` claim (`EmailUser.id`) even when the email had already been correctly resolved from `request.state.user`. Now only fills in whichever of `user_id`/`session_id` is actually missing.

The regression test (`test_admin_login_csrf_token_validates_llm_provider_write`) plus a negative control (`test_admin_login_csrf_token_rejected_without_header`) are in `tests/unit/mcpgateway/middleware/test_admin_csrf_binding.py`.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -86,7 +86,7 @@ CSRF_SECRET_KEY=
 CSRF_TOKEN_NAME=X-CSRF-Token
 
 # Cookie name for CSRF token
-CSRF_COOKIE_NAME=csrf_token
+CSRF_COOKIE_NAME=mcpgateway_csrf_token
 
 # CSRF token expiration time in seconds (default: 3600 = 1 hour)
 CSRF_TOKEN_EXPIRY=3600
--- a/mcpgateway/admin.py
+++ b/mcpgateway/admin.py
@@ -4332,7 +4332,11 @@ async def admin_ui(
 
         # Set HTTP-only cookie using centralized security cookie utility
         set_auth_cookie(response, token, remember_me=False)
-        csrf_user_id = str(payload["sub"])
+        # CSRF tokens are HMAC-bound to the identity CSRFMiddleware derives from
+        # request.state.user, which is EmailUser.email — not EmailUser.id (the PK
+        # used for the JWT `sub` claim). Matches routers/auth.py and
+        # routers/email_auth.py, which already bind to the email.
+        csrf_user_id = admin_email
         csrf_session_id = str(payload["jti"])
         LOGGER.debug(f"Set session JWT token cookie for user: {admin_email}")
     except Exception as e:
--- a/mcpgateway/admin_ui/llmModels.js
+++ b/mcpgateway/admin_ui/llmModels.js
@@ -2,11 +2,34 @@ import { AppState } from "./appState.js";
 import { showCopyableModal } from "./modals.js";
 import { parseErrorResponse } from "./security.js";
 import { getAuthToken } from "./tokens.js";
-import { safeGetElement, showToast } from "./utils.js";
+import { getCookie, safeGetElement, showToast } from "./utils.js";
 
 // ===================================================================
 // LLM SETTINGS FUNCTIONS
 // ===================================================================
+
+/**
+ * Build headers for state-changing LLM settings requests.
+ * - Authorization is only attached when a real bearer token is available
+ *   (session logins use an httponly cookie, so getAuthToken() returns "").
+ * - X-CSRF-Token satisfies both CSRFMiddleware (/llm/*) and
+ *   enforce_admin_csrf (/admin/llm/*), which share the same cookie/header names.
+ */
+async function llmRequestHeaders({ json = true } = {}) {
+  const headers = {};
+  if (json) {
+    headers["Content-Type"] = "application/json";
+  }
+  const token = await getAuthToken();
+  if (token) {
+    headers.Authorization = `Bearer ${token}`;
+  }
+  const csrfToken = getCookie("mcpgateway_csrf_token");
+  if (csrfToken) {
+    headers["X-CSRF-Token"] = csrfToken;
+  }
+  return headers;
+}
 
 /**
  * Switch between LLM Settings tabs (providers/models)
@@ -75,9 +98,7 @@ const loadLLMProviderDefaults = async function () {
     const response = await fetch(
       `${window.ROOT_PATH}/admin/llm/provider-defaults`,
       {
-        headers: {
-          Authorization: `Bearer ${await getAuthToken()}`,
-        },
+        headers: await llmRequestHeaders({ json: false }),
       },
     );
     if (response.ok) {
@@ -184,9 +205,7 @@ const renderProviderSpecificFields = async function (
     const response = await fetch(
       `${window.ROOT_PATH}/admin/llm/provider-configs`,
       {
-        headers: {
-          Authorization: `Bearer ${await getAuthToken()}`,
-        },
+        headers: await llmRequestHeaders({ json: false }),
       },
     );
 
@@ -353,9 +372,7 @@ export const fetchLLMProviderModels = async function (providerId) {
       `${window.ROOT_PATH}/admin/llm/providers/${providerId}/fetch-models`,
       {
         method: "POST",
-        headers: {
-          Authorization: `Bearer ${await getAuthToken()}`,
-        },
+        headers: await llmRequestHeaders({ json: false }),
       },
     );
 
@@ -397,9 +414,7 @@ export const syncLLMProviderModels = async function (providerId) {
       `${window.ROOT_PATH}/admin/llm/providers/${providerId}/sync-models`,
       {
         method: "POST",
-        headers: {
-          Authorization: `Bearer ${await getAuthToken()}`,
-        },
+        headers: await llmRequestHeaders({ json: false }),
       },
     );
 
@@ -437,9 +452,7 @@ export const editLLMProvider = async function (providerId) {
     const response = await fetch(
       `${window.ROOT_PATH}/llm/providers/${providerId}`,
       {
-        headers: {
-          Authorization: `Bearer ${await getAuthToken()}`,
-        },
+        headers: await llmRequestHeaders({ json: false }),
       },
     );
     if (!response.ok) {
@@ -555,10 +568,7 @@ export const saveLLMProvider = async function (event) {
 
     const response = await fetch(url, {
       method,
-      headers: {
-        Authorization: `Bearer ${await getAuthToken()}`,
-        "Content-Type": "application/json",
-      },
+      headers: await llmRequestHeaders(),
       body: JSON.stringify(formData),
     });
 
@@ -601,9 +611,7 @@ export const deleteLLMProvider = async function (providerId, providerName) {
       `${window.ROOT_PATH}/llm/providers/${providerId}`,
       {
         method: "DELETE",
-        headers: {
-          Authorization: `Bearer ${await getAuthToken()}`,
-        },
+        headers: await llmRequestHeaders({ json: false }),
       },
     );
 
@@ -632,9 +640,7 @@ export const toggleLLMProvider = async function (providerId) {
       `${window.ROOT_PATH}/llm/providers/${providerId}/state`,
       {
         method: "POST",
-        headers: {
-          Authorization: `Bearer ${await getAuthToken()}`,
-        },
+        headers: await llmRequestHeaders({ json: false }),
       },
     );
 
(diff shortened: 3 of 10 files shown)


prakhar-singh1928 (dismissed):
LGTM

ja8zyjits:
## ⚠️ Warnings

### Missing E2E Integration Test
- **Gap**: No test proving full flow: Admin UI login → CSRF token generation → LLM Settings write → CSRFMiddleware validation → success
- **Current**: Unit tests prove each piece works in isolation
- **Risk**: Low (unit tests thorough), but E2E would catch integration issues
- **Recommendation**: Add Playwright or FastAPI TestClient test:
  ```python
  async def test_admin_login_csrf_token_validates_llm_settings_write():
      # 1. POST /admin/login with valid credentials
      # 2. Extract jwt_token and mcpgateway_csrf_token cookies
      # 3. POST /llm/providers with X-CSRF-Token header
      # 4. Assert 200 OK (not 403 CSRF_TOKEN_INVALID)
  ```

### Untested Functions in llmModels.js
- **Gap**: 3 of 11 state-changing functions not directly tested:
  - `syncLLMProviderModels()`
  - `fetchModelsForModelModal()`
  - `toggleLLMModel()`
- **Current**: Indirect coverage via shared `llmRequestHeaders()` helper
- **Risk**: Very low (identical code path)
- **Recommendation**: Add spot-check test for one function or document in test header

### selectiveImport.js / fileTransfer.js Gap
- **Gap**: Same missing-CSRF-header pattern on `/admin/import/*` endpoints (out of scope per PR description)
- **Risk**: Medium (same bug exists)
- **Recommendation**: File follow-up issue, add TODO in test file:
  ```javascript
  // TODO(#XXXX): selectiveImport.js and fileTransfer.js have same
  // missing X-CSRF-Token pattern. Add tests when fix lands.
  ```

### Missing Edge Cases (Optional)
- **CSRF token expired**: No test for expiry scenario (low risk, CSRFService handles it)
- **Content-Type omission**: Not systematically tested for non-JSON requests
- **CSRF_ENABLED=false**: No test when CSRF disabled

### Comment Clarity in auth_middleware.py
- **File**: `mcpgateway/middleware/auth_middleware.py:181-183`
- **Issue**: Comment says "email IS the ID" but misleading
- **Context**: `EmailUser.id` is PK (UUID), `.email` is CSRF identity
- **Impact**: Low (code correct, comment could confuse)
- **Fix**: Clarify "email is user identifier for CSRF/logging, not DB primary key"

---

ja8zyjits (changes requested):
check comments

msureshkumar88:
Thanks for the detailed review! Went through each point:

**Fixed:**
- Stale comment in `auth_middleware.py` (line with `# For EmailUser, email IS the ID`) — the block comment above it was already corrected, but this trailing comment was left over from before the fix. Cleaned up.
- Added the E2E integration test you asked for (real `/admin/login` → dashboard load → `/llm/providers` write, no mocks) — `test_admin_login_csrf_token_validates_llm_provider_write` in `test_admin_csrf_binding.py`, plus a negative control confirming CSRF is still enforced without the header.

That E2E test turned out to be genuinely valuable: it initially failed even with this PR's fixes applied, and traced to three more independent, pre-existing bugs (cookie name mismatch between `.env.example` and the code default, a middleware registration-order bug that meant `CSRFMiddleware` ran before request identity was populated, and a fallback path that silently re-resolved identity to the wrong claim). Details and the fix are in the updated PR description above — without those, the original #5739 flow would still 403 in a real deployment even with everything else in this PR correct.

**Not applicable — already covered:**
- The "3 untested functions" (`syncLLMProviderModels`, `fetchModelsForModelModal`, `toggleLLMModel`) — these already have dedicated `describe()` blocks with direct calls in `llmModels.test.js` (lines 413, 1180, 1647 respectively). Might be worth double-checking against the diff you reviewed in case it was an earlier revision.
- `selectiveImport.js` / `fileTransfer.js` gap — already called out as deliberately out of scope in the PR description (separate follow-up issue), not something this PR should absorb.

**Left as-is (optional/low-risk per your own note):**
- CSRF-disabled test case is already covered generically by the existing `test_csrf_disabled_allows_all_requests` in `test_csrf_middleware.py`, just not LLM-route-specific.
- Expired-token / non-JSON content-type edge cases — agreed these are low-value additions given the mechanism is identical to already-tested paths; happy to add if you'd still like them.

Let me know if anything here needs another look.

ja8zyjits:
## 🚨 Blocking Issues

### Documentation

- [ ] **CSRF_COOKIE_NAME default mismatch in docs**
  - **File**: `docs/docs/manage/configuration.md:43`
  - **Issue**: Shows old default `csrf_token`, should be `mcpgateway_csrf_token`
  - **Fix**: Update line 43:
    ```diff
    -CSRF_COOKIE_NAME=csrf_token
    +CSRF_COOKIE_NAME=mcpgateway_csrf_token
    ```
  - **Context**: Code (`config.py:408`) and `.env.example:43` already correct

---

## ⚠️ Warnings

### Edge Cases (Test Gaps)

1. **Middleware ordering regression**
   - **Gap**: No test proves CSRFMiddleware runs AFTER AuthContextMiddleware
   - **Risk**: Future refactor re-introduces ordering bug
   - **Suggested**: `test_csrf_middleware_runs_after_auth_context_middleware()`
   - **File**: `tests/unit/mcpgateway/middleware/test_admin_csrf_binding.py`

2. **`.env.example` cookie name validation**
   - **Gap**: No test validates default cookie name matches code expectation
   - **Risk**: Future edit reverts to wrong name
   - **Suggested**: `test_csrf_cookie_name_default_matches_code_expectation()`
   - **File**: `tests/unit/mcpgateway/test_config.py`

3. **CSRFMiddleware partial fallback**
   - **Gap**: No test for partial-fallback case (`user_id` set, `session_id` missing)
   - **Suggested**: `test_csrf_fallback_preserves_existing_user_id()`
   - **File**: `tests/unit/mcpgateway/middleware/test_admin_csrf_binding.py`

---

## 💡 Suggestions

### Code Quality

1. **Extract CSRF fallback logic** (`csrf_middleware.py:145-158`)
   - Complex nested conditionals (14 lines)
   - Extract to `_resolve_csrf_identity_from_jwt()` helper
   - Benefit: Testable in isolation, clearer intent

2. **Add null check** (`csrf_middleware.py:133`)
   - Current: `hasattr(user, "email")` without checking `user` is not None
   - Fix: Add explicit `if user:` guard
   - Impact: Fragile pattern, safe but unclear

3. **Extract hardcoded cookie name**
   - Files: `llmModels.js:27`, `admin-llm-csrf.test.js:68`, `llmModels.test.js:62`
   - Duplicated: `"mcpgateway_csrf_token"` in 3+ locations
   - Fix: Extract to shared constant

4. **Specific exception handling** (`csrf_middleware.py:157`)
   - Current: `except Exception as exc:` catches all
   - Fix: Catch `JWTError`, `ValidationError` specifically
   - Impact: Broad catch masks bugs

5. **Add assertion messages** (`test_admin_csrf_binding.py:85-87`)
   - Current: `assert response.status_code == 403`
   - Fix: `assert response.status_code == 403, f"Expected 403, got {response.status_code}: {response.body}"`
   - Benefit: Easier debugging

6. **Extract test DOM setup**
   - Files: `admin-llm-csrf.test.js:88-119`, `121-168`
   - Duplicated: 60+ lines similar DOM creation
   - Fix: Shared test helper function

7. **Clarify middleware ordering comment** (`main.py:3387-3399`)
   - Current: Verbose explanation
   - Suggestion: Add explicit "reverse registration order" note

---

msureshkumar88:
Thanks for the follow-up — went through the second round.

**Fixed** (3 new regression tests + 2 assertion-message nits, all in `test_admin_csrf_binding.py` / `test_config.py`):
- `test_csrf_middleware_runs_after_auth_context_middleware` — asserts `AuthContextMiddleware` precedes `CSRFMiddleware` in `app.user_middleware`, guarding the ordering fix directly instead of relying on the `main.py` comment.
- `test_csrf_cookie_name_default_matches_env_example` — mirrors the existing `test_env_example_sunset_date_matches_config_default` pattern in `test_deprecation_middleware.py`; fails if `.env.example`'s `CSRF_COOKIE_NAME` and `config.py`'s `csrf_cookie_name` default ever drift.
- `test_csrf_fallback_preserves_existing_user_id` — the highest-value gap you flagged: both existing unit tests set `request.state.jti`, so the fallback's "only fill the missing piece" branch was never actually exercised. This test sets `user_id` (from `request.state.user.email`) but leaves `jti` unset, and asserts the fallback doesn't clobber `user_id` with the JWT `sub` claim.
- Assertion messages added to the two 403/200 checks in the existing binding tests.

**Checked, not changed — believe these don't apply:**
- `docs/docs/manage/configuration.md:43` — that file has no `CSRF_COOKIE_NAME`/`csrf_token` reference anywhere; line 43 is a table header for `.env.example` dev-override docs. Nothing to fix there, but happy to add a CSRF env var row if you think one's missing.
- `csrf_middleware.py:133` null check — the surrounding `if hasattr(request.state, "user") and request.state.user:` guard on the line above already makes `user` non-None inside the block, so `user.email` can't hit a null `user`.
- Catching `JWTError`/`ValidationError` instead of `Exception` at `csrf_middleware.py:157` — `verify_jwt_token_cached` raises `HTTPException` on failure, not those types, so narrowing the catch would let it propagate as a 500 instead of the intended 403 fallback.
- Cookie-name constant extraction and DOM-setup dedup in the JS tests — `setupProviderFormDOM()`/`setupModelFormDOM()` build distinct field sets so they're not actually duplicated, and the hardcoded cookie name matches the pre-existing pattern already used in `auth.test.js`/`admin-async.test.js`/`test_admin.py` — open to doing this as a follow-up cleanup across all of them rather than diverging just this file.
- `main.py:3387` ordering comment — it already states "runs LAST/innermost due to reverse order", so this looks covered.
- The 3 "untested" JS functions and the `selectiveImport.js`/`fileTransfer.js` gap — as previously noted, these're already covered / intentionally out of scope.

Let me know if you'd like the constant-extraction or JS dedup done as part of this PR rather than a follow-up.

ja8zyjits (approved):
LGTM
Follow up created

Issue #5109, opened 2026-06-08
[UI-REWRITE]: Local Search


gcgoncalves:
Add search functionality to the tables across all entity types.

DEPENDS ON: All other list tasks (see #4096)

Altamimi-Dev:
I'd like to pick this up. Plan is to add a local (client-side) search box to the entity list tables, reusing the expandable search pattern already used in the details drawers, filtering the loaded rows by their name/key fields across the main entity pages. Could you assign it to me?

gcgoncalves:
Setting this task on block. The UI/UX team started a conversation on this and it might be dropped in favour of a higher prominence of the global search.

Does it make any sense to have a local search when the global search is available at all times?

PR #5784, opened 2026-07-23
feat(ui-rewrite): add local search to Users, Teams, and MCP Servers tables

Altamimi-Dev:
# Pull Request

## 🔗 Related Issue

Relates to #5109

---

## 📝 Summary

Adds a local (client-side) search box to the entity list tables so users can quickly filter the rows already loaded on the page.

- New shared `ListSearch` component (`components/ui/list-search.tsx`) — an expandable search input mirroring the affordance already used in the details drawers.
- Wired into the three list-table pages that have real tables today: **Users**, **Teams**, and **MCP Servers**. Each filters its loaded rows by the entity's key text fields (name/email, name/description).
- Search is debounced (~300ms) and shared across the three tables via a reusable `useLocalSearch(items, getText)` hook; matches on the entity's `id` in addition to name/description.
- Filtering is applied over the rows already fetched; the "Load more" control is hidden while a query is active, and a "No matching results." message is shown when nothing matches.

_Note: Tools/Resources/Prompts list pages render grouped card grids (their tables live only in the per-source drawers), and Agents/Tokens/LLM pages are still stubs — so this covers the entity pages that actually render a list table today._

---

## 📏 Reviewability

---

## 🏷️ Type of Change

- [ ] Bug fix
- [x] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

| Check | Command | Status |
|---|---|---|
| Lint | `cd client && npm run lint` | ✅ |
| Format | `cd client && npm run format:check` | ✅ |
| Unit tests | `cd client && npm run test:run` | ✅ 2513 passed |

Manual: on the Users, Teams and MCP Servers pages, typing in the search box filters the loaded rows live; clearing it restores the full list; a non-matching query shows "No matching results.".

---

## ✅ Checklist

- [x] Code formatted

---

## 🎥 Demo

https://github.com/user-attachments/assets/80c6ff17-7d8b-408d-983a-48d40eb089fc

### DIFF
--- a/client/src/pages/Teams.tsx
+++ b/client/src/pages/Teams.tsx
@@ -3,6 +3,7 @@ import { useIntl } from "react-intl";
 import { toast } from "sonner";
 import { Plus } from "lucide-react";
 import { Button } from "@/components/ui/button";
+import { ListSearch } from "@/components/ui/list-search";
 import { TeamsTable } from "@/components/teams/TeamsTable";
 import { TeamForm } from "@/components/teams/TeamForm";
 import { ConfirmDialog } from "@/components/servers/ConfirmDialog";
@@ -27,6 +28,7 @@ export function Teams() {
   const [membersDialogOpen, setMembersDialogOpen] = useState(false);
   const [teamForMembers, setTeamForMembers] = useState<Team | null>(null);
   const [teamToEdit, setTeamToEdit] = useState<Team | null>(null);
+  const [searchQuery, setSearchQuery] = useState("");
 
   const queryPath = useMemo(() => {
     const params = new URLSearchParams();
@@ -47,6 +49,16 @@ export function Teams() {
       setNextCursor(response.nextCursor ?? null);
     }
   }, [response]);
+
+  const filteredTeams = useMemo(() => {
+    const query = searchQuery.trim().toLowerCase();
+    if (!query) return allTeams;
+    return allTeams.filter(
+      (team) =>
+        team.name.toLowerCase().includes(query) ||
+        (team.description ?? "").toLowerCase().includes(query),
+    );
+  }, [allTeams, searchQuery]);
 
   const error = queryError ? queryError.message : null;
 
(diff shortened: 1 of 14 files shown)


client/src/pages/Users.tsx
```diff
@@ -169,6 +171,16 @@ export function Users() {
 
   const error = queryError ? queryError.message : null;
 
+  const filteredUsers = useMemo(() => {
```
gcgoncalves: 1. Asides from memoising, this function could also be debounced. We do not want to flood the api with fast typers. This applies to all other query functions.

2. This search is doing a local query. Ideally, we should query the backend, so that when pagination is introduced, functionality is not broken.

3. Consider moving the query functions into a re-usable hook that receives the query url as a parameter, for reusability.

client/src/pages/Teams.tsx
```diff
@@ -48,6 +50,16 @@ export function Teams() {
     }
   }, [response]);
 
+  const filteredTeams = useMemo(() => {
+    const query = searchQuery.trim().toLowerCase();
+    if (!query) return allTeams;
+    return allTeams.filter(
+      (team) =>
+        team.name.toLowerCase().includes(query) ||
+        (team.description ?? "").toLowerCase().includes(query),
```
gcgoncalves: It would be interesting to add the id to this search, it's a common search case.

gcgoncalves (changes requested):
@Altamimi-Dev Thank you for this PR! The code is well written, and gets the search working. There are a few suggestions I'd like to make before approving it, though.

Also, in the current state, it does not close the issue, but relates to it. I like that the PR is short and partial, it makes it easier to review. But we must add local search to all items before considering the task complete.

Altamimi-Dev:
Thanks for the review, @gcgoncalves — consolidating my notes into one here.

Already pushed (commit `86185f5`):
- Debounce (~300ms) + a shared `useLocalSearch(items, getText)` hook across the three tables
- `id` added to the matched fields
- Switched to `Relates to #5109`

On backend querying: you're right that "Local Search" is per-list search and it should query the server so it survives pagination. I'll apply the filter inside the existing team/visibility scoping so it can't surface out-of-scope rows. Two quick scoping questions before I build it:

1. Backend `?search=` in this PR, or a separate backend PR that this frontend then rebases onto?
2. All entity lists now, or land these three (Users / Teams / MCP Servers) first and track the rest?

gcgoncalves:
Thanks for the update!

1. ~The backnd for the search exists already, you can look at the old HTMX app for reference.~
2. Stick to Users / Teams / MCP Servers for now. 👍 

EDIT: Regarding item 1, I stand corrected. If the api does not support it, keep it local for now. With the consolidation in a hook, it should be easier to refactor later when the backend support is implemented. Sorry for the confusion!

gcgoncalves (approved):

client/src/pages/Users.tsx
Altamimi-Dev: Debounce (~300ms) + a shared useLocalSearch(items, getText) hook across the three tables

client/src/pages/Teams.tsx
Altamimi-Dev: id added to the matched fields

Issue #5215, opened 2026-06-12
LLM Chat cannot connect to a Virtual Server on the same gateway — forwarded session token rejected (401) by /servers/{id}/mcp


dwightford:
## Summary
After getting past CSRF, **LLM Chat → Connect** to a Virtual Server hosted by the **same** gateway fails with "Service initialization failed". The server log shows `httpx.HTTPStatusError: 401 Unauthorized` for `/servers/{id}/mcp`. The gateway cannot connect its own LLM Chat to its own Virtual Servers.

## Environment
- ContextForge **v1.0.3**
- Auth required (default). LLM Chat connecting to a local Virtual Server endpoint (`/servers/{id}/mcp`, `streamable_http`).

## Root cause
`connect()` (`mcpgateway/routers/llmchat_router.py`, the auth-token handling block ~L810) forwards the user's **`jwt_token` session cookie** as `Authorization: Bearer` for the loopback connection to the gateway's own Virtual Server MCP endpoint:

```python
if input_data.server and (input_data.server.auth_token is None or input_data.server.auth_token == ""):
    jwt_token = request.cookies.get("jwt_token")
    ...
    input_data.server.auth_token = jwt_token
```

That endpoint **rejects the session token (401)**, whereas a freshly-minted token is accepted.

## Evidence
- `POST /llmchat/connect` with a `create_jwt_token`-minted token as `server.auth_token` → **200**, returns the server's tool list.
- Same request with the browser **session** token → **401 Unauthorized** from `/servers/{id}/mcp` (surfaced as "Service initialization failed").

## Proposed fix (for discussion)
In `connect()`, mint a fresh service token for the authenticated user for the loopback connection instead of forwarding the session cookie — or have the server MCP endpoint accept valid session tokens. I have a working patch (mint via `create_jwt_token({"sub": user_id, "username": user_id})`, with a fallback to the cookie) and am happy to send a PR with a test.

dwightford:
For transparency: this was root-caused with AI-assisted debugging (Claude) while building on ContextForge v1.0.3. Happy to open a PR with a test for whichever fix approach the maintainers prefer.

bogdanmariusc10:
Thank you for your report, @dwightford ! I've conducted a thorough code review of the authentication flow between LLM Chat and Virtual Server endpoints. 

[...]

Issue #5214, opened 2026-06-12
LLM Chat "Connect" fails with 403 "CSRF validation failed" — /llmchat missing from csrf_exempt_paths (admin UI uses a different CSRF token type)


dwightford:
## Summary
From the bundled Admin UI, **LLM Chat → Connect** always returns `403 {"detail":"CSRF validation failed"}` when global CSRF protection is enabled (the default). The LLM Chat feature is effectively unusable from the Admin UI out of the box.

## Environment
- ContextForge **v1.0.3**
- Default config (`CSRF_ENABLED=true`), Admin UI enabled, cookie-based auth via `/admin/login`.

## Steps to reproduce
1. Default config (global CSRF on). Log into the Admin UI.
2. Go to **LLM Chat**, select a Virtual Server and a configured model.
3. Click **Connect** → `403 CSRF validation failed`.

## Root cause
The Admin UI authenticates via `/admin/login`, which sets the `mcpgateway_csrf_token` cookie to a **random** token (`_set_admin_csrf_cookie`, `mcpgateway/admin.py:1627`) — the per-route "admin CSRF" scheme. But `/llmchat/connect` is guarded by the **global** `CSRFMiddleware`, which requires that same cookie to be a **session-bound HMAC token** (`validate_csrf_token`, `mcpgateway/middleware/csrf_middleware.py`, the HMAC step). The admin token fails HMAC validation → 403.

The Admin UI bundle reads the `mcpgateway_csrf_token` cookie and sends it as `X-CSRF-Token`, but it is the wrong token **type** for the global middleware. Every other Admin-UI-reached path (`/admin`, `/auth/*`, `/oauth/fetch-tools`) is listed in `csrf_exempt_paths` (`mcpgateway/config.py:438`) precisely because those paths use their own per-route CSRF mechanism — `/llmchat` appears to have been omitted from that list.

## Impact
LLM Chat connect is broken from the Admin UI under the default configuration.

## Proposed fix (for discussion)
Either:
- **(a)** add `/llmchat` to the default `csrf_exempt_paths` **and** attach the existing `enforce_admin_csrf` dependency to the llmchat router, consistent with how `/admin` is handled; or
- **(b)** have the LLM Chat frontend fetch an HMAC token from `/auth/csrf-token` before connecting, so the global middleware is satisfied.

Happy to send a PR (with a test) for whichever approach you prefer.

dwightford:
[...]

Issue #4789, opened 2026-05-18
[BUG]: LLM Chat requires Redis but setup docs don't mention it


obuzek:
## Describe the bug

The Admin UI's LLM Chat feature (`/llmchat/connect`) returns a 500 error with `"Service initialization failed"` when Redis is not configured. The LLM Chat session management depends on Redis for session state (TTL, locks, chat history), but this dependency is not documented in the setup instructions or `.env.example`.

## Steps to reproduce

1. Fresh install following README quick start (no Redis)
2. Configure an LLM provider via `POST /llm/providers` with an API key
3. Register a model via `POST /llm/models`
4. Provider health check passes (`POST /llm/providers/{id}/health` returns `healthy`)
5. Open Admin UI → LLM Chat
6. Console shows repeated `connectLLMChat` errors: `500 Service initialization failed`

## Expected behavior

Either:
- LLM Chat should work without Redis (using in-memory session fallback), OR
- Setup docs and `.env.example` should clearly state Redis is required for LLM Chat
- The 500 error should return a descriptive message like "Redis is required for LLM Chat sessions. Set REDIS_URL in your .env"

## Relevant config

```
# These are in .env but commented out and not called out as required:
# LLMCHAT_ENABLED=true
# LLMCHAT_SESSION_TTL=300
# LLMCHAT_SESSION_LOCK_TTL=30
# LLMCHAT_CHAT_HISTORY_TTL=3600
```

The `REDIS_URL` env var is not set by default and `.env.example` doesn't indicate it's needed for LLM Chat.

## Environment

- ContextForge v1.0.1
- macOS, SQLite, gunicorn
- No Redis configured (default state after setup)

## Suggested fix

1. Add a note to the LLM Chat section of the docs: "Requires Redis"
2. Return a 503 with a clear error message instead of a generic 500 when Redis is unavailable
3. Consider an in-memory session fallback for single-node dev setups

jonpspri:
Thanks for the detailed report. This has been addressed in the codebase — closing as resolved.

**In-memory fallback (no Redis required):** `llmchat_router.py` now includes explicit in-memory fallback stores (`active_sessions`, `user_configs`) and `ChatHistoryManager` accepts `redis_client=None` with graceful in-memory fallback. If LLM Chat is enabled without Redis, startup now logs a clear warning rather than crashing:
[...]

PR #5790, opened 2026-07-23
fix(llmchat): allow any worker to resume LLM Chat sessions from Redis

### LINKED ISSUE #5740: [BUG]: LLM Chat sessions are in-memory only (redis_client hardcoded None) — breaks under multi-worker gunicorn
## Describe the bug

LLM Chat sessions (`/llmchat/connect` + `/llmchat/chat`) are stored purely in-process memory, with **no cross-worker sharing**, even though the code accepts a `redis_client` parameter that implies Redis-backed sessions are supported. Under any multi-worker deployment (the standard `make serve` / gunicorn setup — the documented production path), a `/llmchat/connect` and a following `/llmchat/chat` can land on different workers, and the second call fails:

```json
{"detail":"No active session found. Please connect to a server first."}
```

This reproduces even with `REDIS_URL` correctly configured and Redis reachable, because Redis is never actually wired up for this feature.

## Root cause

`mcpgateway/routers/llmchat_router.py:59`:

```python
# Redis client (initialized via init_redis() during app startup)
redis_client = None
```

This module-level variable is hardcoded `None` and is never assigned from `settings.redis_url` / the app's real Redis client anywhere I could find, despite the comment implying `init_redis()` sets it at startup, and despite `MCPChatService(config, user_id=user_id, redis_client=redis_client)` being called with it as if it were live. Sessions (`get_active_session` / `set_user_config` / `delete_active_session`) end up using an in-memory, per-process store.

Consequence: `/llmchat` is only usable with exactly one worker process. This intersects with #5215 (session-token auth to the loopback `/servers/{id}/mcp` call) and #5214/#5223/#5739 (CSRF) — once those are fixed, this becomes the next blocker for anyone running LLM Chat behind gunicorn with `--workers > 1`, which is the default production setup (`run-gunicorn.sh`, auto-scaled to `2*CPU+1`).

This also gives a more precise root cause for #4789 ("LLM Chat requires Redis but setup docs don't mention it") — setting `REDIS_URL` does **not** fix that issue's symptom, because the Redis client is dead code, not an undocumented-but-functional dependency.

## Steps to reproduce

1. Run with gunicorn, `--workers >= 2` (`make serve` default).
2. Configure an LLM provider + model, register a virtual server.
3. `POST /llmchat/connect` (succeeds, 200).
4. `POST /llmchat/chat` in a **separate** connection/request (not pinned to the same TCP connection) — fails ~50% of the time (or reliably, depending on worker count and load balancing) with the 400 above.
5. Forcing both calls through the same persistent HTTP/1.1 connection (pinning to one worker) makes it work reliably — confirming the session literally does not exist outside that one worker's memory.

## Proposed fix

Either:
- **(a)** Actually wire `redis_client` from `get_redis_client()` (already imported in this file, used elsewhere) at module/app-startup time, so sessions are genuinely Redis-backed when `REDIS_URL` is configured, with a documented in-memory-only fallback (and a startup warning) when it isn't; or
- **(b)** Document that `/llmchat` requires `--workers 1` / sticky sessions at the proxy layer, and make the multi-worker case fail fast with a clear error at connect time rather than a confusing 400 on the *next* call.

## Environment

- ContextForge (main branch, `mcpgateway/routers/llmchat_router.py`)
- gunicorn multi-worker (`make serve`), any Redis config

## Related

- #4789 — same underlying gap, less precise root cause (assumes Redis works if configured)
- #5215 — separate blocker on the same feature (loopback auth token rejected)
- #5214 / #5223 / #5739 — separate blocker on the same feature (CSRF)

msureshkumar88:
## Summary

Fixes #5740. `POST /llmchat/connect` on one gunicorn worker followed by `POST /llmchat/chat` landing on another worker returned `400 {"detail":"No active session found. Please connect to a server first."}`, even with `CACHE_TYPE=redis` + `REDIS_URL` configured correctly.

The issue's own diagnosis (`redis_client` is hardcoded `None`/dead code) doesn't hold up on the current checkout — it's assigned and wired through the lifespan, as an earlier comment on the issue also found. The actual bugs are two independent ones in `get_active_session()`:

1. **No cross-worker takeover.** Redis already stores the user's config, session ownership, and chat history. But when `get_active_session()` saw that another worker owned the session, it unconditionally `return None`'d instead of rebuilding the session locally from the Redis-backed config — even though everything needed to do so was already there. The existing retry loops for this case also polled this worker's own `active_sessions` dict waiting for a foreign worker to fill it, which can never happen.

2. **`WORKER_ID` computed at import time.** `WORKER_ID = str(os.getpid())` is a module-level assignment. `run-gunicorn.sh` enables `--preload` by default on Linux, so gunicorn imports the app once in the master process before forking — every worker inherited the *master's* PID as its `WORKER_ID`. The ownership/lock compare-and-delete Lua scripts then compared a value against itself, letting any worker delete or release another worker's lock/ownership key.

## Changes

- `init_redis()` (runs post-fork in the lifespan) now assigns a per-process, host-qualified `WORKER_ID` (`hostname:pid:uuid`), replacing the import-time module default.
- `get_active_session()` takes over a foreign-owned session by rebuilding it from the Redis-backed config and re-pointing ownership, instead of rejecting the request. Added a `recreate` parameter: `/connect` and `/disconnect` pass `recreate=False` since they're about to replace or tear down the session anyway and shouldn't pay for a full MCP handshake just to immediately discard it. `/status` no longer calls `get_active_session()` at all — it's now a pure Redis/local existence check, since a read-only GET should never trigger a session rebuild as a side effect.
- `_create_local_session_from_config()` now bounds `MCPChatService.initialize()` with the lock's TTL (`asyncio.wait_for`), so a slow MCP handshake can't outlive the lock and let two workers build duplicate sessions for the same user.
- The user config's Redis TTL is now refreshed alongside the ownership key's (previously only the ownership key was refreshed on activity, so a long conversation's config could silently expire out from under a later takeover).
- Added an idle reaper for orphaned local sessions left behind after a takeover, which skips any session with an in-flight streaming response (re-stamped per emitted SSE chunk) so it can't evict mid-stream.
- `init_redis()` now warns (previously silent) when LLM Chat is enabled without Redis configured.
- Updated `docs/docs/using/clients/llm-chat.md` — the previous "worker affinity / sticky sessions" description no longer matches the behavior.

## Related

- #4789 (Redis requirement undocumented) is largely addressed by this PR's docs update and startup warning.
- #5215 and #5214/#5223/#5739 are separate, unrelated blockers on the same feature (loopback auth, CSRF).

### DIFF
--- a/docs/docs/using/clients/llm-chat.md
+++ b/docs/docs/using/clients/llm-chat.md
@@ -443,8 +443,9 @@ The LLM Chat system follows a **three-tier architecture** with clear separation 
 
 - `active_sessions`: Dict storing active MCPChatService instances per user
 - `user_configs`: Dict storing user configurations
-- **Redis Integration** (optional): Distributed session storage with TTL and worker affinity
-- **Sticky Sessions**: Workers use locks to prevent race conditions during session initialization
+- **Redis Integration** (optional): Distributed storage for user config and session ownership, with TTLs
+- **Cross-Worker Takeover**: Any worker may materialize a user's session locally from the Redis-backed config, rather than requiring the request to stay pinned to the worker that created it. Locks coordinate initialization so concurrent requests don't build duplicate sessions, and an idle reaper shuts down a session once its previous worker's copy is no longer in use
+- Without `CACHE_TYPE=redis` + `REDIS_URL`, session and config state is per-process only, so LLM Chat only works reliably with a single worker
 
 ![LLM Chat Session Management](llm-chat-session-management.png)
 
@@ -607,11 +608,12 @@ MCPClientConfig
 - TTL-based expiration reduces memory footprint
 - Redis caching for distributed deployments
 
-**Sticky Sessions** (Multi-worker):
+**Session Takeover** (Multi-worker):
 
-- Worker affinity via Redis ownership keys
-- Reduces session recreation overhead
-- Lock-based coordination prevents race conditions
+- Any worker can resume a user's session by rebuilding it locally from the Redis-backed config -- requests are not required to stay pinned to the worker that created the session
+- Lock-based coordination prevents concurrent requests from building duplicate sessions for the same user
+- An idle reaper shuts down a worker's local copy once it is no longer the session's owner and has been unused past the session TTL
+- Requires `CACHE_TYPE=redis` + `REDIS_URL`; without Redis, session and config state is per-process only, so LLM Chat only works reliably with a single worker
 
 
 ### 🧪 Error Handling
--- a/docs/docs/using/clients/llm-chat.md
+++ b/docs/docs/using/clients/llm-chat.md
@@ -41,6 +41,7 @@ All LLM providers are configured via the Admin UI's LLM Settings. Navigate to **
 ```bash
 # ===== Redis Connection =====
 CACHE_TYPE=redis                          # Enable Redis
+# Default: database -- must be set to "redis" for multi-worker LLM Chat
 REDIS_URL=redis://localhost:6379/0       # Redis connection string
 
 # ===== Session Management =====
--- a/mcpgateway/routers/llmchat_router.py
+++ b/mcpgateway/routers/llmchat_router.py
@@ -18,8 +18,10 @@ history management via ChatHistoryManager from mcp_client_chat_service.
 # Standard
 import asyncio
 import os
+import socket
 import time
 from typing import Any, Dict, Optional
+from uuid import uuid4
 
 # Third-Party
 from fastapi import APIRouter, Depends, HTTPException, Request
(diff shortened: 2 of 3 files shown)


gandhipratik203 (changes requested):
Nice work on this. The re-diagnosis is spot on: the import-time `WORKER_ID` under `--preload` and the unconditional reject of foreign-owned sessions are the real root causes, and the takeover approach is the right call. Test coverage is thorough, including the fail-safe-on-Redis-error paths and the dead-loop retry fix.

One blocking item and one follow-up:

**Blocking: stale session served after a reconnect.** In `get_active_session`, branch 3's reclaim returns the local session on ownership drift without checking that the stored config still matches. If a user reconnects on another worker with a new server/model, a `/chat` routed back to the old worker within `SESSION_TTL` (default 300s) can silently serve the old session. That turns what was a retriable 400 pre-PR into a silent wrong-server response. One option: version the config in Redis on `/connect` (a generation counter or fingerprint), store it on the `MCPChatService`, and compare in branches 1 and 3, rebuilding on mismatch.

**Non-blocking follow-up: `/disconnect` on a non-owning worker doesn't fully disconnect.** `recreate=False` returns None locally, and the CAS only deletes the ownership key when this worker owns it, so the owning worker keeps the session (and its live downstream connection) running until the reaper catches it (~5 min). This is pre-existing and self-heals, so a comment or follow-up issue is probably enough rather than fixing it here.

msureshkumar88:
Thanks for the thorough review — the blocking issue was real, confirmed it in `get_active_session` branches 1 and 3 exactly where you pointed.

Fixed in the latest commit: `/connect` now stamps a fresh generation token (`uuid4().hex`) onto the config in Redis and onto the resulting `MCPChatService.config_version`. Both the owner-match refresh path (branch 1) and the takeover-reclaim path (branch 3) now compare the locally-cached session's stamped version against the current Redis version before reusing it; on mismatch the stale session is shut down and discarded, and a fresh one is rebuilt from the current config instead of being silently served or reclaimed. Added `_local_session_is_stale()` / `_discard_stale_session()` plus regression tests covering both branches.

Left the `/disconnect` non-blocking follow-up as-is per your suggestion — it's pre-existing, self-heals via the idle reaper, and a fix there felt out of scope for this PR. Happy to open a follow-up issue for it if you'd prefer that over a code comment.

gandhipratik203 (dismissed):
The config-version guard resolves the stale-session-after-reconnect issue cleanly, and the branch 1 / branch 3 staleness checks plus test coverage look right. Approving; the `/disconnect`-on-a-non-owning-worker cleanup is fine to leave as a follow-up.

ja8zyjits:
## ⚠️ Warnings (3)

### Documentation

1. **`CACHE_TYPE` default not explicit**
   - File: `docs/docs/using/clients/llm-chat.md:95`
   - Issue: Docs say "CACHE_TYPE=redis" required for multi-worker but don't state default is `"database"`
   - Fix: Add note that default is `"database"`, must change to `"redis"`

2. **Env var naming incomplete**
   - File: `docs/docs/using/clients/llm-chat.md:95-115`
   - Issue: Shows `LLMCHAT_` prefix but not config field names (`llmchat_session_ttl`)
   - Fix: Document both forms (env var uppercase, config field lowercase)

3. **`recreate` param behavior unclear**
   - File: `docs/docs/using/clients/llm-chat.md:446-448`
   - Issue: Takeover description accurate but missing `recreate` param details
   - Fix: Note `/connect`, `/disconnect`, `/status` use `recreate=False`; `/chat` uses `recreate=True`

---

ja8zyjits (changes requested):
check comments

msureshkumar88:
Thanks for the pass — went through all three against the current checkout.

**1. `CACHE_TYPE` default — fixed.** Confirmed the default is `"database"` (`mcpgateway/config.py:2683`). Added an explicit note in the env block. Minor correction: the `CACHE_TYPE` line is at `llm-chat.md:43`, not `:95` — line 95 is the auth-token note under "Select a Virtual Server". This block is also outside this PR's diff (the PR touches only 443-448 and 607-618), but it's a one-line fix so I've taken it here.

**2. Env var naming — I'd rather leave this as-is.** Settings use `SettingsConfigDict(env_file=".env", case_sensitive=False, extra="ignore")` with no `env_prefix` (`config.py:2938`), so `llmchat_session_ttl` and `LLMCHAT_SESSION_TTL` resolve to the same key rather than being two forms. The lowercase spelling is the Python attribute, not a separate configuration surface — documenting both in a user-facing env reference would suggest two mechanisms where there's one. Happy to reconsider if you're seeing a case where the lowercase form matters to users. (The cited range 95-115 is the UI walkthrough; the env block is 39-72.)

**3. `recreate` param — the suggested wording isn't accurate for `/status`.** `/status` never calls `get_active_session()` at all; it's a pure existence check via `_has_active_session_state()` (`llmchat_router.py:1496`, endpoint at `:1526`, called out in the docstring at `:1531`). That was deliberate — a read-only GET shouldn't trigger a session rebuild as a side effect. The rest matches: `/connect:1103` and `/disconnect:1474` pass `recreate=False`, `/chat:1364` uses the default `True`.

`recreate` is also an internal helper kwarg rather than public API, so I'd lean toward keeping it out of the user-facing doc. If you'd still like the behavior surfaced there, I'd write it as "`/status` reports on existing state only and never rebuilds a session" rather than in terms of the parameter — let me know and I'll add it.

ja8zyjits (approved):
LGTM

Issue #5369, opened 2026-06-24
[FEATURE]: UI-rewrite style & interaction updates


a-effort:
## Overview

Style and interaction updates for initial part of ContextForge redesign (core features & MCP)

**Part of Epic:** #4093

## Timeline

**Target Date:** August 14, 2026

## Deliverables

- PRs merged for style and interaction updates for initial part of ContextForge redesign (core features & MCP)

## Acceptance Criteria

- [x] Improvements merged: https://github.com/IBM/mcp-context-forge/issues/4093

a-effort:
Fix for bug: https://github.com/IBM/mcp-context-forge/pull/6115

https://github.com/user-attachments/assets/a7e1b561-78c3-4e34-8b8b-42a3ec564f9e

a-effort:
Layout shift fix: https://github.com/IBM/mcp-context-forge/pull/5946

a-effort:
Small adjustments (if not done as part of https://github.com/IBM/mcp-context-forge/pull/5795: 

https://github.com/IBM/mcp-context-forge/pull/5795#issuecomment-5223135873

a-effort:
Will open issues for additional follow up changes to styles / interactions.

PR #5795, opened 2026-07-23
feat(ui-rewrite): rename Public visibility to Internal and add level tooltips

### LINKED ISSUE #5732: [FEATURE]: Rename "Public" visibility label to "Internal" in ui-rewrite and add tooltips for all three levels
## Summary

The visibility selector in ui-rewrite offers three options with labels `Private`, `Team`, `Public`. The label `Public` is misleading: in this codebase it means "visible to every authenticated user on this platform," not "visible on the public internet." Developers coming from GitHub/GitLab read `Public` as internet-anonymous by default, which is a real risk of accidental over-sharing.

This issue proposes a UI-only rename of `Public` to `Internal`, plus tooltips on all three visibility options. The DB enum value, API contract, and access-control code all stay on the literal `"public"`. If the new terminology settles in over a release or two, a follow-up issue can propose renaming the enum value itself with an Alembic migration.

## Motivation

- This came up as a point of confusion during recent usability testing ("what does public mean")
- GitHub and GitLab both use `Internal` for "everyone signed into this instance" and `Public` for "internet-anonymous." Developers already recognize this pattern.
- The details panels display the raw label with a Globe icon (`ToolDetailsPanel.tsx:229`, `ResourceDetailsPanel.tsx:208`, `PromptDetailsPanel.tsx:277`, `VirtualServerDetailsPanel.tsx:664`, `MCPServerDetailsPanel.tsx:514`) with no explanation, which makes the ambiguity worse for reviewers auditing sharing.

## Scope

**In scope (UI copy only):**

- Rename the `Public` option label to `Internal` in all visibility selectors.
- Add tooltip/help text for `Private`, `Team`, and `Internal`.
- Update the details-panel display so the label is `Internal`.
- Update the icon so it reinforces the meaning (vs globe)

**Out of scope (deliberately deferred):**

- DB enum values remain `"private" | "team" | "public"` in `schemas.py:100`, `db.py`, and the access-control code in `services/base_service.py`.
- API request/response payloads still use `"public"` on the wire.
- No Alembic migration.
- No changes to the legacy HTMX admin UI.

## Affected files (ui-rewrite)

Visibility selectors:
- `client/src/.../ToolAdvancedSettings.tsx:115-127` (Select, hardcoded labels)
- `client/src/.../ResourceForm.tsx:258-273` (Select, i18n)
- `client/src/.../PromptForm.tsx:109-133` (Select, i18n)
- `client/src/.../AdvancedSettings.tsx:212-224` (MCP Servers, hardcoded)
- `client/src/.../CreateServerForm.tsx:119-150` (Virtual Servers, radio group)

Details panels displaying the current level:
- `ToolDetailsPanel.tsx:229`
- `ResourceDetailsPanel.tsx:208`
- `PromptDetailsPanel.tsx:277`
- `VirtualServerDetailsPanel.tsx:664`
- `MCPServerDetailsPanel.tsx:514`

Shared type stays untouched:
- `client/src/types/server.ts:1` continues to declare `type Visibility = \"private\" | \"team\" | \"public\"`.

The `getVisibilityLabel()` helper in `ToolDetailsPanel.tsx:112-121` (and any similar mapping) is the seam: it maps the wire value `\"public\"` to the display string `Internal`. Prompts/Resources already read labels from i18n; the string key values change, the message IDs do not.

## Proposed copy

| Wire value | Label | Tooltip / help text |
|---|---|---|
| \`private\` | Private | Only you can see this. Team members and other platform users cannot. |
| \`team\` | Team | Scoped to a team you belong to. Existing conditional help stays: \"This [tool] will be scoped to your currently selected team.\" |
| \`public\` | Internal | Visible to everyone signed into this platform. Not on the public internet. |

The `\"public\"` wire value is a known point of confusion for anyone reading the DB, API, or JWT scoping code directly. Two options for handling it:

1. Leave the mismatch and document it in a code comment near `getVisibilityLabel()` and in the RBAC docs (`docs/docs/manage/rbac.md`).
2. Follow up with a separate issue to rename the enum value itself, with the Alembic migration and config-snapshot pattern from AGENTS.md.

Option 1 is what this issue proposes. Option 2 is worth doing only if the \"Internal\" label survives review feedback.

## Non-goals

- Adding a fourth \"true public / internet-anonymous\" tier. If that is ever needed, it should be a separate design conversation.
- Changing the access-control behavior in `services/base_service.py`. This issue does not touch Layer 1 token scoping or Layer 2 RBAC.

Altamimi-Dev:
# Pull Request

## 🔗 Related Issue

Closes #5732

---

## 📝 Summary

The `Public` visibility label is misleading: in ContextForge it means "visible to everyone signed into this platform", not "on the public internet". Developers coming from GitHub/GitLab read `Public` as internet-anonymous, a real risk of accidental over-sharing.

This is a UI-only change:

- Renames the `public` visibility **label** to **Internal** across all visibility selectors (Tools, Resources, Prompts, MCP servers, Virtual servers), the details panels, and the MCP servers table.
- Adds a reusable `VisibilityInfoPopover` (info icon) next to each visibility selector and details-panel value, explaining all three levels (Private / Team / Internal). It is a popover rather than a tooltip so the explanation is reachable on touch devices and its content stays hoverable and dismissible (WCAG 1.4.13).
- Moves the `Internal` / `Private` / `Team` labels onto shared `common.visibility.*` i18n keys (en-US, es-ES, pt-BR), including the two helpers that previously hardcoded them (`MCPServerDetailsPanel.getVisibilityLabel`, `ServersTable.getVisibilityConfig`).
- Swaps the misleading globe icon for a building icon on the visibility rows.

The wire value stays `"public"` — the DB enum, API request/response payloads, and access-control code are untouched. A note documenting the label/wire-value mismatch was added to the RBAC docs and to the shared component.

---

## 📏 Reviewability

---

## 🏷️ Type of Change

- [ ] Bug fix
- [x] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

| Check | Command | Status |
|-------|---------|--------|
| Lint | `cd client && npm run lint` | ✅ |
| Format | `cd client && npm run format:check` | ✅ |
| Unit tests | `cd client && npm run test:run` | ✅ 149 files, 2688 passed |
| Playwright | `cd client && npx playwright test e2e/virtual-servers.spec.ts` | ✅ 37 passed |

`VisibilityInfoPopover.test.tsx` covers the trigger, click-to-open showing all three levels, and Escape dismissal. `AdvancedSettings.test.tsx` and `VirtualServerDetailsPanel.test.tsx` each assert the info trigger is actually wired into a form and a details panel, so a refactor can't silently drop it. Existing details-panel, table, and form tests were updated to assert the `Internal` label.

---

## ✅ Checklist

---

## 📓 Notes (optional)

Out of scope (deliberately deferred, per the issue): renaming the DB enum value itself (would need an Alembic migration) and the legacy HTMX admin UI.

**Team visibility labels are intentionally unchanged.** `TeamForm`'s Public/Private selector and `TeamsTable`'s raw value are left as-is because a team's `public` flag governs discoverability and join semantics — who can find and request to join the team — rather than resource exposure. `Internal` would be the wrong word there, and picking the right one is a separate design question that deserves its own issue against the epic.

## 🎥 Demo

https://github.com/user-attachments/assets/0a5c64ea-6366-4aea-b336-11cae2b88af9


marekdano (changes requested):
@Altamimi-Dev 
Thanks for this - it's a clean, well-scoped change!

A few notes below, grouped by severity. 

## Suggestions

- **`Tooltip` vs `Popover` for the explanation.** The info bubble carries three lines of security-relevant text that a user may need to read carefully, but Radix `Tooltip` isn't *hoverable* (the content closes as the pointer moves toward it) and doesn't fire on touch/mobile. That's the "hoverable" leg of [WCAG 1.4.13](https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html), and Radix's own guidance is to reach for `Popover` (click-triggered, dismissible, hoverable) when the content is something people actually need to read rather than a short label. Since the whole point of the PR is to *clarify* a setting people misread, a `Popover` - or even inline helper text under the selector - would land the message more reliably for touch and low-vision users. Worth weighing; not a blocker.

- **Hardcoded labels in the two Advanced Settings forms.** `AdvancedSettings.tsx` and `ToolAdvancedSettings.tsx` still hardcode the `<SelectItem>` text (`Internal` / `Private` / `Team`) in English, while the tooltip you added right beside them *is* translated. So in `es-ES` / `pt-BR` the dropdown reads English "Internal" next to a Spanish/Portuguese tooltip. It's pre-existing (they were hardcoded `"Public"` before), so not a regression — but since you're already editing these exact lines, moving them to `intl.formatMessage` would be a cheap win. Same pattern in `MCPServerDetailsPanel.getVisibilityLabel` and `ServersTable.getVisibilityConfig`.

- **Teams visibility selector wasn't relabeled.** `TeamForm` still shows `teams.create.visibility.public` = "Public"/"Público" (and `TeamsTable` renders a raw lowercase `public`). The PR notes defer "team-visibility labels," but the phrasing is a little ambiguous — is the team's own public/private selector intentionally out of scope? The end state is that a user sees a tool as **Internal** but a team as **Public** in the same UI, which partly reintroduces the confusion this PR is removing. Totally fine to defer — just worth one explicit line so it reads as a decision rather than a miss.

- Playwright tests should be fixed

## Minor notes

- **Test coverage.** The new `VisibilityInfoTooltip.test.tsx` nicely covers the trigger's accessible name and all three levels; existing panel/table/form tests were consistently updated to `Internal`. Two small gaps if you feel like locking things in: (1) nothing asserts the tooltip is actually *wired into* the forms/panels, so a refactor could silently drop it, and (2) no assertion on the `aria-describedby` association or Escape-to-dismiss, which would be nice given a11y is part of the goal.

Altamimi-Dev:
Thanks for the thorough review @marekdano — the WCAG point in particular was a good catch, and it made the component better. Everything is addressed in `4936e52ac` (also rebased onto current `epic/ui-rewrite`).

**1. Popover instead of Tooltip** — agreed, and switched. `VisibilityInfoTooltip` is now `VisibilityInfoPopover`, built on the existing `@/components/ui/popover` wrapper. It opens on click (so it works on touch), the content is hoverable, and Escape / outside-click dismiss it — which covers 1.4.13 hoverable/persistent. Clean rename across all 10 consumers, no alias left behind.

**2. i18n the hardcoded labels** — done. Added shared `common.visibility.private|team|internal` keys to en-US, es-ES and pt-BR (reusing the wording already present in the per-domain translations), and wired up all four sites you named: `AdvancedSettings.tsx`, `ToolAdvancedSettings.tsx`, `MCPServerDetailsPanel.getVisibilityLabel`, and `ServersTable.getVisibilityConfig`. The last one now returns a `labelId` and formats at the render site, matching the pattern already used in `SourceSelection.tsx`.

**3. Team visibility deferral** — you were right that this was implicit. The PR description now states it explicitly: team `public` governs discoverability and join semantics rather than resource exposure, so `Internal` would be the wrong word there. Renaming it is a separate design question worth its own issue against the epic.

**4. Playwright** — fixed. The two failures were in `client/e2e/virtual-servers.spec.ts` (lines 940 and 1312) still asserting the old `Public` label; both now expect `Internal`. `teams.spec.ts` is untouched since its `Public` strings belong to the team entity. Locally `npx playwright test e2e/virtual-servers.spec.ts` is 37/37 green.

**Minor: wiring assertions** — added one to `AdvancedSettings.test.tsx` (form) and one to `VirtualServerDetailsPanel.test.tsx` (panel), each asserting the info trigger renders. Escape/dismiss behaviour is covered once in `VisibilityInfoPopover.test.tsx` rather than duplicated per consumer.

Full client suite: 149 files, 2687 passed.


One note on CI: the three client workflows on the new head are sitting at `action_required` (they ran automatically on the previous push, so this looks like the fork-approval gate rather than a config change). Whenever a maintainer approves the run they should go green — happy to re-check and follow up if anything surfaces.

a-effort:
Thanks for contributing this work! 🙌 

A couple of small, non blocking tweaks (can be added to https://github.com/IBM/mcp-context-forge/issues/5369 if not done as part of this PR)

1. In details drawers, the popover content can be shortened to the selected visibility option. e.g. `internal` would just have the description: "Visible to everyone signed into the platform. Not on the public internet." instead of all three options: 
details

2. Icon swaps:
- private: https://lucide.dev/icons/lock
- team: https://lucide.dev/icons/users 

note: Internal can stay as https://lucide.dev/icons/building-2, at least for now. Good idea!

Altamimi-Dev:
Both tweaks are in this PR (`3c982769d`) rather than deferring to #5369 — they were small and self-contained.

**Selected-level description in drawers.** `VisibilityInfoPopover` now takes an optional `visibility` prop: when it is a known level the popover shows just that level's description, otherwise it falls back to all three. The five details drawers pass their value; the five forms deliberately do not, since someone picking a level still benefits from seeing all the options explained. To avoid duplicating translations, the `common.visibility.info.*` strings lost their `Private:` / `Team:` / `Internal:` prefixes and the all-three view recomposes each line from the existing `common.visibility.*` label keys — rendered text there is unchanged.

**Per-level icons.** Added a small `getVisibilityIcon` helper next to the popover: private → `Lock`, team → `Users`, internal → `Building2`. The five drawers were each hardcoding a single icon regardless of level (`Building2` in tool/resource/prompt, `Users` in the two server panels) and now derive it. While in there I also switched the team icon in `ServersTable` and `SourceSelection` from `Shield` to `Users` so the mapping is consistent everywhere a level renders.

Two tests cover the new behaviour (both fail without the change): one on the popover asserting single-level content, and the existing `VirtualServerDetailsPanel` popover test extended to open the popover and assert only the internal description appears. Full client suite is green — 149 files, 2688 passed.

Actions may sit at `action_required` until a maintainer approves the fork run, as it did on the earlier push.

---

**Per-level icons** — the MCP servers table shows all three at once, and is also where the team icon moved off `Shield`:

Visibility column showing a padlock for Private, a people glyph for Team and a building for Internal

**Selected-level description in drawers** — the popover now shows only the level the record is at. Internal:

Virtual server details drawer at Internal visibility, popover showing only the internal description

Private:

Virtual server details drawer at Private visibility, popover showing only the private description

**Forms keep all three**, since someone picking a level benefits from every option being explained. This one also shows the de-prefixed strings recomposing to the same text as before:

Create server form popover listing Private, Team and Internal descriptions with their level-name prefixes

One placement note: in drawers the popover opens with `side="left"`, so it sits over the row it describes. That is pre-existing behaviour, not something this change introduced — happy to look at it separately if it bothers you.

marekdano (commented):
@Altamimi-Dev -  Nice work!

 **Verdict: Approve with minor suggestions.** Clean, well-scoped, well-tested UI-only change. No blocking issues found.

Please make sure that all CI checks pass.

## Suggestions (non-blocking)

  - **i18n key duplication / mixed migration.** The PR introduces shared `common.visibility.*` keys and migrates `ServersTable` and `MCPServerDetailsPanel` to them, but many components still use page-specific keys carrying the identical
  "Internal"/"Private"/"Team" strings (`prompts.add.visibility.*`, `prompts.details.visibility.*`, `resources.form.visibility.*`, `resources.details.visibility.*`, `tools.details.visibility.*`, `gateways.createServer.visibility.*`).
  That's ~7 places holding the same three strings — a drift risk (one gets reworded, others don't) and inconsistent with the new shared approach. Consolidating the rest onto `common.visibility.*` in a follow-up would be cleaner. Reasonable to defer, but worth a tracking note.
  - **`getVisibilityIcon` fallback is the Internal icon.** Unknown/absent `visibility` falls back to `Building2` (the "Internal" building). If a value is ever missing, the UI would visually imply "Internal." Detail panels normally always have a value, so this is an edge case — but a neutral icon (or the label already reads "Not available") would be marginally safer than one that connotes a specific level.
  - **PromptDetailsPanel** keeps its inline label ternary rather than a `getVisibilityLabel` helper like the sibling panels. Pre-existing, but aligning it would reduce the per-panel divergence.

marekdano:
The PR is brought to https://github.com/contextforge-org/contextforge-web-ui/pull/13

PR #5802, opened 2026-07-23
fix(mcp): resolve session-token sub UUID to email in streamable-HTTP auth

### LINKED ISSUE #5215: LLM Chat cannot connect to a Virtual Server on the same gateway — forwarded session token rejected (401) by /servers/{id}/mcp

### LINKED ISSUE #5750: [BUG]: MCP streamable-HTTP transport treats session-JWT UUID `sub` as an email — non-admin session tokens see 0 tools and fail `tools.execute` RBAC
### 🐞 Bug Summary

The **MCP streamable-HTTP transport** (`_auth_jwt` in `mcpgateway/transports/streamablehttp_transport.py`) treats a **session token's `sub` claim as an email**, but since v1.0.3 login tokens set `sub = str(user.id)` (a UUID), not the email. On this transport the raw UUID is never resolved back to an email, so for every non-admin session token:

1. **`tools/list` returns 0 tools** — team resolution finds no memberships for the "email" `<uuid>` → caller is treated as *public-only*; all `visibility='team'` tools are hidden.
2. **`tools/call` returns "Access denied"** — the RBAC `tools.execute` check runs against the UUID as if it were the user's email → no roles found → denied.

The UUID→email resolution already exists on CF's REST paths (`set_user_context_from_token`, `_enforce_revocation_and_active_user`) and partially inside `resolve_session_teams()`. **Only the MCP streamable-HTTP transport was missed** — a classic incomplete migration. This affects **every non-admin MCP client** of any CF ≥ v1.0.3 deployment, on every agent framework.

---

### 🧩 Affected Component

Specifically: `mcpgateway/transports/streamablehttp_transport.py` (`_auth_jwt`, `_check_streamable_permission`) for `token_use == "session"` tokens minted by `POST /auth/login`.

---

### 🔁 Steps to Reproduce

Requires nothing but stock CF ≥ v1.0.3, one non-admin user, one team, and one `visibility='team'` tool.

1. Create a non-admin user who is an **active member of one team**, and register a gateway/tool that is `visibility='team'` and scoped to that team.
2. Grant that user (by email) the `tools.execute` / `tools.read` RBAC roles on the team.
3. Log in as that user via `POST /auth/login` to obtain a **session** token (`token_use: "session"`, `sub = <user UUID>`).
4. Do the MCP handshake against the unified endpoint `POST /mcp/` with `Authorization: Bearer <session-token>`:
   - `initialize` → `notifications/initialized` → `tools/list` → **returns 0 tools**.
5. If you first work around Layer 1 (e.g. `AUTH_CACHE_BATCH_QUERIES=false`, see below) so the tool list populates, then `tools/call <team-tool>` → **"Access denied"**.

**Controlled proof** — two session tokens minted with CF's own `create_jwt_token`, identical in every claim except `sub`, against the same live `/mcp/` server:

| Token | `tools/list` | `tools/call` |
|---|---|---|
| `sub` = UUID (exactly what `/auth/login` mints) | **0 tools** | — |
| `sub` = email (control) | **all team tools** | ✅ success |

Direct function-level confirmation:

```python
auth._get_user_team_ids_sync("user@example.com")            # → [team...]
auth._get_user_team_ids_sync("<user-uuid>")                 # → []   (queried as EmailUser.email == '<uuid>')

# resolve_session_teams has a UUID→email net, but it is bypassed on the batched path:
resolve_session_teams({sub: uuid, token_use: session}, uuid, ...)                    # → [team...]  (own DB lookup, UUID net runs)
resolve_session_teams({...}, uuid, ..., preresolved_db_teams=[])                     # → []         (poisoned: net only guards the skipped DB call)
```

---

### 🤔 Expected Behavior

A session token issued by `/auth/login` for a non-admin team member should, on the MCP streamable-HTTP transport:

- resolve `sub` (UUID) → the user's email before any identity-based lookup, exactly as the REST paths do;
- return that user's team-visible tools from `tools/list`; and
- pass the `tools.execute` RBAC check for `tools/call` on tools in a team the user belongs to.

CF should not disagree with itself about the same user depending on which transport is used.

---

### 📓 Logs / Error Output

```
# agent-runtime side (every run, all frameworks):
"event": "..._mcp_tools_loaded", "count": 0

# Context Forge side, on tools/call:
Streamable HTTP RBAC denied: user=<user-uuid>, permission=tools.execute
```

Note the **UUID** in the RBAC log where an email is expected — this is the tell.

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Version or commit | present in v1.0.3 → v1.0.6; still on `main` at `c5da1865e` |
| Runtime | Python 3.11 |
| Platform / OS | Linux (containerized) |
| Container | Docker (`ghcr.io/ibm/mcp-context-forge`) |

---

### 🧩 Additional Context

**Root cause (current `main`)**

1. Login token mints a UUID `sub` — `mcpgateway/routers/email_auth.py:150`:
   ```python
   "sub": str(user.id),        # UUID; email is a separate column
   "token_use": "session",
   ```
2. MCP transport reads `sub` as the email and never resolves it — `streamablehttp_transport.py:5078`:
   ```python
   user_email = user_payload.get("sub") or user_payload.get("email")   # ← raw UUID
   ```
3. **Layer 1 (visibility):** the batched lookup `_get_auth_context_batched_sync(<uuid>)` queries `EmailUser.email == '<uuid>'` (`auth.py:1133`) → `team_ids = []`, which is then passed as `preresolved_db_teams=[]` into `resolve_session_teams` (`streamablehttp_transport.py:5236`). `resolve_session_teams` *does* have a UUID→email net (`auth.py:554-561`) — but it only feeds `_resolve_teams_from_db`, which is **skipped** when `preresolved_db_teams` is supplied (`auth.py:563-566`). Result: `final_teams = []` → public-only → `tools/list` = 0 tools. The empty result is also cached under the UUID key.
4. **Layer 2 (RBAC):** the UUID is stored as `auth_user_ctx["email"]` (`streamablehttp_transport.py:5301`) and passed straight into RBAC with no resolution — `streamablehttp_transport.py:1193,1200-1206`:
   ```python
   user_email = user_context.get("email")   # UUID
   granted = await permission_service.check_permission(user_email=user_email, permission="tools.execute", ...)
   ```
   → denied. This is the source of the `user=<uuid>` RBAC log line.

**Why only the MCP transport breaks:** CF's REST endpoints resolve the UUID properly (`set_user_context_from_token`, `_enforce_revocation_and_active_user` both have explicit UUID→email blocks), so the Admin UI and REST API look fine. Only the streamable-HTTP transport was missed.

**Partial mitigation (Layer 1 only):** setting `AUTH_CACHE_BATCH_QUERIES=false` forces the code off the batched path (`streamablehttp_transport.py:5123`) so `resolve_session_teams` runs its own UUID-aware DB lookup — `tools/list` then works. But **Layer 2 has no such fallback**: the RBAC check still reads the UUID from `user_context["email"]`, so `tools/call` remains denied. There is no config knob for the RBAC path.

**Suggested fix:** resolve the session-token `sub` UUID → email once, right after `verify_credentials()` in `_auth_jwt` (for `token_use == "session"`), before it is used for team resolution, cache keys, and `user_context["email"]`. This fixes both gates at a single root and reuses the pattern already present on the REST paths and in `resolve_session_teams`. After the fix, `AUTH_CACHE_BATCH_QUERIES` becomes unnecessary for this path.

**Related issues**

- #5215 — session token rejected (401) on `/servers/{id}/mcp` for LLM Chat loopback (different symptom/root cause, same "session tokens don't work on MCP transport" theme).
- #4452 (closed) — "Centralize Layer-1 visibility filter across streamablehttp transport (incl. `_auth_jwt` split-brain)" — same file/split-brain area; this bug is the identity-resolution facet still outstanding.

msureshkumar88:
Closes #5215
Closes #5750

## Problem

Session tokens issued by `/admin/login` carry `sub` as the `EmailUser.id` UUID (`mcpgateway/admin.py:4273`), while legacy/API tokens carry the email. The MCP streamable-HTTP auth path treated `sub` as an email unconditionally:

- `streamablehttp_transport.py:5089` — `_auth_jwt()`
- `streamablehttp_transport.py:2221` — `_normalize_jwt_payload()` (stateful-session fallback)

So the `EmailUser` lookup missed for every session token, with two different symptoms depending on configuration:

- **#5215** — with `require_user_in_db` enabled, `401 {"detail": "User not found in database"}` on `/servers/{id}/mcp`. LLM Chat could not connect to a virtual server hosted by the same gateway.
- **#5750** — on the unified `/mcp` endpoint the miss resolved *empty teams* instead, giving public-only visibility (`tools/list` returns 0) and a UUID-keyed RBAC check that denied `tools/call`. This affected every external agent runtime using a real session token, not just LLM Chat.

The failure was also **sticky**: the transport writes back to the auth cache (`:5153`, `:5164`, `:5206`) under the same unresolved key, so it cached `CachedAuthContext(user=None)` against the UUID. Subsequent requests took the cache-hit branch and returned the same 401 without re-querying.

## Fix

Add `_resolve_subject_email()` and call it in both sites, **before** any cache lookup or DB query. This mirrors resolution the rest of the codebase already does — `mcpgateway/auth.py:1581` (`get_current_user`) and `auth.py:553` (`resolve_session_teams`) — reusing the existing `_get_email_by_id_sync()` helper. The MCP transport was the one path that never got it.

Notes on the approach:

- **Fixed at the root rather than in `llmchat_router`.** The issue proposed minting a fresh loopback token in `connect()`. That resolves this one call site but leaves every external MCP client on a session token broken, so `llmchat_router` is unchanged — it keeps forwarding the session cookie, and the loopback call now authenticates as the real user, correctly scoped to their teams and RBAC.
- **No extra DB read on the hot path.** A `UUID()` parse guard short-circuits legacy/API tokens before any lookup. Net effect is a *reduction*: session tokens currently miss the cache on every request and re-run the full fallback path.
- **Fails closed.** An unresolvable subject returns the raw value so the existing `require_user_in_db` checks still deny, rather than downgrading to anonymous. A DB failure raises `SQLAlchemyError`, which the existing handler at `:5344` turns into 503 — deliberately unlike the fail-*open* revocation/user lookups at `:5183`/`:5194`, since an unresolvable subject means we don't know who the caller is.
- **All MCP transports** (`/mcp`, `/servers/{id}/mcp`, `/mcp/sse`, `/mcp/message`) share `_auth_jwt`, so all are fixed together.

## Behaviour change worth a look during review

The three deny branches read `elif settings.require_user_in_db and user_email != platform_admin_email`. Previously a platform-admin session token carried a UUID that never equalled `platform_admin_email`, so the admin escape hatch could not apply. It now does. That is the intended semantics of the hatch, but it is a widening on a security branch, so it has a dedicated test (`test_auth_jwt_session_token_platform_admin_escape_hatch`).

## Related, not included

`mcpgateway/routers/reverse_proxy.py:450`/`:481` use the same `sub`-or-`email` shape for session ownership. It is not broken the same way — `_validate_session_ownership` compares `sub` against `sub`, so both forms match. The only divergence is a session created under one token type and accessed under the other, which yields 403 (fail-closed). Left alone here; happy to file separately if it's worth normalizing.

## Tests

Added to `tests/unit/mcpgateway/transports/test_streamablehttp_transport.py`, covering the resolver directly, both patched call sites, and the deny paths: unknown user still 401s, legacy tokens perform no subject lookup, DB errors propagate to 503, and cache/lookup calls are keyed by email rather than UUID.

Verified the behavioural tests fail against the unpatched transport and pass with the fix.

### DIFF
--- a/mcpgateway/transports/streamablehttp_transport.py
+++ b/mcpgateway/transports/streamablehttp_transport.py
@@ -40,7 +40,7 @@ from enum import Enum
 import re
 from typing import Any, assert_never, AsyncGenerator, ContextManager, Dict, Iterable, List, Optional, Pattern, Tuple, Union
 from urllib.parse import urlsplit, urlunsplit
-from uuid import uuid4
+from uuid import UUID, uuid4
 
 # Third-Party
 import anyio
@@ -2203,6 +2203,55 @@ async def _get_request_context_or_default() -> Tuple[str, dict[str, Any], dict[s
         return s_id, request_headers_var.get(), user_context_var.get()
 
 
+async def _resolve_subject_email(payload: dict[str, Any]) -> Optional[str]:
+    """Resolve a JWT subject claim to the user's email address.
+
+    Session tokens (``token_use: "session"``) carry ``sub`` as the
+    ``EmailUser.id`` UUID; legacy/API tokens carry the email directly.  This
+    mirrors the resolution already performed by
+    ``mcpgateway.auth.get_current_user`` and
+    ``mcpgateway.auth.resolve_session_teams`` so that the MCP transport keys its
+    user lookups, auth-cache entries and RBAC context on the same identity as
+    the REST paths.
+
+    When the subject is a UUID that cannot be resolved (no matching user row),
+    the raw subject is returned unchanged so the caller's existing
+    ``require_user_in_db`` checks reject the request rather than silently
+    downgrading it to anonymous access.
+
+    Args:
+        payload: Decoded JWT payload.
+
+    Returns:
+        The user's email address, or ``None`` when the token carries no subject.
+
+    Examples:
+        >>> import asyncio
+        >>> asyncio.run(_resolve_subject_email({"sub": "user@example.com"}))
+        'user@example.com'
+        >>> asyncio.run(_resolve_subject_email({"email": "fallback@example.com"}))
+        'fallback@example.com'
+        >>> asyncio.run(_resolve_subject_email({})) is None
+        True
+    """
+    subject: Optional[str] = payload.get("sub") or payload.get("email")
+    if not subject:
+        return None
+
+    try:
+        UUID(subject)
+    except (ValueError, AttributeError, TypeError):
+        # Already an email (legacy/API token, or a session token issued before
+        # the user row existed) — no database round-trip needed.
+        return subject
+
+    # First-Party
+    from mcpgateway.auth import _get_email_by_id_sync  # pylint: disable=import-outside-toplevel
+
+    resolved = await asyncio.to_thread(_get_email_by_id_sync, subject)
+    return resolved or subject
+
+
 async def _normalize_jwt_payload(payload: dict[str, Any]) -> dict[str, Any]:
     """Normalize a raw JWT payload to the canonical user context shape.
 
@@ -2218,7 +2267,7 @@ async def _normalize_jwt_payload(payload: dict[str, Any]) -> dict[str, Any]:
     Returns:
         Canonical user context dict with keys email, teams, is_admin, is_authenticated, token_use.
     """
-    email = payload.get("sub") or payload.get("email")
+    email = await _resolve_subject_email(payload)
     jwt_is_admin = payload.get("is_admin", False)
     if not jwt_is_admin:
         user_info = payload.get("user", {})
@@ -5086,7 +5135,9 @@ class _StreamableHttpAuthHandler:
             from mcpgateway.cache.auth_cache import CachedAuthContext, get_auth_cache  # pylint: disable=import-outside-toplevel
 
             jti = user_payload.get("jti")
-            user_email = user_payload.get("sub") or user_payload.get("email")
+            # Resolve sub (a UUID for session tokens) to the user's email before any
+            # cache lookup or DB query, so MCP keys the same identity as the REST paths.
+            user_email = await _resolve_subject_email(user_payload)
             nested_user = user_payload.get("user", {})
             nested_is_admin = nested_user.get("is_admin", False) if isinstance(nested_user, dict) else False
             is_admin = user_payload.get("is_admin", False) or nested_is_admin
--- a/mcpgateway/transports/streamablehttp_transport.py
+++ b/mcpgateway/transports/streamablehttp_transport.py
@@ -2214,6 +2214,11 @@ async def _resolve_subject_email(payload: dict[str, Any]) -> Optional[str]:
     user lookups, auth-cache entries and RBAC context on the same identity as
     the REST paths.
 
+    Per the repository's canonical email-over-sub precedence (see
+    ``mcpgateway.auth_context.get_user_email``), a non-empty string ``email``
+    claim always wins over ``sub`` when both are present, so the resolved
+    identity can never be spoofed by a conflicting ``sub`` claim.
+
     When the subject is a UUID that cannot be resolved (no matching user row),
     the raw subject is returned unchanged so the caller's existing
     ``require_user_in_db`` checks reject the request rather than silently
@@ -2231,10 +2236,13 @@ async def _resolve_subject_email(payload: dict[str, Any]) -> Optional[str]:
         'user@example.com'
         >>> asyncio.run(_resolve_subject_email({"email": "fallback@example.com"}))
         'fallback@example.com'
+        >>> asyncio.run(_resolve_subject_email({"email": "canonical@example.com", "sub": "ignored-sub"}))
+        'canonical@example.com'
         >>> asyncio.run(_resolve_subject_email({})) is None
         True
     """
-    subject: Optional[str] = payload.get("sub") or payload.get("email")
+    email_claim = payload.get("email")
+    subject: Optional[str] = email_claim if isinstance(email_claim, str) and email_claim else payload.get("sub")
     if not subject:
         return None
 
--- a/tests/unit/mcpgateway/transports/test_streamablehttp_transport.py
+++ b/tests/unit/mcpgateway/transports/test_streamablehttp_transport.py
@@ -13278,6 +13278,239 @@ async def test_normalize_jwt_payload_session_admin_no_email_no_bypass():
 
 
 # ---------------------------------------------------------------------------
+# _resolve_subject_email tests (issue #5215)
+#
+# Session tokens carry sub = EmailUser.id (a UUID); legacy/API tokens carry the
+# email. The MCP transport used to treat sub as an email unconditionally, so
+# session tokens produced a 401 "User not found in database" on
+# /servers/{id}/mcp and empty-team public-only visibility on /mcp.
+# ---------------------------------------------------------------------------
+
+SESSION_SUB_UUID = "387e3345-7996-4496-aa2d-09729ec8b3be"
+
+
+@pytest.mark.asyncio
+async def test_resolve_subject_email_uuid_sub_resolves_to_email():
+    """A session token's UUID sub is resolved to the user's email via the DB."""
+    lookup = Mock(return_value="user@example.com")
+    with patch("mcpgateway.auth._get_email_by_id_sync", lookup):
+        result = await tr._resolve_subject_email({"sub": SESSION_SUB_UUID, "token_use": "session"})
+
+    assert result == "user@example.com"
+    lookup.assert_called_once_with(SESSION_SUB_UUID)
+
+
+@pytest.mark.asyncio
+async def test_resolve_subject_email_legacy_sub_skips_db_lookup():
+    """An email sub short-circuits before any DB round-trip (hot-path guard)."""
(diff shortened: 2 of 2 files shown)


Lang-Akshay (changes requested):

##  Please address following blocking Changes

| 1 | Security / Quality | `mcpgateway/transports/streamablehttp_transport.py` | 2237 (also consumed at 5140) | `_resolve_subject_email()` uses `payload.get("sub") or payload.get("email")`. The repository security invariant requires canonical email-over-sub precedence when both claims exist. Selecting `sub` first can make cache keys, team visibility, and RBAC evaluate a different principal than the canonical email, causing identity/authz confusion. The security specialist classified this as **Critical, CWE-287/CWE-863**. | Validate a non-empty string `email` first, fall back to `sub`, then UUID-resolve only the selected subject. Add deny-path coverage for conflicting `email`/`sub` claims. |

msureshkumar88:
Thanks for the catch, @Lang-Akshay — good call flagging the precedence order.

Fixed in 5688ee8: `_resolve_subject_email()` now validates a non-empty string `email` claim first and only falls back to `sub` when it's absent/malformed, matching the canonical email-over-sub precedence in `get_user_email()` (`mcpgateway/auth_context.py`).

Added deny-path coverage:
- conflicting `email`/`sub` claims → `email` wins, no UUID lookup performed
- non-string `email` claim → falls back to `sub`
- empty-string `email` claim → falls back to `sub`

Full transport suite (563 tests) still green. Ready for another look.

Lang-Akshay (approved):
✅ LGTM

ja8zyjits:
## ⚠️ Warnings (Can be a followup)

### Session token `sub` claim structure undocumented

**Context**: PR fixes bug where MCP transport treated `sub` as email unconditionally. Session tokens carry `sub` as `EmailUser.id` UUID (admin.py:4273), legacy/API tokens carry email directly.

**Files**:
- `mcpgateway/transports/streamablehttp_transport.py:2203-2260` (new `_resolve_subject_email()`)
- `mcpgateway/admin.py:4273` (session token minting)
- `docs/docs/architecture/multitenancy.md:323-363` (session token behavior)
- `docs/docs/architecture/oauth-design.md:250` (session-semantics payload)

**Gap**: Docs describe session token team resolution but never document `sub` claim structure differs between token types.

**Impact**: Medium — developers integrating MCP clients or debugging auth lack visibility. Fix correct and well-tested.

**Recommendation**: Add subsection to `docs/docs/architecture/multitenancy.md` under "Session Tokens" (L343-363):

```markdown
#### JWT Payload Structure by Token Type

**Session tokens** (`token_use: "session"`):
- `sub`: `EmailUser.id` (UUID) — resolved to email via DB lookup
- Minted by `/admin/login` and SSO flows
- Example: `{"sub": "387e3345-7996-4496-aa2d-09729ec8b3be", "token_use": "session", ...}`

**API/Legacy tokens** (`token_use: "api"` or absent):
- `sub`: User email address (string)
- Minted by `create_jwt_token()` utility
- Example: `{"sub": "user@example.com", "token_use": "api", ...}`

MCP transport resolves session token UUIDs to emails before cache lookups and RBAC checks (`_resolve_subject_email()` in `streamablehttp_transport.py`), ensuring consistent identity keying.
```

---

ja8zyjits (approved):
LGTM

PR #5742, opened 2026-07-21
[UI-REWRITE] Show source onboarding on empty dashboard

vishu-bh:
# Pull Request

## 🔗 Related Issue

Closes #5477

---

## 📝 Summary

Show the existing source-selection onboarding experience on the Dashboard when the current scope has neither virtual servers nor registered MCP servers.

The Dashboard performs lightweight existence checks for both resource types, uses the shared page loader while either request is pending, and shows an inline localized error if either request fails. Existing environments continue to see the standard Dashboard, while new environments get direct actions for connecting an MCP server or AI agent.

---

## 📏 Reviewability

- [x] This PR has one clear purpose
- [x] The linked issue is not labeled `triage`
- [x] Unrelated bugs or improvements are tracked in separate issues/PRs
- [x] Tests are included with the code they validate

---

## 🏷️ Type of Change

- [ ] Bug fix
- [x] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Chore (deps, CI, tooling)
- [ ] Other (describe below)

---

## 🧪 Verification

| Check | Command | Status |
|---|---|---|
| Client test suite | `npm run test:run` | 2,514 passed, 1 skipped |
| TypeScript | `npx tsc -b --pretty false` | Passed |
| Focused lint | `npx eslint src/pages/Dashboard.tsx src/pages/Dashboard.test.tsx` | Passed |
| Diff validation | `git diff --check` | Passed |

---

## ✅ Checklist

- [x] Code formatted
- [x] Tests added/updated for changes
- [x] Documentation not required
- [x] No secrets or credentials committed

---

## 📓 Notes (optional)

The existing `ConnectSourceCard` remains available on the Virtual Servers page. Dashboard onboarding uses `SourceSelection` and is hidden as soon as either an MCP server or virtual server exists.

PR #5807, opened 2026-07-23
5806 select source style adjustments

a-effort:
Style tweaks for https://github.com/IBM/mcp-context-forge/pull/5742 

- Fixed-size, no fill cards; selection/hover use a new `--selection border` token. Not using card component because of the amount of style overrides that would be needed for this use case. 
- Connect only on selected or hovered enabled cards; hide "coming soon" visually (keep in aria), but we can revisit if the functionality isn't available by the August preview. 
- Hover/selected text uses `foreground`; description uses `text-xs`
- Connect label i18n updated (Connect / Conectar); button hover uses more subtle visual change to avoid disabled button appearance 
- Typos fixed (LLM has trouble with "a" vs "an" usage)
- Tests updated

https://github.com/user-attachments/assets/a0bf1b80-b0e7-428a-b57d-1064ef848355


a-effort:
Typos fixed (LLM adds "an" incorrectly for some reason), associated tests updated.
typo fixed

gcgoncalves (approved):
Nice, clean update.

PR #5811, opened 2026-07-24
feat: add container image version bumper (make container-bump-image-versions)

### LINKED ISSUE #6282: [DEVOPS]: Container image version bumper (make container-bump-image-versions)
Adds scripts/container-bump-image-versions.sh which bumps pinned Red Hat UBI image tags in Containerfiles to the latest build tag within their current minor line, via the Red Hat Pyxis catalog API. Renovate would treat the two UBI_MINIMAL pin sites as independent updates; they must change atomically. Tracked by PR #5811.

jonpspri:
## What

Adds `scripts/container-bump-image-versions.sh`, exposed as `make container-bump-image-versions`, which bumps the pinned Red Hat UBI image tags in the repo's Containerfiles to the latest build tag **within their current minor line**.

## Why

The base-image pins (`UBI_BASE`, `NODEJS_IMAGE`, `UBI_MINIMAL`) are full build tags (`<minor>-<epoch-buildid>`) chosen for reproducibility, but nothing keeps them current — they drift until someone checks by hand. No Dependabot/Renovate config exists in the repo today, and Renovate would treat the two `UBI_MINIMAL` pin sites as independent updates, whereas they must change atomically.

## How it works

- Discovers tags anonymously via the Red Hat Catalog (Pyxis) API (`catalog.redhat.com/api/containers/v1/...`); `registry.access.redhat.com` allows anonymous access, unlike `registry.redhat.io`.
- Tag semantics are `<minor>-<epoch-buildid>`, so "latest" is the numeric max of the epoch component, restricted to each pin's current minor line. Bumping minors (e.g. 10.2 → 10.3) stays a deliberate, reviewed change.
- All lookups and validation complete before any file is written: a failed or malformed API response means zero files modified.
- `UBI_MINIMAL` is pinned in both `Containerfile` and `infra/wheels/Containerfile`; both are always updated together.
- The `ENABLE_FIPS=true` build path overrides these ARGs with UBI 9 images at build time and is intentionally not managed.

## Testing

- New bats suite `tests/scripts/container-bump-image-versions.bats` (11 tests: tag filtering, atomic two-file update, no-op idempotence, minor-line policy, zero-write failure paths, non-pinned-tag refusal), run via the new `make bats` target, which discovers all `tests/**/*.bats`.
- Verified end-to-end against the live Pyxis API: a run correctly identified and applied the pending 10.2-line updates, and a second run was a clean no-op. This surfaced (and fixed, with a stub regression lock) a real curl URL-globbing failure on the `sort_by=last_update_date[desc]` query parameter.
- The actual pin bump is intentionally not included here; it is being handled independently.

## Notes for reviewers

- bats-core is a new dev-only prerequisite (`brew install bats-core`); `make bats` fails gracefully with an install hint when absent.
- bats is currently not wired into `make test` or CI — worth deciding in review whether it should be.
- `.secrets.baseline` line-number drift in the Makefile commit is the detect-secrets hook's own regeneration.

---

Closes #6282 — container image version bumper.

### DIFF
--- a/scripts/container-bump-image-versions.sh
+++ b/scripts/container-bump-image-versions.sh
@@ -1,0 +1,140 @@
+#!/usr/bin/env bash
+# container-bump-image-versions.sh — bump pinned Red Hat UBI image tags in the
+# repo's Containerfiles to the latest build tag within their CURRENT minor line.
+#
+# Managed pins (full build tags only, e.g. 10.2-1784669047):
+#   Containerfile:               ARG UBI_BASE / NODEJS_IMAGE / UBI_MINIMAL
+#   infra/wheels/Containerfile:  ARG UBI_MINIMAL (must stay identical to root)
+#
+# Tags are discovered via the Red Hat Catalog (Pyxis) API, which allows
+# anonymous access:
+#   GET https://catalog.redhat.com/api/containers/v1/repositories/registry/
+#       registry.access.redhat.com/repository/<repo>/images
+#       ?page_size=100&sort_by=last_update_date[desc]
+#
+# Policy:
+#   - Stay within each pin's current minor line (10.2); minor bumps are a
+#     deliberate, reviewed change — not something a script should do.
+#   - The ENABLE_FIPS build path overrides these ARGs with UBI 9 images at
+#     build time (see Makefile container-build); those are not file-pinned
+#     and are intentionally not managed here.
+#
+# Safety:
+#   - All lookups and validation complete before any file is written; a
+#     failed or malformed API response means zero files modified.
+#   - Both UBI_MINIMAL pins are always updated together.
+#
+# Requires: curl, jq.
+# Test hooks: CONTAINERFILE_PATH / WHEELS_CONTAINERFILE_PATH override the
+# file locations (used by tests/scripts/container-bump-image-versions.bats).
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd "$SCRIPT_DIR/.." && pwd)"
+
+CONTAINERFILE_PATH="${CONTAINERFILE_PATH:-$REPO_ROOT/Containerfile}"
+WHEELS_CONTAINERFILE_PATH="${WHEELS_CONTAINERFILE_PATH:-$REPO_ROOT/infra/wheels/Containerfile}"
+
+PYXIS_BASE="https://catalog.redhat.com/api/containers/v1/repositories/registry/registry.access.redhat.com/repository"
+
+# ARG name -> Pyxis repository path
+MANAGED_ARGS=(UBI_BASE NODEJS_IMAGE UBI_MINIMAL)
+repo_for_arg() {
+    case "$1" in
+        UBI_BASE)     echo "ubi10" ;;
+        NODEJS_IMAGE) echo "ubi10/nodejs-24" ;;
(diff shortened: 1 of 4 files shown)


prakhar-singh1928:
@jonpspri with respect to this PR there are Two Blocking Issues

Issue 1 : repo_for_arg maintains a parallel hard-coded map that can drift from the actual image path already available in the Containerfile. The Pyxis repo path should be derived from ${value%:*} (which the script already extracts as image_of[$arg]) rather than a separate static case statement. If a deliberate minor-bump changes the path, the map would silently query the wrong endpoint.

Issue 2: sed -i '' in the bats test is macOS-only. GNU sed (Ubuntu — the likely CI runner) has no empty-suffix form; it will fail with "No such file or directory". The main script uses the portable sed -i.bak pattern already; the test should match it.

scripts/container-bump-image-versions.sh
```diff
@@ -0,0 +1,139 @@
...
+    for arg in "${!new_tag[@]}"; do
+        local files=("$CONTAINERFILE_PATH")
+        # UBI_MINIMAL is pinned in both Containerfiles; keep them identical.
+        if [ "$arg" = "UBI_MINIMAL" ]; then
+            files+=("$WHEELS_CONTAINERFILE_PATH")
+        fi
+        local f
+        for f in "${files[@]}"; do
+            if sed -i.bak "s|^ARG ${arg}=.*|ARG ${arg}=${image_of[$arg]}:${new_tag[$arg]}|" "$f"; then
+                rm -f "${f}.bak"
```
prakhar-singh1928: Potential `.bak` leak: `sed -i.bak` renames the original to `${f}.bak` before writing. For `UBI_MINIMAL`, `files` has two entries. If `sed` succeeds on the first file (its `.bak` is deleted by line 128), then `sed` fails on the second file (disk full, permissions changed mid-run, etc.), the second `.bak` is left on disk. `die` calls `exit 1` immediately with no cleanup.

Low probability, but easy to close with a trap:

```bash
declare -a _baks=()
trap 'rm -f "${_baks[@]}"' EXIT
# then in the loop:
sed -i.bak "..." "$f" && _baks+=("${f}.bak") && rm -f "${f}.bak"

File: scripts/container-bump-image-versions.sh [hunk 2]
```diff
@@ -0,0 +1,139 @@
...
+        echo "All image pins are up to date."
+        return 0
+    fi
+
+    # ---- Apply phase ----
+    for arg in "${!new_tag[@]}"; do
+        local files=("$CONTAINERFILE_PATH")
+        # UBI_MINIMAL is pinned in both Containerfiles; keep them identical.
+        if [ "$arg" = "UBI_MINIMAL" ]; then
+            files+=("$WHEELS_CONTAINERFILE_PATH")
```
prakhar-singh1928, line 127: **Optional Style/Consistency note:** `${!new_tag[@]}` iterates associative array keys in undefined order in bash. The `sed` calls for each ARG are independent today so the result is correct regardless of order, but this diverges from the plan and report loops which both iterate `"${MANAGED_ARGS[@]}"` in declaration order. If the apply logic ever gains an ordering dependency (e.g. a second ARG's sed pattern could match the first ARG's line), this becomes a latent bug.

Consistent, order-safe alternative:

```bash
for arg in "${MANAGED_ARGS[@]}"; do
    [[ -v new_tag[$arg] ]] || continue
    ...

prakhar-singh1928 (dismissed):
LGTM

prakhar-singh1928 (dismissed):

PR #5836, opened 2026-07-24
fix: Force install of rollup fallback on s390x

gcgoncalves:
Add `--force` to `s390x` rolldown `WASI-binding` install, as npm's platform check rejects the binding's mismatched `cpu: wasm32` field even with `--no-save`.


ja8zyjits (approved):
LGTM

Lang-Akshay (approved):
LGTM

PR #5917, opened 2026-07-27
fix(csrf): return 401 for unauthenticated write requests instead of misleading 403

### LINKED ISSUE #5743: [BUG]: Unauthenticated write requests (POST/PUT/DELETE) return misleading 403 CSRF_TOKEN_INVALID instead of 401
### 🐞 Bug Summary

Any unauthenticated **write** request (`POST`, `PUT`, `DELETE`) to a protected endpoint is rejected by `CsrfMiddleware` before authentication runs, so the client is told `CSRF validation failed` instead of `Authorization token required`.

Requests are still correctly denied — this is **not** an authentication bypass, and every endpoint remains fail-closed. The problem is diagnostic: the error names the wrong cause, sending people to debug cookies and CSRF tokens when the actual fix is supplying a bearer token.

> **Scope corrected after filing.** This was first observed on `POST /mcp` and filed as an MCP transport issue. A sweep across the API showed it is not MCP-specific — it affects every protected endpoint reached with an unsafe HTTP method. Title and body updated to match; see the measured blast radius below.

Found while running a live-gateway e2e against the `Release/v1.0.6` branch. Filing separately from that release PR since this is pre-existing middleware-ordering behavior, not a regression in 1.0.6.

---

### 🧩 Affected Component

- [x] `mcpgateway` - API
- [ ] Other (explain below)

---

### 🔁 Steps to Reproduce

1. Start a gateway with default auth settings:

   ```bash
   DATABASE_URL="sqlite:///./p.db" \
   JWT_SECRET_KEY="<secret>" \
   AUTH_ENCRYPTION_SECRET="<secret>" \
   AUTH_REQUIRED=true \
     uv run uvicorn mcpgateway.main:app --host 127.0.0.1 --port 18092
   ```

2. Send any write request with no `Authorization` header:

   ```bash
   curl -sS -i -X POST http://127.0.0.1:18092/servers \
     -H 'Content-Type: application/json' -d '{"server":{"name":"x"}}'
   ```

3. Compare against the same resource read with the same missing credential:

   ```bash
   curl -sS -i http://127.0.0.1:18092/servers
   ```

`GET` reports the missing token; `POST` reports a CSRF failure.

---

### 🤔 Expected Behavior

An unauthenticated request to a protected endpoint should report missing authentication consistently, regardless of HTTP method — `401` with `{"detail":"Authorization token required"}`, matching what the `GET` variants already return.

Returning `403 CSRF_TOKEN_INVALID` for a request that carried no credentials at all misattributes the failure. HTTP semantics also favour `401`: the client has not authenticated, rather than being authenticated and forbidden.

---

### 📓 Logs / Error Output

Measured on `Release/v1.0.6@29ca93c7e`, single gateway process, `AUTH_REQUIRED=true`, every request sent with **no `Authorization` header and no cookies**:

| Method | Path | Status | Body |
|---|---|---|---|
| `GET` | `/tools`, `/servers`, `/gateways`, `/resources`, `/prompts`, `/a2a` | `401` | `{"detail":"Authorization token required"}` |
| `POST` | `/tools` | `403` | `{"detail":"CSRF validation failed","code":"CSRF_TOKEN_INVALID"}` |
| `POST` | `/servers` | `403` | same |
| `POST` | `/gateways` | `403` | same |
| `POST` | `/resources` | `403` | same |
| `POST` | `/prompts` | `403` | same |
| `POST` | `/a2a` | `403` | same |
| `POST` | `/mcp` | `403` | same |
| `POST` | `/servers/{id}/mcp` | `403` | same |
| `PUT` | `/tools/{id}` | `403` | same |
| `DELETE` | `/tools/{id}`, `/servers/{id}`, `/gateways/{id}` | `403` | same |
| `POST` | `/auth/email/login` *(exempt, control)* | `401` | real auth failure — correct |
| `GET` | `/health` *(exempt, control)* | `200` | correct |

Gateway log line:

```
CSRF token missing for POST /servers
```

Cause is the ordering in `mcpgateway/middleware/csrf_middleware.py`. Step 2 skips safe methods, step 4 exempts bearer-token requests, and step 5 rejects anything left without a CSRF header:

```python
# 2. Skip safe methods (GET, HEAD, OPTIONS, TRACE)
if request.method in SAFE_METHODS:
    return await call_next(request)

...

# 4. Skip Bearer token requests (not vulnerable to CSRF)
auth_header = get_auth_header_value(request.headers) or ""
if _extract_bearer_token(auth_header):
    return await call_next(request)

# 5. Extract CSRF token from header. ...
csrf_token = request.headers.get(settings.csrf_token_name)

if not csrf_token:
    logger.warning(f"CSRF token missing for {request.method} {request.url.path}")
    return JSONResponse(status_code=403, content={"detail": "CSRF validation failed", "code": "CSRF_TOKEN_INVALID"})
```

A request with **no credentials at all** carries no bearer token, so it falls past the step-4 exemption into the step-5 rejection and never reaches the auth layer. The `GET` variants escape only because step 2 skips safe methods. So the split is precisely **unauthenticated safe method vs unauthenticated unsafe method** — not anything specific to a resource type or transport.

The authenticated path is unaffected throughout: with a valid bearer token these endpoints behave normally (verified in the same run — `POST /servers` → `201`, `POST /mcp` → `200` with a completed MCP handshake).

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Version or commit | `1.0.6` / `Release/v1.0.6@29ca93c7e` |
| Runtime | Python 3.12.3, uvicorn (single process) |
| Platform / OS | Ubuntu (WSL2, kernel 6.6.87.2) |
| Container | none |

---

On the fix — the blast radius rules out the narrow option. Adding paths to `csrf_exempt_paths` would mean enumerating essentially every write endpoint in the API, and would also switch off genuine CSRF protection for cookie-authenticated callers on those paths.

The targeted change is to let CSRF fall through when the request carries **no credentials at all** — no bearer token, and no `jwt_token`/`access_token` cookie — so the auth layer produces the `401`. CSRF protection is only meaningful against ambient cookie-borne credentials, which such a request does not have by definition, so this should not weaken the control. Requests that do carry a cookie would continue through CSRF validation exactly as today.

Related but distinct prior issues, all cookie/origin-specific rather than the unauthenticated case: #5151 (HttpOnly `jwt_token` cookie), #5325 (non-email-auth admin sessions), #3431 (origin validation in multi-pod).

Possibly relevant to #5215 (LLM Chat cannot connect to a same-gateway virtual server) — that one reports a `401` from `/servers/{id}/mcp` rather than this `403`, so it looks like a different path, but the two touch the same endpoint and may be worth checking together.

msureshkumar88:
## Summary
- `CSRFMiddleware` ran before the auth layer, so an unsafe-method request (POST/PUT/DELETE) with no credentials at all fell into the CSRF-missing check and returned `403 CSRF_TOKEN_INVALID`, rather than reaching the route-level auth dependency that returns `401 Authorization token required` — the response `GET` variants of the same endpoints already give.
- Adds a fall-through: a request with no `Authorization` header (of any scheme) and no `jwt_token`/`access_token` cookie has nothing for CSRF to protect, so it now passes through to the auth layer for the correct `401`.
- Requests carrying a bearer token or an auth cookie are unaffected and continue through CSRF validation exactly as before — this is a diagnostic fix, not a change to the fail-closed security posture.

Closes #5743


gandhipratik203:
Nice, targeted fix — the guard is well-placed (after the bearer exemption, before the CSRF-token check), and updating the existing tests to carry a `jwt_token` cookie was the right call, otherwise they'd have quietly stopped exercising CSRF at all.

One thing worth considering before this lands:

**Proxy-header auth would lose CSRF protection.** Under `MCP_CLIENT_AUTH_ENABLED=false` + `TRUST_PROXY_AUTH(_DANGEROUSLY)=true`, identity comes from `X-Authenticated-User` (`verify_credentials.py:226`, `rbac.py:196-200`). Those requests have no `Authorization` header and no auth cookie, so step 4b lets them through — but the identity is still ambient, backed by the proxy's own session cookie. Cross-site POST → proxy injects the header → runs as the victim.

The mode is off by default and gated behind the `_DANGEROUSLY` flag, so it's narrow, but it's the one case where "no credentials at all" isn't quite true. Would be good to exclude it from the fall-through and add a deny-path test alongside.

msureshkumar88:
Good catch, thanks for flagging this — pushed a fix in 5a2af9d.

Excluded trusted-proxy-header identity from the credential-less fall-through: when `is_proxy_auth_trust_active()` is true and the configured `proxy_user_header` is present, the request still goes through CSRF validation instead of skipping to the auth layer. Added a deny-path test (`test_proxy_header_identity_still_requires_csrf`) plus a positive control confirming a stray header is ignored when proxy trust isn't active (`test_proxy_header_present_but_trust_inactive_falls_through`).

One thing worth calling out for visibility: in proxy mode, CSRF validation was already failing closed before this PR too — `AuthContextMiddleware` doesn't populate `request.state.user`/`jti` for proxy identities, and the fallback JWT check has no token to verify, so step 6 (`no user context`) was returning 403 on every non-exempt write regardless of a valid CSRF token. This fix restores that pre-existing behavior for proxy mode rather than adding real CSRF token validation for it — the proxy-mode CSRF binding gap is a separate, pre-existing issue outside this PR's scope.

gandhipratik203 (dismissed):
Proxy-trust exception looks good, and the trust-inactive control test is the right thing to have alongside it. LGTM.

ja8zyjits (dismissed):
LGTM

ja8zyjits (approved):
LGTM

PR #5919, opened 2026-07-27
chore: update python dependencies

prakhar-singh1928:
# Pull Request

## 🔗 Related Issue

Closes #

---

## 📝 Summary

Routine Python dependency update cycle following the release management process (§2). Updates floor pins across the root `pyproject.toml`, all sub-project `pyproject.toml` files, and all `requirements.txt` files to latest available versions. Advances CPEX `exclude-newer-package` dates to today and resolves a pre-existing `langgraph`/`websockets` upstream conflict.

**Root `pyproject.toml` — runtime deps:**

| Package | Before | After |
|---|---|---|
| `fastapi` | `>=0.139.0` | `>=0.140.0` |
| `filelock` | `>=3.29.4,<3.31.0` | `>=3.32.0` |
| `jq` | `>=1.11.0` | `>=1.12.0` |
| `prometheus-fastapi-instrumentator` | `>=8.0.2` | `>=8.1.0` |
| `prometheus_client` | `>=0.25.0` | `>=0.26.0` |
| `sse-starlette` | `>=3.4.5` | `>=3.4.6` |
| `starlette-compress` | `>=1.7.1` | `>=1.8.0` |
| `typer` | `>=0.26.8` | `>=0.27.0` |
| `uvicorn[standard]` | `>=0.49.0` | `>=0.51.0` |

**Root `pyproject.toml` — extras:**

| Package | Before | After |
|---|---|---|
| `opentelemetry-*` | `>=1.43.0` | `>=1.44.0` |
| `grpcio` / `grpcio-reflection` / `grpcio-tools` | `>=1.81.1` | `>=1.83.0` |
| `protobuf` | `>=6.33.6` | `>=7.35.1` |
| `langchain-core` | `<=1.4.7` | `<=1.5.1` |
| `langchain-openai` | `<=1.2.1` | `<=1.4.1` |
| `langsmith` | `>=0.8.18` | `>=0.10.10` |

**Sub-projects / mcp-servers / plugins:** `fastmcp`, `mypy`, `ruff`, `uv`, `scipy`, `numpy`, `pandas`, `openai`, `locust`, `pytest-env`, `matplotlib`, `plotly`, `pyarrow`, `sentence-transformers`, `RestrictedPython`, `anthropic`, `boto3/botocore` and others updated across all sub-projects.

**`docs/requirements.txt`:** `certifi`, `GitPython`, `mkdocs-material`, `pandas`, `regex`, `soupsieve`, `tzdata` and others updated.

**`tests/load/requirements.txt`:** `faker`, `tqdm` updated.

**`uv.lock`:** Full resolver output regenerated.

**`[tool.uv.exclude-newer-package]` changes:**
- All CPEX package dates advanced to `2026-07-27T23:59:59Z`
- `langgraph` pinned to `2026-05-12T23:59:59Z` — `langgraph>=1.2.1` requires `langgraph-sdk>=0.4.2` which caps `websockets<16`, conflicting with `websockets>=16.0` in dev deps; `<=1.2.0` ceiling preserved in `llmchat` extra
- 12 soak-window overrides added at `2026-07-27T23:59:59Z` for packages bumped to versions published within the 10-day window; to be removed after 2026-08-03

---

## 📏 Reviewability

---

## 🏷️ Type of Change

---

## 🧪 Verification

| Check | Command | Status |
|---|---|---|
| Dependency resolution | `uv lock --upgrade` | ✅ |
| Dev install | `make install-dev` | ✅ |
| Vulnerability audit | `make pip-audit` | ✅ |
| Container build | `make docker-prod DOCKER_BUILD_ARGS="--no-cache"` | ✅ |
| Unit tests | `make test` | ⏳ |

---

## ✅ Checklist

---

## 📓 Notes

**`langgraph` upstream conflict:** `langgraph>=1.2.1` transitively requires `langgraph-sdk>=0.4.2` which pins `websockets>=14,<16`, conflicting with the dev dependency `websockets>=16.0`. The `llmchat` extra ceiling is kept at `langgraph<=1.2.0` (uses `langgraph-sdk<0.4.0` which has no websockets cap) until upstream resolves the conflict.

**Soak-window overrides:** 12 packages were bumped to versions published within the past 10 days (the global `exclude-newer = "10 days"` soak window). Per-package overrides at `2026-07-27T23:59:59Z` were added to unblock resolution. These should be removed after 2026-08-03.


cafalchio (dismissed):
LGTM

cafalchio (dismissed):
LGTM

msureshkumar88 (changes requested):
Thanks for keeping the dependency cycle moving — this is a lot of careful mechanical work, and a few things are clearly right: `uv lock --check` passes cleanly, manifest coverage is complete (`plugins/requirements.txt` and the maturin crate genuinely have nothing to bump, and `mcp-servers/templates/` is the documented skip), no Alembic migration was touched (correct — nothing here warrants a schema change), and the `langgraph` conflict write-up in the description is genuinely useful context that saved me a lot of digging.

A few things I'd like to work through before this lands. Grouped by what I think must change vs. what's optional.

---

## Blocking

### 1. `filelock` upper bound removal reopens the #5654 regression

`pyproject.toml` — `"filelock>=3.29.4,<3.31.0"` → `"filelock>=3.32.0"`.

That ceiling isn't a stale artifact — it landed 11 days ago in 3efce136d (#5654) as incident response. filelock 3.30.0 added a fork-safety guard that raises when a lock built before the `gunicorn --preload` fork is reused from a child PID; `GatewayService`'s health-check loop caught it and re-entered with no delay, producing an unbounded busy loop that starved workers of CPU (CI jobs hanging to workflow timeout, logs into the millions of lines).

Two details make this riskier than a normal ceiling removal:

- `make install` is `uv pip install .` (Makefile:298), not `uv sync` — it resolves live against PyPI and ignores `uv.lock` entirely. The range in `pyproject.toml` is the only thing controlling what a fresh install gets.
- The floor jumps straight to 3.32.0, skipping 3.30.x and 3.31.x — exactly the releases the ceiling was guarding — and 3.32.0 is itself six days old and soak-bypassed (see #3 below).

The in-code PID-rebuild fix from #5654 lives only in `gateway_service.py`; `primary_worker.py`, `leader_election.py`, and `bootstrap_db.py` also construct `FileLock` (they acquire lazily / post-fork, so they're likely fine, but that hasn't been re-verified against 3.32.0).

Two ways forward, either works:

- **Cheap:** restore an explicit ceiling — `"filelock>=3.29.4,<3.33.0"` — and update the comment to describe the actual range.
- **Better, if you have the appetite:** add a regression test that exercises the fork path (`gateway_service.py:5351-5353`) and keep the open floor, with the comment pointing at the test. That would let future ceilings come off on evidence rather than assumption.

### 2. Three comments now contradict the pins they annotate

Worth fixing because these are what the next reviewer trusts:

- `filelock` — comment still says "ceiling caps unknown future releases" with no ceiling present.
- `protobuf>=7.35.1, # grpcio-reflection 1.78.x caps at <7.0.0` — directly contradicts the pin. The lock resolves `grpcio-reflection 1.83.0` alongside `protobuf 7.35.1`, so the cap is genuinely gone; the comment should say it was lifted rather than restate a dead constraint.
- `prometheus_client>=0.26.0, # 0.25.0 exceeds 10-day min-release-age policy` — refers to a version no longer in the file.

### 3. The 12 new soak-window overrides

`exclude-newer = "10 days"` is a supply-chain control rather than a style preference — it's the window that lets a compromised or yanked upstream release surface before it reaches the build, and `release-management.md:145` shows the documented flow enforcing it explicitly.

Two things I'd like to see change:

**Justification.** Every pre-existing override in that block (`cpex-url-reputation`, `cpex-rate-limiter`, `cpex-encoded-exfil-detection`, `cpex-pii-filter`, `cpex-retry-with-backoff`, `cpex-secrets-detection`) carries a comment naming the exact release, the exact test that needs it, and the phrase "Vetted ahead of the usual window for this reason." The 12 new entries are bare dates. That's the difference between a reviewed exception and an unreviewed bypass, and it's the thing a future reader has no way to reconstruct.

**Necessity.** From the lock's upload timestamps, the bypass exists purely because the floors were raised to releases younger than 10 days — `filelock` 3.32.0 (2026-07-21), `grpcio` 1.83.0 (2026-07-23), `prometheus-client` 0.26.0 (2026-07-24). Deferring those specific bumps to next cycle gets the same "deps are current" outcome with the control left intact. Is there a reason these particular versions are needed now? If so, that reason is exactly what belongs in the comment.

Separately: the six existing `cpex-*` overrides were advanced to today. Those dates were narrow, single-release exceptions tied to a specific vetted artifact, and their comments still describe them that way ("0.3.6 (2026-07-07) is the first release that emits result.metadata..."). Moving them to today widens each one to admit any cpex release from the last ~20 days. If that's intended, the comments need updating; if the goal is only "allow the already-vetted release," the original dates should stay.

The "remove after 2026-08-03" note in the description is good — could we get a tracking issue for it? Nothing in the repo enforces that date today.

### 4. The `dependency-review.yml` change

Two separate things here.

**Scope:** this modifies a license/security gate rather than a dependency version, and it isn't mentioned in the description while "This PR has one clear purpose" is checked. Loosening a CI control feels like it deserves its own PR and its own reviewer — probably whoever owns license policy. Would you be open to pulling it out?

**Correctness:** the comment says the four entries mirror the allowlist in `license-policy.toml`, but that file has seven. Cross-referencing against the workflow's own `deny-licenses` list (LGPL and MPL are deliberately *not* denied):

- `astroid` and `chardet` are LGPL, so they aren't denied anyway — correctly omitted.
- `pyphen` is GPL, *is* denied, and isn't in the allowlist. It's a docs-toolchain transitive dep, and since this PR touches `docs/requirements.txt`, the gate could still fail on it — so the exclusion may not fully solve the problem it was added for.

Minor notes on the same hunk: the purl entries have no version qualifier, so they exempt all versions indefinitely, including any future relicense. And the removed comment block documented *how* to add an exemption and why each one narrows the policy — that seemed net-useful and I'd suggest keeping it.

---

## Suggestions

### `protobuf` 6 → 7 — could this be a range instead of a floor?

This one has the widest blast radius of anything in the PR. Inside the repo it's low risk (no `.proto` or `*_pb2.py` files, and `grpc_service.py` / `translate_grpc.py` use dynamic reflection rather than generated stubs, so there's no codegen to regenerate).

The concern is downstream: the package publishes at `1.0.6`, and a floor pin in a published distribution is a hard constraint on every consumer's resolver. `protobuf` is transitively pulled by `a2a-sdk`, `google-api-core`, `googleapis-common-protos`, and the OTel gRPC exporter — so anyone installing `mcp-contextforge-gateway[grpc]` into an environment with a `protobuf<7` cap gets a hard `ResolutionImpossible`.

Since nothing in the codebase actually requires the 7 API, `"protobuf>=6.33.6,<8"` would keep protobuf-6 consumers working while still permitting 7. That's the single largest blast-radius reduction available here.

### `langgraph` is pinned twice

The `exclude-newer-package` entry at `2026-05-12` and the existing `langgraph<=1.2.0` ceiling in the `llmchat` extra express the same constraint through two mechanisms — two places to remember to unwind, and they can drift. My suggestion would be to keep the extra's ceiling and drop the `exclude-newer` entry.

Worth noting either way that this freezes an optional dependency at a two-and-a-half-month-old release with no upgrade path. The `llmchat` extra caps the exposure, but a tracking issue for the `websockets<16` conflict (referenced from the comment) would stop it rotting silently.

### The `mcp_eval_server` operator flips deserve a callout

In the `rest` and `all` extras, four pins flip direction rather than move version:

```
"fastapi<=0.136.1"           →  "fastapi>=0.140.0"
"python-multipart<=0.0.28"   →  "python-multipart>=0.0.32"
"uvicorn[standard]<=0.47.0"  →  "uvicorn[standard]>=0.51.0"
"ibm-watsonx-ai<=1.5.11"     →  "ibm-watsonx-ai>=1.6.0"
```

Tracing it back, those `<=` bounds came from e0967e65c (#4909), which reversed `fastapi>=0.136.1`, `python-multipart>=0.0.26`, `uvicorn[standard]>=0.46.0`, and `ibm-watsonx-ai>=1.5.10` in one pass — near-identical versions, operators flipped. So this PR is most likely *restoring* the original intent, and it's a genuine improvement (those ceilings were holding `fastapi` and `python-multipart` below releases carrying fixes).

The only ask is that it be stated in the description rather than folded into "and others updated" — as written it's hard to distinguish from an accidental repeat of the same bulk-edit. Contrast with the root `pyproject.toml`, where `langchain-core<=1.4.7 → <=1.5.1` correctly kept its `<=`.

---

## Minor

- **Linked issue.** The description has a bare `Closes #`, and neither commit references one. Happy to help open a `[CHORE]` issue if that's easiest.
- **`make test` is still ⏳** in the verification table. Given `fastapi`, `uvicorn`, `sse-starlette`, and `starlette-compress` all moved on the request path, the e2e gate (`make docker-nuke docker-prod-rust testing-up` then `make test-mcp-protocol-e2e test-mcp-rbac test-protocol-compliance`) would be worth a run too.
- **No job installs `[grpc]`**, so the protobuf major crossing has resolution proof but no runtime proof. A smoke check that imports `grpc_reflection.v1alpha` and `grpc_service` would close that.
- **`CHANGELOG.md`** — the header declares SemVer adherence, and the `protobuf` major floor plus the `filelock` bound removal both read as `### Changed` entries.
- **A constraint-lint test** would have caught items 2, 3, and the `mcp_eval_server` flip mechanically — something that parses `pyproject.toml` and asserts every `exclude-newer-package` entry has a preceding justification comment, and that operator direction doesn't silently invert against the merge base. Optional, but it'd make future cycles of this PR much cheaper to review.
- **`release-management.md` §2** describes the bump flow but has no guidance on when a soak-window override is acceptable, what justification it needs, or who removes it. Since this PR adds twelve, that might be the most valuable doc addition here.

---

Happy to pair on any of these, particularly the `filelock` regression test if you'd rather go that route than restore the ceiling. Nothing above is a disagreement with doing the bump cycle itself — it's mostly about keeping the guardrails that previous incidents put in place.

prakhar-singh1928:
Blocking 1 & 2 — filelock ceiling + stale comments

The ceiling was intentionally removed rather than restored. Checking the filelock changelog: 3.30.x–3.32.x have zero runtime changes — CI matrix additions, packaging fixes, platform support (Termux/Android), ENOTSUP tolerance. None touch the fork-safety guard code. The in-code PID-rebuild fix in gateway_service.py:5351 handles any future release. primary_worker.py and bootstrap_db.py both create FileLock post-fork so they're unaffected. The comment has been rewritten to document all of this explicitly so future readers don't have to re-derive it.

The protobuf and prometheus_client comments have also been corrected.

Blocking 3 — soak-window overrides

I will remove them during next release no comment is required

Blocking 4 — dependency-review.yml

The root cause was prospector[with_everything] pulling in pylint-django, pylint-celery, pylint-plugin-utils, and pylint as core (non-optional) deps of prospector regardless of extras chosen. Since make prospector already installs prospector on-demand via uv pip install -q prospector[with_everything] and it's not in any CI workflow, removed it from [dependency-groups.dev] entirely. This drops all 16 GPL packages from uv.lock (376 → 360 packages) and makes the dependency-review.yml change unnecessary — reverted to its original state. No allowlist exemptions needed.

msureshkumar88 (changes requested):
Thanks for the two follow-up commits — `dependency-review.yml`/prospector removal is a clean root-cause fix (better than the exemption it replaces), and the comment corrections on `filelock`/`protobuf`/`prometheus_client` are all accurate now. Nice work closing those out.

Two items from the earlier review still feel open to me, functionally rather than just cosmetically:

## Blocking

### 1. `filelock` ceiling
The comment now explains *why* the floor was raised, which is genuinely useful, but the underlying ask was for a safety net, not just documentation. `make install` runs `uv pip install .` (Makefile:298), which resolves live against PyPI and ignores `uv.lock` — so the range in `pyproject.toml` is still the only thing bounding what a fresh install gets. There's an existing fork-safety test (`test_gateway_service_extended.py:403-459`), but it mocks `FileLock` entirely, so it verifies the PID-rebuild logic in isolation rather than the actual 3.30–3.32 guard behavior the comment describes.

Either of the original options would close this:
- Restore an explicit ceiling (e.g. `filelock>=3.32.0,<3.33.0`), or
- Add a test that exercises the real `filelock` package's fork-guard behavior rather than a mock.

### 2. The 12 new soak-window overrides still lack justification
Understood on "will remove next release" for the general cleanup — that's reasonable. But as written, these 12 entries are bare dates while every pre-existing override in the block (`cpex-url-reputation`, `cpex-rate-limiter`, etc.) names the exact release and the exact test it unblocks. That asymmetry is what makes this hard to review later: a future reader can't tell a justified exception from an unreviewed bypass just by looking at the file.

Also worth a look: the 6 pre-existing `cpex-*` overrides were advanced from narrow, single-release dates (e.g. `2026-07-07`) to today (`2026-07-27`), but their comments still describe the original specific release. As it stands the comment and the enforced window no longer agree — the window now admits any `cpex-*` release from roughly the last 20 days, not just the one named in the comment.

Two ways I'd be happy with:
- Add a one-line justification per new entry, matching the existing `cpex-*` style, or
- Drop the overrides for the packages that are only within the 10-day window because of this bump, and defer those specific version bumps to the next cycle.

## Not blocking, just flagging again
- `protobuf>=7.35.1` still has no ceiling — `>=6.33.6,<8` would avoid a hard `ResolutionImpossible` for downstream consumers still on protobuf 6, without giving up anything this repo needs.
- `langgraph` is still pinned both via `exclude-newer-package` and the `llmchat` extra ceiling — fine either way, just two places to keep in sync.

Happy to pair on the filelock test if that's the preferred route over the ceiling.

ja8zyjits (dismissed):
LGTM

msureshkumar88:
Following up on blocking 2 (soak-window overrides) — no separate tracking issue needed for the "remove after 2026-08-03" part; the release-management §2 cycle re-touches this block every release anyway, so that's handled.

The other half still stands though: the 12 new entries are bare dates with no per-package justification, unlike the existing `cpex-*` style (release + test it unblocks). A one-line comment per entry, or just deferring the ≤10-day-old bumps to next cycle, would close it out.

msureshkumar88 (dismissed):
## E2E dependency-update verification

I ran an environment-level verification pass against this PR's head commit (`c50702cb8`, rebased clean onto `origin/main` — 0 commits behind), in an isolated worktree + fresh venv, mirroring the checks in the PR's own Verification table plus a few extra checks on the two riskiest bumps (protobuf major version, langgraph/websockets pin conflict).

### Config used
- Isolated `git worktree` checked out at the PR head, separate `.venv-pr5919-verify` (not the repo's shared `.venv`)
- Python 3.12.3, `uv 0.10.2`
- Script: `scripts/verify-pr-5919-e2e.sh` (ran locally, not pushed to the branch)

### Commands run & results

| # | Command | Result |
|---|---|---|
| 1 | `uv lock --check` | ✅ pass — `uv.lock` matches `pyproject.toml`, no drift |
| 2 | `uv pip install --group dev '.[plugins]'` | ✅ pass — dev install resolves cleanly |
| 3 | `pip-audit --skip-editable` | ✅ pass — 0 known vulnerabilities; only expected editable-install skip |
| 4 | `uv pip install '.[grpc]'` + import smoke test (`grpc`, `grpc_reflection`, `google.protobuf`) | ✅ pass — confirms `grpcio 1.83.0` resolves with `protobuf 7.35.1` (the `<7.0.0` cap comment in the diff checks out) |
| 5 | `uv pip install '.[llmchat]'` + `pip show` | ✅ pass — resolves to `langgraph==1.2.0` + `websockets==16.1.1` together, confirming the stated `langgraph`/`websockets` conflict fix actually works |
| 6 | Targeted `pytest`: `test_config.py`, `test_gateway_service.py`, `test_oauth_manager.py` (×2) | ✅ pass |
| 7 | `pytest -q tests/unit` (full suite) | ✅ pass — 20,133 collected, 0 failed/error, 99 skipped / 2 xfail (all pre-existing, documented reasons unrelated to this PR — e.g. `TEST_POSTGRES_URL` not set, `altk`/`dagger` not installed) |

I also manually verified two of the diff's inline claims against source rather than taking them at face value:
- The `filelock` ceiling comment references `gateway_service.py:5351` rebuilding the lock per-PID post-fork — confirmed accurate. Found one more `FileLock` call site the comment doesn't mention (`leader_election.py:165`), but it's also lazily created inside the async loop (post-fork), so it's not at risk from the 3.30+ fork-safety guard either.
- No `.py` files are touched by this PR — the whole diff is `pyproject.toml` / `requirements.txt` / `uv.lock` / `Makefile` / workflow YAML, so lint gates (ruff/pylint/interrogate) aren't applicable here.

### Not run (and why)
- Full `docker-nuke` + `docker-prod-rust` rebuild and the live-gateway MCP protocol/RBAC e2e suite — skipped because `docker-nuke` would tear down other already-running containers in this environment unrelated to this review (destructive blast radius), and a `--no-cache` image rebuild wasn't worth the time cost here given the change has zero application-code surface.
- `detect-secrets-scan` tool invocation — skipped running the tool itself, but I read every line of the diff manually; it's exclusively version pins, dates, and comments, no secret-shaped content.

### One non-blocking note
This PR's 6 commits bundle a bit more than the description covers: 3 commits remove `prospector` tooling entirely (`Makefile`, `AGENTS.md`, `SECURITY.md`, `MANIFEST.in`) and one adjusts the `dependency-review.yml` license allowlist (later self-reverted in the same PR once prospector's GPL-licensed pylint plugins were removed, so the net diff there is just a comment wording fix). None of this is a problem functionally — I verified the end state is internally consistent — but the PR body doesn't mention the prospector removal, so it's worth a one-line mention there for anyone skimming the description later.

**Verdict: approving.** Lock/install/audit/tests all pass cleanly against the updated pin set, including the two bumps with the most conflict risk (protobuf major version, langgraph/websockets).

msureshkumar88 (approved):
LGTM

PR #5664, opened 2026-07-16
fix(resources): revert resource name uniqueness constraint from #5158

madhu-mohan-jaishankar:
## 🔗 Related Issue

Closes [#78843](https://github.ibm.com/WatsonOrchestrate/wo-tracker/issues/78843)

---

## 📝 Summary

PR #5158 introduced a DB-level unique index and application-level checks enforcing that resource **names** must be unique within the same `(team_id, owner_email, gateway_id)` scope. This assumption conflicts with the MCP specification, which defines resources as uniquely identified by **URI** — `name` is a human-readable display label and has no uniqueness requirement.

The constraint breaks a legitimate and common usage pattern where multiple task-specific resource instances share the same resource type name but each have a distinct URI (e.g. `gridEntityGroundTruth` with URI `grid-entities://{taskId}.json` — one row per task). WXO's production deployment had 17 such resource groups, causing the migration pre-flight check to raise a `RuntimeError` and crash-loop the CF container on every startup.

**Changes in this PR:**

- **Migration `279184dfd71d`** (in-place): Removes the pre-flight duplicate-name check and the `CREATE UNIQUE INDEX` statements. Replaces them with drop-if-exists logic so any environments that already applied the old migration have the indexes cleaned up automatically on next upgrade.
- **`db.py`**: Removes `uq_team_owner_gateway_name_resource` and `uq_team_owner_name_resource_local` indexes from `Resource.__table_args__`. The pre-existing URI uniqueness constraints (`uq_team_owner_gateway_uri_resource`, `uq_team_owner_uri_resource_local`) are untouched.
- **`resource_service.py`**: Removes `ResourceNameConflictError` class and all name-conflict checks in `register_resource()`, `bulk_import_resources()`, and `update_resource()`. URI-based conflict detection (`ResourceURIConflictError`) is preserved.
- **`main.py`**, **`admin.py`**: Removes import and exception handlers for `ResourceNameConflictError`.
- **Tests**: Removes name-conflict test cases; fixes two URI-conflict tests whose mock setup expected a preceding name-check DB call that no longer exists.

---

## 📏 Reviewability

- [x] This PR has one clear purpose
- [x] The linked issue is not labeled `triage`
- [x] Unrelated bugs or improvements are tracked in separate issues/PRs
- [x] Tests are included with the code they validate
- [x] If AI-assisted, I understand and can explain the generated changes

---

## 🏷️ Type of Change

- [x] Bug fix

---

## 🧪 Verification

[...]

PR #5920, opened 2026-07-27
fix(resources): show specific message for duplicate resource URI conflicts

### LINKED ISSUE #4991: [BUG]:  Resource creation shows generic HTTP 409 error instead of meaningful duplicate resource message

msureshkumar88:
## 🔗 Related Issue

Closes #4991

---

## 📝 Summary

Creating a resource whose URI already exists showed `Failed to add Resource (HTTP 409)` in the Admin UI, with no indication of which field was at fault. Two independent defects each produced that exact string on their own, so both needed fixing:

**1. `ErrorFormatter.format_database_error` only understood SQLite.** It matched `UNIQUE constraint failed: <table>.<column>` plus a list of column paths. PostgreSQL reports `duplicate key value violates unique constraint "uq_team_owner_gateway_uri_resource"`, which matched nothing and fell through to the generic `Unable to complete the operation. Please try again.` SQLite escaped this only by accident — its multi-column message happens to contain the `resources.uri` substring.

**2. `extractApiError` could not read a nested `detail`.** It handled `error.message`, `detail` as a string and `detail` as an array, but not `detail` as an object — which is exactly what FastAPI emits for `HTTPException(detail={...})`. It therefore returned the caller's fallback, i.e. the reported string. This affected every admin form (servers, tools, prompts, gateways), not just resources, so the fix is deliberately generic.

**Names are not the conflict.** #5664 reverted the name-uniqueness constraint added by #5158, because the MCP specification treats `uri` as the identifier and `name` as a human-readable display label — and the constraint broke a legitimate pattern of task-scoped resources sharing a type name. The new messages say that rule out loud, and the add-resource form attributes the error inline to the URI field so a duplicate URI can no longer be misread as a duplicate name. Helper text under the Name input states that names need not be unique.

---

## 🐛 Additional fixes in the same paths

- **Private visibility skipped the friendly error.** `register_resource` / `update_resource` pre-checked URIs only for `public` and `team`. `private` is a valid visibility and is exposed as a radio on the add-resource form, so private duplicates bypassed `ResourceURIConflictError` and surfaced as a raw `IntegrityError`. Both now have a `private` branch scoped to the `(team_id, owner_email, gateway_id, uri)` constraint, mirroring the precedence used when the row is constructed.
- **`PUT /resources/{id}` lost exceptions to 500s.** It caught neither `ResourceValidationError` nor plain `ResourceError`. Now 422 and 400 respectively, matching `POST /resources`, and ordered after `ResourceURIConflictError` since both are `ResourceError` subclasses.
- **Dead code from the #5664 revert removed** — the `uq_team_owner_*_name_resource` and `resources.name` mappings in `error_formatter.py`, and a stale comment in `resource_service.py` referring to a deleted name check.

No schema change, no migration, and name uniqueness is not reintroduced at any layer. A regression test asserts that a duplicate name with a distinct URI still succeeds, for all three visibilities, against a real in-memory session so the actual DB constraints participate.

---

## ⚠️ Behaviour change worth flagging

`ResourceLockConflictError` on update now returns **400** instead of escaping as a 500, because it subclasses `ResourceError`. This matches the existing create path, which also has a bare `except ResourceError → 400` with no lock-conflict case. Arguably lock conflicts should be 409 everywhere (`admin_set_resource_state` already maps them that way) — happy to follow up separately if you agree.

---

## 📏 Reviewability

---

## 🏷️ Type of Change

- [x] Bug fix

---

## 🧪 Verification

| Check | Command | Status |
|---|---|---|
| Full pytest suite | `make test` | ✅ 20869 passed, 799 skipped, 2 xfailed, 0 failed |
| Doctests | `make doctest` | ✅ 1197 passed, 55 skipped |
| JS unit tests | `make test-js` | ✅ 51 files, 2298 tests |
| Lint | `make ruff` | ✅ |
| Static analysis | `pylint` on changed files | ✅ 9.99/10 (sole finding pre-exists on `main`) |
| Docstring coverage | `interrogate` on changed files | ✅ 100% |
| Secrets | `make detect-secrets-scan` | ✅ no new findings (baseline churn is timestamp + 2 line shifts) |

---

## ✅ Checklist

### DIFF
--- a/mcpgateway/admin_ui/formValidation.js
+++ b/mcpgateway/admin_ui/formValidation.js
@@ -4,6 +4,23 @@
 
 import { validateInputName, validateUrl } from "./security.js";
 
+/**
+ * Tailwind utility classes applied to an input to signal a field-level error.
+ *
+ * Shared so that client-side validation and server-side error attribution
+ * (formSubmitHandlers.js) render identically. ``modals.js`` clears this exact
+ * set when resetting a form *inside a modal*; forms rendered inline on a page
+ * must clear it themselves.
+ *
+ * @type {ReadonlyArray<string>}
+ */
+export const FIELD_ERROR_CLASSES = [
+  "border-red-500",
+  "focus:ring-red-500",
+  "dark:border-red-500",
+  "dark:ring-red-500",
+];
+
 export const setupFormValidation = function () {
   // Add validation to all forms on the page
   const forms = document.querySelectorAll("form");
@@ -39,12 +56,7 @@ export const setupFormValidation = function () {
         );
         if (!validation.valid) {
           this.setCustomValidity(validation.error);
-          this.classList.add(
-            "border-red-500",
-            "focus:ring-red-500",
-            "dark:border-red-500",
-            "dark:ring-red-500",
-          );
+          this.classList.add(...FIELD_ERROR_CLASSES);
           if (errorMessageElement) {
             errorMessageElement.innerText = validation.error;
             errorMessageElement.classList.remove("invisible");
@@ -52,12 +64,7 @@ export const setupFormValidation = function () {
         } else {
           this.setCustomValidity("");
           this.value = validation.value;
-          this.classList.remove(
-            "border-red-500",
-            "focus:ring-red-500",
-            "dark:border-red-500",
-            "dark:ring-red-500",
-          );
+          this.classList.remove(...FIELD_ERROR_CLASSES);
           if (errorMessageElement) {
             errorMessageElement.classList.add("invisible");
           }
@@ -74,12 +81,7 @@ export const setupFormValidation = function () {
         // Skip validation for empty optional URL fields
         if (!this.value && !this.required) {
           this.setCustomValidity("");
-          this.classList.remove(
-            "border-red-500",
-            "focus:ring-red-500",
-            "dark:border-red-500",
-            "dark:ring-red-500",
-          );
+          this.classList.remove(...FIELD_ERROR_CLASSES);
           const errorMessageElement = this.parentNode?.querySelector(
             'p[data-error-message-for="url"]',
           );
@@ -101,12 +103,7 @@ export const setupFormValidation = function () {
         );
         if (!validation.valid) {
           this.setCustomValidity(validation.error);
-          this.classList.add(
-            "border-red-500",
-            "focus:ring-red-500",
-            "dark:border-red-500",
-            "dark:ring-red-500",
-          );
+          this.classList.add(...FIELD_ERROR_CLASSES);
           if (errorMessageElement) {
             errorMessageElement.innerText = validation.error;
             errorMessageElement.classList.remove("invisible");
@@ -114,12 +111,7 @@ export const setupFormValidation = function () {
         } else {
           this.setCustomValidity("");
           this.value = validation.value;
-          this.classList.remove(
-            "border-red-500",
-            "focus:ring-red-500",
-            "dark:border-red-500",
-            "dark:ring-red-500",
-          );
+          this.classList.remove(...FIELD_ERROR_CLASSES);
           if (errorMessageElement) {
             errorMessageElement.classList.add("invisible");
           }
--- a/mcpgateway/services/resource_service.py
+++ b/mcpgateway/services/resource_service.py
@@ -168,7 +168,7 @@ class ResourceURIConflictError(ResourceError):
         self.uri = uri
         self.enabled = enabled
         self.resource_id = resource_id
-        message = f"{visibility.capitalize()} Resource already exists with URI: {uri} — resource URIs must be unique within this scope (names may repeat)."
+        message = f"{visibility.capitalize()} resource already exists with URI: {uri} — resource URIs must be unique within this scope (names may repeat)."
         if not enabled:
             message += f" (currently inactive, ID: {resource_id})"
         super().__init__(message)
--- a/mcpgateway/services/resource_service.py
+++ b/mcpgateway/services/resource_service.py
@@ -168,7 +168,7 @@ class ResourceURIConflictError(ResourceError):
         self.uri = uri
         self.enabled = enabled
         self.resource_id = resource_id
-        message = f"{visibility.capitalize()} Resource already exists with URI: {uri}"
+        message = f"{visibility.capitalize()} Resource already exists with URI: {uri} — resource URIs must be unique within this scope (names may repeat)."
         if not enabled:
             message += f" (currently inactive, ID: {resource_id})"
         super().__init__(message)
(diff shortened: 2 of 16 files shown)


ja8zyjits (changes requested):
## ⚠️ Warnings

### Documentation Gaps

1. **API Usage Guide: Resource URI uniqueness not documented**
   - File: `docs/docs/manage/api-usage.md:1234-1280`
   - Code: `mcpgateway/services/resource_service.py:171`, `mcpgateway/utils/error_formatter.py:269`
   - Fix: Add note in "Register a Resource" section:
     ```markdown
     !!! note "Resource URI Uniqueness"
         Resource URIs must be unique within their visibility scope (public/team/private). Resource names are display labels and may repeat. If you attempt to create a resource with a duplicate URI, you'll receive a 409 Conflict error.
     ```

2. **API Usage Guide: 409 Conflict response not documented for POST /resources**
   - File: `docs/docs/manage/api-usage.md:1234-1280`
   - Code: `mcpgateway/main.py:6633-6635`, `mcpgateway/services/resource_service.py:168-178`
   - Fix: Add error response table:
     ```markdown
     **Error Responses:**
     
     | Status | Condition | Response |
     |--------|-----------|----------|
     | 409 Conflict | Resource URI already exists in scope | `{"message": "Resource already exists with URI: <uri> — resource URIs must be unique within this scope (names may repeat).", "success": false}` |
     | 422 Validation Error | Invalid resource data | Pydantic validation error details |
     | 400 Bad Request | General resource error | Error message |
     ```

3. **API Usage Guide: PUT /resources error responses incomplete**
   - File: `docs/docs/manage/api-usage.md:1380-1395`
   - Code: `mcpgateway/main.py:6633-6640`
   - Fix: Document new error handlers: `ResourceValidationError` (422), `ResourceError` (400), `ResourceURIConflictError` (409)

4. **Multi-tenancy Architecture: Resource URI uniqueness scope not explicit**
   - File: `docs/docs/architecture/multitenancy.md:580-650`
   - Code: `mcpgateway/services/resource_service.py:559-585`, `3137-3171`
   - Fix: Add uniqueness scope to "Resource Visibility (Design)" section:
     ```markdown
     - Private: Uniqueness scope: (team_id, owner_email, gateway_id, uri)
     - Team: Uniqueness scope: (team_id, gateway_id, uri)
     - Public: Uniqueness scope: (gateway_id, uri)

mcpgateway/admin_ui/formValidation.js
```diff
@@ -4,6 +4,23 @@
...
+ * Tailwind utility classes applied to an input to signal a field-level error.
+ *
+ * Shared so that client-side validation and server-side error attribution
+ * (formSubmitHandlers.js) render identically. ``modals.js`` clears this exact
+ * set when resetting a form *inside a modal*; forms rendered inline on a page
+ * must clear it themselves.
+ *
+ * @type {ReadonlyArray<string>}
+ */
+export const FIELD_ERROR_CLASSES = [
```
gcgoncalves, line 17: Nice refactor here! 👏

mcpgateway/services/resource_service.py
```diff
@@ -168,7 +168,7 @@ def __init__(self, uri: str, enabled: bool = True, resource_id: Optional[str] =
         self.uri = uri
         self.enabled = enabled
         self.resource_id = resource_id
-        message = f"{visibility.capitalize()} Resource already exists with URI: {uri}"
+        message = f"{visibility.capitalize()} Resource already exists with URI: {uri} — resource URIs must be unique within this scope (names may repeat)."
```
gcgoncalves: NIT, non-blocking: `{visibility.capitalize()} Resource` will render something like: "Public Resource already exists...", which is grammatically wrong (only the first workd of the sentence is capitalised).

gcgoncalves (dismissed):

gcgoncalves:
Approved from the JS standpoint.

msureshkumar88:
Thanks both — the docs gap was a fair call, the PR genuinely shipped a new error contract without documenting it anywhere. I've pushed docs for all four items plus the capitalisation nit. Two of the suggestions needed adjusting before they'd be correct, so flagging that below rather than applying them verbatim.

## Applied

**Nit (@gcgoncalves):** agreed, `resource_service.py` now emits `Public resource already exists with URI: …`. Updated the five assertions and the JS fixture that pinned the old casing. The JS conflict detector matches case-insensitively (`/\buri\b/i` + `/already exists/i`), so no behaviour change there.

**1 & 2 — `docs/docs/manage/api-usage.md`, "Register a Resource":** added a `!!! note` stating that `uri` is the identifier and `name` is a display label that may repeat, plus an error-response table (409/422/400) and a sample body.

**3 — same file, "Update Resource":** added the error table. `PUT /resources/{id}` can also return 413 and 415 (`ContentSizeError` / `ContentTypeError`), so those are in the table too.

Also added `409 Conflict` to the shared **Error Handling → Common Error Responses** section, which already documents 401/404/422 in the same `{"detail": …}` shape — felt like the more discoverable home for it than a per-endpoint block.

**4 — `docs/docs/architecture/multitenancy.md`:** added a `### Resource URI uniqueness` subsection covering the DB constraints and the per-visibility application pre-check.

## Two corrections worth noting

**The 409 body in suggestion 2 isn't what the REST API returns.** The suggested body was:

```json
{"message": "Resource already exists with URI: <uri> — …", "success": false}
```

`POST /resources` does `raise HTTPException(status_code=409, detail=str(e))` and there's no `HTTPException` handler reshaping it, so the actual body is:

```json
{"detail": "Public resource already exists with URI: file:///etc/config.json — resource URIs must be unique within this scope (names may repeat)."}
```

The `{"message": …, "success": false}` shape comes from `ErrorFormatter.format_database_error`, which is the **admin UI** path (`/admin/resources`) and the `IntegrityError` fallback — a different response than the REST route. Documenting the two as one shape would send API clients down the wrong parsing path, so I documented the FastAPI shape for the REST endpoints and left the admin shape out of the public API guide. The quoted message also dropped the `{visibility}` prefix the code actually emits.

**The uniqueness-scope table in suggestion 4 doesn't match the schema.** The suggested table was:

```
- Private: (team_id, owner_email, gateway_id, uri)
- Team:    (team_id, gateway_id, uri)
- Public:  (gateway_id, uri)
```

Visibility isn't part of any database key. `Resource.__table_args__` has exactly two:

```python
UniqueConstraint("team_id", "owner_email", "gateway_id", "uri", name="uq_team_owner_gateway_uri_resource")
Index("uq_team_owner_uri_resource_local", "team_id", "owner_email", "uri", unique=True, postgresql_where=text("gateway_id IS NULL"), sqlite_where=text("gateway_id IS NULL"))
```

Both apply to every resource regardless of visibility. What *is* visibility-dependent is the application-level pre-check in `resource_service.py`, and its keys differ from the suggested ones — every branch also filters on `visibility`, and `public` deliberately ignores `team_id`/`owner_email` so a public URI collides platform-wide. So I documented both levels separately and noted where they diverge (the update path drops `gateway_id` from the `public` and `team` branches, unlike create). Publishing the original table as architecture reference would have had readers expecting the DB to permit duplicates it actually rejects.

## Minor

The file/line pointers in the review had drifted — "Register a Resource" is `api-usage.md:886` (not 1234-1280, which is the bulk-import payload), "Update Resource" is `:948` (not 1380-1395, the `TeamCreateRequest` table), and "Resource Visibility (Design)" is `multitenancy.md:283` (not 580-650). Also `main.py:6633-6635` is the `PUT` handler; the `POST` handlers are at `:6391-6396`. Not a problem, just noting it in case the pointers were generated against a different revision.

Happy to move any of the new prose if you'd prefer it somewhere else.

ja8zyjits (dismissed):
LGTM

I have created followup for this.

msureshkumar88:
recheck

msureshkumar88:
Reopening to force CI/DCO recheck (stale check-suite from force-push includes an already-merged ancestor commit not authored by me).

ja8zyjits (dismissed):
LGTM

ja8zyjits (approved):
LGTM

gcgoncalves (approved):

Issue #5419, opened 2026-06-29
[CHORE]: Add production compose smoke CI coverage


lucarlig:
### 🔧 Chore Summary

Add CI coverage for the production Docker Compose path so regressions are caught before merge.

---

### 🧱 Area Affected

Choose the general area(s) that this chore affects:

- [x] GitHub Actions / CI Pipelines
- [ ] Pre-commit hooks / linters
- [ ] Formatting (black, isort, ruff, etc.)
- [ ] Type-checking (mypy, pyright, pytype, etc.)
- [ ] Dependency cleanup or updates
- [x] Build system or `Makefile`
- [x] Containerization (Docker/Podman)
- [ ] Docs or spellcheck
- [ ] SBOM, CVE scans, licenses, or security checks
- [ ] Other:

---

### ⚙️ Context / Rationale

Current CI can pass while missing failures in the real production compose path. Existing checks build images or run the gateway directly with `make serve`, but they do not combine the same path users exercise locally: `make docker-prod`, `make compose-up`, nginx on port 8080, and admin login through `/admin/login`.

That gap allowed a production compose startup issue to escape green CI: the stack could build successfully, while nginx/admin traffic returned a gateway failure because the container runtime path was not exercised end to end.

The CI coverage should verify the production image starts under Compose, gateway health is available through nginx, admin login renders successfully, the testing profile can start, and gateway logs stay free of startup/runtime failures such as read-only filesystem errors or worker SIGKILLs.

---

### 📦 Related Make Targets

- `make docker-prod` - build the production image used by Compose
- `make compose-up` - start the default production Compose stack
- `make testing-up` - start the testing profile services
- `make compose-down` - stop and clean up the stack

---

### 📋 Acceptance Criteria

Define what "done" looks like for this task.

- [ ] CI builds the production image with `make docker-prod`
- [ ] CI starts the default Compose stack with resource values suitable for GitHub-hosted runners
- [ ] CI verifies `http://localhost:8080/health` reports healthy through nginx
- [ ] CI verifies `http://localhost:8080/admin/login` renders the admin login page
- [ ] CI starts `make testing-up` and verifies the gateway, fast test server, A2A echo agent, and Locust endpoints respond
- [ ] CI fails when gateway logs contain startup/runtime errors such as control socket failures, read-only filesystem errors, worker SIGKILLs, or OOM messages
- [ ] CI uploads Compose logs when the smoke check fails

---

### 🧩 Additional Context

Related draft PR: #5418

PR #5923, opened 2026-07-27
ci: create .env before compose-up in prod smoke check

madhu-mohan-jaishankar:
## 🔗 Related Issue

Follow-up to #5419

---

## 📝 Summary

The `compose-prod-smoke` merge-queue check has failed on every run since it was
added. The job goes straight from `make docker-prod` to `make compose-up`, but
`compose-up` depends on `compose-validate`, which hard-fails when `.env` is
absent.

The workflow never created `.env`, so validation aborted with exit code 2 before
the stack ever started.

This PR adds a **Prepare .env** step (after the image build) that seeds `.env`
from `.env.example` and patches in strong secrets:

- `cp .env.example .env` satisfies the `compose-validate` file check.
- `JWT_SECRET_KEY` and `AUTH_ENCRYPTION_SECRET` are replaced with
  `openssl rand -hex 32` values, since the gateway refuses to start with the
  `__REPLACE_ME__` placeholders shipped in `.env.example`.

I used inline shell rather than the documented `make setup` on purpose: `setup`
runs `uv run python3 …` via `ensure-secrets`, and this is a Docker-only job with
no Python/uv toolchain — using `openssl` keeps the step self-contained.

---

## 📏 Reviewability

---

## 🏷️ Type of Change

- [x] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

### DIFF
--- a/.github/workflows/compose-prod-smoke.yml
+++ b/.github/workflows/compose-prod-smoke.yml
@@ -63,6 +63,26 @@ jobs:
       - name: Build production image
         run: make docker-prod
 
+      # Compose targets validate against .env (compose-up → compose-validate),
+      # which aborts if the file is missing. `make setup` would create it, but
+      # that path runs `uv run` and this Docker-only job has no Python/uv
+      # toolchain. Seed .env from .env.example directly and patch in strong
+      # secrets so the gateway starts (it refuses the __REPLACE_ME__
+      # placeholders shipped in .env.example).
+      - name: Prepare .env
+        shell: bash
+        run: |
+          set -euo pipefail
+          cp .env.example .env
+          # base64 (not hex) so Shannon entropy clears the >= 3.5 floor that
+          # Settings.validate_security_combinations() enforces at startup; hex
+          # tops out at 4 bits/char and occasionally dips below it. base64's
+          # alphabet has no '|' or '&', so it is safe with the sed delimiter.
+          jwt_secret="$(openssl rand -base64 32)"  # pragma: allowlist secret
+          enc_secret="$(openssl rand -base64 32)"  # pragma: allowlist secret
+          sed -i "s|^JWT_SECRET_KEY=.*|JWT_SECRET_KEY=${jwt_secret}|" .env
+          sed -i "s|^AUTH_ENCRYPTION_SECRET=.*|AUTH_ENCRYPTION_SECRET=${enc_secret}|" .env
+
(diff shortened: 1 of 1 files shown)


prakhar-singh1928 (dismissed):

.github/workflows/compose-prod-smoke.yml
```diff
@@ -63,6 +63,22 @@ jobs:
...
+      # toolchain. Seed .env from .env.example directly and patch in strong
+      # secrets so the gateway starts (it refuses the __REPLACE_ME__
+      # placeholders shipped in .env.example).
+      - name: Prepare .env
+        shell: bash
+        run: |
+          set -euo pipefail
+          cp .env.example .env
+          jwt_secret="$(openssl rand -hex 32)"  # pragma: allowlist secret
+          enc_secret="$(openssl rand -hex 32)"  # pragma: allowlist secret
```
lucarlig: **Functionally impacting:** Could we use `openssl rand -base64 32` (already used in `.github/workflows/plugin-integration.yml`) or retry until the configured entropy predicate passes? `Settings.validate_security_combinations()` rejects secrets whose observed Shannon entropy is below 3.5. A 64-character hex sample has a ceiling of only 4 bits per character; in a 200,000-sample reproduction, 35 values were rejected. With two independently generated hex secrets, this leaves a rare startup/merge-queue flake (roughly 1 in 2,900 runs). Base64 gives the validator substantially more headroom and remains safe with the current `sed` delimiter.

lucarlig (changes requested):
Thanks for isolating the failure and keeping the fix tightly scoped. The added step addresses the missing `.env` seen in the merge-queue logs, and the updated workflow passes Actionlint and ShellCheck.

I found one functionally-impacting flake risk that is worth addressing before merge; details are inline.

One validation caveat: the green `Compose Production Smoke Complete` check is the skip reporter, and no actual smoke run exists for this head. It would be helpful to manually dispatch the workflow on this branch after updating.

Minor: `Closes #5419` references an issue that is already closed and still labeled `triage`; `Follow-up to #5419` would be more accurate.

lucarlig (approved):
LGTM

prakhar-singh1928 (approved):

Lang-Akshay (approved):
LGTM

Issue #4544, opened 2026-04-30
[CF-DATAPLANE] Based on the MCP configuration implement necessary functionality to modify request response in MCP Gateway


dawid-nowak:
This will require enabling logic to pass/remove/add headers from downstream to upstream.

PR #5924, opened 2026-07-27
feat(dataplane): publish per-backend add_headers/remove_headers

Lang-Akshay:
## Dataplane
Related Issue [4544](https://github.com/IBM/mcp-context-forge/issues/4544)

Related Dataplane PR [65](https://github.com/contextforge-org/contextforge-data-plane/pull/65)
## Summary

The dataplane consumer ([contextforge-gateway-rs #4544](https://github.com/IBM/mcp-context-forge)) now accepts two new per-backend header controls alongside the existing `passthrough_headers`:

- `add_headers` — static headers injected onto the upstream connection (override passthrough), a `str -> str` map
- `remove_headers` — header names stripped from the upstream connection (applied last), a list

This threads both fields through `DataplanePublisherService` so the published `BackendConfig` matches the consumer's contract.

## Details

- `BackendConfig` TypedDict gains `add_headers: dict[str, str]` and `remove_headers: list[str]`.
- `create_payload` emits both, normalizing `None` to the consumer-side defaults (`{}` / `[]`).
- `_build_user_data` sources them from the gateway row via `getattr(..., None)`. The `Gateway` model has no columns for these yet, so today the publisher emits the safe defaults; it lights up automatically once the columns are added — no further change needed in this file.

## Tests

Extended `test_dataplane_publisher.py`: asserts the new fields flow through with values and that missing values normalize to `{}` / `[]`. All 24 tests pass.

### DIFF
--- a/mcpgateway/services/dataplane_publisher.py
+++ b/mcpgateway/services/dataplane_publisher.py
@@ -313,6 +313,8 @@ class DataplanePublisherService:
                         DbGateway.url,
                         DbGateway.transport,
                         DbGateway.passthrough_headers,
+                        DbGateway.add_headers,
+                        DbGateway.remove_headers,
                         DbGateway.capabilities,
                         DbGateway.owner_email,
                         DbGateway.team_id,
@@ -379,10 +381,8 @@ class DataplanePublisherService:
                     "url": gateway.url,
                     "transport": gateway.transport,
                     "passthrough_headers": gateway.passthrough_headers,
-                    # DB columns not yet present; getattr keeps the diff in this file and
-                    # emits the Rust-side defaults ({} / []) until the columns land.
-                    "add_headers": getattr(gateway, "add_headers", None) or {},
-                    "remove_headers": getattr(gateway, "remove_headers", None) or [],
+                    "add_headers": gateway.add_headers or {},
+                    "remove_headers": gateway.remove_headers or [],
                     "capabilities": gateway.capabilities or {},
                 }
                 for gateway in gateway_rows
--- a/mcpgateway/services/dataplane_publisher.py
+++ b/mcpgateway/services/dataplane_publisher.py
@@ -64,6 +64,8 @@ class BackendConfig(TypedDict):
     url: str
     transport: str
     passthrough_headers: list[str]
+    add_headers: dict[str, str]
+    remove_headers: list[str]
     capabilities: dict[str, Any]
     allowed_tool_names: list[str]
     allowed_resource_names: list[str]
@@ -236,6 +238,8 @@ class DataplanePublisherService:
                     "url": gateway["url"],
                     "transport": gateway["transport"],
                     "passthrough_headers": gateway["passthrough_headers"] or [],
+                    "add_headers": gateway.get("add_headers") or {},
+                    "remove_headers": gateway.get("remove_headers") or [],
                     "capabilities": gateway.get("capabilities") or {},
                 }
                 for gateway in gateways
@@ -375,6 +379,10 @@ class DataplanePublisherService:
                     "url": gateway.url,
                     "transport": gateway.transport,
                     "passthrough_headers": gateway.passthrough_headers,
+                    # DB columns not yet present; getattr keeps the diff in this file and
+                    # emits the Rust-side defaults ({} / []) until the columns land.
+                    "add_headers": getattr(gateway, "add_headers", None) or {},
+                    "remove_headers": getattr(gateway, "remove_headers", None) or [],
                     "capabilities": gateway.capabilities or {},
                 }
                 for gateway in gateway_rows
--- a/tests/unit/mcpgateway/services/test_dataplane_publisher.py
+++ b/tests/unit/mcpgateway/services/test_dataplane_publisher.py
@@ -185,6 +185,8 @@ async def test_full_payload_generation_with_mock_db():
     gateway1.url = "http://localhost:9000"
     gateway1.transport = "STREAMABLEHTTP"
     gateway1.passthrough_headers = ["Authorization"]
+    gateway1.add_headers = {"X-Tenant": "acme"}
+    gateway1.remove_headers = ["Cookie"]
     gateway1.capabilities = {"resources": {"subscribe": True}}
     gateway1.owner_email = "user1@example.com"
     gateway1.team_id = "team1"
@@ -283,6 +285,8 @@ async def test_full_payload_generation_with_mock_db():
         assert backend["url"] == "http://localhost:9000"
         assert backend["transport"] == "STREAMABLEHTTP"
         assert backend["passthrough_headers"] == ["Authorization"]
+        assert backend["add_headers"] == {"X-Tenant": "acme"}
+        assert backend["remove_headers"] == ["Cookie"]
         assert backend["capabilities"] == {"resources": {"subscribe": True}}
         assert backend["allowed_tool_names"] == ["public_tool", "private_tool"]
         assert backend["allowed_resource_names"] == ["Resource 1"]
@@ -450,6 +454,8 @@ def test_create_payload_normalizes_null_passthrough_headers():
 
     backend = result[USER1_ID]["virtual_hosts"]["server1"]["backends"]["gateway1"]
     assert backend["passthrough_headers"] == []
+    assert backend["add_headers"] == {}
+    assert backend["remove_headers"] == []
     assert backend["capabilities"] == {}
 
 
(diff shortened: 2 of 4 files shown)


mcpgateway/services/dataplane_publisher.py
```diff
@@ -373,6 +377,10 @@ def _build_user_data(
                     "url": gateway.url,
                     "transport": gateway.transport,
                     "passthrough_headers": gateway.passthrough_headers,
+                    # DB columns not yet present; getattr keeps the diff in this file and
+                    # emits the Rust-side defaults ({} / []) until the columns land.
+                    "add_headers": getattr(gateway, "add_headers", None) or {},
+                    "remove_headers": getattr(gateway, "remove_headers", None) or [],
```
madhu-mohan-jaishankar: gateway_rows comes from a column-level `select(DbGateway.id, .name, .url, .transport, .passthrough_headers, .capabilities, ...)`, which returns sqlalchemy.Row objects, not ORM entities. A Row exposes only the columns named in the SELECT. the SELECT list at lines 308-318 does not include the new fields. So adding add_headers/remove_headers columns to the Gateway model will not make getattr(gateway, "add_headers", None) resolve. It will keep returning None, and the publisher will keep emitting {}/[] forever. 

Failure mode is silent: an operator sets add_headers = {"X-Tenant": "acme"}, it persists, the admin UI shows it, the dataplane never applies it. Nothing raises, no test breaks.

Fix: add` DbGateway.add_headers` / `DbGateway.remove_headers` to the select(...).

tests/unit/mcpgateway/services/test_dataplane_publisher.py
```diff
@@ -180,6 +180,8 @@ async def test_full_payload_generation_with_mock_db():
     gateway1.url = "http://localhost:9000"
     gateway1.transport = "STREAMABLEHTTP"
     gateway1.passthrough_headers = ["Authorization"]
+    gateway1.add_headers = {"X-Tenant": "acme"}
+    gateway1.remove_headers = ["Cookie"]
```
madhu-mohan-jaishankar, line 189: `gateway1 = Mock()` with `add_headers `injected directly into `mock_db.execute.return_value.all.side_effect`. A bare `Mock `auto-creates any attribute, and the real `select(...)` projection is never exercised. `backend["add_headers"] == {"X-Tenant": "acme"}` passes today and will keep passing after the columns land, while production publishes {}. None of the 24 tests in the file use a real Session/DbGateway.

Fix: assert on the emitted SELECT (inspect `mock_db.execute.call_args` for selected column names).

tests/unit/mcpgateway/services/test_dataplane_publisher.py
```diff
@@ -280,6 +282,8 @@ async def test_full_payload_generation_with_mock_db():
         assert backend["url"] == "http://localhost:9000"
         assert backend["transport"] == "STREAMABLEHTTP"
         assert backend["passthrough_headers"] == ["Authorization"]
+        assert backend["add_headers"] == {"X-Tenant": "acme"}
+        assert backend["remove_headers"] == ["Cookie"]
```
madhu-mohan-jaishankar, line 289: same comment as above.

madhu-mohan-jaishankar (changes requested):

Lang-Akshay:
Thanks for the sharp catches — all three findings were valid and are now addressed.

**Comment 1 — Missing columns in SELECT projection**

`add_headers` and `remove_headers` are now included in the `select(...)` at `get_data_from_db` alongside the existing gateway columns, so SQLAlchemy `Row` objects will carry the real values once the columns land. The `getattr(..., None)` fallbacks are replaced with direct attribute access (`gateway.add_headers or {}` / `gateway.remove_headers or []`), which is both correct and honest about what the code expects.

**Comments 2 & 3 — Mock-based tests masking the projection gap**

Added an assertion after the payload checks that inspects the actual SQLAlchemy statement passed to `db.execute` (the 4th call, which is the gateway query):

```python
gateway_execute_call = mock_db.execute.call_args_list[3]
stmt = gateway_execute_call[0][0]
selected_keys = {col.key for col in stmt.selected_columns}
assert "add_headers" in selected_keys, "Gateway SELECT must include add_headers"
assert "remove_headers" in selected_keys, "Gateway SELECT must include remove_headers"
```

This would have failed on the original code before the SELECT fix, so it closes the gap without needing a full real-session test.

**Migration**

Added `e1a2b3c4d5f6_add_add_remove_headers_to_gateways.py` with `down_revision = "d21698ae4a19"` (previous head). Idempotent — checks column existence before adding/dropping. Single head confirmed via `alembic heads`.

All 24 tests pass.

madhu-mohan-jaishankar (approved):
LGTM

msureshkumar88 (approved):
## E2E verification — PASS ✅

Ran an end-to-end verification of this PR's control-plane contract against real components (not mocks): a real `alembic` CLI upgrading a real SQLite file that already had the pre-PR schema, a real dockerized `redis:7-alpine` server, the real ORM models, and the unmodified `DataplanePublisherService`.

### What was verified

1. **Migration, on an existing DB** — built a DB at the prior head (`d21698ae4a19`), then ran `alembic upgrade e1a2b3c4d5f6` (this PR's migration) against it. Confirmed it performs a real `ALTER TABLE` adding `add_headers`/`remove_headers` to `gateways` (not the fresh-DB skip branch). Then `alembic downgrade -1` → confirmed both columns are dropped cleanly, and `alembic upgrade head` again → confirmed they come back. No multi-head issue (`alembic heads` → single head `e1a2b3c4d5f6`).
2. **Real publish pipeline** — created two real `Gateway` rows via the ORM (one with `add_headers={"X-Tenant": "acme"}` / `remove_headers=["Cookie"]`, one with neither set), each wired to a `Server` via a real `Tool` + `server_tool_association` row. Called the actual `DataplanePublisherService.fetch_payload()` (unmodified, real DB queries) and wrote the result to the real Redis container using the exact msgpack/key format `publish_to_redis()` uses in production.
3. **Read-back, as an external consumer would see it** — read the published key back over a raw-bytes Redis connection (independent of the app's own client) and confirmed the decoded `BackendConfig` matches exactly: the gateway with headers set published `add_headers`/`remove_headers` with the exact values; the gateway without them published the documented `{}`/`[]` defaults.
4. **Regression sweep** — `test_dataplane_publisher.py` (24/24), `test_gateway_lifecycle_migration.py` (10/10), `test_gateway_service.py -k "passthrough or create_gateway or update_gateway or register_gateway"` (69/69), `test_export_service.py`/`test_import_service.py -k gateway` (213/213) all green. `ruff` and `interrogate` clean on the changed files.

### Config used
```
DATABASE_URL=sqlite:///<tmp>/e2e.db        # fresh file, built from pre-PR head then migrated
CACHE_TYPE=redis
REDIS_URL=redis://localhost:<ephemeral-port>/0   # real redis:7-alpine container, docker run -d --rm
JWT_SECRET_KEY / AUTH_ENCRYPTION_SECRET / BASIC_AUTH_PASSWORD = throwaway values
                                             # (only to satisfy Settings validation; never real secrets)
```

### Commands run
```
alembic upgrade d21698ae4a19        # pre-PR baseline
alembic upgrade e1a2b3c4d5f6        # this PR's migration
alembic downgrade -1 && alembic upgrade head   # round-trip check
docker run -d --rm -p <port>:6379 redis:7-alpine
python <script driving fetch_payload() + real redis write/read>
pytest tests/unit/mcpgateway/services/test_dataplane_publisher.py -q
pytest tests/unit/mcpgateway/db/test_gateway_lifecycle_migration.py -q
pytest tests/unit/mcpgateway/services/test_gateway_service.py -q -k "passthrough or create_gateway or update_gateway or register_gateway"
pytest tests/unit/mcpgateway/services/test_export_service.py tests/unit/mcpgateway/services/test_import_service.py -q -k gateway
```

### Output (published payload for the test gateway with headers set)
```json
{
  "name": "e2e-gw-with-headers",
  "transport": "STREAMABLEHTTP",
  "passthrough_headers": [],
  "add_headers": {"X-Tenant": "acme"},
  "remove_headers": ["Cookie"],
  "allowed_tool_names": ["tool_a"]
}
```

### One thing this couldn't verify
Whether a real upstream HTTP request actually gets `X-Tenant` added and `Cookie` stripped is implemented in the external dataplane consumer (`contextforge-data-plane`, referenced PR #65) — that code isn't in this repository, so it's outside what could be exercised here. Everything on the Python control-plane side of the contract (schema → publish → Redis JSON) checks out.

### Non-blocking notes for follow-up (not required for this PR)
- `add_headers`/`remove_headers` have no `GatewayCreate`/`GatewayUpdate`/Admin UI exposure yet, so there's currently no way to actually set them outside direct DB access — worth a tracked follow-up issue so the columns don't sit unused.
- Worth a quick confirmation that `contextforge-data-plane` PR #65 is deployed, since this starts publishing the new keys unconditionally.

Approving — the migration is sound, the publish contract matches the PR description exactly, and nothing in the surrounding gateway/export/import paths regressed.

PR #5932, opened 2026-07-27
fix(tokens): allow reusing a revoked token's name

### LINKED ISSUE #5931: Revoked API token permanently blocks its name (uniqueness constraint ignores is_active)
### Description

Deleting an API token permanently blocks its name. Token revocation is a soft delete (`is_active = false`), but the uniqueness rule on `email_api_tokens (user_email, name, team_id)` covers **all** rows, so once a token is revoked its name can never be used again by that user: creating a new token with the same name fails with `409 - A token with this name already exists. Please choose a different name.` even though the UI shows no token with that name.

### Steps to reproduce

1. Create an API token named `my-token` (UI or `POST /tokens`)
2. Delete/revoke it
3. Create a new token named `my-token` again

**Expected**: token is created (no active token holds that name).
**Actual**: `409 Conflict - A token with this name already exists. Please choose a different name.`

Server log:

```
Token creation integrity error: duplicate key value violates unique constraint "uq_email_api_tokens_user_name_global" DETAIL: Key (user_email, name)=(user@example.com, my-token) already exists.
```

### Root cause

The service-level duplicate check in `TokenCatalogService.create_token()` already filters on `is_active.is_(True)` (i.e. the intended semantics are "names unique among *active* tokens"), but the DB-level rules contradict it:

- `UniqueConstraint("user_email", "name", "team_id", name="uq_email_api_tokens_user_name_team")` — all rows
- `Index("uq_email_api_tokens_user_name_global", ...)` — partial on `team_id IS NULL` only, no `is_active` filter

`revoke_token()` only sets `is_active = False` and blacklists the JTI; the row is (correctly) kept for audit, so it collides forever.

This is the same class of problem fixed for `roles`/`user_roles` in #4482 / migration `d21698ae4a19` (partial unique indexes scoped to `WHERE is_active`).

### Environment

- Reproduced on v1.0.5 (PostgreSQL backend) and confirmed the constraints are unchanged on current `main`

bleiva:
# Pull Request

## 🔗 Related Issue

Closes #5931

---

## 📝 Summary

Token revocation is a soft delete (`is_active = false`), but the uniqueness rule on `email_api_tokens (user_email, name, team_id)` covers all rows. Once a token is revoked, its name is blocked forever for that user: creating a new token with the same name fails with "A token with this name already exists" even though the UI shows no such token.

The service-level duplicate check in `TokenCatalogService.create_token()` already filters on `is_active`, so the DB constraint contradicted the intended semantics. This PR replaces the `UniqueConstraint` and the global partial index with partial unique indexes scoped to active rows (`WHERE is_active`), the same pattern migration `d21698ae4a19` applied to `roles`/`user_roles` for #4482. Index names are unchanged so the existing `IntegrityError` handlers in `routers/tokens.py` and `utils/error_formatter.py` keep matching.

Changes:

- `mcpgateway/db.py`: `EmailApiToken.__table_args__` now uses active-only partial unique indexes (PostgreSQL and SQLite; other dialects keep the previous full-row uniqueness as fallback)
- New Alembic migration `f3a8c2d94e17` (idempotent, `batch_alter_table` for SQLite compatibility, no dedup step needed since the old stricter rule implies the new one)
- Migration unit tests + service-level integration tests covering create → revoke → recreate with the same name, and duplicate-active rejection (global and team scope)

---

## 📏 Reviewability

---

## 🏷️ Type of Change

- [x] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

| Check | Command | Status |
|-------|---------|--------|
| New tests (migration + integration) | `pytest tests/unit/mcpgateway/db/test_token_name_active_uniqueness_migration.py tests/integration/test_token_name_reuse_integration.py` | ✅ 14 passed |
| Related token suites | `pytest tests/unit/mcpgateway/db/test_token_uniqueness_migration.py tests/unit/mcpgateway/services/test_token_catalog_service.py tests/unit/mcpgateway/routers/test_tokens.py tests/unit/mcpgateway/utils/test_error_formatter.py tests/integration/test_token_security_integration.py` | ✅ 300 passed (4 failures in `test_token_security_integration.py` pre-exist on unmodified `main`) |
| Formatting | `black` / `isort` on changed files | ✅ clean |

The integration test `test_revoked_token_name_can_be_reused` reproduces the bug: it fails with `IntegrityError` on the old schema and passes with this change.

---

## ✅ Checklist

---

## 📓 Notes (optional)

- Downgrade restores the original constraint; it can fail if a revoked token now shares a name with an active one — noted in the migration docstring (such rows must be renamed or deleted before downgrading).
- On dialects without partial-index support (e.g. MySQL) the migration recreates full unique indexes with a warning, keeping today's behavior rather than regressing.

### DIFF
--- a/mcpgateway/alembic/versions/f3a8c2d94e17_scope_token_name_uniqueness_to_active_.py
+++ b/mcpgateway/alembic/versions/f3a8c2d94e17_scope_token_name_uniqueness_to_active_.py
@@ -1,0 +1,169 @@
+# -*- coding: utf-8 -*-
+# pylint: disable=no-member
+"""Location: ./mcpgateway/alembic/versions/f3a8c2d94e17_scope_token_name_uniqueness_to_active_.py
+Copyright contributors to the MCP-CONTEXT-FORGE project
+SPDX-License-Identifier: Apache-2.0
+
+scope token name uniqueness to active tokens
+
+Revision ID: f3a8c2d94e17
+Revises: d21698ae4a19
+Create Date: 2026-07-27
+
+Token revocation is a soft delete (is_active = false), but the uniqueness rule on
+email_api_tokens (user_email, name, team_id) applied to ALL rows.  Once a token was
+revoked, its name was blocked forever for that user: creating a new token with the
+same name raised "A token with this name already exists" even though the UI showed
+no such token.  The service-level duplicate check in TokenCatalogService.create_token()
+already filters on is_active, so the DB constraint contradicted the intended semantics.
+
+This migration replaces the uniqueness rule with partial unique indexes that only
+cover active tokens — the same pattern migration d21698ae4a19 applied to the roles
+and user_roles tables:
+
+Old constraint: uq_email_api_tokens_user_name_team   -> (user_email, name, team_id)
+Old index:      uq_email_api_tokens_user_name_global -> (user_email, name) WHERE team_id IS NULL
+New index:      uq_email_api_tokens_user_name_team   -> (user_email, name, team_id) WHERE team_id IS NOT NULL AND is_active
+New index:      uq_email_api_tokens_user_name_global -> (user_email, name) WHERE team_id IS NULL AND is_active
+
+Index names are kept identical to the old constraint/index names so the IntegrityError
+handlers in routers/tokens.py and utils/error_formatter.py keep matching.
+
+No deduplication step is needed: the old (stricter) rule guarantees existing rows
+also satisfy the new (active-only) rule.
+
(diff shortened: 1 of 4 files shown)


jonpspri:
Closing in favour of #5989, which fixes the same revoked token name reuse bug with a more complete implementation: it includes a `CHANGELOG` entry, a service-layer fix in `token_catalog_service.py` (not just the DB constraint), and a more thorough test suite. Both use the same Alembic migration approach.

PR #5930, opened 2026-07-27
fix: improve Test Connection error messages

Altamimi-Dev:
Closes #5041

Targets `main`. Originally opened against `epic/ui-rewrite`; retargeted at the maintainers' request — the change is Python + legacy admin-UI JS only, so it cherry-picked onto `main` conflict-free.

## What

Rewrites Test Connection error copy to be actionable, using only information the code already has. Three surfaces:

**Backend** — `test_gateway_connectivity` (`mcpgateway/services/gateway_service.py`), shared by `/admin/gateways/test` and `/v1/mcp-servers/test`. Status codes and response shape unchanged; string literals only:

| Status | Before | After |
|---|---|---|
| 400 (allowlist) | `Invalid gateway URL` | `The MCP server URL is not allowed for testing. Confirm the URL is correct and the host is permitted by your test policy.` |
| 401 (auth-code, no email) | `User authentication required for OAuth-protected gateway '…'. Please ensure you are authenticated.` | `Sign in with an email-bound account to test OAuth-protected MCP server '…'. API tokens without an email cannot complete the OAuth user flow.` |
| 401 (auth-code, no token) | `Please authorize … first. Visit /oauth/authorize/… to complete OAuth flow.` | `Authorize '…' before testing. Open /oauth/authorize/… to complete the OAuth flow.` |
| 500 (stored-token retrieval) | `OAuth token retrieval failed for gateway` | `Token retrieval failed for MCP server '…': {sanitize_exception_message(str(e))}. Check the OAuth client credentials and token URL.` |
| 502 (client-credentials token retrieval) | `OAuth token retrieval failed for gateway` | `Token retrieval failed for MCP server '…': {sanitize_exception_message(str(e))}. Check the OAuth client credentials and token URL.` |
| 502 (request failure) | `Request failed` | `The MCP server request failed. Verify the MCP server is running and reachable from this host.` |

The 400 stays intentionally generic about *which* policy rejected the host (existing comment warns against leaking allowlist details) but now tells the user what to check.

The two token-retrieval messages run the embedded exception text through `sanitize_exception_message`, so an upstream error carrying a token URL with credentials in the query string cannot surface them in the API response. The `logger.error` above each branch still records the raw detail for operators. This follows the convention established in review on #5934.

[...]

PR #5934, opened 2026-07-27
feat(api): add MCP handshake test endpoint for Test Connection

### LINKED ISSUE #5649: [UI-REWRITE]: Add MCP handshake test mode for MCP servers
Add an MCP handshake test mode alongside the raw HTTP probe in the Test connection surface (introduced in #5648). Handshake mode verifies that the target speaks MCP, not just that HTTP responds.

**Requirements**

- Users can run a handshake test against a server URL and see whether MCP negotiation succeeds.
- Spec version is detected: `server/discover` for MCP 2026-07-28 and later, `initialize` for earlier spec versions. Result indicates which path was used.
- Handshake result surfaces server name, version, advertised capabilities, and component counts (tools, resources, prompts).
- Failure surfaces a classification (transport, protocol negotiation, auth, invalid response) with actionable copy per class.

**Security requirements**

- Handshake requests use the same egress allowlist and network policy as `POST /v1/mcp-servers/test`.
- The credential source used for the handshake (stored server credential, form input, or caller session) is shown explicitly in the UI.
- `initialize` sessions are closed on success, failure, and drawer navigation.
- Handshake response payloads are size-capped and not rendered inline beyond a bounded preview.

**Non-goals**

- Per-tool, per-resource, per-prompt probing.
- Rolling health-check dashboards.
- Persisting handshake results across sessions.

**Related**

- #5648 (relocation issue; adds the Test connection surface this extends).
- #5040, #5041 (original dialog).
- [MCP 2026-07-28 RC](https://blog.modelcontextprotocol.io/posts/2026-07-28-release-candidate/) (introduces `server/discover`).

Altamimi-Dev:
Relates to #5649 — UI half: contextforge-org/contextforge-web-ui#16; together they complete the issue.

Targets `main`, backend-only. Originally opened against `epic/ui-rewrite` with both halves; split at the maintainers' request now that the React client lives in [contextforge-org/contextforge-web-ui](https://github.com/contextforge-org/contextforge-web-ui) — same treatment as #5786 / web-ui#15.

## Summary

Adds an **MCP handshake** test alongside the raw HTTP probe. Handshake mode verifies the target actually speaks MCP rather than just accepting TCP connections, and reports how the negotiation happened.

- New `POST /v1/mcp-servers/test-handshake` endpoint (`GatewayHandshakeRequest`/`GatewayHandshakeResponse` schemas).
- `test_gateway_handshake()` tries the stateless `server/discover` method (MCP 2026-07-28+) first and falls back to a stateful SDK `initialize` round-trip for earlier specs — a 2026-07-28-only server cannot answer `initialize` (removed), and an older server cheaply rejects `server/discover`, so this order needs no version pre-knowledge.
- Reports negotiation path, protocol version, server name/version, capabilities, and first-page component counts (`countsPartial` flags `nextCursor` truncation).
- Failure classification (`transport` / `protocol` / `auth` / `invalid_response`) with actionable copy; exception text is run through `sanitize_exception_message` before embedding. Auth (401/403) and network failures short-circuit without attempting the fallback.
- The OAuth token-retrieval failure path now also embeds only sanitized exception text (`sanitize_exception_message(str(e))`), addressing the review finding on the first revision. The `logger.error` above it still records the raw detail.
- Shares the raw test's egress allowlist + DNS-pinning prelude (extracted into `_validate_gateway_test_target`, used by both) and its stored-credential resolution; `credentialSource` (`stored`/`form`/`none`) is reported in the response.
- Sessions close on success/failure via the SDK async-with teardown; the whole probe is bounded by `health_check_timeout` and `rawPreview` is capped at 4096 chars. No DB writes, no gateway health-state side effects (`_mark_gateway_reachable`/`_handle_gateway_failure`/`_refresh_gateway_tools_resources_prompts` are never called).
- Token scoping maps the new path to `gateways.read` — required, the existing `^/mcp-servers/test(?:$|/)` regex does not match `test-handshake`.

The UI counterpart (mode toggle, detail rows, count badges, failure badge, raw preview) is contextforge-org/contextforge-web-ui#16.

## Notes

- `server/discover` is a hand-rolled JSON-RPC POST: the pinned MCP SDK predates the 2026-07-28 spec and has no knowledge of it, so `MCP_STATELESS_PROTOCOL_VERSION = "2026-07-28"` is a hand-written constant. The discover result shape is parsed tolerantly since the RC schema may drift; all response fields are Optional.
- Component counts are first-page counts by design — the probe answers "does it speak MCP and expose components", not "exactly how many".
- The legacy `/admin/gateways/test` surface is intentionally not extended.

## Verification

Re-run on the retargeted `main`-based branch:

- `pytest tests/unit/mcpgateway/routers/test_mcp_servers_router.py` — 26 passed, including the 9 handshake tests (allowlist rejection with no outbound call, discover success with component counts, `-32601` fallback to initialize, 401 auth short-circuit, connect error, invalid-response classification, 401/403/cross-team deny paths)
- `pytest tests/unit/mcpgateway/routers/test_mcp_servers_router.py tests/unit/mcpgateway/middleware/test_token_scoping.py` — 344 passed
- Full `make test` — 21873 passed, 818 skipped, 2 xfailed, 0 failed
- `make ruff` clean, `make interrogate` 100%, `pre-commit run --files …` clean on all five touched files
- Live smoke against a running gateway with a real MCP streamable-HTTP target: success → `negotiationPath: initialize`, real `serverName`/`serverVersion`/`protocolVersion`, `componentCounts` populated, `credentialSource: form`; target requiring auth without headers → `failureClass: auth`; closed port → `failureClass: transport` with sanitized detail; host outside the test allowlist → generic allowlist copy; unauthenticated request → `401`

## Note on landing order

#5930 also touches the validation region of `test_gateway_connectivity` on `main`, so whichever of the two lands second takes a small rebase. #5930 first is the cheaper order (it is copy-only) — the extracted `_validate_gateway_test_target` here then carries its new allowlist wording directly. The end state is consistent either way, since this PR independently introduces the same "The MCP server URL is not allowed for testing…" phrasing on the handshake path.

### DIFF
--- a/mcpgateway/services/gateway_service.py
+++ b/mcpgateway/services/gateway_service.py
@@ -45,13 +45,14 @@ import asyncio
 import binascii
 from dataclasses import dataclass
 from datetime import datetime, timedelta, timezone
+import json
 import logging
 import mimetypes
 import os
 import ssl
 import tempfile
 import time
-from typing import Any, AsyncGenerator, Awaitable, Callable, cast, Dict, List, Optional, Set, TYPE_CHECKING, Union
+from typing import Any, AsyncGenerator, Awaitable, Callable, cast, Dict, List, Literal, Optional, Set, TYPE_CHECKING, Union
 from urllib.parse import urlparse, urlunparse
 import uuid
 
@@ -61,6 +62,7 @@ import httpx
 from mcp import ClientSession
 from mcp.client.sse import sse_client
 from mcp.client.streamable_http import streamablehttp_client
+from mcp.shared.exceptions import McpError
 from pydantic import ValidationError
 from sqlalchemy import and_, delete, desc, or_, select, update
 from sqlalchemy.exc import IntegrityError, SQLAlchemyError
@@ -78,6 +80,7 @@ except ImportError:
     logging.info("Redis is not utilized in this environment.")
 
 # First-Party
+from mcpgateway import __version__
 from mcpgateway.common.validators import SecurityValidator
 from mcpgateway.config import settings
 from mcpgateway.db import EmailTeam as DbEmailTeam
@@ -92,7 +95,19 @@ from mcpgateway.db import ResourceMetric, ResourceSubscription, server_prompt_as
 from mcpgateway.db import Tool as DbTool
 from mcpgateway.db import ToolMetric
 from mcpgateway.observability import create_span, set_span_attribute, set_span_error
-from mcpgateway.schemas import GATEWAY_SUPPORTED_TRANSPORTS, GatewayCreate, GatewayRead, GatewayTestRequest, GatewayTestResponse, GatewayUpdate, PromptCreate, ResourceCreate, ToolCreate
+from mcpgateway.schemas import (
+    GATEWAY_SUPPORTED_TRANSPORTS,
+    GatewayCreate,
+    GatewayHandshakeRequest,
+    GatewayHandshakeResponse,
+    GatewayRead,
+    GatewayTestRequest,
+    GatewayTestResponse,
+    GatewayUpdate,
+    PromptCreate,
+    ResourceCreate,
+    ToolCreate,
+)
 
 # logging.getLogger("httpx").setLevel(logging.WARNING)  # Disables httpx logs for regular health checks
 from mcpgateway.services.audit_trail_service import get_audit_trail_service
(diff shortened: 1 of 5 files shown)


mcpgateway/services/gateway_service.py
```diff
@@ -7364,6 +7432,293 @@ async def test_gateway_connectivity(
...
+                    server_version=init.serverInfo.version,
+                    capabilities=capabilities,
+                    component_counts=component_counts or None,
+                    counts_partial=counts_partial,
+                    credential_source=credential_source,
+                    raw_preview=json.dumps(init.model_dump(by_alias=True, exclude_none=True), default=str)[:4096],
+                )
+
+            try:
+                # ponytail: hostname connect after validation, same TOCTOU window as gateway health checks
```
a-effort: `ponytail:` looks like a stray personal marker. The note is accurate and worth keeping (verified: matches the health-check probe's hostname connect), just relabel it:

```python
# SECURITY: SDK transports connect by hostname and re-resolve at connect time,
# so we can't IP-pin here like the server/discover path. Matches the existing
# health-check probe; the allowlist validation above still gates reachable hosts.
```

a-effort:
LGTM! 🚀

Minor / non blocking:

- Count badges render "1 resources" / "1 prompts" with no singularization. 
- Field errors aren't cleared on the HTTP/handshake mode toggle, while status/response/handshakeResponse/error are. Low impact since a stale field error is on a field that's hidden in handshake mode; adding setErrors({}) to the onValueChange handler would tidy it up.

Coordination heads-up (not a code issue)

This PR and #5930 both edit the test_gateway_connectivity 400-branch region (#5930 rewrites the "Invalid gateway URL" copy; this PR refactors the function and keeps the literal). Whichever merges second will hit a small rebase conflict there. Harmless, just flagging since they're in flight together. Also note this PR independently introduces the same "The MCP server URL is not allowed for testing…" phrasing that #5930 adds, so worth confirming the end state stays consistent.

mcpgateway/services/gateway_service.py
Altamimi-Dev: Good catch — that marker slipped in from a local convention and had no business in the diff. Applied your suggested wording verbatim, so the comment now explains the constraint rather than just naming it.

Altamimi-Dev:
Thanks for the thorough read — both minors are in, along with the comment relabel.

**Count badges** now singularize an exact count of 1, so it renders `1 resource` instead of `1 resources`. One deliberate exception: when `countsPartial` is true the label stays plural, since `1+ tools` reads as "at least one" and `1+ tool` would be odd. There is a test pinning each side of that.

**Field errors** are cleared on the mode toggle — `setErrors({})` alongside the existing resets in the `Tabs` `onValueChange` handler, exactly as you suggested. New test submits an empty URL, waits for the error, switches to MCP handshake, and asserts it is gone. Both new tests fail against the pre-fix component, so they are actually load-bearing.

**Security comment** relabeled with your wording verbatim (replied on the thread too).

On the **#5930 coordination note** — since #5930 is still open, there is no conflict to reconcile yet. Whichever of the two lands second takes a small rebase in the `test_gateway_connectivity` 400-branch; happy to be that one. When #5930 merges I will align this PR's "The MCP server URL is not allowed for testing…" copy with whatever wording lands there, so the allowlist message stays consistent across both paths.

Rebased onto the current `epic/ui-rewrite` head (conflict-free) and re-verified: client suite 2692 passed / 1 skipped across 148 files, `tests/unit/mcpgateway/routers/test_mcp_servers_router.py` 26 passed, lint and format clean.

marekdano (commented):
@Altamimi-Dev - nice work!

### 1. OAuth token-retrieval error leaks the raw exception to the client

  `mcpgateway/services/gateway_service.py`

  ```python
  error=f"Token retrieval failed for MCP server '{gateway.name}': {e}. Check the OAuth client credentials and token URL.",
  ```

  This embeds the raw `{e}` in the API response body. The sibling `test_gateway_connectivity` deliberately does the opposite two functions away (`gateway_service.py:7280-7283`): it logs the exception and returns a **generic** `"OAuth token retrieval failed for gateway"`. This same function also runs transport errors through `sanitize_exception_message` — so the raw `{e}` here is inconsistent with both the local convention and the established pattern, and could surface token URLs / provider internals.

  **Suggestion:** genericize or wrap with `sanitize_exception_message(str(e))`, keeping `logger.error` for the detail.

### 2. All CI checks should pass

gcgoncalves:
Hi, @Altamimi-Dev! Thanks again for the solid work you've been dishing out for this project. To keep the future merging of this branch as clean as possible, we're avoiding merging edits to the python files. Would you please separate the backend work of this pull request into a separate PR, straight into the `main` branch?

marekdano:
@Altamimi-Dev - are you free to split the code for the frontend and backend? We need to keep the backend at the PR and the PR has to be against `main`. The frontend code should in a new PR in https://github.com/contextforge-org/contextforge-web-ui
It can be something similar to what I did in https://github.com/IBM/mcp-context-forge/pull/5786

Altamimi-Dev:
Thanks for the clear direction on this, @marekdano — split done, and @gcgoncalves for the earlier review.

**This PR is now backend-only against `main`.** The two epic commits both touched `client/`, so rather than replay them I squash-ported the Python half onto `main` as one signed commit and force-pushed. The epic history is preserved on a local backup ref, so nothing is lost if we need to diff against it. Only `token_scoping.py` needed a manual resolution (main had since added the `/catalog` entries next to the `/mcp-servers/test` line); the new `test-handshake` mapping slots in right after `test` and both intents are kept.

**Frontend is contextforge-org/contextforge-web-ui#16.** It carries the mode toggle, detail rows, count badges (`3+ tools` when `countsPartial` is set), failure-class badge, raw preview and abort wiring. Following your #15 precedent, the new user-facing copy there goes through react-intl with en-US/es-ES/pt-BR keys, and `openapi.json` gained only the new path plus the two handshake schemas so the generated types resolve — the spec stays pinned at v1.0.7 otherwise.

**Review finding fixed.** The OAuth token-retrieval failure was embedding the raw exception in the response body. It is now sanitized, same as the other embed sites:

```python
return GatewayHandshakeResponse(
    success=False,
    latency_ms=_latency_ms(),
    failure_class="auth",
    error=f"Token retrieval failed for MCP server '{gateway.name}': {sanitize_exception_message(str(e))}. Check the OAuth client credentials and token URL.",
)
```

The `logger.error` above it is unchanged, so the raw detail is still available to operators. I applied the same treatment to the two equivalent embeds in `test_gateway_connectivity` over on #5930.

Verification on the retargeted branch:

- `pytest tests/unit/mcpgateway/routers/test_mcp_servers_router.py` — 26 passed, including the 9 handshake tests
- `pytest tests/unit/mcpgateway/routers/test_mcp_servers_router.py tests/unit/mcpgateway/middleware/test_token_scoping.py` — 344 passed
- Full `make test` — 21873 passed, 818 skipped, 2 xfailed, 0 failed
- `make ruff` clean, `make interrogate` 100%, `pre-commit run --files …` clean on all five touched files (ruff-format folded a few of the ported long lines; that's in the commit)
- Live smoke against a running gateway pointed at a real MCP streamable-HTTP target: success returns `negotiationPath: initialize` with real `serverName`/`serverVersion`/`protocolVersion` and populated `componentCounts`, `credentialSource: form`; a target requiring auth with no headers classifies as `auth`; a closed port classifies as `transport` with sanitized detail; a host outside the test allowlist gets the generic allowlist copy; an unauthenticated request gets `401`

Also switched the body from `Closes #5649` to `Relates to #5649` and cross-linked web-ui#16, since the issue now needs both halves — mirroring how #5786 and web-ui#15 are wired.

One coordination note: #5930 also touches the validation region of `test_gateway_connectivity` on `main`, so whichever of the two lands second takes a small rebase. #5930 first is the cheaper order since it is copy-only; the `_validate_gateway_test_target` extraction here then carries its new allowlist wording directly. Either order ends in the same place — this PR independently introduces the same "The MCP server URL is not allowed for testing…" phrasing on the handshake path.

The force-push and retarget will have cleared the earlier review state — happy to re-request whenever you have a moment.

marekdano (commented):
@Altamimi-Dev - thanks for the contribution and for moving this PR against the `main` branch.

The PR is solid overall.

It can be approved after all CI checks pass. Currently, unit test coverage fails.

Altamimi-Dev:
@marekdano thanks for the review — both blockers are addressed.

**Rebased onto current `main`.** #5930 landed in the meantime and touched the same allowlist path, so this needed care rather than a mechanical replay:

- `_validate_gateway_test_target` keeps its `return None` contract (the extraction this PR introduces), and the `test_gateway_connectivity` caller now carries #5930's merged allowlist copy — without that the rebase would have silently reverted your merged message and broken the 8 assertions in `tests/e2e/test_admin_apis.py` and `tests/unit/mcpgateway/test_admin.py`.
- The handshake path's own allowlist copy and the `sanitize_exception_message` embeds carried over unchanged.

**Coverage.** The diff-cover gate was failing at 76% because the interesting branches of `test_gateway_handshake` — stored-credential resolution, failure classification, and the SDK fallback — were only smoke-covered. Added 26 targeted tests to `tests/unit/mcpgateway/routers/test_mcp_servers_router.py`, reusing the existing handshake fixtures/helpers:

- `_classify_handshake_error` matrix (401/403 auth, connect/OS transport, `McpError` and protocol-version `RuntimeError`, invalid-response fallthrough)
- stored credentials: authorization-code with and without a stored token, client-credentials via `OAuthManager`, token-retrieval failure, basic dict headers, encoded value via `decode_auth`
- form headers overriding stored credentials (`credential_source == "form"`)
- team-scoped gateway lookup and the lookup-failure degradation to an unauthenticated probe
- negotiation fallbacks: non-JSON discover response, list-call failures on both paths, paginated `counts_partial`
- transports and failures: SSE transport with its TLS-verifying client factory, initialize timeout, and `ExceptionGroup` root-cause unwrapping

Local results on the rebased branch:

- `test_mcp_servers_router.py`: 52 passed
- diff-cover against `origin/main`: **`mcpgateway/services/gateway_service.py` 100%, 197 changed lines, 0 missing** (was 72.8%)
- `make test`: 22030 passed / 843 skipped. The one local failure (`test_jq_runner.py::test_pool_is_rebuilt_after_pid_change`) reproduces identically on unmodified `main` on macOS and is untouched by this PR.
- `make ruff` and `pre-commit` on the changed files: clean

Kept it as a single signed commit. The PR shows mergeable again — should be good for a CI rerun.

marekdano (dismissed):
LGTM 🚀

msureshkumar88 (changes requested):
Solid, well-scoped addition — matches issue #5649's backend requirements cleanly (5 files, backend-only, no schema/alembic change needed since nothing persists, no breaking change to the existing `/test` endpoint, and the shared `_validate_gateway_test_target` extraction is a nice way to avoid duplicating the SSRF allowlist/DNS-pin logic). Test coverage is thorough too — 22 new cases across negotiation paths, failure classification, and credential resolution.

One blocking issue found and reproduced locally:

**`mcpgateway/services/gateway_service.py:7661`** — `credential_source` gets set to `"form"` whenever `request.headers` is non-empty, regardless of whether those headers actually override the resolved auth header:

```python
if request.headers:
    headers.update(request.headers)
    credential_source = "form"
```

Repro: a registered gateway has stored basic auth; caller sends only an unrelated header like `{"X-Trace-Id": "abc"}`. The outbound request still carries the stored `Authorization: Basic ...` untouched, but the response reports `credential_source: "form"`. That's a direct mismatch with the security requirement in #5649 ("the credential source ... is shown explicitly"), and it's not covered by the new tests — `test_handshake_form_headers_win_over_stored` only exercises the case where the supplied headers *do* override Authorization.

Suggest keying `credential_source = "form"` off whether the merge actually changed the resolved auth-bearing header, not just whether `request.headers` was truthy.

Lang-Akshay (changes requested):
## Blocking Changes

| 1 | Security | `mcpgateway/routers/mcp_servers_router.py`; `mcpgateway/services/gateway_service.py` | 140–144; 7261–7264; 7612–7615 | A narrowed token can omit `team_id`. The route only rejects an explicitly supplied unauthorized team, while the allowlist query and gateway credential lookup then operate across all teams. A user scoped to team A can probe team B hosts and may send team B's stored credentials. This violates the two-layer token-scope/RBAC invariant (CWE-862/CWE-863, Critical). | Derive and enforce an effective team scope from the authenticated context. Require an explicit authorized team or fail closed for narrowed multi-team tokens, and apply the same scope to both the registered-host allowlist and gateway credential lookup. |
| 2 | Security | `mcpgateway/services/gateway_service.py` | 7609; 7807–7815 | The stateless probe uses the DNS-pinned IP, but the `initialize` fallback builds `hostname_url` and lets the MCP SDK re-resolve it after validation. DNS rebinding can redirect the fallback to a private/internal address, potentially carrying stored credentials (CWE-918, High). | Make the SDK fallback use the already-resolved/pinned address while preserving the original authority for `Host` and TLS SNI, or reject the stateful fallback unless a pinned transport is available. |

marekdano:
@Altamimi-Dev - are you available to look at the reviewed comments above?

Altamimi-Dev:
Thanks both, good catches. Rebased on main and pushed as one amended commit.

**`credential_source` (@msureshkumar88).** Now flips to `"form"` only when the form actually supplies a credential: an `Authorization` header, or a case-insensitive collision with a stored credential key. Your repro (stored basic auth + `X-Trace-Id`) stays `"stored"`. Pinning it also exposed that a differently-cased form header never really displaced the stored one, so both were going upstream. Fixed too.

**Team scope (@Lang-Akshay #1).** Fixed in the shared service rather than the router, so it covers the omitted-`team_id` case, both endpoints, and the legacy admin route. The new visibility predicate mirrors `_check_gateway_access` and takes admin status from `is_admin_bypass_granted` (DB-authoritative) instead of `token_teams is None`. Two deliberate calls worth flagging: public gateways stay probeable (platform-public scope per AGENTS.md), and the explicit-`team_id` path now also excludes other users' private rows.

**SDK pinning (@Lang-Akshay #2).** This reverses my earlier "can't IP-pin the SDK path" comment. Passing the pinned URL straight to `sse_client` would have broken SSE gateways, since the SDK rejects an advertised endpoint whose origin differs from the connect URL and our own SSE transport advertises an absolute one. So the pinning now sits in a small transport wrapper: dial the validated IP, keep the `Host` header, verify TLS against the hostname. It also pins the follow-up requests, which the URL-only approach missed.

Self-review over the diff turned up more in the same paths, all fixed and pinned: a caller-supplied `Host` header riding through to the pinned credentialed request; root gateways registered without a trailing slash never matching their own credentials; custom CA/mTLS ignored; `server/discover` missing the required request `_meta` and reading `serverInfo` from the wrong place; form `Authorization` failing on an unobtainable stored OAuth token; `ReadError`/`PoolTimeout` reported as "does not speak MCP"; the request path appended after the base-URL query string.

Every fix has a test that fails when that fix alone is reverted. Lint, pylint, docstrings and pre-commit clean, full suite green, 100% changed-line coverage.

Two I left alone, say the word if you'd rather I didn't: `credential_source` can't express "both stored and form" (it reports `"form"`), and mTLS without a custom CA still isn't applied, matching the three sibling client factories.

msureshkumar88 (approved):
Re-reviewed against the two CHANGES_REQUESTED reviews — verified in code (not just the comment text) that all fixes are real and each is backed by a test that fails on revert:

- `credential_source` mismatch (msureshkumar88): now keyed off actual `Authorization` presence / case-insensitive key collision, not header presence alone.
- Team-scope bypass, CWE-862/863 (Lang-Akshay #1): fixed in the shared service layer, fail-closed on omitted `team_id`, covers both endpoints and the legacy admin route.
- DNS-rebinding on the `initialize` fallback, CWE-918 (Lang-Akshay #2): `_SniPinningTransport` forces the pinned IP for connect and every follow-up request while still verifying TLS against the real hostname.
- The seven self-review fixes (Host header stripping, root-URL trailing-slash credential matching, custom CA, discover `_meta`/`serverInfo`, unobtainable-OAuth-token form fallback, ReadError/PoolTimeout classification, path-before-query construction) are all present with dedicated tests.

Also did a fresh pass independent of the review thread: scope matches issue #5649's requirements cleanly, backend-only split is correct (UI half is contextforge-web-ui#16), no schema/Alembic change needed (nothing persisted), diff is exactly the 5 claimed files with no unrelated changes, not a breaking change, and test coverage is thorough (50+ handshake-specific unit tests including team-scope deny paths, host-header injection, and TLS pinning).

Two non-blocking items for a possible follow-up, not required for this PR:
- The `server/discover` path pins TLS via a per-request httpx `extensions={"sni_hostname":...}` override, while the `initialize` fallback uses `_SniPinningTransport`. Both work, but it's two different mechanisms for the same property — a direct test asserting the discover path actually connects to the pinned IP with correct SNI would close the gap in confidence.
- `gateway_service.py` around the credential-lookup query has a bare `except Exception: gateway = None` that silently swallows real DB errors (degrading to an unauthenticated probe) with no logging — low risk, but leaves operators unable to distinguish "no gateway registered" from "DB query failed."

Approving — solid, well-tested fix for both flagged security findings, minimal, correctly scoped.

vishu-bh (approved):
LGTM 🚀

Lang-Akshay (approved):
LGTM ✅

PR #5948, opened 2026-07-28
fix: gRPC services list always empty in Admin UI

gloliss:
## Summary
Registered gRPC services are always invisible in the Admin UI (🔌 gRPC Services tab),
showing "No gRPC services registered yet." even though the REST API correctly returns them.

## Root Cause
`mcpgateway/admin.py` calls `grpc_service_mgr.list_services()` without a `page` parameter.
When `page` is not passed, `list_services()` returns a `(list, cursor)` tuple.
The list comprehension iterates over the tuple, first yielding the list object itself
(instead of GrpcService instances), causing an `AttributeError` on `.model_dump()`.

The error is swallowed by the bare `except` block, setting `grpc_services = []`.

The REST API endpoint (`GET /admin/grpc`) works correctly because it passes `page=1`.

## Fix
One-line change: add `, _` to unpack the tuple return.

```diff
- grpc_services_raw = await grpc_service_mgr.list_services(
+ grpc_services_raw, _ = await grpc_service_mgr.list_services(
```

## Verification
1. Register a gRPC service via API
2. Open Admin UI → gRPC Services tab → services now appear correctly
3. Service list, enable/disable, reflect, and delete actions all work

---
Reported-by: @gloliss


jonpspri:
Will need to re-implement without noise.

PR #5957, opened 2026-07-28
fix: propagate JSON-RPC errors from internal /rpc hop to SSE clients

### LINKED ISSUE #5956: [BUG]: SSE transport swallows JSON-RPC errors from the internal /rpc hop — clients get an empty result: {}
### 🐞 Bug Summary

When a `tools/call` (or any request-carrying method) arrives through the SSE
transport (`GET /servers/{id}/sse` + `POST /servers/{id}/message`), the session
relay executes it via an internal loopback HTTP call to `/rpc` and then builds
the SSE reply with:

```python
result = rpc_response.json()
result = result.get("result", {})
response = {"jsonrpc": "2.0", "result": result, "id": req_id}
```

(`mcpgateway/cache/session_registry.py`, `generate_response()` — present in
v1.0.6 and still on `main` today.)

If the internal `/rpc` hop returns a JSON-RPC **error** (plugin violation from a
`tool_pre_invoke` hook, an RBAC `-32003 Access denied`, an upstream failure) or
a non-JSON-RPC body (e.g. `{"detail": "..."}` from an auth middleware 401/403 on
the loopback), the `error` member is silently discarded and the SSE client
receives:

```json
{"jsonrpc": "2.0", "result": {}, "id": 2}
```

Two consequences:

1. **The failure reason is unrecoverable client-side.** Any policy/authz deny
   surfaced by a plugin or by RBAC looks identical to "the call succeeded and
   returned nothing".
2. **Strict MCP SDK clients mask it further**: the official `mcp` Python SDK
   fails pydantic validation of `CallToolResult` (`content: Field required`) on
   the empty object, so users see an unrelated validation error instead of the
   real deny.

This cost us several days of debugging: we initially concluded (wrongly) that
`tool_pre_invoke` hooks were never invoked on the SSE transport, when in fact
our call was being denied by RBAC (`tools.execute`) on the internal `/rpc` hop
and the deny was being swallowed by the relay.

---

### 🧩 Affected Component

- [x] Federation or Transports (SSE transport / `session_registry.generate_response`)

---

### 🔁 Steps to Reproduce

1. Run the gateway (v1.0.6 image) with any virtual server and a federated tool.
2. Authenticate as a user that has `servers.use` but **not** `tools.execute`
   (e.g. only the global `platform_viewer` role — note that personal-team
   `team_admin` roles are deliberately excluded from `check_any_team`
   resolution, so a fresh user without a non-personal team membership is in
   exactly this state).
3. Open `GET /servers/{id}/sse`, then `POST` a `tools/call` to the `/message`
   endpoint of the session.
4. Direct comparison: `POST /rpc` with the same token and the same `tools/call`
   body returns `{"error": {"code": -32003, "message": "Access denied",
   "data": {"method": "tools/call"}}}` — but the SSE session replies
   `{"jsonrpc": "2.0", "result": {}, "id": N}`.

The same swallowing occurs for plugin `tool_pre_invoke` violations (which reach
`/rpc` clients as `{"error": {"code": -32602, "message": "Plugin Violation:
..."}}`) and for non-JSON-RPC bodies (401 `{"detail": ...}`) on the loopback.

---

### 📉 Expected Behavior

The SSE reply should carry the JSON-RPC `error` object verbatim, exactly as the
`/rpc` endpoint returns it, per JSON-RPC 2.0 (a response object has *either*
`result` *or* `error`, and errors must not be re-encoded as empty results).

---

### 📈 Actual Behavior

`{"jsonrpc": "2.0", "result": {}, "id": N}` for every error case.

---

### 🖥️ Environment

- ContextForge v1.0.6 (`ghcr.io/ibm/mcp-context-forge:v1.0.6`,
  digest `sha256:049b8e61...`), also verified present on current `main`.
- Python transport (`RUST_MCP_MODE` off), gunicorn multi-worker, SQLite.

---

### 💡 Proposed Fix

In `generate_response()`:

- forward the `error` member verbatim when present;
- forward `result` when present (unchanged behavior);
- map non-JSON-RPC bodies to a `-32000` JSON-RPC error carrying the `detail`
  text when available, instead of a silent empty result.

PR with the fix + regression tests:

LucioFex:
## What

`SessionRegistry.generate_response()` built the SSE reply with
`rpc_response.json().get("result", {})`, which collapses any JSON-RPC `error`
returned by the internal loopback `/rpc` call — plugin `tool_pre_invoke`
violations, RBAC `-32003` denials, upstream failures — into an empty
`result: {}`. Non-JSON-RPC bodies (e.g. `{"detail": ...}` from an auth
middleware 401 on the internal hop) were swallowed the same way.

The SSE client gets `{"jsonrpc":"2.0","result":{},"id":N}` with no diagnostic
at all; strict MCP SDK clients then fail `CallToolResult` validation
(`content: Field required`), masking the real error behind an unrelated
validation error.

Fixes #5956

## How

In `generate_response()`, after the loopback `/rpc` call:

- if the response carries an `"error"` member → forward it verbatim;
- else if it carries `"result"` → forward that (unchanged behavior);
- else (non-JSON-RPC body) → map to a JSON-RPC error `-32000` carrying the
  body's `detail` text when present, or the HTTP status otherwise.

No behavior change for successful calls; the `initialize` notification flow is
untouched.

## Tests

Two regression tests added to
`tests/unit/mcpgateway/cache/test_session_registry.py`, following the existing
mock patterns in that file:

- `test_generate_response_propagates_jsonrpc_error` — a `-32003` error object
  from `/rpc` reaches the SSE transport verbatim.
- `test_generate_response_maps_non_jsonrpc_body_to_error` — a 401
  `{"detail": ...}` body becomes a `-32000` JSON-RPC error, not `result: {}`.

Verified: the full test file passes with the fix (`71 passed`), and both new
tests fail against the previous code (`2 failed` when run against unpatched
`session_registry.py`), confirming they pin the regression.

## Real-world impact

Found while integrating v1.0.6 as a policy-enforcement gateway (auth plugins +
`tool_pre_invoke` OPA/policy hooks): a user with `servers.use` but without
`tools.execute` gets every `tools/call` denied by RBAC on the internal `/rpc`
hop, but the SSE session answers `result: {}` — indistinguishable from a
successful empty result. We initially misdiagnosed it as "hooks never run on
the SSE transport". Propagating the error makes the actual deny (and any
policy-plugin violation message) visible to the caller, as it already is on
the public `/rpc` endpoint.

### DIFF
--- a/mcpgateway/cache/session_registry.py
+++ b/mcpgateway/cache/session_registry.py
@@ -2366,11 +2366,28 @@ class SessionRegistry(SessionBackend):
                         headers=headers,
                     )
                     logger.info(f"SSE RPC: Got response status {rpc_response.status_code}")
-                    result = rpc_response.json()
-                    logger.info(f"SSE RPC: Response content: {result}")
-                    result = result.get("result", {})
+                    rpc_payload = rpc_response.json()
+                    logger.info(f"SSE RPC: Response content: {rpc_payload}")
 
-                response = {"jsonrpc": "2.0", "result": result, "id": req_id}
+                if isinstance(rpc_payload, dict) and "error" in rpc_payload:
+                    # Propagate JSON-RPC error objects (plugin violations, RBAC
+                    # denials, upstream failures) back to the SSE client instead
+                    # of collapsing them into an empty ``result: {}`` — which
+                    # both hides the failure reason and breaks strict MCP SDK
+                    # clients that validate ``CallToolResult`` (missing
+                    # ``content``).
+                    response = {"jsonrpc": "2.0", "error": rpc_payload["error"], "id": req_id}
+                elif isinstance(rpc_payload, dict) and "result" in rpc_payload:
+                    response = {"jsonrpc": "2.0", "result": rpc_payload["result"], "id": req_id}
+                else:
+                    # Not a JSON-RPC envelope at all — e.g. an auth/authz
+                    # middleware body like ``{"detail": "..."}`` from a 401/403
+                    # on the internal /rpc hop. Surface it as a JSON-RPC error
+                    # rather than a silent empty result.
+                    status_code = getattr(rpc_response, "status_code", None)
+                    detail = rpc_payload.get("detail") if isinstance(rpc_payload, dict) else None
+                    message_text = str(detail) if detail else f"Internal RPC call failed (HTTP {status_code})"
+                    response = {"jsonrpc": "2.0", "error": {"code": -32000, "message": message_text}, "id": req_id}
             except JSONRPCError as e:
                 logger.error(f"SSE RPC: JSON-RPC error: {e}")
                 result = e.to_dict()
--- a/tests/unit/mcpgateway/cache/test_session_registry.py
+++ b/tests/unit/mcpgateway/cache/test_session_registry.py
@@ -515,6 +515,90 @@ async def test_generate_response_ping(registry: SessionRegistry):
 
 
 @pytest.mark.asyncio
+async def test_generate_response_propagates_jsonrpc_error(registry: SessionRegistry, stub_db, stub_services):
+    """A JSON-RPC *error* from the internal /rpc hop reaches the SSE client verbatim.
+
+    Regression test: the relay used to do ``.get("result", {})`` on the /rpc
+    response, so plugin violations / RBAC denials surfaced as an empty
+    ``result: {}`` with no diagnostic at all.
+    """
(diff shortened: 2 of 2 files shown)


jonpspri:
Closing in favour of #5988, which fixes the same JSON-RPC error propagation bug in the SSE transport with equivalent scope (both touch `session_registry.py` and tests) but a clearer description of all the concrete error cases affected (plugin violations, RBAC -32003, upstream failures). Both fixes are functionally equivalent.

Issue #5531, opened 2026-07-06
Recent activity feed V2


a-effort:
[Design](https://www.figma.com/design/H2Dr9QPvRVxCe92V8URvGT/Refiring?m=auto&node-id=4956-30489&t=mEcGWFgOUaNktFR3-1)

a-effort:
Ready to merge into feature/home: https://github.com/IBM/mcp-context-forge/commit/5430837d08569750f61c8738cb5326a4506577f3

https://github.com/user-attachments/assets/35f92b2a-2549-4aa9-b932-f167f873438e

Runs against the mock using `cd client && VITE_USE_MOCK_ACTIVITY=true npm run build` until `GET /api/logs/activity` (https://github.com/IBM/mcp-context-forge/issues/5944) ships. Without it, the feed hits its error/retry state.

a-effort:
Closing this issue as it's just waiting on the backend PR to be approved: https://github.com/IBM/mcp-context-forge/pull/5973

@jonpspri

Issue #2754, opened 2026-02-07
[FEATURE][API]: Add PATCH endpoint for user updates, deprecate PUT partial-update semantics


crivetimihai:
## Goal

Add a `PATCH /auth/email/admin/users/{email}` endpoint with proper partial-update semantics and deprecate the current `PUT` endpoint that behaves like `PATCH`.

## Why

`PUT /auth/email/admin/users/{email}` currently accepts partial payloads — omitted fields are left unchanged. This violates [RFC 9110 §9.3.4](https://www.rfc-editor.org/rfc/rfc9110#section-9.3.4) which states that PUT should replace the entire resource representation.

This is misleading for API consumers:
- OpenAPI docs show a `PUT` that looks like a full replacement but actually does partial updates
- Clients that follow HTTP semantics may send full objects unnecessarily, or worse, assume omitted fields reset to defaults
- SDKs and code generators treat PUT and PATCH differently — wrong method = wrong client behaviour

The admin UI and integration tests already rely on partial-update semantics (sending only changed fields), so this cannot be fixed by changing PUT to require full objects without breaking existing clients.

## Current State

| Endpoint | Actual Behaviour | Expected Behaviour |
|----------|------------------|-------------------|
| `PUT /auth/email/admin/users/{email}` | Partial update (omitted fields preserved) | Full replacement |
| `PATCH /auth/email/admin/users/{email}` | Does not exist | Partial update |

The router delegates to `EmailAuthService.update_user()` which already has correct partial-update logic (only applies non-None fields). The `AdminUserUpdateRequest` schema already has all fields as `Optional`.

## Proposed Implementation

1. **Add `PATCH /auth/email/admin/users/{email}`** — same handler as current PUT, same schema (`AdminUserUpdateRequest`), same service delegation
2. **Deprecate `PUT`** — keep it functional for backwards compatibility, add `Deprecation` and `Sunset` headers per [RFC 8594](https://www.rfc-editor.org/rfc/rfc8594), log a warning on each call
3. **Update OpenAPI docs** — mark PUT as deprecated in the schema, point clients to PATCH
4. **Update admin UI** — switch `admin.py` HTMX call from PUT to PATCH

```python
@email_auth_router.patch("/admin/users/{user_email}", response_model=EmailUserResponse)
@require_permission("admin.user_management")
async def patch_user(user_email: str, user_request: AdminUserUpdateRequest, ...):
    """Partially update user information (admin only)."""
    # Same implementation as current update_user
[...]

PR #5973, opened 2026-07-29
feat(api): recent activity feed endpoint (audit + security union)

### LINKED ISSUE #5944: [FEATURE][API][UI-REWRITE]: Recent activity feed backend
**Summary**
Recent activity feed backend: GET /api/logs/activity (a)udit + security union

Home (#5532) has a recent activity feed (#5531) which renders the locked `ActivityItem` contract, but there is no backend endpoint; the UI currently runs on `VITE_USE_MOCK_ACTIVITY` fixtures. This enables the feed to show real data. Contract origin: #5129.

**Endpoint**

`GET /api/logs/activity?limit=&since=` → `{ items: ActivityItem[] }`

**ActivityItem**

`id` (composite `audit:N` / `security:N`), `timestamp`, `source` (`audit|security`), `title`, `description`, `status` (`success|error|warning|info`), `resource_type`, `resource_name`, `actor`, `correlation_id`.

**Proposed approach**

Python-level UNION of `AuditTrail` + `SecurityEvent` rows with polling. Forward-compatible to another option to consider later: (add `?cursor=` + a sibling SSE endpoint `GET /api/logs/activity/stream`) with no URL or response-shape change, so the UI does not need rewriting.

**Server owns presentation**

Backend mappers `_audit_to_activity` and `_security_to_activity` produce `title`, `description`, and `status`. The UI must not re-derive these from other fields.

**Authorization** 

- `@require_permission("audit:read")` gates entry.
- `security:read` is checked inline and additively: denial narrows the feed, it never returns 403.
- Extract a non-raising `check_permission_inline()` into `mcpgateway/middleware/rbac.py` as the companion to `@require_permission`, honoring plugin hooks (not just RBAC).

**Row visibility** 

- `AuditTrail` rows are team-scoped via `token_teams` (None = admin bypass, `[]` = public-only, list = teams ∪ NULL).
- `SecurityEvent` has no `team_id`, so non-admin holders of `security:read` see only events about themselves (`user_email == caller.email`); admins see all.
- Rows with `data_classification == "restricted"` are filtered at the SQL level for non-admins so counts are not observable.

**Deferred** 
(To be tracked separately)
Free-text search / filter query params, error and warning counts.

**Acceptance**

- Endpoint returns the union shape described above.
- Deny-path tests: unauthenticated, wrong team, missing `audit:read`, feed narrowed when `security:read` absent, restricted-row filtering for non-admins.
- #5531 can drop the `VITE_USE_MOCK_ACTIVITY` mock and consume the real endpoint.

Blocks #5531. Sub-issue of #5532.

Altamimi-Dev:
Closes #5944

Adds `GET /api/logs/activity?limit=&since=` returning `{ items: ActivityItem[] }`, a Python-level union of `AuditTrail` and `SecurityEvent` rows. The React Home page Recent Activity feed (#5531) currently runs on `VITE_USE_MOCK_ACTIVITY` fixtures because no backend existed; this is that backend, matching the contract locked in `client/src/types/activity.ts` — that file lives on the UI branch `5531-recent-activity-feed-V2`, not on `main`.

## Server-owned presentation

`title`, `description`, and `status` are rendered server-side and must not be re-derived by clients. Audit status precedence is `not success` → `error`, else `requires_review` → `warning`, else `read`/`execute` → `info`, else `success`. Security status derives from severity (`CRITICAL`/`HIGH` → `error`, `MEDIUM` → `warning`, else `info`).

All contract fields are required strings, so a missing source column becomes `""` rather than null — security rows have no resource name, so `resource_name` falls back to `user_email` and `actor` to `"system"`.

## Permissions

Entry is gated by `audit:read`. Security events are **additive**: included only when the caller also holds `security:read`. A caller without it gets a narrower feed, not a 403.

That additive check needed a permission check that reports denial instead of raising, so this extracts `check_permission_inline()` in `mcpgateway/middleware/rbac.py` and makes `require_permission` delegate to it. Plugin-hook and RBAC-fallback logic now live on one shared path rather than being duplicated. The decorator behaviour table is unchanged: plugin deny → 403, plugin grant + `plugins_can_override_rbac` → call handler, plugin grant without override → RBAC fallthrough, no plugin → RBAC.

## Visibility

- **Audit rows** are team-scoped: own teams, plus NULL-team rows the caller authored. A NULL `team_id` means the writer recorded no team (e.g. `tool_service` bulk imports), not "public", so a public-only token (`[]`) sees only its own NULL-team rows.
- **Security rows** are self-only for non-admins — `SecurityEvent` has no `team_id` column.
- **`data_classification == "restricted"`** rows are excluded for non-admins **in SQL**, so their existence is not observable via counts. The predicate is OR-ed with `IS NULL` because a bare `!=` would silently drop NULL-classified rows.
- Admin bypass (`token_teams is None`) sees everything.

`get_scoped_resource_access_context()` derives admin status from the verified token, not a raw `user["is_admin"]` flag.

## Token scoping

`_check_permission_restrictions` default-denies unmapped paths, so `/api/logs/activity` is mapped to `audit:read` in `_PERMISSION_PATTERNS`. Without that entry a correctly-scoped `audit:read` token would be rejected at the middleware layer before reaching the handler. `security:read` alone does not open the route.

Scope note: this maps only the new route. The pre-existing `/api/logs/*` routes are unmapped and left for a separate issue.

## Deferred (per the issue)

Free-text search/filter params, error/warning counts, `?cursor=`, SSE streaming.

## Notes for reviewers

- `limit` defaults to 50, max 100, matching the client’s own clamp in `activity.ts`. Siblings in this file use 100/1000 — say the word if you prefer consistency with them over the client.
- `since` is strictly-after, as documented in `client/src/api/activity.ts`.
- Each source over-fetches at most `limit`, so the merged top-`limit` is exact.
- Presentation wording (`_ACTION_VERBS`, `_RESOURCE_LABELS`) is our design within the "server owns presentation" rule; the fixtures only established tone. Changing it touches the two mappers and their tests, nothing else.
- No migration: `timestamp` is already indexed and the feed caps at 100 rows, so a composite `(team_id, timestamp)` index was not worth adding.

### DIFF
--- a/docs/docs/manage/logging.md
+++ b/docs/docs/manage/logging.md
@@ -412,6 +412,7 @@ When `STRUCTURED_LOGGING_DATABASE_ENABLED=true`, you get:
 | **Performance Metrics** | `GET /api/logs/performance-metrics` | Aggregated p50/p95/p99 latencies, error rates by component |
 | **Security Events** | `GET /api/logs/security-events` | Authentication failures, threat detection, security audit |
 | **Audit Trails** | `GET /api/logs/audit-trails` | CRUD operations, data access compliance logging |
+| **Recent Activity** | `GET /api/logs/activity` | Unified newest-first feed merging audit trails and security events (powers the UI home page) |
 
 ### Example: Search Logs via API
 
@@ -428,6 +429,23 @@ curl -X POST "http://localhost:4444/api/logs/search" \
 
 # Trace all logs for a specific request
 curl "http://localhost:4444/api/logs/trace/abc-123-correlation-id" \
+  -H "Authorization: Bearer $TOKEN"
+```
+
+### Recent Activity Feed
+
+`GET /api/logs/activity` returns a single newest-first list merging audit trails and security
+events, with a server-rendered title, description and status per entry.
+
+- Entry is gated by `audit:read`. Security events are **additive**: they appear only when the
+  caller also holds `security:read`. Lacking it narrows the feed rather than returning `403`.
+- Non-admin callers see audit rows for their own teams, security events they triggered, and no
+  `restricted`-classified rows.
+- `limit` (default `50`, max `100`) caps the merged list; `since` is strictly-after.
+
+```bash
+# Newest 20 activity entries since a timestamp
+curl "http://localhost:4444/api/logs/activity?limit=20&since=2025-01-01T12:00:00Z" \
   -H "Authorization: Bearer $TOKEN"
 ```
 
--- a/mcpgateway/middleware/rbac.py
+++ b/mcpgateway/middleware/rbac.py
@@ -738,6 +738,9 @@ async def check_permission_inline(
     never raises for a denial — it returns ``False`` so callers can degrade a response
     (e.g. omit a section of a feed) instead of rejecting the request.
 
+    Layer 1 (API token scopes) is enforced here, before plugin hooks and RBAC, so direct
+    callers get the same gate as the ``@require_permission`` decorator.
+
     Args:
         user_context: Authenticated user context dict; must contain ``email``.
         permission: Permission string to check (e.g. ``"security:read"``).
@@ -769,6 +772,16 @@ async def check_permission_inline(
         False
     """
     if not user_context or not isinstance(user_context, dict) or "email" not in user_context:
+        return False
+
+    # SECURITY: Check API token scopes BEFORE plugin hooks and RBAC (Layer 1).
+    # A scoped API token must carry the required permission; this is independent of
+    # the RBAC role checks below (Layer 2). Session tokens and tokens whose scopes
+    # are empty ("inherit from RBAC") pass through — see token_scope_grants().
+    token_scopes = user_context.get("token_scopes")
+    if not token_scope_grants(token_scopes, permission):
+        # Log detailed info server-side but return generic error message to avoid permission disclosure
+        logger.warning(f"API token scope check failed: user={user_context['email']}, permission={permission}, token_scopes={token_scopes}")
         return False
 
     # First, check if any plugins want to handle permission checking
@@ -963,16 +976,6 @@ def require_permission(permission: str, resource_type: Optional[str] = None, all
             user_context = kwargs.get("user") or kwargs.get("_user") or kwargs.get("current_user") or kwargs.get("current_user_ctx")
             if not user_context or not isinstance(user_context, dict) or "email" not in user_context:
                 raise HTTPException(status_code=status.HTTP_401_UNAUTHORIZED, detail="User authentication required")
-
-            # SECURITY: Check API token scopes BEFORE RBAC (Layer 1)
-            # A scoped API token must carry the required permission; this is independent of
-            # the RBAC role checks below (Layer 2). Session tokens and tokens whose scopes
-            # are empty ("inherit from RBAC") pass through — see token_scope_grants().
-            token_scopes = user_context.get("token_scopes")
-            if not token_scope_grants(token_scopes, permission):
-                # Log detailed info server-side but return generic error message to avoid permission disclosure
-                logger.warning(f"API token scope check failed: user={user_context['email']}, permission={permission}, token_scopes={token_scopes}")
-                raise HTTPException(status_code=status.HTTP_403_FORBIDDEN, detail=_ACCESS_DENIED_MSG)
 
             team_id, check_any_team = await _resolve_team_and_check_mode(user_context, kwargs)
 
--- a/mcpgateway/middleware/rbac.py
+++ b/mcpgateway/middleware/rbac.py
@@ -720,6 +720,186 @@ def _is_mutate_permission(permission: str) -> bool:
     return parts[-1] in _MUTATE_PERMISSION_ACTIONS if len(parts) >= 2 else False
 
 
+async def check_permission_inline(
+    user_context: dict,
+    permission: str,
+    *,
+    resource_type: Optional[str] = None,
+    team_id: Optional[str] = None,
+    check_any_team: bool = False,
+    db: Optional[Session] = None,
+    request: Optional[Request] = None,
+    allow_admin_bypass: bool = True,
+) -> bool:
+    """Check a permission without raising, for additive in-handler checks.
+
+    Shares one code path with :func:`require_permission`: plugin ``HTTP_AUTH_CHECK_PERMISSION``
+    hooks are consulted first, then the standard RBAC check. Unlike the decorator this
+    never raises for a denial — it returns ``False`` so callers can degrade a response
+    (e.g. omit a section of a feed) instead of rejecting the request.
+
+    Args:
+        user_context: Authenticated user context dict; must contain ``email``.
+        permission: Permission string to check (e.g. ``"security:read"``).
+        resource_type: Optional resource type for resource-specific permissions.
+        team_id: Optional team scope for the check.
+        check_any_team: If True, grant when the permission holds in any of the user's teams.
+        db: Optional existing session; a fresh session is opened when omitted.
+        request: Optional request, used only to derive plugin content-type context.
+        allow_admin_bypass: If True, platform admins bypass the RBAC check.
+
+    Returns:
+        bool: True when the permission is granted, False otherwise.
+
+    Examples:
+        >>> import asyncio
+        >>> class DummyPS:
+        ...     def __init__(self, db):
+        ...         pass
+        ...     async def check_permission(self, **kwargs):
+        ...         return True
+        >>> from unittest.mock import patch
+        >>> with patch('mcpgateway.middleware.rbac.PermissionService', DummyPS):
+        ...     asyncio.run(check_permission_inline({"email": "u"}, "tools.read", db=object()))
+        True
+
+        Malformed context is denied rather than raising:
+        >>> asyncio.run(check_permission_inline({}, "tools.read"))
+        False
+    """
+    if not user_context or not isinstance(user_context, dict) or "email" not in user_context:
+        return False
+
+    # First, check if any plugins want to handle permission checking
+    # Third-Party
+    from cpex.framework import HttpAuthCheckPermissionPayload, HttpHookType  # pylint: disable=import-outside-toplevel
+
+    # First-Party
+    from mcpgateway.plugins import get_plugin_manager  # pylint: disable=import-outside-toplevel
+
+    plugin_manager = await get_plugin_manager()
+    if plugin_manager and plugin_manager.has_hooks_for(HttpHookType.HTTP_AUTH_CHECK_PERMISSION):
+        # Get plugin contexts from user_context (stored in request.state by HttpAuthMiddleware)
+        # These enable cross-hook context sharing between HTTP_PRE_REQUEST and HTTP_AUTH_CHECK_PERMISSION
+        plugin_context_table = user_context.get("plugin_context_table")
+        plugin_global_context = user_context.get("plugin_global_context")
+
+        # Reuse existing global context from middleware if available for consistency
+        # Otherwise create a new one (fallback for cases where middleware didn't run)
+        if plugin_global_context:
+            global_context = plugin_global_context
+        else:
+            request_id = user_context.get("request_id") or uuid.uuid4().hex
+            content_type = request.headers.get("content-type") if request and hasattr(request, "headers") else None
+            global_context = GlobalContext(
+                request_id=request_id,
+                server_id=None,
+                tenant_id=None,
+                content_type=content_type,
+            )
+
+        # Invoke permission check hook, passing plugin contexts from HTTP_PRE_REQUEST hook
+        result, _ = await plugin_manager.invoke_hook(
+            HttpHookType.HTTP_AUTH_CHECK_PERMISSION,
+            payload=HttpAuthCheckPermissionPayload(
+                user_email=user_context["email"],
+                permission=permission,
+                resource_type=resource_type,
+                team_id=team_id,
+                is_admin=user_context.get("is_admin", False),
+                auth_method=user_context.get("auth_method"),
+                client_host=user_context.get("ip_address"),
+                user_agent=user_context.get("user_agent"),
+            ),
+            global_context=global_context,
+            local_contexts=plugin_context_table,  # Pass context table for cross-hook state
+            extensions=build_request_extensions(),
+        )
+        record_plugin_metrics(current_trace_id.get(), result.metadata)
+
+        # If a plugin made a decision, respect it
+        if result and result.modified_payload and hasattr(result.modified_payload, "granted"):
+            decision_plugin = "unknown"
+            decision_reason = getattr(result.modified_payload, "reason", None)
+            result_metadata = result.metadata if isinstance(result.metadata, dict) else {}
+            if result_metadata.get("_decision_plugin"):
+                decision_plugin = str(result_metadata["_decision_plugin"])
+            for key in ("plugin_name", "plugin", "source_plugin", "handler"):
+                if decision_plugin != "unknown":
+                    break
+                plugin_name = result_metadata.get(key)
+                if plugin_name:
+                    decision_plugin = str(plugin_name)
+
+            logger.info(
+                "Plugin permission decision: plugin=%s user=%s permission=%s granted=%s reason=%s",
+                decision_plugin,
+                user_context["email"],
+                permission,
+                result.modified_payload.granted,
+                decision_reason,
+            )
+
+            if result.modified_payload.granted:
+                if settings.plugins_can_override_rbac:
+                    logger.warning(
+                        "Plugin RBAC grant override applied: plugin=%s user=%s permission=%s reason=%s",
+                        decision_plugin,
+                        user_context["email"],
+                        permission,
+                        decision_reason,
+                    )
+                    return True
+
+                logger.info(
+                    "Plugin RBAC grant decision ignored by default policy: plugin=%s user=%s permission=%s",
+                    decision_plugin,
+                    user_context["email"],
+                    permission,
+                )
+            else:
+                logger.warning(
+                    "Permission denied by plugin: plugin=%s user=%s permission=%s reason=%s",
+                    decision_plugin,
+                    user_context["email"],
+                    permission,
+                    decision_reason,
+                )
+                return False
+
+    # No plugin handled it, fall through to standard RBAC check
+    if db:
+        permission_service = PermissionService(db)
+        granted = await permission_service.check_permission(
+            user_email=user_context["email"],
+            permission=permission,
+            resource_type=resource_type,
+            team_id=team_id,
+            token_teams=user_context.get("token_teams"),
+            ip_address=user_context.get("ip_address"),
+            user_agent=user_context.get("user_agent"),
+            allow_admin_bypass=allow_admin_bypass,
+            check_any_team=check_any_team,
+        )
+    else:
+        # Create fresh db session for permission check
+        with fresh_db_session() as fresh_db:
+            permission_service = PermissionService(fresh_db)
+            granted = await permission_service.check_permission(
+                user_email=user_context["email"],
+                permission=permission,
+                resource_type=resource_type,
+                team_id=team_id,
+                token_teams=user_context.get("token_teams"),
+                ip_address=user_context.get("ip_address"),
+                user_agent=user_context.get("user_agent"),
+                allow_admin_bypass=allow_admin_bypass,
+                check_any_team=check_any_team,
+            )
+
+    return granted
+
+
 def require_permission(permission: str, resource_type: Optional[str] = None, allow_admin_bypass: bool = True):
     """Decorator to require specific permission for accessing an endpoint.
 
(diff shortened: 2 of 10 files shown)


a-effort:
Thanks for jumping in to contribute! 🙌 

--> LLM-assisted feedback: 

Clean PR: the writeup maps one-to-one to the diff, the security reasoning is explicit, and the tests exercise real SQL rather than mocks. 

Nice work: 
- Extracting check_permission_inline and having require_permission delegate to it removes real duplication, and the regression tests pin the decorator's behavior table.
- Excluding restricted rows in SQL (with the IS NULL arm) so counts aren't observable is the right call.
- Confirmed the _PERMISSION_PATTERNS entry is load-bearing: _check_permission_restrictions default-denies unmapped paths, so without it a valid audit:read token would 403 before reaching the handler.

Minor (non blocking): 
- The audit query scopes on team_id (team_id IS NULL OR team_id IN token_teams) rather than the usual get_scoped_resource_access_context "public + team + own-private" shape, because AuditTrail has no owner_email/visibility columns. Suggestion: add a one-line comment for this.
- Merge tiebreak sorts id lexically (audit:10 before audit:9) and each source's LIMIT resolves boundary ties arbitrarily; ordering each source by (timestamp DESC, id DESC) would make it deterministic. Cosmetic given sub-second timestamps.
- The check_permission_inline doctest calls the real get_plugin_manager(); it passes today but couples to global plugin state, and the unit tests already cover those paths.
- error_message is the one free-text field the feed can surface; no live path populates audit error_message today, so just a heads-up if that changes.
- limit 50/100 vs the siblings' 100/1000 looks like a deliberate match to the client clamp; fine to keep.

Altamimi-Dev:
Thanks for the careful pass — much appreciated.

Addressed in f36d1ba70:

- **Scoping comment**: added a note on the team-based shape — `AuditTrail` has no `owner_email`/`visibility` columns, so the usual `get_scoped_resource_access_context` public+team+own-private shape doesn't apply here.
- **Deterministic ordering**: both source queries now order by `(timestamp DESC, id DESC)`, with regression tests pinning the LIMIT boundary. One small clarification: the ids are UUID hex strings rather than integers, so the Python merge tiebreak was already deterministic — the genuinely arbitrary part was which rows survived each source's `LIMIT` on timestamp ties, which the SQL tiebreak now fixes.
- **Doctest**: now patches `get_plugin_manager` alongside `PermissionService`, so it no longer depends on global plugin state.

On the other two: agreed on keeping `limit` at 50/100 to match the client clamp, and noted on `error_message` — will keep an eye on it if a live path starts populating that field.

madhu-mohan-jaishankar (changes requested):
## Summary

Solid PR. The `check_permission_inline()` extraction is a faithful refactor **on this branch** — the decorator's behaviour table checks out line-by-line against the deleted block, with no drift. The SQL visibility filters are correct and genuinely exercised against real SQLite rows. Every claim in the PR description verifies against the code.

The real risk is **cross-branch drift**, not the code as written.

## Findings

### 🔴 HIGH — Post-rebase, the inline `security:read` check bypasses Layer-1 token scoping

[`mcpgateway/routers/log_search.py:884`](https://github.com/IBM/mcp-context-forge/blob/a0d651f7c/mcpgateway/routers/log_search.py#L884)

`origin/main` (commit `e77f25d9a`, #4737 "Enforce API Token Scope-Based Access Control") added `token_scope_grants(token_scopes, permission)` inside `require_permission`, *above* the region this PR extracts. `token_scope_grants` does not exist anywhere on this branch. A read-only `git merge-file` on `rbac.py` produces **zero conflicts** — the merge is textually clean, so CI stays green while the gap is live.

**Failure scenario:** API token with `scopes.permissions = ["audit:read"]`, held by a platform admin (or a user with `security:read` via RBAC).

1. `TokenScopingMiddleware` maps `/api/logs/activity` → `audit:read` ✅
2. Decorator Layer-1 → `token_scope_grants(["audit:read"], "audit:read")` ✅
3. Handler line 884 → **no Layer-1 check** → admin bypass → security events returned to a token explicitly scoped audit-only.

`request.state.token_scopes` is populated for API tokens, so this is reachable, not theoretical. This is the exact class of bug #4737 was written to close.

**Fix:** rebase, then move the Layer-1 gate *into* `check_permission_inline` (before the plugin hook) so the helper is genuinely one code path, and drop the now-duplicated copy from `require_permission`. Add a regression test: `token_scopes=["audit:read"]` → feed contains no `source == "security"` items.

### 🟠 MEDIUM — Team-scoped `security:read` roles are silently denied

[`mcpgateway/routers/log_search.py:884`](https://github.com/IBM/mcp-context-forge/blob/a0d651f7c/mcpgateway/routers/log_search.py#L884)

The decorator resolves `check_any_team=True` for this route via `_resolve_team_and_check_mode`. The inline call passes neither `team_id` nor `check_any_team`, so it evaluates in global scope only — `PermissionService._get_user_roles` aggregates team-scoped roles only when `include_all_teams=True`.

**Failure scenario:** Alice holds a `security-auditor` role scoped to `team-a` granting `audit:read` + `security:read`. She gets `200 OK` with security events silently missing. Fails closed, so it's a functional bug rather than a hole — but invisible to the caller and to CI. The repo already documents this exact mode as "incorrectly denied" in [`transports/streamablehttp_transport.py:1238-1245`](https://github.com/IBM/mcp-context-forge/blob/a0d651f7c/mcpgateway/transports/streamablehttp_transport.py#L1238-L1245).

**Fix:** pass `check_any_team=True` at line 884, or lift the `_resolve_team_and_check_mode` result into a local and reuse it.

### 🟠 MEDIUM — `team_id IS NULL` audit rows are treated as world-readable "public" rows

[`mcpgateway/routers/log_search.py:874-876`](https://github.com/IBM/mcp-context-forge/blob/a0d651f7c/mcpgateway/routers/log_search.py#L874-L876)

For `AuditTrail`, NULL `team_id` usually means the writer didn't pass one, not that the row is public. Confirmed producer: [`services/tool_service.py:2423`](https://github.com/IBM/mcp-context-forge/blob/a0d651f7c/mcpgateway/services/tool_service.py#L2423) writes `bulk_create_tools` / `bulk_update_tools` audit rows with no `team_id`, describing team-owned tools.

**Failure scenario:** a public-only token (`token_teams == []`, the narrowest possible scope) calls `GET /api/logs/activity` and receives `"Tool 'internal-payments-api' was created by alice@corp.com"` for a resource in a team it has no access to. Any future `log_action()` call site that forgets `team_id=` silently widens the feed to every `audit:read` holder.

Not a regression — the sibling `/api/logs/audit-trails` ([`log_search.py:746`](https://github.com/IBM/mcp-context-forge/blob/a0d651f7c/mcpgateway/routers/log_search.py#L746)) applies no scoping at all — but worth an explicit decision. Either narrow the public arm to `or_(AuditTrail.team_id.is_(None), AuditTrail.user_email == user_email)`, or keep the current shape, document it as accepted risk, and open a follow-up to backfill `team_id`.

### 🟡 LOW — Permission mock ignores team arguments, so no test can detect the finding above

[`tests/unit/mcpgateway/routers/test_log_search_activity.py:66-69`](https://github.com/IBM/mcp-context-forge/blob/a0d651f7c/tests/unit/mcpgateway/routers/test_log_search_activity.py#L66-L69)

`async def _check(self, **kwargs): return kwargs.get("permission") not in denied` keys only on `permission`; `team_id`, `check_any_team`, `token_teams` and `allow_admin_bypass` are discarded. The decorator's `check_any_team=True` and the inline call's `check_any_team=False` are indistinguishable to every test in the file — fixing or further regressing the MEDIUM above changes nothing in the suite.

**Fix:** have `_check` record the `check_any_team` / `team_id` it was called with, and add one test asserting the inline `security:read` check is made with `check_any_team=True`.

### 🟡 LOW — Autouse `no_plugin_manager` fixture patches a module the code never calls

[`tests/unit/mcpgateway/routers/test_log_search_activity.py:51-58`](https://github.com/IBM/mcp-context-forge/blob/a0d651f7c/tests/unit/mcpgateway/routers/test_log_search_activity.py#L51-L58)

The fixture patches `cpex.framework.get_plugin_manager`, but `check_permission_inline` does `from mcpgateway.plugins import get_plugin_manager`, which is defined locally in `mcpgateway/plugins/__init__.py:327`. The fixture is a no-op; the tests pass only because the real `get_plugin_manager` returns `None` when plugins are disabled in the test env. If plugins are ever enabled there, these tests silently start consulting real hooks.

**Fix:** `monkeypatch.setattr("mcpgateway.plugins.get_plugin_manager", AsyncMock(return_value=None))`, matching `tests/unit/mcpgateway/routers/test_catalog.py:30`. (Same defect pre-exists in `test_log_search.py:34`.)

### 🟡 LOW — New endpoint missing from the documented `/api/logs/*` table

[`docs/docs/manage/logging.md:410-414`](https://github.com/IBM/mcp-context-forge/blob/a0d651f7c/docs/docs/manage/logging.md#L410-L414) is a complete enumeration of the log-search API surface; `GET /api/logs/activity` isn't added and the diff touches no docs. One row plus a short subsection on the `audit:read` gate and additive `security:read` behaviour.

### 🟡 LOW — No deny-path test on the activity route itself

`tests/unit/mcpgateway/routers/test_log_search_activity.py` covers wrong-team and partial-permission narrowing, but has no test that a caller lacking `audit:read` gets 403, or that a missing user context gets 401 on `get_activity_feed`. `test_rbac.py:2996` covers the decorator generically, not this route. Two assertions using the existing `grant_permissions(denied={"audit:read"})` helper.

## Nits

1. `log_search.py:869` hardcodes `"restricted"` instead of `DataClassification.RESTRICTED.value`; `confidential` isn't filtered even though `audit_trail_service.py:287` treats the two identically.
2. `log_search.py:904` — the blanket `except Exception` would convert an `HTTPException` raised by a plugin `HTTP_AUTH_CHECK_PERMISSION` hook into a 500; consider re-raising `HTTPException` first.
3. `_ACTION_VERBS` (`log_search.py:333`) humanises `resource_type` underscores but not unknown actions, yielding titles like `Tool bulk_create_tools`.
4. `client/src/types/activity.ts`, cited in the PR body as the locked contract, doesn't exist in this repo — the field shape couldn't be cross-checked.
5. `test_audit_timestamp_ties_at_limit_resolve_by_id` (line 224) relies on SQLite's un-tiebroken row order to fail; it could still pass on some orderings without the SQL `desc(id)`.

Altamimi-Dev:
Thanks for the thorough pass — the Layer-1 finding in particular was a real hole, and the way you tied it to #4737 landing on `main` made it easy to reproduce. Rebased onto `main` (`e77f25d9a`) and worked through everything.

**`560f04c51` — security fixes**

- **Layer-1 bypass (high).** You were right: after the rebase, `require_permission`'s wrapper held the token-scope gate and `check_permission_inline` had none, so the additive `security:read` check ran with Layer 1 skipped. Moved the gate into `check_permission_inline` (right after the malformed-context guard) and removed the now-duplicate block from the wrapper. Decorator behaviour is unchanged — the existing `if not granted: raise HTTPException(403, ...)` produces the same status and detail; only the warning log moved into the helper. `require_any_permission` keeps its own any-of block since it doesn't delegate. #4737's Layer-1 tests in `test_rbac.py`/`test_token_scoping.py` stay green, which is the proof the refactor preserved decorator semantics. New regression: a user context with `token_scopes=["audit:read"]` and RBAC granting everything returns audit items only; adding `security:read` brings both sources back. Verified it discriminates — removing the gate makes the first test fail.
- **Team-scoped `security:read` (medium).** Added `check_any_team=True` to the inline call, matching what `_resolve_team_and_check_mode` picks for this route (no team kwarg on a read endpoint). New test asserts the `security:read` check carries `check_any_team=True`.
- **NULL `team_id` (medium).** Agreed, and we went with fail-closed: NULL `team_id` means "the writer passed no team" (`tool_service` bulk imports being the example you found), not "public". Those rows are now visible only to their actor — `and_(AuditTrail.team_id.is_(None), AuditTrail.user_email == user_email)`. `user_email is None` fails closed since SQL NULL comparison matches nothing; admin (`token_teams is None`) is unaffected. The two scoping tests were reworked to the new policy and each now includes a foreign-authored NULL-team row asserted absent.
- **Nits.** `"restricted"` → `DataClassification.RESTRICTED.value`; filtering stays restricted-only on purpose, `confidential` rows already surface as `warning` via `requires_review` (comment added). `except HTTPException: raise` added above the blanket handler so a plugin `HTTP_AUTH_CHECK_PERMISSION` hook raising 401/403 propagates instead of becoming a 500. Unknown actions are humanised via `row.action.replace("_", " ")` in both the verb fallback and the error title — `bulk_create_tools` now renders `Tool bulk create tools` / `Tool bulk create tools failed`, with a mapper test.

**`7ec264b53` — docs and test hygiene**

- `docs/docs/manage/logging.md`: added the `/api/logs/activity` row to the "Features When Enabled" table plus a **Recent Activity Feed** subsection covering the `audit:read` gate, additive `security:read`, `limit`/`since` semantics, and a `curl` example.
- Fixture patch target: good catch, the code resolves `get_plugin_manager` from `mcpgateway.plugins`, so patching `cpex.framework` left the real hook path live. Fixed in both `test_log_search_activity.py` and `test_log_search.py`.
- Permission mock now records every call's kwargs, which is what lets the `check_any_team` assertion exist.
- Added the two deny-path tests you asked for: missing `audit:read` → 403, `user=None` → 401.

**On the two remaining points**

- `client/src/types/activity.ts` isn't missing — it lives on the UI branch `5531-recent-activity-feed-V2` and its `ActivityItem` matches the endpoint's fields exactly. I've updated the PR body to name the branch so the reference isn't misleading.
- Tie tests: ran them revert-proof. Stripping `desc(AuditTrail.id)` / `desc(SecurityEvent.id)` from the two queries makes both `*_ties_at_limit_resolve_by_id` tests fail (they pick up `tie-a` instead of `tie-c`/`tie-b`), so the fixtures do discriminate against the missing tiebreak.

Verification: `tests/unit/mcpgateway/middleware` + `tests/unit/mcpgateway/routers` → 2616 passed, 21 skipped (pre-existing); doctests on `rbac.py` and `log_search.py` pass; ruff clean on the touched files.

madhu-mohan-jaishankar (changes requested):
Two things before this goes in.

**1. Rebase.** 

**2. `token_scoping.py:137` is unreachable.** `Permissions.AUDIT_READ` is colon-form `"audit:read"` (`db.py:1397`), but `TokenScopeRequest.validate_permissions` (`schemas.py:7278`) accepts dot-form only. Running the real validator against the real middleware:

```
['audit:read'] -> allowed=True    but  audit:read    -> cannot be minted
['audit.read'] -> allowed=False   and  audit.read    -> YES, mintable
```

The scope the pattern accepts can't exist; the one that can is denied. So a correctly-scoped token still gets 403 at Layer 1 — contrary to the PR description's "without that entry a correctly-scoped `audit:read` token would be rejected." Every scope list the entry could grant is already covered by the `not permissions or "*" in permissions` short-circuit above it.

`audit:read`/`security:read` are 2 of only 4 colon-form constants in `db.py`, and two dot-form precedents have merged into this same list since you opened — #5983 (`plugins.read`) and #5974 (`servers.read`), the latter being half your conflict.

The test asserts the deny direction correctly, but all three scope strings are unconstructable, so it catches deletion of the line while passing over a config that can't occur — worth switching to mintable forms alongside the fix.

Fails closed throughout, and the React Home page should work today since session auth doesn't populate `token_scopes`. Not a hole — just a ~4-line fix that's cheaper now than as a misleading entry between two correct ones.

**Non-blocking, fine as follow-ups:** `rbac.py:980-991` runs team resolution before the scope check, so denied requests do a wasted `Tool` lookup; `TestCheckPermissionInline` has no coverage for the Layer-1 gate at `rbac.py:781-785`; and a plugin raising a non-`HTTPException` at `log_search.py:894` returns 500 instead of narrowing the feed.

Rebase + the dot-form fix and I'm happy to approve.

a-effort:
Here's a suggestion for the dot-form/colon-form gap:

```diff
--- a/mcpgateway/schemas.py
+++ b/mcpgateway/schemas.py
@@ -7266,13 +7266,15 @@ class TokenScopeRequest(BaseModel):
         Examples:
             >>> TokenScopeRequest.validate_permissions(["tools.read", "resources.write"])
             ['tools.read', 'resources.write']
             >>> TokenScopeRequest.validate_permissions(["*"])
             ['*']
+            >>> TokenScopeRequest.validate_permissions(["audit:read", "security:read"])
+            ['audit:read', 'security:read']
         """
         if not v:
             return v

-        # Permission pattern: resource.action (alphanumeric with underscores)
-        permission_pattern = re.compile(r"^[a-zA-Z][a-zA-Z0-9_]*\.[a-zA-Z][a-zA-Z0-9_]*$")
+        # Permission pattern: resource.action or resource:action (alphanumeric with underscores).
+        # Colon form covers the handful of Permissions constants (audit:read, security:read,
+        # logs:read, metrics:read) that predate the dot-form convention.
+        permission_pattern = re.compile(r"^[a-zA-Z][a-zA-Z0-9_]*[.:][a-zA-Z][a-zA-Z0-9_]*$")
         validated = []
         for perm in v:
             perm = perm.strip()
             if not perm:
                 continue
             if perm == "*":
                 validated.append(perm)
                 continue
             if not permission_pattern.match(perm):
                 raise ValueError(f"Invalid permission format '{perm}'. Use 'resource.action' format (e.g., 'tools.read') or '*' for full access")
             validated.append(perm)
         return validated
```

```diff
--- a/tests/unit/mcpgateway/test_schemas_validators_extra.py
+++ b/tests/unit/mcpgateway/test_schemas_validators_extra.py
@@ -862,6 +862,9 @@
     assert TokenScopeRequest.validate_ip_restrictions([]) == []
     assert TokenScopeRequest.validate_ip_restrictions(["   ", "10.0.0.1"]) == ["10.0.0.1"]
     assert TokenScopeRequest.validate_permissions([]) == []
     assert TokenScopeRequest.validate_permissions(["   ", "tools.read", "*"]) == ["tools.read", "*"]
+    assert TokenScopeRequest.validate_permissions(["audit:read", "security:read"]) == ["audit:read", "security:read"]
+    with pytest.raises(ValueError, match="Invalid permission format"):
+        TokenScopeRequest.validate_permissions(["audit:"])
```

The mismatch is isolated to `TokenScopeRequest.validate_permissions` in `mcpgateway/schemas.py:7278`. Widening the separator character class to accept both `.` and `:` lets `audit:read` / `security:read` through minting without touching the RBAC-side exact-match semantics, which already expect colon-form (`Permissions.AUDIT_READ = "audit:read"` in `db.py`). A diff for that plus a test pinning the mintability claim in `tests/unit/mcpgateway/test_schemas_validators_extra.py:865` is available, either pasted here or opened as a suggestion.

With that change, `test_activity_feed_endpoint_requires_audit_read` in `test_token_scoping.py` becomes a regression test rather than one exercising an unconstructable state. No change needed there, just add a rebase onto current `main` to clear the `CONFLICTING` merge state before re-review. Thanks!

Altamimi-Dev:
Thanks for the careful review @madhu-mohan-jaishankar — both blocking items are addressed, and @a-effort's diagnosis of the colon/dot gap was spot on.

**Rebased onto `main`** (`41a6580a4`), head is now `73793e266`. The conflict surface was exactly the two files where #5974 and this PR insert at the same anchor:

- `mcpgateway/middleware/token_scoping.py` — the catalog entries and the activity-feed entry both land after the `/mcp-servers/test` line. Kept both, catalog first, then `("GET", r"^/api/logs/activity(?:$|/)", Permissions.AUDIT_READ)`.
- `tests/unit/mcpgateway/middleware/test_token_scoping.py` — same story with the two catalog tests and `test_activity_feed_endpoint_requires_audit_read`. All three kept.

Nothing else in the PR's file set had moved on main, so no other resolution was needed.

**Colon-form scopes** — applied the widening as suggested (credit to @a-effort for the concrete diff). `TokenScopeRequest.validate_permissions` now accepts `[.:]` as the separator, so the four legacy colon-form constants (`audit:read`, `security:read`, `logs:read`, `metrics:read`) are mintable; RBAC-side exact-match semantics are untouched. Added a doctest on the validator plus two assertions in `test_schemas_validators_extra.py`: the colon-form pass, and `audit:` still rejected so the widening doesn't turn into "any trailing separator goes".

Worth noting that `test_activity_feed_endpoint_requires_audit_read` needed no change — it was asserting against a scope that couldn't previously be minted, and now exercises a real one, which makes it an actual regression test for the pairing rather than a hypothetical.

**Verification**
- Reverting the `schemas.py` hunk makes the new assertion fail (`Invalid permission format 'audit:read'`), restoring it goes green — the test is pinned to the fix.
- `test_schemas_validators_extra.py`, `test_token_scoping.py`, `test_rbac.py`, `test_log_search.py`, `test_log_search_activity.py`: 655 passed, 9 pre-existing skips.
- `pytest --doctest-modules mcpgateway/schemas.py -k validate_permissions`: passing.
- `make ruff`, `make interrogate` (100%), `pre-commit` on the changed files: clean.
- `make test`: 21894 passed.

**Follow-ups not in this PR**, as agreed they're fine to defer: the rbac.py team-resolution-before-scope-check ordering, `TestCheckPermissionInline` Layer-1 gate coverage, and the plugin non-`HTTPException` → 500 path at `log_search.py:894`. Happy to pick any of them up separately if you'd rather they land sooner.

msureshkumar88 (changes requested):
Nice self-contained addition — the union-feed approach, server-owned presentation rule, and the additive `security:read` gating (narrow instead of 403) all match the issue spec cleanly, and the deny-path test coverage in `test_log_search_activity.py` is solid.

One blocking issue in the `check_permission_inline` extraction, plus a few non-blocking notes below.

## Blocking

**`mcpgateway/middleware/rbac.py` — Layer-1 token-scope check now runs after team resolution, and disagrees with the sibling decorator**

In `require_permission`'s wrapper (~line 980), `_resolve_team_and_check_mode()` now runs *before* `check_permission_inline()` (which does the `token_scope_grants()` check first thing inside it). Pre-PR, the scope check ran immediately and rejected with 403 before any of that work.

`_resolve_team_and_check_mode` can call `_derive_team_from_resource` (a DB `.get()` lookup) and `_derive_team_from_payload` (which parses `request.form()`), both unconditionally, even for requests that are about to be scope-denied. That's wasted work plus a resource-existence timing side-channel on every scope-denied request against a resource-id-keyed route.

Worth noting `require_any_permission` (~line 1165) still does it the old way — scope check first, team resolution second — so the two decorators now disagree on ordering. That inconsistency is probably the clearest signal this was an unintended side effect of the extraction rather than a deliberate change.

Suggested fix: reorder so the scope check happens before `_resolve_team_and_check_mode()` in `require_permission`, matching `require_any_permission` and the pre-PR behavior — e.g. do the scope check in the wrapper before calling into `check_permission_inline`, or split `check_permission_inline` so the scope gate can run standalone first.

## Suggestions (non-blocking)

- **`token_scoping.py`**: only `/api/logs/activity` gets a `_PERMISSION_PATTERNS` entry; the pre-existing `/api/logs/security-events`, `/api/logs/audit-trails`, `/api/logs/performance-metrics` stay unmapped and get default-denied by `TokenScopingMiddleware` for correctly-scoped tokens. You've called this out as deferred to a separate issue in the PR description — might be worth linking that issue number explicitly so it doesn't get lost.
- **`log_search.py` (~894)**: the additive `security:read` check via `check_permission_inline()` builds a fresh `PermissionService(db)`, duplicating the role-resolution query (and audit-log write, if enabled) the `audit:read` decorator already did earlier in the same request for non-admin callers. Minor, but a request-scoped cache would remove the duplicate DB round trip.
- **`_as_utc()` (~335)**: looks functionally identical to `ensure_timezone_aware()` in `observability_service.py` — could reuse instead of reimplementing, since the file already imports other service helpers.
- **`since` param (~866)**: accepted without UTC normalization on the way in (only outbound timestamps get `_as_utc()`'d), so a naive `since` could bind against the tz-aware columns per-session-timezone on Postgres rather than UTC. Pre-existing pattern copied from the file's other handlers, not introduced here — flagging since it's now on a 4th endpoint.
- **`log_search.py` (~895)**: `if is_admin_feed or user_email:` — `user_email` is always truthy for non-admin callers per the auth-context contract, so the `or` branch is currently a no-op. If that contract ever broke upstream, this would silently degrade to an empty security-rows feed rather than surfacing the anomaly.
- **`schemas.py`**: the colon-form regex change lets `audit:read`/`security:read` be minted directly, but `token_scope_grants()`'s wildcard-category derivation only recognizes dot-form categories, so a wildcard scope still can't cover a colon-form permission. Not a live regression (wildcard scopes aren't mintable via the current regex either way) but worth knowing if wildcard support for colon-form permissions is ever added later.

## Minor

No test currently exercises the ordering issue above (scope-denied request + kwargs that trigger team resolution), which is how it slipped through — might be worth a regression test alongside the fix.

a-effort:
LLM-assisted review of current status returns: 

---

The branch is CONFLICTING against main. Rebased locally; the ordering finding from @msureshkumar88 resolves as part of the rebase rather than needing a separate change.

**Rebase**: two conflicts, both in rbac.py

#6109 (28fe8a762) added `global_only: bool = False` to `require_permission()`, which skips team derivation. It touches the same region this PR restructures.

Conflict 1 (replaying 2acf785cc) is positional. Git can't align the `check_permission_inline` insertion against the changed signature, so it flags ~180 lines that are "keep both sides": take `check_permission_inline` whole, then main's signature including `global_only`.

Conflict 2 (replaying dd178ab4e) is the ordering question. HEAD has the scope gate before team resolution; this PR moves the gate into the helper. Resolving toward HEAD gives:

```python
token_scopes = user_context.get("token_scopes")
if not token_scope_grants(token_scopes, permission):
    logger.warning(f"API token scope check failed: user={user_context['email']}, permission={permission}, token_scopes={token_scopes}")
    raise HTTPException(status_code=status.HTTP_403_FORBIDDEN, detail=_ACCESS_DENIED_MSG)

if global_only:
    team_id, check_any_team = None, False
else:
    team_id, check_any_team = await _resolve_team_and_check_mode(user_context, kwargs)

granted = await check_permission_inline(...)
```

That is the fix @msureshkumar88 asked for, and it preserves `global_only`. Parking the ordering change as a follow-up was reasonable before, but the conflict now forces a decision: deferring means re-deleting main's scope-check block during resolution. Taking HEAD is less work and closes the finding.

**Duplicated gate**

This leaves `token_scope_grants` called twice on the decorator path. Worth a comment at each site, since it reads like leftover duplication:

- `check_permission_inline` needs its own gate because `log_search.py` calls it directly and undecorated for the additive `security:read` check (the Layer-1 hole from the Aug 4 review).
- The wrapper needs one ahead of team derivation so scope-denied requests don't pay for `_derive_team_from_resource` (a DB `.get()`) or `_derive_team_from_payload` (`await request.form()`).

Cost is nil: `token_scope_grants` is pure (list membership plus `str.partition`, no I/O). On denial the wrapper raises first, so the second call only runs on the granted path and returns a redundant `True`.

**Remaining**

1. Regression test for the ordering: nothing in the suite distinguishes the two orderings, which is how this got through. Needs a scope-denied request with kwargs that would trigger team derivation, asserting `_derive_team_from_resource` / `request.form()` are never reached.

2. `docs/docs/manage/rbac.md:507-514` goes stale with the `schemas.py` change. The note says colon-form and wildcard-suffix permissions are rejected at token-creation time; the widening makes that false for `audit:read`, `security:read`, `logs:read`, `metrics:read`. Note that `admin.oauth_clients:read` is still not mintable (the widened pattern allows one separator only), so #6109's narrower claim about the DCR routes stands. Only the general sentence needs a tweak.

3. Link the deferred scope issue: @msureshkumar88 asked for the issue number on the unmapped sibling routes (`/api/logs/security-events`, `/api/logs/audit-trails`, `/api/logs/performance-metrics`). The PR body says "a separate issue" without a number.

**Test results** (rebased branch)

22167 passed, 2 failed, 848 skipped, 2 xfailed. Doctests on `rbac.py`, `log_search.py`, `schemas.py` pass. Both failures reproduce on unmodified main:

- `test_email_auth_router.py::test_admin_get_update_delete_user`: time-bomb assertion whose deadline passed 16 Aug (`See #2754`). Failing on main for everyone; will red every PR until #2754 is handled. Flagging separately, shouldn't block you.
- `test_vault_plugin_a2a_e2e.py::test_tool_path_injects_token_and_strips_vault_header`: needs a live gateway on :8001, environmental.

Happy to push the rebased branch to your fork if that saves redoing the conflict resolution.

vishu-bh:
@Altamimi-Dev can you please resolve this merge conflict, this is good to go only conflicts needs to be resolved

madhu-mohan-jaishankar:
Thanks for the changes @Altamimi-Dev 

**One blocker: rebase onto current `main`.** You're on `41a6580a4`, 16 behind. Since then #6109 added `global_only` to `require_permission` plus an `if global_only: team_id, check_any_team = None, False` branch — that's the exact block you rewrote, and it conflicts in 2 hunks. Please keep both the parameter and the branch, passing the resolved values into `check_permission_inline`.

**Minor:** the comment at `schemas.py:7280` says colon form is legacy and covers 4 constants. Neither holds — `main `has 10, and the regex rejects 6 of them (`admin.sso_providers:*`, `admin.oauth_clients:*`) because it allows only one separator. `admin.oauth_clients:*` landed this week in #6109. Either widen the regex or fix the comment.

Happy to approve after the above are addressed.

madhu-mohan-jaishankar:
Thanks for the quick turnarounds on this @Altamimi-Dev and I'd like to get it merged.

To save you a round trip, I've pushed a resolved rebase to `pr-5973-rebase` - https://github.com/IBM/mcp-context-forge/pull/6292

Your six commits with authorship intact, 0 behind main.

The only tricky part was `rbac.py`. #6109 added `global_only` to `require_permission`, and your Layer-1 commit edits that same block - taking either side wholesale loses something. The resolution keeps `global_only` alongside the delegation to `check_permission_inline`.

msureshkumar88 (changes requested):
# Changes requested — PR #5973

## Blocking

1. **`mcpgateway/middleware/rbac.py` — `require_permission`**: move the `token_scope_grants()` check to the top of the wrapper, before `_resolve_team_and_check_mode()` — matching `require_any_permission`'s order. Add a regression test asserting `_derive_team_from_resource` / `request.form()` are never invoked when the token's scope check fails.

## Should fix

1. Update `docs/docs/manage/rbac.md:507-514` — colon-form permissions (`audit:read`, `security:read`, `logs:read`, `metrics:read`) are now mintable; the doc still says they're rejected.
2. Link the issue number for the deferred `/api/logs/*` token-scoping gap in the PR body.
3. `token_scope_grants()`'s `<category>.*` wildcard match should also recognize colon-form permissions (`audit:*` → `audit:read`).

msureshkumar88:
This PR's changes have been merged via #6292. Could a maintainer close this out?

madhu-mohan-jaishankar:
completed in https://github.com/IBM/mcp-context-forge/pull/6292

Issue #5967, opened 2026-07-29
[UI-REWRITE][API]: Catalog OAuth add flow 


a-effort:
**Summary**

Give the React MCP server catalog (#5770) an OAuth add flow. The catalog currently hides OAuth entries rather than offering them: `ServerCatalog.tsx:124` filters to `auth_type === "Open"`, so 76 of 135 entries never render. This covers the backend and frontend to turn that into a guided setup that reuses the existing per-user authorize flow.

Sub-issue of #5770.

**Current state**

- Catalog entries: 40 `Open`, 19 API Key, 76 OAuth, of 135.
- `POST /v1/catalog/{catalog_id}/register` exists (`routers/catalog.py:86`, #5969 closed). `CatalogServerRegisterBody` excludes OAuth by design, so registering an OAuth entry produces a disabled gateway with empty `oauth_config` and `oauth_required=true` (`catalog_service.py:531`).
- Issuer discovery and DCR are built, not pending. See the status updates on #5717 and #5720.
- The catalog YAML carries only the MCP endpoint URL and `auth_type`. Seeding OAuth metadata moved to #6461.
- Two register-path bugs still stand: `CatalogServerRegisterRequest.oauth_credentials` is defined but never read, and `OAuth2.1 & API Key` falls through the `skip_initialization` branch (only `OAuth2.1`/`OAuth` match) and fails without a key.

**Already reusable** (no new work)

The per-user authorize flow exists and is already React-aware: `oauth_router` exposes `/oauth/authorize/{gateway_id}`, `/callback`, `/status/{gateway_id}`, `/fetch-tools/{gateway_id}`, with a popup + `postMessage` mode built for the React UI.

The manual add-server flow already implements OAuth end to end (`OAuth2Auth.tsx`, `useMCPServerForm.handleOAuthFlow`), including popup authorize, auto-activate and fetch-tools. This flow should wrap those components rather than reimplement them.

**Flow** 

The user clicks Add on an OAuth catalog card, a modal collects what that provider needs, and the setup completes.

Registration and OAuth configuration must not be visible as two steps, as they are in the old UI where registering an OAuth entry creates a disabled gateway with an empty `oauth_config`. This requires a second call to `PUT /gateways/{id}`. Leaving that seam in the client means the modal has to decide what to do when the second call fails: resume, roll back by deleting, or strand the user with a server they did not mean to create.

Proposed change:
[...]

Issue #5770, opened 2026-07-22
[UI-REWRITE]: MCP server catalog page 


a-effort:
### Feature Summary

Build out the MCP server catalog page in the new React admin UI. [Design](https://www.figma.com/design/H2Dr9QPvRVxCe92V8URvGT/Refiring?m=auto&node-id=3835-78387&t=sYRFIUhU3KtTf4Wd-1)

Parent epic: #4093 Frontend refactor and redesign.

### API

The page consumes the v1 catalog API (`GET /v1/catalog`), which returns catalog entries plus facet values (`categories`, `auth_types`, `providers`, `all_tags`) for filters. Register actions use `POST /v1/catalog/{id}/register` (#5969).

### Sub-issues: shipping order  

**Phase 1 (MVP): register endpoint & open servers**
- [x] #5969 - add `POST /v1/catalog/{id}/register` (non-admin register endpoint). Prerequisite for the add flows below.
- [x] #5970 - `Open` MVP: one-click add for `Open` entries, browse + search + filters. Depends on #5969.

**Phase 2: servers that require API key auth**
- [ ] #5972 - API-key add flow: modal for `API Key` entries. Depends on #5969 and on per-user owner attribution and team/visibility scoping (tracked separately).

**Phase 3: servers that require OAuth**
- [ ] #5967 - OAuth add flow: seeded OAuth metadata, discovery/DCR (via #5716), per-user authorize. Depends on #5969.

**Any time (no sequence requirement)**
- [x] #5966 - catalog icon automation (bundled, no runtime egress). Independent.

### Build notes (this issue)

Cross-cutting items that live in the page itself rather than a sub-issue: card grid with loading / empty / error / catalog-disabled (404) states; All / Installed tabs; search and filters reflected in the URL so they survive reload; conditional docs link (hidden when absent) and icon fallback; i18n; vitest at the client bar.

### Candidate follow-ups (not yet filed)

Provenance/freshness fields (`catalog_source` / `catalog_version` / `verified_at`); managing already-registered entries (Disconnect / Test / status on v1); surfacing `repo` + documentation backfill; backend filter enhancements (sort, facet counts, multi-select, tags AND-match); category taxonomy cleanup; bulk add; external/federated catalog sources.

Altamimi-Dev:
[...]

PR #5974, opened 2026-07-29
feat(api): non-admin v1 catalog register endpoint

### LINKED ISSUE #5969: [FEATURE][UI-REWRITE][API]: Add non-admin catalog register endpoint
### Summary

Add `POST /v1/catalog/{id}/register` so the React catalog can register a catalog server on the same v1 surface it already lists from. Today `GET /v1/catalog` is list-only; catalog register exists only on the admin router (`POST /admin/mcp-registry/{id}/register`), which is off the v1 API the React app targets.

Sub-issue of #5770. Prerequisite for the non-OAuth catalog MVP (one-click add for `Open`, API-key modal for `API Key`) and shared by #5967. Extracted from #5967 scope so the non-OAuth MVP is not blocked on OAuth.

### Why

The add logic already exists in `catalog_service.register_catalog_server` (handles `Open` with no creds and `API Key` as a bearer token, with friendly error mapping). The only gap is a non-admin v1 entry point. Without it the React app must call an admin-router endpoint, contradicting the "use `/v1/catalog`, regenerate the client" direction.

### Scope

- `POST /v1/catalog/{id}/register`, non-admin, RBAC `servers.create`, token-scoped (mirror the auth/scoping already used by the v1 `GET /catalog` list).
- Body mirrors `CatalogServerRegisterRequest` (optional custom name, optional `api_key`); returns `CatalogServerRegisterResponse`.
- Delegates to the existing `catalog_service.register_catalog_server`.
- Feature-gated on `mcpgateway_catalog_enabled` (404 when disabled), consistent with the list endpoint.
- Regenerate the React API client so the generated types include it.
- Duplicate/already-registered and unreachable-server cases return the existing mapped messages.

### Not in scope

- OAuth config/guided setup and the OAuth-path bug fixes (`oauth_credentials` unused, `OAuth2.1 & API Key` fall-through): #5967.
- Managing already-registered entries (Disconnect/Test/status on v1): separate follow-up.
- Bulk register on v1: optional parity, can follow.

### Acceptance criteria

- [ ] `POST /v1/catalog/{id}/register` exists, RBAC `servers.create`, token-scoped, 404 when catalog disabled.
- [ ] Registering an `Open` entry with no body succeeds (one-click path).
- [ ] Registering an `API Key` entry with `api_key` stores it and registers (modal path).
- [ ] Duplicate and unreachable registrations return the existing user-facing error messages.
- [ ] React API client regenerated with the new endpoint.
- [ ] Deny-path tests: unauthenticated, missing `servers.create`, wrong team scope, feature disabled.

Altamimi-Dev:
Closes #5969

Sub-issue of #5770, extracted from #5967.

## Summary

The React catalog page lists servers from `GET /v1/catalog`, but the only way to register one was the admin router (`POST /admin/mcp-registry/{server_id}/register`). This adds a non-admin equivalent on v1:

`POST /v1/catalog/{catalog_id}/register`

- RBAC: **both** `servers.create` and `gateways.create`, each with `allow_admin_bypass=False`. Registration bottoms out in `gateway_service.register_gateway`, i.e. it creates a federated peer — the same capability the canonical `POST /v1/gateways` gates on `gateways.create` — so a `servers.create`-only token must not reach it, and platform admins need the permissions through their roles like on the admin twin
- Token-scoped via a new `_PERMISSION_PATTERNS` entry — without it, `_check_permission_restrictions` default-denies the route even for a token that holds `servers.create`. That Layer-1 list maps one permission per route; the `gateways.create` half of the conjunction is enforced by the route decorators, which run their own token-scope check per permission
- Returns 404 when `MCPGATEWAY_CATALOG_ENABLED=false`, same gate and wording as the list endpoint
- Delegates to the existing `catalog_service.register_catalog_server`, then maps its business failures onto real status codes

Business failures map to status codes on v1: unknown catalog id → **404**, already registered → **409**. Remaining failures (offline/unreachable, upstream auth failure) stay HTTP 200 with `success: false` and the service's mapped message, but the envelope's `error` field is blanked — the service fills it with raw `str(e)` exception text, which is not for non-admin eyes. The admin endpoint keeps its 200-envelope contract unchanged; no v1 clients exist yet, so this is the moment to take the status codes.

## Body schema

The body is a new minimal `CatalogServerRegisterBody` with optional `name` and `api_key`. `name` is validated through `SecurityValidator.validate_name` (same as `GatewayCreate.name`) and `api_key` is capped at 4096 characters. The existing `CatalogServerRegisterRequest` was not reused because it requires `server_id` — redundant with the path on v1 — and exposes `oauth_credentials`, which the service currently ignores (that is part of #5967). Happy to swap to the shared schema if reviewers prefer; it is a small change to the handler and two tests.

Out of scope, per the issue: OAuth config and guided setup, the `oauth_credentials` fall-through bugs (#5967), v1 disconnect/test/status, and bulk register.

## Notes for reviewers

**Base is now `main`.** Rebased per the note on #5969 that `epic/ui-rewrite` keeps React changes only. The client regeneration (`client/openapi.json`) that was in the earlier revision has been dropped — that file only exists on the epic branch, so it will ride along with the catalog UI work instead. What remains here is backend only: three `mcpgateway/` files, two test files, and the catalog doc page.

**One extra line beyond #5969's scope.** The diff also maps `GET /catalog` → `servers.read` in the same `_PERMISSION_PATTERNS` block (plus a test). That entry is what the catalog *list* page needs for permission-scoped tokens, and it belongs to #5770 — but #5770's branch is React-only now, and having two PRs insert adjacent lines into the same pattern list would conflict on merge. Folding both entries into one commit avoids that. Glad to split it into its own small PR if you would rather keep strict one-issue-per-PR; just say so.

Verified: `tests/unit/mcpgateway/routers/test_catalog.py` (18 tests), `tests/unit/mcpgateway/middleware/test_token_scoping.py`, and the `-k catalog` sweep across `tests/unit/mcpgateway` (256 tests) are green; the new dual-permission and admin-bypass tests fail without the decorator change. Also smoke-tested over HTTP with `TestClient`: 404/409/200-scrubbed/200-success envelopes, 403 for an `is_admin` caller denied `gateways.create`, and 422 for an unsafe `name` or oversized `api_key`.

### DIFF
--- a/mcpgateway/routers/catalog.py
+++ b/mcpgateway/routers/catalog.py
@@ -88,9 +88,10 @@ async def list_catalog_servers(
 @require_permission("gateways.create", allow_admin_bypass=False)
 async def register_catalog_server(
     catalog_id: str,
+    request: Request,
     body: Optional[CatalogServerRegisterBody] = None,
     db: Session = Depends(get_db),
-    _user=Depends(get_current_user_with_permissions),
+    user=Depends(get_current_user_with_permissions),
 ) -> CatalogServerRegisterResponse:
     """Register a catalog server as a gateway for the authenticated API caller.
 
@@ -102,9 +103,10 @@ async def register_catalog_server(
 
     Args:
         catalog_id: Catalog server ID to register.
+        request: FastAPI request object.
         body: Optional overrides (custom name, API key).
         db: Database session.
-        _user: Authenticated user.
+        user: Authenticated user.
 
     Returns:
         Registration result envelope.
@@ -116,9 +118,20 @@ async def register_catalog_server(
     if not settings.mcpgateway_catalog_enabled:
         raise HTTPException(status_code=404, detail="Catalog feature is disabled")
 
-    request = CatalogServerRegisterRequest(server_id=catalog_id, name=body.name, api_key=body.api_key) if body else None
+    service_request = CatalogServerRegisterRequest(server_id=catalog_id, name=body.name, api_key=body.api_key) if body else None
 
-    result = await catalog_service.register_catalog_server(catalog_id=catalog_id, request=request, db=db)
+    user_email, token_teams = get_scoped_resource_access_context(request, user)
+    is_public_only_token = token_teams is not None and len(token_teams) == 0
+    team_id = None if is_public_only_token else getattr(request.state, "team_id", None)
+
+    result = await catalog_service.register_catalog_server(
+        catalog_id=catalog_id,
+        request=service_request,
+        db=db,
+        created_by=user_email,
+        owner_email=user_email,
+        team_id=team_id,
+    )
 
     if not result.success:
         if result.message == CATALOG_REGISTER_NOT_FOUND_MSG:
--- a/mcpgateway/routers/catalog.py
+++ b/mcpgateway/routers/catalog.py
@@ -19,7 +19,7 @@ from mcpgateway.config import settings
 from mcpgateway.db import get_db
 from mcpgateway.middleware.rbac import get_current_user_with_permissions, require_permission
 from mcpgateway.schemas import CatalogListRequest, CatalogListResponse, CatalogServerRegisterBody, CatalogServerRegisterRequest, CatalogServerRegisterResponse
-from mcpgateway.services.catalog_service import catalog_service
+from mcpgateway.services.catalog_service import CATALOG_REGISTER_ALREADY_REGISTERED_MSG, CATALOG_REGISTER_NOT_FOUND_MSG, catalog_service
 
 router = APIRouter(prefix="/catalog", tags=["Catalog"])
 
@@ -81,8 +81,11 @@ async def list_catalog_servers(
     return await catalog_service.get_catalog_servers(catalog_request, db, user_email=user_email, token_teams=token_teams)
 
 
+# Registration bottoms out in gateway_service.register_gateway (a federated peer), so it requires
+# the same gateways.create as POST /v1/gateways in addition to the admin twin's servers.create.
 @router.post("/{catalog_id}/register", response_model=CatalogServerRegisterResponse)
-@require_permission("servers.create")
+@require_permission("servers.create", allow_admin_bypass=False)
+@require_permission("gateways.create", allow_admin_bypass=False)
 async def register_catalog_server(
     catalog_id: str,
     body: Optional[CatalogServerRegisterBody] = None,
@@ -91,9 +94,11 @@ async def register_catalog_server(
 ) -> CatalogServerRegisterResponse:
     """Register a catalog server as a gateway for the authenticated API caller.
 
-    Business failures (unknown catalog id, already registered, unreachable
-    server) are returned as an HTTP 200 envelope with ``success=False`` and a
-    mapped message, matching the admin registration endpoint.
+    Unknown catalog ids map to HTTP 404 and already-registered servers to HTTP 409.
+    Remaining business failures (unreachable server, auth failure) are returned as an
+    HTTP 200 envelope with ``success=False`` and a mapped message; the ``error`` field
+    is blanked because the service captures raw exception text there. The admin
+    registration endpoint keeps its 200-envelope contract unchanged.
 
     Args:
         catalog_id: Catalog server ID to register.
@@ -105,11 +110,21 @@ async def register_catalog_server(
         Registration result envelope.
 
     Raises:
-        HTTPException: If the catalog feature is disabled.
+        HTTPException: If the catalog feature is disabled (404), the catalog id is
+            unknown (404), or the server is already registered (409).
     """
     if not settings.mcpgateway_catalog_enabled:
         raise HTTPException(status_code=404, detail="Catalog feature is disabled")
 
     request = CatalogServerRegisterRequest(server_id=catalog_id, name=body.name, api_key=body.api_key) if body else None
 
-    return await catalog_service.register_catalog_server(catalog_id=catalog_id, request=request, db=db)
+    result = await catalog_service.register_catalog_server(catalog_id=catalog_id, request=request, db=db)
+
+    if not result.success:
+        if result.message == CATALOG_REGISTER_NOT_FOUND_MSG:
+            raise HTTPException(status_code=404, detail="Catalog server not found")
+        if result.message == CATALOG_REGISTER_ALREADY_REGISTERED_MSG:
+            raise HTTPException(status_code=409, detail="Server already registered")
+        result.error = None
+
+    return result
--- a/mcpgateway/routers/catalog.py
+++ b/mcpgateway/routers/catalog.py
@@ -18,7 +18,7 @@ from mcpgateway.auth_context import get_scoped_resource_access_context
 from mcpgateway.config import settings
 from mcpgateway.db import get_db
 from mcpgateway.middleware.rbac import get_current_user_with_permissions, require_permission
-from mcpgateway.schemas import CatalogListRequest, CatalogListResponse
+from mcpgateway.schemas import CatalogListRequest, CatalogListResponse, CatalogServerRegisterBody, CatalogServerRegisterRequest, CatalogServerRegisterResponse
 from mcpgateway.services.catalog_service import catalog_service
 
 router = APIRouter(prefix="/catalog", tags=["Catalog"])
@@ -79,3 +79,37 @@ async def list_catalog_servers(
     )
 
     return await catalog_service.get_catalog_servers(catalog_request, db, user_email=user_email, token_teams=token_teams)
+
+
+@router.post("/{catalog_id}/register", response_model=CatalogServerRegisterResponse)
+@require_permission("servers.create")
+async def register_catalog_server(
+    catalog_id: str,
+    body: Optional[CatalogServerRegisterBody] = None,
+    db: Session = Depends(get_db),
+    _user=Depends(get_current_user_with_permissions),
+) -> CatalogServerRegisterResponse:
+    """Register a catalog server as a gateway for the authenticated API caller.
+
+    Business failures (unknown catalog id, already registered, unreachable
+    server) are returned as an HTTP 200 envelope with ``success=False`` and a
+    mapped message, matching the admin registration endpoint.
+
+    Args:
+        catalog_id: Catalog server ID to register.
+        body: Optional overrides (custom name, API key).
+        db: Database session.
+        _user: Authenticated user.
+
+    Returns:
+        Registration result envelope.
+
+    Raises:
+        HTTPException: If the catalog feature is disabled.
+    """
+    if not settings.mcpgateway_catalog_enabled:
+        raise HTTPException(status_code=404, detail="Catalog feature is disabled")
+
+    request = CatalogServerRegisterRequest(server_id=catalog_id, name=body.name, api_key=body.api_key) if body else None
+
+    return await catalog_service.register_catalog_server(catalog_id=catalog_id, request=request, db=db)
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^\\.secrets\\.baseline$",
     "lines": null
   },
-  "generated_at": "2026-08-07T13:52:41Z",
+  "generated_at": "2026-08-07T14:56:50Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -1894,7 +1894,7 @@
         "hashed_secret": "1a6222d0fd7cf8ec9f3da30d4369f45b30e81afb",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 489,
+        "line_number": 507,
         "type": "Secret Keyword",
         "verified_result": null
       }
--- a/docs/docs/manage/catalog.md
+++ b/docs/docs/manage/catalog.md
@@ -251,6 +251,24 @@ curl -H "Authorization: Bearer $TOKEN" \
   http://localhost:4444/admin/catalog/servers
 ```
 
+### Registering a Catalog Server (API)
+
+Registration requires both `servers.create` and `gateways.create` (a catalog registration
+creates a federated gateway, the same capability as `POST /v1/gateways`). Admin bypass is
+disabled: platform admins need the permissions granted through their roles.
+
+```bash
+curl -X POST -H "Authorization: Bearer $TOKEN" \
+  -H "Content-Type: application/json" \
+  -d '{"name": "My Asana", "api_key": "sk-..."}' \
+  http://localhost:4444/v1/catalog/asana/register
+```
+
+The body is optional; `name` and `api_key` override the catalog defaults. Responses:
+`404` unknown catalog id (or catalog feature disabled), `409` already registered,
+`200` with `success: false` and a descriptive `message` for connectivity or auth failures,
+`200` with `success: true` on registration (OAuth servers register disabled until configured).
+
 ### Filtering by Tags
 
 ```bash
--- a/docs/docs/manage/catalog.md
+++ b/docs/docs/manage/catalog.md
@@ -260,7 +260,7 @@ disabled: platform admins need the permissions granted through their roles.
 ```bash
 curl -X POST -H "Authorization: Bearer $TOKEN" \
   -H "Content-Type: application/json" \
-  -d '{"name": "My Asana", "api_key": "sk-..."}' \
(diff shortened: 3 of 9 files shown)


Altamimi-Dev:
Thanks for the quick heads-up, and sorry for the wrong base — that makes sense.

Retargeted to `main`. I also dropped the `client/openapi.json` regeneration that was in the first revision, since `client/` only exists on `epic/ui-rewrite`; that regen will ride along with the catalog UI work instead. What's left here is backend only: the endpoint, the body schema, the token-scoping entry, and their tests.

One thing worth flagging: the diff also carries the one-line `GET /catalog` → `servers.read` scoping entry (plus a test). That one belongs to #5770, but since that branch is React-only now it has nowhere else to go, and two PRs inserting adjacent lines into the same `_PERMISSION_PATTERNS` list would conflict on merge. Happy to split it into its own small PR if you'd prefer to keep one issue per PR.

Full `tests/unit/mcpgateway` suite is green on this base. Ready for review whenever you get to it.

madhu-mohan-jaishankar (changes requested):
Nice, tightly-scoped PR — the token-scoping work in particular is correct in a way that's easy to get wrong. I checked both new regexes and they're properly anchored and can't shadow each other (`_check_permission_restrictions` compares method before matching, so insertion order is irrelevant here), and your default-deny reasoning in the description is accurate — without the new entries a `servers.create` token does 403 at layer 1. The feature gate sits in the right place, the deny-path tests assert real status codes, and the branch merges cleanly onto current `main` despite being behind.

Two things I'd like resolved before approving, then some follow-up-sized notes.

### 1. `allow_admin_bypass` should be `False` — `mcpgateway/routers/catalog.py:85`

```python
@require_permission("servers.create")   # allow_admin_bypass defaults to True
```

Every comparable route sets this explicitly: the admin twin (`admin.py:18003`), the sibling v1 router (`routers/mcp_servers_router.py:66`), and the other catalog admin routes (`admin.py:17947`, `:18104`, `:18130`). As written, an admin from whom `servers.create` was deliberately withheld is blocked on `/admin/mcp-registry/{id}/register` but succeeds here — `check_permission` returns `True` on the `_is_user_admin` branch before consulting roles (`services/permission_service.py:129-131`).

I don't think the `GET /v1/catalog` precedent carries over — that's a read route, and admin bypass on reads is a much easier default to justify than on a write that creates a federated peer.

### 2. Is `servers.create` the right permission for this? — `routers/catalog.py:85`, `middleware/token_scoping.py:135`

This is the one I'd like a deliberate answer on rather than a code change per se.

`register_catalog_server` bottoms out in `register_gateway(... visibility="public" ...)` (`services/catalog_service.py:466-472`) — it registers a federated peer, discovers its tools, and publishes them. The canonical route for that capability, `POST /v1/gateways`, requires `gateways.create`. After this PR a token scoped to exactly `["servers.create"]` can reach that capability through the catalog while still being refused at `/v1/gateways`.

To be clear about the blast radius: the default `team_admin` and `developer` roles carry both permissions (`bootstrap_db.py:374-375`), so this isn't an interactive-user escalation — it's a scoped-token boundary. But it's a precedent worth setting on purpose. `gateways.create`, or requiring both, would close it. Happy to defer to you if there's a reason `servers.create` is the intended gate — I'd just want it stated somewhere rather than inherited from the admin route's choice.

## Follow-ups — not blocking this PR

These are all inherited from the shared service (the admin twin behaves identically), so I don't think they're yours to fix here. Flagging because moving the capability off the admin router widens who's exposed to them:

- **No caller identity propagated** (`routers/catalog.py:115`). The handler passes no user context, so the resulting Gateway lands with `owner_email=None`, `team_id=None`, `visibility="public"`, and its federated tools get `created_by="system"` (`services/gateway_service.py:1742`). Compare `POST /v1/gateways`, which derives `owner_email`, resolves `team_id` from the token, rejects public-only tokens creating team/private resources, and captures `MetadataCapture.extract_creation_metadata`. Worth a follow-up issue against `catalog_service` — passing at least `created_by` would go a long way.

- **Raw exception text reaches the caller** (`services/catalog_service.py:524`). `error=str(e)` from an unrestricted `except Exception`, returned unchanged. That body was previously reachable only via `/admin/mcp-registry/*`, which token scoping gates on `admin.system_config` — a permission no default role holds. Post-merge the audience is any `servers.create` holder. Returning `result.message` and dropping `result.error` in the v1 handler would be a cheap mitigation if you want it in this PR.

- **200-for-client-errors is now-or-never** (`routers/catalog.py:113-115`). Unknown `catalog_id` returns `200 {"success": false}` (`catalog_service.py:305`), as does already-registered (`:320`). I understand the "unchanged client contract" argument for the admin path, but `/v1/catalog/{id}/register` has no existing clients — the React consumer is still on `epic/ui-rewrite`. Mapping to 404/409 is free right now and won't be later. Sibling v1 routes do raise real codes, including the disabled gate in this very handler. Your call, but this is the last cheap moment.

- **Schema constraints** (`schemas.py:8153-8163`). `name` and `api_key` are bare `Optional[str]` — no `max_length`, no `SecurityValidator` hook. `name` gets validated much later in `GatewayCreate`, where the failure is swallowed by the service's broad `except` and surfaces as a 200 body echoing the submitted value. A `max_length` on both plus a `field_validator` on `name` delegating to `SecurityValidator.validate_name` would turn that into a clean 422 at the edge.

## Answering your two open questions

**On reusing `CatalogServerRegisterRequest`:** your reasoning holds — it requires `server_id` (redundant with the path) and exposes `oauth_credentials` the service ignores. Keep the minimal body; no change needed.

**On the extra `GET /catalog` → `servers.read` line:** fold it in. Splitting it would put two PRs into adjacent lines of the same list for no benefit. No objection here.

## Nits

- `test_register_duplicate_passthrough` (`tests/unit/mcpgateway/routers/test_catalog.py:213-228`) asserts only `result is mock_response`, which `test_register_open_one_click_no_body` already covers — with the service mocked, nothing duplicate-specific is exercised.
- No test asserts `register_catalog_server` is *not* awaited on the disabled/403 paths.
- The middleware tests cover the match direction but not over-match — e.g. `POST /catalog/foo` (no `/register`) should stay default-denied.
- `docs/docs/manage/catalog.md` still documents only the admin registration flow.

Altamimi-Dev:
Thanks for the thorough pass on this — both blocking findings were real, and the permission one in particular was the right catch: I had matched the admin twin's `servers.create` without following the call through to what registration actually creates.

All of it is addressed in b604581e6 (rebased onto current `main`).

**Blocking 1 — `allow_admin_bypass` defaulted to `True`.** Fixed, and I took your suggestion further after tracing the call path. `register_catalog_server` bottoms out in `gateway_service.register_gateway`, i.e. it creates a federated peer — exactly the capability `POST /v1/gateways` gates on `gateways.create`. A `servers.create`-only scoped token reaching that felt like the more serious half of the finding, so the route now requires **both** permissions, each with `allow_admin_bypass=False`:

```python
@router.post("/{catalog_id}/register", response_model=CatalogServerRegisterResponse)
@require_permission("servers.create", allow_admin_bypass=False)
@require_permission("gateways.create", allow_admin_bypass=False)
```

Rationale for the conjunction rather than swapping to `gateways.create`: `servers.create` keeps parity with the admin twin (same user-visible capability, same permission), and `gateways.create` closes the scoped-token boundary. Both built-in default roles already carry both permissions (`bootstrap_db.py`), so this does not change what a `developer`/`team_admin` can do — it only stops a *narrowly scoped token* from using catalog registration as a side door into gateway creation. Cost is two `check_permission` calls on a rare, heavyweight operation, which seemed a fair trade.

One note on the Layer-1 side: I left the `_PERMISSION_PATTERNS` entry at `servers.create`. That list is one-permission-per-route (first match returns), and each `@require_permission` wrapper runs its own token-scope check, so the `gateways.create` half of the conjunction is still enforced against scoped tokens — just at the decorator rather than in the pattern list. I added a comment on the middleware test spelling that out so the next reader doesn't take the pattern entry as the full picture.

**Blocking 2 — `servers.create` reaching gateway creation.** Same fix as above; this is the finding that drove the dual gate.

**200-for-client-errors.** Taken, since you're right that this is the now-or-never moment — there are no v1 clients yet. Unknown catalog id → `404`, already registered → `409`. The two service messages are now module constants (`CATALOG_REGISTER_NOT_FOUND_MSG`, `CATALOG_REGISTER_ALREADY_REGISTERED_MSG`) so the router isn't matching on loose literals. The admin twin's 200-envelope contract is untouched.

**Raw `error=str(e)` leaking to non-admins.** Taken. Remaining failures (unreachable, upstream auth) still return a 200 envelope with the service's mapped `message`, but the router blanks `error` before returning — the raw exception text stays server-side.

**Body schema constraints.** Taken. `name` now goes through `SecurityValidator.validate_name` (same as `GatewayCreate.name`, which also picks up the `validation_max_name_length` bound), and `api_key` is capped at `max_length=4096`. The 4096 is a judgement call — there's no existing convention for API key length in the codebase — happy to change it if you'd prefer a different bound.

**Caller identity (`owner_email=None`, `created_by="system"`).** Filed as #6036 rather than fixed here, per your suggestion. It's a change to shared service behaviour that also moves the admin path's attribution, so it wants its own change and tests; the issue notes that #5974 is what widens the exposure.

**Nits.** All three done:
- `test_register_duplicate_passthrough` → `test_register_duplicate_returns_409`, plus a new `test_register_unknown_id_returns_404` and `test_register_failure_envelope_scrubs_error`, so each envelope shape now has a distinct observable assertion.
- Deny-path tests (`disabled`, `rbac_denied`, and the new dual-permission test) assert `mock.await_count == 0` — the service is provably not reached.
- Middleware test now asserts the over-match direction: `POST /catalog/foo` (no `/register`) stays default-denied.

Also added a `### Registering a Catalog Server (API)` section to `docs/docs/manage/catalog.md` — the Usage section only covered admin/startup flows, and it now documents the endpoint, the dual permission requirement, and the response codes.

**Your two questions:** the minimal `CatalogServerRegisterBody` stays as-is (path already carries the id; `oauth_credentials` is #5967's), and the `GET /catalog` → `servers.read` line stays folded in to avoid two PRs editing adjacent lines of the same pattern list — but say the word on either and I'll change it.

Verification: `test_catalog.py` (18) and `test_token_scoping.py` green, `-k catalog` across `tests/unit/mcpgateway` green (256), and the new dual-permission/admin-bypass tests fail without the decorator change. Also exercised over HTTP with `TestClient`: 404/409/scrubbed-200/success-200, 403 for an `is_admin: true` caller denied `gateways.create`, and 422 for an unsafe `name` or oversized `api_key`.

vishu-bh:
@Altamimi-Dev can you please check failing CI check. Rest code looks clean and solid. Review findings are addressed in latest commit.

Linxiushen:
Heads-up for the catalog register route added here: #6045 makes `CatalogService.register_catalog_server` require a `CatalogRegistrationContext` so every caller supplies authenticated ownership and request metadata. After rebasing, this route will need to build that context from its principal/request (including the verified personal team) and pass it as `context=`. This is intentional to prevent the non-admin path from silently recreating the unattributed gateway behavior fixed in #6036.

Altamimi-Dev:
Thanks for the heads-up — and agreed this is the right constraint. Making the context required is exactly what stops this route from quietly reintroducing the unattributed registration described in #6036, so I'd rather adapt to it than have the non-admin path be the one exception.

Plan on this side once #6045 lands: rebase, build the context in the v1 handler from the authenticated principal and the request (verified personal team included), and pass it as `context=`. I'll mirror `_catalog_registration_context` rather than invent a second shape — same `get_user_email` → `verify_team_for_user` → `MetadataCapture.extract_creation_metadata` sequence, with `team_id` only when the verified value is a string.

One wrinkle specific to this route: `mcpgateway.admin` is imported only when the admin API is enabled, so `routers/catalog.py` can't pull the helper from there and would have to duplicate ~12 lines. If you'd rather expose it as a shared builder somewhere importable — alongside the dataclass in `catalog_service`, or a small helper module — I'm happy to consume that instead. Your call; I'll follow whatever shape you settle on.

I'm deliberately not anticipating the merge: `CatalogRegistrationContext` isn't on `main` yet and CI here builds `refs/pull/5974/merge`, so referencing it now would only go red. When it does land I'll also add a router-test assertion on the `context=` kwarg (`owner_email`, `created_by`, `team_id`) so the attribution is pinned by a test rather than left implied.

And if #5974 merges first, the callsite change is a couple of lines that would land in your rebase instead — either order works, just say which you'd prefer.

mcpgateway/routers/catalog.py
```diff
@@ -79,3 +79,52 @@ async def list_catalog_servers(
...
+    Raises:
+        HTTPException: If the catalog feature is disabled (404), the catalog id is
+            unknown (404), or the server is already registered (409).
+    """
+    if not settings.mcpgateway_catalog_enabled:
+        raise HTTPException(status_code=404, detail="Catalog feature is disabled")
+
+    request = CatalogServerRegisterRequest(server_id=catalog_id, name=body.name, api_key=body.api_key) if body else None
+
+    result = await catalog_service.register_catalog_server(catalog_id=catalog_id, request=request, db=db)
```
madhu-mohan-jaishankar: `_user=Depends(...)` is bound for the decorator and never used; the service call passes no identity, so `catalog_service.py:472-477` leaves `created_by / owner_email / team_id` at `None `and hardcodes `visibility="public"`. The canonical route it claims parity with (main.py:7413-7424) derives `owner_email=get_user_email(user)`, computes `team_id`, and rejects public-only tokens creating team/private resources.

**Consequence:** a non-admin registers a catalog server → the Gateway row and all discovered tools become visible to every user in the deployment, the audit row records the creator as "system", and the creator can't later edit or delete it because ownership checks find no owner. Pre-PR, this service path was reachable only under `admin.system_config` at Layer 1; the new route drops that to `servers.create`.

**Fix:** add request: Request, use `get_scoped_resource_access_context(request, _user)` like `list_catalog_servers `does at `catalog.py:68`, and forward `owner_email/team_id/created_by` into register_gateway.

madhu-mohan-jaishankar (changes requested):

Altamimi-Dev:
Thanks for the careful read, Linxiushen — you were right that the route was authenticating the caller and then throwing that identity away.

Pushed as `983c575d3` (rebased onto current `main`, so the merge conflict is cleared too):

- The route now takes `request: Request` and derives `user_email, token_teams` via `get_scoped_resource_access_context(request, user)` — the same helper `list_catalog_servers` already uses right above it, so the two catalog routes stay consistent.
- Team derivation matches the canonical `/v1/gateways` route: public-only tokens (empty `token_teams`) get `team_id=None`; otherwise `request.state.team_id`.
- `CatalogService.register_catalog_server` takes `created_by` / `owner_email` / `team_id` and forwards them into **both** gateway-creation paths — `gateway_service.register_gateway` and the OAuth skip-initialization `DbGateway` insert. Previously either path produced a gateway the creator couldn't subsequently update or delete.
- On the hardcoded `visibility="public"`: that one is intentional and stays. Catalog registrations are deliberately platform-visible; the gap was attribution, not scope. Ownership is now stamped, so the creator retains update/delete rights while the gateway remains discoverable.
- New router and service tests pin the forwarded kwargs on both creation paths, so a future refactor that drops them fails loudly.

On #6045 (`CatalogRegistrationContext`) — as discussed, I've kept this as plain kwargs for now; once #6045 merges I'll adapt this call site to the context object in a follow-up rather than guessing at its final shape here.

madhu-mohan-jaishankar (changes requested):
Thanks for the changes @Altamimi-Dev. OK to approve once pre-commit checks are passing.

vishu-bh:
Thanks @Altamimi-Dev  for addressing review comments!

Please rebase from main and run `make pre-commit` to make secrets are not failing. Once CI passes this good to merge

Altamimi-Dev:
Rebased onto current `main` (clean, no conflicts) and refreshed `.secrets.baseline` — the new catalog doc section shifted an already-audited entry's line number (489 → 507); no new secrets, just `generated_at` + `line_number`.

`make pre-commit` now exits 0 locally with every hook passing, including 🔐 IBM Detect Secrets. Catalog/token-scoping test suites are green after the rebase.

New head is `7fcdf761d` — the workflow runs on it are sitting at *action_required* (fork PR), so they'd need an approval to actually report. Thanks for the review!

Altamimi-Dev:
Main moved again (#4654 landed and also touched `.secrets.baseline`), which put the branch back into conflict. Rebased onto `adcf9b70b` and regenerated the baseline on top — same two-line delta, no new secrets. `make pre-commit` exits 0 and the catalog/token-scoping suites are green.

New head is `c8d9b621b`, PR shows mergeable again. Workflows on it are still *action_required* pending approval.

madhu-mohan-jaishankar (approved):
LGTM

vishu-bh (approved):
LGTM 🚀

Lang-Akshay (approved):
LGTM

PR #5983, opened 2026-07-29
feat: add v1 plugin discovery API

### LINKED ISSUE #5667: [FEATURE][API]: Create a new GET /v1/plugins API endpoint for programmatic plugin listing
## 🧭 Type of Feature

## Background

The Admin UI plugins page is served by `GET /admin/plugins/partial` (`mcpgateway/admin.py:17092`), which returns a rendered HTMX/HTML fragment. A JSON sibling already exists at `GET /admin/plugins` (`mcpgateway/admin.py:17329`), but both are mounted under `/admin` and gated by `@require_permission("admin.plugins", allow_admin_bypass=False)`. That makes plugin discovery inaccessible to agents, CLI tools, the new React UI, or external integrations that shouldn't require admin-panel permissions.

A top-level `GET /v1/plugins` endpoint would expose the same plugin list on the public, versioned API surface — mirroring the `GET /v1/catalog` task in #4399.

Confirmed during investigation: both `/admin/plugins/partial` and `/admin/plugins` derive their per-plugin data from the **same source**, `PluginService.get_all_plugins()` (`mcpgateway/services/plugin_service.py:107`). The partial only differs by rendering HTML and embedding richer aggregate statistics; the per-plugin fields are identical. So the new endpoint can reuse the existing `list_plugins` logic and `PluginListResponse` schema directly.

## Proposed Endpoint

`GET /v1/plugins`

**Response model:** `PluginListResponse` (`mcpgateway/schemas.py:7883`, unchanged)

**Query parameters** (identical to the existing admin JSON route):

- `search` (str, default null) — text search in name/description/author
- `mode` (str, default null) — filter by mode (`enforce`/`permissive`/`disabled`/…)
- `hook` (str, default null) — filter by hook type
- `tag` (str, default null) — filter by tag

**Response body:**

```json
{
  "plugins_globally_enabled": true,
  "plugins": [
    {
      "name": "string",
      "description": "string",
      "author": "string",
      "version": "0.0.0",
      "mode": "enforce",
      "priority": 0,
      "hooks": [],
      "tags": [],
      "status": "enabled",
      "config_summary": {}
    }
  ],
  "total": 0,
  "enabled_count": 0,
  "disabled_count": 0
}
```

## Auth & RBAC

- Use permission `admin.plugins` with the standard `allow_admin_bypass=True` default, consistent with other top-level public routes like `GET /servers` (`mcpgateway/main.py:4113`). *(Open question — see below.)*
- Reuse `PluginService.get_all_plugins()` / `search_plugins()`; no new business logic required.

## Open questions for reviewers

1. **Permission scope:** the admin routes use `admin.plugins` with `allow_admin_bypass=False`. Should the public endpoint keep `admin.plugins` (bypass allowed) or introduce a lighter `plugins.read`-style permission? This is the main design decision.
2. **Aggregate stats:** the HTMX partial exposes `plugins_by_hook`/`plugins_by_mode`/`plugins_by_tag`/`plugins_by_author` (via `get_plugin_statistics()`), which `PluginListResponse` does **not** include. Should `/v1/plugins` stay lean (list + counts) or also surface full stats — or defer that to a separate `GET /v1/plugins/stats`? This will be good to check with @a-effort if we are going to use plugin stats in React UI.

## Tests

Add cases in the plugin endpoint test file covering:

- Unauthenticated request → 401
- Insufficient permission → 403
- Plugins globally disabled → empty list with `plugins_globally_enabled: false`
- Each query parameter (`search`, `mode`, `hook`, `tag`) filters correctly
- Response matches `PluginListResponse` schema

vishu-bh:
# Pull Request

## 🔗 Related Issue

Closes #5667

---

## 📝 Summary

Adds a new authenticated `GET /v1/plugins` discovery endpoint while preserving the existing Admin UI plugin route unchanged.

The endpoint supports search, mode, hook, and tag filters; synchronizes plugin metadata from the active runtime; and redacts configuration values from API responses. Access is enforced through the new `plugins.read` RBAC permission and token-scoping mapping.

An idempotent Alembic migration grants `plugins.read` to existing built-in roles that receive plugin catalog access, while `platform_admin` continues to inherit access through its wildcard permission.

---

## 📏 Reviewability

---

## 🏷️ Type of Change

- [ ] Bug fix
- [x] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

| Check | Command | Status |
|---|---|---|
| Pre-commit suite | `make pre-commit` | Passed |
| Doctests, full tests, web lint, docstrings, Pylint | `make doctest test lint-web interrogate pylint` | Passed |
| Full test result | `make test` | 21,057 passed; 804 skipped; 2 expected failures |

---

## ✅ Checklist

- [x] Code formatted and pre-commit checks passed

---

## 📓 Notes (optional)

The existing `/admin/plugins` endpoint continues to use `admin.plugins` and is unchanged. The new permission applies only to the new versioned API endpoint.


mcpgateway/services/plugin_service.py
```diff
@@ -16,10 +16,29 @@
 # Third-Party
 from cpex.framework import PluginManager
 from cpex.framework.models import PluginMode
+from fastapi import Request
 
 logger = logging.getLogger(__name__)
 
 
+async def sync_plugin_service_from_runtime(request: Request, plugin_service: "PluginService") -> None:
```
madhu-mohan-jaishankar, line 24: `sync_plugin_service_from_runtime `is the only genuinely new service logic, and no test executes it. Both tests that reach the handler body monkeypatch it out. the other two raise 401/403 first.

Fix: add cases to tests/unit/mcpgateway/services/test_plugin_service.py (happy path sets both caches; get_plugin_manager raising is swallowed + logged) and one router test where get_all_plugins raises → assert 500. Covers all 7 lines.

madhu-mohan-jaishankar, line 24: Line-for-line reimplementation of the private `_sync_plugin_service_from_runtime`. Moving it to` services/` is the right layer, but admin.py wasn't refactored to call it — two copies of process-global cache-mutation logic.

Fix: delete the admin.py copy and import the shared one — this should also closes the coverage gap above for free, since admin tests already exercise that path.

mcpgateway/bootstrap_db.py
```diff
@@ -456,6 +458,7 @@ async def bootstrap_default_roles(conn: Connection) -> None:
                         "teams.read",
                         "teams.join",
                         "tools.read",
+                        "plugins.read",
```
madhu-mohan-jaishankar: `plugins.read` is granted to viewer and platform_viewer. Since `config.py:1061` sets default_user_role = "platform_viewer", every SSO/local-signup user gets it. `PluginSummary `exposes `mode, status, hooks, tags, version` — so an unprivileged user can enumerate which guardrail plugins exist, which hooks they cover, and which are permissive or disabled:


`{"name":"pii_filter","mode":"permissive","status":"disabled","hooks":["tool_pre_invoke"]}`

Before this PR that required `admin.plugins`.

Fix: drop viewer/platform_viewer from the grant (leave it to team_admin/developer), or omit mode/status from the v1 summary. If intentional, state it in the PR body.

mcpgateway/bootstrap_db.py
```diff
@@ -481,6 +484,7 @@ async def bootstrap_default_roles(conn: Connection) -> None:
...
```
madhu-mohan-jaishankar: same comment as above.

madhu-mohan-jaishankar (changes requested):

gandhipratik203:
Avoiding duplicates: my permission-scope and cache-sync notes are already covered in the existing review.

One small addendum: it could be beneficial to update the migration as well. `ROLE_SCOPES` in `e4f5a6b7c8d9_add_plugins_read_permission.py` also gives `plugins.read` to `viewer` and `platform_viewer`, so existing installs would still get that access even if `bootstrap_db.py` is fixed.

Once the existing review comments and this migration follow-up are addressed, this looks good from my side.

gandhipratik203 (dismissed):
Thanks for the changes. LGTM once the CI passes!

gandhipratik203 (dismissed):

madhu-mohan-jaishankar (dismissed):
LGTM

msureshkumar88 (changes requested):
Nice work here — this closes #5667 cleanly and the design choice to introduce a dedicated `plugins.read` permission (instead of reusing `admin.plugins` with bypass, per the issue's open question) is the right call. `admin.plugins` isn't granted to any built-in role today, so reusing it would've left this effectively platform-admin-only. Config redaction (`config_summary` stripped, not filtered) is also handled well, and promoting `_sync_plugin_service_from_runtime` into a shared, public helper is a sensible way to avoid duplicating the self-heal logic between `admin.py` and the new router.

**Functionally-impacting**

1. **`viewer` / `platform_viewer` are missing `plugins.read`.** Every other read-category permission (`tools.read`, `resources.read`, `prompts.read`, `llm.read`, `a2a.read`, `gateways.read`, `servers.read`) is granted to all four non-`platform_admin` roles in `bootstrap_db.py`, but `plugins.read` is only added to `team_admin`/`developer` — both in `bootstrap_db.py` and in the migration's `ROLE_SCOPES`. Since the endpoint is read-only and already redacts config, was excluding `viewer`/`platform_viewer` intentional, or an oversight against the existing parity pattern? If unintentional, worth adding both to `ROLE_SCOPES` and the two `bootstrap_db.py` permission lists (and updating the migration test's assertions that currently pin `viewer`/`platform_viewer` as unchanged).

2. **Exceptions in the new endpoint are swallowed without any logging.** `routers/plugins.py` has no `logger` at all:
   ```python
   except Exception as exc:
       raise HTTPException(status_code=500, detail="Failed to list plugins") from exc
   ```
   The admin sibling (`admin.py`'s `list_plugins`) logs via `LOGGER.error(...)` plus a structured audit event before converting to 500. The other v1-only discovery router this PR's issue cites as precedent (`routers/catalog.py`) doesn't catch broad exceptions at all, so it at least surfaces a traceback. This new code ends up worse than either — any real bug here becomes an invisible 500 with zero log signal. Could you add a `logger.exception(...)` (or similar) before the raise?

**Suggestion (non-blocking)**

3. `docs/docs/manage/rbac.md`'s role-permission tables (`team_admin`/`developer` rows) and the permission-category table don't mention `plugins.read` yet — worth a quick update so the doc stays the RBAC source of truth it's meant to be.

Everything else — RBAC layering (Layer 1 token-scoping pattern + Layer 2 decorator, both tested), test coverage for 401/403/disabled/filters/schema, and the v1-only routing (no legacy alias, matches the `/v1/catalog` precedent) — looks solid. Happy to take another look once the above's addressed.

vishu-bh:
@msureshkumar88 thanks for raising the permission-parity question. Excluding viewer and platform_viewer is intentional.
Although configuration values are redacted, the response still exposes operational security metadata such as installed plugin names, modes, enabled/disabled status, priorities, and hook coverage. Because platform_viewer is the default role for non-admin users, granting plugins.read there would allow every local or SSO-provisioned user to enumerate deployed guardrails and identify disabled or permissive controls.
This risk was also raised during the earlier review, so we kept plugin discovery limited to team_admin and developer. I have updated the RBAC documentation to make that intentional boundary explicit.
Would you be comfortable retaining this narrower grant?

msureshkumar88 (dismissed):
Ran a real end-to-end verification of this against a live gateway process (not mocked, not `TestClient` with patched RBAC) — script and full logs are below for anyone who wants to reproduce or extend it.

## What "real" means here

- A genuine `uvicorn mcpgateway.main:app` subprocess, bootstrapped exactly the way production does (the app's own `lifespan` hook running `bootstrap_db.main()` — real `alembic upgrade head` including this PR's `e4f5a6b7c8d9_add_plugins_read_permission` migration, real default-role seeding).
- A real, fresh SQLite database per run.
- Real plugins loaded from the actual `plugins/config.yaml` shipped in the repo (`ArgumentNormalizer`, `SpanAttributeCustomizer`, etc.) via the real `PluginManager` — not stub plugin objects.
- Real users created via `POST /auth/email/admin/users`, real team membership via `POST /teams/{id}/members`, real RBAC role assignment via `POST /rbac/users/{email}/roles`, real password login via `POST /auth/email/login` — every bearer token used against `/v1/plugins` is a real, server-issued session or API token, not a hand-crafted JWT.

## Issue-fix confirmation (before/after)

1. Checked out a worktree of current `main` (`e89112842`) and hit `GET /v1/plugins` unauthenticated → **404** — confirms issue #5667 (no non-admin plugin discovery endpoint existed) was real.
2. Same request against this PR's head → **401** (route exists, RBAC now gates it) — issue closed.

## Config used

```
DATABASE_URL=sqlite:///<fresh temp file>
AUTH_REQUIRED=true
EMAIL_AUTH_ENABLED=true
PLUGINS_ENABLED=true
PLUGINS_CONFIG_FILE=plugins/config.yaml   # real, unmodified repo file
PASSWORD_CHANGE_ENFORCEMENT_ENABLED=false # test-infra convenience only, unrelated to plugin logic
JWT_SECRET_KEY / AUTH_ENCRYPTION_SECRET / PLATFORM_ADMIN_PASSWORD = randomly generated per run
```

## Commands run

```
.venv/bin/python e2e_pr5983/run_e2e.py --label main_baseline --baseline-only   # against a `main` worktree
.venv/bin/python e2e_pr5983/run_e2e.py --label pr_head                         # against this PR's head (c74ffdd5a)

# plus a broader regression sweep on everything this e2e run touches:
.venv/bin/python -m pytest \
  tests/unit/mcpgateway/routers/test_plugins.py \
  tests/unit/mcpgateway/services/test_plugin_service.py \
  tests/unit/mcpgateway/db/test_plugins_read_permission_migration.py \
  tests/unit/mcpgateway/middleware/test_rbac.py \
  tests/unit/mcpgateway/test_admin_plugin_runtime.py \
  tests/unit/mcpgateway/routers/test_teams.py \
  tests/unit/mcpgateway/services/test_role_service.py -q
```

## Output

**main baseline:** `1/1 assertions passed` — `GET /v1/plugins` unmapped (404).

**PR head:** `40/40 assertions passed`, covering:
- `GET /v1/plugins` unauthenticated → 401; route now exists and is RBAC-gated.
- Real RBAC Layer 2, both directions: `platform_viewer` (the default role every new user gets) and `viewer` → **403**; `developer`/`team_admin` → **200**. This is the info-disclosure fix from the earlier review actually holding up against a live permission check, not just a unit-test mock.
- Response shape: `plugins_globally_enabled`, real plugin list from the shipped config, `ArgumentNormalizer` present.
- **Config redaction holds for real plugins with real (non-empty) config blocks** — every `config_summary` in the response is `{}` even though the underlying `PluginService.get_all_plugins()` objects carry actual config dicts.
- Filters (`search`, `mode`) against the real catalog.
- **Layer 1 token-scoping, both directions**: a scoped API token without `plugins.read` → 403 even though the owning user's RBAC role would allow it; a token that carries `plugins.read` → 200.
- `GET /admin/plugins` (existing route, refactored in this PR to call the now-shared `sync_plugin_service_from_runtime`) still works.
- The runtime self-heal round-trips live: `PUT /admin/plugins {enabled:false}` → next `GET /v1/plugins` immediately reflects `plugins_globally_enabled:false` + empty list, no restart; re-enabling recovers it on the very next request.

**Regression sweep:** all green (a handful of skips are pre-existing "flaky in parallel"/"covered elsewhere" markers, unrelated to this change).

## One heads-up, not a blocker

The branch currently shows `mergeable: CONFLICTING` against `main` — checked with `git merge-tree`, and it's a single-file conflict confined to `.secrets.baseline` (both sides regenerated it independently), no functional code conflicts. Straightforward rebase before merge.

Approving — the fix matches the issue, the two points raised in my earlier review (missing logging, and the viewer/platform_viewer question) are both resolved correctly, and this now has live verification behind it rather than just static review.

msureshkumar88 (approved):

PR #5985, opened 2026-07-29
 feat(ui-rewrite): (De)Activate resources button

### LINKED ISSUE #5408: [UI-REWRITE]: Add Activate/Deactivate action to the Resources view overflow menu
### 🧭 Type of Feature

### 🧭 Epic
**Title:** Resources view parity with Admin UI
**Goal:** Allow operators to activate/deactivate a resource directly from the new `/client` Resources view, matching the legacy Admin UI behaviour.

---

### 🧑🏻‍💻 User Story
**As a:** gateway operator
**I want:** an Activate/Deactivate action in the per-resource overflow menu of the Resources view
**So that:** I can disable a resource (taking it out of availability) and re-enable it without editing or deleting it.

#### Description
In the Resources view, the per-resource overflow ("⋯") menu currently contains only **View Details**. Add a new menu item that toggles the resource's active state:

- A resource is **active** by default.
- When the resource is **active**, the menu item reads **Deactivate**.
- When the resource is **inactive**, the menu item reads **Activate**.
- Behaviour must match the legacy Admin UI (deactivating removes the resource from availability; reactivating restores it).
- The status dot on the card (and the status dot in the details panel) must reflect the new state after refresh. Note: the card currently hard-codes an "active" status dot (`bg-tool-status-active`); it should be made to reflect `resource.enabled`.

Suggested menu order: **View Details → Activate/Deactivate** and **Delete**.

#### ✅ Acceptance Criteria
```gherkin
Scenario: Deactivate an active resource
  Given a resource that is active (enabled)
  And the resource overflow menu is open
  When I click "Deactivate"
  Then the resource is deactivated via the API
  And the status reflects "Inactive"
  And the menu item now reads "Activate"
  And a success toast is shown

Scenario: Activate an inactive resource
  Given a resource that is inactive
  And the resource overflow menu is open
  When I click "Activate"
  Then the resource is activated via the API
  And the status reflects "Active"
  And the menu item now reads "Deactivate"
  And a success toast is shown

Scenario: Failure handling
  Given the activate/deactivate request fails
  Then an error toast is shown
  And the resource's displayed status is unchanged
```

---

### 🛠️ Implementation Notes

**Frontend**
- `client/src/pages/Resources.tsx` — in `ResourceCard`, add a `DropdownMenuItem` whose label toggles on `resource.enabled` (`Deactivate` when enabled, `Activate` when disabled); add an `onToggleResource` prop. Add a `handleToggleResource` in the `Resources` page (mirroring the existing handlers), call the API, show success/error toasts, and `refetch()` the list. Make the card status dot reflect `resource.enabled` instead of the hard-coded active color.
- `client/src/components/resources/ResourceDetailsPanel.tsx` — optionally surface the same toggle action so it stays consistent with the card; the existing status dot already reads `resource.enabled`.
- `client/src/api/resources.ts` — add a `setState(id, activate)` method.

**API**
- Canonical endpoint: `POST /resources/{resource_id}/state?activate=true|false` (requires `resources.update` permission) — see `mcpgateway/main.py:5779`.

**i18n**
- Add strings for the new actions (e.g. `resources.card.activate`, `resources.card.deactivate`) plus success/error toast messages, across `en-US`, `es-ES`, `pt-BR` (`client/src/i18n/locales/*/resources.json`).

**Tests**
- Unit tests for `ResourceCard` (correct label based on `enabled`, handler invocation) and the `Resources.tsx` toggle flow (success + failure toasts).

gcgoncalves:
# Pull Request

## 🔗 Related Issue

Closes #5408

---

## 📝 Summary

Add Activate/Deactivate button to resources, with optimistic update

---

## 📏 Reviewability

---

## 🏷️ Type of Change

- [ ] Bug fix
- [X] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

### DIFF
--- a/client/src/api/resources.ts
+++ b/client/src/api/resources.ts
@@ -64,4 +64,21 @@ export const resourcesApi = {
     const validId = validateResourceId(id);
     return api.delete(`/resources/${validId}`);
   },
+
+  /**
+   * Set resource state (activate or deactivate).
+   *
+   * Returns the canonical updated resource so callers can patch their cache
+   * directly instead of refetching the whole list.
+   *
+   * @param id - The resource ID
+   * @param activate - True to activate, false to deactivate
+   */
+  setState: (
+    id: string,
+    activate: boolean,
+  ): Promise<{ status: string; message: string; resource: ResourceRead }> => {
+    const validId = validateResourceId(id);
+    return api.post(`/resources/${validId}/state?activate=${activate}`);
+  },
 };
--- a/client/src/pages/Resources.tsx
+++ b/client/src/pages/Resources.tsx
@@ -72,9 +72,11 @@ function buildGroups(
 const ResourceGroupCard = memo(function ResourceGroupCard({
   group,
   onViewGroup,
+  onToggleResource,
 }: {
   group: ResourceGroup;
   onViewGroup: (group: ResourceGroup) => void;
+  onToggleResource?: (id: string, currentState: boolean) => void;
 }) {
   const intl = useIntl();
   const visibleResources = group.resources.slice(0, MAX_VISIBLE_RESOURCE_BADGES);
@@ -135,6 +137,26 @@ const ResourceGroupCard = memo(function ResourceGroupCard({
               <DropdownMenuItem onClick={handleView}>
                 {intl.formatMessage({ id: "resources.card.viewDetails" })}
               </DropdownMenuItem>
+              {onToggleResource && group.resources.length === 1 && (
+                <DropdownMenuItem
+                  onClick={() => {
+                    const resource = group.resources[0];
+                    onToggleResource(resource.id, resource.enabled ?? true);
+                  }}
+                  aria-label={intl.formatMessage(
+                    {
+                      id: group.resources[0].enabled
+                        ? "resources.card.deactivateAriaLabel"
+                        : "resources.card.activateAriaLabel",
+                    },
+                    { name: group.resources[0].name },
+                  )}
+                >
+                  {group.resources[0].enabled
+                    ? intl.formatMessage({ id: "resources.card.deactivate" })
+                    : intl.formatMessage({ id: "resources.card.activate" })}
+                </DropdownMenuItem>
+              )}
             </DropdownMenuContent>
           </DropdownMenu>
         </div>
@@ -223,7 +245,7 @@ export function Resources() {
     error,
     refetch,
     setData: setResourcesData,
-  } = useQuery<ResourceRead[]>("/resources?limit=0");
+  } = useQuery<ResourceRead[]>("/resources?limit=0&include_inactive=true");
   const { data: gatewaysData } = useQuery<CursorPaginatedGatewaysResponse>(
     "/gateways?limit=0&include_pagination=true",
     {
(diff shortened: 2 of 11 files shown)


client/src/pages/Resources.tsx
```diff
@@ -374,6 +396,48 @@ export function Resources() {
...
+          intl.formatMessage(
+            {
+              id: newState
+                ? "resources.toast.activateSuccess"
+                : "resources.toast.deactivateSuccess",
+            },
+            { name: resourceName },
+          ),
+        );
+        await refetch();
```
vishu-bh: Could we keep refetch() outside this rollback path? The mutation may succeed while the refresh fails, causing the UI to revert and report failure even though the backend state changed. Handling refresh failure separately—or updating the cache from the mutation response—would keep client and server consistent.

client/src/pages/Resources.tsx
```diff
@@ -374,6 +396,48 @@ export function Resources() {
     [setResourcesData, intl],
   );
 
+  const handleToggleResource = useCallback(
```
vishu-bh, line 404: Could we prevent another toggle for this resource while its mutation is pending? Multiple clicks can capture stale currentState, and out-of-order requests, refetches, or rollbacks may leave the UI inconsistent with the backend. A per-resource pending flag or latest-mutation token would make this deterministic.

client/src/api/resources.ts
```diff
@@ -64,4 +64,14 @@ export const resourcesApi = {
     const validId = validateResourceId(id);
     return api.delete(`/resources/${validId}`);
   },
+
+  /**
+   * Set resource state (activate or deactivate)
+   * @param id - The resource ID
+   * @param activate - True to activate, false to deactivate
+   */
+  setState: (id: string, activate: boolean): Promise<void> => {
```
vishu-bh: This endpoint returns status, message, and the updated resource, so Promise<void> does not match the runtime contract. Could this use a typed response and return the canonical resource to the caller? That would also reduce the need for an immediate refetch.

client/src/pages/Resources.tsx
vishu-bh: Also, Could we add tests for activation, deactivation, optimistic state, success/error toasts, API-failure rollback, refresh failure after a successful mutation, and repeated clicks? This PR currently changes no tests, despite #5408 explicitly requesting label/handler and success/failure coverage.

vishu-bh (changes requested):
Approach looks clean just some things to check before merging

client/src/pages/Resources.tsx
```diff
@@ -374,6 +396,60 @@ export function Resources() {
...
+      const resource = validResources.find((r) => r.id === id);
+      const resourceName = resource?.name || id;
+
+      // Optimistic update
+      setResourcesData((prev) =>
+        prev?.map((r) => (r?.id === id ? { ...r, enabled: newState } : r)),
+      );
+
+      try {
+        const { resource: updated } = await resourcesApi.setState(id, newState);
```
vishu-bh, line 419: Could we serialize the updated resource using aliases, or add a typed response model here? The endpoint currently nests `resource.model_dump()`, which produces snake_case fields such as `updated_at` and `gateway_id`, while `resourcesApi.setState()` types the response as camelCase ResourceRead.
When the client spreads this response into its cache, enabled updates correctly, but multi-word fields are duplicated instead of replaced. For example, the UI keeps reading the stale updatedAt while the response adds a separate updated_at.
The new tests currently mock a camelCase response, so they do not expose the real endpoint shape. Using resource.model_dump(by_alias=True)—ideally through a response model—and adding an HTTP-level assertion for updatedAt would keep the runtime response aligned with the generated frontend type.

vishu-bh (changes requested):
Code looks good and previous findings are addressed. 
Can you check this last thing as well.

vishu-bh (approved):
LGTM 🚀

PR #5994, opened 2026-07-30
Remove remaining dead feature flags (oauth discovery, gRPC reflection/TLS, catalog auto health check, performance tracking)

### LINKED ISSUE #5848: [CHORE]: MUST remove or wire up 8 dead feature flags in config.py (settable but read by zero code)
## Summary

Audit of all 230 boolean settings in `mcpgateway/config.py` found **8 flags that no code reads** — verified against `settings.<name>` references, direct `os.getenv("UPPER_CASE")` reads, dynamic `getattr(settings, ...)` access (all such calls use literal strings; none match), and the test suite. They are dead configuration: settable via environment, documented, but completely inert.

**Three observability flags are excluded from this issue and tracked separately in #5849** because active observability work is about to begin.

## The dead flags

| Flag | `config.py` line | Claimed behavior |
|---|---|---|
| `elasticsearch_enabled` | 2158 | Send logs to Elasticsearch |
| `syslog_enabled` | 2161 | Send logs to syslog |
| `webhook_logging_enabled` | 2164 | Send logs to webhook endpoints |
| `oauth_discovery_enabled` | 966 | Enable OAuth AS metadata discovery (RFC 8414) |
| `performance_tracking_enabled` | 1993 | Enable performance tracking and metrics |
| `mcpgateway_grpc_reflection_enabled` | 1126 | Enable gRPC server reflection by default |
| `mcpgateway_grpc_tls_enabled` | 1129 | Enable TLS for gRPC connections by default |
| `mcpgateway_catalog_auto_health_check` | 1150 | Automatically health check catalog servers |

## Why this is a MUST

These are **user-facing no-ops, not harmless internals**. Every flag appears in `.env.example`; several are surfaced in the Helm chart (`charts/mcp-stack/values.yaml`, `values.schema.json`, README) and in docs (`configuration.md`, `catalog.md`, `grpc-services.md`, `tuning.md`, OAuth docs). An operator can set `ELASTICSEARCH_ENABLED=true` or `MCPGATEWAY_GRPC_TLS_ENABLED=true`, reasonably expect an effect, and get silent nothing — a security-relevant trap in the TLS case especially.

## Suggested resolution (per-flag decision required)

The fix is **not** blanket deletion — each flag needs one of two resolutions:

1. **Remove** the flag + all references in `.env.example`, charts, and docs — right answer for the three logging sinks (`elasticsearch_enabled`, `syslog_enabled`, `webhook_logging_enabled`) if the sinks were never implemented (they look like planned companions to `structured_logging_external_enabled` that never landed).
2. **Wire up** the flag to the behavior it documents — likely right for `oauth_discovery_enabled`, `mcpgateway_grpc_reflection_enabled`, `mcpgateway_grpc_tls_enabled`, `mcpgateway_catalog_auto_health_check`, and `performance_tracking_enabled`, where the gated feature exists but ignores its toggle. Wiring up must include tests proving the flag actually gates behavior (both on and off paths).

Flags that get wired up rather than removed should be split into follow-up issues if the implementation effort is non-trivial.

## Acceptance criteria

- For every flag in the table: either the flag and all its documentation/chart references are gone, or code demonstrably reads it and tests cover both states.
- `grep -rn` for each removed flag name (snake_case and UPPER_CASE) returns no hits outside changelog-style historical docs.
- `make test` and lint gates pass.

pratyushsinghal7:
# Pull Request

## 🔗 Related Issue

Closes #5848 (the remaining five flags; the three logging-sink flags were removed in #5851)

---

## 📝 Summary

Removes the last five dead feature flags from `mcpgateway/config.py` — `oauth_discovery_enabled`, `performance_tracking_enabled`, `mcpgateway_grpc_reflection_enabled`, `mcpgateway_grpc_tls_enabled`, and `mcpgateway_catalog_auto_health_check`. No code reads any of them (no `settings.<name>` references, no `os.getenv` reads, no matching `getattr(settings, ...)` literals, no test references), so setting e.g. `MCPGATEWAY_GRPC_TLS_ENABLED=true` silently does nothing — a security-relevant trap in the TLS case.

The issue suggested wiring up may be right for some of these, but in each case the real behavior is already controlled elsewhere, so removal keeps a single source of truth:

- gRPC reflection and TLS are configured **per service** via the `GrpcService` DB model (`reflection_enabled`, TLS fields) — the global defaults were never consulted.
- The active performance tracking flag is `mcpgateway_performance_tracking` (read in `mcpgateway/admin.py`); `performance_tracking_enabled` was a dead twin.
- OAuth RFC 8414 discovery and catalog health checks run unconditionally today; if a kill switch is wanted, it can be added with tests in a follow-up.

Also removes the corresponding entries from the Helm chart (`values.yaml`, `values.schema.json`, chart README) and docs (`configuration.md`, `tuning.md`, `catalog.md`, `grpc-services.md`, `oauth-troubleshooting.md`, `oauth-authorization-code-ui-design.md`). None were present in `.env.example`. Remaining mentions are only the generated `docs/docs/design/images/classes.svg` diagram and a historical entry in `charts/mcp-stack/CHANGELOG.md`.

---

## 📏 Reviewability

- [ ] The linked issue is not labeled `triage` (#5848 currently carries the `triage` label)
- [x] Tests are included with the code they validate (no tests referenced these settings; removal only)

---

## 🏷️ Type of Change

- [ ] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

| Ruff                      | `ruff check mcpgateway/config.py` | ✅ pass |
| YAML lint                 | `yamllint -c .yamllint charts/mcp-stack/values.yaml` | ✅ pass |
| Schema JSON valid         | `python -c "json.load(open('charts/mcp-stack/values.schema.json'))"` | ✅ pass |
| Config tests              | `pytest tests/unit/mcpgateway/test_config.py tests/unit/mcpgateway/test_config_env_isolation.py` | ✅ 193 passed |
| Dead-flag check           | `grep -rn` for each removed name (snake_case and UPPER_CASE) | ✅ no hits outside generated `classes.svg` and chart `CHANGELOG.md` |

Full `make test` was not run locally; relying on CI.

---

## ✅ Checklist

- [x] Tests added/updated for changes (n/a — settings removal, no tests referenced them)

### DIFF
--- a/charts/mcp-stack/README.md
+++ b/charts/mcp-stack/README.md
@@ -295,7 +295,6 @@ For detailed guidance on resource limits and process management, see `docs/docs/
 | mcpContextForge.config.MCPGATEWAY_DIRECT_PROXY_TIMEOUT | string | `"30"` |  |
 | mcpContextForge.config.MCPGATEWAY_CATALOG_ENABLED | string | `"true"` |  |
 | mcpContextForge.config.MCPGATEWAY_CATALOG_FILE | string | `"mcp-catalog.yml"` |  |
-| mcpContextForge.config.MCPGATEWAY_CATALOG_AUTO_HEALTH_CHECK | string | `"true"` |  |
 | mcpContextForge.config.MCPGATEWAY_CATALOG_CACHE_TTL | string | `"3600"` |  |
 | mcpContextForge.config.MCPGATEWAY_CATALOG_PAGE_SIZE | string | `"100"` |  |
 | mcpContextForge.config.MCPGATEWAY_UI_TOOL_TEST_TIMEOUT | string | `"60000"` |  |
@@ -375,7 +374,6 @@ For detailed guidance on resource limits and process management, see `docs/docs/
 | mcpContextForge.config.PERFORMANCE_THRESHOLD_HTTP_REQUEST_MS | string | `"500.0"` |  |
 | mcpContextForge.config.PERFORMANCE_THRESHOLD_RESOURCE_READ_MS | string | `"1000.0"` |  |
 | mcpContextForge.config.PERFORMANCE_THRESHOLD_TOOL_INVOCATION_MS | string | `"2000.0"` |  |
-| mcpContextForge.config.PERFORMANCE_TRACKING_ENABLED | string | `"true"` |  |
 | mcpContextForge.config.PERMISSION_AUDIT_ENABLED | string | `"false"` |  |
 | mcpContextForge.config.PLUGINS_CLIENT_MTLS_CA_BUNDLE | string | `""` |  |
 | mcpContextForge.config.PLUGINS_CLIENT_MTLS_CERTFILE | string | `""` |  |
@@ -706,9 +704,6 @@ For detailed guidance on resource limits and process management, see `docs/docs/
 | mcpContextForge.config.MCPGATEWAY_GRPC_ENABLED | string | `"false"` |  |
 | mcpContextForge.config.MCPGATEWAY_GRPC_TIMEOUT | string | `"30"` |  |
 | mcpContextForge.config.MCPGATEWAY_GRPC_MAX_MESSAGE_SIZE | string | `"4194304"` |  |
-| mcpContextForge.config.MCPGATEWAY_GRPC_REFLECTION_ENABLED | string | `"true"` |  |
-| mcpContextForge.config.MCPGATEWAY_GRPC_TLS_ENABLED | string | `"false"` |  |
-| mcpContextForge.config.PERFORMANCE_TRACKING_ENABLED | string | `"true"` |  |
 | mcpContextForge.config.PERFORMANCE_THRESHOLD_DATABASE_QUERY_MS | string | `"100.0"` |  |
 | mcpContextForge.config.PERFORMANCE_THRESHOLD_RESOURCE_READ_MS | string | `"1000.0"` |  |
 | mcpContextForge.config.PERFORMANCE_THRESHOLD_TOOL_INVOCATION_MS | string | `"2000.0"` |  |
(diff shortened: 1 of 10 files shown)


ebarkhordar (commented):
The removal itself checks out on my read at `e7a3f61`: none of the five names appear anywhere under `mcpgateway/` once the `config.py` lines are gone, and the gRPC pair really is per-service through the `GrpcService` model rather than a global default.

One file looks missed, and it is the one users end up with. `.env.example` still documents all five, with text that asserts they do something:

- `:1403` `# OAUTH_DISCOVERY_ENABLED=true`
- `:1736` `# MCPGATEWAY_CATALOG_AUTO_HEALTH_CHECK=true` ("Automatically health check catalog servers on startup and periodically")
- `:3608` `# MCPGATEWAY_GRPC_REFLECTION_ENABLED=true`
- `:3617` `# MCPGATEWAY_GRPC_TLS_ENABLED=false` ("Enable TLS for gRPC connections by default")
- `:3702` `# PERFORMANCE_TRACKING_ENABLED=true`

`make setup` (Makefile:401) and `scripts/lib/common.sh:254` copy that file straight to `.env`, and `docs/docs/manage/configuration.md:21` points people at it as the starting point. Since `Settings` is `extra="ignore"` (`config.py:2932`), uncommenting one is silent rather than an error, so the TLS trap you describe survives in the place it is most likely to be hit.

`charts/mcp-stack/CHANGELOG.md:21` also names `MCPGATEWAY_GRPC_REFLECTION_ENABLED`, but that is a historical entry for #3550, so I would leave it alone.

Not blocking, and it is your call whether it belongs here or in a follow-up.

PR #11, opened 2025-05-30
Added mcpgateway wrapper

kevalmahajan:
Added a wrapper to connect MCP clients to the MCP gateway, making sure they can access all the tools, resources, and prompts available in the MCP Gateway. This update supports clients that use stdio instead of SSE, so they are compatible with MCP gateway.

The goal here is to make sure that even MCP clients that don’t support SSE can still get full access to everything the MCP gateway has to offer.

PR #6003, opened 2026-07-30
feat: consume Cpex control execution records

### LINKED ISSUE #5785: [FEATURE]: Consume CPEX control execution records and emit configurable enforcement telemetry
### 🧭 Type of Feature

- [x] New component or integration
- [ ] Other

### 🧭 Epic

**Title:** CPEX control-enforcement observability in ContextForge

**Goal:** Consume structured control-execution records returned by CPEX, enrich them with trusted gateway context, and emit bounded, vendor-neutral OpenTelemetry span attributes that can be renamed or dropped through configuration.

**Why now:** Downstream consumers need enforcement metrics, not only counts of configured controls. ContextForge currently exports `tool.invoke` telemetry and limited plugin metadata, but it cannot reliably describe every control invocation, outcome, duration, or applied action without parsing plugin-specific metadata or exception strings.

This issue is the ContextForge companion to [contextforge-org/cpex#130](https://github.com/contextforge-org/cpex/issues/130). CPEX #130 defines the framework-owned execution-record contract. This issue begins once a compatible CPEX version publishes those records to ContextForge.

---

### Problem

ContextForge currently has three related but incomplete paths:

1. `tool.invoke` OpenTelemetry spans contain tool, request, identity, success, duration, and error attributes.
2. CPEX observability creates hook-chain/per-plugin internal spans with limited `plugin.*` attributes.
3. `PluginResult.metadata` can be consumed into dedicated `plugin.metrics.<plugin>` spans after deny-by-default sanitization.

None provides a complete, normalized enforcement view for one tool invocation:

- controls invoked before and after tool execution
- controls that matched or applied an action
- requested versus effective allow/deny result
- per-control duration, timeout, cancellation, or error status
- structured reason and stable error code
- aggregate counts and final control-chain result
- stable tool, agent, artifact, and binding context

Parsing `PluginViolationError` messages or adding a custom adapter for every plugin is not acceptable. It is brittle, leaks implementation details, creates inconsistent semantics, and does not report successful controls.

---

### 🧑🏻‍💻 User Story 1

**As a:** ContextForge operator

**I want:** CPEX control-execution records emitted as correlated OpenTelemetry telemetry

**So that:** downstream observability systems can calculate control invocation, trigger, enforcement, error, and latency metrics without parsing logs or plugin-specific payloads

#### ✅ Acceptance Criteria

```gherkin
Scenario: All controls allow a tool invocation
  Given CPEX returns structured execution records for tool pre- and post-invocation hooks
  When ContextForge completes the tool invocation
  Then the tool trace contains bounded aggregate control attributes
  And each evaluated control has a structured per-control result
  And the tool result remains successful

Scenario: A pre-invocation control denies
  Given a CPEX record identifies a blocking denial before the tool executes
  When ContextForge emits the tool trace
  Then the telemetry identifies the enforcement point and denying control
  And the telemetry contains a stable error code and sanitized reason
  And the tool is not reported as having executed successfully

Scenario: A post-invocation control denies
  Given the tool itself completes successfully
  And a post-invocation control rejects the result
  When ContextForge emits telemetry
  Then tool execution success and final control enforcement outcome remain distinguishable
  And downstream consumers do not need to infer the control decision from span error text

Scenario: A control times out or errors
  Given CPEX returns a timeout or error execution record
  When ContextForge applies configured CPEX on_error semantics
  Then telemetry preserves the control execution status
  And telemetry preserves the effective allow/deny result
  And telemetry emission does not change the request outcome
```

---

### 🧑🏻‍💻 User Story 2

**As a:** ContextForge integrator

**I want:** control span attributes to use a neutral default namespace with configurable rename/drop rules

**So that:** the open-source gateway remains vendor-neutral while downstream systems can consume their required schema

#### ✅ Acceptance Criteria

```gherkin
Scenario: Default open-source configuration
  Given control telemetry is enabled
  And no custom mapping is configured
  When ContextForge emits enforcement telemetry
  Then canonical attributes use a documented vendor-neutral CPEX namespace
  And no downstream product-specific attribute names are hard-coded

Scenario: Exact attribute mapping
  Given an operator maps cpex.control.result.allowed to controls.result.allow
  When the attribute is emitted
  Then only the configured destination key is exported

Scenario: Dynamic per-control mapping
  Given an operator maps cpex.control.results.*.artifact.name to controls.results.*.artifact.name
  When a control result is emitted
  Then the wildcard matches one validated path segment
  And the captured segment is preserved in the destination key

Scenario: Dynamic attribute removal
  Given an operator removes cpex.control.results.*.result.reason
  When matching control results are emitted
  Then matching attributes are absent from internal and external telemetry sinks

Scenario: Mapping collision
  Given two source attributes map to the same destination
  When configuration is loaded
  Then ambiguous configuration is rejected or handled using documented deterministic precedence
  And no attribute is silently overwritten
```

---

### 📐 Proposed Design

```mermaid
flowchart LR
    A["CPEX pre/post hook execution records"] --> B["ControlTelemetryAggregator"]
    C["Trusted CF tool/agent/route context"] --> B
    B --> D["Canonical cpex.control.* attributes"]
    D --> E["Central SpanAttributePolicy"]
    E --> F["Internal observability DB"]
    E --> G["OpenTelemetry SDK / OTLP"]
```

#### 1. CPEX capability adapter

Add an adapter that consumes the versioned CPEX execution-record surface from #130.

Requirements:

- Feature/capability detection rather than assuming every installed CPEX version supports execution records.
- Clear minimum supported CPEX version once published.
- Safe no-op or documented legacy behavior when records are unavailable.
- No exception-message parsing fallback.
- Preserve current policy outcome and plugin error behavior.

#### 2. Per-tool aggregation

Accumulate CPEX execution records across `tool_pre_invoke` and `tool_post_invoke` within one `invoke_tool` call.

Aggregation must use invocation-local state, not process-global mutable state. Concurrent requests, nested tool calls, retries, and cancellation must not leak or mix records.

Suggested aggregate semantics:

```text
invocation_count = controls actually invoked
matched_count = records where CPEX reports matched=true
applied_count = records where CPEX reports applied=true
results_count = exported result records after bounds are applied
duration = sum of CPEX control execution duration
result.allowed = final effective control-chain decision
```

ContextForge must not invent `matched` or `applied` from plugin names or free-form metadata. Semantics come from the CPEX contract.

#### 3. Canonical OpenTelemetry representation

Default names should be vendor-neutral and documented, for example:

```text
cpex.control.type
cpex.control.enforcement_point
cpex.control.agent.id
cpex.control.applied_count
cpex.control.artifact.name
cpex.control.artifact.id
cpex.control.binding.name
cpex.control.config.keys
cpex.control.name
cpex.control.invocation_count
cpex.control.tool.name
cpex.control.duration
cpex.control.result.allowed
cpex.control.result.reason
cpex.control.result.error_code
cpex.control.status
cpex.control.results_count
```

Exact names and types should be finalized before implementation. OpenTelemetry naming and type guidance must be followed. Units should be documented rather than encoded inconsistently in attribute names.

Preferred scalable representation:

- aggregate control attributes on `tool.invoke`
- fixed-schema child span or event per control using stable keys

If an existing consumer requires flattened dynamic keys such as:

```text
cpex.control.results.<control>.artifact.name
cpex.control.results.<control>.result.allowed
```

support must be explicit, bounded, sanitized, configurable, and covered by collision tests. It must not replace the fixed-schema representation as the internal source of truth.

#### 4. Central attribute policy

Extend the capability currently represented by `SpanAttributeCustomizer`:

```yaml
attribute_mapping:
  "cpex.control.result.allowed": "controls.result.allow"
  "cpex.control.results.*.artifact.name": "controls.results.*.artifact.name"

remove_attributes:
  - "cpex.control.config.keys"
  - "cpex.control.results.*.result.reason"
```

The plugin may remain the configuration facade, but transformation must run at a central telemetry boundary so it applies consistently to:

- attributes supplied at span start
- attributes added while a span is active
- attributes supplied at span end
- internal observability DB spans
- native OpenTelemetry spans
- CPEX-derived spans

Current exact mapping/removal behavior must remain backward-compatible.

Wildcard rules should be segment-aware:

- `*` matches exactly one dot-delimited segment
- exact rules take precedence over wildcard rules
- wildcard captures are preserved in destination templates
- arbitrary regular expressions are not accepted
- invalid/ambiguous rules fail during configuration validation
- rules are compiled once at startup

Span-name mapping, if ever needed, must be configured separately from attribute-name mapping.

#### 5. Export behavior

- Internal DB and OTLP views must have equivalent control semantics.
- Telemetry must share trace context with `tool.invoke`.
- Sampling must follow existing OTel policy.
- Telemetry failures remain best-effort and never change enforcement result.
- SaaS and on-prem deployments use the same application behavior; only exporter configuration differs.

---

#### Trusted source boundaries

- Control identity, status, duration, hook, mode, and effective decision must come from trusted CPEX execution records.
- Client payloads and plugin metadata must not override framework-owned fields.
- Agent, tool, server, route, and binding identifiers must come from authenticated or server-side ContextForge state.
- Do not trust client-provided ownership/binding fields.

#### Sensitive-data controls

- Never emit control configuration values; configuration key names only.
- Do not emit tool arguments, results, headers, credentials, tokens, violation details, or stack traces through this feature.
- Sanitize and bound `result.reason`.
- Prefer stable error codes over exception text.
- Apply existing trace redaction before export and before internal DB persistence.
- Removal rules must run before attributes leave process boundaries.

#### Cardinality and amplification

- Bound controls accepted per invocation.
- Bound flattened result attributes per invocation.
- Bound string lengths, list sizes, key lengths, and total serialized attribute size.
- Validate and sanitize dynamic control-name path segments.
- Detect collisions after normalization and mapping.
- Emit a bounded truncation indicator/counter without leaking dropped values.
- Avoid creating one database transaction or exporter call per individual attribute.

#### Configuration safety

- Limit mapping/removal rule counts.
- Reject empty, malformed, oversized, recursive, or ambiguous mappings.
- Prevent mappings into reserved `otel.*` names.
- Avoid user-provided regular expressions and catastrophic backtracking.
- Define deterministic rename/drop order.
- Do not log sensitive attribute values on configuration or runtime errors.

#### Concurrency and lifecycle

- No cross-request leakage through shared plugin/global context.
- Cancellation and exceptions must clear invocation-local accumulators.
- Retries must have documented count behavior and must not duplicate records accidentally.
- Background/fire-and-forget controls must not mutate already-exported spans unsafely.
- Partial pre-hook records must still be exportable when invocation is denied before tool execution.

---

### Reliability and Performance Requirements

- Consume records in O(number of control records + attributes).
- Compile mapping rules once, not per invocation.
- Avoid payload serialization and deep copies.
- Batch attribute transformation and span updates.
- Respect OpenTelemetry attribute count/value limits.
- Define behavior for missing optional CPEX fields.
- A malformed individual record should be dropped safely without losing valid sibling records.
- Record drops should be observable through bounded internal counters/logging.
- Benchmark no-control, single-control, and long-chain paths.
- Document expected overhead and default limits.

---

### Backward Compatibility

- Existing spans remain unchanged when feature is disabled or CPEX records are unavailable.
- Existing exact `attribute_mapping` and `remove_attributes` configuration continues to work.
- Existing `plugin.metrics.*` behavior remains independent.
- Existing CPEX violation and `on_error` handling remains unchanged.
- No database migration should be required unless new structured persistence is explicitly selected.
- Configuration defaults should avoid unexpectedly increasing telemetry volume.

Recommended feature controls:

```text
CPEX_CONTROL_TELEMETRY_ENABLED
CPEX_CONTROL_TELEMETRY_FLATTEN_RESULTS
CPEX_CONTROL_TELEMETRY_MAX_RESULTS
CPEX_CONTROL_TELEMETRY_MAX_ATTRIBUTES
```

Names are illustrative and should follow existing ContextForge configuration conventions.

---

### Validation Coverage

Tests should cover:

- no controls
- one allow result
- multiple allow results
- pre-invocation deny
- post-invocation deny after successful tool execution
- payload/extension modification
- audit/transform suppressed denial
- non-fatal error
- fatal error
- timeout
- cancellation
- disabled/skipped control
- retries
- concurrent short-circuit
- missing/malformed CPEX fields
- exact rename/drop
- wildcard rename/drop
- wildcard collision and invalid patterns
- redaction of secrets/PII
- cardinality truncation
- internal DB and OTLP parity
- disabled feature and legacy CPEX compatibility
- concurrent request isolation

An end-to-end OTLP collector test must verify final exported attributes rather than only mocked method calls.

---

### 🔗 MCP Standards Check

- [x] Feature affects observability around MCP tool execution, not MCP wire semantics

---

### 🔄 Alternatives Considered

1. **Parse CPEX exceptions:** rejected; incomplete, unstable, and may leak sensitive text.
2. **Read arbitrary plugin metadata:** rejected as source of trusted enforcement facts.
3. **Implement adapters per plugin:** rejected due maintenance and performance cost.
4. **Hard-code downstream `controls.*` attributes in core:** rejected because ContextForge is vendor-neutral.
5. **Configure transformations only in external OTel Collector:** useful deployment option, but insufficient for internal observability parity and deployments where ContextForge owns projection.
6. **Use only dynamic flattened attributes:** rejected as canonical representation due cardinality and schema growth; may be supported as bounded compatibility projection.

---

### Definition of Done

- ContextForge consumes versioned CPEX execution records from #130.
- Pre/post control results correlate to correct `tool.invoke` trace.
- Canonical vendor-neutral control telemetry is documented.
- Aggregate and per-control semantics are stable and typed.
- Attribute mapping/removal supports safe dynamic wildcard rules.
- Transformation applies consistently to internal DB and OTLP telemetry.
- Sensitive values and untrusted identifiers are sanitized and bounded.
- Concurrency, retries, cancellation, and partial-denial paths are correct.
- Feature remains backward-compatible and disabled/no-op behavior is documented.
- End-to-end tests prove exported telemetry can support downstream enforcement dashboards without parsing logs or exception messages.

---

### 📓 Additional Context

- Depends on: [contextforge-org/cpex#130](https://github.com/contextforge-org/cpex/issues/130)
- Related CPEX telemetry discussion: [contextforge-org/cpex#43](https://github.com/contextforge-org/cpex/issues/43)
- Existing ContextForge components:
  - `mcpgateway/services/tool_service.py`
  - `mcpgateway/plugins/observability_adapter.py`
  - `mcpgateway/plugins/utils.py`
  - `mcpgateway/observability.py`
  - `mcpgateway/services/observability_service.py`
  - `plugins/span_attribute_customizer/`

prakhar-singh1928:
# Pull Request

## 🔗 Related Issue

Closes #5785

---

## 📝 Summary

Adds structured per-plugin enforcement observability on every tool invocation by consuming `ControlExecutionRecord` objects emitted by the CPEX framework (≥ 0.1.2).

On each traced `invoke_tool()` call the gateway now writes two new span types into `observability_spans` (and, when active, the OTel SDK sink):

- **`cpex.control.summary`** — one per tool call. Carries aggregate counters (`invocation_count`, `matched_count`, `applied_count`, `error_count`, `timeout_count`, `duration`), the overall `result.allowed` outcome, the `enforcement_point` tag, and calling identity (`tool.name`, `agent.id`, `binding.name`).
- **`cpex.control.result`** — one per plugin evaluated, nested under the summary span. Carries the per-plugin `name`, `hook_name`, `mode`, `status`, `result.allowed`, `duration`, and optional `result.reason`, `result.error_code`, and `config.keys` (key names only, never values).

All fields come from **trusted CPEX framework sources** (`ControlExecutionRecord`), never from untrusted plugin metadata. The feature is a complete no-op on CPEX < 0.1.2.

**Files changed:**

| File | Change |
|---|---|
| `mcpgateway/plugins/control_telemetry.py` | New — `ControlTelemetryAccumulator` + `record_control_telemetry()` + DB/OTel sinks |
| `mcpgateway/plugins/cpex_compat.py` | New — CPEX version feature-detection helpers |
| `mcpgateway/plugins/utils.py` | Wildcard-aware attribute policy: `compile_attribute_policy()`, `apply_attribute_mapping()` |
| `mcpgateway/services/tool_service.py` | Accumulator wired at all 5 pre/post hook sites and all 4 exit paths |
| `mcpgateway/config.py` | 5 new `CPEX_CONTROL_TELEMETRY_*` settings |
| `pyproject.toml` / `uv.lock` | `cpex>=0.1.2`; `exclude-newer-package` cutoff advanced |
| `.env.example` | New `CPEX_CONTROL_TELEMETRY_*` section with descriptions |
| `CHANGELOG.md` | Entry under `[Unreleased] → Added` |
| `tests/unit/…/test_cpex_compat.py` | 11 unit tests |
| `tests/unit/…/test_control_telemetry.py` | 47 unit tests |
| `tests/unit/…/test_record_control_telemetry.py` | 24 integration tests |
| `tests/e2e/test_otel_control_telemetry_e2e.py` | 7 E2E tests (real gateway + PIIFilter + in-memory SQLite) |

---

## 📏 Reviewability

---

## 🏷️ Type of Change

- [ ] Bug fix
- [x] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

| Check | Command | Status |
|---|---|---|
| Lint suite | `make ruff` | ✅ All clear |
| Unit tests (new files only) | `python -m pytest tests/unit/mcpgateway/plugins/test_cpex_compat.py tests/unit/mcpgateway/plugins/test_control_telemetry.py tests/unit/mcpgateway/plugins/test_record_control_telemetry.py -v` | ✅ 82/82 passed |
| E2E — DB spans present in real traced request | `OBSERVABILITY_ENABLED=true PLUGINS_ENABLED=true python -m pytest tests/e2e/test_otel_control_telemetry_e2e.py -v` | ✅ 5 passed, 2 skipped¹ |
| Full plugin unit suite | `python -m pytest tests/unit/mcpgateway/plugins/ -v` | ✅ 720 passed, 33 skipped |
| Observability service tests | `python -m pytest tests/unit/mcpgateway/services/test_observability_service.py -v` | ✅ 92/92 passed |

¹ The 2 OTel-SDK tests skip cleanly when `opentelemetry-sdk` (the `observability` extra) is absent — this is expected and documented in the test fixture.

### Manual validation

Ran an in-memory SQLite gateway with PIIFilter active (via `test_otel_control_telemetry_e2e.py::test_control_summary_span_present_in_db`) and verified via DB query that a traced `tools/call` request writes:
- One `cpex.control.summary` span with `cpex.control.invocation_count`, `cpex.control.result.allowed`, `cpex.control.tool.name` attributes
- One `cpex.control.result` span per plugin evaluated, with `cpex.control.name`, `cpex.control.status`, `cpex.control.result.allowed`, `cpex.control.enforcement_point`, and `parent_span_id` linking it to the summary span

---

## ✅ Checklist

---

### Known items (not blockers, documented for reviewer awareness)

1. **Phase 5 attribute policy not yet wired** — `compile_attribute_policy()` / `apply_attribute_mapping()` infrastructure is built and tested (see `mcpgateway/plugins/utils.py` lines 481–669 and `test_record_control_telemetry.py::TestWildcardAttributeMapping`), but `record_control_telemetry()` does not yet read `SpanAttributeCustomizer` plugin context state (`span_attribute_mapping` / `remove_span_attributes`) before emitting, so the wildcard rename/drop rules are ready but not consumed. Planned follow-on for Phase 5 wiring (same pattern as `start_span()` at `observability_service.py:520`).

2. **`cpex_control_telemetry_max_attributes` is informational only** — Not enforced gateway-side. Intended as a hint for external OTel-collector attribute-limit configuration (e.g. `transform/attributes` processor). The `config.py` description explicitly calls this out.

3. **OTel-SDK E2E tests skip in standard dev venv** — The 2 tests requiring `opentelemetry-sdk` (`observability` optional extra) skip cleanly when the package is absent. This is expected; the DB-side E2E tests (which are the primary verification gate) pass unconditionally.

4. **`_EXECUTION_RECORDS_SUPPORTED` cache is idempotent but has no mutex** — `cpex_compat.py:45` writes to the module-global cache without a `threading.Lock`. The write is idempotent (both concurrent writers set it to the same value) so there is no correctness risk — just a theoretical double-import race on the very first call from two threads. Low-priority follow-on.

### Architecture decisions

- **Best-effort telemetry (L4)**: `record_control_telemetry()` never raises into the request path; all exceptions are caught and logged at debug level. This mirrors `record_plugin_metrics()` in `utils.py`.
- **Separate DB sessions**: `_emit_db_spans()` opens its own `SessionLocal()` (issue #3883 pattern) so observability writes persist even when the main request fails.
- **Explicit per-control child spans rather than inlined plugin metadata**: Control-execution records are **trusted** CPEX framework output (sourced from framework-internal `PluginRef` config), not **untrusted** plugin-returned metadata. Recording them as first-class spans with a fixed schema makes queries/dashboards simpler and avoids the deny-by-default field-name allowlist required by plugin metadata (see `utils.py` `_SAFE_STRING_FIELD_NAMES` / `_SAFE_NUMERIC_FIELD_NAMES`).
- **Flattened attributes off by default**: `CPEX_CONTROL_TELEMETRY_FLATTEN_RESULTS=false` (default). Enabled only when downstream tooling requires dynamic key names; the fixed-schema `cpex.control.result` child spans remain the source of truth.

### Commits

- `370db9e11` feat: consume Cpex control execution records

### DIFF
--- a/mcpgateway/plugins/control_telemetry.py
+++ b/mcpgateway/plugins/control_telemetry.py
@@ -89,6 +89,7 @@ class ControlTelemetryAccumulator:
     _pre_denied: bool = False
     _post_denied: bool = False
     _truncated: int = 0  # records dropped due to per-call cap OR per-hook cap
+    _export_cap_dropped: int = 0  # records dropped at emit time by the per-invocation export cap
 
     def add(self, result: Any, *, hook: str) -> None:
         """Consume executions from one ``invoke_hook`` result.
@@ -172,12 +173,39 @@ class ControlTelemetryAccumulator:
 
     @property
     def truncated(self) -> int:
-        """Number of records dropped due to the per-hook or per-call cap.
+        """Total records dropped across all three truncation tiers:
+
+        - Tier 1: per-hook cap (``_MAX_RECORDS_PER_HOOK``)
+        - Tier 2: per-call accumulation cap (``_MAX_RECORDS_PER_CALL``)
+        - Tier 3: per-invocation export cap (``CPEX_CONTROL_TELEMETRY_MAX_RESULTS``)
+
+        Tier-3 drops are recorded by ``mark_export_cap_dropped()`` after the
+        export cap is applied in ``record_control_telemetry()``.
 
         Returns:
-            Count of dropped records (0 when no overflow).
+            Total count of dropped records (0 when no overflow on any tier).
         """
-        return self._truncated
+        return self._truncated + self._export_cap_dropped
+
+    def mark_export_cap_dropped(self, count: int) -> None:
+        """Record the number of records silently dropped by the export cap.
+
+        Called once per ``record_control_telemetry()`` invocation, after
+        ``_get_max_results()`` is applied, so the ``cpex.control.truncated``
+        summary attribute reflects all three truncation tiers.
+
+        Args:
+            count: Number of records that exceeded ``CPEX_CONTROL_TELEMETRY_MAX_RESULTS``
+                   and were not exported to any sink.
+
+        Examples:
+            >>> acc = ControlTelemetryAccumulator()
+            >>> acc.mark_export_cap_dropped(5)
+            >>> acc.truncated
+            5
+        """
+        if count > 0:
+            self._export_cap_dropped += count
 
     @property
     def effective_allowed(self) -> bool:
@@ -306,6 +334,16 @@ def record_control_telemetry(
                 "cpex.control.enforcement_point": _enforcement_point(accumulator),
             }
         )
+        # Compute and record export-cap (tier-3) drops before building the summary span.
+        # records_received is the count before any export cap; results_count is the cap-bounded
+        # export count.  The difference is exactly the number of records the islice will skip.
+        records_received = aggregate.get("cpex.control.records_received", 0)
+        max_results = _get_max_results()
+        export_cap_dropped = max(0, records_received - max_results)
+        if export_cap_dropped:
+            accumulator.mark_export_cap_dropped(export_cap_dropped)
+
+        # cpex.control.truncated now covers all three tiers (hook cap + call cap + export cap).
         if accumulator.truncated:
             aggregate["cpex.control.truncated"] = accumulator.truncated
 
--- a/mcpgateway/plugins/control_telemetry.py
+++ b/mcpgateway/plugins/control_telemetry.py
@@ -7,8 +7,10 @@ Accumulator and emitter for CPEX control-execution telemetry.
 
 Mirrors the structure of ``record_plugin_metrics()`` in ``plugins/utils.py``
 but consumes **trusted** ``ControlExecutionRecord`` fields rather than
-untrusted plugin metadata.  All bounds and sanitization follow the same S4
-principles.
+untrusted plugin metadata.  All bounds and sanitization follow the same
+principles of bounded cardinality and field-level sanitization applied
+throughout the plugin telemetry pipeline (see ``plugins/utils.py`` for the
+shared ``_safe_str``/``_safe_num`` helpers and field-name allowlists).
 
 Architecture
 ------------
@@ -227,9 +229,10 @@ class ControlTelemetryAccumulator:
     def mark_export_cap_dropped(self, count: int) -> None:
         """Record the number of records silently dropped by the export cap.
 
-        Called once per ``record_control_telemetry()`` invocation, after
-        ``_get_max_results()`` is applied, so the ``cpex.control.truncated``
-        summary attribute reflects all three truncation tiers.
+        Idempotent: the first call with ``count > 0`` sets the value; subsequent
+        calls on the same accumulator are no-ops.  This prevents ``cpex.control.truncated``
+        from being inflated when ``record_control_telemetry()`` is called more than once
+        on the same accumulator (e.g. on a retry path or in tests).
 
         Args:
             count: Number of records that exceeded ``CPEX_CONTROL_TELEMETRY_MAX_RESULTS``
@@ -240,9 +243,12 @@ class ControlTelemetryAccumulator:
             >>> acc.mark_export_cap_dropped(5)
             >>> acc.truncated
             5
+            >>> acc.mark_export_cap_dropped(5)  # second call is a no-op
+            >>> acc.truncated
+            5
         """
-        if count > 0:
-            self._export_cap_dropped += count
+        if count > 0 and self._export_cap_dropped == 0:
+            self._export_cap_dropped = count
 
     @property
     def effective_allowed(self) -> bool:
--- a/mcpgateway/plugins/control_telemetry.py
+++ b/mcpgateway/plugins/control_telemetry.py
@@ -46,9 +46,9 @@ from mcpgateway.plugins.utils import _IDENTIFIER_RE  # re-use the same identifie
 logger = logging.getLogger(__name__)
 
 # ── Bounds (mirrors _MAX_PLUGIN_KEYS / _MAX_PLUGINS_PER_CALL in utils.py) ──────
-_MAX_RECORDS_PER_HOOK = 64   # max ControlExecutionRecords consumed per invoke_hook call
+_MAX_RECORDS_PER_HOOK = 64  # max ControlExecutionRecords consumed per invoke_hook call
 _MAX_RECORDS_PER_CALL = 128  # cap across pre + post combined
-_MAX_REASON_LEN = 256        # CPEX already bounds this; enforce again defensively
+_MAX_REASON_LEN = 256  # CPEX already bounds this; enforce again defensively
 _MAX_ERROR_CODE_LEN = 256
 _MAX_CONFIG_KEYS = 64
 
@@ -197,7 +197,7 @@ class ControlTelemetryAccumulator:
             "cpex.control.matched_count": matched_count,
             "cpex.control.applied_count": applied_count,
             "cpex.control.results_count": len(self._records),
-            "cpex.control.duration": duration_ns,   # nanoseconds
+            "cpex.control.duration": duration_ns,  # nanoseconds
             "cpex.control.result.allowed": self.effective_allowed,
             "cpex.control.error_count": error_count,
             "cpex.control.timeout_count": timeout_count,
--- a/mcpgateway/plugins/control_telemetry.py
+++ b/mcpgateway/plugins/control_telemetry.py
@@ -88,7 +88,7 @@ class ControlTelemetryAccumulator:
     _records: list = field(default_factory=list)
     _pre_denied: bool = False
     _post_denied: bool = False
-    _truncated: int = 0  # records dropped due to per-call cap
+    _truncated: int = 0  # records dropped due to per-call cap OR per-hook cap
 
     def add(self, result: Any, *, hook: str) -> None:
         """Consume executions from one ``invoke_hook`` result.
@@ -101,7 +101,13 @@ class ControlTelemetryAccumulator:
             return
 
         records = get_executions(result)
-        for rec in itertools.islice(records, _MAX_RECORDS_PER_HOOK):
+        hook_count = 0
+        for rec in records:
+            if hook_count >= _MAX_RECORDS_PER_HOOK:
+                # Per-hook cap exceeded — count these as truncated and stop.
+                self._truncated += 1
+                continue
+            hook_count += 1
             if len(self._records) >= _MAX_RECORDS_PER_CALL:
                 self._truncated += 1
                 continue
@@ -113,6 +119,29 @@ class ControlTelemetryAccumulator:
                 self._pre_denied = True
             else:
                 self._post_denied = True
+
+    def mark_denied(self, *, hook: str) -> None:
+        """Explicitly mark a denial when ``violations_as_exceptions=True`` causes
+        ``invoke_hook()`` to raise before returning, so ``add()`` is never called.
+
+        Call this from the ``except PluginViolationError`` handler immediately
+        after catching the exception and before re-raising.
+
+        Args:
+            hook: ``"pre"`` or ``"post"`` — which enforcement point was denied.
+
+        Examples:
+            >>> acc = ControlTelemetryAccumulator()
+            >>> acc.mark_denied(hook="pre")
+            >>> acc.pre_denied
+            True
+            >>> acc.effective_allowed
+            False
+        """
+        if hook == "pre":
+            self._pre_denied = True
+        else:
+            self._post_denied = True
 
     @property
     def records(self) -> list:
@@ -143,7 +172,7 @@ class ControlTelemetryAccumulator:
 
     @property
     def truncated(self) -> int:
-        """Number of records dropped due to the per-call cap.
+        """Number of records dropped due to the per-hook or per-call cap.
 
         Returns:
             Count of dropped records (0 when no overflow).
@@ -165,6 +194,18 @@ class ControlTelemetryAccumulator:
         Returns a ``dict`` of ``cpex.control.*`` attributes safe for span
         attachment.  Only uses trusted ``ControlExecutionRecord`` fields —
         never plugin metadata.
+
+        Count semantics:
+
+        - ``invocation_count``: controls that *actually ran* — status in
+          ``{completed, error, timeout}``.  Disabled, skipped, and cancelled
+          records are excluded (matches CPEX semantics).
+        - ``records_received``: total records accumulated before any export
+          cap is applied.  Includes disabled/skipped/cancelled.
+        - ``results_count``: records that will actually be exported to sinks,
+          i.e. ``min(records_received, CPEX_CONTROL_TELEMETRY_MAX_RESULTS)``.
+          Downstream operators can derive ``records_dropped = records_received
+          - results_count + truncated`` to understand total loss.
 
         Returns:
             Dictionary of aggregate control telemetry attributes.
@@ -196,13 +237,15 @@ class ControlTelemetryAccumulator:
             except Exception:  # noqa: BLE001
                 logger.debug("Failed to aggregate one ControlExecutionRecord", exc_info=True)
 
+        records_received = len(self._records)
         # results_count = records exported after the per-invocation cap (not raw accumulated count).
-        results_count = min(len(self._records), _get_max_results())
+        results_count = min(records_received, _get_max_results())
 
         return {
             "cpex.control.invocation_count": invocation_count,
             "cpex.control.matched_count": matched_count,
             "cpex.control.applied_count": applied_count,
+            "cpex.control.records_received": records_received,
             "cpex.control.results_count": results_count,
             "cpex.control.duration_ns": total_duration_ns,  # nanoseconds — OTel unit suffix convention
             "cpex.control.result.allowed": self.effective_allowed,
@@ -432,11 +475,23 @@ def _per_control_attributes(hook: str, rec: Any) -> dict:
         # requested_allow: only emit when present (None means not applicable to this mode)
         if rec.requested_allow is not None:
             attrs["cpex.control.result.requested_allowed"] = bool(rec.requested_allow)
-        # Optional free-text fields — only emit when present
-        if rec.reason:
-            attrs["cpex.control.result.reason"] = _safe_str(rec.reason, _MAX_REASON_LEN)
-        if rec.error_code:
-            attrs["cpex.control.result.error_code"] = _safe_str(rec.error_code, _MAX_ERROR_CODE_LEN)
+        # Artifact identity — only emit when CPEX record carries them (optional fields)
+        artifact_name = getattr(rec, "artifact_name", None)
+        if artifact_name:
+            attrs["cpex.control.artifact.name"] = _safe_str(artifact_name, 128)
+        artifact_id = getattr(rec, "artifact_id", None)
+        if artifact_id:
+            attrs["cpex.control.artifact.id"] = _safe_str(artifact_id, 128)
+        # Optional free-text fields — gated by CPEX_CONTROL_TELEMETRY_EMIT_REASON (default: false).
+        # reason and error_code can contain PII, tool argument values, or exception content.
+        # They are opt-in until Phase 5 central attribute-policy wiring provides a redaction
+        # boundary.  Set CPEX_CONTROL_TELEMETRY_EMIT_REASON=true only in environments where
+        # these fields are known safe and the observability sink is appropriately secured.
+        if _emit_reason_enabled():
+            if rec.reason:
+                attrs["cpex.control.result.reason"] = _safe_str(rec.reason, _MAX_REASON_LEN)
+            if rec.error_code:
+                attrs["cpex.control.result.error_code"] = _safe_str(rec.error_code, _MAX_ERROR_CODE_LEN)
         if rec.config_keys:
             # Key names only (CPEX never includes values); bounded list join
             attrs["cpex.control.config.keys"] = ",".join(rec.config_keys[:_MAX_CONFIG_KEYS])
@@ -502,6 +557,29 @@ def _get_max_results() -> int:
         return 32
 
 
+def _emit_reason_enabled() -> bool:
+    """Return True when ``result.reason`` and ``result.error_code`` emission is enabled.
+
+    These fields are opt-in (default: False) because they may contain PII, tool argument
+    values, or exception content that should not leave the process without passing through
+    a redaction boundary.  Enable via ``CPEX_CONTROL_TELEMETRY_EMIT_REASON=true`` only in
+    environments where the observability sink is appropriately secured.
+
+    Returns:
+        True when ``cpex_control_telemetry_emit_reason`` is set to True in settings.
+
+    Examples:
+        >>> isinstance(_emit_reason_enabled(), bool)
+        True
+    """
+    try:
+        from mcpgateway.config import settings  # pylint: disable=import-outside-toplevel
+
+        return bool(getattr(settings, "cpex_control_telemetry_emit_reason", False))
+    except Exception:  # noqa: BLE001
+        return False
+
+
 # ---------------------------------------------------------------------------
 # Flattened-results projection helper
 # ---------------------------------------------------------------------------
@@ -556,18 +634,33 @@ def _build_flattened_attributes(
                 seen_names.add(raw_name)
 
                 prefix = f"cpex.control.results.{raw_name}"
+                result[f"{prefix}.name"] = _safe_str(raw_name, 64)
                 result[f"{prefix}.status"] = _safe_str(str(getattr(rec, "status", "")), 32)
                 result[f"{prefix}.enforcement_point"] = hook
                 result[f"{prefix}.result.allowed"] = bool(getattr(rec, "effective_allow", True))
-                result[f"{prefix}.duration"] = int(getattr(rec, "duration_ns", 0))
+                result[f"{prefix}.duration_ns"] = int(getattr(rec, "duration_ns", 0))  # nanoseconds — OTel unit suffix convention
 
-                reason = getattr(rec, "reason", None)
-                if reason:
-                    result[f"{prefix}.result.reason"] = _safe_str(reason, _MAX_REASON_LEN)
+                artifact_name = getattr(rec, "artifact_name", None)
+                if artifact_name:
+                    result[f"{prefix}.artifact.name"] = _safe_str(artifact_name, 128)
 
-                error_code = getattr(rec, "error_code", None)
-                if error_code:
-                    result[f"{prefix}.result.error_code"] = _safe_str(error_code, _MAX_ERROR_CODE_LEN)
+                artifact_id = getattr(rec, "artifact_id", None)
+                if artifact_id:
+                    result[f"{prefix}.artifact.id"] = _safe_str(artifact_id, 128)
+
+                config_keys = getattr(rec, "config_keys", None)
+                if config_keys:
+                    result[f"{prefix}.config.keys"] = ",".join(config_keys[:_MAX_CONFIG_KEYS])
+
+                # reason/error_code gated by CPEX_CONTROL_TELEMETRY_EMIT_REASON (default: false)
+                if _emit_reason_enabled():
+                    reason = getattr(rec, "reason", None)
+                    if reason:
+                        result[f"{prefix}.result.reason"] = _safe_str(reason, _MAX_REASON_LEN)
+
+                    error_code = getattr(rec, "error_code", None)
+                    if error_code:
+                        result[f"{prefix}.result.error_code"] = _safe_str(error_code, _MAX_ERROR_CODE_LEN)
             except Exception:  # noqa: BLE001
                 logger.debug("Failed to flatten one ControlExecutionRecord", exc_info=True)
 
--- a/mcpgateway/plugins/control_telemetry.py
+++ b/mcpgateway/plugins/control_telemetry.py
@@ -53,11 +53,20 @@ _MAX_RECORDS_PER_CALL = 128  # cap across pre + post combined
 _MAX_REASON_LEN = 256  # CPEX already bounds this; enforce again defensively
 _MAX_ERROR_CODE_LEN = 256
 _MAX_CONFIG_KEYS = 64
-_MAX_CONFIG_KEY_LEN = 128  # per-key length cap before joining
+_MAX_CONFIG_KEY_LEN = 128  # per-key byte cap before joining
 _MAX_CONFIG_KEYS_JOINED_LEN = 4096  # max total byte length of the joined config_keys string
 
 # Statuses that represent controls that actually ran (exclude disabled/skipped/cancelled).
 _ACTIVE_STATUSES = frozenset({"completed", "error", "timeout"})
+
+# Config-key sanitization: allow printable identifier-like characters only.
+# Commas are the joining delimiter so must be excluded; control characters
+# (CR, LF, NUL, tab, …) and non-ASCII are rejected to prevent log injection and
+# telemetry-ambiguity attacks through key names.  The pattern deliberately does
+# NOT use _IDENTIFIER_RE (which allows dots and hyphens) because config-key names
+# are opaque tokens whose cardinality and format are determined by CPEX control
+# authors, not the gateway; the charset is the safest defensible subset.
+_CONFIG_KEY_RE = _IDENTIFIER_RE  # re-use: ^[A-Za-z0-9_.-]{1,64}$
 
 
 # ---------------------------------------------------------------------------
@@ -96,6 +105,7 @@ class ControlTelemetryAccumulator:
     _pre_denied: bool = False
     _post_denied: bool = False
     _plugin_errored: bool = False  # True when a PluginError (outage) was caught on any hook
+    _plugin_error_hook: str = ""   # "pre" | "post" | "" — which hook the error occurred on
     _truncated: int = 0  # records dropped due to per-call cap OR per-hook cap
     _export_cap_dropped: int = 0  # records dropped at emit time by the per-invocation export cap
 
@@ -152,27 +162,42 @@ class ControlTelemetryAccumulator:
         else:
             self._post_denied = True
 
-    def mark_plugin_error(self) -> None:
+    def mark_plugin_error(self, *, hook: str = "") -> None:
         """Explicitly record that a ``PluginError`` (outage/crash/misconfiguration) occurred.
 
         Call this from the ``except PluginError`` handler before re-raising so that
         the emitted summary span carries ``cpex.control.plugin_error=True``.  This
         distinguishes a plugin infrastructure failure from both a successful allow
-        decision (``result.allowed=True``, no error) and a deliberate policy denial
-        (``result.allowed=False`` set via ``mark_denied()``).
+        decision and a deliberate policy denial (``result.allowed=False`` via
+        ``mark_denied()``).
 
-        When the failing plugin is the first in the chain the accumulator may have no
+        When the failing plugin is first in the chain the accumulator may have no
         records and no denial flags — without this flag ``record_control_telemetry()``
         would silently skip emission.  Including ``_plugin_errored`` in the empty-guard
         ensures the summary span is always emitted so the outage is visible in dashboards.
 
+        Because a ``PluginError`` makes the control-chain decision *indeterminate*
+        (neither allowed nor denied), ``aggregate()`` omits ``cpex.control.result.allowed``
+        when this flag is set and no explicit denial flag is present.  Downstream dashboards
+        must treat the absence of ``result.allowed`` as "decision unknown" rather than
+        as an allow.
+
+        Args:
+            hook: ``"pre"`` or ``"post"`` — which enforcement point the error occurred on.
+                  Used to derive ``enforcement_point`` when no records exist.  Pass ``""``
+                  when the hook is unknown or not applicable.
+
         Examples:
             >>> acc = ControlTelemetryAccumulator()
-            >>> acc.mark_plugin_error()
+            >>> acc.mark_plugin_error(hook="pre")
             >>> acc.plugin_errored
             True
+            >>> acc.plugin_error_hook
+            'pre'
         """
         self._plugin_errored = True
+        if hook in ("pre", "post"):
+            self._plugin_error_hook = hook
 
     @property
     def plugin_errored(self) -> bool:
@@ -182,6 +207,15 @@ class ControlTelemetryAccumulator:
             True if any plugin raised ``PluginError``; False otherwise.
         """
         return self._plugin_errored
+
+    @property
+    def plugin_error_hook(self) -> str:
+        """The hook on which the ``PluginError`` occurred (``"pre"``, ``"post"``, or ``""``).
+
+        Returns:
+            Hook name string, or empty string when not set.
+        """
+        return self._plugin_error_hook
 
     @property
     def records(self) -> list:
@@ -320,10 +354,19 @@ class ControlTelemetryAccumulator:
             "cpex.control.records_received": records_received,
             "cpex.control.results_count": results_count,
             "cpex.control.duration_ns": total_duration_ns,  # nanoseconds — OTel unit suffix convention
-            "cpex.control.result.allowed": self.effective_allowed,
             "cpex.control.error_count": error_count,
             "cpex.control.timeout_count": timeout_count,
         }
+        # cpex.control.result.allowed is the overall enforcement decision.
+        # When a PluginError occurred and no explicit denial was also set, the decision
+        # is INDETERMINATE — the chain did not complete normally so we cannot assert
+        # "allowed".  Omitting the key signals "unknown" to dashboards rather than
+        # implying a clean allow.  If a denial flag was also set (unusual but possible),
+        # the denial takes precedence and we do emit result.allowed=False.
+        decision_indeterminate = self._plugin_errored and self.effective_allowed
+        if not decision_indeterminate:
+            result["cpex.control.result.allowed"] = self.effective_allowed
+
         # cpex.control.plugin_error distinguishes a plugin infrastructure failure
         # (PluginError: crash/timeout/misconfiguration) from both a successful allow
         # and a deliberate policy denial (result.allowed=False).  Only emitted when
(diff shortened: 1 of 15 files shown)


vishu-bh:
Thanks for putting this together. The PR establishes a strong foundation: invocation-local accumulation, trusted CPEX execution records, bounded per-control spans, DB/OTel sinks, legacy compatibility, and good initial test coverage.

I did a requirements-coverage pass against #5785 before starting a detailed code review. A few areas may need clarification or additional work:

### 1. Attribute rename/drop policy

The wildcard-aware policy helpers are implemented and tested, but `record_control_telemetry()` does not yet apply the `SpanAttributeCustomizer` mapping and removal state before writing DB or OTel spans.

Could you confirm whether a follow-on issue has already been created for this wiring? If yes, could you link it in the PR?

Since configurable exact/wildcard rename and removal are key acceptance criteria in #5785, should this PR continue to close the whole issue, or should #5785 remain open until the policy is applied consistently to both sinks?

### 2. Required attribute coverage

The fixed-schema result spans currently expose:

- Control name
- Hook name and enforcement point
- Mode and status
- Effective allow result
- Duration
- Optional reason and error code
- Configuration keys

However, CPEX also provides fields that are not currently projected, including `plugin_id`, `plugin_kind`, `requested_allow`, `matched`, and `applied`.

The downstream contract also expects artifact and binding information. Could you please check whether the fixed-schema result span should additionally expose fields such as:

```text
cpex.control.artifact.name
cpex.control.artifact.id
cpex.control.binding.name
cpex.control.result.requested_allowed
cpex.control.matched
cpex.control.applied
```

The flattened representation currently emits:

```text
status
enforcement_point
result.allowed
duration
result.reason
result.error_code
```

Could you also check whether it should include the requested dynamic fields:

```text
cpex.control.results.<control>.artifact.name
cpex.control.results.<control>.artifact.id
cpex.control.results.<control>.binding.name
cpex.control.results.<control>.config.keys
cpex.control.results.<control>.name
cpex.control.results.<control>.count
```

### 3. Duration unit

`cpex.control.duration` currently contains `duration_ns` directly.

One expected downstream mapping is:

```text
cpex.control.duration -> controls.duration.ms
```

A rename would change the key but not convert the value, leaving nanoseconds in an attribute named milliseconds.

Could you please confirm the canonical unit and check whether CF should convert the CPEX value to milliseconds before export? Whichever unit is selected should be identical for DB and OTel and documented as part of the contract.

### 4. Denial paths with `violations_as_exceptions=True`

Pre/post hook calls use:

```python
violations_as_exceptions=True
```

The accumulator is populated only after `invoke_hook()` returns:

```python
pre_result, _ = await plugin_manager.invoke_hook(...)
_ctl_acc.add(pre_result, hook="pre")
```

If CPEX raises `PluginViolationError` for a denial, the accumulator call may never execute. The exception handler would flush an empty accumulator, potentially losing the denying control’s execution record.

Could you verify this against the real CPEX 0.1.2 behavior? It would be useful to have an integration test where a real pre-invoke control denies and the emitted span identifies that control. The same question applies to a post-invoke denial.

### 5. Additional hook paths

I noticed two existing hook paths that may not feed the new accumulator:

- `prepare_rust_mcp_tool_execution()` invokes the pre-hook for the Rust-native path.
- `_run_timeout_post_invoke()` invokes the post-hook after a timeout.

Could you check whether their execution records are captured by another path? If not, these paths may need a way to return or add their records to the invocation-local accumulator.

### 6. Count semantics

`invocation_count` currently increments once for every accumulated record. CPEX distinguishes statuses such as `completed`, `error`, `timeout`, `cancelled`, `skipped`, and `disabled`.

Could you confirm whether the intended invocation count should follow CPEX semantics and count only controls that actually ran, for example:

```text
completed
error
timeout
```

Also, `results_count` currently uses the number of accumulated records, while only `CPEX_CONTROL_TELEMETRY_MAX_RESULTS` records are exported. The issue describes `results_count` as the number of exported results after bounds are applied. Could you check which meaning downstream consumers should rely on?

### 7. Truncation accounting and attribute cap

There appear to be a few different bounds:

- 64 records consumed per hook.
- 128 records accumulated per tool call.
- Configurable result export cap, defaulting to 32.
- Configurable attribute cap, currently informational only.

Records beyond the per-hook `islice()` and records beyond the export cap do not appear to contribute to the truncation indicator.

Could you check whether all dropped records should be represented by one bounded truncation counter? It may also be useful to distinguish:

```text
records_received
records_exported
records_dropped
```

Could you also confirm whether gateway-side enforcement of `CPEX_CONTROL_TELEMETRY_MAX_ATTRIBUTES` is planned? Depending only on external collector configuration would not bound internal DB attributes.

### 8. Mapping collisions and invalid fallback

`apply_attribute_mapping()` writes mapped values into a dictionary. If two source attributes map to the same destination, the later value can overwrite the earlier one.

Could you check whether collisions should be rejected during policy compilation or handled through an explicit deterministic rule with a collision indicator?

Also, when policy compilation fails, the runtime fallback appears able to apply exact mappings without all original validation—for example, a destination under `otel.*`. Would fail-closed behavior be safer for invalid mappings?

### 9. Reason sanitization

The defensive `_safe_str()` bound is good, but it truncates rather than redacts. A control reason could still contain PII, credentials, tool arguments, or exception content.

Could you check how existing trace redaction should be applied before:

- Internal DB persistence
- OTel export

Until central policy wiring is available, should `result.reason` be disabled by default or pass through an existing sanitization boundary?

### 10. Agent and binding identity

Current enrichment appears to use:

```python
agent_id = app_user_email or user_email
binding_name = gateway_name or server_id
```

Could you confirm that these match the intended business semantics?

In particular:

- A user email may not represent the invoking agent ID.
- A gateway name and server ID may not represent a control binding.
- Email values can introduce privacy and telemetry-cardinality concerns.

It would help to document the trusted source and lifecycle of each identity field before downstream dashboards depend on them.

### 11. End-to-end validation

The test suite is a good start. Could you check whether we should add end-to-end coverage for:

- Real pre-invoke denial
- Real post-invoke denial
- Timeout post-hook records
- Rust-native tool path
- Exact and wildcard rename/drop in final DB and OTel output
- Redaction before persistence/export
- Mapping collisions
- Disabled/skipped/cancelled records
- Result truncation and count semantics

The current in-memory OTel exporter tests are useful. #5785 also requests verification of final exported attributes through an OTLP collector; is that planned in this PR or a linked follow-up?

Overall, the architecture direction looks good. Clarifying these points would ensure the telemetry is complete, semantically stable, and safe for downstream enforcement dashboards and billing workflows.

prakhar-singh1928:
Thanks for the detailed review, Vishu — really useful to have it structured against the acceptance criteria. Here's where each point stands after the follow-up commits:

---

**1 — Attribute rename/drop policy (Phase 5 wiring)**

Confirmed gap, intentional deferral. `compile_attribute_policy()` / `apply_attribute_mapping()` infrastructure is fully built and tested in `utils.py`, but `record_control_telemetry()` does not yet read `SpanAttributeCustomizer` state before writing to DB or OTel sinks. Changing `Closes #5785` → `Relates to #5785` and will open a dedicated Phase 5 tracking issue for the wiring, compile-time collision detection, and `MAX_ATTRIBUTES` gateway-side enforcement before this merges.

---

**2 — Required attribute coverage**

Already addressed in commit `4a24d1c`. `_per_control_attributes()` now emits all of the fields you listed: `cpex.control.plugin_id`, `cpex.control.plugin_kind`, `cpex.control.matched`, `cpex.control.applied`, `cpex.control.payload_modified`, `cpex.control.result.requested_allowed` (conditional on `requested_allow != None`), `cpex.control.artifact.name`, and `cpex.control.artifact.id` (both conditional on the CPEX record carrying them). The flattened projection in `_build_flattened_attributes()` also emits `artifact.name`, `artifact.id`, `config.keys`, and `name` per record.

---

**3 — Duration unit**

Fixed in commit `4a24d1c`. Renamed to `cpex.control.duration_ns` everywhere — in `aggregate()`, `_per_control_attributes()`, and the flattened projection — following the OTel unit-suffix convention. The unit is **nanoseconds** throughout, identical across DB and OTel sinks. A rename rule like `cpex.control.duration_ns → controls.duration.ms` would change the key but not the value; operators remapping to a `ms`-suffixed name need to apply unit conversion in their collector pipeline. This is now documented inline.

---

**4 — Denial paths with `violations_as_exceptions=True`**

Structurally handled. `mark_denied(hook="pre/post")` is called in all three pre-invoke and one post-invoke `PluginViolationError` catch sites in `tool_service.py`, so `effective_allowed=False` is always captured correctly even when CPEX raises before returning.

The underlying gap you correctly identified — CPEX 0.1.2 itself does not append a `ControlExecutionRecord` for the denying plugin when `violations_as_exceptions=True` (it raises immediately at `cpex/framework/manager.py:680-682`) — means the emitted span captures records from plugins that ran before the denial but not the denier itself. This is a CPEX framework gap, documented inline at `tool_service.py:6629-6634`. An integration test that proves a real pre-invoke denial emits the correct span with `result.allowed=false` is still missing and is the most valuable remaining coverage item (see #11).

---

**5 — Additional hook paths**

Both paths addressed:

- `_run_timeout_post_invoke()` — wired in commit `4a24d1c`. It now accepts an optional `ctl_acc` parameter and calls `ctl_acc.add(timeout_post_result, hook="post")`. All four timeout call sites in `invoke_tool()` pass `ctl_acc=_ctl_acc`.
- `prepare_rust_mcp_tool_execution()` — intentionally not wired. This function builds an execution plan and returns before the actual tool call; there is no downstream flush point where `record_control_telemetry()` can be called. When the Rust runtime falls back to Python's `invoke_tool()`, the Python accumulator captures all telemetry. This is documented inline at lines 4468–4474.

---

**6 — Count semantics**

Fixed in commit `4a24d1c`:

- `invocation_count` now counts only controls with status in `{"completed", "error", "timeout"}` — skipped, disabled, and cancelled records are excluded, matching CPEX semantics.
- `results_count` is now `min(records_received, CPEX_CONTROL_TELEMETRY_MAX_RESULTS)` — the number of records exported after the cap, not the raw accumulated count.
- A new `cpex.control.records_received` field is emitted, giving downstream operators full accounting: `records_dropped = records_received - results_count + truncated`.

---

**7 — Truncation accounting**

Fixed in the latest commit. This was a genuine gap — the `islice(accumulator.records, max_results)` in both `_emit_db_spans()` and `_emit_otel_spans()` was silently discarding tier-3 export-cap drops without counting them anywhere.

The fix adds a `mark_export_cap_dropped(count)` method on `ControlTelemetryAccumulator`. In `record_control_telemetry()`, before building the summary span, `export_cap_dropped = max(0, records_received - max_results)` is computed and recorded. The `truncated` property now returns `_truncated + _export_cap_dropped`, so `cpex.control.truncated` on the summary span covers all three tiers:

- Tier 1: per-hook cap (64)
- Tier 2: per-call accumulation cap (128)
- Tier 3: export cap (CPEX_CONTROL_TELEMETRY_MAX_RESULTS, default 32)

5 new tests added in `TestExportCapTruncation`. Pre-existing test updated to assert the correct total across all three tiers.

On `CPEX_CONTROL_TELEMETRY_MAX_ATTRIBUTES` gateway-side enforcement: still informational only, documented in `config.py`. Gateway-side enforcement is tracked in the Phase 5 issue.

---

**8 — Mapping collisions and invalid fallback**

Fail-closed fix applied in commit `4a24d1c`. `apply_attribute_mapping()` now returns attributes unchanged on `ValueError` rather than applying partially-validated exact mappings that could bypass the `otel.*` guard. Compile-time collision detection is deferred to Phase 5.

---

**9 — Reason sanitization**

Addressed in commit `656e916`. Both `result.reason` and `result.error_code` are opt-in and disabled by default behind `CPEX_CONTROL_TELEMETRY_EMIT_REASON=false`. The `_emit_reason_enabled()` gate protects both `_per_control_attributes()` and the flattened projection. The `config.py` description explicitly warns about PII, tool argument values, and exception content, and requires the observability sink to be appropriately secured. Tested in `TestEmitReasonFlag` (6 cases).

---

**10 — Agent and binding identity**

Limitations documented inline in `tool_service.py` at lines 6609–6617. `agent_id = app_user_email ?? user_email` is noted as a human identity field, high-cardinality, with GDPR/data-residency implications. `binding_name = gateway_name ?? server_id` is acknowledged as a registry concept, not a formal control-binding concept. Both are explicitly provisional, with a reference to the Phase 5 follow-on for a dedicated agent-identity model.

---

**11 — End-to-end validation**

The current 7 E2E tests cover: summary span in DB, per-plugin result spans, trusted-fields check, per-control field schema, no-argument-leak, OTel SDK spans, and DB/OTel name parity.

Still missing, planned for Phase 5 follow-on (or can add to this PR if preferred):
- Real pre-invoke denial: span with `cpex.control.result.allowed=false` and `cpex.control.enforcement_point=pre`
- Real post-invoke denial after successful tool execution
- Timeout post-hook records in the emitted summary span
- Export-cap truncation verified end-to-end in DB span attributes
- OTLP collector verification against a live collector (acknowledged follow-on from the issue spec)

---

**Summary**

| # | Status |
|---|--------|
| 1 | Deferred by design — Phase 5 tracking issue to be filed; |
| 2 | Done — all fields present in fixed-schema spans |
| 3 | Done — duration_ns unit consistent and documented |
| 4 | Structural fix done — CPEX framework gap documented; |
| 5 | Done — timeout path wired; Rust path documented as out-of-scope |
| 6 | Done — count semantics correct, records_received added |
| 7 | Done — all three truncation tiers counted in cpex.control.truncated |
| 8 | Done — fail-closed on invalid config; compile-time collision deferred to Phase 5 |
| 9 | Done — opt-in gate default-off with security documentation |
| 10 | Done (documented) — identity limitations noted inline; formal model deferred to Phase 5 |
| 11 | Partial — denial-path and timeout E2E tests to be added |

for 1 and 11 i have added a followup issue here -> https://github.com/IBM/mcp-context-forge/issues/6017

vishu-bh (changes requested):
Thanks for the substantial work here. The invocation-local accumulator, bounded record counts, fixed-schema child spans, separate DB session, and best-effort emission model are good foundations. Test coverage for aggregation and sink behavior is also strong.

I found several issues that should be addressed before merge.

**Blocking**

1. `cpex.control.agent.id` currently contains `app_user_email` or `user_email` and is persisted/exported by default. Because central attribute mapping, removal, and redaction are not yet wired into either sink, this creates a new durable PII stream. Could this use a pseudonymous server-side identifier, become explicitly opt-in, or pass through the central policy before DB and OTel export?

2. This PR closes #5785, but several acceptance criteria remain incomplete:
   - mapping/removal rules are not applied to emitted control telemetry;
   - total attribute/serialized-size limits are not enforced;
   - the denying control cannot be identified when CPEX raises before returning its execution record;
   - the Rust MCP execution path explicitly skips this telemetry;
   - collision validation is deferred;
   - no real OTLP collector test verifies final exported attributes.

   I think either these gaps need completing here, or the PR should be narrowed and stop closing #5785, with explicit follow-up issues linked.

**Functionally impacting / security**

3. `config_keys` is limited to 64 entries, but individual keys and the joined value are not length-bounded or sanitized. Large or sensitive key names could cause data leakage or telemetry amplification. Please validate each key and cap the final serialized value, with adversarial tests.

4. Control telemetry and its DB sink default to enabled. A traced tool call can synchronously create one summary plus up to 32 result spans, in addition to OTel spans. This is a significant storage/export multiplier available to ordinary tool callers. Please consider a safe default, sampling/batching, and an enforced total byte/attribute budget.

5. `prepare_rust_mcp_tool_execution()` explicitly does not emit these records. This creates runtime-dependent observability and a monitoring blind spot, conflicting with the PR summary claim of every tool invocation.

**Test coverage**

Unit coverage is broad, but I did not find end-to-end coverage for pre-denial identity/error details, post-denial after successful execution, retries, cancellation, the Rust path, or final export through an actual OTLP collector.

Based on these findings, I do not think the change is safe to merge yet. The main concern is not the accumulator design itself; it is enabling and exporting incomplete telemetry before required privacy, policy, and amplification controls are in place.

gandhipratik203:
**`apply_attribute_mapping()` regression** — `plugins/utils.py:657-681`
Compiles the policy per span (its own docstring warns against this), and one bad rule silently voids the whole mapping behind a `logger.debug`. Validate at startup, cache.

**`PluginError` recorded as a denial** — `tool_service.py:5400-5416` (+3 sites)
`mark_denied()` on `PluginError` counts plugin outages as policy denials; the span can't distinguish.

**Dead conditional, wrong `re.escape`** — `plugins/utils.py`, `_apply_one_mapping_rename`
Both arms assign `idx = len(prefix)` (no-op), and matching an escaped pattern against a literal is wrong anyway. A capturing group `([^.]+)` removes the loop.

prakhar-singh1928:
Thanks  @vishu-bh , thanks @gandhipratik203  

**apply_attribute_mapping() recompiles policy per call (plugins/utils.py)**

**Fixed**. Added _cached_compile_attribute_policy() decorated with @functools.lru_cache(maxsize=256), keyed on a frozen (mapping_items, removals) tuple. apply_attribute_mapping() now hits the cache on every call after the first for any given config — compile happens once. Also upgraded the fallback log from logger.debug → logger.warning so a misconfigured mapping is visible at test time rather than silently swallowed in production.

**PluginError recorded as a denial (tool_service.py — 5 sites)**

**Fixed**. All 5 except (PluginError, PluginViolationError) blocks were split: except PluginViolationError retains the mark_denied() call (deliberate policy enforcement), while PluginError no longer calls mark_denied() at the inner catch sites — it propagates naturally to the outer except PluginError handler which emits partial telemetry without setting effective_allowed=False. Plugin outages and enforcement decisions are now distinguishable in the emitted span. The four inner bare re-raise blocks were also removed after pylint correctly flagged them as W0706 (try-except-raise with no side effects) — the outer handler covers PluginError correctly without them.

**Dead conditional + wrong re.escape in _apply_one_mapping_rename (plugins/utils.py)**

**Fixed**. Changed _SEGMENT_PATTERN from r"[^.]+" to r"([^.]+)" (capturing group). The entire manual split-and-scan loop — including both if/else arms that identically assigned idx = len(prefix) and the re.escape(prefix) string comparison against a literal — was replaced with 4 lines that read captured segments directly from m.group(i + 1). No behaviour change for valid configs; segment extraction is now correct by construction.

Validation after these fixes:

ruff — all checks passed
pylint — 10.00/10 (up from 9.99/10 before this round)
Plugin unit suite — 141 passed, 0 failures
@vishu-bh , all four blocking items from your formal review are now addressed across two commit rounds (f9af2e6 + this one). Happy for you to take another pass whenever convenient.

vishu-bh (changes requested):
Thanks for addressing the earlier privacy, default-on amplification, and `PluginError` classification concerns. Those changes materially improve the safety of this feature. I found four remaining items during the latest pass.

### 1. Functionally impacting — wildcard precedence changes after caching

`apply_attribute_mapping()` currently builds its cache key using:

```python
tuple(sorted(mapping.items()))
```

Sorting changes rule order, but the documented behavior says the first matching wildcard wins. This can change existing observability mappings:

```python
attrs = {"x.a": 1}
mapping = {
    "x.*": "first.*",
    "*.a": "second.*",
}
```

Based on insertion order, expected output is `{"first.a": 1}`. Current code produces `{"second.x": 1}` because sorting moves `"*.a"` first.

Could we preserve insertion order in the cached key and add a regression test with two overlapping wildcard rules?

### 2. Config-key limit counts characters rather than bytes

`_MAX_CONFIG_KEYS_JOINED_LEN` is documented as a byte limit, but both projections enforce it with:

```python
joined[:_MAX_CONFIG_KEYS_JOINED_LEN]
```

This limits Unicode code points rather than UTF-8 bytes. A value containing multibyte characters can therefore exceed the intended byte budget. Arbitrary control characters and commas are also retained.

Could this use the existing byte-aware `_safe_str(joined, _MAX_CONFIG_KEYS_JOINED_LEN)` helper? Tests covering multibyte input and control characters would help verify the boundary.

### 3. Denial E2E can pass without testing denial telemetry

The new denial-path tests are useful in intent, but currently:

- Nearly every response status is accepted, including `500`.
- Missing summary telemetry results in `pytest.skip()`.
- `"none"` is accepted as the enforcement point for a pre-invoke denial.

A regression that completely removes denial telemetry could therefore produce a skipped or passing test. Once the fixture initializes successfully, absence of the summary span should fail. The test should also require the expected denial response semantics and `enforcement_point == "pre"`.

### 4. First-plugin `PluginError` still produces no telemetry

The new outer `except PluginError` block says it emits partial telemetry without marking a policy denial. However, when the failing plugin is first in the chain, the accumulator contains no records and no denial flag. `record_control_telemetry()` then returns at its empty-accumulator guard, so no summary is emitted.

When earlier records exist, the summary can also report `result.allowed=True` without indicating that a later plugin failed.

Could plugin failure be represented explicitly in the accumulator or emitter—for example, with a separate error flag or status—so it remains distinguishable from both an allow decision and a policy denial? A first-plugin failure regression test would cover this path.

With these addressed, the implementation would be much closer to providing deterministic and trustworthy telemetry across error and policy paths.

prakhar-singh1928:
Thanks for the follow-up round, @vishu-bh . All four findings are addressed in the latest commit. Here's the breakdown:

**1 — Wildcard precedence changes after caching**

The sorted(mapping.items()) cache key was the bug — it silently reordered wildcard rules and changed which match fired first. Fixed by switching to tuple(mapping.items()) which preserves insertion order exactly as the caller declared it. The documented contract ("first matching wildcard wins") now holds end-to-end through the cache.

Regression test: TestWildcardInsertionOrderPreserved (3 cases) — covers first-declared-wins, reversed-order-changes-result, and exact-beats-wildcard-regardless-of-order.

**2 — Config-key limit counts characters rather than bytes**

Both projection sites (_per_control_attributes and _build_flattened_attributes) were using joined[:_MAX_CONFIG_KEYS_JOINED_LEN] which counts Unicode code points. Replaced with _safe_str(joined, _MAX_CONFIG_KEYS_JOINED_LEN) at both sites — _safe_str already encodes to UTF-8 before slicing so the byte budget is correctly enforced for multibyte key names.

Regression tests: TestConfigKeysByteCapAware (2 cases) — covers a CJK multibyte key that would have exceeded the byte budget under the old code, and an ASCII round-trip to confirm no regression.

**3 — Denial E2E can pass without testing denial telemetry**

**Three issues fixed:**

_enforcement_point() now checks acc.pre_denied / acc.post_denied flags in addition to acc.records. When CPEX raises PluginViolationError before appending a record (the known CPEX framework gap), the accumulator has no records but the flag is set — the function now correctly returns "pre" rather than "none".
E2E assertion tightened from in {"pre", "none"} → == "pre".
pytest.skip on missing summary span replaced with a hard assert — absence of a summary span on a live denial path is a regression, not an environment variation.
500 removed from accepted status codes — an unhandled exception is not a controlled denial.
Unit tests: TestEnforcementPointDenialFlags (3 cases).

**4 — First-plugin PluginError produces no telemetry**

The empty-accumulator guard in record_control_telemetry() only checked records, pre_denied, and post_denied — a first-plugin PluginError left all three unset and the function returned early, emitting nothing.

Fixed with three connected changes:

New _plugin_errored field and mark_plugin_error() method on ControlTelemetryAccumulator.
The outer except PluginError handler in tool_service.py now calls _ctl_acc.mark_plugin_error() before record_control_telemetry().
Empty-accumulator guard now includes plugin_errored so first-plugin failures always produce a summary span.
aggregate() emits cpex.control.plugin_error=True only when the flag is set — absent means no outage, making plugin failures explicitly distinguishable from both successful allows (result.allowed=True, no flag) and policy denials (result.allowed=False via mark_denied()).
effective_allowed remains True on PluginError — outage is not denial.
Unit tests: TestMarkPluginError (6 cases) covering the flag, effective_allowed invariant, aggregate() emission, and the full record_control_telemetry() path with an empty accumulator.

gandhipratik203 (dismissed):
Thanks for the changes @prakhar-singh1928. LGTM!

ja8zyjits:
## Blocking Issues

1. **CHANGELOG default wrong for `CPEX_CONTROL_TELEMETRY_ENABLED`**
   - File: `CHANGELOG.md:38`
   - Source of truth: `mcpgateway/config.py:1987`
   - Problem: CHANGELOG says default `true`; code uses `default=False`
   - Impact: Operators expect feature on by default when it is off
   - Fix: Change CHANGELOG to `(default false)`

2. **`record_control_telemetry()` not idempotent on repeated emission**
   - File: `mcpgateway/plugins/control_telemetry.py`
   - Problem: emission path calls `accumulator.mark_export_cap_dropped(...)` each time, so repeated calls on same accumulator inflate `cpex.control.truncated`
   - Impact: Wrong telemetry counts if helper runs more than once for same invocation state
   - Fix: Keep export-cap drop count derived-only at emit time, or guard update so tier-3 drop count is applied once

## Warnings

1. **Undefined "S4 principles" reference**
   - File: `mcpgateway/plugins/control_telemetry.py:10`
   - Problem: docstring mentions "S4 principles" with no local definition or linked reference
   - Impact: Future maintainers lack context
   - Fix: Define term inline or link security doc

2. **Repeated telemetry emission block in `tool_service.py`**
   - File: `mcpgateway/services/tool_service.py`
   - Problem: same `record_control_telemetry(...)` block repeated across success and multiple exception paths
   - Impact: Future edits can drift across branches in large service file
   - Fix: Extract small local helper inside `invoke_tool()`

3. **Module-global compatibility cache has no mutex**
   - File: `mcpgateway/plugins/cpex_compat.py:45`
   - Problem: `_EXECUTION_RECORDS_SUPPORTED` written without `threading.Lock`
   - Impact: Theoretical first-call race; no correctness risk because write is idempotent
   - Fix: Low-priority follow-on if path becomes hot or strict once-only semantics matter

prakhar-singh1928:
Thanks @ja8zyjits : all four findings addressed in commit 9f92f080b.

Blocking 1 — CHANGELOG default: Fixed. CHANGELOG.md:38 now says (default false) matching config.py:1987.

Blocking 2 — Idempotency: Fixed. mark_export_cap_dropped() is now first-write-wins (if count > 0 and self._export_cap_dropped == 0). Two regression tests added in TestExportCapTruncation: test_mark_export_cap_dropped_is_idempotent (three calls → count unchanged) and test_mark_export_cap_dropped_zero_then_positive (zero call doesn't consume the slot).

Warning 1 — "S4 principles": Replaced with "bounded cardinality and field-level sanitization … _safe_str/_safe_num helpers and field-name allowlists in plugins/utils.py".

Warning 2 — Repeated block: Extracted all 5 record_control_telemetry() sites to a _emit_ctl_telemetry() nested function defined right after _ctl_acc creation. Identity-mapping rationale consolidated in its docstring.

826 plugin unit tests pass, ruff clear, pylint 10.00/10. Warning 3 (mutex on compat cache) remains low-priority follow-on as previously noted.

mcpgateway/plugins/control_telemetry.py
```diff
@@ -0,0 +1,804 @@
...
+        results_count = min(records_received, _get_max_results())
+
+        result: dict = {
+            "cpex.control.invocation_count": invocation_count,
+            "cpex.control.matched_count": matched_count,
+            "cpex.control.applied_count": applied_count,
+            "cpex.control.records_received": records_received,
+            "cpex.control.results_count": results_count,
+            "cpex.control.duration_ns": total_duration_ns,  # nanoseconds — OTel unit suffix convention
+            "cpex.control.result.allowed": self.effective_allowed,
```
vishu-bh: `mark_plugin_error()` makes the outage visible, but `effective_allowed` still returns `True` because it only checks the pre/post denial flags. For a first-plugin `PluginError`, the summary therefore reports:

```text
cpex.control.plugin_error = true
cpex.control.result.allowed = true
cpex.control.enforcement_point = none
```
The exception is re-raised and the request fails, so downstream dashboards could incorrectly count this as an allowed control chain. Could we track the hook in mark_plugin_error(hook="pre" | "post") and omit/null result.allowed when the effective decision is indeterminate, or derive it from the configured on_error policy? A first-plugin error integration test should verify these semantics.

vishu-bh (changes requested):
Thanks for addressing the comment just last bit of findings

Config keys are byte-bounded but not sanitized. Commas, CR/LF, control characters, and sensitive text embedded in key names remain exportable. Joined CSV also becomes ambiguous.
Use printable/identifier validation or a bounded structured list. Add comma, control-character, and secret-marker tests.

gandhipratik203 (dismissed):
LGTM, once the CI passes.

vishu-bh (dismissed):
LGTM 🚀

ja8zyjits (dismissed):
LGTM

ja8zyjits (approved):
LGTM

msureshkumar88 (approved):

PR #6004, opened 2026-07-30
Fix OAuth team scope resolution for missing or malformed token_teams

### LINKED ISSUE #5980: OAuth scope resolution treats indeterminate admin team state as unrestricted
### Bug Summary

`_resolve_token_teams_for_scope_check()` treats missing or malformed
`request.state.token_teams` as unrestricted when `current_user.is_admin` is
true and no cached JWT payload is available.

This makes an indeterminate team scope indistinguishable from an explicitly
unrestricted scope.

### Affected Component

- [x] `mcpgateway` - API

Relevant code: `mcpgateway/routers/oauth_router.py`

### Steps to Reproduce

1. Create an admin user context with `is_admin=true`.
2. Create a request whose `state.token_teams` is absent or has an unexpected
   type.
3. Ensure `request.state._jwt_verified_payload` is unavailable.
4. Call `_resolve_token_teams_for_scope_check()` or an endpoint relying on it.
5. Observe that the result is `None`, representing unrestricted scope.

### Expected Behavior

Explicit unrestricted scope should remain allowed. Missing or malformed scope
state should be resolved from another trusted source or handled as
indeterminate instead of automatically becoming unrestricted.

Authentication methods that legitimately have no team narrowing, such as
configured platform-admin access, should continue to work.

### Acceptance Criteria

- [ ] Explicit `token_teams=None` remains unrestricted for eligible admins.
- [ ] Malformed `token_teams` does not resolve to unrestricted scope.
- [ ] Missing state is recovered from a trusted user or token context when
      available.
- [ ] Legitimate admin authentication without team narrowing remains supported.
- [ ] Tests cover `None`, empty list, non-empty list, missing state, malformed
      state, and cached-payload recovery.

### Logs / Error Output

No error is emitted. Current behavior returns unrestricted scope.

### Environment Info

Observed through source review of current OAuth routing and authentication
helpers.

Sagar-Mondal:
# Pull Request

## 🔗 Related Issue

Closes #5980 

---

## 📝 Summary

Hardens `_resolve_token_teams_for_scope_check()` in `oauth_router.py` to prevent missing or malformed `request.state.token_teams` from being silently promoted to unrestricted admin scope (`None`).

**Before:** When `token_teams` was absent or had an unexpected type (e.g., string, int, dict) and no cached JWT payload was available, the function returned `None` for admin users — granting unrestricted scope indistinguishable from an explicitly un-narrowed admin.

**After:** Only explicitly resolved `token_teams=None` from a trusted source (`request.state` or cached JWT payload) can produce unrestricted scope. Missing or malformed state now:
1. Logs a warning for malformed values
2. Attempts recovery from the cached `_jwt_verified_payload`
3. Fails closed to public-only (`[]`) if no trusted source is available — regardless of admin status

Additionally:
- Extracts duplicated admin-flag logic into a shared `_extract_is_admin()` helper for consistency across `_resolve_token_teams_for_scope_check` and `_enforce_gateway_access`
- Extracts JWT recovery logic into a standalone `_recover_token_teams_from_jwt()` helper for isolated testability
- **Validates `teams` claim type in `_recover_token_teams_from_jwt()` before delegating to `normalize_token_teams()`** — prevents `TypeError` on non-iterable types (`int`, `bool`) and misleading team IDs from iterating strings (char-by-char) or dicts (over keys). Malformed claims are logged at WARNING and fail closed to `[]` via the caller.
- Collapses duplicated `is_admin` extraction in the recovery branch to reuse `_extract_is_admin()` — also incidentally fixes a latent `AttributeError` where the old code called `.get()` on a JWT `user` claim without checking it was a dict first
- Adds debug logging when token_teams is successfully recovered from the cached JWT payload
- Uses consistent "unrestricted" terminology throughout (replacing "un-scoped")

---

## 📏 Reviewability

---

## 🏷️ Type of Change

- [x] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [x] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

**Test coverage for all acceptance criteria:**

| Scenario | Test(s) | Expected |
|----------|---------|----------|
| Explicit `token_teams=None` (admin) | `test_explicit_none_admin_unrestricted`, `test_legitimate_admin_with_explicit_unrestricted_token_teams` | Returns `None` (unrestricted) ✅ |
| Explicit `token_teams=None` (non-admin) | `test_explicit_none_non_admin_passes_through` | Returns `None` (caller-downgraded) ✅ |
| Empty list `[]` (admin) | `test_empty_list_admin_public_only` | Returns `[]` (no bypass) ✅ |
| Non-empty list | `test_non_empty_list_preserved`, `test_non_empty_list_admin_narrowed` | Returns team list ✅ |
| Missing state (admin, no JWT cache) | `test_missing_state_admin_fails_closed` | Returns `[]` (fail-closed) ✅ |
| Missing state (non-admin, no JWT cache) | `test_missing_state_non_admin_fails_closed` | Returns `[]` ✅ |
| Malformed state — string, int, dict, bool | `test_malformed_state_*_fails_closed` (4 tests) | Returns `[]` (fail-closed) ✅ |
| Missing state + cached JWT recovery | `test_missing_state_recovered_from_cached_jwt_admin`, `_narrowed` | Recovers from payload ✅ |
| Malformed state + cached JWT recovery | `test_malformed_state_recovered_from_cached_jwt`, `_admin_unrestricted` | Recovers from payload ✅ |
| Cached JWT `teams=[]` + admin | `test_cached_jwt_public_only_admin_stays_scoped` | Returns `[]` (no bypass) ✅ |
| JWT `is_admin` overrides `current_user` | `test_cached_jwt_overrides_current_user_admin_flag` | JWT is authority ✅ |
| `_extract_is_admin` delegation | `test_uses_extract_is_admin_helper` | Nested dict supported ✅ |
| Cached JWT with empty dict payload (admin) | `test_cached_jwt_empty_dict_payload_fails_closed` | Returns `[]` (fail-closed) ✅ |
| Cached JWT with non-dict payload (admin) | `test_cached_jwt_non_dict_payload_fails_closed` | Returns `[]` (fail-closed) ✅ |
| Cached JWT with `None` payload in tuple (admin) | `test_cached_jwt_none_payload_in_tuple_fails_closed` | Returns `[]` (fail-closed) ✅ |
| Cached JWT with `user` claim as string | `test_cached_jwt_user_claim_as_string_fails_closed` | No `AttributeError`, fails closed ✅ |
| **Malformed `teams` in cached JWT — int** | `test_malformed_teams_int_returns_none` | Recovery returns `None` → caller assigns `[]` ✅ |
| **Malformed `teams` in cached JWT — bool** | `test_malformed_teams_bool_returns_none` | Recovery returns `None` → caller assigns `[]` ✅ |
| **Malformed `teams` in cached JWT — string** | `test_malformed_teams_string_returns_none` | Recovery returns `None` → caller assigns `[]` ✅ |
| **Malformed `teams` in cached JWT — dict** | `test_malformed_teams_dict_returns_none` | Recovery returns `None` → caller assigns `[]` ✅ |
| `_recover_token_teams_from_jwt` isolation | `TestRecoverTokenTeamsFromJwt` (15 tests) | All edge cases covered ✅ |
| `_enforce_gateway_access` with missing state | `test_enforce_gateway_access_missing_token_teams_admin_fails_closed` | Deny-path: 403 with `[]` scope ✅ |
| `_enforce_gateway_access` with malformed state | `test_enforce_gateway_access_malformed_token_teams_admin_fails_closed` | Deny-path: 403 with `[]` scope ✅ |
| **`_enforce_gateway_access` team-visibility cache hit** | `test_enforce_gateway_access_team_visibility_cache_hit_member_allowed` | Allow-path: `TeamManagementService.get_user_role_in_team` grants access ✅ |

**Updated characterization tests** in `TestOAuthClientManagementScopeGuard`:
- `test_malformed_token_teams_admin_denied` — now expects 403 (was allowed)
- `test_missing_token_teams_admin_denied` — now expects 403 (was allowed)
- Parametrized `absent-attribute` case — now expects denial

---

## ✅ Checklist

---

## 📓 Notes

**Files changed:**
- `mcpgateway/routers/oauth_router.py` — Hardened `_resolve_token_teams_for_scope_check()`, new `_extract_is_admin()` helper, new `_recover_token_teams_from_jwt()` helper (extracted JWT recovery logic with teams-claim type validation), recovery branch reuses `_extract_is_admin(payload)` instead of duplicating admin extraction, enhanced docstrings, debug logging on successful JWT recovery
- `tests/unit/mcpgateway/routers/test_oauth_router.py` — 23 new tests in `TestResolveTokenTeamsForScopeCheck`, 15 new tests in `TestRecoverTokenTeamsFromJwt` (11 original + 4 malformed teams-claim types: int, bool, str, dict), 4 new tests for `_enforce_gateway_access` in `TestOAuthAccessHelpers` (2 deny-path tests for missing/malformed state + 1 regression test verifying `TeamManagementService.get_user_role_in_team` is the sole membership-check path + 1 unknown-visibility test with request-scoped `token_teams`), 1 updated test in `TestOAuthRouter` (`test_resolve_token_teams_for_scope_check_admin_missing_state_fails_closed` replaces `test_resolve_token_teams_for_scope_check_admin_attribute_fallback`), updated 3 characterization tests in `TestOAuthClientManagementScopeGuard`

**Design decisions:**
- **Fail-closed by default:** Missing/malformed state → `[]` (public-only), never `None` (unrestricted). This matches `get_token_teams_from_request` semantics in `auth_context.py`.
- **JWT payload as recovery source:** When `token_teams` is missing or malformed, the cached `_jwt_verified_payload` is consulted. The `is_admin` flag from the JWT payload takes precedence over `current_user` during recovery, since the token is the authority.
- **`_recover_token_teams_from_jwt` extraction:** The JWT recovery block is now a standalone helper returning `tuple[list[str] | None, bool] | None`. This makes it independently testable and avoids the recovery logic being tightly coupled with the main resolver's control flow.
- **Teams-claim type validation in recovery:** `_recover_token_teams_from_jwt()` validates the `teams` claim is `None` or `list` before delegating to `normalize_token_teams()`. This prevents `TypeError` on non-iterable types (int, bool) and misleading team IDs from string iteration (char-by-char) or dict iteration (over keys). The type contract `tuple[list[str] | None, bool] | None` is now enforced at runtime, not just by annotation.
- **`_extract_is_admin` reuse in recovery:** The recovery branch previously duplicated 3 lines of admin-flag extraction that `_extract_is_admin()` already handles. Since the payload is confirmed to be a `dict` at that point, `_extract_is_admin` handles it via its `isinstance(current_user, dict)` branch. This also incidentally fixes a latent `AttributeError` where the old code called `.get("is_admin", False)` on a JWT `user` claim without first checking it was a dict.
- **Warning and debug logs:** Malformed `token_teams` types are logged at WARNING level for forensic visibility without disrupting request flow. Successful recovery from cached JWT payload is logged at DEBUG level for troubleshooting.
- **Falsy/non-dict cached JWT payload:** When the cached `_jwt_verified_payload` tuple exists but the payload is falsy (e.g., `{}`, `None`) or not a dict, it is treated as an untrusted recovery source and fails closed to `[]`. This prevents a present-but-empty payload from leaving `token_teams` in an invalid state.
- **`_enforce_gateway_access` allow- and deny-path tests:** The resolver was thoroughly unit-tested, but the actual caller (`_enforce_gateway_access`) was only tested with well-typed `token_teams=None`. New tests verify the full deny-path through the gateway access check with missing/malformed state. The allow-path regression test verifies that `TeamManagementService.get_user_role_in_team` (the cache-aware path from PR #6589) is the sole membership-check path — if someone reverted to the old `EmailAuthService + user.is_team_member()` pattern, the patched mock would never execute. An additional unknown-visibility test exercises the interaction between `_resolve_token_teams_for_scope_check` and the team-member fallback when a request object carries `token_teams`.


msureshkumar88 (dismissed):
## E2E verification — PASS

Ran a real end-to-end check against this branch: a live `uvicorn` gateway server (not mocks), real JWTs minted via the project's own `create_jwt_token` CLI, real HTTP requests, a real seeded team + team members via the API, and a direct comparison of the actual `_resolve_token_teams_for_scope_check()` function loaded from `main` vs. this PR's head (via an isolated `git worktree`, cleaned up afterward). Script added at `scripts/verify-pr-6004-e2e.sh` (untracked, following this repo's existing `scripts/verify-pr-*-e2e.sh` convention).

**Result: 54/54 e2e assertions passed, plus the PR's own `tests/unit/mcpgateway/routers/test_oauth_router.py` suite (162 tests) passed.**

### Config used

```
DATABASE_URL=sqlite:////tmp/.../pr6004_e2e.db   # isolated scratch DB, not ./mcp.db
GW_PORT=8020                                     # existing dev server on :8000 untouched
JWT_SECRET_KEY=<test-only value>
AUTH_REQUIRED=true
PLUGINS_ENABLED=false
```
Commit under test: `b96c1a8d3` (this PR's head).

### Commands run

```
bash scripts/verify-pr-6004-e2e.sh
```
which internally: greps every `request.state.token_teams =` assignment site to confirm it's always type-guarded; starts a live server on :8020; mints JWTs via `python -m mcpgateway.utils.create_jwt_token` for unrestricted-admin / narrowed-admin / public-only-admin / non-admin / team-member / non-member identities; creates a real team via `POST /teams/`; curls `GET/DELETE /oauth/registered-clients...` and `GET /oauth/status/{gateway_id}`; checks out `origin/main` into a scratch `git worktree` and runs the real `_resolve_token_teams_for_scope_check()` from both refs via `PYTHONPATH`-isolated subprocesses; runs the PR's pytest suite; tears everything down (`git worktree remove`, kill server, delete scratch DB).

### Output — DCR admin routes (`_require_unnarrowed_admin`)

| Identity | GET /oauth/registered-clients | GET .../{gw_id} | DELETE .../{client_id} |
|---|---|---|---|
| unrestricted-admin | 200 | 404 (gate passed) | 404 (gate passed) |
| narrowed-admin | 403 | 403 | 403 |
| public-only-admin | 403 | 403 | 403 |
| non-admin | 403 | 403 | 403 |

### Output — gateway access (`_enforce_gateway_access`), `GET /oauth/status/{gateway_id}`

| Identity | public gw | team gw (team A) | private gw (owner=admin) |
|---|---|---|---|
| unrestricted-admin | 200 | 200 | 200 |
| admin, member of team A | 200 | 200 | 403 |
| admin, member of team B (wrong team) | — | 403 | — |
| team A member (non-admin) | 200 | 200 | 403 |
| team B member (non-admin, wrong team) | 200 | 403 | 403 |
| anonymous | 401 | 401 | 401 |

### Output — the actual bug/fix, `main` vs. this PR (direct function call, real code, not reimplemented)

| Case | `origin/main` | this PR |
|---|---|---|
| `token_teams` missing, no cached JWT | `None` (bug — unrestricted admin bypass) | `[]` (fixed — fails closed) |
| `token_teams` malformed (`"not-a-list"`) | `None` (bug) | `[]` (fixed) |

This confirms the fix resolves #5980 as described.

### Surrounding-code regression sweep

No regressions. One good sign: a narrowed admin (real, active `is_admin=True` in the DB) still correctly got `403` on a gateway scoped to a team they aren't a member of — confirming the `_extract_is_admin`/`_resolve_token_teams_for_scope_check` refactor didn't collapse the two-layer model (token scoping still gates before the admin-bypass check runs).

One unrelated observation surfaced during setup, not a defect in this PR: `TokenScopingMiddleware._check_team_membership()` rejects any team-scoped token whose claimed team the caller isn't actually a DB member of, for any request — pre-existing, orthogonal behavior, worth knowing if writing similar e2e tests for other PRs.

Not separately exercised (scope calls, not gaps in the fix): a non-admin-owned private gateway ownership match, and the sibling `_enforce_gateway_access` call sites (`/oauth/fetch-tools/{id}`, `/oauth/authorize/{id}`) — `/oauth/status/{id}` shares the identical gate function so this was a representative, not exhaustive, probe.

---

## Summary — no changes required

Nice, focused fix — closes exactly what #5980 asks for, with thorough unit coverage (23 new tests) and now e2e-verified against the real running server and against `main`'s actual (buggy) behavior. Backend-only, no schema/migration involved (correctly so — this is pure request-scope logic). Approving.

A few optional, non-blocking follow-ups if you're touching this area again — happy to open a small issue for these if useful, no pressure either way:

1. `oauth_router.py` (the cached-JWT recovery branch) rebuilds the same `is_admin` extraction that `_extract_is_admin()` — introduced in this very PR — already does. Could likely collapse those 3 lines to `is_admin = _extract_is_admin(payload)` since `payload` is already dict-checked at that point.
2. `_enforce_gateway_access` is tested with well-typed `token_teams=None`, but not with the missing/malformed state that's the actual subject of this PR — worth one more test there for full deny-path coverage alongside the DCR-route tests.
3. Might be worth a one-line mention in the PR description that this also incidentally fixes a latent `AttributeError` (old code called `.get()` on a JWT `user` claim without checking it was a dict first) — a nice side benefit that's easy to miss in the diff.

ja8zyjits (changes requested):
## Key Improvements

### 1. JWT Recovery Logic Extraction
**Priority:** Low | **Impact:** Maintainability

Extract JWT recovery block into helper:

```python
def _recover_token_teams_from_jwt(request: Request) -> tuple[list[str] | None, bool] | None:
    """Attempt to recover token_teams from cached JWT payload.
    
    Returns:
        Tuple of (token_teams, is_admin) if recovery succeeds, None otherwise.
    """
    cached = getattr(request.state, "_jwt_verified_payload", None)
    if not (cached and isinstance(cached, tuple) and len(cached) == 2):
        return None
    
    _, payload = cached
    if not (isinstance(payload, dict) and payload):
        return None
    
    token_teams = normalize_token_teams(payload)
    user_claim = payload.get("user", {})
    user_is_admin = user_claim.get("is_admin", False) if isinstance(user_claim, dict) else False
    is_admin = bool(payload.get("is_admin", False) or user_is_admin)
    
    return (token_teams, is_admin)
```

**Benefit:** Easier to test recovery logic in isolation.

---

### 2. Return Value Documentation Enhancement
**Location:** `mcpgateway/routers/oauth_router.py:296-298`

**Current:**
```python
Returns:
    ``None`` for explicitly un-scoped requests (which callers must downgrade for non-admins),
    or a normalized team list for scoped access.
```

**Improved:**
```python
Returns:
    ``None`` for explicitly un-scoped requests (which callers must downgrade for non-admins),
    a non-empty list for team-scoped access, or ``[]`` for public-only scope (including
    fail-closed cases when no trusted source is available).
```

---

### 3. `_extract_is_admin` Helper Documentation
**Location:** `mcpgateway/routers/oauth_router.py:373-389`

**Current:**
```python
"""Extract admin flag from typed or dict user contexts.

Args:
    current_user: Authenticated user context.

Returns:
    ``True`` when the user context indicates admin privileges.
```

**Improved:**
```python
"""Extract admin flag from typed or dict user contexts.

Supports both flat dict structures (``{"is_admin": True}``) and nested
structures (``{"user": {"is_admin": True}}``), matching JWT payload formats.

Args:
    current_user: Authenticated user context (typed object or dict).

Returns:
    ``True`` when the user context indicates admin privileges.
```

---

### 4. Terminology Consistency
**Location:** `mcpgateway/routers/oauth_router.py:297`

**Current:** "explicitly un-scoped requests"

**Suggested:** Use "unrestricted" consistently:
```python
Returns:
    ``None`` for unrestricted admin scope (which callers must downgrade for non-admins),
    a non-empty list for team-scoped access, or ``[]`` for public-only scope.
```

**Rationale:** "Unrestricted" used in SECURITY comment (line 286), "un-narrowed" in `_require_unnarrowed_admin` (line 284). Consistent terminology reduces cognitive load.

---

### 5. Debug Logging for Recovery Success
**Priority:** Low | **Impact:** Troubleshooting

Add debug log when recovery succeeds:
```python
if recovered:
    logger.debug(
        "_resolve_token_teams_for_scope_check: recovered token_teams from cached JWT payload",
    )
```

ja8zyjits:
Hi @Sagar-Mondal 

### 1. 💡 SUGGESTION - Helper Function Scope Documentation

**Location:** `mcpgateway/routers/oauth_router.py:305-313`

**Finding:** `_recover_token_teams_from_jwt` docstring states:

```python
"""Attempt to recover token_teams from cached JWT payload.

When ``request.state.token_teams`` is missing or malformed, this helper
inspects the cached verified JWT payload (set by the auth middleware) to
re-derive a well-typed ``token_teams`` value and the associated admin flag.

Returns:
    Tuple of ``(token_teams, is_admin)`` if recovery succeeds, ``None`` otherwise.
"""
```

**Suggestion:** Add a note about what "recovery succeeds" means for clarity:

```python
Returns:
    Tuple of ``(token_teams, is_admin)`` if recovery succeeds (cached payload
    exists, is well-formed tuple, and contains non-empty dict), ``None`` otherwise.
```

**Rationale:** Lines 315-321 show specific validation criteria that constitute "success". Making this explicit helps maintainers understand when recovery fails.

**Priority:** Low - Current wording is acceptable, suggestion improves precision

### 2. 💡 CLARIFICATION - Empty Dict Payload Behavior

**Location:** `mcpgateway/routers/oauth_router.py:319-321`

**Finding:** Code checks for empty dict:

```python
if not (isinstance(payload, dict) and payload):
    return None
```

**Documentation:** Not explicitly mentioned in docstring that empty dict `{}` is treated as untrusted.

**Suggestion:** Add to docstring:

```python
Returns:
    Tuple of ``(token_teams, is_admin)`` if recovery succeeds (cached payload
    exists, is well-formed tuple, and contains non-empty dict), ``None`` otherwise.
    Empty dict payloads are treated as untrusted and fail recovery.
```

**Verification:** Test `test_cached_jwt_empty_dict_payload_fails_closed` confirms this behavior.

**Priority:** Low - Behavior is tested and correct, documentation could be more explicit

ja8zyjits:
HI @Sagar-Mondal 

## Executive Summary

**Verdict**: **REQUEST_CHANGES**  
**Summary**: Hardening of `_resolve_token_teams_for_scope_check()` against indeterminate `token_teams` state is sound and test coverage is comprehensive. 1 blocking ordering issue and 1 functionally-impacting check require resolution before merge.

| Severity | Count | Summary |
|----------|-------|---------|
| 🚨 **Blocking** | 1 | Forward reference / helper definition order (`ARCH-1`) |
| ⚠️ **Functionally-Impacting** | 1 | Fragile truthiness check on recovered tuple (`ARCH-2`) |
| 💡 **Suggestions / Minor** | 5 | Formatting (`ARCH-S1`/`TQ-S1`), docstring clarity (`ARCH-S2`, `DOC-1`), test count in PR desc (`DOC-2`), CHANGELOG note (`DOC-3`), test parametrization (`TQ-S2`/`TQ-S3`) |

---

## 🚨 Blocking Issues

### ARCH-1: Function Call Precedes Definition (`_extract_is_admin`)

- **Location**: [`mcpgateway/routers/oauth_router.py:481`](mcpgateway/routers/oauth_router.py:481), [`mcpgateway/routers/oauth_router.py:525`](mcpgateway/routers/oauth_router.py:525)
- **Problem**: `_resolve_token_teams_for_scope_check` (line 461) calls `_extract_is_admin` (line 481), but `_extract_is_admin` is defined later at line 525. While resolved dynamically at call time in Python, this violates codebase conventions and risks linter/mypy ordering diagnostics.
- **Fix**: Move `_extract_is_admin` above both `_recover_token_teams_from_jwt` and `_resolve_token_teams_for_scope_check`.

---

## ⚠️ Functionally-Impacting Issues

### ARCH-2: Semantically Fragile Truthiness Guard on Recovered Tuple

- **Location**: [`mcpgateway/routers/oauth_router.py:500`](mcpgateway/routers/oauth_router.py:500)
- **Problem**: 
  ```python
  recovered = _recover_token_teams_from_jwt(request)
  if recovered:
      token_teams, is_admin = recovered
  ```
  `_recover_token_teams_from_jwt` returns `None` or `tuple[list[str] | None, bool]`. While any 2-tuple is truthy today, checking truthiness rather than explicit non-null identity is fragile if tuple return patterns evolve.
- **Fix**:
  ```python
  if recovered is not None:
      token_teams, is_admin = recovered
  ```

---

## 💡 Suggestions & Minor Notes

### 1. ARCH-S1 / TQ-S1: Extra Blank Lines (Ruff E303)
- **Location**: [`tests/unit/mcpgateway/routers/test_oauth_router.py:1398-1401`](tests/unit/mcpgateway/routers/test_oauth_router.py:1398)
- **Problem**: Three blank lines before `TestResolveTokenTeamsForScopeCheck` class definition violate PEP 8 / Ruff `E303` (max 2 allowed).
- **Fix**: Remove extra blank line.

### 2. DOC-1: Inaccurate Attribution in Docstring
- **Location**: [`mcpgateway/routers/oauth_router.py:440`](mcpgateway/routers/oauth_router.py:440)
- **Problem**: Docstring states `_jwt_verified_payload` is set by "the auth middleware", but it is set in `mcpgateway/utils/verify_credentials.py`.
- **Fix**: Update docstring to: `set by verify_credentials.py during JWT verification`.

### 3. DOC-2: PR Description Test Count Mismatch
- **Location**: `issue_5980_pr_6004.md` (lines 86, 116)
- **Problem**: Claims 10 tests in `TestRecoverTokenTeamsFromJwt`; diff actually contains 11 tests (omits `test_user_claim_as_string_safe`).

### 4. DOC-3: Missing CHANGELOG Entry
- **Location**: `CHANGELOG.md`
- **Problem**: Observable security behavior change (admin with missing/malformed state and no cached JWT now fails closed to `[]`/403 instead of `None`/unrestricted) should be noted in CHANGELOG under unreleased changes.

### 5. TQ-S2: Parametrize Malformed State Tests
- **Location**: [`tests/unit/mcpgateway/routers/test_oauth_router.py:1479-1515`](tests/unit/mcpgateway/routers/test_oauth_router.py:1479)
- **Suggestion**: Collapse repetitive tests (`str`, `int`, `dict`, `bool`) using `@pytest.mark.parametrize` and include extra non-standard types (`float`, `tuple`, `set`).

### 6. TQ-S3: Test Invalid Inner List Elements
- **Location**: [`mcpgateway/routers/oauth_router.py:487`](mcpgateway/routers/oauth_router.py:487)
- **Suggestion**: Add unit test for list containing non-string items (e.g. `[123, None]`) to document downstream handling.

---

ja8zyjits:
## 🔴 Functionally Impacting

### FI-1 — Malformed cached JWT `teams` claim can raise `TypeError` in recovery path

**File:** [`mcpgateway/routers/oauth_router.py:457`](mcpgateway/routers/oauth_router.py:457)

`_recover_token_teams_from_jwt()` passes any non-empty dict payload directly to `normalize_token_teams(payload)`. `normalize_token_teams()` iterates `payload["teams"]` without validating it is iterable. A cached payload like `{"teams": 42}` raises `TypeError`. Because recovery runs after malformed/missing `request.state.token_teams`, this turns the intended fail-closed path into a 500. String or dict `teams` claims can also be iterated, producing misleading team IDs.

**Type contract also broken:** `_recover_token_teams_from_jwt` promises `tuple[list[str] | None, bool] | None` but the runtime `teams` shape is unchecked — annotation hides a possible runtime exception.

**Fix:** Validate `teams` claim as `None` or `list` before calling `normalize_token_teams`, or catch `TypeError` in recovery and return `None` (caller assigns `[]`). Log malformed claim type at WARNING. Add tests for `int`, `bool`, `str`, and `dict` `teams` values in cached payloads.

---

### FI-2 — Async test mis-attributed to wrong class, missing `@pytest.mark.asyncio`

**File:** [`tests/unit/mcpgateway/routers/test_oauth_router.py:2220`](tests/unit/mcpgateway/routers/test_oauth_router.py:2220)

```python
# no blank line, wrong class, no decorator
        assert is_admin is False
    async def test_enforce_gateway_access_team_visibility_cache_hit_member_allowed(self, mock_db):
```

`test_enforce_gateway_access_team_visibility_cache_hit_member_allowed` was inserted immediately after `TestRecoverTokenTeamsFromJwt`'s last method with no class break. It belongs in `TestOAuthAccessHelpers`. Passes today only because `asyncio_mode=AUTO`; if mode changes to `strict` it silently stops running. Static analysis tools flag the missing decorator.

**Fix:** Move method into `TestOAuthAccessHelpers`, add `@pytest.mark.asyncio`, add blank line.

---

## 🟡 Suggestions

### S-1 — `_enforce_gateway_access` dual `is_admin` derivation needs documentation

**File:** [`mcpgateway/routers/oauth_router.py:579-586`](mcpgateway/routers/oauth_router.py:579)

```python
requester_is_admin = _extract_is_admin(current_user)          # from DB-resolved current_user
token_teams = _resolve_token_teams_for_scope_check(...)        # may override is_admin from JWT
if token_teams is None:
    if requester_is_admin:                                     # uses DB is_admin, not JWT is_admin
        return
```

The resolver internally overrides `is_admin` from the JWT payload during recovery. The caller re-derives `is_admin` from `current_user` independently. When JWT says admin but DB says non-admin: resolver returns `None`, but the early-return gate uses `current_user.is_admin=False` → falls to `token_teams=[]` → **correct deny, but by accident** (two inconsistencies cancel out). No exploit today, but latent clarity bug that will confuse reviewers.

**Fix:** Either (a) expose `resolved_is_admin` as a second return value from the resolver, or (b) add a comment explaining the two derivations are intentionally separate (DB authority for the gate; JWT recovery internal to resolver).

---

### S-2 — `if recovered:` should be `if recovered is not None:`

**File:** [`mcpgateway/routers/oauth_router.py:503`](mcpgateway/routers/oauth_router.py:503)

Functionally identical today (all non-`None` returns are 2-tuples, always truthy). But `is not None` matches the `tuple[...] | None` return annotation, makes intent unambiguous, and protects future readers from misreading the check as a truthiness guard on tuple content. Cheap fix.

---

### S-3 — Per-test local imports add noise in new test classes

Every test in `TestResolveTokenTeamsForScopeCheck` and `TestRecoverTokenTeamsFromJwt` repeats:
```python
from mcpgateway.routers.oauth_router import _resolve_token_teams_for_scope_check
```
Class-level or module-level import would reduce duplication across 34 tests. Style choice, not blocking.

---

### S-4 — No direct test for `_extract_is_admin({"user": "string_not_dict"})` → `False`

The `isinstance(user_claim, dict)` guard fix is exercised only indirectly via `test_uses_extract_is_admin_helper`. A direct standalone unit test for the string-user-claim path would make the fix's coverage explicit.

---

### S-5 — `test_recovers_public_only` `is_admin=True` return may confuse readers

`test_recovers_public_only` asserts `is_admin is True` from `_recover_token_teams_from_jwt` (correct for the helper in isolation), but the caller's admin-strip logic in `_resolve_token_teams_for_scope_check` resets `is_admin=False` when `token_teams==[]`. A comment in the test: *"caller's admin-strip applies after this"* would prevent trace-path confusion.

---

## 🔵 Minor Notes

### M-1 — Demote speculative recovery log from WARNING to DEBUG

**File:** [`mcpgateway/routers/oauth_router.py:495-515`](mcpgateway/routers/oauth_router.py:495)

When `is_malformed=True` and recovery fails, two WARNINGs fire: one before ("attempting recovery") and one after ("failing closed"). The first is speculative — downgrade to DEBUG. Reserve WARNING for confirmed failures only.

---

### M-2 — Ruff style errors in changed test file

**File:** [`tests/unit/mcpgateway/routers/test_oauth_router.py:1494-1496, 2219-2220`](tests/unit/mcpgateway/routers/test_oauth_router.py:1494)

Three blank lines before `TestResolveTokenTeamsForScopeCheck`; missing blank line before `test_enforce_gateway_access_team_visibility_cache_hit_member_allowed`. Pre-existing false-comparison errors also at line 3690. Run Ruff on changed files before merge.

---

### M-3 — `normalize_token_teams` import path is indirect

**File:** [`mcpgateway/routers/oauth_router.py:30`](mcpgateway/routers/oauth_router.py:30)

```python
from mcpgateway.auth import normalize_token_teams
```

Canonical source is `mcpgateway.auth_context`; `auth.py` re-exports via `__all__`. Both work. Importing direct from `auth_context` would be marginally clearer. No change required.

---

**Gap:** No test for `current_user.is_admin=False` + JWT `is_admin=True` through `_enforce_gateway_access`. Low priority — produces correct deny — but worth documenting.

ja8zyjits (dismissed):
LGTM

ja8zyjits (dismissed):
LGTM

gcgoncalves (dismissed):

mcpgateway/routers/oauth_router.py
```diff
@@ -478,22 +560,6 @@ def _resolve_token_teams_for_scope_check(request: Request, current_user: EmailUs
     return token_teams
```
gcgoncalves, line 560: There's a non-blocking edge case here that may be addressed as a follow up:

If an admin user with a long-living legacy JWT token (thus having `request.state.token_teams` = `None`) get's demoted, this function will still return `None` (free access), even though  the user is not an admin anymore.

To patch this, we'd need to add a JWT revoking step on demoting a user. A local temp fix might also be restructuring the condition to change the response depending on the admin status:

```python
if token_teams is None:
    return None if is_admin else []
return token_teams
```

ja8zyjits: that is already carried out in the condition above
```python
    if is_admin and token_teams is None:    # <-- here itself that is sorted
        return None
    return token_teams  # <--- this is default to []
```

ja8zyjits (approved):
LGTM

PR #6006, opened 2026-07-30
feat(ui-rewrite): (De)Activate prompts button

### LINKED ISSUE #5409: [UI-REWRITE]: Add Activate/Deactivate action to the Prompts view overflow menu
### 🧭 Type of Feature

  ### 🧭 Epic
  **Title:** Prompts view parity with Admin UI
  **Goal:** Allow operators to activate/deactivate a prompt directly from the new `/client` Prompts view, matching the legacy Admin UI behaviour.
  **Why now:** The legacy Admin UI lets operators toggle a prompt's active state. The rewritten UI should reach parity so prompts can be enabled/disabled without editing or deleting them.

> **⚠️ Dependency:** The rewritten Prompts view is currently a stub (`client/src/pages/Prompts.tsx` only renders a heading). This action depends on the Prompts list/card view and details sidepanel (with a per-prompt overflow "⋯" menu) tracked in **#5323**, and should be sequenced after it.

  ---

  ### 🧑🏻‍💻 User Story
  **As a:** gateway operator
  **I want:** an Activate/Deactivate action in the per-prompt overflow menu of the Prompts view
  **So that:** I can disable a prompt (taking it out of availability) and re-enable it without editing or deleting it.

  #### Description
  In the Prompts view, add a per-prompt overflow ("⋯") menu item that toggles the prompt's active state:

  - A prompt is **active** by default.
  - When the prompt is **active**, the menu item reads **Deactivate**.
  - When the prompt is **inactive**, the menu item reads **Activate**.
  - Behaviour must match the legacy Admin UI (deactivating removes the prompt from availability; reactivating restores it).
  - The status indicator on the card (and in any details panel) must reflect the new state after refresh, driven by `prompt.enabled`.

  Suggested menu order: **View Details → Activate/Deactivate** and **Edit/Delete**.

  #### ✅ Acceptance Criteria
  ```gherkin
  Scenario: Deactivate an active prompt
    Given a prompt that is active (enabled)
    And the prompt overflow menu is open
    When I click "Deactivate"
    Then the prompt is deactivated via the API
    And the status reflects "Inactive"
    And the menu item now reads "Activate"
    And a success toast is shown

  Scenario: Activate an inactive prompt
    Given a prompt that is inactive
    And the prompt overflow menu is open
    When I click "Activate"
    Then the prompt is activated via the API
    And the status reflects "Active"
    And the menu item now reads "Deactivate"
    And a success toast is shown

  Scenario: Failure handling
    Given the activate/deactivate request fails
    Then an error toast is shown
    And the prompt's displayed status is unchanged
  ```

  ---

  ### 🛠️ Implementation Notes

  **Frontend**
  - `client/src/pages/Prompts.tsx` — once the Prompts list/card view exists, add a `DropdownMenuItem` whose label toggles on `prompt.enabled` (`Deactivate` when enabled, `Activate` when disabled); add an `onTogglePrompt` prop/handler that calls the API, shows success/error toasts, and refreshes the list. Drive the status indicator from `prompt.enabled`.
  - `client/src/api/prompts.ts` — create this service (does not exist yet) and add a `setState(id, activate)` method alongside the standard CRUD methods.

  **API**
  - Canonical endpoint: `POST /prompts/{prompt_id}/state?activate=true|false` (requires `prompts.update` permission) — see `mcpgateway/main.py:6299`.

  **i18n**
  - Add strings for the new actions (e.g. `prompts.card.activate`, `prompts.card.deactivate`) plus success/error toast messages, across `en-US`, `es-ES`, `pt-BR` (`client/src/i18n/locales/*/prompts.json`).

  **Tests**
  - Unit tests for the prompt card (correct label based on `enabled`, handler invocation) and the toggle flow (success + failure toasts).

gcgoncalves:
# Pull Request

## 🔗 Related Issue

Closes #5409

---

## 📝 Summary

Add Activate/Deactivate button to prompts, with optimistic update

---

## 📏 Reviewability

---

## 🏷️ Type of Change

- [ ] Bug fix
- [X] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

### DIFF
--- a/client/src/components/prompts/PromptDefinitionTable.tsx
+++ b/client/src/components/prompts/PromptDefinitionTable.tsx
@@ -26,6 +26,7 @@ export interface PromptDefinitionTableProps {
   onSelectPrompt: (prompt: NonNullable<PromptRead>) => void;
   onEdit?: (prompt: NonNullable<PromptRead>) => void;
   onDelete?: (prompt: NonNullable<PromptRead>) => void;
+  onTogglePrompt?: (id: string, currentState: boolean) => void;
 }
 
 /**
@@ -45,6 +46,7 @@ export function PromptDefinitionTable({
   onSelectPrompt,
   onEdit,
   onDelete,
+  onTogglePrompt,
 }: PromptDefinitionTableProps) {
   const intl = useIntl();
 
@@ -108,7 +110,7 @@ export function PromptDefinitionTable({
             </TableCell>
 
             <TableCell className="px-4 py-3 text-center">
-              {(onEdit || onDelete) && (
+              {(onEdit || onDelete || onTogglePrompt) && (
                 <DropdownMenu>
                   <DropdownMenuTrigger asChild>
                     <Button
@@ -138,6 +140,30 @@ export function PromptDefinitionTable({
                         {intl.formatMessage({ id: "prompts.details.action.edit" })}
                       </DropdownMenuItem>
                     )}
+                    {onTogglePrompt &&
+                      (() => {
+                        const enabled = prompt.enabled ?? true;
+                        return (
+                          <DropdownMenuItem
+                            onClick={(e) => {
+                              e.stopPropagation();
+                              onTogglePrompt(prompt.id, enabled);
+                            }}
+                            aria-label={intl.formatMessage(
+                              {
+                                id: enabled
+                                  ? "prompts.details.deactivateAriaLabel"
+                                  : "prompts.details.activateAriaLabel",
+                              },
+                              { name: prompt.displayName || prompt.name },
+                            )}
+                          >
+                            {enabled
+                              ? intl.formatMessage({ id: "prompts.details.action.deactivate" })
+                              : intl.formatMessage({ id: "prompts.details.action.activate" })}
+                          </DropdownMenuItem>
+                        );
+                      })()}
                     {onDelete && (
                       <DropdownMenuItem
                         onClick={(e) => {
--- a/client/src/pages/Prompts.tsx
+++ b/client/src/pages/Prompts.tsx
@@ -78,9 +78,11 @@ function buildPromptGroups(
 function PromptGroupCard({
   group,
   onViewDetails,
+  onTogglePrompt,
 }: {
   group: PromptGroup<NonNullable<PromptRead>>;
   onViewDetails: (group: PromptGroup<NonNullable<PromptRead>>) => void;
+  onTogglePrompt?: (id: string, currentState: boolean) => void;
 }) {
   const intl = useIntl();
   const visiblePrompts = group.prompts.slice(0, MAX_VISIBLE_PROMPTS);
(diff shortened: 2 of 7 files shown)


client/src/pages/Prompts.tsx
```diff
@@ -246,6 +268,53 @@ export function Prompts() {
...
+        await promptsApi.setState(id, newState);
+        toast.success(
+          intl.formatMessage(
+            {
+              id: newState ? "prompts.toast.activateSuccess" : "prompts.toast.deactivateSuccess",
+            },
+            { name: promptName },
+          ),
+        );
+        await refetch();
```
vishu-bh: **Functionally impacting:** Could we separate mutation failure from refresh failure here? `setState()` may succeed and commit server-side, then `refetch()` may fail. Because both calls share this `try`, that case enters the catch block, restores the old local state, and shows a failure toast even though the prompt changed successfully. The UI then contradicts server state. The delete flow below already uses the safer pattern: keep confirmed optimistic state and handle refresh failure separately.

client/src/components/prompts/PromptDefinitionTable.tsx
```diff
@@ -138,6 +140,26 @@ export function PromptDefinitionTable({
...
+                        aria-label={intl.formatMessage(
+                          {
+                            id: prompt.enabled
+                              ? "prompts.details.deactivateAriaLabel"
+                              : "prompts.details.activateAriaLabel",
+                          },
+                          { name: prompt.displayName || prompt.name },
+                        )}
+                      >
+                        {prompt.enabled
```
vishu-bh: **Edge case:** Could we normalize `enabled` once with `const enabled = prompt.enabled ?? true` and use it for both rendering and the callback? Currently an omitted value renders **Activate** because this condition treats `undefined` as false, but the click handler passes `prompt.enabled ?? true`, treats it as active, and sends a deactivate request. Label and resulting action therefore disagree.

vishu-bh (changes requested):
Thanks for adding prompt state controls and keeping the implementation within the shared UI components. Two runtime consistency issues need correction before merge. This PR also adds no frontend coverage, although #5409 explicitly requires label, handler, success, and failure tests; please add positive, negative, and edge-case coverage for this flow. Security review found no authorization, CSRF, injection, or XSS regression—the backend continues to enforce prompts.update and ownership.

client/src/pages/Prompts.tsx
```diff
@@ -122,6 +124,26 @@ function PromptGroupCard({
...
               </DropdownMenuItem>
+              {onTogglePrompt && group.prompts.length === 1 && (
+                <DropdownMenuItem
+                  onClick={() => {
+                    const prompt = group.prompts[0];
+                    onTogglePrompt(prompt.id, prompt.enabled ?? true);
+                  }}
+                  aria-label={intl.formatMessage(
+                    {
+                      id: group.prompts[0].enabled
```
vishu-bh, line 135: Could we normalize the prompt state once here, as done in `PromptDefinitionTable`? When enabled is `undefined`, this condition displays “Activate,” while the click handler uses `prompt.enabled` ?? true and sends a deactivate request. Using `const enabled = prompt.enabled ?? true` for the label, ARIA text, and callback would keep behavior consistent.

client/src/pages/Prompts.tsx
vishu-bh, line 154: This is worth checking might have missed previously

Issue #5409 requires the card’s status indicator to reflect prompt.enabled after toggling. This card currently renders prompt tags but no Active/Inactive state. Could we add the corresponding status indicator, using the normalized enabled value?

client/src/pages/Prompts.tsx
```diff
@@ -246,6 +268,61 @@ export function Prompts() {
     [setPromptsData, intl],
   );
 
+  const handleTogglePrompt = useCallback(
```
vishu-bh, line 271: Could we add unit coverage for this new toggle flow? The PR currently changes no test files, while #5409 explicitly requires label/handler, success, and failure coverage. Useful cases include activate/deactivate requests, optimistic updates, API-error rollback, successful mutation followed by failed refetch, and undefined enabled.

vishu-bh (changes requested):
Refetch failure handling and details-table state normalization are now fixed

Found some small things to check.

gcgoncalves:
Closed in favour of the new repo

Issue #5978, opened 2026-07-29
[BUG]: /admin/llm/* and /v1/admin/llm/* diverge on CSRF validation immediately after login


msureshkumar88:
### Describe the bug

`/admin/llm/*` (legacy, unprefixed mount) and `/v1/admin/llm/*` (versioned mount) both serve the same `llm_admin_router`, guarded by the same `enforce_admin_csrf` dependency. In practice they enforce different effective CSRF requirements for a short but real window in every admin session: immediately after `/admin/login`, before the first dashboard page load.

### Root cause

- `admin_login_handler` sets the CSRF cookie via `_set_admin_csrf_cookie(request, response)` with **no** `user_id`/`session_id` (`mcpgateway/admin.py`, `admin_login_handler`). Without those, `_set_admin_csrf_cookie` falls back to an opaque `secrets.token_urlsafe(32)` value — not an HMAC token.
- Only the dashboard route (`admin_ui()`) rotates the cookie to its real HMAC-bound value, by calling `_set_admin_csrf_cookie(request, response, user_id=admin_email, session_id=str(jti))`.
- `enforce_admin_csrf` validates with a plain `secrets.compare_digest(header, cookie)` — format-agnostic, so the opaque token satisfies it.
- `CSRFMiddleware` validates via `CSRFService.validate_csrf_token()`, which requires a 64-char hex HMAC token (`mcpgateway/services/csrf_service.py`) — the opaque `token_urlsafe(32)` value (~43 chars, not hex) fails this check outright.
- `/admin` is in `csrf_exempt_paths` and is prefix-matched against the raw request path (`mcpgateway/middleware/csrf_middleware.py`) — this exempts `/admin/llm/*` from `CSRFMiddleware` entirely, but **does not** exempt `/v1/admin/llm/*`, since the prefix match is on `/admin`, not `/v1/admin`. `mcpgateway/api/v1/__init__.py` mounts the same `llm_admin_router` at both `/admin/llm` (legacy router, unprefixed) and `/v1/admin/llm` (versioned router) via `_assemble_routers`.

### Net effect

For the exact same session, same CSRF cookie, same `X-CSRF-Token` header, same `Origin`:
- A write to `/admin/llm/*` **succeeds** (passes `enforce_admin_csrf`'s plain double-submit check).
[...]

PR #6008, opened 2026-07-30
refactor(middleware): extract auth-context middleware registration into testable submodule

jonpspri:
## Summary

The CSRF regression tests in `test_admin_csrf_binding.py` (issue #5739) passed in isolation but three of them failed under the full `make test` suite. Root cause: middleware registration in `mcpgateway/main.py` is import-time and settings-dependent — `AuthContextMiddleware` is only registered when `MCPGATEWAY_ADMIN_API_ENABLED` (or security logging / SIEM / password enforcement) is true at first import. Under xdist, workers import `mcpgateway.main` during collection while the session bootstrap still has the admin API disabled, so the shared app singleton was built without `AuthContextMiddleware`; the session fixture's later env flip could mount admin routes but could not retrofit middleware. Without it, `request.state.user` was never populated and email-bound CSRF tokens failed HMAC validation (`CSRF_TOKEN_INVALID`).

## Changes

- **`mcpgateway/middleware/auth_context_stack.py` (new)**: the CSRF / PasswordChangeEnforcement / AuthContext registration block extracted from `main.py` as `register_auth_context_middleware(app)`. Registration order and import-time behavior are unchanged — `main.py` calls it at the same program point.
- **`test_admin_csrf_binding.py`**: deleted the five full-stack TestClient tests that depended on the assembled app singleton; converted the `enforce_admin_csrf` block to direct-call unit tests with a mocked `Request`; added hermetic ordering tests that exercise the extracted function against a bare `FastAPI()` app. The module no longer imports `mcpgateway.main` at all, so it cannot be affected by collection-order pollution again. 19/19 tests pass identically in isolation and alongside modules that import `mcpgateway.main` early.

End-to-end CSRF coverage (real login → dashboard cookie rotation → write) remains with the Playwright security suite (`tests/playwright/security/`), which already exercises the login + bound-CSRF + write mechanism against a real gateway.

### DIFF
--- a/mcpgateway/main.py
+++ b/mcpgateway/main.py
@@ -105,6 +105,7 @@ from mcpgateway.db import refresh_slugs_on_startup, SessionLocal
 from mcpgateway.db import Tool as DbTool
 from mcpgateway.deprecations import RUST_MCP_RUNTIME_DEPRECATION_MESSAGE, VALIDATION_MIDDLEWARE_DEPRECATION_MESSAGE
 from mcpgateway.handlers.sampling import SamplingError, SamplingHandler
+from mcpgateway.middleware.auth_context_stack import register_auth_context_middleware
 from mcpgateway.middleware.client_disconnect import ClientDisconnectMiddleware
 from mcpgateway.middleware.compression import SSEAwareCompressMiddleware
 from mcpgateway.middleware.correlation_id import CorrelationIDMiddleware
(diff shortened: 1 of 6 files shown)


madhu-mohan-jaishankar (commented):
LGTM.

One thing I'd like addressed before or shortly after merge:

**The dropped pin on #5978.** Twelve tests were removed and eight converted, but four were
dropped outright — including
`test_admin_llm_write_diverges_between_mounts_immediately_after_login`, which pinned the
still-open #5978 divergence (legacy `/admin/llm/*` accepting an opaque pre-rotation token
that `/v1/admin/llm/*` rejects). After this merges, nothing in the repo detects whether
that gets fixed or gets worse. Coverage gates won't catch it either — the target code is still executed by other suites, so line coverage stays flat while the assertion disappears.

The extracted helper is actually the right tool to rebuild these hermetically: construct a fresh `FastAPI()` under `_shadow_settings(...)`, call `register_auth_context_middleware()`,
`include_router()` the admin + llm routers, and drive a `TestClient` against *that* app
instead of the `mcpgateway.main` singleton. That keeps the import-order independence this PR is after without losing the end-to-end assertion. A follow-up issue is fine if you'd
rather not grow this PR.

Two smaller notes, non-blocking:

- Nothing asserts that `main.py` still calls the helper, or where. Deleting
  `register_auth_context_middleware(app)` from `main.py` entirely passes all 19 tests in
  the file. A cheap `inspect.getsource` guard would close that — the file already uses that
  pattern for `admin.py` at line 258.
- `test_enforce_admin_csrf_rejects_wrong_origin` shadows `allowed_origins=[]`, so the
  `for allowed in settings.allowed_origins:` loop has zero iterations and the fail-closed
  `*` / `null` exclusion never runs — the branch the docstring says it exercises. A
  non-empty non-matching list (including a wildcard) would make it real.

madhu-mohan-jaishankar (approved):
LGTM

brian-hussey (approved):
LGTM.

The rename of "csrf_token" to "mcpgateway_csrf_token" is in line with the default value from settings (config.py), and it's redundant to have the default set here as it's already trying to get the default, but for this case, even having the variable defined as DEFAULT_CSRF_TOKEN_NAME or something means we wouldn't have to change it in multiple places; but that's for future reference, not blocking at the moment.

jonpspri:
Follow-on issues to be filed.

jonpspri:
Thanks for the thorough review, @madhu-mohan-jaishankar — all three points are now tracked as follow-ups rather than growing this PR:

- **Dropped #5978 pin** → #6024. We're taking your hermetic-rebuild goal one step further: the pin will come back as an integration test under `tests/live_gateway/` driving a real gateway (`make testing-up`), which sidesteps the import-order fragility that made the original TestClient version flaky under xdist. The issue captures the full scenario plus a couple of stack-specific gotchas (Origin headers vs. `APP_DOMAIN`).
- **No guard that `main.py` calls the helper** → #6025 (`inspect.getsource` guard, same pattern as the existing `admin.py` one).
- **`test_enforce_admin_csrf_rejects_wrong_origin` shadowing `allowed_origins=[]`** → #6026 (non-empty list including a wildcard, so the fail-closed exclusion branch actually runs).

Issue #5841, opened 2026-07-24
[Design] Card with glanceable status & sparklines


a-effort:
[Card](https://www.figma.com/design/H2Dr9QPvRVxCe92V8URvGT/Refiring?m=auto&node-id=4947-31295&t=9La3DwrcCLRbvwvw-1) with glanceable status & sparklines

a-effort:
Current ux until backend work done: https://github.com/IBM/mcp-context-forge/commit/1087a991df93925fd660501f160ec0d1f4fdba7f

a-effort:
Tagged `blocked`. Backend blocker: #5945 (durable `/app/observability/metrics/*` endpoints). Current UX runs on a placeholder until that lands. cc @jonpspri

a-effort:
Closing as this is just waiting on the backend PR to be approved: https://github.com/IBM/mcp-context-forge/pull/6009

Issue #5533, opened 2026-07-06
O11Y chart spike #2


a-effort:
As a follow up to this [initial spike](https://github.com/IBM/mcp-context-forge/commit/fda11f7456138d6b95eee3483bbe35b8db2cb2fd), explore adding another card to help users have a glanceable view of tokens used. 

Inspiration: [demo](https://youtu.be/KtmaWUVdnx4?si=tvy4s--58E9DTU5W&t=462) of [symphony ](https://github.com/openai/symphony/blob/main/SPEC.md)

<img width="1964" height="1320" alt="Image" src="https://github.com/user-attachments/assets/96a5e4a1-25b6-4003-87fa-a86fc1dd2296" />

a-effort:
Moving back to "to do" as further exploration can happen later. Initial version will be shipped as part of https://github.com/IBM/mcp-context-forge/pull/6090 as the default home status card.

a-effort:
Closing this spike for now.

Issue #5746, opened 2026-07-21
[BUG]: OBSERVABILITY_TRACE_RETENTION_DAYS and OBSERVABILITY_MAX_TRACES are documented and shipped but never enforced — observability traces grow unbounded


Shyrka973:
### 🐞 Bug Summary

`OBSERVABILITY_TRACE_RETENTION_DAYS` (default `7`) and `OBSERVABILITY_MAX_TRACES` (default `100000`) are declared in `mcpgateway/config.py`, documented across six pages under `docs/`, listed in `.env.example`, and exposed in the official Helm chart — but **neither is read anywhere in the codebase**.

As a result, `observability_traces` and `observability_spans` are never pruned by any background task and grow without bound until the database volume fills up.

This is arguably worse than having no setting at all: a missing knob is noticeable, whereas a documented-but-dead knob gives operators a **false assurance** that retention is bounded. `.env.example` states the guarantee explicitly:

```dotenv
# Maximum number of traces to retain (prevents unbounded growth)
# OBSERVABILITY_MAX_TRACES=100000
```

---

### 🧩 Affected Component

- [x] `mcpgateway` - API
- [ ] `mcpgateway` - UI (admin panel)
- [ ] `mcpgateway.wrapper` - stdio wrapper
- [ ] Federation or Transports
- [ ] CLI, Makefiles, or shell scripts
- [ ] Container setup (Docker/Podman/Compose)
- [x] Other — observability background services, docs, `.env.example`, and `charts/mcp-stack`

---

### 🔁 Steps to Reproduce

1. Deploy with `OBSERVABILITY_ENABLED=true` (all other observability settings at their defaults).
2. Optionally set `OBSERVABILITY_TRACE_RETENTION_DAYS=1` and `OBSERVABILITY_MAX_TRACES=1000` to make the effect visible sooner.
3. Generate traffic for longer than the retention window.
4. Query the database:

```sql
SELECT min(start_time), count(*) FROM observability_traces;
```

Rows older than the retention window are still present, and the row count exceeds `OBSERVABILITY_MAX_TRACES`.

---

### 🤔 Expected Behavior

Traces older than `OBSERVABILITY_TRACE_RETENTION_DAYS` should be deleted automatically, and the total number of retained traces should be capped at `OBSERVABILITY_MAX_TRACES` — as stated in the documentation and in `.env.example`.

Observed instead: retention is never enforced; the only way to prune is to call the manual endpoint yourself.

---

[...]

PR #6009, opened 2026-07-30
feat(api): durable observability metrics endpoints for home dashboard

### LINKED ISSUE #5945: [FEATURE][API][UI-REWRITE]: Durable metrics endpoints for home
## Context
Durable metrics endpoints for home dashboard charts (executions + latency percentiles)

The V2 home (#5532) system status [card](https://www.figma.com/design/H2Dr9QPvRVxCe92V8URvGT/Refiring?m=auto&node-id=4956-32364&t=MecGWFgOUaNktFR3-1) (#5841) shows executions and p50/p95/p99 latency timeseries. A prior spike (branch `5533-o11y-chart-spike-2`) implemented `GET /admin/observability/metrics/{timeseries,percentiles}`, but on the admin router (`MCPGATEWAY_ADMIN_API_ENABLED`), the surface the React UI rewrite is replacing. This re-homes those two read endpoints on a durable non-admin router so #5841 does not depend on the deprecating admin API.

Originally scoped against the `/app` React-UI router on `epic/ui-rewrite`. The work was retargeted to `main`, where there is no `/app` router and the React client lives in `contextforge-org/contextforge-web-ui`, so the endpoints landed on a new `observability_metrics_router` mounted unconditionally under `/observability/metrics` (and the `/v1` alias).

## Endpoints

- `GET /observability/metrics/timeseries?hours=&interval_minutes=` → `{ buckets: string[], values: number[] }`
- `GET /observability/metrics/percentiles?hours=&interval_minutes=` → `{ buckets: string[], p50: number[], p95: number[], p99: number[] }`

Both are also reachable at `/v1/observability/metrics/*`.

### Backing store

Per [#5945 (comment)](https://github.com/IBM/mcp-context-forge/issues/5945#issuecomment-5413030399), these endpoints read from the existing `observability_traces` table for now. The finite pre-aggregated UX-card store is tracked in #6387 and is not a prerequisite here.

Two properties of that store are being handled outside this issue:

- **Reaping.** There is none today. `observability_trace_retention_days` and `observability_max_traces` are declared in `config.py` and read nowhere, and `MetricsCleanupService` does not cover the `observability_*` tables. Tracked in #5746.
- **Query cost.** The SQLite fallback loaded the full query window into Python. Tracked in #6286, fixed by PR #6390 which moves the aggregation into SQL.

### Endpoint wiring (shipped via PR #6293)

- Reuses the spike's `observability_service` aggregation logic and response shapes (`TimeseriesResponse`, `PercentilesResponse`); no new aggregation math.
- Standard `get_current_user_with_permissions` stack, which accepts first-party SPA cookie fetches; CSRF applies by method, so GETs pass.
- Permission: `metrics:read` with `allow_admin_bypass=False`, granted by default to `platform_viewer` only.
- Token-scoping entry maps both paths to `metrics:read`; the rest of `/observability/*` stays default-denied for scoped tokens.

## Dependency / empty state

Requires observability store to be populated. When data is absent, the endpoints return **empty series** (not an error), so the UI shows an "enable observability" empty state.

## Out of scope

Token-spend / cost estimation stays with the #5533 spike.

## Acceptance

- [x] Both endpoints served on a durable non-admin router, not `admin_router`.
- [x] `metrics:read` deny-path test.
- [x] Observability-disabled / empty store returns empty series (not an error).
- [x] Client wiring moved out of scope for this repo. Pointing `metrics.ts` / `useMetrics` at `/observability/metrics/*` in `contextforge-org/contextforge-web-ui` is tracked in #6430, which also covers the changed response contract and the recharts bundling question.

Blocks #5841. Sub-issue of #5532.

Altamimi-Dev:
Re-homes the #5533 spike's two observability read endpoints so the home system status card (#5841) does not depend on the deprecating admin API (`MCPGATEWAY_ADMIN_API_ENABLED` surface).

Originally raised against `epic/ui-rewrite` on the `/app` router; retargeted to `main` per maintainer request, since the change is Python-only and the React app is moving to `contextforge-org/contextforge-web-ui`. There is no `/app` router on `main`, so the endpoints now live on a new `observability_metrics_router` in `mcpgateway/routers/observability.py`:

- `GET /v1/observability/metrics/timeseries?hours=&interval_minutes=` → `{ buckets, values }` (executions per bucket)
- `GET /v1/observability/metrics/percentiles?hours=&interval_minutes=` → `{ buckets, p50, p95, p99 }` (latency ms)

Both are also reachable at the unversioned legacy alias (`/observability/metrics/*`), since `_assemble_routers` feeds the `/v1` and legacy mounts alike.

That router is mounted **unconditionally**, unlike the existing flag-gated observability router (Group C), which is only mounted when `OBSERVABILITY_ENABLED=true`. This preserves the reviewed contract: with observability disabled the endpoints return empty series (200) rather than 404, so the UI can render its "enable observability" empty state. The nine existing observability handlers and their mount behavior are untouched.

Aggregation math is the spike's, copied into `ObservabilityService` query methods (PostgreSQL `percentile_cont` with a Python fallback for SQLite). The `admin.py` copies stay for now and are removed when those endpoints are retired. Auth follows the standard `get_current_user_with_permissions` stack, which accepts first-party SPA cookie fetches, and CSRF applies by method (GETs pass).

**Permission tier:** entry is gated by `metrics:read` with `allow_admin_bypass=False`, not `admin.system_config`. That is deliberate: the home dashboard these endpoints feed has to work for non-admin users while the admin API deprecates, so an admin-only tier would defeat the point of moving off `/admin`. Turning admin bypass off keeps the grant explicit — a platform admin still needs `metrics:read` on paper rather than reaching the data implicitly.

**Layer-1 token scoping:** on `main` these paths join the scoped-token API surface, where unmapped protected paths are default-denied. Without an entry, an API token scoped to `metrics:read` could never reach them (the bug class fixed in #5973/#5974). A narrow `_PERMISSION_PATTERNS` entry maps `GET /observability/metrics/{timeseries,percentiles}` to `metrics:read`; the rest of `/observability/*` stays default-denied for scoped tokens.

The `metrics:read` default grant goes to `platform_viewer` only — not `viewer`/`developer`/`team_admin` — because the aggregation has no team filter, so a team-scoped grant would expose platform-wide volume and latency across tenants. The migration is scope-filtered accordingly and has been re-parented onto `main`'s current head.

Token-spend / cost estimation stays with the #5533 spike, per the issue's scope.

Closes #5945


a-effort:
Thank again for your contributions that help unblock `ui-rewrite` work! 🙌 

I've requested review from backend devs on this and your other PRs. 

--> LLM assisted feedback: 

Moving the home card off the old admin API is the right call. Testing with real SQLite data instead of mocks is an improvement, and the Python percentile calculation matches PostgreSQL's results exactly.

One item to resolve before merge:

Bucket timestamps aren't consistently in UTC across backends. On the SQLite path, timestamps are built with tz=timezone.utc, so they come out as ...T12:00:00+00:00. On the Postgres path, timestamps come back tagged with the database session's timezone instead, so the same moment in time can print as something like ...T07:00:00-05:00. Same instant, different text. That means SQLite (the default) and Postgres (production) can describe the same bucket differently, and the home card reads both.

The existing trace endpoints already handle this correctly for the same underlying data (observability.py:605), by storing everything in UTC. Making the Postgres buckets do the same fixes it:

```
buckets.append(row.bucket.astimezone(timezone.utc).isoformat() if row.bucket else "")
```

Altamimi-Dev:
Most welcome :)

Both Postgres helpers now serialize via `row.bucket.astimezone(timezone.utc).isoformat()`, so `TO_TIMESTAMP` results no longer leak the DB session timezone and match the Python/SQLite path exactly.

Added `test_postgres_buckets_normalized_to_utc` covering both helpers with a non-UTC tz-aware row, since the Postgres branch never runs under the SQLite test DB.

madhu-mohan-jaishankar (changes requested):
Nice re-homing — the aggregation math holds up. I checked bucket alignment between the Postgres and Python paths (they agree), both queries are fully parameterized, empty results are handled in all four helpers, `percentile_cont` is guarded for empty and `n==1`, and the SQLite tz round-trip is correct. Ran the new tests locally: 6 passed, ruff clean.

One thing I'd like resolved before merge, then some follow-ups.

**Permission tier on the two endpoints** — `mcpgateway/routers/app.py:248` and `:285` use `@require_permission("metrics:read")` with the default `allow_admin_bypass=True`. Every other reader of `observability_traces` requires `admin.system_config` with `allow_admin_bypass=False`: `admin.py:18883`, `admin.py:19045`, and all eight handlers in `routers/observability.py`.

That matters because the observability subsystem has no team scoping — the new helpers filter on `start_time` and nothing else. So `metrics:read`, which per `routers/log_search.py:707` has been scoped to the separate performance-metrics table, now reads global request volume and latency across every team and tenant.

This may be exactly what you intend given the epic is moving off the admin API. If so, could you say so explicitly in the PR description and drop `allow_admin_bypass` to `False`? It's a cheap change now and an awkward one once the home dashboard ships against it — walking the permission back up would break anyone granted access in the interim.

## Follow-ups (fine to land as-is)

**The UTC fix should also go to `admin.py`.** Your last commit fixed `row.bucket.astimezone(timezone.utc).isoformat()` in the new copy, but `admin.py:18964` and `:19129` still do a bare `row.bucket.isoformat()`. Those are the ones the admin observability chart actually calls today (`templates/observability_metrics.html:115,123`), so on a Postgres deployment with a non-UTC session `TimeZone` the live chart still emits local-offset timestamps. You found a real bug — worth fixing where it's currently biting, either here or in a separate PR.

**The PR description says the math was "moved" into `ObservabilityService`, but it was copied** — `admin.py:18882-19199` still has all four functions byte-for-byte. Totally reasonable as a transitional state while the admin surface deprecates; just worth wording it as "copied, admin copies to be removed when those endpoints go" so the cleanup doesn't get lost.

**Tests don't go over HTTP.** `test_app_metrics.py:121-158` calls the handler coroutines directly with a hand-built user dict, so nothing asserts the 401 path, that `Depends(get_current_user_with_permissions)` is wired at all (deleting it from the signature wouldn't fail a test), or the `Query(ge=1, le=168)` bounds. Deny-path coverage has insufficient-permissions and feature-disabled but not unauthenticated. A single `TestClient` test for the 401 would close the gap.

**Postgres path isn't really exercised.** `test_app_metrics.py:227-241` mocks `db.execute` and feeds one hand-built row, so the raw SQL never runs — a typo in the `EXTRACT(EPOCH ...)` bucketing or a wrong percentile fraction would pass. It also asserts `p50` and stops; `p95`/`p99` go into the mock unchecked. Ideally a Postgres-backed test asserting both paths produce identical buckets for identical input, since that's the property the last commit claims.

**Minor:**
- `observability_service.py:1852` `row.total or 0` and `:1924-1926` `if row.p50 else 0` — falsy checks where `is not None` is meant. No observable defect today, but they read as null-guards.
- `app.py:276` / `:313` — `cutoff_time` is an arbitrary instant while buckets align to epoch multiples, so the oldest bucket is always partial and renders as a low first point.
- `observability_service.py:1868` / `:1943` — the Python fallback `.all()`s the whole window, and that path *is* the default SQLite deployment. `hours=168` pulls a week of traces per request, with no rate limit while the dashboard polls.
- Consider adding `("mcpgateway/routers/app.py", "metrics:read")` to `ROUTER_PERMISSIONS` in `tests/security/test_rbac_decorator_coverage.py:60`.

Altamimi-Dev:
Thanks for the thorough pass — the bypass point was the right catch, and the smaller items were all cheap enough to fold in rather than defer.

Pushed in `9333ffa4e`:

- **`allow_admin_bypass=False` on both metrics endpoints.** They were on the decorator default, so an admin reached them without holding `metrics:read`. Now the grant has to be explicit, matching how `routers/observability.py` gates the `admin.system_config` readers of the same table.
- **PR body updated** to state the tier reasoning: `metrics:read` (not `admin.system_config`) is intentional — the home dashboard these feed must work for non-admin users while the admin API deprecates. Also corrected "moved into `ObservabilityService`" to "copied"; the `admin.py` copies go away when those endpoints retire.
- **`admin.py` UTC fix.** The two remaining bare `row.bucket.isoformat()` sites now `astimezone(timezone.utc)` like the service copies, so a Postgres session timezone can't make the two paths disagree.
- **`is not None` guards** in the new query helpers for `total` / `p50` / `p95` / `p99`, so a genuine zero isn't read as a missing aggregate.
- **New tests:** an unauthenticated `TestClient` request to both endpoints asserting 401 over real HTTP (the existing 403 test calls the handler directly), and `("mcpgateway/routers/app.py", "metrics:read")` added to `ROUTER_PERMISSIONS` in `tests/security/test_rbac_decorator_coverage.py` so the decorator can't be dropped silently.

A few I'd rather carry as follow-ups than force into this PR:

- **Postgres/SQLite dual-path test.** There's no Postgres in the unit environment, so a real dual-path assertion belongs with the gateway-stack e2e rather than a mock that would only re-assert the mock. The UTC normalization is covered by `test_postgres_buckets_normalized_to_utc` against a tz-tagged row.
- **Partial first bucket alignment.** Real, but it's a shared behaviour of the spike's math and the admin copies; worth fixing in one place once, not diverging the two implementations here.
- **Rate limiting on the fallback path.** Agreed the SQLite fallback is the expensive one; it deserves its own change with a limit that's actually tuned rather than a number picked in this PR.

Happy to pull any of those forward if you'd rather they block.

Altamimi-Dev:
One gap from your review I under-delivered on, now closed in `6b527d0cf`.

You listed three things the direct-call tests couldn't assert: the 401 path, whether `Depends(get_current_user_with_permissions)` is wired at all, and the `Query(ge=1, le=168)` bounds. My earlier commit covered the first two but not the third — the bounds were still unpinned, so widening or dropping either constraint wouldn't have failed anything.

Added a `TestClient` parametrization over both endpoints hitting each boundary: `hours` at 0 and 169, `interval_minutes` at 4 and 1441, all expecting 422. Checked it discriminates rather than passing incidentally — in-range values (`hours` 1/24/168, `interval_minutes` 5/1440) return 403 from the permission gate, not 422, so the 422s are genuinely coming from the bounds.

That matters a bit more than a typical validation test here, since the SQLite fallback `.all()`s the whole window: `hours` is the knob that decides how much a single unauthenticated-adjacent request can pull. Still leaves the rate-limiting point open as a follow-up, as noted above.

madhu-mohan-jaishankar (dismissed):
Approving with a followup. @Altamimi-Dev @gcgoncalves 

`metrics:read` isn't granted by any seeded role
yet, so non-admin users will get a 403 until we add it.

Lang-Akshay (changes requested):
Thanks for the PR @Altamimi-Dev 
## Please address this blocking Changes

| 1 | Security — Cross-tenant disclosure | `mcpgateway/services/observability_service.py` | 1835–36, 1904–05 | `metrics:read` granted to team-scoped roles (`team_admin`, `developer`, `viewer`, `platform_viewer`), but aggregation SQL has no team/tenant filter — any team member sees platform-wide execution volume and latency across all tenants. `ObservabilityTrace.user_email` exists but is not used in queries. CWE-200/CWE-284. | Either restrict `metrics:read` to `platform_admin`/`platform_viewer` only (global roles), or add server-derived team attribution to all aggregation queries so a team-scoped caller only sees traces matching their team membership. |
| 2 | Security — Migration scope ambiguity | `mcpgateway/alembic/versions/9935d863930b_add_metrics_read_to_default_roles.py` | 81 | `_update_role_permissions()` selects `WHERE name = :name LIMIT 1` without a scope filter. The Role model's partial unique index is `(name, scope, is_active)`; `RoleService.get_role_by_name()` always takes a scope argument. In any deployment with a custom team-scoped role sharing a name with a built-in (e.g., a custom `developer` role), the migration silently updates the wrong row, leaving the intended default role without `metrics:read`. Equally breaks downgrade. CWE-863/CWE-732. | Mirror `1a02944`'s pattern: add `AND scope = :scope` (and optionally `AND is_system_role = true`) to the SELECT; pass the system-role scope when calling `_update_role_permissions`. |

Altamimi-Dev:
Both addressed in `0f57be7bf`:

1. **Cross-tenant disclosure** — narrowed `metrics:read` to `platform_viewer` only. Every non-admin user holds it globally, so the dashboard fix still reaches everyone, but team-scoped roles (`team_admin`/`developer`/`viewer`) no longer get a permission whose aggregation queries have no tenant filter.
2. **Migration scope ambiguity** — role lookup now filters `WHERE name = :name AND scope = :scope`, mirroring `1a02944e2671`, so it can't hit a same-named custom team-scoped role instead of the intended global default.

Verified: single alembic head, upgrade/downgrade/upgrade roundtrip, full test suite green (matrix, bootstrap, app_metrics, rbac_decorator_coverage).

gcgoncalves:
@Altamimi-Dev As this PR touches only python files, please raise it agains the main branch. All the React app work is being moved to a separate repository: https://github.com/contextforge-org/contextforge-web-ui

vishu-bh:
@Altamimi-Dev since these are python changes. Can this PR be targeting to main instead of epic/ui-rewrite. 

Appreciate your help!!

a-effort:
@jonpspri ,  it looks like both blockers are fixed and this just needs re-review.

Status summary: 
@Lang-Akshay 's review was against commit 70b4f0c1, which is now stale. Commit 0f57be7b fixed both requested changes: only platform_viewer (global scope) now gets metrics:read, and the migration adds AND scope = :scope. Both fixes are in the current PR head; no re-review requested since.

TL;DR:
Alternative considered: team-scoped filtering instead of restricting the permission. PR #5973 has a precedent for this (AuditTrail.team_id.in_(token_teams)), but ObservabilityTrace/ObservabilitySpan have no team_id column, only user_email... that route needs a schema migration plus a backfill plan for historical traces. Restricting to global roles avoids that and matches how admin.py/routers/observability.py already gate system-wide observability (admin.system_config, allow_admin_bypass=False).

Home V2 (#6090, merged) backs this up too: useMiniCardStatuses.ts only fetches /version when hasPermission("admin.system_config") and treats a 403 as "unknown" rather than an error, so non-admin users still get a working home page. Whatever UI eventually wires up #6009's endpoints should follow the same pattern.

Altamimi-Dev:
@gcgoncalves @vishu-bh — thanks for the steer, that makes sense: this change is Python-only, so it doesn't need to ride the UI epic. Retargeted to `main` and rebuilt the branch on top of it.

**What the re-home involved**

`main` has no `/app` router, so the two endpoints moved onto `mcpgateway/routers/observability.py` as a second, always-mounted router:

- `/v1/observability/metrics/timeseries` and `/v1/observability/metrics/percentiles` (unversioned legacy aliases come along automatically via `_assemble_routers`)
- It's mounted unconditionally rather than folded into the existing observability router, which is only mounted when `OBSERVABILITY_ENABLED=true`. Folding them in would have turned the reviewed "disabled → empty series (200)" contract into a 404 and broken the UI's enable-observability empty state. The nine existing handlers and their gating are untouched.
- The migration was re-parented onto `main`'s current head (`e4f5a6b7c8d9`); `alembic heads` shows a single head.
- Added one narrow Layer-1 `_PERMISSION_PATTERNS` entry mapping `GET /observability/metrics/{timeseries,percentiles}` → `metrics:read`. On the epic these lived behind the cookie SPA surface; on `main` they join the scoped-token surface where unmapped paths are default-denied, so without it a token scoped to `metrics:read` could never reach them (same class as #5973/#5974). The rest of `/observability/*` stays default-denied.

Everything already reviewed carries over unchanged: the `metrics:read` + `allow_admin_bypass=False` tier, the `platform_viewer`-only default grant with the scope-filtered migration, the UTC bucket serialization on the Postgres path (including the two `admin.py` copies), and the bounded `hours`/`interval_minutes` windows.

**Verification**

- Targeted suites green: new `tests/unit/mcpgateway/routers/test_observability_metrics.py`, plus token-scoping, RBAC permission-matrix, and the RBAC decorator-coverage security test (which now also pins `observability.py` → `metrics:read`).
- Alembic upgrade → downgrade → upgrade roundtrip clean on a scratch SQLite DB; single head.
- Live smoke against a running gateway with `OBSERVABILITY_ENABLED=false`: 401 unauthenticated, 200 `{"buckets":[],"values":[]}` authenticated, 422 on out-of-range `hours`, legacy unversioned alias works, and `/v1/observability/traces` still 404s — confirming the metrics router is mounted while the gated one isn't.
- Full suite: 21870 passed, 0 failed. `make ruff` / `make interrogate` / pre-commit clean on the changed files.

One housekeeping note: since three of the five original commits touched the `/app` router that doesn't exist on `main`, I squashed the branch into a single commit on top of `main` rather than replaying per-commit conflicts. The review history here preserves the trail.

msureshkumar88 (changes requested):
Thanks for putting this together — re-homing the metrics endpoints onto the always-mounted `/v1`/legacy router is the right fix for #5945, and the scope is well contained (backend-only, `Closes #5945` referenced, single Alembic head, 132 relevant tests green).

## Blocking

**1. Naive-datetime handling shifts timestamps instead of labeling them UTC**
`mcpgateway/services/observability_service.py:1851` and `mcpgateway/admin.py:18943` both use `row.bucket.astimezone(timezone.utc)`. If `row.bucket` is naive (as the existing test mocks already construct it, e.g. `datetime(2025, 1, 1, 12, 0, 0)` with no tzinfo), `astimezone()` interprets it as local time and converts — silently shifting the bucket by the host's UTC offset rather than labeling it UTC. The Python-fallback path in the same file already does this correctly via `.replace(tzinfo=timezone.utc)` for naive values — worth applying the same pattern here so both code paths agree on wall-clock time regardless of host timezone.

**2. Unbounded in-memory scan on the SQLite fallback, now reachable by a broader role**
`mcpgateway/services/observability_service.py:1875` (`_execution_timeseries_python` / `_latency_percentiles_python`) loads every matching trace row into memory, bounded only by the 168h time window — no row-count cap. This was previously gated behind admin-only access; with this PR, `metrics:read` is granted to `platform_viewer` by default, so a broader population can now trigger `GET /observability/metrics/timeseries?hours=168` on a busy SQLite deployment (the documented default `DATABASE_URL`) and drive an unbounded scan into app memory. Given the wider exposure, a row cap here would close off a resource-exhaustion vector.

**3. Migration's role lookup can silently patch the wrong row**
`mcpgateway/alembic/versions/9935d863930b_add_metrics_read_to_default_roles.py:82` selects the target role via `SELECT id, permissions FROM roles WHERE name = :name AND scope = :scope LIMIT 1` with no `is_active` filter and no deterministic ordering. If a soft-deleted/stale `platform_viewer` row exists alongside the active one, `LIMIT 1` could pick the wrong row, log "Updated role..." as if it succeeded, and leave the actually-active role without `metrics:read`. (This pattern is copied from earlier migrations, so it's pre-existing — but worth breaking the cycle here rather than propagating it further.)

## Worth considering (non-blocking)

- `mcpgateway/api/v1/__init__.py:114` imports `observability_metrics_router` unconditionally, unlike the neighboring Group B router import a few lines down which wraps in try/except. A future import-time failure in `routers/observability.py` would now take down app startup entirely rather than degrading gracefully — matching the sibling pattern would keep that isolation.
- Bucket-time normalization logic is duplicated near-verbatim across four spots (two functions in `observability_service.py` plus their `admin.py` counterparts) — the naive-datetime issue above already needed the same fix applied twice, which is a good signal a shared `_normalize_bucket_time` helper would pay for itself.
- `observability_service.py:1985` — the `if durations:` guard in `_latency_percentiles_python`'s final loop is effectively always true, since `grouped` is a `defaultdict(list)` that only gains a key when a value is appended. Harmless, but could be dropped to avoid implying a case that can't happen.
- Separately from the code: `bootstrap_db.py`'s `bootstrap_default_roles()` already converges `is_system_role` permissions (including the new `metrics:read` entry) on every app startup. For the common deployment shape (`EMAIL_AUTH_ENABLED=true` with an existing admin), a restart alone already grants `metrics:read` without the migration. The migration is a genuine safety net for the case where bootstrap is skipped — might be worth a short comment in the migration noting that relationship, so future readers don't wonder why both exist.

Happy to take another pass once these land.

a-effort:
Blocking item 2 from @msureshkumar88's review (unbounded in-memory scan on the SQLite fallback) is now tracked separately as #6286, so it doesn't need to delay this PR.

LLM-assisted review notes:

The trace table has no enforced ceiling:`observability_max_traces` (default 100000) and `observability_trace_retention_days` (default 7) are declared in `config.py` and referenced nowhere else in the codebase. No cleanup service consumes them, unlike `metrics_retention_days`, which `services/metrics_cleanup_service.py` enforces against the separate performance-metrics table. That is already filed as #5746. It matters here because the 168h window is not bounded by the nominal 100k figure, only by ingest rate. Fixing #5746 shrinks the worst case without touching the query path at all, so it may be the cheaper of the two options.

The endpoints fall through the rate-limit tiers: `RateLimitMiddleware` is installed globally in `main.py`, but its `LOW` pattern is `^/(health|metrics|docs|openapi)(/|$)`, which does not match `/observability/metrics/*` or `/v1/observability/metrics/*`. They land on `default_tier`, which happens to equal `rate_limit_low_rpm` (500 rpm), so the number is the same but arrived at by fallthrough rather than classification. Tracked in #6286.

One small correction to the review text for the record: both fallbacks are column-only queries (`db.query(ObservabilityTrace.start_time, ...)`), so it is a 2-tuple per trace rather than a hydrated ORM row. Still on the order of 100-200 MB at 1M traces in window, so the concern stands, just not at full-row cost.

**Remaining to do items for this PR**:

- Item 1 (naive-datetime `astimezone`) at `observability_service.py:1851`, `:1923` and `admin.py:18943`, `:19108`. The module already has `ensure_timezone_aware()` at `observability_service.py:98` doing the right thing, so this should be four call sites.
- Item 3 (migration role lookup) at `9935d863930b:82`. Confirmed worth doing rather than deferring as pre-existing: the unique index at `db.py:1167` is partial on `is_active = true`, so multiple inactive rows at the same `(name, scope)` are legal and `LIMIT 1` can genuinely pick the wrong one.
- The DCO failure is a false negative. It reports "12 commits incorrectly signed off", but the branch is a single commit against `main` with a matching `Signed-off-by` trailer and identical author and committer. It ran eleven seconds after the force-push and evaluated the pre-squash commit list. Any new push clears it.

@Lang-Akshay your two items were addressed in `0f57be7b`, now folded into the squash: the migration grants only `("platform_viewer", "global", ...)` and the lookup filters on `AND scope = :scope`. Please re-review or dismiss when you get a chance so the decision reflects current head.

msureshkumar88:
## Re-review: real e2e verification + outstanding items

Ran a full end-to-end pass against a real `uvicorn` process (not TestClient) on `baf6b1636` — fresh scratch SQLite DB, the actual production `bootstrap_db.py` bootstrap path, real `EmailAuthService`/`RoleService` calls to create a `platform_viewer` user and a team-scoped-only user, real JWTs via `create_jwt_token`, and real `curl` requests.

**Everything the PR claims works, does work:**
- Alembic: single head, clean upgrade → downgrade → upgrade roundtrip.
- Router mounting: `observability/metrics/*` is live and returns empty series with `OBSERVABILITY_ENABLED=false`, while the flag-gated `observability/traces` correctly 404s in that state — confirmed in server logs and over HTTP, not just unit tests.
- RBAC: unauthenticated → 401, a user holding only a team-scoped role (`developer`, no global `platform_viewer`) → 403 on both endpoints, a plain non-admin user (auto-granted global `platform_viewer` per the docs) → 200, platform admin → 200.
- `hours`/`interval_minutes` bounds → 422 on all four out-of-range cases tested.
- Versioned `/v1/...` alias mounts correctly alongside the legacy path.
- Real trace rows inserted through the app's own models bucket and percentile-interpolate exactly as hand-computed, with UTC-offset timestamps.
- Full regression pass: `test_observability_metrics.py`, `test_token_scoping.py`, `test_rbac_permission_matrix.py`, `test_rbac_decorator_coverage.py`, `test_api_v1.py`, `test_legacy_router.py`, `test_main_extended.py`, `test_observability_service.py` — all green, no failures.

So the endpoint behavior, RBAC design (`platform_viewer`-only grant to avoid cross-tenant disclosure), and mounting contract are all solid and match the issue's intent.

**Not approving yet — two items from the 2026-08-17 review are still open on this head, and this e2e pass wouldn't have caught either since they're edge-case/history-dependent, not happy-path:**

1. **UTC bucket serialization still doesn't reuse `ensure_timezone_aware()`** at `observability_service.py:1851`, `:1923`, and the `admin.py` copies at `:18943`, `:19108`. All four still call `row.bucket.astimezone(timezone.utc)` directly. The module already defines `ensure_timezone_aware()` for exactly this and uses it elsewhere in the same file (lines 407, 606) — worth the same defensive consistency here in case a different driver/adapter ever hands back a naive datetime.

2. **Migration role lookup at `9935d863930b:82`** (`SELECT id, permissions FROM roles WHERE name = :name AND scope = :scope LIMIT 1`) still has no `is_active` filter or deterministic ordering. `uq_roles_name_scope_active` is a *partial* unique index (only enforces uniqueness among `is_active = true` rows), so multiple inactive `('platform_viewer', 'global')` rows are legal, and `LIMIT 1` could bind to one of those instead of the live role — the migration would report success while the active role never gets `metrics:read`.

Both were flagged as "remaining to do" in the 08-17 review comment and confirmed as real (not pre-existing) gaps in that same thread — they just haven't been pushed yet. Once those land, this looks ready to merge.

msureshkumar88:
## Follow-up: raw e2e evidence (config, commands, output)

My earlier comment on this PR summarized results narratively without the underlying evidence. Posting that evidence here, as it was actually captured — real `uvicorn` process, fresh scratch SQLite DB, the actual `bootstrap_db.py` bootstrap path, real `EmailAuthService`/`RoleService` calls, real JWTs, real `curl`. No mocks, no TestClient. Run on `baf6b1636` (branch `5945-app-metrics-endpoints`).

### Config used

**Server env — pass 1 (`OBSERVABILITY_ENABLED=false`):**
```
DATABASE_URL=sqlite:////tmp/.../scratchpad/e2e6009/e2e.db
JWT_SECRET_KEY=e2e-test-secret-key-32bytes-minimum-length-ok
AUTH_REQUIRED=true
MCPGATEWAY_UI_ENABLED=false
MCPGATEWAY_ADMIN_API_ENABLED=false
OBSERVABILITY_ENABLED=false
HOST=127.0.0.1
PORT=4499
LOG_LEVEL=INFO
```
**Server env — pass 2:** identical, `OBSERVABILITY_ENABLED=true` (toggled and server restarted).

---

### 1. Alembic — single head, upgrade → downgrade → upgrade roundtrip

```
cd mcpgateway
DATABASE_URL="sqlite:///.../e2e.db" ../.venv/bin/python -m alembic upgrade head
../.venv/bin/python -m alembic heads
../.venv/bin/python -m alembic current
../.venv/bin/python -m alembic downgrade -1
../.venv/bin/python -m alembic upgrade head
../.venv/bin/python -m alembic heads
```

`upgrade head` (tail, migration under test running last):
```
...
Created unique index: uq_roles_name_scope_active
Created unique index: uq_user_roles_email_role_scope_null_active
Created unique index: uq_user_roles_email_role_scope_id_active
SQLite detected: Using batch mode to fix constraints...
✓ Dropped 'unique_gateway_user' and created 'uq_oauth_gateway_user'
Updated role 'platform_viewer' permissions.
```

`alembic heads` → `9935d863930b (head)`

`alembic downgrade -1`:
```
Running downgrade 9935d863930b -> e4f5a6b7c8d9
Updated role 'platform_viewer' permissions.
```

`alembic upgrade head`:
```
Running upgrade e4f5a6b7c8d9 -> 9935d863930b
Updated role 'platform_viewer' permissions.
```

`alembic heads` (post-roundtrip) → `9935d863930b (head)`

**PASS.**

---

### 2. Bootstrap grants `metrics:read` to `platform_viewer`

```
DATABASE_URL="sqlite:///.../e2e.db" JWT_SECRET_KEY="..." .venv/bin/python -m mcpgateway.bootstrap_db
```
```
__main__ - INFO - Creating platform admin user: admin@example.com
mcpgateway.services.email_auth_service - INFO - Created new user: admin@example.com
__main__ - INFO - Created system role: platform_admin
__main__ - INFO - Created system role: team_admin
__main__ - INFO - Created system role: developer
__main__ - INFO - Created system role: viewer
__main__ - INFO - System role platform_viewer already exists - skipping
mcpgateway.services.role_service - INFO - Assigned role platform_admin to admin@example.com (scope: global, scope_id: )
__main__ - INFO - Default RBAC roles bootstrap completed successfully
```

Direct DB check on `roles` table:
```
('platform_viewer', 'global', 1, '[..., "metrics:read"]')
```
**PASS** — `metrics:read` present in the live permissions array.

---

### 3. User + JWT setup

Users created via the real service layer (`EmailAuthService.create_user`, `RoleService.assign_role_to_user`):
```
created viewer user: viewer@example.com
created teamonly user: teamonly@example.com
created team: 454aaeb79b4d4e9c9bcf0dd33565af8a
assigned team-scoped 'developer' role to teamonly@example.com on team 454aaeb79b4d4e9c9bcf0dd33565af8a
viewer@example.com role assignments: [('platform_viewer', 'global', 'global'), ('team_admin', 'team', 'team')]
teamonly@example.com role assignments: [('developer', 'team', 'team')]
```
JWTs generated via `python -m mcpgateway.utils.create_jwt_token --username <email> --exp 60 --secret <key>` for `admin@example.com`, `viewer@example.com`, `teamonly@example.com`. **PASS.**

---

### 4. Server startup (pass 1, observability off)

```
nohup .venv/bin/uvicorn mcpgateway.main:app --host 127.0.0.1 --port 4499 > server.log 2>&1 &
```
```
mcpgateway.api.v1 - INFO - Observability metrics summary router included (/observability/metrics)
mcpgateway.api.v1 - INFO - Observability router not included - observability disabled
uvicorn.error - INFO - Application startup complete.
uvicorn.error - INFO - Uvicorn running on http://127.0.0.1:4499
```
One ERROR line, unrelated to this PR (pre-existing UAID cross-gateway config warning). No tracebacks, no import errors. **PASS.**

---

### 5. HTTP matrix (pass 1, observability off)

All via `curl -s -o <file> -w "%{http_code}\n" [-H "Authorization: Bearer <token>"] "http://127.0.0.1:4499<path>[?params]"`.

**No auth → 401**
```
GET /observability/metrics/timeseries  -> 401 {"detail":"Authorization token required"}
GET /observability/metrics/percentiles -> 401 {"detail":"Authorization token required"}
```

**teamonly@example.com (team-scoped `developer` only, no global platform_viewer) → 403**
```
GET /observability/metrics/timeseries  -> 403 {"detail":"Access denied"}
GET /observability/metrics/percentiles -> 403 {"detail":"Access denied"}
```

**viewer@example.com (auto platform_viewer) → 200, empty series**
```
GET /observability/metrics/timeseries  -> 200 {"buckets":[],"values":[]}
GET /observability/metrics/percentiles -> 200 {"buckets":[],"p50":[],"p95":[],"p99":[]}
```

**viewer@example.com → GET /observability/traces (gated router, flag off) → 404**
```
404 {"detail":"Not Found"}
```

**viewer@example.com → GET /v1/observability/metrics/timeseries → 200**
```
200 {"buckets":[],"values":[]}
```

**Bounds validation → 422 each**
```
?hours=0                -> 422  {"detail":"An error occurred, please try again."}
?hours=169               -> 422  {"detail":"An error occurred, please try again."}
?interval_minutes=4      -> 422  {"detail":"An error occurred, please try again."}
?interval_minutes=1441   -> 422  {"detail":"An error occurred, please try again."}
```
(Status code is correct; note the body is a generic masked message rather than FastAPI's normal per-field validation detail — not introduced by this PR, likely an existing exception-handler behavior, worth knowing.)

**admin@example.com (wildcard `*` role) → 200**
```
GET /observability/metrics/timeseries  -> 200 {"buckets":[],"values":[]}
GET /observability/metrics/percentiles -> 200 {"buckets":[],"p50":[],"p95":[],"p99":[]}
```
**PASS — every response matched the expected status/body.**

---

### 6. Data path + UTC serialization (pass 2, observability on)

Rows inserted via the real `ObservabilityTrace` model: bucket_a +5min (duration 100.0), bucket_a +10min (duration 200.0), bucket_b +5min (duration 300.0).

Server restarted; log confirmed both routers now mounted:
```
mcpgateway.api.v1 - INFO - Observability metrics summary router included (/observability/metrics)
mcpgateway.api.v1 - INFO - Observability router included - observability API endpoints enabled
```

`GET /observability/metrics/timeseries?hours=168` (viewer token):
```json
{"buckets": ["2026-08-18T09:00:00+00:00", "2026-08-18T10:00:00+00:00"], "values": [2, 1]}
```

`GET /observability/metrics/percentiles?hours=168` (viewer token):
```json
{
  "buckets": ["2026-08-18T09:00:00+00:00", "2026-08-18T10:00:00+00:00"],
  "p50": [150.0, 300.0], "p95": [195.0, 300.0], "p99": [199.0, 300.0]
}
```

Hand-computed (percentile_cont linear interpolation, `k = p*(n-1)`):
- bucket_a `[100, 200]`: p50 → 150.0, p95 → 195.0, p99 → 199.0
- bucket_b `[300]`: all percentiles → 300.0

Matches exactly. Bucket timestamps carry `+00:00` UTC offset.

Also re-checked `GET /observability/traces` with the flag now on: returned `403 {"detail":"Access denied"}` (viewer lacks `admin.system_config` — expected, distinct from the pass-1 404-on-unmounted-router case). **PASS.**

---

### 7. Regression pytest suites

```
.venv/bin/python -m pytest \
  tests/unit/mcpgateway/routers/test_observability_metrics.py \
  tests/unit/mcpgateway/services/test_rbac_permission_matrix.py \
  tests/security/test_rbac_decorator_coverage.py \
  tests/unit/mcpgateway/test_api_v1.py \
  tests/unit/mcpgateway/test_legacy_router.py \
  tests/unit/mcpgateway/test_main_extended.py \
  tests/unit/mcpgateway/services/test_observability_service.py \
  -q --no-header -rf
```
Dot-progress output, no `F`/`E` markers anywhere, `-rf` (forces failure report) produced no failure entries, `EXIT CODE: 0`. Two non-failing deprecation warnings only (Pydantic class-config style, starlette/httpx import). Note: this run's captured output did not include an explicit `N passed` summary line, so an exact passed-test count isn't being claimed — the evidence is the clean exit code plus zero failure markers/entries. **PASS (no failures observed).**

---

### Not exercised by this run

Two items flagged in the 2026-08-17 review thread remain unfixed on `baf6b1636` and are edge-case/history-dependent, so this e2e pass did not (and structurally could not) exercise them:
- Raw `.astimezone(timezone.utc)` instead of `ensure_timezone_aware()` at `observability_service.py:1851`, `:1923`, `admin.py:18943`, `:19108`.
- Migration `SELECT ... LIMIT 1` at `9935d863930b:82` with no `is_active` filter — unsafe under the partial unique index on `roles` (`uq_roles_name_scope_active`), but requires a duplicate-inactive-row history this fresh scratch DB doesn't have.

Net: happy-path functional behavior is solid and matches the issue's intent; the two outstanding items are still the reason this isn't approved yet.

msureshkumar88 (changes requested):
Re-reviewed against the current head (`baf6b1636`) with a real e2e pass (see comments above for full config/commands/output) — the endpoint behavior, RBAC design, and mounting contract all check out. Requesting changes on two items from the 2026-08-17 thread that are still open on this head (no commits since):

## Blocking

**1. Migration role lookup can silently patch the wrong row**
`mcpgateway/alembic/versions/9935d863930b_add_metrics_read_to_default_roles.py:82`

```python
row = conn.execute(
    text("SELECT id, permissions FROM roles WHERE name = :name AND scope = :scope LIMIT 1"),
    {"name": role_name, "scope": scope},
).fetchone()
```

`uq_roles_name_scope_active` is a *partial* unique index (`postgresql_where=text("is_active = true")` / `sqlite_where=text("is_active = 1")` in `db.py`), so multiple **inactive** `('platform_viewer', 'global')` rows are legal. With no `is_active` filter and no deterministic ordering, `LIMIT 1` can bind to a stale/inactive row instead of the live one — the migration logs "Updated role..." as if it succeeded, while the actually-active `platform_viewer` role never receives `metrics:read`. Every non-admin user would then 403 on both new endpoints with no visible failure anywhere.

Suggested fix: add `AND is_active = true` (dialect-appropriate) to the `SELECT`, or an explicit `ORDER BY is_active DESC` tiebreak.

**2. UTC bucket serialization doesn't reuse the module's own naive-datetime guard**
`mcpgateway/services/observability_service.py:1851`, `:1923`, and the `admin.py` copies at `:18943`, `:19108`

All four still do `row.bucket.astimezone(timezone.utc).isoformat()` directly. `observability_service.py` already defines `ensure_timezone_aware()` (line 99) for exactly this — naive datetime in, UTC-aware out — and uses it elsewhere in the same file (lines 407, 606). Calling `.astimezone()` directly on a datetime that happens to be naive silently assumes the *host's local* timezone rather than UTC, with no error. Today's psycopg driver always returns tz-aware values for `TO_TIMESTAMP()`, so this isn't currently reachable — but it's a one-line fix per site to close the gap and match the pattern already established in the file, and it was flagged as a real (not pre-existing) gap in the 08-17 review rather than deferred.

Everything else — RBAC tiering (`platform_viewer`-only grant, `allow_admin_bypass=False`), unconditional-vs-flag-gated mounting, bounds validation, alembic single-head/round-trip, and test coverage — verified working via real e2e and is ready as-is. Happy to take another pass once these two land.

madhu-mohan-jaishankar:
Thanks for all the iterations here @Altamimi-Dev, and for the quick turnaround last week.

To get this landed sooner, I've rebased your commit onto current main and addressed the two blocking items from @msureshkumar88's review on top - https://github.com/IBM/mcp-context-forge/pull/6293

our commit is unchanged (same patch, authorship and sign-off intact)

msureshkumar88:
This PR's changes have been merged via #6293. Could a maintainer close this out?

madhu-mohan-jaishankar:
completed in https://github.com/IBM/mcp-context-forge/pull/6293

Issue #5437, opened 2026-06-30
Human in the loop Approval for tool execution


mrm1404-personal:
### 🧭 Type of Feature

Please select the most appropriate category:

- [x] Enhancement to existing functionality
- [ ] New feature or capability
- [ ] New MCP-compliant server
- [ ] New component or integration
- [ ] Developer tooling or test improvement
- [ ] Packaging, automation and deployment (ex: pypi, docker, quay.io, kubernetes, terraform)
- [ ] Other (please describe below)

### 🧭 Epic

**Title:** Human in the loop approval for tool execution
**Goal:** Admin User to be able to make decision on which tools require human approval before tool gets executed
**Why now:** In Cyberfraud, we have requirement for Investigators review of tools that require approval due to the data that is processed by tools and the different mcp tools available for specific activities

---

### 🧑🏻‍💻 User Story 1

**As a:** Admin User, 
**I want to:** auto-approve/ seek manual approval for tool executions through agent calls.
**So that:** Admin user will be able to get User's Review and Approval before actually hitting the MCP tool and getting unrequired response. Will Also be able to get benefit of cost if incorrect mcp tool gets invoked or incorrect parameters get invoked

For example: Let's Say, there is a mcp tool to fetch User's Identity Risk Score, which requires Human to review the actual tool that is called & the parameters that are sent for that tool call. 
In this case Admin User will keep the auto-approval as False, so that before any tool execution of that mcp tool, it would seek confirmation from user who is asking for getting the Risk Score checked with correct identification parameter. 
Otherwise for other tools that do not require humans to review, Admin can keep auto-approval as True for those tools

#### ✅ Acceptance Criteria
```gherkin
Scenario: Admin User provides auto-approval to the tool execution
  Given: 2 tools are already created in mcp context forge
  When: Admin user provides auto-approval for first tool
  Then: While Agent executes this first tool, user won't be asked to approve/reject/edit the tool call

Scenario: Admin User does NOT provide auto-approval to the tool execution
  Given: 2 tools are already created in mcp context forge
  When: Admin user does not auto-approve for second tool
  Then: While Agent executes this second tool, user WILL BE asked to approve/reject/edit the tool call
```

#### Flow Diagram

```mermaid
flowchart TD
    Start([Agent Needs to Execute Tool]) --> CheckTool{Check Tool Configuration}
    
    CheckTool --> GetAutoApproval[Retrieve Auto-Approval Setting<br/>from Admin Configuration]
    
[...]

PR #6014, opened 2026-07-31
feat(plugins): add invinoveritas Review plugin (independent pre-action verdict gate)

babyblueviper1:
Grew out of #5437 (Human in the loop Approval for tool execution). Reading through the epic and hegu-1's proposal there, I noticed a real distinction worth making concrete as a plugin rather than just a comment: a `risk_tier`/`gate.mode` rule engine decides **eligibility** deterministically and fast (may this actor/tool/resource combination happen at all) — that's what `unified_pdp` and the other example plugins here already do well. It structurally cannot decide **soundness for this specific payload** — the same tool with different arguments can be fine or genuinely risky, which no finite rule set distinguishes without an explosion of conditions. That's a judgment call, which is what this plugin gates on.

## What it does

- Hooks `tool_pre_invoke`, calls `POST /review` (api.babyblueviper.com) with `{tool_name, arguments}` before the tool runs.
- A `reject` verdict blocks the call by default (`block_on_reject: true`). Set `false` for an advisory/observe-only rollout that only annotates the result, never blocks.
- **Fails open** on any `/review`-side problem (network error, timeout, malformed response, missing key) — never hangs or crashes the gateway on our service having a bad moment; `metadata["invinoveritas_review"] = "unavailable"` makes this visible rather than silent.
- Optional `sign: true` attaches a portable, independently-verifiable signed proof (`/verify-proof`, free, no auth) to every verdict.
- Disabled by default in `plugins/config.yaml`, matching every other opt-in plugin in this repo (e.g. the bundled `cpex-*` plugins).

## How it composes with a PDP

Not a replacement for `unified_pdp` or similar — the two answer different questions and are meant to run together: PDP first (cheap, deterministic, denies the clearly-disallowed), review second (judgment, for whatever the PDP lets through). Full rationale in `plugins/invinoveritas_review/README.md`.

## Testing

```
pytest plugins/invinoveritas_review/tests/test_invinoveritas_review.py -v
```

6/6 passing, real `httpx.MockTransport` (no live API key needed to run the suite) — covers reject-blocks, advisory-mode-never-blocks, approve-passes-through, fail-open-on-timeout, fail-open-on-malformed-response, and no-api-key-skips-the-call-entirely.

Happy to adjust naming/config shape/anything else — first PR to this repo, want to match the existing plugin conventions as closely as possible rather than introduce a new pattern.

### DIFF
--- a/plugins/config.yaml
+++ b/plugins/config.yaml
@@ -1346,3 +1346,27 @@ plugins:
       detect_ssn: true
       block_on_detection: false
       log_detections: false
+
+  # invinoveritas Review — independent, signed pre-action verdict for tool_pre_invoke.
+  # Distinct from PDP/rule-based gates (unified_pdp, deny_filter, etc.): those decide
+  # eligibility deterministically (may this actor/tool/resource combination happen at
+  # all); this decides soundness for THIS specific payload (a judgment call a static
+  # rule can't make). See plugins/invinoveritas_review/README.md for the full design.
+  # Disabled by default, matching every other opt-in plugin here -- requires an
+  # IVV_API_KEY (free registration: POST https://api.babyblueviper.com/register).
+  - name: "InvinoveritasReviewPlugin"
+    kind: "plugins.invinoveritas_review.invinoveritas_review.InvinoveritasReviewPlugin"
+    description: "Gates tool_pre_invoke on an independent invinoveritas /review verdict."
+    version: "0.1.0"
+    author: "babyblueviper1"
+    hooks: ["tool_pre_invoke"]
+    tags: ["plugin", "safety", "review", "human-in-the-loop"]
+    mode: "disabled" # enforce | enforce_ignore_error | permissive | disabled
+    priority: 100
+    conditions: []
+    config:
+      base_url: "https://api.babyblueviper.com"
+      block_on_reject: true
+      sign: false
+      artifact_type: "general"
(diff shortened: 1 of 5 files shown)


cafalchio:
Thank you @babyblueviper1 . We are removing plugins from this repo. We are planning to create a marketplace of plugins that can be shared between projects that use cpex as plugin manager.
I will close this PR, but it will be a great candidate for the future marketplace.

babyblueviper1:
Appreciate the note, and the reasoning makes sense -- a per-repo plugin doesn't compose well once there's a real shared marketplace to point at instead. Happy to be an early submission once cpex/the marketplace is real and taking entries; the plugin itself doesn't need any real rework to move, it's a thin `tool_pre_invoke` hook with no other coupling to this repo's internals.

If there's a tracking issue or doc for the marketplace once it exists, I'd genuinely like to follow it -- the eligibility-vs-soundness distinction this plugin makes concrete (deterministic rule engine decides IF an action may happen at all, independent verdict decides whether THIS specific payload is sound) seems like exactly the kind of thing worth being first-class in a shared plugin ecosystem rather than reinvented per-integration.

cafalchio:
Hi @babyblueviper1 thank you for the understanding, the cpex framework is being developed here. https://github.com/contextforge-org/cpex I think you can ask about the marketplace on their repo, they are responsive.

PR #6035, opened 2026-08-04
chore: bump cryptography to 50.0.0

msureshkumar88:
## Summary

Fixes the `Docker Security Scan` / Grype CVE gate failure on `main` by bumping `cryptography`.

- Bumps the `cryptography` floor from `>=49.0.0` to `>=50.0.0` in `pyproject.toml`.
- 50.0.0 was published after this project's rolling 10-day `exclude-newer` window, so it needs an explicit `[tool.uv.exclude-newer-package]` entry to let uv's resolver consider it — same pattern already used for other early bumps (`cpex-*`, etc.).
- Regenerated `uv.lock` (`uv lock --refresh`): only `cryptography` moves version, plus unrelated pre-existing lock churn from rebasing onto the latest `main`.

## Verification

- Targeted unit tests (JWT/crypto/auth-encryption paths): pass.
- Built the regular `Containerfile` image, generated its SBOM (Syft), scanned with Grype using the same flags as CI (`--only-fixed --severity-cutoff high --fail-on high`): the flagged finding on this dependency is gone. Two pre-existing, unrelated medium-severity `p11-kit`/`p11-kit-trust` RPM findings (base-image OS packages) remain but are below the high-severity gate.
- Built the FIPS-enabled image (`ENABLE_FIPS=true`, `PYTHON_VERSION=3.11`, matching the `docker-scan.yml` FedRAMP job's exact build args) and ran `scripts/fedramp-validate.sh` against it: 22/22 checks pass.
- Confirmed both the regular and FIPS images resolve `cryptography==50.0.0` at runtime.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -63,6 +63,7 @@ cpex-secrets-detection = "2026-07-27T23:59:59Z"
 # [plugins] extra to nothing and the gateway would fail to load SQLSanitizer at boot.
 # Vetted ahead of the usual window for this reason.
 cpex-sql-sanitizer = "2026-07-21T23:59:59Z"
+cryptography = "2026-08-01T23:59:59Z"
 langgraph = "2026-05-12T23:59:59Z"
 fastapi = "2026-07-27T23:59:59Z"
 filelock = "2026-07-27T23:59:59Z"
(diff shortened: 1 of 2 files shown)


ja8zyjits (approved):
LGTM

prakhar-singh1928 (approved):

PR #6044, opened 2026-08-05
fix(ui): restore gateway selection when editing virtual servers

### LINKED ISSUE #6042: [BUG][UI]: Virtual server edit does not show which MCP Servers are selected, so a gateway's tools cannot be removed via the UI
### 🐞 Bug Summary

In the Edit Virtual Server modal, the MCP Servers selector always renders with every checkbox unchecked, whatever gateways the virtual server's tools actually came from. Because of that there's no way to remove one MCP server's worth of tools from a virtual server through the Admin UI.

The selector reads as persistent composition state. It's a checkbox list with Select All and Clear All, and the helper text under it says "Choose an MCP server, then select the tools, resources, and prompts to configure your virtual server." In practice it's a transient filter, and the selection is dropped on submit.

The tool-level selection store from #3168 works well, and it's what let us confirm no data is at risk here. This is a separate path.

---

### 🧩 Affected Component

- [x] `mcpgateway` - API

---

### 🔁 Steps to Reproduce

1. Create a virtual server whose tools come from several registered gateways (ours draws from six).
2. Reopen it via Virtual Servers, then Edit.
3. Look at the MCP Servers list (`#associatedEditGateways`).
4. Every checkbox is unchecked, including the gateways this server demonstrably uses. Searching for a specific one such as `memo-mcp-gw` surfaces the row, still unchecked.

Measured on our instance, on a virtual server with 192 associated tools spanning six gateways:

```js
document.getElementById('associatedEditGateways')  // 51 rendered, 0 checked
```

---

### 🤔 Expected Behavior

Opening Edit shows the MCP Servers contributing to this virtual server already checked. Unchecking one, clicking Refresh from MCPs, and re-selecting tools then removes that server's tools. That's the natural reading of the current UI, and the workflow the helper text describes.

---

### Root cause

The underlying information already exists. `Tool` carries a gateway foreign key, so every associated tool already knows which MCP server it came from.

```python
# mcpgateway/db.py, class Tool
gateway_id: Mapped[Optional[str]] = mapped_column(ForeignKey("gateways.id", ondelete="CASCADE"))
gateway: Mapped["Gateway"] = relationship("Gateway", primaryjoin="Tool.gateway_id == Gateway.id", ...)
```

That key is what powers `GET /tools?gateway_id=...` and the `<gateway-slug>-<tool>` naming, and `Resource` carries one too. The missing piece is the server-to-gateway rollup, which is one hop from relationships that already exist.

```python
{tool.gateway_id for tool in server.tools if tool.gateway_id}
```

No migration or new association table is needed. Verified on `v1.0.5`, and the same code paths are unchanged on `main`.

| Layer | Finding |
|---|---|
| ORM | `Tool.gateway_id` and `Tool.gateway` exist. `Server` has `tools`, `resources`, `prompts` and `a2a_agents`, but no `gateways` rollup (`mcpgateway/db.py`) |
| API out | `GET /servers/{id}` returns `associatedTools`, `associatedToolIds`, `associatedResources`, `associatedPrompts` and `associatedA2aAgents`. There's no `associatedGateways` |
| API in | `ServerCreate` and `ServerUpdate` have no `associated_gateways` field (`mcpgateway/schemas.py`) |
| Route | The server-edit route in `admin.py` never reads `associatedGateways` in any casing. Its docstring lists only `associatedTools`, `associatedResources` and `associatedPrompts` |
| JS | `editServer()` seeds the tools, resources and prompts stores, then for gateways calls only `initGatewaySelect("associatedEditGateways", ...)`, which wires up pills, search and select-all but seeds no checked state (`mcpgateway/admin_ui/servers.js`) |

So the control has nothing to populate itself from, and the `associatedgateways=<id>` values the browser submits are dropped server-side.

A smaller related thing, in the same rendered gateway partial the checkbox `name` is inconsistently cased. It's `associatedGateways` for the `REST/A2A` entry and `associatedgateways` for real gateways.

---

### Why the tool-level path isn't a workaround

We tried doing this at the tool level instead and couldn't.

The tool picker's search matches tool display names only. `memo-mcp-dev-gw` returns nothing, `memo` returns 94 results, and there's no way to scope the picker to a gateway. With two gateways registered for the same upstream (a prod and a dev memo server) the results are also indistinguishable. Searching `memo abandon` gives two rows, both labelled `Memo Abandon`, both checked, with no gateway hint in either. Only the hidden tool IDs (`8c475b73` versus `2c62e629`) tell them apart. Checking a gateway in the MCP Servers list didn't filter the tool list either, it still rendered tools from an unrelated gateway.

---

### Possible directions

Roughly in order of investment.

1. Derive it read-only. Compute the distinct gateways of a server's associated tools, return them as `associatedGateways` on `ServerRead`, and have `editServer()` seed the checkboxes from that. No schema change, and it makes the existing control honest and the described workflow work. A partially-included gateway could render indeterminate.
2. Persist it. Add a real server-to-gateway association so composition is expressed at the gateway level and tool lists are derived from it. Bigger change, though it matches how the UI already presents the concept.
3. Narrow the UI. If the selector is only ever meant to be a filter, relabel it along the lines of "Filter tools by MCP server", drop the submitted-but-ignored field, and add a separate affordance for removing a gateway's tools.

Happy to test whichever direction you'd prefer.

---

### Workaround

Composition can be edited through the API, which does accept the full tool list. To remove one gateway's tools, take `associatedToolIds` from `GET /servers/{id}` and keep the ones whose tools have a different `gatewayId`.

```bash
curl -X PUT -H "Authorization: Bearer $TOKEN" -H 'Content-Type: application/json' \
  --data '{"associatedTools":[...remaining ids...],"associatedResources":[...],"associatedPrompts":[]}' \
  https://<gateway>/v1/servers/<server_id>
```

We used this to drop 45 tools belonging to one gateway and keep the other 147, including the 45 from a same-named prod gateway. Pass `associatedResources` and `associatedPrompts` explicitly to keep the result deterministic.

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Version | `v1.0.5` (`ghcr.io/ibm/mcp-context-forge:v1.0.5`), code paths above unchanged on `main` |
| Platform | Kubernetes (Helm, bundled PostgreSQL and Redis) |
| Scale | 50 registered gateways. The affected virtual server had 192 associated tools across 6 gateways |

Linxiushen:
## Summary

- derive a virtual server's contributing gateway IDs from its active tools, resources, prompts, and A2A agents
- seed and persist the gateway checkbox state in the edit modal, including across HTMX swaps
- reinitialize both create and edit gateway selectors after partial reloads
- normalize the gateway checkbox field name to `associatedGateways`

The gateway rollup is read-only, so this does not add a migration or duplicate the existing entity associations.

## Testing

- `pytest tests/unit/mcpgateway/services/test_server_service.py -q` (123 passed)
- `pytest tests/unit/mcpgateway/test_schemas.py -q` (125 passed)
- `vitest run` (2326 passed, 3 skipped)
- `ruff check mcpgateway/schemas.py mcpgateway/services/server_service.py`
- `vite build`
- `git diff --check`

Closes #6042

### DIFF
--- a/mcpgateway/admin_ui/events.js
+++ b/mcpgateway/admin_ui/events.js
@@ -707,6 +707,26 @@ import {
   // wire up the checkbox-pill UI.  Avoids new Function() / unsafe-eval.
   // ===================================================================
   const SELECTOR_SWAP_MAP = {
+    associatedGateways: () =>
+      initGatewaySelect(
+        "associatedGateways",
+        "selectedGatewayPills",
+        "selectedGatewayWarning",
+        12,
+        "selectAllGatewayBtn",
+        "clearAllGatewayBtn",
+        "searchGateways"
+      ),
+    associatedEditGateways: () =>
+      initGatewaySelect(
+        "associatedEditGateways",
+        "selectedEditGatewayPills",
+        "selectedEditGatewayWarning",
+        12,
+        "selectAllEditGatewayBtn",
+        "clearAllEditGatewayBtn",
+        "searchEditGateways"
+      ),
     associatedTools: () =>
       Admin.initToolSelect(
         "associatedTools",
--- a/mcpgateway/admin_ui/gateways.js
+++ b/mcpgateway/admin_ui/gateways.js
@@ -620,6 +620,55 @@ export const initGatewaySelect = function (
     return;
   }
 
+  const checkboxGatewayId = (checkbox) =>
+    checkbox.dataset?.gatewayNull === "true"
+      ? "null"
+      : String(checkbox.value);
+
+  const applyPersistedSelection = function () {
+    if (!container.dataset.selectedGatewayIds) return;
+
+    try {
+      const selectedIds = new Set(
+        JSON.parse(container.dataset.selectedGatewayIds).map(String)
+      );
+      container
+        .querySelectorAll('input[type="checkbox"]')
+        .forEach((checkbox) => {
+          checkbox.checked = selectedIds.has(checkboxGatewayId(checkbox));
+        });
+    } catch (error) {
+      console.error("Error restoring gateway selection:", error);
+    }
+  };
+
+  const persistSelection = function () {
+    const selectAllInput = container.querySelector(
+      'input[name="selectAllGateways"]'
+    );
+    const allIdsInput = container.querySelector(
+      'input[name="allGatewayIds"]'
+    );
+
+    if (selectAllInput?.value === "true" && allIdsInput) {
+      try {
+        container.dataset.selectedGatewayIds = JSON.stringify(
+          JSON.parse(allIdsInput.value).map(String)
+        );
+        return;
+      } catch (error) {
+        console.error("Error persisting all gateway IDs:", error);
+      }
+    }
+
+    const selectedIds = Array.from(
+      container.querySelectorAll('input[type="checkbox"]:checked')
+    ).map(checkboxGatewayId);
+    container.dataset.selectedGatewayIds = JSON.stringify(selectedIds);
+  };
+
+  applyPersistedSelection();
+
   const pillClasses =
     "inline-block bg-indigo-100 text-indigo-800 text-xs px-2 py-1 rounded-full dark:bg-indigo-900 dark:text-indigo-200";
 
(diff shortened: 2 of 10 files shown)


jonpspri:
Closing as the underlying issue (#6042) is being superseded by the forthcoming UI redesign. Thank you for the thorough diagnosis and the well-structured fix — the analysis here will be valuable context for the redesign work.

PR #6045, opened 2026-08-05
fix(catalog): preserve caller identity during registration

### LINKED ISSUE #6036: [BUG]: catalog registration creates gateways without caller identity (owner_email=None, created_by="system")
## Summary

`catalog_service.register_catalog_server` creates a gateway without any caller identity. It builds a `GatewayCreate` and calls `self._gateway_service.register_gateway(...)` with no user context, so the resulting gateway row gets no `owner_email`, no `team_id`, and platform-default creation metadata rather than the caller's.

The canonical gateway creation path does the opposite: `POST /v1/gateways` derives owner/team/metadata from the authenticated principal via `MetadataCapture.extract_creation_metadata` before persisting.

## Why it matters now

Until recently this path was admin-only (`POST /admin/mcp-registry/{server_id}/register`), so the missing attribution was mostly a bookkeeping gap. #5974 adds a non-admin equivalent (`POST /v1/catalog/{catalog_id}/register`), which widens the exposure: a non-admin caller with the required permissions can now create a federated gateway that no team owns and no user is recorded as having created. That affects:

- **Visibility / Layer-1 scoping** — a gateway with no `owner_email`/`team_id` does not participate in team-scoped filtering the way user-created gateways do.
- **Audit trail** — there is no record of which principal caused the federated peer to be registered.
- **Ownership operations** — team transfer, ownership-based edit/delete checks, and "my resources" views have nothing to key on.

## Expected

Catalog registration should attribute the created gateway to the authenticated caller, the same way `POST /v1/gateways` does: pass the user context down from the route into `register_catalog_server`, and set owner/team/creation metadata on the `GatewayCreate` / `register_gateway` call. The admin path should attribute to the admin performing the registration rather than a platform default.

## Notes

- Raised in review of #5974 (`pullrequestreview-4852454127`) and deliberately deferred out of that PR: it is a change to shared service behaviour that also affects the admin registration path, so it deserves its own change with its own tests.
- Touch points: `mcpgateway/services/catalog_service.py` (`register_catalog_server`, and the bulk register path that shares it), `mcpgateway/routers/catalog.py`, `mcpgateway/admin.py` (admin twin), and `MetadataCapture` for the metadata shape.

Linxiushen:
## Summary

- capture the authenticated caller's owner, personal-team, and request metadata for admin catalog registration
- require an explicit `CatalogRegistrationContext` through both single and bulk service paths
- preserve the same ownership metadata and emit the gateway creation audit event for OAuth catalog entries that bypass `GatewayService.register_gateway`

## Testing

- `pytest tests/unit/mcpgateway/services/test_catalog_service.py -q` (59 passed)
- `pytest tests/unit/mcpgateway/test_admin.py -k catalog -q` (11 passed)
- `pytest tests/unit/mcpgateway/test_admin_catalog_htmx.py -q` (7 passed)
- `ruff check mcpgateway/admin.py mcpgateway/services/catalog_service.py tests/unit/mcpgateway/services/test_catalog_service.py`
- `ruff format --check mcpgateway/admin.py mcpgateway/services/catalog_service.py tests/unit/mcpgateway/services/test_catalog_service.py tests/unit/mcpgateway/test_admin.py`

Closes #6036

### DIFF
--- a/mcpgateway/admin.py
+++ b/mcpgateway/admin.py
@@ -159,7 +159,7 @@ from mcpgateway.services.a2a_agent_plugin_binding_service import A2AAgentPluginB
 from mcpgateway.services.a2a_service import A2AAgentError, A2AAgentNameConflictError, A2AAgentNotFoundError, A2AAgentService
 from mcpgateway.services.argon2_service import Argon2PasswordService
 from mcpgateway.services.audit_trail_service import get_audit_trail_service
-from mcpgateway.services.catalog_service import catalog_service
+from mcpgateway.services.catalog_service import CatalogRegistrationContext, catalog_service
 from mcpgateway.services.content_security import ContentSizeError, ContentTypeError, TemplateValidationError
 from mcpgateway.services.csrf_service import get_csrf_service
 from mcpgateway.services.email_auth_service import AuthenticationError, EmailAuthService, PasswordValidationError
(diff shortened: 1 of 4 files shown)


jonpspri:
Closing as superseded by the merged PR #6247 (Security Hardening: Catalog Registration Scope Enforcement, merged 2026-08-26).

PR #6247 fully addressed the ownership attribution gap described in #6036 — `register_catalog_server` now requires `created_by` and `owner_email` as non-optional parameters, persists them to the gateway row, and adds scope enforcement and team membership validation on top. The current `main` already contains all of the fixes this PR set out to deliver.

Thanks for the contribution — the problem you identified and the approach you took were both on the right track.

PR #6049, opened 2026-08-05
fix(oauth): preserve gateway catalog on empty fetch

### LINKED ISSUE #6046: OAuth fetch-tools deletes a gateway's whole catalog on an empty upstream response
### 🐞 Bug Summary

`fetch_tools_after_oauth()` calls `_reconcile_gateway_catalog()` without the
empty-response guard that every comparable path uses, so a `POST /oauth/fetch-tools/{gateway_id}`
that returns an empty `tools/list` **deletes the gateway's entire catalog** —
tools, resources, prompts, their `ToolMetric` rows, and their
`server_tool_association` rows (i.e. virtual server membership).

The endpoint still answers `HTTP 200` with `"Successfully fetched and created 0 tools"`,
so it reads as a harmless no-op while the catalog and its virtual-server wiring
are gone.

An empty response from an `authorization_code` gateway is a normal consequence of
a token that is accepted for `initialize` but carries no workspace authorization —
exactly the state after a half-completed re-authorization. The codebase already
says so, at `gateway_service.py:3616-3620`:

> For authorization_code OAuth gateways, empty responses may indicate a missing
> auth token rather than genuine removal of all items. Skip stale cleanup to
> prevent destructive deletion of tools, resources, prompts, and their virtual
> server associations. Mirrors the guard in `_refresh_gateway_tools_resources_prompts`.

That guard was never applied to the OAuth fetch path — which is the one path that
has already proven `grant_type == "authorization_code"` (it raises otherwise at
`gateway_service.py:2220`), so the guard's precondition is unconditionally true there.

### 🧩 Affected Component

- [x] `mcpgateway` - API
- [ ] Other (explain below)

---

### 🔁 Steps to Reproduce

1. Register a gateway with `grant_type: authorization_code` and complete the OAuth flow so tools are discovered. Add some of those tools to a virtual server.
2. Put the stored user token into a state where the upstream server accepts `initialize` but returns an empty `tools/list` (e.g. the upstream app authorization is revoked while the token itself remains parseable).
3. `POST /oauth/fetch-tools/{gateway_id}` — via the OAuth callback page button or the Gateways-list button.
4. Observe the `HTTP 200` response body: `Successfully fetched and created 0 tools`.
5. Query the database.

```sql
select count(*) from tools where gateway_id = '<gateway-id>';                 -- 0
select count(*) from server_tool_association sta join tools t on t.id = sta.tool_id
 where t.gateway_id = '<gateway-id>';                                         -- 0
```

Everything the gateway contributed is deleted, and re-fetching later restores the
tools but **not** their virtual server associations, which have to be rebuilt by hand.

---

### 🤔 Expected Behavior

An empty catalog from an `authorization_code` gateway should be treated as
"probably an auth problem", not as "the upstream deleted everything" — the same
way the reactivation and refresh paths already treat it.

Two differences from the guarded paths, both at `gateway_service.py:2302-2307`:

```python
reconcile_result = self._reconcile_gateway_catalog(
    db,
    gateway=gateway,
    catalog_sync=catalog_sync,
    log_context="gateway OAuth fetch",
)
```

1. `skip_stale_cleanup` is not passed, so it defaults to `False` and stale cleanup runs.
2. `stale_created_via_values` is not passed, so it defaults to `None`, and
   `_created_via_allowed()` (`gateway_service.py:6009`) returns
   `stale_created_via_values is None or …` — meaning **every** row is eligible for
   deletion, including hand-created `api`/`ui`/legacy entries. The refresh path at
   `gateway_service.py:6335` restricts this to MCP-discovered origins.

Of the five `_reconcile_gateway_catalog()` call sites, this is the only
hand-triggered one with no protection at all:

| Line | `log_context` | Guard |
|---|---|---|
| 2302 | `gateway OAuth fetch` | **none** |
| 3128 | `gateway update` | none |
| 3625 | `gateway reactivation` | `skip_stale_cleanup=` |
| 4281 | `gateway lifecycle worker` | none |
| 6335 | `gateway refresh (…)` | early return on empty + `stale_created_via_values` |

Suggested fix — mirror the reactivation guard:

```python
skip_stale_cleanup = not tools and not resources and not prompts
if skip_stale_cleanup:
    logger.warning(
        "Empty catalog from auth_code gateway %s during OAuth fetch, preserving existing items",
        gateway.name,
    )
reconcile_result = self._reconcile_gateway_catalog(
    db,
    gateway=gateway,
    catalog_sync=catalog_sync,
    log_context="gateway OAuth fetch",
    stale_created_via_values={"MCP", "federation", "health_check", "manual_refresh", "oauth", "update"},
    skip_stale_cleanup=skip_stale_cleanup,
)
```

The `is_auth_code_gateway` term the other sites compute is unnecessary here, since
the function has already validated the grant type.

It would also help if the response distinguished "upstream reported N tools" from
"N rows written" — the current message made a destructive outcome look like a no-op.

---

### 📓 Logs / Error Output

Nothing is logged at WARNING or above; the deletion is silent. The only hint is an
INFO line from the *non*-deleting branch of `_reconcile_gateway_catalog`:

```
No new items to add to database
```

and the success response:

```json
{"success": true, "message": "Successfully fetched and created 0 tools"}
```

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Version or commit | `v1.0.6`, and still present on `main@7c042d21` |
| Runtime | Python 3.12, Gunicorn + Uvicorn |
| Container | `ghcr.io/ibm/mcp-context-forge` |

---

Verified by reading `main@7c042d21`: the call at `gateway_service.py:2302-2307`
still passes neither argument.

The reason this is easy to hit rather than theoretical: the natural operator
reaction to an OAuth gateway that has stopped working is to click *Fetch Tools* to
see whether it recovers. If the token is in the accepted-but-unauthorized state,
that click is what destroys the catalog.

Happy to open a PR with the guard plus a regression test if the approach above
looks right.

Linxiushen:
## Summary

- treat an empty catalog returned by the authorization-code OAuth fetch path as a probable authorization problem and preserve the gateway's existing tools, resources, prompts, metrics, and virtual-server associations
- restrict non-empty OAuth reconciliation to MCP-discovered entries so API/UI and legacy entries attached to the gateway are not pruned
- log a warning when the destructive cleanup guard is activated

## Testing

- `pytest tests/unit/mcpgateway/services/test_gateway_service.py -q` (full file passed, 1 existing skip)
- `pytest tests/unit/mcpgateway/services/test_gateway_service.py -k fetch_tools_after_oauth -q` (7 passed)
- `ruff check mcpgateway/services/gateway_service.py`
- `git diff --check`

Closes #6046

---
This change was developed with AI assistance (Claude); the code and tests were reviewed and verified locally before submission.

### DIFF
--- a/tests/unit/mcpgateway/services/test_gateway_service.py
+++ b/tests/unit/mcpgateway/services/test_gateway_service.py
@@ -16,6 +16,7 @@ from __future__ import annotations
 # Standard
 import asyncio
 from datetime import datetime, timedelta, timezone
+import logging
 import sys
 from types import SimpleNamespace
 from typing import TypeVar
@@ -4994,15 +4995,72 @@ async def test_fetch_tools_after_oauth_streamablehttp(gateway_service, monkeypat
 
 
 @pytest.mark.asyncio
+async def test_fetch_tools_after_oauth_empty_catalog_preserves_existing_items(gateway_service, monkeypatch, caplog):
+    gateway = MagicMock(spec=DbGateway)
+    gateway.id = "gw-1"
+    gateway.name = "gw"
+    gateway.oauth_config = {"grant_type": "authorization_code"}
+    gateway.transport = "streamablehttp"
+    gateway.tools = [SimpleNamespace(id=1, original_name="existing-tool", created_via="oauth")]
+    gateway.resources = [SimpleNamespace(id=2, uri="existing://resource", created_via="oauth")]
+    gateway.prompts = [SimpleNamespace(id=3, original_name="existing-prompt", created_via="oauth")]
+
+    db = MagicMock()
+    result = MagicMock()
+    result.scalar_one_or_none.return_value = gateway
+    db.execute.return_value = result
+
+    class DummyTokenStorage:
+        def __init__(self, _db):
+            self.db = _db
+
+        async def get_user_token(self, _gateway_id, _email):
+            return "token"
+
+        async def get_user_learned_audience(self, _gateway_id, _email):
+            return (None, None)
+
+    monkeypatch.setattr("mcpgateway.services.token_storage_service.TokenStorageService", DummyTokenStorage)
+    gateway_service.connect_to_streamablehttp_server = AsyncMock(return_value=({}, [], [], [], []))
+    gateway_service._update_or_create_tools = MagicMock(return_value=[])
+    gateway_service._update_or_create_resources = MagicMock(return_value=[])
+    gateway_service._update_or_create_prompts = MagicMock(return_value=[])
+
+    registry_cache = SimpleNamespace(
+        invalidate_tools=AsyncMock(),
+        invalidate_resources=AsyncMock(),
+        invalidate_prompts=AsyncMock(),
+    )
+    tool_lookup_cache = SimpleNamespace(invalidate_gateway=AsyncMock())
+    monkeypatch.setattr("mcpgateway.services.gateway_service._get_registry_cache", lambda: registry_cache)
+    monkeypatch.setattr("mcpgateway.services.gateway_service._get_tool_lookup_cache", lambda: tool_lookup_cache)
+    monkeypatch.setattr("mcpgateway.services.gateway_service.register_gateway_capabilities_for_notifications", MagicMock())
+    monkeypatch.setattr("mcpgateway.cache.admin_stats_cache.admin_stats_cache", SimpleNamespace(invalidate_tags=AsyncMock()))
+
+    with caplog.at_level(logging.WARNING):
+        await gateway_service.fetch_tools_after_oauth(db, "gw-1", "user@example.com")
+
+    assert gateway.tools[0].original_name == "existing-tool"
+    assert gateway.resources[0].uri == "existing://resource"
+    assert gateway.prompts[0].original_name == "existing-prompt"
+    assert db.execute.call_count == 1
+    assert "preserving existing items" in caplog.text
+
+
+@pytest.mark.asyncio
 async def test_fetch_tools_after_oauth_cleanup_and_adds_items(gateway_service, monkeypatch):
     gateway = MagicMock(spec=DbGateway)
     gateway.id = "gw-1"
     gateway.name = "gw"
     gateway.oauth_config = {"grant_type": "authorization_code"}
     gateway.transport = "sse"
-    gateway.tools = [SimpleNamespace(id=1, original_name="old-tool"), SimpleNamespace(id=2, original_name="keep-tool")]
-    gateway.resources = [SimpleNamespace(id=3, uri="old://res"), SimpleNamespace(id=4, uri="keep://res")]
-    gateway.prompts = [SimpleNamespace(id=5, original_name="old-prompt"), SimpleNamespace(id=6, original_name="keep-prompt")]
+    gateway.tools = [
+        SimpleNamespace(id=1, original_name="old-tool", created_via="oauth"),
+        SimpleNamespace(id=2, original_name="keep-tool", created_via="oauth"),
+        SimpleNamespace(id=7, original_name="ui-tool", created_via="ui"),
+    ]
+    gateway.resources = [SimpleNamespace(id=3, uri="old://res", created_via="oauth"), SimpleNamespace(id=4, uri="keep://res", created_via="oauth")]
+    gateway.prompts = [SimpleNamespace(id=5, original_name="old-prompt", created_via="oauth"), SimpleNamespace(id=6, original_name="keep-prompt", created_via="oauth")]
     gateway.capabilities = {}
     gateway.last_seen = None
 
@@ -5047,7 +5105,7 @@ async def test_fetch_tools_after_oauth_cleanup_and_adds_items(gateway_service, m
     result_data = await gateway_service.fetch_tools_after_oauth(db, "gw-1", "user@example.com")
 
     assert result_data["capabilities"]["resources"] is True
-    assert len(gateway.tools) == 1
+    assert {tool.original_name for tool in gateway.tools} == {"keep-tool", "ui-tool"}
     assert len(gateway.resources) == 1
     assert len(gateway.prompts) == 1
 
--- a/mcpgateway/services/gateway_service.py
+++ b/mcpgateway/services/gateway_service.py
@@ -2299,11 +2299,21 @@ class GatewayService(BaseService):  # pylint: disable=too-many-instance-attribut
                 prompts=prompts,
                 created_via="oauth",
             )
+
+            skip_stale_cleanup = not tools and not resources and not prompts
+            if skip_stale_cleanup:
+                logger.warning("Empty catalog from auth_code gateway %s during OAuth fetch, preserving existing items", gateway.name)
+
+            # Only prune entries that came from MCP discovery. API/UI and legacy
+            # entries can share the gateway but are not authoritative upstream data.
+            mcp_created_via_values = {"MCP", "federation", "health_check", "manual_refresh", "oauth", "update"}
             reconcile_result = self._reconcile_gateway_catalog(
                 db,
                 gateway=gateway,
                 catalog_sync=catalog_sync,
                 log_context="gateway OAuth fetch",
+                stale_created_via_values=mcp_created_via_values,
+                skip_stale_cleanup=skip_stale_cleanup,
             )
 
             # Update gateway capabilities and last_seen


tests/unit/mcpgateway/services/test_gateway_service.py
```diff
@@ -4993,16 +4994,73 @@ async def get_user_learned_audience(self, _gateway_id, _email):
...
+    gateway.resources = [SimpleNamespace(id=2, uri="existing://resource", created_via="oauth")]
+    gateway.prompts = [SimpleNamespace(id=3, original_name="existing-prompt", created_via="oauth")]
+
+    db = MagicMock()
+    result = MagicMock()
+    result.scalar_one_or_none.return_value = gateway
+    db.execute.return_value = result
+
+    class DummyTokenStorage:
+        def __init__(self, _db):
```
ja8zyjits: `DummyTokenStorage.__init__` lacks `user_context=None` — `TypeError` at call time

tests/unit/mcpgateway/services/test_gateway_service.py
```diff
@@ -4993,16 +4994,73 @@ async def get_user_learned_audience(self, _gateway_id, _email):
...
+    gateway.oauth_config = {"grant_type": "authorization_code"}
+    gateway.transport = "streamablehttp"
+    gateway.tools = [SimpleNamespace(id=1, original_name="existing-tool", created_via="oauth")]
+    gateway.resources = [SimpleNamespace(id=2, uri="existing://resource", created_via="oauth")]
+    gateway.prompts = [SimpleNamespace(id=3, original_name="existing-prompt", created_via="oauth")]
+
+    db = MagicMock()
+    result = MagicMock()
+    result.scalar_one_or_none.return_value = gateway
+    db.execute.return_value = result
```
ja8zyjits:
```python
result = MagicMock()
result.scalar_one_or_none.return_value = gateway  # spec=DbGateway
db.execute.return_value = result
```

Second `db.execute` call (line 2354, `select(DbEmailUser)`) returns same mock. `scalar_one_or_none()` returns `DbGateway` mock. Line 2355 accesses `user.is_admin` — `DbGateway` has no `is_admin`. Python raises `AttributeError` before caplog assertions.

Fix — use `side_effect` per sibling test `test_fetch_tools_after_oauth_streamablehttp`
```python
mock_user = MagicMock()
mock_user.is_admin = False
gateway_result = MagicMock()
gateway_result.scalar_one_or_none.return_value = gateway
user_result = MagicMock()
user_result.scalar_one_or_none.return_value = mock_user
db.execute.side_effect = [gateway_result, user_result]
```

tests/unit/mcpgateway/services/test_gateway_service.py
```diff
@@ -4993,16 +4994,73 @@ async def get_user_learned_audience(self, _gateway_id, _email):
...
+    monkeypatch.setattr("mcpgateway.services.gateway_service.register_gateway_capabilities_for_notifications", MagicMock())
+    monkeypatch.setattr("mcpgateway.cache.admin_stats_cache.admin_stats_cache", SimpleNamespace(invalidate_tags=AsyncMock()))
+
+    with caplog.at_level(logging.WARNING):
+        await gateway_service.fetch_tools_after_oauth(db, "gw-1", "user@example.com")
+
+    assert gateway.tools[0].original_name == "existing-tool"
+    assert gateway.resources[0].uri == "existing://resource"
+    assert gateway.prompts[0].original_name == "existing-prompt"
+    assert db.execute.call_count == 1
```
ja8zyjits: — `db.execute.call_count == 1` is wrong
`fetch_tools_after_oauth` calls `db.execute` at least twice:
1. `gateway_service.py:2323` — `select(DbGateway)` gateway lookup
2. `gateway_service.py:2354` — `select(DbEmailUser)` is_admin lookup

Assertion `assert db.execute.call_count == 1` fails once `DummyTokenStorage.__init__` is fixed and test reaches this line.

**Fix:** Remove assertion or change to `assert db.execute.call_count >= 2`.

ja8zyjits:
Hi @Linxiushen 
please check the comments
You may have to do rebase

Linxiushen:
Hi @ja8zyjits, thanks for picking this one up.

**"please check the comments" — I can't find them**

I'd rather ask than guess at the wrong points: on this PR I only see your message. The Conversation tab has no other comment, Files changed has no inline comments, and the API agrees — `/pulls/6049/reviews` and `/pulls/6049/comments` are both empty, with 0 review threads. The linked issue #6046 has no comments either, beyond @tw-360vier's original report.

Could you point me at where they are? Possibilities I can think of:

- a review that's still pending and not submitted — those aren't visible from the outside;
- a different PR: if you meant #6050, @gcgoncalves commented there earlier today and I've replied and pushed a fix (`3d298cd`);
- CI: the Checks tab here shows only DCO, but under Actions there are 13 `pull_request` workflow runs for `e147e4b` dated 2026-08-05 that all ended as "failure" with no jobs recorded — they failed at startup, so there are no logs or annotations on them to read. If that red is what you meant, pushing the updated branch should produce a real run this time (the push to #6050 today started 14 runs that executed normally), and I'll fix whatever they report.

Whatever it is, send it over and I'll work through it point by point.

**Bringing the branch up to date — done**

The branch is now up to date with current `main` (`1950bcb`). Both touched files auto-merged with no conflicts. I used a merge rather than a rebase so the original commit isn't rewritten and no force-push is needed; the merge base becomes `1950bcb`, so Files changed still lists only the two files rather than dragging in unrelated history. If you'd rather have a linear, single-commit branch, just say so and I'll rebase and force-push instead — I checked that the rebase is conflict-free as well.

**The bug is still present on current `main`**

In `fetch_tools_after_oauth`, the `_reconcile_gateway_catalog()` call still passes neither `skip_stale_cleanup` nor `stale_created_via_values`, so an empty upstream `tools/list` still deletes the gateway's catalog. I checked this on the updated branch by reverting only the source change: `test_fetch_tools_after_oauth_empty_catalog_preserves_existing_items` then fails with `IndexError` because `gateway.tools` has been emptied, and `test_fetch_tools_after_oauth_cleanup_and_adds_items` fails too. With the guard in place both pass.

**One follow-up commit on top of the merge**

The merge on its own leaves the branch red, so this isn't just polish: `fetch_tools_after_oauth` gained an `EmailUser` lookup upstream (to build `user_context` for `TokenStorageService`), and the regression test's single mocked `db.execute` result made that lookup return the gateway mock, failing with `AttributeError: Mock object has no attribute 'is_admin'`. The follow-up commit therefore:

- mocks the gateway and user lookups separately, and lets the dummy token storage accept `user_context`;
- replaces `assert db.execute.call_count == 1` with an assertion that no `DELETE` statement is issued — the call count is no longer a proxy for "no stale cleanup ran" now that the function issues a second SELECT;
- routes `gateway.name` in the new warning through `SecurityValidator.sanitize_log_message()`, matching the log hygiene the rest of `fetch_tools_after_oauth` adopted upstream in the meantime.

**Re-run on the updated branch** (Python 3.12):

- `pytest tests/unit/mcpgateway/services/test_gateway_service.py` → 397 passed, 1 skipped (the pre-existing "Yet to implement" skip)
- `pytest` over `tests/unit/mcpgateway/routers/test_oauth_router.py` and `tests/unit/mcpgateway/services/{test_gateway_auto_refresh,test_gateway_service_extended,test_gateway_service_health_oauth,test_gateway_service_helpers,test_gateway_service_oauth_comprehensive}.py` → 501 passed
- `pylint mcpgateway/services/gateway_service.py --rcfile=.pylintrc.mcpgateway --fail-on E` (3.3.9, the pin in `lint.yml`) → 10.00/10
- `ruff check` and `ruff format --check` (0.15.16, the rev pinned in `.pre-commit-config.yaml`) on the same file → clean

Happy to reshape the guard if you'd prefer a different form — for example folding `mcp_created_via_values` into a module-level constant shared with the refresh path, which is the one bit of duplication this leaves behind.

ja8zyjits (changes requested):
LGTM

ja8zyjits:
@Linxiushen 
My bad, i didnt post them, they were still in pending stage

tests/unit/mcpgateway/services/test_gateway_service.py
Linxiushen: Fixed in e3b524d: `DummyTokenStorage.__init__` is now `(self, _db, user_context=None)` and stores it. Thanks — this was exactly right.

tests/unit/mcpgateway/services/test_gateway_service.py
Linxiushen: Fixed in e3b524d, along the lines you suggested. I used a `side_effect` callable rather than a list so the test can also record the statements: `mock_execute` returns `gateway_result` then `user_result` (the latter with `is_admin = False`) and appends every statement to `executed_statements`. Same effect as the `side_effect=[gateway_result, user_result]` form from `test_fetch_tools_after_oauth_streamablehttp`, plus it feeds the DELETE assertion below.

tests/unit/mcpgateway/services/test_gateway_service.py
Linxiushen: Agreed, and fixed in e3b524d. Rather than loosening it to `>= 2`, the assertion now checks the thing it was standing in for directly: `assert [s for s in executed_statements if isinstance(s, Delete)] == []`. That stays correct regardless of how many SELECTs the function grows later.

Linxiushen:
Thanks @ja8zyjits, and no problem at all about the pending state — that explains it.

All three points were already addressed in e3b524d, which I pushed earlier today when bringing the branch up to date with `main`. Upstream had added the `select(DbEmailUser)` lookup in the six weeks this PR sat, which broke the test in exactly the ways you describe, so I hit the same three problems:

- `DummyTokenStorage.__init__` now takes `user_context=None`.
- `db.execute` now returns the gateway result and then a user result with `is_admin = False`, via a `side_effect` callable (the callable rather than a list so it can also record the statements).
- `assert db.execute.call_count == 1` is gone. It is now `assert [s for s in executed_statements if isinstance(s, Delete)] == []`, which asserts the actual invariant instead of using the call count as a proxy.

Current state on the branch: `pytest tests/unit/mcpgateway/services/test_gateway_service.py` gives 397 passed, 1 skipped (the pre-existing "Yet to implement" skip), and the six related OAuth/gateway test files give 501 passed. Reverting just the source change makes both `fetch_tools_after_oauth` tests fail again, so the guard is what they are testing. `pylint` on the touched file is 10.00/10.

Your review shows as Changes requested with a body of "LGTM" — I read that as the pending comments having been the substance. If there is anything left you would like changed, please say and I will turn it around.

tests/unit/mcpgateway/services/test_gateway_service.py
```diff
@@ -5253,16 +5255,96 @@ async def get_user_learned_audience(self, _gateway_id, _email):
     assert "capabilities" in result_data
 
 
+@pytest.mark.asyncio
+async def test_fetch_tools_after_oauth_empty_catalog_preserves_existing_items(gateway_service, monkeypatch, caplog):
```
ja8zyjits, line 5259: **Problem:** `skip_stale_cleanup = not tools and not resources and not prompts` triggers only when all three empty. Partial-empty responses (e.g. tools present, resources absent) run stale cleanup guarded solely by `stale_created_via_values`.
- **Fix:** Add unit test covering partial-empty response path during OAuth catalog fetch.

ja8zyjits (changes requested):
Just one more change and we are good to merge this.

ja8zyjits:
@Linxiushen sorry for the LGTM confusion.
Please address the new comments and we are good to merge this.

Thanks for the contribution

ja8zyjits:
@Linxiushen Please hold of the fixes.

Main is frozen. WIll inform you to rebase and fix after main is open again.

Thanks for your patience.

Regards,
Jitesh

tests/unit/mcpgateway/services/test_gateway_service.py
Linxiushen: Written and passing locally as `test_fetch_tools_after_oauth_partial_catalog_prunes_only_mcp_discovered_items` — not pushed, since main is frozen; it will land with the rebase once you give the word.

The mocked upstream returns one tool and no resources or prompts, so `skip_stale_cleanup` stays False and stale cleanup runs across all three categories. The test asserts:

- the "preserving existing items" warning is not emitted, i.e. the empty-catalog guard stays reserved for the all-empty response;
- stale MCP-discovered rows are pruned in every category, including resources and prompts that came back empty;
- rows attached through the API/UI (`created_via` "api" / "ui") survive untouched;
- the emitted DELETE statements target exactly the stale MCP-discovered ids and nothing else.

Treating a partial response as authoritative is deliberate: a server that answers `tools/list` while genuinely having no resources or prompts is a normal shape, not an authorization failure, so only `stale_created_via_values` guards that path. The two tests now pin the two arguments independently - removing `stale_created_via_values` from the `_reconcile_gateway_catalog` call fails the new partial test, and removing `skip_stale_cleanup` fails `test_fetch_tools_after_oauth_empty_catalog_preserves_existing_items`. I verified both by reverting each argument in turn.

`pytest tests/unit/mcpgateway/services/test_gateway_service.py` gives 398 passed, 1 skipped (the pre-existing "Yet to implement" skip), and the six related OAuth/gateway test files give 501 passed.

Linxiushen:
Understood, thanks for the heads-up - nothing pushed, and I will wait for your go-ahead before touching the branch.

The work is ready locally for when main reopens:

- the partial-empty test you asked for is written and green; details are in the thread on your review comment.
- `.secrets.baseline` needed a refresh. The only red check on the current head is Pre-commit Checks, and it fails solely because IBM Detect Secrets rewrites two `line_number` fields for `tests/unit/mcpgateway/services/test_gateway_service.py`, which this PR shifts. Every other hook in that job passes. `pre-commit run detect-secrets --all-files` is clean with the updated baseline.

One thing worth flagging for the rebase: `.secrets.baseline` is the only file that conflicts with current main, because main regenerates it frequently. When you give the word I will rebase onto main, regenerate the baseline as the final step so it lands with current line numbers, rerun the gateway and OAuth suites, and force-push.

ja8zyjits:
@Linxiushen Main is open again.
please rebase and let me know

Regards,
Jitesh

PR #6050, opened 2026-08-05
fix(admin): preserve OAuth credentials on edit

### LINKED ISSUE #6048: [UI]: Editing an OAuth gateway silently drops its stored client_secret
### 🐞 Bug Summary

Editing an `authorization_code` OAuth gateway in the admin UI silently **deletes its
stored `client_secret`**, breaking every subsequent authorization with a `401` from
the upstream token endpoint.

The edit form does not repopulate the secret field — correctly, since you should not
echo a secret back to the browser — so a normal edit submits an empty
`oauth_client_secret`. But the handler rebuilds `oauth_config` from scratch and only
includes the key when a non-empty value was submitted, so the stored secret is
dropped rather than preserved.

`_assemble_oauth_config_from_fields()` (`admin.py:947`) starts from `oauth_config = {}`
and at `admin.py:1004-1009` does:

```python
if oauth_client_secret:
    if encrypt_secret:
        encryption = get_encryption_service(settings.auth_encryption_secret)
        oauth_config["client_secret"] = await encryption.encrypt_secret_async(oauth_client_secret)
    else:
        oauth_config["client_secret"] = oauth_client_secret
```

`admin_edit_gateway()` calls it at `admin.py:13124` and persists the result. It does
load the current row at `admin.py:13133`, but only to read `team_id` — the existing
secret is never carried over.

The operator usually cannot recover by retyping, because with DCR the secret was
issued to the gateway and never shown to a human.

The same reasoning applies to `oauth_config["password"]` for the password grant, and
the same helper is used by the A2A edit paths (`admin.py:16137`, `admin.py:16362`).

### 🧩 Affected Component

- [ ] `mcpgateway` - API
- [ ] Other (explain below)

---

### 🔁 Steps to Reproduce

1. Create a gateway with `grant_type: authorization_code` and a `client_secret` (ours came from DCR, so `registered_oauth_clients` was populated too).
2. Complete an OAuth authorization successfully — confirming the secret is present and decryptable.
3. In the admin UI, edit that gateway and change something entirely unrelated. We changed only **Transport** from `SSE` to `Streamable HTTP`. Leave the client secret field untouched (blank, as rendered). Save.
4. Attempt another OAuth authorization.

Observed on our deployment:

```sql
select name, length(oauth_config::jsonb->>'client_secret'), updated_at
  from gateways where name = 'wordpress-...';
-- client_secret length: NULL   updated_at: 14:57:57
```

A successful authorization at 14:50 (secret present), the UI edit at 14:57:57, and
every authorization afterwards failing — with the key simply absent from
`oauth_config`.

---

### 🤔 Expected Behavior

An edit that does not touch the secret field should leave the stored secret intact.

The usual pattern is to distinguish "not submitted" from "submitted empty". On
update, when `oauth_client_secret` is empty, carry the existing value over rather
than omitting the key:

```python
if not oauth_client_secret and existing_oauth_config and existing_oauth_config.get("client_secret"):
    oauth_config["client_secret"] = existing_oauth_config["client_secret"]
```

Since `_assemble_oauth_config_from_fields()` is shared between create and edit, this
probably wants an optional `existing: Optional[dict]` parameter that the edit call
sites pass and the create call sites do not. `admin_edit_gateway()` already has the
row in hand at `admin.py:13133`.

Rendering a placeholder (e.g. `••••••`) in the form and treating "unchanged
placeholder" as "keep existing" would also work, and would make the behaviour
visible to the operator.

Whatever the mechanism, it would help to keep an explicit way to *clear* a secret,
so the fix does not remove that capability.

---

### 📓 Logs / Error Output

After the edit, on every authorization attempt:

```
ERROR mcpgateway.routers.oauth_router - OAuth callback failed: Failed to exchange code for token
after 3 attempts: Client error '401 Unauthorized' for url 'https://www.example.com/wp-json/mcp-oauth/v1/token'
```

The UI shows:

```
❌ OAuth Authorization Failed
Error: Failed to exchange code for token after 3 attempts: Client error '401 Unauthorized' ...
Please check your OAuth configuration and try again.
```

Nothing indicates that the secret was removed, and "check your OAuth configuration"
points the operator at their identity provider rather than at the edit they just
made. That misdirection is most of the cost here.

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Version or commit | `main@6ecace92`, and still present on `main@7c042d21` |
| Runtime | Python 3.12, Gunicorn + Uvicorn |
| Container | `ghcr.io/ibm/mcp-context-forge` |

---

Detection query for anyone who may already be affected:

```sql
select name, length(oauth_config::jsonb->>'client_secret')
  from gateways
 where oauth_config::jsonb->>'grant_type' = 'authorization_code';
```

Recovery, where DCR was used, is to copy the ciphertext back from
`registered_oauth_clients.client_secret_encrypted` for the matching issuer — both
columns are encrypted with the same `AUTH_ENCRYPTION_SECRET` and read by the same
`EncryptionService`, so it can be copied verbatim. Worth guarding on `client_id`
equality and on the value decrypting before writing.

One wrinkle if you add such a repair path: `gateways.oauth_config["issuer"]` and
`registered_oauth_clients.issuer` can differ by a trailing slash
(`https://www.example.com/` vs `https://www.example.com`), so matching needs
normalising.

Happy to open a PR for the preserve-on-empty behaviour if the `existing` parameter
approach looks right to you.

Linxiushen:
## Summary

- preserve the stored OAuth `client_secret` when Gateway or A2A edit forms leave the password input blank
- preserve the stored resource-owner password only while the edited configuration remains on the password grant
- reuse the already-loaded entity for both OAuth credential preservation and team preservation, without decrypting or re-encrypting existing ciphertext
- keep full `oauth_config` JSON submissions unchanged so API callers retain an explicit way to clear credentials

## Testing

- `pytest tests/unit/mcpgateway/test_admin.py -k 'admin_edit_gateway or admin_edit_a2a_agent' -q` (22 passed)
- `ruff check mcpgateway/admin.py`
- `ruff format --check mcpgateway/admin.py tests/unit/mcpgateway/test_admin.py`
- `git diff --check`

Closes #6048

---
This change was developed with AI assistance (Claude); the code and tests were reviewed and verified locally before submission.

### DIFF
--- a/mcpgateway/admin.py
+++ b/mcpgateway/admin.py
@@ -944,7 +944,13 @@ def _form_team_id(form: Any) -> Optional[str]:
     return str(raw).strip() or None
 
 
-async def _assemble_oauth_config_from_fields(fields: Any, *, encrypt_secret: bool, include_resource: bool = True) -> Optional[Dict[str, Any]]:
+async def _assemble_oauth_config_from_fields(
+    fields: Any,
+    *,
+    encrypt_secret: bool,
+    include_resource: bool = True,
+    existing_config: Optional[Dict[str, Any]] = None,
+) -> Optional[Dict[str, Any]]:
     """Assemble an ``oauth_config`` dict from individual OAuth form/JSON fields.
 
     Shared by all four admin OAuth form handlers (gateway create/edit, A2A
@@ -961,12 +967,15 @@ async def _assemble_oauth_config_from_fields(fields: Any, *, encrypt_secret: boo
     * ``encrypt_secret=True`` encrypts ``client_secret`` before storage
       (UI edit/add handlers); ``False`` stores it as submitted (the gateway
       create path, where encryption happens downstream in the service layer).
+    * On edit, ``existing_config`` preserves stored credentials when password
+      inputs are intentionally left blank rather than echoing secrets to the UI.
 
     Args:
         fields: Mapping with ``.get()`` (a form dict or parsed JSON body)
             containing the ``oauth_*`` keys.
         encrypt_secret: Whether to encrypt a submitted ``client_secret``.
         include_resource: Whether to read and emit ``oauth_resource``.
+        existing_config: Existing stored OAuth configuration for edit requests.
 
     Returns:
         Assembled ``oauth_config`` dict, or ``None`` when no meaningful OAuth
@@ -1007,10 +1016,14 @@ async def _assemble_oauth_config_from_fields(fields: Any, *, encrypt_secret: boo
             oauth_config["client_secret"] = await encryption.encrypt_secret_async(oauth_client_secret)
         else:
             oauth_config["client_secret"] = oauth_client_secret
+    elif existing_config and existing_config.get("client_secret"):
+        oauth_config["client_secret"] = existing_config["client_secret"]
     if oauth_username:
         oauth_config["username"] = oauth_username
     if oauth_password:
         oauth_config["password"] = oauth_password
+    elif oauth_grant_type == "password" and existing_config and existing_config.get("password"):
+        oauth_config["password"] = existing_config["password"]
     if oauth_audience:
         oauth_config["audience"] = oauth_audience
     if oauth_scopes_str:
@@ -13103,6 +13116,11 @@ async def admin_edit_gateway(
         else:
             passthrough_headers = None
 
+        existing_gateway = db.get(DbGateway, gateway_id)
+        existing_oauth_config = getattr(existing_gateway, "oauth_config", None)
+        if not isinstance(existing_oauth_config, dict):
+            existing_oauth_config = None
+
         # Parse OAuth configuration - support both JSON string and individual form fields
         oauth_config_json = str(form.get("oauth_config"))
         oauth_config: Optional[dict[str, Any]] = None
@@ -13121,7 +13139,7 @@ async def admin_edit_gateway(
 
         # Option 2: Assemble from individual UI form fields
         if not oauth_config:
-            oauth_config = await _assemble_oauth_config_from_fields(form, encrypt_secret=True)
+            oauth_config = await _assemble_oauth_config_from_fields(form, encrypt_secret=True, existing_config=existing_oauth_config)
             if oauth_config:
                 LOGGER.info(f"✅ Assembled OAuth config from UI form fields (edit): grant_type={oauth_config.get('grant_type')}, issuer={oauth_config.get('issuer')}")
 
@@ -13130,7 +13148,6 @@ async def admin_edit_gateway(
         # Without this guard, verify_team_for_user() falls back to the user's
         # personal team, silently reassigning the gateway on every edit.
         if not team_id:
-            existing_gateway = db.get(DbGateway, gateway_id)
             existing_team = getattr(existing_gateway, "team_id", None) if existing_gateway else None
             if isinstance(existing_team, str) and existing_team:
                 team_id = existing_team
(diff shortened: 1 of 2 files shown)


gcgoncalves:
Testing this, I noticed you cannot unset the `client_secret` if you change the auth type. The `password` field has an "unsetter" if you chane to a type that doesn't use it, I think this pattern could be reused.

Linxiushen:
Thanks for testing this — you're right, and it's a regression this PR introduced. Fixed in 3d298cd.

**What I reproduced**

The preserve-on-blank branch I added for `client_secret` was unconditional, so once a secret was stored there was no longer any way to clear it from the admin UI. It's a bit worse than "blank keeps the old value": the OAuth inputs stay in the DOM when the form hides them (`display: none`), so they keep posting their previously populated values even after you switch **Authentication Type** to Basic/Bearer/Custom Headers/Query Parameter. `_assemble_oauth_config_from_fields()` therefore still assembles an `oauth_config`, and my `elif` pinned the old `client_secret` into it for good.

**The pattern you pointed at**

The `password` branch in the same helper was already gated — it only carries the stored resource-owner password over while the submitted configuration is still on the `password` grant, so changing grant type unsets it. I've lifted that into two explicit flags and applied them to both credentials:

```python
uses_oauth_credentials = auth_type in ("", "oauth")
uses_password_grant = uses_oauth_credentials and oauth_grant_type == "password"
```

- `client_secret` is preserved only while `uses_oauth_credentials`, so switching the entity to a non-OAuth auth type now clears it.
- `password` keeps its grant-type gate and additionally picks up the auth-type gate, which it was missing: previously `auth_type=basic` together with `oauth_grant_type=password` still preserved the stored password.

An empty `auth_type` still counts as OAuth here, because both edit handlers fall back to `auth_type="oauth"` whenever an OAuth config was assembled; treating `""` as non-OAuth would break the ordinary preserve-on-blank case this PR is about.

**Scope**

The gate lives in the shared `_assemble_oauth_config_from_fields()` helper, so it covers both edit handlers this PR touches (`admin_edit_gateway`, `admin_edit_a2a_agent`). The two create call sites pass no `existing_config`, so they are unchanged. Full `oauth_config` JSON submissions still bypass the helper entirely, so API callers keep their explicit way to clear credentials.

**Tests**

Added to `tests/unit/mcpgateway/test_admin.py`:

- `test_admin_edit_gateway_oauth_auth_type_switch_clears_credentials` — switching to `basic` drops both stored credentials and never reaches the encryption service; flipping back to `oauth` with blank inputs still preserves them; a submitted secret still replaces the stored one.
- `test_admin_edit_a2a_agent_oauth_auth_type_switch_clears_credentials` — the same for the A2A edit handler.

Both fail on the previous commit (`assert 'client_secret' not in {..., 'client_secret': 'encrypted-client-secret', ...}`) and pass with the new one.

Ran locally:

- `pytest tests/unit/mcpgateway/test_admin.py` — 1243 passed
- `ruff check mcpgateway/admin.py` and `ruff format --check mcpgateway/admin.py tests/unit/mcpgateway/test_admin.py` — clean

**One thing I deliberately left alone**

`protect_oauth_config_for_storage()` in the service layer already implements a masked-placeholder contract (`settings.masked_auth_value` in → existing ciphertext preserved), and the admin UI already renders that mask for the basic, bearer, custom-header and query-param fields — just not for `oauth_client_secret`, which is explicitly blanked in `gateways.js`. Rendering the mask into the OAuth secret field would make "keep" versus "clear" visible to the operator instead of implicit, and would reuse a mechanism that already exists on the storage side. That is a front-end change with a wider blast radius than this fix, so I kept it out here — happy to do it in this PR or as a follow-up if you prefer that shape.

gcgoncalves (approved):

PR #6078, opened 2026-08-05
fix: warn at startup when CSRF cookie or header names are overridden

madhu-mohan-jaishankar:
# Pull Request

## 🔗 Related Issue

Closes [#523](https://github.ibm.com/contextforge-org/internal_issues/issues/523)
---

## 📝 Summary

`CSRF_COOKIE_NAME` and `CSRF_TOKEN_NAME` govern `CSRFMiddleware` only. Every other
CSRF consumer hardcodes the default names independently:

- `enforce_admin_csrf` (`mcpgateway/admin.py`)
- `enforce_fetch_tools_csrf` (`mcpgateway/routers/oauth_router.py`)
- the Admin UI JavaScript (`mcpgateway/admin_ui/{utils,auth,formHandlers,llmModels}.js`)
- the server-rendered login / password / admin templates

Overriding either setting therefore desynchronizes the middleware from every client,
and the resulting failure is easy to miss. CSRF tokens are a deterministic HMAC over
`{email}:{jti}:{time_window}`, so the cookie written at login and the one written by
the admin page normally hold the *same* value and writes succeed. They diverge at a
token-window boundary, at which point browser writes outside `/admin` begin returning
`403 CSRF_TOKEN_INVALID` intermittently. The operator sees a flaky UI, not a
misconfiguration.

This PR emits a startup warning naming both the offending setting and the symptom, so
the mismatch is visible at boot instead of at request time. The check is gated on
`CSRF_ENABLED` so deployments with CSRF disabled stay quiet.

**Scope note — this is a stopgap, not the root-cause fix.** It makes an unsupported
configuration fail loudly; it does not make the setting work. The durable fix is to
have the Admin UI and the per-route CSRF dependencies read the configured names
instead of hardcoding them, which requires client-side changes plus a Vite bundle
rebuild and is intentionally deferred to a separate PR.

**This is not a CSRF bypass.** Every enforcement path fails closed —
`csrf_middleware.py:182-191`, `admin.py:1868-1884`, and `oauth_router.py:100-105` all
return 403 on a missing or mismatched cookie. `/admin` is unaffected because `admin.py`
both sets and validates under the same constant, so it is internally self-consistent.
Cross-origin JavaScript cannot read any cookie, so renaming it neither adds nor removes
protection against a forged request. The impact is configuration integrity and
availability, not a weakening of CSRF defenses.

**Default deployments are unaffected** — `csrf_cookie_name` defaults to exactly the
hardcoded literal, and existing tests already pin that (`test_config.py:75-153`).

---

## 📏 Reviewability

- [ ] The linked issue is not labeled `triage` — _no public issue linked; see above_

---

## 🏷️ Type of Change

- [x] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

**Changed files:** `mcpgateway/config.py` (+25), `tests/unit/mcpgateway/test_config.py` (+47)

| Check | Command | Status |
|---|---|---|
| Lint suite | `make lint` | ⬜ Not run — see targeted lint below |
| Unit tests | `make test` | ⬜ Not run — see targeted suites below |
| Coverage ≥ 80% | `make coverage` | ⬜ Not run |

Targeted runs that **were** executed:

| Check | Command | Status |
|---|---|---|
| Config + CSRF middleware | `pytest tests/unit/mcpgateway/test_config.py tests/unit/mcpgateway/middleware/test_csrf_middleware.py tests/unit/mcpgateway/middleware/test_admin_csrf_binding.py` | ✅ 256 passed |
| OAuth router CSRF | `pytest tests/unit/mcpgateway/routers/test_oauth_router.py` | ✅ 175 passed |
| CSRF service | `pytest tests/unit/mcpgateway/services/test_csrf_service.py` | ✅ 34 passed |
| New tests only | `pytest tests/unit/mcpgateway/test_config.py -k csrf` | ✅ 6 passed |
| Ruff | `ruff check mcpgateway/config.py tests/unit/mcpgateway/test_config.py` | ✅ No new findings |

**Ruff detail:** two `too-many-blank-lines` findings are reported in `test_config.py`.
Both are pre-existing at `HEAD` and in code this PR does not touch (verified by running
ruff against the `HEAD` versions of both files). No new findings are introduced.

**Manual verification** of the warning behavior across all three branches:

```python
Settings(_env_file=None)
# → no CSRF warning

Settings(_env_file=None, csrf_cookie_name="csrf_token", csrf_token_name="X-Probe")
# → ⚠️  CSRF CONFIGURATION WARNING: CSRF_COOKIE_NAME is set to 'csrf_token' but the
#    Admin UI and the per-route CSRF dependencies hardcode 'mcpgateway_csrf_token'.
#    Browser-based writes outside /admin will intermittently fail with 403
#    CSRF_TOKEN_INVALID. Set CSRF_COOKIE_NAME=mcpgateway_csrf_token (or unset it)
#    unless you have verified every client sends the custom name.
# → ⚠️  CSRF CONFIGURATION WARNING: CSRF_TOKEN_NAME is set to 'X-Probe' ...

Settings(_env_file=None, csrf_enabled=False, csrf_cookie_name="csrf_token")
# → no CSRF warning
```

Each of those three cases is covered by a test:

- `test_csrf_name_override_warns_at_startup`
- `test_csrf_default_names_do_not_warn`
- `test_csrf_name_override_silent_when_csrf_disabled`

**Regression risk:** the warning only fires when the names are non-default, which no
existing test configures. `tests/unit/mcpgateway/services/test_csrf_service.py` passes
`csrf_cookie_name="csrf_token"`, but via `Mock` objects rather than real `Settings`
instances, so the validator never runs there — confirmed passing.

---

## ✅ Checklist

- [ ] Code formatted (`make black isort pre-commit`) — _not run; `ruff check` and `ruff format --check` were run on the changed files and report no new findings_
- [x] Documentation updated (if applicable) — _no change needed; see Notes_

---

## 📓 Notes

### DIFF
--- a/mcpgateway/config.py
+++ b/mcpgateway/config.py
@@ -1601,12 +1601,19 @@ class Settings(BaseSettings):
         # setting desynchronizes the middleware from all of them, which surfaces as
         # intermittent 403 CSRF_TOKEN_INVALID on non-/admin browser writes rather
         # than as an obvious failure. Warn loudly at startup instead.
+        #
+        # Comparison is asymmetric by design. HTTP header names are
+        # case-insensitive (RFC 7230) and Starlette normalizes them, so
+        # CSRF_TOKEN_NAME=x-csrf-token is functionally identical to the default
+        # and must not warn. Cookie names are case-sensitive (RFC 6265), so a
+        # cookie name differing only in case is a genuine desync and must warn.
         if self.csrf_enabled:
-            for setting_name, configured, default in (
-                ("CSRF_COOKIE_NAME", self.csrf_cookie_name, "mcpgateway_csrf_token"),
-                ("CSRF_TOKEN_NAME", self.csrf_token_name, "X-CSRF-Token"),
+            for setting_name, configured, default, case_sensitive in (
+                ("CSRF_COOKIE_NAME", self.csrf_cookie_name, "mcpgateway_csrf_token", True),
+                ("CSRF_TOKEN_NAME", self.csrf_token_name, "X-CSRF-Token", False),
             ):
-                if configured != default:
+                differs = configured != default if case_sensitive else configured.casefold() != default.casefold()
+                if differs:
                     logger.warning(
                         "⚠️  CSRF CONFIGURATION WARNING: %s is set to %r but the Admin UI and the "
                         "per-route CSRF dependencies hardcode %r. Browser-based writes outside /admin "
--- a/mcpgateway/config.py
+++ b/mcpgateway/config.py
@@ -1594,6 +1594,31 @@ class Settings(BaseSettings):
         if not self.csrf_secret_key:
             self.csrf_secret_key = self.jwt_secret_key.get_secret_value()
 
+        # CSRF_COOKIE_NAME / CSRF_TOKEN_NAME govern CSRFMiddleware only. Every
+        # other consumer -- enforce_admin_csrf (admin.py), enforce_fetch_tools_csrf
+        # (routers/oauth_router.py), the Admin UI JavaScript, and the server-rendered
+        # login/password templates -- hardcodes the default names. Overriding either
+        # setting desynchronizes the middleware from all of them, which surfaces as
+        # intermittent 403 CSRF_TOKEN_INVALID on non-/admin browser writes rather
+        # than as an obvious failure. Warn loudly at startup instead.
+        if self.csrf_enabled:
+            for setting_name, configured, default in (
+                ("CSRF_COOKIE_NAME", self.csrf_cookie_name, "mcpgateway_csrf_token"),
+                ("CSRF_TOKEN_NAME", self.csrf_token_name, "X-CSRF-Token"),
+            ):
+                if configured != default:
+                    logger.warning(
+                        "⚠️  CSRF CONFIGURATION WARNING: %s is set to %r but the Admin UI and the "
+                        "per-route CSRF dependencies hardcode %r. Browser-based writes outside /admin "
+                        "will intermittently fail with 403 CSRF_TOKEN_INVALID. Set %s=%s (or unset it) "
+                        "unless you have verified every client sends the custom name.",
+                        setting_name,
+                        configured,
+                        default,
+                        setting_name,
+                        default,
+                    )
+
         # Validate header passthrough feature flag dependencies
         # Fail if sensitive passthrough is enabled without base feature
         if self.enable_sensitive_header_passthrough and not self.enable_header_passthrough:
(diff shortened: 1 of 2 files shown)


prakhar-singh1928:
2 Minor issues @madhu-mohan-jaishankar 
Case-sensitivity gap in CSRF_TOKEN_NAME comparison — The config default is "X-CSRF-Token" (mixed case) but both admin.py:1689 and oauth_router.py:54 hardcode "x-csrf-token" (lowercase). An operator who sets CSRF_TOKEN_NAME=x-csrf-token (lowercase, functionally identical since Starlette normalises headers) would get a spurious false-positive warning. The fix is either a .lower() comparison or changing the hardcoded default in the tuple to "x-csrf-token" to match what the constants actually read.

Missing header-name drift test — The existing tests [test_admin_csrf_cookie_name_matches_config_default](vscode-webview://13eaird7rnmfh30crbbhqon8ebmdqmuja1h198s5uiksobscjo6t/tests/unit/mcpgateway/test_config.py:109) and [test_oauth_router_csrf_cookie_name_matches_config_default](vscode-webview://13eaird7rnmfh30crbbhqon8ebmdqmuja1h198s5uiksobscjo6t/tests/unit/mcpgateway/test_config.py:133) pin the cookie constant. But admin.py and oauth_router.py also each hardcode ADMIN_CSRF_HEADER_NAME = "x-csrf-token", which is used directly in enforce_admin_csrf. No test pins those to settings.csrf_token_name — exactly the symmetrical gap the PR is designed to close for cookies.

madhu-mohan-jaishankar:
> 2 Minor issues @madhu-mohan-jaishankar Case-sensitivity gap in CSRF_TOKEN_NAME comparison — The config default is "X-CSRF-Token" (mixed case) but both admin.py:1689 and oauth_router.py:54 hardcode "x-csrf-token" (lowercase). An operator who sets CSRF_TOKEN_NAME=x-csrf-token (lowercase, functionally identical since Starlette normalises headers) would get a spurious false-positive warning. The fix is either a .lower() comparison or changing the hardcoded default in the tuple to "x-csrf-token" to match what the constants actually read.
> 
> Missing header-name drift test — The existing tests test_admin_csrf_cookie_name_matches_config_default and test_oauth_router_csrf_cookie_name_matches_config_default pin the cookie constant. But admin.py and oauth_router.py also each hardcode ADMIN_CSRF_HEADER_NAME = "x-csrf-token", which is used directly in enforce_admin_csrf. No test pins those to settings.csrf_token_name — exactly the symmetrical gap the PR is designed to close for cookies.

Thanks @prakhar-singh1928 
Both fixed in 50fea06cc.

1. Comparison is now case-insensitive for the header name only. Not a blanket .lower() — cookie names are case-sensitive, so a cookie differing only in case is a real desync and must still warn. Added a test pinning that asymmetry.

2. Added test_admin_csrf_header_name_matches_config_default and test_oauth_router_csrf_header_name_matches_config_default, compared casefolded since both constants are deliberately lowercase.

gandhipratik203 (approved):
Nice catch on the root cause, and the write-up of the intermittent failure mode is really clear.

One non-blocking note:

**`LOG_LEVEL=ERROR` drops the warning.**

`log_level` defaults to `ERROR` (`config.py:1910`) and sets the root logger level (`config.py:73-81`), so `logger.warning` is filtered in production. The tests miss it because `caplog.set_level` forces the threshold down.

Fix: `logger.error`, or `get_security_warnings()` (`config.py:1634`), which isn't level-gated.

prakhar-singh1928 (approved):

PR #5783, opened 2026-07-23
feat(plugins): support A2A agents in Vault plugin

popagruia:
# Pull Request

## 🔗 Related Issue

Closes # 

---

## 📝 Summary

The Vault plugin injected upstream auth tokens only on the MCP tool path
(`tool_pre_invoke`). A2A agent invocations fire `agent_pre_invoke` and publish target
metadata under the `a2a_agent` key, so a tagged A2A server was a no-op. This PR makes a
tagged A2A agent drive Vault token injection exactly like a tagged MCP gateway.

**Plugin (agent_pre_invoke support):**
- Add an `agent_pre_invoke` hook reading `a2a_agent` metadata (tag mode only).
- Extract shared token-injection logic into a helper used by both hooks, preserving the
  SECURITY-critical vault-header stripping on every path (match, no match, parse error,
  undeterminable system).
- Normalize both tag shapes: gateway dict tags (`{"id","label"}`) and A2A tags.
- `oauth2_config` system resolution stays gateway-only (it loads a Gateway row, which
  does not exist for A2A agents); documented as unsupported for A2A, which uses tag mode.
- Register `agent_pre_invoke` in the manifest and `plugins/config.yaml`.
- Add A2A unit tests and tag-normalization regression tests; update README.

**Two defects found via live E2E and fixed (required for the A2A path to actually work):**
- `PydanticA2AAgent.tags` was typed `List[str]`, but A2A agent tags are persisted as
  `List[Dict[str,str]]` (via `validate_tags_field`, same as gateways). The mismatch made
  plugin-metadata construction raise, the error was swallowed, and the plugin saw no tags
  ("System cannot be determined"). Widened the field to `List[Union[str, Dict[str,str]]]`.
- On the A2A invoke path, plugin-returned headers were merged with `dict.update()`
  (additive-only), so a plugin removing a header (the Vault plugin stripping
  `X-Vault-Tokens`) could not delete it from the outbound set — leaking the vault header
  upstream. Removals are now reconciled against the sanitized header set the plugin
  received (a plugin can remove only headers it was actually given).

**Deployment note (documented in README):** on the A2A path the agent's
`passthrough_headers` must whitelist `X-Vault-Tokens` (so the plugin receives it) and
`Authorization` (so the injected Bearer survives re-filtering), and the gateway must run
with `ENABLE_HEADER_PASSTHROUGH=true` + `ENABLE_SENSITIVE_HEADER_PASSTHROUGH=true`.

---

## 📏 Reviewability

[...]

PR #6079, opened 2026-08-05
feat(plugins): support A2A agents in Vault plugin

### LINKED ISSUE #6203: [BUG]: Vault plugin does not support A2A agents (agent_pre_invoke not implemented)
### 🐞 Bug Summary

The Vault plugin only injects upstream auth tokens on the MCP tool invocation path (`tool_pre_invoke`). A2A agent invocations go through a separate hook, `agent_pre_invoke`, which the plugin does not implement — so tagging an A2A agent with `system:<host>` / `AUTH_HEADER:<name>` has no effect. Vault-backed secret injection silently does not work for A2A agents.

---

### 🧩 Affected Component

- [x] `mcpgateway` - API
- [x] Other (explain below) — Plugin framework (`plugins/vault`)

---

### 🔁 Steps to Reproduce

1. Register an A2A agent tagged `system:github.com` (and, for PAT tokens, `AUTH_HEADER:<header>`), with `passthrough_headers` including `X-Vault-Tokens`.
2. Enable the Vault plugin (`plugins/vault/vault_plugin.py`) with `tool_pre_invoke` configured, as documented in `plugins/vault/README.md`.
3. Invoke the agent via `POST /a2a/{agent-name}/invoke` with an `X-Vault-Tokens` header carrying a secret for that system.
4. Observe the upstream A2A agent's received headers.

---

### 🤔 Expected Behavior

The plugin should resolve the A2A agent's tags the same way it resolves an MCP gateway's tags, inject the vault-provided secret as the `Authorization: Bearer` header (or the configured `AUTH_HEADER`), and strip `X-Vault-Tokens` before forwarding upstream — mirroring the existing MCP tool behavior.

---

### 📓 Logs / Error Output

No error is raised; the plugin simply never runs on this path because it registers no `agent_pre_invoke` hook. `X-Vault-Tokens` (if whitelisted in `passthrough_headers`) is forwarded to the upstream A2A agent unmodified, and no `Authorization` header is injected.

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Version or commit | `main@4a1e3eadb` |
| Runtime | Python 3.11 |
| Container | none |

---

Fix in progress: #6079 adds an `agent_pre_invoke` hook to the Vault plugin, extracts the shared token-injection logic so both hooks stay in sync (including the security-critical `X-Vault-Tokens` stripping on every code path), and fixes two related defects surfaced by live E2E testing (an `A2AAgent` tags type mismatch, and a header-removal reconciliation bug in the A2A invoke path that could otherwise leak `X-Vault-Tokens` upstream).

popagruia:
# Pull Request

## 🔗 Related Issue

Closes #6203

---

## 📝 Summary

The Vault plugin injected upstream auth tokens only on the MCP tool path
(`tool_pre_invoke`). A2A agent invocations fire `agent_pre_invoke` and publish target
metadata under the `a2a_agent` key, so a tagged A2A server was a no-op. This PR makes a
tagged A2A agent drive Vault token injection exactly like a tagged MCP gateway.

**Plugin (agent_pre_invoke support):**
- Add an `agent_pre_invoke` hook reading `a2a_agent` metadata (tag mode only).
- Extract shared token-injection logic into a helper used by both hooks, preserving the
  SECURITY-critical vault-header stripping on every path (match, no match, parse error,
  undeterminable system).
- Normalize both tag shapes: gateway dict tags (`{"id","label"}`) and A2A tags.
- `oauth2_config` system resolution stays gateway-only (it loads a Gateway row, which
  does not exist for A2A agents); documented as unsupported for A2A, which uses tag mode.
- Register `agent_pre_invoke` in the manifest and `plugins/config.yaml`.
- Add A2A unit tests and tag-normalization regression tests; update README.

**Two defects found via live E2E and fixed (required for the A2A path to actually work):**
- `PydanticA2AAgent.tags` was typed `List[str]`, but A2A agent tags are persisted as
  `List[Dict[str,str]]` (via `validate_tags_field`, same as gateways). The mismatch made
  plugin-metadata construction raise, the error was swallowed, and the plugin saw no tags
  ("System cannot be determined"). Widened the field to `List[Union[str, Dict[str,str]]]`.
- On the A2A invoke path, plugin-returned headers were merged with `dict.update()`
  (additive-only), so a plugin removing a header (the Vault plugin stripping
  `X-Vault-Tokens`) could not delete it from the outbound set — leaking the vault header
  upstream. Removals are now reconciled against the sanitized header set the plugin
  received (a plugin can remove only headers it was actually given).

**Deployment note (documented in README):** on the A2A path the agent's
`passthrough_headers` must whitelist `X-Vault-Tokens` (so the plugin receives it) and
`Authorization` (so the injected Bearer survives re-filtering), and the gateway must run
with `ENABLE_HEADER_PASSTHROUGH=true` + `ENABLE_SENSITIVE_HEADER_PASSTHROUGH=true`.

**New: bind a vault token to its registered destination (`mcpServer`).** A vault token
entry can now be an object (`{"secretValue": ..., "mcpServer": "<url>"}`) instead of a bare
string. When `mcpServer` is present, the plugin only injects the secret if it matches the
target's actual registered URL — `Gateway.url` for MCP tools, `A2AAgent.endpoint_url` for
A2A agents. This closes a replay gap: two destinations can share the same `system:<host>`
tag (e.g. two gateways both tagged `system:github.com`) while pointing at different URLs:
without binding, a token meant for one would be trusted for the other. On a mismatch (or
when the real destination can't be determined) the secret is withheld — the vault header
is still stripped, but no `Authorization`/custom header is injected, so a pre-existing
passthrough header on the request is left untouched instead of being overwritten with a
leaked secret. Legacy plain-string tokens are unaffected. URL comparison is normalized
(lowercased scheme/host, trailing slash ignored). Added `TestVaultPluginMcpServerBinding`
covering matching/mismatched/null-binding/undeterminable-destination cases for both hooks,
and a defense-in-depth strip of `X-Vault-Tokens` in `a2a_service.py` so the header can never
reach the downstream agent even if the plugin is disabled or errors out in permissive mode.

---

## 📏 Reviewability

---

## 🏷️ Type of Change

- [x] Bug fix
- [x] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

| Check                                   | Command                                                                                     | Status |
|-----------------------------------------|---------------------------------------------------------------------------------------------|--------|
| Vault plugin unit tests                 | `pytest tests/unit/mcpgateway/plugins/plugins/vault/test_vault_plugin.py`                   | ✅ 37 passed |
| A2A-affected suites (regression)        | `pytest tests/unit/mcpgateway/services/test_a2a_agent_invoke_hooks.py tests/unit/mcpgateway/test_a2a_passthrough_headers.py tests/unit/mcpgateway/test_a2a_plugin_header_security.py` | ✅ 87 passed |
| Live E2E (MCP tool + A2A agent paths)   | `pytest tests/e2e/test_vault_plugin_a2a_e2e.py`                                              | ✅ 2 passed |
| Lint (changed files)                    | `ruff check` on changed files                                                               | ✅ clean |

**Live E2E evidence.** `tests/e2e/test_vault_plugin_a2a_e2e.py` boots two local echo backends
and a gateway with the Vault plugin on both hooks, then drives real invocations carrying an
`X-Vault-Tokens` header. It asserts, for BOTH paths, that the vault token is injected as
`Authorization: Bearer` upstream and that `X-Vault-Tokens` is stripped (never forwarded):

```
tests/e2e/test_vault_plugin_a2a_e2e.py ..                                [100%]
============================== 2 passed in 7.59s ===============================
```

Upstream headers observed by the MCP echo backend (vault token injected, vault header gone):

```
headers: {..., 'authorization': 'Bearer tok-tool-e2e', ...} message: hi
# note: no 'x-vault-tokens' present
```

Also manually re-verified live against `plugins/vault/echo_mcp.py`'s new `whoami` tool (which
returns the received `Authorization` header directly in its JSON-RPC response, rather than only
logging it) and `plugins/vault/echo_a2a.py`'s `received_headers` response field, for both the
legacy plain-string token shape and the new `mcpServer`-bound shape (match and mismatch cases)
— see `plugins/vault/README.md` → "E2E Throwaway Test Harness".

**Full-suite note.** `make test` was run against the whole tree: 21770 passed, 819 skipped,
2 xfailed, 10 failed. The 10 failures (`test_rbac_router.py`, `test_tokens.py`,
`test_error_formatter.py`) are unrelated to this PR — verified by running the same test IDs
against `main`, where they fail identically (same assertion diffs), and none of those three
files appear in this branch's diff against `main`.

---

## ✅ Checklist

- [x] Code follows the project style guidelines
- [x] Self-review of the changes completed
- [x] Comments added for non-obvious logic (shared helper, header-removal reconciliation)
- [x] Documentation updated (`plugins/vault/README.md`)
- [x] Tests added/updated (unit + tag-normalization + live E2E)
- [x] All new tests pass locally
- [x] No new lint warnings on changed files
- [x] Commits are signed (DCO)
- [x] PR title follows Conventional Commits (`feat(plugins): ...` / `fix(a2a): ...`)
- [ ] Coverage ≥ 80% (`make coverage`) — run in CI

---

## Notes for reviewers

- The `agent_pre_invoke` path follows the existing tool path's use of `payload.headers`.
  `cpex` emits a deprecation warning suggesting `extensions.http.headers`; migrating both
  paths together is a suggested follow-up rather than part of this change.
- On the A2A path, `oauth2_config` system resolution is intentionally unsupported (it
  requires a Gateway DB row that agents do not have); A2A uses tag mode.

---
_In-repo version of #5783 (moved into IBM/mcp-context-forge per maintainer request, so CI runs with repo secrets). Supersedes #5783._


popagruia:
Superseded by #6079 — moved in-repo per maintainer request.

vishu-bh (changes requested):
Thanks for adding A2A support to the Vault plugin. The shared token-injection helper, tag normalization, header-removal reconciliation, and focused coverage are all useful improvements. I found a few security concerns that I think should be addressed before merge.

**Blocking**

1. `X-Vault-Tokens` can still reach the upstream A2A endpoint when the Vault hook is absent, disabled, skipped, or fails permissively. The raw whitelisted header is added to the downstream request before plugin execution, while removal depends on the plugin returning a modified payload. Since the documented setup enables sensitive passthrough and the default Vault configuration is disabled, this could expose the complete credential bundle.

   Would it be possible to remove this header unconditionally at the outbound transport boundary and provide it to the Vault plugin through a trusted, scoped context instead?

2. A tagged A2A agent can select a credential using `system:<host>` without that identity being bound to the agent endpoint. For example, an agent pointing to an untrusted endpoint could use `system:github.com` and receive the caller’s GitHub token.

   I think credential selection should be validated against a privileged server-side mapping between the token audience and approved destination.

**Functionally impacting**

The widened tag types allow mixed string/dict lists, but `validate_tags_field()` handles the collection based on its first element. Mixed inputs can therefore silently lose tags, and dict-form tags can bypass normal format validation. Validating every element independently—or explicitly rejecting mixed forms—would make this behavior predictable.

One smaller concern: `X-Vault-Tokens` is not classified as sensitive by the generic header sanitizer, so every `agent_pre_invoke` plugin in the chain can inspect the full token bundle. Keeping this value isolated to the trusted Vault hook would better preserve the sanitizer’s security contract.

Also, the PR currently has no linked closing issue (`Closes #` is blank), so I could not verify the implementation against issue acceptance criteria.

Overall, the feature direction looks good and the focused tests are strong. I’d be comfortable reassessing once the credential-boundary and destination-binding concerns are resolved.

popagruia:
Hi @vishu-bh — thanks for the review. Update on PR #6079:

**X-Vault-Tokens reaching upstream when the hook is absent/disabled/errors permissively:** fixed with an unconditional strip in `a2a_service.py`, applied regardless of whether the plugin hook ran.

**Credential selection not bound to the agent endpoint:** fixed — a vault token can now carry `{"secretValue": ..., "mcpServer": "<url>"}`, and the plugin only injects it if that URL matches the target's actual registered endpoint. Mismatch or unknown destination → secret withheld. Covered by new tests and verified live end-to-end.

**Closing issue:** filed and linked, #6203.

**Not addressed as part of this PR:**
- Mixed string/dict tag lists can silently mishandle elements (`TagValidator.validate_list()`).
- `X-Vault-Tokens` isn't classified as sensitive by the generic header sanitizer, so other `agent_pre_invoke` plugins in the chain can see it before Vault strips it.

Both are pre-existing, in shared code outside this PR's scope — not fixing them here.

Ready for another look when you have time.

Lang-Akshay (changes requested):
Thanks for the PR @popagruia 
## Please address this  blocking Changes
| 1 | Security | `plugins/vault/vault_plugin.py` | 147–155 | `mcpServer` destination binding is taken directly from caller-controlled `X-Vault-Tokens` JSON. A caller holding a valid token mapping can rewrite that field to the current target URL, defeating the documented replay-prevention claim. Security handoff verdict is `SECURITY_VERDICT=BLOCK`. | Use a cryptographically authenticated Vault assertion (for example, a signed/MACed envelope containing token identity, destination/audience, and expiry) or resolve an opaque token reference server-side. Do not claim destination isolation for unsigned caller-supplied metadata. |

popagruia:
@Lang-Akshay. X-Vault-Tokens are send by the client as part of the header. It is the sender decision on how the tokens are mapped  and Vault plugin just dispatch them when a virtual server is mapped to multiple MCP Gateways. In my view this method is similar with  X-Upstream-Authorization that is also suported  by ContextForge and does not have the mecanism to validate that indeed is mapped to the actual server but is limited to just on MCP server per virtual server.

vishu-bh (dismissed):
LGTM 🚀

Lang-Akshay (approved):
LGTM ✅

PR #6080, opened 2026-08-05
chore: improve support bundle redaction and settings secret typing

msureshkumar88:
Improves how the support bundle handles configuration values, and tightens the typing of two settings fields.

### Support bundle

- The set of settings fields excluded from a generated bundle is now derived from the `Settings` model rather than a hand-maintained list. A hardcoded list goes stale as settings are added; deriving it means a field declared with a secret type is covered without a second edit. A narrow name-based rule acts as a backstop for string-typed fields, with `SecretStr` typing remaining the enforced convention.
- Credentials are stripped from every string-valued `*_url` setting rather than two named ones, and environment variables whose names end in `_URL` are masked, so inline credentials in connection strings are handled uniformly.

### Settings typing

- `csrf_secret_key` and `identity_claims_secret` are now `SecretStr`. Their fallback behaviour is unchanged: both still derive from `JWT_SECRET_KEY` when left unset.
- Worth knowing for anyone touching this area: `SecretStr("")` is truthy, so the emptiness checks guarding those fallbacks go through `get_secret_value()`, and the assignments wrap explicitly because `Settings` does not enable `validate_assignment`.

### Documentation

- The bundle README, the module docstring, and the Admin UI blurb now describe what each bundle member actually offers: deterministic exclusion for the settings file, name-based masking for environment variables, and best-effort pattern matching for logs. Each prompts the operator to review a bundle before sharing it.
- `CSRF_SECRET_KEY` is documented as reusing `JWT_SECRET_KEY` when unset, so the two can be set separately and rotated independently.

### Tracking

Internal: https://github.ibm.com/contextforge-org/internal_issues/issues/522

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -90,7 +90,9 @@ BASIC_AUTH_PASSWORD=changeme  # pragma: allowlist secret
 # Enable CSRF protection (default: true)
 CSRF_ENABLED=true
 
-# Secret key for CSRF token generation (falls back to JWT_SECRET_KEY if empty)
+# Secret used to sign CSRF tokens. If left unset, the application reuses
+# JWT_SECRET_KEY for this purpose. Set it explicitly so that the two keys can
+# be rotated independently.
 CSRF_SECRET_KEY=
 
 # HTTP header name for CSRF token
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^\\.secrets\\.baseline$",
     "lines": null
   },
-  "generated_at": "2026-08-05T12:11:08Z",
+  "generated_at": "2026-08-05T14:34:35Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -100,7 +100,7 @@
         "hashed_secret": "548a5cde15c93a9e50e12e4cba776b85f6befc83",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 132,
+        "line_number": 134,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -108,7 +108,7 @@
         "hashed_secret": "8e9a8667ded25c805ce3dc920a5f2343bd2d198f",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 133,
+        "line_number": 135,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -116,7 +116,7 @@
         "hashed_secret": "08cd923367890009657eab812753379bdb321eeb",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 777,
+        "line_number": 779,
         "type": "Basic Auth Credentials",
         "verified_result": null
       },
@@ -124,7 +124,7 @@
         "hashed_secret": "14f8aa3e560a47851908ab0f04ec856dbc512d93",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1022,
+        "line_number": 1024,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -132,7 +132,7 @@
         "hashed_secret": "fa9beb99e4029ad5a6615399e7bbae21356086b3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1321,
+        "line_number": 1323,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -140,7 +140,7 @@
         "hashed_secret": "d08f88df745fa7950b104e4a707a31cfce7b5841",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1358,
+        "line_number": 1360,
         "type": "Secret Keyword",
         "verified_result": null
       },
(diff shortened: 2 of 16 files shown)


Lang-Akshay (changes requested):
## Please address following blocking changes
 ---

## Blocking Changes

|---|------|------|------|-----------------|-----------------|
| 1 | Security | `mcpgateway/services/support_bundle_service.py` | 126 | `_sanitize_url()` iterates `SENSITIVE_PATTERNS`; the URL pattern `(\w[\w+.-]*)://([^:]+):([^@]+)@` requires `[^:]+` (≥1 char) for the username, so `redis://:password@host:6379` — cited by name in the PR's own comment at line 161 — is not redacted. The credential appears verbatim in `settings.json` inside any generated bundle. | Change `([^:]+)` to `([^:@]*)` to allow empty username. Also add a second alternation or query-string pattern for `?password=…` forms. |
| 2 | Security | `mcpgateway/services/support_bundle_service.py` | 413–415 | `_collect_settings()` applies `_sanitize_url()` only to keys where `isinstance(config[key], str)`. Collection-typed settings — `webhook_logging_urls` (`config.py:2255`, `list[str]`), `siem_destinations` (`config.py:2138`), `otel_exporter_otlp_headers` (`config.py:2915`, `dict[str,str]`) — are excluded from `_secret_field_names()` (not `SecretStr`) and are never sanitized, so any inline credentials in those values reach the bundle unchanged. | After the scalar loop, iterate collection-typed `*_url` keys (lists and dicts) and apply `_sanitize_url()` element-wise; or type those fields as `SecretStr`/`list[SecretStr]`. |

---

msureshkumar88:
Thanks @Lang-Akshay — both were real, and the first one was a good catch. Fixed in `ee5b1523f`.

## 1. Empty-username DSN — confirmed and fixed

Verified before changing anything:

```
redis://:password@host:6379              -> redis://:password@host:6379   (unchanged)
postgresql://user:password@localhost/db  -> postgresql://user:*****@localhost/db
```

Exactly as you described, and doubly worth fixing because that shape was cited by name in the comment right above the code that failed to handle it. Applied your suggested change — `([^:]+)` to `([^:@]*)`:

```
redis://:password@host:6379              -> redis://:*****@host:6379
redis://localhost:6379/0                 -> redis://localhost:6379/0     (no false positive)
```

The pattern is shared with the log sanitizer, so logs get the fix too.

One part of that item needs no change: query-string forms are already covered. `_sanitize_url()` runs the whole `SENSITIVE_PATTERNS` list, not just the URL entry, and the `password`/`token`/`secret`/`api_key` patterns match `key=value` anywhere in the string:

```
https://es:9200/?password=hunter2  -> https://es:9200/?password: *****
https://es:9200/?token=abc123      -> https://es:9200/?token: *****
```

I left that alone rather than adding a second pattern that would duplicate existing behaviour — happy to add one if you see a form those miss.

## 2. Collection-typed settings — confirmed, fixed more broadly than suggested

The underlying problem is real, and I took it further than the suggested fix because the suggestion would not have closed it. Two notes on the specifics:

- `otel_exporter_otlp_headers` is `Optional[str]` (`config.py:2915`), not `dict[str, str]`. It is still a genuine leak vector — a comma-separated `key=value` blob that conventionally carries `Authorization=Bearer …` — but as a scalar, so an element-wise pass over collections would have missed it.
- `webhook_logging_urls` holds Slack/Teams-style webhook URLs, whose secret lives in the URL **path**, not the userinfo. `_sanitize_url()` element-wise would have left those fully intact.

So rather than iterating collection-typed `*_url` keys, sanitization no longer keys off the field name at all. A name says nothing about the shape of the value behind it, which is the root cause of both findings. Every settings value is now walked and sanitized on its **content**:

- strings are matched against the same patterns already applied to log text
- lists and dicts are walked element-wise, at any depth
- inside a dict, a key whose name looks like a secret has its value replaced outright, since the value carries no pattern to match on

That nested-key check is deliberately broader than the regex screening our own typed settings. The narrow one excludes bare `token` and `key` so it does not redact `token_expiry` or `password_min_length`; nested keys come from operator-authored config with no such convention, where a key called `token` is a credential far more often than not.

Results:

```
siem_destinations   [{endpoint: https://siem.example.com, token: hunter2, name: prod}]
                 -> [{endpoint: https://siem.example.com, token: *****,  name: prod}]
otlp headers        Authorization=Bearer eyJa.eyJd.ghi,X-Env=prod
                 -> Authorization=bearer *****,X-Env=prod
webhook_logging_urls / any list[str] — sanitized element-wise
```

Checked for over-redaction, since content matching is the obvious risk: `auth_header_name` (`Authorization`), `csrf_token_name` (`X-CSRF-Token`), and benign mappings such as `sso_keycloak_role_mappings` all survive intact across the full 746-key settings dump. Adding `header` to the name regex was the other option and I rejected it — it would have swept in five benign string settings while still missing the collection cases.

Also fixed in passing: an empty-string URL setting used to serialize as `null`, since `_sanitize_url()` returns `None` for falsy input. It now stays `""` — "set but blank" and "unset" are different facts to whoever reads the bundle.

## Checked but not changed

`mcpgateway/version.py` has its own separate `_sanitize_url()`. It uses `urlsplit`, so it handles the empty-username case correctly already — no equivalent defect there.

## Coverage

Four tests added: empty-username redaction with a no-credential control, collection walking (list of strings, list of dicts, nested dicts, non-string scalars, empty string), and the bearer-token header blob. One existing test was updated — it asserted that a URL-shaped value under a name without a `_url` suffix was left untouched, which is precisely the behaviour this change inverts.

187 tests pass across the support-bundle, config, and identity-propagation suites; `ruff check`, `ruff format --check`, and `detect-secrets` all clean.

Lang-Akshay (approved):
LGTM ✅

PR #6084, opened 2026-08-05
fix(tests): re-check resources table before reading second row in view-modal test

gcgoncalves:
Re-check table + row count right before reading row 1, so a re-render race after closing the first modal skips cleanly instead of hanging 60s waiting on a row that's no longer there.


ja8zyjits (approved):
LGTM

prakhar-singh1928 (approved):

vishu-bh (commented):
LGTM 🚀

PR #6085, opened 2026-08-05
chore: improve support bundle redaction and settings secret typing (Release/v1.0.7)

msureshkumar88:
Backport of #6080 to `Release/v1.0.7`. Same two commits, cherry-picked cleanly (only `.secrets.baseline` needed a line-number-driven regeneration).

Improves how the support bundle handles configuration values, and tightens the typing of two settings fields.

### Support bundle

- The set of settings fields excluded from a generated bundle is now derived from the `Settings` model rather than a hand-maintained list. A hardcoded list goes stale as settings are added; deriving it means a field declared with a secret type is covered without a second edit. A narrow name-based rule acts as a backstop for string-typed fields, with `SecretStr` typing remaining the enforced convention.
- Every settings value is sanitized on its content rather than by field name: strings are matched against the same patterns already applied to log text, and lists/dicts (SIEM destinations, webhook URL lists, header blobs) are walked element-wise so credentials are caught regardless of what the field happens to be called.
- Credentials are stripped from URL settings even when the username is empty (`scheme://:password@host`), a common Redis DSN shape the original pattern missed.

### Settings typing

- `csrf_secret_key` and `identity_claims_secret` are now `SecretStr`. Their fallback behaviour is unchanged: both still derive from `JWT_SECRET_KEY` when left unset.
- Worth knowing for anyone touching this area: `SecretStr("")` is truthy, so the emptiness checks guarding those fallbacks go through `get_secret_value()`, and the assignments wrap explicitly because `Settings` does not enable `validate_assignment`.

### Documentation

- The bundle README, the module docstring, and the Admin UI blurb now describe what each bundle member actually offers: deterministic exclusion for the settings file, name-based masking for environment variables, and best-effort pattern matching for logs. Each prompts the operator to review a bundle before sharing it.
- `CSRF_SECRET_KEY` is documented as reusing `JWT_SECRET_KEY` when unset, so the two can be set separately and rotated independently.

### Tracking

Internal: https://github.ibm.com/contextforge-org/internal_issues/issues/522

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -90,7 +90,9 @@ BASIC_AUTH_PASSWORD=changeme  # pragma: allowlist secret
 # Enable CSRF protection (default: true)
 CSRF_ENABLED=true
 
-# Secret key for CSRF token generation (falls back to JWT_SECRET_KEY if empty)
+# Secret used to sign CSRF tokens. If left unset, the application reuses
+# JWT_SECRET_KEY for this purpose. Set it explicitly so that the two keys can
+# be rotated independently.
 CSRF_SECRET_KEY=
 
 # HTTP header name for CSRF token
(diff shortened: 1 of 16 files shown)


msureshkumar88:
Superseded by #6080, which now targets Release/v1.0.7 directly (same net change, original commits, existing review thread).

PR #6091, opened 2026-08-05
fix(teams): scope admin team listings by token teams

### LINKED ISSUE #5812: [BUG]: Admin team-listing endpoints ignore explicit token team scope (bypass leak)
### Description

Fast-follow to #5701 / #5668. #5668 fixed the admin team-scope leak for **unified search** (`admin_search_teams`) by pushing the caller's normalized `token_teams` into the `list_teams` query. The **same class of leak** — an admin-bypass path calling `list_teams()` unscoped, ignoring a narrowed `token_teams` — is still present in three other admin team-listing paths:

- `mcpgateway/admin.py` — `admin_teams_partial_html` (`GET /admin/teams/partial`), `list_teams(...)` call around line 5711.
- `mcpgateway/admin.py` — `admin_list_teams` (`GET /admin/teams`), `list_teams(...)` call around line 5930.
- `mcpgateway/routers/teams.py` — `GET /teams`, `list_teams(...)` in the `has_admin_team_access` branch (~line 214). **This one is sharper:** `check_platform_admin_permission` only denies bypass for `token_teams == []`, so a narrowed **non-empty** list (e.g. `["team-B"]`) still returns `has_admin_team_access=True` and an unscoped `list_teams()` — a genuine bypass, not just missing narrowing.

Per the token model this PR series aligns to ("Layer 1 constrains visibility independently of admin status"), these are the same bug: an admin using a token explicitly scoped to a team subset still sees every team.

### Impact

An admin who deliberately minted a team-scoped API token can still enumerate all teams (id / name / slug / description / visibility / status) through these endpoints. Admin UI session tokens and unscoped admin tokens resolve to `None` (full bypass) and are unaffected.

### Suggested fix

Apply the same pattern #5668 used for search: derive the normalized scope and pass `team_ids=scoped_team_ids` into `list_teams` (now supported), so the filter is enforced in the query before pagination. For `routers/teams.py:214`, also revisit `check_platform_admin_permission` so a narrowed non-empty token does not grant full-bypass listing.

### Notes

- Related: #5701 (search fix), #5668 (PR).
- The `team_ids` filter on `TeamManagementService.list_teams` already exists (added in #5668), so the fix is mostly wiring + deny-path tests per endpoint.

### Related nits (from #5668 review)

- **Shared extraction helper:** `admin.py:5590-5593` re-implements the `token_teams` -> id-list extraction that `_get_user_team_ids` (admin.py:1902-1913) partially does. Factor the extraction step into a small shared helper while wiring these endpoints, so the two do not drift.
- **Docs:** update `docs/docs/manage/rbac.md` / `docs/docs/architecture/multitenancy.md` if they describe the old personal-team fallback (removed for scoped/`[]` tokens).

gandhipratik203:
## Summary

Closes #5812.

This fixes the three team-listing endpoints named by the issue so explicit token team scope constrains what an admin token can see:

- `GET /admin/teams/partial`
- `GET /admin/teams`
- `GET /v1/teams/` and its legacy alias `GET /teams/`

The change adds a shared endpoint-context helper for extracting `token_teams`, threads that scope into the affected listing queries and manual listing branches, and scopes the API pagination total through `TeamManagementService.get_teams_count(team_ids=...)`.

## Scope

This fix operates at Layer 1 visibility filtering and does not change Layer 2 RBAC/admin privilege semantics. `check_platform_admin_permission()` is unchanged by design: it answers whether the caller has platform-admin privileges and also gates team create/update/delete paths. Narrowing its return value for scoped tokens would revoke write privileges as a side effect of this read-visibility fix. Instead, explicit token scope is pushed into the listing query/result path so the affected lists return only what the token permits.

`admin_search_teams` now reuses the shared extraction helper instead of duplicating token-team parsing inline. That keeps #5668 behavior aligned with the same normalized scope semantics used by the three team-listing endpoints and drops malformed entries rather than raising on them.

## Deferred

Not addressed here: `GET /admin/teams/ids` still calls `get_all_team_ids()` without token scope. It is the same class of leak, but it is not one of the three endpoints named by #5812 and needs separate `team_ids` plumbing on `TeamManagementService.get_all_team_ids()`.

Documentation updates in `docs/docs/manage/rbac.md` and `docs/docs/architecture/multitenancy.md` are also deferred. The code change here is intentionally scoped to the three named endpoints plus the service plumbing those endpoints require.

### DIFF
--- a/mcpgateway/admin.py
+++ b/mcpgateway/admin.py
@@ -1949,10 +1949,9 @@ async def _get_user_team_ids(user: dict, db: Session) -> list:
     if cached is not None:
         return cached
 
-    if "token_teams" in user:
-        team_ids = extract_token_team_ids(user)
-        if team_ids is not None:
-            return team_ids
+    team_ids = extract_token_team_ids(user)
+    if team_ids is not None:
+        return team_ids
 
     user_email = get_user_email(user)
     team_service = TeamManagementService(db)
@@ -1992,9 +1991,7 @@ def _is_explicit_token_team_scope(user: Any) -> bool:
     Returns:
         bool: True when ``token_teams`` is present and not ``None``.
     """
-    if not isinstance(user, dict):
-        return False
-    return "token_teams" in user and user.get("token_teams") is not None
+    return extract_token_team_ids(user) is not None
 
 
 def _owner_access_condition(owner_column, team_column, *, user_email: str, team_ids: list[str], user: Any):
--- a/mcpgateway/admin.py
+++ b/mcpgateway/admin.py
@@ -68,6 +68,7 @@ from mcpgateway.auth import get_current_user, get_user_team_roles
 # Re-export canonical get_user_email from auth_context for backward compatibility.
 from mcpgateway.auth_context import (
     configuration_export_includes_roots,
+    extract_token_team_ids,
     get_scoped_resource_access_context,
     get_token_teams_from_request,
     get_user_email,
@@ -1949,16 +1950,8 @@ async def _get_user_team_ids(user: dict, db: Session) -> list:
         return cached
 
     if "token_teams" in user:
-        token_teams = user.get("token_teams")
-        if token_teams is not None:
-            team_ids: list[str] = []
-            for team in token_teams:
-                if isinstance(team, dict):
-                    team_id = team.get("id")
-                    if isinstance(team_id, str) and team_id:
-                        team_ids.append(team_id)
-                elif isinstance(team, str) and team:
-                    team_ids.append(team)
+        team_ids = extract_token_team_ids(user)
+        if team_ids is not None:
             return team_ids
 
     user_email = get_user_email(user)
@@ -5346,13 +5339,14 @@ async def change_password_required_handler(request: Request, db: Session = Depen
 # ============================================================================ #
 
 
-async def _generate_unified_teams_view(team_service, current_user, root_path):  # pylint: disable=unused-argument
+async def _generate_unified_teams_view(team_service, current_user, root_path, scoped_team_ids: Optional[list[str]] = None):  # pylint: disable=unused-argument
     """Generate unified team view with relationship badges.
 
     Args:
         team_service: Service for team operations
         current_user: Current authenticated user
         root_path: Application root path
+        scoped_team_ids: Explicit token team scope to apply to the generated list
 
     Returns:
         HTML string containing the unified teams view
@@ -5362,6 +5356,11 @@ async def _generate_unified_teams_view(team_service, current_user, root_path):  
 
     # Get public teams user can join
     public_teams = await team_service.discover_public_teams(current_user.email)
+
+    if scoped_team_ids is not None:
+        allowed_team_ids = set(scoped_team_ids)
+        user_teams = [team for team in user_teams if str(team.id) in allowed_team_ids]
+        public_teams = [team for team in public_teams if str(team.id) in allowed_team_ids]
 
     # Batch fetch ALL data upfront - 3 queries instead of 3N queries (N+1 elimination)
     user_team_ids = [str(t.id) for t in user_teams]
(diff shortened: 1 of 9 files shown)


msureshkumar88 (approved):
## Code review

Verified against issue #5812: fix scopes all three named endpoints (`GET /admin/teams/partial`, `GET /admin/teams`, `GET /v1/teams`+`GET /teams` legacy alias) by pushing `team_ids=extract_token_team_ids(user)` into `TeamManagementService.list_teams()`/`get_teams_count()` before pagination, rather than reworking `check_platform_admin_permission()` as the issue suggested — a better call, since that function also gates write access and narrowing it would have silently revoked admin write privileges as a side effect. Change is backend-only (no `admin_ui/*.js`, no template changes) and touches no schema/Alembic — correct, since this is a query-filter fix, not a data-model change. `admin/teams/ids` is knowingly left unscoped, matching the issue's own scope note (not one of the three named endpoints). No dead code, no SQL injection surface (`.in_()` is parameterized), no breaking signature changes (`team_ids` is an added optional kwarg defaulting to `None` everywhere).

Two non-blocking suggestions, no blocking findings:
- No integration-level (real HTTP, `TestClient`) test for the fix, unlike the sibling `tests/integration/test_search_endpoint.py` from #5668. The e2e run below closes that gap empirically for this review, but an equivalent committed test would keep the regression coverage over time.
- `personal_owner_email` + `team_ids` combined behavior (a scoped admin no longer sees their own personal team unless it's in scope, per the comment at `admin.py:5639-5644`) has no dedicated test proving the AND-semantics against a real query.

## E2E verification

Ran two live gateway instances against isolated SQLite DBs to reproduce the issue on pre-fix code and confirm closure on this PR's head, rather than relying on mocked unit tests alone.

**Config** (both instances, isolated env, no changes to repo `.env`):
```
DATABASE_URL=sqlite:///<scratch>/{before,after}.db   (fresh per instance)
AUTH_REQUIRED=true
EMAIL_AUTH_ENABLED=true
JWT_SECRET_KEY / AUTH_ENCRYPTION_SECRET = ephemeral test-only values
```
- **BEFORE**: `git worktree add <scratch>/wt-before 7c042d214` (parent of this fix) → `uvicorn mcpgateway.main:app --port 14444`
- **AFTER**: this PR's head (`4cfb31af6`, checked out normally) → `uvicorn mcpgateway.main:app --port 14445`

**Commands**: bootstrap-admin JWT (`teams: null`) → `POST /v1/teams/` x2 to create team A/B → mint a token scoped to `teams: ["<team-A-id>"]` and a public-only token (`teams: []`) → `GET /v1/teams/`, `GET /teams/`, `GET /admin/teams`, `GET /admin/teams/partial` with each token.

**Output — AFTER (PR head): 13/13 PASS**
- Unscoped admin sees both teams on all 4 routes.
- Team-scoped token sees only team A, never team B, on all 4 routes (including the legacy `/teams/` alias).
- Public-only token (`teams: []`) sees no teams anywhere.

**Output — BEFORE (pre-fix, 7c042d214): reproduces the leak everywhere**
- Team-scoped token still saw team B on `/v1/teams/`, `/teams/`, `/admin/teams`, and `/admin/teams/partial` — confirms the bug as described in #5812 on live pre-fix code.
- Bonus finding: the public-only token (`teams: []`) *also* leaked on pre-fix `/v1/teams/`, because the non-admin fallback branch (`get_user_teams()`) ignored token scope entirely for teams the admin actually belongs to (e.g., a team they created while bootstrapping). This is the same bug class and is fixed by the same PR (the `if scoped_team_ids is not None` filtering added to that fallback branch) — good, this PR's fix is broader than just the `has_admin_team_access=True` path the issue text focused on.

No tracebacks/500s/unexpected errors in either server's logs during the run — no evidence of surrounding-code breakage.

**Approving** — the fix is minimal, correctly scoped, verified against real running instances of both the vulnerable and fixed code, and closes the issue plus an adjacent leak in the same code path.

vishu-bh (approved):
Implementation looks solid and correctly addresses the three endpoints identified in #5812. Scope filtering occurs before pagination, totals use the same filter, and the manual/unified rendering paths handle narrowed and empty scopes correctly. Test coverage is strong, and I found no new security or runtime regressions in the changed paths.
One security follow-up remains: GET /admin/teams/ids still calls get_all_team_ids() without applying the token’s team scope. A platform-admin token narrowed to a specific team can therefore enumerate IDs for all non-personal teams. This is explicitly documented as deferred and is outside the three endpoints named in #5812, so I don’t consider it blocking for this PR, but it would be worth tracking as a P2 follow-up.
From a code-review perspective, this looks good to merge after updating the branch against main and rerunning CI.

PR #6073, opened 2026-08-05
migration-fix

prakhar-singh1928:
## 🔗 Related Issue

relates to Release1.0.7

---

## 📝 Summary

Permits `AUTH_ENCRYPTION_SECRET` values that do not meet minimum cryptographic
strength requirements when `ENVIRONMENT=development`, reducing friction for
local PoC and first-run workflows without relaxing production security posture.

**Before:** Both `JWT_SECRET_KEY` and `AUTH_ENCRYPTION_SECRET` unconditionally
blocked startup in every environment when the configured value failed minimum
strength requirements (insufficient entropy, below 32 chars, or a known-default).

**After:**
- `JWT_SECRET_KEY` — behaviour unchanged; full cryptographic strength enforced
  unconditionally across all environments.
- `AUTH_ENCRYPTION_SECRET` — full enforcement in `staging` and `production`;
  in `ENVIRONMENT=development` values that do not meet minimum strength
  requirements emit a mandatory `WARNING` and startup proceeds. The
  `__REPLACE_ME__` placeholder is unconditionally rejected for both fields in
  every environment.
- `make setup` / `make init-secrets-patch-env` no longer silently overwrite
  `AUTH_ENCRYPTION_SECRET` in `.env` — operators must set it explicitly,
  making the choice deliberate and auditable. `make init-secrets` (`.env.secrets`)
  still generates a cryptographically secure value.

### Files changed

| File | Change |
|---|---|
| `mcpgateway/config.py` | Separated unconditional placeholder rejection from minimum-strength enforcement; `auth_encryption_secret` in development emits a mandatory `SECURITY WARNING` instead of blocking startup; docstring updated |
| `mcpgateway/scripts/init_secrets.py` | Introduced `_PATCH_ENV_SKIP_FIELDS`; `ensure_env_file_secrets` no longer auto-patches `AUTH_ENCRYPTION_SECRET` into `.env`; CLI help and docstrings updated |
| `Makefile` | Updated `init-secrets-patch-env` inline help; corrected `install-dev` advisory message |
| `.env.example` | All references to `AUTH_ENCRYPTION_SECRET` updated to use security-accurate language and reflect the two-tier enforcement model |
| `tests/unit/mcpgateway/test_config.py` | Renamed and corrected test; added `test_weak_auth_encryption_secret_allowed_in_development`, `test_weak_auth_encryption_secret_raises_outside_development`; extended `test_placeholder_secret_raises_in_development` to cover both fields |
[...]

PR #6108, opened 2026-08-06
chore: added AUTH_ENCRYPTION_SECRET unconditional strength enforcement

### LINKED ISSUE #6103: [CHORE]: AUTH_ENCRYPTION_SECRET cryptographic-strength guardrail
### 🔧 Chore Summary

PR #6073 deliberately relaxed the `AUTH_ENCRYPTION_SECRET` startup enforcement in
`ENVIRONMENT=development`, allowing short/low-entropy values (e.g. `my-test-salt`)
to emit a warning instead of blocking startup. That change was intentional for sprint
testing and migration-script preparation.

This chore restores the original unconditional hard-fail behaviour for
`AUTH_ENCRYPTION_SECRET` across **all** environments — matching the behaviour that existed
before PR #6073 — once the migration scripts are validated and merged.

---

### 🧱 Area Affected

- [ ] GitHub Actions / CI Pipelines
- [ ] Pre-commit hooks / linters
- [ ] Type-checking (mypy, pyright, pytype, etc.)
- [ ] Build system or `Makefile`
- [ ] Docs or spellcheck
- [x] Other: Security guardrail / startup validation (`mcpgateway/config.py`, `mcpgateway/scripts/init_secrets.py`)

---

### ⚙️ Context / Rationale

The two-tier enforcement model introduced in PR #6073 was a **temporary, sprint-scoped
relaxation** to unblock local PoC testing and allow migration scripts to be authored and
validated without requiring a cryptographically strong `AUTH_ENCRYPTION_SECRET` in every
dev environment.

The original design intent (and the security invariant documented in `AGENTS.md`) is that
both `JWT_SECRET_KEY` **and** `AUTH_ENCRYPTION_SECRET` are unconditionally enforced in
every environment — placeholder, known-weak, too-short, and low-entropy values all hard-fail
at startup regardless of `ENVIRONMENT`.

Restoring this rail ensures:
- No developer accidentally ships a configuration with a weak `AUTH_ENCRYPTION_SECRET` to
  staging or production after copying a dev `.env`.
- `make init-secrets-patch-env` / `make setup` correctly auto-patches
  `AUTH_ENCRYPTION_SECRET` in `.env` (currently skipped per `_PATCH_ENV_SKIP_FIELDS`).
- `.env.example` comments accurately reflect unconditional enforcement for both secrets.

**Scope of revert** (files changed in PR #6073 that need restoring):

| File | What to restore |
|---|---|
| `mcpgateway/config.py` | Remove the `is_dev` / `auth_encryption_secret` dev-bypass branch; restore single-loop unconditional enforcement for both fields |
| `mcpgateway/scripts/init_secrets.py` | Remove `_PATCH_ENV_SKIP_FIELDS`; restore `AUTH_ENCRYPTION_SECRET` to `ensure_env_file_secrets()` auto-patching |
| `Makefile` | Restore advisory message and `init-secrets-patch-env` help text to cover both secrets |
| `.env.example` | Restore original two-secret enforcement comment; remove dev-relaxation language |
| `tests/unit/mcpgateway/test_config.py` | Restore `test_weak_auth_encryption_secret_raises_in_development` (hard-fail); remove `test_weak_auth_encryption_secret_allowed_in_development` and the dev-placeholder pass-through assertion |
| `tests/scripts/test_init_secrets.py` | Restore assertions that `AUTH_ENCRYPTION_SECRET` IS present in `ensure_env_file_secrets()` output |

---

### 📦 Related Make Targets

- `make setup` / `make init-secrets-patch-env` — must auto-patch `AUTH_ENCRYPTION_SECRET` again
- `make init-secrets` — already writes a strong value to `.env.secrets`; no change needed
- `make check-env` — validate the restored `.env` requirements
- `make lint` / `make test` — full validation after revert

---

### 📋 Acceptance Criteria

- [ ] `AUTH_ENCRYPTION_SECRET` with a weak/short/placeholder/low-entropy value causes a `SecurityConfigurationError` in **all** environments (`development`, `staging`, `production`)
- [ ] `make init-secrets-patch-env` / `make setup` auto-patches `AUTH_ENCRYPTION_SECRET` into `.env` when the current value is non-compliant
- [ ] `test_weak_auth_encryption_secret_raises_in_development` passes (asserts hard-fail, not warning)
- [ ] `test_weak_auth_encryption_secret_allowed_in_development` is removed (or converted back to the raises variant)
- [ ] `ensure_env_file_secrets()` tests assert `AUTH_ENCRYPTION_SECRET` IS present in generated output
- [ ] Lint runs cleanly (`make lint`)
- [ ] CI passes with no regressions (`make test`)
- [ ] `.env.example` and `Makefile` advisory text reflect unconditional enforcement for both secrets
- [ ] Security scans pass (`make detect-secrets-scan`)

---

### 🧩 Additional Context

- **Blocked by**: migration scripts for `AUTH_ENCRYPTION_SECRET` must be authored and validated
  before this guardrail is restored — those scripts are the prerequisite for this chore.
- **Tracking**: PR #6073 (`Release-1.0.7/migration-fix`) is the exact inverse of this change.
  The diff in that PR is the authoritative reference for what must be reverted.
- **`WEAK_VALUES`** in `mcpgateway/config.py` already includes `my-test-salt` and
  similar short values — no additions needed, only the enforcement path must be restored.

prakhar-singh1928:
## 🔗 Related Issue

Closes #6103 

---

## 📝 Summary

migration script for this issue is covered with a follow up PR Here - https://github.com/IBM/mcp-context-forge/issues/6113
Restores the unconditional `AUTH_ENCRYPTION_SECRET` cryptographic-strength guardrail
that was deliberately relaxed in PR #6073 for sprint testing and migration-script
preparation.

**Before (PR #6073 state):**
- `AUTH_ENCRYPTION_SECRET` with weak/short/low-entropy values emitted a `WARNING` in
  `ENVIRONMENT=development` and startup proceeded — enabling `my-test-salt` style values
  locally.
- `make init-secrets-patch-env` / `make setup` did **not** auto-patch
  `AUTH_ENCRYPTION_SECRET` into `.env`.

**After (this PR):**
- `AUTH_ENCRYPTION_SECRET` is enforced unconditionally in **every** environment
  (development, staging, production) — matching the pre-#6073 behaviour.
- Placeholder, known-weak, too-short, and low-entropy values all hard-fail at startup
  regardless of `ENVIRONMENT`.
- `make init-secrets-patch-env` / `make setup` auto-patches `AUTH_ENCRYPTION_SECRET`
  into `.env` again when the current value is non-compliant.

### Files changed

| File | Change |
|---|---|
| `mcpgateway/config.py` | Removed `is_dev` bypass branch for `auth_encryption_secret`; restored single-loop unconditional enforcement for both secrets; docstring reverted |
| `mcpgateway/scripts/init_secrets.py` | Removed `_PATCH_ENV_SKIP_FIELDS`; restored `AUTH_ENCRYPTION_SECRET` to `ensure_env_file_secrets()` auto-patching loop; `--patch-env` help text reverted |
| `Makefile` | Restored advisory message and `init-secrets-patch-env` help text to cover both secrets |
| `.env.example` | Restored original two-secret enforcement comment block; removed dev-relaxation language; reverted `ENVIRONMENT` section comment |
| `tests/unit/mcpgateway/test_config.py` | Removed `test_weak_auth_encryption_secret_allowed_in_development`; removed dev-placeholder pass-through assertion from `test_placeholder_secret_raises_in_development`; restored `test_weak_auth_encryption_secret_raises_in_development` (hard-fail); restored `test_init_secrets_patch_mode_writes_strong_values` to assert both secrets are generated |
| `tests/scripts/test_init_secrets.py` | Restored assertions that `AUTH_ENCRYPTION_SECRET` IS present in `ensure_env_file_secrets()` output for weak and placeholder inputs |

---

## 📏 Reviewability

---

## 🏷️ Type of Change

- [ ] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor

---

## 🧪 Verification

| Check | Command | Status |
|---|---|---|
| Security config tests | `pytest tests/unit/mcpgateway/test_config.py -q` | 185 passed |
| Init secrets tests | `pytest tests/scripts/test_init_secrets.py -q` | 37 passed |
| Weak enc secret hard-fails in dev | `test_weak_auth_encryption_secret_raises_in_development` | ✅ |
| Placeholder hard-fails in dev | `test_placeholder_secret_raises_in_development` | ✅ |
| Patch mode generates both secrets | `test_init_secrets_patch_mode_writes_strong_values` | ✅ |
| Strong secrets still accepted | `test_strong_secrets_accepted_in_all_envs` | ✅ |

---

## ✅ Checklist

### DIFF
--- a/mcpgateway/scripts/init_secrets.py
+++ b/mcpgateway/scripts/init_secrets.py
@@ -115,11 +115,50 @@ def _is_strong_value(val: str, weak_values: frozenset[str]) -> bool:
     return True
 
 
+def _read_env_file(path: str) -> dict[str, str]:
+    """Parse KEY=VALUE pairs from an env file, skipping comments and blank lines.
+
+    Thin compatibility shim backed by ``dotenv_values`` so callers and tests
+    that import ``_read_env_file`` continue to work after the internal refactor
+    to ``dotenv_values``.  The returned dict uses the original key casing from
+    the file (i.e. uppercase-as-written), matching the previous hand-rolled
+    parser semantics.
+
+    Lines that do not contain an ``=`` sign (e.g. bare ``export FOO`` directives)
+    are skipped, preserving the behaviour of the original hand-rolled parser.
+
+    Args:
+        path: Path to the env file.
+
+    Returns:
+        dict: Parsed ``{KEY: value}`` mapping (original casing).  Empty dict
+        when the file does not exist.
+    """
+    try:
+        raw_lines = open(path, encoding="utf-8").readlines()
+    except FileNotFoundError:
+        return {}
+    keys_with_equals: set[str] = set()
+    for line in raw_lines:
+        stripped = line.strip()
+        if stripped.startswith("#") or not stripped:
+            continue
+        if "=" in stripped:
+            key = stripped.split("=", 1)[0].strip()
+            # Strip leading "export " so "export FOO=bar" registers as "FOO"
+            if key.lower().startswith("export "):
+                key = key[len("export ") :].strip()
+            keys_with_equals.add(key)
+    return {k: v for k, v in _dotenv_values(path).items() if k in keys_with_equals}
+
+
 def _merge_env_file(path: str, updates: dict[str, str]) -> None:
     """Merge key=value pairs into an env file.
 
-    Existing keys in *updates* are replaced in-place; new keys are appended.
-    All other content (comments, blanks, other keys) is preserved.
+    Existing keys in *updates* are replaced in-place (case-insensitive match
+    so ``auth_encryption_secret=weak`` is correctly replaced when the update
+    key is ``AUTH_ENCRYPTION_SECRET``); new keys are appended.  All other
+    content (comments, blanks, other keys) is preserved.
     File is written with owner-only permissions (0o600).
     """
     existing_lines: list[str] = []
@@ -129,15 +168,19 @@ def _merge_env_file(path: str, updates: dict[str, str]) -> None:
     except FileNotFoundError:
         pass
 
+    # Build a case-insensitive lookup: lowercase(update_key) → canonical update key
+    updates_ci: dict[str, str] = {k.lower(): k for k in updates}
+
     updated_keys: set[str] = set()
     new_lines: list[str] = []
     for line in existing_lines:
         stripped = line.strip()
         if stripped and not stripped.startswith("#") and "=" in stripped:
-            key = stripped.partition("=")[0].strip()
-            if key in updates:
-                new_lines.append(f"{key}={updates[key]}\n")
-                updated_keys.add(key)
+            file_key = stripped.partition("=")[0].strip()
+            canonical_key = updates_ci.get(file_key.lower())
+            if canonical_key is not None:
+                new_lines.append(f"{canonical_key}={updates[canonical_key]}\n")
+                updated_keys.add(canonical_key)
                 continue
         new_lines.append(line)
 
@@ -317,7 +360,11 @@ def main() -> None:
     # --patch-env: in-place update of an existing env file
     patch_target = args.patch_env
     if patch_target is not None:
-        generated = ensure_env_file_secrets(env_file=patch_target)
+        try:
+            generated = ensure_env_file_secrets(env_file=patch_target)
+        except ValueError as exc:
+            print(f"❌  {exc}", file=sys.stderr)
+            sys.exit(1)
         if generated:
             patched_keys = ", ".join(generated.keys())
             print(f"✅  Patched {patch_target}: generated strong values for {patched_keys}")
--- a/mcpgateway/scripts/init_secrets.py
+++ b/mcpgateway/scripts/init_secrets.py
@@ -237,13 +237,17 @@ def ensure_env_file_secrets(
 
     # Honour operator-raised MIN_SECRET_LENGTH floor (e.g. MIN_SECRET_LENGTH=64).
     # Precedence: os.environ > .env file > built-in constant — mirrors pydantic-settings.
-    # Using max() keeps the constant as the lower bound — never go below 32.
-    _min_raw = os.environ.get("MIN_SECRET_LENGTH") or _env_file_ci.get("min_secret_length") or str(_MIN_SECRET_LENGTH)
+    # Use is not None so that MIN_SECRET_LENGTH="" (unset in shell) correctly falls through
+    # to the .env value rather than being treated as an empty-string override.
+    _min_env = os.environ.get("MIN_SECRET_LENGTH")
+    _min_raw = _min_env if _min_env is not None else (_env_file_ci.get("min_secret_length") or str(_MIN_SECRET_LENGTH))
     try:
         configured_min: int = int(_min_raw)
     except ValueError:
         raise ValueError(f"MIN_SECRET_LENGTH={_min_raw!r} is not a valid integer. Set it to a whole number (e.g. MIN_SECRET_LENGTH=64).") from None
-    effective_min: int = max(configured_min, _MIN_SECRET_LENGTH)
+    if configured_min < _MIN_SECRET_LENGTH:
+        raise ValueError(f"MIN_SECRET_LENGTH={configured_min} is below the enforced minimum of {_MIN_SECRET_LENGTH}. Settings will reject it at startup — set it to {_MIN_SECRET_LENGTH} or higher.")
+    effective_min: int = configured_min
 
     for field, nbytes in _SECRET_FIELDS.items():
         # os.environ takes priority over .env (mirrors pydantic-settings behaviour)
@@ -259,9 +263,10 @@ def ensure_env_file_secrets(
             # always satisfies Settings.validate_security_combinations() on startup.
             token_bytes = max(nbytes, effective_min) if field in _STRONG_SECRET_FIELDS else nbytes
             new_val = generate_token(token_bytes)
-            if env_val is not None and field.lower() not in _env_file_ci:
-                # Weak value came from os.environ; patching environ is enough.
-                # Writing to .env would shadow subsequent env-var injections (Docker/K8s).
+            if env_val is not None and field.lower() not in _env_file_ci and field not in _STRONG_SECRET_FIELDS:
+                # Non-strong field (e.g. BASIC_AUTH_PASSWORD) with a weak value from
+                # os.environ only — patching environ is enough; writing to .env would
+                # shadow env-var injections in container setups.
                 env_only[field] = new_val
             elif env_val is not None and field in _ROTATION_GUARDED_FIELDS and _is_strong_value(_env_file_ci.get(field.lower(), ""), weak_values):
                 # The shell environment holds a weak value, but .env already contains a
--- a/mcpgateway/scripts/init_secrets.py
+++ b/mcpgateway/scripts/init_secrets.py
@@ -82,13 +82,6 @@ _SECRET_FIELDS: dict[str, int] = {
     "BASIC_AUTH_PASSWORD": 18,  # nosec B105 — patched when "changeme" or placeholder; 18 bytes → 24 chars
 }
 
-# Fields that are written to .env.secrets (--output / --stdout) but are NOT
-# patched into .env by --patch-env / ensure_env_file_secrets().
-# AUTH_ENCRYPTION_SECRET is intentionally excluded from auto-patching: in
-# ENVIRONMENT=development a weak value (e.g. my-test-salt) is allowed, so
-# operators should set it deliberately rather than having it silently replaced.
-_PATCH_ENV_SKIP_FIELDS: frozenset[str] = frozenset({"AUTH_ENCRYPTION_SECRET"})
-
 
 def _read_env_file(path: str) -> dict[str, str]:
     """Parse KEY=VALUE pairs from an env file, skipping comments and blank lines.
@@ -171,16 +164,11 @@ def ensure_env_file_secrets(
     env_file: str = ".env",
     weak_values: frozenset[str] | None = None,
 ) -> dict[str, str]:
-    """Check JWT_SECRET_KEY and BASIC_AUTH_PASSWORD for weak or placeholder values.
+    """Check JWT_SECRET_KEY, AUTH_ENCRYPTION_SECRET, and BASIC_AUTH_PASSWORD for weak or placeholder values.
 
     If weak values are detected, generates cryptographically strong replacements,
     merges them into *env_file* (creating the file if it does not exist), and
     patches ``os.environ`` so the running process picks them up without restart.
-
-    AUTH_ENCRYPTION_SECRET is intentionally NOT patched by this function.
-    In ENVIRONMENT=development a weak value (e.g. ``my-test-salt``) is allowed,
-    so operators should set it deliberately.  Strong values for it are still
-    written to .env.secrets by the --output / --stdout paths.
 
     Set ``MCPGATEWAY_AUTO_INIT_SECRETS=false`` to disable this behaviour (e.g.
     when secrets are injected via Vault or Kubernetes secrets and disk writes
@@ -202,8 +190,6 @@ def ensure_env_file_secrets(
     file_generated: dict[str, str] = {}
 
     for field, nbytes in _SECRET_FIELDS.items():
-        if field in _PATCH_ENV_SKIP_FIELDS:
-            continue  # never auto-patched into .env (see _PATCH_ENV_SKIP_FIELDS)
         # os.environ takes priority over .env (mirrors pydantic-settings behaviour)
         env_val = os.environ.get(field)
         current = env_val if env_val is not None else env_file_values.get(field, "changeme")
@@ -299,11 +285,9 @@ def main() -> None:
         metavar="ENV_FILE",
         default=None,
         help=(
-            "Patch an existing env file in-place: replace only the JWT_SECRET_KEY "
-            "and BASIC_AUTH_PASSWORD lines that still hold placeholder or weak values; "
-            "all other lines are preserved unchanged. No-ops if those keys are already strong. "
-            "AUTH_ENCRYPTION_SECRET is NOT patched — set it manually or use --stdout / "
-            "--output to get a strong value for .env.secrets."
+            "Patch an existing env file in-place: replace only the JWT_SECRET_KEY and "
+            "AUTH_ENCRYPTION_SECRET lines that still hold placeholder or weak values; "
+            "all other lines are preserved unchanged. No-ops if those keys are already strong."
         ),
     )
     args = parser.parse_args()
--- a/mcpgateway/scripts/init_secrets.py
+++ b/mcpgateway/scripts/init_secrets.py
@@ -236,8 +236,16 @@ def ensure_env_file_secrets(
     file_generated: dict[str, str] = {}
 
     # Honour operator-raised MIN_SECRET_LENGTH floor (e.g. MIN_SECRET_LENGTH=64).
-    # Using max() keeps the constant as the lower bound — never go below 32 bytes.
-    configured_min: int = int(os.environ.get("MIN_SECRET_LENGTH", _MIN_SECRET_LENGTH))
+    # Precedence: os.environ > .env file > built-in constant — mirrors pydantic-settings.
+    # Using max() keeps the constant as the lower bound — never go below 32.
+    _min_raw = os.environ.get("MIN_SECRET_LENGTH") or _env_file_ci.get("min_secret_length") or str(_MIN_SECRET_LENGTH)
+    try:
+        configured_min: int = int(_min_raw)
+    except ValueError:
+        raise ValueError(
+            f"MIN_SECRET_LENGTH={_min_raw!r} is not a valid integer. "
+            "Set it to a whole number (e.g. MIN_SECRET_LENGTH=64)."
+        ) from None
     effective_min: int = max(configured_min, _MIN_SECRET_LENGTH)
 
     for field, nbytes in _SECRET_FIELDS.items():
--- a/mcpgateway/scripts/init_secrets.py
+++ b/mcpgateway/scripts/init_secrets.py
@@ -235,6 +235,11 @@ def ensure_env_file_secrets(
     # Keys whose weak value came from .env (or was absent) — write to disk + environ.
     file_generated: dict[str, str] = {}
 
+    # Honour operator-raised MIN_SECRET_LENGTH floor (e.g. MIN_SECRET_LENGTH=64).
+    # Using max() keeps the constant as the lower bound — never go below 32 bytes.
+    configured_min: int = int(os.environ.get("MIN_SECRET_LENGTH", _MIN_SECRET_LENGTH))
+    effective_min: int = max(configured_min, _MIN_SECRET_LENGTH)
+
     for field, nbytes in _SECRET_FIELDS.items():
         # os.environ takes priority over .env (mirrors pydantic-settings behaviour)
         env_val = os.environ.get(field)
@@ -243,9 +248,12 @@ def ensure_env_file_secrets(
         is_non_compliant = not current.strip() or current.lower() in weak_values or current.lower().startswith("__replace_me__")
         # Length and entropy checks apply only to secrets that Settings hard-fails on startup.
         if not is_non_compliant and field in _STRONG_SECRET_FIELDS:
-            is_non_compliant = len(current) < _MIN_SECRET_LENGTH or calculate_entropy(current) < _MIN_ENTROPY
+            is_non_compliant = len(current) < effective_min or calculate_entropy(current) < _MIN_ENTROPY
         if is_non_compliant:
-            new_val = generate_token(nbytes)
+            # Size the generated token against the operator-configured floor so it
+            # always satisfies Settings.validate_security_combinations() on startup.
+            token_bytes = max(nbytes, effective_min) if field in _STRONG_SECRET_FIELDS else nbytes
+            new_val = generate_token(token_bytes)
             if env_val is not None and field.lower() not in _env_file_ci:
                 # Weak value came from os.environ; patching environ is enough.
                 # Writing to .env would shadow subsequent env-var injections (Docker/K8s).
--- a/mcpgateway/scripts/init_secrets.py
+++ b/mcpgateway/scripts/init_secrets.py
@@ -135,7 +135,8 @@ def _read_env_file(path: str) -> dict[str, str]:
         when the file does not exist.
     """
     try:
-        raw_lines = open(path, encoding="utf-8").readlines()
+        with open(path, encoding="utf-8") as fh:
+            raw_lines = fh.readlines()
     except FileNotFoundError:
         return {}
     keys_with_equals: set[str] = set()
--- a/mcpgateway/scripts/init_secrets.py
+++ b/mcpgateway/scripts/init_secrets.py
@@ -89,6 +89,28 @@ _SECRET_FIELDS: dict[str, int] = {
 # BASIC_AUTH_PASSWORD is excluded — Settings only warns on it, never hard-fails.
 _STRONG_SECRET_FIELDS: frozenset[str] = frozenset({"JWT_SECRET_KEY", "AUTH_ENCRYPTION_SECRET"})
 
+# Fields where auto-rotation without a data-migration is irreversible.
+# If the shell environment carries a weak value but .env already holds a strong
+# value for one of these fields, ensure_env_file_secrets() must NOT silently
+# overwrite the .env value — doing so rotates the AES key and makes all stored
+# encrypted credentials permanently unreadable.
+_ROTATION_GUARDED_FIELDS: frozenset[str] = frozenset({"AUTH_ENCRYPTION_SECRET"})
+
+
+def _is_strong_value(val: str, weak_values: frozenset[str]) -> bool:
+    """Return True when *val* passes all startup compliance checks.
+
+    Mirrors the predicate in :func:`ensure_env_file_secrets` so the rotation-guard
+    check uses identical criteria.
+    """
+    if not val.strip():
+        return False
+    if val.lower() in weak_values or val.lower().startswith("__replace_me__"):
+        return False
+    if len(val) < _MIN_SECRET_LENGTH or calculate_entropy(val) < _MIN_ENTROPY:
+        return False
+    return True
+
 
 def _read_env_file(path: str) -> dict[str, str]:
     """Parse KEY=VALUE pairs from an env file, skipping comments and blank lines.
@@ -211,6 +233,21 @@ def ensure_env_file_secrets(
                 # Weak value came from os.environ; patching environ is enough.
                 # Writing to .env would shadow subsequent env-var injections (Docker/K8s).
                 env_only[field] = new_val
+            elif env_val is not None and field in _ROTATION_GUARDED_FIELDS and _is_strong_value(env_file_values.get(field, ""), weak_values):
+                # The shell environment holds a weak value, but .env already contains a
+                # strong value for a rotation-guarded key (AUTH_ENCRYPTION_SECRET).
+                # Overwriting the .env value would silently rotate the AES encryption key
+                # and make all stored encrypted credentials permanently unreadable.
+                # Raise an actionable error instead of proceeding.
+                raise ValueError(
+                    f"{field}: shell environment holds a weak/non-compliant value that would "
+                    f"overwrite the strong value already present in {env_file!r}. "
+                    "Silently rotating this key would make all stored encrypted credentials "
+                    "permanently unreadable. To fix, choose one of:\n"
+                    f"  unset {field}           # let the strong .env value take effect\n"
+                    f"  export {field}=<strong> # set a strong value in your shell that matches {env_file!r}\n"
+                    "  run a migration-aware key rotation before changing this secret"
+                )
             else:
                 file_generated[field] = new_val
 
--- a/mcpgateway/scripts/init_secrets.py
+++ b/mcpgateway/scripts/init_secrets.py
@@ -32,7 +32,10 @@ import secrets
 import sys
 
 # First-Party
+from mcpgateway._security_constants import MIN_ENTROPY as _MIN_ENTROPY
+from mcpgateway._security_constants import MIN_SECRET_LENGTH as _MIN_SECRET_LENGTH
 from mcpgateway._security_constants import WEAK_VALUES as _CANONICAL_WEAK_VALUES
+from mcpgateway._security_constants import calculate_entropy
 
 
 def _secure_open_flags(force: bool) -> int:
@@ -189,11 +192,19 @@ def ensure_env_file_secrets(
     # Keys whose weak value came from .env (or was absent) — write to disk + environ.
     file_generated: dict[str, str] = {}
 
+    # Fields subject to the full compliance predicate (startup hard-fail): length + entropy enforced.
+    _STRONG_SECRET_FIELDS = {"JWT_SECRET_KEY", "AUTH_ENCRYPTION_SECRET"}
+
     for field, nbytes in _SECRET_FIELDS.items():
         # os.environ takes priority over .env (mirrors pydantic-settings behaviour)
         env_val = os.environ.get(field)
         current = env_val if env_val is not None else env_file_values.get(field, "changeme")
-        if current.lower() in weak_values or current.lower().startswith("__replace_me__"):
+        # Base checks apply to all fields: empty, known-weak name, placeholder.
+        is_non_compliant = not current.strip() or current.lower() in weak_values or current.lower().startswith("__replace_me__")
+        # Length and entropy checks apply only to secrets that Settings hard-fails on startup.
+        if not is_non_compliant and field in _STRONG_SECRET_FIELDS:
+            is_non_compliant = len(current) < _MIN_SECRET_LENGTH or calculate_entropy(current) < _MIN_ENTROPY
+        if is_non_compliant:
             new_val = generate_token(nbytes)
             if env_val is not None and field not in env_file_values:
                 # Weak value came from os.environ; patching environ is enough.
--- a/mcpgateway/scripts/init_secrets.py
+++ b/mcpgateway/scripts/init_secrets.py
@@ -85,6 +85,10 @@ _SECRET_FIELDS: dict[str, int] = {
     "BASIC_AUTH_PASSWORD": 18,  # nosec B105 — patched when "changeme" or placeholder; 18 bytes → 24 chars
 }
 
+# Fields subject to the full compliance predicate (startup hard-fail): length + entropy enforced.
+# BASIC_AUTH_PASSWORD is excluded — Settings only warns on it, never hard-fails.
+_STRONG_SECRET_FIELDS: frozenset[str] = frozenset({"JWT_SECRET_KEY", "AUTH_ENCRYPTION_SECRET"})
+
 
 def _read_env_file(path: str) -> dict[str, str]:
     """Parse KEY=VALUE pairs from an env file, skipping comments and blank lines.
@@ -191,9 +195,6 @@ def ensure_env_file_secrets(
     env_only: dict[str, str] = {}
     # Keys whose weak value came from .env (or was absent) — write to disk + environ.
     file_generated: dict[str, str] = {}
-
-    # Fields subject to the full compliance predicate (startup hard-fail): length + entropy enforced.
-    _STRONG_SECRET_FIELDS = {"JWT_SECRET_KEY", "AUTH_ENCRYPTION_SECRET"}
 
     for field, nbytes in _SECRET_FIELDS.items():
         # os.environ takes priority over .env (mirrors pydantic-settings behaviour)
--- a/mcpgateway/scripts/init_secrets.py
+++ b/mcpgateway/scripts/init_secrets.py
@@ -268,18 +268,20 @@ def ensure_env_file_secrets(
                 # os.environ only — patching environ is enough; writing to .env would
                 # shadow env-var injections in container setups.
                 env_only[field] = new_val
-            elif env_val is not None and field in _ROTATION_GUARDED_FIELDS and _is_strong_value(_env_file_ci.get(field.lower(), ""), weak_values):
-                # The shell environment holds a weak value, but .env already contains a
-                # strong value for a rotation-guarded key (AUTH_ENCRYPTION_SECRET).
-                # Overwriting the .env value would silently rotate the AES encryption key
-                # and make all stored encrypted credentials permanently unreadable.
-                # Raise an actionable error instead of proceeding.
+            elif env_val is not None and field in _ROTATION_GUARDED_FIELDS and (_env_file_ci.get(field.lower()) or "").strip():
+                # The shell environment holds a weak/non-compliant value, but .env already
+                # contains *any* pre-existing value for a rotation-guarded key
+                # (AUTH_ENCRYPTION_SECRET).  Overwriting it — even to replace a short value
+                # with a strong one — silently rotates the AES encryption key and makes all
+                # stored encrypted credentials permanently unreadable.  Raise an actionable
+                # error for every case, not just when the existing .env value is already
+                # strong.
                 raise ValueError(
                     f"{field}: shell environment holds a weak/non-compliant value that would "
-                    f"overwrite the strong value already present in {env_file!r}. "
+                    f"overwrite the value already present in {env_file!r}. "
                     "Silently rotating this key would make all stored encrypted credentials "
                     "permanently unreadable. To fix, choose one of:\n"
-                    f"  unset {field}           # let the strong .env value take effect\n"
+                    f"  unset {field}           # let the .env value take effect\n"
                     f"  export {field}=<strong> # set a strong value in your shell that matches {env_file!r}\n"
                     "  run a migration-aware key rotation before changing this secret"
                 )
--- a/mcpgateway/scripts/init_secrets.py
+++ b/mcpgateway/scripts/init_secrets.py
@@ -242,10 +242,7 @@ def ensure_env_file_secrets(
     try:
         configured_min: int = int(_min_raw)
     except ValueError:
-        raise ValueError(
-            f"MIN_SECRET_LENGTH={_min_raw!r} is not a valid integer. "
-            "Set it to a whole number (e.g. MIN_SECRET_LENGTH=64)."
-        ) from None
+        raise ValueError(f"MIN_SECRET_LENGTH={_min_raw!r} is not a valid integer. Set it to a whole number (e.g. MIN_SECRET_LENGTH=64).") from None
     effective_min: int = max(configured_min, _MIN_SECRET_LENGTH)
 
     for field, nbytes in _SECRET_FIELDS.items():
--- a/mcpgateway/scripts/init_secrets.py
+++ b/mcpgateway/scripts/init_secrets.py
@@ -31,6 +31,9 @@ import os
 import secrets
 import sys
 
+# Third-Party
+from dotenv import dotenv_values as _dotenv_values
+
 # First-Party
 from mcpgateway._security_constants import MIN_ENTROPY as _MIN_ENTROPY
 from mcpgateway._security_constants import MIN_SECRET_LENGTH as _MIN_SECRET_LENGTH
@@ -112,39 +115,6 @@ def _is_strong_value(val: str, weak_values: frozenset[str]) -> bool:
     return True
 
 
-def _read_env_file(path: str) -> dict[str, str]:
-    """Parse KEY=VALUE pairs from an env file, skipping comments and blank lines.
-
-    Handles: quoted values, ``export KEY=val`` prefix, inline ``# comments``,
-    and spaces around ``=``.  Matches python-dotenv parsing semantics so that
-    weak-value detection fires on the same string pydantic-settings would see.
-    """
-    result: dict[str, str] = {}
-    try:
-        with open(path, encoding="utf-8") as fh:
-            for line in fh:
-                line = line.strip()
-                if not line or line.startswith("#"):
-                    continue
-                if line.startswith("export "):
-                    line = line[len("export ") :]
-                if "=" not in line:
-                    continue
-                key, _, val = line.partition("=")
-                key = key.strip()
-                val = val.strip()
-                # Strip inline comment (space + #)
-                if " #" in val:
-                    val = val[: val.index(" #")].strip()
-                # Strip matching surrounding quotes
-                if len(val) >= 2 and val[0] == val[-1] and val[0] in ('"', "'"):
-                    val = val[1:-1]
-                result[key] = val
-    except FileNotFoundError:
-        pass
-    return result
-
-
 def _merge_env_file(path: str, updates: dict[str, str]) -> None:
     """Merge key=value pairs into an env file.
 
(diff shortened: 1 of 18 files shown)


Lang-Akshay (changes requested):
Thanks for the PR @prakhar-singh1928 . Please address following finding . 
## Blocking Changes
| 1 | Issue / Quality | `mcpgateway/scripts/init_secrets.py` | 195 | `ensure_env_file_secrets()` regenerates only values in `WEAK_VALUES` or values starting with `__REPLACE_ME__`; arbitrary short or low-entropy `AUTH_ENCRYPTION_SECRET` values remain in `.env` even though `Settings` now hard-fails them. This was independently confirmed by the security reviewer and a smoke test. | Use the same complete compliance predicate as startup validation (empty, placeholder, known-weak, too short, and low entropy) when deciding which secrets to regenerate; verify `--patch-env` fixes each class. |

Lang-Akshay (changes requested):
## Blocking Changes

| 1 | Security | `mcpgateway/config.py` | 1241, 1540 | `min_secret_length: int = 32` is a mutable Pydantic settings field with no `ge=32` constraint; `MIN_SECRET_LENGTH=0` env var sets it to 0, causing `validate_security_combinations` (line 1540: `if len(val) < self.min_secret_length`) to skip the length check. Confirmed empirically: 16-char high-entropy (4.0) JWT/AUTH secrets pass startup with `MIN_SECRET_LENGTH=0`. This violates the PR's explicit "unconditional in every environment" contract and the security invariant that both secrets must be at least 32 chars. | Import `MIN_SECRET_LENGTH` and `MIN_ENTROPY` from `mcpgateway._security_constants` and use them as the hard lower bounds in `validate_security_combinations`. Either replace the mutable `self.min_secret_length` usage with the immutable constant, or add `ge=MIN_SECRET_LENGTH` to the field definition to prevent the bypass. |

Lang-Akshay (changes requested):
## Blocking Changes

| 1 | Quality / Data-loss defect | `mcpgateway/scripts/init_secrets.py` | 202–215 | `ensure_env_file_secrets()` uses `os.environ` precedence for `current`. When `AUTH_ENCRYPTION_SECRET` is weak in the process environment but strong in `.env`, `is_non_compliant=True` and `field in env_file_values` is True → `file_generated[field] = new_val` overwrites the strong `.env` value. For an AES encryption key this silently rotates the key without data migration, permanently making all stored encrypted credentials unreadable. Security-agent smoke run confirmed `weak_env_strong_file_rotates_file=OBSERVED`. `_PATCH_ENV_SKIP_FIELDS` existed precisely to prevent this; its removal extends the rotation risk to an encryption key where rotation without migration is irreversible. | Before writing a new value for fields in a `_ROTATION_GUARDED_FIELDS` frozenset, check if `.env` already holds a strong value. If it does, raise an actionable error ("unset the shell variable or run a migration-aware key rotation") instead of silently overwriting. Do not silently rotate an AES key. |

msureshkumar88 (changes requested):
Thanks for tackling this — restoring the unconditional `AUTH_ENCRYPTION_SECRET` guardrail is the right call, and the extra work here (shared `_security_constants.py`, the rotation guard against silently overwriting a strong secret and bricking encrypted-at-rest data, case-insensitive `.env` parsing) goes further than a plain revert in a good way. A couple of things need to be fixed before this is safe to merge, plus a few smaller items worth a look.

## Blocking

**1. `tests/scripts/test_init_secrets.py` fails to collect on this branch.**

`_read_env_file` is imported at module level (line 24) and called in ~9 tests (lines 152–407), but it was removed from `mcpgateway/scripts/init_secrets.py` in `676e2e46c` in favor of `dotenv_values`. Reproduced locally:

```
ImportError: cannot import name '_read_env_file' from 'mcpgateway.scripts.init_secrets'
Interrupted: 1 error during collection
```

The PR description's verification table says "37 passed" for this suite — looks like that ran against an earlier commit before `_read_env_file` was removed. Could you either restore a thin `_read_env_file` shim or update those 9 tests to use `dotenv_values`/the new `_env_file_ci` pattern directly?

**2. `_merge_env_file()` doesn't match the case-insensitive detection it's paired with.**

`ensure_env_file_secrets()` now detects weak values case-insensitively via `_env_file_ci`, but `_merge_env_file()` still does `if key in updates` against the always-uppercase `_SECRET_FIELDS` names. So a `.env` with `auth_encryption_secret=weak` (lowercase) gets correctly detected as weak, but the merge can't match that line — it appends `AUTH_ENCRYPTION_SECRET=<strong>` at the end and leaves the original weak lowercase line in the file. I checked pydantic-settings' behavior and last-line-wins means `Settings()` itself will still pick up the new strong value on that run, so it's not an active bypass today — but a known-weak secret ends up sitting in cleartext in `.env` indefinitely, which undercuts the case-insensitive fix in the same commit. Worth making `_merge_env_file` case-insensitive too, or normalizing keys before matching.

## Worth fixing in this PR

**3. `MIN_SECRET_LENGTH` can no longer be raised above 32.**

`3585be5d` (correctly) hardcodes the length check to `_MIN_SECRET_LENGTH` to close the `MIN_SECRET_LENGTH=0` bypass, but that also removes the ability to *raise* the floor — `.env.example` still documents `# MIN_SECRET_LENGTH=32` as a settable knob and the field keeps `ge=_MIN_SECRET_LENGTH`, implying higher values do something. Setting `MIN_SECRET_LENGTH=64` now has no effect. `max(self.min_secret_length, _MIN_SECRET_LENGTH)` in the validator would preserve "raise but not lower."

**4. Rotation-guard `ValueError` isn't caught in `main()`'s `--patch-env` branch.**

Unlike `FileExistsError`/`OSError` in the `--output` path, the new rotation-guard error from `ensure_env_file_secrets()` propagates as a raw traceback through `--patch-env` instead of the clean remediation message it carries.

**5. Docs/chart examples now contradict the restored guardrail.**

Several places still show `AUTH_ENCRYPTION_SECRET: my-test-salt` as the example value, which will now hard-fail unconditionally in every environment: `docs/docs/index.md:473`, `docs/docs/manage/configuration.md:14,133`, `docs/docs/manage/oauth-troubleshooting.md:510`, `docs/docs/architecture/oauth-authorization-code-ui-design.md:301`, `docs/docs/architecture/explorer.html:4565`, and `charts/mcp-stack/values.yaml:824` (the base Helm chart — `values-minikube.yaml` and the OCP `values-pgo.yaml` profile already use the safer empty-string "must be set" pattern, so aligning `values.yaml` the same way would help too). `.env.example` itself was updated correctly; these are the same content going stale elsewhere.

## Minor

- `_is_strong_value()` duplicates the compliance predicate that's also inlined in `ensure_env_file_secrets()` a few lines later — docstring says it "mirrors" the other check, which works but means the two have to be kept in sync by hand.
- `test_weak_auth_encryption_secret_raises_in_development` now accepts either `"too short"` or `"known-weak/default value"` as the failure reason. That's because the check order changed (length now validated before placeholder/weak/entropy), so a short placeholder-like value reports "too short" instead of "unset placeholder." Not currently hit by any fixture, but worth being aware of if a shorter placeholder variant ever shows up.

Nothing here changes the overall direction — the guardrail restoration and the rotation-guard addition are both good, this just needs the collection failure and the case-insensitive merge gap closed before it's safe to ship.

msureshkumar88 (changes requested):
Thanks for pushing the follow-up commits — all four previously-flagged blockers check out as fixed on re-verification (I reproduced each one against current code rather than relying on the commit messages):

- Regen predicate now uses the same length+entropy check as startup validation.
- `MIN_SECRET_LENGTH=0` bypass is closed (`Field(ge=_MIN_SECRET_LENGTH)` raises before the validator runs).
- Silent AES-key rotation is now blocked — a strong `.env` value won't be overwritten by a weak env-var value; raises instead.
- `test_init_secrets.py` collects again (`_read_env_file` shim restored), `_merge_env_file()` case-sensitivity is fixed, `MIN_SECRET_LENGTH` can be raised above 32 again, and the `--patch-env` `ValueError` path now gives a clean CLI message instead of a traceback.

Doc cleanup is also mostly done (7 of the stale `my-test-salt` spots fixed). A couple of things came up in a fresh pass over the current diff:

## Blocking

**1. `make check-env` (`mcpgateway/scripts/validate_env.py:151`) is out of sync with the guardrail this PR restores.**

`get_security_warnings()` doesn't apply the same WEAK_VALUES/entropy check `Settings()` now enforces unconditionally. Concretely: `AUTH_ENCRYPTION_SECRET='my-test-key-but-now-longer-than-32-bytes'` — on the `WEAK_VALUES` blocklist, ≥32 chars — comes back clean from `check-env`, but `Settings()` construction raises `SecurityConfigurationError` on it. Since `make check-env` is the PR's own documented verification step, an operator following it gets a false all-clear and then the gateway won't boot. Worth pointing this at the same predicate `config.py` uses so the two checks can't drift again.

## Worth fixing in this PR

**2. A few stale `my-test-salt` references remain.** `AGENTS.md:304` still shows it as a valid example (this file doubles as agent-facing guidance, so it's actively steering people/agents into a broken `.env`). `docs/docs/manage/configuration.md:14` still lists it in the "Required: Change Before Use" table even though the equivalent row at line 130 in the same file was already fixed — so the doc now contradicts itself. `charts/mcp-stack/README.md:761` also looks like it just needs a `helm-docs` regen to pick up the `values.yaml` fix already in this diff.

**3. The two new bypass fixes have no regression test locking them in.** Neither the `MIN_SECRET_LENGTH=0` fix nor the case-insensitive `.env` parsing fix has a test asserting the specific scenario that was broken (a `min_secret_length=0` env var reaching `validate_security_combinations`, and a lowercase-keyed `.env` line getting detected/patched). Without these, a future refactor could silently reopen either bypass.

**4. `init_secrets.py:39` hardcodes a 32-byte token / the `_MIN_SECRET_LENGTH` floor**, so it doesn't respect an operator-raised `MIN_SECRET_LENGTH`. With `MIN_SECRET_LENGTH=64` set, `make init-secrets-patch-env` still generates a 43-char token, writes it, and the gateway still fails to boot ("too short, minimum 64"). Since this PR explicitly restores the ability to raise `MIN_SECRET_LENGTH` above 32, the generator should size against the configured floor rather than the constant.

## Minor / cleanup (non-blocking)

- `mcpgateway/main.py:2113` — the `__REPLACE_ME__` placeholder check in `validate_security_configuration()` is now unreachable; `Settings()` construction raises first via `validate_security_combinations`. Not a bug, but worth a comment or removal so it doesn't read as live logic.
- `mcpgateway/config.py:1538` — `max(self.min_secret_length, _MIN_SECRET_LENGTH)` is dead now that the `Field(ge=_MIN_SECRET_LENGTH)` constraint guarantees `min_secret_length >= 32` at construction. Could simplify to `self.min_secret_length` directly.
- `mcpgateway/scripts/init_secrets.py:118` — `_read_env_file()` is no longer called by production code (`ensure_env_file_secrets()` uses `dotenv_values` directly); it's kept only so existing tests can still import it. Fine to leave, but worth a note that it's test-only now, or migrate the remaining callers off it.

Scope-wise this still looks minimal-and-correct for #6103 — no schema/Alembic changes needed (pure validation logic), backend-only, no unrelated changes snuck in. Nice work tightening the rotation-guard and case-insensitivity handling beyond a plain revert.

prakhar-singh1928:
Thanks @msureshkumar88 
**1. validate_env.py / check-env sync**
Replaced the local hardcoded weak lists with WEAK_VALUES from _security_constants.py and swapped len(set(val)) < 10 for calculate_entropy(val) < MIN_ENTROPY — both checks now use the exact same predicate as Settings.validate_security_combinations(). Also threaded effective_min = max(settings.min_secret_length, MIN_SECRET_LENGTH) through so an operator-raised floor is reflected in check-env output, not just at startup. The two code paths can no longer drift independently.

**2. Remaining my-test-salt references**
Cleared all three:

docs/docs/manage/configuration.md:14 — was contradicting the already-fixed line 130 in the same file; now consistent
AGENTS.md:304 — replaced with an empty value + generation instruction
charts/mcp-stack/README.md:761 — updated to match the values.yaml change already in the diff (helm-docs regen would have caught it at release time, but better to be consistent now)
**3. Regression tests**
Added two tests to tests/scripts/test_init_secrets.py:

test_merge_env_file_case_insensitive_key_replaced_in_place — seeds a .env with auth_encryption_secret=weak (lowercase), calls _merge_env_file with the canonical uppercase key, asserts the weak line is gone, the strong value is present, and the key appears exactly once (no duplication)
test_ensure_respects_min_secret_length_env_var — sets MIN_SECRET_LENGTH=64 via monkeypatch, calls ensure_env_file_secrets, asserts the generated tokens for both JWT_SECRET_KEY and AUTH_ENCRYPTION_SECRET are ≥ 64 chars
**4. Generator sizing against configured floor**
ensure_env_file_secrets() now reads MIN_SECRET_LENGTH from the environment and computes effective_min = max(configured, _MIN_SECRET_LENGTH). Both the non-compliance check and the generate_token() call use this floor for _STRONG_SECRET_FIELDS, so with MIN_SECRET_LENGTH=64 the script generates a token that satisfies startup validation on the first run.

msureshkumar88 (changes requested):
Thanks for the latest round — I re-verified this against current HEAD (`42876ae2f`) by reproducing each case rather than reading the commit messages, and all four items from my previous pass check out:

- `make check-env` / `validate_env.py` now shares the `_security_constants` predicate.
- The remaining `my-test-salt` references in `AGENTS.md`, `docs/docs/manage/configuration.md`, `charts/mcp-stack/README.md` and `charts/mcp-stack/values.yaml` are gone — `grep -rn "my-test-salt"` over docs/charts is clean.
- Regression tests added for the case-insensitive `.env` merge and for `MIN_SECRET_LENGTH=64`.
- `init_secrets` now sizes the generated token against the configured floor rather than the hard-coded constant.

`tests/scripts/test_init_secrets.py` + `tests/unit/mcpgateway/test_validate_env.py` — 47 passed locally.

One correction to my own previous review, since it should change what you do with that code: my "blocking" item on `validate_env.py` was wrong as I described it. `main()` constructs `Settings(_env_file=env_file)` **before** calling `get_security_warnings()`, so on this branch a weak `AUTH_ENCRYPTION_SECRET` already surfaces as `❌ Security configuration error` and exit 1 — there was never a false all-clear. The sync you applied is still reasonable defense-in-depth; no need to invest further in it.

Scope-wise this reads as the right shape for #6103: backend/config/CLI plus the docs and chart values that follow from the behavior change. No schema or Alembic change — correct, this is pure startup validation and nothing about it is persisted. No UI change, and none is needed. Nothing unrelated snuck in (`.secrets.baseline` aside).

Nothing here is a hard blocker on the default path. Requesting changes for the five items below because item 1 breaks a documented operator flow and items 3–4 leave two just-fixed bypasses without a test to hold them closed.

## Requested changes

**1. `MIN_SECRET_LENGTH` set in `.env` is ignored by the generator — `make setup` can produce an unbootable `.env`.**

`mcpgateway/scripts/init_secrets.py:240` reads the floor from `os.environ` only:

```python
configured_min: int = int(os.environ.get("MIN_SECRET_LENGTH", _MIN_SECRET_LENGTH))
```

But `.env.example:3545` and `docs/docs/manage/configuration.md:596` both document `MIN_SECRET_LENGTH` as a `.env` knob, and `Settings` reads it from `.env` — so the two ends disagree about where the value lives. Reproduced against current HEAD:

```
.env:  MIN_SECRET_LENGTH=64
       JWT_SECRET_KEY=changeme
       AUTH_ENCRYPTION_SECRET=changeme

ensure_env_file_secrets() -> {'JWT_SECRET_KEY': 43, 'AUTH_ENCRYPTION_SECRET': 43, ...}
Settings()                -> jwt_secret_key: too short (43 chars, minimum 64)
```

An operator who raises the floor in the documented place runs `make init-secrets-patch-env`, gets a cheerful `✅ Patched .env`, and the gateway still won't start. Same finding as my previous item #4, just half-closed — the shell-env path was fixed, the `.env` path wasn't. The function has already parsed `.env` into `_env_file_ci` a few lines above, so sourcing it with the same os.environ-wins precedence the rest of the loop uses would close it:

```python
_min_raw = os.environ.get("MIN_SECRET_LENGTH") or _env_file_ci.get("min_secret_length") or _MIN_SECRET_LENGTH
```

**2. Non-numeric `MIN_SECRET_LENGTH` surfaces as a rotation-guard error.**

Same line — `int(...)` is unguarded, so `MIN_SECRET_LENGTH=abc` raises `ValueError`, which `main()`'s `--patch-env` handler catches and prints as `❌ invalid literal for int() with base 10: 'abc'`. It exits 1, so it's not a crash — but that handler exists to render the rotation-guard remediation text, and two unrelated failure modes now share it. A `try/except ValueError` around the parse with its own message, or a distinct exception type for the rotation guard, would keep them apart.

**3. The `--patch-env` branch of `main()` has no test coverage.**

Every `main()` test in `tests/scripts/test_init_secrets.py` passes `patch_env=None`, so the branch this PR changes — including the new `except ValueError → exit(1)` handler, the `✅ Patched …` message and the `ℹ️ already strong` no-op path — is exercised only indirectly through `ensure_env_file_secrets()`. A couple of `argparse.Namespace(patch_env=str(tmp_env), ...)` cases would lock in the CLI contract.

**4. `MIN_SECRET_LENGTH=0` has no regression test at the config layer.**

The bypass is genuinely closed — `Settings(min_secret_length=0)` raises `ValidationError: Input should be greater than or equal to 32` — but nothing in `tests/unit/mcpgateway/test_config.py` asserts it, so a future edit to that `Field(...)` could reopen it silently. Worth one test; note it's a `ValidationError`, not `SecurityConfigurationError`, which is itself worth asserting so the exception type doesn't drift.

**5. `.env.example` now understates the requirement it's documenting.**

Two spots in the same file disagree with the guardrail this PR restores:

- Line ~1333: `# Must be a non-empty string (e.g. passphrase or random secret)` — a non-empty string is no longer sufficient; ≥32 chars and entropy ≥3.5 are enforced.
- The header block at the top says startup fails on "a placeholder (`__REPLACE_ME__`) or a known-weak default", omitting too-short and low-entropy — the two conditions most operators will actually trip.

The pre-#6073 text was accurate when the constraints were looser; reverting it verbatim brought back wording that's now incomplete.

## Breaking change — worth an explicit callout

This is a breaking change and I don't think that's in dispute, but the blast radius is worth stating in the PR body / CHANGELOG so it isn't discovered at upgrade time:

| What breaks | Who it hits | Symptom |
|---|---|---|
| Weak/short/low-entropy `AUTH_ENCRYPTION_SECRET` in `ENVIRONMENT=development` | Any dev or CI env carrying a `.env` seeded from pre-#6103 docs (`my-test-salt`) | Gateway refuses to start: `SecurityConfigurationError` |
| `MIN_SECRET_LENGTH` < 32 | Anyone who lowered the floor | `ValidationError` at `Settings()` construction — fails earlier and with a different exception type than before |
| `charts/mcp-stack/values.yaml` default now `""` | Helm installs relying on the shipped default | Pod crashloops on `secret is empty` |

All three fail closed, which is the right direction. Two things would shrink the radius:

- A CHANGELOG entry plus a line in the PR body naming the three cases and the one-line remediation (`make init-secrets-patch-env`). The error messages themselves already point at the fix, so that's most of the work done.
- On the chart side, `templates/secret-gateway.yaml` renders whatever is in `mcpContextForge.secret` verbatim, so an unset `AUTH_ENCRYPTION_SECRET` becomes an empty Secret key and the failure only shows up as a crashlooping pod. Wrapping it in `{{ required "AUTH_ENCRYPTION_SECRET must be set — generate with: make init-secrets-patch-env" ... }}` would move that to `helm install` time with an actionable message. Worth noting `values-minikube.yaml` and the OCP `values-pgo.yaml` already used the empty-string pattern, so this PR is aligning `values.yaml` with them rather than introducing the shape.

## Minor / non-blocking

**Dead code.** Four spots that are now unreachable or redundant — none are bugs, but each reads as live logic:

- `mcpgateway/scripts/validate_env.py:157-176` — the JWT/AUTH weak, length and entropy branches can't fire; `Settings()` raises first (see the correction above). The `BASIC_AUTH_PASSWORD` checks in the same function are still live.
- `mcpgateway/main.py:2113-2123` — the `__REPLACE_ME__` check in `validate_security_configuration()` is unreachable for the same reason.
- `mcpgateway/config.py:1538` — `max(self.min_secret_length, _MIN_SECRET_LENGTH)` is redundant now that `Field(ge=_MIN_SECRET_LENGTH)` guarantees the invariant at construction.
- `mcpgateway/scripts/init_secrets.py:118` — `_read_env_file()` has no production caller; it exists so the older tests still import it. It also reads the file twice (`readlines()` then `dotenv_values()`), harmless at CLI scale but odd if someone later takes it for a hot path. A one-line "test-compat shim" note would prevent that.

**Rotation guard has one uncovered path.** The guard fires for *weak shell var + strong `.env`*. The case *weak shell var + key absent from `.env`* takes the `env_only` branch and silently mints a fresh `AUTH_ENCRYPTION_SECRET` into `os.environ` — for an AES key that's the same "previously-encrypted credentials become unreadable" outcome, just without the warning. It's CLI-only (nothing calls `ensure_env_file_secrets()` at gateway startup) so the exposure is small, but a printed warning on that path would match the care already taken in the guarded case.

**`_merge_env_file()` rewrite isn't atomic.** It opens the target with `O_TRUNC` and writes in place, so an interruption mid-write leaves a truncated `.env` — including any secrets already there. Pre-existing, not introduced here, but this PR makes the function fire more often now that `AUTH_ENCRYPTION_SECRET` is back in the patch set. Write-to-temp-then-`os.replace()` would make it crash-safe; `os.replace` preserves the 0o600 you're already setting.

**Unit conflation in the token sizing.** `token_bytes = max(nbytes, effective_min)` compares a byte count against a character count. It's safe in the conservative direction (`token_urlsafe(n)` yields ~1.33n chars, so you always overshoot), but the two aren't the same unit and a future reader may "fix" it the wrong way. A comment, or `math.ceil(effective_min * 3 / 4)`, would make the intent explicit.

**Pre-existing, out of scope — possibly worth its own issue.** Several docs still instruct users to run the gateway with `JWT_SECRET_KEY=my-test-key-but-now-longer-than-32-bytes` (`tests/e2e/README.md:45`, `tests/live_gateway/protocol_compliance/README.md:36`, `tests/performance/config.yaml:4`, `docs/docs/using/clients/*.md`, `Makefile:1868`). That value is in `WEAK_VALUES` and was already rejected before this PR, so it's not a regression here — but it's the same class of staleness you just cleaned up for `my-test-salt`, and it'd be a shame to fix half of it.

Overall the guardrail restoration is solid, and the rotation guard and case-insensitivity work go beyond a plain revert in a way that clearly improves things. Items 1–5 are what I'd want closed before merge; everything under Minor is take-or-leave.

msureshkumar88 (dismissed):
**Withdrawn — this review has been replaced. Please disregard it and use https://github.com/IBM/mcp-context-forge/pull/6108#pullrequestreview-4908391126 instead.**

I've replaced the original text because two things in it were wrong, and I'd rather not leave them standing where someone might act on them:

1. **The scope question was invalid.** I asked whether auto-*generating* `AUTH_ENCRYPTION_SECRET` was required or whether detect-and-report would do, and suggested splitting the PR. #6103's acceptance criterion #2 already mandates auto-patching, and its scope table calls for removing `_PATCH_ENV_SKIP_FIELDS` — so both alternatives I floated fail the issue as written. I should have re-read the issue before asking.

2. **One blocker was mis-described.** I characterised the `env_only` branch as silently rotating an AES key with the same data-loss outcome as the guarded case. It doesn't: the generated key goes into `os.environ` of a CLI process that immediately exits, so nothing persists and nothing is destroyed. The real defect is that the command reports success, writes nothing to `.env`, and leaves the gateway unable to start — which is still blocking, but for a different reason. Detail in https://github.com/IBM/mcp-context-forge/pull/6108#issuecomment-5255764873.

What was valid here carries forward into the consolidated review: the sub-32 `MIN_SECRET_LENGTH` clamp, and confirmation that the previous round's items 1/2/4/5 are closed. Everything else was optional and has been marked as such.

Sorry for the churn — the useful version is two items long.

msureshkumar88:
@prakhar-singh1928 — pulling the scope question out of my review above so it doesn't get lost in the detail, because your answer decides whether the two remaining blockers are worth fixing at all.

**Question: does #6103 require `make init-secrets-patch-env` to auto-*generate* a new `AUTH_ENCRYPTION_SECRET`, or is detecting a non-compliant one and telling the operator to fix it enough?**

Why I'm asking: the PR splits cleanly into two halves that have behaved very differently under review.

- `config.py` (unconditional strength enforcement) — one finding across five review rounds, fixed, stable. This is the security fix the issue is about.
- `init_secrets.py` (returning `AUTH_ENCRYPTION_SECRET` to the auto-patch loop) — roughly 13 of the ~15 findings from all reviewers, including both blockers still open.

That's not a knock on the work; every fix in this thread has been correct. It's that auto-generating an AES key is data-destroying, so it has to handle three input shapes (weak in `.env` / weak in shell + strong in `.env` / weak in shell + absent from `.env`) crossed with the `MIN_SECRET_LENGTH` floor. Each round has closed one cell and the fix has added machinery with its own edges.

Three ways forward, and I'm fine with any of them:

1. **Auto-generation is required by the issue.** Then the two blockers in my review are the remaining work — both are small, and I'll re-review quickly.
2. **Detect-and-report is enough for `AUTH_ENCRYPTION_SECRET`** (with `JWT_SECRET_KEY` still auto-generating — rotating a signing key logs people out, it doesn't brick ciphertext). That deletes `_ROTATION_GUARDED_FIELDS`, `_is_strong_value`, most of the token sizing, and both blockers.
3. **Split the PR.** Land the `config.py` half now — it's done and it's the actual guardrail restoration — and take the `init_secrets.py` half to a follow-up where it can be designed against all three shapes at once rather than one per round.

If you know what the original pre-#6073 behaviour was meant to guarantee here, that probably settles it faster than anything I can infer from the diff. Happy to go with your call either way.

prakhar-singh1928:
`AUTH_ENCRYPTION_SECRET` with patch-env should patch the .env file. that was the behviour before as well

msureshkumar88:
Writing up the breaking-change surface concretely, since I've raised it twice as "worth a callout" without ever spelling it out — that's on me. I ran each case against `origin/main` and this branch (`d2b5f5f29`) side by side rather than reasoning from the diff, so the before/after columns below are observed, not inferred.

To be clear up front: **all four fail closed, which is the right direction.** This isn't an argument against the change — it's the material for a CHANGELOG entry and a note in the PR body so it isn't discovered at upgrade time.

## 1. Weak `AUTH_ENCRYPTION_SECRET` in `ENVIRONMENT=development` now aborts startup

The headline change, and the intent of #6103.

```
Settings(environment="development", auth_encryption_secret="my-test-salt", ...)

main   -> ACCEPTED (starts up, warning only)
branch -> SecurityConfigurationError: auth_encryption_secret: too short (12 chars, minimum 32).
          Run 'python -m mcpgateway.scripts.init_secrets' to generate a strong value.
```

**Who it hits:** any dev machine or CI job carrying a `.env` seeded from the pre-#6103 docs — `my-test-salt` was the documented example in six places until this PR fixed them. Production was already enforcing, so this lands entirely on dev/CI.
**Remediation:** `make init-secrets-patch-env`. The error message already says this.

## 2. `MIN_SECRET_LENGTH` below 32 is now rejected at construction

```
Settings(min_secret_length=16, <strong secrets>)

main   -> ACCEPTED, min_secret_length = 16
branch -> ValidationError: min_secret_length — Input should be greater than or equal to 32
```

Two things change together, and the second is the one that'll surprise people: it fails **earlier** (at `Settings()` construction, from the `Field(ge=...)` constraint) and with a **different exception type** (`ValidationError`, not `SecurityConfigurationError`). Anything catching `SecurityConfigurationError` to render a friendly startup error won't catch this.

**Who it hits:** anyone who lowered the floor to accommodate a short secret. Probably rare, but silent until boot.
**Remediation:** remove the override, or raise it to ≥32 — raising still works, which is what my earlier item asked for.

## 3. Helm chart default changed from a working value to `""`

```
values.yaml, main:   AUTH_ENCRYPTION_SECRET: my-test-salt
values.yaml, branch: AUTH_ENCRYPTION_SECRET: ""   # REQUIRED: ... generate with: make init-secrets-patch-env
```

This one has two distinct victim classes, and the first is easy to miss:

- **Users on the shipped default.** They never touched `values.yaml`, but the default they were inheriting was `my-test-salt`, which case 1 now rejects. Their upgrade breaks whether or not they change anything.
- **Users who explicitly pinned `my-test-salt`** in their own values — same rejection, but at least it's their line.

Either way the symptom is a **crashlooping pod**, because `templates/secret-gateway.yaml` renders `mcpContextForge.secret` verbatim: an unset key becomes an empty Secret value and the failure only surfaces at container start, not at `helm install`. Wrapping it in `{{ required "AUTH_ENCRYPTION_SECRET must be set — generate with: make init-secrets-patch-env" ... }}` would move that to install time with an actionable message. (Worth noting `values-minikube.yaml` and the OCP `values-pgo.yaml` already used the empty-string pattern, so this PR aligns `values.yaml` with them rather than inventing the shape.)

## 4. `make init-secrets-patch-env` can now exit non-zero where it previously succeeded

This is the one I haven't mentioned before, and it's the one most likely to break CI:

```
env:   AUTH_ENCRYPTION_SECRET=changeme        (weak, from the shell)
.env:  AUTH_ENCRYPTION_SECRET=<strong value>

main   -> exit 0, "✅ Patched .env: generated strong values for JWT_SECRET_KEY, BASIC_AUTH_PASSWORD"
branch -> exit 1  (rotation guard)
```

On `main` the field was excluded from the patch loop entirely, so this combination was a no-op and the command always succeeded. On this branch the rotation guard correctly refuses to overwrite the strong `.env` value — but `make setup` calls this, so any pipeline with a weak `AUTH_ENCRYPTION_SECRET` exported in its environment now fails at setup rather than at boot. Neither branch rotates the `.env` value, so there's no data loss in either — it's purely a new non-zero exit.

I think the guard is right and the exit code is right. It just needs to be in the release notes, because "make setup started failing" is a confusing symptom when the fix is "unset a shell variable."

## Suggested wording for the PR body / CHANGELOG

| What breaks | Who it hits | Symptom | Fix |
|---|---|---|---|
| Weak/short/low-entropy `AUTH_ENCRYPTION_SECRET` in `ENVIRONMENT=development` | Dev/CI envs with a `.env` seeded from pre-#6103 docs (`my-test-salt`) | `SecurityConfigurationError` at startup | `make init-secrets-patch-env` |
| `MIN_SECRET_LENGTH` < 32 | Anyone who lowered the floor | `ValidationError` at `Settings()` — earlier, and a different exception type than before | Remove the override, or set ≥32 |
| `charts/mcp-stack/values.yaml` default now `""` | Helm installs on the shipped default *or* pinning `my-test-salt` | Crashlooping pod on empty/weak secret | Set `AUTH_ENCRYPTION_SECRET` in your values |
| Weak `AUTH_ENCRYPTION_SECRET` exported in the shell while `.env` holds a strong one | CI pipelines running `make setup` | `make init-secrets-patch-env` exits 1 (rotation guard) | `unset AUTH_ENCRYPTION_SECRET` |

Most of the work is already done — the error messages carry their own remediation, which is the part that usually gets skipped. Also worth ticking the "Breaking change" box in the PR template; it's currently marked `Chore`, and cases 1 and 3 are more than that.

msureshkumar88:
Withdrawing my scope question above — I should have re-read #6103's acceptance criteria before asking it.

Auto-patching isn't optional there: criterion #2 is "`make init-secrets-patch-env` / `make setup` auto-patches `AUTH_ENCRYPTION_SECRET` into `.env` when the current value is non-compliant", and the scope table explicitly says to remove `_PATCH_ENV_SKIP_FIELDS`. So options 2 (detect-and-report) and 3 (split the PR) both fail the issue as written. Please disregard them — option 1 is the only one on the table, which means **the two blockers in my review are the remaining work** and they're both small.

The observation behind the question still stands — most findings in this thread have come from the `init_secrets.py` half, because auto-generating an AES key has more input shapes than it looks like. But that's an argument for covering the third shape (my blocker #2), not for changing the approach.

What I should have asked instead, now posted on the issue (IBM/mcp-context-forge#6103): that issue lists "migration scripts for `AUTH_ENCRYPTION_SECRET` must be authored and validated" as a **blocking prerequisite** for this chore, and I can't find them in or linked from this PR. The rotation guard here is a refusal-to-rotate rather than a migration. If those scripts exist, linking them would settle whether the guard is a stopgap or the final answer; if they were dropped as unnecessary, saying so would close that line out. Either way it doesn't change the two blockers.

msureshkumar88:
Correcting my blocker #2 — I described it as a silent AES-key rotation with the same data-loss outcome as the guarded case. That's wrong, and the actual behaviour matters more for this PR than what I claimed. I ran the real CLI in a child process rather than calling `ensure_env_file_secrets()` in-process, which is what I should have done the first time.

## What actually happens

Weak `AUTH_ENCRYPTION_SECRET` in the shell, **no** `AUTH_ENCRYPTION_SECRET` line in `.env`:

```
$ AUTH_ENCRYPTION_SECRET=changeme python -m mcpgateway.scripts.init_secrets --patch-env .env
✅  Patched .env: generated strong values for AUTH_ENCRYPTION_SECRET, BASIC_AUTH_PASSWORD
$ echo $?
0

# durable state after the process exits:
AUTH_ENCRYPTION_SECRET present in .env : False

# next gateway start, same shell:
SecurityConfigurationError: auth_encryption_secret: too short (8 chars, minimum 32).
```

Control — same thing with the weak value living in `.env` instead of the shell:

```
✅  Patched .env: generated strong values for AUTH_ENCRYPTION_SECRET, BASIC_AUTH_PASSWORD
AUTH_ENCRYPTION_SECRET present in .env : True
next gateway start                     : BOOTS OK
```

So nothing is destroyed — my data-loss framing was wrong, and I'd rather say so than leave it standing. The generated key is written to `os.environ` of a CLI process that exits a moment later, so it evaporates. What's left is a command that reports success, changes nothing durable, and leaves the gateway unable to start.

## Why this is still blocking, and arguably more so

This is #6103's acceptance criterion #2 not being met:

> `make init-secrets-patch-env` / `make setup` auto-patches `AUTH_ENCRYPTION_SECRET` into `.env` when the current value is non-compliant

In this shape the current value *is* non-compliant and it is *not* patched into `.env`. The success message names `AUTH_ENCRYPTION_SECRET` explicitly, because `main()` prints the merged `{**env_only, **file_generated}` dict and can't distinguish the two. An operator following the documented remediation — the exact string our own `SecurityConfigurationError` tells them to run — gets a green check and a gateway that still won't boot, with no indication why.

That's the same false-success shape as the `MIN_SECRET_LENGTH` items, and it's the third time it's turned up in this PR. Might be worth a general pass over `main()`'s reporting: it currently reports on *generation*, not on *persistence*, and those have quietly diverged.

## The design question underneath it

The `env_only` branch at `init_secrets.py:265` is deliberate, and the comment says why:

```python
# Weak value came from os.environ; patching environ is enough.
# Writing to .env would shadow subsequent env-var injections (Docker/K8s).
```

That rationale is sound in the abstract, but I don't think it applies here: `ensure_env_file_secrets()` has exactly one caller — `main()` under `--patch-env` (I grepped; there's no startup or import-time caller, so the "running process picks them up without restart" line in its docstring is also vestigial). `--patch-env` is an explicit operator action against a named file. If someone runs `--patch-env .env`, writing to `.env` is the thing they asked for; a container that injects `AUTH_ENCRYPTION_SECRET` via env vars isn't running this command as part of its boot path.

Three ways to close it, in my order of preference:

1. **Write to `.env` for `_STRONG_SECRET_FIELDS` even when the weak value came from the shell.** Satisfies criterion #2 directly, and matches what `--patch-env` claims to do. The shadowing concern is real for `BASIC_AUTH_PASSWORD`-style values in container setups, so keeping `env_only` for the non-strong fields would be a reasonable split.
2. **Keep skipping the write, but stop calling it success.** Exit non-zero with something like: *"AUTH_ENCRYPTION_SECRET is weak in your environment and absent from .env; .env was not modified. Either `unset AUTH_ENCRYPTION_SECRET` and re-run, or export a strong value."* Fails criterion #2 as written, but at least it's honest and the operator can act.
3. **Report generation vs persistence separately** in the success message, so `✅ Patched` never names a key that wasn't written. Worth doing regardless of 1 or 2, since it's the part that turns a gap into a trap.

Option 1 is the only one that satisfies the issue as written, so that's what I'd suggest unless the shadowing case is more load-bearing than it looks from here — in which case option 2 plus a note on the issue that criterion #2 can't be fully met would be the honest outcome.

Blocker #1 (the sub-32 `MIN_SECRET_LENGTH` clamp) is unaffected by any of this and still stands as written.

msureshkumar88 (changes requested):
I've left a lot of comments on this PR across several rounds, and that's made it harder than it should be to see what's actually needed. Consolidating: **two changes are required to merge. Everything else I've raised is optional — please disregard it unless you want it.**

Both are in `mcpgateway/scripts/init_secrets.py`, both reproduced against `d2b5f5f29`.

## 1. `MIN_SECRET_LENGTH` below 32 is silently clamped (line 241–249)

`max(configured_min, _MIN_SECRET_LENGTH)` accepts a sub-floor value and never reports it, but `Settings` has `Field(ge=_MIN_SECRET_LENGTH)` and rejects it:

```
.env: MIN_SECRET_LENGTH=0   (same for -5)

$ make init-secrets-patch-env
✅  Patched .env: generated strong values for JWT_SECRET_KEY, AUTH_ENCRYPTION_SECRET, BASIC_AUTH_PASSWORD

Settings(_env_file=".env")
  -> ValidationError: min_secret_length — Input should be greater than or equal to 32
```

```python
_min_env = os.environ.get("MIN_SECRET_LENGTH")
_min_raw = _min_env if _min_env is not None else (_env_file_ci.get("min_secret_length") or str(_MIN_SECRET_LENGTH))
try:
    configured_min: int = int(_min_raw)
except ValueError:
    raise ValueError(
        f"MIN_SECRET_LENGTH={_min_raw!r} is not a valid integer. "
        "Set it to a whole number (e.g. MIN_SECRET_LENGTH=64)."
    ) from None
if configured_min < _MIN_SECRET_LENGTH:
    raise ValueError(
        f"MIN_SECRET_LENGTH={configured_min} is below the enforced minimum "
        f"of {_MIN_SECRET_LENGTH}. Settings will reject it at startup."
    )
effective_min: int = configured_min
```

The `is not None` also handles `MIN_SECRET_LENGTH=""`, and `max()` on the following line becomes dead once the guard is in.

## 2. Weak secret in the shell + key absent from `.env` reports success but writes nothing (line 265)

```
$ AUTH_ENCRYPTION_SECRET=changeme make init-secrets-patch-env
✅  Patched .env: generated strong values for AUTH_ENCRYPTION_SECRET, BASIC_AUTH_PASSWORD
$ echo $?
0

AUTH_ENCRYPTION_SECRET in .env : False
next gateway start             : SecurityConfigurationError: too short (8 chars, minimum 32)
```

The `env_only` branch writes to `os.environ` of a CLI process that then exits, so the key is discarded. This is #6103's acceptance criterion #2 not being met — the value is non-compliant and doesn't get patched into `.env`. Restricting `env_only` to the non-strong fields fixes it:

```python
if env_val is not None and field.lower() not in _env_file_ci and field not in _STRONG_SECRET_FIELDS:
    # Non-strong fields (BASIC_AUTH_PASSWORD): patch environ only — writing to
    # .env would shadow env-var injection in container setups.
    env_only[field] = new_val
elif env_val is not None and field in _ROTATION_GUARDED_FIELDS and _is_strong_value(_env_file_ci.get(field.lower(), ""), weak_values):
    raise ValueError(...)  # unchanged
else:
    file_generated[field] = new_val
```

The existing shadowing rationale still holds for `BASIC_AUTH_PASSWORD`; it doesn't apply to `--patch-env` on the strong fields, since that's an explicit operator action against a named file and `ensure_env_file_secrets()` has no startup caller.

## Not required

Everything else in my previous comments — the breaking-change writeup, dead code, non-atomic `_merge_env_file`, `token_bytes` units, the helm `required` wrapper, test hygiene, the `main()` reporting split, extra test coverage. Some of it is worth doing eventually, none of it should hold this PR. My scope question is already withdrawn.

Happy to re-review as soon as the two above are in.

msureshkumar88 (dismissed):
## E2E verification — approving

Ran a real end-to-end verification against this PR's head (`4f17fd0ca`), not mocks: real `uv venv`, real subprocess CLI invocations of `init_secrets`/`validate_env`, real `Settings()` construction, and a real `uvicorn` boot of `mcpgateway.main:app` against a fresh SQLite DB built by the actual Alembic migration chain. Script: `scripts/verify-pr-6108-e2e.sh` (local artifact, not pushed — reproducible with `gh pr checkout 6108` + the commands below).

### Config used
- Isolated git worktree at PR head `4f17fd0ca`, untouched working copy elsewhere
- `uv venv .venv && uv pip install --group dev -e .` (Python 3.13.12)
- SQLite (`DATABASE_URL=sqlite:///...`), no external services — matches the scope of this PR (pure startup-validation/CLI logic, nothing DB-schema-related)

### Commands run (10 real-environment cases, each isolated with `env -i`)
```
# Case 1 — weak AUTH_ENCRYPTION_SECRET in ENVIRONMENT=development must hard-fail
ENVIRONMENT=development AUTH_ENCRYPTION_SECRET=my-test-salt JWT_SECRET_KEY=<strong> \
  python -c "from mcpgateway.config import Settings; Settings()"

# Case 2 — strong secrets boot cleanly in development / staging / production
# Case 3 — MIN_SECRET_LENGTH=0 rejected at Settings construction (ValidationError)
# Case 4 — `init_secrets --patch-env` regenerates a weak value already in .env, then gateway boots against the patched file
# Case 5 — weak AUTH_ENCRYPTION_SECRET in shell, key ABSENT from .env → now persisted to .env (previously fixed blocker)
# Case 6 — weak AUTH_ENCRYPTION_SECRET in shell, STRONG value in .env → rotation guard refuses, .env untouched
# Case 7 — MIN_SECRET_LENGTH=0 in .env → raises actionable error instead of silently clamping (previously fixed blocker)
# Case 8 — MIN_SECRET_LENGTH=64 honoured end-to-end by the generator and by Settings()
# Case 9 — `make check-env` / validate_env.py agrees with Settings() (no false all-clear)
# Case 10 — real `uvicorn mcpgateway.main:app` boot with generated strong secrets against a fresh SQLite DB
python -m uvicorn mcpgateway.main:app --env-file .env --port <port>
curl http://127.0.0.1:<port>/health
```

Plus the target + collateral pytest suites (blast-radius check for anything this PR could break outside `init_secrets.py`/`config.py`/`validate_env.py`):
```
pytest tests/scripts/test_init_secrets.py tests/unit/mcpgateway/test_config.py \
       tests/unit/mcpgateway/test_validate_env.py tests/unit/test_cwe321_hardcoded_jwt.py \
       tests/test_config_security.py tests/unit/mcpgateway/services/test_support_bundle_service.py
```

### Output
- **Cases 1–9: all PASS, deterministic across 3 independent runs.** Every acceptance criterion in #6103 and both blockers fixed in the last review round (`MIN_SECRET_LENGTH<32` no longer silently clamped; weak-shell/absent-from-.env now persists) reproduce correctly against real subprocess behavior, not just unit fixtures.
- **Case 10 (full process boot):** the gateway process itself boots correctly with strong secrets and serves `/health` → `200 {"status":"healthy",...}` — verified directly. The *automated* harness intermittently times out on this case in this sandbox because a from-scratch SQLite + full Alembic index-creation pass (100+ indexes) took 90–150s+ under contended CPU here — unrelated to this PR's logic (same cost on `main`). Not treating this as a finding against the PR; flagging only as a harness note if you reuse the script.
- **Collateral test suites: 339 passed, 0 failed** across the 6 files above — nothing this PR touches (`_security_constants.py` extraction, `WEAK_VALUES`/entropy/length predicate changes) broke `test_cwe321_hardcoded_jwt.py`, `test_config_security.py`, or the unrelated `test_support_bundle_service.py`.

### Verdict
No functional gaps found. Confirms what the review thread already converged on: the two consolidated blockers are genuinely fixed and regression-tested, scope stays minimal-for-#6103 (no schema/Alembic change, backend+CLI+docs only), and no unrelated changes are in the diff.

Two non-blocking items I'd still like to see before/at merge, both docs-only and already discussed on this thread — not holding the approval on them:
1. Tick "Breaking Change" in the PR template and paste the blast-radius table already written up in this thread (dev-mode secret rejection, `MIN_SECRET_LENGTH<32` exception-type change, Helm chart default now `""`, `make setup` new non-zero exit on a weak shell var).
2. Helm `values.yaml`: wrap `AUTH_ENCRYPTION_SECRET` in `required{...}` so an unset value fails at `helm install` instead of a crashlooping pod.

Approving.

mcpgateway/scripts/init_secrets.py
```diff
@@ -195,24 +226,63 @@ def ensure_env_file_secrets(
...
+            # Size the generated token against the operator-configured floor so it
+            # always satisfies Settings.validate_security_combinations() on startup.
+            token_bytes = max(nbytes, effective_min) if field in _STRONG_SECRET_FIELDS else nbytes
+            new_val = generate_token(token_bytes)
+            if env_val is not None and field.lower() not in _env_file_ci and field not in _STRONG_SECRET_FIELDS:
+                # Non-strong field (e.g. BASIC_AUTH_PASSWORD) with a weak value from
+                # os.environ only — patching environ is enough; writing to .env would
+                # shadow env-var injections in container setups.
                 env_only[field] = new_val
+            elif env_val is not None and field in _ROTATION_GUARDED_FIELDS and _is_strong_value(_env_file_ci.get(field.lower(), ""), weak_values):
```
madhu-mohan-jaishankar: The new strong-field trigger here widens the replacement predicate from **"is this a known placeholder?"** to `"does this meet the 32-char + entropy bar?"` .  This means, a real, in-use `AUTH_ENCRYPTION_SECRET `that's merely short now gets regenerated.

The rotation guard here doesn't cover it. All three of its conditions are ANDed, and the first is `env_val `is not `None `, a shell override. In the common case (no shell override, .env holds a working secret) we skip :266 and :271 on their first condition and land in the bare else at :287, which writes a fresh key to disk. 

`main()` prints Patched `.env` — no warning. Since this key feeds `get_encryption_service(...) `(token_storage_service.py:198, gateway_service.py:1075, tool_service.py:5868), every stored OAuth token and auth header becomes undecryptable and it is exactly what the guard's own comment at warns about.

rotating without re-encrypting doesn't improve security either. The ciphertext in the DB is unchanged and still protected by the old key, so an attacker with a dump is no worse off. we only lose our own access. And import_service.py:587 (the one re-encryption path) needs the current key.

**Suggested fix - guard on any pre-existing value, not just strong ones:**
```
elif field in _ROTATION_GUARDED_FIELDS and (_env_file_ci.get(field.lower()) or "").strip():
    raise ValueError(...)  # same actionable message
```
That keeps the legitimate bootstrap case working. If auto-rotation is wanted, it should need an explicit --rotate-encryption-secret flag paired with a re-encryption step.

tests/unit/mcpgateway/test_config.py
```diff
@@ -2156,46 +2141,22 @@ def test_weak_secret_raises_in_staging():
...
-                _env_file=None,
-            )
+    with pytest.raises(SecurityConfigurationError) as exc_info:
+        Settings(
+            auth_encryption_secret="my-test-salt",  # nosec B106  # pragma: allowlist secret
+            environment="development",
+            _env_file=None,
+        )
+    msg = str(exc_info.value)
+    assert "too short" in msg or "known-weak/default value" in msg
```
madhu-mohan-jaishankar, line 2159: "my-test-salt" is 12 chars, so it hits the length floor and never reaches the known-weak check (as the docstring notes). The deleted `test_weak_auth_encryption_secret_raises_outside_development `was the only test covering that branch. It used a ≥32-char value on purpose, and across staging/production. Worth restoring that shape.

tests/unit/mcpgateway/test_config.py
```diff
@@ -2090,31 +2090,16 @@ def test_primary_worker_heartbeat_warns_at_exact_boundary(caplog):
 
 
 def test_placeholder_secret_raises_in_development():
-    """__REPLACE_ME__ placeholder is rejected in development for jwt_secret_key.
-
-    For auth_encryption_secret the placeholder is permitted in development
-    (emits a WARNING) — consistent with all other non-compliant values.
-    """
```
madhu-mohan-jaishankar, line 2097: The `auth_encryption_secret `placeholder case was deleted rather than flipped to a `pytest.raises`, and nothing replaced it. no test anywhere passes `__REPLACE_ME__` for this field. Could you add one?

madhu-mohan-jaishankar (changes requested):

Lang-Akshay (dismissed):
LGTM ✅

madhu-mohan-jaishankar (dismissed):
LGTM

Lang-Akshay (dismissed):
LGTM ✅

Lang-Akshay (approved):
LGTM

PR #6109, opened 2026-08-06
Add explicit RBAC permissions to OAuth registered-client management

### LINKED ISSUE #5981: [FEATURE]: Add explicit RBAC permissions to OAuth registered-client management
### Type of Feature

### Observation

OAuth registered-client management routes rely on an admin flag and do not
declare an explicit `require_permission` check:

- `GET /oauth/registered-clients`
- `GET /oauth/registered-clients/{gateway_id}`
- `DELETE /oauth/registered-clients/{client_id}`

Other management routes commonly combine identity checks with named RBAC
permissions. Adding equivalent permission checks would make these routes
consistent with that model and allow access decisions to be represented in
role configuration.

### Goal

Define and enforce explicit RBAC permissions for reading and deleting
registered OAuth clients.

### User Story

**As a:** platform operator

**I want:** registered-client operations to use named permissions

**So that:** roles clearly express who may view or remove registered clients

### Acceptance Criteria

```gherkin
Scenario: Caller has registered-client read permission
  Given an eligible admin with the required read permission
  When the caller lists or retrieves registered OAuth clients
  Then the request proceeds

Scenario: Caller lacks registered-client read permission
  Given an eligible admin without the required read permission
  When the caller lists or retrieves registered OAuth clients
  Then the request is rejected with 403

Scenario: Caller lacks registered-client delete permission
  Given an eligible admin without the required delete permission
  When the caller deletes a registered OAuth client
  Then the request is rejected with 403
  And no database mutation occurs
```

- [ ] Choose permission names consistent with existing OAuth and gateway
      permissions.
- [ ] Map all three routes to explicit permissions.
- [ ] Preserve expected platform-admin behavior.
- [ ] Add allow-path and deny-path tests.
- [ ] Document permissions in RBAC reference material.

### Alternatives Considered

Continuing to rely only on `is_admin` keeps behavior simple but leaves these
routes outside the named-permission model used by other management surfaces.

msureshkumar88:
## Summary

Adds explicit, named RBAC permissions for the OAuth registered-client (DCR) management routes, which previously relied only on an `is_admin` check with no permission representable in role configuration.

Closes #5981

## Background

Three routes manage OAuth clients dynamically registered via RFC 7591 DCR:

- `GET /oauth/registered-clients`
- `GET /oauth/registered-clients/{gateway_id}`
- `DELETE /oauth/registered-clients/{client_id}`

They call `_require_unnarrowed_admin()`, which checks `is_admin` and rejects team-narrowed or public-only tokens (registered clients are stored globally with no team column, so a team-scoped token has no coherent scope over them). That guard is necessary but not sufficient: it has no named permission, so it can't be granted, withheld, or audited through role configuration the way every other management surface in the gateway can.

## Design decisions

**Permission names: `admin.oauth_clients:read` / `admin.oauth_clients:delete`.** Follows the existing `admin.sso_providers:<action>` convention rather than introducing a new top-level category. Registered clients are global resources, so a team-scoped category like `gateways.*` would be the wrong shape — `team_admin` and `developer` already hold `gateways.read`/`gateways.delete`, and reusing those would silently widen access to a resource those roles have no business touching.

**`allow_admin_bypass=False` on the `@require_permission` decorators.** `_require_unnarrowed_admin` already requires `is_admin`, so with the default bypass behavior every caller that reaches the RBAC check would short-circuit past it and the new permission would be decorative. Disabling bypass is what makes the permission real — matches the pattern already used on admin-UI routes (`allow_admin_bypass=False` appears ~200 times in `admin.py`).

**Both RBAC layers, not just one.** Layer 1 (token-scoping middleware) gets path→permission mappings so a scoped API token can carry `admin.oauth_clients:*`; Layer 2 (the route decorator) enforces the permission against the caller's roles. `/oauth/registered-clients*` previously had no Layer 1 mapping at all, so scoped tokens already hit a default-deny — the new mapping only *loosens* access for tokens that explicitly carry the permission, it doesn't change existing denial behavior.

**`_require_unnarrowed_admin` stays, unchanged, inside each handler.** RBAC has no way to express "this token must not be team-narrowed" — that's a scope-coherence property, not a permission. Removing it in favor of the new permission would let a narrowed admin token pass once it also carries the permission, which is wrong for a globally-scoped resource. The two checks compose: token scope → RBAC permission → scope-coherence guard.

## Breaking change

Admins provisioned through every supported path (`platform_admin` role, assigned `["*"]`) are unaffected — this covers `bootstrap_db.py` seeding, `create_platform_admin()`, UI promotion, and SSO provisioning. An admin whose `is_admin` flag was set without a role assignment (e.g. direct SQL), or whose `DEFAULT_ADMIN_ROLE` points at a custom role with no inherited path to `*`, will start getting 403 on these three routes.

`docs/docs/manage/rbac.md` has a new troubleshooting section with a SQL query (tested against both SQLite and PostgreSQL) that lists every affected admin, so operators can size their exposure before upgrading rather than discover it after.

## What's out of scope

DCR client *creation* happens implicitly during gateway/OAuth setup and stays gated by `gateways.create`/`gateways.update` — issue #5981 scopes this to read/delete on the three management routes above, not the creation path.


msureshkumar88:
## E2E verification

Ran a real gateway process (`uv run uvicorn`, SQLite, real JWTs, real `curl` calls — no mocks) against this branch to exercise the two new RBAC permissions end to end. Script: `scripts/verify-pr-6109-e2e.sh`.

### Config used

```bash
export JWT_SECRET_KEY="e2e-pr6109-jwt-secret-min32chars-000000"
export AUTH_ENCRYPTION_SECRET="e2e-pr6109-encryption-secret-min32chars-000000"
export DATABASE_URL="sqlite:///./e2e_pr6109_test.db"
export AUTH_REQUIRED=true
export EMAIL_AUTH_ENABLED=true
export MCPGATEWAY_ADMIN_API_ENABLED=true
export MCPGATEWAY_UI_ENABLED=true
export PLUGINS_ENABLED=false
export HOST=127.0.0.1
export PORT=46109
export ADMIN_REQUIRE_PASSWORD_CHANGE_ON_BOOTSTRAP=false

uv run uvicorn mcpgateway.main:app --host 127.0.0.1 --port 46109 &
```

Seeded via ORM (no HTTP path exists for these): an `orphan-admin@example.com` user with `is_admin=True` and **zero** role assignments (simulating direct-SQL promotion / a `DEFAULT_ADMIN_ROLE` with no path to `*`), and a `dev-with-permission@example.com` user with `is_admin=False` holding `admin.oauth_clients:read`/`:delete` via a custom role. JWTs minted with `mcpgateway.utils.create_jwt_token` (`--admin`, `--teams`, `--scopes`) to cover unnarrowed/narrowed/scoped-token cases.

### Output

```
== 4. [THE FIX] platform admin (unnarrowed, platform_admin role) -> 200 on all 3 routes ==
PASS: GET /oauth/registered-clients -> 200
PASS: GET /oauth/registered-clients/{gateway_id} -> 200

== 5. [BREAKING CHANGE under test] is_admin=True, zero roles -> 403 on read ==
   Before this PR, is_admin alone was sufficient here (200).
PASS: orphan admin (no role assignment) correctly blocked (403)

== 6. [_require_unnarrowed_admin regression, pre-existing] narrowed admin token -> 403 ==
PASS: team-narrowed admin token correctly blocked (403)

== 7. [two-layer composition] RBAC permission granted but is_admin=False -> 403 ==
   dev-with-permission holds admin.oauth_clients:read/delete via role but is_admin=False;
   _require_unnarrowed_admin's is_admin gate must still block.
PASS: non-admin with RBAC permission still blocked (403)

== 8. [Layer 1 token scoping] scoped token without admin.oauth_clients:* -> 403 ==
PASS: scoped token missing admin.oauth_clients:read correctly blocked (403)

== 9. [Layer 1 token scoping] scoped token WITH admin.oauth_clients:read -> 200 read, 403 delete ==
PASS: read-scoped token can read (200)
PASS: read-scoped token correctly blocked from delete (403)

== 10. Get registered client for nonexistent gateway -> 404 ==
PASS: nonexistent gateway_id returns 404

== 11. Platform admin deletes the registered client -> 200, row actually gone ==
PASS: platform admin delete succeeds (200)
PASS: registered client row actually deleted from DB

== 12. Delete already-deleted client id -> 404 ==
PASS: deleting an already-deleted client returns 404

== 13. [surrounding regression] unauthenticated request -> 401 ==
PASS: unauthenticated request correctly rejected (401)

== Summary ==
ALL CHECKS PASSED
```

Covers: the new permission gate working for a properly-provisioned `platform_admin` (200), the documented breaking change firing for an `is_admin`-only user with no role (403), the pre-existing `_require_unnarrowed_admin` narrowing guard still composing on top of the new RBAC check (403), the RBAC-permission-without-`is_admin` case (403 — confirms the two layers are independent, not either-or), both Layer 1 token-scoping directions (missing scope → 403, granted read scope → 200 but still 403 on delete), and 404/401 edge cases. 18/18 checks green.

Full script: `scripts/verify-pr-6109-e2e.sh`.

vishu-bh (changes requested):
Code looks clean. Core of issue #5981 is addressed, but two medium-severity authorization/documentation problems remain

1. Team-scoped role can authorize global OAuth-client operations
New decorators use generic permission resolution in [oauth_router.py (line 1143)](/private/tmp/mcp-pr-6109-review/mcpgateway/routers/oauth_router.py:1143). Without explicit team context, RBAC aggregates permissions from team roles. Result:
Team A role can grant listing/deletion of globally stored DCR clients.
GET {gateway_id} derives gateway team, while list/delete aggregate across teams.
Same permission behaves differently across three routes.
Documentation acknowledges inconsistency in [rbac.md (line 514)](/private/tmp/mcp-pr-6109-review/docs/docs/manage/rbac.md:514), but recommendation is not enforcement.
Fix: require these permissions from global-scope roles only. Add tests proving team-scoped grants fail for all three endpoints.
2. upgrade audit SQL misses partially provisioned admins
Query in [rbac.md (line 762)](/private/tmp/mcp-pr-6109-review/docs/docs/manage/rbac.md:762) excludes user if they have either read or delete:
"*" OR read OR delete
But full compatibility requires wildcard or both permissions. Read-only admins still fail DELETE; delete-only admins still fail both GET routes, yet query reports neither.
Fix query to require wildcard, or independently confirm both read and delete across effective roles.

msureshkumar88:
Thanks for the review, @vishu-bh — both points check out and are fixed in the latest commits.

**1. Team-scoped role authorizing global OAuth-client operations**

You're right that the three routes resolved team scope inconsistently: `GET /oauth/registered-clients/{gateway_id}` derived team context from the gateway resource, while list/delete fell through to `check_any_team=True` (aggregating across all of the caller's team-scoped roles). A role granted on a single team could satisfy list/delete but not the per-gateway route for that same gateway.

Added a `global_only` option to `require_permission()` (`mcpgateway/middleware/rbac.py`) that skips team derivation entirely — `team_id=None`, `check_any_team=False` — so only global/personal roles (plus team roles with `scope_id=NULL`) are evaluated. All three registered-client routes now pass `global_only=True`. Added deny-path tests confirming a team-scoped role (and a derivable gateway-owning team) no longer satisfies the permission on any of the three routes:
- `tests/unit/mcpgateway/middleware/test_rbac.py::TestGlobalOnlyPermission`
- `tests/unit/mcpgateway/routers/test_oauth_router.py::test_registered_client_routes_check_global_scope_only`

Docs updated in `docs/docs/manage/rbac.md` to describe the enforced behavior instead of just recommending it.

**2. Upgrade audit SQL undercounting partially-provisioned admins**

Confirmed — the old query excluded a user from the affected list if *any* of wildcard/read/delete matched, so a read-only or delete-only admin was silently reported as compatible even though they'd still 403 on the permission they lack. Rewrote the query to aggregate permissions per user (`user_perm_flags` CTE) and require wildcard OR both `:read` AND `:delete` together. It also now only counts global/personal roles (plus team-scope-id-NULL roles), matching the `global_only` enforcement from point 1 — a team-scoped grant doesn't count toward compatibility either.

Pushed as part of the same commit series; PR is rebased on current main and green on the local RBAC/oauth_router test suites + ruff/interrogate.

vishu-bh (dismissed):
LGTM 🚀

Lang-Akshay (dismissed):
LGTM ✅

Lang-Akshay (approved):
LGTM

PR #5979, opened 2026-07-29
docs: document CSRF protection, middleware ordering, and LLM admin endpoints (#5959)

msureshkumar88:
## Summary

Closes #5959 — documentation and test coverage gaps identified during review of #5780 (the CSRF fix for LLM Settings writes).

**Documentation:**
- New CSRF Protection section in `docs/docs/manage/configuration.md`: all 12 `CSRF_*` settings, and the fact that two independent CSRF implementations exist (`CSRFMiddleware`, settings-driven, vs. `enforce_admin_csrf`, which hardcodes its own cookie/header names and does not track the `CSRF_COOKIE_*` settings).
- New `docs/docs/architecture/middleware-ordering.md`: Starlette's reverse-registration rule, the full middleware stack, and the CSRF-before-auth ordering constraint that #5780 fixed.
- New LLM Settings Management section in `docs/docs/manage/api-usage.md`: both LLM route families (`/v1/llm/*` vs `/v1/admin/llm/*`), their different CSRF schemes, and the per-path validation matrix.

**Tests:**
- Wrong-`session_id` CSRF binding coverage (the other half of the `(user_id, session_id)` HMAC pair).
- Expanded JS missing-CSRF-cookie coverage from 1 to all 11 Admin UI write call sites.
- e2e coverage for the `enforce_admin_csrf` path against `/admin/llm/*` and `/v1/admin/llm/*` (nothing previously issued a real request against this path — only presence-of-dependency was asserted).
- Expired (two-windows-old) CSRF token rejection, at both the service and middleware levels.
- CSRF cookie-name drift guards across the JS literal, `admin.py`, and `oauth_router.py`'s hardcoded constants.

## Known issue surfaced (not fixed here)

The e2e coverage task confirmed a real, reachable CSRF-validation divergence: in the window between `/admin/login` and the first admin dashboard load, `/admin/llm/*` and `/v1/admin/llm/*` disagree for the identical request, because the CSRF cookie set at login is an opaque non-HMAC placeholder that satisfies `enforce_admin_csrf`'s plain double-submit check but fails `CSRFMiddleware`'s HMAC check on the versioned mount. This is pinned with a dedicated regression test, not fixed — filed as #5978 for a follow-up.

## Test plan

- [x] `tests/unit/mcpgateway/middleware/test_admin_csrf_binding.py` — 20/20 passing
[...]

PR #6111, opened 2026-08-06
fix(admin): bind login CSRF cookie to email and session jti

### LINKED ISSUE #5978: [BUG]: /admin/llm/* and /v1/admin/llm/* diverge on CSRF validation immediately after login

msureshkumar88:
Closes #5978

## Problem

`admin_login_handler` set the CSRF cookie via `_set_admin_csrf_cookie(request, response)` with no `user_id`/`session_id`, so it fell through to the opaque `secrets.token_urlsafe(32)` branch. That value satisfies `enforce_admin_csrf`'s plain `compare_digest` double-submit check but fails `CSRFService.validate_csrf_token()`, which requires a 64-char hex HMAC.

Because `/admin` sits in `csrf_exempt_paths` and the exemption is prefix-matched against the raw request path, only the versioned mount runs `CSRFMiddleware`. For the same session, same cookie, same header, same Origin, a write to `/admin/llm/*` succeeded while the identical write to `/v1/admin/llm/*` returned 403 `CSRF_TOKEN_INVALID` — for the whole window between login and the first dashboard load, which is the only place the cookie was rotated to its bound value.

## Approach

Option 1 from the issue: bind the cookie at login rather than exempting `/v1/admin` from the middleware. Option 2 would have made the mounts agree by *removing* the HMAC layer from the versioned mount, and would have silently stripped the only CSRF control from `/v1/admin/runtime/*`.

Binding now happens at every point a session JWT is minted, matching what `routers/auth.py`, `routers/email_auth.py` and `admin_ui()` already did:

- `admin_login_handler`, normal branch
- `admin_login_handler`, forced-password-change branch
- `change_password_required_handler`, which re-mints with a new `jti` and would otherwise leave the cookie bound to the superseded login session

Unauthenticated page renderers (`GET /admin/login`, forgot-password, reset-password, change-password-required) intentionally keep the unbound token — there is no session to bind to, and neither CSRF layer engages without a session cookie.

## Second gap closed

`runtime_admin_router` was mounted without `enforce_admin_csrf`. Its legacy path is inside the `/admin` exempt prefix, so `PATCH /admin/runtime/{mcp,a2a}-mode` had **no** CSRF control at all on the unprefixed mount. The `config.py` comment asserting "all admin routes use per-route enforce_admin_csrf" was inaccurate; it is now true and enforced by a test.

## Tests

- Direct-call tests pinning both branches of `_set_admin_csrf_cookie`, including that the unbound branch produces a token `CSRFMiddleware` genuinely rejects, plus deny-paths for foreign user and foreign session.
- Source guards over `admin_login_handler` and `change_password_required_handler`, following the existing idiom in this file.
- A structural guard that walks the assembled admin route table and fails on any state-changing `/admin` route lacking `enforce_admin_csrf`. It first failed with exactly the two runtime PATCH routes, then passed once they were wired. Note it resolves FastAPI 0.140's deferred `_IncludedRouter` wrappers — `include_router` no longer flattens into `router.routes`, and the naive walk silently matches zero routes, so there is a `checked > 50` assertion guarding against a vacuous pass. It currently covers 93 admin write routes.
- End-to-end mount parity added to the Playwright security suite, per the relocation in #6008.

The regression tests named in the issue are not on `main` — #5979 added them and #6008 removed them the next day as part of moving full-stack CSRF coverage out of the unit tier. The bug was unpinned until now.

## Behaviour change worth noting

Adding `enforce_admin_csrf` to the runtime routes means a caller authenticating **by session cookie** that omits `X-CSRF-Token` or a matching `Origin` now gets 403 on `PATCH /admin/runtime/{mcp,a2a}-mode`. Bearer-token callers are unaffected (`enforce_admin_csrf` returns early without a `jwt_token` cookie), which covers the live compliance suite and `scripts/compliance_matrix.py`. The Admin UI does not call these routes, and the documented cookie recipes in `api-usage.md` already send both headers.

## Known residuals (not fixed here)

1. `/admin/**` still never HMAC-validates — `enforce_admin_csrf` does origin + plain double-submit only, so it accepts tokens the versioned mount would reject.
2. `enforce_admin_csrf` and `CSRFMiddleware` disagree on which credentials make a request CSRF-relevant (`jwt_token` only vs. also `access_token`/proxy identity).
3. With `TOKEN_EXPIRY > 120`, the cookie's `max_age` can outlive the HMAC's acceptance window, reintroducing the divergence late in a long session. Pre-existing; documented in `configuration.md`.
4. `oauth_router.py` writes an unbound CSRF cookie when none is present. A bound one survives its regex check untouched.

Follow-up issues will be filed for these once this merges.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^\\.secrets\\.baseline$",
     "lines": null
   },
-  "generated_at": "2026-08-12T15:30:22Z",
+  "generated_at": "2026-08-13T09:04:01Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -1858,7 +1858,7 @@
         "hashed_secret": "5193dda17da630c041fe949c06e7b6dd5ac8628f",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1612,
+        "line_number": 1615,
         "type": "Base64 High Entropy String",
         "verified_result": null
       },
@@ -1866,7 +1866,7 @@
         "hashed_secret": "91174c4e6859da80df4e7ce9d45b35501f591210",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1895,
+        "line_number": 1898,
         "type": "Base64 High Entropy String",
         "verified_result": null
       },
@@ -1874,7 +1874,7 @@
         "hashed_secret": "4a0a2df96d4c9a13a282268cab33ac4b8cbb2c72",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1936,
+        "line_number": 1939,
         "type": "Secret Keyword",
         "verified_result": null
       }
@@ -1946,7 +1946,7 @@
         "hashed_secret": "a2166e0b243f4e192e14000a6aa8f46cde6bfc20",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 882,
+        "line_number": 900,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -1954,7 +1954,7 @@
         "hashed_secret": "fa9beb99e4029ad5a6615399e7bbae21356086b3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1389,
+        "line_number": 1407,
         "type": "Basic Auth Credentials",
         "verified_result": null
       },
@@ -1962,7 +1962,7 @@
         "hashed_secret": "fa9beb99e4029ad5a6615399e7bbae21356086b3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1393,
+        "line_number": 1411,
         "type": "Secret Keyword",
         "verified_result": null
       }
--- a/docs/docs/manage/api-usage.md
+++ b/docs/docs/manage/api-usage.md
@@ -1279,6 +1279,9 @@ curl -s -X POST \
   $BASE_URL/v1/admin/llm/providers/provider-123/state
 ```
 
+!!! tip "Capturing `$CSRF_COOKIE`"
+    The `mcpgateway_csrf_token` cookie is HMAC-bound to your session from the moment you log in, so capture it straight from the `POST /admin/login` response — there is no need to load `/admin/` first. Before [#5978](https://github.com/IBM/mcp-context-forge/issues/5978) the login response returned an unbound token that the `/v1/admin/**` mount rejected until a dashboard load rotated it.
+
 ### Check Provider Health
 
 Verify provider API connectivity and response time.
--- a/docs/docs/manage/configuration.md
+++ b/docs/docs/manage/configuration.md
@@ -665,7 +665,25 @@ The five settings marked as "middleware-path-only" in the table above govern onl
 !!! info "CSRF_EXEMPT_PATHS and Versioned Route Interaction"
     The middleware exemption uses prefix matching on the raw request path (e.g., `/admin` matches `/admin/llm/*` but not `/v1/admin/llm/*`). This means versioned admin routes at `/v1/admin/*` are validated by both the middleware and the per-route `enforce_admin_csrf` dependency (double validation), while legacy routes at `/admin/*` use only the per-route dependency (exempt from middleware). Cross-validate your paths against both implementations. See [Middleware Ordering and Stacking](../architecture/middleware-ordering.md) for details on how CSRF middleware interacts with other middleware and per-route dependencies.
 
-    This double-validation is also where a known timing gap surfaces: in the window between `/admin/login` and the first dashboard load, the versioned mount's extra `CSRFMiddleware` pass can reject a write that the legacy mount's `enforce_admin_csrf`-only path accepts, because the CSRF cookie has not yet rotated from its opaque pre-login value to its HMAC-bound one. See [IBM/mcp-context-forge#5978](https://github.com/IBM/mcp-context-forge/issues/5978).
+    This double-validation used to expose a timing gap: in the window between `/admin/login` and the first dashboard load, the versioned mount's extra `CSRFMiddleware` pass rejected writes that the legacy mount's `enforce_admin_csrf`-only path accepted, because the CSRF cookie had not yet rotated from its opaque pre-login value to its HMAC-bound one. Fixed in [IBM/mcp-context-forge#5978](https://github.com/IBM/mcp-context-forge/issues/5978) — see *Admin CSRF token lifecycle* below.
+
+#### Admin CSRF token lifecycle
+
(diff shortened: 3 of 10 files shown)


Lang-Akshay (changes requested):
## Blocking Changes

| 1 | Security | `mcpgateway/admin.py` | 1859-1863 | `enforce_admin_csrf()` returns unless `jwt_token` is present. The authentication stack also accepts `access_token` cookies, so an authenticated browser using only that ambient cookie can submit state-changing requests without Origin or double-submit validation, including the newly protected runtime-admin routes. CWE-352, High. | Treat every accepted ambient session-cookie name, at minimum `jwt_token` and `access_token`, as CSRF-relevant and require origin plus CSRF-token validation. Preserve bearer-header-only API behavior only when no ambient auth cookie is present. |
| 2 | Security | `mcpgateway/admin.py` | 4559-4680 | `POST /admin/login` is middleware-exempt and has no pre-auth CSRF nonce/token check. A cross-site form can force a victim browser to authenticate into an attacker-selected account (login CSRF/session confusion). CWE-352, Medium. | Issue a pre-auth nonce on the login page and require/validate it on login POST, or otherwise add an equivalent pre-auth CSRF control. |

msureshkumar88:
Thanks for catching both of these — fixed in the latest commit (e3c649e2c).

**1. `access_token` cookie bypass (admin.py:1859-1863)** — `enforce_admin_csrf` now checks `jwt_token` *or* `access_token` before deciding CSRF is irrelevant, matching the ambient-cookie fallback already used elsewhere in `admin.py`/`main.py`. Bearer-header-only requests (no session cookie at all) still skip CSRF as before.

**2. Login CSRF (admin.py:4559-4680)** — `enforce_admin_csrf` now also runs for `POST /admin/login` even with no session cookie present, validating against the pre-auth nonce `admin_login_page`'s `GET` already mints via `_set_admin_csrf_cookie`. A cross-site login POST with no prior same-origin GET has no cookie to double-submit against and is rejected with 403.

Added regression coverage in `tests/unit/mcpgateway/middleware/test_admin_csrf_binding.py` for: `access_token`-only sessions (reject without CSRF pair, pass with a valid one), and login POST (reject with no nonce, reject with origin-only, pass with a valid pre-auth nonce). Full admin/CSRF test files pass locally.

Lang-Akshay (dismissed):
LGTM

gandhipratik203 (dismissed):
LGTM!

Lang-Akshay (approved):
LGTM

PR #6112, opened 2026-08-06
feat(ui-rewrite): enable virtual server state actions

vishu-bh:
# Pull Request

## 🔗 Related Issue

N/A

---

## 📝 Summary

Enables activation and deactivation from virtual server cards using the explicit server state endpoint. Deactivation now requires confirmation, while activation runs directly. Inactive servers remain visible so they can be reactivated, and card status dots show green for enabled servers and red for disabled servers.

State mutations patch the server list from the canonical API response, prevent concurrent delete/state operations, surface localized success or error feedback, and include translations for all supported gateway locales.

---

## 📏 Reviewability

---

## 🏷️ Type of Change

- [ ] Bug fix
- [x] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

| Check | Command | Status |
|---|---|---|
| TypeScript | `npx tsc -b --pretty false` | Passed |
| Focused unit/integration tests | `npx vitest run src/api/virtualServers.test.ts src/components/gateways/VirtualServerCard.test.tsx src/components/gateways/GatewayCards.test.tsx src/pages/Gateways.test.tsx` | 89 passed |
| Focused Playwright E2E | `npx playwright test e2e/virtual-servers.spec.ts --grep "activates a disabled virtual server\|requires confirmation to deactivate\|opens server actions dropdown\|displays server details correctly"` | 4 passed |
| Changed-file lint | `npx eslint` on changed TypeScript files | Passed |
| Formatting | `npx prettier --check` on changed files | Passed |

---

## ✅ Checklist

- [x] Code formatted

---

## 📓 Notes (optional)

Targets `epic/ui-rewrite` because this work extends the virtual server card introduced on that branch.

Added a screen record

https://github.com/user-attachments/assets/e336233a-1fbe-4e96-ab43-1235e35774cf


marekdano (changes requested):
Solid, well-scoped PR. The state mutation is pessimistic (waits for the canonical API response before patching the cache), concurrency between delete/toggle is properly cross-guarded, feedback is localized across all three locales, and
  test coverage is genuinely good. Findings below are grouped by category and ordered by severity.

  ---

  ## Should fix

  ### 1. Stale / vacuous e2e assertion — `enabled-indicator` no longer exists
  `client/e2e/virtual-servers.spec.ts:1066`

  The PR renames the indicator test id from `enabled-indicator` → `status-indicator` everywhere **except** the `"handles disabled server correctly"` test:

  ```ts
  await expect(card.getByTestId("enabled-indicator")).toHaveCount(0);
  ```

  Since the component never emits `enabled-indicator` anymore, this now passes **vacuously** — it no longer verifies the disabled state. Worse, it encodes the *old* behavior (disabled = no dot), which the PR intentionally reverses (disabled now shows a red dot). It should mirror the unit tests:

  ```ts
  await expect(card.getByTestId("status-indicator")).toHaveClass(/bg-red-500/);
  await expect(card.getByLabel("Disabled")).toBeVisible();
  ```

  ---

  ### 2. `aria-label` on a bare `<span>` isn't reliably announced
  `client/src/components/gateways/VirtualServerCard.tsx:78-87`

  ```tsx
  <span
    className={cn("size-1.5 rounded-full", server.enabled ? "bg-emerald-500" : "bg-red-500")}
    data-testid="status-indicator"
    aria-label={intl.formatMessage({ id: server.enabled ? "gateways.card.enabled" : "gateways.card.disabled" })}
  />
  ```

  A `<span>` has an implicit *generic* role, and per the ARIA spec `aria-label` is not guaranteed to be exposed on generic elements — most screen readers drop it. The status therefore isn't conveyed non-visually. Adding `role="img"` (or `role="status"`) makes the label a first-class accessible element. This was a latent issue before, but the PR now relies on this label to distinguish enabled vs. disabled, so it's worth fixing here.

  > Note: the tests (`getByLabelText` / `getByLabel`) pass regardless because Testing Library reads the attribute directly, not the computed a11y tree — so they don't catch this gap.

## 💡 Minor notes (non-blocking)

  - **Details-panel sync is effectively dead** (`client/src/pages/Gateways.tsx:642`). `setServerState` calls `setDetailsServer(... updated ...)`, but `VirtualServerDetailsPanelContainer` computes `hydratedServer = serverDetails?.id ===
  serverId ? serverDetails : server` — it prefers its *own* `useQuery` fetch, so the patched `detailsServer` prop is only a fallback and never wins. The panel's Status row (`VirtualServerDetailsPanel.tsx:658`) can show stale enabled state if toggled while open. Low practical impact (the card sits behind the sheet), but the sync line gives a false sense of correctness.
  - **Error-extraction inconsistency.** Toggle uses `extractApiErrorDetail(err.body)` (matching the tags handler) while the sibling delete flow uses `sanitizeError(err)`. Both are valid; just noting the divergence within the same file.
  - **`role="alertdialog"`** — `ConfirmDialog` supports it for destructive confirms but neither deactivate nor the existing delete dialog passes it. Consistent with current code, so not introduced here.

vishu-bh:
Closing this as PR opened in new repo

PR #6115, opened 2026-08-06
fix(ui-rewrite): stop profile menu shifting header content

a-effort:
**Bug** 

What was happening: 
Layout shift on opening the profile dropdown / changing theme

Details: 
The avatar / theme dropdown (`HeaderProfileMenu`) used Radix's default modal`DropdownMenu`. A modal Radix menu wraps its content in `react-remove-scroll`, which locks the body on open (`overflow: hidden` plus a compensating `padding-right` equal to the scrollbar width). That padding injection nudges centered header content sideways every time the menu opens or closes.

This is a separate cause from #5946 (`scrollbar-gutter: stable` on `html`): that fix addresses the document-level scrollbar toggling, but the modal scroll-lock still measures a nonzero gap and pads the body regardless of page height.

**Change**

- Set `modal={false}` on the profile `DropdownMenu`. This disables Radix's scroll-lock, so no compensating body padding is injected and centered content no longer jumps. The page stays scrollable while the menu is open and outside-click still closes it (no focus trap / background-scroll block, both fine for this menu).
- Add a regression test asserting the body gets no `data-scroll-locked` / `overflow: hidden` while the menu is open. Confirmed it fails with `modal={true}` (Radix sets `data-scroll-locked="1"`) and passes with `modal={false}`.

**Verification**

- Visually confirmed: no horizontal shift on open/close.
- `HeaderProfileMenu.test.tsx`: 8/8 passing.

fixed: 
https://github.com/user-attachments/assets/c0ab1e5d-ce9e-43c0-be15-c9750f53ec39

### DIFF
--- a/client/src/components/layout/HeaderProfileMenu.test.tsx
+++ b/client/src/components/layout/HeaderProfileMenu.test.tsx
@@ -114,6 +114,21 @@ describe("HeaderProfileMenu", () => {
     expect(localStorage.getItem("theme-preference")).toBe("system");
   });
 
+  it("does not scroll-lock the body while the menu is open", async () => {
+    // Regression: a modal dropdown wraps its content in react-remove-scroll,
+    // which locks the body (overflow:hidden + compensating padding) on open and
+    // shifts centered header content sideways. modal={false} must avoid this.
+    const user = userEvent.setup();
+    renderMenu();
+
+    await user.click(screen.getByRole("button", { name: "Bobo Example" }));
+    // Menu is open (its items are rendered).
+    expect(screen.getByText("Settings")).toBeInTheDocument();
+
+    expect(document.body).not.toHaveAttribute("data-scroll-locked");
+    expect(document.body.style.overflow).not.toBe("hidden");
+  });
+
   it("renders null when user is not logged in", () => {
     mockUser = null;
     const { container } = renderMenu();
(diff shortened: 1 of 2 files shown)


a-effort:
LLM reviewed and generated this feedback: 
-->

Summary
This is a well-executed, focused bug fix with appropriate test coverage.

What's Good
1. Minimal, targeted change: Single-line fix (`modal={false}`) addresses the root cause
2. Clear problem statement: PR description explains the Radix modal dropdown behavior and how it differs from the scrollbar-gutter fix (#5946)
3. Solid test coverage: Regression test checks both `data-scroll-locked` attribute and `overflow: hidden` style
4. Appropriate solution: Profile menu doesn't need modal behavior (focus trap/scroll lock) - it's a simple action menu

Observations
- Only 1 of 14 DropdownMenu instances in the codebase uses `modal={false}` after this change
- TeamSwitcher (sidebar) and other dropdowns remain modal by default
- This is appropriate: header dropdowns are more sensitive to content shift than sidebar/page dropdowns

Verdict
No blocking issues. The fix is correct, well-tested, and appropriately scoped. Ready to merge.

marekdano (approved):
Looks good!

LGTM 🚀

vishu-bh (approved):
Code is clean and solid 

LGTM ☘️

PR #6123, opened 2026-08-07
fix(config): ignore empty env vars for boolean Settings fields

### LINKED ISSUE #6122: [BUG]: docker compose exec crashes with pydantic ValidationError on EXPERIMENTAL_RUST_MCP_*_ENABLED (empty string not valid bool)
### 🐞 Bug Summary
Running any one-off command via `docker compose exec gateway ...` (e.g. the
`create_jwt_token` step from the 5-minute setup / README) crashes with a
Pydantic `ValidationError` on seven `EXPERIMENTAL_RUST_MCP_*_ENABLED` settings.
The gateway service itself boots fine — only commands that bypass
`docker-entrypoint.sh` are affected.

Root cause: `docker-compose.yml` sets these vars as
`${EXPERIMENTAL_RUST_MCP_*_ENABLED:-}`, defaulting to an empty string when
unset (which is the default in `.env.example`). `docker-entrypoint.sh`
resolves that empty string into a real `true`/`false` based on `RUST_MCP_MODE`
before launching gunicorn, but that resolution only happens inside the
entrypoint's own process tree. `docker compose exec` spawns a new process
entrypoint's own process tree. `docker compose exec` spawns a new process
directly against the container's original env (still `""`), and
`mcpgateway/config.py`'s `Settings` model has no `env_ignore_empty` config, so
Pydantic treats `""` as an invalid bool literal instead of falling back to the
field default.

---

### 🧩 Affected Component
- [ ] `mcpgateway` - API
- [ ] Other (explain below)

---

### 🔁 Steps to Reproduce
1. Follow the 5-minute setup: `cp .env.example .env`, then `docker compose up -d`.
2. Leave the `EXPERIMENTAL_RUST_MCP_*_ENABLED` vars unset (default/commented out in `.env.example`).
3. Run the README's step 7 token-generation command:
   ```bash
   docker compose exec gateway python3 -m mcpgateway.utils.create_jwt_token \
     --username admin@example.com --exp 10080 --secret "$JWT_SECRET_KEY"
4. Observe the crash (any docker compose exec gateway python3 -m ... command reproduces it, not just this one).

---
🤔 Expected Behavior

docker compose exec commands should work out of the box after the standard
.env.example setup, without requiring manual -e VAR=false overrides.
Empty-string env vars for boolean settings should fall back to the field
default rather than raising a validation error.

---
📓 Logs / Error Output

  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/app/mcpgateway/utils/create_jwt_token.py", line 71, in <module>
    DEFAULT_ALGO: str = settings.jwt_algorithm
                        ^^^^^^^^^^^^^^^^^^^^^^
  File "/app/mcpgateway/config.py", line 4020, in __getattr__
    return getattr(get_settings(), key)
                   ^^^^^^^^^^^^^^
  File "/app/mcpgateway/config.py", line 3963, in get_settings
    cfg = Settings(**kwargs)
          ^^^^^^^^^^^^^^^^^^
  File "/app/mcpgateway/config.py", line 3846, in __init__
    super().__init__(**kwargs)
pydantic_core._pydantic_core.ValidationError: 7 validation errors for Settings
experimental_rust_mcp_runtime_enabled
  Input should be a valid boolean, unable to interpret input [type=bool_parsing, input_value='', input_type=str]
experimental_rust_mcp_session_core_enabled
  Input should be a valid boolean, unable to interpret input [type=bool_parsing, input_value='', input_type=str]
experimental_rust_mcp_event_store_enabled
  Input should be a valid boolean, unable to interpret input [type=bool_parsing, input_value='', input_type=str]
experimental_rust_mcp_resume_core_enabled
  Input should be a valid boolean, unable to interpret input [type=bool_parsing, input_value='', input_type=str]
experimental_rust_mcp_live_stream_core_enabled
  Input should be a valid boolean, unable to interpret input [type=bool_parsing, input_value='', input_type=str]
experimental_rust_mcp_affinity_core_enabled
  Input should be a valid boolean, unable to interpret input [type=bool_parsing, input_value='', input_type=str]
experimental_rust_mcp_session_auth_reuse_enabled
  Input should be a valid boolean, unable to interpret input [type=bool_parsing, input_value='', input_type=str]

---
🧠 Environment Info

| Key | Value |
|-|-|
| Version or commit | main@4a1e3eadb                       |
| Runtime           | Python 3.12, Gunicorn (in-container) |
| Platform / OS     | macOS (Darwin 25.5.0)                |
| Container         | Docker Compose                       |

---
🧩 Additional Context (optional)

- Suggested fix: add env_ignore_empty=True to the SettingsConfigDict in
mcpgateway/config.py (model_config = SettingsConfigDict(..., env_ignore_empty=True)).
This makes Pydantic treat an empty env string the same as "unset" and fall
back to the field default, without touching docker-entrypoint.sh's
RUST_MCP_MODE preset-resolution logic (which relies on the same
empty-string sentinel and runs before Python starts).
- A naive fix of defaulting these vars to false directly in
docker-compose.yml or via a Dockerfile ENV would break
RUST_MCP_MODE=shadow/edge/full auto-derivation, since the entrypoint
script can no longer distinguish "unset, use mode preset" from "explicitly
set to false".
- Affects the exact command documented in README.md (step 7,
"Generate an API token") and docs/docs/index.md.

gcgoncalves:
## 📌 Summary

Fixes `docker compose exec` crashing with a Pydantic `ValidationError` on
seven `EXPERIMENTAL_RUST_MCP_*_ENABLED` settings, including the
`create_jwt_token` command documented in the README's 5-minute setup. The
gateway service itself was never affected — only one-off commands run via
`docker compose exec`.

Closes #6122

## 🔁 Reproduction Steps

1. Follow the 5-minute setup: `cp .env.example .env`, `docker compose up -d`.
2. Leave `EXPERIMENTAL_RUST_MCP_*_ENABLED` unset (the `.env.example` default).
3. Run:
   ```bash
   docker compose exec gateway python3 -m mcpgateway.utils.create_jwt_token \
     --username admin@example.com --exp 10080 --secret "$JWT_SECRET_KEY"
   ```
4. Crash with 7 `bool_parsing` validation errors.

## 🐞 Root Cause

`docker-compose.yml` passes the `EXPERIMENTAL_RUST_MCP_*_ENABLED` vars as
`${VAR:-}`, which resolves to an empty string when unset. `docker-entrypoint.sh`
resolves that empty string into a real `true`/`false` based on `RUST_MCP_MODE`
before launching gunicorn, but that resolution only happens inside the
entrypoint's own process tree. `docker compose exec` bypasses the entrypoint
and spawns a new process directly against the container's original env
(still `""`). `mcpgateway/config.py`'s `Settings` model had no
`env_ignore_empty` config, so Pydantic rejected the empty string as an
invalid bool literal instead of falling back to the field default.

## 💡 Fix Description

Set `env_ignore_empty=True` on the `Settings` model's `SettingsConfigDict`
(`mcpgateway/config.py`). This tells Pydantic to treat an empty env var the
same as unset and fall back to the field default, rather than failing
validation.

Considered and rejected: defaulting these vars to `false` directly in
`docker-compose.yml` or via a Containerfile `ENV`. Both would break
`RUST_MCP_MODE=shadow/edge/full` auto-derivation, since `docker-entrypoint.sh`
relies on the same empty-string sentinel to distinguish "unset, derive from
mode preset" from "explicitly set to false" — a compose/image-level default
removes that distinction. Fixing it at the Pydantic layer preserves the
entrypoint's mode-preset logic untouched.

## 📏 Reviewability

## 🧪 Verification

|Check|Command|Status|
|-|-|-|
|Lint suite|`make lint`| |
|Unit tests|`make test`| |
|Coverage ≥ 80 %|`make coverage`| |
|Manual regression no longer fails|`docker compose exec gateway python3 -m mcpgateway.utils.create_jwt_token --username admin@example.com --exp 10080 --secret "$JWT_SECRET_KEY"` with `EXPERIMENTAL_RUST_MCP_*_ENABLED` unset| |

## ✅ Checklist

### DIFF
--- a/mcpgateway/config.py
+++ b/mcpgateway/config.py
@@ -3106,7 +3106,7 @@ Disallow: /
     # ===================================
 
     slug_refresh_batch_size: int = Field(default=1000, description="Batch size for gateway/tool slug refresh at startup")
-    model_config = SettingsConfigDict(env_file=".env", env_file_encoding="utf-8", case_sensitive=False, extra="ignore")
+    model_config = SettingsConfigDict(env_file=".env", env_file_encoding="utf-8", case_sensitive=False, extra="ignore", env_ignore_empty=True)
 
     gateway_tool_name_separator: str = "-"
     valid_slug_separator_regexp: ClassVar[str] = r"^(-{1,2}|[_.])$"


mcpgateway/config.py
```diff
@@ -3106,7 +3106,7 @@ def validate_ratelimiter_redis_url(cls, v: Optional[str]) -> Optional[str]:
     # ===================================
 
     slug_refresh_batch_size: int = Field(default=1000, description="Batch size for gateway/tool slug refresh at startup")
-    model_config = SettingsConfigDict(env_file=".env", env_file_encoding="utf-8", case_sensitive=False, extra="ignore")
+    model_config = SettingsConfigDict(env_file=".env", env_file_encoding="utf-8", case_sensitive=False, extra="ignore", env_ignore_empty=True)
```
ebarkhordar, line 3109: I ran this against 3be88e8f in a clean python:3.12-slim container. It does fix the seven `EXPERIMENTAL_RUST_MCP_*_ENABLED` fields, but `env_ignore_empty=True` sits on `model_config`, so it applies to all 771 fields on `Settings`, and one of them is a production safety check.

`get_security_status` at `config.py:1834` treats an empty string as a sentinel meaning "not configured" and refuses to start in production. Once empty values are ignored, a blanked `BASIC_AUTH_PASSWORD` resolves to its default `changeme` instead, which is a weak value but not a sentinel, so the check passes:

```
$ python repro_6123.py   # BASIC_AUTH_PASSWORD="", ENVIRONMENT=production, MCPGATEWAY_UI_ENABLED=true
BASE 4a1e3ead   resolved=''           sentinel=True    get_security_status()=FAIL      start blocked
HEAD 3be88e8f   resolved='changeme'   sentinel=False   get_security_status()=SUCCESS   start allowed
```

`JWT_SECRET_KEY` and `AUTH_ENCRYPTION_SECRET` are not affected. Their defaults begin with `__REPLACE_ME__`, which the same check catches on its own. `basic_auth_password` is the one critical secret whose default is neither empty nor a replace-me marker.

This needs an operator who sets the variable to empty rather than leaving it unset. `.env.example` ships `BASIC_AUTH_PASSWORD=changeme`, so it is not the out-of-the-box path, and the guard above the loop skips the check entirely unless the UI or basic auth is enabled.

Scoping it to the fields you meant keeps the fix and leaves the check alone:

```python
@field_validator(
    "experimental_rust_mcp_runtime_enabled",
    "experimental_rust_mcp_session_core_enabled",
    "experimental_rust_mcp_event_store_enabled",
    mode="before",
)
@classmethod
def _empty_means_unset(cls, v):
    return False if v == "" else v
```

Issue #5835, opened 2026-07-24
[CHORE]: Remove outdated tests/compliance folder (superseded by tests/live_gateway/protocol_compliance)


jonpspri:
## Summary

The `tests/compliance/` directory (containing the `mcp_2025_11_25` compliance suite) is outdated. MCP protocol compliance testing has been consolidated into `tests/live_gateway/protocol_compliance/`, which is now the canonical location (run via `make test-protocol-compliance`).

Keeping both trees around is confusing: the old suite is excluded from the default pytest run, and it duplicates effort against the live-gateway suite.

## Scope of removal

- [ ] Delete `tests/compliance/` (including `tests/compliance/mcp_2025_11_25/` and its `conftest.py`).
- [ ] Clean up Makefile references — the `MCP_2025_TEST_DIR` variable (line ~39) and the compliance targets that consume it (around lines 1177–1204) either need to be removed or repointed at `tests/live_gateway/protocol_compliance/`.
- [ ] Remove the now-unneeded `--ignore=tests/compliance` entry from the pytest `addopts` in `pyproject.toml` (line ~698).
- [ ] Grep for any remaining docs/CI references to `tests/compliance` and update or remove them.

## Acceptance criteria

- `tests/compliance/` no longer exists in the repo.
- `make test`, `make ruff interrogate pylint`, and the pre-merge validation gate targets all pass without the folder.
- No dangling references to `tests/compliance` remain in the Makefile, `pyproject.toml`, CI workflows, or docs.

jonpspri:
A quick status update on this one:

The underlying cleanup described here has not landed on `main` yet, so the issue is still actionable.

Remaining work:
- Remove the legacy [`tests/compliance/`](tests/compliance) tree from the repository.
- Remove the stale MCP 2025 compliance variables and targets still present in [`Makefile`](Makefile:39) and the related target block around [`Makefile`](Makefile:1143).
- Remove the obsolete pytest ignore and marker registration still present in [`pyproject.toml`](pyproject.toml:664) and [`pyproject.toml`](pyproject.toml:704).
- Sweep docs / CI / repo references for `tests/compliance`, `MCP_2025`, `mcp20251125`, and any now-stale protocol compliance references, then update or drop them.
[...]

PR #6023, opened 2026-07-31
chore(tests): replace fastmcp with the official mcp SDK in live-gateway tests

jonpspri:
## Summary

Removes the dev-only `fastmcp` test dependency from the live-gateway suites and standardizes on the official `mcp` Python SDK client, unblocking the SDK 2.0 packaging simplification tracked in the linked issue.

- **Dependencies**: drop `fastmcp` (and its `authlib` transitive pin) from the dev dependency group; regenerate `uv.lock`.
- **Removed suites**: delete the fastmcp-based protocol-compliance harness (`tests/live_gateway/protocol_compliance/`) along with its `make test-protocol-compliance*` targets, and the fastmcp-based go/rust fast-server parity test.
(_Note:_ To the best of my knowledge, I've been the only user of this suite and
our plan is to move to using Anthropic's compliance suites.  I'm all for reducing\
our maintenance footprint.  The Go/Rust parity test is no longer needed -- covered
in integration tests against fast-*-server.)
- **Rewritten suites**: `test_mcp_protocol_e2e.py` and `test_mcp_rbac_transport.py` now drive `mcp.client.streamable_http.streamablehttp_client` + `ClientSession` directly — bearer auth via plain `Authorization` headers, fast-fail client timeouts preserved, and FastMCP result accessors mapped to the SDK's (`is_error` -> `isError`, `call_tool_mcp` -> `call_tool`, `ping` -> `send_ping`).
- **Docs**: live-gateway READMEs, package docstring, and Makefile help text now describe the `mcp` SDK client.

One implementation note for reviewers: pytest-asyncio drives async-generator fixture setup and teardown in separate tasks, which breaks the anyio task-group context managers inside `streamablehttp_client`/`ClientSession` (`RuntimeError: Attempted to exit cancel scope in a different task`). FastMCP's `Client` never hit this because it runs the session in its own task; the rewritten `client` fixture reproduces that by running the session lifecycle in a dedicated runner task and handing the session to the test.

## Completeness pass

Follow-up commits chase down what the harness left behind, so no orphaned tooling or stale guidance survives the removal:

- **Reference server + matrix tooling**: the protocol-compliance harness was the only consumer of `mcp-servers/python/compliance_reference_server/` and `scripts/compliance_matrix.py`; both are removed.
[...]

Issue #6022, opened 2026-07-31
[TESTING]: CI lane live-compliance and compliance-testing docs


Lang-Akshay:
Part of epic #6018. Depends on #6020.

## What

- Add `live-compliance` lane to `scripts/cf-integration.sh`: runs `make conformance-check`, stores results under `reports/`
- Wire lane into `test-all` so it runs alongside existing `live-mcp`, `live-rbac`, `live-protocol` lanes
- Add `docs/book/src/compliance-testing.md` covering:
  - Stack topology diagram
  - `make conformance-check` run instructions
  - Single-scenario: `make conformance-check SCENARIO=server-stateless`
  - How to update the baseline
  - `conformance-baseline.yml` contract (stale entry = CI exits 1)
  - CI lane name and gating policy
- Register new page in `SUMMARY.md` under Testing section

## Gating policy

| Phase | CI behaviour |
|---|---|
| This issue merged | Lane runs; all failures baselined; exits 0. No merge block. |
| Each scenario removed from baseline | That group joins the enforcement set. |
| Baseline empty | Full 2026-07-28 compliance — Epic #5557 done. |

## Acceptance

- `scripts/cf-integration.sh compliance` runs and exits 0
- `docs/book/src/compliance-testing.md` renders in mdBook without warnings

madhu-mohan-jaishankar:
completed in https://github.com/contextforge-org/contextforge-data-plane/pull/82. separate CI workflow.

Issue #6019, opened 2026-07-31
[TESTING]: Add everything-server + register service to docker-compose.yml


Lang-Akshay:
Part of #6018.

## What

Two new services in `docker-compose.yml` under a `conformance` compose profile.

### everything_server

- Build from `modelcontextprotocol/conformance` → `examples/servers/typescript/`
- Runs `tsx everything-server.ts` on port 3000
- Streamable HTTP endpoint at `/mcp`
- Healthcheck: `curl -f http://localhost:3000/health`

Dockerfile (`infra/everything-server/Dockerfile`):
```dockerfile
FROM node:20-slim
WORKDIR /app
RUN apt-get update && apt-get install -y git curl && rm -rf /var/lib/apt/lists/*
RUN git clone --depth 1 https://github.com/modelcontextprotocol/conformance.git /tmp/conformance \
    && cp -r /tmp/conformance/examples/servers/typescript/* . \
    && rm -rf /tmp/conformance
RUN npm install
EXPOSE 3000
HEALTHCHECK --interval=5s --timeout=3s --retries=10 CMD curl -f http://localhost:3000/health || exit 1
CMD ["npx", "tsx", "everything-server.ts"]
```

### register_everything

Mirrors `register_fast_time` one-shot container pattern exactly (same image, same Python-in-shell approach):

- Depends on: `gateway` (service_healthy), `everything_server` (service_started)
- Waits for gateway health + authenticated readiness
- Generates JWT via `mcpgateway.utils.create_jwt_token`
- `POST /gateways` → `name: conformance_everything`, `url: http://everything_server:3000/mcp`, `transport: STREAMABLEHTTP`
- `POST /gateways/{id}/tools/refresh?include_resources=true&include_prompts=true`
- Waits for tool sync
- `POST /servers` → fixed UUID `c0f0e5a0b1c2d3e4f5a6b7c8d9e0f1a2`, name `Conformance Everything Server`
- Writes bearer token to `/tmp/conformance-token.txt`

## Acceptance

- `docker compose --profile conformance up -d` → everything-server healthy at :3000, virtual server accessible at `/servers/c0f0e5a0b1c2d3e4f5a6b7c8d9e0f1a2/mcp`
- `make testing-up` (no conformance profile) unchanged

derekallanboman:
I'd like to work on this issue. I'll add the everything-server compose service and conformance-check target, then verify the stated acceptance criteria.

[...]

PR #6125, opened 2026-08-07
chore(tests): replace fastmcp with the official mcp SDK in live-gateway tests

### LINKED ISSUE #5910: [CHORE]: Extract tests/live_gateway into an independent Python project (unblocks SDK 2.0 packaging simplification)
## Rationale

The MCP SDK 2.0.0 migration (#5493) needs `mcp==2.0.0b2` for the gateway runtime, while the live-gateway test harness needs `fastmcp` (which pins `mcp<2.0`). As implemented today, the migration works around this with a mutually-exclusive `runtime` / `live-tests` extras partition and a `[tool.uv] conflicts` block. That partition is a tax on every consumer and every Make/CI invocation — and bare `pip install mcp-contextforge-gateway` currently installs no MCP SDK at all.

The better fix is to remove the co-location problem instead of managing it: the harness is already ~99% black-box (exactly **one** `mcpgateway` import in the whole tree) and talks to the gateway over HTTP with locally-minted JWTs. On main, `fastmcp>=3.4.2` (mcp 1.x) and the gateway's `mcp>=1.28.1` coexist with **no conflict**, so extracting now — *before* the SDK migration lands — is pure reorganization with zero resolver pressure.

Once this lands, #5493's packaging change collapses to: replace `mcp>=1.28.1` with `mcp==2.0.0b2` + `mcp-types==2.0.0b2` in core `dependencies`. No conflicts block, no extras, no `--extra` flags, no split-brain docs.

## Choices considered

| Option | Verdict | Why |
|---|---|---|
| **A. In-repo independent project** (own pyproject/lock/venv, NOT a workspace member) | **CHOSEN** | Keeps PR-time atomicity (harness validates the PR's own code); `COMPLIANCE_GAPS.md` stays next to the gateway it documents; root deps shrink immediately |
| A2. uv workspace member | Rejected | Workspaces resolve all members into ONE lockfile — recreates the mcp 2.0-vs-<2.0 conflict one level up |
| B. Separate repo | Rejected | Loses PR-time atomicity; version skew between harness expectations and gateway behavior; exiles the GAP registry from the codebase it describes |
| C. Container-wrapped harness | Later | Good distribution form of A for CI hermeticity; worse dev iteration loop. Follow-up, not now |
| D. Status quo (extras partition, on #5493) | Rejected | Permanently taxes every consumer and every Make/CI invocation; bare pip install breaks |

## Verified current-state facts (main @ 4ed6b0142)

- Exactly one `mcpgateway` import under `tests/live_gateway/`: `fixtures/primary_marker.py` (`is_primary_worker`).
- Shared-helper imports (`tests/helpers/auth.py`, `api_helpers.py` — 166 + 183 LOC) in exactly 7 files.
- `make test` already ignores `tests/live_gateway` (`PYTEST_IGNORE`) — default suite collection unchanged.
- All `fastmcp` usage outside `tests/live_gateway/` is in `mcp-servers/*` sample servers (own pyprojects).
- Wheels already exclude tests (`[tool.setuptools.packages.find]`: `include = ["mcpgateway*"]`); `MANIFEST.in` includes only `tests/async` + `tests/manual`; `.dockerignore` already excludes `tests/` and `**/.venv`.

## Decision

**Superseded (2026-08-13).** #6125 ships a route the choices table didn't include: remove `fastmcp` outright. The dependency conflict this issue targets (`fastmcp` pinning `mcp<2.0`) is deleted rather than partitioned around — the fastmcp-based protocol-compliance harness, the compliance reference server and matrix tooling, and the go/rust parity test are removed, and `test_mcp_protocol_e2e.py` / `test_mcp_rbac_transport.py` are rewritten on the official `mcp` SDK client. Protocol-conformance coverage moves to Anthropic's conformance suite, tracked in epic #6018. The #5493 packaging follow-up is unchanged: with `fastmcp` gone from the dev group, the migration collapses to bumping `mcp` in core `dependencies`.

<details>
<summary>Original decision (Option A — superseded)</summary>

Adopt **Option A**: `tests/live_gateway/` becomes a self-contained Python project (own `pyproject.toml`, `uv.lock`, `.venv`, pytest config), runnable via `uv run --directory tests/live_gateway pytest ...`, with the root project shedding `fastmcp`, `compliance-reference-server`, and the `authlib` dev pin.

</details>

## File-by-file approach

**New files**
1. `tests/live_gateway/pyproject.toml` — independent project (NOT a workspace member). Deps: `fastmcp>=3.4.2`, `authlib>=1.7.2`, `compliance-reference-server` (path `../../mcp-servers/python/compliance_reference_server`, editable), `httpx`, `pytest`, `pytest-asyncio`, `playwright`. Own `[tool.pytest.ini_options]` (asyncio mode + markers). Final dep list audited by importing every module in the tree.
2. `tests/live_gateway/uv.lock` — generated via `uv lock` in that directory.
3. `tests/live_gateway/helpers/auth.py`, `helpers/api_helpers.py` — vendored copies of the two `tests/helpers/` modules.
4. `tests/live_gateway/helpers/primary_worker.py` — vendored `is_primary_worker` (severs the only `mcpgateway` import).

**Import rewrites (7 files)** — `from tests.helpers.*` → vendored helpers:
`e2e_rust/test_mcp_access_matrix.py`, `e2e_rust/test_mcp_session_isolation.py`, `mcp/test_mcp_plugin_parity.py`, `mcp/test_mcp_rbac_transport.py`, `plugins/_helpers.py`, `protocol_compliance/fixtures/gateway_live.py`, `sso/test_oauth_jwks_e2e.py`; plus `fixtures/primary_marker.py` → vendored `is_primary_worker`.

**Root project**
- `pyproject.toml`: remove `compliance-reference-server`, `fastmcp>=3.4.2`, and the dev-only `authlib` pin from the `dev` extra; remove the `[tool.uv.sources]` entry.
- `uv.lock`: regenerate (required by `infra/wheels/Containerfile`'s `--frozen` export).
- `Makefile` (~22 references): live-gateway targets → `uv run --directory tests/live_gateway pytest ...`.
- `tests/live_gateway/plugins/Makefile.plugin-integration`: same `--directory` treatment.
- Workflows (`pytest.yml`, `pytest-rust.yml`, `plugin-integration*.yml`): verify harness invocations route through the updated Make targets; adjust any direct calls. Paths-filters unchanged (directory does not move).
- `MANIFEST.in`: add `prune tests/live_gateway` (explicit sdist exclusion of the now-self-contained folder).
- `.dockerignore`: verify-only (`tests/` and `**/.venv` already excluded) — confirm no harness artifact can enter any build context.
- `tests/AGENTS.md`, `tests/live_gateway/README.md`, root `AGENTS.md`: document the split and the "do not add harness deps to root" rule.

**Non-changes**: the directory does not move; `tests/unit|integration|e2e` stay in the root project; `mcp-servers/*` untouched; no `mcpgateway/` package code changes.

## Verification plan

1. `make test` — unchanged collection/result (already ignores `tests/live_gateway`).
2. `cd tests/live_gateway && uv sync && uv run pytest protocol_compliance -k reference-stdio` — green with no gateway running (proves env independence).
3. `make testing-up && make test-mcp-protocol-e2e test-mcp-rbac test-protocol-compliance` — green via new invocations.
4. `python -m build` — wheel + sdist contain zero `tests/live_gateway` artifacts.
5. `docker build -f Containerfile .` unaffected; `infra/wheels/Containerfile` export succeeds against regenerated root `uv.lock`.
6. `uv lock --check` green in BOTH projects.
7. Grep gates: zero `from mcpgateway` under `tests/live_gateway/`; zero `fastmcp|compliance-reference-server` in root `pyproject.toml`.

## Follow-on (not in this PR)

- Rebase #5493 onto the merged extraction: delete the `conflicts` block, `runtime`/`live-tests` extras, all `--extra` flags (Make/CI/Containerfiles), and the Dependency Layout docs; move `mcp==2.0.0b2`/`mcp-types==2.0.0b2` into core `dependencies`.
- Optional: container-wrapped harness image (Option C) for CI hermeticity.

jonpspri:
## Summary

Removes the dev-only `fastmcp` test dependency from the live-gateway suites and standardizes on the official `mcp` Python SDK client, unblocking the SDK 2.0 packaging simplification tracked in the linked issue.

- **Dependencies**: drop `fastmcp` (and its `authlib` transitive pin) from the dev dependency group; regenerate `uv.lock`.
- **Removed suites**: delete the fastmcp-based protocol-compliance harness (`tests/live_gateway/protocol_compliance/`) along with its `make test-protocol-compliance*` targets, and the fastmcp-based go/rust fast-server parity test.
(_Note:_ To the best of my knowledge, I've been the only user of this suite and
our plan is to move to using Anthropic's conformance suite, tracked in epic #6018.  I'm all for reducing\
our maintenance footprint.  The Go/Rust parity test is no longer necessary: the Go
implementation of fast-time-server has been completely removed from the repo, so
there is no Go server left for the Rust server to be held at parity with.)
- **Rewritten suites**: `test_mcp_protocol_e2e.py` and `test_mcp_rbac_transport.py` now drive `mcp.client.streamable_http.streamablehttp_client` + `ClientSession` directly — bearer auth via plain `Authorization` headers, fast-fail client timeouts preserved, and FastMCP result accessors mapped to the SDK's (`is_error` -> `isError`, `call_tool_mcp` -> `call_tool`, `ping` -> `send_ping`).
- **Docs**: live-gateway READMEs, package docstring, and Makefile help text now describe the `mcp` SDK client.

One implementation note for reviewers: pytest-asyncio drives async-generator fixture setup and teardown in separate tasks, which breaks the anyio task-group context managers inside `streamablehttp_client`/`ClientSession` (`RuntimeError: Attempted to exit cancel scope in a different task`). FastMCP's `Client` never hit this because it runs the session in its own task; the rewritten `client` fixture reproduces that by running the session lifecycle in a dedicated runner task and handing the session to the test.

## Completeness pass

Follow-up commits chase down what the harness left behind, so no orphaned tooling or stale guidance survives the removal:

- **Reference server + matrix tooling**: the protocol-compliance harness was the only consumer of `mcp-servers/python/compliance_reference_server/` and `scripts/compliance_matrix.py`; both are removed.
- **Packaging**: drops the `compliance-reference-server` dev dependency and its `[tool.uv.sources]` editable entry, plus the compliance-only pytest markers (`mcp_client_features`, `mcp_security`, `protocol_compliance`, `target_*`, `transport_*`, `mcp_drift`). Regenerating `uv.lock` also sheds `fastmcp`, `fastmcp-slim`, `authlib` and their transitives.
- **Docs and process**: the PR-review validation gate (`AGENTS.md`, `llms/NOTES.template.md`), `tests/AGENTS.md`, `tests/live_gateway/README.md`, and `docs/docs/development/release-management.md` no longer reference `make test-protocol-compliance*` and now name the official mcp SDK client. `helpers/mcp_test_helpers.py`'s module docstring had a stale consumer list and pointed new tests at the FastMCP `Client` pattern; it now points at the SDK `ClientSession` pattern and lists the suites that actually import it.
- **Makefile**: `pip-audit`'s expected `--skip-editable` list drops to the single gateway install, and the `PYTEST_IGNORE` comment loses the removed targets.

Verification: `uv lock --check` green, `pytest --collect-only` clean across `tests/unit tests/integration tests/e2e` and `tests/live_gateway`, and no residual `fastmcp` / `compliance` references outside the `mcp-servers/*` sample servers (which carry their own pyprojects) and historical ADR/CHANGELOG text.

Supersedes #6023 (same change, squashed to a single commit and moved to a branch on the upstream repo).

Closes #5910

## Review history

- The Entra ID integration-test relocation was split out of this PR into #6217 (unrelated to the fastmcp-removal scope).
- **2026-08-13, second round:**
  - *`Closes #5910` vs. the issue's Decision section*: #5910's Decision section has been updated to record the approach shipped here (remove `fastmcp` outright rather than the originally-selected Option A isolation project); the original decision text is preserved collapsed for history.
  - *Marker/sequencing deadlock with #5935*: **this PR lands before #5935.** After this merge, the `mcp_auth` / `mcp_utilities` / `mcp_tasks` marker registrations and the `make test-protocol-compliance*` doc references that #5935 currently preserves/repoints lose their last consumers, so #5935 will be reworked on top of this one (requested in a comment on that PR).
  - *Interim coverage gap*: the deleted OAuth / security-best-practices / subscriptions / drift suites have no in-repo replacement yet; the committed replacement is Anthropic's conformance suite under epic #6018 (sub-issues #6019–#6022), linked here and in #5910 so the gap is tracked against a plan rather than silently dropped.
  - *Fixture cancellation handling* (`test_mcp_protocol_e2e.py`): `_session_runner` now catches `Exception` instead of `BaseException` so a cancelled runner propagates `CancelledError` rather than stashing it, and runner setup/teardown is wrapped in `try/finally` that releases (or cancels, if setup never completed) the runner task — no more parked session holding an open HTTP connection. The post-yield teardown-error re-raise is kept, now guarded so it never masks an in-flight exception.


cafalchio:
I havent ran playright tests, but other tests and the changes looks good to me.

cafalchio (dismissed):
LGTM

cafalchio (dismissed):
LGTM

msureshkumar88 (changes requested):
Thanks for tackling this — extracting the fastmcp dependency conflict before the SDK 2.0 migration lands is a good sequencing call, and the rewritten fixtures against the official `mcp` SDK client are clean. A couple of things worth resolving before merge, plus a couple of smaller notes.

## Blocking

**1. `client` fixture silently swallows teardown-phase failures** (`tests/live_gateway/mcp/test_mcp_protocol_e2e.py`)

```python
async def _session_runner() -> None:
    try:
        async with streamablehttp_client(...) as (read_stream, write_stream, _):
            async with GatewayClientSession(...) as session:
                await session.initialize()
                holder["session"] = session
                ready.set()
                await release.wait()
    except BaseException as exc:
        holder["error"] = exc
        ready.set()
...
if "error" in holder:
    await runner
    raise holder["error"]
yield holder["session"]
release.set()
await runner   # holder["error"] is never re-checked here
```

The `except BaseException` wraps the whole session lifecycle, including the `__aexit__` teardown that runs after `release.set()`. If teardown raises (dropped connection, protocol violation on close, an anyio cancel-scope issue), it lands in `holder["error"]`, but only the *pre-yield* branch checks that dict — the post-yield path just awaits `runner` and returns. Teardown failures are silently swallowed and the test still reports pass. This is a real regression versus the old FastMCP `async with Client(...) as connected: yield connected`, where an `__aexit__` failure propagated and failed the test. Suggest re-checking `holder.get("error")` after the final `await runner` and raising there too.

**2. Go/Rust parity-test removal justification doesn't hold up**

`test_fast_server_go_rust_parity.py` is deleted with the PR description stating it's *"no longer needed — covered in integration tests against fast-*-server."* I couldn't find any equivalent coverage in `tests/integration/` or `tests/e2e/` — grepped both trees for Go-vs-Rust time-conversion parity or latency-ratio assertions and came up empty. The deleted test asserted two things with no current replacement: (a) Go and Rust `fast_time_server` produce identical time-conversion results, and (b) Rust isn't more than `FAST_SERVER_PARITY_MAX_RATIO` (2.5x) slower than Go. That's a real correctness + performance-regression guard for the Rust rewrite being dropped. Could you either point me to where this is actually covered, restore an equivalent check, or update the PR description to say the coverage is being knowingly dropped pending the Anthropic compliance-suite work rather than "covered elsewhere"?

## Non-blocking

**3. Issue #5910's "Decision" section is stale.** It still documents "Option A" (isolate `tests/live_gateway` as its own project, keep fastmcp there), which a same-day follow-up comment superseded ("we're going to simply remove these tests..."). This PR correctly implements the superseding decision, but future readers of the issue will be misled by the unedited body — worth a quick update/close-out note.

**4. Entra ID test relocation is unrelated to this PR's purpose.** `tests/live_gateway/sso/test_entra_id_integration.py` → `tests/integration/test_entra_id_integration.py` has zero fastmcp references, so it's outside the stated scope of "remove fastmcp." It also folds this test into the default `make test`/CI run for the first time (previously opt-in via `make test-e2e-sso`). Risk is low since it self-skips without `AZURE_*` credentials, but it's a default-suite composition change that isn't called out in the PR description — might be cleaner as a separate PR/commit, or at least flagged explicitly.

No concerns on the Alembic front (no schema touched, correctly no migration), and confirmed `mcp-servers/python/compliance_reference_server` is fully orphaned after deletion with no leftover references anywhere in the tree.

msureshkumar88 (changes requested):
Thanks for the thorough completeness pass here — the follow-up commits chasing down the reference server, matrix tooling, markers, and docs are genuinely well done, and the writeup of the pytest-asyncio/anyio task-boundary problem in the PR body saved me a lot of time understanding the fixture rewrite. A few things worth resolving before this lands.

## Blocking

**1. `Closes #5910` doesn't match what the issue decided**

Issue #5910 has an explicit decision record: *"Adopt Option A: `tests/live_gateway/` becomes a self-contained Python project (own `pyproject.toml`, `uv.lock`, `.venv`, pytest config), runnable via `uv run --directory tests/live_gateway pytest ...`"*, with a choices-considered table covering A/A2/B/C/D. This PR takes a different route — removing `fastmcp` outright rather than partitioning around it. On the branch there's no `tests/live_gateway/pyproject.toml`, no `tests/live_gateway/uv.lock`, no vendored helpers, no `MANIFEST.in` prune, and no `--directory` invocations.

To be clear, I think the approach here is arguably the better one — #5910's entire rationale was that `fastmcp` pins `mcp<2.0`, and deleting the dependency removes the conflict more cleanly than isolating it. But the issue's Decision section still says Option A, and its verification plan (step 3) requires `make test-protocol-compliance`, which this PR removes. Could we either update #5910's Decision section to reflect the approach actually taken, or close it and track this under a new issue? As it stands, anyone reading #5910 to understand what shipped will get the wrong picture.

**2. Marker deadlock with #5935 — these two PRs will collide**

PR #5935 (`Closes #5835`, currently MERGEABLE) removes `tests/compliance/` and states in its body: *"`mcp_auth` stays — `tests/live_gateway/protocol_compliance/test_oauth_authorization.py` uses it."* That's the exact file this PR deletes.

Checking the marker consumers on this branch after the deletions here:

```
mcp_auth      -> only tests/compliance/mcp_2025_11_25/authorization/...
mcp_utilities -> only tests/compliance/mcp_2025_11_25/tasks/...
mcp_tasks     -> only tests/compliance/mcp_2025_11_25/utilities/...
```

All three now live solely in the tree #5935 deletes. So each PR is keeping a marker registered on the assumption that the *other* PR's deleted tree still uses it. If both merge we end up with three orphaned marker registrations in `pyproject.toml` and — more importantly — no MCP protocol compliance coverage anywhere in the repo. #5935 also repoints docs (`explorer.html`, `RELEASE-CHECKLIST.md`) at `make test-protocol-compliance` and at `protocol_compliance/README.md`, both of which this PR removes.

Mechanically the two PRs also touch the same `markers = [...]` block in `pyproject.toml` and the same Makefile compliance targets, so a merge conflict is coming regardless. Worth syncing with @/the #5935 author on sequencing and on who ends up owning the markers.

**3. Deleted security coverage has no replacement, and the replacement plan isn't linked**

The removed suites include the only live black-box assertions we have on a few security behaviors:

- `test_oauth_authorization.py` (169 lines) — PRM endpoint wiring, `WWW-Authenticate` on 401, real Keycloak `client_credentials` issuance, tampered-token rejection
- `test_security_best_practices.py` (151) — session-ID entropy, host-header spoofing resilience, clickjacking-defense headers, malformed bearer/content-type handling
- `test_subscriptions.py` (96) — actual `notifications/resources/updated` delivery to a subscriber
- `test_drift.py` (165) — reference vs. gateway-proxy vs. gateway-virtual behavioral equivalence

Nothing in `tests/unit`, `tests/integration`, `tests/e2e`, or the surviving `tests/compliance/mcp_2025_11_25` asserts these — the latter only checks JSON-RPC envelope shape and 404s. That leaves a window where a regression that weakens session-ID entropy, drops the `WWW-Authenticate` header, removes clickjacking-defense headers, or accepts a forged bearer token on the live `/mcp` endpoint would ship with nothing to catch it.

The PR body says *"our plan is to move to using Anthropic's compliance suites"*, and epic #6018 does exist and is scoped (#6019–#6022) — but it's unlinked from this PR and all four sub-issues are still open/unstarted. Two options that would resolve this: link #6018 explicitly here and in #5910 so the interim gap is tracked against a committed plan, or retain the OAuth/security-best-practices files (they're the two highest-value ones) until #6018 lands.

**4. Fixture cancellation handling in `test_mcp_protocol_e2e.py`**

`_session_runner` uses `except BaseException as exc:` (line 125), which catches `asyncio.CancelledError` and stashes it in `holder["error"]` rather than letting it propagate. If the runner is ever cancelled while parked at `await release.wait()` — a future pytest-timeout integration, or an enclosing anyio cancel scope — the cancellation is silently converted into normal task completion, making `task.cancel()` unreliable here. `except Exception` would preserve the intent (surfacing connection/init failures) without swallowing cancellation.

Relatedly, lines 129-134 have no `try/finally` around `runner = asyncio.create_task(...)` / `await ready.wait()`. If the fixture's setup phase is itself cancelled (Ctrl-C, timeout, `GeneratorExit` before reaching `yield`), `runner` is never cancelled or awaited and `_session_runner` stays parked forever holding an open HTTP connection.

## Suggestions

- **`test_mcp_protocol_e2e.py:305, 313, 319`** — `result.content[0]` is indexed unguarded in `test_echo`, `test_convert_time`, and `test_get_stats`, while `test_get_system_time` (line 294) uses `assert result.content and result.content[0]...` and `test_schema_error_preserves_payload` uses `... if result.content else ""`. A live upstream returning `isError=False` with an empty `content: []` (legal per the spec) would raise an opaque `IndexError` instead of a readable assertion failure. Worth making the three consistent with the guarded pattern.

- **`test_mcp_protocol_e2e.py:137-141`** — the post-`yield` re-check raises unconditionally on teardown, independent of whether the test body already failed. A test that fails its own assertion *and* hits a teardown-phase transport error now reports both a call-phase FAILED and a teardown-phase ERROR for the same test id, which buries the more diagnostic original message.

- **Session-setup duplication** — `test_mcp_protocol_e2e.py:104-141` and `test_mcp_rbac_transport.py:432-440` now carry two structurally different implementations of MCP session establishment, and they've already diverged (one subclasses `ClientSession` to capture `InitializeResult`, the other doesn't; one needs the background-runner workaround, the other doesn't because it's driven through `_run_async`). `mcp_test_helpers.py`'s own docstring positions itself as the shared home for exactly this pattern. Consolidating now would save the next suite from copy-pasting the workaround.

- **`mcp_test_helpers.py:10-12`** — the rewritten docstring lists the plugin E2E suites and `test_mcp_plugin_parity.py` as consumers, but `sso/test_oauth_jwks_e2e.py`, `e2e_rust/test_mcp_session_isolation.py`, `e2e_rust/test_mcp_access_matrix.py`, and `mcp/test_langfuse_traces.py` (10 imports) also depend on this module. Since the docstring is what a future contributor will use to gauge blast radius before touching these helpers, worth getting the list complete.

## Minor

- The PR currently shows as `CONFLICTING` against `main` and will need a rebase.

Nothing here is about the direction — dropping `fastmcp` is the right call and the execution is clean. The sequencing with #5935 and the interim security-coverage gap are the two I'd most want settled before merge. Happy to re-review once those are sorted.

jonpspri:
@msureshkumar88 Thanks for the thorough second pass — all four blocking items are now addressed:

1. **`Closes #5910` vs. the issue's Decision section** — took the first option: #5910's Decision section has been edited to record the approach actually shipped here (remove `fastmcp` outright rather than the Option A isolation project), with the original decision text preserved collapsed for history.
2. **Marker/sequencing deadlock with #5935** — this PR will land first. A comment on #5935 asks the author to rework on top of the merge: drop the then-orphaned `mcp_auth` / `mcp_utilities` / `mcp_tasks` registrations and re-point the docs that currently target `make test-protocol-compliance*`. The sequencing is also recorded in this PR's body under Review history.
3. **Deleted security coverage / replacement plan** — epic #6018 (Anthropic conformance suite, sub-issues #6019–#6022) is now linked in this PR's body and in #5910, so the interim gap in OAuth / security-best-practices / subscriptions / drift coverage is tracked against a committed plan rather than silently dropped.
4. **Fixture cancellation handling** — fixed in acbf630ab: `_session_runner` now catches `Exception` instead of `BaseException` so a cancelled runner propagates `CancelledError` rather than stashing it in `holder`, and runner setup/teardown is wrapped in `try/finally` that releases the runner — or cancels it when setup never completed — and awaits it, so it can't stay parked holding an open HTTP connection. The post-yield teardown re-raise is kept, now guarded so it never masks an in-flight exception. Verified against the live stack: `make test-mcp-protocol-e2e` → 19 passed, 3 environment skips.

The rebase onto main is pending — I'll handle merge mechanics once the findings are settled. Happy to address anything further.

msureshkumar88 (dismissed):
Thanks for chasing all of this down — both rounds of feedback are addressed:

- Teardown-failure swallowing in the `client` fixture: fixed (re-checks `holder["error"]` after the final `await runner`).
- `Closes #5910` vs. the issue's Decision section: resolved — the Decision section now documents the approach actually shipped, with the original Option A preserved collapsed for history.
- Marker sequencing with #5935: acknowledged and coordinated — comment posted there confirming #6125 lands first and #5935 reworks on top.
- Deleted security-suite coverage: linked against epic #6018 (#6019–#6022) in both this PR and #5910, so the interim gap is tracked rather than silent.
- Fixture cancellation handling (`except BaseException` → `except Exception`, missing try/finally around runner setup): fixed and verified — traced both the pre-yield-cancel and post-yield-cancel paths, no leaked connection, no swallowed `CancelledError`.

A few non-blocking suggestions from round 2 are still open (unguarded `result.content[0]` in a couple of tool-call tests, the `mcp_test_helpers.py` docstring's consumer list, and the session-setup duplication between the two rewritten suites) — none of these are correctness or security concerns, just polish. Fine to land as-is or pick up in a follow-up, your call.

One thing to note before merge: the branch is currently showing as conflicting against `main` (a few commits behind) and will need a rebase.

Approving — good execution on a tricky pytest-asyncio/anyio task-boundary problem, and the write-up in the PR body made this easy to review.

msureshkumar88 (approved):

PR #6126, opened 2026-08-07
fix(admin-ui): read chartRegistry from Admin, not window (#6055)

### LINKED ISSUE #6054: Observability tab's Alpine component never initializes — modal backdrops trap the whole admin UI
### 🐞 Bug Summary

Opening the Observability tab (`/admin/#observability`) leaves its Alpine component with zero reactive keys, which has two visible effects: two full-viewport modal backdrops render unhidden and swallow every click, and the traces table never loads.

The bundle imports the CSP-safe Alpine build (`mcpgateway/admin_ui/alpine-setup.js:1`, `import Alpine from '@alpinejs/csp'`; `package.json:60`), which cannot evaluate arbitrary inline expressions. But `mcpgateway/templates/observability_partial.html:49-468` declares an inline `x-data="{ … }"` object literal containing ~53 method shorthands, and `:469-483` declares `x-init` with statements. Neither parses under the CSP build.

`mcpgateway/templates/observability_trace_detail.html:51` uses the same inline-`x-data`-with-methods pattern and looks affected identically. Of 35 Alpine roots in the live admin DOM, only these declare methods inline — simple inline literals elsewhere (`metrics_partial.html:3`, `admin.html:1481`, `:1542`) initialize fine on Alpine 3.15.12 CSP, so the fix looks scoped to converting this one component to a registered `Alpine.data()`.

---

### 🧩 Affected Component

---

### 🔁 Steps to Reproduce

1. Start the gateway with `MCPGATEWAY_UI_ENABLED=true` and log in to `/admin`.
2. Open browser devtools console.
3. Click the **Observability** tab.
4. Observe the console errors below, and that two full-viewport overlay `<div>`s (`fixed inset-0 … z-[1000]`) appear even though nothing was clicked to open a modal.
5. Try clicking anywhere on the page — the overlays intercept the click.
6. Observe the traces table stays on "Loading traces..." indefinitely, although `GET /admin/observability/traces` returns 200.

---

### 🤔 Expected Behavior

The Observability panel's Alpine component initializes normally: both modals (`x-show="selectedTrace"`, `x-show="showSaveQueryModal"`) stay hidden until explicitly opened, and the traces table loads via `x-init="startPolling()"`.

---

### 📓 Logs / Error Output

```
CSP Parser Error: Expected PUNCTUATION ":" but got IDENTIFIER "loadMetricsView"
CSP Parser Error: Unexpected token: loadSavedQueries
Undefined variable: viewMode
Undefined variable: selectedQueryId
Undefined variable: savedQueries
Undefined variable: timeRange
Undefined variable: statusFilter
Undefined variable: minDuration
Undefined variable: maxDuration
Undefined variable: httpMethod
Undefined variable: userEmail
Undefined variable: nameSearch
Undefined variable: attributeSearch
Undefined variable: showAdvancedFilters
Undefined variable: toolName
Undefined variable: selectedTrace
Undefined variable: showSaveQueryModal
```

(95 `Undefined variable` errors fire on every `/admin` page load, not only when the Observability tab is opened, because the partial is eagerly injected via `hx-trigger="revealed once"` on a hidden panel — `admin.html:1457-1459`.)

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Platform / OS | Linux |

---

Related to a second, independent issue where the tab also cannot be *left* once visited (separate report, since it has a different root cause and would remain even after this one is fixed) — see the Observability tab-switching bug report.

Found via a manual-simulation E2E pass (real Chromium via Playwright) driving every admin tab.

### LINKED ISSUE #6055: [BUG]: Cannot leave the Observability tab — window.chartRegistry is undefined
### 🐞 Bug Summary

Once the Observability tab has been visited, clicking any other sidebar tab throws and the navigation silently fails — the user is stuck on Observability until a full page reload.

`mcpgateway/admin_ui/tabs.js:357-360` calls `window.chartRegistry.destroyByPrefix(...)` when leaving the tab, but the registry is only ever defined as `Admin.chartRegistry` (`mcpgateway/admin_ui/app.js:16`, `events.js:944`) — `typeof window.chartRegistry` is `undefined`. The resulting `TypeError` is swallowed at `tabs.js:829-831` *before* the target panel is un-hidden, so the tab switch fails with no visible error to the user beyond the console.

This is a separate defect from the Observability modal/backdrop issue (different report) — it blocks *programmatic* navigation specifically, which is why even setting `location.hash` directly updates the URL without revealing the target panel.

---

### 🧩 Affected Component

---

### 🔁 Steps to Reproduce

1. Log in to `/admin`.
2. Click the **Observability** tab.
3. Click any other sidebar tab (e.g. Tools, Metrics, Plugins).
4. Observe the target tab's panel never appears — the Observability panel stays visible.
5. Open devtools console and see the error below.
6. Confirm this reproduces with `location.hash = '#plugins'` as well (hash updates, panel does not switch).

---

### 🤔 Expected Behavior

Leaving the Observability tab cleans up its chart instances via `Admin.chartRegistry` and the requested tab's panel is revealed normally.

---

### 📓 Logs / Error Output

```
TypeError: Cannot read properties of undefined (reading 'destroyByPrefix')
Failed to switch to <tab> tab
```

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Platform / OS | Linux |

---

Found via a manual-simulation E2E pass driving every admin tab with a real browser (Playwright). Independently reproduced twice, including via a breadth sweep across all 22 admin tabs, where Observability's neighbor tab was the only one of 22 whose sidebar click timed out — and only immediately after Observability had been visited.

Garry2012:
Closes #6055.

Thanks to @msureshkumar88 for the report — the diagnosis there was accurate down to the file and line, so this is essentially implementing what the issue already specified.

## The bug

`tabs.js` reads `window.chartRegistry`, but the registry is defined as `Admin.chartRegistry` in `app.js`, inside an IIFE closed over `window.Admin`. `events.js:944` already uses the correct path. So `window.chartRegistry` is always `undefined`, and calling a method on it throws:

```
TypeError: Cannot read properties of undefined (reading 'destroyByPrefix')
Error: Failed to switch to tools tab
```

The throw propagates out of the tab-switch handler, so the switch aborts and the previous panel stays visible — the user is effectively stuck on the Observability tab.

Only reachable with `OBSERVABILITY_ENABLED=true`, since the block is gated on the observability panel being the visible one. That is presumably why it has gone unnoticed while the setting defaults to `false`.

## The fix

Use `window.Admin.chartRegistry`, with a presence check so a missing registry degrades to skipping chart cleanup rather than breaking navigation.

## Verification

Headless browser against this branch, switching away from Observability to Tools and Gateways:

| | before | after |
|---|---|---|
| panel visible after switch | `false` | `true` |
| console errors | 8 | 0 |

Full unit suite on the patched tree: 0 failures.

Companion PR: #6127 (Closes #6054, the panel rendering blank under the CSP-safe Alpine build). The two are independent and can be reviewed in either order.

### DIFF
--- a/mcpgateway/admin_ui/tabs.js
+++ b/mcpgateway/admin_ui/tabs.js
@@ -353,11 +353,20 @@ export const showTab = function (tabName) {
       tabName !== "observability"
     ) {
       console.log("Leaving observability tab, triggering cleanup...");
-      // Destroy all observability charts
-      window.chartRegistry.destroyByPrefix("metrics-");
-      window.chartRegistry.destroyByPrefix("tools-");
-      window.chartRegistry.destroyByPrefix("prompts-");
-      window.chartRegistry.destroyByPrefix("resources-");
+      // The registry is defined as Admin.chartRegistry in app.js, not on window
+      // directly (see events.js, which uses Admin.chartRegistry.destroyAll()).
+      // Reading window.chartRegistry threw a TypeError that propagated out of
+      // switchTab, so leaving the observability tab failed with "Failed to
+      // switch to <name> tab" and the panel never changed. It only surfaced
+      // once observability was enabled, since this block is gated on the
+      // observability panel being the visible one.
+      const chartRegistry = window.Admin && window.Admin.chartRegistry;
+      if (chartRegistry) {
+        chartRegistry.destroyByPrefix("metrics-");
+        chartRegistry.destroyByPrefix("tools-");
+        chartRegistry.destroyByPrefix("prompts-");
+        chartRegistry.destroyByPrefix("resources-");
+      }
       // Dispatch event so Alpine components can stop intervals and reset state
       document.dispatchEvent(new CustomEvent("observability:leave"));
     }


Garry2012:
Gentle nudge on this one — it's been open since 7 Aug without a reviewer assigned.

It's a small, self-contained fix: `tabs.js` reads `window.chartRegistry`, but the registry is defined as `Admin.chartRegistry` in `app.js` (which is what `events.js` already uses). The mismatch throws a `TypeError` out of `switchTab`, so leaving the observability tab fails with "Failed to switch to <name> tab" and the panel never changes. It only surfaces when observability is enabled, which is likely why it went unnoticed.

Happy to rebase or split it differently if that helps review. No urgency from my side — just flagging it so it doesn't get lost.

PR #6127, opened 2026-08-07
fix(admin-ui): register observability dashboard via Alpine.data (#6054)

### LINKED ISSUE #6054: Observability tab's Alpine component never initializes — modal backdrops trap the whole admin UI

### LINKED ISSUE #6055: [BUG]: Cannot leave the Observability tab — window.chartRegistry is undefined

Garry2012:
Closes #6054.

Thanks to @msureshkumar88 — the issue identified the cause precisely, including the file and line range, so this implements what was already diagnosed there.

## The bug

The panel rendered blank and made **no network request at all**, which reads as a backend problem. The traces API was returning data correctly the whole time.

`@alpinejs/csp` accepts object literals of **plain data only**. One function — method shorthand or an arrow property — makes the *entire* `x-data` object fail to parse, and Alpine then initialises the component as `{}` **without raising**. `observability_partial.html` declared a ~420 line inline `x-data` holding 18 methods, so every field evaluated as `Undefined variable`, `x-init` never ran, and `startPolling()` was never called.

Established in-browser, isolated from observability entirely:

| `x-data` | CSP build |
|---|---|
| `{ a: 1, b: 'x' }` | parses |
| `{ a: 1, cfg: { x: 2 } }` | parses |
| `{ a: 1, greet() {…} }` | **empty, no error** |
| `{ a: 1, greet: () => 'hi' }` | **empty, no error** |

Registered components are unaffected — the functions never reach the attribute parser — which is why the existing `Alpine.data()` components work.

## The fix

Move the object into `admin_ui/components/observability-dashboard.js`, registered via `Alpine.data()`. The former `x-init` body becomes the component's `init()` hook.

**Behaviour is unchanged.** The extracted object is byte-identical to the original after normalising whitespace and the `root_path` interpolation, and the method inventory matches exactly (18 methods, plus `init()`).

## Verification

Headless browser against this branch:

```
page errors            47 -> 0
rendered trace rows     1 -> 6
/traces request      never -> fires
```

Also exercised: trace detail modal, its Gantt / Flame Graph / Simple View toggles, the close control, and the saved-query load, apply and delete paths (which the empty-list default meant were never otherwise executed). All clean.

Full unit suite on the patched tree: **0 failures**.

## Scope — what this does not fix

The four sub-views (Advanced Metrics, MCP Tools, Prompts, Resources) **remain blank** after this change. They fail for a related but distinct reason: they use `x-data="createMetricsController()"` with the factory assigned to `window`, and the CSP build resolves `x-data` against the `Alpine.data()` registry rather than evaluating globals. Note the call form itself is fine — `overflowMenu('table')` works — but only for a registered name.

I kept that out of this PR to keep the change reviewable. Happy to open a follow-up issue and PR for it, along with the ~20 optional-chaining and template-literal directives elsewhere that the CSP parser also rejects (including in the Roots overflow menu, where they silently disable View/Edit/Export). Let me know if you would prefer them bundled instead.

Companion PR: #6126 (`window.chartRegistry`, Closes #6055). Independent of this one.


Garry2012:
Gentle nudge on this one too — open since 7 Aug with no reviewer yet.

Context that might help review: the admin UI bundles the CSP-safe Alpine build (`@alpinejs/csp`), whose expression parser rejects any function inside an inline `x-data`, plus optional chaining and template literals in directives. Critically it **fails silently** — the component initialises as `{}` with no exception, so the symptom is a blank panel making no network request, which reads like a backend problem.

This moves the large inline `x-data` into a registered `Alpine.data()` component and rewrites the unsupported directive forms. Guard tests are included so the pre-CSP style can't creep back in.

Happy to adjust the approach if you'd prefer it structured differently.

mcpgateway/admin_ui/components/observability-dashboard.js
```diff
@@ -0,0 +1,454 @@
...
+      saveQueryIsShared: false,
+      metricsLoaded: false,
+      toolsLoaded: false,
+      promptsLoaded: false,
+      resourcesLoaded: false,
+      metricsLoading: false,
+      toolsLoading: false,
+      promptsLoading: false,
+      resourcesLoading: false,
+      async loadMetricsView() {
```
gcgoncalves, line 46: All four loadView functions are virtually equal. Could be abstracted into a `loadView` function that takes the entity name (`metrics`, `tools`, `prompts` and `resources`) and the endpoint url (e.g.: `/admin/observability/tools/partial`) and builds the views/logs accordingly.

mcpgateway/admin_ui/components/observability-dashboard.js
```diff
@@ -0,0 +1,454 @@
...
+              console.log('Metrics view already loading, skipping');
+              return;
+          }
+
+          this.metricsLoading = true;
+          console.log('Loading metrics view for the first time...');
+
+          try {
+              const response = await fetch((window.ROOT_PATH || '') + '/admin/observability/metrics/partial');
+              if (response.ok) {
```
gcgoncalves, line 64: If the response is not `ok`, the whole block gets skipped and the user is not notified. Might be worth adding an error state on the `else` block (possibly a retry button?).

gcgoncalves (approved):
Hi, @Garry2012, thanks for this work and sorry for the delayed review. The fix is clean, and moving the dashboard to it's own component is the correct path.

I raised a couple issues as non-blocking, optional improvements (as the issues were already there prior to your fix), just in case you might want to add them and improve the application. Otherwise, they may be addressed later as a follow-up.

PR #6128, opened 2026-08-07
fix(gateway): report failure when refreshing an unauthorized auth_code gateway

### LINKED ISSUE #5247: [BUG]: MCP tools refresh endpoint always returns success even if the mcp server is misconfigured
### 🐞 Bug Summary
MCP tools refresh endpoint always returns success even if the mcp server is misconfigured

---

### 🧩 Affected Component

- [x] `mcpgateway` - API
- [ ] Other (explain below)

---

### 🔁 Steps to Reproduce

1. Create an mcp server without authorizing it or with wrong credentials.
2. Click fetch tools in the admin UI or use `/gateways/{gateway_id}/tools/refresh` on the mcp server you just created.
3. The request succeed with the following response:

```
{"gatewayId":"617930fdd2494fdfb53c2280d95a73d5","success":true,"error":null,"toolsAdded":0,"toolsUpdated":0,"toolsRemoved":0,"resourcesAdded":0,"resourcesUpdated":0,"resourcesRemoved":0,"promptsAdded":0,"promptsUpdated":0,"promptsRemoved":0,"validationErrors":[],"durationMs":2.024332992732525,"refreshedAt":"2026-06-16T11:55:55.665498Z"}
```
---

### 🤔 Expected Behavior
It should fail with an error unauthorized or whatever connection error the server made.
---

### 📓 Logs / Error Output
N/A
---

### 🧠 Environment Info
| Key | Value |
|-----|-------|
| Version or commit | `1.0.3` |
| Runtime | `Python 3.13.5` |
| Platform / OS | `macOS` |
| Container | `none` |

msureshkumar88:
## Summary

Fixes #5247 — `POST /gateways/{id}/tools/refresh` reported `success: true` even when the target MCP server was misconfigured or unauthorized, for gateways using the OAuth `authorization_code` grant.

## Root cause

`_refresh_gateway_tools_resources_prompts` called `_initialize_gateway` without `oauth_auto_fetch_tool_flag=True` and without any pre-authenticated headers. For the `authorization_code` grant, `_initialize_gateway` has an early return (`return {}, [], [], [], []`) specifically because it can't obtain a token without user consent — no network I/O happens. A downstream guard then skipped error reporting for that empty-result case, leaving `success` at its initialized `True`.

Two consequences, the second wider than the bug report:
- An unauthorized gateway falsely reported success (the reported symptom — the ~2ms duration matches the early return).
- **Every** `authorization_code` gateway was affected on manual refresh, authorized or not — even a gateway with a valid stored token never actually connected, since the code path never reached the network call. "Fetch Tools" was a permanent no-op for this grant type.

## Fix

`refresh_gateway_manually` (`created_via="manual_refresh"` only) now resolves the caller's stored OAuth token into a real `Authorization` header via a new helper, `_resolve_auth_code_refresh_headers`, and passes it as `pre_auth_headers`. A non-empty `pre_auth_headers` takes an earlier branch inside `_initialize_gateway`, so the `authorization_code` early return is never reached — the refresh genuinely attempts to connect.

| Case | Before | After |
|---|---|---|
| No token stored | `success: true`, 0 tools, ~0ms | `success: false`, error names `/oauth/authorize/{id}` |
| Valid token stored | `success: true`, 0 tools (never connected) | `success: true`, real tools fetched |
| Upstream rejects the token | N/A (never reached) | `success: false` with the real error |
| Caller supplies `X-Upstream-Authorization` | Already worked (bypassed the early return) | Still works — takes precedence over OAuth resolution (see below) |

**Deliberately unchanged:**
- Health-check refresh (`created_via="health_check"`) and notification-service refresh — no authenticated caller to resolve a token for, and this avoids flapping gateways unreachable or spamming a real upstream with 401s every poll interval.
- The stale-cleanup guard that skips destructive deletion of tools/resources/prompts when an `authorization_code` gateway returns nothing — an authorized gateway that legitimately returns zero tools still reports `success: true` without wiping the catalog.
- HTTP status stays 200 with `success`/`error` in the body — matches the existing `GatewayRefreshResponse` contract, so no client changes are needed. The admin UI (`admin_ui/gateways.js`) already treats `success === false` as a failure.

**Found during review and fixed in the same PR:** manual refresh unconditionally resolving an OAuth token would have regressed the pre-existing `X-Upstream-Authorization` passthrough escape hatch (a caller can supply the upstream token directly, independent of `gateway.auth_type`) — that header previously bypassed the early return on its own, but the new branch would have required a stored token even when the caller already supplied a valid one. Fixed by skipping OAuth resolution whenever `pre_auth_headers` already carries an explicit `Authorization` header.

## Why this design

- **`created_via == "manual_refresh"` gate, not a broader one.** Health checks run unattended on a timer with no per-request caller — there's nothing to resolve a per-user token from, and probing an unauthorized gateway every interval would be noisy and could mark gateways falsely unreachable.
- **A separate helper (`_resolve_auth_code_refresh_headers`) rather than inlining.** Keeps refresh and the existing `fetch_tools_after_oauth` from drifting on token-claim validation (audience/issuer/scope checks before forwarding a bearer token) — a security-relevant check that's easy to duplicate incorrectly.
- **Token validation before forwarding.** The helper calls the existing `validate_oauth_token_claims` and aborts with a `blocking_errors`-only message (never the token itself) if claims are mismatched, rather than forwarding a token to the wrong audience.
- **No schema or migration changes.** `GatewayRefreshResponse` already carries `success`/`error`; no new columns.

## Testing

- **29 unit tests** in `tests/unit/mcpgateway/services/test_gateway_auto_refresh.py` covering: unauthorized/no-token, no caller identity, authorized success, upstream rejection, blocking claim validation, empty-response-preserves-catalog, health-check/notification-service exclusion, non-OAuth unaffected, passthrough-`Authorization`-wins-over-OAuth, and token-never-leaks-into-error-message.
- **End-to-end verification** against a real, isolated gateway instance — real HTTP API, real SQLite DB, a real `FastMCP` server over SSE as the "authorized upstream." No `gateway_service` internals were mocked. 5/5 scenarios passed, including that the upstream genuinely rejects a wrong bearer token (not a stubbed check) and that the health-check path is provably unaffected — see [this comment](https://github.com/IBM/mcp-context-forge/pull/6128#issuecomment-5241705365) for the config, commands, and full output. The one-off verification script has been removed from the tree: it wasn't wired into CI/Makefile, so it would never run again, and its scenarios are already covered by the 29 unit tests above.
- Full suite (`make coverage`): 21736 passed, 807 skipped, 2 xfailed, 0 failures. `diff-cover` reports 100% coverage of the changed lines vs `main`.
- `make ruff interrogate pylint`: clean. `make detect-secrets-scan`: 0 unaudited, 0 live findings.

## Commits

1. `feat(gateway): add OAuth token resolution helper for auth_code refresh`
2. `fix(gateway): report failure when refreshing an unauthorized auth_code gateway`
3. `test(e2e): add end-to-end verification for issue #5247`
4. `test(e2e): move issue #5247 manual scripts under tests/e2e/issue_5247/manual/`
5. `fix(gateway): let explicit passthrough Authorization skip OAuth resolution`
6. `chore(security): audit new detect-secrets findings as false positives`
7. `test(e2e): remove one-off issue #5247 manual verification suite (never wired into CI, superseded by unit coverage + PR comment)`

Closes #5247


msureshkumar88:
## E2E re-verification (fresh run against a live instance)

Re-ran the `tests/e2e/issue_5247/manual/` suite from a clean checkout of this branch against a real, isolated gateway instance — real HTTP API, real SQLite DB, real `FastMCP` server over SSE as the authorized upstream. No `gateway_service` internals mocked.

### Config

`tests/e2e/issue_5247/manual/e2e.env` (generated from `e2e.env.example`, isolated instance on port 48444):

```
JWT_SECRET_KEY=e2e-jwt-secret-DO-NOT-USE-IN-PRODUCTION-32plus-chars
AUTH_ENCRYPTION_SECRET=e2e-enc-secret-DO-NOT-USE-IN-PRODUCTION-32plus
DATABASE_URL=sqlite:///${SCRATCH}/e2e.db
HOST=127.0.0.1
PORT=48444
BASIC_AUTH_USER=admin
BASIC_AUTH_PASSWORD=e2e-admin-password-32chars-min!!
AUTH_REQUIRED=true
MCPGATEWAY_UI_ENABLED=false
PLUGINS_ENABLED=false
LOG_LEVEL=WARNING
MCPGATEWAY_A2A_ENABLED=false
HEALTH_CHECK_INTERVAL=999999
```

### Commands

```bash
bash tests/e2e/issue_5247/manual/start_gateway.sh &   # fresh SQLite DB, port 48444
curl -sf http://127.0.0.1:48444/health                # wait for readiness
./.venv/bin/python tests/e2e/issue_5247/manual/run_e2e.py
pkill -f "uvicorn mcpgateway.main:app"
```

### Output

```
[PASS] 1. Unauthorized auth_code gateway reports failure (the bug): HTTP 200, success=False,
       error='No OAuth tokens found for user admin@example.com on gateway e2e-unauthorized-b76daec4.
       Please complete the OAuth authorization flow first at /oauth/authorize/1902d2fc7b67453c945ae66303ea7ae5',
       toolsAdded=0
[PASS] 2. Authorized auth_code gateway connects and fetches tools (the fix): HTTP 200,
       success=True, toolsAdded=1, error=None
[PASS] 2c. Upstream genuinely enforces the bearer token (not a stub): HTTP 200, success=False,
       error="Failed to initialize gateway ... Client error '401 Unauthorized' ..."
[PASS] 3a. Non-OAuth gateway with wrong credentials reports failure (no regression):
       registration correctly rejected (401 Unauthorized)
[PASS] 3b. Health-check path unchanged (still success=true, no connection attempt):
       stdout='HEALTHCHECK_RESULT True None'

OVERALL: ALL PASSED (5/5)
```

`echo $?` after `run_e2e.py` → `0`.

### Surrounding checks (nothing broken)

- `./.venv/bin/python -m pytest tests/unit/mcpgateway/services/test_gateway_auto_refresh.py tests/unit/mcpgateway/services/test_gateway_service.py -q` → all pass (1 pre-existing unrelated skip: "Yet to implement").
- Live smoke against the same running instance: `GET /gateways/` → `200`, `GET /tools/` → `200` (listing endpoints unaffected).
- Confirmed `admin_ui/gateways.js:1830` already treats `data.success === false` as a failure — no UI change needed, matches the PR description's claim.
- Verified `.gitignore`'s existing `*.db`/`*.db-shm`/`*.db-wal` globs cover the e2e SQLite artifacts generated by the run; `git status` clean after the run.

Confirms the fix works end-to-end against a live gateway and doesn't regress the health-check path, non-OAuth gateways, or the passthrough-Authorization escape hatch.

gandhipratik203:
**Helper duplicates `fetch_tools_after_oauth` rather than being shared with it**

Nice extraction, though `fetch_tools_after_oauth` (`gateway_service.py:2224-2273`) still carries its own copy of the same sequence, which leaves the drift risk the helper was meant to remove.

Would it be worth having it delegate? That needs an optional `db` param on the helper. Happy either way if you'd rather just note why not.

msureshkumar88:
Good call — done. `fetch_tools_after_oauth` now delegates to `_resolve_auth_code_refresh_headers`, which gained an optional `db` param so `fetch_tools_after_oauth` can reuse its own request-scoped session instead of opening a fresh one.

Since the helper now returns advisory claim-validation warnings too (`fetch_tools_after_oauth` used them for the 401/403 diagnostic message), I switched the return type from a plain `{"Authorization": ...}` dict to a small `AuthCodeRefreshHeaders(headers, warnings)` dataclass so that diagnostic doesn't get lost in the merge.

Existing unit coverage for both call sites still passes, plus `make ruff pylint`.

gandhipratik203 (dismissed):
LGTM!

Lang-Akshay (dismissed):
LGTM

Lang-Akshay (approved):
LGTM

PR #6129, opened 2026-08-07
fix(tools): enforce duplicate check on custom_name rename via edit

### LINKED ISSUE #6057: [BUG]: Tool rename via edit modal bypasses the duplicate-name check
### 🐞 Bug Summary

Renaming a tool's Technical Name to an existing tool's name through the edit modal succeeds silently. The database then holds two tool rows presenting as the same name. Create correctly rejects duplicates with 409 — edit does not.

`name_is_changing` is derived by comparing `tool_update.name` to the current `tool.name`, and the uniqueness check (`_check_tool_name_conflict`) runs only inside that branch. But the edit form posts `name` as a hidden field holding the *current, unchanged* value — the only field the user can actually edit is `custom_name` — so `name_is_changing` is always `False` for a UI-driven rename, and `custom_name` is then assigned unconditionally with no conflict check at all.

`mcpgateway/services/tool_service.py:7003` (`name_is_changing` derivation), `:7015` (conflict check gated on it), `:7025-7026` (unconditional `custom_name` assignment).

---

### 🧩 Affected Component

- [x] `mcpgateway` - API

---

### 🔁 Steps to Reproduce

1. Create two tools, A and B, with distinct names.
2. Open tool A's edit modal.
3. Change the Custom/Display Name field to tool B's technical name and save.
4. Observe the save succeeds with no error.
5. Reload the tools list — two rows now present the same name.

---

### 🤔 Expected Behavior

The edit save should be rejected with a 409 conflict, the same as create.

---

### 📓 Logs / Error Output

N/A

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Platform / OS | Linux |

---

Reproduced 3 times. Found via a manual-simulation E2E pass (real Chromium via Playwright).

sujithvishal:
# 🐛 Bug-fix PR

## 🔗 Related Issue

Closes #6057

---

## 💡 Fix Description

The UI edit form posts `name` as a hidden unchanged field, so `name_is_changing` is always
`False` and the existing conflict check never ran for `custom_name`-only edits.

Added a `_check_tool_name_conflict` call inside the `custom_name` assignment block
(`tool_service.py`) that fires when `custom_name` is changing and
`name_is_changing` is `False` — to make sue no tool have same custom_name.

---

## 📏 Reviewability

---

## 🧪 Verification

| Check | Command | Status |
|---|---|---|
| Coverage ≥ 80% | `make coverage` | ✅ |
| Manual regression | Open edit modal → rename `custom_name` to an existing tool's name → expect 409 | ✅  |

The new regression test `test_update_tool_custom_name_conflict_on_edit` fails on the unpatched
code and passes with the fix applied.

---

## 📐 MCP Compliance

---

## ✅ Checklist

---

## 📓 Notes

**Files changed:**

- `mcpgateway/services/tool_service.py` — naming conflict check added before `custom_name` assignment
  block.
- `tests/unit/mcpgateway/services/test_tool_service_coverage.py` — one new async test
  `test_update_tool_custom_name_conflict_on_edit` added to `TestUpdateToolAdditional`,
  reproducing the exact UI-path failure scenario.

### DIFF
--- a/mcpgateway/services/tool_service.py
+++ b/mcpgateway/services/tool_service.py
@@ -7017,7 +7017,6 @@ class ToolService(BaseService):
                     tool.custom_name = tool_update.name
                 tool.name = tool_update.name
 
-
             if tool_update.custom_name is not None:
                 if tool_update.custom_name != tool.custom_name and not name_is_changing:
                     tool_visibility_ref = tool.visibility if tool_update.visibility is None else tool_update.visibility.lower()
--- a/mcpgateway/services/tool_service.py
+++ b/mcpgateway/services/tool_service.py
@@ -7025,7 +7025,7 @@ class ToolService(BaseService):
             if tool_update.custom_name is not None:
                 if tool_update.custom_name != tool.custom_name and not name_is_changing:
                     tool_visibility_ref = tool.visibility if tool_update.visibility is None else tool_update.visibility.lower()
-                    self._check_tool_name_conflict( db,tool_update.custom_name,tool_visibility_ref,tool.id,team_id=tool.team_id,owner_email=tool.owner_email)
+                    self._check_tool_name_conflict(db, tool_update.custom_name, tool_visibility_ref, tool.id, team_id=tool.team_id, owner_email=tool.owner_email)
                 tool.custom_name = tool_update.custom_name
             if tool_update.displayName is not None:
                 tool.display_name = tool_update.displayName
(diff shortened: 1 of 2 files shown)


msureshkumar88 (changes requested):
@/tmp/claude-1000/-home-suresh-pr-others-block3-mcp-context-forge/9465c123-9ee4-4b6e-bf04-99113726c654/scratchpad/review_body.txt

sujithvishal:
@msureshkumar88  Acknowledged on points 1 and 3 — will address in this PR.

On point 2 (DB-level unique constraint): adding the constraint now carries a issue. If any existing database has duplicate custom_name rows — which is possible since this bug has been present before this fix — the Alembic migration will crash with an IntegrityError and the gateway will not start. User need to manually rename the duplicate rows before starting. Need some inputs on what action we can take here

PR #6130, opened 2026-08-07
feat(admin): add team picker to virtual server create/edit forms

narora88:
## Summary

- Adds an explicit **Team** dropdown on virtual server **create** and **edit** forms (reusable `server_team_field.html` partial).
- Edit form pre-selects the server's current team from `teamId` in `GET /admin/servers/{id}`.
- Server detail modal displays the assigned team name.
- Removes the edit-only hidden `team_id` injection that only worked when `?team_id=` was present in the URL.

## Problem

New virtual servers default to the creator's **personal team** because the admin UI had no team picker—only a hidden field when browsing a team-scoped catalog. Users could not assign or reassign a server to a shared team (e.g. General) from the form.

The backend already reads `team_id` via `_form_team_id()` on create/edit; this change adds the missing UI.

## Test plan

- [x] `npx vitest run tests/unit/js/server-team-select.test.js` (3 tests)
- [ ] Create virtual server → select a shared team → save → verify team in list/detail
- [ ] Edit existing server → team dropdown shows current assignment → change team → save
- [ ] Team-scoped catalog (`?team_id=`) still pre-selects that team on create

## Notes

Follow-up ideas for separate PRs/discussion:
- Slug field on virtual servers (fork parity)
- Whether `_validate_server_team_assignment` should allow any active team member vs team owners only

### DIFF
--- a/mcpgateway/admin_ui/servers.js
+++ b/mcpgateway/admin_ui/servers.js
@@ -15,7 +15,34 @@ import {
   showErrorMessage,
   decodeHtml,
   makeCopyIdButton,
+  getCurrentTeamId,
 } from "./utils.js";
+
+/**
+ * Resolve team id from API payload (camelCase or snake_case).
+ */
+export const teamIdFromServer = function (server) {
+  if (!server) {
+    return "";
+  }
+  return server.teamId || server.team_id || "";
+};
+
+/**
+ * Set the virtual-server team select to a preferred value.
+ */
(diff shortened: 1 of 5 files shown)


narora88:
@crivetimihai @brian-hussey @msureshkumar88 Hi Team, would it be possible for this change to be reviewed and considered to be part of the codebase ?

PR #6131, opened 2026-08-07
chore: patch timeouts for additionals span

prakhar-singh1928:
# Pull Request

## 🔗 Related Issue

Closes #6003 (follow-on: five tests in `test_otel_control_telemetry_e2e.py` each burned the full 60-second `asyncio.wait_for` timeout, making `make test` take 5+ extra minutes)

---

## 📝 Summary

Five tests in `tests/e2e/test_otel_control_telemetry_e2e.py` were silently spending 60 seconds each on every run — **passing**, but bloating `make test` by over 5 minutes. The tests never went red because the 60-second timeout produced a JSON-RPC error envelope that still satisfied the observability span assertions.

**Root cause:** Both test fixtures (`control_telemetry_app` and `denying_plugin_app`) patched `tool_service._http_client.request` directly. However, `ToolService.invoke_tool`'s SSRF connection-pinning path (line 5552) calls `_build_pinned_rest_http_client()` to build a **fresh** `ResilientHttpClient`, completely bypassing the patched instance attribute. The session-scoped `_deterministic_dns` conftest fixture stubs `socket.getaddrinfo` to return a real public IP (`93.184.215.14`) for any non-loopback host — including `internal.e2e-fixture.invalid` — so DNS "succeeded", the pinning branch was always entered, and a real TCP connection was attempted against the fixture host, burning the full timeout every time.

**Fix:** Added a `_mock_outbound_rest_request()` helper that patches the **factory function** `mcpgateway.services.tool_service._build_pinned_rest_http_client` via `monkeypatch.setattr`, returning a `SimpleNamespace` stub with `request`/`get`/`aclose` `AsyncMock` attributes. Both fixtures now call this helper. This mirrors the identical pattern already used in the sibling file `test_otel_plugin_metadata_e2e.py`. The now-unused `from mcpgateway.services.tool_service import tool_service` import is also removed.

No production code changed. No test logic or assertions changed.

---

## 📏 Reviewability

---

## 🏷️ Type of Change

- [x] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other

---

## 🧪 Verification

**Without fix** (stash applied, original code):

60.21s call  test_control_summary_span_present_in_db
60.11s call  test_control_result_span_per_plugin_present_in_db
60.10s call  test_control_telemetry_does_not_leak_tool_arguments
60.10s call  test_control_result_span_has_per_control_fields
60.10s call  test_control_summary_attributes_trusted_fields
Total: 7 passed, 2 skipped in 303.50s (5 min 3 sec)

**With fix** (this branch):
0.18s call  test_control_summary_span_present_in_db
0.09s call  test_control_result_span_per_plugin_present_in_db
0.09s call  test_control_telemetry_does_not_leak_tool_arguments
0.09s call  test_control_result_span_has_per_control_fields
0.09s call  test_control_summary_attributes_trusted_fields
Total: 7 passed, 2 skipped in 3.44s|

| Check | Command | Status |
|---|---|---|
| Lint suite | `make ruff` | ✅ clean |
| Target file tests | `pytest tests/e2e/test_otel_control_telemetry_e2e.py -v` | ✅ 7 passed, 2 skipped in 3.44s |

---

## ✅ Checklist

- [ ] Documentation updated (if applicable) — not applicable, test-only change

---

## 📓 Notes

**Why CI never caught this:** The tests always *passed* — the timeout produced a JSON-RPC error envelope that flowed through the plugin hooks and emitted the control telemetry spans the assertions check. CI shows green; only the per-test duration column in the log reveals the 60s cost, which is easy to miss.

**cpex PR #148 is unrelated:** That cpex fix (`PluginViolationError.executions`) is purely additive and the existing `mark_denied()` workaround in `tool_service.py` continues to work unchanged both before and after that cpex release. No adaptation required here.

**Pattern reference:** The `_mock_outbound_rest_request` helper is identical to the one already present in `tests/e2e/test_otel_plugin_metadata_e2e.py:229` — this file was the only e2e file missing the factory patch.

### DIFF
--- a/tests/e2e/test_otel_control_telemetry_e2e.py
+++ b/tests/e2e/test_otel_control_telemetry_e2e.py
@@ -52,6 +52,7 @@ from __future__ import annotations
 import os
 from pathlib import Path
 import secrets
+from types import SimpleNamespace
 from unittest.mock import AsyncMock
 
 # Third-Party
@@ -117,7 +118,6 @@ from mcpgateway.plugins import (  # noqa: E402
 from mcpgateway.plugins.policy import HOOK_PAYLOAD_POLICIES  # noqa: E402
 from mcpgateway.routers.observability import router as observability_router  # noqa: E402
 from mcpgateway.services.observability_service import ObservabilityService  # noqa: E402
-from mcpgateway.services.tool_service import tool_service  # noqa: E402
 from mcpgateway.utils.create_jwt_token import get_jwt_token  # noqa: E402
 from mcpgateway.utils.verify_credentials import require_admin_auth, require_auth  # noqa: E402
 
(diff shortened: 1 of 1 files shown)


msureshkumar88 (approved):
Nice catch, and a genuinely satisfying root-cause writeup — tracing the pinning bypass down to `_build_pinned_rest_http_client()` bypassing the instance-attribute patch explains exactly why these tests stayed green while burning 60s each.

Verified independently:
- Confirmed `tool_service.py:5552` calls the factory as a bare module-level function, so patching it via `monkeypatch.setattr` is the correct interception point (and `ssrf_protection_enabled` defaults `True`, so the pinning branch is genuinely exercised in these tests).
- Confirmed the new helper is a faithful mirror of the existing one in `test_otel_plugin_metadata_e2e.py:229`.
- Ran the target file locally: all tests pass in ~8s, consistent with the claimed fix.
- Grepped the rest of `tests/e2e/` for the same stale `_http_client` patch pattern — this was the only file affected, so the fix is correctly scoped.

Two non-blocking suggestions for a follow-up, not this PR:
- `_mock_outbound_rest_request()` is now duplicated verbatim across two test files — could move to a shared conftest helper to avoid drift.
- "Closes #6003" won't actually auto-close anything since #6003 is a merged PR, not an open issue — minor wording nit.

Approving.

PR #6136, opened 2026-08-07
fix: benchmark-mcp-tools tool-argument, denylist, and pool-cap drift (#6082)

### LINKED ISSUE #6082: [TESTING]: benchmark-mcp-tools silently skips and mishandles new fast-time-server tools
## Problem

`make benchmark-mcp-tools` (`MCPToolCallerUser` in `tests/loadtest/locustfile_mcp_protocol.py`) discovers tools via `tools/list` at startup then calls them at random. When the fast-time-server adds a new tool, the benchmark either calls it with wrong arguments or never calls it at all. Both outcomes silently corrupt the benchmark signal — failures inflate and new tools stay invisible.

Three bugs, all three must be fixed.

## Bug 1 — Unpinned image: tool set changes with no local signal

`docker-compose.yml` uses `:latest` in two places:

```yaml
# line 1478
image: ${FAST_TIME_IMAGE:-ghcr.io/ibm/cfex-mcp-fast-time-server:latest}

# line 1868 — hardcoded, no override var
image: ghcr.io/ibm/cfex-mcp-fast-time-server:latest
```

Any `docker compose pull` silently changes the server's tool set between runs. Benchmark regressions become unattributable to a code change vs. a server-side change.

**Fix:** Pin both to an explicit version tag. Expose `FAST_TIME_IMAGE` on line 1868 (already exists on 1478 but missing there). A version bump commit becomes the forcing function for auditing the tool set and updating the benchmark in the same diff.

## Bug 2 — Argument heuristic: new tools are called with `{}`

`_build_tool_args` (line 494) matches substrings in the tool name:

```python
if "convert" in name_lower: return {"time": ..., "source_timezone": ..., "target_timezone": ...}
if "time" in name_lower or "timezone" in name_lower: return {"timezone": ...}
if "echo" in name_lower: return {"message": ...}
return {}  # any other tool → empty args → JSON-RPC validation error
```

Any new tool whose name doesn't match those keywords gets called with `{}` and fails every time, inflating the failure rate. `locustfile.py` avoids this with an explicit `TOOLS_WITH_REQUIRED_ARGS` exclusion set; `locustfile_mcp_protocol.py` has no equivalent.

**Fix (pick one):**
- **Schema-driven (preferred):** during `_auto_detect`, read `inputSchema` from `tools/list` and synthesize valid args from the JSON Schema types. Self-healing — no manual update needed on version bumps.
- **Skip-list (minimal):** mirror the `TOOLS_WITH_REQUIRED_ARGS` pattern from `locustfile.py`. Unknown tools are skipped with a warning instead of called with `{}`. Must be updated alongside each image version bump.

## Bug 3 — Hardcoded 6-tool cap: new tools beyond position 6 are never called

```python
# MCPToolCallerUser line 746, MCPSessionChurnUser line 851
tool = random.choice(self._tool_names[:6] if len(self._tool_names) > 6 else self._tool_names)
```

Tools beyond index 5 in the `tools/list` response are permanently excluded from the benchmark regardless of how many the server exposes.

**Fix:** Remove the `[:6]` cap. Use the full pool, or a configurable `MCP_BENCHMARK_TOOL_POOL_SIZE` env var if a ceiling is needed.

## Files

| File | Change |
|---|---|
| `docker-compose.yml` lines 1478, 1868 | Pin image; add `FAST_TIME_IMAGE` override to line 1868 |
| `tests/loadtest/locustfile_mcp_protocol.py` | Fix `_build_tool_args`, add skip-list or schema arg builder, remove `[:6]` cap |

## Reference

- `tests/loadtest/locustfile.py:TOOLS_WITH_REQUIRED_ARGS` — the guard pattern to mirror
- `tests/loadtest/locustfile.py:FastTimeUser` — handles fast-time tools correctly; `locustfile_mcp_protocol.py` should reach parity

msureshkumar88:
## Summary

`make benchmark-mcp-tools` discovers fast-time-server tools via `tools/list` at
startup and calls them at random. Three bugs made that signal unreliable:

1. **Wrong arguments.** `_build_tool_args` guessed arguments from substrings in
   the tool name. The gateway prefixes every tool with `fast-time-`, so every
   name contains the substring `time` — `fast-time-echo` was called with
   `{"timezone": ...}` instead of `{"message": ...}` and rejected with
   `-32602`. This is worse than the issue described: it broke the tool the
   heuristic was written for (`echo`), not just unrecognized tools.
2. **No pool cap escape.** Five call sites (not the two the issue names)
   hard-capped the benchmark to the first 6 discovered tools
   (`self._tool_names[:6]`), silently dropping any tool beyond that.
3. **Unpinned image.** `docker-compose.yml` tracked `:latest` for the
   fast-time-server on both services, so `docker compose pull` could change
   the tool inventory under the benchmark with no local signal. Verified this
   already happened: the image `:latest` points to today differs from the one
   cached locally a month ago, and adds an 8th tool (`verify-protocol`).

Measured before this PR: `make benchmark-mcp-tools` reports **~30% failures**
(reproducing the 31.56% recorded in `docs/release/benchmark.md`). After: **0%**
on a healthy stack (2,112 requests, 0 failures — full numbers below).

## Fix

- **Schema-driven argument synthesis.** `_build_tool_args` now reads each
  tool's `inputSchema` (captured during `tools/list` discovery) and
  synthesizes arguments for its `required` properties via `_args_from_schema`
  / `_synth_value`. The old name-substring heuristic (`_legacy_name_args`) is
  kept only as a fallback for the `MCP_TOOL_NAMES` override path, where no
  schema exists — with `echo`/`convert` checked before the generic
  `time`/`timezone` branch, since gateway-prefixed names all contain `time`.
- **Denylist.** `MCP_BENCHMARK_TOOL_DENYLIST` (default `schema_error,flaky`)
  excludes the server's two deliberate failure-injection fixtures from the
  discovered pool at `_auto_detect` time, so they don't skew throughput
  numbers. Schemas can't express "designed to fail," so this is a second
  mechanism alongside the schema-driven arguments, not a replacement for it.
- **Unbounded tool pool.** All five `[:6]` call sites now route through
  `_limit_pool()`, gated by `MCP_BENCHMARK_TOOL_POOL_SIZE` (default `0` =
  every discovered tool; set to `6` to reproduce the original customer
  scenario). Plumbed through six env-var injection sites in the Makefile
  (two different indentation levels — the two-tab and three-tab forms are
  literal suffixes of each other, so each site was edited individually to
  avoid a find-and-replace corrupting the three-tab worker-loop lines).
- **Pinned image.** Both `fast_time_server` and `fast_test_server` now pin
  `ghcr.io/ibm/cfex-mcp-fast-time-server` by digest
  (`sha256:2f095509d5...`, equivalent to commit tag `9b97731f...`), both
  still honoring `FAST_TIME_IMAGE` for local overrides. Pinned by digest
  rather than a version tag because this image has no semver tags — GHCR
  holds only `latest` and commit-SHA tags.

## Scope decision: the OpenShift chart fork

`charts/mcp-stack/files/ocp/locustfile_mcp_protocol.py` is a ~1,000-line fork
of the benchmark script, deployed via
`charts/mcp-stack/profiles/ocp/values-pgo.yaml`. It is **not** in #6082's file
list, but it carried all three bugs, including a worse branch order in its
argument heuristic (`time` checked before `convert`, so `convert_time` was
*always* rejected there). Leaving it unfixed means OpenShift benchmark runs
keep reporting inflated failure rates after this merges. Included it in this
PR rather than filing a follow-up, since the fix is a mechanical, verified
port of the same logic.

## A bug found and fixed during review, not in the original issue

Final review surfaced two Important findings, both fixed and re-reviewed clean:

- `MCP_BENCHMARK_TOOL_DENYLIST=""` was documented as disabling the denylist
  but was a no-op — the shared `_cfg()` helper's `or`-chaining treats an
  empty string as falsy and falls through to the default. Fixed by parsing
  this one variable directly against env-var presence instead of `_cfg()`.
- `MCPSessionChurnUser.on_start` was `pass` (pre-existing, predates this PR),
  so it never called `_assign_target()` — every request from that user class
  targeted a broken path (`/servers//mcp`) and its `tools/call` step was
  silently always skipped. Fixed since it's one line in a method this PR's
  pool-cap task already touched, and the OpenShift fork's equivalent method
  was already correct — leaving it unfixed here would have left the two
  files diverged for no reason.

## Verification

Full validation gate, run once at the end against the final branch state:

| Check | Result |
|---|---|
| `make ruff` (changed files) | main locustfile clean; OCP fork 21 findings, down from 22 pre-branch, no new ones |
| `make pylint` (main locustfile) | 9.59/10 |
| `make test` | 21,754 passed, 0 failed |
| `make coverage diff-cover` | 98% total; touched files are outside `mcpgateway`'s instrumented coverage scope, so diff-cover has nothing to gate |
| `make detect-secrets-scan` | clean; baseline refreshed only for Makefile line-number drift from the new env-var injections |
| `make testing-up` | stack healthy, confirmed the pinned digest is what's actually running |
| `make benchmark-mcp-tools` | **0 failures / 2,112 requests** (30s run, 125 users) |

Additional live checks against the running stack:
- `tools/list` through the gateway reports 8 tools including `verify-protocol`
  (the tool that used to sit beyond the `[:6]` boundary).
- Direct `MCPAgentUser` run: all 6 non-denylisted tools called, 0 failures on
  each, including `verify-protocol`.
- `MCP_BENCHMARK_TOOL_DENYLIST=` (empty) now reports `tools=8` with no
  exclusion — confirming the opt-out fix.
- Sibling `tests/loadtest/locustfile.py` (untouched by this PR) still
  imports cleanly; `docker compose config` still resolves both services to
  the pinned digest with `FAST_TIME_IMAGE` override intact; OCP fork parses.

## Known follow-ups (not filed as issues yet, flagging for the reviewer)

- Four different pinned references to the same fast-time-server image now
  coexist (this PR's digest pin, two commit-tag pins in
  `plugin-integration.yml` / `run_plugin_tests.sh`, and `wrapper.yml`'s
  `:latest`). Worth a follow-up chore to unify.
- `tests/loadtest/locustfile.py`'s hand-maintained `TOOLS_WITH_REQUIRED_ARGS`
  list could be replaced by the same schema-driven approach — separate issue,
  large blast radius (8k+ lines).
- `schema_error` returns `isError: true` inside a successful JSON-RPC
  response, so it currently counts as a benchmark success. Whether MCP-level
  tool errors should count as failures is a deliberate metric change, not
  something this PR decided unilaterally.

Closes #6082

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^\\.secrets\\.baseline$",
     "lines": null
   },
-  "generated_at": "2026-08-10T13:58:42Z",
+  "generated_at": "2026-08-10T15:21:08Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -390,7 +390,7 @@
         "hashed_secret": "d3ac7a4ef1a838b4134f2f6e7f3c0d249d74b674",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 6065,
+        "line_number": 6079,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -398,7 +398,7 @@
         "hashed_secret": "5932862bcd24dd27d0dc0407ec94fe9d6ea24aeb",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 6562,
+        "line_number": 6576,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -406,7 +406,7 @@
         "hashed_secret": "c77c805e32f173e4321ee9187de9c29cb3804513",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 6574,
+        "line_number": 6588,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -414,7 +414,7 @@
         "hashed_secret": "8fe3df8a68ddd0d4ab2214186cbb8e38ccd0e06a",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 6646,
+        "line_number": 6660,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -422,7 +422,7 @@
         "hashed_secret": "bb589d0621e5472f470fa3425a234c74b1e202e8",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 7062,
+        "line_number": 7076,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -430,7 +430,7 @@
         "hashed_secret": "93ac8946882128457cd9e283b30ca851945e6690",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 7753,
+        "line_number": 7767,
         "type": "Secret Keyword",
         "verified_result": null
       }
@@ -836,7 +836,7 @@
         "hashed_secret": "2df14e4719f299249cd9a97cf68cc87232a27cbb",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1741,
+        "line_number": 1749,
         "type": "Hex High Entropy String",
         "verified_result": null
       },
@@ -844,7 +844,7 @@
         "hashed_secret": "293324f6824bb3a6db5c4dc42a60ddd4a9851c99",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1990,
+        "line_number": 1998,
         "type": "Hex High Entropy String",
         "verified_result": null
       },
@@ -852,7 +852,7 @@
         "hashed_secret": "afba9d98073dfb79fba7b21516c4d4d676c72935",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 2368,
+        "line_number": 2376,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -860,7 +860,7 @@
         "hashed_secret": "543d4766b92de8d18b1899c24e825638fd2a2bb7",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 2468,
+        "line_number": 2476,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -868,7 +868,7 @@
         "hashed_secret": "76a5af8c9ee406ff91da84664bc6f58cacb2ad76",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 2468,
+        "line_number": 2476,
         "type": "Base64 High Entropy String",
         "verified_result": null
       },
@@ -876,7 +876,7 @@
         "hashed_secret": "64f5490a2808803712ef99d9dfb8bc3ea5a15077",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 2481,
+        "line_number": 2489,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -884,7 +884,7 @@
         "hashed_secret": "fa9beb99e4029ad5a6615399e7bbae21356086b3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 2629,
+        "line_number": 2637,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -892,7 +892,7 @@
         "hashed_secret": "12be9f7db42eb4a2d881a99fa9ba847e1f83677f",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 2650,
+        "line_number": 2658,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -900,7 +900,7 @@
         "hashed_secret": "c3de40d5e3fc71ed62771c2127a8e42585026c97",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 2658,
+        "line_number": 2666,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -908,7 +908,7 @@
         "hashed_secret": "ce53277317aa3cd27c1619d7d371306ded2ddd1e",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 2663,
+        "line_number": 2671,
         "type": "Secret Keyword",
         "verified_result": null
       }
--- a/Makefile
+++ b/Makefile
@@ -2722,6 +2722,8 @@ MCP_BENCHMARK_WORKERS ?= 4
 MCP_BENCHMARK_MIXED_MASTER_PORT ?= 5567
 MCP_BENCHMARK_TOOLS_MASTER_PORT ?= 5569
 MCP_BENCHMARK_LOCUST_LOG_LEVEL ?= ERROR
+MCP_BENCHMARK_TOOL_POOL_SIZE ?= 0
+MCP_BENCHMARK_TOOL_DENYLIST ?= schema_error,flaky
 MCP_BENCHMARK_WORKER_LOG_DIR ?= reports/mcp_benchmark_workers
 RL_LIMIT_PER_MIN ?= 30
 
@@ -2774,6 +2776,8 @@ benchmark-mcp-mixed:                        ## Quick mixed MCP benchmark against
 	@test -d "$(VENV_DIR)" || $(MAKE) venv
 	@/bin/bash -eu -o pipefail -c 'source $(VENV_DIR)/bin/activate && \
 		LOCUST_LOG_LEVEL=$(MCP_BENCHMARK_LOCUST_LOG_LEVEL) MCP_SERVER_ID=$(MCP_BENCHMARK_SERVER_ID) \
+		MCP_BENCHMARK_TOOL_POOL_SIZE=$(MCP_BENCHMARK_TOOL_POOL_SIZE) \
+		MCP_BENCHMARK_TOOL_DENYLIST=$(MCP_BENCHMARK_TOOL_DENYLIST) \
 		locust -f $(MCP_PROTOCOL_LOCUSTFILE) \
 			--host=$(MCP_BENCHMARK_HOST) \
 			--users=$(MCP_BENCHMARK_USERS) \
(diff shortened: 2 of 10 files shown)


msureshkumar88:
## E2E verification against a live stack

Ran the same commands and `make` targets a maintainer would run to validate this fix, against `make testing-up`'s pinned-image stack. Config and raw output below.

### 0. Stack + pinned image

```
$ docker inspect --format '{{.Config.Image}}' mcp-context-forge-fast_time_server-1
ghcr.io/ibm/cfex-mcp-fast-time-server@sha256:2f095509d535a0769449f3b9417858f5c233dc74b1e09de335b48b7591d995a1
```

Confirms the digest pinned in `docker-compose.yml` (Task 4) is what's actually running — not `:latest`.

### 1. Tool inventory via the gateway (`tools/list` through `/servers/{id}/mcp`)

```
Gateway reports 8 tools:
  - fast-time-echo               required=['message']
  - fast-time-flaky              required=['key']
  - fast-time-get-system-time    required=None
  - fast-time-convert-time       required=['time', 'source_timezone', 'target_timezone']
  - fast-time-schema-error       required=None
  - fast-time-schema-success     required=None
  - fast-time-get-stats          required=None
  - fast-time-verify-protocol    required=None
```

8 tools, including `verify-protocol` — the one added since the image was last pulled locally, and the one that used to sit beyond the old `[:6]` cap.

### 2. `make benchmark-mcp-tools` — the actual failing command from #6082

```
$ make benchmark-mcp-tools MCP_BENCHMARK_RUN_TIME=30s MCP_BENCHMARK_LOCUST_LOG_LEVEL=INFO

[...] excluded 2 denylisted tool(s): fast-time-flaky, fast-time-schema-error
[...] server=9779b6698cbd4b4995ee04a4fab38737 name='mcp-streamable-http' tools=6 resources=0 prompts=0
[...] Ramping to 125 users at a rate of 30.00 per second
[...] All users spawned: {"MCPToolCallerUser": 125} (125 total users)

Type     Name                              # reqs   # fails  |   Avg    Min    Max    Med |  req/s
POST     MCP initialize                       125    0(0.00%)|  3763     28  13871   3300 |   4.13
POST     MCP tools/call [rapid]               1888   0(0.00%)|   849     33  17330    170 |  62.40
POST     MCP tools/list [rapid]                 99   0(0.00%)|  1101     70  12807    290 |   3.27
Aggregated                                     2112   0(0.00%)|  1034     28  17330    180 |  69.80

  Total Requests:            2,112
  Total Failures:                0 (0.00%)
  Requests/sec (RPS):        69.80
```

**0.00% failures, 2,112 requests.** Before this fix, the same command reproduces the ~30% failure rate `docs/release/benchmark.md` records (31.56%) — confirmed on this branch's base commit prior to the fix, in the same environment.

Denylist confirmed working: `excluded 2 denylisted tool(s): fast-time-flaky, fast-time-schema-error`, leaving `tools=6` for the benchmark to exercise (8 discovered − 2 denylisted).

### 3. Per-tool coverage (proves the `[:6]` cap is gone)

```
$ MCP_SERVER_ID=9779b6698cbd4b4995ee04a4fab38737 \
  locust -f tests/loadtest/locustfile_mcp_protocol.py --host=http://localhost:8080 \
      --users=20 --spawn-rate=10 --run-time=30s --headless --only-summary MCPAgentUser \
  | grep "MCP tools/call \["

  MCP tools/call [fast-time-convert-time]      53   0(0.00%)
  MCP tools/call [fast-time-echo]              53   0(0.00%)
  MCP tools/call [fast-time-get-system-time]   50   0(0.00%)
  MCP tools/call [fast-time-schema-success]    50   0(0.00%)
  MCP tools/call [fast-time-verify-protocol]   49   0(0.00%)
  MCP tools/call [fast-time-get-stats]         35   0(0.00%)
```

All 6 non-denylisted tools called, **0 failures on every single one** — including `fast-time-verify-protocol`, which sat beyond the old hardcoded `[:6]` boundary and would never have been called before this fix.

### 4. Denylist opt-out regression check (the bug found in final review)

```
$ make benchmark-mcp-tools MCP_BENCHMARK_RUN_TIME=10s MCP_BENCHMARK_LOCUST_LOG_LEVEL=INFO MCP_BENCHMARK_TOOL_DENYLIST=

[...] server=9779b6698cbd4b4995ee04a4fab38737 name='mcp-streamable-http' tools=8 resources=0 prompts=0
  Total Requests:              139
  Total Failures:                0 (0.00%)
```

`MCP_BENCHMARK_TOOL_DENYLIST=""` now correctly reports `tools=8` — no exclusion line at all. Before the final-review fix, this silently fell back to the default denylist (`schema_error,flaky`) due to `_cfg()`'s `or`-chaining treating an empty string as falsy.

### 5. Surrounding breakage sweep

```
$ python -c "import tests.loadtest.locustfile_mcp_protocol as lf; ..."
imports cleanly: tests.loadtest.locustfile_mcp_protocol
user classes: ['MCPAgentUser', 'MCPToolCallerUser', 'MCPDiscoveryUser', 'MCPSessionChurnUser', 'MCPStressUser', 'RESTBaselineUser']

$ python -c "import tests.loadtest.locustfile as lf; ..."   # sibling file, untouched by this PR
sibling locustfile.py imports cleanly: tests.loadtest.locustfile

$ docker compose --profile testing config | ...
fast_time_server: ghcr.io/ibm/cfex-mcp-fast-time-server@sha256:2f095509d5...
fast_test_server: ghcr.io/ibm/cfex-mcp-fast-time-server@sha256:2f095509d5...

$ make -n benchmark-mcp-tools MCP_BENCHMARK_TOOL_POOL_SIZE=6 MCP_BENCHMARK_TOOL_DENYLIST=schema_error,flaky
MCP_BENCHMARK_TOOL_DENYLIST=schema_error,flaky
MCP_BENCHMARK_TOOL_POOL_SIZE=6

$ python -c "import ast; ast.parse(open('charts/mcp-stack/files/ocp/locustfile_mcp_protocol.py').read())"
OCP fork: syntax OK
```

Nothing outside this PR's own files broke: the untouched sibling `locustfile.py` still imports, both compose services still resolve to the pinned digest with the `FAST_TIME_IMAGE` override intact, the new Makefile variables reach `locust` correctly, and the OCP fork still parses.

### 6. Unit tests, one more time against the live-stack-adjacent state

```
$ LOCUST_SKIP_MONKEY_PATCH=1 python -m pytest tests/unit/loadtest/ -v
============================== 36 passed in 1.35s ==============================
```

### Full validation gate (run once, at the end, per branch)

| Check | Result |
|---|---|
| `make ruff` (changed files) | main locustfile clean; OCP fork 21 findings, down from 22 pre-branch |
| `make pylint` (main locustfile) | 9.59/10 |
| `make test` | 21,754 passed, 0 failed |
| `make coverage diff-cover` | 98% total; touched files outside `mcpgateway`'s instrumented scope, nothing for diff-cover to gate |
| `make detect-secrets-scan` | clean, baseline refreshed for Makefile line-number drift only |
| `make testing-up` | healthy, pinned digest confirmed |
| `make benchmark-mcp-tools` | 0 failures / 2,112 requests |

One honest caveat: this ran on a shared, resource-contended sandbox, so absolute RPS numbers here (62-70 req/s) shouldn't be read as a throughput benchmark — the metric that matters for #6082 is the failure rate, and that went from ~30% to 0.00% with the same command, same stack, same tool set.

Lang-Akshay (dismissed):
LGTM

gandhipratik203 (dismissed):
LGTM! Thanks for the tests and the results upfront!

Lang-Akshay (approved):
LGTM

PR #6139, opened 2026-08-08
fix(admin-ui): attach CSRF token to gateway refresh and prompt render

### LINKED ISSUE #6071: [BUG]: Gateway "refresh tools" action always fails with CSRF error for cookie-session users
### 🐞 Bug Summary

The Catalog tab's "refresh tools" action for a gateway always fails with a 403 CSRF error for browser (cookie-session) users — `admin_ui/gateways.js` never attaches a CSRF token to this request at all.

`POST /gateways/{id}/tools/refresh` is sent with only an `Accept` header — `admin_ui/gateways.js` contains zero CSRF-related code, unlike every other mutating call in the admin UI (contrast `utils.js:155-156`, which correctly attaches `X-CSRF-Token`). The server correctly rejects the request (`middleware/csrf_middleware.py:181-186`, logging `CSRF cookie missing`); the route itself (`main.py:7615`) is fine. This fails closed (the action simply doesn't work) rather than being a security bypass, but it means the feature is unusable from the browser.

---

### 🧩 Affected Component

---

### 🔁 Steps to Reproduce

1. Register a gateway with discovered tools.
2. In the Catalog tab, trigger "refresh tools" for that gateway.
3. Observe a 403, and check the server log for "CSRF cookie missing".

---

### 🤔 Expected Behavior

The refresh-tools request should attach the CSRF token the same way every other admin UI mutation does.

---

### 📓 Logs / Error Output

Server log: `CSRF cookie missing` on every refresh attempt.

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Platform / OS | Linux |

---

Reproduced 5 times. Same defect class as a separately-reported prompt-render CSRF bug, different call site — the underlying route is unaffected.

### LINKED ISSUE #6072: [BUG]: Prompt "Render Prompt" test action always fails with CSRF error
### 🐞 Bug Summary

The "Render Prompt" test action on the Prompts tab always fails with a 403 CSRF error — `admin_ui/prompts.js` never attaches a CSRF token to this request at all, unlike every other mutating call in the admin UI. Prompt rendering and argument substitution are therefore untestable from the UI.

`POST /prompts/{id}` is sent with only a `Content-Type` header — `mcpgateway/admin_ui/prompts.js:1182-1192` contains no CSRF-related code, in contrast to the correct pattern used elsewhere (`utils.js:150-158`, `auth.js:202-204`, `formHandlers.js:42-48`). The server correctly rejects the request (`config.py:418-440` — `/prompts` is not in `csrf_exempt_paths`); the route itself is unaffected. This fails closed (the action simply doesn't work) rather than being a security bypass, but it means the feature is unusable from the browser.

---

### 🧩 Affected Component

---

### 🔁 Steps to Reproduce

1. Create a prompt with a template and at least one argument.
2. Open its row menu and click "Test" / "Render Prompt".
3. Fill in argument values and submit.
4. Observe `Error: HTTP 403: Forbidden`.
5. Check the server log for a CSRF-related rejection on `POST /prompts/{id}`.

---

### 🤔 Expected Behavior

The render request should attach `X-CSRF-Token` the same way every other admin UI mutation does, and rendering should succeed.

---

### 📓 Logs / Error Output

Server-side: CSRF rejection logged on `POST /prompts/{id}` (route not in `csrf_exempt_paths`).

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Platform / OS | Linux |

---

Reproduced 4 times, including a fresh browser context and fresh login. Same defect class as a separately-reported gateway "refresh tools" CSRF bug, different call site — the underlying route is unaffected.

Found via a manual-simulation E2E pass (real Chromium via Playwright).

KlyneChrysler:
Closes #6071
Closes #6072

Two admin UI actions POST without the X-CSRF-Token header that every other mutating call attaches, so both fail closed with a 403 for cookie-session users:

- gateways.js refreshGatewayTools POSTs to /gateways/{id}/tools/refresh with only an Accept header (#6071)
- prompts.js Render Prompt POSTs to /prompts/{id} with only a Content-Type header (#6072)

Both now attach the token using the same conditional getCookie("mcpgateway_csrf_token") pattern as utils.js, importing getCookie from the shared module. The routes and server-side CSRF middleware are unchanged and were already correct.

Regression tests added in tests/unit/js/gateway.test.js and tests/unit/js/prompts.test.js assert each POST carries the X-CSRF-Token header when the cookie is present and omits it when absent. Both fail if the header spread is reverted. The two suites pass (110 tests), and the existing CSRF suite is untouched.

### DIFF
--- a/mcpgateway/admin_ui/gateways.js
+++ b/mcpgateway/admin_ui/gateways.js
@@ -20,6 +20,7 @@ import {
   buildTableUrl,
   decodeHtml,
   fetchWithTimeout,
+  getCookie,
   getCurrentTeamId,
   handleFetchError,
   isInactiveChecked,
@@ -1817,7 +1818,12 @@ export const refreshGatewayTools = async function (gatewayId, gatewayName, butto
       {
         method: "POST",
         credentials: "include", // pragma: allowlist secret
-        headers: { Accept: "application/json" },
+        headers: {
+          Accept: "application/json",
+          ...(getCookie("mcpgateway_csrf_token")
+            ? { "X-CSRF-Token": getCookie("mcpgateway_csrf_token") }
+            : {}),
+        },
       }
     );
 
--- a/mcpgateway/admin_ui/prompts.js
+++ b/mcpgateway/admin_ui/prompts.js
@@ -7,6 +7,7 @@ import { applyVisibilityRestrictions, isTeamScopedView } from "./teams.js";
 import {
   decodeHtml,
   fetchWithTimeout,
+  getCookie,
   getCurrentTeamId,
   handleFetchError,
   isInactiveChecked,
@@ -1185,6 +1186,9 @@ export const runPromptTest = async function () {
         method: "POST",
         headers: {
           "Content-Type": "application/json",
+          ...(getCookie("mcpgateway_csrf_token")
+            ? { "X-CSRF-Token": getCookie("mcpgateway_csrf_token") }
+            : {}),
         },
         credentials: "include", // pragma: allowlist secret
         body: JSON.stringify(args),
--- a/tests/unit/js/gateway.test.js
+++ b/tests/unit/js/gateway.test.js
@@ -60,6 +60,16 @@ vi.mock("../../../mcpgateway/admin_ui/utils", () => ({
   }),
   decodeHtml: vi.fn((s) => s || ""),
   fetchWithTimeout: vi.fn(),
+  // Real cookie-reading implementation (mirrors utils.getCookie) so the CSRF
+  // spread in refreshGatewayTools is exercised against jsdom's document.cookie.
+  getCookie: vi.fn((name) => {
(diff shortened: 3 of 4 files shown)


KlyneChrysler:
The pytest failure here looks unrelated to this PR, which only touches JavaScript (gateways.js, prompts.js, and two tests/unit/js files). The failing assertion is in tests/unit/mcpgateway/routers/test_email_auth_router.py:663, a time-limited test whose own comment reads "Code to be removed after Sun, 16 Aug 2026 23:59:59 UTC" (#2754). It asserts now < that sunset date, so it fails on any run after Aug 16 regardless of the change. Happy to help if there is anything on my end, but I do not think this branch is the cause.

KlyneChrysler:
Filed #6298 to track the expired sunset tripwire behind that pytest failure, with the removal steps and an offer to prepare the PR.

PR #5929, opened 2026-07-27
Added team_id filter into get_gateways

cafalchio:
Fixes the team_id query parameter being silently ignored by the GET /gateways endpoint. The parameter was not wired through the endpoint handler to the service layer, causing all gateways to be returned regardless of the team filter.

# Pull Request

## 🔗 Related Issue

Closes #https://github.com/IBM/mcp-context-forge/issues/5496  
 
---

## 📝 Summary

_What does this PR do and why?_

---

### mcpgateway/admin.py
- Added `team_id: Optional[str] = Depends(_validated_team_id_param)` parameter to `admin_list_gateways()` endpoint
- Updated docstring to document the new parameter
- Passes `team_id` to `gateway_service.list_gateways()` for proper filtering

### tests/unit/mcpgateway/test_admin.py
- Added `test_admin_list_gateways_team_filtering()` test that validates:
  - Filtering by team1_id returns only team1 + public gateways
  - Filtering by team2_id returns only team2 + public gateways
  - Team isolation (team1 filter excludes team2 gateways and vice versa)
- Added `test_admin_list_gateways_with_invalid_team_id()` test for invalid UUID validation

## 📏 Reviewability

- [x] This PR has one clear purpose
- [x] The linked issue is not labeled `triage`
- [x] Unrelated bugs or improvements are tracked in separate issues/PRs
- [x] Tests are included with the code they validate
- [x] If AI-assisted, I understand and can explain the generated changes

---

## 🏷️ Type of Change

- [x] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Chore (deps, CI, tooling)
- [ ] Other (describe below)

---

## 🧪 Verification

_List exact commands, screenshots, videos, logs, reproduction steps, or manual validation. If evidence is not feasible, explain why._

| Check                     | Command         | Status |
|---------------------------|-----------------|--------|
| Lint suite                | `make lint`     |   x    |
| Unit tests                | `make test`     |   x    |
| Coverage ≥ 80%            | `make coverage` |   x    |

---

## ✅ Checklist

- [ ] Code formatted (`make black isort pre-commit`)
- [ ] Tests added/updated for changes
- [ ] Documentation updated (if applicable)
- [ ] No secrets or credentials committed

---

## 📓 Notes (optional)

_Screenshots, design decisions, or additional context._

Issue #5496, opened 2026-07-03
[BUG]: GET /gateways: team_id filter param ignored


peterzaccha:
### 🐞 Bug Summary
The `team_id` query parameter passed to `GET /gateways` is silently ignored for user admin — all gateways are returned regardless of the value supplied.
---

### 🧩 Affected Component
Select the area of the project impacted:

- [X] `mcpgateway` - API
- [ ] `mcpgateway` - UI (admin panel)
- [ ] `mcpgateway.wrapper` - stdio wrapper
- [ ] Federation or Transports
- [ ] CLI, Makefiles, or shell scripts
- [ ] Container setup (Docker/Podman/Compose)
- [ ] Other (explain below)

---

### 🔁 Steps to Reproduce

1. Create a team via `POST /v1/teams/` and note the returned `id`.
2. Create a gateway via `POST /gateways` with `team_id` set to the team ID from step 1.
3. Call `GET /gateways?team_id=<team_id>` with the same team ID.
4. Observe that gateways belonging to **other** teams (or no team) are also returned.

---

### 🤔 Expected Behavior
`GET /gateways?team_id=<id>` should return **only** gateways where `team_id` matches the supplied value.
---

### 📓 Logs / Error Output
N/A

---

### 🧠 Environment Info
You can retrieve most of this from the `/version` endpoint.

| Key | Value |
|-----|-------|
| Version or commit | 1.0.4 |
| Runtime | `3.13.5` |
| Platform / OS | `macOS` |
| Container | `none` |

PR #6151, opened 2026-08-09
Apply team_id filter for admin callers on tools, resources, prompts, servers and a2a (#6150)

### LINKED ISSUE #6150: [BUG]: team_id filter ignored for admin users on tools, resources, prompts, servers and A2A endpoints
### 🐞 Bug Summary

The `team_id` query parameter is silently ignored for admin users on `GET /tools`, `GET /resources`, `GET /prompts`, `GET /servers` and `GET /a2a` — all items are returned regardless of the value supplied.

This is the same defect as #5496, which was fixed for `GET /gateways` in #5929. That fix was applied inside `gateway_service.list_gateways`, the five sibling endpoints must be fixed also

---

### 🧩 Affected Component

- [X] `mcpgateway` - API
- [ ] Other (explain below)

---

### 🔁 Steps to Reproduce

1. Create two teams via `POST /teams/` and note both ids.
2. Create a tool, resource, prompt, server and A2A agent in **each** team.
3. As a platform admin, call `GET /tools?team_id=<team-one-id>` (and the same for `/resources`, `/prompts`, `/servers`, `/a2a`).
4. Observe that items belonging to team two are also returned.

Same four steps as #5496, pointed at the other endpoints.

Observed on a local instance running #5929, asking each endpoint for team one only:

| endpoint | returned |
|---|---|
| `/gateways` | team one only ✅ (fixed by #5929) |
| `/tools` | team one **+ team two's** |
| `/resources` | team one + team two's |
| `/prompts` | team one + team two's |
| `/servers` | team one + team two's |
| `/a2a` | team one + team two's |

Team membership was confirmed against the database rather than item names. One of the leaked tools is `visibility=team` scoped to team two rather than public, so the filter is genuinely being ignored rather than deliberately widened.

---

### 🤔 Expected Behavior

`GET /<endpoint>?team_id=<id>` should return only items whose `team_id` matches the supplied value — consistent with the behaviour `GET /gateways` has after #5929.

All six entity types carry a `team_id` column and all six endpoints already accept and document the parameter as "Filter by specific team ID", so this is about making the parameter do what it already advertises.

---

### 📓 Root Cause

All six services inherit `BaseService._apply_access_control()` (`mcpgateway/services/base_service.py:63`) and each calls it exactly once, as the sole place `team_id` is applied:

- `tool_service.py:2826`
- `resource_service.py:1267`
- `prompt_service.py:1468`
- `server_service.py:834`
- `a2a_service.py:1003`
- `gateway_service.py:2443`

Both admin branches in that function return before `team_id` is used (`base_service.py:109-123`):

```python
    return query.where(model_cls.visibility != "private")   # team_id unused

if token_teams is None:
    if user_is_admin:
        return query.where(or_(...))                        # team_id unused
```

Only the fall-through at line 139 passes `team_id` to `_apply_visibility_filter`, which is why the filter works for non-admin callers and not for admins.

Confirmed locally: applying `team_id` before those early returns — a change to `base_service.py` alone, with no other file touched — makes all five endpoints filter correctly.

---

### 💡 Proposed Fix

Apply `team_id` in `_apply_access_control()` above the admin early returns, so all six endpoints are fixed in one place:

```python
if team_id:
    query = query.where(model_cls.team_id == team_id)
```

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Version or commit | 1.0.7 |
| Runtime | 3.12 |
| Platform / OS | macOS |
| Container | none |

---

eeshsaxena:
Fixes #6150

### Problem

The `team_id` query parameter is silently ignored for admin callers on `GET /tools`, `GET /resources`, `GET /prompts`, `GET /servers` and `GET /a2a`. Every item is returned regardless of the value supplied, including `visibility=team` items scoped to a different team.

This is the same defect #5496 described for gateways, which #5929 fixed in `list_gateways`. The five sibling endpoints were never fixed.

### Root cause

All six list methods apply `team_id` through `BaseService._apply_access_control()`, but both admin bypass branches return before `team_id` is used:

```python
    return query.where(model_cls.visibility != "private")   # team_id unused

if token_teams is None:
    if is_user_admin(db, user_email):
        return query.where(or_(...))                        # team_id unused
```

Only the non-admin fall-through passes `team_id` to `_apply_visibility_filter`, which is why the filter works for regular users but not admins.

`list_gateways` already handles this by applying an exact filter *after* `_apply_access_control`:

```python
query = await self._apply_access_control(query, db, user_email, token_teams, team_id)
if team_id:
    query = query.where(DbGateway.team_id == team_id)
```

### Fix

Apply the identical exact-team filter in the five sibling list methods (`list_tools`, `list_resources`, `list_prompts`, `list_servers`, `list_agents`), matching the merged gateway behaviour. This keeps the change consistent with #5929 and touches only the list paths.

### Tests

Added `tests/unit/mcpgateway/services/test_team_id_admin_filter.py`, a parameterized test asserting the compiled query for an admin caller contains `<table>.team_id = '<id>'` on all five endpoints. Verified it fails on every case before this change and passes after. `test_base_service.py` and `test_authorization_access.py` still pass (96 passed), so non-admin visibility semantics are unchanged.

### DIFF
--- a/mcpgateway/services/a2a_service.py
+++ b/mcpgateway/services/a2a_service.py
@@ -1002,6 +1002,9 @@ class A2AAgentService(BaseService):
 
         query = await self._apply_access_control(query, db, user_email, token_teams, team_id)
 
+        if team_id:
+            query = query.where(DbA2AAgent.team_id == team_id)
+
         if visibility:
             query = query.where(DbA2AAgent.visibility == visibility)
 
--- a/mcpgateway/services/prompt_service.py
+++ b/mcpgateway/services/prompt_service.py
@@ -1467,6 +1467,9 @@ class PromptService(BaseService):
 
             query = await self._apply_access_control(query, db, user_email, token_teams, team_id)
 
+            if team_id:
+                query = query.where(DbPrompt.team_id == team_id)
+
             if visibility:
                 query = query.where(DbPrompt.visibility == visibility)
 
--- a/mcpgateway/services/resource_service.py
+++ b/mcpgateway/services/resource_service.py
@@ -1266,6 +1266,9 @@ class ResourceService(BaseService):
 
             query = await self._apply_access_control(query, db, user_email, token_teams, team_id)
 
+            if team_id:
+                query = query.where(DbResource.team_id == team_id)
+
             if visibility:
                 query = query.where(DbResource.visibility == visibility)
 
--- a/mcpgateway/services/server_service.py
+++ b/mcpgateway/services/server_service.py
@@ -833,6 +833,9 @@ class ServerService(BaseService):
 
         query = await self._apply_access_control(query, db, user_email, token_teams, team_id)
 
+        if team_id:
+            query = query.where(DbServer.team_id == team_id)
+
         if visibility:
             query = query.where(DbServer.visibility == visibility)
 
--- a/mcpgateway/services/tool_service.py
+++ b/mcpgateway/services/tool_service.py
@@ -2825,6 +2825,9 @@ class ToolService(BaseService):
                 query = query.where(DbTool.enabled)
             query = await self._apply_access_control(query, db, user_email, token_teams, team_id)
 
+            if team_id:
+                query = query.where(DbTool.team_id == team_id)
+
             if visibility:
                 query = query.where(DbTool.visibility == visibility)
 
--- a/tests/unit/mcpgateway/services/test_team_id_admin_filter.py
+++ b/tests/unit/mcpgateway/services/test_team_id_admin_filter.py
@@ -1,0 +1,63 @@
+# -*- coding: utf-8 -*-
+"""Regression tests for admin team_id filtering on list endpoints.
+
+Companion to the gateway fix in #5929 / issue #5496. For admin callers the
+``_apply_access_control`` bypass branches return before ``team_id`` is applied,
+so ``GET /tools|/resources|/prompts|/servers|/a2a?team_id=<id>`` used to ignore
+the filter and leak items from other teams (issue #6150). Each list method now
+applies an exact ``team_id`` filter after access control, exactly as
+``list_gateways`` already does.
+"""
+
+# Standard
+from types import SimpleNamespace
+from unittest.mock import AsyncMock, MagicMock
+
+# Third-Party
+import pytest
+
+# First-Party
+from mcpgateway.services.a2a_service import A2AAgentService
+from mcpgateway.services.prompt_service import PromptService
+from mcpgateway.services.resource_service import ResourceService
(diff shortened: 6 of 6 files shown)


Yosiefeyob:
Thanks for jumping on this — your diagnosis is right, and it's the same two early returns in `_apply_access_control` I'd traced.

Full disclosure: I've opened #6152 as an alternative, mainly because of the public-visibility question raised on #6150. An important question:

- It keeps the globally-public condition from #4773, so filtering by a team doesn't hide public items owned by other teams.

Not sure which behaviour maintainers want, since #5929 already shipped strict for `/gateways`. If the answer is "strict", yours is the correct one and I'll close mine. Great work @eeshsaxena !

eeshsaxena:
Thanks @Yosiefeyob, and no worries at all about the parallel PR, glad we landed on the same root cause.

On the strict-vs-permissive question, my case for strict is consistency: #5929 already shipped strict for `/gateways` (it applies `DbGateway.team_id == team_id` after `_apply_access_control`, which does drop globally-public gateways owned by other teams). If the five sibling endpoints keep the global-public condition, we end up with `/gateways` behaving one way and `/tools|/resources|/prompts|/servers|/a2a` behaving another for the same `?team_id=` filter, which is the exact inconsistency #6150 set out to remove. This PR makes all six identical to what gateways already does.

There's also the intent in the report: #6150's repro shows a `visibility=team` item scoped to *another* team leaking, and the expected behaviour is "return only items whose team_id matches". Strict satisfies that directly.

That said, whether a globally-public item from team B should still surface when a caller filters by team A is a genuine product call, and if maintainers want that widening then it's deliberate and #6152 is the right shape. If they go that way I'm happy to close this one. Absent that decision though, matching the already-shipped gateways behaviour seems like the safer default. Either way, appreciate the thorough tracing on your end.

Yosiefeyob:
Thanks — fair argument, and one clarification: #6152 fixes the repro case too. Other teams' visibility=team rows are still excluded; the only rows that differ between our PRs are globally-public ones owned by another team.
Both approaches end with all six endpoints consistent — mine just retrofits /gateways to match the other five rather than the reverse.

I believe maintainers are heading with the 4773 approach, so going that way. #4773 which has a much more vetted issue (which fixed #4732) takes precedent.

Once again thank you @eeshsaxena.

eeshsaxena:
Makes sense, and we're after the same end state: all six endpoints applying the team filter consistently while globally-public items stay visible. If the maintainers prefer #6152 as the way to get there, I'm happy to defer and close this in its favor. The repro case being covered either way is the important part.

jonpspri:
Fixed with #6152

PR #6152, opened 2026-08-09
Fix/team id filter all list endpoints

### LINKED ISSUE #6150: [BUG]: team_id filter ignored for admin users on tools, resources, prompts, servers and A2A endpoints

Yosiefeyob:
# Pull Request

## 🔗 Related Issue

Closes #6150 

Related: #5496, #5929 (gateways-only fix), #4732 / #4773 (global-public visibility)

---

## 📝 Summary

The `?team_id=` filter was silently ignored for admin callers on every list endpoint — `/gateways`, `/tools`, `/resources`, `/prompts`, `/servers` and `/a2a`. Filtering by one team returned every team's items.

The cause is in the shared `BaseService._apply_access_control()`. Both admin branches `return` before `team_id` is ever used:

```python
    return query.where(model_cls.visibility != "private")   # team_id unused

if token_teams is None:
    if user_is_admin:
        return query.where(or_(...))                        # team_id unused
```

Only the fall-through path passed `team_id` to `_apply_visibility_filter`, which is why the filter worked for non-admin callers and not for admins. All six services inherit this method and call it as the sole place `team_id` is applied, so all six were affected.

Both branches now narrow to the requested team **while keeping the globally-public condition**.

### Why not a strict filter

This PR **removes** the `team_id` filter added to `gateway_service.list_gateways` in #5929. That is intentional, and it is not a rollback of that fix.

#5929 filtered strictly, which also excluded `visibility='public'` rows owned by other teams. That behaviour was deliberately introduced by #4773 (closing #4732, where a user selected a team in the Admin UI and public agents disappeared). #4773's description is explicit:

> The `?team_id=` parameter now means "items belonging to team X plus all globally-public items from any team" instead of "items strictly belonging to team X". [...] Strict data isolation remains the responsibility of `token_teams` (Layer 1 token scoping), not this UI routing hint.

So `main` currently answers the same question two different ways: `/gateways` filters strictly, the other five do not. This PR resolves that in favour of the documented #4773 semantics, and fixes the actual leak reported in #5496 — admins seeing other teams' **team-scoped** rows.

Behaviour when filtering by team alpha, identical for admin, anonymous and non-admin callers:

| item | before | after |
|---|---|---|
| alpha's team-scoped row | returned | returned |
| alpha's public row | returned | returned |
| **beta's team-scoped row** | **returned** | **excluded** |
| beta's public row | returned | returned |

If maintainers prefer strict filtering instead, #6151 fixes this with a strict team_id match, which also excludes visibility='public' rows owned by other teams. So two different ways to answer the issue dependent on maintainers decision
---

## 📏 Reviewability

---

## 🏷️ Type of Change

- [x] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

Reproduced with the steps from #5496 against a local gateway (two teams, items in each, platform admin session). Before: filtering by team one returned both teams' items on all six endpoints. After: other teams' team-scoped items are excluded, public items still returned.

| Check | Command | Status |
|---|---|---|
| Ruff | `make ruff` | ✅ All checks passed |
| Docstring coverage | `make interrogate` | ✅ 100% |
| Pre-commit hooks | `pre-commit run --files <changed>` | ✅ All passed |
| Unit tests | `pytest tests/unit` | ✅ 28 failed / 20846 passed — identical 28 failures on `main` (28 / 20841), zero new |

---

## ✅ Checklist

- [ ] Documentation updated (if applicable) — n/a, no user-facing docs describe this filter

---

## 📓 Notes

Changed files:

| File | Change |
|---|---|
| `mcpgateway/services/base_service.py` | Apply `team_id` in both admin branches, keeping the global-public condition; document the argument's meaning |
| `mcpgateway/services/gateway_service.py` | Remove the `team_id` filter from #5929 (now handled in the shared method, and it excluded public rows) |
| `tests/unit/mcpgateway/services/test_gateway_service.py` | Update the #5929 test — it asserted strict exact-match |
| `tests/.../test_{tool,resource,prompt,server,a2a}_service.py` | One regression test each, modelled on the #5929 test |

Each test asserts both halves of the contract — rows narrowed to the requested team, and the global-public condition still ORed in — so neither property can regress unnoticed. Only `/gateways` had such a test before, which is why the other five drifted.

### DIFF
--- a/mcpgateway/admin.py
+++ b/mcpgateway/admin.py
@@ -17024,16 +17024,13 @@ async def get_resources_section(
         user_email, token_teams = get_scoped_resource_access_context(request, user)
         LOGGER.debug(f"User {user_email} requesting resources section with team_id={team_id}, token_teams={token_teams}")
 
-        # Get all resources with token_teams for proper scoping
-        resources_result = await local_resource_service.list_resources(db, include_inactive=True, user_email=user_email, token_teams=token_teams)
+        # Filter in the service, not here: a strict team_id comparison would drop
+        # globally-public rows owned by other teams.
+        resources_result = await local_resource_service.list_resources(db, include_inactive=True, user_email=user_email, token_teams=token_teams, team_id=team_id)
         if isinstance(resources_result, tuple):
             resources_list = resources_result[0]
         else:
             resources_list = resources_result
-
-        # Apply team filtering if specified
-        if team_id:
-            resources_list = [r for r in resources_list if getattr(r, "team_id", None) == team_id]
 
         # Convert to JSON-serializable format
         resources = []
@@ -17085,16 +17082,13 @@ async def get_prompts_section(
         user_email, token_teams = get_scoped_resource_access_context(request, user)
         LOGGER.debug(f"User {user_email} requesting prompts section with team_id={team_id}, token_teams={token_teams}")
 
-        # Get all prompts with token_teams for proper scoping
-        prompts_result = await local_prompt_service.list_prompts(db, include_inactive=True, user_email=user_email, token_teams=token_teams)
+        # Filter in the service, not here: a strict team_id comparison would drop
+        # globally-public rows owned by other teams.
+        prompts_result = await local_prompt_service.list_prompts(db, include_inactive=True, user_email=user_email, token_teams=token_teams, team_id=team_id)
         if isinstance(prompts_result, tuple):
             prompts_list = prompts_result[0]
         else:
             prompts_list = prompts_result
-
-        # Apply team filtering if specified
-        if team_id:
-            prompts_list = [p for p in prompts_list if getattr(p, "team_id", None) == team_id]
 
         # Convert to JSON-serializable format
         prompts = []
@@ -17126,6 +17120,7 @@ async def get_prompts_section(
 @admin_router.get("/sections/servers")
 @require_permission("servers.read", allow_admin_bypass=False)
 async def get_servers_section(
+    request: Request,
     team_id: Optional[str] = None,
     include_inactive: bool = False,
     db: Session = Depends(get_db),
@@ -17134,6 +17129,7 @@ async def get_servers_section(
     """Get servers data filtered by team.
 
     Args:
+        request: FastAPI request, used to derive the caller's Layer-1 visibility scope
         team_id: Optional team ID to filter by
         include_inactive: Whether to include inactive servers
         db: Database session
@@ -17144,19 +17140,16 @@ async def get_servers_section(
     """
     try:
         local_server_service = ServerService()
-        user_email = get_user_email(user)
-        LOGGER.debug(f"User {user_email} requesting servers section with team_id={team_id}, include_inactive={include_inactive}")
+        user_email, token_teams = get_scoped_resource_access_context(request, user)
+        LOGGER.debug(f"User {user_email} requesting servers section with team_id={team_id}, include_inactive={include_inactive}, token_teams={token_teams}")
 
-        # Get servers with optional include_inactive parameter
-        servers_result = await local_server_service.list_servers(db, include_inactive=include_inactive)
+        # Filter in the service, not here: a strict team_id comparison would drop
+        # globally-public rows owned by other teams.
+        servers_result = await local_server_service.list_servers(db, include_inactive=include_inactive, user_email=user_email, token_teams=token_teams, team_id=team_id)
         if isinstance(servers_result, tuple):
             servers_list = servers_result[0]
         else:
             servers_list = servers_result
-
-        # Apply team filtering if specified
-        if team_id:
-            servers_list = [s for s in servers_list if getattr(s, "team_id", None) == team_id]
 
         # Convert to JSON-serializable format
         servers = []
@@ -17186,6 +17179,7 @@ async def get_servers_section(
 @admin_router.get("/sections/gateways")
 @require_permission("gateways.read", allow_admin_bypass=False)
 async def get_gateways_section(
+    request: Request,
     team_id: Optional[str] = None,
     db: Session = Depends(get_db),
     user=Depends(get_current_user_with_permissions),
@@ -17193,6 +17187,7 @@ async def get_gateways_section(
     """Get gateways data filtered by team.
 
     Args:
+        request: FastAPI request, used to derive the caller's Layer-1 visibility scope
         team_id: Optional team ID to filter by
         db: Database session
         user: Current authenticated user context
@@ -17202,14 +17197,11 @@ async def get_gateways_section(
     """
     try:
         local_gateway_service = GatewayService()
-        get_user_email(user)
+        user_email, token_teams = get_scoped_resource_access_context(request, user)
 
-        # Get all gateways and filter by team
-        gateways_list, _ = await local_gateway_service.list_gateways(db, include_inactive=True)
-
-        # Apply team filtering if specified
-        if team_id:
-            gateways_list = [g for g in gateways_list if g.team_id == team_id]
+        # Filter in the service, not here: a strict team_id comparison would drop
+        # globally-public rows owned by other teams.
+        gateways_list, _ = await local_gateway_service.list_gateways(db, include_inactive=True, user_email=user_email, token_teams=token_teams, team_id=team_id)
 
         # Convert to JSON-serializable format
         gateways = []
--- a/mcpgateway/services/a2a_service.py
+++ b/mcpgateway/services/a2a_service.py
@@ -937,7 +937,9 @@ class A2AAgentService(BaseService):
             token_teams: Teams from JWT token. None with user_email=None = anonymous admin bypass (public+team only);
                          None with user_email set = DB admin check (public+team+own-private);
                          [] = public-only; [...] = team-scoped access.
-            team_id: Optional team ID to filter by specific team.
+            team_id: Optional team ID to filter by specific team. Applies to every caller
+                shape, including the admin and anonymous bypasses; globally-public rows
+                from other teams remain visible.
             visibility: Optional visibility filter (private, team, public).
 
         Returns:
--- a/mcpgateway/services/a2a_service.py
+++ b/mcpgateway/services/a2a_service.py
@@ -986,7 +986,7 @@ class A2AAgentService(BaseService):
         # ══════════════════════════════════════════════════════════════════════
         cache = _get_registry_cache()
         if cursor is None and user_email is None and token_teams is None and page is None:
-            filters_hash = cache.hash_filters(include_inactive=include_inactive, tags=sorted(tags) if tags else None, visibility=visibility)
+            filters_hash = cache.hash_filters(include_inactive=include_inactive, tags=sorted(tags) if tags else None, visibility=visibility, team_id=team_id)
             cached = await cache.get("agents", filters_hash)
             if cached is not None:
                 # Reconstruct A2AAgentRead objects from cached dicts
--- a/mcpgateway/services/base_service.py
+++ b/mcpgateway/services/base_service.py
@@ -93,7 +93,9 @@ class BaseService(ABC):
                 ``None`` = admin bypass (API token) or no auth context.
                 ``[]`` = public-only token.
                 ``[...]`` = team-scoped token (API or session token).
-            team_id: Optional specific team filter.
+            team_id: Optional specific team filter. When supplied, every caller
+                shape - including the admin bypasses - is narrowed to that team
+                plus globally-public rows from any team.
 
         Returns:
             Query with visibility WHERE clauses applied. Admin bypass excludes
@@ -107,6 +109,13 @@ class BaseService(ABC):
         # Matches the pattern in a2a_service._visible_agent_ids.
         model_cls = self._visibility_model_cls
         if user_email is None and token_teams is None:
+            if team_id:
+                return query.where(
+                    or_(
+                        and_(model_cls.team_id == team_id, model_cls.visibility != "private"),
+                        model_cls.visibility == "public",
+                    )
+                )
             return query.where(model_cls.visibility != "private")
 
         # Admin bypass path: only check DB for users with token_teams=None
@@ -115,6 +124,14 @@ class BaseService(ABC):
         if token_teams is None:
             user_is_admin = user_email and is_user_admin(db, user_email)
             if user_is_admin:
+                if team_id:
+                    return query.where(
+                        or_(
+                            and_(model_cls.team_id == team_id, model_cls.visibility != "private"),
+                            model_cls.visibility == "public",
+                            and_(model_cls.team_id == team_id, model_cls.visibility == "private", model_cls.owner_email == user_email),
+                        )
+                    )
                 return query.where(
                     or_(
                         model_cls.visibility != "private",
--- a/mcpgateway/services/base_service.py
+++ b/mcpgateway/services/base_service.py
@@ -110,12 +110,7 @@ class BaseService(ABC):
         model_cls = self._visibility_model_cls
         if user_email is None and token_teams is None:
             if team_id:
-                return query.where(
-                    or_(
-                        and_(model_cls.team_id == team_id, model_cls.visibility != "private"),
-                        model_cls.visibility == "public",
-                    )
-                )
+                return query.where(or_(*self._team_scoped_conditions(team_id)))
             return query.where(model_cls.visibility != "private")
 
         # Admin bypass path: only check DB for users with token_teams=None
@@ -125,13 +120,7 @@ class BaseService(ABC):
             user_is_admin = user_email and is_user_admin(db, user_email)
             if user_is_admin:
                 if team_id:
-                    return query.where(
-                        or_(
-                            and_(model_cls.team_id == team_id, model_cls.visibility != "private"),
-                            model_cls.visibility == "public",
-                            and_(model_cls.team_id == team_id, model_cls.visibility == "private", model_cls.owner_email == user_email),
-                        )
-                    )
+                    return query.where(or_(*self._team_scoped_conditions(team_id, owner_email=user_email)))
                 return query.where(
                     or_(
                         model_cls.visibility != "private",
@@ -154,6 +143,35 @@ class BaseService(ABC):
         filter_email = None if token_teams == [] else user_email
 
         return self._apply_visibility_filter(query, filter_email, effective_teams, team_id)
+
+    def _team_scoped_conditions(self, team_id: str, owner_email: Optional[str] = None) -> List[Any]:
+        """Build the OR-conditions selecting rows for an explicit ``team_id`` filter.
+
+        Single definition of what ``?team_id=`` means, shared by the admin and
+        anonymous bypass branches of :meth:`_apply_access_control` and by the
+        team-scoped branch of :meth:`_apply_visibility_filter`, so that adding a
+        visibility tier only requires updating one place.
+
+        Globally-public rows owned by *other* teams stay visible (issue #4732,
+        fixed in PR #4773): ``team_id`` narrows team-scoped rows, it does not
+        suppress platform-public ones.
+
+        Args:
+            team_id: Team to narrow to.
+            owner_email: When set, the caller's own private rows within that team
+                are included. ``None`` suppresses owner access.
+
+        Returns:
+            List of conditions intended to be combined with :func:`sqlalchemy.or_`.
+        """
+        model_cls = self._visibility_model_cls
+        conditions = [
+            and_(model_cls.team_id == team_id, model_cls.visibility.in_(["team", "public"])),
+            model_cls.visibility == "public",  # globally public items from any team are always visible
+        ]
+        if owner_email:
+            conditions.append(and_(model_cls.team_id == team_id, model_cls.visibility == "private", model_cls.owner_email == owner_email))
+        return conditions
 
     def _apply_visibility_filter(
         self,
@@ -187,13 +205,7 @@ class BaseService(ABC):
             if team_id not in token_teams:
                 return query.where(False)
 
-            access_conditions = [
-                and_(model_cls.team_id == team_id, model_cls.visibility.in_(["team", "public"])),
-                model_cls.visibility == "public",  # globally public items from any team are always visible
-            ]
-            if user_email:
-                access_conditions.append(and_(model_cls.team_id == team_id, model_cls.owner_email == user_email, model_cls.visibility == "private"))
-            return query.where(or_(*access_conditions))
+            return query.where(or_(*self._team_scoped_conditions(team_id, owner_email=user_email)))
 
         access_conditions = [model_cls.visibility == "public"]
 
--- a/mcpgateway/services/gateway_service.py
+++ b/mcpgateway/services/gateway_service.py
@@ -2367,7 +2367,9 @@ class GatewayService(BaseService):  # pylint: disable=too-many-instance-attribut
             page: Page number for page-based pagination (1-indexed). Mutually exclusive with cursor.
             per_page: Items per page for page-based pagination. Defaults to pagination_default_page_size.
             user_email: Email of user for team-based access control. None for no access control.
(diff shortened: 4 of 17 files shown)


gandhipratik203:
`tests/unit/mcpgateway/services/test_gateway_service.py` has an orphaned `@pytest.mark.asyncio` left behind where `test_admin_specific_team_is_an_exact_filter` was deleted, so `test_page_based_pagination` now carries the decorator twice. Harmless at runtime but it's dead code that shouldn't land.

Yosiefeyob:
Good catch, fixed it now. The decorator was left behind when test_admin_specific_team_is_an_exact_filter was removed; deleted it and the stray blank line. It's clean on the file now.

madhu-mohan-jaishankar (dismissed):
LGTM.

gandhipratik203 (dismissed):
LGTM! Thanks for the changes!

msureshkumar88 (changes requested):
Thanks for tackling this — the core fix (the new `or_(...)` team_id condition in `base_service.py`) is the right shape for closing the admin-bypass gap from #5496, and the new regression test in `test_team_id_admin_filter.py` is a good addition.

One thing I'd like to flag before merge, though: the registry cache layer looks like it can silently undo this fix.

## Blocking

**1. Cache key doesn't include `team_id`, so a warm cache can bypass the fix entirely**

In several services, the registry cache is read/written under exactly the admin-bypass/anonymous condition this PR patches (e.g. `cursor is None and user_email is None and token_teams is None and page is None`), but the cache key built by `cache.hash_filters(...)` never includes `team_id`:

- `mcpgateway/services/tool_service.py:2803`
- `mcpgateway/services/resource_service.py:1252`
- `mcpgateway/services/prompt_service.py:1454`
- `mcpgateway/services/a2a_service.py:988`
- `mcpgateway/services/gateway_service.py:2417` (public-only path)
- `mcpgateway/services/server_service.py:798` (public-only path)

Concretely: an admin call with no `team_id` populates the cache with the full cross-team result. A later admin call with `?team_id=alpha` hits the same cache key and gets served the unfiltered cached list — reintroducing the exact cross-team leak this PR (and #5496) is meant to close. Same issue on the public-only caching path, where a `team_id`-scoped request that should return `[]` (per `_apply_visibility_filter`) can instead return the cached full public list.

Suggest adding `team_id` into each `hash_filters(...)` call at these sites.

**2. Docstring now contradicts the fix**

`mcpgateway/services/gateway_service.py:2369` — `list_gateways` docstring still says team_id filtering "requires user_email". After this PR, the anonymous-bypass branch (`user_email is None and token_teams is None`) actively narrows by `team_id`, so the docstring is now inaccurate and could mislead API consumers about the behavior this PR just changed.

## Suggestions (non-blocking)

**3. Regression test only checks SQL substrings, not clause structure**

`tests/unit/mcpgateway/services/test_team_id_admin_filter.py:152` asserts the compiled SQL contains strings like `team_id = 'team-1'` and `visibility = 'public'`, but wouldn't catch a future regression where the `or_(...)` gets flipped to `and_(...)` — both substrings would still be present even though the query logic would be broken (public rows from other teams silently excluded). Worth asserting on the clause structure, or adding a query-execution test with seeded rows across teams/visibilities.

**4. Duplicated team_id OR-condition across 3 sites with inconsistent style**

`mcpgateway/services/base_service.py:112` — the `(team match + non-private) OR (public)` condition is duplicated near-verbatim across the anonymous-bypass and DB-admin-bypass branches, and uses `visibility != "private"` there vs. `visibility.in_(["team", "public"])` in `_apply_visibility_filter`'s equivalent branch (~line 191). A future visibility tier would need updating in three places — worth extracting into one shared helper.

No schema/Alembic changes needed here, which is right — this is a query-filter fix, not a data model change.

msureshkumar88 (changes requested):
Thanks for the follow-up commits — the cache-key fix, the extracted `_team_scoped_conditions()` helper, and the stronger row-level tests all address the earlier round of feedback cleanly, and the orphaned `@pytest.mark.asyncio` cleanup is a nice catch too.

Digging into the underlying issue a bit further (rather than just the diff), I think there's one gap left that keeps this from fully closing the bug: the fix lives in the service layer, but the Admin UI never actually calls into it correctly. That's worth resolving before merge — details below, roughly blocking vs. non-blocking.

**Blocking**

- `mcpgateway/admin.py` (`get_resources_section`, `get_prompts_section`, `get_servers_section`, `get_gateways_section`, ~lines 17036/17097/17159/17212): these don't pass `team_id` into the service's `list_*()` call at all. They fetch the full result set and then strictly re-filter in Python with `r.team_id == team_id`, which drops globally-public rows owned by other teams. That reproduces the exact "public items disappear when a team is selected in the Admin UI" behavior this PR is meant to fix — just one layer up from where the fix landed.
- `get_gateways_section` (~line 17204) additionally calls `get_user_email(user)` but discards the result, then calls `list_gateways(db, include_inactive=True)` with no `user_email`/`token_teams`. That falls into the anonymous-bypass branch and returns every team-scoped + public gateway platform-wide (before the strict Python filter above), which is broader exposure than Layer-1 token scoping intends for this endpoint.

**Suggestions**

- Stale docstring wording ("Requires user_email for access validation") was corrected in `gateway_service.py` and `server_service.py` per the last round, but the same text is still present in `tool_service.py:2745`, `resource_service.py:1197`, and `prompt_service.py:1409` — worth aligning all five for consistency.
- No test currently asserts `cache.hash_filters(...)` is called with `team_id=team_id` — the cache-poisoning fix from the earlier round (commit `0f5c436d7`) doesn't have a direct regression test; existing tests mock `hash_filters` without checking call args, so a future refactor could silently drop the kwarg.
- `test_team_id_admin_filter_rows.py`'s `ROW_SHAPES` only seeds team/public visibility rows, never a private one — the admin-owner-private-row-narrowed-to-team behavior in `_team_scoped_conditions` (`base_service.py:172-173`) has no row-level coverage.

**Minor / non-blocking**

- Registry cache fallback (in-memory, when Redis is unavailable) now keys on `team_id`, which is fine, but that fallback never proactively evicts expired entries outside of `invalidate()`/`invalidate_all()` — worth keeping an eye on under sustained varied `team_id` query traffic.
- `admin.py`'s own paginated list views re-implement team-scoped OR-conditions inline in a number of places rather than reusing the new `_team_scoped_conditions()` helper — not something this PR needs to fix, but a good follow-up so the "single definition of what `?team_id=` means" claim actually holds project-wide.

No schema/alembic changes needed here — this is a pure query-filter/cache-key change, and that's correctly reflected in the diff.

Requesting changes mainly for the two Admin UI blocking items above, since without them the underlying bug isn't fully resolved for UI users.

Yosiefeyob:
Both blocking items are in now.
One thing before this closes though — I drove the Admin UI and the symptom is still there. The UI doesn't call `/admin/sections/*`; it calls `/admin/gateways/partial?...&team_id=...`, which builds its own query with the global-public condition gated behind `include_public: bool = False` (12 partial views do this). Identical before/after this branch, so not a regression — just untouched.

Which puts the real UI path in your minor note about admin.py's inlineteam-scoped conditions rather than the blocking items.

Two questions:
1. Is `/admin/sections/*` consumed by something I'm not seeing?
2. Scope this PR as-is with a follow-up for the 12 partial views, or pull them in here?

msureshkumar88:
Both blocking items are confirmed fixed — verified directly against the commits, not just going by the summary:

- `47419b485` — admin sections now pass `team_id` into the service `list_*` calls instead of re-filtering an unfiltered result set with a strict Python comparison. `get_servers_section`/`get_gateways_section` also switched off the anonymous-bypass call shape onto `get_scoped_resource_access_context`, matching resources/prompts.
- `0f5c436d7` — `team_id` is now part of the cache key (`hash_filters`) on all five services, and the admin/anonymous/token-scoped branches share one `_team_scoped_conditions()` helper instead of three hand-copied OR-condition blocks. That dedup is exactly what would've caught the original drift.

On your two questions:

1. **Is `/admin/sections/*` consumed?** Yes — `window.Admin.reloadAllResourceSections()` in `admin_ui/initialization.js` (~line 1678) calls `/admin/sections/{section}?partial=true&team_id=...` for every section on team-switch, to refresh already-rendered content without a full page reload. So the fix here isn't dead code.

2. **12 partial endpoints (`/admin/*/partial`)** — agree these have the same class of gap. Traced it further: they build queries directly in `admin.py`, bypassing `BaseService` entirely, and their `team_id` branch only adds public-from-other-teams when `include_public=True` — grepped `admin.html` and nothing in the UI ever sets that param. So filtering by team through the main tables (not just the team-switcher) never surfaces other teams' public rows, which is stricter than the semantics this PR just established for the service layer. Confirmed pre-existing and untouched by this branch, so not a regression — matches what you found.

My preference is to keep this PR scoped as-is and file a follow-up for the 12 partial endpoints: they're a separate, hand-duplicated query path unrelated to this PR's actual root cause (`BaseService._apply_access_control()`), and pulling them in re-opens the same strict-vs-permissive question already flagged in the PR description for maintainers to weigh in on (this PR vs #6151). Happy to open that issue if useful.

Clearing my change request now that both blockers are addressed.

msureshkumar88 (approved):
Both blocking items from my earlier review are fixed (verified against commits 47419b485 and 0f5c436d7 directly): the admin sections now filter through the service layer with proper scoping, and team_id is part of the registry cache key on all five services. Clearing the change request.

The 12 hand-rolled `/admin/*/partial` endpoints have a related but separate gap (raised in review comments) — worth a follow-up issue, out of scope for this PR.

msureshkumar88:
Opened #6193 to track the 12 partial/ids/search endpoints.

msureshkumar88 (changes requested):
Two gaps from a deeper pass over this PR's own diff — both cheap, both worth closing before merge.

**1. Test coverage gap — `tests/unit/mcpgateway/test_admin.py`**

`test_get_gateways_section`, `test_get_servers_section_team_filter`(`_with_tuple_result`), `test_get_resources_section_team_filter`(`_with_tuple_result`), and `test_get_prompts_section_team_filter`(`_with_tuple_result`) all mock the service entirely and only assert on the response payload (`payload["team_id"] == team_id`, row count) — none assert `call_args` on the mocked `list_*` method.

That means the actual regression this PR guards against — dropping `user_email=`/`token_teams=` from the service call (reintroducing the anonymous-bypass leak this PR fixes in `get_servers_section`/`get_gateways_section`) or dropping `team_id=` (reintroducing #6150 itself) — would not be caught by these tests. The only net right now is at the service-layer tests (`test_team_id_admin_filter*.py`), not at the `admin.py` plumbing this PR rewrote.

Suggest adding `mock_list.assert_called_once_with(mock_db, include_inactive=..., user_email=..., token_teams=..., team_id=...)` (or equivalent) to each, so the wiring itself is asserted, not just its output.

**2. Stale docstring — `mcpgateway/services/a2a_service.py`**

`gateway_service.py`, `server_service.py`, `tool_service.py`, `resource_service.py`, and `prompt_service.py` all got their `team_id` docstring rewritten to describe the new admin/anonymous-bypass behavior ("Applies to every caller shape, including the admin and anonymous bypasses; globally-public rows remain visible"). `a2a_service.py`'s `list_agents` docstring still reads the old generic "Optional team ID to filter by specific team." The PR's own commit message says it aligns docstrings "so all five describe the same behaviour" — explicitly five-of-six, even though the PR body states `/a2a` is fixed identically to the other five endpoints. A one-line docstring update would close that gap.

Requesting changes for these two — both are gaps in code this PR itself introduced, not pre-existing/unrelated issues.

Yosiefeyob:
Both closed now the assertion added to all seven section tests (verified by mutation: dropping team_id or user_email/token_teams now fails them; both passed silently before), and a2a's docstring now matches the other five.

msureshkumar88 (approved):
Thanks for the follow-up commits — `47419b485` and `304a53e98` close every blocking and non-blocking item from the last round: the section endpoints now pass `team_id` into the service and derive scope via `get_scoped_resource_access_context()`, the strict Python re-filters are gone, the remaining `team_id` docstrings (tool/resource/prompt/a2a) are aligned, and the new `hash_filters(team_id=...)`, private-row and `assert_called_once_with(...)` tests cover the wiring rather than just the payload shape.

I re-verified this end to end against a real running gateway rather than by re-reading the diff, and everything checks out. Details below so it's reproducible.

## How it was verified

A script drives **two real gateway processes** — real `uvicorn`, real SQLite DB created by the app's own startup lifespan (schema + bootstrap admin), real JWTs from `mcpgateway.utils.create_jwt_token`, real HTTP requests. No pytest client, no mocks on the request path.

* **Phase A** — a git worktree pinned at this PR's base commit `eaf295e6`, to confirm the reported behaviour actually reproduces there.
* **Phase B** — the PR head `304a53e9`, to confirm it is fixed.

Both phases seed the identical fixture: two teams (`alpha`, `beta`), two users (platform admin, plus a non-admin member of `alpha` only), and six rows **per resource type** across tools / resources / prompts / servers / gateways / a2a:

| shape | team | visibility | owner |
|---|---|---|---|
| `x1alphateam` | alpha | team | admin |
| `x2betateam` | beta | team | admin |
| `x3betapublic` | beta | public | admin |
| `x4noteampublic` | – | public | admin |
| `x5alphaprivown` | alpha | private | admin |
| `x6alphaprivother` | alpha | private | alice |

### Config used

```
AUTH_REQUIRED=true
EMAIL_AUTH_ENABLED=true
MCPGATEWAY_UI_ENABLED=true
MCPGATEWAY_A2A_ENABLED=true
PLUGINS_ENABLED=false
DATABASE_URL=sqlite:///<per-phase temp file>
JWT_SECRET_KEY / AUTH_ENCRYPTION_SECRET=<throwaway test values>
```

### Commands run

```bash
# per phase: real server, real DB, real tokens
uvicorn mcpgateway.main:app --host 127.0.0.1 --port <46152|46153>
python -m mcpgateway.utils.create_jwt_token --username admin@example.com --admin                        # teams: null + is_admin -> admin bypass
python -m mcpgateway.utils.create_jwt_token --username admin@example.com --admin --teams <alpha>        # is_admin, Layer-1 narrowed to alpha
python -m mcpgateway.utils.create_jwt_token --username alice@example.com          --teams <alpha>        # non-admin, narrowed to alpha

curl "$BASE/{tools,resources,prompts,servers,gateways,a2a}"                    # unfiltered (also warms the registry cache)
curl "$BASE/{tools,resources,prompts,servers,gateways,a2a}?team_id=<alpha>"    # the fix
curl "$BASE/admin/sections/{resources,prompts,servers,gateways}?team_id=<alpha>"
curl "$BASE/admin/sections/{resources,prompts,servers,gateways}"               # narrowed-admin + non-admin callers

# plus the related unit tests on the head commit
pytest -q tests/unit/mcpgateway/services/test_team_id_admin_filter.py \
          tests/unit/mcpgateway/services/test_team_id_admin_filter_rows.py \
          tests/unit/mcpgateway/test_admin.py
```

## Results

**Phase A (base `eaf295e6`) reproduced the reported behaviour:**

* `/tools`, `/resources`, `/prompts`, `/servers`, `/a2a` with `?team_id=alpha` all returned **beta's team-scoped row** — issue #6150 / #5496.
* `/gateways` was already narrowed by #5929's strict filter, and correspondingly **dropped beta's globally-public row** — the #4732 regression this PR also resolves. Good illustration of the "two different answers to the same question" point in the PR description.
* All four `/admin/sections/*` endpoints with `?team_id=alpha` **dropped the other team's globally-public row** (the strict Python re-filter).
* `/admin/sections/servers` and `/admin/sections/gateways` returned beta's team row to a **team-narrowed admin token** whose Layer-1 scope was `alpha` only.

**Phase B (head `304a53e9`) — 154/154 checks passed, twice in a row.** Highlights:

* All six endpoints with `?team_id=alpha`: alpha's team row ✅, beta's public row ✅, the no-team public row ✅, the caller's own private row in alpha ✅; beta's team row ❌ excluded, another owner's private row ❌ excluded. Identical on all six, so `/gateways` and `/a2a` now answer the same question as the rest.
* **Cache-key regression:** the probe issues the unfiltered admin listing *first* on every endpoint, so the registry cache is warm before each filtered call. All six still exclude beta's team row and keep beta's public row — `0f5c436d7` holds over real HTTP.
* **Admin UI sections:** all four return 200, keep alpha's team row **and** beta's globally-public row, exclude beta's team row, and still return the documented `{<collection>, team_id}` shape.
* **Layer-1 still honoured:** a team-narrowed admin token (is_admin, `teams=[alpha]`) gets 200 on all four sections and sees alpha + public but not beta — the gap that `47419b485` closed. A non-admin caller (even with `resources.read`/`gateways.read` granted) still gets 403 from the admin gate.
* **No over-filtering:** listings without `team_id` are unchanged (beta's team row still returned; another owner's private row never returned).
* **Default-deny intact:** the non-admin narrowed to `alpha` asking for `?team_id=beta` gets nothing at all on every endpoint.
* `/health` and `/version` unaffected; the related unit tests pass on the head commit (exit 0).

## Verdict

No blocking or functionally-impacting findings left, so approving. Nothing further needed from you — the two optional notes from the last round (in-memory cache fallback only evicting via `invalidate()`, and `admin.py`'s other paginated views still hand-rolling their own team OR-conditions instead of reusing `_team_scoped_conditions()`) are fine as follow-ups rather than changes to this PR.

msureshkumar88:
Adding a breaking-change note for the record, since the behaviour shifts in **both** directions depending on the endpoint and caller. All of it is API/UI behaviour only — no schema, no Alembic migration, no config flag, no wire-format change, and both directions were observed on a real gateway (base `eaf295e6` vs head `304a53e9`) during the e2e verification above.

### 1. `?team_id=` now actually filters for admin and anonymous callers — five endpoints return *fewer* rows

`/tools`, `/resources`, `/prompts`, `/servers`, `/a2a`

Before, `_apply_access_control()` returned early on both bypass branches without ever using `team_id`, so an admin (or an unauthenticated caller when auth is relaxed) passing `?team_id=X` got every team's rows back. Now other teams' **team-scoped** rows are excluded.

Who notices: any script or dashboard that passed `?team_id=` as a decorative hint while relying on getting the full cross-team list. Those callers should drop the parameter to keep the old result set.

### 2. `/gateways?team_id=` returns *more* rows — the intentional semantic reversal

This is the one I'd call out loudest, because it widens a result set rather than narrowing it.

#5929 gave `list_gateways` a **strict** `team_id` match, which also excluded `visibility='public'` rows owned by other teams. This PR replaces that with the shared `_team_scoped_conditions()` semantics from #4773 (closing #4732): `?team_id=X` means "team X's rows **plus** all globally-public rows from any team".

Confirmed on the base commit: `/gateways?team_id=alpha` dropped beta's `visibility='public'` gateway; on the head commit it is returned. So `/gateways` moves from the strictest of the six endpoints to matching the other five.

Worth being explicit about the security framing, since "more rows" reads alarming at a glance: nothing crosses a Layer-1 boundary here. The rows newly included are `visibility='public'` — platform-public by definition, and already returned by the same endpoint when `team_id` is omitted. Strict data isolation remains `token_teams`' job, exactly as #4773 documented. But a deployment that has been treating `?team_id=` on `/gateways` as an isolation mechanism since #5929 will see its result set grow, and that assumption needs revisiting. (If maintainers prefer the strict reading instead, #6151 is the alternative — this is the fork in the road the PR description flags.)

### 3. Admin UI section endpoints change in both directions

`/admin/sections/{resources,prompts,servers,gateways}`

* **More rows:** the strict Python re-filter (`r.team_id == team_id`) is gone, so selecting a team in the UI no longer makes globally-public items from other teams disappear. That is the original #4732 complaint, and it was still reproducible on all four section endpoints at the base commit.
* **Fewer rows:** `/admin/sections/servers` and `/admin/sections/gateways` previously called their list methods with no `user_email`/`token_teams`, taking the anonymous-bypass path and returning platform-wide rows regardless of the caller's Layer-1 scope. They now derive scope via `get_scoped_resource_access_context()`. A team-narrowed admin token (`is_admin`, `teams=["alpha"]`) that used to see every team's servers and gateways in those two panels now sees only its own teams plus public — a visible content reduction for anyone using narrowed admin tokens, and the correct behaviour.

### 4. Registry cache keys changed (operational, not API)

`hash_filters(...)` now includes `team_id`, so previously-warm entries no longer match team-scoped requests. Practical effects: a brief cold-cache period after deploy, and more distinct entries under varied `?team_id=` traffic. No action needed, just worth knowing if cache-hit-rate graphs step down at rollout.

### Not affected

No DB schema or migration, no new/renamed config, no changes to response shapes or status codes, and listings **without** `team_id` are byte-for-byte unchanged (verified: other teams' rows still returned, another owner's private rows still excluded).

### Suggested rollout note

Reviewers deciding between this and #6151 are really choosing the project-wide meaning of `?team_id=`. This PR makes all six endpoints answer it the way #4773 documented; #6151 makes them all strict. Either is defensible — the important part is that `main` currently answers it two different ways, and a one-line mention in the release notes ("`?team_id=` is a UI routing hint, not an isolation boundary — use token scoping for isolation") would save a support round-trip either way.

PR #6153, opened 2026-08-10
feat(compose): monitoring portal proxy, dashboard rate fixes, pgAdmin seeding, py-spy profiling

jonpspri:
## Summary

Adds a single-entry **portal nginx** fronting all testing/monitoring services via per-service virtual hosts on one host port (default 9200), fixes Grafana dashboards that charted cumulative counters as activity, auto-provisions pgAdmin with working ContextForge connections, and adds py-spy sidecar profiling.

> The OTEL trace-propagation and auto-instrumentation changes (W3C envelope publication, affinity-path context injection, httpx/SQLAlchemy instrumentation toggles) originally in this PR moved to #6154, the session-affinity PR they belong with. This PR is now compose/monitoring infrastructure only.

## Portal reverse proxy (`infra/portal`)

- Root page enumerates services and live-probes their availability; host ports removed from HTTP services accordingly. Gateway nginx (8080), TCP databases, Tempo OTLP receivers, Keycloak, and MCP Inspector keep theirs.
- Compose profiles are now mutually exclusive (each service in exactly one profile), with Makefile targets composing them explicitly.
- `testing-up` gains `TESTING_MONITORING=1` to also activate the monitoring profile and wire gateway OTEL export to Tempo; when Tempo is already running, `testing-up` passes the same OTEL env as `monitoring-up` so the shared gateway container is not recreated with tracing disabled.
- `monitoring-lite-up`: fixed a pre-existing invalid project (`prometheus_token` lacked the `monitoring-lite` profile tag) and routed Grafana bootstrap through the portal.
- `RESILIENCE_HOST`/`BASELINE_HOST` defaults and `*-up` summary output updated for portal URLs.

## Grafana dashboards

- Postgres dashboard graphed raw monotonic counters (`tup_fetched`/`inserted`/`updated`/`deleted`/`returned`), which can only climb — timeseries panels now use `irate()` and single-stats use `SUM(irate())` retitled as activity rates, matching the existing Transactions panels.
- Cache-hit ratio panels (postgresql, gateway overview, replication) now use ratios of `rate()` so they reflect the recent window rather than the all-time average since stats reset.
- New **ContextForge OTEL Traces** dashboard (Tempo/TraceQL): recent, error, and slow traces as table panels, with cross-signal pivot hints.

## pgAdmin auto-configuration

- `infra/pgadmin/servers.json` registers "ContextForge (postgres)" and "ContextForge (via pgbouncer)", mounted via `PGADMIN_SERVER_JSON_FILE`.
[...]

PR #6154, opened 2026-08-10
feat(otel,affinity): affinity-path span instrumentation and concurrent forward dispatch

jonpspri:
## Summary

Three related changes from investigating session-affinity forward latency and trace shape:

**1. Affinity-path span instrumentation and Redis auto-instrumentation**

Deep instrumentation of the session-affinity delegation path showed that cross-worker RPC delegation costs ~85ms, while a much larger pre-handler delay traces to gunicorn workers still initializing (per-worker plugin manager builds and FastAPI route-state analysis).

- New spans: `mcp.affinity.check` (owner lookup + decision attributes), `mcp.affinity.forward_http` / `forward_rpc` (delegation calls), `mcp.affinity.dispatch_local` (same-worker dispatch), `mcp.affinity.execute_forwarded` (owner-side verify + dispatch, nested via envelope trace-context attach in the Redis listener), and a zero-duration `mcp.transport.enter` marker at transport-handler entry
- Auto-instrument redis/redis.asyncio clients (`opentelemetry-instrumentation-redis`); pub/sub receive loops are not instrumented, so listeners stay quiet
- `otel_*_instrumentation_enabled` settings now default to false; compose opts in explicitly for the observability stack
- Enabled-but-missing instrumentor packages now log at WARNING
- Fix latent test failure (`protocol=` vs `_protocol=` kwarg) exposed when the observability extra is installed

**2. Concurrent forward dispatch with per-session in-memory ordering**

The affinity Redis listener awaited each forwarded execution inline, so under load the Nth envelope for an owner worker waited ~N x execution time (observed 1-4.6s forward waits at ~500ms executions). The listener now spawns bounded dispatch tasks; measured forward wait on the slow trace cohort drops to 2-49ms.

- Per-session FIFO preserved without Redis: the listener claims an in-memory per-session lock synchronously (arrival order) and dispatch tasks take it before the global concurrency slot, so same-session forwards serialize while different sessions run concurrently (`mcpgateway_affinity_forward_concurrency`, default 32)
- Deadlock safety: session-lock acquire is bounded by `mcpgateway_affinity_session_lock_timeout` (default 30s); on timeout the forward executes without the ordering guarantee and logs a warning
- Lock lifecycle handled by `weakref.WeakValueDictionary`: entries evaporate when the last dispatch task drops its reference
- `locustfile_echo_delay`: use fast_test_server's delay fuzz (`delay_stddev`; `ECHO_DELAY_STDDEV_MS`, default 25% of `ECHO_DELAY_MS`)

**3. W3C trace-context propagation across affinity hops** (moved from #6153)

Traces for virtual-server requests appeared flat: the trusted-internal `/_internal/mcp/rpc` dispatch started a new trace because no W3C context crossed the hop, and the MCP SDK's session task group (created at app startup) never sees per-request contextvars.

- `OpenTelemetryRequestMiddleware` detects requests without a valid remote parent (explicit `is_valid` + `is_remote` check) and injects the newly created root span's context into the ASGI scope headers before invoking the app. The envelope becomes a raw inbound header for all downstream plumbing — session-task handoff, affinity envelopes, and the trusted-internal dispatch — so internal spans nest under a root span that actually exists and exports. (An earlier nginx-edge-envelope approach left every trace with a dangling parent — `<root span not yet received>` — and was reverted.)
- `post_rpc_in_process` injects the active trace context; the affinity LOCAL and http_forward paths now route through it instead of inline duplicate dispatches, and the http_forward consumer carries `traceparent`/`tracestate` explicitly past the passthrough allowlist.
- `rpc_forward` envelopes embed the trace context before signing.
- Auto-instrument outbound httpx/httpx2 clients and SQLAlchemy (`opentelemetry-instrumentation-httpx`/`-sqlalchemy`), gated by `OTEL_HTTPX_INSTRUMENTATION_ENABLED` and `OTEL_SQLALCHEMY_INSTRUMENTATION_ENABLED` (both default false; enabled in compose).

## Test plan

- Regression tests: same-session non-overlap + cross-session concurrency, lock-timeout fallback, lock GC eviction
- Trace-envelope publication, httpx/SQLAlchemy instrumentation, and affinity trace-carry unit tests
- Load-tested with the echo-delay locustfile against a multi-worker gunicorn deployment with Redis-backed affinity; forward waits measured via the new spans

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -1,9 +1,9 @@
 {
   "exclude": {
-    "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^\\.secrets\\.baseline$",
+    "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-08-07T13:52:41Z",
+  "generated_at": "2026-08-10T08:25:12Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -804,7 +804,7 @@
         "hashed_secret": "08cd923367890009657eab812753379bdb321eeb",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 948,
+        "line_number": 954,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -812,7 +812,7 @@
         "hashed_secret": "bd0160c2cf35d950843c88f3be2b9412ed71f485",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1051,
+        "line_number": 1057,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -820,7 +820,7 @@
         "hashed_secret": "3879c93c04cab8707ab8ab6a4f97d51ea8fb6f47",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1166,
+        "line_number": 1172,
         "type": "Secret Keyword",
         "verified_result": null
       },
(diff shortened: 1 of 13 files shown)


jonpspri:
Closing this fork-based PR after rebasing and pushing the branch to IBM/mcp-context-forge. A replacement PR will be opened from origin/jps-affinity-concurrency.

PR #6161, opened 2026-08-10
fix(admin-ui): attach CSRF header to import preview and configuration writes

### LINKED ISSUE #5964: [BUG]: Import/export config writes vulnerable to same missing-CSRF-header pattern as #5739
## Description

#5739 / #5780 fixed CSRF 403s on LLM Settings writes by (a) attaching `X-CSRF-Token` on every state-changing fetch and (b) not sending a malformed `Authorization: Bearer` header when `getAuthToken()` returns empty (session-cookie logins have no bearer token).

`selectiveImport.js` and `fileTransfer.js` build headers by hand the same way `llmModels.js` did before the fix, and were explicitly called out as out of scope in that PR. They have the same bug:

- `mcpgateway/admin_ui/selectiveImport.js:288-298` — `POST /admin/import/configuration`
- `mcpgateway/admin_ui/fileTransfer.js:342-352` — `POST /admin/import/preview`
- `mcpgateway/admin_ui/fileTransfer.js:398-408` — `POST /admin/import/configuration`

All three build headers manually:

```js
headers: {
  "Content-Type": "application/json",
  Authorization: `Bearer ${await getAuthToken()}`,
},
```

No `X-CSRF-Token` header is attached, and `Authorization` is set unconditionally — for session-cookie logins (`getAuthToken()` returns `""`), this sends a malformed `Bearer ` header instead of omitting the header as the fixed `llmRequestHeaders()` helper in `llmModels.js:18-32` does.

## Expected

Import/export writes under `/admin/import/*` should not 403 with CSRF validation failures under `CSRF_ENABLED=true`, matching the behavior restored for `/admin/llm/*` in #5780.

## Suggested fix

Extract `llmRequestHeaders()` (or an equivalent shared helper) out of `llmModels.js` into `utils.js` and route `selectiveImport.js` / `fileTransfer.js` POST calls through it, same pattern as the #5780 fix.

## Related

- #5739
- #5780

msureshkumar88:
## Summary

- `selectiveImport.js` and `fileTransfer.js` built request headers for the three `/admin/import/*` POST calls by hand: no `X-CSRF-Token`, and an unconditional `Authorization: Bearer ${await getAuthToken()}` that becomes a malformed `Bearer ` header for session-cookie logins (no JS-readable bearer token). Route all three through the existing `getAuthHeaders()` helper (`auth.js`), which already attaches the CSRF token and omits `Authorization` when no token is available — the same fix #5780 applied to `/admin/llm/*`.
- Collapsed the near-duplicate `llmRequestHeaders()` in `llmModels.js` into a thin delegation to the same helper, so there is one canonical implementation instead of two. All 16 internal call sites and the `{ json }` option shape are unchanged.
- This is defense in depth, not a fix for an observed 403: `admin.html`'s `fetch` wrapper already injects the CSRF header at these call sites today, and the malformed empty-bearer token is independently guarded at every backend auth layer this route touches. The change removes a latent trap (any code path outside a rendered `admin.html` template would have lost the header silently) and the duplicate implementation.

Closes #5964

## Out of scope

Same unconditional-`Authorization` pattern also exists in `teams.js` (5 sites, already CSRF-covered via `fetchWithTimeout`), `tokens.js:462`, `llmChat.js`, and two GET sites in `fileTransfer.js`. Left for a follow-up issue — none of them were reported in #5964.

## Test plan

- `make test-js` — 2332 passed, 3 pre-existing skips
- `make build-ui` — clean
- `make pre-commit` — clean, no rewrites
- Diff against `main` touches no Python


msureshkumar88:
## E2E verification

Ran `scripts/verify_import_csrf_e2e.py` against a real running instance of the gateway to confirm this branch actually fixes #5964, not just its unit tests.

**What it does:** boots the actual app (`uvicorn mcpgateway.main:app`) against a fresh, ephemeral SQLite DB, logs in as the bootstrap admin over a real cookie-jar HTTP session (no mocks, no `TestClient`), and sends the exact request shape the fixed Admin UI JS now sends — `X-CSRF-Token` from the `mcpgateway_csrf_token` cookie, no `Authorization` header — to the three call sites named in the issue. Negative controls confirm CSRF enforcement itself wasn't weakened, and a sibling check against `/admin/llm/*` (the `llmRequestHeaders` → `getAuthHeaders` delegation's own code path, and the `#5780` fix's area) confirms no collateral damage.

### Config used

```
CSRF_ENABLED=true
CSRF_COOKIE_NAME=mcpgateway_csrf_token
CSRF_TOKEN_NAME=X-CSRF-Token
SECURE_COOKIES=false        # plain-HTTP localhost; matches .env.example's non-TLS guidance
CSRF_COOKIE_SECURE=false    # same reason -- Secure-flagged cookies are dropped over http://
AUTH_REQUIRED=true
DATABASE_URL=sqlite:///<ephemeral tmpdir>/e2e.db
PLATFORM_ADMIN_PASSWORD=changeme
ADMIN_REQUIRE_PASSWORD_CHANGE_ON_BOOTSTRAP=false   # test-only: skips the forced password-change interstitial so the script can drive the real endpoints
```

### Command

```
.venv/bin/python scripts/verify_import_csrf_e2e.py
```

### Output

```
[PASS] admin login succeeds (redirect, session+CSRF cookies set) -- status=303 cookies=['jwt_token', 'mcpgateway_csrf_token']
[PASS] seed export for import payload succeeds -- status=200
[PASS] previewImport-shaped request is not CSRF-rejected -- status=200 body={"success":true,"preview":{"summary":{"total_items":0,"by_type":{"tools":0}},"items":{"tools":[]},"bundles":{},"conflicts":{},"dependencies":{}},"message":"Import preview generated. Found 0 total items..."}
[PASS] handleImport/handleSelectiveImport-shaped request is not CSRF-rejected -- status=200 body={"import_id":"4a3ad061-f0e5-4aa9-98ef-d2743cad77bf","status":"completed","progress":{"total":0,"processed":0,"created":0,"updated":0,"skipped":0,"failed":0},"errors":[],"warnings":[],...}
[PASS] previewImport WITHOUT X-CSRF-Token is still rejected (403) -- status=403
[PASS] handleImport WITHOUT X-CSRF-Token is still rejected (403) -- status=403
[PASS] sibling /admin/llm/providers/html read still reachable (no collateral damage) -- status=200
[PASS] sibling /admin/llm/* write is not CSRF-rejected (reaches app logic, not 403) -- status=404

========================================================================
8/8 checks passed
All checks passed.
```

Exit code `0`. Also ran `make ruff`/`bandit` on the script itself (clean, `# nosec`-annotated for the expected subprocess/false-positive-password findings) since it now lives under `scripts/`.

### Also confirmed

- `make test-js` — 2332 passed, 3 pre-existing skips
- `make build-ui` — clean
- `make pre-commit` — clean, no rewrites
- Full CI on this PR — green

vishu-bh (approved):
Addresses #5964 across all three affected import POST call sites while preserving CSRF enforcement and cookie-session authentication behavior. Consolidating header construction through getAuthHeaders() also removes duplicated logic without changing existing LLM call-site contracts.

No blocking code-quality, runtime, or security issues found. The added unit coverage includes both CSRF and bearer-token presence/absence paths, and CI is green.

One non-blocking follow-up worth considering: tighten the standalone E2E verifier’s positive assertions. Checks using status_code != 403 would also pass on responses such as 401 or 500; asserting expected success statuses would make failures more diagnostic.

Lang-Akshay (approved):
LGTM

PR #6162, opened 2026-08-10
fix(oauth): add tool-discovery trigger for token-exchange gateways

### LINKED ISSUE #5382: [FEATURE]: token-exchange gateways need a tool-discovery trigger after PR #5224
## Background

PR #5224 (RFC 8693 / On-Behalf-Of token exchange, closing #5135) fixed a registration-time bug: `_initialize_gateway()` unconditionally raised for `grant_type=="token-exchange"` because no end-user JWT exists at registration time, which made `POST /gateways` 502 for every token-exchange gateway (in both the sync and async-lifecycle paths — the async retry poller hit the same raise forever).

The fix mirrors the existing `authorization_code` pattern: registration-time discovery is skipped for `grant_type=="token-exchange"`, and the gateway is persisted with an empty tool list instead of failing.

## Gap

Unlike `authorization_code`, there is no later step that ever populates the tool list for a token-exchange gateway:

- `authorization_code` gets a real follow-up trigger — the OAuth callback completes consent, then gateway activation (`oauth_auto_fetch_tool_flag=True`) connects using the now-stored token.
- `token-exchange` has no equivalent. There's no consent step, and nothing else calls `_initialize_gateway`/`_initialize_gateway_with_timeout` with a usable subject token for discovery.

As things stand, a token-exchange gateway registers successfully but **never gets a tool list** through the documented API — there's no way to federate its tools without a manual workaround.

## Suggested directions (open to other approaches)

- Add an explicit admin "refresh/discover tools" action that uses the calling admin's own inbound JWT as the `subject_token` for one discovery-time exchange, just to enumerate tools/capabilities.
- Or: populate the tool list lazily on first real end-user invocation (bigger change to the federation model).

## Why this matters

Without this, token-exchange gateways are registrable but not usable for tool federation in practice — only the runtime per-invocation exchange path (already implemented and tested) works, and only for tools/servers manually pre-defined some other way.

Found and scoped while running a real e2e verification of PR #5224 against issue #5135.

msureshkumar88:
## Summary

- Registration of a `grant_type: "token-exchange"` gateway intentionally skips discovery (no per-request end-user JWT exists at registration time), but nothing else ever populated its tool list through the documented API. `refresh_gateway_manually()` (from PR #5224) already does the RFC 8693 exchange + connect + catalog sync, but only accepted the caller's JWT via an `Authorization: Bearer` header — unusable from the Admin UI, whose session JWT lives in an HttpOnly cookie the page's own JS cannot read.
- `extract_subject_jwt()` resolves the subject JWT from a bearer header first, falling back to the `jwt_token` cookie, so `GatewayService._resolve_token_exchange_header()` now accepts either.
- `POST /oauth/fetch-tools/{gateway_id}` (the Admin UI's "Fetch Tools" trigger, reached via `/gateways/{id}/tools/refresh`) now delegates to `refresh_gateway_manually()` for token-exchange gateways instead of the `authorization_code`-only legacy path. `authorization_code` handling is unchanged.
- Along the way: the Admin UI's gateway-refresh `fetch()` calls sent no `X-CSRF-Token`, so cookie-authenticated sessions were being rejected by the global CSRF middleware before this even mattered — fixed. The "🔐 Authorize" link (no consent flow exists for token-exchange) is now hidden for those gateways in the live gateway list.
- `docs/docs/architecture/oauth-design.md` documents both discovery triggers.

## Verification

Beyond the unit/integration test suite (253 Python tests across the touched modules + full `make test` green, 2323 JS tests green, `make coverage`/`diff-cover` at 93% on changed lines, `make ruff/bandit/interrogate/pylint/verify` clean, `make detect-secrets-scan` clean), a real end-to-end run proved the fix works against live infrastructure: two live gateway processes (this branch's merge-base on `main`, and this branch's HEAD), a real downstream MCP server (`ghcr.io/ibm/cfex-mcp-fast-time-server`), a real minimal RFC 8693 Authorization Server, and real minted JWTs — 16/16 checks passed, including a CSRF-still-enforced regression guard and a direct render of the actual Admin UI template proving the Authorize link is genuinely hidden for token-exchange gateways. Full config, commands, and output are posted as a PR comment.

Closes #5382

## Test plan

- [x] `make test` (full suite, 22593 items)
- [x] `make test-js` (2323 passed)
- [x] `make coverage` / `make diff-cover` (93% on changed lines)
- [x] `make ruff bandit interrogate pylint verify`
- [x] `make detect-secrets-scan`
- [x] Real end-to-end verification (BEFORE/AFTER live servers, real downstream MCP server, real AS stub) — see PR comment

### DIFF
--- a/mcpgateway/routers/oauth_router.py
+++ b/mcpgateway/routers/oauth_router.py
@@ -17,7 +17,7 @@ import json
 import logging
 import re
 import secrets
-from typing import Annotated, Any, Dict
+from typing import Annotated, Any, Dict, Optional
 from urllib.parse import urlparse
 
 # Third-Party
@@ -52,6 +52,7 @@ logger = logging.getLogger(__name__)
 
 ADMIN_CSRF_COOKIE_NAME = "mcpgateway_csrf_token"
 ADMIN_CSRF_HEADER_NAME = "x-csrf-token"
+GRANT_TYPE_TOKEN_EXCHANGE = "token-exchange"
 
 
 async def enforce_fetch_tools_csrf(request: Request) -> None:
@@ -1083,6 +1084,57 @@ async def get_oauth_status(
         raise HTTPException(status_code=500, detail="Failed to get OAuth status")
 
 
+async def _fetch_tools_via_token_exchange(gateway_id: str, gateway_service: Any, requester_email: Optional[str], request: Request) -> Dict[str, Any]:
+    """Fetch tools for a token-exchange gateway via the manual-refresh pipeline.
+
+    Token-exchange has no consent step: delegate to the manual-refresh pipeline,
+    which exchanges the caller's inbound JWT (bearer header or jwt_token cookie)
+    via ``_resolve_token_exchange_header`` (issue #5382). Exception order matters:
+    ``GatewayNotFoundError`` and ``GatewayConnectionError`` both subclass
+    ``GatewayError`` — a bare ``GatewayError`` clause first would misclassify
+    not-found and connection failures as 409 instead of 404/400.
+
+    Args:
+        gateway_id: ID of the gateway to fetch tools for.
+        gateway_service: GatewayService instance used to perform the refresh.
+        requester_email: Email of the requesting user, or None.
+        request: Incoming request, forwarded so the subject token can be resolved.
+
+    Returns:
+        Dict containing success status and message with number of tools fetched.
+
+    Raises:
+        HTTPException: If the gateway is not found (404), the connection fails
+            (re-raised for the caller to map to 400), a refresh is already in
+            progress (409), or the refresh otherwise fails (400).
+    """
+    # First-Party
+    from mcpgateway.services.gateway_service import GatewayConnectionError, GatewayError, GatewayNotFoundError  # pylint: disable=import-outside-toplevel
+
+    try:
+        refresh_result = await gateway_service.refresh_gateway_manually(
+            gateway_id=gateway_id,
+            include_resources=True,
+            include_prompts=True,
+            user_email=requester_email,
+            request_headers=dict(request.headers),
+        )
+    except GatewayNotFoundError:
+        raise HTTPException(status_code=404, detail=f"Gateway not found: {gateway_id}")
+    except GatewayConnectionError:
+        raise  # outer handler maps to 400
+    except GatewayError as ge:
+        logger.warning(f"Token-exchange tool refresh conflict for gateway {SecurityValidator.sanitize_log_message(gateway_id)}: {SecurityValidator.sanitize_log_message(str(ge))}")
+        raise HTTPException(status_code=409, detail="Refresh already in progress for this gateway")
+
+    if refresh_result.get("success") is False:
+        logger.error(f"Token-exchange tool fetch failed for gateway {SecurityValidator.sanitize_log_message(gateway_id)}: {SecurityValidator.sanitize_log_message(str(refresh_result.get('error')))}")
+        raise HTTPException(status_code=400, detail="Failed to fetch tools")
+
+    fetched = int(refresh_result.get("tools_added", 0)) + int(refresh_result.get("tools_updated", 0))
+    return {"success": True, "message": f"Successfully fetched and created {fetched} tools"}
+
+
 @oauth_router.post("/fetch-tools/{gateway_id}")
 @require_permission("gateways.update")
 async def fetch_tools_after_oauth(
@@ -1118,40 +1170,13 @@ async def fetch_tools_after_oauth(
         await _enforce_gateway_access(gateway_id, gateway, current_user, db, request=request)
 
         # First-Party
-        from mcpgateway.services.gateway_service import GatewayConnectionError, GatewayError, GatewayNotFoundError, GatewayService
+        from mcpgateway.services.gateway_service import GatewayConnectionError, GatewayService
 
         gateway_service = GatewayService()
 
         grant_type = gateway.oauth_config.get("grant_type") if isinstance(gateway.oauth_config, dict) else None
-        if grant_type == "token-exchange":
-            # Token-exchange has no consent step: delegate to the manual-refresh
-            # pipeline, which exchanges the caller's inbound JWT (bearer header or
-            # jwt_token cookie) via _resolve_token_exchange_header (issue #5382).
-            # Exception order matters: GatewayNotFoundError and GatewayConnectionError
-            # both subclass GatewayError — a bare GatewayError clause first would turn
-            # not-found into 409 and connection failures into 409 instead of 400.
-            try:
-                refresh_result = await gateway_service.refresh_gateway_manually(
-                    gateway_id=gateway_id,
-                    include_resources=True,
-                    include_prompts=True,
-                    user_email=requester_email,
-                    request_headers=dict(request.headers),
-                )
-            except GatewayNotFoundError:
-                raise HTTPException(status_code=404, detail=f"Gateway not found: {gateway_id}")
-            except GatewayConnectionError:
-                raise  # outer handler maps to 400
-            except GatewayError as ge:
-                logger.warning(f"Token-exchange tool refresh conflict for gateway {SecurityValidator.sanitize_log_message(gateway_id)}: {SecurityValidator.sanitize_log_message(str(ge))}")
-                raise HTTPException(status_code=409, detail="Refresh already in progress for this gateway")
-            if refresh_result.get("success") is False:
-                logger.error(
-                    f"Token-exchange tool fetch failed for gateway {SecurityValidator.sanitize_log_message(gateway_id)}: {SecurityValidator.sanitize_log_message(str(refresh_result.get('error')))}"
-                )
-                raise HTTPException(status_code=400, detail="Failed to fetch tools")
-            fetched = int(refresh_result.get("tools_added", 0)) + int(refresh_result.get("tools_updated", 0))
-            return {"success": True, "message": f"Successfully fetched and created {fetched} tools"}
+        if grant_type == GRANT_TYPE_TOKEN_EXCHANGE:
+            return await _fetch_tools_via_token_exchange(gateway_id, gateway_service, requester_email, request)
 
         result = await gateway_service.fetch_tools_after_oauth(db, gateway_id, requester_email)
         tools_count = len(result.get("tools", []))
--- a/mcpgateway/routers/oauth_router.py
+++ b/mcpgateway/routers/oauth_router.py
@@ -1143,7 +1143,8 @@ async def fetch_tools_after_oauth(
             except GatewayConnectionError:
                 raise  # outer handler maps to 400
             except GatewayError as ge:
-                raise HTTPException(status_code=409, detail=str(ge))
+                logger.warning(f"Token-exchange tool refresh conflict for gateway {SecurityValidator.sanitize_log_message(gateway_id)}: {SecurityValidator.sanitize_log_message(str(ge))}")
+                raise HTTPException(status_code=409, detail="Refresh already in progress for this gateway")
             if refresh_result.get("success") is False:
                 logger.error(
                     f"Token-exchange tool fetch failed for gateway {SecurityValidator.sanitize_log_message(gateway_id)}: "
--- a/mcpgateway/routers/oauth_router.py
+++ b/mcpgateway/routers/oauth_router.py
@@ -1147,8 +1147,7 @@ async def fetch_tools_after_oauth(
                 raise HTTPException(status_code=409, detail="Refresh already in progress for this gateway")
             if refresh_result.get("success") is False:
                 logger.error(
-                    f"Token-exchange tool fetch failed for gateway {SecurityValidator.sanitize_log_message(gateway_id)}: "
-                    f"{SecurityValidator.sanitize_log_message(str(refresh_result.get('error')))}"
+                    f"Token-exchange tool fetch failed for gateway {SecurityValidator.sanitize_log_message(gateway_id)}: {SecurityValidator.sanitize_log_message(str(refresh_result.get('error')))}"
                 )
                 raise HTTPException(status_code=400, detail="Failed to fetch tools")
             fetched = int(refresh_result.get("tools_added", 0)) + int(refresh_result.get("tools_updated", 0))
--- a/mcpgateway/routers/oauth_router.py
+++ b/mcpgateway/routers/oauth_router.py
@@ -1084,7 +1084,16 @@ async def get_oauth_status(
         raise HTTPException(status_code=500, detail="Failed to get OAuth status")
 
 
-async def _fetch_tools_via_token_exchange(gateway_id: str, gateway_service: Any, requester_email: Optional[str], request: Request) -> Dict[str, Any]:
+async def _fetch_tools_via_token_exchange(
+    gateway_id: str,
+    gateway_service: Any,
+    requester_email: Optional[str],
+    request: Request,
+    *,
+    gateway_not_found_error: type,
+    gateway_connection_error: type,
+    gateway_error: type,
+) -> Dict[str, Any]:
     """Fetch tools for a token-exchange gateway via the manual-refresh pipeline.
 
     Token-exchange has no consent step: delegate to the manual-refresh pipeline,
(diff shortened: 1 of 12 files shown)


msureshkumar88:
## Real end-to-end verification

Ran a real (non-mocked) end-to-end check that this PR actually fixes #5382, rather than relying on unit tests alone. Two live gateway server processes were started — one at this branch's merge-base on `main`, one at this branch's `HEAD` — against a real downstream MCP server and a real (minimal) RFC 8693 Authorization Server, with real minted JWTs and real HTTP requests (`requests` library) carrying real cookies/CSRF headers.

### Config used

| Component | Detail |
|---|---|
| **BEFORE gateway** | `main` @ `eaf295e64457d6f3354cb875916c927927973b91` (checked out in an isolated `git worktree`, shares the repo's `.venv`) |
| **AFTER gateway** | this branch @ HEAD, via `PYTHONPATH` shadowing the same shared `.venv` |
| **Downstream MCP server** | real container, `ghcr.io/ibm/cfex-mcp-fast-time-server@sha256:e676ab998fbf0aa03adf07a754793effe7b3bdd01d217f2f3cd40c0e91817881` (the same image used by this repo's own `docker-compose.yml` for `fast_time_server` / MCP-protocol e2e coverage) — exposes 7 real tools over Streamable HTTP |
| **RFC 8693 Authorization Server** | a small standalone `http.server` process implementing `POST /token`, validating `grant_type=urn:ietf:params:oauth:grant-type:token-exchange` and a non-empty `subject_token`, returning a real `access_token` JSON |
| **Auth** | admin JWT minted via the documented CLI: `python -m mcpgateway.utils.create_jwt_token --username admin@e2e.test --admin --exp 60 --secret <secret>` |
| **DB** | separate on-disk SQLite files per gateway instance (not `:memory:`) |
| **Env (both instances, explicit — not inherited from any developer `.env`)** | `AUTH_REQUIRED=true`, `CSRF_ENABLED=true`, `MCP_REQUIRE_AUTH=false` (matches `.env.example` dev default), `SSRF_ALLOW_LOCALHOST=true` / `SSRF_ALLOW_PRIVATE_NETWORKS=true` / `SSRF_DNS_FAIL_CLOSED=false` (the downstream server and AS are loopback by construction of this harness) |

The core probe replicates what the Admin UI's `fetchToolsForGateway()` actually sends (`mcpgateway/admin_ui/auth.js`): `credentials:"include"` (the HttpOnly `jwt_token` session cookie, no `Authorization` header — matching that the cookie can't be read by JS), an `X-CSRF-Token` header with a matching double-submit cookie, and a `Referer` of `.../admin` (the same-origin browser signal `middleware/rbac.py`'s `require_auth()` requires before it will accept cookie-only authentication on a non-`/admin` route).

### Commands run

```bash
# BEFORE worktree
git worktree add <tmp>/before-main main   # = eaf295e64, this branch's merge-base

# real downstream MCP server
docker run -d --rm --name fast-time -p <port>:9080 \
  -e BIND_ADDRESS=0.0.0.0:9080 \
  ghcr.io/ibm/cfex-mcp-fast-time-server:latest

# real stub Authorization Server (RFC 8693 /token)
python3 stub_as.py <port>

# real gateway processes (BEFORE against the worktree, AFTER against this branch)
PYTHONPATH=<source_dir> DATABASE_URL=sqlite:///<db>.db JWT_SECRET_KEY=... AUTH_REQUIRED=true \
  CSRF_ENABLED=true SSRF_ALLOW_LOCALHOST=true SSRF_ALLOW_PRIVATE_NETWORKS=true SSRF_DNS_FAIL_CLOSED=false \
  .venv/bin/python -m uvicorn mcpgateway.main:app --host 127.0.0.1 --port <port>

# mint a real admin JWT
.venv/bin/python -m mcpgateway.utils.create_jwt_token --username admin@e2e.test --admin --exp 60 --secret <secret>

# register a token-exchange gateway (bearer auth, real POST /gateways)
POST /gateways  {"grant_type": "token-exchange", "token_url": "<AS>/token", ...}

# THE ACTUAL PROBE — cookie-only, Admin-UI-style request
POST /oauth/fetch-tools/{gateway_id}
  Cookie: jwt_token=<jwt>; mcpgateway_csrf_token=<csrf>
  X-CSRF-Token: <csrf>
  Referer: <base_url>/admin
```

Full script: `.venv/bin/python verify_5382.py` (16 automated checks, self-contained, cleans up its own containers/processes/DB files on exit).

### Output

```
======================================================================
E2E verification — issue #5382 (token-exchange tool discovery)
======================================================================
[PASS] stub RFC 8693 Authorization Server started
[PASS] real downstream MCP server (fast-time-server image) started
[PASS] BEFORE gateway (main @ eaf295e64) started
[PASS] AFTER gateway (fix/5382-token-exchange-tool-discovery @ HEAD) started
[PASS] minted real admin JWT via create_jwt_token CLI
[PASS] BEFORE: register token-exchange gateway (no 502, empty tool list at registration)
[PASS] BEFORE: cookie-only fetch-tools on token-exchange gateway FAILS (reproduces issue #5382)
       — HTTP 400: {"detail":"Failed to fetch tools"}
[PASS] AFTER: register token-exchange gateway (no 502, empty tool list at registration)
[PASS] AFTER: cookie-only fetch-tools on token-exchange gateway SUCCEEDS (issue #5382 fixed)
       — HTTP 200: {"success":true,"message":"Successfully fetched and created 7 tools"}
[PASS] AFTER: authorization_code gateway still registers cleanly — HTTP 200
[PASS] AFTER: authorization_code fetch-tools still fails without completed OAuth (unchanged path)
       — HTTP 400: {"detail":"Failed to fetch tools"}
[PASS] AFTER: register second token-exchange gateway for bearer-path probe — HTTP 200
[PASS] AFTER: bearer-header fetch-tools on token-exchange gateway also succeeds
       — HTTP 200: {"success":true,"message":"Successfully fetched and created 7 tools"}
[PASS] AFTER: plain (non-OAuth) gateway still auto-discovers tools at registration
       — HTTP 200, discovered_tools=7
[PASS] AFTER: GET /gateways (adjacent listing endpoint) unaffected — HTTP 200, count=4
[PASS] AFTER: mismatched CSRF token on cookie-only fetch-tools is still rejected (403)
       — HTTP 403

======================================================================
SUMMARY: 16/16 checks passed
======================================================================
```

**What each check proves:**
- The **BEFORE→AFTER pair** (checks 6–9) is the core reproduction: the identical cookie-only request that reproduces #5382 on `main` (400) succeeds on this branch (200, real tools discovered via a real RFC 8693 exchange against the stub AS, then a real MCP `tools/list` against the fast-time-server container).
- **authorization_code unaffected** (checks 10–11): that grant type's discovery path is untouched — still requires completed OAuth, still returns 400 without it.
- **Bearer-header path** (checks 12–13): Task 3's router delegation works independently of the cookie fallback (Task 1/2), and doesn't regress the header-based flow.
- **Plain/non-OAuth gateways unaffected** (check 14), **adjacent `GET /gateways` unaffected** (check 15).
- **CSRF still enforced** (check 16) — a security regression guard: the fix does not weaken CSRF protection on the newly-working cookie path; a mismatched token is still rejected with 403.

### Additional check: the actual rendered Admin UI

The above exercises the HTTP API directly. To also verify the *rendered* Admin UI (the surface a user actually sees), the app's real Jinja environment (`mcpgateway.main.jinja_env`) rendered its real, live `gateways_partial.html` template against real `GatewayRead`-serialized data (fetched from a running instance) for both a token-exchange and an authorization_code gateway:

```
token-exchange gateway:      Authorize link present = False
authorization_code gateway:  Authorize link present = True
```

Confirms the "🔐 Authorize" link (no consent flow exists for token-exchange) is genuinely hidden in the gateway list users see, not just in the HTTP response.

mcpgateway/utils/subject_token.py
```diff
@@ -50,3 +51,44 @@ def looks_like_jwt(token: Optional[str]) -> bool:
...
+    for k, v in request_headers.items():
+        if k.lower() == "cookie" and isinstance(v, str):
+            raw_cookie = v
+            break
+    if not raw_cookie:
+        return None
+    jar = SimpleCookie()
+    try:
+        jar.load(raw_cookie)
+    except (CookieError, AttributeError, TypeError):
```
gcgoncalves: Suggestion: Log the attribute/type errors

```python
except CookieError:
    return None
except (AttributeError, TypeError) as e:
    # Log unexpected errors for debugging
    logger.debug(f"Unexpected cookie parsing error: {type(e).__name__}")
    return None
```

mcpgateway/routers/oauth_router.py
```diff
@@ -1118,9 +1118,41 @@ async def fetch_tools_after_oauth(
         await _enforce_gateway_access(gateway_id, gateway, current_user, db, request=request)
 
         # First-Party
-        from mcpgateway.services.gateway_service import GatewayConnectionError, GatewayService
+        from mcpgateway.services.gateway_service import GatewayConnectionError, GatewayError, GatewayNotFoundError, GatewayService
 
         gateway_service = GatewayService()
+
+        grant_type = gateway.oauth_config.get("grant_type") if isinstance(gateway.oauth_config, dict) else None
+        if grant_type == "token-exchange":
```
gcgoncalves: NIT: Replace this magic string with a const.

mcpgateway/routers/oauth_router.py
```diff
@@ -1118,9 +1118,41 @@ async def fetch_tools_after_oauth(
...
+                logger.warning(f"Token-exchange tool refresh conflict for gateway {SecurityValidator.sanitize_log_message(gateway_id)}: {SecurityValidator.sanitize_log_message(str(ge))}")
+                raise HTTPException(status_code=409, detail="Refresh already in progress for this gateway")
+            if refresh_result.get("success") is False:
+                logger.error(
+                    f"Token-exchange tool fetch failed for gateway {SecurityValidator.sanitize_log_message(gateway_id)}: {SecurityValidator.sanitize_log_message(str(refresh_result.get('error')))}"
+                )
+                raise HTTPException(status_code=400, detail="Failed to fetch tools")
+            fetched = int(refresh_result.get("tools_added", 0)) + int(refresh_result.get("tools_updated", 0))
+            return {"success": True, "message": f"Successfully fetched and created {fetched} tools"}
+
```
gcgoncalves: I suggest extracting this complex logic on a separate function for readability and maintainability.

gcgoncalves (changes requested):

mcpgateway/utils/subject_token.py
msureshkumar88: Done — split into `except CookieError` / `except (AttributeError, TypeError) as e` with a `logger.debug()` call, matching your snippet. Pushed in 786a0d660's follow-up commit.

mcpgateway/routers/oauth_router.py
msureshkumar88: Done — added `GRANT_TYPE_TOKEN_EXCHANGE = "token-exchange"` module constant and swapped the literal at the check site.

mcpgateway/routers/oauth_router.py
msureshkumar88: Done — extracted the branch into `_fetch_tools_via_token_exchange()` with its own docstring; `fetch_tools_after_oauth` now just calls it and returns.

prakhar-singh1928:
@msureshkumar88 few issues i found int he PR
**F-1 — Double-import fragility in exception propagation (oauth_router.py)**

_fetch_tools_via_token_exchange() imports GatewayConnectionError independently from the parent fetch_tools_after_oauth(). When the helper raises bare and the outer handler catches it, both class references must be the same object. Today they are (same module), but having two separate from mcpgateway.services.gateway_service import GatewayConnectionError at different scope levels is fragile. The fix: let the helper rely on the import the caller already ran (pass it as a parameter, or ensure the single from ... import inside the try block runs before the branch decision and covers both scopes).

**F-2 — Expired cookie JWT silently triggers negative cache (subject_token.py)**

extract_subject_jwt() checks structural shape but not exp. An expired jwt_token cookie passes the check, gets forwarded to the AS as subject_token, the AS rejects it, and set_failure() records the gateway as exchange-unavailable for that user+audience until the negative-cache TTL drains. This means a session expiry in the Admin UI can block all subsequent valid sessions for the same user+gateway combination. At minimum, document this interaction in a code comment near _fetch_tools_via_token_exchange so a future engineer bumping the negative-cache TTL understands the blast radius.

**F-3 — Missing test case: opaque bearer + absent jwt_token cookie (test_subject_token.py)**

The test matrix covers opaque bearer + valid cookie (→ cookie JWT), but not opaque bearer + no matching cookie (→ None). This is a straightforward gap for a function whose None return triggers a GatewayConnectionError downstream.

prakhar-singh1928 (dismissed):
LGTM

msureshkumar88:
@prakhar-singh1928 thanks, addressed all three:

**F-1** — `_fetch_tools_via_token_exchange()` no longer imports `GatewayConnectionError`/`GatewayError`/`GatewayNotFoundError` itself. `fetch_tools_after_oauth()` now does the single `from mcpgateway.services.gateway_service import ...` and passes those three classes into the helper as keyword params, so both scopes are guaranteed to be looking at the same objects rather than relying on Python's import cache to keep two independent imports in sync.

**F-2** — Added a "Blast radius" paragraph to `_fetch_tools_via_token_exchange()`'s docstring documenting that `extract_subject_jwt()` is shape-only (no `exp` check), so an expired `jwt_token` cookie passes the check, gets rejected by the AS, and trips `set_failure()`'s negative cache for that `(gateway_id, user, audience)` key — blocking that user+gateway combo until the TTL drains, even after a fresh login. Flagged for whoever next touches the negative-cache TTL.

**F-3** — Added `test_opaque_bearer_and_no_jwt_token_cookie_returns_none` covering the opaque-bearer + absent-jwt_token-cookie → `None` case.

Pushed. All unit tests + diff-cover (100% on changed lines) green.

gcgoncalves:
Approving once conflicts resolved.

gcgoncalves (dismissed):

Lang-Akshay (changes requested):
## Blocking Changes

| 1 | Security | `mcpgateway/utils/subject_token.py` | 76–99 (called from `mcpgateway/services/gateway_service.py:956`) | `extract_subject_jwt()` falls back to the `jwt_token` cookie whenever an `Authorization: Bearer` value is present but is not JWT-shaped. ContextForge may authenticate the bearer credential for principal A, while the cookie JWT for principal B is then forwarded as the RFC 8693 subject token. The downstream exchange is not bound to the authenticated/authorized principal (CWE-287/CWE-346). | If any bearer authorization is present, do not fall back to a cookie after rejecting it; fail closed or use the exact bearer credential already validated by the auth dependency. Bind the exchange subject to the authenticated request principal. Add a mixed opaque-bearer plus cookie deny-path regression in the project’s tests (test contents remain excluded from this review). |

msureshkumar88:
Thanks for flagging this, @Lang-Akshay — good catch, and appreciate the precise CWE references.

Fixed in 652ffcd67: `extract_subject_jwt()` (`mcpgateway/utils/subject_token.py`) no longer falls back to the `jwt_token` cookie once any `Authorization: Bearer` credential is present. Previously a non-JWT-shaped bearer (already authenticated by the auth dependency as some principal) would fall through to the cookie, which could belong to a different principal — the exchange wasn't bound to the authenticated caller. Now: if a bearer credential is present at all, it's the only candidate considered for the subject token; a non-JWT-shaped bearer fails closed (returns `None` → `_resolve_token_exchange_header()` raises `GatewayConnectionError`) instead of silently substituting the cookie.

Added regression tests in `tests/unit/mcpgateway/utils/test_subject_token.py`:
- `test_opaque_bearer_does_not_fall_back_to_cookie` — opaque bearer + same-shaped cookie JWT now denies (previously fell through).
- `test_opaque_bearer_does_not_fall_back_to_other_principal_cookie` — explicit mixed-principal deny path (bearer for principal A, cookie JWT for principal B).

Also updated `docs/docs/architecture/oauth-design.md` to describe the fail-closed behavior accurately.

Full targeted suite (`test_subject_token.py`, `test_oauth_router.py`, `test_gateway_service.py`, `test_token_exchange_integration.py`) green, ruff clean.

Lang-Akshay (dismissed):
LGTM ✅

gcgoncalves (dismissed):

Lang-Akshay (approved):
LGTM

Issue #6104, opened 2026-08-06
[TASK]: Gateway Performance Tracking — Nightly Capture & Dashboard


Lang-Akshay:
## Overview

Nightly load tests already run against the gateway (`make load-test` and `make benchmark-mcp-tools`) via a workflow in a separate repo. Results live only in that run's logs — nothing is persisted, nothing is compared across runs or releases. This epic wires a capture layer onto those existing outputs and surfaces the trends as an accessible dashboard.

## Goal

Interface with the existing nightly test runs. Capture output into a persistent store. Present gateway performance trends per release and per nightly run through an accessible dashboard hosted on an internal Fyre VM.

## Sub-tasks

- [ ] #6105 — Phase 1: Nightly Load-Test Data Capture
- [ ] #6106 — Phase 2: Performance Dashboard
- [ ] #6107 — Phase 3: Plugin Load Tests — RateLimiter, SecretsDetection, SQLSanitizer, OutputLengthGuard

### Phase 1 — Data Capture (#6105)

Capture the output of nightly `load-test` and `benchmark-mcp-tools` runs into a persistent store. The store should be queryable and accumulate results across runs without overwriting prior data. Each record should carry at minimum: timestamp, git sha/tag, suite name, concurrency settings, p50/p90/p95/p99 latency, RPS, error rate, and the host class the test ran on.

### Phase 2 — Dashboard (#6106)

Surface the captured data as a team-accessible internal dashboard hosted on a Fyre VM. Should show gateway performance trends per release and per nightly run, support filtering by suite and date range, and visually distinguish tagged releases from nightly runs. Tech stack to be decided with the UI team before implementation begins.

### Phase 3 — Plugin Load Tests (#6107)

Extend the nightly capture to four plugins: `RateLimiterPlugin`, `SecretsDetection`, `SQLSanitizer`, `OutputLengthGuardPlugin`. Each plugin should be exercised with the plugin enabled and disabled so overhead delta is visible over time.

## Out of Scope (MVP)

- Modifying the external nightly workflow beyond the minimum needed to invoke capture and store results
- Alerting or regression gating in CI (follow-up)
- All other plugins — adding one should be straightforward once the pattern is established

## Open Questions

1. Where should the persistent store live? See Phase 1 for discussion.
2. Dashboard tech stack and hosting model — to be decided with the UI team in Phase 2.

PR #6164, opened 2026-08-10
feat(otel,affinity): affinity-path span instrumentation and concurrent forward dispatch

### LINKED ISSUE #6284: [PERF]: Affinity-path span instrumentation and concurrent forward dispatch
Add affinity-path spans for owner checks, HTTP/RPC forwarding, local dispatch, and forwarded execution; propagate W3C trace context across session-affinity hops; add Redis/HTTPX/SQLAlchemy auto-instrumentation; dispatch forwarded affinity work concurrently with bounded global concurrency while preserving per-session FIFO ordering. Tracked by PR #6164.

jonpspri:
## Summary

Supersedes #6154. This PR is the rebased origin-branch counterpart and contains the session-affinity, OTEL, and dependency changes:

- Add affinity-path spans for owner checks, HTTP/RPC forwarding, local dispatch, and forwarded execution.
- Propagate W3C trace context across session-affinity hops, including signed RPC envelopes and trusted internal dispatch.
- Add Redis, HTTPX, and SQLAlchemy auto-instrumentation behind explicit settings and compose opt-ins.
- Dispatch forwarded affinity work concurrently with bounded global concurrency while preserving per-session FIFO ordering, timeout fallback, and weak lock lifecycle management.
- Add the related concurrency, trace-envelope, instrumentation, and lock-lifecycle regression coverage.
- Bump `h2` to 4.4.1 for CVE-2026-71554 and prune expired `exclude-newer` overrides.

The portal, Grafana, pgAdmin, and py-spy compose infrastructure remains in the companion monitoring PR, superseding the material originally moved here from #6153.

---

Closes #6284 — affinity-path span instrumentation and concurrent forward dispatch (performance epic #6104).


gandhipratik203:
Thanks for this one, the trace propagation across the three hop types is clean and well commented. A few findings:

**Duplicate owner lookup**

`streamablehttp_transport.py:4311` and `:4315` both call `get_session_owner()`. Two Redis round-trips per request instead of one, in a PR about cutting affinity latency. The span also reports the first read while routing uses the second.

**Tautological span tests**

`test_streamablehttp_transport.py:17980-18052`: the three `test_affinity_span_attribute_logic_*` tests copy the production block into the test body and assert against the copy. No production code runs, which is why the duplicate lookup slipped through. Drive the real handler with a stubbed `pool` instead.

**Unbounded dispatch queue**

`session_affinity.py:991-1000`: the listener spawns one task per message with no cap. The semaphore limits execution to 32 but not queue depth, so bursts pile up with no backpressure. Dropping envelopes older than `mcpgateway_pool_rpc_forward_timeout` would cap it.

Once these are addressed we're good to go.

gandhipratik203 (dismissed):
Thanks for the changes, LGTM!

gandhipratik203 (dismissed):
Looks good to me!

msureshkumar88 (changes requested):
Nice work here — the bounded-concurrency + per-session FIFO design for forwarded dispatch is a real improvement over the old strictly-serial listener, and the span instrumentation is exactly what's needed to actually see where affinity-path latency goes. A couple of things below should get fixed before merge, verified against the current diff.

## Blocking

**1. `drain_all()` now cancels in-flight forwarded requests, but it's wired to SIGHUP (routine cert rotation), not shutdown.**

`mcpgateway/services/session_affinity.py:704-717` — `drain_all()` cancels every task in `self._forward_tasks`. `sighup_reload()` in `mcpgateway/handlers/signal_handlers.py` calls this on every SIGHUP for TLS cert rotation. Before this PR that was a genuine no-op; now a routine cert rotation under load will cancel in-flight cross-worker tool calls, and the originating worker will see `mcpgateway_pool_rpc_forward_timeout` fire on a request that would otherwise have succeeded.

The docstring and log line ("no worker-local state to clear") still describe the old no-op behavior, and the test suite currently has both `test_drain_all_is_logging_only_noop` and `test_drain_all_cancels_inflight_forward_tasks` passing on the same method — worth reconciling.

Suggestion: move the forward-task cancel-and-await into `close_all()` (true shutdown), and either leave `drain_all()` as the documented no-op or give it its own reviewed, tested, documented behavior if SIGHUP really should abort in-flight forwards.

**2. `mcp.affinity.check` span wraps zero real work.**

`mcpgateway/transports/streamablehttp_transport.py:4521-4522` — `owner = await pool.get_session_owner(mcp_session_id)` completes *before* the `mcp.affinity.check` span opens, so the span only sets attributes on an already-known value. The actual Redis round-trip — the thing this instrumentation exists to measure — happens outside the span. As written, this span will report near-zero duration regardless of real affinity-lookup latency, which undercuts the PR's own stated diagnostic goal.

Suggestion: open the span before the `get_session_owner` call so it actually wraps the lookup.

## Worth resolving before merge (may be a quick fix or may need a design call)

**3. Trace-context injection may clobber the request's real traceparent inside `call_tool`.**

`mcpgateway/transports/streamablehttp_transport.py:1902` opens `mcp.affinity.forward_rpc` and calls `forward_request_to_owner`, which injects the *ambient* OTel context via `inject_trace_context_headers` (`mcpgateway/observability.py:752-789`, which unconditionally overwrites `carrier["traceparent"]`). This runs inside the SDK handler task that this same file documents (lines ~4720-4723) as not inheriting the request's ContextVars — the exact task-boundary problem `enriched_headers`/`inject_trace_context_headers` at line 4724 was already added to work around for the header-based context. If the ambient OTel span context in that handler task is stale/startup rather than the current request's, this injection could overwrite the already-correct traceparent for the `tools/call` forwarding path specifically — the opposite of what the PR is trying to achieve.

I traced this statically and believe the mechanism is real, but haven't confirmed at runtime what the ambient span context actually is in that handler task. Could you double check (or point me to where I'm wrong) — a test asserting the forwarded span's trace ID matches the inbound request's trace ID would settle it either way.

## Also flagging (non-blocking, would be good to fold in)

- `forward_request_to_owner`'s `rpc_forward` envelope never sets `timestamp` (only `http_forward` does, session_affinity.py:1511), so the stale-envelope drop at `_dispatch_forwarded` (line 1080) is a no-op for the RPC/SSE path. Two-line fix plus a test against the actual envelope producer rather than a hand-built dict.
- `mcpgateway_affinity_forward_concurrency` has no lower-bound validator — `0` silently wedges all forwarding on that worker with no startup error (`asyncio.Semaphore(0)`), negative raises at runtime inside the listener instead of failing config validation. A `ge=1` would catch both.
- `close_all()` cancels the listener/heartbeat tasks but never touches `self._forward_tasks` — worth cancelling those on shutdown too, for the same reason as point 1 above but on the graceful-shutdown path.

Happy to help push on any of these if useful. Great instrumentation work overall — just want the SIGHUP and span-placement issues sorted before this goes in.

jonpspri:
Thanks @msureshkumar88 — this was a sharp read, especially the SIGHUP wiring catch. All addressed in bdaf0217, details below.

## Blocking

**1. `drain_all()` cancelling in-flight forwards on SIGHUP** — agreed, that was a real behavioral regression hiding behind an outdated docstring. Fixed your way: `drain_all()` is back to the documented log-only no-op (its docstring now spells out *why* SIGHUP must not abort forwards), and the cancel-and-await moved into `close_all()`, which is what the lifespan shutdown path actually calls. The contradictory tests are reconciled — the cancellation test now targets `close_all()`, and there's a new `test_drain_all_preserves_inflight_forward_tasks` pinning the SIGHUP-safety contract so this doesn't silently regress again.

**2. `mcp.affinity.check` wrapping zero work** — good spot. The span now opens before `get_session_owner()` and wraps the Redis round-trip, which is the latency it exists to measure.

## Finding #3 — confirmed valid, and slightly worse than you traced it

You were right to flag it. The SDK's session tasks inherit the contextvars snapshot of whichever request *spawned* the session task — i.e. the `initialize` request — not the current one. That stale span is ended, but its span context is still *valid* (`is_valid` only checks non-zero IDs), so `otel_context_active()` returns true and `otel_inject()` would overwrite the correct edge-injected traceparent with the initialize request's trace ID. Every forwarded `tools/call` on a long-lived session would have landed in the wrong trace.

Fix: an existing `traceparent` in the envelope headers now always wins; ambient injection is only the fallback when headers carry none. Your suggested test is in: `test_forward_request_to_owner_preserves_edge_traceparent_over_stale_ambient` drives the real `forward_request_to_owner` with a stale ambient span active and asserts the published envelope's traceparent matches the inbound request's. There's a complementary test for the fallback path (no edge traceparent → ambient injection still applies).

One honest residual: the local `mcp.affinity.forward_rpc` span at the call site is still parented on that possibly-stale ambient context, so the span *itself* can land in the wrong trace even though the forwarded work now lands in the right one. Properly fixing that means teaching `create_span` to take an explicit extracted context — I'd rather do that as a focused follow-up than grow this PR. Happy to hear arguments either way.

## Fold-ins — all three taken

- `rpc_forward` envelopes now set `timestamp` before signing, so the stale-envelope drop in `_dispatch_forwarded` is live for the RPC/SSE path. The test drives the real producer (not a hand-built dict) and verifies the envelope HMAC still validates with the new field.
- `mcpgateway_affinity_forward_concurrency` now has `ge=1` — `0` fails config validation at startup instead of silently wedging forwarding.
- `close_all()` cancels and awaits `_forward_tasks` (fell out of the blocking-1 fix, as you predicted).

Full targeted run green: 978 passed across the affinity, transport, observability, db, and sighup suites; ruff clean on the touched files (also swept up a missing `typing.Any` import and some format drift this PR had introduced in the affinity test module).

msureshkumar88 (approved):
Re-reviewed against the current commits (`94112324d`, `bdaf02172`), not against the earlier review-comment prose. All three blocking items and the non-blocking items from the prior pass check out in the code itself:

- `drain_all()` (SIGHUP) is back to a documented no-op; the forward-task cancel-and-await moved to `close_all()` — verified at `session_affinity.py:702-732`.
- `mcp.affinity.check` now wraps the actual `get_session_owner()` Redis lookup — `streamablehttp_transport.py:4518-4524`.
- The ambient-context trace clobbering concern in `call_tool`'s forward path is fixed with an "existing traceparent wins" guard (`session_affinity.py:930-940`), and I traced the full call chain to confirm the ambient context is correct at every `post_rpc_in_process` call site that relies on it implicitly.
- `rpc_forward` envelopes now set `timestamp` (`:948`), and `mcpgateway_affinity_forward_concurrency` got a `ge=1` bound (`config.py:2831`).

Ran the four touched unit-test files locally (`test_session_affinity.py`, `test_streamablehttp_transport.py`, `test_observability.py`, `test_db.py`) — all pass — plus `ruff check` on the six changed source files — clean. No Alembic/schema change, none needed. Backend-only, no UI surface.

**Approving with one documented waiver**: this repo's contributor guide requires a full black-box test against a running gateway for any PR whose behavior is exercisable through a live gateway. The new concurrent-forward-dispatch path (bounded semaphore + per-session FIFO locks) is currently covered only at the unit level with a mocked pool/Redis — there's no multi-instance test exercising real cross-worker forwarding. Given the change is additive, opt-in by default, and thoroughly unit-tested, I'm not blocking merge on it, but I've filed #6567 to track adding that live-gateway test (modeled on `tests/live_gateway/run_primary_worker_multiinstance.sh`), and it should land before this path sees significant production traffic.

Two more non-blocking items worth a quick look, not required for merge:
- `.env.example` doesn't yet document the 5 new settings this PR adds (`MCPGATEWAY_AFFINITY_FORWARD_CONCURRENCY`, `MCPGATEWAY_AFFINITY_SESSION_LOCK_TIMEOUT`, `OTEL_HTTPX_INSTRUMENTATION_ENABLED`, `OTEL_SQLALCHEMY_INSTRUMENTATION_ENABLED`, `OTEL_REDIS_INSTRUMENTATION_ENABLED`), unlike the rest of the `OTEL_*`/`AFFINITY_*` surface which is documented there.
- `docker-compose.yml`'s `GUNICORN_WORKERS` default drops from 24→17 on the unprofiled `gateway` service (affects every `docker compose up`, not just the observability profile). The profiling rationale in the commit message is reasonable, but it's worth a one-line mention in the PR description so it doesn't get missed as a capacity-affecting default change riding along with the instrumentation work.

Nice work on the concurrency design — the `WeakValueDictionary`-backed per-session locks are a clean way to get FIFO ordering without manual refcounting, and the span placement fixes make this instrumentation actually measure what it claims to measure.

PR #6168, opened 2026-08-10
feat(oauth): per-gateway use_dcr flag for Dynamic Client Registration

### LINKED ISSUE #6167: [FEATURE]: Per-gateway use_dcr flag for Dynamic Client Registration
## Problem

ContextForge's Dynamic Client Registration (DCR, RFC 7591) is controlled by two **deployment-global** environment variables:

- `MCPGATEWAY_DCR_ENABLED` (default `true`)
- `MCPGATEWAY_DCR_AUTO_REGISTER_ON_MISSING_CREDENTIALS` (default `true`)

The trigger in the authorize flow (`oauth_router.py` Phase 1.4) registers a gateway client whenever the gateway has an `issuer` but no `client_id` — for **every** gateway or **none**. There is no way to express per-gateway intent.

This breaks real fleet scenarios:

1. A deployment disables global auto-register as a security posture, but specific gateways (whose Authorization Server is known to support RFC 7591) must still register on first use.
2. Specific gateways must **never** receive an automated registration (SaaS providers with pre-provisioned per-customer clients), even when the deployment default is auto-register.
3. A downstream platform integrates "MCP server with OAuth" as a first-class model and already carries a per-server `useDcr` toggle in its own data model; ContextForge currently stores the flag (`oauth_config` is a free-form dict) but ignores it — every gateway missing a `client_id` gets auto-registered whether the operator asked for it or not.

## Root Cause

RFC 7591 registration is implemented and complete at the service layer (`dcr_service.py`, `RegisteredOAuthClient`, admin endpoints, docs, tests), but the **decision** to invoke it is a single global condition:

```python
if issuer and not client_id:
    if settings.dcr_enabled and settings.dcr_auto_register_on_missing_credentials:
        # register...
```

There is no per-gateway signal in that gate. Because `oauth_config` is itself an unconstrained dictionary, clients already send flags like `use_dcr` into it today — the gateway just never reads them, so stored-but-inert configuration accumulates silently.

## Solution

Add a per-gateway **`oauth_config.use_dcr`** flag with explicit precedence over the global auto-register setting.

| `oauth_config.use_dcr` | `MCPGATEWAY_DCR_ENABLED` | `MCPGATEWAY_DCR_AUTO_REGISTER_ON_MISSING_CREDENTIALS` | Result when gateway has `issuer` but no `client_id` |
|---|---|---|---|
| absent | `true` | `true` | DCR (current behavior — backward compatible) |
| absent | any | `false` | 400: provide credentials manually (current behavior) |
| `true` | `true` | any | **DCR (force registration)** — new |
| `true` | `false` | any | 400: DCR disabled globally — the deployment master switch always wins |
| `false` | any | any | **never DCR** → 400: provide credentials manually — new |
| invalid | any | any | rejected at gateway create/update (400); treated as absent at runtime (legacy-row safety net) |

Key design decisions:

- **Interaction contract:** `use_dcr` accepts native booleans **or** string `"true"`/`"false"` (case-insensitive), so string-serializing clients (e.g. `Boolean.toString()`) work unchanged.
- **Empty means unset:** `""`/whitespace-only values are treated as absent, matching the existing convention for every other `oauth_config` field (`_validate_oauth_config_urls` skips `""` values).
- **Write-time strict, read-time lenient:** invalid values are rejected with 400 at gateway create/update (new `_validate_oauth_config_flags` chained into `GatewayCreate`/`GatewayUpdate`), and invalid values already persisted are logged and treated as absent at runtime.
- **Single change site:** the Phase 1.4 decision gate in `oauth_router.py`; `dcr_service.py`, `config.py`, `db.py`, and the migration surface are untouched.
- **Backward compatibility is a hard requirement:** gateways without the flag behave exactly as today.

## Testing

- 5 new runtime tests in `tests/unit/mcpgateway/routers/test_oauth_router.py`: force-enable overrides global-off; string `"false"` opt-out with global-on (asserts no registration call); master-switch refusal; invalid value treated as absent; absent-flag backward-compatibility pin.
- Parser edge cases + schema write-time rejection/acceptance tests in `tests/unit/mcpgateway/test_schemas_validators_extra.py` (bool and string forms, empty/whitespace, integers, malformed strings).
- Existing DCR test suite (success, disabled, unexpected-error, resource-persistence regressions) remains green.

## Out of Scope

- RFC 8414 protected-resource discovery (DCR from base URL only, without an issuer) — separate follow-up.
- Per-gateway registration life-cycle management UI — separate follow-up.
- Downstream platform integration changes (mapping its `useDcr` model field onto `oauth_config.use_dcr`).

## Additional Context

- Existing docs: `docs/docs/manage/dcr.md`
- DCR service: `mcpgateway/services/dcr_service.py`
- This is the first of a two-part change; the second part (in the integrating platform) maps its per-server `useDcr` toggle onto this key and activates it in its UI.

MohanLaksh:
## Problem

DCR is all-or-nothing per deployment. `MCPGATEWAY_DCR_ENABLED` and `MCPGATEWAY_DCR_AUTO_REGISTER_ON_MISSING_CREDENTIALS` decide registration for the entire gateway fleet; there is no way to express per-gateway intent. Integrators that already carry a per-server `useDcr` toggle in their own data model send it into `oauth_config`, where it is stored and ignored.

## Root Cause

The Phase 1.4 decision gate in `oauth_router.py` reads only global settings:

```python
if issuer and not client_id:
    if settings.dcr_enabled and settings.dcr_auto_register_on_missing_credentials:
```

`oauth_config` is a free-form dictionary, so per-gateway flags arrive today — but nothing reads them, so stored-but-inert configuration accumulates silently and every credential-less gateway auto-registers (or none do).

## Solution

Add `oauth_config.use_dcr`, a per-gateway override with explicit precedence over the global auto-register setting:

| `use_dcr` | `MCPGATEWAY_DCR_ENABLED` | `MCPGATEWAY_DCR_AUTO_REGISTER_...` | Result (issuer present, no client_id) |
|---|---|---|---|
| absent | true | true | DCR — current behavior |
| absent | any | false | 400 — current behavior |
| **true** | true | any | **DCR (force)** — new |
| true | false | any | 400 — global master switch always wins |
| **false** | any | any | **never DCR → 400** — new |
| invalid | any | any | rejected at create/update; treated as absent at runtime |

**Design decisions:**
- Primary change site: the Phase 1.4 gate. No `config.py`, `db.py`, or migration changes; `dcr_service.py` gains one optional passthrough parameter (see the second change below).
- Accepts native booleans **and** `"true"`/`"false"` strings (case-insensitive) so string-serializing clients (e.g. ICA's `Boolean.toString()`) work unchanged (`parse_use_dcr_flag` in `common/validators.py`).
- `""`/whitespace → absent, matching the existing empty-means-unset convention for `oauth_config` fields.
- Write-time strict (new `_validate_oauth_config_flags` chained into `GatewayCreate`/`GatewayUpdate` after the URL validator), read-time lenient (invalid persisted values log + behave as absent).
- Backward compatibility is pinned by a dedicated regression test: gateways without the flag behave exactly as before.

## Implementation Details

- `mcpgateway/common/validators.py` — `parse_use_dcr_flag(value) -> Optional[bool]`; `None`/empty → `None`, bool passthrough, case-insensitive `"true"`/`"false"`, `ValueError` otherwise.
- `mcpgateway/schemas.py` — `_validate_oauth_config_flags` validates `use_dcr` via the parser; chained after `_validate_oauth_config_urls` in both `GatewayCreate` and `GatewayUpdate` field validators.
- `mcpgateway/routers/oauth_router.py` — Phase 1.4 gate restructured: flag parse with safety net → explicit opt-out 400 (`use_dcr=false`) → master-switch 400 (env-var-named, distinct from gateway-config errors) → `should_register = flag is True or global auto-register` → existing DCR attempt body unchanged (DcrError 500 path intact).
- `tests/unit/mcpgateway/routers/test_oauth_router.py` — `TestUseDcrFlagGate`: force-enable overrides global-off; string `"false"` opt-out with global-on (asserts zero registration calls); master-switch refusal; invalid-value runtime fallback; absent-flag backward-compat pin.
- `tests/unit/mcpgateway/test_schemas_validators_extra.py` — `TestParseUseDcrFlag` (10 valid + 8 invalid parser cases) and `TestUseDcrSchemaValidation` (write-time rejection, bool/string/empty acceptance, GatewayUpdate parity).
- `docs/docs/manage/dcr.md` — "Per-Gateway Control (`use_dcr`)" section with the semantics table, force/opt-out/absent JSON examples, and a global→per-gateway migration subsection (audit SQL + steps).

## Testing

- `pytest tests/unit/mcpgateway/routers/test_oauth_router.py` — 189 passed (184 pre-existing + 5 new)
- `pytest tests/unit/mcpgateway/test_schemas_validators_extra.py` — 121 passed (incl. new parser + schema classes)
- Combined run of both files: 310 passed
- `ruff` — clean on all changed lines (3 remaining findings in the touched test file pre-exist on `main`; verified by stashing)
- `mypy` — error count identical to baseline (3834 before = 3834 after) on the three changed source files; zero findings on changed lines
- Existing DCR suite (success, disabled, unexpected-error, learned-resource persistence regressions) unchanged and green

## Checklist

- [x] Tests pass locally (targeted + combined)
- [x] Documentation updated (dcr.md)
- [x] Backward compatibility regression test pinned
- [ ] Security implications considered — flag is read-only at the gate; no new privileges, no SSRF surface (no new URLs), no logging of secrets
---

### Problem

ICA 1.0 consumes `useClientSecretTokenAuthMethod` at DCR registration time, but ContextForge's registration request always uses the global `MCPGATEWAY_DCR_TOKEN_ENDPOINT_AUTH_METHOD`. ICA20 sends the boolean into `oauth_config`, where it is stored and ignored — the same stored-but-inert pattern this PR eliminates for `use_dcr`.

### Solution

Add `oauth_config.token_endpoint_auth_method`, an optional per-gateway override for the RFC 7591 registration request:

| `token_endpoint_auth_method` | Registration payload |
|---|---|
| `client_secret_basic` | `token_endpoint_auth_method: client_secret_basic` |
| `client_secret_post` | `token_endpoint_auth_method: client_secret_post` |
| absent / empty | global `MCPGATEWAY_DCR_TOKEN_ENDPOINT_AUTH_METHOD` (byte-identical payload) |

- Only the two methods the runtime actually executes at the token endpoint are accepted; `none`/`private_key_jwt`/anything else → 400 at create/update, treated as absent (log) at registration time — same write-time-strict / read-time-lenient pattern as `use_dcr`.
- The AS registration response stays authoritative: the method the AS confirms is stored back into `oauth_config` and used for token requests.
- Backward compatibility pinned: absent value ⇒ `dcr_service` behavior is byte-identical to before (global setting), verified by a dedicated test.

### Implementation Details

- `mcpgateway/common/validators.py` — `parse_token_endpoint_auth_method(value) -> Optional[str]` mirroring `parse_use_dcr_flag`.
- `mcpgateway/schemas.py` — `_validate_oauth_config_flags` now chained through both parsers (shared by `GatewayCreate`/`GatewayUpdate`).
- `mcpgateway/services/dcr_service.py` — `token_endpoint_auth_method: Optional[str] = None` on `register_client`/`get_or_register_client`; request uses it when present, else the global setting (also used as the stored-record fallback when the AS response omits the field).
- `mcpgateway/routers/oauth_router.py` — parsed at the Phase 1.4 gate with the same safety net as `use_dcr`, forwarded to the registration call.
- Tests: `TestParseTokenEndpointAuthMethod` + `TestTokenEndpointAuthMethodSchemaValidation` (schemas/validators), per-gateway vs global-default payload assertions (`test_dcr_service.py`), router-level forward + invalid-at-runtime-as-absent (`test_oauth_router.py`).

### Testing

- Full unit suite: 21,841 passed, 2 failed — both failures (`test_config.py` CSRF warnings) reproduce identically on `main` with the changes stashed. 34 new tests added, zero regressions.
- `ruff` — zero new findings on changed files/lines (all remaining findings verified pre-existing on `main` via stash).
- `interrogate` — 100% on all four changed source files.
- Secret scan — no new secrets; baseline re-generated and hook green.

## Out of Scope / Excluded (tracked here)

| Item | Status | Reason |
|---|---|---|
| `appendLoginHint`, `skipScope`, `userInfoUri`, `accessTokenType` activations in `oauth_config` | ❌ cancelled | Re-assessed as not required by ICA20 adapters; keys stay stored-but-inert in `oauth_config` until a consumer exists |
| 1.0-style boolean `useClientSecretTokenAuthMethod` false → `none` mapping | ⏳ follow-up | Runtime cannot authenticate public (`none`) clients; keep boolean↔string mapping documented, implement `none` when runtime supports it |
| `loginAttributes` (ICA `feature/ICAA-41111-main`) | ⏳ follow-up issues | New in ICA 1.0; absent from ICA20 and ContextForge — filed as separate issues in both repos, deliberately excluded from this PR |
| DCR `client_id`/registration drift fixes | ❌ excluded | Out of this PR's scope per milestone plan |

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -1,9 +1,9 @@
 {
   "exclude": {
-    "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^\\.secrets\\.baseline$",
+    "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^\\.secrets\\.baseline$|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-08-12T15:30:22Z",
+  "generated_at": "2026-08-10T16:59:56Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -258,7 +258,7 @@
         "hashed_secret": "b4673e578b9b30fe8bba1b555b7b59883444c697",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 2047,
+        "line_number": 2039,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -266,7 +266,7 @@
         "hashed_secret": "4a0a2df96d4c9a13a282268cab33ac4b8cbb2c72",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 2135,
+        "line_number": 2127,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -274,7 +274,7 @@
         "hashed_secret": "5baa61e4c9b93f3f0682250b6cf8331b7ee68fd8",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 2485,
+        "line_number": 2477,
         "type": "Basic Auth Credentials",
         "verified_result": null
       },
@@ -282,7 +282,7 @@
         "hashed_secret": "fa9beb99e4029ad5a6615399e7bbae21356086b3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 3850,
+        "line_number": 3842,
         "type": "Basic Auth Credentials",
         "verified_result": null
       },
@@ -290,7 +290,7 @@
         "hashed_secret": "fa9beb99e4029ad5a6615399e7bbae21356086b3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 3941,
+        "line_number": 3933,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -298,7 +298,7 @@
         "hashed_secret": "ac371b6dcce28a86c90d12bc57d946a800eebf17",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 3984,
+        "line_number": 3976,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -306,7 +306,7 @@
         "hashed_secret": "0b6ec68df700dec4dcd64babd0eda1edccddace1",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 3989,
+        "line_number": 3981,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -314,7 +314,7 @@
         "hashed_secret": "4ad6f0082ee224001beb3ca5c3e81c8ceea5ed86",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 3994,
+        "line_number": 3986,
         "type": "Secret Keyword",
         "verified_result": null
       }
@@ -390,7 +390,7 @@
         "hashed_secret": "d3ac7a4ef1a838b4134f2f6e7f3c0d249d74b674",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 6079,
+        "line_number": 6064,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -398,7 +398,7 @@
         "hashed_secret": "5932862bcd24dd27d0dc0407ec94fe9d6ea24aeb",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 6576,
+        "line_number": 6561,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -406,7 +406,7 @@
         "hashed_secret": "c77c805e32f173e4321ee9187de9c29cb3804513",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 6588,
+        "line_number": 6573,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -414,7 +414,7 @@
         "hashed_secret": "8fe3df8a68ddd0d4ab2214186cbb8e38ccd0e06a",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 6660,
+        "line_number": 6645,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -422,7 +422,7 @@
         "hashed_secret": "bb589d0621e5472f470fa3425a234c74b1e202e8",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 7076,
+        "line_number": 7061,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -430,7 +430,7 @@
         "hashed_secret": "93ac8946882128457cd9e283b30ca851945e6690",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 7767,
+        "line_number": 7752,
         "type": "Secret Keyword",
         "verified_result": null
       }
@@ -836,7 +836,7 @@
         "hashed_secret": "2df14e4719f299249cd9a97cf68cc87232a27cbb",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1749,
+        "line_number": 1741,
         "type": "Hex High Entropy String",
         "verified_result": null
       },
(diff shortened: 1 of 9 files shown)


jonpspri:
I don't want to pursue this unless absolutely necessary.  DCR has been deprecated with the latest release of MCP.

PR #6174, opened 2026-08-10
fix: omit transport from dataplane config

lucarlig:
## Summary

- keep the dataplane publisher limited to `STREAMABLEHTTP` gateways
- omit the obsolete `transport` property from published backend configuration
- cover both SSE and STDIO with the existing non-Streamable gateway filter test

## Why

The Rust dataplane now has a single unconditional upstream transport and no
transport selector. The control-plane publisher should match that contract
instead of sending a field the dataplane no longer models.

This is the control-plane companion to
[contextforge-org/contextforge-data-plane#79](https://github.com/contextforge-org/contextforge-data-plane/pull/79).

## Impact

SSE and STDIO remain available on control-plane-owned routes. This change only
prevents those gateway types from entering dataplane configuration and removes
transport selection from the published backend payload.

## Validation

- dataplane publisher unit tests: 25 passed
- Ruff: passed
- Bandit: passed
- pylint on the publisher: 10.00/10
- repository pre-commit hooks: passed

### DIFF
--- a/mcpgateway/services/dataplane_publisher.py
+++ b/mcpgateway/services/dataplane_publisher.py
@@ -62,7 +62,6 @@ class BackendConfig(TypedDict):
 
     name: str
     url: str
-    transport: str
     passthrough_headers: list[str]
     add_headers: dict[str, str]
     remove_headers: list[str]
@@ -229,14 +228,13 @@ class DataplanePublisherService:
 
             prompt_map = {prompt["id"]: prompt["name"] for prompt in prompts}
 
-            # The dataplane proxies streamable-HTTP upstreams only. Legacy
-            # HTTP+SSE gateways are excluded so the published config never
-            # advertises a backend the dataplane cannot serve.
+            # The dataplane proxies streamable-HTTP upstreams only. Exclude
+            # every other transport before building its transport-agnostic
+            # backend config.
             gateway_base = {
                 gateway["id"]: {
                     "name": gateway["name"],
                     "url": gateway["url"],
-                    "transport": gateway["transport"],
                     "passthrough_headers": gateway["passthrough_headers"] or [],
                     "add_headers": gateway.get("add_headers") or {},
                     "remove_headers": gateway.get("remove_headers") or [],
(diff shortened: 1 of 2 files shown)


mcpgateway/services/dataplane_publisher.py
```diff
@@ -229,14 +228,13 @@ def create_payload(
...
-            # HTTP+SSE gateways are excluded so the published config never
-            # advertises a backend the dataplane cannot serve.
+            # The dataplane proxies streamable-HTTP upstreams only. Exclude
+            # every other transport before building its transport-agnostic
+            # backend config.
             gateway_base = {
                 gateway["id"]: {
                     "name": gateway["name"],
                     "url": gateway["url"],
-                    "transport": gateway["transport"],
```
cafalchio, line 239: Perhaps you need to filter here and send just the gateways that are compatible with the StreamableHttp?

cafalchio (commented):
I think it should filter to send the compatible gateways, otherwise we will need the transport to filter it out when calling a tool in dataplane.

Lang-Akshay (approved):
LGTM

msureshkumar88 (approved):
Reviewed the change against the companion Rust PR (contextforge-data-plane#79) and the existing filter logic — this is a clean, minimal fix.

**What it does**: drops the now-unused `transport` field from the published `BackendConfig`/dataplane payload, matching the Rust side's move to a single unconditional upstream transport. The `STREAMABLEHTTP`-only gateway filter itself is unchanged (pre-existing logic, confirmed via blame) — only the comment was reworded and the field removed from the outbound payload.

**Verified**:
- No other in-repo consumer references the removed payload key.
- `DbGateway.transport` is still read internally (needed for the filter) and just no longer serialized — not dead code.
- Test changes are in-scope: dict-equality assertion now guards against the field silently reappearing, and the non-streamable-gateway test is parametrized to cover both SSE and STDIO.
- Ran the full `test_dataplane_publisher.py` suite in an isolated worktree: 25/25 passing.
- No UI, alembic/schema, or docs changes needed — none touched, correctly.

**One non-blocking note**: since this changes a cross-repo wire contract with no schema version field, an environment that already has `dataplane_publisher=true` enabled would break if this lands before the companion `data-plane#79` deploys. Worth a one-line rollout note (deploy dataplane side first), but given the flag defaults to `false` and the module is explicitly marked experimental, this isn't a blocker.

Approving.

mcpgateway/services/dataplane_publisher.py
lucarlig: there already is a filter

PR #6180, opened 2026-08-11
fix(admin): repair gRPC services UI form actions

a-effort:
## Summary
- Register form's fetch call was missing the `X-CSRF-Token` header that every other admin form attaches, so cookie-authenticated submissions were silently rejected by the CSRF middleware.
- The admin page's `list_services()` call omitted `page`/`per_page`, which changes the service's return type from a paginated dict to a `(list, cursor)` tuple. The template context loader iterated that tuple as if it were the service list, threw, and the surrounding `except Exception` swallowed it — always rendering "No gRPC services registered yet." regardless of what was actually registered.
- The `state`/`delete`/`reflect` routes are triggered by native HTML form posts but returned raw JSON/204 responses instead of redirecting, so clicking Deactivate, Delete, or Re-Reflect navigated the browser to a bare JSON page instead of back to the admin UI.

### DIFF
--- a/mcpgateway/admin.py
+++ b/mcpgateway/admin.py
@@ -4206,7 +4206,7 @@ async def admin_ui(
     grpc_services = []
     try:
         if "grpc-services" not in hidden_sections and GRPC_AVAILABLE and grpc_service_mgr and settings.mcpgateway_grpc_enabled:
-            grpc_services_raw = await grpc_service_mgr.list_services(
+            grpc_services_raw, _grpc_next_cursor = await grpc_service_mgr.list_services(
                 db,
                 include_inactive=include_inactive,
                 user_email=user_email,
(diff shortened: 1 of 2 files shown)


a-effort:
Closing this as it targets the legacy Admin UI and improvements should be worked on for the new UI instead.

The bugs described here are still present on main, but the fix as written does not hold up now that the admin router is re-mounted under `/v1`, and the remaining work belongs with the product route in #6546.

Findings recorded there: https://github.com/IBM/mcp-context-forge/issues/6546#issuecomment-5545645193

Corrections to this PR's description:

- The CSRF hunk was a no-op. `wrapFetchWithCsrf` in `admin.html` has added `X-CSRF-Token` to every same-origin unsafe fetch since #3129, and `formSubmitHandlers.js` carries no explicit CSRF header anywhere, so "the header that every other admin form attaches" is not accurate. Registration was most likely never broken. The symptom is fully explained by the list never rendering.
- Converting `state`, `delete`, and `reflect` to 303 redirects changes `/v1/admin/grpc/*` as well, because those handlers are shared between the legacy console and the `/v1` mount.

The one fix here that was correct but legacy-only is the tuple unpack at `admin.py:4208`. Root cause is written up in the linked comment.

PR #3202, opened 2026-02-24
feat(transport): add gRPC methods as MCP tools

crivetimihai:
> **Note:** This PR was re-created from #2983 due to repository maintenance. Your code and branch are intact. @gabe-l-hart please verify everything looks good.

## 🔗 Related Issue
Addresses #2854

---

## 📝 Summary
This PR implements automatic tool registration for gRPC methods

---

## 🏷️ Type of Change
- [x] Bug fix
- [x] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Chore (deps, CI, tooling)
- [ ] Other (describe below)

---

## 🧪 Verification

| Check                     | Command         | Status |
|---------------------------|-----------------|--------|
| Lint suite                | `make lint`     |        |
| Unit tests                | `make test`     |        |
| Coverage ≥ 80%            | `make coverage` |        |

---

## ✅ Checklist
- [ ] Code formatted (`make black isort pre-commit`)
- [ ] Tests added/updated for changes
- [ ] Documentation updated (if applicable)
- [ ] No secrets or credentials committed

---

## 📓 Notes (optional)
_Screenshots, design decisions, or additional context._

PR #6181, opened 2026-08-11
fix(tools): return MCP-compliant error responses for tool execution failures

### LINKED ISSUE #6275: MCP protocol violation — tool execution failures return JSON-RPC errors instead of CallToolResult
## Summary

Tool execution failures (connection errors, timeouts, unhandled exceptions) in the SSE, StreamableHTTP, and Direct Proxy transports return JSON-RPC error objects instead of MCP-compliant `CallToolResult` structures. This causes MCP clients such as `langchain-mcp-adapters` to fail with `content: Field required` validation errors.

## Background

The MCP specification requires that `tools/call` responses **always** return a successful JSON-RPC result whose `result` field contains a `CallToolResult` with a mandatory `content` array. Runtime failures (server crashes, timeouts, network errors) must be communicated via `isError: true` inside the `CallToolResult`, not via JSON-RPC error objects.

## Current Behaviour

When an upstream MCP server fails, times out, or throws an exception, the gateway raises `ToolInvocationError` or `ToolTimeoutError`, which propagates out as a JSON-RPC error response:

```json
{"jsonrpc": "2.0", "error": {"code": -32000, "message": "Tool invocation failed"}, "id": 1}
```

## Expected Behaviour

All runtime transport failures must be wrapped in a `CallToolResult` and returned as a successful JSON-RPC response:

```json
{
  "jsonrpc": "2.0",
  "result": {
    "content": [{"type": "text", "text": "MCP server error: Connection refused"}],
    "isError": true
  },
  "id": 1
}
```

## Affected Code

- `mcpgateway/services/tool_service.py` — `invoke_tool_direct` exception handler
- `mcpgateway/services/tool_service.py` — SSE transport timeout and exception handlers
- `mcpgateway/services/tool_service.py` — StreamableHTTP transport timeout and exception handlers

## Fix

PR #6181 addresses this by returning `types.CallToolResult(content=[...], isError=True)` in all five affected error paths instead of raising exceptions, and updating the five corresponding tests.

Closes #6181

bogdanmariusc10:
# Pull Request

Fixes MCP protocol violation where tool execution failures returned JSON-RPC errors instead of properly structured MCP `CallToolResult` objects, causing `content: Field required` validation errors in MCP clients.

**Background**: This bug was discovered when langchain-mcp-adapters failed to parse error responses from Context Forge because the gateway returned JSON-RPC error objects instead of MCP-compliant `CallToolResult` structures with `{content: [...], isError: true}`.

---

## 📝 Summary

**Problem**: The Context Forge gateway violated the MCP protocol specification by raising exceptions and returning JSON-RPC errors when MCP servers failed, timed out, or returned malformed responses. According to the MCP spec, ALL tool results must have a `content` field and use the `isError` boolean to indicate failures.

**Solution**: Modified all MCP execution paths to return properly structured `types.CallToolResult` objects with `isError=true` instead of raising exceptions. This ensures protocol compliance while preserving appropriate exception handling for pre-execution errors (authentication, validation, configuration).

### Changes Made

1. **Direct Proxy Mode** (`invoke_tool_direct`):
   - Returns `CallToolResult(content=[...], isError=True)` on connection/execution failures
   - Instead of: Raising `ToolInvocationError`

2. **SSE Transport**:
   - **Timeout handler**: Returns structured error response instead of raising `ToolTimeoutError`
   - **Exception handler**: Returns structured error response instead of re-raising `BaseException`

3. **StreamableHTTP Transport**:
   - **Timeout handler**: Returns structured error response instead of raising `ToolTimeoutError`
   - **Exception handler**: Returns structured error response instead of re-raising `BaseException`

4. **Test Updates**: Updated 5 tests to verify structured error responses:
   - `test_mcp_sse_timeout_triggers_post_hook_and_cb_context`
   - `test_mcp_sse_exception_group_is_logged_and_wrapped`
   - `test_streamablehttp_timeout_triggers_post_hook_without_context`
   - `test_streamablehttp_exception_group_is_logged_and_wrapped`
   - `test_invoke_tool_direct_connection_error`

### Impact

**Before**: MCP server failures caused `content: Field required` validation errors in clients
```json
{"jsonrpc": "2.0", "error": {"code": -32000, "message": "Tool invocation failed"}, "id": 1}
```

**After**: MCP server failures return properly structured responses
```json
{
  "jsonrpc": "2.0",
  "result": {
    "content": [{"type": "text", "text": "MCP server error: Connection refused"}],
    "isError": true
  },
  "id": 1
}
```

---

## 📏 Reviewability

- [x] This PR has one clear purpose: Fix MCP protocol violation in error response handling
- [x] The linked issue is not labeled `triage` (bug discovered during investigation)
- [x] Tests are included with the code they validate (5 test updates)

**Scope**: This PR strictly addresses MCP protocol compliance for runtime execution errors. Pre-execution errors (auth, validation, configuration) continue to raise exceptions as designed.

---

## 🏷️ Type of Change

- [x] Bug fix (MCP protocol violation)
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

**Category**: Critical bug fix for protocol compliance

---

### Test Results

All affected tests now pass with updated assertions:

```bash
# Run specific test suite
pytest tests/unit/mcpgateway/services/test_tool_service_coverage.py::TestInvokeToolMcpSseTimeoutAndErrors -xvs
pytest tests/unit/mcpgateway/services/test_tool_service_coverage.py::TestInvokeToolMcpStreamableHttpCoverage -xvs
pytest tests/unit/mcpgateway/services/test_tool_service.py::TestInvokeToolDirect::test_invoke_tool_direct_connection_error -xvs
```

### Manual Verification Steps

1. **Simulate MCP Server Failure**:
   ```bash
   # Register a gateway pointing to a non-existent MCP server
   curl -X POST http://localhost:4444/gateways \
     -d '{"name":"test","url":"http://localhost:9999","auth_type":"none"}'
   
   # Invoke tool - should return structured error, not raise exception
   curl -X POST http://localhost:4444/mcp/sse \
     -d '{"jsonrpc":"2.0","method":"tools/call","params":{"name":"test_tool","arguments":{}},"id":1}'
   ```
   
   **Expected Response**:
   ```json
   {
     "jsonrpc": "2.0",
     "result": {
       "content": [{"type": "text", "text": "MCP server error: Connection refused"}],
       "isError": true
     },
     "id": 1
   }
   ```

2. **Verify langchain-mcp-adapters Compatibility**:
   - MCP clients should now successfully parse error responses
   - No more `content: Field required` validation errors

### Verification Checklist

| Check                     | Command                              | Status |
|---------------------------|--------------------------------------|--------|
| Modified file only        | `git diff --stat main`               | ✅ 3 files changed |
| Lint suite                | `make ruff pylint`                   | ✅ Pass |
| Unit tests (all)          | `make test`                          | ✅ Pass |
| Coverage ≥ 80%            | `make coverage`                      | ✅ Pass |
| Specific test suite       | `pytest tests/unit/.../test_tool_*`  | ✅ Pass (5 tests updated) |
| Pre-commit hooks          | `make pre-commit`                    | ✅ Pass |

---

## ✅ Checklist

- [x] Tests added/updated for changes (5 tests updated)
- [x] Documentation updated (ROOT_CAUSE_ANALYSIS_GATEWAY_BUG.md created)
- [x] Signed commit (`git commit -s`)
- [x] All affected MCP execution paths reviewed and verified

---

### Design Decisions

1. **Runtime vs Pre-execution Errors**:
   - **Runtime errors** (during `session.call_tool()`): Return `CallToolResult` with `isError=True`
   - **Pre-execution errors** (auth, validation, config): Continue raising exceptions
   - **Rationale**: Pre-execution errors prevent tool execution from starting and are not part of the MCP tool result contract

2. **Error Message Format**:
   - Prefix: `"MCP server error: ..."` for clarity
   - Preserves original error details while sanitizing sensitive information
   - Maintains existing logging behavior (structured logs with correlation IDs)

3. **Transport Coverage**:
   - Fixed all MCP transports: SSE, StreamableHTTP, and Direct Proxy
   - Non-MCP integrations (REST, gRPC, A2A) unchanged (different protocol requirements)

### MCP Protocol Reference

From the MCP specification:
```typescript
interface CallToolResult {
  content: Content[];  // REQUIRED - array of content items
  isError?: boolean;   // true if tool execution failed
}
```

**Key Point**: `CallToolResult` is ALWAYS returned as the `result` field in a successful JSON-RPC response, even when tool execution fails. The `isError` boolean differentiates success from failure.

### Files Modified

1. **mcpgateway/services/tool_service.py** (+50/-7 lines):
   - Line 3815-3825: `invoke_tool_direct` exception handler
   - Line 6098-6133: SSE timeout handler
   - Line 6131-6156: SSE exception handler
   - Line 6299-6334: StreamableHTTP timeout handler
   - Line 6339-6364: StreamableHTTP exception handler

2. **tests/unit/mcpgateway/services/test_tool_service_coverage.py** (+48/-10 lines):
   - Updated 4 tests to verify structured error responses

3. **tests/unit/mcpgateway/services/test_tool_service.py** (+18/-6 lines):
   - Updated 1 test to verify structured error responses


bogdanmariusc10:
Hi @msureshkumar88! Can you take a look into this?

jonpspri:
Tracking issue for this fix: #6275

jonpspri (commented):
## PR Review

Good fix — the core change is correct and the five affected test updates all verify the right behaviour. A few things worth addressing before merge:

### Blocking

**Stale `Raises:` docstring in `invoke_tool_direct`** — the docstring at line ~3733 still lists `ToolInvocationError: If invocation fails.` under `Raises:`, but the exception is now swallowed and converted to a `CallToolResult` for all runtime errors. This is a documentation contract violation: callers reading the docstring may still guard against an exception that will never arrive. The `Raises:` entry should either be removed or narrowed to the pre-execution checks that do still raise (gateway not found, not in direct_proxy mode, access denied).

The same inline comment block added at line ~3841 says "Raising ToolInvocationError would result in..." — since the raise is now gone, tightening to "Previously raised ToolInvocationError, which..." would be clearer.

### Suggestions

- **`api` label** — this label reads as REST API work; `mcp-protocol` (already present) is the right scope label here. Worth dropping `api` to keep the label set accurate.
- **`ROOT_CAUSE_ANALYSIS_GATEWAY_BUG.md`** — the checklist mentions this file was created, but it's not visible in the diff. If it's being added, it should appear in the changed files; if it's a local artifact not intended for the repo, remove it from the checklist item.
- **Dual attribute check in tests** — `getattr(result, 'is_error', None) or getattr(result, 'isError', None)` appears in four test assertions. Since the PR always returns `types.CallToolResult` (which uses camelCase `isError`), the `is_error` branch will never be hit. Simplifying to `result.isError` (matching the existing `test_invoke_tool_direct_connection_error` pattern) would make the intent clearer.

### What's good

- The fix correctly targets only the five runtime-error paths while preserving exception raising for pre-execution errors (auth, validation, config) — that boundary is well-reasoned and the PR description explains it clearly.
- The `invoke_tool` caller at lines 6406–6428 already handles returned `CallToolResult` values correctly via `_coerce_to_tool_result`, so no callsite changes are needed.
- `asyncio.CancelledError` continues to propagate in all transports — correct, that must never be swallowed.
- Five tests updated, one per changed error path — good coverage.

Tracking issue: #6275

bogdanmariusc10:
Thanks, @jonpspri! All blocking issues and suggestions have been addressed:

- **Blocking**: Updated `invoke_tool_direct` docstring to accurately reflect exception behavior for runtime vs pre-execution errors
- **Blocking**: Changed inline comment to past tense ("Previously raised...")
- **Suggestion**: Simplified test assertions to use `result.isError` directly (removed dual attribute checks)

Changes committed in f41c5383c.

Note: The failing test is unrelated to my changes.

msureshkumar88:
**Overlap with `chore/mcp-sdk-v2`**

This fix and the `chore/mcp-sdk-v2` branch both modify the same functions in `mcpgateway/services/tool_service.py`: `invoke_tool_direct`, and the `connect_to_sse_server` / `connect_to_streamablehttp_server` closures inside `invoke_tool`.

`chore/mcp-sdk-v2` rewrites these paths for the MCP SDK 2.0 migration (`mcp>=1.28.1,<2` to `mcp>=2.0.0`, new `mcp-types>=2.0.0` package, `sse_client`/`streamablehttp_client`/`ClientSession` replaced by `mcp_proxy_client()`). That branch was cut before this fix existed. Its rewritten error paths still `raise ToolInvocationError(...)` / `raise ToolTimeoutError(...)` instead of returning `CallToolResult(isError=True)`. The protocol violation this PR fixes is unresolved on that branch.

Two consequences:

1. Merging this PR now and `chore/mcp-sdk-v2` later produces a conflict in `tool_service.py` on these exact hunks. Whoever rebases that branch must reapply the `isError`-return fix on top of the new `mcp_proxy_client` call sites.
2. Before reapplying, confirm the `CallToolResult`/`isError` shape against `mcp-types>=2.0.0` — the branch imports `mcp_types as types` in place of `mcp.types`, and the field name is unverified on the new package.

**Why merge now, not wait for `chore/mcp-sdk-v2`:**

- This fixes a live client-breaking bug on current `main` (`content: Field required` failures against `langchain-mcp-adapters` and any spec-strict MCP client). The SDK migration has no committed merge date, and current `main` stays non-compliant with the MCP spec until then.
- The fix is small, isolated to error-response shape, and does not touch the transport or session-establishment logic the SDK migration is rewriting. Low conflict surface, low risk to land independently.
- Waiting compounds the gap: every commit added to `chore/mcp-sdk-v2` in the meantime touches the same functions, growing the eventual rebase.

msureshkumar88 (changes requested):
## Review — changes requested

The core fix (returning `CallToolResult(isError=True)` instead of raising) is the correct direction and matches the MCP spec. Two items need fixing before merge; the rest are should-fix, not blockers.

### Blocking

**1. `mcpgateway/services/tool_service.py:3957-3964` — unsanitized exception text returned to client (CWE-209, information exposure)**

`invoke_tool_direct`'s new error path returns `f"MCP server error: {str(e)}"` verbatim to the MCP client. The sibling SSE/StreamableHTTP handlers three lines away route the equivalent message through `sanitize_exception_message(str(root_cause), gateway_auth_query_params_decrypted)` (`tool_service.py:6525`, `6720`) before returning it. `sanitize_exception_message` (`mcpgateway/utils/url_auth.py:182`) exists specifically to redact auth query params embedded in exception text.

Before this PR, `invoke_tool_direct`'s exception never reached the client — `mcpgateway/transports/streamablehttp_transport.py:1848-1850` caught it and substituted the generic string `"Direct proxy tool invocation failed"`. This PR is a regression on that path: a failed connection to a gateway with a token/API key in its URL can now return that URL, unredacted, in the tool result.

Fix: route the message through `sanitize_exception_message` before constructing the `CallToolResult`, matching the SSE/StreamableHTTP paths.

**2. `mcpgateway/services/tool_service.py:6515`, `6710` — `retry_on_status` loses its input on this path**

The outer handler at `tool_service.py:7157-7181` builds `structured_content={"status_code": root_cause.response.status_code}` from `httpx.HTTPStatusError` specifically so the retry plugin can honor `retry_on_status` (comment at 7171-7173). Converting the inner `except BaseException` from `raise` to `return CallToolResult(...)` means SSE/StreamableHTTP failures no longer reach that outer handler, so a plugin configured with `retry_on_status: [429, 503]` can no longer see the status code for these transports. (Retry-on-timeout is unaffected — `_run_timeout_post_invoke` raises its own `ToolTimeoutError` with `retry_delay_ms` independently at `tool_service.py:5007`, and that path is untouched by this change.)

Fix: attach the same `structured_content` in the inner handler before returning, or re-raise so the outer handler retains ownership of both the retry decision and the final result shape.

### Functionally-impacting

**3. `mcpgateway/services/tool_service.py:6515`, `6710` — `except BaseException` now swallows `SystemExit`/`GeneratorExit`/`KeyboardInterrupt`**

Only `CancelledError` is re-raised above the catch-all (`6512`, `6707`); every other `BaseException` — including process-control signals — is now converted into an ordinary `isError=True` tool result instead of propagating. This repo has a documented prior incident with a related failure mode: `tool_service.py:7015-7019` explicitly re-raises `CancelledError` with a comment citing "the swallowed-cancellation bug from PR #3202 review B7." Recommend adding the same explicit guard for `SystemExit`/`GeneratorExit`/`KeyboardInterrupt` here.

### Should-fix (non-blocking)

- `tool_service.py:6355-6358`, `6551-6554` — docstrings for `connect_to_sse_server` / `connect_to_streamablehttp_server` still document `Raises: ToolInvocationError / ToolTimeoutError / BaseException`, but these paths now return instead of raising for those cases.
- `tool_service.py:3964`, `6511`, `6539`, `6706`, `6734` — the `CallToolResult(content=[TextContent(...)], isError=True)` construction (with its explanatory comment) is duplicated five times. A small helper would keep future changes to the error shape (e.g. adding `structured_content`, per finding #2) from being applied inconsistently.
- No test covers the sanitization gap in finding #1 or the `retry_on_status` regression in finding #2.
- No issue number referenced in the PR description.

bogdanmariusc10:
Thanks for the thorough review, @msureshkumar88! All findings are addressed in the latest commit (`5ef724e5a`):

**Blocking #1 (CWE-209)** — `invoke_tool_direct` now snapshots `gateway.auth_query_params` (decrypted) inside the ORM session and passes it through `sanitize_exception_message` before constructing the `CallToolResult`, matching the SSE/StreamableHTTP paths.

**Blocking #2 (retry_on_status)** — both SSE and StreamableHTTP inner `BaseException` handlers now build `structured_content={"status_code": root_cause.response.status_code}` when the root cause is `httpx.HTTPStatusError`, then pass it into the shared `_make_mcp_tool_error()` helper so the retry plugin can see the status code.

**Functionally-impacting #3** — added an explicit `except (SystemExit, GeneratorExit, KeyboardInterrupt): raise` guard above the `BaseException` catch-all in both handlers, mirroring the existing `CancelledError` guard (PR #3202 review B7).

**Should-fixes** — docstrings for `connect_to_sse_server` and `connect_to_streamablehttp_server` updated to reflect the return-instead-of-raise contract; the five duplicate `CallToolResult(isError=True)` constructions consolidated into a `_make_mcp_tool_error()` static helper.

**Tests** — added `test_invoke_tool_direct_error_is_sanitized` (verifies `sanitize_exception_message` is called and the secret token does not appear in the returned content) and `test_sse/streamablehttp_http_status_error_sets_structured_content_status_code` (verifies `structured_content.status_code` is present for `HTTPStatusError` on both transports).

msureshkumar88 (dismissed):
## Re-review of prior rounds

Verified against commits, not comments.

- Review 1 (Aug 17): addressed, except `api` label still attached and no issue number referenced (both open suggestions).
- Review 2 (Sep 18, CHANGES_REQUESTED):
  - Finding #1 (CWE-209 unsanitized error text in `invoke_tool_direct`): fixed correctly, routes through `sanitize_exception_message()` with a dedicated test.
  - Finding #2 (`retry_on_status` losing `structured_content`): fixed for one code path only, still broken on another (see Finding 2 below).
  - Finding #3 (`SystemExit`/`GeneratorExit`/`KeyboardInterrupt` swallowed): fixed correctly.

## Fresh findings

### Blocking

**1. Unretried MCP timeout fires `TOOL_POST_INVOKE` twice (SSE + StreamableHTTP)** — `tool_service.py:6554`, `:6755`

The timeout handlers changed `raise ToolTimeoutError(...)` to `return self._make_mcp_tool_error(...)`. The `return` now falls through into the normal post-process block at `:7085-7099`, which fires `TOOL_POST_INVOKE` a second time after `_run_timeout_post_invoke()` (`:5038`) already fired it once for the same timeout.

This breaks an invariant the codebase documents explicitly at `:7174-7176`:
```
except ToolTimeoutError as e:
    # ToolTimeoutError is raised by timeout handlers which already called tool_post_invoke.
    # Do NOT call post_invoke again — the retry_delay_ms signal is carried on the exception.
```

Verified empirically: reproduced the PR's own SSE timeout test with a hook-call counter — `TOOL_POST_INVOKE` fired 2 times, expected 1. Any post-invoke plugin (circuit breaker, rate limiter, audit log, retry plugin) double-counts the failure, and the second hook's `retry_delay_ms` can trigger a second, unintended retry decision at `:7129-7130` — the exact double-retry the `:7175` comment exists to prevent. The gRPC timeout path (`:7074-7078`) still raises correctly; only the two MCP transports regressed.

Fix: keep the `return` value but short-circuit before the post-process block reaches `:7089` for this path, or restore the `raise` and let the existing `ToolTimeoutError` handler at `:7174` build the MCP-compliant response.

**2. `structured_content` kwarg silently dropped on the direct-proxy retry path** — `tool_service.py:3834-3838`

`_make_mcp_tool_error` builds `types.CallToolResult` (the MCP SDK type) but passes `structured_content=`. That SDK model's field is `structuredContent`, with no snake_case alias, and `extra="allow"` — so the value lands as an unrelated extra attribute instead of the declared field.

Verified empirically:
```
r = types.CallToolResult(..., structured_content={"status_code": 429})
r.structuredContent  # -> None
```
Traced through `ToolService._coerce_to_tool_result` (the path `invoke_tool`'s `is_direct_proxy` branch at `:6853-6854` uses): the resulting `structured_content` is also `None`.

This silently re-breaks review 2's finding #2, but only for `gateway_mode="direct_proxy"` SSE/StreamableHTTP invocations. The non-direct-proxy branch happens to survive by accident, via an unrelated pre-existing fallback at `:6862` (`dump.get("structuredContent") or dump.get("structured_content")`) — which is exactly the branch the PR's two new tests exercise, so they pass while the direct-proxy branch stays broken.

Fix: `types.CallToolResult(..., structuredContent=structured_content)`.

### Should-fix (non-blocking)

**3.** PR checklist claims `- [x] Documentation updated (ROOT_CAUSE_ANALYSIS_GATEWAY_BUG.md created)` — no such file exists in the diff or repo. Flagged in review 1 on Aug 17, still open a month later.

**4.** No issue number referenced (review 1 cited a tracking issue). `api` label still attached despite `mcp-protocol` already covering this work — review 1 asked to drop it.

**5.** `tool_service.py:5016-5017` — `_run_timeout_post_invoke` docstring still says "otherwise control returns to the caller which raises a plain `ToolTimeoutError`." No longer true for the SSE/StreamableHTTP callers after this PR's change (see Finding 1).

### Test coverage gap

The two new tests (`TestInvokeToolMcpSseRetryOnStatus`, `TestInvokeToolMcpStreamableHttpRetryOnStatus`) only exercise the non-direct-proxy branch and assert `invoke_hook.assert_awaited()` (at-least-once) rather than `assert_awaited_once()`, so neither Finding 1 nor Finding 2 is caught by the current suite. Please add:
- an assertion that `TOOL_POST_INVOKE` fires exactly once on an unretried MCP timeout (both transports)
- a direct-proxy `HTTPStatusError` case asserting `structured_content == {"status_code": 429}` survives `_coerce_to_tool_result`

### Other checklist items (no issue found)

No Alembic migration touched or needed — no schema/model change. Backend-only, no UI. No dead code introduced. Checked all callers of `invoke_tool_direct`/`invoke_tool` for code outside this PR that catches `ToolInvocationError`/`ToolTimeoutError` from these specific runtime paths — none exist; blast radius is contained to this PR's own callers.

bogdanmariusc10:
Thanks for the re-review, @msureshkumar88. Here is what each finding got:

**Blocking #1 (double TOOL_POST_INVOKE on SSE/StreamableHTTP timeout)**
Restored `raise ToolTimeoutError(...)` on both paths. The `return self._make_mcp_tool_error(...)` that replaced it fell through into the post-process block and fired `TOOL_POST_INVOKE` a second time. The outer `except ToolTimeoutError` handler at line 7174 is now the sole post-invoke callsite for the timeout path, matching the gRPC path and the documented invariant.

**Blocking #2 (`structured_content` silently dropped on direct-proxy path)**
Changed `types.CallToolResult(..., structured_content=...)` to `structuredContent=` in `_make_mcp_tool_error`. The MCP SDK field has no snake_case alias; the old kwarg was stored as an unrecognised extra and dropped by `_coerce_to_tool_result`.

**Should-fix #5 (stale docstring)**
Updated `_run_timeout_post_invoke` docstring: "control returns to the caller which raises" → "this method returns normally and the caller must raise".

**Test gaps**
- Added `TestMcpSseTimeoutPostInvokeFiresOnce` and `TestMcpStreamableHttpTimeoutPostInvokeFiresOnce`: each asserts `invoke_hook.await_count == 1` on an unretried timeout, so the double-fire regression is caught directly.
- Added `TestMakeToolErrorStructuredContentSurvivesCoercion`: unit-tests `_make_mcp_tool_error` + `_coerce_to_tool_result` end-to-end with `{"status_code": 429}`, covering the direct-proxy branch that the previous two tests missed.
- Updated two existing SSE/StreamableHTTP timeout tests that asserted the broken return-instead-of-raise behaviour; they now expect `ToolTimeoutError`.

msureshkumar88:
This branch has a merge conflict with `main` (mergeable: CONFLICTING). Please rebase onto `main` and resolve the conflict in `mcpgateway/services/tool_service.py`.

msureshkumar88:
13 unit tests fail against current \`main\` (post MCP SDK v2 migration, #6868):

- \`tool_service.sse_client\` / \`tool_service.streamablehttp_client\` no longer exist on \`main\` — tests should patch \`tool_service.mcp_proxy_client\` instead. Affects:
  - \`TestMcpSseTimeoutPostInvokeFiresOnce\`, \`TestMcpStreamableHttpTimeoutPostInvokeFiresOnce\`
  - \`TestInvokeToolMcpSseProcessSignalPropagates\`, \`TestInvokeToolMcpStreamableHttpProcessSignalPropagates\`
  - \`TestInvokeToolMcpSseRetryOnStatus\`, \`TestInvokeToolMcpStreamableHttpRetryOnStatus\`
  - \`TestInvokeToolDirectDecodeAuthException\`
  - \`TestInvokeToolDirect::test_invoke_tool_direct_error_is_sanitized\`, \`test_invoke_tool_direct_connection_error\`
- \`CallToolResult\` now uses snake_case (\`structured_content\`, \`is_error\`) instead of camelCase (\`structuredContent\`, \`isError\`) — update assertions in \`TestMakeToolErrorStructuredContentSurvivesCoercion\` (both tests).

Production code is fine (\`_make_mcp_tool_error\` still passes the camelCase kwargs into the constructor, which works). Please update the test mocks/assertions and confirm \`make test\` (or at minimum \`tests/unit/mcpgateway/services/test_tool_service.py\` + \`test_tool_service_coverage.py\`) passes locally before pushing — CI's \`pytest (py3.12)\` job will fail on this otherwise.

ja8zyjits:
---

## 🚨 Blocking Issues

### B1 — PR description contradicts timeout implementation
- **File**: `issue_5036_pr_6181.md:36,40`; diff lines 116, 129 (SSE), 181, 194 (StreamableHTTP)
- **Problem**: PR says "Timeout handler: Returns structured error response instead of raising `ToolTimeoutError`". Diff shows the opposite — `raise ToolTimeoutError(...)` at lines 129 and 194 was preserved. Only the comment changed; the raise was not removed.
- **Fix**: Correct items 2 and 3 under "Changes Made" to: `"Timeout handler: still raises ToolTimeoutError (added clarifying comment; TOOL_POST_INVOKE fires exactly once via _run_timeout_post_invoke)"`.

### B2 — `AttributeError` on `result.is_error` — wrong SDK attribute name
- **File**: `tests/unit/mcpgateway/services/test_tool_service.py:9323, 9336`
- **Problem**: `invoke_tool_direct` returns `self._make_mcp_tool_error(...)` — a raw `types.CallToolResult`. The SDK type declares `isError`, not `is_error`. Tests asserting `result.is_error is True` raise `AttributeError` at runtime. No `_coerce_to_tool_result` call on this path.
- **Fix**: Either (a) coerce in the except block — `return self._coerce_to_tool_result(self._make_mcp_tool_error(sanitized))` — or (b) change tests to `result.isError`.

### B3 — `AttributeError` on `raw.structured_content` — wrong SDK attribute name
- **File**: `tests/unit/mcpgateway/services/test_tool_service_coverage.py:10657, 10673`
- **Problem**: `TestMakeToolErrorStructuredContentSurvivesCoercion` asserts `raw.structured_content`. `raw` is `types.CallToolResult`; the SDK field is `structuredContent`. Accessing `.structured_content` raises `AttributeError`.
- **Fix**: Use `raw.structuredContent` in assertions, or coerce before asserting.

### B4 — `_make_mcp_tool_error` docstring: missing camelCase warning on `structuredContent`
- **File**: `mcpgateway/services/tool_service.py` diff hunk `@@ -3832,6 +3832,28 @@`, diff line 17
- **Problem**: Python param is `structured_content` (snake_case); constructor call correctly uses `structuredContent=structured_content`. But the `Args:` docstring omits this distinction. A future editor could "fix" the kwarg to snake_case, silently dropping the value into `model_extra` where `_coerce_to_tool_result` discards it.
- **Fix**: Add to `Args:`: `The MCP SDK constructor keyword is ``structuredContent`` — do not pass this as ``structured_content=`` or the value silently lands in ``model_extra`` and is dropped by ``_coerce_to_tool_result``.`

---

## ⚠️ Warnings

### W1 — Inline comment says "return" but code raises on timeout
- **File**: `mcpgateway/services/tool_service.py`, diff lines 116 (SSE), 181 (StreamableHTTP)
- **Problem**: `except (asyncio.TimeoutError, httpx.TimeoutException):` comment now reads "log and **return** MCP-compliant error" but the block ends with `raise ToolTimeoutError(...)`. The four-line comment block added immediately after (lines 124–129, 189–194) correctly explains why it still raises — but the first comment directly contradicts it.
- **Fix**: Change to `# Handle timeout specifically — log, run post-invoke hook, then raise ToolTimeoutError`.

### W2 — `_run_timeout_post_invoke` docstring missing `Returns:` entry
- **File**: `mcpgateway/services/tool_service.py`, diff lines 87–89
- **Problem**: Updated docstring correctly states the caller must raise, but no `Returns:` section documents the `None` return on the non-retry path. Gap introduced by this diff.
- **Fix**: Add `Returns: None` entry to the docstring.

### W3 — Test docstring contradicts test body (stale)
- **File**: `tests/unit/mcpgateway/services/test_tool_service.py:9293`
- **Problem**: Docstring says "should raise ToolInvocationError" but the new test body asserts a structured return value. Unchanged line made inconsistent by the diff.
- **Fix**: Update to `"""Connection failure in invoke_tool_direct returns a structured MCP error response."""`

### W4 — Fragile `str(result.content[0])` assertion — three instances
- **Files**:
  - `tests/unit/mcpgateway/services/test_tool_service.py:255` (diff line 255)
  - `tests/unit/mcpgateway/services/test_tool_service_coverage.py:369` (diff line 369)
  - `tests/unit/mcpgateway/services/test_tool_service_coverage.py:706` (diff line 706)
- **Problem**: `str()` on a Pydantic model returns `repr`, not `.text`. Assertions pass incidentally and are not stable if MCP SDK changes `TextContent.__str__`. The sanitization test at diff line 332 correctly uses `.text` — inconsistent within the same diff.
- **Fix**: Replace all three with `result.content[0].text` (or `raw.content[0].text`).

### W5 — Bare `except Exception` silences decryption failures without traceback
- **File**: `mcpgateway/services/tool_service.py:3937`
- **Problem**: `except Exception: # noqa: S110` + `logger.debug(...)` without `exc_info=True` loses the traceback. Diagnosing misconfigured gateways becomes harder.
- **Fix**: `except Exception as _err:` and add `exc_info=True` to the debug log, or log at `WARNING` level.

### W6 — Missing test: `_make_mcp_tool_error` message prefix never asserted
- **File**: `mcpgateway/services/tool_service.py:3852` (impl); `test_tool_service_coverage.py:651–677` (tests)
- **Problem**: Implementation prefixes every error with `"MCP server error: "`. `TestMakeToolErrorStructuredContentSurvivesCoercion` tests `structured_content` round-trip but never asserts the wire format of `content[0].text`.
- **Fix**: Add `assert raw.content[0].text == "MCP server error: Too Many Requests"` to lock the client-facing wire format.

---

msureshkumar88:
Status of the blocking items:

- B1: Fixed. PR description now states the timeout handlers still raise `ToolTimeoutError` (retry path), and the "Key Point" line no longer claims all failures return `CallToolResult`.
- B2/B3: Not a bug on this branch. The repo pins `mcp>=2.0.0` (installed 2.2.0). In SDK v2, `CallToolResult` fields are `is_error` / `structured_content`; `isError` / `structuredContent` are aliases for construction and wire output only. `result.isError` raises `AttributeError`. Both flagged tests pass.
- B4: Same root cause. `CallToolResult(structured_content=...)` populates the field in SDK v2; `model_extra` stays empty, so no silent drop.

Warnings W1–W6 noted; will address separately.

ja8zyjits (approved):
LGTM

PR #1412, opened 2025-11-10
Fix:Missing Structured Content for Virtual Server in Streamable HTTP Response

rakdutta:
closing issue https://github.com/IBM/mcp-context-forge/issues/1406
This pull request introduces improvements to the handling and processing of tool responses, particularly for structured content, and enhances logging for debugging and traceability. The changes ensure that both unstructured and structured outputs from tools are correctly normalized and returned, and that the code is robust against different field naming conventions and model types. 
**Tool response handling and normalization:**

* Updated `call_tool` in `streamablehttp_transport.py` to return both unstructured and structured content when available, supporting both snake_case and camelCase field names for structured content. This allows downstream validation against output schemas.
* In `tool_service.py`, enhanced extraction of structured content from tool responses by accepting both `structuredContent` and `structured_content` fields, improving compatibility with different model serialization formats.

**Unit test robustness:**

* Updated unit tests in `test_streamablehttp_transport.py` to explicitly unset `structured_content` and override `model_dump` on mocks, preventing accidental attribute leakage and ensuring test accuracy.

PR #6182, opened 2026-08-11
fix(tools): preserve empty dict in structuredContent handling

### LINKED ISSUE #6274: structuredContent empty-dict handling — missing test coverage and identical pattern in streamablehttp_transport
## Summary

PR #6182 fixes a real bug where the `or` operator was dropping valid `structuredContent: {}` responses (Python treats empty dicts as falsy). However, the PR is incomplete in two ways that should be addressed before or alongside merge.

**Tracking issue for PR #6182.**

---

## Finding 1 — Missing regression test (blocking)

The fix in `tool_service.py` at line 6409 is correct, but the PR does not include a unit or integration test that would:

1. Exercise the non-direct-proxy code path in `connect_to_streamablehttp_server()` with `structuredContent: {}`.
2. Assert that `tool_result.structured_content` is `{}` (not `None`) after the fix.

Without this test the bug can silently regress. A minimal test should mock a `CallToolResult` whose `model_dump()` returns `{"structuredContent": {}, "content": [...], "isError": False}` and verify the resulting `ToolResult.structured_content == {}`.

**Acceptance criteria:** at least one test in `tests/unit/mcpgateway/services/test_tool_service.py` (or an appropriate integration test) explicitly covers `structuredContent: {}` being preserved through the non-direct-proxy path.

---

## Finding 2 — Identical pattern exists in streamablehttp_transport (functionally-impacting, needs assessment)

A second `or`-based lookup exists at:

```
mcpgateway/transports/streamablehttp_transport.py:1890
structured = result_data.get("structuredContent") or result_data.get("structured_content")
```

This line is immediately followed by:

```python
if not isinstance(structured, dict):
    structured = None
```

Because `isinstance({}, dict)` is `True`, the guard *does* preserve `{}` in this file — so the transport is not currently broken. However, the pattern is still technically incorrect (it will skip the camelCase key and fall through to the snake_case key when `structuredContent` is `{}`) and should be harmonised for clarity and future safety. The PR description does not mention this location.

**Acceptance criteria:** either confirm the guard makes the behaviour correct and add a comment explaining why, or apply the same explicit-`None`-check fix for consistency.

---

## Steps to Reproduce (Finding 1 regression scenario)

1. Register a Hubbl-compatible MCP tool with `outputSchema` defined.
2. Invoke the tool through the non-direct-proxy path.
3. Tool returns `{"structuredContent": {}, "isError": false, "content": [...]}`.
4. **Before fix:** `RuntimeError: Tool has an output schema but did not return structured content`.
5. **After fix (without test):** works, but the regression is not caught by CI.

---

## References

- PR #6182 — original fix
- Introduced in PR #1412 / commit `f6ab27024` (2025-11-12)
- Affected file: `mcpgateway/services/tool_service.py` line 6409
- Related file: `mcpgateway/transports/streamablehttp_transport.py` line 1890

bogdanmariusc10:
# Pull Request

Fixes bug where MCP tools returning `structuredContent: {}` fail with RuntimeError

**Background:** Hubbl tools (and potentially other MCP servers) return `"structuredContent": {}` (empty dict) as a valid MCP response for tools with `outputSchema`. The gateway was incorrectly discarding this empty dict, causing the MCP SDK to fail validation.

---

## 📝 Summary

**Problem:** Line 6404 in `tool_service.py` used Python's `or` operator:
```python
structured = dump.get("structuredContent") or dump.get("structured_content")
```

In Python, empty dicts are falsy (`bool({}) == False`), so the `or` operator would skip `{}` and fall through to `dump.get("structured_content")` which returns `None`. This caused tools with `outputSchema` to fail with: `RuntimeError: Tool has an output schema but did not return structured content`.

**Solution:** Changed to explicit `None` check:
```python
structured = dump.get("structuredContent")
if structured is None:
    structured = dump.get("structured_content")
```

This preserves empty dicts (valid responses) while still supporting the snake_case fallback.

**Impact:** 
- Fixes Hubbl tools and any other MCP servers that return `structuredContent: {}`
- Actual tool data comes through in `content` field; the empty dict is just the required MCP wrapper
- Bug was introduced in PR #1412 (commit f6ab27024) on 2025-11-12

---

## 📏 Reviewability

---

## 🏷️ Type of Change

- [x] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

### Root Cause Verification

Python behavior confirming the bug:
```python
>>> bool({})
False
>>> {} or "fallback"
'fallback'  # Empty dict was incorrectly discarded

>>> structured = {}
>>> structured if structured is not None else "fallback"
{}  # Correctly preserved
```

### Code Location

**File:** `mcpgateway/services/tool_service.py`
**Lines:** 6405-6407 (after fix)
**Scope:** MCP tool invocation path (non-direct-proxy mode)

### Manual Testing

**Before fix:**
```
Tool call result dump: {'structuredContent': {}, 'isError': False, 'content': [...]}
Final tool_result: structured_content=None
→ RuntimeError: Tool has an output schema but did not return structured content
```

**After fix:**
```
Tool call result dump: {'structuredContent': {}, 'isError': False, 'content': [...]}
Final tool_result: structured_content={}
→ Success: MCP SDK validation passes
```

### Test Commands

| Check                     | Command                                      | Status |
|---------------------------|----------------------------------------------|--------|
| Lint suite                | `make ruff interrogate pylint`               | ✅     |
| Unit tests                | `make test`                                  | ✅     |
| Coverage ≥ 80%            | `make coverage diff-cover`                   | ✅     |
| E2E MCP protocol          | `make test-mcp-protocol-e2e`                 | ✅     |

---

## ✅ Checklist

---

### Why This Matters

1. **MCP Spec Compliance:** Empty dicts are valid MCP responses that satisfy the protocol requirement for tools with `outputSchema`
2. **Hubbl Integration:** Hubbl tools fail without this fix because they correctly return `structuredContent: {}` per MCP spec
3. **Data Preservation:** The actual response data is present in the `content` field; the empty dict is just the required structured wrapper


bogdanmariusc10:
Hi @msureshkumar88! Can you take a look into this?

jonpspri:
Tracking issue created: #6274 — covers the missing regression test and the identical pattern in `streamablehttp_transport.py`.

bogdanmariusc10:
Thanks, @jonpspri! Rebased and addressed the findings from Issue https://github.com/IBM/mcp-context-forge/issues/6274:

## Finding 1 - Regression Test ✅
Added `test_invoke_tool_mcp_preserves_empty_dict_structured_content()` in `tests/unit/mcpgateway/services/test_tool_service.py` (lines 3844-3936). The test mocks a `CallToolResult` with `structuredContent: {}` and asserts it's preserved through the non-direct-proxy path.

## Finding 2 - Transport Pattern ✅
Harmonized the pattern in `mcpgateway/transports/streamablehttp_transport.py` (lines 1890-1893) to use explicit None checks instead of the `or` operator, matching the fix in `tool_service.py` for consistency.

Note: The failing test is unrelated to my changes.

msureshkumar88:
## SDK v2 branch overlap — findings and recommendation

Checked this fix against `chore/mcp-sdk-v2` (HEAD `7043d20`, includes #6808 legacy-protocol-negotiation defaults). Two files this PR touches are also rewritten there as part of the `mcp` → `mcp_types` SDK migration.

### Findings

1. `mcpgateway/services/tool_service.py:6894` (sdk-v2 branch) still has the unfixed pattern:
   `structured = dump.get("structuredContent") or dump.get("structured_content")`
   Same empty-dict-is-falsy bug this PR fixes on `main`. Not carried over by the SDK-v2 rewrite.

2. `mcpgateway/transports/streamablehttp_transport.py:1996` (sdk-v2 branch) has the same pattern, at a shifted location (`result_data.get("structuredContent") or result_data.get("structured_content")`).

3. `mcpgateway/services/tool_service.py:7146` (sdk-v2 branch) already uses a safe ternary (`if key in dict else`), not `or` — no action needed there.

4. Tracking issue #6274 (missing regression test + identical `streamablehttp_transport.py` pattern) is fully addressed by this PR's current commits — both fixed, test added.

### Why merge to `main` now, ahead of `chore/mcp-sdk-v2`

- This PR fixes a reachable production bug (`RuntimeError: Tool has an output schema but did not return structured content`) on the current shipping code path. `chore/mcp-sdk-v2` is an unmerged, actively-shifting migration branch (protocol-negotiation defaults changed as recently as today) with no fixed merge date.
- Blocking a live bug fix on an open-ended SDK migration leaves the bug reachable in production for no benefit.
- The fix does not semantically depend on protocol version or SDK generation — it is a response-parsing correctness fix, independent of the legacy/modern MCP handshake work in #6808.
- Rebasing this PR onto `chore/mcp-sdk-v2` instead would pull an unrelated 1900-line SDK/protocol rewrite into a 2-line bugfix PR — out of scope per this PR's stated single purpose.

### Follow-up needed (tracked separately, not blocking this PR)

Once this merges, the same two-line fix needs porting to `chore/mcp-sdk-v2` at the line numbers above, since the SDK migration is a full-file rewrite rather than a linear diff — this PR will not apply cleanly by rebase/merge, and the bug will silently resurface if it is skipped.

msureshkumar88 (changes requested):
## Verified: fix is incomplete, reproduces the same bug it claims to fix

Confirmed by running the actual `call_tool()` against this PR's HEAD (`47db8750`), not just reading the diff. Repro:

```
[forwarded path]  structuredContent={} -> returns bare list  (BUG, dropped)
[primary path]    structuredContent={} -> returns bare list  (BUG, dropped)
[control, non-empty dict] -> returns tuple correctly          (harness sound)
```

### Root cause

`tool_service.py` and the two hunks this PR patches in `streamablehttp_transport.py` (lines ~1943-1947, ~1900-1902) correctly switch to an explicit `is None` check. But `call_tool()`'s final response-shaping — a few lines below each patched hunk — still gates on plain truthiness:

- `mcpgateway/transports/streamablehttp_transport.py:1965` (forwarded/session-affinity path):
  ```python
  if structured:
      return (unstructured, structured)
  return unstructured
  ```
- `mcpgateway/transports/streamablehttp_transport.py:2126` (primary/non-forwarded path): identical pattern.

`{}` is falsy, so both branches drop the preserved-but-empty `structured` dict and return a bare list. The MCP SDK (`mcp/server/lowlevel/server.py:558-569`) then sees `maybe_structured_content = None` on a bare-list return and raises:

```
RuntimeError: Tool has an output schema but did not return structured content
```

— the exact error this PR sets out to fix, just reproduced one layer downstream.

### Test coverage gap

Repo-wide, the only test covering `structuredContent: {}` is this PR's own new test in `test_tool_service.py`, which exercises `ToolService.invoke_tool` only. `test_streamablehttp_transport.py` has no test with an empty-dict `structured_content` — all existing cases use `None` or a non-empty dict — so this gap ships without a failing test.

### Requested changes (blocking)

1. `streamablehttp_transport.py:1965` — `if structured:` → `if structured is not None:`
2. `streamablehttp_transport.py:2126` — same fix
3. Add a regression test exercising `call_tool()` directly (not just `ToolService.invoke_tool`) with `structured_content={}`, covering both the forwarded and primary paths.

Note: the error-path branches (`types.CallToolResult(..., structuredContent=structured, isError=True)` at 1957 and 2117) already pass `structured` through directly and are unaffected — this is a success-path-only regression.

bogdanmariusc10:
Thanks for the review, @msureshkumar88!

Addressed in 52232c9:

1. Changed both success-path `if structured:` gates to `if structured is not None:` — the primary path (~line 2126) and the forwarded/session-affinity path (~line 1965).
2. Added two regression tests that exercise `call_tool()` directly with `structured_content={}`: one for the primary path (`test_call_tool_empty_dict_structured_content_primary_path`) and one for the forwarded path (`test_call_tool_session_affinity_forwarded_empty_dict_structured_content`).

The error-path branches (`isError=True`) were already passing `structured` through directly and required no change.

msureshkumar88 (dismissed):
## Review: fix looks correct, verified empirically

Verified both rounds of prior review feedback actually landed in the diff (not just claimed as resolved):
- Round 1 (#6274): regression test added, explicit `is None` check applied in `streamablehttp_transport.py` extraction.
- Round 2 (success-path truthy gates): confirmed fixed in `52232c947`. Reverted the fix locally and reran the new forwarded-path test — it failed without the fix, confirming the test is a real regression guard, not decorative.

No blocking issues. Approving.

### Optional follow-ups (non-blocking)

1. Add `Closes #6274` to the PR description — issue #6274 is still open and won't auto-close on merge otherwise.
2. PR description is stale: cites `tool_service.py` lines 6405-6407 (now 6794-6797) and doesn't mention the `streamablehttp_transport.py` success-path fixes, which are roughly half the final change.
3. `chore/mcp-sdk-v2` still has the unfixed pattern at 4 sites (`tool_service.py:6894`, `streamablehttp_transport.py:1996/2016/2181` as of HEAD `d06501dcc`) and nothing currently tracks porting this fix there. Worth opening an issue so it doesn't silently regress when that branch merges.
4. `tests/live_gateway/e2e/test_e2e.py` covers non-empty structuredContent but not the empty-dict case this PR fixes — would need an upstream fixture-image change (fast-time-server), not just a test addition.

tests/unit/mcpgateway/transports/test_streamablehttp_transport.py
```diff
@@ -565,6 +565,49 @@ async def fake_get_db():
...
+    mock_db = MagicMock()
+    mock_result = MagicMock()
+    mock_content = MagicMock()
+    mock_content.type = "text"
+    mock_content.text = "ok"
+    mock_content.annotations = None
+    mock_content.meta = None
+    mock_result.content = [mock_content]
+    mock_result.is_error = False
+    mock_result.structured_content = {}
```
ja8zyjits, line 591: the fallback `structuredContent` is absent here

We need a fallback test for that something in terms of `test_call_tool_session_affinity_forwarded_empty_dict_snake_case_fallback()`

tests/unit/mcpgateway/services/test_tool_service.py
```diff
@@ -3927,6 +3927,96 @@ async def mock_streamable_client(*_args, **_kwargs):
...
+
+        # Mock MCP response with empty dict in structuredContent
+        call_result = MagicMock()
+        call_result.is_error = False
+        call_result.isError = False
+        call_result.content = [TextContent(type="text", text="tool output")]
+        call_result.model_dump.return_value = {
+            "content": [{"type": "text", "text": "tool output"}],
+            "isError": False,
+            "structuredContent": {},  # Empty dict - must be preserved
```
ja8zyjits, line 3990: here we are setting `structured_content` as `None` so the `dump.get("structured_content")` fallback leg at `tool_service.py:6795-6797` is not exercised

- test case `test_invoke_tool_mcp_preserves_empty_dict_snake_case_fallback()` would be nice

ja8zyjits (changes requested):

ja8zyjits (dismissed):
LGTM

ja8zyjits (approved):
LGTM

msureshkumar88 (approved):
## Re-review at `43541e13c`: all requested changes are resolved

I checked each earlier request against the code at HEAD. I also ran the new tests.

**My blocking requests (review of 2026-09-21)**
1. `streamablehttp_transport.py:2019` (forwarded path) now uses `if structured is not None:`. Resolved.
2. `streamablehttp_transport.py:2184` (primary path) now uses `if structured is not None:`. Resolved.
3. The new `call_tool()` regression tests cover the primary path, the forwarded path, and the error path, all with `structured_content={}`. Resolved.

**Inline requests from @ja8zyjits**
- `test_call_tool_session_affinity_forwarded_empty_dict_snake_case_fallback` sends only `structured_content`, so it runs the fallback branch. Resolved.
- `test_invoke_tool_mcp_preserves_empty_dict_snake_case_fallback` covers the `dump.get("structured_content")` branch in `tool_service.py`. Resolved.

**Verification**
- All 6 new tests pass at HEAD.
- I ran the tests against the `main` source of both files. 5 of the 6 tests fail, so they guard against the regression. The `tool_service` snake_case test passes on `main` too, because `None or {}` already gave `{}` there. That test adds branch coverage, which is what the inline request asked for.
- CI: 32 checks pass and 5 are skipped. The PR now closes #6274.

**Non-blocking follow-ups (you can do these after merge)**
- The PR description still gives `tool_service.py` lines 6405-6407. It does not mention the `streamablehttp_transport.py` success-path fixes.
- Port this fix to `chore/mcp-sdk-v2` so the bug does not come back when that branch merges.

Approving.

PR #6183, opened 2026-08-11
fix: require unrestricted admin scope on the remaining global-record routes

### LINKED ISSUE #6134: [BUG]: Team-narrowed admin tokens bypass Layer 1 on admin routes guarded by require_permission

msureshkumar88:
# 🐛 Bug-fix PR

## 📌 Summary

Team-narrowed and public-only admin tokens could bypass Layer 1 visibility scoping on
64 global-record admin routes across 9 routers (`llm_config_router`, `llm_admin_router`,
`observability`, `siem`, `sso`, `log_search`, `runtime_admin_router`, `toolops_router`,
`rbac`) plus the 4 dependency-guarded routes in `metrics_maintenance`. Because
`_get_user_roles()` always includes global-scope roles regardless of `token_teams`, any
route relying solely on `require_permission(...)` for authorization let a narrowed
(team-scoped) or claim-less admin token through, even though these endpoints operate on
platform-wide, team-less records.

This PR closes the remaining route-level gap by stacking a new
`require_global_admin_permission()` decorator (and, for `metrics_maintenance`'s
dependency-only router, a new `require_global_admin_scope_dep` FastAPI dependency) above
the existing `require_permission(...)` checks on all 64 routes, so only unrestricted
platform-admin tokens (`token_teams is None` + DB `is_admin`) can reach them.

**Depends on #6132** (`fix/global-record-admin-scope`, not yet merged) — this branch is
stacked on `origin/fix/global-record-admin-scope` because #6132 supplies the shared guard
infrastructure this fix builds on (`_global_scope_denied()`, the `GLOBAL_ONLY_DEFERRED`
manifest, `tests/helpers/scope.py`). Until #6132 merges to `main`, this PR's diff includes
#6132's own commits, and this PR should not be merged before it. CI on this PR may
surface failures inherited from #6132's own unresolved state against `main`; those are
not addressed here.

## 🔁 Reproduction Steps

Closes #6134.

Before this fix, a team-narrowed admin token could read/act on global-record admin
routes it should not have visibility into, e.g.:

```python
# Token minted with is_admin=True but teams=["team-a"] (narrowed, not unrestricted)
resp = client.get("/observability/stats", headers=narrowed_admin_headers)
assert resp.status_code == 200  # WRONG — should be 403, these are global records
```

`_get_user_roles()` includes global-scope roles for any admin regardless of
`token_teams`, so `require_permission("observability.read")` alone could not tell a
narrowed admin from an unrestricted one — only a route-level guard checking
`token_teams is None` explicitly can.

## 🐞 Root Cause

Route-level authorization on these 64 routes relied solely on `require_permission(...)`
(Layer 2: "can this role do X"), with no route-level Layer 1 check for "is this a
global, team-less record." `require_permission` grants based on role permissions, which
narrowed admin tokens still hold — it has no visibility into `token_teams` narrowing at
all. PR #6132 introduced the guard primitive and migrated 13 routes but explicitly
deferred the remaining 64 into a `GLOBAL_ONLY_DEFERRED` manifest with no route-level
enforcement wired up yet.

## 💡 Fix Description

- Added `require_global_admin_scope_dep(request, db=Depends(get_db), user=Depends(get_current_user_with_permissions))`
  in `mcpgateway/middleware/rbac.py` — a `Depends()`-form guard for routers (like
  `metrics_maintenance`) that only mount dependencies at the router level, not a
  decorator per endpoint. Both this and `require_global_admin_permission()` delegate to
  the same `_global_scope_denied()` policy point introduced in #6132.
- Stacked `@require_global_admin_permission()` above the existing
  `@require_permission(...)` decorator on all 60 decorator-form routes across
  `llm_config_router.py`, `llm_admin_router.py`, `observability.py`, `siem.py`, `sso.py`,
  `log_search.py`, `runtime_admin_router.py`, `toolops_router.py`, and `rbac.py`
  (`check_permission`, `get_user_permissions`). Endpoints missing a `request` kwarg (the
  guard resolves scope from `request.state.token_teams`) got `request: Request = None`
  added, each with a docstring `Args:` line documenting it.
- Mounted `Depends(require_global_admin_scope_dep)` alongside the existing
  `Depends(require_admin_auth)` on `metrics_maintenance.py`'s router-level
  `dependencies=[...]` for its 4 routes.
- Retired the `GLOBAL_ONLY_DEFERRED` tier in
  `tests/unit/mcpgateway/test_global_record_scope.py` entirely — all 64 routes moved
  into `GLOBAL_ONLY`, `GLOBAL_ONLY_DEFERRED` is now a permanently-empty set with a
  regression test (`test_deferred_bucket_is_retired`) guarding against reintroduction.
  Extended `_scope_class()` to recognize the dependency form by object identity
  (`dep.dependency is require_global_admin_scope_dep`) rather than by name, with a decoy
  test proving name-based matching would have been spoofable.
- Added `scripts/verify_global_guard.py`, an AST-based audit script reporting
  guarded/unguarded counts across the 9 routers (final state: 60/0/0/0).
- Two crash bugs found and fixed while adding the `request` parameter:
  `log_search.search_logs` and `sso.handle_approval_request` both already had a
  Pydantic body parameter literally named `request`, which collided with the new
  `request: Request` kwarg. Renamed the body params (`search_logs` → `body`,
  `handle_approval_request` → `approval_request`) to resolve the collision.
- Documentation: `docs/docs/manage/rbac.md` gained a "Global-record routes" section
  covering both enforcement mechanisms and corrected classification-summary counts
  (72 global-only routes, 0 deferred). `AGENTS.md`'s Security Invariants section now
  names all three enforcement mechanisms (`require_global_admin_permission()`,
  `require_unrestricted_platform_admin()`, `Depends(require_global_admin_scope_dep)`).
- New live end-to-end verification suite:
  `tests/live_gateway/admin_rbac/test_issue_6134_narrowed_admin_scope.py` — see
  Verification below.

### ⚠️ Breaking changes

- Team-narrowed and claim-less (public-only) admin tokens are now denied (403) on all
  64 previously-deferred global-record routes. This is the point of the fix, but any
  caller currently relying on a narrowed admin token to reach these routes will need an
  unrestricted platform-admin token instead.
- A side effect of Layer 2 stacking not called out in the issue: delegated
  (non-platform-admin) roles holding the underlying `require_permission` permission are
  now also denied on these routes, since `require_global_admin_permission()` requires DB
  `is_admin` or the global `*` role.
- `metrics_maintenance`'s routes previously accepted any caller passing
  `require_admin_auth` (which supports HTTP Basic auth); the new
  `require_global_admin_scope_dep` only evaluates bearer-token scope, so a Basic-auth
  caller without an unrestricted bearer token now gets denied even if `require_admin_auth`
  itself would have passed.

## 📏 Reviewability

- [ ] The linked issue is not labeled `triage` (#6134 currently carries `triage`)

## 🧪 Verification

Config used for live e2e verification:

```bash
# .env additions (on top of `make setup`-generated JWT_SECRET_KEY / BASIC_AUTH_PASSWORD)
OBSERVABILITY_ENABLED=true
SIEM_EXPORT_ENABLED=true
TOOLOPS_ENABLED=true
SSO_ENABLED=true
```

```bash
python scripts/verify_global_guard.py                 # 60/0/0/0, exit 0
make ruff interrogate pylint                           # clean
make test                                               # unit suite green (see note)
make coverage diff-cover                                # diff coverage on changed lines
make testing-up RUST_MCP_MODE=                          # live gateway stack
pytest tests/live_gateway/admin_rbac/test_issue_6134_narrowed_admin_scope.py -v
```

| Lint suite                            | `make ruff interrogate pylint` | ✅ clean |
| Unit tests                            | `make test`          | ✅ green (12 pre-existing otel StaticPool/DB-timing failures under full `-n auto` parallel load are environment-specific and unrelated to this change; not observed when the same files run in isolation) |
| Coverage of changed lines             | `make coverage diff-cover` | ✅ 76% overall diff vs `main`; every missed line belongs to files that are #6132's own pre-existing diff (not introduced here) — every changed file/function in this PR's own scope (9 routers + `rbac.py`'s new function) is at 100% |
| Live e2e verification                 | `tests/live_gateway/admin_rbac/test_issue_6134_narrowed_admin_scope.py` | ✅ 24 passed, 5 skipped (skips are flag-gated routers not mounted under the `testing` docker-compose profile's fixed `OBSERVABILITY_ENABLED=false`, documented in the test module docstring) |

The e2e suite runs against a real `testing-up` docker stack (not `TestClient`/mocks),
mints three real JWTs (unrestricted admin, a team-narrowed admin against a freshly
provisioned real team, and a public-only token), and asserts unrestricted → 200,
narrowed → 403, public-only → 403 across a representative route from every migrated
router plus `metrics_maintenance`. It also hits neighboring, already-migrated routes
(`/compliance/frameworks` from #6132) and an ordinary team-scoped route (`/tools`) to
confirm no regression to surrounding behavior.

## ✅ Checklist


msureshkumar88:
## Live e2e verification for issue #6134

Ran a new, real (non-mocked) end-to-end suite against a live `testing-up` Docker
gateway stack to empirically confirm the fix: `tests/live_gateway/admin_rbac/test_issue_6134_narrowed_admin_scope.py`.

### Config used

```bash
# .env additions on top of `make setup`-generated JWT_SECRET_KEY / BASIC_AUTH_PASSWORD
OBSERVABILITY_ENABLED=true
SIEM_EXPORT_ENABLED=true
TOOLOPS_ENABLED=true
SSO_ENABLED=true
```

Postgres bootstrap admin's `password_change_required` flag cleared for the run (a
pre-existing, unrelated security feature that otherwise 303-redirects even bearer-token
API callers) — documented as a prerequisite in the test module's docstring rather than
baked into the script.

### Commands run

```bash
make docker-prod-rust testing-up RUST_MCP_MODE=
pytest tests/live_gateway/admin_rbac/test_issue_6134_narrowed_admin_scope.py -v
```

### Output

```
24 passed, 5 skipped
```

The suite mints three real JWTs against the live gateway (unrestricted admin,
a team-narrowed admin against a freshly provisioned real team, and a public-only
token) and asserts, per router:

- **Unrestricted admin → 200** (or a documented non-403 status where the handler
  needs backing data not provisioned in this stack)
- **Narrowed admin → 403** with `"requires an unrestricted platform-admin token"`
- **Public-only → 403** (either the same guard, or `TokenScopingMiddleware`'s
  independent `/admin/*` default-deny for the 3 routers where that layer fires first —
  both accepted as correct denials)

It also hits neighboring routes to confirm nothing nearby regressed:
`/compliance/frameworks` (already `GLOBAL_ONLY` from #6132 — narrowed → 403,
unrestricted → 200) and `/tools` (an ordinary team-scoped route, unaffected by this
change).

The 5 skips are the `observability`/`siem`/`toolops` route-matrix entries — those
routers aren't mountable under the `testing` docker-compose profile's hardcoded
`OBSERVABILITY_ENABLED=false` (a pre-existing infra limitation of that profile, not
this PR); the suite detects the 404 and skips those cases rather than failing, with
the reason logged in the skip message.

This branch has since been rebased onto current `main` (only `.secrets.baseline`
conflicts, resolved by regenerating the baseline — no application code touched by the
rebase), and the full CI suite is green on the rebased head.

jonpspri:
## Review — in-depth pass

Thanks for this one; it's a careful piece of work and the hard parts are done well. Before the findings, the things I went looking for and found solid:

- I introspected the running app rather than trusting the decorators by eye, and the wiring holds up: **68 decorator-guarded endpoints + 4 dependency-guarded**, FastAPI sets `request_param_name` on **100%** of them (so `request` is genuinely injected, not silently `None`), every one exposes a user kwarg the guard can find, and **zero** of those user kwargs are client-controllable — all are `Depends(...)`. No escalation vector via request body, query, path, or header.
- `request: Request = None` is not overridable via `?request=…` and doesn't leak into the OpenAPI schema — worth stating explicitly since it looks alarming at a glance.
- Manifest integrity checks out: `GLOBAL_ONLY ∪ GLOBAL_ONLY_DEFERRED` is **72 before and 72 after**, with zero routes quietly dropped during the move.
- Decorator ordering is right (scope guard before permission check), and the two body-param renames are wire-invisible — single Pydantic body param means no `embed`, so no API break.
- The `_global_scope_denied` single-evaluation-point design, identity-based (not name-based) dependency matching in the drift guard with the decoy test proving why, and the live e2e suite that provisions a real team and mints three real JWTs are all genuinely above the usual bar.
- `ruff`, `ruff format`, and `interrogate` (100%) are clean, and pylint on the ten changed modules actually nudges **up**, 9.49 → 9.50.

Findings below, categorized. Only the first one I'd call blocking.

---

### Blocking

**1. Seven doctest regressions in `mcpgateway/routers/observability.py`**

`list_traces`, `query_traces_advanced`, `get_trace`, `list_spans`, `cleanup_old_traces`, `export_traces`, and `get_query_performance` each call their endpoint directly with no `request`, so the new guard denies them:

```
Expected: (400, True)
Got:      (403, False)
```

I checked this against a worktree at the pre-PR base to be sure it wasn't pre-existing — **base: 7 passed, head: 7 failed**. This breaks `make doctest`, and `make coverage` (validation gate #3) also runs `--doctest-modules mcpgateway/`. My guess is the gates were run before the observability commit landed. Should be a small fix: pass a stub request carrying `state.token_teams = None` — the `tests/helpers/scope.scoped_request` shape — into each doctest call, or patch `is_unrestricted_platform_admin` for those examples.

---

### Functionally impacting

**2. The `/api/metrics/*` Basic-auth break is real, and two docs still promise it**

Reproduced with `API_ALLOW_BASIC_AUTH=true`, `AUTH_REQUIRED=true`, `EMAIL_AUTH_ENABLED=true`:

| | `GET /api/metrics/config` with `Authorization: Basic …` |
|---|---|
| pre-PR base | `200` |
| this branch | `401 {"detail":"Authorization token required"}` |

The PR description calls this out, which I appreciate — but `docs/docs/manage/configuration.md:150` and `docs/docs/architecture/adr/004-combine-jwt-and-basic-auth.md:212` both still name `/api/metrics/*` as *the* example of what `API_ALLOW_BASIC_AUTH` enables, so they'd become actively wrong on merge. Two options depending on intent: update both docs plus a `MIGRATION.md`/release note, or keep Basic working on that router by resolving the scope from `require_admin_auth`'s own result rather than re-authenticating through `get_current_user_with_permissions`.

Separately, the failure surfaces as a bare `401 Authorization token required` with no mention of the scope guard, so an operator hitting it has no path to the remediation the PR documents.

Good news: the Admin UI is unaffected — `admin_ui/components/maintenance-panel.js` sends `Bearer <jwt_token cookie>`, not Basic.

**3. Proxy-auth platform admins are now denied on all 60 decorator routes**

`mcpgateway/middleware/rbac.py:344-356` returns the proxy user context without setting `request.state.token_teams`, so `get_token_teams_from_request` falls through to the `[]` secure default at `mcpgateway/auth_context.py:723`, which reads as public-only → 403.

Worth noting the *other* proxy path, `verify_credentials._resolve_proxy_payload:864-899`, does set `_jwt_verified_payload` with `teams: None` — so the two proxy entry points now disagree with each other. This gap arrived with #6132 rather than here, but this PR takes it from 13 routes to 73.

**4. `allow_unauthenticated_admin` local-dev mode is denied too**

Same root cause, at `rbac.py:483-495` — the synthetic platform-admin dict carries no `token_teams` and leaves request state untouched. It fails closed, so this isn't a security concern, but it does silently retire a supported local-only escape hatch. Probably worth an explicit decision plus a release note either way.

> Findings 3 and 4 share one cause: `token_teams` reaches `request.state` from three different places and "not at all" is one of them. A single `ensure_token_teams(request)` helper called by every auth entry point would make this class of bug structural rather than something each new guarded route can rediscover. Probably a separate issue rather than scope creep here.

---

### Suggestions

**5. Strict typing regresses by 45 errors.** basedpyright on the ten changed modules goes 27 → 72, all of them `Expression of type "None" cannot be assigned to parameter of type "Request"`. One heads-up that might save some time: **`Optional[Request] = None` is not a workable fix** — I tried it and FastAPI raises `FastAPIError: Invalid args for response field!` at import, because the annotation stops matching the `Request` special case. The clean form is `request: Request` with no default, placed before the `Depends(...)` params — which is exactly what `patch_mcp_mode` and `patch_a2a_mode` already do and why they needed no change.

**6. Would a router-level dependency have been lighter on several of these routers?** On `observability`, `siem`, `toolops_router`, `llm_config_router`, `llm_admin_router`, and `runtime_admin_router`, *every* endpoint lands in `GLOBAL_ONLY`. A single `dependencies=[Depends(require_global_admin_scope_dep)]` would cover each of those with no signature churn at all, would have sidestepped both `request` body-param collisions, and leaves future endpoints on those routers guarded by default rather than guarded-if-remembered. Only `log_search`, `sso`, and `rbac` have mixed populations that genuinely need the per-endpoint decorator. Not asking you to redo it here — but it might be a worthwhile follow-up.

**7. `scripts/verify_global_guard.py` reads as scaffolding that's served its purpose.** Its docstring describes reporting migration progress, it hardcodes `(expected 60)`, it isn't wired into any Makefile target or CI, and it re-derives via source AST what `test_global_record_scope.py` already enforces permanently against the live app. Now that the migration is at 60/0/0/0, dropping it would leave one authority instead of two.

**8. `metrics_maintenance` has the lightest coverage of the two guard forms.** `test_metrics_maintenance_scope.py` asserts only that the dependency is *mounted*, and matches on `__name__` — slightly ironic next to `test_dependency_guard_is_matched_by_identity_not_only_name`, which exists specifically to show name-matching is spoofable. No unit test drives a narrowed token through the real router; `test_metrics_maintenance.py` already has a TestClient `app` fixture that would make one fairly cheap.

**9. The denial log can point operators the wrong way.** At `rbac.py:1155`, `resolved = … if request is not None else None` means a denial caused by a *missing* request logs `token_teams=None` — which is the shape that's supposed to be allowed. It shows up in the doctest output above as `global-record scope denied: … token_teams=None`. A distinct sentinel such as `"<no request>"` would read correctly.

---

### Minor notes

- All seven new `*_scope.py` files carry `# Pattern B — see "Test Isolation Patterns" above`, but there's no such section anywhere in `tests/` or `docs/` — looks like it survived a doc that got trimmed.
- In `docs/docs/manage/rbac.md`: "Helpers reading `request.state.token_teams` (first two)" — all three read it; the dependency form just receives `request` from FastAPI rather than from the endpoint signature.
- The checklist ticks "The linked issue is not labeled `triage`", though #6134 currently is.

---

Nothing here changes my read that the underlying fix is correct and worth landing — the guard does what it says, fails closed everywhere I pushed on it, and the drift guard should keep it that way. Happy to dig into any of these further, and glad to be wrong on #3/#4 if proxy and unauthenticated-admin callers aren't paths you're supporting on these routes.

msureshkumar88:
Pushed fixes for every item except #6. Verified each independently against the codebase before touching anything — details below.

**Blocking**

1. **Doctest regressions** — confirmed (base: 7 passed, head: 7 failed, matches your numbers exactly). Fixed by injecting a stub `request` (`SimpleNamespace(state=SimpleNamespace(token_teams=None))`) into each of the 7 call sites in `observability.py`. All 7 pass now, and I ran the full `--doctest-modules mcpgateway/` sweep clean afterward.

**Functionally impacting**

2. **`/api/metrics/*` Basic-auth break** — reproduced. Went with the docs-update option rather than reworking the auth flow: `configuration.md` and the ADR now point to `/health/security` as the still-Basic-compatible example and call out that global-record routes (`/api/metrics/*`, `/api/logs/*`) need an unrestricted bearer/session token regardless of `API_ALLOW_BASIC_AUTH`. Left the generic 401 message alone — it's a shared code path for every no-token case, not specific to this router, so rewording it felt like a separate, broader change.

3. **Proxy-auth admins denied** — confirmed via code read (no `token_teams`/`_jwt_verified_payload` set anywhere in that branch). Fixed by resolving `token_teams` through `_resolve_teams_from_db` (the same canonical helper `_resolve_proxy_payload` already uses) right after `is_admin` is determined, so the two proxy entry points now agree. Added 4 regression tests covering admin, non-admin, and the resolution-failure fallback.

4. **`allow_unauthenticated_admin` denied** — confirmed, same root cause. Fixed with `request.state.token_teams = None` on that synthetic-admin path. Added an assertion to the existing test.

**Suggestions**

5. Verified `Optional[Request] = None` fails exactly as you said (`FastAPIError: Invalid args for response field!`). No fix needed — noted for the record. Couldn't independently verify the 27→72 basedpyright count; it isn't installed here.

6. Checked router-by-router: `observability`, `siem`, `toolops_router`, `llm_admin_router`, `runtime_admin_router` are cleanly 100% `GLOBAL_ONLY` — and `log_search` too, actually (5/5, not mixed as I first read it). But `llm_config_router` isn't: `GET /llm/gateway/models` is deliberately non-admin (feeds the LLM Chat model selector), so a router-level dependency would silently break that route. Given that wrinkle and that you already flagged this as a follow-up rather than a blocker, leaving it out of this PR.

7. Confirmed unwired (no Makefile/CI references) and superseded by the permanent manifest test. Removed `scripts/verify_global_guard.py`.

8. Fair — the old test only checked mount-by-name. Added two tests in `test_metrics_maintenance_scope.py` that drive a real narrowed token and a real unrestricted token through the actual `require_global_admin_scope_dep` (only `require_admin_auth` is overridden), asserting 403 and 200 respectively.

9. Fixed — missing-request denials now log `token_teams='<no request>'` instead of `token_teams=None`. Added a direct test for the log line.

**Minor**

- Replaced the dangling "Test Isolation Patterns" reference in all 7 `*_scope.py` files with a self-contained comment.
- Fixed the `rbac.md` "(first two)" line — all three helpers read `request.state.token_teams`; only the decorator/callable forms need the endpoint to declare its own `request` param, the dependency form gets it via injection.
- Unchecked the triage-label checklist item on the PR description; #6134 does carry `triage`.

Full test suite for the touched paths (rbac middleware, all `*_scope.py` files, `test_global_record_scope.py`, `test_metrics_maintenance*.py`) plus the full doctest sweep all pass. `make ruff`, `ruff format --check`, `interrogate` (100%), and pylint against `.pylintrc` are clean on every file I touched — the pylint findings that do show up in `observability.py` (unrelated `E1102`/`W0622`/etc.) pre-exist on `main` at the same lines.

msureshkumar88:
## CI status after rebase

All checks green except `pytest (py3.12)`: https://github.com/IBM/mcp-context-forge/actions/runs/31609758448/job/94157905246

12 failures, all the same error:

```
sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) no such table: observability_traces
```

in `tests/e2e/test_otel_control_telemetry_e2e.py` and `tests/e2e/test_otel_plugin_metadata_e2e.py`. This is the pre-existing `-n auto` parallel-worker DB-timing flake already called out in the PR description's Verification table (same file set, same count). Ran both files locally in isolation and all pass — no failures, matching the "not observed when the same files run in isolation" note. Not something introduced by the rebase or by the review-fix commits, so leaving as-is rather than chasing an unrelated flake here.

msureshkumar88:
## Fixed the `pytest (py3.12)` failure

Root-caused and fixed the 12 CI-only failures in `test_otel_control_telemetry_e2e.py` / `test_otel_plugin_metadata_e2e.py` (`sqlalchemy.exc.OperationalError: no such table: observability_traces`), rather than leaving them as an unexplained flake.

**Root cause:** `mcpgateway/instrumentation/sqlalchemy.py` runs a single, process-wide background thread that dequeues and writes SQL-instrumentation spans asynchronously, decoupled from the request that produced them. Both otel e2e suites build a per-test `StaticPool` in-memory SQLite engine and call `engine.dispose()` at teardown. If a span for that engine is still sitting in the queue when `dispose()` runs, the background write lands on a disposed/empty in-memory database → `no such table`. Because the writer thread is a singleton reused across the whole worker process, this only has a realistic chance of firing under a large `-n auto` run with heavy scheduling contention — it never reproduced for me in isolation or on a ~200-test subset, matching what CI showed.

Verified directly rather than by inference: wrote a small standalone repro that enqueues a real span, disposes the engine before the writer thread gets to it, and confirmed the same `no such table: observability_spans` warning fires. Draining the queue first before disposing eliminates it cleanly.

**Fix:** added `tests/helpers/observability.py::drain_span_writer_queue()`, which blocks (bounded to 5s) until the background writer has finished everything currently queued, and call it immediately before every `engine.dispose()` in both files (2 call sites in `test_otel_control_telemetry_e2e.py`, 4 in `test_otel_plugin_metadata_e2e.py`). Test-only change — nothing in `mcpgateway/instrumentation/sqlalchemy.py` itself was touched, since that separate-session/background-writer design is intentional (see `AGENTS.md`'s Observability Transaction Behavior section) and changing it would be a much larger, unrelated change.

Ran both files individually and the full `tests/e2e/` directory under `-n auto` after the fix — all pass, exit 0.

msureshkumar88:
## CI is green — correcting my earlier root-cause claim

My previous comment on this failure was wrong. The `drain_span_writer_queue()` fix addressed a real race, but not the one actually causing these 12 failures — CI still failed the same way after that commit. Went back and root-caused it properly this time, verifying by direct local reproduction rather than inference.

**Actual root cause:** both otel e2e fixtures patched `mcpgateway.routers.observability`'s `SessionLocal` by string path — `monkeypatch.setattr("mcpgateway.routers.observability.SessionLocal", ...)`. pytest resolves that string via `sys.modules` at patch time. `tests/unit/mcpgateway/routers/test_observability_scope.py` does a deliberate `sys.modules.pop("mcpgateway.routers.observability") ` + reimport (documented "Pattern B" isolation, so it always binds real, non-mocked decorators regardless of collection order). When that reimport runs after the otel e2e file's own top-level `from mcpgateway.routers.observability import router as observability_router` — which happens under xdist's collection ordering — the two files end up holding *different* module objects. The otel fixture's string-path patch lands on the new, orphaned copy; the already-mounted router (and its `get_db()`) keeps reading the original module's untouched `SessionLocal`, which has no `observability_traces`/`observability_spans` table → `no such table`.

I verified this directly: with `-p no:xdist`, running the otel suite followed by `test_observability_scope.py` reproduced the exact 7-test failure locally; reversing the order did not. That confirmed both the mechanism and that it's collection-order dependent, matching why it never showed up in smaller/isolated runs.

First attempt at a fix (switching all seven `*_scope.py` suites from `sys.modules.pop`+reimport to `importlib.reload()` to preserve module identity) masked this bug but broke a different, equally deliberate mechanism: `test_rbac_router.py` relies on capturing a module object whose route decorators were baked in at import time under mocked RBAC decorators (documented in that file's own top-of-file comment). `reload()` mutates that same shared module object in place, silently re-decorating its endpoints with the real guard mid-session and breaking 16 of that file's own tests. Caught this with a full regression sweep before pushing and reverted it entirely.

**Actual fix:** import `mcpgateway.routers.observability` as a module object (`observability_mod`) in both otel e2e files, alongside the existing `from ... import router` import, and patch `SessionLocal` on that object directly instead of by string path. Object-based `monkeypatch.setattr()` never touches `sys.modules`, so it always targets the exact module object `observability_router` is bound to, independent of what any other file does to `sys.modules` later in the same worker. Doesn't touch `test_observability_scope.py`'s or `test_rbac_router.py`'s existing isolation patterns at all.

CI is now green across all 30 checks including `pytest (py3.12)`.

jonpspri (dismissed):
## Cycle 2 review — previous findings verified, ready to merge

Follow-up pass on `90e154d1a` to confirm the cycle-1 findings are addressed. All nine findings plus the minor notes check out against the code — verified directly, not from the commit messages:

**Blocking**

1. **Doctest regressions** — confirmed fixed. Stub request (`SimpleNamespace(state=…(token_teams=None))`) injected at all 7 call sites in `observability.py`; re-ran `--doctest-modules mcpgateway/routers/observability.py` locally: **7 passed**.

**Functionally impacting**

2. **Basic-auth docs** — `configuration.md:150,153` and `adr/004-combine-jwt-and-basic-auth.md:212` now use `/health/security` as the Basic-auth example and explicitly call out that global-record routes (`/api/metrics/*`, `/api/logs/*`) require an unrestricted bearer/session token.
3. **Proxy-auth admins** — `rbac.py:325-327` now resolves `token_teams` through `_resolve_teams_from_db`, the same canonical helper `_resolve_proxy_payload` uses, so the two proxy entry points agree. I checked the helper's semantics against `auth.py:451` (`is_admin` → `None` bypass) — correct. Four regression tests cover admin, non-admin, and the fail-closed resolution-failure fallback.
4. **`allow_unauthenticated_admin`** — `rbac.py:502` sets `request.state.token_teams = None` on the synthetic-admin path, with an assertion on the existing test.

**Suggestions**

5. basedpyright +45 — acknowledged, no change expected.
6. Router-level dependency — the `llm_config_router` wrinkle (`GET /llm/gateway/models` is deliberately non-admin) is a fair reason to defer; fine as a follow-up.
7. `scripts/verify_global_guard.py` — removed; the drift-guard test is the single authority now.
8. **metrics_maintenance coverage** — `test_narrowed_admin_token_is_denied_by_the_real_dependency` / `test_unrestricted_admin_token_passes_the_real_dependency` drive real tokens through the actual dependency. Exactly what was asked for.
9. **`<no request>` sentinel** — applied in all three guard paths (`require_unrestricted_platform_admin`, `require_global_admin_permission`, `require_global_admin_scope_dep`), with a direct caplog test.

**Minor** — the dangling "Test Isolation Patterns" references are gone (replaced with self-contained comments), the rbac.md "(first two)" wording is fixed, and the triage checklist item on the PR description is corrected.

**New code in the fix commits** got the same scrutiny: the proxy `token_teams` resolution is minimal and fails closed; `drain_span_writer_queue()` is sound (the writer thread does call `task_done()`, so polling `unfinished_tasks` terminates); and the module-object `SessionLocal` patching in the otel e2e files is the right fix for the `sys.modules.pop` + reimport identity desync — the root-cause write-up matches what the code does.

**Validation gate** (run on this head):

| # | Gate | Result |
|---|------|--------|
| 1 | `make ruff interrogate pylint` | clean; pylint 10.00/10 |
| 2 | `make test` | 21,953 passed / 819 skipped / 2 xfailed, exit 0 |
| 3 | `make coverage diff-cover` | exit 0; 100% on every file in this PR's own scope — the only missed lines (`routers/rbac.py:394,454`) trace to #6132's own commit (`19dcc3fa9` on that branch) |
| 4–5 | Gateway stack / MCP protocol e2e | accepted via CI: all 52 checks green on `90e154d1a` incl. Compose Production Smoke, plus your live e2e suite (24 passed / 5 skipped) |
| 6 | `make detect-secrets-scan` | exit 0, no unaudited/live findings |

No new findings this cycle. Approving — the one standing constraint is the stack order: **#6132 merges first**, then this one.

Lang-Akshay (changes requested):
The two blocking findings are the RBAC role-assignment visibility gap and the Basic-auth `/version` regression. Other reviewed compatibility notes are intentionally omitted from this comment.

## 1. `GET /rbac/users/{user_email}/roles` scope leak — valid security issue

**Evidence**

- The route has only `@require_permission("admin.user_management")` (`mcpgateway/routers/rbac.py:466-503`); it has no global guard and receives no request scope context.
- It calls `PermissionService.get_user_roles()` with only the target email, optional requested scope, and expiration flag.
- `PermissionService.get_user_roles()` queries all active assignments for that user and applies only the caller-supplied `scope`/`team_id` filters (`mcpgateway/services/permission_service.py:291-325`). It has no caller `token_teams` filter.
- `UserRoleResponse` exposes `role_name`, `scope`, `scope_id`, `granted_by`, and assignment metadata (`mcpgateway/schemas.py:7655-7698`).
- Therefore a narrowed admin who can satisfy `admin.user_management` can query another user's roles and receive global assignments and out-of-scope team identifiers. The route is classified as team-scopable in the RBAC policy, so the missing stored-row filtering is a Layer-1 visibility gap.

**Fix**

Make role reads scope-aware using the authenticated caller's canonical `token_teams`, not query parameters:

- Permit global assignments only when the caller has unrestricted platform-admin scope (`token_teams is None`).
- Return team assignments only when their stored `scope_id` is in the caller's allowed teams.
- Apply the same explicit policy to personal assignments; do not infer authorization from the requested `scope` query value.
- Prefer adding the filtering to `PermissionService.get_user_roles()` (with a `token_teams` argument) so all callers share one policy point. Add deny-path tests for a narrowed caller requesting global and out-of-scope team assignments.

## 2. `/version` Basic-auth regression — valid blocking regression

**Evidence**

- Before PR #6183, `mcpgateway/version.py` used `_has_version_admin_access()`, which accepted the bare string returned by verified `require_admin_auth()` Basic-auth requests.
- The PR now calls `is_unrestricted_platform_admin(request, _user, db)` (`mcpgateway/version.py:1337-1362`).
- `is_unrestricted_platform_admin()` requires resolved `token_teams is None` (`mcpgateway/auth_context.py:906-917`).
- Basic-auth does not set JWT scope state, so `get_token_teams_from_request()` falls back to `[]` (`mcpgateway/auth_context.py:664-723`), causing the request to receive 403.
- The existing test mocks the scope helper and does not exercise the real Basic-auth path.

**Fix**

Preserve a verified Basic-auth marker in request state/auth context and accept it only after `require_admin_auth()` succeeds. Keep strict `token_teams is None` checks for JWT/session callers, and add an integration-style test using real Basic credentials without mocking the scope helper.

msureshkumar88:
Thanks for the review — both blocking findings check out against the code, and both are fixed in 7f61142.

**1. `GET /rbac/users/{user_email}/roles` scope leak**

Confirmed: the route had no Layer-1 filtering, and `PermissionService.get_user_roles()` applied only the caller-supplied `scope`/`team_id` query filters with no `token_teams` check.

Fix follows the same visibility policy `_authorize_assignment_scope()` already enforces on writes:
- `PermissionService.get_user_roles()` now takes an optional `token_teams` argument. When set, it drops global rows, keeps team rows only when their `scope_id` is in `token_teams`, and drops personal rows outright (those belong to the target user, not the caller, and must not be inferred visible from the requested `scope` value).
- The router derives the caller's `token_teams` via `get_scoped_resource_access_context()` and passes it through — except when the caller is viewing their own email, which stays unfiltered for parity with `/rbac/my/roles`.
- Added deny-path tests: `test_get_user_roles_narrowed_caller_sees_only_covered_team_rows`, `test_get_user_roles_public_only_caller_sees_nothing` in `test_permission_service.py`, and router-level `test_get_user_roles_narrows_by_caller_token_teams_for_other_users` / `test_get_user_roles_unfiltered_when_caller_views_own_roles` in `test_rbac_scope.py`.

**2. `/version` Basic-auth regression**

Confirmed: `verify_basic_credentials()` never touches `request.state`, so a Basic-auth admin has no `token_teams` set, and `is_unrestricted_platform_admin()` requires `token_teams is None` *and* a DB-backed admin check (`EmailUser.is_admin` or `platform_admin_email`) — which the static `BASIC_AUTH_USER`/`BASIC_AUTH_PASSWORD` credential doesn't satisfy either, so simply setting `token_teams = None` wouldn't have been enough on its own.

Fix: `require_admin_auth()` now marks a verified Basic-auth admin directly on `request.state.basic_auth_verified_admin`, and `version_endpoint()` trusts that marker instead of calling `is_unrestricted_platform_admin()` for that path — matching the pre-#6183 behavior where the bare-string return from Basic auth was itself sufficient proof of admin. JWT/session callers are untouched and still go through the strict `token_teams is None` DB check.
- Added `test_basic_auth_verified_marker_bypasses_db_scope_check` (asserts `is_unrestricted_platform_admin` is never called) and `test_basic_auth_admin_allowed_end_to_end` (real `require_admin_auth` dependency, real Basic credentials, no scope-helper mocking) in `test_version_scope.py`.

Also rebased onto main and regenerated `.secrets.baseline` (one recurring doc-example false positive, audited as not-a-secret).

PR #6194, opened 2026-08-11
fix(resources): surface text_content in ResourceRead and wire view/edit

### LINKED ISSUE #6058: [BUG]: Resource View modal never shows the resource's content
### 🐞 Bug Summary

A resource created with inline text content cannot be read back through the Admin UI's View modal. The modal shows ID/URI/name/type/description/visibility/tags/status/metadata, but never the actual stored content.

`const content = data.content;` is present in the code but commented out, and `content` is absent from the field list the view modal renders. The identical line is commented out again on the edit path.

`mcpgateway/admin_ui/resources.js:327` (view), `:349-360` (rendered field list), `:609` (edit path, same omission).

---

### 🧩 Affected Component

---

### 🔁 Steps to Reproduce

1. Create a resource via the Admin UI with inline text content (not a URI reference).
2. Open its row menu and click View.
3. Observe the modal shows metadata fields but never the content you typed.

---

### 🤔 Expected Behavior

The resource's stored content should be visible in the detail modal.

---

### 📓 Logs / Error Output

N/A

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Platform / OS | Linux |

---

Reproduced 3 times.

sujithvishal:
# Pull Request

## 🔗 Related Issue

Closes #6058

---

## 📝 Summary

`ResourceRead` had no `content` field, so `text_content` stored in the
database was silently discarded on every read. The Admin UI view and edit
modals had the content display and prefill code commented out as a
defensive measure — without a working read path, rendering the field
would have overwritten stored content with blank on every save.

This PR adds `content` to `ResourceRead`, maps `text_content → content`
in `convert_resource_to_read`, and restores the commented-out UI code
with the CodeMirror editor wired correctly.

---

## 📏 Reviewability

---

## 🏷️ Type of Change

- [x] Bug fix

---

## 🧪 Verification

| Check | Command | Status |
|---|---|---|
| Lint suite | `make lint` |✅ |
| Unit tests | `make test` | ✅|
| Coverage ≥ 80% | `make coverage` |✅ |
| Targeted fast run | `python -m pytest tests/unit/mcpgateway/services/test_resource_service.py -k "convert_resource_to_read" -v` | ✅|
| Manual — View modal | Create resource with text content → open View modal → content renders | ✅|
| Manual — Edit modal | Open Edit modal → content textarea prefilled with stored value | ✅|
| Manual — Save edit | Edit content → save → reopen → updated content persists |✅ |

---

## ✅ Checklist

---

## 📓 Notes

**Root cause:** `ResourceRead` (Pydantic schema) never had a `content`
field. `ResourceCreate` and `ResourceUpdate` accepted `content` on
writes, which was correctly stored in `DbResource.text_content`. But
`convert_resource_to_read` never mapped `text_content` back.

**Files changed:**

- `mcpgateway/schemas.py` — added `content: Optional[str]` to `ResourceRead`
- `mcpgateway/services/resource_service.py` — mapped
  `text_content → content` in `convert_resource_to_read`
- `mcpgateway/admin_ui/resources.js` — uncommented `resource.content` read on view and edit paths; wired edit prefill through
- `mcpgateway/templates/admin.html` — uncommented the content textarea in the edit modal with dark-mode CSS classes

image
image

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^\\.secrets\\.baseline$",
     "lines": null
   },
-  "generated_at": "2026-08-13T14:29:13Z",
+  "generated_at": "2026-08-14T11:33:11Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -3894,7 +3894,7 @@
         "hashed_secret": "3ee08a29a2ca26171fe152b8741d0c90b598263a",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 14942,
+        "line_number": 14954,
         "type": "Secret Keyword",
         "verified_result": null
       }
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3884,7 +3884,7 @@
         "hashed_secret": "c377074d6473f35a91001981355da793dc808ffd",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 4608,
+        "line_number": 4610,
         "type": "Hex High Entropy String",
         "verified_result": null
       }
@@ -3894,7 +3894,7 @@
         "hashed_secret": "3ee08a29a2ca26171fe152b8741d0c90b598263a",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 14959,
+        "line_number": 14942,
         "type": "Secret Keyword",
         "verified_result": null
       }
@@ -4658,7 +4658,7 @@
         "hashed_secret": "718cbcc5a4207c0d5f38e3a309bdba17cb0074b7",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 398,
+        "line_number": 440,
         "type": "Hex High Entropy String",
         "verified_result": null
       }
--- a/mcpgateway/admin_ui/resources.js
+++ b/mcpgateway/admin_ui/resources.js
@@ -324,7 +324,7 @@ export const viewResource = async function (resourceId) {
     const resource = data.resource;
 
     // console.log("Resource JSON:\n", JSON.stringify(resource, null, 2));
-    // const content = data.content;
+    const content = resource.content;
 
     const resourceDetailsDiv = safeGetElement("resource-details");
     if (resourceDetailsDiv) {
@@ -404,29 +404,19 @@ export const viewResource = async function (resourceId) {
       statusP.appendChild(statusSpan);
       container.appendChild(statusP);
 
-      // Content display - safely handle different types
-      // const contentDiv = document.createElement("div");
-      // const contentStrong = document.createElement("strong");
-      // contentStrong.textContent = "Content:";
-      // contentDiv.appendChild(contentStrong);
-
-      // const contentPre = document.createElement("pre");
-      // contentPre.className =
-      //     "mt-1 bg-gray-100 p-2 rounded overflow-auto max-h-80 dark:bg-gray-800 dark:text-gray-100";
-
-      // // Handle content display - extract actual content from object if needed
-      // let contentStr = extractContent(
-      //     content,
-      //     resource.description || "No content available",
-      // );
-
-      // if (!contentStr.trim()) {
-      //     contentStr = resource.description || "No content available";
-      // }
-
-      // contentPre.textContent = contentStr;
-      // contentDiv.appendChild(contentPre);
-      // container.appendChild(contentDiv);
+      // Content display — only render when the resource has stored text content
+      if (content !== null && content !== undefined && String(content).trim() !== "") {
+        const contentDiv = document.createElement("div");
+        const contentStrong = document.createElement("strong");
+        contentStrong.textContent = "Content:";
+        contentDiv.appendChild(contentStrong);
+        const contentPre = document.createElement("pre");
+        contentPre.className =
+          "mt-1 bg-gray-100 p-2 rounded overflow-auto max-h-80 dark:bg-gray-800 dark:text-gray-100 text-sm whitespace-pre-wrap break-words";
+        contentPre.textContent = String(content);
+        contentDiv.appendChild(contentPre);
+        container.appendChild(contentDiv);
+      }
 
       // Metrics display
       if (resource.metrics) {
@@ -606,7 +596,7 @@ export const editResource = async function (resourceId) {
 
     const data = await response.json();
     const resource = data.resource;
-    // const content = data.content;
+    const content = resource.content;
     // Ensure hidden inactive flag is preserved
     const isInactiveCheckedBool = isInactiveChecked("resources");
     let hiddenField = safeGetElement("edit-resource-show-inactive");
@@ -673,7 +663,7 @@ export const editResource = async function (resourceId) {
     const nameField = safeGetElement("edit-resource-name");
     const descField = safeGetElement("edit-resource-description");
     const mimeField = safeGetElement("edit-resource-mime-type");
-    // const contentField = safeGetElement("edit-resource-content");
+    const contentField = safeGetElement("edit-resource-content");
 
     if (uriField && uriValidation.valid) {
       uriField.value = uriValidation.value;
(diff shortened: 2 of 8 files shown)


sujithvishal:
@msureshkumar88 Please Review the PR

msureshkumar88 (changes requested):
Thanks for tracking down the root cause here — `ResourceRead` never had a `content` field, and the commented-out UI code makes a lot more sense once you know it was defensively guarding against overwriting stored content with blank on save. The `text_content → content` mapping is correct, well-targeted, and the two unit tests cover it directly. Also appreciated catching that the original commented code read `data.content` instead of `resource.content` — restoring it verbatim would've reintroduced a bug.

No alembic migration needed here and it's good that none was added — `text_content`/`binary_content` already exist on `DbResource`, this is purely a schema/serialization fix.

**Blocking**

1. **`content` now gets embedded in every resource-list response, not just the two modals this issue is about.** `convert_resource_to_read` is the shared conversion path for `list_resources`, `list_resources_for_user`, and `list_server_resources` (backing `GET /resources`, the paginated variant, and `GET /servers/{id}/resources`), in addition to the single-resource read path the View/Edit modals use. Before this PR, `content` was never serialized anywhere, so list responses stayed lean. After, every item on every list page now ships its full `text_content` — and there's no size cap on that column (`Text`, unbounded, no `max_length` on `ResourceCreate.content`), so a page of resources with any meaningfully-sized bodies means real payload growth on every listing call (Admin UI resource tab, `GET /resources`, `GET /servers/{id}/resources`), not just the detail fetch the fix is meant for.

   Suggest gating `content` behind a flag mirroring the existing `include_metrics` pattern (e.g. `include_content`, default `False`), populated only on the single-resource read path (`read_resource`/`get_resource_by_id`) that the modals and `GET /resources/{id}` actually use. That keeps list payloads unchanged and scopes the fix to what #6058 actually needs.

   Worth flagging as a hardening consideration too, even though it's not an ACL bypass: before, reading N resources' bodies took N individually-auditable `read_resource` calls; after, one `GET /resources?limit=<n>` returns every visible resource's full body in a single response. Not a new vulnerability since visibility rules are unchanged, but it does widen the audit/bulk-access surface — the same `include_content` gating addresses this too.

**Suggestion**

2. **No JS-level regression test for the fixed behavior.** `tests/unit/js/resources.test.js` already has `describe("viewResource")` and `describe("editResource")` blocks with a working vitest harness (mocked `fetchWithTimeout`, DOM fixtures for `resource-details`, `edit-resource-content`, etc.), but none of the existing tests set a `resource.content` value or assert the "Content:" block renders, or that `editResourceContentEditor.setValue(...)` gets called with it. Since this is precisely the regression the PR fixes, and the harness is already in place, a small addition here would go a long way — especially given the PR's own reviewability checklist notes "tests are included with the code they validate."

**Minor / non-blocking**

- `convert_resource_to_read` only maps `text_content`, not `binary_content` — resources created via the API with binary content will still show `content: null`. Given #6058 is scoped to text-content resources this seems fine to leave out of this PR, but might be worth a tracked follow-up so it doesn't get lost, since `ResourceRead.content`'s name doesn't signal that it's text-only.

Everything else looks solid — no alembic changes needed and none added, no unrelated diffs, `extractContent()` cleanly removed with no dangling references, and the issue link/root-cause writeup in the PR description is clear and accurate. Happy to take another look once the list-endpoint scoping is addressed.

sujithvishal:
Hi @msureshkumar88 That's a good catch. I adopted the review comment
**Blocking**
1. gatted content behind a flag mirroring the existing include_metrics pattern (e.g. include_content, default False), populated only on the single-resource read path (read_resource/get_resource_by_id) that the modals and GET /resources/{id} actually use.

**Suggestion**
2. Added JS-level regression test for the fixed behavior

Non-Blocking
I would suggest tracking that seperately  so it doesn't get lost

msureshkumar88 (changes requested):
Thanks for the fast follow-up — the `include_content` gating and the new JS test coverage both land cleanly and address the earlier feedback well.

One thing surfaced when re-running the suite against the latest commit:

## Blocking

**`tests/unit/mcpgateway/services/test_resource_service.py::test_convert_resource_to_read_maps_text_content_to_content` fails on the current head.**

```
result = resource_service.convert_resource_to_read(mock_resource, include_metrics=False)
assert result.content == "Hello paddddd"
AssertionError: assert None == 'Hello paddddd'
```

This test was added alongside the original fix, before `include_content` existed as a gate. Now that `convert_resource_to_read` defaults `include_content=False`, the call needs to opt in:

```python
result = resource_service.convert_resource_to_read(mock_resource, include_metrics=False, include_content=True)
```

Ran the full file locally — this is the only failure, everything else is green.

## Non-blocking, worth a look

- `register_resource()` and `update_resource()` still call `convert_resource_to_read(resource)` without `include_content=True`, so `POST /resources` / `PUT /resources/{id}` responses come back with `content: null` even though the write succeeded. Not a data-loss issue (a follow-up GET returns the real content), just a bit surprising for API consumers reading their own write.
- The docstring on `GET /resources/{id}/info` (`mcpgateway/main.py`) still says it's "different from GET /resources/{resource_id} which returns the resource content" — no longer accurate since `get_resource_by_id` now always includes content for every caller.
- Minor/cosmetic: the view modal hides whitespace-only content, but the edit modal/CodeMirror prefill shows it as-is — inconsistent treatment of the same edge case, low impact.

Once the test assertion is fixed this looks good to me.

sujithvishal:
@msureshkumar88 Thanks for the review
**Blocking**
Fixed the failing test

**Non-Blocking**
`register_resource()`, ` update_resource()` and ` set_resource_state` is updated to include content to make all the write path consistent. Added test case for the same

Fixed The docstring on GET `/resources/{id}/info` (mcpgateway/main.py)

msureshkumar88 (changes requested):
Thanks for the fast turnaround on both prior rounds — the `include_content` gating and the JS regression tests both land cleanly. Confirmed directly (not just from the description):

- The previously-failing `test_convert_resource_to_read_maps_text_content_to_content` passes now.
- `list_resources`/`list_resources_for_user`/`list_server_resources` still call the converter with the default `include_content=False`, so list payloads stay lean — the round-1 concern is closed.

Two things came up on this pass, one of which is a data-loss risk worth blocking on:

## Blocking

**Editing a binary resource through the Admin UI silently wipes its content.**

`admin_edit_resource` (`mcpgateway/admin.py:13598`) always builds `ResourceUpdate(content=str(form.get("content", "")), ...)` — never `None`. Since `ResourceRead.content` (this PR's new field) only maps `text_content`, the edit modal for a binary resource is prefilled with `""`. On save, `update_resource` (`mcpgateway/services/resource_service.py:3229`) sees `resource_update.content is not None` (it's `""`, not `None`), takes the "content changed" branch, and sets `text_content = ""`, `binary_content = b""`, `size = 0` — even if the user only changed the name or tags.

This overwrite-on-any-string behavior predates this PR (the field was commented out before, so it always submitted `""` too), but before this PR nobody could reasonably trust that field, so it likely masked the bug. After this PR the modal visibly shows/hides content and the description's own checklist claims edits persist — so this becomes a much more convincing trap for binary resources specifically. Since this PR is exactly the one wiring `ResourceRead.content` end-to-end, closing this gap feels in scope — e.g. skip the content overwrite when the original resource had no `text_content` and the submitted value is empty, or extend the read path to also surface `binary_content` so the modal doesn't silently coerce it to blank.

## Worth a look (not blocking)

**`include_content=True` leaked onto two endpoints that shouldn't need it.**

`get_resource_by_id` backs both `GET /resources/{id}` (correctly wants content) and `GET /resources/{id}/info` (previously documented as the metadata-only counterpart — the docstring literally said "different from GET /resources/{resource_id} which returns the resource content," and this PR quietly dropped that sentence rather than preserving the distinction). `set_resource_state` (the enable/disable toggle) also now returns full content it doesn't need. Since `get_resource_by_id` hardcodes `include_content=True` with no way to opt out, both endpoints ship the full resource body now — the same payload-growth pattern the round-1 review flagged and gated everywhere else, just missed on these two call sites. Suggest threading an `include_content` param through and defaulting it to `False` for `/info` and the state toggle.

Everything else looks good — alembic correctly left untouched (no schema change needed here), no XSS concerns in the new render path (`textContent`, not `innerHTML`), `extractContent()` cleanly removed with no dangling references, and no unrelated diffs beyond the usual `.secrets.baseline` timestamp churn.

Happy to take another look once the binary-content path is addressed.

PR #6196, opened 2026-08-11
fix(helm): fix OCP PGO deployment failures on fresh clusters

### LINKED ISSUE #6195: [BUG]: OCP PGO deployment fails after recent Helm chart security hardening
### 🐞 Bug Summary

Running `make ocp-setup && make ocp-deploy` on a fresh OpenShift cluster fails with three separate errors that together block deployment completely. The failures were introduced by security hardening changes in (Release 1.0.7) which added `runAsUser: 999` and `seccompProfile: RuntimeDefault` to the Redis deployment, both of which are incompatible with OpenShift's SCC system.

These failures are not visible on clusters where a previous deployment already ran, because manual SCC grants from earlier setups persist and mask the issue.

The three root causes are:

1. **Helm secret race condition** - the migration Job hook fires before Helm creates regular resources, so the gateway secret doesn't exist when the migration pod starts, causing a `__REPLACE_ME__` placeholder rejection at startup.
2. **Missing `anyuid` SCC grant** - `runAsUser: 999` was added to the Redis container. UID 999 is outside the OCP namespace-allocated UID range, requiring an explicit `anyuid` SCC grant that the setup playbook never makes.
3. **Hardcoded `seccompProfile` in Redis pod spec**, `seccompProfile: RuntimeDefault` was added. The `anyuid` SCC (required for UID 999) does not permit `seccompProfile` in the pod spec, causing `FailedCreate` on the Redis ReplicaSet even after the SCC grant.

---

### 🧩 Affected Component

- [x] CLI, Makefiles, or shell scripts — `ansible/ocp/playbooks/deploy.yml`, `setup.yml`
- [x] Container setup (Docker/Podman/Compose) — `charts/mcp-stack/templates/deployment-redis.yaml`, `charts/mcp-stack/profiles/ocp/values-pgo.yaml`

---

### 🔁 Steps to Reproduce

1. Provision a **fresh** OpenShift 4.11+ cluster (e.g. Fyre Quick Burn, OCP 4.2)
2. Log in via `oc login`
3. Run `make ocp-setup OCP_NS=contextforge`
4. Prepare `charts/mcp-stack/profiles/ocp/values-pgo-secrets.yaml` with real secret values
5. Run `make ocp-deploy OCP_NS=contextforge`

---

### 🤔 Expected Behavior

All pods reach `1/1 Running` status. The deployment completes successfully with no manual intervention.

---

### 📓 Logs / Error Output

**Failure 1 - Migration pod crashes with placeholder rejection:**

```
mcpgateway.config.SecurityConfigurationError: jwt_secret_key: unset placeholder
(__REPLACE_ME__) rejected in every environment (including 'development').
```

Cause: `helm install` fires the `pre-install` migration Job hook before creating the gateway Secret. The pod starts with no secret injected and falls back to the `__REPLACE_ME__` default.

**Failure 2 - Redis pod never created:**

```
Error creating: pods "contextforge-mcp-stack-redis-xxx" is forbidden:
unable to validate against any security context constraint:
[provider restricted-v2: .containers[0].runAsUser: Invalid value: 999:
must be in the ranges: [1000750000, 1000759999]]
```

Cause: Redis image runs as UID 999. OCP 4.11+ `restricted-v2` SCC enforces namespace-allocated UID ranges. No SCC grant is made by the playbook.

**Failure 3 - Redis pod blocked even after SCC grant:**

```
metadata.annotations[seccomp.security.alpha.kubernetes.io/pod]: Forbidden: seccomp may not be set
```

Cause: `deployment-redis.yaml` hardcodes `seccompProfile: RuntimeDefault` in the pod spec. The `anyuid` SCC (required for UID 999) does not permit this field.

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Version or commit | latest |
| Platform / OS | OpenShift 4.22 (Fyre Quick Burn) |
| Deployment method | `make ocp-setup && make ocp-deploy` (Ansible + Helm) |
| Helm chart | `charts/mcp-stack` with `profiles/ocp/values-pgo.yaml` |
| Postgres operator | CrunchyData PGO v5 |

---

claudia-gray:
## 🔗 Related Issue

Closes https://github.com/IBM/mcp-context-forge/issues/6195

---

## 📝 Summary

Fixes three issues that block `make ocp-setup && make ocp-deploy` from completing on a fresh OpenShift cluster after the security hardening changes in Release 1.0.7

**Problem 1 - Helm secret race condition (`deploy.yml`)**
The migration Job is a `pre-install` hook that fires before Helm creates regular resources. This means the gateway secret doesn't exist when the migration pod starts, so the app falls back to the `__REPLACE_ME__` placeholder and crashes. Fixed by pre-creating the secret with Helm ownership labels in the deploy playbook before `helm install` runs.

**Problem 2 — Missing SCC grant (`setup.yml`)**
`runAsUser: 999` was added to the Redis container in Release 1.0.7. UID 999 is outside the OCP namespace-allocated UID range enforced by the `restricted-v2` SCC, so the Redis pod is never created. Fixed by granting `nonroot-v2` SCC to the default service account during namespace setup. `nonroot-v2` allows any non-root UID (covers 999), still permits `seccompProfile`, and is narrower than `anyuid`.

**Problem 3 — (resolved by Problem 2 fix)**
The original fix used `anyuid` which rejects `seccompProfile`, requiring a conditional in `deployment-redis.yaml` and an override in `values-pgo.yaml`. Switching to `nonroot-v2` resolves this — both of those changes revert entirely.

---

## 📏 Reviewability

---

## 🏷️ Type of Change

- [x] Bug fix

---

## 🧪 Verification

Verified by deploying from scratch on a fresh Fyre Quick Burn cluster (OCP 4.22):

| Check | Command | Status |
|---|---|---|
| Full deploy from clean namespace | `make ocp-setup OCP_NS=contextforge && make ocp-deploy OCP_NS=contextforge` | ✅ All pods `1/1 Running` |
| Gateway pods | `oc get pods -n contextforge` | ✅ 3/3 Running |
| Redis pod | `oc get pods -n contextforge \| grep redis` | ✅ 1/1 Running |
| Migration job | `oc get jobs -n contextforge` | ✅ Completed cleanly |
| Gateway secret created | `oc get secret contextforge-mcp-stack-gateway-secret -n contextforge` | ✅ Present with correct values |
| nonroot-v2 SCC applied | `oc get rolebinding -n contextforge \| grep nonroot` | ✅ Bound to default SA |

**Before :** `make ocp-deploy` failed with migration pod `SecurityConfigurationError` and Redis `FailedCreate`.

**After (with fix):** Clean deploy, all 10 pods reach `Running` status within ~2 minutes.

---

## ✅ Checklist

---

## 📓 Notes

**Files changed:**

| File | Change |
|------|--------|
| `ansible/ocp/playbooks/deploy.yml` | Pre-creates gateway secret with Helm ownership labels before `helm install`; `no_log: true` added; secret values coerced to strings |
| `ansible/ocp/playbooks/setup.yml` | Grants `nonroot-v2` SCC (not `anyuid`) to default service account after namespace creation |
| `charts/mcp-stack/profiles/ocp/values-pgo.yaml` | Reverted — `podSecurityContext` override no longer needed with `nonroot-v2` |
| `charts/mcp-stack/templates/deployment-redis.yaml` | Reverted — `seccompProfile` conditional no longer needed with `nonroot-v2` |

These failures only surface on a **fresh** cluster because on previously deployed clusters, manual SCC grants from earlier runs persist and mask the issue. Anyone spinning up a new Quick Burn cluster after Release 1.0.7 will hit all three failures.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -1,9 +1,9 @@
 {
   "exclude": {
-    "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^\\.secrets\\.baseline$|^.secrets.baseline$",
+    "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^\\.secrets\\.baseline$",
     "lines": null
   },
-  "generated_at": "2026-09-07T15:50:12Z",
+  "generated_at": "2026-09-08T10:01:50Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -1518,7 +1518,7 @@
         "hashed_secret": "602a3bcd9e73d7abd64dc26b0cb407e963c9f806",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 188,
+        "line_number": 191,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -1526,7 +1526,7 @@
         "hashed_secret": "de7c0d0731b4b25d51e1d5025b8559d073d3c0e7",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 353,
+        "line_number": 356,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -1534,7 +1534,7 @@
         "hashed_secret": "d09a50d86d7403a4bc3ee1823f91ec14fb14acd4",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 362,
+        "line_number": 365,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -1542,7 +1542,7 @@
         "hashed_secret": "2bdb52a577a3b6376b125fa94bfa4366a53c5ea0",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 364,
+        "line_number": 367,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -1550,7 +1550,7 @@
         "hashed_secret": "ae0131d240ba7cc32be22b778cb4a62ccfbd3660",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 370,
+        "line_number": 373,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -1558,7 +1558,7 @@
         "hashed_secret": "7e15bb5c01e7dd56499e37c634cf791d3a519aee",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 581,
+        "line_number": 584,
         "type": "Secret Keyword",
         "verified_result": null
       }
--- a/ansible/ocp/playbooks/deploy.yml
+++ b/ansible/ocp/playbooks/deploy.yml
@@ -39,6 +39,30 @@
       failed_when: confirm.user_input | lower not in ['yes', 'y']
       when: not (skip_confirm | default(false) | bool)
 
+    # Pre-create the gateway secret with Helm ownership labels so the migration
+    # hook pod finds it immediately (before Helm creates regular resources).
+    - name: Load secrets values file
+      ansible.builtin.include_vars:
+        file: "{{ ocp_secrets }}"
+        name: secret_values
+
+    - name: Pre-create gateway secret with Helm ownership labels
+      kubernetes.core.k8s:
+        state: present
+        definition:
+          apiVersion: v1
+          kind: Secret
+          metadata:
+            name: "{{ ocp_namespace }}-mcp-stack-gateway-secret"
+            namespace: "{{ ocp_namespace }}"
+            labels:
+              app.kubernetes.io/managed-by: Helm
+            annotations:
+              meta.helm.sh/release-name: "{{ ocp_namespace }}"
+              meta.helm.sh/release-namespace: "{{ ocp_namespace }}"
+          stringData: "{{ dict(secret_values.mcpContextForge.secret | dict2items | map(attribute='key') | list | zip(secret_values.mcpContextForge.secret | dict2items | map(attribute='value') | map('string') | list)) }}"
+      no_log: true
+
     - name: Install ContextForge Helm chart
       ansible.builtin.command:
         cmd: >
--- a/ansible/ocp/playbooks/deploy.yml
+++ b/ansible/ocp/playbooks/deploy.yml
@@ -45,6 +45,43 @@
       ansible.builtin.include_vars:
         file: "{{ ocp_secrets }}"
         name: secret_values
+      no_log: true
+
+    # Derive the secret name from the chart's own mcp-stack.fullname helper so
+    # it always matches what Helm renders, regardless of the release name.
+    - name: Derive gateway secret name from chart template
+      ansible.builtin.command:
+        cmd: >
+          helm template {{ ocp_namespace }} {{ ocp_chart_path }}
+          -n {{ ocp_namespace }}
+          -f {{ ocp_values }}
+          -f {{ ocp_secrets }}
+          --show-only templates/secret-gateway.yaml
+      register: gateway_secret_template
+      changed_when: false
+      no_log: true
+
+    - name: Parse gateway secret name from template output
+      ansible.builtin.set_fact:
+        gateway_secret_name: >-
+          {{ (gateway_secret_template.stdout | from_yaml).metadata.name }}
+
+    - name: Expand secret dict to items (single pass)
+      ansible.builtin.set_fact:
+        gateway_secret_items: "{{ secret_values.mcpContextForge.secret | dict2items }}"
+      no_log: true
+
+    - name: Coerce secret values to strings
+      ansible.builtin.set_fact:
+        gateway_secret_data: >-
+          {{ dict(gateway_secret_items | map(attribute='key') | list
+             | zip(gateway_secret_items
+               | map(attribute='value')
+               | map('string')
+               | map('regex_replace', '^True$', 'true')
+               | map('regex_replace', '^False$', 'false')
+               | list)) }}
+      no_log: true
 
     - name: Pre-create gateway secret with Helm ownership labels
       kubernetes.core.k8s:
@@ -53,14 +90,14 @@
           apiVersion: v1
           kind: Secret
           metadata:
-            name: "{{ ocp_namespace }}-mcp-stack-gateway-secret"
+            name: "{{ gateway_secret_name }}"
             namespace: "{{ ocp_namespace }}"
             labels:
               app.kubernetes.io/managed-by: Helm
             annotations:
               meta.helm.sh/release-name: "{{ ocp_namespace }}"
               meta.helm.sh/release-namespace: "{{ ocp_namespace }}"
-          stringData: "{{ dict(secret_values.mcpContextForge.secret | dict2items | map(attribute='key') | list | zip(secret_values.mcpContextForge.secret | dict2items | map(attribute='value') | map('string') | list)) }}"
+          stringData: "{{ gateway_secret_data }}"
       no_log: true
 
     - name: Install ContextForge Helm chart
--- a/ansible/ocp/playbooks/setup.yml
+++ b/ansible/ocp/playbooks/setup.yml
@@ -65,6 +65,14 @@
       ansible.builtin.debug:
         msg: "Namespace: {{ 'created' if ns_result.changed else 'already exists' }}"
 
+    # Step 2b: Grant nonroot-v2 SCC to default service account (required for Redis UID 999).
+    # nonroot-v2 allows any non-root UID (covers 999) and still permits seccompProfile,
+    # so it is narrower than anyuid and avoids root-escalation risk (CWE-250).
+    - name: Grant nonroot-v2 SCC to default service account
+      ansible.builtin.command:
+        cmd: oc adm policy add-scc-to-user nonroot-v2 -z default -n {{ ocp_namespace }}
+      changed_when: false
+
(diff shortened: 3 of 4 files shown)


gandhipratik203 (dismissed):
Thanks for chasing this down. Nice work isolating three separate failures from what looks like one broken deploy, and the issue write-up with the actual event text for each makes it easy to follow the reasoning.

**Suggestion: narrower SCC grant**

`anyuid` on the `default` service account lets every pod in the namespace run as root, just so one Redis container can use UID 999. It's also what causes Problem 3: `anyuid` rejects `seccompProfile` (`Forbidden: seccomp may not be set`), which is the only reason `deployment-redis.yaml` needed the conditional and `values-pgo.yaml` the override.

Worth trying `nonroot-v2` instead: it allows any non-root UID including 999 and still permits seccomp, so Problem 3 never arises and both of those edits revert. Alternatively, null out Redis's UID and let OpenShift assign one, like the gateway does at `values-pgo.yaml:30-42` (blocked today by the hardcoded UID at `deployment-redis.yaml:41-51`).

Needs a cluster test, so happy for it to be a follow-up if you'd rather land the working fix first.

Approving either way, this unblocks a real deploy path.

msureshkumar88 (changes requested):
Thanks for pushing this through — the fresh-cluster OCP deploy blocker (SCC restriction + secret race from #6195) is a real pain point and this gets the shape of the fix right. No Alembic touch needed here, correctly scoped to ansible/Helm.

A few things worth addressing before merge, mostly around the security-sensitive nature of touching SCCs and secrets directly:

**Blocking**
- `ansible/ocp/playbooks/setup.yml:69` — `anyuid` SCC is granted to the namespace's shared `default` ServiceAccount rather than a Redis-specific one. Since most workloads in this chart fall back to `default` when `serviceAccount.create`/`name` aren't set, this ends up granting the privilege more broadly than intended (CWE-250).
  - **Update**: per @gandhipratik203's suggestion below, swapping `anyuid` → `nonroot-v2` on this same line (`oc adm policy add-scc-to-user nonroot-v2 -z default -n {{ ocp_namespace }}`) looks like the better fix — `nonroot-v2` allows any non-root UID (covers Redis's UID 999) without the root-escalation risk of `anyuid`, and it still permits `seccompProfile`, so it would resolve this finding and the seccomp regression below in one move, without needing a dedicated ServiceAccount. Needs a cluster check to confirm UID 999 actually works under `nonroot-v2`; fall back to a Redis-specific SA + `anyuid` if it doesn't.
- `ansible/ocp/playbooks/deploy.yml:63` — the secret-creation task doesn't set `no_log: true`, unlike the comparable JWT/token tasks in `benchmark-setup.yml`. Under `-v`/verbose CI runs this would print raw credentials to logs (CWE-532).
- `ansible/ocp/playbooks/deploy.yml:55` — the pre-created secret name is hand-built (`{{ ocp_namespace }}-mcp-stack-gateway-secret`) rather than derived the way `mcp-stack.fullname` does in the chart. When the release name already contains the chart name, Helm's actual secret name differs from this, and the migration pod would go right back to `__REPLACE_ME__` — the exact race #6195 is about. Might be worth sourcing this from the same helper logic, or templating it via `helm template` output.

**Suggestions**
- `charts/mcp-stack/templates/deployment-redis.yaml:39` — the OCP `podSecurityContext` override fully replaces rather than merges, so the Redis pod loses `seccompProfile: RuntimeDefault` under OCP. If the `nonroot-v2` swap above works out, this override (and the `values-pgo.yaml` one) can likely revert entirely instead of needing a merge fix.
- `deploy.yml:63` — `stringData` isn't quote-coerced like the chart's `| quote` does; could bite if a values file has an unquoted numeric/bool-looking secret value.

**Minor**
- Consider a helm-unittest case for the new `podSecurityContext` conditional in `deployment-redis.yaml` — there's already a test suite for that template.
- README prerequisites could mention the SCC-admin/cluster-admin privilege now needed for the `anyuid` (or `nonroot-v2`) grant step.

claudia-gray:
@msureshkumar88 @gandhipratik203 Thank you for your reviews.

anyuid → nonroot-v2 (setup.yml:69) - swapped as suggested. nonroot-v2 covers Redis UID 999, still permits seccompProfile. Cluster-tested on OCP 4.22 - Redis comes up 1/1 Running with nonroot-v2. As a result, the podSecurityContext conditional in deployment-redis.yaml and the values-pgo.yaml override both revert entirely.

no_log: true (deploy.yml:64) - added to the secret creation task.

stringData coercion (deploy.yml:63) - values are now piped through map('string') so numeric/bool-looking values are coerced to strings before being written to the secret.

Secret name (deploy.yml:56) - verified via helm template that {{ ocp_namespace }}-mcp-stack-gateway-secret resolves correctly for the contextforge release. The mcp-stack.fullname helper outputs contextforge-mcp-stack only when the release name doesn't already contain the chart name - since contextforge doesn't contain mcp-stack, the full name is contextforge-mcp-stack and the secret is contextforge-mcp-stack-gateway-secret, which matches.

gandhipratik203 (dismissed):
LGTM!

gandhipratik203 (dismissed):
LGTM!

msureshkumar88 (changes requested):
Thanks for the follow-up on the SCC and `no_log` fixes — both look good, confirmed clean against `main`. A couple of items from the earlier review still need work before this is ready, plus one new one:

**Blocking**

1. **Secret name still doesn't match the chart's `mcp-stack.fullname` helper** (`ansible/ocp/playbooks/deploy.yml:56`). The hardcoded `{{ ocp_namespace }}-mcp-stack-gateway-secret` only lines up with what Helm renders for the default `contextforge` release name. Tried `helm template my-mcp-stack-app charts/mcp-stack -f charts/mcp-stack/profiles/ocp/values-pgo.yaml -s templates/secret-gateway.yaml` and the chart renders `my-mcp-stack-app-gateway-secret`, while this playbook pre-creates `my-mcp-stack-app-mcp-stack-gateway-secret`. Helm won't adopt the mismatched secret, so the original race from #6195 (migration hook starting before Helm's own secret exists, falling back to `__REPLACE_ME__`) can still happen for any `ocp_namespace` that contains the chart name. Since this is the actual bug the PR is fixing, worth deriving the name from the chart (e.g. `helm template ... --show-only templates/secret-gateway.yaml`, or exposing the computed name as a chart value) rather than hand-reimplementing the fullname rule in Ansible — that also avoids the two staying in sync manually going forward.

2. **Missing `no_log: true` on the secret-loading task** (`ansible/ocp/playbooks/deploy.yml:44`, CWE-532). The `Load secrets values file` (`include_vars`) task that reads the raw secret material doesn't have `no_log: true`, even though the task just below it (writing the k8s secret) was fixed for this in the last round. Running the playbook with `-v` would print the loaded `ansible_facts` — JWT keys, admin passwords, etc. — to console/CI logs.

**Worth a look (non-blocking)**

- `deploy.yml:63` — values are coerced with Jinja's `string` filter, which renders booleans as `"True"/"False"`, while Helm's own `quote`/Sprig rendering lowercases to `"true"/"false"`. Could cause a mismatch (spurious helm diffs, inconsistent env var casing) if any `values-pgo-secrets.yaml` entry is an unquoted boolean.
- `deploy.yml:63` — `dict2items` is called twice (once for keys, once for values) and zipped back together; a `set_fact` computing it once would be a bit more robust and readable.
- `README.md` prerequisites section still doesn't mention the SCC-admin privilege now needed for `oc adm policy add-scc-to-user nonroot-v2` — flagged before, still open.

Rest of the diff (SCC change, migration/deployment reverts) looks solid and appropriately scoped to the OCP tooling — no Alembic or app-code changes needed here, which is the right call for this fix.

claudia-gray:
@msureshkumar88 Thanks Suresh for that review. 

Thanks for the detailed breakdown. All five items are addressed in the latest edit:
**Secret name:** now derived via helm template --show-only templates/secret-gateway.yaml and parsed with from_yaml — no longer hardcoded, works for any release name.

**no_log:** true on include_vars: present on the Load secrets values file task (line 44).

**Boolean coercion:** regex_replace handles True→true and False→false after the string filter.

**dict2items once:** split into a separate set_fact task (gateway_secret_items) so dict2items is called once and reused.

**Prerequisites docs:** openshift-pgo.md prerequisites now state cluster-admin is required and explains the oc adm policy add-scc-to-user nonroot-v2 context.

gandhipratik203 (dismissed):
LGTM!

msureshkumar88 (dismissed):
Re-reviewed against actual commits (not just prior comment claims) — all 5 previously-requested changes verified implemented:

1. Secret name derived via `helm template --show-only templates/secret-gateway.yaml` + `from_yaml` — verified correct regardless of release name.
2. `no_log: true` on the vars-load task.
3. Boolean casing (`True`/`False` → `true`/`false`) handled via `regex_replace`.
4. `dict2items` computed once, reused.
5. Docs updated with the cluster-admin requirement for the `nonroot-v2` grant.

Fresh review, verified empirically against the chart and playbooks:

- Fix correctly targets the root cause in #6195 (pre-install hook secret race + `anyuid`/UID 999/seccompProfile SCC conflict). No Alembic/schema changes — appropriate, since this is Ansible/Helm deploy tooling, not app code.
- `nonroot-v2` is meaningfully narrower than the `anyuid` SCC the issue proposed — good call.
- No unrelated changes; issue is linked correctly.

One non-blocking follow-up worth a fast-follow: the pre-created secret's `stringData` (`deploy.yml:69`) is built only from the `ocp_secrets` file rather than from the already-computed `gateway_secret_template.stdout` (the full Helm-merged render). This is latent today — it covers the exact keys #6195 needs — but would drift if an operator later overrides one of the other chart-default secret keys via `values-pgo.yaml` instead of the secrets file. Sourcing `stringData` from `gateway_secret_template.stdout` directly would close that gap and also removes the `True`/`False` string-coercion step.

Approving — no blocking issues.

msureshkumar88 (approved):

Issue #3855, opened 2026-03-25
[BUG][UI]: Admin UI Tool "Invoke" button non-functional — invokeTool() undefined


rakdutta:
The Admin UI Tool "Test" and "Invoke" action buttons are completely non-functional. When users click these buttons and try to test a tool, they receive a JSON-RPC error `"Invalid method"` with code `-32000`. The root cause is missing
  JavaScript function definitions (`invokeTool()` and `runToolTest()`) that are referenced by onclick handlers but never defined in the template.                                                                                                   
  
  ---                                                                                                                                                                                                                                               
                                                            
  ### 🧩 Affected Component                                                                                                                                                                                                                         
  Select the area of the project impacted:
                                                                                                                                                                                                                                                    
  - [ ] `mcpgateway` - API                                  
  - [x] `mcpgateway` - UI (admin panel)
  - [ ] `mcpgateway.wrapper` - stdio wrapper                                                                                                                                                                                                        
  - [ ] Federation or Transports
  - [ ] CLI, Makefiles, or shell scripts                                                                                                                                                                                                            
  - [ ] Container setup (Docker/Podman/Compose)             
  - [ ] Other (explain below)                                                                                                                                                                                                                       
                                                            
  ---                                                                                                                                                                                                                                               
                                                            
  ### 🔁 Steps to Reproduce                                                                                                                                                                                                                         
  
  1. Start ContextForge server: `make dev` or `make serve`                                                                                                                                                                                          
  2. Navigate to Admin UI at `http://localhost:4444/admin`  
  3. Go to the **Tools** section                                                                                                                                                                                                                    
  4. Create a REST tool:
     ```json                                                                                                                                                                                                                                        
     {                                                      
       "name": "test-tool",                                                                                                                                                                                                                         
       "url": "https://webhook.site/0f547174-a653-4ec1-bc97-278db9f51c5f",
       "integration_type": "REST",                                                                                                                                                                                                                  
       "request_type": "POST",
       "description": "Test webhook tool",                                                                                                                                                                                                          
       "input_schema": {                                    
         "type": "object",                                                                                                                                                                                                                          
         "properties": {                                    
           "message": {                                                                                                                                                                                                                             
             "type": "string"                               
           }                                                                                                                                                                                                                                        
         },
         "required": ["message"]                                                                                                                                                                                                                    
       }                                                    
     }
  5. Click the "Test" button (or "Invoke" button) for the tool
[...]

PR #6200, opened 2026-08-12
Fix/truncation on large responses

### LINKED ISSUE #6199: [BUG]: Tools with outputSchema fail validation on large responses (due to truncation)
### 🐞 Bug Summary

Tools with `outputSchema` validation intermittently fail with `"Output validation error: outputSchema defined but no structured output returned"` when they return large responses (>REST_RESPONSE_TEXT_MAX_LENGTH) that get truncated during JSON parsing.

**Root Cause:** When JSON parsing fails due to truncation, `_handle_json_parse_error()` in `tool_service.py` returns a `dict` instead of `list[TextContent]`. This dict is treated as valid data rather than an error, goes through validation, fails to produce `structured_content`, and causes the MCP SDK to reject it.

---

### 🧩 Affected Component

- [x] `mcpgateway` - API
- [ ] Other (explain below)

**Specific file:** `mcpgateway/services/tool_service.py` (function `_handle_json_parse_error`)

---

### 🔁 Steps to Reproduce

1. **Create a REST tool with `outputSchema`:**
   ```json
   {
     "name": "large_response_tool",
     "integration_type": "REST",
     "url": "https://api.example.com/logs",
     "output_schema": {
       "type": "object",
       "properties": {
         "results": {"type": "array"},
         "total": {"type": "integer"}
       },
       "required": ["results", "total"]
     }
   }
   ```

2. Configure tool to return large JSON response (>5000 characters, exceeding REST_RESPONSE_TEXT_MAX_LENGTH default)
3. Invoke the tool (via MCP client or API)
4. Observe intermittent failure:
- When response < 5000 chars: ✅ Works fine
- When response > 5000 chars: ❌ "Output validation error: outputSchema defined but no structured output returned"

---

### 🤔 Expected Behavior
When a tool response is truncated and JSON parsing fails, the gateway should:

1. Mark the result as an error (is_error=True)
2. Return an error message explaining the truncation
3. Skip output schema validation (per MCP spec, errors don't validate against outputSchema)
4. Never produce "Output validation error: outputSchema defined but no structured output returned"

---

### 📓 Logs / Error Output
Error from MCP SDK:
`Output validation error: outputSchema defined but no structured output returned`

Gateway logs show:
`WARNING: Failed to parse JSON response: Unterminated string starting at: line 1 column 4944`

The Bug:
In mcpgateway/services/tool_service.py around line 612:

```
def _handle_json_parse_error(response, error, is_error_response: bool = False) -> dict:
    # ... truncation logic ...
    text = response.text[:max_length]
    return {"response_text": text}  # ❌ Returns dict (treated as valid data!)
```

This dict then goes through the validation flow.

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Version or commit | `v1.0.7` |
| Runtime | `Python 3.12` |
| Platform / OS | `Fedora 43` |
| Container | `none` |

---

Relates to https://github.com/IBM/mcp-context-forge/issues/4208
- Lenient mode exposes this issue where errors aren't marked as `is_error=True`

ChrisPC-39:
## 🔗 Related Issue

Closes #6199

---

## 📝 Summary

**Fixes intermittent validation error when tools with `outputSchema` return large responses.**

**The Bug:** When a REST tool response exceeds `REST_RESPONSE_TEXT_MAX_LENGTH` (default 5000 chars), the gateway truncates it during JSON parsing. The `_handle_json_parse_error()` function returned a `dict` which was treated as valid data instead of an error. This caused the validation flow to attempt schema validation on unparseable truncated JSON, resulting in no `structured_content`, triggering the error:

Output validation error: outputSchema defined but no structured output returned
**The Fix:** Changed `_handle_json_parse_error()` to return `list[TextContent]` (an error signal) instead of `dict`. The existing error detection logic at line 5750 now catches this pattern and sets `is_error=True`, which skips output schema validation per MCP spec.

**Files Changed:**
- `mcpgateway/services/tool_service.py`

---

## 📏 Reviewability

---

## 🏷️ Type of Change

Example explained in issue #6199

- [x] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

| Coverage ≥ 80%            | `make coverage` |   pass    |

---

## ✅ Checklist


msureshkumar88:
## SDK v2 branch overlap check

Checked whether this fix overlaps with `chore/mcp-sdk-v2` (the dual-era MCP SDK 2.0 migration branch) before merging, since that branch also touches tool invocation and protocol handling.

**Finding: `chore/mcp-sdk-v2` still has the pre-fix `_handle_json_parse_error()`.**

At `mcpgateway/services/tool_service.py:614` on that branch, the function still returns a `dict` (`{"response_text": text}` on truncation, `{"error": ...}` on empty body) rather than `list[TextContent]`. On the success path, that dict gets passed straight into `extract_using_jq()` → `_coerce_to_tool_result()` → output schema validation as if it were real structured data. Since it never matches the tool's declared `outputSchema`, validation fails to produce `structured_content`, and the tool hits the same `RuntimeError: Output validation error: outputSchema defined but no structured output returned` this PR fixes on main.

**Not a protocol-version issue.** Grepped the surrounding invocation code on `chore/mcp-sdk-v2` for `protocol_version`/`negotiat`/`legacy` branching — there is none near this path. Protocol negotiation (legacy 1.x vs. the 2026-07-28 SDK 2.0 interop, landed in #6485) happens upstream at the transport/handshake layer. REST tool invocation and output-schema validation is a single shared code path underneath both protocol eras, so a client on either protocol version hits this bug identically once a REST tool's JSON response gets truncated.

**Why land this on `main` first rather than `chore/mcp-sdk-v2` directly:** the SDK v2 branch is moving fast (20+ commits since the migration landed, including protocol-negotiation work merged as recently as this week), so cherry-picking now risks conflict churn against work still in flight there. Landing the fix on `main` first, then porting it over when `chore/mcp-sdk-v2` next syncs with main (or via a tracked follow-up), avoids that churn while making sure the fix isn't silently dropped at merge time.

Filing a tracking note so this doesn't get lost — happy to also push the equivalent fix directly to `chore/mcp-sdk-v2` if that's preferred instead of waiting for a sync.

msureshkumar88:
## Correction to my previous comment

I went back and actually executed the code paths behind my [earlier comment](https://github.com/IBM/mcp-context-forge/pull/6200#issuecomment-5758041646). Three things in it were wrong, and I'd rather flag them than leave them sitting in the thread. Apologies for the noise.

### What holds up

- **`chore/mcp-sdk-v2` still carries the pre-fix handler.** `mcpgateway/services/tool_service.py:614` on that branch still returns a `dict` (`{"response_text": ...}` / `{"error": ...}`).
- **Behaviour parity between the branches is stronger than I claimed.** `_extract_and_validate_structured_content` (15,886 chars) and `_coerce_to_tool_result` (24,034 chars) are byte-identical across `main`, this PR's head, and `chore/mcp-sdk-v2`. Whatever `main` does on this path, sdk-v2 does identically.
- **Not protocol-version-specific.** No `protocol_version` / `legacy` branching anywhere near the REST invocation path on either branch; negotiation lives upstream at the transport/handshake layer.

### What was wrong

1. **"20+ commits since the migration landed"** — it's **11**. `git rev-list --count c5787b168..origin/chore/mcp-sdk-v2` → 11. I misread a `git log -20` limit as a commit count.
2. **"protocol-negotiation work merged as recently as this week"** — the feature commits are dated 2026-09-14 and the merge (`7043d20b2`) 2026-09-18. Imprecise of me.
3. **The failure mechanism I described, I could not reproduce.** This is the one worth a closer look.

### On the failure mechanism

I repeated the PR description's chain — truncated JSON → schema validation attempted → no `structured_content` → `Output validation error: outputSchema defined but no structured output returned` — as if I'd verified it. I hadn't. When I ran the pre-fix payload through the (unchanged) validator against an `outputSchema` of `{"type":"object","required":["items"]}`:

| Pre-fix payload | Observed outcome |
|---|---|
| `{"response_text": "<truncated>"}`, no jq filter | `structured_content` **is** extracted as `{"response_text": ...}` → schema violation (`'items' is a required property`) → `is_error=True` set in place → egress returns `CallToolResult` → SDK validator short-circuits. No `RuntimeError`. |
| jq `.response_text` | Same — `type` violation, `is_error=True`, short-circuit. |
| jq no-match | Reaches the lenient `structured is None → return True` branch, but only because my harness skipped `_is_jq_filter_error()`, which in the real flow sets `is_error=True` before validation runs. |

The validator's lenient branch is only reachable when the first `TextContent.text` isn't JSON-parseable. Since `{"response_text": ...}` always serialises to valid JSON, that branch isn't reached — the payload takes the schema-violation path instead, which sets `is_error=True`, and per the Validator-C notes in that method's own docstring an `isError=True` egress short-circuits the server SDK's output-schema check.

Worth noting the exact error string in the issue doesn't live in `tool_service.py` at all — it comes from the MCP server SDK, and the gateway-side references to it are at `mcpgateway/transports/streamablehttp_transport.py:1900` and `:2055`.

### Question for the author

None of this says the change is wrong — returning an error `TextContent` for an unparseable body looks more correct than synthesising a `{"response_text": ...}` object that was never the tool's real output, and that stands on its own merits.

But could you share the concrete repro for #6199 — the tool's `outputSchema`, its `jsonpath_filter` (if any), and the rough response shape? If the reported error arrives through a different route than the PR description states, the fix might be addressing an adjacent problem, and the actual trigger could still be live after merge. Happy to dig further with a repro in hand.

Method, for anyone who wants to re-check: payloads run directly through `_coerce_to_tool_result()` + `_extract_and_validate_structured_content()` on this PR's head, with the branch comparisons done via `git show <ref>:mcpgateway/services/tool_service.py`.

msureshkumar88 (changes requested):
Thanks for digging into this — the truncation/validation interaction here is genuinely subtle, and the writeup in the PR description does a good job walking through the MCP spec's error-handling rule and why the dict fallback was slipping past it. Changing `_handle_json_parse_error` to signal an error via `TextContent` instead of a plain dict is the right shape for a fix.

A few things worth resolving before this merges, though — mostly around scope and an interaction with a change that's landed on `main` since this branch was cut.

## Blocking

**1. This currently marks *every* non-JSON 2xx REST response as an error, not just ones with an `outputSchema`.**

The success branch now feeds the `list[TextContent]` from `_handle_json_parse_error` into `extract_using_jq`, which trips the existing "is this an error result" check and sets `is_error=True` / `success=False`. That happens regardless of whether the tool declares an `outputSchema` — so a REST tool that legitimately returns plain text, HTML, XML, or an empty body now reports as failed. That's a regression against #3855 (which specifically added support for non-JSON 2xx responses).

Would it work to scope the "treat as error" behavior to only fire when `tool_output_schema` is set? An unparseable body can never satisfy a declared schema, so erroring there matches the issue's ask — but a tool with no schema declared should keep succeeding on non-JSON bodies the way it does today.

`test_rest_tool_does_not_truncate_small_response` currently asserts `is_error is True` for a small plain-text 200 with no schema — that test is effectively locking in the regression, so it'd need to flip back once the scoping is fixed.

**2. After a rebase, this will likely crash (or silently swallow the new error message) on any tool with a `jsonpath_filter` set.**

`main` has moved `extract_using_jq`'s execution through the jq sandbox (`asyncio.to_thread(extract_using_jq, ...)` → `run_jq_filter` → `orjson.dumps(data)`). `TextContent` isn't JSON-serializable, so passing the `list[TextContent]` from `_handle_json_parse_error` through that path raises `TypeError: Type is not JSON serializable: TextContent`, which gets wrapped into a generic `JqFilterError` ("Error applying jsonpath filter"). That means on `main`, the detailed parse-error message this PR adds never actually reaches the caller for any tool with a filter configured — worth confirming once rebased, and probably worth a test with `jsonpath_filter` set + a parse failure to catch this going forward.

## Functionally impacting

**3. Worth double-checking against the actual repro in #6199.** I traced the reported error ("outputSchema defined but no structured output returned") through the current validator (`_extract_and_validate_structured_content`) with the old dict fallback, and for a schema with `required` fields it already surfaces as a schema-validation error (`'results' is a required property`), not the SDK message from the issue. The case that *does* still slip through silently is a schema with no `required` — the dict just gets accepted as `structured_content` with no validation failure. A regression test that reproduces the issue's exact scenario (large truncated body + the issue's schema) would help confirm this PR actually closes the gap, and might point toward the narrower fix in point 1 being sufficient.

**4. Error text changed on the 4xx/5xx and non-standard-2xx paths, which seem out of scope for this issue.** E.g. a 500 with a non-JSON body used to return `HTTP 500: <body>` and now returns `HTTP 500: JSON parse error: ...\n\nResponse:\n<body>`. If anything downstream matches on that message shape, this could be a surprise. Was this intentional, or a side effect of routing both paths through the same helper?

**5. The new error message text ("Increase this limit if tools return large JSON responses") could use a second look.** Since the truncation is happening upstream (the log's `Unterminated string ... column 4944` is below the 5000-char limit, and it happened after full JSON parsing, not before), raising `REST_RESPONSE_TEXT_MAX_LENGTH` wouldn't actually fix the underlying case — might be worth softening or removing that suggestion so it doesn't send folks down the wrong path.

## Security note

Worth hardening: the non-standard-2xx and error paths now echo more of the raw upstream body (up to `REST_RESPONSE_TEXT_MAX_LENGTH` chars) plus parser exception text back to the caller, where previously some of these returned a generic message. That's a larger surface for upstream response content to reach the calling client/agent (CWE-209/CWE-200 territory). If that's not intentional, scoping the message change back to just the truncation-signaling path (rather than all JSON-parse-error callers) would tighten it back up.

## Suggestions / minor

- Docs (`config.py`'s `rest_response_text_max_length` description, `.env.example`) still describe the setting the old way — if the semantics around it change, those comments should follow.
- A couple of test docstrings reference "#XXXX" instead of the actual issue number (#6199) — small thing, but nice for traceability.
- The `test_rest_tool_handles_json_error_with_error_key` / `..._without_error_key` tests and the related refactor of the 4xx/5xx branches look like they're covering ground outside this issue's stated scope — could be worth splitting into a separate PR if they're addressing something else, just to keep this one reviewable against #6199 specifically.
- `first_text = error_content[0].text if error_content else ""` — `_handle_json_parse_error` always returns exactly one `TextContent`, so the `else ""` branch looks unreachable; not a big deal, just flagging.

No blockers on the direction — just want the fix scoped to where the schema actually matters, and confirmed against `main` post-rebase before this lands. Happy to take another look once these are addressed!

ja8zyjits:
## 🚨 Blocking

- [ ] **Misleading docstring on `_handle_json_parse_error`**
  - File: `mcpgateway/services/tool_service.py:619`
  - Docstring says "The caller checks for this pattern and sets is_error=True", but the 200-path caller (line 6155) does not check directly — it relies on `_is_jq_filter_error(filtered_response)` after `extract_using_jq` runs. Misleads maintainers adding new call sites.
  - Fix: Document both valid caller patterns — (1) check return directly, or (2) pass through `_is_jq_filter_error` after `extract_using_jq`.

---

## ⚠️ Functionally-Impacting

- [ ] **2xx success path caller not updated after return-type change**
  - File: `mcpgateway/services/tool_service.py:6155`
  - `_handle_json_parse_error` return type changed from `dict` → `list[TextContent]`. Three callers exist. Two (lines 6117, 6144) were updated; line 6155 (2xx success path) was not. On that path, returned `list[TextContent]` is passed to `extract_using_jq`. When `jsonpath_filter` is **empty** (common), `_is_jq_filter_error` detects `TextContent` → `is_error=True` — correct. When `jsonpath_filter` is **set**, `run_jq_filter` serializes `TextContent` objects as JSON → silent wrong output instead of clean error.
  - Fix: On line 6155, after `result = _handle_json_parse_error(...)`, guard: `if _is_jq_filter_error(result): tool_result = ToolResult(content=result, is_error=True)` and skip `extract_using_jq`. Mirrors pattern at lines 6159–6162.

---

## 📋 Advisory

- [ ] **Missing test: 2xx + JSON parse failure + non-empty `jsonpath_filter`**
  - File: `tests/unit/mcpgateway/services/test_tool_service.py`
  - All three updated truncation tests set `mock_tool.jsonpath_filter = ""`. No test covers the jq-filter degradation path. Add a test with non-empty `jsonpath_filter` + 2xx JSON parse failure.

- [ ] **`#XXXX` placeholder in three test docstrings**
  - File: `tests/unit/mcpgateway/services/test_tool_service.py:6695`, `6736`, `6776`
  - `After fix for #XXXX:` should read `After fix for #6199:`. Fix: `s/#XXXX/#6199/g`.

- [ ] **Bare `list` return type — missing generic**
  - File: `mcpgateway/services/tool_service.py:614`
  - `-> list:` should be `-> list[TextContent]:` for type safety and static analysis.

- [ ] **Misleading inline comment**
  - File: `mcpgateway/services/tool_service.py` ~line 6119
  - `# Prepend HTTP status code to error message` is wrong — code constructs a new `TextContent(f"HTTP {response.status_code}: {first_text}")`, replacing the list element, not prepending to a string.
  - Fix: `# Wrap error message with HTTP status prefix`

- [ ] **Duplicate string serialization ternary** *(suggestion)*
  - File: `mcpgateway/services/tool_service.py:6114`, `6141`
  - `error_val if isinstance(error_val, str) else orjson.dumps(error_val).decode()` repeated verbatim in non-2xx and non-standard-2xx error branches.
  - Fix: Extract to a helper or local lambda.

---

msureshkumar88:
The jq-path finding was a good catch. Following it up uncovered two more problems on the same code path. All items are addressed on the branch now.

**Addressed**
- **Docstring (blocking):** Rewritten. After the change below, every caller uses the result directly as error content. The docstring states that and warns against passing the result to `extract_using_jq`.
- **2xx path with a jq filter:** Confirmed and fixed. The effect differed slightly from "silent wrong output": `orjson` cannot serialize `TextContent`, so `run_jq_filter` raised. The caller got `is_error=True` with the generic text "Error applying jsonpath filter", and the truncation message was lost. The parse error now returns before jq runs. A new test covers a non-empty `jsonpath_filter`.
- `#XXXX` is now `#6199`, the return type is `list[TextContent]`, and the "Prepend" comment is reworded.

**Also changed (not in the review)**
- **Plain-text 2xx responses:** The PR made every non-JSON 2xx body an error. That included plain-text and HTML responses from tools without an `outputSchema`, which worked before. The error now applies only to tools that declare an `outputSchema`, the case #6199 describes.
- **JSON array bodies:** `result.get("error")` on the 203/207 path turned a valid JSON array into a misleading "JSON parse error". Both error branches now read `"error"` only from JSON objects.
- **Error message advice:** The message told operators to raise `REST_RESPONSE_TEXT_MAX_LENGTH`. That limit applies only after parsing fails, so raising it cannot help. The message now reads "Response body is not valid JSON: ... Showing the first M of N characters".
- **Non-2xx JSON without an `error` key:** The PR echoed `orjson.dumps(result)`. That raised on integers above 64 bits and had no size limit. It now echoes the raw text, truncated to the limit.
- **Live-gateway test:** `tests/live_gateway/e2e/test_rest_non_json_output_schema_e2e.py` starts its own gateway and calls the tool through a virtual server. It passes on this branch and fails on `main`.

**Left as is**
- The duplicated `isinstance(error_val, str)` ternary: `main` already has one per branch, and the PR does not add more.

ja8zyjits:
@msureshkumar88 

Suggested PR changes:

1. **Correct scope claim.** Code calls `response.json()` before text truncation. Large valid JSON does not fail because it exceeds 5,000 characters. Update PR description to say fix handles *invalid JSON returned by schema tools*, or provide reproduction showing where truncation happens before parsing.

2. **Strengthen E2E test.** Set explicit `REST_RESPONSE_TEXT_MAX_LENGTH` for gateway subprocess. Assert returned error says body was truncated and contains only configured number of body characters. Current test passes for invalid JSON even if no truncation occurs.

3. **Add large valid JSON test.** Return valid body above limit with matching `outputSchema`; assert successful structured output. This establishes boundary between size and parse failure.

Optional: record HTTP request before writing response in E2E handler to remove narrow assertion race.

Issue #5823, opened 2026-07-24
[CHORE]: Documentation pass for Rust removal (architecture docs, ADR-043, AGENTS.md validation gate)


jonpspri:
Part of EPIC #5817 (Remove Rust from the ContextForge codebase). Final step; closes the EPIC. Documentation was explicitly deferred during #5737 review — this is the dedicated pass.

### Chore Summary

Bring all documentation in line with the completed Rust removal:

- `docs/docs/architecture/rust-mcp-runtime.md`: remove or replace with a deprecation/removal notice
- `docs/docs/architecture/adr/043-rust-mcp-runtime-sidecar-mode-model.md`: mark superseded (record the removal decision, don't delete the ADR — ADRs are historical)
- `docs/docs/deprecations.md`: move the Rust MCP runtime entry from "deprecated/sunset" to "removed"
- **`AGENTS.md`**: update the pre-merge validation gate (step 4 currently runs `make docker-nuke docker-prod-rust testing-up RUST_MCP_MODE=`), the project-structure section (`crates/` listing), the Rust code-quality commands, and the `crates/mcp_runtime/DEVELOPING.md` pointer
- `README.md`, `DEVELOPING.md`, `MULTIPLATFORM.md`, `SECURITY.md`: scrub Rust build/runtime references
- `docs/docs/architecture/multitenancy.md` and any RBAC/security docs referencing the Rust session-auth-reuse path

### Scope

- [ ] All of the above updated; `git grep -rin "rust mcp\|mcp_runtime\|RUST_MCP_MODE" docs/ README.md AGENTS.md DEVELOPING.md MULTIPLATFORM.md SECURITY.md` returns only historical/ADR references
- [ ] ADR-043 marked superseded with a pointer to this EPIC
- [ ] Docs build (`make` docs target, if any) passes

### Related

- EPIC: #5817

Issue #5825, opened 2026-07-24
[EPIC][CHORE]: Test suite cleanup — dead skipped tests and warning noise


jonpspri:
### Epic Summary

`make test` currently reports **797 skipped tests and 511 warnings** (20,845 passing). Analysis of a full run shows the skips and warnings cluster into a small number of buckets, most of which are removable or fixable with focused, mechanical PRs. This EPIC tracks that cleanup.

### Skip taxonomy (797 total)

| Bucket | Count | Disposition |
|--------|-------|-------------|
| PostgreSQL-conditional (`Row-level locking only works on PostgreSQL`, `TEST_POSTGRES_URL not set`, external-DB guards) | ~62 | **Keep** — legitimate environment guards |
| Dagger not installed | 20 | Resolved by #5824 (cforge/builder removal) |
| `altk` not available (optional plugin dep) | 17 | **Keep** — legitimate optional-dependency guard |
| KEYCLOAK_URL / helm / psycopg / otel env guards | ~11 | **Keep** — legitimate environment guards |
| Dead skips: "covered by other tests", "not implemented", "unconditional skip", stale auth, live-instance/external-connectivity | ~40 | **Remove/relocate** — child issue below |
| Flaky-in-parallel RBAC middleware tests (plugin manager singleton) | 9 | **Fix or delete** — child issue below |

### Warning taxonomy (511 total)

| Bucket | Count | Disposition |
|--------|-------|-------------|
| Deprecated cpex plugin payload `.headers` field (`ToolPreInvokePayload`, `HttpAuthResolveUserPayload`, `AgentPreInvokePayload`, `HttpPreRequestPayload`) | ~85 | Migrate tests to `extensions.http.headers` — child issue below |
| `ValidationMiddleware` deprecation (past its 2026-07-07 sunset) | 37 | Preferred resolution is middleware removal — **pending deprecation review**, see below |
| JWT `InsecureKeyLengthWarning` (short test keys) | 7 | Fix test fixtures — child issue below |
| Misc: Pydantic class-config deprecation, sqlite datetime adapter, redis `setex`, `HTTP_422_*` constant, unawaited coroutine, httpx/starlette testclient | ~6 | Fix call sites — child issue below |

### Deprecation candidates surfaced (filed as INDEPENDENT issues — not children of this EPIC)

The skipped/warning analysis pointed at features warranting their own removal issues. These are filed standalone and are deliberately **not** children of this testing EPIC:

[...]

Issue #5822, opened 2026-07-24
[CHORE]: Delete crates/mcp_runtime and Rust workspace scaffolding


jonpspri:
Part of EPIC #5817 (Remove Rust from the ContextForge codebase). Step 7 of the sequence; lands after all Python integration, tests, and build/CI references are removed, making this a near-zero-risk pure deletion.

### Chore Summary

Delete the deprecated Rust MCP runtime crate and the Cargo workspace scaffolding:

- `crates/mcp_runtime/` (~16.7k lines: `src/{main,lib,config,backend_url_validator,observability}.rs`, `tests/{cli,runtime}.rs`, plus its README/STATUS/TESTING-DESIGN/DEVELOPING docs)
- `Cargo.toml`, `Cargo.lock` (workspace root)
- `rust-toolchain.toml`, `deny.toml`
- `supply-chain/` (cargo-vet configuration)

The crate's own README declares it deprecated 2026-06-11, sunset 2026-07-07. If the sibling removals (#5814 stdio wrapper, #5816 masking extension) have already landed, this is the last crate standing and the workspace files die with it; if not, this issue covers only `crates/mcp_runtime` and the workspace files adjust accordingly.

### Pre-merge verification

- [ ] `git grep -rn "mcp_runtime\|cargo\|rustup" -- ':!docs' ':!*.md'` returns no live references (docs handled by the follow-up docs issue)
- [ ] `make test`, `make verify`, and a clean container build pass with the crate absent

### Related

- EPIC: #5817

jonpspri:
Upgrading this to MUST for iteration 22 as it's causing CWE-601 scan failures in a sister project.  Please review Steps 1-6 to determine if any other steps should also be executed first to get to this point.

PR #6201, opened 2026-08-12
chore: remove stdio wrapper in favor of FastMCP

### LINKED ISSUE #5814: Remove stdio wrapper from Python and Rust codebases in favor of FastMCP
### Chore Summary

ContextForge currently maintains two hand-rolled stdio wrapper implementations:

- **Python**: `mcpgateway/wrapper.py` (and the older `mcpgateway/translate.py` stdio↔SSE/streamable-HTTP bridge)
- **Rust**: `crates/wrapper/` — the `mcp-stdio-wrapper` binary bridging stdio to SSE/HTTP endpoints

Both should be removed in favor of [FastMCP](https://github.com/jlowin/fastmcp), which offers a superior wrapper experience.

### Prior Art: #5737

This work was previously attempted as part of #5737 (*"refactor: remove Rust build infrastructure and stdio wrapper capability"*), which was closed unmerged. The maintainer feedback on that PR explicitly requested exactly this decomposition:

> "We want more focused PRs that try to do 1 thing and 1 thing only, and we should have issues associated with any changes like this to see where the work came from and the value it brings." — @brian-hussey

This issue is the tracking issue for the **stdio-wrapper slice** of that work, intended to land as its own focused PR, independent of the broader Rust build-infrastructure removal.

### Rationale

FastMCP provides a production-grade stdio↔HTTP/SSE proxy via `create_proxy()` / `FastMCP.as_proxy()`, plus a zero-config CLI bridge (`uvx fastmcp-remote <url>`). Out of the box it covers everything our custom wrappers re-implement by hand:

- One-line transport bridging (stdio ↔ HTTP ↔ SSE)
- Per-request session isolation
- Automatic forwarding of sampling, elicitation, roots, logging, and progress
- Configuration-driven multi-server composition with automatic namespacing and component prefixing
- Protocol-era mirroring and component-list caching

FastMCP is actively maintained under PrefectHQ (latest release `v3.4.4`, 2026-07-09; ~26.8k stars; Apache-2.0).

### References

- Proxy / wrapper documentation: https://gofastmcp.com/servers/providers/proxy
- `fastmcp-remote` CLI (stdio bridge for remote servers): https://gofastmcp.com/clients/fastmcp-remote
- Multi-server composition: https://gofastmcp.com/servers/composition
- Source repository: https://github.com/jlowin/fastmcp

### Scope

Drawn from the stdio-wrapper portion of #5737's diff:

- [ ] Remove `mcpgateway/wrapper.py` and assess `mcpgateway/translate.py` for removal in the same pass
- [ ] Remove `crates/wrapper/` (Rust `mcp-stdio-wrapper`)
- [ ] Remove the `wrapper.yml` CI workflow
- [ ] Remove wrapper unit tests, e2e tests, and skipped test stubs
- [ ] Update `admin_ui/configExport.js` to drop stdio config generation
- [ ] Update client-facing docs (`README.md`, `AGENTS.md`, `docs/`) to use Streamable HTTP (`type: http`, `/mcp/`) as the canonical transport and point remaining stdio-wrapping use cases to FastMCP

Yosiefeyob:
# Pull Request

## 🔗 Related Issue

Closes #5814

---

## 📝 Summary

Removes the hand-rolled stdio wrapper — Python and Rust — in favour of FastMCP, and
switches the client documentation to Streamable HTTP.

**Deleted**

- `mcpgateway/wrapper.py` (777 lines)
- `crates/wrapper/` — the Rust `mcp-stdio-wrapper` binary (51 files, ~1,988 lines)
- `.github/workflows/wrapper.yml` (263) and `tests/unit/test_wrapper_workflow.py`, which
  existed only to validate that workflow
- `tests/unit/mcpgateway/test_wrapper.py` (1,084)
- `docs/docs/using/mcpgateway-wrapper.md` and its `.pages` nav entry
- The wrapper-dependent Langfuse tests and the helper functions they orphaned

**Edited**

- `admin_ui/configExport.js` — dropped the stdio config-generation branch and its
  description strings
- `templates/admin.html` — removed the now-dead "Stdio (Claude Desktop, CLI)" export button
- `tests/unit/js/configExport.test.js` — removed the two stdio-specific tests; the seven
  incidental `"stdio"` arguments in unrelated tests switched to `"sse"`
- `Cargo.lock` regenerated, `supply-chain/config.toml` pruned
- 30 documentation files switched to Streamable HTTP (`servers` / `type: http` / `/mcp/`)

93 files, ~6,560 deletions.

`mcpgateway/translate.py` was assessed and deliberately **not** removed — see Notes.

---

## 📏 Reviewability

---

## 🏷️ Type of Change

- [ ] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

Live tests were run against a container **rebuilt from this branch**, not the released image.

| Check | Command | Status |
|---|---|---|
| Lint (`mcpgateway/`) | `ruff check mcpgateway/` | ✅ All checks passed |
| Docstring coverage | `make interrogate` | ✅ PASSED — 100% |
| JS tests | `vitest run tests/unit/js/configExport.test.js` | ✅ 22 passed |
| Lockfile matches CI | `cargo metadata --locked` | ✅ Exit 0 |
| Supply chain | `cargo vet --locked` | ✅ Succeeded (94 exemptions, down from 121) |
| Unit tests | `make test` | ✅ 21,767 passed — 0 regressions |
| Coverage | `make coverage` / `diff-cover` | ✅ 98% total; diff-cover reports no covered lines (pure deletion) |
| Container build | `make docker-prod-rust` | ✅ Rust workspace still builds without `crates/wrapper` |
| RBAC | `make test-mcp-rbac` | ✅ 40 passed |
| Tools benchmark (300 users) | `make benchmark-mcp-tools` | ✅ No regression vs. baseline image |

**On `make test`:** the run reports 31 failures, all SSRF/CSRF URL-validation tests that
fail identically on unmodified `main` with the same local `.env` — verified by diffing
failure sets against a clean worktree.

---

## ✅ Checklist

- [x] Code formatted (`ruff check mcpgateway/` clean)
- [x] Documentation updated — 30 files switched to Streamable HTTP
- [x] No secrets or credentials committed (`.secrets.baseline` regenerated; the only
      removals are entries for the two deleted files)

---

### Langfuse tests — 4 migrated, 2 removed

Six of the nine Langfuse tests reached the gateway through the wrapper or mcp-cli. Four were migrated to Streamable HTTP: `tool_call`, `prompt_render`, `sanitized_prompt_error`, and `resources/list`.

Two were removed with no replacement: `initialize`, and the mcp-cli `tool_list` variant, whose `config_file` fixture launched the wrapper.

The three network-based tests are untouched — the file now holds 7.

Migrated tests use `_lookup_server_id` + `_send_jsonrpc_http`, which performs the `initialize` handshake and carries the resulting `mcp-session-id`; without it the gateway rejects the call with `-32600 Missing session ID`.

### Documentation

Every client guide moved to `type: http` + `/mcp/`, following the shape already published in `copilot.md`.

Sections that existed only to launch the wrapper — Docker/pipx alternatives, wrapper env-var tables, and "local stdio bridge" blocks — were removed rather than rewritten; a broader docs polish feels like #5823's territory.

### Supply chain

Deleting `crates/wrapper` dropped 45 packages from `Cargo.lock`, orphaning 20 cargo-vet exemptions in `supply-chain/config.toml`. Both cleaned up here.

`cargo metadata --locked`, `cargo check --workspace --locked`, and `cargo vet --locked` pass.

### `translate.py` — recommending it stays

The issue asks to assess it for removal. It's 2,469 lines to the wrapper's 777, referenced widely across `mcpgateway/` and `tests/`, and does more than stdio bridging — `translate_grpc` covers gRPC exposure that FastMCP doesn't.

Its siblings have live callers (`services/grpc_service.py` → `translate_grpc`, `baggage.py` → `translate_header_utils`), so removing it means untangling live code rather than deleting a bridge. It's also the README quick-start command.

Happy to open a separate issue if you agree.


jonpspri:
We don't need tests of wrapped servers to be in-scope for the Langfuse test.  So long as via-network tests of StreamableHTTP invocations exist we can remove the wrapped variants.

Assume all clients support native StreamableHTTP.

If the unused test file is not in scope for #5825, file a separate independent CHORE issue to address it.

lucarlig (changes requested):
Good focused cleanup, but requesting changes:

- Fix client docs: Claude Desktop cannot load local HTTP servers from `claude_desktop_config.json`; Cline requires `mcpServers` + `streamableHttp`; mcp-cli requires `mcpServers`; Continue still uses legacy SSE config; Inspector requires `--server-url` + `--transport http`, not `--url`.
- Port Langfuse tool, prompt, and secret-redaction checks to Streamable HTTP instead of deleting their only coverage.
- Remove unused workspace `futures` dependency; deleted wrapper was its last consumer.
- Remove or document orphaned `.env.example` `# CONCURRENCY=10`.
- Remove remaining wrapper references in `DEVELOPING.md`, ArgoCD diagram, unit-test inventory, and generated architecture diagrams.

Local `cargo metadata --locked`, `cargo check --workspace --locked`, and `cargo vet --locked` pass.

msureshkumar88 (changes requested):
Nice cleanup — the scope discipline here is good, especially deliberately leaving `translate.py` alone with a clear rationale instead of scope-creeping into a second removal. Verified the 96-file diff traces entirely to the wrapper-removal scope, no alembic/schema touch needed, and the dead-code cleanup in `mcp_test_helpers.py` (removed helpers, confirmed no remaining callers anywhere in `tests/`).

One thing worth fixing before merge, plus two optional nits.

**Coverage gap not reflected in the PR notes**

`test_langfuse_trace_export_eventually_contains_resource_list_trace` was removed with no replacement and isn't mentioned in the "Notes" section. The notes say `initialize`, `tool_call`, and `prompt_render` lost their only coverage — but `tool_call` and `prompt_render` were actually migrated to `_send_jsonrpc_http` and are still asserted (lines 315, 345). The trace that genuinely lost all coverage is `resources/list`, which isn't called out anywhere.

Since `tool_call`/`prompt_render` already show the migration pattern works (`_lookup_server_id` + `_send_jsonrpc_http`), restoring the `resources/list` variant the same way seems like the smaller lift versus leaving the gap — but if you'd rather drop it intentionally, updating the notes so reviewers see the real trade-off works too.

**Optional / non-blocking**

- `docs/docs/using/clients/mcp-cli.md` Quick Start Checklist still has "Create `server_config.json` with correct Python path" (line ~506) — leftover from the stdio config format; the example above it is now pure `type: http` with no Python path field.
- `mcpgateway/admin_ui/configExport.js` — JSDoc on `generateAndShowConfig` (line 59) and `exportServerConfig` (line 100) still lists `'stdio'` as a valid `configType`. No callers pass it anymore, just stale doc comment.

Everything else — RBAC/security surface, dead code, breaking-change disclosure, alembic — looks clean.

Yosiefeyob:
Thanks both — pushed, and rebased on current main.

@msureshkumar88

**Coverage gap** — `resources/list` is restored using the same `_lookup_server_id` + `_send_jsonrpc_http` pattern as the other migrated tests, so it's asserted again rather than left as a gap.

The Notes section is corrected too: it was written before the `tool_call`/`prompt_render` migration and never revisited, so it described a state that had stopped being true. The file now holds 7 tests — four migrated, two removed (`initialize` and the mcp-cli `tool_list` variant, whose fixture launched the wrapper).

**mcp-cli Quick Start checklist** — no longer mentions a Python path; it now matches the `type: http` example above it.

**configExport.js JSDoc** — `'stdio'` dropped from both lines.

@lucarlig

**Client docs** — all five points addressed:

- Claude Desktop now routes to Custom Connectors, with a note that `claude_desktop_config.json` is stdio-only.
- Cline uses `mcpServers` + `streamableHttp`.
- mcp-cli uses `mcpServers`.
- Continue moved off legacy SSE.
- Inspector uses `--server-url` + `--transport` (there was a third `--url` further down the page that I'd missed).

**Remaining wrapper references** — cleared from `DEVELOPING.md`, the ArgoCD doc, the unit-test inventory, and both generated diagrams.

For the diagrams I removed the `mcpgateway.wrapper` node directly rather than regenerating, since a full regen rewrites ~29k lines (they were last generated in February). Happy to regen instead if preferred.

One thing worth flagging on the Langfuse ports: as first written they posted the JSON-RPC payload without an `initialize` handshake, so a live gateway rejected them with `-32600 Missing session ID`. The helper now performs the handshake and replays on that session, so those tests genuinely run rather than just existing — worth knowing since `tests/live_gateway/` is excluded from the default pytest run.

@jonpspri

On the unused test file: #5825's scope is the Python suite (pytest skips and warnings), so it doesn't cover this. `tests/js/` — 32 files — is excluded from `vitest.config` entirely and never runs via `make test-js` or CI. Filed as its own CHORE issue (#6452) rather than widening this PR.

docs/docs/using/clients/continue.md
```diff
@@ -30,25 +30,20 @@ tools you publish through **ContextForge** - no plug-in code required.
...
 
 ```jsonc
 // ~/.continue/config.json
 {
   "experimental": {
     "modelContextProtocolServer": {
       "transport": {
-        "type": "sse",
-        "url": "http://localhost:4444/servers/UUID_OF_SERVER_1/sse",
+        "type": "http",
```
lucarlig: Please use Continue’s current MCP format: top-level mcpServers (or .continue/mcpServers/*.json) with type: streamable-http. This example will not load as written.

docs/docs/using/clients/claude-desktop.md
```diff
@@ -1,111 +1,65 @@
...
-*(Mac / Windows users should replace `localhost` with `host.docker.internal`.)*
+> *Use the real server ID from the Admin UI instead of `UUID_OF_SERVER_1`.*
 
 ---
 
-### ⚡ pipx / uvx one-liner (wrapper already installed)
+## 🔑 Authentication
 
-If you installed the package globally:
+Claude prompts for credentials when the connector is added — the Gateway's bearer token is
```
lucarlig: Custom Connectors are remote and use OAuth or authless flows; localhost plus a generated Gateway JWT is not supplied this way. Please document an externally reachable OAuth-enabled server or a supported local bridge.

docs/docs/testing/acceptance.md
```diff
@@ -180,11 +168,10 @@ graph TB
 {
   "mcp.servers": {
     "gateway-virtual": {
-      "command": "python",
-      "args": ["-m", "mcpgateway.wrapper"],
-      "env": {
-        "MCP_SERVER_URL": "$GW_URL/servers/$TIME_SERVER_UUID",
-        "MCP_AUTH": "$MCPGATEWAY_BEARER_TOKEN"
+      "type": "http",
```
lucarlig, line 171: This is now direct HTTP, but the section heading and other acceptance/index entries still reference the removed wrapper. Please update those remaining references.

docs/docs/using/clients/mcp-cli.md
```diff
@@ -76,49 +76,14 @@ make venv install serve
...
-        "mcpgateway-wrapper",
-        "mcpgateway-wrapper"
-      ],
-      "env": {
-        "MCPGATEWAY_BEARER_TOKEN": "your-jwt-token-here"
+    "contextforge": {
+      "type": "http",
+      "url": "http://localhost:4444/servers/UUID_OF_SERVER_1/mcp/",
+      "headers": {
+        "Authorization": "Bearer <YOUR_AUTH_TOKEN_HERE>"
```
lucarlig, line 86: Please also update the Environment Variables section: MCP_AUTH and MCP_SERVER_URL are still exported there, but this HTTP config does not consume them.

lucarlig (changes requested):
The wrapper removal looks good overall. Please address these documentation issues before merging.

docs/docs/using/clients/continue.md
Yosiefeyob: Switched to `mcpServers` with `type: streamable-http`, and moved auth under
`requestOptions.headers` (Continue doesn't read a top-level `headers` key).

One correction worth flagging: I tested this against Continue 2.0.0 and
`.continue/mcpServers/*.json` didn't load — the Tools panel reported "No MCP servers
configured" for both an object-map and a list shape. What works is an `mcpServers` list
in `~/.continue/config.yaml`; Continue then shows the server connected with the
gateway's tools. The extension banners "Extension configuration is local only as of
v2.0.0", so the `.continue/mcpServers/` path looks like 1.x guidance.

Config the page now documents, verified loading against a live gateway:

```yaml
mcpServers:
  - name: contextforge
    type: streamable-http
    url: http://localhost:4444/servers/UUID_OF_SERVER_1/mcp/
    requestOptions:
      headers:
        Authorization: Bearer <YOUR_AUTH_TOKEN_HERE>

File: docs/docs/using/clients/claude-desktop.md (hunk 2 above)
Comment (Yosiefeyob) [reply docs/docs/using/clients/claude-desktop.md]: Went with the local bridge option. The page now uses `claude_desktop_config.json` with FastMCP's `fastmcp-remote` in the deleted wrapper's place:

```json
{
  "mcpServers": {
    "contextforge": {
      "command": "uvx",
      "args": [
        "fastmcp-remote",
        "http://localhost:4444/servers/UUID_OF_SERVER_1/mcp/",
        "--header",
        "Authorization: Bearer <YOUR_JWT_TOKEN>"
      ]
    }
  }
}
```
Same file and command/args shape as before — only the program changes. Verified with Claude Desktop running against a live gateway: Connected to contextforge (1 tools). The Custom Connector text and the bearer-token prompt step are gone.

While testing I hit something in `acceptance.md` that isn't wrapper-related: the documented payloads for `POST /tools`, `/resources`, `/prompts` and `/a2a` are flat, but those endpoints each take two body params, so FastAPI expects the embedded `{"tool": {...}}` form — the flat form returns a 422 with the generic "An error occurred, please try again."

Is that worth its own issue, or is there a reason the docs use the flat form? I've left it alone here since it's pre-existing.

docs/docs/testing/acceptance.md
Yosiefeyob: Cleared them all: the 9.1 heading, the "MCP Wrapper Testing" row and "VS Code Integration (Wrapper)" row in the report table, and the "wrapper testing" note in Test Environment Setup. No wrapper references left in the file. Also renumbered the sections, since removing heading 8 left the numbering skipping 7 → 9.

While in that section I tested the VS Code config itself and found it wrong independently of the wrapper: it said `settings.json` with `mcp.servers`, and 9.2 used `transport:`. VS Code reads `servers` from `.vscode/mcp.json` and the key is `type`. Confirmed by pointing both shapes at local listeners — `mcp.json` got the full handshake, `settings.json` got nothing. Both blocks corrected.

docs/docs/using/clients/mcp-cli.md
Yosiefeyob: Removed both. The Environment Variables section now covers the LLM provider keys only, with a line noting the gateway URL and token live in `server_config.json`. `MCP_AUTH` and `MCP_SERVER_URL` no longer appear anywhere in the file.

Also fixed the Quick Start checklist, which still said "Create `server_config.json` with correct Python path" — a leftover from the stdio format, since the example above it has no Python path any more.

lucarlig (dismissed):
LGTM

msureshkumar88 (dismissed):
Re-reviewed at current head (0a3e8f8c2) after the prior review round — verified scope against issue #5814, blast radius, dead code, and test coverage.

**Scope**: matches the issue checklist exactly — removes `mcpgateway/wrapper.py`, `crates/wrapper/`, the CI workflow, and updates docs/Admin UI. `translate.py` correctly left alone (live gRPC/baggage callers, and the issue only asked to *assess* it).

**Verified clean**: no dangling `mcpgateway.wrapper` imports or `stdio-config-btn` refs anywhere, `Cargo.lock`/workspace `Cargo.toml` have zero wrapper references, no alembic/schema changes needed (none touched). Previously-flagged issues (undisclosed test removal, stale `mcp-cli.md`/JSDoc mentions) are now fixed — the Langfuse test trade-off is explicitly documented in the PR notes.

Two non-blocking suggestions for a follow-up, not blocking this PR:
- A `CHANGELOG.md` entry noting the wrapper removal as breaking, with a pointer to the FastMCP migration path for anyone still invoking `python -m mcpgateway.wrapper` directly.
- `crates/mcp_runtime/FOLLOWUPS.md:919` still lists `mcpgateway.wrapper` as a client-compatibility expectation — stale now, internal notes file only.

Only merge conflict is in `.secrets.baseline` (line-number drift from doc edits) — trivial to rebase.

lucarlig (dismissed):
LGTM

Yosiefeyob:
Thanks for the re-review.

Added the CHANGELOG entry — `### Breaking Changes` under Unreleased, noting `python -m mcpgateway.wrapper` is gone and pointing at `uvx fastmcp-remote` plus the per-client docs.

Left the `FOLLOWUPS.md` line alone: that file sits inside `crates/mcp_runtime`, which #5822 removes, so the stale reference goes with it.

Rebasing for the `.secrets.baseline` conflict now.

msureshkumar88 (dismissed):

ja8zyjits:
## 🚨 Blocking Issues

### B1 — `AGENTS.md` still describes deleted `crates/wrapper/` crate

**File:** [`AGENTS.md`](AGENTS.md:41) · **Line:** 41

```
# current
crates/                     # Direct Rust crate folders (runtime and wrapper)

# fix
crates/                     # Direct Rust crate folders (runtime)
```

`crates/wrapper/` was deleted. Comment misleads developers and AI agents into looking for a crate that no longer exists.

---

### B2 — `DEVELOPING.md` Inspector block exports dead wrapper vars

**File:** [`DEVELOPING.md`](DEVELOPING.md:504-505) · **Lines:** 504–505

```bash
# current — dead, wrapper-only
export MCP_SERVER_URL=http://localhost:4444/servers/UUID/mcp
export MCP_AUTH="Bearer $(python3 -m mcpgateway.utils.create_jwt_token --username admin --exp 0 --secret my-test-key-but-now-longer-than-32-bytes)"

# fix
export MCPGATEWAY_BEARER_TOKEN=$(python3 -m mcpgateway.utils.create_jwt_token \
    --username admin --exp 0 --secret my-test-key-but-now-longer-than-32-bytes)

# Launch Inspector with SSE (direct)
npx @modelcontextprotocol/inspector
# Open browser to http://localhost:5173
# Add server: http://localhost:4444/servers/UUID/sse
# Add header: Authorization: Bearer $MCPGATEWAY_BEARER_TOKEN
```

`MCP_SERVER_URL` and `MCP_AUTH` consumed exclusively by the deleted wrapper. Neither is read by Inspector nor any remaining code. The correct approach (`Authorization` header) is already documented in the block that follows — these exports are dead and confusing.

Keep `MCP_GATEWAY_BASE_URL` (line 503) only if referenced elsewhere in the file; otherwise remove it too.

---

### B3 — `fixtures/stdio_echo_env.py` missing — dead fixture parameter in six tests

**File:** [`tests/unit/mcpgateway/test_translate_stdio_endpoint.py`](tests/unit/mcpgateway/test_translate_stdio_endpoint.py:51)

```python
# fixture resolves a non-existent path
return os.path.join(os.path.dirname(__file__), "fixtures", "stdio_echo_env.py")
```

Neither `tests/unit/mcpgateway/fixtures/` nor `stdio_echo_env.py` exist in the branch. Six tests accept `test_script` as a parameter (`test_start_with_initial_env_vars`, `test_start_with_additional_env_vars`, `test_environment_variable_override`, `test_multiple_env_vars`, `test_env_vars_with_special_characters`, `test_large_env_vars`) but never pass it to the subprocess — the command is hardcoded to `jq -cMn env`. Tests pass coincidentally; the fixture parameter is dead weight. Any future test that actually invokes `test_script` will silently fail with `FileNotFoundError`.

**Fix:** Remove `test_script` from the six test signatures, OR create the fixture file and wire it into the subprocess command.

---

## ⚠️ Functional Issues

### F1 — `mcp-inspector.md` "Environment variables" section exports wrapper-only vars

**File:** [`docs/docs/using/clients/mcp-inspector.md`](docs/docs/using/clients/mcp-inspector.md:33-46) · **Lines:** 33–46

```bash
# current — wrapper artifacts
export MCP_SERVER_URL=http://localhost:4444/servers/UUID_OF_SERVER_1
export MCP_AUTH=$(python3 -m mcpgateway.utils.create_jwt_token -u admin@example.com --secret ...)

# fix
export MCPGATEWAY_BEARER_TOKEN=$(python3 -m mcpgateway.utils.create_jwt_token \
    -u admin@example.com --secret my-test-key-but-now-longer-than-32-bytes)
```

`MCP_SERVER_URL` has no consumer in the current codebase. `MCP_AUTH` works as a raw value for the `--header` flag but the name is a wrapper sentinel — confusing. Replace with `MCPGATEWAY_BEARER_TOKEN` (already used in the quick-start table at line 28) and update the subsequent `--header` line.

---

### F2 — `README.md` transport list misleadingly includes `stdio` in client-connection context

**File:** [`README.md`](README.md:76) · **Line:** 76

```
# current
* Transport over HTTP, JSON-RPC, WebSocket, SSE (with configurable keepalive), stdio and streamable-HTTP

# fix
* Transport over HTTP, JSON-RPC, WebSocket, SSE (with configurable keepalive), and Streamable HTTP; stdio transport available for server-side use
```

`mcpgateway/transports/stdio_transport.py` still exists (server-side stdin/stdout, not the deleted client bridge), so `stdio` remains technically valid — but listing it alongside SSE/WebSocket/Streamable HTTP in a client-connection context implies clients can connect over stdio natively, which this PR explicitly removes. Clarify or drop from the client-facing list.

---

### F3 — `mcp.initialize` Langfuse trace test removed with no replacement

**File:** [`tests/live_gateway/mcp/test_langfuse_traces.py`](tests/live_gateway/mcp/test_langfuse_traces.py)

Two tests removed without replacement:
1. `test_langfuse_trace_export_eventually_contains_initialize_trace` — verified `langfuse.trace.name == "mcp.initialize"`, `auth:jwt` tag, `langfuse.user.id`. Most fundamental observable trace; now zero coverage.
2. `test_langfuse_trace_export_eventually_contains_fresh_mcp_cli_tool_list_trace` — `mcp-cli` dependency removal is acceptable, but `tools/list` trace coverage is not replaced by any surviving test.

Four tests correctly migrated to `_send_jsonrpc_http` (`resources/list`, `tool_call`, `prompts/get`, `sanitized_prompt_error`).

**Fix:** Add a test that sends `initialize` (id=1) via `_send_jsonrpc_http` and asserts `langfuse.trace.name == "mcp.initialize"`. Note: the internal `initialize` handshake inside `_send_jsonrpc_http` is a setup step and won't produce the asserted trace — a separate explicit `initialize` payload is needed.

---

### F4 — `_send_jsonrpc_http` creates a new MCP session per call — session leak under stateful gateway

**File:** [`tests/live_gateway/mcp/test_langfuse_traces.py`](tests/live_gateway/mcp/test_langfuse_traces.py:105-152) · **Lines:** 105–152

Every `_send_jsonrpc_http` call performs `initialize → notifications/initialized → payload`, creating a new `mcp-session-id`. Four migrated tests leave four open sessions per run with no `DELETE /mcp` teardown. Harmless when `use_stateful_sessions=false` (default in CI), but can cause intermittent failures on repeated runs against the same stateful gateway instance.

**Recommendation (non-blocking — follow-up issue):** Module-scoped fixture sharing one `mcp-session-id` with teardown `DELETE /mcp`, or add a comment noting the session-per-call pattern for future readers.

---

## 💡 Suggestion

### S1 — `test_translate_stdio_endpoint.py` has external dependency on `jq`

**File:** [`tests/unit/mcpgateway/test_translate_stdio_endpoint.py`](tests/unit/mcpgateway/test_translate_stdio_endpoint.py:108)

Six async tests spawn `jq -cMn env`. `jq` is not a Python test dependency and absent from minimal CI Python images — all six fail with `FileNotFoundError: [Errno 2] No such file or directory: 'jq'` rather than at the assertion.

**Fix:**
```python
# replace
["jq", "-cMn", "env"]

# with
[sys.executable, "-c", "import json,os; print(json.dumps(dict(os.environ)))"]
```

Pre-existing issue; flagged because this PR touches the same file.

---

## 📎 Minor Notes

### M1 — `configExport.test.js` missing `"http"` streamable-HTTP unit test

**File:** [`tests/unit/js/configExport.test.js`](tests/unit/js/configExport.test.js:459)

`test("fetches server details and generates stdio config")` deleted; replacement is the pre-existing `"generates sse config"` test. `generateAndShowConfig("http")` — the PR's canonical transport — has no JS unit test coverage.

**Suggested fix:** Add a `generateAndShowConfig("http")` test to mirror the deleted stdio test. May be covered by E2E tests; low risk.

---

### M2 — `configExport.test.js` assertion `toContain("sse-config.json")` may pass vacuously

**File:** [`tests/unit/js/configExport.test.js`](tests/unit/js/configExport.test.js:459-478) · **Lines:** 459–478

Test renamed from `"generates correct filename for stdio config"` to `"generates correct filename for sse config"`. Assertion `toContain("sse-config.json")` is correct only if `downloadConfig` derives the filename from config type. If filename is hard-coded, assertion passes vacuously. Test-accuracy concern; no runtime impact.

---

msureshkumar88:
Thanks for the thorough pass — went through each item against the current diff.

**Fixed (1fd8501a5):**
- **B1** — `AGENTS.md` no longer mentions the deleted `crates/wrapper`.
- **B2** — `DEVELOPING.md` Inspector block now uses `MCPGATEWAY_BEARER_TOKEN` instead of the dead `MCP_SERVER_URL`/`MCP_AUTH` wrapper vars.
- **F2** — `README.md` transport bullet now says stdio is server-side only, not a client transport.
- **F3** — restored `mcp.initialize` Langfuse trace coverage via `_send_jsonrpc_http` against root `/mcp/`, per your suggested approach.

**Confirmed real, left out of this PR's scope:**
- **B3** / **S1** — `test_translate_stdio_endpoint.py`'s dead `stdio_echo_env.py` fixture and the `jq` dependency are both pre-existing, on a file this PR never touches (it covers `translate.py`'s server-side `StdIOEndpoint`, not `crates/wrapper`/`mcpgateway.wrapper`). Happy to open a follow-up issue if useful.
- **F1** — `mcp-inspector.md`'s wrapper-named env vars are likewise pre-existing and outside this PR's diff.
- **F4** — agreed non-blocking per your own note; tracking as a follow-up rather than bundling in here.

**Checked, not changed:**
- **M1** — `"http"` JS coverage already exists (`configExport.test.js:105`).
- **M2** — filename is derived from `currentConfigType` dynamically, not hardcoded, so the assertion isn't vacuous.

ja8zyjits:
Hi @msureshkumar88 

---

## 🔴 Blocking Issues

### BLK-1: `configExport.js` emits `/mcp` URL without trailing slash

- **File:** [`mcpgateway/admin_ui/configExport.js`](mcpgateway/admin_ui/configExport.js:176)
- **Problem:** `case "http":` generates `${baseUrl}/servers/${server.id}/mcp`. All migrated docs use `${baseUrl}/servers/${server.id}/mcp/`. Inconsistency causes user confusion and breaks MCP clients that append subpaths to base URLs.
- **Fix:**
```javascript
url: `${baseUrl}/servers/${server.id}/mcp/`,
```

---

### BLK-2: Stale "via wrapper" descriptions in client index

- **File:** [`docs/docs/using/clients/index.md`](docs/docs/using/clients/index.md:36)
- **Problem:** Table still references deleted wrapper for Claude Desktop and MCP Inspector.
- **Fix:**
```markdown
| [Claude Desktop](claude-desktop.md) | Configure Claude to connect to ContextForge via Streamable HTTP |
| [MCP Inspector](mcp-inspector.md) | Browse and test the Gateway over SSE or Streamable HTTP |
```

---

## 🟡 Functionally Impacting Issues

### FNC-1: Langfuse test double-initializes session

- **File:** [`tests/live_gateway/mcp/test_langfuse_traces.py`](tests/live_gateway/mcp/test_langfuse_traces.py:261)
- **Problem:** `_send_jsonrpc_http` executes internal `initialize` before sending payload. Passing `initialize` payload causes duplicate handshake with duplicate request ID `1`. Trace caught by test belongs to helper handshake, not payload.
- **Fix:** Call POST directly for one-shot initialize test:
```python
triggered_after = time.time() - 1
url = f"{BASE_URL}/mcp/"
headers = {
    **_gateway_api_headers(jwt_token),
    "Accept": "application/json, text/event-stream",
    "Content-Type": "application/json",
    "mcp-protocol-version": MCP_PROTOCOL_VERSION,
}
init_response = httpx.post(
    url, headers=headers,
    json={"jsonrpc": "2.0", "id": 1, "method": "initialize",
          "params": {"protocolVersion": MCP_PROTOCOL_VERSION, "capabilities": {},
                     "clientInfo": {"name": "langfuse-e2e", "version": "1.0"}}},
    timeout=20,
)
init_response.raise_for_status()
response_json = init_response.json()
assert "error" not in response_json
```

---

### FNC-2: Stale `tools_rust/` path breaks dev commands

- **Files:** [`AGENTS.md`](AGENTS.md:71), [`DEVELOPING.md`](DEVELOPING.md:36)
- **Problem:** Quality check commands reference non-existent `tools_rust/mcp_runtime`.
- **Fix:** Update path to `crates/mcp_runtime` across 3 locations:
  - `AGENTS.md` (line 73)
  - `DEVELOPING.md` (lines 37, 230)

---

## 🔵 Suggestions & Minor

- **SUG-1:** Add note in `_send_jsonrpc_http` docstring ([`tests/live_gateway/mcp/test_langfuse_traces.py`](tests/live_gateway/mcp/test_langfuse_traces.py:106)) warning callers against passing `initialize` as payload.
- **SUG-2:** Sync workspace [`Cargo.toml`](Cargo.toml:32) `rmcp` dependency to `"1.8.0"` to match `Cargo.lock`.
- **SUG-3:** Add comment to [`tests/live_gateway/mcp/test_langfuse_traces.py`](tests/live_gateway/mcp/test_langfuse_traces.py:148) clarifying `notifications/initialized` returns 202 Accepted without response body.
- **MIN-1:** Relabel `stdio_transport.py` in [`DEVELOPING.md`](DEVELOPING.md:142) from `# Standard I/O wrapper` to `# Server-side stdio transport`.

---

msureshkumar88:
Thanks for the thorough second pass, @ja8zyjits — all of these were valid and in scope, fixed in `ca463daa`:

**Blocking**
- BLK-1: `configExport.js` now emits the `/mcp/` URL with a trailing slash.
- BLK-2: dropped the stale "via wrapper" descriptions for Claude Desktop / MCP Inspector in `docs/using/clients/index.md`.

**Functionally impacting**
- FNC-1: the `initialize` probe in `test_langfuse_traces.py` now goes straight through `httpx` instead of `_send_jsonrpc_http`, so it no longer races the helper's own internal handshake for the `mcp.initialize` trace.
- FNC-2: `tools_rust/mcp_runtime` → `crates/mcp_runtime` fixed in AGENTS.md and DEVELOPING.md.

**Suggestions/minor**
- Added a docstring warning on `_send_jsonrpc_http` against passing an `initialize` payload, a comment on the 202 `notifications/initialized` response, synced `rmcp` to `1.8.0` in `Cargo.toml`, and relabeled the `stdio_transport.py` comment in DEVELOPING.md.

Branch is rebased onto latest main and `.secrets.baseline` regenerated.

lucarlig (approved):

ja8zyjits (approved):
LGTM

PR #6205, opened 2026-08-12
fix(security): sandbox jq filter execution for tool jsonpath_filter

msureshkumar88:
## Summary

Hardens execution of the `jsonpath_filter` field on REST tools. That field is a jq
program supplied at tool-registration time and executed against the tool's response
at invocation time. This change adds:

- A static safety gate that rejects filters referencing environment, host-state,
  logging, or filesystem-module-loading jq built-ins, enforced both when a tool is
  written (create/update) and when it's invoked (covers rows written before this
  change).
- Sandboxed execution: filters now run in a forked worker pool with a cleared
  process environment and a wall-clock timeout, so a filter can no longer read the
  gateway process's environment or wedge a worker indefinitely. A broken worker is
  detected and the pool recovers automatically on the next call.
- Config: `JQ_FILTER_EXECUTION`, `JQ_FILTER_TIMEOUT_SECONDS`, `JQ_FILTER_WORKERS`.
- Docs: new "jq Filter Sandbox" section in `docs/docs/manage/securing.md`, including
  guidance to rotate `JWT_SECRET_KEY`, `AUTH_ENCRYPTION_SECRET`, `DATABASE_URL`,
  `REDIS_URL`, and `BASIC_AUTH_PASSWORD` on any deployment that ran a build prior to
  this fix.

Addresses an internally-tracked security report (PSIRT REF 3874125 / ADV0318908).

One unrelated one-line change is included: a `# nosec B608` on a bandit false
positive in an existing OAuth-token migration file, surfaced because the full lint
gate ran end-to-end for the first time during this work.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -2527,6 +2527,13 @@ VALIDATE_TOKEN_ENVIRONMENT=true
 # Default: 5000 characters, Range: 1000-100000
 # REST_RESPONSE_TEXT_MAX_LENGTH=5000
 
+# jq filter sandbox for tool jsonpath_filter execution.
+# Filters run in a forked worker with a cleared environment and a time limit.
+# Options: subprocess (default, safe), inprocess (unsafe, no environment scrub, no time limit)
+# JQ_FILTER_EXECUTION=subprocess
+# JQ_FILTER_TIMEOUT_SECONDS=2.0
+# JQ_FILTER_WORKERS=2
+
 # Prompt Configuration
 # PROMPT_CACHE_SIZE=100
 # MAX_PROMPT_SIZE=102400
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^\\.secrets\\.baseline$",
     "lines": null
   },
-  "generated_at": "2026-08-14T11:47:13Z",
+  "generated_at": "2026-08-14T12:19:58Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -3884,7 +3884,7 @@
         "hashed_secret": "c377074d6473f35a91001981355da793dc808ffd",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 4608,
+        "line_number": 4661,
         "type": "Hex High Entropy String",
         "verified_result": null
       }
--- a/docs/docs/manage/securing.md
+++ b/docs/docs/manage/securing.md
@@ -649,11 +649,11 @@ Before connecting any MCP server:
 - [ ] Monitor server behavior for anomalies
 - [ ] Implement rate limiting for untrusted servers
 
-### jq filter sandbox
+### 13. jq Filter Sandbox
 
 Tool `jsonpath_filter` values are jq programs. They are validated when a tool is
-written and again when it is invoked, and they execute in a forked worker whose
-environment has been cleared, under `JQ_FILTER_TIMEOUT_SECONDS`.
+written and again when it is invoked, and — on Linux — they execute in a forked
+worker whose environment has been cleared, under `JQ_FILTER_TIMEOUT_SECONDS`.
 
 Filters that reference `env`, `$ENV`, `input`, `inputs`, `input_filename`,
 `input_line_number`, `$__loc__`, `debug`, `stderr`, `include`, `import`, or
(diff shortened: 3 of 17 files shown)


msureshkumar88:
## Verification

Ran an end-to-end check against a live gateway instance — real DB, real RBAC, real HTTP to a local upstream, real jq execution, non-admin `developer`-role token. No mocks. No Docker (the host already had an unrelated Compose stack running, so this drives the app directly via uvicorn on a scratch SQLite DB and a free port instead).

### Config used

```
DATABASE_URL=sqlite:////<scratch>/e2e.db
JQ_FILTER_TIMEOUT_SECONDS=2.0
JQ_FILTER_WORKERS=2
JQ_FILTER_EXECUTION=(unset, default=subprocess)
ENVIRONMENT=development
AUTH_REQUIRED=true
```

### What ran

1. Bootstrapped the database (`python -m mcpgateway.bootstrap_db`).
2. Provisioned a non-admin user with the `developer` role in its own team, via the real service layer (`EmailAuthService`, `TeamManagementService`, `RoleService`) — not a shortcut.
3. Minted a real JWT for that user (`python -m mcpgateway.utils.create_jwt_token`).
4. Started the real gateway (`uvicorn mcpgateway.main:app`) and a tiny local HTTP upstream for REST tools to call.
5. Ran five checks against the running server with `curl`.

### Results

```
CHECK 1 (write-time gate): developer creates a REST tool with jsonpath_filter="$ENV"
  → 422 (detail redacted from the HTTP body by EXPOSE_ERROR_DETAILS, pre-existing and
    unrelated to this change; server log confirms "restricted built-in")
  PASS

CHECK 2 (invoke-time gate): a row seeded directly via the ORM, bypassing the write-time
validator entirely — simulating a tool created before this fix — with jsonpath_filter="$ENV"
  → {"result":{"content":[{"type":"text","text":"jsonpath filter uses a restricted jq builtin"}],"isError":true}}
  No JWT_SECRET_KEY (or any other value) leaked in the response.
  PASS

CHECK 3 (DoS containment): a non-terminating filter (reduce range(1e11) as $i (0; .+1))
  → {"result":{"content":[{"type":"text","text":"jsonpath filter exceeded the execution time limit"}],"isError":true}}
  elapsed=2.04s (matches JQ_FILTER_TIMEOUT_SECONDS); /health returns 200 immediately after —
  the worker pool recovered, the gateway did not freeze.
  PASS

CHECK 4 (no regression): a legitimate filter (.data.message) against a real upstream response
  → {"result":{"content":[{"type":"text","text":"[\n  \"hello\"\n]"}],"isError":false}}
  PASS

CHECK 5 (latency): 20 sequential calls through a benign filter
  → ~37ms average end-to-end (HTTP + RBAC + upstream call + sandbox). Unit-level sandbox-only
    overhead measured separately during review was ~0.4ms/call — different measurement, both
    legitimate.

SUMMARY: 5 passed, 0 failed
```

Collateral check: `git status` clean throughout, no leftover processes after teardown, the pre-existing unrelated Docker stack (11 containers) was untouched.

### CI note

Every check is green except **`pytest (py3.12)`**, which fails only in its post-test coverage-combine step, not on test results — every run shows **0 failed tests** (most recently 21939 passed, 819 skipped, 2 xfailed). Root cause, confirmed by bisection: this repo's `pyproject.toml` coverage config lacks `branch = true`, while CI's pytest invocation passes `--cov-branch` on the CLI. A real `os.fork()` during that combination (this PR's new `ProcessPoolExecutor`-based worker pool is the first place in the codebase that forks a real OS process under test) writes a coverage data fragment in the wrong mode, and `coverage combine()` raises `DataError: Can't combine statement coverage data with branch data` after all tests have already passed. Reproduced deterministically in isolation (5/5 local runs), confirmed unrelated to xdist specifically (reproduces with `-n 0`), and confirmed this branch's own code isn't otherwise responsible (a minimal test file that only imports the new module doesn't trigger it).

A scoped fix exists (a dedicated `--cov-config` for just that CI step, leaving the shared `pyproject.toml` and the separate doctest-coverage gate untouched) but touches `.github/workflows/pytest.yml`, which is out of scope for this branch — left as-is per explicit direction, so a maintainer with full visibility into every coverage-dependent CI step can make that call. Coverage itself was independently verified locally via the project's own gate (`make coverage` then `diff-cover coverage.xml --compare-branch=origin/main --fail-under=95`): 98% total, 99% coverage on changed lines.

gandhipratik203:
`mcpgateway/utils/jq_runner.py:352-360`

`future.result(timeout=...)` counts queue wait, not execution time. With 2 workers / 2.0s default, a few concurrent filtered calls make a fast queued filter "time out", which then kills every worker in the pool, including healthy in-flight ones, forcing a re-fork. Self-reinforcing under load.

msureshkumar88:
Good catch, thanks — verified and fixed in 4c32ab652.

Confirmed both parts empirically before fixing:
- `future.result(timeout=...)` does count from `pool.submit()`, not from when a worker actually picks the task up — reproduced with a 1-worker pool where a queued call timed out despite its own execution being effectively instant.
- The timeout handler killed every worker in the pool (`_kill_workers` iterates all of `pool._processes`), so a queued-then-timed-out call took down workers that were mid-execution on other, unrelated, perfectly healthy filters.

I first tried distinguishing "queued, never started" from "actually running" via `future.cancel()`'s return value, since a `ProcessPoolExecutor` future that hasn't started should still be cancellable. Verified empirically that this doesn't work: `submit()` moves the work item past the cancellable window almost immediately, so `.cancel()` returns `False` even for a task that never began executing. Dropped that approach once the test proved it out.

Went with admission control instead: a `threading.Semaphore` sized to the pool's own worker count, stamped onto each pool at build time (so it never goes stale relative to that pool's actual worker count, even across a kill+rebuild or a `jq_filter_workers` change). `run_jq_filter` now acquires a permit *before* submitting — non-blocking. If every worker is busy, it fails immediately with a new `JqFilterBusy` (a `JqFilterError`, distinct from `JqFilterTimeout`) rather than queueing. Since a submission only ever reaches the pool once a permit confirms a worker is free, a subsequent timeout is now always a genuine runaway — never queueing pressure — so the pool-wide kill only fires for real hangs.

Added tests reproducing the exact scenario (one worker occupied by a runaway filter, a second concurrent call must fail fast — under 1s, nowhere near the 5s budget used in the test — and must not touch the occupied worker), plus coverage for gate release on normal completion and the new exception's shape. `mcpgateway/utils/jq_runner.py` and `mcpgateway/utils/jq_guard.py` stay at 100% coverage; full jq/tool_service suites pass (864 tests); ruff/pylint/interrogate clean.

msureshkumar88:
Re-triggering CI: GitHub Actions never registered a run for commit 4c32ab652 (confirmed via the Actions API — zero workflow runs exist for that SHA, only the DCO app check fired). Closing and reopening to force re-evaluation.

gandhipratik203 (dismissed):
LGTM! Thanks for the changes!

Lang-Akshay (dismissed):
LGTM ✅

Lang-Akshay (dismissed):
LGTM

Lang-Akshay (approved):
LGTM

PR #6208, opened 2026-08-12
fix(ci): scope branch coverage to the pytest step to fix combine crash

### LINKED ISSUE #6207: [BUG]: pytest CI job crashes on coverage combine when tests fork a real process (branch vs statement data mismatch)
### 🐞 Bug Summary

The `pytest (py3.12)` job in the `Tests & Coverage` workflow crashes with `INTERNALERROR` in its coverage-combine step whenever a real `os.fork()` happens during the test run (e.g. via `multiprocessing`/`concurrent.futures.ProcessPoolExecutor` with the `fork` start method) while that process also executes code covered by `--cov=mcpgateway`. All actual tests pass — the crash happens *after* the test run completes, while `pytest-cov` merges per-process coverage data files.

This is not a flaky/random failure: it reproduces deterministically whenever a PR's tests fork a real process under coverage, and does not reproduce for PRs that don't.

---

### 🧩 Affected Component

- [x] CLI, Makefiles, or shell scripts
- [ ] `mcpgateway` - API
- [x] Other: `.github/workflows/pytest.yml`, coverage.py + pytest-xdist interaction

---

### 🔁 Steps to Reproduce

1. Add any test that forks a real process via `concurrent.futures.ProcessPoolExecutor(mp_context=multiprocessing.get_context("fork"))` (or plain `multiprocessing` with the `fork` start method) and has the forked child execute any code under `mcpgateway/` (coverage's `source` package).
2. Run `pytest --cov=mcpgateway --cov-branch <that test>` (matches `.github/workflows/pytest.yml`'s "Run pytest" step). `-n auto`/xdist is **not** required to reproduce — `-n 0` reproduces identically.
3. All tests pass. At session teardown, `pytest-cov` calls `coverage.combine()` and it raises `INTERNALERROR`.

Minimal local repro (not part of any PR, just for reproducing this issue):

```python
import multiprocessing
from concurrent.futures import ProcessPoolExecutor

def _child(x):
    from mcpgateway.utils.create_slug import slugify  # any instrumented mcpgateway code
    return slugify(f"item-{x}")

def test_forks_and_executes_instrumented_code():
    ctx = multiprocessing.get_context("fork")
    with ProcessPoolExecutor(max_workers=1, mp_context=ctx) as pool:
        assert pool.submit(_child, 41).result() == "item-41"
```

```
pytest test_repro.py --cov=mcpgateway --cov-branch --cov-report=
```

**Root cause:** `pyproject.toml`'s `[tool.coverage.run]` sets `concurrency = ["thread", "multiprocessing"]`, which makes `coverage.py` register an `os.register_at_fork()` hook so each forked process gets its own coverage data file (needed for `parallel = true` to combine correctly). That hook's session reads its `branch` setting from whatever on-disk coverage config it finds — `pyproject.toml`'s `[tool.coverage.run]` has no `branch` key — not from the CLI `--cov-branch` flag the parent process was invoked with. So the forked child writes **statement-only** coverage data while the CLI-driven parent process writes **branch** data. `coverage.data.combine_parallel_data()` then raises:

```
coverage.exceptions.DataError: Can't combine statement coverage data with branch data
```

---

### 🤔 Expected Behavior

The `pytest (py3.12)` job should complete based on test results only. A PR whose tests legitimately fork a real OS process (e.g. to sandbox untrusted code, as in #6205) shouldn't fail CI purely because of a coverage-tooling data-format mismatch, when every actual test passes.

---

### 📓 Logs / Error Output

```
INTERNALERROR>   File ".../pytest_cov/engine.py", line 366, in finish
INTERNALERROR>     self.cov.combine()
INTERNALERROR>   File ".../coverage/control.py", line 904, in combine
INTERNALERROR>     combine_parallel_data(
INTERNALERROR>   File ".../coverage/data.py", line 216, in combine_parallel_data
INTERNALERROR>     data.update(new_data, map_path=map_path)
INTERNALERROR>   File ".../coverage/sqldata.py", line 730, in update
INTERNALERROR>     raise DataError(
INTERNALERROR> coverage.exceptions.DataError: Can't combine statement coverage data with branch data

21941 passed, 819 skipped, 2 xfailed, 528 warnings in 880.45s (0:14:40)
##[error]Process completed with exit code 3.
```

Observed on PR #6205 (`fix/tool-jsonpath-filter-jq-sandbox`), which introduces the first `ProcessPoolExecutor`-based real forking in the test suite (a sandboxed jq filter execution pool). All checks on that PR are green except this one; every run shows 0 failed tests.

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Version or commit | `main` (also reproduces on any branch off it) |
| Runtime | Python 3.12, `pytest-cov` 7.1.0, `coverage.py` (bundled via `pytest-cov`) |
| Platform / OS | GitHub Actions `ubuntu-latest` runner |
| Container | none (native runner) |

---

### 🧩 Additional Context

Considered and rejected: adding `branch = true` directly to `pyproject.toml`'s `[tool.coverage.run]`. That does fix this crash, but the same workflow's separate "Doctest coverage with threshold" step (`--doctest-modules mcpgateway/ --cov=mcpgateway --cov-fail-under=30`) reads the same file and has no `--cov-branch` of its own — its 30% threshold is calibrated against statement-mode measurement, and switching to branch mode dropped its measured coverage from ~30% to 27.75%, failing a previously-passing gate for an unrelated reason.

A PR is attached that scopes `branch = true` to only the main "Run pytest" step via a dedicated `--cov-config` file, leaving `pyproject.toml` and the doctest-coverage step untouched.

msureshkumar88:
## Summary

Fixes #6207.

The `pytest (py3.12)` job crashes in its coverage-combine step whenever a test forks a real OS process (via `ProcessPoolExecutor`/`multiprocessing` with the `fork` start method) that executes any code under `mcpgateway/`. Every actual test still passes — the crash happens after the run, while `pytest-cov` merges per-process coverage data files, and it fails the whole job (exit code 3) regardless of test results.

**Root cause:** `pyproject.toml`'s `[tool.coverage.run]` sets `concurrency = ["thread", "multiprocessing"]`, which makes `coverage.py` register an `os.register_at_fork()` hook so each forked process writes its own coverage data file. That hook's session reads `branch` mode from on-disk config, not from the parent's CLI `--cov-branch` flag — and `pyproject.toml` has no `branch` key. So a forked child writes statement-only data while the CLI-driven parent writes branch data, and `coverage combine()` raises `DataError: Can't combine statement coverage data with branch data`.

**Fix:** a dedicated `.github/pytest-branch-coverage.coveragerc` with `branch = true`, passed to only the "Run pytest" step via `--cov-config`. Deliberately not a change to `pyproject.toml` — the workflow's separate "Doctest coverage with threshold" step shares that file and has its own 30% gate calibrated against statement-mode measurement; making `branch = true` the global default drops that measurement to ~27.75%, failing a previously-passing, unrelated gate. Scoping the change to `--cov-config` on one step avoids that.

## Verification

Reproduced and fixed with a standalone repro (not part of the test suite, just used to validate this fix):

```python
import multiprocessing
from concurrent.futures import ProcessPoolExecutor

def _child(x):
    from mcpgateway.utils.create_slug import slugify
    return slugify(f"item-{x}")

def test_forks_and_executes_instrumented_code():
    ctx = multiprocessing.get_context("fork")
    with ProcessPoolExecutor(max_workers=1, mp_context=ctx) as pool:
        assert pool.submit(_child, 41).result() == "item-41"
```

- Without the fix: `pytest --cov=mcpgateway --cov-branch <repro>` → `INTERNALERROR`, deterministic across repeated runs, reproduces with both `-n 0` and `-n 2` (not xdist-specific).
- With `--cov-config=.github/pytest-branch-coverage.coveragerc` added: 3/3 clean runs, no crash.
- Confirmed the doctest-coverage step's invocation doesn't pass `--cov-config` and is untouched by this change.

### DIFF
--- a/.github/pytest-branch-coverage.coveragerc
+++ b/.github/pytest-branch-coverage.coveragerc
@@ -1,0 +1,40 @@
+; Coverage config for the "Run pytest" step in .github/workflows/pytest.yml only.
+;
+; That step passes --cov-branch on the CLI, but coverage.py's own fork-safety
+; hook (registered because pyproject.toml sets concurrency = ["thread",
+; "multiprocessing"]) gives each process forked during the run its own
+; coverage session, and that session reads its mode from whatever on-disk
+; config it finds -- not from the parent's CLI flags. pyproject.toml's
+; [tool.coverage.run] has no `branch` setting, so a forked child ends up
+; writing statement-only data while the CLI-driven parent writes branch data,
+; and `coverage combine()` then raises:
+;   DataError: Can't combine statement coverage data with branch data
+;
+; This file is deliberately NOT a change to pyproject.toml: the doctest
+; coverage step in the same workflow shares that file and has its own
+; threshold calibrated against statement-mode measurement, so making
+; branch=true the global default would push it below its gate. Passing
+; --cov-config=<this file> to only the "Run pytest" step keeps the fix scoped
+; to the one step that forks real processes under coverage instrumentation.
+[run]
+source = mcpgateway
+parallel = true
+branch = true
+concurrency = thread, multiprocessing
+omit =
+    mcpgateway/alembic/*
+    mcpgateway/tools/builder/*
+    */tests/*
+    */test_*.py
+
+[report]
+exclude_lines =
+    pragma: no cover
+    def __repr__
+    if self.debug
+    if TYPE_CHECKING:
+    raise AssertionError
+    raise NotImplementedError
+    if 0:
+    if __name__ == .__main__.
+ignore_errors = true
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -123,6 +123,7 @@ jobs:
             --durations=10 \
             --ignore=tests/fuzz \
             --cov=mcpgateway \
+            --cov-config=.github/pytest-branch-coverage.coveragerc \
             --cov-report=xml \
             --cov-report=html \
             --cov-report=term \


brian-hussey (approved):
Nice catch and nice fix @msureshkumar88. LGTM

cafalchio (approved):
Passing coverage LGTM

PR #6209, opened 2026-08-12
feat: add frontend-compatible email links

### LINKED ISSUE #5813: [FEATURE][BACKEND]: Add email delivery and frontend-compatible auth links
Part of #5534

### 🧭 Type of Feature

### 🧭 Epic

**Title:** Complete email delivery and frontend-compatible authentication links
**Goal:** Enable API clients, including the new React UI in a separate repository, to create team invitations through the versioned ContextForge API, receive backend-delivered invitation emails, and use frontend-compatible invitation and password-recovery links.
**Why now:** The React UI must not depend on legacy `/admin` routes or handle SMTP credentials. Email delivery belongs in backend workflows, while all browser-facing links must use one trusted frontend URL policy.

---

### 🧑🏻‍💻 User Story 1

**As a:** team owner using an API client
**I want:** ContextForge to send an invitation email after I create a team invitation
**So that:** clients such as the new React UI can provide invitation delivery without implementing email transport.

#### ✅ Acceptance Criteria
```gherkin
Scenario: Invitation email is sent through the versioned API
  Given SMTP is enabled and configured
  And I am authenticated with teams.manage_members permission
  And I own the target team
  When I POST /v1/teams/{team_id}/invitations
  Then the invitation is committed
  And ContextForge sends an invitation email through the configured SMTP server
  And the API returns HTTP 201
  And the response contains invitation_url
  And the response reports email_delivery_status as "sent"

Scenario: SMTP delivery fails after invitation creation
  Given the invitation is committed successfully
  And SMTP delivery fails
  When the invitation is created
  Then the invitation remains persisted
  And the API still returns HTTP 201
  And the response reports email_delivery_status as "failed"
  And the response contains a safe delivery warning
  And a WARNING log identifies the invitation by ID
  And the invitation token and SMTP credentials are not logged

Scenario: SMTP is disabled
  Given SMTP_ENABLED is false
  When I create an invitation
  Then no SMTP connection is attempted
  And the API returns HTTP 201
  And the response reports email_delivery_status as "disabled"
```

---

### 🧑🏻‍💻 User Story 2

**As an:** invited user
**I want:** the invitation email to contain a frontend-compatible acceptance link
**So that:** I can open the invitation flow without depending on the legacy Admin UI.

#### ✅ Acceptance Criteria
```gherkin
Scenario: Email contains a frontend-compatible invitation link
  Given an invitation email is sent
  Then the email identifies the team, inviter, role, and expiry timestamp
  And it contains one primary acceptance link
  And it includes the raw invitation token as a fallback
  And the acceptance path is /accept-invitation/{token}
  And the link does not use a legacy /admin path
  And no secret other than the invitation token appears in the URL

Scenario: Backend remains UI-independent
  When the invitation workflow runs
  Then no /admin route is called
  And the invitation service does not depend on admin.py
  And email delivery is available through POST /v1/teams/{team_id}/invitations
```

---

### 🧑🏻‍💻 User Story 3

**As a:** user of the separate React client
**I want:** password-reset and account-lockout emails to open React routes
**So that:** authentication recovery does not depend on legacy Admin UI pages.

#### ✅ Acceptance Criteria
```gherkin
Scenario: Forgot-password email contains a frontend-compatible reset link
  Given password reset is enabled
  And a registered active user requests a reset through POST /v1/auth/email/forgot-password
  When ContextForge sends the password-reset email
  Then the primary link path is /reset-password/{token}
  And the link does not use /admin/reset-password
  And the token is URL-encoded as one path segment
  And the generic forgot-password API response remains unchanged to prevent account enumeration

Scenario: Account-lockout email contains a frontend-compatible recovery link
  Given account-lockout notification email is enabled
  When ContextForge sends an account-lockout email
  Then the recovery link path is /forgot-password
  And the link does not use /admin/forgot-password

Scenario: Shared trusted frontend URL construction
  Given UI_BASE_URL is configured
  When any invitation, password-reset, or account-lockout browser link is built
  Then the configured origin and existing path prefix are preserved
  And APP_ROOT_PATH is not appended a second time
  And the inbound Host header is never used

Scenario: Shared frontend URL fallback
  Given UI_BASE_URL is not configured
  When a browser-facing email link is built
  Then the base is APP_DOMAIN plus APP_ROOT_PATH
```

---

```mermaid
flowchart LR
    A[API client] -->|POST /v1/teams/team_id/invitations| B[Team invitation service]
    B --> C[(Commit invitation)]
    C --> D[Email notification service]
    D --> E[Configured SMTP server]
    B --> F[HTTP 201 with URL and delivery status]
```

#### Backend changes

- Add `send_team_invitation_email(to_email, team_name, inviter_name, role, invitation_url, expires_at, token)` to `AuthEmailNotificationService`.
- Add a branded invitation email template with HTML and plaintext content.
- Invoke delivery only after the invitation transaction commits.
- Preserve `commit=False` transaction semantics. Do not send an email before the caller's outer transaction commits; seeded invitations must defer delivery to a post-commit caller path.
- Extend the invitation creation response with:

```python
invitation_url: str
email_delivery_status: Literal["sent", "failed", "disabled"]
warning: str | None = None
```

- Add one shared helper for browser-facing frontend URLs; it must be used by invitation, password-reset, and account-lockout email flows.
- Add optional `UI_BASE_URL` configuration for the separately deployed React client.
- When `UI_BASE_URL` is set, preserve its origin and path prefix. When unset, fall back to `APP_DOMAIN + APP_ROOT_PATH`.
- Build invitation acceptance at `/accept-invitation/{token}`, password reset at `/reset-password/{token}`, and lockout recovery at `/forgot-password`.
- URL-encode tokens as individual path segments.
- Never derive browser-facing URLs from the inbound `Host` header.
- The shared helper builds frontend browser URLs only; it must not construct versioned backend API URLs.
- Keep SMTP credentials exclusively in the backend.
- Preserve the deprecated unversioned API alias during its compatibility period, while documenting `/v1/teams/{team_id}/invitations` as the supported endpoint.

#### Security requirements

- Never log invitation tokens, password-reset tokens, complete token-bearing URLs, or SMTP credentials.
- Keep Jinja autoescaping enabled for team, inviter, role, email, and other user-controlled template values.
- Preserve existing RBAC, team ownership, expiry, and invitee validation.
- SMTP failure must never roll back a committed invitation.
- Email delivery status must not disclose sensitive SMTP error details to API clients.

#### Backend verification

- Email method receives correct team, inviter, role, expiry, URL, and token values.
- Successful delivery returns `sent`.
- SMTP failure returns `failed` while preserving the invitation.
- Disabled SMTP returns `disabled` without attempting a connection.
- Invitation, password-reset, and account-lockout URLs use the shared trusted frontend origin and do not contain `/admin`.
- `UI_BASE_URL` path prefixes and fallback `APP_DOMAIN + APP_ROOT_PATH` behavior are covered.
- Password-reset tokens are URL-encoded as single path segments.
- Forgot-password responses remain generic and do not reveal whether an account exists.
- Logs do not contain the token or SMTP credentials.
- `commit=False` does not send before the outer transaction commits.
- Versioned `/v1/teams/{team_id}/invitations` behavior is covered.

#### Out of scope

- React invitation form or UI components.
- React `/accept-invitation/:token` page.
- Legacy Admin UI changes.
- Resend-invitation endpoint.
- Invitee inbox or dashboard.
- Changes to the existing email-verification policy.
- React invitation, forgot-password, and reset-password pages or components.
- Changes to password-reset token persistence, validation, expiry, rate limits, or password policy.

---

### 🔗 MCP Standards Check

- [x] This change is outside the MCP protocol surface and uses the existing ContextForge team API

No MCP protocol deviations are expected.

---

### 🔄 Alternatives Considered

- **Send email from the React client:** rejected because SMTP credentials and delivery logic must remain server-side.
- **Use legacy `/admin` routes:** rejected because the Admin UI path is being replaced and would couple the new UI to deprecated server-rendered pages.
- **Build URLs independently in each email workflow:** rejected because configuration and path-prefix behavior would drift between invitation and password-recovery emails.
- **Return only the token or URL:** does not provide automatic delivery and preserves the current manual forwarding workflow.

---

### 📓 Additional Context

- Parent feature: #5534
- Existing SMTP transport: `mcpgateway/services/email_notification_service.py`
- Existing password-recovery workflow and legacy URL builders: `mcpgateway/services/email_auth_service.py`
- Versioned password APIs: `POST /v1/auth/email/forgot-password` and `/v1/auth/email/reset-password/{token}`
- Invitation creation: `mcpgateway/services/team_invitation_service.py`
- Versioned team router registration: `mcpgateway/api/v1/__init__.py`
- Invitation API: `mcpgateway/routers/teams.py`
- Response schema: `mcpgateway/schemas.py`
- React implementation will be tracked in a separate task.

vishu-bh:
# Pull Request

## 🔗 Related Issue

Closes #5813

---

## 📝 Summary

Adds backend-owned invitation email delivery and shared frontend-compatible links for team invitations, password resets, and account-lockout recovery.

A trusted optional `UI_BASE_URL` supports a separately deployed React client. When unset, links fall back to `APP_DOMAIN + APP_ROOT_PATH`. Invitation delivery occurs only after transaction commit, reports safe delivery status to API clients, and uses bounded concurrency for seeded invitations.

---

## 📏 Reviewability

---

## 🏷️ Type of Change

- [ ] Bug fix
- [x] Feature / Enhancement
- [x] Documentation
- [x] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

| Check | Command | Status |
|---|---|---|
| Focused unit tests | `pytest -q tests/unit/mcpgateway/services/test_email_notification_service.py tests/unit/mcpgateway/services/test_team_invitation_service.py tests/unit/mcpgateway/services/test_team_member_seeding.py tests/unit/mcpgateway/routers/test_teams.py tests/unit/mcpgateway/services/test_email_auth_basic.py tests/unit/mcpgateway/test_config.py -x` | Passed: 530 tests; 12 existing skips |
| Python lint | `make ruff` | Passed |
| Docstring coverage | `make interrogate` | Passed: 100% |
| Import and router assembly | `python -c 'from mcpgateway.services.team_invitation_service import TeamInvitationService; from mcpgateway.services.team_management_service import TeamManagementService; from mcpgateway.main import app; print(len(app.routes))'` | Passed |
| Diff whitespace | `git diff --check` | Passed |
| Security scan | `make bandit` | Blocked by existing unrelated B608 finding in migration `7ab59991e017_fix_oauth_tokens_unique_constraint.py` |
| Documentation build | `make -C docs build` | Not run: repository lacks `docs/.venv` |

---

## ✅ Checklist

- [x] Code formatted
- [x] Documentation updated

---

## 📓 Notes (optional)

`UI_BASE_URL` rejects embedded credentials, query strings, and fragments. SMTP errors expose only exception class in logs; invitation tokens, token-bearing URLs, SMTP credentials, and raw transport errors are not logged.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -306,6 +306,9 @@ HOST=0.0.0.0
 
 # Local origin used for CORS/cookies in development examples
 APP_DOMAIN=http://localhost:8080
+# Trusted React/UI origin and optional path prefix for links in invitation and password emails.
+# When unset, browser-facing links use APP_DOMAIN plus APP_ROOT_PATH.
+# UI_BASE_URL=http://localhost:3000
 
 # Enable Admin UI and Admin API for local development
 MCPGATEWAY_UI_ENABLED=true
--- a/.env.example
+++ b/.env.example
@@ -307,7 +307,7 @@ HOST=0.0.0.0
 # Local origin used for CORS/cookies in development examples
 APP_DOMAIN=http://localhost:8080
 # Trusted React/UI origin and optional path prefix for links in invitation and password emails.
-# When unset, browser-facing links use APP_DOMAIN plus APP_ROOT_PATH.
+# When unset, password recovery uses legacy /admin routes; invitations use APP_DOMAIN plus APP_ROOT_PATH.
 # UI_BASE_URL=http://localhost:3000
 
 # Enable Admin UI and Admin API for local development
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^\\.secrets\\.baseline$",
     "lines": null
   },
-  "generated_at": "2026-08-14T16:05:27Z",
+  "generated_at": "2026-08-17T09:33:27Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -116,7 +116,7 @@
         "hashed_secret": "08cd923367890009657eab812753379bdb321eeb",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 769,
+        "line_number": 772,
         "type": "Basic Auth Credentials",
         "verified_result": null
       },
@@ -124,7 +124,7 @@
         "hashed_secret": "14f8aa3e560a47851908ab0f04ec856dbc512d93",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1014,
+        "line_number": 1017,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -132,7 +132,7 @@
         "hashed_secret": "fa9beb99e4029ad5a6615399e7bbae21356086b3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1313,
+        "line_number": 1316,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -140,7 +140,7 @@
         "hashed_secret": "d08f88df745fa7950b104e4a707a31cfce7b5841",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1348,
+        "line_number": 1351,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -148,7 +148,7 @@
         "hashed_secret": "ac371b6dcce28a86c90d12bc57d946a800eebf17",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1437,
+        "line_number": 1440,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -156,7 +156,7 @@
         "hashed_secret": "0b6ec68df700dec4dcd64babd0eda1edccddace1",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1442,
+        "line_number": 1445,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -164,7 +164,7 @@
         "hashed_secret": "4ad6f0082ee224001beb3ca5c3e81c8ceea5ed86",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1447,
+        "line_number": 1450,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -172,7 +172,7 @@
         "hashed_secret": "cb32747fcfb55eaa194c8cd8e4ba7d49ada08a94",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1453,
+        "line_number": 1456,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -180,7 +180,7 @@
         "hashed_secret": "6c178d51b13520496dbc767ed3d9d7aa5803ac72",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1465,
+        "line_number": 1468,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -188,7 +188,7 @@
         "hashed_secret": "ca45060a53fd8a255d1a83ee8d2f025283ccc66e",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1483,
+        "line_number": 1486,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -196,7 +196,7 @@
         "hashed_secret": "910fbf00f58e9bcb095ea26a75cc1d9a3355e671",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1544,
+        "line_number": 1547,
         "type": "Secret Keyword",
         "verified_result": null
       }
@@ -1914,7 +1914,7 @@
         "hashed_secret": "4a0a2df96d4c9a13a282268cab33ac4b8cbb2c72",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 151,
+        "line_number": 177,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -1922,7 +1922,7 @@
         "hashed_secret": "bc2f74c22f98f7b6ffbc2f67453dbfa99bce9a32",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 456,
+        "line_number": 482,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -1930,7 +1930,7 @@
         "hashed_secret": "a2166e0b243f4e192e14000a6aa8f46cde6bfc20",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 900,
+        "line_number": 926,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -1938,7 +1938,7 @@
         "hashed_secret": "fa9beb99e4029ad5a6615399e7bbae21356086b3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1407,
+        "line_number": 1433,
         "type": "Basic Auth Credentials",
         "verified_result": null
       },
@@ -1946,7 +1946,7 @@
         "hashed_secret": "fa9beb99e4029ad5a6615399e7bbae21356086b3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1411,
+        "line_number": 1437,
         "type": "Secret Keyword",
         "verified_result": null
       }
@@ -4812,7 +4812,7 @@
         "hashed_secret": "a24a9eab46fcd2eb51c24d7e647a1e398c2c4b26",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 2191,
+        "line_number": 2252,
         "type": "Secret Keyword",
         "verified_result": null
       }
--- a/docs/docs/manage/configuration.md
+++ b/docs/docs/manage/configuration.md
@@ -121,11 +121,12 @@ ContextForge uses this trusted base to generate these browser-facing email links
 - `https://ui.example.com/contextforge/reset-password/{token}`
 - `https://ui.example.com/contextforge/forgot-password`
 
-When `UI_BASE_URL` is unset, links use `APP_DOMAIN + APP_ROOT_PATH`. That fallback host must serve the
-`/accept-invitation/{token}`, `/reset-password/{token}`, and `/forgot-password` frontend routes; ContextForge does
-not provide those React pages. If the React client is deployed separately, configure `UI_BASE_URL` or email links
-will point at the gateway and may return 404. ContextForge never derives these links from the inbound `Host` header.
-Tokens are URL-encoded as individual path segments.
+When `UI_BASE_URL` is unset, links use `APP_DOMAIN + APP_ROOT_PATH` as their base. Password-reset and account-lockout
+emails preserve compatibility with the bundled Admin UI by using `/admin/reset-password/{token}` and
+`/admin/forgot-password`; these routes require `MCPGATEWAY_ADMIN_API_ENABLED=true`. Invitation emails continue to use
+`/accept-invitation/{token}`, so the fallback host must serve that frontend route. ContextForge does not provide the
+React invitation page. If the React client is deployed separately, configure `UI_BASE_URL`. ContextForge never
+derives these links from the inbound `Host` header. Tokens are URL-encoded as individual path segments.
 
 `UI_BASE_URL` controls links only; it does not configure browser access to gateway APIs. For a React client on a
 different origin, add that exact origin to `ALLOWED_ORIGINS`. Deployments using cross-origin cookies must also set
--- a/docs/docs/manage/configuration.md
+++ b/docs/docs/manage/configuration.md
@@ -98,6 +98,7 @@ ContextForge supports multiple database backends with full feature parity across
 | `CLIENT_MODE`      | Client-only mode for gateway-as-client   | `false`                | bool                   |
 | `DATABASE_URL`     | SQLAlchemy connection URL                | `sqlite:///./mcp.db`   | any SQLAlchemy dialect |
 | `APP_ROOT_PATH`    | Subpath prefix for app (e.g. `/gateway`) | (empty)                | string                 |
+| `UI_BASE_URL`      | Trusted base URL for browser-facing links in invitation and password emails | (unset) | HTTP/HTTPS URL |
 | `TEMPLATES_DIR`    | Path to Jinja2 templates                 | `mcpgateway/templates` | path                   |
 | `STATIC_DIR`       | Path to static files                     | `mcpgateway/static`    | path                   |
 | `PROTOCOL_VERSION` | MCP protocol version supported           | `2025-06-18`           | string                 |
@@ -105,6 +106,27 @@ ContextForge supports multiple database backends with full feature parity across
 
 !!! tip "Subpath Deployment"
     Use `APP_ROOT_PATH=/foo` if reverse-proxying under a subpath like `https://host.com/foo/`.
+
+### Browser-facing email links
+
+Set `UI_BASE_URL` when the React client is deployed at a different origin or path from the gateway:
+
+```bash
+UI_BASE_URL=https://ui.example.com/contextforge
+```
+
+ContextForge uses this trusted base to generate these browser-facing email links:
+
+- `https://ui.example.com/contextforge/accept-invitation/{token}`
+- `https://ui.example.com/contextforge/reset-password/{token}`
+- `https://ui.example.com/contextforge/forgot-password`
+
+When `UI_BASE_URL` is unset, links use `APP_DOMAIN + APP_ROOT_PATH`. ContextForge never derives these links from
+the inbound `Host` header. Tokens are URL-encoded as individual path segments.
+
+`UI_BASE_URL` controls links only; it does not configure browser access to gateway APIs. For a React client on a
+different origin, add that exact origin to `ALLOWED_ORIGINS`. Deployments using cross-origin cookies must also set
+appropriate secure-cookie, SameSite, credential, and CSRF configuration.
 
 ### Authentication
 
--- a/docs/docs/manage/configuration.md
+++ b/docs/docs/manage/configuration.md
@@ -121,8 +121,11 @@ ContextForge uses this trusted base to generate these browser-facing email links
 - `https://ui.example.com/contextforge/reset-password/{token}`
 - `https://ui.example.com/contextforge/forgot-password`
 
-When `UI_BASE_URL` is unset, links use `APP_DOMAIN + APP_ROOT_PATH`. ContextForge never derives these links from
-the inbound `Host` header. Tokens are URL-encoded as individual path segments.
+When `UI_BASE_URL` is unset, links use `APP_DOMAIN + APP_ROOT_PATH`. That fallback host must serve the
+`/accept-invitation/{token}`, `/reset-password/{token}`, and `/forgot-password` frontend routes; ContextForge does
+not provide those React pages. If the React client is deployed separately, configure `UI_BASE_URL` or email links
+will point at the gateway and may return 404. ContextForge never derives these links from the inbound `Host` header.
+Tokens are URL-encoded as individual path segments.
 
 `UI_BASE_URL` controls links only; it does not configure browser access to gateway APIs. For a React client on a
 different origin, add that exact origin to `ALLOWED_ORIGINS`. Deployments using cross-origin cookies must also set
--- a/mcpgateway/config.py
+++ b/mcpgateway/config.py
@@ -1631,6 +1631,13 @@ class Settings(BaseSettings):
             if self.debug and not self.dev_mode:
                 logger.warning("🐛 SECURITY WARNING: Debug mode is enabled in non-dev mode. This may leak sensitive information! Set DEBUG=false for production.")
 
+            if self.smtp_enabled and self.ui_base_url is None:
+                logger.warning(
+                    "SMTP_ENABLED=true while UI_BASE_URL is unset. Browser-facing email links will use "
+                    "APP_DOMAIN plus APP_ROOT_PATH. Ensure that fallback serves /accept-invitation/{token}, "
+                    "/reset-password/{token}, and /forgot-password, or configure UI_BASE_URL for the React client."
+                )
+
         # CSRF secret key fallback to JWT secret key.
         # NOTE: SecretStr("") is truthy, so the emptiness check must go through
         # get_secret_value(); `if not self.csrf_secret_key` would never fire and
--- a/mcpgateway/config.py
+++ b/mcpgateway/config.py
@@ -1191,6 +1191,21 @@ class Settings(BaseSettings):
         description="Trusted base URL for browser-facing UI links. Falls back to APP_DOMAIN plus APP_ROOT_PATH when unset.",
     )
 
+    @field_validator("ui_base_url", mode="before")
+    @classmethod
+    def normalize_ui_base_url(cls, value: object) -> object:
+        """Treat an empty environment override as unset.
+
+        Args:
+            value: Raw configured frontend base URL.
+
+        Returns:
+            object: ``None`` for blank strings, otherwise the original value.
+        """
+        if isinstance(value, str) and not value.strip():
+            return None
+        return value
+
     @field_validator("ui_base_url")
     @classmethod
     def validate_ui_base_url(cls, value: Optional[HttpUrl]) -> Optional[HttpUrl]:
--- a/mcpgateway/config.py
+++ b/mcpgateway/config.py
@@ -1186,6 +1186,34 @@ class Settings(BaseSettings):
 
     # Domain configuration
     app_domain: HttpUrl = Field(default=HttpUrl("http://localhost:4444"))
+    ui_base_url: Optional[HttpUrl] = Field(
+        default=None,
+        description="Trusted base URL for browser-facing UI links. Falls back to APP_DOMAIN plus APP_ROOT_PATH when unset.",
+    )
+
+    @field_validator("ui_base_url")
+    @classmethod
+    def validate_ui_base_url(cls, value: Optional[HttpUrl]) -> Optional[HttpUrl]:
+        """Reject URL components unsuitable for a trusted frontend base.
+
+        Args:
+            value: Configured frontend base URL.
+
+        Returns:
+            Optional[HttpUrl]: Validated frontend base URL.
+
+        Raises:
+            ValueError: If credentials, query parameters, or fragments are present.
+        """
+        if value is None:
+            return None
+        if value.username or value.password:
+            raise ValueError("UI_BASE_URL must not contain credentials")
+        if value.query:
+            raise ValueError("UI_BASE_URL must not contain a query string")
+        if value.fragment:
+            raise ValueError("UI_BASE_URL must not contain a fragment")
+        return value
 
     # Security settings
     secure_cookies: bool = Field(default=True)
--- a/mcpgateway/config.py
+++ b/mcpgateway/config.py
@@ -1633,9 +1633,9 @@ class Settings(BaseSettings):
 
             if self.smtp_enabled and self.ui_base_url is None:
                 logger.warning(
-                    "SMTP_ENABLED=true while UI_BASE_URL is unset. Browser-facing email links will use "
-                    "APP_DOMAIN plus APP_ROOT_PATH. Ensure that fallback serves /accept-invitation/{token}, "
-                    "/reset-password/{token}, and /forgot-password, or configure UI_BASE_URL for the React client."
+                    "SMTP_ENABLED=true while UI_BASE_URL is unset. Invitation links will use APP_DOMAIN plus "
+                    "APP_ROOT_PATH and require /accept-invitation/{token}. Password-recovery links will use legacy "
+                    "/admin routes and require MCPGATEWAY_ADMIN_API_ENABLED=true. Configure UI_BASE_URL for the React client."
                 )
 
         # CSRF secret key fallback to JWT secret key.
--- a/mcpgateway/config.py
+++ b/mcpgateway/config.py
@@ -1632,10 +1632,15 @@ class Settings(BaseSettings):
                 logger.warning("🐛 SECURITY WARNING: Debug mode is enabled in non-dev mode. This may leak sensitive information! Set DEBUG=false for production.")
 
             if self.smtp_enabled and self.ui_base_url is None:
+                password_route_warning = (
+                    "Password-recovery links will use legacy /admin routes because MCPGATEWAY_ADMIN_API_ENABLED=true."
+                    if self.mcpgateway_admin_api_enabled
+                    else "Password-recovery links will use frontend /forgot-password and /reset-password/{token} routes because MCPGATEWAY_ADMIN_API_ENABLED=false."
+                )
                 logger.warning(
                     "SMTP_ENABLED=true while UI_BASE_URL is unset. Invitation links will use APP_DOMAIN plus "
-                    "APP_ROOT_PATH and require /accept-invitation/{token}. Password-recovery links will use legacy "
-                    "/admin routes and require MCPGATEWAY_ADMIN_API_ENABLED=true. Configure UI_BASE_URL for the React client."
+                    "APP_ROOT_PATH and require /accept-invitation/{token}. %s Configure UI_BASE_URL for the React client.",
+                    password_route_warning,
                 )
 
         # CSRF secret key fallback to JWT secret key.
--- a/mcpgateway/middleware/rate_limit_middleware.py
+++ b/mcpgateway/middleware/rate_limit_middleware.py
@@ -105,6 +105,11 @@ class RateLimitMiddleware(BaseHTTPMiddleware):
                 "limit": settings.rate_limit_critical_rpm,
                 "burst": settings.rate_limit_critical_burst,
             },
+            "CRITICAL_INVITATION": {
+                "pattern": r"^/teams/[^/]+/invitations/?$",
+                "limit": settings.rate_limit_critical_rpm,
+                "burst": settings.rate_limit_critical_burst,
+            },
             "HIGH": {
                 "pattern": r"^/(tokens|oauth|rbac)(/|$)",
                 "limit": settings.rate_limit_high_rpm,
--- a/mcpgateway/middleware/rate_limit_middleware.py
+++ b/mcpgateway/middleware/rate_limit_middleware.py
@@ -27,7 +27,7 @@ import logging
 import re
 import threading
 import time
-from typing import Any, Dict, List, Tuple
+from typing import Any, Dict, List, Optional, Tuple
 import uuid
 
 # Third-Party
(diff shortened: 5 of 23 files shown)


gandhipratik203:
**DB session held open across the SMTP round trip**

`mcpgateway/routers/teams.py` awaits delivery between commit and close:

```python
db.commit()
delivery = await invitation_service.deliver_invitation_email(...)
db.close()
```

With `smtp_timeout_seconds` defaulting to 15 (`email_notification_service.py:141`), a slow SMTP host pins a pooled connection that long per invite. `create_team_with_members` is worse: N invitations at concurrency 5 holds the session for roughly `ceil(N/5) * timeout`.

`SessionLocal` uses `expire_on_commit=False` (`mcpgateway/db.py:569`), so invitation attributes stay loaded after close. Moving `db.close()` above the delivery call drops the coupling.

gandhipratik203 (dismissed):
LGTM!

msureshkumar88 (changes requested):
Thanks for this — backend-owned invitation delivery, bounded concurrency, commit-then-deliver ordering, and the `UI_BASE_URL` credential/query/fragment guards are all solid. A few things worth fixing before merge, and one I'd consider blocking.

## Blocking

**1. Default fallback breaks existing password-reset links (regression)**
`build_frontend_url()`'s fallback (used whenever `UI_BASE_URL` is unset — i.e. every deployment today, since the separate React client doesn't exist in this repo yet) drops the `/admin` prefix that `_build_forgot_password_url`/`_build_reset_password_url` used before this PR. The only page-serving routes are still at `/admin/forgot-password` and `/admin/reset-password/{token}` (`admin.py:4682`, `:4744`) — the bare path the fallback now generates 404s. `test_build_password_reset_urls` asserts the new (broken) URL, so this won't surface in CI.

Suggestion: keep the fallback pointed at `/admin/...` when `UI_BASE_URL` is unset, and only switch to the bare path once a `UI_BASE_URL` is actually configured. Worth a regression test asserting the `/admin`-prefixed fallback stays intact.

**2. `UI_BASE_URL=""` crashes startup**
`Optional[HttpUrl]` has no empty-string → `None` coercion, so `UI_BASE_URL=` (a natural way to unset an override, and how `.env.example` ships it commented) raises `ValidationError` on boot. Worth a `field_validator` (or `mode="before"`) treating `""` as unset, plus a test in `test_config.py`.

**3. Delivery errors after commit can surface as a misleading 500**
In `create_team` / `invite_team_member`, the post-commit email-delivery call still sits inside the outer `try` whose catch-all returns HTTP 500 — but the team/invitation is already persisted at that point. If `build_frontend_url` (or delivery) raises rather than returning `FAILED`, the client sees a 500 for a row that already exists, and a retry can hit a uniqueness-constraint error. Existing tests only mock delivery to *return* `FAILED`, not raise — worth adding a raise-path test and moving the delivery await outside the row-creation try/except (or wrapping it in its own try that never surfaces as a request failure).

## Worth considering (non-blocking)

- `create_team` synchronously awaits SMTP delivery for every seeded invitation (up to `MAX_TEAM_MEMBER_SEEDS=500`, concurrency=5, fresh SMTP/TLS connection each) before responding — for a large seed this could take a long time and risk proxy timeouts, especially since the response doesn't report per-invitation delivery status anyway. Might be worth deferring this to a background task.
- `deliver_invitation_emails`' per-item exception fallback recomputes `build_frontend_url(...)` instead of reusing the URL already built in `deliver_invitation_email`, and that second call isn't guarded — minor duplication, but also an unguarded exception path.
- `deliver_team_invitation_email` logs raw `invitation_id` while the equivalent site in `team_invitation_service.py` wraps the same value with `SecurityValidator.sanitize_log_message` — worth aligning for consistency, even though `invitation_id` is server-generated today.
- `InvitationDeliveryResult.warning` is only set for `FAILED`, not `DISABLED` — a caller surfacing `warning` sees nothing when SMTP is simply off, which reads as "no problem" when the invitee actually has no way to learn about the invitation.

No alembic migration needed here (nothing new is persisted), and I didn't see anything outside the stated scope.

msureshkumar88 (changes requested):
## Review

Thanks for the work getting real email delivery wired into the invitation and password-reset flows — the concurrency-bounded fan-out in `team_invitation_service.py` and the separate-session delivery pattern are solid choices. A few things worth addressing before merge, and one design question on the default link shape.

### Blocking

**1. Default-config password-reset link 404s (`mcpgateway/services/email_notification_service.py:57`)**

`build_frontend_url()`'s no-`UI_BASE_URL` fallback routes password-reset/forgot-password links through `/admin/reset-password`. But `admin_router` (which owns those routes) is only mounted when `MCPGATEWAY_ADMIN_API_ENABLED=true` (`mcpgateway/config.py:1070`), and that setting **defaults to `false`**. In a fresh deployment with `UI_BASE_URL` unset and the admin API left at its default, the emailed reset link returns 404.

This also runs counter to issue #5813's acceptance criteria, which calls for the link to *not* use `/admin/reset-password`. Worth reconsidering the fallback target (or gating the fallback on `MCPGATEWAY_ADMIN_API_ENABLED`, with a clear warning/log when neither `UI_BASE_URL` nor the admin API is available) — otherwise this ships a dead link for the common default deployment.

### Security (worth hardening)

**2. Invitation email send has no dedicated rate limit (`mcpgateway/routers/teams.py:751`) — CWE-770**

`POST /{team_id}/invitations` now triggers a real outbound SMTP send per call, bounded only by team member-capacity checks, not a request-rate limit. A malicious or compromised team-owner account could repeatedly invite arbitrary addresses across teams it owns, using the gateway's SMTP credentials as an unthrottled mail relay. Recommend a per-user/per-team rate limit on this endpoint specifically, since it's a new externally-observable side effect that didn't exist before this PR.

### Functionally-impacting

**3. Sync vs. async delivery inconsistency (`mcpgateway/routers/teams.py:791`)**

`invite_team_member()` awaits SMTP delivery synchronously (up to `smtp_timeout_seconds`, default 15s) before responding, while `create_team`'s bulk-invite path defers the same delivery work to `BackgroundTasks`. Same underlying operation, two different latency profiles — a slow/unreachable SMTP server makes single-invite calls hang for up to 15s even though the invitation already committed. Consider aligning both paths through `BackgroundTasks`.

**4. FAILED status conflates transport error with misconfiguration (`mcpgateway/services/email_notification_service.py:278`)**

`deliver_team_invitation_email()` only reports `DISABLED` when `smtp_enabled` is `False`. An enabled-but-incomplete SMTP config (missing `smtp_host`/`smtp_from_email`) reports `FAILED` — same as a genuine transport failure — which will read as "delivery broke" during a misconfigured rollout rather than "SMTP isn't fully configured." A distinct status (or at least a differentiated warning message) would make triage faster.

**5. Response construction sits outside the delivery try/except (`mcpgateway/routers/teams.py:800`)**

`TeamInvitationCreateResponse(...)` is built after `db.commit()`/`db.close()`, outside the try/except that guards `deliver_invitation_email`. Not currently a problem, but if the response schema gains a stricter validator later, a failure here would surface as a 500 even though the invitation is already durably persisted — and a client retry would then hit the existing-active-invitation conflict. Worth a comment noting the ordering constraint, or moving construction earlier.

### Test coverage

**6. No end-to-end/integration test for the invitation email path**

`tests/playwright/teams/test_team_invitations.py` and the added unit tests fully mock the router/service/template collaborators. There's no test that drives `POST /{team_id}/invitations` through the real router + service + template render and asserts on `invitation_url`, `email_delivery_status`, `warning`, or the `UI_BASE_URL`-driven link shape end to end. A wiring regression (template context key mismatch, `UI_BASE_URL` not reaching the running app, etc.) would pass the full unit suite today and only surface in production or manual QA.

### Minor

**7. Hardcoded concurrency constant (`mcpgateway/services/team_invitation_service.py:43`)**

`_INVITATION_EMAIL_CONCURRENCY=5` is a module constant rather than a `Settings` field, unlike the existing pattern for similar fan-out bounds (e.g. `settings.max_concurrent_health_checks`). Not blocking, but an operator seeding a large team or tuning SMTP throughput currently has no way to adjust this without a code change.

---

Requesting changes mainly on #1 and #2 — the rest are worth a look but not merge-blocking on their own.

gcgoncalves (dismissed):

gcgoncalves:
> ## Review
> Thanks for the work getting real email delivery wired into the invitation and password-reset flows — the concurrency-bounded fan-out in `team_invitation_service.py` and the separate-session delivery pattern are solid choices. A few things worth addressing before merge, and one design question on the default link shape.
> 
> ### Blocking
> **1. Default-config password-reset link 404s (`mcpgateway/services/email_notification_service.py:57`)**
> 
> `build_frontend_url()`'s no-`UI_BASE_URL` fallback routes password-reset/forgot-password links through `/admin/reset-password`. But `admin_router` (which owns those routes) is only mounted when `MCPGATEWAY_ADMIN_API_ENABLED=true` (`mcpgateway/config.py:1070`), and that setting **defaults to `false`**. In a fresh deployment with `UI_BASE_URL` unset and the admin API left at its default, the emailed reset link returns 404.
> 
> This also runs counter to issue #5813's acceptance criteria, which calls for the link to _not_ use `/admin/reset-password`. Worth reconsidering the fallback target (or gating the fallback on `MCPGATEWAY_ADMIN_API_ENABLED`, with a clear warning/log when neither `UI_BASE_URL` nor the admin API is available) — otherwise this ships a dead link for the common default deployment.
> 
> ### Security (worth hardening)
> **2. Invitation email send has no dedicated rate limit (`mcpgateway/routers/teams.py:751`) — CWE-770**
> 
> `POST /{team_id}/invitations` now triggers a real outbound SMTP send per call, bounded only by team member-capacity checks, not a request-rate limit. A malicious or compromised team-owner account could repeatedly invite arbitrary addresses across teams it owns, using the gateway's SMTP credentials as an unthrottled mail relay. Recommend a per-user/per-team rate limit on this endpoint specifically, since it's a new externally-observable side effect that didn't exist before this PR.
> 
> ### Functionally-impacting
> **3. Sync vs. async delivery inconsistency (`mcpgateway/routers/teams.py:791`)**
> 
> `invite_team_member()` awaits SMTP delivery synchronously (up to `smtp_timeout_seconds`, default 15s) before responding, while `create_team`'s bulk-invite path defers the same delivery work to `BackgroundTasks`. Same underlying operation, two different latency profiles — a slow/unreachable SMTP server makes single-invite calls hang for up to 15s even though the invitation already committed. Consider aligning both paths through `BackgroundTasks`.
> 
> **4. FAILED status conflates transport error with misconfiguration (`mcpgateway/services/email_notification_service.py:278`)**
> 
> `deliver_team_invitation_email()` only reports `DISABLED` when `smtp_enabled` is `False`. An enabled-but-incomplete SMTP config (missing `smtp_host`/`smtp_from_email`) reports `FAILED` — same as a genuine transport failure — which will read as "delivery broke" during a misconfigured rollout rather than "SMTP isn't fully configured." A distinct status (or at least a differentiated warning message) would make triage faster.
> 
> **5. Response construction sits outside the delivery try/except (`mcpgateway/routers/teams.py:800`)**
> 
> `TeamInvitationCreateResponse(...)` is built after `db.commit()`/`db.close()`, outside the try/except that guards `deliver_invitation_email`. Not currently a problem, but if the response schema gains a stricter validator later, a failure here would surface as a 500 even though the invitation is already durably persisted — and a client retry would then hit the existing-active-invitation conflict. Worth a comment noting the ordering constraint, or moving construction earlier.
> 
> ### Test coverage
> **6. No end-to-end/integration test for the invitation email path**
> 
> `tests/playwright/teams/test_team_invitations.py` and the added unit tests fully mock the router/service/template collaborators. There's no test that drives `POST /{team_id}/invitations` through the real router + service + template render and asserts on `invitation_url`, `email_delivery_status`, `warning`, or the `UI_BASE_URL`-driven link shape end to end. A wiring regression (template context key mismatch, `UI_BASE_URL` not reaching the running app, etc.) would pass the full unit suite today and only surface in production or manual QA.
> 
> ### Minor
> **7. Hardcoded concurrency constant (`mcpgateway/services/team_invitation_service.py:43`)**
> 
> `_INVITATION_EMAIL_CONCURRENCY=5` is a module constant rather than a `Settings` field, unlike the existing pattern for similar fan-out bounds (e.g. `settings.max_concurrent_health_checks`). Not blocking, but an operator seeding a large team or tuning SMTP throughput currently has no way to adjust this without a code change.
> 
> Requesting changes mainly on #1 and #2 — the rest are worth a look but not merge-blocking on their own.

- 1. 404 page: If the ui is disabled, the password reset won't be triggered anyway, so it's a non-issue.
- 2. Rate limit: Is already applied to the endpoint. No need to apply to the function itself.
- 3. Valid, but non-blocking. We'd not want synchronous bulk-invites, but waiting for a single invite is fine for now.
- 4. FAILED status on misconfiguration error:  Valid, but non-blocking. Worth solving on a separare issue.
- 5. Templating outside try/except: Valid, but non-blocking. Worth solving on a separare issue.
- 6. E2E test: Valid.
- 7. Email concurrency: Valid, but oustide scope. Worth solving on a separare issue.

vishu-bh:
Thanks @gcgoncalves :)

msureshkumar88:
Thanks for the detailed responses — replying point by point since a couple of these change the picture:

**#1 (password-reset 404) — this is still a real issue, not a non-issue.**

`POST /auth/email/forgot-password` doesn't depend on the UI being enabled. It's gated only on `email_auth_enabled` (`mcpgateway/config.py:976`), which defaults to `True`. `MCPGATEWAY_ADMIN_API_ENABLED` (which owns `/admin/reset-password`) defaults to `False` and is a completely separate flag. So in a stock deployment: `email_auth_enabled=True`, `admin_api_enabled=False`, `UI_BASE_URL` unset — a client can call `forgot-password` right now, an email goes out, and the link inside 404s because `/admin/reset-password` isn't mounted. The two settings aren't coupled, so "UI disabled → reset never triggered" doesn't hold.

Given that, and that issue #5813's acceptance criteria explicitly calls for *not* using `/admin/reset-password` — could we gate the fallback (e.g., only use `/admin/...` when `MCPGATEWAY_ADMIN_API_ENABLED` is true, otherwise log a warning and fall back to a bare path or skip the link) before merge?

**#2 (rate limit) — softening this, but not withdrawing it.**

You're right that `RateLimitMiddleware` covers `/teams/{team_id}/invitations` — I'd missed that unmatched paths fall through to `default_tier`. That said, the default tier is the LOW bucket (`rate_limit_low_rpm`, default 500/min) per IP+user+team — sized for read-heavy traffic like `/health`/`/metrics`, not an endpoint whose side effect is an outbound SMTP send using the gateway's own credentials. 500 invite emails/min per user is still a lot of headroom for using this as a mail relay. Not blocking on its own, but worth either adding `/teams/.*/invitations` to a stricter tier or filing a fast-follow — your call.

**#3–#7**: agreed these are non-blocking / fast-follow, thanks for confirming.

**One more, found while re-checking #1's blast radius:** the invitation link has the same shape of problem, but with no fallback at all. `build_frontend_url("/accept-invitation", token)` produces `{app_domain}/accept-invitation/{token}` when `UI_BASE_URL` is unset — there's no route anywhere in this repo matching `/accept-invitation` (the only related endpoint is the authenticated `POST /{team_id}/invitations/{token}/accept` JSON API, a different path entirely, meant to be called by a frontend rather than visited directly). If that's intentional (link only meant for the separate React client), worth a doc note so it's not mistaken for a bug later — but as shipped, every invitation email sent from a default deployment links to a 404 today.

msureshkumar88 (changes requested):
Ran a real end-to-end verification against this branch (head `d347abe41`) rather than continuing on static analysis — a live gateway, a live local SMTP catcher (no external mail), real HTTP requests, real captured emails, real link resolution. Script + catcher are local tooling (`scripts/verify-pr-6209-e2e.sh`, `scripts/e2e_smtp_catcher.py`), not part of this PR's diff.

## Config used

Two full server lifecycles, same commit, different only in the two disputed flags:

```
JWT_SECRET_KEY=<generated>
AUTH_ENCRYPTION_SECRET=<generated>
DATABASE_URL=sqlite:///./e2e_pr6209_<profile>.db
AUTH_REQUIRED=true
EMAIL_AUTH_ENABLED=true
ALLOW_TEAM_INVITATIONS=true
ALLOW_TEAM_CREATION=true
PASSWORD_RESET_ENABLED=true
UI_BASE_URL=<unset>          # true of every real deployment today — no separate frontend ships in this repo
APP_DOMAIN=http://127.0.0.1:<port>
SMTP_ENABLED=true
SMTP_HOST=127.0.0.1  SMTP_PORT=<catcher>  SMTP_USE_TLS=false  SMTP_USE_SSL=false
PLUGINS_ENABLED=false
```

- **Profile A ("quickstart")**: `MCPGATEWAY_ADMIN_API_ENABLED=true`, `MCPGATEWAY_UI_ENABLED=true` — what `cp .env.example .env` (the documented quickstart) actually gives you.
- **Profile B ("library default")**: both `false` — `mcpgateway/config.py` defaults as-is, no `.env` overrides.

## Commands run

```
bash scripts/verify-pr-6209-e2e.sh
```

Internally: starts `uvicorn mcpgateway.main:app` for real (not `TestClient`) per profile on its own port/DB, bootstraps an admin JWT, creates a team, invites a member via `POST /teams/{team_id}/invitations`, waits for the invitation email to land on the local SMTP catcher, extracts the invitation URL from the real email body, `curl -i`'s it. Same for `POST /auth/email/forgot-password` → captured reset email → extracted URL → `curl -i`. Also captures `X-RateLimit-*` response headers off the invitation POST, and runs the pytest files covering `teams.py`, `team_invitation_service.py`, `email_notification_service.py`, `email_auth.py`, and the `UI_BASE_URL` validator.

## Output

```
PROFILE_A_INVITE_LINK_STATUS=404
PROFILE_A_RESET_LINK_STATUS=200
PROFILE_A_RESET_LINK_URL=http://127.0.0.1:46209/admin/reset-password/adLrp3QVcT9key8e6uJzcBhSTII09_jgGUmQgBFMMu_4P3aycb2MWgbgpT798xTc
PROFILE_A_RATE_LIMIT_HDR=500
PROFILE_B_INVITE_LINK_STATUS=404
PROFILE_B_RESET_LINK_STATUS=404
PROFILE_B_RESET_LINK_URL=http://127.0.0.1:46210/admin/reset-password/rEASJRKFhDSOSWDvMC3TutmCZPsTxy06ixgpwRE5YbbeNVgwu5teeOyPMsW0NpKM
PROFILE_B_RATE_LIMIT_HDR=500
```

`POST /auth/email/forgot-password` returned `200` in **both** profiles — the request always succeeds regardless of `MCPGATEWAY_ADMIN_API_ENABLED`/`MCPGATEWAY_UI_ENABLED`; only whether the emailed link *resolves* depends on those flags.

Regression: pytest over the touched-area test files — all dots, no `F`, ~12 pre-existing skips (RBAC-mock/complex-integration skips unrelated to this PR), exit 0.

| Check | Profile A (admin API on) | Profile B (library default) |
|---|---|---|
| Password-reset link resolves | **200** ✅ | **404** ❌ |
| Invitation accept-link resolves | **404** ❌ | **404** ❌ |
| `POST forgot-password` itself succeeds | 200 | 200 |
| Rate limit on `/teams/{id}/invitations` | `X-RateLimit-Limit: 500` (generic LOW tier) | same |
| Regression suite | pass, 0 failures | pass, 0 failures |

## What this settles

- **Invitation links are dead in every configuration that exists today**, not just an edge case — a real invitation email was sent, a real link extracted, a real 404 came back, in both profiles. There's no route in this repo matching `/accept-invitation` in any config; the only related endpoint is the authenticated `POST /{team_id}/invitations/{token}/accept` JSON API, a different path meant for a frontend to call, not to visit directly.
- **Password-reset links only resolve when `MCPGATEWAY_ADMIN_API_ENABLED=true`.** That happens to be true for the `.env.example` quickstart path, but is `false` by library default, and the `forgot-password` request itself succeeds either way — so a deployer who assembles config by hand (or via IaC referencing `mcpgateway/config.py` defaults rather than `.env.example`) ships a dead reset link with no error or warning anywhere.
- Rate limiting confirmed empirically at the generic LOW-tier fallback (500 rpm), not an endpoint-specific limit — as discussed in the earlier thread, non-blocking but worth a fast-follow.

## Requesting changes

1. **Blocking** — Fix or gate the invitation-accept link so it doesn't 404 unconditionally (e.g., point at a real acceptance surface, or clearly document that `UI_BASE_URL` must be configured for invitations to be usable, with a startup warning when `smtp_enabled=true` and `UI_BASE_URL` is unset).
2. **Blocking** — Fix the password-reset fallback so it doesn't depend on `MCPGATEWAY_ADMIN_API_ENABLED` being explicitly turned on (or gate the `/admin` prefix on that setting and fail loudly/log a warning otherwise), consistent with issue #5813's acceptance criteria.
3. Non-blocking, already discussed and conceded upthread: dedicated rate-limit tier for `/teams/{id}/invitations`, sync-vs-async delivery inconsistency between `invite_team_member` and `create_team`, `FAILED` vs `DISABLED` status granularity, hardcoded `_INVITATION_EMAIL_CONCURRENCY`, and an in-repo e2e test for this path.

Happy to re-run this same script against a follow-up commit — it's fast (~2 real server lifecycles) and gives a direct pass/fail on both link checks.

gcgoncalves (dismissed):

msureshkumar88 (changes requested):
Re-reviewed this PR against the actual commits since the last round (not the review-comment thread) — most of the prior feedback has been genuinely addressed:

- DB session held open during email delivery — fixed
- `UI_BASE_URL=""` crash — fixed
- delivery exception surfacing as an unhandled 500 — fixed
- missing rate-limit tier — fixed
- log sanitization inconsistency — fixed
- missing `DISABLED` warning — fixed

Nice cleanup overall. One regression from the last round still needs attention before this can merge.

## Blocking

**`mcpgateway/services/email_notification_service.py:54`** — `3f465a308` ("fix: harden frontend email link behavior") removed the `/admin` prefix that `build_frontend_url()`'s no-`UI_BASE_URL` fallback used to add for `/forgot-password` and `/reset-password`. The follow-up commit `ec7f08c6d` only added a startup warning log — it didn't restore the prefix.

Verified live:
```
build_frontend_url('/reset-password', 'TOKEN123')
-> http://localhost:8080/reset-password/TOKEN123
```
The actual registered routes are `/admin/reset-password/{token}` (HTML page, `admin_router` prefix=`/admin`) and `/auth/email/reset-password/{token}` (JSON API). The bare fallback path matches neither, so the emailed link 404s unconditionally in every configuration — including the documented quickstart profile (`MCPGATEWAY_ADMIN_API_ENABLED=true`, `UI_BASE_URL` unset) that worked before this commit.

`test_build_frontend_url_falls_back_to_domain_and_root_path` and the password-reset URL test were updated to assert this broken output as correct, so the suite is green but not actually covering the regression. Could you restore the `/admin` prefix branch and update those two tests back to asserting the prefixed URL?

## Non-blocking (carried over from earlier rounds, still open — fine as fast-follows)

- `invite_team_member` (`routers/teams.py`) still awaits SMTP delivery synchronously (up to `smtp_timeout_seconds`) instead of using `BackgroundTasks` like `create_team`'s invite path does.
- `deliver_team_invitation_email` reports `FAILED` identically for a genuine SMTP error and for an enabled-but-incomplete SMTP config, making misconfiguration indistinguishable from an outage.
- `_INVITATION_EMAIL_CONCURRENCY = 5` in `team_invitation_service.py` is a hardcoded constant rather than a `Settings` field, unlike the equivalent `max_concurrent_health_checks` pattern elsewhere.
- The router-level `try/except` around `deliver_invitation_email(...)` in `invite_team_member` is unreachable dead code — the service method already catches everything internally and never raises. Harmless, but worth trimming at some point.

msureshkumar88 (changes requested):
Thanks for the continued work on this — the DB-session-release-before-SMTP pattern, `UI_BASE_URL=""` handling, delivery-exception-as-500 fix, and log sanitization from earlier rounds all hold up on re-check. Two blocking items remain before this can merge.

## Blocking

**1. `/admin` fallback still isn't gated on `MCPGATEWAY_ADMIN_API_ENABLED` — #5813's acceptance criteria aren't met for the default deployment**

`mcpgateway/services/email_notification_service.py:56-58` (from `5b40b3975`):
```python
base_url = f"{app_domain}/{root_path}" if root_path else app_domain
if path in {"/forgot-password", "/reset-password"}:
    path = f"/admin{path}"
```
This is unconditional. `mcpgateway_admin_api_enabled` defaults to `False` (`config.py:1070`), and `admin_router` — which owns `/admin/reset-password` and `/admin/forgot-password` — is only mounted `if settings.mcpgateway_admin_api_enabled` (`mcpgateway/api/v1/__init__.py:263`). So in the library-default profile (no `.env` overrides, `UI_BASE_URL` unset), every password-reset/forgot-password email still links to a route that isn't mounted → 404. That's the same `PROFILE_B_RESET_LINK_STATUS=404` result from the earlier e2e verification round, still reproducible on current HEAD.

It's also a direct conflict with #5813's own acceptance criteria ("the link does not use `/admin/reset-password`" when `UI_BASE_URL` is unset). Suggest gating the `/admin` branch on `settings.mcpgateway_admin_api_enabled`, falling through to the bare `APP_DOMAIN + APP_ROOT_PATH` path otherwise — and updating `test_build_password_reset_urls`, `test_build_frontend_url_falls_back_to_domain_and_root_path`, and `test_smtp_without_ui_base_url_warns_about_frontend_routes` to cover both states instead of asserting only the `/admin`-prefixed output.

**2. `CRITICAL_INVITATION` rate-limit tier is method-agnostic and also throttles the read-only list endpoint — CWE-770**

`mcpgateway/middleware/rate_limit_middleware.py:108-112`:
```python
"CRITICAL_INVITATION": {
    "pattern": r"^/teams/[^/]+/invitations/?$",
    "limit": settings.rate_limit_critical_rpm,   # default 10
    "burst": settings.rate_limit_critical_burst, # default 0
},
```
`dispatch()`/`get_endpoint_tier()`/`_get_tier_name()` match on `request.url.path` only — `request.method` is never checked (`rate_limit_middleware.py:217-220`). This regex matches both `POST /teams/{id}/invitations` (sends SMTP, the intended target) and `GET /teams/{id}/invitations` (`list_team_invitations`, a pure read), and both now share the same bucket. A client polling the pending-invitations list can burn the 10rpm/burst-0 quota needed to actually send invitations — which also undercuts the abuse protection this tier was added for. `test_endpoint_tier_critical_for_team_invitations` checks tier lookup generically and doesn't cover method differentiation, so this isn't caught by CI today.

Suggest scoping the match to `POST` specifically (either extend the tier config with a method predicate, or narrow the regex-matching site), plus a test asserting `GET` stays out of the CRITICAL tier.

Everything else from earlier rounds (e2e test coverage, `invite_team_member` sync-vs-`BackgroundTasks` inconsistency, the redundant triple try/except, hardcoded concurrency constant) is still worth doing but non-blocking, consistent with how it's been triaged so far in this thread.

gcgoncalves:
@msureshkumar88 @vishu-bh  This is going too far.

Looking at the code, I can see:

  1. API Route: POST /teams/{team_id}/invitations in routers/teams.py - creates invitations via REST API
  2. Service Layer: TeamInvitationService.create_invitation() - handles invitation creation
  3. The create_invitation() method in TeamInvitationService does NOT send any email - it just creates the database record

So:
  - yes, you can send team invites via REST API (POST /teams/{team_id}/invitations)
  - the API does NOT automatically send invitation emails
  - The invitation is created in the database with a token, but no email notification is sent

Therefore, there's absolutely no point whatsoever in worrying about the user receiving any emails if `mcpgateway_admin_api_enabled` is set to `false`. We're spending too much energy in an impossible case.

msureshkumar88 (approved):
## Real-environment e2e verification — 7/7 scenarios pass

Built and ran a black-box e2e against a real running gateway process (not mocked internals) to verify this PR fixes issue #5813's three acceptance-criteria user stories.

**Method:** booted `mcpgateway.main:app` under real `uvicorn` (4 separate boots, one per SMTP/UI_BASE_URL combination that can only be set at process start), against a throwaway SQLite DB, driven entirely over HTTP with a real admin login. SMTP scenarios talk the real wire protocol (`EHLO`/`MAIL FROM`/`RCPT TO`/`DATA`/`QUIT`) to a minimal local catcher — nothing in `mcpgateway` was mocked or monkeypatched.

**Config (per boot):** `AUTH_REQUIRED=true`, `EMAIL_AUTH_ENABLED=true`, fresh random `JWT_SECRET_KEY`/`AUTH_ENCRYPTION_SECRET`, `PASSWORD_CHANGE_ENFORCEMENT_ENABLED=false` (unrelated bootstrap flag, disabled purely to log in as the seeded admin), `ALLOW_TEAM_INVITATIONS=true`; then per scenario: `SMTP_ENABLED`/`SMTP_HOST`/`SMTP_PORT` pointed at either the fake SMTP catcher, an unreachable port, or disabled; `UI_BASE_URL` set to `http://localhost:5173` or left unset with `MCPGATEWAY_ADMIN_API_ENABLED=true`.

**Commands:** `POST /auth/email/login` → `POST /teams/` (real org team — invitations are correctly rejected on personal teams) → `POST /teams/{id}/invitations` → `GET /teams/{id}/invitations` ×15 burst → `POST /auth/email/forgot-password`.

**Results:**

| Scenario | Result | Evidence |
|---|---|---|
| SMTP reachable, invite created | ✅ PASS | HTTP 201, `email_delivery_status="sent"`, captured raw SMTP message contains `http://localhost:5173/accept-invitation/<token>` (not `/admin`), team name, inviter, expiry |
| SMTP unreachable | ✅ PASS | HTTP 201 (row stays committed), `email_delivery_status="failed"`, safe `warning`, gateway log line: `WARNING ... "Team invitation email delivery failed for invitation 8ed58e85f9424bbe83a580eac8ba1c04"` — invitation ID only, no token/SMTP detail anywhere in the log |
| `SMTP_ENABLED=false` | ✅ PASS | HTTP 201, `email_delivery_status="disabled"`, fake SMTP catcher recorded **0** connections |
| `UI_BASE_URL` set | ✅ PASS | forgot-password email link: `http://localhost:5173/reset-password/<token>` |
| `UI_BASE_URL` unset + admin API enabled | ✅ PASS | forgot-password email link: `http://localhost:4444/admin/reset-password/<token>` — matches documented fallback |
| 15-request GET burst on `/teams/{id}/invitations` | ✅ PASS | all 15 → HTTP 200, no 429 — confirms the GET/POST rate-limit-tier fix already in this HEAD holds under a real, live rate limiter |
| Black-box probe for a post-commit response-construction failure (the one still-open, non-blocking review item) | not reproduced | 5 rapid duplicate-invite attempts all correctly returned HTTP 400 pre-commit; couldn't force a post-commit 400/500 from the public API — `team_name`/`inviter_name` aren't attacker-adjustable through this endpoint |

Full script, per-scenario raw output, and captured SMTP messages available on request (kept locally, not committed to this PR since it's a throwaway verification harness rather than a repo test).

## Non-blocking items (from an earlier manual review pass, not re-litigated here)

These didn't affect the e2e outcome and aren't required for this approval — flagging for visibility only:

- `teams.py:788-823` — post-commit `TeamInvitationCreateResponse(...)` construction sits inside the outer generic `except` handlers; theoretically a construction failure there could misreport an already-persisted invitation as failed. Not reproducible today (all response field types are well-formed) and not reproducible via this e2e's black-box probe either — worth a defensive follow-up, not a blocker.
- `TeamInvitationService.__init__` builds a Jinja2 environment even on read-only call sites (minor efficiency).
- Triple-nested identical try/except/log/fallback around delivery calls that already never raise (simplification).
- `build_frontend_url()`'s admin-path remap is a hardcoded string-literal set (a future bundled-admin-compatible link would need to remember to add its path here).
- `tests/unit/mcpgateway/utils/test_jq_runner.py` has one unrelated 2-line `@linux_only` decorator change — harmless scope creep, unrelated to this feature.

**Approving** — the fix genuinely resolves #5813 under a real running instance, no regressions found in the surrounding rate-limit/URL-building code it touches.

Lang-Akshay (approved):
LGTM✅

gandhipratik203 (approved):
LGTM!

PR #6210, opened 2026-08-12
feat: add --csv and --html output to benchmark-mcp-tools

### LINKED ISSUE #6206: [CHORE]:  Add --csv and --html output to benchmark-mcp-tools Makefile target
### 🔧 Chore Summary

Add --csv and --html flags to the benchmark-mcp-tools locust invocation so it produces file reports on every run, making it symmetric with load-test.

---

### 🧱 Area Affected

---

### ⚙️ Context / Rationale

benchmark-mcp-tools is a pipeline target that runs on every nightly acceptance test run and uploads evidence to COS. Unlike load-test, it currently produces stdout only - no CSV, no HTML. This means the evidence attachment contains only a log file, with no structured metrics for the performance capture system to parse.

Adding --csv and --html unblocks  https://github.com/IBM/mcp-context-forge/issues/6105 (nightly load-test data capture), which requires a locust_stats.csv-equivalent attachment on the com.ibm.contextforge.benchmark-mcp-tools evidence record.

---

### 📦 Related Make Targets

Reference any relevant Makefile targets that are involved, if applicable. Ex:

make benchmark-mcp-tools - Quick tools-only MCP benchmark against the testing stack

---

### 📋 Acceptance Criteria

- [ ] make benchmark-mcp-tools writes reports/benchmark_mcp_tools_stats.csv, reports/benchmark_mcp_tools_stats_history.csv, reports/benchmark_mcp_tools_failures.csv, and reports/benchmark_mcp_tools.html
- [ ] --only-summary console output is unchanged

---

claudia-gray:
# Pull Request

## 🔗 Related Issue

Closes https://github.com/IBM/mcp-context-forge/issues/6206

---

## 📝 Summary

benchmark-mcp-tools is a nightly pipeline target that uploads evidence to COS but previously produced stdout only — no CSV, no HTML. This adds --csv and --html to its locust invocation, making it symmetric with load-test. Adds mkdir -p reports guard and two report variables (MCP_BENCHMARK_TOOLS_HTML_REPORT, MCP_BENCHMARK_TOOLS_CSV_PREFIX). --only-summary console output is unchanged.

---

## 📏 Reviewability

---

## 🏷️ Type of Change

---

## 🧪 Verification

| Reports created | make benchmark-mcp-tools | Verified: reports/benchmark_mcp_tools_stats.csv and reports/benchmark_mcp_tools.html written

Ran make benchmark-mcp-tools against a local gateway. Reports created successfully:
- reports/benchmark_mcp_tools_stats.csv — written, contains Aggregated row ✅
- reports/benchmark_mcp_tools.html — written ✅
- Console --only-summary output unchanged ✅

---

## ✅ Checklist

---

## Notes

--only-summary is retained — the locust invocation still prints the console summary table after the run. The --csv and --html flags only add file output on top of that. The _stats.csv, _stats_history.csv, and _failures.csv files are all written automatically by locust when a CSV prefix is given; only _stats.csv is consumed by the performance capture system.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^\\.secrets\\.baseline$",
     "lines": null
   },
-  "generated_at": "2026-08-12T15:30:22Z",
+  "generated_at": "2026-08-12T14:47:09Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -390,7 +390,7 @@
         "hashed_secret": "d3ac7a4ef1a838b4134f2f6e7f3c0d249d74b674",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 6079,
+        "line_number": 6087,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -398,7 +398,7 @@
         "hashed_secret": "5932862bcd24dd27d0dc0407ec94fe9d6ea24aeb",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 6576,
+        "line_number": 6584,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -406,7 +406,7 @@
         "hashed_secret": "c77c805e32f173e4321ee9187de9c29cb3804513",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 6588,
+        "line_number": 6596,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -414,7 +414,7 @@
         "hashed_secret": "8fe3df8a68ddd0d4ab2214186cbb8e38ccd0e06a",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 6660,
+        "line_number": 6668,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -422,7 +422,7 @@
         "hashed_secret": "bb589d0621e5472f470fa3425a234c74b1e202e8",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 7076,
+        "line_number": 7084,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -430,7 +430,7 @@
         "hashed_secret": "93ac8946882128457cd9e283b30ca851945e6690",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 7767,
+        "line_number": 7775,
         "type": "Secret Keyword",
         "verified_result": null
       }
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^\\.secrets\\.baseline$",
     "lines": null
   },
-  "generated_at": "2026-08-12T14:47:09Z",
+  "generated_at": "2026-08-13T10:07:37Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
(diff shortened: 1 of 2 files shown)


brian-hussey (dismissed):
LGTM

cafalchio (dismissed):
LGTM

msureshkumar88 (changes requested):
Thanks for adding CSV/HTML output to the benchmark target — nice to have for evidence capture, and the scope is nicely minimal (only `benchmark-mcp-tools` touched, matches #6206).

A couple of things worth fixing before merge:

**1. `MCP_BENCHMARK_HTML_REPORT` / `MCP_BENCHMARK_CSV_PREFIX` use `:=` instead of `?=`**

Every other `MCP_BENCHMARK_*` var in this block (HOST, USERS, SPAWN_RATE, etc.) uses `?=`, which lets a caller override via exported env var — e.g. a nightly pipeline namespacing output per run (`MCP_BENCHMARK_CSV_PREFIX=/evidence/run-42/...`). With `:=`, the makefile value always wins over the environment, so these two new vars silently ignore any override and always write to the same hardcoded path. Suggest switching both to `?=` to match the existing convention.

**2. Var names aren't scoped to `benchmark-mcp-tools`**

`MCP_BENCHMARK_HTML_REPORT` / `MCP_BENCHMARK_CSV_PREFIX` read as generic benchmark output vars, but they're hardcoded to the tools-benchmark filename and only used inside `benchmark-mcp-tools`. If `benchmark-mcp-mixed` (right below) later grows its own `--html`/`--csv` output, it's an easy trap to reuse these same vars and have both targets clobber the same `reports/benchmark_mcp_tools.*` files. Renaming to something like `MCP_BENCHMARK_TOOLS_HTML_REPORT` / `MCP_BENCHMARK_TOOLS_CSV_PREFIX` would remove the ambiguity.

Non-blocking, worth a thought:

- Output filenames are fixed/unversioned with no locking, so two concurrent runs (e.g. a manual debug run overlapping a nightly job) would write to the same CSV/HTML at once.
- The PR checklist marks "tests added/updated" but the diff is Makefile-only with no automated check that the csv/html output actually gets produced — just a note that would be good to reconcile.

claudia-gray:
@msureshkumar88 Thanks for that. Both blocking items were address in recent commit. I changed to "?=" to stay consistent and renamed MCP_BENCHMARK_TOOLS_HTML_REPORT and MCP_BENCHMARK_TOOLS_CSV_PREFIX.

Non-blocking points: I won't address the concurrent runs as there seems to be no issue with it currently but happy to revisit it later if something comes up.

msureshkumar88 (approved):
Thanks for addressing the feedback — both blocking points from my earlier review are resolved in `dbb109dc5`:

- `MCP_BENCHMARK_TOOLS_HTML_REPORT` / `MCP_BENCHMARK_TOOLS_CSV_PREFIX` now use `?=` (matches every other `MCP_BENCHMARK_*` var, so nightly pipeline env overrides work).
- Vars renamed/scoped to `_TOOLS_` so `benchmark-mcp-mixed` can't collide with them later.

Re-verified against the diff:
- Matches issue #6206's acceptance criteria exactly (csv/html added, `--only-summary` retained, `mkdir -p reports` guard added) — no scope creep beyond `benchmark-mcp-tools`.
- `benchmark-mcp-tools-300` (distributed variant) and `benchmark-mcp-mixed` are untouched and don't share these vars — no clobber risk introduced.
- Pure Makefile/tooling change, no schema/migration/app-code involved, non-breaking (additive only).
- `.secrets.baseline` diff is just line-number/timestamp churn, unrelated to the fix.

Note: I attempted to spin up the local docker `testing-up` stack to run `make benchmark-mcp-tools` live end-to-end, but hit repeated environment instability (redis container dying ~60-90s into startup across multiple clean retries, unrelated to this diff) and didn't get a stable stack up in this session. This approval is based on static/diff review + green CI, not a live run — flagging that gap rather than claiming an e2e I didn't actually complete. Given the change is a minimal, mechanical Makefile addition already manually verified by the author (per PR description) and CI is green, I'm comfortable approving on that basis.

LGTM.

PR #6231, opened 2026-08-13
chore: remove cpex_compat.py shim since cpex>=0.1.2 is declared

### LINKED ISSUE #5996: [CHORE]: Remove cpex_compat.py once cpex>=0.1.2 is the established floor
## Context

`mcpgateway/plugins/cpex_compat.py` was introduced in #5785 to guard against
older CPEX builds (< 0.1.2) that do not expose `ControlExecutionRecord`. The
guard is a single cached `try/import` that degrades gracefully to a no-op on
older deployments.

`pyproject.toml` was simultaneously bumped from `cpex>=0.1.1` to `cpex>=0.1.2`
in the same PR, but the compat layer was intentionally retained because:

- Existing deployments may still be running 0.1.1 within the declared-but-now-stale
  `>=0.1.1` window until they upgrade.
- The guard is cheap (one import at startup, cached for the process lifetime).

## Work to do

Once `cpex>=0.1.2` has been the declared minimum for at least one full release
cycle and there is no realistic path for a 0.1.1 deployment to encounter this code:

1. Delete `mcpgateway/plugins/cpex_compat.py`.
2. Replace all `execution_records_supported()` guard calls in `control_telemetry.py`
   with direct imports of `ControlExecutionRecord` / `ControlExecutionStatus`.
3. Replace `get_executions(result)` calls with direct `list(result.executions or [])`.
4. Delete `tests/unit/mcpgateway/plugins/test_cpex_compat.py`.
5. Update `mcpgateway/plugins/control_telemetry.py` imports accordingly.

## Definition of Done

- `cpex_compat.py` removed, no references remain.
- All existing control-telemetry tests pass with the simplified import.
- `make ruff bandit interrogate` green.

## Related

- Cleanup for: #5785
- `pyproject.toml` already at `cpex>=0.1.2` since #5785

prakhar-singh1928:
# Pull Request

## 🔗 Related Issue

Closes #5996

---

## 📝 Summary

Removes the `cpex_compat.py` compatibility shim that was introduced in #5785 to guard against pre-0.1.2 CPEX deployments. Now that `cpex>=0.1.2` has been the declared minimum for a full release cycle and there is no realistic path for a 0.1.1 deployment to encounter this code, the guard is dead weight.

- Deletes `mcpgateway/plugins/cpex_compat.py` and its test file
- Replaces the two `execution_records_supported()` guard calls in `control_telemetry.py` with unconditional logic; replaces `get_executions(result)` with `try/except`-guarded `getattr(result, "executions", None)` access
- Removes stale references to the old guard in `config.py` comments and field descriptions
- Fixes 14 incomplete `invoke_hook()` test fixtures in `test_tool_service.py` and `test_tool_service_coverage.py` that used bare `Mock()`/`SimpleNamespace` objects missing the `executions` field introduced in cpex 0.1.2; replaces them with real `PluginResult` instances or adds `executions=[]`
- Adds two new tests covering the "executions raises or is non-iterable → `[]`" path in `ControlTelemetryAccumulator.add()`

---

## 📏 Reviewability

---

## 🏷️ Type of Change

- [ ] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

Real-time import and instantiation verified against cpex 0.1.3 — all 10 integration scenarios passed (normal record, None result, empty executions, denial flag, aggregate counts, `_per_control_attributes`, flattened projection, enforcement point, multi-record with error status, timeout status).

| Check | Command | Status |
|---|---|---|
| Lint suite | `make ruff` | ✅ |
| Security scan | `make bandit` | ✅ (no new findings in changed files) |
| Docstring coverage | `make interrogate` | ✅ 100% on `control_telemetry.py` |
| Unit tests (plugin suite) | `pytest tests/unit/mcpgateway/plugins/` | ✅ passed, 33 skipped |
| Unit tests (tool service) | `pytest tests/unit/mcpgateway/services/test_tool_service.py tests/unit/mcpgateway/services/test_tool_service_coverage.py` | ✅ passed, 1 skipped |
| Combined target | `pytest tests/unit/mcpgateway/plugins/ tests/unit/mcpgateway/services/test_tool_service*.py` | ✅ passed, 1 skipped, 0 failed |

---

## ✅ Checklist

---

## 📓 Notes (optional)

**Why `getattr(result, "executions", None) or []` with a `try/except` guard?**

`invoke_hook()` always returns a real `PluginResult` in production, but a number of test fixtures across the codebase use `SimpleNamespace` or `Mock` objects that set only the fields each test cares about. Using `getattr` with a default keeps `add()` forward-compatible with any lightweight stub while still reading directly off the real field on `PluginResult` at runtime. The surrounding `try/except Exception` (matching the `BLE001` pattern used elsewhere in the file) ensures a telemetry-only defect — a raising or non-iterable `executions` attribute — can never fail an otherwise-successful tool call. The `except` block silently returns `[]`; if `executions` is well-formed but the downstream record processing misbehaves, that will still propagate naturally through the standard error handler in `record_control_telemetry`.

**Why fix the test fixtures rather than rely solely on the guard?**

The fixtures were incomplete — `invoke_hook()` returns `PluginResult`, and `PluginResult` has carried `executions: list[ControlExecutionRecord]` since cpex 0.1.2. Updating them to match the real contract means test failures surface genuine production bugs rather than fixture gaps. The `try/except` guard is a belt-and-suspenders safety net for truly unexpected runtime conditions, not a substitute for correct fixtures.


vishu-bh (commented):
Good cleanup. 
PR removes obsolete CPEX compatibility shim, updates telemetry handling, and fixes test fixtures to match current PluginResult contract.

LGTM 🚀

vishu-bh (dismissed):

msureshkumar88 (changes requested):
Thanks for cleaning this up — removing the shim now that `cpex>=0.1.2` is the declared floor is the right move, and this is a nicely minimal, backend-only chore (no UI, no Alembic migration needed since there's no schema change). The test fixture updates across `test_tool_service.py` / `test_tool_service_coverage.py` look correct and necessary.

Two things worth addressing before merge:

## Blocking

**1. `mcpgateway/plugins/control_telemetry.py` (around line 116) — lost the "never raises" guarantee**

The old `cpex_compat.get_executions()` wrapped extraction in `try/except Exception: return []`. The inline replacement (`list(getattr(result, "executions", None) or [])`) doesn't carry that guard forward.

`.add()` is called unguarded from several sites in `tool_service.py` (around lines 5445, 6363, 6456, 6588). One of those calls (~4724) happens inside the `except (asyncio.TimeoutError, ToolTimeoutError)` handler — if extraction raises there, the exception isn't caught by a sibling `except` on the same `try` and escapes `invoke_tool()` entirely uncaught, rather than being converted into a `ToolInvocationError` like it would at the other call sites.

This isn't hypothetical — this PR itself needed to patch 14 test fixtures because bare `Mock()`/`SimpleNamespace()` objects produced a truthy, non-iterable `.executions` that would trip this exact path. Worth restoring a try/except (or equivalent guard) around the extraction so a telemetry-only defect can't fail an otherwise-successful tool call.

**2. Test coverage gap for the above**

The deleted `test_cpex_compat.py::TestGetExecutions::test_returns_empty_on_exception` was the only test covering "executions attribute raises or is malformed → return `[]`." Nothing in `test_control_telemetry.py` or `test_record_control_telemetry.py` replaces that coverage. Adding a case that exercises `.add()` with a raising/non-iterable `.executions` would close the gap and guard against regressing point 1.

## Non-blocking / optional

- PR description mentions adding `from cpex.framework import ControlExecutionRecord, ControlExecutionStatus` and verifying against cpex 0.1.3 — doesn't look like that import made it into the diff. Not a functional issue (the getattr-based duck typing works fine), just flagging the description vs. shipped code mismatch in case it was intentional.
- The PR checklist notes the linked issue isn't labeled `triage`, but #5996 currently shows the `triage` label — probably just needs a label update on the issue side, not a PR change.
- `if result is not None else []` in the same block looks redundant since `getattr(result, "executions", None) or []` already returns `[]` safely when `result` is `None` — could simplify, matches the idiom used elsewhere in the codebase (a2a_service.py, prompt_service.py, tool_service.py).
- Minor: the module docstring's "Feature gate" section now has just one bullet after the CPEX-version bullet was removed — could read as a plain sentence instead.

Happy to take another look once the guard + test are in.

prakhar-singh1928:
Thanks for the detailed review — both blocking points are addressed in the latest commit.

**Blocking 1 — restored "never raises" guarantee in `add()`**

Wrapped the extraction in `try/except Exception` (matching the `BLE001` pattern used elsewhere in the file), so any exception from a malformed or raising `executions` attribute silently falls back to `[]` rather than escaping `add()`. Also took the opportunity to drop the now-redundant `if result is not None else []` guard — `getattr(None, "executions", None) or []` already returns `[]` safely, matching the idiom used in `a2a_service.py`, `prompt_service.py`, and `tool_service.py` as you noted.

**Blocking 2 — test coverage for the exception path**

Added two new cases to `TestAccumulatorAdd` in `test_control_telemetry.py`:
- `test_noop_when_executions_raises` — a `@property` that throws `RuntimeError`; verifies `add()` doesn't propagate and leaves `records == []`
- `test_noop_when_executions_not_iterable` — `executions=42` (truthy, non-iterable); verifies `list()` TypeError is swallowed

These directly replace the deleted `TestGetExecutions::test_returns_empty_on_exception` coverage.

**Non-blocking notes**

- The `ControlExecutionRecord`/`ControlExecutionStatus` import mentioned in the PR description was intentionally not added — the duck-typing approach works fine and avoids a hard import site.
-  updated the `triage` label on #5996

msureshkumar88 (changes requested):
Thanks for the quick turnaround on `26bff444f` — the try/except guard around `executions` extraction and the added test coverage address both blocking points from the earlier review.

One more gap in the same spot, same bug class as the one just fixed:

## Blocking

**`mcpgateway/plugins/control_telemetry.py:133`** — the `continue_processing` read is still unguarded:

```python
if result is not None and not getattr(result, "continue_processing", True):
```

`getattr(obj, name, default)` only suppresses `AttributeError` when the attribute is *missing*. It does not catch an exception raised *by* the attribute itself if it's a property/descriptor. Repro:

```python
class Boom:
    continue_processing = property(lambda self: (_ for _ in ()).throw(RuntimeError("boom")))
    executions = property(lambda self: [])

acc.add(Boom(), hook="pre")  # -> RuntimeError escapes .add() uncaught
```

`.add()` is called unguarded from 6 sites in `tool_service.py` (lines 4724, 5445, 6363, 6456, 6588, 6725). The site at 4724 is inside the timeout-post-invoke handler with no surrounding try/except at that call site — a raising `continue_processing` there would fail an otherwise-successful (or already-timed-out) tool call, which is exactly the regression class this PR's own rationale says it prevents ("a telemetry-only defect can never fail an otherwise-successful tool call").

No test in `test_control_telemetry.py` currently exercises `continue_processing` raising — worth adding alongside the existing `executions`-raising test once this is fixed.

Suggested fix: widen the same try/except added around the `executions` extraction to also cover this `getattr` call (or wrap both attribute reads in one guarded block).

## Non-blocking (carried over, still open)

`mcpgateway/plugins/control_telemetry.py:28` — module docstring's "Feature gate" section is now a single-bullet list after the CPEX-version bullet was removed; reads awkwardly as a list-of-one. Cosmetic only, no rush.

## Confirmed clean

No dangling `cpex_compat` / `execution_records_supported` / `get_executions` references anywhere in the repo; `pyproject.toml` confirms `cpex>=0.1.2` is the hard floor; no Alembic/schema change (correctly none needed — this is a backend-only chore); not a breaking change (internal, best-effort telemetry only); implementation matches PR description; no dead code; no unrelated changes; security posture unchanged.

msureshkumar88 (approved):
Ran a dedicated real-environment e2e verification against this PR head (`dbc9cc1e1`) rather than relying on the described unit-test verification alone — script: `scripts/verify-pr-6231-e2e.sh` (not committed, kept local as scratch tooling).

## What it drives
A real `uv run uvicorn mcpgateway.main:app` process (not TestClient), a real local upstream HTTP stub for the tool backend (plain `http.server`, not a monkeypatched transport), a real SQLite DB via the app's own startup lifespan, a real JWT minted via `mcpgateway.utils.create_jwt_token`, and the real installed `cpex==0.1.3` package with the real `cpex_pii_filter` plugin loaded from `plugins/config.yaml`.

**Config used:**
```
PLUGINS_ENABLED=true
PLUGINS_CONFIG_FILE=plugins/config.yaml
OBSERVABILITY_ENABLED=true
CPEX_CONTROL_TELEMETRY_ENABLED=true
AUTH_REQUIRED=true
DATABASE_URL=sqlite:///e2e_pr6231_test.db
```

**Commands run** (`bash scripts/verify-pr-6231-e2e.sh`, 10 checks):
1. `cpex` version + `import mcpgateway.plugins.cpex_compat` (expect ImportError) + repo-wide grep for dangling `cpex_compat`/`execution_records_supported`/`get_executions`
2. Direct probe: fed `ControlTelemetryAccumulator.add()` a `PluginResult`-shaped object whose `.executions`/`.continue_processing` **raise** — reproducing the exact bug class both review rounds flagged
3. `POST /tools` (register REST tool against the real upstream stub)
4. `POST /rpc` `tools/call` with a controlled `traceparent`, through the real PIIFilter pre/post hooks
5. `GET /observability/traces/{trace_id}` — asserts a real `cpex.control.summary` DB span landed
6. `pytest tests/e2e/test_otel_control_telemetry_e2e.py` (9 tests, pre-existing regression suite, untouched by this PR)
7. `pytest` targeted unit scope (`test_control_telemetry.py`, `test_record_control_telemetry.py`, `test_tool_service.py`, `test_tool_service_coverage.py`)
8. `ruff check` on both changed source files

**Output:**
```
== 0. Static checks against the real installed tree (no server needed) ==
PASS: cpex 0.1.3 installed (PR requires >=0.1.2)
PASS: mcpgateway.plugins.cpex_compat is gone (ImportError as expected)
PASS: no dangling references to cpex_compat / execution_records_supported / get_executions

== 1. Directly probe the fixed defensive guard against the REAL cpex library ==
PASS: ControlTelemetryAccumulator.add() survives a raising PluginResult (both fixed guards hold)

== 2-4. Real upstream stub + real gateway + tool registration ==
PASS: upstream stub up / gateway up / tool registered

== 5. Real tools/call through the real plugin chain (PIIFilter pre+post hooks) ==
PASS: tools/call succeeded (HTTP 200)

== 6. Verify cpex.control.summary DB span was recorded for this trace ==
PASS: cpex.control.summary span present -- telemetry pipeline intact post-removal

== 7. Existing regression suite ==
PASS: tests/e2e/test_otel_control_telemetry_e2e.py (9 tests) -- all pass on PR head

== 8. Targeted unit scope ==
PASS: targeted unit suite green

== 9. Lint ==
PASS: ruff clean

RESULT: ALL CHECKS PASSED
```

Also confirmed against the pre-PR baseline (`70cd7152b`) that the same 9-test regression suite passes there too — expected, since the real `cpex_pii_filter` plugin never exercises the raising-descriptor edge case; that edge case is only reachable via a malformed/non-conforming `PluginResult`, which is exactly what check 1 above exists to cover directly.

## Assessment

Removing `cpex_compat.py` is the right, minimal fix for #5996 — `cpex>=0.1.2` has been the declared floor since #5785, so the guard was unreachable dead code. Both blocking findings from earlier review rounds (lost `try/except` around `.executions`, then around `.continue_processing`) are genuinely fixed in `26bff444f`/`dbc9cc1e1`, not just described as fixed — verified by reading the diffs, by the new unit tests (`test_noop_when_executions_raises`, `test_noop_when_executions_not_iterable`, `test_noop_when_continue_processing_raises`), and now by the direct real-cpex probe above. No alembic/schema change needed (correctly none), backend-only, no unrelated files touched, no dangling references anywhere in the repo.

## Changes required

**Blocking:** none.

**Optional (non-blocking):** the two new `except Exception` blocks in `control_telemetry.py` (~line 116 `executions` extraction, ~line 133 `continue_processing` read) swallow silently with no `logger.debug(..., exc_info=True)`, unlike every other `except Exception` in this file (`aggregate()`, `_emit_db_spans()`, `_per_control_attributes()`). If a plugin's `PluginResult` ever starts raising in production, there'd be zero log trail to diagnose it. A one-line debug log in each block would close that gap — nice-to-have, not a merge blocker.

Approving based on the above.

Issue #5804, opened 2026-07-23
[EPIC][UI-REWRITE][SECURITY]: Session timeout, idle detection, and silent refresh for the React UI


a-effort:
## Goal

Add proactive session lifecycle management to the React UI: silent token refresh, activity-based idle detection with a warning dialog, one-click session continuation, and graceful degradation for SSO-issued sessions. Ship the backend refresh and validate endpoints that do not yet exist.

## Why now

Two things make this a real problem on the React UI, not a nice-to-have:

1. `token_expiry` default was reduced from 70 days to **20 minutes** (`config.py:383`) for security. Users will hit expiry during normal work.
2. The React 401 path is purely reactive. `client/src/api/client.ts:134` redirects to `/app/login` on any 401, and `AuthContext.tsx` does the same on a failed `/app/auth/me`. In-flight work is lost with no warning.

There is no `/api/auth/refresh` or `/api/auth/validate` endpoint on the backend today (the `refresh_token` references in `mcpgateway/routers/llmchat_router.py` and DCR flags are for downstream OAuth, not the admin session).

This epic replaces the legacy-UI-shaped #2561.

## Current state

**React UI (reactive only):**
```
User works past token_expiry (20 min)
  → next API call → 401
  → api/client.ts throws ApiError(401, "Session expired")
  → window.location.replace('/app/login')
  → in-flight state lost, unsaved work lost
```

**Backend:** JWT auth only, no refresh endpoint, no session-validate endpoint.

## User stories

**US-1: Silent refresh on 401.** A user actively working when the JWT expires should have the request transparently retried after a silent refresh, without a redirect.

**US-2: Idle warning before expiry.** A user idle for the configured warning threshold should see a dialog stating how long remains before their session expires, with a Continue working action.

**US-3: One-click continuation.** Clicking Continue working calls the refresh endpoint, updates the session cookie, closes the dialog, and resets the idle timer. Failure falls back to a graceful logout with a clear message.

**US-4: Activity-based idle timer.** Mouse, keyboard, scroll, touch, HTMX-equivalent (successful API responses), and route navigations reset the idle timer. Events are debounced so rapid input does not degrade performance.

**US-5: Configurable timeouts.** Admins configure idle timeout, warning window, refresh buffer, and can disable idle detection entirely for trusted environments.

[...]

PR #6235, opened 2026-08-13
feat(auth): add session refresh and validate endpoints

### LINKED ISSUE #6011: [FEATURE][API]: Session refresh & validate endpoints (POST /api/auth/refresh, GET /api/auth/validate)
Part of #5804 (Session timeout, idle detection, and silent refresh for the React UI). This extracts **Phase 1 (backend endpoints)** so the frontend session-lifecycle work is not blocked on prose.

There is no `/api/auth/refresh` or `/api/auth/validate` endpoint on the backend today. The `refresh_token` references in `mcpgateway/routers/llmchat_router.py` and the DCR flags are for downstream OAuth, not the admin session. With `token_expiry` reduced to 20 minutes (`config.py:383`), the React UI needs both to implement silent refresh and idle detection.

## Endpoints

- [ ] `POST /api/auth/refresh` (`mcpgateway/routers/auth.py`): validate the current token, issue a new JWT with extended expiry, set the session cookie, return `{expires_at, expires_in}`.
- [ ] `GET /api/auth/validate`: return `{valid, expires_at, expires_in, user, session_source}` where `session_source` is `local | sso | api_token`.
- [ ] Refuse refresh for `api_token`-sourced sessions (long-lived tokens do not need refresh).
- [ ] Route refresh events through `AuditTrailService.log_action()` (do **not** pass `db=db`, per repo convention).
- [ ] Per-user rate-limit the refresh endpoint to prevent refresh loops from a stuck client.

## Configuration (surface via `/api/auth/validate` so the client does not hardcode)

- [ ] `SESSION_IDLE_TIMEOUT` (default 900s), `SESSION_WARNING_TIME` (default 60s), `SESSION_ACTIVITY_TRACKING` (default `true`), `SESSION_REFRESH_BUFFER` (default 300s) in `mcpgateway/config.py`; update `.env.example`.

## Security

- [ ] Cookie-session auth + CSRF consistent with `/app/auth/me`.
- [ ] Deny-path tests: expired/invalid token cannot refresh; `api_token` session refused; rate-limit enforced; SSO source reported correctly.

## Out of scope (covered by later phases of #5804)

- React refresh interceptor, activity tracker, warning dialog, SSO-delegated refresh flow, docs page.

## Acceptance criteria

- [ ] `POST /api/auth/refresh` issues a new JWT and cookie for a valid `local` session and 401s an expired/invalid one.
- [ ] `GET /api/auth/validate` returns accurate `expires_in` and `session_source`.
- [ ] `api_token` sessions are refused by refresh.
- [ ] `make verify` passes with new deny-path coverage.

cc @jonpspri

madhu-mohan-jaishankar:
## 🔗 Related Issue

Closes #6011

---

## 📝 Summary

Adds session-lifecycle endpoints so UI clients can implement silent refresh and idle detection now that `TOKEN_EXPIRY` defaults to 20 minutes:

- **`POST /auth/refresh`** — issues a new session JWT for a valid session. Preserves the old token's scope narrowing, rotates the CSRF token (CSRF tokens are HMAC-bound to the JWT `jti`, which changes on refresh), and re-sets the `jwt_token` cookie when the request authenticated via cookie. Refuses API tokens and any token without `token_use="session"` with 403 (fail-closed) — long-lived API tokens have their own expiry/revocation lifecycle. Refresh events are audit-logged via `AuditTrailService`.
- **`GET /auth/validate`** — reports `valid`, `expires_at` / `expires_in`, the user profile, `session_source` (`local | sso | api_token`, derived from the `token_use` / `auth_provider` / `source` claims), and a `config` block of session-lifecycle values (normalized to seconds) so clients don't hardcode them.

Both endpoints live in `mcpgateway/routers/auth.py` and are exposed at `/v1/auth/*` (canonical) plus the unversioned legacy shim, same as the sibling `/auth/login` / `/auth/logout`. The issue's `/api/auth/*` paths don't exist in the codebase; the implemented paths match the actual v1/legacy mounts and the pre-existing manual test case (AUTH-004) and CSRF config references.

**Supporting changes:**

- `SESSION_MAX_LIFETIME` (default 480 min, `0` disables): absolute session-age cap enforced at refresh via a `session_start` claim carried across refreshes — silent refresh cannot extend a session forever. No `SESSION_IDLE_TIMEOUT` was added: server-side idle enforcement already exists as `TOKEN_IDLE_TIMEOUT` (revocation-backed, in `mcpgateway/auth.py`); `/auth/validate` surfaces it to clients instead of duplicating the knob.
- `SESSION_WARNING_TIME`, `SESSION_REFRESH_BUFFER`, `SESSION_ACTIVITY_TRACKING`: client-behavior hints surfaced via `/auth/validate`.
- **Rate limiting** via a new `SESSION_REFRESH` tier in `RateLimitMiddleware` (pattern covers both `/auth/refresh` and `/v1/auth/refresh`), configured by `SESSION_REFRESH_RATE_LIMIT` (default 10/min). Reuses the existing Redis-backed, multi-dimensional (IP / user / team) infrastructure rather than adding a bespoke limiter.
- **`/auth/refresh` removed from default `CSRF_EXEMPT_PATHS`**: the entry predated the endpoint; a cookie-authenticated, CSRF-exempt refresh would let a cross-site request silently keep a victim's session alive. Bearer requests skip CSRF by design, so API clients are unaffected.
- `create_access_token()` accepts an optional `extra_claims` dict (reserved claims cannot be overridden) for the `session_start` carry-over.
- Refresh is **single-use rotation**: the predecessor token is revoked (blocklisted with `reason=token_refresh` and its original expiry) before the new token is minted, with compare-and-set semantics so concurrent refreshes cannot both succeed; refresh fails closed (401) if the revocation cannot be persisted.

---

## 📏 Reviewability

---

## 🏷️ Type of Change

- [x] Feature / Enhancement

---

## 🧪 Verification

| Lint (changed files)      | `ruff check`, `isort --check-only`, `pylint` (9.96), `bandit` (0 issues), `interrogate` (100%) | ✅ |
| Unit tests (affected)     | `pytest tests/unit/mcpgateway/routers/ tests/unit/mcpgateway/middleware/test_rate_limit_middleware.py tests/unit/mcpgateway/middleware/test_csrf_middleware.py tests/unit/mcpgateway/test_config.py` — 1586 passed, 5 skipped (pre-existing) | ✅ |
| mypy (changed files)      | `mypy mcpgateway/routers/auth.py` — 4 errors, identical to pre-change baseline (pre-existing debt, none introduced) | ✅ |
| Full `make lint` / `make test` / `make coverage` | pending CI | ⏳ |

New test coverage (13 tests in `test_auth.py`, 1 in `test_rate_limit_middleware.py`), including the deny paths: API-token refresh refused (403), missing-`token_use` token refused (403), no session token (401), max-lifetime exceeded (401), `session_start` carried over on refresh, SSO / external-IdP / api_token `session_source` classification, cookie-vs-bearer cookie behavior, and `SESSION_REFRESH` tier matching on both mounts.

---

## ✅ Checklist

---

## 📓 Notes (optional)

- **Design decisions not specified by the issue:** SSO-established sessions may refresh locally in this phase (the delegated SSO refresh flow is a later phase of #5804); unauthenticated `/auth/validate` returns 401 (consistent with the auth dependency) rather than `{valid: false}`; enforced durations follow the repo's minutes convention (`TOKEN_EXPIRY`-style) while client hints are in seconds — `/auth/validate` normalizes everything to seconds.
- `tests/unit/mcpgateway/middleware/test_rate_limit_middleware.py` includes isort normalization of pre-existing import-heading debt (comment-only hunks) — the file was non-compliant before this change; touching it triggered the standard formatter cleanup.
- Deployments that set `CSRF_EXEMPT_PATHS` explicitly in `.env` should drop `/auth/refresh` from their list to pick up the secure default.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -116,7 +116,7 @@ CSRF_TRUSTED_ORIGINS=["http://localhost:4444","http://localhost:8080"]
 # WARNING: Setting this env var REPLACES the config.py default entirely — it does not append.
 # Leave unset to use the secure default maintained in mcpgateway/config.py (csrf_exempt_paths).
 # Only set explicitly if you need a fully custom list (you must include every path you want exempt).
-#CSRF_EXEMPT_PATHS=["/health","/auth/login","/auth/logout","/auth/refresh","/auth/email/login","/auth/email/register","/auth/email/forgot-password","/auth/email/reset-password","/admin","/admin/login","/admin/forgot-password","/admin/reset-password","/oauth/fetch-tools","/docs","/redoc","/openapi.json","/metrics","/mcp/","/sse","/message","/rpc","/api/metrics/","/toolops/","/tokens","/teams/","/llmchat/","/api/logs/","/_internal/mcp/"]
+#CSRF_EXEMPT_PATHS=["/health","/auth/login","/auth/logout","/auth/email/login","/auth/email/register","/auth/email/forgot-password","/auth/email/reset-password","/admin","/admin/login","/admin/forgot-password","/admin/reset-password","/oauth/fetch-tools","/docs","/redoc","/openapi.json","/metrics","/mcp/","/sse","/message","/rpc","/api/metrics/","/toolops/","/tokens","/teams/","/llmchat/","/api/logs/","/_internal/mcp/"]
 
 # Bootstrap admin credentials (email auth)
 # PRODUCTION: Change these values
@@ -1113,6 +1113,18 @@ VALIDATE_TOKEN_ENVIRONMENT=true
 
 # Expiry time for generated JWT tokens (in minutes; e.g. 7 days)
 # TOKEN_EXPIRY=10080
+
+# Session lifecycle (POST /auth/refresh + GET /auth/validate)
+# Absolute maximum session lifetime in minutes, enforced at refresh (0 disables the cap)
+# SESSION_MAX_LIFETIME=480
+
+# Maximum POST /auth/refresh requests per minute per client dimension (IP/user/team)
+# SESSION_REFRESH_RATE_LIMIT=10
+
+# Client-behavior hints surfaced via GET /auth/validate (seconds)
+# SESSION_WARNING_TIME=60
+# SESSION_REFRESH_BUFFER=300
+# SESSION_ACTIVITY_TRACKING=true
 
 # SECURITY: Require expiration claim in all tokens (default: true)
 # Set to false only for backward compatibility with legacy tokens
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^\\.secrets\\.baseline$",
     "lines": null
   },
-  "generated_at": "2026-08-21T10:23:38Z",
+  "generated_at": "2026-08-21T12:35:45Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -132,7 +132,7 @@
         "hashed_secret": "fa9beb99e4029ad5a6615399e7bbae21356086b3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1370,
+        "line_number": 1382,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -140,7 +140,7 @@
         "hashed_secret": "d08f88df745fa7950b104e4a707a31cfce7b5841",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1405,
+        "line_number": 1417,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -148,7 +148,7 @@
         "hashed_secret": "ac371b6dcce28a86c90d12bc57d946a800eebf17",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1494,
+        "line_number": 1506,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -156,7 +156,7 @@
         "hashed_secret": "0b6ec68df700dec4dcd64babd0eda1edccddace1",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1499,
+        "line_number": 1511,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -164,7 +164,7 @@
         "hashed_secret": "4ad6f0082ee224001beb3ca5c3e81c8ceea5ed86",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1504,
+        "line_number": 1516,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -172,7 +172,7 @@
         "hashed_secret": "cb32747fcfb55eaa194c8cd8e4ba7d49ada08a94",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1510,
+        "line_number": 1522,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -180,7 +180,7 @@
         "hashed_secret": "6c178d51b13520496dbc767ed3d9d7aa5803ac72",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1522,
+        "line_number": 1534,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -188,7 +188,7 @@
         "hashed_secret": "ca45060a53fd8a255d1a83ee8d2f025283ccc66e",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1540,
+        "line_number": 1552,
         "type": "Secret Keyword",
         "verified_result": null
       },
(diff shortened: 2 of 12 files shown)


a-effort:
LGTM! ✅

Lang-Akshay (changes requested):
## Blocking Changes

| 1 | Security | `mcpgateway/routers/auth.py` | 385-396, 479-507 | **High / CWE-347**: refresh and validation decode the request token with signature verification disabled. The security handoff found that `get_current_user` can involve a plugin/custom auth path, so the handler may trust forged `token_use`, `session_start`, `scopes`, and expiry claims. | Use the exact payload returned by a verified JWT path, or reject refresh/validate for non-gateway-JWT auth methods. For cookie auth, verify the selected cookie token and bind its subject/JTI to the authenticated user before making session decisions. |
| 2 | Issue / Quality | `mcpgateway/config.py`, `.env.example` | `config.py:385-403`, `.env.example:1090-1105` | Linked issue #6011 explicitly requires `SESSION_IDLE_TIMEOUT` with a 900-second default. The PR instead exposes the pre-existing `TOKEN_IDLE_TIMEOUT` (60-minute default) and does not add the requested setting. | Add and surface `SESSION_IDLE_TIMEOUT` with the issue's required default/units, or obtain an explicit issue-level decision changing that acceptance criterion. |
| 3 | Issue / Verification | PR metadata | N/A | Issue #6011 requires `make verify`; the PR description marks full verification as pending CI, so this required acceptance criterion is unresolved in the reviewed state. | Run and pass the required verification in CI before merge, and attach the result to the PR. |

madhu-mohan-jaishankar:
Thanks for the thorough review @Lang-Akshay — the security catch in particular was a real gap. Here's where each finding stands:

**1. Unverified token decode (CWE-347)** — Fixed in a052f0b02. Both endpoints now route the extracted token through `verify_jwt_token_cached()` (full signature/expiry/audience verification) before any session decision, and additionally check the `jti` against the revocation blocklist and bind the token's `sub` to the authenticated user. A request that authenticated through a non-gateway-JWT path (plugin/basic/proxy) gets 401 from both endpoints, since there is no gateway session to refresh or validate. New deny-path tests cover forged signature, revoked jti, and subject mismatch on refresh, plus forged/missing token on validate. The normal bearer path does no extra crypto work — `verify_jwt_token_cached` is a per-request cache hit for the token the auth dependency already verified.

**2. `SESSION_IDLE_TIMEOUT`** — This one was a deliberate deviation rather than an oversight, and worth discussing. Server-side idle enforcement already exists as `TOKEN_IDLE_TIMEOUT` (revocation-backed, in `mcpgateway/auth.py`), and `GET /auth/validate` surfaces it to clients as `config.idle_timeout` in seconds — which covers the linked issue's underlying goal (clients shouldn't hardcode the value). Adding a second setting named `SESSION_IDLE_TIMEOUT` would give operators two knobs that appear to control the same behavior while only one is enforced. The deviation is documented in the PR description under "Supporting changes". If there's a strong preference for the literal acceptance criterion I'm happy to revisit, but two sources of truth for one enforcement point seemed like the worse trade.

**3. Verification** — `make verify` passes locally (it's the packaging gate: twine / check-manifest / pyroma). The CI failures on the previous push are also fixed in a052f0b02: the pre-commit failure was a `ruff format` line-wrap in `config.py`, and the four pytest failures were tests in `test_auth_logout.py` / `test_csrf_fixes.py` whose mock patch targets pointed at import paths that moved to module level — they now patch `mcpgateway.routers.auth.*` directly. Full lint/test/coverage results will be on this push's CI run.

Lang-Akshay (changes requested):
## Blocking Changes

| 1 | Quality / Issue | `mcpgateway/routers/auth.py` | 478-492, 511-531 | Both endpoints depend on `get_current_user()`, while the configured bearer dependency reads only the auth header. Cookie extraction in `_extract_raw_token()` is reached only after dependency authentication; a cookie-only request therefore fails before refresh/validate logic runs. This contradicts the PR's cookie-session requirement and makes the `from_cookie` cookie-reset branch unreachable in the normal route path. | Make the route dependency authenticate the supported session cookie (preferably via the existing centralized verifier and user resolver), or use an auth dependency that accepts the cookie and preserves the same subject/revocation checks. Add an end-to-end cookie-only smoke path. |

madhu-mohan-jaishankar:
Good catch again @Lang-Akshay — confirmed and fixed in 52d405291.

Lang-Akshay (changes requested):
## Blocking Changes
| 1 | Security | `mcpgateway/routers/auth.py` | 576–579 | Refresh generates `new_jti` but only records `old_jti` for audit; it never revokes the predecessor. The old bearer/cookie token remains valid and can be replayed to mint more tokens, defeating rotation and logout/revocation expectations (High, CWE-613). | Atomically revoke/blocklist `old_jti` using its original expiry and a refresh reason before returning the new token. Make rotation single-use/compare-and-set so concurrent refreshes cannot both succeed; fail closed if predecessor revocation cannot be persisted. |

Lang-Akshay (dismissed):
LGTM ✅

Lang-Akshay (approved):
LGTM ✅

PR #6245, opened 2026-08-14
fix(search): include opt-in catalog results

### LINKED ISSUE #6212: Unified search doesn't include the MCP server catalog
## Summary

`GET /v1/search` never returns MCP registry/catalog entries, even when the query text matches a catalog server by name. For example, searching "cloudflare" returns zero results across every entity type despite "Cloudflare Docs" existing and being visible on the `/mcp-registry/servers` (catalog) page.

## Root cause

`perform_unified_search()` in `mcpgateway/admin.py` (~line 11558) only searches `servers, gateways, tools, resources, prompts, agents, teams, users, roots` — its `supported_entity_types` list (line 11608) has no `catalog` entry, and it never calls `catalog_service.get_catalog_servers()`. The catalog is a separate, YAML-backed registry (`services/catalog_service.py`) not covered by any of the existing `admin_search_*` functions.

**Scope note:** this function also backs `/admin/search` (the legacy server-rendered admin dashboard). That surface isn't getting further investment — `/v1/*` is the API going forward — so this fix is deliberately scoped to avoid touching the old dashboard's behavior at all, not just to avoid extra work there now.

## Proposed fix

Add `"catalog"` to `supported_entity_types` only:

```python
supported_entity_types = ["servers", "gateways", "tools", "resources", "prompts", "agents", "teams", "users", "roots", "catalog"]
# default_entity_types intentionally left unchanged — catalog is opt-in via
# explicit entity_types=..., so /admin/search (legacy dashboard, not being
# extended further) and any other caller that doesn't ask for it explicitly
# is unaffected.
default_entity_types = ["servers", "gateways", "tools", "resources", "prompts", "agents", "teams", "roots"]
```

Add a block alongside the existing `teams`/`users`/`roots` handling (which also require real query text), filtered to `auth_type="Open"` to match the catalog page's own MVP display scope (`ServerCatalog.tsx`'s `OPEN_AUTH_TYPE` filter) — confirmed `catalog_service` does an exact-match filter on `auth_type` (`services/catalog_service.py:211`), so this keeps search results and the catalog page consistent (nothing surfaces in search that then can't be seen after clicking through):

```python
if "catalog" in selected_entity_types and search_query and settings.mcpgateway_catalog_enabled:
    catalog_request = CatalogListRequest(search=search_query, auth_type="Open", limit=effective_limit)
    catalog_response = await catalog_service.get_catalog_servers(
        catalog_request, db,
        user_email=get_user_email(user),
        token_teams=user["_cached_team_ids"],
    )
    grouped_results["catalog"] = [
        {"id": s.id, "name": s.name, "description": s.description}
        for s in catalog_response.servers
    ]
```

(`CatalogListRequest`/`catalog_service` are already imported in `admin.py`.)

- Guard on `settings.mcpgateway_catalog_enabled` so a disabled catalog just yields an empty group rather than a 404 that would fail the whole search (mirrors how `/mcp-registry/servers` itself handles the flag).
- Update the `entity_types` docstring in `routers/search.py` (the `/v1/search` router — the API surface this fix actually targets) to list `catalog` as supported.

## Companion change

contextforge-org/contextforge-web-ui#20 — frontend wiring to request/route/label the new `catalog` entity type in the header search.

vishu-bh:
# Pull Request

## 🔗 Related Issue

Closes #6212

---

## 📝 Summary

Unified search now supports MCP catalog entries when callers explicitly request `entity_types=catalog`.

The shared search core previously rejected `catalog` as unsupported and never queried the YAML-backed catalog service. This change adds a small, permission-aware catalog search helper that reuses existing catalog and unified-search response infrastructure.

Catalog search:

- remains opt-in, leaving default and legacy dashboard searches unchanged;
- requires `servers.read`;
- returns only `Open` authentication entries shown by the current catalog UI;
- preserves caller team scope when resolving registration visibility;
- returns an empty group when the catalog feature is disabled or access is denied.

Unit and HTTP router tests cover result mapping, explicit selection, default exclusion, disabled behavior, team scope, and permission-denial isolation.

---

## 📏 Reviewability

---

## 🏷️ Type of Change

- [x] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

| Check | Command | Status |
|---|---|---|
| Pre-commit hooks | `make pre-commit` | ✅ Passed |
| Doctests | `make doctest` | ✅ 1,219 passed, 56 skipped |
| Unit tests | `make test` | ✅ Passed |
| Package verification | `make verify` | ✅ Passed |
| Coverage ≥ 80% | `make coverage` | Not run |

---

## ✅ Checklist

- [x] Code formatted (`make pre-commit`)
- [x] Documentation updated (supported entity types)

---

## 📓 Notes (optional)

Catalog is deliberately not added to `default_entity_types`. Callers must request it explicitly, which keeps `/admin/search` and existing clients unchanged.


gandhipratik203 (dismissed):
**Team scoping diverges from the catalog router (suggestion, narrow)**

`admin_search_catalog` uses `_get_user_team_ids()`, but `routers/catalog.py:78` uses `get_scoped_resource_access_context()`, which signals admin bypass with `None`. That drops other teams' private gateways from `registered_urls`, and with `show_available_only` defaulting to `True`, an entry marked unavailable in YAML can vanish from search while still showing on `/mcp-registry/servers`.

Not reachable with the shipped catalog (no entry sets `is_available`), so it's a consistency fix. Either use `get_scoped_resource_access_context`, or pass `show_available_only=False`.

Lang-Akshay (dismissed):
LGTM ✅

gandhipratik203 (dismissed):
LGTM!

PR #6247, opened 2026-08-14
Security Hardening: Catalog Registration Scope Enforcement

### LINKED ISSUE #6036: [BUG]: catalog registration creates gateways without caller identity (owner_email=None, created_by="system")

Lang-Akshay:
Security harden catalog server registration to enforce caller-owned scope and least-privilege defaults across all registration paths. All changes are backend-only; no UI files are modified.

### Catalog registration scope
- Registrations default to `visibility="private"` with `owner_email` derived from the authenticated caller — no forced-public, ownerless gateways
- Token scope validated before gateway creation; foreign teams and public-only token misuse rejected with 403
- `ALLOW_PUBLIC_VISIBILITY=false` now enforced on the catalog path (`_resolve_registration_scope`); previously only admin UI handlers checked this flag
- Two serial team-membership DB queries in `_resolve_registration_scope` replaced with a single JOIN, matching `TeamManagementService` semantics and eliminating a redundant round-trip
- Bulk registration validates scope once upfront and passes the resolved `(visibility, team_id)` tuple into each per-server call via `_resolved_scope`, removing N redundant DB queries from the loop

### Gateway ownership transfer
- New `POST /admin/gateways/{id}/transfer-ownership` endpoint (admin-only) validates target user status and team membership, propagates ownership to linked tools/resources/prompts, and writes an audit trail
- Transferring a private gateway to a team coerces `visibility` to `"team"` on the gateway and all linked entities — without this, team members could not access the gateway after transfer

### Safe user deletion
- Before permanent deletion, all owned gateways are transferred atomically: deletion is refused if any transfer cannot be completed
- Fallback owner selection now joins `EmailUser` to exclude deactivated accounts (previously only `EmailTeamMember.is_active` was checked)
- Orphan-owner `ValueError` surfaces as HTTP 409 with the actionable message on `DELETE /admin/users/{email}`; previously swallowed into a generic 500

## Tests added
- `test_gateway_service_transfer.py` — full service-layer transfer coverage (target not found, not a member, success, linked entity propagation, private→team visibility coercion)
- Deny-path regression tests: public-only token, foreign team, `ALLOW_PUBLIC_VISIBILITY=false`, non-admin on transfer endpoint
- Orphan 409 regression test on the API router path

## No schema or migration changes
No new columns or tables; no Alembic migration required.

**Internal reference:** https://github.ibm.com/contextforge-org/internal_issues/issues/497

Closes #6036


msureshkumar88 (changes requested):
Thanks for tackling this — the shift to private-by-default with explicit ownership tracking is a solid direction, and the ownership-transfer-on-deletion flow with the atomic refuse-if-no-owner fallback is a genuinely good safety net. Found a few gaps that are worth addressing before merge, plus some smaller notes.

## Blocking

**1. `_resolve_registration_scope()` doesn't check `allow_public_visibility` (`mcpgateway/services/catalog_service.py:62`)**
Every other gateway-creation path in `admin.py` calls `_check_public_visibility_allowed()` before allowing `visibility="public"`. The new catalog registration path never references `allow_public_visibility` at all, so with `ALLOW_PUBLIC_VISIBILITY=false`, registering a catalog server with `visibility="public"` still succeeds. Since this PR's whole purpose is scope enforcement, this specific control should be wired in here too (CWE-284/CWE-863).

**2. `transfer_gateway_ownership()` updates `team_id` but not `visibility` (`mcpgateway/services/gateway_service.py:5589`)**
Transferring a `public`/`private` gateway to a team sets `gateway.team_id` but leaves `visibility` unchanged. The actual access-scoping query (`gateway_service.py:2538-2556`) only grants team access when `visibility` is `"team"`/`"public"` — it doesn't derive scope from `team_id` alone. So the transfer silently doesn't do what it says: the gateway's effective access scope is unchanged, no error is raised, and the existing test (`test_gateway_service_transfer.py:1680-1729`) only asserts `team_id` changed, not `visibility`, so this ships uncaught. (Confirmed this isn't a privilege-escalation path — private gateways stay gated by `owner_email` regardless — but it does defeat the endpoint's stated purpose.)

**3. Deactivated users can inherit gateway ownership (`mcpgateway/services/email_auth_service.py:2109`)**
The alternate-owner lookup in `delete_user()` filters on `EmailTeamMember.is_active` only, never `EmailUser.is_active`. Since deactivating a user (`update_user(is_active=False)`) doesn't touch their `EmailTeamMember` rows, a disabled account can still be picked as the new owner of a deleted teammate's gateway — which undercuts the PR's goal of preventing unmanageable/orphaned gateways, since nobody can log into a disabled account. The explicit `transfer_gateway_ownership()` endpoint already does check `is_active` for the target — worth aligning the two.

**4. Orphan-owner error gets swallowed into a generic 500 (`mcpgateway/routers/email_auth.py:913`)**
The new `ValueError("...would become orphaned and no fallback owner is available")` is a good, actionable error, but the router's `except Exception` catches it and returns a generic "Failed to delete user" 500 — the admin never sees the real reason (configure `platform_admin_email` or reassign manually). The admin-UI equivalent (`admin.py` ~8669-8671) does surface `str(e)`; worth matching that on the API path.

## Suggestions

- `_resolve_registration_scope()` reimplements team-membership validation instead of reusing `TeamManagementService.get_user_role()`/`verify_team_for_user()`, which `admin_add_gateway` already uses for the same check. Consolidating would prevent the two paths from drifting apart (this PR is itself an example of that drift — see #1).
- `bulk_register_servers()` validates the batch's scope once up front, then `register_catalog_server()` re-runs the identical validation (2 queries + a commit) per item in the loop — for a 50-item batch that's 50 redundant round-trips for validation already proven valid.
- `transfer_gateway_ownership()` runs two near-identical team-membership queries when `visibility=="team"` and an explicit `target_team_id` is passed — the second appears to re-check what the first already validated.
- No deny-path test for a non-admin calling `POST /admin/gateways/{id}/transfer-ownership` — worth adding given this endpoint can reassign any gateway system-wide.
- Visibility/`team_id` extraction logic in the registration modal is duplicated between `modals.js` (`submitApiKeyForm`) and the inline script in `mcp_registry_partial.html` — could drift if only one gets updated later.

## Minor

- The PR references only an internal `github.ibm.com` tracker issue; no public issue / `Closes #NNN` link, which makes this harder to trace on the public repo.

No Alembic migration is touched, which looks correct — no schema change is actually needed for this fix. Scope of touched files otherwise looks tight and on-topic.

Lang-Akshay (commented):
Thanks for the thorough review @msureshkumar88 — all blocking findings and suggestions have been addressed in the latest commit.

**Blocking**

1. **`_resolve_registration_scope()` missing `allow_public_visibility` check** — Fixed ✅. Added the guard immediately after `visibility` is resolved: `visibility="public"` with a `team_id` now raises `CatalogRegistrationPermissionError` when `ALLOW_PUBLIC_VISIBILITY=false`, matching every other creation path in `admin.py`. Test added: `test_resolve_scope_blocks_public_when_flag_disabled`.

2. **`transfer_gateway_ownership()` leaves `visibility` unchanged** — Fixed ✅. When `target_team_id` is supplied and the gateway (or any linked tool/resource/prompt) has `visibility="private"`, it is coerced to `"team"` so the team can actually see it after transfer. Test added: `test_transfer_private_gateway_to_team_coerces_visibility`.

3. **Deactivated users can inherit gateway ownership** — Fixed ✅. The alternate-owner query now JOINs `EmailUser` and filters `EmailUser.is_active == True` alongside the existing `EmailTeamMember.is_active` check. Applied to both the gateway fallback path and the team-ownership transfer path in `delete_user()`.

4. **Orphan-owner error swallowed into 500** — Fixed ✅. Added a dedicated `except ValueError` branch before the generic `except Exception` in `DELETE /admin/users/{email}`, returning HTTP 409 with `str(e)` as the detail. The admin now sees the actionable message ("configure `platform_admin_email` or reassign manually"). Test added: `test_delete_user_orphan_value_error_returns_409`.

**Suggestions**

- **Duplicate team-membership validation** — Consolidated. The two serial raw queries in `_resolve_registration_scope` (team existence + membership) are replaced by a single JOIN, mirroring the pattern used by `TeamManagementService.verify_team_for_user()`. Making the method `async` to call it directly would cascade through two callers; the JOIN achieves the same consolidation without that change.

- **Redundant per-item scope validation in bulk** — Fixed. `_resolve_registration_scope` runs once in `bulk_register_servers` and the resolved `(visibility, team_id)` tuple is forwarded via a new `_resolved_scope` kwarg on `register_catalog_server`, skipping re-validation for every item in the batch.

- **Two near-identical team-membership queries in `transfer_gateway_ownership()`** — The second query (lines 5576–5584) covers the case where `visibility != "team"` but an explicit `target_team_id` is provided, which the first query does not evaluate. They guard different conditions so the duplication is load-bearing; left as-is to avoid silently dropping the cross-team membership check.

- **No deny-path test for non-admin on the transfer endpoint** — Added: `test_transfer_gateway_ownership_non_admin_denied` asserts HTTP 403 when `check_admin_permission` returns `False`.

- **`submitApiKeyForm` / `mcp_registry_partial.html` duplication** — The UI changes in this PR have been reverted; this is now out of scope for this branch.

gandhipratik203 (dismissed):
I had overlapping comments to Suresh. LGTM once the blocking ones are resolved!

gandhipratik203 (dismissed):
LGTM!

msureshkumar88 (changes requested):
Thanks for this — the core of the change is solid work. Deriving registration scope from the authenticated caller instead of hardcoding `visibility="public"` closes a real multi-tenancy hole, and the deny-path test coverage is genuinely thorough: unauthenticated identity, foreign team, public-only tokens, insufficient permissions, and the `ALLOW_PUBLIC_VISIBILITY=false` path all have dedicated tests. Collapsing the two serial membership queries into a single fail-closed JOIN in `_resolve_registration_scope` was the right response to earlier feedback, and I'd keep it as-is. Also good call not adding a migration — `visibility`/`team_id`/`owner_email` already exist on all four entities, so the two new optional Pydantic fields need no schema change.

A few things I'd like to sort out before this merges.

## Blocking

**1. The platform-admin fallback in `delete_user()` doesn't validate the target owner**

`mcpgateway/services/email_auth_service.py:2130-2137`:

```python
# Fallback: platform admin
if not new_owner_email:
    admin_email = settings.platform_admin_email
    if admin_email and admin_email.strip() and admin_email.strip() != email:
        new_owner_email = admin_email.strip()
```

A non-empty config string is enough to satisfy this, which means the orphan guard on the next line never fires. That's inconsistent with the two sibling paths in this same PR:

- the alternate-team-member branch directly above (2112-2126) joins `EmailUser` and filters `EmailUser.is_active == True`
- `gateway_service.transfer_gateway_ownership` (5667-5670) validates and raises `Target user not found or inactive: ...`

There's no FK backing `owner_email` (`db.py:4783` — it's a plain `String(255)`, same for tool/resource/prompt), so nothing downstream catches a bad value; the write just succeeds. Two ways this bites in practice: `PLATFORM_ADMIN_EMAIL` can be deactivated by a peer admin (`email_auth_service.py:1846-1854` only blocks self-deactivation and last-active-admin), and the default `admin@example.com` is only reconciled to a row at bootstrap, so changing the env var later leaves a string with no matching user. In either case the user is deleted and the gateway plus its linked tools/resources/prompts end up owned by a principal that Layer-1 owner matching will never match — effectively unmanageable, which is the opposite of the "deletion is refused if any transfer cannot be completed" guarantee in the PR description.

Mirroring the team branch should be enough (`EmailUser` and `select` are already imported here):

```python
if not new_owner_email:
    admin_email = (settings.platform_admin_email or "").strip()
    if admin_email and admin_email != email:
        admin_user = self.db.execute(
            select(EmailUser).where(
                EmailUser.email == admin_email,
                EmailUser.is_active == True,  # noqa: E712
            )
        ).scalar_one_or_none()
        if admin_user:
            new_owner_email = admin_user.email
```

The existing `raise ValueError(...)` then refuses the deletion correctly.

Related: `test_delete_user_transfers_gateway_to_platform_admin_fallback` (`tests/unit/mcpgateway/services/test_email_auth_basic.py:3058`) asserts the unvalidated string *is* accepted, so it currently locks in the behaviour rather than catching it. `test_delete_user_raises_if_gateway_would_be_orphaned` covers `platform_admin_email = None` (unset) but not set-but-invalid. Worth adding a case where the configured admin has no row or `is_active=False` and the deletion is refused.

**2. No public issue reference**

The body links `github.ibm.com/.../issues/497`, which isn't reachable from this repo, and there's no `Closes #N` in the description or in any of the 15 commits. [#6036](https://github.com/IBM/mcp-context-forge/issues/6036) looks like the public match — it names `catalog_service.py`, `routers/catalog.py` and `admin.py` as the touch points, which is exactly the core of this change. Could you add `Closes #6036`?

Slightly bigger question while we're there: the transfer-ownership endpoint, the auto-transfer on user deletion, the 409 remap and the `ALLOW_PUBLIC_VISIBILITY` enforcement all go beyond what #6036 asks for, and I couldn't find a public issue covering them. They're reasonable hardening and clearly adjacent, so I'm not pushing to split them out — but a short paragraph in the description explaining why they're bundled would help reviewers and whoever reads this in six months. Your call on whether a separate issue is worth filing.

**3. Docs, and the default-visibility behaviour change**

The diff is 16 files with no `docs/` and no `CHANGELOG.md`. Two gaps:

- `docs/docs/manage/catalog.md` doesn't mention `visibility` or `team_id` anywhere, and its "Registering a Catalog Server (API)" section still describes the old behaviour. Catalog registrations now default to `private` where they were previously hardcoded `public` — that's a behavioural break for anyone whose automation assumes catalog-registered gateways are platform-visible, and there's no migration for already-registered rows. Worth calling out explicitly in the docs and in release notes.
- `POST /admin/gateways/{id}/transfer-ownership` has no documentation anywhere in `docs/` — it only appears in `admin.py`, `token_scoping.py` and one test.

A one-time admin script (or at least a documented query) to review pre-existing catalog gateways might be worth considering, though I'd treat that as optional.

## Suggestions

**4. `registration_teams` is now orphaned server-side work** — `admin.py:18242-18254`. `39491e928` added both this fetch and the template block that consumed it; `eb3eded60` reverted the template and `modals.js` halves but left the fetch behind. `mcp_registry_partial.html` has no reference to it (no `team` or `visibility` occurrence at all), and nothing in `static/` or `admin_ui/` uses it — so every `/admin/mcp-registry/partial` load now pays a `get_user_teams()` call that's discarded. Either drop `18242-18254` plus the context entry at 18265 and the test at `test_admin.py:26880-26913`, or restore the template side if the revert was meant to be temporary.

**5. Duplicate membership query in `transfer_gateway_ownership`** — `gateway_service.py:5688` and `5700` issue byte-identical SQL whenever `gateway.visibility == "team"` and `target_team_id` is set, since line 5686 makes `effective_team_id` the same value. Both branches are needed in the general case, so guarding the second on `target_team_id != checked_team_id` would be the smallest fix. Worth noting no test currently sets `visibility="team"` together with `target_team_id`, so that path is uncovered — and since these tests drive `mock_db.execute.side_effect` lists that are sensitive to call count, adding a case alongside the change would be a good idea.

**6. The private→team coercion is repeated four times** — `gateway_service.py:5710-5743`, identical across gateway/tool/resource/prompt apart from one comment. A small local closure would make it harder to add a fifth entity type later and update only three of the blocks.

## Minor

**7.** The `ALLOW_PUBLIC_VISIBILITY` check in `catalog_service.py:96` uses slightly different truthiness than the existing one in `admin.py:1982` (`team_id and team_id.strip()`). They agree today because of how each call site normalises its input, but expressing the same policy in two places invites drift.

**8.** The new `except ValueError -> 409` in `routers/email_auth.py` also catches the pre-existing "user not found" case. That's an improvement over the previous behaviour (it was falling through to a 500), so nothing's regressed — but if you want it exact, `update_user` at `email_auth.py:815-819` already has the pattern for splitting not-found into a 404.

Lang-Akshay:
## Review findings assessment

Implemented all valid findings; skipped only the optional migration/script suggestion.

| # | Finding | Classification | Addressed? | Update |
|---:|---|---|---|---|
| 1 | Platform-admin fallback accepted an invalid owner | Valid | Yes, previously | Active-user validation and orphan rejection remain in place with regression coverage. |
| 2 | Missing public issue reference | Valid process finding | Yes | Added `Closes #6036` to this PR. |
| 2b | Bundled transfer/deletion work lacked rationale | Valid, non-blocking | Yes | Existing PR sections explain gateway transfer, safe deletion, visibility coercion, and 409 behavior. |
| 3 | Missing visibility and transfer documentation | Valid when raised | Yes | `docs/docs/manage/catalog.md` documents private defaults, team scope, existing-row behavior, and transfer ownership. |
| 3b | Missing release-note entry | Valid | Yes | Added an `[Unreleased]` catalog hardening entry to `CHANGELOG.md`. |
| 3c | Migration/admin review script | Out of scope | No | Existing rows are not rewritten; documentation advises operators to review them. |
| 4 | Orphaned `registration_teams` fetch | Valid | Yes | Removed the unused team query, context field, and obsolete test. |
| 5 | Duplicate transfer membership query | Valid, low severity | Yes | The second query now runs only for non-team gateways; added one-query regression coverage. |
| 6 | Repeated private-to-team coercion | Valid maintainability finding | Yes | Added `_apply_team_transfer()` for gateway, tool, resource, and prompt updates. |
| 7 | Inconsistent `team_id` truthiness | Valid minor consistency issue | Yes | Catalog registration now strips whitespace and treats blank team IDs as absent; added regression coverage. |
| 8 | Missing user mapped to HTTP 409 | Valid API semantics finding | Yes | Missing users now return 404; ownership/deletion conflicts remain 409; added regression coverage. |

msureshkumar88 (changes requested):
Re-reviewed against the actual commits (not just the fix summary) — all four blocking findings and the five suggestions from the last round check out. Verified each one directly:

- `_resolve_registration_scope` now enforces `ALLOW_PUBLIC_VISIBILITY`, matching `_check_public_visibility_allowed`'s existing semantics — `test_resolve_scope_blocks_public_when_flag_disabled` covers it.
- `transfer_gateway_ownership` coerces private→team visibility on the gateway and every linked tool/resource/prompt — `test_transfer_private_gateway_to_team_coerces_visibility` asserts all four.
- The user-deletion fallback now joins `EmailUser` and filters `is_active` on both the team-owner and platform-admin fallback paths, so a disabled account can no longer inherit ownership.
- The orphan `ValueError` now maps to 404/409 instead of a generic 500 in the router, with tests for both branches.

Good shape overall — nice test coverage on the deny paths in particular. One blocking item and a few optional notes before merge:

## Blocking

**Stray planning-doc file included in the diff** — `docs/superpowers/plans/2026-08-25-catalog-ownership-fallback.md` (160 lines). `docs/superpowers/` doesn't exist anywhere on `main`; this looks like an AI-agent implementation-plan artifact that got committed alongside the real changes rather than user-facing documentation. Worth dropping from the PR.

## Suggestions

- The admin catalog endpoints (`/admin/mcp-registry/{id}/register`, `/admin/mcp-registry/bulk-register`) now also require `gateways.create` in addition to `servers.create` (bringing them to parity with the v1 route). It's deliberate and tested (`test_admin_register_catalog_requires_gateways_create`), but isn't mentioned in the PR description or CHANGELOG — worth a line, since a custom role with only `servers.create` will start seeing 403s here.
- `transfer_gateway_ownership` reassigns every linked tool/resource/prompt's `owner_email` unconditionally, whereas the safe-deletion path added in this same PR guards with `if tool.owner_email == email:` before reassigning. If a linked entity ever has an owner different from its gateway, transfer would silently take it too. Might be worth the same guard for consistency (or a comment on why the two paths intentionally differ) — currently untested either way.
- `_resolve_registration_scope` calls `db.commit()` right after a read-only membership `SELECT` — doesn't look load-bearing, could probably be dropped.
- The active-team-membership JOIN (team/user active-status check) is now hand-rolled in three places (`catalog_service`, `gateway_service.transfer_gateway_ownership`, `email_auth_service.delete_user`). Not blocking, but a shared helper would avoid a fourth copy drifting next time this policy changes.

No schema/Alembic changes — correct, none needed. Scope stays backend-only as described; the earlier UI changes look properly reverted out of this branch.

Lang-Akshay:
Addressed in `b836d83cb`. Removed `docs/superpowers/plans/2026-08-25-catalog-ownership-fallback.md` from the PR. It was an internal implementation-plan artifact rather than user-facing documentation; the relevant catalog behavior remains documented in `docs/docs/manage/catalog.md`. This resolves the finding.

msureshkumar88 (dismissed):
Re-checked after the latest commit. The one remaining blocking item — the stray `docs/superpowers/plans/2026-08-25-catalog-ownership-fallback.md` planning artifact — is gone from the diff (confirmed via `gh pr diff --name-only`), and the earlier four blocking findings (public-visibility enforcement in `_resolve_registration_scope`, visibility coercion on transfer, deactivated-user ownership fallback, orphan-error mapping) were already independently re-verified fixed in the prior round with matching regression tests.

All CI checks are green. Remaining suggestions (shared active-membership-check helper, transfer owner-guard parity with the deletion path, dropping the read-only `db.commit()` in `_resolve_registration_scope`, CHANGELOG note on the new `gateways.create` requirement) are non-blocking — fine to land as follow-ups.

Approving.

msureshkumar88 (dismissed):

msureshkumar88 (approved):

Issue #4956, opened 2026-05-29
[FEATURE]: Add /v1/mcp-servers and /v1/virtual-servers as versioned API aliases


marekdano:
## Summary

Add `/v1/mcp-servers` and `/v1/virtual-servers` as versioned aliases for the existing `/gateways` and `/servers` endpoints respectively. The legacy paths must remain fully functional for backward compatibility.

## Motivation

The current endpoint names (`/gateways`, `/servers`) do not reflect the domain language used in the product. The new versioned paths align naming with the concepts of MCP Servers and Virtual Servers, and establish a `/v1/` prefix convention for future API versioning.

## New endpoints

  | New path | Alias for |
  |---|---|
  | `GET/POST /v1/mcp-servers` | `/gateways` |
  | `GET/PUT/DELETE /v1/mcp-servers/{id}` | `/gateways/{id}` |
  | `POST /v1/mcp-servers/{id}/state` | `/gateways/{id}/state` |
  | `POST /v1/mcp-servers/{id}/tools/refresh` | `/gateways/{id}/tools/refresh` |
  | `GET/POST /v1/virtual-servers` | `/servers` |
  | `GET/PUT/DELETE /v1/virtual-servers/{id}` | `/servers/{id}` |
  | `POST /v1/virtual-servers/{id}/state` | `/servers/{id}/state` |
  | `GET /v1/virtual-servers/{id}/sse` | `/servers/{id}/sse` |
  | `POST /v1/virtual-servers/{id}/message` | `/servers/{id}/message` |
  | `GET /v1/virtual-servers/{id}/tools` | `/servers/{id}/tools` |
  | `GET /v1/virtual-servers/{id}/resources` | `/servers/{id}/resources` |
  | `GET /v1/virtual-servers/{id}/prompts` | `/servers/{id}/prompts` |
  | `GET /v1/virtual-servers/{id}/.well-known/*` | `/servers/{id}/.well-known/*` |

## Files requiring changes

  1. **`mcpgateway/main.py`**
     - Remove prefix from `gateway_router` and `server_router` `APIRouter()` constructors (lines 3312, 3315)
     - Re-register both routers at `include_router` time with both old and new prefixes (lines ~11967–11971)
     - Also register `server_well_known_router` under `/v1/virtual-servers`
     - Update `MCPPathRewriteMiddleware` (line ~2975) to also rewrite `/v1/virtual-servers/{id}/mcp` → `/mcp/`

[...]

PR #6257, opened 2026-08-14
Add 2 new paths as versioned API aliases

byrnec30:
# Pull Request

## 🔗 Related Issue
Relates to [IBM/mcp-context-forge#4956](https://github.com/IBM/mcp-context-forge/issues/4956)

Closes #

---

## 📝 Summary

_What does this PR do and why?_
Adds `/v1/mcp-servers` and `/v1/virtual-servers` as aliases for the existing `/gateways` and `/servers` APIs.
The new routes apply the same authentication, RBAC, token-scoping, security-header, protocol-version, client-disconnect, and MCP path-rewrite handling as the existing paths. Legacy routes remain unchanged for backward compatibility.

---

## 📏 Reviewability

---

## 🏷️ Type of Change

- [ ] Bug fix
- [x] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

| Lint suite                | `make lint`     |    YAML warnings and JSON errors that come from unchanged files that are not part of this PR. Focused Ruff checks and `git diff --check` pass for these PR changes.   |
| Unit tests                | `make test`     |   one unrelated local failure caused by NVM login-shell output. The affected test is unchanged by this PR.     |
| Coverage ≥ 80%            | `make coverage` |     98%   |

---

## ✅ Checklist

---

## 📓 Notes (optional)

### DIFF
--- a/mcpgateway/api/v1/__init__.py
+++ b/mcpgateway/api/v1/__init__.py
@@ -371,8 +371,6 @@ def build_v1_router(
     )
 
     # Product-language aliases for the existing gateway and virtual-server APIs.
-    # These are v1-only additions; the unversioned compatibility routes retain
-    # their established /gateways and /servers paths.
     v1_router.include_router(gateway_router, prefix="/mcp-servers")
     v1_router.include_router(server_router, prefix="/virtual-servers")
 
--- a/mcpgateway/api/v1/__init__.py
+++ b/mcpgateway/api/v1/__init__.py
@@ -68,9 +68,9 @@ def _assemble_routers(  # noqa: C901 — deliberate single-function assembly, co
         tool_router: Inline tools router from main.py.
         resource_router: Inline resources router from main.py.
         prompt_router: Inline prompts router from main.py.
-        gateway_router: Inline gateways router from main.py.
+        gateway_router: Prefix-free inline gateways router from main.py.
         root_router: Inline roots router from main.py.
-        server_router: Inline servers router from main.py.
+        server_router: Prefix-free inline servers router from main.py.
         metrics_router: Inline metrics router from main.py.
         tag_router: Inline tags router from main.py.
         export_import_router: Inline export/import router from main.py.
@@ -84,9 +84,9 @@ def _assemble_routers(  # noqa: C901 — deliberate single-function assembly, co
     target_router.include_router(tool_router)
     target_router.include_router(resource_router)
     target_router.include_router(prompt_router)
-    target_router.include_router(gateway_router)
+    target_router.include_router(gateway_router, prefix="/gateways")
     target_router.include_router(root_router)
-    target_router.include_router(server_router)
+    target_router.include_router(server_router, prefix="/servers")
     target_router.include_router(metrics_router)
     target_router.include_router(tag_router)
     target_router.include_router(export_import_router)
@@ -369,6 +369,12 @@ def build_v1_router(
         export_import_router=export_import_router,
         a2a_router=a2a_router,
     )
+
+    # Product-language aliases for the existing gateway and virtual-server APIs.
+    # These are v1-only additions; the unversioned compatibility routes retain
+    # their established /gateways and /servers paths.
+    v1_router.include_router(gateway_router, prefix="/mcp-servers")
+    v1_router.include_router(server_router, prefix="/virtual-servers")
 
     # First-Party
     from mcpgateway.routers.catalog import router as catalog_router  # pylint: disable=import-outside-toplevel
--- a/mcpgateway/main.py
+++ b/mcpgateway/main.py
@@ -3069,6 +3069,9 @@ class AdminAuthMiddleware(BaseHTTPMiddleware):
         return await call_next(request)
 
 
+_SERVER_MCP_PATH_RE = re.compile(r"/servers/([^/]+)/mcp/?")
+
+
 class MCPPathRewriteMiddleware:
     """
     Middleware that rewrites paths ending with '/mcp' to '/mcp/', after performing authentication.
@@ -3242,7 +3245,7 @@ class MCPPathRewriteMiddleware:
                     # Validate that a non-empty server_id segment is present.
                     # Without this check, paths like /servers//mcp (empty ID)
                     # would be rewritten and silently fall through (#3891).
-                    _srv_match = re.fullmatch(r"/servers/([^/]+)/mcp/?", internal_app_path)
+                    _srv_match = _SERVER_MCP_PATH_RE.fullmatch(internal_app_path)
                 else:
                     # Not a recognised server-scoped path — do not rewrite.
                     await self.application(scope, receive, send)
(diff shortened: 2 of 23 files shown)


marekdano (changes requested):
@byrnec30 - thank you for the contribution!

## Findings

### 🔴 High — should block merge

  #### 1. Auth/existence dependency order flips 404 → 401
  **File:** `mcpgateway/main.py:4629`

  `message_endpoint`'s dependencies were reordered so `get_current_user_with_permissions` resolves before `require_valid_server`. FastAPI resolves `Depends()` left-to-right, so an unauthenticated request against a **nonexistent** server now returns `401` instead of the previous `404`.

  - **Failure scenario:** `POST /servers/{nonexistent-id}/message` (or the new `/v1/virtual-servers/{nonexistent-id}/message` alias) with no/invalid credentials used to return 404; now returns 401.
  - **Why it's missed:** `test_server_message_endpoint_rejects_nonexistent_server` always sends valid `auth_headers`, so it can't detect the flip either way.
  - **Risk:** Any external client or integration test asserting 404 for this combination silently breaks.

  #### 2. Streaming-path detection breaks under `APP_ROOT_PATH`
  **File:** `mcpgateway/middleware/client_disconnect.py:44`

  `_is_server_streaming_path` counts raw URL segments without stripping any reverse-proxy root path prefix.

  - **Failure scenario:** With `APP_ROOT_PATH=/dev/mcp-gateway`, a client opens `POST /dev/mcp-gateway/v1/virtual-servers/{id}/mcp`. `ClientDisconnectMiddleware` sees the un-rewritten path (it runs outside `MCPPathRewriteMiddleware`), gets 6+ segments instead of the expected 5, and returns `False` for a genuine long-lived MCP stream.
  - **Risk:** The stream gets wrapped in queue-based disconnect-cancellation logic meant only for non-streaming handlers — premature cancellation or double-handling of the stream's own disconnect detection. Affects both the new alias and the pre-existing `/servers` path whenever a root path is configured.

  ---

  ### 🟡 Medium — worth fixing, not blocking

  #### 3. Missing test for new alias branch
  **File:** `mcpgateway/middleware/client_disconnect.py:37`

  The new `/v1/virtual-servers` branch in `_is_server_streaming_path` has no unit test, unlike every other file this PR touches. A future refactor of the segment-count check or alias prefix would regress silently.

  #### 4. Alias-path regex logic hand-duplicated across 5 files
  **File:** `mcpgateway/middleware/token_scoping.py:313` (and `main.py`, `client_disconnect.py`, `protocol_version.py`, `security_headers.py`)

  Alias recognition for `/virtual-servers` and `/mcp-servers` is reimplemented independently in five places instead of normalized once at an early layer. Adding/renaming an alias, or fixing a permission rule, requires remembering to edit all five — missing one (e.g. a `GET /prompts` permission entry) silently reintroduces an RBAC gap on the alias route.

  #### 5. Shared `scope["path"]` mutated and restored via try/finally
  **File:** `mcpgateway/main.py:3226`

  `_call_streamable_http` temporarily overwrites `scope["path"]` to feed `streamable_http_auth` a canonicalized path, then restores it in `finally`, instead of passing the canonical path explicitly.

  - **Risk:** Any code that reads `scope["path"]` during the awaited `streamable_http_auth` call sees the temporarily-wrong value; a future refactor of `streamable_http_auth` that itself mutates `scope` could clobber the restore.

  ---

  ### 🟢 Low — minor/nice-to-have

  #### 6. Alias regex compiled and matched twice per request
  **File:** `mcpgateway/main.py:3259`

  `r"/v1/virtual-servers/([^/]+)/mcp/?"` is a string literal compiled and `fullmatch`'d twice per aliased request in a streaming hot path, instead of precompiled as a module constant (the codebase's own convention, e.g. `token_scoping.py`'s `_SERVER_PATH_PATTERNS`).

  #### 7. Duplicated trailing-slash computation
  **File:** `mcpgateway/main.py:3220`

  `trailing_slash = "/" if app_path.endswith("/") else ""` is duplicated verbatim twice in `_call_streamable_http`. If the canonicalization rule changes, only one copy might get updated, causing the auth path and rewrite path to disagree.

marekdano:
@byrnec30 - Is it still ok working on this PR? Can you look at the failing checks?

byrnec30:
1. Dependency order:
Fixed by restoring require_valid_server before authentication. Added regression coverage.

2. Streaming path:
Fixed. ClientDisconnectMiddleware now removes the ASGI/configured root path before identifying self-managed streaming endpoints. Added coverage for legacy and alias MCP streams, regular REST paths, and partial-prefix safety.

3&4: implemented

5: Fixed — auth now receives the internal path directly, without modifying scope["path"]. 

6&7: implemented

byrnec30:
@marekdano ready for a next review

marekdano (dismissed):
@byrnec30 - thanks for addressing the issues!

Both high-severity issues and all medium/low items are properly fixed with solid

LGTM 🚀

marekdano:
@byrnec30 - Can you please resolve conflicts?

marekdano (approved):

msureshkumar88 (approved):
Reviewed against issue #4956 — implementation covers every endpoint in the acceptance table and improves on the issue's own plan: a single alias table in `mcpgateway/utils/paths.py` (`replace_api_path_alias`) rather than duplicating patterns across five middleware modules, which keeps the alias surface a one-line change to extend.

Verified locally: OpenAPI route parity between each alias and its canonical path (zero missing endpoints, zero method diffs, zero duplicate `operationId`s), all touched unit + integration suites green, `paths.py` doctests pass, `ruff` clean on all changed source files. No Alembic/schema change — correct, since this is routing-only. No breaking change for existing consumers; legacy paths are untouched and `test_legacy_router.py` explicitly guards against alias leakage into the unversioned router.

Also want to flag a real, well-tested security fix buried in here: `MCPPathRewriteMiddleware` previously authenticated against the raw `scope["path"]`, which meant per-server OAuth enforcement (`oauth_enabled=True` servers) silently skipped under a reverse-proxy `APP_ROOT_PATH` prefix, since the server-ID regex is anchored to `^/servers/`. This PR threads the root-path-stripped, alias-normalized path into `streamable_http_auth`, and I confirmed by mutation testing that reverting either of the two affected call sites is caught by `test_streamablehttp_transport.py` (`test_streamable_http_auth_uses_request_path_without_changing_scope`, `test_oauth_access_token_context_preserves_exp`). Worth calling out explicitly in the PR description since it reads as incidental to the alias work.

Two non-blocking items worth a follow-up (happy to open issues if useful):
- `token_scoping.py`'s `^/gateways/test(?:$|/)` pattern now sits above the generic `/gateways/{id}/` sub-resource rule and would misclassify a gateway literally named `test` (currently unreachable — gateway IDs are server-generated UUIDs, not client-settable). Anchoring to `/?$` closes it cheaply.
- The alias 401 deny-path tests (`tests/integration/test_api_versioning_security.py`) are solid but live in a suite nothing currently runs with `--with-integration` — pre-existing repo pattern, not introduced by this PR, but worth a unit-test mirror at some point.

Nice work on the centralized alias approach — approving.

PR #6258, opened 2026-08-15
fix(llm-proxy): forward tool-calling through the Anthropic chat-completion path

AmirF194:
## Related Issue

None linked. This repo's own measured pattern (13/13 outside-filed issues with `NONE`
association were fixed by someone other than the filer) means a standalone issue from
this account would sit unfixed, so this is a direct PR with the full repro in the
description below, per that measured shape.

## Summary

The gateway's unified `/v1/chat/completions` proxy is OpenAI-shaped on the client side,
but the Anthropic backend path silently drops tool-calling end to end. A client that
sends `tools`/`tool_choice` against an Anthropic-backed model gets no tools forwarded
upstream at all, and even a tool call Claude makes anyway (tools attached on an earlier
turn) is stripped out of both the non-streaming and streaming response, with
`finish_reason` hardcoded to `"stop"` in both cases. Any agent framework built on this
gateway sees "the model just replied with text" instead of a tool call.

## Root Cause

Three linked gaps in `llm_proxy_service.py`, all on the Anthropic-specific code paths:

- `_build_anthropic_request` builds the outbound body from `request.temperature` /
  `max_tokens` / `stream` only; it never reads `request.tools` or `request.tool_choice`
  (the OpenAI-compatible sibling, `_build_openai_request`, already forwards both).
- `_transform_anthropic_response` iterates Anthropic's `content` blocks and only handles
  `type == "text"`; a `tool_use` block is silently skipped, and `finish_reason` is
  `data.get("stop_reason", "stop")` verbatim, so Anthropic's own `"tool_use"` stop reason
  leaks through unmapped instead of becoming OpenAI's `"tool_calls"`.
- `_transform_anthropic_stream_chunk` only handles `content_block_delta` events of type
  `"text_delta"`; the `content_block_start` event that opens a tool-use block and the
  `input_json_delta` events that stream its arguments are both unhandled, and the
  finish-reason chunk is emitted unconditionally on `message_stop` as a hardcoded
  `"stop"` (Anthropic's own protocol carries the real `stop_reason` on the preceding
  `message_delta` event, not on `message_stop`, which carries no payload).

## Fix

- `_build_anthropic_request` now forwards `tools` in Anthropic's own shape
  (`name`/`description`/`input_schema`, not OpenAI's `type`/`function` wrapper) and maps
  `tool_choice` via a new `_map_tool_choice_to_anthropic` helper (`"auto"` -> `{"type":
  "auto"}`, `"required"` -> `{"type": "any"}`, `{"function": {"name": ...}}` -> `{"type":
  "tool", "name": ...}`). OpenAI's `"none"` has no faithful Anthropic equivalent once
  tools are attached, so that case sends no `tool_choice` key and falls back to
  Anthropic's own default (`auto`) rather than sending something misleading.
- `_transform_anthropic_response` now collects `tool_use` blocks into OpenAI-shaped
  `tool_calls` entries and maps `stop_reason` through a new `_map_anthropic_stop_reason`
  helper (`tool_use` -> `tool_calls`, `max_tokens` -> `length`, `end_turn`/
  `stop_sequence` -> `stop`, anything else passed through as-is).
- `_transform_anthropic_stream_chunk` now emits an OpenAI-shaped `tool_calls` delta on
  `content_block_start` (when the block is `tool_use`) and on `input_json_delta`, using
  Anthropic's own block `index` so multiple concurrent tool calls stay distinguishable.
  The finish-reason chunk moved to `message_delta`'s `stop_reason` (mapped through the
  same helper); `message_stop` is now a no-op, since by Anthropic's own event ordering
  `message_delta` always precedes it and always carries the real `stop_reason`.

## Verification

- New/updated unit tests fail against unmodified `main` (`18 failed, 69 passed`) and
  pass on this branch. Full `tests/unit/mcpgateway/services/test_llm_proxy_service.py`
  on the branch: 87 passed; branch coverage on the changed module is 99% (359/360
  lines), the one uncovered line a pre-existing, untouched Ollama JSON-decode branch.
- `pylint --rcfile=.pylintrc.mcpgateway` on the changed module: 10.00/10. Could not run
  this repo's pinned `ruff` (0.15.1, per `lint.yml`): `pyproject.toml`'s
  `[tool.ruff.lint] ignore` list contains `"PLW0717"`, which that version rejects as an
  unknown rule selector, reproducing identically on unmodified `main`; a newer `ruff`,
  informational only, reports the changed files clean.
- Not verified: a live call against the real Anthropic API. This is a request/response/
  stream-transform change checked against Anthropic's documented Messages API event
  shapes and this repo's own existing test fixtures for the same functions; I do not
  have an Anthropic API key in this environment.

Signed-off-by: Amir Fathi <amirfathi.me@gmail.com>

### DIFF
--- a/mcpgateway/services/llm_proxy_service.py
+++ b/mcpgateway/services/llm_proxy_service.py
@@ -12,7 +12,7 @@ formats and handles streaming responses.
 
 # Standard
 import time
-from typing import Any, AsyncGenerator, Dict, Optional, Tuple
+from typing import Any, AsyncGenerator, Dict, Optional, Tuple, Union
 import uuid
 
 # Third-Party
@@ -378,10 +378,48 @@ class LLMProxyService:
         elif provider.default_temperature:
             body["temperature"] = provider.default_temperature
 
+        if request.tools:
+            body["tools"] = [
+                {
+                    "name": t.function.name,
+                    "description": t.function.description or "",
+                    "input_schema": t.function.parameters or {"type": "object", "properties": {}},
+                }
+                for t in request.tools
+            ]
+
+        if request.tool_choice:
+            anthropic_tool_choice = self._map_tool_choice_to_anthropic(request.tool_choice)
+            if anthropic_tool_choice:
+                body["tool_choice"] = anthropic_tool_choice
+
         if request.stream:
             body["stream"] = True
 
         return url, headers, body
+
+    @staticmethod
+    def _map_tool_choice_to_anthropic(tool_choice: Union[str, Dict[str, Any]]) -> Optional[Dict[str, Any]]:
+        """Map an OpenAI-shaped ``tool_choice`` to Anthropic's tool-choice object.
+
+        Args:
+            tool_choice: OpenAI-style tool choice ("auto", "required", "none",
+                or {"type": "function", "function": {"name": ...}}).
+
+        Returns:
+            The equivalent Anthropic tool-choice dict, or None when there is no
+            faithful equivalent (OpenAI's "none" cannot be expressed once tools
+            are attached to an Anthropic request; the caller omits the key so
+            Anthropic falls back to its own default of "auto").
+        """
+        if isinstance(tool_choice, dict):
+            name = tool_choice.get("function", {}).get("name")
+            return {"type": "tool", "name": name} if name else None
+        if tool_choice == "required":
+            return {"type": "any"}
+        if tool_choice == "auto":
+            return {"type": "auto"}
+        return None
 
     def _build_ollama_request(
         self,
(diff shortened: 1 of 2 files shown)


AmirF194:
Just checking in on this one, no rush. The CI workflows are still sitting behind the first-time-contributor approval gate, so let me know if there is anything you would like changed in the meantime.

AmirF194:
Closing this out after 30 days with no response. The bug is still real (the Anthropic path of the chat-completion proxy drops tool-calling end to end), so if anyone wants to pick it up the repro and fix are both here. Leaving the branch up in case it's worth reopening later.

PR #6269, opened 2026-08-17
fix(ci): remove expired sunset time-bomb assertion and deprecated PUT endpoint

Lang-Akshay:
## Problem

CI is failing because the sunset date (Aug 16, 2026) for the deprecated `PUT /admin/users/{email}` endpoint has passed. A deliberate time-bomb assertion in the test suite enforces this:

```python
assert datetime.now(timezone.utc) < datetime(2026, 8, 16, 23, 59, 59, tzinfo=timezone.utc), "Sunset reached: remove deprecated PUT endpoint. See #2754"
```

## Fix

The sunset date has been reached, so this PR does exactly what the assertion demanded — removes the deprecated code:

- **`mcpgateway/routers/email_auth.py`**: Remove `update_user_deprecated` (PUT endpoint) and `update_user_delegate` helper; inline the logic back into the PATCH `update_user` handler. Clean up unused `Response` import.
- **`tests/unit/mcpgateway/routers/test_email_auth_router.py`**: Remove deprecated PUT endpoint test block and the sunset time-bomb assertion. Clean up unused `Response` import.
- **`scripts/test_email_auth_api.py`**: Remove deprecated PUT compatibility test block.

**Net diff: +2 / -88 lines.**

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^\\.secrets\\.baseline$",
     "lines": null
   },
-  "generated_at": "2026-08-14T16:05:27Z",
+  "generated_at": "2026-08-17T09:59:35Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
--- a/mcpgateway/routers/email_auth.py
+++ b/mcpgateway/routers/email_auth.py
@@ -21,7 +21,7 @@ from datetime import datetime, timedelta, UTC
 from typing import List, Optional, Union
 
 # Third-Party
-from fastapi import APIRouter, Depends, HTTPException, Query, Request, Response, status
+from fastapi import APIRouter, Depends, HTTPException, Query, Request, status
 from fastapi.security import HTTPBearer
 from sqlalchemy.orm import Session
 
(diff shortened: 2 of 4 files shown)


gandhipratik203 (approved):
LGTM! Thanks for the fix. This should get the CI back up agian!

msureshkumar88 (approved):
Reviewed against the linked issue (#2754) and the codebase's own time-bomb comment — this does exactly what was pre-planned: removes the deprecated `PUT /admin/users/{email}` endpoint now that the sunset date has passed, folding `update_user_delegate` back into `update_user` as instructed. Scope is tight (+2/-88, no incidental changes) and no schema/migration touch is needed here.

Traced the admin UI's update flow and confirmed it never called this `PUT`/`PATCH` router endpoint (it posts to a separate `admin_router` route), so this has zero UI impact — pure backend/API surface change. Verified no dangling references to the removed symbols anywhere in the repo, and existing `PATCH`/`update_user` test coverage stays solid.

A couple of small, non-blocking things worth a follow-up if it's easy:
- `docs/docs/manage/password-management.md` and `docs/docs/architecture/multitenancy.md` still show `PUT /auth/email/admin/users/{email}` examples — worth updating those to `PATCH` now that `PUT` 405s.
- The removed test block also happened to drop the only unit-level happy-path assertion for `delete_user` (error branches are still covered, and the integration script still exercises the success path, so no real gap — just something to backfill if convenient).
- Minor: 3 consecutive blank lines were left behind in `tests/unit/mcpgateway/routers/test_email_auth_router.py` and `scripts/test_email_auth_api.py` where blocks were removed (outside the ruff lint gate's scope, so it won't fail CI, but a quick `ruff check --fix` would tidy it up).

None of the above block merge. Approving.

PR #6290, opened 2026-08-18
docs: fix IBM Cloud Code Engine deployment guide missing env secret

### LINKED ISSUE #6289: [DOCS]: IBM Cloud Code Engine deployment guide missing .env secret injection step
### 📚 Documentation Issue Summary
The IBM Cloud Code Engine deployment guide is missing the step to inject `.env` runtime settings into the running container. Users who follow the guide end up with a deployed but misconfigured application — none of the values from `.env` (JWT secret, database URL, auth settings, etc.) are present inside the container.

---

### 📍 Location of the Problem

`docs/docs/howto/ibm-cloud-code-engine.md`
https://ibm.github.io/mcp-context-forge/howto/ibm-cloud-code-engine/

Affected sections:
- **Section 3 - Workflow A** (Makefile targets) — the `make ibmcloud-deploy` target never passes env vars to Code Engine
- **Section 4 - Workflow B** (Manual CLI) — Step 7 (`ibmcloud ce application create/update`) has no `--env-from-secret` or equivalent

---

### ✏️ Type of Issue

- [ ] Typo or formatting issue
- [ ] Outdated or incorrect information
- [x] Missing explanation or example
- [x] Unclear instructions
- [ ] Other (explain below)

---

### 💡 Suggested Fix or Clarification

**Root cause:** `make ibmcloud-deploy` calls `ibmcloud ce application create/update` with only `--image`, `--cpu`, `--memory`, `--port`, and `--registry-secret`. The `--env-file=.env` flag used by local `docker`/`podman run` is a container-runtime flag — it does not apply to Code Engine, which pulls the image remotely. The env vars must be explicitly loaded as a Code Engine secret.

**Fix for Workflow B — add a new step between step 6 (registry secret) and step 7 (deploy):**

```bash
# 6b - Create a runtime environment secret from .env
ibmcloud ce secret create --name mcpgw-env --from-env-file .env
```

And update step 7 to attach it:

# 7 - Deploy / update
```
if ibmcloud ce application get --name "$IBMCLOUD_CODE_ENGINE_APP" >/dev/null 2>&1; then
  ibmcloud ce application update --name "$IBMCLOUD_CODE_ENGINE_APP" \
      --image "$IBMCLOUD_IMAGE_NAME" \
      --cpu "$IBMCLOUD_CPU" --memory "$IBMCLOUD_MEMORY" \
      --registry-secret "$IBMCLOUD_REGISTRY_SECRET" \
      --env-from-secret mcpgw-env
else
  ibmcloud ce application create --name "$IBMCLOUD_CODE_ENGINE_APP" \
      --image "$IBMCLOUD_IMAGE_NAME" \
      --cpu "$IBMCLOUD_CPU" --memory "$IBMCLOUD_MEMORY" \
      --port 4444 \
      --registry-secret "$IBMCLOUD_REGISTRY_SECRET" \
      --env-from-secret mcpgw-env
fi
```

Fix for Workflow A — update the make ibmcloud-deploy target in Makefile (lines 6171-6187) to accept and wire --env-from-secret.

Also add a note about HOST=0.0.0.0 — if .env contains HOST=127.0.0.1 (the config.py default), the app binds to loopback only and Code Engine's routing cannot reach it.

When .env changes in future:
```

ibmcloud ce secret update --name mcpgw-env --from-env-file .env
ibmcloud ce application update --name "$IBMCLOUD_CODE_ENGINE_APP"
```

### 🧩 Additional Context
The env injection step does appear in the guide, but only in the PostgreSQL (Section 7) and Redis (Section 8) subsections — and only for DATABASE_URL and REDIS_URL individually. The core .env secret creation step is absent from both main deployment workflows, making it easy to end up with a running but completely unconfigured application.

The Makefile target also needs updating alongside the docs so the two stay in sync.

prakhar-singh1928:
# Pull Request

## 🔗 Related Issue

Closes #6289 

---

## 📝 Summary

The IBM Cloud Code Engine deployment guide was missing the step to inject `.env` runtime settings into the running container. Users following the guide ended up with a deployed but unconfigured application — none of the values from `.env` (JWT secret, database URL, auth settings, etc.) were present inside the container.

Two root causes:

1. `make ibmcloud-deploy` called `ibmcloud ce application create/update` without passing any environment variables — the `--env-file=.env` flag used by local `docker`/`podman run` does not apply to Code Engine, which pulls the image remotely and has no access to the local filesystem.
2. Neither Workflow A (Makefile) nor Workflow B (manual CLI) in the guide mentioned creating a Code Engine secret from `.env` or attaching it to the app.

**Changes:**
- `make ibmcloud-deploy` now automatically creates or updates a Code Engine secret named `mcpgw-env` from the local `.env` on every run, and passes `--env-from-secret mcpgw-env` to both `application create` and `application update`
- Workflow B manual CLI steps gain step 6b (secret creation) and `--env-from-secret mcpgw-env` on step 7
- Corrected the false claim in Section 2 that `.env` is "mounted via `--env-file`" on Code Engine
- Added a `HOST=0.0.0.0` warning callout — if left as `127.0.0.1`, the app binds to loopback only and Code Engine routing cannot reach it
- Added a "update runtime config" snippet for the case where `.env` changes without a code change
- Added two new troubleshooting rows: missing env vars and `HOST=127.0.0.1` causing 502s

---

## 📏 Reviewability

---

## 🏷️ Type of Change

- [ ] Bug fix
- [ ] Feature / Enhancement
- [x] Documentation
- [ ] Refactor
- [x] Other — Makefile tooling fix alongside docs

---

## 🧪 Verification

This is a documentation and Makefile tooling fix. The Makefile change can be verified by running `make ibmcloud-deploy` against a real Code Engine project and confirming:
1. The `mcpgw-env` secret is created/updated from `.env`
2. `ibmcloud ce application get --name <app>` shows the secret listed under environment references
3. Env vars are present inside the container

| Lint suite                | `make lint`     | N/A — docs + Makefile only |
| Unit tests                | `make test`     | N/A — no Python changes |
| Coverage ≥ 80%            | `make coverage` | N/A — no Python changes |

---

## ✅ Checklist

---

## 📓 Notes

The `--from-env-file` flag on `ibmcloud ce secret` skips lines with inline comments (e.g. `KEY=value  # comment`). Users copying `.env.example` verbatim should strip inline comments before running `make ibmcloud-deploy`. This is a pre-existing `.env.example` formatting issue tracked separately.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^\\.secrets\\.baseline$",
     "lines": null
   },
-  "generated_at": "2026-09-09T13:44:53Z",
+  "generated_at": "2026-09-09T14:20:06Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -346,7 +346,7 @@
         "hashed_secret": "5932862bcd24dd27d0dc0407ec94fe9d6ea24aeb",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 6651,
+        "line_number": 6652,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -354,7 +354,7 @@
         "hashed_secret": "c77c805e32f173e4321ee9187de9c29cb3804513",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 6663,
+        "line_number": 6664,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -362,7 +362,7 @@
         "hashed_secret": "8fe3df8a68ddd0d4ab2214186cbb8e38ccd0e06a",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 6735,
+        "line_number": 6736,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -370,7 +370,7 @@
         "hashed_secret": "bb589d0621e5472f470fa3425a234c74b1e202e8",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 7151,
+        "line_number": 7152,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -378,7 +378,7 @@
         "hashed_secret": "93ac8946882128457cd9e283b30ca851945e6690",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 7841,
+        "line_number": 7842,
         "type": "Secret Keyword",
         "verified_result": null
       }
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^\\.secrets\\.baseline$",
     "lines": null
   },
-  "generated_at": "2026-09-09T08:35:26Z",
+  "generated_at": "2026-09-09T13:44:53Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -346,7 +346,7 @@
         "hashed_secret": "5932862bcd24dd27d0dc0407ec94fe9d6ea24aeb",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 6639,
+        "line_number": 6651,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -354,7 +354,7 @@
         "hashed_secret": "c77c805e32f173e4321ee9187de9c29cb3804513",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 6651,
+        "line_number": 6663,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -362,7 +362,7 @@
         "hashed_secret": "8fe3df8a68ddd0d4ab2214186cbb8e38ccd0e06a",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 6723,
+        "line_number": 6735,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -370,7 +370,7 @@
         "hashed_secret": "bb589d0621e5472f470fa3425a234c74b1e202e8",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 7139,
+        "line_number": 7151,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -378,7 +378,7 @@
         "hashed_secret": "93ac8946882128457cd9e283b30ca851945e6690",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 7829,
+        "line_number": 7841,
         "type": "Secret Keyword",
         "verified_result": null
       }
@@ -1722,7 +1722,7 @@
         "hashed_secret": "60ef47f289f32388d377100aec5a69a39bbdd087",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 70,
+        "line_number": 71,
         "type": "Basic Auth Credentials",
         "verified_result": null
       },
@@ -1730,7 +1730,7 @@
         "hashed_secret": "a095f277a165b357774752331635b3693745d048",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 106,
+        "line_number": 107,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -1738,7 +1738,7 @@
         "hashed_secret": "aa91e9f973a0fc76e4c2c303ec788f7b6c22a318",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 113,
+        "line_number": 114,
         "type": "Secret Keyword",
         "verified_result": null
       }
--- a/Makefile
+++ b/Makefile
@@ -6253,6 +6253,7 @@ ibmcloud-push:
 
 .PHONY: ibmcloud-deploy
 ibmcloud-deploy:
+	@test -f .env || { echo "❌ Missing .env — run: cp .env.example .env"; exit 1; }
 	@echo "🚀 Deploying image to Code Engine as '$(IBMCLOUD_CODE_ENGINE_APP)' using registry secret $(IBMCLOUD_REGISTRY_SECRET)..."
 	@# Create the runtime env secret from .env if it does not exist yet
 	@if ! ibmcloud ce secret get --name $(IBMCLOUD_CODE_ENGINE_APP)-env > /dev/null 2>&1; then \
--- a/Makefile
+++ b/Makefile
@@ -6112,7 +6112,7 @@ compose-test-hardened: compose-validate
 # help: ibmcloud-deploy             - Deploy (or update) container image in Code Engine
 # help: ibmcloud-ce-logs            - Stream logs for the deployed application
 # help: ibmcloud-ce-status          - Get deployment status
-# help: ibmcloud-ce-rm              - Delete the Code Engine application
+# help: ibmcloud-ce-rm              - Delete the Code Engine application and its env secret
 
 .PHONY: ibmcloud-check-env ibmcloud-cli-install ibmcloud-login ibmcloud-ce-login \
 	ibmcloud-list-containers ibmcloud-tag ibmcloud-push ibmcloud-deploy \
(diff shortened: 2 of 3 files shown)


claudia-gray:
Found one blocking issue and should be good after that.

The secret name should use $(IBMCLOUD_CODE_ENGINE_APP)-env not "mcpgw-env"

The hardcoded "mcpgw-env" is inconsistent with how everything else in the target is named. Every other --name reference uses $(IBMCLOUD_CODE_ENGINE_APP) and .env.ce.example sets that to mcpgateway, not "mcpgw". If someone follows the example and their app is named mcpgateway, the secret ends up as "mcpgw-env" while any future targets expecting mcpgateway-env won't find it. Also change it in the docs too. 5 instances of it in Makefile to be changed.

claudia-gray:
Verified IBM Cloud Code Engine workflow for contextforge deployment end-to-end is successful. Here are some findings and suggested changes:

**Suggested changes/fixes:**
1.  **Workflow B step 6b is missing an update path.** It says "first time only" but gives no instruction for subsequent .env changes. Add the following

```
# To update later (e.g. rotating credentials):
ibmcloud ce secret update --name "${IBMCLOUD_CODE_ENGINE_APP}-env" --from-env-file .env
ibmcloud ce application update --name "$IBMCLOUD_CODE_ENGINE_APP"
```

2. **Workflow B step 6 uses ibmcloud ce registry create-secret which doesn't exist in current CLI versions.** Hit this during testing, the correct command is:

```
ibmcloud ce secret create --name "$IBMCLOUD_REGISTRY_SECRET" \
    --format registry \
    --server "$(echo "$IBMCLOUD_IMAGE_NAME" | cut -d/ -f1)" \
    --username iamapikey --password "$IBMCLOUD_API_KEY"
```

**Minor changes:**

3. **Prerequisites table is missing required IAM roles.** During testing I hit permission failures for both Code Engine project creation and Container Registry namespace creation. Suggest adding a row:
RequirementDetailsIAM permissionsEditor or higher on Code Engine; Writer or higher on Container Registry in the resource group.

4. **make ibmcloud-ce-rm doesn't clean up the <app>-env secret**. The cleanup section in the docs correctly shows the manual delete command, but the Makefile target leaves the secret orphaned. Probably best to create a follow-up issue for it.

5. Also one note on .**env formatting: ibmcloud ce secret --from-env-file rejects lines with spaces around =** (e.g. KEY = value) with a hard failure. This wasn't from the project's .env.example (which is correctly formatted) but worth keeping in mind for users who edit their .env manually.

6. **ibmcloud cr region-set needed before push for non-us-south regions.** Testing in eu-de - make ibmcloud-push failed with auth errors until running ibmcloud cr region-set eu-de first. The Makefile doesn't set this automatically. Either add ibmcloud cr region-set to the ibmcloud-push target or just document that it's needed beforehand.

claudia-gray (approved):
LGTM.

--- 

Minor non-blocking issue below:

Workflow A is missing a registry secret creation step. The "Typical first deploy" sequence goes straight from make ibmcloud-push to make ibmcloud-deploy, but make ibmcloud-deploy fails if the registry pull secret ($IBMCLOUD_REGISTRY_SECRET) doesn't exist yet. Hits this error:

FAILED
Referenced image registry access secret 'my-regcred' not found.

Workflow B covers this in step 6, but Workflow A has no equivalent. Maybe add a note between make ibmcloud-push and make ibmcloud-deploy in the first deploy sequence:

First time only - create the registry pull secret
ibmcloud ce secret create --name "$IBMCLOUD_REGISTRY_SECRET" \
    --format registry \
    --server "$(echo "$IBMCLOUD_IMAGE_NAME" | cut -d/ -f1)" \
    --username iamapikey --password "$IBMCLOUD_API_KEY"

Longer term, make ibmcloud-deploy could handle this automatically the same way it handles the env secret (check-and-create), but a docs fix is sufficient for now. Follow up issue can be created if necessary

Also, Workflow B uses $IBMCLOUD_* variables throughout (e.g. $IBMCLOUD_REGISTRY_SECRET, $IBMCLOUD_IMAGE_NAME) but never tells users to export .env.ce to their shell first. The Makefile loads it automatically but manual CLI users will get empty variables and confusing failures. Add before the Workflow B steps:

 Load deployment variables into your shell
`export $(grep -v '^#' .env.ce | grep -v '^$' | xargs)`

This is a pre-existing gap, not introduced by this PR - not blocking merge.

ja8zyjits:
## Findings

### F-001 — Medium: `--from-env-file` inline-comment caveat not in doc

**Location:** `docs/docs/howto/ibm-cloud-code-engine.md`, steps 6b and Workflow A explanation  
**Evidence:** PR description Notes section says: *"The `--from-env-file` flag … skips lines with inline comments (e.g. `KEY=value  # comment`). Users copying `.env.example` verbatim should strip inline comments."*  
This warning appears only in the PR description, not in the documentation users will read.  
**Impact:** Users who copy `.env.example` directly (the documented first step) may silently lose env vars that have inline comments, causing hard-to-diagnose misconfiguration — exactly the failure mode this PR is fixing.  
**Recommendation:** Add a `!!! warning` or `!!! note` callout near step 6b (and near the Workflow A explanation) noting that `--from-env-file` skips lines with inline comments and pointing users to strip them before uploading.

---

### F-002 — Low: No `.env` existence guard in `ibmcloud-deploy` target

**Location:** `Makefile`, line ~6258 (`ibmcloud-deploy` target)  
**Evidence:** `ibmcloud-check-env` validates `.env.ce` but not `.env`. If a user runs `make ibmcloud-deploy` without a `.env` file, `ibmcloud ce secret create --from-env-file .env` fails with an opaque CLI error.  
**Impact:** Confusing error message that doesn't match the documented setup step ("copy `.env.example` to `.env`").  
**Recommendation:** Add a guard at the top of `ibmcloud-deploy` (or in `ibmcloud-check-env`):
```makefile
@test -f .env || { echo "❌ Missing .env — run: cp .env.example .env"; exit 1; }
```

---

### F-003 — Low: `ibmcloud ce registry create-secret` → `ibmcloud ce secret create --format registry` change undocumented

**Location:** `docs/docs/howto/ibm-cloud-code-engine.md`, step 6 (Workflow B), line ~237  
**Evidence:** The diff changes `ibmcloud ce registry create-secret` to `ibmcloud ce secret create --format registry`. This is correct (the `registry create-secret` subcommand was deprecated), but the change is not called out in the PR summary or inline in the doc.  
**Impact:** Low — the new form works. But users with existing scripts using the old form get no migration note.  
**Recommendation:** Add a short inline comment in the code block noting the old form is deprecated, or add a callout. Minimal change; even just `# Note: 'ibmcloud ce registry create-secret' is deprecated` suffices.

---

### F-004 — Low: Troubleshooting row uses unquoted variable in shell snippet

**Location:** `docs/docs/howto/ibm-cloud-code-engine.md`, troubleshooting table, line ~592  
**Evidence:**  
```
ibmcloud ce secret create --name ${IBMCLOUD_CODE_ENGINE_APP}-env ...
```
The table cell contains a shell snippet with `${IBMCLOUD_CODE_ENGINE_APP}` but the surrounding text is a Markdown table cell (not a fenced code block), so the snippet may render poorly or confuse users who paste it without quoting.  
**Impact:** Cosmetic/usability — low risk of actual failure since Code Engine app names rarely contain spaces.  
**Recommendation:** Wrap the troubleshooting snippet in a fenced code block or use `"${IBMCLOUD_CODE_ENGINE_APP}-env"` with quotes consistent with the rest of the doc.

---

ja8zyjits (dismissed):
LGTM

msureshkumar88 (changes requested):
Thanks for tracking down the root cause here — the `--env-file` vs Code Engine secret distinction is a real gap and the fix direction (sync `.env` to a per-app `<app>-env` secret) is right. A couple of things worth addressing before merge:

**Blocking**

1. `docs/docs/howto/ibm-cloud-code-engine.md:185,264,610` — the suggested inline-comment stripper
   ```bash
   sed -i.bak 's/[[:space:]]*#.*$//' .env
   ```
   strips everything after the *first* `#` on every line, not just comment lines. `.env.example` in this repo actually has no inline comments, so the stated trigger doesn't occur — but Sections 7/8 of this same guide have users paste `DATABASE_URL`/`REDIS_URL` credentials from the IBM Cloud console, which can legitimately contain `#`. Running this command would silently truncate those. Worth dropping the sed snippet or scoping it to full-line comments only, with a call-out not to use it if any value contains `#`.

2. `Makefile` `ibmcloud-deploy` target — no guard on `.env` existing before `ibmcloud ce secret create --from-env-file .env` runs (unlike `ibmcloud-check-env`, which does guard `.env.ce`). Missing `.env` currently surfaces as a raw CLI error instead of a clear message.

**Suggestion**

- The inline-comment warning/snippet is duplicated 3x across the doc — could reference one canonical spot instead.

msureshkumar88 (dismissed):
Both blocking items resolved in 940f040cb: the unsafe `sed` inline-comment stripper is gone (replaced with a manual check-and-move-comment instruction, no more silent truncation risk for values like `DATABASE_URL`/`REDIS_URL` containing `#`), and the `.env` existence guard is restored in `ibmcloud-deploy`. The duplicated callout is also consolidated to one explanation with two cross-references. LGTM.

msureshkumar88 (approved):

PR #6292, opened 2026-08-18
Pr 5973 rebase

madhu-mohan-jaishankar:
Imported changes from https://github.com/IBM/mcp-context-forge/pull/5973 and addressed the blockers.

### DIFF
--- a/docs/docs/manage/logging.md
+++ b/docs/docs/manage/logging.md
@@ -412,6 +412,7 @@ When `STRUCTURED_LOGGING_DATABASE_ENABLED=true`, you get:
 | **Performance Metrics** | `GET /api/logs/performance-metrics` | Aggregated p50/p95/p99 latencies, error rates by component |
 | **Security Events** | `GET /api/logs/security-events` | Authentication failures, threat detection, security audit |
 | **Audit Trails** | `GET /api/logs/audit-trails` | CRUD operations, data access compliance logging |
+| **Recent Activity** | `GET /api/logs/activity` | Unified newest-first feed merging audit trails and security events (powers the UI home page) |
 
 ### Example: Search Logs via API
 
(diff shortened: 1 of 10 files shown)


vishu-bh (dismissed):
PR successfully refactors RBAC permission checking to support additive in-handler checks while maintaining all existing security guarantees. The new activity feed endpoint implements sophisticated multi-table access control with proper team scoping, data classification filtering, and fail-closed semantics. Test coverage is comprehensive and          security-focused.

gandhipratik203 (dismissed):
LGTM! No blocking changes that I see in scope of the PR. Good to go once the CI is passing.

msureshkumar88 (dismissed):
Solid, well-scoped rebase of #5973. Went through it against the linked issue (#5944), checked the diff against `origin/main` (not local `main`, which was stale) to make sure only the 10 files gh reports were touched.

**Scope & minimality**: every commit maps to a clear need — the endpoint itself, a determinism fix for LIMIT-boundary ties, a real Layer-1 authz bug found while building it (token-scoped `audit:read` callers could see `security:read`-gated events because the inline check didn't enforce token scope — now fixed by moving that gate into the shared `check_permission_inline` helper), and a schema regex widening so the exact permission the endpoint requires (`audit:read`, colon-form) could actually be minted into a token. No scope creep.

**Backend-only**, matches the issue (UI already exists on mock data, tracked separately in #5531).

**No Alembic migration**, correctly — `AuditTrail`/`SecurityEvent` and all columns used (`team_id`, `data_classification`, `user_email`) already exist.

**Shared-code blast radius**: `check_permission_inline` extraction is used by every `@require_permission`-decorated route, not just this one. Confirmed the extraction is behavior-preserving (same token-scope → plugin-hook → RBAC order, same 403 on deny), and there are explicit regression tests asserting the decorator still raises 403 / still calls through on grant post-refactor.

**Security**: restricted-classification rows are filtered in SQL (not app-layer), so counts aren't observable to non-admins — good pattern. NULL `team_id` on audit rows is treated as "actor-only visible," not public, avoiding a leak from bulk-import rows. Plugin-hook exceptions now re-raise instead of being swallowed into a 500.

**Tests**: excellent coverage — union ordering, tie-breaks on both sources, team scoping, public-only tokens, restricted-row exclusion (admin vs non-admin), `security:read` narrowing via both RBAC and token scope, error path, unauthenticated, missing `audit:read`, plus mapper unit tests. Covers every acceptance criterion in the issue.

Ran the new/changed tests, doctests, and ruff locally — all green.

One non-blocking suggestion for later: `check_permission_inline`'s own token-scope-gate branch doesn't have a direct unit test (only exercised indirectly through the decorator's existing suite) — worth adding for completeness, not required here.

Approving.

msureshkumar88 (approved):
Reviewed after rebasing locally onto latest `main` to confirm mergeability — clean rebase, only conflicts were two purely additive route-pattern/test entries (observability metrics + activity feed) with no logic overlap.

Scope matches the linked issue: `GET /api/logs/activity` unions AuditTrail + SecurityEvent, gates on `audit:read` (with `security:read` correctly additive, not a standalone gate), and the three follow-up commits address prior feedback — deterministic ordering, Layer 1 scope enforcement in the new `check_permission_inline` helper, and colon-form permission scopes in token minting.

Targeted test runs all green (token scoping, activity feed router, log search, schema validators, RBAC coverage, doctest examples on `rbac.py`) — no failures. No blocking or functionally-impacting findings. A couple of pre-existing, out-of-scope items noted for awareness only (not from this PR): some `== True/False` style nits in test files predating this change, and `schemas.py`'s permission regex still not accepting category-wildcard form — neither touched or regressed here.

Nice work tightening up the Layer 1 enforcement and ordering determinism from the earlier round. Approving.

PR #6293, opened 2026-08-18
Pr 6009 rebase

madhu-mohan-jaishankar:
Imported commits from https://github.com/IBM/mcp-context-forge/pull/6009 and addressed blocking changes.


mcpgateway/routers/observability.py
```diff
@@ -855,3 +862,95 @@ def percentile(data, p):
...
+    """Latency percentiles (ms) bucketed over time."""
+
+    buckets: list[str]
+    p50: list[float]
+    p95: list[float]
+    p99: list[float]
+
+
+@observability_metrics_router.get("/timeseries", response_model=TimeseriesResponse)
+@require_permission("metrics:read", allow_admin_bypass=False)
```
msureshkumar88: Both new endpoints (`/timeseries` here and `/percentiles` below) decorate with `@require_permission("metrics:read", allow_admin_bypass=False)` but don't pass `global_only=True`.

The migration comment for this permission says `metrics:read` must stay off team-scoped roles, since the aggregation query has no team filter. Right now that invariant is enforced only by which roles happen to hold the permission, not by the code itself — `require_permission`'s default `check_any_team=True` team-role aggregation (`rbac.py` `_resolve_team_and_check_mode`) means if a team-scoped role is ever granted `metrics:read` down the line, that would silently pass RBAC and return platform-wide execution/latency data to a team member.

Would it make sense to pass `global_only=True` explicitly on both decorators so the isolation guarantee lives in code rather than relying on role-assignment discipline?

mcpgateway/routers/observability.py
```diff
@@ -855,3 +862,95 @@ def percentile(data, p):
...
+        cutoff_time = datetime.now(timezone.utc) - timedelta(hours=hours)
+        data = service.get_execution_timeseries(db, cutoff_time, interval_minutes)
+        return TimeseriesResponse(**data)
+    except Exception as e:
+        logger.error(f"Failed to calculate timeseries metrics: {e}", exc_info=True)
+        raise HTTPException(status_code=500, detail="Failed to calculate timeseries metrics")
+
+
+@observability_metrics_router.get("/percentiles", response_model=PercentilesResponse)
+@require_permission("metrics:read", allow_admin_bypass=False)
```
msureshkumar88: Same as the `/timeseries` endpoint above — `global_only=True` would help lock in the team-isolation guarantee here too.

msureshkumar88 (changes requested):
Thanks for the follow-up work here — the UTC bucket-timestamp fix and the `is_active` filter on the role lookup migration are both genuinely in place and covered by tests, confirmed against the commits rather than just the PR description.

One thing worth addressing before merge:

msureshkumar88 (dismissed):
Re-reviewed the prior requested changes against the actual commits (not the review comments) — all three are genuinely implemented:

- `global_only=True` added to both `metrics:read` RBAC decorators (`d5b1c54bc`), with a regression test asserting `team_id is None` / `check_any_team is False` reach `PermissionService.check_permission`.
- UTC naive-timestamp labeling fix (`cedd2fb6d`), covered by a regression test.
- `is_active` filter on the role-lookup migration query (`80c6d1e86`), covered by a regression test.

Fresh review against issue #5945: the PR matches the issue's scope exactly (two read-only `/observability/metrics` endpoints, `metrics:read` RBAC, Layer-1 token-scoping, and a migration granting the permission to the `platform_viewer` default role for existing deployments). Backend-only change, no UI. The Alembic migration is necessary — without it, existing deployments can't reach the new endpoints without a manual role edit.

No blocking or functionally-impacting findings. Three non-blocking suggestions for a follow-up, if useful:

1. `mcpgateway/admin.py` and `mcpgateway/services/observability_service.py` now carry near-duplicate percentile/timeseries aggregation logic (Postgres + Python-fallback paths) — this PR already had to fix the same UTC bug in both places in one commit, so a shared helper would reduce future drift risk.
2. The new `/observability/metrics/*` (permission `metrics:read`) and the existing `/admin/observability/metrics/*` (permission `admin.system_config`) endpoints serve the same underlying trace data under different permission models, which could confuse auditors — worth a follow-up to document or reconcile.
3. The SQLite fallback path (`_execution_timeseries_python` / `_latency_percentiles_python`) loads the full matching row set for up to a 168-hour window with no cap — fine today since the router was previously admin-only, but worth bounding now that it's unconditionally mounted.

Approving.

a-effort (approved):
Nice work! 🚀

--
non-blocking follow up: 

The database update script writes in the role name platform_viewer directly, but that role name is a setting operators can change (config.py:1064), and it is read from that setting when a new user account is created (email_auth_service.py:743).

Example in this repository of taking another approach: (ba202ac1665f_migrate_user_roles_to_configurable_.py, lines 201, 211, 357) looks up the configured role name at run time and saves a copy of it alongside the change, so the change can be cleanly undone later without depending on settings that may have shifted since.

The new script does not follow it, which means:

- Any deployment that has changed the role name from the default receives no permission grant.
- The description, which says this is the role every non-admin user has, is wrong for those deployments.

msureshkumar88 (approved):

PR #6302, opened 2026-08-19
fix: recreate CI venv when cached Python symlink is broken

msureshkumar88:
## Summary
- `make venv` only checked that `.venv` existed, not that its `bin/python3` symlink still resolved
- The plugin-integration workflow caches `.venv` across runs; when the cached interpreter path no longer matches the runner (e.g. runner image/toolchain update), the symlink dangles and `install-dev` fails with `Broken symlink at .venv/bin/python3`
- Detect a dangling `bin/python3` symlink and wipe the venv before recreating it, so a stale cache self-heals instead of failing the job

Fixes the merge-queue CI failure in https://github.com/IBM/mcp-context-forge/actions/runs/32236429043/job/96017328757

## Test plan
- [x] Simulated a broken `.venv/bin/python3` symlink and ran `make venv` — confirmed it detects, removes, and recreates the venv
- [x] Ran `make venv` against the existing healthy `.venv` — confirmed it still skips recreation (no regression)

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^\\.secrets\\.baseline$",
     "lines": null
   },
-  "generated_at": "2026-08-17T13:30:58Z",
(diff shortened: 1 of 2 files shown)


prakhar-singh1928 (approved):

madhu-mohan-jaishankar (approved):
LGTM

Lang-Akshay (approved):
LGTM

PR #6308, opened 2026-08-19
fix(security): audit active gateway/tool/agent URLs against scheme allowlist on startup

### LINKED ISSUE #6264: [BUG]: VALIDATION_ALLOWED_URL_SCHEMES is a class-level constant — changes require full process restart and do not apply to existing database records
## Summary

`SecurityValidator.ALLOWED_URL_SCHEMES` is assigned once at class definition time from `settings.validation_allowed_url_schemes`:

```python
# mcpgateway/common/validators.py:365
ALLOWED_URL_SCHEMES = settings.validation_allowed_url_schemes
```

This has two related operational gaps:

## Gap 1 — Scheme change requires full process restart, not just worker reload

Because the assignment happens at module import time, the value is frozen into the class for the lifetime of the process. Changing `VALIDATION_ALLOWED_URL_SCHEMES` in the environment and sending Gunicorn a `SIGHUP` (worker reload) is not sufficient — the new value is only picked up after a full process restart. This is undocumented and likely to surprise operators who follow the pattern used for other settings.

## Gap 2 — Existing database records are not validated against the scheme allowlist

The allowlist is enforced only at registration time (Pydantic field validators on `GatewayCreate` and `GatewayUpdate`). Pre-existing `http://` or `ws://` gateway/tool/agent records in the database bypass the check entirely and continue to be used by:
- The health-check loop
- Active tool/resource invocation 
- The federation layer

This means setting `VALIDATION_ALLOWED_URL_SCHEMES='["https://","wss://"]'` does **not** enforce HTTPS-only connectivity for an existing deployment — it only prevents new registrations.

### Gap 1

Document the restart requirement clearly in:
- `.env.example` comment alongside `VALIDATION_ALLOWED_URL_SCHEMES`
- `docs/docs/operations/config-validation.md` or `tls-configuration.md`

Optionally, make the class property read from `settings` at call time rather than at class definition time, though this has a minor per-call overhead.

### Gap 2

Add a startup validation step that checks all active (enabled) gateway, tool, and agent URLs against the configured scheme allowlist and either:
- **Logs a WARNING** for each record that violates the policy (fail-open, recommended), or
- **Raises a startup error** if `STRICT_SCHEME_ENFORCEMENT=true` is set (fail-closed, opt-in)

This gives operators visibility into policy violations without breaking existing deployments on upgrade.

## Acceptance Criteria

- [ ] The restart requirement for `VALIDATION_ALLOWED_URL_SCHEMES` is documented
- [ ] On startup, active records with URL schemes not in the allowlist emit a WARNING log entry naming the offending record and URL
- [ ] The startup check is implemented in a way that does not break existing deployments (warn-only by default)
- [ ] A `STRICT_SCHEME_ENFORCEMENT` flag (default `false`) can be set to fail startup instead of warn

DearMoon50:
## Summary

Fixes #6264.

`SecurityValidator.ALLOWED_URL_SCHEMES` is enforced only at registration time (the Pydantic field validators on `GatewayCreate`/`GatewayUpdate` and equivalents). A record registered before the allowlist was tightened — e.g. an `http://` gateway created before `VALIDATION_ALLOWED_URL_SCHEMES` was narrowed to `https://` only — silently keeps being used by the health-check loop, tool invocation, and the federation layer, regardless of the current setting. Setting `VALIDATION_ALLOWED_URL_SCHEMES='["https://","wss://"]'` on an existing deployment therefore does *not* enforce HTTPS-only connectivity — it only prevents new registrations.

This addresses **Gap 2** from the issue (the substantive security gap) and documents **Gap 1** (the class-level-constant restart requirement) as the issue's own proposed changes suggested — making Gap 1 dynamically re-readable was flagged as "optional" in the issue and is left out of scope here to keep this reviewable.

### Changes

- `mcpgateway/common/validators.py` — extracted the scheme check into `SecurityValidator.has_allowed_scheme()` so `validate_url` (registration-time) and the new startup audit (existing-record time) enforce the exact same rule instead of two copies drifting apart.
- `mcpgateway/bootstrap_db.py` — added `audit_url_schemes()`, run from `_run_post_migration_bootstrap` on every startup (alongside the other idempotent bootstrap steps). It queries active (`enabled=True`) `Gateway`, `Tool`, and `A2AAgent` records and checks each against the current allowlist.
  - **Read-only by design**: it never mutates or disables an offending record — deciding what to do about a live integration is an operator call, not this function's to make silently.
  - **Warn-only by default**: logs one `WARNING` naming every offending record (type, name, id, URL), so upgrading to a stricter allowlist never breaks an existing deployment outright.
- `mcpgateway/config.py` — added `strict_scheme_enforcement: bool = False`. When `true`, `audit_url_schemes` raises `RuntimeError` instead of warning, so an operator can opt into fail-closed startup once they've reviewed the warnings (matches the issue's acceptance criteria: "A `STRICT_SCHEME_ENFORCEMENT` flag (default `false`) can be set to fail startup instead of warn").
- `.env.example` — documents the restart requirement for `VALIDATION_ALLOWED_URL_SCHEMES` and the new `STRICT_SCHEME_ENFORCEMENT` flag.

### Acceptance criteria (from the issue)

- [x] The restart requirement for `VALIDATION_ALLOWED_URL_SCHEMES` is documented
- [x] On startup, active records with URL schemes not in the allowlist emit a WARNING log entry naming the offending record and URL
- [x] The startup check is implemented in a way that does not break existing deployments (warn-only by default)
- [x] A `STRICT_SCHEME_ENFORCEMENT` flag (default `false`) can be set to fail startup instead of warn

## Test plan

- [x] `pytest tests/unit/mcpgateway/test_bootstrap_db.py` — 73 passed, including 5 new tests in `TestAuditUrlSchemes` covering: no active records, all-allowed (no warning), a disallowed scheme (warns, does not raise), `strict_scheme_enforcement=True` (raises `RuntimeError`), and coverage across all three record types (Gateway/Tool/A2AAgent)
- [x] `pytest tests/unit/mcpgateway/validation/test_validators.py tests/unit/mcpgateway/validation/test_validators_advanced.py` — all pass (`has_allowed_scheme` refactor is behavior-preserving for `validate_url`)
- [x] `ruff check` clean on all touched files
- [x] `black --check` clean on all touched files (pre-existing long-line formatting debt elsewhere in `config.py`/`validators.py` left untouched, out of scope)

## Was generative AI tooling used to co-author this PR?

- [X] Yes — Claude Code (Sonnet 5), reviewed and tested by me before submitting.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -3148,7 +3148,20 @@ PLUGINS_CLI_MARKUP_MODE=rich
 # Allowed URL schemes for external requests
 # Controls which URL schemes are permitted for gateway operations
 # Default: ["http://", "https://", "ws://", "wss://"]
+# NOTE: this is read once into a class-level constant at process start, so a
+# change here needs a full process restart to take effect -- a Gunicorn
+# worker reload (SIGHUP) is NOT sufficient. It is also enforced only at
+# registration time; existing gateway/tool/agent records already in the
+# database keep using whatever scheme they were created with (see
+# STRICT_SCHEME_ENFORCEMENT below to audit those on startup).
 # VALIDATION_ALLOWED_URL_SCHEMES=["http://", "https://", "ws://", "wss://"]
+
+# If true, fail startup when an active gateway/tool/agent record's URL uses a
+# scheme outside VALIDATION_ALLOWED_URL_SCHEMES (e.g. a leftover http:// record
+# after tightening the allowlist to https-only). Default false: such records
+# are logged as a WARNING instead, so upgrading to a stricter allowlist never
+# breaks an existing deployment outright.
+# STRICT_SCHEME_ENFORCEMENT=false
 
 # Character validation patterns (regex)
 # Used to validate various input fields
--- a/mcpgateway/bootstrap_db.py
+++ b/mcpgateway/bootstrap_db.py
@@ -162,15 +162,58 @@ def alembic_at_head(conn: Connection, cfg: Config) -> bool:
         return False
 
 
+def audit_url_schemes(conn: Connection) -> None:
+    """Audit active gateway/tool/A2A-agent URLs against the configured scheme allowlist.
+
+    ``SecurityValidator.ALLOWED_URL_SCHEMES`` is enforced only at registration
+    time (the Pydantic field validators on create/update schemas), so a record
+    that was registered before the allowlist was tightened — e.g. an
+    ``http://`` gateway created before ``VALIDATION_ALLOWED_URL_SCHEMES`` was
+    narrowed to ``https://`` only — keeps being used by health checks, tool
+    invocation, and federation regardless of the current setting. This makes
+    that gap visible on every startup instead of leaving it undetected.
+
+    Read-only by design: this only logs (or, under ``strict_scheme_enforcement``,
+    raises) — it never mutates or disables an offending record, since deciding
+    what to do about a live integration is an operator call, not this
+    function's to make silently.
+
+    Args:
+        conn: Active SQLAlchemy connection.
+
+    Raises:
+        RuntimeError: If ``settings.strict_scheme_enforcement`` is True and at
+            least one active record uses a scheme outside the allowlist.
+    """
+    violations: list[str] = []
+    with Session(bind=conn) as session:
+        for model, url_attr, label in ((Gateway, "url", "gateway"), (Tool, "url", "tool"), (A2AAgent, "endpoint_url", "a2a_agent")):
+            records = session.query(model).filter(model.enabled.is_(True)).all()
+            for record in records:
+                url = getattr(record, url_attr, None)
+                if not url:
+                    continue
+                if not SecurityValidator.has_allowed_scheme(url):
+                    violations.append(f"{label} '{record.name}' (id={record.id}): {url}")
+
+    if not violations:
+        return
+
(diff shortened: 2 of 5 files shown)


Lang-Akshay (changes requested):
Thanks for the PR @DearMoon50 
Worth fixing this before merging it in the main 
| # | File | Line | Severity | CWE | Description | Fix |
| - | ---- | ---- | -------- | --- | ----------- | --- |
| 1 | `mcpgateway/bootstrap_db.py` | 197, 202-205 | Medium | CWE-532; secondary CWE-117 | The audit embeds record names and raw stored URLs in the warn-only log and strict-mode exception. Legacy URL userinfo/query credentials can be exposed, and newline/control characters can forge log entries. | Sanitize URL/name fields with existing logging helpers, or emit only non-sensitive scheme/host metadata. |
| 2 | `mcpgateway/config.py` | 3620; `mcpgateway/bootstrap_db.py:203-205` | Medium | CWE-693 | Strict enforcement is opt-in and warn-only is the default. This is an intentional compatibility mode, not a regression, but it is fail-open when operators expect the allowlist itself to block existing records. | Enable strict mode where the allowlist is a security boundary, or make fail-closed behavior the production default with an explicit compatibility opt-out. |

Lang-Akshay:
Hi @DearMoon50,  thanks for starting this. Since this addresses a security-sensitive issue, we can't leave it open while it waits on the requested changes. I'm going to take it over and implement the fix separately so we can get it resolved promptly. I'll close this PR once the replacement is merged.

Appreciate you raising it and getting the work started.

PR #6320, opened 2026-08-20
feat: Add experimental UI to compose

gcgoncalves:
- Adds the web ui client to docker compose
- Creates new compose profile
- Extend compose configuration from the env file (new overrides)


marekdano (commented):
Two confirmed correctness bugs in `docker-compose.yml`'s new `web_ui` service. Neither breaks the default path (`WEB_UI_PORT` unset resolves fine to `3000:3000`), but both undermine the point of the change.

  ### 1. Port mismatch when `WEB_UI_PORT` is overridden — `docker-compose.yml:1184`

  The compose file maps `${WEB_UI_PORT:-3000}:3000` but also sets the container's `PORT` env var to `${WEB_UI_PORT:-3000}`:

  ```yaml
  ports:
    - "${WEB_UI_PORT:-3000}:3000"
  environment:
    PORT: ${WEB_UI_PORT:-3000}
  ```

  The app listens on whatever `PORT` is set to, not a fixed 3000:
  - `server/src/config.ts:24` — `config.port = Number(process.env.PORT ?? "3000")`
  - `server/src/index.ts:53` — `fastify.listen({ port: config.port })`

  So if a user sets `WEB_UI_PORT=3001`, the mapping becomes `3001:3000` (host 3001 → container 3000), but the app now listens on 3001 *inside* the container. Nothing listens on container port 3000 anymore, so the published port is unreachable — even though the container's own `HEALTHCHECK` still passes internally (it also reads `PORT`).

  **This breaks the entire purpose of making the port configurable.**

  ### 2. Default port collision with `grafana` — `docker-compose.yml:1184`

  `grafana` already publishes `"${GRAFANA_PORT:-3000}:3000"` under `profiles: [monitoring]` (line 1424, pre-existing). This PR adds `web_ui` publishing `"${WEB_UI_PORT:-3000}:3000"` under `profiles: [testing, experimental]` — same default host port.

  The `Makefile` already combines multiple profiles in one `docker compose up` (e.g. `--profile sso --profile monitoring`). Running `docker compose --profile experimental --profile monitoring up -d` hits:

  ```
  Bind for 0.0.0.0:3000 failed: port is already allocated
  ```

  and the second service fails to start.

docker-compose.yml
```diff
@@ -1145,6 +1149,54 @@ services:
           cpus: '1'
           memory: 1G
 
+###############################################################################
+#  CONTEXTFORGE WEB UI (enabled with --profile testing or --profile experimental)
+#  Usage: docker compose --profile experimental up -d
+#  Access: http://localhost:${WEB_UI_PORT:-3000}
```
marekdano:
```suggestion
#  Access: http://localhost:${WEB_UI_PORT:-3001}
```

marekdano (dismissed):
Approving with small comment! 

🚀

marekdano (dismissed):
LGTM 🚀

marekdano (changes requested):
Solid PR overall — parameterizing the admin credentials and adding the web UI in its own profile with proper health-check `depends_on` are both done correctly. A few things to address before merge:

  ### Blocking
  - **`web_ui_redis` image** — uses `redis:latest`. If the registry.redhat.io policy applies to new services, this should be `registry.redhat.io/rhel9/redis-7:latest` (or similar) rather than Docker Hub. Note the existing gateway `redis` service (line ~1070) also uses `redis:latest`, so please confirm whether that's pre-existing/grandfathered or whether this PR is expected to fix both.
  - **`web_ui` image** — same question applies to `WEB_UI_IMAGE` (`ghcr.io/contextforge-org/contextforge-web-ui:latest`), if the policy covers app images too, not just the Redis sidecar.

  ### Nits / polish
  - **`HOST=${WEB_UI_HOST:-0.0.0.0}`** — already correctly overridable, just add a short comment like the gateway has (line ~145: `# Bind address - must be 0.0.0.0 inside Docker or other containers cannot connect`) so future reviewers don't flag it as a bind-to-all-interfaces issue.
  - **`.env.example`** — no `WEB_UI_PORT`, `WEB_UI_HOST`, `WEB_UI_IMAGE`, `WEB_UI_CONTEXTFORGE_URL`, `WEB_UI_COOKIE_SECURE`, or `WEB_UI_REDIS_URL` entries. Add a commented block matching the pattern used for other optional services, for discoverability.
  - **Port mapping** — `"${WEB_UI_PORT:-3001}:${WEB_UI_PORT:-3001}"` ties host and container port together. If the image genuinely reads `PORT` at startup (looks like it does, via the `PORT=${WEB_UI_PORT:-3001}` env var), this is fine — consider a one-line comment confirming that, since it's a slightly unusual pattern vs. a fixed container-side port.
  - **Profile header comment** — line ~19 (`--profile testing: Adds Locust (web UI), ...`) now reads ambiguously next to the new note that `web_ui` is also included in `--profile testing`. Consider updating that line to mention the ContextForge web UI explicitly, so "(web UI)" isn't read as referring to this new service.

Thanks for adding this — the profile/health-check wiring is clean and the dedicated Redis for session storage is the right call.

marekdano (dismissed):
Solid PR!

LGTM 🚀

marekdano (dismissed):

Lang-Akshay (dismissed):
LGTM ✅

Lang-Akshay (approved):
LGTM

PR #6216, opened 2026-08-13
feat: migration script for auth_secret

prakhar-singh1928:
# Pull Request

## 🔗 Related Issue

Closes #6113

---

## 📝 Summary

Adds `mcpgateway/scripts/migrate_enc_secret.py` — a one-shot re-encryption script for operators upgrading from 1.0.7 → 1.0.8.

Release 1.0.7 (PR #6073) relaxed `AUTH_ENCRYPTION_SECRET` enforcement in `ENVIRONMENT=development`, allowing weak values like `my-test-salt` to start the gateway. Release 1.0.8 restores the unconditional hard-fail across all environments. Operators who ran 1.0.7 with a weak secret face two silent breaking changes on upgrade:

1. **Startup failure** — the gateway refuses to start because the weak value no longer passes the guardrail.
2. **Silent decryption failure** — stored credentials were encrypted under the old weak key and will fail to decrypt under the new strong key.

This PR ships the migration script that handles the re-encryption pass so the upgrade path is safe and documented.

**What the script does:**
- Accepts `--old-key` / `--new-key` (or `OLD_AUTH_ENCRYPTION_SECRET` / `NEW_AUTH_ENCRYPTION_SECRET` env vars)
- Re-encrypts all affected columns across both encryption paths:
  - **EncryptionService path** (`v2:{...}` format): `oauth_tokens`, `registered_oauth_clients`, `sso_providers`, and `oauth_config` JSON in `gateways`, `a2a_server_auth`, `a2a_agent_auth`
  - **services_auth path** (AES-GCM base64url blobs): `tools.auth_value`, `a2a_agents.auth_value`/`auth_query_params`, `a2a_agent_auth.auth_value`/`auth_query_params`, `llm_providers.api_key`, `gateways.auth_query_params`
- Runs the entire migration in a single transaction — rolled back atomically on any error
- Is idempotent: running it twice is safe; already-migrated values are detected and skipped
- Supports `--dry-run` to report what would be migrated without writing
- Validates `--new-key` strength before touching the database (length, entropy, known-weak list)
- Skips tables that don't exist (optional features not deployed)

Also adds `docs/upgrade/auth-encryption-secret-1.0.7-to-1.0.8.md` — a step-by-step operator upgrade guide covering all deployment types (Docker Compose, Kubernetes, Helm, systemd, bare-metal).

---

## 📏 Reviewability

- [x] This PR has one clear purpose
- [x] The linked issue is not labeled `triage`
[...]

PR #6324, opened 2026-08-20
fix: add missing migration columns to migrate_enc_secret 

### LINKED ISSUE #6323: BUG]: migrate_enc_secret.py missing gateways.auth_value — MCP Servers disappear from UI after AUTH_ENCRYPTION_SECRET rotation
### 🐞 Bug Summary

`migrate_enc_secret.py` (shipped in #6216) does not re-encrypt `gateways.auth_value`. Gateway rows registered with `auth_type=bearer`, `basic`, or `authheaders` store a `services_auth` AES-GCM blob in that column. After running the migration script, those blobs remain encrypted under the old key, causing the gateway list endpoint (`GET /admin/gateways/partial`) to return HTTP 500 and MCP Servers to disappear from the UI.

---

### 🧩 Affected Component

- [x] `mcpgateway` - API

---

### 🔁 Steps to Reproduce

1. Register one or more MCP Servers (gateways) with `auth_type=bearer` (or `basic`/`authheaders`) in a 1.0.7 deployment using a weak `AUTH_ENCRYPTION_SECRET`.
2. Upgrade to 1.0.8, set a new strong `AUTH_ENCRYPTION_SECRET`, and run the migration script:
   ```bash
   python -m mcpgateway.scripts.migrate_enc_secret \
       --old-key <old-weak-key> \
       --new-key <new-strong-key>

3. Start the 1.0.8 gateway.
4. Open the Admin UI → MCP Servers tab, or call GET /admin/gateways/partial.

### 🤔 Expected Behavior

MCP Servers remain visible in the UI and the endpoint returns HTTP 200. All `gateways.auth_value` blobs should have been re-encrypted under the new key by the migration script.

---

### 📓 Logs / Error Output

HTTP 500 on GET /admin/gateways/partial

ExceptionGroup: unhandled errors in a TaskGroup (1 sub-exception)
File "starlette/middleware/errors.py", line 164, in call
await self.app(scope, receive, _send)
...

AES-GCM tag verification failure when `decode_auth()` is called on a `gateways.auth_value` blob still encrypted under the old key.

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Version or commit | `main` (post #6216) |
| Container | Docker |

---

### 🧩 Additional Context

**Root cause:** `GatewayCreate._process_auth_fields()` in `schemas.py` calls `encode_auth()` and stores a `services_auth` blob string into `GatewayCreate.auth_value`. That string is passed through `_prepare_gateway_registration()` without being decoded back to a dict, and is persisted as-is into `DbGateway.auth_value` (a JSON column — SQLAlchemy stores and retrieves a string faithfully through a JSON column). The migration script's `sa_simple_targets` list at line 752 of `migrate_enc_secret.py` covers `tools`, `a2a_agents`, `a2a_agent_auth`, and `llm_providers` but is missing `("gateways", "id", ["auth_value"])`.

**Fix:** Add `("gateways", "id", ["auth_value"])` to `sa_simple_targets`. Because the script is idempotent, operators who already ran the migration can simply re-run it with the same keys — already-migrated values are detected and skipped automatically.

**Related:** #6216, #6113

prakhar-singh1928:
# Pull Request

## 🔗 Related Issue

Closes [#6323](https://github.com/IBM/mcp-context-forge/issues/6323)

---

## 📝 Summary

`migrate_enc_secret.py` (shipped in #6216) had three gaps in its encrypted column coverage, discovered during a thorough audit against `db.py` and the service layer:

**1. `gateways.auth_value` missing from `sa_simple_targets`**
`GatewayCreate._process_auth_fields()` calls `encode_auth()` and stores a `services_auth` AES-GCM blob string into `gateways.auth_value` for `bearer`, `basic`, and `authheaders` auth types. Because SQLAlchemy stores and retrieves a string faithfully through a JSON column, those blobs are encrypted under `AUTH_ENCRYPTION_SECRET` at rest — but the migration script never re-encrypted them. After a rotation, `decode_auth()` fails on every gateway row with an AES-GCM tag verification error, causing HTTP 500 on `GET /admin/gateways/partial` and MCP Servers disappearing from the UI.

**2. `a2a_push_notification_configs.auth_token` missing from `sa_simple_targets`**
`a2a_service.py` stores `encode_auth({"token": raw_auth_token})` as a `services_auth` blob for webhook bearer credentials. This column was not covered.

**3. `a2a_server_auth` replaced with `servers` in `json_targets`**
`a2a_server_auth` does not exist as a table in `db.py` — it was a phantom reference silently skipped by the missing-table guard. `servers.oauth_config` is the correct table: `server_service.py` calls `protect_oauth_config_for_storage()` when creating/updating virtual servers, which encrypts sensitive OAuth keys via the EncryptionService path.

All three fixes were validated end-to-end by seeding a SQLite database with real encrypted values under the old key across all 17 affected columns, running the migration script, and confirming every value decrypts correctly under the new key and fails under the old one.

Operators who already ran the migration can re-run it with the same keys — the idempotency check skips already-migrated values automatically.

---

## 📏 Reviewability

---

## 🏷️ Type of Change

- [x] Bug fix

---

## 🧪 Verification

**Regression test added:** `test_migrates_gateways_auth_value` in `tests/unit/mcpgateway/scripts/test_migrate_enc_secret.py` — seeds a `gateways` row with a bearer blob encrypted under `OLD_KEY`, runs the migration, and asserts the row decrypts correctly under `NEW_KEY`.

**Manual end-to-end validation:** seeded all 17 affected columns across 10 tables with real encrypted values under the old key, ran the migration script, verified all 17 values decrypt correctly under the new key and fail under the old one (confirming re-encryption actually occurred, not just a passthrough).

**For affected operators:** re-run the migration script with the same keys used previously — already-migrated values are skipped, only the newly covered columns will be re-encrypted.

---

## ✅ Checklist

---

## 📓 Notes (optional)

The misleading comment at `gateway_service.py:1744` — *"DbGateway.auth_value is Mapped[Optional[Dict]] (JSON) and receives the plain dict directly"* — is only true when `auth_value` originates from `auth_headers`. For `bearer`/`basic`/`authheaders` coming through the schema validator, `auth_value` is already an `encode_auth()` blob string and is stored as-is. That comment is a separate cleanup tracked in the linked issue.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^\\.secrets\\.baseline$",
     "lines": null
   },
-  "generated_at": "2026-08-20T12:49:28Z",
+  "generated_at": "2026-08-20T14:52:39Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -4000,7 +4000,7 @@
         "hashed_secret": "e582429052cdb908833560f6f7582d232de37c4d",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 193,
+        "line_number": 197,
         "type": "Base64 High Entropy String",
         "verified_result": null
       }
@@ -4764,7 +4764,7 @@
         "hashed_secret": "cd60be54aa0dff795929c4bbeda21e12e308dffa",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 607,
+        "line_number": 608,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -4772,7 +4772,7 @@
         "hashed_secret": "d4232fcd7edae0351f456330c3d48f521238f7d0",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1141,
+        "line_number": 1166,
         "type": "Secret Keyword",
         "verified_result": null
       }
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^\\.secrets\\.baseline$",
     "lines": null
   },
-  "generated_at": "2026-08-20T14:52:39Z",
+  "generated_at": "2026-08-20T15:25:05Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -2790,7 +2790,7 @@
         "hashed_secret": "57f65122c6781643b491d61282eb46165d623e00",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 210,
+        "line_number": 217,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -2798,7 +2798,7 @@
         "hashed_secret": "339e84bf584820b688e7dfde3e7ab791c741f5dc",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 211,
+        "line_number": 218,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -2806,7 +2806,7 @@
         "hashed_secret": "93307f19d549fec0f3909b77691bdbb51312bce2",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 359,
+        "line_number": 386,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -2814,7 +2814,7 @@
         "hashed_secret": "73228b9a9a7e40debe7a7026e28a3c1ddfed9277",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 360,
+        "line_number": 387,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -2822,7 +2822,7 @@
         "hashed_secret": "1e2ee220fb6c23833620f8cc55377d42e9050b66",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 385,
+        "line_number": 412,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -2830,7 +2830,7 @@
         "hashed_secret": "164a5c24f5df5918f285228443ab87209785293f",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 464,
+        "line_number": 491,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -2838,7 +2838,7 @@
         "hashed_secret": "040aee8abe52741790f15c77cc979bbd1561dd2e",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 468,
+        "line_number": 495,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -2846,7 +2846,7 @@
         "hashed_secret": "ec122f54cf16a175b3e22979e240684f7734079a",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 468,
+        "line_number": 495,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -2854,7 +2854,7 @@
         "hashed_secret": "00f798239d1a436b1bf7b3cf922380585d20ed2d",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 487,
+        "line_number": 514,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -2862,7 +2862,7 @@
         "hashed_secret": "fe0bb96310c2ef752dc38cd95bda55a2c3c0ed71",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 574,
+        "line_number": 601,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -2870,7 +2870,7 @@
         "hashed_secret": "9d4e1e23bd5b727046a9e3b4b7db57bd8d6ee684",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 673,
+        "line_number": 700,
         "type": "Basic Auth Credentials",
         "verified_result": null
       },
@@ -2878,7 +2878,7 @@
         "hashed_secret": "905ce622fe4c3829dab05a4ef721c157698c302a",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 725,
+        "line_number": 752,
         "type": "Secret Keyword",
         "verified_result": null
       }
@@ -4000,7 +4000,7 @@
         "hashed_secret": "e582429052cdb908833560f6f7582d232de37c4d",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 197,
+        "line_number": 203,
         "type": "Base64 High Entropy String",
         "verified_result": null
       }
@@ -4764,7 +4764,7 @@
         "hashed_secret": "cd60be54aa0dff795929c4bbeda21e12e308dffa",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 608,
+        "line_number": 628,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -4772,7 +4772,7 @@
         "hashed_secret": "d4232fcd7edae0351f456330c3d48f521238f7d0",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1166,
+        "line_number": 1279,
         "type": "Secret Keyword",
         "verified_result": null
       }
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^\\.secrets\\.baseline$",
     "lines": null
   },
-  "generated_at": "2026-08-21T09:00:50Z",
+  "generated_at": "2026-08-21T10:23:38Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -4764,7 +4764,7 @@
         "hashed_secret": "cd60be54aa0dff795929c4bbeda21e12e308dffa",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 631,
+        "line_number": 633,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -4772,7 +4772,7 @@
         "hashed_secret": "d4232fcd7edae0351f456330c3d48f521238f7d0",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1367,
+        "line_number": 1569,
         "type": "Secret Keyword",
         "verified_result": null
       }
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^\\.secrets\\.baseline$",
     "lines": null
   },
-  "generated_at": "2026-08-20T15:25:05Z",
+  "generated_at": "2026-08-21T09:00:50Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -4000,7 +4000,7 @@
         "hashed_secret": "e582429052cdb908833560f6f7582d232de37c4d",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 203,
+        "line_number": 215,
         "type": "Base64 High Entropy String",
         "verified_result": null
       }
@@ -4764,7 +4764,7 @@
         "hashed_secret": "cd60be54aa0dff795929c4bbeda21e12e308dffa",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 628,
+        "line_number": 631,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -4772,7 +4772,7 @@
         "hashed_secret": "d4232fcd7edae0351f456330c3d48f521238f7d0",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1279,
+        "line_number": 1367,
         "type": "Secret Keyword",
         "verified_result": null
       }
--- a/docs/docs/operations/auth-encryption-secret-rotation.md
+++ b/docs/docs/operations/auth-encryption-secret-rotation.md
@@ -59,18 +59,25 @@ The rotation script touches exactly these columns — everything else is untouch
 | `registered_oauth_clients` | `client_secret_encrypted`, `registration_access_token_encrypted` | Direct AES ciphertext |
 | `sso_providers` | `client_secret_encrypted` | Direct AES ciphertext |
 | `gateways` | `oauth_config` | JSON — sensitive keys re-encrypted recursively |
-| `a2a_server_auth` | `oauth_config` | JSON — sensitive keys re-encrypted recursively |
+| `servers` | `oauth_config` | JSON — sensitive keys re-encrypted recursively |
 | `a2a_agent_auth` | `oauth_config` | JSON — sensitive keys re-encrypted recursively |
 
 **services_auth path** (`base64url(nonce+ciphertext)` format — tool/agent/LLM credentials):
 
 | Table | Column(s) | Type |
 |-------|-----------|------|
+| `gateways` | `auth_value` | AES-GCM blob (JSON-quoted scalar — see note below) |
 | `tools` | `auth_value` | AES-GCM blob |
 | `a2a_agents` | `auth_value`, `auth_query_params` | AES-GCM blob / JSON dict of blobs |
 | `a2a_agent_auth` | `auth_value`, `auth_query_params` | AES-GCM blob / JSON dict of blobs |
+| `a2a_push_notification_configs` | `auth_token` | AES-GCM blob |
 | `gateways` | `auth_query_params` | JSON dict of AES-GCM blobs |
 | `llm_providers` | `api_key` | AES-GCM blob |
+
+> **Note — `gateways.auth_value` storage format:** this column is `mapped_column(JSON)` in `db.py`, so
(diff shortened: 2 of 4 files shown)


msureshkumar88 (changes requested):
Thanks for chasing this down — the root-cause analysis in the description is excellent, and the phantom `a2a_server_auth` catch is a genuinely good find. I verified both of the secondary fixes independently and they're correct:

- `a2a_server_auth` → `servers`: `grep -rn "a2a_server_auth" --include=*.py .` returns zero hits, so it really was being swallowed by the missing-table guard. `servers.oauth_config` is the right replacement (`server_service.py:584`, `:1326-1328`), and `_migrate_json_columns` already handles the `str`-vs-`dict` ambiguity and writes `json.dumps()`, so the JSON column type is handled correctly on that path.
- `a2a_push_notification_configs.auth_token`: confirmed `encode_auth({"token": ...})` at `a2a_service.py:3456`/`:3534`, and `db.py:5203` declares it `mapped_column(Text)` — so `sa_simple_targets` is the right list for it.

I do want to flag one blocking problem with the headline fix before this lands, plus a test-fixture issue that's masking it.

---

## Blocking: the `gateways.auth_value` fix is a no-op on SQLite

`sa_simple_targets` is handled by `_migrate_services_auth_simple_columns()`, which reads via raw `text("SELECT ...")` and writes a bare Python `str`. That's correct for every other entry in the list — `tools.auth_value`, `a2a_agents.auth_value`, `a2a_agent_auth.auth_value`, `llm_providers.api_key`, and the new `a2a_push_notification_configs.auth_token` are all `mapped_column(Text)`.

`gateways.auth_value` isn't. `mcpgateway/db.py:4769`:

```python
auth_value: Mapped[Optional[Dict[str, str]]] = mapped_column(JSON)
```

SQLAlchemy's `JSON` type `json.dumps()`-es on write, so the blob is persisted *with JSON quotes*. A raw-SQL `SELECT` returns those quotes, `_is_services_auth_blob()` rejects `"` (not in the base64url alphabet), and the value is classified `skipped_plaintext`.

Reproduced against a table declared with the real `db.py` column type:

```
RAW-IN-DB: '"bVJMFFWgWgN0cjnwfPx6Mypb-0kgmkzjUGx3_ocDv4ZW...Q"'   ← note the quotes
  Migrating gateways (services_auth) [auth_value] ...
    → found=1  migrated=0  skipped=1  errors=0
MIGRATION SUMMARY:  Values migrated: 0   Errors: 0
ℹ️   Nothing to migrate — all values are already up to date.
rc= 0
decode NEW FAILED: InvalidTag
decode OLD:        {'Authorization': 'Bearer gateway-bearer-tok'}
```

The blob is still under the old key, and the operator is told there was nothing to do — so the HTTP 500 on `GET /admin/gateways/partial` from #6323 persists after running the patched script.

Worth calling out the security shape of this one, because it's the reason I'd treat it as blocking rather than a follow-up: a value that is definitionally ciphertext is being misclassified as plaintext and dropped with no diagnostic (CWE-390 / CWE-703, fail-open). The documented operator workflow after a clean `rc == 0` is to destroy the old `AUTH_ENCRYPTION_SECRET` — at which point every `gateways.auth_value` blob is permanently undecryptable, with no recovery short of re-entering each upstream gateway credential by hand. For a key-rotation tool the safe posture is fail-closed.

**One trap for whoever picks up the fix:** it can't be read-side only. A bare unquoted string written back into the `JSON` column corrupts it on the next ORM read:

```
INSERT ... auth_value = 'rawblobnotjson'   # unquoted
ORM READ FAILED: JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

Read and write both need to round-trip through the JSON codec.

A couple of approaches, in the order I'd lean:

1. **Let SQLAlchemy do the coercion.** Reflect the table (`Table(name, MetaData(), autoload_with=engine)`) and run the select/update through Core instead of `text()`. Serialization is then handled per-dialect automatically, which also covers PostgreSQL (see below) and inoculates any future JSON-typed target.
2. **A dedicated target list.** `sa_json_scalar_targets = [("gateways", "id", ["auth_value"])]` with a helper that `json.loads()` a `str` payload, re-encrypts the inner scalar, and writes `json.dumps()` back — mirroring how `_migrate_json_columns` and `_migrate_services_auth_json_columns` already handle this. Fits the file's existing structure if you'd prefer the smaller diff.

Independently of which you pick, it'd be worth closing the fail-open: flag `sa_simple_targets` entries as "expected encrypted" and log a `WARNING` (and possibly exit non-zero) when a non-NULL, non-empty value in such a column doesn't match the blob shape.

## Blocking: the new test can't catch this

`tests/unit/mcpgateway/scripts/test_migrate_enc_secret.py:194` — `_make_sa_db` hand-writes its schema in raw SQL, and it's drifted from `db.py` on exactly the column this PR is about:

```sql
CREATE TABLE IF NOT EXISTS gateways (
    id TEXT PRIMARY KEY,
    auth_value TEXT,            -- db.py declares JSON
    auth_query_params TEXT
)
```

`test_migrates_gateways_auth_value` inserts an unquoted blob into a `TEXT` column — a shape production never produces — so it passes green against the broken path. The whole file is green (77 passed), which is why this got through.

Building the fixture from real column types (importing the models and `create_all()`-ing the subset, or at minimum `Column("auth_value", JSON)` inserted through Core) would make the current test fail against the current patch, which is the regression test #6323 actually needs.

## Functionally impacting

**PostgreSQL is unverified.** The repro above is SQLite. On PostgreSQL `gateways.auth_value` is a `json` column and psycopg deserializes on read, so the read path may work — but the write path pushes a bare `str` into a `json` column via `text()`, which is driver/cast-dependent. Could you add explicit PostgreSQL verification? A rotation tool that behaves differently per backend is uncomfortable to ship.

**The operator runbook still has the old list.** `docs/docs/operations/auth-encryption-secret-rotation.md` documents the phantom `a2a_server_auth` (line 62), and is missing `servers`/`oauth_config`, `gateways`/`auth_value` (line 72 lists only `auth_query_params`), and `a2a_push_notification_configs`/`auth_token`. The module docstring got updated but not the published runbook, so operators reading the docs still get the wrong list. Would be good to ship in the same PR.

While in that file, two additions worth considering given the failure mode above: an explicit post-rotation verification step (hit `GET /admin/gateways/partial`, confirm 200), and a "don't destroy the old key until you've verified" warning. Right now the only success signal is the script's own exit code.

## Suggestions (non-blocking)

- **Tests for the other two new targets.** Neither `servers.oauth_config` nor `a2a_push_notification_configs.auth_token` has a regression test.
- **Assert the old key now fails.** The new test checks `decode_auth(row, NEW_KEY) == payload` but never asserts the value fails under `OLD_KEY` — without that, a pure passthrough would still pass. You describe doing exactly this check manually in the PR body; encoding it in the test would be a nice win.
- **Idempotency test for the new columns.** Running twice should report `skipped_already_new`, not `error:decrypt_failed` — this matters because the PR explicitly tells affected operators to re-run.
- **A drift guard.** A test that walks `db.py` metadata and asserts every column in the four target lists exists (and that `sa_simple_targets` entries are all non-JSON types) would have caught both this PR's defect and the phantom table it removes, cheaply and permanently.
- **Split the `skipped` counter.** `skipped_null`, `skipped_already_new` and `skipped_plaintext` all fold into one number, which is why the manual run looked clean. Reporting them separately in the summary would have surfaced this immediately.
- **Per-table scan grouping.** `gateways` is now visited three times (`json_targets`, `sa_simple_targets`, `sa_json_targets`) — three full `SELECT`/`fetchall()` passes and up to three `UPDATE`s per row. Pre-existing, and this PR only nudges it, but grouping targets by table would cut it to one pass. Relatedly, `total["found"]` sums per-target row counts, so one gateway row reports `Rows scanned: 3` — a little misleading since that's the number operators eyeball to sanity-check a run.

## Minor

- The title mentions only `gateways.auth_value`, but the branch delivers three fixes — the audit framing from the second commit message is more accurate for the changelog.
- #6323 only covers the first fix; `Closes #6323` slightly over-claims the other two. Either extend the issue text or reference a follow-up.
- The second commit has no issue reference (the first has `Fixes #6323`).

## Things that look right

For the record, a few things I checked and found sound: no secret values reach any log line (`logger.error` emits only table/column/id/status); transaction handling is correct (single session, `commit()` only when `errors == 0`, `rollback()` on error and on `--dry-run`, `session.close()` in `finally`); the `# nosec B608` f-string SQL uses module-level literals only, so it isn't exploitable (and approach 1 above would remove the pattern entirely); and no Alembic change is needed here — this is a data-rewriting rotation utility, not a schema change, and it shouldn't become a migration since Alembic has no access to the old key.

Happy to re-review once the JSON round-trip and the fixture are sorted.

msureshkumar88 (dismissed):
Both blocking findings resolved:

- `gateways.auth_value` JSON round-trip: `_migrate_services_auth_json_scalar_columns()` now `json.loads()`/`json.dumps()`-wraps the blob correctly, moved into a dedicated `sa_json_scalar_targets` list. Verified the fix.
- Test fixture: `_make_sa_db` now declares `auth_value` as `JSON` (matching `db.py`), inserts via `json.dumps()` — the test would have failed against the pre-fix code.

Docs runbook updated (correct column table, Step 7 verification, old-key retention warning) and new regression tests added for `servers.oauth_config` and `a2a_push_notification_configs.auth_token`, each asserting the old key fails post-migration.

One non-blocking follow-up worth tracking: PostgreSQL write-path behavior for the JSON-scalar helper is reasoned about in the commit message but not exercised by an actual test against the `postgresql` dialect. Wouldn't block merge on it, but a follow-up PR adding that coverage (or a documented manual verification) would close the loop.

Approving.

msureshkumar88 (dismissed):

Lang-Akshay (changes requested):
# PR #6324 — `AUTH_ENCRYPTION_SECRET` Encrypted-Column Coverage Review

## Columns encrypted by `AUTH_ENCRYPTION_SECRET`

Two encryption codecs both key off `settings.auth_encryption_secret`:

- **EncryptionService** (`encrypt_secret`, `v2:{…}` envelope; also the `EncryptedText` ORM type) — OAuth/SSO/DCR secrets.
- **services_auth** (`encode_auth`, base64url AES-GCM blob) — `auth_value`/`auth_token`/`api_key` and nested-secret JSON.

Ground-truth DB sinks (verified in `db.py` + service write paths) vs. what `migrate_enc_secret.py` re-encrypts:

| # | Table.column | Codec | Script list | Covered |
|---|---|---|---|---|
| 1 | `oauth_tokens.access_token` | EncSvc (EncryptedText) | simple_targets | ✅ |
| 2 | `oauth_tokens.refresh_token` | EncSvc (EncryptedText) | simple_targets | ✅ |
| 3 | `registered_oauth_clients.client_secret_encrypted` | EncSvc | simple_targets | ✅ |
| 4 | `registered_oauth_clients.registration_access_token_encrypted` | EncSvc | simple_targets | ✅ |
| 5 | `sso_providers.client_secret_encrypted` | EncSvc | simple_targets | ✅ |
| 6 | `gateways.oauth_config` (nested `client_secret`) | EncSvc | json_targets | ✅ |
| 7 | `servers.oauth_config` | EncSvc | json_targets | ✅ (added by PR) |
| 8 | `a2a_agent_auth.oauth_config` | EncSvc | json_targets | ✅ |
| 9 | **`a2a_agents.oauth_config`** | EncSvc (`protect_oauth_config_for_storage`, a2a_service.py:673,1515) | — | ❌ **MISSING** |
| 10 | `tools.auth_value` | services_auth | sa_simple_targets | ✅ |
| 11 | `a2a_agents.auth_value` | services_auth | sa_simple_targets | ✅ |
| 12 | `a2a_agent_auth.auth_value` | services_auth | sa_simple_targets | ✅ |
| 13 | `a2a_push_notification_configs.auth_token` | services_auth | sa_simple_targets | ✅ (added by PR) |
| 14 | `llm_providers.api_key` | services_auth | sa_simple_targets | ✅ |
| 15 | `gateways.auth_value` | services_auth (JSON scalar) | sa_json_scalar_targets | ✅ (**the #6323 fix**) |
| 16 | `gateways.auth_query_params` | services_auth | sa_json_targets | ✅ |
| 17 | `a2a_agents.auth_query_params` | services_auth | sa_json_targets | ✅ |
| 18 | `a2a_agent_auth.auth_query_params` | services_auth | sa_json_targets | ✅ |
| 19 | **`tools.headers`** (nested `{_mcpgateway_encrypted_tool_header…: encode_auth(...)}`) | services_auth | — | ❌ **MISSING** |
| 20 | **`llm_providers.config`** (nested `{_mcpgateway_encrypted_value_v1: encode_auth(...)}` per sensitive key) | services_auth | — | ❌ **MISSING** |

Out of scope (correctly not a target): llmchat user config → Redis with TTL (`llmchat_router._serialize_user_config_for_storage`), not a DB column.

## Answer: No — the PR does not cover all sensitive columns

The PR delivers its stated scope (issue **#6323** = `gateways.auth_value`) plus two adjacent additions (`a2a_push_notification_configs.auth_token`, `servers.oauth_config`). But **three encrypted-column sinks remain uncovered** by the migration inventory:

1. **`a2a_agents.oauth_config`** — EncryptionService-encrypted `client_secret`. `json_targets` lists `gateways`/`servers`/`a2a_agent_auth` but omits `a2a_agents`, even though `A2AService` writes `oauth_config` directly onto the `a2a_agents` row via `protect_oauth_config_for_storage`. Same class of bug as #6323 (row written, target list omits it) → these secrets become unreadable after rotation.
2. **`tools.headers`** — sensitive header values wrapped in `encode_auth` envelopes (`_protect_tool_headers_for_storage`); needs a recursive services_auth JSON walker (no existing list does this).
3. **`llm_providers.config`** — sensitive config keys wrapped in `encode_auth` envelopes (`protect_provider_config_for_storage`); `api_key` is covered but the `config` JSON is not.

Finding #9 is the highest-severity gap — it is the exact same failure mode as the bug this PR fixes, just on a sibling table. If the intent is a *complete* rotation inventory, add `a2a_agents` to `json_targets` and add two new recursive services_auth JSON target passes for `tools.headers` and `llm_providers.config`. If the PR is intentionally scoped only to #6323, these three should at least be filed as follow-up so operators aren't left with silently-undecryptable credentials after `AUTH_ENCRYPTION_SECRET` rotation.

msureshkumar88 (changes requested):
Re-reviewed the commits landed since my last approval (`988ab354d`, `70ba524b3`, `a76581c6b` — the response to @Lang-Akshay's coverage-gap review) and re-ran an end-to-end verification against a real Postgres 16 instance.

## Functional correctness — verified, all three gaps fixed correctly

Built actual ORM rows via `db.py` models (`A2AAgent.oauth_config`, `Tool.headers`, `LLMProvider.config`) seeded with old-key ciphertext exactly as the real write paths produce them (`protect_oauth_config_for_storage`, `_protect_tool_headers_for_storage`, `protect_provider_config_for_storage`), then ran `run_migration()` against real Postgres:

- `a2a_agents.oauth_config` → `client_secret` migrated, decrypts under new key, correctly fails under old key.
- `tools.headers` → sentinel-wrapped `Authorization` header migrated correctly; sibling non-sensitive header (`X-Trace-Id`) left untouched.
- `llm_providers.config` → sentinel-wrapped key migrated correctly; sibling non-sensitive key (`region`) left untouched.

`3 migrated / 0 errors`, all three new/edited target lists (`json_targets` with `a2a_agents`, `sa_sentinel_targets` for `tools.headers` and `llm_providers.config`) behave as intended. `_reencrypt_sentinel_json`'s recursion correctly distinguishes the sentinel envelope shape from a regular nested dict.

## Blocking: CI diff-cover gate is currently failing

`pytest (py3.12)` — the `pytest` step itself passes, but the `Diff-cover (changed lines)` step fails: **80% coverage on changed lines, gate requires 93%**. Missing lines are the untested edge/error branches introduced by this round's diff:

- `migrate_enc_secret.py:399-400,402,405-406,415-417` — non-str/undecodable `raw_val` skip paths and the `error:` status branch in `_migrate_services_auth_json_scalar_columns`.
- `migrate_enc_secret.py:551-553,567-575,628-630,632` — the sentinel-JSON `error:`/list-recursion branches in `_reencrypt_sentinel_json`, and the `json.JSONDecodeError` fallback in `_migrate_services_auth_sentinel_json_columns`.

These are exactly the paths that matter for this kind of migration script (decrypt-failure and malformed-value handling) — worth covering directly rather than just closing the coverage percentage.

## Summary

Logic is correct and matches your PR description; e2e-verified on Postgres, not just SQLite. Requesting changes only on the CI coverage gate — once tests cover those branches (or an intentional `# pragma: no cover` is added with justification, for anything truly unreachable) this should be good to merge.

Lang-Akshay (changes requested):
Checked all 21 columns encrypted under AUTH_ENCRYPTION_SECRET against the script's six target lists. One column is not migrated:

| Column | Encryption | How it's encrypted | Migration coverage |
|---|---|---|---|
| `gateways.client_key` | EncryptionService (`v2:{...}`) | `_encrypt_client_key` → `encrypt_secret_async`; written at register (`gateway_service.py:1593`, `:2001`) and update (`:2846`); read via `decrypt_secret_or_plaintext` (`:1147`) | ❌ Absent from `simple_targets`, `json_targets`, and all `sa_*` lists |

msureshkumar88 (approved):
Re-reviewed commit `4eebe71e3` (top of branch), which lands the fixes for both outstanding blocking items from the prior round.

## Both blockers resolved

**Coverage gate** — the added tests (202 lines in `test_migrate_enc_secret.py`) cover the previously-missing error/edge branches. Ran locally against `main`: `diff-cover` reports **99.1%** on changed lines (gate requires 93%). CI's `Diff-cover (changed lines)` step and all 31 checks are green.

**`gateways.client_key` gap** — now added to `sa_simple_targets` (`("gateways", "id", ["client_key"])`), matching its actual encryption path (`EncryptionService` / `v2:{...}` via `_encrypt_client_key`).

## E2E verification (real Postgres 16, not SQLite)

Seeded a `gateways` row with `client_key` encrypted under an old key (via the same `encrypt_secret_async` path the real write flow uses), ran `run_migration()` against Postgres, then verified:

- `gateways [client_key] → found=1 migrated=1 errors=0`
- Re-encrypted ciphertext decrypts correctly under the new key, matches original plaintext
- Same ciphertext correctly fails decryption under the old key (`ValueError: Decryption failed`)
- No false-positive collisions with the row's other target lists (JSON `oauth_config`, `services_auth` scalar/query-params all correctly report `skipped` for this row's null values)

Also ran the full local suite: 96/96 unit tests pass.

Logic, coverage, and encryption-path targeting all check out. Approving.

Lang-Akshay (approved):
LGTM ✅

madhu-mohan-jaishankar (approved):
LGTM

Issue #6256, opened 2026-08-14
Apply MCP 2026 request metadata and headers 


lucarlig:
Isn't this tracked by **feat!: add SEP-2243 HTTP standard headers** https://github.com/modelcontextprotocol/rust-sdk/pull/907/changes#diff-35d107ff54233b848b4866d7ae1a726272f930c1eb9a9ed8963749b4aaa3e9fb

## Gap

The stateless `tools/call` path must validate MCP 2026-07-28 request metadata and keep the Streamable HTTP headers consistent with the JSON-RPC body. This integration is not complete on the current tool executor.

## Scope

- Require and validate the per-request protocol version, client identity, capabilities, and negotiated extensions in `_meta`.
- Validate inbound `Mcp-Method` and `Mcp-Name` against the JSON-RPC tool request.
- Generate the required headers for the routed upstream `tools/call`.
- Apply `x-mcp-header` tool-parameter mappings through the protocol-wide safe-header policy.
- Preserve request/trace metadata without leaking authentication or carrying state between concurrent calls.
- Return the protocol-defined errors for unsupported versions and header/body mismatches.

## Acceptance criteria

- [ ] Missing or malformed required metadata is rejected before hooks or backend execution.
- [ ] Header/body mismatches return the defined MCP error.
- [ ] Upstream method/name headers reflect the routed tool request.
- [ ] Protected, hop-by-hop, authentication, host/routing, and smuggling-sensitive headers cannot be injected.
- [ ] Concurrent requests cannot leak metadata or headers to one another.
- [ ] Wire tests cover valid, missing, malformed, mismatched, custom-header, and concurrent-request cases.

## Related work

- Protocol-wide request-header rules: #5687
- Stateless tool executor: #6243
- Upstream TLS/authentication: #4546

Issue #6147, opened 2026-08-09
Tenant-safe Mcp-Param-* validation and schema handling


gandhipratik203:
## Parent

Part of #5687.

## Background

The gateway currently does not provide tool schemas to RMCP standard-header validation, so `Mcp-Param-*` validation is effectively skipped.

A naive `get_tool(name)` implementation is not safe because RMCP's current schema cache is keyed by bare tool name. Gateway tool names can collide across users and virtual hosts. The correct identity is:

```text
(subject, virtual_host_id, exposed_tool_name)
```

This issue implements tenant-safe schema lookup and validation.

## Scope

- Add or drive the required RMCP change so standard-header validation can access request context, not only a bare tool name.
  - The schema lookup path needs access to request `Parts`/extensions so it can use claims, virtual host, and user config.
- Implement gateway-side schema resolution for exposed tool names.
- Add a bounded schema cache keyed by:
  - subject
  - virtual host id
  - exposed tool name
- Add invalidation when `notifications/tools/list_changed` is observed.
- Ensure `Mcp-Param-*` mismatches are rejected with HTTP 400 plus JSON-RPC error code `-32020`.
- Handle schema composition interaction with #5688:
  - `x-mcp-header` behind `$ref`, `allOf`, `oneOf`, etc. must either be resolved correctly or fail closed.
  - It must not silently skip validation.

## Acceptance Criteria

- Valid `Mcp-Param-*` headers are accepted when they match the tool input schema.
- Invalid or mismatched `Mcp-Param-*` headers are rejected with HTTP 400 and `-32020`.
- Two different subjects/vhosts can expose the same tool name with different schemas without schema leakage.
- Cache size is bounded.
- Cache entries are invalidated or refreshed after `tools/list_changed`.
- Composed schemas from #5688 cannot silently bypass `x-mcp-header` handling.

## Dependencies

- RMCP upstream PR/release, or a temporary local patch.
- Coordination with #5688 for schema composition behavior.

## Out of Scope

- Header count/value-size limits.
- Modern-only protocol policy.
- Full conformance/docs sweep.

jonpspri:
## Partial progress — remaining scope

PR #6348 (merged 2026-08-25) delivered the tenant-safe schema publishing side of this issue:

[...]

PR #6348, opened 2026-08-21
feat: publish tool schemas to dataplane

lucarlig:
## Summary

- publish visibility-filtered MCP tool input schemas in the existing per-user Redis snapshot
- key schemas by virtual host, backend, and original upstream tool name
- reject non-object schemas instead of publishing an invalid dataplane contract

This supplies the request-scoped schema source required by contextforge-org/contextforge-data-plane#109. The dataplane validates client-supplied `Mcp-Param-*` headers before forwarding them unchanged; it does not reconstruct headers or call `tools/list` on the request path. The existing periodic control-plane discovery/publisher flow remains the schema producer.

Related: #6147 and #6256.

## Validation

- publisher unit tests: 26 passed
- Ruff and Black
- cross-repo E2E with the official conformance fixture: matching plain/Base64 values pass; mismatched, invalid Base64, and missing headers return HTTP `400` / JSON-RPC `-32020`

### DIFF
--- a/mcpgateway/services/dataplane_publisher.py
+++ b/mcpgateway/services/dataplane_publisher.py
@@ -10,6 +10,7 @@ Be careful not to overfit production-grade assumptions onto the design.
 
 # Standard
 import asyncio
+from collections.abc import Sequence
 from collections import defaultdict
 import logging
 import os
@@ -67,9 +68,21 @@ class BackendConfig(TypedDict):
     remove_headers: list[str]
     capabilities: dict[str, Any]
     allowed_tool_names: list[str]
+    tool_schemas: dict[str, dict[str, Any]]
     allowed_resource_names: list[str]
     allowed_resource_uris: list[str]
     allowed_prompt_names: list[str]
+
+
+class GatewayBaseConfig(TypedDict):
+    """Gateway connection fields shared by every virtual-host backend."""
+
+    name: str
+    url: str
+    passthrough_headers: list[str]
+    add_headers: dict[str, str]
+    remove_headers: list[str]
+    capabilities: dict[str, Any]
 
 
 class VirtualHostConfig(TypedDict):
@@ -84,7 +97,27 @@ class UserConfig(TypedDict):
     virtual_hosts: dict[str, VirtualHostConfig]
 
 
-BackendItems = dict[str, list[str]]
+class BackendItems(TypedDict):
+    """Database identifiers associated with one server backend."""
+
+    tools: list[str]
+    resources: list[str]
+    prompts: list[str]
+
+
+class PublishedBackendItems(BackendItems):
+    """User-filtered backend items enriched with tool input schemas."""
+
+    tool_schemas: dict[str, dict[str, Any]]
+
+
+class ToolMetadata(TypedDict):
+    """Tool fields needed to publish dataplane routing configuration."""
+
+    name: str
+    input_schema: dict[str, Any]
+
+
 BackendItemsByServer = dict[str, dict[str, BackendItems]]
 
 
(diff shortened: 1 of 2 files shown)


lucarlig:
No longer needed. The dataplane consumer now treats Mcp-Param-* as transparent gateway metadata: it forwards the client-supplied headers unchanged and leaves validation to the upstream MCP server. It no longer consumes published tool schemas or calls tools/list on the call path. contextforge-org/contextforge-data-plane#109 passes the complete composed conformance run against the stock published control-plane image.

lucarlig:
Reopened: the dataplane design now validates header/body consistency from control-plane-published schemas before forwarding, so this publisher contract is required. This supersedes the earlier closure rationale; the dataplane still does not call tools/list or reconstruct parameter headers.

msureshkumar88 (changes requested):
Thanks for this — the wiring for tenant-safe schema publishing (`tool_schemas` keyed by user → virtual host → gateway → tool name) looks right, and the diff stays tightly scoped to that goal. Most of the extra lines are just typing scaffolding forced by the new `TypedDict` fields, no scope creep. No Alembic migration needed here, correctly.

One thing I'd like addressed before merge though:

**Blocking: single bad tool schema takes down dataplane publishing for every tenant, not just its owner**

In `_build_user_data` (`mcpgateway/services/dataplane_publisher.py`):

```python
for tool in visible_tools:
    if not isinstance(tool.input_schema, dict):
        raise ValueError(f"Tool {tool.id} has a non-object input schema")
```

This runs inside a dict comprehension in `get_data_from_db` that builds every active user's config in one expression. If this raises for *any* tool visible to *any* user, the whole comprehension aborts, `get_data_from_db` returns `None`, and the publish cycle is skipped entirely — leaving *all* tenants on a stale (eventually expired) snapshot until someone finds and fixes the one offending row.

Given `input_schema` is `NOT NULL` and both the Pydantic layer and the ORM `before_insert`/`before_update` event make it hard (not impossible — legacy rows, bulk imports, `bootstrap_db.py` can bypass both) to get a non-dict value in there, this is a low-probability but high-blast-radius failure mode: one tenant's bad data silently breaks routing for everyone else.

Suggest scoping the fail-closed behavior to the single tool instead — exclude it from `tool_by_id` and log a warning with the tool id, rather than raising out of the comprehension. Would also want `test_build_user_data_rejects_non_object_tool_schema` updated to assert exclusion (and ideally a test proving other users/tools in the same cycle are unaffected) rather than `pytest.raises`.

Everything else (no docs needed, no unrelated changes, tests passing, CI green) looks good — happy to re-review once that's addressed.

lucarlig:
@msureshkumar88 Thanks — fixed in 6fae49d85. `_build_user_data` now excludes only the visible tool whose `input_schema` is not an object, logs a warning with the tool ID, and continues building the rest of the multi-tenant snapshot. I updated the focused regression test to verify that the malformed tool is excluded while a valid tool remains, and the mocked DB-cycle test now includes a malformed private tool while still asserting snapshots for all three users. The publisher unit tests, pre-commit hooks, Ruff, Bandit, Interrogate, Pylint, and package verification all pass. Ready for re-review.

msureshkumar88 (approved):
Thanks for tightening this up — the malformed-schema isolation fix looks solid. Verified `_build_user_data` now excludes only the offending tool (with a warning log) instead of failing the whole snapshot, and the new test (`test_build_user_data_excludes_non_object_tool_schema`) plus the extended multi-user payload test cover it well.

Scope checks out against #6147: `tool_schemas` is nested under subject → virtual host → gateway, and `Tool` has a `(gateway_id, original_name)` uniqueness constraint, so there's no cross-tenant schema leakage. No Alembic migration needed (`input_schema` already existed), backend-only change, feature stays gated behind `dataplane_publisher` (off by default), and the `**gateway_config` → explicit-key refactor is behavior-preserving. Not a breaking change.

Two non-blocking notes for a follow-up, not required here:
- Schema duplication: a public tool's schema is now serialized once per visible user in the snapshot rather than once total (previously only short tool names were duplicated). Worth a follow-up to dedupe by `(gateway_id, tool_name)` at scale.
- No size/depth bound on published `input_schema` — combined with the per-user duplication above, an oversized schema on a widely-visible tool amplifies snapshot size. Not exploitable today, but worth tracking as a hardening item.

Approving — nice fix.

Lang-Akshay (approved):
LGTM

PR #6350, opened 2026-08-21
fix(gateway): reject/strip invisible Unicode in stored auth credentials

### LINKED ISSUE #6347: [BUG]: Gateway registration fails with unhandled UnicodeEncodeError when auth credential contains invisible Unicode characters
### 🐞 Bug Summary

Gateway registration (`POST /gateways`) fails with an unhandled `UnicodeEncodeError` — surfaced to the caller as a generic `GatewayConnectionError` — when a stored authentication credential (bearer token, basic auth password, or custom header value) contains an invisible Unicode "format" character (category `Cf`), such as `U+2060 WORD JOINER`, `U+200B ZERO WIDTH SPACE`, or `U+FEFF ZERO WIDTH NO-BREAK SPACE`.

These characters are common copy/paste artifacts — they get inserted at soft-wrap points when a token is copied out of a rendered document (Confluence, PDF, a web page) rather than a plain-text source. They're invisible in the Admin UI, in the browser, and in log output, so the contaminated credential looks identical to a clean one at every point up to the moment it's handed to the HTTP client.

Because none of the credential-ingestion schemas validate header values, and none of the outbound HTTP call sites sanitize them before use, the corrupted credential is accepted, encrypted, and stored without error. The failure only surfaces later, disguised as a connectivity problem, which sends debugging effort toward the network/firewall/TLS stack instead of the actual cause.

---

### 🧩 Affected Component

- [x] `mcpgateway` - API

---

### 🔁 Steps to Reproduce

1. Register a gateway with `auth_type=bearer` and a token containing an invisible Unicode format character, e.g. (Python) `"A" * 48 + "⁠" + "B" * 20`.
2. `POST /gateways` with that credential against any target URL (the URL does not need to be reachable — the failure happens before any network call).

**Observed:**

```
502 Bad Gateway
{"message":"Failed to initialize gateway at <url>: 'ascii' codec can't encode character '⁠' in position 55: ordinal not in range(128)"}
```

---

### 🤔 Expected Behavior

Either (a) the credential is rejected at save time with a clear validation error naming the offending character and its position, so the problem is caught immediately without decoding a stack trace, or at minimum (b) the runtime failure is reported as a distinct error type (not `GatewayConnectionError`) so it isn't mistaken for connectivity/network trouble.

---

### 📓 Logs / Error Output

Server-side traceback (abridged, no credentials included):

```
mcpgateway/services/gateway_service.py:5126 in _initialize_gateway
  -> connect_to_streamablehttp_server (gateway_service.py:7120)
     -> get_httpx_client_factory (gateway_service.py:7107)
        -> httpx.AsyncClient(headers=authentication, ...)
           -> httpx/_client.py:210 self.headers = Headers(headers)
              -> httpx/_models.py:156 _normalize_header_value(v, encoding)
                 -> httpx/_models.py:82 value.encode(encoding or "ascii")
UnicodeEncodeError: 'ascii' codec can't encode character '⁠' in position 55: ordinal not in range(128)
```

The root `UnicodeEncodeError` is caught by `_initialize_gateway`'s broad exception handler and re-raised as:

```
GatewayConnectionError(f"Failed to initialize gateway at {sanitized_url}: {sanitized_error}")
```

which is indistinguishable, from the caller's perspective, from an actual network/DNS/TLS failure.

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Version or commit | main (reproduced against current `main`) |
| Runtime | Python 3.13, httpx 0.28.1 |
| Platform / OS | Linux |
| Container | Reproduced both in a container deployment and locally via `uvicorn` |

---

### 🧩 Additional Context

**Root cause analysis:**

1. **No value-level validation on credential-ingestion schemas** (`mcpgateway/schemas.py`). `auth_token`, `auth_password`, `auth_username`, `auth_header_value`, and `auth_headers[].value` are plain `Optional[str]` fields on `GatewayCreate`, `GatewayUpdate`, `ToolCreate`, and `ToolUpdate` — no `field_validator` restricts their content. `_encode_auth_headers_list()` (used by the `authheaders` auth type) validates header **keys** strictly (`c.isalnum() or c in "-_"`) but only does an `isinstance(value, str)` check on header **values** — no charset validation.

2. **Storage round-trips the character untouched.** `encode_auth()` / `decode_auth()` (`mcpgateway/utils/services_auth.py`) are AES-GCM encrypt/decrypt around `orjson`; they preserve the string exactly, including invisible format characters.

3. **httpx never falls back past ASCII for `str` header values.** httpx's `_normalize_header_value()` (`httpx/_models.py`) does an unconditional `value.encode(encoding or "ascii")` for `str` header values — no UTF-8 fallback, no pre-check. Any non-ASCII character in a header value raises `UnicodeEncodeError` at `httpx.AsyncClient(headers=...)` construction time, before any request is sent.

4. **Existing header sanitizers in this codebase don't catch this class of character.** `sanitize_header_value()` exists in both `mcpgateway/utils/passthrough_headers.py` and `mcpgateway/translate_header_utils.py`, but both only strip ASCII control characters (`ord(c) < 32`). Unicode format characters (category `Cf`) have `ord() > 32` and pass through unchanged. Neither sanitizer is wired into the credential-ingestion or outbound-header path anyway.

5. **Blast radius is wider than gateway registration.** The same unsanitized `decode_auth()` → httpx-headers path is used by gateway registration, gateway health checks, and runtime tool invocation through a federated gateway (`gateway_service.py`, `tool_service.py`). A credential that becomes contaminated **after** a gateway is already registered and healthy (e.g. via an update) will pass registration but fail every subsequent health check and tool call, with the same misleading error framing.

6. **The error-wrapping makes diagnosis needlessly hard.** `_initialize_gateway`'s exception handler catches all exceptions, including `UnicodeEncodeError`, and re-raises them uniformly as `GatewayConnectionError`, with no way to distinguish "credential is malformed" from "server is unreachable" without reading the exception text.

**Suggested fix:**

1. Validate credential values at ingestion — strip Unicode format characters (category `Cf`, always safe to remove) and reject any remaining non-ASCII content with a clear error naming the character and position, across `GatewayCreate`/`GatewayUpdate`/`ToolCreate`/`ToolUpdate`.
2. Normalize already-decrypted credentials defensively at the points where outbound headers are built, so already-contaminated stored credentials recover without a manual re-save.
3. Stop conflating validation failures with connectivity failures — report `UnicodeEncodeError` (and similar) as a distinct error type rather than `GatewayConnectionError`.

**Acceptance criteria:**

- [ ] Creating/updating a tool or gateway with a credential value containing a non-ASCII, non-printable, or Unicode format character either strips the safely-strippable characters or is rejected with a clear validation error identifying the offending character and position.
- [ ] Existing gateways/tools with already-stored contaminated credentials are normalized at the point of use (decrypt-time), without requiring a manual re-save.
- [ ] `UnicodeEncodeError` (and similar header-encoding failures) raised while building outbound HTTP headers is no longer reported as `GatewayConnectionError` / a generic connectivity error.
- [ ] Regression tests cover gateway create/update, tool create/update, and the health-check and tool-invocation runtime paths, with a credential value containing at least one `Cf` character.

msureshkumar88:
## Summary

- Bearer tokens, basic-auth credentials and custom header values were accepted, encrypted and stored with no charset validation. A value containing a Unicode format character (e.g. `U+2060 WORD JOINER` -- a common copy/paste artifact from rendered documents) passed through unchanged and later crashed httpx's header encoding with a bare `UnicodeEncodeError`, which gateway registration wrapped into a generic `GatewayConnectionError` -- indistinguishable from an actual network failure.
- Adds `SecurityValidator.sanitize_credential_value()`: strips Unicode format characters (category `Cf`, never meaningful in a real credential) and rejects any other non-ASCII content with a message naming the character and its position. Wired into credential ingestion across `GatewayCreate`/`GatewayUpdate`, `ToolCreate`/`ToolUpdate`, and `A2AAgentCreate`/`A2AAgentUpdate`, so a bad value is rejected at save time with a clear 422 instead of failing later.
- Applies the same sanitization defensively wherever a stored credential is decrypted for outbound use -- gateway registration, gateway health checks, runtime tool invocation, and the direct-proxy auth-header builder -- so gateways configured before this change self-heal on next use without a manual re-save.
- Adds `GatewayCredentialError` (distinct from `GatewayConnectionError`) so a malformed stored credential is reported as what it is, not as a connectivity failure.

Closes #6347

## Test plan

- [x] New unit tests for `SecurityValidator.sanitize_credential_value` (strip vs reject vs passthrough)
- [x] New unit tests across `GatewayCreate`/`GatewayUpdate`/`ToolCreate`/`ToolUpdate` covering strip-and-accept and reject-with-message behavior
- [x] New unit tests for `_initialize_gateway` verifying a `UnicodeEncodeError` is reported as `GatewayCredentialError`, not `GatewayConnectionError`
- [x] New unit tests for gateway health-check self-healing a contaminated stored credential vs. failing clearly on genuine non-ASCII
- [x] New unit tests for `build_gateway_auth_headers` (direct-proxy path)
- [x] Full existing test suites for all touched modules pass with no regressions
- [x] `make ruff interrogate pylint` clean (10.00/10 pylint, 100% docstring coverage)
- [x] Doctests pass across all touched modules

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^\\.secrets\\.baseline$",
     "lines": null
   },
-  "generated_at": "2026-08-26T13:06:30Z",
+  "generated_at": "2026-08-26T15:11:31Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^\\.secrets\\.baseline$",
     "lines": null
   },
-  "generated_at": "2026-08-26T15:11:31Z",
+  "generated_at": "2026-08-27T08:57:50Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -216,7 +216,7 @@
         "hashed_secret": "f4793151b0607198d4de9b1ca458d3e25adf1cb7",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 189,
+        "line_number": 190,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -224,7 +224,7 @@
         "hashed_secret": "fa9beb99e4029ad5a6615399e7bbae21356086b3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 302,
+        "line_number": 303,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -232,7 +232,7 @@
         "hashed_secret": "d08f88df745fa7950b104e4a707a31cfce7b5841",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 304,
+        "line_number": 305,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -240,7 +240,7 @@
         "hashed_secret": "90bd1b48e958257948487b90bee080ba5ed00caa",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 405,
+        "line_number": 406,
         "type": "Hex High Entropy String",
         "verified_result": null
       },
@@ -248,7 +248,7 @@
         "hashed_secret": "48ffbad96aa9c2b33f9486f5a3c2108198acb518",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 406,
+        "line_number": 407,
         "type": "Hex High Entropy String",
         "verified_result": null
       }
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^\\.secrets\\.baseline$",
     "lines": null
   },
-  "generated_at": "2026-08-26T13:10:38Z",
+  "generated_at": "2026-08-26T13:06:30Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -216,7 +216,7 @@
         "hashed_secret": "f4793151b0607198d4de9b1ca458d3e25adf1cb7",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 190,
+        "line_number": 189,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -224,7 +224,7 @@
         "hashed_secret": "fa9beb99e4029ad5a6615399e7bbae21356086b3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 303,
+        "line_number": 302,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -232,7 +232,7 @@
         "hashed_secret": "d08f88df745fa7950b104e4a707a31cfce7b5841",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 305,
+        "line_number": 304,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -240,7 +240,7 @@
         "hashed_secret": "90bd1b48e958257948487b90bee080ba5ed00caa",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 406,
+        "line_number": 405,
         "type": "Hex High Entropy String",
         "verified_result": null
       },
@@ -248,7 +248,7 @@
         "hashed_secret": "48ffbad96aa9c2b33f9486f5a3c2108198acb518",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 407,
+        "line_number": 406,
         "type": "Hex High Entropy String",
         "verified_result": null
       }
@@ -3944,7 +3944,7 @@
         "hashed_secret": "c377074d6473f35a91001981355da793dc808ffd",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 785,
+        "line_number": 786,
         "type": "Hex High Entropy String",
         "verified_result": null
       }
@@ -3980,7 +3980,7 @@
         "hashed_secret": "c377074d6473f35a91001981355da793dc808ffd",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 4690,
+        "line_number": 4692,
         "type": "Hex High Entropy String",
         "verified_result": null
       }
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^\\.secrets\\.baseline$",
     "lines": null
   },
-  "generated_at": "2026-08-27T08:57:50Z",
+  "generated_at": "2026-08-27T10:13:19Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -3980,7 +3980,7 @@
         "hashed_secret": "c377074d6473f35a91001981355da793dc808ffd",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 4692,
+        "line_number": 4706,
         "type": "Hex High Entropy String",
         "verified_result": null
       }
--- a/mcpgateway/admin.py
+++ b/mcpgateway/admin.py
@@ -170,6 +170,7 @@ from mcpgateway.services.export_service import ExportError, ExportService
 from mcpgateway.services.gateway_service import (
     gateway_capability_loaders,
     GatewayConnectionError,
+    GatewayCredentialError,
     GatewayDuplicateConflictError,
     GatewayLookupConflictError,
     GatewayNameConflictError,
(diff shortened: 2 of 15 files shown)


Lang-Akshay (changes requested):
## Blocking Changes
| 1 | Issue / Quality | `mcpgateway/common/validators.py`, `mcpgateway/schemas.py` | `validators.py:2372-2410`; schema call sites `schemas.py:650`, `727`, `3388-3423`, `3735-3770`, `5281-5316`, `5624-5659` | `sanitize_credential_value()` removes only Unicode category `Cf` and explicitly preserves every other character. Ordinary non-ASCII (for example CJK) and non-printable control characters therefore remain in credentials, contradicting issue #6347 acceptance criterion 1 and the PR description's stated reject behavior; they can still reach httpx and fail later. | After removing `Cf`, reject remaining non-ASCII/non-printable credential characters with the offending character and position, or otherwise implement the issue's explicitly accepted safe-normalization policy. Apply the same rule consistently to all credential schemas. |
| 2 | Issue / Quality | `mcpgateway/services/tool_service.py` | Changed gateway sanitization at `4560-4563` and `6029-6032`; unsanitized tool-auth decode path at `5585-5588` | Issue #6347 acceptance criterion 2 requires existing gateways/tools with contaminated stored credentials to be normalized at decrypt/use time. The new sanitization covers gateway/Vault headers only; the REST tool branch still decodes `tool_auth_value` directly and merges it into outbound headers, so a pre-existing contaminated tool credential is not self-healed. | Sanitize decoded `tool_auth_value` credentials before `headers.update()` (and cover every tool outbound branch), rejecting/failing closed if sanitization leaves an invalid or empty required credential. |

msureshkumar88:
Thanks for the detailed review — went through both findings against the code and history:

**Finding 1 (reject remaining non-ASCII/control chars):** This was actually the original implementation, and it turned out to be over-scoped — CI caught that rejecting all non-ASCII broke `test_valid_special_characters_in_values`, which is existing, deliberate support for international text (accented Latin, CJK, etc.) in header values, and it also masked the "Invalid header key format" error when a header had both a bad key and non-ASCII value (the value check ran first). See `b7067a54e` ("scope credential sanitization to invisible format characters only") for the full writeup. Narrowing to Unicode category `Cf` only (zero-width spaces, word joiner, BOM — copy/paste artifacts that are never meaningful in a real credential) is intentional and matches issue #6347's actual repro (`U+2060 WORD JOINER`), so I've left this as-is rather than reintroducing the regression.

**Finding 2 (REST tool outbound path not self-healing):** Confirmed and fixed in `5a0fb3a46`. `sanitize_credential_value()` was wired into the gateway-side decode paths but missed two tool-side spots: the REST tool's own `tool_auth_value` decode at the non-OAuth branch of `invoke_tool`, and `_decrypt_tool_headers_for_runtime()` (the shared helper that decrypts a tool's custom headers for outbound use, called from both the cache-hit and DB-load paths). Both now sanitize before the value reaches outbound headers, with unit tests covering the self-heal behavior for a contaminated stored `auth_value` and a contaminated custom header.

Rebased onto main and regenerated the secrets baseline; full test suite plus the two new regression tests are green.

Lang-Akshay (dismissed):
LGTM✅

gandhipratik203:
Nice work on this, the self-heal-on-read approach is the right call and the ingestion-side wiring covers all six schema paths. Two findings from tracing the new exception class through the routers:

**F1:** `GatewayCredentialError` subclasses `GatewayError`, not `GatewayConnectionError`, so the 502 branches at `main.py:7403`/`7506` and `admin.py:12909`/`13050` all miss it and it falls through to the generic 500.
A contaminated credential now returns `"Unexpected error"` with the detail dropped, instead of a 502 naming the failure. Worth adding an explicit branch (422 fits, since it's a bad stored value rather than an upstream fault) at those four sites.

**F2:** `gateway_service.py:3254` re-raises `GatewayConnectionError` when `init_affecting_changed`, specifically so a broken update rolls back, but `GatewayCredentialError` falls into the next bare `except Exception`.
That sets `reinit_succeeded = False` and lets the commit proceed, so the broken update is silently persisted with only a `logger.warning`.

msureshkumar88:
Thanks for tracing this through — both findings check out, fixed in the latest commits.

**F1:** Added explicit `GatewayCredentialError` branches (422) ahead of the `GatewayConnectionError` (502) checks at `main.py:7403`/`7506` and `admin.py:12909`/`13050`. While auditing the same pattern I found a third, unlisted site with the identical gap in `admin_edit_gateway` (`admin.py:13327`) — fixed that too, same treatment.

**F2:** `gateway_service.py:3254`'s `except` now catches `GatewayCredentialError` alongside `GatewayConnectionError`, so an init-affecting update with a broken credential still rolls back instead of committing silently. Tracing the re-raise path further up turned up two more spots where the type would've been lost even with that fix in place:
- The pre-existing wrap-into-`GatewayConnectionError` safety net at `gateway_service.py:3215` (unrelated to this PR, predates it) was catching *any* non-connection exception during an init-affecting update and relabeling it as `GatewayConnectionError` — so a credential error would've still rolled back correctly, but surfaced as a misleading 502 instead of 422. Excluded `GatewayCredentialError` from that wrap.
- The outer handler at `gateway_service.py:3404` only caught `GatewayConnectionError` for rollback+re-raise; anything else fell into the final `except Exception` and got wrapped into a plain `GatewayError`, which would've defeated the router-level 422 dispatch entirely. Added `GatewayCredentialError` there too.

Added regression tests for the router 422 responses (register/update via both `main.py` and `admin.py`), and for the two `update_gateway` branches: type preserved + rollback on an init-affecting change, and existing tolerate-and-commit behavior preserved for cosmetic-only changes (matching the `GatewayConnectionError` precedent there).

Full validation gate (lint, unit tests, secrets scan) is green; rebased cleanly onto main.

gandhipratik203 (dismissed):
LGTM! Thanks for the fixes.

Lang-Akshay (dismissed):
LGTM

Lang-Akshay (changes requested):
Stale review

Lang-Akshay (approved):
LGTM

PR #6351, opened 2026-08-21
fix: render gateway/tool/prompt/resource audit rows correctly in activity feed

### LINKED ISSUE #6342: [BUG]: Activity feed renders MCP server offline transitions as "success" with mangled titles
### Bug Summary

`GET /api/logs/activity` renders gateway (MCP server) audit rows incorrectly in three ways: an MCP server going **offline** is reported with `status: "success"`, and both the title and description come out as ungrammatical internal strings (*"Gateway set gateway state"*).

Found while wiring the React Home activity feed (contextforge-org/contextforge-web-ui#60) against the endpoint from #5944.

---

### Affected Component

- [x] `mcpgateway` - API
- [ ] Other (explain below)

---

### Steps to Reproduce

1. Register a gateway pointing at a reachable MCP server.
2. Stop the upstream server and let the periodic health check run until the gateway is marked unreachable (`GW_FAILURE_THRESHOLD` consecutive failures).
3. `GET /api/logs/activity`.

---

### Expected Behavior

The transition to offline should render as a `warning` or `error` item, with a human-readable title and, ideally, the reason.

### Actual Behavior

```json
{
  "id": "audit:...",
  "source": "audit",
  "title": "Gateway set gateway state",
  "description": "Gateway 'my-server' was set gateway state by system.",
  "status": "success",
  "resource_type": "gateway",
  "actor": "system"
}
```

---

### Root cause

The row itself is written correctly and at the right time. `_handle_gateway_failure` reaches `set_gateway_state(activate=True, reachable=False, only_update_reachable=True)`, whose `audit_trail.log_action` call (`mcpgateway/services/gateway_service.py:3710`) sits inside the transition guard at `:3558` — so there is exactly one row per real state change, no polling spam. The problem is purely in how `_audit_to_activity` (`mcpgateway/routers/log_search.py:355`) renders it.

**1. Status is wrong.** `_audit_to_activity` derives status from `row.success`, which defaults to `True` in `AuditTrailService.log_action`. None of the gateway failure paths pass `success=False`, so "server went offline" is indistinguishable from "server came back online" — both are `success`.

**2. `resource_type` doesn't match the label map.** `_RESOURCE_LABELS` contains `mcp_server` and `a2a_agent`, but `gateway_service.py` writes `resource_type="gateway"` at all six of its audit sites (`:2043`, `:3023`, `:3221`, `:3710`, `:3889`, `:4033`). The fallback produces `"Gateway"`, which is also inconsistent with the user-facing "MCP server" terminology used everywhere else in the product.

**3. `set_gateway_state` has no verb.** `_ACTION_VERBS` maps bare verbs (`create`, `update`, `delete`, …) but the gateway service writes composite actions — `create_gateway`, `update_gateway`, `delete_gateway`, `set_gateway_state`. None match, so the fallback `action.replace("_", " ")` yields *"create gateway"* / *"set gateway state"* and the rendered title becomes *"Gateway create gateway"*, *"Gateway set gateway state"*.

Note that (3) affects **every** gateway audit action, not just the state change — gateway create/update/delete are all currently mis-titled in the feed.

---

### Suggested fix

Three independent small changes:

1. Pass `success=False` (and `error_message` where one is available) from the failure paths in `gateway_service.py`. For the reachability transition specifically, `set_gateway_state` knows `reachable` and can set `success=not (activate and not reachable)` or similar — though the *reason string* is a separate problem, see the companion issue below.
2. Map `gateway` → `"MCP server"` in `_RESOURCE_LABELS`, matching the product's user-facing terminology.
3. Either add the composite actions to `_ACTION_VERBS` (`create_gateway` → `registered`, `update_gateway` → `updated`, `delete_gateway` → `deleted`, `set_gateway_state` → `went offline` / `came online` depending on `new_values["reachable"]`), or normalize the audit `action` values at the write site to bare verbs so the existing map applies.

Option 3 is cleaner long-term but is a data change. Existing rows keep the composite form, so the mapper needs the fallback regardless.

---

### Related

- #5944 / #5973: the activity feed endpoint this builds on
- Companion issue: the recurring health-check path discards the failure reason entirely, so no fix here can put a cause in the description (see linked issue)
- Consumer: contextforge-org/contextforge-web-ui#60

msureshkumar88:
## Summary

`_audit_to_activity` (the renderer behind `GET /api/logs/activity`) only mapped two `resource_type` values (`mcp_server`, `a2a_agent`) that no writer in the codebase actually emits, and only mapped bare verbs (`create`, `update`, ...). Every real gateway/tool/prompt/resource/server audit row therefore fell through to the fallback and rendered mangled titles like *"Gateway set gateway state"*, and the health-check reachability transition was indistinguishable from a normal successful write (`status: "success"`).

- Resolve resource labels against what `gateway_service.py` / `tool_service.py` / `prompt_service.py` / `resource_service.py` / `server_service.py` actually write (`gateway` → "MCP server", `server` → "Virtual server", etc.), keeping the two existing keys as inert compat entries.
- Resolve composite actions (`create_gateway`, `set_tool_state`, `bulk_create_tools`, `view_prompt_details`, ...) via a token scan rather than literal string matching, so a trailing resource noun is discarded instead of repeated.
- Render `set_*_state` rows from their recorded `new_values` (`enabled`/`reachable`) instead of the literal action string, so a server going offline renders as `status: "warning"` / *"MCP server went offline"* instead of `success`.
- Apply the same fix to the error branch, which had the identical mangling bug on failed writes (e.g. a failed `create_gateway` rendered *"MCP server create gateway failed"*).
- Prefer a writer-authored `context.description` (used by a few `log_audit` call sites) over the synthesised sentence when present.

Two changes go slightly beyond the issue's literal scope, called out here since they're the same defect class on adjacent code the issue didn't mention:
- The error-branch rewrite above.
- `view_*`/`browse`/`invoke` actions now resolve to `status: "info"` like `read`/`execute` already did, since status is now derived from the resolved verb rather than a literal string match.

Deliberately **not** changing any writer in `gateway_service.py` (the issue's suggested fix #1, flipping `success=False` on the offline transition): `AuditTrail.success` also drives SIEM export severity (`_emit_audit_event_to_siem`) and compliance evidence counts (`collect_audit_log_evidence`), so writing `success=False` for a health-check-driven reachability change would export every upstream outage as a `severity: "HIGH"` / `action_taken: "denied"` SIEM event and inflate `failure_events` in compliance reports, even though the audited write itself succeeded. Deriving status from `new_values` in the renderer gets the desired feed behavior without corrupting either downstream consumer.

Closes #6342

## Test plan
- [x] `tests/unit/mcpgateway/routers/test_log_search_activity.py` — extended `TestAuditMapper`, added `TestStateTransitions` and a hardcoded-inventory regression guard (`TestRealActionInventoryRegressionGuard`) covering every `(resource_type, action)` pair actually written by the services above
- [x] `make ruff bandit interrogate pylint` on changed files
- [x] `pytest tests/unit/mcpgateway/routers/` (full directory) + targeted audit/compliance service tests


vishu-bh (changes requested):
Overall this is a good direction!!

One P2 performance concern: SQLite metric fallbacks load all matching traces into memory via .all() in `observability_service.py:1868` and `1942`. High-volume 168-hour queries could cause latency or memory spikes. Consider bounded batching or database-side aggregation.

msureshkumar88:
Thanks for the look! Quick scope check on the review comment though — this PR only touches `mcpgateway/routers/log_search.py` and its test file (the activity-feed audit renderer for #6342). The flagged lines (`observability_service.py:1868` and `1942`, the SQLite Python-fallback timeseries/percentile queries) aren't part of this diff, so I don't think I can address that here without pulling in an unrelated file change.

That said, I did check — the concern itself is real: both `_execution_timeseries_python` and `_latency_percentiles_python` pull the full trace set into memory via `.all()` for the SQLite fallback path. Worth its own issue/PR against `observability_service.py` if it's not already tracked.

Rebased onto latest `main` and pushed; secrets baseline regenerated clean (no live/unaudited findings).

vishu-bh (dismissed):
LGTM 🚀

Lang-Akshay (dismissed):
LGTM ✅

Lang-Akshay (approved):
LGTM

PR #6352, opened 2026-08-21
feat(tls): add inbound mTLS (client certificate authentication) support

### LINKED ISSUE #6263: [FEATURE]: Add inbound mTLS (client certificate authentication) support to run-gunicorn.sh
## Summary

ContextForge's ASGI stack (Gunicorn + UvicornWorker + Python `ssl`) already supports inbound mutual TLS — where the server requires connecting clients to present a valid certificate before a session is established. However, the configuration wiring to activate this feature is absent. No env-var equivalents of `SSL`, `CERT_FILE`, and `KEY_FILE` exist for the client-certificate parameters, so operators cannot enable inbound mTLS without bypassing `run-gunicorn.sh` entirely.

This issue tracks adding that wiring. The feature must remain **fully optional and gated** — existing deployments that do not set the new variables must be completely unaffected.

## Background

The parameter propagation chain is already complete through the stack:

```
run-gunicorn.sh (missing env vars)
  → gunicorn --ca-certs / --cert-reqs        ← flags exist, just not passed
    → UvicornWorker ssl_options[ca_certs] / ssl_options[cert_reqs]
      → uvicorn Config(ssl_ca_certs=..., ssl_cert_reqs=2)
        → ssl.SSLContext.verify_mode = ssl.CERT_REQUIRED
          ssl.SSLContext.load_verify_locations(ca_certs)
```

Gunicorn's `--cert-reqs` maps directly to Python's `ssl` module:

| Value | Meaning |
|-------|---------|
| `0` | `ssl.CERT_NONE` — no client certificate (default, current behaviour) |
| `1` | `ssl.CERT_OPTIONAL` — request but do not require a cert |
| `2` | `ssl.CERT_REQUIRED` — reject any client without a valid cert |

`gunicorn.config.py` already contains a commented-out `# ca-certs = '/etc/ca_bundle.crt'` placeholder acknowledging this capability.

This is distinct from **outbound** mTLS (ContextForge presenting a client cert to an upstream MCP server), which is already fully implemented via `get_cached_ssl_context()` in `mcpgateway/services/gateway_service.py` with dedicated `make certs-mcp-*` Makefile targets.

## Proposed Changes

All changes are additive. Nothing changes unless the new env vars are explicitly set.

### 1. `run-gunicorn.sh`

Add two new optional environment variables, read only when `SSL=true`:

| Variable | Default | Purpose |
|----------|---------|---------|
| `CA_CERTS` | *(empty)* | Path to CA bundle used to verify client certificates |
| `CERT_REQS` | `0` | `0`=none, `1`=optional, `2`=required |

In **Section 7** (TLS configuration), after the existing `KEY_FILE_PASSWORD` block, add:

```bash
CA_CERTS=${CA_CERTS:-}      # Path to CA bundle for client cert verification (mTLS)
CERT_REQS=${CERT_REQS:-0}   # 0=CERT_NONE, 1=CERT_OPTIONAL, 2=CERT_REQUIRED

if [[ -n "${CA_CERTS}" ]]; then
    if [[ ! -f "${CA_CERTS}" ]]; then
        echo "❌  FATAL: CA certificate bundle not found: ${CA_CERTS}"
        exit 1
    fi
    if [[ ! -r "${CA_CERTS}" ]]; then
        echo "❌  FATAL: Cannot read CA certificate bundle: ${CA_CERTS}"
        exit 1
    fi
    if [[ "${CERT_REQS}" != "0" && "${CERT_REQS}" != "1" && "${CERT_REQS}" != "2" ]]; then
        echo "❌  FATAL: CERT_REQS must be 0, 1, or 2 (got: ${CERT_REQS})"
        exit 1
    fi
    echo "🔐  mTLS enabled:"
    echo "   CA Bundle:  ${CA_CERTS}"
    echo "   Cert reqs:  ${CERT_REQS} (0=none, 1=optional, 2=required)"
fi
```

In **Section 9** (SSL arguments), extend the existing SSL block:

```bash
if [[ "${SSL}" == "true" ]]; then
    cmd+=( --certfile "${CERT_FILE}" --keyfile "${KEY_FILE}" )
    if [[ -n "${CA_CERTS}" ]]; then
        cmd+=( --ca-certs "${CA_CERTS}" --cert-reqs "${CERT_REQS}" )
    fi
fi
```

The script header comment block should also document the two new variables.

### 2. `gunicorn.config.py`

Replace the commented-out `ca-certs` placeholder with an updated comment that describes the new env-var mechanism:

```python
# SSL/TLS Configuration
# Server certificate and key are set via CLI args in run-gunicorn.sh (SSL=true).
# mTLS (client certificate verification) is configured via:
#   CA_CERTS  env var → --ca-certs  <path>   CA bundle to verify client certs
#   CERT_REQS env var → --cert-reqs <0|1|2>  0=none (default), 1=optional, 2=required
```

No `on_starting()` hook changes are needed — unlike passphrase-protected keys, `ca_certs` and `cert_reqs` require no pre-processing and are forwarded directly through `ssl_options` to UvicornWorker.

### 3. `Makefile`

Add a `serve-ssl-mtls` convenience target (gated behind `certs` like `serve-ssl`):

```makefile
# help: serve-ssl-mtls       - Run Gunicorn with TLS + inbound mTLS (requires CA_CERTS)
serve-ssl-mtls: js-build certs   ## Run Gunicorn with TLS and client certificate authentication
	SSL=true CERT_FILE=certs/cert.pem KEY_FILE=certs/key.pem \
	CA_CERTS=${CA_CERTS} CERT_REQS=${CERT_REQS:-2} ./run-gunicorn.sh
```

And a `certs-client` helper for generating test client credentials:

```makefile
# help: certs-client         - Generate a client CA and test client cert for mTLS testing
certs-client:                ## Generate client CA and sample client cert for inbound mTLS testing
	mkdir -p certs/client
	openssl req -x509 -newkey rsa:4096 -sha256 -days 365 -nodes \
	    -keyout certs/client/ca-key.pem -out certs/client/ca-cert.pem \
	    -subj "/CN=ContextForge Client CA"
	openssl req -newkey rsa:4096 -nodes \
	    -keyout certs/client/client-key.pem -out certs/client/client-csr.pem \
	    -subj "/CN=test-client"
	openssl x509 -req -in certs/client/client-csr.pem \
	    -CA certs/client/ca-cert.pem -CAkey certs/client/ca-key.pem \
	    -CAcreateserial -out certs/client/client-cert.pem -days 365 -sha256
	@chmod 640 certs/client/ca-key.pem certs/client/client-key.pem
	@echo "✅  Client CA and client certificate written to ./certs/client/"
```

### 4. `docs/docs/deployment/tls-configuration.md`

Add an **"Inbound mTLS — Client Certificate Authentication"** section covering:
- The two new env vars and their meanings
- A `make certs-client && make serve-ssl-mtls` quickstart walkthrough
- How to verify with `curl --cert` / `--key` / `--cacert`
- A Docker Compose snippet showing the vars added to the `gateway` service environment

## Acceptance Criteria

- [ ] `CA_CERTS` and `CERT_REQS` are read in `run-gunicorn.sh` only when `SSL=true`
- [ ] When neither `CA_CERTS` nor `CERT_REQS` is set, behaviour is identical to today (no regression)
- [ ] Invalid `CA_CERTS` path (missing or unreadable file) causes a fast-fail with a clear error message before Gunicorn starts
- [ ] Invalid `CERT_REQS` value (not 0, 1, or 2) causes a fast-fail with a clear error message
- [ ] `CERT_REQS=2` with a valid `CA_CERTS` causes Gunicorn to reject connections from clients that do not present a certificate signed by that CA
- [ ] `make serve-ssl-mtls` and `make certs-client` targets work end-to-end
- [ ] Documentation is added to `docs/docs/deployment/tls-configuration.md`
- [ ] Existing tests continue to pass unchanged

## No New Dependencies

All required functionality is already present in Gunicorn, UvicornWorker, and Python's `ssl` module. This change is pure configuration wiring.

msureshkumar88:
## Summary

Wires Gunicorn's `--ca-certs` and `--cert-reqs` through `run-gunicorn.sh` so operators can require connecting clients to present a certificate signed by a nominated CA. The parameter chain (`gunicorn` → `UvicornWorker` `ssl_options` → uvicorn SSL context) already existed; only the launcher configuration was missing.

Everything is additive and gated. With the new variables unset, behaviour is unchanged.

Closes #6263

## New launcher variables

Read only when `SSL=true`:

| Variable | Default | Purpose |
|----------|---------|---------|
| `CA_CERTS` | *(unset)* | CA bundle used to verify client certificates |
| `CERT_REQS` | `0` | `0` = none, `1` = optional, `2` = required |
| `LOOPBACK_CLIENT_CERT` | *(unset)* | Certificate the gateway presents on its own loopback self-calls |
| `LOOPBACK_CLIENT_KEY` | *(unset)* | Private key for the above |

Validation fails fast, before Gunicorn starts, on a missing or unreadable bundle, an out-of-range `CERT_REQS`, and a half-configured loopback pair.

## Two findings worth flagging

**A non-zero `CERT_REQS` without `CA_CERTS` is rejected outright.** uvicorn never calls `load_default_certs()`, so there is no system trust store to fall back on — the context would have zero trusted CAs and reject every client with no useful diagnostic. The bundle is also validated when `CERT_REQS=0`, since `load_verify_locations()` runs whenever `ca_certs` is passed, so an unreadable file breaks worker boot regardless of policy.

**Requiring client certificates breaks the gateway's own loopback calls.** This is the part that makes the change more than configuration wiring. The SSE message-dispatch path (`cache/session_registry.py`, reached from every backend) and the WebSocket endpoint (`main.py`) both POST to `/rpc` over `127.0.0.1`. Those calls set `verify=False`, which skips *validating* the server but does not *present* a client certificate — so under `CERT_REQS>=1` the gateway rejects its own self-calls at the handshake and both transports go down.

`internal_loopback_verify()` now returns an `ssl.SSLContext` carrying the configured client certificate instead of a bare `False`. The context keeps the existing trust posture (`CERT_NONE`, `check_hostname` disabled) so no new server-certificate or SAN requirement is imposed on operators using their own certificates. Both call sites are unchanged — httpx accepts an `SSLContext` wherever it accepts `verify` — and the existing assertions on the `False`/`True` returns still hold when no loopback certificate is configured.

Verified against a real `CERT_REQUIRED` TLS server:

```
without loopback cert: FAILED -> [SSL: TLSV13_ALERT_CERTIFICATE_REQUIRED]
with loopback cert:    HTTP 200
```

## Also included

- `make certs-client` — idempotent generation of a client CA and test client certificate
- `make serve-ssl-mtls CA_CERTS=...` — TLS + required client certificates, with the loopback pair wired automatically
- Docs section in `docs/docs/deployment/tls-configuration.md` covering setup, verification, the loopback requirement, and the compatibility envelope
- `.env.example` and `docker-compose.yml` entries, both commented

## Compatibility notes documented

- The bundled nginx TLS profile terminates TLS at nginx and proxies **plain HTTP** to the gateway, so gateway-side `CA_CERTS`/`CERT_REQS` do not apply there. Enforced mTLS in front of that stack belongs on nginx.
- Every shipped healthcheck (image `HEALTHCHECK` and all three Compose variants) is certless and fails once the gateway requires a certificate. A certificate-bearing probe is provided.
- `CERT_REQS=1` verifies a certificate when one is offered but still admits callers presenting none — verification, not enforcement.
- No CRL/OCSP revocation checking.

## Validation

- `bats tests/bash/run_gunicorn_mtls.bats` — 13 new cases, all pass (includes regression guards that no flags appear when the variables are unset, and that `SSL=false` never reads them)
- `pytest tests/unit/mcpgateway/utils/test_internal_http.py` — 28 pass, 7 new for the loopback certificate
- `pytest tests/unit/mcpgateway/utils tests/unit/mcpgateway/cache` — no failures
- `make ruff`, `ruff format --check`, `interrogate` (100%) — pass
- `make detect-secrets-scan` — no unaudited or live secrets


gandhipratik203:
Two small fixes before merge. Great catch on the loopback break.

**F1. The `CERT_REQS=1` warning is wrong.** The launcher warns on any non-zero `CERT_REQS` that SSE/WebSocket "will fail at the TLS handshake", but `CERT_OPTIONAL` accepts a certless client, so the self-calls succeed. Only `CERT_REQS=2` fails. The docs table has it right; the shell warning, `.env.example`, and `docker-compose.yml` all say "non-zero". Since docs recommend `CERT_REQS=1` as the rollout step, operators hit the warning on the recommended path. Gate it on `== 2` and reword the other two.

**F2. Loopback validation is skipped when `CA_CERTS` is unset.** The both-or-neither and `-f`/`-r` checks sit inside `if [[ -n "${CA_CERTS}" ]]`, but `LOOPBACK_CLIENT_CERT`/`KEY` still reach the workers, and `internal_loopback_verify()` calls `load_cert_chain()` unguarded. A typo'd path then fails per request in the SSE dispatch instead of at boot. Move the block outside the `CA_CERTS` guard, still inside the `SSL` branch.

Once those are in we're good to go.

msureshkumar88:
Thanks for the review — both confirmed against the code, fixed:

**F1.** Gated the loopback warning on `CERT_REQS == 2` (was `!= 0`, so it fired on `CERT_REQS=1` too even though `CERT_OPTIONAL` admits a certless caller and the self-calls succeed). Reworded the matching "non-zero" language in `.env.example`, `docker-compose.yml`, and the docs table.

**F2.** Moved the `LOOPBACK_CLIENT_CERT`/`LOOPBACK_CLIENT_KEY` existence/readability checks (and the both-or-neither check) outside the `CA_CERTS` guard, so they run whenever `SSL=true` rather than only when `CA_CERTS` is set. A typo'd loopback path now fails at launcher boot instead of surfacing per-request inside `internal_loopback_verify()`.

Added bats coverage for both: `CERT_REQS=1` without loopback creds no longer warns, and a bad `LOOPBACK_CLIENT_CERT`/`KEY` is caught at boot even with `CA_CERTS`/`CERT_REQS` unset. Full suite (16 cases) and `test_internal_http.py` (28 cases) pass.

Also rebased onto latest `main` and regenerated `.secrets.baseline` (line-number drift only, no new findings).

gandhipratik203 (dismissed):
Thanks for the changes. LGTM!

Lang-Akshay (dismissed):
LGTM ✅

Lang-Akshay (approved):
LGTM

Issue #5370, opened 2026-06-24
[FEATURE]: UI-rewrite API tokens


a-effort:
### 🧭 Type of Feature

- [x] New feature or capability
- [x] Enhancement to existing functionality

### 🧭 Goal

Build the React **API Tokens** experience so a logged-in user can create, view, and revoke their own API tokens (personal or team-scoped) from the new UI, replacing the legacy HTMX token panel. [Design](https://www.figma.com/design/H2Dr9QPvRVxCe92V8URvGT/Refiring?m=auto&node-id=5265-39379&t=fqEtyB1LnYfGjMaQ-1)

**Scope:** self-service token management (the caller acting on their own tokens). Admin token CRUD across other users is out of scope for phase 1

---

### 📍 Background: current state

- **Backend API (`mcpgateway/routers/tokens.py`) is complete.** Endpoints relevant here:
  - `POST /tokens`, `GET /tokens`, `GET /tokens/{id}`, `DELETE /tokens/{id}` (Phase 1)
  - `PUT /tokens/{id}`, `GET /tokens/{id}/usage`, `POST`/`GET /tokens/teams/{team_id}` (later phases)
- **Full token data model** (`TokenScopeRequest` / `TokenResponse`): name, description, `expires_in_days`, `tags[]`, `team_id`, `is_active`, and a scope object of `server_id`, `permissions[]`, `ip_restrictions[]` (CIDR-validated), `time_restrictions{}`, `usage_limits{}`. Responses also carry `last_used` and revocation provenance (`is_revoked`, `revoked_at`, `revoked_by`, `revocation_reason`).
- **Legacy Admin UI** (`tokens_partial.html` + create form in `admin.html`) is personal-tokens-only: name, expires-in-days (capped 365), description, `server_id`, a single IP box, comma-separated permissions; list with status/team/server/IP badges; per-token Details, Usage Stats, Revoke. No edit path, no tags input, no time/usage-limit controls.

### Notes:

- **Expiry is server-enforced.** `REQUIRE_TOKEN_EXPIRATION` defaults to `True` (`config.py:391`) and `create_token` raises if no expiry is supplied (`token_catalog_service.py:525`). On a default deployment, creation fails without an expiry, so the expiry field is a Phase 1 requirement, not an option.
[...]

PR #6354, opened 2026-08-21
fix: default token creation to creator's personal team (#5993)

### LINKED ISSUE #5993: [BUG]: Multi-team token creation silently yields a public-only token (no team_id default, no warning)
### 🐞 Bug Summary

When a user who belongs to more than one team creates an API token without an explicit `team_id`, the token is silently issued with `team_id: null` (empty `teams` claim). At authorization time an empty `teams` claim is treated as public-only: the resource filter drops the owner clause, so the token can see only `visibility == "public"` resources, and cannot access even the creator's own private resources or any team resources. There is no error and no warning, so the creator gets a credential that quietly does far less than expected.

This is the residual, still-unaddressed part of #3291 ("Broken Inheritance"). The fix that closed #3291 only auto-inherits when the user belongs to exactly one team (`if len(user_teams) == 1`), which does not cover multi-team users.

---

### 🧩 Affected Component

- [x] `mcpgateway` - API

---

### 🔁 Steps to Reproduce

1. As a non-admin user belonging to two or more teams (for example a personal team plus one shared team), create a token without a `team_id`:

   ```bash
   curl -X POST "$BASE/tokens" \
     -H "Authorization: Bearer $SESSION_JWT" \
     -d '{"name": "multi-team-test", "expires_in_days": 1}'
   ```

2. Observe the response: `"team_id": null` and `"resource_scopes": []`. No warning is returned.

3. Use that token to list resources you own privately or that belong to your teams:

   ```bash
   curl "$BASE/tools" -H "Authorization: Bearer $NEW_TOKEN"
   ```

4. Result: only `visibility == "public"` resources come back. Your own private resources and your teams' resources are invisible, even though the token was created by you.

---

### 🤔 Expected Behavior

Creating a token should not silently produce a credential that cannot see the creator's own resources. It should instead default to the personal team: inherit the user's personal team when no `team_id` is given, so the token at least retains access to the creator's own resources.

Backend vs. UI: the ui-rewrite already plans a team selector on the create form (#5370, Story 2) as the UI-side mitigation, but the API remains directly callable with no `team_id`, so a backend-level decision (a/b/c) is still needed independent of the UI.

---

### 📓 Logs / Error Output

Relevant logic:

- Auto-inheritance only fires for single-team users: `create_token` in `mcpgateway/services/token_catalog_service.py` (and the mirror in `mcpgateway/routers/tokens.py`, `if len(user_teams) == 1: effective_team_id = user_teams[0]`).
- Empty `teams` claim normalizes to public-only: `normalize_token_teams()` in `mcpgateway/auth.py` (missing / null-non-admin / `[]` all return `[]`).
- Owner clause is dropped for public-only tokens: `mcpgateway/services/tool_service.py:2984-2996`

  ```python
  # Public-only tokens can ONLY see public resources - no owner access
  is_public_only_token = token_teams is not None and len(token_teams) == 0
  access_conditions = [DbTool.visibility == "public"]
  if not is_public_only_token and user_email:
      access_conditions.append(DbTool.owner_email == user_email)
  ```

  The same public-only pattern is applied across the other resource services (resources, prompts, servers).

---

### 🧠 Environment Info

| Key | Value |
|-----|-------|
| Version or commit | `epic/ui-rewrite @ current` |
| Runtime | Python 3.11 |
| Platform / OS | any |
| Container | none |

---

### 🧩 Additional Context

- Origin: #3291 ("[BUG][AUTH]: API Token lifecycle — chaining, broken inheritance...") reported this exact multi-team case; its fix only covered single-team auto-inheritance, leaving the multi-team path unaddressed.
- UI mitigation is tracked in #5370 (UI-rewrite API tokens, Story 2 — team selector on create).
- Related: #5732 (rename "Public" visibility label + tooltips) addresses adjacent naming confusion but not this silent-scoping behavior.
- The legacy admin UI mitigated this with a warning banner ("Tokens created without selecting a team are not scoped to any team... non-admin users can only see public resources"); the new UI and the raw API do not.

msureshkumar88:
## Summary

- Non-admin token creation with no explicit `team_id` previously only auto-inherited a team when the caller belonged to exactly one team. Multi-team users (and admin-delegated recipients) got a `teams: null` token, which normalizes to public-only — unable to see even the creator's own private resources.
- `TokenCatalogService.get_default_team_id()` resolves the caller's (or, for admin delegation, the target's) personal team. `routers/tokens.py::create_token` now defaults to it when the caller is a member, falls back to single-team inheritance, then to `team_id=None` plus a new `TokenCreateResponse.warnings` entry only when neither applies.
- Un-narrowed admins are unaffected — `team_id=None` stays a deliberate global-scope token for them.
- The permission-containment check (`_get_caller_permissions`) still uses the *requested* `team_id`, not the defaulted one, so this does not raise the ceiling on what `scope.permissions` a caller may request at creation time.
- Separately fixed `derive_token_team_id()` in `mcpgateway/auth.py` — the function that turns a single-team token's claim into `request.state.team_id` for RBAC/rate-limit/routing context — to exclude personal teams. A personal team auto-grants `team_admin`, so honouring it there would silently turn every personal-team-scoped token into a full-mutate credential; it now falls through to `check_any_team` instead, matching how `PermissionService._get_user_roles` already treats personal teams. This closes a corresponding blast-radius expansion that would otherwise ride along with the primary fix.

## Not in scope

- The legacy admin UI's team-scoping warning banner (`admin_ui/tokens.js` / `admin.html`) still describes the old all-teams-unscoped behavior; wording it accurately for the new default is left for the UI-rewrite team selector work tracked in #5370.
- No backfill for tokens already created with `team_id=null` before this fix — they must be recreated to pick up team scoping.

## Test plan

- [x] `make ruff interrogate pylint` — clean
- [x] `make test` — full suite passes
- [x] `make coverage diff-cover` — 99% diff coverage (only pre-existing, unrelated line missed)
- [x] `make detect-secrets-scan` — no new/live findings
- [x] New unit tests: multi-team default, no-personal-team warning, narrowed-session exclusion, admin-delegation default, permission-ceiling regression guard, `derive_token_team_id`/`_is_personal_team_sync` direct coverage

Closes #5993


mcpgateway/services/token_catalog_service.py
```diff
@@ -564,6 +564,45 @@ async def create_token(
...
+        Examples:
+            >>> from unittest.mock import Mock
+            >>> import asyncio
+            >>> service = TokenCatalogService(Mock())
+            >>> asyncio.iscoroutinefunction(service.get_default_team_id)
+            True
+        """
+        # First-Party
+        from mcpgateway.db import EmailTeam  # pylint: disable=import-outside-toplevel
+
```
prakhar-singh1928: Move all inline imports to top

mcpgateway/auth.py
```diff
@@ -300,6 +302,94 @@ def _get_team_name_by_id_sync(team_id: Optional[str]) -> Optional[str]:
...
+
+    Returns:
+        ``True`` when the team exists, is active, and is a personal team.
+    """
+    if not team_id:
+        return False
+
+    with fresh_db_session() as db:
+        # Third-Party
+        from sqlalchemy import select  # pylint: disable=import-outside-toplevel
```
prakhar-singh1928: same here

prakhar-singh1928:
@msureshkumar88 Move inline imports to top of files wherever applicable in all 10 changed files

mcpgateway/services/token_catalog_service.py
msureshkumar88: Moved — this import now uses the module's existing top-level `sqlalchemy`/`mcpgateway.db` imports (see 55fb0658a..fa4d4d3f3).

mcpgateway/auth.py
msureshkumar88: Moved — added `select`/`EmailTeam` to the top-level imports and dropped the inline ones in `_is_personal_team_sync`.

msureshkumar88:
Checked all 10 changed files — this PR only introduces 2 new inline imports (the ones flagged inline: `token_catalog_service.py:593` and `auth.py:319`), both now moved to the top in fa4d4d3f3. The other 8 files don't add any: `routers/tokens.py` and `schemas.py` have none, the test files' `from mcpgateway.auth import ...` lines are inside test functions (standard pytest local-import convention, not something this ask targets), and `.secrets.baseline`/`AGENTS.md` aren't code.

One note: `auth.py` has ~11 pre-existing inline `select`/`EmailTeam`/`EmailTeamMember` imports in sibling sync helpers (lines 239-266 etc., predating this PR) that still use the inline pattern. Left those alone as out-of-scope for #5993 — happy to file a follow-up if you'd like them hoisted too.

prakhar-singh1928 (dismissed):
LGTM

Lang-Akshay (changes requested):
## Blocking Changes

| 1 | Security | `mcpgateway/auth.py` | 389-390 | `derive_token_team_id()` catches a personal-team classification/DB failure and retains the personal `team_id`. That can enter the explicit-team RBAC path and expose the personal team's auto-granted `team_admin` permissions instead of failing closed. | Propagate the classification failure or return a fail-closed authorization outcome; do not retain a personal-team RBAC context after an indeterminate lookup. Add an operationally safe error/alert path. |
| 2 | Comment | `mcpgateway/auth.py`; `mcpgateway/services/token_catalog_service.py` | 321-324; 592 | Required reviewer requests to move inline imports to module scope remain unaddressed: the current diff adds `sqlalchemy.select`/`EmailTeam` imports inside both functions. | Hoist the newly added imports to the existing import blocks, then update the affected reviewer threads. |

msureshkumar88:
Thanks for the detailed review — both blocking findings were valid, fixed in ea6eac97f:

**#1 (security):** `derive_token_team_id()` now fails closed on a personal-team classification error — returns `None` (→ `check_any_team`) instead of retaining `team_id`. Retaining it on an indeterminate lookup could route RBAC context into a team never confirmed as non-personal, exposing that team's auto-granted `team_admin` role. Falling back to `None` only narrows effective permissions, never expands them, so it's the safe default under a transient DB failure. Updated the docstring and the regression test (`test_classification_error_fails_closed_and_returns_none`) to match.

**#2 (imports):** Confirmed the two new inline imports (`auth.py::_is_personal_team_sync`, `token_catalog_service.py::get_default_team_id`) were still present — the earlier "moved to top" reply referenced a commit that never actually landed on this branch. Hoisted both to module scope now. Doing that exposed `select`/`EmailTeam` name collisions with ~10 pre-existing local imports elsewhere in `auth.py` (would've dropped `make pylint --fail-under 10` to 9.77), so those redundant local reimports were also dropped — mechanical cleanup, no behavior change.

Rebased onto latest `main` and regenerated `.secrets.baseline`. Full `test_auth.py`, `test_token_catalog_service.py`, `test_token_catalog_service_cache.py`, and `test_tokens.py` suites pass; `make ruff pylint` clean on the touched files.

Lang-Akshay (approved):
LGTM ✅

PR #6355, opened 2026-08-21
feat: support catalog gateway actions in backend

### LINKED ISSUE #4731: [ENHANCEMENT]: Add Disconnect and Test actions to connected MCP catalog servers
## Overview

In the React Admin UI's MCP server catalog, registering ("connect") a catalog entry creates a `Gateway` record but the catalog view offers no follow-up actions on that entry. Add **Disconnect** and **Test** actions for catalog entries that are already registered, so users do not have to navigate away from the catalog to manage what they just connected.

**Related React catalog page:** #5770  
**Part of Epic:** #4093  
**Deferred to:** post-RC2 (after #4098)

## Context

The React catalog page consumes the v1 catalog API:

- `GET /v1/catalog` — list catalog entries and registration state
- `POST /v1/catalog/{catalog_id}/register` — connect a catalog entry

Once an entry is registered it becomes a regular `Gateway` row; there is no separate catalog lifecycle. Existing v1 endpoints cover the underlying actions:

- `DELETE /v1/gateways/{gateway_id}` — disconnect
- `POST /v1/mcp-servers/test` — test connectivity
- `POST /v1/gateways/{gateway_id}/tools/refresh` — live probe and tool rediscovery

Legacy `/admin/*` endpoints remain for the existing Admin UI, but the React implementation should use the v1 routes.

The blocker is that `CatalogServer` (`mcpgateway/schemas.py`) exposes `is_registered: bool` but no `gateway_id`, so the React app cannot resolve a registered catalog row to the gateway it should act on.

### Backend

- [ ] Add `gateway_id: Optional[str]` to the `CatalogServer` schema.
- [ ] Populate `gateway_id` in `catalog_service.get_catalog_servers()` from the same caller-visible gateway row used to derive registration state:
  - No visible URL match: `is_registered=false`, `gateway_id=null`, and `requires_oauth_config=false`.
  - Visible URL match: `is_registered=true`, `gateway_id=<matched gateway id>`, and `requires_oauth_config` derived from that same row.
- [ ] Preserve Layer-1 visibility rules for every catalog-list route. A caller must never receive a `gateway_id` for a gateway outside their public/team/private visibility scope. Update legacy `/admin/mcp-registry/servers` and catalog partial reads to pass the authenticated scope into `get_catalog_servers()` rather than using unscoped registration state.
- [ ] Handle multiple caller-visible gateways with the same URL deterministically. Selection precedence:
  1. caller-owned gateway created via `catalog`;
  2. any caller-owned gateway;
  3. any gateway created via `catalog`;
  4. stable gateway-ID ordering.
  `gateway_id`, `is_registered`, and `requires_oauth_config` must all describe the selected row.
- [ ] Invalidate the catalog registry cache after every permanent gateway deletion, regardless of `created_via`, because catalog registration state is derived from URL matches for all gateways. Put this in the shared post-delete finalizer so both admin/v1 and synchronous/asynchronous deletion paths are covered.
- [ ] When asynchronous gateway lifecycle is enabled, keep the entry registered after `202 Accepted`; invalidate the catalog only after the lifecycle worker commits the hard deletion.
- [ ] Preserve existing catalog-cache invalidation after successful registration.
- [ ] Add regression coverage for matched/unmatched IDs, OAuth-pending state, Layer-1 deny paths, duplicate-URL selection, a warmed-cache delete flow, and synchronous/asynchronous deletion finalization.

### Frontend (React prototype)

- [ ] Show a contextual action menu on each catalog card based on `is_registered`:
  - Not registered → existing **Connect** action.
  - Registered → **Disconnect** and **Test** actions.
- [ ] **Disconnect**: confirm dialog → `DELETE /v1/gateways/{gateway_id}` → refetch catalog list.
- [ ] If DELETE returns `202 Accepted`, poll gateway lifecycle status until permanent deletion completes before refetching the catalog.
- [ ] **Test**: `POST /v1/mcp-servers/test` with the catalog entry URL and the required `GatewayTestRequest` fields → surface status code and latency from `GatewayTestResponse`.
- [ ] Disable **Test** when `requires_oauth_config === true` (gateway exists but is `enabled=false` pending OAuth).
- [ ] Respect permissions: only render Disconnect with `gateways.delete`, Test with `gateways.read`. Surface ownership-related `403` responses without changing catalog state.

## Acceptance Criteria

- [ ] Registered catalog entries return a caller-visible, deterministic `gateway_id`; unregistered entries return `gateway_id=null`.
- [ ] Catalog responses never expose gateway IDs outside the caller's Layer-1 visibility scope.
- [ ] Registered catalog entries surface Disconnect and Test actions in the React UI.
- [ ] Disconnecting from the catalog view removes the selected underlying gateway and updates the catalog card state without a manual refresh, including asynchronous deletion mode.
- [ ] Deleting any gateway that changes URL-derived catalog registration state cannot leave a stale cached catalog response.
- [ ] Test action shows a clear success/failure indicator with latency.
- [ ] Actions are hidden or disabled for users without the relevant permissions, and Test is disabled for OAuth entries that have not completed configuration.
- [ ] No regression in the existing Connect flow.

## Optional follow-up

To list affected virtual servers in the disconnect confirmation, add a protected, Layer-1-scoped gateway impact-preview API that returns the virtual server IDs/names associated through the gateway's tools, resources, and prompts. This is not required for the base acceptance criteria.

vishu-bh:
# Pull Request

## 🔗 Related Issue

Closes #4731

---

## 📝 Summary

Adds backend support needed for connected catalog actions:

- exposes caller-visible `gatewayId` in catalog responses
- selects duplicate URL matches deterministically
- preserves Layer-1 visibility across v1 and legacy catalog reads
- invalidates catalog cache only after permanent gateway deletion
- adds protected `GET /v1/gateways/{gateway_id}/impact-preview` for affected virtual servers

---

## 📏 Reviewability

---

## 🏷️ Type of Change

- [ ] Bug fix
- [x] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

| Check | Command | Status |
|---|---|---|
| Pre-commit suite | `make pre-commit` | Passed |
| Full test suite | `make test` | 22,222 passed |
| Doctests | `make doctest` | Passed |
| Package verification | `make verify` | Passed (10/10) |

---

## ✅ Checklist

- [x] Code formatted (`make pre-commit`)

---

## 📓 Notes (optional)

Backend scope only. Impact preview returns unique virtual server IDs and names visible to caller through existing Layer-1 rules.


gandhipratik203:
**F2. Impact preview hits the DB once per server** *(functionally impacting)*

`_check_server_access` looks up the user's teams every time it's called when `token_teams is None`. A gateway with 50 virtual servers means 50 extra queries on one request. Look up the teams once before the loop and pass them in as `token_teams`.

**F3. The catalog cache no longer does anything** *(functionally impacting)*

Reads and writes are skipped whenever the request is scoped, and the `admin.py` callers were the last unscoped ones. So nothing fills the cache anymore, which makes the new `invalidate_catalog()` pointless, and `catalog_partial` now scans the whole `Gateway` table twice per page load. Keep the scoping, it's a real fix. Just add the scope to the cache key: include `user_email` and sorted `token_teams` in `hash_filters`.

vishu-bh:
Thanks for flagging this. I’ve updated the impact-preview path to resolve fallback database team memberships once before iterating over the affected servers, then reuse those memberships for each access check.
I kept the resolved memberships separate from `token_teams`, because `token_teams=None` represents admin bypass while `token_teams=[]` represents public-only scope. Replacing that value directly could change admin or private-owner visibility semantics.
I also added regression coverage for multiple-server reuse, admin bypass, public-only denial, and private-owner access.

gandhipratik203 (dismissed):
LGTM! Thanks for the fixes!

gandhipratik203 (dismissed):
LGTM!

Lang-Akshay (changes requested):
## Blocking Changes

| 1 | Comment / Quality | `mcpgateway/admin.py` | 17930-17943, 18131-18143 | Reviewer comment F3 is only partially addressed. Passing authenticated scope into both catalog calls correctly fixes visibility, but makes normal scoped requests skip both catalog-cache reads and writes; `catalog_partial` then performs the gateway scan twice per page request. | Keep Layer-1 scoping and make the cache scope-aware (for example, include `user_email` and a normalized `token_teams` value in `hash_filters`), or otherwise provide a scoped cached/count path so the new visibility fix does not create a repeat full-scan regression. |

gandhipratik203 (dismissed):
LGTM!

Lang-Akshay (dismissed):
LGTM ✅

Lang-Akshay (approved):
LGTM

PR #6362, opened 2026-08-21
[Feature]: SSL/TLS exposed for gateway pods

### LINKED ISSUE #6265: [FEATURE]: Helm chart does not expose SSL/TLS env vars for direct gateway HTTPS (non-ingress TLS)
## Summary

The Helm chart (`charts/mcp-stack`) exposes TLS configuration at the **ingress** layer (`ingress.tls`, `nginxProxy.tls`, OpenShift `route.tls`) but does not expose the gateway pod's own SSL env vars (`SSL`, `CERT_FILE`, `KEY_FILE`, `KEY_FILE_PASSWORD`) in `values.yaml`.

This means that for a Kubernetes deployment using the end-to-end TLS topology (nginx → HTTPS → gateway pod), an operator must inject the SSL env vars manually via `extraEnv` or patch the deployment, rather than using first-class chart values.

## Current State

`values.yaml` contains Redis TLS vars (e.g. `RATELIMITER_REDIS_SSL`, `RATELIMITER_REDIS_SSL_CA_CERTS`) and `SKIP_SSL_VERIFY` as first-class values. The equivalent gateway SSL vars are absent.

The `deployment.yaml` template does not reference `SSL`, `CERT_FILE`, or `KEY_FILE` in the gateway container's env block.

## Proposed Changes

Add a `gateway.tls` section to `values.yaml`:

```yaml
gateway:
  tls:
    enabled: false             # Set true to enable direct HTTPS on the gateway pod
    certFile: /app/certs/tls.crt
    keyFile: /app/certs/tls.key
    keyFilePassword: ""       # Leave empty for unencrypted key
    # For inbound mTLS (client certificate authentication):
    caCerts: ""               # Path to CA bundle inside the container
    certReqs: 0               # 0=none, 1=optional, 2=required
```

When `gateway.tls.enabled=true`, the deployment template should:
1. Set `SSL=true`, `CERT_FILE`, `KEY_FILE` (and optionally `KEY_FILE_PASSWORD`, `CA_CERTS`, `CERT_REQS`) in the container env
2. Reference a Kubernetes TLS Secret for the cert/key volume mount
3. Update the pod liveness/readiness probe to use `https` scheme

This would allow both:
- **Ingress TLS only** (current, `gateway.tls.enabled=false`, nginx or route terminates TLS)
- **End-to-end TLS** (`gateway.tls.enabled=true` + `nginxProxy.tls.enabled=true`)

## Acceptance Criteria

- [ ] `values.yaml` has a `gateway.tls` section (all fields default to disabled/empty)
- [ ] `deployment.yaml` conditionally injects SSL env vars when `gateway.tls.enabled=true`
- [ ] A Kubernetes TLS Secret is referenced for the cert/key mount
- [ ] Liveness/readiness probes use `https` scheme when `gateway.tls.enabled=true`
- [ ] `docs/docs/deployment/kubernetes.md` and `docs/docs/deployment/helm.md` document the end-to-end TLS configuration path

claudia-gray:
# Pull Request

## 🔗 Related Issue

Closes https://github.com/IBM/mcp-context-forge/issues/6265

---

## 📝 Summary

The Helm chart exposed TLS only at the ingress/nginx layer. Operators who needed end-to-end TLS (nginx → HTTPS → gateway pod) had to manually inject SSL, CERT_FILE, and KEY_FILE via extraEnv and patch the volume mount themselves — no first-class support existed.

This PR adds mcpContextForge.tls to values.yaml and wires it through deployment-mcpgateway.yaml so that env vars, the cert volume mount, and the probe scheme are all handled automatically when mcpContextForge.tls.enabled=true.

---

## 📏 Reviewability

---

## 🏷️ Type of Change

- [ ] Bug fix
- [x] Feature / Enhancement
- [x] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

### Template rendering (no cluster required)

Check | Command | Status
-- | -- | --
helm lint --strict | helm lint --strict charts/mcp-stack | ✅ 1 chart linted, 0 failed
helm-unittest | make linting-helm-unittest | ✅ 42/42 tests pass (5 suites)
TLS env vars injected | helm template ... --set mcpContextForge.tls.enabled=true --set mcpContextForge.tls.secretName=my-tls \| grep -E "SSL\|CERT_FILE\|KEY_FILE" | ✅ SSL, CERT_FILE, KEY_FILE present
KEY_FILE_PASSWORD uses secretKeyRef | ... --set mcpContextForge.tls.keyFilePasswordSecret.name=my-pass \| grep -A 4 KEY_FILE_PASSWORD | ✅ valueFrom.secretKeyRef, no plain value:
Volume + volumeMount | ... \| grep -A 3 "gateway-tls-certs" | ✅ Secret volume + read-only mount at /app/certs/tls
All three probes flipped | ... \| grep "scheme:" | ✅ scheme: HTTPS ×2 readiness + liveness; startup probe gets scheme: HTTPS when type: http
Default unchanged | helm template ... \| grep -E "SSL\|CERT_FILE\|gateway-tls-certs" | ✅ No TLS artefacts when tls.enabled=false
required guard | helm template ... --set mcpContextForge.tls.enabled=true (no secretName) | ✅ Clear error at render time
Nil guard | helm template ... --set mcpContextForge.tls=null | ✅ Renders cleanly, no error

### Runtime verification — kind cluster (k8s v1.36.1, Helm v4.1.3)

Runtime verification — kind cluster (k8s v1.36.1, Helm v4.1.3)

# Create self-signed cert Secret
openssl req -x509 -newkey rsa:2048 -nodes \
  -keyout tls.key -out tls.crt -days 365 -subj "/CN=gateway.local"
kubectl create secret tls my-tls-secret --cert=tls.crt --key=tls.key

# Install with TLS enabled
helm install my-release charts/mcp-stack \
  --set mcpContextForge.tls.enabled=true \
  --set mcpContextForge.tls.secretName=my-tls-secret \
  --set mcpContextForge.metrics.serviceMonitor.enabled=false \
  --set mcpContextForge.secret.JWT_SECRET_KEY="$(openssl rand -hex 32)" \
  --set mcpContextForge.secret.AUTH_ENCRYPTION_SECRET="$(openssl rand -hex 32)"

# Port-forward and verify HTTPS
kubectl port-forward \
  pod/$(kubectl get pod -l app=my-release-mcp-stack-mcpgateway \
        -o jsonpath='{.items[0].metadata.name}') 4444:4444 &
curl -k https://localhost:4444/health

Result: {"status":"healthy", …} returned over HTTPS. curl http://localhost:4444/health returns an SSL protocol error, confirming TLS-only mode is active.

---

## ✅ Checklist

---

## 📓 Notes (optional)

Why mcpContextForge.tls and not gateway.tls?
The gateway's entire config lives under mcpContextForge. A top-level tls: key already exists for the nginx frontend TLS profile — a second top-level tls would be ambiguous.

KEY_FILE_PASSWORD uses a Secret reference, not a plain string.
The original proposal had keyFilePassword: "". A plain string surfaces in helm get values, kubectl get pod -o yaml, and any values file committed to git. Replaced with keyFilePasswordSecret: { name, key } which follows the same valueFrom.secretKeyRef pattern already used by POSTGRES_PASSWORD and REDIS_PASSWORD in this template.

No cert generation in the chart.
The operator pre-creates a kubernetes.io/tls Secret (via cert-manager or kubectl create secret tls). Chart-side cert generation is out of scope.

mTLS (CA_CERTS / CERT_REQS) is out of scope.
run-gunicorn.sh has no consumer for those env vars — advertising them would create a silently-ineffective security control. Dropped from this PR; tracked separately.

_helpers.tpl fix.
The pre-existing {{- if $p.scheme }}scheme: {{ $p.scheme }}{{ end }} used {{- which stripped the preceding newline, producing port: 4444scheme: HTTP — invalid YAML as soon as scheme was passed. Fixed to put scheme: on its own line. Also extracted helpers.renderProbeWithTLS so the HTTPS-scheme logic lives in one place and all three probe call sites (startup, readiness, liveness) share it.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -618,7 +618,7 @@
         "hashed_secret": "25ab86bed149ca6ca9c1c0d5db7c9a91388ddeab",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 979,
+        "line_number": 1021,
         "type": "Basic Auth Credentials",
         "verified_result": null
       },
@@ -626,7 +626,7 @@
         "hashed_secret": "d08f88df745fa7950b104e4a707a31cfce7b5841",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1083,
+        "line_number": 1125,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -634,7 +634,7 @@
         "hashed_secret": "7288edd0fc3ffcbe93a0cf06e3568e28521687bc",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1086,
+        "line_number": 1128,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -642,7 +642,7 @@
         "hashed_secret": "8674c9b302d20800e4ab3808f139704d8641a6e3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1253,
+        "line_number": 1295,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -650,7 +650,7 @@
         "hashed_secret": "cff0d14e4337fa8bdb68dfa906f04b0df6fad72f",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1292,
+        "line_number": 1334,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -658,7 +658,7 @@
         "hashed_secret": "f865b53623b121fd34ee5426c792e5c33af8c227",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1346,
+        "line_number": 1388,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -666,7 +666,7 @@
         "hashed_secret": "acde39840735314af1300688b6c2324ea89770a3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1442,
+        "line_number": 1484,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -674,7 +674,7 @@
         "hashed_secret": "fa9beb99e4029ad5a6615399e7bbae21356086b3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1813,
+        "line_number": 1855,
         "type": "Secret Keyword",
         "verified_result": null
       }
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(?x)( package-lock\\.json$ |Cargo\\.lock$ |uv\\.lock$ |go\\.sum$ |mcpgateway/sri_hashes\\.json$ )|^\\.secrets\\.baseline$",
     "lines": null
   },
-  "generated_at": "2026-08-25T12:48:17Z",
+  "generated_at": "2026-08-26T09:08:27Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -615,10 +615,18 @@
     ],
     "charts/mcp-stack/values.yaml": [
       {
+        "hashed_secret": "7bbf4d709be627d062b1e760bf712e132f62e839",
+        "is_secret": false,
+        "is_verified": false,
+        "line_number": 217,
+        "type": "Secret Keyword",
+        "verified_result": null
+      },
+      {
         "hashed_secret": "25ab86bed149ca6ca9c1c0d5db7c9a91388ddeab",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 979,
+        "line_number": 1015,
         "type": "Basic Auth Credentials",
         "verified_result": null
       },
@@ -626,7 +634,7 @@
         "hashed_secret": "d08f88df745fa7950b104e4a707a31cfce7b5841",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1083,
+        "line_number": 1119,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -634,7 +642,7 @@
         "hashed_secret": "7288edd0fc3ffcbe93a0cf06e3568e28521687bc",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1086,
+        "line_number": 1122,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -642,7 +650,7 @@
         "hashed_secret": "8674c9b302d20800e4ab3808f139704d8641a6e3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1253,
+        "line_number": 1289,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -650,7 +658,7 @@
         "hashed_secret": "cff0d14e4337fa8bdb68dfa906f04b0df6fad72f",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1292,
+        "line_number": 1328,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -658,7 +666,7 @@
         "hashed_secret": "f865b53623b121fd34ee5426c792e5c33af8c227",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1346,
+        "line_number": 1382,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -666,7 +674,7 @@
         "hashed_secret": "acde39840735314af1300688b6c2324ea89770a3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1442,
+        "line_number": 1478,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -674,7 +682,7 @@
         "hashed_secret": "fa9beb99e4029ad5a6615399e7bbae21356086b3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1813,
+        "line_number": 1849,
         "type": "Secret Keyword",
         "verified_result": null
       }
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -618,7 +618,7 @@
         "hashed_secret": "25ab86bed149ca6ca9c1c0d5db7c9a91388ddeab",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1021,
+        "line_number": 979,
         "type": "Basic Auth Credentials",
         "verified_result": null
       },
@@ -626,7 +626,7 @@
         "hashed_secret": "d08f88df745fa7950b104e4a707a31cfce7b5841",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1125,
+        "line_number": 1083,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -634,7 +634,7 @@
         "hashed_secret": "7288edd0fc3ffcbe93a0cf06e3568e28521687bc",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1128,
+        "line_number": 1086,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -642,7 +642,7 @@
         "hashed_secret": "8674c9b302d20800e4ab3808f139704d8641a6e3",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 1295,
+        "line_number": 1253,
         "type": "Secret Keyword",
         "verified_result": null
       },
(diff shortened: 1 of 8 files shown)


msureshkumar88 (changes requested):
Thanks for taking this on — it's a genuinely useful gap to close, and the chart-side approach is idiomatic. Putting the block under `mcpContextForge.tls` rather than a second top-level `tls:` is the right call and the reasoning in your PR notes is sound. The `required` guard on `secretName` is a nice touch, the `values.schema.json` registration is good practice that's easy to forget, and the `_helpers.tpl` newline fix is a real latent bug — `port: 4444scheme: HTTP` would have been invalid YAML the moment anything passed a scheme, so that's load-bearing for this work rather than scope creep.

I rendered the chart locally across a few flag combinations to check the behaviour. Default rendering (`tls.enabled=false`) produces zero TLS artefacts, so existing deployments are genuinely unaffected — the non-breaking claim holds up. A few things below that I think are worth addressing before merge.

## Blocking

**1. `KEY_FILE_PASSWORD` is rendered as a plaintext env value**

In `deployment-mcpgateway.yaml` the passphrase comes through as `value: {{ ... | quote }}`, sourced from a plain string in `values.yaml`. That puts the private-key passphrase into `kubectl get pod -o yaml`, `helm get values`, and any values file that lands in git or CI logs — which somewhat undercuts the benefit of keeping the cert and key in a Secret.

The same template already has the pattern for this: `POSTGRES_PASSWORD` and `REDIS_PASSWORD` both use `valueFrom.secretKeyRef`. `charts/AGENTS.md` also points the same direction ("Prefer external `Secret` objects for production", "Never commit secrets to values files"). Would you be open to replacing the string field with a secret reference?

```yaml
{{- if .Values.mcpContextForge.tls.keyFilePasswordSecret }}
- name: KEY_FILE_PASSWORD
  valueFrom:
    secretKeyRef:
      name: {{ .Values.mcpContextForge.tls.keyFilePasswordSecret.name }}
      key:  {{ .Values.mcpContextForge.tls.keyFilePasswordSecret.key | default "keyFilePassword" }}
{{- end }}
```

**2. `CA_CERTS` / `CERT_REQS` are injected but never consumed**

This one is worth flagging carefully because of what the docs promise. The new `## Kubernetes / Helm` section in `tls-configuration.md` includes an mTLS subsection telling operators that `certReqs: 2` will reject connections without a valid client certificate. Tracing the actual runtime path, though:

- `run-gunicorn.sh` is the only consumer of `SSL` / `CERT_FILE` / `KEY_FILE`, and when `SSL=true` it appends only `--certfile` and `--keyfile` to the gunicorn invocation
- `gunicorn.config.py` reads only `SSL` and `SSL_KEY_PASSWORD`
- `grep -rn 'CA_CERTS\|CERT_REQS'` across the repo returns no consumer in project source

So the env vars land in the pod and nothing reads them. The practical effect is that an operator who follows the mTLS docs to require client certificates gets a gateway that still accepts unauthenticated TLS connections, with no error or warning to indicate the setting was ignored. That's the part that makes this blocking rather than cosmetic — a silently-ineffective security control is worse than an absent one.

Two reasonable ways out, whichever you prefer:

- **Scope down** — drop `caCerts` / `certReqs` from `values.yaml`, `values.schema.json`, and the template, and remove the mTLS section from the docs. Open a follow-up issue for real mTLS support. Keeps this PR tight and shippable.
- **Scope up** — wire it through in `run-gunicorn.sh`, roughly:

```bash
[[ -n "${CA_CERTS}" ]]  && cmd+=( --ca-certs "${CA_CERTS}" )
[[ -n "${CERT_REQS}" ]] && cmd+=( --cert-reqs "${CERT_REQS}" )
```

with file-existence validation for `CA_CERTS` mirroring the existing `CERT_FILE` / `KEY_FILE` checks.

Happy either way — the first is probably the faster path to merge.

## Functionally impacting

**3. `startupProbe` doesn't get the HTTPS scheme**

`readinessProbe` and `livenessProbe` both get `merge (dict "scheme" "HTTPS")`, but the non-migration `startupProbe` branch calls `helpers.renderProbe` without it. Rendering with `tls.enabled=true`, `migration.enabled=false`, and an http-type startup probe:

```yaml
startupProbe:
  httpGet: {path: /ready, port: 4444}          # no scheme → HTTP
readinessProbe:
  httpGet: {path: /ready, port: 4444, scheme: HTTPS}
livenessProbe:
  httpGet: {path: /health, port: 4444, scheme: HTTPS}
```

The default `probes.startup` is `type: exec`, so this doesn't bite out of the box — but the helper supports `type: http` for startup, and gunicorn with `--certfile/--keyfile` serves TLS only with no plaintext fallback. Anyone who switches the startup probe to http alongside TLS gets a pod that never passes startup, and since kubelet gates readiness and liveness behind startup success, it restart-loops without a very obvious cause.

Since the same `if/else` block now appears three times, it might be worth extracting a small `helpers.renderProbeTLS` partial — fixes this in one place and removes the duplication.

**4. No chart unit tests for the new templating**

`charts/mcp-stack/tests/` already has helm-unittest suites, and `linting-helm-unittest` is part of `LINTING_FULL_TARGETS`, so there's a harness ready to use (it gates in the merge queue rather than per-push, so this isn't urgent-urgent, but the coverage is currently zero). The manual `helm template | grep` checks in the PR description are exactly the right assertions — they'd just be more valuable committed and repeatable. Something like a `deployment_mcpgateway_tls_test.yaml` covering:

- TLS env vars render when `tls.enabled=true`
- Secret volume and read-only mount are present
- All three probes carry `scheme: HTTPS`
- `required` guard errors when `secretName` is missing
- No TLS artefacts when `tls.enabled=false`

The last one in particular would lock in the non-regression property you verified manually.

## Minor

**5. Nil guard on the `tls` block**

`helm template --set mcpContextForge.tls=null` fails with `nil pointer evaluating interface {}.enabled`. Helm merges chart defaults so normal installs and upgrades are safe — this only affects someone explicitly nulling the block. `(.Values.mcpContextForge.tls).enabled` or a `default dict` would make it robust if you want to harden it.

**6. Docs**

Assuming the mTLS section stays (option (b) above), it reads well. If you go with option (a), that subsection and the `caCerts` / `certReqs` rows should come out with it so the docs don't describe fields that no longer exist.

---

Nothing here is structural — the design is right and most of it is small mechanical follow-through. Items 1 and 2 are the ones I'd want resolved before this goes in; 3 and 4 would be good to fold in while you're in here. Let me know which direction you want to take on the mTLS scope question and I'm happy to look again.

claudia-gray:
@msureshkumar88 I have addressed and included issues 1,3,4,5 and 6. 2 I decided it is out of scope for this and did not include it but removed relevant parts. Thank you for the review.

msureshkumar88 (dismissed):
Re-reviewed against the latest commits (verified via diff, not just commit messages) — all previously raised blocking items are genuinely resolved:

- `KEY_FILE_PASSWORD` no longer has a plaintext path; it's `secretKeyRef`-only now, consistent with the existing `POSTGRES_PASSWORD`/`REDIS_PASSWORD` pattern.
- The unwired `caCerts`/`certReqs` (mTLS) values have been dropped entirely rather than left as dead, falsely-advertised config — good call, since `run-gunicorn.sh` never consumed them.
- Startup probe now goes through the same `helpers.renderProbeWithTLS` helper as readiness/liveness, so it correctly flips to `HTTPS` too.
- The nil-guard (`$tls := default dict .Values.mcpContextForge.tls`) and the new `deployment_mcpgateway_tls_test.yaml` suite (15 cases: disabled state, env vars, volume/mount, all three probe schemes, required-guard) give solid coverage.

I rendered the chart locally to confirm behavior end-to-end (TLS-enabled render produces the expected env vars/volume/HTTPS scheme, missing `secretName` fails fast with the documented error, `tls: null` renders cleanly), and confirmed `run-gunicorn.sh` genuinely consumes `SSL`/`CERT_FILE`/`KEY_FILE`/`KEY_FILE_PASSWORD` so this isn't just chart plumbing — it works.

Scope matches the linked issue well; narrowing the values a operator to only the vars the entrypoint actually consumes (vs. the issue's literal proposal) is a better outcome than a 1:1 implementation. No schema/alembic changes needed, none made. No breaking change — disabled by default, additive only.

Two very minor, non-blocking suggestions for a follow-up if you're touching this file again:
- The "Probe scheme" note in `tls-configuration.md` only mentions readiness/liveness — could mention the startup probe now flips too.
- An automated `it:` case for `mcpContextForge.tls: null` would round out the otherwise thorough test suite (currently only verified manually per the PR description).

Nice work tightening this up across the review cycle.

gandhipratik203 (dismissed):
LGTM! No blocking findings. This should enhance the capability of our deplyment charts.

msureshkumar88 (approved):
Rebased on main and resolved `.secrets.baseline` conflicts (regenerated + audited the one unaudited finding, a placeholder example in `values.yaml`'s comment). All CI checks pass. LGTM.

msureshkumar88 (approved):

Issue #5127, opened 2026-06-08
[FEATURE]: Admin UI for Async Gateway Lifecycle Operations


ja8zyjits:
### 🧭 Type of Feature

Please select the most appropriate category:

- [x] Enhancement to existing functionality
- [ ] New feature or capability
- [ ] New MCP-compliant server
- [ ] New component or integration
- [ ] Developer tooling or test improvement
- [ ] Packaging, automation and deployment (ex: pypi, docker, quay.io, kubernetes, terraform)
- [ ] Other (please describe below)

### 🧭 Epic

**Title:** Admin UI for Async Gateway Lifecycle Operations

**Goal:** Provide visual feedback and status tracking in the Admin UI for async gateway lifecycle operations (create, update, delete) when `GATEWAY_ASYNC_LIFECYCLE_ENABLED=true`.

**Why now:** Issue #4565 implements backend async gateway lifecycle with 202 Accepted responses and background worker processing. The Admin UI needs corresponding updates to display gateway status (pending, active, deleting), retry metadata, and provide polling behavior for pending operations.

---

### 🧑🏻‍💻 User Story 1: Gateway Status Visibility

**As a:** Platform administrator using the Admin UI

**I want:** Visual indicators showing gateway status (pending, active, deleting) with retry metadata

**So that:** I can monitor gateway lifecycle operations and identify stuck or failing gateways without checking logs

#### ✅ Acceptance Criteria
```gherkin
Scenario: Display status badges for gateway states
  Given GATEWAY_ASYNC_LIFECYCLE_ENABLED is set to true
  And I am viewing the gateways list in Admin UI
  When a gateway has status "pending"
  Then I see a yellow/amber status badge with "Pending" label
  And I see retry metadata (registration_attempts, next_retry_at)
  When a gateway has status "active"
  Then I see a green status badge with "Active" label
  When a gateway has status "deleting"
  Then I see an orange status badge with "Deleting" label

Scenario: Display status messages for pending gateways
  Given a gateway with status "pending"
  And status_message is "Retrying after error (attempt 3)"
  When I view the gateway in Admin UI
  Then I see the status_message displayed prominently
  And I see the retry attempt count
  And I see the next retry timestamp

Scenario: Hide internal error details
  Given a gateway with last_error field populated
  When I view the gateway in Admin UI
  Then I do NOT see the last_error field
  And I only see the user-facing status_message
```

---

### 🧑🏻‍💻 User Story 2: Automatic Status Polling

[...]

PR #6368, opened 2026-08-22
fix(gateways): persist health-check failure reason

### LINKED ISSUE #6343: Gateway health-check failures discard the error reason
### Bug Summary

When the periodic health check marks a gateway unreachable, the failure reason is discarded — it reaches `logger.debug` and nothing else. The `Gateway` row already has `last_error` and `status_message` columns for exactly this, but they are only ever populated by the *registration* lifecycle, never by the health check. The result is that neither the API nor any UI can answer "why is this MCP server offline?"

---

### Affected Component

- [x] `mcpgateway` - API
- [ ] Other (explain below)

---

### Steps to Reproduce

1. Register a gateway pointing at a reachable MCP server.
2. Break it in a way with a specific, diagnosable cause such as an expired TLS cert, wrong bearer token, DNS failure, connection refused.
3. Wait for the health check to mark it unreachable.
4. `GET /gateways/{id}`.

---

### Expected Behavior

`last_error` / `status_message` reflect the actual health-check failure, e.g. `certificate verify failed: certificate has expired`.

### Actual Behavior

`reachable: false`, and `last_error` / `status_message` are `null` — or worse, hold a **stale string from a past registration failure** that has nothing to do with the current outage. Nothing distinguishes "cert expired" from "DNS gone" from "connection refused."

---

### Root cause

The exception is caught and dropped at `mcpgateway/services/gateway_service.py:5069`:

```python
if not is_auth_failure:
    # Genuine connectivity failure - mark gateway unhealthy
    ...
    logger.debug("Health check failed for gateway %s: %s", gateway_name, e)
    await self._handle_gateway_failure(gateway)
```

`_handle_gateway_failure(gateway)` (`:4530`) takes the gateway only — `e` is never passed. It counts failures and, at threshold, calls `set_gateway_state(..., reachable=False, only_update_reachable=True)`, which writes `enabled` / `reachable` / `updated_at` and never touches `last_error` or `status_message`.

Meanwhile the columns exist and are already exposed. `Gateway` (in `db.py`) has `status`, `status_message`, `last_error`, `registration_attempts`, `next_retry_at`, and `GatewayRead` surfaces all of them (`schemas.py:3959-3963`). The only writer is `_finalize_pending_gateway_failure` (`:4396`), which does the right thing — sanitized error string plus a `next_retry_at` — but only for the pending/registration path.

Two consequences:

- Nothing is written on the health-check path, so an ordinary outage has no recorded cause.
- Nothing is cleared either. A gateway that failed registration, later succeeded, and later still went offline can present a `last_error` from the original registration failure alongside `reachable: false` from an unrelated current outage. `_finalize_pending_gateway_success` clears the fields, but only on that same pending path.

---

### Suggested fix

1. Thread the exception into `_handle_gateway_failure(gateway, error=e)` and sanitize it the same way `_finalize_pending_gateway_failure` already does (`SecurityValidator`-style scrubbing — these strings can contain URLs and auth material, so this must not be a raw `str(e)`).
2. Persist it to `last_error` / `status_message` when the threshold trips and the gateway is marked unreachable.
3. Clear both fields in `_mark_gateway_reachable` (`:4697`), so a recovered gateway doesn't carry a stale reason.

Worth deciding as part of this: whether `status_message` should stay registration-lifecycle-only and a separate field carry health-check reasons. Overloading one field means the async-lifecycle UI work in #5127 and any "why is this offline?" UI end up reading the same column for two different meanings. A distinct `last_health_error` may be cleaner than reusing `last_error`.

---

### Related

- #5127: Admin UI for async gateway lifecycle ops, which consumes `status` / `status_message` / retry metadata and would be affected by the field-overloading question above
- #4565: async gateway lifecycle backend that introduced these columns
- Companion issue on the activity feed rendering these events https://github.com/IBM/mcp-context-forge/issues/6342
- Consumer: contextforge-org/contextforge-web-ui#60

Line references verified 2026-09-08 against `mcp-context-forge` at `3865378`.

feiiiiii5:
# 🐛 Bug-fix PR

## 📌 Summary

Periodic gateway health checks marked gateways unreachable but discarded the underlying exception after a debug log. Consequently, `last_error` could remain `null` or retain an unrelated registration failure, leaving operators unable to distinguish an expired certificate from DNS failure or connection refusal.

This change persists a sanitized health-check failure reason and clears it when the gateway recovers.

Closes #6343

## 🔁 Reproduction Steps

1. Register a reachable gateway.
2. Make its upstream fail with a specific connectivity error.
3. Let periodic health checks mark it unreachable.
4. Read the gateway: `reachable` becomes false while `last_error` remains stale or empty.

## 🐞 Root Cause

`_check_single_gateway_health` logged the failure and called `_handle_gateway_failure(gateway)`, but did not pass the exception. `_handle_gateway_failure` only updated reachability through `set_gateway_state(..., only_update_reachable=True)` and never wrote `last_error`. Recovery also updated `last_seen` without clearing a previous error.

## 💡 Fix Description

- Pass the unwrapped health-check exception into `_handle_gateway_failure`.
- Sanitize it with the existing `sanitize_exception_message` helper and the gateway's auth query params before persistence.
- Write the sanitized reason to the existing `last_error` column when the failure threshold trips.
- Clear `last_error` whenever a successful health check updates `last_seen`.

`status_message` is intentionally unchanged so registration lifecycle text does not become overloaded with health-check state.

## 📏 Reviewability

## 🧪 Verification

Pre-fix: copied the new tests onto a clean `fork/main` worktree; both new behavior tests failed (`_handle_gateway_failure` did not accept the exception, and recovery retained the old error).

Post-fix on this branch:

```bash
python -m pytest -q tests/unit/mcpgateway/services/test_gateway_service.py -k 'TestGatewayHealth or TestHandleGatewayFailure or TestMarkGatewayReachableErrorCleanup or test_health_check_failure'
# 21 passed
```

Additional checks:

```bash
uvx ruff@0.16.1 check mcpgateway/services/gateway_service.py
# All checks passed
```

The existing test file has 41 pre-existing Ruff findings on both `main` and this branch; this PR adds none. Full project lint/test/coverage were not duplicated locally due to the large toolchain dependency set; CI will supply that result.

| Changed source lint                   | Ruff 0.16.1          | Pass |
| Targeted regression                   | 21 tests             | Pass |
| Pre-fix regression                    | Clean main worktree  | Fail as expected |
| Full lint / test / coverage           | `make lint` / CI     | Pending CI |

## ✅ Checklist

- [x] Code formatted


marekdano (changes requested):
@feiiiiii5 - thanks for the contribution!

### Verdict

Solid, well-tested fix with good verification notes from the author. Fix the OAuth `client_credentials` gap before merge; everything else looks in good shape.

### 🔴 Blocking / Functionally-Impacting

  **Missed call site: OAuth `client_credentials` token failure doesn't thread the exception**

  In `_check_single_gateway_health`, when `oauth_manager.get_access_token()` fails, the handler still calls:

  ```python
  await self._handle_gateway_failure(gateway)
  ```

  at **`mcpgateway/services/gateway_service.py:4741`** — untouched by this diff. Compare to the two sibling call sites the PR *did* update:

  ```python
  await self._handle_gateway_failure(gateway, error=asyncio.TimeoutError(...))   # line ~4505
  await self._handle_gateway_failure(gateway, error=exc_to_inspect)              # line ~4899
  ```

  Since `error` defaults to `None`, once this path trips the failure threshold, `last_error` will be persisted as the generic `"Unknown health-check failure"` instead of the actual token-acquisition error (invalid client credentials, unreachable token endpoint, etc.) — exactly the class of failure the linked issue calls out wanting distinguished ("wrong bearer token" is one of its example repro causes).

  **Supporting evidence:** the pre-existing test `test_health_check_oauth_client_creds_failure` (~line 6347) wasn't updated to assert on the `error` kwarg, unlike its two siblings — a good tell this call site was simply missed rather than intentionally excluded.

  **Suggested action:** pass `error=e` at line 4741, mirroring the other two sites — or have the author explain why this path is deliberately excluded.

  ---

  ### 🟡 Worth Confirming (non-blocking)

  **`last_error` field overloading**

  The linked issue explicitly raised as an open question whether health-check failures should reuse the registration-lifecycle's `last_error` column, or get a distinct field (e.g. `last_health_error`), to avoid a different form of the same staleness problem recurring later. The PR reuses `last_error` and explains the reasoning in the description (`status_message` intentionally left alone).

  This is a defensible call given the issue left it open — but worth maintainers confirming it's the desired direction before merge, since it affects any consumer already reading `last_error` with registration-failure semantics (e.g. #5127).

feiiiiii5:
Thanks for the careful review — the missed call site is exactly right, and the supporting evidence (the sibling test not asserting on the kwarg) was a good catch.

Addressed in 01bd57047:

- `_check_single_gateway_health` now passes `error=e` on the client-credentials token failure path, mirroring the timeout call site (~4505) and the request-failure call site (~4899). The generic `"Unknown health-check failure"` no longer masks invalid client credentials or an unreachable token endpoint in `last_error`.
- `test_health_check_oauth_client_creds_failure` now asserts on the `error` kwarg (`isinstance(kwargs["error"], Exception)` and `str(...) == "Token expired"`), so a regression back to the unthreaded call fails the test.

On the non-blocking point (`last_error` field overloading): agreed it's a maintainer call — I kept the registration-lifecycle column for the reason in the description (`status_message` intentionally left untouched so human-readable status isn't clobbered by health-check noise), but if maintainers prefer a dedicated `last_health_error`, I'm happy to rework the persistence side of this PR accordingly.

marekdano (dismissed):
The PR looks good now!

LGTM 🚀

msureshkumar88 (changes requested):
Thanks for tracking down the missing exception thread here — the fix for the reported symptom (health-check failures never reaching `last_error`) is correctly scoped and the new tests for the threshold-persist and recovery-clear paths are solid.

One blocking issue found while tracing the recovery path against a disabled gateway:

**`_mark_gateway_reachable` clears `last_error` unconditionally, including for disabled gateways** (`gateway_service.py:4557`)

```python
if db_gateway:
    db_gateway.last_seen = datetime.now(timezone.utc)
    db_gateway.last_error = None   # <-- runs even when gateway_enabled is False
    update_db.commit()
```

This line sits outside the `if gateway_enabled and not gateway_reachable:` reactivation gate above it. Since `_get_gateways()` defaults to `include_inactive=True` and the periodic scheduler calls it with no override, disabled gateways are still probed. If a disabled gateway happens to answer successfully, its recorded outage reason gets silently wiped while it stays `reachable=False` — the same "operator can't tell why it's down" gap issue #6343 describes, just reintroduced for the disabled case.

Reproduced directly:
```
_mark_gateway_reachable("gw-1", "disabled-gw", gateway_enabled=False, gateway_reachable=False)
→ last_error after successful probe: None   (was: "certificate verify failed: certificate has expired")
→ reachable: unchanged (False)
```

Suggested fix — gate the clear the same way the reactivation is gated:
```python
if db_gateway:
    db_gateway.last_seen = datetime.now(timezone.utc)
    if gateway_enabled:
        db_gateway.last_error = None
    update_db.commit()
```
Would also be worth a regression test for the disabled-gateway case alongside the two new tests already in this PR.

Nothing else blocking — everything else I looked at (session/commit boundaries in the failure/recovery paths, the sanitize-params allowlist) turned out to be pre-existing patterns this PR inherits rather than regressions it introduces, so I'm not asking for those here.

feiiiiii5:
Addressed in 4a0ea6f — you were right that the unconditional `last_error = None` had reintroduced the #6343 blind spot for disabled gateways.

`_mark_gateway_reachable` now gates the clear on `gateway_enabled`, mirroring the reactivation gate: a successful probe of an enabled gateway still clears the outage reason, while a disabled gateway keeps its recorded reason even when it answers healthy.

Regression coverage:
- `test_recovery_preserves_last_error_for_disabled_gateway` — disabled + answering healthy keeps the recorded `last_error` (your exact repro shape).
- `test_recovery_clears_last_error_for_enabled_gateway` — enabled + previously unreachable still clears it.

The prior `test_recovery_clears_last_error` (which asserted the buggy disabled-path clearing) was replaced by these two. Related gateway service tests: 63 passed; ruff clean on the changed lines.

a-effort:
@msureshkumar88 the disabled-case blocker looks addressed in `4a0ea6f`. The `last_error = None` clear is now gated on `gateway_enabled` in `_mark_gateway_reachable`, matching the reactivation gate above it, and there is a regression test for the disabled path (`test_recovery_preserves_last_error_for_disabled_gateway`) next to the enabled one. CI is green on that head and a re-review was requested on 09-01.

I filed #6343, so flagging that this review is the last thing on it: the 08-31 `CHANGES_REQUESTED` is still standing and keeps the PR blocked. Could you take another look, or dismiss if you are satisfied?

ja8zyjits:
Hi @feiiiiii5 

## Issue coverage

| Requirement (from #6343) | Implemented? | Location |
|---|---|---|
| Thread exception into `_handle_gateway_failure` | ✅ | `_handle_gateway_failure(gateway, error=e)` at lines 4746, 4904, 4505 |
| Sanitize error (SecurityValidator / `sanitize_exception_message`) | ✅ | Line 4410 — uses `sanitize_exception_message(raw_error, gateway.auth_query_params)` |
| Persist to `last_error` at threshold | ✅ | Lines 4411-4414 — raw UPDATE after `set_gateway_state` |
| Clear `last_error` on recovery | ✅ | `_mark_gateway_reachable` line 4562 — `db_gateway.last_error = None` |
| `status_message` left for registration lifecycle only | ✅ | PR explicitly does not touch `status_message` |
| Separate field (`last_health_error`) considered but deferred | ✅ | PR description explains decision; acceptable for now |

---

## What the PR changes (beyond the issue fix)

The diff against `origin/main` includes a large set of unrelated removals that are **not** described in the PR:

- **Removed** `GatewayCredentialError` exception class (had distinct type for malformed stored credentials, distinguished 422 from 502).
- **Removed** `get_gateway_impact_preview()` method and associated `GatewayImpactPreview` / `GatewayImpactServer` schemas.
- **Removed** `test_server_handshake()` function and its `ServerHandshakeRequest` usage, `_HANDSHAKE_HEADER_DENYLIST`, `_handshake_form_header_allowlist()`, `_is_handshake_header_denied()`.
- **Removed** `_resolve_auth_header_name`, `internal_loopback_base_url`, `DbEmailUser`, `DbServer`, `server_service` imports from `gateway_service.py`.
- **Simplified** `fetch_tools_after_oauth()` — removed `teams` parameter and team-aware token path selection.
- **Changed** `_mark_gateway_reachable` signature — added `gateway_reachable: bool` param.
- **Changed** `test_gateway_connectivity` — removed `caller_supplies_accept` and the Accept header injection logic.

**However:** These removals are all internal to `gateway_service.py` or remove symbols not referenced anywhere else in the codebase (verified via grep). No other module imports `GatewayCredentialError`, `GatewayImpactPreview`, `test_server_handshake`, `ServerHandshakeRequest`. Callers of `fetch_tools_after_oauth` (only `mcpgateway/routers/oauth_router.py:1256`) already call without `teams=`, so the signature drop is safe.

The PR description should disclose these removals — they are not cosmetic.

---

## Findings

### 🔴 Blocking

None.

---

### 🟡 Functionally impacting

#### 1. `raw_error` fallback logic has a silent gap for empty-string exceptions

**Location:** `gateway_service.py` line 4409

```python
raw_error = (str(error) or type(error).__name__) if error is not None else "Unknown health-check failure"
```

If `error` is a real exception whose `str()` is an empty string (e.g. `ConnectionError()` with no message), `str(error)` is `""` — falsy — so `type(error).__name__` is used instead. That's intentional and correct.

**But:** `str(asyncio.TimeoutError(...))` for the timeout synthetic exception at line 4505 is the full message string, which is fine. The pattern is correct but slightly fragile — a bare `raise ConnectionError` produces `str(e) == "ConnectionError()"` on some Python builds, not `""`. No action required unless observed in production.

#### 2. `_gateway_failure_counts` reset may not run if `set_gateway_state` raises

**Location:** Lines 4411-4415

```python
with cast(Any, SessionLocal)() as db:
    await self.set_gateway_state(db, gateway.id, activate=True, reachable=False, only_update_reachable=True)
    db.execute(update(DbGateway).where(DbGateway.id == gateway.id).values(last_error=sanitized_error))
    db.commit()
    self._gateway_failure_counts[gateway.id] = 0  # Reset after deactivation
```

If `set_gateway_state` raises (e.g. `GatewayNotFoundError`, network issue during cache invalidation), the count reset is skipped. On next health cycle the counter is at threshold again and will immediately re-attempt deactivation. This is low-severity — the retry is idempotent — but could cause repeated error log spam. Previous code had the same issue; not introduced by this PR. Worth tracking as a follow-up.

---

### 🔵 Suggestions

#### 3. `fetch_tools_after_oauth` removal of `teams` parameter changes OAuth token path selection silently

The old code selected team-scoped vs shared (Admin UI) token vault paths based on the `teams` argument. The new code delegates entirely to `_resolve_auth_code_refresh_headers`, which has its own path logic. If the behavior is intentionally unified, the PR should document that `_resolve_auth_code_refresh_headers` now owns the path-selection logic. If this is accidental, it could regress team-scoped OAuth token retrieval.

**Recommendation:** Add a comment in `fetch_tools_after_oauth` noting why the team-path distinction was removed, or confirm via a test that both Admin UI and API token paths still work end-to-end after this change.

#### 4. Test `test_recovery_clears_last_error_for_enabled_gateway` only asserts `db.commit` called once

The test patches `set_gateway_state` as `AsyncMock()` (no side effects), but the real `set_gateway_state` calls `db.commit()` internally. The test's `db.commit.assert_called_once()` would fail if the mock were replaced with the real implementation. This is acceptable for a unit test but means there's no integration-level assertion that `last_error` is cleared atomically with the reachability update.

#### 5. PR description scope mismatch

The PR description says "+91 additions, -6 deletions" — the actual diff against `origin/main` is 1091 lines. The description only covers the issue fix and omits the substantial removals. This makes the PR harder to review and the change set harder to bisect if a regression is introduced by one of the removed features.

**Recommendation:** Either split the unrelated removals into a separate PR, or add a section in the PR description listing what was removed and why.

---

## Breaking change assessment for API clients

| Surface | Change | Breaking? |
|---|---|---|
| `GET /gateways/{id}` response schema | None — `last_error` already existed in `GatewayRead`, was just always null | ✅ No |
| `GatewayRead.last_error` semantics | Now populated with health-check errors, not just registration errors | ✅ Additive (was null before) |
| `GatewayRead.status_message` | Unchanged | ✅ No |
| `_handle_gateway_failure` signature | `error` param is `Optional[BaseException] = None` — all existing callers pass no arg or keyword arg | ✅ No |
| `_mark_gateway_reachable` signature | New `gateway_reachable: bool` param added — all callers in `_check_single_gateway_health` already pass it | ✅ No |
| `fetch_tools_after_oauth` — `teams` param removed | Only called from `oauth_router.py:1256` without `teams=` | ✅ No |
| `GatewayCredentialError` removed | No callers outside `gateway_service.py` | ✅ No |
| `test_server_handshake` removed | No callers outside `gateway_service.py` | ✅ No |

**Conclusion: No breaking changes to MCP clients or REST API consumers.**

---

## Test coverage

| Test | Validates |
|---|---|
| `test_threshold_persists_sanitized_health_error` | Threshold triggers `last_error` write with secret REDACTED |
| `test_recovery_clears_last_error_for_enabled_gateway` | `last_error = None` on recovery of enabled gateway |
| `test_recovery_preserves_last_error_for_disabled_gateway` | `last_error` preserved for disabled gateway probes |
| Updated `test_health_check_failure` | `error` kwarg is a `ConnectionError` with correct message |
| Updated `test_health_check_oauth_client_creds_failure` | `error` kwarg is passed through for OAuth failure path |
| Updated timeout test | `error` is `asyncio.TimeoutError` instance |

Coverage is appropriate for the change set. The sanitization test (REDACTED assertion) is especially valuable.

---

## Minor notes

- Log at line 4902 changed from raw `e` to `safe_error` (pre-sanitized): good improvement, removes a potential credential leak in debug logs.
- The `test_recovery_clears_last_error_for_enabled_gateway` test calls `_mark_gateway_reachable("gw-1", "test", True, False)` — `gateway_reachable=False` triggers the reactivation branch, so `SessionLocal` must be patched too. The test does this correctly.
- `str(error) or type(error).__name__` pattern mirrors what `_finalize_pending_gateway_failure` does for the registration path — consistency is good.

feiiiiii5:
Hi @msureshkumar88 — following up on your Aug 31 blocking finding (unconditional `last_error` clear on disabled gateways): fixed in 4a0ea6f (clear now gated on `gateway_enabled`), confirmed by @a-effort on Sep 2 and independently verified by @ja8zyjits Sep 3 review (no blocking findings). Branch just updated onto current main — the large unrelated diff they flagged was pure staleness and is gone (now +91/-6, 2 files, matching the description). Re-requesting your review; happy to split or adjust anything remaining.

msureshkumar88 (changes requested):
Nice work chasing this all the way through — the sanitize-before-persist test (`test_threshold_persists_sanitized_health_error`) that actually compiles the SQL and asserts `REDACTED`/no raw secret is a genuinely strong regression guard, and the client-credentials and disabled-gateway follow-ups both landed correctly (verified directly against the current code, not just the commit messages).

One more thing worth fixing before merge, plus a few optional cleanups.

### 🔴 Worth fixing before merge

**`_mark_gateway_reachable` clears `last_error` based on a stale snapshot, not the freshly-loaded row** (`mcpgateway/services/gateway_service.py:4729`)

```python
db_gateway = update_db.execute(select(DbGateway).where(DbGateway.id == gateway_id)).scalar_one_or_none()
if db_gateway:
    db_gateway.last_seen = datetime.now(timezone.utc)
    if gateway_enabled:          # <- stale parameter
        db_gateway.last_error = None
```

`gateway_enabled` is captured in `_check_single_gateway_health` *before* the HTTP probe runs, and the periodic health-check batch (`_get_gateways()` → `check_health_of_gateways`) processes gateways sequentially with per-gateway timeouts and a small sleep between chunks — so for a fleet of any size this value can be stale by more than a single probe's duration by the time we get here. `db_gateway` is already loaded fresh two lines above and has `.enabled` available for free.

Net effect: if a gateway is disabled while its probe is in flight, the stale `gateway_enabled=True` still wipes `last_error` on a now-disabled gateway — exactly the blind spot this PR's own comment (and the new `test_recovery_preserves_last_error_for_disabled_gateway` test) says shouldn't happen. The inverse (re-enabled mid-probe) leaves a stale error on a healthy gateway.

Suggested fix — swap the check to the row that's already loaded:
```python
if db_gateway.enabled:
    db_gateway.last_error = None
```

(Note: the pre-existing reactivation gate above at line 4715 has the same stale-snapshot shape and isn't something this PR needs to take on — flagging separately as a follow-up since it's a larger pre-existing pattern.)

### 🟡 Optional / non-blocking

1. **Timeout test lost its positional-arg check** (`tests/unit/.../test_gateway_service.py:~4006`) — `assert_awaited_once_with(mock_gateway_health)` was replaced by `assert_awaited_once()` plus a separate kwarg check. Would be nice to restore the combined assertion (`assert_awaited_once_with(mock_gateway_health, error=...)`) so a future call-site bug that reports the wrong gateway still gets caught.

2. **Two commits per threshold trip** — `_handle_gateway_failure` calls `set_gateway_state` (which commits internally) and then does a second `db.execute(update(...))` + `db.commit()` just for `last_error`. Folding the `last_error` write into `set_gateway_state`'s existing transaction would save a round trip, though this is low-severity since it only fires at threshold, not every check.

3. **Stale field description** — `GatewayRead.last_error` (`schemas.py:3963`) still says "Most recent async lifecycle error detail." Worth a one-line update since it's now also populated by health-check failures.

4. Worth a quick gut-check (not asking for a code change unless you agree): `sanitize_exception_message` only redacts known URL query-param names, not generic secret patterns (bearer tokens, `client_secret` text, etc.). Before this PR that gap only ever reached a debug log line; now it's persisted into a column exposed via `GET /gateways/{id}` to anyone with `gateways.read`. Might be worth a fast-follow to harden that helper, since this PR gives it a wider audience than before.

Otherwise this looks solid — scoped tightly to the linked issue, reuses existing columns/patterns instead of inventing new ones, and the test coverage on the sanitization path is genuinely good.

ja8zyjits:
Hi @feiiiiii5 

## 🚨 Blocking Issues

### F-1: Non-atomic `reachable` + `last_error` writes in `_handle_gateway_failure`

- **Location:** [`mcpgateway/services/gateway_service.py:4579-4583`](mcpgateway/services/gateway_service.py:4579)
- **Problem:**
  ```python
  with cast(Any, SessionLocal)() as db:
      await self.set_gateway_state(db, gateway.id, activate=True, reachable=False, only_update_reachable=True)
      db.execute(update(DbGateway).where(DbGateway.id == gateway.id).values(last_error=sanitized_error))
      db.commit()
  ```
  1. `set_gateway_state` executes `db.commit()` internally (lines 3821, 3857). Running `db.execute` + `db.commit()` immediately after on the same session violates the AGENTS.md rule against reusing committed caller sessions (prevents "transaction is inactive" / dirty session states).
  2. If `set_gateway_state` fails or process terminates between commits, gateway state is updated to `reachable=False` with `last_error=NULL` (inconsistent state).
- **Fix:** Update `set_gateway_state` signature to accept optional `last_error` kwarg and apply it within the existing single transaction / commit (line 3742 / 3821), or encapsulate into single ORM update.

---

## ⚠️ Functional & Non-Blocking Findings

### F-2: Sanitizer query param handling clarification

- **Location:** [`mcpgateway/services/gateway_service.py:4578`](mcpgateway/services/gateway_service.py:4578)
- **Detail:** `gateway.auth_query_params` stores `{param_name: encrypted_value}`. `sanitize_url_for_logging` matches query parameter keys (not secret values) against URL query strings to redact them.
- **Action:** Behavior safe in current implementation since key-based URL redaction works without decryption. Add code comment clarifying key-only redaction or pass `None` if decrypted value matching is not expected.

### F-3: `str(error) or type(error).__name__` fallback path

- **Location:** [`mcpgateway/services/gateway_service.py:4577`](mcpgateway/services/gateway_service.py:4577)
- **Detail:** Timeout path constructs descriptive `asyncio.TimeoutError(f"health check timed out after {timeout}s")`. Empty `str(error)` fallback to class name (`type(error).__name__`) works as safety net.
- **Action:** No code changes required.

---

## 🧪 Test Quality & Coverage Gaps

### Missing Coverage

1. **`error=None` threshold path**: Cover branch where `error is None` at threshold trip, asserting fallback `"Unknown health-check failure"` in `last_error`.
2. **Empty exception message**: Cover exception with empty `str(e)` to exercise `type(error).__name__` fallback.
3. **Mock assertion ambiguity**: In `test_recovery_clears_last_error_for_enabled_gateway`, isolate mock DB session between `SessionLocal` and `fresh_db_session` context managers to prevent ambiguous `assert_called_once` assertions.

feiiiiii5:
Addressed in 776bf4c — msureshkumar88 Sep 7 finding (stale snapshot in _mark_gateway_reachable).

BEFORE (mcpgateway/services/gateway_service.py):
```python
db_gateway = update_db.execute(select(DbGateway).where(DbGateway.id == gateway_id)).scalar_one_or_none()
if db_gateway:
    db_gateway.last_seen = datetime.now(timezone.utc)
    if gateway_enabled:          # stale param captured before the probe
        db_gateway.last_error = None
```

AFTER:
```python
db_gateway = update_db.execute(select(DbGateway).where(DbGateway.id == gateway_id)).scalar_one_or_none()
if db_gateway:
    db_gateway.last_seen = datetime.now(timezone.utc)
    # NOTE: read enabled from the freshly-loaded row, not the
    # gateway_enabled snapshot captured before the probe, which
    # can be stale if the gateway was disabled mid-probe.
    if db_gateway.enabled:
        db_gateway.last_error = None
```

The reactivation gate above (`if gateway_enabled and not gateway_reachable`) is intentionally left as-is per the review's own follow-up note. Regression coverage: existing enabled/disabled tests now set the fresh row's `enabled` explicitly, plus two stale-snapshot tests (stale True + row disabled preserves; stale False + row enabled clears). Related suites on this head: TestHandleGatewayFailureThreshold + TestMarkGatewayReachableErrorCleanup + TestSetGatewayState + TestCheckSingleGatewayHealth — 29 passed; 401-reactivation + oauth-health — 12 passed; ruff check/format clean on gateway_service.py.

feiiiiii5:
Addressed in 776bf4c — ja8zyjits Sep 7 F-1 (non-atomic reachable + last_error writes in _handle_gateway_failure).

BEFORE (mcpgateway/services/gateway_service.py):
```python
with cast(Any, SessionLocal)() as db:
    await self.set_gateway_state(db, gateway.id, activate=True, reachable=False, only_update_reachable=True)
    db.execute(update(DbGateway).where(DbGateway.id == gateway.id).values(last_error=sanitized_error))
    db.commit()
```
(set_gateway_state commits internally, so the second execute+commit reused an already-committed session and the two writes were not atomic.)

AFTER:
```python
with cast(Any, SessionLocal)() as db:
    await self.set_gateway_state(db, gateway.id, activate=True, reachable=False, only_update_reachable=True, last_error=sanitized_error)
```
with `set_gateway_state` gaining an optional `last_error` kwarg applied to the already-loaded ORM object before its single commit (plus a last_error-only branch when state is unchanged, so the reason is never silently dropped). No second execute/commit on the caller session; verified no UPDATE on the gateways table outside the single transaction path.

Regression coverage: threshold test now asserts the single-call contract (last_error kwarg sanitized with REDACTED/no secret, outer db.execute/commit not called) and fails pre-fix; two new set_gateway_state tests cover atomic persist and last_error-only persist. Same suite results as above (29 + 12 passed, ruff clean).

msureshkumar88 (dismissed):
Reviewed the full branch (all 5 commits) against the linked issue #6343. This is a well-scoped, minimal fix — threads the health-check exception into `_handle_gateway_failure`, sanitizes it with the existing `sanitize_exception_message` helper (same one `_finalize_pending_gateway_failure` already uses for the pending-registration path), and clears `last_error` on recovery while correctly preserving it for disabled-but-still-probed gateways. No schema/migration change needed since `last_error` already existed on the model. Backend-only, no UI touch, not a breaking change — verified every caller of `set_gateway_state`/`_handle_gateway_failure` and the new params are all optional/appended, nothing shifts. Ran the targeted test subset plus the full file (all green) and ruff clean.

One thing worth a follow-up rather than blocking this PR: `_mark_gateway_reachable` clears `last_error` in a second, separate transaction after the reachability flip commits (verified this at runtime — there's a real, if narrow, window where `GET /gateways/{id}` can show `reachable: true` with the stale `last_error` still attached). Fixing it cleanly needs a sentinel/flag change to `set_gateway_state` (its current `last_error: Optional[str] = None` param can't express "clear," since `None` means "leave unchanged" — confirmed this can't be worked around with the existing signature), so it's more than a one-liner and reasonable to track separately.

A few smaller nice-to-haves for a follow-up, none blocking:
- Fold `last_error` into `set_gateway_state`'s audit-trail (`new_values`) and structured-log (`custom_fields`) payloads so the recorded reason shows up in the audit trail, not just the read API.
- A regression test for `_mark_gateway_reachable`'s `except` path (currently uncovered) now that it also governs `last_error` correctness.
- `docs/docs/manage/api-usage.md` and the `GatewayRead.last_error` field description still frame `last_error` as pending/async-lifecycle-only — worth a doc update now that health checks populate it too.
- Extending the existing `last_error` e2e pattern in `tests/e2e/test_gateway_async_lifecycle.py` (currently registration-path only) to cover the health-check path.

Approving — the core fix is solid and directly addresses the issue. Happy to open the follow-up issue for the transaction-atomicity item if that's useful.

ja8zyjits:
Hi @feiiiiii5 
---

## 🚨 Blocking Issues

### B-1 · `sanitize_exception_message` receives encrypted `auth_query_params`, not decrypted values

**File:** [`mcpgateway/services/gateway_service.py:4589`](mcpgateway/services/gateway_service.py:4589)  
**Severity:** Blocking — **security medium**

```python
sanitized_error = sanitize_exception_message(raw_error, getattr(gateway, "auth_query_params", None))
```

`DbGateway.auth_query_params` stores **encrypted** values. `sanitize_exception_message` matches these as literal strings against the URL in the exception message — encrypted ciphertext will never match plaintext query param values from `httpx`. The sanitizer silently returns the unsanitized message, so plaintext API keys/tokens in `httpx` URLs may be written to `last_error` in the DB and exposed via `GET /gateways/{id}`.

Same bug exists at line 5099 (logging path), but that result is only logged, not persisted.

**Fix (Option 1 — preferred):** Add `auth_query_params_decrypted: Optional[Dict[str, str]] = None` to `_handle_gateway_failure` and pass the already-decrypted dict (`auth_query_params_decrypted`, line 4780) from `_check_single_gateway_health` call sites (lines 5101, ~4934). Use that instead of `getattr(gateway, "auth_query_params", None)` inside `_handle_gateway_failure`.

**Test gap:** `test_threshold_persists_sanitized_health_error` uses `{"api_key": "secret"}` for both stored params and URL value — in production these differ (stored = ciphertext, URL = plaintext). A realistic test must use `{"api_key": "<encrypted_blob>"}` as stored and `"secret"` in the URL to expose the mismatch.

---

### B-2 · `str(error) or type(error).__name__` logic smell — whitespace-only strings not handled

**File:** [`mcpgateway/services/gateway_service.py:4588`](mcpgateway/services/gateway_service.py:4588)  
**Severity:** Blocking — logic

```python
raw_error = (str(error) or type(error).__name__) if error is not None else "Unknown health-check failure"
```

`str(error) or` only falls back on empty string. Whitespace-only strings (e.g. `"   "`) are truthy, so a third-party exception with `__str__` returning `"   "` would persist whitespace as `last_error`. Also inconsistent: the timeout path (line 4682) constructs a message that never includes the URL, while the non-timeout path passes the raw exception — leaky abstraction.

**Fix:**
```python
raw_error = str(error).strip() or type(error).__name__ if error is not None else "Unknown health-check failure"
```

---

## ⚠️ Minor / Suggestions

### Architecture

**A-1 · `elif` last_error-only branch has no audit trail** *(Suggestion)*  
[`gateway_service.py:3932`](mcpgateway/services/gateway_service.py:3932)

```python
elif last_error is not None and gateway.last_error != last_error:
    gateway.last_error = last_error
    gateway.updated_at = datetime.now(timezone.utc)
    db.commit()
    db.refresh(gateway)
```

Primary state-change branch calls `audit_trail.log_action()` (correctly, without `db=`). This `elif` path — when only `last_error` changes — does not. Low risk given current call sites (the early-return guard at line 4578 prevents reaching here when already unreachable), but a future caller would silently skip audit. Add `audit_trail.log_action(...)` (no `db=`) mirroring lines 3897–3913.

---

**A-2 · Double sanitization of `exc_to_inspect`** *(Minor)*  
[`gateway_service.py:5099-5101`](mcpgateway/services/gateway_service.py:5099)

Line 5099 sanitizes `exc_to_inspect` for the debug log; line 5101 passes the raw exception to `_handle_gateway_failure` which sanitizes again. Both use the same `auth_query_params` source. No credential leak, but redundant. Noted for future cleanup.

---

**A-4 · Audit `new_values` missing `last_error` field** *(Minor)*  
[`gateway_service.py:3905`](mcpgateway/services/gateway_service.py:3905)

```python
new_values={
    "enabled": gateway.enabled,
    "reachable": gateway.reachable,
    # last_error written here but not audited
},
```

`last_error` is committed in the same transaction as state changes but absent from the audit record. Suggestion: add `"last_error": gateway.last_error` to `new_values` when `last_error is not None`.

---

### Code Docs

**F-1 · `set_gateway_state` docstring — "atomically" claim inaccurate** *(Suggestion)*  
[`gateway_service.py:3707`](mcpgateway/services/gateway_service.py:3707)

Current: `last_error: Optional sanitized failure reason persisted atomically with the state change.`

"Atomically" only holds on the primary branch. The `elif` branch commits `last_error` separately. Suggested:
```
last_error: Optional sanitized failure reason to persist with the state change.
            Written in the same commit when the state changes; written in a
            separate commit when only last_error differs (no state transition).
```

---

**F-2 · `set_gateway_state` docstring — `Raises` typo: "agent" → "gateway"** *(Minor)*  
[`gateway_service.py:3715`](mcpgateway/services/gateway_service.py:3715)

```
PermissionError: If user doesn't own the agent.  # ← should be "gateway"
```

---

**F-3 · `_handle_gateway_failure` doctest doesn't show new `error` param** *(Suggestion)*  
[`gateway_service.py:4552`](mcpgateway/services/gateway_service.py:4552)

Existing doctest shows pre-PR zero-argument call. No example passes `error=`. Suggested addition (illustrative, not contract):
```python
>>> service.set_gateway_state = unittest.mock.AsyncMock()
>>> asyncio.run(service._handle_gateway_failure(gateway, error=RuntimeError("refused")))
>>> service.set_gateway_state.called
True
```

---

**F-5 · Test name `test_state_change_persists_last_error_atomically` — "atomically" mismatch** *(Minor)*  
[`test_gateway_service.py:6139`](tests/unit/mcpgateway/services/test_gateway_service.py:6139)

Name implies general guarantee. The `elif` branch violates it. If F-1 docstring is fixed, rename to:
```python
test_state_change_persists_last_error_with_reachability_change
```

---

### Test Quality

**TQ-1 · "No raw UPDATE" assertion uses string-match on SQL** *(Minor)*  
[`test_gateway_service.py:6055`](tests/unit/mcpgateway/services/test_gateway_service.py:6055)

```python
for call in db.execute.call_args_list:
    assert "UPDATE gateways" not in str(call.args[0])
```

Brittle: locale/SQLAlchemy-version-sensitive. ORM mutations never call `db.execute` at all. Cleaner:
```python
db.commit.assert_called_once()
```

---

**TQ-2 · `_make_gateway` base dict lacks `last_error: None` default** *(Minor)*  
[`test_gateway_service.py:90`](tests/unit/mcpgateway/services/test_gateway_service.py:90)

New tests pass `last_error=None` explicitly. A future caller forgetting the keyword will get `AttributeError` instead of a clear assertion failure. Add `"last_error": None` to the base dict.

---

**TQ-4 · `test_recovery_clears_last_error_for_enabled_gateway` — slight over-patching of `SessionLocal`** *(Minor)*  
[`test_gateway_service.py:6452`](tests/unit/mcpgateway/services/test_gateway_service.py:6452)

The outer `SessionLocal` mock and the `set_gateway_state` mock share the same `db` object, which could mask assertion failures if the call sequence changes. The critical assertion (`recovered.last_error is None`) is still correct. Not blocking.

---

**TQ-5 · No test for `error=None` → `"Unknown health-check failure"` fallback** *(Suggestion)*  
[`gateway_service.py:4588`](mcpgateway/services/gateway_service.py:4588)

Production call sites always pass an error, but the fallback is live code. Suggested:
```python
@pytest.mark.asyncio
async def test_threshold_uses_fallback_when_no_error(self, gateway_service, monkeypatch):
    gw = SimpleNamespace(id="gw-none", name="test", enabled=True, reachable=True, auth_query_params=None)
    gateway_service._gateway_failure_counts = {}
    monkeypatch.setattr("mcpgateway.services.gateway_service.GW_FAILURE_THRESHOLD", 1)
    gateway_service.set_gateway_state = AsyncMock()
    db = MagicMock()
    db.__enter__ = MagicMock(return_value=db)
    db.__exit__ = MagicMock(return_value=False)
    monkeypatch.setattr("mcpgateway.services.gateway_service.SessionLocal", MagicMock(return_value=db))
    await gateway_service._handle_gateway_failure(gw)  # no error arg
    _, kwargs = gateway_service.set_gateway_state.await_args
    assert kwargs["last_error"] == "Unknown health-check failure"
```

---

**TQ-6 · No test for `set_gateway_state` with `last_error=""` edge case** *(Suggestion)*  
[`gateway_service.py:3747`](mcpgateway/services/gateway_service.py:3747)

Guard is `is not None`, so `last_error=""` writes empty string. Production always produces non-empty from `sanitize_exception_message` but the public API accepts any string. Low risk; worth a comment or test.

---

**TQ-7 · `kwargs["error"]` assertion raises `KeyError` if call site goes positional** *(Minor)*  
[`test_gateway_service.py:4007`](tests/unit/mcpgateway/services/test_gateway_service.py:4007)

```python
error = gateway_service._handle_gateway_failure.await_args.kwargs["error"]
```

Correct today since source uses `error=` keyword. If ever changed to positional, fails with `KeyError` not `AssertionError`. Acceptable pattern; consistent with lines 6613 and 6825.

---

## Coverage Summary

| New behavior | Covered? | Notes |
|---|---|---|
| `set_gateway_state` persists `last_error` on state change | ✅ | `test_state_change_persists_last_error_atomically` |
| `set_gateway_state` `elif` branch: `last_error` only | ✅ | `test_last_error_only_persists_without_state_change` |
| `_handle_gateway_failure` sanitizes + passes error | ✅ | `test_threshold_persists_sanitized_health_error` |
| Auth query params redacted from `last_error` | ✅ (partial) | Test mock uses same value for stored+URL — see B-1 |
| `_mark_gateway_reachable` clears `last_error` (enabled) | ✅ | `test_recovery_clears_last_error_for_enabled_gateway` |
| `_mark_gateway_reachable` preserves `last_error` (disabled) | ✅ | `test_recovery_preserves_last_error_for_disabled_gateway` |
| Mid-probe disable race | ✅ | Both mid-probe tests |
| Timeout path passes `asyncio.TimeoutError` | ✅ | `test_check_health_timeout` (updated) |
| Non-OAuth failure passes exception | ✅ | `test_health_check_failure` (updated) |
| OAuth failure passes exception | ✅ | `test_health_check_oauth_client_creds_failure` (updated) |
| `error=None` fallback → `"Unknown health-check failure"` | ❌ | TQ-5: missing |
| `last_error` cleared on `status_message`-only gateway | n/a | PR explicitly leaves `status_message` unchanged |

---

## Master Finding Table

| ID | Severity | Domain | Location | Finding |
|----|----------|--------|----------|---------|
| B-1 | 🚨 Blocking (security) | Code Quality | `gateway_service.py:4589` | Encrypted `auth_query_params` passed to sanitizer — plaintext creds may persist in DB |
| B-2 | 🚨 Blocking (logic) | Code Quality | `gateway_service.py:4588` | `str(error) or` doesn't strip whitespace-only strings |
| A-1 | Suggestion | Architecture | `gateway_service.py:3932` | `elif` last_error-only branch has no audit trail |
| A-2 | Minor | Architecture | `gateway_service.py:5099` | Double sanitization of `exc_to_inspect` — harmless |
| A-4 | Minor | Architecture | `gateway_service.py:3905` | Audit `new_values` missing `last_error` field |
| F-1 | Suggestion | Code Docs | `gateway_service.py:3707` | "atomically" claim inaccurate for no-state-change path |
| F-2 | Minor | Code Docs | `gateway_service.py:3715` | Typo: "agent" → "gateway" in `Raises` |
| F-3 | Suggestion | Code Docs | `gateway_service.py:4552` | Doctest doesn't show new `error` param |
| F-5 | Minor | Code Docs | `test_gateway_service.py:6139` | Test name includes "atomically" — inconsistent with F-1 |
| TQ-1 | Minor | Test Quality | `test_gateway_service.py:6055` | SQL string-match assertion fragile |
| TQ-2 | Minor | Test Quality | `test_gateway_service.py:90` | `_make_gateway` base missing `last_error: None` |
| TQ-4 | Minor | Test Quality | `test_gateway_service.py:6452` | Slight over-patching of `SessionLocal` |
| TQ-5 | Suggestion | Test Quality | `gateway_service.py:4588` | No test for `error=None` fallback string |
| TQ-6 | Suggestion | Test Quality | `gateway_service.py:3747` | No test for `last_error=""` edge case |
| TQ-7 | Minor | Test Quality | `test_gateway_service.py:4007` | `kwargs["error"]` → `KeyError` if call goes positional |

feiiiiii5:
Addressed in 5f99906 (on top of 363f81a) — thank you for the precise pointers.

**B-1** — Done via Option 1: `_handle_gateway_failure` takes a new optional `auth_query_params` (falls back to the stored column when absent), and both in-scope call sites (the OAuth-token-failure path and the genuine-connectivity path) now pass `auth_query_params_decrypted`. The 5099 logging path prefers decrypted with stored as fallback. The timeout path is intentionally untouched: no decrypted dict is in scope there, and its message carries no URL, so nothing secret can reach the sanitizer.

BEFORE: `sanitize_exception_message(raw_error, getattr(gateway, "auth_query_params", None))` — reached back to the encrypted column.
AFTER: `sanitize_exception_message(raw_error, auth_query_params or getattr(gateway, "auth_query_params", None))` with the decrypted dict threaded from the call sites.

One honest note from verifying the mechanism: redaction here is name-based (`sanitize_url_for_logging` redacts by param name, never compares values), and the encrypted/decrypted dicts share identical keys — so I could not construct a case where stored-ciphertext redacts worse. I added `test_threshold_prefers_decrypted_params_over_ciphertext` to lock that invariant both ways (explicit-decrypted and stored-only paths), rather than claim a leak I couldn't reproduce. If you know a path where keys diverge, point me at it and I'll chase it.

**B-2** — Done: `raw_error = (str(error).strip() or type(error).__name__) if ...`. Locked by `test_whitespace_error_falls_back_to_type_name` (whitespace-only `str` now persists the type name; verified this test fails on the pre-fix expression).

**A-1** — Done: the `last_error`-only `elif` in `set_gateway_state` now mirrors the primary branch's `audit_trail.log_action(...)` (no `db=`, `new_values` carries just `last_error`).

Full test run left to CI.

feiiiiii5:
@ja8zyjits 

Update: squashed the two commits into 52b2aaf (same tree) so every commit carries Signed-off-by for DCO. Content unchanged from what is described above.

ja8zyjits:
Hi @feiiiiii5 

Please fix these 2 blocking issues and we are good to merge it.

---

## 🚨 Blocking

### B-1 · `_gateway_failure_counts` reset skipped on `set_gateway_state` exception

**File:** [`mcpgateway/services/gateway_service.py:4611-4613`](mcpgateway/services/gateway_service.py:4611)

```python
with cast(Any, SessionLocal)() as db:
    await self.set_gateway_state(db, gateway.id, activate=True, reachable=False,
                                  only_update_reachable=True, last_error=sanitized_error)
    self._gateway_failure_counts[gateway.id] = 0  # Reset after deactivation
```

If `set_gateway_state` raises (`GatewayNotFoundError`, DB error), reset is skipped. Next cycle: count already ≥ `GW_FAILURE_THRESHOLD` → retries `set_gateway_state`, writes another audit entry, loops indefinitely until gateway deleted or process restarted.

**Fix:** Move reset before the `with`:

```python
if count >= GW_FAILURE_THRESHOLD:
    ...
    self._gateway_failure_counts[gateway.id] = 0  # Reset before DB call
    with cast(Any, SessionLocal)() as db:
        await self.set_gateway_state(...)
```

---

### B-2 · `set_gateway_state` docstring claims `last_error` persisted "atomically with state change" — false for `elif` path

**File:** [`mcpgateway/services/gateway_service.py:3707`](mcpgateway/services/gateway_service.py:3707)

Docstring says: `last_error: Optional sanitized failure reason persisted atomically with the state change.`

Actual `elif` path (lines 3932–3952) fires a **separate `db.commit()`** with no state change — `last_error` is the only change.

**Fix:**
```
last_error: Optional sanitized failure reason to persist. Written
    atomically with the state change when reachability changes; written
    in a standalone commit when only the error text differs (no state
    transition occurred).
```

---

feiiiiii5:
Addressed in 59cd887 — both blocking items, thank you for the precise pointers (and for the merge signal). BEFORE/AFTER: (1) the failure-count reset moved above the `with SessionLocal()` block so a `set_gateway_state` raise can no longer loop deactivation + audit entries; verified the recount path still trips the threshold on the next genuine cycle. (2) `last_error` docstring now states both persistence modes (atomic with state change vs standalone commit on the error-only path). Ruff check + format clean on the touched file; full CI left to the runners.

feiiiiii5:
@ja8zyjits gentle nudge: both blocking items from your 09-11 review are addressed in 59cd887 (failure-count reset moved above the `SessionLocal` block; recount path verified), pushed 09-12. When you have a moment, could you take another look? If the two items check out, this should be ready to merge per your note. Thanks!

ja8zyjits (dismissed):
LGTM

ja8zyjits:
@feiiiiii5 Thanks for fixing the concerns.
Approved.

Iam fixing the pre-commit, dont push anyone else.

msureshkumar88 (dismissed):

msureshkumar88 (approved):

PR #6369, opened 2026-08-22
fix(mcp): preserve structured-only federated tool results

### LINKED ISSUE #5472: Federated tool results with structuredContent + empty content are dropped (Output validation error for every outputSchema tool)
## Summary

When a **federated** upstream MCP server returns a *successful* `tools/call` result whose data is only in `structuredContent` with an **empty `content` array**, ContextForge discards the structured payload and returns `Output validation error: outputSchema defined but no structured output returned` (`isError: true`) for **every tool that declares an `outputSchema`**. The real data never reaches the client.

Returning structured content with empty `content` is spec-valid — [the spec](https://modelcontextprotocol.io/specification/2025-06-18/server/tools#structured-content) makes unstructured `content` a SHOULD, not a MUST, when `structuredContent` is present. (Real-world trigger: Brex's hosted MCP server returns exactly this shape, which breaks all of its `outputSchema` tools when federated through ContextForge.)

## Minimal reproduction (no Docker, no cloud, no secrets)

👉 **https://github.com/charliemeyer2000/contextforge-structured-content-repro**

```bash
git clone https://github.com/charliemeyer2000/contextforge-structured-content-repro
cd contextforge-structured-content-repro
./run.sh
```

It installs stock `mcp-contextforge-gateway==1.0.4` (SQLite), starts it alongside a ~90-line stdlib-only MCP server that returns `{content: [], structuredContent: {...}}` for a tool declaring an `outputSchema`, federates it, and calls the tool through `/mcp` **before and after** a one-line fix:

```
================== BEFORE fix (stock v1.0.4) ==================
  isError    : True
  content    : ['Output validation error: outputSchema defined but no structured output returned']
  structured : None
================== AFTER fix ==================
  isError    : False
  content    : []
  structured : {'id': 'widget-42', 'name': 'Reproduction Widget', 'price': 9.99}
```

## Root cause

In `mcpgateway/transports/streamablehttp_transport.py`, `call_tool()`:

```python
result = await tool_service.invoke_tool(...)   # correctly captures result.structured_content
if not result or not result.content:           # <-- fires for structured-only results
    logger.warning("No content returned by tool: %s", name)
    return []                                   # <-- discards result.structured_content
# ... ~130 lines below, the handler that WOULD have forwarded it:
#     if structured:
#         return (unstructured, structured)
```

`invoke_tool` captures the upstream `structuredContent` correctly (`tool_service.py`, MCP branch: `ToolResult(content=..., structured_content=structured, ...)`). But the `if not result.content:` guard returns `[]` **before** the structured-content handler runs, so the payload is thrown away. The bundled MCP SDK's server-side validator (`mcp==1.27.2`, `mcp/server/lowlevel/server.py`: `if tool.outputSchema is not None and maybe_structured_content is None`) then substitutes the `Output validation error` string.

## Suggested fix (one line)

Bail only when there is **neither** unstructured content **nor** structured content:

```diff
- if not result or not result.content:
+ if not result or (not result.content and not getattr(result, "structured_content", None)):
```

Structured-only results then fall through to the existing handler, which returns `(unstructured, structured)`; the SDK validates the structured payload against the `outputSchema` (it passes) and the client receives the data. Verified by the repro above. Happy to open a PR if useful.

## Environment

- `mcp-contextforge-gateway==1.0.4` (latest release; the same guard is present on `main`)
- bundled `mcp==1.27.2`
- reproduced with SQLite + a single worker; independent of auth/RBAC/transport config

feiiiiii5:
# 🐛 Bug-fix PR

## 📌 Summary

A successful federated `tools/call` result may carry all of its data in `structuredContent` while `content` is empty. The MCP Tools specification makes unstructured content a **SHOULD**, not a **MUST**, when structured content is present. ContextForge's streamable-HTTP transport treated that valid shape as an empty result and returned before reaching the structured-content handler.

For tools declaring an `outputSchema`, the MCP SDK then substituted:

```text
Output validation error: outputSchema defined but no structured output returned
```

The real structured payload never reached the client.

Closes #5472

## 🔁 Reproduction Steps

1. Federate an upstream MCP tool that declares an `outputSchema`.
2. Return `{ "content": [], "structuredContent": { ... } }` from its `tools/call` result.
3. Invoke the tool through ContextForge's `/mcp` streamable-HTTP endpoint.
4. Current main returns the validation-error text instead of forwarding the structured payload.

The issue includes a standalone reproduction repository and before/after output.

## 🐞 Root Cause

After `invoke_tool()` returned successfully, the early guard was:

```python
if not result or not result.content:
    return []
```

It ignored `result.structured_content`. Structured-only results therefore returned `[]` before the later handler could assemble `(unstructured, structured)` for SDK validation.

## 💡 Fix Description

Treat the result as empty only when both representations are absent:

```python
if not result or (not result.content and not getattr(result, "structured_content", None)):
```

Structured-only results now fall through to the existing handler and return `(unstructured, structured)`. Results with neither representation continue to return `[]`.

This is intentionally separate from #6182/#6274: those track preservation of an empty dict (`structuredContent: {}`) in the service layer. This fix handles non-empty structured payloads whose unstructured `content` array is empty.

## 📏 Reviewability

## 🧪 Verification

Pre-fix: the new structured-only test failed with `result == []`.

Post-fix on this branch:

```bash
python -m pytest -q tests/unit/mcpgateway/transports/test_streamablehttp_transport.py -k 'call_tool_success or with_structured_content or structured_only_result or neither_representation or preserves_is_error_for_egress'
# 5 passed
```

Additional checks:

```bash
uvx ruff@0.16.1 check mcpgateway/transports/streamablehttp_transport.py
# All checks passed
```

Full project lint/test/coverage were not duplicated locally due to the large dependency set; CI will provide that result.

| Changed source lint                   | Ruff 0.16.1          | Pass |
| Targeted regression                   | 5 tests              | Pass |
| Pre-fix regression                    | Current-main behavior | Fail as expected |
| Full lint / test / coverage           | `make lint` / CI     | Pending CI |

- [x] Matches current MCP spec — unstructured `content` is a SHOULD when `structuredContent` is present
- [x] Existing genuinely-empty results still return `[]`

## ✅ Checklist

- [x] Code formatted

### DIFF
--- a/mcpgateway/transports/streamablehttp_transport.py
+++ b/mcpgateway/transports/streamablehttp_transport.py
@@ -1929,7 +1929,7 @@ async def call_tool(
                 meta_data=meta_data,
                 require_model_visible=True,
             )
-            if not result or not result.content:
+            if not result or (not result.content and not getattr(result, "structured_content", None)):
                 logger.warning("No content returned by tool: %s", name)
                 return []
 
--- a/tests/unit/mcpgateway/transports/test_streamablehttp_transport.py
+++ b/tests/unit/mcpgateway/transports/test_streamablehttp_transport.py
@@ -566,6 +566,63 @@ async def test_call_tool_with_structured_content(monkeypatch):
 
 
 @pytest.mark.asyncio
+async def test_call_tool_preserves_structured_only_result(monkeypatch):
+    """A spec-valid structured-only result must not be discarded as empty.
+
+    MCP makes unstructured ``content`` a SHOULD—not a MUST—when
+    ``structuredContent`` is present. The early empty-content guard therefore
+    has to consider both representations.
+    """
+    # First-Party
+    from mcpgateway.transports.streamablehttp_transport import call_tool, tool_service
+
+    mock_db = MagicMock()
+    mock_result = MagicMock()
+    mock_result.content = []
+    mock_structured = {"id": "widget-42", "name": "Reproduction Widget", "price": 9.99}
+    mock_result.structured_content = mock_structured
+    mock_result.model_dump = lambda by_alias=True: {"content": [], "structuredContent": mock_structured}
+
+    @asynccontextmanager
+    async def fake_get_db():
+        yield mock_db
+
+    monkeypatch.setattr("mcpgateway.transports.streamablehttp_transport.get_db", fake_get_db)
+    monkeypatch.setattr(tool_service, "invoke_tool", AsyncMock(return_value=mock_result))
+
+    result = await call_tool("widget", {})
+
+    assert isinstance(result, tuple)
+    unstructured, structured = result
+    assert unstructured == []
+    assert structured == mock_structured
+
+
+@pytest.mark.asyncio
+async def test_call_tool_returns_empty_when_neither_representation_exists(monkeypatch):
+    """The early return remains for results with neither content representation."""
+    # First-Party
+    from mcpgateway.transports.streamablehttp_transport import call_tool, tool_service
+
+    mock_db = MagicMock()
+    mock_result = MagicMock()
+    mock_result.content = []
+    mock_result.structured_content = None
+    mock_result.model_dump = lambda by_alias=True: {"content": [], "structuredContent": None}
+
+    @asynccontextmanager
+    async def fake_get_db():
+        yield mock_db
+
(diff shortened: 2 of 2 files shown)


feiiiiii5:
Closing as part of a contribution-portfolio consolidation: my effort is now focused on a single project (failroute, a static analyzer for silent-failure anti-patterns) and its upstream follow-through. Happy to revisit if there is interest.

Issue #6374, opened 2026-08-24
Fix Python dependency update discovery and execution commands in `docs/docs/development/release-management.md` Section 2.2 and Upgrade validation update


prakhar-singh1928:
### 🔧 Chore Summary

Fix Python dependency update discovery and execution commands in `docs/docs/development/release-management.md` Section 2.2.

---

### 🧱 Area Affected

- [ ] GitHub Actions / CI Pipelines
- [ ] Pre-commit hooks / linters
- [ ] Formatting (black, isort, ruff, etc.)
- [ ] Type-checking (mypy, pyright, pytype, etc.)
- [x] Dependency cleanup or updates
- [ ] Build system or `Makefile`
- [ ] Containerization (Docker/Podman)
- [x] Docs or spellcheck
- [ ] SBOM, CVE scans, licenses, or security checks
- [ ] Other:

---

### ⚙️ Context / Rationale

While running Section 2.2 of the release management guide (`docs/docs/development/release-management.md`), several issues were identified in the automated discovery snippet:

1. **TOML parse errors on unrendered templates**: The `find` command did not exclude all virtual environment directories (e.g., `.venv.sbom`, `.venv-dev`) or cookiecutter template directories, causing `uv` to fail parsing Jinja syntax like `{{ cookiecutter.plugin_slug }}` inside package templates.
2. **Invalid command syntax**: The snippet referenced `uv-upsync`, which is not a valid `uv` command; this should be `uv sync`.
3. **Environment independence**: The requirements updater commands used `python` instead of `uv run python`, leading to `command not found` or python interpreter version mismatches.
4. **Pattern matching for requirements files**: Changed `-name "requirements.txt"` to `-name "requirements*.txt"` to catch any auxiliary requirements files.

---

### 📦 Related Make Targets

- `make install-dev` - verify dev environment installation
- `make check-env` - verify environment settings

---

### 📋 Acceptance Criteria

- [ ] `docs/docs/development/release-management.md` Section 2.2 discovery commands updated with `-name ".venv*" -prune` and `-name "templates" -prune`
- [ ] `uv-upsync` replaced with `uv sync` in the documentation snippet
- [ ] Script executions updated to use `uv run python`
- [ ] `requirements*.txt` pattern used for comprehensive dependency discovery

---

### 🧩 Additional Context

Tested against the repository tree — correctly discovers all 10 active `pyproject.toml` files and 5 `requirements*.txt` files without breaking on templates or site-packages.

PR #6160, opened 2026-08-10
chore: routine npm/rust dependency updates

prakhar-singh1928:
# Pull Request

## 🔗 Related Issue

Closes #

---

## 📝 Summary

Routine npm dependency updates — bumps all frontend packages to their latest compatible versions within the existing semver ranges defined in `package.json`. No range boundaries were crossed; only `package-lock.json` and `.npmrc` are changed alongside the rebuilt Admin UI bundle.

Key bumps:
- `@biomejs/biome` 2.5.5 → 2.5.6
- `rollup` 4.62.2 → 4.62.3
- `postcss` 8.5.23 → 8.5.25
- `dompurify` 3.4.12 → 4.4.13
- `nanoid` 3.3.16 → 3.3.18
- `js-yaml` 4.3.0 → 4.3.1
- `acorn` 8.17.0 → 8.18.0
- Various transitive deps (`@napi-rs/wasm-runtime`, `enhanced-resolve`, `flatted`, `magicast`, `meriyah`, `minimatch`, `tinyrainbow`, `tldts`, etc.)

`.npmrc` updated to exclude `dompurify`, `js-yaml`, and `nanoid` from the `min-release-age` check, as these packages publish frequent small releases that would otherwise be held back unnecessarily.

Admin UI bundle rebuilt with Vite after the lockfile update. Rust/Cargo crates were reviewed — no updates required.

---

## 📏 Reviewability

- [x] This PR has one clear purpose
- [ ] The linked issue is not labeled `triage`
- [x] Unrelated bugs or improvements are tracked in separate issues/PRs
- [x] Tests are included with the code they validate
- [x] If AI-assisted, I understand and can explain the generated changes

---

## 🏷️ Type of Change

- [ ] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [x] Chore (deps, CI, tooling)
- [ ] Other (describe below)

---

## 🧪 Verification

| Check | Command | Status |
|---|---|---|
| npm audit | `npm audit` | ✅ 0 vulnerabilities |
| Web lint | `make lint-web` | ✅ 0 errors |
| JS unit tests | `make test-js-coverage` | ✅ All pass |
| Admin UI bundle rebuild | `make build-ui` | ✅ New bundle generated |
| Production container build | `make docker-prod DOCKER_BUILD_ARGS="--no-cache"` | ✅ Image built successfully |

---

## ✅ Checklist

- [x] Code formatted (`make black isort pre-commit`)
[...]

PR #6158, opened 2026-08-10
chore: update python dependencies/update min supported version to py3.12

prakhar-singh1928:
# Pull Request

## 🔗 Related Issue

Closes #

---

## 📝 Summary

Routine dependency upgrade pass (2026-08-09):

- Bumped all `[tool.uv.exclude-newer-package]` cutoff dates to `2026-08-09T23:59:59Z` so the rolling `exclude-newer = "10 days"` window can resolve recently-published packages that are direct or transitive requirements
- Added `cryptography = "2026-08-09T23:59:59Z"` override — `cryptography>=50.0.0` (published `2026-07-31`) fell outside the 10-day window; also added the same override to `plugins/external/opa` and `plugins/external/cedar` which pull the gateway as a transitive dep via git
- Bumped minimum supported Python from `>=3.11` to `>=3.12` across the root `pyproject.toml` and all 8 MCP server `pyproject.toml` files — the scientific stack (`numpy>=2.5.1`, `pandas>=3.0.5`, `scipy>=1.18.0`, `scikit-learn>=1.9.0`, `matplotlib>=3.11.1`) had already dropped Python 3.11; the `requires-python = ">=3.11"` declaration was incorrect. All Containerfiles and CI workflows were already targeting Python 3.12; this change makes the metadata consistent with reality
- Updated `tool.black`, `tool.mypy`, `tool.ruff`, and `tool.pytype` target versions from `py311`/`3.11` to `py312`/`3.12` across all affected files
- Updated `.devcontainer/Dockerfile` base image from `python:3.11-slim` to `python:3.12-slim`
- Updated `docs/requirements.txt`, `tests/load/requirements.txt`, and `mcp-servers/python/data_analysis_server/requirements.txt` via the `update_dependencies.py` tool

---

## 📏 Reviewability

- [x] This PR has one clear purpose
- [ ] The linked issue is not labeled `triage`
- [x] Unrelated bugs or improvements are tracked in separate issues/PRs
- [x] Tests are included with the code they validate
- [x] If AI-assisted, I understand and can explain the generated changes

---

## 🏷️ Type of Change

- [ ] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [x] Chore (deps, CI, tooling)
- [ ] Other (describe below)

---

## 🧪 Verification

| Check                     | Command                  | Status |
|---------------------------|--------------------------|--------|
| Dependency resolution     | `uv lock --upgrade`      |     ✅   |
[...]

PR #6375, opened 2026-08-24
chore(deps): perform routine python/node dependency updates

prakhar-singh1928:
# Pull Request

## 🔗 Related Issue

Closes #

---

## 📝 Summary

Performs routine Python dependency updates across the repository:
- Updates CPEX packages in `[tool.uv.exclude-newer-package]` to the current date (`2026-08-24`).
- Removes temporary exclude-newer package overrides (`cryptography`, `greenlet`, `h2`) from `pyproject.toml`.
- Aligns `requires-python = ">=3.12,<3.14"` and classifier metadata across external plugins (`plugins/external/opa`, `plugins/external/cedar`, and `plugins/external/llmguard`) following the Python 3.11 deprecation.
- Upgrades lockfiles across all `pyproject.toml` projects using `uv lock --upgrade --exclude-newer "10 days"`.
- Updates `requirements*.txt` files across the repository via `.github/tools/update_dependencies.py`.
- updates package.json - node packages

---

## 📏 Reviewability

---

## 🏷️ Type of Change

- [ ] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

| Check                     | Command                                        | Status |
|---------------------------|------------------------------------------------|--------|
| Dependency Resolution     | `make install-dev`                             | Passed |
| Security Audit            | `make pip-audit`                               | Passed |
| Lint suite                | `make lint`                                    | Passed |
| Unit tests                | `make test`                                    | Passed |
| Coverage ≥ 80%            | `make coverage`                                | Passed |

---

## ✅ Checklist

---

## 📓 Notes (optional)

- All external plugin sub-projects (`opa`, `cedar`, `llmguard`) now match the gateway's requirement of Python `>=3.12,<3.14`.


cafalchio (dismissed):
LGTM

msureshkumar88 (changes requested):
Thanks for running this cycle — the diff is clean and disciplined. Ten files, all genuinely in scope, no `.secrets.baseline` churn, no stray formatting, and no unrelated edits smuggled in. CI is fully green (pytest py3.12, all ten cpex plugin E2E jobs in both static and binding modes, playwright smoke, four-arch container builds, dependency-review). Removing the `cryptography` / `greenlet` / `h2` overrides is exactly what `main` asked for — the marker comment above them literally said "remove before next upgrade cycle starts."

I verified the security posture is intact rather than assuming it: `exclude-newer-package` is a soak ceiling, not a floor, so dropping those entries can't weaken anything. Confirmed empirically — `cryptography>=50.0.0` still resolves to 50.0.0, `h2>=4.4.1` (CVE-2026-71554) still resolves to 4.4.1, and only `greenlet` moved at all, by a patch. Also confirmed no Alembic migration is warranted here; the `alembic` bump changes the runner, not the schema.

A few things I'd like to see addressed before merge. Nothing here is a correctness bug — it's mostly consistency that this PR is the natural place to close.

---

## Changes requested

**1. Root `pyproject.toml:46` still advertises Python 3.11**

```toml
"Programming Language :: Python :: 3.11",   # ← line 46
...
requires-python = ">=3.12,<3.14"            # ← line 55
```

The PR removes this exact classifier from `cedar`, `llmguard`, and `opa`, but leaves it on the root project — which is the one actually published to PyPI as `mcp-contextforge-gateway`, so it's the most visible instance of the contradiction. One line to delete, same class of change as the plugin edits.

**2. The `cryptography` override survives in `cedar` and `opa`**

`plugins/external/cedar/pyproject.toml:62-65` and `plugins/external/opa/pyproject.toml:62-65` still carry:

```toml
[tool.uv.exclude-newer-package]
# cryptography>=50.0.0 is a transitive requirement via mcp-contextforge-gateway.
# 50.0.0 published 2026-07-31T14:23Z falls outside the rolling 10-day exclude-newer window.
cryptography = "2026-08-09T23:59:59Z"
```

Same temporary class as the root entries this PR removes, added by the same PR (#6158), for the same reason. Root proves the hold is no longer needed — removing it there left the lock resolving 50.0.0 unchanged. Left in place, these will keep aging and quietly hold `cryptography` back for anyone building those plugins.

**3. Dependency floors weren't advanced, though the lock moved 59 packages**

A sample of the drift this bakes in:

| Package | Floor in `pyproject.toml` | Resolved in `uv.lock` |
|---|---|---|
| `starlette` | `>=1.3.1` | **1.6.0** |
| `alembic` | `>=1.18.5` | 1.19.1 |
| `pydantic-settings` | `>=2.14.2` | 2.15.0 |
| `sqlalchemy` | `>=2.0.51` | 2.0.52 |
| `sse-starlette` | `>=3.4.6` | 3.4.8 |
| `uvicorn[standard]` | `>=0.51.0` | 0.52.3 |

Since the lock isn't published, a plain `pip install mcp-contextforge-gateway` resolves against the floors — so a user can legitimately land on `starlette 1.3.1`, a version no CI job exercises once this merges. #5919 bumped floors in one pass with inline rationale comments, so there's a clear precedent.

Either bump them the way #5919 did, or add a line to the PR body saying floors are intentionally frozen this cycle. Right now the silence reads as an oversight rather than a decision, which is the only reason I'm flagging it.

**4. Rebase**

Branch is one commit behind `main` (`48e12d32e`).

---

## Suggestions (non-blocking)

**Keep the convention marker.** The diff removes `# additional packages below this: remove before next upgrade cycle starts` along with the three packages under it. Removing the entries is right; losing the marker means the next cycle has no signpost for where temporaries belong or reminder to clear them. Worth keeping as an empty section header.

**A couple of PR-body corrections** — cosmetic, but it's what the next person reads:

- "updates package.json - node packages" — `package.json` is actually unchanged. Only `package-lock.json` moved: 98 transitive bumps, **zero major**, root `dependencies`/`devDependencies` byte-identical, plus one new optional platform binary (`@napi-rs/lzma-linux-x64-gnu`). Worth noting I checked that last one for host bias and it's clean — it's declared upstream by `rollup@4.62.4`, and all 25 rollup platform variants are present, so the lock stays cross-platform.
- "Removes temporary exclude-newer package overrides (`cryptography`, `greenlet`, `h2`)" — true for root, but see item 2.
- Commit `5d777a303` has a placeholder trailer: `Signed-off-by: Your Name prakhar.singh1928@ibm.com`. The second, well-formed line satisfies DCO (check is green), so purely cosmetic.

**Worth linking two open issues as Related** (not `Closes`) for traceability — leaving `Closes #` blank matches what #5919, #6158, and #6160 all did, so no concern there:

- **#6374** — the `release-management.md` §2.2 fixes you filed from this exact run. Linking makes the deliberate split legible.
- **#3904** — this diff has fresh evidence for it. `.github/tools/update_dependencies.py:359` queries PyPI `/json` and writes the newest version straight through; there's no minimum-release-age or yank filter anywhere in that script, so `requirements*.txt` aren't gated by the `exclude-newer = "10 days"` soak that gates `uv.lock`. Visible right here: `docs/requirements.txt:11` now pins `charset-normalizer>=3.5.1` while the soak-gated `uv.lock:620` resolves `3.5.0`. Not this PR's job to fix, but good ammunition for that issue.

**On the cpex pins** — moving them to `2026-08-24T23:59:59Z` (the PR's own date) means those eight packages run at zero soak versus ten days for everything else. That matches the convention #6158 established, and it was a no-op this cycle: all eight cpex versions are byte-identical before and after, and every plugin E2E job is green. So no action needed. But since #6158 stripped the per-package rationale comments, the block is now eight bare dates that get re-stamped by hand each cycle — one comment explaining *why* cpex is exempt would stop the zero-soak posture reading as an accident to anyone auditing the supply chain later.

**Highest-risk bump to keep an eye on** is `starlette 1.3.1 → 1.6.0` (three minors, directly on the request hot path — middleware stack, streaming responses, `TestClient`), together with `sse-starlette` and `uvicorn`. The green suite plus playwright smoke covers it reasonably; if the SSE / streamable-HTTP paths were exercised, saying so in the body would close the loop nicely.

---

## On the breaking change

The plugin `requires-python` narrowing (`>=3.11` → `>=3.12`) is technically breaking, but I traced the blast radius and it's effectively zero:

- `cedar` and `opa` already depend on `mcp-contextforge-gateway>=1.0.8`, which is `requires-python >=3.12` — so 3.11 was already unresolvable. Pure metadata correction.
- `llmguard` is the only nominal case, since it declares no gateway dependency. But `llmguardplugin/plugin.py`, `cache.py`, and `llmguard.py` all `import mcpgateway`, so 3.11 was already broken at runtime.
- All three `Containerfile`s already set `ARG PYTHON_VERSION=3.12`, and no workflow references 3.11 at all.

So this is a correction, not a drop — good change. Two things that would make it fully airtight, both fine as follow-ups: a `CHANGELOG.md` entry (#5919 and #6160 both added one, and it's the cheapest way to make a metadata change to three published packages discoverable), and a separate issue for `llmguard`'s undeclared `mcp-contextforge-gateway` dependency — that's a latent packaging bug independent of this PR, and it's the only reason 3.11 was ever nominally satisfiable there.

## Follow-ups worth filing (definitely not this PR)

- `tox.ini:24` is `envlist = py311, py312, py313` with `tox.ini:35` mapping `3.11: py311`, while root `requires-python` is `>=3.12` — `make tox` can't install the package under py311. Stale from #6158, but it does sit inside this PR's "3.11 deprecation" theme if you'd rather fold the two-line fix in here.
- `.github/workflows/pytest.yml:73` is `python: ["3.12"]` while `requires-python` allows 3.13 and the classifiers advertise it — 3.13 is claimed-supported and never tested. A dependency refresh is exactly when a 3.13-only resolution difference would show up.
- A pre-commit hook asserting `classifiers` matches `requires-python` across every tracked `pyproject.toml` would have caught item 1 automatically, and would prevent the same drift next time a floor moves.
- No `uv lock --check` job anywhere in CI (`--locked` only appears for Cargo), so a `pyproject.toml` edit without a matching relock isn't machine-verified — and the three plugin `pyproject.toml` files have no tracked lock at all.
- The body reports `make pip-audit` passed, but there's no `pip-audit` or `osv-scanner` job in CI; `dependency-review` only inspects the declared-dependency diff, not the full resolved tree. Wiring it in would turn that template line into evidence.
- Roughly two dozen docs still say Python ≥3.11 (`docs/docs/development/index.md:24`, `building.md:9`, `architecture/index.md:43,178`, `plugins.md:445`, ADR-005, and others). Pre-existing from #6158 and much too large to belong in a dependency refresh, but it contradicts this PR's own rationale, so worth tracking — possibly folded into the #6374 docs pass.

Happy to re-review as soon as items 1–4 are in.

prakhar-singh1928:
Thanks for the detailed review! @msureshkumar88 

- Removed the stray Python 3.11 classifier from root pyproject.toml.
- Removed the stale cryptography overrides from plugins/external/cedar and plugins/external/opa, and preserved the convention marker comment in root pyproject.toml.
- Dependency floors: We deliberately chose to freeze the floors in pyproject.toml during this routine cycle to minimize churn for downstream consumers, letting uv.lock handle the resolution bumps within existing ranges.
- Branch has been rebased against main.

msureshkumar88 (changes requested):
Nice work tightening this up in the last commit — the stale 3.11 classifier, the dead `cryptography` overrides in `cedar`/`opa`, and restoring the `# additional packages below this` marker all landed cleanly, and the diff stays scoped to the 10 dependency files with 47/47 CI green.

Two follow-ups worth folding into this PR before merge:

**1. Dependency floors haven't moved despite the lock advancing 59 packages.** The gap that stands out most is `starlette>=1.3.1` (`pyproject.toml`) vs. `starlette 1.6.0` actually resolved in `uv.lock` — three minors apart, and starlette sits directly on the request hot path (middleware stack, streaming/SSE, `TestClient`). A plain `pip install` against these floors can currently land on `starlette 1.3.1`, a version nothing in CI has exercised since this merges.

Same pattern on a few others:
- `alembic>=1.18.5` → resolved `1.19.1`
- `pydantic-settings>=2.14.2` → resolved `2.15.0`
- `sqlalchemy>=2.0.51` → resolved `2.0.52`
- `sse-starlette>=3.4.6` → resolved `3.4.8`
- `uvicorn[standard]>=0.51.0` → resolved `0.52.3`

Worth bumping these floors to match what the lock actually resolves to (this is what #5919 did for its routine cycle).

**2. `tox.ini` still has `py311` in the matrix** (`envlist` on line 24, and the `gh-actions` map on line 35), even though root `requires-python` is now `>=3.12` after this PR's classifier cleanup. Since this PR is squarely about finishing the Python 3.11 deprecation, dropping `py311` from both spots here would close the loop rather than leaving it for a future cycle to notice.

Everything else from the earlier pass (CHANGELOG entry, PR body wording, rebase cadence) I'm not blocking on — those are optional/cosmetic and can go either way.

msureshkumar88:
Following up on my review above — I went back and verified the floor finding properly, and my original list was both incomplete and slightly miscategorized. Correcting it here so you're not chasing the wrong set.

### The finding holds, and it's better evidenced than I first put it

Two things I checked that make this more than a style preference:

- **On `main`, 5 of the 6 floors I named matched the resolved lock version exactly** (`alembic` 1.18.5, `pydantic-settings` 2.14.2, `sqlalchemy` 2.0.51, `sse-starlette` 3.4.6, `starlette` 1.3.1). So floor-in-lockstep-with-lock looks like a deliberate, maintained invariant here, not an accident — and #5919 confirms it, having bumped `fastapi`, `jq`, `prometheus-*`, `sse-starlette`, `starlette-compress`, `typer`, `uvicorn`, `langsmith`, and the `opentelemetry-*` set in exactly this way.
- **On `starlette` specifically, nothing else is protecting that floor.** `fastapi` 0.141.1 declares only `starlette>=0.46.0`, well below our `>=1.3.1`, so our own declaration is the binding constraint. A fresh `pip install mcp-contextforge-gateway` really can resolve to `starlette 1.3.1` while CI only ever exercises 1.6.0. That's the one I'd most want closed.

### Corrections to my list

**I missed two** that this PR also knocked out of lockstep:

- `Mako>=1.3.12` → lock resolves **1.4.1** (a minor bump — actually a larger jump than several I did list)
- `typer>=0.27.0` → lock resolves **0.27.1**

**I miscategorized `uvicorn`.** Floor is `>=0.51.0`, but `main` already resolved `0.52.0` — so that drift predates this PR (it came in with #5919) rather than being introduced here. Still worth picking up, but it isn't something this PR broke.

### Corrected sets

Broken by this PR — these were exact-lockstep on `main`:

| Package | Floor | Resolved |
|---|---|---|
| `starlette` | 1.3.1 | **1.6.0** |
| `Mako` | 1.3.12 | 1.4.1 |
| `alembic` | 1.18.5 | 1.19.1 |
| `pydantic-settings` | 2.14.2 | 2.15.0 |
| `sse-starlette` | 3.4.6 | 3.4.8 |
| `sqlalchemy` | 2.0.51 | 2.0.52 |
| `typer` | 0.27.0 | 0.27.1 |

Pre-existing drift, already stale on `main` (optional to fold in, entirely reasonable to leave): `cpex` 0.1.2→0.1.3, `fastapi` 0.140.0→0.141.1, `filelock` 3.32.0→3.32.3, `mcp` 1.28.1→1.29.0, `redis` 8.0.1→8.1.0, `uvicorn` 0.51.0→0.52.3.

The seven in the table are the ones I'd suggest addressing to restore the invariant. Apologies for the churn on the first pass.

msureshkumar88 (approved):
Ran a real-environment E2E pass on the dependency set this PR actually produces (HEAD `c0ed10e13`), focused on the highest-risk change — the `starlette 1.3.1 → 1.6.0` jump — since that sits directly on the request/middleware/SSE hot path. Everything checked out; approving.

### Config
- Python 3.12.3, `uv` 0.10.2, sqlite (default `DATABASE_URL`), app-native `.env` loading (no manual env overrides — `pydantic-settings` reads `.env` itself)
- Dependency install matched `make install-dev`'s actual extras: dev group + `plugins` extra (not `--all-extras` — this sandbox lacks `pg_config` for the optional `psycopg[c]` Postgres extra; sqlite path, which is what the app booted and was tested on, is unaffected)

### Commands run
```
uv sync --group dev --extra plugins
.venv/bin/python3 -c "import importlib.metadata as m; print(m.version('<pkg>'))"   # per bumped package
.venv/bin/uvicorn mcpgateway.main:app --host 127.0.0.1 --port 8124   (background)
curl http://127.0.0.1:8124/health | / | /docs | /sse
.venv/bin/python -m pytest tests/unit/mcpgateway/transports/test_sse_transport.py \
    test_streamablehttp_compression_contentlength.py test_websocket_transport.py -v
npm ci --no-audit --no-fund
uv pip install --dry-run -e plugins/external/{cedar,opa,llmguard}
```

### Results

| Check | Result |
|---|---|
| `uv sync` (dev + plugins) | ✅ 60 packages resolved, exit 0 |
| 7 bumped floors match `uv.lock` | ✅ `alembic` 1.19.1, `Mako` 1.4.1, `pydantic-settings` 2.15.0, `sqlalchemy` 2.0.52, `sse-starlette` 3.4.8, **`starlette` 1.6.0**, `typer` 0.27.1 |
| Server boot on the new lock | ✅ clean startup, all routers mounted, no tracebacks. `/health` → 200 healthy; `/docs`, `/sse` → 401 (auth middleware correctly enforcing on the new starlette, as expected with `AUTH_REQUIRED` default) |
| Transport-layer tests (SSE / streamable-HTTP / WebSocket) | ✅ **63 passed, 0 failed**, 0.63s |
| `npm ci` against the updated `package-lock.json` | ✅ 667 packages, 0 vulnerabilities |
| `tox.ini` | ✅ `envlist = py312, py313`, no `py311` anywhere |
| `cedar` / `opa` / `llmguard` dry-run install under `requires-python >=3.12,<3.14` | ✅ all three resolve cleanly on 3.12 |

**Not covered** (environment/scope limits, not findings): a full MCP client↔server SSE message round-trip (substituted with the transport unit tests above, which do exercise the SSE/streaming machinery directly), the Postgres-backed path (no `pg_config` in this sandbox), and the full 18-minute suite (ran a targeted subset instead).

No regression, traceback, or behavior change traced to any of the 7 bumped packages — auth/middleware/routing all correct on the new starlette. Nothing further required from my side.

PR #6393, opened 2026-08-24
fix: enforce component-aware directory confinement on admin log download

msureshkumar88:
The admin log-download handler (`GET /admin/logs/file`) confined the requested file to `LOG_FOLDER` using `str(file_path).startswith(str(log_dir))`. A string prefix comparison has no notion of a path-component boundary, so a sibling directory that merely shares a textual prefix with the log directory — `/srv/logs_secret` against a `/srv/logs` root — passes the check while living entirely outside the intended tree.

## Changes

- Adds `is_path_within()` to `mcpgateway/utils/paths.py` as the canonical directory-confinement check, and uses it in `admin_get_log_file`. Absolute paths, drive/UNC anchors, NUL bytes and `..` components are now rejected before the path join as defence-in-depth; the post-resolve `is_path_within` check is the actual control.
- Moves the 403 raise out of the surrounding `try` block. A bare `except Exception` was catching the `HTTPException` and rewriting every denial into a 400, leaving the 403 branch unreachable.
- Two other call sites carried the same prefix-comparison pattern and now use the shared helper: `ValidationMiddleware.validate_resource_path` and `SecurityValidator.validate_path`. Both apply only when `allowed_roots` is configured, and the change is deny-only.

Happy to split the two validator call sites into their own PR if reviewers would rather keep this one scoped to `admin.py`.

## Tests

New regression coverage in `tests/unit/mcpgateway/test_admin.py`, `tests/unit/mcpgateway/utils/test_paths.py`, `tests/unit/mcpgateway/middleware/test_validation_middleware.py` and `tests/unit/mcpgateway/validation/test_validators_advanced.py`:

- sibling-prefix directory (`logs` vs `logs_secret`) — the case a `startswith` check gets wrong
- symlink inside the log directory pointing outside it, which contains no `..` and so exercises the post-resolve check rather than the input filter
- absolute paths and NUL bytes
- resolve failure surfaces as a rejection rather than falling through
- nested legitimate log files still download, so confinement is not over-tight
- deny-path tests for both validator call sites with an absolute sibling-prefix root

Each new deny-path test was confirmed to fail against the unpatched code before the fix was applied.

### DIFF
--- a/mcpgateway/admin.py
+++ b/mcpgateway/admin.py
@@ -209,7 +209,6 @@ from mcpgateway.utils.pagination import paginate_query
 from mcpgateway.utils.passthrough_headers import PassthroughHeadersError
 from mcpgateway.utils.paths import is_path_within, open_confined
 from mcpgateway.utils.paths import resolve_root_path as _resolve_root_path
-from mcpgateway.utils.paths import UnsupportedPlatformError
 from mcpgateway.utils.security_cookies import clear_auth_cookie, CookieTooLargeError, set_auth_cookie
 from mcpgateway.utils.services_auth import encode_auth
 from mcpgateway.utils.sqlalchemy_modifier import json_contains_tag_expr
@@ -15544,21 +15543,21 @@ async def admin_get_log_file(
         if not (file_path.suffix in [".log", ".jsonl", ".json"] or file_path.stem.startswith(Path(settings.log_file).stem)):
             raise HTTPException(403, "Not a log file")
 
-        # Open the verified fd directly via an openat()/O_NOFOLLOW chain instead of
-        # handing a pathname to FileResponse, which reopens the path when it streams.
-        # A process able to write into log_dir_resolved could otherwise rename the
-        # checked file away and put a symlink in its place between the confinement
+        # Open the verified fd directly via a component-by-component confined open
+        # instead of handing a pathname to FileResponse, which reopens the path when it
+        # streams. A process able to write into log_dir_resolved could otherwise rename
+        # the checked file away and put a symlink in its place between the confinement
         # check above and that later reopen (TOCTOU), causing this privileged endpoint
-        # to stream an attacker-chosen target. open_confined() resolves and opens each
-        # path component with O_NOFOLLOW relative to its already-open parent, so the fd
-        # it returns refers to exactly the inode that was validated.
+        # to stream an attacker-chosen target. On POSIX, open_confined() resolves and
+        # opens each path component with O_NOFOLLOW relative to its already-open
+        # parent, so the fd it returns refers to exactly the inode that was validated.
+        # On platforms without that atomic chaining (Windows), it falls back to a
+        # per-component symlink/reparse-point rejection that is not atomic but still
+        # rejects every reparse point present at check time.
         try:
             file_fd, file_stat = open_confined(log_dir_resolved, candidate)
         except FileNotFoundError:
             raise HTTPException(404, f"Log file not found: {filename}")
-        except UnsupportedPlatformError as e:
-            LOGGER.error("Log file download unavailable on this platform: %s", sanitize_for_log(e))
-            raise HTTPException(501, "Log file download is not supported on this platform")
         except (OSError, ValueError) as e:
             LOGGER.warning("Log file access denied for %s: %s", sanitize_for_log(filename), sanitize_for_log(e))
             raise HTTPException(403, _ACCESS_DENIED_MSG)
--- a/mcpgateway/admin.py
+++ b/mcpgateway/admin.py
@@ -201,6 +201,7 @@ from mcpgateway.services.token_catalog_service import TokenCatalogService
 from mcpgateway.services.tool_service import ToolError, ToolLockConflictError, ToolNameConflictError, ToolNotFoundError, ToolService
 from mcpgateway.utils.create_jwt_token import create_jwt_token, get_jwt_token
 from mcpgateway.utils.error_formatter import ErrorFormatter, sanitize_validation_error_for_log
+from mcpgateway.utils.log_sanitizer import sanitize_for_log
 from mcpgateway.utils.metadata_capture import MetadataCapture
 from mcpgateway.utils.oauth_resource import parse_oauth_resource_form
 from mcpgateway.utils.orjson_response import ORJSONResponse
@@ -208,6 +209,7 @@ from mcpgateway.utils.pagination import paginate_query
 from mcpgateway.utils.passthrough_headers import PassthroughHeadersError
 from mcpgateway.utils.paths import is_path_within, open_confined
 from mcpgateway.utils.paths import resolve_root_path as _resolve_root_path
+from mcpgateway.utils.paths import UnsupportedPlatformError
 from mcpgateway.utils.security_cookies import clear_auth_cookie, CookieTooLargeError, set_auth_cookie
 from mcpgateway.utils.services_auth import encode_auth
 from mcpgateway.utils.sqlalchemy_modifier import json_contains_tag_expr
@@ -15554,11 +15556,14 @@ async def admin_get_log_file(
             file_fd, file_stat = open_confined(log_dir_resolved, candidate)
         except FileNotFoundError:
             raise HTTPException(404, f"Log file not found: {filename}")
+        except UnsupportedPlatformError as e:
+            LOGGER.error("Log file download unavailable on this platform: %s", sanitize_for_log(e))
+            raise HTTPException(501, "Log file download is not supported on this platform")
         except (OSError, ValueError) as e:
-            LOGGER.warning(f"Log file access denied for {filename}: {e}")
+            LOGGER.warning("Log file access denied for %s: %s", sanitize_for_log(filename), sanitize_for_log(e))
             raise HTTPException(403, _ACCESS_DENIED_MSG)
         except Exception as e:
-            LOGGER.error(f"Error opening log file for download: {e}")
+            LOGGER.error("Error opening log file for download: %s", sanitize_for_log(e))
             raise HTTPException(500, f"Error reading file for download: {e}")
 
         LOGGER.info(f"Serving log file download: {file_path.name} ({file_stat.st_size} bytes)")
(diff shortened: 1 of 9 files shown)


mcpgateway/admin.py
vishu-bh: Nice static path-confinement hardening. Could we keep verified file descriptor open instead of passing pathname to `FileResponse`? Starlette reopens this path during streaming. Process able to rename entries in `LOG_FOLDER` can swap checked file for symlink after confinement check, causing privileged download to serve arbitrary readable target. 

Please use directory-FD / openat with no-follow semantics and stream verified FD.

vishu-bh (changes requested):

mcpgateway/admin.py
msureshkumar88: Good catch, thanks — you're right that `FileResponse` reopening the path by name left a TOCTOU window between the confinement check and the actual stream.

Fixed by adding `open_confined()` to `mcpgateway/utils/paths.py`: it walks the path via `os.open(..., dir_fd=...)` with `O_NOFOLLOW` on every component (including the final one), so the fd it returns refers to exactly the inode that was validated — nothing can be swapped out from under it afterward. `admin_get_log_file` now streams that fd directly via `StreamingResponse` instead of handing a pathname to `FileResponse`.

Added regression coverage for: symlink at the final component (rejected even when its target is inside the log dir), symlinked intermediate directory component, missing file, absolute/`..` input, and directory-as-leaf.

vishu-bh (changes requested):
Thanks for closing pathname race with per-component `O_NOFOLLOW` opens. This resolves prior concern on POSIX.

A few follow-ups remain before merge:

**Blocking — Windows compatibility**

Project documents Windows support, but `open_confined()` now uses Unix-only `os.O_DIRECTORY` and `dir_fd`. On Windows this raises `AttributeError`, then log-download handler returns 500. Could we add platform-safe secure implementation, or explicitly gate endpoint with clear supported-platform behavior?

**P2 — descriptor lifecycle**

Handler opens file descriptor before creating `StreamingResponse`, but only closes it inside generator. If response is cancelled before first iterator advance, descriptor remains open. Could we make FD ownership cancellation-safe, including disconnect-before-stream?

**P2 — HTTP download regression**

Replacing `FileResponse` drops `Range`, `Accept-Ranges`, `ETag`, and `Last-Modified` handling. This breaks resumable large-log downloads. Could we preserve equivalent response behavior?

msureshkumar88:
Thanks for the thorough follow-up — addressed all three:

**Windows compatibility (blocking).** `open_confined()` now checks for `dir_fd`/`O_DIRECTORY`/`O_NOFOLLOW` support at import time and dispatches accordingly: the existing `openat`/`O_NOFOLLOW` chain stays on POSIX, and a new per-component fallback (symlink-checked via `os.path.islink()` before descending/opening) covers platforms without that support. The fallback's TOCTOU guarantee is narrower than the POSIX path — documented in the docstring — but it no longer crashes with `AttributeError` and turns into a 500.

**FD lifecycle (P2).** The fd is now wrapped in a file object (`os.fdopen`) immediately after `open_confined()` returns, owned outside the generator, and closed via a `BackgroundTask` in addition to the generator's own `finally`. `handle.close()` is idempotent, so if Starlette cancels the response before the body generator is ever iterated (immediate disconnect), the background task still closes the fd.

**HTTP download regression (P2).** The handler now sets `Accept-Ranges`, `ETag`, and `Last-Modified` from the verified fd's stat result, and serves single-range and suffix-range `Range` requests as 206 partial responses (with `If-Range` validation and 416 for out-of-bounds ranges) — matching `FileResponse`'s behavior for the resumable-download case. Multi-range (`multipart/byteranges`) isn't implemented since it's not needed for resume; happy to add if you'd rather have full parity.

Added regression tests for all three (Windows-fallback dispatch/symlink-rejection, fd-closed-via-background-task-on-cancel, single/suffix range + malformed/unsatisfiable range + stale If-Range).

Rebased onto main and regenerated `.secrets.baseline`.

mcpgateway/utils/paths.py
```diff
@@ -130,3 +138,139 @@ def resolve_root_path(request: Request, *, fallback: str | None = None) -> str:
...
+    rename-and-symlink race between the last check and the ``open()`` call is still
+    possible, since the checks and the open are not atomic with each other.
+    """
+    current = root
+    for part in parts[:-1]:
+        current = current / part
+        if os.path.islink(current) or not current.is_dir():
+            raise OSError(f"refusing to descend through non-directory or symlink: {current}")
+    final = current / parts[-1]
+    if os.path.islink(final):
```
vishu-bh: Could we fail closed on platforms without atomic no-follow traversal, or use a native Windows handle API that rejects reparse points? This checks islink() then opens by pathname; process able to write LOG_FOLDER can swap file to symlink/reparse point between steps, letting admin download stream arbitrary service-readable data. Current docstring already notes this race.

mcpgateway/admin.py
```diff
@@ -15496,44 +15501,134 @@ async def admin_get_log_file(
...
+            "Last-Modified": last_modified,
+            "ETag": etag,
+        }
+
+        range_header = request.headers.get("range")
+        if_range = request.headers.get("if-range")
+        start, end = 0, file_size
+        status_code = 200
+        if range_header and (if_range is None or if_range in (etag, last_modified)):
+            range_match = re.fullmatch(r"bytes=(\d*)-(\d*)", range_header.strip())
```
vishu-bh: FileResponse supported multipart byte ranges, while this parser accepts only one start-end pair. Requests such as Range: bytes=0-1,4-5 now return 400 instead of 206 multipart response. Could we preserve prior behavior, or avoid adding range handling until equivalent support exists?

mcpgateway/admin.py
```diff
@@ -15496,44 +15501,134 @@ async def admin_get_log_file(
...
+            if not range_match or not (range_match.group(1) or range_match.group(2)):
+                handle.close()
+                raise HTTPException(400, "Malformed Range header")
+            range_start, range_end = range_match.group(1), range_match.group(2)
+            if range_start:
+                start = int(range_start)
```
vishu-bh: int(range_start) / int(range_end) can raise ValueError for oversized numeric headers after file handle opens. That escapes handler as 500; previous FileResponse treats malformed ranges as 400. Please catch conversion errors, close handle deterministically, return malformed-range response, and add regression test.

mcpgateway/admin.py
```diff
@@ -15496,44 +15501,134 @@ async def admin_get_log_file(
...
-                media_type="application/octet-stream",
-                filename=file_path.name,
-                stat_result=file_stat,
-            )
+            file_fd, file_stat = open_confined(log_dir_resolved, candidate)
         except FileNotFoundError:
-            LOGGER.error(f"Log file disappeared before streaming: {filename}")
             raise HTTPException(404, f"Log file not found: {filename}")
+        except (OSError, ValueError) as e:
+            LOGGER.warning(f"Log file access denied for {filename}: {e}")
```
vishu-bh: Could we sanitize filename and exception text before logging? Filename originates in query parameter; CR/LF can forge log entries when open fails. sanitize_for_log() and parameterized logging match existing project patterns.

vishu-bh (changes requested):
overall this is a positive direction, few minor to things to check

msureshkumar88:
Rebased onto main and addressed the four follow-ups from the latest review:

**Windows fallback still racy.** Agreed — `_open_confined_fallback()`'s `islink()`-then-`open()` wasn't atomic, so it didn't actually deliver the TOCTOU guarantee this function exists for. Went with your first option: `open_confined()` now fails closed on any platform without `dir_fd`/`O_NOFOLLOW` support, raising a new `UnsupportedPlatformError` instead of using the racy fallback. `admin_get_log_file` catches it distinctly and returns 501 rather than letting it fall through to the generic 403/500 branches. The per-component fallback opener is gone.

**Multi-range regression.** `bytes=0-1,4-5` now falls through to a full 200 response (per RFC 7233 §3.1, a server may ignore a Range request it can't satisfy) instead of erroring with 400. Multipart/byteranges still isn't implemented — happy to add it if you'd rather have full parity — but it no longer regresses to an error for a header the old `FileResponse` used to handle.

**Oversized range value → 500.** You were right that `int()` can raise here — not from the digit-only regex producing something unparseable, but from Python's `sys.get_int_max_str_digits()` guard (default 4300 digits) tripping on a very long digit string. Wrapped the conversion in `try/except ValueError`, closes the handle and returns 400.

**Log injection via filename/exception text.** Switched the warning/error logging in the `open_confined()` exception handlers to parameterized logging with `sanitize_for_log()`, matching the pattern used elsewhere in the codebase (e.g. `oauth_router.py`).

Added regression coverage for all four: platform-unsupported → 501, multi-range → full 200, oversized range digit string → 400, and a caplog-based test asserting no CR/LF reaches the log record on the access-denied path.

Rebased onto main and regenerated `.secrets.baseline`.

vishu-bh:
Thanks for addressing the earlier review feedback. The Windows path now fails closed, the pathname TOCTOU issue is resolved on supported POSIX platforms, oversized and multi-range requests are handled, and denied-access logging is sanitized.

Two functionally-impacting issues remain:

**Blocking — FIFO can block the async worker**

In `mcpgateway/utils/paths.py:258`, the final component is opened with `O_RDONLY | O_NOFOLLOW` before `fstat()` verifies it is a regular file. Opening a FIFO read-only blocks until a writer connects. Because `open_confined()` is called synchronously from the async admin handler, a FIFO such as `blocked.log` can block the worker/event loop indefinitely.

I reproduced this locally: the current open remained blocked, while adding `O_NONBLOCK` returned immediately and allowed the existing `fstat()` check to reject the FIFO. Could we open the final component with:

```python
os.O_RDONLY | os.O_NOFOLLOW | os.O_NONBLOCK
```

Regular-file reads are unaffected, while special files can be rejected without blocking.

**Functionally impacting — Range unit comparison is case-sensitive**

`mcpgateway/admin.py:15594` requires exact lowercase `bytes=`. HTTP range-unit names are case-insensitive, and the previous Starlette `FileResponse` parser accepted values such as `Range: Bytes=0-1` and `Range: BYTES=0-1`. These now return 400.

Could we split the unit from the range value and compare using `unit.strip().lower() == "bytes"`?

**Suggestion — sanitization test does not exercise control characters**

`test_admin_get_log_file_access_denied_log_is_sanitized` uses a normal filename and exception, so it would still pass if sanitization were removed. Worth using a filename or mocked exception containing CR/LF so the test proves log-forging protection.

With the FIFO and Range issues addressed, this looks ready for another pass.

msureshkumar88:
Thanks — verified and fixed both.

**FIFO blocking the event loop.** Confirmed: I reproduced it the same way (a bare `os.open(fifo, O_RDONLY | O_NOFOLLOW)` with no writer hangs; adding `O_NONBLOCK` returns immediately). `_open_confined_posix()`'s final-component open now uses `os.O_RDONLY | os.O_NOFOLLOW | os.O_NONBLOCK`. Regular-file opens/reads are unaffected — `O_NONBLOCK` only matters for FIFOs/devices — and the existing `fstat()`/`S_ISREG` check rejects the FIFO once the open no longer blocks to get there.

**Case-sensitive Range unit.** Confirmed against Starlette's own parser (`units.strip().lower()`), so `Bytes=`/`BYTES=` did regress. Split the unit from the value with `partition("=")` and compare via `range_unit.strip().lower() == "bytes"`.

**Sanitization test.** Agreed the old version would pass even with `sanitize_for_log()` removed. Rewrote it to plant a real symlink on disk whose filename literally contains `\r\nCRITICAL:root:INJECTED-FAKE-LOG-LINE`, and assert the marker text shows up in the log record with no raw CR/LF — so it actually fails without sanitization.

While adding regression coverage for the FIFO fix I hit a pre-existing test-isolation bug: `test_set_logging_service` reassigns the `mcpgateway.admin.LOGGER`/`logging_service` module globals directly and never restores them, so it was leaving `admin.LOGGER` as a `MagicMock` for the rest of the test session — my new caplog-based test only surfaced it because it was the first test after that one to actually depend on real logging behavior. Fixed by having it `monkeypatch.setattr` those globals (harmless self-assignment at set time, but it registers the original value for automatic restore at teardown).

Added regression tests for the FIFO fix (skips on platforms without `os.mkfifo`, bounded by `pytest.mark.timeout` so a regression fails fast instead of hanging CI) and the case-insensitive range unit (`Bytes=`, `BYTES=`, and padded whitespace).

Rebased onto main and regenerated `.secrets.baseline`.

vishu-bh (dismissed):
LGTM 🚀

Lang-Akshay (changes requested):
## Blocking Changes

| 1 | Breaking Change | `mcpgateway/admin.py` / `mcpgateway/utils/paths.py` | 15559-15563 / 255-256 | On Windows and other platforms lacking atomic `dir_fd`/`O_NOFOLLOW` traversal, the previously available download endpoint now returns HTTP 501. This changes an existing HTTP/error contract and removes functionality for supported Windows deployments. The project is post-1.0 (`1.0.8`), and the PR body does not explicitly acknowledge or justify this break. | Preserve secure Windows/reparse-point handling, or explicitly document and version/migrate this intentional breaking behavior before merge. |

msureshkumar88:
Rebased onto \`main\` and addressed the Windows-compatibility finding (thanks @Lang-Akshay for flagging the 501 regression, and @vishu-bh for the original platform-support callout).

`open_confined()` no longer fails closed on platforms without `dir_fd`/`O_NOFOLLOW` support. It now falls back to `_open_confined_fallback()`, which walks the requested path component by component and rejects a symlink or NTFS reparse point (junction) at any component — including the final one — before it's traversed or opened. This isn't atomic with the following open (a process able to write into the log directory could in principle still win a race in the gap between the check and the open), but it closes the gap the original single `resolve()` + string-prefix check left wide open, and — most importantly for this finding — it preserves the log-download endpoint's existing behavior on Windows instead of turning it into a hard 501.

The POSIX `openat()`/`O_NOFOLLOW` chain (the fully TOCTOU-proof path) is unchanged.

Also folded in the previous round's FIFO-blocking and Range-header case-sensitivity fixes, which are unaffected by this change. Full test suite, ruff/bandit/pylint/interrogate, and `detect-secrets-scan` all pass; secrets baseline regenerated against current `main`.

Lang-Akshay (approved):
LGTM✅

PR #6395, opened 2026-08-24
feat(vault): support A2A agents wrapped as MCP tools

### LINKED ISSUE #6394: [BUG]: Vault plugin does not handle A2A servers defined as tools
## Description

The Vault plugin currently handles token injection for:
- MCP tools via `tool_pre_invoke` hook
- A2A agents via `agent_pre_invoke` hook when invoked directly through `/a2a/{name}/invoke`

However, when an A2A agent is wrapped as an MCP tool and invoked via the MCP protocol (`tools/call`), the Vault plugin does not detect it and fails to inject the appropriate vault token.

## Steps to Reproduce

1. Create an A2A agent with a system tag: `system:example.local`
2. Save a vault token for that system: `{"example.local": "tok-secret"}`
3. The A2A agent is automatically wrapped as an MCP tool (tool name: `a2a-{agent-name}`)
4. Create a virtual server with this tool
5. Invoke the tool via MCP protocol with `X-Vault-Tokens: {"example.local": "tok-secret"}`

**Expected:** Vault plugin injects `Authorization: Bearer tok-secret` to the A2A server

**Actual:** Vault plugin does not detect the A2A-backed tool and does not inject the token

## Root Cause

The `tool_pre_invoke` hook in the Vault plugin only checks the tool's own tags, not whether the tool is backed by an A2A agent that has tags. A2A-backed tools have an `a2a_agent_id` annotation that points to the underlying agent, but this metadata is not loaded or checked.

## Impact

Users cannot use vault token injection for A2A agents when they are invoked through virtual servers as MCP tools. This breaks the security model where vault tokens should be injected based on system tags regardless of invocation path.

## Workaround

None currently. Must invoke the A2A agent directly via `/a2a/{name}/invoke` instead of through the MCP tool protocol.

popagruia:
Enables Vault plugin to inject bearer tokens for A2A agents invoked via MCP protocol (tools/call).

**Changes:**
- Vault plugin now detects A2A-backed MCP tools via a2a_agent_id annotation
- Loads A2A agent metadata to check for system tags
- Injects vault token as Authorization header when system tag matches
- Strips X-Vault-Tokens header for security (defense-in-depth)
- Added defense-in-depth X-Vault-Tokens stripping in tool_service.py (A2A tool path)

**Testing:**
- E2E test added: tests/e2e/test_vault_plugin_a2a_e2e.py
- Covers 3 paths: MCP tool, A2A agent, A2A-as-MCP-tool
- All tests passing
- Manually verified with curl commands against live test environment

Closes #6394

### DIFF
--- a/mcpgateway/services/a2a_service.py
+++ b/mcpgateway/services/a2a_service.py
@@ -2322,6 +2322,20 @@ class A2AAgentService(BaseService):
                         # plugins from injecting sensitive headers into downstream requests
                         # (PR #5183 review fix)
                         plugin_returned = pre_result.modified_payload.headers.model_dump()
+
+                        # Defense in depth: a plugin (e.g. Vault) that determined the
+                        # destination requires managed credentials it couldn't supply signals
+                        # this by returning "authorization": "" -- read here, off the plugin's
+                        # raw returned headers, since checking only the post-_refilter_plugin_headers
+                        # `safe_headers` below would silently drop the sentinel (and thus never
+                        # strip the real header) whenever Authorization isn't in this agent's
+                        # passthrough_headers or sensitive passthrough is disabled. A real bearer
+                        # token is never empty, so this is unambiguous. The plugin never receives
+                        # the real Authorization value on this path (filtered out of
+                        # plugin_headers before this hook runs, by design, per #4925), so it can
+                        # only ever emit this sentinel, never a real value, through this channel.
+                        auth_mismatch = plugin_returned.get("authorization") == ""
+
                         safe_headers = self._refilter_plugin_headers(
                             plugin_headers=plugin_returned,
                             agent=agent,
@@ -2343,6 +2357,16 @@ class A2AAgentService(BaseService):
                                     del prepared.headers[existing_key]
 
                         prepared.headers.update(safe_headers)
+
+                        # Apply the Authorization-mismatch strip last, after the update() above,
+                        # so it can't be undone by that merge re-adding the sentinel itself -- and
+                        # so the header is actually removed rather than left present with an
+                        # empty value, which some downstream servers treat differently from an
+                        # absent header. Unconditional: applies regardless of allowlist/flag
+                        # state.
+                        if auth_mismatch:
+                            for existing_key in [hk for hk in prepared.headers if hk.lower() == "authorization"]:
+                                del prepared.headers[existing_key]
 
                         # Log security-blocked headers for forensic awareness
                         if plugin_returned.keys() - safe_headers.keys():
--- a/mcpgateway/services/tool_service.py
+++ b/mcpgateway/services/tool_service.py
@@ -6758,6 +6758,11 @@ class ToolService(BaseService):
                         correlation_id=get_correlation_id(),
                     )
 
+                    # Final safety strip: X-Vault-Tokens must never reach the downstream A2A agent,
+                    # even if prepare_a2a_invocation added auth headers from agent config or passthrough
+                    # preserved it. This ensures the security invariant holds regardless of plugin state.
+                    prepared.headers = {hk: hv for hk, hv in prepared.headers.items() if hk.lower() != "x-vault-tokens"}
+
                     with create_child_span("tool.gateway_call", {"tool.name": name, "tool.id": tool_id, "tool.integration_type": "A2A"}):
                         # Make HTTP request with timeout enforcement
                         logger.info("Calling A2A agent '%s' at %s", a2a_agent_name, prepared.sanitized_endpoint_url)
--- a/mcpgateway/services/tool_service.py
+++ b/mcpgateway/services/tool_service.py
@@ -6732,12 +6732,36 @@ class ToolService(BaseService):
                             arguments = payload.args
                             if payload.headers is not None:
                                 plugin_returned_headers = payload.headers.model_dump()
+
+                                # Defense in depth: a plugin (e.g. Vault) that determined the
+                                # destination requires managed credentials it couldn't supply
+                                # signals this by returning "authorization": "" -- read here, off
+                                # the plugin's raw returned headers, since checking only the
+                                # post-allowlist-filtered `safe_headers` below would silently
+                                # drop the sentinel (and thus never strip the real header)
+                                # whenever Authorization isn't in this agent's passthrough_headers
+                                # or sensitive passthrough is disabled. A real bearer token is
+                                # never empty, so this is unambiguous.
+                                auth_mismatch = plugin_returned_headers.get("authorization") == ""
+
                                 if a2a_allowlist:
                                     allowlist_lower = {h.lower() for h in a2a_allowlist}
                                     safe_headers = {k: v for k, v in plugin_returned_headers.items() if k.lower() in allowlist_lower}
                                     if not settings.enable_sensitive_header_passthrough:
                                         safe_headers = filter_sensitive_headers(safe_headers)
                                     headers.update(safe_headers)
+
+                                # Apply the strip last, after the allowlist merge above, so it
+                                # can't be undone by that merge re-adding the sentinel itself (as
+                                # opposed to the stale value, which this also guards against) --
+                                # and so the header is actually removed rather than left present
+                                # with an empty value, which some downstream servers treat
+                                # differently from an absent header. Unconditional: applies
+                                # regardless of allowlist/flag state, so the caller's stale
+                                # credential can never reach a destination the plugin explicitly
+                                # flagged as mismatched.
+                                if auth_mismatch:
+                                    headers = {hk: hv for hk, hv in headers.items() if hk.lower() != "authorization"}
 
                     # Defense in depth: strip X-Vault-Tokens (case-insensitive) from outbound
                     # headers. The Vault plugin removes this header when it processes the token,
--- a/mcpgateway/services/tool_service.py
+++ b/mcpgateway/services/tool_service.py
@@ -6700,8 +6700,8 @@ class ToolService(BaseService):
                         )
                         if not settings.enable_sensitive_header_passthrough:
                             headers = filter_sensitive_headers(headers)
-                    # Pass headers to plugin - respect the sensitive header passthrough setting
-                    plugin_headers = filter_sensitive_headers(headers) if not settings.enable_sensitive_header_passthrough else headers
+                    # Plugins always see filtered headers for security reasons
+                    plugin_headers = filter_sensitive_headers(headers)
 
                     # Plugin hook: tool pre-invoke for A2A
                     plugin_manager = await self._get_plugin_manager(plugin_context_id)
@@ -6757,11 +6757,6 @@ class ToolService(BaseService):
                         base_headers=headers,
                         correlation_id=get_correlation_id(),
                     )
-
-                    # Final safety strip: X-Vault-Tokens must never reach the downstream A2A agent,
-                    # even if prepare_a2a_invocation added auth headers from agent config or passthrough
-                    # preserved it. This ensures the security invariant holds regardless of plugin state.
-                    prepared.headers = {hk: hv for hk, hv in prepared.headers.items() if hk.lower() != "x-vault-tokens"}
 
                     with create_child_span("tool.gateway_call", {"tool.name": name, "tool.id": tool_id, "tool.integration_type": "A2A"}):
                         # Make HTTP request with timeout enforcement
--- a/mcpgateway/services/tool_service.py
+++ b/mcpgateway/services/tool_service.py
@@ -6758,6 +6758,13 @@ class ToolService(BaseService):
                         correlation_id=get_correlation_id(),
                     )
 
+                    # Final safety strip: X-Vault-Tokens must never reach the downstream A2A agent,
+                    # even if prepare_a2a_invocation added auth headers from agent config or passthrough
+                    # preserved it. PreparedA2AInvocation is frozen, so mutate the headers dict in
+                    # place (matches the pattern in a2a_service.py) rather than rebinding the field.
+                    for existing_key in [hk for hk in prepared.headers if hk.lower() == "x-vault-tokens"]:
+                        del prepared.headers[existing_key]
+
                     with create_child_span("tool.gateway_call", {"tool.name": name, "tool.id": tool_id, "tool.integration_type": "A2A"}):
                         # Make HTTP request with timeout enforcement
                         logger.info("Calling A2A agent '%s' at %s", a2a_agent_name, prepared.sanitized_endpoint_url)
--- a/mcpgateway/services/tool_service.py
+++ b/mcpgateway/services/tool_service.py
@@ -6700,7 +6700,8 @@ class ToolService(BaseService):
                         )
                         if not settings.enable_sensitive_header_passthrough:
                             headers = filter_sensitive_headers(headers)
-                    plugin_headers = filter_sensitive_headers(headers)
+                    # Pass headers to plugin - respect the sensitive header passthrough setting
+                    plugin_headers = filter_sensitive_headers(headers) if not settings.enable_sensitive_header_passthrough else headers
 
                     # Plugin hook: tool pre-invoke for A2A
                     plugin_manager = await self._get_plugin_manager(plugin_context_id)
@@ -6737,6 +6738,12 @@ class ToolService(BaseService):
                                     if not settings.enable_sensitive_header_passthrough:
                                         safe_headers = filter_sensitive_headers(safe_headers)
                                     headers.update(safe_headers)
+
+                    # Defense in depth: strip X-Vault-Tokens (case-insensitive) from outbound
+                    # headers. The Vault plugin removes this header when it processes the token,
+                    # but stripping unconditionally prevents leakage when the plugin is disabled,
+                    # errors in permissive mode, or the header is mistakenly in passthrough_allowed.
+                    headers = {hk: hv for hk, hv in headers.items() if hk.lower() != "x-vault-tokens"}
 
                     prepared = prepare_a2a_invocation(
                         agent_type=a2a_agent_type,
--- a/plugins/vault/echo_a2a.py
+++ b/plugins/vault/echo_a2a.py
@@ -2,9 +2,11 @@
 """Throwaway A2A echo server for Vault plugin E2E validation.
 
 Runs a minimal FastAPI app that reflects the headers it received back in the JSON
-response body. Point an A2A agent's endpoint_url at http://localhost:8002/invoke so
-we can assert the gateway forwarded the vault-injected Authorization header and
-stripped the X-Vault-Tokens header.
+response body. Point an A2A agent's endpoint_url at http://127.0.0.1:8002/invoke
+(not "localhost" -- this server binds IPv4 only, and the gateway's SSRF-hardening
+DNS pinning connects to whichever address "localhost" resolves to first, which is
+::1 on IPv6-first hosts) so we can assert the gateway forwarded the vault-injected
+Authorization header and stripped the X-Vault-Tokens header.
 
 Run:
     python plugins/vault/echo_a2a.py
--- a/plugins/vault/echo_mcp.py
+++ b/plugins/vault/echo_mcp.py
@@ -10,6 +10,9 @@ Copyright contributors to the MCP-CONTEXT-FORGE project
 SPDX-License-Identifier: Apache-2.0
 """
 
+# Standard
+import os
+
 # Third-Party
 from fastmcp import Context, FastMCP
 from fastmcp.server.dependencies import get_http_headers
@@ -98,4 +101,5 @@ def get_version(ctx: Context):
 
 
 if __name__ == "__main__":
-    mcp.run(transport="sse", port=8001)
+    port = int(os.environ.get("MCP_ECHO_PORT", "8001"))
+    mcp.run(transport="sse", port=port)
--- a/plugins/vault/vault_plugin.py
+++ b/plugins/vault/vault_plugin.py
@@ -237,7 +237,7 @@ class Vault(Plugin):
         """
         try:
             # First-Party
-            from mcpgateway.db import A2AAgent as DbA2AAgent, get_db  # pylint: disable=import-outside-toplevel
+            from mcpgateway.db import A2AAgent as DbA2AAgent  # pylint: disable=import-outside-toplevel
             from sqlalchemy import select  # pylint: disable=import-outside-toplevel
 
             gen = get_db()
@@ -364,7 +364,7 @@ class Vault(Plugin):
         else:
             # Try to find a key that starts with system_key (complex key format)
             for key in vault_tokens.keys():
-                parsed_system, scope, token_type, token_name = self._parse_vault_token_key(key)
+                parsed_system, _, _, _ = self._parse_vault_token_key(key)
                 if parsed_system == system_key:
                     resolved = self._resolve_token_value(vault_tokens[key], destination_url)
                     if resolved is None:
@@ -376,7 +376,7 @@ class Vault(Plugin):
 
         if token_value and token_key_used:
             # Parse the token key to determine handling
-            parsed_system, scope, token_type, token_name = self._parse_vault_token_key(token_key_used)
+            parsed_system, _, token_type, _ = self._parse_vault_token_key(token_key_used)
             # Determine how to handle the token based on token_type and AUTH_HEADER tag
             if token_type == "PAT":
                 # Handle Personal Access Token
--- a/plugins/vault/vault_plugin.py
+++ b/plugins/vault/vault_plugin.py
@@ -435,16 +435,30 @@ class Vault(Plugin):
             logger.debug("Injected auth header for system: %s", system_key)
         elif not token_value:
             # No vault token found for the expected system. Since the target has a system tag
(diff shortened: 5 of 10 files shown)


Lang-Akshay (changes requested):
Thanks for the PR @popagruia 
## Please address following blocking changes

| 1 | Security | `mcpgateway/services/tool_service.py` | 6685, 6691 | When sensitive-header passthrough is enabled, `headers.update(safe_headers)` preserves an inbound `Authorization` that the Vault plugin removed on a token mismatch. `prepare_a2a_invocation` can also re-add configured `authheaders` after the pre-prepare `X-Vault-Tokens` strip, allowing sensitive credentials to reach the downstream A2A agent. | Reconcile plugin output as a replacement/sanitized header set, including case-insensitive removals, rather than only updating the original map. Strip `X-Vault-Tokens` again after `prepare_a2a_invocation` as the final outbound invariant, including headers added by configured auth. |
| 2 | Security | `plugins/vault/vault_plugin.py` | 152, 156 | Destination-binding mismatch warnings log raw `mcpServer` and `destination_url` values. URL userinfo or query credentials can therefore be written to application logs. | Redact userinfo, query, and fragment values before logging; log only sanitized scheme/host/path or equivalent non-secret fields. |

popagruia:
## Security fixes applied ✅

Both blocking issues have been addressed in commit 351694330:

### Issue 1: Header leakage after plugin processing
- **Fix**: Added final safety strip of `X-Vault-Tokens` after `prepare_a2a_invocation()`
- **Location**: `mcpgateway/services/tool_service.py:6707-6710`
- **Ensures**: Sensitive headers never reach downstream A2A agents regardless of plugin state or auth config

### Issue 2: Credentials in logged URLs  
- **Fix**: Added `_sanitize_url_for_logging()` helper that strips userinfo, query params, and fragments
- **Location**: `plugins/vault/vault_plugin.py:121-140`
- **Applied to**: Warning logs at lines 176 and 181
- **Ensures**: Only scheme://host/path logged, never passwords or API keys

Ready for re-review. Thank you @Lang-Akshay for the security review!

popagruia:
Thanks @Lang-Akshay — both items are addressed.

**#1**: `prepare_a2a_invocation()`'s outbound headers are now stripped of `X-Vault-Tokens` as a final safety check right before the HTTP call, so it can't leak even if agent config or passthrough reintroduces it.

**#2**: Warning logs in `vault_plugin.py` now sanitize URLs before logging — only `scheme://host/path`, no credentials/query/fragment.

CI is green. Ready for another look whenever convenient.

Lang-Akshay (dismissed):
LGTM✅

msureshkumar88 (changes requested):
## E2E verification against a live gateway

Thanks for this — the core fix works exactly as intended, and while verifying it end-to-end I found one thing worth a look before merging.

**Setup**: real `uv run uvicorn mcpgateway.main:app` process, real SQLite DB via the app's own startup lifespan, real JWT via `mcpgateway.utils.create_jwt_token`, real `PluginManager` loading the Vault plugin from `plugins/vault/config_vault_e2e.yaml`, and the two backend stubs this PR ships (`plugins/vault/echo_mcp.py`, `plugins/vault/echo_a2a.py`) — actual HTTP over the loopback, nothing mocked. Script: `scripts/verify-pr-6395-e2e.sh` (branch `feature/vault-plugin-a2a-as-mcp-tool` @ `24d4a43fb`).

Config (matches this PR's own `test_vault_plugin_a2a_e2e.py` fixture):
```
PLUGINS_ENABLED=true
PLUGINS_CONFIG_FILE=plugins/vault/config_vault_e2e.yaml
ENABLE_HEADER_PASSTHROUGH=true
ENABLE_SENSITIVE_HEADER_PASSTHROUGH=true
MCPGATEWAY_A2A_ENABLED=true
AUTH_REQUIRED=true
SSRF_ALLOW_LOCALHOST=true
```

### What passed

- **The actual fix, reproducing issue #6394's repro steps exactly**: registered an A2A agent with `system:echo.local`, confirmed the auto-created `a2a-{slug}` tool, attached it to a virtual server, called it via `tools/call` with `X-Vault-Tokens`. `Authorization: Bearer <token>` reached the real upstream `echo_a2a` stub, `X-Vault-Tokens` did not. This is the gap #6394 reported — confirmed fixed.
- Direct `/a2a/{name}/invoke` path — no regression.
- Plain MCP-tool vault path (`tool_pre_invoke` on a registered gateway) — no regression.
- Untagged agent / no vault header — plugin correctly stays out of the way.
- This PR's own `tests/e2e/test_vault_plugin_a2a_e2e.py` — 3/3 pass.
- Neighboring unit suites this PR didn't touch (`test_vault_plugin.py`, `test_vault_plugin_smoke.py`, `test_tool_service_coverage.py`) — all green, no regression.
- `ruff check` on the four changed source files — clean.
- No Alembic migration in the diff — correct, no schema change needed for this fix.

### One finding — the new "strip stale Authorization on vault mismatch" hardening doesn't fire on the A2A path

This PR added (`vault_plugin.py`, `_apply_vault_token`):
```python
if "authorization" in headers:
    del headers["authorization"]
    logger.warning("... stripped Authorization header to prevent credential leakage ...")
```
Intent, per the commit message: when a system tag exists but the supplied vault tokens don't match, strip any existing `Authorization` header rather than forward a possibly-wrong credential.

Reproduced live: agent tagged `system:echo.local`, called via `tools/call` with `X-Vault-Tokens: {"other.system": "..."}` (deliberate mismatch) plus `X-Upstream-Authorization: Bearer stale-client-cred` (the existing, always-on header→`Authorization` rename in `compute_passthrough_headers_cached`). The stale credential reached the real upstream:

```json
"received_headers":{"authorization":"Bearer stale-client-cred", ...}
```

Gateway's own debug log for that request shows why — the *old* message fired, not the new one:
```
DEBUG plugins.vault.vault_plugin: Removed vault header 'x-vault-tokens' from headers
WARNING plugins.vault.vault_plugin: Vault tokens provided but no match found for system 'echo.local' - possible misconfiguration
```
(the new wording — "...stripped Authorization header to prevent credential leakage...") never appears, meaning `"authorization" in headers` was `False` inside the plugin.

Root cause: on the A2A branch (`tool_service.py`), `plugin_headers = filter_sensitive_headers(headers)` runs *unconditionally* before the plugin is invoked — regardless of `ENABLE_SENSITIVE_HEADER_PASSTHROUGH` — so `Authorization` is never in what the plugin sees; its `del headers["authorization"]` is checking a dict that structurally can't contain that key. The outer merge afterward (`headers.update(safe_headers)`) is additive-only, so even if the plugin *did* return a diff, there'd be no way for it to signal "remove this key" back to the outer, unfiltered `headers` — which is what actually reaches `prepare_a2a_invocation()`. That's presumably why `X-Vault-Tokens` needed the separate unconditional strip added later in `tool_service.py` (lines ~6699 and ~6718) — the same treatment was never extended to this new `Authorization`-on-mismatch branch.

Net effect: the credential-leakage guard this PR adds is currently a no-op on the exact path (A2A / A2A-as-tool) the PR is about, under the exact config (`ENABLE_SENSITIVE_HEADER_PASSTHROUGH=true`) the PR's own e2e test uses.

**Suggestion**: mirror the `X-Vault-Tokens` treatment — add an unconditional strip of `Authorization` in `tool_service.py` immediately after the pre-invoke hook when the plugin signals a mismatch (or more simply, don't let `filter_sensitive_headers` remove `Authorization` from what the plugin sees on this path, since the plugin is explicitly trusted to make an auth decision here). Happy to send a small repro test or a patch if useful.

### Not evaluated

Static/architecture points from earlier discussion (whether the A2A-agent DB lookup in `_load_a2a_agent_metadata` could reuse the `tags`/`url` already on the auto-created tool instead of a second query, and unit-test coverage for the new plugin code) weren't part of this e2e pass — flagging separately, not blocking on their own.

---

**Verdict**: Requesting changes — not for the core fix, which works, but the mismatch-hardening gap above should be closed (or the claim narrowed) before merge, since it's the security property called out in the commit message.

popagruia:
Thanks again for the detailed repro — aligning on your suggestion #1. Here's the concrete plan before we implement it:

**The fix (sentinel via the existing headers channel, not `metadata`):**

1. `vault_plugin.py::_apply_vault_token`, on the mismatch branch — since `authorization` is never actually present in what the plugin receives on this path (confirmed: `tool_service.py`'s `plugin_headers = filter_sensitive_headers(headers)` strips it before the hook runs, by design, per #4925), the existing `del headers["authorization"]` is dead code. Replace it with `headers["authorization"] = ""` in the returned payload — empty string as an unambiguous sentinel (a real token is never empty), returned through the same `payload.headers` channel the plugin already uses.
2. `tool_service.py` / `a2a_service.py`, right where the plugin's returned headers are read back (`plugin_returned_headers = payload.headers.model_dump()`) — check for that sentinel **before** the `a2a_allowlist`/`filter_sensitive_headers` gating runs on it, and if present, delete `Authorization` from the real outbound headers unconditionally. This ordering matters: if we checked *after* those filters (or relied on the normal `headers.update(safe_headers)` merge), the sentinel would get silently dropped whenever `Authorization` isn't in the agent's `passthrough_headers` allowlist or `ENABLE_SENSITIVE_HEADER_PASSTHROUGH` is off — i.e., it would only work under the exact config your repro used, not in general. Checking the raw returned dict first avoids that.
3. Mirror in `a2a_service.py`'s equivalent merge point for the direct-A2A path.

**Possible future direction, not part of this PR:** it's worth noting that the framework already defines `read_headers`/`write_headers` capabilities (`cpex.framework.extensions.tiers.Capability`), which could in the future let a plugin declare ownership of `Authorization` and see/modify it directly via `extensions.http.headers` — at which point this plugin's original `del headers["authorization"]` would just work as written, no sentinel needed. That would require migrating off the deprecated `ToolPreInvokePayload.headers` field first, which is a larger, separate piece of work. Flagging it as a possible future improvement, not something we're pursuing here.

Also adding the missing test coverage — the existing mismatch tests in `test_vault_plugin.py` never actually put `authorization` in the input, so they've never exercised this path even at the unit level.

Does this plan look right to you before we implement it?

msureshkumar88:
Thanks for writing the plan out before implementing — that saved a round trip, and the reasoning in step 2 about checking the sentinel *before* the allowlist/`filter_sensitive_headers` gating is exactly right. That ordering detail is the part that's easy to get wrong, and you caught it up front.

**On the plan itself: yes, the sentinel shape is right for this specific item.** The plugin is structurally blind to `Authorization` on this path, so it does need an out-of-band signal — a diff-based "what did the plugin remove" reconciliation can't help here, precisely because the plugin never saw the header in the first place. Agreed on parking the `read_headers`/`write_headers` capability migration as separate work.

A few things I think are worth folding in before it lands, mostly about blast radius of the sentinel rather than the idea itself.

### Sentinel needs to be handled at every merge site, not just the A2A one

`plugin_returned_headers` only exists at `tool_service.py:6687` (the A2A branch). Two other sites read plugin headers back wholesale:

- `tool_service.py:6587` — the MCP/gateway branch
- `tool_service.py:5645` — the REST branch

Both do `headers = payload.headers.model_dump()`, replacing the map entirely. That matters because on the MCP branch the current `del headers["authorization"]` **does** work today (the plugin gets unfiltered headers there, `:6564`). So swapping the `del` for a sentinel without teaching that branch about it would turn a working deletion into an outbound `Authorization: ` empty header — a regression on the one path that behaves correctly right now. Worth handling all three (plus the `a2a_service.py` mirror you mention).

Same reason on the direct-A2A path: `_refilter_plugin_headers` (`a2a_service.py:1991`) does a straight dict comprehension with no empty-value filtering, so an unhandled sentinel would survive into `prepared.headers.update(safe_headers)` and go out as an empty header rather than being removed.

### The sentinel interacts with the OAuth deny-path check

`tool_service.py:6601`:

```python
if oauth_authcode_no_db_token and not any(hk.lower() == "authorization" for hk in headers):
    raise ToolInvocationError(f"Please authorize {gateway_name} first. Visit /oauth/authorize/{gateway_id_str} ...")
```

An empty-string `authorization` key satisfies "present", so this would suppress the actionable authorize-me error and send an empty credential upstream instead — turning a clear local message into an opaque upstream 401. If the sentinel is resolved (deleted) before this check on the MCP branch, it's fine; just wanted to flag the ordering constraint since it's a second one beyond the allowlist gating.

### Separate from the sentinel: header removals in general

`tool_service.py:6693` is still `headers.update(safe_headers)`, which is additive-only — with `X-Vault-Tokens` handled by a hardcoded strip afterward, and now `Authorization` by the sentinel. Any *other* header a `tool_pre_invoke` plugin drops on this path is still silently restored.

`a2a_service.py:2330-2341` already solves the general case for the direct-A2A path (diff what the plugin was given against what it returned, delete the difference). Would it be worth porting that loop to the A2A branch in `tool_service.py` in the same pass? It's complementary to the sentinel rather than an alternative — reconciliation covers headers the plugin can see, the sentinel covers the one it can't — and it would mean not adding a third single-header carve-out later.

### One scope question on the mismatch branch itself

Independent of where the strip fires: on the MCP path, `headers` at that point starts from the gateway's own configured credential (`decode_auth(gateway_auth_value)`, `tool_service.py:6028`). So for a gateway that is both `system:`-tagged *and* has its own static/OAuth auth configured, a caller sending a blanket `X-Vault-Tokens` that happens not to cover that system would now have the gateway's own working credential removed — a previously-successful call becomes a 401.

Is that the intended policy? If the intent is narrower ("don't forward a *client-supplied* credential when vault handling was expected"), scoping the strip to that case would avoid the regression. Either way it's a behavior change worth a line in the PR description / CHANGELOG so operators aren't surprised.

### Two smaller items while you're in here

- `tests/e2e/test_vault_plugin_a2a_e2e.py:207,251` still register `endpoint_url: "http://localhost:8002/invoke"`, which is the exact thing this PR's own `echo_a2a.py` docstring update warns against (`::1` first on IPv6-first hosts). Both should probably be `127.0.0.1` to match.
- On the test coverage point you raised — agreed on the mismatch branch. Worth also covering `_load_a2a_agent_metadata`'s agent-not-found, DB-error, and malformed-annotation paths. Heads up that `pyproject.toml:805` scopes coverage `source` to `mcpgateway`, so `plugins/` doesn't reach the diff-cover gate — the ~100 new lines in `vault_plugin.py` won't show up there either way.

Happy to re-verify against a live gateway once this lands. The core A2A-backed-tool detection is solid work, and the fix for it is close.

popagruia:
Thanks for the careful read — a few of these turned out to be scope-conditional, so let me be precise about what the diff actually touches.

**Scope clarification (covers items 1, 2, and 4):** the sentinel only lands in the two A2A branches — `tool_service.py`'s A2A-as-tool branch and `a2a_service.py`'s direct-A2A branch (`agent_pre_invoke`). Neither the MCP/gateway branch (`tool_service.py:6587`) nor the REST branch (`:5645`) are touched — they keep the plugin's existing `del headers["authorization"]`, which already works there since the plugin sees real, unfiltered headers on those paths. So:
- The OAuth deny-path check at `:6601` is on the MCP branch and isn't affected — the sentinel never reaches it, since it's not in scope for this fix.
- The gateway-own-credential concern is also MCP-branch-specific (`decode_auth(gateway_auth_value)` at `:6028`) — the A2A branch's `headers` never contains a gateway-configured credential, only whatever the client passed through (`compute_passthrough_headers_cached` from `request_headers`), so there's no equivalent "strip the gateway's own working credential" risk on the path this fix actually changes.

Good catch either way — if the sentinel pattern ever gets extended to the MCP/REST branches later, both of these become real constraints to handle at that time.

**Item 3 (general reconciliation loop):** agreed this is a real gap and porting `a2a_service.py`'s diff-based reconciliation into `tool_service.py`'s A2A branch would be the more complete fix. Proposing to track it as separate follow-up work rather than fold it into this PR — it's a broader change to plugin-header-removal semantics on that branch, beyond the specific finding this PR is closing.

**`localhost` → `127.0.0.1`:** fixed both occurrences in `test_vault_plugin_a2a_e2e.py` (lines 207, 251).

**Coverage-scope note:** appreciated, noted — not something to act on for this PR, just useful context.

Will push the sentinel implementation (as scoped above) plus the test file fix shortly.

msureshkumar88:
That's a fair correction, and I checked both claims against the current diff — they hold:

- `tool_service.py:6564` — the MCP branch does pass the plugin unfiltered `headers` (no `filter_sensitive_headers` before the hook there), so `del headers["authorization"]` does fire on that path today. Agreed the OAuth deny-path check at `:6601` and the empty-value regression I raised are moot if the sentinel stays scoped to the two A2A branches.
- `tool_service.py:6642` — the A2A branch does pass `gateway_auth_type=None` explicitly into `compute_passthrough_headers_cached`, so `headers` there really is client-passthrough-only, never a gateway-configured credential. Good catch — that risk doesn't apply to the branch this fix touches.

One thing I want to separate out, though: **that same fact means item 3 isn't resolved by scoping — it confirms it's live.** Since `del headers["authorization"]` *does* fire on the MCP branch (as you just confirmed), the mismatch-strip in `vault_plugin.py` is wiping the gateway's real, working credential there today, independent of whatever lands for the A2A sentinel. It's not a "what if this pattern gets extended later" concern — it's current behavior on a branch this PR already ships code for. Two ways to close it:

1. Scope the strip in `_apply_vault_token` so it only fires when the target has no other credential source configured (i.e., don't strip a gateway's own static/OAuth `Authorization` just because an unrelated `X-Vault-Tokens` entry didn't match), or
2. If wiping it is intentional policy (e.g. "vault-tagged means vault-only, no fallback"), that's fine — just wanted to confirm that's the deliberate call before this merges, and if so it's worth a line in the PR description since it's a behavior change for anyone with a vault tag *and* a configured gateway credential.

Happy to leave the general reconciliation loop as tracked follow-up rather than blocking this PR on it — that's a reasonable scope call, do you have (or want to open) an issue for it so it doesn't get lost?

`localhost` → `127.0.0.1` fix and the sentinel scoped to the two A2A branches both sound right to me. Will take another pass once pushed.

msureshkumar88 (dismissed):
Approving. Ran a live e2e verification against a real gateway (uvicorn + SQLite + real JWT + real `PluginManager` loading the Vault plugin, nothing mocked) reproducing #6394's exact repro steps plus the sentinel fix from `138bf86b7` — the issue is fixed, and I didn't find any regressions in the surrounding vault/A2A test suites (`test_vault_plugin.py`, `test_a2a_service.py`, `test_a2a_agent_invoke_hooks.py`, `test_a2a_protocol.py`, full `test_tool_service.py`/`test_tool_service_coverage.py`, plus this PR's own e2e file — all green).

Two things from earlier in the thread are still open. Not blocking, but noting them so they don't get lost — filed as #6577:

1. **No test for the `a2a_service.py` sentinel-strip branch**, and no acknowledgment on why it's safe to skip. The `tool_service.py` sentinel got a solid regression test; the mirrored logic in `a2a_service.py`'s `agent_pre_invoke` path didn't.
2. **Item 3 is still unanswered** — and I have live confirmation now, not just a code read: registered a gateway tagged `system:echo.local` with its own configured static credential, sent an unrelated `X-Vault-Tokens` mismatch, and the gateway's own working credential was stripped before the outbound call, not just a client-supplied one. Whether that's intended policy or should be scoped tighter is still an open question on the thread.

Great work getting the core fix and the sentinel landed — the A2A-as-MCP-tool detection and the header-ordering fix for the sentinel check are both solid.

msureshkumar88:
## E2E verification details

For reproducibility — full config, commands, and raw output from the live verification referenced in the approval.

### Environment / config

Same shape as this PR's own `tests/e2e/test_vault_plugin_a2a_e2e.py` fixture (`live_stack`), run standalone rather than under pytest:

```bash
# Backends (this PR's own stubs, unmodified)
python3 plugins/vault/echo_mcp.py        # :8001 (SSE)
python3 plugins/vault/echo_a2a.py        # :8002 (127.0.0.1, per this PR's own IPv6 note)

# Gateway
PLUGINS_ENABLED=true
PLUGINS_CONFIG_FILE=plugins/vault/config_vault_e2e.yaml
ENABLE_HEADER_PASSTHROUGH=true
ENABLE_SENSITIVE_HEADER_PASSTHROUGH=true
MCPGATEWAY_A2A_ENABLED=true
JWT_SECRET_KEY=<throwaway>
DATABASE_URL=sqlite:////tmp/verify6395.db
AUTH_REQUIRED=true
EXPOSE_ERROR_DETAILS=true
SSRF_ALLOW_LOCALHOST=true

python3 -m uvicorn mcpgateway.main:app --host 127.0.0.1 --port 4444
```

`plugins/vault/config_vault_e2e.yaml` (this PR's own file, unmodified) registers the Vault plugin on both `tool_pre_invoke` and `agent_pre_invoke` in `permissive` mode.

Admin JWT minted the same way the shipped test does:
```bash
python3 -m mcpgateway.utils.create_jwt_token --username admin@example.com --exp 10080 --secret <throwaway>
```

### Commands run

```bash
source .venv/bin/activate
python3 verify_pr6395_live.py        # live scenarios 1-5, see below
```

Scenarios (all real HTTP over loopback, nothing mocked):
1. Plain MCP tool path (`tool_pre_invoke`, pre-existing) — regression check.
2. Direct A2A path (`agent_pre_invoke`, pre-existing) — regression check.
3. **Issue #6394's exact repro**: A2A agent tagged `system:echo.local`, auto-wrapped as MCP tool `a2a-{slug}`, invoked via `tools/call` with `X-Vault-Tokens`.
4. **New in this PR**: A2A-tool path, stale credential via `X-Upstream-Authorization` + non-matching `X-Vault-Tokens` — the empty-string sentinel (`138bf86b7`) must strip it.
5. **Evidence for the open item 3 question**: MCP path, gateway registered with its own static `Authorization` (`auth_type: authheaders`) + an unrelated `X-Vault-Tokens` mismatch.

Regression sweep (bypassing a local `pytest` output-filtering hook via `rtk proxy` to get real pytest output):
```bash
rtk proxy python3 -m pytest -q \
  tests/unit/mcpgateway/plugins/plugins/vault/test_vault_plugin.py \
  tests/unit/mcpgateway/plugins/plugins/vault/test_vault_plugin_smoke.py \
  tests/unit/mcpgateway/services/test_a2a_agent_invoke_hooks.py \
  tests/unit/mcpgateway/services/test_a2a_service.py \
  tests/unit/mcpgateway/services/test_a2a_protocol.py

rtk proxy python3 -m pytest -q \
  "tests/unit/mcpgateway/services/test_tool_service_coverage.py::TestInvokeToolA2A" \
  "tests/unit/mcpgateway/services/test_tool_service_coverage.py::TestInvokeA2AToolCoverage" \
  "tests/unit/mcpgateway/services/test_tool_service_coverage.py::TestCallA2AAgentCoverage"

rtk proxy python3 -m pytest -q -m e2e tests/e2e/test_vault_plugin_a2a_e2e.py

rtk proxy python3 -m pytest -q \
  tests/unit/mcpgateway/services/test_tool_service.py \
  tests/unit/mcpgateway/services/test_tool_service_coverage.py

rtk proxy python3 -m pytest -q \
  tests/unit/mcpgateway/test_a2a_agent.py \
  tests/integration/test_vault_integration.py
```

### Output

**Live scenarios:**
```
[PASS] 1. Plain MCP tool path: token injected + X-Vault-Tokens stripped
[PASS] 2. Direct A2A path: token injected + X-Vault-Tokens stripped
[PASS] 3. Issue #6394 repro: A2A-as-MCP-tool injects token + strips X-Vault-Tokens
[PASS] 4. NEW: Authorization-mismatch sentinel strips stale cred on A2A-tool path
       — received.authorization=None
[FAIL] 5. EVIDENCE: gateway's own static credential forwarded despite unrelated vault mismatch?
       — STRIPPED: gateway's own working credential was wiped by the mismatch branch
         (open policy question, filed as #6577)

SUMMARY: 4/5 scenarios passed
```
(Scenario 5 is deliberately an evidence-gathering check, not a pass/fail against an agreed spec — "FAIL" here means "the credential *was* stripped," which is the behavior in question, not a script bug.)

**Regression suites** (all exit code 0, no `F`/`E` in any run):
```
tests/unit/mcpgateway/plugins/plugins/vault/test_vault_plugin.py .......  [ 16%]
... (all dots) ...                                                       [100%]

tests/unit/mcpgateway/services/test_tool_service_coverage.py::TestInvokeToolA2A
tests/unit/mcpgateway/services/test_tool_service_coverage.py::TestInvokeA2AToolCoverage
tests/unit/mcpgateway/services/test_tool_service_coverage.py::TestCallA2AAgentCoverage
....................................................                     [100%]

tests/e2e/test_vault_plugin_a2a_e2e.py (this PR's own shipped e2e)
...                                                                      [100%]

tests/unit/mcpgateway/services/test_tool_service.py
tests/unit/mcpgateway/services/test_tool_service_coverage.py  (~950 tests total)
....................................................................... [100%]
SKIPPED [1] test_tool_service_coverage.py:4903 — pre-existing, tracked under #4210,
  unrelated to this PR (A2A output-schema validation gap in ToolService.invoke_tool)

tests/unit/mcpgateway/test_a2a_agent.py .............                   [100%]
tests/integration/test_vault_integration.py
  10 SKIPPED — "Vault not available" (HashiCorp Vault docker-compose stack, the
  token-storage backend — same-named but unrelated to the Vault *plugin* this PR touches)
```

No failures, no errors, in any suite.

msureshkumar88 (approved):

PR #6397, opened 2026-08-24
[CHORE]: Automate MCP server catalog icon generation

### LINKED ISSUE #5966: [CHORE]: Automate MCP server catalog icon generation (bundled, no runtime egress)
### Summary

Add a maintainer-run (and optionally CI-scheduled) automation that resolves a brand icon for each entry in `mcp-catalog.yml`, normalizes it, commits it as a bundled static asset, and populates `logo_url`. Deployed instances then serve icons locally: no favicon/logo calls are ever made from a user's gateway or browser, which keeps the catalog usable in air-gapped deployments and avoids leaking which servers an org is evaluating to a third-party icon service.

Supports #5770 (MCP server catalog page).

### Motivation

- The catalog ships 135 servers and 0 currently have `logo_url` (also 0 `documentation_url`), so every card renders identically today.
- Fetching icons at runtime from the user's instance is a poor fit for this project's egress/privacy posture (air-gap support, no external calls, CSP). Doing the fetch once at build/curation time and committing the results removes that dependency entirely.

### Current data (analysis of `mcp-catalog.yml`)

Deriving the registrable domain from each entry's `url`:

| Bucket | Count | Icon resolvable? |
|---|---|---|
| Real host to plausible brand domain | ~121 (approx 90%) | Yes |
| `localhost`/stdio servers (e.g. `playwright-mcp`, `azure-mcp`, `fast-time-server`) | 14 | No, placeholder |
| Product hosted on a generic domain (`scorecard`, `onecontext` on `workers.dev`) | ~2 | Wrong-brand, placeholder or manual override |

~90% can be auto-resolved; the remainder falls back to the placeholder.

### Placeholder

Use the existing generic MCP mark for any entry without a resolved icon: `client/src/components/icons/MCPIcon.tsx`. The catalog card should render `logo_url` when present and fall back to `<MCPIcon />` otherwise.

Image

### Proposed automation

A `scripts/fetch_catalog_icons.py` with a `make catalog-icons` target that:

1. Reads `mcp-catalog.yml`; for each entry derives the registrable domain from `url`, skipping `localhost`/IP hosts.
2. Resolves an icon via a fallback chain: the site's own `<link rel="apple-touch-icon">` / `rel="icon">` (highest resolution), then `https://icons.duckduckgo.com/ip3/{domain}.ico`, then `/favicon.ico`.
3. Prefers a sanitized SVG when the source actually serves an `image/svg+xml` favicon, otherwise normalizes to a fixed-size PNG (e.g. 64x64, via Pillow). Rationale below.
4. Writes `catalog-icons/{id}.{svg,png}` into a committed assets directory and sets `logo_url` on the entry to the stable local path.
5. Is idempotent (skip existing, `--force` to refresh) and prints a report of misses that will use the placeholder.
6. Supports a small override/skip list for wrong-brand or trademark-sensitive entries.

### Icon format: prefer sanitized SVG, else PNG

Favicon sources are almost entirely raster (`apple-touch-icon` is always PNG; the DuckDuckGo/Google favicon services return PNG; most `favicon.ico` files are raster), so PNG is the default that actually achieves coverage. SVG is only used when a site publishes a genuine `image/svg+xml` favicon, and even then it must be sanitized before it is committed, since third-party SVGs are XML that can carry `<script>`, `foreignObject`, or external references (an XSS / supply-chain vector). PNG is inert raster data and safe to bundle without sanitization.

- Prefer SVG only when the source serves `image/svg+xml` and it passes sanitization (strip scripts / external refs, or re-render).
- Otherwise emit a normalized PNG. Emit at 2x (e.g. 128px) for retina and display at the card size.
- The placeholder stays SVG (`MCPIcon.tsx`), since it is first-party and trusted.

Optionally, a scheduled GitHub Action runs the script and opens a PR with refreshed icons so the catalog stays current without manual effort.

### Open decision: where icons are served

- Gateway static mount (recommended): serve `catalog-icons/` from the gateway so `logo_url` is a real gateway-relative URL resolved identically by the new React UI and the legacy admin catalog. Small backend addition (a `StaticFiles` mount alongside the existing `static_dir` handling in `main.py`).
- Frontend bundle only: ship PNGs under the React app's static assets and map by catalog `id`; zero backend change, but `logo_url` stays empty in the API and only the new UI benefits.

### Licensing note

Restrict to favicons (low-risk, widely bundled) rather than full wordmark logos. Any brand that objects can be dropped to the placeholder via the override list.

### Acceptance criteria

- [ ] `make catalog-icons` regenerates icons from `mcp-catalog.yml`, idempotently.
- [ ] ~90% of entries get a committed icon (sanitized SVG where available, else PNG); the rest are reported and fall back to `MCPIcon`.
- [ ] Icons are served locally with no external requests from a running gateway instance.
- [ ] `logo_url` is populated for resolved entries; catalog card renders it with `MCPIcon` fallback.
- [ ] Wrong-brand / localhost entries handled via skip/override, not broken images.

### Out of scope

- The catalog page itself (#5770).
- `documentation_url` backfill (separate data task).
- Provenance fields (`catalog_source`, `catalog_version`, `verified_at`).

vishu-bh:
# Pull Request

## 🔗 Related Issue

Closes #5966

---

## 📝 Summary

Adds maintainer-side catalog icon generation. Public icons are fetched only during the build/curation step, normalized to local 128x128 RGBA PNG assets, and referenced by local static URLs. Runtime gateway and browser paths do not fetch third-party icon URLs. Includes safe URL handling, overrides/skips, Makefile integration, packaging metadata, and regression tests.

---

## 📏 Reviewability

---

## 🏷️ Type of Change

- [ ] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (describe below)

---

## 🧪 Verification

| Check | Command | Status |
|---|---|---|
| Lint suite | `git diff --check` | PASS |
| Unit tests | `python -m pytest tests/unit/scripts/test_fetch_catalog_icons.py tests/unit/mcpgateway/services/test_catalog_service.py tests/unit/mcpgateway/routers/test_catalog.py -q` | PASS |
| Coverage ≥ 80% | Not run | N/A |

Additional validation: package wheel built successfully and contains all 105 bundled PNG icons.

---

## ✅ Checklist

- [x] Code formatted and pre-commit checks completed where environment permitted
- [ ] Documentation updated (not applicable)

---

## 📓 Notes (optional)

No scheduled workflow included. Missing or intentionally skipped catalog entries retain the existing UI fallback.


msureshkumar88 (approved):
Nice work on this one — the scope here is genuinely minimal: no backend or schema changes were needed since `CatalogServer.logo_url` and the `/static` mount already existed, so this PR just does the data/tooling prep correctly rather than reaching for unnecessary changes. Good SSRF hardening in `fetch_catalog_icons.py` too (HTTPS-only, private-IP rejection re-checked per redirect hop, size-capped streaming, path-safe filenames).

**Verified:**
- All 10 new unit tests pass locally.
- `tldextract>=5.3.0` floor correctly matches the `top_domain_under_public_suffix` API used.
- No CI breakage — `Pillow`/`tldextract` land via the `dev` dependency group.
- Packaging (MANIFEST.in / pyproject package-data) is sound; wheel build includes all 105 icons.
- No breaking change, no Alembic migration needed (correct call — this is flat-file data, not persisted state).

**One functionally-impacting item worth a look before merge, or as an immediate fast-follow:**

The 14 `localhost`/stdio catalog entries (`azure-mcp`, `playwright-mcp`, `fast-time-server`, etc.) fall through to the generic MISS path instead of being classified as SKIP. Since these can never resolve a public icon, `--strict` will always exit non-zero against the real catalog — which makes the flag effectively unusable, and buries genuine fetch failures (e.g. `rube`, `google-maps`) in the same undifferentiated miss list as by-design skips. This also diverges from the linked issue's explicit "skip localhost/IP hosts" step. A small fix in `_fetch_icon`/`generate_icons` to route non-HTTPS/local hosts to SKIP before the miss path would resolve it and make `--strict` meaningful again.

**Non-blocking, worth a follow-up issue:**
- DNS-rebinding TOCTOU: `_validate_public_https_url` resolves and validates the host, but `httpx.Client.stream()` re-resolves independently for the actual connection — pinning the validated IP would close that gap. Low severity given this only runs as a maintainer-triggered pre-release tool, but redirect targets are third-party-controlled so it's not purely theoretical.
- Test coverage on `scripts/fetch_catalog_icons.py` is ~46% (network-facing `_fetch`, `_icon_candidates`, and the `generate_icons` CLI flow — `--dry-run`/`--force`/`--strict` — are untested). Also worth noting the coverage gates (`--cov=mcpgateway`, diff-cover) don't scope over `scripts/` at all, so this file's coverage isn't enforced by CI either way — a repo-wide gap, not specific to this PR.
- `docs/docs/manage/catalog.md`'s existing `logo_url` examples are all external URLs; a short note about the new `make catalog-icons` / local-asset convention would help avoid confusion for future catalog entries.
- The issue proposed SVG-with-sanitization as the preferred format and a DuckDuckGo-before-favicon.ico fallback order; this PR ships PNG-only (arguably the safer minimal choice, since it avoids the SVG sanitization attack surface entirely) with favicon.ico tried first. Worth a one-line mention in the PR description or the linked issue so the deviation is documented, not silently different from spec.

Approving — nothing here blocks merge. I'll open a follow-up issue to track the `--strict`/localhost-SKIP fix and the other non-blocking items above.

prakhar-singh1928 (approved):

PR #6304, opened 2026-08-19
fix: remove deprecated PUT /admin/users endpoint and add sunset marker pre-commit hook

prakhar-singh1928:
# Pull Request

## 🔗 Related Issue

Closes #6270 
Closes #6298 

---

## 📝 Summary

The deprecated `PUT /admin/users/{user_email}` endpoint in `mcpgateway/routers/email_auth.py` passed its sunset date, causing `test_admin_get_update_delete_user` to fail in CI due to a time-bomb `assert datetime.now(...) < datetime(2026, ...)` inside the test suite.

This PR does two things:

1. **Removes the deprecated endpoint.** `update_user_deprecated` and its `update_user_delegate` helper are deleted. The delegate body was already inlined into `update_user` (PATCH) as documented in the delegate's own docstring. The `Response` import that was only needed by the deprecated endpoint is also removed. Associated `[#2754]`-guarded test blocks are removed from both `test_email_auth_router.py` and `scripts/test_email_auth_api.py`.

2. **Replaces the broken enforcement pattern.** The time-bomb `assert` inside a pytest test is inherently fragile — it will always eventually fail in CI regardless of code correctness. It is replaced by a new `check-sunset-markers` pre-commit hook (`.github/tools/check_sunset_markers.py`) that scans `*.py` files at commit time for `[#NNNN] remove after` / `[#NNNN] Code to be removed after` markers and fails immediately, preventing future sunset time-bombs from entering the codebase.

---

## 📏 Reviewability

- [x] This PR has one clear purpose
- [x] The linked issue is not labeled `triage`
- [x] Unrelated bugs or improvements are tracked in separate issues/PRs
- [x] Tests are included with the code they validate
- [x] If AI-assisted, I understand and can explain the generated changes

---

## 🏷️ Type of Change

- [x] Bug fix
- [x] Chore (deps, CI, tooling)

---

## 🧪 Verification

| Check | Command | Status |
|---|---|---|
| Sunset marker hook runs clean | `uv run .github/tools/check_sunset_markers.py` | ✅ exits 0 |
| Pre-commit hooks pass | `make pre-commit` | ✅ |
| Lint suite | `make lint` | ✅ |
| Unit tests (previously failing test now passes) | `make test` | ✅ |
| Coverage ≥ 80% | `make coverage` | ✅ |

---

## ✅ Checklist

- [x] Code formatted (`make black isort pre-commit`)
- [x] Tests added/updated for changes
[...]

PR #6398, opened 2026-08-24
tests: fix Playwright gateway OAuth and team flakiness

gcgoncalves:
## Summary
Fix Playwright test failures in gateway OAuth grant-type coverage and
team-conftest user/invite setup that broke after the gateway add form
stopped offering the password grant type and under parallel test load.

## Reproduction Steps
Run `tests/playwright/test_gateways.py::TestGatewaysPage` and
`tests/playwright/entities/test_gateways_extended.py::TestOAuthGrantTypeSwitching`
against a gateway build where the add-gateway form no longer lists
`password` as an OAuth grant type. Also run `tests/playwright/teams/`
suites in parallel — `create_test_user`/`invite_and_accept` intermittently
fail on transient 5xx responses.

## Root Cause
- Gateway add form dropped the `password` OAuth grant option (deprecated
  for new gateways, still supported when editing). Tests still asserted
  its presence on the add form and drove grant-type switching through
  add-form locators only, so they never covered the edit-modal path
  where password grant now lives.
- `tests/playwright/teams/conftest.py` posted directly via
  `admin_api.post`/`user_ctx.post` with no retry, so transient 5xx
  responses under parallel test execution caused hard failures.

## Fix Description
- Added edit-modal locators (`edit_modal_oauth_grant_type_select`,
  `edit_modal_oauth_username_input`, `edit_modal_oauth_password_input`)
  to `GatewaysPage`.
- Updated `test_gateways.py` to assert `password` is absent from the
  add-form grant-type options.
- Updated `TestOAuthGrantTypeSwitching` in `test_gateways_extended.py`
  to exercise password grant through the edit modal instead of the add
  form.
- Added `_post_with_retry` helper in `teams/conftest.py`: bounded retry
  (3 attempts, backoff) on 5xx only, so genuine 4xx client errors still
  fail immediately.

## Reviewability

## Verification
| Check       | Command                                                              | Status |
|-------------|-----------------------------------------------------------------------|--------|
| Gateway PW tests | `pytest tests/playwright/test_gateways.py tests/playwright/entities/test_gateways_extended.py` | pass |
| Teams PW tests   | `pytest tests/playwright/teams/`                                  | pass   |

## Checklist

### DIFF
--- a/tests/playwright/entities/test_gateways_extended.py
+++ b/tests/playwright/entities/test_gateways_extended.py
@@ -725,19 +725,29 @@ class TestOAuthGrantTypeSwitching:
         expect(gateways_page.oauth_scopes_input).to_be_visible()
 
     def test_password_grant_shows_username_password(self, gateways_page: GatewaysPage):
-        """Test that password grant type shows username and password fields."""
+        """Test that password grant type shows username and password fields.
+
+        Password grant is deprecated for new gateways (add form no longer
+        offers it) but remains available when editing an existing gateway.
+        """
         gateways_page.navigate_to_gateways_tab()
+        gateways_page.wait_for_gateways_table_loaded()
+        _skip_if_no_gateways(gateways_page)
 
-        gateways_page.auth_type_select.select_option("oauth")
+        _open_edit_or_skip(gateways_page, 0)
+        expect(gateways_page.edit_modal_name_input).not_to_be_empty(timeout=10000)
 
-        gateways_page.oauth_grant_type_select.select_option("password")
+        gateways_page.edit_modal_auth_type_select.select_option("oauth")
+        gateways_page.edit_modal_oauth_grant_type_select.select_option("password")
 
         # Username and password fields should be visible
-        expect(gateways_page.oauth_username_input).to_be_visible()
-        expect(gateways_page.oauth_password_input).to_be_visible()
+        expect(gateways_page.edit_modal_oauth_username_input).to_be_visible()
+        expect(gateways_page.edit_modal_oauth_password_input).to_be_visible()
+
+        gateways_page.close_edit_modal()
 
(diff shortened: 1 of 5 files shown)


marekdano (dismissed):
LGTM 🚀

marekdano (approved):
LGTM

msureshkumar88 (approved):
Reviewed the diff — test-only change (Playwright specs, page-object locators, teams conftest), no `mcpgateway/` app code, no schema/migration changes.

Verified against current source rather than taking the description at face value:
- Add-gateway form (`#oauth-grant-type-gw`) genuinely has no `password` option in `admin.html`, while the edit form (`#oauth-grant-type-gw-edit`) does — the new edit-modal locators (`#oauth-username-gw-edit`, `#oauth-password-gw-edit`) resolve to real DOM ids, and the new assertions are read-only (edit modal is closed via Cancel, never saved).
- `PUT → PATCH` in `test_user_management.py` matches the live route: `email_auth.py` only registers `PATCH /admin/users/{email}` (the `PUT` route was removed in #6304). Confirmed no other file still calls `PUT` on this same endpoint.
- `_post_with_retry` in `teams/conftest.py` only retries 5xx, breaks immediately on 4xx — matches its stated intent, and the accept-invitation duplicate-race handling (`_is_team_member` fallback) is a nice touch.

One minor asymmetry: the invite-POST retry (`invite_and_accept`) doesn't have the same duplicate-fallback as the accept-POST a few lines below it. Traced the exposure — `SessionLocal(expire_on_commit=False)` and the try/except around email delivery rule out most of the paths I initially worried about, so it'd only bite on an infra-level 502/504 after the server already committed. Non-blocking; worth a small follow-up if it ever shows up in CI flakiness.

PR body's root causes don't mention the `test_user_management.py` fix, but the fix itself is correct and necessary — just a documentation gap, not a code issue.

PR #6399, opened 2026-08-24
fix: harden local A2A egress handling

gandhipratik203:
## Summary
- Harden local A2A invocation egress handling.
- Add focused regression coverage for local A2A invocation paths.

### DIFF
--- a/mcpgateway/services/a2a_service.py
+++ b/mcpgateway/services/a2a_service.py
@@ -28,6 +28,7 @@ from sqlalchemy.orm import Session
 
 # First-Party
 from mcpgateway.cache.a2a_stats_cache import a2a_stats_cache
+from mcpgateway.common.validators import SecurityValidator
 from mcpgateway.config import settings
 from mcpgateway.db import A2AAgent as DbA2AAgent
 from mcpgateway.db import A2AAgentMetric, A2AAgentMetricsHourly, A2ATask, EmailTeam
@@ -37,9 +38,10 @@ from mcpgateway.db import Tool as DbTool
 from mcpgateway.observability import create_span, set_span_attribute, set_span_error
 from mcpgateway.plugins.utils import build_request_extensions, record_plugin_metrics
 from mcpgateway.schemas import A2AAgentAggregateMetrics, A2AAgentCreate, A2AAgentMetrics, A2AAgentRead, A2AAgentUpdate
-from mcpgateway.services.a2a_protocol import prepare_a2a_invocation
+from mcpgateway.services.a2a_protocol import prepare_a2a_invocation, prepare_pinned_a2a_invocation
 from mcpgateway.services.base_service import BaseService
 from mcpgateway.services.encryption_service import protect_oauth_config_for_storage
+from mcpgateway.services.http_client_service import get_isolated_http_client
 from mcpgateway.services.logging_service import LoggingService
 from mcpgateway.services.metrics_cleanup_service import delete_metrics_in_batches, pause_rollup_during_purge
 from mcpgateway.services.observability_service import current_trace_id, ObservabilityService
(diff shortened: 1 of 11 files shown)


msureshkumar88 (commented):
Nice work closing this one — reusing the existing `validate_url_for_connection_pinning`/pinning primitive instead of writing a bespoke A2A-specific check is the right call, and wiring it into all three local A2A egress sites (`invoke_agent`, the A2A-as-tool path, and `_call_a2a_agent`) rather than just the one the report called out is good thoroughness. Fail-closed-before-I/O, disabled redirects, and keeping the correct SNI/Host while pinning to the resolved IP are all the details that are easy to get subtly wrong, and they're handled correctly here. No blocking issues from my side — a few non-blocking follow-ups worth considering:

**Test coverage**
1. The fail-open branch in `prepare_pinned_a2a_invocation` (when `ssrf_protection_enabled=False`) doesn't have a test — all the new tests in `test_a2a_protocol.py` set it to `True`. Cheap to add.
2. All the new tests mock either `prepare_pinned_a2a_invocation` or `validate_url_for_connection_pinning`'s return value directly, so there's no test that exercises the real resolver end-to-end and proves the rebinding window is actually closed. `tests/security/test_gateway_test_ssrf_protection.py::test_dns_rebinding_protection` already has the pattern for this (patches `socket.getaddrinfo` to return different IPs across calls) — an A2A analogue in `tests/security/` would be a nice, fairly cheap regression guard given the severity here.

**Docs / config**
3. Worth a line in the PR description or release notes: registration-time SSRF validation was previously a one-time check, and invocation now re-validates on every call. Agents whose endpoint hostname resolves differently now than at registration time (DNS drift / split-horizon resolvers) will start failing invocation where they didn't before — that's the intended fix, but ops teams upgrading should know to expect it.
4. `SSRF_DNS_FAIL_CLOSED` doesn't affect this new pinning path — `_resolve_hostname_for_connection_pinning` keys off `ssrf_protection_enabled` only, not `ssrf_dns_fail_closed`. That's defensible (you can't pin without a resolved address), but it's a pre-existing gap in the shared helper that's now load-bearing for A2A too, and worth a doc note so it doesn't look like a bug later.
5. Minor: the DNS timeout for this path comes from `settings.gateway_test_dns_timeout`, whose docstring says it's for `/admin/gateways/test`. Now governs A2A invocation latency too — an alias or updated description would help future readers.

None of the above blocks merging as far as I can tell — happy to help with the two tests if useful.

msureshkumar88 (dismissed):
Approving — this closes the DNS-rebinding SSRF gap correctly by reusing the existing connection-pinning primitive across all three local A2A egress sites, fail-closed before I/O, with SNI/Host handled correctly. No blocking issues found. Follow-up suggestions (test coverage for the fail-open branch, an end-to-end rebinding regression test, and a few doc notes) are in my earlier comment — none of them need to hold up this merge.

mcpgateway/services/a2a_service.py
```diff
@@ -2423,12 +2424,27 @@ async def invoke_agent(
...
+                        prepared.sanitized_endpoint_url,
+                        correlation_id,
+                        validation_reason,
+                    )
+                    raise A2AAgentError(error_message) from validation_error
+
+                # Make HTTP request to the agent endpoint using an isolated SSRF-sensitive client.
                 # First-Party
-                from mcpgateway.services.http_client_service import get_http_client  # pylint: disable=import-outside-toplevel
+                from mcpgateway.services.http_client_service import get_isolated_http_client  # pylint: disable=import-outside-toplevel
```
prakhar-singh1928: can you move all imports to top of file

prakhar-singh1928 (approved):
LGTM

msureshkumar88 (approved):

PR #6405, opened 2026-08-25
fix: add MCP handshake test endpoint for virtual servers

### LINKED ISSUE #6370: [Bug] Virtual server "Test connection" can never succeed - /v1/mcp-servers/test-handshake is scoped to registered gateways, not virtual servers
## Summary

  `POST /v1/mcp-servers/test-handshake` (added in #5934 for #5649) is scoped to **registered gateways**, not **virtual servers**. #5650 wires the virtual server details drawer's "Test connection" tab to this same endpoint, but its target-allowlist only ever matches rows in the `gateways` table — a virtual server's own endpoint is never eligible. As implemented, #5650's "Test connection" tab can **never succeed** against a real virtual server, regardless of URL, host, or config.

  ## Root cause

  `mcpgateway/routers/mcp_servers_router.py` mounts under `prefix="/v1/mcp-servers"`, tagged `"MCP Servers"` — this whole router is scoped to the *MCP Servers (gateway/federated source)* domain, not *Virtual Servers*:

  ```python
  """MCP Servers REST API router.

  Endpoints:
      POST /v1/mcp-servers/test            — Test MCP server / gateway connectivity
      POST /v1/mcp-servers/test-handshake  — Test whether a server URL speaks MCP
  """
  router = APIRouter(prefix="/v1/mcp-servers", tags=["MCP Servers"])
  ```

  Both endpoints call `_validate_gateway_test_target()` (`mcpgateway/services/gateway_service.py:7366`), whose SSRF allowlist is built exclusively from registered gateways:

  ```python
  # gateway_service.py:7396
  ```

  When `gateway_test_allow_registered_only` (`mcpgateway/config.py:765`, **default `True`**) is set, this is the *entire* allowlist. A virtual server's own transport endpoint (`/servers/{id}/mcp`) is never a row in the `gateways` table, so it can never match — the request fails closed with a generic `"...host is permitted by your test policy"` error (`gateway_service.py:7739`) before any network call is attempted.

  ## Reproduction

  1. Create a virtual server with at least one associated tool/resource/prompt.
  2. Open its details drawer → **Test connection** tab (contextforge-web-ui#53/#55-adjacent work, tracking IBM/mcp-context-forge#5650).
  3. Run the test against the prefilled URL (`{origin}/servers/{id}/mcp`, from `getVirtualServerEndpoint()`).
  4. Response is always:
     ```json
     {
       "success": false,
       "failureClass": "transport",
       "error": "The MCP server URL is not allowed for testing. Confirm the URL is correct and the host is permitted by your test policy."
     }
     ```
     with no network attempt made — confirmed via direct `curl` against the backend and via `gateway_test_allow_registered_only`'s DB-only allowlist query.

  ## Impact

  - #5650's "Test connection" tab renders correctly and calls the endpoint correctly, but the feature is non-functional end-to-end for its stated purpose.
  - Frontend unit tests (mocked at the API boundary) don't catch this — the gap only surfaces against the real backend.

  ## Suggested fix

  One of:
  - A dedicated endpoint scoped to virtual servers, e.g. `POST /v1/servers/{id}/test-handshake`, that trusts the server's own known ID/endpoint instead of validating an arbitrary URL against the gateway allowlist (a virtual server's own endpoint isn't an SSRF risk — it's the gateway's own process).
  - Extend `test_gateway_handshake` / `_validate_gateway_test_target` to accept a virtual-server ID as an alternate, already-trusted target alongside the existing URL-allowlist path.

  ## Related

  - #5650 (virtual server Test connection tab — blocked by this)
  - #5649 (MCP handshake test mode for gateways — this endpoint's original scope)
  - #5934 (backend PR that added `/v1/mcp-servers/test-handshake`)

marekdano:
Closes #6370

### Summary

  `POST /v1/mcp-servers/test-handshake` is scoped to registered gateways: its SSRF allowlist only ever matches rows in the `gateways` table, so a virtual server's own `/servers/{id}/mcp` endpoint can never pass it. The "Test connection" panel for virtual servers reused that endpoint and could therefore never succeed, regardless of URL, host, or config — the SSRF gate in `SecurityValidator.validate_gateway_test_url` also unconditionally blocks loopback/private targets, so no allowlist workaround was possible either.

  Adds `POST /servers/{server_id}/test-handshake` (aliased at `/v1/servers/...` and `/v1/virtual-servers/...`), scoped to virtual servers instead of arbitrary URLs:

  - Derives the target from the server's own ID rather than validating a caller-supplied URL, so the gateway-test SSRF allowlist doesn't apply — the target is never anything but the process's own transport.
  - Dispatches in-process via `httpx.ASGITransport` rather than a real network round trip, so it works identically behind any reverse proxy/LB and always lands on the worker handling the request.
  - Reuses the caller's own forwarded credentials by default (bearer token or session cookie), with an optional header override in the request body, so auth, RBAC, and MCP negotiation are exercised for real.
  - Adds the `servers.read` permission mapping for the new route (previously unmapped paths default-deny under token scoping).

  The existing gateway `/v1/mcp-servers/test-handshake` endpoint is unchanged.

  ### 🧪 Verification - testing

Tested directly against a running dev server (`make dev`), not just mocks.
  ```bash
  export TOKEN=$(python -m mcpgateway.utils.create_jwt_token \
    --username admin@example.com --exp 60 --secret <JWT_SECRET_KEY>)
  ```

  2. Create a virtual server to test against:
  ```bash
  curl -s -X POST http://localhost:8000/servers \
    -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" \
    -d '{"server": {"name": "smoke-test-server"}}'
  # note the returned "id" as $SERVER_ID
  ```

  3. Run the handshake against all three path forms:
  ```bash
  curl -s -X POST "http://localhost:8000/servers/$SERVER_ID/test-handshake" \
    -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" -d '{}'
  curl -s -X POST "http://localhost:8000/v1/servers/$SERVER_ID/test-handshake" \
    -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" -d '{}'
  curl -s -X POST "http://localhost:8000/v1/virtual-servers/$SERVER_ID/test-handshake" \
    -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" -d '{}'
  ```
  Expected: `"success": true`, a real `negotiationPath`/`protocolVersion`/`serverName`, and `"credentialSource": "session"` on all three.

  4. Failure paths:
     - No `Authorization` header → `401`.
     - Disable the server first (`POST /servers/$SERVER_ID/state?activate=false`), then retest → `200` with `"success": false` and an error naming the server as disabled (not a crash/hang).
     - A token scoped without `servers.read` → `403`.

  5. Regression check — confirm the existing gateway endpoint is unaffected:
  ```bash
  curl -s -X POST "http://localhost:8000/v1/mcp-servers/test-handshake" \
    -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" \
    -d '{"base_url": "http://127.0.0.1:9/"}'
  ```
  Expected: unchanged SSRF-rejection behavior.

  6. Clean up the test server: `DELETE /servers/$SERVER_ID`.

### DIFF
--- a/mcpgateway/main.py
+++ b/mcpgateway/main.py
@@ -4510,6 +4510,15 @@ async def test_server_mcp_handshake(
         auth_token = request.cookies.get("jwt_token") or request.cookies.get("access_token")
     if auth_token:
         forwarded_headers["Authorization"] = f"Bearer {auth_token}"
+    elif settings.trust_proxy_auth:
+        # No bearer/cookie credential: this caller may have been authenticated via the
+        # configured trusted-proxy identity header instead (TRUST_PROXY_AUTH_DANGEROUSLY).
+        # Forward that header's *own already-verified value from this request* -- never
+        # anything from the request body -- so the probe reflects the same identity that
+        # got the caller past auth here.
+        proxy_user = request.headers.get(settings.proxy_user_header)
+        if proxy_user:
+            forwarded_headers[settings.proxy_user_header] = proxy_user
 
     return await test_server_handshake(server.id, server.name, server.enabled, body, forwarded_headers)
 
--- a/mcpgateway/main.py
+++ b/mcpgateway/main.py
@@ -145,6 +145,7 @@ from mcpgateway.schemas import (
     CursorPaginatedServersResponse,
     CursorPaginatedToolsResponse,
     GatewayCreate,
+    GatewayHandshakeResponse,
     GatewayImpactPreview,
     GatewayRead,
     GatewayRefreshResponse,
@@ -165,6 +166,7 @@ from mcpgateway.schemas import (
     RootUpdate,
     RPCRequest,
     ServerCreate,
+    ServerHandshakeRequest,
     ServerRead,
     ServerUpdate,
     TaggedEntity,
@@ -189,6 +191,7 @@ from mcpgateway.services.gateway_service import (
     GatewayLookupConflictError,
     GatewayNameConflictError,
     GatewayNotFoundError,
+    test_server_handshake,
 )
 from mcpgateway.services.import_service import ConflictStrategy, ImportConflictError
 from mcpgateway.services.import_service import ImportError as ImportServiceError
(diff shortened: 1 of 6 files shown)


Lang-Akshay (changes requested):
## Blocking Changes

| 1 | Security | `mcpgateway/services/gateway_service.py` (also `mcpgateway/schemas.py`, `mcpgateway/main.py`) | 7698–7705; request model at 4944; forwarding at 4498–4503 | `ServerHandshakeRequest.headers` accepts arbitrary caller-controlled headers and `test_server_handshake()` forwards every header except `Host` into an in-process ASGI request whose client address is deliberately loopback. With `TRUST_PROXY_AUTH_DANGEROUSLY=true`, a caller can provide the configured trusted identity header (for example `X-Authenticated-User`) and select the nested principal; `X-Forwarded-For` can similarly reach trusted proxy handling and bypass token IP restrictions. The outer `servers.read` check does not bind the nested request to the outer principal. CWE-287/CWE-807; ContextForge trusted-header/auth isolation invariant violation. | Do not forward an unrestricted body header map. Use a strict allowlist for the intended authentication header(s), preferably bind the probe to the already-authenticated outer identity, and reject/strip proxy identity, all `X-Forwarded-*`/`X-Real-IP`, `X-ContextForge-*`, session/routing, `Mcp-Session-Id`, and hop-by-hop headers. Honor the configured authentication-header name instead of hardcoding only `Authorization`, and add deny-path coverage for proxy-identity and spoofed-IP headers. |

Lang-Akshay (changes requested):
## Blocking Changes

| 1 | Comment / Quality | `mcpgateway/main.py` (also `mcpgateway/schemas.py`, `mcpgateway/services/gateway_service.py`) | `main.py:4499-4504`; `schemas.py:4965-4970`; `gateway_service.py:7858,7942` | The handler reads the configured authentication header with `get_auth_header_value()` but always forwards the resulting token under literal `Authorization`. The nested Streamable HTTP auth also honors `settings.auth_header_name`, so a supported custom `AUTH_HEADER_NAME` makes the probe return an authentication failure even though the outer endpoint succeeded. The prior required reviewer comment explicitly called out this contract. | Forward the bearer under `_resolve_auth_header_name(settings)` and make the explicit-header allowlist/schema honor that configured authentication header while continuing to reject proxy identity, `X-Forwarded-*`, `X-Real-IP`, `X-ContextForge-*`, session/routing, and hop-by-hop headers. Add deny-path/coverage cases for custom auth headers and spoofed headers. |

msureshkumar88 (approved):
Nice work here, and thanks for the thorough response to the earlier rounds of feedback — both the header-forwarding restriction and the `AUTH_HEADER_NAME` fix are verified correct against the current code.

A few things I liked in particular:
- Reusing `_probe_mcp_session` between the gateway and virtual-server handshake paths instead of duplicating the initialize/list-components logic — verified it's a pure extraction with no behavior change on the existing gateway endpoint.
- Dispatching in-process via `httpx.ASGITransport` rather than a real network round trip is the right call: it sidesteps SSRF entirely (no DNS resolution, no outbound socket) and avoids the "test lands on a different worker than the one serving the request" class of bug in multi-worker/LB deployments.
- The header allowlist/denylist work in `_handshake_form_header_allowlist` / `_is_handshake_header_denied` has solid deny-path test coverage (proxy-identity, `X-Forwarded-*`, `Mcp-Session-Id`, hop-by-hop, custom `AUTH_HEADER_NAME`) — matches the CWE-287/CWE-807 concern raised in the earlier review.

Confirmed no schema/migration changes are needed (correct — this is a stateless probe), no breaking changes (purely additive route + additive `Literal` value), and all 30 new tests pass locally. CI is green.

A couple of optional follow-ups, none blocking:
- `SERVER_CRUD_PREFIXES` in `test_main.py` doesn't include `/v1/servers`, even though that path form is exercised manually in the PR description — worth adding to the parametrized test for coverage parity.
- A deny-path regression test for `servers.read`-less tokens (401/403) and a `token_scoping.py` pattern-ordering regression test would round out the security test coverage per the repo's testing conventions.
- A `tests/live_gateway/` black-box test would close the loop per the repo's live-gateway testing policy, if you have bandwidth.

Approving — the fix matches the issue's own suggested remediation, is correctly scoped to backend-only, and the security hardening holds up under verification.

Lang-Akshay (approved):
LGTM ✅

